PR #288, opened 2016-12-21
Updated nmap.tb

doriclazar:
Fixes: "live packet capture not supported on this system."
Compiles without unnecessary zenmap.


thedamian:
Tested it with:
sudo nmap -sS 192.168.0.1
and found my router is printer capable. Very nice.

MERGED

PR #289, opened 2016-12-23
calculate SHORTARCH from ARCH - fix for missing getconf

minektur:
remove getconf as a dependency for install, case statement to figure out long size from arch.


thedamian:
But you're not using SHORTARCH

or do you use it in some packages?

PR #291, opened 2016-12-23
Fixed a typo in crew script

jam7:
uname -m returns armv7l instead of arm7l.
This doesn't cause failure since default value is 32 but it's good to fix.

### DIFF
--- a/crew
+++ b/crew
@@ -22,7 +22,7 @@ case ARCH
     SHORTARCH="32"
   when "x86_64"
     SHORTARCH="64"
-  when "arm7l"
+  when "armv7l"
     SHORTARCH="32"
   else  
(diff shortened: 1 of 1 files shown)


crew
```diff
@@ -22,7 +22,7 @@ case ARCH
     SHORTARCH="32"
   when "x86_64"
     SHORTARCH="64"
-  when "arm7l"
+  when "armv7l"
```
finwo: Are you sure this is a typo, and not an extra (more common) type?

lyxell:
Thanks!

minektur:
I was remembering wrong on what I put in there - I dont have a arm chromebook in dev mode I can test on and was.... shooting from the hip.

Issue #290, opened 2016-12-23
go fails to build


mattddowney:
```
Building from source, this may take a while...
##### Building Go bootstrap tool.
cmd/dist
ERROR: Cannot find /home/chronos/user/go1.4/bin/go.
Set $GOROOT_BOOTSTRAP to a working Go tree >= Go 1.4.
go failed to install: `TMPDIR=/usr/local/tmp ./make.bash` exited with 1
```

Go needs go to build as described here:
https://golang.org/doc/install/source#go14

Looks like this could be fixed either with a binary package, or by creating another package with the go 1.4 toolchain as a dependency.

mattddowney:
Resolved by #292

PR #292, opened 2016-12-24
Bootstrap Go

mattddowney:
Satisfies #290 

Building any version of go > 1.4 requires go itself. This PR allows compilation of the latest go package for systems without an existing go binary by requiring a dependency, go_bootstrap (essentially go v1.4, which has the ability to be built with gcc).


thedamian:
Having a language fepend on itself is crazy! 

I like the solution.

PR #293, opened 2016-12-24
add package for nethack4

minektur:
Great, old game.... "nethack4" fork of nethack.

### DIFF
--- a/packages/nethack4.rb
+++ b/packages/nethack4.rb
@@ -25,7 +25,6 @@ class Nethack4 < Package
   
   def self.install
     target=CREW_BREW_DIR+"nethack4-4.3-beta2/build"
-    system "mkdir -p " + target
     Dir.chdir target  do
       #install in destdir so package manager can keep track
       system "/usr/local/bin/perl ../aimake --install-only -i #{CREW_DEST_DIR}/usr/local/ --directory-layout=prefix --without=gui"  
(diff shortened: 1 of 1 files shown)


thedamian:
You're making a directory thr build lrocess and in the install process? Isnt the system going to move the files for you or are you doing that in the temp dirsgotry?

thedamian:
Typo on my phone. Meant directory.

minektur:
I am just following their build instructions - make an empty directory,
call aibuild by relative path, specifying on the command line where you
want it to be installed.  Since we install to a temp directory and then it
gets moved into the right place, you have to do it in two steps - the first
step is to build with the eventual install directory, but not install, and
then cal lit again with a special "just do the install part to where I
specify without rebuilding".

Doing it this way gets the rpath set correctly (to
/usr/local/[lib|bin|etc]) but lets you do the ./aibuild install.... stuff
into /usr/local/tmp/crew/dest.


So the 'build' block builds it as if it is going into /usr/local and the
'install' block temporarily installs into /usr/local/tmp/crew/dest and then
crew does the rest - making the manifest files
(/usr/local/etc/crew/nethack4.filelist and nethack4.dirlist) and then puts
everything into /usr/local/

You could just grab the one file and try an install - check your /usr/local
filesystem before and after and you can see what files change.

minektur:
Oh, and mkdir -p is idempotent...

thedamian:
Alright. since it's an independent game. it's merged. 

thanks for these great golden nudgets as always @minektur

PR #294, opened 2016-12-24
New arm binary including totally recompiled gcc

jam7:
Hi, this time I update all binaries including gcc for ARM architecture.  Therefore, an annoying symbolic link from ~/Downloads/tools is no more necessary.  I also remove unnecessary binaries from ARM packages which I was thinking they are necessary wrongly.  Afther this PR, ARM has very similar package structure to other x86 and x86_64 architectures.

I put all binaries to dropbox.com, but I appreciate if anyone migrate them to github or other binary distributing site.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -99,27 +99,6 @@ for file in git zlibpkg libssh2 perl curl expat gettext python readline ruby bui
   wget -N -c $URL/packages/$file.rb
 done
 
-#install gcc on arm only.  It requires special treatments
-case "$architecture" in
-"armv7l")
-  echo y | crew install gcc
-  if [ ! -d $HOME/Downloads/tools ]; then
-    mkdir $HOME/Downloads/tools
-    mkdir $HOME/Downloads/tools/bin
-    mkdir $HOME/Downloads/tools/lib
-    mkdir $HOME/Downloads/tools/libexec
-    mkdir $HOME/Downloads/tools/share
-    ln -s /usr/local/armv7a-cros-linux-gnueabi $HOME/Downloads/tools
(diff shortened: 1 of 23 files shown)


jam7:
In addition, it is possible to try it from https://github.com/jam7/chromebrew by typing:

  $ wget -q -O - https://raw.github.com/jam7/chromebrew/master/install.sh | bash

bfayers:
@jam7 If I install this can i safely remove ~/Downloads/tools ?

jam7:
Basically, yes.  You may need to recompile what you compiled by old ~/Downloads/tools compilers.  If those are in crew packages, crew remove PKG and crew install PKG performs the recompilation.

thedamian:
Looks like a fantastic PR.

Loving it. Its merged!!!

bfayers:
It's not important, but I'd like to know how you got GCC down to a 65MB package.

jam7:
I didn't do anything special, so I'm not sure how.  Maybe because I used chromium gcc fork?

Comparing to previous gcc for ARM, previous one contains glibc and others in gcc package.  This time, I divided them.  Comparing to x86 or x86_64 gcc, I'm not sure why.

If you have an interest or want to try, I created a repository, https://github.com/jam7/chrome-cross, cross compile script of toolchains.  I used this to make toolchains for ARM.  I also would like to ask others to use this toolchains for x86 and x86_64 in near future if toolchains works well.

lyxell:
@jam7 Great work! It would be fantastic if we could use this for x86 as well!

jam7:
@bfayers: I forgot to mention.  I'm using xz instead of gz.  It reduces the size very well.

PR #300, opened 2016-12-27
Add package for jq

cstrouse:
This adds jq (1.5) which is a command line filter for processing JSON. Configure warns about missing Ruby dependencies which I'm ignoring since we don't have proper manpage support yet and I don't see a need for bringing in additional bloat when disk space is limited for us to begin with. Also, regex support requires oniguruma which will come shortly.

Tested sucessfully on Samsung Chromebook 3 (XE500C13-K01US).


lyxell:
I added a switch to disable the docs and omit the warning.

Issue #146, opened 2016-02-25
Can't install vim


Tenn1518:
I installed chromebrew today and it's been working well for me so far. Nano installed without a hitch. However, vim is not installing for me. I attempted an install with "crew install vim", and here is a log of what happened. I'm using an x86-64 Dell Chromebook 11.

chronos@localhost ~ $ crew install vim
Found vim, version 7.4
Following packages also need to be installed: 
binutils gmp mpc mpfr glibc gcc make linuxheaders buildessential readline zlibpkg libsigsegv diffutils bz2 ncurses perl python27 ruby 
Do you agree? [Y/n]
y
Proceeding...
Found binutils, version 2.23.2
Package binutils already installed, skipping...
Found gmp, version 5.1.2
Package gmp already installed, skipping...
Found mpc, version 1.0.1
Package mpc already installed, skipping...
Found mpfr, version 3.1.2
Package mpfr already installed, skipping...
Found glibc, version 2.17.90-baseline
Package glibc already installed, skipping...
Found gcc, version 4.8.1-baseline
Package gcc already installed, skipping...
Found make, version 3.82
Package make already installed, skipping...
Found linuxheaders, version 3.4.0
Package linuxheaders already installed, skipping...
Found buildessential, version 1.0
Package buildessential already installed, skipping...
Found readline, version 6.2
Package readline already installed, skipping...
Found zlibpkg, version 1.2.8
Package zlibpkg already installed, skipping...
Found libsigsegv, version 2.10
Package libsigsegv already installed, skipping...
Found diffutils, version 3.3
Package diffutils already installed, skipping...
Found bz2, version 1.4.17
Package bz2 already installed, skipping...
Found ncurses, version 5.9
Package ncurses already installed, skipping...
Found perl, version 5.16.3
Package perl already installed, skipping...
Found python27, version 2.7.10
Package python27 already installed, skipping...
Found ruby, version 2.0.0p247-chromeos1
Package ruby already installed, skipping...
Found vim, version 7.4
No precompiled binary available for your platform, downloading source...
WARNING: timestamping does nothing in combination with -O. See the manual
for details.

--2016-02-24 20:28:47--  ftp://ftp.vim.org/pub/vim/unix/vim-7.4.tar.bz2
           => 'vim-7.4.tar.bz2'
Resolving ftp.vim.org... 192.87.102.42, 192.87.102.43
Connecting to ftp.vim.org|192.87.102.42|:21... connected.
[...]

PR #301, opened 2016-12-27
Disabling selinux

jam7:
solve #146 problem on arm and x86_64.  I'm not sure why configure script think chromebook should use selinux, though...


jam7:
I made issue #303 to investigate situation.  It may be better to merge this after more discussion.

lyxell:
Looks good to me.

PR #302, opened 2016-12-27
Add package for netcat

cstrouse:
This adds GNU netcat (0.7.1). netcat is a utility for managing network connections.

Tested sucessfully on Samsung Chromebook 3 (XE500C13-K01US).


cstrouse:
Actually we may not need this. I was poking around a bit in the Chromium git repos and nc is enabled in the busybox config. Do we give people both options?

thedamian:
This installed perfectly on my chromebook pixel. I'm going to let it in.

Can I ask you rename the package to netcat.rb because I don't know if it will be immediately known to people looking for this package that it's called like the command itself is.

@cstrouse  I couldn't test what you're saying but uninstalling the package didn't give me NC back. I couldn't find it in my system except the version installed by crew by this script.
What am I missing that you say is already in our system? and if it's there, how do we find it?

ALso it's good to have a version on chromebrew even if it's already installed so that we can upgrade it when a new version comes out.

cstrouse:
@thedamian Sure I'll rename it to netcat. I forgot that GNU spells it out while the original hobbit tool was always simply nc.

thedamian:
Youve renamed the file but not the class. They must match  

The command will still be nc, so dont worry about that.

cstrouse:
@thedamian That wasn't intentional. They match now.

thedamian:
Flawless! Merged

PR #305, opened 2016-12-28
Disable selinux support

cstrouse:
Missing selinux.h was causing emacs abort the build. ChromeOS doesn't currently use selinux so it's not needed anyway.

### DIFF
--- a/packages/emacs.rb
+++ b/packages/emacs.rb
@@ -11,7 +11,7 @@ class Emacs < Package
   depends_on "automake"
 
   def self.build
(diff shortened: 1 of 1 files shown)


thedamian:
It sounded like several people were asking for adding the .h just like we did with open ssl.

but a fix now is better than a promise. 

It's merged!

cstrouse:
On my machine it appears that these are all present. I did run dev_install at one point though so I suppose it could have come from that.

https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/master/chromeos-base/busybox-config/files/busybox

PR #306, opened 2016-12-29
Add php7 package

uberhacker:
Compiled successfully on Google Chrome OS:
```
Platform 8872.73.0 (Official Build) stable-channel kip
```

### DIFF
--- a/packages/php7.rb
+++ b/packages/php7.rb
@@ -1,13 +1,12 @@
 require 'package'
 
 class Php7 < Package
-  version '7.0.14'
-  source_url 'http://php.net/get/php-7.0.14.tar.gz/from/this/mirror' # software source tarball url
-  source_sha1 '7d098167c63446011e87b1d1578f2f3ac0bab01a'  # source tarball sha1 sum
-  
-  depends_on 'nginx'
+  version '7.1.0'
+  source_url 'http://php.net/distributions/php-7.1.0.tar.xz' # software source tarball url
+  source_sha1 'c74c920256b9c6873bae696fbb0ec14a02dc8495'  # source tarball sha1 sum
+
   depends_on 'libxml2'
-  
+
   def self.build                                                  # self.build contains commands needed to build the software from source
     system "./configure"
     system "make"                                                 # ordered chronologically
--- a/packages/php7.rb
+++ b/packages/php7.rb
@@ -1,0 +1,19 @@
+require 'package'
+
+class Php7 < Package
+  version '7.0.14'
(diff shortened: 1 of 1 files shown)


thedamian:
I tried to get this to work myself and failed. (there was a requirement to a package that would "BORK" my system. 

but with your code I was able to create a .rb file that works.

Unfortunately the way you have it won't work.

1) You have an older version of the stable code (you have 7.0.14 and they're at 7.1)
2) (And more importnatly) Your url has /mirror at the end, and with our latest changes that allows tar or zip files we don't yet make the assumption what the file is, or what it's called (always assume it's the last thinig after the last / (and hence it thinks it's mirror"

HERE: Since I don't just want to fix your PR and check it in, here's the changes that will make work.
  version '7.1.0'
  source_url 'http://php.net/distributions/php-7.1.0.tar.xz' # software source tarball url
  source_sha1 'c74c920256b9c6873bae696fbb0ec14a02dc8495' 

ALSO I DON'T think that nginx should be a requirement.

Although I agree PHP without a web server is silly. it's still something the user MIGHT want. and also if they have ANOTHER web server (say apache) Magically installed, then they certainly DO NOT want NGINX (probably)

PLEASE REMOVE NGINX as one of the requirements.

After these changes it worked perfectly on my box. and I will merge it.

cstrouse:
Just use the built-in one that Facebook claimed in either 6 or 7 was fast enough to run production services on. I believe you use -S host:port

uberhacker:
@thedamian: One of the primary reasons I wanted PHP was so I could use Composer.  However, my main motivation was to use it via the CLI, so your point about Nginx not being a requirement is definitely valid.  The decision to use version 7.0.14 was based on the fact that 7.1.0 is relatively new and has some major changes but I guess we should always be moving forward anyway.  @cstrouse: Excellent point about the baked-in web server.  I'm definitely going to give that a try once I get this working the way I need it.  I imagine as more library packages are added, we will be able to start enabling more configure options.  As soon as my compile is finished and I test it locally, I'll update the PR.  Thanks for your quick response.

thedamian:
ah! the old "should we use the new version or the stable"

I'm a node developer (by trade) so I know how you feel. Node 7 kills most of the work i've done. So I understand.

I'm NOT a php developer. Just saw php.net call 7.1 their "Current Stable" and the 7.0 the "old stable"

When node does it at least they call 6 their "LTS" and 7 their "current" version.

I appreciate you did what I asked and will merge it asap. But if you really feel we should have 7.0 then just do another PR and I'll accept that too 

BUT make sure the url ends with .gz or .xz or .zip or else the crew installer gets cranky.

PR #308, opened 2016-12-30
Add configure options to php7 package

uberhacker:
- This adds a sane number of configure options to get Composer working
      - Many of these options are requirements for other PHP projects as well


uberhacker:
Closing to replace with PR without the Libtiff package commit

PR #309, opened 2016-12-30
Add configure options to php7 package

uberhacker:
- This adds a sane number of configure options to get Composer working
  - Many of these options are requirements for other PHP projects as well
  - The openssl package (a necessary dependency so the `--with-openssl` configure option can compile) was not compatible with my Chrome OS version, but I removed the package via `sudo crew remove openssl` after `crew install php7` and then php commands with https requests started working again.


thedamian:
Check if it requires openssl or openssl_devel which are the headers.

Or both.

uberhacker:
Only openssl package was necessary to compile successfully.  You are welcome to test it out and see if you can install [Composer](https://getcomposer.org/download/).  Worked for me (tm).

thedamian:
Thats a deal. Ill try it out myself in 15min and will merge it if works.

uberhacker:
If your system has a different openssl version than the package version (like mine does), don't forget to remove the package via `sudo crew remove openssl`, otherwise, the first Composer installation request to https might fail.

thedamian:
It didn't install on my box. and I can't understand what doing a remove would do since the openssl is a dependancy and will be installed again before.

I'm testing an update top openssl to 1.0.2j (the latest 1.0.2) and hopefully this will fix the issue for THIS (the php7 package)

thedamian:
Ok. I just updated OpenSSL to 1.0.2j and it's headers are in my system and me doing "openssl version" gives me version 1.0.2j (pull here: https://github.com/skycocker/chromebrew/pull/311 )

But PHP7 still gives me the same error I got when I clean installed chromebook (from scratch) installed chromebrew from scratched and then install PHP7 (I can't think of a better way to do a test for this project.)

Here's the error I get "configure: error: Cannot find OpenSSL's libraries"

what am I doing wrong?

uberhacker:
Hey @thedamian: Is the issue that you cannot install or that PHP throws that error during HTTPS requests on program execution?

thedamian:
When getting to the configure.... line it wrros out with:

"configure: error: Cannot find OpenSSL's libraries"

Ive tried several changes like the 
--with-openssl=/usr/local

Ive even tried

--with-openssl-dir=yes

And even --prefix=/usr/local

But with and without it it foes not install. Also i have a vrand new erased box. And also with and without both openssl and openssl_devel installed.

cstrouse:
Package installs properly on Samsung Chromebook 3 (XE500C13-K01US).

```
Platform 8872.73.0 (Official Build) stable-channel celes
ARC Version 3583806
Firmware Google_Celes.7287.92.60
```

uberhacker:
@cstrouse: Thanks for testing the package install.  Glad you had success!   Did you get a chance to try a [Composer](https://getcomposer.org/download/) install yet?  Just cut and paste the lines in the `Download Composer` section.  On my system, I had to remove the openssl package so php would work with https requests.

@thedamian: I'm not sure what options would work best.  Since php looks in the `ext` directory to configure extensions, that might be a directory we can target as a `--with-openssl(-dir)=` option.

Thanks guys!

thedamian:
@cstrouse how did you test this?

@uberhacker the php cant find openssl and that has to be fixed.

cstrouse:
@thedamian crew install php7 ; php -v ; php -S localhost:3000 and was able to view a phpinfo.php I created. I didn't try anything with SSL though as I was just trying to see if it would error out during compilation like you experienced. I'll do some more extensive testing and see if I find anything.

uberhacker:
@thedamian: Not sure why your system couldn't find openssl but installing the openssl package worked in my instance.  Apparently, it also worked for @cstrouse as well.  I don't have time right now (work always gets in the way, doesn't it? :-), but I can look at it later this evening and see if I can find out what php is expecting in terms of openssl library stuff.

uberhacker:
@thedamian: Could this be a case of 7.1.0 being too "unstable"?  What happens if you try the 7.0.14 version?

thedamian:
@cstrouse did you edit your php7.rb file with these changes?

cstrouse:
@thedamian I started with just ./configure and have been experimenting. The openssl headers are found properly by addng pkgconfig as a dependency and adding --with-opensl to configure line.

If you don't want to pull in pkgconfig then try adding --with-openssl --with-openssl-dir=/usr/local/ssl --with-lib-dir=/usr/local/lib64. The code searching for the paths skips past where things actually are for some reason.

It compiles but when installing composer produces the following errors:
```
Warning: copy(): SSL operation failed with code 1. OpenSSL Error messages:
error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed in Command line code on line 1

Warning: copy(): Failed to enable crypto in Command line code on line 1

Warning: copy(https://getcomposer.org/installer): failed to open stream: operation failed in Command line code on line 1
```
So now the issue is having several other places that paths are incorrect and that we'll need to add to the configure line also.

cstrouse:
With pkgconfig added as a dependency and this configure line composer will install.
```
system "./configure --with-openssl --with-openssl-dir=/usr/local/ssl"
```

thedamian:
I got it to run. You need these extra parameters to the ./config line:

      --prefix=/usr/local \
      CPPFLAGS=\"-I/usr/local/include\" \
      LDFLAGS=\"-L/usr/local/lib\" "

PLEASE tell me if there's something you see there as to WHY it's what _I_ needed but not what YOU needed (perhapts you have something in your .bash_profile or bashrc file?

Please discuss.

thedamian:
I forgot to mention. You CAN install composer WITHOUT OPENSSL (without https) by simply changing that first line in the composer install script from HTTPS to HTTP I had to do it for a class that another teacher was teaching

cstrouse:
Sure you can but then https URLs won't work right when you use libcurl, file_get_contents, etc.

thedamian:
anything wrong with my changes.

Anyone has successfully installed this patch? (and I mean you physically editing your /usr/local/lib/crew/packages/php7.rb and THEN doing a crew install php7)

Let me know if my changes are needed and what you think of them.

I want php to use openssl but even with my changes the first line in composer still doesn't work.

uberhacker:
@cstrouse is right.  Basically, what he is saying is you need https or the php experience is not fun. :)  @thedamian: Maybe we should discuss what other packages we have installed?  It's possible some other package has the necessary components which openssl requires.  Not sure why you would need the extra configure options since those basically look like the default settings.

thedamian:
Yes. Ive upgraded the headers and regular openssl to the more secure 1.0.2j thats why you have the difference. 

But maybe we should have another package for those that want rheir headers to maybe the older version that matches the built in version that care more about compatibility than with security

thedamian:
Im not merging this until this builds on a virgin install if chromebook.

uberhacker:
OK guys (@thedamian, @cstrouse): I have successfully installed php7 with full support of openssl.  I have modified `php7` and the `openssl` package as submitted by @thedamian.  Please test it out when you get a chance.  You should be able to do all of the following:
```
$ openssl version
OpenSSL 1.0.2j  26 Sep 2016
$ php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
$ echo '<?php phpinfo();' > phpinfo.php
$ php -S localhost:3000
Visit http://localhost:3000/phpinfo.php with the following in the openssl section:
OpenSSL support	enabled
OpenSSL Library Version	OpenSSL 1.0.2j 26 Sep 2016
OpenSSL Header Version	OpenSSL 1.0.2j 26 Sep 2016
Openssl default config	/usr/local/ssl/openssl.cnf
```

lyxell:
There's packages depending on openssl_devel, in its current state this PR breaks installation of those packages.

uberhacker:
OK, I can hold off on removal but I really want it removed eventually.  I agree with @thedamian that we should have the ability to override the pre-installed openssl version and "upgrade" when we deem worthy rather than wait on the Chrome OS team.  You can still use the headers from the `openssl` package.  I think if we keep the `openssl_devel` package, we should make sure it is updated with the latest version as @minektur mentioned in #311, otherwise, it's pointless having it.  Keeping just one package updated is much less of a pita to maintain.

thedamian:
My pull that im waitingnon you guys to comment on upgrades openssl froma 6 year old version to the current version that is only months old. It does this for both openssl and openssl_devel

I think this PR ahluld do what the title says. Make php7 work on a chromebook that has a fresh install if chromebrew and isnt setup and isnt setup any different 

It DOES have a dependenacy on openssl which is the crew installed version so make it work with that version.

uberhacker:
@thedamian: Have you tried the latest PR update yet on a fresh install?  It should work.  But as every developer always claims: "It worked on my system." :tm:

thedamian:
Alright guys.

Here's my latest work on this: https://github.com/skycocker/chromebrew/pull/311/

I'm am VERY impress by your work @uberhacker

I think you've done great work with openssl and getting it built in a way that's integrated with the system as much as possible.

and your work on php7 is VERY appriciated because I've wanted PHP (and several people have been asking for php) on our system for a long time.

But there where a couple of things missing in both packages. missing dependencies and missing sudo commands. and new parameters for SSL to actually work.

I've tested my work (on my box) and I completely in agreement  with you, I want to make sure it doesn't just "work on my box"

BUT more importantly. How do we do this.

my work is based on all you've done here.
I've tested it by having a file with <?PHP phpinfo() ?> and doing the old: (PHP -S localhost:8000)
and the info shows "SSL Version	OpenSSL/1.0.2j" so it looks like it's worked.

the  depends_on 'libpng'  for  --with-pcre-regex
and the "--with-ssl=/usr/local/bin/openssl \" is the only way it seems to see the new openssl successfully.

uberhacker:
Hey guys (@thedamian, @lyxell, @cstrouse, @jam7): After experimenting a bit with different options, I have fixed the `openssl` and `php7` packages to install `The crew way` :tm: (aka elegantly).  I'm hoping that a fresh install will "just work".  Please test it when you get a chance and report back if you had success (or not).  It would have been nice if it didn't take nearly 30 minutes to compile each time during testing but patience pays off in the long run. :)

minektur:
Ed - you might find that running a virtual machine with cloudready on some
beefy desktop machine might get you faster testing results.  My desktop
machine running a virtual chromeos via cloudready + chromebrew is
significantly faster than my HP chromebook 14 "falco".

uberhacker:
Thanks @minektur.  I'll have to give that a try sometime.  This Celeron processor on my Chromebook is SSSLLLOOOWWW.

jam7:
@uberhacker: Thank you for your efforts. I tried it and noticed a problem.  The filelist, `/usr/local/etc/crew/meta/php7.filelist`, is empty after the installation.  It is the sign of poblem.  I've researched and found php makefile uses INSTALL_ROOT instead of DESTDIR.

I modified it and found another problem of php.  php runs `pear/install-pear-nozlib.phar` and this creates files at wrong place, just under #{CREW_DEST_DIR}, instead of #{CREW_DEST_DIR}/#{PREFIX}/php.

jam7:
I corrected PHP install problem.  Following codes should work fine with php7.rb.

```
  def self.install
    system "make", "INSTALL_ROOT=#{CREW_DEST_DIR}", "install"

    # clean up some files created under #{CREW_DEST_DIR}. check http://pear.php.net/bugs/bug.php?id=20383 for more details
    system "mv", "#{CREW_DEST_DIR}/.depdb", "#{CREW_DEST_DIR}/usr/local/lib/php"
    system "mv", "#{CREW_DEST_DIR}/.depdblock", "#{CREW_DEST_DIR}/usr/local/lib/php"
    system "rm", "-rf", "#{CREW_DEST_DIR}/.channels", "#{CREW_DEST_DIR}/.filemap", "#{CREW_DEST_DIR}/.lock", "#{CREW_DEST_DIR}/.registry"
  end
```

uberhacker:
@jam7: That is a great analysis.  Have you confirmed that both install and remove work with these changes?  I wonder if we can add build options where we don't need to move things around after make?

jam7:
Yes.  I confirmed both install and remove.  Files moving around are created as a part of php configuration, so I think it is difficult to avoid making them with configure or build options.

BTW, this php7.rb works fine with openssl_devel too.  You can confirm it by following commands:
```
$ sudo crew remove openssl
$ crew install openssl_devel     # 1.0.2f is good enough
$ vi php7.rb                     # modify dependency to use openssl_devel instead of openssl
$ crew install php7
```

packages/openssl.rb
```diff
@@ -1,13 +1,21 @@
...
+  def self.build                                                  # self.build contains commands needed to build the software from source
+    system './config --prefix=/usr/local'
+    system 'make'                                                 # ordered chronologically
+  end
+
+  def self.install                                                # self.install contains commands needed to install the software on the target system
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"          # remember to include DESTDIR set to CREW_DEST_DIR - needed to keep track of changes made to system
+    system 'cd /usr/local/ssl && \                                # make sure cert.pem exists
+      wget http://curl.haxx.se/ca/cacert.pem && \
+      mv cacert.pem cert.pem'
```
jam7: `system "wget", "http://curl.haxx.se/ca/cacert.pem", "-O", "#{CREW_DEST_DIR}/usr/local/ssl/cert.pem"`

is better.  If we leave files under CREW_DEST_DIR, crew manges them (correctly removed at crew remove openssl)

uberhacker:
@jam7: That is awesome.  Thank you for your testing efforts.  I will give those changes a try when I get off work this evening and then I'll update the PR so we can finally get this merged into master!  BTW, I think `crew` is a better solution than `crouton` since there is no chroot layer.  I'm thinking about creating a repository for binaries so folks don't need to compile packages each time.  What do you think?

lyxell:
@uberhacker This have been up for discussion earlier and I think the general consensus is that binaries would be good if there's a way to verify the integrity of them in an automated fashion. I think there's an open issue about it in the issue tracker.

packages/openssl.rb
```diff
@@ -1,13 +1,21 @@
...
+  source_url 'ftp://ftp.openssl.org/source/openssl-1.0.2j.tar.gz' # software source tarball url
+  source_sha1 'bdfbdb416942f666865fa48fe13c2d0e588df54f'          # source tarball sha1 sum
+
+  def self.build                                                  # self.build contains commands needed to build the software from source
+    system './config --prefix=/usr/local'
+    system 'make'                                                 # ordered chronologically
+  end
+
+  def self.install                                                # self.install contains commands needed to install the software on the target system
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"          # remember to include DESTDIR set to CREW_DEST_DIR - needed to keep track of changes made to system
```
jam7: I forgot to mention here.  openssl makefile uses INSTALL_PREFIX.
```
    system "make", "INSTALL_PREFIX=#{CREW_DEST_DIR}", "install"
```

packages/openssl.rb
```diff
@@ -1,13 +1,21 @@
...
-    i686: 'cadea32ec770c4b44d565b7e5fdf96a469a05757',
-    x86_64: '3cf4defb11fc2fccce77736d0f4559e56d9d7e05'
-  })
+  version '1.0.2j'
+
+  source_url 'ftp://ftp.openssl.org/source/openssl-1.0.2j.tar.gz' # software source tarball url
+  source_sha1 'bdfbdb416942f666865fa48fe13c2d0e588df54f'          # source tarball sha1 sum
+
+  def self.build                                                  # self.build contains commands needed to build the software from source
+    system './config --prefix=/usr/local'
```
jam7: It's better to create shared library as what @thedamian did.
```
    system "./config", "--prefix=/usr/local", "shared", "zlib-dynamic"
```

packages/openssl.rb
uberhacker: @jam7: Maybe it would be better to remove the `openssl` package from this PR?  Then we could just merge @thedamian's version when it is ready?

jam7: I think so.  I'm going to make a PR about only openssl.

lyxell:
Looks good to me in it's current state.

packages/php7.rb
```diff
@@ -3,16 +3,26 @@
...
+      --with-openssl \
+      --with-pcre-regex \
+      --with-zlib'
+    system 'make'                                            # ordered chronologically
   end
 
-  def self.install                                                # self.install contains commands needed to install the software on the target system
-    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"          # remember to include DESTDIR set to CREW_DEST_DIR - needed to keep track of changes made to system
+  def self.install                                           # self.install contains commands needed to install the software on the target system
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"     # remember to include DESTDIR set to CREW_DEST_DIR - needed to keep track of changes made to system
```
jam7: @lyxell, @uberhacker: Need to use INSTALL_ROOT for php make.  And php make install has bug, so need little modification to handle it correctly.

```
    system "make", "INSTALL_ROOT=#{CREW_DEST_DIR}", "install"

    # clean up some files created under #{CREW_DEST_DIR}. check http://pear.php.net/bugs/bug.php?id=20383 for more details
    system "mv", "#{CREW_DEST_DIR}/.depdb", "#{CREW_DEST_DIR}/usr/local/lib/php"
    system "mv", "#{CREW_DEST_DIR}/.depdblock", "#{CREW_DEST_DIR}/usr/local/lib/php"
    system "rm", "-rf", "#{CREW_DEST_DIR}/.channels", "#{CREW_DEST_DIR}/.filemap", "#{CREW_DEST_DIR}/.lock", "#{CREW_DEST_DIR}/.registry"
```

lyxell:
Tested and working for me if I add `libpng` as a dependency.

uberhacker:
@lyxell: I suppose `libpng` is a dependency required by `--with-gd`?

lyxell:
@uberhacker Sounds reasonable

uberhacker:
@lyxell: Please test again at your leisure.  Thanks!

lyxell:
Nice work!

PR #310, opened 2016-12-30
change method names to conform to ruby conventions

mattddowney:
See:
https://github.com/bbatsov/ruby-style-guide#snake-case-symbols-methods-vars

Should probably make variable names conform to this as well, but that is for another day.

### DIFF
--- a/crew
+++ b/crew
@@ -24,7 +24,7 @@ case ARCH
     SHORTARCH="64"
   when "armv7l"
     SHORTARCH="32"
-  else  
+  else
     SHORTARCH="32"
 end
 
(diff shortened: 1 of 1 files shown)


thedamian:
Ill try it on ny box and would like others to try this too hut doesnt this fall under "if it aint broken....."

mattddowney:
Lots of manual testing going on. Any ideas for some test automation?

minektur:
Maybe we could do something with cloudready - automated vm cloning, automated deploying, etc... but it would be a lot of work.

We now support 3 platforms... I think among the regular contributors we have pretty good coverage of platforms...

What we could maybe do is have a 2-stage release process. a "Beta" where we push stuff, and all can test, and then weekly, or monthly releases or something...  This would fit with our desire to have binaries hosted and release-tagged on github.

jam7:
@mattddowney Hi, is it possible to rebase this PR to the latest one?  I think modifying crew like this way is very important.

PR #181, opened 2016-06-02
Openssl_devel  - headers to allow building against system openssl

minektur:
Installs JUST the headers to match the chromeos supplied libraries so that you can build things 
that link with openssl. Needs to be kept version-synced with chromeos releases
Could detect current version and grab one of many different packages, compare to saved
hashes and support multiple versions  maybe - call it a TODO

update the two 'from-source' python pacakges and libevent to depend on openssl_devel

PR #311, opened 2016-12-30
adding PHP5 and pkg_config removed

thedamian:
Alright. I've updated openSSL to the latest (well the latest 1.0.2 version which is 1.0.2j)

Tell me if I've done something that will screw up anything.

This is because the PHP7 package was missing openssl for it to install with openssl capability.


jam7:
Do we need to update openssl_devel?  Regarding to #181, I thought we would like to keep the version of openssl_devel as same as what the system has.

BTW, flip c100pa has 1.0.2h instead of 1.0.2f.  Does x86_64 or x86 have 1.0.2f?
```
$ /usr/bin/openssl version
OpenSSL 1.0.2h  3 May 2016
```

lyxell:
Shouldn't the Php-package use openssl_devel rather than openssl?

thedamian:
In /usr/bin/openssl the x86 version has 1.0.2h

Im updating both here but noticed openssl_devel just gets the latest version and extracts just the headers and nothing else.

Since the openssl package now has the headers maybe we dont beed openssl_devel any more.

@lyxell correct. I think php sbould need only the headers but at this point no matter what i vant get php to gind the openssl libraries.

But yourr right. Or is it that php uses the actual application?

lyxell:
PHP does not use the actual application.

lyxell:
And I would rather argue that it is the openssl package that is not needed, rather than the openssl_devel package. Since the OpenSSL executable and shared libraries is included in Chrome OS. It's only the headers that are missing.

uberhacker:
Is there any way we can have a variable set in the openssl(_devel) packages that matches the currently installed version?  It appears everything breaks if the versions don't match (ie. php https requests, git operations via ssh, etc.).  Unless we can completely overwrite the version that comes pre-installed?

thedamian:
@lyxell  but why not have the package so that we can upgrade it when WE want and not when google deems it?

lyxell:
@thedamian To have multiple executables seems to me like begging for trouble, and it did break https-negotiation in the past, not sure if that has been fixed

minektur:
We can't control what version of openssl libraries the base OS uses.  openssl-devel exists so that we can get the headers of the (theoretically) same version that the base OS has installed.

This allows other software that wants to build against the openssl libraries that come with the base system.  We should be able to build any software with openssl if we have the matching headers installed.

We could, as you suggest, make our own installation of openssl.  This would mean that there were two versions of openssl installed on the system - one in /usr/lib and one in /usr/local/lib.  This is the situation we had when we just had an openssl package.

The problem that we faced is that configure scripts, and auto-build scripts would find the wrong openssl library - it would use the headers in /usr/local/include (version X of openssl) and it would find the library in /usr/lib (version Y of openssl) and stuff would break.  So, people making packages to build software with openssl need to be extra careful to make sure that it builds using both the library and the headers in /usr/local.  For a lot of software you can specify the location of openssl so that you get the right version.  Some dont - some make it a giant pain to specify a version...

That problem goes away if we just use the system openssl, and keep headers around to match what the base OS already provides.

We DO have a problem when openssl-devel gets out of sync with the base - and should update that quickly.  (This presupposes that the different chromeos platforms all use the same version, which so far has been the case).

So, for PHP7 build, we need to A) update the openssl-devel package - which needs to be done anyway, and B) then re-configure/re-build openssl.

This should make the issues go away.

I'd vote for keeping openssl-devel and not making a full openssl package - things would be simpler, more likely to work, and it will be easier to make packages, with of course the issue that we need to be uptight and on the ball about keeping it up to date.


Some random proverb I remember reading from when I was a kid - "A man with one watch always knows what time it is.  A man with two watches is never sure."

uberhacker:
@minektur: Excellent analysis of the situation.  I have taken your suggestions with a bit of a caveat.  I suggest keeping the `openssl` package and nixing the `openssl_devel` package. :)  I have successfully installed the `php7` package with openssl where both the library and header versions match the pre-installed Chrome OS version.  I think it is not necessary to manage both `openssl` and `openssl_devel` packages based on the wisdom of your random proverb.

lyxell:
@minektur What are your thoughts on this?

jam7:
@thedamian I'm writing down here.

- double make line
- "make INSTALL_PREFIX=#{CREW_DEST_DIR} install" should work
- do we really need to install cacert.pem as cert.pem?

uberhacker:
I think you need `DESTDIR=/usr/local/ssl` so you know exactly where the installation puts the files but if you do set it to `#{CREW_DEST_DIR}`, you will also need to change the `php7` package `--with-openssl=` option with the matching path so php knows where to find the correct openssl location.  You need `cert.pem` since that is the default filename.  I suppose you could change the value with something like `openssl.cafile=/usr/local/ssl/cacert.pem` but that would be an extra configuration step.  You can find out the default locations by executing ` php -r "print_r(openssl_get_cert_locations());"`.  This is a bit of a chicken before the egg scenario, however, since you need php installed before you can find that out.  Hope that helps.

jam7:
@uberhacker DESTDIR for autoconf or INSTALL_PREFIX for openssl config is not --prefix stuff.  It specifies where files should be copied temporary to make install package.  After that, the crew script moves files under /usr/local as previously specified by --prefix.  Does this make sense?

uberhacker:
@jam7: Yes, that does make sense.  However, we should know the ultimate DESTDIR, right?  By explicit setting it, we define exactly where were we want the files to be installed and then we know where to point our paths when configuring other packages.

jam7:
@uberhacker I guess you are talking about OPENSSL directory.  It will be /usr/local/ssl if it is configured with --prefix=/usr/local option or /etc/ssl for prebuild openssl.  I'm not sure what openssl_get_cert_locations() returns.  It may contains "/certs" suffix like "/usr/local/ssl/certs".  Is this OPENSSL directory what you need?

```
$ /usr/local/bin/openssl version -d
OPENSSLDIR: "/usr/local/ssl"
```

```
$ /usr/bin/openssl version -d
OPENSSLDIR: "/etc/ssl"
```

thedamian:
completely recreated the openssl and openssl_devel projects based on the amazing work by @uberhacker  and also used his great php7 work to get it working with full SSL support

jam7:
It's my personal opinion, but it's better to have separate pull requests.  Especially, pkg_config is very different from others.

thedamian:
absolutely @jam7 
I've done this mistake before but if it helps pkg_config is part of how I was trying to fix php7 and it's integration with OPENSSL so hopefully this mistake can be overlooked.

uberhacker:
Were getting closer guys but I have to agree with @jam7 regarding separating the pull requests.  It is much easier to track after the fact if you separate out.  It took me awhile to figure this out but you can create separate PRs by committing on separate branches.  This will allow you to have more than one pending PR in the queue with a single locally cloned repo.  @thedamian, I appreciate the kind comments and your commitment to make this all happen.

lyxell:
I'm pretty sure not using `#{CREW_DEST_DIR}` breaks uninstallation?
And I still want the thoughts of @minektur on this since he has the most insight in how `openssl_devel` works.

uberhacker:
I've noticed `sudo crew remove <some_package>` doesn't actually remove the app with some packages.  I will have to look into this process a bit more to edumacate myself how this is all supposed to work.  Thanks for the insight @lyxell.

lyxell:
@uberhacker After running `make install`, the name and directory structure of all files in `#{CREW_DEST_DIR}` are saved in a text file in `/usr/local/etc/crew/`, this way, when uninstalling the packages, we know exactly which files to remove.

packages/openssl.rb
```diff
@@ -1,13 +1,29 @@
...
+
+  depends_on 'perl'
+
+  def self.build
+	system "./config","--prefix=/usr/local","shared","zlib-dynamic"
+	system "make" #depend
+  end
+
+  def self.install
+    system "make"
```
jam7: we dont need this second duplicated make line

packages/openssl.rb
```diff
@@ -1,13 +1,29 @@
...
+    system "sudo make DESTDIR=/usr/local/ssl install"
```
jam7: "make", "INSTALL_PREFIX=#{CREW_DEST_DIR}", "install" is OK.
no need sudo here.  openssl makefile uses INSTALL_PREFIX instead of DESTDIR.

packages/openssl.rb
```diff
@@ -1,13 +1,29 @@
...
+    system 'cd /usr/local/ssl && \                                # make sure cert.pem exists
+    	    sudo wget http://curl.haxx.se/ca/cacert.pem && \
+    	    sudo mv cacert.pem cert.pem'                                # && \
+    	    #ln -s /usr/local/ssl/bin/openssl /usr/local/bin/openssl'    # override the system version (in case of a version mismatch)
+                                                                  	# the caveat here is /usr/local/bin should come before /usr/bin in $PATH (the default)
```
jam7: I'm not sure why we need cert.pem here.  At least, isn't it possible to make these lines to echo commands?

uberhacker:
@jam7: `cert.pem` is needed to avoid the dreaded ssl error while making php https requests via the cli like below:
```
PHP Warning:  include_once(): SSL operation failed with code 1. OpenSSL Error messages:
error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed in /files/somefile.php on line x
```
By default, php looks for a file called `cert.pem` in `/usr/local/ssl` and if it's not there, complains like above.

minektur:
(There is a TL;DR at the bottom)

I'm probably OK with us having our own build of openssl libraries/headers/etc.  This will make package creation a bit harder but also more flexible.  It would also remove the 'hurry and update after chromeos changed the library version underneath us' issue.

For simple-ish software that uses openssl,  we should just be able to put openssl as a dependency, and then run ./configure and the configure script will find it. 

This may or may not "just work" depending on rpath in the binaries, on your environment (e.g. LD_LIBRARY_PATH) and depending on how the software actually builds itself.  You just have to make sure, using command line, environment, etc or patching later with patchelf that it is looking at openssl in /usr/local and not the system one.

For more complicated software that builds it's own libraries, installs them and then expects to find them, and system libraries, it is more complicated.  The example we've been discussing is php - there is a configure option "--with-openssl=/usr/local/lib/ssl" or whatever the actual path is...  Where you can explicitly tell it where to find the libraries.  Some software is not so smart and you'll maybe have to patch build scripts to force the right one - I dont have any examples off the top of my head...

As an aside, some software that builds it's own libraries is difficult to build because of our CREW_DEST_DIR installation target directory.  Some configure and build scripts for software that builds and installs it's own libraries will assume if your --prefix=  is set that is where things will eventually be found and will force it's libraries to look there.  So you have to configure it like it is going to live in /usr/local (so that library paths get hard coded) and then really install it in CREW_DEST_DIR so that the manifests get made right and then it will get automagically installed.

nethack4.rb is a perfect example of this.  It has it's own not-very-smart perl script that handles all the complicated building and installing of it's own libraries.  it has the equivalent of --prefix - "-i"

If you specify -i #{CREW_DEST_DIR) then a bunch of binaries and libraries get installed in to CREW_DEST_DIR expecting that that is the final location of the files and that path gets hard-coded into the rpath of the binaries.

You'll note that there is a separate build and install phase and that I had to specify two different sets of install locations to get it to work - the build phase tells it that it will eventually live in /usr/local and the install phase tells it to NOT DO ANYTHING but install the files in CREW_DEST_DIR.

nethack doesn't use openssl, but other software using openssl may have a similar problem - and might end up using the wrong library because it found the wrong one.  A careful examination of versions is important in package development.

tmux is the reason that I made openssl_devel in the first place - it depends on libevent, which subsequently depends on openssl, and because of path problems it kept finding the system openssl and I didn't have the time or patience to track down how to make it do the right thing. 

Making tmux work with openssl instead of openssl_devel would be the litmus test to show we have got it working right.

--------------------------
TL;DR: It's probably OK to have a full openssl install for crew and to get rid of openssl_devel, but it will be a pain some times, so be careful.

jam7:
I have two problems with php7.rb.

- `--with-ssl` option is not recognized.  I got following message.  Created php7 is compiled without ssl but with curl when I check `php -m`.  It might work with ssl through curl since curl supports ssl.
```
configure: WARNING: unrecognized options: --with-ssl
```

- `--with-pcre-regex` is not working well on my system since it doesn't have pcre.h.  Do your system have pcre.h?  php7 is compiled with internal pcre library on my machine.
```
checking pcre install prefix... no
```

uberhacker:
@jam7: Yeah, that should be `--with-openssl` instead of `--with-ssl`.  More about the configure options can be found here: http://php.net/manual/en/configure.about.php.
Not sure why your system is not finding the pcre library.  Maybe we should exclude that option?  It's more of a nice to have than a need to have anyway.

jam7:
@uberhacker: my system has pcre library but no pcre.h header file.  It is similar to the openssl.  Lzma and bz2 have the similar problem too.  So, I was just wondering where other ppl get pcre.h header file.

uberhacker:
@jam7: That is a very good question.  Mine is located at `/usr/local/include/pcre.h`.  It would nice if `crew` had a command like `dpkg -S` on Debian systems where you could enter a file as a parameter and it would tell you which package included that file.  Anybody have an idea how this could work?

jam7:
grep pcre.h /usr/local/etc/crew/meta/* works in general.

2017年1月4日(水) 9:18 Ed Reel <notifications@github.com>:

uberhacker:
@jam7:  Then you could have answered your own question. :)  I found it in the `pcre` package.
```
pcre.filelist:/usr/local/include/pcre.h
```

uberhacker:
So if we include the `--with-pcre-regex` configure option in the `php7` package, we should also include the `pcre` package as a dependency.

jam7:
@uberhacker thank you.  pcre.h in pcre package pretty make sense.  I agree about the dependency.

thedamian:
Extra Make line removed.

--with-ssl seems to work while --with-openssl does not.

You're getting an error? If you edit your file to change it to "with-opessl" and leave everything else the same does it work?

thedamian:
this --with-ssl on php7 works and I get the following confirmation information you see below but does NOT work on PHP5

checking for cURL support... yes
checking for cURL in default path... found in /usr/local
checking for cURL 7.10.5 or greater... libcurl 7.32.0
checking for SSL support in libcurl... yes
checking how to run the C preprocessor... cc -E
checking for openssl support in libcurl... yes
**checking openssl/crypto.h usability... yes
checking openssl/crypto.h presence... yes
checking for openssl/crypto.h... yes**
checking for gnutls support in libcurl... no
checking for curl_easy_perform in -lcurl... yes
checking for curl_easy_strerror in -lcurl... yes

jam7:
@thedamian: Php7 searches ssl automaticallly.  Those you mentions are from that functionality.  `--with-ssl` is just ignored by `./configure` since it is not known option.  If you type `php -m` after the installation, you will not see `openssl` in the compiled module list.  `--with-openssl` option, php7 correctly configured with openssl and `php -m` shows `openssl`

jam7:
@thedamian: to answer your questions, it is not an error.  it is an warning.

```
configure: WARNING: unrecognized options: --with-ssl
```

I copied php7.rb from your branch by

```
$ wget https://raw.githubusercontent.com/thedamian/chromebrew/master/packages/php7.rb -O /usr/local/lib/crew/packages/php7.rb
```

You can gather the output of `crew install php7` by using `script` command and investigate it later.

uberhacker:
Since we already merged the `openssl` and `php7` package updates into master, this PR will need to be updated to remove these packages.  I might suggest closing this PR and opening a separate one called `Add php5 package`?

thedamian:
php5 isn't working and the php7 would have to change to work.

But who merged the other PR that wasn't working.

Guys please test this. 

I'm going right now to test php7.rb and will gladly close this if it's already done.

lyxell:
The other pull request is working and merged.

jam7:
@thedamian: I think php7 on x86_64 requires pkg_config as you mentioned with keen insight.  php doesn't look inside of lib64.  It is possible to let php do so by adding `--with-lib=lib64`, but this conflicts with other architectures.  Using pkg_config solves it as you noticed.  Please make a PR for pkg_config.

In addition, pkg-config searches `/usr/local/lib/pkgconfig` and `/usr/local/share/pkgconfig` only by default.  So, it is required to let pkg-config check `/usr/local/lib64/pkgconfig` also since openssl places it's information in `lib64/pkgconfig`.  I don't find the way to do so yet.  At least, user can specify it by PKG_CONFIG_PATH environment variable.

Anyway, it is possible to use php7 with openssl on x86_64 by following steps.

1. install new pkg_config in you PR
1. set environment variable as `export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig:/usr/local/lib64/pkgconfig:/usr/local/share/pkgconfig`
1. install merged php7

lyxell:
Alright, my bad, I thought it was tested on x86_64 as well.

jam7:
I didn't know it too at that time.  It is difficult to test on all devices without having them all. :(

lyxell:
The reason for needing pkg_config is to correctly locate openssl, right?
How about `--with-openssl=/usr/local/ssl`?

lyxell:
But this really is a flaw of the gcc toolchain of x86_64 which has an incorrect search order when linking the libraries. It would be great if we could update the toolchains for x86 as well.

lyxell:
@jam7 Are you positive? You're probably right, but if I recall correctly, issuing `gcc -Wl,--verbose` on x86_64, showed that there was some issues with SEARCH_DIR, but this might be unrelated.

jam7:
@lyxell: Yes, you are right.  There is a real problem in gcc and ld search paths which uses /usr/lib64 prior to /usr/local/lib.  However, skycocker binutils solved this problem already by a trick making a symbolic link from /usr/local/x86_64-*/lib64 to /usr/local/lib.  So, /usr/local/lib was parsed at the beginning in fact.  :)

jam7:
I've updated pkgconfig instead of pkg_config at #341.  Just information since I discussed about pkg_config here.

thedamian:
Thanks @jam7 

I'll probably close this PR and create a new one because right now PHP7 isn't installing.

uberhacker:
@thedamian: Are you saying PHP7 is not installing based on this PR or what is currently in master?

thedamian:
What is currently merged into master is NOT installing on my machine. This PR was to help fix that but It looks like the code on the last PR from (I think it was you @uberhacker  no?) 

But it looks like @lyxell merged it.

I'm on my mac now for work so I'll have to get home to figure out how to help you guys update the PHP7.rb so it installs on people's chromebook which I believe (and no one has written otherwise) does NOT install.

uberhacker:
@thedamian: I have installed it fine during my testing.  I believe @jam7, @lyxell and @cstrouse had success with installation also.  It even uninstalls cleanly; leaving no residue behind.  What are your exact Chromebook system specs?  You can see from the `About Chrome OS` link in `Settings`.  My specs are below:
```
Platform 8872.73.0 (Official Build) stable-channel kip
```

lyxell:
It works on arm, and it looks like it works on i686 as well? The only issue right now is with x86_64 not finding openssl because of pkg_config.

PR #319, opened 2017-01-02
Add crew search -i and -a options

uberhacker:
-i = List all installed packages
  -a = List all available packages


lyxell:
I don't really see the use case for this
```bash
crew search | grep '(i)'
crew search | grep -v '(i)'
```

uberhacker:
Which would you rather type?  `crew search -i` or `crew search | grep '(i)'`.  Always naysayers. :)

lyxell:
Personally I think there's a greater value in keeping the code base small and manageable, but that's just my opinion.

uberhacker:
Not every user is going to know how to pipe to grep but point taken.

PR #321, opened 2017-01-03
Changed to use #{CREW_DEST_DIR} in libffi.rb

jam7:
libffi was not using CREW_DEST_DIR.  As a result, it didn't make a file list correctly in /usr/local/etc/crew/meta/libffi.filelist.  By this modification, the problem will be solved.

### DIFF
--- a/packages/libffi.rb
+++ b/packages/libffi.rb
@@ -13,6 +13,6 @@ class Libffi < Package
   end
 
   def self.install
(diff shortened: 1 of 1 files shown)


jam7:
Should I increment version number like `3.0.13-upd1`?

lyxell:
Looks good to me.

PR #322, opened 2017-01-03
Fix the usage of CREW_DEST_DIR in mono.rb.

jam7:
Similar to #321.  I modified two lines.

- `./configure` line, prefix argument is ignored since it doesn't start with `--`, so it is configured by default value, `/usr/local`, anyway.  Add `--` and fix the path.
- insall line, wrong variable name doesn't modify the install directory, so everything were being copied to configured directory, /usr/local, directly.  Modify variable name and fix the path.

### DIFF
--- a/packages/mono.rb
+++ b/packages/mono.rb
@@ -1,7 +1,7 @@
 require 'package'
 
 class Mono < Package
-  version '4.4.0.148'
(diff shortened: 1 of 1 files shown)


lyxell:
Looks good to me.

PR #327, opened 2017-01-04
Change to use openssl instead of openssl_devel

jam7:
I'm still not sure either openssl or openssl_devel is better, but I made a PR merging #309 and #311 for the case.

- openssl.rb using 1.0.2j is from both of them.  I modified little to match CREW_DEST_DIR mechanism.
- removing openssl_devel.rb since it has conflict with openssl.rb.  Both try to install header files into /usr/local/include/openssl/
- modify all packages to use openssl instead of openssl_devel.  Compiled binaries uses shared library and they use the identical API, so leave them as is without re-compiling.

You may have following errors in the middle if you tries to intall openssl previously.
```
$ wget http://curl.haxx.se/ca/cacert.pem
URL transformed to HTTPS due to an HSTS policy
--2017-01-05 02:15:42--  https://curl.haxx.se/ca/cacert.pem
Resolving curl.haxx.se... 80.67.6.50
Connecting to curl.haxx.se|80.67.6.50|:443... connected.
ERROR: cannot verify curl.haxx.se's certificate, issued by 'CN=Let\'s Encrypt Authority X3,O=Let\'s Encrypt,C=US':
  Unable to locally verify the issuer's authority.
To connect to curl.haxx.se insecurely, use `--no-check-certificate'.
```
 If so, please remove previous shared libraries by hand.
```
$ sudo rm /usr/local/lib/libssl* /usr/local/lib/libcrypto*
```


packages/openssl.rb
```diff
@@ -1,13 +1,27 @@
...
+  def self.build
+    system "./config","--prefix=/usr/local","shared","zlib-dynamic"
+    system "make"
+  end
+
+  def self.install
+    system "make", "INSTALL_PREFIX=#{CREW_DEST_DIR}", "install"
+
+    # make sure cert.pem exists
+    system "wget", "http://curl.haxx.se/ca/cacert.pem",
```
lyxell: Can we use https here?

uberhacker:
I can make a case for `openssl` instead of the `openssl_devel` package.  With `openssl` (full package), it gives us more control over what version is available instead of relying on the Chrome OS team to keep openssl updated (think latest security updates).  Also, it assures the binaries, libraries and headers are all consistent.  This is especially important with packages that depend on `openssl`.  It's possible that some Chromebooks have different binary versions of openssl and the `openssl_devel` package would need to accommodate all these different versions (a nightmare to manage).  With that said, if we do get rid of `openssl_devel`, we will need to update all packages that depend on `openssl_devel` to use `openssl` as a dependency instead.  Hopefully, that explains it well.

packages/openssl.rb
jam7: Yes.  Actually, wget uses https automatically through HSTS policy.

lyxell: You're right, but an upgrade via HSTS does not prevent a man-in-the-middle attack.

jam7: I see.  I've updated it.  Thanks.

uberhacker:
How about using `wget https://curl.haxx.se/ca/cacert.pem --no-check-certificate ...`?  That way it still works over https regardless.

lyxell:
@uberhacker The integrity of the certificates is a pretty big deal.

uberhacker:
True if you were transmitting private data that you wanted encrypted.  This is public data.

lyxell:
@uberhacker There's probably more appropriate forums to have this discussion, but anyway
https://en.wikipedia.org/wiki/Man-in-the-middle_attack
http://stackoverflow.com/questions/14907581/ssl-and-man-in-the-middle-misunderstanding

uberhacker:
What the `--no-check-certificate` switch does basically is says "I trust the ssl certificate owned by the curl.haxx.se domain is valid".  Is there any reason we cannot trust this domain?

lyxell:
@uberhacker the `--no-check-certificate` rather means "Ḯ'm positive there's noone intercepting the connection between me and curl.haxx.se right now, so there's no reason to validate the integrity of our communication"

uberhacker:
That is not my understanding based on the switch definition:
```
‘--no-check-certificate’
Don’t check the server certificate against the available certificate authorities. Also don’t require the URL host name to match the common name presented by the certificate.

As of Wget 1.10, the default is to verify the server’s certificate against the recognized certificate authorities, breaking the SSL handshake and aborting the download if the verification fails. Although this provides more secure downloads, it does break interoperability with some sites that worked with previous Wget versions, particularly those using self-signed, expired, or otherwise invalid certificates. This option forces an “insecure” mode of operation that turns the certificate verification errors into warnings and allows you to proceed.

If you encounter “certificate verification” errors or ones saying that “common name doesn’t match requested host name”, you can use this option to bypass the verification and proceed with the download. Only use this option if you are otherwise convinced of the site’s authenticity, or if you really don’t care about the validity of its certificate. It is almost always a bad idea not to check the certificates when transmitting confidential or important data. For self-signed/internal certificates, you should download the certificate and verify against that instead of forcing this insecure mode. If you are really sure of not desiring any certificate verification, you can specify –check-certificate=quiet to tell wget to not print any warning about invalid certificates, albeit in most cases this is the wrong thing to do.
```

lyxell:
@uberhacker Well, if you read the links I provided you, you would change your understanding.

uberhacker:
I did read the link you gave me and I still don't understand it any differently.

lyxell:
@uberhacker One of the features of TLS is that it (supposing none of its preconditions is broken) defends against MITM-attacks, if we use the `--no-check-certificate` flag, this no longer holds, and we are vulnerable against MITM-attacks.

uberhacker:
Even with a "verified" ssl certificate, there is still the possibility of a MITM attack.  If you truly want an unaltered cacert.pem file from curl.haxx.se, you should perform a sha1 check.  That is really what are concern is here, right?

uberhacker:
It all comes down to trust one way or another.  How do we know the cacert.pem coming from curl.haxx.se doesn't validate fake CAs?

lyxell:
@uberhacker The contents of the `cacert.pem` is constantly changing, so we can't really use a static sha1, that's why we at least should use TLS.

Regarding your later concern, in my opinion both haxx and Mozilla is to be considered trusted entities, so the most important thing is to validate the integrity of the connection.

uberhacker:
I don't trust either entity but that doesn't stop me from using the Internet because I believe most people have good intentions.

lyxell:
If that's a concern for a lot of people we could use:
```bash
cat /usr/share/chromeos-ca-certificates/*.pem > cacert.pem
```
to use the ca certs which comes with the operating system.

uberhacker:
@lyxell: Now you are trusting Google? LoL  Just kidding.  That is actually a really great idea.

minektur:
@lyxell - I like your idea of using the system CAs.

jam7:
@lyxell : I like that idea too.  What I really want to do is use system certs at /etc/ssl/certs directly, but I'm not sure how to do it.  Your cat technique achieves it, although it is not possible to catch the system update.

lyxell:
@jam7

The configure flag `--with-ssl-dir=/etc/ssl` would achieve that.

minektur:
wont it try to write it's files into that directory then?  or only read?

jam7:
Do some of `openssl` commands try to write data or certificates into OPENSSLDIR too?  If so, last modification is not good even if `/usr/bin/openssl` have the same problem though.

lyxell:
@jam7 I think it should be fine, but I'm only guessing.

lyxell:
Tested and working. Merging. :)

PR #329, opened 2017-01-05
Add crew source command

uberhacker:
This is basically the equivalent of `grep <filepath> /usr/local/etc/crew/meta/*` when you want to find out which package includes &lt;filepath&gt;.  The full path to the filename should be used.

`Usage: crew source [filepath]`

Essentially, this is the same concept as `dpkg -S <filepath>` on Debian-based or `rpm -qf <filepath>` on RedHat-based Linux distributions when you want to find out which package contains a certain file on the system.

### DIFF
--- a/crew
+++ b/crew
@@ -416,8 +416,8 @@ when "remove"
 when nil
   puts "Chromebrew, version 0.2.1"
   puts "Usage: crew [command] [package]"
-  puts "Available commands: search, download, install, remove"
+  puts "Available commands: search, download, install, remove, source"
 else
   puts "I have no idea how to do #{@command} :("
-  puts "Available commands: search, download, install, remove"
+  puts "Available commands: search, download, install, remove, source"
 end
--- a/crew
+++ b/crew
@@ -28,7 +28,6 @@ case ARCH
     SHORTARCH="32"
 end
 
-
 $LOAD_PATH.unshift "#{CREW_LIB_PATH}lib"
 
 USER = `whoami`.chomp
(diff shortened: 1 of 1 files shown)


uberhacker:
Any feedback guys?  What do you think?  @lyxell, @cstrouse, @thedamian, @jam7, @minektur?  If you like this command, I have another one in mind that complements this that is even more flexible. :)

minektur:
I'm fine with the functionality - I use the rpm/yum equivalents on centos
once a month or so and I can see why people would want it.

I personally would change the command name to something like 'whatprovides'
 (e.g. like yum) or 'origin' or something because the the word 'source' is
somewhat overloaded.  But honestly this is a minor nit and I don't have
strong feelings about the name.

uberhacker:
@minektur: Yeah, I agree that `source` is a bit vague and can have many different interpretations.  I like `origin` better than `whatprovides` since I'm lazy and hate typing out long commands. :)  I appreciate the feedback.

thedamian:
I was about to merge it when I figured I should test it first.

How do you test it. the ruby.rb package has  io.h

but crew source io.h doesn't bring anything up.

Tell me how to test it. Because it doesn't seem to be working for now.

uberhacker:
If I make changes to the `crew` command, should I be incrementing the version number?  How are new features treated?  Do you increment the minor or patch number?  Is there a long drawn-out political process to change?

uberhacker:
@thedamian: You will need to enter the full path of the file.  Something like `/usr/local/bin/ruby`.  It doesn't handle partial strings (ie. no path).

minektur:
I personally would prefer whatprovides because that is what YUM uses....
but I'd also be happy with origin.  I'm know I'm picky...

uberhacker:
I'm not picky.  Closing this PR in favor of `crew whatprovides`.  New PR coming soon...

PR #334, opened 2017-01-06
Add crew whatprovides command

uberhacker:
This is basically the equivalent of `grep <pattern> /usr/local/etc/crew/meta/*` when you want to find out which package includes &lt;pattern&gt;.
```
Usage: crew whatprovides [pattern]
```
Essentially, this is the same concept as `dpkg -S <filepath>` on Debian-based or `rpm -qf <filepath>` on RedHat-based Linux distributions when you want to find out which package contains a certain file on the system.

Changed from `crew source` to `crew whatprovides` per @minektur's preference.  He was the loudest so he wins. :)

### DIFF
--- a/crew
+++ b/crew
@@ -67,40 +67,12 @@ end
 
 def whatprovides (pkgName)
   fileArray = []
-  op = '*'
-  search = pkgName.dup
-  if search[/\*/]
-    search = search.gsub! '*', ''
-  end
-  abort if not search
-  if search[0] == '^'
-    op = '^'
-    search = search[1..search.length - 1]
-  end
-  if search[-1] == '$'
-    if op == '^'
-      op = '^$'
-    else
-      op = '$'
-    end
-    search = search[0..search.length - 2]
-  end
   Find.find (CREW_CONFIG_PATH + 'meta/') do |packageList|
     if File.file? packageList
       if packageList[/\.filelist$/]
         packageName = File.basename packageList, '.filelist'
         File.readlines(packageList).each do |line|
-          found = nil
-          case op
-          when '*'
-            found = line[/#{Regexp.quote(search)}/]
-          when '^'
-            found = line[/^#{Regexp.quote(search)}/]
-          when '$'
-            found = line[/#{Regexp.quote(search)}$/]
-          when '^$'
-            found = line[/^#{Regexp.quote(search)}$/]
-          end
+          found = line[/#{Regexp.new(pkgName)}/]
           if found
             fileLine = packageName + ': ' + line
             if not fileArray.include? fileLine
--- a/crew
+++ b/crew
@@ -70,18 +70,15 @@ def whatprovides (pkgName, silent = false)
   packageName = ''
   fileName = pkgName + "\n"
   Find.find (CREW_LIB_PATH + 'packages') do |filename|
-    Find.find (CREW_CONFIG_PATH + 'meta/') do |packageList|
-      packageName = File.basename filename, '.rb'
-      fileList = CREW_CONFIG_PATH + 'meta/' + packageName + '.filelist'
-      if File.file?(fileList)
-        File.readlines(fileList).each do |line|
-          if line == fileName
-            found = true
-            break
-          end
+    packageName = File.basename filename, '.rb'
+    fileList = CREW_CONFIG_PATH + 'meta/' + packageName + '.filelist'
+    if File.file?(fileList)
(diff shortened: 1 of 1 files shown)


thedamian:
wait? didn't I try to merge this already? This wasn't working before.
I did
crew source ehv.h

or any other file and it doesn't reply with anything. How do you test this?

uberhacker:
Hey @thedamian: It works if you supply the full path to the file.  Try something like `crew whatprovides /usr/local/lib/ehv.h`.  It would take a bit more work, but if you would like to search for a single file without the path, I can add that functionality.  Since you now have the `mlocate` package available, you can easily search for ehv.h and it will instantly report the full path in the search results.  Then you can use that path in the `crew whatprovides` command to figure out which package installed it.

thedamian:
No no. Ita your changes. I just didnt know how to use it.

thedamian:
Ill test it tobight or if someone else can test it ill a merge it.

thedamian:
Im just saying there use to be a lot of talk on this change and now theres none. Did you delete your last PR and resubmit?

uberhacker:
Yeah, I figured it was easier to just close out the old PR since the name change.  I found a slight bug (it actually doesn't affect the functionality at all) that hampers the efficiency but I can fix that tonight after I get off work. I'm really addicted to this Chromebrew project.  I have a lot of ideas flowing through my head right now of how I want to add to and improve the project.  I gravitated here as a disgruntled Crouton user which is kind of funny, actually.

uberhacker:
@thedamian: Should I be incrementing the version number of Chromebrew since we're changing the `crew` command?  If so, are there any basic semver rules the project has adopted?

uberhacker:
@thedamian: Did you ever get this working?  I just pushed up the fix I was talking about earlier so it should be ready to merge unless someone else has issues.  @jam7, @minektur, @lyxell, @cstrouse?

jam7:
I think @thedamian is talking about the possibility of reopening #329 instead of using this #334 since we discussed at #329 already and we may need to discuss same topics here again if we uses #334 instead.

uberhacker:
@jam7, @thedamian: I wanted to change the commit message, but since you cannot rewrite history (or at least not recommended unless you really know what you are doing), I was left with no other alternative.  There wasn't much discussion on the closed PR anyway.  It was @minektur basically saying he prefers the `whatprovides` command name over `source`.  With that said, do you think this PR can be merged?

jam7:
If so, I will leave my opinion here.  I personally wanna check them with someting like `crew whatprovides aio.h`, `crew whatprovides include/openssl` or `crew whatprovides *.la` way.  You can do it by using `system "grep #{name} /usr/local/etc/crew/meta/..."`.  (actually *.la doesn't work well with grep tho)

uberhacker:
@jam7: I just pushed up a commit that searches based on a regex pattern instead of the full path to the file.  The caveat is `*` acts as a literal, so instead of `crew whatprovides *.la`, you should enter `crew whatprovides .la` instead.  You can use the `^` and `$` in pattern searches, however.  So you could enter `crew whatprovides .so$` and it will only return files that end in `.so`, not `.so.0` or `.so.1` or anything else for that matter.  I wouldn't suggest a search like `crew whatprovides ^/usr`, however, since depending on the number of packages you have installed, it could take awhile.  Please test when you get a change and let me know if it works as advertised.
Thanks

uberhacker:
@jam7, @thedamian, @minektur: Go ahead and add stars in your search pattern.  I removed that restriction. :)

lyxell:
Is it possible to use many operators in the same search command? E.g. `^pkg*.h`

uberhacker:
Thanks for the feedback @lyxell.  Yeah, I knew there was probably a better way.  Not a ruby expert here. :/.  I put items in an array so I could sort the output.

uberhacker:
@lyxell: Thanks for the suggestion.  It really reduced the code. :)  Interestingly, I get 3 different results based on the regex patterns as shown below:
```
$ crew whatprovides \.pc
Total found: 530
$ crew whatprovides "\.pc"
Total found: 31
$ crew whatprovides "\.pc$"
Total found: 27
```
But this seems to be expected based on the patterns.  Please test again when you get a chance.
Thanks

uberhacker:
@lyxell ?

lyxell:
@uberhacker Looks good to me!

uberhacker:
:+1: merge away!

PR #339, opened 2017-01-06
Only require package name for removal

### LINKED ISSUE #335: package openssl_devel not found
As the package name is appended with 'devel', this shouldn't come as a surprise to me, but the package has been removed on the master branch.

This shouldn't be a problem to me, as I don't depend on the package (non-devel requirements), but some unexpected behavior made chromebrew unusable:

Whenever I'm trying to update or upgrade, I get notified that the `openssl_devel` package was not found & it refuses to update/upgrade anything. Same thing goes for trying to remove the package, which is more of an issue.

The actual issue:
It is not possible to remove unknown packages after they've dissapeared from the package list. 

Suggested fix:
Keeping a copy of the package file in a local directory to indicate it as installed & track the installed version (I don't know how crew currently keeps track of that) & using that to run the 'remove' command could resolve this issue. It would then also be possible to show the currently installed version of a package, as a bonus feature. ( Almost like described in #332 )

lyxell:
Fixes #335

### DIFF
--- a/crew
+++ b/crew
@@ -328,12 +328,15 @@ end
 
 def remove (pkgName)
 
+  #make sure the package is actually installed
   unless @device[:installed_packages].any? { |pkg| pkg[:name] == pkgName }
     puts "Package #{pkgName} isn't installed."
     return
   end
 
   Dir.chdir CREW_CONFIG_PATH do
+
+    #remove all files installed by the package
     File.open("meta/#{pkgName}.filelist").each_line do |line|
       begin
         File.unlink line.chomp
@@ -341,6 +344,7 @@ def remove (pkgName)
       end
     end
 
+    #remove all directories installed by the package
     File.readlines("meta/#{pkgName}.directorylist").reverse.each do |line|
       begin
         Dir.rmdir line.chomp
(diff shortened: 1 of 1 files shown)


lyxell:
Now handles packages marked as `is_fake` correctly.
This should be ready for merge.
Any additional thoughts? @jam7

crew
```diff
@@ -334,28 +334,31 @@ def remove (pkgName)
...
   end
 
-  Dir.chdir CREW_CONFIG_PATH do
-
-    #remove all files installed by the package
-    File.open("meta/#{pkgName}.filelist").each_line do |line|
-      begin
-        File.unlink line.chomp
-      rescue => exception #swallow exception
+  #if the package is marked as 'is_fake', the file/directorylists doesn't exist
```
jam7: It is possible to have both an `is_fake` mark and file/directory list if it was regular package before but someone decide to convert it to empty place holder with `is_fake` mark.  I tried it first when I made a place holder last time.  Anyway, this is a comment to comment.  Not a real issue.

lyxell: Can you try to reproduce that? It shouldn't be possible.
If a package gets upgraded, the filelist and directorylist gets removed via https://github.com/skycocker/chromebrew/blob/956e5e58a4084d34f033aade2dfb9f7e4df4cd92/crew#L351-L352
And if the package has turned into a fake package, crew will not create new filelist/directorylists.

jam7: I'm talking about update not upgrade.  update changes `.rb` files without file/directory list modification.

1. test.rb is a regular package and there are file/directory lists in users environment
1. test.rb is modified to `is_fake` at the master
1. `crew update` changes all `.rb` files without file/directory lists modifications
1. when user type `crew upgrade test`, there still be file/directory lists but `test.rb` has `is_fake` mark

Last time I tried to make `is_fake` empty place holder few hours ago, it happens AFAIK.

lyxell: Oh, yeah, you're right. Do we need to do anything about this?

jam7: Don't think so since we would like to remove files and directories anyway.  Maybe it is possible to write something we've considered it in the comment?  :)

lyxell: Updated the comment to reflect this fact.

lyxell:
I'm pretty confident in merging this. Is there any objections?

Issue #328, opened 2017-01-04
Why do we have pkg_config and pkgconfig?


uberhacker:
There are two versions of the same package?  Am I missing something here?  Shouldn't we only have one with the latest available stable version?

lyxell:
@thedamian

uberhacker:
@lyxell: You pointing fingers? LoL

lyxell:
:)

uberhacker:
https://github.com/skycocker/chromebrew/issues/344#issuecomment-271133351

thedamian:
Oh absolutely. If there was already a pkg-confit they I did a boo boo.

I'll delete it if it's the same.

uberhacker:
@thedamian: Before you delete `pkg_config`, make sure there are no other packages that currently depend on it.  If there are any, those will need to be converted over to use `pkgconfig` instead (if necessary).  We had a situation with `openssl_devel` after it was removed from the repo.  See #335.

uberhacker:
Can we close out this issue?

PR #341, opened 2017-01-06
Update pkgconfig and correct default search paths

jam7:
Now pkg-config searches all /usr/local/lib, /usr/local/lib64, /usr/local/share without PKG_CONFIG_PATH environment variable.  This searches lib64 on any architectures, but it will be not a problem since armv7l and i686 just ignore it.

Tested on both armv7 and x86_64, not on i686 (don't own).

I didn't know there are two identical package with different name until I read #328.  Only pkgconfig was depended by others, so I've just modified it.

### DIFF
--- a/packages/pkgconfig.rb
+++ b/packages/pkgconfig.rb
@@ -1,14 +1,15 @@
 require 'package'
 
 class Pkgconfig < Package
-  version '0.28'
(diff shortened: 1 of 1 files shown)


lyxell:
Nice!

PR #349, opened 2017-01-12
Change crew version to match release tag

uberhacker:
How about tagging another release tag of 0.4.1 so it matches what the crew command outputs? :)

After merging this PR, of course!


thedamian:
Ok.

PR #351, opened 2017-01-12
make ruby_latest depend on openssl and add the include directory

richard-fisher:
without openssl at compile time ruby will be unable to install gems from an https source


jam7:
That's just my 2 cents.  I can compile ruby_latest including openssl without CPPFLAGS option.  So, I guess only dependency to openssl is needed here.

richard-fisher:
Confirmed, only the dependency is necessary, I've updated the PR, thanks.

PR #353, opened 2017-01-12
Fix problem caused by #339

jam7:
`crew remove` doesn't remove actual files since it misses filelist path.


lyxell:
Ah, good catch! Thanks!

PR #260, opened 2016-12-13
Remove git dependencies

lyxell:
Remove the unnecessary dependencies for Git. In essence; if the user only wants Git and, lets say, a binary package, nothing else needs to be downloaded or installed. Earlier discussed in #167 

Tested on ARM, needs testing on x86.

PR #354, opened 2017-01-12
Add dependencies to make sure pkgconfig is installed prior to source code compilations

jam7:
This is my solution to @lyxell's question, https://github.com/skycocker/chromebrew/pull/347#issuecomment-271432465.

**Problem**: x86_64 really needs `pkgconfig` to figure out library path and others, but it's really really easy to forget adding dependency to `pkgconfig` since other architectures don't need it.

**Solution**: add `pkgconfig` as a part of `buildessential`.  This won't break other architectures since having pkgconfig helps configure process on not only x86_64 but also arm and i686.

Some packages forget to add `buildessential` dependency, so this PR doesn't solve problems on all packages.  Once #260 is merged and `buildessential` dependency is hardcoded, this will work correctly with all packages.

### DIFF
--- a/packages/buildessential.rb
+++ b/packages/buildessential.rb
@@ -8,4 +8,5 @@ class Buildessential < Package
   depends_on 'gcc'
   depends_on 'make'
   depends_on 'linuxheaders'
+  depends_on 'pkgconfig'
 end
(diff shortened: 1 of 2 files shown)


lyxell:
Good idea Jam, I think this is the most reasonable way to do it.

thedamian:
I like it. MERGED!

PR #363, opened 2017-01-14
Add package for xz-utils (5.2.3)

cstrouse:
Adds xz-utils (5.2.3). CrOS dev tools provides these tools primarily for reducing the size of image files to preserve what little disk space we have available on our tiny eMMC drives. For those who don't wish to run dev_install this will allow you to work with xz archives.


jam7:
works like a charm on both c100pa (arm) and c202sa (x86_64)

PR #366, opened 2017-01-14
Update Python from 2.7.12 to 2.7.13

cstrouse:
Update to 2.7.13 and also ensure that the pip module is being installed. This is suppoed to be default behavior when installing from either source or binary and for some reason wasn't happening.


thedamian:
with pip. I like it!

MERGED!

PR #367, opened 2017-01-14
fontconfig requires freetype2

jam7:
When I try `crew install fontconfig`, it fails because `freetype2` is not installed.  So, I added dependency.

Tested on ARM and X86_64.


cstrouse:
Your fix works for me on x86_64.

PR #368, opened 2017-01-14
Need some modification to install ffcall.

jam7:
When I try `crew install ffcall`, it fails because its makefile doesn't create intermediate directory to copy several files in `ffcall` package.  I add `mkdir -p #{CREW_DEST_DIR}/usr/local` to fix the problem.

Tested on X86_64.
`ffcall` itself has problems on ARM, so it is not tested on ARM.


cstrouse:
Works for me on x86_64 as well.

PR #371, opened 2017-01-14
Remove --with-tiff option to make mapserver compilable

jam7:
Add dependencies.

Need to remove --with-tiff option since mapserver supports only tiff3 but we need tiff4 to use BigTIFF.  We disable tiff option and use gdal to operate tiff files as described at http://lists.osgeo.org/pipermail/mapserver-users/2009-June/061792.html.

### DIFF
--- a/packages/mapserver.rb
+++ b/packages/mapserver.rb
@@ -5,8 +5,19 @@ class Mapserver < Package
   source_url 'http://download.osgeo.org/mapserver/mapserver-5.6.9.tar.gz'
   source_sha1 '3f05bf6c9a32b34408e36ddd32f61d68d65cf01c'
 
+  depends_on 'gdal'
+  depends_on 'libpng'
+  depends_on 'libjpeg'
+  depends_on 'proj4'
+  depends_on 'geos'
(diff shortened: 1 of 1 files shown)


jam7:
Need few more tries, so close this once.

jam7:
Updated.

uberhacker:
@jam7: Do we need to remove the `--with-tiff` option when we have a `libtiff` library package now?

jam7:
Yes.  We cannot compile mapserver with libtiff4.

>

uberhacker:
Should we have both libtiff4 and libtiff3 packages?

jam7:
Please do it if you want.  I don't care if you submit new PR.  My purpose is make this compilable.

uberhacker:
Do we need it?

jam7:
resolve conflicts

jam7:
I've noticed mapserver is compileable anymore after libgd.rb is updated in #483.  Updating mapserver to 6 or 7 may solve that problem, but I'm not sure.

I'm closing this PR since updating mapserver is not my intent.

Issue #355, opened 2017-01-13
Error using mandb


uberhacker:
```
$ sudo mandb -c
mandb: error while loading shared libraries: libgdbm.so.4: cannot open shared object file: No such file or directory
```
Possible fix?
```
$ sudo ln -s /usr/local/lib/libgdbm.so.4 /usr/local/lib/man-db/libgdbm.so.4
$ sudo ln -s /usr/local/lib/libpipeline.so.1 /usr/local/lib/man-db/libpipeline.so.1
```

lyxell:
Why sudo?

uberhacker:
Because I don't know what I'm doing? ;)

jam7:
This is common problem we have.  The sources of problem are:

- sudo doesn't pass LD_LIBRARY_PATH
- /etc/ld.so.cache is not updated
- mandb command doesn't have full rpath
- libgdbm.so and libpipeline.so don't exist at /usr/local/lib/man-db where mandb's rpath points

I think 3rd fixes as described in #49 is the best for all of us since it doesn't require remounting of root or a lot of symbolic links.  How do you think?

lyxell:
@uberhacker What errors do you get when omitting sudo?
@jam7 I agree with you, I think we should generally avoid symbolic links when possible.

uberhacker:
@lyxell: I get the same error without sudo.  I thought maybe adding it to the command would help but apparently, it doesn't make a difference.  You are welcome to try it out yourself to see if I'm not just talking out of my @$$.
Thanks

jam7:
@uberhacker My x86_64 works like this.  How about yours?
```
chronos@localhost / $ sudo mandb -c
mandb: error while loading shared libraries: libgdbm.so.4: cannot open shared object file: No such file or directory
chronos@localhost / $ mandb -c
mandb: warning: $MANPATH set, ignoring /usr/local/etc/man_db.conf
Processing manual pages under /usr/local/man...
mandb: can't chmod /var/cache/man/oldlocal/CACHEDIR.TAG: No such file or directory
mandb: warning: cannot create catdir /var/cache/man/oldlocal
mandb: can't create index cache /var/cache/man/oldlocal/1431: No such file or directory
mandb: can't update index cache /var/cache/man/oldlocal/1431: No such file or directory
done.
chronos@localhost / $ env | grep LD_LIBRARY_PATH
LD_LIBRARY_PATH=/usr/local/lib:/usr/local/lib64
[...]

PR #372, opened 2017-01-15
Fix mandb install so it 'just works'

uberhacker:
See #355.


jam7:
This is my personal opinion, but modifying user environment (.bashrc) quietly is not good.  I prefer old way, showing messages.  mkdir is OK since it is part of system environment, but .bashrc is...  For example, how do you think user performs install & remove & install...

On the other hand, I'm not sure about the new explanation of `mandb -c`.  Isn't it just making cache db?  man runs without it little slowly but it shows all newly added man pages without db updates.

uberhacker:
You have a good point about modifying the user environment.  It is quite possible (although not very likely) someone may prefer a different shell other than bash.  My thought process is to make it as easy as possible for the user to configure and use the command.  In regards to `mandb -c`, here is what `man mandb` says about it:
```
       -c, --create
              By  default,  mandb  will  try  to update any previously created databases.  If a database does not exist, it will create it.  This
              option forces mandb to delete previous databases and re-create them from scratch, and implies --no-purge.  This may be necessary if
              a database becomes corrupt or if a new database storage scheme is introduced in the future.
```
Having an updated database of manual pages is pretty important.  This makes sure everything is up to date so there are no surprises.

lyxell:
Isn't it more reasonable to use another directory instead of chowning the one in `/var`?

uberhacker:
@lyxell: Based on the errors when mandb attempts to update cache, those were the directories targeted.  Not sure if we can easily change the cache paths.  BTW, @jam7, I restored the PAGER env variable set up instructions in the latest commit.

uberhacker:
@lyxell: Is there any hope of this PR getting merged into master?  If not, what do I need to change?

lyxell:
Yeah, I'll try this out soon

uberhacker:
Closing due to unresolvable merge conflicts.

Issue #332, opened 2017-01-06
New package manifest proposal


uberhacker:
Hey guys,
What are your thoughts on creating a manifest of all the installed packages?  Basically, the idea is to store metadata about each package such as name, description, version, etc.  It would be kind of like `package.rb` but more human-readable and include data about every installed package.  I was thinking it would be nice to be able to list all the packages and a short description (for mere mortals).  I thought about maybe adding the description to each package but then you would need to parse each package everytime you wanted information.  Storing it all in one file would be more efficient.
Thanks

minektur:
We could just insert that information as some kind of tag or string in each
package.... and then somehow display that on demand... I haven't looked at
the code - how do we decide if somehting is already installed now?

Presumably you would want this to have the actual info for the actually
installed version so that if you updated crew and it's packages that the
information would still be correct - we'd need to store it at package
install time.   Stash it in a single file?  it would probably be easier to
make a 3rd manifest file for each package (right now we have a file list
manifest and a directory list manifest )...

It seems like a good idea to me.

lyxell:
Right now we are populating a json-file: `/usr/local/etc/crew/device.json`

uberhacker:
@lyxell: This is exactly what I had envisioned.  Not sure how I missed it.  Maybe because I was expecting a file named `packages.json` instead of `device.json`? :)  Do you think it would be difficult to sneak in a short description for each installed package as well?  I imagine if the .rb packages had a description, this would be easier.

lyxell:
@uberhacker Sounds like a good idea to me.

thedamian:
I love it.

who's going to take the lead on this?

I'll volunteer to change our http://skycocker.github.io/chromebrew/ to add a search or just pull from that .json file and list all the packages available and that short description.

uberhacker:
@thedamian: I'll take on the challenge!  The only concern I have is how to handle the transition.  We will need a way to elegantly parse existing packages (that folks have already installed) without a description as well as the new packages with the description.  We could just ask folks to update all their packages but that would be a bit extreme.  Maybe a script that inserts the description into existing packages that don't have one yet?

thedamian:
Yeah. Just create a file and copy the names, and title and descriptions as just the name of the package for now. You push that. I'll merge it right away and then we can either do a couple at a time or I'll make sure NOT to approve anything that doesn't also have their changes on your file. slowly we'll fill it up.

[...]

PR #373, opened 2017-01-16
Add packages manifest

uberhacker:
This PR addresses #332 and adds the packages manifest file packages.json.  The `crew info` command has been added to pull data from the manifest.  Also, install.sh was updated to copy packages.json to `$CREW_CONFIG_PATH`.  To test, execute `bash update.sh`.  Then type `crew info [package]` to verify the package manifest data is available.


uberhacker:
Anybody get a chance to try this out yet?  @jam7, @lyxell, @thedamian, @cstrouse, @prhero?

lyxell:
I still think it would be a lot more reasonable to use the .rb files we already have to store information.

uberhacker:
@lyxell: The problem with that approach is you will need all users to upgrade their existing packages in order to parse the new data.  If you can outline a plan that makes sense and gets this working going forward, I'm all ears.

lyxell:
We could use hash-sums of the package files to keep track of changes, this would also update the packages if we change the build steps without changing the package version, which would be good.

uberhacker:
Right, but that still doesn't address how this will work going forward.  I was hoping for an outline and a plan that works. :)

lyxell:
I could whip up a pull request if that's what you mean? @uberhacker

uberhacker:
@lyxell: Sure, you could pull the data from packages.json.  It was A LOT of work and I hope it was appreciated.

thedamian:
@lyxell : The problem @uberhacker addresses here is the details about the pages. This is needed

Putting that detail in the .rb file makes it easy for people doing a package but hard for the system to search (and for a search site that I plan to build) hard to parse. 

Having ONE package.json file that can be parsed by "crew" is easy (and done) and that ONE file can searched for with a basic javascript script in our github.io site or a full site to give that search that brew has.

thedamian:
I'm ready to merge but I like this conversation.

uberhacker:
What do you guys think about this proposal?  Why not just go ahead and use packages.json for now and we can still update all the .rb files?  Over time, we will eventually get them all populated and at that point we can start parsing the .rb files instead.  However, keep in mind, what @thedamian is stating above is very true.  It's easy (and faster) to pull data from one file rather than parsing potentially hundreds of .rb files to get a list of all the data.  Suggestion: Why not have both and the ability to automatically generate packages.json by parsing all the .rb files ftw?

thedamian:
@uberhacker I like the idea of automatically parsing, but we'd have to have another app to do that, and it can be then run by anything and the changes can be submitted as a PR.

someone will have to make it ٩(̾●̮̮̃̾•̃̾)۶

cstrouse:
You may want to look at how Google's done the packaging for [webports](https://chromium.googlesource.com/webports/+/master/lib/webports/bsd_pkg.py). They use the FreeBSD pkg-ng format. The manifest uses ucl (universal configuration language) and is pretty nice.

uberhacker:
@lyxell: Were you going to create a PR that replaces my proposal?  It's not easy, I know. :)  What do you think of @cstrouse's idea of Google webports?

uberhacker:
Hey guys,
@lyxell @jam7 @thedamian @cstrouse 
I have the entire code in place to provide package info.  A new command has been added to crew called `crew refresh`.  This function is also called anytime packages are upgraded.  It will parse all the `.rb` files in `chromebrew/packages`, extract out the package info and add it to `/usr/local/etc/crew/packages.json`.  If we can get all the info in the packages and start pushing up PRs, eventually we will have everything up to date.  Below is an example package summary with all the info populated:
```
require 'package'

class Php5 < Package
  description 'PHP is a popular general-purpose scripting language that is especially suited to web development.'
  homepage 'http://www.php.net/'
  type 'package'
  category 'development'
  version '5.6.29'
  source_url 'http://php.net/distributions/php-5.6.29.tar.xz'  # software source tarball url
  source_sha1 'df3301d11d56fbeabd8f6336e9c8f4c47af45562'       # source tarball sha1 sum
```
Let me know if this is a feasible solution or if you can think of a better way to do it.
Thanks

crew
```diff
@@ -209,30 +239,26 @@ end
...
+    @pkg.dependencies.each do |dep|
+      if not search(dep, true)
+        if not @dependencies.include? dep
+          @dependencies.push(dep)
+        end
       end
     end
   end
-  
-  pushDeps
```
jam7: Does this pushDeps refactoring works correctly?  What does it happens if A depends_on B, B depends_on C, and C depends_on D?  How did you test your modifications?

It looks like way too much modifications at once for me.

uberhacker: @jam7: Thank you for the input.  We don't need packages.json but since I already spent the effort to create it, we have the data ready to pull from there to the ruby packages.  You are correct about the recursive dependencies.  My thinking was only prompt for dependency packages if they aren't already installed.  Currently, the process doesn't take that into account.  I'll keep working on it.

uberhacker: @jam7: I refactored pushDeps so it recursively calls itself to drill down and get every possible dependency.  Only dependencies that have not been installed yet will be prompted.

uberhacker:
Replaced by #719.

PR #383, opened 2017-01-21
(FIX) Installing over an old crew install.

bfayers:
Otherwise you get errors like "Read error at byte 692/690 ((null)). Retrying." since -N -c is only download if newer + carry on from where left off it seems to get confused. Atleast on my machine it fixed the issue. Just incase others run into it.


thedamian:
where are you seeing an error? 

I literally just wiped by Chromebook pixel and I'm installing Node (where I found the node error I just fixed) but install.sh worked flawlessly.

what error are you fixing and when do you get it?

bfayers:
It fixes an issue if you run install.sh over a current crew install, i had so many conflicting issues in the end a powerwash and restart was better.

Issue #384, opened 2017-01-21
(ISSUE) (ARM) pkgconfig


bfayers:
I cannot install cmake because of pkgconfig.
```Archive downloaded
Unpacking archive, this may take a while...
gcc (GCC) 4.9.x-google 20150123 (prerelease)
Copyright (C) 2014 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Building from source, this may take a while...
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /bin/mkdir -p
checking for gawk... no
checking for mawk... mawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking build system type... armv7l-unknown-linux-gnueabihf
checking host system type... armv7l-unknown-linux-gnueabihf
checking how to print strings... printf
checking for style of include used by make... GNU
checking for gcc... gcc
checking whether the C compiler works... no
configure: error: in `/usr/local/tmp/crew/pkg-config-0.29.1':
configure: error: C compiler cannot create executables
See `config.log' for more details
pkgconfig failed to install: `./configure --with-internal-glib --with-pc-path=/usr/local/lib/pkgconfig:/usr/local/lib64/pkgconfig:/usr/local/share/pkgconfig CFLAGS=" -fPIC"` exited with 77```

however, the system already has pkgconfg 0.28. I also cnnot build 0.28 from source...

**Comment by bfayers** [comment]
@jam7 Ideas?

**Comment by bfayers** [comment]
+ openssl, if i comment out buildessential (pkgconfig is a dep of that) openssl fails like this


```
make[3]: Entering directory `/usr/local/tmp/crew/openssl-1.0.2j'
make[4]: Entering directory `/usr/local/tmp/crew/openssl-1.0.2j'
/usr/local/lib/gcc/armv7a-cros-linux-gnueabi/4.9.x-google/../../../../armv7a-cros-linux-gnueabi/bin/ld: cannot find /home/chronos/user/Downloads/tools/armv7a-cros-linux-gnueabi/lib/libc_nonshared.a
collect2: error: ld returned 1 exit status
make[4]: *** [link_a.gnu] Error 1
make[4]: Leaving directory `/usr/local/tmp/crew/openssl-1.0.2j'
[...]

PR #386, opened 2017-01-21
cmake requires ncurses, otherwise bootstrap fails

jam7:
When I try `crew install cmake` to check #384, I've just noticed ncurses is also required to build cmake.  So, adding dependency for it.

Tested on ARM.


thedamian:
Confirmed. I found you're exactly right!

MERGED!

Issue #337, opened 2017-01-06
Usage of shortarch


jam7:
It looks like several packages uses it, but is it working on all architectures?

## shortarch values

|arch|shortarch|
|---|---|
|armv7l|32|
|i686|32|
|x86_64|64|

## gcc availability

|arch|gcc -m32|gcc -m64|gcc|
|---|---|---|---|
|armv7l|not work|not work|work|
|i686|?|?|work|
|x86_64|not work|work|work|

I'm not sure how i686 gcc works.  X86_64 gcc accepts -m32 option but doesnt work correctly since we don't have 32bit glibc and stub-32.h.

Don't we need to specify -m flag at all?

## library files placements

|arch|lib32|lib64|lib|
|---|---|---|---|
|armv7l|not exist|not exist|exist|
|i686|may exist|not exist|exist|
|x86_64|not used|exist|exist|

I686 may have lib32 directory since some packages may use shortarch and place files there.  Armv7l crew automatically convert lib32 to lib on the fly in the middle of compilation, so lib32 doesn't exist.  X86_64 has lib32 for 32 bit gcc libraries but not used by others.

I think specifying lib32 or lib64 is not necessary, but not sure what we should do for search directories.  Using pkgconfig is one good solution I think.

jam7:
The latest pkgconfig searches both /usr/local/lib and /usr/local/lib64 on x86_64.  So now, basically it should be OK to let each library decide the install destination either lib or lib64 and link them through pkgconfig.

It means we can remove SHORTARCH use from many packages.

jam7:
Closing since #393 has been merged

PR #393, opened 2017-01-22
Remove all SHORTARCH related stuff

jam7:
Removing SHORTARCH variable from crew as discussed/described in #337.

This modifies bc, clisp, ffcall, flex, fontconfig, freetype, gdal, geos, imagemagick, libgd, libjpeg, libpng, libsigsegv, libtiff, libxml2, libxslt, mapserver, openconnect and proj4.

I tested them all and all other packages depend on them on X86_64.  I also tested all except ffcal and clisp on ARM (those two are not compilable on ARM)

### DIFF
--- a/crew
+++ b/crew
@@ -23,17 +23,6 @@ else
 end
 
 ARCH = `uname -m`.strip
-#SHORTARCH = `getconf LONG_BIT`.strip # will refactor this line in the future
-case ARCH
-  when "i686"
-    SHORTARCH="32"
-  when "x86_64"
-    SHORTARCH="64"
-  when "armv7l"
-    SHORTARCH="32"
-  else  
-    SHORTARCH="32"
-end
 
 $LOAD_PATH.unshift "#{CREW_LIB_PATH}lib"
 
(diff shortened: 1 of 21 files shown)


thedamian:
@jam7 I love this.

BUT... I think you shouldn't update the package. 

if you have the new crew that doesn't have SHORTARCH you don't have the package that uses it. and vice versa.

jam7:
@thedamian Yes, therefore I modified packages and crew.  I forgot to mention crew modification but I did.  ;)

lyxell:
Looks good to me!

cstrouse:
Looks good other than the tacking the suffix on the package versions. I'm with @thedamian on not liking that.

jam7:
The reason of incrementing the version number is this modification may change the place of libraries installed.  For example, libraries installed to lib64 previously may be installed to lib after this modification.  Both should work fine because we have pkg-config.  But, I just thought it is not bad to force to upgrade by incrementing the version number.

But, if it is not acceptable, I can revert them.  Please let me know.  Thanks.

cstrouse:
@jam7 that makes sense. I'm good with it.

jam7:
Resolve conflicts.

PR #399, opened 2017-01-25
Refactor logic for extracting archives

### LINKED ISSUE #394: problem with creating a package for lshw
While attempting to create a package for `lshw` I noticed a bug in the way `crew` handles determining the `topdir` for an archive. I've only ran into this problem so far once but it could potentially show up with other projects that do weird things with their tarballs.

The code running ends up being (see [crew.rb](https://github.com/skycocker/chromebrew/blob/master/crew#L290)):
```ruby
topdir = `tar -tf lshw-B.02.18 | sed -e 's@/.*@@' | uniq`.chomp!
```

Which results in the following:
```
chronos@localhost /usr/local/tmp/crew $ tar -tf lshw-B.02.18.tar.gz | sed -e 's@/.*@@' | uniq | head
lshw-B.02.18
lshw.spec
```

`crew` attempts to use `lshw.spec` rather than `lshw-B.02.18`

I can get things working properly by modifying the code to read:
```ruby
topdir = `tar -tf lshw-B.02.18 | sed -e 's@/.*@@' | uniq | head -1`.chomp!
```

Here's my package file so far:
```ruby
require 'package'

class Lshw < Package
  version 'B.02.18'
  source_url 'http://www.ezix.org/software/files/lshw-B.02.18.tar.gz'
  source_sha1 'c0240f5e53cf40769d52e316719e8010ea6bdea3'

  def self.build
    FileUtils.chdir("src") do
      system "PREFIX=/usr/local", "make"
    end
  end

  def self.install
    FileUtils.chdir("src") do
      system "PREFIX=/usr/local make", "DESTDIR=#{CREW_DEST_DIR}", "install"
    end
  end
end
```

Is adding `head -1` a workable fix or is there a better way to handle this?

cstrouse:
Extracting files from the downloaded archive were handled differently depending on their type. Tarballs ran tar twice; once to extract the files, and once to test for the containing directory name. Tarballs that had a file and a directory at their top level were causing problems with determining the topdir for the build step. This changes the archive extraction to work the same way for both zip and tar.

This fixes #394.

### DIFF
--- a/crew
+++ b/crew
@@ -268,27 +268,18 @@ def install
     topdir = ""
     Dir.chdir CREW_BREW_DIR do
       puts "Unpacking archive, this may take a while..."
-      if meta[:filename][-4,4] == ".zip"
+      Dir.mkdir("tmpzip") unless Dir.exist?("tmpzip")
+      if File.extname(meta[:filename]) == ".zip"
         system "unzip", "-qq", "-d", "tmpzip", meta[:filename]
       else
-        system "tar", "xf", meta[:filename]
+        system "tar", "xf", meta[:filename], "-C", "tmpzip", "--strip", "1"
       end
       if meta[:source] == true
         abort "You don't have a working C compiler. Run 'crew install buildessential' to get one and try again." unless system("gcc", "--version")
-        if meta[:filename][-4,4] == ".zip"
-          Dir.chdir "tmpzip" do
-            entries=Dir["*"]
-            if entries.length == 0
-              abort "empty zip archive: #{meta[:filename]}"
-            elsif entries.length == 1
-              topdir = "tmpzip/" + entries.first
-            else 
-              topdir = "tmpzip/"
-            end
-          end
-        else
-          topdir = `tar -tf #{meta[:filename]} | sed -e 's@/.*@@' | uniq`.chomp!
+        if File.extname(meta[:filename]) == ".zip"
+          abort "empty zip archive: #{meta[:filename]}" if Dir.empty? "tmpzip"
         end
+        topdir = Dir.entries("tmpzip") == 1 ? "tmpzip/#{entries.first}" : "tmpzip/"
         Dir.chdir CREW_BREW_DIR + topdir do
           puts "Building from source, this may take a while..."
           @pkg.build


crew
```diff
@@ -268,27 +268,18 @@ def install
...
-            else 
-              topdir = "tmpzip/"
-            end
-          end
-        else
-          topdir = `tar -tf #{meta[:filename]} | sed -e 's@/.*@@' | uniq`.chomp!
+        if File.extname(meta[:filename]) == ".zip"
+          abort "empty zip archive: #{meta[:filename]}" if Dir.empty? "tmpzip"
         end
+        topdir = Dir.entries("tmpzip") == 1 ? "tmpzip/#{entries.first}" : "tmpzip/"
```
jam7: What does happen if we have more than one directory here?  Does it abort around here?  Do we use tmpzip as a topdir?

cstrouse: It always uses tmpzip as a topdir.

cstrouse: @jam7 have you tried this out yet?

jam7: No.  Not tested this yet.

jam7:
working fine on ARM too.

jam7:
I've noticed a problem.  It is something like original zip support problem, tho.

In the middle of `crew install autoconf`, `/usr/local/tmp/crew` may contain
```
autoconf-2.69/
autoconf-2.69.tar.xz
dest/
m4-1.4.17/
m4-1.4.17.tar.xz
usr/
```
So, in the middle of multiple source packages compilation caused by dependencies, tmpzip may be occupied by multiple instances.  I think easiest way to fix this is to remove extracted source files each time.

jam7:
This doesn't work with binary package like perl also.  Do you have an interest to fix this?  I also would like to refactor crew install mechanism for another purpose.  So if you don't mind, I would.
```
perl-5.18.1-chromeos-x86_64.tar.gz  100%[===================================================================>]  15.21M  11.5MB/s    in 1.3s    

2017-01-29 23:15:47 (11.5 MB/s) - 'perl-5.18.1-chromeos-x86_64.tar.gz' saved [15952847/15952847]

Archive downloaded
Installing...
/usr/local/lib/ruby/2.0.0/fileutils.rb:1551:in `stat': No such file or directory - dlist (Errno::ENOENT)
        from /usr/local/lib/ruby/2.0.0/fileutils.rb:1551:in `block in fu_each_src_dest'
        from /usr/local/lib/ruby/2.0.0/fileutils.rb:1567:in `fu_each_src_dest0'
        from /usr/local/lib/ruby/2.0.0/fileutils.rb:1549:in `fu_each_src_dest'
        from /usr/local/lib/ruby/2.0.0/fileutils.rb:508:in `mv'
        from /usr/local/bin/crew:342:in `block in install'
<<<chop>>>
```

cstrouse:
@jam7 go for it.

cstrouse:
Closing in favor of #399.

PR #401, opened 2017-01-25
(UPDATE) Remove python34 & replace with python3

### LINKED ISSUE #397: Add python 3.6
[Python 3.6](https://www.python.org/downloads/release/python-360/) is the latest release and would be nice to have.

bfayers:
Changelog:

1. Remove python34.rb
2. Add python3.rb with python version 3.6.0

Why?

Latest python update. python3 is probably a better idea than python34 for python 3.6.0
Also resolves #397 

At some point I plan to update the binaries in the python.rb to 3.6.0 however I only have an arm chromebook so I would only be able to create an ARM binary, unless someone wants to help with that.


thedamian:
great idea. but name the package python3 not just the file.

jam7:
As described in #260, what we really need as binary distributions are only ruby and git.  It is possible to change both python3 and python to source distribution if we once apply #260.

cstrouse:
@bfayers Were there supposed to be changes to readline? That commit is empty.

thedamian:
@bfayers 

The line still says:

class Python27 < Package

when it should say
 
class Python3 < Package

to match the filename.

bfayers:
@thedamian solved.

@cstrouse no, don't know how that got there

PR #405, opened 2017-01-26
removed unnecessary dependencies related to git

jam7:
This is originally promoted by @lyxell at #260.

Git itself is distributed as binary form on all architectures and it depends on several run time libraries but all of them are available on chromeos out of the box.  So, all of them required at compile time is not necessary to use it.

### DIFF
--- a/crew
+++ b/crew
@@ -241,12 +241,12 @@ def resolveDependencies
   
   pushDeps
 
-  #add buildessential as dependency if we're building from source
+  # Add buildessential and solve its dependencies if any of dependent
+  # packages are made from source
   if @source_package > 0
     @dependencies.unshift 'buildessential'
     search 'buildessential', true
   end
-
   pushDeps
   
   return if @dependencies.empty?
--- a/crew
+++ b/crew
@@ -214,10 +214,21 @@ def resolveDependencies
   @dependencies = []
 
   def pushDeps
-    if @pkg.dependencies
-      @dependencies.unshift @pkg.dependencies
+    if @pkg.binary_url && @pkg.binary_url.has_key?(@device[:architecture])
+      # Use binary dependencies if it exists, use source dependencies otherwise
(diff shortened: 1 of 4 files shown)


cstrouse:
Can we bump the version, too? I need a newer version for Google's cloud SDK. Or should we have a separate package for 2.x+ git?

jam7:
I like idea to renew git to 2.x+.  Can someone compile it for x86?  I can compile git for both arm and x86_64 but not for x86.

jam7:
Anyway, please merge this after evaluation.  We can update git binary later.

jam7:
Closing this since I've found a way to bump git version.  BTW, pending PR makes modifications really difficult.

jam7:
Add git update to 2.11.0.  I provides binary package for both ARM and X86_64.  For the case of X86, it is required to compile git from scratch.  I modified crew, git.rb and others to compile git automatically.  #428 is required to compile from source.

This time, I add following features by this PR.

- Add `rdepends_on` to specify run-time dependencies.  For binary package, crew checks rdepends_on first.  If rdepends_on is not provided, crew checks depends_on.  For source package, crew checks only depends_on.
- Add dependency to `buildessential` if the package is source package for the ease of development.  It's really easy to forget to add such dependencies.
- Change `install.sh` to check availability of binary package of git.  It install only appropriate dependencies.
- Modify git.rb to give information how to compile it.

jam7:
I have two branches for the ease of evaluation.

- for distribution with git binary
```
wget -q -O - https://raw.github.com/jam7/chromebrew/only-git-test/install.sh | bash
```
- for distribution without git binary
```
wget -q -O - https://raw.github.com/jam7/chromebrew/only-git-source-test/install.sh | bash
```

jam7:
I'm going to close this PR once since there are several modification to crew and this confuse those modifications.  I'll submit similar PR later.

PR #407, opened 2017-01-27
Add glib-2.0 library

uberhacker:
GLib provides the core application building blocks for libraries and applications written in C. It provides the core object system used in GNOME, the main loop implementation, and a large set of utility functions for strings and common data structures.  This is also a dependency of the sshfs command.

### DIFF
--- a/packages/glib.rb
+++ b/packages/glib.rb
@@ -1,6 +1,6 @@
 require 'package'
 
-class Glib2 < Package
+class Glib < Package
   version '2.0.7'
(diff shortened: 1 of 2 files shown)


lyxell:
Maybe just call it glib?

cstrouse:
Works for me on Samsung XE50013-K01US (x86_64).

uberhacker:
@lyxell: Renamed glib2.rb to glib.rb.
@cstrouse: Thank you for testing!

PR #419, opened 2017-01-27
Add gc (7.2g) package

cstrouse:
gc is also known as the Boehm-Demers-Weiser conservative garbage
collector. This is a dependency for multiple developer tools that I am
creating packages for.

### DIFF
--- a/packages/gc.rb
+++ b/packages/gc.rb
@@ -1,0 +1,16 @@
+require 'package'
+
+class Gc < Package
+  version '7.2g'
+  source_url 'http://www.hboehm.info/gc/gc_source/gc-7.2g.tar.gz'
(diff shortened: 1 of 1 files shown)


uberhacker:
Nice, @cstrouse.  This is a good one!

kirbyfan64:
@cstrouse Is there a reason you didn't use version 7.6 (the latest) instead?

cstrouse:
Sorry for the delayed reply @kirbyfan64. The wording on the project's page made it seem like the 7.2g was their stable and 7.6 possibly a development version. I don't know much about that package so I just went with that. If you happen to use that library and think 7.6.0 is ok I'd be happy to update it.

PR #422, opened 2017-01-28
Add attr (2.4.47) package

cstrouse:
The attr package contains utilities to administer the extended
attributes on filesystem objects.


uberhacker:
Another nice utility!

cstrouse:
@uberhacker The only downside to acl on CrOS i that it requires a rebuild of coreutils for the bit to show up in `ls` output.

Issue #362, opened 2017-01-14
multi core compilation


jam7:
It is possible to use multiple cores embedded in CPU with -j flag of make.

For example, crew install php7 takes 34 min on c202sa which contains 2 cores.  It takes only 16 min with make -j2 (I modified php7.rb locally).

How can we use these multiple cores?  Modify crew or modify each package?  I wish we have good solution for all of us.

BTW, the number of cores is taken by `nproc` linux command.

cstrouse:
@jam7 I agree that adding additional jobs would be a great idea so long as it doesn't break things for anyone. Make is supposed to check the load and scale the number of processes based on load when just using `make -j` with no job count so do we really need to set that using the core count? I recommend that we go with `make -j`.

jam7:
Maybe that's fine.  I was thinking someone might want to use only few of their cores like 2 out of 4.

kirbyfan64:
Maybe I'm the only person this has happened to, but `make -j` frequently overestimates the power of my computer and locks the whole thing up. It tries to schedule far too many jobs at once for my 2-core.

jam7:
How about this?  Make `crew` to set the number of processor to CREW_NPROC variable and to add `-j#{CREW_NPROC}` as additional argument to `system "make"` command.  Then, we also make `crew` to prioritize $CREW_NPROC environment variable.

So, if user want to specify the number of processor to use, set above environment variable.  Otherwise, `crew` uses all processors.

jam7:
#430 is merged, so closing this issue.

jam7:
@kirbyfan64 please let me know if you have problem with #430.

jam7:
After applying #430, `crew install openssl` fails on flip c100pa.  It works on other machines AFAIK.  I've investigated and found the source of problem is to use -j# at install time.  I'm going to send PR for remedy.

In the mean time, we should either of following items.

- compile using only one core by default?
- specify additional flag to not use multi cores on only broken packages?
- use 2 by default?  (if using 4 cores is the source of problem)

How do you think?  Thanks.

lyxell:
@jam7 What kind of error is reported on c100pa?

jam7:
Openssl is compiled without any errors, but when I run a command using either libcrypto or libssl, system says libxxx is empty or something.  I can reproduce it and post mode details tonight.

jam7:
[...]

PR #430, opened 2017-01-28
Change to use -j option in make to reduce the compile time

jam7:
Discussed in #362.  Php7 compile time is only 16 min after applying this.  It took 30 min before.

Users can specify their own usage of multiple cores by CREW_NPROC environment variable.  If a user sets it 2, crew uses only 2 cores.


bfayers:
how many cores does is use by default? All avaliable?

jam7:
@bfayers all.  It uses what `nproc` command returns.

uberhacker:
Awesome!  My Chromebook is slow so this is a very welcomed change.

cstrouse:
Thanks @jam7 this is a very welcome improvement.

PR #432, opened 2017-01-29
Add p7zip (16.02) package

cstrouse:
7-Zip  is  a  file archiver with the highest compression ratio. The
program supports 7z (that implements LZMA compression algorithm), ZIP,
CAB, ARJ, GZIP, BZIP2, TAR, CPIO, RPM and DEB formats. Compression
ratio in the new 7z format is 30-50% better than ratio in ZIP format.

Tested as working on Samsung XE50013-K01US.

### DIFF
--- a/packages/p7zip.rb
+++ b/packages/p7zip.rb
@@ -1,0 +1,16 @@
+require 'package'
+
+class P7zip < Package
+  version '16.02'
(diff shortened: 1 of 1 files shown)


cstrouse:
Here's the log from `make test` passing 100%.
https://gist.github.com/cstrouse/cb4539f9268bb2359184cee327cac7ab

uberhacker:
Nice!  I like the gist of `make test`.  We should do this with all our packages.

PR #433, opened 2017-01-29
Add libnettle (3.3) package

cstrouse:
Nettle is a cryptographic library that is designed to fit easily in more
or less any context: In crypto toolkits for object-oriented languages
(C++, Python, Pike, ...), in applications like LSH or GNUPG, or even in
kernel space.

Tested as working on Samsung XE50013-K01US.

Log of successful test run:
https://gist.github.com/cstrouse/e984a02502e2903550b710eb67abab6d

### DIFF
--- a/packages/libnettle.rb
+++ b/packages/libnettle.rb
@@ -1,0 +1,18 @@
+require 'package'
+
+class Libnettle < Package
+  version '3.3'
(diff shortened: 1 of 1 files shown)


cstrouse:
Didn't see that we already had a package for nettle. Closing this and will reopen as an update to the existing nettle.

PR #444, opened 2017-01-29
Update install.sh

metaquanta:
Install finished "successfully" but with this output: https://gist.github.com/metaquanta/9c35d032373ac12bb35668db7c380006

I wasn't user chronos.
I don't think this is a particularly good idea either. Anyone who touches your computer is chronos. Even guests. I actually added NOPASSWD to chronos' sudoers line and then about 24h later this occurred to me.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -109,7 +109,7 @@ echo y | crew install git
 sudo ln -s /usr/local/bin/gcc /usr/local/bin/cc
 
 #This will allow a lot of things to work without sudo
-sudo chown -R chronos:chronos /usr/local
+sudo chown -R `whoami`:`whoami` /usr/local
 
 #prepare sparse checkout .rb packages directory and do it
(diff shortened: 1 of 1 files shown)


cstrouse:
@metaquanta sure everyone is chronos but no one has access to your cryptohome if they're logged in as themselves or in guest mode so they wouldn't have access to your installed crew anyway.

metaquanta:
Other users will have access to /usr/local. 
Regardless, it's good to support users besides chronos.

jam7:
I'm fine with this modification.  This doesn't break existing users who doesn't modify user name.  This also fix a problem for users who have been modified their user name.

jam7:
@metaquanta BTW, it should be something like below.  `whoami` doesn't return your group name.
```
sudo chown -R `id -u`:`id -g` /usr/local
```

PR #464, opened 2017-02-02
Add geoip package

uberhacker:
GeoIP Legacy C API is a dependency of the goaccess package.


uberhacker:
Replacing with PR of squashed commits

PR #475, opened 2017-02-03
Fix broken dependency

cstrouse:
The recent update to ncurses built the shared libraries deprecating the
ncurses_so package which ncdu was dependent on. This changes the
dependency to use the current ncurses package.

Tested as working on Samsung XE50013-K01US (x86_64).


cstrouse:
Closing as other packages also need changes for the new include search path.

jam7:
@cstrouse Don't you mind to re-open this PR since #490.  This PR works fine on arm now.

PR #476, opened 2017-02-03
Fix include path for ncurses-dependent packages

cstrouse:
This is a temporary fix until it's safe to assume everyone's using the
updated ncurses 6.0 with utf-8 support.

The other option would be to create a symlink from /usr/local/include/ncurses to /usr/local/include/ncursesw and leave the packages alone.

### DIFF
--- a/packages/iftop.rb
+++ b/packages/iftop.rb
@@ -9,7 +9,11 @@ class Iftop < Package
   depends_on "ncurses"
 
   def self.build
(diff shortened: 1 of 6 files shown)


cstrouse:
On second thought there will be too many changes so going with the link during ncurses install seems to be the way to go. Any reason not to?

cstrouse:
Closing in favor of #477.

PR #477, opened 2017-02-03
Symlink ncursesw (utf-8 version) to ncurses

cstrouse:
Many packages which rely on ncurses have trouble finding the library
headers with the utf-8 version (/usr/local/include/ncursesw) so rather
than edit a ton of packages we create a symlink after ncurses has been
installed.

### DIFF
--- a/packages/ncurses.rb
+++ b/packages/ncurses.rb
@@ -20,5 +20,6 @@ class Ncurses < Package
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    FileUtils.ln_s '/usr/local/include/ncursesw', '/usr/local/include/ncurses'
   end
 end


jam7:
Don't we make both ncursesw and ncurses?  It looks like there is no binary compatibility.  Linux distribution also provides ncurses-devel and ncursesw-devel.

cstrouse:
I've had it work fine as a replacement for every package I've compiled after a powerwash. The issue has been it not using the correct search path.

jam7:
However, ncurses-6.0/INSTALL describes:
```
    --enable-widec
        Compile with wide-character code.  This makes a different version of
        the libraries (e.g., libncursesw.so), which stores characters as
        wide-characters,

        NOTE: applications compiled with this configuration are not compatible
        with those built for 8-bit characters.  You cannot simply make a
        symbolic link to equate libncurses.so with libncursesw.so
```
If we create both ncurses.rb and ncursesw.rb, user and application can choice whatever they need.

cstrouse:
Didn't notice that. Good catch!

jam7:
FYI, if I compile iftop with the latest ncurses and your symbolic link patch, compiled binary uses /lib/libncurses.so.
```
$ ldd /usr/local/sbin/iftop
        libpcap.so.1 => /usr/local/lib/libpcap.so.1 (0xb6f7d000)
        libm.so.6 => /lib/libm.so.6 (0xb6f03000)
        libncurses.so.5 => /lib/libncurses.so.5 (0xb6ec7000)
        libpthread.so.0 => /usr/local/lib/libpthread.so.0 (0xb6eb3000)
        libc.so.6 => /usr/local/lib/libc.so.6 (0xb6dbd000)
        /lib/ld-linux-armhf.so.3 (0xb6fc1000)
```

jam7:
@cstrouse Don't you modify or close this PR?  I believe this PR is making ncurses package as a combination of ncurses_w and ncurses_so.  :(

uberhacker:
I think the reason we didn't have ncursesw and the other symbols is because the base nurses packages was compiled with the `--no-debug` configure option.  I needed the symbols to successfully compile goaccess.  See #463.

jam7:
The problem @cstrouse mentioned here is missing `/usr/local/include/ncurses`.

It is not exist since he modified ncurses.rb to install only ncursesw.  We have only `/usr/local/include/ncursesw` after that.

He tired to recover it by creating symbolic link from ncursesw to ncurses.  It works as remedy, but what compiled application does is reading ncursesw header files (the API is ncursesw v6) and links `/lib/libncurses.so.5.0.0` (it's v5 API).  I believe this will big problem later, so I'm claiming it here.

Regarding to your problem, let's discuss at your thread.

cstrouse:
Fixed by #490.

PR #482, opened 2017-02-03
Fix nano

Thesola10:
It looked like ncurses installs in /usr/local/include/ncursesw instead of /usr/local/include/ncurses. Fixed this case

### DIFF
--- a/packages/nano.rb
+++ b/packages/nano.rb
@@ -9,7 +9,7 @@ class Nano < Package
   depends_on 'ncurses' 
   
   def self.build                                                  # self.build contains commands needed to build the software from source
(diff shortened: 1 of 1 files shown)


cstrouse:
@Thesola10 This is due to including utf-8 support in the recent ncurses update. Multiple packages are affected by this. #477 addresses this in a way that doesn't require edits to every dependent package.

Thesola10:
Well I still had the problem nonetheless

Le ven. 3 févr. 2017 à 10:26, Casey Strouse <notifications@github.com> a
écrit :

Thesola10:
Nevermind will wait for #477

PR #443, opened 2017-01-29
Change crew to use same logic for zip and tar

jam7:
This PR is originally created to compile lshw #400 as we discussed in #394.  This PR will replace #399.

This PR has relations to #276 and #284 also.  I refactor crew carefully, so I think I'm not breaking #276.  But who knows, please test this PR carefully.  Thanks.

I tested on X86_64 using following scenarios.

- lshw #400, as a test of tar file containing files at top  (EDIT: need modification after https://github.com/skycocker/chromebrew/pull/443/commits/0c435ec6dd56fae1df3ffbf3086415e7ed723b66)
- perl, as a test of binary package
- automake and m4 as a test of multiple package installation at once
- Makefile.zip #276 as a test of zip file without directories inside
- a.zip #276 as a test of zip file with top directory

PR #488, opened 2017-02-04
Use multicore for only compile-time

jam7:
Modified it to use multiple cores for only compile-time as discussed #362 

This PR includes #443 also, so please apply #443 first.

### DIFF
--- a/crew
+++ b/crew
@@ -218,7 +218,9 @@ end
 def build_and_preconfigure (target_dir)
   Dir.chdir target_dir do
     puts "Building from source, this may take a while..."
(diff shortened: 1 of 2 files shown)


jam7:
Squashed modifications since #443 is merged.

jam7:
Change to use class instance variable instead of global variable.

jam7:
Squashed and cleaned codes.  Removed a mechanism to clean extract directory since it modifies crew behavior.  Hope this and #526 will be merged soon since `crew isntall ...` fails on some packages without them.

PR #489, opened 2017-02-04
Update golang to 1.7.5

### LINKED ISSUE #479: Go package tries to build with bootstrap with go installed
I have go installed.
```bash
chronos@localhost /usr/local/lib/crew/packages $ which go
/usr/local/lib/go/bin/go
```
One issue is that the code checks the output of `which go` and compares it to an empty string. When using backticks in ruby the output has `\n` appended. 

Another is that the package doesn't appear to have access to the user's bash configuration and so the GOROOT is not set. I've tried in both ~/.bashrc and ~/.bash_profile with no luck.

Last, after running `crew install go` and it in turn install `go_bootstrap` only `go_bootstrap` is reported by `crew search` as being installed even though they both were installed successfully.

I updated the go.rb to 1.7.5 but it tries to build with the boostrap even though it shouldn't as a working go compiler is present.
```bash
##### Building Go bootstrap tool.
cmd/dist
ERROR: Cannot find /home/chronos/user/go1.4/bin/go.
Set $GOROOT_BOOTSTRAP to a working Go tree >= Go 1.4.
go failed to install: `TMPDIR=/usr/local/tmp ./make.bash` exited with 1
```

jam7:
Modify instructions to build.  It doesn't use go but go_bootstrap for go > 1.4.

This will fix #479.

### DIFF
--- a/packages/go.rb
+++ b/packages/go.rb
@@ -1,25 +1,20 @@
 require 'package'
 
 class Go < Package
-  version '1.7.4'
-  source_url 'https://storage.googleapis.com/golang/go1.7.4.src.tar.gz'
-  source_sha1 '0fb305c827c8794cfda7e437befa6101a2d06b2e'
+  version '1.7.5'
+  source_url 'https://storage.googleapis.com/golang/go1.7.5.src.tar.gz'
+  source_sha1 '4a36ad56b11579a8d534817b70a9baf754e5369a'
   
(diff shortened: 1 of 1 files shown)


jam7:
Closing this since there is a correct way to build go with existing go as described in https://github.com/skycocker/chromebrew/issues/479#issuecomment-278619173.

PR #490, opened 2017-02-04
Change to instal both ncurses ncursesw

jam7:
This is a PR to fix ncurses problem.  This PR has conflicts to #477.  I was waiting for @cstrouse to modify his PR, but not.  So, I've just decided to submit it by myself.

I think this PR solves ncurses and ncrusesw problem.

What I did:
- This reverts ncurses.rb to install ncurses similar to previous ncurses.
- This add ncursesw.rb to install wide character ncurses for user who want it.
- This add dependency so `depends_on ncurses` requires both ncurses and ncursesw now.
- Incresed the version number of ncurses to force to reinstall it.

3rd modification is for the ease of development.  For example, less works well even if there is only ncurses but it choose ncursesw if both are installed.  It is difficult to decide which we should install.  Therefore, 3rd modification forces to install both if developers simply write `dependes_on ncurses`.  Then, each applications configuration script will choose what they want.

Problem:
- ncurses and ncursesw writes several identical files like man files.  So, removing one of them may cause problem.

Other options:
- It is also possible to use `--enable-widec` and `--disable-suffix`, but I'm not sure whether that works well or not.

### DIFF
--- a/packages/ncurses.rb
+++ b/packages/ncurses.rb
@@ -6,6 +6,7 @@ class Ncurses < Package
   source_sha1 'acd606135a5124905da770803c05f1f20dd3b21c'
 
   depends_on "diffutils"
+  depends_on "ncursesw"
 
   def self.build
     system './configure ' \
--- a/packages/ncurses.rb
+++ b/packages/ncurses.rb
@@ -1,7 +1,7 @@
 require 'package'
 
 class Ncurses < Package
-  version '6.0'
+  version '6.0-1'
   source_url 'ftp://ftp.gnu.org/pub/gnu/ncurses/ncurses-6.0.tar.gz'
   source_sha1 'acd606135a5124905da770803c05f1f20dd3b21c'
 
--- a/packages/ncursesw.rb
+++ b/packages/ncursesw.rb
@@ -8,6 +8,16 @@ class Ncursesw < Package
   depends_on "diffutils"
 
   def self.build
+    # Check ncurses doesn't conflict with ncrusesw
+    if File.exist? CREW_CONFIG_PATH + "meta/ncurses.filelist"
+      if `grep include/ncursesw #{CREW_CONFIG_PATH}meta/ncurses.filelist` != ''
+	puts
+	puts "PLEASE PERFORMS `crew upgrade ncurses` OR `sudo crew remove ncurses` FIRST"
+	puts
+	exit 1
+      end
+    end
+    # Build ncursesw
     system './configure ' \
 	    'CFLAGS=" -fPIC" ' \
 	    '--without-debug ' \
--- a/packages/ncursesw.rb
+++ b/packages/ncursesw.rb
@@ -1,6 +1,6 @@
 require 'package'
 
-class Ncurses < Package
+class Ncursesw < Package
   version '6.0'
   source_url 'ftp://ftp.gnu.org/pub/gnu/ncurses/ncurses-6.0.tar.gz'
   source_sha1 'acd606135a5124905da770803c05f1f20dd3b21c'
@@ -13,7 +13,8 @@ class Ncurses < Package
 	    '--without-debug ' \
 	    '--prefix=/usr/local ' \
 	    '--with-shared ' \
-	    '--with-cxx-shared '
+	    '--with-cxx-shared ' \
+	    '--enable-widec'
     system "make"
   end
 


finwo:
@jam7 Tested on Acer chromebook 14 (x86_64), release 8872.76.0 (stable channel).

This PR appears to be working for nano but devastating for less:

```
chronos:~% less <some file>
less: error while loading shared libraries: libncursesw.so.6: cannot open shared object file: No such file or directory
[127] chronos:~%
```

jam7:
@finwo Thank your for the try.  Is it possible to show me what `crew search | grep ncurses` appears on your test environment?  I make this PR to force crew to install both ncurses and ncursesw, but it looks like your environment has only ncurses.  So, I just want to make it clear.

cstrouse:
Works great for me just had to rebuild dependent packages. @jam7 Perhaps add a note to the user post-install that they may need to rebuilt packages as well as show the list of package names.

finwo:
@cstrouse A notice would probably not be sufficient for users with zsh instead of bash, as zsh refuses to start without rebuilding.

@jam7 Just recovered my chromebook, as my shell (zsh) refused to start after this fix.

Installed ncurses manually & now `crew search | grep ncurses` show (i) on both ncurses & ncursesw.
Surprisingly, `crew install nano` indicates it can't find `ncurses.h` anymore again.

jam7:
@finwo please performs `crew upgrade ncurses` first.  I guess it will fix your problem.

finwo:
@jam7 Must've forgotten that step. After upgrading ncurses, nano (a.k.a. a new program) installs successfully & zsh (a.k.a. old programs) can't find the ncursesw library anymore.

So it boils down to re-compiling the programs using ncurses.

jam7:
@finwo sorry, my suggestion was wrong.  Please remove both ncurses and ncursesw once, then install ncurses (ncursesw will be installed automatically).  I'll write down details soon.

finwo:
@jam7 Output:
```
chronos% crew remove ncurses
Ncurses removed!
chronos% crew remove ncursesw
Ncursesw removed!
chronos% crew install ncurses ; crew search | grep ncurses
Found ncurses, version 6.0-1
Following packages also need to be installed:
libsigsegv diffutils ncursesw
Do you agree? [Y/n]
libsigsegv already installed
diffutils already installed
building ncursesw
building ncurses
Ncurses installed!
(i) ncurses
(i) ncursesw
chronos%
```

jam7:
Story is this.  TL;DR

ncurses.rb 6.0 install ncursesw.
I modified ncurses.rb to 6.0-1 and reverted it to install ncurses.  I also made ncursesw.rb 6.0 to install ncursesw.  This ncurses.rb 6.0-1 depends on ncursesw.rb 6.0.  So, I thought crew install both.
When user performs `crew update` and `crew install nano`, crew installs only ncursesw.rb 6.0.  It doesn't upgrade ncurses.rb.  So, we have only ncursesw here.
When user performs `crew upgrade ncurses` after that, crew removes ncurses 6.0 but it is actually ncursesw.  So, crew breaks ncursesw at this time.  Then, crew isntalls ncurses 6.0-1.

To solve this, need to remove both ncurses and ncursesw once, then install both.

jam7:
@finwo yeah.  Now you can use zsh and nano without recompile them I think.

finwo:
@jam7 As the current nano was compiled with the new ncurses 6.0-1, so I've tried it with less which was compiled with the 'old' lib.

Less before the upgrade:
```
chronos% less
WARNING: terminal is not fully functional
Missing filename ("less --help" for help)
chronos% 
```

Less after the upgrade (not re-compiled):
```
chronos% less
Missing filename ("less --help" for help)
chronos% 
```

Seems to be working.

jam7:
I've modified ncursesw to warn if it finds conflicting ncurses.  Warning message is something like below.

```
$ crew install ncurses   # actually ncursesw
$ crew update
$ crew install nano
Found nano, version 2.7.1
Following packages also need to be installed: 
libsigsegv diffutils ncursesw binutils gmp mpfr mpc isl cloog glibc gcc make linuxheaders pkgconfig buildessential ncurses 
Do you agree? [Y/n]
y
...

PLEASE PERFORMS `crew upgrade ncurses` OR `sudo crew remove ncurses` FIRST
```

jam7:
@cstrouse Rebuild dependent packages is not needed if both ncurses and ncursesw are upgraded once.  So, I add warning messages like above.  Now, it works like below.  :)

```
$ crew install ncurses  # 6.0, so ncursesw actually
$ crew install less     # uses ncursesw
$ crew install zsh      # uses ncursesw
$ crew install nano     # fails
...
nano.h:92:20: fatal error: curses.h: No such file or directory
 #include <curses.h>
                    ^
compilation terminated.
...
$ crew update           # apply this PR
$ crew install nano
...
PLEASE PERFORMS `crew upgrade ncurses` OR `sudo crew remove ncurses` FIRST

$ crew upgrade ncurses  # ncurses is upgraded to 6.0-1 and ncursesw is installed
$ zsh                   # OK
$ less                  # OK
$ crew install nano     # OK
```

Thank you @finwo and @cstrouse for good suggestions.

finwo:
@lyxell Could this be merged somewhat soon? Currently I'm not getting updates & fancy new features without manually merging the repository locally.

PR #491, opened 2017-02-05
Update php5 from 5.6.29 to 5.6.30

uberhacker:
Add `--with-readline` configure option to enable interactive shell.
 
Test with `$ php -a`:

```
Interactive shell

php > echo 'test';
test
php > exit
```

http://www.php.net/ChangeLog-5.php#5.6.30

### DIFF
--- a/packages/php5.rb
+++ b/packages/php5.rb
@@ -1,9 +1,9 @@
 require 'package'
 
 class Php5 < Package
-  version '5.6.29'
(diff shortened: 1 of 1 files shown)


cstrouse:
Nice addition but should probably squash the commits before merge.

uberhacker:
@cstrouse: I cannot rebase on the current feature branch.  I would need to create a whole new branch.

PR #497, opened 2017-02-06
(ADD) Go unstable version

bfayers:
Add's a GO Unstable version currently 1.8rc3


bfayers:
I will now test installing this alongsie go 1.7.5 to see if they can work together if not would there be a way to add a flip of depends_on such as breaks_on or not_compatible_with perhaps?

bfayers:
closing until i can work this out.

PR #502, opened 2017-02-08
Switch to ```all.bash``` for existing go

seitensei:
Regarding #479: For building go using go, we ```all.bash``` is used instead of ```make.bash```.


minektur:
did they finally fix the build system so that the tests didn't expect to be
able to run stuff from /tmp and write files as root to / ?  This is why we
changed in the first place - we skipped the tests because the tests were
not well behaved.

PR #504, opened 2017-02-08
Update imagemagick from 6.9.2-10-1 to 7.0.4-7

cstrouse:
This is a general bugfix/maintenance release.

Tested on Samsung XE50013-K01US (x86_64).

### DIFF
--- a/packages/imagemagick.rb
+++ b/packages/imagemagick.rb
@@ -1,9 +1,9 @@
 require 'package'
 
 class Imagemagick < Package
-  version '6.9.2-10-1'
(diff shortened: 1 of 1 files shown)


nuovodna:
link to imagemagick package doesn't work

http://www.imagemagick.org/download/releases/ImageMagick-7.0.4-7.tar.xz

jam7:
It's gone.  We can use https://github.com/ImageMagick/ImageMagick/archive/7.0.4-7.tar.gz instead.
Or https://github.com/ImageMagick/ImageMagick/archive/7.0.5-0.tar.gz for the latest.

PR #518, opened 2017-02-14
Add libcap package

uberhacker:
Libcap implements the user-space interfaces to the POSIX 1003.1e capabilities available in Linux kernels. These capabilities are a partitioning of the all powerful root privilege into a set of distinct privileges.  See http://sites.google.com/site/fullycapable/.  This is a dependency of the cdrkit package.

### DIFF
--- a/packages/libcap.rb
+++ b/packages/libcap.rb
@@ -1,0 +1,25 @@
+require 'package'
+
+class Libcap < Package
+  version '2.25'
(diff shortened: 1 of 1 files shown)


uberhacker:
@lyxell: This is a dependency package for cdrkit which was already merged into master.

uberhacker:
Thank you @thedamian!  Now if I can only get my dvd rw drive to actually burn, then I might be onto something. :100:

Issue #429, opened 2017-01-28
Uninstall script?


yoshifan509:
Is there a way to make a script for ChromeBrew that uninstalls crew? I looked at the issue from 3 years ago, but I wouldn't really know how.

cstrouse:
@yoshifan509 you can either remove everything under /usr/local or if you don't like that idea you can concatenate all of the filelists under /usr/local/etc/crew/meta and remove all of those files plus the ones that install.sh creates.

jam7:
How about this script?
```
#!/bin/bash
for i in `crew search|grep '(i)'|grep -v ruby|sed -e 's/(i) *//'`; do sudo crew remove $i; done; sudo crew remove ruby
```
If we have this script as uninstall.sh next to install.sh, we can perform `wget -O - ...uninstall.sh | bash`

This script doesn't work correctly if a user install several ruby like ruby and ruby3.

dragon788:
Is there a reason other than that Homebrew does it that Chromebrew puts everything in /usr/local instead of in the user's home directory? Doing it in the home directory may be slightly safer to "clean", though on a multi-user system it would result in a bit of wasted space if multiple users had their own installs.

I believe Linuxbrew either defaults to the user's home directory (to avoid permissions issues with /usr/local) or can be set to a custom prefix.

jam7:
Any files under user's home directory are not exacutable in chromebook environment.  Therefore, we don't use home directory.

It is possible to use `/usr/local/.local` or something instead like @cstrouse said in #536.

uberhacker:
See #877.

dragon788:
If you really want to "uninstall" and you don't have Linuxbrew or anything else in `/usr/local` you can use `dev_install --uninstall` which will automatically wipe any non-hidden folders from `/usr/local`. After discovering it in the comments in an issue on this repo I've been using it to recover from any shenanigans that break my Chromebrew/Linuxbrew installs.

uberhacker:
@dragon788: I have never used `dev_install`.  Can you specify a directory other than `/usr/local` to uninstall?  Since we have the `#{CREW_PREFIX}` constant, we could override the `/usr/local` default directory.

yoshifan509:
fuck you all bitches.....

uberhacker:
And your point is?

Issue #340, opened 2017-01-06
The ruby version in install.sh does not reflect the one in ruby.rb


lyxell:
The ruby version in install.sh is `2.0.0p247`, while the one in ruby.rb is `2.0.0p247-chromeos1`.
This results in crew wanting to upgrade ruby when running `crew upgrade` on a fresh install even though ruby is not outdated.

thedamian:
I think this was done because of a typo and to help people wth older versions.

I also believe that ruby has always had an older version with the install.sh than with the package because we didnt want to recreate all the great work skycoker did but still wanted users to be able to upgrade to the latest and greatest.

thedamian:
I see that it was done when adding support for x64
https://github.com/skycocker/chromebrew/commit/17e8cf16ee1936b0b10f07fb2888503364fa9bfb

do you want to put a PR to push it (and Ncurses) back to without the '-cromeos1' since you found this? @lyxell

PR #522, opened 2017-02-16
Change to not only overwrite device.json but also update it if it exists

jam7:
`install.sh` overwrites existing device.json, so we need to install everything again if we executes it accidentally.  Moreover, as described in #236, if we remove `ruby` accidentally, we also need to install everything again from scratch.

This PR fixes it.  New `install.sh` creates device.json if it doesn't exist or update it if it exists.

Tested on flip (ARM).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -81,28 +81,30 @@ sudo ln -s `pwd`/crew $CREW_PREFIX/bin
 mkdir $CREW_LIB_PATH/lib && cd $CREW_LIB_PATH/lib
 wget -N -c $URL/lib/package.rb
 wget -N -c $URL/lib/package_helpers.rb
-#create the device.json file
+
+#create or update the device.json file
+ruby_version='2.0.0p247'
 cd $CREW_CONFIG_PATH
 if [ ! -f device.json ]; then
-  echo "Creating device.json newly..."
+  echo "Creating new device.json..."
   echo '{' > device.json
   echo '  "architecture": "'$architecture'",' >> device.json
   echo '  "installed_packages": [' >> device.json
   echo '    {' >> device.json
   echo '      "name": "ruby",' >> device.json
-  echo '      "version": "2.0.0p247"' >> device.json
+  echo '      "version": "'$ruby_version'"' >> device.json
   echo '    }' >> device.json
   echo '  ]' >> device.json
   echo '}' >> device.json
 elif grep '"name": "ruby"' device.json > /dev/null; then
   echo "Updating version number of existing information in device.json..."
-  sed -i device.json -e '/"name": "ruby"/N;//s/"version": ".*"/"version": "2.0.0p247"/'
+  sed -i device.json -e '/"name": "ruby"/N;//s/"version": ".*"/"version": "'$ruby_version'"/'
 else
   echo "Adding new information to device.json..."
   sed -i device.json -e '/    }$/s/$/,\
     {\
       "name": "ruby",\
-      "version": "2.0.0p247"\
+      "version": "'$ruby_version'"\
     }/'
 fi
 
--- a/install.sh
+++ b/install.sh
@@ -83,15 +83,28 @@ wget -N -c $URL/lib/package.rb
 wget -N -c $URL/lib/package_helpers.rb
 #create the device.json file
 cd $CREW_CONFIG_PATH
-echo '{' > device.json
-echo '  "architecture": "'$architecture'",' >> device.json
-echo '  "installed_packages": [' >> device.json
-echo '    {' >> device.json
-echo '      "name": "ruby",' >> device.json
-echo '      "version": "2.0.0p247"' >> device.json
-echo '    }' >> device.json
-echo '  ]' >> device.json
-echo '}' >> device.json
+if [ ! -f device.json ]; then
+  echo "Creating device.json newly..."
+  echo '{' > device.json
+  echo '  "architecture": "'$architecture'",' >> device.json
+  echo '  "installed_packages": [' >> device.json
+  echo '    {' >> device.json
+  echo '      "name": "ruby",' >> device.json
+  echo '      "version": "2.0.0p247"' >> device.json
+  echo '    }' >> device.json
+  echo '  ]' >> device.json
+  echo '}' >> device.json
+elif grep '"name": "ruby"' device.json > /dev/null; then
+  echo "Updating version number of existing information in device.json..."
+  sed -i device.json -e '/"name": "ruby"/N;//s/"version": ".*"/"version": "2.0.0p247"/'
+else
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -83,15 +83,28 @@ wget -N -c $URL/lib/package.rb
...
-echo '  "architecture": "'$architecture'",' >> device.json
-echo '  "installed_packages": [' >> device.json
-echo '    {' >> device.json
-echo '      "name": "ruby",' >> device.json
-echo '      "version": "2.0.0p247"' >> device.json
-echo '    }' >> device.json
-echo '  ]' >> device.json
-echo '}' >> device.json
+if [ ! -f device.json ]; then
+  echo "Creating device.json newly..."
```
uberhacker: Better wording would be `Creating new device.json...`.

install.sh
```diff
@@ -83,15 +83,28 @@ wget -N -c $URL/lib/package.rb
...
+  echo '  "installed_packages": [' >> device.json
+  echo '    {' >> device.json
+  echo '      "name": "ruby",' >> device.json
+  echo '      "version": "2.0.0p247"' >> device.json
+  echo '    }' >> device.json
+  echo '  ]' >> device.json
+  echo '}' >> device.json
+elif grep '"name": "ruby"' device.json > /dev/null; then
+  echo "Updating version number of existing information in device.json..."
+  sed -i device.json -e '/"name": "ruby"/N;//s/"version": ".*"/"version": "2.0.0p247"/'
```
uberhacker: I'm not a big fan of hardcoded version numbers.  Can this be extracted instead?

install.sh
```diff
@@ -83,15 +83,28 @@ wget -N -c $URL/lib/package.rb
...
+else
+  echo "Adding new information to device.json..."
+  sed -i device.json -e '/    }$/s/$/,\
+    {\
+      "name": "ruby",\
+      "version": "2.0.0p247"\
```
uberhacker: Can this version number be extracted instead?

jam7:
@uberhacker Ok.  Changed to follow your suggestions.  Thanks.

install.sh
```diff
@@ -81,17 +81,32 @@ sudo ln -s `pwd`/crew $CREW_PREFIX/bin
 mkdir $CREW_LIB_PATH/lib && cd $CREW_LIB_PATH/lib
 wget -N -c $URL/lib/package.rb
 wget -N -c $URL/lib/package_helpers.rb
-#create the device.json file
+
+#create or update the device.json file
+ruby_version='2.0.0p247'
```
jam7: The point here is we don't install actuall ruby package.

For example, we can make minimum bootstrap ruby and install it with a version number like "bootstrap".  Then, we can update ruby.rb to 2.4.0 safely.  @thedamian gave me this enlighten idea in #340.

Therefore, I want to hard code the version number in somewhere of install.sh.

uberhacker: I still think it makes sense to not hardcode the version.  The whole point of `device.json` is to store the versions of the _*ACTUAL*_ packages.  Why not be consistent?  What if we removed all the hardcoded links to ruby from `install.sh` and extracted from the ruby.rb package instead?  Would that make sense?

jam7: First, it's still consistent.  `install.sh` install ruby-2.0.0p347.  It is just there is a possibility to upgrade ruby to 2.4.0 after chromebrew installation.

If we decide to install the latest ruby in `install.sh`, we need to check whole installation process on every machines whenever we want to upgrade ruby.

Does this make sense?

jam7: On the other hand, it is possible to extract appropriate version number from the name of binary file.  However, their name is already not consistent.  So, I just decide it hard-coded.  ;)

uberhacker: OK, so let's not update device.json then.  It makes no sense to store a version that does not match the same package.  Consistency matters.  If we hardcode versions all over the place, it makes the whole process much more difficult to maintain.  I see nothing wrong with testing if the entire installation process with the version inside ruby.rb.  What kind of instabilities are introduced any time you upgrade any package?

jam7:
@uberhacker  It matches to the version `install.sh` installed.  Then, you can upgrade them to the latest if you want.

If we hard-code the version number and binary file in install.sh, it doesn't mean difficult to maintain.  It means we don't need to maintain install.sh while we upgrade actual packages safely and freely.

For example, when I upgrade newtoolchains, I needed to modify `install.sh` to download isl.rb and cloog.rb since `install.sh` doesn't automatically download them.  Another example, I made ruby-2.4.  This newly depends on gmp, so it is required to modify `install.sh` again.

uberhacker:
@jam7: I still think we shouldn't be updating `device.json` in `install.sh`. :)

jam7:
@uberhacker Sry, I'm not sure what you means by the work `update` here.  Even `install.sh` installs ruby-2.0.0p247 over existing ruby, but it should not not update device.json?  For what purpose?

uberhacker:
`device.json` should only include _packages_ that can be installed and uninstalled.  You cannot uninstall the ruby version from `install.sh`, correct?

jam7:
Yes we can.  Didn't you try `sudo crew remove ruby`?

uberhacker:
Well, then that is a bug.  You would make `crew` worthless if you did that.  It is the equivalent of running `apt-get remove apt` on a Debian system and then wondering why you couldn't install any more software. :)

jam7:
Then, how do you upgrade apt?

uberhacker:
If you uninstalled it, then you couldn't install it again unless you downloaded the `.deb` package and installed via `dpkg`.

jam7:
Currently, we can remove ruby but it is not possible to recover it.  Although it happens often in the middle of ruby upgrades.

Therefore, I'm modifying `install.sh` to add recovering method from such situation.

However, you are saying don't update device.json?

uberhacker:
Yes, I'm saying don't install ruby as a "package" in `install.sh` and store that version in `device.json`.  You shouldn't be able to uninstall ruby and kill crew.

jam7:
If I did so, users still can `crew install ruby` and `sudo crew remove ruby`.  So, such modification doesn't worth in my opinion.  On the other hand, it introduces new situation like somewhat ruby is installed.  It requires developers to consider about such new situation.  I prefer simple hard-coding over that.

uberhacker:
What happens if someone wants to uninstall crew?  Heaven forbid, but shouldn't that be an option also?  I'm a big fan of uninstall scripts that remove every trace of residue everywhere including any packages that were installed.  Should we add `uninstall.sh`?

jam7:
Yes, I think so.  `uninstall.sh` script is discussed in #429 once.

jam7:
Closing this PR since this is merged into #529

PR #523, opened 2017-02-16
Add Dart

kirbyfan64:


thedamian:
I don't like references to zips in people's drop boxes.

any way to do it from the official repo. Everyone can trust that.

cstrouse:
I'd like to have build logs for the binaries available somewhere as well. We can probably have some Travis build them and release to Github and then put build logs in another repo or create them as gists or something like that. I have a bunch of my GCP trial left too so if anyone wants me to launch a decently spec'd workspace to bulk build binaries for every package I'd be more than happy to get that rolling.

kirbyfan64:
Well, I was going to grab them from the official repo, but...

Chromebrew proviees *no* way of doing anything with the compiled binaries that arr downloaded. Dart's come in an "odd" format that needs to be toyed with before installation.

FWIW the script I used to build the tgz is available here: https://github.com/kirbyfan64/chromebrew_pkgs/blob/master/dart.sh

jam7:
@kirbyfan64 How about this template?  You need to fill `build` and `install` method.
```
require 'package'

class Dart < Package
  version '1.22.0'
  case ARCH
  when "armv7l"
    source_url "https://storage.googleapis.com/dart-archive/channels/stable/release/#{self.version}/sdk/dartsdk-linux-arm-release.zip"
    source_sha1 '9878f97a48526141e31955e9d72a6c68d677f205'
  when "i686"
    source_url "https://storage.googleapis.com/dart-archive/channels/stable/release/#{self.version}/sdk/dartsdk-linux-ia32-release.zip"
    source_sha1 '9c06e9ae5edfb4522fa457c3bf46f2513cbc07c8'
  when "x86_64"
    source_url "https://storage.googleapis.com/dart-archive/channels/stable/release/#{self.version}/sdk/dartsdk-linux-x64-release.zip"
    source_sha1 'ee6f91625758202f47ab8249fcf8df19b682ecbb'
  end

  def self.build
    system "./configure", ...
  end

  def self.install
    system "mkdir", "-p", "#{CREW_DEST_DIR}/usr/local/..."
    system "cp", "-r", "xxx", "#{CREW_DEST_DIR}/usr/local/..."
  end
end
```

thedamian:
@kirbyfan64  The problem with prepackages binaries is that it's hosted in a place outside the source code and outside of this repo. 

@jam7 you are an amazing person. I agree with you that something like this would be better. I think we could use somethign like that \to just copy the binaries (since that's what dart gives us. Binaries) to be copied instead of "built" and "installed"

PR #528, opened 2017-02-18
Remove .gitignore

jam7:
.gitignore contains ".DS_Store".  It may work on Mac but doesn't on others.  Commit log says:
```
commit 8b773b1b46d7a2b8b404f591f1768f022c7aa51d
    Add --no-check-certificate for wget in install script
```
So, I guess this file is added accidentally.

Personally, I like to not have project-wide .gitignore, since without it, I can write whatever there without considering commit or accidental removal when switching between branches.  So, if collaborators don't mind, please merge this PR.  Thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,0 @@
-.DS_Store


uberhacker:
@jam7: I think it would be a good idea to keep `.gitignore`.  It prevents developers from accidentally committing certain files that should never be included in the repo.  It think might even be a good idea to include the following also:

jam7:
@uberhacker IMHO, those are what you should put in your private ignore list.  In general, you need to ignore them on any or your working directory.  Do you want to ask all projects to put those stuff in every project-wide .gitignore?

Issue #434, opened 2017-01-29
Update perl from 5.15.3 to 5.24.1


cstrouse:
Perl could use an update.

lyxell:
Solved by #525

PR #529, opened 2017-02-18
Refactor/install.sh

jam7:
This PR acieves:

 - perforamance of execution
   - it takes only 30 sec to install chromebrew
   - install only git, ruby and libssh2
 - make it possible to recover from accidental ruby removal
   - executing install.sh over existing chromebrew recover ruby
 - separate management of packages and management of install.sh
   - update packages don't affect install.sh behavior
   - it is possible to update install.sh if developer wants
   - or it is also possible to leave install.sh as is until ChromeOS update causes trouble
   - moreover, it is possible to provide ruby.rb or git.rb as a source package if we want

This PR solves:

 - #236 partially since it is possible to recover from #236 with this PR
 - #434 and similar problems since we can make it as a source package

Tested on armv7l and x86_64 Chromebook.  Tested on i686 Cloudready too.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -30,47 +30,117 @@ for dir in $CREW_LIB_PATH $CREW_CONFIG_PATH $CREW_CONFIG_PATH/meta $CREW_BREW_DI
   sudo mkdir -p $dir && sudo chown -R $USER:$USER $dir
 done
 
-#cd into the brew directory, everything will happen there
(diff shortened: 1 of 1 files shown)


bfayers:
This is a great idea.

lyxell:
Great work as usual Jam!

PR #530, opened 2017-02-19
Update ruby to 2.4.0 with binaries

jam7:
Update ruby.rb to 2.4.0 instead of ruby_latest.rb, since it is possible to update ruby.rb itself while remaining install.sh to use ruby 2.0.0p247.  It may be possible to install ruby 2.4.0 in install.sh, but I don't test it.  If updating ruby.rb is not good, please let me know.

This PR also provides binaries for all architectures.

Tested on armv7l and x86_64 using chromebook and i686 using cloudready.  Passed `make check` on all architectures.


Cyan101:
Having both ruby and ruby_latest seems alright in my opinion

Issue #517, opened 2017-02-13
Autoconf fails to compile from Perl error.


ryanpcmcquen:
`perl` is installed.
```
checking for perl... /usr/local/bin/perl
/usr/local/bin/perl: relocation error: /usr/local/bin/perl: symbol __get_cpu_features, version GLIBC_PRIVATE not defined in file libc.so.6 with link time reference
configure: error: Perl 5.006 or better is required
autoconf failed to install: `./configure` exited with 127
```

jam7:
Is it possible to provide details like architecture or using only original gcc or newtoolchains?  Thanks.

ryanpcmcquen:
On an Acer R11. Same error on new toolchain and original.

jam7:
On your machine, does `uname -m` shows x86_64?

I tried to reproduce your problem on my c202sa (x86_64), but no luck.

It works fine without errors with either original gcc and newtoolchains.  When I googled your error message, it says something related to glibc.  So, I try to remove glibc to see what happens with system's glibc, but it also works fine.  What I did:

1. install original chromebrew from scratch (just remove all /usr/local/* stuff)
1. install autoconf
1. install newtoolchains with `for ...` described at #461 
1. remove and install autoconf
1. remove glibc.  then, remove and install autoconf
1. revert to original gcc
1. remove glibc.  then, remove and install autoconf

All tries work fine.

jam7:
You can try remove and install `glibc` and `perl`.  It may solve your problem.

ryanpcmcquen:
OK, I am out of dev mode for now.

jam7:
I see.  The dev channel and beta channel now has glibc-2.23 and it conflicts with glibc-2.17 or glibc-2.19.  Need to update chromebrew's glibc to 2.23 soon.

ryanpcmcquen:
Aha!

ryanpcmcquen:
Is there any workaround in the meantime?

jam7:
I have glibc223.rb and perl.rb in my repo.  They are not tested enough, but it works on x86_64 beta channel.  If you don't mind, try them.  I'll make PR soon.

ryanpcmcquen:
So I can just install from your repo?

jam7:
Ahh, never mind.  It is possible to switch my repo but switching back is difficult for now.  So, waiting PR is better I think.  I'll make a PR in a day or so.

ryanpcmcquen:
[...]

PR #533, opened 2017-02-19
Add new experimental package named glibc223 to support glibc-2.23

jam7:
Chromebook dev channel and beta channel is using glibc-2.23 now.  This new glibc requires libmvec_nonshared.a for x86_64 at compile-time, but chromebook doesn't have these .a files.  So, having glibc-2.23 may cause new problems like #517 and https://github.com/jam7/chromebrew/issues/2 describe.

It will be available in a month or two in stable channel.  So, I just made glibc223.rb to prepare for it.

How to use: perform `crew install glibc223`.  This overwrites existing `glibc`.
It is also possible to perform `sudo crew remove glibc` then `crew install glibc223` if you want.

How much test: Not tested enough yet.  I don't want to use dev/beta channel for daily purpose.  Tested little on x86_64.  Not tested yet on armv7l nor i686.  This is independent and experimental package glibc223.rb, so I think it is OK.

How to build your own glibc: Methods are available at https://github.com/jam7/chrome-cross.

### DIFF
--- a/packages/glibc223.rb
+++ b/packages/glibc223.rb
@@ -1,0 +1,15 @@
+require 'package'
+
+class Glibc223 < Package
+  version '2.23'
+  binary_url ({
+    armv7l: 'https://github.com/jam7/chromebrew/releases/download/newtoolchains/glibc-2.23-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


lyxell:
Good idea!

PR #534, opened 2017-02-19
Add `zlibpkg` dep to Emacs.

ryanpcmcquen:
It fails to compile otherwise.

### DIFF
--- a/packages/emacs.rb
+++ b/packages/emacs.rb
@@ -5,6 +5,7 @@ class Emacs < Package
   source_url 'ftp://ftp.gnu.org/gnu/emacs/emacs-25.1.tar.xz'
   source_sha1 '983e457971e3e3c8964d039c113033f98132b8a8'
 
+  depends_on "zlibpkg"
   depends_on "diffutils"
   depends_on "m4"
   depends_on "autoconf"


cstrouse:
zlibpkg is installed as part of a fresh chromebrew installation.

ryanpcmcquen:
I am hoping that this will get merged: https://github.com/skycocker/chromebrew/pull/529

In which case, `zlibpkg` will _not_ be part of a fresh install.

cstrouse:
It is failing to compile for me after a fresh install but for a different reason.
screenshot 2017-02-21 at 11 57 43 am

ryanpcmcquen:
If you are on `beta` or `dev`, this should fix it: https://github.com/skycocker/chromebrew/pull/533

jam7:
Currently zlibpkg is installed by install.sh, so not having zlibpkg dependency doesn't cause problems.

However, it's possible to remove zlibpkg by hand.  #529 also changes to not install zlibpkg by default.  And anyway emacs requires zlibpkg.  So, adding such dependencies don't cause any trouble.  I prefer to add such dependencies whenever we notice it.

ryanpcmcquen:
Any update here? #529 was merged so we really should merge this, it is broken right now.

ryanpcmcquen:
@lyxell will you take a look at this?

PR #538, opened 2017-02-23
Fix python34.rb to symlink /usr/local/bin/python

uberhacker:
If I `crew install python34`, the `which python` command returns version 2.7.13 instead of 3.4.1 as expected.  This PR recreates the symlink to `/usr/local/bin/python` to reference `/usr/local/bin/python3.4` instead.


lyxell:
As far as i know, `which python` is expected to return python27 on most linux distros for compability reasons.

uberhacker:
Whatever :)

Issue #487, opened 2017-02-04
Add `:build` options to specify compile-time only dependencies


jam7:
Hi, I'd like to add new functionality to write run-time dependency.  Let me know you guys thought.

Originally, crew has ability to write one package information file (`packages/*.rb`) for both source compile and binary distribution.  Skycocker's example at https://github.com/skycocker/chromebrew/wiki/Creating-a-package already has both.

If we write both, following updates will be much easier.

The problem is that it requires to write only one kind of dependencies for both.  For example, git requires perl and python at compile-time for extracting its perl and python API.  However, git doesn't requires perl or python until you want to use those API at run-time.  So, dependencies for compile-time and run-time are different.

We can write ruby in `packages/*.rb`, so it is possible to write something like:
```
  if self.binary_url && self.binary_url.has_key?(:"#{ARCH}")
    # run-time dependencies
    depends_on ...
  else
    # compile-time dependencies
    depends_on ...
  end
```
But, it is really really difficult.  I've spent 1 hour to write above example correctly.  :(

Or we can modify crew to understand something like:
```
  # compile-time dependencies
  depends_on

  # run-time dependencies
  rdepends_on
```

I prefer latter.  How do you guys think?  If you like this, I'd like to make a PR.

jam7:
I recently have a chance to check how homebrew does at http://www.rubydoc.info/github/Homebrew/brew/master/Formula.depends_on.  They does:
```
# only compile-time dependencies
depends_on "cmake" => :build

# compile-time and run-time dependencies
depends_on "libssh2"
```
It is very difficult to support all of homebrew ways, but maybe some?  How do you think?

`rdepens_on` is good enough as a dialect?  Or we would like to follow little what homebrew does?

jam7:
I think `:build` option is better for us all since `rdepends_on` requires new lines in package file, so it requires twice much lines for dependencies but `:build` option is not.  Moreover, it is easier to add more options like `:option` as homebrew does in future.

jam7:
Closing this sine #574 is merged.

Issue #140, opened 2016-01-17
Binary Packages or compiled?


thedamian:
Alright. it's time to brawl!

Do you create packages with binaries that install quickly but need a i686 (32bit) and x64 (64bit) versions?

Or do we have everyone compile everything. 

Binary: 
PRO - Fast, REALLY FAST! you can be done in maybe 2 minutes)
CON - 32bit and 64bit versions. No one realized java had it's 64bit version missing until I tried it (because the 32bit version works flawlessly)

Compile:
PRO - One version for everyone. If I test it on my 64bit version. your 32bit machine will work. And it's EASY for ANYONE to update to the latest version of a package. NODE updated? Just find the new link. find the sha1sum of the file and do a pull request. Everyone has the latest and greatest faster.
CON - You're compiling each and every single time. SLOW! I mean multiply EVERY computer with the time wasted compiling. Did I mention slow?

Thoughts? @skycocker @jbaum98 @Tivoli you guys have done the most work here. What do you guys think?

Tivoli:
I prefer compiling, not just for the easier updating, but as you get into more complex packages, like postgresql for example, that have multiple binaries for its tools would have to be supported and even more of a complex update process. With nodejs, you have npm and node that get installed, so you would have to do multi package updates with that possibly.

jbaum98:
I feel like we can emulate Homebrew here, which uses "formulas," which are compiled on every computer, as well as "bottles" which I think are precompoled binaries. You just can have a list of binaries and a way of determining whether or not they are applicable on the machine and compile it yourself as a fallback.

Tivoli:
I like that idea. :+1:

thedamian:
cool! In brew do the forumals have these "bottles" ready to go and fall back on them? or you choose "binary or compiled" somehow?

jbaum98:
I think if the precompiled "bottle" is available, it will be used unless the user specifies not to. They describe exactly how they decide which to use [here](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/FAQ.md#why-do-you-compile-everything]).

lyxell:
One idea could be to set up a server with a hook to this repository which compiled things automatically as soon as the master branch is updated, it's a bit of work but totally possible.

jbaum98:
Homebrew uses [Bintray](https://bintray.com/homebrew/bottles).

skycocker:
[...]

PR #539, opened 2017-02-23
Add `rdepends_on` rule to specify run-time dependencies.

jam7:
Implemented `rdepends_on` described in #487.

After this PR, we can say for example git depends on only `libssh2` at run-time as below.
```
  # compile-time dependencies
  depends_on 'zlibpkg'
  depends_on 'libssh2'
  depends_on 'openssl'
  depends_on 'perl'
  depends_on 'curl'
  depends_on 'expat'
  depends_on 'gettext'
  depends_on 'python27'     # requires python2

  # run-time dependencies
  # use system zlibpkg, openssl, curl, expat
  rdepends_on 'libssh2'
```

Algorithm this PR takes:
 - For binary installation, modified crew checks rdepends_on instead of depends_on.
 - For the ease of maintenance, new crew also checks depends_on if rdepends_on is not specified.
 - For source installation, new crew checks depends_on as old one is.

### DIFF
--- a/crew
+++ b/crew
@@ -286,16 +286,21 @@ def resolveDependencies
 
   def pushDeps
     if @pkg.binary_url && @pkg.binary_url.has_key?(@device[:architecture])
-      @check_deps = @pkg.dependencies
+      # Use binary dependencies if it exists, use source dependencies otherwise
+      if @pkg.rdependencies
+        @check_deps = @pkg.rdependencies
+      else
+        @check_deps = @pkg.dependencies
+      end
     else
-      # Use source dependencies
+      # Use only source dependencies
       @check_deps = @pkg.dependencies
       @source_package += 1
     end
     if @check_deps && !@check_deps.empty?
       @dependencies.unshift @check_deps
       
-      @pkg.dependencies.each do |dep|
+      @check_deps.each do |dep|
         search dep, true
         pushDeps
       end
--- a/lib/package.rb
+++ b/lib/package.rb
@@ -4,7 +4,7 @@ class Package
   property :version, :binary_url, :binary_sha1, :source_url, :source_sha1, :is_fake
   
   class << self
-    attr_reader :dependencies, :is_fake
+    attr_reader :dependencies, :rdependencies, :is_fake
     attr_accessor :name, :in_build
   end
   def self.depends_on (dependency = nil)
@@ -13,6 +13,14 @@ class Package
       @dependencies << dependency
     end
     @dependencies
+  end
+
+  def self.rdepends_on (dependency = nil)
+    @rdependencies = [] unless @rdependencies
+    if dependency
+      @rdependencies << dependency
+    end
+    @rdependencies
   end
   
   def self.is_fake


uberhacker:
@jam7: I'm not sure what value this adds.  Even with the `rdepends_on` line, you still need the compile-time package.  Am I correct?  I'm thinking just one `depends_on` line would be good enough?

jam7:
This PR assumes `.rb` package files which contains both binary package URL and source compile options.  For example, you have this at the very beginning.
```
require 'package'

class Git < Package
  version '1.8.4'
  binary_url ({
    armv7l: 'https://dl.dropboxusercontent.com/s/lnz5hmjv48d14f2/git-1.8.4-chromeos-armv7l.tar.xz',
    i686: 'https://dl.dropboxusercontent.com/s/g3binxopw5nfky1/git-1.8.4-chromeos-i686.tar.gz?token_hash=AAEWnMNBfozSIzLD1unbYGJzT4FGkEfJmLFQ-3uzydfT_A&dl=1',
    x86_64: 'https://dl.dropboxusercontent.com/s/i7vs9wfk94tsrzt/git-1.8.4-chromeos-x86_64.tar.gz?token_hash=AAHyvjayN7THoxlryZaxQ2Ejm9xyD6bZCqXZM81hYRC8iQ&dl=1'
  })
  binary_sha1 ({
    armv7l: '084a3b9bb90c572e7c5b12aae485715f145053e5',
    i686: '8c1bf4bcffb0e9c17bf20dd05981e86ea34d5d65',
    x86_64: '067cb6c36616ac30999ab97e85f3dc0e9d1d57f4'
  })

  depends_on 'zlibpkg'
  depends_on 'libssh2'
  depends_on 'perl'
  depends_on 'curl'
  depends_on 'expat'
  depends_on 'gettext'
  depends_on 'python'
end
```
At this time, we don't need rdepends_on.  This is what our chromebrew is currently.

If you decide to upgrade git to 2.11 but you have only x86_64 machine.  Then, you can write like this.
```
require 'package'

class Git < Package
  version '2.11.0'
  source_url 'https://github.com/git/git/archive/v2.11.0.tar.gz'
  source_sha1 'ca1187843b7dd2bb3b1c5e275f04c17fa70e7100'
  binary_url ({
    x86_64: 'https://github.com/jam7/chromebrew/releases/download/binaries/git-2.11.0-chromeos-x86_64.tar.xz',
  })
  binary_sha1 ({
    x86_64: 'd8345b00742cadf4f34585f61cf49da5d5f7a695',
  })

  # compile-time dependencies
  depends_on 'zlibpkg'
  depends_on 'libssh2'
  depends_on 'openssl'
  depends_on 'perl'
  depends_on 'curl'
  depends_on 'expat'
  depends_on 'gettext'
  depends_on 'python27'     # requires python2

  # run-time dependencies
  # use system zlibpkg, openssl, curl, expat
  rdepends_on 'libssh2'

  def self.build
    # need to build using single core
    system "make", "-j1", "prefix=/usr/local", "CC=gcc", "PERL_PATH=perl", "PYTHON_PATH=python2", "all"
  end

  def self.install
    system "make", "DESTDIR=#{CREW_DEST_DIR}", "prefix=/usr/local", "CC=gcc", "PERL_PATH=perl", "PYTHON_PATH=python2", "install"
  end
end
```
For users using x86_64 needs to install only `libssh2`.  For users using arm or i686, needs others to compile git from source.  Someone who own arm or i686 may be able to add binary file for them later.

jam7:
In addition, there is a discussion #140.

I'd like to say both.  As much as binaries but source for the case binary files are not available.  This PR fills a gap between them, I think.

uberhacker:
Oh, so you are saying with the `rdepends_on` line you don't need the `depends_on 'libssh2'` line so it will use the system binaries instead of compiling the package?  In your example above, you include the `depends_on` line also so I guess I'm not quite understanding.  That leads to another question that may need to be discussed in another issue.  What happens if you install a binary package and decide later you want to compile that package instead?  Would we have the option to choose between a binary or compiled package install?

jam7:
Depends_on means you need it at compile-time.  For example, you need openssl package to use its head files since Chromeos openssl doesn't provide them.  Once you compiled your program, you don't need header files.  Actually, having the latest openssl library is the best for the security point of view, but it is not necessary.

For the case of `perl` and `python27`, they requires to decide the version of API needed to provide at compile time.  They are not necessary at run time until users decide to use perl or python27 API.

The decision making between binary or compile is completely different story.  Current version of crew doesn't have such option.  Therefore, it is completely outside of this PR.  I personally already made such options, though..  I'll submit it later.

jam7:
rebased to solve conflicts.

jam7:
closing this one to modify user interface to match homebrew interface.

PR #547, opened 2017-02-28
Add aarch64 support

awafaa:
As ChromeOS user land is 32bit, we can re-use the existing builds. Only the kernel is 64bit. I've duplicated the binary and hash lines so they can act as place holders for when Chrome switches fully to 64bit.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -18,7 +18,7 @@ if [ $EUID -eq 0 ]; then
 fi
 
 case "$architecture" in
-"i686"|"x86_64"|"armv7l")
+"i686"|"x86_64"|"armv7l"|"aarch64")
   ;;
 *)
   echo 'Your device is not supported by Chromebrew yet.'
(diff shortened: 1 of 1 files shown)


jam7:
Is this modification required to run `chromebrew` on Chromebook Plus (OP1 processor)?  I've noticed it just a few min before.  If so, it's good to prioritize this PR since Plus hits the market already.

I've not tested this on aarch64 since I don't own it, but I've tested this on armv7l without any problem.  I've reviewed this one also.  I think this is fine to merge.

Alanaktion:
I can confirm this does work on my aarch64 machine. Package installation fails in most cases since the packages don't include aarch64 binaries, but PR #556 seems to fix that just fine.

awafaa:
Sorry for the delay in response, @jam7 yes this would support both the Samsung Plus and the Acer R13. Note @Alanaktion's comment about PR #556, I was hoping to work on why package installation was failing but was beaten to it.

PR #548, opened 2017-03-08
Add buildessential as dependency if we're building from source

jam7:
Currently, chromebrew doesn't install buildessential by default.  It is required to install it if we want to compile from source.  However, almost all packages don't have such dependencies.

This PR modifies crew command to automatically add buildessential dependency if not only target but also any of dependet packages of the target are build from source.

Tested on ARM and X64.  Tested on X86 also with Cloudready.

### DIFF
--- a/crew
+++ b/crew
@@ -279,9 +279,19 @@ end
 def resolveDependencies
   @dependencies = []
 
+  # check source packages existance
+  @source_package = 0
+
   def pushDeps
-    if @pkg.dependencies
-      @dependencies.unshift @pkg.dependencies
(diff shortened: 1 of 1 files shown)


lyxell:
Is there a specific reason to count the number of "source packages"?
Wouldn't something like this suffice? https://github.com/lyxell/chromebrew/blob/492b88e32d8cbf43878170c760de8345b1a8d2dc/crew#L180-L184

jam7:
@lyxell yes.  counting is not necessary, but checking all dependent packages is needed as you notice.

PR #551, opened 2017-03-10
Add gzip package

uberhacker:
Can't believe we don't have this package yet. :)  Adds gzip and gunzip commands.


uberhacker:
Wait..we do have it in /bin/gzip.  Nevermind.

PR #553, opened 2017-03-11
add less 481 binaries

jam7:
add less 481 binaries.   tested on arm and x86_64.  tested on x86 with cloudready too.


finwo:
@jam7 I assume the purpose is to reduce installation time?

jam7:
Yes.  I'd like to add more binaries for that purpose.

PR #554, opened 2017-03-18
Added bin heroku 4.28.3

finwo:
Yes, I know the current stable version is 5.7.3 (35ab144) and not 4.28.3. I just couldn't find a proper binary download for that version. ( I inspected https://cli-assets.heroku.com/ to determine the most recent version I could download ).

When starting the binary, it will automatically update itself to 5.7.3.

I was only able to test this on x86_64.

### DIFF
--- a/packages/heroku.rb
+++ b/packages/heroku.rb
@@ -1,0 +1,15 @@
+require 'package'
+
+class Heroku < Package
+  version '4.28.3'
+  binary_url ({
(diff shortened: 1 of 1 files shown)


thedamian:
Love it. I'd love the newer version but we'll take this cause this is VERY useful to developers

finwo:
@thedamian I'll look for a more up-to-date version of binaries for all 3 platforms later this week

Issue #555, opened 2017-03-19
issue installing cromebrew on samsung chromebook plus


mprorock:
`uname -m` identifies as aarch64

If however you add an arch entry for it that dupes the existing armv7l, or force 'architecture=armv7l' it works.

jam7:
Does #547 fix it?  We also need to duplicates all armv7l lines in every package though.

mprorock:
I went with a more temp workaround patch for now, alerting that things have not been fully tested yet...
[0001-fix-around-aarch64.patch.txt](https://github.com/skycocker/chromebrew/files/853249/0001-fix-around-aarch64.patch.txt)

jnsquire:
So far, I've noticed the libsigsegv isn't installing due to running it's own 'uname -m' invocation not recognizing the arch during configure.

jam7:
Does providing libsigsegv in binary format solve your problem?

jnsquire:
I was able to get it to build and install by forcing --build and --target to armv7a-cros-linux-gnueabi in the configure command used

thedamian:
Before I merge the fixes we have for this https://github.com/skycocker/chromebrew/pull/556

You guys have to discuss why not force it to armv7l
and ALSO

why I have to host the packages on the amazing @jam7 repo.

mprorock:
I am ok with forcing arm7l.  most binaries seem to run fine there, but I have noticed compile problems on several packages that run their own architecture detection.  I currently have armv7l forced on my system.

jam7:
@thedamian It's a good point.  I was thinking aarch64 can execute both armv7hf and aarch64 binaries.  So, I choose the way to not force it.  I'm not completely sure though.

mprorock:
it certainly can run aarch64 bins in a chroot env, as it is a 64bit kernel, but i have had issues with some binaries (non static ones for instance) since the userspace appears to be 32 bit at the moment.  not 100% sure on that though

jam7:
@thedamian Regarding to second part, hosting, I'm not sure your point.  Are you talking about binary files URL?  I simply use my github repo to post all binaries I submitted instead of my dropbox since those are too much these days.  If you want to copy them to under official repo or official bintray, please do it.  Only you and @lyxell can perform that, so we are waiting for you guys.  We have no permission to write to the official site.

If it is not what you point to, please let me know.

awafaa:
[...]

PR #556, opened 2017-03-19
Add aarch64 binary packages by duplicating existing armv7l packages.

jam7:
This is additional PR to #547.  The combination of this and #547 will fix problem described in #555.

This PR duplicates all armv7l binaries as aarch64 binaries to solve problem after the installation.  I don't own aarch64 machine, so I've not tested this on the real machine though.

### DIFF
--- a/packages/binutils.rb
+++ b/packages/binutils.rb
@@ -3,13 +3,15 @@ require 'package'
 class Binutils < Package
   version '2.25-1'
   binary_url ({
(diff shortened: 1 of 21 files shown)


Alanaktion:
This works wonderfully on my aarch64 machine. I had considered taking a different approach and writing `armv7l` to device.json, but this is probably a more future-proof way of handling the new architecture.

awafaa:
Can confirm this works perfectly on my Samsung Plus. Huge thanks for building upon my initial PR!

mprorock:
working on my side as well.

PR #570, opened 2017-04-10
Added bind version 9.10.4 and Subversion 1.9.5 (with dependencies)

ntchambers:
This currently works on x86_64. I don't have access to any other chromebook arch, but I don't see any reason why it wouldn't work.

### DIFF
--- a/packages/apr.rb
+++ b/packages/apr.rb
@@ -1,0 +1,18 @@
+require 'package'
+
+class Apr < Package
+    version '1.5.2'
+    source_url 'http://apache.claz.org//apr/apr-1.5.2.tar.bz2'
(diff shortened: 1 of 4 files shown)


minektur:
Is there any reason you have to specify the arch on the build line? It
would make life easier if it could autodetect.... It might do it
automatically... Or there are some packages that handle multiple
architectures....

Any chance?

ntchambers:
No. My apologies. I added it for some initial testing and forgot to remove it. Its updated with a proper build script.

ntchambers:
On my quest to package swift, I ended up making packages for apr, apr-util, and svn (which depends on the former too) and have added them to this pull request.

jam7:
Nice.  All of them works fine on armv7l too.

Issue #552, opened 2017-03-10
Automatic build binary package feature


jam7:
Do we like to have something like the topic?  It creates binary package for your architecture automatically from a package's `.rb` file.  For example,
```
$ crew build git
...
$ ls
git-2.11.0-chromeos-armv7l.tar.xz
git-2.11.0-chromeos-armv7l.tar.xz.sha1
```
We can build any binary packages easily with this.

It is possible to use this to create all binary packages of all `.rb` files if you have time and disk space for that. :)

I've modified crew to do that.  I'd like to hear any opinions about something like this feature before submitting a PR.  For example, is `build` command name common for something like this purpose?  Thanks.

finwo:
Yes, I would like this feature.

---

An extra request though: Would it be feasible to create binary packages out of structured folders (as we would when building packages now) by providing a directory path (of which the name is not a known package name)?

Binary package building would become a lot faster with this, as one would be able to simply download all required binary files & point crew to the directory for each architecture.

jam7:
It uses original mechanism for compile, so it extracts files under /usr/local/tmp/crew and build binaries there.  It moves files to /usr/local/tmp/crew/dest and modified crew make binary package file at working directory even if it is your home directory.

Does this what you want?

You explanation sounds like you would like to gather binary package files in particular directory and install packages from there instead of downloading...

finwo:
Your interpretation of my explanation is almost correct. I do want to gather the files in a particular directory.

Instead of installing from there, I'd like to build the arm, x86 & x86_64 archives based on the directories. ( automatically build the dlist & flist, archive them & output the checksums in separate files ).

This removes some manual steps from creating binary packages. I could try to build an example script for it, but I'm not that experienced with Ruby.

jam7:
Closing since #573 is merged.

PR #573, opened 2017-04-12
Add `crew build` command to build binary archives

jam7:
This PR adds feature described in #552.

If a user types `crew build package-name`, crew compile it from source as described in `package.rb` file and leave binary package and sha1 file at the current working directory like below.
```
$ crew build git
...
$ ls
git-2.11.0-chromeos-armv7l.tar.xz
git-2.11.0-chromeos-armv7l.tar.xz.sha1
``` 

There are two additional features.

 - Developer can specify the level of compression by `CREW_XZ_OPT` environment variable.  See xz manual for details.  Use "-7e" by default.
 - Developer can add aggressive test in the check method in each package.  This check method is executed only in this build mode.  So, we can add any kind of time consuming test cases here.  Even if it takes 2 days, regular `crew install` will be finished quick.

In order to add latter check method, need to modify `.rb` files.  For example, we can add python check method like below.  This test takes long time, but only users who use `crew build python` gets affected.
```
  def self.check
    system "TMPDIR=/usr/local/tmp http_proxy= https_proxy= ftp_proxy= make test"
  end
```

### DIFF
--- a/crew
+++ b/crew
@@ -22,6 +22,14 @@ else
   CREW_NPROC = ENV["CREW_NPROC"]
 end
 
+# Set XZ_OPT environment variable for build command.
+# If CREW_XZ_OPT is defined, use it by default.  Use `-7e`, otherwise.
+if ENV["CREW_XZ_OPT"].to_s == ''
+  ENV["XZ_OPT"] = "-7e"
+else
+  ENV["XZ_OPT"] = ENV["CREW_XZ_OPT"]
+end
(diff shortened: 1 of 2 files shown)


jam7:
Well, what is a problem to merge this PR?  I understand all of us are busy, so I don't want to rush us but...  If more explanation is needed or fewer modifications at once is required, please let me know.  Thanks.

PR #580, opened 2017-04-17
DBus + newer Glib + autoconf-archive + pango

flyingP0tat0:
I added DBus and autoconf-archive packages and modified Glib to use a newer version.

### DIFF
--- a/packages/autoconf_archive.rb
+++ b/packages/autoconf_archive.rb
@@ -2,7 +2,7 @@ require 'package'
 
 class Autoconf_archive < Package
   version '2017-03-21'
-  source_url 'https://ftp.wrz.de/pub/gnu/autoconf-archive/autoconf-archive-2017.03.21.tar.xz'
+  source_url 'http://ftpmirror.gnu.org/autoconf-archive/autoconf-archive-2017.03.21.tar.xz'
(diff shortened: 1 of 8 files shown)


packages/autoconf_archive.rb
```diff
@@ -0,0 +1,20 @@
+require 'package'
+
+class Autoconf_archive < Package
+  version '2017-03-21'
+  source_url 'https://ftp.wrz.de/pub/gnu/autoconf-archive/autoconf-archive-2017.03.21.tar.xz'
```
jam7: I think it's better to use public url `http://ftpmirror.gnu.org/autoconf-archive/autoconf-archive-2017.03.21.tar.xz` as described in https://www.gnu.org/software/autoconf-archive/Downloads.html.  This ftpmirror.gnu.org redirects us to the nearest mirror ftp site.

flyingP0tat0: Yes, my fault. Changed it.

flyingP0tat0:
Added some more packages, pango and its dependencies.

flyingP0tat0:
Alright, I feel like this pull request is somewhat unorganized. I will close this one and reopen multiple specific to the packages added.

PR #599, opened 2017-04-24
Update newlisp to 10.7.1

Kriskras99:
Update newlisp to 10.7.1, 10.7.0 is not available for download anymore.
Tested on armv7l

### Release notes:
This stable release adds new functionality to existing functions, adds one new function and fixes bugs.

**Additions and changes**
The now function on MS Windows, reports the type of daylight savings used in the GMT offset field.
The new history functions reports the call history of a function.
The maximum length of a symbol has been extended to 1022 characters from 255.
When creating symbols with a hash functor an error message will be thrown on symbol strings longer 1022 characters.
**Bug fixes**
The GMT offset field in the now function would report the wrong value when not in daylight savings time. This bug was introduced in 10.7.0 and did not exist before.
On some platforms file positions went out of sync when using search, seek and read-line.
mat with a scalar second argument was broken on 64-bit versions when the scalar parameter was not a float.
fixed a problem in --, ++ when passing a nil argument which was not content of a symbol.
**Compatibility with previous versions**
This version is compatible with the previous version 10.7.0.

### DIFF
--- a/packages/newlisp.rb
+++ b/packages/newlisp.rb
@@ -1,7 +1,7 @@
 require 'package'
 
 class Newlisp < Package
-  version '10.7.0'
+  version '10.7.1'
   source_url 'http://www.newlisp.org/downloads/newlisp-10.7.1.tgz'
   source_sha1 '724e7fd1c0512a4236fde022825dfd7ef859ca96'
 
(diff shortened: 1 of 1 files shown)


jam7:
Thank you for the update.

Just my 2 cents.  How about using `https://github.com/kosh04/newlisp/archive/10.7.1.tar.gz` as download URL?  I think URL under www.newlisp.org is not stable since it is often removed at the release of next version.  I think URL under github.com is little more stable.  ;)

Kriskras99:
That's a great idea, thank you.

PR #608, opened 2017-04-25
Add Libtasn1

skiman6010:
This is required to install GnuTLS since it was unbundled in the newer versions of that package.

I did not test this, just wrote it really quickly to get us moving in the right direction. We might need to change it up a little bit if needed for it to install correctly.


lyxell:
Closed in favour of #611

Issue #612, opened 2017-04-28
Checksum mismatch :/ try again


thedogb:
just as #93, i get same problem with vim again.
when i run 'crew install vim' ,i returns error information as fellow:
Found vim, version 8.0
Following packages also need to be installed:
glibc223 binutils gmp mpfr mpc isl cloog glibc gcc make linuxheaders pkgconfig buildessential readline zlibpkg perl libsigsegv diffutils ncursesw bz2 ncurses openssl python27 ruby
Do you agree? [Y/n]

Proceeding...
Found glibc223, version 2.23
Package glibc223 already installed, skipping...
Found binutils, version 2.25-1
Package binutils already installed, skipping...
Found gmp, version 6.1.2
Package gmp already installed, skipping...
Found mpfr, version 3.1.5
Package mpfr already installed, skipping...
Found mpc, version 1.0.3
Package mpc already installed, skipping...
Found isl, version 0.14.1
Package isl already installed, skipping...
Found cloog, version 0.18.4
Package cloog already installed, skipping...
Found glibc, version 1.0
Package glibc already installed, skipping...
Found gcc, version 4.9.x
Package gcc already installed, skipping...
Found make, version 3.82
Package make already installed, skipping...
Found linuxheaders, version 3.18
Package linuxheaders already installed, skipping...
Found pkgconfig, version 0.29.1
Package pkgconfig already installed, skipping...
Found buildessential, version 1.0
Package buildessential already installed, skipping...
Found readline, version 6.3
Package readline already installed, skipping...
Found zlibpkg, version 1.2.8
Package zlibpkg already installed, skipping...
Found perl, version 5.24.1
Package perl already installed, skipping...
Found libsigsegv, version 2.11
Package libsigsegv already installed, skipping...
Found diffutils, version 3.5
Package diffutils already installed, skipping...
Found ncursesw, version 6.0
Package ncursesw already installed, skipping...
Found bz2, version 1.4.17
Package bz2 already installed, skipping...
Found ncurses, version 6.0-1
Package ncurses already installed, skipping...
Found openssl, version 1.0.2k
Package openssl already installed, skipping...
Found python27, version 2.7.13
Package python27 already installed, skipping...
Found ruby, version 2.0.0p247-chromeos1
Package ruby already installed, skipping...
Found vim, version 8.0
No precompiled binary available for your platform, downloading source...
[...]

PR #613, opened 2017-04-30
Package updates and additions

ntchambers:
* added diffutils as a dependency to bind

* fixed flex's 410 error

* added serf HTTP library for subversion

* added scons build system

* updated subversion to depend on serf,
  because in my infinite wisdom my
  original build didn't include HTTP(S)
  support. Clang package to come soon!

* added openldap. This does not include
  slapd because I couldn't figure out
  how to build Berkely DB4.

* added pacparser

### DIFF
--- a/packages/bind.rb
+++ b/packages/bind.rb
@@ -9,6 +9,7 @@ class Bind < Package
   depends_on "openssl"
   depends_on "libcap"
   depends_on "readline"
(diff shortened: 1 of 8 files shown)


ntchambers:
Second commit is to fix #612

lyxell:
Thanks!

Issue #615, opened 2017-05-02
Web Server


Vannrith:
Hi, thanks for your work, Chromebrew.
this is not an error or issue ..
I just have a question. How can I install Apache and full lamp stack?
I found nginx and php7, but still not enough, I want Apache and mysql as well.
thanks again :)

ntchambers:
I am working on mysql right now for #211. I'll add apache to the list.

Vannrith:
i'd love to know when it's ready.
thanks so much

ntchambers:
I was able to get apache built (I'll submit a PR later tonight), but I'm still having trouble with mysql/mariadb, so that will have to come at a later time

Vannrith:
it's near ^^

thedamian:
@Vannrith httpd added in https://github.com/skycocker/chromebrew/pull/618

Vannrith:
thanks so much

dihuta:
@ntchambers 
Any update on MySQL?
Thank you.

flyingP0tat0:
@dihuta Working on it, but will take some time. Please use the open MySQL issue for that.

hutzlibu:
I installed apache ... but how do I start it? (And how to auto-start it with login?)
Also in general, where do packages get installed to?

uberhacker:
Hey @hutzlibu: All great questions and we should definitely do a better job of documenting this.  PRs are certainly welcome.
How do I start it?
```
$ sudo /usr/local/apache2/bin/apachectl start
```
How do auto-start it with login?
```
$ echo 'sudo /usr/local/apache2/bin/apachectl start' >> ~/.bashrc
```
Where do packages get installed to?  Usually the files of most packages are installed in a subdirectory of `/usr/local` which is the default prefix for most configure scripts.  In the case of httpd, it is installed in `/usr/local/apache2`.  Other things worth mentioned about httpd:
How do I stop the server?
```
$ sudo /usr/local/apache2/bin/apachectl stop
```
Where is the document root?
```
/usr/local/apache2/htdocs
```
How do I view files served by Apache?
```
Visit localhost or 127.0.0.1 in your browser
```

hutzlibu:
Thanks a lot for the detailed answer.
I think this belongs into Wikipage, no?
I will try to create a PR (never done that before)

uberhacker:
We could create a wiki page or it might even be a good idea to create something like README.md and add it to the package.

Issue #132, opened 2016-01-03
New Feature Request Thread


thedamian:
I suggest making only ONE thread for new features so that we don't fill the "issues" section with REQUESTS.

Also I'll try to make a Wiki (unless someone beats me to it) to show how to create a new Package yourself.

So Please feel free to use this thread to make a request of a package you want. WHERE we can find it. and WHY you need it.

I suggest closing all the other request Issues unless Skycocker likes having 26 open issues where a lot of them are just a request for X package or Y package.

Kriskras99:
Good idea, here are the current package/feature requests:

Current open package requests:
#10 [Cups](https://www.cups.org/)
#27 Man-pages ([man-db](http://man-db.nongnu.org/) and [less](https://www.gnu.org/software/less/))
#63 [Mosh](https://mosh.mit.edu/)
#66 [rdesktop](http://www.rdesktop.org/) / [freerdp](http://www.freerdp.com/)
#128 [xclip](http://sourceforge.net/projects/xclip/)

Current open feature requests:
#4 ARM support
#48 Error handling when download fails
#49 Recompiling binaries with rpath
#68 More binaries
#70 Support for tinycore packages
#72 Better error handling when git gives an error

ealgase:
One package I'd like to add is the X windows manager.  You could also install Xfce and Gnome as separate packages.

jabyrd3:
I'd kill for arm support. Can we raise a bounty or something?

bfayers:
> ealgase commented on 8 Feb
One package I'd like to add is the X windows manager. You could also install Xfce and Gnome as separate packages.

I'm sure this can be done relatively easily, just with some compiling, heck you could probably do it yourself? Gnome has its source avaliable anyways. [https://github.com/GNOME/gnome-desktop](https://github.com/GNOME/gnome-desktop) [https://github.com/GNOME/gnome-shell](https://github.com/GNOME/gnome-shell) or is it harder than i think?

bfayers:
>ealgase commented on 8 Feb
One package I'd like to add is the X windows manager. You could also install Xfce and Gnome as separate packages.

>bfayers commented 3 days ago
I'm sure this can be done relatively easily, just with some compiling, heck you could probably do it yourself? Gnome has its source avaliable anyways. https://github.com/GNOME/gnome-desktop https://github.com/GNOME/gnome-shell or is it harder than i think?

@ealgase 
[...]

Issue #63, opened 2015-03-02
Package mosh client


colemickens:
Hello,

I attempted to do this myself but ran into enough pitfalls that I decided to just file an issue and see if someone else has the chops for it.

I'd like to see "mosh" built for Chromebrew. I got part of the way there and then got stuck trying to get protocol buffers to build for Chrome OS.

thedamian:
mosh added in https://github.com/skycocker/chromebrew/pull/618

PR #618, opened 2017-05-04
Fix for a few requests

ntchambers:
* added mutt client for #132 . Currently, `make install` ignores errors,
  since it tries to install to /usr/local/share/man/man5 which is owned
  by root.

* added google protobuf (needed by mosh client)

* added mosh for #63. Although it is provided natively via
  https://chrome.google.com/webstore/detail/mosh/ooiklbnjmhbcgemelgfhaeaocllobloj

* added boost v1.59.0 needed by MySQL (yet to be built)

* added apache httpd for #615. Currently, `make install` ignores errors,
  since it tries to install to /usr/local/share/man/man8 which is owned by
  root.


thedamian:
awesome PR

ntchambers:
Thanks! More to come soon (next on my list is clang, mariadb/mysql, firefox, and swift)

bfayers:
@ntchambers Firefox? how are you going to display that??

PR #626, opened 2017-05-07
Change to not create static libraries for p11kit

jam7:
Changed to not create static libraries to increase disk space. :)

Tested on armv7l and x86_64.


thedamian:
Should all these "change to not create static libraries"  be done in ONE pull?

opinions?

jam7:
I have no opinion about it.  I used to separate them in my repo for the ease of maintenance.  If one PR is better, I'll merge them.

PR #629, opened 2017-05-07
Update zlibpkg to 1.2.11

jam7:
Update it and add build, install and check methods to compile it from source code.
Also changed to not install static libraries to increase disk space. :)

Tested on armv7l and x86_64.


thedamian:
I am always happy to have a package built from source! 

MERGED!

PR #630, opened 2017-05-07
Change to remove previous version just before install when it upgrades

jam7:
This solves the problem #236.

`crew upgrade` sets @pkg.in_upgrade true while it is upgrading target package instead of removing packages at the beginning.  Then, `install` function recognizes the flag and removes target package just before the install.  Therefore, even if it fails in compile, the target package is not removed.


thedamian:
merged

PR #640, opened 2017-05-10
Update git to 2.13.0

jam7:
It is not need to distribute binary packages, so I add source compile methods and update git to 2.13.0.

Tested on armv7l.

### DIFF
--- a/packages/git.rb
+++ b/packages/git.rb
@@ -1,25 +1,26 @@
 require 'package'
 
 class Git < Package
-  version '1.8.4'
-  binary_url ({
-    aarch64: 'https://dl.dropboxusercontent.com/s/lnz5hmjv48d14f2/git-1.8.4-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


thedamian:
Damn. I always forget.

I think we have issue3s updating GIT, and anything else in the original files that make the initial install.

am I wrong?

jam7:
Thank you for the review.  You are partially wrong.  It was the problem before.  Recently, we modified install.sh to not use .rb file at install.  Install.sh extracts tar files directly.  Therefore, what we need now is to keep the original tar files of git/ruby/libssh2 as is and make new .rb files freely.  :)

thedamian:
@jam7 thank you for the explanations.

I love it! 
I'm going to bring in the tars from dropbox (I think it's YOUR drop box, correct?) into github.

ANY OBJECTION on that?

ANd of course this is approved. I always want GIT updated.

jam7:
@thedamian No objection. 
 git/ssh2/ruby binaries for armv7l are in my dropbox.  Others are in skycocker's?  I'm not sure.

PR #642, opened 2017-05-11
Update pcre to 8.40 and optimized for size.

jam7:
Tested on armv7l.


thedamian:
I love ANYTHING where we go from a tarball to source!

PR #643, opened 2017-05-11
Update make to 4.2

jam7:
Add source compile and check methods also.

Tested on armv7l.


thedamian:
Such Amazing PR's @jam7 
You're amazing! That or a whole team in south america!!!

PR #176, opened 2016-06-01
Remove need for sudo

minektur:
For install, assumes that /usr/local is already writable by chronos and is rwx
chronos:chronos

fixes #172

(and a bunch of auto-changed whitespace...)

There are migration concerns - anyone who installed previously would need to manually make chronos own all the stuff that got installed.

roughly:

chown -R chronos:chronos /usr/local/*

PR #546, opened 2017-02-28
Remove need for sudo

uberhacker:
This is a reissue of PR #176.  Installs cleanly on a power washed Chromebook with specs below:
```
Version 56.0.2924.110 (64-bit)
Platform 9000.91.0 (Official Build) stable-channel kip
Firmware Google_Kip.5216.227.78
```
Hopefully, this will take less than 8 months to get merged into master. :)

PR #645, opened 2017-05-12
Update crew and install.sh to remove the need for sudo

uberhacker:
This is a reissue of PR #176. Installs cleanly on a power washed Chromebook with specs below:

Version 56.0.2924.110 (64-bit)
Platform 9000.91.0 (Official Build) stable-channel kip

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -25,9 +25,12 @@ case "$architecture" in
   exit 1;;
 esac
 
+#This will allow things to work without sudo
+sudo chown -R `id -u`:`id -g` /usr/local
+
 #prepare directories
 for dir in $CREW_LIB_PATH $CREW_CONFIG_PATH $CREW_CONFIG_PATH/meta $CREW_BREW_DIR $CREW_DEST_DIR $CREW_PACKAGES_PATH; do
-  sudo mkdir -p $dir && sudo chown -R $USER:$USER $dir
+  mkdir -p $dir
 done
 
 #prepare url and sha256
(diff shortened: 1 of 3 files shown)


uberhacker:
@thedamian: I reopened this PR since I had trouble updating the previous #546.  I updated the crew version, so we will also need to tag a new [release](https://github.com/skycocker/chromebrew/releases).  We probably need a new release anyway since there have been so many PRs merged into master recently. :)

install.sh
```diff
@@ -96,7 +99,7 @@ function extract_install () {
     rm -rf ./usr
     tar -xf $2
     echo "Installing $1 (this may take a while)..."
-    sudo tar cf - ./usr/* | (cd /; sudo tar xp --keep-directory-symlink -f -)
+    tar cf - ./usr/*
```
uberhacker: Do we need this?  From what I can tell, it serves no purpose and will never execute anyway.

uberhacker:
@jam7:  It took me awhile to understand what was going on here but now I get it.  I thought initially we could just simply extract the files but the original logic was trying to preserve the symlinks.  I have pushed the change you suggested.  Please test again when you get a chance.
Thanks

jam7:
It looks like working great here on armv7l.  #657 was needed to apply, though.  Nice work.

uberhacker:
@lyxell: We'll need to tag a new [release](https://github.com/skycocker/chromebrew/releases) since the crew version changed, right?

PR #650, opened 2017-05-12
Change perl.rb to Add source compile method and check method

jam7:
This is what I used to use to compile perl, so I'm not going to increment version number or remove binary packages here.  I hope this will be helpful in the future to compile next version of perl.  :)

I've tried to strip files also, but it changes only few ten KB.  So, I didn't add strip stuff this time.

Tested on armv7l.  It passes almost all `make check`. :)


jam7:
I was wrong.  If we compile perl with shared library, libperl.so saves space for other programs like vim which link libperl.

PR #658, opened 2017-05-14
Update make.rb

nbougues:
The configure script needs limits.h in the preprocessor test, which is part of linuxheaders


lyxell:
Fixed in #657

PR #662, opened 2017-05-18
Change all toolcain to use stripped binaries and libraries

jam7:
All toolchaiins are compiled by using https://github.com/jam7/chrome-cross as previous one was.  This time, I stripped all unnecessary symbols from binaries and libraries as suggested by @lyxell in https://github.com/jam7/chrome-cross/issues/2.

Reducing file size introduces more disk space, more performance, less battery usage.  :)

Tested on armv7 and x86_64.

### DIFF
--- a/packages/binutils.rb
+++ b/packages/binutils.rb
@@ -1,17 +1,17 @@
 require 'package'
 
 class Binutils < Package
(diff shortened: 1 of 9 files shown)


lyxell:
Great work!

PR #663, opened 2017-05-18
Add /usr/lib64 to VPATH to make traceroute compilable on x86_64.

jam7:
This is small patch to make traceroute compilable on x86_64.

Without this patch, `crew isntall traceroute` on x86_64 causes following error:
```
make[1]: *** No rule to make target '-lm', needed by 'traceroute'.  Stop.
```
With this patch, `crew install traceroute` works fine on x86_64.


uberhacker:
@jam7: Thank you for this fix.  I experienced the same issue and was working on a patch but you beat me to the punch. :)

PR #675, opened 2017-05-22
Add wine package

flyingP0tat0:


finwo:
Has this been tested on platforms other than amd64?

flyingP0tat0:
No, it has not. I only have a x86_64 chromebook.

lyxell:
The `--enable-win64` will likely break installation on other architectures, I'll wait with merging this until it's tested on arm/aarch64.

flyingP0tat0:
Yes, I already thought about that. Should I make two sperate packages? (something like wine and wine64)

flyingP0tat0:
@lyxell Yes, I already thought about that. Should I make two sperate packages? (something like wine and wine64)

lyxell:
If the --enable-win64 flag really is needed for x86_64, a case switch inside wine.rb determining what to do for the specific architectures would be needed.

flyingP0tat0:
I deleted my wine branch, will create a new pull request.

PR #676, opened 2017-05-23
Add help command

uberhacker:
- Provide information about usage for each command
  - Update README.md to add commands and correct grammar

### DIFF
--- a/crew
+++ b/crew
@@ -70,7 +70,7 @@ def search (pkgName, silent = false)
 end
 
 def help (pkgName)
-  case @pkgName
+  case pkgName
   when "build"
     puts "Build a package."
     puts "Usage: crew build [package]"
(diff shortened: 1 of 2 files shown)


crew
```diff
@@ -69,6 +69,49 @@ def search (pkgName, silent = false)
   abort "package #{pkgName} not found :("
 end
 
+def help (pkgName)
+  case @pkgName
```
jam7: you can write `case pkgName` here.  @pkgName is an instance variable.  And pkgName is a local variable.

crew
```diff
@@ -531,14 +581,14 @@ when "whatprovides"
   if @pkgName
     whatprovides @pkgName
   else
-    puts "Usage: crew whatprovides [pattern]"
+    help @pkgName = "whatprovides"
```
jam7: Once you use a local variable in help function, you can pass it here like `help "whatprovides"`, here and below.

uberhacker:
@jam7: Thank you for the pointers.  I have updated as you suggested.

Issue #668, opened 2017-05-20
"No F77 compiler found" when installing r


wangguansong:
When installing r, it exited with a configure error "No F77 Compiler found".
I thought gcc covers f77. Anyway, is there an available package for this in crew?

jam7:
#662 adds gfortran also.  I could compile r with it.  Please apply #662 by hand or wait until the merge.

jam7:
#662 is merged.  so, I think you can compile R now.  Please close this issue if it works.  Thanks.

wangguansong:
I tried again, it passed f77 checking, while stopped at zlib:

`checking if zlib version >= 1.2.5... no
checking whether zlib support suffices... configure: error: zlib library and headers are required
r failed to install: `./configure --with-x=no` exited with 1`

jam7:
you can install zlib by typing `crew install zlibpkg`.  you will need `pcre` package also.

wangguansong:
I added zlibpkg, which is already installed, and pcre. Then I tried again and got the same error.

jam7:
Ya.  R-3.3.2 has a problem in the way to recognize zlib version.  It is fixed in R-3.4.0.  So, apply #683 first please.

wangguansong:
Success! Thanks!

PR #683, opened 2017-05-26
Change r.rb to update it to 3.4.0 and correct dependencies to solve #668

jam7:
Also change to strip binaries.

Tested on armv7l and x86_64.  R can be compiled on both without errors.  However, R on armv7l doesn't pass all tests, so it may have problems.  R on x86_64 passes all tests.

### DIFF
--- a/packages/r.rb
+++ b/packages/r.rb
@@ -1,9 +1,14 @@
 require 'package'
 
 class R < Package
-  version '3.3.2'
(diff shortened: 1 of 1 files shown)


BobbyBobster:
While the issues mentioned in #668 have been resolved, I still run into an error on my Toshiba Chromebook 2 (x86_64). The problem occurs when checking for the bzlib. 
I have uploaded a screenshot of the error message.
chromebrewerrormessager

PR #685, opened 2017-05-27
Add yarn

flyingP0tat0:
Automatically chown's the npm directories (for `-g` installs). I think we should consider doing this aswell in the node package rather than telling the user to do so.

### DIFF
--- a/packages/yarn.rb
+++ b/packages/yarn.rb
@@ -1,0 +1,17 @@
+require 'package'
+
+class Yarn < Package
+  version '0.25.3'
+  source_url 'https://github.com/yarnpkg/yarn/releases/download/v0.25.3/yarn-v0.25.3.tar.gz'
+  source_sha1 'a25d2eba9d40479a795422a23af8db9e950f36d0'
+
+  depends_on 'node'
+
+  def self.build
+    system "sudo bash chown -R $(whoami) $(npm config get prefix)/{lib/node_modules,bin,share}"
+  end
+
+  def self.install
+    system "npm install -g"
+  end
+end
--- a/packages/yarn.rb
+++ b/packages/yarn.rb
@@ -2,7 +2,7 @@ require 'package'
 
 class Yarn < Package
   version '0.25.4'
-  source_url 'https://github.com/yarnpkg/yarn/releases/download/v0.25.3/yarn-v0.25.3.tar.gz'
+  source_url 'https://github.com/yarnpkg/yarn/releases/download/v0.25.4/yarn-v0.25.4.tar.gz'
   source_sha1 '94b08478cf06652a337aef5742cf32472c767924'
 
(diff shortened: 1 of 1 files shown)


jam7:
Do we need to perform chown each time?  If not, it's a better to modify `node.rb` to perform chown only once, I think.  We are going to reduce the number of `sudo` as discussed in #176, #546 and #645.  Thanks.

lyxell:
Does it really make sense to bundle npm packages with chromebrew? Shouldn't `npm install yarn -g` suffice?

flyingP0tat0:
@jam7 Yes, it would make more sense. But I wasn't sure about that, thats why I waited for somebody to share some thoughts about that. I will make that change later today.

flyingP0tat0:
@lyxell Via npm is not the recommended way of installing yarn. It will even tell you that if you do.

lyxell:
@flyingP0tat0 Wouldn't it make sense then to not use npm at all in the installation? Line 15 in this pull request currently installs Yarn via npm by `system "npm install -g"`.

This is essentially the same thing as just issuing `npm install yarn -g` outside of crew.

flyingP0tat0:
It is the only way to install yarn, as it is a node module.
My point was, that it is a node module, but should be treared as any other program, also update in that way. Nobody really does an global upgrade via npm, so we can just release a new version here, and everyone updates it with all the other programs. Thats also the reason why there are yarn packages for the major distributions and if you do a `npm install -g yarn`, yarn will tell you that it want's itself to be installed via package managers or alternative scripts.

lyxell:
Alright, one problem installing via npm is that it overrides the DESTDIR intermediate step, so crew remove yarn would do nothing. Would it be possible to just install it via cp?

flyingP0tat0:
@lyxell Yes that should be possible, it will just do a `npm install` and then `cp` the files to their locations. I will take a look at that.

thedamian:
I don't know if we need this. yOu can install this with npm  by simply running: 
npm install --global yarn 
(As per the installation instructions on their own website: https://yarnpkg.com/en/docs/install#alternatives-tab )

PLUS following their directions the latest (maybe official?) version is; yarn-v0.24.6.tar.gz

You're installing a newer one than that. 

Does this script even work (didn't have a chance to test it) 

Their basic instructions (see below) don't seem to require npm and if you're creating a pcakge for it don't make it depend on node JUST to use npm

(from their own website)
cd /usr/opt
wget https://yarnpkg.com/latest.tar.gz
tar zvxf latest.tar.gz

flyingP0tat0:
Well as I said before, it is not recommended to install yarn via npm. From their website: 

>  Installation of Yarn via npm is generally not recommended

But what I am doing at the moment is really nothing more than a npm install, so I will drop this one. Thanks for your feedback!

PR #688, opened 2017-05-28
Fix openssl source url

uberhacker:
Apparently, the ftp site is no longer in service.


uberhacker:
Not needed with PR #690.

Issue #665, opened 2017-05-19
Installing openSSL breaks ssh/wget


bdodroid:
EDIT: When I try and install anything requiring OpenSSL on my Samsung Chromebook Plus it fails and breaks wget/ssh until I remove openssl.

Here is the last few lines of the install:
```
Installing...
Groff installed!
Found curl, version 7.54.0
No precompiled binary available for your platform, downloading source...
--2017-05-18 19:31:47--  https://curl.haxx.se/download/curl-7.54.0.tar.bz2
Resolving curl.haxx.se... 151.101.2.49, 151.101.66.49, 151.101.130.49, ...
Connecting to curl.haxx.se|151.101.2.49|:443... connected.
*** Error in `wget': free(): invalid pointer: 0xb9059b78 ***
Checksum mismatch :/ try again
```

Trying any wget after this fails with *** Error in `wget': free(): invalid pointer: 0xb2c1f0e0 ***

Im running - Version 60.0.3100.0 (Official Build) dev (32-bit) (same results installing php on version 58.0.3029.112)

*edit was to change "Crew" to "anything requiring OpenSSL" as Crew itself doesn't cause any problems. **Also removed Crew from the title.

ksunden:
I had this issue as well, turned out to just be my disk was full, so wget ran out of space

bdodroid:
I was wondering if that was it but I have 20gb free space :(
Turns out it also breaks ssh for me as well, with an invalid pointer error. 
So weird, I've never had any problems on my laptop (neverware) or my last Asus chromebook flip, but for some reason my chromebook plus just wont do it.

Anyone have this working on a Samsung Chromebook Plus?

ksunden:
try "sudo crew remove openssl"

For me, python had that as a dependency, but installing had the same effect (aarch64 acer R13)
Removing (and even installing python without openssl in the first place), made it work

lyxell:
Related to #614

bdodroid:
@ksunden you where right, openssl is the problem and removing fixes my invalid pointer issue.
Is it possible to install php5/7 without openssl? 
I tried removing references to it in the php.rb package file but it still ends up installing it.

Ulexus:
I have an Acer CB5-312T, arm64/armv8.  Same problem.  If the chromebrew openssl package is installed, it breaks any tool which uses openssl (wget, ssh, etc).

```
$ uname -a
[...]

PR #690, opened 2017-05-28
Update to 1.0.2l / V 1.0.2k FTP Link offline

beli3ver:


Ulexus:
For me, this breaks the system `wget`'s TLS support, which, among other things, causes `crew` to fail to download.

jam7:
It works fine on armv7l and x86_64.  Passed all test cases also.

beli3ver:
> For me, this breaks the system wget's TLS support, which, among other things, causes crew to fail to download https-sourced files.

On my system, I have no problems.

Ulexus:
I'm on arm64/armv8.  Perhaps that is the distinction, though I don't see why architecture should make a difference.

flyingP0tat0:
I just saw this one after creating #709...
Why does this one have conflicts, while mine doesn't?

flyingP0tat0:
@Ulexus Could you try it with 1.0.2k manually? Although I don't see any reason why this should cause a broken wget.

Ulexus:
@flyingP0tat0 same problem with 1.0.2k manually.  I'm sure it's not related to the version bump per se, but rather that the openssl version itself is not compatible with whatever the system wget is compiled against.

Thus, there is no reason to hold this PR on my issue.  It looks like that deserves a separate issue.

flyingP0tat0:
@Ulexus Yes, open a new issue with your problem.

flyingP0tat0:
I think we should fix this problem with the openssl URL as fast as possible, as many packages depend on openssl. I already resloved the conflict in #709.

flyingP0tat0:
@jam7's package with binaries for arm (#665) seems to be working, so we might just drop this one and he needs to push his package (with new version and via `http`).

jam7:
@flyingP0tat0 I make #714.  Please let me know there is any problems.

PR #701, opened 2017-06-01
Change to use tar for the performance

jam7:
Modify crew to use `"tar cf - | tar xf -"` style to reduce the installation time, twice or several times faster.

This mechanism is used in install.sh and I was testing this for several months on armv7l and x86_64.  Please check it out. :)


lyxell:
Really nice!

Issue #707, opened 2017-06-03
Fatal error when installing R


BobbyBobster:
Tried to install R on my Toshiba Chromebook 2 (x86_64) but got a fatal error. I first ran into some other minor problems, namely I had to install `bz2`, `curl` and `openssl` (after applying #690) first.

Then I got this error message:

![errormessager](https://cloud.githubusercontent.com/assets/23128298/26757176/6e42302a-48b4-11e7-9932-8747bbe72b8d.png)

BobbyBobster:
Turns out I also had to install `readline`. Now it has installed without error. 
I don't know if this means that the dependencies of `r.rb` should be changed, but I am closing the issue.

flyingP0tat0:
Well it should definetly be added as a dependency.

PR #708, opened 2017-06-03
Add readline dependency to R

flyingP0tat0:
As seen in #707.


jam7:
I guess adding `bz2`, `curl` and `openssl` is also needed as he said.

flyingP0tat0:
Yes, have not seen that.

PR #709, opened 2017-06-05
IMPORTANT: Update openssl

flyingP0tat0:
The openssl package is not working due to some error when downloading the package. Maybe `ftp` does not work anymore. Anyway, here is the (newer) `http` version.


flyingP0tat0:
This is not neccessary anymore.

PR #710, opened 2017-06-05
Update NodeJS to auto chown

flyingP0tat0:
This will automatically chown the npm directories.


thedamian:
Yeah, I see what you're doing but just copying the instructions did work initially, have you tried your changes (of just putting the line I made into a comment really run as a command?)

Issue #614, opened 2017-05-01
python27 install fails because openssl won't build


brucem1976:
On Samsung Chromebook Plus:

Configured for linux-aarch64.
making all in crypto...
make[1]: Entering directory `/usr/local/tmp/crew/openssl-1.0.2k.tar.gz.dir/openssl-1.0.2k/crypto'
make[1]: warning: jobserver unavailable: using -j1.  Add `+' to parent make rule.
/usr/local/bin/perl ../util/mkbuildinf.pl "gcc -I. -I.. -I../include  -fPIC -DOPENSSL_PIC -DZLIB_SHARED -DZLIB -DOPENSSL_THREADS -D_REENTRANT -DDSO_DLFCN -DHAVE_DLFCN_H -Wa,--noexecstack -O3 -Wall -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM" "linux-aarch64" >buildinf.h
gcc -I. -I.. -I../include  -fPIC -DOPENSSL_PIC -DZLIB_SHARED -DZLIB -DOPENSSL_THREADS -D_REENTRANT -DDSO_DLFCN -DHAVE_DLFCN_H -Wa,--noexecstack -O3 -Wall -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM   -c -o cryptlib.o cryptlib.c
gcc -I. -I.. -I../include  -fPIC -DOPENSSL_PIC -DZLIB_SHARED -DZLIB -DOPENSSL_THREADS -D_REENTRANT -DDSO_DLFCN -DHAVE_DLFCN_H -Wa,--noexecstack -O3 -Wall -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM   -c -o mem.o mem.c
gcc -I. -I.. -I../include  -fPIC -DOPENSSL_PIC -DZLIB_SHARED -DZLIB -DOPENSSL_THREADS -D_REENTRANT -DDSO_DLFCN -DHAVE_DLFCN_H -Wa,--noexecstack -O3 -Wall -DSHA1_ASM -DSHA256_ASM -DSHA512_ASM   -c -o mem_dbg.o mem_dbg.c
[...]

Issue #577, opened 2017-04-15
openSSL install fails on aarch64


ksunden:
Workaround: add no-asm to the ./config options

brucem1976:
Thanks that worked - but now I get:

No precompiled binary available for your platform, downloading source...
--2017-05-02 06:42:02--  https://www.python.org/ftp/python/2.7.13/Python-2.7.13.tar.xz
Resolving www.python.org... 151.101.172.223
Connecting to www.python.org|151.101.172.223|:443... connected.
*** Error in `wget': free(): invalid pointer: 0xb9112888 ***
Checksum mismatch :/ try again

lyxell:
This was probably solved by #591

brucem1976:
So this got it compiling and installing - however it messes up the crypto setup on the system - github stops authenticating, ssh-keygen won't run without sudo, and ssh-add doesn't work....

jam7:
How about performing test?

```
$ sudo crew remove openssl
$ crew install openssl keep
$ cd /usr/local/tmp/crew/openssl...    # cd to working directory crew left as is
$ make test
```

ksunden:
Why is openssl a required dependency for python2.7?
I am able to install python2.7 without the openssl dependency. (Python 2.7 was what prompted me to open this issue in the first place, as in #614 )

All problems go away without openssl installed.

brucem1976:
How do you install without openssl?

lyxell:
@ksunden OpenSSL gives Python features which is considered to be a part of the standard installation

jam7:
Please refer https://github.com/skycocker/chromebrew/issues/614#issuecomment-305946120 if you can try new toolchains for aarch64

jam7:
Close in favor of #1025.

PR #714, opened 2017-06-06
Update openssl to 1.0.2l and distribute with pre-built arm binaries

jam7:
This avoids the aarch64 openssl problem.  This doesn't fix it since we are not sure the source of problem.  At least we know armv7l binary works fine on aarch64, so I'm distributing it.  Also update the version of openssl.

I'm using official github distribution instead of official http distribution, since later removes old version at every release.

Tested on armv7l and x86_64.  Also tested on aarch64 as described in https://github.com/skycocker/chromebrew/issues/665#issuecomment-306335119.

This will fix problems in #665, #614, #577.


lyxell:
Thanks Jam!

PR #717, opened 2017-06-08
Refactor dependency expansion

jam7:
Change crew to remove already installed packages from dependencies, so it won't show something like following messages if toolchains are already installed.  :)
```
$ crew install xzutils
Found xzutils, version 5.2.3-2
Following packages also need to be installed: 
glibc223 binutils gmp mpfr mpc isl cloog glibc gcc linuxheaders make pkgconfig 
Do you agree? [Y/n]
```
Also modified following stuff.
 - Change to check only run-time dependencies if a target is already installed
 - Change to remove unnecessary push_dependency call

Tested on armv7l and x86_64.

### DIFF
--- a/crew
+++ b/crew
@@ -336,7 +336,8 @@ def resolveDependencies
   @source_package = 0
 
   def push_dependencies
-    if @pkg.is_binary?(@device[:architecture])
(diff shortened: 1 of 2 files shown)


lyxell:
Great PR!

PR #719, opened 2017-06-10
Add info and refresh commands

uberhacker:
- The crew info command describes a package
  - The crew refresh command rebuilds the package metadata
  - Update README.md to include the new commands

This PR offers a complete solution to #332.  The package metadata is updated with the `crew update` command as well.


jam7:
Is doing something like below not enough?  I'm not sure why we need to build json database.
```
when "info"
  search @pkgName, true
  puts @pkg.description
  puts @pkg.homepage
```

lyxell:
I agree with @jam7 here

uberhacker:
Agreed but after thinking about it a bit, why don't we just enhance the `crew search` command?

uberhacker:
Replaced by #720.

PR #730, opened 2017-06-12
removing the @@debug_symbol line that was preventing updating

thedamian:


lyxell:
It would be great if you let people review your pull requests before merging.

PR #733, opened 2017-06-13
JDK to u131 -  Node to 6.11.0 - Node_Current to 8.1.0 - Inetutils Package (for telnet, BBS TIME!) added

thedamian:
Just updating Java, and Node to all the current versions.

NOTE: In Java I was able to find the x86 and 64x official links from oracle (instead of dropbox) but I couldn't find the ones for arm, so actually arm is back at 8u22 until I can find the links.

### DIFF
--- a/packages/inetutils.rb
+++ b/packages/inetutils.rb
@@ -1,20 +1,29 @@
 require 'package'
 
 class Inetutils < Package
-  description 'Inetutils - GNU network utilities is a collection of common network programs.'
+  description 'The Inetutils package contains programs for basic networking. Such as dnsdomainname, ftp, hostname, ifconfig, ping, 
+ping6, talk, telnet, tftp, traceroute'
   homepage 'https://www.gnu.org/software/inetutils/'
   version '1.9.4'
-  source_url 'http://ftp.gnu.org/gnu/inetutils/inetutils-1.9.4.tar.gz'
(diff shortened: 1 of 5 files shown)


thedamian:
I will change node and node_current to pull from the binaries. 

I think I was trying to make it easy for me by having everyone compile from code but BOY does my Google Pixel get hot when THAT'S happening, and it probably takes a good half an hour on my i5 1.8Ghz CPU so I can't even think how long it takes on an arm or x86 machine.

lyxell:
The -m64 flag looks problematic to me

thedamian:
@lyxell yeah. that's why I didn't automatically merge it. I needed it for my Pixel (with intel i5) so I'm guessing it's a 64bit version.

So, do I do one of these so that I can run -m64 on 64bit lines?

case ARCH
    when "x86_64" || "aarch64"

lyxell:
I don't think it is possible to use -m64 on aarch64. According to GCC documentation "The 64-bit environment sets int to 32 bits and long and pointer to 64 bits and generates code for AMD's x86-64 architecture". Also, aarch64 uses armv7-gcc, which is 32 bit.

thedamian:
when I tried to upgrade node on my box using crew the -m64 was the only thing that allow it and didn't give me the errors with some libraries. 

My box is the intel i5 (google pixel) so maybe only x64 intel needs it?

I'd have to try it without it to see what the error was and report it.

PR #403, opened 2017-01-26
Refactor search logic to support regex

cstrouse:
`crew search` use to only return a single result if your query string
matched a package name exactly.

`crew search` now supports regex.

```bash
chronos@localhost / $ crew search ^lib
Found libpcap, version 1.8.1
Found libjpeg, version 9.1
Found libunwind, version 1.1
Found libxslt, version 1.1.28
Found libffi, version 3.0.13-1
Found libssh2, version 1.4.3
Found libtool, version 2.4.6
Found libpipeline, version 1.4.1
Found libevent, version 2.0.22
Found libxml2, version 2.9.2
Found libgd, version 2.0.33
Found libtiff, version 4.0.7
Found libpng, version 1.6.26
Found libsigsegv, version 2.10
```

Also includes some minor cleanup to make things more idiomatic.

PR #735, opened 2017-06-13
Refactor search logic to support regex

jam7:
This is revised version of #403.

I modified @cstrouse PR to match the latest upstream/master and change to add new `regexp_search` function instead of modifying existing `search` function.  I also add `extra` option.  So, it is possible to use this like below now.
```
$ crew search vi
(i) vim: Vim is a highly configurable text editor built to make creating and changing any kind of text very efficient.
(i) vidstab: Transcode video stabilization plugin.
vifm: Vifm is an ncurses based file manager with vi like keybindings/modes/options/commands/configuration, which also borrows some useful ideas from mutt.
(i) libxvid: The free video codec that is strong in compression and quality.
neovim: Neovim is a refactor, and sometimes redactor, in the tradition of Vim (which itself derives from Stevie).
$ crew search ^vim$ extra
(i) vim: Vim is a highly configurable text editor built to make creating and changing any kind of text very efficient.
http://www.vim.org/
version 8.0-1
```

### DIFF
--- a/crew
+++ b/crew
@@ -157,7 +157,7 @@ def regexp_search(pkg_name)
   results = Dir["#{CREW_LIB_PATH}packages/*.rb"] \
 	  .select  { |f| File.basename(f, '.rb') =~ Regexp.new(pkg_name, true) } \
 	  .collect { |f| File.basename(f, '.rb') } \
-	  .each    { |f| set_package(f) }
+	  .each    { |f| print_package(f, ARGV[2] == "extra") }
   abort "Package not found :(" unless results.length > 0
 end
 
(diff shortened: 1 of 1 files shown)


cstrouse:
Looks good to me @jam7.

uberhacker:
This completely overrides the print_package call and will never print the description or homepage in the results of a `crew search <pattern>`.  It also won't print descriptions in the output of `crew search` without any parameters.  More work needed. :)

jam7:
that's good features to add, but I prefer to merge this PR as is since it won't break crew.  then, you can raise new PR with such improvements.

uberhacker:
Fair enough.  Merge away.  Good job btw.

uberhacker:
@jam7: I noticed we're going to have a merge conflict with PR #736.  I can resubmit after you merge this PR or vice versa.

PR #739, opened 2017-06-15
Add composer package

uberhacker:
Composer is a tool for dependency management in PHP. It allows you to declare the libraries your project depends on and it will manage (install/update) them for you.  See https://getcomposer.org/.

### DIFF
--- a/packages/composer.rb
+++ b/packages/composer.rb
@@ -1,0 +1,19 @@
+require 'package'
+
+class Composer < Package
+  description 'Dependency Manager for PHP'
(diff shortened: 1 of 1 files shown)


thedamian:
composer is installable by php, but you're right. it's much easier in a package. I love it. MERGED

lyxell:
This looks broken to me, manually copying files via `cp` tends to break uninstallation.

uberhacker:
@lyxell: Uninstall worked perfectly when I tested.  Can you provide any errors you experienced while testing?

PR #741, opened 2017-06-15
Fix depends_on issue as pointed out by @lyxell

uberhacker:
Don't require php unless it isn't currently installed.  Reference #739.

### DIFF
--- a/packages/composer.rb
+++ b/packages/composer.rb
@@ -1,5 +1,4 @@
 require 'package'
-require 'fileutils'
 
 class Composer < Package
   description 'Dependency Manager for PHP'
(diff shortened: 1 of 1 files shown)


thedamian:
Wait WHAT?

1) Why would composer NOT require PHP (it makes no sense to be without PHP)
2) you're removed the "PHP5 or PHP7" and think requiring EITHER is important.

lyxell:
@thedamian It works, but it is not super pretty.
The code could be read as:
```
if (php (any version) is not installed)
  require php7
```

uberhacker:
Hey @thedamian: What @lyxell mentioned is correct.  I suppose we could also do something like `depends_on 'php7' not File.exists? '/usr/local/bin/php'`.

lyxell:
The ruby idiom would probably be to use `unless`.

uberhacker:
You mean like this? `depends_on 'php7' unless File.exists? '/usr/local/bin/php'`

thedamian:
sorry guys. so "depends_on 'php7' is the same as check for ANY VERSION?

I definitely am NOT a ruby expert but this doesn't seem like the 7 would be ignored.

Just confirm. 

when you say "php7" the system is smart enough to ignore the 7 part. yes?

uberhacker:
Hey @thedamian: What this line will do, basically, is _*ONLY*_ install php7 if there is no version of php installed currently.  Otherwise, if you have either the php5 or php7 packages installed, this line is skipped.

thedamian:
@uberhacker  sorry. I guess it was the revert thinking of python that confused me.

I see: 1) It checks if ANY php is installed by checking for the directory. if it's NOT there, then it tries installs PHP7

NOW I get it. It's merged.

Issue #763, opened 2017-06-19
Unable to upgrade jdk8


uberhacker:
```
$ crew upgrade
Updating packages...
(i) jdk8: The JDK is a development environment for building applications, applets, and components using the Java programming language.
http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html
version 8u131
Precompiled binary available, downloading...
--2017-06-19 08:39:23--  http://javadl.oracle.com/webapps/download/AutoDL?BundleId=220305_d54c1d3a095b4ff2b6607d096fa80163
Resolving javadl.oracle.com... 137.254.120.23
Connecting to javadl.oracle.com|137.254.120.23|:80... connected.
HTTP request sent, awaiting response... 302 Moved Temporarily
Location: http://sdlc-esd.oracle.com/ESD6/JSCDL/jdk/8u131-b11/d54c1d3a095b4ff2b6607d096fa80163/jre-8u131-linux-x64.tar.gz?GroupName=JSC&FilePath=/ESD6/JSCDL/jdk/8u131-b11/d54c1d3a095b4ff2b6607d096fa80163/jre-8u131-linux-x64.tar.gz&BHost=javadl.sun.com&File=jre-8u131-linux-x64.tar.gz&AuthParam=1497880764_d6b1fb882a621bd0c820508630c132f9&ext=.gz [following]
--2017-06-19 08:39:24--  http://sdlc-esd.oracle.com/ESD6/JSCDL/jdk/8u131-b11/d54c1d3a095b4ff2b6607d096fa80163/jre-8u131-linux-x64.tar.gz?GroupName=JSC&FilePath=/ESD6/JSCDL/jdk/8u131-b11/d54c1d3a095b4ff2b6607d096fa80163/jre-8u131-linux-x64.tar.gz&BHost=javadl.sun.com&File=jre-8u131-linux-x64.tar.gz&AuthParam=1497880764_d6b1fb882a621bd0c820508630c132f9&ext=.gz
Resolving sdlc-esd.oracle.com... 23.218.156.10, 23.218.156.40
Connecting to sdlc-esd.oracle.com|23.218.156.10|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 75920010 (72M) [application/x-gzip]
Saving to: 'AutoDL'

AutoDL                                               100%[====================================================================================================================>]  72.40M  3.96MB/s    in 23s     

[...]

PR #764, opened 2017-06-19
Update jdk8 package

uberhacker:
Fix for issue #763.

### DIFF
--- a/packages/jdk8.rb
+++ b/packages/jdk8.rb
@@ -4,16 +4,26 @@ class Jdk8 < Package
   description 'The JDK is a development environment for building applications, applets, and components using the Java programming language.'
   homepage 'http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html'
   version '8u131'
-  binary_url ({
-    i686: "http://javadl.oracle.com/webapps/download/AutoDL?BundleId=220303_d54c1d3a095b4ff2b6607d096fa80163",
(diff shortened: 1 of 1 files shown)


thedamian:
Again, or either my python is not up to snuff (possibly)

or you're just making a directory called /usr/local/usr/local and copying all the JDK8 stuff up there in this line:
    system "mkdir -p #{CREW_DEST_DIR}/usr/local"

am I wrong?

uberhacker:
My understanding is crew will copy everything in `#{CREW_DEST_DIR}/usr/local` to the actual `/usr/local` directory during install.  This is what I needed to do to get it to actually install on my system.  Did you experience the same issue I did as explained in issue #763?

thedamian:
I was just under the impression that #{CREW_DEST_DIR} represented "/usr/local"

If I'm write then you'll create a diretory called:
/usr/local/usr/local

and install java there.

there but there they won't be able to be called.

That's all I'm saying. I'm ok if my new JDK doesn't work. and I REALLY appriciate the help fixing it. BUT I don't think your code will put the files in the correct place unless I don't understand python correctly (which Very well may be true)

uberhacker:
Actually, if you look at the defined constants at the top of the `crew` source, `CREW_DEST_DIR` is `/usr/local/tmp/crew/dest`.  When you install a package, all files that are extracted here will be tracked in `<package>.directorylist` and `<package>.filelist` in `/usr/local/etc/crew/meta` so they can be uninstalled cleanly.  Did you test this PR yet to see if it installs without error?

thedamian:
I did not test this PR. 
I'll test it when I get home. 

Thank you for the fix. I appreciate it.

PR #776, opened 2017-06-20
Update nano from 2.7.1 to 2.8.4

cstrouse:
This is a general maintenance and bugfix release which also includes several
enhancements to usability.


thedamian:
tested. Flawless update!

PR #777, opened 2017-06-21
Add pip package

uberhacker:
The pip executable is available in the `python27` and `python3` packages, however, the executable is not installed in `/usr/local/bin` which makes it difficult to actually use out-of-the-box.  Also, the `python` package does not have pip.  I felt this package was necessary because of all these issues.


uberhacker:
Not needed since `pip` is installed via `python27` or `python3` packages.

PR #778, opened 2017-06-21
Add storm package

uberhacker:
storm is a command line tool to manage your ssh connections.  See http://stormssh.readthedocs.io/en/master/.

features

- adding, editing, deleting, listing, searching across your SSHConfig.
- command alias support for your CLI preferences.
- support for custom SSH directives.
- scriptable as a python library.
- user interfaces besides cli. (wxpython, unity(ubuntu) indicator.)

This package installs the storm executable in /usr/local/bin so it's accessible anywhere in the terminal.

### DIFF
--- a/packages/storm.rb
+++ b/packages/storm.rb
@@ -1,0 +1,17 @@
+require 'package'
+
+class Storm < Package
+  description 'Manage your SSH like a boss.'
+  homepage 'http://stormssh.readthedocs.org/en/master/'
+  version '0.6.9'
+  source_url 'https://github.com/emre/storm/archive/0.6.9.tar.gz'
+  source_sha1 'ebc82a5ab77039de16e3ca0d7df6e24ff1950927'
+
+  depends_on 'pip'
+
+  def self.install
+    system "mkdir -p #{CREW_DEST_DIR}/usr/local/bin"
+    system "pip install stormssh --prefix=/usr/local"
+    system "cp /usr/local/bin/storm #{CREW_DEST_DIR}/usr/local/bin"
+  end
+end


lyxell:
Why not use CREW_DEST_DIR as prefix and let Chromebrew handle the copying?

uberhacker:
@lyxell: Because the installation method is `pip`?  I'm basically letting `crew` know after the fact that `storm.filelist` and `storm.directorylist` files exist by storing the executable in `CREW_DEST_DIR`.

lyxell:
It doesn't really make sense to me to wrap a package manager inside another package manager.

However, wouldn't something like (1):
```
  system "pip install stormssh --prefix=/usr/local --target=#{CREW_DEST_DIR}"
```
be preferred over (2):
```
  system "pip install stormssh --prefix=/usr/local"
  system "cp /usr/local/bin/storm #{CREW_DEST_DIR}/usr/local/bin"
```
If the user cancels the installation, I think problems would be handled correctly by 1 but not by 2. I haven't tested this though.

lyxell:
Also, is the storm binary really the only thing that "pip install stormssh" installs?

uberhacker:
@lyxell: What you suggested is definitely more elegant.  I'll give that a whirl and see how well it works.  AFAIK, the storm executable is the only file installed from this pip package.  I basically put this package out as a proof of concept of the possibility of using crew as a package manager wrapper.  You have to admit, it's kind of cool. :)

lyxell:
I'm a bit worried that the two package managers will collide, what happens if you uninstall via pip and then via chromebrew or vice versa?

uberhacker:
I'm just going to close this as a bad idea.  PoC rejected.

PR #779, opened 2017-06-22
Add gdrive package

uberhacker:
Google Drive CLI Client.  Drive starts you with 15 GB of free Google online storage, so you can keep photos, stories, designs, drawings, recordings, videos – anything.  See https://www.google.com/drive/.

### DIFF
--- a/packages/gdrive.rb
+++ b/packages/gdrive.rb
@@ -1,6 +1,8 @@
 require 'package'
 
 class Gdrive < Package
+  description 'Google Drive CLI Client'
+  homepage 'https://www.google.com/drive/'
   version '2.1.0'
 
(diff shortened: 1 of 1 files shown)


uberhacker:
Hey @jam7: Can you please try installing this package on your armv7l system when you get a chance?  It doesn't need to compile, so it should only take a minute.

lyxell:
This is uninstallable since it doesn't use CREW_DEST_DIR

uberhacker:
@lyxell: What makes you think it doesn't use `CREW_DEST_DIR`?  If you look in `/usr/local/etc/crew/meta` after installation, you will notice the `gdrive.directorylist` and `gdrive.filelist` files exist.  This will enable the package to be uninstalled cleanly also.

PR #781, opened 2017-06-22
Add aircrack-ng package and its dependencies

wwwiiilll:
Because why not.
Did some quick tests (on my Acer CB3-131) and it does seem to work well.


lyxell:
Nice work

PR #786, opened 2017-06-22
Add libsqlite3 package

uberhacker:
SQLite is a software library that implements a self-contained, serverless, zero-configuration, transactional SQL database engine.  See https://github.com/LuaDist/libsqlite3.


uberhacker:
@lyxell: Just so you know, this package is a dependency of aria2. :)

Issue #790, opened 2017-06-23
ruby failed to install: `make -j2` exited with 2


gishnanigans:
After trying to install `rsync` again after [issue 766](https://github.com/skycocker/chromebrew/issues/766) and getting the same error as before, I decided to do an `update` and `upgrade` on crew and its packages. The result of the upgrade was a lot of processing followed by a similar error and exit code, but this time for `ruby`.

Thought you would like to know.

>... many lines ...
compiling math.c
compiling node.c
compiling numeric.c
compiling object.c
compiling pack.c
compiling parse.c
compiling proc.c
compiling process.c
compiling random.c
In file included from /usr/local/include/sys/syscall.h:31:0,
                 from /usr/local/include/syscall.h:1,
                 from random.c:81:
random.c: In function ‘fill_random_bytes_syscall’:
random.c:525:16: error: ‘__NR_getrandom’ undeclared (first use in this function)
  ret = syscall(SYS_getrandom, seed, size, flags);
                ^
random.c:525:16: note: each undeclared identifier is reported only once for each function it appears in
random.c: At top level:
cc1: warning: unrecognized command line option "-Wno-self-assign" [enabled by default]
cc1: warning: unrecognized command line option "-Wno-constant-logical-operand" [enabled by default]
cc1: warning: unrecognized command line option "-Wno-parentheses-equality" [enabled by default]
cc1: warning: unrecognized command line option "-Wno-tautological-compare" [enabled by default]
make: *** [random.o] Error 1
make: *** Waiting for unfinished jobs....
ruby failed to install: `make -j2` exited with 2

thedamian:
It was a dependency issue, it looks like one of our packages (that you had installed) has this missing. 
I'll look to see what it was before. You might have to update all your /usr/local/lib/crew/packages/ directory manual from github.
and then the upgrade should work. if it doesn't please tell us which package this broke on.

gishnanigans:
I'll give it a try tonight or tomorrow ASAP. Thank you.

jam7:
@gishnanigans you may need to install linuxheaders.  `__NR_getrandom` is defined in `/usr/local/include/bits/syscall.h`.  Please try `crew upgrade linuxheaders` by hand at first.

I think the source of problem is that `crew upgrade` doesn't upgrade files belong to dependency.  So, you may need to install them by hand belong to dependency sometime.

gishnanigans:
[...]

PR #791, opened 2017-06-23
Correct dependency problems

jam7:
This solves #790 problem, but this makes `crew upgrade` little slow if you install a lot of packages.  :(

This also solves a problem that :build tagged packages are not installed automatically when `crew build target` if the target is already installed.

Tested on armv7l and x86_64.


thedamian:
I'm merging this, but I had a similar problem (but with me 'make -j4' ) and only with the php7 package. (again which may be different.

Issue #689, opened 2017-05-28
Dev-channel LD_LIBRARY_PATH does not include /usr/local/lib


dihuta:
I install the lasted chromebrew and install nano via command: crew install nano. 
Then run nano, it display this error:

`nano: error while loading shared libraries: libncursesw.so.6: cannot open shared object file: No such file or directory`

Please help me.
Thank you.

jam7:
What does `crew whatprovides libncursesw` show?  And what does `ls -l /usr/local/lib*/libncurses*` show?  And please let us know your machine's architecture also which is shown by `uname -m`.

dihuta:
>> crew whatprovides libncursesw
`chronos@localhost / $ crew whatprovides libncursesw
ncursesw: /usr/local/lib/libncursesw.a
ncursesw: /usr/local/lib/libncursesw.so
ncursesw: /usr/local/lib/libncursesw.so.6
ncursesw: /usr/local/lib/libncursesw.so.6.0

Total found: 4
chronos@localhost / $ uname
x86_64
ls -l /usr/local/lib*/libncurses*
lrwxrwxrwx. 1 chronos chronos     15 May 28 18:01 /usr/local/lib/libncurses.so -> libncurses.so.6
lrwxrwxrwx. 1 chronos chronos     17 May 28 18:01 /usr/local/lib/libncurses++.so -> libncurses++.so.6
lrwxrwxrwx. 1 chronos chronos     17 May 28 18:01 /usr/local/lib/libncurses.so.6 -> libncurses.so.6.0
lrwxrwxrwx. 1 chronos chronos     19 May 28 18:01 /usr/local/lib/libncurses++.so.6 -> libncurses++.so.6.0
-rwxr-xr-x. 1 chronos chronos 449792 May 28 18:01 /usr/local/lib/libncurses.so.6.0
-rwxr-xr-x. 1 chronos chronos 315672 May 28 18:01 /usr/local/lib/libncurses++.so.6.0
-rw-r--r--. 1 chronos chronos 689462 May 28 18:00 /usr/local/lib/libncurses++w.a
-rw-r--r--. 1 chronos chronos 870760 May 28 18:00 /usr/local/lib/libncursesw.a
lrwxrwxrwx. 1 chronos chronos     18 May 28 18:00 /usr/local/lib/libncurses++w.so -> libncurses++w.so.6
[...]

PR #794, opened 2017-06-24
Support lib64 installation on x86_64

jam7:
This PR modify `crew` to add `CREW_LIB_PREFIX` constant as described in #689.  This also modify several packages to demonstrate how to use it.  By applying this PR, some packanges start to install their libraries into `/usr/local/lib64` instead of `/usr/local/lib` on x86_64.

This PR follows the direction what the latest ChromeOS on x86_64 takes.  However, this PR won't break anything.  We still can choose use only lib64 or both lib64 and lib as we are discussing in #689.  This PR just shows the possibility that we can use only lib64 way as what ChromeOS is choosing.

Tested on armv7l and x86_64.

### DIFF
--- a/packages/curl.rb
+++ b/packages/curl.rb
@@ -13,7 +13,7 @@ class Curl < Package
   depends_on 'groff' => :build
 
   def self.build
-    system "./configure", "--disable-static"
+    system "./configure", "--lidir=#{CREW_LIB_PREFIX}", "--disable-static"
     system "make"
   end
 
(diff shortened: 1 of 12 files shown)


packages/curl.rb
```diff
@@ -13,15 +13,15 @@ class Curl < Package
   depends_on 'groff' => :build
 
   def self.build
-    system "./configure", "--disable-static"
+    system "./configure", "--lidir=#{CREW_LIB_PREFIX}", "--disable-static"
```
lyxell: There's a typo here at line 16.

jam7: Thank you for the review.  I've corrected it.

Issue #697, opened 2017-05-30
GCC internal compiler error 


tempdc:
Hi, when I try to compile qemu, I am getting an internal compiler error. Any hint on what to do to address?

Making all in test
  CC       utils.lo
  CC       utils-prng.lo
utils-prng.c: In function ‘prng_randmemset_r’:
utils-prng.c:272:6: internal compiler error: in emit_move_insn, at expr.c:3609
 void prng_randmemset_r (prng_t                  *prng,
      ^

utils-prng.c:272:6: internal compiler error: Aborted
gcc: internal compiler error: Aborted (program cc1)
../libtool: line 1749: 10309 Aborted                 (core dumped) gcc -DHAVE_CONFIG_H -I. -I.. -I../pixman -I../pixman -fopenmp -O2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -g -fPIC -m32 -Wall -Wdeclaration-after-statement -fno-strict-aliasing -fvisibility=hidden -pthread -O2 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -g -fPIC -m32 -Wall -Wdeclaration-after-statement -fno-strict-aliasing -fvisibility=hidden -MT utils-prng.lo -MD -MP -MF .deps/utils-prng.Tpo -c utils-prng.c -o utils-prng.o
make[3]: *** [Makefile:1044: utils-prng.lo] Error 1
make[3]: *** Waiting for unfinished jobs....
make[2]: *** [Makefile:503: all-recursive] Error 1
make[1]: *** [Makefile:410: all] Error 2
make: *** [Makefile:330: subdir-pixman] Error 2
qemu failed to install: `make -j2` exited with 2

lyxell:
Is this on Samsung Chromebook Plus?

tempdc:
I am using Cloudready 32 bit version 56 on a x686 platform. I used the standard ruby file that download qemu 2.9.0.  I have made some attempts with lower versions (2.7.1 and 2.8.1) with the same error.

flyingP0tat0:
Maybe try to install the pixman package (as the error seems to be connected to the compilation of pixman) and see what happens could be a good starting point.

tempdc:
Unfortunately, still get the same error compiling pixman...

make[2]: Entering directory '/usr/local/tmp/crew/pixman-0.34.0.tar.gz.dir/pixman-0.34.0/test'
  CC       utils.lo
[...]

PR #795, opened 2017-06-24
Change to use gcc-4.9.4.

jam7:
Change toolchains to use gcc-4.9.4 instead of google internal gcc-4.9.x.  This may solve the problem #697.  This time I also modify binutils to install stuff into `/usr/local/lib64` on x86_64 to not cause any problems like #689.  So, this toolchains work flawlessly on ChromeOS 60 x86_64 out of the box.

Tested on armv7l and x86_64.  Not tested on i686 nor aarch64, but I hope it works well since we used to use these cross-compiled toolchains on them.


lyxell:
Nice work!

PR #798, opened 2017-06-25
Fix Dropbox for ARM Devices

RMcNeely:
Recently submitted Dropbox package "breaks" `crew search` for the ARM devices by aborting before search is completed. Simple fix allows search to continue on ARM.

### DIFF
--- a/packages/dropbox.rb
+++ b/packages/dropbox.rb
@@ -12,7 +12,7 @@ class Dropbox < Package
     source_url 'https://clientupdates.dropboxstatic.com/dbx-releng/client/dropbox-lnx.x86_64-28.4.14.tar.gz'
     source_sha1 '40f4f37b64394d42f4fa3d6b3d53553f854587e4'
   else
-    abort 'Unable to install dropboxd.  Supported architectures include x86 and x86_64 only.'.lightred
+    'Unable to install dropboxd.  Supported architectures include x86 and x86_64 only.'.lightred
   end
 
   depends_on 'python'


thedamian:
Help me out. You've removed the "abort" method and now there's just a string with no command.
Is this valid python?

uberhacker:
@RMcNeely: Thank you for pointing out this issue.  I think we will probably want to use something like `puts` so the message is printed.  Do you get this message every time you run `crew search`?

jam7:
@RMcNeely Nice to point it out.  Thanks.
@thedamian yeah.  it's valid and ignored.  I prefere to remove `else` and the string itself, though.
@uberhacker `crew search` parses all package files.  For the case of dropbox, it abort parsing and crew itself on arm.

uberhacker:
@jam7: The rub here is during installation, you will want the install to abort, otherwise, it will continue and attempt to compile resulting in an epic failure.  What would happen if we put the condition inside the `def self.build` section?

jam7:
@uberhacker `crew install` aborts if it cannot find source files nor binary files.  It abort at line 354 of crew.  You can modify that error message as you like instead of writing abort in each package file.

uberhacker:
@jam7: If we remove the message, then we will get some kludgy error when an arm system user attempts to install the package, right?  What if we just replace `abort` with `puts`?  Would that work?

jam7:
@uberhacker No worry.  If you didn't write any source_url nor binary_url for arm, `crew` abort in the middle of compilation.  After applying this PR, crew works like below.
```
$ crew install dropbox
dropbox: Dropbox simplifies the way you create, share and collaborate. Bring your photos, docs, and videos anywhere and keep your files safe.
https://www.dropbox.com/
version 28.4.14
No precompiled binary for armv7l nor source is available.
```

uberhacker:
@jam7: Ok, that is great.  What if we just turn the abort line into a comment instead?

jam7:
@uberhacker that works fine also.  👍

uberhacker:
I am unable to reproduce this behavior on x86_64.  @RMcNeely: If what you say is true, the jdk8 package should affect `crew search` also.  Do you experience the same issue as with this package?

RMcNeely:
@EdReel this is the package I traced my issues to locally (I'm still
rocking a 1st Gen Samsung Chromebook) but I'll look at jdk8 also. You
shouldn't see the issue if you're on x86_64 for Dropbox BC that case is
accountd for.

On Jun 25, 2017 9:42 AM, "Ed Reel" <notifications@github.com> wrote:

I am unable to reproduce this behavior on x86_64. @RMcNeely
<https://github.com/rmcneely>: If what you say is true, the jdk8 package
should affect crew search also. Do you experience the same issue as with
this package?

—
<https://github.com/skycocker/chromebrew/pull/798#issuecomment-310903314>,
or mute the thread

RMcNeely:
Did some checking, the JDK8 package doesn't have the same effect b/c whomever wrote it accounted for cases of both `arm7l` and `aarch64`which might be a preferable design pattern to enforce.

uberhacker:
@RMcNeely: I was able to reproduce this error on x86_64.  I didn't copy the package with the commented lines to /usr/local/lib/crew/packages before testing. doh!  I think in your case, the other packages don't present a problem because your architecture is included in the conditionals for those packages.  I will follow @jam7's advice and just remove the abort lines in the other packages just in case.  The crew generated message is fine without the else condition in all packages.

uberhacker:
@RMcNeely: PR #830 resolves this issue so we can probably close this request.  I appreciate you bringing this issue to our attention.

PR #799, opened 2017-06-25
Change install to fix quoting problem.

jam7:
I've found quoting problem in install.sh.  With single quote, it doesn't expand `$1` correctly.  This PR fixes the problem.

Tested on armv7l.


uberhacker:
@jam7: Nice find.  Yeah, the `$1` will print a literal without the double quotes.

PR #800, opened 2017-06-25
Change README.md to modify armv7l support and add aarch64 support

jam7:
Modify README.md as described in the title.  :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,8 @@ Supported Systems
 |:--------------:|:----------:|
 |     x86_64     |     Yes    |
 |      i686      |     Yes    |
-| arm (Exynos 5) |     Yes    |
+|     armv7l     |     Yes    |
+|     aarch64    |     Yes(use armv7l binaries)    |
 
 Overview
 --------
--- a/README.md
+++ b/README.md
@@ -6,12 +6,12 @@ Package manager for Chrome OS
 Supported Systems
 -----------------
 
-|  Architecture  | Supported? |
-|:--------------:|:----------:|
-|     x86_64     |     Yes    |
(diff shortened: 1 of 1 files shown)


uberhacker:
Does this corrupt the spacing of the box?

jam7:
No.  Browser is adjusting the spacing.

README.md
```diff
@@ -10,7 +10,8 @@ Supported Systems
 |:--------------:|:----------:|
 |     x86_64     |     Yes    |
 |      i686      |     Yes    |
-| arm (Exynos 5) |     Yes    |
+|     armv7l     |     Yes    |
+|     aarch64    |     Yes(use armv7l binaries)    |
```
flyingP0tat0: I think a whitespace after `Yes` should make it more readable.

flyingP0tat0: Second: although it does not break the table, I guess it is considered good practise to fix the spacing in the Markdown.

uberhacker:
Do we even need the box?  Wouldn't a bulleted list work also?  The yes column is redundant anyway.

README.md
jam7: Thanks.  I modified it.

jam7:
@uberhacker Maybe yes, but I don't care it.  Please modify it once this PR is merged.

RMcNeely:
Does this mean that Chromebrew is expected to work on ARM Chromebooks even without Exynos chipsets?

jam7:
Originally, I ported chromebrew to c100pa which is RK3288 regarding to https://www.chromium.org/chromium-os/developer-information-for-chrome-os-devices.  Is it Exynos chipset?

PR #801, opened 2017-06-25
Add crew shrink command

uberhacker:
Introducing `crew shrink`.  This command will reduce the size of manpage and shared library object files and recover valuable disk space.  If the name reminds you too much of a Seinfeld episode, we can change it to something else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ Where available commands are:
   * install [install a package along with its dependencies after prompting for confirmation]
   * remove [remove a package]
   * search [look for a package]
+  * shrink [reduce the size of manpage and shared library object files]
   * update [update crew itself]
   * upgrade [update all or a specific package]
   * whatprovides [regex search for package(s) that contains file(s)]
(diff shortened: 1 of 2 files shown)


jam7:
Nice idea.  I was working to modify crew to strip binaries and libraries at install time, though. :)

BTW, compressing manual files is not enough.  You need to re-link all symbolic links referring the compressing files.

jam7:
If you have an interest, please check http://www.linuxfromscratch.org/blfs/view/5.1/postlfs/compressdoc.html to see how to handle symbolic and hard links on manual pages correctly.

uberhacker:
@jam7: Thanks for the input.  I was thinking it might help to avoid the need to put strip commands in every package. :)  But I understand if you want to control it on a package by package basis.  Maybe we can have a flag that ignores certain packages where this might create an issue?

uberhacker:
I like your idea better, actually.  Can we include the manpage compression in the crew command during install time also?  I will make a comment on your PR.

PR #802, opened 2017-06-25
Change crew to strip binaries and libraries at install time

jam7:
This PR modifies `crew` to strip binaries and libraries at install time unless `CREW_NOT_STRIP` environment variable is defined.

I modified several packages to do so, but it makes those files complicated and difficult to understand.  Therefore, I was looking for different method to strip them without modifying package files.  This PR is it.

I also reverted package files modified to strip binaries and libraries by themselves.

Tested on armv7l and x86_64.

### DIFF
--- a/crew
+++ b/crew
@@ -27,6 +27,9 @@ else
   CREW_NPROC = ENV["CREW_NPROC"]
 end
 
+# Set CREW_NOT_STRIP from environment variable
+CREW_NOT_STRIP = ENV["CREW_NOT_STRIP"]
+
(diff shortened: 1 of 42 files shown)


uberhacker:
Can we please add manpage compression to this PR also?  However, if you want to make it a separate PR, I understand completely.

jam7:
@uberhacker compressing manpage is difficult.  I'd like to that but different PR.  Or I appreciate if you would work on that.

lyxell:
Nice work as always

PR #804, opened 2017-06-25
Remove abort line from packages

uberhacker:
- Fix dropbox package to depend on python only if not installed yet
This also fixes the issue as discussed in PR #798.


thedamian:
Help me understand.

what's wrong with aborting if your platform is not supported. Got knows we're getting new platforms all the time!

uberhacker:
@thedamian: Thank you for looking into this issue.  If you follow the discussion thread on PR #798, @jam7 mentioned crew will output a message if the source or binaries are unavailable.  We both feel this message is suitable to handle any architectures that aren't included in the package.

jam7:
putting abort in a package file aborts not only crew install but also crew
search.

2017年6月26日(月) 13:15 Ed Reel <notifications@github.com>:

uberhacker:
I will add these changes in with the sha1 to sha256 PRs.

PR #806, opened 2017-06-26
Add compressdoc package

uberhacker:
Compress (with bzip2 or gzip) all man pages in a hierarchy and update symlinks.  See https://github.com/ojab/BLFS/blob/master/auxfiles/compressdoc.

### DIFF
--- a/packages/compressdoc.rb
+++ b/packages/compressdoc.rb
@@ -3,14 +3,15 @@ require 'package'
 class Compressdoc < Package
   description 'Compress (with bzip2 or gzip) all man pages in a hierarchy and update symlinks'
   homepage 'https://github.com/ojab/BLFS/blob/master/auxfiles/compressdoc'
-  version '8.0'
-  source_url 'https://github.com/ojab/BLFS/archive/8.0.tar.gz'
-  source_sha1 '1a4bcb8bd771701e91d9d525d8348dd6ec1c7db8'
+  version '9b2b12'
+  source_url 'https://github.com/ojab/BLFS/archive/9b2b12c0d809e287e1ea3fa4790a73a71feffbe4.zip'
+  source_sha1 'dd55e750128972ffd1022808b1b26cff2de242b1'
+
+  depends_on 'unzip'
 
   def self.install
-    system "wget https://raw.githubusercontent.com/ojab/BLFS/master/auxfiles/compressdoc"
-    system "chmod +x compressdoc"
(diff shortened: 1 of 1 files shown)


lyxell:
Is there a way to install this package with the standard procedure?
I.e. download a tar-ball, copy the files needed to CREW_DEST_DIR, and be done?

uberhacker:
@lyxell: I thought that is what I was doing?  The GitHub tarball is just a throw away to get to the compressdoc file.  I actually use this same technique with a few other packages.  If you can suggest a better way and include a code snippet, that would be helpful.

lyxell:
The "throw away" seems like a bad practice to me, why download files and check their hashsum if you're not using them? Isn't there a way to only download the files you're actually using? And let Chromebrew check the hashsum automatically?

uberhacker:
Suggestions?  I'm certainly open to "what works".

lyxell:
Untested, but maybe something like this:
```
require 'package'

class Compressdoc < Package
  source_url 'https://github.com/ojab/BLFS/archive/9b2b12c0d809e287e1ea3fa4790a73a71feffbe4.zip'
  source_sha1 'dd55e750128972ffd1022808b1b26cff2de242b1'
  version '9b2b12c0d809e287e1ea3fa4790a73a71feffbe4'

  depends_on 'unzip'

  def self.install
    system "mkdir -p #{CREW_DEST_DIR}/usr/local/bin"
    system "chmod +x auxfiles/compressdoc"
    system "cp auxfiles/compressdoc #{CREW_DEST_DIR}/usr/local/bin"
  end
end
```

uberhacker:
@lyxell: I appreciate your enthusiasm but the problem with this approach is the zip archive is a moving target unless it has a release tag.  If this archive is from the master branch then any new commits will render the sha1 hash invalid.  With that said, you could probably say the same thing about `auxfiles/compressdoc`.  We might be better off just removing the whole manual checksum line.  If the download is corrupted and you can't install then no biggie.  You just try again.

jam7:
I think any new commits take its own tag, so I guess this particularly
tagged zip file won't modified.  just thought, I've not tried it yet.

2017年6月26日(月) 14:44 Ed Reel <notifications@github.com>:

uberhacker:
OK, you guys convinced me.  I just learned how to grab the latest archive from the project page.  I have pushed up a new commit with the proposed changes.  Thank you for the help.

lyxell:
@jam7 Is correct.

PR #809, opened 2017-06-26
Dart version bump, arch update, update binary source, fix #780

### LINKED ISSUE #780: Dart package needs work
This package does not install.  After examining the output, it appears `/usr/bin`, `/usr/include` and `/usr/lib` directories in the tarball should be changed to `/usr/local/...` instead.  Also, the dart.rb package file is missing description and homepage metadata.

dantrevino:
Dart version bump to current.  Add arm7 architecture.  Download from official repositories. Fix directory install location issues  (#780).

### DIFF
--- a/packages/dart.rb
+++ b/packages/dart.rb
@@ -1,6 +1,5 @@
 require 'package'
 
-depends_on 'unzip'
 
 class Dart < Package
   description 'The Dart SDK is a set of tools and libraries for the Dart programming language.  You can find information about Dart online at dartlang.org.'
@@ -27,6 +26,9 @@ class Dart < Package
 #  else
 #    abort 'Unable to install dart sdk.  Architecture not supported.'.lightred
   end
+
+  depends_on 'unzip'
+
   def self.install
     system "mkdir -p #{CREW_DEST_DIR}/usr/local"
     system "cp -r bin/ #{CREW_DEST_DIR}/usr/local"
--- a/packages/dart.rb
+++ b/packages/dart.rb
@@ -1,5 +1,7 @@
 require 'package'
 
+depends_on 'unzip'
+
 class Dart < Package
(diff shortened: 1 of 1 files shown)


uberhacker:
@dantrevino: Thank you for this PR.  I think we should probably remove the `else ... abort` lines.  If you are interested, you can follow the discussion over on PR #798.

dantrevino:
Nice.  I'll update.

lyxell:
Nice!

uberhacker:
This package gets my vote as merge worthy.

uberhacker:
Actually, I just tested this and it doesn't install.  See below:
```
cp: cannot stat ‘man/’: No such file or directory
dart failed to install: `cp -r man/ /usr/local/tmp/crew/dest/usr/local` exited with 1
```
Please change as shown below:
```
- system "cp -r man/ #{CREW_DEST_DIR}/usr/local"
- system "cp -r include/ #{CREW_DEST_DIR}/usr/local"
```
Apparently, there are no manpages in the source.

dantrevino:
Odd.  I committed that change as part of the original.  Not sure what I
missed.

dantrevino:
Commit 3 fix to man directory.  thx @uberhacker

uberhacker:
@dantrevino: You will need to put the `depends_on 'unzip'` line inside the `class Dart < Package` block.  The following error occurs with that line outside the class.
```
/usr/local/lib/crew/packages/dart.rb:3:in `<top (required)>': undefined method `depends_on' for main:Object (NoMethodError)
```
You will probably want the depends_on line right above the `def self.install` line.  That should fix it.  Thanks for persevering. :)

uberhacker:
@dantrevino:  Nice work.  You found the end of the tunnel!! :)

PR #810, opened 2017-06-26
Change crew to compress manual files at install time

jam7:
I was looking for a way to install `compressdoc`.  Thank you @uberhacker to make it as a package.  This PR change crew to use `compressdoc` at install time if `compressdoc` is available and `CREW_NOT_COMPRESS` environment variable is not defined.

Tested on x86_64.

### DIFF
--- a/crew
+++ b/crew
@@ -28,7 +28,7 @@ else
 end
 
 # Set CREW_NOT_COMPRESS from environment variable
-CREW_NOT_COMPRESS = `bash -c "type compressdoc 2> /dev/null"`.empty? || ENV["CREW_NOT_COMPRESS"]
+CREW_NOT_COMPRESS = ENV["CREW_NOT_COMPRESS"]
 
 # Set CREW_NOT_STRIP from environment variable
 CREW_NOT_STRIP = ENV["CREW_NOT_STRIP"]
@@ -425,13 +425,23 @@ def build_and_preconfigure (target_dir)
   end
 end
 
+def compress_doc (dir)
+  # check whether crew should compress
+  return if CREW_NOT_COMPRESS || !File.exist?("#{CREW_PREFIX}/bin/compressdoc")
+
+  if Dir.exist? dir
+    system "find #{dir} -type f ! -perm -200 | xargs -r chmod u+w"
+    system "compressdoc --gzip -9 #{dir}"
+  end
+end
+
 def prepare_package (destdir)
   Dir.chdir destdir do
     # compress manual files
-    system "compressdoc --gzip -9 #{destdir}#{CREW_PREFIX}/man > /dev/null" unless CREW_NOT_COMPRESS
-    system "compressdoc --gzip -9 #{destdir}#{CREW_PREFIX}/info > /dev/null" unless CREW_NOT_COMPRESS
-    system "compressdoc --gzip -9 #{destdir}#{CREW_PREFIX}/share/man > /dev/null" unless CREW_NOT_COMPRESS
-    system "compressdoc --gzip -9 #{destdir}#{CREW_PREFIX}/share/info > /dev/null" unless CREW_NOT_COMPRESS
+    compress_doc "#{destdir}#{CREW_PREFIX}/man"
+    compress_doc "#{destdir}#{CREW_PREFIX}/info"
+    compress_doc "#{destdir}#{CREW_PREFIX}/share/man"
+    compress_doc "#{destdir}#{CREW_PREFIX}/share/info"
 
     # create directory list
     system "find . -type f > ../filelist"
--- a/crew
+++ b/crew
@@ -28,7 +28,7 @@ else
 end
 
 # Set CREW_NOT_COMPRESS from environment variable
-CREW_NOT_COMPRESS = ENV["CREW_NOT_COMPRESS"]
+CREW_NOT_COMPRESS = `bash -c "type compressdoc 2> /dev/null"`.empty? || ENV["CREW_NOT_COMPRESS"]
 
 # Set CREW_NOT_STRIP from environment variable
 CREW_NOT_STRIP = ENV["CREW_NOT_STRIP"]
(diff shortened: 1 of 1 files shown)


crew
```diff
@@ -27,6 +27,9 @@ else
   CREW_NPROC = ENV["CREW_NPROC"]
 end
 
+# Set CREW_NOT_COMPRESS from environment variable
+CREW_NOT_COMPRESS = `bash -c "type compressdoc 2> /dev/null"`.empty? || ENV["CREW_NOT_COMPRESS"]
+
```
uberhacker: Can we use more Ruby-like logic here instead of the bash test?

crew
```diff
@@ -420,15 +423,23 @@ end
 
 def prepare_package (destdir)
   Dir.chdir destdir do
-    #create directory list
+    # compress manual files
+    system "compressdoc --gzip -9 #{destdir}#{CREW_PREFIX}/man > /dev/null" unless CREW_NOT_COMPRESS
+    system "compressdoc --gzip -9 #{destdir}#{CREW_PREFIX}/info > /dev/null" unless CREW_NOT_COMPRESS
+    system "compressdoc --gzip -9 #{destdir}#{CREW_PREFIX}/share/man > /dev/null" unless CREW_NOT_COMPRESS
+    system "compressdoc --gzip -9 #{destdir}#{CREW_PREFIX}/share/info > /dev/null" unless CREW_NOT_COMPRESS
```
uberhacker: I ran across a situation in my testing where some of the man page files were read-only.  Might want to set permissions and ownership just in case, especially since you are redirecting standard error to null. :)

crew
```diff
@@ -420,15 +423,23 @@ end
...
+    # create directory list
     system "find . -type f > ../filelist"
     system "find . -type l >> ../filelist"
     system "cut -c2- ../filelist > filelist"
-    #create file list
+
+    # create file list
     system "find . -type d > ../dlist"
     system "cut -c2- ../dlist > dlistcut"
     system "tail -n +2 dlistcut > dlist"
```
uberhacker: Does this handle updating the package .filelist and .directorylist files with the new `.gz` manpages in `/usr/local/etc/crew/meta`?

crew
jam7: I found something at stackoverflow but I'm one liner, so I just preferred this way. :)

crew
jam7: Yes.  In order to update filelist correctly, I put compressdoc here.

crew
jam7: Thanks.  I'll modify crew to fix this issue.

crew
uberhacker: What is wrong with this?
```
CREW_NOT_COMPRESS = ENV["CREW_NOT_COMPRESS"] unless File.exists? '/usr/local/bin/compressdoc'
```

jam7:
@uberhacker Thank you for the review.  I updated commit.

crew
jam7: ahh, maybe that's good enough for this.  I looked at ruby implemented which function.

crew
```diff
@@ -27,6 +27,9 @@ else
   CREW_NPROC = ENV["CREW_NPROC"]
 end
 
+# Set CREW_NOT_COMPRESS from environment variable
+CREW_NOT_COMPRESS = `bash -c "type compressdoc 2> /dev/null"`.empty? || ENV["CREW_NOT_COMPRESS"]
+
 # Set CREW_NOT_STRIP from environment variable
 CREW_NOT_STRIP = ENV["CREW_NOT_STRIP"]
```
uberhacker: Now that I think of it, shouldn't we have a test to see if the `strip` command is available first and then install the `binutils` package if not?

crew
```diff
@@ -27,6 +27,9 @@ else
   CREW_NPROC = ENV["CREW_NPROC"]
 end
 
+# Set CREW_NOT_COMPRESS from environment variable
+CREW_NOT_COMPRESS = ! File.exist?("#{CREW_PREFIX}/bin/compressdoc") || ENV["CREW_NOT_COMPRESS"]
```
uberhacker: I know I'm probably overthinking this and should just get back to my day job but why don't we just install the `compressdoc` package if the command is not found?  Then you don't need any of this logic.  Also, if users are unaware that they need to install `compressdoc`, then we potentially get a bunch of packages that are installed that need to be reinstalled so the manpages are compressed.  If this is the case, then we would need a `crew compress` command to change all installed packages in one process.  Do you understand what I'm getting at here?

jam7: Since I don't prefer that way, installing `compressdoc` automatically.  It's user's choice.

Regarding to making `crew compress` command, I like it.  I would like to upgrade filelist also if manual pages are compressed, but it's too much work to do.  So, I'm not asking you to do so.  I also like to have `crew shrink`.

uberhacker: Since when did users know what is best? :-P  Sorry, now I'm sounding like a certain company in the northwest corner of the US.  Compressing manpages by default is actually not a bad idea in this case since Chromebooks have notoriously small hdds.

jam7:
Well, I updated whatever I can although this conversation doesn't seems so.  :(

Tested on x86_64 and working well.

crew
```diff
@@ -439,14 +457,14 @@ def install_package (pkgdir)
...
+    system "find . -type f -name 'lib*.a' -print | xargs -r sh -c 'for i in \"$0\" \"$@\"; do case \"$(head -c 4 $i)\" in ?ELF|\!?ar) echo \"$i\";; esac ; done' | xargs -r strip -S" unless CREW_NOT_STRIP
+    system "find . -type f -name 'lib*.so*' -print | xargs -r chmod u+w" unless CREW_NOT_STRIP
+    system "find . -type f -name 'lib*.so*' -print | xargs -r sh -c 'for i in \"$0\" \"$@\"; do case \"$(head -c 4 $i)\" in ?ELF) echo \"$i\";; esac ; done' | xargs -r strip -S" unless CREW_NOT_STRIP
 
     # Strip binaries
-    system "find . -type f -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d' | xargs chmod u+w 2>/dev/null" unless CREW_NOT_STRIP
-    system "find . -type f -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d' | xargs strip 2>/dev/null" unless CREW_NOT_STRIP
+    system "find . -type f -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d' | xargs -r chmod u+w" unless CREW_NOT_STRIP
+    system "find . -type f -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d' | xargs -r sh -c 'for i in \"$0\" \"$@\"; do case \"$(head -c 4 $i)\" in ?ELF) echo \"$i\";; esac ; done' | xargs -r strip" unless CREW_NOT_STRIP
```
uberhacker: Is it a safe assumption to assume _ALL_ shared library object files start with `^lib`?  Wouldn't we potentially miss other library files that need stripped?

jam7: Why not safe?  What is danger?

uberhacker: No danger, just wondering if some files might be missed.

jam7:
Rebased to avoid conflicts.

PR #812, opened 2017-06-26
Change crew comand to check strip command existence

jam7:
This PR fixes https://github.com/skycocker/chromebrew/pull/810/files/bb62f2cd4d17dc7a8efda4767b58f15b00a82287#r124056419 issue.

### DIFF
--- a/crew
+++ b/crew
@@ -28,7 +28,7 @@ else
 end
 
 # Set CREW_NOT_STRIP from environment variable
-CREW_NOT_STRIP = ENV["CREW_NOT_STRIP"]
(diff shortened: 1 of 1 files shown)


jam7:
Closing this PR since executing strip command even if it is not exist doesn't hurts.  Moreover, it works well if `strip` command is installed in the middle of compile because of dependencies.  If we modify `crew` to check strip command existence at the beginning, files are not stripped.

1. $ crew install something
1. ... (somthing requires buildessential, so binutils is installed)
1. ... (at the end of compile, `strip` command is executed and worked well)

uberhacker:
I think we should check if the command exists before attempting to execute.  That is the more elegant way to do it.

Issue #797, opened 2017-06-25
make build fails


dantrevino:
chronos@localhost / $ crew install make keep
...
./config.status: line 3183: cmp: command not found
pkgconfig failed to install: `make -j4` exited with 127

jam7:
It is caused by a difficult to solve dependency problem.  :(

You can avoid this problem by `crew install buildessential`.

jam7:
This issue will be fixed once #815 is merged.

jam7:
Close in favor of #815.

PR #815, opened 2017-06-27
Change crew to not automatically install buildessential

jam7:
Once I modified to crew to automatically install `buildessential` if users try to compile package from source, but this behavior causes #797 problem because `crew install make` tries to install `pkgconfig` as a part of `buildessential` although `make` is required to compile `pkgconfig`.

This PR solves this problem by modifying `crew` to simply not install `buildessential` automatically.  Even with this PR, `crew` still shows error message to ask users to install `buildessential` when users try to install source packages.  So, this PR won't break anything.

Tested on x86_64.

### DIFF
--- a/crew
+++ b/crew
@@ -476,9 +476,6 @@ end
 def expand_dependencies
   @dependencies = []
 
-  # check source packages existance
-  @source_package = 0
-
   def push_dependencies
     if @pkg.is_binary?(@device[:architecture]) ||
        (!@pkg.in_upgrade && !@pkg.build_from_source && @device[:installed_packages].any? { |pkg| pkg[:name] == @pkg.name })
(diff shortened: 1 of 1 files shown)


uberhacker:
I agree but if `buildessential` is needed, why don't we just include it `install.sh`?  Then we don't even need a package.

jam7:
`buildessential` is not needed.

uberhacker:
That's like saying users will never need to install anything except binary packages. :-P

jam7:
I prefer to distribute more and more binary packages like what debian/homebrew does.  Don't you?

uberhacker:
As long as _YOU_ maintain it, fine. :-P  I think we should always give users the ability to build any package from source even if the binary is available.  Having that flexibility makes it possible to introduce new architectures as more manufacturers produce Chromebooks.  That is one thing I really like about this project.

jam7:
I'm not understand what you want.  Yes, it is possible to install `buildessential`.  Yes, user can choose to install it.  Why are you asking to force `install.sh` to install `buildessential`?  Is it really matter?

uberhacker:
As long as you make it easy for users to compile packages, I'm fine with doing it that way.  I agree we should only install when necessary.

jam7:
Is there any problem on this PR?  Please let me know if any.  Thanks.

PR #822, opened 2017-06-28
Add aspell package

uberhacker:
GNU Aspell is a Free and Open Source spell checker designed to eventually replace Ispell. It can either be used as a library or as an independent spell checker. Its main feature is that it does a superior job of suggesting possible replacements for a misspelled word than just about any other spell checker out there for the English language.  See http://aspell.net/.


uberhacker:
@lyxell:  Just so you know, aspell_en and aspell_es packages depend on this package and those were merged into master 2 days ago. :)

PR #824, opened 2017-06-28
Add shtool package

uberhacker:
GNU shtool is a compilation of small but very stable and portable shell scripts into a single shell tool. All ingredients were in successful use over many years in various free software projects.  See http://savannah.gnu.org/projects/shtool.


thedamian:
There's conflicts @uberhacker  and I can't seem to be able to resolve it online.

uberhacker:
@thedamian: Can you please post the conflict errors you see?

Issue #796, opened 2017-06-24
Wine on aarch64


flyingP0tat0:
If I undertstand #780 correctly, `aarch64` uses 32-bit mode?
The thing is, that I set up wine to build with 64-bit support on `aarch64`. Could someone please build wine on `aarch64` and let me know about the result?

lyxell:
Correct, wine installation for aarch64 is probably broken.
Another way to test this is to try to install the package on armv7 with the 64-bit flag enabled.

flyingP0tat0:
I guess I will just edit the package...

flyingP0tat0:
Possible problem is acknowledged, change is proposed.

PR #832, opened 2017-06-29
Update wine

flyingP0tat0:
Change aarch64 to use 32-bit install method.
Hopefully solves #796.

### DIFF
--- a/packages/wine.rb
+++ b/packages/wine.rb
@@ -15,9 +15,9 @@ class Wine < Package
 
   def self.build
     case ARCH
-    when "i686" || "armv7l"
+    when "i686" || "armv7l" || "aarch64"
       system "./configure --without-x"
-    when "x86_64" || "aarch64"
+    when "x86_64"
       system "./configure --without-x --enable-win64"
     else
       abort "Error getting your device configuration."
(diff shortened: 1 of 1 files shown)


uberhacker:
@flyingP0tat0: Good call.  Thanks for this PR.  Can you please add another commit to avoid a merge conflict with PR #836?  See below:
```
-  source_sha1 '886bc24c315b2accbc2ad41d921e92177768f3a5'
 +  source_sha256 '488df7ffd2e81da455bf428fc9eb784bb4273a890334500895665711bd52f179'
```

flyingP0tat0:
@uberhacker Done.

lyxell:
@uberhacker It wouldn't have created a conflict, git is smart enough to automatically merge files where there's changes but on different lines.

uberhacker:
@lyxell: Right, but if my PR was merged in first, his PR would have reverted the sha256 change back to sha1 which would have been even worse than a merge conflict.

lyxell:
@uberhacker No, both pull requests would have merged fine.

uberhacker:
@lyxell: My understanding is either of us would have needed to resolve merge conflicts if one PR was merged in before the other.  The reason it would create a merge conflict is the code being merged would not match master without resolving locally and pushing again.  Believe me, I've ran across this situation many times regarding pending PRs with changes on the same file.

uberhacker:
I might be wrong about this but if git works anything like patches, if any of the line numbers don't match in the patch, it won't apply successfully.  In our case, we were lucky since there weren't any lines changed above the requested change(s).  In any event, we should be careful when merging multiple pending PRs that request changes on the same file.

Issue #818, opened 2017-06-27
Fake package bundles do not show '(i)' (aka installed) in crew search


uberhacker:
An example package is `buildessential`.  I think the expected behavior is to show fake packages as installed if all the `depend_on` packages are currently installed.

uberhacker:
Fixed in PR #833.

PR #833, opened 2017-07-01
Change crew to check internal hash table instead of directory

jam7:
This PR fixes issue #818.  This also increase the performance of print_package.

Tested on x86_64.


jam7:
I've noticed the output from regexp_search is not sorted.  So, add sort call there too.

uberhacker:
I like it!  Nice work @jam7.

PR #841, opened 2017-07-01
Add help2man package

uberhacker:
help2man produces simple manual pages from the ‘--help’ and ‘--version’ output of other commands.  See https://www.gnu.org/software/help2man/.

### DIFF
--- a/packages/help2man.rb
+++ b/packages/help2man.rb
@@ -4,8 +4,8 @@ class Help2man < Package
   description "help2man produces simple manual pages from the '--help' and '--version' output of other commands."
   homepage 'https://www.gnu.org/software/help2man/'
   version '1.47.4'
-  source_url 'http://gnu.mirror.globo.tech/help2man/help2man-1.47.4.tar.xz'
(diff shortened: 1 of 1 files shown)


lyxell:
Why is the globo.tech mirror used?

uberhacker:
@lyxell: I believe that was the default mirror I received during the download request.  What other mirror(s) would you suggest?

lyxell:
@uberhacker http://ftpmirror.gnu.org/help2man/help2man-1.47.4.tar.xz picks a mirror close to the user automatically and is more likely to be available at all times than using a definite mirror.

uberhacker:
Thanks, I wasn't aware of that.  I will make sure I use that url on other projects from gnu.org.  New commit with the change requested has been pushed.

lyxell:
@uberhacker It should probably use https:// as well, sorry for missing that.

uberhacker:
@lyell: Change pushed.  I also changed the pending PRs for coreutils, datamash and shtool packages.

PR #842, opened 2017-07-02
Add parallel package

uberhacker:
Run multiple jobs at once.  This `is_fake` package. :)  See https://joeyh.name/code/moreutils/.

### DIFF
--- a/packages/parallel.rb
+++ b/packages/parallel.rb
@@ -1,0 +1,11 @@
+require 'package'
+
+class Parallel < Package
+  description 'Run multiple programs simultaneously.'
+  homepage 'https://joeyh.name/code/moreutils/'
+  version '0.60'
+
+  is_fake
+
+  depends_on 'moreutils'
+end


lyxell:
?

uberhacker:
@lyxell: LoL, check the source. :)

uberhacker:
@lyxell: Maybe this will help clear things a bit.  See below:
```
moreutils is a growing collection of the unix tools that nobody thought to write long ago when unix was young.

It began when I blogged:

I'm a fan of the unix tools philosophy, but I sometimes wonder if there's much room for new tools to be added to that toolbox. I've always wanted to come up with my own general-purpose new unix tool.

Well, after lots of feedback documented in the many followups (1 2 3) in my blog, I've concluded:

Maybe the problem isn't that no-one is writing them, or that the unix toolspace is covered except for specialised tools, but that the most basic tools fall through the cracks and are never noticed by people who could benefit from them.

And so the moreutils collection was born, to stop these programs from falling through the cracks.

What's included

Probably the most general purpose tool in moreutils so far is sponge(1), which lets you do things like this:

% sed "s/root/toor/" /etc/passwd | grep -v joey | sponge /etc/passwd
There are lots more listed below, and I'm always interested to add more to the collection, as long as they're suitably general-purpose, and don't duplicate other well-known tools.

chronic: runs a command quietly unless it fails
combine: combine the lines in two files using boolean operations
errno: look up errno names and descriptions
ifdata: get network interface info without parsing ifconfig output
ifne: run a program if the standard input is not empty
isutf8: check if a file or standard input is utf-8
lckdo: execute a program with a lock held
mispipe: pipe two commands, returning the exit status of the first
parallel: run multiple jobs at once  <<<< Notice the command is included in the moreutils package
pee: tee standard input to pipes
sponge: soak up standard input and write to a file
ts: timestamp standard input
vidir: edit a directory in your text editor
vipe: insert a text editor into a pipe
zrun: automatically uncompress arguments to command
```

thedamian:
OH! this is one of those "alias" packages, that all it does is load the main large package that will have "this parallel" package for you"

PR #855, opened 2017-07-05
imagemagick@6 formula

vince:
Homebrew has an equivalent formula called [imagemagick@6 ](https://github.com/Homebrew/homebrew-core/blob/master/Formula/imagemagick%406.rb)which is used because the latest rmagick gem is not compatible with ImageMagick 7. If anyone is using chromebrew to do ruby development this may come in handy.


vince:
I shut down this fork so will make a new PR

PR #862, opened 2017-07-08
Add pip package

uberhacker:
Unless you install the python27 package, pip is unavailable even if the python and/or python3 packages are installed.  Sometimes you need python version 3 which justifies having this package available.

### DIFF
--- a/packages/pip.rb
+++ b/packages/pip.rb
@@ -7,12 +7,16 @@ class Pip < Package
   source_url 'https://github.com/pypa/pip/archive/9.0.1.tar.gz'
   source_sha256 'd03fabbc4fbf2fbfc2f97307960aef2b3ca4c880ecda993dcc35957e33d7cd76'
 
+  # Don't overwrite the currently installed version.
+  abort 'pip is already installed.'.lightgreen if File.exists? '/usr/local/bin/pip'
+
+  # Install python if not found.
   depends_on 'python' unless File.exists? '/usr/local/bin/python'
 
   def self.install
     system "wget https://bootstrap.pypa.io/get-pip.py"
     system "mkdir -p #{CREW_DEST_DIR}/usr/local/bin"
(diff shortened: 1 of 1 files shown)


jam7:
I thought I replied this one but it doesn't appear here.  Writing it again from web UI.

I think python.rb is out of date.  You can install python3.rb instead and it install pip from the beginning.  We may think to remove python.rb.

uberhacker:
What do you think about removing the python.rb package and renaming python3.rb to python.rb?  I think the confusion for users is they will not know that they need to enter the `python3` and `pip3` commands to execute.  But if we can create symlinks from python3.6 to python and pip3.6 to pip if the python27.rb package is not installed, that might be the ticket.

jam7:
As @lyxell explained you several weeks ago, /usr/bin/python is python 2.7 on many linux distributions.  So, I think pretending that python3 is python is not a good idea.  I already saw a few packages has wrong depends_on line because of command name conflicts.  Making such symbolic link increase the possibility of wrong depends_on lines in the future.

I have no opinion about the package name.  I have above opinion about making such symbolic links.

uberhacker:
The symlinks could work with the following caveats.  Both packages would need to look for existing installations and NOT add the symlinks/binaries if they're already installed.  The only time you would need the symlinks added is if the python3.rb package is the only python version installed.  Likewise, if the python27.rb package is installed after python3.rb, then python27 should NOT add /usr/local/bin/python or /usr/local/bin/pip.

uberhacker:
Closing since apparently, we want Chromebrew to be like every other package manager.

jam7:
FYI, python3 is not compatible with python2.  It's the reason why ppl leave `python` as is and install python 3.x as `python3`.  You can google about differences.

Anyway, leave `pyhon` intact is good manner for few more years, I think.

PR #864, opened 2017-07-08
Update elixir from 1.3.1 to 1.4.5

uberhacker:
This should also fix elixir so it installs.


jam7:
Nice.  I could compile it on armv7l.  I can try elixir now.  :)

PR #866, opened 2017-07-08
Add metasploit package

MD5HashBrowns:
From [Wikipedia](https://en.wikipedia.org/wiki/Metasploit_Project): "[The] Metasploit Framework [is] a tool for developing and executing exploit code against a remote target machine."

This package "works" (runs msfconsole, can connect to the postgres database, etc), but not I'm sure I built it the proper way. I used `bundler install` to install required dependency gems, which aren't removed upon `crew remove metasploit`. I also moved copied the entire framework to /usr/local/share, which seemed to spit out quite a few errors regarding `strip`, which I was unable to resolve. Everything _appears_ to work fine, however.

### DIFF
--- a/packages/metasploit.rb
+++ b/packages/metasploit.rb
@@ -1,0 +1,37 @@
+require 'package'
+
+class Metasploit < Package
+  description 'The Metasploit Framework is a tool for developing and executing exploit code against a remote target machine.'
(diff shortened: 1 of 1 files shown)


uberhacker:
Shady. :-P

PR #871, opened 2017-07-09
Add dependency to postgres package

uberhacker:


MD5HashBrowns:
I noticed after installing postgres that you also need `readline`

uberhacker:
Right you are.  I just confirmed.  Thanks for the heads up.  Dependency added.

PR #873, opened 2017-07-09
Change install.sh to check LD_LIBRARY_PATH and show guide message.

jam7:
This PR does what @flyingP0tat0 said in https://github.com/skycocker/chromebrew/issues/689#issuecomment-309211334.

I tried to modify several packages to install libraries into lib64 on x86_64.  It works.  However, it is not perfect, since anyone can possible to add new package and install libraries into lib by accident.

Therefore, I made this PR.  Tested on x86_64 and arvm7l.  install.sh shows following message on x86_64 after this PR.
```
Chromebrew installed successfully and package lists updated.

Several packages may install their libraries into /usr/local/lib, so adding below to your ~/.bash_profile is recommended.
  export LD_LIBRARY_PATH=/usr/local/lib64:/usr/local/lib
$
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -182,3 +182,14 @@ echo crew >> .git/info/sparse-checkout
 git fetch origin master
 git reset --hard origin/master
 echo "Chromebrew installed successfully and package lists updated."
+
+#check LD_LIBRARY_PATH on x86_64
+case "$architecture" in
+"x86_64")
+  (echo $LD_LIBRARY_PATH | grep '/usr/local/lib[^6]\|/usr/local/lib$' > /dev/null) || cat << EOF
+
+Several packages may install their libraries into $CREW_PREFIX/lib, so adding below to your ~/.bash_profile is recommended.
+  export LD_LIBRARY_PATH=/usr/local/lib64:/usr/local/lib
+EOF
+  ;;
+esac


flyingP0tat0:
Why not automatically setting the path at install instead of telling the user to do so?

jam7:
@flyingP0tat0 Because it's difficult to write perfect mechanism.  :(

install.sh
```diff
@@ -182,3 +182,14 @@ echo crew >> .git/info/sparse-checkout
 git fetch origin master
 git reset --hard origin/master
 echo "Chromebrew installed successfully and package lists updated."
+
+#check LD_LIBRARY_PATH on x86_64
+case "$architecture" in
+"x86_64")
+  (echo $LD_LIBRARY_PATH | grep '/usr/local/lib[^6]\|/usr/local/lib$' > /dev/null) || cat << EOF
```
flyingP0tat0: In this case, we could do a `echo "export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib" >> ~/.bashrc` and a `source ~/.bashrc`. This should do the trick. I don't think this could break anything, because it just adds a path to the library path, nothing will be overwritten.

jam7: IMHO, it breaks several stuff.
 - how about ppl share .bashrc among several machines
 - how about ppl who don't want to add /usr/local/lib

you can read discussions in the web.  For example, https://softwareengineering.stackexchange.com/questions/241257/should-a-package-manager-modify-your-bashrc-file.

uberhacker: The problem with adding configuration in `~/.bashrc`, `~/.zshrc`, etc. is it is not easily removed when the package requiring that configuration is removed.  Also, it requires some user interaction to get things working (a la `source ~/.bashrc`).  I'm of the opinion that any configuration specific to a package should be added in a separate file that Chrome OS can read when commands requiring that specific configuration are executed.  With that said, where is the ideal location?  We don't want to edit `/etc/profile` because that would require crew to use sudo to install and remove the package.  It would be nice if Chrome OS could read configuration files from somewhere like the `/usr/local/etc/profile.d` directory.

flyingP0tat0: But it is the actual program, crew, modfiying the path. Of course, no package should do that.
But given the problems ChromiumOS gives us, it seems reasonable for a program to add stuff to enviroment. If the crew gets delteted, the additional path won't hurt in any way.
@uberhacker `~/.bashrc` is automatically read every time a terminal gets opened, no need for further interactions.
@jam7 If someone really shares a ~/.bashrc between, he really knows what he is doing, I don't think this is recommended.
Why would you not want to add one single library path, if this is the only way to get your programs working?

lyxell: There's also the case of users using their Chromebooks in Guest mode, their `.bashrc` is reset on reboot (The only thing retained is `/usr/local`).

flyingP0tat0:
Ok, ok, I see that adding to `PATH` via `~/.bashrc` is not the best idea. (But to be honest, I see @lyxell 's point as the only valid one :D )
So what to do then? It has to be a non-crew-specific solution.
Maybe we could do something with a script at startup-time? I am not into that sort of things, but this just came to my mind...

thedamian:
I've had a LOT of problems modifying  LD_LIBRARY_PATH 

so I'm a little worried about this branch ,and at the same time i'd love to have it done right.

flyingP0tat0:
Well, there seems no other way than to modify `LD_LIBRARY_PATH`.
Just the way of doing it is what needs to be done. Do you agree on that?

jam7:
Please don't hijack my PR.  I just wanted to add a mechanism for notification message at install time in this PR.  If you want to discuss different mechanism, please open your own issue.

BTW, I don't agree any way to modify user's environment automatically.  I hate it.

flyingP0tat0:
Yes, will open a new one.

thedamian:
@jam7  I love this PR and I will merge it. I just wanted help understanding why this line (that i've added myself in the past, but now have commented out because it caused problems) why it's needed and what does it do. That's all.

jam7:
@thedamian This PR add mechanism to check LD_LIBRARY_PATH and let users know the need of adding `/usr/local/lib64` as a part of LD_LIBRARY_PATH on x86_64 architecture.  The source of this problem is that ChromsOS version 60 decides to use only `/usr/local/lib64` as LD_LIBRARY_PATH by the default.  Several libraries still install their libraries into `/usr/local/lib` on x86_64 architecture, so those won't work on new ChromeOS.

ChromeOS used to use `/usr/local/lib:/usr/local/lib64` as LD_LIBRARY_PATH on x86_64, so there was no problem before.  But, ChromeOS 60 doesn't, so we need to add some mechanism to let users know this fact. 
 This PR only let users know it on particular architecture to not cause any conflicts.  This checks LD_LIBRARY_PATH on x86_64 only, and this shows warning message if this finds a problem.

As a next step, it may be possible to change user's environment automatically, but it's different topic from this PR.

On the other hand, I modified several libraries to install their libraries into `/usr/local/lib64` on x86_64 in #794 and #816.  We can perform this for several important libraries, but not all.

Issue #536, opened 2017-02-19
Change installation directory to prevent dev_install problems


cstrouse:
A lot of Chromebook users who are trying to get a package manager setup aren't aware that the dev_install script will blow away anything they have under /usr/local during a reinstall or removal. We can do one simple thing to reduce the likelihood of a new Chromebrew user having a bad experience due to not understanding how Chromium's dev_install works.

I had installed rustup manually a while back and ended up with the rust package manager installed under /usr/local/.cargo. While working on Chromebrew I regularly used dev_install --uninstall as a quick way to blast away Chromebrew without doing a powerwash. One day noticed that .cargo was left intact during this process.

I think it would be a good idea to startup putting Chromebrew in /usr/local/.chromebrew by default or perhaps to check for the presence of dev_install during installation and using the alternate path if it's installed. If these two options aren't acceptable then adding an interactive prompt to the installer might work.

jam7:
After placing everything under /usr/local/.chromebrew, what do you think we need to do?

Adding /usr/local/.chromebrew/bin to PATH environment variable?
Or just making symbolic links from /usr/local/.chromebrew/* to /usr/local?

cstrouse:
A symlink is probably better than requiring path modifications.

jam7:
I see.  It's easier also since it doesn't require recompile all binaries.

On the other hand, symbolic link method is probably difficult to avoid conflicts.  For example, it requires uses to make symbolic links first.

Basically I agree with changing install directories.  I like such idea.  But, there are several problems need to solve.

dragon788:
Linuxbrew simply provides instructions for adding an export to your shell profile to update the PATH variable. This would probably be the best approach as then it would completely avoid conflicts with dev install/emerged software.

jam7:
IMHO, I also prefer to put everything under .chromebrew and force users to modify their PATH.  But, I understand it's really really difficult...

lyxell:
There is people using their Chromebooks in guest-mode only, they cannot change their PATH.

cstrouse:
@lyxell @jam7 Would it make sense then to offer it as an option during installation or have it use .chromebrew if they're not in guest-mode?

uberhacker:
[...]

PR #877, opened 2017-07-10
Add uninstall.sh script

uberhacker:
This might not be the most elegant way but it works and it makes testing a lot easier than the extreme measure of the dreaded powerwash.

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 IFS=' '
-packages="$(crew search | grep '(i)' | cut -d':' -f1 | cut -d')' -f2 | sed 's,\[0m,,g' | tr -dc '[:print:]' | sed 's, ruby,,')"
+packages="$(crew search | grep '(i)' | cut -d':' -f1 | cut -d')' -f2 | sed 's,\[0m,,g' | tr -dc '[:print:]' | sed 's, ruby , ,')"
 for package in $packages; do
   crew remove $package
 done
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -1,7 +1,18 @@
 #!/bin/bash
-IFS=' '
 packages="$(crew search | grep '(i)' | cut -d':' -f1 | cut -d')' -f2 | sed 's,\[0m,,g' | tr -dc '[:print:]' | sed 's, ruby , ,')"
 for package in $packages; do
   crew remove $package
+  rm -rf /usr/local/db/$package
+  rm -rf /usr/local/etc/$package
+  rm -rf /usr/local/lib/$package
+  rm -rf /usr/local/lib64/$package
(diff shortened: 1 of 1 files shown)


thedamian:
shouldn't the last line be to remove the directory where crew lives as well:
/usr/local/lib/crew/

and the link in /usr/local/bin/crew

uberhacker:
@thedamian:  Good point.  It might be a good idea to recursively remove other directories under `/usr/local` as well since many packages build up residue with extra files post-installation.

uberhacker:
@thedamian: I added your suggestions and a few more cleanup lines.  Sad that it takes hours to install all your favorite packages and less than a minute to uninstall everything. :)

lyxell:
We can't be sure we're actually cleaning up everything, and the rm -rf's introduced might as well remove stuff that didn't originate from crew. I'm skeptic.

thedamian:
@lyxell 

It removes the packages installed by crew how it is.
The next thing would simply to remove /usr/local/lib/crew itself and it's link in /usr/localbin

I'm not suggesting we clean the whole /usr/local/lib directory.

If we're truly uninstalling crew, this script is wonderful, but In my opinion 90% there.

uberhacker:
@lyxell: I understand your concern.  However, some packages will add additional files after installation and we need a way to clean up this "residue" as part of the uninstall process.  In the case of node, python or ruby, for example, additional "packages" (ie. gems, npm and pip packages) can be installed post-installation.  We need an elegant way to remove these in addition to the main package.  I'm open to suggestions.

lyxell:
I think it is better to continue the discussion on https://github.com/skycocker/chromebrew/issues/536
This would make removal straightforward, we need to find a way to work around the problems with a custom installation path though.

jam7:
My suggestion is to stop removing files directly like `rm -rf /usr/local/...` way.  Nobody think something like that is safe.

You can simply performs only `crew remove ...`.  If you know several files are remain, you can shows warning messages.  Or you can remove them with prompt.  Or you can make different program named `uninstall-force` to remove additional files.

uberhacker:
@jam7: I'm of the opinion that residue should be removed without prompting.  `rm -rf /usr/local/...` is perfectly safe if you know the implications.  I have tested this on CloudReady which includes symlinks everywhere (that can certainly have unintended consequences).  Let me know specifically which `rm -rf` lines in this script are not safe and I'll remove them.

andzaytsev:
This script works great for me and is very useful. Is there anything that stops it from being merged?
Should there be a warning and a prompt before everything is removed?

uberhacker:
Why are we so concerned about recursively wiping out entire directories when we can simply Powerwash and restore everything?  The whole point of this script is so you don't need to go through the extreme measure of a Powerwash.

uberhacker:
Closing in favor of dev_install --uninstall

PR #881, opened 2017-07-11
Add package for The Silver Searcher.

vince:
The Silver Searcher is an alternative to ack that is a magnitude faster. It's wonderful for searching your local filesystem and is necessary if you're going to use the vim plugin of the same name. You can read all about the package [here](https://github.com/ggreer/the_silver_searcher).

### DIFF
--- a/packages/ag.rb
+++ b/packages/ag.rb
@@ -1,7 +1,7 @@
 require 'package'
 
-class The_silver_searcher < Package
-  description 'Code-search similar to ack. (ag)'
+class Ag < Package
+  description 'The Silver Searcher. Very fast search similar to ack or grep. (ag)'
   homepage 'https://github.com/ggreer/the_silver_searcher'
   version '2.0.0'
   source_url 'https://github.com/ggreer/the_silver_searcher/archive/2.0.0.tar.gz'
--- a/packages/the_silver_searcher.rb
+++ b/packages/the_silver_searcher.rb
@@ -1,0 +1,26 @@
+require 'package'
+
+class The_silver_searcher < Package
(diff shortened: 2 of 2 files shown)


uberhacker:
@vince: Thank you for adding this package.  I just tried it out and it looks amazing!  Much faster than using `grep -r something`.  I'm thinking you might want to change the package object name from `The_silver_searcher` to `Ag` instead.  Usually, it's a good idea to name the package the same as the command to avoid confusion.  You can still include silver searcher in the description so it can be found in searches.  If you change the package name and include `silver searcher` in the description, `crew search ag` and `crew search silver` will both return results.

vince:
Arch and most other distros list it as "the_silver_searcher" which is why I changed it from Ag. Debian uses silversearcher-ag. Would that be a good compromise?

    silversearcher-ag - very fast grep-like program, altenative to ack-grep

Wow, just noticed that they have a typo in their description. :)

uberhacker:
I never understood why we needed to be like other distros but since that has `ag` in the name, it would be easily searchable.  `crew install ag` is much easier to type. :)

flyingP0tat0:
+1 for `ag`

vince:
Ok, this is done. Do I need to do a new PR or will [d3d8fce](https://github.com/skycocker/chromebrew/pull/881/commits/d3d8fce6877fc6c72e0cbbc26d43de3ad04ac860) automatically be applied?

uberhacker:
No need for another PR.  Just push with the new commit.

lyxell:
Thanks!

PR #885, opened 2017-07-14
Add putty package

uberhacker:
Free Telnet, SSH, and Rlogin clients plus a terminal emulator

PuTTY is a free implementation of Telnet, SSH, and Rlogin for Win32 and Unix platforms, along with an xterm terminal emulator. The package includes: 

pscp - SCP client (command-line secure file copy)
psftp - SFTP client (general file transfer sessions much like FTP)
plink - Command-line interface to the PuTTY back ends
puttygen - RSA and DSA key generation utility

See https://www.chiark.greenend.org.uk/~sgtatham/putty/.

### DIFF
--- a/packages/putty.rb
+++ b/packages/putty.rb
@@ -1,0 +1,18 @@
+require 'package'
+
+class Putty < Package
+  description 'Free Telnet, SSH, and Rlogin clients plus a terminal emulator'
+  homepage 'https://www.chiark.greenend.org.uk/~sgtatham/putty/'
+  version '0.70'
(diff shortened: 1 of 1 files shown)


thedamian:
@uberhacker 

I LOVE putty. But it's only for windows no? I kean gen, sftp and scp are already built in.

I don't know if we want to add this (Really. I had to look at putty 3 times because I was SURE it was simply a replacement for the built in linux apps but for windows.)

what am I missing.

uberhacker:
Putty works on all systems.  This package provides alternative versions of sftp, scp and ssh-keygen.

thedamian:
weird. But alright. It's merged.

PR #886, opened 2017-07-14
Add qt package

uberhacker:
Qt is a comprehensive cross-platform framework and toolkit that helps you create and build native applications and user interfaces for all the screens of your end user.  See https://info.qt.io/download-qt-for-application-development.

Guys, I think this is the first gui app installable via crew and the coolest thing is it lets you create cross-platform gui apps. :)

### DIFF
--- a/packages/qt.rb
+++ b/packages/qt.rb
@@ -1,0 +1,83 @@
+require 'package'
+
+class Qt < Package
+  description 'Qt is a comprehensive cross-platform framework and toolkit that helps you create and build native applications and user interfaces for all the screens of your end user.'
+  homepage 'https://info.qt.io/download-qt-for-application-development'
+  version '5.9.1'
+  source_url 'http://download.qt.io/official_releases/qt/5.9/5.9.1/Qt-591-qtbase-patch-for-IOS-QTBUG-61690.zip'
+  source_sha256 'a4cfa103b480e67ebc0c91d02845f94bf443c60f847980fa5b38a1303bd6e985'
+
+  depends_on 'unzip'
+
+  def self.install
+    if ( ARCH == 'i686' || ARCH == 'x86_64' )
+      system "echo 'function Controller() {' > qt-installer-script.qs"
+      system "echo '    installer.autoRejectMessageBoxes();' >> qt-installer-script.qs"
+      system "echo '    installer.installationFinished.connect(function() {' >> qt-installer-script.qs"
+      system "echo '        gui.clickButton(buttons.NextButton);' >> qt-installer-script.qs"
+      system "echo '    })' >> qt-installer-script.qs"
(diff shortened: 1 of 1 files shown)


flyingP0tat0:
There is no way this would work in the current situation, as we have no X11.
Have you tried to install it?

uberhacker:
@flyingP0tat0: Dude, it totally works.  I have tested it and all the static libraries necessary to get it working are embedded into the package.  For what it's worth, this emulates a RedHat environment and only works on x86_64 and i686 systems.

flyingP0tat0:
Well, I tested it on my x86_64 HP Chromebook 13 G1 and it did not work because of missing `libx`.

flyingP0tat0:
Are you sure you have not compiled any libraries for that before?

uberhacker:
OK, so it installs on CloudReady since the libX11 library files exist.  I just tried it on my HP Chromebook and the libX11 libraries don't exist.  So you are right about that.  However, now that I know it works with CloudReady, I'm thinking X11 is very possible.

flyingP0tat0:
Yes it may be. But I guess the tricky thing is, that Chrome OS uses some sort of Wayland-Port, if i remember correctly. So after installing libX11, it wont find any display, X-org-server is needed. Which probably also won't work given the Wayland thing. So we will probably need Weston. I am trying that out right now, but that will take some time.

thedamian:
This would be revolutionary but I agree with flyingP0tat0 in that I don't know if this will work, and we might just get Xlib working first. since THAT will make EVERYTHING work on chrome!

flyingP0tat0:
@thedamian The issue is way bigger than just getting libX11, there is a lot to be done. I will open a new issue to keep track of getting all of this to work on Chrome OS.

thedamian:
do do we close this PR without merging?

uberhacker:
Yes, but there is a glimmer of hope for gui apps since I could install this in CloudReady.

JL2210:
Hey, @uberhacker, open this again and merge it.

sajjaadf:
Please

uberhacker:
@sajjaadf: Been there, done that.  Please try to install the latest qt and submit an issue if you have any trouble.

sajjaadf:
Later Today
is it Crew Install qt?

uberhacker:
Yes, `crew install qt`.  The command to execute after installed is `qtcreator`.

sajjaadf:
Thanks, Btw See my The issue with creating Gui App store and Link apps to
the icon. I'm working on it.

>

sajjaadf:
And Android Studio?

uberhacker:
I did.  Thank you for adding the issues.  Nice to see interest in improving Chromebrew.

PR #891, opened 2017-07-16
Update inetutils package

uberhacker:
- Add ncurses dependency
- Disable traceroute

### DIFF
--- a/packages/inetutils.rb
+++ b/packages/inetutils.rb
@@ -8,11 +8,13 @@ class Inetutils < Package
   source_sha256 '849d96f136effdef69548a940e3e0ec0624fc0c81265296987986a0dd36ded37'
 
   depends_on 'buildessential'
+  depends_on 'ncurses'
 
   def self.build
-    system "./configure --prefix=/usr/local        \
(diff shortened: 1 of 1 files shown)


thedamian:
All look good (especially the ncurses requirements if it's needed) 
but why the disable of traceroute?

uberhacker:
That is a good question.  The traceroute installed from this package is just a stub file that is not executable.  The same is true with whois.  That is why we have separate packages.  I'm guessing the maintainers did this so they would not interfere with the executables installed from the main package manager.  In our case, Chrome OS does not install these executables out-of-the-box.

jam7:
Is it possible to hear more about why you want to disable `traceroute`?  I'm not sure what you meant by `stub file`.  `traceroute` command itself is built by inetutils but +rx flags are disabled on previous `inetutils.rb`.

uberhacker:
The `traceroute` included in this package is an older version and currently doesn't install with execute permissions.  Since we already have a `traceroute` package (more updated version), hence why it was disabled.

jam7:
I see your point.  I'll check more detail later.  Thanks.

PR #895, opened 2017-07-16
Change install.sh to use the latest cross-compiled libssh2, git and ruby

jam7:
This PR updates libssh2, git and ruby binary packages used by install.sh to the latest one.

Tested on x86_64 and armv7l.  Tested i686 on clooudready.

All packages are cross-compiled on x86_64 ubuntu by using https://github.com/jam7/chrome-cross.


lyxell:
Great work!

Issue #889, opened 2017-07-16
GUI


janbaudisch:
This issue is intended to keep track of the progress made in the case of GUI applications. It is a result of #886.

- [x] libX11 (#930)
- [x] X-org protocols (#906)
- [x] Wayland (#961)
- [x] Mesa (#962)
- [x] udev / eudev (#1228)
- [x] XWayland (#1229)
- [x] GTK
- [x] Sommelier

smerkousdavid:
I think that's an amazing idea. Here are a few packages/dependencies for X11. 
https://github.com/CodeBuds/Crevos/tree/32c0200b6b2bc08152b31d1d952c58e0875748a6/packages.

Cheers

bfayers:
Super cool, once this is working I may revert one of my chromebooks to running chromeos again instead of Arch.

uberhacker:
How is this coming along @flyingP0tat0?  You making any progress?

flyingP0tat0:
I don't really have time to finish this project unfortunately.
Feel free to check out [my fork](https://github.com/flyingP0tat0/chromebrew), especially the test directory. There already are Weston and GTK scripts. I did not get these working, but it is possible.
Although I don't really have time at the moment, you can tag me everytime you want me to take a look at something.

uberhacker:
Yes, I understand.  This is a very ambitious project and I appreciate the effort you have put into it so far.  I'll see if I have time to nudge this along when I get a chance.
Thanks

littlej247:
The work you guys are doing is absolutely amazing. I know this isn't the place for this but I didn't know where else to say it. Thanks for all your hard work to everyone involved in chromebrew. Honestly Thank You!

nskaggs:
@flyingP0tat0, can you summarize what's left to make this a reality? Where exactly did you get stuck?

flyingP0tat0:
@nskaggs I guess the most important thing is to get GTK and Weston to work.

VitaliyKulikov:
ping me if need to test something on Google Pixelbook )) meantime, i will take a look on codebase.

beliefless-dogmasless:
I'm linking this:
https://www.reddit.com/r/unixporn/comments/7zns09/chromebook_wayland_going_full_native/

bfayers:
Whoah, cool!

tomas:
@uberhacker Do you think it would be possible to get something working, now that the gtk3 package has been merged?

[...]

PR #898, opened 2017-07-17
Add xorg protocols

flyingP0tat0:
Adds all available xorg protocls as a packages each + `xorg_proto` fake package to install all.
This is not helpful at the moment, but one step towards getting GUIs to work.
See #889.


flyingP0tat0:
Oops, some other commits went into this one...

PR #899, opened 2017-07-17
Add nix dependencies

flyingP0tat0:
These are (some) of [nix](https://github.com/NixOS/nix)' dependencies.
I did not manage to get nix itself working yet, but here are at least some new packages for now.

This includes updating perl to install cpanminus, which allows one to install perl modules easliy. Plus one perl module with a patch to make it working.


flyingP0tat0:
Removed perl `WWW:Curl` package as I found a better way to install it.

PR #901, opened 2017-07-17
Add dmidecode package

uberhacker:
Dmidecode reports information about your system's hardware as described in your system BIOS according to the SMBIOS/DMI standard (see a sample output). This information typically includes system manufacturer, model name, serial number, BIOS version, asset tag as well as a lot of other details of varying level of interest and reliability depending on the manufacturer. This will often include usage status for the CPU sockets, expansion slots (e.g. AGP, PCI, ISA) and memory module slots, and the list of I/O ports (e.g. serial, parallel, USB).  See http://www.nongnu.org/dmidecode/.

### DIFF
--- a/packages/dmidecode.rb
+++ b/packages/dmidecode.rb
@@ -8,6 +8,10 @@ class Dmidecode < Package
   source_sha256 'd766ce9b25548c59b1e7e930505b4cad9a7bb0b904a1a391fbb604d529781ac0'
 
   def self.build
+    case ARCH
+    when 'aarch64', 'armv7l'
+      abort "#{ARCH} architecture not supported.".lightred
+    end
     system 'make'
   end
 
--- a/packages/dmidecode.rb
+++ b/packages/dmidecode.rb
@@ -1,0 +1,21 @@
+require 'package'
+
+class Dmidecode < Package
+  description "Dmidecode reports information about your system's hardware as described in your system BIOS according to the SMBIOS/DMI standard (see a sample output)."
+  homepage 'http://www.nongnu.org/dmidecode/'
+  version '3.1'
+  source_url 'http://download.savannah.gnu.org/releases/dmidecode/dmidecode-3.1.tar.xz'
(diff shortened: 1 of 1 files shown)


jam7:
When I try this PR, it shows:
```
$ sudo /usr/local/sbin/dmidecode
# dmidecode 3.1
Scanning /dev/mem for entry point.
/dev/mem: Operation not permitted
/dev/mem: Bad file descriptor
```
I think there is a problem on installation into armv7l.

uberhacker:
@jam7: If you sudo a command and get errors, that is a permission issue instead of an installation issue.  Does the `/dev/mem` node exist on armv7l?

jam7:
It exists.
```
$ ls -l /dev/mem 
crw-r-----. 1 root kmem 1, 1 Jul 29 14:45 /dev/mem
```

jam7:
Ok.  I researched it little.  The answer is dmidecode doesn't work on arm.  I guess physical memory mapping to access BIOS information is different.

@uberhacker is it possible to add a comment like `# dmidecode doesn't support ARM atm` in dmidecode.rb?  Thanks.

uberhacker:
@jam7: Should we make a stronger effort to get it working on arm?  Are there some switch options that we could use to make it work?

uberhacker:
@jam7: I pushed up a new commit I think you will like.

jam7:
it's way too much.
```
$ crew install healthcheck
healthcheck: Linux system health check.
https://github.com/SimplyLinuxFAQ/health-check-script
version 1cace5
dmidecode 
Do you agree? [Y/n] y
Proceeding...
dmidecode: Dmidecode reports information about your system's hardware as described in your system BIOS according to the SMBIOS/DMI standard (see a sample output).
http://www.nongnu.org/dmidecode/
version 3.1
Archive downloaded
gcc (GCC) 4.9.4

armv7l architecture not supported.
```

uberhacker:
@jam7: I changed the abort line to puts instead.  This should allow healthcheck to be installed.  Keep in mind, you will see errors when running healthcheck on arm.

PR #920, opened 2017-07-20
Update cairo from 1.14.8 to 1.14.10

uberhacker:


jam7:
Can you add `fontconfig` as a dependency?  Cairo itself doesn't depend on fontconfig, but pango requires fontconfig enabled cairo.

uberhacker:
Done.

PR #905, opened 2017-07-18
Update SDL to SDL2

flyingP0tat0:
Updating to version 2.0.5

PR #929, opened 2017-07-21
Add SDL2 (libsdl2) and fix libtheora

flyingP0tat0:
Cleares up my mistake from #905. Also fixes `libtheora`, which caused an error before.

### DIFF
--- a/packages/libsdl2.rb
+++ b/packages/libsdl2.rb
@@ -1,0 +1,18 @@
+require 'package'
+
+class Libsdl2 < Package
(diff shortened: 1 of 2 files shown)


uberhacker:
Be careful about changing libtheora.  The ffmpeg package requires specific versions of libraries to compile correctly.  Please check that this doesn't break ffmpeg and other package builds.

flyingP0tat0:
That is exactly the reason why I changed it. I needed it to work in order to compile `ffmpeg`.

PR #906, opened 2017-07-18
Add xorg protocols

flyingP0tat0:
One step towards #889.

PR #930, opened 2017-07-21
Add libx11

flyingP0tat0:
Next step towards #889.

### DIFF
--- a/packages/libx11.rb
+++ b/packages/libx11.rb
@@ -1,0 +1,18 @@
+require 'package'
+
+class Libx11 < Package
+  description 'C interface to the X window system'
+  homepage 'https://x.org'
+  version '1.6.5'
+  source_url 'https://www.x.org/archive/individual/lib/libX11-1.6.5.tar.gz'
+  source_sha256 '3abce972ba62620611fab5b404dafb852da3da54e7c287831c30863011d28fb3'
+
+  def self.build
+    system "./configure"
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+end
--- a/packages/libx11.rb
+++ b/packages/libx11.rb
@@ -7,6 +7,13 @@ class Libx11 < Package
   source_url 'https://www.x.org/archive/individual/lib/libX11-1.6.5.tar.gz'
   source_sha256 '3abce972ba62620611fab5b404dafb852da3da54e7c287831c30863011d28fb3'
 
+  depends_on 'kbproto'
+  depends_on 'libxcb'
+  depends_on 'xproto'
+  depends_on 'inputproto'
+  depends_on 'xextproto'
+  depends_on 'libxtrans'
+  
   def self.build
     system "./configure"
     system "make"
--- a/packages/libx11.rb
+++ b/packages/libx11.rb
@@ -13,6 +13,8 @@ class Libx11 < Package
   depends_on 'inputproto'
   depends_on 'xextproto'
   depends_on 'libxtrans'
+  depends_on 'libxau'
+  depends_on 'pthread_stubs'
   
   def self.build
     system "./configure"
--- a/packages/libx11.rb
+++ b/packages/libx11.rb
@@ -13,8 +13,6 @@ class Libx11 < Package
   depends_on 'inputproto'
   depends_on 'xextproto'
   depends_on 'libxtrans'
-  depends_on 'libxau'
-  depends_on 'pthread_stubs'
   
   def self.build
     system "./configure"
--- a/packages/libxau.rb
+++ b/packages/libxau.rb
@@ -1,0 +1,18 @@
+require 'package'
+
+class Libxau < Package
+  description 'xau library for libX11'
+  homepage 'https://x.org'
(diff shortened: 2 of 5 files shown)


uberhacker:
👍

flyingP0tat0:
@uberhacker You mean for `libx11`?

uberhacker:
Yes, it requires a ton of dependencies.  It did at least when I tried compiling it awhile back.

flyingP0tat0:
Ok. I could compile it without anything. But maybe I just have the dependencies already installed. But I thought it is just a kind of interface nothing funky like displaying. Maybe someone else could try to compile it?

flyingP0tat0:
@uberhacker I found the dependecies from AUR. I will add them.

uberhacker:
Did you compile on CloudReady?  If you did, that library is already present.  That's how I could install qt.

flyingP0tat0:
No, ChromOS stable.

flyingP0tat0:
Now it should have all dependencies.

uberhacker:
And each one of the dependencies you added has a ton of dependencies also.  It's running down a rabbit hole.  You will soon find out. :)

flyingP0tat0:
But seriously, when I installed them, it just worked.
I will probably reset my Chromebook soon and then I can try again. Maybe someone else with a somewhat clean installation?

jam7:
`crew build libx11` shows following errors on my environment.  Crew was trying to compile libxcb.
```
checking if gcc -std=gnu99 supports -Werror... yes
checking if gcc -std=gnu99 supports -Werror=attributes... yes
Package xorg-macros was not found in the pkg-config search path.
Perhaps you should add the directory containing `xorg-macros.pc'
No package 'xorg-macros' found
checking whether make supports nested variables... (cached) yes
checking whether to build developer documentation... yes
checking for doxygen... no
configure: WARNING: doxygen not found - documentation targets will be skipped
configure: WARNING: dot not found - doxygen targets will be skipped
checking for CHECK... no
checking for XCBPROTO... yes
checking for NEEDED... no
configure: error: Package requirements (pthread-stubs xau >= 0.99.2) were not met:

No package 'pthread-stubs' found
No package 'xau' found

Alternatively, you may set the environment variables NEEDED_CFLAGS
and NEEDED_LIBS to avoid the need to call pkg-config.
libxcb failed to build: `./configure` exited with 1
```

flyingP0tat0:
@jam7 Thank you, I will work that out.

flyingP0tat0:
@jam7 These dependencies should now be fixed.

jam7:
Above errors are from libxcb.  So, it is required to update libxcb.rb dependencies at least.  I add dependes_on to libxcb.rb.  Then, I got no xproto erros while compiling libxau.

jam7:
After applying following patch, `crew install libx11` success.
```
diff --git a/packages/libxau.rb b/packages/libxau.rb
index dae03e8..b13b583 100644
--- a/packages/libxau.rb
+++ b/packages/libxau.rb
@@ -7,6 +7,8 @@ class Libxau < Package
   source_url 'https://www.x.org/archive/individual/lib/libXau-1.0.8.tar.gz'
   source_sha256 'c343b4ef66d66a6b3e0e27aa46b37ad5cab0f11a5c565eafb4a1c7590bc71d7b'
 
+  depends_on 'xproto'
+
   def self.build
     system "./configure"
     system "make"
diff --git a/packages/libxcb.rb b/packages/libxcb.rb
index 35e1683..52f9a5d 100644
--- a/packages/libxcb.rb
+++ b/packages/libxcb.rb
@@ -8,6 +8,8 @@ class Libxcb < Package
   source_sha256 '092f147149d8a6410647a848378aaae749304d5b73e028ccb8306aa8a9e26f06'
 
   depends_on 'xcb_proto'
+  depends_on 'libxau'
+  depends_on 'pthread_stubs'
 
   def self.build
     system "./configure"
```

flyingP0tat0:
@jam7 Now it should acutually be correct.

uberhacker:
The only way this will install on a fresh system is if dependency packages `kbproto`, `xproto`, `inputproto` and `xextproto` are also added in with this PR.  I suspect that each of these aforementioned packages requires several additional dependencies.
```
$ crew install libx11
libx11: C interface to the X window system
https://x.org
version 1.6.5
Package kbproto not found. :(
```

flyingP0tat0:
@uberhacker They are added in #906. They don't require any other dependencies, as they are just coping some files, no building at all.

uberhacker:
Ah, gotcha.  Sorry, missed that in the maze of PRs we have in the queue. LoL

yancouto:
Wow, I was a couple of days late. I was actually trying to see if I could install [love2d](https://love2d.org/), and ended up installing a bunch of other dependencies and related stuff (A LOT), including this.

In the end it didn't even work 😞. I installed xorg-server and got some weird errors that I couldn't get past. I'm not even sure if all this I'm installing is necessary. You seem to know what you're doing way better than me hahah.

I hope it all works out in the end. If you need any help testing the new packages, I can try them in my chromebook.

flyingP0tat0:
@yancouto Of course, trying and reporting is always great.

Issue #934, opened 2017-07-22
fuse and sshfs issues


jam7:
When I try to use sshfs, the setuid is not enabled.  So, I need to perform following steps.
```
$ sudo chown root:root /usr/local/bin/sshfs
$ sudo chmod u+s /usr/local/bin/sshfs
```
Then, I can mount my workstation through ssh as below but with warning messages.  Is there any way to fix this?
```
$ mkdir work
$ sshfs workstation:work work
fuse: warning: library too old, some operations may not not work
```

BTW, sshfs is really nice.  Thanks for it! 👍

jam7:
The source of the problem is version mismatch between /usr/lib/libfuse.so.2.8.6 and /usr/local/lib/libfuse.so.2.9.7.  We must use `/sbin/mount.fuse` which uses libfuse.so.2.8.6 and it is not overwritten, so it is required to use libfuse 2.8.6 to avoid conflicts.

jam7:
I'll update package files later.  I'm re-checking the behavior of sshfs without setuid.  It may work without it.

jam7:
Updated #935.  setuid is not required at all.

jam7:
close in favor of #968

PR #935, opened 2017-07-22
Update fuse to solve problem described in #934

jam7:
Update sshfs and fuse to solve problem described in #934.

Tested on armv7l, x86_64 and i686.

We may need to think about adding something like `postinstall` method to perform setuid and similar works.  We don't want to setuid before copying files from `/usr/local/tmp/crew` to `/usr/local` since it may intorduces too many root owned fiels.  Therefore, we removed automatic sudo stuff from `crew`.  But, how about doing setuid after copying files?  Maybe it's OK...


jam7:
I'm closing this one once to submit a new PR again.  When I wrote this PR, I misunderstood about sshfs and description about it confuse us.

PR #936, opened 2017-07-23
Update emacs from 25.1 to 25.3

uberhacker:
Added `--localstatedir=/usr/local/share` option to resolve:
```
tar: ./usr/local/var/games: Cannot mkdir: Permission denied
```


jam7:
I think it's better to solve the source of problems caused by cloudready instead of patching every packages show conflicts.

uberhacker:
@jam7:  What do you mean?  The issue is /var is symlinked to /usr/local/var.  Since /var is mounted read-only, this is what produces the error.  I simply changed the path to be writable.

jam7:
/usr/local/var symbolic link exist on only cloudready, doesn't it?  No such link on Chrome OS or Chromium OS.

uberhacker:
True, but we want it working in any environment, right?  This fix shouldn't do any harm and it solves the problem.
HP Chromebook x86_64
```
$ df /usr/local/var
Filesystem     1K-blocks    Used Available Use% Mounted on
/dev/mmcblk0p1  10801712 8367444   1865848  82% /usr/local
```
CloudReady x86_64
```
```

jam7:
that's right. your modification doesn't hurts.  but you need to modify all packages using /usr/local/var.

how about `sudo chown -R chronos:chronos /usr/local/var` in install.sh if it exists?

jam7:
you may need to specfy /usr/local/var/.

uberhacker:
#953 is related.

jam7:
Please stop to use `/usr/local/var` for something like below by your own reason.
```
‘localstatedir’
The directory for installing data files which the programs modify while they run, and that pertain to one specific machine. Users should never need to modify files in this directory to configure the package’s operation; put such configuration information in separate files that go in $(datadir) or $(sysconfdir). $(localstatedir) should normally be /usr/local/var, but write it as $(prefix)/var. (If you are using Autoconf, write it as ‘@localstatedir@’.)
```

thedamian:
@jam7 I thnk statements like "PLEASE STOP USING" sound like you're the dictator here. and nor you nor I fall under than even if you and uberhacker are the top contributors. 

You guys make this an AMAZING repo. Please treat each other as equals and since we're all rowing the same boat we might get to land faster

jam7:
I treated he as equals.  I suggested different solution in #953, removing a symbolic link from `/usr/local/var` to `/var`.  He is just pushing his PR again and again without explanation.  :(

Anyway, it's a good time to discuss about it.  I'll make an issue.

cstrouse:

cstrouse:
emacs rev'd to 25.3

uberhacker:
@cstrouse: Updated to 25.3, tested and working on x86_64.  Ready to merge after ARMv8 testing.

cstrouse:
Works properly on Samsung Chromebook Plus (aarch64).

PR #942, opened 2017-07-23
Add libatomic_ops package

uberhacker:
The atomic_ops project (Atomic memory update operations portable implementation).  See https://github.com/ivmai/libatomic_ops.


lyxell:
Merged in #899

PR #953, opened 2017-07-23
Fix grabinfo.lock error

uberhacker:
Resolves:
```
Can't create lock file (/usr/local/share/grabinfo/grabinfo.lock): No such file or directory at /usr/local/bin/grabinfo line 61.
```

### DIFF
--- a/packages/grabinfo.rb
+++ b/packages/grabinfo.rb
@@ -3,30 +3,30 @@ require 'package'
 class Grabinfo < Package
   description 'Navigates the /proc filesystem and executes commands in order to gather information about the state of your machine.'
   homepage 'https://sourceforge.net/projects/grabinfo/'
-  version '1.71'
+  version '1.71-1'
   source_url 'https://downloads.sourceforge.net/project/grabinfo/grabinfo/1.71/grabinfo-1.71.tar.gz'
   source_sha256 '3cf1e56aff10ef182065943a10ab815f2e8bf5ab77716dd97c2e91feb8db4278'
 
   depends_on 'perl'
 
   def self.build
-    system "sed -i 's,/usr/bin,/usr/local/bin,g' grabinfo"
-    system "sed -i 's,/var/log,/usr/local/tmp,g' grabinfo"
-    system "sed -i 's,/var/log,/usr/local/tmp,g' grabinfo.1"
-    system "sed -i 's,/usr/lib,/usr/local/lib,g' Makefile"
-    system "sed -i 's,/usr/share,/usr/local/share,g' Makefile"
-    system "sed -i 's,/var/log,/usr/local/tmp,g' Makefile"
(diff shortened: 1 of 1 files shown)


jam7:
Why is this program using `/usr/local/share/grabinfo/grabinfo.lock` files instead of using `/usr/local/var/grabinfo/grabinfo.lock`?

uberhacker:
@jam7: The `/usr/local/var` "folder" is a symlink to `/var` on CloudReady.  Since this directory is not writeable, hence the change.  We could write to `/usr/local/tmp/grabinfo` as an alternative.

jam7:
Regarding to solve your CloudReady only problem, there are several ways to solve it.
1. change every programs you noticed to use different path.
1. perform `sudo chown -R chronos:chronos /usr/local/var/.` at install time.
1. perform `sudo rm /usr/local/var` at compile time.

1st choice affects your modification to all other's environment.  2nd choice would be the most safe.  3rd choice is the most simple, I used to use this option every time I use CloudReady or Chromium OS.

uberhacker:
I'm not so sure #2 would be the safest.  It might be possible that certain programs running under a user other than chronos need write access to `/var`.  If the process(es) are running under root then it shouldn't be a problem.  Since this only affects CloudReady, #3 seems the most feasible.

jam7:
@uberhacker I agree with you.  Then, what will you do?  Just want to know your opinion.

If you would like to use 3rd choice above, is it possible to perform follwing?  Thanks.

1. Submit new a PR to modiify install.sh for ChromeReady.
2. Modify this PR to use /usr/local/var

jam7:
BTW, you are right.  `chown -R chronos:chronos /var` caused several problems.  For example, `/var/db/sudo` must be owned by root.  ;)

uberhacker:
@jam7: I will work on this after work this evening.

uberhacker:
The more I think about this, the better I like option 1 since removing the symlink might break CloudReady.  This issue happens rarely anyways and changing each package on a case by case basis is better, I think.  I have pushed another commit to store the lock file in `/usr/local/tmp/grabinfo` instead which probably makes more sense.

jam7:
The `tmp` is better than `share`, so I'm merging this one, but I appreciate if you sotp to not use var as I explained above.

jam7:
ah.. please increment version number before the merge.  thanks.

uberhacker:
I have updated the PR.  Please check when you get a chance.  Thanks

thedamian:
@uberhacker @jam7  This looks like a great compromise. I didn't like the idea of programming for anything other than chromeos (not chromium, nor cloudready) 

Until I saw that a LOT of schools are converting their old laptops to use it.

I love  compromises and will merge it now.

LOVE bug fixes!

Issue #956, opened 2017-07-25
Cyclic dependencies


yancouto:
Some packages have cyclic dependencies.
[Perl](https://github.com/skycocker/chromebrew/blob/master/packages/perl.rb) depends on [curl](https://github.com/skycocker/chromebrew/blob/master/packages/curl.rb) that depends on [openssl](https://github.com/skycocker/chromebrew/blob/master/packages/openssl.rb) that depends on perl again.

When I cleaned up my chromebook and tried installing everything again I ran into a stack overflow problem when generating dependencies that happened because of this. I think this originated in commit 164aa4aa912a56e3c2cd842a567a010ff6909bc6 .

Deleting the curl dependency from perl worked, everything installed correctly. (Probably because of [this](https://github.com/skycocker/chromebrew/issues/882#issuecomment-316898564))

(There should be a way to test if package dependencies are correct, like uninstalling (or just moving around) everything else but the dependencies. I tried doing that quick and dirty but ended up deleting packages that were necessary for chromebrew. This should probably be another issue, though)

flyingP0tat0:
Indeed, there should be no problem in removing cURL as a dependency.

PR #957, opened 2017-07-25
Removing curl dependency from perl

yancouto:
This should close issue #956, and will work because curl is already installed by default.


flyingP0tat0:
@lyxell This is a quite important fix (breaks git upgrade for example), please do it as soon as soon as possible.
This was somewhat my own fault, sorry for that.

PR #958, opened 2017-07-25
Adding manpages package

yancouto:
I wasn't sure if I should put `mandb` as a dependency, so I decided to not do it and put a friendly text for the following reasons:
1. It really doesn't need `mandb` to be installed, just used.
1. If someone installed this package directly, the helper text in the end of the `mandb` installation (that shows how to configure it) would be lost in the installation text.

Feel free to change it, though.

### DIFF
--- a/packages/manpages.rb
+++ b/packages/manpages.rb
@@ -1,0 +1,14 @@
+require 'package'
+
+class Manpages < Package
+  description 'The Linux man-pages project documents the Linux kernel and C library interfaces that are employed by user-space programs.'
+  homepage 'https://www.kernel.org/doc/man-pages/'
+  version '4.12'
+  source_url 'https://www.kernel.org/pub/linux/docs/man-pages/man-pages-4.12.tar.xz'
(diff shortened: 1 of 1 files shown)


uberhacker:
@yancouto: Do we need this package?  I can read manpages just fine with only mandb?

yancouto:
Really? Weird. I installed mandb (and followed the given instructions) and wasn't able to access the manpages for C functions.

I cleaned up my chromebook today so there shouldn't be much interference. To be sure, I'll uninstall this and mandb and install it again and see what happens.

yancouto:
Nope. I get "no manual entry for XXX" when trying C functions like printf, strlen, ...

flyingP0tat0:
Isn't that expected, as mandb is just the database and man-pages  specifically provides some documentation?

jam7:
Nice.  This PR installs man pages for C functions and more, man2, man3, man4, man5, man6 and man7.  I'm going to merge this one if there is no objection.

@uberhacker you need to install both, mandb and manpages.

PR #961, opened 2017-07-26
Add Wayland

flyingP0tat0:
Next step towards #889.
Adds Wayland library and protocols + fake package to install both.

### DIFF
--- a/packages/libwayland.rb
+++ b/packages/libwayland.rb
@@ -1,0 +1,22 @@
+require 'package'
+
+class Libwayland < Package
+  description 'Wayland is intended as a simpler replacement for X, easier to develop and maintain.'
+  homepage 'https://wayland.freedesktop.org'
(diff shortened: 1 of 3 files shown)


jam7:
`crew install wayland_protocols` shows following error.
```
checking for WAYLAND_SCANNER... no
configure: error: Package requirements (wayland-scanner) were not met:

No package 'wayland-scanner' found

Alternatively, you may set the environment variables WAYLAND_SCANNER_CFLAGS
and WAYLAND_SCANNER_LIBS to avoid the need to call pkg-config.
wayland_protocols failed to install: `./configure` exited with 1
```
Once I install `libwayland` by hand, it is possible to compile.

flyingP0tat0:
@jam7 Yes, this is included in `libwayland`. Should be fixed now.

PR #963, opened 2017-07-26
Test for dependency cycle

yancouto:
I made a test to check if there is some dependency cycle. Running the program will show a dependency cycle or tell if there is none.

I'm new to ruby, and don't know much about this project, so it's probably not the prettiest code. I'll be happy to close this PR in favor of a better one. 

Things worth discussing
- Should tests be in a `tests/` folder?
- I just used a regular ruby program that returns 1 on error. Is there a test library that would be better?
- Should tests be automated? (Travis?)
- `load '../crew'` is probably not the best way to set up the environment correctly.

(Maybe this discussion should be moved to an issue?)

### DIFF
--- a/crew
+++ b/crew
@@ -5,8 +5,11 @@ require 'uri'
 require 'digest/sha2'
 require 'json'
 require 'fileutils'
+require_relative 'lib/color'
+require_relative 'lib/const'
 
-# Constants definitions
+# Add lib to LOAD_PATH
+$LOAD_PATH.unshift "#{CREW_LIB_PATH}lib"
 
 DOC = <<DOCOPT
 Chromebrew - Package manager for Chrome OS http://skycocker.github.io/chromebrew/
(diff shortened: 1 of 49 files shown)


jam7:
I think it's good idea to add such tools to make maintain chromebrew easy.  However, I have no big picture about these tools.  Maybe we should discuss about it once by making issue about it.

On the other hand, about this PR, please give me few more days to review it.  Thanks.

jam7:
It's really difficult to use currently.  It performs `load ../crew` so it is required to run script from `/usr/local/lib/crew/tests`.  One way to solve this is merging this ability into `crew` itself like making a function to check cyclic dependency path.  Or you can performs `load /usr/local/lib/crew/crew`.  Or we should gather all constants to `/usr/local/lib/crew/lib/config.rb` or something.

Having tests folder is good idea to gather several tests I think by the way.

yancouto:
It doesn't need to be run from `/usr/local/lib/crew`, it can be run from your local repository, it will just use the "local" crew instead of the one in `/usr/local`.

Having constants in another file or calling the tests from inside crew does sound better though.

yancouto:
How about this? I made a `crew test` command. It just runs all files in the `tests` folder.

Please don't merge it yet, it is just an idea. And if it is decided to do this, the install file needs to be changed to download the `tests` folder and something needs to be done about the crews that were already installed. (maybe moving the tests to `lib/tests` since `lib` is downloaded, but this doesn't make much sense)

yancouto:
Any input on this? Positive or negative

jam7:
Sorry I didn't check this one.

I have a question.  What is the purpose of this program?  If it is to test in CI process, it is better to not use crew since it cause following errors.  Using crew requires everything installed correctly.  I prefer to use more handy tool for CI process.
```
$ ./crew test      # I'm running this on linux server like what we will do in CI process
./crew:124:in `read': No such file or directory @ rb_sysopen - /usr/local/etc/crew/device.json (Errno::ENOENT)
        from ./crew:124:in `<main>'
```
If you think this is for handy check on Chrome OS, then using crew is not a problem.  I'll check more details.

thedamian:
I love the idea of this PR. I'm reading what others have said and are waiting on more discussion before the merge, but I've thought of setting up a test on every new and changed package at least so that I'd known if sily typos are at least taken care of without installing all packages.

I'll wait a little while long for the merge.

yancouto:
Oops. I had messed up the merge and had a lot of useless commits, so I forced everything back to normal.

@jam7, as you said, it would be good to be able to run the tests (at least this one) without crew installed, so I made the test look on the local packages (inside the same repository) instead of the ones in `/usr/local/lib/crew`.

I also removed the testing from `crew`. Now, to run all tests, you need to run the `tests/test_all` file (from within that directory). I added a `README. md` to the `tests` directory explaining everything.

I had to change the `crew` file a little to allow it to set up the constants and not actually do anything, since creating another (common) file would be a problem because of the way the sparse repository is set up.

**tl;dr**
```
$ cd tests
$ ./test_all
Running tests...

Test Name: no_cycles
No dependency cycles found.

All tests were successful.
```

yancouto:
Note: If we create a `.travis.yml` with this content
```yml
language: ruby
script:
  - cd tests
  - ruby test_all
```
and enable the repository on [travis-ci.org](travis-ci.org) (not sure if it can be any collaborator or just the owner) then it will test if every commit passes all tests. (I think it also tests pull requests even before they are merged)

jam7:
It looks perfect to use this test in CI tools.  I'll try it later.

Regarding to your questions, only owner can integrate those CI tools.  If one of them are integrated, it is possible to apply those CI tools to pull requests as well.

jam7:
You still modified `crew` command.

jam7:
I see.  You decide to modify crew to make `load 'crew'` passes.

It's much better to separate constant definitions and colorize functions to other files like `lib/const.rb`  and `lib/color.rb`.  Then, load them from your test script.  If you are not sure how to do that, I can submit a PR to do that.

yancouto:
Ok. I moved the constants and color to lib.
I also made so that test_all can be run from any directory (previously it only worked if it was run from tests/ folder).

I tested in a computer without crew installed and the tests run fine.

crew
```diff
@@ -84,80 +49,6 @@ end
...
-  def lightcyan
-    colorize(36, 1)
-  end
-
-  def white
-    colorize(37, 1)
-  end
-end
-
-#disallow sudo
```
jam7: You removed sudo check also.  Please revert this part.  Thanks.

jam7: Never mind, you moved sudo check to different place.  I see.

yancouto: I moved it before the docopt check. Sorry I didn't say anything. Before, if you wrote a wrong command as sudo it would say the command was wrong instead of complaining you're sudo.

If you prefer the old way I can revert it.

jam7:
Thanks.  LGTM.  I appreciate to hear what other collaborator thinks since this PR requires modification of code structures (separate crew to `crew, lib/const.rb and lib/color.rb`).  I prefer this modifications, but I want to hear what others think.

And I appreciate other collaborator merges #1153 before this one since it will show a lot of conflicts...

tests/no_cycles.rb
```diff
@@ -0,0 +1,50 @@
+require 'find'
```
uberhacker: Do you need `require_relative '../lib/color'` since you colorized the last line?

tests/no_cycles.rb
```diff
@@ -0,0 +1,50 @@
...
+# Looking for cycles. @path will keep the current dependency path.
+# @state will store :on_path for vertices on the current dependency path
+# and :visited for vertices that have already been checked not to lead to
+# cycles.
+@state = {}
+@path = []
+def dfs(pkg)
+  @path.push(pkg.name)
+  if @state[pkg] == :on_path
+    puts "Found dependency cycle!".lightred
```
uberhacker: Should this be abort instead of puts?  Otherwise, it continues on and you get the success message at the end, correct?

uberhacker: Never mind.  I see you have the exit 1 at the end.

uberhacker:
@yancouto: This is a really awesome PR.  I like the refactoring.  I'm just wondering if it should be two separate PRs.  1 for the refactoring and another for the new tests.  Makes it more incremental and easier to evaluate.

uberhacker:
Actually, we probably should split it up into separate PRs anyway since it will create a merge conflict with #1153.  Here's hoping the other co-maintainer comes out of his cave and merges it. :)

tests/no_cycles.rb
yancouto: Well, `../lib/color` is required by the `test_all` script, so it isn't necessary here. But I'm not sure if I should rely on that. I guess it's better to be sure, as the `test_all` script may change in the future.

tests/no_cycles.rb
yancouto: Yes, I still need to print the dependency cycle.

tests/no_cycles.rb
yancouto: Added.

PR #975, opened 2017-07-29
Update heroku from 4.28.3 to 6.13.9

### LINKED ISSUE #829: Should we remove the heroku package?
This package is currently broken.  Heroku CLI (see https://devcenter.heroku.com/articles/heroku-cli) can be installed via `npm -g install heroku-cli`.  The legacy version can also be installed via `gem install heroku` (see https://blog.heroku.com/the_heroku_toolbelt).

uberhacker:
Resolves #829.   Depends on #1027.

### DIFF
--- a/packages/heroku.rb
+++ b/packages/heroku.rb
@@ -3,9 +3,9 @@ require 'package'
 class Heroku < Package
   description 'The Heroku Command Line Interface (CLI), formerly known as the Heroku Toolbelt, is a tool for creating and managing Heroku apps from the command line / shell of various operating systems.'
   homepage 'https://devcenter.heroku.com/articles/heroku-cli'
-  version '6.13.5'
-  source_url 'https://github.com/heroku/cli/archive/v6.13.5.tar.gz'
-  source_sha256 '769943a2e272f9b9332d85cfd5d4f49a7a3070a06b81e6da1cb33b4723a85607'
+  version '6.13.6'
+  source_url 'https://github.com/heroku/cli/archive/v6.13.6.tar.gz'
+  source_sha256 '395c7e22d16b1fce2d5ed4ac1e022dbd3c9aa8fdbc172d9dadf86ba9c1ad6978'
 
   depends_on 'nvm'
   depends_on 'yarn'
--- a/packages/heroku.rb
+++ b/packages/heroku.rb
@@ -7,11 +7,9 @@ class Heroku < Package
   source_url 'https://github.com/heroku/cli/archive/v6.13.9.tar.gz'
   source_sha256 '76699029d86fc10326f8cf39b0c0a3b1f322dee39bd7c96fa81e34e52ff454f9'
 
-  depends_on 'nodebrew' unless `node --version`
   depends_on 'yarn'
 
   def self.build
-    system 'nodebrew install-binary latest && nodebrew use latest' unless `node --version`
     system 'yarn install'
   end
 
--- a/packages/heroku.rb
+++ b/packages/heroku.rb
@@ -3,15 +3,26 @@ require 'package'
 class Heroku < Package
   description 'The Heroku Command Line Interface (CLI), formerly known as the Heroku Toolbelt, is a tool for creating and managing Heroku apps from the command line / shell of various operating systems.'
   homepage 'https://devcenter.heroku.com/articles/heroku-cli'
-  version '4.28.3'
-  binary_url ({
-    armv7l: "https://drive.google.com/uc?export=download&id=0ByCixsDmZPzxOHFIMzQyNHNFUWc",
-    i686: "https://drive.google.com/uc?export=download&id=0ByCixsDmZPzxd3NULTRkMWlHQTA",
-    x86_64: "https://drive.google.com/uc?export=download&id=0ByCixsDmZPzxLURkMktpREpDZk0"
-  })
-  binary_sha256 ({
-    armv7l: "b7fab53d3c8cfd42d41c1b07db069c9fd5f7261fcea48fd99114981cf1a293d7",
-    i686: "1e73fbff3a10c0c95dd1aa1cf68952a95a330634fa410765962df0396315a42c",
-    x86_64: "9259cd8b5d7aaeb9172142b5956ad767a3d4f1bcf126e51a7f4a8c055e53c068"
-  })
+  version '6.13.5'
+  source_url 'https://github.com/heroku/cli/archive/v6.13.5.tar.gz'
+  source_sha256 '769943a2e272f9b9332d85cfd5d4f49a7a3070a06b81e6da1cb33b4723a85607'
+
+  def self.install
+    case ARCH
+    when 'i686'
+      system 'wget https://cli-assets.heroku.com/heroku-cli/channels/stable/heroku-cli-linux-x86.tar.gz -O heroku-cli.tar.gz'
+    when 'x86_64'
+      system 'wget https://cli-assets.heroku.com/heroku-cli/channels/stable/heroku-cli-linux-x64.tar.gz -O heroku-cli.tar.gz'
+    else
+      abort 'Architecture not supported.'.lightred
+    end
+    system "mkdir -p #{CREW_DEST_DIR}/usr/local/bin"
+    system "mkdir -p #{CREW_DEST_DIR}/usr/local/lib"
+    system "tar xvf heroku-cli.tar.gz"
+    system "mv heroku-cli-v* heroku-cli-linux-#{ARCH}"
+    system "mv heroku-cli-linux-#{ARCH} #{CREW_DEST_DIR}/usr/local/lib"
+    FileUtils.cd("#{CREW_DEST_DIR}/usr/local/bin") do
+      system "ln -s /usr/local/lib/heroku-cli-linux-#{ARCH}/bin/heroku heroku"
+    end
+  end
 end
--- a/packages/heroku.rb
+++ b/packages/heroku.rb
@@ -7,22 +7,19 @@ class Heroku < Package
   source_url 'https://github.com/heroku/cli/archive/v6.13.5.tar.gz'
   source_sha256 '769943a2e272f9b9332d85cfd5d4f49a7a3070a06b81e6da1cb33b4723a85607'
 
+  depends_on 'yarn'
+
+  def self.build
+    system '. ~/.bashrc && nvm install stable'
+    system 'yarn install'
+  end
+
   def self.install
-    case ARCH
-    when 'i686'
-      system 'wget https://cli-assets.heroku.com/heroku-cli/channels/stable/heroku-cli-linux-x86.tar.gz -O heroku-cli.tar.gz'
-    when 'x86_64'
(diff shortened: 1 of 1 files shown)


jam7:
Is it difficult to support armv7l?  This is just a question.  If it is difficult, I'll merge this one as is after some evaluation.

uberhacker:
Unfortunately, they only have binaries available for i686 and x86_64.  I'll try and compile from source.

uberhacker:
@jam7: Please test the latest commit on arm when you get a chance.  It works on x86_64.

jam7:
Thank you for your effort, I can compile it on arm well.  But it shows following error on other machine.
```
Archive downloaded                                                          
.: 1: Can't open /home/chronos/user/.bashrc                                                                                           
heroku failed to build: `\. ~/.bashrc && nvm install stable && yarn install` exited with 127
```
How about updating node to 8.2.1 and use something like below you used in yarn.rb.
```
  depends_on 'node' unless File.exists? '/usr/local/bin/node'
```
Chromebrew has too many ways to install node atm.  :(

uberhacker:
Yes, node is everywhere these days.  Do I need `depends_on 'node'` here if it's already in `yarn`?  Maybe we can put `depends_on 'nvm'` in `yarn`. ;)

packages/heroku.rb
```diff
@@ -3,15 +3,27 @@
...
-  binary_sha256 ({
-    armv7l: "b7fab53d3c8cfd42d41c1b07db069c9fd5f7261fcea48fd99114981cf1a293d7",
-    i686: "1e73fbff3a10c0c95dd1aa1cf68952a95a330634fa410765962df0396315a42c",
-    x86_64: "9259cd8b5d7aaeb9172142b5956ad767a3d4f1bcf126e51a7f4a8c055e53c068"
-  })
+  version '6.13.6'
+  source_url 'https://github.com/heroku/cli/archive/v6.13.6.tar.gz'
+  source_sha256 '395c7e22d16b1fce2d5ed4ac1e022dbd3c9aa8fdbc172d9dadf86ba9c1ad6978'
+
+  depends_on 'nvm'
```
jam7: What we need here is node 8.1.4 or above.  Not nvm.  I prefer the way you did in yarn.rb like below.
```
  depends_on 'node' unless File.exists? '/usr/local/bin/node'                                                                         
```
Then, user can choose nvm to install node 8.1.4 or above.

packages/heroku.rb
```diff
@@ -3,15 +3,27 @@
...
+  def self.install
+    system "mkdir -p #{CREW_DEST_DIR}#{CREW_PREFIX}/bin"
+    system "mkdir -p #{CREW_DEST_DIR}#{CREW_PREFIX}/share/heroku"
+    system "cp -r . #{CREW_DEST_DIR}#{CREW_PREFIX}/share/heroku"
+    FileUtils.cd("#{CREW_DEST_DIR}#{CREW_PREFIX}/bin") do
+      system "ln -s /usr/local/share/heroku/bin/run heroku"
+    end
+    puts ""
+    puts "To finish the installation, execute the following:".lightblue
+    puts "source ~/.bashrc".lightblue
```
jam7: Do these message really needed?

packages/heroku.rb
```diff
@@ -3,15 +3,27 @@
...
-  })
+  version '6.13.6'
+  source_url 'https://github.com/heroku/cli/archive/v6.13.6.tar.gz'
+  source_sha256 '395c7e22d16b1fce2d5ed4ac1e022dbd3c9aa8fdbc172d9dadf86ba9c1ad6978'
+
+  depends_on 'nvm'
+  depends_on 'yarn'
+
+  def self.build
+    system '\. ~/.bashrc && nvm install stable && yarn install'
```
jam7: This should be just `yarn install`.

packages/heroku.rb
uberhacker: We need nvm to install node stable (aka version >= 8.1.4).  The message is needed, otherwise heroku is not found.  See below:
```
To finish the installation, execute the following:
source ~/.bashrc

Installing...
Heroku installed!
chronos@localhost ~/chromebrew/packages (update-heroku-package) $ heroku
/usr/local/bin/heroku: line 19: node: command not found
chronos@localhost ~/chromebrew/packages (update-heroku-package) $ source ~/.bashrc
chronos@localhost ~/chromebrew/packages (update-heroku-package) $ heroku
Usage: heroku COMMAND

Help topics, type heroku help TOPIC for more details:

 access          manage user access to apps
 addons          tools and services for developing, extending, and operating your app
 apps            manage apps
 auth            heroku authentication
...
```

jam7: It's the requirement of nvm not heroku.

uberhacker: If nvm is not installed initially, this message is necessary.  However, it doesn't hurt a thing to execute it again just in case.

jam7: Installing `nvm` is not necessary if node v8.1.4 or above is isntealled already.  In this PR, you assume many ppl don't install new node yet.  It may correct atm.  But, how about 3 months later?

Regarding to make `heroku` depend on `nvm`, there are many other version management tools for node like nodebrew, nbenv, anyenv, ...  I'm still not sure what is the best way to solve this situation.  But, choosing only one version management tool is not the best way I think...

jam7: On the other hand, following method you used in yarn.rb allows users to choose users' own method to install node.  I think this is great idea.  To use this with nvm, we need to modify `nvm.rb` to make a symbolic link from `/usr/local/bin/node` to `~/.nvm/.../node`, though.  
```
depends_on 'node' unless File.exists? '/usr/local/bin/node'
```

uberhacker: This whole exercise would be a lot easier if we just got rid of the `node` and `node_current` packages. :)  I would ultimately like to do something like the following:
```
depends_on 'insert best node version manager' unless "$(node -v)" >= "v8.1.4"
```

jam7: Here is another issue for me.  What chromebrew and other package management tools do is to suggest dependencies.  For example, we suggest to install `node` if it doesn't exist.  It's not needed to force to use the latest node.  We can suggest to install new node if it doesn't exist.  But, choosing `nvm` over `node` is too much to do in my opinion.  Maybe we can suggest to use nvm, but installing it is way too much.

If the installed node is too old, heroku installation shows it's own errors.  It's fine.  With those errors, each user can choose what each want.

packages/heroku.rb
jam7: Please change this line something like:
```
  depends_on 'node' unless `node --version`
```
Then, users can choose whatever they want.  For example, `node`, `node_current`, `nvm` or `nodebrew`.

packages/heroku.rb
```diff
@@ -3,15 +3,27 @@
...
-  })
+  version '6.13.6'
+  source_url 'https://github.com/heroku/cli/archive/v6.13.6.tar.gz'
+  source_sha256 '395c7e22d16b1fce2d5ed4ac1e022dbd3c9aa8fdbc172d9dadf86ba9c1ad6978'
+
+  depends_on 'nvm'
+  depends_on 'yarn'
+
+  def self.build
+    system '. ~/.bashrc && nvm install stable && yarn install'
```
jam7: Please modify this something like:
```
    system 'yarn install'
```
We don't need to configure `nvm` here.

packages/heroku.rb
```diff
@@ -3,15 +3,27 @@
...
+  end
+
+  def self.install
+    system "mkdir -p #{CREW_DEST_DIR}#{CREW_PREFIX}/bin"
+    system "mkdir -p #{CREW_DEST_DIR}#{CREW_PREFIX}/share/heroku"
+    system "cp -r . #{CREW_DEST_DIR}#{CREW_PREFIX}/share/heroku"
+    FileUtils.cd("#{CREW_DEST_DIR}#{CREW_PREFIX}/bin") do
+      system "ln -s /usr/local/share/heroku/bin/run heroku"
+    end
```
jam7: Please use following command lines here:
```
    system "mkdir -p #{CREW_DEST_DIR}#{CREW_LIB_PREFIX}/heroku"
    system "cp -r . #{CREW_DEST_DIR}#{CREW_LIB_PREFIX}/heroku"
    system "mkdir -p #{CREW_DEST_DIR}#{CREW_PREFIX}/bin"
    system "ln -s #{CREW_LIB_PREFIX}/heroku/bin/run #{CREW_DEST_DIR}#{CREW_PREFIX}/bin/heroku"
```
Point is using lib directory instead of share directory.  Basically, share directory is used to contain files we can share among multiple architectures.  It's the reason of that perl, python, ruby, etc. place their directory at lib.

packages/heroku.rb
```diff
@@ -3,15 +3,27 @@
...
+    puts ""
+    puts "To finish the installation, execute the following:".lightblue
+    puts "source ~/.bashrc".lightblue
+    puts ""
```
jam7: Please remove these puts lines to suggest to configure nvm.

uberhacker: Although I agree we shouldn't force the user to install node in any particular manner, there are several issues with this line:
```
depends_on 'node' unless `node --version`
```
1. We need node version 8 and `node` installs version 6.  We should instead determine what version is currently installed and suggest the user install (or better yet, install automatically) a more updated version that meets the minimum requirement.
2. `node --version` does not work when `nvm` is installed unless `~/.bashrc` is sourced first.
3. Compiling node from source takes an eternity on a slow Chromebook (like over 2 hours).

I would recommend we use nvm or nodebrew instead of compiling from source.  I'd like to update the `yarn` package to install this way eventually also.

jam7: 1. I understand your point but determining the version number is really difficult.  In order to determine version number, we need some kind of framework.  Many OSSes just compre them as a string and cause a log bugs later when the version number bumps from 9 to 10 since string comparison answers 10 is smaller than 9.  Even if there is no version check in chromebrew, `heroku` itself checks the correct version.  So, I just think we don't need to check the version here.
2. It is user's problem who didn't modify their environment after the installation of `nvm`.
3. Yes.  We can provides pre-compiled node here.  Or, we can rename current `node.rb` to `node_from_source.rb` and make new `node.rb` to choose one of `node` version manager.  But, it is different topic I think.

jam7: BTW, these are what I think.  You can submit any PR and I'll merge them if it works fine and allows user's choice (e.g. `nvm` or `node`).

uberhacker:
I have updated the PR.  Please merge this soon before I run out of energy. :-P

uberhacker:
@jam7: I removed the nodebrew dependency here since yarn has the dependency already.  I also updated yarn to use nodebrew.

jam7:
In my opinion, you are exhausting yourself by merging several problems into one and making the problem bigger.  You don't need to solve heroku problem and selecting the best way to install `node` problem at once.  It's a piece of pie if you divide the problem into several problems like this.

jam7:
Reminder: it is required to install the latest node using `nvm`, #1000 or `nodebrew` with #1027 before trying to install this heroku.  Thanks.

PR #977, opened 2017-07-29
Update httpd from 2.2.32 to 2.4.27

uberhacker:


jam7:
is it difficult to change this to use apr and apr-util package previously installed instead of getting those source codes?

uberhacker:
It won't compile with `apr` and `aprutil` packages instead.  Needs the source code.  You are welcome to give it a try.

jam7:
As described in https://github.com/apache/httpd/blob/trunk/INSTALL, it is possible.  However, it is not possible if we use github tar.gz file which is not the official release file.  It doesn't contain `configure` file.

Please change it to use:
```
  source_url 'http://www-us.apache.org/dist/httpd/httpd-2.4.27.tar.gz'
  source_sha256 '346dd3d016ae5d7101016e68805150bdce9040a8d246c289aa70e68a7cd86b66'
```
and remove `subversion` dependency and add `apr`, `aprutil` and `expat` depdendencies.  Then, modify it to perform only `./configure` and `make`.  This will solve our problem.  Thanks.

uberhacker:
@jam7: Good observation.  I have submitted the changes as you suggested.  Really appreciate your efforts.  Chromebrew will continue to improve if you keep it up.

packages/httpd.rb
```diff
@@ -3,21 +3,22 @@
...
-    system './buildconf'
-    system './configure --prefix=/usr/local'
+    system './configure'
     system 'make'
   end
 
   def self.install
-    system 'make -i install'
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "libtool --finish #{CREW_DEST_DIR}#{CREW_PREFIX}/apache2/modules"
```
jam7: This last line shows:
```
libtool: finish: PATH="/usr/local/bin:/usr/bin:/bin:/opt/bin:/sbin" ldconfig -n /usr/local/tmp/crew/dest/usr/local/apache2/modules
----------------------------------------------------------------------
Libraries have been installed in:
   /usr/local/tmp/crew/dest/usr/local/apache2/modules

If you ever happen to want to link against installed libraries
in a given directory, LIBDIR, you must either use libtool, and
specify the full pathname of the library, or use the '-LLIBDIR'
flag during linking and do at least one of the following:
   - add LIBDIR to the 'LD_LIBRARY_PATH' environment variable
     during execution
   - add LIBDIR to the 'LD_RUN_PATH' environment variable
     during linking
   - use the '-Wl,-rpath -Wl,LIBDIR' linker flag
   - have your system administrator add LIBDIR to '/etc/ld.so.conf'

See any operating system documentation about shared libraries for
more information, such as the ld(1) and ld.so(8) manual pages.
----------------------------------------------------------------------
```
I'm just wondering we need this information?  I'm not sure how to use libtool, so this information may be correct and important.  I'm asking this question since I don't know about it.  Thanks.

uberhacker: I just followed the instructions from the message during compilation:
```
libtool: warning: remember to run 'libtool --finish /usr/local/apache2/modules'
```
I'm not sure if the `#{CREW_DEST_DIR}` prefix adversely affects the install or not but it appears to be working fine.

jam7: I researched libtool.  AFAIK, `libtool --finish` is used to modify `*.la` to point library path correctly.  It updates `*.la` files, but httpd doesn't install any `*.la` files, so it is not needed to perform `libtool --finish` here.  Please remove that line.  Thanks.

uberhacker: Yes, you are correct.  I have removed the line and updated the PR.
Thanks

PR #978, opened 2017-07-29
Add openssl dependency to serf package

uberhacker:


jam7:
Please increment version number like `1.3.9-1`

uberhacker:
Done.

PR #982, opened 2017-07-30
Add xmlclitools package

uberhacker:
xmlclitools provides four command-line tools for searching, modifying, and formating XML data. The tools are designed to work in conjunction with standard *nix utilities such as grep, sort, and shell scripts.  See http://freecode.com/projects/xmlclitools.

### DIFF
--- a/packages/xmlclitools.rb
+++ b/packages/xmlclitools.rb
@@ -1,0 +1,22 @@
+require 'package'
+
+class Xmlclitools < Package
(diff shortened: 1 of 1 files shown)


jam7:
URL looks like one of mirror sites.  I googled for a while but I cannot find better source URL, so I'm going to merge this one as is soon if there is no objection or no other opinions.  Please let me know if any.  Thanks.

uberhacker:
Yeah, this was the best location I could find also.  I don't have a problem downloading from a Russian site.  I like the Russians. :)

PR #987, opened 2017-07-31
Add gox package

uberhacker:
A dead simple, no frills Go cross compile tool.  See https://github.com/mitchellh/gox.

### DIFF
--- a/packages/gox.rb
+++ b/packages/gox.rb
@@ -11,6 +11,6 @@ class Gox < Package
 
   def self.install
     system "mkdir -p #{CREW_DEST_DIR}/usr/local/bin"
-    system "go build -ldflags \"-X main.VERSION=v0.3.0\" -o #{CREW_DEST_DIR}/usr/local/bin/gox"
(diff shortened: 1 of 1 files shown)


packages/gox.rb
```diff
@@ -0,0 +1,16 @@
...
+  homepage 'https://github.com/mitchellh/gox'
+  version '0.3.0'
+  source_url 'https://github.com/mitchellh/gox/archive/v0.3.0.tar.gz'
+  source_sha256 '29dc6b689f670a5444cc54cd9111549ccb01501901bc9197d0e1325a35157802'
+
+  depends_on 'go'
+
+  def self.install
+    system "mkdir -p #{CREW_DEST_DIR}/usr/local/bin"
+    system "go build -ldflags \"-X main.VERSION=v0.3.0\" -o #{CREW_DEST_DIR}/usr/local/bin/gox"
```
jam7: This `-ldflags -X main.VERSION` trick doesn't work with gox.  It is required to prepare a variable having the identical name in main.go.

uberhacker: OK, I removed the version flag and pushed another commit.

PR #993, opened 2017-08-01
Add waf package

uberhacker:
The meta build system.  See https://waf.io/.


jam7:
Waf installs many python files directly into `/usr/local/bin/.waf-1.9.12-389471dbd161471eece4b5393b9b9cc6/waflib/`.  I'm not sure what we should do.  Maybe it's OK if it is too difficult to solve.

Do you have any idea?

uberhacker:
I figured out how to avoid the `/usr/local/bin/.waf-...` directory.  Please test when you get a chance.

PR #995, opened 2017-08-01
Change imagemagick6.rb to update it to 6.9.9-4

jam7:
Tested on arm and x86_64.


jam7:
@lyxell Thank you for mergint them.  👍

PR #1000, opened 2017-08-03
Change node.rb to modify the way of configuration

jam7:
This time, I modified only the way of configuration by following the way in node_current.  This is required to compile node on arm.

Tested on arm, x86_64 and i686.

### DIFF
--- a/packages/node.rb
+++ b/packages/node.rb
@@ -11,7 +11,12 @@ class Node < Package
   depends_on 'python27'
 
   def self.build
-    system "CC='gcc -m64' python2.7 ./configure --without-snapshot"
+    case ARCH
+    when "x86_64"
+      system "CC='gcc -m64' python2.7 ./configure --without-snapshot"
+    else
+      system "CC='gcc' python2.7 ./configure --without-snapshot"
+    end
     system "make"
   end
 


uberhacker:
I would like to propose we remove `node` and `node_current` in favor of the `nvm` package.  It takes over 2 hours to compile node on my HP Chromebook.  It takes less than one minute to install any version of node with `nvm` and I can change the default version easily as well.  Also, I don't have the potential permission issues when installing node modules globally since it always installs with the current user permissions.  There are reasons https://github.com/creationix/nvm has over 20k stars.

jam7:
did you try nodebrew?  it works similar but doesn't modify users' env.  ;)

2017年8月4日(金) 6:28 Ed Reel <notifications@github.com>:

uberhacker:
I have not tried nodebrew but from my initial inspection, it looks like a replica of nvm.  This is not modifying the users' environment?
```
export PATH=$HOME/.nodebrew/current/bin:$PATH
```
BTW, the way I install nvm via crew is in `/usr/local/share/nvm` instead of `~/.nvm`.  This takes it out of the user directory.  I think we should seriously think about installing other packages like this that create hidden directories in the user environment.  I HATE, HATE, HATE needing to add additional paths unnecessarily.  There is a project called rvm (aka, not surprisingly, ruby version manager) that installs in `~/.local`.  This is definitely non-standard.

jam7:
for me, asking to change PATH is ok since user knows it.  but modifying
PATH hiddenly is not ok since it may open a security hole.  modifying
.bashrc exists between them both but I don't like it since it is easy to
not notice.

for package developer, I dislike to ask users to modify something.  so I
make package work without users' help.  for example, make a symbolic link
from /usr/local/bin like what jdk8 does.  I think we can do same thing for
nvm if it doesn't modify .bashrc hiddenly.

2017年8月4日(金) 7:45 Ed Reel <notifications@github.com>:

jam7:
Update node to 8.2.1 also.  We can choose `node` to compile it from source or `nvm` to use node version manager and precompiled binaries.

thedamian:
Node has two versions LTS and CURRENT

jam7:
@thedamian there are several node version manager like `rbenv`.  These days, there are `nodebrew`, `nvm`, `nbenv` and `anyenv` to manage node version.  Chromebrew supports `nodebrew` and `nvm` currently.  By using `nodebrew`, you can do something like below (sorry for no nvm example):
```
$ nodebrew list
v8.2.1
v8.3.0

current: v8.3.0
$ nodebrew ls-all
remote:
v0.0.1    v0.0.2    v0.0.3    v0.0.4    v0.0.5    v0.0.6

v0.1.0    v0.1.1    v0.1.2    v0.1.3    v0.1.4    v0.1.5    v0.1.6    v0.1.7
...
v6.11.1   v6.11.2

v7.0.0    v7.1.0    v7.2.0    v7.2.1    v7.3.0    v7.4.0    v7.5.0    v7.6.0
v7.7.0    v7.7.1    v7.7.2    v7.7.3    v7.7.4    v7.8.0    v7.9.0    v7.10.0
v7.10.1

v8.0.0    v8.1.0    v8.1.1    v8.1.2    v8.1.3    v8.1.4    v8.2.0    v8.2.1
v8.3.0
...
$ nodebrew install-binary v6
Fetching: https://nodejs.org/dist/v6.11.2/node-v6.11.2-linux-x64.tar.gz
Installed successfully
$ nodebrew use v6
use v6.11.2
$ node --version
v6.11.2
$ nodebrew use v8
use v8.3.0
$ node --version
v8.3.0
```
So, we are discussing to make `node.rb` as a wrapper to install the latest node.  Maybe we can make another `node6.rb` as a wrapper to install the latest node version 6.

uberhacker:
One goal I would like to achieve eventually is not having any version specific packages.  If we have the need for multiple versions, we should have a single package that provides options to install whatever version is needed.  In the case of node, nodebrew or nvm provides this ability.  I'm thinking we should modify node.rb as a fake package to either nodebrew or npm.  There are advantages and disadvantages to both of these options.  I'd like to remove all the disadvantages if possible.  Disadvantages include modifying the user environment and creating package specific hidden directories.  If we can avoid doing this sort of thing, that would be ideal.

jam7:
@uberhacker Anyenv, https://github.com/riywo/anyenv, is yours.  It is a version manager for ruby, perl, python, php, node, lua, go, etc.

By the way, modifying PATH is really required for them because for example `npm install new` installs a `new` executable file into their own bin directory.  We cannot make symbolic links each time.  So, really modifying PATH is needed if someone want to develop using them.

jam7:
Reverted the modification to update it to 8.2.1 because of https://github.com/skycocker/chromebrew/pull/1000#issuecomment-322088206.  Now, this PR is about only fixes of the way to compile.

We can discuss about how to organize node in #1056.

PR #1002, opened 2017-08-05
Update mapserver from 7.0.4 to 7.0.6

uberhacker:


jam7:
Thank you for fixing configuraion option and dependencies.  Still having following problem.
```
$ crew build mapserver
mapserver: MapServer is an Open Source platform for publishing spatial data and interactive mapping applications to the web.
http://mapserver.org/
version 7.0.6
Downloading source...
Archive downloaded
Scanning dependencies of target mapserver
[  2%] Building C object CMakeFiles/mapserver.dir/fontcache.c.o
[  2%] Building C object CMakeFiles/mapserver.dir/cgiutil.c.o
[  3%] Building C object CMakeFiles/mapserver.dir/mapgeos.c.o
[  3%] Building C object CMakeFiles/mapserver.dir/maporaclespatial.c.o
In file included from /usr/local/include/fribidi/fribidi.h:35:0,
                 from /usr/local/tmp/crew/mapserver-7.0.6.tar.gz.dir/mapserver-7.0.6/fontcache.h:13,
                 from /usr/local/tmp/crew/mapserver-7.0.6.tar.gz.dir/mapserver-7.0.6/fontcache.c:32:
/usr/local/include/fribidi/fribidi-common.h:61:20: fatal error: glib.h: No such file or directory
 #  include <glib.h>
                    ^
compilation terminated.
make[2]: *** [CMakeFiles/mapserver.dir/build.make:63: CMakeFiles/mapserver.dir/fontcache.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:105: CMakeFiles/mapserver.dir/all] Error 2
make: *** [Makefile:130: all] Error 2
mapserver failed to build: `make -j4` exited with 2
```
I even installed glib by hand, but it install `/usr/local/include/glib-2.0/glib.h` not `/usr/local/include/glib.h`.

uberhacker:
@jam7: See the fix in #1008.  Let me know if this resolves your issue.

jam7:
You are just opening a wound.  The source of problem is not a hidden glib.h but that fribidi is not supporting glib2.0.  Please fix fribidi  or related stuff.  There are several discussions like http://lists.osgeo.org/pipermail/mapserver-dev/2014-January/013872.html.

uberhacker:
@jam7: Thank you for bringing this issue to my attention, however, the link you gave me does not help me understand the problem.  Could you please explain it a bit more or better yet, provide a solution?

jam7:
I've researched about this problem deeply, so now I can explain details.

The problem is that `cmake` doesn't recognize the configuration required for fribidi correctly through `pkg-config`.  In order to compile files with fribidi, `-I/usr/local/include/fribidi -I/usr/local/include/glib-2.0 -I/usr/local/lib/glib-2.0/include` options are required as what `pkg-config` shows like below.  This result may different depends on each environment.
```
$ pkg-config fribidi --cflags-only-I
-I/usr/local/include/fribidi -I/usr/local/include/glib-2.0 -I/usr/local/lib/glib-2.0/include
```
However, `cmake` cannot retrieve them all correctly.  `cmake` confuse and think only `-I/usr/local/include` is needed for fribidi by unknown reasons.

To fix this problem, please add following `-DFRIBIDI_INCLUDE_DIR=...` line as a part of cmake command line.  Thanks.
```
      system "cmake \
              -DFRIBIDI_INCLUDE_DIR=`pkg-config fribidi --cflags-only-I|sed -e 's/^-I//' -e 's/ -I/;/g'` \
              -DWITH_CLIENT_WFS=1 \
              -DWITH_CLIENT_WMS=1 \
...
```

uberhacker:
I tested with this additional configure option and it works well.  I have updated the PR.
Thanks

PR #1008, opened 2017-08-05
Fix to make glib visible

uberhacker:
This should resolve the issue in #1002.


jam7:
This won't fix the problem.  This creates new problem.  :(

uberhacker:
What new problem is created?  Please elaborate.

uberhacker:
Fixed in PR #1002.

PR #1009, opened 2017-08-06
Update mosh from 1.3.0 to 1.3.2

uberhacker:


jam7:
Is it possible to change `make install` to `make DESTDIR=#{CREW_DEST_DIR} install`?  Thanks.

uberhacker:
Good call.  Missed that.

jam7:
Thanks.

PR #1010, opened 2017-08-06
Update mutt from 1.8.2 to 1.8.3

uberhacker:
Added libxslt dependency to resolve:
```
/bin/bash: xsltproc: command not found
```

### DIFF
--- a/packages/mutt.rb
+++ b/packages/mutt.rb
@@ -3,17 +3,17 @@ require 'package'
 class Mutt < Package
   description 'Mutt is a small but very powerful text-based mail client for Unix operating systems.'
   homepage 'http://mutt.org/'
-  version '1.8.2'
-  source_url 'ftp://ftp.mutt.org/pub/mutt/mutt-1.8.2.tar.gz'
-  source_sha256 '1d057bf1b565f2c38ee50c9a661654cbbe4165f98e25bfa361ebbd707d96f235'
+  version '1.8.3'
+  source_url 'ftp://ftp.mutt.org/pub/mutt/mutt-1.8.3.tar.gz'
+  source_sha256 '9b81746d67ffeca5ea44f60893b70dc93c86d4bc10187d4dd360185e4d42ed42'
 
   def self.build
-    system 'mkdir /usr/local/mail'
     system './configure --with-mailpath=/usr/local/mail --prefix=/usr/local'
     system 'make'
   end
 
   def self.install
-    system 'make -i install'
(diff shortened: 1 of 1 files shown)


packages/mutt.rb
```diff
@@ -3,17 +3,17 @@
...
   homepage 'http://mutt.org/'
-  version '1.8.2'
-  source_url 'ftp://ftp.mutt.org/pub/mutt/mutt-1.8.2.tar.gz'
-  source_sha256 '1d057bf1b565f2c38ee50c9a661654cbbe4165f98e25bfa361ebbd707d96f235'
+  version '1.8.3'
+  source_url 'ftp://ftp.mutt.org/pub/mutt/mutt-1.8.3.tar.gz'
+  source_sha256 '9b81746d67ffeca5ea44f60893b70dc93c86d4bc10187d4dd360185e4d42ed42'
 
   def self.build
-    system 'mkdir /usr/local/mail'
```
jam7: it looks like removing this line cause some problems.
```
$ crew install mutt
...
checking for va_copy... va_copy                                             
checking for ftruncate... yes                                                                                                         
checking for strftime... yes                                                
checking for fchdir... yes                                                                                                            
checking for regcomp... yes                                                 
checking whether your system's regexp library is completely broken... no                                                              
checking if /usr/local/mail is world writable... no                         
checking if /usr/local/mail is group writable... no                                                                                   
checking where to put the documentation... ${datarootdir}/doc/mutt
...
$ mutt
...
/var/spool/mail/chronos: No such file or directory (errno = 2)
```

uberhacker: I didn't see that error during my testing but I put it back and pushed another commit.

jam7:
Thank you for modifying that.  However, I was wrong.  `mkdir /usr/local/mail` just before `./configure` doesn't solve the problem I mentioned.  Moreover, it cause following different error.  Please remove it again.  Sorry for that.
```
$ crew install mutt
...
mkdir: cannot create directory '/usr/local/mail': File exists
mutt failed to install: `mkdir /usr/local/mail` exited with 1
```

uberhacker:
I removed that line again.  Strange that we're both experiencing different situations during the compilation process.

jam7:
I'm not sure how to use mutt yet.  However, install works perfectly.  And I can confirm that the given mail path `/usr//local/mail` is passed to compiled mutt correctly.   So, I'm merging this.  Thanks.

PR #1012, opened 2017-08-06
Fix to find library files in /usr/local/lib

uberhacker:
Needed for php7 update #1013.


jam7:
Nothing is changed if I apply this before #1013 or not.

uberhacker:
Closing because of suspected corrupt testing environment.

PR #1013, opened 2017-08-06
Update php7 from 7.1.7 to 7.1.8

uberhacker:
Please merge #1012 before this update.  EDIT: Please disregard.


jam7:
I merged 1012 before this, but nothing changed.  I don't understand why `libssh2` is needed here.  What is changed?

uberhacker:
The file `/usr/local/lib/libssh2.la` was not found and caused the compilation to abort, although the file was installed in `/usr/local/lib64/include/libssh2.la`.

jam7:
Can you show those error messages?  When I tried to compile php7 with not modified openssh2 on x86_64, I didn't see any error messages.  Thanks.

uberhacker:
I tried to reproduce but was unable.  I'll just chalk it up as another case of a corrupt testing environment.  The libssh2 dependency is not needed since the curl package already has it.

PR #1017, opened 2017-08-06
Update nginx from 1.11.6-1 to 1.13.3

uberhacker:
- Highlight help text in blue to stand out

### DIFF
--- a/packages/nginx.rb
+++ b/packages/nginx.rb
@@ -37,7 +37,7 @@ class Nginx < Package
     puts
     puts "To completely remove nginx, perform the following:".lightblue
     puts "crew remove nginx".lightblue
-    puts "sudo rm -rf /usr/local/nginx".lightblue
+    puts "sudo rm -rf #{CREW_PREFIX}/nginx".lightblue
     puts
   end
 end
--- a/packages/nginx.rb
+++ b/packages/nginx.rb
@@ -3,9 +3,9 @@ require 'package'
 class Nginx < Package
   description 'nginx [engine x] is an HTTP and reverse proxy server, a mail proxy server, and a generic TCP/UDP proxy server, originally written by Igor Sysoev.'
   homepage 'http://nginx.org/'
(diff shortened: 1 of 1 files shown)


jam7:
How do you think about making two scripts in `/usr/local/bin` like below instead of modifying user's .bashrc?
 - startnginx
```
#!/bin/bash
sudo nginx
```
 - stopnginx
```
#!/bin/bash
sudo nginx -s quit
```
I can submit PR later if you don't want that.

uberhacker:
I like it.  I'll try and update the PR.

uberhacker:
The PR has been updated.  Please test when you get a chance.

packages/nginx.rb
```diff
@@ -16,14 +16,28 @@ def self.build
...
-    system "echo pages are in /usr/local/nginx/html"
-    system "echo adding bash aliases so you can easily start/stop nginx"
-    system "echo startnginx starts nginx and stopnginx stops nginx"
-    system "sed -i \'/^alias startnginx/d\' ~/.bashrc"
-    system "sed -i \'/^alias stopnginx/d\' ~/.bashrc"
-    system "sed -i '$ a alias startnginx=\"sudo nginx\"' ~/.bashrc"
-    system "sed -i '$ a alias stopnginx=\"sudo nginx -s quit\"' ~/.bashrc"
+    system "mkdir -p #{CREW_DEST_DIR}#{CREW_PREFIX}/bin"
+    FileUtils.cd("#{CREW_DEST_DIR}#{CREW_PREFIX}/bin") do
+      system "sudo ln -sf #{CREW_PREFIX}/nginx/sbin/nginx nginx"
```
jam7: Please remove this sudo.  It works without sudo since making symbolic link under #{CREW_DEST_DIR}.
It is also possible to move `system "echo..."` commands here.

uberhacker: Another good call.  PR updated.

PR #1023, opened 2017-08-07
Corrected dependencies of several package files

jam7:
I just noticed several dependency problems while installing packages.  This fixes them.

### DIFF
--- a/packages/cairo.rb
+++ b/packages/cairo.rb
@@ -7,7 +7,6 @@ class Cairo < Package
   source_url 'https://www.cairographics.org/releases/cairo-1.14.10.tar.xz'
   source_sha256 '7e87878658f2c9951a14fc64114d4958c0e65ac47530b8ac3078b2ce41b66a09'
 
-  depends_on 'fontconfig'
   depends_on 'libpng'
   depends_on 'pixman'
   depends_on 'freetype'   # pango requires cairo with freetype
--- a/packages/cairo.rb
+++ b/packages/cairo.rb
@@ -10,9 +10,11 @@ class Cairo < Package
   depends_on 'fontconfig'
   depends_on 'libpng'
   depends_on 'pixman'
+  depends_on 'freetype'   # pango requires cairo with freetype
+  depends_on 'fontconfig' # pango requires cairo with fontconfig
 
   def self.build
-    system "./configure"
+    system "./configure --disable-xlib"
     system "make"
   end
 
--- a/packages/glib.rb
+++ b/packages/glib.rb
@@ -9,6 +9,8 @@ class Glib < Package
 
   depends_on 'libffi'
   depends_on 'gettext'
+  depends_on 'zlibpkg'
+  depends_on 'python27'
 
   def self.build
     system "./configure"
--- a/packages/httpd.rb
+++ b/packages/httpd.rb
@@ -11,6 +11,7 @@ class Httpd < Package
   depends_on 'aprutil'
   depends_on 'libtool'
   depends_on 'pcre'
+  depends_on 'expat'
 
   def self.build
     system './configure'
--- a/packages/libarchive.rb
+++ b/packages/libarchive.rb
@@ -10,6 +10,7 @@ class Libarchive < Package
   depends_on 'acl'
   depends_on 'attr'
   depends_on 'lz4'
+  depends_on 'xzutils' => :build
 
   def self.build
     system "./configure --prefix=/usr/local --disable-static"
--- a/packages/pango.rb
+++ b/packages/pango.rb
@@ -11,6 +11,7 @@ class Pango < Package
   depends_on 'fontconfig'
   depends_on 'harfbuzz'
   depends_on 'cairo'
+  depends_on 'glib'
 
   def self.build
     system "./configure"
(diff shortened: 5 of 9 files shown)


packages/cairo.rb
```diff
@@ -10,9 +10,11 @@ class Cairo < Package
   depends_on 'fontconfig'
   depends_on 'libpng'
   depends_on 'pixman'
+  depends_on 'freetype'   # pango requires cairo with freetype
+  depends_on 'fontconfig' # pango requires cairo with fontconfig
```
uberhacker: Duplicate `depends_on 'fontconfig'` lines?

packages/cairo.rb
```diff
@@ -10,9 +10,11 @@ class Cairo < Package
   depends_on 'fontconfig'
   depends_on 'libpng'
   depends_on 'pixman'
+  depends_on 'freetype'   # pango requires cairo with freetype
```
uberhacker: I believe freetype is a dependency of fontconfig so it's not needed here.

packages/cairo.rb
jam7: Thanks.  Fixed it.

packages/cairo.rb
jam7: If cairo acesses only fontconfig, I agree with you.  But it accesses both freetype and fontconfig.  Cairo generates multiple functions depends on existence of other packages, so leave them both increase the safeness against to the future modifications.

packages/cairo.rb
```diff
@@ -7,12 +7,13 @@ class Cairo < Package
   source_url 'https://www.cairographics.org/releases/cairo-1.14.10.tar.xz'
   source_sha256 '7e87878658f2c9951a14fc64114d4958c0e65ac47530b8ac3078b2ce41b66a09'
 
-  depends_on 'fontconfig'
   depends_on 'libpng'
   depends_on 'pixman'
+  depends_on 'freetype'   # pango requires cairo with freetype
```
uberhacker: Package freetype is not needed here since it's a dependency of fontconfig.

packages/glib.rb
```diff
@@ -9,6 +9,8 @@ class Glib < Package
 
   depends_on 'libffi'
   depends_on 'gettext'
+  depends_on 'zlibpkg'
```
uberhacker: Package zlibpkg is not needed here since it's a dependency of python27.

packages/httpd.rb
```diff
@@ -11,6 +11,7 @@ class Httpd < Package
   depends_on 'aprutil'
```
uberhacker: Package apr is not needed here since it's a dependency of aprutil.

packages/httpd.rb
```diff
@@ -11,6 +11,7 @@ class Httpd < Package
...
   depends_on 'libtool'
   depends_on 'pcre'
+  depends_on 'expat'
```
uberhacker: Package expat is not needed here since it's a dependency of aprutil.

packages/libarchive.rb
```diff
@@ -10,6 +10,7 @@ class Libarchive < Package
   depends_on 'acl'
   depends_on 'attr'
```
uberhacker: Package attr is not needed here since it's a dependency of acl.

packages/pango.rb
```diff
@@ -11,6 +11,7 @@ class Pango < Package
   depends_on 'fontconfig'
```
uberhacker: Package freetype is not needed here since it's a dependency of fontconfig.

uberhacker: Package fontconfig is not needed here since it's a dependency of cairo.

packages/cairo.rb
jam7: Didn't see https://github.com/skycocker/chromebrew/pull/1023#discussion_r132813769?

packages/glib.rb
jam7: Didn't see https://github.com/skycocker/chromebrew/pull/1023#discussion_r132813769?

packages/httpd.rb
jam7: Didn't see https://github.com/skycocker/chromebrew/pull/1023#discussion_r132813769?

packages/httpd.rb
jam7: It is outside of this PR.  I don't want to optimize acyclic paths of dependencies to avoid causing new problem by this PR.

packages/libarchive.rb
jam7: It is outside of this PR.  I don't want to optimize acyclic paths of dependencies to avoid causing new problem by this PR.

packages/pango.rb
jam7: It is outside of this PR.  I don't want to optimize acyclic paths of dependencies to avoid causing new problem by this PR.

jam7:
Add comments with a hope that @thedamian catch them and merge this PR soon.  ;)

packages/cairo.rb
uberhacker: You cannot install fontconfig without freetype so I don't understand your point.

jam7: I modified this like below at https://github.com/skycocker/chromebrew/pull/1023#discussion_r132813714.  There is no order problem like you points.
```
  depends_on 'libpng'
  depends_on 'pixman'
  depends_on 'freetype'   # pango requires cairo with freetype
  depends_on 'fontconfig' # pango requires cairo with fontconfig
```

jam7:
Add necessary dependency to harfbuzz discussed in #1059.

PR #1024, opened 2017-08-07
Fix wrong option in cmatrix.rb

jam7:
I cannot compile cmatrix with original package file.  I needed to change the way to set environment variable like this PR.

Tested on armv7l and x86_64.

### DIFF
--- a/packages/cmatrix.rb
+++ b/packages/cmatrix.rb
@@ -11,7 +11,7 @@ class Cmatrix < Package
   depends_on 'ncurses'
 
   def self.build
-    system './configure --prefix=/usr/local CPPFLAGS="-I/usr/local/include/ncurses"'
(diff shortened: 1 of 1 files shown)


packages/cmatrix.rb
```diff
@@ -11,7 +11,7 @@ class Cmatrix < Package
   depends_on 'ncurses'
 
   def self.build
-    system './configure --prefix=/usr/local CPPFLAGS="-I/usr/local/include/ncurses"'
+    system 'CPPFLAGS=-I/usr/local/include/ncurses ./configure --prefix=/usr/local'
```
uberhacker: `system "CPPFLAGS=-I#{CREW_PREFIX}/include/ncurses ./configure --prefix=#{CREW_PREFIX}"` ?

jam7: done

uberhacker:
Bump version to detect changes?

jam7:
done

PR #1027, opened 2017-08-08
Add nodebrew 0.9.7

jam7:
This PR adds nodebrew 0.9.7, another `node` version manager similar to `nvm`.

Once installed nodebrew, use can install their favorite version of `node` like:
```
$ nodebrew install-binary latest
Fetching: https://nodejs.org/dist/v8.2.1/node-v8.2.1-linux-armv7l.tar.gz
Installed successfully
$ nodebrew use latest
use v8.2.1
```

Tested on armv7l and x86_64.

### DIFF
--- a/packages/nodebrew.rb
+++ b/packages/nodebrew.rb
@@ -21,7 +21,7 @@ class Nodebrew < Package
 
     puts ""
     puts "Please set PATH environment variable."
-    puts "\texport PATH=$HOME/.nodebrew:$PATH"
+    puts "\texport PATH=$HOME/.nodebrew/current/bin:$PATH"
     puts ""
     puts "To install the latest node, please execute:"
     puts "\tnodebrew install-binary latest"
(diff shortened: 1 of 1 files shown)


packages/nodebrew.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    system "mkdir -p #{CREW_DEST_DIR}#{CREW_PREFIX}/bin"
+    system "ln -s #{CREW_LIB_PREFIX}/nodebrew/nodebrew #{CREW_DEST_DIR}#{CREW_PREFIX}/bin/"
+    system "ln -s #{CREW_LIB_PREFIX}/nodebrew/current/bin/node #{CREW_DEST_DIR}#{CREW_PREFIX}/bin/"
+
+    system "ln -s #{CREW_LIB_PREFIX}/nodebrew $HOME/.nodebrew"
+
+    puts ""
+    puts "Please set PATH environment variable."
+    puts "\texport PATH=$HOME/.nodebrew/current/bin:$PATH"
```
uberhacker: Do we need?
```
puts "\texport PATH=$HOME/.nodebrew/current/bin:$PATH && source ~/.bashrc"
```

jam7: It's up to how you think about the user level of chromebrew.  You can add them by another PR like what you did in go.rb.

uberhacker: OK, I think I understand what is going on here.  Is it possible to change the path to use `/usr/local/lib/nodebrew` instead of `~/.nodebrew` and symlink npm like node?  This would remove the need to add an extra directory to PATH that needs sourced.  Did I ever mention I _HATE_ adding extra directories to the PATH if it can be avoided? :)

jam7: Nodebrew modifies ~/.nodebrew.  So, in order to change version of node using nodebrew later, ~/.nodebrew is required.  It is possible to specify the path using NODEBREW_ROOT, but I prefer symbolic link from ~/.nodebrew since it's simple.

BTW, I've add symbolic link for npm and npx to run them without modifying PATH.  Future node may provide more binaries, so modifying PATH is better in general, I think.

PR #1032, opened 2017-08-09
Update yarn from 0.27.0-20170629.1443 to 1.0.0-20170808.1058

uberhacker:


uberhacker:
One thing I wanted to mention about nodebrew is it doesn't install npm like nvm does.  Boo!

jam7:
depends on #1027 

@uberhacker it installs npm.  What you need to do is update your PATH environment variable.

uberhacker:
@jam7: Is this an acceptable solution or should I change something?

jam7:
I cannot merge #1027 since it is my own PR and this PR depends on it, so I cannot merge this PR also.  That's the reason why I don't merge this one yet.

If you want to merge this one quickly, please don't use `nodebrew` dependency on this PR.  And please don't run `nodebrew` automatically to install the latest node since users may have chosen different version of node for users' work already.

uberhacker:
This goes against an ideal package manager policy which automatically installs dependencies as needed to complete the installation, but I have removed the nodebrew dependency.

packages/yarn.rb
```diff
@@ -3,11 +3,9 @@
...
   description 'Yarn is a new package manager for JavaScript and an alternative to npm.'
   homepage 'https://yarnpkg.com/en/'
-  version '0.27.0-20170629.1443'
-  source_url 'https://nightly.yarnpkg.com/yarn-v0.27.0-20170629.1443.tar.gz'
-  source_sha256 'e1086d267151cd05896b15023e8c4e253a64e675a80bb38dc294b5b1c86c6656'
-
-  depends_on 'node' unless File.exists? '/usr/local/bin/node'
+  version '1.0.0-20170808.1058'
+  source_url 'https://nightly.yarnpkg.com/yarn-v1.0.0-20170808.1058.tar.gz'
+  source_sha256 '7056c5ee12aa8d014d59b377be2e494ce49a6059e2678d8f79f1fc9904eb008b'
```
jam7: Can you add
```
depends_on 'node' unless `node --version`
```

uberhacker: I don't think it's a good idea to "force" install the compiled version of node if it isn't installed already.  A better way would be either via nvm or nodebrew.  Like you said before, it will just produce an error that node is missing and the user can decide how they want to install node.

jam7: I don't want to "force" to install the compiled version of node here.  What I am asking here is not change the common name to install `node`.  It's important to couple packages as loosely as possible.

If do so, you can make different PR to rename current `node` to `node_source` or something.  Then, you can make new `node` which install the latest node using `nvm` or `nodebrew`.  In this way, you don't need to write `nvm install ...` or `nodebrew install-binary latest` in each dependent package.  And everyone can install the latest node without confusion and spending time for source compile.

uberhacker: Agreed.  I will work on overhauling the node package to install via binary instead of compiling from source.  In the meantime, please hold off from merging this PR.

PR #1033, opened 2017-08-09
Add CREW_DEST_PREFIX and CREW_DEST_LIB_PREFIX constants to crew command

uberhacker:
This should simplify paths like `#{CREW_DEST_DIR}#{CREW_PREFIX}` and `#{CREW_DEST_DIR}#{CREW_LIB_PREFIX}` in packages.

### DIFF
--- a/crew
+++ b/crew
@@ -19,6 +19,8 @@ CREW_LIB_PATH = CREW_PREFIX + '/lib/crew/'
 CREW_CONFIG_PATH = CREW_PREFIX + '/etc/crew/'
 CREW_BREW_DIR = CREW_PREFIX + '/tmp/crew/'
 CREW_DEST_DIR = CREW_BREW_DIR + 'dest'
+CREW_DEST_PREFIX = CREW_DEST_DIR + CREW_PREFIX
+CREW_DEST_LIB_PREFIX = CREW_DEST_DIR + CREW_LIB_PREFIX
 
 # Set CREW_NPROC from environment variable or `nproc`
 if ENV["CREW_NPROC"].to_s == ''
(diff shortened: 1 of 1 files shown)


jam7:
Adding those new variable is OK, but I'm not sure why you increase the version number here.  How do you think about it?  Can you explain any reason why you think it should be increased?  Thanks.

uberhacker:
I'm of the opinion that _ANY_ changes to the `crew` command warrants a version bump.  Otherwise, you cannot differentiate between older versions.

jam7:
For the case of version number of each package file, I could say yes.  However, for the case of chromebrew itself, I don't agree to increment the version number for any changes.  I think we need to decide whether increase the version number or not for this.  I appreciate if you can explain reasons why you think increasing version number is better for this PR.

On the other hand, those are just different opinions.  The real problem here is that we don't have a concrete thought to update the version number of chromebrew.  Like why or when @skycocker updated it in the past.

I check the history of `crew` commit log, only you increased version number so often.  At the beginning you asked to increase it to other collaborators.  Then, you change to increase it by yourself silently including it in your PRs.  Is it possible to ask it again through issue page to increase version number?  Thanks.

In addition, it is also possible to discuss how/when we increase the version number in issue too.

uberhacker:
See #1035.

uberhacker:
@jam7: If the crew version is holding back this PR, I can take it out.  It's pointless to argue about it.

jam7:
Yes, it's the reason why I don't merge this yet as I asked you to change it above.

uberhacker:
OK, OK.  I didn't increment the Chromebrew version.

PR #1034, opened 2017-08-09
Fix wrong homepage resize

### LINKED ISSUE #819: Homepage looks wonky when screen is wider than 1200 pixels
The boxes shrink when the screen is wider than 1200 pixels at http://skycocker.github.io/chromebrew/.  This appears to be a possible css media query logic error.  Do any of you front-end guys want to take a stab at this?

flyingP0tat0:
Just changed the type of Bootstrap columns. You can check the fixes by applying the same changes in your DevTools' inspector.
Fixes #819.

### DIFF
--- a/index.html
+++ b/index.html
@@ -39,7 +39,7 @@
       </div>
 
       <div class="row marketing">
-        <div class="col-lg-6">
+        <div class="col-6">
           <div class="bs-example bs-example-type">
             <h4>Chromebrew installs what you need with its dependencies:</h4>
           </div>
@@ -80,7 +80,7 @@
       </div>
 
       <div class="row marketing">
-        <div class="col-lg-6">
+        <div class="col-6">
           <div class="bs-example bs-example-type">
             <h4>Install Chromebrew (along with Ruby and Git).</h4>
             <h5>Available on aarch64, armv7l, i686 and x86_64.</h5>
--- a/index.html
+++ b/index.html
@@ -39,7 +39,7 @@
       </div>
 
       <div class="row marketing">
-        <div class="col-6">
+        <div class="col-12">
           <div class="bs-example bs-example-type">
             <h4>Chromebrew installs what you need with its dependencies:</h4>
           </div>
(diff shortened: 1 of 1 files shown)


uberhacker:
LGTM. http://htmlpreview.github.io/?https://github.com/flyingP0tat0/chromebrew/blob/patch-12/index.html

index.html
```diff
@@ -39,7 +39,7 @@ <h3 class="text-muted">Chromebrew</h3>
       </div>
 
       <div class="row marketing">
-        <div class="col-lg-6">
+        <div class="col-6">
```
jam7: Can you remove this line completely and remove the line closing this `div` at https://github.com/flyingP0tat0/chromebrew/blob/5b00fb210cf93e95ca08305b9cf6a3c24e1e4bf7/index.html#L79 line?

index.html
```diff
@@ -80,7 +80,7 @@ <h3 class="text-muted">Chromebrew</h3>
...
```
jam7: Can you remove this line and https://github.com/flyingP0tat0/chromebrew/blob/5b00fb210cf93e95ca08305b9cf6a3c24e1e4bf7/index.html#L91?

index.html
uberhacker: Are you sure we should to do this?  I think a bootstrap max column width of 6 would work well.  Otherwise, it would span the entire page width.  Probably not desirable on a wide screen.  However, we can try it out and see.

jam7: I think those lines are under previous line's `div class="row marketing"`, so it won't span the entire page width.  Please try that one to see what it does.  Thanks.

We can say something like `div class="col"` if we don't want to remove these lines completely.  I just thought `"col-6"` is strange.  If it is the common way to mark this situation, I accept that one too.  Please let me know.

flyingP0tat0: @jam7 `col-6` is a valid option. See the [grid system manual](https://v4-alpha.getbootstrap.com/layout/grid/#grid-options).

index.html
flyingP0tat0: @jam7 See my response above.

index.html
jam7: Thank you.  I couldn't find it by my google search.

Now I understand it is valid, but still not sure what purpose exists.  Why do you choose `col-6` instead of `col-12`?  Thanks.

flyingP0tat0: We can do another `col` size, I just did it this way because it was this way before.
But yes, `col-12` would make more sense.

jam7: Thanks for modifying that.  I also need some sense that I fee I understand and agree with in order to merge it by my name.  This is small change but important for me to agree with this PR.  Thanks a lot.

PR #1036, opened 2017-08-10
LuaJIT dependencies and binary

yancouto:
LuaJIT had missing dependencies. Also added binary for x86_64.

### DIFF
--- a/packages/luajit.rb
+++ b/packages/luajit.rb
@@ -16,7 +16,6 @@ class Luajit < Package
   })
 
   depends_on 'buildessential' => :build
-  depends_on 'make' => :build
 
   def self.build
     system 'make'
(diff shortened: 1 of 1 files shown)


yancouto:
If this binary is ok, I can add binaries (compiled in my machine) for many other packages too.

uberhacker:
@yancouto: Thank you for adding this.  I think it might make sense to submit the binary to somewhere like bintray.  More on that discussion [here](#140).

packages/luajit.rb
```diff
@@ -7,6 +7,17 @@ class Luajit < Package
...
+  binary_url ({
+    x86_64: 'https://github.com/yancouto/chromebrew/releases/download/bin/luajit-2.0.5-chromeos-x86_64.tar.xz'
+  })
+
+  binary_sha256 ({
+    x86_64: 'bba89cae837d6e88a4f1b9dcaa135ef4af1d6ecbd20358c33b034813a51597b4'
+  })
+
+  depends_on 'buildessential' => :build
+  depends_on 'make' => :build
```
jam7: Thanks.  Can you remove this `make` line since `buildessential` includes `make` also.

Issue #1026, opened 2017-08-07
Multiple arguments in a single command?


uberhacker:
```
$ crew install pkg1 pkg2 pkg3 ...
$ crew remove pkg1 pkg2 pkg3 ...
$ crew upgrade pkg1 pkg2 pkg3 ...
```
Possible?

flyingP0tat0:
That would need a lot of rewriting, implementing some sort of framework. So yes, it is possible, with some new code and packages.

yancouto:
Wouldn't it be possible to use one of [these](http://www.awesomecommandlineapps.com/gems.html)? Or would implementing one from scratch be better?

It would be cool to give extra commands with dashes, something like `crew install --from-source --keep pkg1 pkg2`

uberhacker:
@yancouto: I'm getting a blank green screen at http://www.awesomecommandlineapps.com/gems.html.

yancouto:
It was just a page with some gems to make it easier to deal with command line stuff. [This](https://www.sitepoint.com/ruby-command-line-interface-gems/) has less options but seems more relevant to this issue, and more detailed.

flyingP0tat0:
[Thor](https://github.com/erikhuda/thor) seems well documented and under continious development.

uberhacker:
Should we close this issue now that #1079 has been merged?

PR #1038, opened 2017-08-11
Update crew command to accept multiple arguments

uberhacker:
This is a proof of concept to add multiple argument handling to the `crew install` and `crew remove` commands.
```
$ crew install ascii cpustat di dos2unix sysstat
...
Ascii installed!
...
Cpustat installed!
...
Di installed!
...
Dos2unix installed!
...
Sysstat installed!
$ crew remove ascii cpustat di dos2unix sysstat
Ascii removed!
Cpustat removed!
Di removed!
Dos2unix removed!
Sysstat removed!
```
Addresses #1026.


jam7:
This PR breaks `crew upgrade` since it calls remove commands several times.
```
$ crew upgrade libssh2 nodebrew
(i) libssh2: libssh2 is a client-side C library implementing the SSH2 protocol.
https://www.libssh2.org/
version 1.8.0
Updating libssh2...
Archive downloaded
Removing since upgrade...
Libssh2 removed!
Nodebrew removed!    <---- ***nodebrew should not remvoed at here***
Installing...
Libssh2 installed!
Archive downloaded
```

You can solve this problem by modifying crew like below instead of modifying each functions to accept ARGV.  You also not need to re-indent every functions.
```
when "install"
  if ARGV.length > 1
    for i in 1...ARGV.length
      @pkgName = ARGV[i]
      search @pkgName
      resolve_dependencies_and_install
    end
  else
    help "install"
  end
...
when "remove"
... similar to above.
```

uberhacker:
I don't think that will work either.  It will break this line:
```
unless ARGV[2] == 'keep'
```

jam7:
I'm not sure why you submitted this PR while all others were talking about what kind of framework is good to treat such switches in #1026.  Please check that issue again.  Thanks.

uberhacker:
I did mention that it was a proof of concept.  I think using a framework is overkill.  I will update this PR soon and hopefully, it will be more than just a proof of concept.

uberhacker:
Please test the latest commit when you get a chance.

uberhacker:
Opening per @thedamian's request.

uberhacker:
Not considered as a feasible solution.

PR #1040, opened 2017-08-12
Fix libX11 install of SDL, SDL2 and Cairo

### LINKED ISSUE #1037: cairo install fail
Cairo installation fails with the message

`cairo-xlib-surface.c:71:33: fatal error: X11/extensions/XShm.h: No such file or directory
 #include <X11/extensions/XShm.h>`

I'm not sure which dependency should be added.

flyingP0tat0:
Only a temporary fix for building these packages while we are building the GUI.

Fixes #1037.

### DIFF
--- a/packages/cairo.rb
+++ b/packages/cairo.rb
@@ -3,7 +3,7 @@ require 'package'
 class Cairo < Package
   description 'Cairo is a 2D graphics library with support for multiple output devices.'
   homepage 'https://www.cairographics.org'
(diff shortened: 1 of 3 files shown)


jam7:
Thank you for the quick fixes.  Please increment version number something like this.
```
   version '1.14.10-1'
```

flyingP0tat0:
@jam7 Done.

jam7:
Thanks.

PR #1041, opened 2017-08-12
Update subversion to 1.9.7 and correct the install method of serf

jam7:
This PR updates `subversion` to 1.9.7.

While I was trying to install it, I noticed `serf` is installed into `/usr/local/tmp/crew/dest/usr/local` directly instead of the combination of compiling it for `/usr/local` and installing it to `/usr/local/tmp/crew/dest/usr/local`.  So, I fixed it.  Once I fixed it, several hacks to `subversion` are not needed anymore, so I removed them also.

When I check `get-dep.sh` script, I notice we don't have `zlibpkg` dependency.  So, I add that one too.

Tested on armv7l and x86_64.

### DIFF
--- a/packages/serf.rb
+++ b/packages/serf.rb
@@ -3,7 +3,7 @@ require 'package'
 class Serf < Package
   description 'The serf library is a high performance C-based HTTP client library built upon the Apache Portable Runtime (APR) library.'
   homepage 'https://serf.apache.org/'
-  version '1.3.9-1'
+  version '1.3.9-2'
   source_url 'https://www.apache.org/dist/serf/serf-1.3.9.tar.bz2'
   source_sha256 '549c2d21c577a8a9c0450facb5cca809f26591f048e466552240947bdf7a87cc'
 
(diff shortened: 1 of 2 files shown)


packages/serf.rb
```diff
@@ -3,16 +3,19 @@
...
+  version '1.3.9-2'
   source_url 'https://www.apache.org/dist/serf/serf-1.3.9.tar.bz2'
   source_sha256 '549c2d21c577a8a9c0450facb5cca809f26591f048e466552240947bdf7a87cc'
 
   depends_on 'scons'
   depends_on 'aprutil'
   depends_on 'openssl'
 
+  def self.build
+    system "scons APR=/usr/local APU=/usr/local OPENSSL=/usr/local/include/openssl PREFIX=#{CREW_PREFIX} LIBDIR=#{CREW_LIB_PREFIX}"
```
uberhacker: `system "scons APR=#{CREW_PREFIX} APU=#{CREW_PREFIX} OPENSSL=#{CREW_PREFIX}/include/openssl ...`
Better?

jam7: done

PR #1043, opened 2017-08-12
Changes to use lib64 on x86_64

jam7:
Change several packages to use lib64 on x86_64
Also disable making static libraries.

Tested on armv7l and x86_64.

### DIFF
--- a/packages/finch.rb
+++ b/packages/finch.rb
@@ -3,34 +3,38 @@ require 'package'
 class Finch < Package
   description 'Finch is a chat program which lets you log in to accounts on multiple chat networks simultaneously.'
   homepage 'http://pidgin.im/'
-  version '2.12.0'
+  version '2.12.0-1'
   source_url 'https://downloads.sourceforge.net/project/pidgin/Pidgin/2.12.0/pidgin-2.12.0.tar.bz2'
   source_sha256 '8c3d3536d6d3c971bd433ff9946678af70a0f6aa4e6969cc2a83bb357015b7f8'
 
   depends_on 'glib'
   depends_on 'ncursesw'
   depends_on 'tcl'
+  depends_on 'perl'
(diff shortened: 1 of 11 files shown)


packages/finch.rb
```diff
@@ -3,34 +3,38 @@
...
+      '--disable-gstreamer-interfaces',
+      '--disable-idn',
+      '--disable-meanwhile',
+      '--disable-screensaver',
+      '--disable-sm',
+      '--disable-tk',
+      '--disable-vv',
+      '--enable-consoleui=yes',
+      '--includedir=/usr/local/include',
+      '--oldincludedir=/usr/local/include',
```
uberhacker:
```
"--includedir=#{CREW_PREFIX}/include",
"--oldincludedir=#{CREW_PREFIX}/include",
```
Better?

uberhacker:
I'm wondering if we should just automatically add the following configure options to every package:
```
"--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}", "--includedir=#{CREW_PREFIX}/include"
```

packages/finch.rb
jam7: done

jam7:
@uberhacker Nice idea.  Why don't you try that?
@flyingP0tat0 BTW, x11 libraries looks like CREW_LIB_PREFIX is needed also.

thedamian:
I hate updates that "do it right" 

Is it right if it's already working? What HUGE benefits do we get with these changes? I'm all for progress but don't want to screw up anything while doing that.

jam7:
It's worth to do for only Chrome OS 60 or above on x86_64.  Chrome OS 60 is in beta channel.  It will arrives stable channel soon.  This PR is preparing for it.

PR #1045, opened 2017-08-12
Add vdev

flyingP0tat0:
Works as a replacement for `udev`, allowing to compile many other programs.


jam7:
Don't we need `depends_on vdev`?

flyingP0tat0:
@jam7 Where?

jam7:
@flyingP0tat0 not sure.  I was thinking all others tools in this PR may depends on vdev.  I'll check tonight.

flyingP0tat0:
@jam7 They are vdev dependencies.

jam7:
Ahh... never mind.  I just didn't notice `depends_on` stuff in `vdev`.

Several problems:
 - All of libraries are installed into hard-coded lib directory.  I appreciate if you can fix them.
 - Lvm2 tries to install files into /etc.  We may need to modify it to /usr/local/etc.
 - Iproute2 tries to install files into /etc, /sbin, /usr/lib, /usr/include, /usr/share and /var.  Actually it install nothing to /usr/local.  Do we still need iproute2?

flyingP0tat0:
@jam7 Yes, we need these packages for udev compatibility. I will try to fix this.

jam7:
@flyingP0tat0 check my comment to #1050.  In order to use glib, we can disable libmount option for now

flyingP0tat0:
@jam7 I need to get `util_linux` to work either way, so we can wait until that works.

jam7:
Ok.  I'll test it again tomorrow.

thedamian:
@jam7 the "i'll test it tomorrow" was for the other package right? the one I haven't merged yet, or these new ones?

jam7:
@thedamian Tha'ts correct.  I mentioned about #1050.  Anywa, #1050 depends on this PR, so I'm waiting for fixes of this PR.

thedamian:
so I'm waiting here on other PRs right? 

If there's a PR that's just waiting that's simple and I can merge to allow this to be merged, plese let me know.

flyingP0tat0:
@jam7 Now it should work.

packages/lvm2.rb
```diff
@@ -0,0 +1,20 @@
...
+  description 'LVM2 refers to the userspace toolset that provide logical volume management facilities on linux.'
+  homepage 'https://sourceware.org/lvm2'
+  version '2.02.174'
+  source_url 'ftp://sources.redhat.com/pub/lvm2/releases/LVM2.2.02.174.tgz'
+  source_sha256 'fb6ecec0408cb14ac4923c87cb0a356481122d7d0ae3fee6e94d2edb9d170a8b'
+
+  depends_on 'readline'
+
+  def self.build
+    system "./configure --disable-selinux --prefix=#{CREW_PREFIX}"
```
jam7: Don't you need to install configuration files into `/usr/local/etc/lvm` instead of `/etc/lvm`?  In order to do so, we need to modify this line like below.
```
    system "./configure --disable-selinux --prefix=#{CREW_PREFIX} --with-confdir=#{CREW_PREFIX}/etc"
```

jam7:
And, I've just noticed you are using your github repository as source URLs for libpstat, fskit and vdev.  If you are using your repository in order to prepare tar.gz files without modifications, you can use original repository like this, https://github.com/jcnelson/fskit/archive/2feb33a84932980e4ed9691242537771dd4e80f4.zip.  Please use `2feb33a` as their version number if you use this method.  Thanks.

uberhacker:
@flyingP0tat0: You could also use https://github.com/jcnelson/fskit/archive/2feb33a84932980e4ed9691242537771dd4e80f4.tar.gz and it would not require `depends_on 'unzip'` in your package.

flyingP0tat0:
@jam7 @uberhacker Done.

packages/fskit.rb
```diff
@@ -0,0 +1,19 @@
+require 'package'
+
+class Fskit < Package
+  description 'Filesystem utility library and SDK'
+  homepage 'https://github.com/jcnelson/fskit'
+  version '1.0.0'
```
uberhacker: Should the version be `2feb33`?

packages/vdev.rb
```diff
@@ -0,0 +1,33 @@
+require 'package'
+
+class Vdev < Package
+  description 'A device-file manager for *nix'
+  homepage 'https://github.com/jcnelson/vdev'
+  version '1.0.0'
```
uberhacker: Should the version by `ceb7a6`?

packages/libpstat.rb
```diff
@@ -0,0 +1,17 @@
+require 'package'
+
+class Libpstat < Package
+  description 'Library for getting information about running processes'
+  homepage 'https://github.com/jcnelson/libpstat'
+  version '1.0.0'
+  source_url 'https://github.com/flyingP0tat0/libpstat/archive/1.0.0.tar.gz'
```
uberhacker: Still using personal repo?

PR #1046, opened 2017-08-13
Change crew to accept multiple arguments by using docopt

jam7:
This is another proof of concept implementation of multiple arguments supports to show using a framework is not overkill.  Related to #1026 and #1038.

Pros:
 - automated error check
 - easy handling

Cons:
 - require installation of `docopt` individually (it is possible to install it in install.sh though)

How to handle new commands:
  1. make `command`_command function like:
```
def build_command (args)
  args["<name>"].each do |name|
    @pkgName = name
    search @pkgName
    resolve_dependencies_and_build
  end
end
```
 2. that's it.  following code call each function:
```
def is_command (name)
  return false if name =~ /^[-<]/
  return true
end

command_name = args.find { |k, v| v && is_command(k) } [0]
function = command_name + "_command"
send(function, args)
```
Note: In order to try this code, `gem install docopt` is required.

### DIFF
--- a/crew
+++ b/crew
@@ -5,9 +5,38 @@ require 'uri'
 require 'digest/sha2'
 require 'json'
 require 'fileutils'
+require 'docopt'
 
-@command = ARGV[0]
-@pkgName = ARGV[1]
+doc = <<DOCOPT
+Chromebrw - Package manager for Chrome OS http://skycocker.github.io/chromebrew/
+
+Usage:
+  #{__FILE__} build [-k|--keep] <name>...
+  #{__FILE__} download <name>...
+  #{__FILE__} help [<command>]
+  #{__FILE__} install [-k|--keep] <name>...
+  #{__FILE__} remove <name>...
+  #{__FILE__} search [-d|--desc] [<name>...]
+  #{__FILE__} update
+  #{__FILE__} upgrade [-k|--keep] [<name>...]
+  #{__FILE__} whatprovides <name>...
+
+  -k --keep  Keep extracted files as is.
+  -d --desc  Show extra information.
+  -h --help  Show this screen.
+
+version 0.4.3
+DOCOPT
+
+begin
+  args = Docopt::docopt(doc)
+rescue Docopt::Exit => e
+  puts e.message
+  exit 1
+end
+
+@opt_keep = args["--keep"]
+@opt_desc = args["--desc"]
 
 ARCH = `uname -m`.strip
 ARCH_LIB = if ARCH == 'x86_64' then 'lib64' else 'lib' end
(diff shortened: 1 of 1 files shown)


uberhacker:
My PR is a complete solution including documentation, required no dependencies and the code is less complicated.  I still think this is overkill but you are welcome to take over the responsibility.

yancouto:
I like this solution. Docopt seems nice. New commands are easily added and documented, and everything works as expected.

Instead of doing `gem install docopt` on `install.sh` (and how to do it for current users?), wouldn't it be better to just add the library to a folder `externals` inside `lib`? Everything would work fine, no worries about the user removing docopt or it being updated and breaking stuff. Also, it would work for current users and the library seems to be [just one file](https://github.com/docopt/docopt.rb/tree/master/lib).

thedamian:
@uberhacker can you refer to your solution so we can compare?

uberhacker:
@thedamian: See #1038.

jam7:
@yancouto Thanks.  Is docopt safe to copy in?  They use MIT license and we use GPL-3.0.  I'm not sure what is needed to copy their files without license issues.

yancouto:
@jam7 I don't think there is a problem. MIT Licence allows us to copy and distribute that code. The only restriction is we need to distribute a copy of the license with it. Just adding a `docopt.LICENSE` file in the same folder or adding the license as a comment in the beginning of the `docopt.rb` file should be fine.

flyingP0tat0:
Although I get the point of this being to overkill, to make it future-proof and reliable, the best way is to implement a tool like this one. It will be useful later on. Also remember: Do few things and do them good. We don't need to reinvent the wheel, if there is a tested gem out there.
+1 from me for this solution

_Sent from my OnePlus ONEPLUS A3003 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

crew
```diff
@@ -731,62 +760,77 @@ def remove (pkgName)
...
+end
+
+def is_command (name)
+  return false if name =~ /^[-<]/
+  return true
+end
+
+command_name = args.find { |k, v| v && is_command(k) } [0]
+function = command_name + "_command"
+send(function, args)
```
uberhacker: What happens if the switch is somewhere in-between or at the end of the argument list?
```
crew install pkg1 --keep pkg2 ...
```
or
```
crew install pkg1 pkg2 ... --keep
```
Does your logic treat it as a switch or a package?  These kinds of things should be considered.

uberhacker:
Don't get me started on the myth that a framework will help you not re-invent the wheel.  How much extra logic is needed to get this to actually work that is not even related to the framework in this PR?

crew
jam7: So what behavior you want?  It smells like you choose opposite whatever I answer.

Anyway, docopt consideres both are switches.

uberhacker: That doesn't answer my question.  When looping through the arguments, is the switch removed so the process operates on packages and not switches?

jam7: Yes.  args['name'] contains only pkg1 and pkg2.

flyingP0tat0:
@uberhacker It is about not making some mistakes. Not today, but in the future. There is a reason ruby has gems. If you are using ruby, you should make use of gems.
When writing a node app, you don't write your own module because you can. If there is a actively maintained gem from guys who now what they are doing, why not use it?

Issue #848, opened 2017-07-04
Dropbox fails to start


vince:
Just installed Dropbox and running dropbox start fails with:

```
Starting Dropbox...Traceback (most recent call last):
  File "<string>", line 5, in <module>
zipimport.ZipImportError: not a Zip file: '/usr/local/bin/dropbox-lnx.x86_64-28.4.14/library.zip'
^[c^CTraceback (most recent call last):
  File "/usr/local/bin/dropbox.py", line 1535, in <module>
    ret = main(sys.argv)
  File "/usr/local/bin/dropbox.py", line 1524, in main
    result = commands[argv[i]](argv[i+1:])
  File "/usr/local/bin/dropbox.py", line 1395, in start
    if not start_dropbox():
  File "/usr/local/bin/dropbox.py", line 741, in start_dropbox
    time.sleep(interval)
```

Trying to unzip it manually I see something similar:

```
chronos@localhost /usr/local/bin/dropbox-lnx.x86_64-28.4.14 $ unzip library.zip 
Archive:  library.zip
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
note:  library.zip may be a plain executable, not an archive
unzip:  cannot find zipfile directory in one of library.zip or
        library.zip.zip, and cannot find library.zip.ZIP, period.
```

Edit: I ended up downloading that zip file and manually copying it over and then it launched. Now just trying to figure out how to use the package..

uberhacker:
Hi @vince: Thank you for reporting this issue.  Could you please send your system specs?  You can find them in `Settings` at the top by clicking on `About Chromium OS`.  Also, please send the output of `uname -a` from the terminal.

vince:
Hey there, thanks for the response!

Here's uname -a
    Linux localhost 3.18.0-15776-g24ef2013948e #1 SMP PREEMPT Tue Jul 4 02:52:28 PDT 2017 x86_64 Intel(R) Core(TM) m3-6Y30 CPU @ 0.90GHz GenuineIntel GNU/Linux

About Chrome
    Version 61.0.3147.0 (Official Build) canary (64-bit)

I ended up installing the headless version of dropbox using [these instructions](https://www.dropbox.com/install-linux) and that seemed to work so I no longer need this fixed but happy to help debug!

uberhacker:
[...]

PR #1047, opened 2017-08-13
Strip binaries but not compressed archives

uberhacker:
This fixes the issue in #848.

### DIFF
--- a/crew
+++ b/crew
@@ -475,8 +475,8 @@ def install_package (pkgdir)
     strip_find_files "find . -type f -name 'lib*.a' -print", "-S"
     strip_find_files "find . -type f -name 'lib*.so*' -print", "-S"
 
-    # Strip binaries
-    strip_find_files "find . -type f -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d'"
+    # Strip binaries but not compressed archives
+    strip_find_files "find . -type f ! -iname '*\.zip' ! -iname '*\.tar' ! -iname '*\.gz' ! -iname '*\.tgz' ! -iname '*\.bz2' -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d'"
 
     system "tar cf - ./usr/* | (cd /; tar xp --keep-directory-symlink -f -)"
   end
--- a/crew
+++ b/crew
@@ -476,7 +476,7 @@ def install_package (pkgdir)
     strip_find_files "find . -type f -name 'lib*.so*' -print", "-S"
 
     # Strip binaries but not compressed archives
-    strip_find_files "find . -type f ! -iname '*\.zip' ! -iname '*\.tar' ! -iname '*\.gz' ! -iname '*\.tgz' ! -iname '*\.bz2' -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d'"
+    strip_find_files "find . -type f ! -iname '*\.bz2' ! -iname '*\.gz' ! -iname '*\.lha' ! -iname '*\.rar' ! -iname '*\.tar' ! -iname '*\.tgz' ! -iname '*\.xz' ! -iname '*\.zip' -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d'"
 
     system "tar cf - ./usr/* | (cd /; tar xp --keep-directory-symlink -f -)"
   end


jam7:
Nice notice.  I'll check details soon.  Thanks.

jam7:
Thank you for your insight.  It's identify the source of problem.

However, this PR does solve the problem if and only if zip file is not hard or soft linked to binary file.  Because, checking name solves the problem from zip file side, but doesn't from binary file side.  So, I'm still looking for better solution about this.  Maybe we can check elf format size and actual file size...

uberhacker:
The best solution, IMHO, would be to identify only the relevant binary files that need stripped and operate on those only.  However, identifying the gamut of possible binary files that need stripped is likely not attainable, so excluding files that shouldn't be stripped may be the only feasible option.

jam7:
@uberhacker You are thinking problem differently.  Think about situation like below:
```
$ touch a b
$ zip test.zip a b
...
$ echo 'main() {puts("hello");}' > hello.c
$ gcc hello.c
...
$ cat a.out test.zip > binary
$ chmod a+x binary
$ ln binary library.zip
$ ./binary
hello
$ unzip -l library.zip 
Archive:  library.zip
warning [library.zip]:  8600 extra bytes at beginning or within zipfile
  (attempting to process anyway)
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2017-08-13 23:11   a
        0  2017-08-13 23:11   b
---------                     -------
        0                     2 files
```
It's what dropbox provides.  It is a signle file.  Therefore, it is impossible to strip only newbinary.

This PR works for `dropbox.rb` situation.  The reason is this `dropbox.rb` copies `dropbox` and `library.zip` separately by using `cp -r` for copying contents.  It's the reason I'm looking for better solution.

uberhacker:
I guess I'm confused why you think library.zip is symlinked.  I was able to extract the contents fine from the source_url downloaded tarball.

jam7:
It is actually hard-link to dropbox like below.  I've never said it is a symbolic-link, though.
```
$ crew download dropbox
$ tar tvf  /usr/local/tmp/crew/dropbox-lnx.x86_64-30.4.22.tar.gz|grep library
hrwxr-xr-x releng/releng        0 2017-07-12 12:48 .dropbox-dist/dropbox-lnx.x86_64-30.4.22/library.zip link to .dropbox-dist/dropbox-lnx.x86_64-30.4.22/dropbox
```

jam7:
Not sure which library.zip you are talking about.  It is ELF format binary file at the beginning.
```
$ file library.zip
library.zip: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.15, BuildID[sha1]=4993d9df69a7bd249187d38df9a6fd8431293cc7, stripped
$ od -c library.zip |head
0000000 177   E   L   F 002 001 001  \0  \0  \0  \0  \0  \0  \0  \0  \0
0000020 002  \0   >  \0 001  \0  \0  \0  \0   `   E  \0  \0  \0  \0  \0
0000040   @  \0  \0  \0  \0  \0  \0  \0     022   O  \0  \0  \0  \0  \0
0000060  \0  \0  \0  \0   @  \0   8  \0  \t  \0   @  \0 036  \0 035  \0
0000100 006  \0  \0  \0 005  \0  \0  \0   @  \0  \0  \0  \0  \0  \0  \0
0000120   @  \0   @  \0  \0  \0  \0  \0   @  \0   @  \0  \0  \0  \0  \0
0000140 370 001  \0  \0  \0  \0  \0  \0 370 001  \0  \0  \0  \0  \0  \0
0000160  \b  \0  \0  \0  \0  \0  \0  \0 003  \0  \0  \0 004  \0  \0  \0
0000200   8 002  \0  \0  \0  \0  \0  \0   8 002   @  \0  \0  \0  \0  \0
0000220   8 002   @  \0  \0  \0  \0  \0 034  \0  \0  \0  \0  \0  \0  \0
$ unzip -l library.zip |head
Archive:  library.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
     3265  06-27-2017 14:53   __main__.pyc
      257  07-12-2017 12:46   __main__dropbox__.pyc
    15857  06-27-2017 14:33   BaseHTTPServer.pyc
    20929  06-27-2017 14:33   ConfigParser.pyc
    16849  06-27-2017 14:33   Cookie.pyc
    14065  06-27-2017 14:33   HTMLParser.pyc
      113  10-25-2015 04:41   PyQt5/__init__.pyc
```

jam7:
Well, it is required to copy linked file to different separated files, but this works with `dropbox.rb`.  I'm merging this.  Maybe we can find better solution later.

jam7:
Just an additional info.  It is possible to check whether a file is already stripped or not by `objdump -t library.zip|head|grep "no symbols"`, but this is not perfect too...

PR #1048, opened 2017-08-13
Update dropbox from 30.4.22 to 32.4.23

uberhacker:


jam7:
This and previous `dropbox` install `/usr/local/bin/VERSION` file.  Don't you mind to put them another directory?  Like this way?  Thanks.
```
    system "sed -i 's,~/.dropbox-dist,#{CREW_LIB_PREFIX}/dropbox/.dropbox-dist,g' dropbox.py"
...
    system "mkdir -p #{CREW_DEST_LIB_PREFIX}/dropbox"
    system "cp -r .dropbox-dist #{CREW_DEST_LIB_PREFIX}/dropbox"
    system "cp dropbox.py #{CREW_DEST_LIB_PREFIX}/dropbox"
    system "cp dropbox #{CREW_DEST_PREFIX}/bin"
```

thedamian:
I'll have to test it out myself and see the answer to @jam7's question before I merge

uberhacker:
@jam7: Agreed.  Installing everything in /usr/local/bin is a bit sloppy.  I have pushed up a new commit.  Please test again when you get a chance.

uberhacker:
BTW, I noticed that the `dropbox` shell script was not accepting more than one argument so I fixed that also.

jam7:
@uberhacker Thank you for the update.  BTW, the best way to pass all arguments in shell script is to use `prg "$@"`.  This pass all arguments with spaces correctly even if user runs this script like `script "a b c"`.

Issue #1051, opened 2017-08-13
Compile errors for many packages


yancouto:
I was trying to build all packages, but some did not work. Here are their errors:

## LIBCAP

gcc -O2 -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -Dlinux -Wall -Wwrite-strings -Wpointer-arith -Wcast-qual -Wcast-align -Wstrict-prototypes -Wmissing-prototypes -Wnested-externs -Winline -Wshadow -g  -fPIC -I/usr/local/tmp/crew/libcap-2.25.tar.xz.dir/libcap-2.25/libcap/../libcap/include/uapi -I/usr/local/tmp/crew/libcap-2.25.tar.xz.dir/libcap-2.25/libcap/../libcap/include -c cap_file.c -o cap_file.o
make[1]: *** [Makefile:63: cap_text.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/usr/local/tmp/crew/libcap-2.25.tar.xz.dir/libcap-2.25/libcap'
make: *** [Makefile:12: all] Error 2
libcap failed to build: `make -j2` exited with 2

## CLAMAV

configure: error: The installed zlib version may contain a security bug. Please upgrade to 1.2.2 or later: http://www.zlib.net. You can omit this check with --disable-zlib-vcheck but DO NOT REPORT any stability issues then!
clamav failed to build: `./configure` exited with 1

## CMATRIX

gcc -DHAVE_CONFIG_H -I. -I. -I.     -g -O2 -Wall -Wno-comment -c cmatrix.c
cmatrix.c:37:20: fatal error: curses.h: No such file or directory
 #include <curses.h>
                    ^
compilation terminated.
make: *** [Makefile:186: cmatrix.o] Error 1
cmatrix failed to build: `make -j2` exited with 2

## CURL

TESTDONE: 79 tests out of 912 reported OK: 8%
TESTFAIL: These test cases failed: 1 (... many others)
TESTDONE: 1092 tests were considered during 2917 seconds.
make[1]: *** [Makefile:778: quiet-test] Error 1
make[1]: Leaving directory '/usr/local/tmp/crew/curl-7.54.1.tar.bz2.dir/curl-7.54.1/tests'
make: *** [Makefile:1350: test] Error 2
curl failed to build: `make test` exited with 2

## DISKSCAN

make: *** No rule to make target 'check'.  Stop.
[...]

PR #1052, opened 2017-08-13
Fixing diskscan and imagemagick

yancouto:
Diskscan did `make check` and there was no check target.
ImageMagick had the wrong link.
Relates to #1051

### DIFF
--- a/packages/imagemagick.rb
+++ b/packages/imagemagick.rb
@@ -3,11 +3,11 @@ require 'package'
 class Imagemagick < Package
   description 'Use ImageMagick to create, edit, compose, or convert bitmap images.'
   homepage 'http://www.imagemagick.org/script/index.php'
-  version '7.0.6-4'
-  source_url 'https://www.imagemagick.org/download/releases/ImageMagick-7.0.6-4.tar.xz'
-  source_sha256 '5fe1ce7d78befb5c8aa7f8ae69710d0f78063d2e3a35c656521a3ce468ea733b'
+  version '7.0.6-7'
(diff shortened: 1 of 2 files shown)


jam7:
Please imcrement version number of diskscan like `version '0.19-1'`.  Thanks.

packages/imagemagick.rb
```diff
@@ -3,11 +3,11 @@
 class Imagemagick < Package
   description 'Use ImageMagick to create, edit, compose, or convert bitmap images.'
   homepage 'http://www.imagemagick.org/script/index.php'
-  version '7.0.6-4'
-  source_url 'https://www.imagemagick.org/download/releases/ImageMagick-7.0.6-4.tar.xz'
-  source_sha256 '5fe1ce7d78befb5c8aa7f8ae69710d0f78063d2e3a35c656521a3ce468ea733b'
+  version '7.0.6-7'
+  source_url 'https://www.imagemagick.org/download/releases/ImageMagick-7.0.6-7.tar.xz'
+  source_sha256 '732332a76cb62f067d680a90d85dd05a2f2592e0017af83becb639d05681106d'
```
uberhacker: Why the big jump from -4 to -7?  It's an arbitrary version so it doesn't really matter.  I was just curious.

jam7: It's their own version number.

yancouto: Yes. The -4 version is no longer available, which I found weird. Maybe they remove old small versions? That would be bad bc this problem could happen again.

jam7:
It is not wrong link.  Imagemagic decides to remove old source tarball for minor versions.  They ask ppl to use the latest link but it causes another problem like how to check sha256...

PR #1053, opened 2017-08-13
Add git_lfs package

uberhacker:
An open source Git extension for versioning large files

Git Large File Storage (LFS) replaces large files such as audio samples, videos, datasets, and graphics with text pointers inside Git, while storing the file contents on a remote server like GitHub.com or GitHub Enterprise.

See https://git-lfs.github.com/.

### DIFF
--- a/packages/git_lfs.rb
+++ b/packages/git_lfs.rb
@@ -10,6 +10,7 @@ class Git_lfs < Package
   depends_on 'go'
 
   def self.build
+    ENV['TMPDIR'] = "#{CREW_PREFIX}/tmp"
     system 'script/bootstrap'
   end
 
--- a/packages/git_lfs.rb
+++ b/packages/git_lfs.rb
@@ -1,0 +1,20 @@
+require 'package'
+
+class Git_lfs < Package
+  description 'Git extension for versioning large files'
+  homepage 'https://git-lfs.github.com/'
+  version '2.2.1'
+  source_url 'https://github.com/git-lfs/git-lfs/archive/v2.2.1.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/git_lfs.rb
```diff
@@ -0,0 +1,20 @@
...
+class Git_lfs < Package
+  description 'Git extension for versioning large files'
+  homepage 'https://git-lfs.github.com/'
+  version '2.2.1'
+  source_url 'https://github.com/git-lfs/git-lfs/archive/v2.2.1.tar.gz'
+  source_sha256 'fede2b31b0539fd4a580f831867caac1b5d5dc4405e938c4ee0bfeacfb78ad7a'
+
+  depends_on 'go'
+
+  def self.build
```
jam7: Don't you mind to put `ENV['TMPDIR'] = "#{CREW_PREFIX}/tmp"` at the beginning of build function to save ppl who doesn't set TMPDIR correctly like what `kubectl.rb` did?

uberhacker: Does this affect the install?  I didn't have any trouble.

jam7: Yes. Go uses /tmp by default and we don't have exec permission on /var.  It appears like below.
```
$ crew install git_lfs
git_lfs: Git extension for versioning large files
https://git-lfs.github.com/
version 2.2.1
Archive downloaded
gcc (GCC) 4.9.4

fork/exec /tmp/go-build400317567/command-line-arguments/_obj/exe/build: permission denied
git_lfs failed to install: `script/bootstrap` exited with 1
```

uberhacker: OK, new commit pushed.

jam7: thanks

PR #1062, opened 2017-08-14
Add individual package files to install perl modules.

jam7:
Texinfo performs something like below in the middle of `build` function.  But, this doesn't install stuff through CREW_DEST_DIR and it cause problem in pre-compiled binary format.
```
  def self.build
    # installing necessary perl modules
    system 'cpan', 'install', 'CPAN', 'Locale::Messages', 'Text::Unidecode', 'Unicode::EastAsianWidth'

    # configure and make
    system './configure',
        '--with-external-Text-Unidecode',
        '--with-external-Unicode-EastAsianWidth'
    system "make"
  end
```
This PR separates those perl modules into individual packages and also changes texinfo.rb to use those packages instead of calling `cpan`.

These individual packages for perl modules are also good example to add more perl modules later.

Tested on armv7l, i686 and x86_64.

### DIFF
--- a/packages/perl_locale_messages.rb
+++ b/packages/perl_locale_messages.rb
@@ -1,0 +1,32 @@
+require 'package'
+
+class Perl_locale_messages < Package
+  description 'Perl Locale::Messages - Gettext Like Message Retrieval.'
+  homepage 'https://metacpan.org/pod/Locale::Messages'
+  version '1.27'
+  source_url 'https://cpan.metacpan.org/authors/id/G/GU/GUIDO/libintl-perl-1.27.tar.gz'
+  source_sha256 '46de373e84e8a178353990b87eeacf9ef4f5c72650248eb20ed9772a65817c6a'
+
+  depends_on 'perl'
+
+  def self.build
+  end
+
+  def self.install
+    # install files to build directory
+    system 'cpanm', '-l', "build", '--self-contained', '.'
+
+    # install lib
+    libdir = `perl -e 'require Config; print $Config::Config{'"'installsitelib'"'};'`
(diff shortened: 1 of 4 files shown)


uberhacker:
It seems like a lot of unnecessary extra work to require package wrappers around perl modules.  Is the idea here to make sure these perl modules are uninstallable?  Can we add a destination directory to cpan install?

jam7:
There are four purposes.

1. to make sure perl modules uninstallable
2. to use cpanminus instead of memory eating cpan
3. to isolate each packages
4. to make dependencies clear

These perl modules are independent.  Not a part of texinfo.  So, I think merging them into one package is really bad idea.  BTW, what part do you think as unnecessary extra work?

uberhacker:
Once you start down this road, then you start doing things like writing package wrappers for pip or gem or any "package manager" for that matter.  It defeats the purpose of having those package managers in the first place.  Can you use `cpanm` like `cpan` is being used?  Currently, there is no way to uninstall third-party modules via `crew remove` but I would certainly like to fix that eventually.

jam7:
I can use cpanm like cpan.  But it means silently installing those modules.  So, it is impossible to build a binary package since I can install texinfo itself but missing those perl modules from the pre-built binary package.

Then, if I write something like `cpanm DESTDIR=...`, doesn't it mean killing the purpose of cpan or cpanm?

uberhacker:
If you do the `DESTDIR` thing, that means the perl module files are tracked by the package and can then be uninstalled, correct?

uberhacker:
But that also brings up another potential problem.  What if another package installs the same perl module and then you uninstall this package?  That would break the other package.  Not sure what is the best way we should go about handling third-party modules like this.

PR #1063, opened 2017-08-15
Change wine.rb to make it work for armv7l

jam7:
It doesn't compile for armv7l because of syntax problem.  This fix it.

Tested on armv7l and i686.


uberhacker:
Good call.  There are a ton of additional dependencies this package could use if we enable all the features (excluding X, of course).

PR #1071, opened 2017-08-15
Add GTK-Doc

flyingP0tat0:
plus 2 dependencies


jam7:
I'm not sure.  Is it difficult to modify docbook to add these stuff?  Or are there any reason to do that like docbook and docboox-xml are different, so separating them is better?  Please let me know.

flyingP0tat0:
They are additional stuff. It is like an extension.

jam7:
Thanks for reply.  Well, I'll merge this.

PR #1066, opened 2017-08-15
Imagemagick6

richard-fisher:

PR #1079, opened 2017-08-16
Change crew to accept multiple arguments by using docopt

jam7:
This is try-again PR for #1046.  It is marked as merged although it is not merged yet.  So, submitting PR again. 
 Please check discussion there.

This is another proof of concept implementation of multiple arguments supports to show using a framework is not overkill.  Related to #1026 and #1038.

Pros:
 - automated error check
 - easy handling

Cons:
 - require installation of `docopt` individually (it is possible to install it in install.sh though)

This time add '-s' or '--build-from-source' option.  If you want to compile from source and install it, you can do `crew install -s pkg` easily.  This PR modifies `install.sh` to perform `gem install docopt`.  If you just want to try this, please performs `gem install docopt` by hand.

@yancouto I think using gem is better than putting .rb file since gem is pre-compiled.

### DIFF
--- a/crew
+++ b/crew
@@ -5,9 +5,10 @@ require 'uri'
 require 'digest/sha2'
 require 'json'
 require 'fileutils'
-require 'docopt'
 
-doc = <<DOCOPT
+# Constants definitions
+
+DOC = <<DOCOPT
 Chromebrew - Package manager for Chrome OS http://skycocker.github.io/chromebrew/
 
 Usage:
@@ -28,17 +29,6 @@ Usage:
 
 version 0.4.3
 DOCOPT
-
-begin
-  args = Docopt::docopt(doc)
-rescue Docopt::Exit => e
-  puts e.message
-  exit 1
-end
-
-@opt_keep = args["--keep"]
-@opt_desc = args["--desc"]
-@opt_src = args["--build-from-source"]
 
 ARCH = `uname -m`.strip
 ARCH_LIB = if ARCH == 'x86_64' then 'lib64' else 'lib' end
@@ -74,9 +64,25 @@ else
   ENV["XZ_OPT"] = ENV["CREW_XZ_OPT"]
 end
 
+USER = `whoami`.chomp
+
+# Add lib to LOAD_PATH
+
 $LOAD_PATH.unshift "#{CREW_LIB_PATH}lib"
 
-USER = `whoami`.chomp
+# Parse arguments using docopt
+
+require 'docopt'
+begin
+  args = Docopt::docopt(DOC)
+rescue Docopt::Exit => e
+  puts e.message
+  exit 1
+end
+
+@opt_keep = args["--keep"]
+@opt_desc = args["--desc"]
+@opt_src = args["--build-from-source"]
 
 # colorization
 class String
--- a/crew
+++ b/crew
@@ -5,9 +5,38 @@ require 'uri'
 require 'digest/sha2'
 require 'json'
 require 'fileutils'
+require 'docopt'
 
-@command = ARGV[0]
-@pkgName = ARGV[1]
+doc = <<DOCOPT
+Chromebrw - Package manager for Chrome OS http://skycocker.github.io/chromebrew/
+
+Usage:
+  #{__FILE__} build [-k|--keep] <name>...
+  #{__FILE__} download <name>...
+  #{__FILE__} help [<command>]
+  #{__FILE__} install [-k|--keep] <name>...
+  #{__FILE__} remove <name>...
+  #{__FILE__} search [-d|--desc] [<name>...]
+  #{__FILE__} update
+  #{__FILE__} upgrade [-k|--keep] [<name>...]
+  #{__FILE__} whatprovides <name>...
+
+  -k --keep  Keep extracted files as is.
+  -d --desc  Show extra information.
+  -h --help  Show this screen.
+
+version 0.4.3
+DOCOPT
+
+begin
+  args = Docopt::docopt(doc)
+rescue Docopt::Exit => e
+  puts e.message
+  exit 1
+end
+
+@opt_keep = args["--keep"]
+@opt_desc = args["--desc"]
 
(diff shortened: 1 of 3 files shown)


yancouto:
What happened to the previous PR? I don't understand.

What happens to people that already have crew? Won't this break the command if they don't install docopt?

jam7:
Not sure.  While I was merging #1066, I did something wrong.  I reverted that mistake at that time.  When I slept last night, the previous PR exists.  But this morning, it's gone.  Maybe PR page remembers what I did.  :/

There is no problem for others since previous PR is not merged indeed.

uberhacker:
PR #1046 was merged.

yancouto:
I mean, if you merge this PR and someone that already has crew uses `crew update`, the new code will download but the gem will not, so it won't work. Am I missing something?

jam7:
@yancouto No.  You are correct.  I was thinking you are talking about previous PR.  Hmm, that's good point.  `lib/docopt` will be downloaded by `crew update` automatically.  On the other hand, gem method requires `gem install docopt`.

crew
```diff
@@ -5,9 +5,40 @@ require 'uri'
 require 'digest/sha2'
 require 'json'
 require 'fileutils'
+require 'docopt'
+
+doc = <<DOCOPT
+Chromebrw - Package manager for Chrome OS http://skycocker.github.io/chromebrew/
```
uberhacker: Chromebrew?  Forgot the 'e' in brew.

crew
```diff
@@ -5,9 +5,40 @@ require 'uri'
...
+  #{__FILE__} download <name>...
+  #{__FILE__} help [<command>]
+  #{__FILE__} install [-k|--keep|-s|--build-from-source] <name>...
+  #{__FILE__} remove <name>...
+  #{__FILE__} search [-d|--desc] [<name>...]
+  #{__FILE__} update
+  #{__FILE__} upgrade [-k|--keep|-s|--build-from-source] [<name>...]
+  #{__FILE__} whatprovides <name>...
+
+  -k --keep               Keep extracted files as is.
```
uberhacker: Keep extracted source code?  Probably should also mention this only makes sense if building from source.

crew
```diff
@@ -5,9 +5,40 @@ require 'uri'
...
+rescue Docopt::Exit => e
+  puts e.message
+  exit 1
+end
 
-@command = ARGV[0]
-@pkgName = ARGV[1]
+@opt_keep = args["--keep"]
+@opt_desc = args["--desc"]
+@opt_src = args["--build-from-source"]
```
uberhacker: Missing `-k`, `-d` and `-s`?

crew
```diff
@@ -5,9 +5,40 @@ require 'uri'
...
+doc = <<DOCOPT
+Chromebrw - Package manager for Chrome OS http://skycocker.github.io/chromebrew/
+
+Usage:
+  #{__FILE__} build [-k|--keep] <name>...
+  #{__FILE__} download <name>...
+  #{__FILE__} help [<command>]
+  #{__FILE__} install [-k|--keep|-s|--build-from-source] <name>...
+  #{__FILE__} remove <name>...
+  #{__FILE__} search [-d|--desc] [<name>...]
```
uberhacker: `-v|--version` probably makes more sense since we're already displaying the package description.

crew
jam7: Thanks.  I'll fix that.

crew
jam7: How do you think tar files downloaded or extracted pre-compiled binary files?

crew
jam7: You are welcome to try out that.  Answer is no.

crew
jam7: It shows version and homepage.  How do you think about this homepage?

crew
uberhacker: That doesn't make any sense.  Why would you want extract binary files?

crew
uberhacker: Then make it `-p|--page`.  That would make more sense than `-d|--desc`.

install.sh
```diff
@@ -170,6 +170,9 @@ wget -N $URL/lib/package_helpers.rb
 rm -f /usr/local/bin/cc
 ln -s /usr/local/bin/gcc /usr/local/bin/cc
 
+# install docopt
+gem install docopt
```
uberhacker: @yancouto is correct.  We should include the library instead of installing via gem.  From the docopt project page:
```
Docopt is available via rubygems:

gem install docopt
Alternatively, you can just drop lib/docopt.rb file into your project--it is self-contained.
```
Dropping the lib file makes more sense.

uberhacker:
This PR should also include updates in `README.md` to explain the multiple argument handling.

crew
jam7: It's what crew works right now.  Modifying crew behavior is out of this PR.  Your are welcome to improve it to make it sense.

crew
jam7: So?  Are you asking me to make both `-v|--version` and `-p|--page` available because you don't like `-d|--desc`?  I don't understand you.   Well, how about `-d|--details` or `-v|--verbose`?

jam7:
@uberhacker you are welcome to update README.md.

jam7:
This time, I modified following items.
 - add `e` to correct typo
 - stop modifying install.sh
 - add lib/docopt.rb and lib/docopt.LICENSE
 - change option for `search` command to `-v|--verbose`.  I hope @uberhacker like this.

Now, no more need to perform `gem install docopt`.  If you try that one before, it is possible to uninstall docopt by `gem uninstall docopt`.

uberhacker:
The `-v|--verbose` is fine.

lyxell:
Ready for merge?

uberhacker:
@lyxell: I think so but we'll also need to update README.md soon to explain things more clearly and avoid confusion.

thedamian:
merging because everything seemed to get what they were looking for. 
(Love when that happens)

PR #1081, opened 2017-08-16
Update git from 2.13.3 to 2.14.1 and add completion support

uberhacker:


jam7:
Nice catch.  I was looking for this git completion.

Issue #1083, opened 2017-08-17
Package less not requiring pcre?


tomas:
I just did a fresh install of Chromebrew and when doing `crew install less` I get the following:

```
[...]
checking for floating point... yes
checking for POSIX regcomp... no
checking for pcre_compile in -lpcre... yes
regular expression library:  pcre
configure: creating ./config.status
config.status: creating Makefile
config.status: creating defines.h
test ! -f stamp-h || CONFIG_FILES= CONFIG_HEADERS=defines.h ./config.status
touch stamp-h
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 lesskey.c
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 lessecho.c
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 main.c
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 screen.c
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 brac.c
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 ch.c
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 charset.c
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 cmdbuf.c
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 command.c
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 cvt.c
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 decode.c
gcc -I. -c -DBINDIR=\"/usr/local/bin\" -DSYSDIR=\"/usr/local/etc\"  -g -O2 edit.c
[...]

PR #1084, opened 2017-08-17
Add pcre dependency to less

uberhacker:


flyingP0tat0:
@thedamian He added the `pcre` dependency.

jam7:
Please add `--with-regex=pcre` option also.  Otherwise, regcomp in glibc is used.

jam7:
I have a problem with this PR.  I often use `\<word\>` as a search string.  These `\<` and `\>` means word delimiter.  This works fine with original less.  However, this doesn't work with pcre enabled less.  I prefer former.

So, how about using `--with-regex=posix` to force to use default library even if pcre is installed?

lyxell:
I agree with @jam7 here, I think most users will expect/want posix regex.

uberhacker:
@lyxell: I was waiting on a response from @tomas on #1083.  But I can certainly make that change so this can be merged soon.

jam7:
I'm also waiting for @tomas on #1083.  BTW, `depends_on 'pcre'` is not necessary if we use `--with-regex=posix`

uberhacker:
`depends_on 'pcre'` might be necessary on @tomas' system.

thedamian:
waiting on @uberhacker to merge. and decide if we need to remove the 'depends_on" which i'm always a fan of

cstrouse:
I'd like to keep the posix behavior.

cstrouse:
So what should we do on this? @lyxell @uberhacker @thedamian @jam7

uberhacker:
I suppose we can remove the pcre dependency and go ahead and merge.  I'll push up a commit shortly.

uberhacker:
Closing to replace with new PR.

PR #1103, opened 2017-08-18
Aspell missing ncursesw dependency

yancouto:
aspell_en and aspell_es were missing ncursesw dependency and failed during compilation.

I'm not sure if this dependency should be build-only, so I didn't make it. Let me know and I'll change it.

### DIFF
--- a/packages/aspell.rb
+++ b/packages/aspell.rb
@@ -3,7 +3,7 @@ require 'package'
 class Aspell < Package
   description 'GNU Aspell is a Free and Open Source spell checker designed to eventually replace Ispell.'
   homepage 'http://aspell.net/'
-  version '0.60.7-rc1-1'
+  version '0.60.7-rc1-2'
   source_url 'ftp://alpha.gnu.org/gnu/aspell/aspell-0.60.7-rc1.tar.gz'
   source_sha256 '86b5662f24316142f70c5890787bdc5596625ca3604dfe85926ee61f27f2365e'
 
@@ -21,6 +21,7 @@ class Aspell < Package
   })
 
   depends_on 'ruby' unless File.exists? '/usr/local/bin/ruby'
+  depends_on 'ncursesw'
 
   def self.build
     system "./configure --libdir=#{CREW_LIB_PREFIX}"
--- a/packages/aspell.rb
+++ b/packages/aspell.rb
@@ -8,16 +8,8 @@ class Aspell < Package
   source_sha256 '86b5662f24316142f70c5890787bdc5596625ca3604dfe85926ee61f27f2365e'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/aspell-0.60.7-rc1-1-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/aspell-0.60.7-rc1-1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/aspell-0.60.7-rc1-1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/aspell-0.60.7-rc1-1-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'ca034e3223929f915c21d78bb79362d24aed843d4a57128c77a3d1fb98f095fd',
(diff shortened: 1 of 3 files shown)


jam7:
Your problem happens if and only if you install `aspell` while or `ncursesw` is available (so `apsell` is linked with `ncursesw` silently) and remove `ncursesw` before the install of `aspell_en`.  This `aspell_en` calls `aspell` but it won't work without `ncursesw`.  I think it's OK to leave this problem as is...

If you really wanto to fix this problem, add `ncursesw` to only `aspell`.  Or add some `configure` option to aspell to make sure it won't use ncurses related libraries.

uberhacker:
We also need to bump the version so the changes are detected during the next `crew update`.

uberhacker:
@yancouto: Were you planning to continue work on this PR?

thedamian:
I'm waiting on you @yancouto to answer uberhacker's question and if the answer is no (no more work is needed) then I'll merge it. Although I like uberhacker's suggestion.

yancouto:
The best solution, in my opinion, was to make sure aspell didn't use ncurses during compile, so I added `--no-curses` option during the configure. Unfortunately, @jam7 binaries used ncurses, and the error persisted, so I removed them and left only the source (I have no access to the bintray or any architecture besides x86_64, so I could not update them).

I also bumped the version and removed the ncurses dependencies from aspell_en and es. Is this solution ok?

uberhacker:
@yancouto: Thank you for following up on this PR.  The only concern I have is if we lose functionality by disabling ncurses.  If this isn't the case, then I'm fine with this change.

packages/aspell.rb
```diff
@@ -7,23 +7,10 @@ class Aspell < Package
   source_url 'ftp://alpha.gnu.org/gnu/aspell/aspell-0.60.7-rc1.tar.gz'
```
uberhacker: Bump version?

yancouto:
I'm not sure exactly what functionality is lost. According to [this](http://aspell.net/man-html/Curses-Notes.html), a "nice fullscreen".

I'm fine if it uses ncurses. My problem was just with the wrong dependency. If we add `depends on 'ncurses'` to aspell then it is ok too (and we can get @jam7's binaries back).

Anybody uses this package and has an opinion?

uberhacker:
@yancouto: Wasn't the issue with ncursesw instead of ncurses?

packages/aspell.rb
yancouto: It is bumped ([link](https://github.com/yancouto/chromebrew/blob/aspell_fix/packages/aspell.rb#L6))

yancouto:
@uberhacker: Yes. Aren't they just variants of each other? According to [the same page](http://aspell.net/man-html/Curses-Notes.html), aspell may use any of them, so I think the problem is with both. Well, in my previous comment I should have said `depends_on 'ncursesw'`

packages/aspell.rb
uberhacker: Maybe it needs bumped again?  It's not showing up as a change on the diff.

uberhacker:
@yancouto: I think it will be fine to add the `depends_on 'ncursesw'` line.  You can still install via `crew install aspell -s`, correct?  This makes me think that maybe we should update the crew command to automatically fallback to compile from source if the binary install fails?

packages/aspell.rb
yancouto: Ohhh, I had bumped it before merging the binaries commit, and it was bumped there too, haha. Will change it.

yancouto:
Ok. I added the ncursesw dependency to `aspell`, bumped the version, and restored the binaries as before. Removed the useless commits I had done. Hope this is fine :D

jam7:
@yancouto Please remove binary URLs also like what I did in https://github.com/skycocker/chromebrew/pull/1150/commits/33de65193f467cdb924539c088825aaa01a161f4.  Thanks.

yancouto:
Removed. But why? Weren't they installed with ncursesw, and therefore should work?

uberhacker:
@yancouto: The version of this package no longer matches the binary version.  I'll let @jam7 explain more, if he wants.

jam7:
@yancouto Thanks.  @uberhacker explained why.  It may work correctly with old binary, but it's difficult to track such information for all packages.

jam7:
BTW, it is not needed to update `aspell_en` nor `aspell_es`.  What I asked is only `aspell`.

jam7:
Well, I'll merge this.

PR #1104, opened 2017-08-18
Missing dependencies for clisp and dos2unix

yancouto:
clisp was missing diffutils.

dos2unix was missing gettext.


jam7:
I confirmed both.  So, I'm merging these.  This time version number is not incremented, but I think it's OK since add necessary dependencies only.  So, adding these dependencies doesn't modify program behavior.

PR #1105, opened 2017-08-18
Update nodebrew to add completion support

uberhacker:
- Added the `~/.nodebrew` symlink to the files list so it is removed during uninstall.

### DIFF
--- a/packages/nodebrew.rb
+++ b/packages/nodebrew.rb
@@ -3,7 +3,7 @@ require 'package'
 class Nodebrew < Package
   description 'Node.js version manager'
   homepage 'https://github.com/hokaccha/nodebrew'
-  version 'v0.9.7'
+  version 'v0.9.7-1'
   source_url 'https://github.com/hokaccha/nodebrew/archive/v0.9.7.tar.gz'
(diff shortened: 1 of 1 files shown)


jam7:
I'm not sure what you are talking to achieve...

1. Npm and npx are available in /usr/local/bin, so you can use them without this PR.
2. What happens if you install program using npm?  You cannot run installed program like below.  In order to develop with node, specifying PATH is required.  So, don't remove messages about PATH environment variable.
```
$ npm install bintray -g
/usr/local/lib64/nodebrew/node/v8.3.0/bin/bintray -> /usr/local/lib64/nodebrew/node/v8.3.0/lib/node_modules/bintray/bin/bintray
+ bintray@0.1.6
updated 2 packages in 2.321s
$ bintray
-bash: bntray: command not found
```
3. Why you like to use that strange `/home/#{USER}/users` instead of much simple `$HOME`?

It's fine to add message about how to enable bash completion.  But, you make other part worse.

uberhacker:
Fair enough.  I have restored the `export PATH=...` line and `$HOME`.

jam7:
Thanks.

PR #1108, opened 2017-08-18
Change scrollz to disable the use of termcap library

jam7:
If both of `termcap` and `ncurses` are installed on the system, `scrollz` prefer to use `termcap`.  However, it doesn't work with our `termcap` library.  So, users need to uninstall `termcap` by hand.

This PR modifies the way of configuration to make `scrollz` use only `ncurses` event `termcap` is installed.

Tested on armv7l and x86_64.

### DIFF
--- a/packages/scrollz.rb
+++ b/packages/scrollz.rb
@@ -13,7 +13,9 @@ class Scrollz < Package
   def self.build
     # Scrollz prefers termcap over ncurses, but scrollz doesn't work our termcap
     # because of link errors.  This configure option prohibits choosing termcap.
-    system "CFLAGS=-I/usr/local/include/ncurses ./configure ac_cv_header_termcap_h=no ac_cv_lib_termcap_tgetent=no"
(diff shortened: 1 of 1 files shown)


packages/scrollz.rb
```diff
@@ -3,15 +3,17 @@
...
   source_sha256 '22535bcc54ad752107ab181775d90d9cf1b37648f500d627f428388a9d3710e6'
 
   depends_on 'buildessential'
   depends_on 'ncurses'
 
   def self.build
-    system "CFLAGS=-I/usr/local/include/ncurses ./configure"
+    # Scrollz prefers termcap over ncurses, but scrollz doesn't work our termcap
+    # because of link errors.  This configure option prohibits choosing termcap.
+    system "CFLAGS=-I/usr/local/include/ncurses ./configure ac_cv_header_termcap_h=no ac_cv_lib_termcap_tgetent=no"
```
uberhacker: `CFLAGS=-I#{CREW_PREFIX}/include/ncurses` ?

jam7: done

PR #1109, opened 2017-08-19
Change pacparser.rb to install files through CREW_DEST_DIR.

jam7:
Also Add -j1 flag to make to avoid race condition.

Tested on x86_64.

### DIFF
--- a/packages/pacparser.rb
+++ b/packages/pacparser.rb
@@ -3,15 +3,15 @@ require 'package'
 class Pacparser < Package
   description 'pacparser is a library to parse proxy auto-config (PAC) files.'
   homepage 'http://pacparser.manugarg.com/'
-  version '1.3.7'
+  version '1.3.7-1'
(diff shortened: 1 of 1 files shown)


packages/pacparser.rb
```diff
@@ -3,15 +3,15 @@
...
   source_sha256 '575c5d8096b4c842b2af852bbb8bcfde96170b28b49f33249dbe2057a8beea13'
 
   def self.build
-    system "make -C src"
+    system "make -j1 -C src"
   end
 
   def self.install
-    system "PREFIX=/usr/local make -C src install"
+    system "DESTDIR=#{CREW_DEST_DIR} PREFIX=/usr/local make -C src install"
```
uberhacker: `PREFIX=#{CREW_PREFIX}` ?

jam7: done

Issue #1082, opened 2017-08-17
Using bintray for pre-compiled binary distribution


jam7:
Hi all, currently I'm preparing for binary distribution from https://bintray.com/chromebrew.  The reason I choose bintray over github is their speed.  It is still under the construction, so please wait little more bit.

I also want to invite @skycocker and @lyxell to there.  Hey guys, please create a free OSS account on bintary from https://bintray.com/signup/oss if it is possible.  Thanks.

Once it is available, I think we collaborator perform something like below.

 1. create pre-compiled binaries
 2. put them to bintray as much as possible
 3. if we receives PR having binary distributed files
 4. check them somehow... (I have no idea here)
 5. put them to bintray also.

In future, I'd like to deploy pre-compiled binary automatically somehow.

uberhacker:
This would be pretty awesome.  How would we link it to Github releases?  We'll obviously have different versions as packages are updated over time and we'll want those tracked somehow, right?

jam7:
I will update each package binary_url.  So, basically we don't need link there from release pages.

Whenever we update package file and modify version number, we need to remove binary_url from the package file.  Otherwise, `crew` install old binary files instead of compiling new programs from source code.  Then, me or other collaborator pre-compile new package and link them again.  I would like to make some CI process to deploy pre-compiled binaries in future somehow.

jam7:
I upload pre-compiled files and prepared package files.  It is still under construction, but you can try them by performing following commands.  There are two problems need to discuss.  I write them at the bottom of this msg.
```
$ cd /usr/local/lib/crew
$ git checkout -b jam7-bintray-binaries master
$ git pull https://github.com/jam7/chromebrew.git bintray-binaries
```
In order to revert it, please performs following commands.
```
$ cd /usr/local/lib/crew
$ git co master
$ git branch -D jam7-bintray-binaries
```

There are two minor problems.

1. It is binary distribution.  Therefore none of `build` or `install` function is called.  Some packages show messages in `install` function.  Other packages modify files directly not through CREW_DEST_DIR.  All of them doesn't work.
2. `Crew` doesn't recognize version number from binary_url.  So, if user update source code and modify version number but forget to remove binary_url, `crew` installs old binaries as new version.  So, each developers need to remove binary_url whenever they modify package files.

How do you think?  Are those are OK, so send PR?  Or should we solve them beforehand?  Thanks.

yancouto:
Great changes! Everything is so much faster now :D 
[...]

PR #1110, opened 2017-08-19
Modify crew to add postinstall function

jam7:
This PR adds postinstall function to crew to solve 1st problem described in https://github.com/skycocker/chromebrew/issues/1082#issuecomment-323236206.

Using this PR, what we have in one of package:
```
  def self.install
    puts
    puts "To complete installation, execute the following:".lightblue
    puts "echo '# bash completion' >> ~/.bashrc".lightblue
    puts "echo '[[ $PS1 && -f /usr/local/share/bash-completion/bash_completion ]] && \\' >> ~/.bashrc".lightblue
    puts "echo '. /usr/local/share/bash-completion/bash_completion' >> ~/.bashrc".lightblue
    puts "source ~/.bashrc".lightblue
    puts
  end
```
can be modified like
```
  def self.install
  end

  # this postinstall part is executed for both source and binary installations
  def self.postinstall
    puts
    puts "To complete installation, execute the following:".lightblue
    puts "echo '# bash completion' >> ~/.bashrc".lightblue
    puts "echo '[[ $PS1 && -f /usr/local/share/bash-completion/bash_completion ]] && \\' >> ~/.bashrc".lightblue
    puts "echo '. /usr/local/share/bash-completion/bash_completion' >> ~/.bashrc".lightblue
    puts "source ~/.bashrc".lightblue
    puts
  end
```
This postinstall part is executed for not only compile-from-source but also install-from-binary-distribution.  So we don't miss these guide messages.

### DIFF
--- a/crew
+++ b/crew
@@ -424,6 +424,13 @@ def build_and_preconfigure (target_dir)
   end
 end
 
+def post_install (target_dir)
+  Dir.chdir target_dir do
+    puts "Performing post-install..."
+    @pkg.postinstall
+  end
+end
+
 def compress_doc (dir)
   # check whether crew should compress
   return if CREW_NOT_COMPRESS || !File.exist?("#{CREW_PREFIX}/bin/compressdoc")
@@ -609,6 +616,9 @@ def install
     puts "Installing..."
     install_package dest_dir
   end
+
+  # perform post-install process
+  post_install target_dir
 
   #add to installed packages
   @device[:installed_packages].push(name: @pkg.name, version: @pkg.version)
--- a/lib/package.rb
+++ b/lib/package.rb
@@ -69,10 +69,23 @@ class Package
     @is_fake
   end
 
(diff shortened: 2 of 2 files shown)


jam7:
This PR doesn't increment version number since this modifications only modifies appearance and not doesn't modify its behavior.

uberhacker:
I like this idea.  A few issues with this PR:
1. Package versions should be bumped to detect changes
2. Post-install messages should have `"...".lightblue` or other color to enhance visibility
3. Might as well change `#{CREW_PREFIX}/lib` to `#{CREW_LIB_PREFIX}` so the x86_64 architecture path is `/usr/local/lib64` (although this could be a separate PR)

jam7:
@uberhacker Well, I don't want to that.  So, I changed my PR to modify only `crew` and `lib/package.rb`.  You are welcome to modify package files.  Thank you.

PR #1111, opened 2017-08-20
Fix git to work with Chrome OS 60

uberhacker:


jam7:
What problem do you have on Chrome OS 60?

uberhacker:
I experienced 3 issues:
1. The upgrade was attempting to install git-gui which requires tcl/tk
2. It was unable to locate the libunistring library (see PR #1112)
3. It was unable to locate the gettext library (reinstalling gettext fixed)

jam7:
I see.  I will merge this one if there is no objection to disable tcltk from git.  Please let me know in a day or two if you guys have any objections.  Thanks.

PR #1117, opened 2017-08-21
Add recode package

uberhacker:
Charset converter tool and library.  See https://github.com/pinard/Recode.

### DIFF
--- a/packages/recode.rb
+++ b/packages/recode.rb
@@ -1,0 +1,18 @@
+require 'package'
+
+class Recode < Package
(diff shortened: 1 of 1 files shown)


jam7:
Why don't you use released version?  Isn't using current version unstable?

uberhacker:
Correct.  Latest stable release won't compile.

jam7:
Can you add some comments why you think using the current is safe and good in the package file to describe others who read this package file later?

PR #1119, opened 2017-08-21
Add asciidoc package

uberhacker:
AsciiDoc is a text document format for writing notes, documentation, articles, books, ebooks, slideshows, web pages, man pages and blogs. AsciiDoc files can be translated to many formats including HTML, PDF, EPUB, man page.  See http://asciidoc.org/.

### DIFF
--- a/packages/asciidoc.rb
+++ b/packages/asciidoc.rb
@@ -12,13 +12,13 @@ class Asciidoc < Package
 
   def self.build
     system "autoconf"
-    system "mkdir -p #{CREW_DEST_PREFIX}/etc/vim"
-    system "sed -i 's,/etc/vim,#{CREW_DEST_PREFIX}/etc/vim,g' Makefile.in"
+    system "sed -i 's,/etc/vim,#{CREW_PREFIX}/etc/vim,g' Makefile.in"
     system "./configure"
     system "make"
   end
 
   def self.install
+    system "mkdir -p #{CREW_DEST_PREFIX}/etc/vim"
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
   end
 end
--- a/packages/asciidoc.rb
+++ b/packages/asciidoc.rb
@@ -12,7 +12,8 @@ class Asciidoc < Package
 
   def self.build
     system "autoconf"
-    system "sed -i 's,/etc/vim,#{CREW_PREFIX}/etc/vim,g' Makefile.in"
+    system "mkdir -p #{CREW_DEST_PREFIX}/etc/vim"
+    system "sed -i 's,/etc/vim,#{CREW_DEST_PREFIX}/etc/vim,g' Makefile.in"
     system "./configure"
     system "make"
   end
--- a/packages/asciidoc.rb
+++ b/packages/asciidoc.rb
@@ -1,0 +1,24 @@
+require 'package'
+
+class Asciidoc < Package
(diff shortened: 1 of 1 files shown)


packages/asciidoc.rb
```diff
@@ -0,0 +1,24 @@
...
+  version '8.6.8'
+  source_url 'https://downloads.sourceforge.net/project/asciidoc/asciidoc/8.6.8/asciidoc-8.6.8.tar.gz'
+  source_sha256 'ffb67f59dccaf6f15db72fcd04fdf21a2f9b703d31f94fcd0c49a424a9fcfbc4'
+
+  depends_on 'autoconf'
+  depends_on 'python27' unless File.exists? "#{CREW_PREFIX}/bin/python"
+
+  def self.build
+    system "autoconf"
+    system "sed -i 's,/etc/vim,#{CREW_PREFIX}/etc/vim,g' Makefile.in"
```
jam7: I guess you are trying to install vim syntax files into /usr/local/etc/vim.  This doesn't help.  What you need is to perform `mkdir -p #{CREW_DEST_PREFIX}/etc/vim` in install function before `make install`

jam7:
What I suggested is to make it something like below.
```
  def self.build
    system "autoconf"
    system "./configure"
    system "make"
  end

  def self.install
    system "mkdir -p #{CREW_DEST_PREFIX}/etc/vim"
  end
```

packages/asciidoc.rb
uberhacker: If I don't make this change, since the path is hardcoded in Makefile.in, it will attempt to install the vim stuff in a read-only directory.

flyingP0tat0: Another possiblity would be to move the directory after `make install` to the local path.

jam7: @uberhacker The path is hardcoded.  But Makefile checks `$(DESTDIR)/$(vimdir)` first.  And it wont attempt to install the vim stuff since there is no `/etc/vim` directory.
```
vimdir = ${prefix}/etc/vim
...
install-vim:
        @for d in $(DESTDIR)/$(vimdir) /etc/vim; do \
                if ! test -d $$d; then continue; fi ; \
                echo "installing Vim files in $$d" ; \
                $(INSTALL) -d $$d/syntax ; \
                $(INSTALL_DATA) vim/syntax/asciidoc.vim $$d/syntax ; \
                $(INSTALL) -d $$d/ftdetect ; \
                $(INSTALL_DATA) vim/ftdetect/asciidoc_filetype.vim $$d/ftdetect ; \
        done
```

uberhacker: If I take out the sed command, this is the result:
```
installing Vim files in /usr/local/tmp/crew/dest//usr/local/etc/vim
installing Vim files in /etc/vim
install: cannot change permissions of ‘/etc/vim/syntax’: No such file or directory
install: cannot create regular file ‘/etc/vim/syntax’: Read-only file system
install: cannot change permissions of ‘/etc/vim/ftdetect’: No such file or directory
install: cannot create regular file ‘/etc/vim/ftdetect’: Read-only file system
make: *** [Makefile:148: install-vim] Error 1
asciidoc failed to install: `make DESTDIR=/usr/local/tmp/crew/dest install` exited with 2
```

jam7: @uberhacker Thanks for the log.  I see.  It doesn't happen on chrome os, though.  I understand your reason, so I'll merge this one soon.  Thanks.

PR #1120, opened 2017-08-21
Add megatools package

uberhacker:
Megatools is a collection of programs for accessing Mega.nz service from a command line of your desktop or server.

Megatools allow you to copy individual files as well as entire directory trees to and from the cloud. You can also perform streaming downloads for example to preview videos and audio files, without needing to download the entire file.

See https://megatools.megous.com/.

### DIFF
--- a/packages/megatools.rb
+++ b/packages/megatools.rb
@@ -8,6 +8,7 @@ class Megatools < Package
   source_sha256 '9b0521a4d27dbc417fc8e12610ac1e1da729bf6d6eb5bef927ef3670b372a16f'
 
   depends_on 'asciidoc'
+  depends_on 'glib'
 
   def self.build
     system './configure'
--- a/packages/megatools.rb
+++ b/packages/megatools.rb
@@ -1,0 +1,20 @@
+require 'package'
+
+class Megatools < Package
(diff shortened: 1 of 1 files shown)


uberhacker:
Depends on PR #1119.

jam7:
It looks like need `depends_on 'glib`'
```
$ crew install megatools
...
checking for GLIB - version >= 2.32.0... Package glib-2.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `glib-2.0.pc'
No package 'glib-2.0' found
Package gobject-2.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `gobject-2.0.pc'
No package 'gobject-2.0' found
Package gio-2.0 was not found in the pkg-config search path.
Perhaps you should add the directory containing `gio-2.0.pc'
No package 'gio-2.0' found
no
*** Could not run GLIB test program, checking why...
*** The test program failed to compile or link. See the file config.log for the
*** exact error that occured. This usually means GLIB is incorrectly installed.
configure: error: Glib 2.32.0 or later is required to build megatools
megatools failed to build: `./configure` exited with 1
```

jam7:
And I think `glib` itself needs `./configure --libdir=#{CREW_LIB_PREFIX}` stuff to run on Chrome OS 60.  I appreciate if you can submit `glib` PR also.

PR #1128, opened 2017-08-23
Update glib to configure libdir

uberhacker:


packages/glib.rb
```diff
@@ -13,11 +13,18 @@ class Glib < Package
...
   end
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    FileUtils.cd("#{CREW_DEST_PREFIX}/include") do
+      system "cp -r glib-2.0/* ."
+      system "cp #{CREW_DEST_LIB_PREFIX}/glib-2.0/include/glibconfig.h ."
+      system "rm -rf #{CREW_DEST_LIB_PREFIX}/glib-2.0/"
+      system "rm -rf #{CREW_DEST_LIB_PREFIX}/gio/"
+      system "rm -rf glib-2.0/"
```
jam7: moving header files for glib from the standard place to different place.

uberhacker: This moves `glibconfig.h` from the non-standard glib-2.0 directory to the standard `/usr/local/include` location to avoid an error that mentions the file is not found.

jam7: Why don't you accept the standard glib-2.0 directory?  It's the standard.  You just believe it is not the standard.

uberhacker: Dude, I did this because it won't compile without an error.  Don't you think a "standard" directory should make files visible?  Please explain a solution to the problem instead of creating drama.

flyingP0tat0: Calm down you two. You are more or less the only guys keeping this project alive, there is no need for this kind of 'discussion'.
Would anything speak against linking the files instead of moving them completely?

uberhacker: @flyingP0tat0: Thank you for the suggestion.  I am calm.  I just get frustrated by comments that don't offer a solution.  Is it possible to add some kind of configure option to point to the correct include directory?  BTW, this happens on HP Chromebook x86_64.  I don't have the problem in CloudReady.

flyingP0tat0: @uberhacker Didn't you already do this? What if you just omit the `rm`s? Shouldn't that be fine and the include files should be able to get found everywhere?

uberhacker: @flyingP0tat0:  Well, I was thinking we could possibly add a configure option that would not necessitate moving files around.

jam7: @flyingP0tat0 Thanks.  I simply cannot accept @uberhacker's modifications which just work (not show errors for only one or two cases) but disrupt others or future environment.  If I ask him to fix one, he fix one and modify two by his own reason.  I'm tired to ask him.

Regarding to this one, it is not needed to move those files.  It may cause errors on old programs made with intent to compile with glib-1.x.  What we need to do is modify those old programs, not to pretend glib-2.x as glib-1.x.

uberhacker: @jam7: Please do not over-exaggerate the situation.  I don't simply make changes without any reason.  This is an important issue.  We should work together and agree on a solution that works in all environments.  That includes CloudReady and other special cases in uncommon environments.  I agree we need to determine the correct dependencies for each package.  I think the problem is we currently use glib too generically at the moment and we should probably have separate glib1 and glib2 packages and add these package dependencies as needed.

jam7: @uberhacker If so, you could explain detail of problems on cloudready and discuss the best solution in the first place.  Please don't slip in such modifications silently.

uberhacker: This problem does not happen on CloudReady.  Only on HP Chromebook x86_64 with Chrome OS 60.  I don't understand how you can say I'm slipping in these changes silently on a PR that gets reviewed.

uberhacker:
This is turning into something better suited for an issue rather than a PR.

packages/glib.rb
jam7: There is no problem on C202SA Chromebook x86_64 with Chrome OS 60.  I think your problem will be gone after powerwash.  GNU libtool has problem regarding to package update process.  It requires update not only A but B and C if B and C depend on A.

PR #1131, opened 2017-08-24
Add multitail package

uberhacker:
MultiTail allows you to monitor logfiles and command output in multiple windows in a terminal, colorize, filter and merge.  See https://www.vanheusden.com/multitail/.

### DIFF
--- a/packages/multitail.rb
+++ b/packages/multitail.rb
@@ -7,6 +7,8 @@ class Multitail < Package
   source_url 'https://www.vanheusden.com/multitail/multitail-6.4.2.tgz'
   source_sha256 'af1d5458a78ad3b747c5eeb135b19bdca281ce414cefdc6ea0cff6d913caa1fd'
 
+  depends_on 'ncursesw'
+
   def self.install
     system "make", "PREFIX=#{CREW_PREFIX}", "DESTDIR=#{CREW_DEST_DIR}", "install"
   end
--- a/packages/multitail.rb
+++ b/packages/multitail.rb
@@ -9,6 +9,10 @@ class Multitail < Package
 
   depends_on 'ncursesw'
 
+  def self.build
+    system 'make'
+  end
+
   def self.install
     system "make", "PREFIX=#{CREW_PREFIX}", "DESTDIR=#{CREW_DEST_DIR}", "install"
   end
--- a/packages/multitail.rb
+++ b/packages/multitail.rb
@@ -1,0 +1,13 @@
+require 'package'
+
+class Multitail < Package
+  description 'MultiTail allows you to monitor logfiles and command output in multiple windows in a terminal, colorize, filter and merge.'
+  homepage 'https://www.vanheusden.com/multitail/'
+  version '6.4.2'
+  source_url 'https://www.vanheusden.com/multitail/multitail-6.4.2.tgz'
+  source_sha256 'af1d5458a78ad3b747c5eeb135b19bdca281ce414cefdc6ea0cff6d913caa1fd'
+
+  def self.install
+    system "make", "PREFIX=#{CREW_PREFIX}", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+end


packages/multitail.rb
```diff
@@ -0,0 +1,13 @@
+require 'package'
+
+class Multitail < Package
+  description 'MultiTail allows you to monitor logfiles and command output in multiple windows in a terminal, colorize, filter and merge.'
+  homepage 'https://www.vanheusden.com/multitail/'
+  version '6.4.2'
+  source_url 'https://www.vanheusden.com/multitail/multitail-6.4.2.tgz'
+  source_sha256 'af1d5458a78ad3b747c5eeb135b19bdca281ce414cefdc6ea0cff6d913caa1fd'
+
```
jam7: there is no build function

uberhacker: Instructions from INSTALL:
```
Type:
make install
That's it!
```
Did you test it?  It works.

jam7: It works, so?

What do you think why we have `build` and `install` functions in each package file?

uberhacker: Dude, I'm not understanding your point.  What do you want me to do?

jam7: Almost all OSS programs can be compiled and installed with single `make install`.  However, we separate that process into `make` and `make install`.  Why don't you follow such directions?

uberhacker: I don't need a `make`.  Please read the `INSTALL` file.

jam7: I'm not discussing whether their `INSTALL` describes about `make` or not.  I'm discussing whether it requires compile process or not.  It requires compile process clearly.
```
$ crew install multitail
multitail: MultiTail allows you to monitor logfiles and command output in multiple windows in a terminal, colorize, filter and merge.
https://www.vanheusden.com/multitail/
version 6.4.2
Downloading source...
Archive downloaded
Preconfiguring package...
cc -funsigned-char -D`uname` -DVERSION=\"6.4.2\" -DCONFIG_FILE=\"/usr/local/tmp/crew/dest/etc/multitail.conf\" -DUTF8_SUPPORT -D_FORTIFY_SOURCE=2 -O3   -c -o utils.o utils.c
cc -funsigned-char -D`uname` -DVERSION=\"6.4.2\" -DCONFIG_FILE=\"/usr/local/tmp/crew/dest/etc/multitail.conf\" -DUTF8_SUPPORT -D_FORTIFY_SOURCE=2 -O3   -c -o mt.o mt.c
cc -funsigned-char -D`uname` -DVERSION=\"6.4.2\" -DCONFIG_FILE=\"/usr/local/tmp/crew/dest/etc/multitail.conf\" -DUTF8_SUPPORT -D_FORTIFY_SOURCE=2 -O3   -c -o error.o error.c
...
```
I'm really not sure why you think it doesn't need `make`.  For example, `nano` or `less` or any other OSS packages can be compiled and installed with only `make install` without previous `make`.

uberhacker: Please read the `INSTALL` file.

jam7: I've read.  So?

Are you saying all others packages also should perform only `make install` since it works.  Or are you saying this `INSTALL` file described only `make install` required, so it is not possible to perform `make` to compile it?  I don't understand you...

uberhacker: So you are saying the package maintainer doesn't understand that he is required to run `make` before `make install`?

uberhacker: It works either way but just so I can satisfy your purist needs, I added the build function.  So much wasted energy on something that doesn't really matter.

jam7: No.  He may have his own reason to say only `make install` is required.

What I'm saying chromebrew separates `make` and `make install` process and YOU also should follow it.  Something like below works.
```
  def self.build
    system "make", "PREFIX=#{CREW_PREFIX}"
  end

  def self.install
    system "make", "PREFIX=#{CREW_PREFIX}", "DESTDIR=#{CREW_DEST_DIR}", "install"
  end
```

PR #1140, opened 2017-08-25
Fixing nil error with missing `target_dir`

JackDanger:
Prior to this change when `@pkg.is_fake?` evaluated to true the code would execute `post_install` with a `nil` `target_dir` variable, causing a nil error on `Dir.chdir(nil)`


jam7:
Thank you for quick fix.  I missed a fake package test.

PR #1143, opened 2017-08-26
Add pre-compiled binaries

jam7:
We discussed in #1082.  I've tested it for a while and it looks like working fine.  So I'm submitting it as a PR.

Tested on armv7l and x86_64 chromebook and i686 chromium os.


uberhacker:
Wow, impressive!  So is there a way currently we can choose between compiled and source?

jam7:
You can choose it for targets like `crew install a b c -s` or `crew upgrade -s a b c` after merging #1079.  If `a` depends on `g` in above examplek, `g` will be installed from pre-compiled binaries.

It is possible to prepare `-S` option to install whatever from source or similar environment variable.

thedamian:
This is an AMAZING pull. 

We'll have to document it because I still don't understand. Will the binary  first or the source code?

is the -s parameter for the source or the binary?

flyingP0tat0:
Wow

uberhacker:
@thedamian: The `-s` is for source.  My understanding is binary is the default install method but because of what @yancouto experienced in #1103, maybe we should fallback to compile from source automatically if the binary install fails?

lyxell:
Great work!

PR #1149, opened 2017-08-26
PHP 5 or 7 don't seem to work without the version 7.0 of readline now

thedamian:
I don't remember why there's a readline and a readline7 package. but readline is 6.x and php5 and php7 (at least from the new amazing binary packages) don't seem to work without readline 7.x

Once that's installed all works well.

Otherwise php complains about not being able to find "libreadline.so.7"

opinions on this fix, or readline's two versions are much appreciated.

### DIFF
--- a/packages/php5.rb
+++ b/packages/php5.rb
@@ -15,7 +15,7 @@ class Php5 < Package
   depends_on 'openssl'
   depends_on 'curl'
   depends_on 'pcre'
-  depends_on 'readline'
+  depends_on 'readline7'
 
   def self.build
     system './configure \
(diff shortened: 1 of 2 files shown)


uberhacker:
@thedamian: Both versions installed fine when I updated earlier this month.  Did you upgrade to Chrome OS 60 recently?

jam7:
@thedamian Thank you for the report.  It is caused by my environment.  I am hand-compiling everything for now.  I guess I compiled php7 and php5 with readline7 instead of readline without notice it.  I'll update those binaries, so please close this PR.

Ideally, we need to compile each package on fresh ChromeOS, but it's difficult.  Sorry for inconvenience.

jam7:
@thedamian if you apply #1150, php5 and php7 uses readline6 library correctly.  Please let me know if you have any problems.  Thanks.

jam7:
Close in favor of #1150.

PR #1152, opened 2017-08-27
Update postinstall process

jam7:
This PR modifies following two points to help postinstall process development.

1. Change working directory of postinstall process to DESTDIR for source build or extracted dir for binary distribution.  So, the contents of working directory are identical for both cases source build and pre-built binary distribution.  This modification is for the ease of development.
2. Change comments in lib/packages.rb file.

Tested on armv7l.

### DIFF
--- a/crew
+++ b/crew
@@ -424,8 +424,8 @@ def build_and_preconfigure (target_dir)
   end
 end
 
-def post_install (target_dir)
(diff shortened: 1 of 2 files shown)


uberhacker:
I like this change and I now fully understand why you asked me to change the location of weatherrc.

PR #1163, opened 2017-08-28
Update libuv from 1.9.1 to 1.14.0

cstrouse:

This also adds two missing dependencies: automake and libtool.


uberhacker:
@jam7: Before we go crazy and merge all these updates with the binary urls removed, I would like to point out that we're just going to need to change them again and it will require needless version bumping and multiple PRs.

PR #1167, opened 2017-08-28
Add tools to create pre-built binary and upload

jam7:
This PR adds tools which I am using to build pre-built binary and upload them.
This PR also moves several tools from top directory to under tools directory.


tools/Makefile
```diff
@@ -0,0 +1,51 @@
+CREW_PREFIX = /usr/local
+CREW_LIB_PATH = $(CREW_PREFIX)/lib/crew
```
uberhacker: Should this take into account the `/usr/local/lib64` directory on x86_64 architecture?

tools/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+case $1 in
+-k) keep="keep"; shift;;
+*)  keep="";;
+esac
+set -o pipefail
+# yes | crew install buildessential $keep
+arch=`uname -m`
+for i in "$@"; do
+    version=`grep '\<version ['"'"'"]' /usr/local/lib/crew/packages/$i.rb | head -1 | sed -e 's/^.*version *['"'"'"]//' -e 's/['"'"'"].*$//'`
```
uberhacker: Should we use `CREW_LIB_PREFIX` constant?

tools/genpkgname.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+version=`grep '\<version ['"'"'"]' /usr/local/lib/crew/packages/$1.rb | head -1 | sed -e 's/^.*version *['"'"'"]//' -e 's/['"'"'"].*$//'`
```
uberhacker: `CREW_LIB_PREFIX` constant?

tools/Makefile
jam7: No.  We use /usr/local/lib/crew directory to place crew related stuff even on x86_64 currently.

tools/build.sh
jam7: This is shell script.  How can I use `CREW_LIB_PREFIX`?

tools/genpkgname.sh
jam7: This is shell script.  How can I use `CREW_LIB_PREFIX`?

tools/Makefile
uberhacker: Fair enough.

tools/build.sh
uberhacker: You can define a variable in shell with the same logic as ruby.  Different syntax but the logic to determine the directory is the same.

uberhacker: But in this case, since all the crew stuff goes into `/usr/local/lib/crew` anyway, it's pointless.

PR #1168, opened 2017-08-28
Add new pre-built binaries.

jam7:
Add pre-built binaries for packages newly added recently.


jam7:
Update to add few more binaries.

PR #1181, opened 2017-08-30
Update git and xz in the install script

andzaytsev:
I modified the install script to have the latest versions of `git` and `xz` that are available in `packages/git.rb` and `packages/xzutil.rb`. This will eliminate the need to do `crew upgrade` every time after a fresh install.

I'm wondering if there's a way to avoid duplication of precompiled links and shas in `install.sh` and `packages/`. This is a temporary fix, and the next time any packages listed in the install script are updated, install.sh will have to be modified as well. Should I create a separate issue for that?


jam7:
Thank you for the patch.  There are two problems.  One is need to solve.  The other is possible problem.

 1. Installing xzutils is for the environment which doesn't support xz, so tar.xz is not possible to use in such environment.  Please use `https://bintray.com/chromebrew/chromebrew-cross/download_file?file_path=xz-5.2.3-chromeos-armv7l.tar.gz` instead if you want to use bintray stuff.
 1. Those compiled binaries on user environment may require unnecessary libraries.  For example, compiled ruby requires `libgmp`.  Did you check `git` and `libexec/git-core/*` to see they don't require unnecessary libraries?  If not, we need to check them in our review.  This is just a possible problem which our collaborators might need to check.

andzaytsev:
Hi @jam7, thanks for your comments!

1. I see the problem. :) Reverted to using the `.tar.gz` version.
2.  What do you think is the best way to check? I did a clean install with my changes, and it seems like `git` works fine. Since nothing else was installed, does it mean it's all good?

jam7:
@andzaytsev That's fine.  We collaborators will check it.

If you really want to check it through the correct way, it requires to check all binaries and shared libraries using ldd on all architectures.  For example, `ldd /usr/local/bin/git`, `ldd /usr/loca/libexec/git-core/git-pull`, `ldd /usr/local/libexec/git-core/git-upload-pack`, etc...  It's difficult to do so, though.  This difficulty is the reason why I didn't update binary programs installed by `install.sh`.  ;)

jam7:
I checked all shared library dependencies and it looks good, so I'm merging this PR.
For a record, I checked it like below on each architecture.
```
$ mkdir tmp
$ tar xfp .../git-2.14.1-1-chromeos-ARCH.tar.xz
$ find . -type f -perm +111 | xargs ldd|grep -v '\(librt\|libpthread\|libc\|ld-linux.*\|linux-vdso\|linux-gate\|libz\|libexpat\|libcurl\|libssl\|libcrypto\|libssh2\|libdl\|libgcc_s\)\.so'
```

PR #1121, opened 2017-08-21
Add cmus package

uberhacker:
cmus is a small, fast and powerful console music player for Unix-like operating systems.  See https://cmus.github.io/.

PR #1182, opened 2017-08-30
[WIP] Support custom user CREW_PREFIX

dragon788:
The `${VAR:-/some/path}` will use the path on the right side if the variable on the left side is null or empty (ie if the user didn't set it).

I tested this on my machine as I wanted to set the prefix to `/usr/local/.chromebrew` to prevent it from getting wiped by `dev_install --uninstall` per #536 and #429 though it may require some work on #877 to remove symlinks only if they were created by Chromebrew (if it would attempt to symlink a binary into `/usr/local/bin` only if it didn't already exist).


dragon788:
Some caveats, the install script gets crew in the requested spot, but there are many places paths to config files or other bits appears to be hardcoded or incorrectly assembled. I'll be working through these as I get time, as I'd really like to have Chromebrew and Linuxbrew working nicely side by side in addition to the native ChromeOS `dev_install` packages.

jam7:
It might be possible to solve the path problem by opposite way.  First, we change to use `/usr/local/.chromebrew` as the default place to install and compile everything.  Then, we can install everything into `/usr/local` if users want to stick `/usr/local`.  In this way, we can make a symbolic link from `/usr/local/.chromebrew` to `/usr/local` to avoid those path problem.

This requires drastic changes, though...

jam7:
LGTM.  Is it OK to merge this one?  I appreciate hear something from other collaborators and owner.  Thanks.

uberhacker:
@dragon788: Thank you for submitting this PR.  I agree with this concept in theory but in order for it to actually work, we need to update every package to configure via `--prefix=#{CREW_PREFIX}`.  Right now, that is not the case.  I would like to eventually modify the crew command to do this automagically along with `--libdir=#{CREW_LIB_PREFIX}` if these configure options are not provided in the `def self.build` section of the package.  I just haven't figured out how to do it elegantly yet.  Obviously, we would like the ability to override the auto-configured options and we also need to consider cases where "non-standard" configure scripts are used.  A prime example is PR #1121 which is a total PITA to compile.

dragon788:
Does crew support reporting its current prefix with `crew prefix` similar to how brew's is available via `brew prefix`? That could make testing some of these things easier.

spoelstraethan:
@uberhacker I just ran into it today where CREW_PREFIX wasn't being honored by the Ruby and other Bootstrap packages in spite of this PR being merged, have you made any progress on updating crew to handle fixing this for all packages to use the path specified by the user in .bashrc or exported in the command line? It may also be good to capture the value to an env file (which I think the installer does).

PR #1183, opened 2017-09-01
Updating and fixing libszl

yancouto:
Fixes part of #1051 (that was closed before everything was fixed)

I picked the latest commit in the [repository](http://git.1wt.eu/web/libslz.git/). Should I have picked the last commit tagged as a release? It may be stabler...


jam7:
Released version is better.  But if you have any reason and you describe it, developing current version is also good.  But, it is required to modify version number to short hash number since it is not 1.1.0 anymore.

yancouto:
No objections. Switched to release version.

PR #1188, opened 2017-09-02
Add msttcorefonts package

uberhacker:
An easy way to install Microsoft's TrueType core fonts on Linux.  See http://corefonts.sourceforge.net/.

### DIFF
--- a/packages/msttcorefonts.rb
+++ b/packages/msttcorefonts.rb
@@ -15,6 +15,8 @@ class Msttcorefonts < Package
   depends_on 'cabextract'
 
   def self.build
+    # Since there is no configure script available, some modifications to prevent a 'file not found'
+    # message and writes to read-only directories were necessary in the update-ms-fonts script.
     system "sed -i '61icp cabfiles.sha256sums \$SCRATCHDIR/' update-ms-fonts"
     system "sed -i 's,/var/lib/msttcorefonts/,,g' update-ms-fonts"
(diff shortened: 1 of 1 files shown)


jam7:
@uberhacker Can you add some comments to those `sed` command lines what you want to achieve in each line.  It helps reviewer and someone who may need to modify/update this package file.  Thanks.

uberhacker:
Done.

uberhacker:
Is there anything else holding this PR back from getting merged?

cstrouse:
This package installs several duplicate .ttf files that are 0 bytes.

For example
```bash
chronos@localhost /usr/local/share/fonts/truetype/msttcorefonts $ du -sh Impact.ttf 
136K    Impact.ttf
chronos@localhost /usr/local/share/fonts/truetype/msttcorefonts $ du -sh impact.ttf 
0       impact.ttf
```

uberhacker:
Remember, in *nix systems, files are case sensitive so the files are not technically duplicates.  However, I agree there shouldn't be any empty files.

cstrouse:
Should we do something like run find on the directory and remove any 0 length files in the postinstall step or is there a better way to handle this?

uberhacker:
It might be better to determine how these files are getting generated and then stop it before it happens.

cstrouse:
Sorry, I was mistaken about the package installing empty files; I didn't see that they were symlinks. Ready to merge.

PR #1189, opened 2017-09-02
Add pmd package

uberhacker:
PMD is a source code analyzer. It finds common programming flaws like unused variables, empty catch blocks, unnecessary object creation, and so forth. It supports Java, JavaScript, Salesforce.com Apex and Visualforce, PLSQL, Apache Velocity, XML, XSL.

Additionally it includes CPD, the copy-paste-detector. CPD finds duplicated code in Java, C, C++, C#, Groovy, PHP, Ruby, Fortran, JavaScript, PLSQL, Apache Velocity, Scala, Objective C, Matlab, Python, Go, Swift and Salesforce.com Apex and Visualforce.

See https://pmd.github.io/.

### DIFF
--- a/packages/pmd.rb
+++ b/packages/pmd.rb
@@ -12,6 +12,7 @@ class Pmd < Package
   binary_sha256 ({
   })
 
+  depends_on 'jdk8'
   depends_on 'unzip'
 
   def self.install
--- a/packages/pmd.rb
+++ b/packages/pmd.rb
@@ -16,10 +16,15 @@ class Pmd < Package
 
   def self.install
     system "mkdir -p #{CREW_DEST_PREFIX}/bin"
+    FileUtils.cd("#{CREW_DEST_PREFIX}/bin") do
+      system "echo '#!/bin/bash' > pmd"
+      system "echo 'PWD=$(pwd)' >> pmd"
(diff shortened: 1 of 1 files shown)


jam7:
@uberhacker When I execute pmd like their example, I have following errors.
```
$ pmd -d /usr/src -R java-basic -f text
-d is NOT a valid application name, valid options are:pmd, cpd, cpdgui, designer, bgastviewer
Error: Could not find or load main class 
```
I think you need to modify pmd script something like below.  You may need to prepare similar script for `cpd`, `cpdgui`, `designer` and `bgastviewer` also.
```
bin/run.sh pmd "$@"
```

uberhacker:
The bash script in `/usr/local/bin/pmd` should handle that already.  I think the intended way is to provide the pmd command that you want to execute as the first parameter like below:
```
$ pmd [cpd|...] -d /usr/src -R java-basic -f text
```

uberhacker:
After analyzing it a bit more, I think you are correct.  I added a cpd script and modified the pmd script to work as separate commands.  I didn't include the graphics commands since we don't have X11 working yet.  I'll just need to remember to add those once @flyingP0tat0 gets the X Window System :tm: fully operational.

Issue #992, opened 2017-07-31
nano refresh loop


flywebguy:
Immediately after installing nano in Chromebrew I tried to open a test project's index.html file and got this: https://photos.app.goo.gl/PsrlFhPqVwsCKomF3

This is on a Samsung Chromebook Plus on beta version 60.0.3112.80

jam7:
@flyingP0tat0 do you have this problem?  I'm having the same issue on armv7l Chrome OS 59.

As a remedy, it is possible to avoid this problem by running `nano` inside `screen`.  I tested `nano` inside `screen`, therefore I didn't notice this problem.

flywebguy:
@jam7 you're awesome. Thanks for that quick fix. 

It'd be nice if screen wasn't required, but great find nonetheless!

flyingP0tat0:
@jam7 @flywebguy Yes I am having the same issue.
Of course, the `screen` fix won't stay forever, we need to find a solution.

jam7:
This problem still happens on the latest 2.8.7.

lyxell:
Do we know more about the source of this bug?

Kriskras99:
It was solved for me by installing the `readline` package, the binary was probably build with the `readline` package installed so it was automatically linked. I will open a pull request to fix the dependency.

qualityshepherd:
This is a problem on my chromebook asus c201 (rockchip). Installing `readline` didn't fix the issue.

cstrouse:
I have `readline` installed and can't reproduce this on the Samsung Chromebook Plus running stable channel.

hursey013:
This is still an issue, happening on Asus C302, ChromeOS Version 63.0.3239.140 (Official Build) (64-bit).

PR #1191, opened 2017-09-02
Fix nano on arm

Kriskras99:
See #992.
Shouldn't cause any problems on the other archs

### DIFF
--- a/packages/nano.rb
+++ b/packages/nano.rb
@@ -23,6 +23,7 @@ class Nano < Package
   depends_on 'buildessential'
   depends_on 'ncurses'
   depends_on 'filecmd'
+  depends_on 'readline'
 
   def self.build
     system "./configure CPPFLAGS=\"-I/usr/local/include/ncurses\""


jam7:
Thank you for the PR.  I tried this, but it won't work as expected on both armv7l and x86_64.  It looks like nano doesn't use readline after this PR also.
```
$ ldd /usr/local/bin/nano
        libz.so.1 => /usr/local/lib/libz.so.1 (0xaed21000)
        libmagic.so.1 => /usr/local/lib/libmagic.so.1 (0xaecf2000)
        libncursesw.so.6 => /usr/local/lib/libncursesw.so.6 (0xaec87000)
        libc.so.6 => /usr/local/lib/libc.so.6 (0xaeba1000)
        /lib/ld-linux-armhf.so.3 (0xaed4b000)
```
Can you explain more details?

I did:
```
$ crew remove nano
$ crew install readline
$ crew install nano -s     # compile only nano from the source code
```

Kriskras99:
I just used the provided binary and it worked. But after rebooting it's failing again.

PR #1192, opened 2017-09-02
Add fop package

uberhacker:
Apache™ FOP (Formatting Objects Processor) is a print formatter driven by XSL formatting objects (XSL-FO) and an output independent formatter. It is a Java application that reads a formatting object (FO) tree and renders the resulting pages to a specified output. Output formats currently supported include PDF, PS, PCL, AFP, XML (area tree representation), Print, AWT and PNG, and to a lesser extent, RTF and TXT. The primary output target is PDF.  See https://xmlgraphics.apache.org/fop/.

### DIFF
--- a/packages/fop.rb
+++ b/packages/fop.rb
@@ -22,6 +22,9 @@ class Fop < Package
       system "echo 'cd $PWD' >> fop"
       system "chmod +x fop"
     end
+  end
+
+  def self.postinstall
     puts
     puts "To complete the installation, set the environment variable as follows:".lightblue
     puts "echo 'export JAVA_HOME=#{CREW_LIB_PREFIX}/jdk8' >> ~/.bashrc && source ~/.bashrc".lightblue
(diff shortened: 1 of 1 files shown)


packages/fop.rb
```diff
@@ -0,0 +1,30 @@
...
+      system "echo 'PWD=$(pwd)' >> fop"
+      system "echo 'cd #{CREW_LIB_PREFIX}/fop/fop' >> fop"
+      system "echo './fop \"$@\"' >> fop"
+      system "echo 'cd $PWD' >> fop"
+      system "chmod +x fop"
+    end
+    puts
+    puts "To complete the installation, set the environment variable as follows:".lightblue
+    puts "echo 'export JAVA_HOME=#{CREW_LIB_PREFIX}/jdk8' >> ~/.bashrc && source ~/.bashrc".lightblue
+    puts
```
jam7: @uberhacker Don't you mind to use postinstall function to put these messages at install time for either source build or pre-built binary distribution?

uberhacker:
Good call.  Thought I added it already but apparently not.

PR #1199, opened 2017-09-03
Adding texlive package

yancouto:
Texlive has all the useful tex and latex stuff to create documents.

The installation is not an usual one so I had to move a bunch of things by hand.

There are still some problems:
- I'm not sure if it works on other architectures. I think it does, the install script probably figures it out. But it would be great if someone could test it.
- The script does not install the binaries to a reasonable location. It installs to INSTALL_TEXLIVE_PREFIX/texlive/2017/bin/ARCHITECTURE/. I made the install prefix be /usr/local/share and my solution was to symlink all the binaries there to /usr/local/bin. **This is not great**, since if you install more binaries using `tlmgr` (their package manager), they will go to their binaries directory, and not be symlinked to /usr/local/bin. Their recommendation is to change the PATH but as I've seen that packages here usually do not do that. **Is there a better solution to this?**

### DIFF
--- a/packages/texlive.rb
+++ b/packages/texlive.rb
@@ -11,8 +11,8 @@ class Texlive < Package
 
   def self.build
     system "rm -rf *"
-    system "wget mirror.ctan.org/systems/texlive/tlnet/install-tl-unx.tar.gz"
-    system "wget mirror.ctan.org/systems/texlive/tlnet/install-tl-unx.tar.gz.sha512"
+    system "wget ftp://ftp.fu-berlin.de/tex/CTAN/systems/texlive/tlnet/install-tl-unx.tar.gz"
+    system "wget ftp://ftp.fu-berlin.de/tex/CTAN/systems/texlive/tlnet/install-tl-unx.tar.gz.sha512"
     system "cat install-tl-unx.tar.gz.sha512 | xargs | cut -d' ' -f1 > sha512"
     sha512 = open('sha512').read.chomp
     abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA512.hexdigest( File.read('install-tl-unx.tar.gz') ) == "#{sha512}"
--- a/packages/texlive.rb
+++ b/packages/texlive.rb
@@ -5,7 +5,7 @@ class Texlive < Package
   homepage 'https://www.tug.org/texlive/'
   version '2017'
   source_url 'http://mirror.ctan.org/systems/texlive/tlnet/install-tl-unx.tar.gz'
-  source_sha256 '9bc5c663bf5fb7066b4bf3c54e5ce93251923d4edaeadb485224621a02297a39'
+  source_sha256 '13e36ba42cf6fb3b4eb96878bdb2828b38709c9538e9446c723ca82fd00a6ac4'
 
   def self.build
   end
@@ -18,15 +18,15 @@ class Texlive < Package
            TEXLIVE_INSTALL_TEXMFHOME=#{dir}/local \
            ./install-tl --scheme=basic --no-cls"
     system "find #{dir} -iname '*.pdf' -delete" # saving some space
-    system "mkdir -p #{CREW_DEST_PREFIX}/man"
-    system "mv #{dir}/2017/texmf-dist/doc/man/* #{CREW_DEST_PREFIX}/man/" # copying manpages
-    system "rm -rf #{dir}/2017/texmf-dist/doc"
-    system "mkdir -p #{CREW_DEST_PREFIX}/bin"
-    system "find #{dir}/2017/bin/ -xtype f -exec ln -rs {} #{CREW_DEST_PREFIX}/bin ';'" # linking binaries
-    system "#{CREW_DEST_PREFIX}/bin/tlmgr init-usertree"
+    system "find #{dir}/2017/texmf-dist/doc -type f -and -not -path '*man*' -delete"
+    system "find #{dir} -name 'tlmgr' -exec {} init-usertree ';'"
   end
 
   def self.postinstall
-    puts "This is a very small installation, with only the basic packages. To install more, use `tlmgr install <package>`.".lightblue
+    puts "\nPlease add texlive to your PATH and MANPATH to be able to use the executables and manpages. Use the following commands:".lightblue
+    puts " echo \"export PATH=\$PATH:#{CREW_PREFIX}/share/texlive/2017/bin/#{ARCH}-linux\" >> ~/.bashrc".lightblue
+    puts " echo \"export MANPATH=\$MANPATH:#{CREW_PREFIX}/share/texlive/2017/bin/texmf-dist/doc/man\" >> ~/.bashrc".lightblue
+    puts " source ~/.bashrc".lightblue
+    puts "\nThis is a very small installation, with only the basic packages. To install more, use `tlmgr install <package>`.".lightblue
   end
 end
--- a/packages/texlive.rb
+++ b/packages/texlive.rb
@@ -1,0 +1,32 @@
+require 'package'
+
+class Texlive < Package
+  description 'TeX Live is an easy way to get up and running with the TeX document production system.'
+  homepage 'https://www.tug.org/texlive/'
+  version '2017'
+  source_url 'http://mirror.ctan.org/systems/texlive/tlnet/install-tl-unx.tar.gz'
+  source_sha256 '9bc5c663bf5fb7066b4bf3c54e5ce93251923d4edaeadb485224621a02297a39'
+
+  def self.build
+  end
+
+  def self.install
+    dir = "#{CREW_DEST_PREFIX}/share/texlive"
+    system "yes I | TEXLIVE_INSTALL_PREFIX=#{dir} \
+           TEXLIVE_INSTALL_TEXMFVAR=#{dir}/local/texmf-var \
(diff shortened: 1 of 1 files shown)


jam7:
Regarding to your qestions.

- Which architecture did you try this?  I can try another if you let me know which one.
- Modifying PATH is a simplest way to solve your situation.  Otherwise, you need to prepare crew package files to install each texlive packanges like what I did in perl_*.rb.  For me, it is required since packages uses those perl_*.rb is installed by crew.  I needed to make texinfo package work in binary distribution.  For your case, it is not necessary at the moment.

I will merge this PR if following points are modified.

- Please change to use `ln -s` insteal of `mv` for manuals like what you did for binaries.
- Please add some guide message in postinstall to let users modify their PATH and MANPATH.  Modifying MANPATH is needed since `tlmgr` will install not only binaries but also manuals.

yancouto:
I added a message explaining how to add the directories to PATH and MANPATH. I removed the linking to /usr/local/bin since it doesn't make any sense if the directories are already added to PATH.

I've only tested in x86_64. Would be grateful if you could test in another, since I'm not sure if the installer works. There is a directory whose name depends on the architecture, but I used `find` to avoid dealing with this. The explanation message in the end assumes the name of the directory is `#{ARCH}-linux`, which it is at least for x86_64.

jam7:
Thank you for the update.  Unfortunately, their sha256 is changed everytime I download their file...

yancouto:
This happened here too. Maybe the file changes depending on the mirror used?

yancouto:
I fixed a mirror. Should work now, but will probably be slower.

jam7:
Thank you for the update, but their sha256sum is now `1e09ffe84046af5fa68c788a3fd3c517f9c7e131d9d08603e2856cba22adf36e`...

jam7:
Today's their sha256 value is `ee4841a90b7e8248c13a0eeace4bfb737d1fc2421cfefdf45f9b96753264bf17`  :/

jam7:
They are updating their install script every day.  They also update their sha512 for the archive everyday.  It just doesn't match what chromebrew does.  We need different mechanism to create a package file for texlive.  I have no idea at the moment, though.

jam7:
On the other hand, I could install this one into armv7l well.  The path for armv7l is `/usr/local/share/texlive/2017/bin/armhf-linux`, though.  MANPATH also has small typo.  It is better to modify this package file like below.
```
    binpath=`echo #{CREW_PREFIX}/share/texlive/2017/bin/*-linux`.strip
    puts " echo \"export PATH=\$PATH:#{binpath}\" >> ~/.bashrc".lightblue
    puts " echo \"export MANPATH=\$MANPATH:#{CREW_PREFIX}/share/texlive/2017/texmf-dist/doc/man\" >> ~/.bashrc".lightblue
```

idifahs:
I tried to add the .rb file to my current install as I need tex, and the SHA256 is still changing.

uberhacker:
Hey @idifahs: I'm not sure I understand what you are trying to do here.  Could you please elaborate a bit?  Thanks

idifahs:
The issue that @jam7 pointed out about the SHA signature not remaining consistent still exists, despite @yancouto stating it should be fixed. I was just trying to let them know that the issue does still exist.

uberhacker:
Ah, I see what's happening now.  You could do something like this:
```
source_url 'ftp://ftp.fu-berlin.de/tex/CTAN/systems/texlive/tlnet//update-tlmgr-r45305.sh'
source_sha256 '...'

def self.build
  system "wget ftp://ftp.fu-berlin.de/tex/CTAN/systems/texlive/tlnet/install-tl-unx.tar.gz"
  system "wget ftp://ftp.fu-berlin.de/tex/CTAN/systems/texlive/tlnet/install-tl-unx.tar.gz.sha512"
  TL_SHA512 = `cat install-tl-unx.tar.gz.sha512`
  abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA512.hexdigest( File.read('install-tl-unx.tar.gz') ) == "#{TL_SHA512}"
  system "tar xvf install-tl-unx.tar.gz"
end

...
```
The source_url is just a throw away that gets discarded but since this file is versioned, it won't be a moving target.

uberhacker:
@yancouto:  Actually, I just tried installing this and I believe what you need to do is shown below.  I also noticed perl should be added as a dependency.  Hackish, I know, but it works. :)
```
require 'package'                                                                                                                                                                                                 
                                                                                                                                                                                                                  
class Texlive < Package                                                                                                                                                                                           
  description 'TeX Live is an easy way to get up and running with the TeX document production system.'                                                                                                            
  homepage 'https://www.tug.org/texlive/'                                                                                                                                                                         
  version '2017'                                                                                                                                                                                                  
  source_url 'ftp://tug.org/historic/systems/texlive/2017/texlive-20170524-extra.tar.xz'                                                                                                                          
  source_sha256 'afe49758c26fb51c2fae2e958d3f0c447b5cc22342ba4a4278119d39f5176d7f'                                                                                                                                
                                                                                                                                                                                                                  
  depends_on 'perl'                                                                                                                                                                                               
                                                                                                                                                                                                                  
  def self.build                                                                                                                                                                                                  
    system "rm -rf *"                                                                                                                                                                                             
    system "wget ftp://ftp.fu-berlin.de/tex/CTAN/systems/texlive/tlnet/install-tl-unx.tar.gz"                                                                                                                     
    system "wget ftp://ftp.fu-berlin.de/tex/CTAN/systems/texlive/tlnet/install-tl-unx.tar.gz.sha512"                                                                                                              
    system "cat install-tl-unx.tar.gz.sha512 | xargs | cut -d' ' -f1 > sha512"                                                                                                                                    
    sha512 = open('sha512').read.chomp                                                                                                                                                                            
    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA512.hexdigest( File.read('install-tl-unx.tar.gz') ) == "#{sha512}"                                                                        
    system "tar xvf install-tl-unx.tar.gz"                                                                                                                                                                        
    system "mv install-tl-2017*/* ."                                                                                                                                                                              
    system "rm -rf install-tl-2017*/"                                                                                                                                                                             
  end                                                                                                                                                                                                             
```
Let me know if you need any more help.

flyingP0tat0:
There has to be another option than downloading everything twice. We should look for constant sources. Also using a mirror is not really good for people in other parts of the world...

uberhacker:
@flyingP0tat0: Agreed but in this case I don't think they include the version in the filename of the latest release.  We're actually not downloading twice.  In my example, I'm just downloading a tiny file available from an earlier release.  The footprint is actually quite small.  We could use older releases but it would always be one version behind the latest.  I've seen some repos not even offer earlier releases.  We have a few packages in that situation.  In those cases, if a version is bumped, there is just a 404 on the file and that becomes the clue that we need to upgrade.

uberhacker:
@yancouto: Did you get a chance to look at this yet?

yancouto:
Heya, I was travelling so couldn't check this out. I like the solution, it is not ideal, but with how things are structured I think it is the best that can be done. Thanks @uberhacker .

I changed to use `mirror.ctan.org` instead of a specific mirror. I just realized that it may not be a good thing, since I tried once and it downloaded the file and the sha512 from different mirrors and got Checksum mismatch.

Since the installer is small I don't think there is a problem with fixing a mirror. I'm pretty sure when the installer downloads the files it will look for a new mirror again. So I will change it tomorrow if there are no objections. ( @flyingP0tat0 ?)

Actually, I think the best solution may be to keep a copy of the installer in the bintray. No more problems with checksums then. I don't have access so I can't do it though.

yancouto:
Fixed the mirror. Anything left to do to get this accepted?

uberhacker:
@yancouto: Thank you for the update.  Changing the mirror doesn't fix the issue with the install tarball though, right?  Since they don't tag releases, this is always a moving target.

yancouto:
Yes, the file is always changing, but that's why we're downloading it manually. It works as it is, but the moving file might bring problems in the future (when the version changes to 2018 maybe).

Another solution is to store the installer in bintray, do you think it is better?

uberhacker:
The solution in https://github.com/skycocker/chromebrew/pull/1199#issuecomment-333324699 is probably better since it overcomes the moving target issue.  You are welcome to replace with the new mirror urls.

yancouto:
Did that (and similar stuff for other parts that used 2017).

uberhacker:
@yancouto: Thank you for the update.  It installs fine on x86_64.  @cstrouse: Can you please confirm if this works on aarch64?

cstrouse:
The install completes but the **bin** directory is missing in `/usr/share/local/texlive`.

yancouto:
The `bin` should actually be in `/usr/share/local/texlive/2017/bin` and the executables should be in `/usr/local/share/local/texlive/2017/bin/aarch64-linux` (or something like that for your architecture). Are they not there?

I'm not sure how to fix it in this case. Sorry it took so long to reply.

uberhacker:
@yancouto: Does `tlmgr` work with this package?

yancouto:
Yes. This installs only a small number of latex packages, and the rest you should install as needed with tlmgr.

uberhacker:
@yancouto: OK, thanks.  Do you think this is ready to be merged @cstrouse?

PR #1153, opened 2017-08-27
Update crew to explain multiple argument handling

uberhacker:
Also updated comments for clarity.

PR #1200, opened 2017-09-03
Refactoring crew

yancouto:
Changelog:
- Moved string colorization to lib/color.rb
- Moved constants to lib/const.rb
- Moved sudo checking before docopt checking.
- Changed docopt require to be relative
- Fixed stray tabs

First step to #963

### DIFF
--- a/crew
+++ b/crew
@@ -1,11 +1,15 @@
 #!/usr/bin/env ruby
+require_relative 'lib/color'
+
+# Disallow sudo
+abort "Chromebrew should not be run as root.".lightred if Process.uid == 0
+
 require 'find'
 require 'net/http'
 require 'uri'
 require 'digest/sha2'
 require 'json'
 require 'fileutils'
-require_relative 'lib/color'
 require_relative 'lib/const'
 
 # Add lib to LOAD_PATH
@@ -32,9 +36,6 @@ Usage:
 
 version 0.4.3
 DOCOPT
-
-# Disallow sudo
-abort "Chromebrew should not be run as root.".lightred if Process.uid == 0
 
 # Parse arguments using docopt
 require_relative 'lib/docopt'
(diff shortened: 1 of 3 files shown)


crew
```diff
@@ -30,49 +33,11 @@ Usage:
...
-  ENV["XZ_OPT"] = ENV["CREW_XZ_OPT"]
-end
-
-USER = `whoami`.chomp
-
-# Add lib to LOAD_PATH
-
-$LOAD_PATH.unshift "#{CREW_LIB_PATH}lib"
+# Disallow sudo
+abort "Chromebrew should not be run as root.".lightred if Process.uid == 0
```
uberhacker: I'm actually thinking this should go above everything.  We want this to fail as early as possible.

uberhacker:
@yancouto:  Hopefully, the crew changes in #963 get merged before #1153.  That way you don't need to resolve merge conflicts. :)

crew
yancouto: Well, it has to be after the string colorization anyway, since it is displayed in red. So, as of now, before the sudo check there are just some require's and the constants definition, is that a problem?

uberhacker: I was the one who added the colorization section so I definitely understand.  I was just thinking since you have the `require_relative 'lib/color'` line now, it could go in immediately after that section.

yancouto: Done

jam7:
Nice PR.  I'm going to merge in a day or two if there is no objection.  Thanks.

crew
```diff
@@ -30,49 +37,8 @@ Usage:
...
-USER = `whoami`.chomp
-
-# Add lib to LOAD_PATH
-
-$LOAD_PATH.unshift "#{CREW_LIB_PATH}lib"
-
 # Parse arguments using docopt
-
-require 'docopt'
+require_relative 'lib/docopt'
```
uberhacker: Might be good if this was up with the rest of the `require...` lines.  Not a biggie but just helps keeps things more organized.

PR #1201, opened 2017-09-03
Add banner package

cstrouse:
Banner is an implementation of the traditional Unix-program used to display
large characters.

### DIFF
--- a/packages/banner.rb
+++ b/packages/banner.rb
@@ -1,0 +1,25 @@
+require 'package'
+
+class Banner < Package
+  description 'An implementation of the traditional Unix-program used to display large characters.'
+  homepage 'http://shh.thathost.com/pub-unix/#banner'
+  version '1.3.2'
+  source_url 'http://shh.thathost.com/pub-unix/files/banner-1.3.2.tar.gz'
+  source_sha256 '0dc0ac0667b2e884a7f5ad3e467af68cd0fd5917f8c9aa19188e6452aa1fc6d5'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'shhmsg'
+  depends_on 'shhopt'
+
+  def self.build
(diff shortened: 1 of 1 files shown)


packages/banner.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  depends_on 'shhmsg'
+  depends_on 'shhopt'
+
+  def self.build
+    system 'make'
+  end
+
+  def self.install
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
```
uberhacker: You might add `PREFIX=#{CREW_PREFIX}` here since I'm assuming there is no configure script available?

cstrouse: No configure script but the Makefile defaults to /usr/local. Good point though that it should be explicit.

uberhacker:
According to `man make`, the variable is `prefix` (lowercase).  Not sure if that makes a difference or not but if it's case insensitive, then it would matter.

jam7:
Now sure how their makefile work.
```
$ crew build banner
...
$ tar tvf banner-1.3.2-chromeos-x86_64.tar.xz
-rw-r--r-- chronos/chronos   0 2017-09-05 02:59 dlist
-rw-r--r-- chronos/chronos   0 2017-09-05 02:59 filelist
```
It looks like shhmsg and shhopt also doesn't install binaries through CREW_DEST_DIR.

uberhacker:
@cstrouse: Did you get a chance to look into this issue yet?

cstrouse:
@uberhacker I forgot all about this. I'll see if I can get this resolved today.

uberhacker:
It looks like the dependencies `shhmsg` and `shhopt` both don't abstract the library files but that is not banner's fault.  I'll go ahead and merge this and hopefully, we can get the dependency packages updated eventually.

PR #1202, opened 2017-09-03
Add miniupnpc package

cstrouse:
miniupnpc is a lightweight upnp client library. Added as a prereq for bitcoin.

### DIFF
--- a/packages/miniupnpc.rb
+++ b/packages/miniupnpc.rb
@@ -1,0 +1,23 @@
+require 'package'
+
+class Miniupnpc < Package
+  description 'UPnP IGD client lightweight library'
+  homepage 'http://miniupnp.free.fr/'
+  version '2.0'
+  source_url 'http://miniupnp.free.fr/files/download.php?file=miniupnpc-2.0.tar.gz'
+  source_sha256 'd434ceb8986efbe199c5ca53f90ed53eab290b1e6d0530b717eb6fa49d61f93b'
+
+  binary_url ({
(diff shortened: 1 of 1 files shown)


packages/miniupnpc.rb
```diff
@@ -0,0 +1,23 @@
...
+  source_sha256 'd434ceb8986efbe199c5ca53f90ed53eab290b1e6d0530b717eb6fa49d61f93b'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.build
+    system "sed -i '139s,/usr,,' Makefile"
+    system "make", "PREFIX=#{CREW_PREFIX}", "DESTDIR=#{CREW_DEST_DIR}"
```
uberhacker: We would only need the `PREFIX` and `DESTDIR` options in the `def self.install` section, correct?

cstrouse: It was putting things in the wrong location without it added there for some reason.

uberhacker: I just read this in the make man page:
```
Variable names are case-sensitive. The names ‘foo’, ‘FOO’, and ‘Foo’ all refer to different variables.
```
So I guess we should be using `prefix=#{CREW_PREFIX}` instead of `PREFIX=#{CREW_PREFIX}`.  See https://www.gnu.org/software/make/manual/html_node/Using-Variables.html.  I wonder if that is the issue here?

PR #1204, opened 2017-09-04
Added a package for Nethack 3.6.

indianajohn:
No package manager is complete without NetHack.

Stuff I thought about while making this package:
1. There does not currently seem to be a way to change `CREW_PREFIX`. If there ever is one, this package will not work without modification, because the wrapper that is created to change to `HACKDIR` contains the absolute path to `HACKDIR` when `make install` is run.

2. It would be nice to add some binary packages. I figure they could be added to the Bintray hosting. There doesn't seem to be any documentation on how to do this, so I left that part out for now.

3. This version of Nethack is console-only. A version that includes tiles and linkage to X11 should probably be a separate package.

### DIFF
--- a/packages/nethack.rb
+++ b/packages/nethack.rb
@@ -12,17 +12,21 @@ class Nethack < Package
   depends_on 'flex'
   depends_on 'ncurses'
 
-  def self.build
+  def self.patch
     # Apply a patch to set a hints file that will work for the current build environment. 
     system "curl -L 'https://gist.githubusercontent.com/anonymous/77b05c6cd87628ab3cb944e75ecc45b7/raw/d5f327450aa6b4d50cafb5352fad06ed60f91b69/nethack_chromebrew.patch' | patch -p0"
+  end
+
+  def self.build
+    self.patch
     system "./sys/unix/setup.sh", "./sys/unix/hints/linux"
     system "make"
   end
 
   def self.install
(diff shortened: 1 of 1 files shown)


uberhacker:
@indianajohn: Thank you for submitting this PR.  You might want to change `#CREW_DEST_DIR}/usr/local` to `#{CREW_DEST_PREFIX}` instead.  The reason is we might not always use in `/usr/local` per PR #1182.  Also, I'm not sure you are aware but we have `def self.patch` instead of `def self.build` available for applying patches.

indianajohn:
Hey,

Thanks for the comments. My latest commit should address both of them.

I assumed that `self.patch` would be called automatically before `self.build`, but that appears to not be the case. Maybe a recent change? Anyway, I restructured the formula so `self.build` is a separate method, and was forced to make `self.install` call `self.build`. That was the way that the only other reference in the source tree to `self.build` ( https://github.com/skycocker/chromebrew/blob/8d9605251aa2bd751efda813edc74987d119ec4f/packages/unzip.rb ) seemed to do it.

jam7:
Wow, nice.  It works.  Thanks.  I cannot play `nethack4.rb` well on `crosh`, so I was looking for working nethack.  :)

Regarding to your question, we never use `self.patch` AFAIK.  So, I guess it is for only contributor of `unzip.rb`.  But anyway, tt is possible to define it and call it from `self.build` whenever.

PR #1205, opened 2017-09-04
Update postgres from 9.5.0 to 9.6.5

uberhacker:
This also changes `/usr/local` references to `#{CREW_PREFIX}` instead and simplifies usage.

### DIFF
--- a/packages/postgres.rb
+++ b/packages/postgres.rb
@@ -3,21 +3,13 @@ require 'package'
 class Postgres < Package
   description 'PostgreSQL is a powerful, open source object-relational database system.'
   homepage 'https://www.postgresql.org/'
-  version '9.5.0'
-  source_url 'https://ftp.postgresql.org/pub/source/v9.5.0/postgresql-9.5.0.tar.bz2'
-  source_sha256 'f1c0d3a1a8aa8c92738cab0153fbfffcc4d4158b3fee84f7aa6bfea8283978bc'
+  version '9.6.5'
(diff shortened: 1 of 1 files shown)


jam7:
Not sure creating `/usr/local/pgsql/data` database automatically.  Some user may want to use `/usr/local/share/pgsql/data` instead.  I prefer old logic, just showing guide messages and let user choose.

uberhacker:
The old logic creates the directory automatically also.  Actually, I was thinking a good default location could be `/usr/local/data/pgsql`.  What if we add a message to show where the databases will be stored.  Will that work?

uberhacker:
Maybe there is some confusion of what initdb actually does.  It doesn't create a database.  All it does is initialize the database directory.  I think most users won't care which directory stores the databases but I'm sure they do care and want to know where the directory resides.

jam7:
@uberhacker You are right.  Old logic creates the `/usr/local/pgsql/data` directory.  But you are performing not only `mkdir` but also `initdb` without explanations.  You also adding `pgctl` script to controll pgsql without explanations.  Can you explain what you did in your PR first?

uberhacker:
This package installs just as the official documentation recommends.  It is not necessary to put the burden on the user to execute these post-install steps manually.

uberhacker:
@jam7: I added comments to provide an explanation to the madness.  I also removed the `pgctl` script.

cstrouse:

PR #1206, opened 2017-09-06
Change shhmsg.rb and shhopt.rb to install files through CREW_DEST_PREFIX

jam7:
Tested on x86_64.

I think something similar modifications are required for #1201 too.


uberhacker:
Nice!  So would this replace the install following line?  We would only need this if the configure script is not available, correct?
```
system "make", "prefix=#{CREW_PREFIX}", "DESTDIR=#{CREW_DEST_DIR}", "install"
```

PR #1207, opened 2017-09-06
Fix possibly incorrect `chown` on /usr/local

AravK:
Should be on CREW_PREFIX

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -26,7 +26,7 @@ case "$architecture" in
 esac
 
 #This will allow things to work without sudo
-sudo chown -R `id -u`:`id -g` /usr/local
+sudo chown -R `id -u`:`id -g` "${CREW_PREFIX}"
 
 #prepare directories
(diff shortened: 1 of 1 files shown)


uberhacker:
Good call!

install.sh
```diff
@@ -26,7 +26,7 @@ case "$architecture" in
 esac
 
 #This will allow things to work without sudo
-sudo chown -R `id -u`:`id -g` /usr/local
+sudo chown -R `id -u`:`id -g` "${CREW_PREFIX}"
```
uberhacker: Should be `"#{CREW_PREFIX}"`.

jam7: it's a shell script.  not a ruby script.

jam7:
Thank you for your PR.

install.sh
uberhacker: Oops, missed that.

PR #1215, opened 2017-09-16
Update php7 from 7.1.8-1 to 7.1.11

uberhacker:
- Replaced /usr/local with #{CREW_PREFIX}
  - Replaced #{CREW_DEST_DIR}/usr/local/lib with #{CREW_DEST_LIB_PREFIX}
  - Added --libdir and --localstatedir configure options

`--localstatedir=#{CREW_PREFIX}/tmp` was needed to prevent the following errors on CloudReady:
```
tar: ./usr/local/var/log: Cannot utime: Operation not permitted
tar: ./usr/local/var/log: Cannot change mode to rwxr-xr-x: Operation not permitted
```

### DIFF
--- a/packages/php7.rb
+++ b/packages/php7.rb
@@ -3,21 +3,13 @@ require 'package'
 class Php7 < Package
   description 'PHP is a popular general-purpose scripting language that is especially suited to web development.'
   homepage 'http://www.php.net/'
-  version '7.1.8-1'
-  source_url 'https://php.net/distributions/php-7.1.8.tar.xz'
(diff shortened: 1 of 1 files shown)


cstrouse:
php just rev'd to 7.1.10 fyi @uberhacker

uberhacker:
Thanks for the heads up @cstrouse.  This is what happens when PRs stay in the queue for weeks at a time. :) Hint: @lyxell @thedamian @jam7 ;)

cstrouse:
php just rev'd to 7.1.11.

uberhacker:
Of course. :)

cstrouse:
The updated version works on both Samsung Chromebook Plus (aarch64) and Samsung Chromebook 3 (x86_64).

PR #1219, opened 2017-09-16
Add execute permissions to .sh scripts

uberhacker:
- Added comment to explain how to install prerequirement commands in upload_github_release.sh


uberhacker:
Closing due to conflict with #1167.

PR #1221, opened 2017-09-18
Add poppler package

uberhacker:
Poppler is a PDF rendering library based on the xpdf-3.0 code base.  See https://poppler.freedesktop.org/.

### DIFF
--- a/packages/poppler.rb
+++ b/packages/poppler.rb
@@ -13,7 +13,6 @@ class Poppler < Package
   })
 
   depends_on 'cairo'
-  depends_on 'fontconfig'
   depends_on 'harfbuzz'
   depends_on 'libjpeg'
   depends_on 'libpng'


cstrouse:
When building on x86_64 I get the following error.

```
checking for jpeglib.h... no
configure: WARNING:
There is an installation error in jpeg support. You seem to have only one
of either the headers _or_ the libraries installed. You may need to either
provide correct --with-extra-... options, or the development package of
libjpeg6b. You can get a source package of libjpeg from http://www.ijg.org/
Disabling JPEG support.

checking libjpeg.h works correctly... no
checking libjpeg.h problem can be fixed by XMD_H macro... no, disable libjpeg
configure: error: Install libjpeg before trying to build poppler. You can also decide to use the internal unmaintained DCT decoder or none at all. See --help.
poppler failed to install: `./configure             --prefix=/usr/local             --libdir=/usr/local/lib64` exited with 1
```

cstrouse:
Missing several dependencies:
- libjpeg
- harfbuzz

uberhacker:
Good catch @cstrouse.  I also noticed this package needs libpng and cairo to enable full functionality.  I'll try and update this when I get a chance. Thanks

packages/poppler.rb
```diff
@@ -13,7 +13,6 @@ class Poppler < Package
   })
 
   depends_on 'cairo'
-  depends_on 'fontconfig'
   depends_on 'harfbuzz'
   depends_on 'libjpeg'
```
cstrouse: It also needs `libopenjpeg`.

uberhacker:
Replaced by #1368.

PR #1222, opened 2017-09-18
Add pdfgrep package

uberhacker:
A commandline utility to search text in PDF files.  See https://pdfgrep.org/.  Depends on #1221.


cstrouse:

PR #1225, opened 2017-09-19
Add missing libxcb dependency

### LINKED ISSUE #1224: ffmpeg missing dependency
After installing the ffmpeg package it fails to run on XE500C13-K01US (stable channel, X86_64) with the following error:
```
chronos@localhost ~ $ ffmpeg 
ffmpeg: error while loading shared libraries: libxcb.so.1: cannot open shared object file: No such file or directory
```

cstrouse:
This adds libxcb to the dependency list as it was missing. Resolves #1224.


jam7:
Sorry for inconvininence.  This problem is caused by wrongly compiled pre-built binary for x86_64.  I'll update binary files soon.  Please use `crew install ffmpeg -s` in the meantime.

PR #1230, opened 2017-09-19
Add pass password manager

Alanaktion:
This adds the standard unix password manager from https://www.passwordstore.org/.

### DIFF
--- a/packages/pass.rb
+++ b/packages/pass.rb
@@ -15,6 +15,6 @@ class Pass < Package
   depends_on 'gnupg'
 
   def self.install
-    system  "make", "PREFIX=/usr/local", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system  "make", "PREFIX=#{CREW_PREFIX}", "DESTDIR=#{CREW_DEST_DIR}", "install"
   end
 end
--- a/packages/pass.rb
+++ b/packages/pass.rb
@@ -1,0 +1,20 @@
+require 'package'
+
+class Pass < Package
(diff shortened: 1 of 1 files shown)


uberhacker:
Cool!  Thank you for adding this @Alanaktion.  Might want to change `--prefix=/usr/local` to `--prefix=#{CREW_PREFIX}` instead to abstract the install location.

Alanaktion:
Thanks, I've added it. I knew there was a variable for it, but couldn't remember what it was, and saw several hard-coded `/usr/local`s in other packages.

uberhacker:
Yes, we need to change those other packages also.  @jam7 has a pending PR to handle similar cases like this without a configure script available.  See https://github.com/skycocker/chromebrew/pull/1206/files.

uberhacker:
He will probably ask you to add this section also. :)
```
def self.build
   system "make"
end
```

Alanaktion:
This program specifically does not have a `make` though, the Makefile just prints:
> Password store is a shell script, so there is nothing to do. Try `make install` instead.

uberhacker:
Well, in that case we don't need that section.

flyingP0tat0:
Just a little style thing: Could you please move the dependencies above the binaries (or remove the empty binaries for now)? Just to ensure the same syntax :)

Alanaktion:
I've removed the binary lines. I just matched what I saw in other packages without binaries, but it's definitely cleaner to not have the empty hashes :)

PR #1231, opened 2017-09-19
Add GPG pinentry

Alanaktion:
This installs the GPG Pinentry program required to use private keys with the `gnupg` package. By default it will install the `tty` and `curses` versions, but the upstream installer detects dependencies for each supported version automatically so other options could work if their dependencies are installed.


jam7:
Some errors in the middle of compile.
```
$ crew install pinentry
...
checking if Unix domain socket is supported... yes
checking for pkg-config... /usr/local/bin/pkg-config
checking for gtk+-2... no
configure: WARNING: pkg-config could not find the module gtk+-2.0
checking for gcr... no
configure: WARNING: pkg-config could not find the module gcr-base-3
checking for pkg-config... (cached) /usr/local/bin/pkg-config
checking for libsecret... no
configure: WARNING: pkg-config could not find the modules libsecret-1
checking for Qt5Core >= 5.0.0 Qt5Gui >= 5.0.0 Qt5Widgets >= 5.0.0... no
Package Qt5Core was not found in the pkg-config search path.
Perhaps you should add the directory containing `Qt5Core.pc'
No package 'Qt5Core' found
checking for QtCore >= 4.6.0 QtGui >= 4.6.0... no
configure:
***
*** You need libgpg-error to build this program.
**  This library is for example available at
***   ftp://ftp.gnupg.org/gcrypt/libgpg-error
*** (at least version 1.16 is required.)
***
configure:
***
*** You need libassuan to build this program.
*** This library is for example available at
***   ftp://ftp.gnupg.org/gcrypt/libassuan/
*** (at least version 2.1.0 (API 2) is required).
***
configure: error: 
***
*** Required libraries not found. Please consult the above messages
*** and install them before running configure again.
***
pinentry failed to build: `CPPFLAGS=-I/usr/local/include/ncurses ./configure --prefix=/usr/local` exited with 1
```

packages/pinentry.rb
```diff
@@ -7,6 +7,7 @@ class Pinentry < Package
   source_url 'https://gnupg.org/ftp/gcrypt/pinentry/pinentry-1.0.0.tar.bz2'
   source_sha256 '1672c2edc1feb036075b187c0773787b2afd0544f55025c645a71b4c2f79275a'
 
+  depends_on 'libcap'
   depends_on 'ncurses'
```
cstrouse: Please add the missing dependencies:

```
depends_on 'libgpgerror'
depends_on 'libassuan'
```

Alanaktion:
I've added the missing dependencies. I hadn't noticed that failure since I assumed it would be installed after `gnupg`, but this makes more sense.

uberhacker:
Should `gnupg` be a dependency also?

cstrouse:
Yes, should have included that one in my suggestion sorry.

uberhacker:
No worries at all @cstrouse.  I appreciate your review and concern to get this pushed through.

flyingP0tat0:
Just adding `gnupg` would be enough, the other libraries are dependencies of gpg.

uberhacker:
Excellent suggestion @flyingP0tat0.

Alanaktion:
The pinentry program can technically be used without gnupg, but it's not a common usage. I've updated the dependencies.

PR #1232, opened 2017-09-20
Abstract prefix, libdir and localstatedir configure options

uberhacker:
This commit updates all `a*.rb` packages.


cstrouse:
xmlclitools needs this as well

PR #1236, opened 2017-09-21
Update mutt to build with smtp/imap

heavyimage:
This allows mutt to work with gmail


jam7:
Thank you for submitting PR.  Please increase version number like `version '1.8.3-1'` to let crew know it is new package.  You also need to add dependencies for libraries you specified like `depends_on 'openssl'` and `depends_on 'perl'`.  `openssl` is needed in order to use `--with-ssl`.  `perl` is probably missing dependency from the original mutt.rb.

heavyimage:
Okay -- I think I've addressed your concerns and more (but this has gotten a bit more complicated!) -- let me know what I can do to help :-)

packages/libcyrussasl.rb
```diff
@@ -0,0 +1,22 @@
...
+class Libcyrussasl < Package
+  description 'Simple Authentication and Security Layer (SASL) is a specification that describes how authentication mechanisms can be plugged into an application protocol on the wire. Cyrus SASL is an implementation of SASL that makes it easy for application developers to integrate authentication mechanisms into their application in a generic way.'
+  homepage 'https://www.cyrusimap.org/sasl'
+  version '2.1.26'
+  source_url 'ftp://ftp.cyrusimap.org/cyrus-sasl/cyrus-sasl-2.1.26.tar.gz'
+  source_sha256 '8fbc5136512b59bb793657f36fadda6359cae3b08f01fd16b3d406f1345b7bc3'
+
+  def self.build
+    system './configure',
+           '--prefix=/usr/local',
```
uberhacker: Can you please change this to `--prefix=#{CREW_PREFIX}`?

packages/mutt.rb
```diff
@@ -20,15 +20,20 @@ class Mutt < Package
...
 
+  depends_on 'perl'
   depends_on 'libxslt'
+  depends_on 'openssl'
+  depends_on 'libcyrussasl' => :build
+  depends_on 'gdbm'
 
   def self.build
-    system './configure --with-mailpath=/usr/local/mail --prefix=/usr/local'
+    system './configure --with-mailpath=/usr/local/mail --prefix=/usr/local --enable-imap --with-ssl --enable-smtp --with-sasl=/usr/local/lib/sasl2 --enable-hcache'
```
uberhacker: Can you please change this to:
```
system "./configure \
             --prefix=#{CREW_PREFIX} \
             --enable-hcache \
             --enable-imap \
             --enable-smtp \
             --with-ssl \
             --with-mailpath=#{CREW_PREFIX}/mail \
             --with-sasl=#{CREW_LIB_PREFIX}/sasl2"
```

packages/mutt.rb
```diff
@@ -20,15 +20,20 @@ class Mutt < Package
...
     system 'make'
   end
 
   def self.install
     system "mkdir -p #{CREW_DEST_DIR}#{CREW_PREFIX}/mail"
```
uberhacker: This can be simplified as `system "mkdir -p #{CREW_DEST_PREFIX}/mail"`.

packages/mutt.rb
```diff
@@ -20,15 +20,20 @@ class Mutt < Package
...
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "echo 'export SASL_PATH=/usr/local/lib/sasl2' >> ~/.bashrc"
```
uberhacker: Please move this line to the postinstall section and mention to the user that they need to add the `SASL_PATH` environment variable in their shell.  Something like:
```
  def self.postinstall
    puts
    puts "echo 'export SASL_PATH=#{CREW_LIB_PREFIX}/sasl2' >> ~/.bashrc && source ~/.bashrc".lightblue
    puts
  end
```

cstrouse:
This fails to install on aarch64 (ARMv8). See #1351.

PR #1237, opened 2017-09-24
Add binclock package

uberhacker:
Ncurses clock, with time displayed in colourful binary.  See https://github.com/JohnAnthony/Binary-Clock.


cstrouse:
Works properly on XE500C13-K01US (x86_64).

cstrouse:

PR #1239, opened 2017-09-24
Add collectl package

uberhacker:
Collectl is a light-weight performance monitoring tool capable of reporting interactively as well as logging to disk.  See http://collectl.sourceforge.net/.


cstrouse:
Works properly on Samsung Chromebook Plus (ARMv8). Also, they've rev'd to 4.2.0.

cstrouse:
4.2.0 also works properly on Samsung Chromebook Plus (aarch64).

PR #1246, opened 2017-09-29
Add stow package

wwwiiilll:
Manage installation of multiple softwares in the same directory tree

### DIFF
--- a/packages/stow.rb
+++ b/packages/stow.rb
@@ -1,0 +1,21 @@
+require 'package'
+
+class Stow < Package
+  description 'Manage installation of multiple softwares in the same directory tree'
+  homepage 'https://www.gnu.org/software/stow/'
+  version '2.2.2'
+  source_url 'https://ftp.gnu.org/gnu/stow/stow-2.2.2.tar.gz'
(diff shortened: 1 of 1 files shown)


uberhacker:
@wwwiiilll: Thank you for the new package!  Can you please abstract the prefix with `system "./configure", "--prefix=#{CREW_PREFIX}"`?  Also, if there are any libraries installed those will need to be abstracted as well.  Something like `system "./configure", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}"`.
Thanks

packages/stow.rb
```diff
@@ -0,0 +1,21 @@
+require 'package'
+
+class Stow < Package
+  description 'Manage installation of multiple softwares in the same directory tree'
+  homepage 'https://www.gnu.org/software/stow/'
+  version '2.2.2'
+  source_url 'http://ftp.gnu.org/gnu/stow/stow-2.2.2.tar.gz'
```
uberhacker: Might be a good idea to use https instead of http.

wwwiiilll:
Changed some stuff, I'll test it again later today, especially for the Perl modules stuff

wwwiiilll:
Looks good now

PR #1249, opened 2017-09-29
Update erlang from 20.0 to 20.1

cstrouse:

### DIFF
--- a/packages/erlang.rb
+++ b/packages/erlang.rb
@@ -3,28 +3,20 @@ require 'package'
 class Erlang < Package
   description 'Erlang is a programming language used to build massively scalable soft real-time systems with requirements on high availability.'
   homepage 'http://www.erlang.org/'
-  version '20.0'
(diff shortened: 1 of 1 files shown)


packages/erlang.rb
```diff
@@ -3,28 +3,20 @@
...
-       i686: 'cd06eacaa56bba7fd3ec72840f5e6e1c9e893684837b8af2b889832a212c539d',
-     x86_64: '807f78fae8ec997e891a443f982ccad5820b42777ec6c7020660a89597e4d368',
   })
 
   depends_on 'flex' => :build
 
   def self.build
     system 'export ERL_OTP=`pwd`'
-    system './configure --prefix=/usr/local'
+    system "./configure --prefix=#{CREW_PREFIX}"
```
uberhacker: Does this package install any libraries?  You can find out with the new `crew files erlang` command.  Those should be abstracted also via `--libdir=#{CREW_LIB_PREFIX}`.

PR #1256, opened 2017-10-02
Update R from 3.4.1 to 3.4.2

cstrouse:


uberhacker:
We'll need prefix and libdir configure options, correct?
```
system './configure', "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}", '--with-x=no' # X is not available
```

PR #1257, opened 2017-10-02
Update libxml2 from 2.9.4 to 2.9.5

cstrouse:
This is a security and bugfix release. No CVEs referenced.

Tested on XE500C13-K01US.


uberhacker:
Missing `--prefix=#{CREW_PREFIX}` configure option?

PR #1261, opened 2017-10-02
Update tmux from 2.2 to 2.5

cstrouse:


uberhacker:
Please add prefix configure option:
```
   depends_on 'ncurses'
 
   def self.build
-    system "CPPFLAGS=-I#{CREW_PREFIX}/include/ncurses ./configure"
+    system "CPPFLAGS=-I#{CREW_PREFIX}/include/ncurses ./configure --prefix=#{CREW_PREFIX}"
     system 'make'
   end
```

PR #1265, opened 2017-10-02
Update texinfo from 6.4 to 6.5

cstrouse:


uberhacker:
Please add prefix and libdir configure options:
```
   def self.build
     # configure and make
     system './configure',
+        "--prefix=#{CREW_PREFIX}",
+        "--libdir=#{CREW_LIB_PREFIX}",
         '--with-external-Text-Unidecode',
         '--with-external-Unicode-EastAsianWidth'
```

PR #1276, opened 2017-10-03
Add mdp package

cstrouse:
Mdp is a command-line based markdown presentation tool.

### DIFF
--- a/packages/mdp.rb
+++ b/packages/mdp.rb
@@ -1,0 +1,24 @@
+require 'package'
+
+class Mdp < Package
+  description 'A command-line based markdown presentation tool.'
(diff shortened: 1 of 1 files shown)


packages/mdp.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'ncurses'
+
+  def self.build
+    system "CPPFLAGS=-I#{CREW_PREFIX}/include/ncursesw make"
```
uberhacker: Don't we need prefix and libdir options?  Might not need libdir if it doesn't install any library files.

PR #1277, opened 2017-10-04
Fix htop+protobuf dependencies for build only

bjoern-r:
mark build time dependencies of htop and protobuf pakages so they dont get istalled if we install from binary

### DIFF
--- a/packages/htop.rb
+++ b/packages/htop.rb
@@ -8,16 +8,8 @@ class Htop < Package
   source_sha256 '179be9dccb80cee0c5e1a1f58c8f72ce7b2328ede30fb71dcdf336539be2f487'
 
   binary_url ({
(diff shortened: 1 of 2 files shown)


uberhacker:
@bjoern-r: Thank you for the update.  Can you please bump the versions so the changes are detected?  Something like `2.0.2-1` for htop and `3.3.0-2` for protobuf.
Thanks

bjoern-r:
if the binaru_url and binary_sha256 values are removed the package need to be build on client side right? at leas that is what happend after i removed this section. 
My intention was to fix the package so that binary only installation without uneeded dependency wpuld be possible..

uberhacker:
That is correct.  However, since we need to add the prefix configure options, a recompile is necessary.  We need the binary_url and binary_sha256 values removed so we can replace with new versions.  Please review #1082 to get a better understanding of how this works.

PR #1282, opened 2017-10-08
Add wget package

cstrouse:
Wget is a network utility to retrieve files from the Web.

---

We may want to create packages for libpsl and libidn2 and then add those to the dependencies for this package. Any reason why we would not need or want to do that?

### DIFF
--- a/packages/wget.rb
+++ b/packages/wget.rb
@@ -1,0 +1,29 @@
+require 'package'
+
+class Wget < Package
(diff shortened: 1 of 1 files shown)


uberhacker:
I suppose it would be better than waiting on Google to update.  If you want to go ahead and add those library dependencies, I can wait to merge this.

cstrouse:
I think it's important for people to have the choice of running the current version or of sticking with the one that Google bundles.

The package seems to work properly without the additional packages.

PR #1284, opened 2017-10-08
Add intltool package

uberhacker:
intltool is a set of tools to centralize translation of many different file formats using GNU gettext-compatible PO files.  See https://freedesktop.org/wiki/Software/intltool/.


cstrouse:
Works on both Samsung Chromebook 3 (x86_64) and Samsung Chromebook Plus (aarch64).

PR #1285, opened 2017-10-08
Add perl_xml_parser package

uberhacker:
Perl XML::Parser - A perl module for parsing XML documents.  See https://metacpan.org/pod/XML::Parser.

### DIFF
--- a/packages/perl_xml_parser.rb
+++ b/packages/perl_xml_parser.rb
@@ -1,0 +1,30 @@
+require 'package'
+
+class Perl_xml_parser < Package
+  description 'Perl XML::Parser - A perl module for parsing XML documents'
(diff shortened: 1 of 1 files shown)


cstrouse:
Installs and works with example code from CPAN docs on Samsung Chromebook Plus (ARMv8).

cstrouse:
Fails to build/install on x864_64 without adding `expat` as a dependency.

uberhacker:
Good catch.  We actually need perl as a dependency also.  Duh...right?

cstrouse:
Works on Samsung Chromebook Plus (aarch64) and Samsung Chromebook 3 (x86_64).

PR #1286, opened 2017-10-08
Add libsndfile package

uberhacker:
Libsndfile is a C library for reading and writing files containing sampled sound (such as MS Windows WAV and the Apple/SGI AIFF format) through one standard library interface.  See http://www.mega-nerd.com/libsndfile/.

### DIFF
--- a/packages/libsndfile.rb
+++ b/packages/libsndfile.rb
@@ -8,7 +8,10 @@ class Libsndfile < Package
   source_sha256 '1ff33929f042fa333aed1e8923aa628c3ee9e1eb85512686c55092d1e5a9dfa9'
 
   def self.build
-    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
+    system "./configure \
+            --prefix=#{CREW_PREFIX} \
+            --libdir=#{CREW_LIB_PREFIX} \
+            --disable-dependency-tracking"
     system "make"
   end
 


packages/libsndfile.rb
```diff
@@ -0,0 +1,18 @@
...
+
+class Libsndfile < Package
+  description 'Libsndfile is a C library for reading and writing files containing sampled sound (such as MS Windows WAV and the Apple/SGI AIFF format) through one standard library interface.'
+  homepage 'http://www.mega-nerd.com/libsndfile/'
+  version '1.0.28'
+  source_url 'http://www.mega-nerd.com/libsndfile/files/libsndfile-1.0.28.tar.gz'
+  source_sha256 '1ff33929f042fa333aed1e8923aa628c3ee9e1eb85512686c55092d1e5a9dfa9'
+
+  def self.build
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
```
cstrouse: It makes sense to add `--disable-dependency-tracking` here.

cstrouse:
I noticed in the build output the following:

```bash
checking for flac >= 1.3.1 ... no
checking for ogg >= 1.1.3 ... no
checking for vorbis >= 1.2.3 ... no
checking for vorbisenc >= 1.2.3 ... no

configure: WARNING: *** One or more of the external libraries (ie libflac, libogg and
configure: WARNING: *** libvorbis) is either missing (possibly only the development
configure: WARNING: *** headers) or is of an unsupported version.
configure: WARNING: ***
configure: WARNING: *** Unfortunately, for ease of maintenance, the external libs
configure: WARNING: *** are an all or nothing affair.
```

How useful is this package without the external audio libraries?

uberhacker:
I'll add the missing libraries giving the warning messages and see if still works since this is a dependency package of pulseaudio.

cstrouse:
Also, do you know how we can test that it works properly? `make check` fails with a bunch of errors and I'm not sure how to get it to play audio as there's no option for passing a device path to `sndfile-play`.

`sndfile-info` works properly, however:
```bash
chronos@localhost /usr/local/tmp $ sndfile-info /usr/local/lib/python2.7/test/audiodata/pluck-pcm16.wav
========================================
File : /usr/local/lib/python2.7/test/audiodata/pluck-pcm16.wav
Length : 13370
RIFF : 13362
WAVE
fmt  : 16
  Format        : 0x1 => WAVE_FORMAT_PCM
  Channels      : 2
  Sample Rate   : 11025
  Block Align   : 4
  Bit Width     : 16
  Bytes/sec     : 44100
LIST : 90
  INFO
    INAM : Pluck
    IART : Serhiy Storchaka
    ICMT : Audacity Pluck + Wahwah
    ICRD : 2013
data : 13228
End

----------------------------------------
Sample Rate : 11025
Frames      : 3307
Channels    : 2
Format      : 0x00010002
Sections    : 1
Seekable    : TRUE
Duration    : 00:00:00.300
Signal Max  : 32768 (0.00 dB)
```

uberhacker:
My original intention for pulseaudio was to provide another i/o option for the `cmus` package.  However, this package is still very much a work in progress.  Definitely hold off from merging PR #1121 until we can solve these dependency issues first.

PR #1287, opened 2017-10-08
Add speexdsp package

uberhacker:
Speex is an Open Source/Free Software patent-free audio compression format designed for speech.  See https://speex.org/.

### DIFF
--- a/packages/speexdsp.rb
+++ b/packages/speexdsp.rb
@@ -8,7 +8,12 @@ class Speexdsp < Package
   source_sha256 '4ae688600039f5d224bdf2e222d2fbde65608447e4c2f681585e4dca6df692f1'
 
   def self.build
-    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
+    system "./configure \
+            --prefix=#{CREW_PREFIX} \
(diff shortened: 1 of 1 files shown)


packages/speexdsp.rb
```diff
@@ -0,0 +1,18 @@
...
+
+class Speexdsp < Package
+  description 'Speex is an Open Source/Free Software patent-free audio compression format designed for speech.'
+  homepage 'https://speex.org/'
+  version '1.2rc3'
+  source_url 'http://downloads.xiph.org/releases/speex/speexdsp-1.2rc3.tar.gz'
+  source_sha256 '4ae688600039f5d224bdf2e222d2fbde65608447e4c2f681585e4dca6df692f1'
+
+  def self.build
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
```
cstrouse: We should probably add `--disable-maintainer-mode` here as well as `--disable-dependency-tracking` and possibly `--disable-examples`.

cstrouse:
Works on both Samsung Chromebook Plus (aarch64) and Samsung Chromebook 3 (x86_64).

PR #1291, opened 2017-10-12
Add package for ncat which is a 'netcat on steroids'

minektur:
It has a bunch of extra features...  Binary package for x86_64 included

Note that this is included with the nmap source code, but not built/installed by our package.

It makes sense to have a separate package for this apart from nmap.

similarly, I have an nping package from the same suite that I'm going to submit a PR for.

The x86_64 binary package is currently hosted on my personal website, but I'd be happy to have you upload it to some official place and I'll change the URL


uberhacker:
Thank you for the new package @minektur and welcome back!  We have started to include all our binary packages at Bintray.  See https://github.com/skycocker/chromebrew/issues/1082#issuecomment-325168555.  We're also abstracting the install location and configure options with crew constants.  Please change to something like below:
```
  binary_url ({
  })
  binary_sha256 ({
  })

  depends_on 'openssl' 
  depends_on 'filecmd' => :build   #configure uses file 

  def self.build
    #fixup "/usr/bin/file" -> "#{CREW_PREFIX}/bin/file" in the configure scripts
    system "sed -i s,/usr/bin/file,#{CREW_PREFIX}/bin/file,g libdnet-stripped/configure"
    system "cd ncat && make"
  end
```
Thanks

minektur:
Ah - I see - I'll update and resubmit PRs.

On the bintray thing.... someone controls access to that?  how do I get my
binaries into bintray?  I kind of figured that whoever has access would get
a copy from my URL,  verify the sha256 hash, and then upload to the real
place, and then... either I or they would update the URL in the package?

or do I misunderstand?  Please explain how it is supposed to work.  I may
update some of the docs in a future PR because it wasn't terribly clear on
the wiki/in the tools/README.md

uberhacker:
Hey @minektur: All great questions.  I don't know the best answers to give you.  Ultimately, we would like package committers like yourself to not have to worry about binaries at all.  There are basically 3 different Chromebook platforms that require separate binaries for each and it would be great if every package supported all architectures.  We have another Github repository at https://github.com/chromebrew that we would like to set up with a build process using CI which would automate the binary package compilation process.  We would also like to automate package testing.  Bintray has an API that we could use to integrate with Github.  If you are interested and would like to help, I can send you an invitation to join the Github and Bintray projects.

minektur:
So is there a CREW_DEST_XXXX variable that represents where files will eventually end up and not where they temporarily get installed?  I have the following build-time command to patch a configure script:

    system "sed -i s#/usr/bin/file#/usr/local/bin/file#g libdnet-stripped/configure"

the first /usr/bin/file is what is in the script now - and the /usr/local/bin/file is the theoretical place where the chromebrew filecmd will be installed.  There is no other reference to "/usr/local" here and changing /usr/local/bin/file to #{CREW_DEST_PREFIX}/bin isn't the right thing because that is not where the bin command will be found as already installed...

In this case I can perhaps assume that the /usr/local/bin or wherever-else chromebrew has been installed in is already in the user's path and just replace /usr/bin/file with file and let their path take care of it...

Should I also change the binary_url path https://dl.bintray.com/chromebrew/chromebrew/ncat-7.60-chromeos-x86_64.tar.xz   and let you guys upload? 

As an aside, I have the ability to get involved with the CI efforts you mention but not the personal time...  I'm a devops guy for my job, so I'm very familiar with automating things...  but I will probably be radio silent for big chunks of time.

minektur:
I screwed up my git tree and so I made a new branch and new PR:  Leaving this one open for a short while because the discussion is here....

https://github.com/skycocker/chromebrew/pull/1295

uberhacker:
I hate when that happens. :)

minektur:
Ok - is this what you're looking for?  We need better docs :)

packages/ncat.rb
```diff
@@ -0,0 +1,33 @@
...
+    x86_64: '5660ea321436f1b9642d318a77e9b0b5c379825ac0611daaf5f82dd794753598',
+  })
+
+  depends_on 'buildessential' => :build
+  depends_on 'filecmd' => :build   #configure uses file 
+
+  def self.build
+    #fixup "/usr/bin/file" -> "file" in the configure script
+    
+    system "sed -i s#/usr/bin/file##{CREW_PREFIX}/bin/file#g libdnet-stripped/configure"
```
uberhacker: I think we will want a different delimiter than `#`, correct?  If this currently works, I would be surprised.

packages/ncat.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    #without-zenmap in configure removes openssl dependency
+    system "./configure --without-zenmap --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
```
uberhacker: This looks great.

packages/ncat.rb
minektur: it works ... I tested it.... I confess that when I inserted the ruby variable in the string I was concerned - sed can use plenty of other delimiters....  but it does work...  originally I picked something other than the normal / because I wanted to swap out paths with that character in it... # is always my goto secoind choice with regex :)

packages/ncat.rb
minektur: Ill send those other 3 PRs this evening.

uberhacker: Can you please post the output from `crew files ncat` when you get a chance?  I'm curious which files this package installs.

packages/ncat.rb
uberhacker: I guess if the replacement occurs prior to the regex evaluation, it would work.  Interesting discovery.

packages/ncat.rb
minektur: I must have an old enough 'crew' binary that I dont have the files command.... but I did a 'find /usr/local' before and after - and this is the (cleaned up) diff of the output:

> /usr/local/etc/crew/meta/ncat.directorylist
> /usr/local/etc/crew/meta/ncat.filelist
> /usr/local/bin/ncat
> /usr/local/share/man/man1/ncat.1
> /usr/local/share/ncat
> /usr/local/share/ncat/ca-bundle.crt

uberhacker: OK great.  So we can drop the `--libdir=#{CREW_LIB_PREFIX}` configure option then.  BTW, if you `cat /usr/local/etc/crew/meta/ncat.filelist`, this essentially does the same thing as `crew files ncat`.  If you execute `crew update` followed by `crew upgrade`, you can get the latest updated versions of all packages including the crew command.

minektur: you want me to remove the --libdir option?  it doesn't hurt even though the package doesn't install libraries...

uberhacker: Yes, although it doesn't hurt to leave it, it would be better to remove it so we can tell just by looking at the configure options that we don't have any libraries.

uberhacker:
@minektur: I'll be impressed if you can figure out a way to include the `armv7l` and `i686` binaries also.

minektur:
uberhacker if you can build them, I'll include them - I lack the hardware to do the build...  Well, I might be able to borrow an armv7l but it might be difficult to dev-mode, install, etc and then wipe to return it... I guess old arm chromebooks are showing up on craiglist for really cheap.... one of these days....

minektur:
Ok - I've removed the libdir option as it is unneeded.  Should be ready for a merge...

uberhacker:
Thank you @minektur.  We can add the other binaries later.

PR #1292, opened 2017-10-12
Add nping network diagnostic tool from nmap suite

minektur:
This is another cool tool from the nmap suite.

Binary package for x86_64 provided (hosted on my website)


uberhacker:

PR #1293, opened 2017-10-12
add filecmd binary package for x86_64

minektur:
adding a binary package... figured it would be nice since I use it in the ncat/nping stuff and configure uses it all over the place


uberhacker:

PR #1295, opened 2017-10-12
add ncat package - netcat on steroids - from the nmap suite

minektur:
Here is what I was talking about  - this removes the reference to the 'eventual install directory' of filecmd which needs to be preinstalled for configure to be happy...

and changes the URL to point bintray, where you can upload it.

The original is at: 

    https://www.clift.org/fred/chromebrew/ncat-7.60-chromeos-x86_64.tar.xz

Does that work?  If so, I'll make new PRs for the other packages

### DIFF
--- a/packages/ncat.rb
+++ b/packages/ncat.rb
@@ -1,0 +1,31 @@
+require 'package'
+
+class Ncat < Package
+  description 'Ncat - Netcat for the 21st Century - a modern netcat with extra features from the makers of nmap.'
+  homepage 'https://nmap.org/ncat'
+  version '7.60'
+  source_url 'https://nmap.org/dist/nmap-7.60.tar.bz2'
+  source_sha256 'a8796ecc4fa6c38aad6139d9515dc8113023a82e9d787e5a5fb5fa1b05516f21'
+
+  binary_url ({
+    x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/ncat-7.60-chromeos-x86_64.tar.xz',
+  })
+  binary_sha256 ({
+    x86_64: '5660ea321436f1b9642d318a77e9b0b5c379825ac0611daaf5f82dd794753598',
+  })
+
+  depends_on 'buildessential' => :build
+  depends_on 'openssl' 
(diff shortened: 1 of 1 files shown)


uberhacker:
That might work but it's probably safer to include the full path of `#{CREW_PREFIX}/bin`.  We also need the `--prefix=#{CREW_PREFIX}` and `--libdir=#{CREW_LIB_PREFIX}` configure options if there are any library files stored in `/usr/local/lib`.  Can you please merge these changes into your original PR #1291 and then close this PR?

minektur:
I'm not sure what you mean, or I didn't explain myself well enough.  I need a variable that points to the actual eventual install destination of an already installed package.


Are you saying that #{CREW_PREFIX}/bin is where I can expect to find the already installed filecmd?

uberhacker:
Eventually, all compiled from source packages should include the `#{CREW_PREFIX}` constant to abstract the install location which is how we can install in a directory other than `/usr/local`.  The default is currently `/usr/local` which probably won't affect the current install.  This request basically just "future proofs" current packages.  I do agree we need better docs.  I'm guessing you want to volunteer? ;)

minektur:
I can probably whip up some quick "getting started" how to make a new package guide...  "real soon now"...  is my most recent edit what you're hoping for?  I'll change the other packages ready asap...  

Is there some drive/desire to have the rest of the pacakges updated similarly?  It seems like it would be pretty easy to do a big set of them in one big PR....

As for the design, I assume the plan is that for CHROMEBREW_PREFX=/usr/local we can have binary packages, but if you want to install somewhere else you'd have to build from source?    I wouldn't think that binaries built intended to go into one dir would necessarily work in another...

uberhacker:
I did create a starter PR #1232 but the other collaborators have been sleeping on my request. :)  Eventually, when we migrate this project to https://github.com/chromebrew, my plan was to install all the binaries in a directory such as `/usr/local/share/chromebrew` so the uninstall will be painless.  This would require the `PATH` environment variable to be modified to include this new directory in the search, however.  But you are correct.  If you want to install binaries in a different directory, it would require recompiling all the packages.

PR #1297, opened 2017-10-13
Nping: a network diagnostic tool from the Nmap suite

minektur:
the binary referred to in the package is currently hosted at 

https://www.clift.org/fred/chromebrew/ncat-7.60-chromeos-x86_64.tar.xz

if you want to drop it on bintray...


minektur:
arg - branch problems again... I'll close and reopen once I unhork my tree

PR #1300, opened 2017-10-15
Add rlwrap package

uberhacker:
rlwrap is a 'readline wrapper', a small utility that uses the GNU readline library to allow the editing of keyboard input for any command. I couldn't find anything like it when I needed it, so I wrote
this one back in 1999.  By now, there are (and, in hindsight, even then there were) a number of good readline wrappers around, like rlfe, distributed as part of the GNU readline library, and the amazing socat (http://freecode.com/projects/socat). You should consider rlwrap especially when you need user-defined completion (by way of completion word lists) and persistent history, or if you want to program 'special effects' using the filter mechanism. rlwrap compiles and runs on a fairly wide range of Unix-like systems.  See https://github.com/hanslub42/rlwrap.

### DIFF
--- a/packages/rlwrap.rb
+++ b/packages/rlwrap.rb
@@ -1,0 +1,26 @@
+require 'package'
+
+class Rlwrap < Package
+  description 'A readline wrapper'
+  homepage 'https://github.com/hanslub42/rlwrap'
+  version '0.43'
+  source_url 'https://github.com/hanslub42/rlwrap/archive/v0.43.tar.gz'
+  source_sha256 '29e5a850fbe4753f353b0734e46ec0da043621bdcf7b52a89b77517f3941aade'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'autoconf'
+
+  def self.build
+    system "autoreconf --install"
+    system "./configure --prefix=#{CREW_PREFIX}"
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+end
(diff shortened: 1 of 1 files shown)


cstrouse:
Install is failing for me on x86_64.

```
chronos@localhost /usr/local/lib/crew/packages $ crew install rlwrap
rlwrap: A readline wrapper
https://github.com/hanslub42/rlwrap
version 0.43
autoconf 
Do you agree? [Y/n] y
Proceeding...
autoconf: Autoconf is an extensible package of M4 macros that produce shell scripts to automatically configure software source code packages.
http://www.gnu.org/software/autoconf/
version 2.69
Archive downloaded
Installing...
Performing post-install...
Autoconf installed!
Archive downloaded
gcc (GCC) 4.9.4

Can't exec "aclocal": No such file or directory at /usr/local/share/autoconf/Autom4te/FileUtils.pm line 326.
autoreconf: failed to run aclocal: No such file or directory
rlwrap failed to install: `autoreconf --install` exited with 1
```

uberhacker:
@cstrouse: Thanks for the feedback.  Apparently, we also need automake as a dependency.  Please see if adding automake fixes this issue.

cstrouse:
Works with `automake` added.

uberhacker:
Hey @cstrouse: I was wondering the same about @lyxell and @thedamian.  They must be busy.

lyxell:
I'm busy with studies for the time being, you can always ping me if you want my opinions though.

uberhacker:
What is your opinion on 30+ PRs that need merged? :)

lyxell:
My opinion is that we obviously need more collaborators. :)

cstrouse:
@lyxell Can you merge this so the repl isn't awful under Clozure CL?

PR #1302, opened 2017-10-16
Update bacon from 3.5.4 to 3.6

cstrouse:

### DIFF
--- a/packages/bacon.rb
+++ b/packages/bacon.rb
@@ -3,33 +3,27 @@ require 'package'
 class Bacon < Package
   description 'BaCon is a free BASIC to C translator for Unix-based systems.'
   homepage 'http://www.basic-converter.org/'
-  version '3.5.4'
-  source_url 'http://www.basic-converter.org/stable/bacon-3.5.4.tar.gz'
-  source_sha256 '7b1c72fd46daaa43d19e1bfac2f9bcd9decc5b8443d8f5640e903bfc35e122b9'
+  version '3.6'
+  source_url 'http://www.basic-converter.org/stable/bacon-3.6.tar.gz'
+  source_sha256 'c09d7a69ac1f9dd4c607ca5248e2b20a91c279779bef7a8887e93b77f982f855'
 
   binary_url ({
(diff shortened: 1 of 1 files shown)


packages/bacon.rb
```diff
@@ -3,33 +3,27 @@
...
-     armv7l: '3415da0b1303b42ef2b2b4385a9aec58f02ae237b05e857f4bbb1807290a5b41',
-       i686: '2cf7f4c550c50898fbadeabb792e2dc22423532e20241967e193c97fd138d767',
-     x86_64: '023778b5d5ccbbd276bd97a49077633ca5c8c249778b0cfce2a720d74f0efc66',
   })
 
-  def self.build
-    system "./configure --prefix=/usr/local --disable-gui"
-    system 'sed -i "45s,/usr/share,/usr/local/share," Makefile'
-    system 'sed -i "46s,/usr/share,/usr/local/share," Makefile'
+  depends_on 'buildessential'
```
uberhacker: We could probably include this line on every package, however, there is a pending PR #1217 that will essentially make this unnecessary.  Since that PR hasn't been merged yet, this is fine.  Just be aware that this requirement will likely go away in the future.

cstrouse: Noted. Thanks for letting me know as I had not looked at #1217 yet.

packages/bacon.rb
```diff
@@ -3,33 +3,27 @@
...
 
-    # force to compile in sequential since bacon Makefile doesn't work in parallel
-    system "make", "-j1"
+  def self.build
+    system 'sed -i "s,/usr/share,\$\(DATADIR\)," Makefile.in'
```
uberhacker: Do we need a global replacement here?  Just asking since there were multiple sed lines earlier.

cstrouse: The updated version had another broken Makefile that would have required another modification so I just switched to operating on Makefile.in instead of Makefile and to globally replace those paths to use $(DATADIR) as he does elsewhere.

I'm going to send a patch to the Makefile.in upstream which should remove the need for this when the next update happens.

uberhacker: @cstrouse: We need more guys like you that help fix these upstream issues.  Keep up the awesome work.

PR #1303, opened 2017-10-17
Add nping from the nmap suite - a network diagnostic tool.

minektur:
Another tool from the nmap team.

(you can find a copy of my x86_64 package at https://www.clift.org/fred/chromebrew/nping-7.60-chromeos-x86_64.tar.xz  to upload to bintray)

### DIFF
--- a/packages/nping.rb
+++ b/packages/nping.rb
@@ -1,0 +1,37 @@
+require 'package'
+
+class Nping < Package
+  description 'Nping is part of the nmap suite - a network diagnostics tool`.'
+  homepage 'https://nmap.org/nping/'
+  version '7.60'
+  source_url 'https://nmap.org/dist/nmap-7.60.tar.bz2'
+  source_sha256 'a8796ecc4fa6c38aad6139d9515dc8113023a82e9d787e5a5fb5fa1b05516f21'
+
+  binary_url ({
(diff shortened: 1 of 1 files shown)


packages/nping.rb
```diff
@@ -0,0 +1,37 @@
...
+    
+    system "sed -i s#/usr/bin/file##{CREW_PREFIX}/bin/file#g libdnet-stripped/configure"
+
+    #without-zenmap in configure removes openssl dependency
+    system "./configure --without-zenmap --prefix=#{CREW_PREFIX}"
+
+    # Make files in the project dont have the right dependencies to build this by itself
+    # so just make the whole thing and then install nping - not effecient, but
+    # more likely to work than a bunch of Makefile dependency edits
+    system "make"
```
uberhacker: Do you want to `system "cd nping && make"`.  I'm only asking because your make install has `cd nping`.

minektur: The make files are missing some dependencies, which makes making nping separately problematic.  There are 3 options 

1. manually running make in several peer directories in the right order
2. editing the make files to add the necessary dependencies
3. building the whole thing

Building the whole thing is what I chose - Pros:  more likely to be "correct" because this is the normal way it gets built, probably less package maintaince in the long run because it's simple.  Cons:  It takes a little longer to build from source, and it doesn't look obviously correct as you noticed.  (I did add the comment to try to explain)

PR #1305, opened 2017-10-17
fix build to match new ENV requirements, fix filecmd for configure

minektur:
has new configure option to match new env var standard, and a fix for borked filecmd which configure depends on...

### DIFF
--- a/packages/nmap.rb
+++ b/packages/nmap.rb
@@ -20,10 +20,13 @@ class Nmap < Package
      x86_64: '12eb815ec4bc47ac0d778575c7125907421937396fa491745d5ab48486380550',
   })
 
-  depends_on 'buildessential'
+  depends_on 'buildessential' => :build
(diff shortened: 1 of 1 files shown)


minektur:
The version upstream hasn't changed...  I dont know if changing the install prefix prefix will make a difference in the binary.

Do we have a standard for bumping build versions when the upstream hasn't changed?

As we update packages to use the install prefix are we going to make new binary packages for all?

uberhacker:
Right, but the build has so we should bump the version.  Something like `7.60-1` would work.  The reason we want to do this is in case somebody already has the previous version installed, a `crew update` will detect changes and enable that user to update to the new version via `crew upgrade`.  Eventually, I would like to get all the packages updated to use crew constants so we can compile everything to use another prefix besides `/usr/local`.

minektur:
Ok I bumped the version...

PR #1308, opened 2017-10-19
Add jfrog package

uberhacker:
JFrog CLI is a compact and smart client that provides a simple interface that automates access to Artifactory, Bintray and Mission Control through their respective REST APIs.  See https://github.com/JFrogDev/jfrog-cli-go.

### DIFF
--- a/packages/jfrog.rb
+++ b/packages/jfrog.rb
@@ -12,7 +12,6 @@ class Jfrog < Package
   end
 
   def self.install
-    system "mkdir -p #{CREW_DEST_PREFIX}/bin"
-    system "cp jfrog #{CREW_DEST_PREFIX}/bin"
+    system "install -Dm755 jfrog #{CREW_DEST_PREFIX}/bin/jfrog"
   end
 end


packages/jfrog.rb
```diff
@@ -0,0 +1,18 @@
...
+  source_url 'https://github.com/JFrogDev/jfrog-cli-go/archive/1.11.2.tar.gz'
+  source_sha256 'b57f85205097c194d23dae07b225b492839ddf941e0711e7ebdc3e44b13d195d'
+
+  def self.build
+    system "bash getCli.sh"
+  end
+
+  def self.install
+    system "mkdir -p #{CREW_DEST_PREFIX}/bin"
+    system "cp jfrog #{CREW_DEST_PREFIX}/bin"
```
cstrouse: This can be simplified into a single step as
```ruby
system "install -Dm755 jfrog #{CREW_DEST_PREFIX}/bin/jfrog"
```

cstrouse:
The binary doesn't run properly on ARMv8.

```bash
chronos@localhost /usr/local/lib/crew/packages $ jfrog
bash: /usr/local/bin/jfrog: cannot execute binary file: Exec format error
```

We might want to add `go` as a build-time dependency and build from source so that things work properly on all architectures.

uberhacker:
Replaced with #1370.

PR #1312, opened 2017-10-20
Update go from 1.9 to 1.9.1

cstrouse:

### DIFF
--- a/packages/go.rb
+++ b/packages/go.rb
@@ -3,9 +3,9 @@ require 'package'
 class Go < Package
   description 'Go is an open source programming language that makes it easy to build simple, reliable, and efficient software.'
   homepage 'https://golang.org/'
-  version '1.9'
-  source_url 'https://storage.googleapis.com/golang/go1.9.src.tar.gz'
-  source_sha256 'a4ab229028ed167ba1986825751463605264e44868362ca8e7accc8be057e993'
+  version '1.9.1'
+  source_url 'https://storage.googleapis.com/golang/go1.9.1.src.tar.gz'
+  source_sha256 'a84afc9dc7d64fe0fa84d4d735e2ece23831a22117b50dafc75c1484f1cb550e'
 
   binary_url ({
   })


uberhacker:
Thanks for the update @cstrouse.  Can we clean the build and install sections up a bit?  We could change `#{CREW_PREFIX}/lib` to `#{CREW_LIB_PREFIX}` and `#{CREW_DEST_DIR}#{CREW_PREFIX}` to `#{CREW_DEST_PREFIX}`.

cstrouse:
Does this make sense even though `go_bootstrap` is installing binaries there which aren't actually libraries and so probably don't belong under `/usr/local/lib` or `/usr/local/lib64`?

uberhacker:
That makes me think that maybe we shouldn't be installing binaries in `/usr/local/lib`?  Is that where they belong?  What happens if you take `/usr/local/lib` out of the library search path `LD_LIBRARY_PATH` environment variable on a x86_64 build?  Does it install and execute without error?

uberhacker:
@cstrouse: I'm thinking we should install go in `/usr/local/share` instead of `/usr/local/lib` and move all the stuff installed in `$HOME` to there as well.  Thoughts?

maazghani:
Could the default golang installation path `/usr/local` work here? Source: [Golang documentation](https://golang.org/doc/install#tarball)

Resulting `$GOROOT` would be `/usr/local/go`.

uberhacker:
@maazghani: While I would normally agree with the official documentation, in this case I think they're "wrong". :)  IMHO, I think the `/usr/local` directory should only include subdirectories such as bin, lib, include, share, etc. and not be "cluttered".  Most projects install in `/usr/local/share` and go should be no different in order to keep things organized.

uberhacker:
@cstrouse: Any update on this PR?

cstrouse:
They've rev'd to 1.9.2 in the meantime. The holdup has been that it doesn't build properly on aarch64. I've been unable to get it to build and am thinking of having it use the official ARM binaries for the time being. Any objections?

thedamian:
@cstrouse binaries sound better than compiling it if it works. If you do that, my friend, then I'll merge it!

uberhacker:
@thedamian: I can take this PR over and generate the binaries if you like.  Do you have an aarch64 system you can test?

thedamian:
No @uberhacker I don't.
So if you're taking this over I think I should merge it and you'd make the changes?
I'd suggest a quick fix is just get the old version of the binary for the aarch64 people. so at least they have an installable build.

I'll wait for your answer before I merge.

uberhacker:
@thedamian: Go ahead and merge this.  I think we have an update anyway but the changes here are minimal and don't affect the current situation we are in with this package anyway.

thedamian:
merged

PR #1313, opened 2017-10-20
Update nginx from 1.13.5 to 1.13.6

cstrouse:


uberhacker:
Thank you for the update @cstrouse.  Really appreciate your efforts.  We could change `#{CREW_DEST_DIR}#{CREW_PREFIX}` to `#{CREW_DEST_PREFIX}` in the install section.

cstrouse:
Done but I squashed the commits so we can have a cleaner history. We should probably start doing that more often to avoid clutter in the history from making these minor tweaks after review.

uberhacker:
I agree with not cluttering up the commit history but if you want to squash, it might be helpful to include comments in the final commit message that mention all the changes from the squashed commits.

PR #1317, opened 2017-10-23
Update manpages from 4.12 to 4.13

uberhacker:
- Update manpages and mandb packages to include a postinstall section
- Abstract installation directories with Chromebrew constants


cstrouse:

PR #1322, opened 2017-10-25
Update coreutils from 8.27 to 8.28

uberhacker:
- Add arch command.  This is to override the existing `/usr/bin/arch` command to use `/usr/local/bin/uname -m`.


cstrouse:
Works properly on Samsung Chromebook 3 (x86_64).

PR #1323, opened 2017-10-25
Add dfc package

uberhacker:
dfc displays file system space usage using graphs and colors. In some ways, it is a modernized version of df as it is able to use colors, draw graphs and export its output to different formats such as CSV or HTML.  See https://projects.gw-computing.net/projects/dfc.


cstrouse:
Works on Samsung Chromebook Plus (ARMv8).

PR #1330, opened 2017-10-27
Adding elinks package

Alanaktion:
This adds the [ELinks](http://elinks.cz/) web browser. I'm using the latest 0.13 git release since the stable releases are very old and missing a lot of features and modern architecture support.

One thing that may not work with this package is that the `source_url` file is rebuilt every day, so the hash likely won't match. Would it be a good idea to mirror this file somewhere else so the hash stays valid?

### DIFF
--- a/packages/elinks.rb
+++ b/packages/elinks.rb
@@ -3,10 +3,10 @@ require 'package'
 class Elinks < Package
   description 'Full-Featured Text WWW Browser'
   homepage 'http://elinks.or.cz/'
-  version 'f86be65971'
+  version 'f86be659'
 
-  source_url 'https://git.phpizza.com/alan/elinks/archive/f86be659718c0cd0a67f88b42f07044c23d0d028.tar.gz'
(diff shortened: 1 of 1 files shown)


uberhacker:
@Alanaktion: Thank you for the new package.  There is a github project with the source.  You could try something like:
```
require 'package'

class Elinks < Package
  description 'Full-Featured Text WWW Browser'
  homepage 'http://elinks.or.cz/'
  version 'f4a58b'

  source_url 'https://github.com/nabetaro/elinks/archive/f4a58ba3b574a478fd5954ba2c5b29e8b809ff9b.tar.gz'
  source_sha256 'f53679e736d102ea65aeb18b1f15014fc65b398df912f901a2bfe592911110cb'

...
```

Alanaktion:
We could use that repository, but it is two years (48 commits) behind the upstream. I'll mirror it myself with a service that'll keep it up-to-date and still allow downloading archives of specific commits.

uberhacker:
That is a good idea.  Github has lots of tools that could help automate something like that.

Alanaktion:
I've updated it to use a GitLab mirror that should stay up-to-date automatically.

PR #1338, opened 2017-10-29
Update screen from 4.6.1 to 4.6.2

cstrouse:
This is bugfix and maintenance release.

Tested as working on Samsung Chromebook Plus (ARMv8).

### DIFF
--- a/packages/screen.rb
+++ b/packages/screen.rb
@@ -3,31 +3,24 @@ require 'package'
 class Screen < Package
   description 'Screen is a full-screen window manager that multiplexes a physical terminal between several processes, typically interactive shells.'
   homepage 'https://www.gnu.org/software/screen/'
-  version '4.6.1'
-  source_url 'ftp://ftp.gnu.org/gnu/screen/screen-4.6.1.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/screen.rb
```diff
@@ -3,31 +3,24 @@
 class Screen < Package
   description 'Screen is a full-screen window manager that multiplexes a physical terminal between several processes, typically interactive shells.'
   homepage 'https://www.gnu.org/software/screen/'
-  version '4.6.1'
-  source_url 'ftp://ftp.gnu.org/gnu/screen/screen-4.6.1.tar.gz'
-  source_sha256 'aba9af66cb626155d6abce4703f45cce0e30a5114a368bd6387c966cbbbb7c64'
+  version '4.6.2'
+  source_url 'ftp://ftp.gnu.org/gnu/screen/screen-4.6.2.tar.gz'
```
uberhacker: Can we change this to `https://ftp.gnu.org/gnu/screen/screen-4.6.2.tar.gz`?

cstrouse: Sure.

PR #1340, opened 2017-10-30
Update php5 from 5.6.31-1 to 5.6.32

uberhacker:
- Also added crew constants and --libdir configure option.


cstrouse:

PR #1345, opened 2017-10-31
Adding tree dependency to pass

Alanaktion:
The pass program requires tree to list the stored passwords correctly.


uberhacker:
@Alanaktion: Thank you for the PR.  Can you please bump the version so the changes can be detected?  Something like `version '1.7.1-1'`.  Thanks

Alanaktion:
Ah, right. I've updated the version number.

PR #1346, opened 2017-11-01
Disable MPI assembler modules for ARMv8

cstrouse:
The included MPI assembler code doesn't work for the ARMv8 processor in
the Samsung Chromebook Plus (and possibly others). Until upstream adds
support for these processors they must be disabled for the package to
build properly.

In the future support will probably be added for these processors
enabling us to re-enable support for MPI ASM modules.

Closes #1344

Issue #1123, opened 2017-08-21
Version number of each package file


jam7:
The purpose of this issue is to make a decision and write FAQ about this topic later.

Question is whether we need to modify version number in package files when we modify them.  I think there are 3 options we can choose.

1. Modify version number whenever you modify package files.
2. Modify version number if and only if you modification changes behavior of the package.
3. Don't modify version number.

1 is the most simple solution.  So, if you are not sure, choosing this is the best.
2 is good to reduce unnecessary upgrade.  For example, adding dependencies to make it compile is this.  Who already make it install don't need to upgrade.
3 is bad.  So, no need to choose I think.

How do you think?  Should we choose 1 whenever we modify package files?  I prefer 2 since it reduces unnecessary upgrade and we can still choose 1 if we are not sure.

uberhacker:
@jam7: I think 1 is probably the best choice since we can't always assume the user has installed the package.  There is also the case of updating postinstall messages.  In these cases, although it doesn't directly change the package behavior (except if the message is instructions that modify the user environment), you still want the user to see the new messages after they upgrade.

jam7:
@uberhacker Can you explain more details?  I don't understand what kind of trouble you are assuming.

lyxell:
If we choose 1 it is probably better to skip version numbers and just use a sha256-sum of the .rb-file. This way versioning is handled automatically.

uberhacker:
@lyxell: I like automatic version handling. :)

jam7:
Automatic version handling may requires a lot of unnecessary downloads of pre-built binary files.  Bintray limits the amount of OSS free account transfer to 1TB/month.  I'm worried about it.

lyxell:
1 TB sounds like a lot to me, do we have some kind of estimation on how many downloads the most popular packages could expect to have per month?

uberhacker:
@lyxell: I imagine every package in `buildessential` will get the most action.  You can examine the statistics here: https://bintray.com/chromebrew/chromebrew.  For example, to find out the popularity of `gcc`, see https://bintray.com/chromebrew/chromebrew/gcc#statistics.

jam7:
Please check this https://bintray.com/chromebrew/chromebrew-cross/gcc#statistics.

JL2210:
I'm fairly sure this can be closed. We chose 1.

JL2210:
@uberhacker

PR #1352, opened 2017-11-02
Use Ruby ENV for libgcrypt build

Alanaktion:
This fixes issues with building libgcrypt on aarch64 without manually setting an environment variable. See #1346.

I didn't bump the version number here since the package would not have built properly on the affected architecture. That seems correct to me until a decision is reached on #1123.


cstrouse:
LGTM.

PR #1357, opened 2017-11-02
Remove sudo from crew commands in glibc package

uberhacker:
Replaces #1216.


cstrouse:
Works on Samsung Chromebook 3 (x86_64).

PR #1220, opened 2017-09-16
Update tests/README.md

uberhacker:
Updated instructions based on my experience attempting to run tests.

Issue #1307, opened 2017-10-17
new doc idea....


minektur:
I'm thinking that I should write up my workflow I use when making a bunch of package updates, combination of github interface, git commands, branching, PRs etc.... 

I have a very very rough outline here:

https://gist.github.com/minektur/ad88076a179b8c6ceb12fec329506d79

I'm thinking of whipping it up into a fully-fleshed out MD to augment/replace our package build docs on the wiki....

Is this approximately what you guys do?  Do you have any feedback on the outline/process before I write up details and flesh it out...

uberhacker:
@minektur: This is a great start.  Thank you for taking the time to put this together.  I definitely agree we need better docs.  Instead of lines 10 and 11, I would just do `git pull upstream master`.  You could start out by creating your new package in `/usr/local/lib/crew/packages` and then copy back to the main `packages` directory when you are ready to commit your changes.  You might want to mention a step after `crew install -s mycoolpackage` to examine the output of `crew files mycoolpackage` and determine if there are any library files that should be abstracted.  For instance, on a x86_64 system, you may end up with files installed in `/usr/local/lib` instead of `/usr/local/lib64`.  This would be an indicator that you need the `--libdir=#{CREW_LIB_PREFIX}` configure option.  Likewise, if you don't notice any library files, you could remove the `--libdir=...` configure option.  After this, you should probably `crew remove mycoolpackage` and make sure all the files that were installed initially are actually removed.  If you don't see any files from `crew files mycoolpackage` after installation, this is a good indicator that the install failed even if it was reported as installed successfully.  You will not likely need to git checkout specific commits.  The way I would start working on a new package is something like below:
```
$ git checkout master
$ git pull upstream master
$ git checkout -b mynewpackage
```
This will work if you don't make any local commits on your master branch.  Keep your master branch intact and updated regularly with the upstream and you should be fine.

Issue #997, opened 2017-08-02
Implementing Tests


yancouto:
This issue is to discuss whether tests should be implemented.

- Should we create tests?
- Which kind of tests? (In #963 I implemented a test to check if there is a dependency cycle)
- Should some kind of testing framework be used?
- Could they be automated? (Travis?)

jam7:
In my opinion

- yes, definitely
- there are two.  to check the correctness of .rb file including dependencies and to check the correctness of compile/install
- we have check function to self.check the correctness of compile/install.  see go.rb, perl.rb, python27.br, etc.  we didn't have any for .rb file correctness.
- yes.  I'd like to try some soon.

jam7:
I have a question about automated test execution environment for chromebrew.  Is there any good tools we can use free...
 - to check correctness of .rb files like #963?
 - to check and build pre-compiled binaries for at least x86_64?
 - to check and build pre-compiled binaries for other architecture?
Thanks.

yancouto:
In Travis it is possible to automatically run tests like #963.

Building stuff from source is possible, but I'm not sure if is possible to do in those specific architectures, I've only used it in small projects.

jam7:
I see.  Just a test with several scripts should be easy.  Maybe we can try some of them.  Then, we may try building stuff later.  I'd like to try some in near future definitely.

uberhacker:
@jam7: I found a few executables available in our very own package system that may be helpful.  The `fslint` package will assist in identifying issues with files.  I ran it recently on my system and found this section interesting:
```
-------------------------Non Stripped executables
-rwxr-xr-x 1 chronos chronos    14104 Aug  2 20:58 /usr/local/bin/checkgid
-rwxr-xr-x 1 chronos chronos    23624 Aug  2 20:58 /usr/local/bin/httxt2dbm
-rwxr-xr-x 1 chronos chronos    27464 Aug  2 20:58 /usr/local/bin/htdigest
-rwxr-xr-x 1 chronos chronos    27712 Aug  2 20:58 /usr/local/bin/logresolve
-rwxr-xr-x 1 chronos chronos    29464 Aug  2 20:58 /usr/local/bin/rotatelogs
-rwxr-xr-x 1 chronos chronos    36184 Aug  2 20:58 /usr/local/bin/htdbm
-rwxr-xr-x 1 chronos chronos    39368 Aug  2 20:58 /usr/local/bin/htpasswd
[...]

PR #1363, opened 2017-11-03
Adding more documentation to tests

yancouto:
As @cstrouse suggested in #1220. Is this fine? Anything still needs better explaining?

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -8,12 +8,6 @@ Running these tests will make sure that crew is working correctly, or explain th
 
 The tests can be used at any time to help with developing, but make sure to run them before submitting a Pull Request.
 
-## How to run tests
-
-Execute `ruby test_all` from this directory (`tests/`). If all tests pass, it will print `All tests successful.`, otherwise it will show what went wrong.
-
-Running tests manually may not work!
-
 ## Preparing new tests
 
 To prepare a new test, create a `.rb` file. It may use crew constants (created in `lib/const.rb`) and import packages (be sure to use `require_relative`).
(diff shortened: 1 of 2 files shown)


yancouto:
@uberhacker Any suggestions?

uberhacker:
@yancouto: I agree with @cstrouse.  I definitely think there is value in having local tests for new packages and updates.  You might take a look at #997 and #1307.  Also, the suggestions in #1220.
Thanks

cstrouse:
Thanks @yancouto that's a great improvement. Thanks for your help in both developing the first testing and for improving the documentation along with @uberhacker.

uberhacker:
@yancouto: Yes, this is a huge improvement and we really appreciate your efforts.  Is it possible to explain where the tests should be run (like which directory)?  Also, I'm thinking it might make sense if you swapped the order of the `Preparing new tests` and `How to run tests` sections.  If you can provide a complete example test suite, that would be even better.
Thanks

yancouto:
> Is it possible to explain where the tests should be run (like which directory)?

Done.

> Also, I'm thinking it might make sense if you swapped the order of the Preparing new tests and How to run tests sections.

I don't think so, if this README is for newcomers, they should worry more about running the tests than creating new. But if you wish, I can change it.

> If you can provide a complete example test suite, that would be even better.

What do you mean exactly? There is an example test in the README.
Or do you mean creating more real tests?

uberhacker:
I just thought it might be helpful to provide a complete set of tests as an example to explain how a practical test suite should be generated.  Also, I was thinking if you rearranged the order chronologically, it might be a bit easier to understand.  Kind of more like a step-by-step guide.

yancouto:
Ok, switched the sections.

I'll not create more tests for now though, I have no good ideas, that one was because of the problem we had.

uberhacker:
OK, I'll merge this for now.  We can add more content later as we get more familiar with writing tests.

PR #1364, opened 2017-11-03
Command suggestions

yancouto:
When you mistype a command, it shows similar commands.

For example
```
$ crew updstr
Could not understand "crew updstr".
Did you mean?
  update
Usage:
  crew build [-k|--keep] <name> ...
  ...
```

### DIFF
--- a/crew
+++ b/crew
@@ -11,6 +11,7 @@ require 'digest/sha2'
 require 'json'
 require 'fileutils'
 require_relative 'lib/const'
+require_relative 'lib/util'
 
 # Add lib to LOAD_PATH
 $LOAD_PATH.unshift "#{CREW_LIB_PATH}lib"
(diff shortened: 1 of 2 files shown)


uberhacker:
Nice! :+1:

crew
```diff
@@ -51,6 +52,16 @@ require_relative 'lib/docopt'
 begin
   args = Docopt::docopt(DOC)
 rescue Docopt::Exit => e
+  puts "Could not understand \"crew #{ARGV.join(' ')}\".".lightred
+  cmds = ["build", "download", "files", "help", "install", "remove", "search", "update", "upgrade", "whatprovides"]
```
uberhacker: I'm wondering if we can use this array generically so we only need to manage all the commands in one place.  It might even be a good idea to store it in the config directory like `CREW_CONFIG_PATH + 'device.json'` with all the other metadata like each help command message strings and even the docopt usage section.  It would require some serious refactoring, however.

yancouto: Yes, that would be good. While that isn't implemented, I think this is the simplest solution though :P

uberhacker: Agreed but we should keep this thought in the back of our minds moving forward.

PR #1365, opened 2017-11-03
Update jdk8 from 8u144 to 8u152

uberhacker:


cstrouse:
Works on both Samsung Chromebook 3 (x86_64) and Samsung Chromebook Plus (aarch64).

uberhacker:
Unable to squash commits.

PR #1366, opened 2017-11-03
Update jdk8 from 8u144 to 8u152

uberhacker:
- Change urls from http to https
- Change install location, crew constants and get manpages working
- Combine aarch64 and armv7l installs


cstrouse:
The Samsung Chromebook Plus (and possibly other aarch64 models) ships with a 32-bit glibc. The 64-bit version of the JDK requires a 64-bit glibc so we're stuck using the 32-bit until Google changes what they ship.

PR #1370, opened 2017-11-04
Add jfrog package

uberhacker:
JFrog CLI is a compact and smart client that provides a simple interface that automates access to Artifactory, Bintray and Mission Control through their respective REST APIs. See https://github.com/JFrogDev/jfrog-cli-go.


cstrouse:
Works on Samsung Chromebook 3 (x86_64).

PR #1288, opened 2017-10-08
Add mp4v2 package

uberhacker:
MP4v2 Library: This library provides functions to read, create, and modify mp4 files.  See https://code.google.com/archive/p/mp4v2/.

PR #1373, opened 2017-11-05
Add mp4v2 package

uberhacker:
MP4v2 Library: This library provides functions to read, create, and modify mp4 files. See https://code.google.com/archive/p/mp4v2/.  Replaces #1288.


cstrouse:
Works on Samsung Chromebook Plus (aarch64) and Samsung Chromebook 3 (x86_64).

PR #1376, opened 2017-11-05
Abstract paths with crew constants in xmlclitools package

uberhacker:
Hopefully, this works on aarch64 as well.


cstrouse:

PR #1377, opened 2017-11-05
Add zeromq package

cstrouse:
ZeroMQ is a high-performance asynchronous messaging library, aimed at
use in distributed or concurrent applications.

Tested as working on Samsung Chromebook Plus (aarch64). All tests pass.

### DIFF
--- a/packages/zeromq.rb
+++ b/packages/zeromq.rb
@@ -1,0 +1,29 @@
+require 'package'
+
+class Zeromq < Package
+  description 'Fast messaging system built on sockets. C and C++ bindings. aka 0MQ, ZMQ.'
+  homepage 'http://zeromq.org/'
+  version '4.2.2'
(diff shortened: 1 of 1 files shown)


cstrouse:
Builds properly on Samsung Chromebook 3 (x86_64) but all 79 tests fail.

uberhacker:
Should I hold off on the merge until you can figure out how to execute the tests error free on x86_64?

skiman6010:
I think this brings up and interesting point. If all 79 tests fail, but the package still build properly and works, then are our tests not sufficient? 

It seems to me like we should overhaul the tests to make sure this doesn't happen.

flyingP0tat0:
This is about ZMQ's own tests, right? We only have one test for crew...

cstrouse:
Yes this is about their tests. It may not be a big deal as it could be a bug in their tests but should be investigated before merging.

uberhacker:
Installs fine on Hisense C11 (armv7l).

uberhacker:
Installs fine on CloudReady (x86_64).

PR #1388, opened 2017-11-08
Add crew -V|--version and -v|--verbose options

### LINKED ISSUE #1385: crew -v|--version should display the version
Currently, `crew -v|--version` does not work.  However, to complicate matters, the `-v` switch is also used to get verbose output from `crew search <package> -v`.  I propose we repurpose the `-v|--verbose` switch to something like `-x|--expand` instead so we can get `crew -v|--version` working.

uberhacker:
Closes #1385.

### DIFF
--- a/crew
+++ b/crew
@@ -20,23 +20,24 @@ DOC = <<DOCOPT
 Chromebrew - Package manager for Chrome OS http://skycocker.github.io/chromebrew/
 
 Usage:
-  crew build [-k|--keep] <name> ...
-  crew download <name> ...
+  crew build [-k|--keep] [-v|--verbose] <name> ...
+  crew download [-v|--verbose] <name> ...
   crew files <name> ...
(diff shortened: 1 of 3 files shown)


skiman6010:
When testing this commit, I get the following error when typing `crew [command] `: 
```
/usr/local/bin/crew:40:in `<main>': uninitialized constant CREW_VERSION (NameError)
Did you mean?  CREW_NPROC
```

Seems to be an issue with const.rb for me.

uberhacker:
Before you test, make sure `lib/const.rb` and `lib/package.rb` are copied to `/usr/local/lib/crew/lib` and crew should be copied to `/usr/local/bin`.

skiman6010:
They were, but I nuked my crew installation and started from scratch and then it worked fine.

cstrouse:
LGTM.

PR #1389, opened 2017-11-08
Package: mongodb - fix to ensure we use the right src code tag to compile mongo-tools

jamesjnadeau:
I was having issues getting mongodb to install, I think this is a fix for my issue as the mongo-tools src is out of sync with what we are installing. It's checking out the latest and greatest code when we should be getting a tagged version.

haven't figured out how to test this locally, but that's up next, should work...

### DIFF
--- a/packages/mongodb.rb
+++ b/packages/mongodb.rb
@@ -29,8 +29,8 @@ class Mongodb < Package
   def self.build
     # prepare mongo-tools
     system "git clone https://github.com/mongodb/mongo-tools"
-    system "git checkout tags/r3.2.16"
-    system "cd mongo-tools; ./build.sh ssl || echo '***YOU NEED TO SET UP GO ENVIRONMENT IN ORDER TO COMPILE MONGO-TOOLS***'"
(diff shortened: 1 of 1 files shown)


jamesjnadeau:
ok, made some updates after testing, works on my machine ;)

packages/mongodb.rb
```diff
@@ -29,7 +29,8 @@ class Mongodb < Package
   def self.build
     # prepare mongo-tools
     system "git clone https://github.com/mongodb/mongo-tools"
-    system "cd mongo-tools; ./build.sh ssl || echo '***YOU NEED TO SET UP GO ENVIRONMENT IN ORDER TO COMPILE MONGO-TOOLS***'"
+    system "cd mongo-tools; git checkout tags/r3.2.16; ./build.sh ssl || echo '***YOU NEED TO SET UP GO ENVIRONMENT IN ORDER TO COMPILE
+MONGO-TOOLS***'"
```
uberhacker: Hey @jamesjnadeau: Thank you for the update.  It might be better if you used `&&` instead of `;` to separate the commands.  This will ensure that if the git checkout is not successful, it won't continue and try to execute build.sh.  Also, since this is a modification, please bump the version to something like `3.2.16-1` so users that have installed this version can get this update applied.  Otherwise, the change is not detected.
Thanks

jamesjnadeau:
ok @uberhacker, your requested changes are up

PR #1398, opened 2017-11-11
Update httpd from 2.4.27 to 2.4.29

uberhacker:
- Add configure options
- Add postinstall section with instructions


cstrouse:

PR #1399, opened 2017-11-11
Add wput package

cstrouse:
wput is a command line file upload tool, the opposite of wget.

### DIFF
--- a/packages/wput.rb
+++ b/packages/wput.rb
@@ -1,0 +1,28 @@
+require 'package'
+
+class Wput < Package
+  description 'wput is a command line file upload tool, the opposite of wget'
+  homepage 'http://wput.sourceforge.net/'
+  version '0.6.1'
+  source_url 'https://prdownloads.sourceforge.net/wput/wput-0.6.1.tgz'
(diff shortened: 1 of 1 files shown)


packages/wput.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def self.preinstall
+    system "curl -Ls -o config.guess 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD'"
+    system "curl -Ls -o config.sub 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD'"
+  end
+
+  def self.build
+    system './configure',
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}"
```
uberhacker: Do we need anything like `--disable-maintainer-mode` or `--disable-debugger-checking`?  If those options are not available, no worries.

cstrouse: They're not available.

PR #1403, opened 2017-11-11
Add octane package

uberhacker:
The JavaScript Benchmark Suite for the modern web https://developers.google.com/octane/.  See https://github.com/chromium/octane.

### DIFF
--- a/packages/octane.rb
+++ b/packages/octane.rb
@@ -1,0 +1,28 @@
+require 'package'
+
+class Octane < Package
+  description 'The JavaScript Benchmark Suite for the modern web https://developers.google.com/octane/.'
+  homepage 'https://github.com/chromium/octane'
+  version '2.0'
+  source_url 'https://github.com/chromium/octane/archive/570ad1ccfe86e3eecba0636c8f932ac08edec517.tar.gz'
+  source_sha256 'c355dd9a37df421826f3d8028d89e8c8b58faa12a49da88fa2641788f9482fe7'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.install
+    system "mkdir -p #{CREW_DEST_PREFIX}/share/octane"
+    system "cp -r . #{CREW_DEST_PREFIX}/share/octane"
+  end
+
+  def self.postinstall
+    puts
+    puts "To perform a benchmark test, execute the following:".lightblue
+    puts "ruby -run -e httpd #{CREW_PREFIX}/share/octane/ -p 8080".lightblue
+    puts "Visit http://localhost:8080".lightblue
+    puts "Click 'Start Octane 2.0'".lightblue
+    puts
+  end
+end


cstrouse:
This is an unmaintained project that no longer serves a purpose according to the developers. Also, it says v1 but the readme for the project says it's v2; any idea why the disconnect? Also, wouldn't running the webserver on the machine being benchmarked skew the scores? Why not just run it from their website?

flyingP0tat0:
Agreed, there is no need for such a package.

cstrouse:
If @uberhacker has a reason that he thinks it's a package worth having vs running it on the website (which offers a 2.0 version) then we'll get it into the packages. I just thought this one might warrant some discussion.

uberhacker:
I thought it might be helpful to compare benchmarks locally vs over the web since the bandwidth could play a factor in the benchmark results.  Also, this is version 1 which can give a comparison with the website version 2.

cstrouse:
I would imagine that the test loads client-side and then runs so bandwidth wouldn't matter as much. I'll get this tested shortly.

packages/octane.rb
```diff
@@ -0,0 +1,36 @@
...
+    system "mkdir -p #{CREW_DEST_PREFIX}/share/nginx/html/octane"
+    system "cp -r . #{CREW_DEST_PREFIX}/share/nginx/html/octane"
+  end
+
+  def self.postinstall
+    system "ps ax | grep httpd | grep -v grep | head -1 2> /dev/null > /tmp/httpd.pid"
+    if not File.zero? "/tmp/httpd.pid" then
+      puts
+      puts "Make sure httpd is not running on port 80".lightblue
+      puts "sudo apachectl stop".lightblue
```
cstrouse: Shouldn't this be controlling nginx instead of apache?

packages/octane.rb
```diff
@@ -0,0 +1,36 @@
...
+  version 'v1'
+  source_url 'https://github.com/chromium/octane/archive/v1.tar.gz'
+  source_sha256 '51e9002e8391f65809c19200e44277d0e5a0adf9f7c03af46676cd321f965d68'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'nginx'
```
cstrouse: Can we do this via Ruby's webrick, python's built-in webserver, or darkhttpd to make this a smaller install?

packages/octane.rb
uberhacker: This logic only tests if Apache is using port 80 since it would create a conflict with Nginx.

packages/octane.rb
uberhacker: I suppose that would be worth trying to see if it has any play in the performance factors.  We could also run php as a server as well.

cstrouse: Yeah but php is pretty big and most people won't have it installed. We already have ruby installed and most people will have python as well. Just a thought.

packages/octane.rb
cstrouse: Can we check this another way? How about netstat?

uberhacker: I just thought it would be easier using port 80 since it wouldn't require any extra configuration.  We could try netstat.  How would that work?

flyingP0tat0:
Bandwidth definitely is not a factor for this benchmark. Didn't intend it to sound rude, but I really don't think that a browser (!) benchmark belongs here.
That said, if you still think this is a good idea, please don't ship it with a webserver. The benchmark consists out of JavaScript and CSS only, so you could just open it in your browser as any other file.

uberhacker:
I compared tests locally and at http://chromium.github.io/octane/ and although the tests varied between runs, the locally run tests were consistently higher scores than the web tests at http://chromium.github.io/octane/ in the range of %10-%20 higher.

uberhacker:
The webserver is needed because Chrome OS restricts access to `file:///...` requests in the browser.  Nginx is lightweight so I thought that was a bit better choice than Apache.  I am certainly open to other options that "just work".

cstrouse:
You can try the darkhttpd package we have which is lighter.

cstrouse:
Using webrick works fine and adds no dependencies. Works on Samsung Chromebook 3 (x86_64).

PR #1404, opened 2017-11-11
Add pdfcrack package

cstrouse:
pdfcrack is a tool for recovering passwords for PDF files.

### DIFF
--- a/packages/pdfcrack.rb
+++ b/packages/pdfcrack.rb
@@ -1,0 +1,22 @@
+require 'package'
+
+class Pdfcrack < Package
+  description 'pdfcrack is a command line, password recovery tool for PDF-files.'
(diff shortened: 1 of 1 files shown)


packages/pdfcrack.rb
```diff
@@ -0,0 +1,22 @@
...
+  })
+  binary_sha256 ({
+  })
+
+  def self.build
+    system 'make'
+  end
+
+  def self.install
+    system "install -Dm755 pdfcrack #{CREW_DEST_PREFIX}/bin/pdfcrack"
```
uberhacker: No manpage or other supporting files?

cstrouse: Just a single binary.

PR #1405, opened 2017-11-11
Update nginx package

uberhacker:
- Fix to get the nginx process id
- Return exit code of 1 on error
- Bump version number to detect changes
- Add zlibpkg dependency


cstrouse:

PR #1408, opened 2017-11-12
Update libpng from 1.6.33 to 1.6.34

uberhacker:


cstrouse:
Works on Samsung Chromebook Plus (aarch64). All 33 tests pass.

PR #1409, opened 2017-11-12
Update cairo package

uberhacker:
- Remove freetype dependency since fontconfig already includes it
- Abstract --prefix and --libdir configure options with crew constants


cstrouse:

PR #1410, opened 2017-11-12
Add zlibpkg dependency to ag package

uberhacker:
- Designate some dependencies as build only

### DIFF
--- a/packages/ag.rb
+++ b/packages/ag.rb
@@ -3,7 +3,7 @@ require 'package'
 class Ag < Package
   description 'The Silver Searcher. Very fast search similar to ack or grep. (ag)'
   homepage 'https://github.com/ggreer/the_silver_searcher'
-  version '2.1.0-1'
+  version '2.1.0-2'
   source_url 'https://github.com/ggreer/the_silver_searcher/archive/2.1.0.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/ag.rb
```diff
@@ -12,11 +12,12 @@ class Ag < Package
...
-  depends_on "pkgconfig"
+  depends_on "autoconf" => :build
+  depends_on "automake" => :build
+  depends_on "pkgconfig" => :build
   depends_on "pcre"
   depends_on "xzutils"
+  depends_on "zlibpkg"
 
   def self.build
     system "autoreconf", "-fiv"
```
cstrouse: Do you know why this is necessary? Usually packagers should not run autoreconf.

uberhacker: I'm guessing because of the autoreconf line.  @vince, can you please provide more details?

cstrouse: I was asking specifically about the autoreconf line. I may have marked the wrong line.

vince: I largely took this recipe from the homebrew package. Could be it's not needed but I stopped fiddling when the package worked.

cstrouse:

PR #1411, opened 2017-11-12
Change --prefix configure option in sqlite package

uberhacker:
Several packages depend on sqlite and do not install correctly unless this fix is in place.


cstrouse:

PR #1412, opened 2017-11-12
Update apriconv from 1.2.1-1 to 1.2.2

uberhacker:
This update is necessary since the 1.2.1 source is no longer available.

### DIFF
--- a/packages/apriconv.rb
+++ b/packages/apriconv.rb
@@ -3,9 +3,9 @@ require 'package'
 class Apriconv < Package
   description 'a portable implementation of the iconv() library'
   homepage 'http://apr.apache.org/'
-  version '1.2.1-1'
-  source_url 'http://apache.claz.org/apr/apr-iconv-1.2.1.tar.bz2'
-  source_sha256 'c46c919bc2a36a705f91f4dea444b18a83236eef97a417528a988113b3a7e46e'
+  version '1.2.2'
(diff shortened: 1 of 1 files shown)


packages/apriconv.rb
```diff
@@ -3,9 +3,9 @@
 class Apriconv < Package
   description 'a portable implementation of the iconv() library'
   homepage 'http://apr.apache.org/'
-  version '1.2.1-1'
-  source_url 'http://apache.claz.org/apr/apr-iconv-1.2.1.tar.bz2'
-  source_sha256 'c46c919bc2a36a705f91f4dea444b18a83236eef97a417528a988113b3a7e46e'
+  version '1.2.2'
+  source_url 'http://apache.claz.org/apr/apr-iconv-1.2.2.tar.bz2'
```
cstrouse: This is available via https. Please switch the protocol.

cstrouse:
This builds without httpd installed. Is that really a necessary dependency?

packages/apriconv.rb
uberhacker: Agreed.  Will fix.

uberhacker:
Well, I suppose it was this statement in the configure script which swayed the decision:
```
By default, `make install' will install all the files in
`/usr/local/apache2/bin', `/usr/local/apache2/lib' etc.  You can specify
an installation prefix other than `/usr/local/apache2' using `--prefix',
for instance `--prefix=$HOME'.
```

PR #1413, opened 2017-11-12
Add libplist package

cstrouse:
libplist is a library for manipulating Apple's Property List file
format.

Tested as working on Samsung Chromebook Plus (aarch64). All tests
passing.


cstrouse:
FYI There's a problem with their SSL cert so http is the only option for downloading the package at this point.

PR #1417, opened 2017-11-13
Update ImageMagick to 7.0.7-11

rumpaar1:
They removed 7.0.7-9 from the server.


packages/imagemagick.rb
```diff
@@ -4,8 +4,8 @@ class Imagemagick < Package
   description 'Use ImageMagick to create, edit, compose, or convert bitmap images.'
   homepage 'http://www.imagemagick.org/script/index.php'
   version '7.0.7-9'
```
uberhacker: Thank you for the update @rumpaar1.  Can you please bump the version to match the source?

rumpaar1: That would be what I forgot to do. Done now.

PR #1424, opened 2017-11-15
Add lzo package

uberhacker:
- LZO is a portable lossless data compression library written in ANSI C.
- Offers pretty fast compression and *extremely* fast decompression.
- One of the fastest compression and decompression algorithms around. See the ratings for lzop in the famous Archive Comparison Test .
- Includes slower compression levels achieving a quite competitive compression ratio while still decompressing at this very high speed.
- Distributed under the terms of the GNU General Public License (GPL v2+).

See http://www.oberhumer.com/opensource/lzo/.

PR #1425, opened 2017-11-15
Add lzop package

uberhacker:
lzop is a file compressor which is very similar to gzip. lzop uses the LZO data compression library for compression services, and its main advantages over gzip are much higher compression and decompression speed (at the cost of some compression ratio).  See http://www.lzop.org/.  Depends on PR #1424.


cstrouse:

PR #1426, opened 2017-11-16
Update aspell_en from 2017.01.22-0 to 2017.08.24-0

uberhacker:


cstrouse:
Builds on Samsung Chromebook Plus (aarch64) but intermittently segfaults when spell checking files.

uberhacker:
That might be a result of the `aspell` package and not this data only update.  Do you get segfaults with the current dictionary version?

cstrouse:
It has intermittent issues with the current dictionary as well but seems to only happen when spell checking markdown-formatted text files. No the end of the work I suppose.

thedamian:
i'm going to approve this and hope the errors are due to something else

PR #1428, opened 2017-11-16
Update python3 from 3.6.0 to 3.6.4

uberhacker:


cstrouse:
Installs fresh on Samsung Chromebook 3 (x86_64).

packages/python3.rb
```diff
@@ -39,12 +31,12 @@ def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
 
     # remove static library
-    system "find #{CREW_DEST_DIR}/usr/local -name 'libpython*.a' -print | xargs -r rm"
+    system "find #{CREW_DEST_PREFIX} -name 'libpython*.a' -print | xargs -r rm"
 
     # create symbolic links in lib64 for other applications which use libpython
```
cstrouse: This should no longer refer to lib64.

thedamian: @cstrouse do you mean it isn't? or it is? do you like this fix or are pointing out something wrong?

uberhacker: @thedamian: I'm guessing he means we should be using `-L#{CREW_LIB_PREFIX}` instead of `-L#{CREW_PREFIX}/lib`.

uberhacker: @thedamian: I understand what @cstrouse was talking about with the upgrade.  You lose pip3 if you perform `crew upgrade python3`.  I will just make sure the pip executables are included in the pre-built binaries and then it will be a non-issue.  Go ahead and merge this when you get a chance and I'll follow it with the binaries PR soon afterwards.
Thanks

PR #1431, opened 2017-11-16
Update icu4c from 58.2 to 60.1

uberhacker:


packages/icu4c.rb
```diff
@@ -28,7 +20,11 @@ def self.build
         # discussed in https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=653457
         system "sed -e '/LDFLAGSICUDT=/cLDFLAGSICUDT=' -i config/mh-linux"
       end
-      system "./configure", "--without-samples", "--without-tests"
+      system "./configure",
+             "--prefix=#{CREW_PREFIX}",
+             "--libdir=#{CREW_LIB_PREFIX}",
+             "--without-samples",
+             "--without-tests"
```
cstrouse: Just curious why you're disabling tests. Are they large? Do they take forever? Do they tend to fail on CrOS?

cstrouse:

packages/icu4c.rb
uberhacker: Tests were disabled before I made this change.  There's probably nothing wrong with doing this as long as the initial tests were successful on all architectures.

cstrouse: I'm not saying we need to enable the tests. I was just curious if you knew what the reason was for them being disabled. No worries.

PR #1432, opened 2017-11-17
Update perl from 5.24.1-3 to 5.26.1

cstrouse:

*NOTE*: This does break vim because it can't find libperl.so. I'm not sure if vim needs to be modified or if this package does. Perl works correctly and all tests pass.

### DIFF
--- a/packages/perl.rb
+++ b/packages/perl.rb
@@ -3,21 +3,13 @@ require 'package'
 class Perl < Package
   description 'Perl 5 is a highly capable, feature-rich programming language with over 29 years of development.'
   homepage 'https://www.perl.org/'
-  version '5.24.1-3'
-  source_url 'http://www.cpan.org/src/5.0/perl-5.24.1.tar.gz'
-  source_sha256 'e6c185c9b09bdb3f1b13f678999050c639859a7ef39c8cad418448075f5918af'
+  version '5.26.1'
+  source_url 'http://www.cpan.org/src/5.0/perl-5.26.1.tar.gz'
+  source_sha256 'e763aa485e8dc1a70483dbe6d615986bbf32b977f38016480d68c99237e701dd'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/perl-5.24.1-3-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/perl-5.24.1-3-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


uberhacker:
@cstrouse: Thank you for the update.  I noticed this was missing the prefix abstraction?  Can we do something like below?  I hate, hate, hate when packages install in `/usr/local/lib`.
```
diff --git a/packages/perl.rb b/packages/perl.rb
index 1f4413c..34a5b8a 100644
--- a/packages/perl.rb
+++ b/packages/perl.rb
@@ -18,7 +18,7 @@ class Perl < Package
     # Use system zlib and bzip2
     # Create shared library
     # Install manual files into /usr/local/share/man/man* even if groff is not installed.
-    system "BUILD_ZLIB=False BUILD_BZIP2=0 ./Configure -de -Duseshrplib -Dman1dir=#{CREW_PREFIX}/share/man/man1 -Dman3dir=#{CREW_PREFIX}/share/man/man3"
+    system "BUILD_ZLIB=False BUILD_BZIP2=0 ./Configure -de -Duseshrplib -Dprefix=#{CREW_PREFIX}/share/perl -Dman1dir=#{CREW_PREFIX}/share/man/man1 -Dman3dir=#{CREW_PREFIX}/share/man/man3"
     system "make"
     system "curl -o cpanm https://cpanmin.us"
     system "chmod +x cpanm"
@@ -28,6 +28,9 @@ class Perl < Package
     system "mkdir -p #{CREW_DEST_PREFIX}/bin/"
     system "cp cpanm #{CREW_DEST_PREFIX}/bin/"
+    FileUtils.cd("#{CREW_DEST_PREFIX}/share/perl/bin") do
+      system "find . -exec ln -s #{CREW_PREFIX}/share/perl/bin/{} #{CREW_DEST_PREFIX}/bin \\;"
+    end
   end
```

cstrouse:
Yeah I'll do that and then we should also figure out how to make it play nice with vim. Now might also be a good time to make sure Perl support in other packages doesn't break also.

thedamian:
like in other Pulls. I LOOOOVE the upgrade. and this is SUPER important. I'm sure VIM can be upgraded and might take that task on. 

BUT We're loosing all the pre-build bin's. What can we do about that. I'm looking for advice, not bins *bins would be nice but I"m trying to open a discussion here.

uberhacker:
@thedamian: I got the pre-built binaries covered.  I'm going to make a strong effort to keep up with them as quickly as possible so we don't get too big of a backlog.  Most of the packages have pre-built binaries now which makes it much more convenient.  No more waiting, waiting, waiting...

PR #1434, opened 2017-11-19
Update qemu from 2.9.0 to 2.10.1

uberhacker:


cstrouse:

PR #1435, opened 2017-11-19
Update freetype from 2.8-1 to 2.8.1

uberhacker:


cstrouse:

PR #1441, opened 2017-11-19
Update libx264 from 20170604-2245-stable to 20171118-2245-stable

uberhacker:


thedamian:
love the upgrade. Hate that we're getting rid of the built in compiled versions.

How can we resolve this? I'll check it in but want to hear some suggestions.

uberhacker:
@thedamian is back!  This is definitely a moving target since these are basically nightly updates.  IIRC, this update was needed as a dependency for ffmpeg.

thedamian:
ok ok. I'm merging because I love your work. I just want to keep the conversation in the perl PR because we have to figure out how to automate the binary builds!

thedamian:
p.s. I"m back cause I just got a Samsung Chromebook Pro to replace my dead CHromebook Pixel (1) It's a shame cause I think the motherboard is dead.

I can't afford it but it's the closest thing to a pixelbook.

uberhacker:
Bummer, Pixels are top-of-the-line.  Can you get it replaced?  The thought of having 16 gb of ram and a 512 gb ssd drive is mouth watering.  I just wish they had a screen bigger than 12.3 inches.  Way too small for my bad eyes. :)  If you have any ideas how we can perform the automated pre-built binary builds, I'm all ears.  The jfrog package which you just merged is a helpful interface to bintray.com.

PR #1443, opened 2017-11-20
Update nettle from 3.3 to 3.4

uberhacker:


cstrouse:

PR #1446, opened 2017-11-20
Add zsync package

cstrouse:
zsync is a client-side file transfer program similar to rsync.

Tested as working on Samsung Chromebook Plus (aarch64). All tests
passing.

### DIFF
--- a/packages/zsync.rb
+++ b/packages/zsync.rb
@@ -1,0 +1,26 @@
+require 'package'
+
+class Zsync < Package
+  description 'zsync is a client-side file transfer program similar to rsync.'
+  homepage 'http://zsync.moria.org.uk/'
+  version '0.6.2'
+  source_url 'https://zsync.moria.org.uk/download/zsync-0.6.2.tar.bz2'
+  source_sha256 '0b9d53433387aa4f04634a6c63a5efa8203070f2298af72a705f9be3dda65af2'
+
+  def self.preinstall
+    system "curl -Ls -o autotools/config.guess 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD'"
+    system "curl -Ls -o autotools/config.sub 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD'"
+  end
+
+  def self.build
+    system './configure',
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}",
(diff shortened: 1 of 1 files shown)


packages/zsync.rb
```diff
@@ -0,0 +1,26 @@
...
+class Zsync < Package
+  description 'zsync is a client-side file transfer program similar to rsync.'
+  homepage 'http://zsync.moria.org.uk/'
+  version '0.6.2'
+  source_url 'https://zsync.moria.org.uk/download/zsync-0.6.2.tar.bz2'
+  source_sha256 '0b9d53433387aa4f04634a6c63a5efa8203070f2298af72a705f9be3dda65af2'
+
+  def self.preinstall
+    system "curl -Ls -o autotools/config.guess 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD'"
+    system "curl -Ls -o autotools/config.sub 'https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD'"
```
uberhacker: Should we checksum these to make sure they downloaded correctly?

thedamian: Yes. Not for security. Just to make sure you didn't try to download it while on a wifi-capture page  or a bad connection.

yes uber
@cstrouse can you add a quick checksum check?

cstrouse: This is something that affects many packages. They're using old versions of config.guess and config.sub that were created prior to the ARMv8-a processors (and possibly others) which aren't able to determine the build/host type.

Most Linux distributions use the files from their libtool package and then copy them into place whenever a package requiring a configure step is built. We should do something along those lines.

cstrouse:
Also, on a side note now that we have squashfs tools once this gets merged we can make use of tinycore binaries like someone suggested in the new features thread.

packages/zsync.rb
uberhacker: Can we add these updated versions of config.guess and config.sub to the libtool package and then wash our hands?

cstrouse: libtool already has current versions of them. We would need to install libtool in base or as a build-time dep anytime a package uses configure.

uberhacker: Can the package be reconfigured to look for these files in the libtool location instead of the current package versions?

cstrouse:
Can we merge this and open an issue regarding the handling of config.guess/config.sub for all packages requiring such a thing?

uberhacker:
Sure, did you want to open the issue?

PR #1448, opened 2017-11-23
Downgrade bashdb package

uberhacker:
According to the instructions, the version of bashdb to use has to be compatible with the version of bash used.  I have checked `bash --version` on several Chromebooks and the standard version shipped with Chrome/Chromium OS appears to be 4.3.42(1).  This version may change on updates so we should watch and update accordingly.


thedamian:
I hate that this has to be done, because now we have to remember to upgrade when bash is upgraded :(

I'll check it in!

PR #1450, opened 2017-11-24
Add syncthing package

allenbenz:
Syncthing is a free, open-source peer-to-peer file synchronization application. See https://syncthing.net/.

Tested on a Samsung Chromebook Pro (x86_64)

### DIFF
--- a/packages/syncthing.rb
+++ b/packages/syncthing.rb
@@ -23,6 +23,12 @@ class Syncthing < Package
     system "mkdir -p #{CREW_DEST_PREFIX}/bin"
     system "mv bin/syncthing #{CREW_DEST_PREFIX}/bin/syncthing-bin"
 
+    [ 1, 5, 7 ].each do |i|
+        system "mkdir -p #{CREW_DEST_PREFIX}/share/man/man#{i}"
+        system "gzip man/*.#{i}"
+        system "cp -R man/*.#{i}.gz #{CREW_DEST_PREFIX}/share/man/man#{i}"
+    end
+
     #syncthing requires some ports to be open
     system %Q(echo '#!/bin/bash
 if [[ $EUID == 0 ]]; then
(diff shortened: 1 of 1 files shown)


uberhacker:
@allenbenz: Thank you for the new package.  Is the `syncthing` executable the only file we need from the build?  Are there any help files or other supporting files?

allenbenz:
@uberhacker The binary (and now the wrapper script) are the only files needed.

There are some man pages and an upstart script. But man isn't installed so I assumed they wouldn't be needed. And the web gui has a help link that sends you to the documentation.

I realized I had set up some iptables rules to make local network discovery work.
```
/sbin/iptables -I INPUT -p tcp --dport 22000 -j ACCEPT
/sbin/iptables -I INPUT -p udp --dport 21025 -j ACCEPT
```
So I added that wrapper script to the install.

uberhacker:
@allenbenz: If you can include those man pages also, that would be helpful.  You can install man by executing `crew install mandb`.

allenbenz:
@uberhacker man pages are now included in the install.

uberhacker:
@allenbenz: Excellent work.  We will need another small change:
```
- depends_on 'go' => build
+ depends_on 'go' => :build
```
This is necessary because the binary install will still require the go package the way it is now.

@cstrouse: Can you please check this on aarch64 when you get a chance?  I tested on armv7l and it works well.

allenbenz:
@uberhacker the :build change was made

uberhacker:
@allenbenz: I'm waiting on @cstrouse to check out how this works on aarch64 architecture when he gets a chance.  From my testing experience, it works fine on i686, x86_64 and armv7l.

cstrouse:
The Go package doesn't compile on aarch64 still so I can't test this unless I manually install the official Go armv7 binaries. I can do that this evening.

uberhacker:
@cstrouse: Did you get a chance to test this on aarch64?

PR #1451, opened 2017-11-25
Update stack from 1.4.0 to 1.5.1

uberhacker:
This package will not need pre-built binaries since the source is basically installing the binary already.  Not sure how we make a note of this during the automated build process.

### DIFF
--- a/packages/stack.rb
+++ b/packages/stack.rb
@@ -3,32 +3,26 @@ require 'package'
 class Stack < Package
   description 'The Haskell Tool Stack - Stack is a cross-platform program for developing Haskell projects. It is aimed at Haskellers both new and experienced.'
   homepage 'https://docs.haskellstack.org/en/stable/README/'
-  version '1.4.0'
+  version '1.5.1'
 
   case ARCH
+  when 'aarch64', 'armv7l'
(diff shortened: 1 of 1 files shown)


packages/stack.rb
```diff
@@ -3,32 +3,27 @@
...
-     armv7l: '5d372b2453b33a0e22e929e50b0eef9893c881b65a6a00f8a3c722ede57a9bf4',
-       i686: '2b1103372503767cf3982ce07a2fc9f7ea97bc3e4cee0e27703285f4a9561eea',
-     x86_64: '85be48d282df764cec7139ffd6e2c9f4b05e437b5b24cba371cdc5dab982c41f',
   })
 
   def self.install
-    system "mkdir -p #{CREW_DEST_DIR}/usr/local/bin"
-    system "cp stack #{CREW_DEST_DIR}/usr/local/bin"
+    system "mkdir -p #{CREW_DEST_PREFIX}/bin"
+    system "cp stack #{CREW_DEST_PREFIX}/bin"
```
cstrouse: Can you also swap these two lines out to use the install command?

cstrouse:

PR #1452, opened 2017-11-25
Add kotlin package

uberhacker:
Statically typed programming language for modern multiplatform applications.  See https://kotlinlang.org/.


cstrouse:
Works on Samsung Chromebook Plus (aarch64). I was able to compile and run the hello world program from their docs.

PR #1457, opened 2017-11-27
Add glib dependency to irssi

uberhacker:


cstrouse:

PR #1464, opened 2017-11-30
piping 'y' to `crew install` to handle prompt: 'do you agree? y/n'

MatrixManAtYrService:
I'm not sure what buildessential wanted me to agree to, but the install script got hung up asking me if I wanted to agree.  I realized after the fact that I was pretty much done with the install and could simply have installed buildessential interactively, but at the time I thought it was hung up earlier.

This change says yes to the wayward prompt, which allows installation to complete.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -185,8 +185,8 @@ echo lib >> .git/info/sparse-checkout
 echo crew >> .git/info/sparse-checkout
 git fetch origin master
 git reset --hard origin/master
-crew install buildessential
-crew install less
+yes | crew install buildessential
+yes | crew install less
 echo
 echo "You will have to set the default PAGER environment variable to be able to use less:"
 echo "echo \"export PAGER=$CREW_PREFIX/bin/less\" >> ~/.bashrc && . ~/.bashrc"


uberhacker:
Hey @MatrixManAtYrService: How did you run the install.sh script?  The README.md install line pipes yes already.  I suppose it wouldn't hurt adding the pipes here but my original intention was to have some way to alternatively prompt.  This would remove that possibility.

MatrixManAtYrService:
Oh hey, so it does.

I was following this: https://skycocker.github.io/chromebrew/ which doesn't quite match the README.

uberhacker:
Yes, thank you for pointing that out.  If you want to update the README.md at https://skycocker.github.io/chromebrew/ with another PR, that would be awesome. :)

minektur:
If we don't want to bother asking if the user agrees then we should just remove that prompt code completely rather than suggest a way to ignore it.  We could add a command line option to 'agree' and skip the prompt, mentioning it in our docs...

uberhacker:
@minektur:  Thank you for the feedback.  We basically do that with the most recent install one-liner.  The issue here is the main home page needs updated to match this installer line.

PR #1470, opened 2017-12-01
Add tldr package.

skiman6010:
Tldr is a collection of simplified and community-driven man pages.

### DIFF
--- a/packages/tldr.rb
+++ b/packages/tldr.rb
@@ -1,0 +1,16 @@
+require 'package'
+  
+class Tldr < Package
+    description 'A collection of simplified and community-driven man pages.'
+    homepage 'http://tldr.sh/'
+    version '93cfb27'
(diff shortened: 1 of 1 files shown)


uberhacker:
Thank you for the new package @skiman6010!  You can simplify it a bit and only require one download if you do the following:
```
require 'package'
  
class Tldr < Package
  description 'A collection of simplified and community-driven man pages.'
  homepage 'http://tldr.sh/'
  version '93cfb27'
  source_url 'https://github.com/raylee/tldr/archive/93cfb278d2c56c3518a39ba3e01cbfc40423cc3b.tar.gz'
  source_sha256 'd565c6bdc594b87bf204fa35b3059011931a3cb59b935812c8a8a5fbe1fbd5e5'

  def self.install
    system "install -Dm755 tldr #{CREW_DEST_PREFIX}/bin/tldr"
  end
end
```

skiman6010:
Tested and the simpler way works. Thanks @uberhacker :)

Issue #1458, opened 2017-11-27
Checksum mismatch in libusbmuxd package


uberhacker:
```
libusbmuxd: USB Multiplex Daemon
http://www.libimobiledevice.org/
version 1.0.9
Downloading source...
Checksum mismatch. :/ Try again.
```

uberhacker:
Fixed.

uberhacker:
Fixed in #1471.

PR #1471, opened 2017-12-01
Update libusbmuxd.rb

skiman6010:
Fix for #1458. Correct sha256sum added.


uberhacker:
Thank you!  I was hoping someone would grab this easy issue.

PR #1474, opened 2017-12-02
Update tig package to compile correctly

uberhacker:
This package wasn't compiling on any architecture.  This update fixes it so it does.


cstrouse:

PR #1477, opened 2017-12-02
Remove node and node_latest packages

uberhacker:
These packages are no longer needed since we have the nvm and nodebrew packages.  Compiling node on Chrome OS is a painstaking process and since the nvm and nodebrew packages already have the ability to install binaries, there is no added value supporting them anymore.


cstrouse:
We need to patch nodebrew before we can remove these packages so that aarch64 users can still install node.js. The `system_info` subroutine needs to be modified so that aarch64 is treated as armv7l.

uberhacker:
I just merged @skiman6010's nodebrew PR which removes the corrupt binaries, so if you want to patch nodebrew so that it works on aarch64, please give it a go when you get a chance.

zacharysang:
I noticed that after running `crew remove` on `node`, `node_current`, nodebrew no longer worked correctly. Is anyone else seeing this (Maybe I just have a weird config going on)? If other people are seeing this though, then there may be a dependency here that would need to be accounted for

skiman6010:
@zacharysang did you run `nodebrew use <version>` to set an active node installation?

zacharysang:
Yeah, I did. Here is a screenshot showing my issue more clearly:

screenshot 2017-12-03 at 13 53 57

zacharysang:
@skiman6010  Additionally, after reinstalling node_current, the issue persists and I need to uninstall and reinstall nodebrew for it to start recognizing the node command again

uberhacker:
@zacharysang @skiman6010: The conflict issue should be resolved by this PR.  I imagine you might have a conflict with nodebrew and nvm as well since both packages basically do the same thing.

skiman6010:
@uberhacker  It seems to me like the conflict is from using the node package and Nodebrew/nvm. How should we go about correcting it?

uberhacker:
`crew remove nodebrew && crew install nodebrew` or `crew remove nvm && crew install nvm`

gangstead:
Agreed. One time I accidentally tried to build from source on my arm7 Chromebook flip. I have up after 45 minutes.

PR #1480, opened 2017-12-02
Add urlwatch package

uberhacker:
urlwatch: a tool for monitoring webpages for updates
urlwatch is intended to help you watch changes in webpages and get notified (via email, in your terminal or with a custom-written reporter class) of any changes. The change notification will include the URL that has changed and a unified diff of what has changed.  See https://thp.io/2008/urlwatch/.

### DIFF
--- a/packages/urlwatch.rb
+++ b/packages/urlwatch.rb
@@ -1,0 +1,46 @@
+require 'package'
+
+class Urlwatch < Package
+  description 'A tool for monitoring webpages for updates'
+  homepage 'https://thp.io/2008/urlwatch/'
+  version '2.7'
+  source_url 'https://github.com/thp/urlwatch/archive/2.7.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/urlwatch.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'python3'
+
+  def self.preinstall
+    abort "Unable to install urlwatch.  /home/#{USER}/user/.local directory exists.".lightred if Dir.exist? "/home/#{USER}/user/.local"
```
cstrouse: Composer and possibly other packages use subdirectories in this location. I don't think refusing to install if other packages have written data to ~/.local is the best way to do things.

uberhacker: If you have a better suggestion on how this should be installed, I'm all ears.  What I don't want is for this package to end up being installed in `~/.local`. :)

cstrouse:

PR #1481, opened 2017-12-03
Add template package

zacharysang:
Added a boilerplate package to serve as a starting point for adding new packages


uberhacker:
Hey @zacharysang:  Thank you for the PR!  Nice idea.  We will only need the one file, however.  It looks like you pushed up two by accident?  It might be good to add more to the template to help explain what functions are available.  I pulled the following snippet from `lib/package.rb`:
```
  # Function to perform pre-install operations prior to build from source.
  def self.preinstall

  end

  # Function to perform patch operations prior to build from source.
  def self.patch

  end

  def self.build

  end

  # Function to perform install from source build.
  def self.install

  end

  # Function to perform post-install for both source build and binary distribution
  def self.postinstall

  end

  # This is execute if and only if `crew build`.
  def self.check

  end
```
Also, it might be helpful to put a comment in the template as a suggestion to view some of the current packages to get an idea of what is possible.
Thanks

zacharysang:
Yeah, I was originally developing as template.rb and the rename didn't work as I wanted. Fixed now!

Additionally, I added boilerplate code for the the other methods inherited from Package and added a comment at the top as recommended. 

Let me know what you think!

uberhacker:
Looks good.  I'll let it percolate a bit and see what others have to say.

PR #1486, opened 2017-12-04
Add guile package

uberhacker:
Guile is a programming language

Guile is designed to help programmers create flexible applications that can be extended by users or other programmers with plug-ins, modules, or scripts.

With Guile you can create applications and games for the desktop, the Web, the command-line, and more.

See http://www.gnu.org/software/guile/.

### DIFF
--- a/packages/guile.rb
+++ b/packages/guile.rb
@@ -1,0 +1,33 @@
+require 'package'
+
+class Guile < Package
+  description 'Guile is designed to help programmers create flexible applications that can be extended by users or other programmers with plug-ins, modules, or scripts.'
+  homepage 'http://www.gnu.org/software/guile/'
+  version '2.2.3'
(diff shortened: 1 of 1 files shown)


packages/guile.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'libtool' => :build
+  depends_on 'bdwgc'
+  depends_on 'libffi'
+  depends_on 'libunistring'
```
cstrouse: Most lisps need readline or rlwrap for their REPL to be useful.

uberhacker: Those weren't explicitly required in the configure script.  Can you please give an example of something that doesn't work using one of the guile executables?

cstrouse:

PR #1495, opened 2017-12-06
Update less to include posix regex support

uberhacker:


cstrouse:

PR #1502, opened 2017-12-08
Add .github templates

flyingP0tat0:
This will make it easier to create and understand issues and pull requests.


uberhacker:
@flyingP0tat0: Thank you for these templates.  Great idea.  We can tweak them a bit later but I'll merge them now since something is better than nothing.

PR #1503, opened 2017-12-08
Correct install line

uberhacker:
The previous install line was not working.

Preview: http://htmlpreview.github.io/?https://github.com/uberhacker/chromebrew/blob/gh-pages-change-install-line/index.html

### DIFF
--- a/index.html
+++ b/index.html
@@ -86,7 +86,7 @@
             <h5>Available on aarch64, armv7l, i686 and x86_64.</h5>
           </div>
           <div class="highlight">
-            <pre><code>wget -q -O - https://raw.github.com/skycocker/chromebrew/master/install.sh | bash</code></pre>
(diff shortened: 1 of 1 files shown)


thedamian:
So the line that's currently doesn't work on my mac (your new one DOES work) 

But on my chromebook pro I ruan the line as it is on the skycocker.github.io/chromebook and it right away started with "Downloading ruby...." 

So I think this PR is NOT needed. Please let me know otherwise.

uberhacker:
I believe this works because you merged #1464.  We probably should close this PR and fix the install lines at https://github.com/skycocker/chromebrew, although now we don't have the option to prompt during installs anymore.

thedamian:
I don't mind closing this PR. That's fine (you can do that too no?)

BUT I'd like to know if you feel the problem is fixed and that the line works the way we have it on the https:// skycocker.github.io/chromebook  which is from where everyone usually installs from.

PR #1507, opened 2017-12-12
Remove corrupt binaries from lsb_release

uberhacker:
Reworked the install section to use the install command.


thedamian:
I like the install change. but WHY remove the binaries for nothing?

PR #1510, opened 2017-12-13
Update heroku.rb

skiman6010:
Update to v6.14.41. 

Need pre-compiled binaries for aarch64, armv7l and i686.

### DIFF
--- a/packages/syncthing.rb
+++ b/packages/syncthing.rb
@@ -1,0 +1,56 @@
+require 'package'
+
+class Syncthing < Package
+  description 'An application that lets you synchronize your files across multiple devices'
+  homepage 'https://github.com/syncthing/syncthing'
+  version '0.14.41'
+  source_url 'https://github.com/syncthing/syncthing/archive/v0.14.41.tar.gz'
+  source_sha256 '9c406897443465e03d3259df1dda10bd86e49ae4286f8c855d9f35e47026af2f'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'go' => :build
+
+  def self.build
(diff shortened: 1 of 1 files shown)


uberhacker:
Thank you for the update @skiman6010!  Can you please remove the pre-compiled binaries from your commit?  Also, if you could please change the install section like below, that would be better:
```
    system "mkdir -p #{CREW_DEST_PREFIX}/share/heroku"
    system "cp -r . #{CREW_DEST_PREFIX}/share/heroku"
    system "ln -s #{CREW_DEST_PREFIX}/share/heroku/bin/run #{CREW_DEST_PREFIX}/bin/heroku"
```
Thanks

skiman6010:
Should I remove the updated x86_64 binary I made and uploaded?

uberhacker:
Yes, the url can be simplified. :)  We should be adding binaries for all architectures on one commit, otherwise it creates confusion.  The versions of the binaries should always match the source in the package.

skiman6010:
I was hoping someone else could compile the binaries for the extra architectures before the pull request was merged, but we can just do a new pull request for those then. I made your changes.

uberhacker:
Can you please remove the indentation?  It makes the changes easier to detect.  Also, this might be a good time to learn how to squash your commits.  See https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git.  You will need to force the commit when you push it back to the repo using the `-f` flag.

skiman6010:
Wow that's cool. Didn't know you could squash commits like that. How's that looking?

uberhacker:
Yeah, squashing doesn't clutter up the commit log as much.  Looks like we have some blank lines with trailing whitespace and `Her  oku` in the description.  Can you please clean this up a bit? :)

PR #1511, opened 2017-12-13
Update Heroku.rb

skiman6010:
Update Heroku to 6.14.42. Change install to match other packages.


uberhacker:
There is a extra indented line but I can fix that when I build the pre-built binaries.  Nice work!

PR #1515, opened 2017-12-14
Remove corrupt pre-built binaries and simplify install of urlwatch

uberhacker:


uberhacker:
@thedamiam: we need the changes merged and the revision bumped before we create the binaries.

PR #1518, opened 2017-12-16
Update ocaml.rb

skiman6010:
Forgot to bump version number up. Whoops!


uberhacker:
Good catch.  My bad for missing it the first time.

PR #1519, opened 2017-12-16
Update expect.rb

skiman6010:


uberhacker:
Can we use https instead http for the source url?  I just noticed it works.

PR #1520, opened 2017-12-16
Update scons.rb

skiman6010:


uberhacker:
Can we use https instead http for the source url?  I just noticed it works.

skiman6010:
Yeah there we go! Sorry, I should have checked that.

uberhacker:
Sorry, I just noticed `--prefix=/usr/local`.  Can you please change that to `--prefix=#{CREW_PREFIX}` instead?

PR #1522, opened 2017-12-16
Add pre-built binaries for expect, miniupnpc, ocaml, ruby and scons

uberhacker:


thedamian:
Good god do I ever love this PR! YES YES YES!

PR #1524, opened 2017-12-17
Add libuuid package

uberhacker:
Portable uuid C library.  See https://sourceforge.net/projects/libuuid/.  Note that this package is a dependency of gptfdisk.


beliefless-dogmasless:
'libbuid' is 'absent' from the list:			https://github.com/skycocker/chromebrew/tree/master/packages

PR #1537, opened 2017-12-21
Update lynis.rb

skiman6010:


uberhacker:
Can you please change the following?
```
- system "echo './lynis \$1 \$2 \$3 \$4 \$5 \$6 \$7 \$8 \$9' >> lynis"
+ system "echo './lynis \"\$@\"' >> lynis"
```

PR #1580, opened 2017-12-28
Add fswatch package

uberhacker:
fswatch is a file change monitor that receives notifications when the contents of the specified files or directories are modified.  See https://github.com/emcrisostomo/fswatch.


uberhacker:
Tested and working on:
- [x] x86_64
- [x] i686
- [x] armv7l

PR #1587, opened 2017-12-29
Update unzip package to fix patch issue and compress man pages

uberhacker:


thedamian:
@uberhacker  help me understand this. I love the improvements. But I don't see a version change. Just a removal of the binaries. Some dependencies (which I'm sure are needed) and just improvements.

Why remove the binaries?

uberhacker:
@thedamian: The current binaries are not working.  Also, the new binaries are needed due to the patch and compressed man pages.

Issue #1585, opened 2017-12-29
Packages need updated


uberhacker:
Updates needed:
- [ ] abcde https://abcde.einval.com/download/abcde-2.9.3.tar.gz
- [x] acl https://bigsearcher.com/mirrors/nongnu/acl/acl-2.2.53.tar.gz
- [x] ag https://github.com/ggreer/the_silver_searcher/archive/2.2.0.tar.gz
- [ ] aide https://github.com/aide/aide/releases/download/v0.16.1/aide-0.16.1.tar.gz
- [ ] aircrack_ng https://download.aircrack-ng.org/aircrack-ng-1.5.2.tar.gz
- [ ] alsa_lib ftp://ftp.alsa-project.org/pub/lib/alsa-lib-1.1.9.tar.bz2
- [ ] alsa_plugins ftp://ftp.alsa-project.org/pub/plugins/alsa-plugins-1.1.9.tar.bz2
- [ ] ansible https://releases.ansible.com/ansible/ansible-2.7.10.tar.gz
- [ ] ansifilter http://www.andre-simon.de/zip/ansifilter-2.14.tar.bz2
- [x] ant https://www.apache.org/dist/ant/binaries/apache-ant-1.10.5-bin.tar.gz
- [ ] antlr4 https://github.com/antlr/antlr4/archive/4.7.2.tar.gz
- [x] apr https://apache.claz.org/apr/apr-1.7.0.tar.bz2
- [ ] aria2 https://github.com/aria2/aria2/releases/download/release-1.34.0/aria2-1.34.0.tar.xz
- [x] armadillo https://downloads.sourceforge.net/project/arma/armadillo-9.6.00.5.tar.xz
- [ ] asciinema https://github.com/asciinema/asciinema/archive/v2.0.2.tar.gz
- [x] aspell_en https://ftp.gnu.org/gnu/aspell/dict/en/aspell6-en-2018.04.16-0.tar.bz2
- [ ] at_spi2_atk https://github.com/GNOME/at-spi2-atk/archive/AT_SPI2_ATK_2_32_0.tar.gz
- [ ] at_spi2_core https://github.com/GNOME/at-spi2-core/archive/AT_SPI2_CORE_2_32_1.tar.gz
- [ ] atk https://download.gnome.org/sources/atk/2.32/atk-2.32.0.tar.xz
[...]

PR #1596, opened 2017-12-31
Updated libassuan to v 2.5.1, plus added support to put libraries in lib64 for x86_64

duckunix:
x86_64 hosts

Supports #1585 

## Description
Updated libassuan to v2.5.1.  Also, added support to put library files in .../lib64 for x86_64 hosts
## Additional information

Built & tested on:
- [x] x86_64
- [x]  armv7l
- [x] i686 (No hardware to test)
- [ ] aarch64 (No hardware to test)

 Also, removed pre-built stanzas

### DIFF
--- a/packages/libassuan.rb
+++ b/packages/libassuan.rb
@@ -8,12 +8,7 @@ class Libassuan < Package
   source_sha256 '47f96c37b4f2aac289f0bc1bacfa8bd8b4b209a488d3d15e2229cb6cc9b26449'
 
   def self.build
-    case ARCH
-      when 'x86_64'
-        system './configure --prefix=/usr/local --libdir=/usr/local/lib64'
-      else
-        system './configure --prefix=/usr/local'
-    end
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
     system 'make'
   end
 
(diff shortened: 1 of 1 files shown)


packages/libassuan.rb
```diff
@@ -3,25 +3,17 @@
...
+  source_sha256 '47f96c37b4f2aac289f0bc1bacfa8bd8b4b209a488d3d15e2229cb6cc9b26449'
 
   def self.build
-    system './configure --prefix=/usr/local'
+    case ARCH
+      when 'x86_64'
+        system './configure --prefix=/usr/local --libdir=/usr/local/lib64'
+      else
+        system './configure --prefix=/usr/local'
+    end
```
uberhacker: You can replace this entire conditional logic using crew constants with a single line:
```
```

uberhacker:
@duckunix: Thank you for the update!  Please see comment in the `Files changed` section.

packages/libassuan.rb
duckunix: Still learning ruby and chromebrew way.  That does make for cleaner code...I will update...

duckunix:
Updated per comment above.

uberhacker:
While testing on i686, I noticed there was a missing dependency:
```
checking for GPG Error - version >= 1.17... no
configure: error: libgpg-error was not found
libassuan failed to install: `./configure --prefix=/usr/local --libdir=/usr/local/lib` exited with 1
```
It appears we need `depends_on 'libgpgerror'` added.

duckunix:
Added depends_on 'libgpgerror', still builds for me. :)

uberhacker:
I didn't test on aarch64 but most of the time if it works on armv7l, it works on aarch64.

PR #1602, opened 2018-01-01
Supporting 1585 - update ascii, asciidoc, datamash, whois, avocado, aria2, wput, bdwgc

duckunix:
Updates #1585 

## Description

Programatically updated following packages:
- asciidoc - update to 8.6.9
- ascii - update to 3.18
- datamash - update to 1.2
- whois - update to 5.2.20
- avocado - update to 57.0
- aria2 - update to 1.33.1
- wput - update to 0.6.2
- bdwgc - update to 7.6.2

## Addtional information
At this point, these have been tested only on x86_64 that the packages build.

### DIFF
--- a/packages/aria2.rb
+++ b/packages/aria2.rb
@@ -4,8 +4,8 @@ class Aria2 < Package
   description 'aria2 is a lightweight multi-protocol & multi-source, cross platform download utility operated in command-line. It supports HTTP/HTTPS, FTP, SFTP, BitTorrent and Metalink.'
   homepage 'https://aria2.github.io/'
   version '1.33.1'
-  source_url 'https://github.com/aria2/aria2/releases/download/release-1.33.1/aria2-1.33.1.tar.bz2'
-  source_sha256 '32e8dc76da0c630da18d6a656afb117fef8f2f71d6f38e326a6e09fa113e2109'
+  source_url 'https://github.com/aria2/aria2/releases/download/release-1.33.1/aria2-1.33.1.tar.xz'
+  source_sha256 '2539e4844f55a1f1f5c46ad42744335266053a69162e964d9a2d80a362c75e1b'
 
   binary_url ({
   })
--- a/packages/aria2.rb
+++ b/packages/aria2.rb
@@ -3,9 +3,9 @@ require 'package'
 class Aria2 < Package
   description 'aria2 is a lightweight multi-protocol & multi-source, cross platform download utility operated in command-line. It supports HTTP/HTTPS, FTP, SFTP, BitTorrent and Metalink.'
   homepage 'https://aria2.github.io/'
-  version '1.33.0'
-  source_url 'https://github.com/aria2/aria2/releases/download/release-1.33.0/aria2-1.33.0.tar.xz'
-  source_sha256 '996e3fc2fd07ce2dd517e20a1f79b8b3dbaa5c7e27953b5fc19dae38f3874b8c'
+  version '1.33.1'
+  source_url 'https://github.com/aria2/aria2/releases/download/release-1.33.1/aria2-1.33.1.tar.bz2'
+  source_sha256 '32e8dc76da0c630da18d6a656afb117fef8f2f71d6f38e326a6e09fa113e2109'
 
   binary_url ({
   })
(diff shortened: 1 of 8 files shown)


uberhacker:
@duckunix: Thank you for the updates!  When you say programmatically updated, what do you mean?  I'm curious.  Please see comments in the `Files changed` section.

packages/ascii.rb
```diff
@@ -8,8 +8,16 @@ class Ascii < Package
...
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/ascii-3.16-1-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/ascii-3.16-1-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/ascii-3.16-1-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/ascii-3.16-1-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
+    aarch64: '8afd3dd8d8bbb56f039bd51dd2e0a65fdd23449b3858fb328fd0d2cbb75382b9',
+     armv7l: '8afd3dd8d8bbb56f039bd51dd2e0a65fdd23449b3858fb328fd0d2cbb75382b9',
+       i686: '9c5f93b333c1b40e558c76ab082bf5cfd7541a6f36f3c4561250e7dc2db45d26',
+     x86_64: 'cfce0c56efacea56a77d481d6f5001645972fa4780a2a8dea921f9435dffc466',
```
uberhacker: Don't we want this package updated to version 3.18?

packages/aria2.rb
```diff
@@ -3,9 +3,9 @@
 class Aria2 < Package
   description 'aria2 is a lightweight multi-protocol & multi-source, cross platform download utility operated in command-line. It supports HTTP/HTTPS, FTP, SFTP, BitTorrent and Metalink.'
   homepage 'https://aria2.github.io/'
-  version '1.33.0'
-  source_url 'https://github.com/aria2/aria2/releases/download/release-1.33.0/aria2-1.33.0.tar.xz'
-  source_sha256 '996e3fc2fd07ce2dd517e20a1f79b8b3dbaa5c7e27953b5fc19dae38f3874b8c'
+  version '1.33.1'
+  source_url 'https://github.com/aria2/aria2/releases/download/release-1.33.1/aria2-1.33.1.tar.bz2'
```
uberhacker: It might be better if we used https://github.com/aria2/aria2/releases/download/release-1.33.1/aria2-1.33.1.tar.xz instead since the compression is a bit more efficient.

packages/datamash.rb
```diff
@@ -3,21 +3,13 @@
 class Datamash < Package
   description 'GNU Datamash is a command-line program which performs basic numeric,textual and statistical operations on input textual data files.'
   homepage 'http://savannah.gnu.org/projects/datamash'
-  version '1.1.1'
-  source_url 'https://ftpmirror.gnu.org/datamash/datamash-1.1.1.tar.gz'
-  source_sha256 '420819b3d7372ee3ce704add847cff7d08c4f8176c1d48735d4a632410bb801b'
+  version '1.2'
+  source_url 'https://mirror.sergal.org/gnu/datamash/datamash-1.2.tar.gz'
```
uberhacker: Probably should stick with https://ftpmirror.gnu.org/datamash/datamash-1.2.tar.gz since it optimizes based on locality.

packages/aria2.rb
duckunix: Just pulled the URL from #1585.  I can update.

packages/datamash.rb
duckunix: Just pulled the URL from #1585. I can update.

packages/ascii.rb
duckunix: arg....should be...

duckunix:
So, when I say programmatically, what I mean is that I wrote a script which checks the .rb file, and if it is a simple one (no build dependencies, no case statements for different archs), then I use a little sed magic to clean the prebuilt links, update the version & URL, pull it, and try to build it.

Assuming it builds, I check it in to my local git repo, and then I will batch the up to create a PR.

duckunix:
Okay...URL fix, compress changes, and actually updated ascii.

PR #1634, opened 2018-01-10
Update heroku.rb

skiman6010:


uberhacker:
This version is probably good for only another few hours. LoL

PR #1664, opened 2018-01-16
Create libxshmfence.rb

lovedheart:
libxshmfence: A library that exposes a event API on top of Linux futexes

Works properly:
- [ ] x86_64 (not tested)
- [x] armv7l

---

### DIFF
--- a/packages/libxshmfence.rb
+++ b/packages/libxshmfence.rb
@@ -1,0 +1,20 @@
+require 'package'
+
+class Libxshmfence < Package
+  description 'libxshmfence: A library that exposes a event API on top of Linux futexes'
+  homepage 'http://t2sde.org/packages/libxshmfence.html'
+  version '1.32'
+  source_url 'http://xorg.freedesktop.org/releases/individual/lib/libxshmfence-1.2.tar.bz2'
+  source_sha256 'd21b2d1fd78c1efbe1f2c16dae1cb23f8fd231dcf891465b8debe636a9054b0c'
+
+  def self.build
+    system './configure',
+      "--prefix=#{CREW_PREFIX}",
+      "--libdir=#{CREW_LIB_PREFIX}"
+    system 'make'
+  end
+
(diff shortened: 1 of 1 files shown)


packages/libxshmfence.rb
```diff
@@ -0,0 +1,20 @@
+require 'package'
+
+class Libxshmfence < Package
+  description 'libxshmfence: A library that exposes a event API on top of Linux futexes'
+  homepage 'http://t2sde.org/packages/libxshmfence.html'
+  version '1.32'
+  source_url 'http://xorg.freedesktop.org/releases/individual/lib/libxshmfence-1.2.tar.bz2'
```
uberhacker: Can you please change the `http:` to `https:`?  This will make the url more secure.

packages/libxshmfence.rb
```diff
@@ -0,0 +1,20 @@
+require 'package'
+
+class Libxshmfence < Package
+  description 'libxshmfence: A library that exposes a event API on top of Linux futexes'
```
uberhacker: Can you please change this to `description 'A library that exposes a event API on top of Linux futexes'`?  The reason is `crew search libxshmfence` will return `libxshmfence: libxshmfence: A library that exposes a event API on top of Linux futexes`.  A bit redundant. :)

uberhacker:
@lovedheart: Thank you for the new package.  Please see the inline comments.

packages/libxshmfence.rb
```diff
@@ -0,0 +1,20 @@
+require 'package'
+
+class Libxshmfence < Package
+  description 'A library that exposes a event API on top of Linux futexes'
+  homepage 'http://t2sde.org/packages/libxshmfence.html'
+  version '1.32'
```
uberhacker: Should this be `version '1.2'`?

lovedheart:
yes I made an error. It should be v1.2

PR #1667, opened 2018-01-17
Add libomxil_bellagio

lovedheart:
Bellagio is an opensource implementation of the Khronos OpenMAX Integration Layer
API to access multimedia components.

## Addtional information

Works properly:
- [ ] x86_64 (no tested)
- [x] armv7l

### DIFF
--- a/packages/libomxil_bellagio
+++ b/packages/libomxil_bellagio
@@ -9,7 +9,7 @@ class Libomxil_bellagio < Package
 
   def self.build
     system "./configure","--prefix=#{CREW_PREFIX}","--libdir=#{CREW_LIB_PREFIX}"
-    system "make", "CFLAGS=-Wno-error=switch"    # only -j1 possible (tested on armv7l)
+    system "make", "-j1 CFLAGS=-Wno-error=switch"    # only -j1 possible (tested on armv7l)
   end
 
   def self.install
--- a/packages/libomxil_bellagio
+++ b/packages/libomxil_bellagio
@@ -1,0 +1,19 @@
+require 'package'
+
+class Libomxil_bellagio < Package
+  description 'Bellagio is an opensource implementation of the Khronos OpenMAX Integration Layer
+API to access multimedia components.'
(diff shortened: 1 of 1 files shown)


packages/libomxil_bellagio
```diff
@@ -0,0 +1,19 @@
+require 'package'
+
+class Libomxil_bellagio < Package
+  description 'Bellagio is an opensource implementation of the Khronos OpenMAX Integration Layer
+API to access multimedia components.'
```
uberhacker: Can you please make the description a single line?

packages/libomxil_bellagio
```diff
@@ -0,0 +1,19 @@
...
+  homepage 'https://sourceforge.net/projects/aa-project/'
+  version '0.9.3'
+  source_url 'https://sourceforge.net/projects/omxil/files/omxil/Bellagio%200.9.3/libomxil-bellagio-0.9.3.tar.gz'
+  source_sha256 '593c0729c8ef8c1467b3bfefcf355ec19a46dd92e31bfc280e17d96b0934d74c'
+
+  def self.build
+    system "./configure","--prefix=#{CREW_PREFIX}","--libdir=#{CREW_LIB_PREFIX}"
+    system "make", "CFLAGS=-Wno-error=switch"    # only -j1 possible (tested on armv7l)
```
uberhacker: Should we just go ahead and add the `, "-j1"` option here for all architectures?

lovedheart: I am not sure if that must be down on x86_64. But with -j1 is better for all. It does not take much time. (less than 1 min)

uberhacker: Do you need the `-j1` option for `armv7l` or does the current line work?

lovedheart: current line does not work.  Plz  put -j1 for armv7l

uberhacker: Can you please add it?  It's your PR. ;)

PR #1668, opened 2018-01-17
Added a set of formulae for llvm5 and family.

indianajohn:
## Description
GCC 4.9.1 lets us build a lot of packages, but it's not only on the old side (not supporting some modern compiler features), it also lacks some development features - libclang is pretty nice and enables things like YouCompleteMe (https://valloric.github.io/YouCompleteMe/). Finally, some people just plain prefer it.

This package contains the entire LLVM/Clang 5.0.1 toolchain, with the exception of `lldb`, `llvm-test-suite`, `clang-tools-extra`, and `libc++-abi`. Those packages could be added in the future if they were needed by someone.

## Addtional information
* The actual build of these packages takes forever on many Chromebooks. On my ARM Asus Chromebook Flip, compiling the entire set of packages would probably take 5-6 hours. One of my first priorities is to upload some binary packages. After I build these, do you how would I go about getting them on your BIntray?

* For speed of build and space considerations (my Chromebook was actually running out of disk space during the build), I've limited the targets to x86 and ARM, since this means that clang can target all (?) Chromebooks.
* The install adds the shortcuts `clang++w` and `clangw` (w is for "wrapper"), which automatically inserts command line arguments for the standard include and library paths and passes the remaining arguments for whatever variant of Clang is being used.

* This supports the C++17 standard.

* I've successfully built simple programs with this. More complicate projects such as ones built with CMake will require debugging as we proceed.

* the build of `lldb` version 5 will require stashing the LLVM source somewhere and then providing it for the build. I'll probably just have `llvm5` copy the source code to the `share` directory for dependent package(s) to reference. This pull request was getting big enough as it was. 

* This puts the old `lldb` package in a weird place. You might just want to get rid of it and drop the "5" suffix here. Let me know if that would be best.

* I have Spacemacs up and running with full semantic completion in C++ through YouCompleteMe. That might be a package for the future, as it makes my Chromebook viable for native C++ development.

Works properly:
- [? ] x86_64
- [* ] aarch64 (reasons why it doesn't)

### DIFF
--- a/packages/lld.rb
+++ b/packages/lld.rb
@@ -1,0 +1,25 @@
+require 'package'
+
+class Lld < Package
+  description 'LLD is a linker from the LLVM project.'
+  homepage 'https://lld.llvm.org/'
+  version '5.0.1'
+  source_url 'https://releases.llvm.org/5.0.1/lld-5.0.1.src.tar.xz'
+  source_sha256 'd5b36c0005824f07ab093616bdff247f3da817cae2c51371e1d1473af717d895'
+
+  depends_on 'cmake' => :build
+
+  def self.build
+    Dir.mkdir 'mybuilddir'
+    Dir.chdir "mybuilddir" do
+      system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
+      system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+      system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_PREFIX} -P cmake_install.cmake"
+    end
+  end
+end
--- a/packages/llvm.rb
+++ b/packages/llvm.rb
@@ -2,38 +2,24 @@ require 'package'
 
 class Llvm < Package
   description 'The LLVM Project is a collection of modular and reusable compiler and toolchain technologies.'
-  homepage 'http://llvm.org/'
-  version '3.8.1-1'
-  source_url 'http://llvm.org/releases/3.8.1/llvm-3.8.1.src.tar.xz'
-  source_sha256 '6e82ce4adb54ff3afc18053d6981b6aed1406751b8742582ed50f04b5ab475f9'
+  homepage 'https://llvm.org/'
+  version '5.0.1'
+  source_url 'https://releases.llvm.org/5.0.1/llvm-5.0.1.src.tar.xz'
+  source_sha256 '5fa7489fc0225b11821cab0362f5813a05f2bcf2533e8a4ea9c9c860168807b0'
 
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/llvm-3.8.1-1-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/llvm-3.8.1-1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/llvm-3.8.1-1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/llvm-3.8.1-1-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: '94255baa409229bbd0ad5f23981ee807fc39cb0c5160451daa0c3bfb1ebb78ae',
-     armv7l: '94255baa409229bbd0ad5f23981ee807fc39cb0c5160451daa0c3bfb1ebb78ae',
-       i686: '1d906912c090feeb81bb1777b8ff1ec7dcdadea08adea68c8a9cf9dfdcd8f2bb',
-     x86_64: 'efd72e40446eb7fb266b39a2e20df8398846ed76c96197b65dd31cb0fe023438',
-  })
-
-  depends_on 'buildessential'
-  depends_on 'cmake'
+  depends_on 'cmake' => :build
 
   def self.build
-    system "mkdir mybuilddir"
+    Dir.mkdir 'mybuilddir'
     Dir.chdir "mybuilddir" do
- 	    system "cmake .. -DLLVM_BUILD_LLVM_DYLIB=true"
-    	system "cmake --build ."
+      system "cmake .. -DBUILD_SHARED_LIBS=true -DCMAKE_BUILD_TYPE=Release -DLLVM_OPTIMIZED_TABLEGEN=ON"
+      system "cmake --build . --  -j#{CREW_NPROC}"
     end
   end
 
   def self.install
     Dir.chdir "mybuilddir" do
-       system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_DIR}/usr/local -P cmake_install.cmake"
+      system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_PREFIX} -P cmake_install.cmake"
     end
   end
 end
--- a/packages/clang.rb
+++ b/packages/clang.rb
@@ -1,0 +1,45 @@
+require 'package'
+
+class Clang < Package
+  description 'C language family frontend for LLVM.'
+  homepage 'https://clang.llvm.org/'
+  version '5.0.1'
+  source_url 'https://releases.llvm.org/5.0.1/cfe-5.0.1.src.tar.xz'
+  source_sha256 '135f6c9b0cd2da1aff2250e065946258eb699777888df39ca5a5b4fe5e23d0ff'
+
+  depends_on 'cmake' => :build
+  depends_on 'llvm'
+  depends_on 'clang_libcpp'
+  depends_on 'clang_openmp'
+  depends_on 'llvm_compiler_rt'
+  depends_on 'llvm_polly'
+  depends_on 'llvm_unwind'
+  depends_on 'lld'
+
+  def self.build
+    Dir.mkdir 'mybuilddir'
+    Dir.chdir "mybuilddir" do
+      system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
+      system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+      system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_PREFIX} -P cmake_install.cmake"
+      FileUtils.cd("#{CREW_DEST_PREFIX}/bin") do
+        system "echo '#!/bin/bash' > clangppw"
+        system "echo 'GCCVERSION=`gcc --version | grep ^gcc | sed '\\''s/^.* //g'\\''`' >> clangppw"
+        system "echo 'TARGET=`gcc -dumpmachine`' >> clangppw"
+        system "echo '#{CREW_PREFIX}/bin/clang++ -I#{CREW_PREFIX}/include/c++/v1 -B#{CREW_PREFIX}/lib/gcc/${TARGET}/${GCCVERSION} -lc++ $@' >> clangppw"
+        system "chmod 755 clangppw"
+        system "echo '#!/bin/bash' > clangw"
+        system "echo 'GCCVERSION=`gcc --version | grep ^gcc | sed '\\''s/^.* //g'\\''`' >> clangw"
+        system "echo 'TARGET=`gcc -dumpmachine`' >> clangw"
+        system "echo '#{CREW_PREFIX}/bin/clang -B#{CREW_PREFIX}/lib/gcc/${TARGET}/${GCCVERSION} $@' >> clangw"
+        system "chmod 755 clangw"
+      end
+    end
+  end
+end
+
--- a/packages/clang_libcpp.rb
+++ b/packages/clang_libcpp.rb
@@ -1,0 +1,26 @@
+require 'package'
+
+class Clang_libcpp < Package
+  description 'Standard library for Clang5.'
+  homepage 'https://libcxx.llvm.org/'
+  version '5.0.1'
+  source_url 'https://releases.llvm.org/5.0.1/libcxx-5.0.1.src.tar.xz'
+  source_sha256 'fa8f99dd2bde109daa3276d529851a3bce5718d46ce1c5d0806f46caa3e57c00'
+
+  depends_on 'cmake' => :build
+  depends_on 'llvm'
+
+  def self.build
+    Dir.mkdir 'mybuilddir'
+    Dir.chdir "mybuilddir" do
+      system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
+      system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+      system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_PREFIX} -P cmake_install.cmake"
+    end
+  end
+end
--- a/packages/clang_openmp.rb
+++ b/packages/clang_openmp.rb
@@ -1,0 +1,25 @@
+require 'package'
+
+class Clang_openmp < Package
+  description 'The OpenMP subproject of LLVM contains the components required to build an executable OpenMP program that are outside the compiler itself.'
+  homepage 'https://openmp.llvm.org/'
+  version '5.0.1'
+  source_url 'https://releases.llvm.org/5.0.1/openmp-5.0.1.src.tar.xz'
+  source_sha256 'adb635cdd2f9f828351b1e13d892480c657fb12500e69c70e007bddf0fca2653'
+
+  depends_on 'cmake' => :build
+
+  def self.build
+    Dir.mkdir 'mybuilddir'
+    Dir.chdir "mybuilddir" do
+      system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
+      system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+      system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_PREFIX} -P cmake_install.cmake"
+    end
+  end
+end
--- a/packages/llvm_compiler_rt.rb
+++ b/packages/llvm_compiler_rt.rb
@@ -1,0 +1,25 @@
+require 'package'
+
+class Llvm_compiler_rt < Package
+  description 'Part of the LLVM project.'
+  homepage 'https://compiler-rt.llvm.org/'
+  version '5.0.1'
+  source_url 'https://releases.llvm.org/5.0.1/compiler-rt-5.0.1.src.tar.xz'
+  source_sha256 '4edd1417f457a9b3f0eb88082530490edf3cf6a7335cdce8ecbc5d3e16a895da'
+
+  depends_on 'cmake' => :build
+
+  def self.build
+    Dir.mkdir 'mybuilddir'
+    Dir.chdir "mybuilddir" do
+      system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
+      system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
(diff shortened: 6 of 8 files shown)


uberhacker:
@indianajohn: Wow, this is an awesome package suite.  Thank you for your effort in putting this together.  Currently, there are no other packages that have `lldb` as a dependency, so just replacing what we have there might be the best option.  I will include some inline comments also but some things that stand out are below:
- Please change the `source_url`s to use https instead of http
- Please replace all instances of `/usr/local` with `#{CREW_PREFIX}` and `#{CREW_DEST_DIR}/usr/local` with `#{CREW_DEST_PREFIX}`
- You could use `Dir.mkdir 'mybuilddir'` instead of `system "mkdir -p mybuilddir"`
- You don't need the `depends_on 'buildessential'` lines since that package is part of crew "core"
- Please change `depends_on 'cmake'` to `depends_on 'cmake' => :build` to only require that dependency when compiling the package

packages/clang5.rb
```diff
@@ -0,0 +1,46 @@
...
+  depends_on 'clang5_openmp'
+  depends_on 'llvm5_compiler_rt'
+  depends_on 'llvm5_polly'
+  depends_on 'llvm5_unwind'
+  depends_on 'lld5'
+
+  def self.build
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
```
uberhacker: Please indent only 2 spaces below Dir.chdir ...

packages/clang5.rb
```diff
@@ -0,0 +1,46 @@
...
+        system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+      system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_DIR}/usr/local -P cmake_install.cmake"
```
uberhacker: Missing `.` or `..` target directory?

packages/clang5.rb
```diff
@@ -0,0 +1,46 @@
...
+      FileUtils.cd("#{CREW_DEST_PREFIX}/bin") do
+        system "echo '#!/bin/bash' > clang++w"
```
uberhacker: Probably shouldn't use special characters in filenames.  Can you use `clangppw` instead?

packages/clang5.rb
```diff
@@ -0,0 +1,46 @@
...
+        system "echo 'GCCVERSION=`gcc --version | grep ^gcc | sed '\\''s/^.* //g'\\''`' >> clang++w"
+        system "echo 'TARGET=`gcc -dumpmachine`' >> clang++w"
+        system "echo '/usr/local/bin/clang++ -I/usr/local/include/c++/v1 -B/usr/local/lib/gcc/${TARGET}/${GCCVERSION} -lc++ $@' >> clang++w"
```
uberhacker: You will want to escape the `$`s with `\$`, otherwise it gets evaluated inside the double quoted string.

packages/clang5.rb
```diff
@@ -0,0 +1,46 @@
...
+        system "chmod 755 clang++w"
+        system "echo '#!/bin/bash' > clangw"
+        system "echo 'GCCVERSION=`gcc --version | grep ^gcc | sed '\\''s/^.* //g'\\''`' >> clangw"
+        system "echo 'TARGET=`gcc -dumpmachine`' >> clangw"
+        system "echo '/usr/local/bin/clang -B/usr/local/lib/gcc/${TARGET}/${GCCVERSION} $@' >> clangw"
```
uberhacker: Again, you will want to escape both `$` and `@`.

packages/clang5_openmp.rb
```diff
@@ -0,0 +1,26 @@
...
+  source_url 'http://releases.llvm.org/5.0.1/openmp-5.0.1.src.tar.xz'
+  source_sha256 'adb635cdd2f9f828351b1e13d892480c657fb12500e69c70e007bddf0fca2653'
+
+  depends_on 'buildessential'
+  depends_on 'cmake'
+
+  def self.build
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
```
uberhacker: Please indent with only 4 spaces below Dir.chdir ...

packages/clang5_openmp.rb
```diff
@@ -0,0 +1,26 @@
...
+        system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+       system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_DIR}/usr/local -P cmake_install.cmake"
```
uberhacker: Missing `.` or `..` target directory?

packages/llvm5_compiler_rt.rb
```diff
@@ -0,0 +1,26 @@
...
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
+        system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+       system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_DIR}/usr/local -P cmake_install.cmake"
```
uberhacker: Missing `.` or `..` target directory?

packages/llvm5.rb
```diff
@@ -0,0 +1,26 @@
...
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake .. -DBUILD_SHARED_LIBS=true -DLLVM_TARGETS_TO_BUILD=\"X86\;ARM\" -DCMAKE_BUILD_TYPE=Release -DLLVM_OPTIMIZED_TABLEGEN=ON"
+    	system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+       system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_DIR}/usr/local -P cmake_install.cmake"
```
uberhacker: Missing `.` or `..` target directory?

packages/lld5.rb
```diff
@@ -0,0 +1,27 @@
...
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
+        system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+       system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_DIR}/usr/local -P cmake_install.cmake"
```
uberhacker: Missing `.` or `..` target directory?

packages/llvm5_polly.rb
```diff
@@ -0,0 +1,27 @@
...
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
+        system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+       system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_DIR}/usr/local -P cmake_install.cmake"
```
uberhacker: Missing `.` or `..` target directory?

packages/llvm5_unwind.rb
```diff
@@ -0,0 +1,27 @@
...
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
+        system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+       system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_DIR}/usr/local -P cmake_install.cmake"
```
uberhacker: Missing `.` or `..` target directory?

packages/clang5_libcpp.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  depends_on 'buildessential'
+  depends_on 'cmake'
+  depends_on 'llvm5'
+  depends_on 'clang5'
+
+  def self.build
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
```
uberhacker: Please indent with only 4 spaces below Dir.chdir

packages/lld5.rb
```diff
@@ -0,0 +1,27 @@
...
+  source_sha256 'd5b36c0005824f07ab093616bdff247f3da817cae2c51371e1d1473af717d895'
+
+  depends_on 'buildessential'
+  depends_on 'cmake'
+  depends_on 'llvm5'
+
+  def self.build
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
```
uberhacker: Please indent with only 4 spaces below Dir.chdir

packages/llvm5.rb
```diff
@@ -0,0 +1,26 @@
...
+  source_url 'http://releases.llvm.org/5.0.1/llvm-5.0.1.src.tar.xz'
+  source_sha256 '5fa7489fc0225b11821cab0362f5813a05f2bcf2533e8a4ea9c9c860168807b0'
+
+  depends_on 'buildessential'
+  depends_on 'cmake'
+
+  def self.build
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake .. -DBUILD_SHARED_LIBS=true -DLLVM_TARGETS_TO_BUILD=\"X86\;ARM\" -DCMAKE_BUILD_TYPE=Release -DLLVM_OPTIMIZED_TABLEGEN=ON"
```
uberhacker: Please indent with only 4 spaces below Dir.chdir

packages/llvm5_compiler_rt.rb
```diff
@@ -0,0 +1,26 @@
...
+  source_url 'http://releases.llvm.org/5.0.1/compiler-rt-5.0.1.src.tar.xz'
+  source_sha256 '4edd1417f457a9b3f0eb88082530490edf3cf6a7335cdce8ecbc5d3e16a895da'
+
+  depends_on 'buildessential'
+  depends_on 'cmake'
+
+  def self.build
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
```
uberhacker: Please indent with only 4 spaces below Dir.chdir

packages/llvm5_polly.rb
```diff
@@ -0,0 +1,27 @@
...
+  source_sha256 '9dd52b17c07054aa8998fc6667d41ae921430ef63fa20ae130037136fdacf36e'
+
+  depends_on 'buildessential'
+  depends_on 'cmake'
+  depends_on 'llvm5'
+
+  def self.build
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
```
uberhacker: Please indent with only 4 spaces below Dir.chdir

packages/llvm5_unwind.rb
```diff
@@ -0,0 +1,27 @@
...
+  source_sha256 '6bbfbf6679435b858bd74bdf080386d084a76dfbf233fb6e47b2c28e0872d0fe'
+
+  depends_on 'buildessential'
+  depends_on 'cmake'
+  depends_on 'llvm5'
+
+  def self.build
+    system "mkdir -p mybuilddir"
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
```
uberhacker: Please indent with only 4 spaces below Dir.chdir

packages/clang5_libcpp.rb
```diff
@@ -0,0 +1,28 @@
...
+        system "cmake --build . --  -j#{CREW_NPROC}"
+    end
+  end
+
+  def self.install
+    Dir.chdir "mybuilddir" do
+       system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_DEST_DIR}/usr/local -P cmake_install.cmake"
```
uberhacker: Missing `.` or `..` target directory?

packages/llvm5_polly.rb
indianajohn: In these cases, the install command works fine - there's no need to specify the directory if the -P option is specified I think (which gives the path to the install script.

indianajohn: It's equivalent to running `make install` in the build directory. I'm not sure why the original creator of the llvm package did it that way, but I just followed his/her lead.

uberhacker: OK great.  Just wanted to make sure.  Nice to know.

packages/clang5.rb
uberhacker: Actually, you probably will want to quote the `$@` also.  Sometimes arguments have spaces.  So the final string should look like: `\"\$\@\"`.  So much escaping, I know.

packages/clang5.rb
indianajohn: I'll test this again, but I think that anything enclosed inside the single `s will not be processed by the shell. The single quote is ruby, right?

I had to fight with it before I got it to work, of course.

packages/clang5.rb
indianajohn: See above. The file seems to end up being created properly in my tests.

packages/clang5.rb
uberhacker: From my experience, anything inside a double quoted string is evaluated.  Even if there is a single quoted string inside the double quoted string, everything will still get evaluated.  Does the file get created with the literal `$` character still visible?  I would be surprised if it did.

indianajohn: I don't pretend to know how shell escaping inside a subshell inside a Ruby string works, but oddly enough this is the exact file that got created:
```
#!/bin/bash
GCCVERSION=`gcc --version | grep ^gcc | sed 's/^.* //g'`
TARGET=`gcc -dumpmachine`
/usr/local/bin/clang++ -I/usr/local/include/c++/v1 -B/usr/local/lib/gcc/${TARGET}/${GCCVERSION} -lc++ $@
```

indianajohn:
Tonight I'll run a package build on all of these and hopefully produce some binary packages. That will also serve to test the most recent changes and make sure I didn't muck anything up. Let me know if you notice anything else that should be changed.

packages/llvm.rb
```diff
@@ -2,38 +2,24 @@
...
-  depends_on 'cmake'
+  depends_on 'cmake' => :build
 
   def self.build
-    system "mkdir mybuilddir"
+    Dir.mkdir 'mybuilddir'
     Dir.chdir "mybuilddir" do
- 	    system "cmake .. -DLLVM_BUILD_LLVM_DYLIB=true"
-    	system "cmake --build ."
+ 	    system "cmake .. -DBUILD_SHARED_LIBS=true -DLLVM_TARGETS_TO_BUILD=\"X86\;ARM\" -DCMAKE_BUILD_TYPE=Release -DLLVM_OPTIMIZED_TABLEGEN=ON"
```
uberhacker: I'm thinking we should leave out the `-DLLVM_TARGETS_TO_BUILD=\"X86\;ARM\"` option so this can be built on `x86_64` architecture also (the most popular Chromebook platform).

uberhacker: Please change your indentation also.  If you are using tabs, please use spaces instead.  I know this is a bit nit picky but we strive for clean code.

indianajohn: I'll change the indentation in a bit. The indentation is weird because I used just plain old vi to do it, because I broke my VIm/Spacemacs configuration by removing libclang over and over again. :/

The thing about the targets... I thought the same thing you did, but then I looked at CMakeLists.txt in Clang and noted this:

```

**Comment by indianajohn** [comment]
Note the lack of an arch for x86_64 in the list. I haven't found any docs with state this explicitly, but I think the X86 target supports the x86_64 architecture.

File: packages/llvm.rb (hunk 20 above)
Comment (indianajohn) [reply packages/llvm.rb]: Note the lack of an arch for x86_64 in the list. I haven't found any docs with state this explicitly, but I think the X86 target supports the x86_64 architecture.

I'll dig around more to confirm that. Anyway, the reason why I did it in the first place is that building all arches filled up my hard drive. I later discovered that the default mode for the build was set to debug, which made everything bigger, so it's possible we can turn all architectures back on without removing my computer's ability to build it.

It will take a lot longer to build though.

Comment (indianajohn) [reply packages/llvm.rb]: Looks like aarch64 Chromebooks exist too. I'll just try to turn on all architectures and building again.

File: crew [hunk 21]
```diff
@@ -697,6 +697,7 @@ def build_package (pwd)
 
   # build package from filelist, dlist and binary files in CREW_DEST_DIR
   puts "Archiving..."
+  install_package pwd
```
uberhacker: I don't think we need this, do we?  Currently, crew works as expected without doing this.

packages/llvm.rb
```diff
@@ -12,14 +12,14 @@ class Llvm < Package
   def self.build
     Dir.mkdir 'mybuilddir'
     Dir.chdir "mybuilddir" do
- 	    system "cmake .. -DBUILD_SHARED_LIBS=true -DLLVM_TARGETS_TO_BUILD=\"X86\;ARM\" -DCMAKE_BUILD_TYPE=Release -DLLVM_OPTIMIZED_TABLEGEN=ON"
-    	system "cmake --build . --  -j#{CREW_NPROC}"
+      system "cmake .. -DBUILD_SHARED_LIBS=true -DCMAKE_BUILD_TYPE=Release -DLLVM_OPTIMIZED_TABLEGEN=ON"
+      system "cmake --build . --  -j#{CREW_NPROC}"
```
uberhacker: Thank you for fixing the indentation.  It looks like the other packages in this suite have a similar issue with the indentation also.

crew
indianajohn: That wasn't supposed to make it into this PR. It would be nice to have a 'crew build --install <package_name' (spoiler: that doesn't work.)

indianajohn:
I think I caught all of the indentation problems. I'm going to finish up my local build here and in the process remove some unnecessary local dependencies (most of this stuff only depends on llvm). I'll let you know when that test has gone successfully, and also upload packages to Google Drive.

uberhacker:
Hey @indianajohn: No need to rebase but you might want to squash your commits.  Have you tried that in git yet?

packages/lld.rb
```diff
@@ -0,0 +1,26 @@
...
+  source_sha256 'd5b36c0005824f07ab093616bdff247f3da817cae2c51371e1d1473af717d895'
+
+  depends_on 'cmake' => :build
+  depends_on 'llvm'
+
+  def self.build
+    Dir.mkdir 'mybuilddir'
+    Dir.chdir "mybuilddir" do
+ 	    system "cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Release .."
+        system "cmake --build . --  -j#{CREW_NPROC}"
```
uberhacker: Indentation fix still needed?

indianajohn:
I was able to build everything successfully, and upload the packages I created:

https://drive.google.com/drive/folders/1yB9kROE80UEVsJK_ksrw-atEkUVD1Lhe?usp=sharing

We can put those in Bintray and add links if you'd like, or wait until we can build a package for all arches in a separate PR. I'm not sure how your package build infrastructure works so I'll leave that abstract for now.

I think that pretty much wraps this up?

uberhacker:
Hey @indianajohn: Don't worry about the binaries.  I can take care of that.  Thank you for your hard work on this.

PR #1670, opened 2018-01-17
Update mesa.rb

lovedheart:
## Description
modification for mesa build on armv7l

## Addtional information

Works properly:
- [x] x86_64 (binary available)
- [x] armv7l (tested)

### DIFF
--- a/packages/mesa.rb
+++ b/packages/mesa.rb
@@ -26,13 +26,23 @@ class Mesa < Package
   depends_on 'libxtrans'
   depends_on 'libxkbfile'
   depends_on 'wayland'
-  depends_on 'python'
+  depends_on 'python27'
   depends_on 'bison'
 
   # tested on armv7l
   def self.build
     system "pip install Mako"
-    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --enable-shared-glapi --with-gallium-drivers=nouveau --with-dri-drivers=nouveau --enable-egl --with-platforms=x11,drm,wayland --enable-gbm --enable-xa --enable-dri"
+    system "./configure \
+            --prefix=#{CREW_PREFIX} \
+            --libdir=#{CREW_LIB_PREFIX} \
+            --enable-shared-glapi \
+            --with-gallium-drivers=nouveau \
+            --with-dri-drivers=nouveau \
+            --enable-egl \
+            --with-platforms=x11,drm,wayland \
+            --enable-gbm \
+            --enable-xa \
+            --enable-dri"    
     system "make"
   end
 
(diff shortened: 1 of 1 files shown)


packages/mesa.rb
```diff
@@ -27,8 +27,12 @@ class Mesa < Package
   depends_on 'libxkbfile'
   depends_on 'wayland'
 
+  depends_on 'bison'
```
uberhacker: You will need `depends_on 'python27'` since you are using `pip install` below.

lovedheart: yes

packages/mesa.rb
```diff
@@ -26,9 +26,13 @@ class Mesa < Package
   depends_on 'libxtrans'
   depends_on 'libxkbfile'
   depends_on 'wayland'
+  depends_on 'python'
```
uberhacker: There is no package named `python`.  Based on your `pip install` line below, you should add `python27` as a dependency.

packages/mesa.rb
```diff
@@ -26,9 +26,13 @@ class Mesa < Package
...
+  depends_on 'bison'
 
+  # tested on armv7l
   def self.build
-    system "./configure --enable-gles2 --with-egl-platforms=x11,wayland,drm --enable-gbm --enable-shared-glapi --with-gallium-drivers=swrast --disable-dri3"
+    system "pip install Mako"
```
uberhacker: Please add `--prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR}` options to the pip install line so the package can track this dependency and uninstall when the package is removed.

packages/mesa.rb
```diff
@@ -26,9 +26,13 @@ class Mesa < Package
...
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --enable-shared-glapi --with-gallium-drivers=nouveau --with-dri-drivers=nouveau --enable-egl --with-platforms=x11,drm,wayland --enable-gbm --enable-xa --enable-dri"
```
uberhacker: It would be good for readability if you put each option on a separate line.  Something like:
```
system "./configure \
  --prefix=#{CREW_PREFIX} \
  --libdir=#{CREW_LIB_PREFIX} \
  --enable-shared-glapi \
  --with-gallium-drivers=nouveau \
  --with-dri-drivers=nouveau \
  --enable-egl \
  --with-platforms=x11,drm,wayland \
  --enable-gbm \
  --enable-xa \
  --enable-dri"
```
This also makes it easier to add or remove options simply by adding/removing the option line.

uberhacker:
I'll go ahead and merge this. I noticed the configure options were changed a bit but that is fine.  I'm sure we will need to tweak this a few more times before it's all said and done.

uberhacker:
@lovedheart: I noticed there was an update for this package.  Can you please update to ftp://ftp.freedesktop.org/pub/mesa/mesa-17.3.2.tar.xz when you get a chance?

PR #1671, opened 2018-01-17
Update cairo.rb

lovedheart:
add dependency


uberhacker:
@lovedheart: This dependency is not needed.  I just removed automake on my armv7l system and it compiled fine.  Also, there are binaries available with PR #1665 so you won't need to compile after it is merged.

PR #1674, opened 2018-01-18
Rename libomxil_bellagio to libomxil_bellagio.rb

lovedheart:
Fix package name


uberhacker:
Oh, wow.  I totally missed this.  Good catch!

PR #1676, opened 2018-01-18
Update mesa to 17.3.2 (tested on armv7l)

lovedheart:
Tested on armv7l

### DIFF
--- a/packages/mesa.rb
+++ b/packages/mesa.rb
@@ -4,7 +4,7 @@ class Mesa < Package
   description 'Open-source implementation of the OpenGL specification'
   homepage 'https://www.mesa3d.org'
   version '17.3.2'
-  source_url 'ftp://ftp.freedesktop.org/pub/mesa/mesa-17.3.2.tar.xz'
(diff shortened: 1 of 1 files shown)


packages/mesa.rb
```diff
@@ -3,17 +3,15 @@
 class Mesa < Package
   description 'Open-source implementation of the OpenGL specification'
   homepage 'https://www.mesa3d.org'
-  version '17.1.5'
-  source_url 'ftp://ftp.freedesktop.org/pub/mesa/mesa-17.1.5.tar.xz'
-  source_sha256 '378516b171712687aace4c7ea8b37c85895231d7a6d61e1e27362cf6034fded9'
+  version '17.3.2'
+  source_url 'ftp://ftp.freedesktop.org/pub/mesa/mesa-17.3.2.tar.xz'
```
uberhacker: Sorry, you must have missed my correction.  Can we use https://mesa.freedesktop.org/archive/mesa-17.3.2.tar.xz instead since this url is secure?

PR #1677, opened 2018-01-20
Allow -j# overrides in lib/package.rb

### LINKED ISSUE #1666: How to compile code with -j1 using crew?
I tried to compile libomxil-bellagio. It has been shown that only "make -j1" works. How to compile code with single job. I am not much familiar with ruby.

uberhacker:
Fixes #1666.  Is this the fix you were looking for @lovedheart?

### DIFF
--- a/lib/package.rb
+++ b/lib/package.rb
@@ -101,21 +101,42 @@ class Package
   end
 
   def self.system(*args)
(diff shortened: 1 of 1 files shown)


lovedheart:
This works when I remove "-j1" in libomxil_bellagio.rb

PR #1680, opened 2018-01-22
Update libomxil_bellagio.rb

lovedheart:
remove compile option flag -j1 to make it work together with #1677 

Works properly:
- [] x86_64
- [x] armv7l


uberhacker:
@lovedheart: You probably want to do something like this:
```
system "make", "-j1", "CFLAGS=-Wno-error=switch"
```
The argument parsing is based on what crew considers arguments.  This would likely work also:
```
system "make -j1 CFLAGS=-Wno-error=switch"
```
Hope that makes sense.

thedamian:
@uberhacker I like your latest suggestions, but I'm merging it and someone can clean it up when they have a minute.

PR #1683, opened 2018-01-23
Add mc package

uberhacker:
GNU Midnight Commander is a visual file manager, licensed under GNU General Public License and therefore qualifies as Free Software. It's a feature rich full-screen text mode application that allows you to copy, move and delete files and whole directory trees, search for files and run commands in the subshell. Internal viewer and editor are included.

Midnight Commander is based on versatile text interfaces, such as Ncurses or S-Lang, which allows it to work on a regular console, inside an X Window terminal, over SSH connections and all kinds of remote shells.

See http://midnight-commander.org/.

### DIFF
--- a/packages/mc.rb
+++ b/packages/mc.rb
@@ -1,0 +1,30 @@
+require 'package'
+
+class Mc < Package
+  description 'GNU Midnight Commander is a visual file manager'
+  homepage 'http://midnight-commander.org/'
+  version '4.8.20'
+  source_url 'https://github.com/MidnightCommander/mc/archive/4.8.20.tar.gz'
+  source_sha256 'f6d6aebc5ab835f0593b9c30debb39bec05c8073a486aed661e09da3ebe488f9'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'automake' => :build
+  depends_on 'libtool' => :build
+  depends_on 'glib' => :build
+  depends_on 'aspell'
+  depends_on 'slang'
+
+  def self.build
+    system './autogen.sh'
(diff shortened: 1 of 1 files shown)


uberhacker:
@VitaliyKulikov: Does this PR add what you were trying to build?

VitaliyKulikov:
@uberhacker yep. hmm, is it working !? awesome 👍 
still need few packages to not use `chroot` or `crouton` :
- maven (not tried yet, seems should work ... just shell script based on jdk) and gradle (same)
- vscode or atom (seems challenging)
- meld (sometimes for three-way-compare during git rebase)

uberhacker:
@VitaliyKulikov: Yes, however, the `slang` package is not building correctly for x86_64 systems since the libraries are not in the correct location.  I'll push a PR to fix soon.

uberhacker:
@VitaliyKulikov: I would :heart: to get vscode or atom working in Chromebrew.  However, the problem is these editors require GUI support which is not currently available.  You might be able to get these apps working in CloudReady.  I've had luck with several GUI apps.  If you are interested in helping with the effort to get the GUI working in Chromebrew, see [this issue[(#889).

uberhacker:
It looks like meld is a graphic app also.  You could use vimdiff in Chromebrew as an alternative.  Not as visually appealing as meld, I'm sure, but I use it all the time and find it "good enough" in most cases.

VitaliyKulikov:
@uberhacker thank you for your feedback )) btw, could you provide correct URL for `get the GUI working in Chromebrew, see this issue.`

uberhacker:
@VitaliyKulikov: Sorry, the link has been corrected.  GitHub markdown doesn't work quite like I expected.

PR #1685, opened 2018-01-23
Add maven package

uberhacker:
Apache Maven is a software project management and comprehension tool. Based on the concept of a project object model (POM), Maven can manage a project's build, reporting and documentation from a central piece of information.  See https://maven.apache.org/.


uberhacker:
@VitaliyKulikov: Does this look like what you need?

PR #1686, opened 2018-01-23
Add gradle package

uberhacker:
From mobile apps to microservices, from small startups to big enterprises, Gradle helps teams build, automate and deliver better software, faster.  See https://gradle.org/.


uberhacker:
@VitaliyKulikov: Does this look like what you need?

VitaliyKulikov:
@uberhacker you are awesome )) thanks.

PR #1696, opened 2018-01-24
Add wol package

gergesh:
wol implements Wake On LAN functionality in a small program. It wakes up hardware that is Magic Packet compliant.

### DIFF
--- a/packages/wol.rb
+++ b/packages/wol.rb
@@ -10,7 +10,7 @@ class Wol < Package
   depends_on 'compressdoc' => :build
 
   def self.build
-    system "./configure", "--prefix=#{CREW_PREFIX}", "--mandir #{CREW_PREFIX}/share/man"
+    system "./configure", "--prefix=#{CREW_PREFIX}", "--mandir=#{CREW_PREFIX}/share/man"
     system "make"
   end
 
--- a/packages/wol.rb
+++ b/packages/wol.rb
@@ -3,28 +3,20 @@ require 'package'
 class Wol < Package
   description 'Wake up hardware that is Magic Packet compliant'
   homepage 'http://ahh.sourceforge.net/wol/'
-  version '0.7.1-r1'
+  version '0.7.1'
   source_url 'https://downloads.sourceforge.net/ahh/wol-0.7.1.tar.gz'
   source_sha256 'e0086c9b9811df2bdf763ec9016dfb1bcb7dba9fa6d7858725b0929069a12622'
 
   depends_on 'compressdoc' => :build
 
   def self.build
-     system"./configure \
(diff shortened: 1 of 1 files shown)


gergesh:
Little fuckup: I couldn't figure out how adding to the manpage works, so it's in the dependencies but doesn't actually do anything. The man file in the source is in /doc/wol.1

uberhacker:
@gergesh: Thank you for the new package!  You could try `bash configure --help | less` and see what options are available.  There might be something like `--mandir` where you can set the man file location.

gergesh:
I've fixed the manpage compression and removed a lot of unnecessary configure options, I think it should be okay now.

packages/wol.rb
```diff
@@ -3,28 +3,20 @@
...
-            --prefix=#{CREW_PREFIX} \
-            --enable-gui=no \
-            --with-features=huge \
-            --without-x \
-            --disable-nls \
-            --enable-multibyte \
-            --with-tlib=ncurses \
-            --enable-perlinterp \
-            --disable-selinux"
+    system "./configure", "--prefix=#{CREW_PREFIX}", "--mandir #{CREW_PREFIX}/share/man"
```
uberhacker: Missing `=` in between `--mandir` and `#{CREW_PREFIX}/share/man`?

gergesh: Whoops, guess that's what happens when you make last-minute cosmetic changes without checking. I've fixed this and will do one last test to make sure everything functions properly.

gergesh:
Program itself is working and locales are placed in the correct directory, but the manpages don't get installed and I have no idea why,

uberhacker:
@gergesh: I think you have it correct.  Did you install the `manpages` package?  That lets you access the man stuff.  I tested and it looks fine:
```
$ crew files wol
(i) wol: Wake up hardware that is Magic Packet compliant
http://ahh.sourceforge.net/wol/
version 0.7.1
/usr/local/bin/wol
/usr/local/bin/wol-bootptab
/usr/local/bin/wol-dhcpdconf
/usr/local/info/wol.info.gz
/usr/local/share/locale/de/LC_MESSAGES/wol.mo
/usr/local/share/locale/es/LC_MESSAGES/wol.mo
/usr/local/share/locale/fr/LC_MESSAGES/wol.mo
/usr/local/share/locale/it/LC_MESSAGES/wol.mo
/usr/local/share/locale/sv/LC_MESSAGES/wol.mo
/usr/local/share/man/man1/wol.1.gz
Total found: 10
```

gergesh:
H‌uh, guess that does it. Thanks a lot for the help!

PR #1717, opened 2018-02-03
Add dlang support (ldc2,ldmd2,dub)

Kozzi11:

### DIFF
--- a/packages/dub.rb
+++ b/packages/dub.rb
@@ -1,0 +1,26 @@
+require 'package'
+
+class Dub < Package
+  description 'Developer package manager for D programming language'
+  homepage 'https://github.com/D-Programming-Language/dub'
+  version '1.7.1'
+  source_url 'https://github.com/dlang/dub/archive/v1.7.1.tar.gz'
+  source_sha256 'baa8c533f59d83f74e89c06f5ec7e52daf3becb227c7177a9eeab7159ba86dbc'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'ldc'
+
+  def self.build
+    system "echo \"module dub.version_;\" > source/dub/version_.d"
+    system "echo 'enum dubVersion = \"1.7.1\";' >> source/dub/version_.d"
+    system "ldmd2 -ofbin/dub -w -O -g -version=DubUseCurl -Isource -L-lcurl @build-files.txt"
+  end
+
+  def self.install
+    system "install", "-Dm755",  "bin/dub", "#{CREW_DEST_PREFIX}/bin/dub"
+  end
+end
--- a/packages/ldc.rb
+++ b/packages/ldc.rb
@@ -1,0 +1,40 @@
+require 'package'
+
+class Ldc < Package                 # The first character of the class name must be upper case
+  description 'D language compiler using LLVM.'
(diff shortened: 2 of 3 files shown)


packages/ldc.rb
```diff
@@ -0,0 +1,39 @@
...
+  depends_on 'ncurses'
+  depends_on 'zlibpkg'
+  depends_on 'libconfig' => :build
+  depends_on 'cmake' => :build
+  depends_on 'libedit' => :build
+
+  def self.build                   # the steps required to build the package
+    system "mkdir", "build"
+    system "curl -L https://github.com/ldc-developers/ldc/releases/download/v0.17.5/ldc-0.17.5-src.tar.gz \
+            | tar xzf - -C build"
```
uberhacker: I guess I'm not understanding why this step is needed.  Isn't this the same project but a different version?

packages/dub.rb
```diff
@@ -0,0 +1,26 @@
...
+  depends_on 'ldc'
+
+  def self.build
+    system "echo \"module dub.version_;\" > source/dub/version_.d"
+    system "echo 'enum dubVersion = \"1.7.1\";' >> source/dub/version_.d"
+    system "ldmd2 -ofbin/dub -w -O -g -version=DubUseCurl -Isource -L-lcurl @build-files.txt"
+  end
+
+  def self.install
+    system "install", "-Dm755",  "bin/dub", "#{CREW_DEST_DIR}/usr/local/bin/dub"
```
uberhacker: Please use the crew constant `#{CREW_DEST_PREFIX}` instead of `#{CREW_DEST_DIR}/usr/local`.

packages/libconfig.rb
```diff
@@ -0,0 +1,23 @@
...
+  source_url 'https://hyperrealm.github.io/libconfig/dist/libconfig-1.7.2.tar.gz'
+  source_sha256 '7c3c7a9c73ff3302084386e96f903eb62ce06953bb1666235fac74363a16fad9'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.build
+    system './configure --prefix=/usr/local'
```
uberhacker: Please use `#{CREW_PREFIX}` instead of `/usr/local` and enclose in double quotes.  Also, if there are any library files installed in `/usr/local/lib(64)`, those should be abstracted with `--libdir=#{CREW_LIB_PREFIX}`.

packages/ldc.rb
Kozzi11: Because this version is last C++ version, so it is needed for bootstraping, because current ldc codebase is written in D language itself, so there is a need of some D compiler to compile it.

uberhacker:
@Kozzi11: Awesome!  Thank you for the new packages.  Please see inline comments in the `Files changed` section.

packages/ldc.rb
uberhacker: Thanks for the explanation.  That makes sense.  We should probably do a checksum on this tarball just to make sure it downloaded correctly.  Something like:
```
abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('ldc-0.17.5-src.tar.gz') ) == '12345abcde...'
```

Kozzi11: Thats a good idea. I will add it

PR #1718, opened 2018-02-04
Update jdk8 from 8u152 to 8u162

uberhacker:
Tested and working on:
- [x] armv7l
- [x] i686
- [x] x86_64


thedamian:
I don't like that we're using the real java now a day instead of openJDK. but I DO like it's the latest. I'm merging this!

PR #1725, opened 2018-02-05
Update heroku.rb

skiman6010:


uberhacker:
I'll expect another update within the next day or so. LoL

PR #1726, opened 2018-02-05
Update gdb.rb

skiman6010:


uberhacker:
Excellent!  This should fix the issue of the previous version which could not be compiled.

PR #1733, opened 2018-02-06
Create graphene.rb

lovedheart:
Dependency for building gtk

#889

### DIFF
--- a/packages/graphene.rb
+++ b/packages/graphene.rb
@@ -18,8 +18,14 @@ class Graphene < Package
 
   def self.install
     Dir.chdir("_build") do
-     system "ninja test"
      system "ninja install"
     end
   end
+  
+  def self.check
+    Dir.chdir("_build") do
+     system "ninja test"
+    end
+  end  
+  
 end
--- a/packages/graphene.rb
+++ b/packages/graphene.rb
@@ -1,0 +1,25 @@
+require 'package'
+
+class Graphene < Package
+  description 'A thin layer of graphic data types'
+  homepage 'https://github.com/ebassi/graphene'
+  version '1.6.0'
+  source_url 'https://github.com/ebassi/graphene/archive/1.6.0.tar.gz'
+  source_sha256 '98970f859e452ce421b72726ca727fdf3ac27cb4804b62bfe520157fa46aa2fd'
+
+  depends_on 'ninja'
+  depends_on 'python3'
+  depends_on 'gobject_introspection'
+
+  def self.build
+    system "pip3 install --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR} meson"    # depends_on meson
+    system "meson _build --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} -Denable-arm-neon=false"
+  end
+
+  def self.install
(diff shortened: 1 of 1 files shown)


packages/graphene.rb
```diff
@@ -0,0 +1,25 @@
...
+  version '1.6.0'
+  source_url 'https://github.com/ebassi/graphene/archive/1.6.0.tar.gz'
+  source_sha256 '98970f859e452ce421b72726ca727fdf3ac27cb4804b62bfe520157fa46aa2fd'
+
+  depends_on 'ninja'
+  depends_on 'python3'
+  depends_on 'gobject_introspection'
+
+  def self.build
+    system "pip3 install --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR} meson"    # depends_on meson
```
uberhacker: I'm thinking meson is a good candidate for a separate package.  I can see this being used as a dependency for other packages as well.  Also, isolating it out as a separate package helps with maintenance.  There is no way to uninstall meson if you install it in this manner.

packages/graphene.rb
```diff
@@ -0,0 +1,25 @@
...
+    system "meson _build --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} -Denable-arm-neon=false"
+  end
+
+  def self.install
+    Dir.chdir("_build") do
+     system "ninja test"
```
uberhacker: This test line should probably split into the `def self.check` section.

uberhacker:
Awesome work on chipping away at #889 @lovedheart!

packages/graphene.rb
lovedheart: How to crea a separate package for meson? @uberhacker

uberhacker: Would this work?
```
require 'package'

class Meson < Package
  description 'The Meson Build System'
  homepage 'http://mesonbuild.com/'
  version '0.44.0'
  source_url 'https://github.com/mesonbuild/meson/releases/download/0.44.0/meson-0.44.0.tar.gz'
  source_sha256 '50f9b12b77272ef6ab064d26b7e06667f07fa9f931e6a20942bba2216ba4281b'

  depends_on 'python3'

  def self.install
    system "pip3 install --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR} meson"
  end
end
```
You might also want to take a look at https://github.com/mesonbuild/meson to see if it would be worth compiling from source.  Either way would be fine, I think.

lovedheart: It works

packages/graphene.rb
```diff
@@ -0,0 +1,30 @@
+require 'package'
+
+class Graphene < Package
+  description 'A thin layer of graphic data types'
+  homepage 'https://github.com/ebassi/graphene'
+  version '1.6.0'
+  source_url 'https://github.com/ebassi/graphene/archive/1.6.0.tar.gz'
+  source_sha256 '98970f859e452ce421b72726ca727fdf3ac27cb4804b62bfe520157fa46aa2fd'
+
+  depends_on 'ninja'
```
uberhacker: Since ninja is a dependency of meson, you don't need to add it here.

PR #1734, opened 2018-02-06
Create util_macros.rb

lovedheart:
The util-macros package contains the m4 macros used by all of the Xorg packages

### DIFF
--- a/packages/util_macros.rb
+++ b/packages/util_macros.rb
@@ -1,0 +1,19 @@
+require 'package'
+
+class Util_macros < Package
+  description 'The util-macros package contains the m4 macros used by all of the Xorg packages'
+  homepage 'https://www.linuxfromscratch.org/blfs/view/svn/util-macros.html'
+  version '1.19.1'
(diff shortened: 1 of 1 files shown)


packages/util_macros.rb
```diff
@@ -0,0 +1,19 @@
...
+  description 'The util-macros package contains the m4 macros used by all of the Xorg packages'
+  homepage 'https://www.linuxfromscratch.org/blfs/view/svn/util-macros.html'
+  version '1.19.1'
+  source_url 'https://www.x.org/pub/individual/util/util-macros-1.19.1.tar.bz2'
+  source_sha256 '18d459400558f4ea99527bc9786c033965a3db45bf4c6a32eefdc07aa9e306a6'
+
+
+  def self.build
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
+    #system "make"
```
uberhacker: It won't hurt to uncomment this or is there a specific reason you don't need to execute?

PR #1735, opened 2018-02-06
Create meson.rb

lovedheart:
Add meson

### DIFF
--- a/packages/meson.rb
+++ b/packages/meson.rb
@@ -1,0 +1,15 @@
+require 'package'
+
+class Meson < Package
+  description 'The Meson Build System'
+  homepage 'http://mesonbuild.com/'
+  version '0.44.0'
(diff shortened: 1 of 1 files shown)


packages/meson.rb
```diff
@@ -0,0 +1,15 @@
+require 'package'
+
+class Meson < Package
+  description 'The Meson Build System'
+  homepage 'http://mesonbuild.com/'
+  version '0.44.0'
+  source_url 'https://github.com/mesonbuild/meson/releases/download/0.44.0/meson-0.44.0.tar.gz'
+  source_sha256 '50f9b12b77272ef6ab064d26b7e06667f07fa9f931e6a20942bba2216ba4281b'
+  
+  depends_on 'python3'
```
uberhacker: After looking at https://github.com/mesonbuild/meson, it appears ninja is also a dependency.  Sorry I didn't mention it in my example.

PR #1736, opened 2018-02-06
Update intltool.rb

lovedheart:
Since intltool-update fails with perl 5.26, see https://bugs.launchpad.net/intltool/+bug/1696658,   
applying patch from https://github.com/Alexpux/MSYS2-packages/blob/master/intltool/perl-5.22-compatibility.patch

### DIFF
--- a/packages/intltool.rb
+++ b/packages/intltool.rb
@@ -13,6 +13,7 @@ class Intltool < Package
 
   def self.patch
     system "wget https://raw.githubusercontent.com/Alexpux/MSYS2-packages/master/intltool/perl-5.22-compatibility.patch"
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('perl-5.22-compatibility.patch') ) == '9c6527072aada6e3cb9aceb6e07cfdf51d58839a2beb650168da0601a85ebda3'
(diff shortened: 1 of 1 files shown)


packages/intltool.rb
```diff
@@ -7,22 +7,14 @@ class Intltool < Package
...
-       i686: 'f2177c902c6bc94da49ec78c436f01d99abc68e35650228ac9835f789d123c66',
-     x86_64: 'e6a6069282bf20d1fdd70b6619a839bcbd8a311ac840eed36418a143515e291e',
-  })
-
   depends_on 'libtool'
   depends_on 'perl_xml_parser'
+  depends_on 'patch' => 'build'
 
+  def self.patch
+    system "wget https://raw.githubusercontent.com/Alexpux/MSYS2-packages/master/intltool/perl-5.22-compatibility.patch"
```
uberhacker: Can you please add the following line?  This will make sure the file has been downloaded successfully before continuing to apply the patch.
```
abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('perl-5.22-compatibility.patch') ) == '9c6527072aada6e3cb9aceb6e07cfdf51d58839a2beb650168da0601a85ebda3'
```

uberhacker:
Nice find, btw.

lovedheart:
intltool is required to build some packages

PR #1740, opened 2018-02-06
Create libxi.rb

lovedheart:
Dependency for building gtk

#889


packages/libxi.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class Libxi < Package
+  description 'X.org libXi Client library for XInput'
+  homepage 'https://x.org'
+  version '1.7.9'
+  source_url 'https://github.com/mirror/libXi/archive/libXi-1.7.9.tar.gz'
+  source_sha256 'e3bc48654d4c21ac37592e8b41c87a5de73872a243e7b0fb39ebd565be5b943d'
+
+
+  depends_on 'libx11'
```
uberhacker: In order to use `autogen.sh` there are several dependencies needed.  You will need at least `automake` and maybe a few other packages.  Please test on a fresh Chromebrew install and add the build dependencies.

PR #1746, opened 2018-02-06
Update cairo.rb

lovedheart:
cairo-xlib is requied to build gtk

#889

### DIFF
--- a/packages/cairo.rb
+++ b/packages/cairo.rb
@@ -22,3 +22,4 @@ class Cairo < Package
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
   end
+end
--- a/packages/cairo.rb
+++ b/packages/cairo.rb
@@ -14,7 +14,12 @@ class Cairo < Package
   depends_on 'mesa'
 
   def self.build
-    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --enable-xlib --enable-xlib-xcb --enable-glesv2"
+    system "./configure \
+            --prefix=#{CREW_PREFIX} \
+            --libdir=#{CREW_LIB_PREFIX} \
+            --enable-xlib \
+            --enable-xlib-xcb \
+            --enable-glesv2"
     system "make"
   end
 
--- a/packages/cairo.rb
+++ b/packages/cairo.rb
@@ -1,3 +1,4 @@
+
 require 'package'
 
 class Cairo < Package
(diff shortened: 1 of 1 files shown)


packages/cairo.rb
```diff
@@ -7,32 +8,17 @@ class Cairo < Package
...
   depends_on 'fontconfig' # pango requires cairo with fontconfig
+  depends_on 'gobject_introspection'
+  depends_on 'mesa'
 
   def self.build
-    system './configure',
-           "--prefix=#{CREW_PREFIX}",
-           "--libdir=#{CREW_LIB_PREFIX}",
-           '--disable-xlib'
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --enable-xlib --enable-xlib-xcb --enable-glesv2"
```
uberhacker: Nice!  Can you please keep the separation of one configure option per line?  It just makes it easier to examine and add/remove options in the future.

packages/cairo.rb
```diff
@@ -7,32 +8,17 @@ class Cairo < Package
...
     system "make"
   end
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
   end
-end
```
uberhacker: Don't we need the ending `end` here?

packages/cairo.rb
lovedheart: I got error if I do that. I don't know why

uberhacker:
@lovedheart: I really appreciate your efforts to add all these new packages and improvements to existing packages but please test before submitting PRs.  Most the packages you submitted earlier, I wasn't able to compile either due to missing dependencies or missing packages altogether.  If you test before you commit, it makes it much easier on me since I have to generate binaries.
Thanks

packages/cairo.rb
uberhacker: You need a trailing comma on each line or you can concatenate with a trailing `\`.  Also, you need all the crew constants to be enclosed in double quotes or the variables won't expand.  Hope that helps.

packages/cairo.rb
```diff
@@ -7,28 +8,14 @@ class Cairo < Package
   source_url 'https://www.cairographics.org/releases/cairo-1.14.12.tar.xz'
```
uberhacker: Please bump the version to detect the changes, otherwise `crew update` will not update previous installs of cairo.

lovedheart: How to do that?

uberhacker: `version '1.14.12-1'` should do it.

packages/cairo.rb
uberhacker: Please change to:
```
system "./configure \
  --prefix=#{CREW_PREFIX} \
  --libdir=#{CREW_LIB_PREFIX} \
  --enable-xlib \
  --enable-xlib-xcb \
  --enable-glesv2"
```

uberhacker:
Please email a subscribe message to `crewonslack@gmail.com` and I will send you an invite to join our Slack group.

PR #1747, opened 2018-02-06
Update pango.rb

lovedheart:
Update pango build to avoid compilation error to build gtk+

Works properly:
- [ ] x86_64
- [x] armv7l

### DIFF
--- a/packages/pango.rb
+++ b/packages/pango.rb
@@ -7,19 +7,6 @@ class Pango < Package
   source_url 'https://ftp.gnome.org/pub/gnome/sources/pango/1.41/pango-1.41.0.tar.xz'
   source_sha256 '1f76ef95953dc58ee5d6a53e5f1cb6db913f3e0eb489713ee9266695cae580ba'
 
-#  binary_url ({
-#    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/pango-1.41-chromeos-armv7l.tar.xz',
-#     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/pango-1.41-chromeos-armv7l.tar.xz',
-#       i686: 'https://dl.bintray.com/chromebrew/chromebrew/pango-1.41-chromeos-i686.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/pango.rb
```diff
@@ -3,35 +3,39 @@
...
+#     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/pango-1.41-chromeos-armv7l.tar.xz',
+#       i686: 'https://dl.bintray.com/chromebrew/chromebrew/pango-1.41-chromeos-i686.tar.xz',
+#     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/pango-1.41-chromeos-x86_64.tar.xz',
+#  })
+#  binary_sha256 ({
+#    aarch64: '67f8f261f2a069225b2858fe6dc622a0a1eb4a04e1172be6e3bb16ce09c21785',
+#     armv7l: '67f8f261f2a069225b2858fe6dc622a0a1eb4a04e1172be6e3bb16ce09c21785',
+#       i686: '032ddf476182d20ee2db104abae1bf918db4d14ef931e8011231def15a9eaffc',
+#     x86_64: '88e9f5e2c23542edbb724011fb1bd9190770de53fcf0c2a382ea4e26ed5579b3',
+#  })
```
uberhacker: If you could just remove these lines altogether please, that would be better.

packages/pango.rb
```diff
@@ -3,35 +3,39 @@
...
+  depends_on 'python27'                # need six package from python to install Pango
+
 
   def self.build
     system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
     system "make"
   end
 
   def self.install
+    system "pip install six"
```
uberhacker: Should this be a separate package?  Also, you will need `--prefix #{CREW_PREFIX}` and `--root #{CREW_DEST_DIR}` so the files can be tracked by the package.

lovedheart: Thanks for the tip

lovedheart:
Based on this mod , I built the gtk+ locally

PR #1748, opened 2018-02-06
Update gobject_introspection.rb

lovedheart:
Correct dependency

Tested from a clean crew


uberhacker:
Right, bison is already a dependency of flex.

PR #1750, opened 2018-02-07
Update pango.rb

lovedheart:
"pip install --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR} six" does not work.  six packages not found.
Fix this error using normal installation of six in python

### DIFF
--- a/packages/pango.rb
+++ b/packages/pango.rb
@@ -22,7 +22,8 @@ class Pango < Package
   end
 
   def self.install
-    system "pip install --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR} six"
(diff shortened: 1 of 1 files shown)


uberhacker:
Try this:
`"pip install --target=#{CREW_DEST_DIR} six"`

uberhacker:
I suppose if you are going to uninstall it anyway, it doesn't matter.  However, is the python six package a run-time dependency?  How does it work with the other packages that depend on this one?

lovedheart:
I think it is a run-time dependency. The last step of installation requires six to finish the process, not the compilation procedure.

uberhacker:
If it's a run-time dependency then it shouldn't be uninstalled, right?  If it's a build dependency on the other hand, it might be OK.

PR #1753, opened 2018-02-07
Update cairo.rb

lovedheart:
Fix compilation error


uberhacker:
Please bump the version.  You should get in the habit of incrementing the version any time there are changes and new binaries are needed.  Something like `version '1.14.12-2'` should be fine.

PR #1755, opened 2018-02-07
Update xkeyboard_config.rb

lovedheart:
Fix dependencies

### DIFF
--- a/packages/xkeyboard_config.rb
+++ b/packages/xkeyboard_config.rb
@@ -10,6 +10,8 @@ class Xkeyboard_config < Package
 
   depends_on "util_macros"
   depends_on "intltool"
+  depends_on "libx11"
+  depends_on "gettext" => :build
 
   def self.build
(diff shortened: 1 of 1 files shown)


packages/xkeyboard_config.rb
```diff
@@ -10,6 +10,8 @@ class Xkeyboard_config < Package
 
   depends_on "util_macros"
   depends_on "intltool"
```
uberhacker: This is likely a build dependency.

lovedheart: Is there any difference if  marked as "build"?

uberhacker: The difference is this package is not installed needlessly if the binary build doesn't require it as a runtime dependency.  Since this is a small package, it doesn't really make much difference but when you include the go or jdk8 packages needlessly, this can take up quite a bit of disk space.  An example would be the hub package which only requires go to compile, not to run.

Issue #1531, opened 2017-12-20
Binaries for perl_xml_parser are not compatible with current perl


Alanaktion:
The `perl_xml_parser` package includes the latest upstream code, but the binaries included in the package install to the old perl directory, preventing the library from being used with the latest `perl` package. Installing from source works around the issue, but the binaries need to be updated to use the current perl lib path.

Fixing this in a way that keeps the version number matching the upstream while still triggering the package to install again on existing setups might need some discussion, which was started in #1123.

uberhacker:
@Alanaktion: Thank you for reporting this issue.  I imagine we need to update all the perl packages.  PRs welcome. :)

JL2210:
I think this has been fixed by #1756.

PR #1756, opened 2018-02-07
Corrupted binary for armv7l  (perl_xml_parser.rb)

lovedheart:
Force to build binary from source code for armv7l


uberhacker:
I would prefer to include the binaries as a complete set, if possible.  If the current binary does not install correctly, please make any changes, remove the current binaries, increment the version and we will generate another complete set of new binaries.

lovedheart:
It should fix #1531

PR #1757, opened 2018-02-07
Update gdk_pixbuf.rb

lovedheart:
Fix installation error


packages/gdk_pixbuf.rb
```diff
@@ -11,13 +11,16 @@ class Gdk_pixbuf < Package
...
 
   def self.build
     system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
     system "make"
   end
 
   def self.install
+    system "pip install six" # for installation process
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "pip uninstall --yes six"
```
uberhacker: Are you sure this is only a build dependency?

PR #1758, opened 2018-02-07
Create gtk3.rb

lovedheart:
GTK+3 is available now.
#889  

Works properly:
- [] x86_64
- [x] armv7l

### DIFF
--- a/packages/gtk3.rb
+++ b/packages/gtk3.rb
@@ -1,0 +1,34 @@
+require 'package'
+
+class Gtk3 < Package
+  description 'Gtk3 is a cross-platform widget toolkit for creating graphical user interfaces.'
+  homepage 'https://developer.gnome.org/gtk3/3.0/'
+  version '3.26'
+  source_url 'http://ftp.gnome.org/pub/gnome/sources/gtk+/3.22/gtk+-3.22.26.tar.xz'
+  source_sha256 '61eef0d320e541976e2dfe445729f12b5ade53050ee9de6184235cb60cd4b967'
+
+  depends_on 'glib'
+  depends_on 'gobject_introspection'
+  depends_on 'gdk_pixbuf'
+  depends_on 'cairo'
+  depends_on 'pango'
+  depends_on 'libepoxy'
+  depends_on 'graphene'
+  depends_on 'atk'
+  depends_on 'libxkbcommon'
+  depends_on 'meson'
+  depends_on 'ninja'
+  depends_on 'python3'
+  depends_on 'libx11'
+  depends_on 'at_spi2_atk'
+
+  def self.build
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+
+end
--- a/packages/gtk3.rb
+++ b/packages/gtk3.rb
@@ -21,6 +21,7 @@ class Gtk3 < Package
   depends_on 'python3'
   depends_on 'libx11'
   depends_on 'at_spi2_atk'
+  depends_on 'python27'
 
   def self.build
     system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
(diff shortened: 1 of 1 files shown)


packages/gtk3.rb
```diff
@@ -0,0 +1,36 @@
+require 'package'
+
+class Gtk3 < Package
+  description 'Gtk3 is a cross-platform widget toolkit for creating graphical user interfaces.'
+  homepage 'https://developer.gnome.org/gtk3/3.0/'
+  version '3.26'
+  source_url 'http://ftp.gnome.org/pub/gnome/sources/gtk+/3.22/gtk+-3.22.26.tar.xz'
```
uberhacker: Please change to use https.

packages/gtk3.rb
```diff
@@ -0,0 +1,36 @@
...
+  depends_on 'glib'
+  depends_on 'gobject_introspection'
+  depends_on 'gdk_pixbuf'
+  depends_on 'cairo'
+  depends_on 'pango'
+  depends_on 'libepoxy'
+  depends_on 'graphene'
+  depends_on 'atk'
+  depends_on 'libxkbcommon'
+  depends_on 'meson'
```
uberhacker: Since meson already includes python27 and libx11 as dependencies, they don't need to be included here.

packages/gtk3.rb
```diff
@@ -0,0 +1,36 @@
...
+  depends_on 'ninja'
+  depends_on 'python3'
```
uberhacker: Since python3 is already a dependency of ninja, it doesn't need to be included here.

beliefless-dogmasless: @uberhacker:
how do you keep the list of dependency, do you have a excel thing/or a script to know, what has 'already' been done/compiled.

packages/gtk3.rb
```diff
@@ -0,0 +1,36 @@
...
+  description 'Gtk3 is a cross-platform widget toolkit for creating graphical user interfaces.'
+  homepage 'https://developer.gnome.org/gtk3/3.0/'
+  version '3.26'
+  source_url 'http://ftp.gnome.org/pub/gnome/sources/gtk+/3.22/gtk+-3.22.26.tar.xz'
+  source_sha256 '61eef0d320e541976e2dfe445729f12b5ade53050ee9de6184235cb60cd4b967'
+
+  depends_on 'glib'
+  depends_on 'gobject_introspection'
+  depends_on 'gdk_pixbuf'
+  depends_on 'cairo'
```
uberhacker: Since pango already includes cairo, you don't need to include it here.

packages/gtk3.rb
```diff
@@ -0,0 +1,36 @@
+require 'package'
+
+class Gtk3 < Package
+  description 'Gtk3 is a cross-platform widget toolkit for creating graphical user interfaces.'
+  homepage 'https://developer.gnome.org/gtk3/3.0/'
+  version '3.26'
+  source_url 'http://ftp.gnome.org/pub/gnome/sources/gtk+/3.22/gtk+-3.22.26.tar.xz'
+  source_sha256 '61eef0d320e541976e2dfe445729f12b5ade53050ee9de6184235cb60cd4b967'
+
+  depends_on 'glib'
```
uberhacker: Since pango already includes glib, you don't need to include it here.

packages/gtk3.rb
```diff
@@ -0,0 +1,36 @@
...
+  depends_on 'gobject_introspection'
+  depends_on 'gdk_pixbuf'
+  depends_on 'cairo'
+  depends_on 'pango'
```
uberhacker: Since gdk_pixbuf already includes pango, you don't need to include it here.

packages/gtk3.rb
uberhacker: Before you include a dependency, you should drill down and find if it's already included in another dependency.  Obviously, this is easier when you have fewer dependencies.

uberhacker: I should point out also that you want to avoid dependency loops.  Currently, both the graphene and gdk_pixbuf packages have dependency loops which prevent them from compiling so it's important that we only include the dependencies that are actually needed in a package.

lovedheart: Yeah, the dependency loops should be avoided. The reason why I add all of them is just to show which package should be installed.

PR #1769, opened 2018-02-07
[intltool] Fix depedency

lovedheart:
Add 'wget' for downloading required patch


uberhacker:
Please bump the version.  This also accounts for changes made on earlier commits.

PR #1771, opened 2018-02-07
Update gtk3.rb

lovedheart:
Bump version


uberhacker:
This is more like correcting the version instead of bumping it. :)

PR #1774, opened 2018-02-07
Add itstool.rb

lovedheart:
Add itstool package for further build as dependency


uberhacker:
I'm curious, what is the output of `crew files itstool` after you install?

PR #1775, opened 2018-02-07
Add yelp_xsl

lovedheart:


packages/yelp_xsl.rb
```diff
@@ -0,0 +1,27 @@
+
+require 'package'
+
+class Yelp_xsl < Package
+  description 'yelp-xsl is a collection of programs and data files to help you build, maintain, and distribute documen$
```
uberhacker: It looks like you missed the last part of the description and there is no closing quote at the end.

packages/yelp_xsl.rb
```diff
@@ -0,0 +1,27 @@
+
+require 'package'
+
+class Yelp_xsl < Package
+  description 'yelp-xsl is a collection of programs and data files to help you build, maintain, and distribute documen'
```
uberhacker: Still missing the last part of the description.  Should it be `documentation`?

uberhacker: Please slow down and pay attention to detail. :)

lovedheart: I have tested it several times.

PR #1776, opened 2018-02-07
Add yelp_tools.rb

lovedheart:


packages/yelp_tools.rb
```diff
@@ -0,0 +1,21 @@
+require 'package'
+
+class Yelp_tools < Package
+  description 'yelp-tools is a collection of scripts and build utilities to help create, manage, and publish documentation for Yelp and the web'
+  homepage 'https://github.com/GNOME/yelp-tools'
+  version '3.27.90'
+  source_url 'https://github.com/GNOME/yelp-tools/archive/3.27.90.tar.gz'
+  source_sha256 'da584450bc2373d1a27a922e851f88256c49c9a49a15ed2762e493f37b52a692'
+
+  depends_on 'yelp_xsl'
```
uberhacker: Don't we also need at least `depends_on 'automake'` since we're using autogen.sh?

PR #1778, opened 2018-02-08
Add hicolor_icon_theme

lovedheart:
Add icon-theme


beliefless-dogmasless:
hicolor_icon_theme
don't have .rb  'at the end' so can't be found, nor installed.

Issue #1772, opened 2018-02-07
GTK3 needs improvement


lovedheart:
I have tried to run gtk3-daemon, something works and something not. It needs more effort to improve it.

lovedheart:
#2128

uberhacker:
Latest gtk3 upgrade produces:
```
Performing post-install...
Error opening directory “/usr/local/tmp/crew/dest/usr/local/share/glib-2.0/schemas”: No such file or directory
gtk3 failed to install: `glib-compile-schemas /usr/local/tmp/crew/dest/usr/local/share/glib-2.0/schemas` exited with 1
```

uberhacker:
Installing gtk3 from source on `i686` produces:
```
hecking for BASE_DEPENDENCIES... no
configure: error: Package requirements (glib-2.0 >= 2.49.4    atk >= 2.15.1    pango >= 1.37.3    cairo >= 1.14.0    cairo-gobject >= 1.14.0    gdk-pixbuf-2.0 >= 2.30.0) were not met:

Package 'xxf86vm', required by 'egl', not found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables BASE_DEPENDENCIES_CFLAGS
and BASE_DEPENDENCIES_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.
gtk3 failed to install: `./configure --prefix=/usr/local --libdir=/usr/local/lib --enable-broadway-backend --enable-x11-backend --enable-wayland-backend` exited with 1
```

lovedheart:
I am trying the workaround

uberhacker:
It appears gtk3 is missing 2 dependencies:
```
depends_on 'libxxf86vm'
depends_on 'libxrender'
```

lovedheart:
i prefer to include xorg_lib

uberhacker:
Would it make sense to include `system "glib-compile-schemas #{CREW_DEST_PREFIX}/share/glib-2.0/schemas"` in the `glib` package instead?

lovedheart:
No. I dont' think so

beliefless-dogmasless:
Hy,

Here is a list of 'some that works':
caja-connect-server
caja-file-management-properties
dconf-editor
gcr-viewer
geany
gentoo        # gentoo file manager
gtk3-demo
gtk3-demo-application
gtk3-icon-browser
gtk3-widget-factory
[...]

Issue #1742, opened 2018-02-06
Add more package dependencies to Chromebrew "core"


uberhacker:
I wanted to open this discussion to talk about adding a few more packages to the base Chromebrew install.  The reason I want to introduce new dependencies is several packages require `autogen.sh` to generate a `configure` script.  Rather than add all these necessary dependencies in with each package, would it make sense to just have these packages available by default on a fresh Chromebrew install?

lovedheart:
@uberhacker  It is meaningful to do that.   glib and gobject_introspection should be in the core, since the later compilation highly depends on these.

skiman6010:
I think we should only install the necessary packages needed on a fresh install. It might just be me, but the less "bloat" installed against my will, the better. If `autogen.sh` can be installed later without any problems, I think that is the best way to go.

uberhacker:
@skiman6010: The problem we run into is every package that requires `autogen.sh` needs a few extra dependencies to install without error.  I understand your point but there is a trade-off either way.  We could just include everything on a fresh Chromebrew install or require that every package include the extra dependencies.  The latter could get messy depending on how many packages require `autogen.sh`.

skiman6010:
@uberhacker: So you're saying that `autogen.sh` has extra packages that would also need to be installed before `autogen.sh` could be installed? In that case, I agree with adding them to the fresh install.

uberhacker:
Exactly.  See #1789.  If you can think of other dependencies needed for `autogen.sh`, please mention them in that PR.

uberhacker:
Resolved by #1789.

PR #1781, opened 2018-02-08
Add gusb

lovedheart:
gusb is depencency of colord for gtk3 enhancement

#1772

### DIFF
--- a/packages/gusb.rb
+++ b/packages/gusb.rb
@@ -11,6 +11,8 @@ class Gusb < Package
   depends_on 'glib'
   depends_on 'gtk_doc'
   depends_on 'libusb'
+  depends_on 'libtool' => :build
+  depends_on 'intltool' => :build
 
   def self.build
     system "./autogen.sh"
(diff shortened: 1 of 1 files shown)


packages/gusb.rb
```diff
@@ -0,0 +1,25 @@
+require 'package'
+
+class Gusb < Package
+  description 'GUsb is a GObject wrapper for libusb1'
+  homepage 'https://www.openhub.net/p/gusb'
+  version '0.2.11'
+  source_url 'https://github.com/hughsie/libgusb/archive/gusb_0_2_11.tar.gz'
+  source_sha256 '090eb605e75f8a5b0b3df7ff29d96dd51730850ac89417378d4a8d39fab13702'
+
+  depends_on 'automake' => :build
```
uberhacker: Do we also need `depends_on 'libtool'` or `depends_on 'intltool'` since we're executing autogen.sh?

lovedheart: Could automake , libtool and intltool be merged into buildessential ?

uberhacker: That is basically what I proposed in #1742.

PR #1789, opened 2018-02-08
Update buildessential.rb

lovedheart:
Add more automake, libtool and intltool into buildessential

### DIFF
--- a/packages/buildessential.rb
+++ b/packages/buildessential.rb
@@ -12,8 +12,10 @@ class Buildessential < Package
   depends_on 'make'
   depends_on 'pkgconfig'
   
-  depends_on 'automake' #
-  depends_on 'libtool'  #
-  depends_on 'intltool' # These three are mostly required for "./autogen.sh"
+  # typically required library to compile source code using "./autogen.sh"
+  depends_on 'automake' 
+  depends_on 'libtool'  
+  depends_on 'intltool' 
+  depends_on 'patch'
   
 end


packages/buildessential.rb
```diff
@@ -3,12 +3,17 @@
...
   is_fake
 
   depends_on 'gcc'
   depends_on 'linuxheaders'
   depends_on 'make'
   depends_on 'pkgconfig'
+  
+  depends_on 'automake' #
+  depends_on 'libtool'  #
+  depends_on 'intltool' # These three are mostly required for "./autogen.sh"
```
uberhacker: We might need `patch` and some other packages as well.  We should try recompiling a few more packages that require autogen.sh on a fresh Chromebrew install to confirm.  But yeah, this is essentially what I proposed in #1742.

lovedheart: How about meson build?

lovedheart:
I think we should also need some packages that are required to install the compiled code, such as six from python27.

packages/buildessential.rb
```diff
@@ -3,12 +3,19 @@
...
   depends_on 'gcc'
   depends_on 'linuxheaders'
   depends_on 'make'
   depends_on 'pkgconfig'
+  
+  # typically required library to compile source code using "./autogen.sh"
+  depends_on 'automake' 
+  depends_on 'libtool'  
+  depends_on 'intltool' 
+  depends_on 'patch'
```
uberhacker: I'm going to make a bold proposal and suggest we add these also since we have so many xorg packages and since these seem to common build dependencies on several packages.
```
depends_on 'diffutils'
depends_on 'bison'
depends_on 'flex'
depends_on 'util_macros'
```
This might not be a complete list, however.  Just a start.

lovedheart: I think we need also xorg_proto. Building these xorg packages, I would suggest all the protocols are installed before. It can save depedency checking

uberhacker:
Closing due to duplicate of #1833.

PR #1792, opened 2018-02-08
Update aria2 to support ssl

uberhacker:


uberhacker:
Since ssl is already enabled by default, no need for this update.

PR #1794, opened 2018-02-08
Upgrade glib from 2.52.3 to 2.54.3

lovedheart:
glib >=2.54.0 is required to build gobject_introspection

@uberhacker could you maybe build binary first for this one? Or you can tell me how to build a binary and upload it.

### DIFF
--- a/packages/glib.rb
+++ b/packages/glib.rb
@@ -3,22 +3,10 @@ require 'package'
 class Glib < Package
   description 'GLib provides the core application building blocks for libraries and applications written in C.'
   homepage 'https://developer.gnome.org/glib'
-  version '2.52.3'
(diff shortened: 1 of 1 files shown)


uberhacker:
If you only have an arm Chromebook, this will be difficult.  If you have a Linux, Mac or Windows system, you can install VirtualBox or VMWare and then boot a ChromeOS image.

You can find the `i686` image here: https://github.com/crosbuilder/CustomBuilds/releases

You can find the `x86_86` image (CloudReady) here: http://go.neverware.com/freeVMWare

lovedheart:
I can't bear the compile time on tegra k1

uberhacker:
This is why building binaries is so important.  We need to figure out how to build python so it includes pip because that is a dependency for a lot of other packages and it will take more time to when performing a clean Chromebrew install if you have to continually build from source.

PR #1795, opened 2018-02-08
Update yelp_tools.rb

### LINKED ISSUE #1793: [yelp_tools] Unable to build
```
checking for xsltproc... no
configure: error: xsltproc not found
yelp_tools failed to build: `./autogen.sh` exited with 1
```

lovedheart:
Fix dependency

It should fix #1793 

Confirm it works, since libxslt is the last dependency.


uberhacker:
That was too fast. :)  Please test before you submit a PR.

PR #1800, opened 2018-02-09
Create xcb_util_renderutil.rb

lovedheart:
Add xcb_util_renderutil

### DIFF
--- a/packages/xcb_util_renderutil.rb
+++ b/packages/xcb_util_renderutil.rb
@@ -1,0 +1,20 @@
+require 'package'
+
+class Xcb_util_renderutil < Package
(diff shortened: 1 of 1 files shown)


packages/xcb_util_renderutil.rb
```diff
@@ -0,0 +1,20 @@
+require 'package'
+
+class Xcb_util_renderutil < Package
+  description 'The xcb-util-renderutil package provides additional extensions to the XCB library'
+  homepage 'http://www.linuxfromscratch.org/blfs/view/cvs/x/xcb-util-renderutil.html'
```
uberhacker: Isn't the homepage https://xcb.freedesktop.org/?

PR #1804, opened 2018-02-09
Create libsigcplucpluc.rb

lovedheart:
Add libsigc++

### DIFF
--- a/packages/libsigcplucpluc.rb
+++ b/packages/libsigcplucpluc.rb
@@ -1,0 +1,20 @@
+require 'package'
+
+class Libsigcplusplus < Package
+  description 'http://libsigc.sourceforge.net/'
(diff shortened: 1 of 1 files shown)


packages/libsigcplucpluc.rb
```diff
@@ -0,0 +1,20 @@
+require 'package'
+
+class Libsigcplusplus < Package
+  description 'http://libsigc.sourceforge.net/'
+  homepage 'libsigc++ implements a typesafe callback system for standard C++.'
```
uberhacker: Looks like description and homepage are swapped.

lovedheart: No. This is the official link from libsigc.sourceforge.net

PR #1806, opened 2018-02-09
Create graphviz.rb

lovedheart:


packages/graphviz.rb
```diff
@@ -0,0 +1,25 @@
+require 'package'
+
+class Graphviz < Package
+  description 'Graphviz is open source graph visualization software.'
+  homepage 'https://www.graphviz.org/'
+  version '2.40.1'
+  source_url 'https://graphviz.gitlab.io/pub/graphviz/stable/SOURCES/graphviz.tar.gz'
```
uberhacker: Can we get a url that is versioned?  I think this url could be volatile.

lovedheart: It's difficult to find that

uberhacker: This is the download from the `stable_release_2.40.1` tagged commit hash:
https://gitlab.com/graphviz/graphviz/repository/67cd2e5121379a38e0801cc05cce5033f8a2a609/archive.tar.bz2

packages/graphviz.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class Graphviz < Package
+  description 'Graphviz is open source graph visualization software.'
+  homepage 'https://www.graphviz.org/'
+  version '2.40.1'
+  source_url 'https://gitlab.com/graphviz/graphviz/repository/67cd2e5121379a38e0801cc05cce5033f8a2a609/archive.tar.bz2'
+  source_sha256 '8dddc80b4194b17c1a5bc1cc7b5e001e9a3ec27272287dc16f278c736a29a9b9'
+
+  depends_on 'fontconfig'
+  depends_on 'pango'
```
uberhacker: The cairo package is a dependency of pango and fontconfig is already a dependency of cairo, so no need to include fontconfig here.

packages/graphviz.rb
```diff
@@ -0,0 +1,26 @@
...
+  depends_on 'flex'
+  depends_on 'gdk_pixbuf'
+  depends_on 'libxrender'
+  depends_on 'poppler'
```
uberhacker: Since we're using autogen.sh, we should be including at least `depends_on 'automake' => :build`, correct?

packages/graphviz.rb
```diff
@@ -0,0 +1,27 @@
...
+
+class Graphviz < Package
+  description 'Graphviz is open source graph visualization software.'
+  homepage 'https://www.graphviz.org/'
+  version '2.40.1'
+  source_url 'https://gitlab.com/graphviz/graphviz/repository/67cd2e5121379a38e0801cc05cce5033f8a2a609/archive.tar.bz2'
+  source_sha256 '8dddc80b4194b17c1a5bc1cc7b5e001e9a3ec27272287dc16f278c736a29a9b9'
+
+  depends_on 'automake' => :build
+  depends_on 'fontconfig'
```
uberhacker: We can remove fontconfig, right?

lovedheart: Why? I have not tested it without fontconfig. I follow the suggested depedencies from website.

uberhacker: I already explained the reason.  It's because pango and poppler include cairo as a dependency and fontconfig is a dependency of cairo.

uberhacker: So what this means basically is you don't need to include it since it will be included by default from the other packages.

uberhacker: As a matter of fact, you don't even need pango since it's already a dependency of gtk_pixbuf.

PR #1811, opened 2018-02-09
Create perl_text_wrapi18n.rb

lovedheart:


packages/perl_text_wrapi18n.rb
```diff
@@ -0,0 +1,32 @@
+require 'package'
+
+class Perl_text_wrapi18n < Package
+  description 'Text::WrapI18N - Line wrapping module with support for multibyte, fullwidth, and combining characters and languages without whitespaces betw$
```
uberhacker: This looks like another case where the description got truncated.  There is no ending quote.

PR #1822, opened 2018-02-10
python3 depends on libffi and readline

dhiltonp:
Python3 couldn't build on aarch64, stopping on `ModuleNotFoundError: No module named '_ctypes'`

I used -k, then ran `make` in the build dir.

```
Failed to build these modules:
_ctypes               readline
```

https://bugs.python.org/issue30090 says that _ctypes is provided by libffi, I also installed readline.

### DIFF
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -3,7 +3,7 @@ require 'package'
 class Python3 < Package
   description 'Python is a programming language that lets you work quickly and integrate systems more effectively.'
   homepage 'https://www.python.org/'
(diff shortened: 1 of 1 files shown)


uberhacker:
Hey @dhiltonp:  Thank you for the update!  Can you please bump the version?  Otherwise, the changes will not be detected during an update.
```
- version '3.6.4-0'
+ version '3.6.4-1'
```
Thanks

dhiltonp:
Done!

Issue #1825, opened 2018-02-10
Enhance xorg


lovedheart:
I think it is beneficial to enhance xorg library. I am working on this.

uberhacker:
I really appreciate all the hard work you have done to get the X window system working.  I just hope we can find a good home for all these new binaries!

jrjaggari:
Looks like xauth is missing. I get the following error, startx: xauth: not found

beliefless-dogmasless:
Yes it is missing:
https://www.x.org/archive/individual/app/xauth-1.0.10.tar.gz

uberhacker:
Would you like to add the package @beliefless-dogmasless? :)

beliefless-dogmasless:
The package works, but there is always something which I don't understand, why some gtk applications work and most don't with the "Cannot open display", I'm working on it. But until now, now luck. It's the same with guix(the package manager), packages install themselves, but there is a problem linking the 'interface'.

uberhacker:
Have you looked at CloudReady yet @beliefless-dogmasless?  They include all the X Window libraries out-of-the-box.  You could probably examine the directory structure and determine what is missing.  See https://www.neverware.com/freedownload.

beliefless-dogmasless:
http://www.linuxfromscratch.org/blfs/view/svn/index.html
propose all the X Window libraries out-of-the-box, too, but it's not about the libraries, it's about linking the interface. The last thing that pops up today was checking about iptables, if something blocks it, maybe some applications use "some services" to access the web and block their start, don't know, need to check that further.

PR #1830, opened 2018-02-10
Create xf86dgaproto.rb

lovedheart:
#1825

### DIFF
--- a/packages/xf86dgaproto.rb
+++ b/packages/xf86dgaproto.rb
@@ -1,0 +1,18 @@
+require 'package'
+
+class Xf86dgaproto < Package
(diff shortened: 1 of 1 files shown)


packages/xf86dgaproto.rb
```diff
@@ -0,0 +1,18 @@
+require 'package'
+
+class Xf86dgaproto < Package
+  description 'X.org XF86DGA protocol headers'
+  homepage 'http://www.x.org'
+  version '1.1.2'
```
uberhacker: This version doesn't appear to match the source_url version.

lovedheart:
Fix it

PR #1832, opened 2018-02-10
[libxrender] Fix dependency

lovedheart:
Fix dependencies #1825


uberhacker:
I know it's probably a total pain but when testing new packages but please start with a fresh Chromebrew install each time so missing dependencies will become more obvious.

lovedheart:
The problem is that for each test for xorg packages. It should install python27 that takes time, since python binary is corrupted.

PR #1833, opened 2018-02-10
Update buildessential.rb

lovedheart:
Enhance buildessential

### DIFF
--- a/packages/buildessential.rb
+++ b/packages/buildessential.rb
@@ -11,4 +11,20 @@ class Buildessential < Package
   depends_on 'linuxheaders'
   depends_on 'make'
   depends_on 'pkgconfig'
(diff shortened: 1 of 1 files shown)


uberhacker:
I was thinking about maybe prompting to add build dependencies during install rather than shoving these down people's throats all at once without a choice.  I would imagine only project developers like us who need to compile from source would need the extra packages.  Otherwise, I would think most users who prefer to work with the binary packages instead would not want all this extra fluff.  thoughts?

lovedheart:
I think the binary in buildessential is so not large...

thedamian:
I think it's not just developers that need these tools since a lot of projects need this "build" and DON'T have binaries.

PR #1837, opened 2018-02-10
Update xcb_proto.rb

lovedheart:
Remove python27 depedency. I have tested it without python27 on a fresh crew. It also works.  Python27 is only a optional package, see http://www.linuxfromscratch.org/blfs/view/svn/x/xcb-proto.html


lovedheart:
For safety, I change it to python => :build

PR #1842, opened 2018-02-10
Add new xorg lib

lovedheart:
tested with fresh crew
using "crew install xorg_lib"

### DIFF
--- a/packages/xorg_lib.rb
+++ b/packages/xorg_lib.rb
@@ -1,0 +1,22 @@
+require 'package'
+
+class Xorg_lib < Package
+  is_fake
+
+  depends_on 'libdmx'
+  depends_on 'libfs'
+  depends_on 'libice'
+  depends_on 'libxaw'
+  depends_on 'libxfont2'
+  depends_on 'libxft'
+  depends_on 'libxmu'
+  depends_on 'libxpm'
+  depends_on 'libxrandr'
+  depends_on 'libxres'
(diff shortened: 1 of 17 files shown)


lovedheart:
#1825

packages/xorg_lib.rb
```diff
@@ -0,0 +1,22 @@
+require 'package'
```
uberhacker: This should have a description, homepage and version.  Take a look at `buildessential` for an example.  Everything else looks great, however.

lovedheart: Fix it.

packages/xorg_lib.rb
```diff
@@ -0,0 +1,27 @@
+require 'package'
+
```
uberhacker: Missing `require 'package'`?

packages/xorg_lib.rb
```diff
@@ -0,0 +1,27 @@
...
+  depends_on 'libxft'
+  depends_on 'libxmu'
+  depends_on 'libxpm'
+  depends_on 'libxrandr'
+  depends_on 'libxres'
+  depends_on 'libxt'
+  depends_on 'libxv'
+  depends_on 'libxvmc'
+  depends_on 'libxxf86dga'
+  depends_on 'libxxf86vm'
```
uberhacker: It might help to sort these alphabetically to ensure there are no duplicates.

PR #1846, opened 2018-02-11
Add pre-built binaries for xorg lib packages

uberhacker:


uberhacker:
@thedamian? @gangstead?

PR #1853, opened 2018-02-11
Add cryptsetup

tblaschke:
Add cryptsetup package

Works properly:
- [x] x86_64 (tested on Samsung Chromebook Pro) 
- [ ] aarch64 (don't have any other arch to test)


packages/cryptsetup.rb
```diff
@@ -0,0 +1,29 @@
+require 'package'
+
+class Cryptsetup < Package
+  description 'Cryptsetup is utility used to conveniently setup disk encryption based on DMCrypt kernel module.'
```
uberhacker: Some might say `Cryptsetup is an utility...` would be the proper grammar but I actually prefer `Cryptsetup is a utility...`.  You could also say `The cryptsetup utility is...`.  Either way would be fine.

packages/cryptsetup.rb
```diff
@@ -0,0 +1,29 @@
...
+  depends_on 'popt'
+
+  def self.build
+    system "./configure", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}"
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    #system "make", "install"
```
uberhacker: Please remove this extraneous line.

uberhacker:
Thank you for the new package @tblaschke!  Please see inline comments in the `Files changed` section.

tblaschke:
I updated the description and removed the extra line.

Cryptsetup provides not only tools for encryption but also for decryption. :) 
I'm currently using it to access an encrypted partition on my sdcard.

sudo /usr/local/sbin/cryptsetup luksOpen /dev/mmcblk1p2 sd_crypt
sudo mount /dev/mapper/sd_crypt /media/removable/SD\ Card/

It seems to work with all basic LUKS container.

PR #1854, opened 2018-02-11
Update jsonc.rb

tblaschke:
Fixes build requirements 
Installs jsonc to /usr/lib64/ instead to /usr/lib/


uberhacker:
Thank you for the update @tblaschke!  There is an update to this version, so if you can get it working with https://github.com/json-c/json-c/archive/json-c-0.13-20171207.tar.gz and resubmit, that would be awesome.  Also, not sure we need the `depends_on 'autoconf' => :build` line since there is no call to `autoconf` in the build section.  Did you need a new config for some reason?

tblaschke:
Closing to resubmit

PR #1856, opened 2018-02-11
Update jsonc to 0.13

tblaschke:
Updates jsonc to 0.13

Works properly:
- [x] x86_64 (tested on Samsung Chromebook Pro)

`depends_on 'autoconf' => :build` is needed. Otherwise the build fails on a fresh crew installation with the following message 

> config.status: executing libtool commands
> (CDPATH="${ZSH_VERSION+.}:" && cd . && /bin/bash /usr/local/tmp/crew/json-c-0.13-nodoc.tar.gz.dir/json-c-0.13/missing autoheader)
> /usr/local/tmp/crew/json-c-0.13-nodoc.tar.gz.dir/json-c-0.13/missing: line 81: autoheader: command not found
> WARNING: 'autoheader' is missing on your system.
>          You should only need it if you modified 'acconfig.h' or
>          'configure.ac' or m4 files included by 'configure.ac'.
>          The 'autoheader' program is part of the GNU Autoconf package:
>          <http://www.gnu.org/software/autoconf/>
>          It also requires GNU m4 and Perl in order to run:
>          <http://www.gnu.org/software/m4/>
>          <http://www.perl.org/>
> make: *** [Makefile:493: config.h.in] Error 127
> jsonc failed to build: `make V=0 -j4` exited with 2

### DIFF
--- a/packages/jsonc.rb
+++ b/packages/jsonc.rb
@@ -3,25 +3,14 @@ require 'package'
 class Jsonc < Package
   description 'JSON-C implements a reference counting object model that allows you to easily construct JSON objects in C, output them as JSON formatted strings and parse JSON formatted strings back into the C representation of JSON objects.'
   homepage 'https://github.com/json-c/json-c/wiki'
-  version '0.12.1-nodoc'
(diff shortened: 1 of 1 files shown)


uberhacker:
Fair enough!

PR #1859, opened 2018-02-11
Update libuuid.rb

tblaschke:
Update libuuid. It's part of util-linux now so we have to download the whole util-linux source.
The old libuuid produced lots of warnings in combination with crouton

`mountpoint: /usr/local/lib64/libuuid.so.1: no version information available (required by /lib64/libblkid.so.1)
mountpoint: /usr/local/lib64/libuuid.so.1: no version information available (required by /lib64/libblkid.so.1)
Unmounting /mnt/stateful_partition/crouton/chroots/xenial...
mountpoint: /usr/local/lib64/libuuid.so.1: no version information available (required by /lib64/libblkid.so.1)
mount: /usr/local/lib64/libuuid.so.1: no version information available (required by /lib64/libblkid.so.1)`

### DIFF
--- a/packages/libuuid.rb
+++ b/packages/libuuid.rb
@@ -2,30 +2,18 @@ require 'package'
 
 class Libuuid < Package
   description 'Portable uuid C library'
-  homepage 'https://sourceforge.net/projects/libuuid/'
-  version '1.0.3'
-  source_url 'https://downloads.sourceforge.net/project/libuuid/libuuid-1.0.3.tar.gz'
-  source_sha256 '46af3275291091009ad7f1b899de3d0cea0252737550e7919d17237997db5644'
-
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/libuuid-1.0.3-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/libuuid-1.0.3-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/libuuid-1.0.3-chromeos-i686.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/libuuid.rb
```diff
@@ -2,30 +2,18 @@
...
+  source_url 'https://www.kernel.org/pub/linux/utils/util-linux/v2.31/util-linux-2.31.1.tar.gz'
+  source_sha256 'd636cb23a2a4aaeeec4b1b52c702366a83978f5a3aaf3029431e7a76a7095323'
 
   def self.build
     system './configure',
            "--prefix=#{CREW_PREFIX}",
            "--libdir=#{CREW_LIB_PREFIX}",
-           '--disable-dependency-tracking'
-    system "make"
+           '--disable-all-programs'
```
uberhacker: You will need a comma after this line since you included the entire option in quotes.

uberhacker:
Thank you for the update @tblaschke.  See inline comment.

uberhacker:
We also have the `util_linux` package so it might make sense just to make this a fake package which references that package instead.

tblaschke:
I tested `gptfdisk`: it seems to work (it lists all partitions). Is there any specific test you have in mind? 

I also created a fake package with depends on `util_linux`. It works and provides also libuuid. However, it also pulls quite large dependencies (mainly python3) which we could avoid by maintaining a separate package. 

`chronos@localhost ~ $ crew install libuuid
libuuid: Portable uuid C library
https://www.kernel.org
version 2.31
libxml2 gettext attr fuse libpstat fskit xzutils libffi python3 libcap_ng vdev util_linux `

We route should we take? Creating a separate package to have a smaller footprint on the device or creating a fake package to have a lower maintenance for this package?

uberhacker:
It might even make sense to remove this package entirely and just change the dependency from libuuid to util_linux in `gptfdisk`.

tblaschke:
Close in favor of #1874

PR #1867, opened 2018-02-11
[libdrm] Upgrade from 2.4.82 to 2.4.89, add tegra and exynos api support

lovedheart:
Add tegra and exynos support and update version #1825


uberhacker:
Excellent!  We need more updates like this.  See #1585 for more packages that need updated.  I checked this off the list. :)

lovedheart:
It might need some depedenceis on x86-64 or i686, but it works on armv7l

PR #1868, opened 2018-02-12
[xorg_lib] Update xorg libraries list

lovedheart:
#1825 

This should be the basic depedency of lidrm, mesa and so on.

### DIFF
--- a/packages/xorg_lib.rb
+++ b/packages/xorg_lib.rb
@@ -3,14 +3,30 @@ require 'package'
 class Xorg_lib < Package
   description 'A collection of xorg libraries.'
   homepage ''
-  version '0.1'  
+  version '0.1-0'  
   
   is_fake
  
+  depends_on 'libxtrans'
+  depends_on 'libx11'
+  depends_on 'libxext'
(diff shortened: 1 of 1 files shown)


packages/xorg_lib.rb
```diff
@@ -23,5 +39,7 @@ class Xorg_lib < Package
   depends_on 'libxvmc'
   depends_on 'libxxf86dga'
   depends_on 'libxxf86vm'
+  depends_on 'libxcomposite'
+  depends_on 'libxscrnsaver'
```
uberhacker: Should we sort these to avoid duplicates or does the order matter?

lovedheart: No duplicates, I have tested it with fresh install.

lovedheart: This kind of list might help us to fix previous error. I can know which packages I have worked on.

uberhacker: I was just thinking the list is so long it would be easy to accidentally add a duplicate but if they were sorted, it would be a bit easier to avoid.

Issue #1870, opened 2018-02-12
[cryptsetup] Unable to build


uberhacker:
```
generate-luks2-area-in-json-hdr-space-json0.img.sh...lib.sh: line 107: xxd: command not found
done
generate-luks2-correct-full-json0.img.sh...lib.sh: line 107: xxd: command not found
lib.sh: line 98: xxd: command not found
Header generator ./generate-luks2-correct-full-json0.img.sh failed: '2'
FAILED
FAIL: luks2-validation-test
WARNING: You must be root to run this test, test skipped.
SKIP: luks2-integrity-test
WARNING: You must be root to run this test, test skipped.
SKIP: verity-compat-test
WARNING: You must be root to run this test, test skipped.
SKIP: reencryption-compat-test
WARNING: You must be root to run this test, test skipped.
SKIP: reencryption-compat-test2
WARNING: You must be root to run this test, test skipped.
SKIP: integrity-compat-test
=======================
1 of 3 tests failed
(17 tests were not run)
=======================
make[2]: *** [Makefile:671: check-TESTS] Error 1
make[2]: Leaving directory '/media/tmp/crew/cryptsetup-2.0.1.tar.xz.dir/cryptsetup-2.0.1/tests'
make[1]: *** [Makefile:795: check-am] Error 2
make[1]: Leaving directory '/media/tmp/crew/cryptsetup-2.0.1.tar.xz.dir/cryptsetup-2.0.1/tests'
make: *** [Makefile:2083: check-recursive] Error 1
cryptsetup failed to build: `make check` exited with 2
```

tblaschke:
My mistake. I didn't notice I had `vim` installed. `xxd` is part of `vim`. Either we update the build dependencies to include `vim` or we just skip the make tests.

tblaschke:
#1875 should fix this

PR #1874, opened 2018-02-12
Remove redundant libuuid package

tblaschke:
As mentioned in #1859 

util_linux provides a libuuid. 
The only packages depending on this lib are `cryptsetup` and `gptfdisk`. Both work with the version provided bu util_linux.


uberhacker:
@tblaschke: This looks great.  Not sure if you noticed yet but I had trouble installing cryptsetup.  See https://github.com/skycocker/chromebrew/issues/1870.

PR #1877, opened 2018-02-13
Add Powered by Bintray banner

### LINKED ISSUE #1826: Alternative for Bintray
OK guys, it appears we have reached the 10 gb limit imposed on open source projects at bintray.com.  We need to start thinking about alternative solutions because Bintray wants to charge $150 per month to start since we need more space.  If anybody has ideas for a better alternative binary repository, please state your case and why it would be a good solution.
Thanks

uberhacker:
Good news guys!  Bintray has lifted the 10 gb limit and is now sponsoring this project!  The only request they made was that we display a `Powered by Bintray` image which this PR provides.  Many thanks to the good folks at Bintray for being dedicated to OSS.  Resolves #1826.


bfayers:
Wow! That's great  of bintray! I don't mind a banner atall

PR #1881, opened 2018-02-14
Add aws_shell package

uberhacker:
An integrated shell for working with the AWS CLI.  See https://github.com/awslabs/aws-shell.


VitaliyKulikov:
@uberhacker thanks!

PR #1892, opened 2018-02-18
Update imagemagick packages

### LINKED ISSUE #1887: Imagemagick dependencies?
I'm not sure if this is on purpose, but both `imagemagick` and `imagemagick6` are missing dependencies for them to work. The former doesn't list [any dependencies](https://github.com/skycocker/chromebrew/blob/master/packages/imagemagick.rb) in the package definition, while [the latter](https://github.com/skycocker/chromebrew/blob/master/packages/imagemagick6.rb) lists some but not all of them (`libwebp` and `openjpeg` are still missing).

This results in nasty error messages when trying to run the `convert` command:

    $ convert  
    convert: error while loading shared libraries: libwebp.so.7: cannot open shared object file: No such file or directory
    $ crew install libwebp
    (...)
    $ convert 
    convert: error while loading shared libraries: libopenjp2.so.7: cannot open shared object file: No such file or directory

uberhacker:
Fixes #1887.

Update imagemagick6 from 6.9.9-27 to 6.9.9-34.
Update imagemagick7 from 7.0.7-15 to 7.0.7-22.


beliefless-dogmasless:
Hello, there is no : 
https://www.imagemagick.org/download/releases/ImageMagick-7.0.7-22.tar.xz

but a    https://www.imagemagick.org/download/releases/ImageMagick-7.0.7-23.tar.xz

uberhacker:
@beliefless-dogmasless: Thank you for reporting this issue.  There is no https://www.imagemagick.org/download/releases/ImageMagick-6.9.9-34.tar.xz either.  I'm working on the newer versions that include the pre-built binaries.  There is an issue with the x86_64 builds and @rmrfslashbin is working on a fix, I think.  I'll continue to try and build the binaries myself also.

beliefless-dogmasless:
I have 'checksum mismatch' 'error' for all ImageMagick-6 & 7...

When I download them, without crew, I don't have any 'problem'.

rmrfslashbin:
Indeed, I will be working on compiling the latest version `ImageMagick-7.0.7-23`.

PR #1903, opened 2018-02-21
Add weechat package

gergesh:
An IRC client, perhaps will suit some users better than irssi.

### DIFF
--- a/packages/weechat.rb
+++ b/packages/weechat.rb
@@ -1,0 +1,35 @@
+require 'package'
+
+class Weechat < Package
+  description "WeeChat is a fast, light and extensible chat client."
+  homepage 'http://www.weechat.org/'
+  version '2.0.1'
+  source_url 'https://weechat.org/files/src/weechat-2.0.1.tar.xz'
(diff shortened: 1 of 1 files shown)


packages/weechat.rb
```diff
@@ -0,0 +1,35 @@
...
+  description "WeeChat is a fast, light and extensible chat client."
+  homepage 'http://www.weechat.org/'
+  version '2.0.1'
+  source_url 'https://weechat.org/files/src/weechat-2.0.1.tar.xz'
+  source_sha256 '6943582eabbd8a6fb6dca860a86f896492cae5fceacaa396dbc9eeaa722305d1'
+
+  depends_on 'curl'
+  depends_on 'gnutls'
+  depends_on 'libgcrypt'
+  depends_on 'aspell' => :build
```
uberhacker: Maybe I'm wrong but this seems like it should be a run-time dependency.

uberhacker:
Thank you for the new package @gergesh!  Please see comment in the `Files changed` section.

packages/weechat.rb
gergesh: It is an optional dependency but not required for running the program (confirmed by removing it and entering weechat).

Issue #1910, opened 2018-02-26
Broken Dependency in SVN


HamishCPratt:
Installing the subversion package does not require the installation of filecmd, however running svn without filecmd gives the following error:
`error: svn: error while loading shared libraries: libmagic.so.1: cannot open shared object file: No such file or directory` 

This was on Chrombrew 0.4.3, and simply running `crew install --keep filecmd` fixed all issues

uberhacker:
Thank you for reporting this issue @HamishCPratt.  We will make sure the next svn update includes this dependency.

thedamian:
Fixed in PR https://github.com/skycocker/chromebrew/pull/1912

PR #1912, opened 2018-03-01
addinv filecmd to subversion per issue 1910

thedamian:
Fixes issue https://github.com/skycocker/chromebrew/issues/1910

## Description
We are just adding the filecmd dependancy that's needed by SVN subversion as per issue #1910 

Works properly:
- [x] x86_64
- [x] i386
- [x] aarch64 (reasons why it doesn't)


uberhacker:
@thedamian:  Thank you for the update!  Can you please bump the version?  Otherwise, when folks execute `crew update`, this new dependency won't be detected.  Something like:
```
- version '1.9.7'
+ version '1.9.7-1'
```

PR #1920, opened 2018-03-05
Add Transmission package

beliefless-dogmasless:
I'm on arm, it works well.

### DIFF
--- a/packages/transmission.rb
+++ b/packages/transmission.rb
@@ -1,0 +1,37 @@
+require 'package'
+
+class Transmission < Package
+  description 'A fast, easy, and free BitTorrent client. https://transmissionbt.com/'
+  homepage 'https://github.com/transmission/transmission-releases'
+  version '2.93'
+  source_url 'https://github.com/transmission/transmission-releases/raw/master/transmission-2.93.tar.xz'
+  source_sha256 '8815920e0a4499bcdadbbe89a4115092dab42ce5199f71ff9a926cfd12b9b90b'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+  
+  depends_on 'libevent'
+  depends_on 'openssl'
+  depends_on 'miniupnpc'
+  depends_on 'curl'
+  depends_on 'zlibpkg'
+  depends_on 'glib'
+  depends_on 'rtmpdump'
+  depends_on 'ctorrent'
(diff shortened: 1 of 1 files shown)


packages/transmission.rb
```diff
@@ -0,0 +1,37 @@
...
+  
+  depends_on 'libevent'
+  depends_on 'openssl'
+  depends_on 'miniupnpc'
+  depends_on 'curl'
+  depends_on 'zlibpkg'
+  depends_on 'glib'
+  depends_on 'rtmpdump'
+  depends_on 'ctorrent'
+  depends_on 'intltool'
```
uberhacker: Build only dependency?

packages/transmission.rb
```diff
@@ -0,0 +1,37 @@
...
+  depends_on 'gettext'
+  depends_on 'pkgconfig'
```
uberhacker: Build only dependency?

packages/transmission.rb
```diff
@@ -0,0 +1,37 @@
...
+  depends_on 'libevent'
+  depends_on 'openssl'
+  depends_on 'miniupnpc'
+  depends_on 'curl'
+  depends_on 'zlibpkg'
+  depends_on 'glib'
+  depends_on 'rtmpdump'
+  depends_on 'ctorrent'
+  depends_on 'intltool'
+  depends_on 'gettext'
```
uberhacker: Build only dependency?

packages/transmission.rb
```diff
@@ -0,0 +1,37 @@
...
+  })
+  
+  depends_on 'libevent'
+  depends_on 'openssl'
+  depends_on 'miniupnpc'
+  depends_on 'curl'
+  depends_on 'zlibpkg'
+  depends_on 'glib'
+  depends_on 'rtmpdump'
+  depends_on 'ctorrent'
```
uberhacker: Is this really needed?

beliefless-dogmasless: I copy paste, the gentoo...:

https://gitweb.gentoo.org/repo/gentoo.git/diff/net-p2p/transmission/transmission-2.93.ebuild?id=4c23097d5cc80886f0ccbc5061caa5d10d08406f

beliefless-dogmasless:
I removed gettext, intltool and pkgconfig.

PR #1923, opened 2018-03-05
Add Gpart & Sharedmimeinfo Packages

beliefless-dogmasless:
Hy,

Both work 'fine' on arm.

### DIFF
--- a/packages/gpart.rb
+++ b/packages/gpart.rb
@@ -1,0 +1,25 @@
+require 'package'
+
+class Gpart < Package
+  description 'Partition table rescue/guessing tool. https://github.com/baruch/gpart'
+  homepage 'https://github.com/baruch/gpart/releases'
+  version '0.3'
+  source_url 'https://github.com/baruch/gpart/archive/0.3.tar.gz'
+  source_sha256 'ec56d12ec9ffdb9877c12692ea6e51620b1ae44473d3d253b27fc31ed9ebb4dd'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.build
+    system "sh autogen.sh"
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install" # the steps required to install the package
+  end
+  
+end
--- a/packages/sharedmimeinfo.rb
+++ b/packages/sharedmimeinfo.rb
@@ -1,0 +1,27 @@
+require 'package'
+
+class Sharedmimeinfo < Package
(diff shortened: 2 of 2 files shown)


packages/gpart.rb
```diff
@@ -0,0 +1,25 @@
+require 'package'
+
+class Gpart < Package
+  description 'Partition table rescue/guessing tool. https://github.com/baruch/gpart'
+  homepage 'https://github.com/baruch/gpart/releases'
```
uberhacker:
```
- homepage 'https://github.com/baruch/gpart/releases'
+ homepage 'https://github.com/baruch/gpart'
```

packages/sharedmimeinfo.rb
```diff
@@ -0,0 +1,27 @@
+require 'package'
+
+class Sharedmimeinfo < Package
+  description 'The Shared MIME-info Database specification. https://freedesktop.org/wiki/Software/shared-mime-info'
```
uberhacker:
```
- description 'The Shared MIME-info Database specification. https://freedesktop.org/wiki/Software/shared-mime-info'
+ description 'The Shared MIME-info Database specification.'
```

packages/gpart.rb
```diff
@@ -0,0 +1,25 @@
+require 'package'
+
+class Gpart < Package
+  description 'Partition table rescue/guessing tool. https://github.com/baruch/gpart'
```
uberhacker:
```
- description 'Partition table rescue/guessing tool. https://github.com/baruch/gpart'
+ description 'Partition table rescue/guessing tool.'
```

uberhacker:
Thank you for the new packages @beliefless-dogmasless!  Please see inline comments.

beliefless-dogmasless:
The 'only problem' with that method, is rarely package have both information(on the same page:downloads and information about the package). So, if I keep, one in the description(information about the package), and one in the homepage(downloads, it leaves 'people' with two information, and the advantage to check relatively rapidly the version, and 'make them' compile/and propose new version.

uberhacker:
Rather than combining the homepage url in with the description, why not separate it out in another field?  Should we propose a new metadata value?  You can also get the url that you have stored in homepage from the source_url by removing the tarball from the end.

beliefless-dogmasless:
<a href="http://fr.tinypic.com?ref=21llgnb" target="_blank">Image and video hosting by TinyPic</a>
It's 'neat' to have the address directly. Then, if people want to collaborate, they don't have to search 'a lot of time' to find it.

uberhacker:
What is wrong with `crew search gpart -v`?

beliefless-dogmasless:
It's perfect, no 'they' 'have all' the information they need to participate.
<a href="http://fr.tinypic.com?ref=258v1bl" target="_blank">Image and video hosting by TinyPic</a>

PR #1930, opened 2018-03-08
Add Pavuk Package

beliefless-dogmasless:
Hy,

It works(compile) well on arm.

### DIFF
--- a/packages/pavuk.rb
+++ b/packages/pavuk.rb
@@ -1,0 +1,24 @@
+require 'package'
+
+class Pavuk < Package
+  description 'Web spider and website mirroring tool. http://www.pavuk.org/'
+  homepage 'https://sourceforge.net/projects/pavuk/files/pavuk/'
+  version '0.9.35'
+  source_url 'https://sourceforge.net/projects/pavuk/files/pavuk/0.9.35/pavuk-0.9.35.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/pavuk.rb
```diff
@@ -0,0 +1,24 @@
+require 'package'
+
+class Pavuk < Package
+  description 'Web spider and website mirroring tool. http://www.pavuk.org/'
+  homepage 'https://sourceforge.net/projects/pavuk/files/pavuk/'
```
uberhacker: Please do not include the homepage url in the description.  If you want to propose that we add a new field to the Package class, then please create a new issue so we can have a discussion.  The homepage url you have there also points at the source, not the homepage of the project.

uberhacker:
Thank you for the new package @beliefless-dogmasless.  Please see inline comment.

packages/pavuk.rb
beliefless-dogmasless: I don't want to 'spend' 'more time' on talking, I will contribute, but remove one link(in the section, the direct link to download). Do, what you want with rest I'm not interested(don't take it personally, it is not).

uberhacker:
If you need the homepage and version of the package, you can execute `crew search <package> -v`.

PR #1936, opened 2018-03-09
Update gdk_pixbuf.rb

### LINKED ISSUE #1935: Gdk_pixbuf - Post Installation Problem
Hy,

When I Compile It, it installs the 'libraries', but can't finish the 'installation'.
So I removed those lines(from the rb file):

 def self.postinstall
    # If the package is installed using a "DESTDIR" method, an important file was not installed and should be copied or generated.
    system "sudo gdk-pixbuf-query-loaders --update-cache"
  end

And It Installs Successfully, 
 then I do a:      sudo gdk-pixbuf-query-loaders --update-cache

lovedheart:
Fixes #1935

### DIFF
--- a/packages/gdk_pixbuf.rb
+++ b/packages/gdk_pixbuf.rb
@@ -3,7 +3,7 @@ require 'package'
 class Gdk_pixbuf < Package
   description 'GdkPixbuf is a library for image loading and manipulation.'
   homepage 'https://developer.gnome.org/gdk-pixbuf'
-  version '2.36.7-1'
+  version '2.36.7-0'
   source_url 'https://ftp.gnome.org/pub/gnome/sources/gdk-pixbuf/2.36/gdk-pixbuf-2.36.7.tar.xz'
   source_sha256 '1b6e5eef09d98f05f383014ecd3503e25dfb03d7e5b5f5904e5a65b049a6a4d8'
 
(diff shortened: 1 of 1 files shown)


packages/gdk_pixbuf.rb
```diff
@@ -39,6 +39,6 @@ def self.install
   
   def self.postinstall
     # If the package is installed using a "DESTDIR" method, an important file was not installed and should be copied or generated.
-    system "sudo gdk-pixbuf-query-loaders --update-cache"
+    puts "please run \"sudo gdk-pixbuf-query-loaders --update-cache\" to generate an important file that was not installed."
```
uberhacker: Is the sudo command necessary?  What happens if you exclude it?

lovedheart: It is necessary. If it is excluded, you can also install it without any errors. But it's highly recommended to run this

uberhacker: What exactly does it install and where?  If it only installs in the `/usr/local` directory, the sudo prefix is not needed.

lovedheart: I don't know it exactly. The only problem is that we use DESTDIR. If use regular destdir, there is no such a problem.

uberhacker: I believe this is what you will want to do:
```
   def self.install
     system "pip install six" # for installation process
+    system "mkdir -p #{CREW_DEST_LIB_PREFIX}/gdk-pixbuf-2.0/2.10.0"
+    system "touch #{CREW_DEST_LIB_PREFIX}/gdk-pixbuf-2.0/2.10.0/loaders.cache"
     system "pip uninstall --yes six"
   end

   def self.postinstall
-    # If the package is installed using a "DESTDIR" method, an important file was not installed and shoul$
-    puts "please run \"sudo gdk-pixbuf-query-loaders --update-cache\" to generate an important file that $
+    system 'gdk-pixbuf-query-loaders --update-cache'
   end
```

uberhacker: Did you get a chance to test this suggestion yet @lovedheart?

lovedheart:
I've changed the code

uberhacker:
Can you bump the version and remove the binaries please?

lovedheart:
@uberhacker  changed it

PR #1938, opened 2018-03-09
Upgrade protobuf to v3.5.1

lovedheart:


uberhacker:
Thank you for the update @lovedheart!  Does this resolve #1926?

lovedheart:
I have not tested x86

uberhacker:
Does it compile without the `system './autogen.sh'` line?  We also don't need the dependencies since they're now part of Chromebrew "core".

lovedheart:
It needs "./autogen.sh"

PR #1941, opened 2018-03-09
[fribidi] Has a wrong source address

beliefless-dogmasless:
Hy,

The Source file is not right.

### DIFF
--- a/packages/fribidi.rb
+++ b/packages/fribidi.rb
@@ -3,22 +3,9 @@ require 'package'
 class Fribidi < Package
   description 'GNU FriBidi is an implementation of the Unicode Bidirectional Algorithm (bidi).'
   homepage 'https://www.fribidi.org/'
-  version '0.19.7-1'
+  version '1.0.1'
   source_url 'https://github.com/fribidi/fribidi/releases/download/v1.0.1/fribidi-1.0.1.tar.bz2'
   source_sha256 'c1b182d70590b6cdb5545bab8149de33b966800f27f2d9365c68917ed5a174e4'
-
-  binary_url ({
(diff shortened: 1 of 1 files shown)


uberhacker:
Thank you for the update @beliefless-dogmasless!  It looks like you have 2 files.  Please sync with the updated master branch locally and push another commit so you only include the `fribidi` package changes.  Also, you will need to update the version to match the source:
```
- version '0.19.7-1'
+ version '1.0.1'
```
Thanks

beliefless-dogmasless:
I haven't change(look to) the binaries, but compiling(source), I noticed the address was dead, had that one instead.

uberhacker:
If you can remove all the binary references also that would be appreciated.

beliefless-dogmasless:
Done.

uberhacker:
I'm still seeing the extra `pavuk` package.  I think this package is already in the master branch.  Please remove and resubmit.

beliefless-dogmasless:
You're right, excuse me, I have not 'updated the master branch'.[I don't even know 'how to do it'].

uberhacker:
Follow these steps:
```
$ cd /path/to/chromebrew/source
$ git checkout master
$ git pull origin master (or crew update)
$ git push origin master
```
I think that should fix it.

beliefless-dogmasless:
Thanks, I will 'try that', the next time:
I deleted the fork, and fork it 'again'.

uberhacker:
If you create a new branch any time you need to make changes and leave the master branch untainted, it will make this process much more manageable.  Don't be afraid to create branches.  That is the power of git.

beliefless-dogmasless:
"If you create a new branch any time you need to make changes and leave the master branch untainted, it will make this process much more manageable. Don't be afraid to create branches. That is the power of git."
Ok, Thanks For the Tip.

PR #1949, opened 2018-03-10
Add core_packages.txt, check_updates.sh,  compare_packages.sh and packages.yaml to the tools directory

uberhacker:


lovedheart:
Should cmake and unzip also be added into "core" ?

uberhacker:
Definitely could make a case.  If you include `unzip`, you should probably include `zip` also.

PR #1952, opened 2018-03-10
Create libinput.rb

lovedheart:
#889  Dependency for weston

### DIFF
--- a/packages/libinput.rb
+++ b/packages/libinput.rb
@@ -7,27 +7,31 @@ class Libinput < Package
   source_url 'https://www.freedesktop.org/software/libinput/libinput-1.10.2.tar.xz'
   source_sha256 '1509766d348efe8c6da4285efad3acff4a4c955defb43309e3e4851849197bb9'
 
-    depends_on 'mtdev'
-    depends_on 'libevdev'
-    depends_on 'libwacom'
-    depends_on 'libunwind'
-    depends_on 'libcheck'    
-    depends_on 'valgrind' => :build
-    depends_on 'meson' => :build
+  depends_on 'mtdev'
+  depends_on 'libevdev'
+  depends_on 'libwacom'
+  depends_on 'libunwind'
+  depends_on 'libcheck'    
+  depends_on 'valgrind' => :build
+  depends_on 'meson' => :build
     
-    # If debug-gui feature is required, uncomment following lines and remove "-Ddebug-gui=false" to enable it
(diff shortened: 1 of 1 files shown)


packages/libinput.rb
```diff
@@ -0,0 +1,33 @@
...
+    
+    # If debug-gui feature is required, uncomment following lines and remove "-Ddebug-gui=false" to enable it
+    #depends_on 'graphviz' => :build
+    #depends_on 'gtk3' => :build
+
+    
+  def self.build
+      system "meson --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} -Ddebug-gui=false -Ddocumentation=false _build"           
+      system "ninja -v -C _build"
+      system "ninja -C _build test"
```
uberhacker: This line could go in the `def self.check` section:
```
def self.check
  system 'ninja -C _build test'
end
```

lovedheart: I think this is recommended to be under test before install.

packages/libinput.rb
```diff
@@ -0,0 +1,33 @@
...
+    depends_on 'libevdev'
+    depends_on 'libwacom'
+    depends_on 'libunwind'
+    depends_on 'libcheck'    
+    depends_on 'valgrind' => :build
+    depends_on 'meson' => :build
+    
+    # If debug-gui feature is required, uncomment following lines and remove "-Ddebug-gui=false" to enable it
+    #depends_on 'graphviz' => :build
+    #depends_on 'gtk3' => :build
```
uberhacker: Please indent using only 2 spaces instead of 4 for the `depends_on ...` lines.

packages/libinput.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    
+  def self.build
+      system "meson --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} -Ddebug-gui=false -Ddocumentation=false _build"
```
uberhacker: Since this line is so long, it would be more readable if you split each option into separate lines:
```
  system "meson \
                --prefix=#{CREW_PREFIX} \
                --libdir=#{CREW_LIB_PREFIX} \
                -Ddebug-gui=false \
                -Ddocumentation=false \
                _build"
```

uberhacker:
Thank you for the new package @lovedheart!  Please see inline comments.

lovedheart:
#1585

PR #1954, opened 2018-03-10
Upgrade wayland_protocols to 1.13

lovedheart:
Required for weston build

#889

### DIFF
--- a/packages/wayland_protocols.rb
+++ b/packages/wayland_protocols.rb
@@ -17,7 +17,7 @@ class Wayland_protocols < Package
   depends_on 'libwayland'
   
   def self.build
-    system "./configure"
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
   end
 
   def self.install
--- a/packages/wayland_protocols.rb
+++ b/packages/wayland_protocols.rb
@@ -18,6 +18,7 @@ class Wayland_protocols < Package
   
   def self.build
     system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
+    system "make"
   end
 
   def self.install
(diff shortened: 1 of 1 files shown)


packages/wayland_protocols.rb
```diff
@@ -3,28 +3,21 @@
...
-       i686: '11d26d2ca2b5e960b71e44573ebeba6c601e3e9e3d0e4614972c66cd8316145e',
-     x86_64: '5c1f8913e2ff2ee59dc8080b8c076383caf693890b20bbf08113d08f7fb6199e',
+
   })
 
   depends_on 'libwayland'
   
   def self.build
     system "./configure"
-    system "make"
```
uberhacker: Why was this line removed?

packages/wayland_protocols.rb
```diff
@@ -3,28 +3,21 @@
...
-     armv7l: 'b1c7bf17d30ff2335dafdbd2c57bb8769bfe54f8a2342d70af06fbf0a2adc51f',
-       i686: '11d26d2ca2b5e960b71e44573ebeba6c601e3e9e3d0e4614972c66cd8316145e',
-     x86_64: '5c1f8913e2ff2ee59dc8080b8c076383caf693890b20bbf08113d08f7fb6199e',
+
   })
 
   depends_on 'libwayland'
   
   def self.build
     system "./configure"
```
uberhacker: Don't we need `--prefix=#{CREW_PREFIX}` and `--libdir=#{CREW_LIB_PREFIX}` configure options?

lovedheart: It works without it. But I can add it.

uberhacker:
Thank you for the new package @lovedheart!  Please see inline comments.

uberhacker:
Most packages will install fine simply with `make install` but it is a "best practice" if you run `make` prior to install.

PR #1968, opened 2018-03-11
Create xkbcomp.rb

lovedheart:
#889

### DIFF
--- a/packages/xkbcomp.rb
+++ b/packages/xkbcomp.rb
@@ -1,0 +1,27 @@
+require 'package'
+
+class Xkbcomp < Package
+  description 'Compile XKB keyboard'
+  homepage 'https://www.x.org'
+  version '1.4.1'
+  source_url 'https://www.x.org/archive//individual/app/xkbcomp-1.4.1.tar.gz'
+  source_sha256 '9cdae907b6e34f1f325f6311020126157037925877f323a880e0c9ecab9628f6'
+
+
+  depends_on 'mesa'
+  depends_on 'xcb_util'
+
+    
+  def self.build
+    system "./configure \
+            --prefix=#{CREW_PREFIX} \
+            --libdir=#{CREW_LIB_PREFIX} \ 
+            --with-xkb-config-root=#{CREW_PREFIX}/X11/xkb "
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "mkdir -p #{CREW_PREFIX}/X11/xkb"
+  end
+end
--- a/packages/xkbcomp.rb
+++ b/packages/xkbcomp.rb
@@ -22,7 +22,7 @@ class Xkbcomp < Package
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
-    system "mkdir -p #{CREW_LIB_PREFIX}/X11/xkb"
+    system "mkdir -p #{CREW_DEST_LIB_PREFIX}/X11/xkb"
   end
   
   def self.postinstall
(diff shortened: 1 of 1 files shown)


packages/xkbcomp.rb
```diff
@@ -0,0 +1,27 @@
...
+    system "./configure \
+            --prefix=#{CREW_PREFIX} \
+            --libdir=#{CREW_LIB_PREFIX} \ 
+            --with-xkb-config-root=#{CREW_PREFIX}/X11/xkb "
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "mkdir -p #{CREW_PREFIX}/X11/xkb"
```
uberhacker: What goes in this directory?  If you use `#{CREW_DEST_PREFIX}/X11/xkb` instead of `#{CREW_PREFIX}/X11/xkb`, this directory can be tracked from `crew files xkbcomp`.

uberhacker: If there is any post installation configuration necessary, please message that in the `def self.postinstall` section.

packages/xkbcomp.rb
```diff
@@ -0,0 +1,30 @@
...
+            --with-xkb-config-root=#{CREW_LIB_PREFIX}/X11/xkb "
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+  
+  def self.postinstall
+    system "mkdir -p #{CREW_LIB_PREFIX}/X11/xkb"
```
uberhacker: Actually, I meant this line can stay in the `def self.install` section but if you needed any messages about the configuration that would be needed to populate this directory, those can go in the `def self.postinstall` section.  Sorry about the confusion.  Just so you know, the `def self.postinstall` section lines are executed regardless of whether you install via source or binary, however, the `def self.install` section is only executed during the source install.  Hopefully, that makes sense.

packages/xkbcomp.rb
```diff
@@ -0,0 +1,31 @@
...
+    system "./configure \
+            --prefix=#{CREW_PREFIX} \
+            --libdir=#{CREW_LIB_PREFIX} \ 
+            --with-xkb-config-root=#{CREW_LIB_PREFIX}/X11/xkb "
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "mkdir -p #{CREW_LIB_PREFIX}/X11/xkb"
```
uberhacker: If you want this directory tracked for installation, use `system "mkdir -p #{CREW_DEST_LIB_PREFIX}/X11/xkb"` instead.

packages/xkbcomp.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+  
+  def self.postinstall
+    puts "xkb configuration files are located in #{CREW_LIB_PREFIX}/X11/xkb"
```
uberhacker: I usually add some color to messages like this so they stand out during installation.
```
puts "xkb configuration files are located in #{CREW_LIB_PREFIX}/X11/xkb".lightblue
```

packages/xkbcomp.rb
uberhacker: Do you understand the reason this should be `#{CREW_DEST_LIB_PREFIX}/X11/xkb` instead of `#CREW_LIB_PREFIX}/X11/xkb`?

lovedheart: What ist that reason?

uberhacker: If you don't use that directory, `crew files xkbcomp` will not include the `#{CREW_LIB_PREFIX}/X11/xkb` directory.

lovedheart: OK. I got it.

PR #1970, opened 2018-03-11
[graphene] Upgrade to 1.8.0,  remove corrupted binaries and modify the compilation code

### LINKED ISSUE #1969: graphene binary corrupted
chronos@localhost ~/Downloads $ crew files graphene
(i) graphene: A thin layer of graphic data types
https://github.com/ebassi/graphene
version 1.6.0
Total found: 0

lovedheart:
Upgrade version 1.8.0 and fix #1969  

I think all packages that require graphene should be recomplied.

### DIFF
--- a/packages/graphene.rb
+++ b/packages/graphene.rb
@@ -3,39 +3,29 @@ require 'package'
 class Graphene < Package
   description 'A thin layer of graphic data types'
   homepage 'https://github.com/ebassi/graphene'
-  version '1.6.0'
-  source_url 'https://github.com/ebassi/graphene/archive/1.6.0.tar.gz'
-  source_sha256 '98970f859e452ce421b72726ca727fdf3ac27cb4804b62bfe520157fa46aa2fd'
+  version '1.8.0'
+  source_url 'https://github.com/ebassi/graphene/archive/1.8.0.tar.gz'
+  source_sha256 '410f2e848952cc5830f39b6f6ea7f9b0a487cfe99dad86eec6f22ccbb3ec635b'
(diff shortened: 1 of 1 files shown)


packages/graphene.rb
```diff
@@ -3,39 +3,29 @@
...
-    end
-  end
-  
-  def self.check
-    Dir.chdir("_build") do
-     system "ninja test"
-    end
-  end  
+      system "DESTDIR=#{CREW_DEST_DIR} ninja -C _build install"
+  end
```
uberhacker: Did you set your tab key to 4 spaces instead of 2? :)

packages/graphene.rb
```diff
@@ -3,39 +3,28 @@
...
   })
 
   depends_on 'meson'
+  depends_on 'gobject_introspection' => :build
 
   def self.build
-    system "meson _build --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} -Denable-arm-neon=false"
+    system "meson --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} -Darm_neon=false _build"           
+    system "ninja -v -C _build"
+    system "ninja -C _build test"
```
uberhacker: This should probably still be included in the `def self.check` section but it's more semantic than incorrect.

lovedheart: If i put it there, it is against meson build procedure.

uberhacker: No worries, I already merged.

Issue #1981, opened 2018-03-11
Recompile all packages that have wrong libdir settings


lovedheart:
Due to some packages that have wrong libdr settings befor, it faces some problems for buildling packages for x86_64.

uberhacker:
I like it!

PR #2021, opened 2018-03-11
Update xproto.rb

lovedheart:
#1981


packages/xproto.rb
```diff
@@ -3,25 +3,19 @@
 class Xproto < Package
   description 'The protocols for the X window system provide extended functionality for communication between a X client and the server.'
   homepage 'https://x.org'
-  version '7.0.31'
+  version '7.0.3-0'
```
uberhacker: version '7.0.31-0'?

PR #2024, opened 2018-03-11
Update libxcb.rb

lovedheart:
#1981

### DIFF
--- a/packages/libxcb.rb
+++ b/packages/libxcb.rb
@@ -3,7 +3,7 @@ require 'package'
 class Libxcb < Package
   description 'library for the X window system'
   homepage 'https://x.org'
-  version '1.12'
+  version '1.12-0'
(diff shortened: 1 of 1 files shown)


packages/libxcb.rb
```diff
@@ -26,7 +26,7 @@ class Libxcb < Package
   depends_on 'pthread_stubs'
 
   def self.build
-    system "python=/usr/local/bin/python2.7 ./configure"
+    system "python=/usr/local/bin/python2.7 ./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
```
uberhacker: system "python=#{CREW_PREFIX}/bin/python2.7 ... ?

lovedheart: It can be removed.

lovedheart:
Have tested

PR #2045, opened 2018-03-12
[libtirpc] fix libdir

lovedheart:
#1981


uberhacker:
Thank you for the update @lovedheart.  Bump version?

PR #2046, opened 2018-03-12
[dbus] Upgrade to 1.13.2 , fix libdir

lovedheart:
#1981

### DIFF
--- a/packages/dbus.rb
+++ b/packages/dbus.rb
@@ -20,7 +20,7 @@ class Dbus < Package
     system "./configure \
     --prefix=#{CREW_PREFIX} \
     --libdir=#{CREW_LIB_PREFIX} \
-    --localstatedir=#{CREW_DEST_DIR}/var \
+    --localstatedir=#{CREW_DEST_PREFIX}/var \
     --disable-static \
     --disable-doxygen-docs \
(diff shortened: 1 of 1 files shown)


packages/dbus.rb
```diff
@@ -3,27 +3,27 @@
...
   })
 
   depends_on 'expat'
 
   def self.build
-    system "./configure"
+    system "./configure \
+    --prefix=#{CREW_PREFIX} \
+    --libdir=#{CREW_LIB_PREFIX} \
+    --localstatedir=#{CREW_DEST_PREFIX}/var \
```
uberhacker: Not sure if you are aware but this `--localstatedir` directory is read-only on CloudReady.  You might notice an error during installation.  This is usually a good indicator of the problem.

uberhacker:
Thank you for the update @lovedheart.  Please see inline comment.

packages/dbus.rb
lovedheart: @uberhacker Remove --localstatedir ?

uberhacker: You could remove or just change to a writable directory like `#{CREW_DEST_PREFIX}/tmp` instead.

PR #2047, opened 2018-03-12
[c_ares] Fix libdir

lovedheart:
#1981


uberhacker:
Thank you for the update @lovedheart.  Can you please bump the version?

PR #2062, opened 2018-03-13
[libarchive] fix libdir

lovedheart:
#1981


uberhacker:
Thank you for the update @lovedheart!  There is an update available for this library.  See http://www.libarchive.org/downloads/libarchive-3.3.2.tar.gz.

PR #2063, opened 2018-03-13
[libass] fix libdir

lovedheart:
#1981

### DIFF
--- a/packages/libass.rb
+++ b/packages/libass.rb
@@ -7,9 +7,9 @@ class Libass < Package
   source_url 'https://github.com/libass/libass/releases/download/0.14.0/libass-0.14.0.tar.xz'
   source_sha256 '881f2382af48aead75b7a0e02e65d88c5ebd369fe46bc77d9270a94aa8fd38a2'
 
-  depends_on 'automake' => :build
-  depends_on 'autoconf' => :build
-  depends_on 'libtool' => :build
+  depends_on 'automake'
+  depends_on 'autoconf'
+  depends_on 'libtool'
   depends_on 'fribidi'
   depends_on 'fontconfig'  
   
--- a/packages/libass.rb
+++ b/packages/libass.rb
@@ -3,22 +3,11 @@ require 'package'
 class Libass < Package
   description 'A portable subtitle renderer for the ASS/SSA (Advanced Substation Alpha/Substation Alpha) subtitle format.'
   homepage 'https://github.com/libass/libass'
-  version '0.13.6-1'
+  version '0.13.6-2'
   source_url 'https://github.com/libass/libass/archive/0.13.6.tar.gz'
   source_sha256 'de68288397d284a955061b3ecba67152cea74d8ace6222d987dd30091b1534b7'
 
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/libass-0.13.6-1-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/libass-0.13.6-1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/libass-0.13.6-1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/libass-0.13.6-1-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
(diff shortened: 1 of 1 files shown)


uberhacker:
There is an update available for this library.  See https://github.com/libass/libass/releases/download/0.14.0/libass-0.14.0.tar.xz.

packages/libass.rb
```diff
@@ -3,22 +3,11 @@
...
-  })
-  binary_sha256 ({
-    aarch64: '0b0996d9626e827c7a22b95c7eb59217b657afa2044934884167deaae217d106',
-     armv7l: '0b0996d9626e827c7a22b95c7eb59217b657afa2044934884167deaae217d106',
-       i686: 'd9ed5ea138dda3cff16e20b33534cf3cd7f3cead250cd7c5b74347bf517938fa',
-     x86_64: 'dbc621f3d5fb1d39338fbd2b7e7f1237b12d67f8e46f95a999550680ab0ecbfd',
-  })
 
   depends_on 'automake' => :build
   depends_on 'autoconf' => :build
```
uberhacker: You can remove all the build dependencies since they're already part of Chromebrew "core".

packages/libass.rb
```diff
@@ -3,34 +3,15 @@
...
-     armv7l: '0b0996d9626e827c7a22b95c7eb59217b657afa2044934884167deaae217d106',
-       i686: 'd9ed5ea138dda3cff16e20b33534cf3cd7f3cead250cd7c5b74347bf517938fa',
-     x86_64: 'dbc621f3d5fb1d39338fbd2b7e7f1237b12d67f8e46f95a999550680ab0ecbfd',
-  })
-
-  depends_on 'automake' => :build
-  depends_on 'autoconf' => :build
-  depends_on 'freetype'
-  depends_on 'fribidi'
-  depends_on 'fontconfig'
```
uberhacker: We probably want to leave the non-build dependencies, correct?

lovedheart: I would like uncomment it. I think it's better.

lovedheart: I think we don't change any dependency-related config.

packages/libass.rb
```diff
@@ -3,34 +3,21 @@
...
-     x86_64: 'dbc621f3d5fb1d39338fbd2b7e7f1237b12d67f8e46f95a999550680ab0ecbfd',
-  })
+  version '0.14.0'
+  source_url 'https://github.com/libass/libass/releases/download/0.14.0/libass-0.14.0.tar.xz'
+  source_sha256 '881f2382af48aead75b7a0e02e65d88c5ebd369fe46bc77d9270a94aa8fd38a2'
 
   depends_on 'automake' => :build
   depends_on 'autoconf' => :build
+  depends_on 'libtool' => :build
   depends_on 'freetype'
```
uberhacker: `freetype` is already a dependency of `fontconfig` so it can excluded.

packages/libass.rb
```diff
@@ -3,34 +3,21 @@
...
-       i686: 'd9ed5ea138dda3cff16e20b33534cf3cd7f3cead250cd7c5b74347bf517938fa',
-     x86_64: 'dbc621f3d5fb1d39338fbd2b7e7f1237b12d67f8e46f95a999550680ab0ecbfd',
-  })
+  version '0.14.0'
+  source_url 'https://github.com/libass/libass/releases/download/0.14.0/libass-0.14.0.tar.xz'
+  source_sha256 '881f2382af48aead75b7a0e02e65d88c5ebd369fe46bc77d9270a94aa8fd38a2'
 
   depends_on 'automake' => :build
   depends_on 'autoconf' => :build
+  depends_on 'libtool' => :build
```
uberhacker: We can remove all the `:build` dependencies, correct?

packages/libass.rb
```diff
@@ -3,34 +3,20 @@
...
-       i686: 'd9ed5ea138dda3cff16e20b33534cf3cd7f3cead250cd7c5b74347bf517938fa',
-     x86_64: 'dbc621f3d5fb1d39338fbd2b7e7f1237b12d67f8e46f95a999550680ab0ecbfd',
-  })
-
-  depends_on 'automake' => :build
-  depends_on 'autoconf' => :build
-  depends_on 'freetype'
+  depends_on 'automake'
+  depends_on 'autoconf'
+  depends_on 'libtool'
```
uberhacker: No, I mean these lines can be removed completely since they're already part of Chromebrew "core".  Sorry for the confusion. :)

PR #2064, opened 2018-03-13
Update libedit.rb

lovedheart:
#1981

### DIFF
--- a/packages/libedit.rb
+++ b/packages/libedit.rb
@@ -3,7 +3,7 @@ require 'package'
 class Libedit < Package
   description 'This is an autotool and libtoolized port of the NetBSD Editline library (libedit).'
   homepage 'http://thrysoee.dk/editline/'
-  version '3.1'
+  version '3.1-0'
   source_url 'https://thrysoee.dk/editline/libedit-20170329-3.1.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/libedit.rb
```diff
@@ -4,24 +4,15 @@ class Libedit < Package
   description 'This is an autotool and libtoolized port of the NetBSD Editline library (libedit).'
   homepage 'http://thrysoee.dk/editline/'
   version '3.1'
-  source_url 'http://thrysoee.dk/editline/libedit-20160903-3.1.tar.gz'
-  source_sha256 '0ccbd2e7d46097f136fcb1aaa0d5bc24e23bb73f57d25bee5a852a683eaa7567'
+  source_url 'http://thrysoee.dk/editline/libedit-20170329-3.1.tar.gz'
```
uberhacker: Can we change to use https instead?

uberhacker: Also, let's make the version `20170329-3.1` so it can be bumped.

packages/libedit.rb
```diff
@@ -4,24 +4,15 @@ class Libedit < Package
   description 'This is an autotool and libtoolized port of the NetBSD Editline library (libedit).'
   homepage 'http://thrysoee.dk/editline/'
   version '3.1'
```
uberhacker: Can we make the version `20170329-3.1` so it can be bumped?

PR #2066, opened 2018-03-13
Update libnl3.rb

lovedheart:
#1981

### DIFF
--- a/packages/libnl3.rb
+++ b/packages/libnl3.rb
@@ -11,11 +11,11 @@ class Libnl3 < Package
 
   def self.build
 	system "./configure \
-			    --prefix=#{CREW_PREFIX} \
-			    --libdir=#{CREW_LIB_PREFIX} \
-          --sysconfdir=#{CREW_DEST_PREFIX}/etc \
-          --sbindir=#{CREW_DEST_PREFIX}/sbin \
-          --disable-static"
+		--prefix=#{CREW_PREFIX} \
+		--libdir=#{CREW_LIB_PREFIX} \
+                --sysconfdir=#{CREW_PREFIX}/etc \
+                --sbindir=#{CREW_PREFIX}/sbin \
+                --disable-static"
     system "make"
   end
 
--- a/packages/libnl3.rb
+++ b/packages/libnl3.rb
@@ -7,7 +7,6 @@ class Libnl3 < Package
   source_url 'http://www.infradead.org/~tgr/libnl/files/libnl-3.2.25.tar.gz'
   source_sha256 '8beb7590674957b931de6b7f81c530b85dc7c1ad8fbda015398bc1e8d1ce8ec5'
 
-  depends_on "buildessential" => :build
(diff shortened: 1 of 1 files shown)


packages/libnl3.rb
```diff
@@ -3,32 +3,20 @@
...
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/libnl3-3.2.25-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: '1eeb2dc6b831749cfdb466fe7daafed0bd4621b799782237ad972a2f46bf6566',
-     armv7l: '1eeb2dc6b831749cfdb466fe7daafed0bd4621b799782237ad972a2f46bf6566',
-       i686: '466eadbbc822bc03308d5f7881510de0cc61e5bfa7647b9600898018168a6d44',
-     x86_64: 'f29506472cccad26547d20afa519ed4823e2898a1822b43e7bc4a41f581d78c7',
-  })
-
   depends_on "buildessential" => :build
```
uberhacker: This dependency can be removed since it's now part of Chromebrew "core".

packages/libnl3.rb
```diff
@@ -3,32 +3,20 @@
...
-    system "./configure",
-      "--prefix=/usr/local",
-      "--sysconfdir=/usr/local/etc",
-      "--sbindir=/usr/local/sbin",
-      "--disable-static"
+	system "./configure \
+			    --prefix=#{CREW_PREFIX} \
+			    --libdir=#{CREW_LIB_PREFIX} \
+          --sysconfdir=#{CREW_DEST_PREFIX}/etc \
+          --sbindir=#{CREW_DEST_PREFIX}/sbin \
```
uberhacker: These should be `#{CREW_PREFIX}` instead of `#{CREW_DEST_PREFIX}`.

packages/libnl3.rb
```diff
@@ -3,32 +3,19 @@
...
-  })
-  binary_sha256 ({
-    aarch64: '1eeb2dc6b831749cfdb466fe7daafed0bd4621b799782237ad972a2f46bf6566',
-     armv7l: '1eeb2dc6b831749cfdb466fe7daafed0bd4621b799782237ad972a2f46bf6566',
-       i686: '466eadbbc822bc03308d5f7881510de0cc61e5bfa7647b9600898018168a6d44',
-     x86_64: 'f29506472cccad26547d20afa519ed4823e2898a1822b43e7bc4a41f581d78c7',
-  })
-
-  depends_on "buildessential" => :build
   depends_on "glibc"
```
uberhacker: This is part of Chromebrew "core" also and can be removed.

lovedheart: glibc also in core?

uberhacker: Yes, see https://github.com/skycocker/chromebrew/pull/1949/files.

PR #2071, opened 2018-03-13
Update libtheora.rb

lovedheart:
#1981


packages/libtheora.rb
```diff
@@ -3,29 +3,17 @@
 class Libtheora < Package
   description 'Theora is a free and open video compression format from the Xiph.org Foundation.'
   homepage 'https://theora.org/'
-  version '1.1.1'
+  version '1.1.1-0'
   source_url 'http://downloads.xiph.org/releases/theora/libtheora-1.1.1.tar.bz2'
```
uberhacker: Can we change to use https instead?

PR #2075, opened 2018-03-13
Update libxvid.rb

lovedheart:
#1981


uberhacker:
Update available.  See https://downloads.xvid.com/downloads/xvidcore-1.3.5.tar.bz2.

PR #2080, opened 2018-03-14
Create freetype_sub.rb

lovedheart:
To handle circular dependency between harfbuzz and freetype

### DIFF
--- a/packages/freetype_sub.rb
+++ b/packages/freetype_sub.rb
@@ -1,0 +1,32 @@
+require 'package'
+
+class Freetype_sub < Package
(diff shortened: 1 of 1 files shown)


packages/freetype_sub.rb
```diff
@@ -0,0 +1,32 @@
...
+  depends_on 'libpng'   # freetype needs zlib optionally. zlib is also the dependency of libpng
+  depends_on 'bz2'
+  
+  def self.build
+      system "./configure \
+              CFLAGS=\" -fPIC\" \
+              --prefix=#{CREW_PREFIX} \
+              --libdir=#{CREW_LIB_PREFIX} \
+              --without-harfbuzz"
+      system "make"
```
uberhacker: Can we please use 2 space instead of 4 space indentation here?

PR #2083, opened 2018-03-14
Update heroku.rb

skiman6010:


uberhacker:
Thank you for the update @skiman6010!  This version should be good another day or so. LoL

skiman6010:
@uberhacker Maybe two hours if we're lucky!!

PR #2096, opened 2018-03-17
Add ansible package

pgdagenais:
## Description
Add the ansible package. Installed with pip. 
See source: https://github.com/ansible/ansible

Works properly:
- [x] x86_64

### DIFF
--- a/packages/ansible.rb
+++ b/packages/ansible.rb
@@ -4,7 +4,7 @@ class Ansible < Package
   description 'Ansible is a radically simple IT automation engine that automates cloud provisioning, configuration management, application deployment, intra-service orchestration, and many other IT needs.'
   homepage 'https://www.ansible.com/'
   version '2.4.3'
-  source_url 'http://releases.ansible.com/ansible/ansible-2.4.3.0.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/ansible.rb
```diff
@@ -0,0 +1,17 @@
+require 'package'
+
+class Ansible < Package
+  description 'Ansible is a radically simple IT automation engine that automates cloud provisioning, configuration management, application deployment, intra-service orchestration, and many other IT needs.'
+  homepage 'https://www.ansible.com/'
+  version '2.4.3'
+  source_url 'http://releases.ansible.com/ansible/ansible-2.4.3.0.tar.gz'
```
uberhacker: Can we use https://releases.ansible.com/ansible/ansible-2.4.3.0.tar.gz? (https instead of http)

uberhacker:
Thank you for the new package @pgdagenais!  Excellent work.  Can we please use https instead of http for the source_url?

packages/ansible.rb
pgdagenais: Of course we can ! Let me change that

PR #2099, opened 2018-03-17
Add bash_completion postinstall method

pgdagenais:
The post installation action was not printed by crew after the installation of bash_completion.

I moved them to the `postinstall` method


uberhacker:
@pgdagenais: Nice!  Since you are making changes, why not update?  The latest version is https://github.com/scop/bash-completion/releases/download/2.8/bash-completion-2.8.tar.xz.

PR #2102, opened 2018-03-17
Update pthread_stubs.rb

lovedheart:


skiman6010:
Thanks for the pull request! Can you please delete the extra lines between ` source_sha256` and `def self.build`?

skiman6010:
Thanks so much @lovedheart!

Issue #2108, opened 2018-03-18
Can not find .la files by compiling packages on x86_64


lovedheart:
The compile error "/usr/local/lib/foo.la is not a valid libtool archieve" or "can not find /usr/local/lib/foo.la" is  a very common problem when building packages on x86_64. It seems that those *.la files located in /usr/local/lib64 can be deleted to resolve the compile errors.

see [https://gnunet.org/faq-la-files](https://gnunet.org/faq-la-files)
[https://blog.flameeyes.eu/2009/09/removing-la-files-for-dum-w-uncertain-people/](https://blog.flameeyes.eu/2009/09/removing-la-files-for-dum-w-uncertain-people/)

#2105

uberhacker:
Fixed in #2110.

PR #2109, opened 2018-03-18
Update package.rb

lovedheart:
add command to remove any .la files in /usr/local/lib64/ before starting compile the packages (x86_64 only)

#2108

### DIFF
--- a/lib/package.rb
+++ b/lib/package.rb
@@ -69,6 +69,13 @@ class Package
     @is_fake
   end
 
+  # Remove /usr/local/lib64/*.la to resolve any "can not find /usr/local/*.la" compile errors when building the packages (only needed for x86_64).
+  # https://gnunet.org/faq-la-files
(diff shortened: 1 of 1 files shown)


lib/package.rb
```diff
@@ -69,6 +69,13 @@ def self.is_fake?
     @is_fake
   end
 
+  # Remove /usr/local/lib64/*.la to resolve any "can not find /usr/local/*.la" compile errors when building the packages (only needed for x86_64).
+  # https://gnunet.org/faq-la-files
+  # https://stackoverflow.com/questions/42963653/libquadmath-la-is-not-a-valid-libtool-archive-when-configuring-openmpi-with-g
+  def self.remove_la_files
+    system "rm -rf #{CREW_LIB_PREFIX}/*.la"
+  end
```
uberhacker: This is not the right location for this logic.  I think you should be able to accomplish the same thing in `crew` instead.

PR #2115, opened 2018-03-18
[cairo] add pre-built packages

lovedheart:


skiman6010:
@uberhacker will be happy to see someone else working on binary packages!

uberhacker:
Yes, my energy level was definitely decreasing in that task. :)

uberhacker:
@lovedheart: You can find out which packages need binaries if you cd into the tools directory:
```
$ bash needs_binaries.sh
```
Some packages are better off without binaries and some don't even work at all with binaries just so you know.  You can ignore fake packages.

Issue #1802, opened 2018-02-09
[gusb] Unable to build


uberhacker:
`i686` error:
```
./configure: line 2969: syntax error near unexpected token `1.29'
./configure: line 2969: `GOBJECT_INTROSPECTION_CHECK(1.29)'
gusb failed to build: `./autogen.sh` exited with 2
```

uberhacker:
It appears the issue on `x86_64` is gusb is looking in the wrong directory for libudev.la.  The correct directory is `/usr/local/lib64/libudev.la`.

lovedheart:
It seems like to need gobject_introspection to build it.

PR #2117, opened 2018-03-18
Update gusb.rb

lovedheart:
Fix dependency

### DIFF
--- a/packages/gusb.rb
+++ b/packages/gusb.rb
@@ -3,16 +3,13 @@ require 'package'
 class Gusb < Package
   description 'GUsb is a GObject wrapper for libusb1'
   homepage 'https://www.openhub.net/p/gusb'
-  version '0.2.11'
+  version '0.2.11-0'
   source_url 'https://github.com/hughsie/libgusb/archive/gusb_0_2_11.tar.gz'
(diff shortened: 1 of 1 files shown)


uberhacker:
Does this fix #1802?

packages/gusb.rb
```diff
@@ -3,16 +3,13 @@
 class Gusb < Package
   description 'GUsb is a GObject wrapper for libusb1'
   homepage 'https://www.openhub.net/p/gusb'
-  version '0.2.11'
+  version '0.2.11-0'
   source_url 'https://github.com/hughsie/libgusb/archive/gusb_0_2_11.tar.gz'
```
uberhacker: Update available?  https://github.com/hughsie/libgusb/archive/0.3.0.tar.gz

lovedheart: @uberhacker  No. I won't update it. I have tried that and get the errors, which I can not fix right now.

uberhacker: No worries.  It looks more like a beta release anyway.

uberhacker:
@lovedheart: Did you plan on adding binaries to this PR or do you want me to merge it and provide later?

lovedheart:
merge it right now

PR #2124, opened 2018-03-19
Update gdk_pixbuf.rb

lovedheart:
fix loaders.cache

### DIFF
--- a/packages/gdk_pixbuf.rb
+++ b/packages/gdk_pixbuf.rb
@@ -17,14 +17,14 @@ class Gdk_pixbuf < Package
   def self.build
     system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --with-x11"
     system "make"
-    # generate cache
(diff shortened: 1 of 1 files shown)


packages/gdk_pixbuf.rb
```diff
@@ -31,17 +17,17 @@ class Gdk_pixbuf < Package
   def self.build
     system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --with-x11"
     system "make"
+    # generate cache
+    system "mkdir -p #{CREW_LIB_PREFIX}/gdk-pixbuf-2.0/2.10.0/"
+    system "sudo #{CREW_DEST_PREFIX}/bin/gdk-pixbuf-query-loaders --update-cache"
```
uberhacker: Have you tried without `sudo`?  I think it should work since that directory is writable by `chronos` user.

lovedheart:
Some problems for i686 and armv7l. Close it

PR #2129, opened 2018-03-20
20march-Some Fixes

beliefless-dogmasless:
Hy,

Here are, some 'Fixes'.

Take 'what You Want/what You Need', (no obligation).

### DIFF
--- a/packages/vala.rb
+++ b/packages/vala.rb
@@ -1,8 +1,8 @@
 require 'package'
 
 class Vala < Package
-  description 'Vala is a programming language that aims to bring modern programming language features to GNOME developers. https://wiki.gnome.org/Projects/Vala'
-  homepage 'https://download.gnome.org/sources/vala/'
+  description 'Vala is a programming language that aims to bring modern programming language features to GNOME developers.'
+  homepage 'https://wiki.gnome.org/Projects/Vala'
   version '0.40.0'
   source_url 'https://download.gnome.org/sources/vala/0.40/vala-0.40.0.tar.xz'
   source_sha256 '15888fcb5831917cd67367996407b28fdfc6cd719a30e6a8de38a952a8a48e71'


lovedheart:
Cairo source link should not be changed. Because official link provides really slow download

packages/binutils.rb
```diff
@@ -4,31 +4,19 @@ class Binutils < Package
   description 'The GNU Binutils are a collection of binary tools.'
   homepage 'http://www.gnu.org/software/binutils/'
   version '2.30'
-  source_url 'https://ftpmirror.gnu.org/binutils/binutils-2.30.tar.xz'
+  source_url 'https://ftp.gnu.org/gnu/binutils/binutils-2.30.tar.xz'
```
uberhacker: We want the original url since it optimizes the mirror to download based on locale.

packages/cairo.rb
```diff
@@ -3,23 +3,10 @@
 class Cairo < Package
   description 'Cairo is a 2D graphics library with support for multiple output devices.'
   homepage 'https://www.cairographics.org'
-  version '1.15.10-0'
-  source_url 'https://fossies.org/linux/misc/cairo-1.15.10.tar.xz'
+  version '1.15.10'
+  source_url 'http://cairographics.org/snapshots/cairo-1.15.10.tar.xz'
```
uberhacker: Is the original download faster?  Also, if you make a change like this, the version should be bumped forward, not backward.
`version '1.15.10-1`

packages/dbus.rb
```diff
@@ -7,28 +7,11 @@ class Dbus < Package
   source_url 'https://dbus.freedesktop.org/releases/dbus/dbus-1.13.2.tar.gz'
```
uberhacker: Bump version?

packages/dbus.rb
```diff
@@ -7,28 +7,11 @@ class Dbus < Package
...
 
   def self.build
-    system "./configure \
-    --prefix=#{CREW_PREFIX} \
-    --libdir=#{CREW_LIB_PREFIX} \
-    --disable-static \
-    --disable-doxygen-docs \
-    --disable-xml-docs"
+    system "sh autogen.sh"
+    system "./configure"
```
uberhacker: We want to keep the `--prefix` and `--libdir` configure options, correct?

packages/ffmpeg.rb
```diff
@@ -41,6 +28,7 @@ class Ffmpeg < Package
   depends_on 'vidstab'
 
   def self.build
+    ENV['TMPDIR'] = "#{CREW_PREFIX}/tmp"
```
uberhacker: This line is unneccessary.  The line below uses a completely different temp directory.

packages/pcre.rb
```diff
@@ -7,21 +7,8 @@ class Pcre < Package
   source_url 'https://ftp.pcre.org/pub/pcre/pcre-8.41.tar.bz2'
```
uberhacker: Bump version?

packages/perl.rb
```diff
@@ -7,19 +7,6 @@ class Perl < Package
...
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/perl-5.26.1-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: 'd905b33e42f8825063eebe17a34bc8011ed9e7a88374373187d55f879453ccc4',
-     armv7l: 'd905b33e42f8825063eebe17a34bc8011ed9e7a88374373187d55f879453ccc4',
-       i686: 'a1394c4caafbca33de118521f850fa4162834f70c4bacbca9b576048a420ffb2',
-     x86_64: 'd6f71ac89fae67ce77d097075b52683cc5bb6367f87ad2dc86c383b4024b4186',
-  })
-
   depends_on 'patch' => :build
```
uberhacker: What is changing here?  I only noticed the binaries were removed with no other changes?

packages/poppler.rb
```diff
@@ -38,6 +38,9 @@ def self.build
   end
 
   def self.install
+    system "pip install six"
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "pip uninstall --yes six"
```
uberhacker: Changes like this will require us to bump the version, especially if the old binaries don't work.

packages/python27.rb
```diff
@@ -3,23 +3,10 @@
 class Python27 < Package
   description 'Python is a programming language that lets you work quickly and integrate systems more effectively.'
   homepage 'https://www.python.org/'
-  version '2.7.14-1'
+  version '2.7.14'
```
uberhacker: Version should be bumped forward, not backward.
`version '2.7.14-2'`

packages/python3.rb
```diff
@@ -3,22 +3,9 @@
...
-  })
-  binary_sha256 ({
-    aarch64: '481fa9f5bdff6d43eeea4ac90bbee6ef5054e2960d5f1f141bb24359af5ac25b',
-     armv7l: '481fa9f5bdff6d43eeea4ac90bbee6ef5054e2960d5f1f141bb24359af5ac25b',
-       i686: '8e7ffe80318afee1e3f3614440d30fb5fd4daf2d2d901b55b48c2216e034059d',
-     x86_64: 'cb75bc5794d06951897906b87ec2fedf316d268040290845809517861a9e7bbe',
-  })
+  version '3.7.0b2'
+  source_url 'https://www.python.org/ftp/python/3.7.0/Python-3.7.0b2.tar.xz'
+  source_sha256 '92082de7fafdcdab61a91b908f32b35f13a7aef3c2671c0fa388eb574c3fc882'
```
uberhacker: We probably want to avoid beta versions until they become official releases.  It's a conservative approach but it assures the most stability.  This is a dependency for a lot of other packages so we need to be careful.

uberhacker:
@beliefless-dogmasless: I really appreciate your efforts to update these packages!  Please see the inline comments.  In the future, please add your PRs for a single package update at a time.  This will make it easier to review and will likely get things merged in quicker.  I would suggest you close this PR since it conflicts with an update @lovedheart submitted recently and add your package update PRs separately.

packages/poppler.rb
lovedheart: six is usually required for x86_64. I think

uberhacker:
Now up to 4 conflicts.  Closing due to inactivity.

PR #2133, opened 2018-03-20
Update gtk3.rb

lovedheart:
Need this patch quickly! @uberhacker

### DIFF
--- a/packages/gtk3.rb
+++ b/packages/gtk3.rb
@@ -20,8 +20,6 @@ class Gtk3 < Package
      x86_64: '449303430fc73dfa177ef7880a02b3972cc65efbbda1256c598710330e9a6bf7',
   })
 
-
-
   depends_on 'gdk_pixbuf'
   depends_on 'libepoxy'
   depends_on 'graphene'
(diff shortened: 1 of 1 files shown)


uberhacker:
LoL, this is the exact same fix I proposed in #2131 but after thinking about it awhile, I think there is a better way.  Try this:
```
  def self.install
    system "pip install six"  # for installation process
+    system "mkdir -p #{CREW_DEST_PREFIX}/share/glib-2.0/schemas"
+    system "glib-compile-schemas #{CREW_DEST_PREFIX}/share/glib-2.0/schemas"
    system "pip uninstall --yes six"
  end

- def self.postinstall
-    system "glib-compile-schemas #{CREW_DEST_PREFIX}/share/glib-2.0/schemas"   # generate gschemas.compiled
- end
```

lovedheart:
This is better. I got error by previous fix

uberhacker:
Would it make sense to include system "glib-compile-schemas #{CREW_DEST_PREFIX}/share/glib-2.0/schemas" in the glib package instead?

uberhacker:
We'll need new binaries and the version bumped also, correct?

lovedheart:
doing it. I do not bump the version

uberhacker:
Keep in mind that if somebody has already installed gtk3 and you don't bump the version, they will not have the changes detected.

PR #2147, opened 2018-03-23
Update imagemagick7.rb

jeffmcneill:
Previous SHA was not correct for this file. Now corrected. Sorry for error.


uberhacker:
No worries.  Glad you corrected it!

PR #2149, opened 2018-03-24
Update imagemagick.rb

jeffmcneill:
Updated to most recent versions of ImageMagick 6+7 install urls and checksums


uberhacker:
Thank you for the update @jeffmcneill! What about `imagemagick6.rb`?

Issue #2145, opened 2018-03-23
Add a crew list installed and crew list available option


pgdagenais:
Going back on this closed pull request by @uberhacker https://github.com/skycocker/chromebrew/pull/319

I do think that this option is useful. Maybe with a new argument like `crew list installed` and `crew list available`

It doesn't really fit with `crew search`

If you don't mind @uberhacker I would take your code and modify it to match match what I said.

uberhacker:
@pgdagenais: Sure, go for it!  If you add a new command, make sure you document it with help.  You can probably get a good idea of how to do that from the existing commands.  I was thinking it might also be good if the `crew files` command included the disk space used (aka size) for all the files in the package.

uberhacker:
Fixed in #2151.

PR #2151, opened 2018-03-24
Add list command to crew

pgdagenais:
This PR add the `crew list` command.

`crew list available` list available package that are not installed.
`crew list installed` list installed package.

Big thanks to @uberhacker for his work in a earlier PR that was closed: https://github.com/skycocker/chromebrew/pull/319

See https://github.com/skycocker/chromebrew/issues/2145 issue.

### DIFF
--- a/crew
+++ b/crew
@@ -25,6 +25,7 @@ Usage:
   crew files <name> ...
   crew help [<command>]
   crew install [-k|--keep] [-s|--build-from-source] [-v|--verbose] <name> ...
+  crew list (available|installed)
   crew remove [-v|--verbose] <name> ...
   crew search [-v|--verbose] [<name> ...]
   crew update
@@ -60,7 +61,7 @@ rescue Docopt::Exit => e
     end
     if ARGV[0] != '-h' and ARGV[0] != '--help' then
       puts "Could not understand \"crew #{ARGV.join(' ')}\".".lightred
-      cmds = ["build", "download", "files", "help", "install", "remove", "search", "update", "upgrade", "whatprovides"]
+      cmds = ["build", "download", "files", "help", "install", "list", "remove", "search", "update", "upgrade", "whatprovides"]
       # Looking for similar commands
       if not cmds.include?(ARGV[0]) then
         similar = cmds.select {|word| edit_distance(ARGV[0], word) < 4}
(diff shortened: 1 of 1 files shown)


crew
```diff
@@ -166,6 +189,9 @@ def help (pkgName)
     puts "If `-k` or `--keep` is present, the `CREW_BREW_DIR` (#{CREW_BREW_DIR}) directory will remain."
     puts "If `-s` or `--build-from-source` is present, the package(s) will be compiled instead of installed via binary."
     puts "If `-v` or `--verbose` is present, extra information will be displayed."
+  when "list"
+    puts "List available or installed packages"
+    puts "Usage: crew list (available|installed)"
```
uberhacker: Can we use brackets instead of parenthesis?
```
Usage: crew list [available|installed]
```

crew
```diff
@@ -25,6 +25,7 @@ Usage:
   crew files <name> ...
   crew help [<command>]
   crew install [-k|--keep] [-s|--build-from-source] [-v|--verbose] <name> ...
+  crew list (available|installed)
```
uberhacker: Can we use brackets instead of parenthesis?
```
Usage: crew list [available|installed]
```

uberhacker:
Thank you for the PR @pgdagenais!  What happens if you type just `crew list` without any parameters?  Does it default to list all packages?

crew
pgdagenais: Based on the [docopt](http://docopt.org/) documentation brackets mean optional elements.

Since using `crew list` only doesn't really make sense by itself, I think that having one of the 2 option a requirement is better

uberhacker: Understood but don't we want a default option if no parameters are supplied?  In that case, it would be optional.  Just sayin'.

pgdagenais: It really depends if we want a default option. I think that the command `crew list` doesn't say anything. 

Same thing with the `crew files` command, it doesn't have a default and without an argument the command doesn't make sense

uberhacker: I suppose if we defaulted to list all the packages it would essentially be the same as `crew search` with no parameters.  I'll go ahead and merge this and we can update later if enough users feel it makes sense to provide a default option.

pgdagenais: Sound good for me, thank you very much.

PR #2155, opened 2018-03-25
Add sshrc package

uberhacker:
Bring your .bashrc, .vimrc, etc. with you when you ssh.  See https://github.com/Russell91/sshrc.


skiman6010:
This is awesome!

PR #2156, opened 2018-03-25
Update imagemagick.rb

jeffmcneill:
Point to .gz source file and sha


uberhacker:
Thank you for the update @jeffmcneill, however, we probably want to stick with the `.tar.xz` extension since the compression is a bit more efficient and will result in a faster download.  In any event, I have this covered in #2152.

PR #2169, opened 2018-03-26
Update miniconda3.rb

skiman6010:
Update Miniconda3. I changed it from pulling the latest to pulling a specific version so we won't have any surprise checksum mismatches.

### DIFF
--- a/packages/miniconda3.rb
+++ b/packages/miniconda3.rb
@@ -3,33 +3,20 @@ require 'package'
 class Miniconda3 < Package
   description 'Anaconda is the world\'s most popular Python data science platform.'
   homepage 'https://conda.io/miniconda.html'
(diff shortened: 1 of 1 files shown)


packages/miniconda3.rb
```diff
@@ -3,33 +3,20 @@
 class Miniconda3 < Package
   description 'Anaconda is the world\'s most popular Python data science platform.'
   homepage 'https://conda.io/miniconda.html'
-  version '5.0.1'
+  version '4.4.10'
```
uberhacker: Downgrade?

skiman6010: No I messed up originally and used the version numbers from Anaconda instead of Miniconda because I'm not all there in the head.

Issue #2059, opened 2018-03-13
Building gcc7.3 from source.


nickjmeyer:
## Description
I am attempting to set up a new package to build gcc from source.  I'm trying to build gcc-7.3 and I've named the packet `Gcc7`.  Below are the contents of `Gcc7.rb`.

```
require 'package'

class Gcc7 < Package
  description 'The GNU Compiler Collection includes front ends for C, C++, Objective-C, Fortran, Ada, and Go.'
  homepage 'https://www.gnu.org/software/gcc/'
  version '7.3.0'

  source_url 'https://github.com/gcc-mirror/gcc/archive/gcc-7_3_0-release.zip'

  source_sha256 'd8cb199f0d98bd83e3dac645126722e44137c24cbe5e87e1113dd4270f7530c6'

  depends_on 'buildessential'
  depends_on 'binutils'
  depends_on 'gmp'
  depends_on 'mpfr'
  depends_on 'mpc'
  depends_on 'isl'
  depends_on 'cloog'
  depends_on 'glibc'

  def self.build
    system "./configure", "-v", "--build=x86_64-linux-gnu", "--host=x86_64-linux-gnu", "--target=x86_64-linux-gnu", "--prefix=#{CREW_DEST_DIR}", "--libdir=#{CREW_LIB_PREFIX}", "--enable-checking=release", "--enable-languages=c,c++,fortran", "--disable-multilib", "--program-suffix=-7.3" 
    system "make"
  end

  def self.install
    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install-strip"
  end

  def self.check
    system "make", "check"
  end
end
```

To install it, I copy `Gcc7.rb` to the installed directory:
```shell
chronos@localhost ~/repos/chromebrew $ cp ~/repos/chromebrew/packages/gcc7.rb /usr/local/lib/crew/packages/
```

Then install:
```shell
chronos@localhost ~/repos/chromebrew $ crew install gcc7
```

The download is successful.  The build starts, but ultimately fails.  Here is the tail of the output:
```
gcc -E ../.././gcc/params-options.h | sed 's/^#.*//;/^$/d' > tmp-params.options
/bin/bash ../.././gcc/../move-if-change tmp-all-tree.def all-tree.def
/bin/bash ../.././gcc/../move-if-change tmp-params.options params.options
/bin/bash ../.././gcc/../move-if-change tmp-params.list params.list
echo timestamp > s-alltree
[...]

PR #2173, opened 2018-04-01
Add gcc 7.3 package.

nickjmeyer:
## Description

This PR adds gcc 7.3 as a new package.  It installs from source, so be prepared to wait a while.

Related to issue #2059 

## Addtional information
This is only configured for x86_64.  All I have is a Pixelbook so I don't have access to the other architectures to set it up for those.  

Works properly:
- [x] x86_64
- [ ] aarch64

### DIFF
--- a/packages/gcc7.rb
+++ b/packages/gcc7.rb
@@ -10,28 +10,31 @@ class Gcc7 < Package
   source_sha256 'd8cb199f0d98bd83e3dac645126722e44137c24cbe5e87e1113dd4270f7530c6'
 
   depends_on 'buildessential'
-  depends_on 'binutils'
-  depends_on 'gmp'
-  depends_on 'mpfr'
-  depends_on 'mpc'
-  depends_on 'isl'
-  depends_on 'cloog'
-  depends_on 'glibc'
   depends_on 'gawk'
+  depends_on 'unzip' => :build
 
   def self.build
     system "mkdir", "build"
     Dir.chdir "build" do
-      ENV["LIBRARY_PATH"] = "/usr/local/lib64"
-      system "../configure", "-v", "--build=x86_64-linux-gnu", "--host=x86_64-linux-gnu", "--target=x86_64-linux-gnu", "--libdir=#{CREW_LIB_PREFIX}", "--prefix=#{CREW_PREFIX}", "--enable-checking=release", "--enable-languages=c,c++,fortran", "--disable-multilib", "--disable-libmpx", "--program-suffix=-7.3"
+      ENV["LIBRARY_PATH"] = "#{CREW_LIB_PREFIX}"
+      system "../configure",
+             "-v",
+             "--libdir=#{CREW_LIB_PREFIX}",
+             "--prefix=#{CREW_PREFIX}",
+             "--enable-checking=release",
+             "--enable-languages=c,c++,fortran",
+             "--disable-multilib",
+             "--disable-libmpx",
+             "--program-suffix=-7.3"
       system "make"
     end
   end
 
   def self.install
     Dir.chdir "build" do
-      ENV["LIBRARY_PATH"] = "/usr/local/lib64"
-      system "make", "install-strip"
+      ENV["LIBRARY_PATH"] = "#{CREW_LIB_PREFIX}"
+      system "make",
+             "install-strip"
     end
   end
 end
--- a/packages/gcc7.rb
+++ b/packages/gcc7.rb
@@ -33,6 +33,7 @@ class Gcc7 < Package
     Dir.chdir "build" do
       ENV["LIBRARY_PATH"] = "#{CREW_LIB_PREFIX}"
       system "make",
+             "DESTDIR=#{CREW_DEST_DIR}",
              "install-strip"
     end
   end
--- a/packages/gcc7.rb
+++ b/packages/gcc7.rb
@@ -9,7 +9,6 @@ class Gcc7 < Package
 
   source_sha256 'd8cb199f0d98bd83e3dac645126722e44137c24cbe5e87e1113dd4270f7530c6'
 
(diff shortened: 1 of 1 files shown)


packages/gcc7.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  depends_on 'buildessential'
+  depends_on 'binutils'
+  depends_on 'gmp'
+  depends_on 'mpfr'
+  depends_on 'mpc'
+  depends_on 'isl'
+  depends_on 'cloog'
+  depends_on 'glibc'
+  depends_on 'gawk'
```
uberhacker: `depends_on 'unzip' => :build`?

packages/gcc7.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def self.build
+    system "mkdir", "build"
+    Dir.chdir "build" do
+      ENV["LIBRARY_PATH"] = "/usr/local/lib64"
```
uberhacker: This should definitely be `ENV["LIBRARY_PATH"] = "#{CREW_LIB_PREFIX}"`.  Otherwise, armv7l and i686 will not work.

packages/gcc7.rb
```diff
@@ -0,0 +1,37 @@
...
+      system "../configure", "-v", "--build=x86_64-linux-gnu", "--host=x86_64-linux-gnu", "--target=x86_64-linux-gnu", "--libdir=#{CREW_LIB_PREFIX}", "--prefix=#{CREW_PREFIX}", "--enable-checking=release", "--enable-languages=c,c++,fortran", "--disable-multilib", "--disable-libmpx", "--program-suffix=-7.3"
```
uberhacker: Please put each option on a separate line to avoid horizontal scrolling.
```
../configure", "-v",
"--build=x86_64-linux-gnu",
"--host=x86_64-linux-gnu",
"--target=x86_64-linux-gnu",
"--prefix=#{CREW_PREFIX}",
"--enable-checking=release",
"--enable-languages=c,c++,fortran",
"--disable-multilib",
"--disable-libmpx",
"--program-suffix=-7.3"
```

packages/gcc7.rb
```diff
@@ -0,0 +1,37 @@
...
+      system "make"
+    end
+  end
+
+  def self.install
+    Dir.chdir "build" do
+      ENV["LIBRARY_PATH"] = "/usr/local/lib64"
```
uberhacker: This should definitely be ENV["LIBRARY_PATH"] = "#{CREW_LIB_PREFIX}". Otherwise, armv7l and i686 will not work.

packages/gcc7.rb
uberhacker: This also makes it easier to add/remove lines in the future.

packages/gcc7.rb
nickjmeyer: The `--build`, `--host`, `--target` flags in the configuration kill this too.  Maybe those aren't necessary to specify?  I can try without.

packages/gcc7.rb
nickjmeyer: Ah yeah good catch.  What does the `=>` syntax mean?  This file is the extent of my Ruby knowledge.

packages/gcc7.rb
uberhacker: You could add:
```
"--build=#{ARCH}-linux-gnu",
"--host=#{ARCH}-linux-gnu",
"--target=#{ARCH}-linux-gnu",
```

uberhacker: But excluding may be better, actually.

nickjmeyer: Sounds good.  Ill remove them.

packages/gcc7.rb
uberhacker: It just means it's a build instead of run-time dependency.  You don't need to include any dependencies that are already in Chromebrew "core".  See https://github.com/skycocker/chromebrew/blob/master/tools/core_packages.txt.

uberhacker:
@lovedheart: You also compiled this successfully, correct?  Is there anything you included that is missing here?

nickjmeyer:
I'm re-installing with the changes to make sure it still works.  Noticed one other issue though.  `crew remove gcc7` doesn't actually remove anything.  All the binaries are still there.

lovedheart:
it was sucessful on arm and i686.

packages/gcc7.rb
```diff
@@ -0,0 +1,40 @@
...
+class Gcc7 < Package
+  description 'The GNU Compiler Collection includes front ends for C, C++, Objective-C, Fortran, Ada, and Go.'
+  homepage 'https://www.gnu.org/software/gcc/'
+  version '7.3.0'
+
+  source_url 'https://github.com/gcc-mirror/gcc/archive/gcc-7_3_0-release.zip'
+
+  source_sha256 'd8cb199f0d98bd83e3dac645126722e44137c24cbe5e87e1113dd4270f7530c6'
+
+  depends_on 'buildessential'
```
uberhacker: Actually, you can remove this also.  It's a fake package that includes all the core dependencies anyway.

uberhacker:
@nickjmeyer: What is the output of `crew files gcc7`?  Do you get anything?

packages/gcc7.rb
nickjmeyer: Fixed

uberhacker:
@lovedheart: Is this good to merge then?

lovedheart:
use DESTDIR option to make install

nickjmeyer:
Added `DESTDIR`.  Installing again now to make sure it works.

nickjmeyer:
`crew files gcc7` shows a lot of stuff.  Not sure if it always showed the files or if it was because of the `DESTDIR` argument.

PR #2174, opened 2018-04-02
[gcc7] Add pre-built binary, minor change

lovedheart:
#2059


packages/gcc7.rb
```diff
@@ -4,37 +4,90 @@ class Gcc7 < Package
...
+                 "--target=#{ARCH}-cros-linux-gnu",
+                 "--enable-checking=release",
+                 "--enable-languages=c,c++,fortran",
+                 "--disable-multilib",
+                 "--enable-threads=posix",
+                 "--disable-bootstrap",
+                 "--disable-libmpx",
+                 "--disable-static",
+                 "--program-suffix=-7.3",
+                 "--with-arch-64=x86-64"
```
uberhacker: Should this be x86_64 instead of x86-64?

lovedheart: I think it should be x86-64

PR #2178, opened 2018-04-02
[libsigcplusplus] upgrade to 2.99.10 with binaries

lovedheart:


packages/libsigcplusplus.rb
```diff
@@ -3,33 +3,46 @@
...
   depends_on 'pkgconfig' => :build
   depends_on 'diffutils' => :build
   depends_on 'm4' => :build
+  depends_on 'gcc7' => :build  # c++14 support
+  depends_on 'cmake' => :build
 
   def self.build
-    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
-    system "make"
+    system "mkdir -p build"
```
uberhacker: No need for the `-p` flag since it's a single directly but it doesn't hurt including it.

lovedheart: @uberhacker gcc7.3 should work I think

PR #2180, opened 2018-04-02
[libsigcplusplus] change name to libsigcplusplus3

lovedheart:
Both sigc++-2.0 and sigc++-3.0 are required for building glibmm.
After changing the name, I will recover the libsigcplusplus for sigc++-2.0

### DIFF
--- a/packages/libsigcplusplus3.rb
+++ b/packages/libsigcplusplus3.rb
@@ -1,6 +1,6 @@
 require 'package'
 
-class Libsigcplusplus < Package
+class Libsigcplusplus3 < Package
   description 'libsigc++ implements a typesafe callback system for standard C++.'
   homepage 'http://libsigc.sourceforge.net/'
   version '2.99.10'


uberhacker:
Are there any other packages that depend on this library?  If that is the case, then those dependencies will need updated also.

lovedheart:
libsigcplusplus will remain. Nothing will be changed

uberhacker:
```
glibmm.rb:24:  depends_on 'libsigcplusplus'
cairomm.rb:24:  depends_on 'libsigcplusplus'
```
These packages will need to be updated to reflect the new name, correct?

lovedheart:
@uberhacker Only cairomm should be changed.  Glibmm need both libsigc++2 and libsigc++3

uberhacker:
Will you be taking `depends_on 'libsigcplusplus'` out of glibmm.rb then?

lovedheart:
@uberhacker I upgrade glibmm to 2.56.0. It requires both libsigc++ versions.

uberhacker:
OK, I will wait for that update and then merge all 3 PRs at that time.

lovedheart:
@uberhacker can you merge this ? Then I can recover libsigcplusplus.rb

uberhacker:
If I merge this now then glibmm.rb will not have the update.  I could merge but if anybody decides to compile from glibmm from source, it will fail.  What is the rush?

lovedheart:
I will make a libsigcplusplus right now. It's only possible after this merge?

uberhacker:
Oh, so you want to create a new `libsigcplusplus.rb` and still have the same dependency in `glibmm.rb`?

lovedheart:
right

uberhacker:
Gotcha.

lovedheart:
depends_on 'libsigcplusplus'
depends_on 'libsigcplusplus3'

PR #2182, opened 2018-04-02
Create libsigcplusplus.rb

lovedheart:
#2180  recover libsigc++-2.0

### DIFF
--- a/packages/libsigcplusplus.rb
+++ b/packages/libsigcplusplus.rb
@@ -1,0 +1,35 @@
+require 'package'
+
+class Libsigcplusplus < Package
+  description 'libsigc++ implements a typesafe callback system for standard C++.'
+  homepage 'http://libsigc.sourceforge.net/'
+  version '2.10.0'
+  source_url 'http://ftp.gnome.org/pub/GNOME/sources/libsigc++/2.10/libsigc++-2.10.0.tar.xz'
+  source_sha256 'f843d6346260bfcb4426259e314512b99e296e8ca241d771d21ac64f28298d81'
+
+  binary_url ({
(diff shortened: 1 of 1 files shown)


packages/libsigcplusplus.rb
```diff
@@ -0,0 +1,35 @@
+require 'package'
+
+class Libsigcplusplus < Package
+  description 'libsigc++ implements a typesafe callback system for standard C++.'
+  homepage 'http://libsigc.sourceforge.net/'
+  version '2.10.0'
+  source_url 'http://ftp.gnome.org/pub/GNOME/sources/libsigc++/2.10/libsigc++-2.10.0.tar.xz'
```
uberhacker: Can we use https instead of http?

packages/libsigcplusplus.rb
```diff
@@ -0,0 +1,35 @@
...
+  binary_sha256 ({
+    aarch64: '526748e3a15fcd316fefc8c0359880b5c89640ad89a01de8c3f6210251e229f4',
+     armv7l: '526748e3a15fcd316fefc8c0359880b5c89640ad89a01de8c3f6210251e229f4',
+       i686: 'cbdf640dbf2462b3b8f24bb5f7cd2345f1ff3f46c983191e7ee31482213d24cd',
+     x86_64: 'c32ff29b0f4a8394b69502cd3dc030848f84c5c3df048beb88966f7b1b6c6935',
+  })
+
+  depends_on 'pkgconfig' => :build
+  depends_on 'diffutils' => :build
+  depends_on 'm4' => :build
```
uberhacker: These are all "core" packages so no need to include here.  It doesn't hurt including but not necessary.

PR #2184, opened 2018-04-02
[cairomm] revert to stable version 1.12.2

lovedheart:


uberhacker:
Mistake?

lovedheart:
That is too new for building stable packages.

PR #2185, opened 2018-04-02
Create pangomm.rb

lovedheart:
#1772

### DIFF
--- a/packages/pangomm.rb
+++ b/packages/pangomm.rb
@@ -4,7 +4,7 @@ class Pangomm < Package
   description 'pangomm is the official C++ interface for the Pango font layout library.'
   homepage 'https://developer.gnome.org/pangomm/stable/'
   version '2.40.1'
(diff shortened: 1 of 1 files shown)


packages/pangomm.rb
```diff
@@ -0,0 +1,43 @@
+require 'package'
+
+class Pangomm < Package
+  description 'pangomm is the official C++ interface for the Pango font layout library.'
+  homepage 'https://developer.gnome.org/pangomm/stable/'
+  version '2.40.1'
+  source_url 'http://ftp.acc.umu.se/pub/GNOME/sources/pangomm/2.40/pangomm-2.40.1.tar.xz'
```
uberhacker: We probably shouldn't use a mirror url for the source.  The official project has a newer version available?  https://ftp.gnome.org/pub/GNOME/sources/pangomm/2.41/pangomm-2.41.5.tar.xz

lovedheart: I prefer stable version. the latest one requires gcc-7.3 to compile.

PR #2186, opened 2018-04-02
[pango] upgrade to 1.42 with pre-built binaries

lovedheart:
#889 
#1772


uberhacker:
Thank you for the update @lovedheart!

PR #2190, opened 2018-04-03
[openjpeg] upgrade to 2.3.0 with binaries

lovedheart:


uberhacker:
Thank you for the updates @lovedheart.   You are doing an awesome job!

PR #2194, opened 2018-04-04
[llvm] Upgrade to 6.0.0 with pre-built binaries

lovedheart:


packages/llvm.rb
```diff
@@ -1,38 +1,56 @@
 require 'package'
 
 class Llvm < Package
-  description 'The LLVM Project is a collection of modular and reusable compiler and toolchain technologies.'
-  homepage 'https://llvm.org/'
-  version '5.0.1'
-  source_url 'https://releases.llvm.org/5.0.1/llvm-5.0.1.src.tar.xz'
-  source_sha256 '5fa7489fc0225b11821cab0362f5813a05f2bcf2533e8a4ea9c9c860168807b0'
+    description 'The LLVM Project is a collection of modular and reusable compiler and toolchain technologies.'
+  homepage 'https://compiler-rt.llvm.org/'
```
skiman6010: Just curious, why did you change the homepage link?

lovedheart: Originally, I was about to fix compiler-rt problem.

packages/llvm.rb
```diff
@@ -2,37 +2,55 @@
 
 class Llvm < Package
   description 'The LLVM Project is a collection of modular and reusable compiler and toolchain technologies.'
-  homepage 'https://llvm.org/'
-  version '5.0.1'
-  source_url 'https://releases.llvm.org/5.0.1/llvm-5.0.1.src.tar.xz'
-  source_sha256 '5fa7489fc0225b11821cab0362f5813a05f2bcf2533e8a4ea9c9c860168807b0'
+  homepage 'http://llvm.org/'
+  version '6.0.0'
+    source_url 'http://llvm.org/releases/6.0.0/llvm-6.0.0.src.tar.xz'
```
uberhacker: We should be using https instead of http for the source, if possible.  Also, this is petty but it would be nice if the indentation was consistent.

skiman6010: I accepted the PR because http redirects to https in this instance. I agree that we should be consistent and should have mentioned the indentation, so that one is on me. I'll fix the indentation and https for you.

uberhacker: If http redirects to https, we still should use https.  The extra redirect could potentially create a problem and cause a delay in the download.  Since this PR already includes pre-built binaries, it won't be a big deal for end-users.

skiman6010: Sounds like I just need to be more picky like you. :)

uberhacker: Picky is better than nit-picky, I guess. :-P  I just try and think about how things might affect future versions.

PR #2197, opened 2018-04-05
Fix armadillo package

nickjmeyer:
## Description

Adds `cmake` as a build dependency and fixes install paths.

## Addtional information

Works properly:
- [x] x86_64

### DIFF
--- a/packages/armadillo.rb
+++ b/packages/armadillo.rb
@@ -21,7 +21,6 @@ class Armadillo < Package
   def self.install
     system "make",
            "DESTDIR=#{CREW_DEST_DIR}",
-           "LIBDIR=#{CREW_LIB_PREFIX}",
            "install"
   end
 
--- a/packages/armadillo.rb
+++ b/packages/armadillo.rb
@@ -8,17 +8,21 @@ class Armadillo < Package
   source_url 'https://downloads.sourceforge.net/project/arma/armadillo-8.400.0.tar.xz'
   source_sha256 '5cb6bc2f457a9d6a0758cfb15c418d48289909daccd79d0e428452029285dd9b'
 
+  depends_on 'cmake' => :build
   depends_on 'openblas'
 
   def self.build
     system "./configure",
-           "--prefix=#{CREW_PREFIX}",
-           "--libdir=#{CREW_LIB_PREFIX}"
+           "-DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX}",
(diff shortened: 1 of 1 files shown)


packages/armadillo.rb
```diff
@@ -8,17 +8,21 @@ class Armadillo < Package
   source_url 'https://downloads.sourceforge.net/project/arma/armadillo-8.400.0.tar.xz'
   source_sha256 '5cb6bc2f457a9d6a0758cfb15c418d48289909daccd79d0e428452029285dd9b'
 
+  depends_on 'cmake' => :build
   depends_on 'openblas'
 
   def self.build
     system "./configure",
```
uberhacker: Should this line be `system "cmake",` instead?

packages/armadillo.rb
```diff
@@ -8,17 +8,21 @@ class Armadillo < Package
...
+           "-DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX}",
+           "-DCMAKE_INSTALL_LIBDIR=#{CREW_LIB_PREFIX}"
     system "make"
   end
 
   def self.install
-    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "make",
+           "DESTDIR=#{CREW_DEST_DIR}",
+           "LIBDIR=#{CREW_LIB_PREFIX}",
```
uberhacker: LIBDIR shouldn't be necessary here since the cmake option was already supplied in the build section.

packages/armadillo.rb
nickjmeyer: Their `configure` script is just a wrapper around `cmake`.  These arguments get forwared on to `cmake`.  I left it as using `configure` because that's what they say in their documentation.

packages/armadillo.rb
nickjmeyer: Fixed

packages/armadillo.rb
uberhacker: But cmake is not executed, only make?

nickjmeyer: `cmake` is executed.  The `configure` bascially just forwards all arguments to `cmake`.  The final line of the `configure` script is `cmake $@ .`.

uberhacker: OK, thanks for the clarification.  It makes sense now.

nickjmeyer: Sure thing.  👍

PR #2198, opened 2018-04-05
Add glog package.

nickjmeyer:
## Description

Adds the Google Logging library (aka Glog).

## Addtional information
This has only been tested on x86_64.

Works properly:
- [x] x86_64
- [ ] aarch64

### DIFF
--- a/packages/glog.rb
+++ b/packages/glog.rb
@@ -1,0 +1,27 @@
+require 'package'
+
+class Glog < Package
+  description 'A C++ implementation of the Google logging module.'
+  homepage 'https://github.com/google/glog'
+  version '0.3.5'
+
+  source_url 'https://github.com/google/glog/archive/v0.3.5.zip'
+  source_sha256 '267103f8a1e9578978aa1dc256001e6529ef593e5aea38193d31c2872ee025e8'
+
+  depends_on 'openblas'
+
+  def self.build
+    system "./configure",
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}"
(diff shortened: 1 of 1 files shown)


packages/glog.rb
```diff
@@ -0,0 +1,27 @@
...
+
+class Glog < Package
+  description 'A C++ implementation of the Google logging module.'
+  homepage 'https://github.com/google/glog'
+  version '0.3.5'
+
+  source_url 'https://github.com/google/glog/archive/v0.3.5.zip'
+  source_sha256 '267103f8a1e9578978aa1dc256001e6529ef593e5aea38193d31c2872ee025e8'
+
+  depends_on 'openblas'
```
uberhacker: Missing `depends_on 'unzip' => :build`?

packages/glog.rb
```diff
@@ -0,0 +1,27 @@
...
+    system "./configure",
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}"
+    system "make"
+  end
+
+  def self.install
+    system "make",
+           "DESTDIR=#{CREW_DEST_DIR}",
+           "LIBDIR=#{CREW_LIB_PREFIX}",
```
uberhacker: LIBDIR shouldn't be necessary here since it was already provided in the build section.

packages/glog.rb
nickjmeyer: Fixed

packages/glog.rb
nickjmeyer: Fixed

PR #2203, opened 2018-04-05
Create libxml2_python.rb

lovedheart:
Add libxml2 python module


uberhacker:
Any current packages need dependencies modified to use libxml2_python instead of libxml2 (no python)?

PR #2205, opened 2018-04-05
[ansible] Upgrade to 2.5.0 with pre-built binaries

### LINKED ISSUE #2104: [ansible] Unable to build
```
    ImportError: /tmp/pip-build-mCADKr/pynacl/.eggs/cffi-1.11.5-py2.7-linux-armv7l.egg/_cffi_backend.so: failed to map segment from shared object
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-mCADKr/pynacl/
You are using pip version 9.0.1, however version 9.0.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
ansible failed to build: `pip install --prefix /usr/local --root /usr/local/tmp/crew/dest -I ansible==2.4.3` exited with 1
```

lovedheart:
Add "check and install dependencies" during post-installation to ensure execution of runtime.

Fixes #2104

### DIFF
--- a/packages/ansible.rb
+++ b/packages/ansible.rb
@@ -32,7 +32,7 @@ class Ansible < Package
   end
   
   def self.postinstall
-    puts "Check and install dependencies ...".blue
+    puts "Check and install dependencies ...".lightblue
     system "yes | pip install ansible"
   end
 end
--- a/packages/ansible.rb
+++ b/packages/ansible.rb
@@ -3,15 +3,36 @@ require 'package'
 class Ansible < Package
   description 'Ansible is a radically simple IT automation engine that automates cloud provisioning, configuration management, application deployment, intra-service orchestration, and many other IT needs.'
   homepage 'https://www.ansible.com/'
-  version '2.4.3'
-  source_url 'https://releases.ansible.com/ansible/ansible-2.4.3.0.tar.gz'
-  source_sha256 '0e98b3a56928d03979d5f8e7ae5d8e326939111b298729b03f00b3ad8f998a3d'
+  version '2.5.0'
+  source_url 'http://releases.ansible.com/ansible/ansible-2.5.0.tar.gz'
+  source_sha256 '714ede4b0f5455eb2dd2536d8cabb42ab439f52b5ac902308ad027b46f68e6dc'
+
+  binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/ansible-2.5.0-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/ansible-2.5.0-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/ansible-2.5.0-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/ansible-2.5.0-chromeos-x86_64.tar.xz',
+  })
+  binary_sha256 ({
+    aarch64: '1101f5205b7310836e095922340ce1d69986d3292e8b354184baef265eb7548e',
+     armv7l: '1101f5205b7310836e095922340ce1d69986d3292e8b354184baef265eb7548e',
+       i686: '2631c1e1b83dfd75021c8dcc28e2301bb267ac1be95857bbe60cab91fcb0b44c',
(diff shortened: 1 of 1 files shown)


packages/ansible.rb
```diff
@@ -3,15 +3,36 @@
...
+     armv7l: '1101f5205b7310836e095922340ce1d69986d3292e8b354184baef265eb7548e',
+       i686: '2631c1e1b83dfd75021c8dcc28e2301bb267ac1be95857bbe60cab91fcb0b44c',
+     x86_64: 'd588644e26d4b54daf55b921914dd2f32ee67ddef7277af89b9726f729cc7f1c',
+  })
 
   depends_on 'python27'
 
-  def self.install
+  def self.build
     system "yes | pip install setuptools"
```
uberhacker: Remove?

packages/ansible.rb
```diff
@@ -3,15 +3,36 @@
...
-    system "yes | pip uninstall setuptools"
+    system "python2 setup.py build"
+  end
+
+  def self.install
+    system "python2 setup.py install --root=#{CREW_DEST_DIR}"
+  end
+  
+  def self.postinstall
+    puts "Check and install dependencies ...".blue
```
uberhacker: .blue is usually a bit dark.  .lightblue might be a better color (stands out better).

packages/ansible.rb
lovedheart: What do you mean??

packages/ansible.rb
```diff
@@ -3,15 +3,36 @@
...
+    system "yes | pip install ansible"
```
uberhacker: Use `depends_on 'ansible'` instead?

uberhacker: Why do we need this if python already installs it in the install section?

packages/ansible.rb
uberhacker: Is this required?

packages/ansible.rb
```diff
@@ -3,15 +3,36 @@
...
+  })
 
   depends_on 'python27'
 
-  def self.install
+  def self.build
     system "yes | pip install setuptools"
-    system "pip install --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR} -I ansible==2.4.3"
-    system "yes | pip uninstall setuptools"
+    system "python2 setup.py build"
```
uberhacker: How do we ensure the correct version is getting installed?

packages/ansible.rb
lovedheart: It will install dependencies.  Using setup.py to build and install to prevent some unknown errors, like #2104

packages/ansible.rb
```diff
@@ -3,15 +3,36 @@
...
+  end
+
+  def self.install
+    system "python2 setup.py install --root=#{CREW_DEST_DIR}"
+  end
+  
+  def self.postinstall
+    puts "Check and install dependencies ...".lightblue
+    system "yes | pip install ansible"
```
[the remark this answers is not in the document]
lovedheart: chronos@localhost ~/Downloads $ pip list
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.
ansible (2.5.0)
itstool (2.0.4)
libxml2-python (2.9.8)
pip (9.0.1)
pyxattr (0.5.5)
setuptools (28.8.0)
You are using pip version 9.0.1, however version 9.0.3 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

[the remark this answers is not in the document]
uberhacker: I guess I still don't understand why `pip install ansible` is necessary.  The version is also not controlled.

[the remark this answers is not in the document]
lovedheart: You can try without this. Then `ansible` can not start due to missing dependencies.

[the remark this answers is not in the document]
uberhacker: What does the install doc say about using a requirements.txt file?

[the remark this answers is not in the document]
uberhacker: I just don't like all these dependencies getting installed with no tracking.  An uninstall will not get everything, leaving behind a bunch of dependency residue.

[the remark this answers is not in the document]
lovedheart: Uninstall is not so smart at the moment.

lovedheart:
It makes only sense to use `python setup.py build (install)` to install external python modules, like #2203

PR #2207, opened 2018-04-05
[cmake] fix dependeny

lovedheart:
remove buildessential to prevent dependency loop, add libuv as dependency, fresh build have been performed.


lovedheart:
#2206

PR #2209, opened 2018-04-05
Clarify prereqs.  Wiki just gets you in dev mode

nealmcb:
When I read the Prerequisites in the README.md the first time, I didn't see anything on the given wiki page telling me anything about crew, so I felt lost. Finally I realized that although the instruction to visit the wiki was in a separate paragraph from the one above which had already said I needed to be in dev mod, it didn't expect me to have already gotten in to dev mode, and was sending me to the wiki to do so.

So I've simply rephrased the README.md in order to help people know what they're looking for on the wiki.
I also gave at least one hint on what is needed for security in dev mode. That could and should be expanded somewhere but I don't know of a good source yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,11 +21,10 @@ Chromebooks with Chrome OS run a Linux kernel. The only missing pieces to use th
 Prerequisites
 -------------
 
-You will need a Chromebook with developer mode enabled.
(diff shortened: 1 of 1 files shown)


uberhacker:
Thank you for clarifying @nealmcb!

PR #2210, opened 2018-04-06
Add gawk to buildessential

lovedheart:


uberhacker:
Bonus points for removing trailing whitespace.

PR #2216, opened 2018-04-08
Add used disk space to the files command

pgdagenais:
Following this issue, https://github.com/skycocker/chromebrew/issues/2145

@uberhacker proposed that we add a line that shows the disk space used by the package that we call using `crew file <package>`

### DIFF
--- a/crew
+++ b/crew
@@ -234,7 +234,7 @@ def files (pkgName)
   if File.exists? "#{filelist}"
     system "sort #{filelist}"
     lines = `wc -l "#{filelist}"`.strip.split(' ')[0].to_i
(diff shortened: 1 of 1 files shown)


uberhacker:
@pgdagenais: Thank you for the update!  Not sure where you got that variable but it looks like it always finds the total disk usage of the current directory?  I think you will probably want something like this:
```
+    size = `for f in $(cat #{filelist}); do du -b $f; done | awk '{ total += $1 }; END { mb = total / 1024 / 1024; printf("%.1f MB\\n", mb) }'`
    puts "Total found: #{lines}".lightgreen
+    puts "Disk usage: #{size}".lightgreen
```

pgdagenais:
Awesome thank you

PR #2218, opened 2018-04-08
Create dejagnu.rb

lovedheart:
The dependency (texinfo) should be recompiled again, since version mistach with perl

### DIFF
--- a/packages/dejagnu.rb
+++ b/packages/dejagnu.rb
@@ -1,0 +1,32 @@
+require 'package'
+
+class Dejagnu < Package
+  description 'DejaGnu is a framework for running test suites on GNU tools.'
+  homepage 'https://www.ansible.com/'
+  version '1.6.1'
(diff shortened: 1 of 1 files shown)


packages/dejagnu.rb
```diff
@@ -0,0 +1,32 @@
+require 'package'
+
+class Dejagnu < Package
+  description 'DejaGnu is a framework for running test suites on GNU tools.'
+  homepage 'https://www.ansible.com/'
```
uberhacker: Wrong homepage?

packages/dejagnu.rb
```diff
@@ -0,0 +1,32 @@
...
+  version '1.6.1'
+  source_url 'https://ftp.gnu.org/gnu/dejagnu/dejagnu-1.6.1.tar.gz'
```
uberhacker: It would be better if you used the mirror url: https://ftpmirror.gnu.org/dejagnu/dejagnu-1.6.1.tar.gz

PR #2232, opened 2018-04-11
[gcc7] minor change, add gcc4-compatible option

lovedheart:


packages/gcc7.rb
```diff
@@ -74,20 +83,52 @@ def self.build
...
   end
-
+  
+  
+  # preserve for check, skip check for current version
+  def self.check
+    Dir.chdir("objdir") do
+      #system "ulimit -s 32768"
+      #system "make -k check -j8"
+      #system "../contrib/test_summary > ~/Downloads/gcc_test_summary.txt"
```
uberhacker: Why commented?  Tests pass or fail?

packages/gcc7.rb
```diff
@@ -74,20 +83,52 @@ def self.build
...
+	    system "install -v -dm755 #{CREW_DEST_LIB_PREFIX}/bfd-plugins"
     end
   end
     
+  def self.postinstall
+    
+    #system "ln -sv gcc-7.3 #{CREW_PREFIX}/bin/gcc"
+	  #system "ln -sv gcc-7.3 #{CREW_PREFIX}/bin/cc"
+	  # Add a compatibility symlink to enable building programs with Link Time Optimization (LTO)
+	  system "ln -sfv #{CREW_PREFIX}/libexec/gcc/$(gcc -dumpmachine)/7.3.0/liblto_plugin.so #{CREW_LIB_PREFIX}/bfd-plugins/"
```
uberhacker: You probably want this in the install section.  I think that target directory will need to be created first also unless it already exists.

packages/gcc7.rb
```diff
@@ -74,20 +83,52 @@ def self.build
...
 	    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+	   
+	    # http://www.linuxfromscratch.org/lfs/view/development/chapter06/gcc.html#contents-gcc
+	    # move a misplaced file
+	    # The installation stage puts some files used by gdb under the /usr/local/lib(64) directory. This generates spurious error messages when performing ldconfig. This command moves the files to another location.
+      system "mkdir -pv #{CREW_DEST_PREFIX}/share/gdb/auto-load/usr/lib"
+      system "mv -v #{CREW_DEST_LIB_PREFIX}/*gdb.py #{CREW_DEST_PREFIX}/share/gdb/auto-load/usr/lib"
+	    
+	    # Install Binary File Descriptor library (BFD)
+	    system "install -v -dm755 #{CREW_DEST_LIB_PREFIX}/bfd-plugins"
```
uberhacker: What are we installing here?  Or are you just creating a directory for the symlink later?

packages/gcc7.rb
lovedheart: most succeed , some fail, bu not so important

packages/gcc7.rb
uberhacker: Can you please clean up the indentation a bit?  I imagine you have tabs instead of spaces maybe?

uberhacker:
Thank you for the update @lovedheart.  Please see inline comments.

packages/gcc7.rb
```diff
@@ -110,25 +110,25 @@ def self.check
...
-    
+  def self.postinstall  
     #system "ln -sv gcc-7.3 #{CREW_PREFIX}/bin/gcc"
-	  #system "ln -sv gcc-7.3 #{CREW_PREFIX}/bin/cc"
-	  # Add a compatibility symlink to enable building programs with Link Time Optimization (LTO)
-	  system "ln -sfv #{CREW_PREFIX}/libexec/gcc/$(gcc -dumpmachine)/7.3.0/liblto_plugin.so #{CREW_LIB_PREFIX}/bfd-plugins/"
+    #system "ln -sv gcc-7.3 #{CREW_PREFIX}/bin/cc"
+	  
+    # Add a compatibility symlink to enable building programs with Link Time Optimization (LTO)	  
+    system "ln -sfv #{CREW_PREFIX}/libexec/gcc/$(gcc -dumpmachine)/7.3.0/liblto_plugin.so #{CREW_LIB_PREFIX}/bfd-plugins/"
```
uberhacker: Why here and not in the install section?  The reason I ask is this will not be included in the tracked files to be uninstalled.

packages/gcc7.rb
```diff
@@ -74,20 +84,49 @@ def self.build
...
+      system "mkdir -pv #{CREW_DEST_PREFIX}/share/gdb/auto-load/usr/lib"
+      system "mv -v #{CREW_DEST_LIB_PREFIX}/*gdb.py #{CREW_DEST_PREFIX}/share/gdb/auto-load/usr/lib"
+	    
+      # Install Binary File Descriptor library (BFD)
+      system "install -v -dm755 #{CREW_DEST_LIB_PREFIX}/bfd-plugins"
+      
+      # Add a compatibility symlink to enable building programs with Link Time Optimization (LTO)
+      system "ln -sfv #{CREW_DEST_PREFIX}/libexec/gcc/$(gcc -dumpmachine)/7.3.0/liblto_plugin.so #{CREW_DEST_LIB_PREFIX}/bfd-plugins/"
+      system "ln -sv gcc-7.3 #{CREW_DEST_PREFIX}/bin/gcc"
+      system "ln -sv gcc-7.3 #{CREW_DEST_PREFIX}/bin/cc"
```
uberhacker: Did you create new pre-built binaries after these changes?

lovedheart: newly recompiled

PR #2234, opened 2018-04-12
[pkgconfig] Recompile binaries

lovedheart:
Reconfigure system-include-path and system-library-path for different architectures.

### DIFF
--- a/packages/pkgconfig.rb
+++ b/packages/pkgconfig.rb
@@ -3,34 +3,48 @@ require 'package'
 class Pkgconfig < Package
   description 'pkg-config is a helper tool used when compiling applications and libraries.'
   homepage 'https://www.freedesktop.org/wiki/Software/pkg-config/'
-  version '0.29.2'
+  version '0.29.2-0'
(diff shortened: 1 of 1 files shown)


packages/pkgconfig.rb
```diff
@@ -3,34 +3,52 @@
...
-      system "make"
+    case ARCH
+    when "x86_64"
+      system "./configure",
+             "--prefix=#{CREW_PREFIX}",
+             "--libdir=#{CREW_LIB_PREFIX}",
+             "--with-internal-glib",
+             "--with-pc-path=#{CREW_PREFIX}/lib/pkgconfig:#{CREW_PREFIX}/lib64/pkgconfig:#{CREW_PREFIX}/share/pkgconfig",
+             "--with-system-include-path=/usr/include:#{CREW_PREFIX}/include",
+             "--with-system-library-path=#{CREW_PREFIX}/lib:#{CREW_PREFIX}/lib64",
```
uberhacker: Do we need both `/lib` and `/lib64`?

lovedheart: yes. compiler-rt will install libs on /usr/local/lib

PR #2235, opened 2018-04-12
[texinfo] add pre-built binaries

lovedheart:


packages/texinfo.rb
```diff
@@ -7,6 +7,19 @@ class Texinfo < Package
   source_url 'http://ftpmirror.gnu.org/texinfo/texinfo-6.5.tar.xz'
```
uberhacker: Can we change this to https instead of http?

PR #2250, opened 2018-04-16
[openblas] add pre-built binaries

lovedheart:
use `cmake` to build binaries for x86_64 to make sure that the libraries are installed in /usr/local/lib64.
normal `cmake` procedure only supports x86_64 right now.


uberhacker:
@lovedheart: Thank you for the update!   Can we test this a bit more on several different targets (aka cpu processor types)  to make sure `ATOM` works as an acceptable target?  Also, do you think we can still use the detection logic of openblas and then fallback to `ATOM` if not found?

PR #2277, opened 2018-04-23
[buildessential] update gcc version to 7.3.0

lovedheart:


packages/buildessential.rb
```diff
@@ -7,7 +7,7 @@ class Buildessential < Package
 
   is_fake
 
-  depends_on 'gcc'
+  depends_on 'gcc7'
```
uberhacker: You sure we're ready for this?  Bump version?

lovedheart: The option gcc4-compatible is added, no need to compile the packages from source code again.

uberhacker: Let's test this a bit more.  Please compile some more packages with gcc7 and push a few PRs.

uberhacker: Tests results good.  Nice work @lovedheart!

PR #2278, opened 2018-04-23
[install.sh] remove symbolic links to cc

lovedheart:
symbolic link has already been made in gcc7 package, consider to remove these two lines if gcc7 comes into core package, see #2277

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -170,9 +170,12 @@ cd $CREW_LIB_PATH/lib
 curl -# -o package.rb $URL/lib/package.rb
 curl -# -o package_helpers.rb $URL/lib/package_helpers.rb
 
-#Making GCC act like CC (For some npm packages out there)
-rm -f $CREW_PREFIX/bin/cc
-ln -s $CREW_PREFIX/bin/gcc $CREW_PREFIX/bin/cc
+#Making GCC act like CC (For some npm packages out there, only required for gcc)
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -170,9 +170,12 @@ cd $CREW_LIB_PATH/lib
...
 
-#Making GCC act like CC (For some npm packages out there)
-rm -f $CREW_PREFIX/bin/cc
-ln -s $CREW_PREFIX/bin/gcc $CREW_PREFIX/bin/cc
+#Making GCC act like CC (For some npm packages out there, only required for gcc)
+#rm -f $CREW_PREFIX/bin/cc
+#ln -s $CREW_PREFIX/bin/gcc $CREW_PREFIX/bin/cc
+
+#package gcc7 has already made symbolic links for cc and gcc, no action required here
+
```
uberhacker: I'm hesitant to make this change.  Can we test a bit more how gcc7 handles compiling packages?  Please update a few packages here: https://github.com/skycocker/chromebrew/issues/1585.  I tried gcc7 recently and it borked on a few things.

lovedheart: Which package?

uberhacker: Let me see if I can reproduce the error.

uberhacker: Wasn't able to reproduce.  I tested a few more packages just to be sure and no problems.  I think we're good.

PR #2290, opened 2018-04-28
remove individual xorg proto

lovedheart:
Remove individual proto


uberhacker:
I like reducing complexity!

PR #2291, opened 2018-04-28
adjust ncurses

lovedheart:
Compile libncurses and libncursesw as a compact package and adjust dependencies on some packages


packages/aspell.rb
```diff
@@ -21,7 +21,7 @@ class Aspell < Package
   })
 
   depends_on 'ruby' unless File.exists? "#{CREW_PREFIX}/bin/ruby"
-  depends_on 'ncursesw'
+  depends_on 'ncurses'
```
uberhacker: If you leave the `ncursesw` dependency, you get both `ncurses` and `ncursesw`.  Any reason not to have `ncursesw`?

lovedheart: the modified ncurses have both ncurses and ncursesw

packages/ncurses.rb
```diff
@@ -3,38 +3,68 @@
...
+             "--with-cxx-shared",
+             "--without-debug",
+             "--enable-pc-files",
+             "--with-pkg-config-libdir=#{CREW_LIB_PREFIX}/pkgconfig",
+             "--disable-db-install",
+             "--without-manpages",
+             "--without-progs",
+             "--without-tack"
+      system "make"
+    end
```
uberhacker: Oh, I see what you are doing now.  You want to combine ncursesw in with the ncurses package?

lovedheart: Yes.

lovedheart: Since they come from the same source code

uberhacker: Yes, that makes a lot of sense.  I like the way you think. :)

packages/ncurses.rb
```diff
@@ -3,38 +3,68 @@
...
-     armv7l: '4e33d4fc68ea5b47bae07aa710b1f801fd80ba85a02c0060c6f157100acadaf4',
-       i686: '16a5f5c03a4e9591146dc56a1ed4b017c09ad9d72da38d104ce282a3d25ef9fd',
-     x86_64: '1eab39eee49e16799ad61def29e69c50a7ea39b04a8da2ff4f4ada033e81273c',
+    aarch64: 'a12e70f779b2375912a5373b5f320c7b4bef8ddcba5144709e2aed67f8011afe',
+     armv7l: 'a12e70f779b2375912a5373b5f320c7b4bef8ddcba5144709e2aed67f8011afe',
+       i686: 'a89d5d79d91c8717cfecf7ebbebae6479f8cded2884798643d061605e449c1df',
+     x86_64: 'fcaf728604029920a14f27f64f498dfc8a3586bc0426ca6a890adf925e39e204',
   })
 
   depends_on 'diffutils' => :build
```
uberhacker: Remove since in core?

PR #2303, opened 2018-04-30
add compilation step for glibc2.23

lovedheart:


packages/glibc223.rb
```diff
@@ -3,17 +3,142 @@
 class Glibc223 < Package
   description 'GNU C Library'
   homepage 'https://www.gnu.org/software/libc/'
-  version '2.23-2'
+  version '2.23-3'
+  source_url 'https://ftp.gnu.org/gnu/glibc/glibc-2.23.tar.xz'
```
uberhacker: Can we use `https://ftpmirror.gnu.org/glibc/glibc-2.23.tar.xz` instead?  This should localize the downloads better.

packages/glibc223.rb
```diff
@@ -3,17 +3,142 @@
...
+
+  depends_on 'libgd' => :build
+  depends_on 'texinfo' => :build
+
+  def self.patch
+    # Apply patch due to new version of binutils  which causes compilation failure
+    # http://lists.busybox.net/pipermail/buildroot/2017-August/199812.html
+    Dir.chdir "misc" do
+      if File.readlines("regexp.c").grep(/monitor/).any?
+        puts "Pachted!"
```
uberhacker: Patched misspelled?

packages/glibc223.rb
```diff
@@ -3,17 +3,142 @@
...
+      else
+        system "sed -i 's,char \\*loc1,char \\*loc1 __attribute__ ((nocommon)),g' regexp.c"
+        system "sed -i 's,char \\*loc2,char \\*loc2 __attribute__ ((nocommon)),g' regexp.c"
+        system "sed -i 's,char \\*locs,char \\*locs __attribute__ ((nocommon)),g' regexp.c"
+      end
+    end
+  end
+
+  def self.build
+    system "mkdir -pv glibc_build"
```
uberhacker: Not sure if you are aware, but this works also:
```
Dir.mkdir "glibc_build"
```

lovedheart: Can I keep it as it is?

packages/glibc223.rb
```diff
@@ -3,17 +3,142 @@
...
+               "--libdir=#{CREW_LIB_PREFIX}",
+               "--with-headers=#{CREW_PREFIX}/include",
+               "--disable-werror",
+               "--disable-sanity-checks",
+               "--enable-shared",
+               "--enable-obsolete-rpc",
+               "libc_cv_forced_unwind=yes"
+      when 'x86_64'
+        system "echo \"slibdir=#{CREW_LIB_PREFIX}\" > configparms"
+        puts "echo \"slibdir=#{CREW_LIB_PREFIX}\" > configparms"
```
uberhacker: Is the `configparms` only needed on x86_64?

packages/glibc223.rb
uberhacker: Sure, but just thought I would mention it in case you didn't know already.

packages/glibc223.rb
lovedheart: yes

packages/glibc223.rb
```diff
@@ -3,17 +3,142 @@
...
+
+  depends_on 'libgd' => :build
+  depends_on 'texinfo' => :build
+
+  def self.patch
+    # Apply patch due to new version of binutils  which causes compilation failure
+    # http://lists.busybox.net/pipermail/buildroot/2017-August/199812.html
+    Dir.chdir "misc" do
+      if File.readlines("regexp.c").grep(/monitor/).any?
+        puts "Patched!"
```
uberhacker: `puts "Patched!".lightgreen`?

uberhacker:
I really appreciate your effort to get the source code and compile it @lovedheart!  Next step is upgrade glibc to 2.27?

lovedheart:
@uberhacker No update to glibc 2.27 until chrome os uses it.

uberhacker:
Isn't this package what dictates the version Chrome OS uses?

lovedheart:
Yes. Otherwise all corrupted.

PR #2306, opened 2018-05-01
[filecmd] update to 5.33

lovedheart:


packages/filecmd.rb
```diff
@@ -3,33 +3,22 @@
...
+  version '5.33'
+  source_url 'ftp://ftp.astron.com/pub/file/file-5.33.tar.gz'
+  source_sha256 '1c52c8c3d271cd898d5511c36a68059cda94036111ab293f01f83c3525b737c6'
 
   def self.build
-    system "./configure --libdir=#{CREW_LIB_PREFIX}"
+    system "./configure",
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}",
+           "--disable-libseccomp"
```
uberhacker: What is the reason to exclude libseccomp support?

lovedheart: No libseccomp available. Cant pass configuration

uberhacker: Add new package?  See https://github.com/seccomp/libseccomp.

lovedheart: Maybe merge this first. This is not critical

uberhacker: Just curious why we need to disable now and not with earlier versions.

lovedheart: It can not automatically detect the existance of libseccomp any more.

PR #2307, opened 2018-05-01
[xorg_libs] add compact xorg library (for testing)

lovedheart:


packages/xorg_libs.rb
```diff
@@ -0,0 +1,272 @@
...
+    ############################################################
+      for index in 0 ... $xorg_lib_source_name.size
+        folder_name = "#{$xorg_lib_source_name[index]}"
+        folder_name.slice! ".tar.gz"  # remove .tar.gz
+  
+        puts "Configuring #{folder_name} ...".lightblue
+        Dir.chdir(folder_name) do
+         
+          if folder_name.include? "libICE"
+            puts "libICE".lightred
```
uberhacker: Yellow instead of red?  Red usually indicates an error.

packages/xorg_libs.rb
```diff
@@ -0,0 +1,272 @@
...
+                   "--disable-static",
+                   "--enable-shared",
+                   "--with-appdefaultdir=#{CREW_PREFIX}/etc/X11/app-defaults"
+          else
+            puts "#{folder_name}".lightred
+            system "./configure",
+                   "--prefix=#{CREW_PREFIX}",
+                   "--libdir=#{CREW_LIB_PREFIX}",
+                   "--disable-static",
+                   "--enable-shared"
```
uberhacker: I don't think you are going to want this else condition.  This will execute every time in the loop when folder_name does not include libXt-.  Use case...when statements instead?

packages/xorg_libs.rb
```diff
@@ -0,0 +1,272 @@
...
+          end
+        
+          system "make"
+          system "make install"  # This step is only
+          system "find #{CREW_LIB_PREFIX} -type f -name *.la -delete"
```
uberhacker: Don't we do this already in crew?

packages/xorg_libs.rb
```diff
@@ -0,0 +1,272 @@
+require 'package'
+
+class Xorg_libs < Package
+  description 'A collection of xorg libraries.'
+  homepage 'https://www.x.org'
+  version '1.0.0'
```
uberhacker: Wrong version?

lovedheart: fake one

packages/xorg_libs.rb
```diff
@@ -0,0 +1,272 @@
...
+  source_url 'http://downloads.sourceforge.net/project/libtirpc/libtirpc/1.0.2/libtirpc-1.0.2.tar.bz2'
+  source_sha256 '723c5ce92706cbb601a8db09110df1b4b69391643158f20ff587e20e7c5f90f5'
```
uberhacker: Why are we using libtirpc source?

packages/xorg_libs.rb
lovedheart: la files should be removed after each installation

packages/xorg_libs.rb
lovedheart: Fake one. How to use a dummy ?

packages/xorg_libs.rb
```diff
@@ -0,0 +1,272 @@
...
+  $xorg_lib_source_name.push("libpciaccess-0.14.tar.gz")
+  $sha256sum_source.push("8d86e64893917be3dfb1c5e837888d1275399c818783474002203d751312b03c")
+  
+  
+  
+  $xorg_lib_source_name.push("libxkbfile-1.0.9.tar.gz")
+  $sha256sum_source.push("95df50570f38e720fb79976f603761ae6eff761613eb56f258c3cb6bab4fd5e3")
+  
+  $xorg_lib_source_name.push("libxshmfence-1.3.tar.gz")
+  $sha256sum_source.push("7eb3d46ad91bab444f121d475b11b39273142d090f7e9ac43e6a87f4ff5f902c")
```
uberhacker: How are we going to manage these "dependency" versions when individual libraries are updated?

uberhacker: I'm thinking it might be better to just use the existing crew packages as dependencies.

lovedheart: The dependencies are automatically handled according to installation procedure.
1) Follow the compilation steps -> build code -> install to host
2) Build related packages (DESTDIR)
3) Remove installed packages on host.

uberhacker: This makes the packaging system more modular so other packages can use the same dependencies.

lovedheart: We only need to see it as one package. The internal dependency should be safely handled if the package can be sucessfully built.

lovedheart: The dependency lies mainly on xorg-related source code.

uberhacker: So basically this is a static build of all the libraries?

lovedheart: Refer to http://www.linuxfromscratch.org/blfs/view/svn/x/x7lib.html

packages/xorg_libs.rb
uberhacker: I thought we do this already in crew?

packages/xorg_libs.rb
lovedheart: Yes. Build all xorg-related libraries.

lovedheart: Currently we have 32 individual xorg libraries.

packages/xorg_libs.rb
uberhacker: I have needed to do this on several packages myself.  What I usually do is grab the README.md file if the source is from github.com.  What you want is something small related to the project that can be versioned.

packages/xorg_libs.rb
lovedheart: But only once

lovedheart:
Packge build from **multiple sources** in a **single package** might be implemented in crew

packages/xorg_libs.rb
lovedheart: Most of these packages are really old. I think only a few of them are still in active status.

lovedheart: The dependency internally is not a big problem. Once the source code updated, we should always check the build procedure. It's true for all packages. Not for this bundle one.

uberhacker: Reason to use weston and/or wayland instead?  I need to get back to my day job but think about this a bit more and we can discuss it later.  I don't think we will need the kitchen sink for GUI stuff.

lovedheart: There are not many software using wayland. Ubuntu 18.04 switch back to xorg

PR #2312, opened 2018-05-01
Update Dart.rb to solve issues with pub

Ewpratten:
# Fixes
`pub get` now requires a file called *version* in `/usr/local/`

### DIFF
--- a/packages/dart.rb
+++ b/packages/dart.rb
@@ -39,5 +39,6 @@ class Dart < Package
     system "cp -r bin/ #{CREW_DEST_PREFIX}"
     system "cp -r lib/* #{CREW_DEST_LIB_PREFIX}"
     system "cp -r include/ #{CREW_DEST_PREFIX}"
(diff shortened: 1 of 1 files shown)


packages/dart.rb
```diff
@@ -39,5 +39,6 @@ def self.install
     system "cp -r bin/ #{CREW_DEST_PREFIX}"
     system "cp -r lib/* #{CREW_DEST_LIB_PREFIX}"
     system "cp -r include/ #{CREW_DEST_PREFIX}"
+    system "cp version /usr/local/version" # This stops 'pub get' from throwing errors
```
uberhacker: Thank you for the update @Ewpratten!  Can we use a crew constant to make sure this file is tracked?
```
system "cp version #{CREW_DEST_PREFIX}/version"
```

Ewpratten: Done!

PR #2327, opened 2018-05-06
Added dpkg.rb to Packages directory.

JL2210:
New file found in ./packages/: dpkg.rb.
Concratulations! Dpkg can now be installed on Chrome OS!
Thank you for this wonderful project. I hope I have given you something valuable.

### DIFF
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -24,7 +24,14 @@ class Dpkg < Package
       system "make install DESTDIR=#{CREW_DEST_DIR}"
       system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
       system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
-      system "touch #{CREW_DEST_PREFIX}/var/lib/dpkg/status"
+      system "echo >> /usr/local/var/lib/dpkg/status"
     end
+  end
+    def self.postinstall
+    puts "You may need to run" 
+    puts "'sudo chromeos-firmwareupdate --mode=todev',"
+    puts "followed by 'sudo /usr/share/vboot/bin/make_dev_ssd.sh --remove_rootfs_verification',"
+    puts "follow the instructions, and restart your Chromebook before installing packages."
+    puts "However, this isn't necessary if you aren't going to use dpkg -i."
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -10,17 +10,18 @@ class Dpkg < Package
     depends_on 'icu4c' => :build
 
     def self.build
-    	    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
-	Dir.chdir ("dpkg")
-	    system "git checkout 1.19.0.5"
-	    system "autoreconf -i -f"
-	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
-	    system "make"
-	end
+          system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
+  Dir.chdir ("dpkg")
+      system "git checkout 1.19.0.5"
+      system "autoreconf -i -f"
+      system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
+      system "make"
+  end
 
-	def self.install
-	    system "make install DESTDIR=#{CREW_DEST_DIR}"
-	    system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
-	    system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
+  def self.install
+      Dir.chdir ("dpkg")
+      system "make install DESTDIR=#{CREW_DEST_DIR}"
+      system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
+      system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -10,9 +10,9 @@ class Dpkg < Package
     depends_on 'icu4c' => :build
 
     def self.build
-	    system "rm -rf {*,.*}"
 	    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
 	Dir.chdir ("dpkg")
+	    system "git checkout 1.19.0.5"
 	    system "autoreconf -i -f"
 	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
 	    system "make"
@@ -20,7 +20,10 @@ class Dpkg < Package
 
 	def self.install 
 	    system "make install DESTDIR=#{CREW_DEST_DIR}"
-	    system "echo \"# Dpkg Alias (Do NOT Touch)\" >> ~/.bashrc"
-	    system "echo \"alias dpkg='dpkg --root=/usr/local'\" >> ~/.bashrc"
+	    system "mkdir -p ~/.config/fish"
+	    system "mkdir -p /usr/local/var/lib/dpkg"
+	    system "echo \"# Dpkg Alias (Do NOT Touch)\" >> ~/.{bashrc,zshrc,config/fish/config.fish}"
+	    system "echo \"alias dpkg='dpkg --root=/usr/local'\" >> ~/.{bashrc,zshrc,config/fish/config.fish}"
+	    system "touch /usr/local/var/lib/dpkg/status"
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -7,21 +7,23 @@ class Dpkg < Package
   source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.19.0.5/dpkg-1.19.0.5.tar.gz'
   source_sha256 'd38308afcd5d7896bbd1f946875b90f9d8510a8a96b44e4f14e781285e5d9641'
 
-    depends_on 'icu4c' => :build
+  depends_on 'icu4c' => :build
 
-    def self.build
-          system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
-  Dir.chdir ("dpkg")
+  def self.build
+    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
+    Dir.chdir ("dpkg")
       system "git checkout 1.19.0.5"
       system "autoreconf -i -f"
       system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
       system "make"
+    end
   end
 
   def self.install
-      Dir.chdir ("dpkg")
+    Dir.chdir ("dpkg")
       system "make install DESTDIR=#{CREW_DEST_DIR}"
       system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
       system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
+    end
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -20,6 +20,7 @@ class Dpkg < Package
 
 	def self.install 
 	    system "make install DESTDIR=#{CREW_DEST_DIR}"
-	    system "alias dpkg='dpkg --root=/usr/local'"
+	    system "echo \"# Dpkg Alias (Do NOT Touch)\" >> ~/.bashrc"
+	    system "echo \"alias dpkg=\"dpkg --root=/usr/local\"\" >> ~/.bashrc"
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -24,7 +24,6 @@ class Dpkg < Package
       system "make install DESTDIR=#{CREW_DEST_DIR}"
       system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
       system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
-      system "echo >> /usr/local/var/lib/dpkg/status"
     end
   end
     def self.postinstall
@@ -33,5 +32,6 @@ class Dpkg < Package
     puts "followed by 'sudo /usr/share/vboot/bin/make_dev_ssd.sh --remove_rootfs_verification',"
     puts "follow the instructions, and restart your Chromebook before installing packages."
     puts "However, this isn't necessary if you aren't going to use dpkg -i."
+    system "echo >> /usr/local/var/lib/dpkg/status"
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -1,0 +1,37 @@
+require 'package'
+
+class Dpkg < Package
+  description 'dpkg, a medium-level package manager for Debian'
+  homepage 'https://anonscm.debian.org/git/dpkg/'
+  version '1.19.0.5'
+  source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/master/dpkg-master.tar.gz'
+  source_sha256 'a64b257d89d95f2e199fb39c6b37bf9de9e15e2d1e1898060fa9c9af3899330e'
+
+    depends_on 'perl' => :build
+    depends_on 'pkgconfig' => :build
+    depends_on 'automake' => :build
+    depends_on 'make' => :build
+    depends_on 'gcc' => :build
+    depends_on 'linuxheaders' => :build
+    depends_on 'flex' => :build
+    depends_on 'bison' => :build
+    depends_on 'icu4c' => :build
+    depends_on 'gettext' => :build
+    depends_on 'autoconf' => :build
+    depends_on 'libtool' => :build
+
+    def self.build
+	    system "sudo mount -o remount,rw -io exec ~/"
+	    system "echo 'Thank you for entering your sudo password. It will be used to run the configure script.'"
+	    system "rm -rf {*,.*}"
+	    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
+	    Dir.chdir ("dpkg")
+	    system "autoreconf -i -f"
+	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
+	    system "make"
+	end
+
+    def self.install 
+	    system "make install DESTDIR=#{CREW_DEST_DIR}"
+  end
+end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -33,5 +33,6 @@ class Dpkg < Package
 
     def self.install 
 	    system "make install DESTDIR=#{CREW_DEST_DIR}"
+	    system "alias dpkg=\"dpkg --root=/usr/local\""
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -20,6 +20,7 @@ class Dpkg < Package
 
 	def self.install 
 	    system "make install DESTDIR=#{CREW_DEST_DIR}"
+	    system "ln -s /usr/ /usr/local/usr"
 	    system "mkdir -p ~/.config/fish"
 	    system "mkdir -p /usr/local/var/lib/dpkg"
 	    system "echo \"# Dpkg Alias (Do NOT Touch)\" >> ~/.{bashrc,zshrc,config/fish/config.fish}"
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -7,18 +7,7 @@ class Dpkg < Package
   source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.19.0.5/dpkg-1.19.0.5.tar.gz'
   source_sha256 'd38308afcd5d7896bbd1f946875b90f9d8510a8a96b44e4f14e781285e5d9641'
 
-    depends_on 'perl' => :build
-    depends_on 'pkgconfig' => :build
-    depends_on 'automake' => :build
-    depends_on 'make' => :build
-    depends_on 'gcc' => :build
-    depends_on 'linuxheaders' => :build
-    depends_on 'flex' => :build
-    depends_on 'bison' => :build
     depends_on 'icu4c' => :build
-    depends_on 'gettext' => :build
-    depends_on 'autoconf' => :build
-    depends_on 'libtool' => :build
 
     def self.build
 	    system "rm -rf {*,.*}"
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -1,11 +1,11 @@
 require 'package'
 
 class Dpkg < Package
-  description 'dpkg, a medium-level package manager for Debian'
+  description 'A medium-level package manager for Debian'
   homepage 'https://anonscm.debian.org/git/dpkg/'
   version '1.19.0.5'
-  source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/master/dpkg-master.tar.gz'
-  source_sha256 'a64b257d89d95f2e199fb39c6b37bf9de9e15e2d1e1898060fa9c9af3899330e'
+  source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.19.0.5/dpkg-1.19.0.5.tar.gz'
+  source_sha256 'd38308afcd5d7896bbd1f946875b90f9d8510a8a96b44e4f14e781285e5d9641'
 
     depends_on 'perl' => :build
     depends_on 'pkgconfig' => :build
@@ -25,13 +25,13 @@ class Dpkg < Package
 	    system "echo 'Thank you for entering your sudo password. It will be used to run the configure script.'"
 	    system "rm -rf {*,.*}"
 	    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
-	    Dir.chdir ("dpkg")
+	Dir.chdir ("dpkg")
 	    system "autoreconf -i -f"
 	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
 	    system "make"
 	end
 
-    def self.install 
+	def self.install 
 	    system "make install DESTDIR=#{CREW_DEST_DIR}"
 	    system "alias dpkg='dpkg --root=/usr/local'"
   end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -11,7 +11,7 @@ class Dpkg < Package
 
   def self.build
     system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
-    Dir.chdir ("dpkg")
+    Dir.chdir ("dpkg") do
       system "git checkout 1.19.0.5"
       system "autoreconf -i -f"
       system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
@@ -20,10 +20,11 @@ class Dpkg < Package
   end
 
   def self.install
-    Dir.chdir ("dpkg")
+    Dir.chdir ("dpkg") do
       system "make install DESTDIR=#{CREW_DEST_DIR}"
       system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
       system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
+      system "touch /usr/local/var/lib/dpkg/status"
     end
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -21,6 +21,6 @@ class Dpkg < Package
 	def self.install 
 	    system "make install DESTDIR=#{CREW_DEST_DIR}"
 	    system "echo \"# Dpkg Alias (Do NOT Touch)\" >> ~/.bashrc"
-	    system "echo \"alias dpkg=\"dpkg --root=#{CREW_PREFIX}\"\" >> ~/.bashrc"
+	    system "echo \"alias dpkg='dpkg --root=/usr/local'\" >> ~/.bashrc"
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -31,7 +31,7 @@ class Dpkg < Package
     puts "'sudo chromeos-firmwareupdate --mode=todev',"
     puts "followed by 'sudo /usr/share/vboot/bin/make_dev_ssd.sh --remove_rootfs_verification',"
     puts "follow the instructions, and restart your Chromebook before installing packages."
-    puts "However, this isn't necessary if you aren't going to use dpkg -i."
+    puts "However, this isn't necessary if you aren't going to install any packages."
     system "echo >> /usr/local/var/lib/dpkg/status"
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -10,21 +10,17 @@ class Dpkg < Package
     depends_on 'icu4c' => :build
 
     def self.build
-	    system "ARCH = `uname -m`.strip"
-            system "ARCH_LIB = if ARCH == 'x86_64' then 'lib64' else 'lib' end"          
-            system "CREW_LIB_PREFIX = /usr/local + '/' + ARCH_LIB"
-	    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
+    	    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
 	Dir.chdir ("dpkg")
 	    system "git checkout 1.19.0.5"
 	    system "autoreconf -i -f"
-	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=/usr/local"
+	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
 	    system "make"
 	end
 
 	def self.install
-	    system "make install DESTDIR=/usr/local/${ARCH_LIB}"
-	    system "ln -s /usr/ /usr/local/usr"
-	    system "mkdir -p /usr/local/var/lib/dpkg"
-	    system "touch /usr/local/var/lib/dpkg/status"
+	    system "make install DESTDIR=#{CREW_DEST_DIR}"
+	    system "mkdir -p /usr/local/usr/"
+	    system "ln -s /usr/local/ /usr/local/usr/local"
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -24,7 +24,7 @@ class Dpkg < Package
       system "make install DESTDIR=#{CREW_DEST_DIR}"
       system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
       system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
-      system "touch /usr/local/var/lib/dpkg/status"
+      system "touch #{CREW_DEST_PREFIX}/var/lib/dpkg/status"
     end
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -20,7 +20,7 @@ class Dpkg < Package
 
 	def self.install 
 	    system "make install DESTDIR=#{CREW_DEST_DIR}"
-	    system "echo \"# Dpkg Alias (Do NOT Touch)\" >> ~/.bashrc"
-	    system "echo \"alias dpkg=\"dpkg --root=/usr/local\"\" >> ~/.bashrc"
+	    system "echo \# Dpkg Alias (Do NOT Touch)\ >> ~/.bashrc"
+	    system "echo \alias dpkg=\"dpkg --root=/usr/local\"\ >> ~/.bashrc"
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -20,7 +20,7 @@ class Dpkg < Package
 
 	def self.install 
 	    system "make install DESTDIR=#{CREW_DEST_DIR}"
-	    system "echo \# Dpkg Alias (Do NOT Touch)\ >> ~/.bashrc"
-	    system "echo \alias dpkg=\"dpkg --root=/usr/local\"\ >> ~/.bashrc"
+	    system "echo \"# Dpkg Alias (Do NOT Touch)\" >> ~/.bashrc"
+	    system "echo \"alias dpkg=\"dpkg --root=/usr/local\"\" >> ~/.bashrc"
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -20,7 +20,7 @@ class Dpkg < Package
 
 	def self.install
 	    system "make install DESTDIR=#{CREW_DEST_DIR}"
-	    system "mkdir -p /usr/local/usr/"
-	    system "ln -s /usr/local/ /usr/local/usr/local"
+	    system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
+	    system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
   end
 end
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -10,21 +10,21 @@ class Dpkg < Package
     depends_on 'icu4c' => :build
 
     def self.build
+	    system "ARCH = `uname -m`.strip"
+            system "ARCH_LIB = if ARCH == 'x86_64' then 'lib64' else 'lib' end"          
+            system "CREW_LIB_PREFIX = /usr/local + '/' + ARCH_LIB"
 	    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
 	Dir.chdir ("dpkg")
 	    system "git checkout 1.19.0.5"
 	    system "autoreconf -i -f"
-	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
+	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=/usr/local"
 	    system "make"
 	end
 
-	def self.install 
-	    system "make install DESTDIR=#{CREW_DEST_DIR}"
+	def self.install
+	    system "make install DESTDIR=/usr/local/${ARCH_LIB}"
 	    system "ln -s /usr/ /usr/local/usr"
-	    system "mkdir -p ~/.config/fish"
 	    system "mkdir -p /usr/local/var/lib/dpkg"
-	    system "echo \"# Dpkg Alias (Do NOT Touch)\" >> ~/.{bashrc,zshrc,config/fish/config.fish}"
-	    system "echo \"alias dpkg='dpkg --root=/usr/local'\" >> ~/.{bashrc,zshrc,config/fish/config.fish}"
 	    system "touch /usr/local/var/lib/dpkg/status"
   end
 end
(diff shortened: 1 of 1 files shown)


uberhacker:
@JL2210:  Thank you for the new package, however, why do we need this?  Chromebrew is already a package manager.  Why do we need Debian's package manager app also?

JL2210:
Isn't Chromebrew based off of Homebrew? Homebrew has dpkg. Also, there are a wide range of packages that are not available with Chromebrew that would be accessible with dpkg.

JL2210:
I also am making the package manager more user-friendly and have a beta apt.rb, which I believe would help.

uberhacker:
I imagine you need a LOT more than just dpkg in order to get Debian packages installed and working on a Chromebook.  With that said, wouldn't Debian replace Chromebrew as the package manager in that case?  Sorry, I guess I'm not seeing the value-add here.

JL2210:
No, dpkg would be more like an add-on to Chromebrew. Just something that you could use if Chromebrew didn't have a package you wanted. I tested it, and after fixing some dependency issues i got checkinstall working. Not in any way am I saying dpkg would be required.

uberhacker:
The dpkg app is useless without packages.  You would still need apt utilities to search and install packages from Debian repositories.  I would definitely need to research this further.  I imagine several Debian packages would possibly conflict with Chromebrew package installs.  Chromebrew probably aligns more with Linuxbrew than Homebrew.  Linuxbrew does a lot of things "incorrectly" and has a philosophy that conflicts with the way I personally think a package manager should work.  The most fundamental thing the project gets wrong is they require you to have build tools available to get Linuxbrew itself installed.  In order to have build tools available, you need to first have a package manager to install them which completely defeats the purpose.

JL2210:
Since it is built from source, theoretically it can install Ubuntu packages too. I downloaded the checkinstall package and its first line of dependencies from packages.debian.org and it worked fine. As long as you don't install libc6 or a package that is currently installed with Chromebrew, you should be fine. Apt will be coming. I will include a dummy package with it so libc6 does not accidentally get installed. Chrome OS already has a libc6 alternative from out of box, and libc6 can break some features, such as sudo, in my case.

JL2210:
The only think stopping me from building apt is a file called Dpkg.pm that cmake cannot seem to find.

uberhacker:
Debian/Ubuntu...potatos/potatoes.  I noticed Linuxbrew has dpkg also but they don't have apt utilities.  Not very useful, I'm sure.  Linuxbrew is total trash (sorry, just my personal opinion).  I don't understand why it is so popular.  They basically just forked Homebrew without any thought or planning.  I'm glad you are thinking about ways to improve Chromebrew.  I just wish there was an easier way to implement these third-party packages more elegantly.  That is the really the only thing holding me back from accepting this Debian stuff.

JL2210:
I'm not offended about your opinion on Linuxbrew, never used it personally, but what exactly do you mean by implementing "these third-party packages more elegantly"?

JL2210:
Oh. Dpkg handles that itself. After a system reboot, all the files that were overwritten by the files in the packages are restored from the secondary root partition on every Chromebook.

JL2210:
The packages don't automatically uninstall on every system reboot.

JL2210:
If I edit the dpkg installer to create an alias for dpkg that sets the root directory to /usr/local, will that work? I just realized that it is set to install everything in /, which wouldn't work, if I am correct. Can I merge a pull request with another pull request?

uberhacker:
Yeah, all the good stuff is installed in `/usr/local` (aka `#{CREW_PREFIX}`).  You won't need to do any permission stuff since Chromebrew install already handles that.  You can simply push another commit to your PR and it will be included.

JL2210:
Explain that last bit? I am new to git.

uberhacker:
Just commit and push on the same branch as you used to create the PR.

JL2210:
Got it. Just did. Thanks.

JL2210:
Do I need single quotes if I am using the system command and need quotes inside it?

uberhacker:
If you use single quotes, anything like variables (eg. `#{CREW_PREFIX}`) that you want evaluated with be literals instead.  Use double quotes if you want the variables evaluated.

JL2210:
I am using /usr/local, so that won't be an issue. Anything else I might need to change before this gets merged?

uberhacker:
I will make comments.   I have asked the other collaborators what their thoughts are about this because I'm a bit apprehensive about merging this myself.

JL2210:
Ok. Thank you for your time.

packages/dpkg.rb
```diff
@@ -0,0 +1,38 @@
+require 'package'
+
+class Dpkg < Package
+  description 'dpkg, a medium-level package manager for Debian'
```
uberhacker: You can remove `dpkg,` from the description since `crew search dpkg` will generate `dpkg: dpkg, a medium-level...` which is a bit redundant.

packages/dpkg.rb
```diff
@@ -0,0 +1,38 @@
...
+    depends_on 'automake' => :build
+    depends_on 'make' => :build
+    depends_on 'gcc' => :build
+    depends_on 'linuxheaders' => :build
+    depends_on 'flex' => :build
+    depends_on 'bison' => :build
+    depends_on 'icu4c' => :build
+    depends_on 'gettext' => :build
+    depends_on 'autoconf' => :build
+    depends_on 'libtool' => :build
```
uberhacker: You won't need to include any of these dependencies since they're already included in Chromebrew "core".

packages/dpkg.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def self.build
+	    system "sudo mount -o remount,rw -io exec ~/"
```
uberhacker: Why do you need to do this?  What are we installing in the user home directory that needs to be executable?

JL2210:
./configure

JL2210:
Look over icu4c, flex, linuxheaders, bison and the others again. Those weren't automatically installed.

packages/dpkg.rb
```diff
@@ -0,0 +1,38 @@
...
+	    system "echo 'Thank you for entering your sudo password. It will be used to run the configure script.'"
+	    system "rm -rf {*,.*}"
+	    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
```
uberhacker: If you plan on using git, you really should be using a tag instead of cloning from the master branch.  This will make the package versionable.  What is wrong with just using the tarball source instead?  You wouldn't need the `rm -rf ...` line then.

JL2210:
Ok. I will remove "dpkg, "

JL2210:
The tarball source fails to get the version from the .git folder in the Git repository. Any suggestions on what I could do to fix that?

packages/dpkg.rb
```diff
@@ -0,0 +1,38 @@
...
+	    Dir.chdir ("dpkg")
+	    system "autoreconf -i -f"
+	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
+	    system "make"
+	end
```
uberhacker: Can you please fix the indentation so the `Dir.chdir ...` and `end` are in the same column?

JL2210:
The only branch is master.

JL2210:
Will fix.

packages/dpkg.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    def self.install 
+	    system "make install DESTDIR=#{CREW_DEST_DIR}"
+	    system "alias dpkg='dpkg --root=/usr/local'"
+  end
```
uberhacker: Please add spaces so the `end` and `def self.install` are in the same column.

uberhacker:
Can you please reply on the comments in the Files changed section instead of in the main comment section?  I will need to research if a version of the source is available.  Regarding the need to mount ~ executable, that won't be necessary if you install in `/usr/local/tmp` instead.

JL2210:
Oh, yeah. I just copied the shell script I made for myself into the ruby file. I will fix that.

packages/dpkg.rb
uberhacker: OK, I will give you `icu4c` as a necessary dependency but all others are not necessary.  Also, `gcc` has been replaced by `gcc7`.  See https://github.com/skycocker/chromebrew/blob/master/tools/core_packages.txt for reference.

JL2210:
Thanks. I will look at that file.

JL2210:
I am ready to merge, if you decide to. Check over my code just in case, please.

JL2210:
Wait, need to change something.

packages/dpkg.rb
```diff
@@ -0,0 +1,26 @@
...
+	Dir.chdir ("dpkg")
+	    system "autoreconf -i -f"
+	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
+	    system "make"
+	end
+
+	def self.install 
+	    system "make install DESTDIR=#{CREW_DEST_DIR}"
+	    system "echo \"# Dpkg Alias (Do NOT Touch)\" >> ~/.bashrc"
+	    system "echo \"alias dpkg='dpkg --root=/usr/local'\" >> ~/.bashrc"
```
uberhacker: I don't think you need the alias.  It should install in `/usr/local` by default.

JL2210:
I want the dpkg packages to install to /usr/local/ as the root directory.
I'm not entirely sure, so I want to be safe with that.

uberhacker:
I tried installing this and couldn't get it working.  My guess is there is some more configuration necessary.  I reworked the package a bit and this is what I came up wtih:
```
require 'package'

class Dpkg < Package
  version '1.19.0.5'
  source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.19.0.5/dpkg-1.19.0.5.tar.gz'
  source_sha256 'd38308afcd5d7896bbd1f946875b90f9d8510a8a96b44e4f14e781285e5d9641'

  depends_on 'icu4c' => :build

  def self.build
    system 'git clone https://salsa.debian.org/dpkg-team/dpkg.git'
    Dir.chdir 'dpkg' do
      system 'git checkout 1.19.0.5'
      system 'autoreconf -i -f'
      system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
      system 'make'
    end
  end

  def self.install
    Dir.chdir 'dpkg' do
    end
  end
end
```
You don't need to `rm -rf` since the git clone creates a separate directory.  Once the package is installed this working directory goes away anyway.
Output after install:
```
$ dpkg -l
dpkg-query: error: failed to open package info file '/usr/local/var/lib/dpkg/status' for reading: No such file or directory
```

JL2210:
Try to install a package other than libc6. That automatically creates that
file.

JL2210:
I will remove the rm -rf line.

packages/dpkg.rb
uberhacker: You probably don't want to force users into using bash for their shell either.  I try and create bash scripts instead of aliases when I can.

JL2210:
The default shell for Chrome OS is bash.

uberhacker:
Make sure you add the `git checkout ...` line also since we want the version and not the latest master branch.

uberhacker:
Right, the default shell is bash but some users might prefer fish or zsh.  In those cases, the alias won't work.

JL2210:
What would the syntax be for git checkout?

uberhacker:
Just like in my example package output above.

JL2210:
How would I make the alias compatible with all types of shell users? Also, I do not believe I can use git checkout to clone a git repo.

uberhacker:
Let me simplify it a bit.  You need an alias or bash shell script. :)

JL2210:
Yes. I want an alias. Bash is always executed upon entering shell in crosh. Wouldn't that always execute bashrc and add the alias upon typing shell in crosh?

uberhacker:
Dude, you don't need the alias.  Just don't create it.

JL2210:
So what would I do? Could you help me with the shell script?

uberhacker:
You don't need a shell script either.

JL2210:
Without the shell script or the alias, package files would install onto the root directory, which most Chromebook users, whether in devmode or not, do not have read write access to.

uberhacker:
Where are you getting your packages and what version of Debian have you tested this with?

JL2210:
I am running on FlintOS and I am getting my packages from packages.debian.org. I am installing Stretch packages.

JL2210:
I am building the source code from salsa.debian.org.

uberhacker:
I don't know anything about Flint OS but I suspect it already has a bunch of extra dependencies (probably Debian) installed that Chrome/Chromium OS doesn't have out-of-the-box.  OK, I tried this and it failed:
```
$ wget http://ftp.us.debian.org/debian/pool/main/a/alien/alien_8.95_all.deb
$ sudo dpkg -i alien_8.95_all.deb 
Password: 
dpkg: error: failed to open package info file '/usr/local/var/lib/dpkg/status' for reading: No such file or directory
```

JL2210:
I will add a line to sudo touch that file once I'm done adding the alias to ~/.bashrc, ~/.zshrc, and ~/.config/fish/config.fish

JL2210:
Flint OS is an old build of Chromium with supposed Android app support. Hint: It doesn't. It has now been aquired by Neverware, the makers of Cloudready. It is just like any old build of Chrome OS.

JL2210:
That suit your needs?

uberhacker:
I think you are going down a very slippery slope with the alias thing.  Please just leave it alone.

JL2210:
The alias is required to have dpkg install the package data to /usr/local instead of /

JL2210:
No alias required. Thanks for trying to warn me. What are the actual paths for the environment variables CREW_LIB_PREFIX, CREW_PREFIX, and CREW_INSTALL_DIR?

uberhacker:
Take a look at https://github.com/skycocker/chromebrew/blob/master/lib/const.rb.

JL2210:
Don't merge that, doesn't work. Will work on it again in the morning.

skiman6010:
Another maintainer jumping in. Just to be transparent, I don't like that we wrap pip and cpan commands already. I think it's redundant. Now dpkg is interesting because it's not necessarily a package manager in the same sense as Aptitude or apt. I feel like the Chromebrew project more mimics a traditional package manager where everything is tested and working well before pushed to users. Dpkg is more manual than that in my opinion. I think at that point, the user is in charge of dependencies and deciding which architecture to install. I don't think it would hurt anything to give an option to use dpkg to our users.

I would say that as long as dpkg is working, we should merge it and see what happens. If it breaks too much or is too "beta", we can pull it and decide on another solution.

JL2210:
Thanks. Please wait, as I currently need to make it more architecture
independent. I am considering splitting it into two packages, dpkg32 and
dpkg64, to fix the possibility of it breaking because of shared library
locations.

JL2210:
Another bug I am currently working on is changing the directory to install
packages from the root directory to /use/local. I have already added a
symlink from /usr/local/usr to /usr.

skiman6010:
@JL2210 Why does it need to be more architecture dependent? 

Could you use a CASE statement like we have in the miniconda package? https://github.com/skycocker/chromebrew/blob/master/packages/miniconda3.rb

JL2210:
The lib and lib64 directories for 32 and 64 bit systems, respectively.

JL2210:
Case would probably be appropriate. I will look into that. Thank you.

uberhacker:
I imagine if you attempted to install a *-i386* deb package on an arm system, it would likely go belly up.

JL2210:
You're right, but that's not what I'm implying. They use the same
directories in /usr/local, correct?

On May 7, 2018 4:26 PM, "Ed Reel" <notifications@github.com> wrote:

I imagine if you attempted to install a *-i386* deb package on an arm
system, it would likely go belly up.

—
<https://github.com/skycocker/chromebrew/pull/2327#issuecomment-387194050>,
or mute the thread

uberhacker:
We have the `#{CREW_LIB_PREFIX}` constant that abstracts the location.

JL2210:
In other words, dpkg wouldn't be able to decipher the constant. Therefore,
we need the absolute path.

uberhacker:
You're not using a ruby package?

uberhacker:
I think you are overthinking this process.  Just compile dpkg for the architecture you want it installed on and wash your hands.

JL2210:
Dpkg isn't ruby. I want to make sure i686, arm and aarch64 Chromebook users
can use dpkg as well. The variables for ruby and crew do not work for me,
let alone with dpkg, on my system.

skiman6010:
I recommend we just get it working on one architecture, then proceed from there.

skiman6010:
@JL2210 Crew's installation system uses the abstract paths to put packages in places. If you install dpkg with the crew command and tell it the installation path is #{CREW_LIB_PREFIX} it will be translated to the location we want it to be installed. I don't understand the problem here.

uberhacker:
Correct me if I'm wrong here @JL2210 but I think you want dpkg configured so that it installs packages into `/usr/local` (aka `#{CREW_PREFIX}`) instead of `/usr`?  The problem isn't with installing dpkg via crew but rather configuration and installation of Debian packages, right?

JL2210:
No, /usr/local instead of the Chrome OS root directory, which most
Chromebook users do not have write access to.

uberhacker:
Right, but if you get everything that would normally be installed in `/usr` from a Debian package moved over into `/usr/local` instead, that would be the solution, correct?  I don't think you want `/usr/local/usr/...`.

JL2210:
That's a symlink required for DPKG to be installed in /usr/local. Dpkg
requires its configuration directories to be installed in /usr/local. So, I
symlinked /usr to /usr/local/usr. I also added that to dpkg.rb.

uberhacker:
Did you research how Homebrew configures dpkg?  I imagine it would be very similar in Chromebrew.  See https://github.com/Homebrew/homebrew-core/blob/8e1ba262383f14c23bc37bfef59ebc66a46a416e/Formula/dpkg.rb.

JL2210:
Thanks. I will look at that. I imagine Macs don't have ARM chips, though...

JL2210:
Look at this:
# Please always keep the Homebrew mirror as the primary URL as the
# dpkg site removes tarballs regularly which means we get issues
# unnecessarily and older versions of the formula are broken.

JL2210:
They don't configure dpkg to install packages. I think that's a necessary
feature. What's the Linuxbrew configuration of dpkg?

JL2210:
Could I resume work on this on the weekend? I need a little time to think
how this would work.

uberhacker:
@JL2210: Absolutely.  There is no rush on this.  I agree with you that it doesn't make sense to have the dpkg app if you can't even use it.   Please don't get me started again about what I think about that project. :)  Apparently, they just reference the Homebrew project directly with dpkg since it's not in the Linuxbrew repository.

JL2210:
Thanks. Linuxbrew does have dpkg.

JL2210:
Any way to run scripts when uninstalling? Right now I have a dirty symlink from /usr/ to /usr/local/usr that would be better to remove than to leave there when uninstalling.

uberhacker:
@JL2210: Just add the symlink in when you install the package.  As long as the file is tracked, it should uninstall fine.

JL2210:
OK. Thanks.

JL2210:
Sorry for waking you up this late at night. Just about done.

uberhacker:
No worries at all.  I'm on CST time so it's not super late yet.

JL2210:
OK. Thanks.

JL2210:
Is the new change showing up? I re-forked to update my side of the repository.

JL2210:
Yes it is. I think I'm ready. Any final suggestions?

JL2210:
It's 12 here, so I'll get back to you in the morning on this. I'm ready.

packages/dpkg.rb
```diff
@@ -0,0 +1,26 @@
...
+	    system "git checkout 1.19.0.5"
+	    system "autoreconf -i -f"
+	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
+	    system "make"
+	end
+
+	def self.install
+	    system "make install DESTDIR=#{CREW_DEST_DIR}"
+	    system "mkdir -p /usr/local/usr/"
+	    system "ln -s /usr/local/ /usr/local/usr/local"
```
uberhacker: Is this the symlink you wanted removed on uninstall?  If so, create the symlink like this instead:
```
system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
```

uberhacker:
`#{CREW_PREFIX}` is `/usr/local` by default so yes unless `#{CREW_PREFIX}` is changed.

JL2210:
It's fine. It works. Anything else? I think that's it.

packages/dpkg.rb
```diff
@@ -0,0 +1,26 @@
...
+	    system "autoreconf -i -f"
+	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
+	    system "make"
+	end
+
+	def self.install
+	    system "make install DESTDIR=#{CREW_DEST_DIR}"
+	    system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
+	    system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
+  end
```
uberhacker: Can you please clean up the indentation a bit to avoid confusion?

packages/dpkg.rb
```diff
@@ -0,0 +1,26 @@
...
+  homepage 'https://anonscm.debian.org/git/dpkg/'
+  version '1.19.0.5'
+  source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.19.0.5/dpkg-1.19.0.5.tar.gz'
+  source_sha256 'd38308afcd5d7896bbd1f946875b90f9d8510a8a96b44e4f14e781285e5d9641'
+
+    depends_on 'icu4c' => :build
+
+    def self.build
+    	    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
+	Dir.chdir ("dpkg")
```
uberhacker: You will need `Dir.chdir ("dpkg") do`.

packages/dpkg.rb
```diff
@@ -0,0 +1,26 @@
...
+	    system "git checkout 1.19.0.5"
+	    system "autoreconf -i -f"
+	    system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
+	    system "make"
+	end
+
+	def self.install
```
uberhacker: You will need `Dir.chdir ("dpkg") do` here also, I believe.

JL2210:
Where, exactly?

packages/dpkg.rb
uberhacker: Right below where this comment was inserted.

JL2210:
        13: from /usr/local/bin/crew:911:in `<main>'
        12: from /usr/local/bin/crew:853:in `install_command'
        11: from /usr/local/bin/crew:853:in `each'
        10: from /usr/local/bin/crew:855:in `block in install_command'
         9: from /usr/local/bin/crew:142:in `search'
         8: from /usr/local/lib64/ruby/2.5.0/find.rb:43:in `find'
         7: from /usr/local/lib64/ruby/2.5.0/find.rb:43:in `each'
         6: from /usr/local/lib64/ruby/2.5.0/find.rb:48:in `block in find'
         5: from /usr/local/lib64/ruby/2.5.0/find.rb:48:in `catch'
         4: from /usr/local/lib64/ruby/2.5.0/find.rb:49:in `block (2 levels) in find'
         3: from /usr/local/bin/crew:143:in `block in search'
         2: from /usr/local/bin/crew:105:in `set_package'
         1: from /usr/local/lib64/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/usr/local/lib64/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require': /usr/local/lib/crew/packages/dpkg.rb:27: syntax error, unexpected end-of-input, expecting keyword_end (SyntaxError)

JL2210:
"do" is what's causing it.

packages/dpkg.rb
uberhacker: Interesting.  I usually have the opposite happen.

uberhacker: What is your Chromebrew version and system specs?

JL2210:
[uname.txt](https://github.com/skycocker/chromebrew/files/1997643/uname.txt)
[lsb-release.txt](https://github.com/skycocker/chromebrew/files/1997644/lsb-release.txt)

JL2210:
How do I find the chromebrew version?

JL2210:
Does [this](https://github.com/skycocker/chromebrew/files/1997649/crew.txt) help?

uberhacker:
Please just paste the output in the comments instead of in a file.  I have to download and then open to see the results.

uberhacker:
The Chrome OS version is in the output of `cat /etc/lsb-version`.

JL2210:
cat: /etc/lsb-version: No such file or directory

uberhacker:
Sorry, `cat /etc/lsb-release`.

JL2210:
CHROMEOS_RELEASE_APPID={1A14631E-DE46-4330-91F4-80D10EE3692C}
CHROMEOS_BOARD_APPID={1A14631E-DE46-4330-91F4-80D10EE3692C}
DEVICETYPE=OTHER
CHROMEOS_RELEASE_BOARD=amd64-flint
CHROMEOS_DEVSERVER=
GOOGLE_RELEASE=9901.80.4
CHROMEOS_RELEASE_BUILD_NUMBER=9901
CHROMEOS_RELEASE_BRANCH_NUMBER=80
CHROMEOS_RELEASE_CHROME_MILESTONE=62
CHROMEOS_RELEASE_PATCH_NUMBER=4
CHROMEOS_RELEASE_TRACK=stable-channel
CHROMEOS_RELEASE_DESCRIPTION=
CHROMEOS_RELEASE_VERSION=9901.80.4
CHROMEOS_AUSERVER=https://up.flintos.xyz/service/update2
uname -a:
Linux localhost 4.4.64-09104-gffd9a4db8e11 #7 SMP Tue Dec 19 21:19:20 CST 2017 x86_64 Pentium(R) Dual-Core CPU T4200 @ 2.00GHz GenuineIntel GNU/Linux

uberhacker:
I believe you are missing an `end` line.  This is why I asked you to clean up the indentation to avoid confusion.  Please remove all tabs and replace with 2 spaces instead.  Once you align everything, you will notice the missing `end` line.

JL2210:
Like this?
require 'package'

class Dpkg < Package
  version '1.19.0.5'
  source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.19.0.5/dpkg-1.19.0.5.tar.gz'
  source_sha256 'd38308afcd5d7896bbd1f946875b90f9d8510a8a96b44e4f14e781285e5d9641'

    depends_on 'icu4c' => :build

    def self.build
          system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
  Dir.chdir ("dpkg")
      system "git checkout 1.19.0.5"
      system "autoreconf -i -f"
      system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
      system "make"
  end

  def self.install
      Dir.chdir ("dpkg")
      system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
      system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
  end
end

uberhacker:
Like this:
```
equire 'package'

class Dpkg < Package
  version '1.19.0.5'
  source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.19.0.5/dpkg-1.19.0.5.tar.gz'
  source_sha256 'd38308afcd5d7896bbd1f946875b90f9d8510a8a96b44e4f14e781285e5d9641'

  depends_on 'icu4c' => :build

  def self.build
    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
    Dir.chdir ("dpkg") do
      system "git checkout 1.19.0.5"
      system "autoreconf -i -f"
      system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
      system "make"
    end
  end

  def self.install
    Dir.chdir ("dpkg") do
      system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
      system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
    end
  end
end
```

JL2210:
Hang on. Let me commit this so you can better understand.

uberhacker:
I tested your package as corrected above and these were the results:
```
chronos@localhost ~/chromebrew/packages (dpkg *%=)$ dpkg -l
dpkg-query: error: failed to open package info file '/usr/local/var/lib/dpkg/status' for reading: No such file or directory
chronos@localhost ~/chromebrew/packages (dpkg *%=)$ mkdir -p /usr/local/var/lib/dpkg
chronos@localhost ~/chromebrew/packages (dpkg *%=)$ touch /usr/local/var/lib/dpkg/status
chronos@localhost ~/chromebrew/packages (dpkg *%=)$ wget http://ftp.us.debian.org/debian/pool/main/b/bats/bats_0.4.0-1.1_all.deb
--2018-05-12 14:11:27--  http://ftp.us.debian.org/debian/pool/main/b/bats/bats_0.4.0-1.1_all.deb
Resolving ftp.us.debian.org... 208.80.154.15, 128.30.2.26, 128.61.240.89, ...
Connecting to ftp.us.debian.org|208.80.154.15|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 14294 (14K) [application/octet-stream]
Saving to: ‘bats_0.4.0-1.1_all.deb’

bats_0.4.0-1.1_all.deb       100%[===========================================>]  13.96K  --.-KB/s    in 0.07s   

2018-05-12 14:11:28 (212 KB/s) - ‘bats_0.4.0-1.1_all.deb’ saved [14294/14294]

chronos@localhost ~/chromebrew/packages (dpkg *%=)$ sudo dpkg -i bats_0.4.0-1.1_all.deb 
Selecting previously unselected package bats.
(Reading database ... 0 files and directories currently installed.)
Preparing to unpack bats_0.4.0-1.1_all.deb ...
Unpacking bats (0.4.0-1.1) ...
dpkg: error processing archive bats_0.4.0-1.1_all.deb (--install):
 unable to clean up mess surrounding './usr/lib/bats' before installing another version: Read-only file system
dpkg-deb (subprocess): decompressing archive member subprocess was killed by signal (Broken pipe)
dpkg-deb: error: <decompress> subprocess returned error exit status 2
Errors were encountered while processing:
 bats_0.4.0-1.1_all.deb
```
Looks like you will need to touch the status file during install or generate somehow.

JL2210:
        13: from /usr/local/bin/crew:911:in `<main>'
        12: from /usr/local/bin/crew:853:in `install_command'
        11: from /usr/local/bin/crew:853:in `each'
        10: from /usr/local/bin/crew:855:in `block in install_command'
         9: from /usr/local/bin/crew:142:in `search'
         8: from /usr/local/lib64/ruby/2.5.0/find.rb:43:in `find'
         7: from /usr/local/lib64/ruby/2.5.0/find.rb:43:in `each'
         6: from /usr/local/lib64/ruby/2.5.0/find.rb:48:in `block in find'
         5: from /usr/local/lib64/ruby/2.5.0/find.rb:48:in `catch'
         4: from /usr/local/lib64/ruby/2.5.0/find.rb:49:in `block (2 levels) in find'
         3: from /usr/local/bin/crew:143:in `block in search'
         2: from /usr/local/bin/crew:105:in `set_package'
         1: from /usr/local/lib64/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/usr/local/lib64/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require': /usr/local/lib/crew/packages/dpkg.rb:28: syntax error, unexpected keyword_end, expecting end-of-input (SyntaxError)
  end
  ^~~

JL2210:
Sorry, forgot to add "do" after Dir.chdir. Works now.

JL2210:
What are you running? Cloudready?

uberhacker:
Currently, I'm running a custom build of Chromium OS from within Virtualbox (i686 32-bit).  I also have x86_64 and armv7l Chromebooks.

JL2210:
Oh. Nice. I'm building my Chromium OS from source right now on the desktop. I'm on a laptop right now with FlintOS. I also have an x86_64 Chromebook. I'm planning to install vanilla Chromium OS on the laptop I'm on right now.

JL2210:
Success! It works!

packages/dpkg.rb
```diff
@@ -0,0 +1,30 @@
...
+      system "make"
+    end
+  end
+
+  def self.install
+    Dir.chdir ("dpkg") do
+      system "make install DESTDIR=#{CREW_DEST_DIR}"
+      system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
+      system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
+      system "touch /usr/local/var/lib/dpkg/status"
```
uberhacker: Please change to:
```
system "mkdir -p ${CREW_DEST_PREFIX}/var/lib/dpkg"
system "touch #{CREW_DEST_PREFIX}/var/lib/dpkg/status"
```
If you do this, the status file will be tracked and removed during uninstall.

JL2210:
Got it.

JL2210:
${CREW_DEST_PREFIX}/var/lib/dpkg is created during build.
#{CREW_DEST_PREFIX}/var/lib/dpkg/status is not.

uberhacker:
What about the `Read-only file system` error from my test results above?

packages/dpkg.rb
```diff
@@ -0,0 +1,30 @@
...
+      system "make"
+    end
+  end
+
+  def self.install
+    Dir.chdir ("dpkg") do
+      system "make install DESTDIR=#{CREW_DEST_DIR}"
+      system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
+      system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
+      system "touch #{CREW_DEST_PREFIX}/var/lib/dpkg/status"
```
uberhacker: You will need `system "mkdir -p #{CREW_DEST_PREFIX}/var/lib/dpkg"` above this line.

JL2210:
If I did that, every time dpkg was upgraded, the installed package list would be overwritten. I might use >> as that wouldn't overwrite the package list.

JL2210:
#{CREW_DEST_PREFIX}/var/lib/dpkg is created automatically.

uberhacker:
I don't think the touch command is that powerful.

JL2210:
I know, I will use >>

JL2210:
No. When dpkg is installed, ```make``` automatically creates /usr/local/var/lib/dpkg. It just does not touch status in that directory.

uberhacker:
Right, but `#{CREW_DEST_PREFIX}/var/lib/dpkg` is not created which is needed to track the status file.

JL2210:
Sorry, I meant that. Let me commit my changes so you can see what I've done.

uberhacker:
`#{CREW_PREFIX}/var/lib/dpkg` is not installed either.  See below:
```
sh: cannot create /usr/local/var/lib/dpkg/status: Directory nonexistent
dpkg failed to install: `echo >> /usr/local/var/lib/dpkg/status` exited with 2
```

JL2210:
Oh, wait. I see. I would need to put that in the postinstall for it to work. Let me do that.

uberhacker:
This works:
```
  def self.install
    Dir.chdir ("dpkg") do
      system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
      system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
      system "mkdir -p #{CREW_DEST_PREFIX}/var/lib/dpkg/"
      system "touch #{CREW_DEST_PREFIX}/var/lib/dpkg/status"
    end
  end
```
```
$ crew install dpkg
... snip ...
$ crew files dpkg|grep status
/usr/local/var/lib/dpkg/status
```

JL2210:
Yes, but chromebrew overwrites the original dpkg status file, thereby eliminating the status file and defeating the purpose of dpkg.

JL2210:
Install a couple of packages and try that. They'll be on your VM forever, at least until you reinstall.

uberhacker:
I got dpkg working.  This is what you need to do:
```
  def self.install
    Dir.chdir ("dpkg") do
      system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
      system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
      system "mkdir -p #{CREW_DEST_PREFIX}/var/lib/dpkg/"
      system "touch #{CREW_DEST_PREFIX}/var/lib/dpkg/status"
    end
    Dir.chdir ("#{CREW_DEST_PREFIX}/bin") do
      system 'mv dpkg dpkg-run'
      system "echo '#!/bin/bash' > dpkg"
      system "echo 'dpkg-run --root=#{CREW_PREFIX} \"$@\"' >> dpkg"
      system 'chmod +x dpkg'
    end
  end
```
```
$ crew install dpkg
... snip ...
$ wget http://ftp.us.debian.org/debian/pool/main/b/bats/bats_0.4.0-1.1_all.deb
$ sudo dpkg -i bats_0.4.0-1.1_all.deb 
Password: 
Selecting previously unselected package bats.
(Reading database ... 0 files and directories currently installed.)
Preparing to unpack bats_0.4.0-1.1_all.deb ...
Unpacking bats (0.4.0-1.1) ...
Setting up bats (0.4.0-1.1) ...
chronos@localhost ~/chromebrew/packages (dpkg *%=)$ bats
Bats 0.4.0
Usage: bats [-c] [-p | -t] <test> [<test> ...]
chronos@localhost ~/chromebrew/packages (dpkg *%=)$ dpkg -l
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                   Version          Architecture     Description
+++-======================-================-================-=================================================
ii  bats                   0.4.0-1.1        all              bash automated testing system
```

JL2210:
CREW_PREFIX is /usr/local, right?

JL2210:
I added a little of my own magic there.

JL2210:
Yes. Is it something I could do with CASE?

packages/dpkg.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+  end
+
+  def self.install
+    Dir.chdir ("dpkg") do
+      system "make install DESTDIR=#{CREW_DEST_DIR}"
+      system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
+      system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
+      system "mkdir -p #{CREW_DEST_PREFIX}/var/lib/dpkg/"
+      system "mv #{CREW_PREFIX}/var/lib/dpkg/status #{CREW_PREFIX}/var/lib/dpkg/status.old"
```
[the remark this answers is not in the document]
uberhacker:
```
system "[ -f #{CREW_PREFIX}/var/lib/dpkg/status ] && mv #{CREW_PREFIX}/var/lib/dpkg/status #{CREW_PREFIX}/var/lib/dpkg/status.old"
```

packages/dpkg.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+    Dir.chdir ("#{CREW_DEST_PREFIX}/bin") do
+      system 'mv dpkg dpkg-run'
+      system "echo '#!/bin/bash' > dpkg"
+      system "echo 'dpkg-run --force-depends --root=#{CREW_PREFIX} \"$@\"' >> dpkg"
+      system 'chmod a+x dpkg'
+    end
+  end
+    def self.postinstall
+    system "cat #{CREW_PREFIX}/var/lib/dpkg/status.old >> #{CREW_PREFIX}/var/lib/dpkg/status"
```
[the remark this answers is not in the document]
uberhacker: The Ruby :tm: way:
```
if File.exists? "#{CREW_PREFIX}/var/lib/dpkg/status.old" then
  system "cat #{CREW_PREFIX}/var/lib/dpkg/status.old >> #{CREW_PREFIX}/var/lib/dpkg/status"
  system "rm #{CREW_PREFIX}/var/lib/dpkg/status.old"
end
```

packages/dpkg.rb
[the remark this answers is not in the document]
uberhacker: After thinking about this a bit, I don't believe this is the correct location for this logic.  We have a preinstall function that would probably work better for this.

JL2210:
```😑```

uberhacker:
The reason we need the preinstall function is because of pre-built binaries that will install files statically without any conditional logic.

JL2210:
Is there something wrong with this that I can't see?
```require 'package'

class Dpkg < Package
  version '1.19.0.5'
  source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.19.0.5/dpkg-1.19.0.5.tar.gz'
  source_sha256 'd38308afcd5d7896bbd1f946875b90f9d8510a8a96b44e4f14e781285e5d9641'

  depends_on 'icu4c' => :build

  def self.build
    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git"
    Dir.chdir ("dpkg") do
      system "git checkout 1.19.0.5"
      system "autoreconf -i -f"
      system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
      system "make"
    end
  end

    def self.preinstall
  if File.exists? "#{CREW_PREFIX}/var/lib/dpkg/status" then
      system "mv #{CREW_PREFIX}/var/lib/dpkg/status #{CREW_PREFIX}/var/lib/dpkg/status.old"
    end
  def self.install
    Dir.chdir ("dpkg") do
      system "mkdir -p #{CREW_DEST_PREFIX}/usr/"
      system "ln -s #{CREW_PREFIX} #{CREW_DEST_PREFIX}/usr/local"
      system "mkdir -p #{CREW_DEST_PREFIX}/var/lib/dpkg/"
      system "touch #{CREW_DEST_PREFIX}/var/lib/dpkg/status"
    end
    Dir.chdir ("#{CREW_DEST_PREFIX}/bin") do
      system 'mv dpkg dpkg-run'
      system "echo '#!/bin/bash' > dpkg"
      system "echo 'dpkg-run --force-depends --root=#{CREW_PREFIX} \"$@\"' >> dpkg"
      system 'chmod a+x dpkg'
    end
  end
      def self.postinstall
    if File.exists? "#{CREW_PREFIX}/var/lib/dpkg/status.old" then
      system "cat #{CREW_PREFIX}/var/lib/dpkg/status.old >> #{CREW_PREFIX}/var/lib/dpkg/status"
      system "rm #{CREW_PREFIX}/var/lib/dpkg/status.old"
    end
  end
end```

JL2210:
/usr/local/lib64/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require': /usr/local/lib/crew/packages/dpkg.rb:47: syntax error, unexpected end-of-input, expecting keyword_end (SyntaxError)

uberhacker:
Yes, it appears you are missing an `end` line.  This is why proper indentation helps.  Please use 2 space indentation instead of tabs and make sure all your conditional starts and ends are lined up in the same column and that should help.

JL2210:
Thanks.

JL2210:
Anything else?

uberhacker:
I think that should do it.  I appreciate all the effort you put into this package.  The next step is to see how well the pre-built binaries work.

JL2210:
Thanks. You really helped me with this. I appreciate it.

PR #2334, opened 2018-05-08
Add krypton package; command-line client, daemon, and SSH integration

danielcbaldwin:
## Description
Krypton is a set of tools and utilities that allow you to have your ssh private key hosted on your phone and when you attempt to connect to a server a message is sent to your phone to confirm the use of your private key to initiate the ssh session. This gives you some additional security when using keys for ssh / git. https://krypt.co

Works properly:
- [x] x86_64
I dont have another architecture to test this against.

### DIFF
--- a/packages/kr.rb
+++ b/packages/kr.rb
@@ -14,6 +14,7 @@ class Kr < Package
   depends_on "psmisc" => :build
 
   def self.install
+    cargo_web_version = "0.6.10"
     cargo_web_sha256 = "c8a92ed0185206ad7f2da8e1692d168b20044de7066b2eb086e4bbb0d06ff787"
     ENV["GOPATH"] = "#{Dir.pwd}/go"
     ENV["RUSTUP_HOME"] = "#{Dir.pwd}/.rustup"
@@ -25,7 +26,10 @@ class Kr < Package
     system "rustup", "default", "stable"
     system "rustup", "target", "add", "wasm32-unknown-emscripten"
 
-    system "cargo web --version || ( curl -Ls https://github.com/koute/cargo-web/archive/0.6.10.zip -o cargo-web.zip; if [ $(sha256sum cargo-web.zip | cut -d ' ' -f1) != '#{cargo_web_sha256}' ]; then echo 'cargo-web sha did not match #{cargo_web_sha256}, exiting.'; exit 1; fi; unzip cargo-web.zip; cd cargo-web-0.6.10; cargo build --release && cargo install )"
+    system "curl -Ls https://github.com/koute/cargo-web/archive/#{cargo_web_version}.zip -o cargo-web.zip"
+    system "if [ $(sha256sum cargo-web.zip | cut -d ' ' -f1) != '#{cargo_web_sha256}' ]; then echo 'cargo-web sha did not match #{cargo_web_sha256}, exiting.'; exit 1; fi"
+    system "unzip cargo-web.zip && cd cargo-web-#{cargo_web_version}"
+    system "cargo build --release && cargo install"
 
     system "mkdir -p #{Dir.pwd}/go/src"
     system "go get github.com/kryptco/kr"
--- a/packages/kr.rb
+++ b/packages/kr.rb
@@ -38,4 +38,11 @@ class Kr < Package
       system "make", "install"
     end
   end
+  
+  def self.postinstall
+    puts "============================"
+    puts "Krypton is now installed!"
+    puts "'kr pair' will automatically configure ~/.ssh/config to use Krypton. To prevent this, export KR_SKIP_SSH_CONFIG=1"
+    puts "============================"
+  end
 end
--- a/packages/kr.rb
+++ b/packages/kr.rb
@@ -7,15 +7,11 @@ class Kr < Package
   source_url 'https://raw.githubusercontent.com/kryptco/kr/master/README.md'
   source_sha256 'a0452783d22218a6fab914abbf5263459aaa1c8745512d7c4ebcf51fe08aaf7a'
 
-  depends_on "gcc7" => :build
   depends_on "rust" => :build
   depends_on "go" => :build
-  depends_on "pkgconfig" => :build
   depends_on "libsodium" => :build
   depends_on "rsync" => :build
-  depends_on "buildessential" => :build
   depends_on "psmisc" => :build
-  depends_on "git" => :build
 
   def self.install
     ENV["GOPATH"] = "#{Dir.pwd}/go"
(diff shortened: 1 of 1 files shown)


packages/kr.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  depends_on "gcc7" => :build
+  depends_on "rust" => :build
+  depends_on "go" => :build
+  depends_on "pkgconfig" => :build
+  depends_on "libsodium" => :build
+  depends_on "rsync" => :build
+  depends_on "buildessential" => :build
+  depends_on "psmisc" => :build
+  depends_on "git" => :build
```
uberhacker: You can remove gcc7, pkgconfig, buildessential and git dependencies since they're already included in Chromebrew "core".

packages/kr.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def self.install
+    ENV["GOPATH"] = "#{Dir.pwd}/go"
+
+    system "rustup", "target", "add", "wasm32-unknown-emscripten"
+    system "cargo web --version || ( curl -Ls https://github.com/koute/cargo-web/archive/0.6.10.zip -o cargo-web.zip; unzip cargo-web.zip; cd cargo-web-0.6.10; cargo build --release && cargo install )"
```
uberhacker: Should we have a checksum on 0.6.10.zip?

packages/kr.rb
danielcbaldwin: Removed as you suggested.

packages/kr.rb
```diff
@@ -0,0 +1,48 @@
...
+      ENV["PREFIX"] = "#{CREW_PREFIX}"
+      ENV["PWD"] = Dir.pwd
+      system "git checkout tags/2.4.4"
+      system "git submodule update --init --recursive"
+      system "git checkout -- Makefile"
+      system "sed -i '/^LINK_LIBSIGCHAIN_LDFLAGS/ s/$/ -fstack-protector/' Makefile"
+      system "sed -i '/^DSTBIN =/ s/.*/DSTBIN = #{CREW_DEST_PREFIX.gsub('/', '\/')}\\/bin/' Makefile"
+      system "sed -i '/^DSTLIB =/ s/.*/DSTLIB = #{CREW_DEST_PREFIX.gsub('/', '\/')}\\/lib/' Makefile"
+      system "make"
+      system "make", "install"
```
uberhacker: Is this sufficient to install and track everything this package requires?  Just wondering if we're not tracking the rustup and cargo line stuff above.  In other words, does everything that gets installed get removed with `crew remove kr`?

uberhacker:
Thank you for the new package @danielcbaldwin!  Please see inline comments.

packages/kr.rb
danielcbaldwin: I added in the checksum check.

packages/kr.rb
danielcbaldwin: The answer is yes and no. Krypton and all of it's dependent files are tracked and removed, I have checked that multiple times, however some of the build environment stuff does not get tracked or removed. For instance cargo-web or the rustup target. I dont know of a way to handle that and I dont see any evidence of that ever being handled in the homebrew tap https://github.com/kryptco/homebrew-tap/blob/master/kr.rb or the aur package https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=kr both of which I used to get this going.

uberhacker: Can you specify a target directory for cargo and rustup?  If you install everything relative to `#{CREW_DEST_DIR}`, that may be the ticket.  By the way, just because homebrew and aur does it doesn't mean we have to do it the same way.  Package managers should do a better job of cleaning up the residue after an uninstall.

danielcbaldwin: Did some quick research and testing and now everything for rustup and cargo is sandboxed in the build directory and doesnt change anything for the user.

packages/kr.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def self.install
+    cargo_web_sha256 = "c8a92ed0185206ad7f2da8e1692d168b20044de7066b2eb086e4bbb0d06ff787"
+    ENV["GOPATH"] = "#{Dir.pwd}/go"
+    ENV["RUSTUP_HOME"] = "#{Dir.pwd}/.rustup"
+    ENV["CARGO_HOME"] = "#{Dir.pwd}/.cargo"
+    system "mkdir -p #{ENV["RUSTUP_HOME"]}"
+    system "mkdir -p #{ENV["CARGO_HOME"]}"
+
+    ENV["PATH"] = "#{ENV["CARGO_HOME"]}/bin:/usr/local/bin:/usr/bin:/bin:/opt/bin"
```
uberhacker: Why do you need `#{ENV["CARGO_HOME"]/bin` in the path?

packages/kr.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    system "rustup", "install", "stable"
+    system "rustup", "default", "stable"
+    system "rustup", "target", "add", "wasm32-unknown-emscripten"
+
+    system "cargo web --version || ( curl -Ls https://github.com/koute/cargo-web/archive/0.6.10.zip -o cargo-web.zip; if [ $(sha256sum cargo-web.zip | cut -d ' ' -f1) != '#{cargo_web_sha256}' ]; then echo 'cargo-web sha did not match #{cargo_web_sha256}, exiting.'; exit 1; fi; unzip cargo-web.zip; cd cargo-web-0.6.10; cargo build --release && cargo install )"
```
uberhacker: Can this long line be split into separate lines for readability?

packages/kr.rb
uberhacker: So basically rustup and cargo stuff is only needed for the build?

packages/kr.rb
danielcbaldwin: Removed it, I checked and all works without it. In trying to get this up and running I had come across something that indicated the need for that for cargo to work. So I had that in my path so I was just trying to replicate that for the sandboxed version. More evidence of me not knowing anything about rust and cargo. Thanks for catching that.

packages/kr.rb
danielcbaldwin: Yes, I can look at switching it to heredoc. I had it this way so that it could be skipped in the case that you had it already but that is unlikely now with it sandboxed so likely this can just be all separate system calls now.

packages/kr.rb
danielcbaldwin: That I am aware of and can figure out, that is true.

packages/kr.rb
danielcbaldwin: Done

uberhacker:
I'll go ahead and merge this and we can address issues in updates going forward.  Nice work!

PR #2340, opened 2018-05-09
Update crew

lovedheart:


crew
```diff
@@ -501,6 +501,13 @@ def prepare_package (destdir)
     compress_doc "#{destdir}#{CREW_PREFIX}/share/man"
     compress_doc "#{destdir}#{CREW_PREFIX}/share/info"
 
+    # Remove *.la before archiving packages
+    if @opt_verbose then
+      system "find #{CREW_DEST_LIB_PREFIX} -type f -name *.la -delete"
+    else
+      system "find #{CREW_DEST_LIB_PREFIX} -type f -name *.la -delete"
+    end
```
uberhacker: Isn't this no longer an issue with the renaming logic?  What does it hurt leaving these files?

lovedheart: We need these files?

uberhacker: Don't we get 404 errors if these files are missing sometimes?

lovedheart: ?

lovedheart: Renaming them before compiling code == *.la files missing during compilation

uberhacker: Right, the renaming logic fixes the libtool error issue but what about other packages that need these files as dependencies?

uberhacker: How about delete the *.la files before and after builds?  Then the renaming is unnecessary.  Also, how does the verbose option change the output to be more descriptive?

lovedheart: The renaming is a comprimised method. The scope is to delete *.la files in all packages. But it should require recompiling all the packages. What I modified is to make sure the later compiled package contains no *.la files and it does also reduce the work to recompile all packages.  If we upgrade all the packge once, there are no *.la files any more.

uberhacker: Right, but it makes no sense if you rename and then delete what you just renamed.  Why not just delete and be done with it?  We are making this more complicated than it needs to be.  If you want me to merge, please rework it to be more efficient and more descriptive when the verbose option is provided.

lovedheart: I don't delete what I have renamed. I delet only the compiled ones.

uberhacker: Not sure I follow.  Why don't we just delete all the *.la files if they aren't needed?  This would require 2 steps instead of 3.  Delete all the *.la files before compiling and then delete all the *.la files after compiling.  This doesn't solve the issue of packages that are expecting the *.la files in the builds and erroring out when removing them but it moves us closer to having all the packages as lean as possible going forward.  We can recompile as needed or when packages have updates.  I need more help in that task, by the way. :)

lovedheart: But what is the problem of my mod? Is there any issue that might occur ? Or something so complicated that nobody understands?

PR #2341, opened 2018-05-09
Update nano.rb

lovedheart:


packages/nano.rb
```diff
@@ -3,34 +3,43 @@
...
-    aarch64: 'c2c3367cab170f5561bda70d0a50b6a214f552133dc9b3f927484167b761b5ae',
-     armv7l: 'c2c3367cab170f5561bda70d0a50b6a214f552133dc9b3f927484167b761b5ae',
-       i686: '63c0d80e9d45e0ce5985d8d15d10a2bc5945d6ac86cbb91d545f5516a1c36539',
-     x86_64: '680533f195e883574b6591bcdba703e1ae36d01066e93f5f3eaacb104ef26702',
-  })
+  version '2.9.5'
+  source_url 'https://www.nano-editor.org/dist/v2.9/nano-2.9.5.tar.xz'
+  source_sha256 '7b8d181cb57f42fa86a380bb9ad46abab859b60383607f731b65a9077f4b4e19'
 
   depends_on 'ncurses'
```
uberhacker: This line can be removed since ncurses is in Chromebrew "core".

packages/nano.rb
```diff
@@ -3,34 +3,43 @@
...
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/nano-2.9.1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/nano-2.9.1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/nano-2.9.1-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: 'c2c3367cab170f5561bda70d0a50b6a214f552133dc9b3f927484167b761b5ae',
-     armv7l: 'c2c3367cab170f5561bda70d0a50b6a214f552133dc9b3f927484167b761b5ae',
-       i686: '63c0d80e9d45e0ce5985d8d15d10a2bc5945d6ac86cbb91d545f5516a1c36539',
-     x86_64: '680533f195e883574b6591bcdba703e1ae36d01066e93f5f3eaacb104ef26702',
-  })
```
uberhacker: Not adding binaries anymore? ;)

packages/nano.rb
```diff
@@ -3,34 +3,43 @@
...
+           "--enable-utf8"
     system 'make'
   end
 
   def self.install
-    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install-strip'
+    
+    system "mkdir -pv #{CREW_DEST_DIR}/home/chronos/user"
+    system "touch #{CREW_DEST_DIR}/home/chronos/user/.nanorc"
```
uberhacker: It would probably be better if we used `$HOME` for `/home/chronos/user` instead to abstract the user home directory.

packages/nano.rb
```diff
@@ -3,34 +3,43 @@
...
+  end
+  
+  def self.postinstall
+    puts "Personal configuration file is located in /home/chronos/user/.nanorc".lightgreen
+    open('/home/chronos/user/.nanorc', 'w') { |f|
```
uberhacker: Replace `/home/chronos/user` with `$HOME`?

packages/nano.rb
```diff
@@ -3,34 +3,43 @@
...
+      f << "set multibuffer\n"
+      f << "set nohelp\n"
+      f << "set nowrap\n"
+      f << "set positionlog\n"
+      f << "set historylog\n"
+      f << "set quickblank\n"
+      f << "set regexp\n"
+      f << "set smooth\n"
+      f << "set suspend\n"
+      }
```
uberhacker: This is an interesting heredoc implementation.  I haven't seen this yet.

lovedheart: xD

uberhacker:
Thank you for the update @lovedheart!  Please see inline comments.

packages/nano.rb
lovedheart: It is an absolute directory for all chrome os !

uberhacker: Not necessarily.  There could be other users.  Don't assume everybody is going to login via chronos.

PR #2345, opened 2018-05-11
fixes #2339 where the cd into the extracted cargo directory is lost h…

### LINKED ISSUE #2339: [kr] Unable to install
```
error: could not find `Cargo.toml` in `/usr/local/tmp/crew/README.md.dir` or any parent directory
kr failed to install: `cargo build --release && cargo install` exited with 101
```

danielcbaldwin:
…en running the build

Fixes #2339 

## Description
The issue was caused by breaking the long line I had previously into smaller lines and not correctly testing it. The problem is that I would cd into a directory in one system call and then try to do the build in that directory but the cd context is not preserved between system calls.

Works properly:
- [x] x86_64


skiman6010:
Perfect! Thanks so much!

PR #2366, opened 2018-05-19
Add fakeroot package

JL2210:
I added the fakeroot package. Any suggestions on how I could make it better?


uberhacker:
Looks good to me.  Thank you for the new package @JL2210!

PR #2367, opened 2018-05-19
Fixed invalid environment variable CREW_LIB_DIR in fakeroot

JL2210:
I accidentally had --libdir as CREW_LIB_DIR instead of CREW_LIB_PREFIX. I fixed this in this pull request. Sorry for not checking about that before I made the first one.


uberhacker:
No worries.  I should have caught it when I reviewed.

PR #2369, opened 2018-05-19
Fixed issue where Dpkg/ and Dpkg.pm weren't in @INC

JL2210:
I added some lines to dpkg.rb, so dpkg-architecture now runs. I wasn't sure if I did it that well, but it works. Any suggestions?


uberhacker:
Thank you for the update @JL2210.  Since we're adding new files, we'll need new binaries.  Please bump the version and remove the old binaries.
```
- version '1.19.0.5'
+ version '1.19.0.5-1'
```

JL2210:
Done.

PR #2370, opened 2018-05-19
Added zstd package

JL2210:
Added zstd to Packages. Hint: Apt coming up soon.

### DIFF
--- a/packages/zstd.rb
+++ b/packages/zstd.rb
@@ -4,14 +4,14 @@ class Zstd < Package
   description 'Zstandard - Fast real-time compression algorithm'
   homepage 'http://www.zstd.net'
   version '1.3.4'
-  source_url 'https://github.com/facebook/zstd/archive/master.tar.gz'
-  source_sha256 '7fba0070789a3c1e343fab4660958b73cc262110d2977589ba158b169d331ad6'
+  source_url 'https://github.com/facebook/zstd/archive/v1.3.4.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/zstd.rb
```diff
@@ -0,0 +1,17 @@
+require 'package'
+
+class Zstd < Package
+  description 'Zstandard - Fast real-time compression algorithm'
+  homepage 'http://www.zstd.net'
+  version '1.3.4'
+  source_url 'https://github.com/facebook/zstd/archive/master.tar.gz'
```
uberhacker: We should use the versioned source at https://github.com/facebook/zstd/archive/v1.3.4.tar.gz.

packages/zstd.rb
```diff
@@ -0,0 +1,17 @@
...
+  source_sha256 '7fba0070789a3c1e343fab4660958b73cc262110d2977589ba158b169d331ad6'
+  
+  def self.build
+    system 'make'
```
uberhacker: Is there a configure script?  If so, we will need the prefix and possibly the libdir abstracted.

packages/zstd.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
```
uberhacker: If there is no configure script, you could add `PREFIX=#{CREW_PREFIX}` here.

JL2210:
No configure script.

PR #2377, opened 2018-05-20
Add perl readkey package

JL2210:
Added perl_read_key to Packages.


uberhacker:

uberhacker:
Please remove the changes to dpkg and gettext packages on this PR.

packages/perl_read_key.rb
```diff
@@ -0,0 +1,19 @@
...
+  source_sha256 '0fa4fb2f8145e3fb2c2129ad28d55be175abcc258f239ba8ddc2cd83790aa8fb'
+
+  def self.build
+    system "perl", "Makefile.PL"
+    system "make"
+    system "perl", "ReadKey_pm.PL"
+  end
+
+  def self.install
+    system "make", "install"
```
uberhacker: Please add `DESTDIR=#{CREW_DEST_DIR}`.

JL2210:
I need some help on this one.

uberhacker:
You need help removing the dpkg and gettext package changes?

JL2210:
No. there should be another file that perl should run inside the source directory, but it requires the perl module that it builds to be installed before it can run.

JL2210:
But that too.

JL2210:
Should I make another package?

JL2210:
Is there in option for doing something between self.build and self.install?

uberhacker:
Based on their install notes, I believe this is all you need?
```
To install, unpack somewhere, type "perl Makefile.PL", and then "make test".
If the compilation and the tests are successful, then change to root and run
"make install".
```
I would suggest you just create a new branch and recommit.
```
$ cd /path/to/packages
$ cp perl_read_key.rb ~/
$ git push origin --delete add-perl-readkey
$ git branch -D add-perl-readkey
$ git remote add chromebrew git@github.com:skycocker/chromebrew.git
$ git fetch chromebrew
$ git checkout chromebrew/master
$ git checkout -b add-perl-readkey
$ cp ~/perl_read_key.rb .
... make changes ...
$ git add perl_read_key.rb
$ git commit -m "Add perl_read_key package"
$ git push origin master
```

uberhacker:
You could add this:
```
def self.check
  system 'make test'
end
```

JL2210:
I think it would be better to create another package. Perl readkey requires PerlReadkey_pm to be run before it adds the module to perl. I believe this is an issue on the source code maker's part.

packages/perl_read_key_pm.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Perl_read_key_pm < Package
+  description 'Character mode terminal access for Perl Term::ReadKey'
+  homepage 'https://metacpan.org/source/JSTOWE/TermReadKey-2.37/'
+  version '2.37'
+  source_url 'https://github.com/jonathanstowe/TermReadKey/archive/v2.37.tar.gz'
+  source_sha256 '0fa4fb2f8145e3fb2c2129ad28d55be175abcc258f239ba8ddc2cd83790aa8fb'
+
```
uberhacker: `depends_on 'perl_read_key'`?

packages/perl_read_key.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+
+  def self.postinstall
+    puts "Please install perl_read_key_pm after installing this package,"
+    puts "This just fills in dependencies so it can compile correctly."
+    puts "If you are currently installing perl_read_key_pm, you are OK."
```
uberhacker: Would adding `depends_on 'perl_read_key'` to perl_read_key_pm.rb remove the need for this message?  Also, you are missing an `end`. :)  Please test to make sure these actually install before creating the PR.  Please make all your commits locally until you are ready.

packages/perl_read_key_pm.rb
```diff
@@ -0,0 +1,27 @@
...
+    system "perl", "Makefile.PL"
+    system "make", "uninstall"
+    system "perl", "ReadKey_pm.PL"
+    system "perl", "Makefile.PL"
+    system "make"
+    system "crew", "remove", "perl_read_key"
+  end
+
+  def self.check
+    system "make", "test"
```
uberhacker: Missing `end`.  Please, please test first. :)

packages/perl_read_key_pm.rb
```diff
@@ -0,0 +1,27 @@
...
+
+class Perl_read_key_pm < Package
+  description 'Character mode terminal access for Perl Term::ReadKey'
+  homepage 'https://metacpan.org/source/JSTOWE/TermReadKey-2.37/'
+  version '2.37'
+  source_url 'https://github.com/jonathanstowe/TermReadKey/archive/v2.37.tar.gz'
+  source_sha256 '0fa4fb2f8145e3fb2c2129ad28d55be175abcc258f239ba8ddc2cd83790aa8fb'
+
+  def self.build
+    system "crew", "install", "perl_read_key"
```
uberhacker: You don't need this if you use `depends_on 'perl_read_key'`.

JL2210:
How do you make a puts line output blue text?

packages/perl_read_key_pm.rb
```diff
@@ -0,0 +1,27 @@
...
+    system "perl", "Makefile.PL"
+    system "make", "uninstall"
+    system "perl", "ReadKey_pm.PL"
+    system "perl", "Makefile.PL"
+    system "make"
+    system "crew", "remove", "perl_read_key"
```
uberhacker: Why do you need to remove?

JL2210:
depends_on perl read key will make perl_read_key install every time there is an update. The files are the same, it is just a messy fix that I could not find another workaround for. Also, if the files are the same, make install would

JL2210:
fail because it cannot overwrite the files.

uberhacker:
Maybe it would be better if you didn't separate the packages?  Just try and get it working in a single package, if possible.  To make the text blue, just add `.lightblue` at the end of the string.  There are other packages you can look at as an example.

JL2210:
I'll try.

JL2210:
Try it now

uberhacker:
Sometimes the simplest way is the best. :)

PR #2378, opened 2018-05-20
Added perl_gcstring_linebreak two-in-one package

JL2210:
I added the perl_gcstring_linebreak package, which is a necessity for po4a. Coming up!

### DIFF
--- a/packages/perl_gcstring_linebreak.rb
+++ b/packages/perl_gcstring_linebreak.rb
@@ -1,0 +1,21 @@
+require 'package'
+
+class Perl_unicode_linebreak < Package
+  description 'UAX 14 Unicode Line Breaking Algorithm - Perl binding Unicode::LineBreak Unicode::GCString'
+  homepage 'http://search.cpan.org/~nezumi/Unicode-LineBreak-2018.003/lib/Unicode/LineBreak.pod'
+  version '2018.003'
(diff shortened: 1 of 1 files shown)


uberhacker:

packages/perl_gcstring_linebreak.rb
```diff
@@ -0,0 +1,21 @@
+require 'package'
+
+class Perl_unicode_linebreak < Package
```
uberhacker: This class name does not match the filename.

packages/perl_gcstring_linebreak.rb
```diff
@@ -0,0 +1,21 @@
...
+  def self.build
+      system "git", "clone", "--recurse-submodules", "https://github.com/hatukanezumi/Unicode-LineBreak", "-b", "Unicode-LineBreak-2018.003"
+    Dir.chdir ("Unicode-LineBreak") do
+      system "perl", "Makefile.PL"
+      system "make"
+    end
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
```
uberhacker: I think we need `Dir.chdir ("Unicode-LineBreak") do` here also.  Please test this to make sure.

PR #2379, opened 2018-05-20
Add po4a package

JL2210:
Added po4a package.

### DIFF
--- a/packages/perl_gcstring_linebreak.rb
+++ b/packages/perl_gcstring_linebreak.rb
@@ -1,0 +1,21 @@
+require 'package'
+
+class Perl_unicode_linebreak < Package
+  description 'UAX 14 Unicode Line Breaking Algorithm - Perl binding Unicode::LineBreak Unicode::GCString'
+  homepage 'http://search.cpan.org/~nezumi/Unicode-LineBreak-2018.003/lib/Unicode/LineBreak.pod'
+  version '2018.003'
+  source_url 'https://github.com/hatukanezumi/Unicode-LineBreak/archive/Unicode-LineBreak-2018.003.tar.gz'
+  source_sha256 '6f8cb4de140f8b63924786df8b0c2389c342bccdee05fbb1b9af8d8b6a8a3fad'
+
+  def self.build
+      system "git", "clone", "--recurse-submodules", "https://github.com/hatukanezumi/Unicode-LineBreak", "-b", "Unicode-LineBreak-2018.003"
+    Dir.chdir ("Unicode-LineBreak") do
+      system "perl", "Makefile.PL"
+      system "make"
+    end
+  end
+
+  def self.install
+    system "make", "install"
+  end
+end
--- a/packages/po4a.rb
+++ b/packages/po4a.rb
@@ -1,0 +1,27 @@
+require 'package'
+
+class Po4a < Package
+  description 'Maintain the translations of your documentation with ease (PO for anything)'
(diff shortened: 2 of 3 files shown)


uberhacker:

uberhacker:
Please remove the conflicting packages.

JL2210:
Wait.

JL2210:
Let me look into this a bit more.

JL2210:
It needs /usr/bin/locale, but the only locale I could find was /usr/local/bin/locale. Help?

uberhacker:
If the path is hardcoded in Build.PL and there are no environment or override options, then you could do something like this:
```
system "sed -i 's,/usr/bin,#{CREW_PREFIX}/bin,g' Build.PL"
```

JL2210:
It would have to be hardcoded to give that error. Let me try that.

JL2210:
If you're wondering why I run the same set of commands twice, the source code is missing a file called META.yml. The first time it runs Build.PL, it notices that that file is missing. The first time it runs Build, it creates META.yml. Then, it runs successfully the second time because all the files it needs are created.

uberhacker:
Gotta <3 perl. ;)

JL2210:
Yeah. You used sed on Build.PL instead of build.

uberhacker:
So this installs fine and you have tested without errors?

JL2210:
No.

JL2210:
```Use of uninitialized value $Text::WrapI18N::charmap in pattern match (m//) at /usr/local/lib/perl5/site_perl/5.26.1/Text/WrapI18N.pm line 137.```

JL2210:
```Can't exec "/usr/bin/locale": No such file or directory at /usr/local/lib/perl5/site_perl/5.26.1/Text/WrapI18N.pm line 134.```

uberhacker:
Does the `distmeta` option fix the issue?  Let me know when you think this is ready.

JL2210:
It makes it so I don't have to run Build with no arguments again, but it still tries to stat /usr/bin/locale. It also doesn't install to /usr/local. What do I do?

uberhacker:
Try `system "sed -i 's,/usr/bin,#{CREW_DEST_PREFIX}/bin,g' Build.PL"`.

JL2210:
Tried that. locale isn't hardcoded.

JL2210:
I symlinked it (removed rootfs verification) and this is the error now:
```I/O error : Attempt to load network entity http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl
warning: failed to load external entity "http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl"
cannot parse http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl
Undefined subroutine &PerlIO::F_UTF8 called at /usr/local/lib/perl5/5.26.1/Pod/Man.pm line 809.
        ...propagated at Po4aBuilder.pm line 273.
po4a failed to install: `perl Build` exited with 4```

uberhacker:
I don't know.  Maybe submit an issue in the queue at https://github.com/mquinson/po4a/issues?

JL2210:
Did already. Evidently, they don't get back to you on issues right away. I didn't realize it, but I think I made a duplicate package: xsltproc. Could you remove it?

JL2210:
Thank you. Any way that you could fix this:
```warning: failed to load external entity:```?

PR #2381, opened 2018-05-20
Fix gettext version number

JL2210:
Changed gettext version number to 0.19.8.1-1 to initiate an update.

### DIFF
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -9,7 +9,6 @@ class Dpkg < Package
 
   depends_on 'bz2'
   depends_on 'xzutils'
-  depends_on 'icu4c' => :build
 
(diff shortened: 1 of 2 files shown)


uberhacker:
I already merged this fix.  No need to increment the version since no new files have been added in the package.

JL2210:
It implements a change. If it doesn't upgrade, it doesn't have the dependency it needs.

uberhacker:
If a package is already installed and a change doesn't affect any of the files in the package, a version bump should not be necessary.  You could have a dependency issue but since this is a core package, it shouldn't be an issue.  With that said, if icu4c is in fact a dependency of gettext then we should be installing it when we install Chromebrew.

PR #2410, opened 2018-06-07
Add fpc package

uberhacker:
Free Pascal is a 32, 64 and 16 bit professional Pascal compiler. It can target many processor architectures: Intel x86 (including 8086), AMD64/x86-64, PowerPC, PowerPC64, SPARC, ARM, AArch64, MIPS and the JVM. Supported operating systems include Linux, FreeBSD, Haiku, Mac OS X/iOS/iPhoneSimulator/Darwin, DOS (16 and 32 bit), Win32, Win64, WinCE, OS/2, MorphOS, Nintendo GBA, Nintendo DS, Nintendo Wii, Android, AIX and AROS. Additionally, support for the Motorola 68k architecture is available in the development versions.  See https://freepascal.org/.

### DIFF
--- a/packages/fpc.rb
+++ b/packages/fpc.rb
@@ -1,0 +1,35 @@
+require 'package'
+
+class Fpc < Package
+  description 'Free Pascal is a 32, 64 and 16 bit professional Pascal compiler.'
+  homepage 'https://www.freepascal.org/'
+  version '3.0.4'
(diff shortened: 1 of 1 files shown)


JL2210:
Awesome! I love this compiler, even if a little outdated. It can compile C on windows and linux.

PR #2415, opened 2018-06-09
Added lines to readline to fix fdisk

JL2210:
Fixed issue where executing fdisk after installing readline gave the error:
 ```
libreadline.so.s: error: /usr/local/lib(64)/libreadline.so.6: undefined symbol: UP
```
Using system libreadline.so.6 now

### DIFF
--- a/packages/readline.rb
+++ b/packages/readline.rb
@@ -34,7 +34,7 @@ class Readline < Package
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
-    system "mv", "#{CREW_DEST_LIB_PREFIX}/libreadline.so.6", "#{CREW_DEST_LIB_PREFIX}/libreadline.so.6.old"
+    system "rm", "#{CREW_DEST_LIB_PREFIX}/libreadline.so.6"
     system "ln", "-s", "/#{ARCH_LIB}/libreadline.so.6", "#{CREW_DEST_LIB_PREFIX}/libreadline.so.6"
   end
 end
--- a/packages/readline.rb
+++ b/packages/readline.rb
@@ -8,16 +8,8 @@ class Readline < Package
   source_sha256 '56ba6071b9462f980c5a72ab0023893b65ba6debb4eeb475d7a563dc65cafd43'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/readline-6.3p8-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/readline-6.3p8-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/readline-6.3p8-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/readline-6.3p8-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: '7b70ee08258af2bd190b30639306aa4d0b71273cfb702a4ae75003317f19702f',
-     armv7l: '7b70ee08258af2bd190b30639306aa4d0b71273cfb702a4ae75003317f19702f',
-       i686: '572baf2f2bba47d719714fc97709bc3932abab4bf1a9d838bb47ccd0f1fa4395',
-     x86_64: 'f235ee73feba6d2c0bb085042db9462225cda1a18f190a47116417cc6993e716',
   })
 
   depends_on 'buildessential' => :build
@@ -42,5 +34,7 @@ class Readline < Package
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "mv", "#{CREW_DEST_LIB_PREFIX}/libreadline.so.6", "#{CREW_DEST_LIB_PREFIX}/libreadline.so.6.old"
+    system "ln", "-s", "/#{ARCH_LIB}/libreadline.so.6", "#{CREW_DEST_LIB_PREFIX}/libreadline.so.6"
   end
 end
--- a/packages/readline.rb
+++ b/packages/readline.rb
@@ -34,7 +34,8 @@ class Readline < Package
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
-    system "rm", "#{CREW_DEST_LIB_PREFIX}/libreadline.so.6"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by uberhacker** [comment]
@JL2210: Thank you for the update.  The reason this package was created was to override the out-of-the-box version of `/lib(64)/libreadline.so.6`.  I think a better solution would be to set `LD_LIBRARY_PATH=/lib64:/usr/local/lib64` or without 64 on 32-bit systems.  Let me know if this fixes the issue with fdisk.  Also, are you having this issue on the stable channel?

**Comment by JL2210** [comment]
Yes. All channels, including stable. It does fix the issue. I checked. It
is a lot harder to change environment variables within crew than to do this
or get the user involved.

**Comment by uberhacker** [comment]
Right, but how many other programs are affected by this change?  I counted over 20 packages that require readline as a dependency.  Are you willing to test all these other packages?  Also, what do we gain by using the system library with all apps that depend on readline?  Why not just get rid of the readline package altogether?

**Comment by JL2210** [comment]
What are the packages? I will test them. If we remove readline, we will have to change them anyway.

**Comment by JL2210** [comment]
Fdisk and Tmux work.

**Comment by JL2210** [comment]
Now that I think about it, those lines would probably be better in the postinstall. We don't want the system libraries showing up in the prebuilt binaries.

**Comment by JL2210** [comment]
Would crew compress the file itself, or compress the symlink?

**Comment by JL2210** [comment]
Dang it. Why do they put the close and comment button so close to the comment button?

**Comment by JL2210** [comment]
I think it's good now. Try it.

**Comment by uberhacker** [comment]
```
bc.rb:  depends_on 'readline'
bind.rb:  depends_on "readline"
clisp.rb:  depends_on 'readline'
gawk.rb:  depends_on 'readline' => :build
gdbm.rb:  depends_on 'readline'
lua.rb:  depends_on 'readline'
lvm2.rb:  depends_on 'readline'
mandb.rb:  depends_on 'readline'
metasploit.rb:  depends_on 'readline'
newlisp.rb:  #depends_on 'readline'
parted.rb:  depends_on 'readline'
php.rb:  depends_on 'readline'
php5.rb:  depends_on 'readline'
php7.rb:  depends_on 'readline'
postgres.rb:  depends_on 'readline'
python3.rb:  depends_on 'readline'
r.rb:  depends_on 'readline'
rlwrap.rb:  depends_on 'readline'
ruby.rb:  depends_on 'readline'
socat.rb:  depends_on 'readline'
sqlite.rb:  depends_on 'readline'
tig.rb:  depends_on 'readline'
tmux.rb:  depends_on 'readline'
```
I'm actually curious why your readline is any different than the compiled version.  I'm hesitant about this "fix".
```
$ crew whatprovides libreadline
readline: /usr/local/lib/libreadline.so
readline: /usr/local/lib/libreadline.so.6
readline: /usr/local/lib/libreadline.so.6.3

Total found: 3
$ ls /lib/libreadline*
/lib/libreadline.so.6  /lib/libreadline.so.6.3
```

**Comment by JL2210** [comment]
The compiled readline breaks fdisk because it has an undefined symbol: `UP`.

**Comment by uberhacker** [comment]
How did you install fdisk, by the way?  Or did you mean gdisk from the gptfdisk package?
```
$ which fdisk
which: no fdisk in (/usr/local/bin:/usr/bin:/bin:/opt/bin)
```

**Comment by JL2210** [comment]
```sudo which fdisk```

**Comment by JL2210** [comment]
It's in an sbin directory.

**Comment by uberhacker** [comment]
Yep, you are right.  I didn't have `/sbin` in my path:
```
$ ls /sbin/fdisk
/sbin/fdisk
```

**Comment by JL2210** [comment]
```
chronos@localhost / $ sudo find / -name fdisk
Password: 
/sbin/fdisk
chronos@localhost / $ fdisk -v
bash: fdisk: command not found
chronos@localhost / $ sudo fdisk -v
fdisk: symbol lookup error: /usr/local/lib64/libreadline.so.6: undefined symbol: UP
chronos@localhost / $ 
```

**Comment by uberhacker** [comment]
For the exact reasons I just explained.  Besides, the system /sbin version is older than the util_linux packaged version:
```
$ sudo /usr/local/sbin/fdisk -v
fdisk from util-linux 2.30
$ sudo /sbin/fdisk -v
fdisk from util-linux 2.28.2
```

**Comment by JL2210** [comment]
util_linux isn't installed with crew, though.

**Comment by JL2210** [comment]
If you are going to suggest that, at least make util_linux a buildessential.

**Comment by uberhacker** [comment]
It's not a build essential but a case could be made that it's a utility essential.

**Comment by uberhacker** [comment]
I'm going to close this PR.  We need a better solution.

**Comment by uberhacker** [comment]
FWIW, if I don't have readline installed, the system fdisk works fine.
```
$ crew search readline
readline: The GNU Readline library provides a set of functions for use by applications that allow users to edit command lines as they are typed in.
readline7: The GNU Readline library provides a set of functions for use by applications that allow users to edit command lines as they are typed in.
$ sudo which fdisk
/sbin/fdisk
$ sudo fdisk -v
fdisk from util-linux 2.28.2
$ sudo fdisk

Usage:
 fdisk [options] <disk>      change partition table
 fdisk [options] -l [<disk>] list partition table(s)

Display or manipulate a disk partition table.

Options:
 -b, --sector-size <size>      physical and logical sector size
 -B, --protect-boot            don't erase bootbits when creating a new label
 -c, --compatibility[=<mode>]  mode is 'dos' or 'nondos' (default)
 -L, --color[=<when>]          colorize output (auto, always or never)
                                 colors are enabled by default
 -l, --list                    display partitions end exit
 -o, --output <list>           output columns
 -t, --type <type>             recognize specified partition table type only
 -u, --units[=<unit>]          display units: 'cylinders' or 'sectors' (default)
 -s, --getsz                   display device size in 512-byte sectors [DEPRECATED]
     --bytes                   print SIZE in bytes rather than in human readable format
 -w, --wipe <mode>             wipe signatures (auto, always or never)

 -C, --cylinders <number>      specify the number of cylinders
 -H, --heads <number>          specify the number of heads
 -S, --sectors <number>        specify the number of sectors per track

 -h, --help     display this help and exit
 -V, --version  output version information and exit

Available columns (for -o):
 gpt: Device Start End Sectors Size Type Type-UUID Attrs Name UUID
 dos: Device Start End Sectors Cylinders Size Type Id Attrs Boot End-C/H/S Start-C/H/S
 bsd: Slice Start End Sectors Cylinders Size Type Bsize Cpg Fsize
 sgi: Device Start End Sectors Cylinders Size Type Id Attrs
 sun: Device Start End Sectors Cylinders Size Type Id Flags

For more details see fdisk(8).
```

JL2210:
Yes. It does. The chromebrew readline breaks it.

uberhacker:
I guess the workaround is to install sudo and util_linux to use `/usr/local/sbin/fdisk` instead.

PR #2423, opened 2018-06-14
Add autoconf build dependency to Weston

lzmartinico:
## Description
Adds a build dependency to prevent Weston installation from failing

### DIFF
--- a/packages/weston.rb
+++ b/packages/weston.rb
@@ -20,6 +20,8 @@ class Weston < Package
   depends_on 'dbus'
   
   depends_on 'libxxf86vm'
+  
+  depends_on 'autoconf' => :build
 
   def self.build
     system "./autogen.sh"


uberhacker:
@lzmartinico: Thank you for the update but we already include autoconf in Chromebrew "core" so no need to include it here since it is added in a standard install already.

lzmartinico:
@uberhacker my local version of Chromebrew (0.4.3) did not come with autoconf installed

uberhacker:
@lzmartinico: Did you try `crew update` followed by `crew upgrade` yet?  If that doesn't fix it, try:
```
$ sudo dev_install --reinstall -y
$ curl -Ls git.io/vddgY | bash
```
I'm guessing you haven't updated in awhile.

lzmartinico:
Yes that's right, I hadn't updated in a while, but there was no indication I hadn't. Homebrew by default updates whenever a new installation is started if the last update was not too recent, maybe Chromebrew should have the same behaviour? Otherwise, I think there might be some merit in preserving dependencies added in recent versions of core crew.

PR #2435, opened 2018-06-29
Build with python3 support

doriclazar:
## Description
Script now checks if python3 is the only version installed, and if so, it switches it with python27 in requirements and configure string (interpret).

### DIFF
--- a/packages/vim.rb
+++ b/packages/vim.rb
@@ -19,18 +19,11 @@ class Vim < Package
        i686: 'c248117c8aba83f20c362999da8ec4b7c7b73aa74e6e4251a0dc114320809df9',
      x86_64: 'c0b2c46297ce1b5d01c42352675e26c980c4f51e1816263150c4c0e99f1899be',
   })
-  if !File.exists?("#{CREW_PREFIX}/bin/python") && File.exists?("#{CREW_PREFIX}/bin/python3")
-  @@only_py3 = true
-  end 
   depends_on 'compressdoc' => :build
   depends_on 'ncurses'
   # vim uses shared library of following languages, so need them isntalled at run-time
   depends_on 'perl'
-  if @@only_py3
   depends_on 'python3'
-  else
-  depends_on 'python27'
-  end 
   depends_on 'ruby'
 
   def self.build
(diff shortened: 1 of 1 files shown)


uberhacker:
@doriclazar: Thank you for the update!  Why not just switch from the python27 to python3 dependency and call it a day?
```
- depends_on 'python27'
+ depends_on 'python3'
- --enable-pythoninterp \
+ --enable-python3interp \
```

uberhacker:
One thing that has really annoyed me about this package is vimdiff doesn't work even with the `--with-vimdiff` option so if this update fixes that, it will be a bonus.

doriclazar:
It does. Even if built with both python versions, `vimdif` will still work with any non-removed.
However, precompiled also works for me with required python27.

skiman6010:
@uberhacker doing it this way is telling vim to support Python3 if I'm reading this correctly. If I remember correctly, vim is weird with its Python support.

uberhacker:
@skiman6010: Yeah, I just thought instead of supporting python27, why not just make python3 a hard requirement.  @doriclazar: I believe the vim issue is on i686 but it could just be a corrupt binary because if I compile from source, vimdiff works.

uberhacker:
@doriclazar: Can you please remove the binaries and increment the version?
```
- version '8.0-2'
+ version '8.0-3'

-   binary_url ({
-     aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/vim-8.0-2-chromeos-armv7l.tar.xz',
-      armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/vim-8.0-2-chromeos-armv7l.tar.xz',
-        i686: 'https://dl.bintray.com/chromebrew/chromebrew/vim-8.0-2-chromeos-i686.tar.xz',
-      x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/vim-8.0-2-chromeos-x86_64.tar.xz',
-   })
-   binary_sha256 ({
-     aarch64: '41c5462bc6a85758c826a3238d1e639beff5aeaad4cad6c279a8916b6d3ae21a',
-      armv7l: '41c5462bc6a85758c826a3238d1e639beff5aeaad4cad6c279a8916b6d3ae21a',
-        i686: 'c248117c8aba83f20c362999da8ec4b7c7b73aa74e6e4251a0dc114320809df9',
-      x86_64: 'c0b2c46297ce1b5d01c42352675e26c980c4f51e1816263150c4c0e99f1899be',
-   })
```

doriclazar:
Done. Thank you very much @uberhacker .

PR #2457, opened 2018-07-18
Ensures core packages are installed during installation.

JL2210:
Added a line to install.sh to install all packages listed in tools/core_packages.txt as a safety net. Also removed glibc223 from core_packages.txt

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -191,6 +191,7 @@ git reset --hard origin/master
 yes | crew install buildessential
 yes | crew install less
 yes | crew install most
+yes | crew install $(curl -Ls https://github.com/skycocker/chromebrew/raw/master/tools/core_packages.txt)
 echo
 echo "To set the default PAGER environment variable to be able to use less:"
 echo "echo \"export PAGER=$CREW_PREFIX/bin/less\" >> ~/.bashrc && . ~/.bashrc"
--- a/tools/core_packages.txt
+++ b/tools/core_packages.txt
@@ -12,7 +12,6 @@ gcc7
 gettext
 git
 glibc
-glibc223
 gmp
 gnutls
 icu4c


install.sh
```diff
@@ -191,6 +191,7 @@ git reset --hard origin/master
 yes | crew install buildessential
 yes | crew install less
 yes | crew install most
+yes | crew install $(curl -Ls https://github.com/skycocker/chromebrew/raw/master/tools/core_packages.txt)
```
uberhacker: Why do we need this?  All of these packages should be installed already at this point.

JL2210:
I've noticed that the installation doesn't install some of these packages. This just ensures that it does.

install.sh
JL2210: I've noticed that the installation doesn't install some of these packages. This just ensures that it does.

uberhacker: How do you know core_packages.txt is correct?  I just used the output of `crew list installed` to create it after a fresh install.  Can you please substantiate the missing packages with some actual diffs?

JL2210: Here is your diff:

[diff](https://github.com/skycocker/chromebrew/files/2207388/diff.txt)

uberhacker: Currently `readline` is not in core_packages.txt so maybe that needs updated instead. :)

JL2210: Here is the command I used to create the diff:
```
diff with-pr without-pr > diff
```

uberhacker:
I created core_packages.txt for the purpose of checking against new packages to determine if depends_on lines are needed or not. 
 @skiman6010: Do you have any input regarding this?

JL2210:
It could always be removed if needed. The core_packages.txt file still needs updating. It needs to include readline and needs to remove glibc223.

skiman6010:
Is there any evidence that crew misses a core package and needs a just in case backup? I would say if you can provide evidence of that happening then we can add this PR, but if not, then it just seems like a waste.

JL2210:
Here is the diff:

https://github.com/skycocker/chromebrew/files/2207388/diff.txt

JL2210:
Also, I removed glibc223 from the core_packages.txt file because it is not a package anymore.

skiman6010:
Right but that doesn't provide any evidence of us needing a just in case backup option running on everyone's system. That just shows me that core_packages.txt is out of date.

JL2210:
When install is cancelled, this could potentially help with that. This only runs once, during the install.

JL2210:
Also, this shows that buildessential is outdated. When using this approach, readline is installed, as a pose to not being installed when this approach is not used.

uberhacker:
@JL2210: Do you think we're missing a needed dependency?  Which package has trouble if readline is not installed?  Maybe the problem goes deeper and we need readline for one of the core packages?  If that is the case then we should add a `depends_on 'readline'` line in that package.  Just my two cents.

JL2210:
Gettext.

JL2210:
I'm closing this. Remember to add readline as a dependency to gettext and remove glibc223 from core_packages.txt

JL2210:
Sure. I'm doing one to extend the functionality of fakeroot right now,
though.

JL2210:
autoreconf: configure.ac: not using Gettext: Readline missing

uberhacker:
@JL2210: FYI:
```
$ crew remove readline
Readline removed!
chronos@localhost ~/chromebrew/tools (update-packages_yaml *%=)$ gettext -h
Usage: gettext [OPTION] [[TEXTDOMAIN] MSGID]
or:    gettext [OPTION] -s [MSGID]...

Display native language translation of a textual message.

  -d, --domain=TEXTDOMAIN   retrieve translated messages from TEXTDOMAIN
  -e                        enable expansion of some escape sequences
  -E                        (ignored for compatibility)
  -h, --help                display this help and exit
  -n                        suppress trailing newline
  -V, --version             display version information and exit
  [TEXTDOMAIN] MSGID        retrieve translated message corresponding
                            to MSGID from TEXTDOMAIN

If the TEXTDOMAIN parameter is not given, the domain is determined from the
environment variable TEXTDOMAIN.  If the message catalog is not found in the
regular directory, another location can be specified with the environment
variable TEXTDOMAINDIR.
When used with the -s option the program behaves like the 'echo' command.
But it does not simply copy its arguments to stdout.  Instead those messages
found in the selected catalog are translated.
Standard search directory: /usr/local/share/locale

Report bugs to <bug-gnu-gettext@gnu.org>.

JL2210:
xgettext

skiman6010:
@JL2210 If you're having a specific package issue, can you please create a detailed issue for us to track the fixes for? 

Thanks.

PR #2459, opened 2018-07-19
Added crew profile for more compatibility between shells

JL2210:
# Fixes

Better compatibility between shells

## Description
I created a profile for crew in /usr/local/etc/profile. I believe this would allow for greater inter-compatibility between shells. I also believe this would fix the issue of your .bashrc clogging up. 

## Addtional Information

Works properly:
- [x] x86_64
- [x] x86
- [x] aarch64
- [x] arm

---

## That's it

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -191,12 +191,15 @@ git reset --hard origin/master
 yes | crew install buildessential
 yes | crew install less
 yes | crew install most
-yes | crew install crew_profile
 echo
 echo "To set the default PAGER environment variable to be able to use less:"
 echo "echo \"export PAGER=$CREW_PREFIX/bin/less\" >> ~/.bashrc && . ~/.bashrc"
 echo
 echo "Alternatively, you could use most.  Why settle for less, right?"
 echo "echo \"export PAGER=$CREW_PREFIX/bin/most\" >> ~/.bashrc && . ~/.bashrc"
+echo
+
+yes | crew install crew_profile
+
 echo
 echo "Chromebrew installed successfully and package lists updated."
--- a/install.sh
+++ b/install.sh
@@ -191,6 +191,7 @@ git reset --hard origin/master
 yes | crew install buildessential
 yes | crew install less
 yes | crew install most
+yes | crew install crew_profile
 echo
 echo "To set the default PAGER environment variable to be able to use less:"
 echo "echo \"export PAGER=$CREW_PREFIX/bin/less\" >> ~/.bashrc && . ~/.bashrc"
--- a/packages/crew_profile.rb
+++ b/packages/crew_profile.rb
@@ -4,8 +4,8 @@ class Crew_profile < Package
   description 'A profile for Chromebrew, to be used instead of ~/.*rc'
   homepage 'https://github.com/skycocker/chromebrew'
   version '0.4.3'
-  source_url 'https://raw.githubusercontent.com/skycocker/chromebrew/master/install.sh'
-  source_sha256 '6c061009df229624485279e8a076f94b61218b8dc67478d49a6b563692e92aef'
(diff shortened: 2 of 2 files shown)


packages/crew_profile.rb
```diff
@@ -0,0 +1,23 @@
...
+    puts
+    puts "To create your very own Crew profile, please execute this:".lightblue
+    puts
+    puts "echo \"if [ -f #{CREW_PREFIX}/etc/profile ]; then\" >> ~/.\{bash,zsh\}rc".lightblue
+    puts "echo \"  # Source crew profile\" >> ~/.\{bash,zsh\}rc".lightblue
+    puts "echo \"  source #{CREW_PREFIX}/etc/profile\" >> ~/.\{bash,zsh\}rc".lightblue
+    puts "echo \"fi\" >> ~/.\{bash,zsh\}rc".lightblue
+    puts "source ~/.\{bash,zsh\}rc"
+    puts
+    puts "Works with Bash and Zsh"
```
uberhacker: This is not going to work without something in `/usr/local/etc/profile`, correct?

install.sh
```diff
@@ -191,6 +191,7 @@ git reset --hard origin/master
 yes | crew install buildessential
 yes | crew install less
 yes | crew install most
+yes | crew install crew_profile
```
uberhacker: I would rather this be optional.

packages/crew_profile.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Crew_profile < Package
+  description 'A profile for Chromebrew, to be used instead of ~/.*rc'
+  homepage 'https://github.com/skycocker/chromebrew'
+  version '0.4.3'
+  source_url 'https://raw.githubusercontent.com/skycocker/chromebrew/master/install.sh'
```
uberhacker: This is a moving target.  As soon as your PR is merged into master, the sha256 of this will be different.  If you want to have something more static, use a hashed version of a file like `https://github.com/skycocker/chromebrew/blob/511ee9cb3138e113df74ddf0dee057ef5d9a46fd/README.md`.

uberhacker: Even if you change the sha256 hash, this is still a moving target.  Any new commits to install.sh on the master branch is going to break this.

JL2210:
OK. Chrome broke yesterday. Managed to fix it, but I was tired. Will fix
the moving target. If this is going to work as I intended, as a total and
complete replacement for your bashrc and zshrc, it will have to be
necessary.

packages/crew_profile.rb
JL2210: It'll work.

install.sh
JL2210: If this is going to work as I intended, as a total and
complete replacement for your bashrc and zshrc, it will have to be
necessary.

packages/crew_profile.rb
JL2210: Got it. Will change.

install.sh
JL2210: A lot of the lines in packages depend on the bashrc. Little, if any, support is given in those packages for zsh and fish. I created this to fix that problem. They can still use their bashrc. It doesn't override it. Same with zshrc and fishrc.

uberhacker: You haven't presented your use case example of how this would work.  Are you suggesting all the current packages that have postinstall configuration steps be included in /usr/local/etc/profile instead?  If that is the case, then maybe a `depends_on 'crew_profile'` line should be added in those packages?

JL2210: Yeah. It would just be easier in install. If it's possible, that's what I want. I don't want the zsh and fish users kept in the dark.

JL2210:
Not possible, will close and rebase pull request as optional. Fish does some wierd stuff with `if`.

PR #2462, opened 2018-07-22
Update packages.yaml in tools

uberhacker:
This update adds activity to mark the project status.  The intention is to make it easier to isolate projects that need updates more often than others.

- high = 2 or more updates in the past 6 months
- medium = 2 or more updates in the past year
- low = less than 2 updates in the past year
- none = no updates in the past year


skiman6010:
Thank you. This will in theory be really useful.

PR #2469, opened 2018-07-27
Revert "Update icu4c from 60.1 to 62.1"

JL2210:
Reverts skycocker/chromebrew#2467

This pull request breaks gettext. Can you revert this or recompile gettext?

```
xgettext: error while loading shared libraries: libicui18n.so.60:
 cannot open shared object file: No such file or directory
```


uberhacker:
Can you please add this as an issue instead of a PR?  I would rather fix than revert, if possible.  Also, you are welcome to provide a PR for gettext.

PR #2480, opened 2018-07-28
Update sed from 4.1.5 to 4.5

uberhacker:


JL2210:
You have a sed package now? Great! Now it might not give me the `'e/r/w' commands disabled in sandbox mode` error.

uberhacker:
Yes, that is exactly what this package fixes.

Issue #2489, opened 2018-07-29
TiLP in Chromebrew? (Attempted solutions)


superloach:
TiLP is a tool for transferring data to/from TI calculators, and would be useful to many students using Chromebooks, such as myself! I'd like to share my experiences trying to install it in Chrome OS, request assistance, and propose adding it to the Chromebrew repos once it's working.

Problems:
 - Automated installer fails
   - Permission is denied when attempting to execute sanity checks.
   - Full output: https://pastebin.com/Ku8t93pa
 - Automated installer fails as root
   - Complains about not being able to work in read-only directories.
   - Modifying `SRCDIR` to somewhere read/write just gives same errors as with no root.
   - Full output: https://pastebin.com/G7wLZZ0B
 - Build instructions are wrong
   - `INSTALL` file says to use `./configure && make && make install`, but there isn't `./configure`.
   - Full output: https://pastebin.com/u43s8vRW
 - Guessed that `autoconf configure.ac` might do the trick
   - Prints the contents of `configure.ac`, and then complains about macros (with or without root).
   - Full output: https://pastebin.com/a5x7pL2n

Possible issues once we get further:
 - Says it depends on GTK
   - Might be the issue with `autoconf`, said something about a KDE-related macro.
   - Compiled, TiLP can run with `--no-gui`.
   - Weston *is* in the works, and could help with this?

It seems like some editing of the installer script might make it work, but I have no clue where to start. I'm willing to dive much deeper into getting this working, but would love to have support from more experienced Cros "porters" 😝 Hope this won't be too hard to get working and into the repo!

TiLP Homepage: http://lpg.ticalc.org/prj_tilp/
TiLP Installer: http://lpg.ticalc.org/prj_tilp/download/install_tilp.sh
TiLP Source: https://www.ticalc.org/pub/unix/tilp.tar.gz

uberhacker:
@superloach: I went ahead and made an attempt to add this package in PR #2492 (depends on PR #2491).  Since the [GUI](#889) is still a work in progress, your mileage may vary.  Please give it a go when you get a chance and let me know if you have any success.  Feel free to post your results.  This project is a great way to help move the GUI work along.

superloach:
Awesome, thanks! I tried installing it, and it worked with just a little tweaking.

[...]

PR #2495, opened 2018-07-31
Added libdir to tilp package.

JL2210:
Fixes comment 3 of #2489

Sets libdir of tilp to #{CREW_LIB_PREFIX}


uberhacker:
Actually, that's not enough.  I will push up a PR with a fix and pre-built binaries.

JL2210:
Could I do it? I think I have an idea.

PR #2497, opened 2018-07-31
Update tilp from 1.18 to 1.19

uberhacker:
Fixes #2498.

### DIFF
--- a/packages/tilp.rb
+++ b/packages/tilp.rb
@@ -3,7 +3,7 @@ require 'package'
 class Tilp < Package
   description 'TiLP is a linking program for Texas Instruments\' graphing calculators.'
   homepage 'http://lpg.ticalc.org/prj_tilp/'
-  version '1.18'
+  version '1.19'
   source_url 'https://www.ticalc.org/pub/unix/tilp.tar.gz'
   source_sha256 '6ba834f7fdbbce9818ccaa864222aed2d1688b210e9ff2c59576d1fde5159cd7'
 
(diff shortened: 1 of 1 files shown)


JL2210:
#2489 not #2498

JL2210:
@uberhacker, can't you just use different delimeters with sed, such as colons instead of slashes? Then you could just sed out /usr/bin/getent with #{CREW_PREFIX}/bin/getent? It would take up less space and be less of a hack, plus it works.

JL2210:
@uberhacker, ever heard of autoconf?

uberhacker:
@JL2210: Yes, but it won't work in this instance.  The install script is poorly written and doesn't allow overrides.  Actually, my fix is not much of a hack at all.  It just allows for the `LIBDIR` override.

JL2210:
Yes, but it completely removes a part instead of replacing the tool it tries to find with the tool it needs to find. BTW, you're using tilp2. Tilp1 has configure scripts.

JL2210:
Also, you can get a versioned tarball from their SourceForge page: https://sourceforge.net/projects/tilp/files

JL2210:
My second one worked just fine. But I closed it in favor of yours, because you seem to know what you're doing.

JL2210:
This doesn't need to be clogging up this PR. I'll talk to you on the user group.

uberhacker:
Actually, I'm not removing anything.  All I'm doing is wrapping the LIBDIR logic with an if statement so it overrides that section if the environment variable is set.  Nice find on the sourceforge.net project.  Does it handle the dependencies needed also?

JL2210:
Yes. The deps are in another tarball.

debrouxl:
Well, I'm not against fixing the upstream user install script, if there's a good reason for it ;)
It already allows overriding SRCDIR and PREFIX (the latter is then passed to autotools, as it should be). The LIBDIR section could easily be made conditional to the LIBDIR environment variable being empty, like SRCDIR and PREFIX are already handled.

You don't need the SF.net project: using releases is a bad idea, the versions in the Git repositories on Github (and now Gitlab, which is mirroring them) always contain fixes and improvements.

uberhacker:
Github repo?  Do you have a url?

debrouxl:
See install_tilp.sh: https://github.com/debrouxl/tilibs , https://github.com/debrouxl/tilp_and_gfm .

PR #2518, opened 2018-08-07
Added Xterm and Xwayland. The X server now works.

JL2210:
See the description. There's really no better way to put it.

### DIFF
--- a/packages/xterm.rb
+++ b/packages/xterm.rb
@@ -1,0 +1,24 @@
+require 'package'
+
+class Xterm < Package
+  description 'The xterm program is a terminal emulator for the X Window System.'
+  homepage 'https://invisible-island.net/xterm/'
+  version '333'
+  source_url 'ftp://ftp.invisible-island.net/xterm/xterm-333.tgz'
+  source_sha256 '2f1d42014e55c8036c6b29a847b31d3b5c1a3a35b126993ae6d3f05e8da0ef78'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.build
+    system "./configure", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}"
+    system "make"
+  end
+
+  def self.install
+    system "make install DESTDIR=#{CREW_DEST_DIR}"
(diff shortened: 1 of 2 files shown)


packages/xwayland.rb
```diff
@@ -39,5 +40,6 @@ def self.build
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "ln", "-s", "#{CREW_PREFIX}/bin/Xwayland", "#{CREW_DEST_PREFIX}/bin/X"
```
uberhacker: This looks good.  Can we bump the version?
```
- version '1.19.3'
+ version '1.19.3-1'
```

JL2210: Done.

packages/xterm.rb
```diff
@@ -0,0 +1,24 @@
+require 'package'
+
+class Xterm < Package
+  description 'The xterm program is a terminal emulator for the X Window System.'
+  homepage 'https://invisible-island.net/xterm/'
+  version '333'
+  source_url 'ftp://ftp.invisible-island.net/xterm/xterm-333.tgz'
```
uberhacker: `source_url 'https://invisible-mirror.net/archives/xterm/xterm-333.tgz'` ?

uberhacker: ftp should die a slow and painful death.

JL2210: Along with versioned packages?

uberhacker: What do you mean?  This is a new package, right?

JL2210: OK. Got it.

uberhacker: Curious to know, what is the time difference between:
`wget https://invisible-mirror.net/archives/xterm/xterm-333.tgz`
versus
`wget ftp://ftp.invisible-island.net/xterm/xterm-333.tgz` ?

JL2210: Done.

JL2210:
The time difference should be less than an hour. If you get a versioned
tarball and not the latest code, you should be fine.

uberhacker:
@JL2210: Xwayland has been upgraded already so we should be good.

PR #2517, opened 2018-08-07
Added sommelier package.

JL2210:
Fixes:

#2401 (I don't want to close this issue just yet.)

Should work. Removed dependencies and lines of incompatible code.

Tested on:

i686 [ ]
x86_64 [x]
armv7l [ ]
aarch64 [ ]

PR #2520, opened 2018-08-08
Added sommelier package for native X support

### LINKED ISSUE #2489: TiLP in Chromebrew? (Attempted solutions)

JL2210:
Re-do of #2517 

Fixes #2489 

Adds #Awesome Functionality :)

### DIFF
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -1,0 +1,39 @@
+require 'package'
+
+class Sommelier < Package
+  description 'Sommelier works by redirecting X11 and Wayland programs to the built-in ChromeOS wayland server.'
+  homepage 'https://chromium.googlesource.com/chromiumos/containers/sommelier'
+  version '0.20'
+  source_url 'https://chromium.googlesource.com/chromiumos/containers/sommelier/+/0.20/README?format=TEXT'
+  source_sha256 'b58d799b16d20abf92369fe0749c73f7398996f0afa9933517051778a8bb16c3'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'mesa'
+  depends_on 'pixman'
+  depends_on 'libxkbcommon'
+  depends_on 'xwayland'
+  depends_on 'xkbcomp'
+  
+  def self.build
+    system 'mkdir sommelier'
+    Dir.chdir ("sommelier") do
+      system 'git clone https://chromium.googlesource.com/chromiumos/containers/sommelier -b 0.20 .'
+      system 'sed -i "/sd_notify/d" sommelier.c'
+      system 'sed -i "/sd-daemon/d" sommelier.c'
+      system 'sed -i "s,libsystemd,,g" Makefile'
+      system "sed -i 's,/lib/,/#{ARCH_LIB}/,g' Makefile"
+      system "sed -i 's,-I.,-I. -I#{CREW_PREFIX}/include/pixman-1,g' Makefile"
+      system "make PREFIX=#{CREW_PREFIX} SYSCONFDIR=#{CREW_PREFIX}/etc"
+    end
+  end
+
+  def self.install
+    Dir.chdir ("sommelier") do
+      system "make", "PREFIX=#{CREW_PREFIX}", "SYSCONFDIR=#{CREW_PREFIX}/etc", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    end
+  end
+end
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -3,7 +3,7 @@ require 'package'
 class Sommelier < Package
   description 'Sommelier works by redirecting X11 and Wayland programs to the built-in ChromeOS wayland server.'
   homepage 'https://chromium.googlesource.com/chromiumos/containers/sommelier'
-  version '0.20'
+  version 'f85b7e1'
   source_url 'https://chromium.googlesource.com/chromiumos/containers/sommelier/+/0.20/README?format=TEXT'
   source_sha256 'b58d799b16d20abf92369fe0749c73f7398996f0afa9933517051778a8bb16c3'
 
@@ -20,11 +20,8 @@ class Sommelier < Package
   
   def self.build
     system 'mkdir sommelier'
+    system 'git clone https://chromium.googlesource.com/chromiumos/containers/sommelier -b f85b7e1ca4882c63467c7b3ec2ee65c654eca0d0'
     Dir.chdir ("sommelier") do
-      system 'git clone https://chromium.googlesource.com/chromiumos/containers/sommelier -b 0.20 .'
-      system 'sed -i "/sd_notify/d" sommelier.c'
-      system 'sed -i "/sd-daemon/d" sommelier.c'
-      system 'sed -i "s,libsystemd,,g" Makefile'
       system "sed -i 's,/lib/,/#{ARCH_LIB}/,g' Makefile"
       system "sed -i 's,-I.,-I. -I#{CREW_PREFIX}/include/pixman-1,g' Makefile"
       system "make PREFIX=#{CREW_PREFIX} SYSCONFDIR=#{CREW_PREFIX}/etc"
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -3,7 +3,7 @@ require 'package'
 class Sommelier < Package
   description 'Sommelier works by redirecting X11 and Wayland programs to the built-in ChromeOS wayland server.'
   homepage 'https://chromium.googlesource.com/chromiumos/containers/sommelier'
-  version 'f85b7e1'
+  version '1382c'
   source_url 'https://chromium.googlesource.com/chromiumos/containers/sommelier/+/0.20/README?format=TEXT'
   source_sha256 'b58d799b16d20abf92369fe0749c73f7398996f0afa9933517051778a8bb16c3'
 
@@ -19,8 +19,9 @@ class Sommelier < Package
   depends_on 'xkbcomp'
   
   def self.build
-    system 'git clone https://chromium.googlesource.com/chromiumos/containers/sommelier -b f85b7e1ca4882c63467c7b3ec2ee65c654eca0d0'
+    system 'git clone https://chromium.googlesource.com/chromiumos/containers/sommelier'
     Dir.chdir ("sommelier") do
+      system "git reset --hard 1382ce084cc40790340d672e8b62ec47733cb860"
       system "sed -i 's,/lib/,/#{ARCH_LIB}/,g' Makefile"
       system "sed -i 's,-I.,-I. -I#{CREW_PREFIX}/include/pixman-1,g' Makefile"
       system "make PREFIX=#{CREW_PREFIX} SYSCONFDIR=#{CREW_PREFIX}/etc"
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -19,7 +19,6 @@ class Sommelier < Package
   depends_on 'xkbcomp'
   
   def self.build
-    system 'mkdir sommelier'
     system 'git clone https://chromium.googlesource.com/chromiumos/containers/sommelier -b f85b7e1ca4882c63467c7b3ec2ee65c654eca0d0'
     Dir.chdir ("sommelier") do
       system "sed -i 's,/lib/,/#{ARCH_LIB}/,g' Makefile"


packages/sommelier.rb
```diff
@@ -0,0 +1,39 @@
+require 'package'
+
+class Sommelier < Package
+  description 'Sommelier works by redirecting X11 and Wayland programs to the built-in ChromeOS wayland server.'
+  homepage 'https://chromium.googlesource.com/chromiumos/containers/sommelier'
+  version '0.20'
+  source_url 'https://chromium.googlesource.com/chromiumos/containers/sommelier/+/0.20/README?format=TEXT'
```
uberhacker: Easier to use `source_url 'https://chromium.googlesource.com/chromiumos/containers/sommelier/+archive/0.20.tar.gz'`?  Then you don't need the git clone, correct?

JL2210: No. The shasum is different every time. The README stays the same, because it's so old. They pack their tarballs on the fly and send them to you.

uberhacker: Wouldn't they bump the version after any new changes?  This is a git repo.

JL2210: No. The README literally doesn't have anything in it but `Sommelier is a nested Wayland compositor with X11 forwarding support.`

ukscone: wouldn't it make sense to pull our own copy once a release and host it ourselves somewhere? that way we avoid shasum changing on each pull?

uberhacker: It sounds like the folks at Google need a lesson about tagging new releases.

JL2210: No. If anything, I'll change the file from the README to a hashed version of `install.sh`. What do you think, @uberhacker?

uberhacker: I don't think this is the correct approach.  This is a hashed version: https://chromium.googlesource.com/chromiumos/containers/sommelier/+/0.20

JL2210: Yes, but it's literally a hash. Try to wget it and cat the file.

uberhacker: Not sure I follow.  Any file from this tagged release should remain static.  Are you sure you aren't pulling from the master branch?

JL2210: No. Try this: `wget -q -Ook https://chromium.googlesource.com/chromiumos/containers/sommelier/+/0.20/README?format=TEXT; cat ok; rm ok;`

JL2210: They tag it just fine. If you create a tarball, then create another tarball with the same contents, the hashes are different. They don't save a copy of that tarball, they just make another one when it's requested from their server.

uberhacker: What kind of repo is Google hosting?  This is all messed up.

JL2210: There hasn't been a release in 4 months and they probably won't. They made a newer one that has an icon on the taskbar.

JL2210: It's not like GitHub, which saves a tarball for each release. It just creates one when you ask for it.

JL2210: Saves space, but not friendly to package managers.

JL2210: It works. Why not merge it?

uberhacker: I'm going to play with this a bit before I decide to merge.  I'm not convinced this is the best approach yet.

JL2210:
[sommelier-0.20-chromeos-x86_64.tar.xz.txt](https://github.com/skycocker/chromebrew/files/2268544/sommelier-0.20-chromeos-x86_64.tar.xz.txt)
[sommelier-0.20-chromeos-x86_64.tar.xz.sha256.txt](https://github.com/skycocker/chromebrew/files/2268545/sommelier-0.20-chromeos-x86_64.tar.xz.sha256.txt)

packages/sommelier.rb
JL2210: There's a newer version from https://chromium.googlesource.com/chromiumos/platform2 that I can extract if you give me the time.

uberhacker: I'm definitely up for a newer version.  Close this PR?

JL2210: I got a new version with no need for the `sed`s. The version is the first five characters of the commit ID, though.

JL2210:
Could I edit tilps in this package because it fixes that issue?

uberhacker:
Dude, you need to slow down and test before making a bunch of wild ass commits and saying it works.  You are working too fast and need to take a breather.

JL2210:
Sorry. I will. Could I edit tilp?

uberhacker:
Ask @debrouxl and @superloach what they think.

JL2210:
Ok, guys. Should I edit tilp to make it work with sommelier?

packages/sommelier.rb
```diff
@@ -0,0 +1,36 @@
...
+  depends_on 'mesa'
+  depends_on 'pixman'
+  depends_on 'libxkbcommon'
+  depends_on 'xwayland'
+  depends_on 'xkbcomp'
+  
+  def self.build
+    system 'git clone https://chromium.googlesource.com/chromiumos/containers/sommelier'
+    Dir.chdir ("sommelier") do
+      system "git reset --hard 1382ce084cc40790340d672e8b62ec47733cb860"
```
uberhacker: This could also be `system 'git checkout 59dd453e6efd2c4a0b2b08767c42f47e66a4aaa1'`.  Why are you not picking the latest hash from master?

JL2210: No. The latest hash has some incompatible code. Git gives me `Unknown ref 59dd453e6efd2c4a0b2b08767c42f47e66a4aaa1` when I try the git checkout thing.

uberhacker: What is incompatible?  Just trying to understand why we need an earlier commit.

JL2210: It has a makefile specific to Google that isn't open source, and it requires a header file that Crew doesn't have.

JL2210: Look at the git log: `gyp` `virtwl` https://chromium.googlesource.com/chromiumos/containers/sommelier

uberhacker: Can we add those packages?

JL2210: No. Virtwl is a linux header file, and I can't find gyp anywhere. More trouble than it's worth to fix something that's not broken.

uberhacker: This is an interesting commit a few down from the latest master commit.
```
commit d0a225e2af73483aff02ab9ddb8a710f56ae4349
Author: David Reveman <reveman@chromium.org>
Date:   Thu Apr 5 23:14:05 2018 +0000
    
    Remove libsystemd dependency and add --sd-notify flag.
```

ukscone:
I already went through later versions and because of problems I went back to the commit listed in the crouton doc. ok I could have gone up a few more commits but once they moved it they started adding stuff we don't have, won't have yet and can't have because not everyone will get it

ukscone:
lets just stick with what we know works "properly" with minimum fuss e.g. the commit listed in the crouton doc or at most the 0.20 tagged release and leave it at that. it works, it works reasonably well and we won't be chasing after a moving target unless and until we want to

JL2210:
The crouton doc is what I use. More recent than 0.20 and works with no sed

uberhacker:
Chasing after moving targets is exactly the thing I want to stay away from.  FWIW, the latest commit from master is only 12 commits ahead of the 0.20 tagged release and only a day later.

JL2210:
It's not. It's the latest version that works. Disregard @ukscone.

uberhacker:
After a bit of analysis, it appears commit `1382ce084cc40790340d672e8b62ec47733cb860` is the sweet spot.  Let's just go with that commit and call it a day.  Sound reasonable?

JL2210:
K. Let me rework tilp and we-re all good.

superloach:
That sounds good to me, sorry I haven't quite understood the issue this whole time. It seems like the best plan is to use the one that works! 😛

JL2210:
@superloach, I'm going to use sommelier to make the X for tilp work. I'll get back to this tomorrow.

superloach:
Good deal, this is all very exciting!
Let me know if you need me to test anything :)

ukscone:
@superloach tested it and blown away chromebrew at least twice already. took me a while to get it working originally yesterday/today and built and tested several things. i was happy with using the commit in the doc as whatever works it's "better" that what we had which was nowt

superloach:
I understand how you feel about using the older version, which is better documented. I do think it would be a good idea to see how it works on that version, and then work our way up to a more recent version for stability purposes? Please let me know if I'm misunderstanding the situation :)

ukscone:
@superloach yes you are misunderstanding the situation. i was the one who found the doc and did the initial tests and as i've said use the commit that is in the doc and we know works reasonably well with only one real tweak needed and leave it at that for the time being. i said i have gone up and down a few commits but none worked as well as the named commit. @JL2210 said he'd got 0.20 working with some sed's and that's what i believe he was packaging but i'm pragmatic. we have a commit known to work not just for us but also for crouton so lets stick with it and workout the wrinkles that it'll cause if any in other packages. so far with a hand built SDL 1.2, crew supplied gtk i've got fltk and wxWidgets and apps using those all working so i'm happy that sommelier of the named commit is good enough

https://twitter.com/ukscone/status/1026680995343466496
https://twitter.com/ukscone/status/1026931561856417795
https://twitter.com/ukscone/status/1026946481377488896

i'm quite happy and just waiting for people to decide it's good enough to add to chromebrew

superloach:
Woah alright, that looks amazing!
Quite looking forward to cheering y'all on as you get this sorted out, let me know if there's anything I can help with 😀

uberhacker:
OK guys, I have confirmed this works well.  Quite amazing actually.  I expect to see a lot more gui packages in this project in the future. :)  Thanks again to @JL2210 and @ukscone for their hard work to get this moving along.

JL2210:
Hey, @uberhacker, what about tilp?

ukscone:
Yeah thanks @JL2210 for taking the rough idea and turning it into something useful for all

ukscone:
btw I had to rebuild SDL with X11 enabled (or seemed to have to) to get a couple of things working so do we need a rebuilt or new SDL package with and without X11 enabled?

JL2210:
No. It all works fine.

JL2210:
Crew packages are good with that sort of stuff.

ukscone:
all looks fine with a clean chromebrew install from scratch except for an annoying error at startup

"_XSERVTransmkdir: ERROR: euid != 0,directory /tmp/.X11-unix will not be created."

still  testing with things like wxwidgets and fltk but so far so good.

JL2210:
Yeah. No root with Chromebrew. Can't fix that. Try adding this to the line
that calls sommelierd `2>/dev/null`

superloach:
... so you're saying TiLP just won't work in Chromebrew?
Why can't it properly work with root? (I've tried, btw)

JL2210:
It does. The error message references the tmpdir for X not being owned by
root. The error's not fatal.

JL2210:
Stop assuming stuff.

superloach:
Oh, I thought you were referring to the libusb permission errors, which actually are pretty much fatal (it's useless otherwise).

JL2210:
No, that will not work without root. Try this for a workaround: `sudo chown
-R chronos:chronos /dev /proc /sys`

On Aug 8, 2018 2:48 PM, "superloach" <notifications@github.com> wrote:

Oh, I thought you were referring to the libusb permission errors, which
actually are pretty much fatal (it's useless otherwise).

—
<https://github.com/skycocker/chromebrew/pull/2520#issuecomment-411512391>,
or mute the thread

superloach:
Sorry about that, I thought I was on the wrong thread. 🤦‍♂️
Anyways, I tried that `chown`, and it still doesn't work (after complaining about literally everything in `/proc`, too)

JL2210:
Those are processes. What do you expect? Try maybe doing it with root and
an exact path, like /usr/local/bin/tilp

superloach:
Dies with `(tilp:1406): Gtk-WARNING **: 14:58:58.169: cannot open display: :0`

JL2210:
`source /home/chronos/user/.bashrc && /usr/local/bin/sommelier -X
/usr/local/bin/tilp`

superloach:
Okay, I'm not really sure what I did, but now `tilp` (not as root) works just fine.
So, uh... *if you're having problems try one of the things JL2210 said, one of them might work* 😅

ukscone:
not sure why but with my self built sommelier my gui progs work with crew package sommelier they hang/crash when trying to use network. flummuxing but only thing that's changed

superloach:
Also, what's the deal with the lack of shelf icons? Will this have to wait until we figure out how to make Sommelier launchers?

JL2210:
No way, impossible without something like Crouton integration. How about
wait for a desktop environment.

ukscone:
just an aside added this to my .bashrc and it although hacky seems to fix most of the ugly warnings on startup


if [ ! -d /tmp/.X11-unix ]; then
    mkdir /tmp/.X11-unix
    chmod 1777 /tmp/.X11-unix
    sudo chown root:root /tmp/.X11-unix
fi
if ! pgrep -x "sommelier" > /dev/null
then 
 /usr/local/sbin/sommelierd &
fi

JL2210:
Good idea. Will try to make a PR.

JL2210:
Sweet. That works!

uberhacker:
Would a better solution be to change the process to allow non-root users access to `/tmp/.X11-unix`?  I have a feeling we're just trying to work around a bug.

ukscone:
I added the -R to the /tmp/.X11-unix to the chmod and no real difference to whether it hangs or not but it does seem to be consistant that closing via corner X it hangs, closing via a menu item it doesn't hang. does feel like a bug with the corner X handling as it seems really consistant

JL2210:
Oh, well. ^C works. This is not a part of this issue anymore.

uberhacker:
Right, this is the reason I want the sommelier update in a separate PR; so we don't clog up this thread.  It also helps when searching for issues in the future.

JL2210:
I got it.

PR #2521, opened 2018-08-08
Update sommelier to add daemon and postinstall config

uberhacker:
Followed the instructions at https://github.com/dnschneid/crouton/wiki/Sommelier-(A-more-native-alternative-to-xiwi) to build a more complete sommelier package.

### DIFF
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -3,13 +3,21 @@ require 'package'
 class Sommelier < Package
   description 'Sommelier works by redirecting X11 and Wayland programs to the built-in ChromeOS wayland server.'
   homepage 'https://chromium.googlesource.com/chromiumos/containers/sommelier'
-  version '1382c'
+  version '1382ce'
   source_url 'https://chromium.googlesource.com/chromiumos/containers/sommelier/+/0.20/README?format=TEXT'
   source_sha256 'b58d799b16d20abf92369fe0749c73f7398996f0afa9933517051778a8bb16c3'
 
   binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/sommelier-1382ce-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/sommelier-1382ce-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/sommelier-1382ce-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/sommelier-1382ce-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
+    aarch64: '54fd051594c0f62a1b8af107c23432b7a12a961450741d2c40fe5b5453ae3a10',
+     armv7l: '54fd051594c0f62a1b8af107c23432b7a12a961450741d2c40fe5b5453ae3a10',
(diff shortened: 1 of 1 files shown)


JL2210:
Good idea. I will add an update to tilp that catches all the files.

uberhacker:
@JL2210: You don't need to do anything with tilp.  The dialog box pops up fine if you have the sommelierd daemon running in the background, :)  This should make @superloach happy.

uberhacker:
If you want to make a change, just add `depends_on 'sommelier'` to the tilp package.

JL2210:
Yes, I do. You have a hackish way to track most of the files. I will get them all with a patch that I got from Gentoo.

uberhacker:
The only way you are going to get that working is if you create separate packages for all the libraries which are needed during the build process.  Is it worth it?  Maybe not.

JL2210:
No, it's not. I'll package them all into one. Why do that when only one package needs the libs? I'll do two `make install`s, one to install the files and one to track them. I'll disable tests, because they seem to break some things, and I'll add sha256 digests for all of them. Not simple, but worthy.

JL2210:
@uberhacker, remake this with the sommelier daemon running 2>/dev/null and 1>/dev/null. I don't want this stuff to show up when I'm starting my shell:
```
_FontTransOpen: Unable to Parse address ${prefix}/share/fonts/X11/misc/
_FontTransOpen: Unable to Parse address ${prefix}/share/fonts/X11/TTF/
_FontTransOpen: Unable to Parse address ${prefix}/share/fonts/X11/OTF/
_FontTransOpen: Unable to Parse address ${prefix}/share/fonts/X11/Type1/
_FontTransOpen: Unable to Parse address ${prefix}/share/fonts/X11/100dpi/
_FontTransOpen: Unable to Parse address ${prefix}/share/fonts/X11/75dpi/
```

uberhacker:
It's not a very good idea to hide warnings just because it's annoying.  Why not fix so the warnings go away instead?  Novel concept, I know.  For now, they're just warnings so we can deal with it.

JL2210:
They aren't warnings. The tests make the build fail, so I disable them.

JL2210:
Oh. I thought you were talking about something else. Along with that, crosh doesn't exit at all anymore with that. When you type exit, it just hangs.

uberhacker:
I don't have that problem.

JL2210:
With either or just one?

uberhacker:
With the _"warnings"_ shown.

JL2210:
What package are those fonts from? We could always manually set all of
those to exact paths.

JL2210:
Saw that. Does tilp fail to compile on your end?

On Aug 8, 2018 11:40 AM, "Ed Reel" <notifications@github.com> wrote:

Not sure but I did find out that xwayland was missing the font_util package
dependency.

—
<https://github.com/skycocker/chromebrew/pull/2521#issuecomment-411451799>,
or mute the thread

uberhacker:
No, I was able to build the pre-built binaries.  Anything wrong with those?

JL2210:
Yeah. You need to change the prefix to crew_dest_dir along with
crew_lib_prefix and others

JL2210:
Or fix the configure script to do it that way.

uberhacker:
You are welcome to submit a PR if you cannot live with my _"hacked"_ solution.

JL2210:
K. Will do that. Almost done.

PR #2522, opened 2018-08-08
Add xterm pre-built binaries

uberhacker:
Depends on #2521.

### DIFF
--- a/packages/xterm.rb
+++ b/packages/xterm.rb
@@ -8,9 +8,19 @@ class Xterm < Package
   source_sha256 '2f1d42014e55c8036c6b29a847b31d3b5c1a3a35b126993ae6d3f05e8da0ef78'
 
   binary_url ({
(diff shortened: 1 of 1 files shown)


JL2210:
How does this depend on #2521? It works with xinit and my wayland modifications. Try it.

JL2210:
If anything, could you edit it and add `depends_on 'xwayland'` and `depends_on 'xinit'`?

uberhacker:
Dude, it depends on either but there were no dependencies when you added the package which, as it turns out, was not considered originally.

JL2210:
Sorry. Could you edit that for me, please?

uberhacker:
No need.  The `depends_on 'sommelier'` includes xwayland as a dependency already.

JL2210:
OK. True. If anything, this should be a dependency of xinit. Next up, gnome-session! lol

PR #2523, opened 2018-08-08
Fix corrupt font_util binary package

uberhacker:


skiman6010:
Dang. I can't actually find anything wrong with this PR. Lucky you... ;)

PR #2526, opened 2018-08-08
Update packages and dependencies.

JL2210:
Helps fix 1585. Should be all very well tested. Hope this helps!


uberhacker:
I appreciate your efforts but can you please submit a single PR for each package?  I don't want to merge 20+ packages into master without having pre-built binaries and to generate this many at once is too big a burden.

PR #2529, opened 2018-08-08
Added code package (Visual Studio), and fix gtk_doc

JL2210:
Fixes #
Ready this time.
All dependencies satisfied.
Great with sommelier, not so much with just wayland.

### DIFF
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -2,13 +2,13 @@ require 'package'
 
 class Code < Package
   description 'Visual Studio Code is a source code editor developed by Microsoft for Windows, Linux and macOS.'
-  homepage 'https://code.visualstudio.com/docs/?dv=linux64'
+  homepage 'https://code.visualstudio.com/'
   version '1.25'
   case ARCH
-  when "x86_64"
+  when 'x86_64'
     source_url 'https://go.microsoft.com/fwlink/?LinkID=620884'
     source_sha256 '5856bbebf38aa05d584da4722869bbe507cf123f69f7ffab5f1532d73dbf3438'
-  when "i686"
+  when 'i686'
     source_url 'https://go.microsoft.com/fwlink/?LinkID=620885'
     source_sha256 'af6adc2e2500e50bfebe7ee7b97d661b6e774a590136bf5f89334132a5b292e2'
   else
@@ -21,14 +21,22 @@ class Code < Package
   binary_sha256 ({
   })
 
-  depends_on 'sommelier'
   depends_on 'xterm'
   depends_on 'libgconf'
   depends_on 'xdg_base'
+  depends_on 'sommelier'
 
   def self.install
     case ARCH
-    when 'armv7l', 'aarch64'
+    when 'x86_64', 'i686'
+      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/share/code"
+      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/bin"
+      system "cp", "-rpa", ".", "#{CREW_DEST_PREFIX}/share/code/"
+      system "echo '#!/bin/bash' > #{CREW_DEST_PREFIX}/bin/code"
+      system "echo >> #{CREW_DEST_PREFIX}/bin/code"
+      system "echo 'xterm -e #{CREW_PREFIX}/share/code/bin/code' >> #{CREW_DEST_PREFIX}/bin/code"
+      system "chmod a+x #{CREW_DEST_PREFIX}/bin/code"
+    else
       puts
       puts 'Visual Studio Code is currently not supported on ARM and AArch64.'.lightred
       puts 'Please try HeadMelted.'.lightred
@@ -37,14 +45,6 @@ class Code < Package
       puts 'Happy coding!'.lightgreen
       puts
       return 1
-    else
-      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/lib/code"
-      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/bin"
-      system "cp", "-rpa", ".", "#{CREW_DEST_PREFIX}/lib/code/"
-      system "echo '#!/bin/bash' > #{CREW_DEST_PREFIX}/bin/code"
-      system "echo >> #{CREW_DEST_PREFIX}/bin/code"
-      system "echo 'xterm -e #{CREW_PREFIX}/lib/code/bin/code' >> #{CREW_DEST_PREFIX}/bin/code"
-      system "chmod a+x #{CREW_DEST_PREFIX}/bin/code"
     end
   end
 
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -1,0 +1,49 @@
+require 'package'
+
+class Code < Package
+  description 'Visual Studio Code is a source code editor developed by Microsoft for Windows, Linux and macOS.'
+  homepage 'https://code.visualstudio.com/docs/?dv=linux64'
+  version '1.25'
+  case ARCH
+  when "x86_64"
+    source_url 'https://go.microsoft.com/fwlink/?LinkID=620884'
+    source_sha256 '5856bbebf38aa05d584da4722869bbe507cf123f69f7ffab5f1532d73dbf3438'
+  when "i686"
+    source_url 'https://go.microsoft.com/fwlink/?LinkID=620885'
+    source_sha256 'af6adc2e2500e50bfebe7ee7b97d661b6e774a590136bf5f89334132a5b292e2'
+  else
+    source_url 'https://raw.githubusercontent.com/Microsoft/vscode/1.25.1/README.md'
+    source_sha256 'c1c5e6ec903730a4e116c1cfd83bb29acd227195d97f2ec8944452808232c310'
+  end
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'sommelier'
+  depends_on 'xterm'
+  depends_on 'libgconf'
+
+  def self.install
+    case ARCH
+    when 'armv7l', 'aarch64'
+      puts
+      puts 'Visual Studio Code is currently not supported on ARM and AArch64.'.lightred
+      puts 'Please try HeadMelted.'.lightred
+      puts 'https://code.headmelted.com'.lihghtred
+      puts
+      puts 'Happy coding!'.lightgreen
+      puts
+      return 1
+    else
+      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/lib/code"
+      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/bin"
+      system "cp", "-rpa", ".", "#{CREW_DEST_PREFIX}/lib/code/"
+      system "echo '#!/bin/bash' > #{CREW_DEST_PREFIX}/bin/code"
+      system "echo >> #{CREW_DEST_PREFIX}/bin/code"
+      system "echo 'xterm -e #{CREW_PREFIX}/lib/code/bin/code' >> #{CREW_DEST_PREFIX}/bin/code"
+      system "chmod a+x #{CREW_DEST_PREFIX}/bin/code"
+    end
+  end
+end
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -21,6 +21,7 @@ class Code < Package
   binary_sha256 ({
   })
 
+  depends_on 'gtk2'
   depends_on 'libgconf'
   depends_on 'xdg_base'
   depends_on 'sommelier'
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -48,7 +48,7 @@ class Code < Package
   def self.postinstall
     puts
     puts 'Congratulations! You have installed Visual Studio Code on Chrome OS!'.lightgreen
-    puts 'Now, please reboot to restart the Sommelier Daemon, and run \'code\' to start Visual Studio.'.lightgreen
+    puts 'Now, please run \'code\' to start Visual Studio.'.lightgreen
     puts 'Happy coding!'.lightgreen
     puts
   end
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -21,7 +21,6 @@ class Code < Package
   binary_sha256 ({
   })
 
-  depends_on 'xterm'
   depends_on 'libgconf'
   depends_on 'xdg_base'
   depends_on 'sommelier'
@@ -32,10 +31,7 @@ class Code < Package
       system "mkdir", "-p", "#{CREW_DEST_PREFIX}/share/code"
       system "mkdir", "-p", "#{CREW_DEST_PREFIX}/bin"
       system "cp", "-rpa", ".", "#{CREW_DEST_PREFIX}/share/code/"
-      system "echo '#!/bin/bash' > #{CREW_DEST_PREFIX}/bin/code"
-      system "echo >> #{CREW_DEST_PREFIX}/bin/code"
-      system "echo 'xterm -e #{CREW_PREFIX}/share/code/bin/code' >> #{CREW_DEST_PREFIX}/bin/code"
-      system "chmod a+x #{CREW_DEST_PREFIX}/bin/code"
+      system "ln", "-s", "#{CREW_PREFIX}/share/code/bin/code", "#{CREW_DEST_PREFIX}/bin"
     else
       puts
       puts 'Visual Studio Code is currently not supported on ARM and AArch64.'.lightred
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -47,4 +47,11 @@ class Code < Package
       system "chmod a+x #{CREW_DEST_PREFIX}/bin/code"
     end
   end
+
+  def self.postinstall
+    puts
+    puts 'Congratulations! You have installed Visual Studio Code on Chrome OS!'.lightgreen
+    puts 'Now, please reboot to restart the Sommelier Daemon, and run \'code\' to start Visual Studio.'.lightgreen
+    puts 'Happy coding!'.lightgreen
+    puts
 end
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -54,4 +54,5 @@ class Code < Package
     puts 'Now, please reboot to restart the Sommelier Daemon, and run \'code\' to start Visual Studio.'.lightgreen
     puts 'Happy coding!'.lightgreen
     puts
+  end
 end
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -36,7 +36,7 @@ class Code < Package
       puts
       puts 'Visual Studio Code is currently not supported on ARM and AArch64.'.lightred
       puts 'Please try HeadMelted.'.lightred
-      puts 'https://code.headmelted.com'.lihghtred
+      puts 'https://code.headmelted.com'.lightred
       puts
       puts 'Happy coding!'.lightgreen
       puts
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -24,6 +24,7 @@ class Code < Package
   depends_on 'sommelier'
   depends_on 'xterm'
   depends_on 'libgconf'
+  depends_on 'xdg_base'
 
   def self.install
     case ARCH
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -41,7 +41,7 @@ class Code < Package
       puts
       puts 'Happy coding!'.lightgreen
       puts
-      return 1
+      exit 1
     end
   end
 
--- a/packages/dbus_glib.rb
+++ b/packages/dbus_glib.rb
@@ -13,6 +13,9 @@ class Dbus_glib < Package
   })
 
   depends_on 'gtk_doc'
+  depends_on 'dbus'
+  depends_on 'glib'
+
 
   def self.build
     system "./configure", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}"
--- a/packages/dbus_glib.rb
+++ b/packages/dbus_glib.rb
@@ -1,0 +1,25 @@
+require 'package'
+
+class Dbus_glib < Package
+  description 'The D-Bus GLib package contains GLib interfaces to the D-Bus API.'
+  homepage 'http://www.linuxfromscratch.org/blfs/view/svn/general/dbus-glib.html'
+  version '0.110'
+  source_url 'https://dbus.freedesktop.org/releases/dbus-glib/dbus-glib-0.110.tar.gz'
+  source_sha256 '7ce4760cf66c69148f6bd6c92feaabb8812dee30846b24cd0f7395c436d7e825'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'gtk_doc'
+
+  def self.build
+    system "./configure", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}"
+    system "make"
+  end
+
+  def self.install
+    system "make install DESTDIR=#{CREW_DEST_DIR}"
+  end
+end
--- a/packages/docbook_xsl.rb
+++ b/packages/docbook_xsl.rb
@@ -19,39 +19,38 @@ class Docbook_xsl < Package
   depends_on 'libxml2'
   depends_on 'libxslt'
   depends_on 'docbook_xml'
-  depends_on 'docbook'
 
   def self.build
   end
 
   def self.install
-    system "install -v -m755 -d #{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1 &&
+    system "install -v -m755 -d #{CREW_DEST_PREFIX}/share/xml/docbook/xsl-stylesheets-1.79.1 &&
             cp -v -R . \
-                #{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1/ &&
+                #{CREW_DEST_PREFIX}/share/xml/docbook/xsl-stylesheets-1.79.1/ &&
             
             install -v -m644 -D README \
-                                #{CREW_DEST_DIR}/usr/local/share/doc/docbook-xsl-1.79.1/README.txt &&
+                                #{CREW_DEST_PREFIX}/share/doc/docbook-xsl-1.79.1/README.txt &&
             install -v -m644    RELEASE-NOTES* NEWS* \
-                                #{CREW_DEST_DIR}/usr/local/share/doc/docbook-xsl-1.79.1"
+                                #{CREW_DEST_PREFIX}/share/doc/docbook-xsl-1.79.1"
     
     system "xmlcatalog --noout --add 'rewriteSystem' \
                        'http://docbook.sourceforge.net/release/xsl/1.79.1' \
-                       '#{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1' \
+                       '#{CREW_DEST_PREFIX}/share/xml/docbook/xsl-stylesheets-1.79.1' \
                 /usr/local/etc/xml/catalog &&
             
             xmlcatalog --noout --add 'rewriteURI' \
                        'http://docbook.sourceforge.net/release/xsl/1.79.1' \
-                       '#{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1' \
+                       '#{CREW_DEST_PREFIX}/share/xml/docbook/xsl-stylesheets-1.79.1' \
                 /usr/local/etc/xml/catalog &&
             
             xmlcatalog --noout --add 'rewriteSystem' \
                        'http://docbook.sourceforge.net/release/xsl/current' \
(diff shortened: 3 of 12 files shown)


uberhacker:
Where is libsecret?  That is a dependency I needed.  I will push up a PR soon.

JL2210:
Strange. What does it say?

packages/code.rb
```diff
@@ -0,0 +1,49 @@
...
+    source_sha256 'c1c5e6ec903730a4e116c1cfd83bb29acd227195d97f2ec8944452808232c310'
+  end
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'sommelier'
+  depends_on 'xterm'
```
uberhacker: Why do we need xterm?

packages/code.rb
```diff
@@ -0,0 +1,49 @@
...
+      puts 'Happy coding!'.lightgreen
+      puts
+      return 1
+    else
+      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/lib/code"
+      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/bin"
+      system "cp", "-rpa", ".", "#{CREW_DEST_PREFIX}/lib/code/"
+      system "echo '#!/bin/bash' > #{CREW_DEST_PREFIX}/bin/code"
+      system "echo >> #{CREW_DEST_PREFIX}/bin/code"
+      system "echo 'xterm -e #{CREW_PREFIX}/lib/code/bin/code' >> #{CREW_DEST_PREFIX}/bin/code"
```
uberhacker: The code executable should work without this crazy xterm stuff.  Where did you get these installation instructions?

packages/code.rb
JL2210: It doesn't work without. For some reason, it has to be executed with xterm.

packages/code.rb
JL2210: Googling stuff.

packages/gtk_doc.rb
```diff
@@ -31,14 +31,17 @@ class Gtk_doc < Package
   depends_on 'docbook_xsl'
 
   def self.build
-    system 'pip3 install six'
-    system 'pip install six'
-    system './autogen.sh --with-xml-catalog=/usr/local/etc/xml/catalog'
-    system './configure --with-xml-catalog=/usr/local/etc/xml/catalog'
+    system "./autogen.sh"
+    system "./configure --with-xml-catalog=/usr/local/etc/xml/catalog --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
```
uberhacker:
```
- --with-xml-catalog=/usr/local/etc/xml/catalog
+ --with-xml-catalog=#{CREW_PREFIX}/etc/xml/catalog
```
You should get in the habit of replacing all hardcoded '/usr/local`s with `#{CREW_PREFIX}` instead.

packages/gtk_doc.rb
```diff
@@ -31,14 +31,17 @@ class Gtk_doc < Package
...
     system 'make'
   end
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
   end
+
+  def self.postinstall
+    system "pip3 install --upgrade six --prefix #{CREW_PREFIX}"
+    system "pip2 install --upgrade six --prefix #{CREW_PREFIX}"
```
uberhacker: Why do we need both sixes?

packages/libgconf.rb
```diff
@@ -0,0 +1,25 @@
+require 'package'
+
+class Libgconf < Package
+  description 'GConf is a system for storing application preferences.'
+  homepage 'https://projects-old.gnome.org/gconf/'
+  version '3.2.6'
+  source_url 'ftp://ftp.gnome.org/pub/GNOME/sources/GConf/3.2/GConf-3.2.6.tar.xz'
```
uberhacker: ftp is evil:
```
- ftp://ftp.gnome.org/pub/GNOME/sources/GConf/3.2/GConf-3.2.6.tar.xz
+ https://ftp.gnome.org/pub/GNOME/sources/GConf/3.2/GConf-3.2.6.tar.xz
```

packages/gtk_doc.rb
JL2210: I didn't do it.

packages/gtk_doc.rb
JL2210: Python 2 and 3

packages/dbus_glib.rb
```diff
@@ -0,0 +1,25 @@
...
+  version '0.110'
+  source_url 'https://dbus.freedesktop.org/releases/dbus-glib/dbus-glib-0.110.tar.gz'
+  source_sha256 '7ce4760cf66c69148f6bd6c92feaabb8812dee30846b24cd0f7395c436d7e825'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'gtk_doc'
```
uberhacker: You are missing `dbus` and  `glib` dependencies.  Test on a fresh crew install.

JL2210: OK.

JL2210: LOL, that makes sense.

packages/gtk_doc.rb
JL2210: Got it anyway.

JL2210:
Might need a reboot to work.

JL2210:
This is awesome! It highlights your syntax errors and even gives you suggestions on how to correct them!

JL2210:
Try it!

JL2210:
[screenshot 2018-08-08 at 8 24 04 pm](https://user-images.githubusercontent.com/38843657/43871466-c8f3b694-9b4a-11e8-8fd5-d205415a8029.png)

[hello.c.txt](https://github.com/skycocker/chromebrew/files/2272443/hello.c.txt)

superloach:
This looks awesome, can't wait! For those who prefer it, what do you think of getting Atom in here?
I tried earlier, but my knowledge of build tools is seriously underpowered. Need to work on that. :/

packages/code.rb
```diff
@@ -0,0 +1,57 @@
...
+      system "chmod a+x #{CREW_DEST_PREFIX}/bin/code"
+    end
+  end
+
+  def self.postinstall
+    puts
+    puts 'Congratulations! You have installed Visual Studio Code on Chrome OS!'.lightgreen
+    puts 'Now, please reboot to restart the Sommelier Daemon, and run \'code\' to start Visual Studio.'.lightgreen
```
uberhacker: Actually, you don't need to go to the extreme of rebooting to start the sommelier "daemon".  All you need to do is exit and re-enter your crosh session.

packages/code.rb
```diff
@@ -0,0 +1,57 @@
...
+  })
+
+  depends_on 'sommelier'
+  depends_on 'xterm'
+  depends_on 'libgconf'
+  depends_on 'xdg_base'
+
+  def self.install
+    case ARCH
+    when 'armv7l', 'aarch64'
```
uberhacker: It would be better to have conditional logic that identifies the architecture that is supported rather than unsupported.  So something like:
````
case ARCH
when 'i686', 'x86_64'
  ...
else
  puts
  puts "Visual Studio Code does not currently support your architecture.".lightred
  ...
end
```

packages/code.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'sommelier'
+  depends_on 'xterm'
+  depends_on 'libgconf'
+  depends_on 'xdg_base'
```
uberhacker: Why do we need `xdg_base`?  Are you storing settings in `~/.config`?  We should probably list `depends_on 'sommelier'` last since we don't want users to miss the postinstall message.

packages/code.rb
```diff
@@ -0,0 +1,57 @@
...
+      puts 'Please try HeadMelted.'.lightred
+      puts 'https://code.headmelted.com'.lihghtred
+      puts
+      puts 'Happy coding!'.lightgreen
+      puts
+      return 1
+    else
+      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/lib/code"
+      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/bin"
+      system "cp", "-rpa", ".", "#{CREW_DEST_PREFIX}/lib/code/"
```
uberhacker: Can we use `#{CREW_DEST_PREFIX}/share/code` instead?

packages/code.rb
uberhacker: We don't need xterm.  Try this: `$ /usr/local/lib/code/bin/code`.

JL2210:
Tried that. Has to be run directly, and xterm -e works for that.

packages/code.rb
JL2210: No, the daemon is run in the background. It stays up, in my experience. Also, the shell still hangs when you try to exit the yellow crosh shell.

uberhacker:
It worked fine for me on a fresh crew install after fixing the errors from above.  Again, please test your PR before submitting.  It will save me a lot of effort.

JL2210:
I did. Uncommitted changes.

packages/libgconf.rb
```diff
@@ -0,0 +1,26 @@
...
+  source_url 'https://ftp.gnome.org/pub/GNOME/sources/GConf/3.2/GConf-3.2.6.tar.xz'
+  source_sha256 '1912b91803ab09a5eed34d364bf09fe3a2a9c96751fde03a4e0cfa51a04d784c'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'dbus_glib'
+  depends_on 'gtk2'
```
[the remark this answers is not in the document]
JL2210: I noticed it saying it needed gtk2 in the build process, so I stuck it in there. It works, but I'll change that. It doesn't really have anything to do with the version.

JL2210:
Try upgrading gtk_doc for the issue with libgconf failing to install.

packages/code.rb
JL2210: OK.

packages/code.rb
JL2210: It fails with a JavaScript error if ~/.config doesn't exist, and this package is a good way to provide that. Not terribly unreasonable.

JL2210:
Sure. Sorry about that, but it worked for me, even after a fresh crew.

JL2210:
How's it work now?

ukscone:
@JL2210 re: crosh hanging when exiting. not 100% sure why but after I added the little /tmp/.X11-unix & pgrep hack to my .bashrc I haven't had crosh hang when exiting since

JL2210:
Nice. I will add that if that fixes the issue.

ukscone:
YMMV and I bet now i've said it it will stop doing it but can't hurt to try :)

JL2210:
VMMV? Where was it that you showed me the .X11.unix hack?

uberhacker:
@JL2210: I shouldn't need to upgrade one package to fix another.  Each package should work independently of the other.  With that said, please add the six pip module to libgconf in case other packages need to use this as a dependency.

ukscone:
@JL2210 can't remember/find the post but here it is again

I put it after the settings needed for sommelier e.g. after the DISPLAY setting

-------
export DISPLAY=:0
if [ ! -d /tmp/.X11-unix ]; then
    mkdir /tmp/.X11-unix
    chmod 1777 /tmp/.X11-unix
    sudo chown root:root /tmp/.X11-unix
fi
if ! pgrep -x "sommelier" > /dev/null
then 
 /usr/local/sbin/sommelierd &
fi

ukscone:
n/m as i said it has suddenly started hanging again. smh

JL2210:
No, it works. Thanks!

JL2210:
@uberhacker, Six is a dependency of gtk_doc. It was in the package before, but it was in the build section. I just moved it to the postinstall, added the prefix and changed /usr/local to CREW_PREFIX.

JL2210:
I also bumped the version number.

JL2210:
If you don't mind, I'd really like to fix sommelier here, because the hanging thing is just unbearable. Notice the chmod -R 1777.

ukscone:
@JL2210 it seems to mostly work. there must be an edgecase related with how you close the prog. seems to be consist if you close via a menu option but closing via the corner X seems a toss

JL2210:
Yeah. Don't care. I hated having to manually close the Chrome tab and go through the warning each time because it hung.

packages/libgconf.rb
```diff
@@ -0,0 +1,27 @@
...
+  source_sha256 '1912b91803ab09a5eed34d364bf09fe3a2a9c96751fde03a4e0cfa51a04d784c'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'dbus_glib'
+  depends_on 'gtk3'
+  depends_on 'gtk2'
```
uberhacker: This dependency is not needed here.  The gtk2 dependency is needed for code, not libgconf.

JL2210: I think libgconf has some gtk2 bindings, too. I looked at the configure log.

packages/gtk_doc.rb
uberhacker: We should only need one six module and it should be in the install section, not postinstall.  I know you didn't make these changes initially but as long as we're "fixing" this package...

Ewpratten:
How are you doing the gui stuff? crostini? crouton?

packages/code.rb
uberhacker: Do you have more than one crosh session open?  That maybe the reason.  Besides, rebooting is totally not necessary to start sommelier.  It starts when you login to your shell session from crosh.

packages/libgconf.rb
uberhacker: It looks for gtk3, not gtk2.

packages/sommelier.rb
```diff
@@ -3,7 +3,7 @@
 class Sommelier < Package
```
uberhacker: Please add this change as a separate PR.  We need to keep these things separate because we don't have the binaries yet.  With that said, you will also need to remove the pre-built binaries.

JL2210:
sommeleir.

packages/sommelier.rb
JL2210: No, I won't. I just changed the postinstall.

packages/sommelier.rb
```diff
@@ -48,14 +48,22 @@ def self.install
...
     puts "echo 'export DISPLAY=:0' >> ~/.bashrc".lightblue
-    puts "echo '[ -f #{CREW_PREFIX}/sbin/sommelierd ] && #{CREW_PREFIX}/sbin/sommelierd &' >> ~/.bashrc".lightblue
+    puts "echo 'if [ ! -d /tmp/.X11-unix ]; then' >> ~/.bashrc".lightblue
+    puts "echo 'mkdir /tmp/.X11-unix' >> ~/.bashrc".lightblue
+    puts "echo 'fi' >> ~/.bashrc".lightblue
+    puts "echo 'chmod -R 1777 /tmp/.X11-unix' >> ~/.bashrc".lightblue
+    puts "echo 'sudo chown root:root /tmp/.X11-unix' >> ~/.bashrc".lightblue
+    puts "echo 'if ! pgrep -x \"sommelier\" > /dev/null' >> ~/.bashrc".lightblue
+    puts "echo 'then' >> ~/.bashrc".lightblue
+    puts "echo '/usr/local/sbin/sommelierd &' >> ~/.bashrc".lightblue
```
uberhacker: Please change `/usr/local` to `#{CREW_PREFIX}`.

JL2210:
An X forwarding mechanism for Wayland.

packages/sommelier.rb
JL2210: K

packages/sommelier.rb
JL2210: Just like with gtk_doc. No difference in the files.

uberhacker: Actually that is correct but I still want a separate PR. :)

packages/code.rb
```diff
@@ -0,0 +1,54 @@
...
+    when 'x86_64', 'i686'
+      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/share/code"
+      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/bin"
+      system "cp", "-rpa", ".", "#{CREW_DEST_PREFIX}/share/code/"
+      system "ln", "-s", "#{CREW_PREFIX}/share/code/bin/code", "#{CREW_DEST_PREFIX}/bin"
+    else
+      puts
+      puts 'Visual Studio Code is currently not supported on ARM and AArch64.'.lightred
+      puts 'Please try HeadMelted.'.lightred
+      puts 'https://code.headmelted.com'.lihghtred
```
uberhacker: lihghtred is misspelled.  Obviously, this hasn't been tested.

packages/sommelier.rb
JL2210: I will revert sommelier in this one, and after this is merged I will create a new one.

packages/libgconf.rb
JL2210: OK. Will change.

packages/code.rb
JL2210: The sommelier daemon runs totally in the background. No. This is null anyway, now that I reverted that.

packages/gtk_doc.rb
JL2210: OK. Python 2.

JL2210: Just a warning, gtk_doc fails to build because of XSL stylesheets. Help?

uberhacker: I've run across that problem also.  I just removed the docs with `--disable-gtk-doc` in other packages that had this issue.   Not sure what to do in this case.  Maybe try updating the libxslt package?

JL2210:
I'll just disable gtk docs. That seems ok.

packages/gtk_doc.rb
uberhacker: I have an idea.  Why don't you just remove this package from this PR and we can deal with it in a separate PR?  This doesn't affect the code package AFAICT.

uberhacker: Also, the pre-built binaries need to be removed. :)

JL2210: What does AFAICT mean?

JL2210: No. That wouldn't work. I'll just revert gtk_doc, add a package called six that it depends on.

uberhacker: AFAICT = as far as I can tell

packages/dbus_glib.rb
uberhacker: Remove the `depends_on 'gtk_doc'` dependency?

packages/code.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'gtk2'
+  depends_on 'libgconf'
+  depends_on 'xdg_base'
```
uberhacker: Why do we need xdg_base?

packages/dbus_glib.rb
JL2210: OK.

packages/code.rb
```diff
@@ -0,0 +1,55 @@
...
+      system "ln", "-s", "#{CREW_PREFIX}/share/code/bin/code", "#{CREW_DEST_PREFIX}/bin"
+    else
+      puts
+      puts 'Visual Studio Code is currently not supported on ARM and AArch64.'.lightred
+      puts 'Please try HeadMelted.'.lightred
+      puts 'https://code.headmelted.com'.lightred
+      puts
+      puts 'Happy coding!'.lightgreen
+      puts
+      return 1
```
uberhacker: This is not bash, it's ruby so I'm not sure this is valid.

JL2210:
Can I have a minute to work this out in my head?

packages/code.rb
JL2210: ~/.config

packages/six.rb
```diff
@@ -0,0 +1,22 @@
+require 'package'
+
+class Setuptools < Package
```
uberhacker: Dude, please don't add stuff without testing it first.  You are WAY too fast.  You need to slow down and test before spewing out random commits.  This class does not match the six.rb filename.

JL2210: Sorry. Fixed that. I'm sort of stressed to get this fixed. I have something to do tomorrow. It's done.

packages/six.rb
```diff
@@ -0,0 +1,22 @@
...
+  description 'Six is a Python 2 and 3 compatibility library.'
+  homepage 'https://github.com/benjaminp/six'
+  version '1.11.0'
+  source_url 'https://github.com/skycocker/chromebrew/raw/cc4f47007b6fe702e03de28347e20d3a0bb692f6/install.sh'
```
uberhacker: Why don't you grab the source from `https://github.com/benjaminp/six/archive/1.11.0.tar.gz` and run the setup.py script?

packages/six.rb
uberhacker: Maybe we should just shelve this until you are ready?  It can wait. :)

packages/six.rb
JL2210: Sure.

packages/code.rb
JL2210: How would you do it? I haven't caught on with all the functions, yet.

packages/six.rb
JL2210: I think I'm good now.

packages/code.rb
uberhacker: How about just remove the line? :)

packages/code.rb
uberhacker: Why do we need to restart?  We only need to start, correct?  A reboot should not be necessary so can you please change the wording?  Maybe just remove and simply say: `Run 'code' to start`?  If they followed the postinstall message, sommelier should be running aready at this point.

packages/six.rb
```diff
@@ -0,0 +1,27 @@
...
+  depends_on 'python27'
+
+  def self.build
+    system "python2.7 setup.py build"
+    system "python3 setup.py build"
+  end
+
+  def self.install
+    system "python2.7 setup.py install --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR}"
+    system "python3 setup.py install --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR}"
```
uberhacker: I'm glad you added this package.  There are several other packages that could be updated to use this as a dependency.  Also might be worth making a `setuptools.rb` package as well.  Something to think about.

JL2210: That's actually what I based this package off of originally. I will make another PR.

packages/code.rb
JL2210: OK. What would I do to make it fail then?

uberhacker: You could add `abort` but I'm thinking it might not even be necessary.  Test it out a bit and see what happens.

JL2210:
How about 'exit'?

packages/code.rb
JL2210: Alright. I'll change that. I didn't want a warning to be put out the first time somebody started VS Code.

uberhacker:
This might help: http://blog.honeybadger.io/how-to-exit-a-ruby-program/

JL2210:
Thanks. All *```
awesome
```* now

File: packages/gtk_doc.rb [hunk 21]
```diff
@@ -29,12 +29,11 @@ class Gtk_doc < Package
   depends_on 'docbook'
   depends_on 'docbook_xml'
   depends_on 'docbook_xsl'
+  depends_on 'six'
```
uberhacker: Can we remove the pre-built binaries and bump the version?  Does the six package fix so this compiles cleanly now?

JL2210:
No. Version bump won't matter, crew recursively checks dependencies anyway.
This package is cr&p.

packages/gtk_doc.rb
uberhacker: Actually, there is an update available.  Upgrade to https://github.com/GNOME/gtk-doc/archive/GTK_DOC_1_28.tar.gz?

JL2210: No. Version bump won't matter, crew recursively checks dependencies anyway.
This package is cr&p.

uberhacker: If this package is not needed to get code working, I would rather it be removed from this PR and submitted separately.

uberhacker: Dude, you completely changed the build section.  This needs new binaries.

JL2210: Maybe. I'll try it.

uberhacker: Try upgrading to version 1.28.  Always gotta think about moving forward. :)

JL2210: Nope.

JL2210: Fixed. No more binaries for docbook_xsl, docbook_xml, or docbook ever.

uberhacker: I still see binaries.

JL2210: It works now. Try it.

uberhacker: Dude, there is no difference between binaries and compiling from source.  Well, the difference is binary installs are a LOT faster but we're not gonna compile these each time.  That doesn't make sense.

packages/docbook_xsl.rb
```diff
@@ -33,13 +26,8 @@ def self.build
...
     system "install -v -m755 -d #{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1 &&
-            cp -v -R VERSION assembly common eclipse epub epub3 extensions fo        \
-                     highlighting html htmlhelp images javahelp lib manpages params  \
-                     profiling roundtrip slides template tests tools webhelp website \
-                     xhtml xhtml-1_1 xhtml5                                          \
-                #{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1 &&
-            
-            ln -s VERSION #{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1/VERSION.xsl &&
+            cp -v -R . \
+                #{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1/ &&
```
uberhacker: As long as we're making changes:
```
- #{CREW_DEST_DIR}/usr/local/share
+ #{CREW_DEST_PREFIX}/share
```

packages/dbus_glib.rb
uberhacker: Can we not just remove the gtk_doc dependency here and submit the docbook madness in another PR please?

JL2210:
@uberhacker, #{CREW_PREFIX}/etc/xml/catalog is overwritten by those binary packages. A tarball isn't smart enough to fix that.

tools/needs_binaries.sh
```diff
@@ -7,7 +7,7 @@ cd ../packages
 arch=$(uname -m)
 exclusions="cabal.rb cros_resize.rb docx2txt.rb dpkg.rb fpc.rb freedos.rb ghc.rb kr.rb"
 exclusions+=" libtinfo.rb lsb_release.rb mysqltuner.rb nodebrew.rb nvm.rb oc.rb stack.rb"
-exclusions+=" txt2regex.rb uwsgi.rb xdg_base.rb"
+exclusions+=" txt2regex.rb uwsgi.rb xdg_base.rb docbook.rb docbook_xsl.rb docbook_xml.rb"
```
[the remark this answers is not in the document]
JL2210: No! Why?

tools/generate_binaries.sh
```diff
@@ -40,6 +40,9 @@ arch=$(uname -m)
 # Exclude fake packages
 exclusions="buildessential glibc parallel wayland xorg_proto"
 
+# Exclude packages that binary packages break
+exclusions="$exclusions docbook_xsl docbook_xml docbook"
```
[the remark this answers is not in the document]
JL2210: Why not? It's true. Installing from source is more dynamic than uncompressing binaries over a filesystem, and that's what this package needs.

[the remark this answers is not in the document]
uberhacker: Maybe the issue is all these "docbook" packages were built at different times.  What if we make fresh binaries out of everything at the same time?  Not sure why you are so worried about the catalog.  In any event, please move these out of this PR into a separate PR.

JL2210:
I changed the title. I need to get this working, or else it will break.
Currently, you have me set up to create at least 20+ PRs for all the
updates and fixes I have. I'd like it if you didn't do that.

JL2210:
Another one?

packages/dbus_glib.rb
JL2210: I can't untrack them now. There will be merge conflicts.

uberhacker:
Just create a new branch and submit a new PR.  I can close this to make it easier.  The simplest solution which I mentioned 3 times now is just remove `depends_on 'gtk_doc'` from dbus_glib and call it a day..

JL2210:
OK.

PR #2530, opened 2018-08-09
Fixes the issue with crosh hanging when sommelier is installed

JL2210:
Pretty self explanatory.


Ewpratten:
Any chance that this could be used to change desktop environments?

JL2210:
No.

JL2210:
Maybe run a separate one inside a window.

JL2210:
It's basically like Crouton Integration, except you can't minimize it.

JL2210:
@dososboicwen

JL2210:
Now, with Xwayland and Xinit, then you could, but you need a desktop environment first.

PR #2534, opened 2018-08-10
Add qt package

uberhacker:
With Qt, you can reach all your target platforms – desktop & embedded – with one technology and one codebase, minimizing your time-to-market and maintenance burden. Qt is available under a dual-licensing model — you choose what’s right for your needs.  See https://info.qt.io/download-qt-for-application-development.  Added per @JL2210's request. :)

Tested on:
- [x] i686
- [x] x86_64
- [ ] armv7l (not available)

### DIFF
--- a/packages/qt.rb
+++ b/packages/qt.rb
@@ -1,0 +1,81 @@
+require 'package'
+
+class Qt < Package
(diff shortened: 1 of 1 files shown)


JL2210:
Sweet! Thanks for reviving this!

JL2210:
For ARM, could you possibly build this from source?

uberhacker:
It might be worth a shot.  I can look into it.

PR #2537, opened 2018-08-10
Add start/stop commands to sommelier package

uberhacker:
This should also help prevent multiple instances of sommelier running in more than one crosh session.


JL2210:
Very good. This is really something.

skiman6010:
Yes this is great!

PR #2539, opened 2018-08-11
Added crew_profile as an optional package.

JL2210:
Does support bash, zsh, and fish, but not all 3 at the same time.

Adds a new file, `#{CREW_PREFIX}/etc/profile`, that is sourced after the shell starts.

### DIFF
--- a/packages/crew_profile.rb
+++ b/packages/crew_profile.rb
@@ -9,11 +9,13 @@ class Crew_profile < Package
 
   depends_on "xdg_base"
 
+  ENV['HOME'] = '/home/chronos/user'
+
   def self.install
     system "mkdir", "-p", "#{CREW_DEST_PREFIX}/etc"
     system "mkdir", "-p", "#{CREW_DEST_HOME}"
-    system "touch", "$HOME/.profile" # To not overwrite already installed configuration
-    system "ln", "-s", "$HOME/.profile", "#{CREW_DEST_PREFIX}/etc/profile"
+    system "touch", "#{HOME}/.profile" # To not overwrite already installed configuration
+    system "ln", "-s", "#{HOME}/.profile", "#{CREW_DEST_PREFIX}/etc/profile"
   end
 
   def self.postinstall
--- a/packages/crew_profile.rb
+++ b/packages/crew_profile.rb
@@ -9,13 +9,11 @@ class Crew_profile < Package
 
   depends_on "xdg_base"
 
-  HOME = "/home/chronos/user"
-
   def self.install
     system "mkdir", "-p", "#{CREW_DEST_PREFIX}/etc"
     system "mkdir", "-p", "#{CREW_DEST_HOME}"
-    system "touch", "#{HOME}/.profile" # To not overwrite already installed configuration
-    system "ln", "-s", "#{HOME}/.profile", "#{CREW_DEST_PREFIX}/etc/profile"
+    system "touch", "$HOME/.profile" # To not overwrite already installed configuration
+    system "ln", "-s", "$HOME/.profile", "#{CREW_DEST_PREFIX}/etc/profile"
   end
 
   def self.postinstall
--- a/packages/crew_profile.rb
+++ b/packages/crew_profile.rb
@@ -9,12 +9,10 @@ class Crew_profile < Package
 
   depends_on "xdg_base"
 
-  ENV['HOME'] = '/home/chronos/user'
-
   def self.install
     system "mkdir", "-p", "#{CREW_DEST_PREFIX}/etc"
     system "mkdir", "-p", "#{CREW_DEST_HOME}"
-    system "ln", "-s", "#{HOME}/.profile", "#{CREW_DEST_PREFIX}/etc/profile"
+    system "ln", "-s", "#{ENV['HOME']}/.profile", "#{CREW_DEST_PREFIX}/etc/profile"
   end
 
   def self.postinstall
--- a/packages/crew_profile.rb
+++ b/packages/crew_profile.rb
@@ -11,7 +11,8 @@ class Crew_profile < Package
 
   def self.install
     system "mkdir", "-p", "#{CREW_DEST_PREFIX}/etc"
-    system "touch", "#{CREW_DEST_PREFIX}/etc/profile"
+    system "touch", "#{CREW_DEST_HOME}/.profile"
+    system "ln", "-s", "#{CREW_DEST_HOME}/.profile", "#{CREW_PREFIX}/etc/profile"
   end
 
   def self.postinstall
--- a/packages/crew_profile.rb
+++ b/packages/crew_profile.rb
@@ -9,10 +9,13 @@ class Crew_profile < Package
 
   depends_on "xdg_base"
 
+  HOME = "/home/chronos/user"
+
   def self.install
     system "mkdir", "-p", "#{CREW_DEST_PREFIX}/etc"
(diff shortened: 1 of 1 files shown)


uberhacker:
Sorry, can you please help me understand how this works again?  If I put all my config in `#{CREW_PREFIX}/etc/profile` and source it from `~/.bashrc` and then run `rm -rf /usr/local`, wouldn't I lose all my config?  Have you heard of dotfiles?  A lot of github developers store their local config in a repo to save and use on other systems.  Maybe we can set up something like that to work with Chromebrew?

JL2210:
Good call. I have an idea.

JL2210:
Could you give me a log like this?

```
ls -la ~/ | grep profile 2>/dev/null 1>home.log
```

uberhacker:
This is the only output I get (x86_64):
```
-rw-r--r--   1 chronos chronos            204 Jun 24 15:42 .bash_profile
```

JL2210:
Great. I'll push my latest commit and this is ready.

JL2210:
Oh, wait. I have the symlink backwards.

packages/crew_profile.rb
```diff
@@ -0,0 +1,47 @@
...
+  version '0.4.3'
+  source_url 'https://github.com/skycocker/chromebrew/raw/511ee9cb3138e113df74ddf0dee057ef5d9a46fd/README.md'
+  source_sha256 '27b201cec82d903a1856972e6d7ff1ac58a67c761d729ecd7fd14f24fa9d9901'
+
+  depends_on "xdg_base"
+
+  def self.install
+    system "mkdir", "-p", "#{CREW_DEST_PREFIX}/etc"
+    system "touch", "#{CREW_DEST_HOME}/.profile"
+    system "ln", "-s", "#{CREW_DEST_HOME}/.profile", "#{CREW_PREFIX}/etc/profile"
```
uberhacker: Does Chrome OS even use this on login?  I know it uses `~/.bash_profile` since the default shell is bash.

JL2210: No. Read the postinstall.

packages/crew_profile.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  depends_on "xdg_base"
+
+  HOME = "/home/chronos/user"
+
+  def self.install
+    system "mkdir", "-p", "#{CREW_DEST_PREFIX}/etc"
+    system "mkdir", "-p", "#{CREW_DEST_HOME}"
+    system "touch", "#{HOME}/.profile" # To not overwrite already installed configuration
+    system "ln", "-s", "#{HOME}/.profile", "#{CREW_DEST_PREFIX}/etc/profile"
```
uberhacker: How about plain old `$HOME`?  That works in ruby, actually.  You could also use `ENV['HOME']`.  In any event, use the environment variable instead of setting it explicitly.

JL2210: OK. Got it.

JL2210:
How is it working for you now?

uberhacker:
Haven't tried it yet.  I just found out the python builds don't have pip. :/

JL2210:
How about add a pip package? The versions that used to come with python
were old anyway.

On Aug 11, 2018 12:29 PM, "Ed Reel" <notifications@github.com> wrote:

Haven't tried it yet. I just found out the python builds don't have pip. :/

—
You are receiving this because you authored the thread.
<https://github.com/skycocker/chromebrew/pull/2539#issuecomment-412285921>,
or mute the thread

uberhacker:
I fixed the python binaries to include pip/pip3.  Lesson learnt: Don't have python installed when you build the package or pip will not be included.  Solution: `crew remove python27 python3` first.

JL2210:
Ok. Add a comment about that or remove it in the file.

uberhacker:
Duly noted in the packages.

JL2210:
So, how is this? It works nicely.

uberhacker:
What about mosh?

packages/crew_profile.rb
```diff
@@ -0,0 +1,48 @@
+require 'package'
+
+class Crew_profile < Package
+  description 'A profile for Chromebrew, to be used instead of ~/.*rc'
+  homepage 'https://github.com/skycocker/chromebrew'
+  version '0.4.3'
```
uberhacker: How did you come up with this version?  Seems like 1.0 would make more sense?

JL2210: No. Since it is `A profile for Chromebrew`, I figured it should be correlated to the Chromebrew version.

JL2210: It was a while ago.

JL2210:
mosh?

JL2210:
How does that relate to this?

packages/crew_profile.rb
```diff
@@ -0,0 +1,48 @@
+require 'package'
+
+class Crew_profile < Package
+  description 'A profile for Chromebrew, to be used instead of ~/.*rc'
+  homepage 'https://github.com/skycocker/chromebrew'
+  version "#{CREW_VERSION}"
```
uberhacker: We're gonna need a static version.  This will change when the crew version changes even if there are no changes in this package.

packages/crew_profile.rb
```diff
@@ -0,0 +1,48 @@
...
+    puts
+    puts "echo \"if [ -f #{CREW_PREFIX}/etc/profile ]; then\" >> ~/.zshrc".lightblue
+    puts "echo \"  # Source crew profile\" >> ~/.zshrc".lightblue
+    puts "echo \"  source #{CREW_PREFIX}/etc/profile\" >> ~/.zshrc".lightblue
+    puts "echo \"fi\" >> ~/.zshrc".lightblue
+    puts "source ~/.zshrc".lightblue
+    puts
+    puts "For Fish, execute this:".lightblue
+    puts
+    puts "echo \"if [ -f #{CREW_PREFIX}/etc/profile ]\" >> ~/.config/fish/config.fish".lightblue
```
uberhacker: If fish isn't installed, this will puke, right?  Can we add another check to make sure `~/.config/fish/config.fish` exists before attempting to append?

JL2210: If fish isn't installed, why would they run that?

packages/crew_profile.rb
JL2210: OK.

packages/crew_profile.rb
```diff
@@ -0,0 +1,48 @@
...
+    puts "echo \"  # Source crew profile\" >> ~/.config/fish/config.fish".lightblue
+    puts "echo \"  source #{CREW_PREFIX}/etc/profile\" >> ~/.config/fish/config.fish".lightblue
+    puts "echo \"end\" >> ~/.config/fish/config.fish".lightblue
+    puts "source ~/.config/fish/config.fish".lightblue
+    puts
+    puts "Works with Bash, Zsh, and Fish".lightblue
```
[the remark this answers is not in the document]
JL2210: I guess.

JL2210:
@uberhacker: This fails to install with $HOME as you suggested.

packages/crew_profile.rb
uberhacker: Just saying.

uberhacker:
Use `#{ENV['HOME']}` instead and don't assign it a value.

JL2210:
@uberhacker, I think this is done. Version is 1, uses, at your suggestion, `#{ENV['HOME']}` and is all light blue.

PR #2540, opened 2018-08-11
Update python packages

uberhacker:
- Add six dependency to all relevant packages

Tested updated python27 and python3 (including pip/pip3) on:
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/atk.rb
+++ b/packages/atk.rb
@@ -21,9 +21,7 @@ class Atk < Package
   })
 
   depends_on 'gobject_introspection'
-  depends_on 'diffutils' => :build
-  depends_on 'perl' => :build
-  depends_on 'python27' => :build
+  depends_on 'six' => :build
 
   def self.build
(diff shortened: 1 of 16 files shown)


JL2210:
@uberhacker: Six already depends on python2.7 and python3, so any Python dependencies can be removed.

JL2210:
Aka, `mate_desktop`

uberhacker:
Good call!  I like the way you think.

JL2210:
Wait, if you have `mate_desktop`, and you have `sommelier` and `xinit`, does that mean you can have a working desktop environment in crew?

uberhacker:
There is only one way you can find out. :)

JL2210:
@uberhacker, you put it way better than I could have. @thedamian, have you seen the code package?

uberhacker:
If it helps any, I will accept full responsibility if this breaks anything. :)

JL2210:
And I'll help. I'm glad you're back. Did you get meltdown and spectre resolved at work? 😆

PR #2416, opened 2018-06-09
Add sudo package

uberhacker:
This package solves the problem when running sudo commands that require the root user PATH to find binaries.  Without this installed, sudo will default to use `/sbin` instead of `/usr/local/sbin`.  This would result in sudo executing the system `/sbin/fdisk` instead of the util_linux packaged (newer) `/usr/local/sbin/fdisk`, for example.

PR #2543, opened 2018-08-11
Revert "Add sudo package"

JL2210:
Reverts skycocker/chromebrew#2416

This package shouldn't be needed anymore. Fdisk works.


uberhacker:
Agreed!

PR #2544, opened 2018-08-11
Added cross compiler script at @uberhacker's request

JL2210:
Adds a cross compiler script, build_cross_gcc, in the tools/ directory. Works well, got an i686 cross compiler working.


uberhacker:
Nice!  It has a few flaws but good start!  I'll add another PR to tidy it up a bit.

PR #2545, opened 2018-08-12
Update build_cross_gcc script to use crew specific variables for versions

uberhacker:
Also added some validation and input constraints.  It definitely needs work.  I have been unable to build for any architecture successfully yet.  My hangup is always here:
```
make[1]: Nothing to be done for 'install-target'.
make[1]: Leaving directory '/media/tmp/build-binutils'
Makefile:610: arch/i686-cros-linux/Makefile: No such file or directory
make: *** No rule to make target 'arch/i686-cros-linux/Makefile'.  Stop.
```
@JL2210: Please look at this when you get a chance and try to debug.

Thanks

For reference: https://gcc.gnu.org/install/specific.html

### DIFF
--- a/tools/build_cross_gcc
+++ b/tools/build_cross_gcc
@@ -1,116 +1,140 @@
 #!/bin/sh
 
- set -e
+set -e
 
- case $1 in
+if ! test $1; then
+  echo USAGE:
+  echo \ $0 [options] target arch_type
+  echo \ \-h, --help: Display help
+  echo \ \-v, --version: Display version
+  echo \ \-c, --clean: Clean current directory
+  exit
+fi
 
- -h) echo USAGE:
- echo \ $0 [options] target arch_type kernel_version
- echo \ \-h, --help: Display help
- echo \ \-v, --version: Display version
- echo \ \-c, --clean: Clean current directory
- shift
- exit 0;;
- --help) echo USAGE:
- echo \ $0 target arch_type kernel_version
- echo \ \-h, --help: Display help
- echo \ \-v, --version: Display version
- echo \ \-c, --clean: Clean current directory
- shift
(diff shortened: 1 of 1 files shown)


JL2210:
No way! What happened to my old PR?

uberhacker:
It got tidied up a bit. :)

JL2210:
I'll make some changes. It doesn't need the crew command.

uberhacker:
The crew command makes sure the versions are the same as the latest packages.  We need to simulate the current Chromebrew environment as closely as we can.

JL2210:
It requires 3 arguments. The target, the arch type, such as x86 or arm64, and the kernel version.

uberhacker:
The kernel version is not necessary.  We can get that from the linuxheaders package version.

JL2210:
Except that you don't have the source for linuxheaders. Can you close this and I'll go off of yours?

JL2210:
Notice `cross`, not `cros`.

JL2210:
BTW, glibc223 fails to compile with this. That's why I switched to 228.

uberhacker:
Closing to let @JL2210 fix it better.

uberhacker:
The `cros` means "Chrome OS` and is totally arbitrary.

JL2210:
`cross`, not `cros`.

uberhacker:
Like I said, the value is arbitrary and only there so gcc considers it a "valid" architecture.

JL2210:
OK. yeah. It doesn't really matter. BTW, I was talking about the name of the script.

JL2210:
The versions I use are newer and they all work very well with each other. I have an updated script that uses a newer version of everything except glibc and the kernel version.

PR #2546, opened 2018-08-12
Update isl from 0.19 to 0.20

uberhacker:
Tested on:
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/isl.rb
+++ b/packages/isl.rb
@@ -3,25 +3,24 @@ require 'package'
 class Isl < Package
   description 'Integer Set Library for manipulating sets and relations of integer points bounded by linear constraints'
   homepage 'http://isl.gforge.inria.fr/'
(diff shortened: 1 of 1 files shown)


JL2210:
Why don't you use the tarball from http://isl.gforge.inria.fr ?

uberhacker:
Because it's easier to grab the tagged release from git?  Sorry, that was a joke.  I'll update with the release tarball.  It also doesn't require autogen.sh to create a configure script.

JL2210:
OK. Well, if it has autogen.sh, that might be nice. Our autoconf doesn't work that well.

uberhacker:
Does that look any better @JL2210?

JL2210:
I couldn't have done any better myself. Good job!

PR #2551, opened 2018-08-14
Fix docbook packages

JL2210:
Fixes docbook packages and circular dependencies.

### DIFF
--- a/packages/docbook_xml.rb
+++ b/packages/docbook_xml.rb
@@ -10,65 +10,66 @@ class Docbook_xml < Package
   source_sha256 '23068a94ea6fd484b004c5a73ec36a66aa47ea8f0d6b62cc1695931f5c143464'
 
   binary_url ({
-  # Please do not create binaries for this package, breaks gtk_doc.
   })
   binary_sha256 ({
   })
 
   depends_on 'docbook'
 
-  def self.build
-  end
-
   def self.install
-    system "install -v -d -m755 #{CREW_DEST_PREFIX}/share/xml/docbook/xml-dtd-4.3"
+
+    xml_dtd = 'xml-dtd-4.3'
+    xml_version = 'xml/4.3'
+    vxml = 'V4.3'
+
+    system "install -v -d -m755 #{CREW_DEST_PREFIX}/share/xml/docbook/#{xml_dtd}"
     system "install -v -d -m755 #{CREW_DEST_PREFIX}/etc/xml"
-    system "cp -rpa . #{CREW_DEST_PREFIX}/share/xml/docbook/xml-dtd-4.3/"
+    system "cp -rpa . #{CREW_DEST_PREFIX}/share/xml/docbook/#{xml_dtd}/"
     system "rm -f #{CREW_PREFIX}/etc/xml/docbook && \
                 xmlcatalog --noout --create #{CREW_DEST_PREFIX}/etc/xml/docbook && \
             xmlcatalog --noout --add 'public' \
-                '-//OASIS//DTD DocBook XML V4.3//EN' \
-                'http://www.oasis-open.org/docbook/xml/4.3/docbookx.dtd' \
+                '-//OASIS//DTD DocBook XML #{vxml}//EN' \
+                'http://www.oasis-open.org/docbook/#{xml_version}/docbookx.dtd' \
                 #{CREW_DEST_PREFIX}/etc/xml/docbook &&
             xmlcatalog --noout --add 'public' \
-                '-//OASIS//DTD DocBook XML CALS Table Model V4.3//EN' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/xml-dtd-4.3/calstblx.dtd' \
+                '-//OASIS//DTD DocBook XML CALS Table Model #{vxml}//EN' \
+                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}/calstblx.dtd' \
                 #{CREW_DEST_PREFIX}/etc/xml/docbook &&
             xmlcatalog --noout --add 'public' \
                 '-//OASIS//DTD XML Exchange Table Model 19990315//EN' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/xml-dtd-4.3/soextblx.dtd' \
+                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}/soextblx.dtd' \
                 #{CREW_DEST_PREFIX}/etc/xml/docbook &&
             xmlcatalog --noout --add 'public' \
-                '-//OASIS//ELEMENTS DocBook XML Information Pool V4.3//EN' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/xml-dtd-4.3/dbpoolx.mod' \
+                '-//OASIS//ELEMENTS DocBook XML Information Pool #{vxml}//EN' \
+                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}/dbpoolx.mod' \
                 #{CREW_DEST_PREFIX}/etc/xml/docbook &&
             xmlcatalog --noout --add 'public' \
-                '-//OASIS//ELEMENTS DocBook XML Document Hierarchy V4.3//EN' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/xml-dtd-4.3/dbhierx.mod' \
+                '-//OASIS//ELEMENTS DocBook XML Document Hierarchy #{vxml}//EN' \
+                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}/dbhierx.mod' \
                 #{CREW_DEST_PREFIX}/etc/xml/docbook &&
             xmlcatalog --noout --add 'public' \
-                '-//OASIS//ELEMENTS DocBook XML HTML Tables V4.3//EN' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/xml-dtd-4.3/htmltblx.mod' \
+                '-//OASIS//ELEMENTS DocBook XML HTML Tables #{vxml}//EN' \
+                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}/htmltblx.mod' \
                 #{CREW_DEST_PREFIX}/etc/xml/docbook &&
             xmlcatalog --noout --add 'public' \
-                '-//OASIS//ENTITIES DocBook XML Notations V4.3//EN' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/xml-dtd-4.3/dbnotnx.mod' \
+                '-//OASIS//ENTITIES DocBook XML Notations #{vxml}//EN' \
+                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}/dbnotnx.mod' \
                 #{CREW_DEST_PREFIX}/etc/xml/docbook &&
             xmlcatalog --noout --add 'public' \
-                '-//OASIS//ENTITIES DocBook XML Character Entities V4.3//EN' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/xml-dtd-4.3/dbcentx.mod' \
+                '-//OASIS//ENTITIES DocBook XML Character Entities #{vxml}//EN' \
+                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}/dbcentx.mod' \
                 #{CREW_DEST_PREFIX}/etc/xml/docbook &&
             xmlcatalog --noout --add 'public' \
-                '-//OASIS//ENTITIES DocBook XML Additional General Entities V4.3//EN' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/xml-dtd-4.3/dbgenent.mod' \
+                '-//OASIS//ENTITIES DocBook XML Additional General Entities #{vxml}//EN' \
+                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}/dbgenent.mod' \
                 #{CREW_DEST_PREFIX}/etc/xml/docbook &&
             xmlcatalog --noout --add 'rewriteSystem' \
-                'http://www.oasis-open.org/docbook/xml/4.3' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/xml-dtd-4.3' \
+                'http://www.oasis-open.org/docbook/#{xml_version}' \
+                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}' \
                 #{CREW_DEST_PREFIX}/etc/xml/docbook &&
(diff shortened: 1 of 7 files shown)


packages/docbook_xml.rb
```diff
@@ -5,100 +5,93 @@
...
   binary_sha256 ({
-    aarch64: 'd52eb60b394b51945c4ebcefe915664fb835308e68e81a03f63da6b524f2b8db',
-     armv7l: 'd52eb60b394b51945c4ebcefe915664fb835308e68e81a03f63da6b524f2b8db',
-       i686: '432c69d99add5b330b8c3a43bbf937e10a2cee93081296e7fb43cd2e35cf3bd6',
-     x86_64: '43652bc703c27c907372133db6e93c39a90b8d671a94e6136a3c2e453672a02d',
   })
 
-  depends_on 'unzip'
+  depends_on 'unzip' => :build
   depends_on 'libxml2'
```
uberhacker: Actually, you could remove both of these dependencies since they're already in Chromebrew "core".

packages/docbook_xml.rb
```diff
@@ -5,100 +5,93 @@
...
   def self.install
-    system "install -v -d -m755 #{CREW_DEST_DIR}/usr/local/share/xml/docbook/xml-dtd-4.3"
-    system "install -v -d -m755 #{CREW_DEST_DIR}/usr/local/etc/xml"
+    system "install -v -d -m755 #{CREW_DEST_PREFIX}/share/xml/docbook/xml-dtd-4.3"
+    system "install -v -d -m755 #{CREW_DEST_PREFIX}/etc/xml"
  
-    system "if [ ! -e #{CREW_DEST_DIR}/usr/local/etc/xml/docbook ]; then
-                xmlcatalog --noout --create #{CREW_DEST_DIR}/usr/local/etc/xml/docbook
+    system "if [ ! -e #{CREW_PREFIX}/etc/xml/docbook ]; then
+                xmlcatalog --noout --create #{CREW_PREFIX}/etc/xml/docbook
```
uberhacker: This should be `#{CREW_DEST_PREFIX}`.

packages/docbook_xsl.rb
```diff
@@ -5,65 +5,50 @@
...
   })
   binary_sha256 ({
-    aarch64: 'dc78b57e5e4219ea92f0c2f077806c8f1f88f66577cec829d23691319d812ad6',
-     armv7l: 'dc78b57e5e4219ea92f0c2f077806c8f1f88f66577cec829d23691319d812ad6',
-       i686: '46834cf7ee7510cd02417a50b874fe0aceb806fc656ca13600e3bfa7dee25c5c',
-     x86_64: '6713dae4909337fde776f610cf3dee9b6b0ca88a8b35ffb02cc213022f6b85ba',
   })
 
-  depends_on 'unzip'
   depends_on 'libxml2'
```
uberhacker: You could remove this also.

packages/libxslt.rb
```diff
@@ -3,30 +3,25 @@
...
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/libxslt-1.1.32-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'c9edcdbefad60a9b0b054fb8570adf6d7d27f711bd78fc7d608e9fca789cbba4',
-     armv7l: 'c9edcdbefad60a9b0b054fb8570adf6d7d27f711bd78fc7d608e9fca789cbba4',
-       i686: 'acd156aaa74e26070d956fb5cdb6d7ef4613b3cb38f7ae7f48c3998987a5e88d',
-     x86_64: '3b34fc2e1c0178f51017d739ccb3765e3dc360f30c5a578942c9da612ed0c92a',
   })
 
   depends_on 'libxml2'
```
uberhacker: This can be removed.

packages/docbook_xml.rb
JL2210: OK.

packages/docbook_xml.rb
JL2210: So? It overwrites the other catalog.

packages/docbook_xsl.rb
JL2210: OK.

packages/libxslt.rb
JL2210: Fine.

packages/docbook_xml.rb
uberhacker: Do you see the catalog with `crew files docbook`.  If not, then we need it installed this way.

tools/generate_binaries.sh
```diff
@@ -40,6 +40,9 @@ arch=$(uname -m)
 # Exclude fake packages
 exclusions="buildessential glibc parallel wayland xorg_proto"
 
+# Exclude packages that binary packages break
+exclusions="$exclusions docbook_xsl docbook_xml docbook"
```
uberhacker: I'm still opposed to doing this.  If you are worried about the docbook catalog, let's try to create and append correctly in the binaries.

tools/needs_binaries.sh
```diff
@@ -7,7 +7,7 @@ cd ../packages
 arch=$(uname -m)
 exclusions="cabal.rb cros_resize.rb docx2txt.rb dpkg.rb fpc.rb freedos.rb ghc.rb kr.rb"
 exclusions+=" libtinfo.rb lsb_release.rb mysqltuner.rb nodebrew.rb nvm.rb oc.rb stack.rb"
-exclusions+=" txt2regex.rb uwsgi.rb xdg_base.rb"
+exclusions+=" txt2regex.rb uwsgi.rb xdg_base.rb docbook.rb docbook_xsl.rb docbook_xml.rb"
```
uberhacker: We don't want to compile these packages everytime they need installed.  That doesn't make sense.

JL2210: What does this file do anyway? I thought it builds binaries.

tools/generate_binaries.sh
JL2210: Append, per say, doesn't work. With this, it could work, but the files won't be tracked. Xml has a very confusing format. It's like C and JSON. It has to have an end.

packages/docbook_xsl.rb
```diff
@@ -5,66 +5,50 @@
...
   def self.install
-    system "install -v -m755 -d #{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1 &&
-            cp -v -R VERSION assembly common eclipse epub epub3 extensions fo        \
-                     highlighting html htmlhelp images javahelp lib manpages params  \
-                     profiling roundtrip slides template tests tools webhelp website \
-                     xhtml xhtml-1_1 xhtml5                                          \
-                #{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1 &&
-            
-            ln -s VERSION #{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1/VERSION.xsl &&
+    system "install -v -m755 -d #{CREW_DEST_PREFIX}/share/xml/docbook/xsl-stylesheets-1.79.1 &&
```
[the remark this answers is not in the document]
JL2210: Where?

packages/docbook_xml.rb
JL2210: I could do that. It wouldn't work well, but it's possible.

JL2210: And again, it wouldn't allow for binaries.

tools/needs_binaries.sh
uberhacker: What you are doing is excluding docbook packages from the binary builds.  We don't want to do that. :)

tools/generate_binaries.sh
uberhacker: What happens if we try to build all the binaries in a certain order.  Would that make sense?

tools/needs_binaries.sh
JL2210: Yes, we do! I could make a workaround, if you are willing to put it in the postinstall.

tools/generate_binaries.sh
JL2210: Maybe. We could also try the postinstall.

tools/needs_binaries.sh
JL2210: I reverted this.

tools/generate_binaries.sh
JL2210: I changed this.

JL2210:
@uberhacker, how does this work now?

tools/needs_binaries.sh
JL2210: BTW, it doesn't 'compile' them, per se. It copies the docbook files to a directory and then creates a catalog.

uberhacker:
You cannot include the pre-made catalog in with the build?  Seems to me this could be done with `#{CREW_DEST_PREFIX}` instead of `#{CREW_PREFIX}`.  Isn't this a "set and forget" catalog build?

JL2210:
No, it would be better if we could let xsltproc deal with it. I say just put it in the postinstall and leave out a lot of pain.

JL2210:
@uberhacker, this should work now, including the precompiled binaries. Ready?

JL2210:
@uberhacker? It's ready. The pre-built binaries work. Could you test this and confirm that?

JL2210:
@uberhacker... this has been ready for a long while. Test these out and see if they work.

uberhacker:
This is not ready.  Please read the comments I added 5 days ago.

JL2210:
Which ones?

JL2210:
I copy the catalog into CREW_DEST_PREFIX afterwards. The catalogs are tracked in both.

JL2210:
The actual catalogs, not placeholders.

JL2210:
I removed the references from `tools/`, and I removed the unneeded dependencies.

packages/docbook_xml.rb
```diff
@@ -5,100 +5,89 @@
...
-                'file:///usr/local/share/xml/docbook/xml-dtd-4.3' \
-                #{CREW_DEST_DIR}/usr/local/etc/xml/docbook"
+                'file://#{CREW_PREFIX}/share/xml/docbook/xml-dtd-4.3' \
+                #{CREW_DEST_PREFIX}/etc/xml/docbook"
     
-    system "if [ ! -e #{CREW_DEST_DIR}/usr/local/etc/xml/catalog ]; then
-                xmlcatalog --noout --create #{CREW_DEST_DIR}/usr/local/etc/xml/catalog
-            fi &&
+    system "rm -f #{CREW_PREFIX}/etc/xml/catalog && \
+                xmlcatalog --noout --create #{CREW_DEST_PREFIX}/etc/xml/catalog && \
```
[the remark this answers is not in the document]
uberhacker: I guess my point is we don't need the `system "rm -f #{CREW_PREFIX}/etc/xml/catalog && \`.

packages/libxslt.rb
```diff
@@ -3,30 +3,24 @@
...
-     armv7l: 'c9edcdbefad60a9b0b054fb8570adf6d7d27f711bd78fc7d608e9fca789cbba4',
-       i686: 'acd156aaa74e26070d956fb5cdb6d7ef4613b3cb38f7ae7f48c3998987a5e88d',
-     x86_64: '3b34fc2e1c0178f51017d739ccb3765e3dc360f30c5a578942c9da612ed0c92a',
   })
 
-  depends_on 'libxml2'
+  depends_on 'libxml2_python'
+  depends_on 'libgcrypt'
+  depends_on 'docbook'
+  depends_on 'docbook_xsl'
```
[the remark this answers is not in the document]
JL2210: Yeah.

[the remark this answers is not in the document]
JL2210: Sure.

packages/docbook_xsl.rb
```diff
@@ -5,66 +5,50 @@
...
+                #{CREW_DEST_PREFIX}/share/xml/docbook/xsl-stylesheets-1.79.1/"
+    system "install -v -m644 -D README \
+                                #{CREW_DEST_PREFIX}/share/doc/docbook-xsl-1.79.1/README.txt &&
             install -v -m644    RELEASE-NOTES* NEWS* \
-                                #{CREW_DEST_DIR}/usr/local/share/doc/docbook-xsl-1.79.1"
-    
+                                #{CREW_DEST_PREFIX}/share/doc/docbook-xsl-1.79.1"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/etc/xml"
+    system "mv #{CREW_PREFIX}/etc/xml/catalog #{CREW_DEST_PREFIX}/etc/xml/"
+    system "mv #{CREW_PREFIX}/etc/xml/docbook #{CREW_DEST_PREFIX}/etc/xml/"
```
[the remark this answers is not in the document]
JL2210: Nope. It adds to it when in `/usr/local` and then modes it to CREW_DEST_PREFIX.

packages/docbook_xml.rb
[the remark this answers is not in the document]
JL2210: It creates the docbook catalog.

JL2210:
@uberhacker, I implemented your fixes. It needs a few more variables than just 1, though.

packages/docbook_xsl.rb
```diff
@@ -5,66 +5,50 @@
...
+                #{CREW_PREFIX}/etc/xml/catalog &&
             
             xmlcatalog --noout --add 'rewriteURI' \
                        'http://docbook.sourceforge.net/release/xsl/current' \
-                       '#{CREW_DEST_DIR}/usr/local/share/xml/docbook/xsl-stylesheets-1.79.1' \
-                /usr/local/etc/xml/catalog"
-
+                       '#{CREW_PREFIX}/share/xml/docbook/#{xsl_stylesheets}' \
+                #{CREW_PREFIX}/etc/xml/catalog"
+    system "mv #{CREW_PREFIX}/etc/xml/catalog #{CREW_DEST_PREFIX}/etc/xml/"
```
uberhacker: Don't we want to create the catalog in a single package instead of 2 separate packages that can tread on each other?

JL2210: No, the catalog for docbook_xml will only ever be in docbook_xml and the catalog for docbook_xsl will only ever be in docbook_xsl.

JL2210: Besides, their needs are different.

uberhacker: Unless I'm looking at this wrong, `#{CREW_PREFIX}/etc/xml/catalog` is copied in 2 separate packages.

JL2210: Exactly. The XSL and XML-enabled catalog is in docbook_xsl, while the XML-only catalog is in docbook_xml.

JL2210: The binaries won't work for either packages if the catalogs aren't tracked in both.

JL2210:
I made a few changes. @uberhacker, what is the status of this now?

PR #2552, opened 2018-08-14
Fixed mesa compilation and updated.

JL2210:
That's about it.

### DIFF
--- a/packages/mesa.rb
+++ b/packages/mesa.rb
@@ -3,21 +3,13 @@ require 'package'
 class Mesa < Package
   description 'Open-source implementation of the OpenGL specification'
   homepage 'https://www.mesa3d.org'
-  version '17.3.6-0'
-  source_url 'https://mesa.freedesktop.org/archive/mesa-17.3.6.tar.xz'
-  source_sha256 'e5915680d44ac9d05defdec529db7459ac9edd441c9845266eff2e2d3e57fbf8'
+  version '18.1.5'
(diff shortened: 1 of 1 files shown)


packages/mesa.rb
```diff
@@ -45,8 +37,8 @@ def self.build
                 --prefix=#{CREW_PREFIX} \
                 --libdir=#{CREW_LIB_PREFIX} \
                 --enable-shared-glapi \
-                --with-gallium-drivers=i915,r300,r600,radeon,radeonsi,nouveau,svga,swrast,virgl \
-                --with-dri-drivers=i915,i965,nouveau,swrast,r200,radeon \
+                --with-gallium-drivers=i915,r300,r600,radeonsi,nouveau,svga,swrast,virgl \
+                --with-dri-drivers=i915,i965,nouveau,swrast,r200 \
```
skiman6010: Just curious, do you know why dropping radeon fixes the package?

uberhacker: What happens if we just remove these lines?  Does it autodetect the current drivers?  If that is the case, then specifying drivers is not necessary.

JL2210:
No, it fails with `'radeon' is not a valid target` with those lines.

JL2210:
No, it fails with `'radeon' is not a valid target` with radeon. It still includes radeonsi.

packages/mesa.rb
JL2210: No. The drivers for all Chromebooks are different. This is good.

uberhacker: Is this driver list exhaustive?

PR #2556, opened 2018-08-15
Removed Python support in vim.

### LINKED ISSUE #2554: Vim Complaining about Libpython
```
vim: error while loading shared libraries: libpython3.6m.so.1.0: cannot open shared object file: No such file or directory
```

skiman6010:
Removed support for Python3 as a stop-gap until the Python issue can be sorted out.

Fixes #2554 but not ideal.

### DIFF
--- a/packages/vim.rb
+++ b/packages/vim.rb
@@ -3,21 +3,15 @@ require 'package'
 class Vim < Package
   description 'Vim is a highly configurable text editor built to make creating and changing any kind of text very efficient.'
   homepage 'http://www.vim.org/'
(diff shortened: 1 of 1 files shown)


uberhacker:
@skiman6010: Thank you for the update.  I think a better solution would be to upgrade to 8.1.  I'll try and submit a new PR later today.

skiman6010:
I think the best solution would be to not break my vim installation!! ;)

uberhacker:
Growing pains, my friend. :)

PR #2560, opened 2018-08-17
Update vim and add xdg_base support

uberhacker:
I didn't bump the version since we merged it eariler today.


skiman6010:
Neat! I was excited to see the rumored rewrite of the vim package!

PR #2561, opened 2018-08-17
Updated wine.rb

skiman6010:
Updated wine to 3.13 and enabled X-server support.

### DIFF
--- a/packages/wine.rb
+++ b/packages/wine.rb
@@ -22,7 +22,7 @@ class Wine < Package
     when "i686", "armv7l", "aarch64"
       system "./configure"
     when "x86_64"
-      system "./configure --without-freetype --enable-win64"
+      system "./configure --enable-win64"
     else
       abort "Error getting your device configuration."
     end
--- a/packages/wine.rb
+++ b/packages/wine.rb
@@ -4,21 +4,13 @@ require 'fileutils'
 class Wine < Package
   description 'Wine (originally an acronym for "Wine Is Not an Emulator") is a compatibility layer capable of running Windows applications on several POSIX-compliant operating systems, such as Linux, macOS, & BSD.'
   homepage 'https://www.winehq.org/'
-  version '2.10-1'
-  source_url 'https://dl.winehq.org/wine/source/2.x/wine-2.10.tar.xz'
-  source_sha256 '488df7ffd2e81da455bf428fc9eb784bb4273a890334500895665711bd52f179'
+  version '3.13'
+  source_url 'https://dl.winehq.org/wine/source/3.x/wine-3.13.tar.xz'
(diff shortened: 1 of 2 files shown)


uberhacker:
@skiman6010: Thank you for the update!  There are a ton of things you can enable in wine.  Have you looked through all the options yet?  I think this project could definitely benefit from a few of those.  Any particular reason you disabled freetype support?  If you received an error during the compile process, please post here.  Maybe we can help resolve.

JL2210:
You could use the stable version.

packages/wine.rb
```diff
@@ -4,21 +4,13 @@
...
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/wine-2.10-1-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'de6c0bdaf7e2b913a4170588eb1474b34ef48cca4de848aabd8e1b875c6323b4',
-     armv7l: 'de6c0bdaf7e2b913a4170588eb1474b34ef48cca4de848aabd8e1b875c6323b4',
-       i686: 'f741ba56e1d18d26813feab82a2997017ae3325e19cf449c32a61589717178dd',
-     x86_64: '217be92db577ede5038f0bceb1800453bad4e89f2af8f26cdab6afe99384ae2a',
   })
 
   depends_on 'bison'
```
uberhacker: Don't we also need `depends_on 'sommelier'`?

skiman6010: Oh yeah good call my dude.

skiman6010:
I debated back and forth with myself, but wine moves so quickly and fixes so much in the development version that I think keeping up with the changes would be worth the extra effort to get the most compatibility out of the package.

packages/wine.rb
```diff
@@ -4,33 +4,26 @@
...
   def self.build
     case ARCH
     when "i686", "armv7l", "aarch64"
-      system "./configure --without-x"
+      system "./configure"
     when "x86_64"
-      system "./configure --without-x --enable-win64"
+      system "./configure --enable-win64"
     else
       abort "Error getting your device configuration."
```
uberhacker: It might be more appropriate to have a message like:
```
abort "#{ARCH} architecture not supported.".lightred
```

skiman6010:
I mean it compiles fine and it runs with sommelier support. Not sure what we should disable until we run into errors that we find by deploying it to the users. I can't test every use case and exe.

uberhacker:
I'm not taking about disabling anything.  I'm just saying we should include more dependencies, otherwise we could be missing out on some important added functionality.

JL2210:
@skiman6010: Capability, not compatibility. This is another thing that you will have to constantly have to update if you do this, you know. Just a warning.

PR #2565, opened 2018-08-18
Add cups package

### LINKED ISSUE #10: cups package
A cups package would be a nice addition if cups can work on chromeos.  It should.

uberhacker:
CUPS is the standards-based, open source printing system developed by Apple Inc. for macOS® and other UNIX®-like operating systems. CUPS uses the Internet Printing Protocol (IPP) to support printing to local and network printers.  See https://www.cups.org/.  Resolves #10.

### DIFF
--- a/packages/cups.rb
+++ b/packages/cups.rb
@@ -1,0 +1,97 @@
+require 'package'
+
+class Cups < Package
+  description 'CUPS is the standards-based, open source printing system'
+  homepage 'https://www.cups.org/'
+  version '2.2.8'
+  source_url 'https://github.com/apple/cups/releases/download/v2.2.8/cups-2.2.8-source.tar.gz'
+  source_sha256 '3968fc1d26fc48727508db1c1380e36c6694ab90177fd6920aec5f6cc73af9e4'
+
+  binary_url ({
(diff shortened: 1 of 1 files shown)


JL2210:
If you don't mind, I'd like to use the stable version. I've had some issues with the unstable.

JL2210:
Also, `depends_on 'libjpeg'`

uberhacker:
What were the issues?  This package has a complex build.

JL2210:
No it doesn't. In my case, I've been using the stable version and it's a simple `./configure  && make && make install` process.

JL2210:
Well, now that I see yours, I realize that's what it is. Could you still by chance use stable? I am making a new package, `hplip`, that doesn't seem to recognize the beta.

uberhacker:
Fair enough.  I didn't like the fact that they hardcoded some paths in the configure script.  It's something that really grinds my gears.  https://www.youtube.com/watch?v=CkK0z2ALU_4

JL2210:
Neither did I.

JL2210:
Sed is a real lifesaver with things like that.

PR #2572, opened 2018-08-20
Added more dependencies to wine.rb for more options.

skiman6010:
Let me know if you think any other packages would be worth adding for support.

### DIFF
--- a/packages/wine.rb
+++ b/packages/wine.rb
@@ -14,9 +14,17 @@ class Wine < Package
   })
 
   depends_on 'bison'
-  depends_on 'sommelier'
   depends_on 'flex'
   depends_on 'freetype'
+  depends_on 'ncurses'
+  depends_on 'cups'
+  depends_on 'gettext'
+  depends_on 'libgudev'
+  depends_on 'mesa'
+  depends_on 'krb5'
+  depends_on 'libxrandr'
+  depends_on 'zlibpkg'
+  depends_on 'sommelier'
 
   def self.build
     case ARCH
@@ -25,7 +33,7 @@ class Wine < Package
     when "x86_64"
       system "./configure --enable-win64"
     else
-      abort "Error getting your device configuration."
+	  abort "#{ARCH} architecture not supported.".lightred
     end
     system "make"
   end


uberhacker:
Not sure if they came along with any of the other dependencies but `alsa_lib`, `libglu`, `libpng`, `libjpeg`, `libtiff`?

packages/wine.rb
```diff
@@ -14,9 +14,17 @@ class Wine < Package
   })
 
   depends_on 'bison'
-  depends_on 'sommelier'
   depends_on 'flex'
   depends_on 'freetype'
+  depends_on 'ncurses'
+  depends_on 'cups'
+  depends_on 'gettext'
```
uberhacker: This is "core" so you can exclude.

packages/wine.rb
```diff
@@ -14,9 +14,17 @@ class Wine < Package
...
+  depends_on 'libgudev'
+  depends_on 'mesa'
+  depends_on 'krb5'
+  depends_on 'libxrandr'
+  depends_on 'zlibpkg'
```
uberhacker: This is "core" so you can exclude.

packages/wine.rb
```diff
@@ -14,9 +14,17 @@ class Wine < Package
...
-  depends_on 'sommelier'
   depends_on 'flex'
   depends_on 'freetype'
+  depends_on 'ncurses'
+  depends_on 'cups'
+  depends_on 'gettext'
+  depends_on 'libgudev'
+  depends_on 'mesa'
+  depends_on 'krb5'
+  depends_on 'libxrandr'
```
uberhacker: This might come along for the ride with sommelier.  Not sure but might be worth checking to avoid a potential dreaded dependency loop.

packages/wine.rb
```diff
@@ -14,9 +14,17 @@ class Wine < Package
   })
 
   depends_on 'bison'
-  depends_on 'sommelier'
   depends_on 'flex'
   depends_on 'freetype'
+  depends_on 'ncurses'
```
uberhacker: This is "core" so you can exclude.

packages/wine.rb
```diff
@@ -14,9 +14,17 @@ class Wine < Package
   })
 
   depends_on 'bison'
```
uberhacker: This is "core" so you can exclude.

packages/wine.rb
```diff
@@ -14,9 +14,17 @@ class Wine < Package
...
-  depends_on 'sommelier'
   depends_on 'flex'
```
uberhacker: This is "core" so you can exclude.

packages/wine.rb
JL2210: No. It doesn't.

packages/wine.rb
```diff
@@ -14,9 +14,17 @@ class Wine < Package
...
   depends_on 'freetype'
+  depends_on 'ncurses'
+  depends_on 'cups'
+  depends_on 'gettext'
+  depends_on 'libgudev'
+  depends_on 'mesa'
```
uberhacker: I do know mesa comes along for the ride with sommelier so we don't need it here.

Issue #2569, opened 2018-08-19
QB64: 32- and 64-bit BASIC Compiler and IDE


superloach:
I've been attempting to compile QB64, with little success.

I modified the install script to use `bash setup_build.sh` instead of `./setup_build.sh`, and that fixed some initial errors. But now I keep getting this error, which is obviously related to #2567:
```
/usr/local/libexec/gcc/x86_64-cros-linux-gnu/4.9.4/cc1: error while loading shared libraries: libisl.so.15: cannot open shared object file: No such file or directory
```
Hopefully we can get that fixed, because I'd love to mess around with some of my old programs 😝

(p.s. - there's an SDL version of QB64 as well, in case there are graphical issues in future)

superloach:
After moving the symlink to `libisl.so.19`, it complains about not finding `GL/glu.h`. I'll mention this on the QB64 forum, but should I open an issue on their GitHub?

FellippeHeitor:
Hi and thanks for you interest in QB64. I just replied to your inquiry at our forum. The setup script attempts to detect the distro you're using to download the appropriate dependencies. You may be interested in trying to download them manually first. 
https://www.qb64.org/forum/index.php?topic=438.msg3140#msg3140

FellippeHeitor:
Hope it works, please let us know.

superloach:
It appears that `libglu1-mesa-dev` isn't in the repositories (or anything mentioning "glu", for that matter).

superloach:
I have all the other requirements, though.
Doesn't appear that any other packages have provided it, either.
```
$ sudo find /usr/local/ -name *glu*
/usr/local/share/gtk-doc/html/gobject/glue.png
/usr/local/share/doc/groff-1.22.3/examples/chem/122/ch4z1_eqn_glutamic.chem
/usr/local/include/gssapi/mechglue.h
/usr/local/lib64/wine/fakedlls/glu32.dll
/usr/local/lib64/wine/libglu32.def
/usr/local/lib64/wine/glu32.dll.so
```
The only mention in this repository is a `--without-glu` flag in a PR for Wine.

FellippeHeitor:
FreeGLUT is in the core of QB64. Looks like a deal breaker. We do have some legacy versions available at qb64.org that use SDL instead, as you mentioned above. Maybe one of those will cut it for you. Last version to use it was 0.954

superloach:
[...]

PR #2577, opened 2018-08-20
Add package Qb64

superloach:
Hooray, I fixed my own issue! (#2569) 🤣

### DIFF
--- a/packages/qb64.rb
+++ b/packages/qb64.rb
@@ -13,12 +13,18 @@ class Qb64 < Package
   depends_on 'alsa_plugins'
   depends_on 'sommelier'
 
-  def self.build
+  def self.patch
     system "find . -name '*.sh' -exec sed -i \"s/\\r//g\" {} \\;"
     system "find . -name '*.sh' -exec sed -i \"s/\.\\/setup/bash setup/g\" {} \\;"
     system "find . -name '*.sh' -exec sed -i \"s/\.\\/qb64 \\&/exec qb64 \\&/g\" {} \\;"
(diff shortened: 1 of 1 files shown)


uberhacker:
I tested this on a i686 Chrome OS VM and it looks like a BSOD (aka Black Screen of Death). LoL

screenshot 2018-08-20 at 3 45 50 pm

I'll try to test on armv7l and provide more feedback in a bit.

superloach:
Hmm, that sucks. Perhaps @FellippeHeitor would know something about this?

Also, `x86_64` screenshots just for brags (`WIDTH` is weird and I'm out of practice, don't judge 😛)
image

FellippeHeitor:
I'm glad to see it working! About the black screen, though: no idea.

superloach:
I can't really tell much from that screenshot, but it appears that the Sommelier window doesn't have a "title" bar (even though it doesn't forward titles 😒). Stupid question, but will other GUI apps run properly in that VM?

superloach:
If I had an i686 machine (or an easy way to emulate one), I'd be totally willing to maintain a fork for it. Unfortunately... *I don't.* 😅

uberhacker:
Do you have a Windows or Mac or even Linux system?  You could run a VM like I have it set up.

superloach:
We have our Windows 10 desktop, but it literally has the same specs as my Acer R11, it's sad.
I also have a broken Thinkpad T410, which I used to have Arch Linux on. Currently working on repairing that, in fact.

In short, not really no.

JL2210:
@uberhacker, are you still using that crappy i686 VM from CrosBuilder?

uberhacker:
Yes, unfortunately.  But it is the only VM I could get to work halfway decently.

Issue #2582, opened 2018-08-22
Six site-packages location


superloach:
The pre-built binaries for Six use `/usr/local/lib/python3.6/site-packages/`, which prevents it from working with Python 3.7.
```
$ ls /usr/local/lib/|grep py
libpython2.7.so
libpython2.7.so.1.0
libpython3.7m.so
libpython3.7m.so.1.0
libpython3.so
python2.7
python3.6
python3.7
```
```
$ ls /usr/local/lib/python3.6/site-packages/
__pycache__  six-1.11.0-py3.6.egg-info  six.py
```
```
Python 3.7.0 (default, Aug 11 2018, 12:25:30) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import six
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'six'
```

uberhacker:
Thanks for pointing this out @superloach.  I'll try and make some new binaries tomorrow.

Issue #2580, opened 2018-08-21
Package Request - GIMP


superloach:
~~Unlike my previous package requests, I haven't made much progress on getting this compiled.~~
Ironically enough, I've done most of the work on this so far! 🤣
This will probably be a pretty big project, lots of libraries from various places. I'll go ahead and make a list of dependencies.
- [x] GLib
- [x] GObject
- [x] Cairo
- [x] Pango
- [x] librsvg
- [x] libpng
- [x] libexif
- [x] libtiff
- [x] libjpeg
- [x] liblcms
- [x] pkg-config
- [x] fontconfig
- [x] aalib
- [x] zlib
- [x] python
- [x] babl - #2581
- [x] Python-Cairo - #2583
- [x] json-glib - #2584
- [x] GEGL - #2586 
- [x] libmypaint-gegl - #2587 
- [x] libexiv2 - #2593 
- [x] gexiv2 - #2594 
- [x] mypaint-brushes - #2595 
- [x] PyGObject - #2596 
- [x] GTK+ - #2589 and #2598 **TEMPORARY FIX**
- [ ] webkit - #2590 
- [x] poppler-data - #2597 
- [x] [libwmf](http://wvware.sourceforge.net/libwmf.html)

Also GIMP is *totally* a dependency for #2579 😆

JL2210:
Sure.

skiman6010:
Hey @superloach, I just wanted to let you know that linking the check boxes to the downloads and package websites is super helpful and I love that you did that! It is a gigantic help! For that, I shall do a package for you.

superloach:
Haha, no problem! Very excited to see where we can take this. 😃

superloach:
It looks like [Python-Cairo is in PyPI](https://pypi.org/project/pycairo/), should we just require Python and run a `pip install` before build?
Or would it make more sense to put that into a separate package?

superloach:
Went ahead and submitted a PR to add `pycairo`, following the same format as `six`.

superloach:
Moving on to GEGL, `./configure` exits with this:
```
checking for GLIB - version >= 2.44.0... yes (version 2.56.1)
checking for DEP... no
configure: error: 
        *** GLib not found. You can find it on ftp://ftp.gtk.org
        *** Errors follow:
            No package 'json-glib-1.0' found
```

superloach:
Moving on to GEGL, `./configure` exits with this:
```
[...]

PR #2583, opened 2018-08-22
Add Python-Cairo

superloach:
Working toward #2580.
Based on [six.rb](https://github.com/skycocker/chromebrew/blob/00a26c9501b858e6f831315b5717bf2adafbe912/packages/six.rb), which currently has some issues (#2582). Package format still works just fine, though.


uberhacker:
This looks good!  Nice work.

PR #2587, opened 2018-08-22
Add Libmypaint/Brushlib

superloach:
Working toward #2580.
We're almost done, this is exciting!


packages/libmypaint-gegl.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Libmypaint < Package
+  description 'This is the brush library used by MyPaint. A number of other painting programs use it too.'
```
uberhacker: Better description?  `Libmypaint is MyPaint’s brushstroke rendering code, in a form that can be used by other programs.`

superloach: Good idea, I just used what I found on the GitHub.

packages/libmypaint-gegl.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Libmypaint < Package
```
uberhacker: The class name should match the filename.  Rename `libmypaint-gegl.rb` to `libmypaint.rb`?

superloach: I had originally called it `libmypaint.rb`, but remembered it was actually called `libmypaint-gegl`.
Should I revert to `libmypaint.rb`, or change the class to `Libmypaint_gegl`?

uberhacker: Either way is fine as long as the class matches the filename, otherwise, crew will get confused.

superloach: I'll just omit the `gegl`, keep it shorter.

superloach:
I'm just eager to get this working 😁

PR #2593, opened 2018-08-22
Add libexiv2

superloach:
Working toward #2580


packages/libexiv2.rb
```diff
@@ -0,0 +1,27 @@
...
+  description 'Exiv2 is a Cross-platform C++ library and a command line utility to manage image metadata.'
+  homepage 'http://exiv2.org/'
+  version '0.26'
+  source_url 'http://www.exiv2.org/builds/exiv2-0.26-trunk.tar.gz'
+  source_sha256 'c75e3c4a0811bf700d92c82319373b7a825a2331c12b8b37d41eb58e4f18eafb'
+
+  depends_on 'zlibpkg'
+  depends_on 'gettext'
+  depends_on 'libiconv'
+  depends_on 'expat'
```
uberhacker: You can exclude all these "core" dependencies. :)

superloach: Argh, I keep forgetting about core.

PR #2594, opened 2018-08-22
Add gexiv2

superloach:
Working toward #2580

### DIFF
--- a/packages/gexiv2.rb
+++ b/packages/gexiv2.rb
@@ -4,7 +4,7 @@ class Gexiv2 < Package
   description 'gexiv2 is a GObject wrapper around the Exiv2 photo metadata library.'
   homepage 'https://wiki.gnome.org/Projects/gexiv2/'
   version '0.10.8'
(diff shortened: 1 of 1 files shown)


packages/gexiv2.rb
```diff
@@ -0,0 +1,25 @@
+require 'package'
+
+class Gexiv2 < Package
+  description 'gexiv2 is a GObject wrapper around the Exiv2 photo metadata library.'
+  homepage 'https://wiki.gnome.org/Projects/gexiv2/'
+  version '0.10.8'
+  source_url 'https://mirror.umd.edu/gnome/sources/gexiv2/0.10/gexiv2-0.10.8.tar.xz'
```
uberhacker: Please use the canonical source url instead of a mirror.  Remember, we also have users in China. :)
```
source_url 'https://download.gnome.org/sources/gexiv2/0.10/gexiv2-0.10.8.tar.xz'
```

superloach: Good point, whoops >.<

PR #2595, opened 2018-08-22
Add mypaint-brushes

superloach:
Working toward #2580

### DIFF
--- a/packages/mypaint_brushes.rb
+++ b/packages/mypaint_brushes.rb
@@ -1,0 +1,26 @@
+require 'package'
+
+class Mypaint_brushes < Package
+  description 'Brushes used by MyPaint and other software using libmypaint.'
+  homepage 'https://github.com/Jehan/mypaint-brushes'
+  version '1.3.0'
+  source_url 'https://github.com/Jehan/mypaint-brushes/archive/v1.3.0/mypaint-brushes-v1.3.0.tar.gz'
+  source_sha256 '704bb6420e65085acfd7a61d6050e96b0395c5eab078433f11406c355f16b214'
+
+  depends_on 'libmypaint'
+
+  def self.patch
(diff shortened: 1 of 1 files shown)


packages/mypaint_brushes.rb
```diff
@@ -0,0 +1,26 @@
...
+  description 'Brushes used by MyPaint and other software using libmypaint.'
+  homepage 'https://github.com/Jehan/mypaint-brushes'
+  version '1.3.0'
+  source_url 'https://github.com/Jehan/mypaint-brushes/archive/v1.3.0/mypaint-brushes-v1.3.0.tar.gz'
+  source_sha256 '704bb6420e65085acfd7a61d6050e96b0395c5eab078433f11406c355f16b214'
+
+  depends_on 'libmypaint'
+
+  def self.patch
+    system "wget http://www.linuxfromscratch.org/patches/blfs/svn/mypaint-brushes-1.3.0-automake_1.16-1.patch"
```
uberhacker: Can we have a checksum to make sure this patch is downloaded fully intact?  Something like:
```
abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('mypaint-brushes-1.3.0-automake_1.16-1.patch') ) == '<sha256 hash goes here>'
```

packages/mypaint_brushes.rb
```diff
@@ -0,0 +1,26 @@
...
+    system "patch -Np1 -i mypaint-brushes-1.3.0-automake_1.16-1.patch"
+  end
+
+  def self.build
+    system "./autogen.sh"
+    system "./configure --prefix=#{CREW_PREFIX}"
```
uberhacker: This is trivial but you could replace these 2 lines with just:
```
system "./autogen.sh --prefix=#{CREW_PREFIX}"
```

packages/mypaint_brushes.rb
superloach: good idea

PR #2598, opened 2018-08-23
Temporary fix for Cairomm

superloach:
Disable binaries, because Bintray links are broken.
Just a temporary fix so that #2580 can be installed from a fresh Chromebrew installation.


uberhacker:
Thank you for noticing the broken binaries.  I'll create new binaries soon so this is only temporary. :)

PR #2600, opened 2018-08-23
Update VS Code shasums and add version information

### LINKED ISSUE #2599: Shasum mismatch of VS Code stable sources
## Description

Running `crew install code` on the Pixelbook fails due to a shasum mismatch of the downloaded sources.

```
chronos@localhost / $ crew download code
code: Visual Studio Code is a source code editor developed by Microsoft for Windows, Linux and macOS.
https://code.visualstudio.com/
version 1.25
Checksum mismatch. :/ Try again.
```

The shasum of the source package served by Microsoft is:
```
5ef2f82e1272b0008a446fc8f62f9693e2af7c83d5922883ecbfd8a9986fd40b
```

I assume that the stable VS Code version has changed, and that the package script requires updating.

chromebrew: 1.0.3
ChromeOS: 10718.71.2
Model: Google Pixelbook
Architecture: x86_64

davidschreiber:
Resolves #2599

I updated the shasums to match the files served by Microsoft servers. To make it simpler to detect file version changes, I also added the file names to the package spec.

### DIFF
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -6,11 +6,13 @@ class Code < Package
   version '1.25'
   case ARCH
   when 'x86_64'
+    # code-stable-1534444642.tar.gz
     source_url 'https://go.microsoft.com/fwlink/?LinkID=620884'
-    source_sha256 '5856bbebf38aa05d584da4722869bbe507cf123f69f7ffab5f1532d73dbf3438'
+    source_sha256 '5ef2f82e1272b0008a446fc8f62f9693e2af7c83d5922883ecbfd8a9986fd40b'
   when 'i686'
+    # code-stable-1534444635.tar.gz 
     source_url 'https://go.microsoft.com/fwlink/?LinkID=620885'
-    source_sha256 'af6adc2e2500e50bfebe7ee7b97d661b6e774a590136bf5f89334132a5b292e2'
+    source_sha256 '16a9eb4582767c31ca92fce9d7ed97b960175d8a97db25fab3c8e56afec2f3bc'
   else
     source_url 'https://raw.githubusercontent.com/Microsoft/vscode/1.25.1/README.md'
     source_sha256 'c1c5e6ec903730a4e116c1cfd83bb29acd227195d97f2ec8944452808232c310'


JL2210:
Have you checked this? They delete the files after a while.

JL2210:
Can you please bump the version number to 1.26? We need to bump it so already installed packages will update.

packages/code.rb
```diff
@@ -6,11 +6,13 @@ class Code < Package
   version '1.25'
```
JL2210: Can you please bump the version number to 1.26? We need to update the already installed packages.

skiman6010: Newest stable version is 1.26.1 now it seems.

packages/code.rb
```diff
@@ -6,11 +6,13 @@ class Code < Package
...
   case ARCH
   when 'x86_64'
+    # code-stable-1534444642.tar.gz
     source_url 'https://go.microsoft.com/fwlink/?LinkID=620884'
```
skiman6010: Can we use the GitHub page to get the sources? https://github.com/Microsoft/vscode/releases

That will stop us from downloading from a moving target and hitting checksum errors every time they update.

packages/code.rb
```diff
@@ -6,11 +6,13 @@ class Code < Package
...
-    source_sha256 '5856bbebf38aa05d584da4722869bbe507cf123f69f7ffab5f1532d73dbf3438'
+    source_sha256 '5ef2f82e1272b0008a446fc8f62f9693e2af7c83d5922883ecbfd8a9986fd40b'
   when 'i686'
+    # code-stable-1534444635.tar.gz 
     source_url 'https://go.microsoft.com/fwlink/?LinkID=620885'
```
skiman6010: Can we use the GitHub page to get the sources? https://github.com/Microsoft/vscode/releases

That will stop us from downloading from a moving target and hitting checksum errors every time they update.

uberhacker: @skiman6010: I think the issue here is the source_url is not really source but binary.  But you are right, we should be using a static url that specifies a version rather than a generic download link.  @JL2210, can you please look into this so we can solve the moving target issue?

uberhacker:
Closing and opening issue #2603 instead.

packages/code.rb
JL2210: No. These versions are better.

PR #2606, opened 2018-08-25
Add pre-built binaries for bz2, gegl, gexiv2, libexiv2, libssh, poppler_data and pygobject

uberhacker:
- Modify bz2 to install with the correct permissions
- Add options to the gexiv2 and libexiv2 packages
- Add dependencies to gegl
- Add more gegl dependencies
- Add libssh package
- Add libssh dependency to libexiv2
- Add dependencies to gegl
- Fix poppler_data install
- Make rust a build dependency in librsvg


superloach:
This is awesome dude, these should fix a lot of the problems I was getting stuck on for GIMP! 🎉
It's kinda 2 AM for me, but I'll definitely be getting back to work on the GIMP package ~~tomorrow~~ later today. 👍

PR #2613, opened 2018-08-27
Update julia package

### LINKED ISSUE #2611: [julia] library not found error on x86_64
```
$ julia
ERROR: could not load library "/usr/local/bin/../lib/julia/sys.so"
/usr/local/bin/../lib/julia/sys.so: cannot open shared object file: No such file or directory
```

uberhacker:
Fixes #2611.

### DIFF
--- a/packages/julia.rb
+++ b/packages/julia.rb
@@ -3,7 +3,7 @@ require 'package'
 class Julia < Package
   description 'Julia is a flexible dynamic language, appropriate for scientific and numerical computing'
   homepage 'https://julialang.org/'
-  version '1.0.0'
+  version '1.0.0-1'
   case ARCH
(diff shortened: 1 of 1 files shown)


JL2210:
Try using `#{CREW_DEST_LIB_PREFIX}` instead of `#{CREW_DEST_PREFIX}/lib`

uberhacker:
The problem we are facing is the executable from the binary (which we have no control over) puts the files in `#{CREW_DEST_PREFIX}/lib` instead of `#{CREW_DEST_LIB_PREFIX}`.  This PR creates symlinks to the libraries in `#{CREW_DEST_LIB_PREFIX}` just in case some other app uses these as dependencies.

PR #2616, opened 2018-08-29
Add libgnomecanvas package

uberhacker:
Canvas widget for Gnome2

'libgnomecanvas' is the library for the GNOME canvas, an engine for structured graphics that offers a rich imaging model, high performance rendering, and a powerful, high-level API. It is based on the Tk toolkit's canvas widget, and is one of the essential Gnome libraries.

See https://developer.gnome.org/libgnomecanvas/.

### DIFF
--- a/packages/libgnomecanvas.rb
+++ b/packages/libgnomecanvas.rb
@@ -1,0 +1,37 @@
+require 'package'
+
+class Libgnomecanvas < Package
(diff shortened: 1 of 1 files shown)


JL2210:
Sweet. We're really close to being able to bootstrap GNOME. Only jhbuild to go.

PR #2637, opened 2018-09-08
Add libmng package

uberhacker:
A PNG-like Image Format Supporting Multiple Images, Animation and Transparent JPEG.  See http://www.libpng.org/pub/mng/.

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/libmng.rb
+++ b/packages/libmng.rb
@@ -1,0 +1,36 @@
+require 'package'
+
+class Libmng < Package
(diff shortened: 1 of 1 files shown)


JL2210:
May I ask what this is needed for?

uberhacker:
It's a dependency of something else I'm working on atm.

JL2210:
What would that be? Could I help?

JL2210:
Nice.

PR #2643, opened 2018-09-12
Add more exclusions to needs_binaries.sh

uberhacker:
More packages that don't require binary builds.


JL2210:
I can make a PR that lets dpkg have a binary build.

PR #2647, opened 2018-09-12
Add checkinstall package

JL2210:
CheckInstall is a computer program for Unix-like operating systems which eases the installation and uninstallation of software compiled from source by making use of package management systems.

### DIFF
--- a/packages/checkinstall.rb
+++ b/packages/checkinstall.rb
@@ -13,8 +13,8 @@ class Checkinstall < Package
   })
 
   def self.patch
-    system 'wget', 'https://github.com/JL2210/patches/raw/f2da9fb13545311e67decef27626c45e4c6583e1/checkinstall-1.6.2-chromebrew.patch'
-    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('checkinstall-1.6.2-chromebrew.patch') ) == '17706ab65698b7baa4cb2c1eda1c9ef3a41453069f6d83843f3c134439d3096a'
+    system 'wget', 'https://github.com/JL2210/patches/raw/267ebd80cc8d2c366cd3adb72e26d4cd51677571/checkinstall-1.6.2-chromebrew.patch'
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('checkinstall-1.6.2-chromebrew.patch') ) == '2a70ef396416c3a405b0c5dea54df867e52318efd9868fa1c0cf25c7d7b5e2f3'
     system 'patch -p1 < checkinstall-1.6.2-chromebrew.patch'
   end
 
@@ -23,14 +23,7 @@ class Checkinstall < Package
   end
 
   def self.install
-    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin/"
-    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/lib/checkinstall/"
-    FileUtils.mkdir_p "#{CREW_DEST_LIB_PREFIX}"
-    system "make", "install", "PREFIX=#{CREW_PREFIX}", "LCDIR=#{CREW_PREFIX}/lib/checkinstall", "CONFDIR=#{CREW_PREFIX}/lib/checkinstall/locale", "LIBDIR=#{CREW_LIB_PREFIX}"
-    system "cp -pa #{CREW_PREFIX}/bin/installwatch #{CREW_DEST_PREFIX}/bin/"
-    system "cp -rpa #{CREW_PREFIX}/lib/checkinstall/ #{CREW_DEST_PREFIX}/lib/"
-    system "cp -pa #{CREW_LIB_PREFIX}/installwatch.so #{CREW_DEST_LIB_PREFIX}/"
-    system "cp -pa #{CREW_PREFIX}/sbin/checkinstall #{CREW_DEST_PREFIX}/bin/"
-    system "cp -pa #{CREW_PREFIX}/sbin/makepak #{CREW_DEST_PREFIX}/bin/"
+    system "make", "install", "PREFIX=#{CREW_DEST_PREFIX}", "LCDIR=#{CREW_DEST_PREFIX}/etc/checkinstall", "CONFDIR=#{CREW_DEST_PREFIX}/share/checkinstall/locale", "LIBDIR=#{CREW_DEST_LIB_PREFIX}"
+    system "sed -i 's,#{CREW_DEST_PREFIX},#{CREW_PREFIX},g' #{CREW_DEST_PREFIX}/bin/installwatch"
   end
 end
--- a/packages/checkinstall.rb
+++ b/packages/checkinstall.rb
@@ -13,9 +13,9 @@ class Checkinstall < Package
   })
 
   def self.patch
-    system 'wget', 'https://github.com/JL2210/patches/raw/da54bdfb4aea5dfbab8c08224ebd9416fc2bc62d/checkinstall.patch'
-    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('checkinstall.patch') ) == '71c36071df8deaf3c14caa58d79fd86eb9262872281226c2ee0e48b640c182ce'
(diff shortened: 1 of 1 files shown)


uberhacker:
I have my doubts about the usefulness of this package.  Chromebrew has a special way of packaging that includes file and directory manifests.  I suppose we could create Debian packages since we have dpkg but that still doesn't handle dependencies.  Can you please provide a practical use-case example and then I might consider it?  Also, I believe there is a configuration file you need to include:
```
...
    system "install -Dm644 #{CREW_DEST_PREFIX}/checkinstallrc #{CREW_DEST_HOME}/.checkinstallrc"
    system "cp #{CREW_DEST_PREFIX}/checkinstallrc $HOME/.checkinstallrc"
  end

  def self.postinstall
    puts
    puts "To configure, modify $HOME/.checkinstallrc".lightblue
    puts
  end
```

JL2210:
It can create tar archives. Choose Slackware packaging.

uberhacker:
Dude, I know that, however, we need Chromebrew specific packaging as previous explained.

JL2210:
Like you said, it would be useful with dpkg. I imagine later we could create a patch that builds Chromebrew packaging.

uberhacker:
How about you create a patch now so it can actually be used? :)

JL2210:
Sure!

JL2210:
@uberhacker, this is ready now. Chromebrew compatibility took a couple of hours to attain.

packages/checkinstall.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.patch
+    system 'wget', 'https://github.com/JL2210/patches/raw/f2da9fb13545311e67decef27626c45e4c6583e1/checkinstall-1.6.2-chromebrew.patch'
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('checkinstall-1.6.2-chromebrew.patch') ) == '17706ab65698b7baa4cb2c1eda1c9ef3a41453069f6d83843f3c134439d3096a'
+    system 'patch -p1 < checkinstall-1.6.2-chromebrew.patch'
```
uberhacker: Instead of trying to reinvent the wheel, why not just execute `crew build <pkgname>` when they select Chromebrew?  Also, you should be using `/usr/local/tmp` for the temp directory instead of `/usr/local/var/tmp` or it will crash and burn on CloudReady.

packages/checkinstall.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def self.build
+    system 'make'
+  end
+
+  def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin/"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/lib/checkinstall/"
```
uberhacker: `#{CREW_DEST_PREFIX}/etc/checkinstall/` is better location IMHO.

packages/checkinstall.rb
```diff
@@ -0,0 +1,36 @@
...
+    FileUtils.mkdir_p "#{CREW_DEST_LIB_PREFIX}"
+    system "make", "install", "PREFIX=#{CREW_PREFIX}", "LCDIR=#{CREW_PREFIX}/lib/checkinstall", "CONFDIR=#{CREW_PREFIX}/lib/checkinstall/locale", "LIBDIR=#{CREW_LIB_PREFIX}"
```
uberhacker: `"LCDIR=#{CREW_PREFIX}/etc/checkinstall"` and `"CONFDIR=#{CREW_PREFIX}/etc/checkinstall/locale"`?

packages/checkinstall.rb
```diff
@@ -0,0 +1,36 @@
...
+    system "cp -pa #{CREW_PREFIX}/bin/installwatch #{CREW_DEST_PREFIX}/bin/"
+    system "cp -rpa #{CREW_PREFIX}/lib/checkinstall/ #{CREW_DEST_PREFIX}/lib/"
```
uberhacker: `system "cp -rpa #{CREW_PREFIX}/etc/checkinstall/ #{CREW_DEST_PREFIX}/etc/"`?

JL2210: No. Not configuration files.

packages/checkinstall.rb
JL2210: Not config files.

packages/checkinstall.rb
JL2210: How badly non-FHS compliant is Cloudready!?

JL2210: Also, that defeats the total purpose of Checkinstall.

packages/checkinstall.rb
JL2210: What does IMHO mean?

JL2210: Plus the patch doesn't allow for that.

uberhacker: IMHO = in my humble opinion

packages/checkinstall.rb
uberhacker: I don't think you understand how Chromebrew packages are built.  Take some time and examine how the build function works in the crew command.  `/usr/local/var` in CloudReady is a symlink to the read-only `/var` directory.

JL2210: `/var/tmp` is supposed to be mounted as tmpfs. When's the last time you used Cloudready? Crew build only accepts the name of a package, not a directory. I could try `crew build .` and all I would get back would be an error: `.: package not found :/`

packages/checkinstall.rb
JL2210: `/usr/local/lib/crew`?

packages/checkinstall.rb
JL2210: The patch breaks this. I could be open to a compromise on `/usr/local/share`.

packages/checkinstall.rb
JL2210:
```
Loading extra module: /usr/share/crosh/dev.d/50-crosh.sh
shWelcome to crosh, the Chrome OS developer shell.

If you got here by mistake, don't panic!  Just close this tab and carry on.

Type 'help' for a list of commands.

If you want to customize the look/behavior, you can use the options page.
Load it by using the Ctrl+Shift+P keyboard shortcut.

crosh> shell
crewchronos@localhost / $ crew build .
Package . not found. :(
chronos@localhost / $ crew build /home/chronos/user/Downloads
Package /home/chronos/user/Downloads not found. :(
chronos@localhost / $ cd
chronos@localhost ~ $ cd Downloads
chronos@localhost ~/Downloads $ ls
Markdown.txt  Takeout  ada.rb  ada_bootstrap.rb  chroot  misc  patches  takeout-20180912T133929Z-001.tgz  update-cross-script-1
chronos@localhost ~/Downloads $ cd misc/
.clang-format               BLFS-BOOK-8.2-systemd.html  crew-build/                 uCoreutils/                 
.gitignore                  LFS-BOOK-8.2-systemd.pdf    hello-chromeos-gles2/       xfce_fat-4.12.tar.lz4       
chronos@localhost ~/Downloads $ cd misc/
chronos@localhost ~/Downloads/misc $ crew build uCoreutils                  
Package uCoreutils not found. :(
chronos@localhost ~/Downloads/misc $ 
```

packages/checkinstall.rb
```diff
@@ -0,0 +1,36 @@
...
+    system "cp -pa #{CREW_LIB_PREFIX}/installwatch.so #{CREW_DEST_LIB_PREFIX}/"
+    system "cp -pa #{CREW_PREFIX}/sbin/checkinstall #{CREW_DEST_PREFIX}/bin/"
+    system "cp -pa #{CREW_PREFIX}/sbin/makepak #{CREW_DEST_PREFIX}/bin/"
```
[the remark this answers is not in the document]
JL2210: K.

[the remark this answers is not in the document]
JL2210: I guess I can't tell you that `make install` performs a few `sed` operations on `installwatch` based on the prefix.

packages/checkinstall.rb
JL2210: I'll change the patch. I see what you mean.

PR #2649, opened 2018-09-12
Add hplip package

JL2210:
Print, Scan and Fax Drivers for Linux

### DIFF
--- a/packages/hplip.rb
+++ b/packages/hplip.rb
@@ -1,0 +1,32 @@
+require 'package'
+
+class Hplip < Package
+  description 'Print, Scan and Fax Drivers for Linux'
+  homepage 'https://developers.hp.com/hp-linux-imaging-and-printing/'
+  version '3.18.7'
+  source_url 'https://sourceforge.net/projects/hplip/files/hplip/3.18.7/hplip-3.18.7.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/hplip.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  depends_on 'libjpeg'
+  depends_on 'cups'
+
+  def self.build
+    system "sed -i 's,/usr/bin/perl,#{CREW_PREFIX}/bin/perl,g' hpijs-drv"
+    system "sed -i 's,/usr/include,#{CREW_PREFIX}/include,g' Makefile.in Makefile.am config.guess installer/core_install.py installer/dcheck.py"
+    system 'sed -i "s,return(true);,return TRUE;,g" prnt/hpcups/genPCLm.cpp'
+    system 'sed -i "s,return true;,return TRUE;,g" prnt/hpcups/genPCLm.cpp'
+    system "sed -i 's,/usr/bin/file,#{CREW_PREFIX}/bin/file,g' configure"
```
uberhacker: FYI: You can also use `system 'filefix'` to do the same thing.  It will even search the entire codebase for every possible configure file to change including subdirectories.

PR #2661, opened 2018-09-14
Add kakoune package

StanislavNikolov:
Kakoune is a text editor, similar to vim.
Official repo: https://github.com/mawww/kakoune
Official website: http://kakoune.org

Works properly on all targets.

I have not researched much about how crew does prebuilt binaries. It looks like I have to upload them somewhere (bintray?) or set up a script to build them automatically. Is this correct?

### DIFF
--- a/packages/kakoune.rb
+++ b/packages/kakoune.rb
@@ -7,20 +7,20 @@ class Kakoune < Package
   version 'v2018.09.04'
   source_url 'https://github.com/mawww/kakoune/releases/download/v2018.09.04/kakoune-2018.09.04.tar.bz2'
   source_sha256 '7a31c9f08c261c5128d1753762721dd7b7fe4bb4e9a3c368c9d768c72a1472e1'
-
+  
   depends_on 'ncurses'  => :build
   depends_on 'asciidoc' => :build
   depends_on 'libxslt'  => :build
 
   def self.build
     Dir.chdir("src") do
-      system "make", "debug=no"
+      system "make", "PREFIX=#{CREW_PREFIX}", "DESTDIR=#{CREW_DEST_DIR}", "debug=no"
     end
   end
 
   def self.install
     Dir.chdir("src") do
-      system "make", "install", "DESTDIR=#{CREW_DEST_DIR}", "debug=no"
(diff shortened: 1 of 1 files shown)


packages/kakoune.rb
```diff
@@ -0,0 +1,26 @@
...
+  source_url 'https://github.com/mawww/kakoune/releases/download/v2018.09.04/kakoune-2018.09.04.tar.bz2'
+  source_sha256 '7a31c9f08c261c5128d1753762721dd7b7fe4bb4e9a3c368c9d768c72a1472e1'
+
+  depends_on 'ncurses'  => :build
+  depends_on 'asciidoc' => :build
+  depends_on 'libxslt'  => :build
+
+  def self.build
+    Dir.chdir("src") do
+      system "make", "debug=no"
```
uberhacker: You generally want to add configure options `--prefix=#{CREW_PREFIX}` and `--libdir=#{CREW_LIB_PREFIX}`.  Does this archive include a configure script?

uberhacker:
Thank you for the new package @StanislavNikolov!  Please see the inline comment.

uberhacker:
You need an account on bintray.com to upload pre-built binaries.  Don't worry about it for now.  We can get those added for you.

packages/kakoune.rb
uberhacker: I examined the codebase and it looks like you could do something like this:
```
system "make PREFIX=#{CREW_PREFIX} debug=no"
```

packages/kakoune.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+  end
+
+  def self.install
+    Dir.chdir("src") do
+      system "make", "install", "DESTDIR=#{CREW_DEST_DIR}", "debug=no"
```
uberhacker:
```
system "make install PREFIX=#{CREW_PREFIX} DESTDIR=#{CREW_DEST_DIR} debug=no"
```

packages/kakoune.rb
StanislavNikolov: My bad, kakoune's Makefile has `PREFIX` already set to `/usr/local`, so I didn't notice.

>Does this archive include a configure script?

As far as I know - no.

PR #2682, opened 2018-09-17
Updated tilp package and split dependencies.

### LINKED ISSUE #2489: TiLP in Chromebrew? (Attempted solutions)

JL2210:
Fixes #2489 

Updates Tilp package, uses Autoconf, and splits dependencies into separate packages.

### DIFF
--- a/packages/libticalcs2.rb
+++ b/packages/libticalcs2.rb
@@ -1,0 +1,32 @@
+require 'package'
+
+class Libticalcs2 < Package
+  description 'Libticalcs2 offers the library used to properly manipulate files on TI calculators.'
+  homepage 'http://lpg.ticalc.org/prj_tilp/'
+  version '1.1.9'
+  source_url 'https://sourceforge.net/projects/tilp/files/tilp2-linux/tilp2-1.18/libticalcs2-1.1.9.tar.bz2'
+  source_sha256 '76780788bc309b647f97513d38dd5f01611c335a72855e0bd10c7bdbf2e38921'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'libticables2'
+
+  def self.build
+    system 'autoreconf -i'
+    system "/usr/bin/env",
+      "CC=clang -fuse-ld=lld",
+      "CXX=clang++ -fuse-ld=lld",
+      "./configure",
+      "--prefix=#{CREW_PREFIX}",
+      "--libdir=#{CREW_LIB_PREFIX}"
+    system 'sed -i "s,tests,,g" Makefile'
+    system 'make'
+  end
+
+  def self.install
+    system "make install DESTDIR=#{CREW_DEST_DIR}"
+  end
+end
--- a/packages/tilp2.rb
+++ b/packages/tilp2.rb
@@ -1,0 +1,32 @@
+require 'package'
+
+class Tilp2 < Package
+  description 'TiLP 2 is a linking program for Texas Instruments\' graphing calculators.'
+  homepage 'http://lpg.ticalc.org/prj_tilp/'
+  version '1.18'
+  source_url 'https://sourceforge.net/projects/tilp/files/tilp2-linux/tilp2-1.18/tilp2-1.18.tar.bz2'
+  source_sha256 '7b3ab363eeb52504d6ef5811c5d264f8016060bb7bd427be5a064c2ed7384e47'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'libticalcs2'
+  depends_on 'libglade'
+  depends_on 'sommelier'
+
+  def self.build
+    system "wget https://github.com/JL2210/patches/raw/04eeee244059b41e0f0f4bf60f38fc4b6cb1e480/tilp2-1.18-autoconf.patch"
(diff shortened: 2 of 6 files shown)


JL2210:
I need to do something. Don't merge yet.

packages/libticalcs2.rb
```diff
@@ -0,0 +1,32 @@
+require 'package'
+
+class Libticalcs2 < Package
+  description 'Libticalcs2 offers the library used to properly manipulate files on TI calculators.'
```
uberhacker: Isn't this the same description of libtifiles2?

packages/tilp2.rb
```diff
@@ -0,0 +1,38 @@
...
+  })
+
+  depends_on 'libticalcs2'
+  depends_on 'libglade'
+  depends_on 'sommelier'
+
+  def self.build
+    system "wget https://github.com/JL2210/patches/raw/04eeee244059b41e0f0f4bf60f38fc4b6cb1e480/tilp2-1.18-autoconf.patch"
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('tilp2-1.18-autoconf.patch') ) == 'dedba5bbd1ceb9cf819ea62c156377b6e4eb4b167f33e2c4c8823c2684fa352b'
+    system 'patch -Np1 -i tilp2-1.18-autoconf.patch'
```
uberhacker: Can we remove these 3 lines from the `def self.build` section and add to `def self.patch`?

packages/tilp2.rb
```diff
@@ -0,0 +1,38 @@
...
+    system 'autoreconf -i'
+    system "/usr/bin/env",
+           "CC=clang -fuse-ld=lld",
+           "CXX=clang++ -fuse-ld=lld",
+           "./configure",
+           "--prefix=#{CREW_PREFIX}",
+           "--without-kde",
+           "--libdir=#{CREW_LIB_PREFIX}"
+    system 'make'
+    system 'make install'
```
uberhacker: This line should be removed.

packages/tilp2.rb
```diff
@@ -0,0 +1,38 @@
...
+  depends_on 'libticalcs2'
+  depends_on 'libglade'
+  depends_on 'sommelier'
+
+  def self.build
+    system "wget https://github.com/JL2210/patches/raw/04eeee244059b41e0f0f4bf60f38fc4b6cb1e480/tilp2-1.18-autoconf.patch"
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('tilp2-1.18-autoconf.patch') ) == 'dedba5bbd1ceb9cf819ea62c156377b6e4eb4b167f33e2c4c8823c2684fa352b'
+    system 'patch -Np1 -i tilp2-1.18-autoconf.patch'
+    system 'autoreconf -i'
+    system "/usr/bin/env",
```
uberhacker: Why do you need `/usr/bin/env`?

packages/libticalcs2.rb
JL2210: Yeah. Let me change that.

packages/tilp2.rb
JL2210: Yeah. Forgot to catch that.

packages/tilp2.rb
JL2210: Oh. That was from my previous strategy.

packages/tilp2.rb
JL2210: Because it doesn't work with multiple lines when you define variables first. Trust me, I tried to avoid this.

JL2210:
How's this now?

PR #2683, opened 2018-09-17
Revert "Add checkinstall package"

JL2210:
Reverts skycocker/chromebrew#2647

Wasn't ready yet. Need to fix a few things (Locales were backwards with config files)


uberhacker:
Too late.  Please don't submit a PR until it's ready.  I merged this and you own it now. :)  Just add changes to the existing package in master.

PR #2686, opened 2018-09-18
Update cross script

JL2210:
Updates cross compiler script.
Tested on x86_64 cross compiling for i686

### DIFF
--- a/tools/build_cros_gcc
+++ b/tools/build_cros_gcc
@@ -95,7 +95,7 @@
    if ! [[ "$1" =~ "^[0-9]+$" ]] &>/dev/null; then
      JOBS="$1"
      shift
-   else if { mkdir -p "$1" &>/dev/null; }; then
+   else
      PREFIX="$1"
      shift
    fi
--- a/tools/build_cros_gcc
+++ b/tools/build_cros_gcc
@@ -1,21 +1,24 @@
 #!/bin/bash
 
+ # '\033[1;31' outputs red text. '\033[0m' reverts it to normal.
+ command -v crew > /dev/null || { echo -e "\033[1;31mcrew command not found.\033[0m"; exit 1; }
+
  GLIBC=glibc-$(crew search glibc -v | tail -1 | cut -d' ' -f2 | cut -d'-' -f1)
- BINUTILS=binutils-2.31.1
+ BINUTILS=binutils-$(crew search binutils -v | tail -1 | cut -d' ' -f2 | cut -d'-' -f1)
  GCC=gcc-8.2.0
- MPFR=mpfr-4.0.1
- GMP=gmp-6.1.2
- MPC=mpc-1.1.0
- ISL=isl-0.20
- CLOOG=cloog-0.18.4
+ MPFR=mpfr-$(crew search mpfr -v | tail -1 | cut -d' ' -f2 | cut -d'-' -f1)
+ GMP=gmp-$(crew search gmp -v | tail -1 | cut -d' ' -f2 | cut -d'-' -f1)
+ MPC=mpc-$(crew search mpc -v | tail -1 | cut -d' ' -f2 | cut -d'-' -f1)
+ ISL=isl-$(crew search isl -v | tail -1 | cut -d' ' -f2 | cut -d'-' -f1)
+ CLOOG=cloog-$(crew search cloog -v | tail -1 | cut -d' ' -f2 | cut -d'-' -f1)
 
  KERNEL_VERSION=$(crew search linuxheaders -v | tail -1 | cut -d' ' -f2 | cut -d'-' -f1).1
 
  MAJOR_VERSION="$(echo $KERNEL_VERSION | cut -c1)"
 
- case $1 in
- -h|--help) echo USAGE:
- echo \ $0 target arch_type prefix
+ case x$1x in
+ x-hx|x--helpx) echo USAGE:
+ echo \ $0 target [arch_type] [prefix]
  echo \ \-h, --help: Display this help
  echo \ \-v, --version: Display version
  echo \ \-t, --const: Display constants
@@ -23,10 +26,10 @@
  echo \ Valid targets: 'aarch64-cros-linux-gnu', 'x86_64-cros-linux-gnu', 'i686-cros-linux-gnu', 'armv7l-cros-linux-gnueabihf'
  shift
  exit 0;;
- -v|--version) echo build_cros_gcc version 1.26
+ x-vx|x--versionx) echo build_cros_gcc version 1.26.1
  shift
  exit 0;;
- -c|--clean) rm -rf build-binutils build-gcc build-glibc
+ x-cx|x--cleanx) rm -rf build-binutils build-gcc build-glibc
  rm -rf \
  ./$GCC \
  ./$CLOOG \
@@ -48,7 +51,7 @@
  ./$MPC.tar.gz || true
  shift
  exit 0;;
- -t|--const)
+ x-tx|x--constx)
  echo KERNEL_VERSION=$KERNEL_VERSION
  echo MAJOR_VERSION=$MAJOR_VERSION
  echo CLOOG=$CLOOG
@@ -61,22 +64,22 @@
  echo GLIBC=$GLIBC
  shift
  exit 0;;
- armv7l-cros-linux-gnueabihf)
+ xarmv7l-cros-linux-gnueabihfx)
  true;;
- aarch64-cros-linux-gnu)
+ xaarch64-cros-linux-gnux)
  true;;
- i686-cros-linux-gnu)
+ xi686-cros-linux-gnux)
  true;;
- x86_64-cros-linux-gnu)
+ xx86_64-cros-linux-gnux)
  true;;
- *)
+ xx)
+ $0 -h
+ exit 1;;
+ x*x)
  echo Invalid target or option \'$1\'
  $0 -h
  exit 1;;
  esac
-
- # '\033[1;31' outputs red text. '\033[0m' reverts it to normal.
- command -v crew > /dev/null || { echo "\033[1;31mcrew command not found.\033[0m"; exit 1; }
 
  if test ! $2; then
    echo "$1" | grep x86_64 1>/dev/null
@@ -105,19 +108,28 @@
    exit 0
  fi
 
- $0 -c
+ rm -rf \
+ ./$GLIBC \
+ ./$GMP \
+ ./$GCC \
+ ./$BINUTILS \
+ ./linux-$KERNEL_VERSION \
+ ./$MPFR \
+ ./$CLOOG \
+ ./$ISL \
+ ./$MPC
 
- echo $1 $2 $3
+ echo $0 $1 $2 $3
 
- wget https://ftpmirror.gnu.org/gnu/binutils/$BINUTILS.tar.xz
- wget https://ftpmirror.gnu.org/gnu/gcc/$GCC/$GCC.tar.xz
- wget https://www.kernel.org/pub/linux/kernel/v"$MAJOR_VERSION".x/linux-$KERNEL_VERSION.tar.xz
- wget https://ftpmirror.gnu.org/gnu/libc/$GLIBC.tar.xz
- wget https://ftpmirror.gnu.org/gnu/mpfr/$MPFR.tar.xz
- wget https://ftpmirror.gnu.org/gnu/gmp/$GMP.tar.xz
- wget https://ftpmirror.gnu.org/gnu/mpc/$MPC.tar.gz
- wget http://isl.gforge.inria.fr/$ISL.tar.bz2
- wget https://www.bastoul.net/cloog/pages/download/$CLOOG.tar.gz
+ wget -c https://ftpmirror.gnu.org/gnu/binutils/$BINUTILS.tar.xz
+ wget -c https://ftpmirror.gnu.org/gnu/gcc/$GCC/$GCC.tar.xz
+ wget -c https://www.kernel.org/pub/linux/kernel/v"$MAJOR_VERSION".x/linux-$KERNEL_VERSION.tar.xz
+ wget -c https://ftpmirror.gnu.org/gnu/libc/$GLIBC.tar.xz
+ wget -c https://ftpmirror.gnu.org/gnu/mpfr/$MPFR.tar.xz
+ wget -c https://ftpmirror.gnu.org/gnu/gmp/$GMP.tar.xz
+ wget -c https://ftpmirror.gnu.org/gnu/mpc/$MPC.tar.gz
+ wget -c http://isl.gforge.inria.fr/$ISL.tar.bz2
+ wget -c https://www.bastoul.net/cloog/pages/download/$CLOOG.tar.gz
 
  for tar in *.tar*; do tar xpf $tar; done
 
@@ -133,6 +145,10 @@
  ./$MPC.tar.gz 
 
  cd $GCC
+ for patch in {isl-0.20,libmpx}; do
+   wget https://github.com/JL2210/patches/raw/master/gcc-8.2.0-$patch.patch
+   patch -p1 < gcc-8.2.0-$patch.patch
+ done
  ln -s ../$CLOOG cloog
  ln -s ../$ISL isl
  ln -s ../$MPC mpc
@@ -142,7 +158,7 @@
  mkdir build-binutils || true
  cd build-binutils
  ../$BINUTILS/configure --prefix=$3 --target=$1 --disable-multilib
- make -j$(nproc)
+ make -j$(( `nproc` + 1 ))
  make install
  cd ..
  cd linux-$KERNEL_VERSION
@@ -150,29 +166,29 @@
  cd ..
  mkdir build-gcc || true
  cd build-gcc
- ../$GCC/configure --disable-libmpx --prefix=$3 --target=$1 --enable-shared --enable-host-shared --enable-languages=c,c++,fortran,go --disable-multilib
- make -j$(nproc) all-gcc
+ ../$GCC/configure --prefix=$3 --target=$1 --enable-shared --enable-host-shared --enable-languages=c,c++,fortran,go --disable-multilib
+ make -j$(( `nproc` + 1 )) all-gcc
  make install-gcc
  cd ..
  mkdir build-glibc || true
  cd build-glibc
- ../$GLIBC/configure --disable-libmpx --prefix=$3/$1 --enable-shared --enable-host-shared --build=$MACHTYPE --host=$1 --target=$1 --with-headers=$3/$1/include --disable-multilib libc_cv_forced_unwind=yes
+ ../$GLIBC/configure --prefix=$3/$1 --enable-shared --enable-host-shared --build=$MACHTYPE --host=$1 --target=$1 --with-headers=$3/$1/include --disable-multilib libc_cv_forced_unwind=yes
  make install-bootstrap-headers=yes install-headers
- make -j$(nproc) csu/subdir_lib
+ make -j$(( `nproc` + 1 )) csu/subdir_lib
  install csu/crt1.o csu/crti.o csu/crtn.o $3/$1/lib
  $1-gcc -nostdlib -nostartfiles -shared -x c /dev/null -o $3/$1/lib/libc.so
  touch $3/$1/include/gnu/stubs.h
  cd ..
  cd build-gcc
- make -j$(nproc) all-target-libgcc
+ make -j$(( `nproc` + 1 )) all-target-libgcc
  make install-target-libgcc
  cd ..
  cd build-glibc
- make -j$(nproc)
+ make -j$(( `nproc` + 1 ))
  make install
  cd ..
  cd build-gcc
- make -j$(nproc)
+ make -j$(( `nproc` + 1 ))
  make install
  cd ..
  $1-gcc -v
--- a/tools/build_cros_gcc
+++ b/tools/build_cros_gcc
@@ -18,7 +18,7 @@
 
  case x$1x in
  x-hx|x--helpx) echo Usage:
- echo \ $0 target [arch_type] [prefix]
+ echo \ $0 target [arch_type] [prefix] [jobs]
  echo \ \-h, --help: Display this help
  echo \ \-v, --version: Display version
  echo \ \-t, --const: Display constants
@@ -107,6 +107,8 @@
    $0 $1 $2 $(crew const CREW_PREFIX | cut -d'=' -f2)
    exit 0
  fi
+ if test ! $4; then
+   $0 $1 $2 $3 $(( `nproc` + 1 ))
 
  rm -rf \
  ./$GLIBC \
@@ -144,9 +146,9 @@
 
  cd $GCC
  GCC_PATCHES="$GCC-$ISL.patch"
- for patch in $GCC_PATCHES; do
-   wget https://github.com/JL2210/patches/raw/master/$patch.patch
-   patch -p1 < $patch.patch
+ for PATCH in $GCC_PATCHES; do
+   wget https://github.com/JL2210/patches/raw/master/$PATCH
+   patch -p1 < $PATCH
  done
  ln -s ../$CLOOG cloog
  ln -s ../$ISL isl
@@ -157,7 +159,7 @@
  mkdir build-binutils || true
  cd build-binutils
  ../$BINUTILS/configure --prefix=$3 --target=$1 --disable-multilib
- make -j$(( `nproc` + 1 ))
+ make -j$4
  make install
  cd ..
  cd linux-$KERNEL_VERSION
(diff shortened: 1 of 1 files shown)


tools/build_cros_gcc
```diff
@@ -61,23 +64,23 @@
...
  true;;
- x86_64-cros-linux-gnu)
+ xx86_64-cros-linux-gnux)
  true;;
- *)
+ xx)
+ $0 -h
+ exit 1;;
+ x*x)
  echo Invalid target or option \'$1\'
```
uberhacker: Why not red text here also? ;)

tools/build_cros_gcc
```diff
@@ -142,37 +158,37 @@
  mkdir build-binutils || true
  cd build-binutils
  ../$BINUTILS/configure --prefix=$3 --target=$1 --disable-multilib
- make -j$(nproc)
+ make -j$(( `nproc` + 1 ))
```
uberhacker: Why add 1?

tools/build_cros_gcc
```diff
@@ -105,19 +108,28 @@
...
- wget https://www.bastoul.net/cloog/pages/download/$CLOOG.tar.gz
+ wget -c https://ftpmirror.gnu.org/gnu/binutils/$BINUTILS.tar.xz
+ wget -c https://ftpmirror.gnu.org/gnu/gcc/$GCC/$GCC.tar.xz
+ wget -c https://www.kernel.org/pub/linux/kernel/v"$MAJOR_VERSION".x/linux-$KERNEL_VERSION.tar.xz
+ wget -c https://ftpmirror.gnu.org/gnu/libc/$GLIBC.tar.xz
+ wget -c https://ftpmirror.gnu.org/gnu/mpfr/$MPFR.tar.xz
+ wget -c https://ftpmirror.gnu.org/gnu/gmp/$GMP.tar.xz
+ wget -c https://ftpmirror.gnu.org/gnu/mpc/$MPC.tar.gz
+ wget -c http://isl.gforge.inria.fr/$ISL.tar.bz2
+ wget -c https://www.bastoul.net/cloog/pages/download/$CLOOG.tar.gz
```
uberhacker: Why do you need `-c`?

tools/build_cros_gcc
```diff
@@ -105,19 +108,28 @@
...
+ ./$GCC \
+ ./$BINUTILS \
+ ./linux-$KERNEL_VERSION \
+ ./$MPFR \
+ ./$CLOOG \
+ ./$ISL \
+ ./$MPC
 
- echo $1 $2 $3
+ echo $0 $1 $2 $3
```
uberhacker: This assumes all arguments are provided.  Aren't `arch_type` and `prefix` optional?

tools/build_cros_gcc
```diff
@@ -133,6 +145,10 @@
  ./$MPC.tar.gz 
 
  cd $GCC
+ for patch in isl-0.20; do
+   wget https://github.com/JL2210/patches/raw/master/gcc-8.2.0-$patch.patch
+   patch -p1 < gcc-8.2.0-$patch.patch
+ done
```
uberhacker: Simpler?
```
wget https://github.com/JL2210/patches/raw/master/gcc-8.2.0-$ISL.patch
patch -p1 < gcc-8.2.0-$ISL.patch
```

tools/build_cros_gcc
JL2210: Because it's not an error that would stop them from moving on totally.

tools/build_cros_gcc
JL2210: It makes the process go faster. Also recommended by Gentoo.

tools/build_cros_gcc
JL2210: If the download fails during the script running, it can complete the download. Otherwise, it overwrites it.

tools/build_cros_gcc
JL2210: Yes, they are. The script executes itself to fill in the arguments.

tools/build_cros_gcc
JL2210: Not really. If I ever create more GCC patches, I could use those.

uberhacker: Then do it the _"right"_ way:
```
PATCHES="$GCC-$ISL.patch"
for PATCH in $PATCHES; do
  wget https://github.com/JL2210/patches/raw/master/$PATCH
  patch -p1 < $PATCH
done
```

tools/build_cros_gcc
uberhacker: Shouldn't this value be up to the user to decide?

tools/build_cros_gcc
JL2210: This is also jut for debug. I can remove it.

tools/build_cros_gcc
uberhacker: But the red text would make it stand out.  BTW, red doesn't necessarily mean a fatal error.

tools/build_cros_gcc
uberhacker: What does that mean?  That doesn't make any sense.

tools/build_cros_gcc
JL2210: OK.

tools/build_cros_gcc
```diff
@@ -133,6 +143,11 @@
  ./$MPC.tar.gz 
 
  cd $GCC
+ GCC_PATCHES="$GCC-$ISL.patch"
+ for patch in $GCC_PATCHES; do
+   wget https://github.com/JL2210/patches/raw/master/$patch.patch
+   patch -p1 < $patch.patch
```
uberhacker: I think you will want `$patch` instead of `$patch.patch`.  Please test. :)

JL2210: Oops. I don't have that kind of Internet connection.

tools/build_cros_gcc
JL2210: Have you ever canceled a download halfway through? With the `-c` flag, you can continue downloading where you left off.

tools/build_cros_gcc
JL2210: OK. It's red now.

tools/build_cros_gcc
JL2210: Not necessarily. I imagine most users would complain if this took days to compile. With this, it takes much less time.

tools/build_cros_gcc
uberhacker: If the downloads abort, does the script exit gracefully?  This is usually why we need sha256 checksums on files.  Why aren't they in this script? :)

JL2210:
Let me do one more thing.

tools/build_cros_gcc
uberhacker: I know some builds require nproc = 1 or they fail.  I still think it should be up to the user to decide.

tools/build_cros_gcc
```diff
@@ -104,20 +107,29 @@
    $0 $1 $2 $(crew const CREW_PREFIX | cut -d'=' -f2)
    exit 0
  fi
+ if test ! $4; then
```
uberhacker: This should probably be `if test $3; then`.  `if test ! $4; then` will always be null even if $3 is not provided.

JL2210: Yeah. That's what I'm working on. Stay tuned.

tools/build_cros_gcc
```diff
@@ -142,37 +159,37 @@
  mkdir build-binutils || true
  cd build-binutils
  ../$BINUTILS/configure --prefix=$3 --target=$1 --disable-multilib
- make -j$(nproc)
+ make -j$4
```
uberhacker: I don't think you can assume the `$4` argument has been provided here.

uberhacker:
I think you need to be careful with the nproc value.  You can't always assume there will be more than 1 processing unit available, especially on notoriously underpowered Chromebooks.
```
Print the number of processing units available to the current process,
which may be less than the number of online processors
```
This means if you have a single processor Chromebook (yes, these exist), you cannot assign more than 1 processing unit.  I think it might be best if we just leave this alone.  The default nproc is 4 with a quad processor.  Would you rather wait twice as long to compile on these Chromebooks by default? ;)

JL2210:
No. The `nproc` command specifies the number of processors on the CURRENT MACHINE. There isn't any drawbacks of using two jobs if you don't have to processors/cores. The Linux kernel limits the processor percentage used by these applications. Who in their right mind would try to compile a cross compiler on a one-core/processor Chromebook, anyway?

JL2210:
In studies done by benchmarking websites, using one more job than processors/cores actually increases the speed of the compiling, even on one-core processors.

uberhacker:
Yes, I think you are confused about the purpose of nproc.  It doesn't mean the number of jobs per processor, it means the number of processors to dedicate to a particular task.  It is physically impossible to assign more than the number of processors available.  That still misses the point, however.  What you want, if possible, is the maximum number of processors dedicated to any given task.  With that said, I think a default of `nproc --all` is better than `nproc + 1`.  BTW, where are these benchmark studies you are talking about?  Don't just spew information without backing it up with factual evidence.

JL2210:
Yes, it is the number of processors. You are wrong about assigning more than the number of processors available though. Try running `top` and counting how many processes are in the list. That's how many processes are running at the moment. There's a lot more processes than there are cores, and last I checked, `gcc` is a process.

uberhacker:
Don't confuse processes with processors.

JL2210:
All right. I give up. I can't argue with you.

JL2210:
I'm not. You can't `allocate` processors or processes. Using one more core than my computer has worked for two weeks to build an entire system from scratch. I used it just today to build `guile`, and it worked just fine. I've used it many a time on an i686 VirtualBox image with only one 'virtual' core. My point is that it could provide some benefits, and if it doesn't, it still doesn't harm anything.

JL2210:
I was looking at a page about building a cross compiler before I wrote this script, and it showed that building with 1.5x your core level was the fastest stable time, about 15s faster than -j'nproc'. The reason `stable` was referenced was because he experienced segmentaion faults when running with -j'nproc * 2'.
I don't remember the exact website, and so I deleted my comment.

JL2210:
It wasn't an organization or a benchmarking website.

tools/build_cros_gcc
JL2210: It continues. Usually, compressed files don't decompress if they fail. All of these also use HTTPS, which automatically checks the files.

JL2210: Also, with the `-c` option, the downloads don't fail. They just remove the corrupted bytes and go on.

uberhacker: HTTPS is just an encrypted transfer protocol.  It has nothing to do with checking the file integrity.  I still think it would be a good idea to include the sha256 checksum hash check.  It would be easy, right?  All the checksums are included in each package already.

tools/build_cros_gcc
JL2210: Not any of these. That would be a major flaw, especially with GCC and Glibc.

tools/build_cros_gcc
JL2210: Except for GCC 8.2.0

JL2210: How do you do that in bash, anyway?

uberhacker: Would this work?
```
SUMS="5d20086ecf5752cc7d9134246e9588fa201740d540f7eb84d795b1f7a93bca86  $BINUTILS.tar.xz"
SUMS+="325adf3710ce2229b7eeb9e84d3b539556d093ae860027185e7af8a8b00a750e  $CLOOG.tar.gz"
SUMS+="196c3c04ba2613f893283977e6011b2345d1cd1af9abeac58e916b1aab3e0080  $GCC.tar.xz"
SUMS+="94efeb00e4603c8546209cefb3e1a50a5315c86fa9b078b6fad758e187ce13e9  $GLIBC.tar.xz"
SUMS+="87b565e89a9a684fe4ebeeddb8399dce2599f9c9049854ca8c0dfbdea0e21912  $GMP.tar.xz"
SUMS+="b587e083eb65a8b394e833dea1744f21af3f0e413a448c17536b5549ae42a4c2  $ISL.tar.bz2"
SUMS+="5833cd780fd96614b722040041d8242c102ad7eb956d6f102760bc0faad0a08e  linux-$KERNEL_VERSION.tar.xz"
SUMS+="6985c538143c1208dcb1ac42cedad6ff52e267b47e5f970183a3e75125b43c2e  $MPC.tar.gz"
SUMS+="67874a60826303ee2fb6affc6dc0ddd3e749e9bfcb4c8655e3953d0458a6e16e  $MPFR.tar.xz"
for SUM in $SUMS; do
  if ! echo "$SUM" | shasum -a 256 -c - || { echo "\033[1;31m$SUM checksum mismatch.  Try again. :/\033[0m"; exit 1; }
done
```

JL2210: Not bad.

JL2210: But your shasum for the Linux sources is off.

uberhacker: How so?

JL2210: I changed the sources from 3.18.1 to 3.18

JL2210: Also, your `if` statement is off.

JL2210: Your `for` statement won't work, because the SUMS variable has spaces in it. It also doesn't have any newlines.

uberhacker: Dude, the `IFS` line makes it work with newlines.

uberhacker: Your comments came after I edited.

JL2210: Yes, but your edits are inside the `Load more...` prompt.

uberhacker: Are you going to add sha256 sums?

tools/build_cros_gcc
```diff
@@ -1,178 +1,210 @@
...
 
+ set -e
+
+ # '\033[1;31m' outputs red text. '\033[0m' reverts it to normal.
+ command -v crew &>/dev/null || { echo -e "\033[1;31mcrew command not found.\033[0m"; exit 1; }
+
  GLIBC=glibc-$(crew search glibc -v | tail -1 | cut -d' ' -f2 | cut -d'-' -f1)
- BINUTILS=binutils-2.31.1
+ BINUTILS=binutils-$(crew search binutils -v | tail -1 | cut -d' ' -f2 | cut -d'-' -f1)
  GCC=gcc-8.2.0
```
uberhacker: `GCC=$(crew search gcc -v | tail -1 | cut -d' ' -f2 | cut -d'-' -f1)`?  Nothing wrong with gcc 7 and it makes sure when gcc is updated, the script doesn't need modified.  Also, it emulates the current crew build environment more closely which is important for testing purposes.

tools/build_cros_gcc
```diff
@@ -1,178 +1,210 @@
...
- wget https://www.kernel.org/pub/linux/kernel/v"$MAJOR_VERSION".x/linux-$KERNEL_VERSION.tar.xz
- wget https://ftpmirror.gnu.org/gnu/libc/$GLIBC.tar.xz
- wget https://ftpmirror.gnu.org/gnu/mpfr/$MPFR.tar.xz
- wget https://ftpmirror.gnu.org/gnu/gmp/$GMP.tar.xz
- wget https://ftpmirror.gnu.org/gnu/mpc/$MPC.tar.gz
- wget http://isl.gforge.inria.fr/$ISL.tar.bz2
- wget https://www.bastoul.net/cloog/pages/download/$CLOOG.tar.gz
+ SUMS="5d20086ecf5752cc7d9134246e9588fa201740d540f7eb84d795b1f7a93bca86  $BINUTILS.tar.xz\n"
+ SUMS+="325adf3710ce2229b7eeb9e84d3b539556d093ae860027185e7af8a8b00a750e  $CLOOG.tar.gz\n"
+ SUMS+="196c3c04ba2613f893283977e6011b2345d1cd1af9abeac58e916b1aab3e0080  $GCC.tar.xz\n"
```
uberhacker: This sha256 hash will change if we use gcc 7.3.0.

tools/build_cros_gcc
```diff
@@ -1,178 +1,210 @@
...
+     ./${ISL} \
+     ./${GLIBC}.tar.xz \
+     ./${GMP}.tar.xz \
+     ./${GCC}.tar.xz \
+     ./${BINUTILS}.tar.xz \
+     ./linux-${KERNEL_VERSION}.tar.xz \
+     ./${MPFR}.tar.xz \
+     ./${CLOOG}.tar.gz \
+     ./${ISL}.tar.bz2 \
+     ./${MPC}.tar.gz || true
```
uberhacker: `./${MPC}.tar.gz ./sha256sum || true`?

tools/build_cros_gcc
```diff
@@ -1,178 +1,210 @@
...
+ rm -rf \
+ ./${GLIBC} \
+ ./${GMP} \
+ ./${GCC} \
+ ./${BINUTILS} \
+ ./linux-${KERNEL_VERSION} \
+ ./${MPFR} \
+ ./${CLOOG} \
+ ./${ISL} \
+ ./${MPC}
```
uberhacker: `./sha256sum`?  We should clean all residue. :)

tools/build_cros_gcc
JL2210: GCC 7 causes errors when cross compiling. Read the GCC 8 ChnageLog.

tools/build_cros_gcc
JL2210: IF. Read previous comments.

tools/build_cros_gcc
JL2210: OK.

tools/build_cros_gcc
JL2210: Yeah.

tools/build_cros_gcc
```diff
@@ -1,178 +1,210 @@
...
+   fi
+ fi
+ if [[ -z "${PREFIX}" ]]; then
+   export PREFIX="$(crew const CREW_PREFIX | cut -d'=' -f2)"
  fi
- if test ! $3; then
-   $0 $1 $2 $(crew const CREW_PREFIX | cut -d'=' -f2)
-   exit 0
+ if [[ -z "${JOBS}" ]]; then
+   export JOBS="1"
```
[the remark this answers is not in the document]
JL2210: The script fails if you use more than one processor, just like you said.

tools/build_cros_gcc
```diff
@@ -1,178 +1,210 @@
...
+   export PREFIX="$1"
+   shift
+ else
+   if ! [[ "$1" =~ "^[0-9]+$" ]] &>/dev/null; then
+     export JOBS="$1"
+     shift
+   else if { mkdir -p "$1" &>/dev/null; }; then
+     export PREFIX="$1"
+     shift
    fi
```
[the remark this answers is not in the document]
JL2210: Not necessarily. This makes the script versatile to the point that you can remove arguments entirely.

tools/build_cros_gcc
uberhacker: There is no mention of cross-compiling fixes anywhere at https://www.gnu.org/software/gcc/gcc-8/changes.html.  Also, no mentions of cross-compiling bugs at https://gcc.gnu.org/bugzilla/buglist.cgi?bug_status=RESOLVED&resolution=FIXED&target_milestone=7.3.

tools/build_cros_gcc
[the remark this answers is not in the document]
uberhacker: Also, no need to export variables.  In fact, you shouldn't in case it affects other scripts that use variables.

tools/build_cros_gcc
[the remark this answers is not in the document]
uberhacker: If the Chromebook has only 1 processor, `nproc --all` will return 1.  There is never a time `nproc --all` will report more processors than physically possible but it will return the most processors available which would be the most efficient default.  Actually, you don't need the export either and `JOBS` is not the true meaning of the variable.   `NPROC` might make more sense.

tools/build_cros_gcc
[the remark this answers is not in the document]
JL2210: Oh. I didn't know that's what `export` did. Will fix.

tools/build_cros_gcc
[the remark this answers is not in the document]
JL2210: JOBS means the number of jobs. What I mean is, building some of these in parallel make the compilation fail, or skip essential parts. I made it default to the safest option, only 1.

JL2210:
I can have a sha256sum or I can have crew-specific versions. I don't think I can do both.

tools/build_cros_gcc
JL2210: Yeah. Sorry, misleading comment. I encountered issues with the startfiles on GCC 7.

uberhacker:
@JL2210: Are you going to do any more work on this or should we just close it?

JL2210:
@uberhacker: What else needs done? I still have the local repo.

tools/build_cros_gcc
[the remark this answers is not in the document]
JL2210: Fixed in latest

tools/build_cros_gcc
JL2210: Done.

tools/build_cros_gcc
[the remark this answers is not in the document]
JL2210: Done.

JL2210:
Done. How is this, @uberhacker?

tools/build_cros_gcc
```diff
@@ -1,178 +1,211 @@
...
+   shift
+ fi
+ if [[ -d $1 ]] &>/dev/null; then
+   PREFIX="$1"
+   shift
+ else
+   if ! [[ "$1" =~ "^[0-9]+$" ]] &>/dev/null; then
+     JOBS="$1"
+     shift
+   else if { mkdir -p "$1" &>/dev/null; }; then
```
uberhacker: This is not recommended.  Why do you need to create a directory?

tools/build_cros_gcc
```diff
@@ -1,178 +1,212 @@
...
- wget https://www.kernel.org/pub/linux/kernel/v"$MAJOR_VERSION".x/linux-$KERNEL_VERSION.tar.xz
- wget https://ftpmirror.gnu.org/gnu/libc/$GLIBC.tar.xz
- wget https://ftpmirror.gnu.org/gnu/mpfr/$MPFR.tar.xz
- wget https://ftpmirror.gnu.org/gnu/gmp/$GMP.tar.xz
- wget https://ftpmirror.gnu.org/gnu/mpc/$MPC.tar.gz
- wget http://isl.gforge.inria.fr/$ISL.tar.bz2
- wget https://www.bastoul.net/cloog/pages/download/$CLOOG.tar.gz
+ SUMS="5d20086ecf5752cc7d9134246e9588fa201740d540f7eb84d795b1f7a93bca86  $BINUTILS.tar.xz\n"
+ SUMS+="325adf3710ce2229b7eeb9e84d3b539556d093ae860027185e7af8a8b00a750e  $CLOOG.tar.gz\n"
+ SUMS+="832ca6ae04636adbb430e865a1451adf6979ab44ca1c8374f61fba65645ce15c  $GCC.tar.xz\n"
```
uberhacker: Has this sha256 sum been updated to match the current crew gcc version?

JL2210: Yes.

tools/build_cros_gcc
JL2210: I don't. I'll fix that.

JL2210: Fixed.

JL2210:
@uberhacker, how is this now?

tools/build_cros_gcc
```diff
@@ -1,178 +1,212 @@
...
- # '\033[1;31' outputs red text. '\033[0m' reverts it to normal.
- command -v crew > /dev/null || { echo "\033[1;31mcrew command not found.\033[0m"; exit 1; }
+ TARGET="$1" && shift
 
- if test ! $2; then
-   echo "$1" | grep x86_64 1>/dev/null
-   if [ $? == 0 ]; then
-   $0 $1 x86 $3;
-   exit 0;
+ if ! [[ "$1" =~ "^[0-9]+$" ]] &>/dev/null; then
```
uberhacker: Shouldn't this conditional exclude the `!`?

tools/build_cros_gcc
```diff
@@ -1,178 +1,212 @@
...
+   JOBS="$1"
+   shift
+ else
+   shift
+ fi
+ if [[ -d $1 ]] &>/dev/null; then
+   PREFIX="$1"
+   shift
+ else
+   if ! [[ "$1" =~ "^[0-9]+$" ]] &>/dev/null; then
```
uberhacker: No `!`?

tools/build_cros_gcc
```diff
@@ -1,178 +1,212 @@
...
+ fi
+ if [[ -z "${ARCH}" ]]; then
+   if { echo "${TARGET}" | grep x86_64 1>/dev/null; }; then
+     ARCH="x86_64";
    fi
-   echo "$1" | grep armv7l 1>/dev/null
-   if [ $? == 0 ]; then
-   $0 $1 arm $3;
-   exit 0;
+   if { echo "${TARGET}" | grep i686 1>/dev/null; }; then
```
uberhacker: These conditionals should probably be exact matches.

tools/build_cros_gcc
```diff
@@ -1,178 +1,212 @@
...
- ln -s ../$ISL isl
- ln -s ../$MPC mpc
- ln -s ../$GMP gmp
- ln -s ../$MPFR mpfr
+ cd ${GCC}
+ GCC_PATCHES="gcc-8.2.0-isl-0.20.patch"
+ for PATCH in ${GCC_PATCHES}; do
+   wget https://github.com/JL2210/patches/raw/master/${PATCH}
+   patch -Np1 -i ${PATCH}
+ done
```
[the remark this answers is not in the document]
JL2210: Oh. There are undefined references to isl variables because isl 0.20 has changed their header files. The patch only adds two `#include`s.

tools/build_cros_gcc
uberhacker: It opens the door to put anything into `TARGET` that has that keyword string in it and that is good enough.  How about something like this?
```
VALID_TARGETS="aarch64-cros-linux-gnu x86_64-cros-linux-gnu i686-cros-linux-gnu armv7l-cros-linux-gnueabihf'"
VALID=
for t in $VALID_TARGETS; do
  if [ "$t" == "$TARGET" ]; then
    VALID=1
  fi
done
if ! $VALID; then
  echo "Not a valid TARGET.  Please try again."
  exit 1
fi
```

JL2210: Then what's this?
```
armv7l-cros-linux-gnueabihf)
 true;;
 aarch64-cros-linux-gnu)
 true;;
 i686-cros-linux-gnu)
 true;;
 x86_64-cros-linux-gnu)
 true;;
 *)
 echo Invalid target or option \'$1\'	
 $0 -h	
 exit 1;;
```

JL2210: Don't worry. I covered all bases on that.

JL2210: Yeah. Can you try it out? Also, to rephrase: I have all the bases covered on blocking all targets but those four.

rafi16jan:
how to use this

PR #2703, opened 2018-10-01
Update yarn to remove pesky outdated text during yarn use.

skiman6010:
Tested on x86_64.


uberhacker:
I see you put a ton of effort into this update. :-P

PR #2704, opened 2018-10-02
Added Sublime Merge.

skiman6010:
Can someone test on a fresh install? The dependencies might put you in a loop. I wasn't able to test. Only works on x86_64 as well as far as I am aware.

### DIFF
--- a/packages/sublime_merge.rb
+++ b/packages/sublime_merge.rb
@@ -1,0 +1,33 @@
+require 'package'
+
+class Sublime_merge < Package
+  description 'A snappy UI, three-way merge tool,side-by-side diffs, syntax highlighting, and more.'
+  homepage 'https://www.sublimemerge.com/'
+  version 'Build 1065'
+  case ARCH
+  when 'x86_64'		  
+	source_url 'https://download.sublimetext.com/sublime_merge_build_1065_x64.tar.xz'
+	source_sha256 'ebe9029ea71543a101dcfd09c5540759759c1e33fea31d4b98256803b7ceaf97'
+  end
+  
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'sommelier'
+
+  def self.install
+    system "mkdir", "-p", "#{CREW_DEST_PREFIX}/share/sublime_merge"
+    system "mkdir", "-p", "#{CREW_DEST_PREFIX}/bin"
+    system "cp", "-rpa", ".", "#{CREW_DEST_PREFIX}/share/sublime_merge/"
+    system "ln", "-s", "#{CREW_PREFIX}/share/sublime_merge/sublime_merge", "#{CREW_DEST_PREFIX}/bin/sublime_merge"
+  end
+
+  def self.postinstall
+    puts
+    puts 'Congratulations! You have installed Sublime Merge on Chrome OS!'.lightgreen
+    puts 'Now, please run \'sublime_merge\' to start Sublime Merge.'.lightgreen
+    puts
+  end
+end


packages/sublime_merge.rb
```diff
@@ -0,0 +1,41 @@
...
+class Sublime_merge < Package
+  description 'A snappy UI, three-way merge tool,side-by-side diffs, syntax highlighting, and more.'
+  homepage 'https://www.sublimemerge.com/'
+  version 'Build 1065'
+  case ARCH
+  when 'x86_64'		  
+	source_url 'https://download.sublimetext.com/sublime_merge_build_1065_x64.tar.xz'
+	source_sha256 'ebe9029ea71543a101dcfd09c5540759759c1e33fea31d4b98256803b7ceaf97'
+  else
+    puts 'Merge is not supported on your system yet.'
```
uberhacker: Probably should just remove this message, otherwise, it shows up when you `crew search merge` on arm or i686.

packages/sublime_merge.rb
```diff
@@ -0,0 +1,41 @@
+require 'package'
+
+class Sublime_merge < Package
+  description 'A snappy UI, three-way merge tool,side-by-side diffs, syntax highlighting, and more.'
```
uberhacker: The downside: _"Sublime Merge may be downloaded and evaluated for free, however a license must be purchased for continued use."_

packages/sublime_merge.rb
```diff
@@ -0,0 +1,41 @@
...
+  depends_on 'xdg_base'
+  depends_on 'sommelier'
+
+  def self.install
+    case ARCH
+    when 'x86_64'
+      system "mkdir", "-p", "#{CREW_DEST_PREFIX}/share/sublime_merge"
+	  system "mkdir", "-p", "#{CREW_DEST_PREFIX}/bin"
+      system "cp", "-rpa", ".", "#{CREW_DEST_PREFIX}/share/sublime_merge/"
+	  system "ln", "-s", "#{CREW_PREFIX}/share/sublime_merge/sublime_merge", "#{CREW_DEST_PREFIX}/bin/sublime_merge"
```
uberhacker: Can we remove the tabs please? ;)

packages/sublime_merge.rb
```diff
@@ -0,0 +1,41 @@
...
+    else
+      puts 'Sublime Merge is currently not supported on your system.'.lightred
+      exit 1
```
uberhacker: This is not needed.  It will abort on the first conditional if the source is not found.

packages/sublime_merge.rb
skiman6010: Friggin VS Code. I'm switching back to vim now...

packages/sublime_merge.rb
skiman6010: Hoping they change their mind about that.

packages/sublime_merge.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+  end
+
+  def self.postinstall
+    puts
+    puts 'Congratulations! You have installed Sublime Merge on Chrome OS!'.lightgreen
+    puts 'Now, please run \'sublime_merge\' to start Sublime Merge.'.lightgreen
```
uberhacker: I wonder where you plagurized this message...could it be from code?

packages/sublime_merge.rb
```diff
@@ -0,0 +1,41 @@
...
+	source_sha256 'ebe9029ea71543a101dcfd09c5540759759c1e33fea31d4b98256803b7ceaf97'
+  else
+    puts 'Merge is not supported on your system yet.'
+  end
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'xdg_base'
```
uberhacker: Why do we need xdg_base?

uberhacker:
Oooo...new package.  Nice!

packages/sublime_merge.rb
skiman6010: Nah man I 'leveraged' that message.

packages/sublime_merge.rb
skiman6010: Do you want me to remove the entire conditional or just the else statement?

packages/sublime_merge.rb
skiman6010: Just a leftover from Code that I was debating whether it would be useful or not. I will remove it and re-add it at a later time if an issue arises.

packages/sublime_merge.rb
uberhacker: Just the else and message lines.

packages/sublime_merge.rb
uberhacker: You can remove the entire conditional here.

skiman6010:
Updated and squashed. How's that look for my shiny new package?

uberhacker:
Looks good.  Where is good ol' sublime text?  Could it be the next package I see from the skimeister?

skiman6010:
I don't use Sublime Text but sounds like it should be my next target!

packages/sublime_merge.rb
JL2210: lol

JL2210: plagiarized

PR #2713, opened 2018-10-02
Remove redundant dependencies to prevent a dependency loop

### LINKED ISSUE #2706: [sommelier] Unable to install due to dependency loop
This happens on a fresh crew install as well.  Fix coming soon...

uberhacker:
Fixes #2706.


JL2210:
Ah. I see the problem. It was Mesa depending on LLVM.

PR #2715, opened 2018-10-02
Update Avocado package

JL2210:
Tested on x86_64

### DIFF
--- a/packages/avocado.rb
+++ b/packages/avocado.rb
@@ -3,36 +3,29 @@ require 'package'
 class Avocado < Package
   description 'Avocado is a next generation testing framework inspired by Autotest and modern development tools such as git.'
   homepage 'http://avocado-framework.github.io/'
-  version '60.0'
-  source_url 'https://github.com/avocado-framework/avocado/archive/60.0.tar.gz'
-  source_sha256 'cadafe363a315a1dd3b332f66407da68748bcbe534153e20c59d588daa83a4ce'
+  version '64.0'
+  source_url 'https://github.com/avocado-framework/avocado/archive/64.0.tar.gz'
+  source_sha256 'bf1434d61754f389771dc38beb66a62ee6cca95f5224b17eb2a4ca2f7abf22ab'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/avocado-60.0-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/avocado-60.0-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/avocado-60.0-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/avocado-60.0-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'f434f34711084c9e425b6c6ea9f6a3412274cb41b2ab59027bc3919c811a2c7a',
-     armv7l: 'f434f34711084c9e425b6c6ea9f6a3412274cb41b2ab59027bc3919c811a2c7a',
-       i686: 'f8284b6aa7b57571759724904979a0052ff42abd0009b2b785c03440ad365437',
-     x86_64: 'c8db96fe48bdade9f974931f6eac557f7e2d65292a335449b4fdbc6c47b4ee53',
   })
 
-  depends_on 'python27' unless File.exists? "#{CREW_PREFIX}/bin/pip"
+  depends_on 'python3'
   depends_on 'xzutils'
+  depends_on 'setuptools'
+
+  def self.build
+    system "python3 setup.py build"
+  end
 
   def self.install
     system "pip install avocado-framework -r requirements.txt --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
+    system "python3 setup.py install --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
     system "mkdir -p $HOME/.config/avocado"
     system "ln -sf #{CREW_PREFIX}/etc/avocado/avocado.conf $HOME/.config/avocado/avocado.conf"
     system "mkdir -p #{CREW_DEST_DIR}/$HOME/.config/avocado"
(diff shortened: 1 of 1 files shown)


packages/avocado.rb
```diff
@@ -3,36 +3,29 @@
...
   })
   binary_sha256 ({
-    aarch64: 'f434f34711084c9e425b6c6ea9f6a3412274cb41b2ab59027bc3919c811a2c7a',
-     armv7l: 'f434f34711084c9e425b6c6ea9f6a3412274cb41b2ab59027bc3919c811a2c7a',
-       i686: 'f8284b6aa7b57571759724904979a0052ff42abd0009b2b785c03440ad365437',
-     x86_64: 'c8db96fe48bdade9f974931f6eac557f7e2d65292a335449b4fdbc6c47b4ee53',
   })
 
-  depends_on 'python27' unless File.exists? "#{CREW_PREFIX}/bin/pip"
+  depends_on 'python3'
```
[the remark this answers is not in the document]
JL2210: It's newer. I figured it's better to switch mow than when Python 2 is deprecated.

packages/avocado.rb
```diff
@@ -3,36 +3,29 @@
...
   depends_on 'xzutils'
+  depends_on 'setuptools'
+
+  def self.build
+    system "python3 setup.py build"
+  end
 
   def self.install
-    system "pip install avocado-framework -r requirements.txt --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
+    system "pip2 install avocado-framework -r requirements.txt --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
```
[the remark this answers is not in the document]
JL2210: Yeah.

packages/avocado.rb
```diff
@@ -3,36 +3,29 @@
...
-    aarch64: 'f434f34711084c9e425b6c6ea9f6a3412274cb41b2ab59027bc3919c811a2c7a',
-     armv7l: 'f434f34711084c9e425b6c6ea9f6a3412274cb41b2ab59027bc3919c811a2c7a',
-       i686: 'f8284b6aa7b57571759724904979a0052ff42abd0009b2b785c03440ad365437',
-     x86_64: 'c8db96fe48bdade9f974931f6eac557f7e2d65292a335449b4fdbc6c47b4ee53',
   })
 
-  depends_on 'python27' unless File.exists? "#{CREW_PREFIX}/bin/pip"
+  depends_on 'python3'
   depends_on 'xzutils'
+  depends_on 'setuptools'
```
uberhacker: Probably should add `depends_on 'xdg_base'` to accommodate `$HOME/.config/avocado/avocado.conf`.

packages/avocado.rb
```diff
@@ -3,36 +3,29 @@
...
+
+  def self.build
+    system "python3 setup.py build"
+  end
 
   def self.install
-    system "pip install avocado-framework -r requirements.txt --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
+    system "pip3 install avocado-framework -r requirements.txt --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
+    system "python3 setup.py install --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
```
uberhacker: You don't need the self.build section or this line if you use pip3 to install.

uberhacker: This should lock down the version:
```
system "pip3 install avocado-framework==64.0 -r requirements.txt --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
```

packages/avocado.rb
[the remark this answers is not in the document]
uberhacker: This line should be removed if using `depends_on 'setuptools'`.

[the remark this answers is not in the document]
JL2210: True.

packages/avocado.rb
JL2210: OK.

packages/avocado.rb
JL2210: This makes sure the avocado script is chmod 755. Why download the sources if you don't use anything but requirements.txt?

uberhacker: Exactly, why use `python3 setup.py install` if you are already installing via `pip3 install`?

JL2210: So I can change the source URL to file:///dev/null and change the sha256sum to the sha256sum of an empty file?

JL2210: I'm simply pulling in the dependencies with pip.

uberhacker: Remove the self.build section and change the line below and I'll merge it:
```
system "pip3 install -r requirements.txt --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
```

JL2210: Why? Isn't it better to build from source, considering that the executable bit isn't set with the pip package?

JL2210: Fair enough. I'll give you write access.

JL2210:
How's this now?

packages/avocado.rb
```diff
@@ -3,36 +3,25 @@
...
   })
 
-  depends_on 'python27' unless File.exists? "#{CREW_PREFIX}/bin/pip"
+  depends_on 'xdg_base'
   depends_on 'xzutils'
+  depends_on 'setuptools'
 
   def self.install
-    system "pip install avocado-framework -r requirements.txt --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
+    system "pip3 install avocado-framework==64.0 -r requirements.txt --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
```
uberhacker: Take out the avocado-framework argument:
```
system "pip3 install -r requirements.txt --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
```

JL2210: OK. I understand now.

JL2210: What I don't understand is why we couldn't have done that with AWS.

uberhacker:
You are right.  We could have done it this way with aws also.  The point I've been trying to drive home is if you want to use pip3 to install the package, then use pip3.  Otherwise, just use pip3 to install the requirements ONLY and then the package via python3 setup.py install.

JL2210:
Yeah. Thank you for explaining. Ready?

PR #2723, opened 2018-10-03
Update dpkg package

JL2210:
Update dpkg to 1.19.1
Tested on x86_64
Version 1.19.1 of dpkg resolves po errors of previous version.
Safe for pre-compiled binaries.

### DIFF
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -3,20 +3,20 @@ require 'package'
 class Dpkg < Package
   description 'A medium-level package manager for Debian'
   homepage 'https://anonscm.debian.org/git/dpkg/'
-  version '1.18.25'
-  source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.18.25/dpkg-1.18.25.tar.gz'
-  source_sha256 '93470770161bb15fd7988da1890aecda322e19c3fc858d6cb8f33417e5fe01e9'
+  version '1.19.1'
+  source_url 'http://http.debian.net/debian/pool/main/d/dpkg/dpkg_1.19.1.tar.xz'
+  source_sha256 'ae3978a6b7bddc3e3196804ae0d49ea008c84a8a7a60b7d212af1e1d469e7ccf'
 
   depends_on 'bz2'
   depends_on 'xzutils'
 
   def self.build
-    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git -b 1.18.25"
-    Dir.chdir ("dpkg") do
-      system "autoreconf -i -f"
-      system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
-      system "make"
-    end
+    system "/usr/bin/env",
+           "PERL_LIBDIR=#{CREW_PREFIX}/lib/perl5/site_perl/",
+           "./configure",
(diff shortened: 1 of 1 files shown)


packages/dpkg.rb
```diff
@@ -3,20 +3,20 @@
 class Dpkg < Package
   description 'A medium-level package manager for Debian'
   homepage 'https://anonscm.debian.org/git/dpkg/'
-  version '1.18.25'
-  source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.18.25/dpkg-1.18.25.tar.gz'
-  source_sha256 '93470770161bb15fd7988da1890aecda322e19c3fc858d6cb8f33417e5fe01e9'
+  version '1.19.1'
+  source_url 'http://http.debian.net/debian/pool/main/d/dpkg/dpkg_1.19.1.tar.xz'
```
uberhacker: What is wrong with `https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.19.1/dpkg-1.19.1.tar.gz`?

packages/dpkg.rb
```diff
@@ -3,20 +3,20 @@
...
 
   def self.build
-    system "git clone https://salsa.debian.org/dpkg-team/dpkg.git -b 1.18.25"
-    Dir.chdir ("dpkg") do
-      system "autoreconf -i -f"
-      system "./configure --libdir=#{CREW_LIB_PREFIX} --prefix=#{CREW_PREFIX}"
-      system "make"
-    end
+    system "/usr/bin/env",
+           "PERL_LIBDIR=#{CREW_PREFIX}/lib/perl5/site_perl/",
```
uberhacker: Can you please clue me in why this is needed?

packages/dpkg.rb
```diff
@@ -3,20 +3,20 @@
...
   homepage 'https://anonscm.debian.org/git/dpkg/'
-  version '1.18.25'
-  source_url 'https://salsa.debian.org/dpkg-team/dpkg/-/archive/1.18.25/dpkg-1.18.25.tar.gz'
-  source_sha256 '93470770161bb15fd7988da1890aecda322e19c3fc858d6cb8f33417e5fe01e9'
+  version '1.19.1'
+  source_url 'http://http.debian.net/debian/pool/main/d/dpkg/dpkg_1.19.1.tar.xz'
+  source_sha256 'ae3978a6b7bddc3e3196804ae0d49ea008c84a8a7a60b7d212af1e1d469e7ccf'
 
   depends_on 'bz2'
   depends_on 'xzutils'
```
uberhacker: `depends_on 'perl'`?

uberhacker: Actually, this is in "core" so nevermind.

packages/dpkg.rb
JL2210: REmember the fact that we had tio use git before? This source tarball has a configure script.

packages/dpkg.rb
JL2210: `PERL_LIBDIR`? That's where dpkg installs its perl module files. They are needed for `dpkg-architecture` to work. If you're talking about `/usr/bin/env`, try doing it without. It fails with `bash: PERL_LIBDIR=/usr/local/lib/perl5/site_perl/: command not found`

PR #2724, opened 2018-10-04
bump emacs to 26.1, enable x11 via lucid

s-urbaniak:
## Description
- bump Emacs to 26.1
- enable emacs X11 (via lucid toolkit) which makes the GUI work via sommelier

## Addtional information
I removed the sha's for the binaries.

Works properly:
- [x] x86_64
- [ ] aarch64 (reasons why it doesn't) - didn't test

### DIFF
--- a/packages/emacs.rb
+++ b/packages/emacs.rb
@@ -3,35 +3,21 @@ require 'package'
 class Emacs < Package
   description 'An extensible, customizable, free/libre text editor - and more.'
   homepage 'http://www.gnu.org/software/emacs/'
-  version '25.3'
-  source_url 'https://ftp.gnu.org/gnu/emacs/emacs-25.3.tar.xz'
-  source_sha256 '253ac5e7075e594549b83fd9ec116a9dc37294d415e2f21f8ee109829307c00b'
-  
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/emacs-25.3-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/emacs-25.3-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/emacs-25.3-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/emacs-25.3-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: '9efee852141ebcb349ec99766739ef2babbb6d7b83d66725e32b74d406ab2e49',
-     armv7l: '9efee852141ebcb349ec99766739ef2babbb6d7b83d66725e32b74d406ab2e49',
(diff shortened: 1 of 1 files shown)


JL2210:
You need to remove the binaries until we can recompile.

packages/go.rb
```diff
@@ -3,15 +3,15 @@
 class Go < Package
```
JL2210: Don't update multiple packages in one PR.

s-urbaniak:
@JL2210 addressed, PTAL

JL2210:
This looks good as well. @uberhacker?

packages/emacs.rb
```diff
@@ -3,35 +3,27 @@
...
+  depends_on 'libtiff'
 
   def self.build
     system "./configure \
             --prefix=#{CREW_PREFIX} \
             --localstatedir=#{CREW_PREFIX}/share \
-            --with-x=no \
+            --with-x=yes \
+            --with-x-toolkit=lucid \
+            --with-gif=no \
```
uberhacker: What is wrong with including libgif?

packages/emacs.rb
```diff
@@ -3,35 +3,27 @@
...
+  version '26.1'
+  source_url 'https://ftp.gnu.org/gnu/emacs/emacs-26.1.tar.xz'
+  source_sha256 '1cf4fc240cd77c25309d15e18593789c8dbfba5c2b44d8f77c886542300fd32c'
 
   depends_on "zlibpkg" => :build
   depends_on "diffutils" => :build
   depends_on "m4" => :build
   depends_on "autoconf" => :build
   depends_on "automake" => :build
   depends_on "ncurses"
```
uberhacker: All `depends_on ...` lines above this comment can be removed since they're already included in chromebrew "core".

uberhacker:
Thank you for the package updates @s-urbaniak!  Please see the comments.

packages/emacs.rb
s-urbaniak: I simply couldn't find it in the packages and disabled it (for now), unless I missed some pkg name. I can try to add lbgif independently.

packages/emacs.rb
s-urbaniak: sounds good :+1:

packages/emacs.rb
```diff
@@ -3,35 +3,27 @@
...
+  depends_on 'sommelier'
```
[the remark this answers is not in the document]
s-urbaniak: Sounds good, I'll put it to the end of the list.

packages/emacs.rb
JL2210: There is no `libgif` in Chromebrew.

uberhacker: There is no libgif package but he could use `--with-gif=yes`.  Is there any reason this would cause a problem @s-urbaniak?

s-urbaniak: @uberhacker the compilation will fail during linkage. I _guess_ the missing feature here will be GIFs not to be displayed in emacs buffers.

When I have a bit more time I can package libgif in a separate PR.

s-urbaniak:
@uberhacker addressed comments minus libgif, PTAL

PR #2742, opened 2018-10-09
Add glfw package

codekrafter:
Adds GLFW as a package

Works properly:
- [x] x86_64
- [ ] aarch64 (has not tested)

### DIFF
--- a/packages/glfw.rb
+++ b/packages/glfw.rb
@@ -10,7 +10,12 @@ class Glfw < Package                 # The first character of the class name mus
   depends_on 'cmake' => :build      # packages with "=> :build" are only required if you're building from source 
 
   def self.build                   # the steps required to build the package
-    system "cmake","-DCMAKE_INSTALL_PREFIX:PATH=#{CREW_PREFIX}", "-DCMAKE_LIB_SUFFIX=64"
+    case ARCH
+    with 'x86_64'
+      system "cmake","-DCMAKE_INSTALL_PREFIX:PATH=#{CREW_PREFIX}", "-DCMAKE_LIB_SUFFIX=64"
+    else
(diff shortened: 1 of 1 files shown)


packages/glfw.rb
```diff
@@ -0,0 +1,24 @@
...
+  description 'GLFW is a multi-platform library for OpenGL, OpenGL ES, Vulkan, window and input '
+  homepage 'http://www.glfw.org/'
+  version '3.2.1'
+  source_url 'https://github.com/glfw/glfw/archive/3.2.1.zip'
+  source_sha256 '0c623f65a129c424d0fa45591694fde3719ad4a0955d4835182fda71b255446f'   # Use the command "sha256sum"
+
+  depends_on 'cmake' => :build      # packages with "=> :build" are only required if you're building from source 
+
+  def self.build                   # the steps required to build the package
+    system "cmake","-DCMAKE_INSTALL_PREFIX:PATH=#{CREW_PREFIX}"
```
uberhacker: We probably want to abstract the library path also so x86_64 Chromebooks look in `/usr/local/lib64` instead of `/usr/local/lib`.  Probably something like `-DCMAKE_LIBRARY_PATH=#{CREW_LIB_PREFIX}`.  You can execute `cmake -LAH` in the extracted source directory to find out what options are available to cmake.  If the above option isn't available, something like `-DCMAKE_LIB_SUFFIX=64` might work also.  Look at some of the other packages that are built with cmake for ideas.

uberhacker:
Thank you for the new package @codekrafter!  See the comment in the `Files changed` tab.

codekrafter:
I added "-DCMAKE_LIB_SUFFIX=64" so that it puts in it lib64.

uberhacker:
To test your build, execute `crew files glfw` and make sure the library files are located in `/usr/local/lib64` on your Chromebook.

PR #2748, opened 2018-10-10
Added ShPkg package

JL2210:
A simple dpkg-compatible package manager implemented in Bash.

This package manager is going to be the basis for my operating system, along with `shapt` (Soon to come!).

### DIFF
--- a/packages/shpkg.rb
+++ b/packages/shpkg.rb
@@ -1,0 +1,30 @@
+require 'package'
+
+class Shpkg < Package
+  description 'A simple dpkg-compatible package manager implemented in Bash.'
+  homepage 'https://github.com/JL2210/shpkg'
+  version '0.1.0'
+  source_url 'https://github.com/JL2210/shpkg/releases/download/v0.1.0/shpkg-0.1.0.tar.gz'
+  source_sha256 'a5a56e24160986d1392cc34efe48b00ace2e8e1df813ea5cddb0e129effc74ba'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.build
+    system 'make',
+           "PREFIX=#{CREW_PREFIX}",
+           "SYSCONFDIR=#{CREW_PREFIX}/etc",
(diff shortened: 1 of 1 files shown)


uberhacker:
Do we really need another package manager? ;)  Oh no!  Not another operating system!

uberhacker:
Why re-invent the wheel?  https://github.com/bpkg/bpkg

JL2210:
Not the same at all. I created a different format, and it only has dpkg COMPATIBILITY mode, for when dpkg isn't working.

JL2210:
The new format has an empty control file, and instead of the `debian-binary` file it has a `sha256sum` with the checksums of `data.tar.xz` and `control.tar.xz`.

uberhacker:
If this is a wrapper for dpkg, shouldn't that be a dependency for this package?  My point is...why?

JL2210:
This is a replacement for `dpkg`.

JL2210:
<whisper>I want to have Chromebrew be the package manager for my operating system, but it might need this first.</whisper>

JL2210:
I want Chromebrew to be the secondary package manager for my operating system. This will just be the upater for the system.

uberhacker:
Where are you getting the packages?  Don't you need a repository?

JL2210:
Yes, I do. I have them in the `binaries/` directory of my chroot right now. Usually, I would build them by hand. This package would just help by allowing me to build them outside of the chroot.

JL2210:
Plus, that would be what `shapt` will be for. If `dpkg` ever breaks, this will also be a fallback.

uberhacker:
I don't think we need this package here then, right?

uberhacker:
I wish I could encourage you to work on other things like maybe an apt package for this project?  I would think that would be a bit more useful than a new package manager that pulls binaries from who knows where, don't you think?

JL2210:
Fine. I suppose with `docbook` fixed, I could. One issue is that it will try to install binaries in `/usr`, though.

PR #2750, opened 2018-10-12
Bump llvm version number

JL2210:
Doesn't remove binaries, just bumps version number so llvm will be upgraded in `crew upgrade`.


JL2210:
Thank you. That's all we needed to do.

PR #2751, opened 2018-10-12
Update google_cloud_sdk from 198.0.0 to 220.0.0

uberhacker:
Tested on:
- [x] i686
- [x] x86_64


JL2210:
High five for removing the error.

PR #2752, opened 2018-10-12
Update julia from 1.0.0-1 to 1.0.1

uberhacker:
Tested on:
- [x] i686
- [x] x86_64

Binary builds for armv7l have been dropped.

### DIFF
--- a/packages/julia.rb
+++ b/packages/julia.rb
@@ -3,19 +3,14 @@ require 'package'
 class Julia < Package
   description 'Julia is a flexible dynamic language, appropriate for scientific and numerical computing'
   homepage 'https://julialang.org/'
-  version '1.0.0-1'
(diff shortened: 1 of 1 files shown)


JL2210:
Is it still possible for armv7l users to install?

uberhacker:
No, but if you are willing to compile from source and submit a PR, have at it.  Warning: Not for the faint of heart.

JL2210:
How bad is it?

uberhacker:
https://github.com/JuliaLang/julia/tree/v1.0.1#required-build-tools-and-external-libraries

JL2210:
These all look like they're in Chromebrew "core", and the ones that are not are automatically downloaded. Let me give it a shot.

PR #2764, opened 2018-10-17
Add gimp package

uberhacker:
GIMP is a cross-platform image editor available for GNU/Linux, OS X, Windows and more operating systems.  See https://www.gimp.org/.  This package does not include webkit.  Basically this means there is no help browser but you can still read the online documentation.  It also means the binary is much smaller and runs faster which is probably worth the tradeoff.  Depends on #2761, #2762 and #2763. 
 @superloach: I expect to see a pixel perfect Chromebrew logo soon! ;)

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

Config tip: If icons aren't rendering, from the menu select `Edit > Preferences > Icon Theme  > Legacy > OK`.

### DIFF
--- a/packages/gimp.rb
+++ b/packages/gimp.rb
@@ -1,0 +1,52 @@
+require 'package'
+
+class Gimp < Package
+  description 'GIMP is a cross-platform image editor available for GNU/Linux, OS X, Windows and more operating systems.'
+  homepage 'https://www.gimp.org/'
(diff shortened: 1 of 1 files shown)


JL2210:
Sweet! I think @superloach isn't feeling well right now. Check the Discord channel.

PR #2765, opened 2018-10-17
Add sysfsutils package

JL2210:
This package's purpose is to provide a set of utilities for interfacing
with sysfs, a virtual filesystem in Linux kernel versions 2.5+ that
provides a tree of system devices. While a filesystem is a very useful
interface, we've decided to provide a stable programming interface
that will hopefully make it easier for applications to query system devices
and their attributes.

Tested on x86_64

### DIFF
--- a/packages/sysfsutils.rb
+++ b/packages/sysfsutils.rb
@@ -1,0 +1,23 @@
+require 'package'
+
+class Sysfsutils < Package
+  description 'Libsysfs provides a stable programming interface to sysfs and eases querying system devices and their attributes.'
+  homepage 'https://github.com/Distrotech/sysfsutils'
+  version '2.1.0'
+  source_url 'http://http.debian.net/debian/pool/main/s/sysfsutils/sysfsutils_2.1.0+repack.orig.tar.gz'
+  source_sha256 'e865de2c1f559fff0d3fc936e660c0efaf7afe662064f2fb97ccad1ec28d208a'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.build
+    system "./configure", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}"
+    system 'make'
+  end
+
+  def self.install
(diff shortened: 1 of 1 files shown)


uberhacker:
Do any Chromebooks use this filesystem?  I guess my question is why do we need this?

JL2210:
Of course. Ever look at `/sys`? It's a dependency for another package.

uberhacker:
What is the "other" package that needs this dependency?

JL2210:
`rng_tools`, a program for gathering random data from device drivers and ahrdware random number generators. It uses this to access the `*random` devices in `/dev`. The `rng_tools` package is useful for generating GPG keys.

packages/sysfsutils.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Sysfsutils < Package
+  description 'Libsysfs provides a stable programming interface to sysfs and eases querying system devices and their attributes.'
```
uberhacker: Shouldn't this description mention sysfsutils instead of libsysfs?

packages/sysfsutils.rb
```diff
@@ -0,0 +1,23 @@
...
+  homepage 'https://github.com/Distrotech/sysfsutils'
+  version '2.1.0'
+  source_url 'http://http.debian.net/debian/pool/main/s/sysfsutils/sysfsutils_2.1.0+repack.orig.tar.gz'
```
uberhacker: Why don't you used the source from the github repo?

packages/sysfsutils.rb
JL2210: If you don't mind the description being slightly long, I could do that.

packages/sysfsutils.rb
JL2210: The github repo doesn't have releases. Debian just packed the latest git source code into a tarball ( From 2010 )

JL2210: How about a compromise? I got the source code from SourceForge.

packages/sysfsutils.rb
JL2210: Done.

PR #2766, opened 2018-10-17
Add rng_tools package

JL2210:
This is a random number generator daemon.

It monitors a set of entropy sources, and supplies entropy from them to the system kernel's /dev/random machinery.

Operation is fully documented in the man page, and should be fairly intuitive

### DIFF
--- a/packages/rng_tools.rb
+++ b/packages/rng_tools.rb
@@ -1,0 +1,76 @@
+require 'package'
+
+class Rng_tools < Package
+  description 'This is a random number generator daemon.'
+  homepage 'https://github.com/nhorman/rng-tools'
+  version '6.5'
+  source_url 'https://github.com/nhorman/rng-tools/archive/v6.5.tar.gz'
+  source_sha256 '5fecd904f7d01262b3209ff78dd3b9594aac8daa41badd5a1e6438658e80c36e'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'curl'
+  depends_on 'libgcrypt'
+  depends_on 'sysfsutils'
+  depends_on 'psmisc'
+
+  def self.build
+    system './autogen.sh'
+    system "./configure", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}"
+    system 'make'
+    system "echo '#!/bin/bash' > startrngd"
+    system "echo 'RNGD=#{CREW_PREFIX}/sbin/rngd' >> startrngd"
+    system "echo 'RNG=\$(pidof \$RNGD 2> /dev/null)' >> startrngd"
+    system "echo 'if [ -z \"\$RNG\" ]; then' >> startrngd"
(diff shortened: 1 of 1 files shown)


packages/rng_tools.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def self.build
+    system './autogen.sh'
+    system "./configure", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}"
+    system 'make'
+    system "echo '#!/bin/bash' > startrngd"
+    system "echo 'RNGD=#{CREW_PREFIX}/sbin/rngd' >> startrngd"
+    system "echo 'RNG=\$(pidof \$RNGD 2> /dev/null)' >> startrngd"
+    system "echo 'if [ -z \"\$RNG\" ]; then' >> startrngd"
+    system "echo '  sudo LD_LIBRARY_PATH=\$LD_LIBRARY_PATH rngd &>/dev/null' >> startrngd"
```
uberhacker: Not sure LD_LIBRARY_PATH is necessary.  Should already be set.

packages/rng_tools.rb
```diff
@@ -0,0 +1,76 @@
...
+    system "echo '  sleep 3' >> stoprngd"
+    system "echo 'fi' >> stoprngd"
+    system "echo 'RNG=\$(pidof \$RNGD 2> /dev/null)' >> stoprngd"
+    system "echo 'if [ -z \"\$RNG\" ]; then' >> stoprngd"
+    system "echo '  echo \"rngd process stopped\"' >> stoprngd"
+    system "echo 'else' >> stoprngd"
+    system "echo '  echo \"rngd process \$RNG is running\"' >> stoprngd"
+    system "echo '  exit 1' >> stoprngd"
+    system "echo 'fi' >> stoprngd"
+    system "chmod a+x startrngd stoprngd"
```
uberhacker: You can get rid of this line if you use the `install` command.  See below.

packages/rng_tools.rb
JL2210: It complains about a missing `libsysfs.so.2`. Not much I can do there.

packages/rng_tools.rb
JL2210: Got it.

packages/rng_tools.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "cp", "startrngd", "stoprngd", "#{CREW_DEST_PREFIX}/bin/"
```
uberhacker:
```
system "install -Dm755 startrngd #{CREW_DEST_PREFIX}/bin/startrngd"
system "install -Dm755 stoprngd #{CREW_DEST_PREFIX}/bin/stoprngd"
```

JL2210: Yeah.

PR #2767, opened 2018-10-17
Update sommelier from 1382ce0 to 1382ce08

uberhacker:
Add --scale option to sommelierd and improve the pid detection in the scripts.

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -3,24 +3,25 @@ require 'package'
 class Sommelier < Package
   description 'Sommelier works by redirecting X11 and Wayland programs to the built-in ChromeOS wayland server.'
   homepage 'https://chromium.googlesource.com/chromiumos/containers/sommelier'
-  version '1382ce0'
+  version '1382ce08'
   source_url 'https://chromium.googlesource.com/chromiumos/containers/sommelier/+/0.20/README?format=TEXT'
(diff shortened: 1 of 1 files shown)


JL2210:
Just a note, the scale would be different on Chromebooks with different DPIs. You might not set that variable.

uberhacker:
Leaving it unset produces the same problem.  At least now they know the option is available if they want to change it or not.  They also don't need to change the script.  They can change the environment variable in `~/.bashrc`.

JL2210:
Fair enough. A suggestion: Maybe you could add a note in the postinstall about that.

uberhacker:
It would have to be just one line.  The current post install message is almost using up the entire viewable screen already.

JL2210:
Maybe install a `bashrc.tmp` file in `/usr/local/tmp` and ask users to `cat /usr/local/tmp/bashrc.tmp >> ~/.bashrc`?

JL2210:
That would save space in the postinstall.

uberhacker:
The thing is, I want users to see what they're appending to `~/.bashrc` and understand what it means.  If you hide it, then they don't see and understand.

JL2210:
Have them first `cat /usr/local/tmp/bashrc.tmp`?

PR #2770, opened 2018-10-18
Add ndisc6 package

Tomo59:
Add ndisc6 package to have tools for IPv6.

Works properly:
- [x] x86_64

I don't have other arch to test. Also I don't know how to add binary archive on bintray (but it doesn't take long to compile).

### DIFF
--- a/packages/ndisc6.rb
+++ b/packages/ndisc6.rb
@@ -1,0 +1,25 @@
+require 'package'
+
+class Ndisc6 < Package
+  description 'Small collection of useful tools for IPv6 networking (ndisc6, rdisc6, tcptraceroute6, traceroute6, rdnssd).'
+  homepage 'https://www.remlab.net/ndisc6/'
+  version '1.0.3'
+  source_url 'https://www.remlab.net/files/ndisc6/ndisc6-1.0.3.tar.bz2'
+  source_sha256 '0f41d6caf5f2edc1a12924956ae8b1d372e3b426bd7b11eed7d38bc974eec821'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+  
+  depends_on 'perl'
+
+  def self.build
+    system "./configure", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}", "--disable-suid-install"
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+end


packages/ndisc6.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+  
+  depends_on 'perl'
+
+  def self.build
+    system "./configure --disable-suid-install"
```
uberhacker: Please add a --prefix option like below:
```
system './configure', "--prefix=#{CREW_PREFIX}", '--disable-suid-install'
```

uberhacker:
Thank you for the new package @Tomo59!  Please see the comment.

packages/ndisc6.rb
JL2210: Another useful flag would be `'--libdir=#{CREW_LIB_PREFIX}'`.

Tomo59:
Thanks for the comment, I just changed the configure command line

packages/ndisc6.rb
Tomo59: Fixed in c91a566

JL2210:
@uberhacker, this looks good! Ready to merge?

Tomo59:
For information, how could I put the bin archive on bintray ?

JL2210:
@Tomo59, it's complicated. Let @uberhacker deal with it.

uberhacker:
Easy for you to say @JL2210. :-P  If you are interested @Tomo59, I can send you an invite to join https://bintray.com/chromebrew.  The biggest challenge is to set up environments to compile on the 3 architectures (armv7l, i686 and x86_64).  If you have Chromebooks (hardware), that is going to be better than VMs (software).  I wasn't able to grab an old i686 Chromebook yet, so I compile via the VM thing.

JL2210:
Just run an i686 Chromium OS image on an x86_64 Chromebook.

JL2210:
And buy an AArch64 Chromebook and run armv7l Chromium OS on it.

uberhacker:
I don't think you can mix and match architectures.  Chromium OS was compiled for a specific hardware and likely won't work on something different.

JL2210:
It still works. Backwards compatibility.

JL2210:
64-bit computers can run 32-bit instructions.

Tomo59:
Thanks @uberhacker but I only have my own Chromebook which is x86_64. I'll let you put the binaries on bintray !

uberhacker:
Got you covered already.  See #2777.

PR #2772, opened 2018-10-19
Add Heroku update script for easy management of Heroku updates

JL2210:
Oh no! @skiman6010's worst fear!

This package pulls the latest tag form the heroku/cli git repository and then creates a functional Chromebrew package from there.

### DIFF
--- a/tools/heroku.sh
+++ b/tools/heroku.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-export PWD_HOLD="${PWD}"
+PWD_HOLD="${PWD}"
 
 cd /tmp
 if [[ -d /tmp/heroku-cli ]]; then
@@ -49,5 +49,3 @@ echo "    system \"mkdir -p #{CREW_DEST_PREFIX}/bin\"" >> heroku.rb
 echo "    system \"ln -s #{CREW_PREFIX}/share/heroku/bin/run #{CREW_DEST_PREFIX}/bin/heroku\"" >> heroku.rb
 echo "  end" >> heroku.rb
 echo "end" >> heroku.rb
-
-unset PWD_HOLD
--- a/tools/heroku.sh
+++ b/tools/heroku.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+
+export PWD_HOLD="${PWD}"
+
+cd /tmp
+if [[ -d /tmp/heroku-cli ]]; then
+  cd heroku-cli
+  git pull &>/dev/null
+  cd ..
+else
+  git clone https://github.com/heroku/cli.git heroku-cli &>/dev/null
+fi
+cd heroku-cli
+export HEROKU_V="$(git describe --abbrev=0 --tags)"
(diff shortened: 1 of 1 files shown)


tools/heroku.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+
+export PWD_HOLD="${PWD}"
```
uberhacker: No need to export if this is used only in this script.

tools/heroku.sh
```diff
@@ -0,0 +1,53 @@
...
+echo >> heroku.rb
+echo "  def self.install" >> heroku.rb
+echo "    system \"mkdir -p #{CREW_DEST_PREFIX}/share/heroku\"" >> heroku.rb
+echo "    system \"cp -r . #{CREW_DEST_PREFIX}/share/heroku\"" >> heroku.rb
+echo "    system \"mkdir -p #{CREW_DEST_PREFIX}/bin\"" >> heroku.rb
+echo "    system \"ln -s #{CREW_PREFIX}/share/heroku/bin/run #{CREW_DEST_PREFIX}/bin/heroku\"" >> heroku.rb
+echo "  end" >> heroku.rb
+echo "end" >> heroku.rb
+
+unset PWD_HOLD
```
uberhacker: No need if not exported.

tools/heroku.sh
JL2210: See below.

uberhacker:
This is a nice idea but what would make it even cooler is if you could create the binaries and upload to bintray.com automagically.  If you could do this, the entire build process could be automated.  It might be even better if the script was general purpose for any package.

tools/heroku.sh
JL2210: Also, it only affects newly formed child processes.

JL2210:
That would be extremely difficult. Right now, I don't see it being possible for any other package, and even if it was, it would take a lot of hand-written scripts. About the binaries, the script would require people to have 4 VMs with Chromium OS installed to work. I gather that that's much too involved for even the average maintainer.

JL2210:
The new scripts would also require a Git repository and marked releases to work. Not very portable, if I do say so myself.

tools/heroku.sh
uberhacker: Any variable you assign will be available in the script without the need to export.  Not sure what weird shenanigans you are talking about but every script I have ever written did not need exported variables.

JL2210: OK.

JL2210: I'm not sure about that anymore. I couldn't reproduce it today.

JL2210: Done.

tools/heroku.sh
JL2210: Done.

JL2210:
@uberhacker, this OK?

uberhacker:
I'm passing the baton to @skiman6010. :)

skiman6010:
Cool script! Thanks!

PR #2774, opened 2018-10-19
Add sg3_utils package

uberhacker:
The sg3_utils package contains utilities that send SCSI commands to devices. As well as devices on transports traditionally associated with SCSI (e.g. Fibre Channel (FCP), Serial Attached SCSI (SAS) and the SCSI Parallel Interface(SPI)) many other devices use SCSI command sets. ATAPI cd/dvd drives and SATA disks that connect via a translation layer or a bridge device are examples of devices that use SCSI command sets.  See http://sg.danny.cz/sg/sg3_utils.html.

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/sg3_utils.rb
+++ b/packages/sg3_utils.rb
@@ -1,0 +1,31 @@
+require 'package'
+
+class Sg3_utils < Package
+  description 'The sg3_utils package contains utilities that send SCSI commands to devices.'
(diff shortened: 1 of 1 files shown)


JL2210:
I just recently had trouble with the ebuilds for this package in Gentoo. The older version had basically masked the older version, and I had to forcefully remove the old version to upgrade.

PR #2775, opened 2018-10-19
Add the Netwide Assembler

JL2210:
The Netwide Assembler is an assembler and disassembler for the Intel x86 architecture. It can be used to write 16-bit, 32-bit and 64-bit programs. NASM is considered to be one of the most popular assemblers for Linux. NASM was originally written by Simon Tatham with assistance from Julian Hall.

Tested on x86_64

### DIFF
--- a/packages/nasm.rb
+++ b/packages/nasm.rb
@@ -1,9 +1,9 @@
 require 'package'
 
 class Nasm < Package
-  version '2.14rc16'
   description 'The Netwide Assembler'
   homepage 'https://www.nasm.us/'
+  version '2.14rc16'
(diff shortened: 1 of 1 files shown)


packages/nasm.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Nasm < Package
+  version '2.14rc16'
+  description 'The Netwide Assembler'
+  homepage 'https://www.nasm.us/'
```
uberhacker: I know it doesn't really matter from Chromebrew's perspective but can you please put the version below the homepage to keep it consistent for humans? :)

JL2210: OK.

JL2210:
@uberhacker: Is this OK?

PR #2784, opened 2018-10-24
Gcc fix

JL2210:
Removes gcc 4 entirely and creates dependency from gcc to gcc7.  GCC 4 is deprecated, and also depends on cloog, breaking some packages. This kills two birds with one stone. Thanks, @superloach, for finding this out.

### DIFF
--- a/packages/gcc.rb
+++ b/packages/gcc.rb
@@ -3,25 +3,10 @@ require 'package'
 class Gcc < Package
   description 'The GNU Compiler Collection includes front ends for C, C++, Objective-C, Fortran, Ada, and Go.'
   homepage 'https://www.gnu.org/software/gcc/'
-  version '4.9.4-1'
+  version '7.3.0'
 
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew-cross/gcc-4.9.4-1-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew-cross/gcc-4.9.4-1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew-cross/gcc-4.9.4-1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew-cross/gcc-4.9.4-1-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: 'ecb047a9f52ea6b24313d4329b00236b6880e12df684e52e34e9664127d447c6',
-     armv7l: 'ecb047a9f52ea6b24313d4329b00236b6880e12df684e52e34e9664127d447c6',
-       i686: '9a8b20b58a564c9557c74b2b2769ddc68c110211588356b71c533aac6d54431b',
-     x86_64: '1ae1f58ca4b9fee3b0f68856e5d49b2feeb90e259e8d885f339c9aa8091f27c8',
-  })
+  is_fake
 
-  depends_on 'binutils'
-  depends_on 'gmp'
-  depends_on 'mpfr'
-  depends_on 'mpc'
-  depends_on 'isl'
-  depends_on 'glibc'
+  depends_on 'gcc7'
+
 end


uberhacker:
Can we rename the package to gcc4 instead?  I would like to keep these older versions, just in case, and eventually give the user the option to install different gcc versions if they want.  Sometimes these older versions are needed, believe it or not.

superloach:
If we do this, would it be possible to have both `gcc7` and `gcc4` binaries to allow both to be installed at the same time, and symlink `gcc7 -> gcc`?

JL2210:
Nope. That's why I removed GCC 4 in this PR. It breaks GCC 7.

superloach:
What prevents us from keeping them separate? I may look into doing this myself, if you'd like.

JL2210:
One thing is that the maintainer of the GCC package removed the sources, and another is conflicting files. Two versions of GCC are NOT meant to go together.

JL2210:
Also, the fact that my Chromebook is crashing like Windows does right now.

superloach:
Ouch, that's no fun. Looks like keeping gcc4 may be a bit of a lost cause, @uberhacker.

uberhacker:
@JL2210: Not true.  There is no reason we cannot install both packages at the same time.  This is another reason I want the source code instead of just binaries.  Ideally, I'd like to rename this package and provide the source code just like we do in gcc7.  @superloach: That is exactly what I had envisioned to swap between versions.  The symlinks would be simple and easy and we could install side-by-side.

JL2210:
No, you couldn't. What's the point of keeping an old compiler that doesn't even work now? Also, why rewrite an old package when we can use a new one?

uberhacker:
The only reason it doesn't work now is we have symlinks all over the place.  We need to rethink this but I don't see why it couldn't be done with some work.

JL2210:
Fine. https://gcc.gnu.org/faq.html#multiple

JL2210:
For now, though, can this be merged? This is a pretty serious issue.

uberhacker:
@superloach: If you can find the source code for gcc4, that may be the solution to help install them side by side.

JL2210:
If we find any package that needs GCC 4, then we can bring this back up.

superloach:
Yeah, I'll try and find it. From that page JL linked, it looks like it'll be pretty easy to separate them. Just some fun sed escapes 😁

superloach:
Well would you look at that, gcc4 sources!
http://mirrors-usa.go-parts.com/gcc/releases/gcc-4.9.4/

uberhacker:
Can we rename this to gcc4 instead?  Ultimately, I'd like to get the source code and compile it correctly.

JL2210:
OK. But for now, unless we see any package that absolutely needs GCC 4 that this would break, can this be merged? I'm having some troubles.

superloach:
The source is taking FOREVER to download, I may have to find another mirror.

JL2210:
`ftp.gnu.org/gnu/gcc`?

uberhacker:
How dare you suggest an ftp site! :-P

superloach:
Hahaha, yeah - ftp sucks.

JL2210:
@uberhacker, let me clarify: `https://ftp.gnu.org/gnu/gcc/`

uberhacker:
OK, that is not so bad then.  If you want the fastest mirror, this is likely your best link: https://ftpmirror.gnu.org/gnu/gcc/.

JL2210:
Maybe. I thought that was random.

uberhacker:
According to https://www.gnu.org/prep/ftp.en.html: _"GNU software is available on ftp.gnu.org under the directory /gnu. Please try to use one of the many mirrors of our site listed below: the mirrors will give you faster response."_

JL2210:
And it reduces the load. Those guys are pretty smart. Tell you what, once I get my Chromebook working, I'll add GCC 6. It has an awesome Java compiler.

uberhacker:
If you can create a gcc6 package, that would be great.

JL2210:
Yeah. You know, even though you guys seem to still want it, I found a big issue with GCC 4: If you compile a c++ app with GCC 4, it won't work with GCC 7 libs. They've changed the symbol mangler since then.

uberhacker:
That makes sense.  Why would you want to do that anyway?  Just work in an isolated environment for each version and it's all good.

JL2210:
OK. I suppose that's right. I'm just saying that most people would need to install both to have old libraries written in c++ work.

JL2210:
I found a need for this package. Building GlibC.

Issue #2785, opened 2018-10-24
VLC Package


superloach:
VLC is a free and open source cross-platform multimedia player and framework that plays most multimedia files, and various streaming protocols.

Source: http://ftp.videolan.org/pub/videolan/vlc/3.0.4/vlc-3.0.4.tar.xz

**Dependencies:**
======
Sources: https://download.videolan.org/contrib/
- [x] [a52](http://liba52.sourceforge.net/) - #2786 
- [x] [aom](https://aomedia.googlesource.com/aom/)
- [x] [aribb24](https://github.com/nkoriyama/aribb24)
- [x] [aribb25](https://code.videolan.org/videolan/aribb25)
- [ ] [asdcplib](https://www.cinecert.com/asdcplib/)
- [ ] [breakpad](https://chromium.googlesource.com/breakpad/breakpad/)
- [x] [cddb](http://libcddb.sourceforge.net/)
- [x] [chromaprint](https://github.com/acoustid/chromaprint)
- [ ] [crystalhd](https://github.com/yeradis/crystalhd)
- [ ] [daala](https://xiph.org/daala/)
- [ ] [dvbpsi](https://www.videolan.org/developers/libdvbpsi.html)
- [ ] [dvdcss](https://www.videolan.org/developers/libdvdcss.html)
- [x] [dvdnav](https://www.videolan.org/developers/libdvdnav.html)
- [x] [dvdread](http://dvdnav.mplayerhq.hu/#download)
- [ ] [ebml](http://matroska-org.github.io/libebml/)
- [ ] [faad2](https://github.com/dsvensson/faad2)
- [ ] [fluidlite](https://github.com/divideconcept/FluidLite)
- [x] [glew](http://glew.sourceforge.net/)
- [ ] [gme](https://github.com/mcfiredrill/libgme)
- [ ] [goom](http://www.fovea.cc/?page=details_produits&id=goom)
- [ ] [gsm](http://www.quut.com/gsm/)
- [ ] [kate](https://wiki.xiph.org/OggKate)
- [x] [libbluray](https://www.videolan.org/developers/libbluray.html)
- [ ] [libbpg](https://bellard.org/bpg/)
- [x] [libdca](https://www.videolan.org/developers/libdca.html)
- [ ] [libfluid](http://opennetworkingfoundation.github.io/libfluid/)
- [ ] libmpeg2
- [ ] libplacebo
- [x] live555
- [ ] matroska
- [ ] mfx
- [ ] microdns
[...]

PR #2786, opened 2018-10-24
Add ragel and a52

superloach:
Working toward #2785

### DIFF
--- a/packages/ragel.rb
+++ b/packages/ragel.rb
@@ -1,10 +1,10 @@
 require 'package'
 
 class Ragel < Package
-  description 'ragel desc stub'
+  description 'Ragel compiles executable finite state machines from regular languages.'
   homepage 'http://www.colm.net/open-source/ragel/'
   version '6.10'
-  source_url 'http://www.colm.net/files/ragel/ragel-6.10.tar.gz'
+  source_url 'https://www.colm.net/files/ragel/ragel-6.10.tar.gz'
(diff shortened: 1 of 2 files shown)


superloach:
Forgot how branches work 🤦‍♀️
These 2 can just go together, haha

packages/ragel.rb
```diff
@@ -0,0 +1,18 @@
+require 'package'
+
+class Ragel < Package
+  description 'ragel desc stub'
```
uberhacker: Can we change this description?  Does `Ragel State Machine Compiler` make it more clear?

packages/ragel.rb
```diff
@@ -0,0 +1,18 @@
...
+  homepage 'http://www.colm.net/open-source/ragel/'
+  version '6.10'
+  source_url 'http://www.colm.net/files/ragel/ragel-6.10.tar.gz'
```
uberhacker: `https://www.colm.net/files/ragel/ragel-6.10.tar.gz` works so can we use that instead?

packages/ragel.rb
superloach: Dang it, forgot to change that. Fixing now.

packages/ragel.rb
superloach: Yes!

uberhacker:
Nice work @superloach!

PR #2790, opened 2018-10-26
pipes.sh and pipesX.sh

superloach:
Adds the pipes.sh terminal screensaver.

### DIFF
--- a/packages/pipes_sh.rb
+++ b/packages/pipes_sh.rb
@@ -1,0 +1,15 @@
+require 'package'
+
+class Pipes_sh < Package
+  description 'Animated pipes terminal screensaver'
+  homepage 'https://pipeseroni.github.io/'
+  version '1.3.0'
+  source_url 'https://github.com/pipeseroni/pipes.sh/archive/v1.3.0.tar.gz'
+  source_sha256 '532976dd8dc2d98330c45a8bcb6d7dc19e0b0e30bba8872dcce352361655a426'
+
+  depends_on 'ncurses'
+
+  def self.install
+    system "make", "PREFIX=#{CREW_DEST_DIR}#{CREW_PREFIX}", "install"
+  end
+end
--- a/packages/pipes_sh.rb
+++ b/packages/pipes_sh.rb
@@ -10,6 +10,6 @@ class Pipes_sh < Package
   depends_on 'ncurses'
 
   def self.install
-    system "make", "PREFIX=#{CREW_DEST_DIR}#{CREW_PREFIX}", "install"
+    system "make DESTDIR=#{CREW_DEST_DIR} PREFIX=#{CREW_PREFIX} install"
   end
 end
--- a/packages/pipesx_sh.rb
+++ b/packages/pipesx_sh.rb
@@ -1,0 +1,15 @@
+require 'package'
+
+class Pipesx_sh < Package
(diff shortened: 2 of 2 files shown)


JL2210:
Just a quick thought, I don't think that VLC will use this over the included one. Nice work, though!

superloach:
Haha, I'm just making some easy packages I've been thinking about for ages.

JL2210:
Lol. I did that once. Five random PRs in a day. I think they were `wireless_tools`, `firejail`, `hplip`, `checkinstall`, and `quakespasm`.

packages/pipes_sh.rb
```diff
@@ -0,0 +1,15 @@
...
+  description 'Animated pipes terminal screensaver'
+  homepage 'https://pipeseroni.github.io/'
+  version '1.3.0'
+  source_url 'https://github.com/pipeseroni/pipes.sh/archive/v1.3.0.tar.gz'
+  source_sha256 '532976dd8dc2d98330c45a8bcb6d7dc19e0b0e30bba8872dcce352361655a426'
+
+  depends_on 'ncurses'
+
+  def self.install
+    system "make", "PREFIX=#{CREW_DEST_DIR}#{CREW_PREFIX}", "install"
```
uberhacker: You can use `#{CREW_DEST_PREFIX}` instead of the two combined constants.

packages/pipesx_sh.rb
```diff
@@ -0,0 +1,15 @@
...
+  description 'Animated pipes terminal screensaver at an angle'
+  homepage 'https://pipeseroni.github.io/'
+  version '1.1.0'
+  source_url 'https://github.com/pipeseroni/pipesX.sh/archive/v1.1.0.tar.gz'
+  source_sha256 '82e60efbe3e8ae1c83a9b7e5df7d67ccdad3905dfd478f27d2b2be488603420c'
+
+  depends_on 'ncurses'
+
+  def self.install
+    system "make", "PREFIX=#{CREW_DEST_DIR}#{CREW_PREFIX}", "install"
```
uberhacker: Same here.  Only `#{CREW_DEST_PREFIX}` is needed.

packages/pipes_sh.rb
uberhacker: With that said, it might be better to use `DESTDIR=#{CREW_DEST_DIR}` and the normal `PREFIX=#{CREW_PREFIX}`, if possible.

uberhacker:
Great, just what we need; more screensavers to lock up our screen randomly when we need to start working. LoL

JL2210:
Awesome!

JL2210:
You know, that's one of the best parts about Windows. :rofl:

packages/pipes_sh.rb
superloach: So `make DESTDIR=#{CREW_DEST_DIR} PREFIX=#{CREW_PREFIX} install` for both packages?

superloach:
Haha, yup. Also, this runs in a terminal, only when you run it. No unexpected lock-ups there!

packages/pipes_sh.rb
uberhacker: If you can swing it, sure.

superloach:
Heck yeah >:3

PR #2791, opened 2018-10-26
Terminal fun! (cowsay, fortune, sl)

superloach:
Cowsay, fortune, lolcat, and sl!
===
- Cowsay is cowsay, I have no idea how we didn't have that yet.
- Not totally sure if lolcat is valid, I can't tell if `gem` is using the current folder or not.
- I did separate packages for `fortune` and `fortunes`, as it should be.
- I hope nobody is trying to do `crew install ls`

If anyone thinks of more fun terminal stuff, I can throw it into this PR.

_I am prepared for lots of criticism from the crew gods, aka Ed and James_ 🙃


packages/cowsay.rb
```diff
@@ -0,0 +1,13 @@
+require 'package'
+
+class Cowsay < Package
+  description 'cowsay is a program that generates ASCII pictures of a cow with a message.'
+  homepage 'https://en.wikipedia.org/wiki/Cowsay'
```
uberhacker: This is not really a home page, per say, but it does adequately describe cowsay.  Probably should use https://github.com/tnalpgge/rank-amateur-cowsay/ instead. :)

packages/cowsay.rb
```diff
@@ -0,0 +1,13 @@
...
+  version '3.04'
+  source_url 'https://github.com/tnalpgge/rank-amateur-cowsay/archive/cowsay-3.04.tar.gz'
+  source_sha256 'd8b871332cfc1f0b6c16832ecca413ca0ac14d58626491a6733829e3d655878b'
+
+  def self.install
+    system "sh install.sh #{CREW_PREFIX}"
```
uberhacker: Does `crew files cowsay` return any results?  I'm thinking not until you use `#{CREW_DEST_PREFIX}` instead.  BTW, `crew files ...` is a great way to test if the package installed correctly or not.

packages/fortunes.rb
```diff
@@ -0,0 +1,14 @@
+require 'package'
+
+class Fortunes < Package
+  description 'Over 15000 cookies for the fortune program.'
+  homepage 'https://packages.debian.org/sid/fortunes'
+  version '1.99.1'
+  source_url 'http://cdn-fastly.deb.debian.org/debian/pool/main/f/fortune-mod/fortune-mod_1.99.1.orig.tar.gz'
```
uberhacker: Debian really should get with the times and use https instead but this is what we're dealt and have to live with.

packages/lolcat.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class Lolcat < Package
+  description 'Rainbows and unicorns!'
+  homepage 'https://github.com/busyloop/lolcat'
+  version '99.9.19'
+  source_url 'https://github.com/busyloop/lolcat/archive/v99.9.19.tar.gz'
+  source_sha256 '2ec5beb8051576600adf3197b9826ce3319a18448bd0dbb8a92a028be8370449'
+
+  def self.install
+    system "gem install lolcat"
```
uberhacker: Probably want to install in `#{CREW_DEST_PREFIX}` somehow, otherwise, files aren't tracked and nothing shows up with `crew files lolcat`.

packages/lolcat.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def self.postinstall
+    system "echo spam blah blah blah>>/tmp/lol"
+    system "echo this is randomness>>/tmp/lol"
+    system "echo uwuwuwuwuwuwuwuwuwu>>/tmp/lol"
+    system "echo rainbows and unicorns!>>/tmp/lol"
+    system "echo cotton candy lol!!!1!>>/tmp/lol"
+    system "echo yay kittens =^~^=>>/tmp/lol"
+    system "lolcat /tmp/lol --freq=0.5"
+    system "rm /tmp/lol"
```
uberhacker: Cute! :)

packages/sl.rb
```diff
@@ -0,0 +1,22 @@
...
+  homepage 'https://github.com/mtoyoda/sl'
+  version '5.02'
+  source_url 'https://github.com/mtoyoda/sl/archive/5.02.tar.gz'
+  source_sha256 '1e5996757f879c81f202a18ad8e982195cf51c41727d3fea4af01fdcbbb5563a'
+
+  def self.build
+    text = File.read("sl.c")
+    new_contents = text.gsub(/curses\.h/, "#{CREW_PREFIX}/include/ncurses/curses.h")
+    File.open("sl.c.t", "w") {|file| file.puts new_contents }
+    system "mv sl.c.t sl.c"
```
uberhacker: Simplify?
```
system "sed -i 's,curses.h,#{CREW_PREFIX}/include/ncurses/curses.h,' sl.c"
```

packages/sl.rb
```diff
@@ -0,0 +1,22 @@
+require 'package'
+
+class Sl < Package
+  description 'Steam Locomotive runs across your terminal when you type "sl" as you meant to type "ls".'
```
uberhacker: I do this a LOT.  Lixing metters. :)

uberhacker:
@superloach:  These are really nice packages!  Should have installed them earlier but...you know.

packages/fortune.rb
```diff
@@ -0,0 +1,24 @@
...
+    text = File.read("fortune.c")
+    new_contents = text.gsub(/\/usr/, "#{CREW_PREFIX}")
+    File.open("fortune.c.t", "w") {|file| file.puts new_contents }
+    system "mv fortune.c.t fortune.c"
+    system "make"
+  end
+
+  def self.install
+    system "mkdir -pv #{CREW_DEST_PREFIX}/bin"
+    system "mv fortune #{CREW_DEST_PREFIX}/bin/"
```
uberhacker: You could replace these two lines with one:
```
system "install -Dm755 fortune #{CREW_DEST_PREFIX}/bin/"
```

packages/lolcat.rb
superloach: Would `gem install lolcat --install-dir #{CREW_DEST_LIB_PREFIX}/ruby/gems/*` work?

superloach: You know what, I think trying to make packages of gem packages just isn't worth it. Just let them use gem 😂
Perhaps someday we could add stuff to the `Package` class that would give us the option to "metapackage" from other package managers... not really worth it yet, though.

packages/fortune.rb
superloach: Did that for `sl` as well :)

superloach:
I've also made the amazing discovery that `pipes.sh|lolcat` and `sl|lolcat` work wonderfully.

uberhacker:
Yeah, installing packages from other package managers can be a bit tricky.  Another thing you need to figure out is how to lock in the version because a simple `gem install something` will always get the latest release.  I don't think packaging from other package managers is always a bad idea, however, especially if there is some complex configuration with the app.

packages/fortunes.rb
JL2210: @uberhacker, they have an HTTPS mirror. `https://deb.debian.org/` is the usual one. It redirects to the closest HTTPS mirror.

JL2210:
@uberhacker, this definitely needs to have the hacktoberfest label.


lol

PR #2792, opened 2018-10-26
Fix install for fortune and sl

uberhacker:
Fixes install: target '/usr/local/tmp/crew/dest/usr/local/bin/' is not a directory: No such file or directory
sl failed to install: `install -Dm755 sl /usr/local/tmp/crew/dest/usr/local/bin/` exited with 1

### DIFF
--- a/packages/fortune.rb
+++ b/packages/fortune.rb
@@ -15,6 +15,6 @@ class Fortune < Package
   end
 
   def self.install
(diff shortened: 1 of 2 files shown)


JL2210:
That's odd. The `D` option is supposed to create the directory that the file is to be installed to. Also, good news, my Chromebook works!

PR #2795, opened 2018-10-26
XScreenSaver

superloach:
Let's hope Jamie Zawinski doesn't have too much of a grudge against ChromeOS 🤣
Also, none of the GLX screensavers work.


superloach:
Pfft, forgot to take the `/pull/2795` off that URL.

PR #2796, opened 2018-10-27
fix(dpkg): 1.19.1 has been removed,so use 1.19.2 instead

fsx950223:
Fixes #
dpkg package not found

## Description
update package version

## Addtional information

Works properly:
- [x] x86_64

---

## That's it

### DIFF
--- a/packages/dpkg.rb
+++ b/packages/dpkg.rb
@@ -3,9 +3,9 @@ require 'package'
 class Dpkg < Package
   description 'A medium-level package manager for Debian'
   homepage 'https://anonscm.debian.org/git/dpkg/'
-  version '1.19.1'
-  source_url 'http://http.debian.net/debian/pool/main/d/dpkg/dpkg_1.19.1.tar.xz'
-  source_sha256 'ae3978a6b7bddc3e3196804ae0d49ea008c84a8a7a60b7d212af1e1d469e7ccf'
+  version '1.19.2'
+  source_url 'http://http.debian.net/debian/pool/main/d/dpkg/dpkg_1.19.2.tar.xz'
+  source_sha256 'f8f2ae2cf8065b81239db960b3794099ec607c94a125cec61c986f68f9861b71'
 
   depends_on 'bz2'
   depends_on 'xzutils'


packages/dpkg.rb
```diff
@@ -4,7 +4,7 @@ class Dpkg < Package
   description 'A medium-level package manager for Debian'
   homepage 'https://anonscm.debian.org/git/dpkg/'
   version '1.19.1'
```
uberhacker: This version should be bumped also.

packages/dpkg.rb
```diff
@@ -4,7 +4,7 @@ class Dpkg < Package
...
-  source_url 'http://http.debian.net/debian/pool/main/d/dpkg/dpkg_1.19.1.tar.xz'
+  source_url 'http://http.debian.net/debian/pool/main/d/dpkg/dpkg_1.19.2.tar.xz'
   source_sha256 'ae3978a6b7bddc3e3196804ae0d49ea008c84a8a7a60b7d212af1e1d469e7ccf'
```
uberhacker: This sha256 hash will change.

uberhacker:
Thank you for the PR @fsx950223!  Please see comments.

packages/dpkg.rb
uberhacker: Please change to:
```
source_url 'https://deb.debian.org/debian/pool/main/d/dpkg/dpkg_1.19.2.tar.xz'
source_sha256 'f8f2ae2cf8065b81239db960b3794099ec607c94a125cec61c986f68f9861b71'
```

packages/code.rb
```diff
@@ -25,7 +26,12 @@ class Code < Package
   depends_on 'libgconf'
   depends_on 'xdg_base'
   depends_on 'sommelier'
-
+  def self.preinstall
+    case ARCH
+    when 'armv7l'
+      system "dpkg", "-i", "vscode-1.28.0.deb"
+    end
+  end
```
uberhacker: Please include the change to code.rb in a separate PR.

fsx950223: ok

packages/code.rb
```diff
@@ -17,6 +17,7 @@ class Code < Package
   end
 
   binary_url ({
+    armv7l:'https://github.com/stevedesmond-ca/vscode-arm/releases/download/1.28.0/vscode-1.28.0.deb'
```
uberhacker: I don't think you will want to this with a .deb package.  This is reserved for Chromebrew packages only.

PR #2797, opened 2018-10-27
feat(code): add vscode-arm.deb

fsx950223:
Fixes #

add vscode-arm binary install package

## Description

## Addtional information

Works properly:
- [x] x86_64

---

## That's it

### DIFF
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -17,6 +17,7 @@ class Code < Package
   end
 
   binary_url ({
+    armv7l:'https://github.com/stevedesmond-ca/vscode-arm/releases/download/1.28.0/vscode-1.28.0.deb'
   })
   binary_sha256 ({
   })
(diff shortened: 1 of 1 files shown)


packages/code.rb
```diff
@@ -25,7 +26,12 @@ class Code < Package
   depends_on 'libgconf'
```
uberhacker: You will need `depends_on 'dpkg'` since this package uses that executable.

packages/code.rb
```diff
@@ -17,6 +17,7 @@ class Code < Package
   end
 
   binary_url ({
+    armv7l:'https://github.com/stevedesmond-ca/vscode-arm/releases/download/1.28.0/vscode-1.28.0.deb'
```
uberhacker: Are you sure this works?  This section expects a Chromebrew packaged archive.

packages/code.rb
```diff
@@ -25,7 +26,12 @@ class Code < Package
   depends_on 'libgconf'
   depends_on 'xdg_base'
   depends_on 'sommelier'
-
+  def self.preinstall
+    case ARCH
+    when 'armv7l'
+      system "dpkg", "-i", "vscode-1.28.0.deb"
```
uberhacker: This technically should go in the `def self.install` section.

fsx950223:
There is too much problems

PR #2801, opened 2018-10-27
Add x86_64 binaries for dpkg package

JL2210:
Tested on x86_64 :)


uberhacker:
If you can't create all the binaries, please do not submit a PR.

JL2210:
I can't create the ARM binaries, because I don't have an ARM Chromebook, and I can't create an i686 binary because my USB broke. Why?

uberhacker:
I want to include all the binaries or none.  I'll see if I can create binaries for all architectures.

uberhacker:
Just so you are not in the dark about me reasons for an all or nothing approach, I want to make sure if there are any issues building the binaries for a certain architecture that those are addressed in the source for future builds.

JL2210:
Oh, that's fine.

PR #2804, opened 2018-10-27
Fix i686 binary for nasm

uberhacker:
Tested on:
- [x] i686


JL2210:
What was wrong with the i686 binary for nasm?

JL2210:
Never mind. It's pretty obvious, now that I look at it.

PR #2805, opened 2018-10-27
Add libav package

uberhacker:
Libav provides cross-platform tools and libraries to convert, manipulate and stream a wide range of multimedia formats and protocols.  See https://www.libav.org/.
Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/libav.rb
+++ b/packages/libav.rb
@@ -1,0 +1,38 @@
+require 'package'
+
+class Libav < Package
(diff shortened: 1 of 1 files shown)


JL2210:
@uberhacker, shouldn't ffmpeg be used instead? As I understand, you can only use one or the other.

uberhacker:
Well, like most things in open source, there is more than one way of doing things.  With that said, this could potentially create a conflict if both packages are installed at the same time.

JL2210:
Exactly what I mean.

PR #2806, opened 2018-10-27
Rebuild alsa_plugins with libav support

uberhacker:
Depends on #2805.
Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64


JL2210:
Are you sure this enables libav? Are there any `--enable-libav` or `--enable-libffmpeg` options?

uberhacker:
It auto detects the libraries if they are available.

uberhacker:
I think I created a dependency loop with the pulseaudio and alsa_plugins packages.

JL2210:
Great. lol. Maybe make an `alsa_plugins_sub` package like you do with Freetype?

JL2210:
@uberhacker, did this work? What about the dependency loop?

PR #2807, opened 2018-10-27
Add tcpwrappers package

uberhacker:
The tcpwrappers package provides daemon wrapper programs that report the name of the client requesting network services and the requested service.  See http://www.linuxfromscratch.org/blfs/view/6.1/basicnet/tcpwrappers.html.

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/tcpwrappers.rb
+++ b/packages/tcpwrappers.rb
@@ -1,0 +1,51 @@
+require 'package'
+
+class Tcpwrappers < Package
+  description 'The tcpwrappers package provides daemon wrapper programs that report the name of the client requesting network services and the requested service.'
+  homepage 'http://www.linuxfromscratch.org/blfs/view/6.1/basicnet/tcpwrappers.html'
+  version '7.6'
+  source_url 'ftp://ftp.porcupine.org/pub/security/tcp_wrappers_7.6.tar.gz'
+  source_sha256 '9543d7adedf78a6de0b221ccbbd1952e08b5138717f4ade814039bb489a4315d'
+
+  binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/tcpwrappers-7.6-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/tcpwrappers-7.6-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


JL2210:
FTP? You know better...

You also might use a newer patch than LFS 6.1.

uberhacker:
What else can you do when that is the only option available that works?

uberhacker:
This release is already over 20 years old.  I don't think newer patches will make that much difference but who knows?

JL2210:
Fair enough. You could still use a newer patch (8.2 is out now.)

uberhacker:
Right, but does 8.2 have build instructions available for this package?

JL2210:
http://www.linuxfromscratch.org/blfs/downloads/6.3/patches/tcp_wrappers-7.6-shared_lib_plus_plus-1.patch

JL2210:
No, but 6.3 does.

JL2210:
http://files.ichilton.co.uk/nfs/tcp_wrappers_7.6.tar.gz

uberhacker:
It's the same exact patch. :)

JL2210:
Yes, but the source is HTTP. Better than nothing. Also, you never know. They could have changed it.

uberhacker:
Have you tried that source?  It's a dead link.

JL2210:
We'd better use the 6.3 one, anyway. They might remove 6.1

uberhacker:
I'll let you change it.  Not worth it.

JL2210:
I'll host the tarball on GitHub, then. I don't want to have to deal with FTP.

uberhacker:
You won't need to deal with ftp once the binaries are available unless you like pain and always want to compile everything from source.

JL2210:
Who says that's pain?

uberhacker:
Only you would say that. :-P  Apparently, you have all the time in the world to do all this compiling crap. :)

JL2210:
https://github.com/JL2210/source-code/raw/master/tcp_wrappers_7.6.tar.gz

JL2210:
I love fixing bad designs in C, build systems, and otherwise.

JL2210:
@uberhacker, did you use my source URL?

PR #2810, opened 2018-10-28
Fix pulseaudio to avoid dependency loop

uberhacker:
This fixes the dependency loop issue @JL2210.

### DIFF
--- a/packages/pulseaudio.rb
+++ b/packages/pulseaudio.rb
@@ -20,7 +20,7 @@ class Pulseaudio < Package
      x86_64: '67608beaffd35ca0f5a3c35769797d2eae4193d5c81c8cad5e6e123f2d667e44',
   })
 
-  depends_on 'alsa_plugins'
+  depends_on 'alsa_plugins' => :build
   depends_on 'dbus'
   depends_on 'intltool'
   depends_on 'libsndfile'


JL2210:
Thanks :)

JL2210:
Did you check if there were bindings to alsa_plugins in pulseaudio? Should you rebuild the binaries?

uberhacker:
I think it's fine for now.  Can you test if pulseaudio kills your sound again?

JL2210:
It's working now. I think it was either the codecs that were messing with my sound, or the fact that my Chromebook was so full of junk that it needed a Powerwash.

uberhacker:
After taking out the dependency, the build no longer has alsa bindings so I just made it a build dependency in order to prevent the dependency loop.  This should be ready to merge @skiman6010.

JL2210:
Wouldn't that still cause a dependency loop if you're trying to build from source?

uberhacker:
It builds fine after fresh Chromebrew install.  No dependency looping.

JL2210:
Also, a binary is still linked to a library whether the library is there or not, so... rebuild.
Take this for example:
```
chronos@localhost ~/Downloads $ cat a.c
extern int lol(int argc, char **argv);

int main(int argc, char **argv)
{
        lol(argc,argv);
        return 0;
}
chronos@localhost ~/Downloads $ cat liba.c
#include <stdio.h>

int lol(int argc, char **argv)
{
        printf("%s", argv[1]);
        return 0;
}
chronos@localhost ~/Downloads $ ldd a 
        linux-vdso.so.1 (0x00007ffe539c8000)
        liba.so => ***not found***
        libc.so.6 => /usr/local/lib64/libc.so.6 (0x00007c861b5e7000)
        /lib64/ld-linux-x86-64.so.2 (0x00007c861b992000)
chronos@localhost ~/Downloads $ ldd liba.so
        linux-vdso.so.1 (0x00007ffd0bfd9000)
        libc.so.6 => /usr/local/lib64/libc.so.6 (0x00007fcac48a3000)
        /usr/local/lib64/ld-linux-x86-64.so.2 (0x00007fcac4c4e000)
chronos@localhost ~/Downloads $ ./a
./a: error while loading shared libraries: liba.so: cannot open shared object file: No such file or directory
```

uberhacker:
No need to rebuild.  The binary will be the same.

JL2210:
OK.

PR #2821, opened 2018-11-01
Update yarn from 1.12.0 to 1.12.1

uberhacker:


JL2210:
No binaries?

uberhacker:
Not needed.  It takes seconds to install.  Amazing that an app this powerful is this simple.

JL2210:
I could see that. Meson, with what an incredible build system it is, takes less than a minute to compile and install, even though it has to precompile the .py files into Python bytecode!

skiman6010:
Meson is incredible.

PR #2825, opened 2018-11-02
Update Python3

JL2210:
Update `python3` package.

Tested on:
- [x] x86_64

Adds more dependencies because of missing shared libraries required in python3 3.7.1

### DIFF
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -3,26 +3,22 @@ require 'package'
 class Python3 < Package
   description 'Python is a programming language that lets you work quickly and integrate systems more effectively.'
   homepage 'https://www.python.org/'
-  version '3.7.0'
-  source_url 'https://www.python.org/ftp/python/3.7.0/Python-3.7.0.tar.xz'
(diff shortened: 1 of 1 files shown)


packages/python3.rb
```diff
@@ -3,26 +3,22 @@
...
-     x86_64: '7c4e07989669cac3233fe2b10d2447208a96c7d2056c20731f46aed90a83622b',
+     x86_64: '3e414c4a01b4f47c75f900435a29b5325900975f525ae905a5d1d073cdb8f332',
   })
 
-  depends_on 'bz2' => :build
-  depends_on 'xzutils' => :build
-  depends_on 'sqlite' => :build
+  depends_on 'bz2'
+  depends_on 'xzutils'
+  depends_on 'sqlite'
```
uberhacker: Why are these runtime dependencies now?

JL2210: The lib-dynload modules are linked to them and require some of their API calls.

PR #2826, opened 2018-11-03
Update fskit package

JL2210:
Update fskit package for compatibility with attr 2.4.48

Tested on:
- [x] x86_64

### DIFF
--- a/packages/fskit.rb
+++ b/packages/fskit.rb
@@ -3,38 +3,36 @@ require 'package'
 class Fskit < Package
   description 'Filesystem utility library and SDK'
   homepage 'https://github.com/jcnelson/fskit'
-  version 'dd9a8'
+  version 'dd9a8d'
   source_url 'https://github.com/jcnelson/fskit/archive/dd9a8dc844b81afc63e5067397786a32ecb66008.tar.gz'
   source_sha256 'fe682890ebab9226d65fc6ebfb8b3619c0d5a93e3161787cea9d01ad23d3a83a'
 
   binary_url ({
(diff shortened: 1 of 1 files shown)


packages/fskit.rb
```diff
@@ -3,38 +3,36 @@
...
 
   depends_on 'attr'
 
   def self.patch
-    # for x86_64 architecture, build library should be modified to #{CREW_PREFIX}/lib64
-    if ARCH == "x86_64"
-      puts "Patch build file for x86_64 ..."
-      `sed -i -- 's/lib/lib64/g' buildconf.mk` # replace lib by lib64
-    end
+    system "wget", "https://github.com/JL2210/patches/raw/master/fskit-dd9a8-xattr.patch"
```
uberhacker: Should we add a sha256 checksum on the patch?

JL2210: No. Right now, I'm planning on improving it. It's kind of crude right now. It just totally removes the header file and replaces it with two others.

uberhacker: That doesn't give me much confidence.  We should probably add the sha256 checksum just in case.

JL2210: For now, I'll use a hashed version and add a sha256sum.

JL2210: BTW, patch fails with something along the lines of `patch: **** Only garbage was found in the patch input.` if the patch is corrupted.

uberhacker: Well, with a sha256 checksum, at least they would get a message that tells them to try again.  I'm afraid they may just think the patch is corrupt if it doesn't download cleanly.

uberhacker: Anyway, with binaries we don't have to worry about any of these edge cases.

JL2210: Done.

PR #2827, opened 2018-11-03
Update Vdev package

JL2210:
Depends on #2826 

Updates fskit package for compatibility with attr 2.4.48
Uses proper lib directory `#{CREW_LIB_PREFIX}`

Tested on:
- [x] x86_64

### DIFF
--- a/packages/vdev.rb
+++ b/packages/vdev.rb
@@ -25,20 +25,20 @@ class Vdev < Package
   end
 
   def self.build
-    system "make LIBDIR=#{CREW_LIB_PREFIX}"
-    system "make LIBDIR=#{CREW_LIB_PREFIX} -C vdevd OS=LINUX"
-    system "make LIBDIR=#{CREW_LIB_PREFIX} -C example"
-    system "make LIBDIR=#{CREW_LIB_PREFIX} -C hwdb"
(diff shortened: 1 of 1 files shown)


packages/vdev.rb
```diff
@@ -3,44 +3,42 @@
...
-    system "make -C example"
-    system "make -C hwdb"
-    system "make -C libudev-compat"
-    system "make -C fs"
+    system "make LIBDIR=#{CREW_LIB_PREFIX}"
+    system "make LIBDIR=#{CREW_LIB_PREFIX} -C vdevd OS=LINUX"
+    system "make LIBDIR=#{CREW_LIB_PREFIX} -C example"
+    system "make LIBDIR=#{CREW_LIB_PREFIX} -C hwdb"
+    system "make LIBDIR=#{CREW_LIB_PREFIX} -C libudev-compat"
+    system "make LIBDIR=#{CREW_LIB_PREFIX} -C fs"
```
uberhacker: I assume this means all the files are installed under `#{CREW_LIB_PREFIX}`?  If anything isn't, we would need `PREFIX=#{CREW_PREFIX}`, right?

JL2210: I suppose so. Let me try that rebasing URL you gave me.

JL2210:
@uberhacker Ready?

PR #2840, opened 2018-11-06
Update Apache Ant

### LINKED ISSUE #2839: Unable to install `ant` package from "source"
Log:
```
chronos@localhost / $ crew install -s ant
ant: Apache Ant is a Java library and command-line tool whose mission is to drive processes described in build files as targets and extension points dependent upon each other.
http://ant.apache.org/
version 1.10.2
Downloading source...
Checksum mismatch. :/ Try again.
```

JL2210:
Fixes #2839 
Fixes Ant on x86_64

Update Apache Ant package to 1.10.5

Tested on:
- [x] x86_64

### DIFF
--- a/packages/ant.rb
+++ b/packages/ant.rb
@@ -3,42 +3,40 @@ require 'package'
 class Ant < Package
   description 'Apache Ant is a Java library and command-line tool whose mission is to drive processes described in build files as targets and extension points dependent upon each other.'
   homepage 'http://ant.apache.org/'
-  version '1.10.2'
-  source_url 'https://www.apache.org/dist/ant/binaries/apache-ant-1.10.2-bin.tar.xz'
-  source_sha256 '361c8ad2ed8341416e323e7c28af10a8297170a80fdffba294a5c2031527bb6c'
+  version '1.10.5'
+  source_url 'http://www.apache.org/dist/ant/source/apache-ant-1.10.5-src.tar.xz'
+  source_sha256 'ec7c704e22a2b0167e1ff35cc3d4198267e6489f142aa3e9da1b7fe6e39ff53d'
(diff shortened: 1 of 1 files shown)


packages/ant.rb
```diff
@@ -3,42 +3,40 @@
...
-    system "mkdir -p #{CREW_DEST_LIB_PREFIX}"
-    system "cp -r bin/ #{CREW_DEST_PREFIX}"
-    system "cp -r lib/* #{CREW_DEST_LIB_PREFIX}"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/jdk8/"
+    system "for dist in lib bin; do \
+              cp -a dist/$dist/ #{CREW_DEST_PREFIX}/share/jdk8/; \
+            done"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin/"
+    system "for bin in ant antRun antRun.pl complete-ant-cmd.pl runant.pl runant.py; do \
+              ln -s ../share/jdk8/bin/$bin #{CREW_DEST_PREFIX}/bin/; \
```
uberhacker: This is not the correct symlink path.  You probably want something like:
```
ln -s #{CREW_PREFIX}/share/jdk8/bin/$bin #{CREW_DEST_PREFIX}/bin/;
```
What happens when you try to run any of these executables?  I imagine they don't work.  Remember, test, test, test.

JL2210: Nope. They do. I say `Tested on x86_64` for a reason. Symlink paths are relative to the file that they are in.

uberhacker: OK, I think I understand what is happening here.  Sorry, long day.

PR #2845, opened 2018-11-06
Add libjpeg dependency to emacs

kaiminglee:
## Description
`emacs` requires `libjpeg.so.9` provided by libjpeg, which is not installed by default. This PR adds `libjpeg` as a dependency.


uberhacker:
Thank you for the PR @kaiminglee!

PR #2848, opened 2018-11-07
Add xz_java package

JL2210:
This aims to be a complete implementation of XZ data compression
    in pure Java. Features:
      - Full support for the .xz file format specification version 1.0.4
      - Single-threaded streamed compression and decompression
      - Single-threaded decompression with limited random access support
      - Raw streams (no .xz headers) for advanced users, including LZMA2
        with preset dictionary

    Threading is planned but it is unknown when it will be implemented.

Tested on:
- [x] x86_64

### DIFF
--- a/packages/xz_java.rb
+++ b/packages/xz_java.rb
@@ -1,0 +1,35 @@
+require 'package'
+
+class Xz_java < Package
+  description 'A complete implementation of XZ data de/compression in pure Java.'
+  homepage 'http://tukaani.org/xz/java.html'
+  version '1.8'
+  source_url 'https://tukaani.org/xz/xz-java-1.8.zip'
+  source_sha256 '86be97adf59d19c000c5bac8c77f2fa2f84fa082d608c7ad2d07a007dce76160'
+
+  binary_url ({
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/xz_java-1.8-chromeos-x86_64.tar.xz',
+  })
+  binary_sha256 ({
+     x86_64: '911b67621907a1fca84fbdf3fce3cb3910a508bdf0b73731e0bf9c9a76920c29',
+  })
+
+  depends_on 'jdk8'
+  depends_on 'ant' => :build
+
+  def self.build
+    system "env", "JAVA_HOME=#{CREW_PREFIX}/share/jdk8", "ant"
+    system "env", "JAVA_HOME=#{CREW_PREFIX}/share/jdk8", "ant", "doc"
+    system "wget", "https://github.com/JL2210/bash-scripts/raw/94fb9fe9a69610c36de0978a1232769bb86b01a6/xz-java.in"
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("xz-java.in") ) == '077d39e511a896f9d9db8328edbf7254c9f8c1746f1d114863422005693835e2'
+    system "sed 's,@PREFIX@,#{CREW_PREFIX},g' xz-java.in > xz-java"
+  end
+
+  def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/doc/xz-java-1.8/"
+    system "cp", "-a", "build/doc/.", "#{CREW_DEST_PREFIX}/share/doc/xz-java-1.8/"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/jdk8/"
+    system "cp", "-a", "build/jar/.", "#{CREW_DEST_PREFIX}/share/jdk8/lib/"
+    system "install", "-Dm755", "xz-java", "#{CREW_DEST_PREFIX}/bin/xz-java"
+  end
+end


packages/xz_java.rb
```diff
@@ -0,0 +1,35 @@
...
+     x86_64: 'c8059feaa552660439ef17bb64a44fc096d9555eb06be8748cebf8ea2a7916dc',
+  })
+
+  depends_on 'jdk8'
+  depends_on 'ant' => :build
+
+  def self.build
+    system "JAVA_HOME=#{CREW_PREFIX}/share/jdk8", "ant"
+    system "JAVA_HOME=#{CREW_PREFIX}/share/jdk8", "ant", "doc"
+    system "wget", "https://github.com/JL2210/bash-scripts/raw/f786f858faae18c80e8c284ad4fc9e25cd14be36/xz-java.in"
```
uberhacker: Why not just call this xz-java without the .in extension?  Also, I would prefer you provide the source for this script in this package instead of on your personal Github account.  It makes it easier to review and change (ie. more open).  There are a few things about your script I think can be improved.  Generally, most compression programs use `-x` instead of `-d` to extract/decompress.  Also, you could add `--compress` and `--decompress` long switch options to your help section.

uberhacker:
Thank you for the new package @JL2210.

packages/xz_java.rb
JL2210: I'm trying to be as `xz` compliant as possible. `xz` has the -d option to decompress. It does have the long options `--compress` and `--decompress`. The file is too large to include in this package file. It installs as `xz-java`. It just has to configure the script with the prefix first.

uberhacker: I'm talking about the help section, not the parsing logic.

packages/xz_java.rb
```diff
@@ -0,0 +1,35 @@
...
+     x86_64: '5015ecd8bee78fda874c6b6cdf4576dcc3a8effe95c545e6841d2d041d916f16',
+  })
+
+  depends_on 'jdk8'
+  depends_on 'ant' => :build
+
+  def self.build
+    system "env", "JAVA_HOME=#{CREW_PREFIX}/share/jdk8", "ant"
+    system "env", "JAVA_HOME=#{CREW_PREFIX}/share/jdk8", "ant", "doc"
+    system "wget", "https://github.com/JL2210/bash-scripts/raw/c28abd3b80716637f708300e4085ed775b525844/xz-java.in"
```
uberhacker: Can you please rename this file with a `.sh` extension?  The `.in` does not indicate the type of file.

packages/xz_java.rb
```diff
@@ -0,0 +1,35 @@
...
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("xz-java.in") ) == '4d07e68c90241b57bff417d77f1faa915774c1e65cc4f4227d015604d9ffec41'
+    system "sed 's,@PREFIX@,#{CREW_PREFIX},g' xz-java.in > xz-java"
+  end
+
+  def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/doc/xz-java-1.8/"
+    system "cp", "-a", "build/doc/.", "#{CREW_DEST_PREFIX}/share/doc/xz-java-1.8/"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/jdk8/"
+    system "cp", "-a", "build/jar/.", "#{CREW_DEST_PREFIX}/share/jdk8/lib/"
+    system "install", "-Dm755", "xz-java", "#{CREW_DEST_PREFIX}/bin/xz-java"
```
uberhacker: You could do this:
```
system "install", "-Dm755", "xz-java.sh", "#{CREW_DEST_PREFIX}/bin/xz-java"
```
No need to keep renaming the same file.

JL2210:
Done.

packages/xz_java.rb
uberhacker: You start out with xz-java.in, then xz-java.  You could just use xz-java.sh and be done.

packages/xz_java.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  depends_on 'jdk8'
+  depends_on 'ant' => :build
+
+  def self.build
+    system "env", "JAVA_HOME=#{CREW_PREFIX}/share/jdk8", "ant"
+    system "env", "JAVA_HOME=#{CREW_PREFIX}/share/jdk8", "ant", "doc"
+    system "wget", "https://github.com/JL2210/bash-scripts/raw/c28abd3b80716637f708300e4085ed775b525844/xz-java.in"
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("xz-java.in") ) == '4d07e68c90241b57bff417d77f1faa915774c1e65cc4f4227d015604d9ffec41'
+    system "sed 's,@PREFIX@,#{CREW_PREFIX},g' xz-java.in > xz-java"
```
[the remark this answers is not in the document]
uberhacker: Why does that matter?  It's still a bash shell script with placeholders.

PR #2856, opened 2018-11-11
Update openssl from 1.0.2o to 1.0.2p

uberhacker:
Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64


JL2210:
Eventually, we'll need to upgrade to 1.1.1. Can we start making preparations now?

uberhacker:
Only if it doesn't break other packages.  There are A LOT of packages that have this as a dependency.

PR #2870, opened 2018-11-13
Change xdg_base package source URL

JL2210:
Changes source URL for xdg_base to `/dev/null` because current source URL is blocked on some networks.

### DIFF
--- a/packages/xdg_base.rb
+++ b/packages/xdg_base.rb
@@ -4,8 +4,8 @@ class Xdg_base < Package
   description 'XDG Base Directory Specification Configuration'
   homepage 'https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html'
   version '0.7-1'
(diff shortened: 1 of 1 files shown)


uberhacker:
If this source url is blocked on some networks, then I'm sure other source urls are blocked as well.  I would rather not change this for special cases.

uberhacker:
I did.  I don't think this is necessary.  The real issue is the network access.  That url should not be blocked.  Maybe port 443 is blocked but that should never happen on a properly configured network.

JL2210:
It's a school network. Supposedly, `Website contains prohibited Forums content.`

JL2210:
I can't install `vim` now, and my phone isn't working, so I can't install it at home.

JL2210:
Even so, we don't use the file in any way.

JL2210:
Absolutely. It's exactly the same. Any empty file has that checksum.

PR #2884, opened 2018-11-18
Update scite.rb scite icons broken

sajjaadf:
Scite Icons are broken. see issue skycocker#2879

### DIFF
--- a/packages/scite.rb
+++ b/packages/scite.rb
@@ -21,10 +21,11 @@ class Scite < Package
   })
 
   depends_on 'gtk3'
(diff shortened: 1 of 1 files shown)


uberhacker:
Thank you for the update @sajjaadf!  Can you please move the `depends_on 'gnome_icon_theme'` above `depends_on 'sommelier'`?  The reason is we need the post install message from sommelier last so it doesn't scroll past the active screen.

sajjaadf:
Okay-Done!

PR #2887, opened 2018-11-20
Add redef.rb and bump Chromebrew version

JL2210:
Adds redef.rb to allow redefining Chromebrew constants and bumps Chromebrew version. Will push a use-case PR soon.

Tested on x86_64


uberhacker:
I assume you think this is necessary because you couldn't set `CREW_NOT_STRIP`?  If you set an environment variable, you can override the default value.  I guess my question is why is this needed?  And why would you need a separate PR to explain a use case?

JL2210:
It's not really a use case, just a package that uses this. Setting the environment variable is useless, because it is only registered before `require 'package'` This could also theoretically be useful for changing the default CREW_PREFIX in packages.

JL2210:
Anything wrong with the Ruby code?

PR #2891, opened 2018-11-26
Add filezilla package and dependencies

uberhacker:
FileZilla Client is a free FTP solution and not only supports FTP, but also FTP over TLS (FTPS) and SFTP.  It is open source software distributed free of charge under the terms of the GNU General Public License.  See https://filezilla-project.org/.  Dependencies include libfilezilla, wxwidgets and xdg_utils.

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

Caveats:
- The i686 build of filezilla does not launch.  This has been a general issue with most of the gui apps tested lately which makes it increasingly difficult to support this architecture moving forward.  Most command line apps still work well, however.  We should probably think about supporting only cli apps on i686.
- The man pages for xdg_utils are useless but this doesn't prevent filezilla from working since the executables are fine.  The i686 man pages work fine, however.  Go figure?

### DIFF
--- a/packages/filezilla.rb
+++ b/packages/filezilla.rb
@@ -1,0 +1,45 @@
+require 'package'
+
+class Filezilla < Package
+  description 'FileZilla Client is a free FTP solution.'
+  homepage 'https://filezilla-project.org/'
+  version '1.6.3'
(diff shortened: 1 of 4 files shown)


skiman6010:
Neat!

PR #2908, opened 2018-12-03
Improve ld_default package

JL2210:
Can tell if `ld` is a symlink.
Allows for installing with no Internet (By changing source URL to `/dev/null`)
Works in a tmpdir
Brackets environment variables.

### DIFF
--- a/packages/ld_default.rb
+++ b/packages/ld_default.rb
@@ -2,54 +2,70 @@ require 'package'
 
 class Ld_default < Package
   description 'Select the default ld executable and check for libraries in #{CREW_LIB_PREFIX} first'
-  homepage 'https://github.com/skycocker/chromebrew'
-  version '1.0'
-  source_url 'https://raw.githubusercontent.com/skycocker/chromebrew/93aa4dc4d9198e28d84dcf9d59b3ab6b6a94d92c/README.md'
-  source_sha256 '3ec9c6f7ba6a138d07dfb4a372d562cd3ad558b58b6bd447c2dbb52e16b697a5'
+  homepage 'https://github.com/skycocker/chromebrew/wiki/FAQ'
+  version '1.1'
+  source_url 'file:///dev/null'
+  source_sha256 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
 
   depends_on 'llvm'
 
   def self.build
-system "cat << 'EOF' > ld_default
+    system "cat << 'EOF' > ld_default
 #!/bin/bash
-cd #{CREW_PREFIX}/bin
-type=\$(file ld | cut -d' ' -f2 | tr -d '\\n')
-if [ \"\$type\" == \"ELF\" ]; then
-  current=\$(find . -inum $(ls -i ld 2> /dev/null | cut -d' ' -f1) 2> /dev/null | fgrep 'ld.' 2> /dev/null | cut -d/ -f2 2> /dev/null)
-else
-  slashes=\$(tr -dc '/' <<<'#{CREW_PREFIX}/bin/ld/' | wc -c)
-  current=\$(tail -1 ld | cut -d'/' -f\${slashes} 2> /dev/null | cut -d' ' -f1 2> /dev/null)
-fi
-echo
-echo \"Current default: \$current\"
-echo
-echo 'Enter the ld default:'
-echo 'b = ld.bfd'
-echo 'g = ld.gold'
-echo 'l = ld.lld'
-echo '0 = Cancel'
-while true; do
-  read default
-  case $default in
-    b)
-      current='ld.bfd'
-      break;;
-    g)
-      current='ld.gold'
-      break;;
-    l)
-      current='ld.lld'
-      break;;
-    0)
-      exit;;
-    *)
-      echo 'Please select from one of the options or enter 0 to cancel.'
-  esac
-done
-echo '#!/bin/bash' > /tmp/ld
-echo '#{CREW_PREFIX}/bin/'$current' --library-path #{CREW_LIB_PREFIX} -rpath #{CREW_LIB_PREFIX} \"\$@\"' >> /tmp/ld
-install -Dm755 /tmp/ld #{CREW_PREFIX}/bin/ld
-rm -f /tmp/ld
+
+  set -e
+
+  mktmp=\$(mktemp -d)
+
+  cd #{CREW_PREFIX}/bin
+
+  type=\$(file ld | cut -d' ' -f2 | tr -d '\\n')
+
+  if [ \"\${type}\" == \"ELF\" ]; then
+    current=\$(find . -inum \$(ls -i ld | cut -d' ' -f1) | fgrep 'ld.')
+  elif [ \"${type}\" == \"symbolic\" ]; then	
+    current=\$(basename \$(readlink ld))
+  else
+    current=\$(basename \$(tail -1 ld | cut -d' ' -f1))
+  fi
+
+  echo
+  echo \"  Current default linker: \${current}\"
+  echo
+  echo '  Enter the new default linker:'
+  echo
+  echo '  b = ld.bfd'
+  echo '  g = ld.gold'
+  echo '  l = ld.lld'
+  echo '  0 = Cancel'
+  echo
+
+  while true; do
+    read default
+    case \${default} in
+      b)
+        new='ld.bfd'
+        break;;
+      g)
+        new='ld.gold'
+        break;;
+      l)
+        new='ld.lld'
+        break;;
+      0)
+        exit;;
+      *)
+        echo 'Please select from one of the options or enter 0 to cancel.'
+        echo
+    esac
+  done
+
+  echo '#!/bin/bash' > \${mktmp}/ld
+  echo \"\${new}\"' --library-path #{CREW_LIB_PREFIX} -rpath #{CREW_LIB_PREFIX} \"\$@\"' >> \${mktmp}/ld
+
+  install -Dm755 \${mktmp}/ld #{CREW_PREFIX}/bin/ld
+
+  rm -rf \${mktmp}
 EOF"
   end
 


uberhacker:
This is not needed.  The else condition works whether the file is a symbolic link or not.  Why make it more complicated?  If you don't have an Internet connection, you shouldn't be attempting to install Chromebrew packages anyway. :)

JL2210:
Yes, they are functional. It can read symlinks to tell you your default `ld`. It works in a dedicated temporary directory that changes every time that you use it. It doesn't hardcode the path of the default linker. It brackets environment variables, which should always be done. Anyway, are cosmetic improvements bad?

JL2210:
It also doesn't hardcode the path to `/tmp`.

JL2210:
@uberhacker, aren't any improvements supposed to be welcome? The package doesn't use the Chromebrew page, why download it?

packages/ld_default.rb
```diff
@@ -2,54 +2,70 @@
...
+  cd #{CREW_PREFIX}/bin
+
+  type=\$(file ld | cut -d' ' -f2 | tr -d '\\n')
+
+  if [ \"\${type}\" == \"ELF\" ]; then
+    current=\$(find . -inum $(ls -i ld 2>/dev/null | cut -d' ' -f1) 2>/dev/null | fgrep 'ld.' 2>/dev/null)
+  elif [ \"${type}\" == \"symbolic\" ]; then
+    current=\$(basename `readlink ld`)
+  else
+    current=\$(tail -1 ld | cut -d' ' -f1 2>/dev/null)
```
uberhacker: There are 2 things that can make this better:
```
if [ \"\${type}\" == \"ELF\" ]; then
     current=\$(find . -inum $(ls -i ld 2>/dev/null | cut -d' ' -f1) 2>/dev/null | fgrep 'ld.' 2>/dev/null)
   elif [ \"${type}\" == \"symbolic\" ]; then
     current=\$(readlink ld)
   else
     current=\$(basename \$(tail -1 ld | cut -d' ' -f1))
   fi
```
The readlink command already strips the full path.  The symlink scenario will only happen if it is created manually.  If somebody already knows how to do that, they don't need this script.

uberhacker:
OK, I'll give you credit for the changes.  Please see the comment.

JL2210:
Thank you. :)

packages/ld_default.rb
JL2210:
```
chronos@localhost ~/Downloads $ ln -s /usr/local/bin/hello file
chronos@localhost ~/Downloads $ ls -la file
lrwxrwxrwx. 1 chronos chronos 22 Dec  3 14:58 file -> /usr/local/bin/hello
chronos@localhost ~/Downloads $ readlink file
/usr/local/bin/hello
chronos@localhost ~/Downloads $
```
True.

JL2210: What do you want me to change?

uberhacker: Your readlink works differently than mine.  OK, I guess leave basename but please use $(...) instead of `...`.  Backticks is bad form.

JL2210: Oh. So you do want to keep the symlink functionality or you don't?

uberhacker: Yes, we can keep it but it will never happen to anybody except yourself. :)

JL2210: I re-added it with `$()` instead of backticks. Is it good now?

uberhacker: Still need `basename` on the last conditional.

uberhacker: You can also remove all the `2>/dev/null`s.  I cannot think of a situation where it would error.

JL2210: I can. If 'ld' is not there or if it's a broken link. If that errors, then `find` will fail because it's not trying to find anything. If that fails, `fgrep` will go forever, as it's trying to read broken input.

JL2210: This looks good right now. After this is merged, `basename` will never be needed on the last conditional. Also, don't backslash out the '$'.

uberhacker: Please use basename on the last line.  That is a full path to ld.
```
current=\$(basename \$(tail -1 ld | cut -d' ' -f1))
```

uberhacker: If find fails, we want to know why, correct?  Please remove.

packages/ld_default.rb
```diff
@@ -2,54 +2,70 @@
...
+      0)
+        exit;;
+      *)
+        echo 'Please select from one of the options or enter 0 to cancel.'
+        echo
+    esac
+  done
+
+  echo '#!/bin/bash' > \${mktmp}/ld
+  echo \"\${new}\"' --library-path #{CREW_LIB_PREFIX} -rpath #{CREW_LIB_PREFIX} \"\$@\"' >> \${mktmp}/ld
```
uberhacker: It would be better to use the full path to `${new}`.  You have a lot of faith in your path.

packages/ld_default.rb
```diff
@@ -2,54 +2,70 @@
...
+  mktmp=\$(mktemp -d)
+
+  cd #{CREW_PREFIX}/bin
+
+  type=\$(file ld | cut -d' ' -f2 | tr -d '\\n')
+
+  if [ \"\${type}\" == \"ELF\" ]; then
+    current=\$(find . -inum $(ls -i ld 2>/dev/null | cut -d' ' -f1) 2>/dev/null | fgrep 'ld.' 2>/dev/null)
+  elif [ \"${type}\" == \"symbolic\" ]; then	
+    current=\$(basename $(readlink ld))
```
uberhacker: `current=\$(basename \$(readlink ld))`

packages/ld_default.rb
```diff
@@ -2,54 +2,70 @@
...
+  set -e
+
+  mktmp=\$(mktemp -d)
+
+  cd #{CREW_PREFIX}/bin
+
+  type=\$(file ld | cut -d' ' -f2 | tr -d '\\n')
+
+  if [ \"\${type}\" == \"ELF\" ]; then
+    current=\$(find . -inum $(ls -i ld 2>/dev/null | cut -d' ' -f1) 2>/dev/null | fgrep 'ld.' 2>/dev/null)
```
uberhacker: `current=\$(find . -inum \$(ls -i ld | cut -d' ' -f1) | fgrep 'ld.')`

packages/ld_default.rb
JL2210: The script will look like this: `ld.<ld> --library-path...`
You need to have your Chromebrew installation in the PATH, obviously.

uberhacker: I know.  If you use the full path like I had set it up originally, THAT would be an improvement.

JL2210: How? Why would we need that? Users shouldn't yet be using Chromebrew in a directory other than `/usr/local`, anyway.

packages/ld_default.rb
JL2210: All right. Done.

packages/ld_default.rb
JL2210: Done.

packages/ld_default.rb
JL2210: Done.

packages/ld_default.rb
JL2210: Sorry. I didn't get what you were saying. I think it's better to search the PATH for your linker instead of hardcoding the path (sort of, in the file).

packages/ld_default.rb
JL2210: Done.

packages/ld_default.rb
uberhacker: The reason you just explained proves my point.

packages/ld_default.rb
```diff
@@ -2,54 +2,70 @@
...
+  cd #{CREW_PREFIX}/bin
+
+  type=\$(file ld | cut -d' ' -f2 | tr -d '\\n')
+
+  if [ \"\${type}\" == \"ELF\" ]; then
+    current=\$(find . -inum \$(ls -i ld | cut -d' ' -f1) | fgrep 'ld.')
+  elif [ \"${type}\" == \"symbolic\" ]; then	
+    current=\$(basename \$(readlink ld))
+  else
+    current=\$(basename $(tail -1 ld | cut -d' ' -f1))
```
uberhacker: Missing `\` at `$(tail...`.

packages/ld_default.rb
JL2210: Exactly. Instead of using `/usr/local/bin/ld.*`, this uses `ld.*` and has the shell search for the binary in the `$PATH`.

JL2210: `--library-path #{CREW_LIB_PREFIX} -rpath #{CREW_LIB_PREFIX}` are both needed options, though.

packages/ld_default.rb
JL2210: Yeah. Fixed.

packages/ld_default.rb
uberhacker: We are making changes in this directory explicitly.  We don't want to use other linkers by searching `$PATH`.

JL2210: What other linkers? Even if there were any, their `ld` would have higher precedence over this one, depending on the `$PATH`. Example: If someone had Ada in the front of the path, the `ld` from ada would take precedence over this one, anyway.

JL2210:
```
chronos@localhost / $ which ld
/usr/local/bin/ld
chronos@localhost / $ PATH=~/.local/bin:$PATH
chronos@localhost / $ which ld
/home/chronos/user/.local/bin/ld
chronos@localhost / $ 
```

uberhacker: You are proving my point exactly why you should put the full path to ld.whatever in the ld bash script.  Although somebody could certainly make some other ld command take precedence over this one, what I don't want is some other ld.whatever to take precedence over the ld.whatever in this script.  Do you understand it yet?

JL2210: Yes. It's not going to.

JL2210: A directory later in the PATH can't take control of an earlier directory. If `ld.gold`, `ld.bfd`, and `ld.lld` are in `/usr/local/bin`, then they will always be executed first, unless the user chooses to add a directory to the front of the PATH. If so, then the `ld` in the directory in front of the path will be executed first, anyway, so this `ld` will not be used.

JL2210: I get what you're saying. However, if the user wants to change their PATH for the express reason of having another ld.whatever override this one, shouldn't they be allowed to?

uberhacker: I don't think you understand yet.

JL2210:
Thank you. If there are any issues, I'll make sure to push another PR.

PR #2926, opened 2018-12-09
Add Broadway

sajjaadf-alt:
Enables GUI apps to run in tab (Enables the use of App icons and more) (By @JL2210, He did not have time to upload)

PR #2937, opened 2018-12-10
Add Broadway package

JL2210:
Fixes #2926 

Broadway is a display manager that allows GTK3 apps to run in a browser window.


uberhacker:
Let's try and fix #2926.  We already have a discussion going on there.

JL2210:
Yes, see my comments on #2926. The recommendations don't work.

PR #2943, opened 2018-12-13
Fix broadway/sommelier interoperability

uberhacker:
The broadway and sommelier window managers cannot run simultaneously so this fix prevents one from running while the other is running.  It also introduces an elegant way to switch the GDK_BACKEND environment variable value as needed.

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64


skiman6010:
There we go. I was waiting for this!

JL2210:
Much better than I could have done.

PR #2947, opened 2018-12-13
Fix Broadway and Sommelier packages

JL2210:
Silences broadway and sommelier daemons (for certain cases in which output from ~/.bashrc is undesirable) 
Exports needed environment variables
Moves `chmod` and `chown` to `if` statement in Sommelier
@uberhacker, think before closing.

### DIFF
--- a/packages/broadway.rb
+++ b/packages/broadway.rb
@@ -3,7 +3,7 @@ require 'package'
 class Broadway < Package
   description 'Run GTK applications in a browser window.'
   homepage 'https://developer.gnome.org/gtk3/stable/gtk-broadway.html'
-  version 'gtk3.22-1'
+  version 'gtk3.22-2'
   source_url 'file:///dev/null'
   source_sha256 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
 
@@ -15,36 +15,32 @@ class Broadway < Package
   depends_on 'gtk3'
 
   def self.build
-    system "echo 'GDK_BACKEND=broadway' > .broadway.env"
-    system "echo 'XDG_RUNTIME_DIR=/var/run/chrome' >> .broadway.env"
-    system "echo 'BROADWAY_DISPLAY=:5' >> .broadway.env"
+    system "echo 'export GDK_BACKEND=broadway' > .broadway.env"
+    system "echo 'export XDG_RUNTIME_DIR=/var/run/chrome' >> .broadway.env"
+    system "echo 'export BROADWAY_DISPLAY=:5' >> .broadway.env"
     system "echo '#!/bin/bash' > initbroadway"
+    system "echo >> initbroadway"
     system "echo 'source ~/.broadway.env' >> initbroadway"
+    system "echo >> initbroadway"
     system "echo 'BROADWAYD=\$(pidof broadwayd 2>/dev/null)' >> initbroadway"
     system "echo 'if [ -z \"\${BROADWAYD}\" ]; then' >> initbroadway"
     system "echo '  [ -f #{CREW_PREFIX}/bin/stopsommelier ] && stopsommelier' >> initbroadway"
-    system "echo '  broadwayd \${BROADWAY_DISPLAY} &' >> initbroadway"
+    system "echo '  broadwayd \${BROADWAY_DISPLAY} &>/dev/null &' >> initbroadway"
     system "echo '  sleep 3' >> initbroadway"
     system "echo 'fi' >> initbroadway"
     system "echo 'BROADWAYD=\$(pidof broadwayd 2>/dev/null)' >> initbroadway"
-    system "echo 'if [ ! -z \"\${BROADWAYD}\" ]; then' >> initbroadway"
-    system "echo '  echo \"broadwayd process \${BROADWAYD} is running\"' >> initbroadway"
-    system "echo 'else' >> initbroadway"
-    system "echo '  echo \"broadwayd failed to start\"' >> initbroadway"
+    system "echo 'if [ -z \"\${BROADWAYD}\" ]; then' >> initbroadway"
     system "echo '  exit 1' >> initbroadway"
     system "echo 'fi' >> initbroadway"
     system "echo '#!/bin/bash' > stopbroadway"
     system "echo >> stopbroadway"
     system "echo 'BROADWAYD=\$(pidof broadwayd 2>/dev/null)' >> stopbroadway"
     system "echo 'if [ ! -z \"\${BROADWAYD}\" ]; then' >> stopbroadway"
-    system "echo '  pkill broadwayd' >> stopbroadway"
+    system "echo '  pkill broadwayd &>/dev/null' >> stopbroadway"
     system "echo '  sleep 3' >> stopbroadway"
     system "echo 'fi' >> stopbroadway"
-    system "echo 'BROADWAYD=\$(pidof broadwayd 2> /dev/null)' >> stopbroadway"
-    system "echo 'if [ -z \"\${BROADWAYD}\" ]; then' >> stopbroadway"
-    system "echo '  echo \"broadwayd process stopped\"' >> stopbroadway"
-    system "echo 'else' >> stopbroadway"
-    system "echo '  echo \"broadwayd process \${BROADWAYD} is running\"' >> stopbroadway"
+    system "echo 'BROADWAYD=\$(pidof broadwayd 2>/dev/null)' >> stopbroadway"
+    system "echo 'if [ ! -z \"\${BROADWAYD}\" ]; then' >> stopbroadway"
     system "echo '  exit 1' >> stopbroadway"
     system "echo 'fi' >> stopbroadway"
   end
@@ -53,7 +49,6 @@ class Broadway < Package
     system "install -Dm755 initbroadway #{CREW_DEST_PREFIX}/bin/initbroadway"
     system "install -Dm755 stopbroadway #{CREW_DEST_PREFIX}/bin/stopbroadway"
     system "install -Dm644 .broadway.env #{CREW_DEST_HOME}/.broadway.env"
-    system "cp .broadway.env ~/"
   end
 
   def self.postinstall
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -24,39 +24,33 @@ class Sommelier < Package
       system "sed -i 's,/lib/,/#{ARCH_LIB}/,g' Makefile"
       system "sed -i 's,-I.,-I. -I#{CREW_PREFIX}/include/pixman-1,g' Makefile"
       system "make PREFIX=#{CREW_PREFIX} SYSCONFDIR=#{CREW_PREFIX}/etc"
-      system "echo 'GDK_BACKEND=wayland' > .sommelier.env"
-      system "echo 'CLUTTER_BACKEND=wayland' >> .sommelier.env"
-      system "echo 'XDG_RUNTIME_DIR=/var/run/chrome' >> .sommelier.env"
-      system "echo 'WAYLAND_DISPLAY=wayland-0' >> .sommelier.env"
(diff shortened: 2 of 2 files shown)


packages/broadway.rb
```diff
@@ -15,36 +15,32 @@ class Broadway < Package
   depends_on 'gtk3'
 
   def self.build
-    system "echo 'GDK_BACKEND=broadway' > .broadway.env"
-    system "echo 'XDG_RUNTIME_DIR=/var/run/chrome' >> .broadway.env"
-    system "echo 'BROADWAY_DISPLAY=:5' >> .broadway.env"
+    system "echo 'export GDK_BACKEND=broadway' > .broadway.env"
+    system "echo 'export XDG_RUNTIME_DIR=/var/run/chrome' >> .broadway.env"
+    system "echo 'export BROADWAY_DISPLAY=:5' >> .broadway.env"
```
uberhacker: This is not how an environment file works.  Please spend some time and learn the correct way.

packages/broadway.rb
```diff
@@ -15,36 +15,32 @@ class Broadway < Package
...
     system "echo '#!/bin/bash' > initbroadway"
+    system "echo >> initbroadway"
     system "echo 'source ~/.broadway.env' >> initbroadway"
+    system "echo >> initbroadway"
     system "echo 'BROADWAYD=\$(pidof broadwayd 2>/dev/null)' >> initbroadway"
     system "echo 'if [ -z \"\${BROADWAYD}\" ]; then' >> initbroadway"
     system "echo '  [ -f #{CREW_PREFIX}/bin/stopsommelier ] && stopsommelier' >> initbroadway"
-    system "echo '  broadwayd \${BROADWAY_DISPLAY} &' >> initbroadway"
+    system "echo '  broadwayd \${BROADWAY_DISPLAY} &>/dev/null &' >> initbroadway"
```
uberhacker: Don't we want to display errors if they exist?  No harm in not including this.

packages/broadway.rb
```diff
@@ -15,36 +15,32 @@ class Broadway < Package
...
-    system "echo '  echo \"broadwayd failed to start\"' >> initbroadway"
+    system "echo 'if [ -z \"\${BROADWAYD}\" ]; then' >> initbroadway"
     system "echo '  exit 1' >> initbroadway"
     system "echo 'fi' >> initbroadway"
     system "echo '#!/bin/bash' > stopbroadway"
     system "echo >> stopbroadway"
     system "echo 'BROADWAYD=\$(pidof broadwayd 2>/dev/null)' >> stopbroadway"
     system "echo 'if [ ! -z \"\${BROADWAYD}\" ]; then' >> stopbroadway"
-    system "echo '  pkill broadwayd' >> stopbroadway"
+    system "echo '  pkill broadwayd &>/dev/null' >> stopbroadway"
```
uberhacker: No harm in excluding this.

packages/broadway.rb
JL2210: Excuse me? I probably know more about environment variables than you do.

packages/broadway.rb
JL2210: > Silences broadway and sommelier daemons (for certain cases in which output from ~/.bashrc is undesirable)

packages/broadway.rb
JL2210: > Silences broadway and sommelier daemons (for certain cases in which output from ~/.bashrc is undesirable)

packages/broadway.rb
```diff
@@ -53,7 +49,6 @@ def self.install
     system "install -Dm755 initbroadway #{CREW_DEST_PREFIX}/bin/initbroadway"
     system "install -Dm755 stopbroadway #{CREW_DEST_PREFIX}/bin/stopbroadway"
     system "install -Dm644 .broadway.env #{CREW_DEST_HOME}/.broadway.env"
-    system "cp .broadway.env ~/"
```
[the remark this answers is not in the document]
JL2210: Well, then `#{CREW_DEST_HOME}` is corrupt. Why go around the issue instead of fixing it?

packages/sommelier.rb
```diff
@@ -79,10 +72,10 @@ def self.postinstall
     puts "echo '# Sommelier environment variables + daemon' >> ~/.bashrc".lightblue
     puts "echo '# See https://github.com/dnschneid/crouton/wiki/Sommelier-(A-more-native-alternative-to-xiwi)' >> ~/.bashrc".lightblue
     puts "echo 'if [ ! -d /tmp/.X11-unix ]; then' >> ~/.bashrc".lightblue
-    puts "echo 'mkdir /tmp/.X11-unix' >> ~/.bashrc".lightblue
+    puts "echo '  mkdir /tmp/.X11-unix' >> ~/.bashrc".lightblue
+    puts "echo '  sudo chmod -R 1777 /tmp/.X11-unix' >> ~/.bashrc".lightblue
+    puts "echo '  sudo chown root:root /tmp/.X11-unix' >> ~/.bashrc".lightblue
     puts "echo 'fi' >> ~/.bashrc".lightblue
-    puts "echo 'sudo chmod -R 1777 /tmp/.X11-unix' >> ~/.bashrc".lightblue
-    puts "echo 'sudo chown root:root /tmp/.X11-unix' >> ~/.bashrc".lightblue
```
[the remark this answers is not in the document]
JL2210: There is no chance of this happening. The directory is removed after a reboot, and is not created by anything other than the sommelier daemon.

packages/sommelier.rb
```diff
@@ -24,39 +24,33 @@ def self.build
...
+      system "echo '  sudo killall sommelier &>/dev/null' >> stopsommelier"
       system "echo '  sleep 3' >> stopsommelier"
       system "echo 'fi' >> stopsommelier"
-      system "echo 'SOMM=\$(pidof sommelier 2> /dev/null)' >> stopsommelier"
-      system "echo 'if [ -z \"\$SOMM\" ]; then' >> stopsommelier"
-      system "echo '  echo \"sommelier process stopped\"' >> stopsommelier"
-      system "echo 'else' >> stopsommelier"
-      system "echo '  echo \"sommelier process \$SOMM is running\"' >> stopsommelier"
+      system "echo 'SOMM=\$(pidof sommelier 2>/dev/null)' >> stopsommelier"
+      system "echo 'if [ ! -z \"\$SOMM\" ]; then' >> stopsommelier"
```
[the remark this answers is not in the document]
JL2210: Exactly.
> Silences broadway and sommelier daemons (for certain cases in which output from ~/.bashrc is undesirable)

JL2210:
Reopen and we will discuss.

JL2210:
```
# This file is sourced by all *interactive* bash shells on startup,
# including some apparently interactive shells such as scp and rcp
# that can't tolerate any output.  So make sure this doesn't display
# anything or bad things will happen !
```

packages/broadway.rb
uberhacker: Whatever.  The `export` is not needed.

packages/broadway.rb
uberhacker: Are you saying any errors that might happen are undesirable also?

packages/broadway.rb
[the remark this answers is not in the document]
uberhacker: I'm certainly open to suggestions.

packages/sommelier.rb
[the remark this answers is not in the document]
uberhacker: So you think.  There might be another X11 package that creates it also.

packages/sommelier.rb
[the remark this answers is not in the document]
uberhacker: Why shouldn't I know the status of the command I just executed?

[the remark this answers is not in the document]
JL2210: `echo $?`

packages/sommelier.rb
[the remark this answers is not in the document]
JL2210: Nope. `fgrep -R .X11-unix /usr/local` doesn't return anything but Sommelier-related files.

packages/broadway.rb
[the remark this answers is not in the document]
JL2210: First of all, enable verbosity in Chromebrew. Forced `make V=0` is not helping.

packages/broadway.rb
JL2210: Not necessarily. I just don't want to interfere with the users' ability to use `scp` and `rcp`.

packages/broadway.rb
JL2210: Yes, it is. Try using it without `export`. Shell variables are different than environment variables. Programs can only interact with environment variables.

packages/sommelier.rb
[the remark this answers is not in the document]
uberhacker: How would I even know I needed to execute that if all the errors were sent to `/dev/null`?  You are not making any sense.

packages/broadway.rb
JL2210: `broadwayd process 4126 is running` is the only thing that it displays, and then it exits.

JL2210:
I have a better idea. We can use a `VERBOSE` environment variable to control whether output is sent out.

packages/broadway.rb
JL2210: If you don't use `export`, the variables only stay in the current shell, not on to child processes. Say if I wrote two bash scripts like this:
`calling.sh`:
```
#!/bin/bash

CHILDREN="how ever many you want"
./how_many.sh
```
`how_many.sh`:
```
#!/bin/bash

if [[ -z "${CHILDREN}" ]]; then
  CHILDREN="none at all"
fi

echo "${CHILDREN}"
```
It would echo "none at all".
However, if I added `export ` at the beginning of line 3 of `calling.sh`, the output would be "how ever many you want".

JL2210: If we only set `DISPLAY` / any of the other variables when we're **calling** the daemon, it will take ***no effect*** on the programs with which we are using the daemon.

uberhacker: I totally understand what you are talking about.  Take a look at the output of `ps ax|grep broadway`.  Do you see `broadwayd :5`?  If you do, then the environment variable is working as expected.

JL2210:
Not exactly what I mean. We can have the offending '`output`' only show when `VERBOSE` is set to `1` in `~/.bashrc`. We can just set the default to `0`.

packages/broadway.rb
JL2210: I do, however GTK applications **do not** show up in the browser window. With this fix, they do.

JL2210:
```
Loading extra module: /usr/share/crosh/dev.d/50-crosh.sh
Welcome to crosh, the Chrome OS developer shell.

If you got here by mistake, don't panic!  Just close this tab and carry on.

Type 'help' for a list of commands.

If you want to customize the look/behavior, you can use the options page.
Load it by using the Ctrl+Shift+P keyboard shortcut.

crosh> shell
broadwayd process 26399 is running
chronos@localhost / $ qemu-system-i386 
chronos@localhost / $ cat ~/.broadway.env 
GDK_BACKEND=broadway
XDG_RUNTIME_DIR=/var/run/chrome
BROADWAY_DISPLAY=:5
chronos@localhost / $ qemu-system-i386 
chronos@localhost / $ echo $GDK_BACKEND
broadway 
chronos@localhost / $ export GDK_BACKEND
chronos@localhost / $ qemu-system-i386 
Unable to init server: Could not connect: Connection refused
gtk initialization failed
chronos@localhost / $ export BROADWAY_DISPLAY
chronos@localhost / $ qemu-system-i386 

** (qemu-system-i386:26736): WARNING **: 11:11:37.736: experimental: using broadway, x11 virtual keysym
mapping - with very limited support. See also
https://bugzilla.gnome.org/show_bug.cgi?id=700105
```
When there's no error from `qemu-system-i386`, it runs in the display, not the browser window.

uberhacker: Fair enough.  I just tested and you are correct about gtk apps in the browser.  See #2950.

PR #2949, opened 2018-12-14
Updated Sublime Merge.

skiman6010:


packages/sublime_merge.rb
```diff
@@ -6,8 +6,8 @@ class Sublime_merge < Package
   version 'Build 1065'
```
uberhacker: Bump version?

skiman6010: Shush.

JL2210: :P

PR #2950, opened 2018-12-15
Fix broadway and sommelier to export environment variables

uberhacker:


uberhacker:
Can you please provide some proof of how this affects the `scp` or `rcp` commands?  The only person who is bothered by the error messages is you.

JL2210:
You **don't** think people are bothered by the error messages?
```
Warning: Unknown scale factor: 1111
Warning: Unknown scale factor: 952
Warning: Unknown scale factor: 909
Warning: Unknown scale factor: 870
Warning: Unknown scale factor: 833
Warning: Unknown scale factor: 769
Warning: Unknown scale factor: 667
sommelier process 3514 is running
/usr/local/sbin/sommelierd: line 2:  3514 Terminated              sommelier -X --x-display=$DISPLAY --scale=$SCALE --no-exit-with-child /bin/sh -c "/usr/local/etc/sommelierrc"
(EE) 
Fatal server error:
(EE) failed to read Wayland events: Broken pipe
(EE) 
Segmentation fault (core dumped)
```

JL2210:
As for `scp` and `rcp`:
> `broadwayd process 4126 is running` is the only thing that it displays, and then it exits.

uberhacker:
We should probably figure out a better way to terminate sommelier.  That `Segmentation fault (core dumped)` message is a bit worrisome.  It's probably filling up the disk with core dump logs.

JL2210:
I have an idea, though it will take a little while to make it stable. We can kill `/bin/bash /usr/local/sbin/sommelierd` instead of `sommelier`.

JL2210:
`SOMM=$(ps ax | grep '/usr/local/sbin/sommelierd' | grep -v 'grep' | cut -d' ' -f2)

JL2210:
We can use `pkill` instead of `killall`. Sorry for the bad explanation.

uberhacker:
If you kill sommelierd, the sommelier process will continue to run.  The killall is used to kill off both processes in one go.  We could use `killall -g sommelier` and it would be a bit more graceful.

JL2210:
OK. To be honest, when I was using `pkill` in the original `broadway` package, I thought it was pretty hilarious that it printed `Terminated`. That brought back some good memories.

PR #2952, opened 2018-12-18
Add libcdio package and dependency

uberhacker:
The GNU Compact Disc Input and Control library (libcdio) contains a library for CD-ROM and CD image access.  See http://www.gnu.org/software/libcdio/.  Depends on libcddb (included).

Libcddb is a C library to access data on a CDDB server (freedb.org).  See http://libcddb.sourceforge.net/.

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/libcddb.rb
+++ b/packages/libcddb.rb
@@ -1,0 +1,31 @@
+require 'package'
+
+class Libcddb < Package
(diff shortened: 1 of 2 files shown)


JL2210:
Ah. Thank you! This is needed for Gstreamer!

PR #2953, opened 2018-12-18
Add uchardet package

uberhacker:
uchardet is an encoding detector library, which takes a sequence of bytes in an unknown character encoding without any additional information, and attempts to determine the encoding of the text.  See https://www.freedesktop.org/wiki/Software/uchardet/.

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/uchardet.rb
+++ b/packages/uchardet.rb
@@ -1,0 +1,30 @@
+require 'package'
+
+class Uchardet < Package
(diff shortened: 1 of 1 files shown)


JL2210:
Have you checked `/usr/local/lib64` on x86_64 for this?

uberhacker:
Good call.  Surprisingly, this is one of those rare times cmake puts the libraries in the correct location without any extra logic.

JL2210:
You can say **that** again.

PR #2957, opened 2018-12-18
Fix CREW_DEST_HOME constant

JL2210:
Fixes CREW_DEST_HOME constant and packages files in `#{ENV['HOME']}`. Also fixes `install.sh` if any packages in the install script ever need files in `#{ENV['HOME']}`. Bumps `crew` version.


uberhacker:
Nice!

PR #2958, opened 2018-12-19
Update Git to 2.20.1

JL2210:
@uberhacker, please create binaries. I will modify `install.sh` afterwards.

### DIFF
--- a/packages/git.rb
+++ b/packages/git.rb
@@ -3,39 +3,46 @@ require 'package'
 class Git < Package
   description 'Git is a free and open source distributed version control system designed to handle everything from small to very large projects with speed and efficiency.'
   homepage 'https://git-scm.com/'
-  version '2.19.0'
-  source_url 'https://github.com/git/git/archive/v2.19.0.tar.gz'
-  source_sha256 'c4146d94d1c74de0d099bcd6ba22cc44f799332a46146ede4ad9c9f67971aa28'
+  version '2.20.1'
+  source_url 'https://github.com/git/git/archive/v2.20.1.tar.gz'
+  source_sha256 'a3fe1d35b00ec2e48f21f690d4d4b2b061ff132cba3d68684c530b12c8ef227f'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/git-2.19.0-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/git-2.19.0-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/git-2.19.0-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/git-2.19.0-chromeos-x86_64.tar.xz',
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/git-2.20.1-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/git-2.20.1-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/git-2.20.1-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/git-2.20.1-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'c8ed38f5e4e8ce268f42c0f65c47b8a18674c9704ed8a9eefb376aedb60ced69',
-     armv7l: 'c8ed38f5e4e8ce268f42c0f65c47b8a18674c9704ed8a9eefb376aedb60ced69',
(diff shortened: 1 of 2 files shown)


packages/git.rb
```diff
@@ -3,39 +3,40 @@
...
+    system "./configure",
+           "--without-tcltk",
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}",
+           "--with-perl=#{CREW_PREFIX}/bin/perl",
+           "--with-pager=#{CREW_PREFIX}/bin/less",
+           "--with-editor=#{CREW_PREFIX}/bin/nano",
+           "--with-python=#{CREW_PREFIX}/bin/python2",
+           "--with-gitconfig=#{CREW_PREFIX}/etc/gitconfig",
+           "--with-gitattributes=#{CREW_PREFIX}/etc/gitattributes"
```
uberhacker: I'm glad you split these options out like this.  Much easier to read.  I don't think we need to set pager or editor options since those are preferences the user may want to change.

packages/git.rb
```diff
@@ -3,39 +3,40 @@
...
     system "#{@make_cmd} all"
   end
 
   def self.install
     system "#{@make_cmd} DESTDIR=#{CREW_DEST_DIR} install"
-    system "mkdir -p #{CREW_DEST_PREFIX}/share/git-completion"
-    system "cp -r contrib/completion/* #{CREW_DEST_PREFIX}/share/git-completion"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/git-completion"
+    system "cp -a contrib/completion/. #{CREW_DEST_PREFIX}/share/git-completion/"
```
uberhacker: Why does this need changed?

packages/git.rb
JL2210: > Actually, those may not make any difference because PAGER and EDITOR are always forced in `/etc/profile` anyway.

I believe Chrome OS forces these variables.

packages/git.rb
```diff
@@ -54,6 +55,6 @@ def self.postinstall
 
   def self.check
     # Skip several t9010-svn-fe and t9011-svn-da tests since they fail.
-    #system "GIT_SKIP_TESTS='t9010.16 t9010.20 t9011.1[49] t9011.2[0346] t9011.31 ' #{@make_cmd} test"
+    system "GIT_SKIP_TESTS='t9010.16 t9010.20 t9011.1[49] t9011.2[0346] t9011.31 ' #{@make_cmd} test"
```
uberhacker: Skipping tests because they fail is not a good way to fix the issues.  You might as well not test at all.

packages/git.rb
JL2210: `FileUtils.mkdir_p "` takes advantage of Ruby more than `system "mkdir -p `.
`cp -a...` preserves file permissions.
Using '.' instead of '*' also copies dotfiles (Files with names starting with '.').

packages/git.rb
JL2210: OK. I didn't write this.

packages/git.rb
```diff
@@ -3,39 +3,40 @@
...
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/git-2.19.0-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/git-2.19.0-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/git-2.19.0-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'c8ed38f5e4e8ce268f42c0f65c47b8a18674c9704ed8a9eefb376aedb60ced69',
-     armv7l: 'c8ed38f5e4e8ce268f42c0f65c47b8a18674c9704ed8a9eefb376aedb60ced69',
-       i686: '3c0e0b6fcfd0c07b2722c6c89998133281a49a5d265b3fb7f3b5309247909bf3',
-     x86_64: '80cc91e879dde9578668ea091a5d5f3129b57231b334bc4aa4cf4a5739f4feee',
   })
```
uberhacker: Thank you for not generating just one binary for x86_64. :)

JL2210: You're welcome. I've quit doing that.

JL2210:
Let me fix these issues and get back to you.

packages/git.rb
uberhacker: Without the `-r` option, it won't recursively copy directories.  If there are no directories, then this is alright.

JL2210:
OK. All good?

packages/git.rb
uberhacker: But you uncommented it.  How do you know it doesn't fail on arm or i686?

packages/git.rb
JL2210: `-a` implies `-R`:
```
  -a, --archive                same as -dR --preserve=all
  -R, -r, --recursive          copy directories recursively
      --preserve[=ATTR_LIST]   preserve the specified attributes (default:
                                 mode,ownership,timestamps), if possible
                                 additional attributes: context, links, xattr,
                                 all
```

JL2210:
@uberhacker Have you seen my comments?

packages/git.rb
uberhacker: News to me.

PR #2965, opened 2018-12-19
Update readline7 package

JL2210:


packages/readline7.rb
```diff
@@ -3,35 +3,26 @@
...
   def self.build
-    system "wget -r -N -nd --no-parent ftp://ftp.gnu.org/gnu/readline/readline-7.0-patches/readline70-001 -P readline-7.0-patches"
-    system "wget -r -N -nd --no-parent ftp://ftp.gnu.org/gnu/readline/readline-7.0-patches/readline70-002 -P readline-7.0-patches"
-    system "wget -r -N -nd --no-parent ftp://ftp.gnu.org/gnu/readline/readline-7.0-patches/readline70-003 -P readline-7.0-patches"
-    # system "for i in readline-7.0-patches/*.sig; do gpg $i; done"
-    system "for i in readline-7.0-patches/readline70-???; do patch < $i; done"
+    system "wget --recursive --no-parent https://ftp.gnu.org/gnu/readline/readline-7.0-patches/"
+    system "mv ftp.gnu.org/gnu/readline/readline-7.0-patches/ ."
+    system "rmdir -p ftp.gnu.org/gnu/readline/"
+    system "for n in 1 2 3 4 5; do patch -Np0 -i readline-7.0-patches/readline70-00$n; done"
```
uberhacker: You could replace the 3 lines above with:
```
system "for i in ftp.gnu.org/gnu/readline/readline-7.0-patches/readline70-*; do patch -p4 < $i; done"
```
No need to move things around.

JL2210: Asterisks don't work correctly in this. It will change every time, even if they did work. `7.0p5` is the version, you know.

JL2210:
Done.

packages/readline7.rb
```diff
@@ -3,35 +3,23 @@
...
   def self.build
-    system "wget -r -N -nd --no-parent ftp://ftp.gnu.org/gnu/readline/readline-7.0-patches/readline70-001 -P readline-7.0-patches"
-    system "wget -r -N -nd --no-parent ftp://ftp.gnu.org/gnu/readline/readline-7.0-patches/readline70-002 -P readline-7.0-patches"
-    system "wget -r -N -nd --no-parent ftp://ftp.gnu.org/gnu/readline/readline-7.0-patches/readline70-003 -P readline-7.0-patches"
-    # system "for i in readline-7.0-patches/*.sig; do gpg $i; done"
-    system "for i in readline-7.0-patches/readline70-???; do patch < $i; done"
-
-    system "CC='gcc' ./configure --libdir=#{CREW_LIB_PREFIX} --disable-static --with-curses"
+    system "wget --recursive --no-parent https://ftp.gnu.org/gnu/readline/readline-7.0-patches/"
+    system "for n in 1 2 3 4 5; do patch -Np0 -i ftp.gnu.org/gnu/readline/readline-7.0-patches/readline70-00$n; done"
```
uberhacker: This will need to be changed if/when new patches are added.  It would be better to use a wildcard to list all the files, otherwise, this will need to be edited with the next update; something that could easily be overlooked.

JL2210: Exactly. All we need to do is bump the version number and add a new patch. The new patches could cause software incompatibilities, therefore we select the patches to be used. (Sorry if I made some spelling mistakes, I'm writing this on a bus.)

uberhacker: You are missing the point.  It would be easy to miss the new patch because you hardcoded 5 patches.  If there are 6 or more patches that need applied next time and you don't remember to add the new patches, that could cause an issue.

uberhacker: IOW, I prefer the way it was done previously. :)

JL2210: We're only downloading a little more data. I don't want the patchlevel to increase if the package version hasn't increased.

JL2210: If new patches need to be applied, we apply the patches and bump the package version to `7.0p6`, `7.0p7`, etc.

uberhacker: Maybe you should add a note so if somebody else updates this package in the future, they would understand how to do that.

JL2210: OK. That's fair.

JL2210:
@uberhacker Done.

packages/readline7.rb
```diff
@@ -1,37 +1,30 @@
...
 class Readline7 < Package
+
+  @patchlevel = 5
+
   description 'The GNU Readline library provides a set of functions for use by applications that allow users to edit command lines as they are typed in.'
   homepage 'http://cnswww.cns.cwru.edu/php/chet/readline/rltop.html'
-  version '7.0p3'
-  source_url 'ftp://ftp.gnu.org/gnu/readline/readline-7.0.tar.gz'
+  version "7.0p#{@patchlevel}"
+  source_url 'https://ftp.gnu.org/gnu/readline/readline-7.0.tar.gz'
```
uberhacker: Can you please change this line to `source_url 'https://ftpmirror.gnu.org/readline/readline-7.0.tar.gz'`?

packages/readline7.rb
```diff
@@ -1,37 +1,30 @@
 require 'package'
 
 class Readline7 < Package
+
+  @patchlevel = 5
```
uberhacker: This is a good idea.

JL2210: Thank you.

PR #2966, opened 2018-12-19
Update ruby from 2.5.1 to 2.5.3

uberhacker:
Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64


JL2210:
Darn. You got there before me. I was just going to create a PR.

JL2210:
Do you mind changing these dependencies? They're all in `buildessential`.

uberhacker:
I was hoping I beat you. :)  Please don't update core packages unless you can generate all the binaries since it is so critical to the install process.  I can do it much faster than trying to grab the update from your repo to update the binaries.

JL2210:
OK. I had some Readline issues building. Did you not?

JL2210:
Anyway,
> Do you mind changing these dependencies? They're all in `buildessential`.

uberhacker:
We don't want to start removing dependencies from core packages.  You don't want to grab the chicken before the egg.  Think about what is happening here.  We can't assume a dependency is going to be available when Chromebrew is installing initially.  For instance, after ruby is installed, it is used by crew to install other core packages.  If those dependencies aren't available to ruby, it could crash.  Not what you want to happen during installation.  Also, keeping the dependencies there helps to document which core packages are necessary.

uberhacker:
I didn't have problems with readline probably because I left the dependency in there. :)

JL2210:
You know that `install.sh` doesn't track dependencies. As for 'keeping the dependencies there helps to document which core packages are necessary', just look at `buildessential`.

JL2210:
I had `readline` installed, as it's ** shocked face ** part of `buildessential`. This is what I'm talking about. You can remove all the dependencies and it would still work.

JL2210:
I also did not have `readline7` installed at the time.

uberhacker:
In a way, install.sh does track dependencies.  Take out libiconv and see how well git works for you.  If we did what you suggest to all the core packages, there would be no core dependencies left.  With that said, buildessential, even though it's a fake package, does include some dependencies in ruby and since that is added in install.sh, it would be fine to remove.  You just have to be very careful about which dependencies you remove.

uberhacker:
Here is a prime example of why removing dependencies from core package is so dangerous.  The core packages gawk, gettext and less all depend on ncurses.  As you know, ncurses is a dependency for many, if not most, of the packages.  If we remove ncurses from all 3 of the other core packages, ncurses would never get installed.  Think about how well that would turn out.

JL2210:
Anyway, we should either include all dependencies for all packages (and maybe remove `buildessential), or include none of the `buildessential` dependencies (and put `core_packages.txt` packages into `buildessential`). I don't think half-and-half will work very well for us in the future.

JL2210:
Thank you. Also see the first suggestion:
> include all dependencies for all packages (and maybe remove `buildessential`)

uberhacker:
You are asking for pain if we do that.  It also increases the likelihood that we'll have more dependency loops.

JL2210:
Possibly. However, we can impose some rules on that. We cannot depend on basic toolchains ( `gcc`, `llvm`, and others ), and libraries that the toolchain depends on ( 'gmp', 'mpfr', 'mpc' ) in the toolchain packages themselves, other than build dependencies (`=> :build`).

uberhacker:
The reason I added `core_packages.txt` is to have a place to check if a dependency should be included in a package or not.

skiman6010:
Hush you two, this is fine.

PR #2970, opened 2018-12-20
Update ruby from 2.5.1 to 2.5.3

uberhacker:
_"Fresh install"_ tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

Also added an EDITOR section and improved comments in install.sh.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -25,22 +25,24 @@ case "$architecture" in
   exit 1;;
 esac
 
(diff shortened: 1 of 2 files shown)


skiman6010:
I don't trust you.

uberhacker:
I accept full responsibility if this breaks anything and you can put all the blame on me if it doesn't work.

PR #2973, opened 2018-12-23
Remove unnecessary sections of all packages that operate on files in $HOME

uberhacker:
Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

Unable to build dia on i686 so rather than mess with it, I decided since we're dropping support for gui apps, it was OK to exclude the pre-built binaries.  FWIW, gvim did not behave either even though the pre-built binaries generated fine.  We'll likely see fewer and fewer pre-built binaries for i686 gui apps in the future.

### DIFF
--- a/packages/asciinema.rb
+++ b/packages/asciinema.rb
@@ -3,21 +3,21 @@ require 'package'
 class Asciinema < Package
   description 'Terminal session recorder'
   homepage 'https://asciinema.org/'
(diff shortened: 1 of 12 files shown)


JL2210:
High five! The only way this PR could be better is if we could figure out a way to install `xdg_base` without Internet!

PR #2975, opened 2018-12-24
Update Readline package

JL2210:
Updates readline package to 6.3p8-2 and uses new versioning scheme seen in readline 7.0p5.

### DIFF
--- a/packages/readline.rb
+++ b/packages/readline.rb
@@ -1,44 +1,43 @@
 require 'package'
 
 class Readline < Package
+
+  @patchlevel = 8
+
   description 'The GNU Readline library provides a set of functions for use by applications that allow users to edit command lines as they are typed in.'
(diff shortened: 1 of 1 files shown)


JL2210:
Oh shoot. Let me fix this for a minute...

uberhacker:
In case you didn't know already, this is a core package.  We need the pre-built binaries before we can merge this, otherwise, it gets compiled on every new install.

JL2210:
Done.

JL2210:
Oh. Can you copy my work into one of your branches and go from there?

uberhacker:
I can but when you make changes to your PR after the fact, it makes me lose confidence that you have actually tested this to make sure it doesn't have any cascading bad effects.

JL2210:
I have, and it compiled fine. I accidentally omitted the leading section of the old `readline` package that had the system libraries override the libraries provided by the package. Wouldn't it be better for me to catch it and fix it then for you to have to do the same?

PR #2976, opened 2018-12-24
fix typo

nosarthur:


uberhacker:
@nosarthur: Thank you for the update!  Even the spellcheckers can't identify that error. :)

PR #2977, opened 2018-12-24
Add komodo package

uberhacker:
_"Yet Another Editor (YAE)"_:  Komodo Edit is a great editor if you\'re looking for something powerful, yet simple.  See https://www.activestate.com/products/komodo-edit/.

Tested on:
- [x] i686
- [x] x86_64

Installs with a bunch of errors because of spaces in the directory/file names but doesn't prevent it from working.  We could probably try and compile from source eventually for armv7l/aarch64 but for some reason, the GitHub repo at https://github.com/Komodo/KomodoEdit hasn't tagged a release for the latest version yet.  This package only includes the ready-made binaries as "source" for now.

### DIFF
--- a/packages/komodo.rb
+++ b/packages/komodo.rb
@@ -1,0 +1,35 @@
+require 'package'
+
+class Komodo < Package
+  description 'Komodo Edit is a great editor if you\'re looking for something powerful, yet simple.'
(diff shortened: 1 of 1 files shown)


sajjaadf:
This is amazing! It can replace vs code on my chromebook that did not like the weird port

PR #2984, opened 2018-12-25
Add SVGAlib package

JL2210:
SVGAlib is a low-level graphics library for Linux. It augments the C programming language, which doesn't provide support for graphics.

But there are lots of graphics programs written in C!

Yes, but they all rely on external library functions. C itself can only give you text. That's because all graphic functions are system dependent and are non-portable. Unfortunately, graphic routines coded for one operating system will not work under another unless they are completely rewritten. For example, graphics originally written for DOS or Windows are useless under Linux.

To code graphics in C under Linux, it is necessary to use an external set of functions which are native to Linux. SVGAlib is one such set.

### DIFF
--- a/packages/svgalib.rb
+++ b/packages/svgalib.rb
@@ -1,0 +1,55 @@
+require 'package'
+
+class Svgalib < Package
+  description 'SVGAlib is a low-level graphics library for Linux. It augments the C programming language, which doesn\'t provide support for graphics.'
+  homepage 'https://www.svgalib.org/'
+  version 'e698'
+  source_url 'https://github.com/ryanmcgrath/svgalib-1/archive/e698d8c11f938bb6039a1791dc54c620c228d308.tar.gz'
+  source_sha256 'ff39a2b375c36b688934d4be88193b1fd02d3809fa398dbf51d0fcd182087452'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.patch
+    system "sed -i 's,#ifndef __GNUC__,#ifdef __GNUC__,g' threeDKit/quickmath.h"
+    system "sed -i '/ldconfig/d' src/Makefile threeDKit/Makefile Makefile lrmi-0.9/Makefile.lrmi"
+    system "sed -i -e 's,#!/usr/bin/perl,#!#{CREW_PREFIX}/bin/perl,g' -e 's,/usr/lib/kbd,#{CREW_PREFIX}/lib/kbd,g' utils/svgakeymap"
+  end
+
+  def self.build
+    make_cmd = "make \
+                NO_ASM=y \
+                NO_HELPER=y \
+                BACKGROUND=y \
+                prefix=#{CREW_PREFIX} \
+                libdir=#{CREW_LIB_PREFIX} \
+                LIBDIR=#{CREW_LIB_PREFIX} \
+                shlibdir=#{CREW_LIB_PREFIX} \
+                datadir=#{CREW_PREFIX}/etc/vga \
+                mandir=#{CREW_PREFIX}/share/man \
+                sharedlibdir=#{CREW_LIB_PREFIX} \
+                rulesdir=#{CREW_PREFIX}/etc/udev/rules.d"
+    system "#{make_cmd} install"
+  end
+
+  def self.install
+    make_cmd = "make \
+                NO_ASM=y \
+                NO_HELPER=y \
+                BACKGROUND=y \
+                prefix=#{CREW_DEST_PREFIX} \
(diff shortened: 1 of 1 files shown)


packages/svgalib.rb
```diff
@@ -0,0 +1,55 @@
...
+           LIBDIR=#{CREW_DEST_LIB_PREFIX} \
+           shlibdir=#{CREW_DEST_LIB_PREFIX} \
+           datadir=#{CREW_DEST_PREFIX}/etc/vga \
+           mandir=#{CREW_DEST_PREFIX}/share/man \
+           sharedlibdir=#{CREW_DEST_LIB_PREFIX} \
+           rulesdir=#{CREW_DEST_PREFIX}/etc/udev/rules.d"
+    system "#{make_cmd} install"
+    system "#{make_cmd} demoprogs"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/lib/kbd/keytables"
+    system "gzip -9 /dev/null -c > #{CREW_DEST_PREFIX}/lib/kbd/keytables/us.map.gz"
```
uberhacker: No need to add `/dev/null` here since we might want to see if the map actually generated without errors.

packages/svgalib.rb
```diff
@@ -0,0 +1,55 @@
...
+           shlibdir=#{CREW_LIB_PREFIX} \
+           datadir=#{CREW_PREFIX}/etc/vga \
+           mandir=#{CREW_PREFIX}/share/man \
+           sharedlibdir=#{CREW_LIB_PREFIX} \
+           rulesdir=#{CREW_PREFIX}/etc/udev/rules.d"
+    system "#{make_cmd} install"
+  end
+
+  def self.install
+make_cmd = "make \
```
uberhacker: Please indent 4 spaces.

packages/svgalib.rb
```diff
@@ -0,0 +1,55 @@
...
+  })
+
+  def self.patch
+    system "sed -i 's,#ifndef __GNUC__,#ifdef __GNUC__,g' threeDKit/quickmath.h"
+    system "sed -i '/ldconfig/d' src/Makefile threeDKit/Makefile Makefile lrmi-0.9/Makefile.lrmi"
+    system "sed -i -e 's,#!/usr/bin/perl,#!#{CREW_PREFIX}/bin/perl,g' -e 's,/usr/lib/kbd,#{CREW_PREFIX}/lib/kbd,g' utils/svgakeymap"
+  end
+
+  def self.build
+make_cmd = "make \
```
uberhacker: Please indent 4 spaces.

uberhacker:
From the project page: `It's not svgalib "dash" 1, it's svgalib negative 1, because if you still use this library there's probably something wrong with you. ` :)

JL2210:
@uberhacker Can't help it. It's a dependency.

packages/svgalib.rb
JL2210: I want it to be lined up.

packages/svgalib.rb
JL2210: See below.

packages/svgalib.rb
JL2210: There is no map. It just fails if that file doesn't exist.

JL2210:
@uberhacker Done.

packages/svgalib.rb
uberhacker: Maybe we should be including rather than skirting around it?  Maybe another package is missing this file?

JL2210: No other package is missing this file, and it doesn't need to be included. Remember that this is an old package, and probably doesn't have the most correct dependencies on files.

packages/svgalib.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.patch
+    system "sed -i 's,#ifndef __GNUC__,#ifdef __GNUC__,g' threeDKit/quickmath.h"
+    system "sed -i '/ldconfig/d' src/Makefile threeDKit/Makefile Makefile lrmi-0.9/Makefile.lrmi"
+    system "sed -i -e 's,#!/usr/bin/perl,#!#{CREW_PREFIX}/bin/perl,g' -e 's,/usr/lib/kbd,#{CREW_PREFIX}/lib/kbd,g' utils/svgakeymap"
```
uberhacker: The file `/usr/local/lib/kbd` doesn't exist so not sure why you are even worried about it.  With that said, could this be a missing dependency?

JL2210: To be honest, I have no idea. It doesn't seem to use the keymap in any way at all, so I have no idea why it would be needed.

PR #2992, opened 2018-12-25
Add kbd package and dependency

uberhacker:
The kbd project contains tools for managing Linux console (Linux console, virtual terminals, keyboard, etc.) – mainly, what they do is loading console fonts and keyboard maps.  See http://kbd-project.org/.  Depends on linux_pam (included).

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/kbd.rb
+++ b/packages/kbd.rb
@@ -1,0 +1,36 @@
+require 'package'
+
+class Kbd < Package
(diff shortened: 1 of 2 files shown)


JL2210:
@uberhacker, use `openpam` instead of `linux_pam`. `linux_pam` is too harsh on restrictions.

PR #2996, opened 2018-12-26
Update Arduino IDE

JL2210:


packages/arduino_ide.rb
```diff
@@ -19,15 +22,18 @@ class Arduino_ide < Package
...
-      platform = 'linux32'
+      @platform = 'linux32'
     when 'armv7l', 'aarch64'
-      platform = 'linuxarm'
+      @platform = 'linuxarm'
     end
+
+  def self.build
+    version = @version
+    platform = @platform
```
uberhacker: These variable assignments shouldn't be necessary.  The instance variable defined at the top could be used anywhere in the class; including functions and methods.

packages/arduino_ide.rb
```diff
@@ -40,26 +46,21 @@ def self.build
...
-      platform = 'linux32'
-    when 'armv7l', 'aarch64'
-      platform = 'linuxarm'
-    end
+    version = @version
+    platform = @platform
     Dir.chdir("build") do
       FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share"
-      system "tar", "xpvf", "linux/arduino-#{version}-#{platform}.tar.xz", "-C", "#{CREW_DEST_PREFIX}/share/"
+      system "tar", "-xpf", "linux/arduino-#{version}-#{platform}.tar.xz", "-C", "#{CREW_DEST_PREFIX}/share/"
```
uberhacker: The dash is not necessary on the tar command.

uberhacker: And why did you remove the verbose option?

packages/arduino_ide.rb
JL2210: Done.

packages/arduino_ide.rb
JL2210: Done. Did not add back `-v` option as not needed.

JL2210:
@uberhacker This is ready.

packages/arduino_ide.rb
```diff
@@ -19,47 +22,41 @@ class Arduino_ide < Package
...
     when 'x86_64'
-      platform = 'linux64'
+      @platform = 'linux64'
     when 'i686'
-      platform = 'linux32'
+      @platform = 'linux32'
     when 'armv7l', 'aarch64'
-      platform = 'linuxarm'
+      @platform = 'linuxarm'
     end
```
uberhacker: This section should be moved up above the description since they are instance variables.  Please use proper indentation.

JL2210:
@uberhacker This is ready.

PR #2997, opened 2018-12-26
Update ld_default package with non-interactive mode support

JL2210:
Adds ability for ld_default to be ran in non-interactive mode (will be useful for packages).
Uses double square brackets, a more modern Bash-only feature that can evaluate strings with spaces in them. See [this link](https://serverfault.com/questions/52034/what-is-the-difference-between-double-and-single-square-brackets-in-bash) for more details.
Provides a fallback if linker type is unknown instead of spewing random binary data.
Adds message to postinstall about aforementioned added features.

### DIFF
--- a/packages/ld_default.rb
+++ b/packages/ld_default.rb
@@ -3,7 +3,7 @@ require 'package'
 class Ld_default < Package
   description 'Select the default ld executable and check for libraries in #{CREW_LIB_PREFIX} first'
   homepage 'https://github.com/skycocker/chromebrew/wiki/FAQ'
-  version '1.1'
+  version '1.2'
   source_url 'file:///dev/null'
   source_sha256 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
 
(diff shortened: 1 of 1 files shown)


uberhacker:
This is a good idea but you overlooked the possibility that you might need to return to the previous linker default after the package process is complete.  You would need a way to assign the current default, do your package processing with the new default and then return to what it was previously.

JL2210:
How would I go about doing that?

uberhacker:
I think you misunderstood what I was trying to communicate.  In the package that needs the new default linker, you would need to doing something like this:
```
@old_default = `ld_default d`
@new_default = `ld_default g`
...
system "ld_default #{@old_default}"
```
The `d` parameter is what grabs the current default linker.

JL2210:
Yes. See this package: 
```
class V4l_utils < Package
  description 'A conglomeration of video4linux libraries and utilities.'
  homepage 'https://linuxtv.org/'
  version '1.16.3'
  source_url 'https://linuxtv.org/downloads/v4l-utils/v4l-utils-1.16.3.tar.bz2'
  source_sha256 '7c5c0d49c130cf65d384f28e9f3a53c5f7d17bf18740c48c40810e0fbbed5b54'

  binary_url ({
  })
  binary_sha256 ({
  })

  #depends_on 'sdl2_image'
  depends_on 'libsdl2'
  depends_on 'libglu'
  depends_on 'alsa_lib'
  depends_on 'ld_default'

  def self.build
    original_default = "`ld_default g`"
    system "./configure",
           "--disable-static",
           "--prefix=#{CREW_PREFIX}",
           "--libdir=#{CREW_LIB_PREFIX}"
    system "ld_default #{original_default}"
    system 'make'
  end

  def self.install
  end
end
```

packages/ld_default.rb
```diff
@@ -21,28 +21,42 @@ def self.build
...
+    echo
+    echo '  Enter the new default linker:'
+    echo
+    echo '  b = ld.bfd'
+    echo '  g = ld.gold'
+    echo '  l = ld.lld'
+    echo '  0 = Cancel'
+    echo
+  else
+    echo \"\${current}\" | cut -d'.' -f2 | cut -c1
```
uberhacker: This could be shortened to `echo \"\${current}\" | cut -c1`.

uberhacker: Nevermind.  This is correct.

packages/ld_default.rb
```diff
@@ -21,28 +21,42 @@ def self.build
...
+  fi
 
   while true; do
-    read default
-    case \${default} in
+    if [[ -z \"\${1}\" ]]; then
+      read default
+    else
+      default=\"\${1}\"
+      if [[ \"\${default}\" != 'b' ]] && [[ \"\${default}\" != 'g' ]] && [[ \"\${default}\" != 'l' ]]; then
```
uberhacker: See if this works:
```
if ! [[ \"\${default}\" =~ "^[b|g|l]$" ]]; then

Comment (JL2210) [reply packages/ld_default.rb]: No. It doesn't.
```
chronos@localhost ~/Download/source/chromebrew $ ld_default b
l
Invalid linker configuration: b
chronos@localhost ~/Download/source/chromebrew $ ld_default g
l
Invalid linker configuration: g
chronos@localhost ~/Download/source/chromebrew $ ld_default l
```

uberhacker: You might need to escape more characters.  I'm thinking `|` is one of them.

uberhacker: Anyway, not worth wrestling with it unless you want to do it.

JL2210: I don't think `|` is a special character in Ruby. I'd have to check, though.

PR #3000, opened 2018-12-27
Update Code package (#3000)

JL2210:
Update Code package.

Yay! #3000!


packages/code.rb
```diff
@@ -1,54 +1,63 @@
...
-      puts
-      puts 'Visual Studio Code is currently not supported on ARM and AArch64.'.lightred
-      puts 'Please try HeadMelted.'.lightred
-      puts 'https://code.headmelted.com'.lightred
-      puts
-      puts 'Happy coding!'.lightgreen
-      puts
-      exit 1
-    end
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share"
```
uberhacker: `FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/code"`?

packages/code.rb
```diff
@@ -1,54 +1,63 @@
...
+    # ^^^ Do not remove this line.
+    system "ln", "-s", "../share/code/bin/code-oss", "#{CREW_DEST_PREFIX}/bin/code"
+    system "ln", "-s", "../share/code/bin/code-oss", "#{CREW_DEST_PREFIX}/bin/code-oss"
   end
 
   def self.postinstall
     puts
     puts 'Congratulations! You have installed Visual Studio Code on Chrome OS!'.lightgreen
-    puts 'Now, please run \'code\' to start Visual Studio.'.lightgreen
+    puts 'Now, please run \'code\' or \'code-oss\' to start Visual Studio.'.lightgreen
```
uberhacker: No need to include both `code` and `code-oss`.  Just include `code`.  Makes it less confusing.

packages/code.rb
JL2210: Nope. If we did that, the logic we use to rename the directory wouldn't work. The directory would be named `#{CREW_PREFIX}/share/code/VSCode-Linux-#{@arch}`.

packages/code.rb
JL2210: I don't find it confusing.

packages/code.rb
```diff
@@ -1,54 +1,63 @@
...
-      puts 'Please try HeadMelted.'.lightred
-      puts 'https://code.headmelted.com'.lightred
-      puts
-      puts 'Happy coding!'.lightgreen
-      puts
-      exit 1
-    end
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
+    system "mv", "../VSCode-linux-#{@arch}", "#{CREW_DEST_PREFIX}/share/code"
```
uberhacker: Your move should be:
```
system "mv", "../VSCode-linux-#{@arch}", "#{CREW_DEST_PREFIX}/share"
```
if you insist that the directory shouldn't be created ahead of time.

packages/code.rb
uberhacker: Please don't self resolve conversations.

packages/code.rb
JL2210: No. It will create incompatibilities with the old `code` package. Anyway, does it matter?

uberhacker: I can tell this wasn't tested.

JL2210: BS. I installed this package five times!

packages/code.rb
uberhacker: Don't be arrogant.

JL2210: All right. Why is it confusing?

packages/code.rb
JL2210: Sorry.

packages/code.rb
JL2210: It creates the directory `#{CREW_DEST_PREFIX}/share` and then renames the directory `../VSCode-linux-#{@arch}` to `#{CREW_DEST_PREFIX}/share/code`. Trust me, it works.

uberhacker: OK, that makes sense.  I didn't notice you were moving (aka renaming) instead of copying.  My bad.

JL2210: It's fine.

packages/code.rb
uberhacker: It goes against the KISS method.  KISS = Keep It Stupid Simple

JL2210: All right. The default executable name is `code-oss`, so I'll keep it at that.

JL2210: Done.

packages/code.rb
JL2210: Can this be resolved?

packages/code.rb
JL2210: Can this be resolved?

packages/code.rb
JL2210: Can this be resolved?

PR #3001, opened 2018-12-27
Rework the vim package

wwwiiilll:
This reworks the vim (and gvim) packages to keep user config safe. I had to spend time twice fighting this package for control over my `.vim` and `.vimrc`, luckily I use stow for my dotfiles so my originals were intact.

So this will:
- keep the runtime files in `share` where they belong
- provide a system-wide `vimrc` that is safely editable, subsequent upgrades will create a `.new` file just like all the other package managers out there (native support for that would be great)
- define some defaults mostly to emulate XDG support
- make the scripting engines optional by dynamic linking to them (maybe we should require py3 at least)

That's basically exactly like the usual setup on most linux distros, though this is mostly drawing from the Arch vim packages.

I still have to finish the gvim package, I don't have a working sommelier setup to test yet.

Feedback welcome, I'm also very curious about the reson behind puttin the runtime files in `~/.vim`


uberhacker:
Thank you for the update @wwwiiilll!  Sorry you experienced so much trouble with your upgrades.  As you probably already know, the current configuration was my failed attempt to make vim XDG compliant.  Please see the inline comments.

packages/gvim.rb
```diff
@@ -3,81 +3,50 @@
...
-    system "cp #{CREW_PREFIX}/share/vim/vim81/gvimrc_example.vim #{configdir}/gvimrc" unless File.size? "#{configdir}/gvimrc"
-    puts
-    puts "To finish the installation, execute the following:".lightblue
-    puts "echo '# Fixes Warning: Missing charsets in String to FontSet conversion' >> ~/.bashrc".lightblue
-    puts "echo 'export LC_ALL=C' >> ~/.bashrc && . ~/.bashrc".lightblue
-    puts
-    puts "The config files are located in $HOME/.vim".lightblue
-    puts
+    # runtime files are provided by the vim package
+    system "rm", "-r", "#{CREW_DEST_PREFIX}/share/vim"
```
uberhacker: No need to remove this folder since crew will automatically delete these working files after installation.

packages/vim.rb
```diff
@@ -3,79 +3,106 @@
...
-    system "cp -r #{CREW_DEST_PREFIX}/share/vim/vim81/* #{configdestdir}"
-    FileUtils.mkdir_p "#{CREW_DEST_HOME}"
-    FileUtils.mkdir_p "#{configdir}" unless Dir.exists? "#{configdir}"
-    system "ln -s #{configdir} #{CREW_DEST_HOME}/.vim"
-    system "ln -s #{configdir}/vimrc #{CREW_DEST_HOME}/.vimrc"
-    system "touch #{configdir}/vimrc #{configdestdir}/vimrc"
+
+    # remove desktop and icon files for the terminal package
+    system "rm", "-r", "#{CREW_DEST_PREFIX}/share/applications"
+    system "rm", "-r", "#{CREW_DEST_PREFIX}/share/icons"
```
uberhacker: No need to delete these folders.  They will be automatically removed by crew after installation.

packages/vim.rb
```diff
@@ -3,79 +3,106 @@
...
+      if ! isdirectory(expand(&g:directory))
+        silent! call mkdir(expand(&g:directory), 'p', 0700)
+      endif
+      if ! isdirectory(expand(&g:backupdir))
+        silent! call mkdir(expand(&g:backupdir), 'p', 0700)
+      endif
+      if ! isdirectory(expand(&g:undodir))
+        silent! call mkdir(expand(&g:undodir), 'p', 0700)
+      endif
+    EOF
```
uberhacker: This is probably a good idea.  I didn't do it since this file is so _"opinionated"_.

packages/vim.rb
```diff
@@ -3,79 +3,106 @@
...
-    system "cp #{CREW_PREFIX}/share/vim/vim81/vimrc_example.vim #{configdir}/vimrc" unless File.size? "#{configdir}/vimrc"
-    puts
-    puts "To finish the installation, execute the following:".lightblue
-    puts "echo 'export VIMRUNTIME=$HOME/.vim' >> ~/.bashrc && . ~/.bashrc".lightblue
-    puts
-    puts "The config files are located in $HOME/.vim".lightblue
-    puts
+    vimrc = "#{CREW_PREFIX}/etc/vimrc"
+    # keep user changes by writing to a new file
+    vimrc += ".new" if File.exists?(vimrc)
```
uberhacker: Might want to create a random extension since `vimrc.new` may already exist from a previous upgrade.

packages/gvim.rb
```diff
@@ -3,81 +3,50 @@
...
 
   def self.build
     system "./configure \
             --prefix=#{CREW_PREFIX} \
-            --disable-nls \
-            --enable-gui \
+            --localstatedir=#{CREW_PREFIX}/var/lib/vim \
+            --with-features=huge \
+            --with-compiledby='Chromebrew' \
+            --with-tlib=ncurses \
```
JL2210: If possible, can we use `ncursesw`?

packages/gvim.rb
```diff
@@ -3,81 +3,50 @@
...
-            --enable-rubyinterp \
-            --with-python3-command=#{CREW_PREFIX}/bin/python3 \
-            --with-ruby-command=#{CREW_PREFIX}/bin/ruby \
-            --with-features=huge \
+            --enable-cscope \
+            --enable-perlinterp=dynamic \
+            --enable-pythoninterp=dynamic \
+            --enable-python3interp=dynamic \
+            --enable-rubyinterp=dynamic \
+            --disable-nls \
```
JL2210: I think it would be better to ENABLE Native Language support.

packages/gvim.rb
JL2210: This is still in `def self.install`. I think I know what he's doing. He's removing those files because they're provided by the `vim` package.

packages/vim.rb
```diff
@@ -3,79 +3,106 @@
...
-      system "rm -rf $HOME/.vim"
-    end
-    if File.exists? "$HOME/.vimrc"
-      system "cp $HOME/.vimrc #{configdir}/vim/vimrc"
-    end
-  end
+  depends_on 'perl' => :build
+  depends_on 'python27' => :build
+  depends_on 'python3' => :build
+  depends_on 'ruby' => :build
```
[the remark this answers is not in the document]
JL2210: I agree with @uberhacker on `xdg_base`.

packages/vim.rb
```diff
@@ -3,79 +3,106 @@
...
-            --enable-rubyinterp \
-            --with-python3-command=#{CREW_PREFIX}/bin/python3 \
-            --with-ruby-command=#{CREW_PREFIX}/bin/ruby \
-            --with-features=huge \
+            --enable-cscope \
+            --enable-perlinterp=dynamic \
+            --enable-pythoninterp=dynamic \
+            --enable-python3interp=dynamic \
+            --enable-rubyinterp=dynamic \
+            --disable-nls \
```
JL2210: See above for NLS comment.

packages/vim.rb
```diff
@@ -3,79 +3,106 @@
...
+  depends_on 'ruby' => :build
 
   def self.build
     system "./configure \
             --prefix=#{CREW_PREFIX} \
-            --enable-gui=no \
+            --localstatedir=#{CREW_PREFIX}/var/lib/vim \
+            --with-features=huge \
+            --with-compiledby='Chromebrew' \
+            --with-tlib=ncurses \
```
JL2210: See above `ncursesw` comment.

packages/vim.rb
JL2210: @uberhacker These are in `def self.install`. He doesn't want them installed.

packages/vim.rb
```diff
@@ -3,79 +3,106 @@
...
-    system "touch #{configdir}/vimrc #{configdestdir}/vimrc"
+
+    # remove desktop and icon files for the terminal package
+    system "rm", "-r", "#{CREW_DEST_PREFIX}/share/applications"
+    system "rm", "-r", "#{CREW_DEST_PREFIX}/share/icons"
+
+    # add some defaults simulating XDG support
+    FileUtils.mkdir_p("#{CREW_DEST_PREFIX}/share/vim/vimfiles")
+    File.write("#{CREW_DEST_PREFIX}/share/vim/vimfiles/chromebrew.vim", <<~EOF)
+      " Global vimrc - setting some sane defaults
```
JL2210: Good job.

packages/gvim.rb
```diff
@@ -3,81 +3,50 @@
...
-     x86_64: 'd8fb8c1ee73295d0e7646235c012f1785c38a67c46db18846faf33b89384cdfc',
   })
 
   depends_on 'vim'
   depends_on 'gtk2'
   depends_on 'sommelier'
+  depends_on 'perl' => :build
+  depends_on 'python27' => :build
+  depends_on 'python3' => :build
+  depends_on 'ruby' => :build
```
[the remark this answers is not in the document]
wwwiiilll: Right, good to know that they are installed in the defined order.

packages/gvim.rb
wwwiiilll: It was `ncurses` previously so I didn't mess with it, I'll try building it with `w` and see what happens.

packages/gvim.rb
wwwiiilll: Again, it was previously disable, I'm pretty sure we can turn that on, I'll build and see

packages/gvim.rb
wwwiiilll: Exactly, that's not obvious from how the diff looks, but that in `install`, I don't want the runtime files to end up in the binary package.

packages/vim.rb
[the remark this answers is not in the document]
wwwiiilll: Python is listed as "optional" on most vim packages, we need need the headers at compile time to build support. It is only required at runtime if the user wishes to use it.

We're no longer using `.config` after this release, no need to have it executable.

packages/vim.rb
wwwiiilll: Exactly, this is reducing the package size, plus they will be provided by the `gvim` package instead.

packages/vim.rb
wwwiiilll: Which is why I've provided an opt-out using `PREFIX/etc/vimrc`, but otherwise those are standard-ish defaults.

packages/vim.rb
wwwiiilll: That's fine, the user is expected to diff that file again the `vimrc` in place and keep what they want, they're free to delete the `.new` afterwards. We just don't want to replace the potentially user-modified `vimrc` file. (which is providing the opt-out for the opiniated config above)

JL2210:
@wwwiiilll Instead of this format:
```
system "configure \
opt 1 \
opt 2"
```
Could you put it in this format? :
```
system "configure",
            "opt 1",
            "opt 2"
```

packages/gvim.rb
wwwiiilll: Build it with nls, seems like it's working fine, I'll keep it on.

packages/gvim.rb
```diff
@@ -12,41 +12,46 @@ class Gvim < Package
...
             --prefix=#{CREW_PREFIX} \
             --localstatedir=#{CREW_PREFIX}/var/lib/vim \
             --with-features=huge \
             --with-compiledby='Chromebrew' \
-            --with-tlib=ncurses \
-            --with-x \
-            --enable-gui \
-            --enable-fontset \
+            --with-x=yes \
+            --enable-gui=gtk2 \
```
JL2210: Use `gtk3` instead of `gtk2`.

packages/gvim.rb
```diff
@@ -12,41 +12,46 @@ class Gvim < Package
   binary_sha256 ({
   })
 
-  depends_on 'vim'
-  depends_on 'gtk2'
-  depends_on 'sommelier'
   depends_on 'perl' => :build
   depends_on 'python27' => :build
   depends_on 'python3' => :build
   depends_on 'ruby' => :build
```
JL2210: Don't need `perl` or `ruby`.

JL2210: Core packages.

packages/gvim.rb
```diff
@@ -3,81 +3,55 @@
...
 
-  def self.postinstall
-    configdir = "#{CREW_PREFIX}/.config/vim"
-    system "cp #{CREW_PREFIX}/share/vim/vim81/gvimrc_example.vim #{configdir}/gvimrc" unless File.size? "#{configdir}/gvimrc"
-    puts
-    puts "To finish the installation, execute the following:".lightblue
-    puts "echo '# Fixes Warning: Missing charsets in String to FontSet conversion' >> ~/.bashrc".lightblue
-    puts "echo 'export LC_ALL=C' >> ~/.bashrc && . ~/.bashrc".lightblue
-    puts
-    puts "The config files are located in $HOME/.vim".lightblue
```
JL2210: Modify this and the corresponding line in `vim.rb` to say `    puts "The config files are located in #{CREW_PREFIX}/etc".lightblue`

packages/vim.rb
```diff
@@ -3,79 +3,113 @@
 class Vim < Package
   description 'Vim is a highly configurable text editor built to make creating and changing any kind of text very efficient.'
   homepage 'http://www.vim.org/'
-  version '8.1-1'
+  version '8.1-2'
   source_url 'ftp://ftp.vim.org/pub/vim/unix/vim-8.1.tar.bz2'
```
JL2210: Change `ftp://` to `http://` in both packages.

packages/vim.rb
```diff
@@ -3,79 +3,113 @@
...
-            --enable-gui=no \
-            --without-x \
-            --disable-nls \
-            --enable-multibyte \
-            --with-tlib=ncurses \
-            --enable-perlinterp \
-            --enable-python3interp \
-            --enable-rubyinterp \
-            --with-python3-command=#{CREW_PREFIX}/bin/python3 \
-            --with-ruby-command=#{CREW_PREFIX}/bin/ruby \
```
JL2210: Might keep the `--with-*-command=` lines.

packages/gvim.rb
wwwiiilll: Yes, I have to bring the instructions back, especially to warn about the `export VIMRUNTIME` that will break things.

packages/vim.rb
wwwiiilll: I don't think they are used with `=dynamic`, but I'll check what's the impact.

wwwiiilll:
I'm still trying to get sommelier to work to test out gvim.

Also, unfortunately there is necessary overlap between the `vim` and `gvim` packages. So removing gvim will take out the vim binaries as well and force to re-install.

Otherwise I think this looks alright for now.

uberhacker:
I actually considered the overlap the last time I worked on these packages.  There wasn't (or at least I didn't notice) the overlap with binaries.

packages/gvim.rb
uberhacker: I think we need to be careful here.  Are you sure the man files are vim and not gvim?

uberhacker: The entire section you removed considered the overlap.

packages/vim.rb
```diff
@@ -3,79 +3,119 @@
...
+
+      set backspace=indent,eol,start
+      set ruler
+      set suffixes+=.aux,.bbl,.blg,.brf,.cb,.dvi,.idx,.ilg,.ind,.inx,.jpg,.log,.out,.png,.toc
+      set suffixes-=.h
+      set suffixes-=.obj
+
+      " Move temporary files to a secure location to protect against CVE-2017-1000382
+      if exists('$XDG_CACHE_HOME')
+        let &g:directory=$XDG_CACHE_HOME
```
uberhacker: This variable will require the xdg_base dependency.  Better make sure this variable is not expanded.  The `$` might need escaped.

packages/vim.rb
uberhacker: Just keep in mind that if they forget about the `.new` file (or if they never knew in the first place), any subsequent installs will overwrite the "saved" file.

packages/gvim.rb
uberhacker: The vim directory has `gvimrc_example.vim` and this will wipe that away.

packages/gvim.rb
```diff
@@ -3,81 +3,63 @@
...
-    system "mv #{tmpdir}/* #{CREW_DEST_PREFIX}/bin"
-    system "rm -rf #{tmpdir}"
-    system "compressdoc --gzip -9 #{CREW_DEST_PREFIX}/share/man/man1"
-    FileUtils.mkdir_p "#{configdestdir}"
-    FileUtils.mkdir_p "#{CREW_DEST_HOME}"
-    system "touch #{configdir}/gvimrc #{configdestdir}/gvimrc"
-    system "ln -s #{configdir}/gvimrc #{CREW_DEST_HOME}/.gvimrc"
+    system "make", "VIMRCLOC=#{CREW_PREFIX}/etc", "VIMRUNTIMEDIR=#{CREW_PREFIX}/share/vim/vim81", "DESTDIR=#{CREW_DEST_DIR}", "install"
+
+    # these files are provided by the vim package
```
uberhacker: Not all of these files are provided by the vim package.

packages/vim.rb
wwwiiilll: It will be used if it's set, `xdg_base` is optional in this case as we'll fallback to `~/.cache` and required directories will be created.

packages/gvim.rb
wwwiiilll: They should be, `make install` will install the same runtime files and man pages regardless of the configure switches.

wwwiiilll: Regarding the overlap, I'll double check this, but the `vim` binary is different when building with GUI support. I actually tried to exclude the common binaries in `gvim` and starting it would just fail with "not enabled at compile".

wwwiiilll:
Regarding the overlap, I'll double check this, but the `vim` binary is different when building with GUI support. I actually tried to exclude the common binaries in `gvim` and starting it would just fail with "not enabled at compile".

packages/gvim.rb
uberhacker: See comment below.  Before we do this, we really need to compare both installs before removing entire directories willy-nilly.

wwwiiilll: I did compare the runtime files, I'll check again for the man pages though.

packages/vim.rb
wwwiiilll: Why would overwriting the `.new` file be an issue? Its only purpose is to allow users to basically manually install changes from an upgrade. It is actually by-design that the "new" file be always overwritten by the package manager. This allows us to propose changes without imposing them.

JL2210:
Try forcefully removing the `vim` binary after build, e.g. `system "rm -f #{CREW_DEST_PREFIX}/bin/vim"`

wwwiiilll:
I validated the overlap situation by diffing the result of both full builds:
- the entire `PREFIX/share/vim` is identical byte for byte whether GUI is enabled or not
- extra man pages are included for the additional binaries when enabling GUI 
- the `vim` binaries are very much different when building with GUI support

Considering the binary incompatibilities, I don't have any other choise but to consider `vim` and `gvim` as distinct packages.

JL2210:
Just a note: "choice", not "choise".

uberhacker:
You didn't have to rewrite.  The package already handled the overlap.  I like what you did with the `~/.vim` and `~/.vimrc` stuff, however.  @JL2210: Canadians spell things differently than the US.

JL2210:
Oh. Sorry about that, @wwwiiilll.

wwwiiilll:
@JL2210 that's not the right spelling even by our standard, not sure how that slipped out

@uberhacker I finally figured out how it was handled previously, I'll re-incorporate that

uberhacker:
No worries.  I appreciate your efforts!

JL2210:
Wish you would say that to **me**...

wwwiiilll:
I got frustrated with the localised man pages for `gvim` and kind of gave up. So I extracted the common runtime in the `vim_runtime` package as I intended on doing initially, this includes the system wide config to avoid repeating that everywhere.

The only downside is that when compiling from source you have to compile things twice, once for the runtime because a vim binary is needed for the install process and a second time for the actual `vim` or `gvim` package. But with pre-built package this will work great.

Also, I changed the sources to the github releases as recommended on the vim homepage itself to get the latest patches.

wwwiiilll:
@uberhacker unfortunately I still can't get sommelier to work to test `gvim`, but otherwise I'm running the `vim` build at the moment with my usual config and plugin setup and I don't see any issues with it.

wwwiiilll:
Oh yeah other thing, this assumes a user won't install both `vim` and `gvim`. I thought about putting a `preinstall` to check if the other variant is installed (or being installed?) and fail with a helpful message, but I'm not too sure about that.

uberhacker:
Well, I imagine some users would want both options.  Maybe work on it some more?

JL2210:
Hint: Have 'gvim' and 'vim' both depend on 'vim_common' and remove
conflicting files from the other packages, e.g. Put all common files in
'vim_runtime'

wwwiiilll:
Well, `gvim` includes `vim`, so there really is no reason to have both.

JL2210:
There we go. Remove `vim` from the `gvim` package.

JL2210:
As I suggested before, `system "rm #{CREW_DEST_PREFIX}/bin/vim"`

wwwiiilll:
@JL2210 that's what we had before and that's what I'm avoiding, from a maintenance perspective it's not helpful.

If you look at how the vim packages are setup on all major distros you'll notice that:

`vim` conflicts with `gvim`
`gvim` conflicts with `vim`, but provides `vim`
both depend on `vim-runtime` or similar

If you look at Debian you even have a `vim-gtk2`, `vim-gtk3`, `vim-athena`, ... for every variant of the GUI which are all conflicting with each other and "providing" vim.

That's exactly how I've laid that out here, except I can't express the "conflict" or "provides" bits, it doesn't really matter here because crew doesn't truly support dependency management.

uberhacker:
If you can convince me the vim executable is needed in gvim, then I'm on board.  Otherwise, it makes no sense to me.

wwwiiilll:
Well, `gvim` *IS* `vim` with some additional compile options, technically there's no such thing as "gvim".

JL2210:
Users can either install the graphical version of `vim` and get `gvim` and `vim`, or they can install the more minimal `vim` package and only get `vim`. How is that?

uberhacker:
I get that but you don't need both `gvim` and `vim` executables in the gvim package.  That doesn't make any sense at all.  It does make sense if vim is a dependency of gvim, however.  What I'm trying to say is you should be able to uninstall vim if gvim is installed or vice versa and have no impact on either package.  I thought that was the reason for having the vim_runtime package?

JL2210:
The `vim_runtime` package contains the files needed by the `vim` and `gvim` packages, not the other way around. How about YOU add a `=> :conflicts` option if you have a problem with it?

uberhacker:
That's what I'm trying to say.  The vim_runtime package should have the shared resources of both vim and gvim packages.

JL2210:
Exactly, however we should have `gvim` and `vim` in the `gvim` package, and only `vim` in the `vim` package. If we want them not to overwrite the other package's files, we *need* a `=> :conflicts` option.

wwwiiilll:
If you compile vim and "vim with GUI" and diff both folders, the intersection is what `vim_runtime` contains.

JL2210:
Bingo.

uberhacker:
_"Exactly, however we should have gvim and vim in the gvim package"_ Wrong!

JL2210:
Why so, if they're not meant to be installed together?

uberhacker:
Why do they _need_ to be installed together?  Just because every major distro does it that way?

JL2210:
Debian and Ubuntu are probably the two most major distros I know of. ***They*** don't do it that way.

uberhacker:
So what?  We've been using gvim and vim currently that way and I haven't heard any complaints from users yet.

JL2210:
So I'm not a user?

JL2210:
I doubt some of our users know what `vim` is, let alone know how it's compiled and installed,

uberhacker:
You have never complained about vim/gvim yet other than arguing that we should be doing the same way everybody else does it without a logical explanation of the reason why.

JL2210:
> the same way everybody else does it
Gee, maybe there's a *reason* other people do it that way?

uberhacker:
Yeah, why don't you explain it instead of arguing in circles?

JL2210:
Because `vim` files differ if `vim` has been compiled with GUI support, and we don't want differing files in packages.

uberhacker:
Dude, you are missing the point.  We don't need vim in the gvim package because we already have gvim.  Duh?

JL2210:
For example:
> the vim binary is different when building with GUI support.

wwwiiilll:
What I can say is that doing it this way is easier on package maintainers. You just `make install` and package that in both cases, no filtering is done aside from the trully common files (aka `$VIMRUNTIME`) and that way upstream changes are automatically handled with low effort for the downstream maintainers.

uberhacker:
I don't care.  When I type vim, I expect a cli interface.  When I type gvim, I expect a gui interface.  I don't want to type vim and get a gui interface.  That is dumb.

uberhacker:
Typical maintainers don't care about users.  That is where we differ.  Chromebrew should be convenient for users, not maintainers.

wwwiiilll:
You get the GUI if whatever command you typed begins with 'g', it's that dumb. See: https://github.com/vim/vim/blob/master/src/main.c#L1769-L1780

JL2210:
> Typical maintainers don't care about users.

Oh, come on. You look at (the Gentoo bugzilla)[https://bugs.gentoo.org/buglist.cgi?chfield=%5BBug%20creation%5D&chfieldfrom=7d&list_id=4133646] and tell me that again.

wwwiiilll:
If you read the man page for vim (https://linux.die.net/man/1/vim), you'll notice:

> -g
> If Vim has been compiled with GUI support, this option enables the GUI. If no GUI support was compiled in, an error message is given and Vim aborts.

So, if we have GUI support, the `vim` binary is expected to support the `-g` flag. Therefore, it wouldn't make sense to install `gvim` but have `vim -g` fail. So `gvim` has to provide `vim`.

uberhacker:
@wwwiiilll: Thank you for the explanation.  That makes more sense now.  How about this idea?  Why not just compile in gui support and use that for the vim package?  So basically, vim and gvim would be identical builds but gvim includes only the gvim binary and vim only includes the vim binary.  Would that work?  BTW, personally I don't care if the `-g` flag works or not.  I never intend to use vim that way.  There is something to be say about keeping things simple.

JL2210:
If we enable GUI support for the `vim` package, a bunch of users that don't use `gvim` will be installing extra dependencies that they don't need.

> BTW, personally I don't care if the -g flag works or not. I never intend to use vim that way.

Some people might. I know my dad does, for instance. Even if nobody else does, it's still what the `vim` maintainers originally designed it to be. Maybe we should figure out something else.

wwwiiilll:
@uberhacker I really don't want or expect to pull in the gtk dependencies when compiling or installing vim and I don't want GUI support in the cli-only package, this doesn't make any sense.

Also, there is no `gvim` binary, it's just a symlink to `vim`. On startup it detects that it was called as "gvim" and starts the GUI, effectively acting as if it was invoked as `vim -g`.

JL2210:
@wwwiiilll Well, if that works, there must be some check in `vim` to see if GUI support might have been installed and enabled in another package. Maybe it would be better to have `gvim` depend on `vim`?

uberhacker:
The solution to this is really simple.  Remove the gvim symlink and rename the gui enabled vim to gvim?

wwwiiilll:
@uberhacker but then `vim -g` won't work, except the man page says it should

JL2210:
Wait. The `gvim` package worked before, when GUI wasn't enabled in the `vim` package. That sounds like a good idea, but can we try and find out why?

uberhacker:
Of course it won't work.  The vim package is only compiled without gui support.

JL2210:
Please read the updated  comment.

uberhacker:
IOW, who cares if the `-g` switch works or not?  I certainly don't give a rats @$$.
```
-g
If Vim has been compiled with GUI support, this option enables the GUI. If no GUI support was compiled in, an error message is given and Vim aborts.
```
Notice the man page says *_If_* Vim has been compiled with GUI support.  I don't see why this is such a big deal.

wwwiiilll:
But the vim package **IS** compiled with GUI support if you want "with with GTK"

uberhacker:
Not if we rename it to gvim so it doesn't conflict with the non-gui vim binary.

JL2210:
If you gave a rat's @$$ about other people's opinions instead of focusing on your opinion, we might get somewhere with this.

uberhacker:
You aren't even listening.

JL2210:
I am. We all need to listen for this conversation to get anywhere.

JL2210:
The way it worked before, `vim -g` worked even when `vim` was compiled without GUI support. Why did it work then?

uberhacker:
The difference between you and me is I question everything.  You just want to do it the same way that every other distro does it without even understanding why they do it that way.

JL2210:
Read this:
> The way it worked before, vim -g worked even when vim was compiled without GUI support. Why did it work then?

wwwiiilll:
@JL2210 Since `cp` was dereferencing `gvim` the current package is essentially making multiple copies of the `vim` binary with GUI support as all the usual `g` aliases. However there is no way `-g` is working with the current setup.

uberhacker:
@wwwiiilll: The current gvim package gvim executable is a binary and not a symlink.  We are making this more complicated than necessary.  Here is what the current gvim package contains:
```
$ crew files gvim
(i) gvim: gVim is a highly configurable text editor built to make creating and changing any kind of text very efficient.
http://www.vim.org/
version 8.1-1
/home/chronos/user/.gvimrc
/usr/local/bin/eview
/usr/local/bin/evim
/usr/local/bin/gview
/usr/local/bin/gvim
/usr/local/bin/gvimdiff
/usr/local/bin/gvimtutor
/usr/local/bin/rgview
/usr/local/bin/rgvim
/usr/local/.config/vim/gvimrc
/usr/local/share/man/man1/eview.1.gz
/usr/local/share/man/man1/evim.1.gz
/usr/local/share/man/man1/gview.1.gz
/usr/local/share/man/man1/gvim.1.gz
/usr/local/share/man/man1/gvimdiff.1.gz
/usr/local/share/man/man1/rgview.1.gz
/usr/local/share/man/man1/rgvim.1.gz
/usr/local/share/vim/vim81/gvimrc_example.vim
Total found: 18
Disk usage: 20M
```

JL2210:
Darn. I have to admit... That looks pretty good. Maybe we *could* sacrifice `vim -g` for that size?

wwwiiilll:
@uberhacker the current package contains binaries because `cp gvim DEST/gvim` will dereference the symlink and copy the `vim` file in the destination. That's fine, but technically not efficient.

With my version of the packages:

```
$ crew files vim
...
Total found: 161
Disk usage: 3.2M
```

and

```
$ crew files gvim
...
Total found: 279
Disk usage: 3.5M
```

So `gvim` is ~300k bigger than `vim` (disregarding the million of dependencies it pulls in).

JL2210:
Oh...

uberhacker:
How big is your `vim_runtime` package?  The problem we are trying to solve here is how to avoid conflicts between vim and gvim packages.  Technically correct or not, it avoids a lot of potential issues for users.

JL2210:
I could see that. If he used `cp` for all the files in the package, therefore copying data that's identical to other files in the package, that would be very inefficient.

wwwiiilll:
@uberhacker 

```
$ crew files vim_runtime
...
Total found: 1674
Disk usage: 32M
```

However that's rather irrelevant since those files were included in the `vim` package previously, so that's constant space no matter which method we use.

JL2210:
`20 + <something greater than 20>  <  38.7`

JL2210:
Just as much as is in the 'vim_runtime' package.

wwwiiilll:
@uberhacker this looks good to me, I really think having gvim provide vim is what makes more sense. However that might be skewed by years of habits with other distros. But from a maintainer perspective, I can see the benefits in the simpler packaging and I don't think it's more inconvenient for the user.

PR #3003, opened 2018-12-27
Allow 'code' to be built if Node.JS has not already been installed

JL2210:
And remove unused `@version` variable.

### DIFF
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -2,8 +2,6 @@ require 'package'
 
 class Code < Package
 
-  @version = "1.30.1"
-
   case ARCH
   when 'x86_64'
     @arch = 'x64'
@@ -15,8 +13,8 @@ class Code < Package
 
   description 'Visual Studio Code is a source code editor developed by Microsoft for Windows, Linux and macOS.'
   homepage 'https://code.visualstudio.com/'
-  version '1.30.1'
-  source_url "https://github.com/Microsoft/vscode/archive/#{@version}.tar.gz"
+  version '1.30.1-1'
+  source_url "https://github.com/Microsoft/vscode/archive/1.30.1.tar.gz"
   source_sha256 'bf558c2818159fd2a47fa80c882794505b5d73d118f94f46d09c98b388d9e203'
 
   binary_url ({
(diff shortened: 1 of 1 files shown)


packages/code.rb
```diff
@@ -35,13 +33,15 @@ class Code < Package
...
+    node_old = `nodebrew ls | fgrep 'current: ' | cut -d' ' -f2 | grep -v none`
     system "nodebrew install 8.15.0 || true"
     system "nodebrew use 8.15.0"
     system "npm install gulp"
     system "yarn install"
     system "yarn run gulp vscode-linux-#{@arch}"
-    system "nodebrew use #{node_old}"
+    if node_old != ''
+      system "nodebrew use #{node_old}"
+    end
```
uberhacker: This is not going to work.  node_old will always be not equal to '' since you didn't chomp.

uberhacker: Have you even tried building binaries?  Apparently, not since `nodebrew use none` always errors out.

JL2210: So all I need to do is add `.chomp` at the end of the line that assigns `#{node_old}`, right?

uberhacker: Anytime you use commands with backticks, you need to chomp the end to get rid of the non-printable characters.

JL2210: Got it. So is that ALL I need to do?

uberhacker: Test it without any node versions installed and see what happens.

JL2210: Done.

uberhacker:
The code package is completely broken on the stable channel.  I think you need more than this to fix it.

JL2210:
@uberhacker True, but this fixes a mistake that I admittedly made.

packages/code.rb
```diff
@@ -35,13 +33,15 @@ class Code < Package
...
   def self.build
-    node_old = `nodebrew ls | fgrep 'current: ' | cut -d' ' -f2`
+    node_old = `nodebrew ls | fgrep 'current: ' | cut -d' ' -f2 | grep -v none`.chomp
     system "nodebrew install 8.15.0 || true"
     system "nodebrew use 8.15.0"
     system "npm install gulp"
     system "yarn install"
     system "yarn run gulp vscode-linux-#{@arch}"
-    system "nodebrew use #{node_old}"
+    if node_old != ''
```
uberhacker: You don't need the ` | grep -v none`.  You could:
`system "nodebrew use #{node_old}" unless "#{node_old}" == "none"`
Also, another consideration is to remove this node version when complete, therefore not impeding on the users node space:
`system "nodebrew uninstall 8.15.0" unless "#{node_old}" == "v8.15.0"`

uberhacker: You might want to use a variable for "v8.15.0" also.

JL2210: Good idea. That sounds perfect.

JL2210: One step ahead of you there, @uberhacker.

uberhacker: How about?
```
    system "nodebrew install #{node_ver}" unless "#{node_ver_installed}" == "#{node_ver}"
    system "nodebrew use #{node_ver}" unless "#{node_old}" == "#{node_ver}"
```

JL2210: That works too.

PR #3005, opened 2018-12-27
Update libSDL package and add X and Wayland bindings

JL2210:
Needed for SDL plugins, such as SDL_image and SDL_net.

### DIFF
--- a/packages/libsdl.rb
+++ b/packages/libsdl.rb
@@ -3,32 +3,29 @@ require 'package'
 class Libsdl < Package
   description 'Simple DirectMedia Layer is a cross-platform development library designed to provide low level access to audio, keyboard, mouse, joystick, and graphics hardware via OpenGL and Direct3D.'
   homepage 'http://www.libsdl.org'
-  version '1.2.15-2'
+  version '1.2.15-3'
   source_url 'http://www.libsdl.org/release/SDL-1.2.15.tar.gz'
   source_sha256 'd6d316a793e5e348155f0dd93b979798933fb98aa1edebcc108829d6474aad00'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/libsdl-1.2.15-2-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/libsdl-1.2.15-2-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/libsdl-1.2.15-2-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/libsdl-1.2.15-2-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'dd572e285fc4485ec46c36f7612f06f78ee2565e4fa5814c2828c520ea7f44ac',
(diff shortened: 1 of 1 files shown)


packages/libsdl.rb
```diff
@@ -3,32 +3,31 @@
...
   })
   binary_sha256 ({
-    aarch64: 'dd572e285fc4485ec46c36f7612f06f78ee2565e4fa5814c2828c520ea7f44ac',
-     armv7l: 'dd572e285fc4485ec46c36f7612f06f78ee2565e4fa5814c2828c520ea7f44ac',
-       i686: '485362d999a6a6ab466c8a544df175b9e93db2b5149ee0f37be4bf33d6adc66b',
-     x86_64: 'f9c9cee413c6acd449fefadb44f8148595124d994b658fedca67642921369b43',
   })
 
   depends_on 'dbus'
   depends_on 'eudev'
```
uberhacker: `xwayland` already has `eudev` as a dependency.

packages/libsdl.rb
```diff
@@ -3,32 +3,31 @@
...
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/libsdl-1.2.15-2-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'dd572e285fc4485ec46c36f7612f06f78ee2565e4fa5814c2828c520ea7f44ac',
-     armv7l: 'dd572e285fc4485ec46c36f7612f06f78ee2565e4fa5814c2828c520ea7f44ac',
-       i686: '485362d999a6a6ab466c8a544df175b9e93db2b5149ee0f37be4bf33d6adc66b',
-     x86_64: 'f9c9cee413c6acd449fefadb44f8148595124d994b658fedca67642921369b43',
   })
 
   depends_on 'dbus'
```
uberhacker: `pulseaudio` already has `dbus` as a dependency.

packages/libsdl.rb
```diff
@@ -3,32 +3,31 @@
...
   depends_on 'eudev'
+  depends_on 'xwayland'
+  depends_on 'alsa_lib'
+  depends_on 'pulseaudio'
+  depends_on 'mesa'
+  depends_on 'nasm' => :build
+
+  def self.patch
+    system "sed -i 's,#ifdef LONG64,#if 0 // Causes a build failure with recent libx11 versions,g' src/video/x11/SDL_x11sym.h"
```
uberhacker: If you do this, do you realize everything inside this block will get ignored?

JL2210: It's only three statements. They all conflict with `libx11`.

JL2210:
@uberhacker Not a duplicate. Two different packages.

packages/libsdl.rb
JL2210: Done.

packages/libsdl.rb
JL2210: Done.

packages/libsdl.rb
JL2210: Conflicting statements:
```
/*
 * Not required...these only exist in code in headers on some 64-bit platforms,
 *  and are removed via macros elsewhere, so it's safe for them to be missing.
 */
#ifdef LONG64
SDL_X11_MODULE(IO_32BIT)
SDL_X11_SYM(int,_XData32,(Display *dpy,register long *data,unsigned len),(dpy,data,len),return)
SDL_X11_SYM(void,_XRead32,(Display *dpy,register long *data,long len),(dpy,data,len),)
#endif
```

PR #3006, opened 2018-12-27
Update libSDL2 package and add X and Wayland bindings

JL2210:
See description for libSDL


JL2210:
@uberhacker Not a duplicate. Two different packages.

PR #3008, opened 2018-12-28
Fill in dependencies for xinit

JL2210:
Add xclock and twm packages, needed for startx.
Works on x86_64.


packages/xinit.rb
```diff
@@ -20,12 +20,13 @@ class Xinit < Package
      x86_64: '5df3455157f222c07641d3ff605ff9365fbf4a19d014c2db09999bf84d128296',
   })
```
uberhacker: This doesn't require new binaries?

uberhacker: Bump version?

JL2210: Nope. Just adds runtime dependendies.

JL2210: It really should also depend on `xterm`, though. Let me fix that.

JL2210: Done?

uberhacker: I think you should ponder about your assumption that new runtime dependencies doesn't remove the need to recreate binaries.

packages/xinit.rb
```diff
@@ -20,12 +20,14 @@ class Xinit < Package
...
 
-  depends_on 'xorg_lib'
+  depends_on 'xclock'
+  depends_on 'xterm'
+  depends_on 'twm'
```
uberhacker: Maybe these apps should have xinit as their dependency instead?

packages/xinit.rb
JL2210:
```
chronos@localhost / $ crew remove xterm twm xclock
Xterm removed!
Twm removed!
Xclock removed!
chronos@localhost / $ startx
/usr/local/bin/startx: 186: /usr/local/bin/startx: xauth: not found
/usr/local/bin/startx: 203: /usr/local/bin/startx: xauth: not found
/usr/local/bin/startx: 203: /usr/local/bin/startx: xauth: not found

_XSERVTransmkdir: ERROR: euid != 0,directory /tmp/.X11-unix will not be created.
xinit: XFree86_VT property unexpectedly has 0 items instead of 1
/usr/local/etc/X11/xinit/xinitrc: 55: exec: xterm: not found
xinit: connection to X server lost

waiting for X server to shut down 
/usr/local/bin/startx: 225: /usr/local/bin/startx: xauth: not found
```
After installing `xterm`:
```
chronos@localhost / $ startx
/usr/local/bin/startx: 186: /usr/local/bin/startx: xauth: not found
/usr/local/bin/startx: 203: /usr/local/bin/startx: xauth: not found
/usr/local/bin/startx: 203: /usr/local/bin/startx: xauth: not found

_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root
xinit: XFree86_VT property unexpectedly has 0 items instead of 1
/usr/local/etc/X11/xinit/xinitrc: 52: /usr/local/etc/X11/xinit/xinitrc: xclock: not found
/usr/local/etc/X11/xinit/xinitrc: 51: /usr/local/etc/X11/xinit/xinitrc: twm: not found
```

packages/xinit.rb
JL2210: Nope. Read above.

packages/xinit.rb
JL2210: I haven't recompiled `xinit`, I promise.

JL2210: If you'd like, I could go ahead and add `xauth`.

JL2210: @uberhacker @skiman6010 @caglewebsolutions With all of these installed, `code` and `filezilla` work!

JL2210: I would think that `xauth` is probably what's fixing it.

uberhacker: Sorry, but even on a fresh install I cannot get `code` working.  I believe `xorg_server` is also a dependency of `xinit`.  There is missing configuration for `xinit` also.

uberhacker:
```
$ startx
xauth:  file /home/chronos/user/.serverauth.20347 does not exist

xinit: unable to run server "/usr/local/bin/X": No such file or directory
```

JL2210: That `xauth:` warning doesn't mean anything. It will create the file when needed. Have you installed all the dependencies? `xorg_server` is a dependency of `twm`. I should probably change that to `Xwayland`, though.

JL2210: Done.

uberhacker: I only installed xinit by itself on a fresh installation to determine if these dependencies like twm and xclock are needed.  Apparently, not.  I don't think this fixes the code issue, regardless.  Why would we need X with sommelier?  Isn't sommelier supposed to redirect X requests to the built-in wayland server already?  Also, since sommelier is a window manager, wouldn't this conflict with a running twm?

JL2210: Xinit starts an X server by itself. It is in no way to be used with sommelier. This wasn't originally supposed to be a fix for `code`. TWM is not supposed to be used with Sommelier. Read this comment and the ~five below it: https://github.com/skycocker/chromebrew/pull/3008#discussion_r244340494

JL2210: However, `filezilla` is definitely tested to work with these.

JL2210: If you wish to use `twm`, you need to stop sommelier.

uberhacker: You are not making any sense.  The filezilla package currently has sommelier as a dependency.  Are you saying we need to change the dependency from sommelier to twm?  I read those comments and on a fresh install, I don't experience the same startx errors.  I only needed xorg_server as a dependency.

JL2210: No. I think the `xauth` package might fix the issue with `filezilla`. I'm not suggesting changing the dependency from `sommelier` to `twm`. `twm` is only a window manager. It needs to be used with an X server.
`xorg_server` needs `/dev/tty0`. That's why I didn't use it.

JL2210: As a side note, `filezilla` does not need `sommelier` as a dependency. It is a GTK application.

JL2210: I guess this is a pointless argument. I'll bump the version and remove the binaries.

JL2210: Done.

packages/xinit.rb
uberhacker: I don't think all these "dependencies" are needed, especially xclock and xterm.  You might be able to make a case for twm, however, the window manager really should be up to the user to decide.  I agree xauth is needed.  Have you tried installing xinit on a fresh Chromebrew yet?

JL2210: `startx` throws errors if those commands aren't found. As `xinit` isn't depended on by any package, it shouldn't matter if it has a few extra dependencies.
Yes, I have installed `xinit` on a fresh Chromebrew installation. It works fine, and does not give errors about missing `xterm`, `xclock`, `xauth`, or `twm`.

uberhacker: You are contradicting yourself.  First you say it throws errors and then you say it works without the dependencies?  I'm confused.

JL2210: `startx` gives errors about `<program>: not found` if you don't have `xterm`, `xclock`, `xauth`, and `twm` installed. When I did a fresh Chromebrew installation, I copied those packages and `xinit` over to `/usr/local/lib/crew/packages/`, and installed them. My modified `xinit` package works fine on a fresh Chromebrew installation.

uberhacker: I don't think you understand what I'm asking.  Did you try to install xinit on a fresh Chromebrew without those "extra" dependencies?  The only real dependency should be xauth and nothing else.

JL2210: If I installed it without those dependencies, `startx` would not even start.
```
chronos@localhost ~/Download/source/chromebrew/packages $ startx
/usr/local/bin/startx: 186: /usr/local/bin/startx: xauth: not found
/usr/local/bin/startx: 203: /usr/local/bin/startx: xauth: not found
/usr/local/bin/startx: 203: /usr/local/bin/startx: xauth: not found

_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root
xinit: XFree86_VT property unexpectedly has 0 items instead of 1
/usr/local/etc/X11/xinit/xinitrc: 52: /usr/local/etc/X11/xinit/xinitrc: xclock: not found
/usr/local/etc/X11/xinit/xinitrc: 51: /usr/local/etc/X11/xinit/xinitrc: twm: not found
```
Without `xterm`:
```
chronos@localhost ~/Download/source/chromebrew/packages $ startx
/usr/local/bin/startx: 186: /usr/local/bin/startx: xauth: not found
/usr/local/bin/startx: 203: /usr/local/bin/startx: xauth: not found
/usr/local/bin/startx: 203: /usr/local/bin/startx: xauth: not found

_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root
xinit: XFree86_VT property unexpectedly has 0 items instead of 1
/usr/local/etc/X11/xinit/xinitrc: 55: exec: xterm: not found
xinit: connection to X server lost
```

PR #3018, opened 2018-12-29
Add sdl2_image package

JL2210:
SDL_image is an image loading library that is used with the SDL library, and almost as portable. It allows a programmer to use multiple image formats without having to code all the loading and conversion algorithms themselves.

### DIFF
--- a/packages/sdl2_image.rb
+++ b/packages/sdl2_image.rb
@@ -1,0 +1,31 @@
+require 'package'
+
+class Sdl2_image < Package
+  description 'SDL2_image is an image loading library that is used with the SDL2 library.'
+  homepage 'https://www.libsdl.org/projects/SDL_image/'
+  version '2.0.4'
+  source_url 'https://www.libsdl.org/projects/SDL_image/release/SDL2_image-2.0.4.tar.gz'
+  source_sha256 'e74ec49c2402eb242fbfa16f2f43a19582a74c2eabfbfb873f00d4250038ceac'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'libsdl2'
+  depends_on 'libjpeg'
+  depends_on 'libpng'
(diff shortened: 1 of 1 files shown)


packages/sdl2_image.rb
```diff
@@ -0,0 +1,31 @@
...
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'libsdl2'
+  depends_on 'libjpeg'
+  depends_on 'libpng'
+  depends_on 'libtiff'
+  depends_on 'libwebp'
```
uberhacker: The `libwebp` package already includes `libpng` and `libtiff` dependencies.

JL2210: It's what, another five seconds to resolve the dependencies? Five seconds for the security of knowing that all the dependencies are installed.

uberhacker: Apparently, you don't think it's a good idea to try and prevent possible dependency loops.

JL2210: I do, but this package compiles and runs correctly. There is no need to prevent "possible dependency loops" if the packages in question cannot even use `SDL2_image`.

JL2210: Just so you know, there's a difference between redundant dependencies and dependency loops.

uberhacker: This is a pointless argument.  I will just remove after adding the pre-built binaries.

JL2210: Nice to know you take into account my opinion.

uberhacker: Thank you for new package.

JL2210:
You're welcome. I'm sorry that I... overreacted.

packages/sdl2_image.rb
JL2210: You're welcome. I'm sorry that I... overreacted.

PR #3027, opened 2018-12-30
Update Mesa package

JL2210:
Depends on #3024


packages/mesa.rb
```diff
@@ -27,55 +19,40 @@ class Mesa < Package
...
   depends_on 'libxcb'
+  depends_on 'libxvmc'
   depends_on 'libxtrans'
   depends_on 'libxkbfile'
   depends_on 'wayland'
   depends_on 'python27'
   depends_on 'bison'
   depends_on 'libxshmfence'
   depends_on 'elfutils'
+  depends_on 'llvm'
```
uberhacker: Shouldn't this be a build dependency?

packages/mesa.rb
```diff
@@ -27,55 +19,40 @@ class Mesa < Package
...
   
   def self.build
-    #system "pip install --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR} Mako"
+    system "pip uninstall -y Mako MarkupSafe || :"
```
uberhacker: Or colon?  What is the purpose of doing that?

packages/mesa.rb
JL2210: No. Mesa uses LLVM libraries.

packages/mesa.rb
```diff
@@ -27,55 +19,40 @@ class Mesa < Package
...
+    system "pip install --install-option=\"--prefix=#{CREW_DEST_PREFIX}\" Mako"
```
uberhacker: This probably should be:
```
system "pip install --root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX} Mako==1.0.7"
```

packages/mesa.rb
JL2210: The command fails if `Mako` and `MarkupSafe` are not installed. `:` is an alias for `true`.

packages/mesa.rb
JL2210: Good idea.

packages/mesa.rb
```diff
@@ -27,55 +19,40 @@ class Mesa < Package
...
-                --enable-glx\
-                --enable-dri \
-                --enable-llvm"
-      end
+    system "./configure",
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}",
+           "--enable-shared-glapi",
+           "--with-gallium-drivers=i915,nouveau,r300,r600,radeonsi,pl111,svga,swrast,swr,tegra,vc4,virgl,freedreno",
+           "--with-dri-drivers=i915,i965,nouveau,radeon,r200,swrast",
```
uberhacker: Why do you always want to change from the original?  We should *NOT* be installing i915 or i965 drivers on arm.  Those are Intel drivers.  If you had an arm system, I might let this slide but you don't so please return to the previous configure options.

JL2210: How about you test it? If it works, then this can be resolved.

uberhacker: Just return it back to the way it was please.  I don't have time to reverse things you mess around with.

JL2210: Actually, I remember why I added the `--disable-asm` option. So stuff like this wouldn't happen.

JL2210: All I'm doing is asking you to test this on ARM before it gets merged. Could you do that for me?

packages/mesa.rb
uberhacker: This line won't be necessary if there is a `python_mako` package dependency.

uberhacker:
This is the result of an attempt to compile mesa on i686:
```
configure: error: LLVM target 'amdgpu' not enabled in your LLVM build. Required by r600.
mesa failed to install: `./configure --prefix=/usr/local --libdir=/usr/local/lib --enable-shared-glapi --with-gallium-drivers=i915,nouveau,r300,r600,radeonsi,pl111,svga,swrast,swr,tegra,vc4,virgl,freedreno --with-dri-drivers=i915,i965,nouveau,radeon,r200,swrast --enable-osmesa --enable-opengl --enable-egl --enable-xvmc --disable-asm --enable-nine --enable-gles1 --enable-gles2 --with-platforms=x11,drm,wayland --enable-gbm --enable-xa --enable-glx --enable-dri --enable-llvm` exited with 1
```

uberhacker:
This is the result of an attempt to compile mesa on armv7l:
```
checking for I915... no
configure: error: Package requirements (libdrm >= 2.4.75 libdrm_intel >= 2.4.75) were not met:

No package 'libdrm_intel' found

Alternatively, you may set the environment variables I915_CFLAGS
and I915_LIBS to avoid the need to call pkg-config.
mesa failed to install: `./configure --prefix=/usr/local --libdir=/usr/local/lib --enable-shared-glapi --with-gallium-drivers=i915,nouveau,r300,r600,radeonsi,pl111,svga,swrast,swr,tegra,vc4,virgl,freedreno --with-dri-drivers=i915,i965,nouveau,radeon,r200,swrast --enable-osmesa --enable-opengl --enable-egl --enable-xvmc --disable-asm --enable-nine --enable-gles1 --enable-gles2 --with-platforms=x11,drm,wayland --enable-gbm --enable-xa --enable-glx --enable-dri --enable-llvm` exited with 1
```
As you can see, it doesn't like the Intel driver.

JL2210:
@uberhacker Did you try compiling Mesa on armv7l after upgrading via. #3024?

packages/mesa.rb
```diff
@@ -27,55 +19,40 @@ class Mesa < Package
...
-                --with-platforms=x11,drm,wayland \
-                --enable-gbm \
-                --enable-xa \
-                --enable-glx\
-                --enable-dri \
-                --enable-llvm"
-      end
+    system "pip uninstall -y Mako MarkupSafe || :"
+    system "pip install --prefix \"#{CREW_PREFIX}\" --root \"#{CREW_DEST_DIR}\" Mako==1.0.7"
+    system "pip install --prefix \"#{CREW_PREFIX}\" Mako==1.0.7"
```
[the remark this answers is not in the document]
JL2210: I feel like it's kind of a waste of resources to create new packages for every Python package out there. I mean, I could see `six` and `setuptools`, but past that, is there any real point?

uberhacker:
Look at the comment above.

sajjaadf:
What is `mesa`?

JL2210:
https://www.mesa3d.org/

packages/mesa.rb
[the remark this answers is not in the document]
uberhacker: There is also something to be said about not including pip installs in specific packages but rather separating them out so they can be used modularly.  It appears Mako might fit in the same category as six and setuptools.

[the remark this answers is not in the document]
JL2210: What I'm saying is that there are thousands of Python packages. `six` and `setuptools` are used in almost all of them. On the contrary, `Mako` is used in very few. Along with that, `Mako` is only used once in all of the Chromebrew packages: `mesa`.

uberhacker:
Please restore the options back to the previous state.  This update breaks everything except x86_64 builds.  The Intel drivers are not compatible with armv7l/aarch64.

JL2210:
@uberhacker Use the new commit on #3024 and see if the issues persist. Also, notice this in the original PR message:
> Depends on #3024

uberhacker:
In case you didn't know already, mesa is a dependency of sommelier and that is a huge deal breaker for this update.

JL2210:
I know. I created the `sommelier` package.

JL2210:
@uberhacker Use the new commit on #3024 and see if the issues persist. Also, notice this in the original PR message:
> Depends on #3024

uberhacker:
I already tried to upgrade libdrm.  It's broken.

JL2210:
Let me see the error message.

uberhacker:
Look above.

JL2210:
@uberhacker You specified that those errors were from mesa. Use `libdrm` from #3024 and try to compile `mesa` again.

uberhacker:
There is a reason the drivers were split out this way.  Previous packages already struggled through what you *WANT* to remove.

uberhacker:
I already did try it and it breaks again.

JL2210:
Let me see the error message.

uberhacker:
Look above.  Please don't keep asking the same question over and over.

JL2210:
> This is the result of an attempt to compile mesa on armv7l:

> This is the result of an attempt to compile mesa on i686:

Nothing about errors compiling `libdrm`.

uberhacker:
Dude, libdrm compiles fine.  The problem is with mesa, not libdrm.  You cannot include Intel drivers on arm systems.

JL2210:
You have to upgrade `libdrm` with the LATEST COMMIT on #3024. I just pushed two new ones. Please try again.

uberhacker:
OK, I refreshed everything and tried to compile libdrm and mesa from the latest commit.  Still doesn't like the x86 processor references:
```
configure: error: LLVM target 'amdgpu' not enabled in your LLVM build. Required by r600.
mesa failed to install: `./configure --prefix=/usr/local --libdir=/usr/local/lib --enable-shared-glapi --with-gallium-drivers=i915,nouveau,r300,r600,radeonsi,pl111,svga,swrast,swr,tegra,vc4,virgl,freedreno --with-dri-drivers=i915,i965,nouveau,radeon,r200,swrast --enable-osmesa --enable-opengl --enable-egl --enable-xvmc --disable-asm --enable-nine --enable-gles1 --enable-gles2 --with-platforms=x11,drm,wayland --enable-gbm --enable-xa --enable-glx --enable-dri --enable-llvm` exited with 1
```

JL2210:
All right. I can work with that.

JL2210:
Can we modify `llvm` to fix this?

uberhacker:
Are you sure you want to run down this rabbit hole?  There is likely a deadly viper waiting at the bottom.

JL2210:
I don't know, I suppose we could add an exception for the `r600` target if it fixes this issue. The only computer I've ever known to have an `r600` is my AMD64 desktop.

JL2210:
I changed these specific lines. Is this OK?
```
    if "#{ARCH}" == "i686"
      r600 = ""
    else
      r600 = "r600,"
    fi
           "--with-gallium-drivers=i915,nouveau,r300,#{r600}radeonsi,pl111,svga,swrast,swr,tegra,vc4,virgl,freedreno",
```

uberhacker:
While that may solve the issue with i686, what about arm (a much larger user base)?

JL2210:
Do you have the same issue on arm?

uberhacker:
Uh, https://github.com/skycocker/chromebrew/pull/3027#issuecomment-450598245?

uberhacker:
Put in another condition for arm to exclude i9xx drivers and then we can return to the original configure options.

JL2210:
Oh. It says "x86 processor *references*".

JL2210:
Using `--enable-intel` in `libdrm` fixed that issue.

uberhacker:
Think about how retarded it is to enable intel on arm systems.

JL2210:
Thank you for calling my ideas retarded.

JL2210:
It's all C/C++. Why would it matter what platform it's compiled for?

uberhacker:
Obviously, you haven't thought about it yet.

JL2210:
So...?

uberhacker:
We don't need the Intel drivers enabled on arm and we don't need the tegra driver enabled on x86.  That doesn't make sense.

JL2210:
> We don't ***need*** the Intel drivers enabled on arm and we don't ***need*** the tegra driver enabled on x86. (edited for clarification)

We might not need them, but if they compile fine, they might save us a lot of time down the road.

uberhacker:
You aren't listening.

JL2210:
All right. I've given up. Which drivers for what platform?

Issue #2902, opened 2018-12-01
Cursor bug (Reference)


sajjaadf:
For any Gtk Gui app gets a black block as cursor

uberhacker:
Can you please provide your system specs?

sajjaadf:
chromebrew: Latest version, just ran crew update

ChromeOS: 72 canary December 1st update

Model: Asus C201 PA

Architecture: ArmV7L

JL2210:
72 is not supported. Revert to stable or do it yourself.

sajjaadf:
where is GDK cursor theme stored?

JL2210:
Try `crew install libxcursor xcb_util_cursor`

sajjaadf:
Thanks

--
Sent with DejaLu (https://dejalu.me?sig)

sajjaadf:
Put a label (Reference)

PR #3029, opened 2018-12-31
Fix cursor Bug

sajjaadf:
Just fix issue #2902

### DIFF
--- a/packages/scite.rb
+++ b/packages/scite.rb
@@ -25,6 +25,8 @@ class Scite < Package
   depends_on 'hicolor_icon_theme'
   depends_on 'shared_mime_info'
   depends_on 'sommelier'
+  depends_on 'libxcursor' 
+  depends_on 'xcb_util_cursor'
 
 
   def self.build


uberhacker:
Thank you for the PR @sajjaadf!  Please see the comment.

JL2210:
Are those Mac line endings?

sajjaadf:
which comment?

packages/scite.rb
```diff
@@ -25,6 +25,8 @@ class Scite < Package
   depends_on 'hicolor_icon_theme'
   depends_on 'shared_mime_info'
   depends_on 'sommelier'
+  depends_on 'libxcursor' 
+  depends_on 'xcb_util_cursor'
```
uberhacker: Maybe these should be included in the sommelier package to other apps don't have cursor issues?  In any event, we should be adding new dependencies above sommelier so the postinstall messages don't scroll past the users' visible screen area.

sajjaadf: okay

PR #3031, opened 2018-12-31
Update the Readme on the GH pages 

sajjaadf:
just that


skiman6010:
What did you even do here? Looks like you broke the images during the copy and didn't even add a link to the gh-pages... Which seems to be the original intent. I'm not sure our gh-pages even need a README so please in your next PR on this topic -if you choose to make one - defend your changes instead of saying "just that". 

Please double check your work before you submit a PR as well. We have a handy PR template to follow. Thank you for understanding.

sajjaadf:
sorry

PR #3046, opened 2019-01-01
Update attr to override system

### LINKED ISSUE #3043: [tilp2] Cannot execute tilp
```
$ tilp
tilp: relocation error: /usr/local/lib/libacl.so.1: symbol lgetxattr, version ATTR_1.0 not defined in file libattr.so.1 with link time reference
```

uberhacker:
Fixes #3043.


uberhacker:
Maybe it's because you make so many assumptions about your updates and totally disregard the fact that others have already beaten down the path that you want to divert away from because you _want_ to.

JL2210:
All right. I don't care. Go right ahead.

uberhacker:
I'm talking about the current PRs.

PR #3052, opened 2019-01-03
Delete dpkg.rb

skiman6010:
This package conflicts with the Crouton install scripts and doesn't seem necessary to keep since it is not working as intended.

@JL2210 feel free to defend this package before it is removed. As far as I can tell, it doesn't work to install deb files.


JL2210:
This package works perfectly fine. If, in the `.deb`, there are only files in `/usr/local`, as a pose to having files in `/usr` and `/` (the standard in distros such as Debian and Ubuntu), then the package installs and works fine. As was originally the purpose of the package, the tools to build packages work fine.

JL2210:
If you want to modify this package, add the `--disable-dselect` configure option. That will disable installing `.deb` files altogether.

skiman6010:
Maybe that should be added to the package then. To me, dpkg is used to install packages and if that isn't working, then it seems useless to me. If the user knows that it only works on select packages and used primarily to build, then it will be fine. Also that it conflicts with Crouton. 

I didn't know that building and select installs work.

JL2210:
>  If the user knows that it only works on select packages and used primarily to build, then it will be fine. 

I can help with that.

uberhacker:
That is not a good enough defense.  Lots of Debian packages install files all over the place outside of `/usr/local`.  What this fails to do is warn the user of incompatible .deb packages and continues to install without the user even knowing if the package installed correctly or not.  The user then gets confused when things don't _"just work"_.  We have been getting a lot of complaints from users and you offer no support other than this lame explanation.

JL2210:
All right. Just add `--disable-dselect` and be done with it. You can still extract packages.

JL2210:
No! What about people who already had `dpkg` installed? Do you know how ***HARD*** I worked on this?

uberhacker:
I understand you put a lot of effort into this but it has created more confusion for users than we are willing to support.  It does not install packages the Chromebrew way and conflicts with the way we intended for installation.

JL2210:
Can we disable `dselect` and re-add it?

sajjaadf:
I agree with @JL2210

JL2210:
All right. I might try sometime.

PR #3063, opened 2019-01-06
Update Xorg server

JL2210:
Makes Xwayland a meta-package as xorg_server contains
the Xwayland binary. Enables more features and X servers.

### DIFF
--- a/packages/xwayland.rb
+++ b/packages/xwayland.rb
@@ -1,49 +1,12 @@
 require 'package'
 
 class Xwayland < Package
-  description 'X server configured to (only) work with weston'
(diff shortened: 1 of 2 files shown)


packages/xwayland.rb
```diff
@@ -1,49 +1,12 @@
 require 'package'
 
 class Xwayland < Package
-  description 'X server configured to (only) work with weston'
+  description 'The Xorg Server is the core of the X Window system.'
```
uberhacker: Probably should mention something about Wayland here.

JL2210: Eh. It's just a meta-package that depends on xorg_server.

PR #3060, opened 2019-01-06
Edit Mesa package

JL2210:
Does not bump version, as Mesa is uninstallable at this point in time.
Fixes #3053

PR #3064, opened 2019-01-06
Update Xorg server

JL2210:
Removes the Xwayland package as xorg_server contains
the Xwayland binary. Enables more features and X servers.
Uses xorg_server instead of xwayland in `libsdl`, `libsdl2`, `libwmf`, `sommelier`, and `xinit`.


uberhacker:
Does this fix the dependency loop issue?  That should be the highest priority fix right now.

JL2210:
Yes. I can confirm that.

uberhacker:
Great!  Can you compile mesa without errors?  That would be the next priority fix.  Currently, I cannot compile that on x86_64. :confused:

JL2210:
The problem is demonstrated in this diagram:
```
xwayland -> xinit -> xterm -> sommelier -> xwayland
```

JL2210:
I pushed a PR to fix Mesa. It's a matter of undoing a patch that fixes LLVM 6. For some reason, that patch broke LLVM 7.

JL2210:
See #3060.

PR #3067, opened 2019-01-08
Add pre-built binaries for packages g-x

uberhacker:
Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

- Added packages libva and libvdpau as dependencies of mesa.
- Most gui apps are still not working on i686 but that is somewhat expected.
- The gvim package needs work.  It will not start.  New issue pending.

### DIFF
--- a/packages/xinit.rb
+++ b/packages/xinit.rb
@@ -8,15 +8,22 @@ class Xinit < Package
   source_sha256 '230835eef2f5978a1e1344928168119373f6df1d0a32c09515e545721ee582ef'
 
   binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/xinit-1.4.0-1-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/xinit-1.4.0-1-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/xinit-1.4.0-1-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/xinit-1.4.0-1-chromeos-x86_64.tar.xz',
   })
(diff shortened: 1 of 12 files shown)


packages/xinit.rb
```diff
@@ -8,14 +8,19 @@ class Xinit < Package
...
+    aarch64: '571b3b0688bb7013a186dc577d6a6ea5b036f08b398a42528512c8990b03f51b',
+     armv7l: '571b3b0688bb7013a186dc577d6a6ea5b036f08b398a42528512c8990b03f51b',
+       i686: '3feebdbc9b9138d8acc6cfcf36459581f6bae6e8e8f26fdbcf6945f66c66743c',
+     x86_64: '943fd77aebd9dc370ca8ba637493e39d4e92e018dcea4b03a0d69566be64cb34',
   })
 
-  depends_on 'xclock'
-  depends_on 'xterm'
   depends_on 'xauth'
-  depends_on 'twm'
```
JL2210: So, I take it you don't care about runtime errors? These are **REQUIRED**. Try running `startx` without them. Since `xinit` is not depended on by any package, it shouldn't matter whether we pull in any more dependencies than "needed". It's the user's choice to install it.

uberhacker: I do care about runtime errors but there is something fundamental about this that you don't understand.  The twm, xclock and xterm apps are all **OPTIONAL** and not required.  If you don't believe me, try editing the last few lines of `/usr/local/etc/X11/xinit/xinitrc` and comment out the xclock or xterm lines and see if you can still execute startx.  You can even swap out twm for another window manager if you want which is what I've been trying to tell you after you submitted this PR but apparently, you weren't listening.  However, I went ahead and put these "dependencies" back in to keep people like you from being confused.

skiman6010: Sounds good to me.

JL2210:
Thanks, @uberhacker :).

PR #3068, opened 2019-01-09
Add xhost

skiman6010:
Not sure how needed this is, but I tried using it to fix VSCode and it didn't work but might as well add it into the repo in case anyone else needs it. 

Description: "Server access control program for X."
https://github.com/freedesktop/xorg-xhost

### DIFF
--- a/packages/xhost.rb
+++ b/packages/xhost.rb
@@ -1,0 +1,23 @@
+require 'package'
+
+class Xhost < Package
+	description 'Server access control program for X'
+	homepage 'https://github.com/freedesktop/xorg-xhost'
+	version '1.0.7'
+	source_url 'https://github.com/freedesktop/xorg-xhost/archive/xhost-1.0.7.tar.gz'
+	source_sha256 '1dd87b189cdd2921aec225eae630cc21113164b9a8654d1ed99613f3cda7be6a'
+
+	def self.build
+		system "./autogen.sh", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}"
+		system "make"
+	end
+
+	def self.install
+		system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+	end
+
+	def self.check
+		system "make", "check"
+	end
+end
+
--- a/packages/sublime_merge.rb
+++ b/packages/sublime_merge.rb
@@ -3,11 +3,11 @@ require 'package'
 class Sublime_merge < Package
   description 'A snappy UI, three-way merge tool,side-by-side diffs, syntax highlighting, and more.'
   homepage 'https://www.sublimemerge.com/'
-  version 'Build 1065'
+  version 'Build 1097'
   case ARCH
   when 'x86_64'		  
-	source_url 'https://download.sublimetext.com/sublime_merge_build_1065_x64.tar.xz'
-	source_sha256 'ebe9029ea71543a101dcfd09c5540759759c1e33fea31d4b98256803b7ceaf97'
+	source_url 'https://download.sublimetext.com/sublime_merge_build_1097_x64.tar.xz'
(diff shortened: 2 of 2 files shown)


JL2210:
Great! I looked for this package for the LONGEST time. You wouldn't believe
it. Thanks.

packages/xhost.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Xhost < Package
+	description 'Server access control program for X'
```
JL2210: Are these tabs? I can't exactly tell on this, but that's what they look like. Could you change them to two spaces?

packages/xhost.rb
```diff
@@ -0,0 +1,23 @@
...
+	homepage 'https://github.com/freedesktop/xorg-xhost'
+	version '1.0.7'
+	source_url 'https://github.com/freedesktop/xorg-xhost/archive/xhost-1.0.7.tar.gz'
```
JL2210: Pleas change `source_url` to `https://www.x.org/releases/individual/app/xhost-1.0.7.tar.bz2`.

packages/xhost.rb
skiman6010: You know, here's the thing about this. So I originally used vim as my only text editor and had it set up how I like it. It would convert my tabs to spaces and make everything pretty. Then uberhacker updated the vim package and broke all my stuff, but that's when I had VSCode working so it was no big deal because I had that set up how I like it and it would do the same thing with tabs to spaces... BUT now everyone broke both vim and VSCode so I had to resort to using Caret and it's a little frustrating. So basically I will convert these, I just wanted to explain to you why my life is the way it is right now.

packages/xhost.rb
```diff
@@ -0,0 +1,23 @@
...
+	source_sha256 '1dd87b189cdd2921aec225eae630cc21113164b9a8654d1ed99613f3cda7be6a'
+
+	def self.build
+		system "./autogen.sh", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}"
```
JL2210: Please change `autogen.sh` to `configure`.

packages/xhost.rb
JL2210: OK. I use `nano`, but it's a tad simple for "expert" users. :)

packages/xhost.rb
uberhacker: Not a good idea to use a homepage with a version in it.  His homepage is fine.

packages/xhost.rb
uberhacker: This is actually needed since configure is not available.

packages/xhost.rb
```diff
@@ -0,0 +1,23 @@
...
+		system "make"
+	end
+
+	def self.install
+		system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+	end
+
+	def self.check
+		system "make", "check"
```
uberhacker: :+1: for the self.check section. 🥇

packages/xhost.rb
JL2210: In the release tarball from `x.org`, the `configure` script is available. See above.

packages/xhost.rb
JL2210: Yeah. Let's just see if it works. :P

JL2210:
@skiman6010 Are you still working on this? It would be lovely to be able to run X applications as a user other than `chronos`. I'm thinking maybe GParted (as `root`)?

packages/xhost.rb
JL2210: `X11R6.8.1` is just the version of the X server. No correlation to the package used. In fact, I believe that's the only version of the X server that has that manpage. It was split out into a separate package in later versions.

JL2210:
Is the update to Sublime Merge supposed to be included in this PR?

skiman6010:
You can do it since you're so picky about all these unimportant things. @JL2210

JL2210:
All right, I'll give it a shot.

PR #3071, opened 2019-01-09
Remove CRLFs and replace with LFs in all packages

uberhacker:
This PR fixes the result of building packages in Windows which should never be used as a development environment for Chromebrew...ever. :)

### DIFF
--- a/packages/libdmx.rb
+++ b/packages/libdmx.rb
@@ -22,7 +22,7 @@ class Libdmx < Package
 
   depends_on 'libxext'
   depends_on 'libx11'
-  
+
   def self.build
     system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
     system "make"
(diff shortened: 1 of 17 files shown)


JL2210:
Wow. Maybe they used Wine or ReactOS? I can't imagine a Chromebrew user
doing this.

skiman6010:
Haha this is funny. 

Thanks @uberhacker!

JL2210:
Did you run `dos2unix`? That command is simply amazing.

uberhacker:
Macs are poo-poo'd also since they use on CRs instead of LFs.  No reason to not develop on a Chromebook if creating Chromebrew packages.  How else would you test?

sajjaadf:
What is a CR and a LF?

uberhacker:
CR = carriage return (MacOS style)
LF = line feed (Unix/Linux style)
CRLF = carriage return + line feed (Windows/DOS style)

@JL2210: Yes, I did use dos2unix.

sajjaadf:
Oh

JL2210:
@sajjaadf Read above explanation by @uberhacker.

PR #3073, opened 2019-01-10
Update ghpages

sajjaadf:
Fix a little problem


uberhacker:
Uh, do we really need this?  The chances of somebody going here accidentally are pretty slim.

JL2210:
Yeah... I agree with @uberhacker.

sajjaadf:
can't we do it for fun?

sajjaadf:
It won't hurt, _right_?

uberhacker:
But here's the thing: they might be in the right place intentionally and this message could end up being more confusing than helpful.

PR #3075, opened 2019-01-11
Update filezilla from 3.38.1 to 3.39.0

uberhacker:
Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64


JL2210:
Darn. I was going to do this, too.

uberhacker:
If you are "looking" for things to do, the code package is still dead in the water.  That really should be a high priority right now.  Lots of users want to use code.

JL2210:
And dead in the water it will stay. That program is probably broken. After
all, it was written by the company that created Windows.

sajjaadf:
Use sublime its better

-- SajjaadF

PR #3076, opened 2019-01-12
Add gyp package

uberhacker:
GYP is a Meta-Build system: a build system that generates other build systems.  See https://gyp.gsrc.io/.

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/gyp.rb
+++ b/packages/gyp.rb
@@ -1,0 +1,28 @@
+require 'package'
+
+class Gyp < Package
+  description 'GYP is a Meta-Build system: a build system that generates other build systems.'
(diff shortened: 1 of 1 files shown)


JL2210:
If you're planning on updating Sommelier, it's not going to work without
MAJOR hackage. Trust me, I tried. Your best bet would be to build the
entire 'platform2' tree in Chromium, which just happens to be above the
storage size of my 2 terabyte hard drive.

JL2210:
Other than that, not bad.

PR #3077, opened 2019-01-13
Update Sed package

JL2210:
Tested on:

- [x] x86_64

Adds `def self.check` section.


JL2210:
@uberhacker, SELinux tests are disabled if you are building without SELinux automatically, so your little "Remove selinux tests" section is unneeded.

JL2210:
Yes, it was. They passed when I created that pre-built binary.

JL2210:
All right. Every platform is different. If that was needed on some other Chromebooks, go right ahead and keep it in there. It's not harming anything.

(Darn Autocorrect)

PR #3078, opened 2019-01-13
Update GDB package

JL2210:
Enables new features in GDB (Python, X, Readline, Ncurses)

Enables GDB Simulator

Builds-in Chromebrew branding

Enables GDB TUI and CLI

Makes `configure` somewhat consistent with binutils

### DIFF
--- a/packages/gdb.rb
+++ b/packages/gdb.rb
@@ -3,28 +3,40 @@ require 'package'
 class Gdb < Package
   description 'GDB, the GNU Project debugger, allows you to see what is going on \'inside\' another program while it executes -- or what another program was doing at the moment it crashed.'
   homepage 'https://www.gnu.org/software/gdb/'
-  version '8.1'
-  source_url 'https://ftpmirror.gnu.org/gdb/gdb-8.1.tar.xz'
-  source_sha256 'af61a0263858e69c5dce51eab26662ff3d2ad9aa68da9583e8143b5426be4b34'
+  version '8.2.1'
+  source_url 'https://ftpmirror.gnu.org/gdb/gdb-8.2.1.tar.xz'
+  source_sha256 '0a6a432907a03c5c8eaad3c3cffd50c00a40c3a5e3c4039440624bae703f2202'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gdb-8.1-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/gdb.rb
```diff
@@ -3,28 +3,40 @@
...
+           '--with-python',
+           '--with-curses',
+           '--enable-shared',
+           '--with-system-zlib',
+           '--enable-64-bit-bfd',
+           '--enable-host-shared',
+           '--with-system-readline',
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}",
+           '--with-pkgversion=Chromebrew'
```
uberhacker: Why do we need this?  Shouldn't a version have a number?

packages/gdb.rb
```diff
@@ -3,28 +3,40 @@
...
+           '--with-x',
+           '--with-lzma',
+           '--enable-sim',
+           '--with-expat',
+           '--enable-tui',
+           '--with-python',
+           '--with-curses',
+           '--enable-shared',
+           '--with-system-zlib',
+           '--enable-64-bit-bfd',
```
uberhacker: Isn't this for x86_64 systems only?  What about arm and i686?

packages/gdb.rb
JL2210: No. If we do not set this, `pkgversion` would be set as `GDB`. If this was not set, the first line outputted when executing `gdb` would be `GNU gdb (GDB) 8.2.1`, which is redundant and does not describe that the package is provided by Chromebrew. If the option is re-added, the first line of output would be `GNU gdb (Chromebrew) 8.2.1`.

packages/gdb.rb
uberhacker: Wouldn't it be better to leave out the option if it's enabled by default?  Otherwise, you are explicitly enabling it on non-64-bit systems.

JL2210: The option is useful for cross-compiling to target a 64-bit system, but has other uses too (among them, one is performance).

uberhacker: We're not cross-compiling so there is no need to set this option.

JL2210: We may in the future. I don't want to close that door off for other people. As I said before, this option was introduced for compatibility with libbfd from binutils, as two versions of libbfd compiled with different options cause... problems.

JL2210: It causes no harm, so there is no reason to remove it.

uberhacker: You have never attempted to compile on other systems so how can you be so sure?

PR #3081, opened 2019-01-13
Update tests.

JL2210:
Allows for identifying more than one dependency cycle in `no_cycles.rb` and removes `All tests successful` from `test_all`.

### DIFF
--- a/tests/no_cycles.rb
+++ b/tests/no_cycles.rb
@@ -20,17 +20,18 @@ end
 # @state will store :on_path for vertices on the current dependency path
 # and :visited for vertices that have already been checked not to lead to
 # cycles.
+@failed = 0
 @state = {}
 @path = []
 def dfs(pkg)
   @path.push(pkg.name)
   if @state[pkg] == :on_path
-    puts "Found dependency cycle!".lightred
+    puts "\nFound dependency cycle!".lightred
     while @path.first != @path.last
       @path.shift
     end
     puts @path.to_s
-    exit 1
+    @failed += 1
   elsif @state[pkg] == nil
     @state[pkg] = :on_path
     if pkg.dependencies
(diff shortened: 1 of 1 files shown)


tests/no_cycles.rb
```diff
@@ -50,4 +51,9 @@ def dfs(pkg)
   dfs(pkg)
 end
 
-puts "No dependency cycles found.".lightgreen
+if @failed == 0 || @failed > 1
+  @cycles = "cycles"
+else
+  @cycles = "cycle"
+end
```
uberhacker:
```
@cycles = "cycles"
@cycles = "cycle" if @failed == 1
```
2 lines  instead of 5.  1 conditional instead of 2.

tests/no_cycles.rb
```diff
@@ -50,4 +51,9 @@ def dfs(pkg)
...
+puts "#{@failed} dependency #{@cycles} found.".lightgreen
```
uberhacker: `abort` instead of `puts` and you don't have to delete the successful message below.

tests/no_cycles.rb
JL2210: Good idea.

uberhacker: What would make this even better is if you identified which dependencies of each package were causing the cycles.

JL2210: It does, it's just not obvious. This line is key: `    puts @path.to_s`

tests/no_cycles.rb
JL2210: Done.

tests/no_cycles.rb
uberhacker: Well it prints something but what if it displayed the package name along with the dependency for clarity?

PR #3083, opened 2019-01-14
Update LZO package

JL2210:
Force shared libs


uberhacker:
What is wrong with static libraries?  Do they get in the way?

PR #3085, opened 2019-01-14
Update php from 5.6.39-7.3.0 to 5.6.40-7.3.1

uberhacker:
- Add more dependencies and configure options
- Fix for tidy support
- Add libzip dependency
- Update php7 to 7.1.26
- Update php7 to 7.2.14
- Update php7 to 7.3.1
- Add pre-built binaries

Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/libzip.rb
+++ b/packages/libzip.rb
@@ -1,0 +1,47 @@
+require 'package'
+
+class Libzip < Package
+  description 'libzip is a C library for reading, creating, and modifying zip archives.'
+  homepage 'https://libzip.org/'
+  version '1.5.1'
+  source_url 'https://libzip.org/download/libzip-1.5.1.tar.xz'
+  source_sha256 '04ea35b6956c7b3453f1ed3f3fe40e3ddae1f43931089124579e8384e79ed372'
+
+  binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/libzip-1.5.1-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/libzip-1.5.1-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/libzip-1.5.1-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/libzip-1.5.1-chromeos-x86_64.tar.xz',
+  })
+  binary_sha256 ({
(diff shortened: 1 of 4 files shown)


JL2210:
Isn't this whole PHP thing getting out of hand? Just provide `php5`, `php6`, and `php7` and make `php` a version selector.

packages/libzip.rb
```diff
@@ -0,0 +1,47 @@
...
+       i686: 'd286172ecf333bed71e0bd1c7713cbbefb5a5434e2592ee5e67cd881bccb5784',
+     x86_64: '4d1fa9fad684b7e4aa92885e247e963ac7ad1a1c85fd7a35cec728491a2e72cc',
+  })
+
+  depends_on 'bz2'
+
+  def self.build
+    Dir.mkdir 'build'
+    Dir.chdir 'build' do
+      system "cmake \
```
JL2210: Can you use `",` instead of backslashes? Same for all of these below, please.

JL2210: It makes for an unpleasant error when you use backslashes.
Example:
```
'foo' failed to install: `really                     long                   configure                         command` exited with 1
```

uberhacker: So here's the thing.  I actually prefer this method to the commas because if you ever need to set some environment variables, this is the most readable way to do it.  Granted this package doesn't need any environment variables but sometimes cmake requires them and using commas doesn't work.

JL2210: Just use `env`. It's a lot prettier, and there is no backslashing required.

uberhacker:
There is no php version 6.  Also, php5 is nearing EoL.  We need the php5 and php7 packages to augment the php package.

packages/libzip.rb
uberhacker: The unpleasant error output with lots of spaces is a fair argument for not using this method.  Why didn't you suggest the commas with the php packages?

skiman6010: It's fine, it can get fixed later.  Not a big deal.

JL2210: I did, though I imagine it didn't come across that way.

JL2210:
You can just have `php` depend on `php5` or `php7`, depending on what was selected for install.

uberhacker:
But what you don't understand is there are 3 different versions of php 7.  We can only have 1 package.

JL2210:
Why are they needed? Just put a warning in and slowly remove them after
time.

skiman6010:
I use different php versions depending on what local dev environment I need since WordPress is awful. I'm the 1% who uses this package constantly.

JL2210:
So what do you use it for? I'm confused at exactly what this does.

PR #3086, opened 2019-01-15
Add pre-built binaries for gdb, lzo, lzop and xhost

uberhacker:
Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

- xhost freezes on i686 but that is not really surprising


JL2210:
What I don't understand is what's so bad about that hompage. Why do you have to change it?

JL2210:
And you had me under the impression that their README was descriptive. It contains one sentence about `xhost`, no usage, and then the rest is links to mailing lists and stuff. :-1:

skiman6010:
My word everyone chill out about the homepage. It doesn't matter at all.

PR #3087, opened 2019-01-15
Add linux_sources package and recreate linuxheaders package

JL2210:
Installs native Linux headers instead of installing 3.18.0 headers on all systems. Allows for building modules.


JL2210:
Please create binaries before merging, as the build fails if no compiler is present.

JL2210:
Let me work on this for a little bit more.

JL2210:
Still needs testing, but looks good now. I'm reinstalling.

JL2210:
This works on a fresh Chromebrew install, @uberhacker.

uberhacker:
I'm really wary of merging this since it's so critical to fresh installs.

JL2210:
It works. You can test if you'd like.

JL2210:
@uberhacker, anything that needs extra work with this?

uberhacker:
Why do we need sources?  Do you actually think we can compile a new kernel and install it?

JL2210:
No way! I have two more reasons for you:
1. Building kernel modules
A lot of packages build kernel modules and require kernel sources to be available. This fixes that.
2. Building UML (User-Mode Linux)
This would allow users to install User-Mode Linux, which is a Linux kernel that compiles as a program and works somewhat like a virtual machine, except faster. It's very useful for debugging and test cases. For more information, see [their SourceForge page](http://user-mode-linux.sourceforge.net/).

uberhacker:
This UML thing looks like good stuff.  I wonder if we can use it to cross-compile?

JL2210:
Sadly, no. It can only be used to run a native operating system.

uberhacker:
I just tried this and it adds 5 min to the install time. :/  It also installs a bunch of architectural specific crap my Chromebook would never need such as arm, mips, m68k, powerpc, etc.  When you say a lot of packages build kernel modules, can you say which specific packages require this?  I think this needs some more work.  At least try and trim it down to only install the files needed by the system architecture and nothing else.

JL2210:
Yes, that's because it's downloading the full sources.

If we don't include all the "architectural specific crap", we wouldn't be providing the full kernel sources, and the compile would probably fail because of a missing file. Also, we can't exclude the files from the download.

I can state one that I'm actively working on: VirtualBox.

How so?

Read above.

uberhacker:
Granted you cannot exclude when extracting the sources but you can filter when adding files to `#{CREW_DEST_PREFIX}`.  The only kernel source files needed should be those compatible with the architecture.

JL2210:
To be honest, I don't want to filter any files out. If somebody wants to install the kernel sources, I imagine that they would expect the full kernel sources. Anyway, with all the `Kconfig` files everywhere, I know that it'll be extremely painful to try and remove anything without breaking something.

JL2210:
So, what do you say, @uberhacker?

uberhacker:
I'm hesitant.  It adds 5 mins to the install with little value added.  Maybe you can include it as a dependency with the packages that really need it?

JL2210:
I'm sorry, but I can't really help that. The `linuxheaders` package we have now is just sort of thrown together, and doesn't have any sources.

> with little value added.

Not true at all. For one, it installs the kernel headers specific to the Chromebook you're working on. For another, it allows building kernel modules, which are very useful for things such as, say, virtualization?

> Maybe you can include it as a dependency with the packages that really need it?

That would break all sorts of things. The headers would be overwritten by newer versions, and the dependency tree would break a lot of packages that use Linux headers. For example, if you remove `linux_sources` after installing it (after making your suggested changes), even the simplest "Hello, World!" program would fail to compile.

skiman6010:
First off, why has your writing changed so much between your normal self and now? Just curious, it's a little creepy. 

Anyway, I'm not convinced this will be useful to enough people to add five minutes at the minimum to a fresh install. That's a long time. That's why we are so hesitant to merge this, you haven't given us good enough reasons to warrant an extra five minutes to the install. We try to keep the installs as minimal and clean as possible, so we want to know in more detail what benefits there are.

JL2210:
Read all my comments on this. What more do you need?

uberhacker:
We need less.  That would be more.

JL2210:
I was hoping on removing all references to `https://bintray.com/chromebrew/chromebrew-cross` from the GitHub repository so we could go ahead and remove it from Bintray.

uberhacker:
That is great but can you please trim down this package?  We don't need all these extra architectures in the source that take up valuable disk space.  Remember, most Chromebooks have a limited size to the hard drive.

uberhacker:
How about this idea?  Just update the linuxheaders package and put the sources minus headers in linux_sources that can be excluded from core?

JL2210:
Won't work. The sources and the kernel headers have to be the same version as the running kernel.

uberhacker:
You mean it won't work because you have already tried it or it won't work because you don't want to try it?

JL2210:
I already tried it.

JL2210:
I could probably add this to the package to the kernel modules will work on all the Chromebooks:
https://raw.githubusercontent.com/D1W0U/ARP-RootKit/master/patch-lkm.py

JL2210:
I think I have a solution: If I change this to use 3.18 exclusively, will it be OK?

uberhacker:
The whole problem with this PR is it adds too much fat to a core package with little added value.  If you split it into 2 separate packages (1 with just the headers for the 3 architectures and the other with the remaining kernel source to be a dependency package for _ONLY_ those that need it), I could accept that.  We also don't want to increase the time required to install fresh, if possible.

JL2210:
All right. I'm on it.

PR #3088, opened 2019-01-16
Update install.sh

JL2210:
quote variables,
respect spaces in install names,
respect choice of branch,
fix logic,
suggest simpler user-friendly editor,
and exit on failure

Update install.sh to help with work for next PR

Tested and working on fresh VM. (I didn't want to Powerwash and lose all my work)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,24 +1,31 @@
+#!/bin/bash
+
+# exit on fail
+set -e
+
 #chromebrew directories
 OWNER="skycocker"
 REPO="chromebrew"
 BRANCH="master"
-URL="https://raw.githubusercontent.com/$OWNER/$REPO/$BRANCH"
-CREW_PREFIX=${CREW_PREFIX:-/usr/local}
-CREW_LIB_PATH=$CREW_PREFIX/lib/crew/
-CREW_CONFIG_PATH=$CREW_PREFIX/etc/crew/
-CREW_BREW_DIR=$CREW_PREFIX/tmp/crew/
-CREW_DEST_DIR=$CREW_BREW_DIR/dest
-CREW_PACKAGES_PATH=$CREW_LIB_PATH/packages
+URL="https://raw.githubusercontent.com/${OWNER}/${REPO}/${BRANCH}"
+CREW_PREFIX="${CREW_PREFIX:-/usr/local}"
+CREW_LIB_PATH="${CREW_PREFIX}/lib/crew/"
+CREW_CONFIG_PATH="${CREW_PREFIX}/etc/crew/"
+CREW_BREW_DIR="${CREW_PREFIX}/tmp/crew/"
+CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
+CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 
-architecture=$(uname -m)
+ARCH="$(uname -m)"
 
-if [ $EUID -eq 0 ]; then
-  echo 'Chromebrew should not be run as root.'
+if [ "${EUID}" == "0" ]; then
+  echo 'Chromebrew should not be installed or run as root.'
   exit 1;
 fi
 
-case "$architecture" in
+case "${ARCH}" in
 "i686"|"x86_64"|"armv7l"|"aarch64")
+  LIB_SUFFIX=
+  [ "${ARCH}" == "x86_64" ] && LIB_SUFFIX='64'
   ;;
 *)
   echo 'Your device is not supported by Chromebrew yet.'
@@ -26,18 +33,20 @@ case "$architecture" in
 esac
 
 # This will allow things to work without sudo
-sudo chown -R `id -u`:`id -g` "${CREW_PREFIX}"
+sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}"
 
 # prepare directories
-for dir in $CREW_LIB_PATH $CREW_CONFIG_PATH $CREW_CONFIG_PATH/meta $CREW_BREW_DIR $CREW_DEST_DIR $CREW_PACKAGES_PATH; do
-  mkdir -p $dir
+for dir in "${CREW_CONFIG_PATH}/meta" "${CREW_DEST_DIR}" "${CREW_PACKAGES_PATH}"; do
+  if [ ! -d "${dir}" ]; then
+    mkdir -p "${dir}"
+  fi
 done
 
 # prepare url and sha256
 # install only gcc7, ruby, libiconv, git and libssh2 (order matters)
 urls=()
 sha256s=()
-case "$architecture" in
+case "${ARCH}" in
 "aarch64")
   urls+=('https://dl.bintray.com/chromebrew/chromebrew/gcc7-7.3.0-3-chromeos-armv7l.tar.xz')
   sha256s+=('c9dd2a2b2d195f0f2021dc673d6317e46b5dfaef582f496b4cea1f3c5c2c17ad')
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -94,133 +104,125 @@ esac
...
-git remote add -f origin https://github.com/$OWNER/$REPO.git
+git remote add -f origin "https://github.com/${OWNER}/${REPO}.git"
 git config core.sparsecheckout true
 echo packages >> .git/info/sparse-checkout
 echo lib >> .git/info/sparse-checkout
 echo crew >> .git/info/sparse-checkout
-git fetch origin master
-git reset --hard origin/master
+git fetch origin "${BRANCH}"
+git reset --hard origin/"${BRANCH}"
```
uberhacker: Where is `${BRANCH}` getting the value?

install.sh
```diff
@@ -94,133 +104,125 @@ esac
...
+echo "echo \"export PAGER='most'\" >> ~/.bashrc && . ~/.bashrc"
 echo
-echo "You may wish to set the default EDITOR.  The current EDITOR is '$EDITOR'."
-echo "For example, to set vim as the default editor, execute the following:"
-echo "echo \"export EDITOR=$CREW_PREFIX/bin/vim\" >> ~/.bashrc && . ~/.bashrc"
-echo "To install the vim editor, run 'crew install vim'"
+echo "You may wish to set the default EDITOR.  The current EDITOR is '${EDITOR}'."
+echo "For example, to set nano as the default editor, execute the following:"
+echo "echo \"export EDITOR='nano'\" >> ~/.bashrc && . ~/.bashrc"
+echo "To install the simple nano editor, run 'crew install nano'"
```
uberhacker: The default `EDITOR` is already nano so that is why we added the vim example.  No need to change to nano since it would be redundant.

install.sh
```diff
@@ -94,133 +104,125 @@ esac
...
 
 # install a base set of essential packages
 yes | crew install buildessential less most
 
+echo
+echo "If you have installed Chromebrew in a directory other than '/usr/local',"
+echo "please run these commands:"
+echo "echo 'export CREW_PREFIX=${CREW_PREFIX}' >> ~/.bashrc && . ~/.bashrc"
+echo "echo 'export PATH=\"\${CREW_PREFIX}/bin:\${CREW_PREFIX}/sbin:\${PATH}\"' >> ~/.bashrc && . ~/.bashrc"
+echo "echo 'export LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}' >> ~/.bashrc && . ~/.bashrc"
```
uberhacker: Where is `${LIB_SUFFIX}` getting the value?

install.sh
JL2210: Up at the top where most of the variables are set.

install.sh
JL2210: Oh. I suppose that I could change that.

install.sh
JL2210: Where the valid architectures are determined. I separated out `x86_64` and added `LIB_SUFFIX=64`.

JL2210: First of all, empty environment variables are expanded to empty strings. Second of all, if users want to set `LIB_SUFFIX`, they should be able to.

install.sh
```diff
@@ -1,43 +1,53 @@
...
   exit 1;
 fi
 
-case "$architecture" in
-"i686"|"x86_64"|"armv7l"|"aarch64")
+case "${ARCH}" in
+"i686"|"armv7l"|"aarch64")
+  ;;
+"x86_64")
+  LIB_SUFFIX=64
```
uberhacker: This could be simpified:
```
LIB_SUFFIX=''
[ "${ARCH}" == "x86_64" ] && LIB_SUFFIX='64'
```

JL2210: I don't think that would work with `set -e`.

uberhacker: Nevermind, this is checking for hardware.

JL2210: Done.

install.sh
uberhacker: I would recommend vim over nano but whatever.  An editor is a preference and should not be recommended.

install.sh
uberhacker: I don't think setting LIB_SUFFIX=64 for x86_64 is giving users the ability to set what they want.

install.sh
JL2210: Yeah, although `nano` is easier to understand. A newbie user wouldn't be able to use `vim`.

install.sh
JL2210: All right. Good point. It should just be an internal thing.

install.sh
```diff
@@ -94,133 +104,125 @@ esac
...
+echo "echo 'export PATH=\"\${CREW_PREFIX}/bin:\${CREW_PREFIX}/sbin:\${PATH}\"' >> ~/.bashrc && . ~/.bashrc"
+echo "echo 'export LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}' >> ~/.bashrc && . ~/.bashrc"
 echo
 echo "To set the default PAGER environment variable to be able to use less:"
-echo "echo \"export PAGER=$CREW_PREFIX/bin/less\" >> ~/.bashrc && . ~/.bashrc"
+echo "echo \"export PAGER='less'\" >> ~/.bashrc && . ~/.bashrc"
 echo
 echo "Alternatively, you could use most.  Why settle for less, right?"
-echo "echo \"export PAGER=$CREW_PREFIX/bin/most\" >> ~/.bashrc && . ~/.bashrc"
+echo "echo \"export PAGER='most'\" >> ~/.bashrc && . ~/.bashrc"
```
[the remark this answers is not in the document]
uberhacker: Yes.  First of all, users can change their `PATH` from the default and secondly, if ChromeOS decides to add one of these commands in future releases, this assures the desired command is executed.

install.sh
JL2210: Anyhow, I already changed this. See above (in the file).

install.sh
JL2210: I was having some troubles setting `${BRANCH}` when I was testing out `linux_sources`. This fixes that.

install.sh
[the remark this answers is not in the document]
JL2210: `/usr/local/bin` is first in `${PATH}`. That guarantees that the locally-installed copy will be used first.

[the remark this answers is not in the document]
JL2210: If they want to change their `${PATH}`, let them do so. I'm not stopping them.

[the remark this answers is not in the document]
JL2210: If Chrome OS does choose to add one of these commands, it shouldn't make a difference. `most` is `most` and `less` is `less`.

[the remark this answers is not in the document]
uberhacker: Except in cases where the system version of less/most in future releases is not updated like, hopefully, the Chromebrew version.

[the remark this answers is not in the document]
JL2210: Keep in mind that this is all speculative. What use would a browser-based OS have for a pager?

install.sh
```diff
@@ -1,43 +1,52 @@
...
+
+if [ "${EUID}" == "0" ]; then
+  echo 'Chromebrew should not be installed or run as root.'
   exit 1;
 fi
 
-case "$architecture" in
+case "${ARCH}" in
 "i686"|"x86_64"|"armv7l"|"aarch64")
+  LIB_SUFFIX=
```
uberhacker: This should be assigned outside the case statement, otherwise it is not initialized for other architectures.

uberhacker: Nevermind.  I see you included all architectures already.

install.sh
```diff
@@ -1,43 +1,52 @@
...
+  [ "${ARCH}" == "x86_64" ] && LIB_SUFFIX='64'
   ;;
 *)
   echo 'Your device is not supported by Chromebrew yet.'
   exit 1;;
 esac
 
 # This will allow things to work without sudo
-sudo chown -R `id -u`:`id -g` "${CREW_PREFIX}"
+sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}"
```
uberhacker: This is definitely a better way to get these values.

install.sh
```diff
@@ -1,43 +1,52 @@
...
 
 # prepare directories
-for dir in $CREW_LIB_PATH $CREW_CONFIG_PATH $CREW_CONFIG_PATH/meta $CREW_BREW_DIR $CREW_DEST_DIR $CREW_PACKAGES_PATH; do
-  mkdir -p $dir
+for dir in "${CREW_CONFIG_PATH}/meta" "${CREW_DEST_DIR}" "${CREW_PACKAGES_PATH}"; do
```
uberhacker: Are you sure the `CREW_LIB_PATH` and `CREW_CONFIG_PATH` directories are not necessary?

install.sh
```diff
@@ -94,133 +103,128 @@ esac
 
 # functions to maintain packages
 function download_check () {
-    cd $CREW_BREW_DIR
+    cd "${CREW_BREW_DIR}"
 
     #download
-    echo "Downloading $1..."
-    curl --progress-bar -C - -# -L --ssl $2 -o "$3"
```
uberhacker: Yeah, this random `#` showing up in the output was annoying. :+1:

install.sh
```diff
@@ -94,133 +103,128 @@ esac
...
+    echo "Downloading ${1}..."
+    curl --progress-bar -C - -L --ssl "${2}" -o "${3}"
 
     #verify
-    echo "Verifying $1..."
-    echo $4 $3 | sha256sum -c -
-    case $? in
+    echo "Verifying ${1}..."
+    echo "${4}" "${3}" | sha256sum -c -
+    case "${?}" in
```
uberhacker: This should always return a number so quoting is not necessary.  In fact, since `0` is used below, it may not work.

install.sh
```diff
@@ -94,133 +103,128 @@ esac
...
+      rm -rf "${dir}"
+    done
+    for file in ./*list; do
+      rm -f "${file}"
+    done
+    tar xpf ../"${2}"
+    echo "Installing ${1} (this may take a while)..."
+    tar cpf - ./*/* | (cd /; tar xp --keep-directory-symlink -f -)
+    mv ./dlist "${CREW_CONFIG_PATH}/meta/${1}.directorylist"
+    mv ./filelist "${CREW_CONFIG_PATH}/meta/${1}.filelist"
```
uberhacker: I don't understand why this section was changed.

install.sh
```diff
@@ -94,133 +103,128 @@ esac
...
 }
 
 function update_device_json () {
-  cd $CREW_CONFIG_PATH
+  cd "${CREW_CONFIG_PATH}"
 
-  if grep '"name": "'$1'"' device.json > /dev/null; then
-    echo "Updating version number of existing $1 information in device.json..."
-    sed -i device.json -e '/"name": "'$1'"/N;//s/"version": ".*"/"version": "'$2'"/'
+  if grep '"name": "'"${1}"'"' device.json > /dev/null; then
```
uberhacker: Quoting is not necessary here and only makes things more confusing.

install.sh
```diff
@@ -94,133 +103,128 @@ esac
...
+    echo "Updating version number of ${1} in device.json..."
+    sed -i device.json -e '/"name": "'"${1}"'"/N;//s/"version": ".*"/"version": "'"${2}"'"/'
```
uberhacker: Quoting overload here too.

install.sh
```diff
@@ -94,133 +103,128 @@ esac
...
   elif grep '^    }$' device.json > /dev/null; then
-    echo "Adding new $1 information to device.json..."
+    echo "Adding new information on ${1} to device.json..."
```
uberhacker: How about?
```
echo "Adding new information for package '${1}' to device.json..."
```
This makes sure if the package name is an ordinary english word, there is no confusion.

install.sh
```diff
@@ -94,133 +103,128 @@ esac
...
     }/'
   else
-    echo "Adding new $1 information to device.json..."
+    echo "Adding new information on ${1} to device.json..."
     sed -i device.json -e '/^  "installed_packages": \[$/s/$/\
     {\
-      "name": "'$1'",\
-      "version": "'$2'"\
+      "name": "'"${1}"'",\
+      "version": "'"${2}"'"\
```
uberhacker: Quote overload.

install.sh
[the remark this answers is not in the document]
JL2210: True, but still: Why would Chrome OS *include* a pager if it's only a browser? They're probable not going to add one. Anyway, like I said before, Chromebrew will almost certainly be in front of the `PATH` if people follow instructions.

install.sh
JL2210: If a user's name has spaces in it, this is most definitely better.

install.sh
JL2210: Yes. Those directories are created recursively with the `-p` option to `mkdir`.

install.sh
JL2210: I don't like the idea of unquoted variables. All it does is add a pair of `"`.

install.sh
JL2210: I don't think `libssh2` or `libiconv` are ordinary English words.

install.sh
JL2210: See above.

install.sh
JL2210: See above.

install.sh
JL2210: Because there might be other directories than just `/usr` and `/home` that have files in them in the future, and because we want to preserve permissions for those.

install.sh
JL2210: Quotes aren't taken into account in the `case` statement itself. It's just that it consolidates the value of `${?}` into one argument, if it contains spaces. I'm not taking any chance here.

install.sh
JL2210: Yeah, the `--progress-bar` long option is much better than `-#`.

install.sh
uberhacker: Have you ever tried creating a user with spaces?  I don't think you will have much success.  Quotes are not necessary.

install.sh
```diff
@@ -94,133 +103,128 @@ esac
...
+    echo "Extracting ${1} (this may take a while)..."
+    for dir in ./*/; do
+      rm -rf "${dir}"
+    done
+    for file in ./*list; do
+      rm -f "${file}"
+    done
+    tar xpf ../"${2}"
+    echo "Installing ${1} (this may take a while)..."
+    tar cpf - ./*/* | (cd /; tar xp --keep-directory-symlink -f -)
```
uberhacker: This might be good enough:
```
tar cpf - * | (cd /; tar xp --keep-directory-symlink -f -)
```

install.sh
```diff
@@ -94,133 +103,128 @@ esac
...
-    fi
-    mv ./dlist $CREW_CONFIG_PATH/meta/$1.directorylist
-    mv ./filelist $CREW_CONFIG_PATH/meta/$1.filelist
+    echo "Extracting ${1} (this may take a while)..."
+    for dir in ./*/; do
+      rm -rf "${dir}"
+    done
+    for file in ./*list; do
+      rm -f "${file}"
+    done
```
uberhacker: This might be more thorough and is a bit easier:
```
find . -type d -exec rm -rf {} \\;
find . -type f -exec rm -f {} \\;
```

install.sh
uberhacker: Uh, dude, the variables are already quoted.  Why do I get the idea you did a global search and replace blindly?

install.sh
uberhacker: Then why change it?

install.sh
JL2210: It's always better to quote variables. Just trust me here.
Anyhow, quotes aren't doing any harm.

install.sh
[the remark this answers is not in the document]
JL2210: Oh, and by the way, `EDITOR` is set to `/bin/nano`, which just happens to be nonexistent. I'm reverting this.

install.sh
JL2210: Oh, and by the way, `EDITOR` is set to `/bin/nano`, which just happens to be nonexistent. I'm reverting this.

install.sh
JL2210: The quotes are cancelled out. Currently, `${1}` was unquoted. Quote stacking is a tricky topic, but I have made sure this specific string works.

install.sh
uberhacker: Sounds perfectly fine to me.

JL2210: "Sounds" doesn't mean it's correct.

install.sh
uberhacker: `if grep '"name": "'${1}'"' device.json > /dev/null; then`

JL2210: Yes. The single quotes surrounded the double quotes and made them literal. The `${1}` was unquoted.

install.sh
JL2210: I'll go for a compromise: `    echo "Adding new information on package '${1}' to device.json..."`

install.sh
uberhacker: Then try this:
`find -type d -not -path . -exec rm -rf {} \\;`

install.sh
uberhacker: Nothing is made literal here.  Look at the line carefully.

install.sh
uberhacker: Actually, `for` is a better preposition than `on` here.  Didn't you learn these things in elementary school?

install.sh
JL2210: The double quotes are taken literally.
I have looked at it carefully.
If you'd like to keep arguing, go ahead but I know it's correct because I've tested it.

install.sh
uberhacker: Actually, the more I think about this, the more wrong I think the decision was to change this.

install.sh
JL2210: I don't appreciate being insulted, @uberhacker. Please look at the general channel in Slack before you post things like this.

install.sh
JL2210: Thank you. I'm glad to know my ideas go unappreciated.

install.sh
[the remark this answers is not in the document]
uberhacker: I change my PATH all the time and sometimes I add stuff in the front so it executes those things first.  The point is you shouldn't assume the PATH is set a certain way since it can be changed by the user at any time.

install.sh
JL2210: I'm sorry if I'm being disrespectful, but it just doesn't make me feel happy to work on this when the work that I spend a lot of time on is rejected, as a pose to your contributions, which are committed with almost no question.

install.sh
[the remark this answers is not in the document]
JL2210: Yes, and if a user chooses to, they should also have the right to change their pager. These are just *suggested defaults*.

install.sh
JL2210: Please see [Quoting](https://www.gnu.org/software/bash/manual/html_node/Quoting.html).

install.sh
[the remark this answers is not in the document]
uberhacker: Suggested defaults that were changed because you wanted to change them when they didn't need changed.

install.sh
uberhacker: I don't think you are listening.

install.sh
JL2210: Well, it's really user choice. I'll just leave it.

install.sh
uberhacker: Nothing is being rejected.  I was saying the idea to delete everything like we're doing here was the wrong idea.  There is stuff in this directory we want to save.

install.sh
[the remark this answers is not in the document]
JL2210: Yes, but it's not just that I want to change them. Users could change their `PATH` and yet their `PAGER` would stay the same. What if they removed their `CREW_PREFIX` but installed Chromebrew in another directory and added that to their `PATH` first?

install.sh
JL2210: I am, but I need to be listened to.
Please see [Quoting](https://www.gnu.org/software/bash/manual/html_node/Quoting.html).
Maybe then it will make more sense :).

install.sh
JL2210: Notice that I added `/dest` to the directory we `cd` into. I didn't want the `dest` directory to be removed, but I did want the files to be removed. If this fix isn't incorporated, all the files in `${CREW_PREFIX}/tmp/crew/` stay there throughout the installation. This approach saves space, and is more on flow with what `crew` (the Ruby script) does.

JL2210:
Any other suggestions?

uberhacker:
This is a scary update and please don't take the microscopic criticism in the wrong way.  This is a critical change and I just want to make sure we get it right.  You have a lot of good improvements here.

install.sh
```diff
@@ -94,133 +103,128 @@ esac
...
     *)
-      echo "Verification failed, something may be wrong with the $1 download."
+      echo "Verification failed, something may be wrong with the download."
       exit 1;;
     esac
 }
 
 function extract_install () {
-    cd $CREW_BREW_DIR
+    cd "${CREW_BREW_DIR}/dest"
```
uberhacker: This could be: `cd "#{CREW_DEST_DIR}"`

JL2210: Good catch.

JL2210: I need to rebase against `master` real quick. Give me some time to re-clone my repo.

install.sh
uberhacker: We're making this more complicated than needed.  This is all you need:
```
function extract_install () {
    #start with a clean slate
    rm -rf "${CREW_DEST_DIR}"
    mkdir -p "${CREW_DEST_DIR}"
    cd "${CREW_DEST_DIR}"

    #extract and install
    tar xpf ../"${2}"
    tar cpf - ./*/* | (cd /; tar xp --keep-directory-symlink -f -)
    mv ./dlist "${CREW_CONFIG_PATH}/meta/${1}.directorylist"
    mv ./filelist "${CREW_CONFIG_PATH}/meta/${1}.filelist"
}
```

install.sh
```diff
@@ -94,133 +103,128 @@ esac
...
 
 # install a base set of essential packages
 yes | crew install buildessential less most
 
+echo
+echo "If you have installed Chromebrew in a directory other than '/usr/local',"
+echo "please run these commands:"
+echo "echo 'export CREW_PREFIX=${CREW_PREFIX}' >> ~/.bashrc && . ~/.bashrc"
+echo "echo 'export PATH=\"\${CREW_PREFIX}/bin:\${CREW_PREFIX}/sbin:\${PATH}\"' >> ~/.bashrc && . ~/.bashrc"
+echo "echo 'export LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}' >> ~/.bashrc && . ~/.bashrc"
```
uberhacker: How about this?
```
if "${CREW_PREFIX}" != "/usr/local" then
  echo "echo 'export CREW_PREFIX=${CREW_PREFIX}' >> ~/.bashrc"
  echo "echo 'export PATH=\"\${CREW_PREFIX}/bin:\${CREW_PREFIX}/sbin:\${PATH}\"' >> ~/.bashrc"
  echo "echo 'export LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}' >> ~/.bashrc"
  source ~/.bashrc
fi
```
No need to print unless they have changed `${CREW_PREFIX}` and no need to source every line until the end.

JL2210: I don't think you want to `source ~/.bashrc` directly from the script :). I'll push this soon.

uberhacker: Yes, edited.

PR #3089, opened 2019-01-16
Allow for custom #{CREW_PREFIX}

JL2210:
Allow for custom `#{CREW_PREFIX}` via. environment variable
Depends on #3088

Bump minor version, new features added
Add `HOME` variable for future PR (Will push when this is merged)

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,11 +1,16 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.0.6'
+CREW_VERSION = '1.1.0'
 
 ARCH = `uname -m`.strip
 ARCH_LIB = if ARCH == 'x86_64' then 'lib64' else 'lib' end
 
-CREW_PREFIX = '/usr/local'
+if ENV['CREW_PREFIX'].to_s == ''
+  CREW_PREFIX = '/usr/local'
+else
+  CREW_PREFIX = ENV['CREW_PREFIX']
+  @pkg.build_from_source = true
+end
 CREW_LIB_PREFIX = CREW_PREFIX + '/' + ARCH_LIB
 
 CREW_LIB_PATH = CREW_PREFIX + '/lib/crew/'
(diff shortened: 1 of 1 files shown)


lib/const.rb
```diff
@@ -1,11 +1,16 @@
...
 ARCH = `uname -m`.strip
 ARCH_LIB = if ARCH == 'x86_64' then 'lib64' else 'lib' end
 
-CREW_PREFIX = '/usr/local'
+if ENV['CREW_PREFIX'].to_s == ''
+  CREW_PREFIX = '/usr/local'
+else
+  CREW_PREFIX = ENV['CREW_PREFIX']
+  @pkg.build_from_source = true
+end
```
uberhacker: Probably a better way:
```
CREW_PREFIX = '/usr/local'
if ENV['CREW_PREFIX'].to_s != '' then
  @pkg.build_from_source = true
end
```

uberhacker: Actually, this might not work since you cannot reassign constants in ruby.  Nevermind.

lib/const.rb
```diff
@@ -1,11 +1,16 @@
...
 
 ARCH = `uname -m`.strip
 ARCH_LIB = if ARCH == 'x86_64' then 'lib64' else 'lib' end
 
-CREW_PREFIX = '/usr/local'
+if ENV['CREW_PREFIX'].to_s == ''
+  CREW_PREFIX = '/usr/local'
+else
+  CREW_PREFIX = ENV['CREW_PREFIX']
+  @pkg.build_from_source = true
```
uberhacker: This is painful in so many ways. :dizzy_face:

lib/const.rb
```diff
@@ -15,6 +20,7 @@
 CREW_DEST_PREFIX = CREW_DEST_DIR + CREW_PREFIX
 CREW_DEST_LIB_PREFIX = CREW_DEST_DIR + CREW_LIB_PREFIX
 CREW_DEST_HOME = CREW_DEST_DIR + ENV['HOME']
+HOME = ENV['HOME']
```
uberhacker: Good idea.

lib/const.rb
JL2210: How so?

uberhacker: Uh, you are requiring the user to compile EVERYTHING from source?

JL2210: Well, yeah. You can't just install software compiled for `/usr/local` in `/some/random/directory`.

PR #3093, opened 2019-01-18
Update HOME and CREW_DEST_HOME variables

JL2210:
Open to all suggestions!

Installs files in `#{CREW_PREFIX}/#{HOME}` if installing to a custom prefix.


uberhacker:
This is a good idea if you want your entire config portable such as a usb stick or external drive.

PR #3094, opened 2019-01-18
Add pre-built binaries for cairo

uberhacker:
Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/cairo.rb
+++ b/packages/cairo.rb
@@ -8,14 +8,22 @@ class Cairo < Package
   source_sha256 '5e7b29b3f113ef870d1e3ecf8adf21f923396401604bda16d44be45e66052331'
 
   binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/cairo-1.16.0-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/cairo-1.16.0-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


JL2210:
Good job. Make sure to explicitly enable those dependencies, though.

uberhacker:
Explicitly enable what?  All the dependencies are already inherited.

JL2210:
Something like this:
```
           '--enable-lzo',
           '--enable-rsvg',
```

uberhacker:
I believe those are already inheritedly enabled by default.

JL2210:
Let me check that...

JL2210:
Just enable them. They may not be enabled by default in the future, and it does no harm.
I don't think LZO is needed:
```
chronos@localhost ~/Downloads/tarballs/cairo-1.16.0 $ ./configure --help | grep -i lzo
chronos@localhost ~/Downloads/tarballs/cairo-1.16.0 $ 
```

uberhacker:
Check your build log with and without lzo and librsvg dependencies.

packages/cairo.rb
```diff
@@ -8,14 +8,22 @@ class Cairo < Package
   source_sha256 '5e7b29b3f113ef870d1e3ecf8adf21f923396401604bda16d44be45e66052331'
 
   binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/cairo-1.16.0-chromeos-armv7l.tar.xz',
```
JL2210: This is good.

PR #3099, opened 2019-01-20
Update lzip package

JL2210:


packages/lzip.rb
```diff
@@ -2,27 +2,19 @@
...
   binary_sha256 ({
-    aarch64: '7cbe7a366aed4da37ffe6fe1cbae057917879559f794803d613d0421f71a92e8',
-     armv7l: '7cbe7a366aed4da37ffe6fe1cbae057917879559f794803d613d0421f71a92e8',
-       i686: 'e64bad3920d72f7e5e5c952c5ee3870f40448c0d833377631dd67e54a3de2c37',
-     x86_64: 'dbbc38602aaeb84e11d971aeb35d64b49fc30d44354b902b767e0529a3ce7d04',
   })
 
   def self.build
-    # default of -O2 causes a run-time segmentation fault on armv7l
-    system "./configure", "CXXFLAGS=-Wall -W -O1"
```
uberhacker: How can you so confidently remove this without testing on armv7l first?

JL2210: Can you test? If it works, I'd imagine it would have to do with one of these:
```
        * lzip_index.cc: Detect some kinds of corrupt trailers.
        * main.cc (main): Check return value of close( infd ).
```
Anyway, it needs binaries (it's a core package).

JL2210:
How does it work?

JL2210:
Make sure to rebase.

PR #3103, opened 2019-01-21
Add xorg_xmessage and xorg_util_macros

superloach:
Adds the xmessage tool, as well as the Xorg util macros it needs.

PR #3104, opened 2019-01-21
Xmessage

superloach:
Fixes #3103

### DIFF
--- a/packages/xmessage.rb
+++ b/packages/xmessage.rb
@@ -2,7 +2,7 @@ require 'package'
 
 class Xmessage < Package
   description 'Xmessage displays a message or query in a window.'
-  homepage 'https://www.x.org/archive/X11R6.9.0/doc/html/xmessage.1.html'
+  homepage 'https://www.x.org/'
   version '1.0.5'
   source_url 'https://www.x.org/releases/individual/app/xmessage-1.0.5.tar.bz2'
   source_sha256 '373dfb81e7a6f06d3d22485a12fcde6e255d58c6dee1bbaeb00c7d0caa9b2029'


packages/xmessage.rb
```diff
@@ -10,7 +10,6 @@ class Xmessage < Package
   depends_on 'xorg_lib'
 
   def self.build
-    system "bash --norc"
```
JL2210: Why?

JL2210: Not a problem with the package, but just curious.

superloach: I used that line for debugging, don't worry about it.

JL2210: OK. I find that I do that sometimes: I download source tarballs when I'm offline and modify the packages to use them, and then switch it back before I push. It's very useful for packages with large sources.

packages/xmessage.rb
```diff
@@ -0,0 +1,22 @@
+require 'package'
+
+class Xmessage < Package
+  description 'Xmessage displays a message or query in a window.'
+  homepage 'https://www.x.org/archive/X11R6.9.0/doc/html/xmessage.1.html'
```
uberhacker: Better homepage? https://sourceforge.net/projects/xmessage/

JL2210: That's old. Not from `x.org`.

JL2210: Look at this: "Last Update: 2013-04-09"

uberhacker:
Welcome back @superloach!

packages/xmessage.rb
JL2210: Also: `A library for binary message parsing and assembling, through message definitions in xml. Especially useful in protocols such as SMPP, in which messages have a header and a body. Also provides functions to transform among diffrent protocols and messages`
Not the same package.

uberhacker: Nevermind.  I think that is another project.

uberhacker: How about https://xorg.freedesktop.org/wiki/?  I want to get away from using man pages for homepages.

JL2210: Nobody ever looks at a homepage, except maybe for doing updates. `https://xorg.freedesktop.org/wiki/` has nothing on `xmessage`.

> I want to get away from using man pages for homepages.

Why? They explain how to use the package.

JL2210: This at least has something explaining it: https://www.x.org/wiki/ModuleDescriptions/#legacyx11platform

JL2210: Sorry for intruding on this, @superloach. What's your opinion?

superloach: I think what we currently have is the best option, because it's a direct link to the latest version of this program.
The Xorg wiki is too general, and no longer has this package. The legacy program list isn't very specific, and doesn't show usage. Similarly, the Xorg file list isn't useful to users trying to learn more about this program.

uberhacker: @superloach: Thank you for the feedback.  The problem I'm trying to solve for the homepage is the canonical location where this package can be found regardless of the version.  When I look for package sources, I need the most up to date location that has the most activity for the package, otherwise, future versions may be difficult to locate.  That said, there is no guarantee this version of the man page will not change in the future and we need a way to find information about the package sources other than a list with a specific version of xorg.

JL2210: There will most certainly not be any new updates to this package. This page probably has as much activity as any other page would have, considering the circumstances. Just look at `source_url` when looking for sources. That's what I do.
There is no unversioned form of this manpage. I scoured `https://x.org/current/` and couldn't find one.

uberhacker: The problem with looking for the source is you don't know where to look if the homepage is versioned.  I don't think you understand the significance of what I'm trying to communicate here.  If you read the wiki page (which I suggest as the homepage), you would learn:
```
The last full release of the entire X.Org stack was X11R7.7 - since then individual X.Org modules have been released independently as needed - see the xorg-announce archives for details of those releases, and https://www.x.org/releases/individual/ for downloads. Information about all releases is available. (Important: If you have an older release, please see the Security page for information on security updates.)
```
So your made up story about this package not having any future updates is bogus.

JL2210: Tell you what: If and when this package has any new updates, you let me know. Otherwise, this is fine.
In this case, and in the case of many other X packages (e.g. `xhost`), I really doubt that there's any "canonical URL" for you.

JL2210: Also, I suggested https://www.x.org/releases/individual/. You rejected it.

uberhacker: Right, because it's not a homepage.  A homepage exists for almost every project including this project.  You could make an argument that https://www.x.org is the homepage because this package is really just a group of packages that make up the whole of X.  In fact, several packages here already have that as a homepage which is technically correct.

superloach: So we should use https://www.x.org?

uberhacker: It redirects to the wiki page, but yeah, that would be better.  Hopefully, you understand what I'm trying to communicate here.

PR #3105, opened 2019-01-21
Prevent repeat dependency checking

JL2210:
Thanks, @dpgettings, for writing this up.

Keep track of the names of dependencies that have already been checked
in the `expand_dependencies`/`push_dependencies` loop. A quick and dirty way
to break out of dependency cycles and speed up the dependency resolution
step in general.


uberhacker:
Thanks for the PR @dpgettings!  I noticed this has changes to `const.rb` and `cairo.rb`.  Please merge the latest master into your working branch to remove these extra changes.  You can modify `const.rb` but it will be a newer version.  There is also another PR #3096 that bumps the crew version so please increment again to the correct version.

JL2210:
I think I might have used the wrong fork.

dpgettings:
I just pushed https://github.com/dpgettings/chromebrew/commit/cb36d20a8125e5ab5009694bd936de068f64668b and https://github.com/dpgettings/chromebrew/commit/842a66d77c6361991cbcb5c6c676d2ed23d3bb06 which should bring both branches even with the latest master.

The two branches are basically equivalent in terms of function. In the `prevent-repeat-dep-checks` branch I used `Set` because I thought the hashing would help with speed. A couple days later when I compared against the same approach using just arrays I realized it was pretty much a wash. So for the sake of simplicity I created the `faster-dep-checks` branch which drops the `Set` require and uses plain arrays for the lists in `expand_dependencies`. 

The only functional difference I've noticed is that the final list of dependencies comes out in a different order between the two branches. The content of the lists is the same, as far as I've checked. It seems like the order could affect the installation of packages that have `:build` tags somewhere along the chain, but I'm not sure.

Feel free to PR whichever branch—it's pretty much just a matter of taste.

uberhacker:
Closing in favor of #3106.

PR #3106, opened 2019-01-22
Faster dep checks

dpgettings:
## Description
Keep track of the names of dependencies that have already been checked
in the `expand_dependencies`/`push_dependencies loop`. A quick and dirty way
to break out of dependency cycles and speed up the dependency resolution
step in general.

## Addtional information
This PR would merge the branch `dpgettings:faster-dep-checks` which uses arrays to store the names of dependencies that have already been checked. This is an alternative to #3105 which would merge `dpgettings:prevent-repeat-dep-checks`, a different branch that accomplishes the same thing as `dpgettings:faster-dep-checks` but which uses the `Set` class to store dependency names. 

The only noticeable functional difference between the two versions is the order of a package's final dependency list, which is often different; the deps named on the list are exactly the same for every current package. In the end, the performance gain due to the use of `Set` vs arrays is negligible.

### DIFF
--- a/crew
+++ b/crew
@@ -627,41 +627,28 @@ end
 
 def expand_dependencies
   @dependencies = []
-  @dontRecheck = []
-
   def push_dependencies
     if @pkg.is_binary?(@device[:architecture]) ||
        (!@pkg.in_upgrade && !@pkg.build_from_source && @device[:installed_packages].any? { |pkg| pkg[:name] == @pkg.name })
       # retrieve name of dependencies that doesn't contain :build tag
       check_deps = @pkg.dependencies.select {|k, v| !v.include?(:build)}.map {|k, v| k}
-    elsif @pkg.is_fake?
-      # retrieve name of all dependencies
-      check_deps = @pkg.dependencies.map {|k, v| k}
     else
       # retrieve name of all dependencies
       check_deps = @pkg.dependencies.map {|k, v| k}
     end
-
-    # remove a dependent package which is equal to the target
-    check_deps.select! {|name| @pkgName != name}
-
-    # add new dependencies at the beginning of array
-    @dependencies = check_deps.clone.concat(@dependencies)
-
     # check all dependencies recursively
     check_deps.each do |dep|
-      # skip a dependency if it's been checked already
-      next if @dontRecheck.include?(dep)
-      search dep, true
-      push_dependencies
-      # Remember dep so we can skip it from now on
-      @dontRecheck << dep
+      # build unique dependencies list
+      unless @dependencies.include?(dep) || dep == @pkgName
+        @dependencies << dep
+        search dep, true
+        push_dependencies
+      end
     end
   end
-
   push_dependencies
-
-  @dependencies.uniq
+  # resolve_dependencies needs expand_dependencies to return @dependencies
+  return @dependencies
 end
 
 def resolve_dependencies
--- a/crew
+++ b/crew
@@ -627,6 +627,7 @@ end
 
 def expand_dependencies
   @dependencies = []
+  @dontRecheck = []
 
   def push_dependencies
     if @pkg.is_binary?(@device[:architecture]) ||
@@ -649,8 +650,12 @@ def expand_dependencies
 
     # check all dependencies recursively
     check_deps.each do |dep|
+      # skip a dependency if it's been checked already
+      next if @dontRecheck.include?(dep)
       search dep, true
       push_dependencies
+      # Remember dep so we can skip it from now on
+      @dontRecheck << dep
     end
   end
 


dpgettings:
Created this additional PR for convenience/disambiguation. Feel free to close if you decide to merge the other branch in #3105.

dpgettings:
@uberhacker: Actually this won't work because it will never fall through to `search dep`. By merging `check_deps` into `@dependencies` before looping over `check_deps`, the condition `@dependencies.include?(dep)` inside the loop is always going to be true. What you're effectively doing is:
```ruby
# Without doing this @dependencies = (check_deps + @dependencies).uniq
check_deps.each do |dep|
  (search dep, etc.) unless @dependencies.include?(dep) || check_deps.include?(dep)
end
```
The search will only go down one level and the final dependency list will always be equal to (or a subset of) `@pkgName`'s direct dependencies.

There is a way to do it with just one list though, and it can simplify the logic even more:
```ruby
def expand_dependencies
  @dependencies = []
  def push_dependencies
    if @pkg.is_binary?(@device[:architecture]) ||
       (!@pkg.in_upgrade && !@pkg.build_from_source && @device[:installed_packages].any? { |pkg| pkg[:name] == @pkg.name })
      # retrieve name of dependencies that doesn't contain :build tag                                                                                                                      
      check_deps = @pkg.dependencies.select {|k, v| !v.include?(:build)}.map {|k, v| k}
    else
      # retrieve name of all dependencies                                                                                                                                                  
      check_deps = @pkg.dependencies.map {|k, v| k}
    end
    # check all dependencies recursively                                                                                                                                                   
    check_deps.each do |dep|
      # build unique dependencies list                                                                                                                                                     
      unless @dependencies.include?(dep) || dep == @pkgName
        @dependencies << dep
        search dep, true
        push_dependencies
      end
    end
  end
  push_dependencies
  # resolve_dependencies needs expand_dependencies to return @dependencies
  return @dependencies
end
```
`check_deps` is still being merged into `@dependencies` but it's happening one at a time and after `@dependencies` gets checked for the preexistence of `dep`. It should be impossible to add a `dep` that's already in `@dependencies` so the final `@dependencies` ought to be unique already. And as long as we're checking names for permission to go to `search dep`, we might as well put the `dep != @pkgName` constraint there. Also, the final `@dependencies` in `expand_dependencies` is a return value. This caught me too at first so it might be worth making the return explicit?

I tested this for a few packages and can verify that it seems to be coming up with the same final package list as the version in https://github.com/dpgettings/chromebrew/commit/8a7d7260b786e6c763badfd139c209fbd0488bc6

superloach:
This looks really useful! Hopefully it'll give the speed boost needed to make things like GIMP or FreeRDP reasonable to install... though the second one has issues of its own 👀

dpgettings:
@superloach: On my machine this modified version of `crew` completes dependency resolution for GIMP in about 3 seconds. Not bad!

superloach:
That's phenomenal! 😍
Is there any way I could help with this?

crew
```diff
@@ -627,36 +627,28 @@ end
...
+        push_dependencies
+      end
     end
   end
-
   push_dependencies
-
-  @dependencies.uniq
+  # resolve_dependencies needs expand_dependencies to return @dependencies
+  return @dependencies
```
uberhacker: Actually, you don't need to return the value here since @dependencies is an instance variable and is available globally.  This looks really good @dpgettings.  Great work!  The only thing this needs now is to bump the `CREW_VERSION` in `lib/const.rb` since this changes the crew command.

dpgettings: I guess I'm confused about the return thing. Looking down at the next method, `resolve_dependencies`, it looks like `dependencies` here is actually a local variable being set by a return value from `expand_dependencies`. It seems like getting rid of the return in `expand_dependencies` would necessitate changing every occurrence of `dependencies` in the `resolve_dependencies` method into a reference to the instance variable `@dependencies`, i.e.: 
```ruby
def resolve_dependencies
  #dependencies = expand_dependencies
  expand_dependencies  # calling without an assignment
  # leave only not installed packages in dependencies
  @dependencies.select! {|name| @device[:installed_packages].none? {|pkg| pkg[:name] == name}}
  puts "The following packages also need to be installed: "
  @dependencies.each do |dep|
    print dep + " "
  end
  puts ""
  print "Do you agree? [Y/n] "
  response = STDIN.getc
  case response
  when "n"
    abort "No changes made."
  when "\n", "y", "Y"
    puts "Proceeding..."
    proceed = true
  else
    puts "I don't understand `#{response}`. :(".lightred
    abort "No changes made."
  end
  if proceed
    @dependencies.each do |dep|
      search dep
      install
    end
  end
end
def install
  if !@pkg.in_upgrade && @device[:installed_packages].any? { |pkg| pkg[:name] == @pkg.name }
    return
  end
```
... Otherwise, `resolve_dependencies`'s local variable `dependencies` would be set to whatever the `expand_dependencies` method happened to implicitly return (if anything), such as the return value of the last call to `push_dependencies`.

Am I totally off base here? I'm definitely not experienced enough in Ruby to be totally sure I know what I'm talking about, haha.

JL2210:
This is like Chromebrew on steroids! Dependency checks that used to take 5 minutes take 3 seconds now!
We should probably fix this without a workaround, but for now this is awesome!

crew
JL2210: @uberhacker If you want to type something without accidentally pinging a user, surround it in backticks, e.g. `@dependencies`.

uberhacker: @dpgettings: Does it work without setting the local variable?

uberhacker: @dpgettings: I tested the code you have there and it works great.  I can merge this if you push those changes along with bumping `CREW_VERSION` in `lib/const.rb`.

skiman6010:
I seriously dig this code! Thanks for fulfilling a need I didn't even know could be filled! @dpgettings

crew
JL2210: @dpgettings?

JL2210: @uberhacker Can we just merge this? I don't think it would require a version bump, as it doesn't remove any dependencies.

uberhacker: It changes the crew command which is what drives the need to increment the version.

JL2210: All right, at least the patch version. Could you do it? It doesn't seem like he's responding.

JL2210: @skycocker

PR #3108, opened 2019-01-22
Add postinstall command to crew

uberhacker:
Should help in those instances when the messages scroll past the viewable screen.

### DIFF
--- a/crew
+++ b/crew
@@ -27,6 +27,7 @@ Usage:
   crew help [<command>]
   crew install [-k|--keep] [-s|--build-from-source] [-S|--recursive-build] [-v|--verbose] <name> ...
   crew list (available|installed)
+  crew postinstall <name> ...
   crew remove [-v|--verbose] <name> ...
   crew search [-v|--verbose] [<name> ...]
   crew update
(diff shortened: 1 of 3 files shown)


JL2210:
Yay!

crew
```diff
@@ -927,6 +932,17 @@ def list_command (args)
   end
 end
 
+def postinstall_command (args)
+  args["<name>"].each do |name|
+    set_package name, true
+    if @device[:installed_packages].any? do |elem| elem[:name] == name end
+      @pkg.postinstall
+    else
+      puts "Package #{name} is not installed. :(".lightred
```
JL2210: I don't think it should require the package to be installed.

uberhacker: Actually, it does in several instances where files are expected to exist.  When you think about it, it does make sense.

JL2210: Yes, it could also overwrite user configuration. Just output a warning if the package isn't installed.

uberhacker: That's what I'm doing.

JL2210: Yes, but you should still go through with the command. You know, there's a difference between a warning and an error.

uberhacker: No, I shouldn't.  See above.

JL2210: All right. Other than that, this is fine.

JL2210: Is there any way you could add an option to force it? I've been looking forward to this for a long time.

uberhacker: Why would you need to run postinstall on an uninstalled package?

JL2210: Testing a new package and making sure you haven't forgotten `.lightblue` without installing it? Making sure there aren't any typos?

uberhacker: You wouldn't see that at the end anyway?

JL2210: I don't want to go through the whole build process again just to see the postinstall when I've changed it.

uberhacker: Wait, testing a package includes installing it which would also output the text at the end.  You really should be testing it with the install anyway.

JL2210: I don't want to build a package again when all I've changed is the postinstall.

PR #3113, opened 2019-01-23
Add RSU-Client package

superloach:
Adds a package for the RuneScape Unix/Linux Client.
RuneScape login in a Sommelier window

### DIFF
--- a/packages/rsu_client.rb
+++ b/packages/rsu_client.rb
@@ -16,7 +16,7 @@ class Rsu_client < Package
   def self.patch
     Dir.chdir 'runescape' do
       system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
-                sed -i 's,#{CREW_PREFIX},/usr,g' \${f}
+                sed -i 's,/usr/local,/usr,g' \${f}
               done"
       system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
                 sed -i 's,/usr,#{CREW_PREFIX},g' \${f}
--- a/packages/rsu_client.rb
+++ b/packages/rsu_client.rb
@@ -16,10 +16,10 @@ class Rsu_client < Package
   def self.patch
     Dir.chdir 'runescape' do
       system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
-                sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' $f
+                sed -i 's,#{CREW_PREFIX},/usr,g' \${f}
               done"
       system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
-                sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' $f
+                sed -i 's,/usr,#{CREW_PREFIX},g' \${f}
               done"
     end
   end
@@ -29,9 +29,9 @@ class Rsu_client < Package
   end
 
   def self.install
-    system 'mkdir', '-pv', "#{CREW_DEST_PREFIX}/share"
-    system 'cp', '-rv', 'runescape', "#{CREW_DEST_PREFIX}/share/"
-    system 'mkdir', '-pv', "#{CREW_DEST_PREFIX}/bin"
-    system 'ln', '-sv', "#{CREW_PREFIX}/share/runescape/runescape", "#{CREW_DEST_PREFIX}/bin/runescape"
+    system 'mkdir', '-p', "#{CREW_DEST_PREFIX}/share"
+    system 'cp', '-a', 'runescape', "#{CREW_DEST_PREFIX}/share/"
+    system 'mkdir', '-p', "#{CREW_DEST_PREFIX}/bin"
+    system 'ln', '-s', "#{CREW_PREFIX}/share/runescape/runescape", "#{CREW_DEST_PREFIX}/bin/runescape"
   end
 end
--- a/packages/rsu_client.rb
+++ b/packages/rsu_client.rb
@@ -9,54 +9,18 @@ class Rsu_client < Package
 
   depends_on 'jdk8'
   depends_on 'p7zip'
-  depends_on 'perl'
   depends_on 'wxwidgets'
   depends_on 'xdg_utils'
   depends_on 'sommelier'
 
   def self.patch
     Dir.chdir 'runescape' do
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./changelog.txt"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./install-desktop-icons"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./rsu-settings"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./rsu/bin/rsu-query-unix"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./rsu/framework/API/client/launch/launcher.pm"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./rsu/framework/Perl5lib/XML/RSSLite.pm"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./rsu/framework/modules/rsu/java/jre.pm"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./rsu/framework/modules/rsu/mains.pm"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./rsu/framework/modules/rsu/nvidia/optimus.pm"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./rsu/framework/modules/updater/download/sysdload.pm"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./rsu/framework/modules/updater/download/wxsysdload.pm"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./rsu/framework/modules/updater/gui/zenity.pm"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./rsu/rsu-query"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./runescape"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./templates/packaging/usr/games/rsu-settings"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./templates/packaging/usr/games/runescape"
-      system "sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' ./templates/packaging/usr/games/update-runescape-client"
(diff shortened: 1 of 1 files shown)


packages/rsu_client.rb
```diff
@@ -0,0 +1,73 @@
...
+      system "sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' ./rsu/framework/modules/updater/download/wxsysdload.pm"
+      system "sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' ./rsu/framework/modules/updater/gui/zenity.pm"
+      system "sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' ./rsu/rsu-query"
+      system "sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' ./runescape"
+      system "sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' ./templates/packaging/usr/games/rsu-settings"
+      system "sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' ./templates/packaging/usr/games/runescape"
+      system "sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' ./templates/packaging/usr/games/update-runescape-client"
+      system "sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' ./templates/packaging/usr/local/bin/xdg-open"
+      system "sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' ./templates/protocol/install-jagex-protocol-linux"
+      system "sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' ./updater"
```
JL2210: Better solution?
```
      system "for file in \$(find . -type f); do
                sed -i 's,/usr,#{CREW_PREFIX},g' \${file}
              done"
```

packages/rsu_client.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def self.build
+    system 'cpan install Archive::Extract exit'
+  end
+
+  def self.install
+    system 'mkdir', '-pv', "#{CREW_DEST_PREFIX}/share"
+    system 'cp', '-rv', 'runescape', "#{CREW_DEST_PREFIX}/share/"
+    system 'mkdir', '-pv', "#{CREW_DEST_PREFIX}/bin"
+    system 'ln', '-sv', "#{CREW_PREFIX}/share/runescape/runescape", "#{CREW_DEST_PREFIX}/bin/runescape"
```
JL2210: Is this open source or otherwise?

uberhacker:
Another time waster! :stuck_out_tongue_closed_eyes:

packages/rsu_client.rb
JL2210: This looks proprietary. If so, please close. Proprietary applications are incompatible with Chromebrew (incompatible licenses).

packages/rsu_client.rb
```diff
@@ -0,0 +1,73 @@
...
+class Rsu_client < Package
+  description 'RSU-Client is a RuneScape Client Launcher written for the now Legacy client and now OldSchool.'
+  homepage 'https://github.com/rsu-client/rsu-client'
+  version '4.3.7'
+  source_url 'https://github.com/rsu-client/rsu-client/archive/v4.3.7.tar.gz'
+  source_sha256 '7cc589de111baad956c0c41a4d7f1b45168886bf623dd3886d07fa3d1dfd604c'
+
+  depends_on 'jdk8'
+  depends_on 'p7zip'
+  depends_on 'perl'
```
uberhacker: No need since perl is already a core package.

packages/rsu_client.rb
uberhacker: Actually, the way she has it is preferable.  I would rather see a full path than a relative path when I examine a symlink.

packages/rsu_client.rb
JL2210: You know, on an unrelated topic, maybe we should stop doing this. I doubt the average user uses the toolchain on a daily basis, and you can't even install a simple text editor without having missing library errors if you don't have `buildessential` installed. Plus, with #3106, dependency loops and speed shouldn't be an issue.

packages/rsu_client.rb
superloach: That would include binary files. The list I got came from searching for `/usr/bin`, for some reason searching for `/usr` causes issues. I'll try to come up with proper patches in the future, but for now it works.

packages/rsu_client.rb
JL2210: Not necessarily. This way, the package will be movable.

superloach: And using the `CREW_*` variables isn't? Not sure what your point is here.

packages/rsu_client.rb
superloach: Alright, I will.
Here's a translation of my search script, if needed.
```ruby
system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
          sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' $f
        done"
```

uberhacker: Nice suggestion.

packages/rsu_client.rb
JL2210: Along with that, who the heck is going to view a symlink to a game engine? :P 
I know relative paths in symlinks are preferred in some cases elsewhere (package managers, for one).

packages/rsu_client.rb
superloach: @JL2210 your version gives these grep warnings:
```
grep: ./share/img/Custom: No such file or directory
grep: Icons.txt: No such file or directory
```
Looks like a quoting issue, but I couldn't figure out how to fix it.

packages/rsu_client.rb
uberhacker: Stop doing what?  Let's not get sloppy by being lazy.

packages/rsu_client.rb
superloach: My solution, on the other hand, surprisingly works!

packages/rsu_client.rb
superloach: Yeah, even if we have speed improvements coming up doesn't mean we shouldn't do our best to remove redundancies.
Though I do understand the annoyance of getting a dozen comments on a file just for redundant dependencies 🤣
Perhaps we could avoid immediately pointing them out, and instead come up with a tool to automatically remove them and run it somewhat regularly?

packages/rsu_client.rb
```diff
@@ -0,0 +1,37 @@
...
+  depends_on 'sommelier'
+
+  def self.patch
+    Dir.chdir 'runescape' do
+      system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
+                sed -i 's,#{CREW_PREFIX},/usr,gw /dev/stdout' $f
+              done"
+      system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
+                sed -i 's,/usr,#{CREW_PREFIX},gw /dev/stdout' $f
+              done"
```
uberhacker: Am I seeing double here? LoL

packages/rsu_client.rb
uberhacker: Who the heck is going to move things around without reinstalling?  Only you, apparently.

packages/rsu_client.rb
JL2210: Yes, but I'd like to have a working Chromebrew installation without installing `buildessential`. Currently, if you omit installing `buildessential` during the initial install phase, you have to hand-pick out the dependencies for some packages.
Example: If you run a fresh Chromebrew installation and remove any instance of `buildessential` from `install.sh`, and then you run `crew install nano && nano`, you get an error about a missing `libmagic.so` (from `filecmd`).

packages/rsu_client.rb
JL2210: Come on. The symlink will point to the same file. This way reduces the package file size and the symlink size.

packages/rsu_client.rb
JL2210: Not quite. First, she replaces `#{CREW_PREFIX}` with `/usr`, and then she replaces `/usr` with `#{CREW_PREFIX}`. No idea why, though.

JL2210: Make sure to backslash and bracket `$f`: `\${f}`

packages/rsu_client.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  def self.build
+    system 'cpan install Archive::Extract exit'
+  end
+
+  def self.install
+    system 'mkdir', '-pv', "#{CREW_DEST_PREFIX}/share"
```
JL2210: Could you please remove the verbosity in this package?
`ln -sv` > `ln -s`, `w /dev/stdout` > ``, `cp -rv` > `cp -a`, and `mkdir -pv` > `mkdir -p`
`cp -a` preserves permissions.

packages/rsu_client.rb
JL2210: Along with that, could you please tell me why you prefer to see a full path? And why the way you like it is preferable?

uberhacker: It clearly indicates the full path to the target.  Your way causes a dangling symlink if the files aren't moved relative to each other and then there would be no way to determine where the original target is located.

JL2210: All right.

packages/rsu_client.rb
JL2210: All right. You don't need to brag. ;P

packages/rsu_client.rb
superloach: If I just replace `/usr` with `/usr/local`, you end up with things like `/usr/local/local`, which I'm fairly sure shouldn't happen.

packages/rsu_client.rb
superloach: Yes, I was going to address that.

skiman6010:
👀  Hey look I can stop using the Android app and use this instead. Neat!

packages/rsu_client.rb
uberhacker: What if you did just this?
```
system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
                sed -i 's,/usr/local,#{CREW_PREFIX},g' \${f}
              done"
```

superloach: That wouldn't turn `/usr/bin` into `/usr/local/bin`.

uberhacker: The danger of having these two search and replaces here is the assumption that `#{CREW_PREFIX}` is `/usr/local` and that isn't always the case.  Would this work?
```
system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
                sed -i 's,/usr/bin,#{CREW_PREFIX}/bin,g' \${f}
              done"
system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
                sed -i 's,/usr/local,#{CREW_PREFIX},g' \${f}
              done"
```

superloach: Hm... I see what you mean about `/usr/local`.
How about this?
```
system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
          sed -i 's,/usr/local,/usr,g' $f
        done"
system "for f in \$(grep -r /usr/bin .|egrep -v '^Binary'|egrep -o '^[^:]+');do
          sed -i 's,/usr,#{CREW_PREFIX},g' $f
        done"
```
This moves everything from `/usr/local` into `/usr`, then from `/usr` to `#{CREW_PREFIX}`.

uberhacker: That could work also.

superloach: It's basically what we already have, just without assuming the file starts out with `#{CREW_PREFIX}` in it rather than `/usr/local`.

PR #3127, opened 2019-01-29
Add gcc8 package

uberhacker:
Tested on:
- [x] aarch64
- [x] armv7l
- [x] i686
- [x] x86_64


packages/gcc8.rb
```diff
@@ -0,0 +1,160 @@
...
+  depends_on 'mpfr'
+  depends_on 'mpc'
+  depends_on 'isl'
+  depends_on 'cloog'
+  depends_on 'glibc'
+
+  def self.patch
+    # See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86724#c4.
+    system "sed -i '40i#include <isl/id.h>' gcc/graphite.h"
+    system "sed -i '41i#include <isl/space.h>' gcc/graphite.h"
```
JL2210: You can always use my patch... Also, very bad idea to hardcode the line names.

packages/gcc8.rb
```diff
@@ -0,0 +1,160 @@
...
+          system "../configure",
+                 "--prefix=#{CREW_PREFIX}",
+                 "--libdir=#{CREW_LIB_PREFIX}",
+                 "--build=armv7l-cros-linux-gnueabihf",
+                 "--host=armv7l-cros-linux-gnueabihf",
+                 "--target=armv7l-cros-linux-gnueabihf",
+                 "--enable-checking=release",
+                 "--disable-multilib",
+                 "--enable-threads=posix",
+                 "--disable-bootstrap",
```
JL2210: Why?

packages/gcc8.rb
```diff
@@ -0,0 +1,160 @@
...
+      system "ln -sv #{CREW_PREFIX}/bin/gcov-tool-8.2 #{CREW_DEST_PREFIX}/bin/gcov-tool"
+      system "ln -sv #{CREW_PREFIX}/bin/gfortran-8.2 #{CREW_DEST_PREFIX}/bin/gfortran"
+
+      system "ln -sv #{CREW_PREFIX}/bin/$(gcc -dumpmachine)-c++-8.2 #{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-c++"
+      system "ln -sv #{CREW_PREFIX}/bin/$(gcc -dumpmachine)-g++-8.2 #{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-g++"
+      system "ln -sv #{CREW_PREFIX}/bin/$(gcc -dumpmachine)-gcc-8.2 #{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc"
+      system "ln -sv #{CREW_PREFIX}/bin/$(gcc -dumpmachine)-gcc-ar-8.2 #{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc-ar"
+      system "ln -sv #{CREW_PREFIX}/bin/$(gcc -dumpmachine)-gcc-nm-8.2 #{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc-nm"
+      system "ln -sv #{CREW_PREFIX}/bin/$(gcc -dumpmachine)-gcc-ranlib-8.2 #{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc-ranlib"
+      system "ln -sv #{CREW_PREFIX}/bin/$(gcc -dumpmachine)-gfortran-8.2 #{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gfortran"
```
JL2210: You don't need half of these. Also, please don't use `#{CREW_PREFIX}/bin/` before any of these. It's unneeded.

install.sh
```diff
@@ -61,11 +61,11 @@ case "${ARCH}" in
   ;;
 "armv7l")
   if ! type "xz" > /dev/null; then
-    urls+=('https://github.com/snailium/chrome-cross/releases/download/v1.8.1/xz-5.2.3-chromeos-armv7l.tar.gz')
-    sha256s+=('4dc9f086ee7613ab0145ec0ed5ac804c80c620c92f515cb62bae8d3c508cbfe7')
+    urls+=('https://dl.bintray.com/chromebrew/chromebrew/xzutils-5.2.3-2-chromeos-armv7l.tar.xz')
```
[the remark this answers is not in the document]
uberhacker: I was hoping nobody noticed. :blush: Anyway, I don't think this will happen on arm anymore with newer releases.

install.sh
```diff
@@ -61,11 +61,11 @@ case "${ARCH}" in
...
+    sha256s+=('839ead7097adee016986a9992d8cf8f4c91554ade2b6e111f7d0379c595be366')
   fi
-  urls+=('https://dl.bintray.com/chromebrew/chromebrew/gcc7-7.3.0-3-chromeos-armv7l.tar.xz')
-  sha256s+=('c9dd2a2b2d195f0f2021dc673d6317e46b5dfaef582f496b4cea1f3c5c2c17ad')
+  urls+=('https://dl.bintray.com/chromebrew/chromebrew/gcc8-8.2.0-chromeos-armv7l.tar.xz')
```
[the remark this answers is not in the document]
uberhacker: We need gcc8 as a "core" package anyway so no biggie.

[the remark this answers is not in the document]
uberhacker: Although, not relying on gcc allows you to uninstall gcc and still execute crew commands which is probably preferable.

packages/buildessential.rb
```diff
@@ -3,11 +3,11 @@
 class Buildessential < Package
   description 'A collection of tools essential to compile and build software.'
   homepage ''
-  version '1.3'
+  version '1.4'
 
   is_fake
 
-  depends_on 'gcc7'
+  depends_on 'gcc8'
```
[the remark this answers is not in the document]
uberhacker: It didn't when I tested.  I was able to compile fine.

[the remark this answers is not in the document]
uberhacker: Upgrading doesn't remove gcc7, however.  Might run into a problem there.

packages/gcc7.rb
```diff
@@ -3,27 +3,27 @@
...
+     armv7l: '07126320a193d947aeac51f1827ff954c322a8f43014f3cff146715b72f9d7f8',
+       i686: '72cd956d055ad418b052dab407ec3a2b16663b09c396073a23bd12828fcf03c3',
+     x86_64: '6cce57e98836ccc3c6773df1627bca2102b4f89c7abf64dc47f7b97120b0e01b',
   })
 
   depends_on 'unzip' => :build
   depends_on 'gawk' => :build
   depends_on 'dejagnu' => :build # for test
-  #depends_on 'gcc' => :build   # gcc version 4.9.4
+  depends_on 'gcc8' => :build # gcc version 8.2.0
```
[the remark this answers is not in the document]
uberhacker: Dude, how are you going to compile gcc without gcc?

packages/gcc8.rb
uberhacker: Where was your patch when I was trying to compile on i686?  This fixes the issue and using line numbers works.  This "patch" may not even be needed for the next version.

packages/gcc8.rb
uberhacker: I disagree that `#{CREW_PREFIX}` is not needed.  If you are suggesting using relative symlinks is better, that train has already passed.

JL2210: This is a case where it is almost certainly better. If you remove the file it's pointing to or move the symlink itself, either way it's going to break stuff. This would just be easier if you moved both files to another directory, i.e. Installing Chromebrew in `/usr/local` and moving it to `/tmp/usb_drive_mount`.

packages/gcc8.rb
```diff
@@ -0,0 +1,160 @@
...
+  depends_on 'icu4c' => :build # icu version 62.1
+  depends_on 'python27' => :build
+  depends_on 'python3' => :build
+
+  depends_on 'binutils'
+  depends_on 'gmp'
+  depends_on 'mpfr'
+  depends_on 'mpc'
+  depends_on 'isl'
+  depends_on 'cloog'
```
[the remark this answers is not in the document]
uberhacker: Carryover from gcc7.

uberhacker:
I believe it enables all available languages by default.  I don't think that specific flag exists anymore either.

packages/gcc8.rb
uberhacker: Why would we need to bootstrap?

packages/gcc7.rb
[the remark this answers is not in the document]
JL2210: It's not like you're ever going to merge a core package without binaries.

packages/gcc8.rb
JL2210: Probably not. They already have it fixed in `master`.

packages/gcc8.rb
JL2210: So GCC 8 can be built with GCC 8 instead of GCC 7. It's a performance thing. It takes longer at build, but it pays off in the long run.

packages/gcc8.rb
JL2210: Probably. What about handling spaces in paths? I know symlinks are really bad at that right now.

JL2210:
`--enable-languages` sure does. It would break sh*t if they removed it. If you're going to say something like that, please check first.
It only enables the needed languages: C and C++. Any other language isn't enabled if it isn't needed.

packages/gcc7.rb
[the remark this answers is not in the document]
uberhacker: And how do you think we compile the binaries?

packages/gcc8.rb
uberhacker: If you put spaces in your directories, you are just doing something unnecessary which will cause pain.

uberhacker:
I did check and there is no documented option.  I think you are referring to gcc7.  This is gcc8.

packages/gcc7.rb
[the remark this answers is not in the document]
JL2210: By installing Chromebrew first? It wouldn't be such an issue if you'd pass `--enable-bootstrap`.

JL2210:
What's this, then?
```
chronos@localhost ~/Downloads/chromebrew/tools/musl_gcc/gcc-8.2.0 $ cat configure | grep enable-languages
# Look if the user specified --enable-languages="..."; if not, use
        echo configure.ac: warning: setting LANGUAGES is deprecated, use --enable-languages instead 1>&2
      echo configure.ac: --enable-languages needs at least one language argument 1>&2
  ac_configure_args=`echo " $ac_configure_args" | sed -e "s/ '--enable-languages=[^ ]*'//g" -e "s/$/ '--enable-languages="$enable_languages"'/" `
chronos@localhost ~/Downloads/chromebrew/tools/musl_gcc/gcc-8.2.0 $ 
```

JL2210:
There may be no documented option, but it's still there and being used.

uberhacker:
Who in their right mind would need a Mac runtime on a Chromebook?  Where are you getting your options list?

PR #3128, opened 2019-01-30
Update FreeRDP package

### LINKED ISSUE #3114: [freerdp] Unable to build
```
[ 38%] Linking C shared library libwinpr2.so
[ 38%] Built target winpr
make: *** [Makefile:152: all] Error 2
freerdp failed to build: `make V=0 -j2` exited with 2
```

JL2210:
Fixes #3114

### DIFF
--- a/packages/freerdp.rb
+++ b/packages/freerdp.rb
@@ -2,29 +2,51 @@ require 'package'
 
 class Freerdp < Package
   description 'FreeRDP is a free implementation of the Remote Desktop Protocol.'
-  homepage 'http://www.freerdp.com/'
-  version '2.0.0-rc4'
+  homepage 'https://www.freerdp.com/'
+  version '2.0.0-rc4-1'
   source_url 'https://github.com/FreeRDP/FreeRDP/archive/2.0.0-rc4.tar.gz'
   source_sha256 '3406f3bfab63f81c1533029a5bf73949ff60f22f6e155c5a08005b8b8afe6d49'
 
-  depends_on 'ffmpeg'
-  depends_on 'gst_plugins_base'
   depends_on 'cups'
+  depends_on 'mesa'
   depends_on 'xmlto'
-  depends_on 'libxslt'
   depends_on 'ninja'
   depends_on 'xprop'
-  depends_on 'sommelier'
+  depends_on 'ffmpeg'
+  depends_on 'linux_pam'
+  depends_on 'pulseaudio'
+  depends_on 'libmp3lame'
+  depends_on 'gst_plugins_base'
 
   def self.build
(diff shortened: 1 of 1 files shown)


packages/freerdp.rb
```diff
@@ -2,29 +2,53 @@
...
   depends_on 'cups'
+  depends_on 'mesa'
   depends_on 'xmlto'
-  depends_on 'libxslt'
   depends_on 'ninja'
   depends_on 'xprop'
-  depends_on 'sommelier'
+  depends_on 'ffmpeg'
+  depends_on 'libsoxr'
+  depends_on 'xorg_lib'
```
uberhacker: libsoxr and xorg_lib come along for the ride with ffmpeg so not needed here.

JL2210: All right. In case you're wondering why I removed the `sommelier` dependency, it's because I feel that users should have a choice on what X server they use.

packages/freerdp.rb
```diff
@@ -2,29 +2,53 @@
...
+  depends_on 'linux_pam'
+  depends_on 'pulseaudio'
+  depends_on 'libmp3lame'
+  depends_on 'gst_plugins_base'
```
uberhacker: You sure we don't need sommelier?

JL2210: Absolutely. It isn't called in the application, and users should have a choice if they want to use another window manager (`Xwayland`). I brought in all the reasonable dependencies of it, so it should build fine.

packages/freerdp.rb
```diff
@@ -2,29 +2,51 @@
...
-  version '2.0.0-rc4'
+  homepage 'https://www.freerdp.com/'
+  version '2.0.0-rc4-1'
   source_url 'https://github.com/FreeRDP/FreeRDP/archive/2.0.0-rc4.tar.gz'
   source_sha256 '3406f3bfab63f81c1533029a5bf73949ff60f22f6e155c5a08005b8b8afe6d49'
 
-  depends_on 'ffmpeg'
-  depends_on 'gst_plugins_base'
   depends_on 'cups'
+  depends_on 'mesa'
```
uberhacker: Actually, mesa is not needed either since it is already indirectly a dependency of gst_plugins_base.

packages/freerdp.rb
uberhacker: Building successfully and actually executing are two separate things.

JL2210: 🙄 It runs fine.

packages/freerdp.rb
JL2210: ... Probably not the *best* idea to remove it...

Issue #185, opened 2016-06-11
Update the toolchains


lyxell:
I've cross compiled a gcc-toolchain for ARM and created a repository for it (see https://github.com/lyxell/chromeos-toolchains, thanks @bfayers for all the testing).

While I was at it I also cross compiled the same toolchain for x86_64 and i686.

These toolchains uses musl (https://www.musl-libc.org/) instead of glibc to create encapsulation inside /usr/local. Unlike glibc, musl will not interfere with the native libraries and the native libraries will not interfere with musl, this should solve #49 #56 #103 #131 #142 #171 and similar issues. It should also get rid of this https://github.com/skycocker/chromebrew/search?utf8=%E2%9C%93&q=shortarch

Musl also makes cross compiling possible which means we are able to update the toolchain for all architectures from a single machine in one go.

GCC is bumped to the latest version of the 4.x branch (4.9.3).

If people would like to try it out the tarballs can be found here: https://github.com/lyxell/chromeos-toolchains/blob/master/README.md. I've not included a C++ compiler included yet, so nodejs won't compile, but everything else in Chromebrew should work fine. If you run into an issue you can file it inside that repository so we don't clutter this thread, searching through #168 could also be helpful.

What are your general thoughts on this? Are people on x86_64 and i686 be interested in updating their toolchains as well?

minektur:
I personally would like a self-hosting compiler build.  I'm fine using a boot-strapped something or other from crouton, but we should be able to get to a gcc build with the gcc we install - e.g. start with the cross-compiled one, build a temporary gcc and use that to build a 'real' gcc which should be the same as the temporary one... Or something like that.

lyxell:
@minektur I'm not really sure on what you mean, could you elaborate? Why would there be any advantages of using a self-hosted build? The rest of the OS is cross compiled with the same settings.

minektur:
well.... call me a purist? but... Ok I found someone else who explained it pretty well.  Read this http://programmers.stackexchange.com/questions/263651/why-are-self-hosting-compilers-considered-a-rite-of-passage-for-new-languages    stackoverflow question and find the answer by DefinitelyNotSteve, where he paraphrases steve yegge.

Here is my summary of his summary and thoughts, well point 3 anyway.  It is an excellent way to make sure it actually works.  Building a compiler is a complicated task.  If the compiler can build itself and get a copy of itself out then you know its working, right?  

Quoting from the 'linux from scratch' build manual:

make bootstrap
[...]

PR #3131, opened 2019-01-30
Fix dependency loops and update gcc7 and gcc8 packages

JL2210:
`librsvg` is not used in Cairo, and `llvm` can build without GraphViz.

Includes an update to gcc 7 and 8 that uses more Ruby than shell and enables all languages.
Changes `/usr/lib` to `#{CREW_LIB_PREFIX}`.
Uses `Dir.glob` instead of asterisks.
Changes symlinks for gcc 7 from `7.3` to `7.4`.
Makes gcc 7 similar to gcc 8.
Removes unneeded dependencies.

### DIFF
--- a/packages/gcc7.rb
+++ b/packages/gcc7.rb
@@ -3,28 +3,16 @@ require 'package'
 class Gcc7 < Package
   description 'The GNU Compiler Collection includes front ends for C, C++, Objective-C, Fortran, Ada, and Go.'
   homepage 'https://www.gnu.org/software/gcc/'
-  version '7.4.0'
+  version '7.4.0-1'
   source_url 'https://ftpmirror.gnu.org/gcc/gcc-7.4.0/gcc-7.4.0.tar.xz'
   source_sha256 'eddde28d04f334aec1604456e536416549e9b1aa137fc69204e65eb0c009fe51'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gcc7-7.4.0-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/gcc7-7.4.0-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/gcc7-7.4.0-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/gcc7-7.4.0-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: '07126320a193d947aeac51f1827ff954c322a8f43014f3cff146715b72f9d7f8',
-     armv7l: '07126320a193d947aeac51f1827ff954c322a8f43014f3cff146715b72f9d7f8',
-       i686: '72cd956d055ad418b052dab407ec3a2b16663b09c396073a23bd12828fcf03c3',
-     x86_64: '6cce57e98836ccc3c6773df1627bca2102b4f89c7abf64dc47f7b97120b0e01b',
   })
 
-  depends_on 'unzip' => :build
-  depends_on 'gawk' => :build
   depends_on 'dejagnu' => :build # for test
-  depends_on 'gcc8' => :build # gcc version 8.2.0
-  depends_on 'icu4c' => :build # icu version 62.1
   depends_on 'python27' => :build
   depends_on 'python3' => :build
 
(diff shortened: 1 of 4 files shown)


packages/gcc7.rb
```diff
@@ -36,125 +24,91 @@ class Gcc7 < Package
...
 
   def self.build
-    # previous compile issue
-    # #{CREW_PREFIX}/bin/ld: cannot find crti.o: No such file or directory
-    # #{CREW_PREFIX}/bin/ld: cannot find /usr/lib64/libc_nonshared.a
-    ENV["LIBRARY_PATH"] = "#{CREW_LIB_PREFIX}"   # fix x86_64 issues
-    system "mkdir -p objdir"
-    Dir.chdir("objdir") do
+    ENV['LIBRARY_PATH'] = CREW_LIB_PREFIX   # fix x86_64 issues
+    system 'mkdir -p build'
```
uberhacker: Since you are in a remove all system commands mood: `Dir.mkdir('build')`?

packages/llvm.rb
```diff
@@ -21,7 +21,6 @@ class Llvm < Package
   })
 
   depends_on 'python27' => :build # for test suite
-  depends_on 'graphviz' => :build
```
uberhacker: You sure this is not needed?  This is required for the `dot` command.  It's only a build dependency also.

JL2210: Yes, but it can be disabled and it causes a load of dependency loops.

uberhacker: It shouldn't cause a dependency loop as a build dependency or are you saying it causes the looping when you try to compile?

JL2210: When I try to compile.

packages/gcc7.rb
JL2210: done

JL2210:
This is mainly the reason why I pass `--enable-bootstrap`: https://github.com/skycocker/chromebrew/issues/185#issuecomment-226679762

JL2210:
Are you absolutely opposed to any of this? Is there anything obviously wrong that wouldn't allow this to be merged (other than not having binaries)?

uberhacker:
Your branch name doesn't describe the changes.

packages/gcc7.rb
```diff
@@ -36,125 +24,91 @@ class Gcc7 < Package
...
-      system "ln -sv #{CREW_PREFIX}/bin/$(gcc -dumpmachine)-gcc-nm-7.3 #{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc-nm"
-      system "ln -sv #{CREW_PREFIX}/bin/$(gcc -dumpmachine)-gcc-ranlib-7.3 #{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc-ranlib"
-      system "ln -sv #{CREW_PREFIX}/bin/$(gcc -dumpmachine)-gfortran-7.3 #{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gfortran"
+      FileUtils.ln_s("$(gcc -dumpmachine)-c++-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-c++")
+      FileUtils.ln_s("$(gcc -dumpmachine)-g++-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-g++")
+      FileUtils.ln_s("$(gcc -dumpmachine)-gcc-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc")
+      FileUtils.ln_s("$(gcc -dumpmachine)-gcc-ar-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc-ar")
+      FileUtils.ln_s("$(gcc -dumpmachine)-gcc-nm-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc-nm")
+      FileUtils.ln_s("$(gcc -dumpmachine)-gcc-ranlib-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc-ranlib")
+      FileUtils.ln_s("$(gcc -dumpmachine)-gfortran-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gfortran")
```
uberhacker: Good catch with the release version.  I'm wondering if we should use a variable to prevent this from happening in the future.

packages/gcc7.rb
```diff
@@ -3,28 +3,16 @@
...
-    aarch64: '07126320a193d947aeac51f1827ff954c322a8f43014f3cff146715b72f9d7f8',
-     armv7l: '07126320a193d947aeac51f1827ff954c322a8f43014f3cff146715b72f9d7f8',
-       i686: '72cd956d055ad418b052dab407ec3a2b16663b09c396073a23bd12828fcf03c3',
-     x86_64: '6cce57e98836ccc3c6773df1627bca2102b4f89c7abf64dc47f7b97120b0e01b',
   })
 
-  depends_on 'unzip' => :build
-  depends_on 'gawk' => :build
   depends_on 'dejagnu' => :build # for test
-  depends_on 'gcc8' => :build # gcc version 8.2.0
```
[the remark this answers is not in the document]
JL2210: We're going to add binaries before we merge this, just like any core package. If you do this, there's a dependency loop. Also, Chromebrew always installs a compiler during installation, so you should have no problem running `crew build gcc...`

JL2210:
See #3134

packages/gcc8.rb
```diff
@@ -20,11 +20,7 @@ class Gcc8 < Package
      x86_64: '6d982f3c35acd3a738e176a3f3b8c4ea093c7be86882153a7711841cf0c80737',
   })
 
-  depends_on 'unzip' => :build
-  depends_on 'gawk' => :build
   depends_on 'dejagnu' => :build # for test
-  depends_on 'gcc7' => :build # gcc version 7.4.0
```
uberhacker: You won't get past the configure script without a compiler.

JL2210: We're going to add binaries before we merge this, just like any core package. If you do this, there's a dependency loop. Also, Chromebrew always installs a compiler during installation, so you should have no problem running `crew build gcc...`

JL2210: One of the main reasons I did this is because I don't want a compiler to depend on another compiler. If we did that, it could break stuff, including having two conflicting compilers installed on one system. If we have two conflicting compilers on one system, there would be library mismatches. Also, if we removed one compiler, half of the other would also be removed. We really need to figure out a way to handle file collisions.

Issue #3100, opened 2019-01-20
[feature request]: New --check option


JL2210:
To execute the `def self.check` section, we currently need to use `crew build <package-name>`. If we added a new `--check` option, we could run this command to run the checks while installing a package: `crew install --check <package-name>`. Is this possible?

It would be preferable to eventually have the `--check` option be the default for `crew install`.

uberhacker:
We definitely don't want to check by default for `crew install`, especially when installing from binaries.  It might make sense to have the option available when installing from source, however.  A new command would be overkill.  The switch option makes more sense.  Just my two cents.

JL2210:
> A new command would be overkill.

Fair enough.

> We definitely don't want to check by default for crew install, especially when installing from binaries. It might make sense to have the option available when installing from source, however.

Sorry, I forgot to elaborate: Only when installing from sources.

> The switch option makes more sense.

That seems reasonable.

skycocker:
Do we want to abort the build if any of the checks fail?

JL2210:
Yes.

uberhacker:
@skycocker: That would be reasonable.  I envision this will only be needed when installing from source, not from binaries.  That said, we should do a good job of checking during builds so there are no errors.

cstrouse:
I've ran into quite a few projects where the tests aren't well-maintained or don't work on odd configurations such as the Samsung Chromebook Plus V1 (64-bit kernel and 32-bit armv7l userland). One that comes to mind was `memcached` where the tests had broken alignment checks in their `Makefile` and would always fail even though the software compiled and worked properly. I certainly don't want the check step enabled by default on my machine.

uberhacker:
Yeah, tests are usually overrated and can provide a false sense of success.  If the tests are thorough, I can see the value but most tests are inadequate.

cstrouse:
They also seem to be less useful on ChromeOS devices since the software and hardware configurations are often quite different from the main configurations that they develop on, test on, and support.

Zopolis4:
Also, for the packages where we do still run tests, we tend to disable shipping them, so the tests wouldn't be available in the precompiled binary and thus cannot be run from crew install.

satmandu:
We could package the build archives. 😅

Zopolis4:
We could...

Zopolis4:
I don't think its worth shipping test binaries to all users so that some tests can be ran without building a package.

PR #3132, opened 2019-01-30
#3100 Allow post-build package checks run during installation

skycocker:
Solves #3100. Aborts the installation if the flag is provided and any of the checks fail. Runs the checks for dependencies as well, if any are being built.

Please test it on actual chromebook before merging, since I currently don't have one.

### DIFF
--- a/crew
+++ b/crew
@@ -722,8 +722,8 @@ def install
       if @opt_check && Dir.exist?(target_dir)
         Dir.chdir target_dir do
           puts "Running post-build checks for package #{@pkg.name}..."
-          abort("#{@pkg.name} failed post-build checks. Aborting.") unless @pkg.check
-          puts "Checks for #{@pkg.name} have passed successfully."
+          abort("#{@pkg.name} failed post-build checks. Aborting.".lightred) unless @pkg.check
+          puts "Checks for #{@pkg.name} have passed successfully.".lightgreen
         end
       end
 
--- a/crew
+++ b/crew
@@ -25,7 +25,7 @@ Usage:
   crew download [-v|--verbose] <name> ...
   crew files <name> ...
   crew help [<command>]
-  crew install [-k|--keep] [-s|--build-from-source] [-S|--recursive-build] [-v|--verbose] <name> ...
+  crew install [-k|--keep] [-s|--build-from-source] [-S|--recursive-build] [-v|--verbose] [-c|--check] <name> ...
   crew list (available|installed)
   crew postinstall <name> ...
   crew remove [-v|--verbose] <name> ...
@@ -36,6 +36,7 @@ Usage:
 
   -k --keep               Keep the `CREW_BREW_DIR` (#{CREW_BREW_DIR}) directory.
   -s --build-from-source  Build from source even if pre-compiled binary exists.
+  -c --check              Run checks on the package before installing
   -S --recursive-build    Build from source, including all dependencies, even if pre-compiled binaries exist.
   -V --version            Display the crew version.
   -v --verbose            Show extra information.
@@ -79,9 +80,10 @@ rescue Docopt::Exit => e
   exit 1
 end
 
-@opt_keep = args["--keep"]
-@opt_verbose = args["--verbose"]
-@opt_src = args["--build-from-source"]
+@opt_keep      = args["--keep"]
+@opt_verbose   = args["--verbose"]
+@opt_src       = args["--build-from-source"]
+@opt_check     = args["--check"]
 @opt_recursive = args["--recursive-build"]
 
 @device = JSON.parse(File.read(CREW_CONFIG_PATH + 'device.json'), symbolize_names: true)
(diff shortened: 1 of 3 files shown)


JL2210:
Checking now.

crew
```diff
@@ -717,6 +719,14 @@ def install
       # CREW_DEST_DIR contains usr/local/... hierarchy
       build_and_preconfigure target_dir
 
+      if @opt_check && Dir.exist?(target_dir)
+        Dir.chdir target_dir do
+          puts "Running post-build checks for package #{@pkg.name}..."
+          abort("#{@pkg.name} failed post-build checks. Aborting.") unless @pkg.check
```
uberhacker: Can we add `.lightred` to the end to highlight the error?

crew
```diff
@@ -717,6 +719,14 @@ def install
...
+          puts "Checks for #{@pkg.name} have passed successfully."
```
uberhacker: Can we add `.lightgreen` to the end to highlight success?

uberhacker:
@skycocker: Thank you for the PR!  Please see inline comments.

JL2210:
It should fail if the `def self.check` section does not exist.

skycocker:
@uberhacker added the colors. 
@JL2210 could you try now?

JL2210:
Nice. Now bump the `crew` version in `lib/const.rb` from `1.1.2` to `1.2.0`.

JL2210:
Testing...

JL2210:
Now could you rebase and squash your commits?

uberhacker:
No need to squash commits @skycocker.  I can do that when I merge the PR.

skycocker:
Ok, so let's just wait until @JL2210 verifies it works.

JL2210:
Wait, you say I have to either squash my commits when I push my PR or only have one commit, but @skycocker is fine?

uberhacker:
I ask for the commits to be squashed if the comments are not meaningful.  @skycocker does a great job of commenting each commit which enhances the git log history.

JL2210:
Oh come on. I try to keep my commits short and sweet so they don't fill up the log.

uberhacker:
Oh, you mean like `updated package` and `fixed error`?  That only clutters up the log history with useless information and is not helpful to determine the actual change.

JL2210:
😆 All right, I'll make my commit logs longer.

JL2210:
Usually, I put all my comments on the third line and below so as not to fill up the subject line.

lib/package.rb
```diff
@@ -97,7 +97,9 @@ def self.postinstall
   # Function to perform check from source build.
   # This is execute if and only if `crew build`.
   def self.check
-
+    # blindly assume the package is OK
+    # if it has no checks at all
+    true
```
uberhacker: Maybe not needed?

crew
```diff
@@ -79,9 +80,10 @@ rescue Docopt::Exit => e
   exit 1
 end
 
-@opt_keep = args["--keep"]
-@opt_verbose = args["--verbose"]
-@opt_src = args["--build-from-source"]
+@opt_keep      = args["--keep"]
+@opt_verbose   = args["--verbose"]
+@opt_src       = args["--build-from-source"]
+@opt_check     = args["--check"]
```
uberhacker: Can we add this to the install section in help?
```
   when "install"
     puts "Install package(s)."
+    puts "Usage: crew install [-c|--check] [-k|--keep] [-s|--build-from-source] [-S|--recursive-build] [-v|--verbose] <package1> [<package2> ...]"
     puts "The package(s) must have a valid name.  Use `crew search <pattern>` to search for packages to install."
+    puts "If `-c` or `--check` is present, run checks on the package before installing from source.  Must also include the `-s` or `-S` option."
```

crew
```diff
@@ -717,6 +719,14 @@ def install
...
       build_and_preconfigure target_dir
 
+      if @opt_check && Dir.exist?(target_dir)
+        Dir.chdir target_dir do
+          puts "Running post-build checks for package #{@pkg.name}..."
+          abort("#{@pkg.name} failed post-build checks. Aborting.".lightred) unless @pkg.check
+          puts "Checks for #{@pkg.name} have passed successfully.".lightgreen
+        end
+      end
+
```
uberhacker: Maybe add this before the `resolve_dependencies_and_install` line in the `def install_package` function?
```
abort "The `-c` or `--check` option must also include either `-s` or `-S`".lightred if @opt_check and !@pkg.build_from_source
```

uberhacker: I was wondering whether it would be worth it to enable checks for every source build if the `-S` option is present?

skycocker:
@uberhacker the whole thing has to be rebuilt, I guess I'll close this PR until we rewrite the whole `crew` file - I'm looking at the method defined inside other method and I have such big WTF in my head that I believe it's kind of pointless to just add junk to it instead of rewriting it properly.

Issue #3126, opened 2019-01-29
Code package problems on armv7l


themoonisacheese:
EDIT: this issue was originally for a problem with computing dependencies for the `code` package, but a fix has since been applied. There are some more issues described in this thread that have not been fixed, please look through the thread before dismissing it as not helpful. Most of these issues appear to be armv7l related, but they may still apply to you anyway.

## Description
i'm trying to install vs code.
when doing
```shell
crew install code
```
nothing happens for a long time (literally half an hour) and it finally fails:

Output:
```shell
~ [ crew install code                                                                                                                       ] 11:10 AM
code: Visual Studio Code is a source code editor developed by Microsoft for Windows, Linux and macOS.
https://code.visualstudio.com/
version 1.30.1-1
Traceback (most recent call last):
        12679: from /usr/local/bin/crew:988:in `<main>'
        12678: from /usr/local/bin/crew:919:in `install_command'
        12677: from /usr/local/bin/crew:919:in `each'
        12676: from /usr/local/bin/crew:923:in `block in install_command'
        12675: from /usr/local/bin/crew:612:in `resolve_dependencies_and_install'
        12674: from /usr/local/bin/crew:669:in `resolve_dependencies'
        12673: from /usr/local/bin/crew:663:in `expand_dependencies'
        12672: from /usr/local/bin/crew:657:in `push_dependencies'
         ... 12667 levels...
            4: from /usr/local/lib/ruby/2.5.0/find.rb:49:in `block (2 levels) in find'
            3: from /usr/local/bin/crew:148:in `block in search'
            2: from /usr/local/bin/crew:109:in `set_package'
            1: from /usr/local/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:39:in `require'
/usr/local/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:123:in `rescue in require': stack level too deep (SystemStackError)
```

## Versions and system information
chromebrew: `crew --version`
1.1.2
ChromeOS: `cat /etc/lsb-release`
CHROMEOS_AUSERVER=https://tools.google.com/service/update2
CHROMEOS_BOARD_APPID={2FC0F0CF-1A55-DF3F-73E6-517389444085}
CHROMEOS_CANARY_APPID={90F229CE-83E2-4FAF-8479-E368A34938B1}
CHROMEOS_DEVSERVER=
[...]

PR #3133, opened 2019-01-30
Update Code package

JL2210:
No shocking fixes in this other than allowing the compilation to finish on ARM. Still trying to get to the bottom of why it's not displaying anything on the screen, though.

Tested on:
- [x] x86_64
- armv7l (partially, see #3126)


uberhacker:
This is already broke so it can't be any worse.

PR #3134, opened 2019-01-31
Fix dependency loops and update gcc7 and gcc8 packages

JL2210:
librsvg is not used in Cairo, and llvm can build without GraphViz.

Includes an update to gcc 7 and 8 that uses more Ruby than shell and enables all languages.
Changes /usr/lib to #{CREW_LIB_PREFIX}.
Uses Dir.glob instead of asterisks.
Changes symlinks for gcc 7 from 7.3 to 7.4.
Makes gcc 7 similar to gcc 8.
Removes unneeded dependencies.


JL2210:
What are you doing?

JL2210:
You missed `      FileUtils.ln_s("../../libexec/gcc/$(gcc -dumpmachine)/<version here>/liblto_plugin.so", "#{CREW_DEST_LIB_PREFIX}/bfd-plugins/")`

uberhacker:
You cannot set ulimit like you are doing in the check section because it generates an error.  Of course, you never tested this so you wouldn't have known.

uberhacker:
I already did and you ignored my question.

JL2210:
When? I opened another PR to fix one of the issues you had: `Your branch name doesn't describe the changes.`
I don't see you saying anything anywhere about anything else. And, along with that, I have to sleep.

JL2210:
This is why I disable "Allow edits from maintainers."

JL2210:
```
chronos@localhost ~/Downloads/chromebrew $ ulimit -s 32768
chronos@localhost ~/Downloads/chromebrew $ echo $?
0
```

uberhacker:
And what happens when you build?  This is why I close PRs that aren't ready and need more work.

JL2210:
All right, I'll test it. Actually, I'll push a binary to Bintray just to show you that it works.

packages/gcc7.rb
```diff
@@ -70,7 +70,6 @@ def self.build
 
   def self.check
     Dir.chdir('build') do
-      system 'ulimit -s 32768'
```
JL2210: You couldn't have just told me that `ulimit` is already set at unlimited by default?

packages/gcc7.rb
```diff
@@ -90,25 +89,25 @@ def self.install
...
+      FileUtils.ln_s("gfortran-#{version}", "#{CREW_DEST_PREFIX}/bin/gfortran")
 
-      FileUtils.ln_s("$(gcc -dumpmachine)-c++-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-c++")
-      FileUtils.ln_s("$(gcc -dumpmachine)-g++-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-g++")
-      FileUtils.ln_s("$(gcc -dumpmachine)-gcc-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc")
-      FileUtils.ln_s("$(gcc -dumpmachine)-gcc-ar-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc-ar")
-      FileUtils.ln_s("$(gcc -dumpmachine)-gcc-nm-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc-nm")
-      FileUtils.ln_s("$(gcc -dumpmachine)-gcc-ranlib-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gcc-ranlib")
-      FileUtils.ln_s("$(gcc -dumpmachine)-gfortran-7.4", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-gfortran")
+      FileUtils.ln_s("$(gcc -dumpmachine)-c++-#{version}", "#{CREW_DEST_PREFIX}/bin/$(gcc -dumpmachine)-c++")
```
JL2210: I'm fairly sure this will break stuff. Packages expect `gcc-8.2`, not `gcc-8.2.0`.
Note: Above is just an example, probably not relevant for this package.

PR #3138, opened 2019-02-03
Fix bz2 installation errors

Ewpratten:
Previously, when installing bz2, it would fail due to `configure: error: C compiler cannot create executables`

It seems that installing gcc8 fixes this.


JL2210:
Are you on the beta, dev, or Canary channels?

JL2210:
`buildessential` depends on `gcc8` now.
I think we should really create a fake `gcc7` package that points to `gcc8` until we can merge them all into a single `gcc` package. Having two GCC versions installed is just not practical.

PR #3139, opened 2019-02-03
Add kcov package

akossomfai:
This is a request to add kcov to chromebrew.

Kcov is a code coverage tester for compiled programs, Python scripts and shell scripts. It allows collecting code coverage information from executables without special command-line arguments, and continuously produces output from long-running applications.

The main features of kcov are:
* Supports both compiled (ELF / Mach-O) binaries, Python programs and shell scripts (sh/bash)
* Collects code coverage information from binaries directly - no compilation options
* HTML and Cobertura XML output without processing steps
* With Cobertura output, kcov coverage data can be easily integrated in the Jenkins continuous integration server.
* kcov can upload coverage data automatically to coveralls.io, with easy integration into travis-ci.org.
* Shared libraries linked to the application (or opened through dlopen) are transparently covered
* kcov can accumulate data from multiple runs of the same program
* kcov also accumulates data from multiple programs, so that common source code can be analyzed in total
* kcov generates an updated web page every second, so long-running programs can be watched "live"
* kcov allows output to be sorted, either by coverage percentage or filename
* kcov allows specifying paths to include and exclude in the coverage, thereby limiting noise from system header files etc
* kcov allows multiple program coverage to be kept in a single HTML output directory, and will automatically add more programs as they are run
* build-id is automatically used to find source code, when applicable

### DIFF
--- a/packages/kcov.rb
+++ b/packages/kcov.rb
@@ -1,0 +1,24 @@
+require 'package'
+
+class Kcov < Package
+  description 'Kcov is a code coverage tester for compiled programs, Python scripts and shell scripts.'
+  homepage 'http://simonkagstrom.github.io/kcov/'
+  version 'v36'
+  source_url 'https://github.com/SimonKagstrom/kcov/archive/v36.tar.gz'
+  source_sha256 '29ccdde3bd44f14e0d7c88d709e1e5ff9b448e735538ae45ee08b73c19a2ea0b'
+
+  depends_on 'curl'
+
+  def self.build
+    system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX}"
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+
+  def self.check
+    system "make", "check"
+  end
+end
--- a/packages/kcov.rb
+++ b/packages/kcov.rb
@@ -8,17 +8,23 @@ class Kcov < Package
   source_sha256 '29ccdde3bd44f14e0d7c88d709e1e5ff9b448e735538ae45ee08b73c19a2ea0b'
 
   depends_on 'curl'
+  depends_on 'elfutils'
 
   def self.build
-    system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX}"
-    system "make"
+    Dir.mkdir 'build'
+    Dir.chdir 'build' do
+      system 'cmake',
+             '-DCMAKE_BUILD_TYPE=Release',
+             "-DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX}",
+             '..'
+      system 'make'
+    end
   end
 
   def self.install
-    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    Dir.chdir 'build' do
+      system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    end
   end
 
-  def self.check
-    system "make", "check"
-  end
 end


JL2210:
Thank you for the PR @akossomfai! Does this package install any libraries?

packages/kcov.rb
```diff
@@ -0,0 +1,24 @@
...
+  homepage 'http://simonkagstrom.github.io/kcov/'
+  version 'v36'
+  source_url 'https://github.com/SimonKagstrom/kcov/archive/v36.tar.gz'
+  source_sha256 '29ccdde3bd44f14e0d7c88d709e1e5ff9b448e735538ae45ee08b73c19a2ea0b'
+
+  depends_on 'curl'
+
+  def self.build
+    system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX}"
+    system "make"
```
JL2210: Could you please use single quotes around `make` instead of double quotes?

packages/kcov.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+
+  def self.check
+    system "make", "check"
```
JL2210: Single quotes, please.

JL2210:
I noticed that you're creating a pull request from the `master` branch. I would highly advise against that, because if you have any need to pull and rebase your Chromebrew branch later on, there may be merge conflicts. If you're working from the command line, you can use `git checkout -b branch-name` to switch to a new branch (`branch-name` in that example) while in your cloned Git repository.

uberhacker:
Yes, thank you for the PR @akossomfai.  I noticed the `kcov` command is missing without elfutils.  The changes below should help resolve the dependency gaps and ensure the package is complete.
```
require 'package'

class Kcov < Package
  description 'Kcov is a code coverage tester for compiled programs, Python scripts and shell scripts.'
  homepage 'http://simonkagstrom.github.io/kcov/'
  version 'v36'
  source_url 'https://github.com/SimonKagstrom/kcov/archive/v36.tar.gz'
  source_sha256 '29ccdde3bd44f14e0d7c88d709e1e5ff9b448e735538ae45ee08b73c19a2ea0b'

  depends_on 'curl'
  depends_on 'elfutils'

  def self.build
    Dir.mkdir 'build'
    Dir.chdir 'build' do
      system "cmake .. \
             -DCMAKE_BUILD_TYPE=Release \
             -DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX} \
             -DLIBBFD_IBERTY_LIBRARY=#{CREW_LIB_PREFIX}/libbsd.so"
      system 'make'
    end
  end

  def self.install
    Dir.chdir 'build' do
    end
  end

  def self.check
    Dir.chdir 'build' do
      system "make", "check"
    end
  end
end
```
See the attached build and files logs for a comparison.

[kcov.log](https://github.com/skycocker/chromebrew/files/2826017/kcov.log)
[kcov-new.log](https://github.com/skycocker/chromebrew/files/2826018/kcov-new.log)

[kcov-files.log](https://github.com/skycocker/chromebrew/files/2826024/kcov-files.log)
[kcov-files-new.log](https://github.com/skycocker/chromebrew/files/2826023/kcov-files-new.log)

uberhacker:
Try this without them and find out what happens.  Also, we don't need to be nitpicking about quotes.

uberhacker:
Interestingly, this works:
```
      system 'cmake',
             '-DCMAKE_BUILD_TYPE=Release',
             "-DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX}",
             "-DLIBBFD_IBERTY_LIBRARY=#{CREW_LIB_PREFIX}/libbsd.so",
             '..'
```
But this doesn't work:
```
      system 'cmake ..',
             '-DCMAKE_BUILD_TYPE=Release',
             "-DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX}",
             "-DLIBBFD_IBERTY_LIBRARY=#{CREW_LIB_PREFIX}/libbsd.so"
```

JL2210:
It should be fine.

> we don't need to be nitpicking about quotes.

Why are *you* telling *me* not to nitpick about quotes?

JL2210:
That's because `cmake ..` isn't a command. `cmake` is.

uberhacker:
That shouldn't make any difference.  The way strings are constructed before they are executed is what matters.  The extra "evaluation" under the hood is not necessary.

JL2210:
Technically, it does. First of all, it's an array of strings we're dealing with, so some strictly string-based rules go away. Basically, what it does is search `PATH` for `cmake ..`, just like it does if the first sting in the array is `VAR=value`. The strings are constructed as seperate arguments based on what element of the array they are in.

uberhacker:
You are missing the point.  Technically, it's inefficient if all it does is reconstruct all the array elements together into a final string that is executed in the end.  There is no need to do all that path searching until the string is constructed at the end.

akossomfai:
Hi, I've just changed it accordingly.

akossomfai:
> I noticed that you're creating a pull request from the `master` branch. I would highly advise against that, because if you have any need to pull and rebase your Chromebrew branch later on, there may be merge conflicts. If you're working from the command line, you can use `git checkout -b branch-name` to switch to a new branch (`branch-name` in that example) while in your cloned Git repository.

Yes, you're right, I'll do this way with any subsequent pull request.

packages/kcov.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  depends_on 'curl'
+  depends_on 'elfutils'
+
+  def self.build
+    Dir.mkdir 'build'
+    Dir.chdir 'build' do
+      system 'cmake',
+             '-DCMAKE_BUILD_TYPE=Release',
+             "-DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX}",
```
uberhacker: You need this line to get rid of the error in the log:
```
"-DLIBBFD_IBERTY_LIBRARY=#{CREW_LIB_PREFIX}/libbsd.so",
```

uberhacker: The error in the log I'm referring to is this:
```
-- Could NOT find LIBBFD (missing: LIBBFD_IBERTY_LIBRARY) 
```

JL2210: There's a difference between `libbfd` and `libbsd`. The first is the Binary File Descriptor library. The other is a collection of miscellaneous functions not found in GNU libc.

JL2210: If we were going to pass that option, we'd need to pass `--enable-shared --enable-install-libbfd --enable-install-libiberty` in Binutils, and it wouldn't use `libbsd.so`, either. In this case, the error is correct.

uberhacker: Yes, libbfd.so was the file.  That was a typo.  Your explanation makes sense.

PR #3143, opened 2019-02-08
Update GTK packages (gtk2, gtk3, gnome_icon_theme)

JL2210:
Removes unneeded dependency `gtk3` from `gnome_icon_theme` and switches the dependencies.
Should fix most issues for missing image box in place of image.

### DIFF
--- a/packages/gtk2.rb
+++ b/packages/gtk2.rb
@@ -1,38 +1,33 @@
 require 'package'
 
 class Gtk2 < Package
-  description 'Gtk+ 2.24 graphical user interface library.'
+  description 'GTK+ is a multi-platform toolkit for creating graphical user interfaces.'
   homepage 'https://www.gtk.org/'
-  version '2.24'
-  source_url 'https://ftp.gnome.org/pub/gnome/sources/gtk+/2.24/gtk+-2.24.32.tar.xz'
+  version '2.24.32'
+  source_url 'https://download.gnome.org/sources/gtk+/2.24/gtk+-2.24.32.tar.xz'
   source_sha256 'b6c8a93ddda5eabe3bfee1eb39636c9a03d2a56c7b62828b359bf197943c582e'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gtk2-2.24-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/gtk2-2.24-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/gtk2-2.24-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/gtk2-2.24-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'd3c7e104ceeb13827016447cfe6f3db5a5c3a8a50fc8be54f67e1893700bd388',
-     armv7l: 'd3c7e104ceeb13827016447cfe6f3db5a5c3a8a50fc8be54f67e1893700bd388',
-       i686: '9d961a6437d896cd1dbde21ebe4a20967ad0267621fbbb7d80cb36309fd7f232',
-     x86_64: '96e236c0057e01ac2790ec5881349a2fa4b6c0821fa3d9ae477a1471f0c1d1af',
   })
 
   depends_on 'gtk_doc'
   depends_on 'atk'
   depends_on 'pango'
-  depends_on 'cairo'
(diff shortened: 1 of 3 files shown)


packages/gtk2.rb
```diff
@@ -1,38 +1,33 @@
 require 'package'
 
 class Gtk2 < Package
-  description 'Gtk+ 2.24 graphical user interface library.'
+  description 'GTK+ is a multi-platform toolkit for creating graphical user interfaces.'
```
uberhacker: Much better description. 👍

packages/gtk2.rb
```diff
@@ -1,38 +1,33 @@
...
   depends_on 'pango'
-  depends_on 'cairo'
   depends_on 'gdk_pixbuf'
+  depends_on 'cups'
   depends_on 'six' => :build
 
   def self.build
-    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
+    system "./configure",
+           "--with-gdktarget=x11",
```
uberhacker: If we add this option shouldn't we also add some X11 dependency or does that get included on a fresh install?

packages/gtk2.rb
```diff
@@ -1,38 +1,33 @@
...
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}"
     system "make"
   end
 
   def self.install
-    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install" # the steps required to install the package
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
```
uberhacker: I'm surprised you didn't change the double quotes to single quotes. ;)

packages/gtk3.rb
```diff
@@ -1,47 +1,50 @@
...
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+
+  def self.postinstall
     # generate schemas
-    system "mkdir -p #{CREW_DEST_PREFIX}/share/glib-2.0/schemas"
-    system "glib-compile-schemas #{CREW_DEST_PREFIX}/share/glib-2.0/schemas"
+    system "glib-compile-schemas #{CREW_PREFIX}/share/glib-2.0/schemas"
+    system "update-mime-database #{CREW_PREFIX}/share/mime"
```
uberhacker: Do applications know to pull data from this directory?

packages/gtk3.rb
```diff
@@ -1,47 +1,50 @@
...
 
   def self.build
     system "./configure",
+           "--enable-cups",
+           "--disable-debug",
+           "--enable-x11-backend",
+           "--enable-introspection",
            "--prefix=#{CREW_PREFIX}",
-           "--libdir=#{CREW_LIB_PREFIX}",
+           "--enable-wayland-backend",
```
uberhacker: No wayland dependency needed?

JL2210: No. My guess is that its dynamically loaded.

packages/gtk3.rb
JL2210: Yes.

packages/gtk2.rb
JL2210: Sorry. Forgot.

packages/gtk2.rb
JL2210: I don't think it's needed.

JL2210: I don't think it's needed.

uberhacker:
You are losing credibility @JL2210.  Turns out the "unneeded" gtk3 dependency is needed after all in gnome_icon_theme:
```
checking for gtk-update-icon-cache... no
configure: error: Could not find gtk-update-icon-cache
gnome_icon_theme failed to build: `./configure --prefix=/usr/local --enable-icon-mapping` exited with 1
```
Please, please, please test these things on a fresh install before you submit a PR.  If you cannot do that, then do not suggest changes.  It wastes a lot of my time.

JL2210:
I removed gtk3 and it worked just fine. It might be one of the dependencies
of gtk3. Maybe gtk3 should be a build dependency?

JL2210:
And you know I CANNOT test things on a fresh install. If I did that every
time I tested something, I'd be able to do this for about two days before
my "unlimited" (screw AT&T) data ran out.

uberhacker:
Don't let your limited bandwidth stop you from a fresh install.  Just tarball your `/usr/local` directory and restore from backup.  That would be faster than downloading everything again anyway:
```
$ tar czvf fresh-crew.tar.gz /usr/local
... do your thing here ...
$ rm -rf /usr/local
$ tar xvf fresh-crew.tar.gz
```
Just remember to create a new tarball every time master is updated.

PR #3148, opened 2019-02-08
Update OpenSSL package

JL2210:
System provides older version for backwards compat. We can fix and propose upstream any source changes needed.

### DIFF
--- a/packages/openssl.rb
+++ b/packages/openssl.rb
@@ -3,70 +3,40 @@ require 'package'
 class Openssl < Package
   description 'OpenSSL is an open source project that provides a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols.'
   homepage 'https://www.openssl.org/'
-  version '1.0.2p-1'
-  source_url 'https://github.com/openssl/openssl/archive/OpenSSL_1_0_2p.tar.gz'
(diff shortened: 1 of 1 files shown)


uberhacker:
I don't think this is a good idea.  Please test on some packages that require it like irssi, php, ruby and tcpdump and report back the results.  Since this is a core package we'll need working binaries also.

JL2210:
Don't worry. All those packages have OpenSSL 1.1 support by now, and there's no binary incompatibility. If this and #3149 are merged, the core packages should be able to be linked statically, which would reduce packages needed during install to the minimum of just Git and Ruby.

uberhacker:
Doesn't ruby still have a gcc dependency or did you figure out how to remove it?  Still doesn't hurt to test.

JL2210:
That's my point: static linking **should** remove any external dependencies from the binary (a.k.a. `libssh2`, `libiconv`, and `gcc`).

JL2210:
Ruby works fine after upgrading.

JL2210:
And any newer PHP version (i.e not 5 or anything 7.2 or below) should compile and work just fine with this.
There also shouldn't be any issue with the binaries as long as the system still has openssl 1.0.

uberhacker:
Sorry, but this breaks everything.  I cannot compile php with this version of openssl.

uberhacker:
@skiman6010: Please do NOT merge this.  It is not ready yet.

skiman6010:
Haha don't tell me what to do @uberhacker. I merge what I want.

PR #3149, opened 2019-02-08
Update libiconv package

JL2210:


packages/libiconv.rb
```diff
@@ -3,32 +3,27 @@
...
   
-  depends_on 'glibc'
-  depends_on 'manpages'
-
   def self.build
-    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
-    system "make"
+    system './configure',
+           '--enable-shared',
+           '--enable-static',
```
uberhacker: Why do we need static libraries?

JL2210: Statically linking the core packages so as to reduce dependencies.

uberhacker: This update is still sitting here since I thought you were going to update ruby.

JL2210: Yeah, I know. Ruby's having some issues doing this right now, look:
```
~snip~
in `require': incompatible library version
~snip~
```

uberhacker: And this is the exactly the reason I close these kinds of PRs.  I waste a lot of my time testing PRs that break other packages.

JL2210: This doesn't break anything. It's just Ruby that's being a pain during build.

uberhacker: You just contradicted yourself.  This breaks ruby.

JL2210: No, Ruby breaks Ruby. This is fine.

JL2210:
@uberhacker Could you please merge this?

uberhacker:
Uh...not until we have binaries since this is a core package and you have not shared your branch.

JL2210:
Sorry about that. I usually disable edits from maintainers unless it's needed.

PR #3150, opened 2019-02-09
Rename zlibpkg -> libz and change packages accordingly

JL2210:


packages/libz.rb
```diff
@@ -0,0 +1,29 @@
+require 'package'
+
+class Libz < Package
```
uberhacker: I didn't like the `zlibpkg` name either but `zlib` is probably a better name since that is the project name.  It's an easy change, right?

uberhacker: Renaming packages is not a simple fix.  You also need to consider how to remove the older version, otherwise, the metadata hangs around and can cause a potential problem.

JL2210: This was gone over before. `zlib` is a reserved keyword in Ruby: https://ruby-doc.org/stdlib-2.6.1/libdoc/zlib/rdoc/Zlib.html

JL2210: You didn't seem to care with GCC and all the fake packages that you're removing.

uberhacker: It'd be nice to update `tools/core_packages.txt` also since this is a core package.

packages/squashfs.rb
```diff
@@ -20,12 +20,11 @@ class Squashfs < Package
      x86_64: 'b9f8bb91f7ab76540d0a8484cd3bc7ce64f90bea6279474d1e4a8f32c79f990f',
   })
 
-  depends_on 'compressdoc' => :build
   depends_on 'help2man' => :build
   depends_on 'lz4'
   depends_on 'lzo'
   depends_on 'xzutils'
-  depends_on 'zlibpkg'
+  depends_on 'libz'
```
uberhacker: This can just be removed altogether, right?

JL2210: I think so.

packages/libz.rb
uberhacker: If I remember correctly, wasn't it YOU who suggested all these changes to gcc in the first place?

uberhacker:
Not really worth it to try and work around reserved commands.  Either way, the package name is incorrect.

JL2210:
What the heck? That's not the only thing I changed.

JL2210:
This adds static libraries.
`libz` should work just fine.
And `zlibpkg` is not exactly an optimal name.

uberhacker:
Just submit a new clean up PR.  I know there are other things you don't like about some of these packages that aren't in this PR anyway.

JL2210:
Like what?

uberhacker:
libzpkg works just fine also.

uberhacker:
You will find something.

JL2210:
How is that any better? That's just appending `pkg` to the name `libz`, which is unneeded. If `libz` works, why not just use that? It's close enough.

uberhacker:
No, my point was it doesn't matter if we rename it to whatever, the package name still going to be wrong.

JL2210:
I'd say that `libz` is descriptive enough, and a search for `libz` on Google changes to `zlib` automatically. The package is clear enough by its description.

packages/libz.rb
JL2210: Not the ones you made. You just copied the already bad package for gcc7 into gcc8 and changed the numbers.

PR #3151, opened 2019-02-09
Update libssh2 package

JL2210:
Update to 1.8.1

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -56,13 +56,13 @@ case "${ARCH}" in
   sha256s+=('047163a67f3ea9e143b465087d5ea4a3d17b963c8ad3579bc27a21934d4a17ab')
   urls+=('https://dl.bintray.com/chromebrew/chromebrew/git-2.20.1-chromeos-armv7l.tar.xz')
   sha256s+=('149a43937a9be2a8af9ebc1ab9b14fb90ff361acf1c9641959fa1c6731f842a7')
-  urls+=('https://dl.bintray.com/chromebrew/chromebrew/libssh2-1.8.0-chromeos-armv7l.tar.xz')
-  sha256s+=('6fa84296583273dd9e749a2c54cb1cf688a7dab032e2528de5944a4d9777f037')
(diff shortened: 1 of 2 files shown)


uberhacker:
I'm confused.  How does git have anything to do with this package?  This openssl 1.1.0 update has opened a Pandora's box of worms (or should I say bugs?).  I would rather wait until a stable version of this package.

JL2210:
This is basically just preparing for the upgrade. If you don't mind, I'd like to keep it open so all the OpenSSL related pull requests can be merged at once.

uberhacker:
OK, but these kinds of updates tend to have a cascading effect to break everything.  If understand correctly, the intention here is to remove the dependency from git?

JL2210:
It's more of a "Kill two birds with one stone" deal, but yes.

JL2210:
Can this be merged? It doesn't break anything.

uberhacker:
Except the install?  Did you forget this is still a dependency in install.sh?  We would also need binaries for all architectures since this is a core package.  If you plan on removing the git dependency, wouldn't we also need an updated git?  All these things would then need to be updated in install.sh.

JL2210:
Can you add binaries, then? It still works with the old OpenSSL.

uberhacker:
See https://github.com/skycocker/chromebrew/pull/3151#issuecomment-464517728.

JL2210:
This enables static archives and paves the way to upgrading OpenSSL to 1.1.1. It's not causing any harm.

uberhacker:
Or so you think.  There must be a reason they haven't released a stable version yet.  Maybe because it's still unstable?

JL2210:
Nah, probably just really sparse release timing.

JL2210:
Stable version was released recently, please merge.

uberhacker:
This is more like it.  I'll try and build the binaries first since this is a core package.

skiman6010:
Good work on the fast update everyone!

PR #3152, opened 2019-02-09
Add cloc package

davidberard98:
Added cloc package; this is a script which counts the number of lines of code in a directory

https://github.com/AlDanial/cloc/releases

### DIFF
--- a/packages/cloc.rb
+++ b/packages/cloc.rb
@@ -7,8 +7,6 @@ class Cloc < Package
   source_url 'https://github.com/AlDanial/cloc/releases/download/1.80/cloc-1.80.tar.gz'
   source_sha256 '082f53530eee3f9ee84ec449eca59a77ff114250cd7daf9519679537b5b21d67'
 
-  depends_on 'perl'
-
   def self.install
     system "install -Dm755 cloc #{CREW_DEST_PREFIX}/bin/cloc"
   end
--- a/packages/cloc.rb
+++ b/packages/cloc.rb
@@ -1,0 +1,15 @@
+require 'package'
+
+class Cloc < Package
+  description 'cloc counts blank lines, comment lines, and physical lines of source code in many programming languages.'
+  homepage 'https://github.com/AlDanial/cloc'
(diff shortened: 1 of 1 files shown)


packages/cloc.rb
```diff
@@ -0,0 +1,15 @@
+require 'package'
+
+class Cloc < Package
+  description 'cloc counts blank lines, comment lines, and physical lines of source code in many programming languages.'
+  homepage 'https://github.com/AlDanial/cloc'
+  version '1.80'
+  source_url 'https://github.com/AlDanial/cloc/releases/download/1.80/cloc-1.80.tar.gz'
+  source_sha256 '082f53530eee3f9ee84ec449eca59a77ff114250cd7daf9519679537b5b21d67'
+
+  depends_on 'perl'
```
JL2210: Perl is already a "core" package, so this is not needed.

JL2210:
Thank you for the PR @davidberard98! Please see above comment.

JL2210:
I noticed that you're creating a pull request from the `master` branch. I would highly advise against that, because if you have any need to pull and rebase your Chromebrew branch later on, there may be merge conflicts. If you're working from the command line, you can use `git checkout -b branch-name` to switch to a new branch (`branch-name` in that example) while in your cloned Git repository.

davidberard98:
Whoops, I'll fix that and submit another pull request. Thanks!

JL2210:
It's not too big of a deal. Just remember to check out a new branch next time you create another pull request.

uberhacker:
Thank you for the new package @davidberard98!  If you add the following, you can create a manpage also:
```
  def self.build
    system 'make -C Unix'
    system 'gzip -9 Unix/cloc.1'
  end

  def self.install
    system "install -Dm755 cloc #{CREW_DEST_PREFIX}/bin/cloc"
    system "install -Dm644 Unix/cloc.1.gz #{CREW_DEST_PREFIX}/share/man/man1/cloc.1.gz"
  end
```

PR #3153, opened 2019-02-09
Add cloc package

davidberard98:
Added cloc package; this is a script which counts the number of lines of code in a directory

Moved to a different branch.

### DIFF
--- a/packages/cloc.rb
+++ b/packages/cloc.rb
@@ -7,8 +7,6 @@ class Cloc < Package
   source_url 'https://github.com/AlDanial/cloc/releases/download/1.80/cloc-1.80.tar.gz'
   source_sha256 '082f53530eee3f9ee84ec449eca59a77ff114250cd7daf9519679537b5b21d67'
 
-  depends_on 'perl'
-
   def self.install
     system "install -Dm755 cloc #{CREW_DEST_PREFIX}/bin/cloc"
   end
--- a/packages/cloc.rb
+++ b/packages/cloc.rb
@@ -1,0 +1,15 @@
+require 'package'
+
+class Cloc < Package
+  description 'cloc counts blank lines, comment lines, and physical lines of source code in many programming languages.'
+  homepage 'https://github.com/AlDanial/cloc'
+  version '1.80'
+  source_url 'https://github.com/AlDanial/cloc/releases/download/1.80/cloc-1.80.tar.gz'
+  source_sha256 '082f53530eee3f9ee84ec449eca59a77ff114250cd7daf9519679537b5b21d67'
+
+  depends_on 'perl'
+
+  def self.install
+    system "install -Dm755 cloc #{CREW_DEST_PREFIX}/bin/cloc"
+  end
+end


JL2210:
@uberhacker 👍?

JL2210:
Is there a testsuite of some sort or any configuration that may be needed?

JL2210:
It looks like you could use the `Makefile` in the `Unix/` directory to build, test, and install this.

JL2210:
May I suggest this?
```
  def self.build
    Dir.chdir('Unix') do
      system 'make'
    end
  end

  def self.install
    Dir.chdir('Unix') do
    end
  end

  def self.check
    Dir.chdir('Unix') do
      system 'make', 'test'
    end
  end
```

uberhacker:
This won't work @JL2210.  If you read the INSTALL, the Unix directory is for distros with package managers and the cloc executable there needs external dependencies.  Try it and find out yourself.

JL2210:
So Chromebrew **isn't** a package manager?

JL2210:
We should be able to package them. I don't see why not.

uberhacker:
Dude, they're not going to have instructions for building on Chromebrew.  Think about that for a moment.

JL2210:
They're just regular Perl packages, just like any other `perl_*` package (including `perl_io_socket_ssl`).

uberhacker:
And that is exactly why make doesn't work.  Are you going to try it?  Obviously, you don't believe me.

JL2210:
Sources:
https://cpan.metacpan.org/authors/id/T/TY/TYEMQ/Algorithm-Diff-1.1903.tar.gz
https://cpan.metacpan.org/authors/id/A/AB/ABIGAIL/Regexp-Common-2017060201.tar.gz
https://cpan.metacpan.org/authors/id/G/GA/GAAS/Digest-MD5-2.55.tar.gz
https://cpan.metacpan.org/authors/id/Y/YA/YANICK/Parallel-ForkManager-2.02.tar.gz

uberhacker:
You don't need to go to that extent to package this thing.

JL2210:
They honestly shouldn't be that hard.

JL2210:
@uberhacker How's that?

davidberard98:
Sorry for the delay; thanks for looking into this.

However, as far as I can tell, it seems like these dependencies aren't actually required when using this perl file for cloc.  It will install these dependencies in a temporary directory on startup if they aren't already installed
https://github.com/AlDanial/cloc#regexpcommon-digestmd5-win32autoglob-algorithmdiff-

Additionally, grepping through the file it appears that most the source code for the dependencies is included in the cloc file.

JL2210:
Yes, but we should be able to package those, too; after all, we are a
package manager. In fact, I have already pushed a PR with the dependencies.

uberhacker:
We could do that but it's total overkill.  @davidberard98 is correct.  The perl dependencies aren't needed with the cloc executable.  If we really want to package like other package managers, we only need the command and manpage.  Here is the output of files installed from the Debian version of the cloc package:
```
$ dpkg -L cloc
/usr/bin/cloc
/usr/share/doc/cloc/changelog.Debian.gz
/usr/share/doc/cloc/changelog.gz
/usr/share/doc/cloc/copyright
/usr/share/man/man1/cloc.1.gz
```

JL2210:
How exactly is it "total overkill"? For one, these could well be needed for other packages, and for another, I don't think that installing those in a temporary directory is any better than packaging them.

uberhacker:
The cloc executable is over 13k lines of perl code that includes all the dependencies already.

JL2210:
Yes, and the stripped-down `cloc` executable is more modular and allows for different versions of dependencies.

uberhacker:
@davidberard98: Have you decided how you want to proceed with this package?  Monolithic or modular?  That is the question.

uberhacker:
Closed due to inactivity.

PR #3154, opened 2019-02-09
Add perl_algorithm_diff, perl_digest_md5, perl_parallel_forkmanager, perl_regexp_common packages

JL2210:
Helps along #3153

### DIFF
--- a/packages/perl_algorithm_diff.rb
+++ b/packages/perl_algorithm_diff.rb
@@ -1,0 +1,23 @@
+require 'package'
+
+class Perl_algorithm_diff < Package
+  description 'Algorithm::Diff - Compute `intelligent\' differences between two files / lists'
+  homepage 'https://metacpan.org/pod/Algorithm::Diff'
+  version '1.1903'
+  source_url 'https://cpan.metacpan.org/authors/id/T/TY/TYEMQ/Algorithm-Diff-1.1903.tar.gz'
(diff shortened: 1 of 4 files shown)


uberhacker:
This is definitely overkill.  What do you think @davidberard98?  I want you to get the credit for adding this package.

JL2210:
I didn't actually add the Cloc package in this PR, just the dependencies for it. @davidberard98 can have as much credit as he wants for the package; after all, he is the one who wrote it.

JL2210:
Besides, all the package managers out there that I can think of also have these packages.

JL2210:
How exactly is this a work in progress? This can be merged and the specific packages work just fine.

uberhacker:
Well, I could merge these but since the cloc package is still a wip, it would be useless if he doesn't decide to go this route.

JL2210:
Let's just merge it. These may be useful some time in the future, anyway.

uberhacker:
I'm hesitant.

JL2210:
It's no big deal. At least they won't break anything.

uberhacker:
We can submit this again if these packages are needed in the future.

JL2210:
Or we could just merge it because there's no reason not to.

JL2210:
Except that `Parallel::ForkManager` requires `Moo::Role`. I'll add that.

JL2210:
There's a HUGE dependency tree for `Moo::Role`... This will take a while.

PR #3155, opened 2019-02-09
Add Linux_Sources package and recreate linuxheaders package based on it

JL2210:
@uberhacker Binaries, please.

I hope this makes you all happy :).

### DIFF
--- a/packages/linux_sources.rb
+++ b/packages/linux_sources.rb
@@ -1,0 +1,18 @@
+require 'package'
+
+class Linux_sources < Package
+  description 'Sources for the Linux kernel'
+  homepage 'https://kernel.org/'
+  version '3.18'
+  source_url 'https://www.kernel.org/pub/linux/kernel/v3.x/linux-3.18.tar.xz'
+  source_sha256 'becc413cc9e6d7f5cc52a3ce66d65c3725bc1d1cc1001f4ce6c32b69eb188cbd'
+
+  def self.install
+    linux_src_dir = CREW_DEST_PREFIX + '/src/linux'
+    FileUtils.mkdir_p(linux_src_dir)
+    FileUtils.cp_r('.', linux_src_dir)
+    Dir.chdir(linux_src_dir) do
+      system 'make', 'defconfig'
+    end
+  end
+end
--- a/packages/linuxheaders.rb
+++ b/packages/linuxheaders.rb
@@ -2,18 +2,37 @@ require 'package'
 
 class Linuxheaders < Package
   description 'Linux headers for Chrome OS.'
-  homepage ''
-  version '3.18-2'
+  homepage 'https://kernel.org/'
+  version '3.18'
+  source_url 'file:///dev/null'
+  source_sha256 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
+
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew-cross/linuxheaders-3.18-2-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew-cross/linuxheaders-3.18-2-chromeos-armv7l.tar.xz',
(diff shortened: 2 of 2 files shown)


packages/linuxheaders.rb
```diff
@@ -2,18 +2,24 @@
...
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew-cross/linuxheaders-3.18-2-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: '811e1bfa4d83e866ac7625c6970d7289b8768bf30b18af8a3b79c0760dfa42af',
-     armv7l: '811e1bfa4d83e866ac7625c6970d7289b8768bf30b18af8a3b79c0760dfa42af',
-       i686: '8ee5c68467cb5e7994993cddf163d59cb61a5f7d70f5bc595cb3fe288131ae59',
-     x86_64: '830b18e54ebd6a2c8701c5ea1efd883e345719770de34ad0ece2f7cfb904df6f',
-  })
+  homepage 'https://kernel.org/'
+  version '3.18-1'
```
uberhacker: This really should be `version '3.18-3'`.

JL2210: Oh, shoot. I meant to change that back to `3.18` after testing. I imagine by now that that version of the package is so ungodly old that nobody will have it installed.

JL2210:
I'm making some progress on Musl support. Once I gave up on multilib, the only issue stopping me now is disk space. It's unhelpful that Gentoo Prefix takes up so much space.

JL2210:
After this is merged, we should be able to delete https://bintray.com/chromebrew/chromebrew-cross

JL2210:
@uberhacker?

packages/linux_sources.rb
```diff
@@ -0,0 +1,24 @@
...
+  source_url 'https://www.kernel.org/pub/linux/kernel/v3.x/linux-3.18.tar.xz'
+  source_sha256 'becc413cc9e6d7f5cc52a3ce66d65c3725bc1d1cc1001f4ce6c32b69eb188cbd'
+
+  def self.install
+    case ARCH
+    when 'x86_64', 'x86'
+      arch = 'x86'
+    when 'armv7l', 'aarch64'
+      arch = 'arm'
+    end
```
uberhacker: Why do we define the `arch` variable and never use it?  Unless I'm missing something here?

JL2210: Sorry. Little mistake.

JL2210: Done.

JL2210:
I'm going to add a quick patch to this that should fix the Sommelier build process. Hang on.

JL2210:
Please build `linuxheaders` and this will be good. The patches that Chrome OS uses for the linuxheaders are way too invasive.

uberhacker:
What do you mean by invasive?  Can you be more specific?

JL2210:
The patches that Chromium applies to the linux sources are too invasive and would require major changes to the programs and libraries that use them.

uberhacker:
You are not being specific enough.  It what ways specifically are they being invasive?  We currently use this supposedly "invasive" linuxheaders package without any major changes.

JL2210:
Dude, read. I'm deciding against applying the Chromium patches for the Linux headers because the patches are invasive and will break stuff, not the headers. Build binaries and this is good.

uberhacker:
Notice during build:
```
rm: cannot remove '.': Is a directory
```
Probably should find a more elegant solution.

JL2210:
Does that make the build fail? It shouldn't affect anything.

uberhacker:
No but do you know how to fix to remove the notice?

JL2210:
Append `-not -type d` to the `find .` command.

JL2210:
Done.

JL2210:
@uberhacker? I've done all that you told me to.

uberhacker:
I held back on this because I want to do more testing.

JL2210:
Well, it works fine on my end. Any issues with the other architectures? I'd be happy to fix.

JL2210:
@uberhacker Just so you know, I've been using this since forever ago and it works fine. Do you mind building binaries for `linuxheaders`?

PR #3156, opened 2019-02-09
Update PCIUtils package

JL2210:


packages/pciutils.rb
```diff
@@ -2,35 +2,38 @@
...
   def self.build
-    system "sed -i 's,SBINDIR=\$(PREFIX)/sbin,SBINDIR=#{CREW_PREFIX}/bin,' Makefile"
-    system "make", "PREFIX=#{CREW_PREFIX}", "SHARED=yes", "ZLIB=yes"
+    system 'make',
+           "LIBDIR=#{CREW_LIB_PREFIX}",
+           'SBINDIR=$(PREFIX)/bin',
+           "PREFIX=#{CREW_PREFIX}",
+           'SHARED=yes',
+           'ZLIB=yes',
+           'DNS=yes'
```
uberhacker: I have a feeling this option will slow down the process.

JL2210: How so? I'm currently testing it on a Commodore 64 emulator.

uberhacker: Uh, dns requires a network to resolve ip addresses into domain names.  That takes time.  Not sure why we even need this option in the first place.

JL2210: It only enables DNS support. It doesn't call DNS unless its needed.

JL2210: The Commodore 64 emulator runs just fine without any network access.

PR #3157, opened 2019-02-10
Update OpenSSL package

JL2210:
Binary compatibility is kept with previous versions of OpenSSL.
**We can fix and/or propose upstream any source changes needed.**

### DIFF
--- a/packages/openssl.rb
+++ b/packages/openssl.rb
@@ -3,70 +3,40 @@ require 'package'
 class Openssl < Package
   description 'OpenSSL is an open source project that provides a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols.'
   homepage 'https://www.openssl.org/'
-  version '1.0.2p-1'
-  source_url 'https://github.com/openssl/openssl/archive/OpenSSL_1_0_2p.tar.gz'
-  source_sha256 '95ca65a25bdd41e127e5f4054539e8532a46be602b43b44af7c7100172e7cd50'
+  version '1.1.1a'
+  source_url 'https://www.openssl.org/source/openssl-1.1.1a.tar.gz'
+  source_sha256 'fc20130f8b7cbd2fb918b2f14e2f429e109c31ddd0fb38fc5d71d9ffed3f9f41'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/openssl-1.0.2p-1-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


JL2210:
Sorry, but OpenSSL 1.0.2 is going to go out of support some time. I don't think we should wait to upgrade.

JL2210:
We can patch PHP.

JL2210:
> Our previous LTS version (1.0.2 series) will continue to be supported until 31st December 2019 (security fixes only during the last year of support).

uberhacker:
I don't think you understand.  We cannot update this version of openssl without fixing other packages along the way.  Here is another example.  You cannot compile curl with this version:
```
In file included from http_ntlm.c:40:
curl_ntlm_core.h:41:12: fatal error: openssl/ssl.h: No such file or directory
 #  include <openssl/ssl.h>
            ^~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [Makefile:1875: libcurl_la-http_ntlm.lo] Error 1
In file included from curl_ntlm_wb.c:51:
curl_ntlm_core.h:41:12: fatal error: openssl/ssl.h: No such file or directory
 #  include <openssl/ssl.h>
            ^~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [Makefile:1882: libcurl_la-curl_ntlm_wb.lo] Error 1
make[2]: Leaving directory '/usr/local/tmp/crew/curl-7.61.0.tar.xz.dir/curl-7.61.0/lib'
make[1]: *** [Makefile:749: all] Error 2
make[1]: Leaving directory '/usr/local/tmp/crew/curl-7.61.0.tar.xz.dir/curl-7.61.0/lib'
make: *** [Makefile:927: all-recursive] Error 1
curl failed to build: `make V=0 -j2` exited with 2
```

JL2210:
This should work for 5.6:
https://bugs.php.net/patch-display.php?bug_id=73741&patch=php56-openssl11&revision=1493415093
I'm looking for a patch for 7.2.
Curl has already been patched:
https://curl.haxx.se/download/curl-7.64.0.tar.xz

JL2210:
We *can* fix other packages if we need to. That's the point.

JL2210:
You're not even giving this a chance. We can't just go using an old version of OpenSSL forever.

uberhacker:
You didn't even try what I asked you to do earlier.  Test other packages like curl and php.  You want to put that burden on me, apparently.  We can't just add packages if it breaks others, especially if it's an important core package like this one.

uberhacker:
We need to fix all those other packages in one go because if we just include this and not the others, things break and that means unhappy users.

JL2210:
I'm sorry...

uberhacker:
No need to apologize.  I appreciate your efforts.  We just need to be careful about these kinds of things.  It's like walking on egg shells.

PR #3165, opened 2019-02-13
Update code.rb

JL2210:


packages/code.rb
```diff
@@ -65,5 +65,8 @@ def self.postinstall
     puts 'Now, please run \'code-oss\' to start Code.'.lightgreen
     puts 'Happy coding!'.lightgreen
     puts
+    puts 'If you\'re having issues with your installation, comment out the Sommelier'.orange
+    puts ' configuration in `~/.bashrc` and try again.'
```
uberhacker: This is not a good suggestion because it will break other apps that run fine with sommelier.

uberhacker: Also, you could update to https://github.com/Microsoft/vscode/archive/1.31.1.tar.gz.

skiman6010: Yeah we should add in some language about how this will break other sommelier apps, but for now it is the only way we can get Code to run right now.

uberhacker: Does `stopsommelier` do the job?

JL2210: I'm tired of how the `sommelier`, `broadway`, and `xdg_base` packages work right now. I'm rewriting them.

skiman6010: @JL2210 Can you put some language in there explaining that this might break other sommelier apps and then I'll merge it?

uberhacker:
This solution is a bit too hackish.

JL2210:
@uberhacker I'm working on it. Please reopen.

JL2210:
And it's also the only solution that works.

skiman6010:
@JL2210 Please re-open as a draft PR if you are going to continue to work on it after submitting. https://github.blog/2019-02-14-introducing-draft-pull-requests/

JL2210:
OK.

PR #3170, opened 2019-02-15
Update Code to 1.31.1

skiman6010:
Updates Code to 1.31.1 which is the newest release.

Does not have any of @JL2210 changes in the open PR.


uberhacker:
Does this work on release 72?

skiman6010:
Yes and also works on 73 as well. 

You have to launch it via `sommelier -X /usr/local/share/code/code-oss` though as before. Which is lame, but works.

uberhacker:
I have an idea how we can make it less lame.

PR #3171, opened 2019-02-16
Update curl from 7.61.0 to 7.64.0

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/curl.rb
+++ b/packages/curl.rb
@@ -3,33 +3,45 @@ require 'package'
 class Curl < Package
   description 'Command line tool and library for transferring data with URLs.'
   homepage 'https://curl.haxx.se/'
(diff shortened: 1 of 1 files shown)


superloach:
Apparently this causes issues with libmetalink, at least on ARM32.

uberhacker:
@superloach:  I have noticed this also.  It seems to be problematic on packages that have curl as a dependency.  What package did you notice this issue?

superloach:
Someone on the Discord attempting to run the crew installer.

uberhacker:
Can you please provide more details (ie. architecture, install or reinstall, etc)?  If this relates to #3174, please comment on that thread.

superloach:
Never mind, #3174 is the same one.

PR #3177, opened 2019-02-17
Update libpng from 1.6.34 to 1.6.36

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/libpng.rb
+++ b/packages/libpng.rb
@@ -3,32 +3,37 @@ require 'package'
 class Libpng < Package
   description 'libpng is the official PNG reference library.'
   homepage 'http://libpng.org/pub/png/libpng.html'
-  version '1.6.34'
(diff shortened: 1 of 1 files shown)


packages/libpng.rb
```diff
@@ -3,32 +3,37 @@
...
+  end
+
   def self.build
-      system './configure',
-        "--prefix=#{CREW_PREFIX}",
-        "--libdir=#{CREW_LIB_PREFIX}",
-        '--disable-dependency-tracking',
-        '--disable-maintainer-mode'
-      system 'make'
+    system './configure',
```
skiman6010: You have mixed single and double quotes here, any specific reason for that?

uberhacker: That is a very good question.  The crew constants require the double quotes in order to be expanded but the other quotes line do not need the double quotes.

PR #3181, opened 2019-02-19
Update code.rb

JL2210:
Draft.

Any suggestions welcome, please do not close.


uberhacker:
This update didn't work when I tested it and as I explained earlier, is not a good solution since it will break other packages that work fine with sommelier.  We need a better solution.

JL2210:
I'm all ears.

PR #3182, opened 2019-02-19
Update sommelier, broadway, and xdg_base packages

JL2210:
This is a work in progress to eliminate the configuration stress with Sommelier, Broadway, and `xdg_base`. Please mark it as such.

Any suggestions are appreciated.

### DIFF
--- a/packages/broadway.rb
+++ b/packages/broadway.rb
@@ -13,65 +13,25 @@ class Broadway < Package
   })
 
   depends_on 'gtk3'
-
-  def self.build
-    system "echo 'GDK_BACKEND=broadway' > .broadway.env"
-    system "echo 'XDG_RUNTIME_DIR=/var/run/chrome' >> .broadway.env"
-    system "echo 'BROADWAY_DISPLAY=:5' >> .broadway.env"
-    system "echo '#!/bin/bash' > initbroadway"
-    system "echo 'BROADWAYD=\$(pidof broadwayd 2>/dev/null)' >> initbroadway"
-    system "echo 'if [ -z \"\${BROADWAYD}\" ]; then' >> initbroadway"
-    system "echo '  [ -f #{CREW_PREFIX}/bin/stopsommelier ] && stopsommelier' >> initbroadway"
-    system "echo '  broadwayd \${BROADWAY_DISPLAY} &' >> initbroadway"
-    system "echo '  sleep 3' >> initbroadway"
-    system "echo 'fi' >> initbroadway"
-    system "echo 'BROADWAYD=\$(pidof broadwayd 2>/dev/null)' >> initbroadway"
-    system "echo 'if [ ! -z \"\${BROADWAYD}\" ]; then' >> initbroadway"
-    system "echo '  echo \"broadwayd process \${BROADWAYD} is running\"' >> initbroadway"
-    system "echo 'else' >> initbroadway"
-    system "echo '  echo \"broadwayd failed to start\"' >> initbroadway"
-    system "echo '  exit 1' >> initbroadway"
-    system "echo 'fi' >> initbroadway"
-    system "echo '#!/bin/bash' > stopbroadway"
-    system "echo >> stopbroadway"
-    system "echo 'BROADWAYD=\$(pidof broadwayd 2>/dev/null)' >> stopbroadway"
-    system "echo 'if [ ! -z \"\${BROADWAYD}\" ]; then' >> stopbroadway"
-    system "echo '  pkill broadwayd' >> stopbroadway"
-    system "echo '  sleep 3' >> stopbroadway"
-    system "echo 'fi' >> stopbroadway"
-    system "echo 'BROADWAYD=\$(pidof broadwayd 2> /dev/null)' >> stopbroadway"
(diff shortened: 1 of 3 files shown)


packages/xdg_base.rb
```diff
@@ -3,41 +3,45 @@
...
     # Save any previous configuration, if it exists.
-    if Dir.exists? "$HOME/.config"
-      system "cp -a $HOME/.config #{CREW_PREFIX} && rm -rf $HOME/.config"
+    if Dir.exists? "#{HOME}/.config"
+      FileUtils.cp_r("#{HOME}/.config", "#{CREW_PREFIX}/")
+      FileUtils.rm_rf("#{HOME}/.config")
     end
-    if Dir.exists? "$HOME/.local"
-      system "mkdir -p #{CREW_PREFIX}/.config" unless Dir.exists? "#{CREW_PREFIX}/.config"
-      system "cp -a $HOME/.local/. #{CREW_PREFIX}/.config && rm -rf $HOME/.local"
```
uberhacker: There is a reason the `&&` is here.  If for any reason the `cp -a ...` fails, the `rm -rf ...` statement is ignored.  Otherwise, this could lead to potential lost data.

packages/broadway.rb
```diff
@@ -13,65 +13,25 @@ class Broadway < Package
...
-    system "install -Dm755 initbroadway #{CREW_DEST_PREFIX}/bin/initbroadway"
-    system "install -Dm755 stopbroadway #{CREW_DEST_PREFIX}/bin/stopbroadway"
-    system "install -Dm644 .broadway.env #{CREW_DEST_HOME}/.broadway.env"
-  end
-
+  
   def self.postinstall
+    puts 'To configure the Broadway daemon, remove the former'.lightblue
+    puts ' "Broadway environment variables + daemon" section'.lightblue
+    puts ' from `~/.bashrc`, and run this:'.lightblue
```
uberhacker: And how exactly does this make it any easier for the end user?  You just made the transition more difficult.  This is what the bash scripts you conveniently removed already do automagically.

packages/sommelier.rb
```diff
@@ -3,109 +3,60 @@
...
-      system "install -Dm644 .sommelier.env #{CREW_DEST_HOME}/.sommelier.env"
+    Dir.chdir ('sommelier') do
+      system 'make', "PREFIX=#{CREW_PREFIX}", "SYSCONFDIR=#{CREW_PREFIX}/etc", "DESTDIR=#{CREW_DEST_DIR}", 'install'
     end
   end
 
   def self.postinstall
+    puts 'To configure the Sommelier daemon, remove the former'.lightblue
+    puts ' "Sommelier environment variables + daemon" section'.lightblue
+    puts ' from `~/.bashrc`, and run this:'.lightblue
```
uberhacker: Dito here.  You made the transition more difficult.

packages/xdg_base.rb
JL2210: OK.

packages/broadway.rb
JL2210: Because the bash scripts screw stuff up. They're not part of the package, and they just make things harder.

packages/sommelier.rb
JL2210: See above. I wish there was a better solution to running X apps then this old version of Sommelier.

packages/broadway.rb
uberhacker: Can you explain exactly what you mean by screwing stuff up?  That is a generalization, not a specific issue that we can address.

JL2210: They set variables that cannot be unset, take ~15 seconds every shell startup, and display annoying debug messages.

packages/sommelier.rb
JL2210: I have an idea: Create a `virtwl` package with the required header and build from there.

packages/xdg_base.rb
JL2210: Would it work to use what I'm doing, but concatenate the lines with an `if` in between?

packages/sommelier.rb
JL2210: Argh. It doesn't work.

JL2210:
Any other suggestions?

uberhacker:
Leave it the way it is? :)

JL2210:
The way it is is just too.... What's the word? Inconsistent, maybe? One thing, for sure, is that there are too many scripts installed, and too many verbose debugging messages not useful for the average user outputted. Along with that, the startup lag makes the current behavior all too unappealing.

uberhacker:
I'm not hearing any other complaints from other users.  Can you explain what you mean about inconsistent?  I haven't noticed that issue myself.

JL2210:
You have to stop one daemon to run another. That's most likely because nobody uses Broadway except for me.

uberhacker:
I think we need a better solution.  I do agree the xdg_base package needs work so if you want to submit a PR for just that package, it would be nice.

JL2210:
Please reopen. It's a draft for a reason, and all your concerns have been addressed.

PR #3183, opened 2019-02-21
Add musl package

JL2210:
musl provides consistent quality and implementation behavior from tiny embedded systems to full-fledged servers. Minimal machine-specific code means less chance of breakage on minority architectures and better success with “write once run everywhere” C development.

### DIFF
--- a/packages/musl.rb
+++ b/packages/musl.rb
@@ -1,0 +1,44 @@
+require 'package'
+
+class Musl < Package
+  description 'A modern, simple, and fast C library implementation that strives to be lightweight, fast, simple, free, and correct in the sense of standards-conformance and safety.'
+  homepage 'https://www.musl-libc.org/'
+  version '1.1.21'
+  source_url 'https://www.musl-libc.org/releases/musl-1.1.21.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/musl.rb
```diff
@@ -0,0 +1,40 @@
...
+  source_url 'https://www.musl-libc.org/releases/musl-1.1.21.tar.gz'
+  source_sha256 'c742b66f6f49c9e5f52f64d8b79fecb5a0f6e0203fca176c70ca20f6be285f44'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.build
+    abi = 'eabihf' if ARCH == 'armv7l'
```
uberhacker: Probably should add `abi = ''` above this line so other architectures have this variable assigned.

JL2210: It works.

uberhacker: It works but it's not good practice.  You could potentially get unexpected results using a variable that is unassigned.

JL2210: OK.

JL2210:
@uberhacker Done.

PR #3184, opened 2019-02-21
Update glibc 2.23 2.27

JL2210:
@uberhacker: Thank you for writing this. It needs some work to merge, but it's a definite solution for the issues we're seeing on the stable channel now.


uberhacker:
I was waiting for the right time to introduce this PR.  Now is almost the time but not quite.  I wanted to wait until the release 72 update has settled in with most devices first.

JL2210:
It has. I got the update arrow over a week ago. It's time.

uberhacker:
It has on YOUR device but I have 2 chromebooks that have not received the update notice yet.  What I don't want is for folks to install this update before it's ready.  Otherwise, we have to tell them to reinstall.

JL2210:
It will be fine. If their Chromebook is still on glibc 2.23 it will install
glibc 2.23. When their device reaches 72 it will upgrade again.

packages/glibc.rb
```diff
@@ -3,29 +3,33 @@
 class Glibc < Package
   description 'The GNU C Library project provides the core libraries for GNU/Linux systems.'
   homepage 'https://www.gnu.org/software/libc/'
-  version '2.23'
-  source_url 'https://ftpmirror.gnu.org/glibc/glibc-2.23.tar.xz'
-  source_sha256 '94efeb00e4603c8546209cefb3e1a50a5315c86fa9b078b6fad758e187ce13e9'
+  version '2.27'
+  source_url 'https://ftpmirror.gnu.org/glibc/glibc-2.27.tar.xz'
```
JL2210: This really should be `https://ftpmirror.gnu.org/libc/glibc-2.27.tar.xz`. `glibc` is a symlink to `libc`.

uberhacker: What exactly should be changed here?

JL2210:
Oh. Can you change that?

PR #3190, opened 2019-02-23
Add android_studio package

uberhacker:
Android Studio is the official IDE for Android development, and includes everything you need to build Android apps.  See https://developer.android.com/studio.

Tested on:
- [x] x86_64

### DIFF
--- a/packages/android_studio.rb
+++ b/packages/android_studio.rb
@@ -1,0 +1,48 @@
+require 'package'
+
+class Android_studio < Package
+  description 'Android Studio is the official IDE for Android development.'
+  homepage 'https://developer.android.com/studio'
+  version '3.3.1.0'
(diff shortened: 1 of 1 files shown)


skiman6010:
Man how does this run on lower end Chromebooks?

uberhacker:
If you have a small hdd, you are out of luck.  You can say the same about arm Chromebooks.

packages/android_studio.rb
```diff
@@ -0,0 +1,48 @@
+require 'package'
+
+class Android_studio < Package
+  description 'Android Studio is the official IDE for Android development.'
+  homepage 'https://developer.android.com/studio'
+  version '3.3.1.0'
+  source_url 'https://dl.google.com/dl/android/studio/ide-zips/3.3.1.0/android-studio-ide-182.5264788-linux.zip'
```
JL2210: ... This is not a good idea. You should at least fail out before downloading on other architectures.

JL2210: http://tools.android.com/build/studio

PR #3192, opened 2019-02-25
Update rust from 1.23.0 to 1.32.0

uberhacker:
Tested on:
- [x] aarch64
- [x] armv7l
- [x] x86_64

Unable to build on i686.


skiman6010:
Oh yeah this should probably be updated.

Issue #3195, opened 2019-02-28
[golang update] go 1.12 is released


nosarthur:
https://blog.golang.org/go1.12

JL2210:
Thank you for letting us know. Not sure if go is our highest priority right
now, but I imagine somebody may try to update it.

nosarthur:
This is not urgent to me either. Maybe I will learn how to update it and hopefully create a PR

uberhacker:
@nosarthur: Apply the attached patch and you can get credit for this update.
```
$ cd
$ git clone git@github.com:skycocker/chromebrew.git
$ cd chromebrew/packages
$ patch -p1 go.rb < ~/Downloads/go-1.12.patch.txt
```
[go-1.12.patch.txt](https://github.com/skycocker/chromebrew/files/2915843/go-1.12.patch.txt)

JL2210:
He's not the one who wrote it, so why should he get credit?

skiman6010:
@JL2210 Because we are trying to get new people involved to create, update and maintain packages so the work isn't all on the few tireless workers. The best way to get people involved is by letting them dip their feet in the water and not just pushing them into the ocean. 

@uberhacker has set up a nice way for @nosarthur to clone and see how the repo is laid out while also providing enough help in that patch to not mess anything up and also to learn how updates are applied. They obviously have some interest in golang or they wouldn't have notified us of the update. In that regard, allowing them to explore how packages are updated without completely jumping in seems like the best way to get more people involved. 

If you have a problem with that, you can PM us in Slack so we don't clutter this issue with more bickering. 

Thanks!

JL2210:
You're right. Sorry.

PR #3198, opened 2019-02-28
Update go from 1.11.1 to 1.12 (#3195)

nosarthur:
Not sure if I should add these binary urls. Maybe they won't work anyways?


uberhacker:
Thank you for the PR @nosarthur!  Those binaries won't work anyway since it won't match the architecture.  We can generate compatible binaries for all Chromebooks.  If you remove those, I can merge this.

nosarthur:
I see, thanks

PR #3203, opened 2019-03-02
Sublime depends on gtk2

helge000:
This was mentioned in the FAQ, however, gtk2 is not installed.

### DIFF
--- a/packages/sublime_text.rb
+++ b/packages/sublime_text.rb
@@ -18,6 +18,7 @@ class Sublime_text < Package
   binary_sha256 ({
   })
 
+  depends_on 'gtk2'
   depends_on 'sommelier'
 
   def self.install


JL2210:
It's not an issue with GTK2, it's just misinformation in the FAQ. Follow
the instructions, just change instances of 2.0 to 3.0.

helge000:
@JL2210, to clarify: We do not need gtk2 for sublime_text to work? Or should we use rather depend on gtk3?

I accidentally wiped my chromebook two times in a row and both times sublime would only work after gtk2 was installed.

I only checked primary deps for [sommelier](https://github.com/skycocker/chromebrew/blob/master/packages/sommelier.rb#L23), it does not seem  to pull in any of the gtk packages.

### Ref
From [FAQ](https://github.com/skycocker/chromebrew/wiki/FAQ): 
> Q: I'm having troubles with font size in apps depending on GTK2. What do I do?
> A: Packages like gimp and sublime_text still depend on GTK2.  [...]

JL2210:
If that's the case, then this is OK. @uberhacker?

packages/sublime_text.rb
```diff
@@ -19,6 +19,7 @@ class Sublime_text < Package
   })
 
   depends_on 'sommelier'
+  depends_on 'gtk2'
```
uberhacker: Can you please move this above `depends_on 'sommelier'`?  The reason is there are post install instructions that would be missed.

helge000: Done

PR #3204, opened 2019-03-02
Add bash_completion support for git

helge000:
### For some reason git seems not be worthy for upstream bash_completion.

Though trivial it my first package submission :). 

### Issues
I went through `crew` and the Package class, however it seems there is no easy way around tar?

### DIFF
--- a/packages/bash_completion_git.rb
+++ b/packages/bash_completion_git.rb
@@ -1,0 +1,16 @@
+require 'package'
+
+class Bash_completion_git < Package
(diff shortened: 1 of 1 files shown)


JL2210:
There's already Bash completion in the `git` package. It's only shown when upgrading, not in `install.sh`, though. How about you add `crew postinstall git` to `install.sh`?

helge000:
@JL2210 - you're right, I did not catch that. Closing!

PR #3206, opened 2019-03-02
Fixes #3205: only expand real archives

### LINKED ISSUE #3205: crew should not try to unpack a non-archive
## Description
crew should not try to extract non-archives

```shell
#crew install mysqltuner
[...]
Archive downloaded
tar: This does not look like a tar archive
tar: Skipping to next header
```

helge000:
Uses regex on File.basename(). Using File.extname() on some
files (ie, archive.tar.gz) can lead to unexpected behavior.

### DIFF
--- a/crew
+++ b/crew
@@ -444,22 +444,17 @@ def unpack (meta)
   extract_dir = "#{meta[:filename]}.dir"
   target_dir = nil
   Dir.chdir CREW_BREW_DIR do
-    puts "Unpacking archive, this may take awhile..."
     Dir.mkdir("#{extract_dir}") unless Dir.exist?("#{extract_dir}")
-    case File.extname meta[:filename]
-    when '.zip'
-      if @opt_verbose then
-        system "unzip", "-v", "-d", "#{extract_dir}", meta[:filename]
-      else
-        system "unzip", "-qq", "-d", "#{extract_dir}", meta[:filename]
-      end
-    else
-      if @opt_verbose then
-        system "tar", "xvf", meta[:filename], "-C", "#{extract_dir}"
-      else
-        system "tar", "xf", meta[:filename], "-C", "#{extract_dir}"
-      end
-    end
+    case File.basename meta[:filename]
+    when /\.zip$/i
+      puts "Unpacking archive using 'unzip', this may take awhile..."
+      _verbopt = @opt_verbose ? 'v' : 'qq'
+      system "unzip", _verbopt, "-d", "#{extract_dir}", meta[:filename]
+    when /\.(tar(\.(gz|bz2|xz))?|tgz|tbz)$/i
+      puts "Unpacking archive using 'tar', this may take awhile..."
+      _verbopt = @opt_verbose ? 'v' : ''
+      system "tar", "x#{_verbopt}f", meta[:filename], "-C", "#{extract_dir}"
+   end
     if meta[:source] == true
       # Check the number of directories in the archive
       entries = Dir["#{extract_dir}/*"]
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.2.1'
+CREW_VERSION = '1.2.2'
 
 ARCH = `uname -m`.strip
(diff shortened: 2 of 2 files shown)


crew
```diff
@@ -444,21 +444,18 @@ def unpack (meta)
...
-    case File.extname meta[:filename]
-    when '.zip'
-      if @opt_verbose then
-        system "unzip", "-v", "-d", "#{extract_dir}", meta[:filename]
-      else
-        system "unzip", "-qq", "-d", "#{extract_dir}", meta[:filename]
-      end
+    case File.basename meta[:filename]
+    when /\.zip$/i
+      puts "Unpacking archive using zip, this may take awhile..."
```
uberhacker: Extracting archive using unzip?

crew
```diff
@@ -444,21 +444,18 @@ def unpack (meta)
...
+      _verbopt = @opt_verbose ? 'v' : 'qq'
+      system "unzip", _verbopt, "-d", "#{extract_dir}", meta[:filename]
+    when /\.(tar|tbz|tgz)/i
+      puts "Unpacking archive using tar, this may take awhile..."
+      _verbopt = @opt_verbose ? 'v' : ''
+      system "tar", "x#{_verbopt}f", meta[:filename], "-C", "#{extract_dir}"
     else
```
uberhacker: There are other archive types such as `.xz` and `.bz2` that should be considered also.  Actually, these extensions are getting to be more common since the compression is more efficient than traditional `.gz`.

crew
```diff
@@ -444,21 +444,18 @@ def unpack (meta)
...
-      if @opt_verbose then
-        system "tar", "xvf", meta[:filename], "-C", "#{extract_dir}"
-      else
-        system "tar", "xf", meta[:filename], "-C", "#{extract_dir}"
-      end
+      puts "Not unpacking #{File.basename meta[:filename]}"
```
uberhacker: No need to output anything here.  Keep in mind whenever you make changes to the `crew` command, you will also need to bump the version in `lib/const.rb`.

crew
helge000: All these extensions are already covered by the regex. Have a look, basically it does mach the glob `*.tar.*`

crew
helge000: Right, I will remove it: I was not sure whatever you guys require `else` clauses in switch statements. I'll bump the version by 0.01, i suppose?

crew
helge000: I basically kept the original calls to the `#system` method here; but if you want to change anything else while we're at it, let me know.

crew
helge000: As a side node: The way you are using tar here was new to me; I am oldschol and was using the `z|j|J` switches with `x` as a habit. It seems, nowadays you do not need to do that any more as tar detects the filer by itself. (I am quite certain I was getting an error some centuries ago feeding a tgz to plain tar xf... :)

Still, it needs do be a tarball to work. Using a blanket `tar xf archive.xz`  does not work on a xz archive, in this instance, which brings be right back to the intention of this PR.

As it replicates the current behaviour with arbitrary, non zip, non tar archives, I think we should focus on covering more archive types in future PR's?

uberhacker: Probably a better regex is `when /\.(bz2|gz|tar|tbz|tgz|xz)$/i`.

uberhacker: All of the pre-built binaries from Bintray are compressed using xz and they extract fine with the current configuration.

helge000: > Probably a better regex is `when /\.(bz2|gz|tar|tbz|tgz|xz)$/i`.

This still leaves us with the issue I am trying to solve: 
Would [`/\.(tar(\.(gz|bz2|xz)?)|tgz|tbz)$/i`](https://www.debuggex.com/r/gkK_Dzy1bTMirjkJ) work for you? This has the benefit that we can add more non-tar extractors later on (ie, xz or gz only)

The current implementation only works because Bintray uses tar with some filter added after the fact. As I stated above, try to use tar on a non-tar archive; it will do nothing.

uberhacker: Yes, that would be better.  I guess you can't assume if the compressed file ends in `.gz`, `.bz2` or `.xz` it is always going to be tarred also.

crew
helge000: Yea, I clearly did not get this at first :) - changed it with the latest version...

PR #3211, opened 2019-03-06
Add bacula package

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/bacula.rb
+++ b/packages/bacula.rb
@@ -1,0 +1,58 @@
+require 'package'
+
+class Bacula < Package
+  description 'Bacula is a set of computer programs that permits the system administrator to manage backup, recovery, and verification of computer data across a network of computers of different kinds.'
+  homepage 'https://www.bacula.org/'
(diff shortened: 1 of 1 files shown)


JL2210:
That `--datarootdir` flag should be `#{CREW_PREFIX}/share`, not `#{CREW_PREFIX}`.

skiman6010:
I agree with how it is done.

JL2210:
Umm... 
```
  --datarootdir=DIR       read-only arch.-independent data root [PREFIX/share]
  --datadir=DIR           read-only architecture-independent data [DATAROOTDIR]
```
This would make it so you wouldn't have to pass `--mandir=#{CREW_PREFIX}/share/man`.

uberhacker:
You are right.  If you want to create another PR to fix it, go ahead but everything is installed in the correct location.  I will fix it on the next version, otherwise.

JL2210:
Nah, it's not that big of a problem. Thanks.

PR #3212, opened 2019-03-08
Add ngrok package

uberhacker:
ngrok exposes local servers behind NATs and firewalls to the public internet over secure tunnels.

How it works
You download and run a program on your machine and provide it the port of a network service, usually a web server.

It connects to the ngrok cloud service which accepts traffic on a public address and relays that traffic through to the ngrok process running on your machine and then on to the local address you specified.

What it's good for
- Demoing web sites without deploying
- Building webhook consumers on your dev machine
- Testing mobile apps connected to your locally running backend
- Stable addresses for your connected devices that are deployed in the field
- Running personal cloud services from your home

Tested on all architectures.

### DIFF
--- a/packages/ngrok.rb
+++ b/packages/ngrok.rb
@@ -1,0 +1,30 @@
+require 'package'
+
+class Ngrok < Package
+  description 'ngrok exposes local servers behind NATs and firewalls to the public internet over secure tunnels.'
+  homepage 'https://ngrok.com/'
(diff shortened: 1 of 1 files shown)


JL2210:
Do you mind upgrading `binutils` to `2.32` and `gcc8` to `8.3.0`? Also, please enable the Objective-C an Objective-C++ languages (Requires `bdwgc` package).

Issue #3213, opened 2019-03-08
Duplicate packages


JL2210:
Duplicate packages: `bdwgc` and `gc`.

Also, please update to `8.0.4`: https://github.com/ivmai/bdwgc/releases/download/v8.0.4/gc-8.0.4.tar.gz

uberhacker:
You are welcome to submit a PR to update and remove the duplicate package.

uberhacker:
Just make sure you also update all the other packages that currently depend on the package you plan on removing and replace the dependency with the other package.  That means testing each package afterwards that is affected also.

JL2210:
There's no need to update if they're both the same package.

uberhacker:
You sound so confident.  Always need to test.

JL2210:
It's a little thing called backwards compatibility that every sane project has.

uberhacker:
Fixed in #3222.

JL2210:
There's a reason I didn't add "Fixes" to the description of that PR: There
may be more duplicate packages in the future. Please reopen.

uberhacker:
I think this is the only duplicate package.

JL2210:
That I've found so far, yes. But there may be more in the future.

uberhacker:
If you find one, let me know and I'll reopen.

PR #3222, opened 2019-03-24
Remove duplicate 'gc' package and update related packages

JL2210:
#3213


uberhacker:
Thank you for the update @JL2210!

PR #3223, opened 2019-03-24
Update neofetch package

JL2210:


uberhacker:
Thank you for the update @JL2210!

PR #3225, opened 2019-03-25
Add pre-built binaries for bdwgc, libatomic_ops, neofetch and tinycc packages

uberhacker:
Tested on all architectures.


JL2210:
Neofetch doesn't need binaries. It's just a script, and it installs quickly.

JL2210:
@uberhacker? I think the very least we could do in this situation is create an "all" architecture that we could use all the same binaries on.

uberhacker:
If a package doesn't need binaries, you should submit a change to [needs_binaries.sh](https://github.com/skycocker/chromebrew/blob/master/tools/needs_binaries.sh) to let the maintainers know to ignore.

PR #3228, opened 2019-03-25
Update tools/needs_binaries.sh

JL2210:
Add exclusions for more packages that
1. Are either platform-independent, and/or
2. Compile and install very quickly

Also change a bit of logic and bracket variables.


uberhacker:
This only includes perl stuff.

JL2210:
I added `neofetch` too. All the perl packages take < 1 min. to install.

PR #3231, opened 2019-03-26
Revert "Fixes #3205: only expand real archives"

### LINKED ISSUE #3205: crew should not try to unpack a non-archive

JL2210:
This has been causing issues ever since it was merged,
including but not limited to: screwing up `.zip` files,
not unpacking `.tar.lz` files, and other bugs.

### DIFF
--- a/crew
+++ b/crew
@@ -441,20 +441,25 @@ def download
 end
 
 def unpack (meta)
-  extract_dir = "#{meta[:filename]}.dir"
+  extract_dir = meta[:filename]+ '.dir'
   target_dir = nil
   Dir.chdir CREW_BREW_DIR do
-    Dir.mkdir("#{extract_dir}") unless Dir.exist?("#{extract_dir}")
-    case File.basename meta[:filename]
-    when /\.zip$/i
-      puts "Unpacking archive using 'unzip', this may take a while..."
-      _verbopt = @opt_verbose ? 'v' : 'qq'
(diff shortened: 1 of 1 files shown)


JL2210:
Argh... Somebody changed it after the original PR. Hang on...

JL2210:
@uberhacker Please mark with [high priority] and/or merge.

uberhacker:
I don't want to revert @helge000's work.  He provided a good suggestion of how we can improve this process.  Rather than reverting, all we need to do is include the `.lz` extension in the regex pattern for the gmp package which is the only package that breaks with what we have currently.

JL2210:
This utterly breaks unpacking zip files.

A better solution would simply be silencing any warnings from the `tar` command (`2>/dev/null`).

uberhacker:
I haven't noticed it breaking zip packages.  Which specific package are you having trouble installing now?  You know already how I feel about silencing warnings.  Not a better solution.

JL2210:
I'd probably say these packages have been broken since merging that PR: "pmd kotlin xz_java dart markdown bcif s oci packer ngrok jam docbook_xml freedos mywanip sejda_console glfw elixir gwt glog glm android_studio gradle binutils ed gmp wget lzip"

PR #3232, opened 2019-03-26
Expand lzip and zip files correctly

JL2210:
Fix #3231

### DIFF
--- a/crew
+++ b/crew
@@ -449,8 +449,8 @@ def unpack (meta)
     when /\.zip$/i
       puts "Unpacking archive using 'unzip', this may take a while..."
       _verbopt = @opt_verbose ? 'v' : 'qq'
-      system "unzip", _verbopt, "-d", "#{extract_dir}", meta[:filename]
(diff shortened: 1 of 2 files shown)


crew
```diff
@@ -449,8 +449,8 @@ def unpack (meta)
     when /\.zip$/i
       puts "Unpacking archive using 'unzip', this may take a while..."
       _verbopt = @opt_verbose ? 'v' : 'qq'
-      system "unzip", _verbopt, "-d", "#{extract_dir}", meta[:filename]
-    when /\.(tar(\.(gz|bz2|xz))?|tgz|tbz)$/i
+      system "unzip", '-' + _verbopt, "-d", "#{extract_dir}", meta[:filename]
```
uberhacker: Why not just include the dash in `_verbopt`?

uberhacker: We also need a version bump in `lib/const.rb`.

JL2210: The dash in `_veropt` would be slightly redundant. Version bumped.

JL2210: That seems fair. Give me a minute...

PR #3235, opened 2019-03-30
Fix issue where gcc8 depends on gcc7 and vice versa

JL2210:
This issue causes any and all packages installed with `-S`
to depend on a version of GCC that's not installed.
It also breaks the dual-versioning support scheme.


uberhacker:
I held off on this since I had planned on upgrading the gcc packages.  It wouldn't hurt to merge this to ease the "pain" of those crazy enough to compile with the `-S` flag.

PR #3247, opened 2019-04-16
Add perl_algorithm_diff, perl_digest_md5, perl_parallel_forkmanager, perl_regexp_common packages

JL2210:
Filling in some Perl packages in case they're needed in the future.


uberhacker:
I would rather wait until they're actually needed.  Why not just include the cloc package also so these packages have a purpose?

JL2210:
Sure.

PR #3248, opened 2019-04-18
Update llvm from 7.0.0-1 to 8.0.0

uberhacker:
Tested on all architectures.


JL2210:
You can get the current GCC version inside the script with `\$(gcc -dumpversion)`. That way this update won't be invalidated when GCC is updated to `8.3.0`.

JL2210:
You can also get the architecture with `\$(gcc -dumpmachine)`.

JL2210:
Nice!

PR #3252, opened 2019-04-22
Update ruby from 2.5.3 to 2.6.3

uberhacker:
Tested on all architectures.  This update provides a self-healing wrapper script around ruby that makes sure crew still works if any of the runtime dependency packages are removed.  In fact, any of the core packages can be removed and crew will continue to work.  You can't remove ruby without requiring a fresh Chromebrew install, however.  A future PR is in the plans to make it possible to remove the ruby package and still keep the crew command working.


JL2210:
Whoa! I do not approve. Just disable stack smashing protection.

uberhacker:
Dude, that doesn't work.  I disapprove your suggestion.

JL2210:
Configure with stack_protector=no and tell me it doesn't work again.

JL2210:
If that doesn't work, try *at least* packaging `libssp` with the `ruby` package.

JL2210:
But I really don't want to force a download, nor the overhead of a shell script, on unsuspecting users.

uberhacker:
You are missing the point.  The only reason you want to remove stack protection is so you can get rid of the gcc dependency.  That is not a good enough reason to take an important feature away.  The overhead of the shell script is minimal.  This is a package manager, not some executable that needs maximum performance.  There is more method to the madness here.  I want to be able to switch gcc versions without breaking crew and this solution fits the bill.

uberhacker:
I have an idea how to package libssp.  I think that is a better approach to be able to switch gcc versions.

PR #3256, opened 2019-04-26
Add bleachbit package

uberhacker:
Tested on all architectures.  The gui version doesn't work on i686 but that is somewhat expected.


JL2210:
This looks interesting. Does it work? How well does it work?

PR #3259, opened 2019-04-28
Recompile lua to remove readline7 dependency

uberhacker:
Tested on all architectures.


JL2210:
While you're at it, could you recompile gdb too?

PR #3265, opened 2019-05-09
Rearrange crew dependency install order

### LINKED ISSUE #3174: curl Missing libmetalink.so.3 after downloading
Happened right after attempting to install clib;

```chronos@localhost ~/Downloads/bluefish/bluefish-2.2.10 $ crew install clib 
clib: C package manager-ish https://github.com/clibs/clib version 1.8.1 
The following packages also need to be installed: curl brotli libmetalink openldap rtmpdump Do you agree? [Y/n] y 
Proceeding... 
curl: Command line tool and library for transferring data with URLs. https://curl.haxx.se/ version 7.64.0 Precompiled binary available, downloading... ################################################################################################################################################ 100.0% 
Archive downloaded Unpacking archive, this may take awhile... 
Installing... 
Performing post-install... 
brotli: Brotli compression format https://github.com/google/brotli version 1.0.4 
curl: error while loading shared libraries: libmetalink.so.3: cannot open shared object file: No such file or directory 
12: from /usr/local/bin/crew:978:in `<main>' 
11: from /usr/local/bin/crew:909:in `install_command' 
10: from /usr/local/bin/crew:909:in `each' 
9: from /usr/local/bin/crew:913:in `block in install_command' 
8: from /usr/local/bin/crew:612:in `resolve_dependencies_and_install' 
7: from /usr/local/bin/crew:687:in `resolve_dependencies'
6: from /usr/local/bin/crew:687:in `each' 
5: from /usr/local/bin/crew:689:in `block in resolve_dependencies' 
4: from /usr/local/bin/crew:701:in `install' 
3: from /usr/local/bin/crew:429:in `download' 
2: from /usr/local/bin/crew:429:in `chdir'
1: from /usr/local/bin/crew:436:in `block in download' 
/usr/local/bin/crew:436:in `read': No such file or directory @ rb_sysopen - ./brotli-1.0.4-chromeos-armv7l.tar.xz (Errno::ENOENT)```

edit: formatting

### PR DESCRIPTION
**Pull request by uberhacker**
- Fixes issue when curl is installed before brotli or libmetalink
- Fixes issue when gtk3 is installed before shared_mime_info
- Fixes issue when sommelier is installed before other dependencies and the postinstall message is missed
- Bump crew version

Fixes #3174.

### DIFF
--- a/crew
+++ b/crew
@@ -659,7 +659,20 @@ def resolve_dependencies
 
   return if @dependencies.empty?
 
+
   puts "The following packages also need to be installed: "
+
+  i = 0
+  last_deps = []
+  last_packages = ["curl", "gtk3", "sommelier"]
+  @dependencies.each do |dep|
+    if last_packages.include?(dep)
+      @dependencies.delete_at(i)
+      last_deps.push(dep)
+    end
+    i += 1
+  end
+  @dependencies.concat last_deps.sort
 
   @dependencies.each do |dep|
     print dep + " "
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.2.3'
+CREW_VERSION = '1.2.4'
 
 ARCH = `uname -m`.strip
 ARCH_LIB = if ARCH == 'x86_64' then 'lib64' else 'lib' end

### CODE REVIEW

File: crew [hunk 1]
```diff
@@ -659,8 +659,21 @@ def resolve_dependencies
 
   return if @dependencies.empty?
 
+
   puts "The following packages also need to be installed: "
 
+  i = 0
+  last_deps = []
+  last_packages = ["curl", "gtk3", "sommelier"]
```
JL2210: I think we should find a better solution than hardcoding package names. I'm sure there are other packages affected by this that need a fix, too.

JL2210:
I have a better option: Use `unshift` instead of `concat` in the dependency resolver.

crew
uberhacker: Agreed this is not the best solution but it solves 3 known problems without a performance penalty with the dependency resolver routine.  If you can identify other packages, we can certainly add them here.  These are the issues I noticed so far.

PR #3266, opened 2019-05-09
Fix dependency order

JL2210:
Also fix a possibly unseen typo in available commands.

Please test and see if this works. As far as I can tell, it does.

### DIFF
--- a/crew
+++ b/crew
@@ -240,7 +240,7 @@ def help (pkgName)
     puts "Usage: crew whatprovides <pattern> ..."
     puts "The <pattern> is a search string which can contain regular expressions."
   else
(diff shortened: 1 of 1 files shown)


uberhacker:
Although I do agree this algorithm is more elegant, there is no guarantee the sommelier dependency will be last and that is the thing I'm really trying to ensure.  Users are confused if things don't work when they miss the postinstall message.

JL2210:
Can you test it and check that sommelier is installed last?

uberhacker:
We really need to figure out a better algorithm to resolve the dependencies without a performance penalty.  With that said, changing from concat to unshift still does not solve the problem.  This is what I get with a fresh `crew install gimp`:

PR #3267, opened 2019-05-11
Update android_studio from 3.3.1.0 to 3.4.0.18

uberhacker:
Tested on all architectures.


JL2210:
Are you sure that this takes 6GB of space? Maybe you should double check that.

uberhacker:
It needs that much during installation.  Maybe more with a fresh install.

PR #3272, opened 2019-05-13
Add nconvert package

uberhacker:
NConvert is a powerful command line multi-platform batch image processor with more than 80 commands. Compatible with 500 image formats.  See https://www.xnview.com/en/nconvert/.

Tested on x86_64 and i686.

### DIFF
--- a/packages/nconvert.rb
+++ b/packages/nconvert.rb
@@ -1,0 +1,19 @@
+require 'package'
+
+class Nconvert < Package
+  description 'Command Line Batch utility for images'
+  homepage 'https://www.xnview.com/en/nconvert/'
+  version '6.88'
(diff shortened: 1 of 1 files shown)


JL2210:
Is this package open source? What's the license?

uberhacker:
NConvert is provided as FREEWARE (NO Adware, NO Spyware) for private or educational use (including non-profit organizations).

JL2210:
How can you be sure? Without the source code, you never know if the
software's spying on you. The owner also seems adamantly against releasing
any kind of source code, including his modified version of QT. Along with
that issue, the wrapper script causes the program not to run correctly, as
it overrides LD_LIBRARY_PATH.

PR #3278, opened 2019-05-17
Add profanity package

Sebatyne:
Profanity is a a console based XMPP client, under GPLv3 Licence : https://profanity-im.github.io/

I could test the package under a x86_64 system.


uberhacker:
Thank you for the new package @Sebatyne!  Nice work.

PR #3279, opened 2019-05-17
Updated Code to 1.34.0

skiman6010:
Works for me... But so did the last updates. Your mileage may vary.


uberhacker:
This still doesn't work for me but since it works for some, I'll go ahead and merge it.

PR #3285, opened 2019-05-20
Update weston.rb

### LINKED ISSUE #3284: Minor Issues with Weston Package
After installing, `weston` has issues with permissions on `XDG_RUNTIME_DIR`. On my system they're 755, instead of 700.
```
[14:25:29.238] warning: XDG_RUNTIME_DIR "/var/run/chrome" is not configured
correctly.  Unix access mode must be 0700 (current mode is 755),
and must be owned by the user (current owner is UID 1000).
```
It is also missing the dependency `libwebp`.
```
[14:26:09.660] Loading module '/usr/local/lib64/libweston-4/wayland-backend.so'
[14:26:09.662] Failed to load module: libwebp.so.7: cannot open shared object file: No such file or directory
[14:26:09.662] fatal: failed to create compositor backend
```
In addition, there's a warning at the end of building that states `remember to run 'libtool --finish /usr/local/lib64/weston'`. That should probably be added as well.

-----
I'll probably do a PR for this myself, just wanted to open an issue to track it in case I do something wrong 😛

superloach:
Fixes #3284

We should look into updating Weston, 4.0.0 has been old for a while!

### DIFF
--- a/packages/weston.rb
+++ b/packages/weston.rb
@@ -3,26 +3,17 @@ require 'package'
 class Weston < Package
   description 'Weston is the reference implementation of a Wayland compositor, and a useful compositor in its own right.'
   homepage 'http://wayland.freedesktop.org'
-  version '4.0.0'
+  version '4.0.0-1'
   source_url 'https://github.com/wayland-project/weston/archive/4.0.0.tar.gz'
   source_sha256 '46b0178cd37b0bf7471c9af12c847c7b8728699ecd5c04ce79be77ea12c98512'
-
-  binary_url ({
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/weston-4.0.0-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/weston-4.0.0-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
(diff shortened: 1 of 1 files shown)


packages/weston.rb
```diff
@@ -43,5 +34,9 @@ def self.build
   def self.install
     system "make DESTDIR=#{CREW_DEST_DIR} install"
   end
-end
 
+  def self.postinstall
+    system "libtool --finish #{CREW_LIB_PREFIX}/weston"
```
JL2210: This bit is unnecessary. It's handled automatically by `LD_LIBRARY_PATH` logic.

packages/weston.rb
```diff
@@ -43,5 +34,9 @@ def self.build
...
+    system "chmod 700 $XDG_RUNTIME_DIR -v"
```
JL2210: This could be considered insecure... Possibly just silence the warning? Also, user `1000` is chronos, even if the warning says it's not.

uberhacker:
Thank you for the update @superloach!  Did you want to go ahead and update weston to the newer version?

packages/weston.rb
JL2210: @superloach?

uberhacker:
@superloach?

uberhacker:
Closing due to inactivity.  There are newer releases available at https://github.com/wayland-project/weston/releases anyway and we should update to one of those instead.

PR #3286, opened 2019-05-21
Add ledger package.

lloydhuang:
Hello Admin,
I added a ledger package - A double-entry accounting system. It is great tools for emacs ledger-mode. This is my first time to package,  If any problem, pls feel free to let me know, thank you.

BRs
Lloyd

### DIFF
--- a/packages/ledger.rb
+++ b/packages/ledger.rb
@@ -20,16 +20,14 @@ class Ledger < Package
      x86_64: '',
   })
 
-  depends_on 'gmp' => :build
   depends_on 'boost' => :build
-  depends_on 'mpfr' => :build
 
   def self.build
-    system "./acprep"
+    system './acprep', "--prefix=#{CREW_PREFIX}"
     system 'make'
   end
 
   def self.install
-    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install/strip'
   end
 end
--- a/packages/ledger.rb
+++ b/packages/ledger.rb
@@ -1,0 +1,35 @@
+require 'package'
+
+class Ledger < Package
+  description 'A double-entry accounting system with a command-line reporting interface'
+  homepage 'https://www.ledger-cli.org/'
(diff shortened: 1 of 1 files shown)


packages/ledger.rb
```diff
@@ -0,0 +1,35 @@
...
+     x86_64: '',
+  })
+
+  depends_on 'gmp' => :build
+  depends_on 'boost' => :build
+  depends_on 'mpfr' => :build
+
+  def self.build
+    system "./acprep"
+    system 'make'
```
uberhacker: We will probably want to add `prefix=#{CREW_PREFIX}` and possibly `libdir=#{CREW_LIB_PREFIX}` to abstract the install location.

packages/ledger.rb
```diff
@@ -0,0 +1,35 @@
...
+  binary_sha256 ({
+    aarch64: '',
+     armv7l: '',
+       i686: '',
+     x86_64: '',
+  })
+
+  depends_on 'gmp' => :build
+  depends_on 'boost' => :build
+  depends_on 'mpfr' => :build
```
uberhacker: All of these dependencies come along for the ride in Chromebrew "core", so they aren't needed here.

uberhacker:
Thank you for the new package @lloydhuang!  Please see comments in the `Files changed` section.

packages/ledger.rb
uberhacker: Actually, `boost` is not in "core" so that line is fine.

packages/ledger.rb
lloydhuang: Option --prefix=#{CREW_PREFIX} is OK. But this package don't provide --libdir option. Is okay that just add --prefix option?

uberhacker: If no `--libdir` configure option is available, that usually indicates there are no library files included in this app.  It is fine to use only `--prefix`.

JL2210:
@ubethacker If this requires Boost as a runtime library, then that should be a regular dependency, not a build-time one.

lloydhuang:
@uberhacker Please help to review it, Thank you.

PR #3289, opened 2019-05-25
Update gptfdisk package

JL2210:
Update to latest version, remove packages that are included in core, remove unneeded hack with newer version.

Use more ruby during install phase and use `FileUtils.glob` for filename-matching.


uberhacker:
What happened to fixparts?

uberhacker:
If install works like I think it does, you shouldn't need `FileUtils.mkdir_p ...`.

JL2210:
`fixparts` is in there. `install` does *not* work like you think it does.

uberhacker:
Where is the manpage?

JL2210:
It's included in the glob `*.8.gz`.

PR #3293, opened 2019-05-29
Add sbcl package. A ANSI Common Lisp compiler.

lloydhuang:
SBCL is a high performance Common Lisp compiler. Please check this url: http://www.sbcl.org/

### DIFF
--- a/packages/sbcl.rb
+++ b/packages/sbcl.rb
@@ -1,0 +1,32 @@
+require 'package'
+
+class Sbcl < Package
+  description 'Steel Bank Common Lisp (SBCL) is a high performance Common Lisp compiler.'
+  homepage 'http://www.sbcl.org/index.html'
+  version '1.5.3'
+  source_url 'http://prdownloads.sourceforge.net/sbcl/sbcl-1.5.3-source.tar.bz2?download'
+  source_sha256 'd2a7901144d227a7194661dc3ccee28873c8d5cbbd167013660c78b9ad63640c'
+
+  binary_url ({
+    aarch64: '',
+     armv7l: '',
+       i686: '',
+     x86_64: '',
+  })
+  binary_sha256 ({
+    aarch64: '',
+     armv7l: '',
+       i686: '',
+     x86_64: '',
+  })
+
+  depends_on 'clisp' => :build
+
+  def self.build
+    system './make.sh', "--prefix=#{CREW_PREFIX}", '--xc-host=clisp --silent'
+  end
+
+  def self.install
+    system "INSTALL_ROOT=#{CREW_DEST_DIR}#{CREW_PREFIX} sh install.sh"
+  end
+end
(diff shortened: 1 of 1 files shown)


packages/sbcl.rb
```diff
@@ -0,0 +1,32 @@
+require 'package'
+
+class Sbcl < Package
+  description 'Steel Bank Common Lisp (SBCL) is a high performance Common Lisp compiler.'
+  homepage 'http://www.sbcl.org/index.html'
+  version '1.5.3'
+  source_url 'http://prdownloads.sourceforge.net/sbcl/sbcl-1.5.3-source.tar.bz2?download'
```
JL2210: Please remove `?download` here.

packages/sbcl.rb
```diff
@@ -0,0 +1,32 @@
...
+  source_sha256 'd2a7901144d227a7194661dc3ccee28873c8d5cbbd167013660c78b9ad63640c'
+
+  binary_url ({
```
JL2210: Please remove the empty `binary_url` and `binary_sha256` blocks.

JL2210:
Please see the above comments. Looks good other than that.

packages/sbcl.rb
```diff
@@ -0,0 +1,32 @@
...
+  })
+
+  depends_on 'clisp' => :build
+
+  def self.build
+    system './make.sh', "--prefix=#{CREW_PREFIX}", '--xc-host=clisp --silent'
+  end
+
+  def self.install
+    system "INSTALL_ROOT=#{CREW_DEST_DIR}#{CREW_PREFIX} sh install.sh"
```
JL2210: Please use `#{CREW_DEST_PREFIX}`.

packages/sbcl.rb
uberhacker: @lloydhuang: It's fine to leave the empty blocks.

uberhacker:
Thank you for the new package @lloydhuang!  We need a lisp compiler that actually works.

packages/sbcl.rb
```diff
@@ -0,0 +1,19 @@
...
+  description 'Steel Bank Common Lisp (SBCL) is a high performance Common Lisp compiler.'
+  homepage 'http://www.sbcl.org/index.html'
+  version '1.5.3'
+  source_url 'http://prdownloads.sourceforge.net/sbcl/sbcl-1.5.3-source.tar.bz2'
+  source_sha256 'd2a7901144d227a7194661dc3ccee28873c8d5cbbd167013660c78b9ad63640c'
+
+  depends_on 'clisp' => :build
+
+  def self.build
+    system './make.sh', "--prefix=#{CREW_PREFIX}", '--xc-host=clisp --silent'
```
uberhacker: Each switch option can be separated by commas like:
`... '--xc-host=clisp', '--silent'`
But this is not needed.

JL2210: @uberhacker I think he wanted the `xc-host` command to be `clisp --silent`.

packages/sbcl.rb
JL2210: @uberhacker It will cause the download to try and download an empty file and verify it with an empty sha256sum, instead of downloading the sources and installing as usual.

packages/sbcl.rb
uberhacker: If that is the case, then that should be enclosed in quotes.  The way it works now is the switches are processed separately.

PR #3311, opened 2019-06-09
Add sbc package

JL2210:
SBC is a digital audio encoder and decoder used to transfer data to Bluetooth audio output devices like headphones or loudspeakers.

Dependency for CRAS (Chromium OS Audio Server).


uberhacker:
Thank you for the new package @JL2210!  Technically, the homepage would be bluez.org but since there is no good information on that site about sbc, this will have to do.

PR #3312, opened 2019-06-09
Add iniparser package

JL2210:
iniparser is a free stand-alone ini file parsing library.
It is written in portable ANSI C and should compile anywhere.
iniparser is distributed under an MIT license.

Another dependency for CRAS.


uberhacker:
LGTM

PR #3313, opened 2019-06-09
Add ladspa package

JL2210:
LADSPA is a standard that allows software audio processors and effects to be plugged into a wide range of audio synthesis and recording packages.

For instance, it allows a developer to write a reverb program and bundle it into a LADSPA "plugin library." Ordinary users can then use this reverb within any LADSPA-friendly audio application. Most major audio applications on Linux support LADSPA.

Yet another dependency for CRAS.

### DIFF
--- a/packages/ladspa.rb
+++ b/packages/ladspa.rb
@@ -1,0 +1,27 @@
+require 'package'
+
+class Ladspa < Package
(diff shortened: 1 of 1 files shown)


uberhacker:
LGTM

PR #3314, opened 2019-06-09
Add cras package

JL2210:
CRAS: Chromium OS Audio Server

Allows ALSA to play audio through Chrome OS' default audio server.

Tested and working on:
- [x] x86-64

### DIFF
--- a/packages/cras.rb
+++ b/packages/cras.rb
@@ -1,0 +1,64 @@
+require 'package'
+
+class Cras < Package
+  description 'ChromeOS Audio Server'
+  homepage 'https://www.chromium.org/chromium-os/chromiumos-design-docs/cras-chromeos-audio-server'
+  version 'stabilize-12249.B'
+  source_url 'https://chromium.googlesource.com/chromiumos/third_party/adhd/+/refs/heads/stabilize-12249.B/cras/README?format=TEXT'
+  source_sha256 'b65b959f4ad842a219a637413b8702676378a0b5d18b5e413c78c211b3fb133c'
+
+  depends_on 'alsa_lib'
+  depends_on 'ladspa'
+  depends_on 'iniparser'
+  depends_on 'speexdsp'
+  depends_on 'sbc'
+  depends_on 'dbus'
+
+  def self.build
+    system 'git', 'clone', 'https://chromium.googlesource.com/chromiumos/third_party/adhd', '-b', version, '.'
+    Dir.chdir('cras') do
+      system 'cat << _EOF_ > 10-cras.conf
+pcm.cras {
+    type cras
+    hint {
+        show on
+        description "Chromium OS Audio Server"
+    }
+}
+ctl.cras {
+    type cras
+}
+# Default: route all audio through the CRAS plugin.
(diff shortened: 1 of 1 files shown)


packages/cras.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  depends_on 'alsa_lib'
+  depends_on 'ladspa'
+  depends_on 'iniparser'
+  depends_on 'speexdsp'
+  depends_on 'sbc'
+  depends_on 'dbus'
+
+  def self.build
+    system 'git', 'clone', 'https://chromium.googlesource.com/chromiumos/third_party/adhd', '-b', version, '.'
```
uberhacker: This should be cloned to a separate directory instead of the current directory since the README file from the source_url is present.

packages/cras.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+  end
+
+  def self.install
+    Dir.chdir('cras') do
+      system 'make', "DESTDIR=#{CREW_DEST_DIR}", "install"
+      FileUtils.mkdir_p CREW_DEST_PREFIX + '/share/alsa/alsa.conf.d/'
+      FileUtils.install '10-cras.conf', CREW_DEST_PREFIX + '/share/alsa/alsa.conf.d/'
+    end
+  end
```
uberhacker: Missing `def self.postinstall` message to explain how to start and configure the server?

packages/cras.rb
JL2210: `crew` creates a directory called `README.dir` and `chdir`s to that before any of this takes place.

packages/cras.rb
JL2210: No need. The server built in to Chrome OS is already running. All you need to do is run an application that uses ALSA.

packages/cras.rb
uberhacker: Is the directory empty?  If not, the git clone will fail.

packages/cras.rb
uberhacker: Apparently, you don't have much experience running daemons.  Any config change will require a server restart to pick up the change.

packages/cras.rb
JL2210: The directory is empty.

packages/cras.rb
JL2210: No, it won't. The reading of configuration is a runtime setting.

uberhacker: Where is this documented?  I have a difficult time believing that is possible.

JL2210: Under the "How" section of the homepage:

> Put an audio server in between Chromium and ALSA that can change how audio flows *on the fly*.

Emphasis added.

JL2210: Also, this is ALSA configuration, not CRAS configuration. ALSA has no server, nor does it have a daemon, so a server reset wouldn't even be possible.

uberhacker: Well, technically it is a cras config since you are passing alsa settings to the server.  But this is very interesting.  I'll try to see if this actually works.

PR #3341, opened 2019-06-16
Update stress-ng from 0.09.02 to 0.09.59

cstrouse:


packages/stressng.rb
```diff
@@ -33,6 +21,7 @@ def self.install
       "BINDIR=#{CREW_PREFIX}/bin",
       "MANDIR=#{CREW_PREFIX}/share/man/man1",
       "JOBDIR=#{CREW_PREFIX}/share/stress-ng/example-jobs",
+      "BASHDIR=#{CREW_PREFIX}/share/bash-completion/completions",
       'install'
   end
```
uberhacker: Maybe add a `def self.postinstall` message to explain how to set up bash completions?

PR #3346, opened 2019-06-17
Add aqemu package

uberhacker:
AQEMU is a GUI for virtual machines using QEMU as the backend. Support for the KVM accelerator on Linux is provided. The application has a user-friendly interface and allows to set a large number of options.  See https://sourceforge.net/projects/aqemu/.  Also includes the libvncserver dependency package.

Features:
- Manage VMs
- Create HDD images
- Convert HDD images
- Run and control VMs

Tested on:
- [x] aarch64
- [x] armv7l
- [x] x86_64

### DIFF
--- a/packages/aqemu.rb
+++ b/packages/aqemu.rb
@@ -1,0 +1,54 @@
+require 'package'
+
+class Aqemu < Package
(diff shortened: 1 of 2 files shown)


JL2210:
KVM doesn't work on Chrome OS without root.

PR #3348, opened 2019-06-18
Updated Code-oss to 1.35.1

skiman6010:
I'm back with another minor update. 

I don't know if I am the only one with a working install, but I'm going with it.


JL2210:
1.36 is out, however i686 support has been dropped (upstream support was removed in Electron).

skiman6010:
We know.  #3354 

I need to add the 1.36.1 patch though.

PR #3354, opened 2019-07-04
Updated code-oss to 1.36.0

skiman6010:
It's that time of the month again when the person with the working Code installation updates the package fruitlessly.


JL2210:
Note to i386 users: This package won't build. Support for i386 Linux systems was dropped in the upstream version of Electron used in this build. To work around this, upgrade to a 64-bit device.

PR #3355, opened 2019-07-05
Update gtk3 from 3.24.3 to 3.24.9

uberhacker:
Tested on all architectures.  All gtk3 apps crash on release 75.  Please revert to release 74, if possible.


skiman6010:
Does gtk3 crash because of this update or just in general?

uberhacker:
Works great on anything except release 75.  As you might already know, Google pulled the release and reverted back to 74.

skiman6010:
Thought so, just wanted to check. Interesting because when 75 was in beta it was fine. Wonder what changed in the stable release. 76 is good as well.

JL2210:
Darn. And I just updated.

PR #3371, opened 2019-07-15
Add codium package

blinderjay:
vs codium is not a fork of vscode.It is a repository of scripts to automatically build Microsoft's vscode repository into freely-licensed(MIT) binaries with a community-driven default configuration.
> I am not the any of the maintainer of the Project https://github.com/VSCodium/vscodium. However I really hope this package can be add to chromebrew

### DIFF
--- a/packages/codium.rb
+++ b/packages/codium.rb
@@ -1,0 +1,110 @@
+require 'package'
+
+class Codium < Package
+
+  case ARCH
+  when 'x86_64'
+    @arch = 'x64'
+  when 'i686'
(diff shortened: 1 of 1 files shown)


uberhacker:
Thank you for the new package @blinderjay!  Eventually I would like to replace the original code package with this and call it day.  I like the idea of stripping out all the Microsoft fluff.

JL2210:
This has some issues. First of all, it's not Visual Studio Codium, it's simply Codium.

PR #3374, opened 2019-07-20
add exa package

grufwub:
Adds exa (https://github.com/ogham/exa), an ls replacement, v0.9.0 package file

Personally I find the color highlighting a lot easier to parse than regular ls.

Works properly:
- [x] x86_64
- [ ] aarch64 (not able to test)

### DIFF
--- a/packages/exa.rb
+++ b/packages/exa.rb
@@ -1,0 +1,36 @@
+require 'package'
+
+class Exa < Package
+  description 'Modern replacement for \'ls\''
+  homepage 'https://the.exa.website'
+  version '0.9.0'
+  source_url 'https://github.com/ogham/exa/archive/v0.9.0.tar.gz'
+  source_sha256 '96e743ffac0512a278de9ca3277183536ee8b691a46ff200ec27e28108fef783'
+
+  depends_on 'cmake'
+  depends_on 'rust'
+  depends_on 'libgit2'
+
+  def self.build
+    case ARCH
+    when 'aarch64'
+      system 'rustup toolchain install 1.25.0-aarch64-unknown-linux-gnu'
+      system 'rustup default 1.25.0-aarch64-unknown-linux-gnu'
+    when 'armv7l'
+      system 'rustup toolchain install 1.25.0-armv7-unknown-linux-gnueabihf'
+      system 'rustup default 1.25.0-armv7-unknown-linux-gnueabihf'
+    else
+      system 'rustup toolchain install stable'
+      system 'rustup default stable'
+    end
+    system 'cargo build --release'
+  end
+
+  def self.check
+    system 'cargo test --all'
+  end
+
+  def self.install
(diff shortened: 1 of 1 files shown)


packages/exa.rb
```diff
@@ -0,0 +1,36 @@
...
+
+class Exa < Package
+  description 'Modern replacement for \'ls\''
+  homepage 'https://the.exa.website'
+  version '0.9.0'
+  source_url 'https://github.com/ogham/exa/archive/v0.9.0.tar.gz'
+  source_sha256 '96e743ffac0512a278de9ca3277183536ee8b691a46ff200ec27e28108fef783'
+
+  depends_on 'cmake'
+  depends_on 'rust'
```
uberhacker: You don't need the `depends_on 'cmake'` line since that is already included in "core".  Also, since rust is only a build dependency, you don't need to require it if installing from binary: `depends_on 'rust' => :build`.

packages/exa.rb
```diff
@@ -0,0 +1,36 @@
...
+    case ARCH
+    when 'aarch64'
+      system 'rustup toolchain install 1.25.0-aarch64-unknown-linux-gnu'
+      system 'rustup default 1.25.0-aarch64-unknown-linux-gnu'
+    when 'armv7l'
+      system 'rustup toolchain install 1.25.0-armv7-unknown-linux-gnueabihf'
+      system 'rustup default 1.25.0-armv7-unknown-linux-gnueabihf'
+    else
+      system 'rustup toolchain install stable'
+      system 'rustup default stable'
```
uberhacker: Do i686 and x86_64 use the same build command?

packages/exa.rb
```diff
@@ -0,0 +1,36 @@
...
+  depends_on 'libgit2'
+
+  def self.build
+    case ARCH
+    when 'aarch64'
+      system 'rustup toolchain install 1.25.0-aarch64-unknown-linux-gnu'
+      system 'rustup default 1.25.0-aarch64-unknown-linux-gnu'
+    when 'armv7l'
+      system 'rustup toolchain install 1.25.0-armv7-unknown-linux-gnueabihf'
+      system 'rustup default 1.25.0-armv7-unknown-linux-gnueabihf'
```
uberhacker: I imagine you don't need to build specifically for aarch64.  The armv7l build should also work on aarch64.

uberhacker:
Thank you for the new package @grufwub!  Please see comments.

packages/exa.rb
uberhacker: Probably should use a versioned architecture also.

skiman6010: Are you sure this works? Rustup is broken on my `crew` install right now.

uberhacker: It works.  I just tested it but the test.sh script fails. :)

uberhacker: I will fix the arm build when I create the binaries.

PR #3381, opened 2019-07-30
Update cpustat from 0.02.04 to 0.02.09

cstrouse:
Tested on Chromebook Plus v1 (ARM).


uberhacker:
Thank you for the update @cstrouse!  Nice catch with the https.

PR #3382, opened 2019-07-30
Update xxhash from 0.6.3 to 0.7.0

cstrouse:
Adds XXH3 hashing. Tests pass. Tested on Samsung Chromebook Plus v1 (ARM).


uberhacker:

PR #3386, opened 2019-08-01
Remove linux_sources binaries

JL2210:
This is pointless, as it literally just copies the patched files.

It's better to just download the sources and patch them on the fly, and it should reduce server load on Bintray.

### DIFF
--- a/packages/linux_sources.rb
+++ b/packages/linux_sources.rb
@@ -7,19 +7,6 @@ class Linux_sources < Package
   source_url 'https://www.kernel.org/pub/linux/kernel/v4.x/linux-4.14.tar.xz'
   source_sha256 'f81d59477e90a130857ce18dc02f4fbe5725854911db1e7ba770c7cd350f96a7'
 
-  binary_url ({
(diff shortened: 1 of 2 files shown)


uberhacker:
This is not pointless.  It packages all the files together as one fast and easy download instead of the slow method of cloning with git and copying files around (needlessly).

JL2210:
@uberhacker That's not true. The `git clone` command only downloads about 3 MB of data (thanks to `--depth=1`). The download is by no means fast (it takes longer than cloning all the patches and downloading the Linux tarball combined) and it duplicates the files four times (once for each architecture). That's what's unneeded.

uberhacker:
_"...it takes longer than cloning all the patches and downloading the Linux tarball combined"_.  If you think this is true, you must be smoking kid crack.  I don't even know why you care so much.  This will only be used to rebuild the linuxheaders package.  By the way, sommelier still crashes on arm with these patches files so more work is needed anyway.  It does seem to work on x86_64, however, but I'm wondering if it is specific to the 3.x kernel.

PR #3389, opened 2019-08-03
Add pre-built binaries for youtubedl

uberhacker:
Tested on all architectures.


cstrouse:
Works properly on ARM.

PR #3393, opened 2019-08-05
Update tk from 8.6.8 to 8.6.9

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/tk.rb
+++ b/packages/tk.rb
@@ -3,23 +3,24 @@ require 'package'
 class Tk < Package
   description 'Tk is a graphical user interface toolkit that takes developing desktop applications to a higher level than conventional approaches.'
   homepage 'http://www.tcl.tk/'
-  version '8.6.8'
-  source_url 'https://prdownloads.sourceforge.net/tcl/tk8.6.8-src.tar.gz'
-  source_sha256 '49e7bca08dde95195a27f594f7c850b088be357a7c7096e44e1158c7a5fd7b33'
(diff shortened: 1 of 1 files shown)


cstrouse:
I just installed this on ARM and while it installs it doesn't work properly.
```
chronos@localhost /usr/local/lib/crew/packages $ tclsh8.6 
% package require Tk
couldn't load file "/usr/local/lib/libtk8.6.so": libXft.so.2: cannot open shared object file: No such file or directory
```

```
chronos@localhost /usr/local/lib/crew/packages $ wish
wish: error while loading shared libraries: libXft.so.2: cannot open shared object file: No such file or directory
```

uberhacker:
Good catch.  I have updated the dependency to xorg_lib instead of libx11.  This should pick up libxft and libxscrnsaver.

cstrouse:
I haven't done anything with Chromebrew since GUI stuff was added. How do you get $DISPLAY set so that Tk things will launch?

uberhacker:
@cstrouse: You need to install and run sommelier via `crew install sommelier`.  Make sure you carefully follow the postinstall instructions so sommelier starts after a reboot.  This should allow you to test `tkdiff`.

cstrouse:
If this requires sommelier why is it not in the dependency list?

uberhacker:
It doesn't need sommelier to build.  You would need sommelier if you wanted to run an app like tkdiff, however.

PR #3395, opened 2019-08-05
Update BaCoN from 3.7.2 to 3.9.1

cstrouse:
Fix incorrect configure flag in addition to updating version. `--disable-gui ` was not a valid option (or is no longer a valid option).


uberhacker:
Nice!  This package hasn't been updated in awhile.

Issue #3359, opened 2019-07-09
Graphical apps shutdown after sleep mode


vincowl:
## Description
Graphical apps are automatically shut down after sleep mode. This occurs at least with Code and Gemacs. It happened after upgrading gtk3 to 3.24.9 and xdg_utils to 1.1.3_1. This happened before this only with gemacs but it let me think it was linkled to this long lasting bug (https://bugzilla.gnome.org/show_bug.cgi?id=85715 and https://gitlab.gnome.org/GNOME/gtk/issues/221)

Output:
```shell
The XKEYBOARD keymap compiler (xkbcomp) reports:
> Error:            Key <CAPS> added to map for multiple modifiers
>                   Using Mod3, ignoring Lock.
Errors from xkbcomp are not fatal to the X server
(EE) failed to read Wayland events: Broken pipe
```

## Versions and system information
chromebrew: `crew --version`
> 1.2.4

ChromeOS: `cat /etc/lsb-release`
> CHROMEOS_ARC_ANDROID_SDK_VERSION=25
> CHROMEOS_ARC_VERSION=5669872
> CHROMEOS_AUSERVER=https://tools.google.com/service/update2
> CHROMEOS_BOARD_APPID={B00DD0BC-D2C9-BAB4-E66C-81AE3F5A7CED}
> CHROMEOS_CANARY_APPID={90F229CE-83E2-4FAF-8479-E368A34938B1}
> CHROMEOS_DEVSERVER=
> CHROMEOS_RELEASE_APPID={B00DD0BC-D2C9-BAB4-E66C-81AE3F5A7CED}
> CHROMEOS_RELEASE_BOARD=cave-signed-mpkeys
> CHROMEOS_RELEASE_BRANCH_NUMBER=75
> CHROMEOS_RELEASE_BUILDER_PATH=cave-release/R75-12105.75.0
> CHROMEOS_RELEASE_BUILD_NUMBER=12105
> CHROMEOS_RELEASE_BUILD_TYPE=Official Build
> CHROMEOS_RELEASE_CHROME_MILESTONE=75
> CHROMEOS_RELEASE_DESCRIPTION=12105.75.0 (Official Build) stable-channel cave 
> CHROMEOS_RELEASE_KEYSET=mp
> CHROMEOS_RELEASE_NAME=Chrome OS
> CHROMEOS_RELEASE_PATCH_NUMBER=0
> CHROMEOS_RELEASE_TRACK=stable-channel
[...]

PR #3396, opened 2019-08-05
Add gn package

vincowl:
GN is a meta-build system that generates NinjaBuild files. It's meant to be faster and simpler than GYP. It outputs only Ninja build files.

Used by Chromium project


packages/sommelier.rb
```diff
@@ -3,21 +3,21 @@
...
   })
   binary_sha256 ({
-    aarch64: '4d0c30b98a82340e43e3e8f32d9914e1726465b3c8ca961c0f764aaa967f8cd7',
-     armv7l: '4d0c30b98a82340e43e3e8f32d9914e1726465b3c8ca961c0f764aaa967f8cd7',
-       i686: '20fb398592c4a8368fd2292faaa599632f5df717d30a8c27cd2a67631563e0dd',
-     x86_64: '29813316bbb6421f5a54ca45ccb52d0ab88f2a4c84bb84b823bb08fc50dd8216',
+#    aarch64: '4d0c30b98a82340e43e3e8f32d9914e1726465b3c8ca961c0f764aaa967f8cd7',
+#     armv7l: '4d0c30b98a82340e43e3e8f32d9914e1726465b3c8ca961c0f764aaa967f8cd7',
+#       i686: '20fb398592c4a8368fd2292faaa599632f5df717d30a8c27cd2a67631563e0dd',
+#     x86_64: '29813316bbb6421f5a54ca45ccb52d0ab88f2a4c84bb84b823bb08fc50dd8216',
```
uberhacker: I would be better if you just removed the binaries altogether instead of commenting the lines.

packages/sommelier.rb
```diff
@@ -27,9 +27,10 @@ class Sommelier < Package
   depends_on 'xsetroot'
 
   def self.build
-    system 'git clone https://chromium.googlesource.com/chromiumos/containers/sommelier'
+#   system 'git clone https://chromium.googlesource.com/chromiumos/containers/sommelier'
+    system 'git clone https://chromium.googlesource.com/chromiumos/platform2/+/master/vm_tools/sommelier/'
     Dir.chdir ("sommelier") do
-      system "git reset --hard 1382ce084cc40790340d672e8b62ec47733cb860"
+#      system "git reset --hard 59dd453e6efd2c4a0b2b08767c42f47e66a4aaa1"
```
uberhacker: Shouldn't this be uncommented?

uberhacker:
Thank you for the new package @vincowl!  Why is the sommelier change included in this PR?

packages/sommelier.rb
vincowl: ouch... Mistake here : sommelier.rb shouldn't have been changed in this branch... This is my working file

vincowl:
This is a mistake : sommelier.rb shouldn't have been changed in this branch... This is my working file. I needed `gn` tool to try building a new sommelier version from the chromium's repo. Cf #3359

packages/gn.rb
```diff
@@ -0,0 +1,30 @@
+require 'package'
+
+class Gn < Package
+  description 'GN is a meta-build system that generates build files for Ninja.'
+  homepage 'https://gn.googlesource.com/gn/'
+  version 'dca877f'
+  source_url 'https://chromium.googlesource.com/chromiumos/containers/sommelier/+/0.20/README?format=TEXT'
+  source_sha256 'b58d799b16d20abf92369fe0749c73f7398996f0afa9933517051778a8bb16c3'
```
uberhacker: From my experience, this source_sha256 will change everytime you try to download from the chromium repo.  I would suggest a different file.

packages/gn.rb
```diff
@@ -0,0 +1,30 @@
...
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+
+  def self.build
+    system "git clone https://gn.googlesource.com/gn"
+    system 'set C_INCLUDE_PATH /usr/local/include/'
+    Dir.chdir ("gn")
```
uberhacker: This should be `Dir.chdir 'gn' do`.

packages/gn.rb
```diff
@@ -0,0 +1,30 @@
...
+    system "python build/gen.py"
+    system "ninja -C out"
```
uberhacker: You also need to end the `Dir.chdir` with `end`.

packages/gn.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def self.install
+    system "cp out/gn #{CREW_PREFIX}/bin/."
```
uberhacker: A better way would be: `system "install -Dm755 out/gn #{CREW_DEST_PREFIX}/bin/gn`.

packages/gn.rb
```diff
@@ -0,0 +1,30 @@
...
+  version 'dca877f'
+  source_url 'https://chromium.googlesource.com/chromiumos/containers/sommelier/+/0.20/README?format=TEXT'
+  source_sha256 'b58d799b16d20abf92369fe0749c73f7398996f0afa9933517051778a8bb16c3'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+
```
uberhacker: Missing `depends_on 'python27'`.

uberhacker: Missing `depends_on 'meson' => :build`.

packages/gn.rb
vincowl: Any idea which one could do the job ? I faced this problem and thus used the `sommelier.rb`'s config

packages/gn.rb
```diff
@@ -0,0 +1,30 @@
...
+  def self.build
+    system "git clone https://gn.googlesource.com/gn"
+    system 'set C_INCLUDE_PATH /usr/local/include/'
+    Dir.chdir ("gn")
+    system "python build/gen.py"
+    system "ninja -C out"
+  end
+
+  def self.install
```
uberhacker: You will also need a `Dir.chdir 'gn' do ... end` in this section.

packages/gn.rb
vincowl: These are installed on my config. So I didn't note these missings. Is there a tool to get these infos ?

packages/gn.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+
+  def self.build
+    system "git clone https://gn.googlesource.com/gn"
+    system 'set C_INCLUDE_PATH /usr/local/include/'
```
uberhacker: I don't believe this will work.  You will need to do something like:
```
env['C_INCLUDE_PATH'] = "#{CREW_PREFIX}/include"
```

packages/gn.rb
uberhacker: Try:
```
  source_url 'file:///dev/null'
```

packages/gn.rb
uberhacker: You can find out which package you don't need to include with: `grep <package> /path/to/cloned/chromebrew/tools/core_packages.txt`.  You can find out which packages you need simply by examining all the commands you are using.  You need `python27` since you are using the `python` command.  You need `meson` since you are using the `ninja` command and the meson build system.

packages/gn.rb
```diff
@@ -0,0 +1,33 @@
...
+    Dir.chdir ("gn") do
+      system "python build/gen.py"
+      system "ninja -C out"
+    end
+  end
+
+  def self.install
+    system "install -Dm755 out/gn #{CREW_DEST_PREFIX}/bin/gn"
```
uberhacker: If you don't want to use `Dir.chdir ...`, you could do this instead:
```
system "install -Dm755 gn/out/gn #{CREW_DEST_PREFIX}/bin/gn"
```

packages/gn.rb
vincowl: Absolutely right. But it also needs to set `CXX` to `g++`. So : a new dependency to gcc8

uberhacker: You don't need to include `gcc8` as a dependency since it is already included in Chromebrew "core".  Like I said earlier, you can find the package(s) you don't need with `grep gcc8 /path/to/chromebrew/repo/tools/core_packages.txt`.  It should return the the same so you know "core" already includes that package.

vincowl: Right : It's late here and I can hardly keep my eyes open.

PR #3401, opened 2019-08-07
Update diffutils from 3.6 to 3.7

cstrouse:
Tested on Samsung Chromebook Plus V1 (ARM).

### DIFF
--- a/packages/diffutils.rb
+++ b/packages/diffutils.rb
@@ -3,21 +3,13 @@ require 'package'
 class Diffutils < Package
   description 'GNU Diffutils is a package of several programs related to finding differences between files.'
   homepage 'http://www.gnu.org/software/diffutils/'
-  version '3.6'
-  source_url 'https://ftp.gnu.org/gnu/diffutils/diffutils-3.6.tar.xz'
-  source_sha256 'd621e8bdd4b573918c8145f7ae61817d1be9deb4c8d2328a65cea8e11d783bd6'
+  version '3.7'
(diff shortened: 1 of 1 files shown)


packages/diffutils.rb
```diff
@@ -3,21 +3,13 @@
 class Diffutils < Package
   description 'GNU Diffutils is a package of several programs related to finding differences between files.'
   homepage 'http://www.gnu.org/software/diffutils/'
-  version '3.6'
-  source_url 'https://ftp.gnu.org/gnu/diffutils/diffutils-3.6.tar.xz'
-  source_sha256 'd621e8bdd4b573918c8145f7ae61817d1be9deb4c8d2328a65cea8e11d783bd6'
+  version '3.7'
+  source_url 'https://ftp.gnu.org/gnu/diffutils/diffutils-3.7.tar.xz'
```
uberhacker: Can we change this to `source_url 'https://ftpmirror.gnu.org/diffutils/diffutils-3.7.tar.xz'`?  The mirror locations are generally closer in proximity where the request was initiated.

packages/diffutils.rb
```diff
@@ -3,21 +3,13 @@
 class Diffutils < Package
   description 'GNU Diffutils is a package of several programs related to finding differences between files.'
   homepage 'http://www.gnu.org/software/diffutils/'
```
uberhacker: This is redirecting to `https://www.gnu.org/software/diffutils/` so might as well change this also.

PR #3403, opened 2019-08-07
Add faultstat package

cstrouse:
Faultstat measures page fault activity and swap utilization of processes. This is another useful utility for kernel developers from the talented Colin King. Tested on ARM.

### DIFF
--- a/packages/faultstat.rb
+++ b/packages/faultstat.rb
@@ -2,7 +2,7 @@ require 'package'
 
 class Faultstat < Package
   description 'Faultstat measures page fault activity and swap utilization of processes'
-  homepage 'http://kernel.ubuntu.com/~cking/faultstat/'
+  homepage 'https://kernel.ubuntu.com/~cking/faultstat/'
   version '0.01.01'
   source_url 'https://kernel.ubuntu.com/~cking/tarballs/faultstat/faultstat-0.01.01.tar.xz'
   source_sha256 '81218818fe7498411797289bdd0967e82665d2065407be8b5335eaf2959b8991'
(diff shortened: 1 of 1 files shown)


packages/faultstat.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Faultstat < Package
+  description 'Faultstat measures page fault activity and swap utilization of processes'
+  homepage 'http://kernel.ubuntu.com/~cking/faultstat/'
```
uberhacker: `https://kernel.ubuntu.com/~cking/faultstat/`?

packages/faultstat.rb
```diff
@@ -0,0 +1,23 @@
...
+  binary_sha256 ({
+  })
+
+  def self.build
+    system "CPPFLAGS=-I#{CREW_PREFIX}/include/ncurses make" 
+  end
+
+  def self.install
+    system "install -Dm755 faultstat #{CREW_DEST_PREFIX}/bin/faultstat"
+    system "install -Dm644 faultstat.8 #{CREW_DEST_PREFIX}/man/man8/faultstat.8"
```
uberhacker: Can we change the man location to `#{CREW_DEST_PREFIX}/share/man/man8/faultstat.8`?  I only say this because that is the default `MANPAGES` path.  Otherwise, we would need to modify the environment variable.

uberhacker:
Thank you for the new package @cstrouse!

packages/faultstat.rb
cstrouse: Sure. This should be done for all of the other packages that have different `MANPATH` settings, too.

packages/faultstat.rb
cstrouse: Changed to HTTPS.

packages/faultstat.rb
cstrouse: Should I do a PR with a bulk edit and a version bump? Or should we wait and address this as updates happen?

uberhacker: Agreed.  In fact, that could be an issue.  If you can locate all the packages that need the man path updated, we could create an issue with a checklist.  You can just push another commit to this PR.  No need to bump the version since this hasn't been merged yet.

PR #3411, opened 2019-08-07
Update aspell from 0.60.7-rc1-3 to 0.60.7

cstrouse:
This updates from an RC to a release. Tested on ARM.

### DIFF
--- a/packages/aspell.rb
+++ b/packages/aspell.rb
@@ -3,21 +3,13 @@ require 'package'
 class Aspell < Package
   description 'GNU Aspell is a Free and Open Source spell checker designed to eventually replace Ispell.'
   homepage 'http://aspell.net/'
-  version '0.60.7-rc1-3'
-  source_url 'ftp://alpha.gnu.org/gnu/aspell/aspell-0.60.7-rc1.tar.gz'
-  source_sha256 '86b5662f24316142f70c5890787bdc5596625ca3604dfe85926ee61f27f2365e'
+  version '0.60.7-1'
+  source_url 'ftp://ftp.gnu.org/gnu/aspell/aspell-0.60.7.tar.gz'
+  source_sha256 '5ca8fc8cb0370cc6c9eb5b64c6d1bc5d57b3750dbf17887726c3407d833b70e4'
 
   binary_url ({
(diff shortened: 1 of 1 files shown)


packages/aspell.rb
```diff
@@ -3,21 +3,13 @@
 class Aspell < Package
   description 'GNU Aspell is a Free and Open Source spell checker designed to eventually replace Ispell.'
   homepage 'http://aspell.net/'
-  version '0.60.7-rc1-3'
-  source_url 'ftp://alpha.gnu.org/gnu/aspell/aspell-0.60.7-rc1.tar.gz'
-  source_sha256 '86b5662f24316142f70c5890787bdc5596625ca3604dfe85926ee61f27f2365e'
+  version '0.60.7-1'
+  source_url 'ftp://ftp.gnu.org/gnu/aspell/aspell-0.60.7.tar.gz'
```
uberhacker: Can we change this to `https://ftpmirror.gnu.org/aspell/aspell-0.60.7.tar.gz`?  I wish ftp would just go away.

uberhacker: It looks like there is an update to the english dictionary also at https://ftpmirror.gnu.org/aspell/dict/en/aspell6-en-2018.04.16-0.tar.bz2.  Can you please update aspell_en.rb also?

packages/aspell.rb
```diff
@@ -3,21 +3,13 @@
 class Aspell < Package
   description 'GNU Aspell is a Free and Open Source spell checker designed to eventually replace Ispell.'
   homepage 'http://aspell.net/'
-  version '0.60.7-rc1-3'
-  source_url 'ftp://alpha.gnu.org/gnu/aspell/aspell-0.60.7-rc1.tar.gz'
-  source_sha256 '86b5662f24316142f70c5890787bdc5596625ca3604dfe85926ee61f27f2365e'
+  version '0.60.7-1'
```
uberhacker: Why the `-1` on the version?

packages/aspell.rb
uberhacker: Actually, don't worry about it.  I'll update in the pre-built binaries PR.

PR #3418, opened 2019-08-08
Add support for ARM

cstrouse:
This adds ARM support using the project's binary release. It also updates the
package from 1.0.1 to 1.0.4.

Note that to update to the current non-LTS release we would need to build out
own armv7l binaries as they only release 64-bit versions.

Tested on Samsung Chromebook Plus V1 (ARM).


cstrouse:
@uberhacker Please test this on the other architectures. I only have an ARM Chromebook right now.

uberhacker:
Nice!  These are binaries and should work.

PR #3425, opened 2019-08-14
Change arm64 to arm in code.rb

JL2210:
#3391 changed arm to arm64 in Code, which breaks the ARM builds. Change this back and restore order.

Error:  
error

Details on the Discord server.

### DIFF
--- a/packages/code.rb
+++ b/packages/code.rb
@@ -24,8 +24,8 @@ class Code < Package
   depends_on 'sommelier'
 
   case ARCH
-  when 'aarch64', 'armv7l'
-    @arch = 'arm64'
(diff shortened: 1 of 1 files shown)


packages/code.rb
```diff
@@ -24,8 +24,8 @@ class Code < Package
   depends_on 'sommelier'
 
   case ARCH
-  when 'aarch64', 'armv7l'
-    @arch = 'arm64'
+  when 'armv7l', 'aarch64'
+    @arch = 'arm'
```
uberhacker: This won't help.  There is only support for arm64.  Take a look at the code sometime.

PR #3426, opened 2019-08-16
Update vim packages

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/gvim.rb
+++ b/packages/gvim.rb
@@ -8,8 +8,16 @@ class Gvim < Package
   source_sha256 'd543a06a075596e10d1ab4964b6a312ac78c7ccd0628d8b1d6b5deeb7b1d252e'
 
   binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gvim-8.1.1852-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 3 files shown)


packages/gvim.rb
```diff
@@ -71,5 +71,15 @@ def self.postinstall
...
     puts
+    puts "If you receive the following error while attempting to launch gvim:".orange
+    puts
+    puts "E233: cannot open display"
+    puts "Press ENTER or type command to continue"
+    puts "E852: The child process failed to start the GUI"
+    puts "Press ENTER or type command to continue"
+    puts
+    puts "Execute the following:".orange
+    puts "echo 'export GDK_BACKEND=x11' >> ~/.bashrc && source ~/.bashrc".orange
```
JL2210: Perhaps we should change this in `sommelier` instead of overwriting the variable in every package that it breaks?

uberhacker: That is a good thought.  This fix is needed for code/codium also.

skiman6010:
Do you want to rework this before I merge it or should I just go ahead and merge it and you can rework it when you rework sommelier?

uberhacker:
Hold off.  I will update to a newer version after work this evening.

PR #3434, opened 2019-08-24
Add libspng package

cstrouse:
libspng is a C library for reading and writing Portable Network Graphics (PNG)
format files with a focus on security and ease of use.

Tested on ARM.

### DIFF
--- a/packages/libspng.rb
+++ b/packages/libspng.rb
@@ -1,0 +1,32 @@
+require 'package'
+
+class Libspng < Package
+  description 'libspng is a C library for reading and writing Portable Network Graphics (PNG) format files with a focus on security and ease of use.'
+  homepage 'https://libspng.org'
+  version '0.5.0'
+  source_url 'https://gitlab.com/randy408/libspng/uploads/3d980bac86c51368f40af2f1ac79a057/libspng-0.5.0.tar.xz'
+  source_sha256 '220a653802559943ae43fd48f03ba6ff3935a5243766d9ee5ff905240d4399a7'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'cmake' => :build
+  depends_on 'zlibpkg'
+
+  def self.build
+    system 'mkdir -p build'
+    Dir.chdir('build') do
+      system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX} .."
+      system "sed -i 's,build\/tests\/libspng\.pc,build\/libspng.pc,' cmake_install.cmake"
+      system 'make'
+    end
+  end
+
+  def self.install
+    Dir.chdir('build') do
+      system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    end
+  end
+end


packages/libspng.rb
```diff
@@ -0,0 +1,32 @@
...
+  })
+
+  depends_on 'cmake' => :build
+  depends_on 'zlibpkg'
+
+  def self.build
+    system 'mkdir -p build'
+    Dir.chdir('build') do
+      system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX} .."
+      system "sed -i 's,build\/tests\/libspng\.pc,build\/libspng.pc,' cmake_install.cmake"
```
uberhacker: Did you do this to prevent the tests build?

packages/libspng.rb
```diff
@@ -0,0 +1,32 @@
...
+  source_url 'https://gitlab.com/randy408/libspng/uploads/3d980bac86c51368f40af2f1ac79a057/libspng-0.5.0.tar.xz'
+  source_sha256 '220a653802559943ae43fd48f03ba6ff3935a5243766d9ee5ff905240d4399a7'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'cmake' => :build
+  depends_on 'zlibpkg'
```
uberhacker: You can exclude these dependencies since they are already in "core".  See https://github.com/skycocker/chromebrew/blob/master/tools/core_packages.txthttps://github.com/skycocker/chromebrew/blob/master/tools/core_packages.txt.

packages/libspng.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def self.build
+    system 'mkdir -p build'
```
uberhacker: This can be "rubized" by:
```
Dir.mkdir 'build'
```
or
```
FileUtils.mkdir 'build'
```

packages/libspng.rb
```diff
@@ -0,0 +1,32 @@
...
+    Dir.chdir('build') do
+      system "cmake -DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX} .."
```
uberhacker: Is there an option to abstract the library directory?  Maybe `-DCMAKE_INSTALL_LIBDIR=#{CREW_LIB_PREFIX}`?

uberhacker:
Thank you for the new package @cstrouse!  See comments.

packages/libspng.rb
cstrouse: It looks for the libspng.pc file within a build/tests directory which isn't created. Instead, the libspng.pc file is dumped into the build dir so I'm telling it where to find the file.

packages/libspng.rb
cstrouse: It only builds library files and they're put into the right places already so not sure if it would make a difference.

packages/libspng.rb
cstrouse: I'll update these tomorrow.

packages/libspng.rb
cstrouse: I'll swap this out tomorrow.

PR #3435, opened 2019-08-24
Update iptables from 1.6.1 to 1.8.3

cstrouse:
Tested on ARM.


uberhacker:
I'm glad you updated this.  Last time I tried the latest version, I wasn't able to build.

PR #3446, opened 2019-09-04
Add wl_clipboard package

JL2210:
This package provides native integration with the Chrome OS clipboard via. the built-in Wayland server.

### DIFF
--- a/packages/wl_clipboard.rb
+++ b/packages/wl_clipboard.rb
@@ -1,0 +1,30 @@
+require 'package'
+
+class Wl_clipboard < Package
+  description 'Command-line copy/paste utilities for Wayland'
+  homepage 'https://github.com/bugaevc/wl-clipboard.git'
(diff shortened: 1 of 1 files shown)


packages/wl_clipboard.rb
```diff
@@ -0,0 +1,30 @@
...
+  def self.build
+    system 'meson', 'build'
+    system 'ninja', '-C', 'build'
+  end
+
+  def self.install
+    # Meson and Ninja sadly do not support a command-line
+    #  argument for this like Make does
+    ENV['DESTDIR'] = CREW_DEST_DIR
+    system 'ninja', '-C', 'build', 'install'
```
uberhacker: You could also do: `system "DESTDIR=#{CREW_DEST_DIR} ninja -C build install"`.  I find this easier to read without the excessive quotes and commas.  What you have works good enough.

PR #3456, opened 2019-09-11
Change download source from FTP to Github release

cstrouse:
The FTP download link no longer functions and the package cannot be installed.
This fixes that by changing the download source to use the GitHub release.


uberhacker:
Good catch!

PR #3475, opened 2019-09-18
Add libraw1394 package

uberhacker:
libraw1394 provides direct access to the IEEE 1394 bus through the Linux 1394 subsystem's raw1394 user space interface.  See https://sourceforge.net/projects/libraw1394/.  Tested on all architectures.


cstrouse:
`config.guess` is outdated on this and doesn't build on Samsung Chromebook Plus v1.

PR #3476, opened 2019-09-18
Add libdc1394 package

uberhacker:
libdc1394 is a library that provides a complete high level application programming interface (API) for developers who wish to control IEEE 1394 based cameras.  See https://damien.douxchamps.net/ieee1394/libdc1394/.  Tested on all architectures.

### DIFF
--- a/packages/libdc1394.rb
+++ b/packages/libdc1394.rb
@@ -1,0 +1,37 @@
+require 'package'
+
+class Libdc1394 < Package
+  description 'libdc1394 is a library that provides a complete high level application programming interface (API) for developers who wish to control IEEE 1394 based cameras'
+  homepage 'https://damien.douxchamps.net/ieee1394/libdc1394/'
+  version '2.2.0'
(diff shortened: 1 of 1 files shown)


cstrouse:
`config.guess` is outdated on this one, too. Configure step fails for Samsung Chromebook Plus v1.

uberhacker:
@cstrouse: Should we sha256sum check these files for integrity?

cstrouse:
It wouldn't be a bad idea. That's just what the output of configure recommends. It's probably a good idea to grab a specific version as well if we're going to use checksums.

Multiple packages would need to be updated to do things this way. Another option would be to download a recent version as part of the `crew` installation and stash it somewhere and then automatically use that recent version whenever we're building something that uses `configure`.

uberhacker:
If we do that, we should probably pull from a hash commit instead of HEAD.
```
https://git.savannah.gnu.org/gitweb/?p=config.git;a=snapshot;h=1912ca50411bb77fb2c610ef55dd91e332663de9;sf=tgz
```

PR #3478, opened 2019-09-19
Add zvbi package

uberhacker:
The Zapping VBI library, in short ZVBI, provides functions to capture and decode VBI data. 
 See http://zapping.sourceforge.net/ZVBI/.  Tested on all architectures.


cstrouse:
Compiles on ARM.

PR #3480, opened 2019-09-20
Update aws from 1.16.115 to 1.16.241

cstrouse:
Tested on ARM.


uberhacker:
Thank you for the update @cstrouse!  This project is like vim and heroku.  There is an update almost daily. LoL

PR #3481, opened 2019-09-20
Update ansible from 2.6.4 to 2.8.5

cstrouse:
Update and also remove hardcoded version in install step.

Tested on ARM.


uberhacker:

PR #3485, opened 2019-09-22
Update protobuf from 3.6.1.3 to 3.9.2

uberhacker:
Tested on all architectures.  Also retested mosh and qtbase packages which depend on protobuf.


cstrouse:
Tested on ARM.

PR #3489, opened 2019-09-22
Update util_linux to 2.34

JL2210:
Fixes multiple issues with fdisk (library versioning).

Some other aesthetic changes and one build fix.


cstrouse:
Worked properly on ARM.

PR #3493, opened 2019-09-23
Update doxygen from 1.8.14 to 1.8.16

cstrouse:
Update package.

Minor stylistic changes. Update project URL as old one is no 404.

Tested on ARM.


packages/doxygen.rb
```diff
@@ -2,41 +2,33 @@
...
   })
 
   depends_on 'cmake' => :build
   depends_on 'python27' => :build
 
   def self.build
-    system "mkdir -p build"
-    Dir.chdir("build") do
-      system "cmake",
+    system 'mkdir -p build'
```
uberhacker: `Dir.mkdir 'build'`?

uberhacker:
We will need to create binaries since this is a core package.

PR #3494, opened 2019-09-23
Update ed from 1.14.2 to 1.15

cstrouse:
Tested on ARM.

### DIFF
--- a/packages/ed.rb
+++ b/packages/ed.rb
@@ -3,36 +3,28 @@ require 'package'
 class Ed < Package
   description 'GNU ed is a line-oriented text editor.'
   homepage 'http://www.gnu.org/software/ed/ed.html'
-  version '1.14.2-1'
-  source_url 'http://ftpmirror.gnu.org/ed/ed-1.14.2.tar.lz'
-  source_sha256 'f57962ba930d70d02fc71d6be5c5f2346b16992a455ab9c43be7061dec9810db'
+  version '1.15'
+  source_url 'https://ftpmirror.gnu.org/gnu/ed/ed-1.15.tar.lz'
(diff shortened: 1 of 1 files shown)


packages/ed.rb
```diff
@@ -3,36 +3,28 @@
 class Ed < Package
   description 'GNU ed is a line-oriented text editor.'
   homepage 'http://www.gnu.org/software/ed/ed.html'
-  version '1.14.2-1'
-  source_url 'http://ftpmirror.gnu.org/ed/ed-1.14.2.tar.lz'
-  source_sha256 'f57962ba930d70d02fc71d6be5c5f2346b16992a455ab9c43be7061dec9810db'
+  version '1.15'
+  source_url 'https://ftpmirror.gnu.org/gnu/ed/ed-1.15.tar.lz'
```
uberhacker: `https://ftpmirror.gnu.org/ed/ed-1.15.tar.lz` works also.

packages/ed.rb
```diff
@@ -3,36 +3,28 @@
...
   })
 
   # only lz archive is available for ed and it requires lzip.
   depends_on 'lzip' => :build
 
   def self.build
-    system "./configure"
-    system "make"
+    system './configure'
+    system 'make'
```
uberhacker: Probably should add `--prefix=#{CREW_PREFIX}` and `--libdir=#{CREW_LIB_PREFIX}` if there are library files.

packages/ed.rb
cstrouse: Done.

packages/ed.rb
cstrouse: There are no libs. I abstracted the prefix.

PR #3501, opened 2019-09-25
Update nmap from 7.70 to 7.80

cstrouse:
Tested on ARM.

### DIFF
--- a/packages/nmap.rb
+++ b/packages/nmap.rb
@@ -3,21 +3,13 @@ require 'package'
 class Nmap < Package
   description 'Nmap (\'Network Mapper\') is a free and open source (license) utility for network discovery and security auditing.'
   homepage 'https://nmap.org/'
-  version '7.70'
-  source_url 'https://nmap.org/dist/nmap-7.70.tar.bz2'
-  source_sha256 '847b068955f792f4cc247593aca6dc3dc4aae12976169873247488de147a6e18'
+  version '7.80'
+  source_url 'https://nmap.org/dist/nmap-7.80.tar.bz2'
+  source_sha256 'fcfa5a0e42099e12e4bf7a68ebe6fde05553383a682e816a7ec9256ab4773faa'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/nmap-7.70-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/nmap.rb
```diff
@@ -3,21 +3,13 @@
...
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/nmap-7.70-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: '7cc3c71f49254d9cd809f7de8ade5c26b9a07e826244fb2df0817fe00c593420',
-     armv7l: '7cc3c71f49254d9cd809f7de8ade5c26b9a07e826244fb2df0817fe00c593420',
-       i686: '5e7cb838b232fb999cb7684e0ba8b21a61850c3107d74c748c0c7ac2e58dfe0f',
-     x86_64: '27d2b3bdf89573224f0c1436e534540e4cb8a8a74aceef0ce5e8a0987462ae94',
   })
 
   depends_on 'buildessential' => :build
```
uberhacker: We can remove both of these depends_on lines since they're already included in "core".

k0gen:
nmap won't run because of lack of libssl.so.1.0.0 (tested on intel)
```
chronos@localhost / $ nmap
nmap: error while loading shared libraries: libssl.so.1.0.0: cannot open shared object file: No such file or directory
```
Checked what ssl library is available:
```
chronos@localhost / $ sudo ldconfig -v | grep libssl
ldconfig: Path `/lib' given more than once
ldconfig: Path `/lib64' given more than once
ldconfig: Can't stat /libx32: No such file or directory
ldconfig: Path `/usr/lib' given more than once
ldconfig: Path `/usr/lib64' given more than once
ldconfig: Can't stat /usr/libx32: No such file or directory
        libssl.so.1.1 -> libssl.so.1.1
        libssl3.so -> libssl3.so
        libssl.so.48 -> libssl.so.48.0.1
```
Turns out version 1.1 is installed. I did qucik workaround by making symbolic links to libssl and libcrypto:
```
cd /usr/lib64
sudo ln -s libssl.so.1.1 libssl.so.1.0.0
sudo ln -s libcrypto.so.1.1 libcrypto.so.1.0.0
```
to make nmap work.

uberhacker:
It looks like there is a new release available.  Did you want to update this package @cstrouse?  https://nmap.org/dist/nmap-7.91.tar.bz2

PR #3504, opened 2019-09-26
Update mandb from 2.8.4 to 2.8.6

cstrouse:
Update the package, DRY up the sed rewrites, and remove dependencies in core.

Tested on ARM.

Yes, there's an update to 2.8.7 available but the files on the mirrors aren't downloading so this will do for now.

### DIFF
--- a/packages/mandb.rb
+++ b/packages/mandb.rb
@@ -39,7 +39,7 @@ class Mandb < Package
     ].each { |file|
       system "sed -i 's,/var/cache/man,#{CREW_PREFIX}/cache/man,g' #{file}"
     }
-    system "sed -i 's,/var/cache/man,#{CREW_PREFIX}/share/man,g' tools/chconfig"
+    system "sed -i 's,/usr/share/man,#{CREW_PREFIX}/share/man,g' tools/chconfig"
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
   end
 
--- a/packages/mandb.rb
+++ b/packages/mandb.rb
@@ -3,27 +3,14 @@ require 'package'
 class Mandb < Package
   description 'mandb is used to initialise or manually update index database caches that are usually maintained by man.'
   homepage 'http://savannah.nongnu.org/projects/man-db'
-  version '2.8.4'
-  source_url 'https://download.savannah.gnu.org/releases/man-db/man-db-2.8.4.tar.xz'
-  source_sha256 '103c185f9d8269b9ee3b8a4cb27912b3aa393e952731ef96fedc880723472bc3'
+  version '2.8.6'
+  source_url 'https://download.savannah.gnu.org/releases/man-db/man-db-2.8.6.tar.xz'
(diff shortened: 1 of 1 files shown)


packages/mandb.rb
```diff
@@ -35,22 +22,25 @@ def self.build
...
+      'man/man1/whatis.man1',
+      'man/man1/apropos.man1',
+      'man/man1/man.man1',
+      'man/man8/accessdb.man8',
+      'man/man8/mandb.man8',
+      'tools/chconfig'
+    ].each { |file|
+      system "sed -i 's,/var/cache/man,#{CREW_PREFIX}/cache/man,g' #{file}"
+    }
     system "sed -i 's,/var/cache/man,#{CREW_PREFIX}/cache/man,g' tools/chconfig"
```
uberhacker: Redundant line?

packages/mandb.rb
```diff
@@ -35,22 +22,25 @@ def self.build
...
+      'manual/files.me',
+      'man/man1/whatis.man1',
+      'man/man1/apropos.man1',
+      'man/man1/man.man1',
+      'man/man8/accessdb.man8',
+      'man/man8/mandb.man8',
+      'tools/chconfig'
+    ].each { |file|
+      system "sed -i 's,/var/cache/man,#{CREW_PREFIX}/cache/man,g' #{file}"
+    }
```
uberhacker: Nice rubyizing!

packages/mandb.rb
```diff
@@ -35,22 +22,25 @@ def self.build
...
     system "sed -i 's,/var/cache/man,#{CREW_PREFIX}/cache/man,g' tools/chconfig"
-    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
```
uberhacker: It might be worth adding this package to "core" since it goes without saying it is vital to any linux cli user.

uberhacker: Oh wait...it's already in core. :)

packages/mandb.rb
cstrouse: Oh, that's actually a mistake. There *should* be a line after the block but not that one.

packages/mandb.rb
```diff
@@ -35,22 +22,25 @@ def self.build
...
+      'man/man1/whatis.man1',
+      'man/man1/apropos.man1',
+      'man/man1/man.man1',
+      'man/man8/accessdb.man8',
+      'man/man8/mandb.man8',
+      'tools/chconfig'
+    ].each { |file|
+      system "sed -i 's,/var/cache/man,#{CREW_PREFIX}/cache/man,g' #{file}"
+    }
+    system "sed -i 's,/var/cache/man,#{CREW_PREFIX}/share/man,g' tools/chconfig"
```
uberhacker: `system "sed -i 's,/var/share/man,#{CREW_PREFIX}/share/man,g' tools/chconfig"`?

cstrouse: Good catch. Should be /usr/share/man -> #{CREW_PREFIX}/share/man.

PR #3505, opened 2019-09-27
Add config package

uberhacker:
This package should help to update source with outdated config.guess and config.sub scripts.  We can eventually include this in "core".  For now, we can just `depends_on 'config'` in the packages with outdated versions.  To update the scripts, cd into the directory with the outdated files and execute `config .`.  This can be done in the `def self.build` section prior to `./configure`.

### DIFF
--- a/packages/config.rb
+++ b/packages/config.rb
@@ -1,0 +1,34 @@
+require 'package'
+
+class Config < Package
+  description 'The GNU config.guess and config.sub scripts.'
+  homepage 'https://savannah.gnu.org/projects/config'
+  version '1912ca5'
+  source_url 'http://git.savannah.gnu.org/cgit/config.git/snapshot/config-1912ca5.tar.gz'
+  source_sha256 '3b9142de353991dada446a1bc55599838c0e066091ed63a184a9abce50f54724'
+
+  def self.build
+    system "cat << 'EOF' > config
+#!/bin/bash
+if ! test \$1; then
+  echo 'Copy config.guess and config.sub to the destination directory.'
+  echo 'Usage: config <destination directory>'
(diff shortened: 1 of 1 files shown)


cstrouse:
Do you think this is the best way to go about this? I was hoping to do something more automated that doesn't require modifying packages.

If we're going to modify packages another option is to change the `configure` step as follows:

```ruby
def self.build
  system './configure',
    "--prefix=#{CREW_PREFIX}",
    "--build=#{`gcc -dumpmachine`}"
  system 'make'
end
```

The output of `gcc -dumpmachine` is as follows on Samsung Chromebook Plus v1:

```bash
chronos@localhost ~ $ gcc -dumpmachine
armv7l-cros-linux-gnueabihf
```

This yields a working build for me. There may be some codebases where updating the config.* files won't be compatible with the version of autotools being used so maybe this would be preferable?

uberhacker:
@cstrouse:  No, I'm not sure this is the best way.  But if we need to control the version of these files, this is the most elegant way I know.  What is your `arch` output?  This issue only appears to be a problem on armv8 architecture, correct?

cstrouse:
`arch` outputs `aarch64` which is what causes the problem. Since this machine only has 4gb of memory they decided that things would perform better with things being 32-bit. Yes, this only happens on ARMv8 hardware.

The two options seem to be updating the `config.guess` and `config.sub` files or pass the `--build` flag to produce ARMv7 binaries.

Some package managers put a recent version of the config files somewhere on the fs like `/usr/local/share/misc` and then use them automatically as part of their package builds.

Another option (which may be best for the project overall) is to leverage a feature of `autoconf` called `config.site` where you can tell it various settings to use site-wide.

uberhacker:
Not the best way to solve this problem.

PR #3506, opened 2019-09-27
Add pre-built binaries for mandb and nginx

uberhacker:
@cstrouse: I tested on all architectures and now `mandb -c` segfaults.  Also, there are 2 nginx files that get added to the files manifest but are not installed.  You can find from `crew files nginx`:
```
du: cannot access '/lib/systemd/system/man-db.service': No such file or directory
du: cannot access '/lib/systemd/system/man-db.timer': No such file or directory
```
The mandb executable has always been a bit flaky but this latest update killed it completely.  I think we can work on these issues later since everything else seems to work fine.  The mandb package is in core so we need the binaries, otherwise, users see mandb compile from source during crew install.

### DIFF
--- a/packages/mandb.rb
+++ b/packages/mandb.rb
@@ -3,13 +3,21 @@ require 'package'
 class Mandb < Package
   description 'mandb is used to initialise or manually update index database caches that are usually maintained by man.'
   homepage 'http://savannah.nongnu.org/projects/man-db'
-  version '2.8.6'
+  version '2.8.6.1'
   source_url 'https://download.savannah.gnu.org/releases/man-db/man-db-2.8.6.tar.xz'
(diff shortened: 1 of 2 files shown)


uberhacker:
FWIW, I downgraded to mandb 8.2.5 and `mandb -c` worked so the issue is isolated to the latest version, apparently.

cstrouse:
There was a point release to fix this in v2.8.6.1. I just opened a PR (#3507) to fix this.

uberhacker:
@cstrouse: mandb binaries updated for 2.8.6.1.

uberhacker:
@cstrouse: Please merge this when you get a chance.  Without the binaries, it slows down the install for crew since we need to wait until mandb is compiled.

Issue #3511, opened 2019-09-30
[ghc] checksum mismatch once again


cstrouse:
This package needs to be modified to use something other than the checksum of the webpage since that's a moving target that's constantly changing. Any ideas?

```
chronos@localhost / $ crew install cabal
cabal: Common Architecture for Building Applications and Libraries
https://www.haskell.org/cabal/
version 2.2.0.0
The following packages also need to be installed: 
ghc 
Do you agree? [Y/n] y
Proceeding...
ghc: The Glasgow Haskell Compiler is a state-of-the-art, open source compiler and interactive environment for the functional language Haskell.
https://www.haskell.org/ghc/
version 8.4.2
No precompiled binary available for your platform, downloading source...
-=#=-#    #     #                                                                                                        
Checksum mismatch. :/ Try again.
```

uberhacker:
This package probably needs updated anyway.  Did you want to submit a PR @cstrouse?

cstrouse:
I just updated it to the latest version. The problem is that they are constantly changing text on the homepage and we shouldn't be using that as the source file to download. What should we use instead? Something that doesn't change would be like:

```https://downloads.haskell.org/~ghc/8.8.1/ghc-8.8.1-src.tar.xz.sig```

uberhacker:
@cstrouse: That file looks like an excellent candidate.  You could also download the source but that would be overkill just to throw it away.

vincowl:
pandoc, ghc and cabal have been upgraded. No binary available at the moment, but does it solve the problem ?

vincowl:
Please have a look at the related PRs and issues : ghc cannot be upgraded to 8.8.1 due to slack incompatibility

saltedcoffii:
Ghc no longer has binary packages, this can be closed @uberhacker

PR #3518, opened 2019-10-04
Switch source_url to a more stable target

cstrouse:
Fix for #3511. I don't have any Intel-based hardware so please test before merging.

### DIFF
--- a/packages/ghc.rb
+++ b/packages/ghc.rb
@@ -4,8 +4,8 @@ class Ghc < Package
   description 'The Glasgow Haskell Compiler is a state-of-the-art, open source compiler and interactive environment for the functional language Haskell.'
   homepage 'https://www.haskell.org/ghc/'
   version '8.4.2'
-  source_url 'https://www.haskell.org/ghc/download_ghc_8_4_2.html'
-  source_sha256 'a0f4ee8bdd284d869a3db0a0c58f83800a880293ccc19abc5e9debd2e7ab93d7'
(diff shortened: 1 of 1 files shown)


packages/ghc.rb
```diff
@@ -4,8 +4,8 @@ class Ghc < Package
   description 'The Glasgow Haskell Compiler is a state-of-the-art, open source compiler and interactive environment for the functional language Haskell.'
   homepage 'https://www.haskell.org/ghc/'
   version '8.4.2'
-  source_url 'https://www.haskell.org/ghc/download_ghc_8_4_2.html'
-  source_sha256 'a0f4ee8bdd284d869a3db0a0c58f83800a880293ccc19abc5e9debd2e7ab93d7'
+  source_url 'https://downloads.haskell.org/~ghc/8.4.2/ghc-8.4.2-src.tar.xz.sig'
+  source_sha256 '59d3aed5e4fee30e01b6710d57592f16f4f75ce67de97877ddaaa3cec9c6e348'
```
uberhacker: The latest version is https://downloads.haskell.org/~ghc/8.8.1/ghc-8.8.1-src.tar.xz.  Can you please update since we're changing anyway?

cstrouse:
I have the latest version update in a branch but it doesn't work on ARM.

packages/ghc.rb
cstrouse: I have the latest version update in a branch but it doesn't work on ARM.

uberhacker: Fair enough.  I'll see if I can get it working.

PR #3520, opened 2019-10-04
Update vala from 0.39.7 to 0.46.2

cstrouse:
Tested on ARM.

FYI, you need to set `TMPDIR=/usr/local/tmp` when compiling, for example:
```bash
$ TMPDIR=/usr/local/tmp vala hello.vala
```


uberhacker:
You can also add `export TMPDIR=/usr/local/tmp` in `~/.bashrc`.  Always a good idea for other use cases as well.

PR #3528, opened 2019-10-09
Heroku update

skiman6010:


uberhacker:
Thank you for the update @skiman6010.  Please resubmit on a branch with only the heroku update.

PR #3530, opened 2019-10-10
Update Sublime to 3.2.2

skiman6010:


packages/sublime_text.rb
```diff
@@ -3,14 +3,14 @@
 class Sublime_text < Package
   description 'A sophisticated text editor for code, markup and prose.'
   homepage 'https://www.sublimetext.com/'
-  version '3.2'
+  version '3.2.2'
```
uberhacker: 3.2.11?

uberhacker: Might also be good to use the same version scheme as sublime merge.

skiman6010: Yeah I don't know why they version differently between the tools, but this is how the website lists the versions.

uberhacker: Weird.

PR #3534, opened 2019-10-12
Update codium from 1.36.1 to 1.39.1

uberhacker:
Tested on x86_64.


skiman6010:
We should probably at least leave the Extensions host configuration in there so at least we can get extensions without configuring that product.json every upgrade ourselves.

uberhacker:
That is true if we're building from source but this is a binary install which includes that config already.

PR #3557, opened 2019-10-23
Update python27 from 2.7.16 to 2.7.17

cstrouse:
Tested on ARM.

### DIFF
--- a/packages/python27.rb
+++ b/packages/python27.rb
@@ -3,21 +3,13 @@ require 'package'
 class Python27 < Package
   description 'Python is a programming language that lets you work quickly and integrate systems more effectively.'
   homepage 'https://www.python.org/'
-  version '2.7.16'
-  source_url 'https://www.python.org/ftp/python/2.7.16/Python-2.7.16.tar.xz'
-  source_sha256 'f222ef602647eecb6853681156d32de4450a2c39f4de93bd5b20235f2e660ed7'
+  version '2.7.17'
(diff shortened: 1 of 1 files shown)


JL2210:
This'll need to be rebuilt when/if we redo OpenSSL.

uberhacker:
This package is scheduled to be deprecated in 2020 so probably not worth it.  See https://pythonclock.org/.

JL2210:
We can't just ignore this package. Python 2 (although to be deprecated) is still widely used.

uberhacker:
And how many of those widely used packages that depend on this will break with the openssl update?  I'm not saying ignore, just that it might not be worth the effort.  Maybe spending more effort to update the packages that currently depend on 2.7 to 3 would make more sense.

JL2210:
2.7 to 3 is a *huge* leap. They're very incompatible. That said, I would guess that most of them depending on OpenSSL either wouldn't work or don't exist, anyway. A rebuild would probably be the most that's needed if any do exist, and that can be done with the "create binaries" PR that you usually do.

cstrouse:
I agree that there are enough projects that don't support Python 3 yet and may not any time soon and we should probably do what we can to support it for a while. I personally still use Python 2.7 for a lot of things that are just way easier than in 3. For instance, many tools for security research (exploit dev, reverse engineering, CTFs, etc) are easier in 2.7 because working with raw bytes and lower-level stuff is just a pain in 3.

@uberhacker If you don't want to invest much time into this I totally understand. I'd be willing to continue working on it if a few others will commit to helping with testing on various architectures.

PR #3560, opened 2019-10-23
Update libunbound to 1.9.4

JL2210:
Will need rebuilt twice; once now and once with OpenSSL update

*sigh*

Also update homepage and source URL.

### DIFF
--- a/packages/libunbound.rb
+++ b/packages/libunbound.rb
@@ -2,22 +2,14 @@ require 'package'
 
 class Libunbound < Package
   description 'Unbound is a validating, recursive, and caching DNS resolver.'
-  homepage 'https://www.unbound.net/'
-  version '1.8.0'
-  source_url 'https://www.unbound.net/downloads/unbound-1.8.0.tar.gz'
-  source_sha256 '78f79d6d3b643fdcd74a14fc76542250da886c82f82bc55b51e189663d61b83f'
+  homepage 'https://nlnetlabs.nl/projects/unbound/about/'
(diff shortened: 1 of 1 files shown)


packages/libunbound.rb
```diff
@@ -2,22 +2,14 @@
 
 class Libunbound < Package
   description 'Unbound is a validating, recursive, and caching DNS resolver.'
-  homepage 'https://www.unbound.net/'
-  version '1.8.0'
-  source_url 'https://www.unbound.net/downloads/unbound-1.8.0.tar.gz'
-  source_sha256 '78f79d6d3b643fdcd74a14fc76542250da886c82f82bc55b51e189663d61b83f'
+  homepage 'https://nlnetlabs.nl/projects/unbound/download/'
```
uberhacker: What is wrong with the original homepage?  A download url is not a homepage.

JL2210: The old homepage redirects to `https://nlnetlabs.nl/projects/unbound/about` and this is the quickest link to get the latest version. I could have the homepage link to the `about` page if that's more appropriate.

uberhacker: Yeah, about is better than download.

JL2210: @uberhacker done

uberhacker: Thanks.  I'm going to hold off on merging this for now until I get a chance to generate the binaries so folks don't have to wait to compile during install.

PR #3561, opened 2019-10-24
Switch from openssl to libressl

### LINKED ISSUE #3463: Reminder: OpenSSL 1.0.2 is to be deprecated at the end of December
OpenSSL 1.0.2 is to be deprecated at the end of December. We should start upgrading our packages to the newer versions.

Some updates:

* curl has released a new version and should only require an update and a rebuild.

* PHP 5.6 has OpenSSL 1.1 support updates [here](https://github.com/microsoft/php-src/tree/PHP-5.6-security-backports-openssl11).

* I can help find any others that need updates.

Also note that [1.0.2t should not be installed yet](https://github.com/openssl/openssl/issues/7679#issuecomment-531591284).

JL2210:
Fixes #3463

Tested on x86_64

### DIFF
--- a/packages/openssl.rb
+++ b/packages/openssl.rb
@@ -1,72 +1,11 @@
 require 'package'
 
 class Openssl < Package
-  description 'OpenSSL is an open source project that provides a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols.'
+  description 'OpenSSL is a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols.'
   homepage 'https://www.openssl.org/'
-  version '1.0.2p-1'
-  source_url 'https://github.com/openssl/openssl/archive/OpenSSL_1_0_2p.tar.gz'
-  source_sha256 '95ca65a25bdd41e127e5f4054539e8532a46be602b43b44af7c7100172e7cd50'
+  version '3.0.2'
 
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/openssl-1.0.2p-1-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/openssl-1.0.2p-1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/openssl-1.0.2p-1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/openssl-1.0.2p-1-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: '9d2a1bdeac0d26ec7a22a91e017064edf1799f1783ebb7bce89e92e95b68f043',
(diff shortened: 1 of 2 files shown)


packages/openssl.rb
```diff
@@ -1,72 +1,33 @@
...
 
     # remove all files under /etc/ssl (use system's /etc/ssl as is)
-    system "rm", "-rf", "#{CREW_DEST_DIR}/etc"
-
-    # compress man pages
-    system "compressdoc --gzip -9 #{CREW_DEST_PREFIX}/share/man/man1"
-    system "compressdoc --gzip -9 #{CREW_DEST_PREFIX}/share/man/man3"
-    system "compressdoc --gzip -9 #{CREW_DEST_PREFIX}/share/man/man5"
-    system "compressdoc --gzip -9 #{CREW_DEST_PREFIX}/share/man/man7"
+    system "rm", "-rf", "#{CREW_DEST_DIR}/etc/ssl"
```
uberhacker: No FileUtils?  Where is your ruby spirit?

uberhacker:
What does tested mean?  Did you test every package that depends on this (or the previous openssl package) to see if anything breaks?

packages/openssl.rb
JL2210: :P

JL2210:
Did a quick run-test of `cmake`, `curl`, `git`, and `wget` and they seem fine. This needs binaries before it's merged.

uberhacker:
Can you create a separate package called libressl.rb and delete openssl.rb?

JL2210:
I would think that that would screw up device.json.

packages/openssl.rb
```diff
@@ -1,72 +1,33 @@
...
-      system "./Configure --prefix=#{CREW_PREFIX} --openssldir=/etc/ssl #{options} linux-armv4 -march=armv7-a"
-    when 'x86_64'
-      system "./config --prefix=#{CREW_PREFIX} --libdir=lib64 --openssldir=/etc/ssl #{options}"
-    else
-      system "./config --prefix=#{CREW_PREFIX} --openssldir=/etc/ssl #{options}"
-    end
+    system './configure',
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}",
+           '--with-openssldir=/etc/ssl'
```
cstrouse: Why are you using `/etc/ssl` instead of `/usr/local/etc/ssl`?

uberhacker:
How about rename to libressl.rb and update openssl.rb to be a fake package that depends on libressl?

packages/openssl.rb
uberhacker: @cstrouse: I think this is fine.  If you look at the build, we remove `/etc/ssl` so nothing ever gets installed there from this package.

cstrouse:
It compiles on ARM. Testing some packages that I regularly use that require it now. The build does seem to hang for a while at the end but that's probably not the end of the world.

JL2210:
@uberhacker Sounds good.

uberhacker:
Just make it the same as libressl.

packages/openssl.rb
JL2210: Fixed.

packages/openssl.rb
JL2210: We use the system's `/etc/ssl` to get the Google-provided certificates. This is mainly just here because it's easier than providing and updating our own certificates.

JL2210:
@uberhacker Done.

PR #3565, opened 2019-10-24
Update libuv from 1.19.2 to 1.33.1

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/libuv.rb
+++ b/packages/libuv.rb
@@ -3,26 +3,22 @@ require 'package'
 class Libuv < Package
   description 'libuv is a multi-platform support library with a focus on asynchronous I/O.'
   homepage 'http://libuv.org/'
-  version '1.19.2'
-  source_url 'https://dist.libuv.org/dist/v1.19.2/libuv-v1.19.2.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/libuv.rb
```diff
@@ -3,27 +3,23 @@
...
+     x86_64: '679bc3ae5d4ff33b335840fff43087fc2144619eeb678f5c047689c74840faf8',
   })
 
-  depends_on 'automake'
-  depends_on 'glibc'
-  depends_on 'libtool'
-
   def self.build
     system './autogen.sh'
     system './configure',
```
cstrouse: Should we be explicit and add `--libdir` and `--includedir` here?

cstrouse:
Compiles on ARM. All but one of the tests pass FWIW.

packages/libuv.rb
uberhacker: We already have `--libdir` and `--includedir` is usually `#{CREW_PREFIX}/include` if you set `--prefix`.

uberhacker:
@cstrouse: Which test failed on aarch64?  Can you please paste the error output?

PR #3567, opened 2019-10-24
Break lua package up into multiple versions

### LINKED ISSUE #3358: lua missing pkgconfig .pc file
## Description
The `lua` package in missing the necessary ".pc" file that allows `pkg-config` to generate project compile time information. According to a project `Makefile` script I have (copyright 2018), that ".pc" file can be named any of the following:
```
          LUAPKG := $(shell for p in lua5.3 lua-5.3 lua53 lua5.2 lua-5.2 lua52 lua5.1 lua-5.1 lua51 lua ; do $(PKG_CONFIG) --exists $$p && echo $$p && break ; done)
          LUACOPT = $(shell $(PKG_CONFIG) $(LUAPKG) --cflags)
          LUALOPT = $(shell $(PKG_CONFIG) $(LUAPKG) --libs)
```  

The current version of `lua` is:
```
$ lua -v
Lua 5.3.5  Copyright (C) 1994-2018 Lua.org, PUC-Rio
```

From the shell, the expected output should be:
```
$ 
```

chromebrew: `1.2.4`

ChromeOS:
```
CHROMEOS_BOARD_APPID={81C04B02-A863-BFBB-F77C-7707347DC958}
CHROMEOS_DEVSERVER=
CHROMEOS_RELEASE_APPID={81C04B02-A863-BFBB-F77C-7707347DC958}
CHROMEOS_RELEASE_BOARD=veyron_speedy-signed-mp-v3keys
CHROMEOS_RELEASE_BRANCH_NUMBER=75
CHROMEOS_RELEASE_BUILDER_PATH=veyron_speedy-release/R75-12105.75.0
CHROMEOS_RELEASE_BUILD_NUMBER=12105
CHROMEOS_RELEASE_CHROME_MILESTONE=75
CHROMEOS_RELEASE_DESCRIPTION=12105.75.0 (Official Build) stable-channel veyron_speedy 
CHROMEOS_RELEASE_KEYSET=mp-v3
CHROMEOS_RELEASE_TRACK=stable-channel
CHROMEOS_RELEASE_VERSION=12105.75.0
DEVICETYPE=CHROMEBOOK
GOOGLE_RELEASE=12105.75.0
```

Model: `Asus CP201`

Architecture: `armv7l`

## Addtional information
`luajit` is also provided, but it ONLY works with `lua-5.1`.

- [ ] step one: change current `lua` package name to `lua-5.3` or `lua53`
- [ ] step two: ensure `lua-5.3.pc` or `lua53.pc` is generated into `./lib/pkgconfig/`
- [ ] step three: add Lua version 5.1 and 5.2 to the package repository

Example `lua53.pc` generation:
```
$ cat > lua53.pc << "EOF"
V=5.3
R=5.3.5

prefix=/usr/local
INSTALL_BIN=${prefix}/bin
INSTALL_INC=${prefix}/include
INSTALL_LIB=${prefix}/lib
INSTALL_MAN=${prefix}/share/man/man1
INSTALL_LMOD=${prefix}/share/lua/${V}
INSTALL_CMOD=${prefix}/lib/lua/${V}
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: Lua
Description: An Extensible Extension Language
Version: ${R}
Requires:
Libs: -L${libdir} -llua53 -lm -ldl
Cflags: -I${includedir}
EOF
```

Cheers

Paul

cstrouse:
Some projects only work with older lua releases. This allows those projects to
function. The missing pkgconfig files now are generated.

Tested on ARM. Fixes #3358.


cstrouse:
@uberhacker Should we add a `lua` fake package that installs `lua53` so that existing packages that depend on `lua` don't break? Or should we update those packages to depend one of the specific lua packages?

packages/lua51.rb
```diff
@@ -0,0 +1,51 @@
...
+INSTALL_MAN=#{CREW_PREFIX}/share/man/man1
+INSTALL_LMOD=#{CREW_PREFIX}/share/lua/5.1
+INSTALL_CMOD=#{CREW_PREFIX}/lib/lua/5.1
+exec_prefix=#{CREW_PREFIX}
+libdir=#{CREW_PREFIX}/lib
+includedir=#{CREW_PREFIX}/include
+
+Name: Lua
+Description: An Extensible Extension Language
+Version: 5.1.5
```
uberhacker: Should we use variables for the versions?

uberhacker:
I was thinking we could create a generic lua package that lets you choose the version.  Something like how we do it with php.

packages/lua51.rb
JL2210: An easy way is to just use `#{version}`; it's set by `version '...'`.

cstrouse: Our package version may not always match the upstream.

uberhacker: Can you add another variable `@ver = 5.1`?

packages/lua51.rb
```diff
@@ -0,0 +1,51 @@
...
+V=5.1
+R=5.1.5
+
+prefix=#{CREW_PREFIX}
+INSTALL_BIN=#{CREW_PREFIX}/bin
+INSTALL_INC=#{CREW_PREFIX}/include
+INSTALL_LIB=#{CREW_LIB_PREFIX}
+INSTALL_MAN=#{CREW_PREFIX}/share/man/man1
+INSTALL_LMOD=#{CREW_PREFIX}/share/lua/5.1
+INSTALL_CMOD=#{CREW_PREFIX}/lib/lua/5.1
```
uberhacker: Use version variable?  Not sure if this works like bash.
```
INSTALL_LMOD=#{CREW_PREFIX}/share/lua/\$V
INSTALL_CMOD=#{CREW_PREFIX}/lib/lua/\$V
```

cstrouse: I had that but for some reason it wasn't getting replaced.

uberhacker: You could do this:
```
    V=5.1
    system "cat > #{CREW_DEST_PREFIX}/lib/pkgconfig/lua51.pc << \"EOF\"
prefix=#{CREW_PREFIX}
INSTALL_BIN=#{CREW_PREFIX}/bin
INSTALL_INC=#{CREW_PREFIX}/include
INSTALL_LIB=#{CREW_LIB_PREFIX}
INSTALL_MAN=#{CREW_PREFIX}/share/man/man1
INSTALL_LMOD=#{CREW_PREFIX}/share/lua/#{V}
INSTALL_CMOD=#{CREW_PREFIX}/lib/lua/#{V}
```
Set the variable outside of the heredoc.

cstrouse:
> I was thinking we could create a generic lua package that lets you choose the version. Something like how we do it with php.

I really don't like how the php packages are setup especially the "5.x.x-7.x.x" versioning. I'm also not a fan of how you install the php virtual package but have to remove php5 or php7. It would be nice to be able to track that you installed via the virtual package and remove via that name.

From what I've read people are pretty commited to either luajit, lua-5.1, lua-5.2, or lua-5.3 for their projects and aren't likely to change. Given this users who work with lua and lua-dependent packages will probably install multiple versions.

uberhacker:
With php, you cannot install multiple versions so lua may require a different approach.

uberhacker:
@cstrouse: The php packages have been reworked to install in a more standard way.  Would it make sense to use that same approach for the lua packages?

uberhacker:
@cstrouse:  Any update on this?  I'd like to get it merged in before EoM so you can get credit for Hacktoberfest (if you joined).

uberhacker:
Closing due to inactivity for over a year.

PR #3587, opened 2019-10-27
v2ray package

dx0eu:
A platform for building proxies to bypass network restrictions. https://www.v2ray.com/ github repo https://github.com/v2ray/v2ray-core


uberhacker:
@dx0eu:  Thank you for the update.  Please create a separate branch (not master) on the next update so we don't have all the updates in the commit history.

dx0eu:
@uberhacker  ok, I will update it in time, forgive me may not be very soon

PR #3589, opened 2019-10-27
Vim 8.1.2225

### LINKED ISSUE #3588: Upgrade vim to 8.1.2225
Vim is actually built against dynamic python3interp, that crashes vim plugins. Upragde to 8.1.2225 resolves this

vincowl:
Fixes #3588 
Removing python27 dependency is optional. I think it should be but what about backward compatibility ?

### DIFF
--- a/packages/vim.rb
+++ b/packages/vim.rb
@@ -3,9 +3,9 @@ require 'package'
 class Vim < Package
   description 'Vim is a highly configurable text editor built to make creating and changing any kind of text very efficient.'
   homepage 'http://www.vim.org/'
-  version '8.1.1915'
-  source_url 'https://github.com/vim/vim/archive/v8.1.1915.tar.gz'
-  source_sha256 '508bcffd340497d6279fb2d5aa98ff43190cf7983a87fe4838fb780446f900a9'
+  version '8.1.2225'
+  source_url 'https://github.com/vim/vim/archive/v8.1.2225.tar.gz'
+  source_sha256 'a88cf5fb07701d92ea79a3cc8f6e020e6b9bf07c0dfa72f0dde8a65f87e30f2d'
 
   if ARGV[0] == 'install'
     gvim = `which gvim 2> /dev/null`.chomp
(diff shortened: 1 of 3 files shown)


packages/vim.rb
```diff
@@ -3,29 +3,20 @@
...
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/vim-8.1.1915-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: '7e9139cbb1940f272ea3414facfbe34e37f59e476868d8ceb7fe2419ac1261e1',
-     armv7l: '7e9139cbb1940f272ea3414facfbe34e37f59e476868d8ceb7fe2419ac1261e1',
-       i686: 'a3714ab96726d3630f6903b535b4af5808fdcc1101aabf08b0d79544d676e984',
-     x86_64: 'b5eeeb48785b05e94227747138b72b8ea48502b32aa0b6b7559e6582fab85a99',
   })
 
-  depends_on 'python27' => :build
```
JL2210: Removing this will not compile in Python support. The way it was done before did make the dependency optional.

packages/vim.rb
```diff
@@ -3,29 +3,20 @@
 class Vim < Package
   description 'Vim is a highly configurable text editor built to make creating and changing any kind of text very efficient.'
   homepage 'http://www.vim.org/'
-  version '8.1.1915'
-  source_url 'https://github.com/vim/vim/archive/v8.1.1915.tar.gz'
-  source_sha256 '508bcffd340497d6279fb2d5aa98ff43190cf7983a87fe4838fb780446f900a9'
+  version '8.1.2225'
```
JL2210: Please update `gvim` and `vim_runtime` too.

packages/vim.rb
vincowl: Not sure to understand what you mean. I removed the dependancy and the `"--enable-pythoninterp=dynamic"` (everything related to python 2.7). Python3 dependancy and interp are kept. This is fully functional on my machine, and allows me to run plugins that need python. Since Python 2.7 ends its life in about 2 months, I thought it could be removed yet. But maybe it would be better to wait 2 months and act in a way that will be explained by vim's team (Hopefully!)

packages/vim.rb
vincowl: Done in the commit 60ba049. I kept the same philosophy about python 2.7. I will apply the same decision once we will agree on what to do (keep or remove the python 2.7 link)

packages/vim.rb
uberhacker: Yes, let's no remove python 2.7 support.  Doesn't hurt to keep it, right?

vincowl: No problem !

PR #3595, opened 2019-10-27
Upgrade Hunspell to 1.7.0 and add french dictionary

vincowl:


packages/hunspell.rb
```diff
@@ -37,14 +31,18 @@ def self.build
...
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+
+  def self.postinstall
     system 'git clone -b libreoffice-6-1 --depth 1 git://anongit.freedesktop.org/libreoffice/dictionaries.git'
     system "install -Dm644 dictionaries/en/en_US.aff #{CREW_DEST_DIR}$HOME/Library/Spelling/en_US.aff"
     system "install -Dm644 dictionaries/en/en_US.dic #{CREW_DEST_DIR}$HOME/Library/Spelling/en_US.dic"
     system "install -Dm644 dictionaries/en/en_US.aff $HOME/Library/Spelling/en_US.aff"
     system "install -Dm644 dictionaries/en/en_US.dic $HOME/Library/Spelling/en_US.dic"
```
vincowl: It seems more logical to me to set these steps in postinstall. Thus one could run postinstall to update dictionaries regularly.

packages/hunspell.rb
```diff
@@ -37,14 +31,18 @@ def self.build
...
-  end
-
-  def self.postinstall
+    system "install -Dm644 dictionaries/fr_FR/fr.aff #{CREW_DEST_DIR}$HOME/Library/Spelling/fr_FR.aff"
+    system "install -Dm644 dictionaries/fr_FR/fr.dic #{CREW_DEST_DIR}$HOME/Library/Spelling/fr_FR.dic"
+    system "install -Dm644 dictionaries/fr_FR/fr.aff $HOME/Library/Spelling/fr_FR.aff"
+    system "install -Dm644 dictionaries/fr_FR/fr.dic $HOME/Library/Spelling/fr_FR.dic"
```
vincowl: How could we allow users to install their own dictionaries ? Is it part of these packages ? This is an easy solution but probably not the cleanest one especially in term of separation of concerns

packages/hunspell.rb
```diff
@@ -3,32 +3,26 @@
...
   })
   binary_sha256 ({
-    aarch64: '2122527ab9f428c6d88d22ea142c134bf648b933963178f368ce5e9832d93317',
-     armv7l: '2122527ab9f428c6d88d22ea142c134bf648b933963178f368ce5e9832d93317',
-       i686: '1e7bd3254580e4a211b1b8208e548675602290e9c7a1105ba66af6d4c9cee8da',
-     x86_64: '76e93eec7e0cec41707fd68f0f21351e94c25e8b993431f3bf400693d3de8251',
   })
 
   depends_on 'readline'
+  depends_on 'ncurses'
```
uberhacker: No need to include either of these dependencies since they're already included in "core".  See https://github.com/skycocker/chromebrew/blob/master/tools/core_packages.txt.

packages/hunspell.rb
```diff
@@ -37,14 +31,18 @@ def self.build
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+
+  def self.postinstall
     system 'git clone -b libreoffice-6-1 --depth 1 git://anongit.freedesktop.org/libreoffice/dictionaries.git'
     system "install -Dm644 dictionaries/en/en_US.aff #{CREW_DEST_DIR}$HOME/Library/Spelling/en_US.aff"
     system "install -Dm644 dictionaries/en/en_US.dic #{CREW_DEST_DIR}$HOME/Library/Spelling/en_US.dic"
```
uberhacker: This package was created before we added the constant but you could replace `#{CREW_DEST_DIR}$HOME` with `#{CREW_DEST_HOME}`.

packages/hunspell.rb
uberhacker: This is actually a great way to separate dictionary updates from the package.  We can run `crew postinstall hunspell` to perform the updates.  We might want to add a postinstall message to explain:
```
puts "To update the dictionaries periodically, execute 'crew postinstall hunspell'.".lightblue
```

uberhacker: We can just update this package when someone wants to add another language.

packages/hunspell.rb
uberhacker: That said, we still want the `system "install ...` lines in the install section.  We could add something like this in the postinstall section:
```
system 'git clone -b libreoffice-6-1 --depth 1 git://anongit.freedesktop.org/libreoffice/dictionaries.git'
FileUtils.c 'dictionaries/en/*', "#{HOME}/Library/Spelling/"
FileUtils.c 'dictionaries/fr/*', "#{HOME}/Library/Spelling/"
...
```

packages/hunspell.rb
```diff
@@ -37,14 +31,18 @@ def self.build
...
-  end
-
-  def self.postinstall
+    system "install -Dm644 dictionaries/fr_FR/fr.aff #{CREW_DEST_DIR}$HOME/Library/Spelling/fr_FR.aff"
+    system "install -Dm644 dictionaries/fr_FR/fr.dic #{CREW_DEST_DIR}$HOME/Library/Spelling/fr_FR.dic"
+    system "install -Dm644 dictionaries/fr_FR/fr.aff $HOME/Library/Spelling/fr_FR.aff"
+    system "install -Dm644 dictionaries/fr_FR/fr.dic $HOME/Library/Spelling/fr_FR.dic"
     puts
     puts "To complete the installation, execute the following:".lightblue
     puts "echo 'export DICTIONARY=en_US' >> ~/.bashrc && source ~/.bashrc".lightblue
```
uberhacker: Might also want to add something like:
```
puts "To change to another language for example, execute:".lightblue
puts "echo 'export DICTIONARY=fr_FR' >> ~/.bashrc && source ~/.bashrc".lightblue
```

packages/hunspell.rb
vincowl: The side effect is that one could have a large amount of installed files he doesn't need : for instance here, if one doesn't need french dictionaries, he will be forced to install them

packages/hunspell.rb
uberhacker: These lines aren't needed, actually, and can be removed.

packages/hunspell.rb
vincowl: > That said, we still want the `system "install ...` lines in the install section. 

Why do you need it in the install section ? Not clear to me

> We could add something like this in the postinstall section:
> 
> ```
> system 'git clone -b libreoffice-6-1 --depth 1 git://anongit.freedesktop.org/libreoffice/dictionaries.git'
> FileUtils.cp Dir.glob('dictionaries/en/*'), "#{HOME}/Library/Spelling/"
> FileUtils.cp Dir.glob('dictionaries/fr/*'), "#{HOME}/Library/Spelling/"
> ...
> FileUtils.rm_rf 'dictionaries/'
> ```

Why not simply replace actual lines by these ones ? 

Why removing the `dictionaries` folder ? Isn't it automatically placed in tmp and then deleted once postinstall ends like it is done with install ?

packages/hunspell.rb
uberhacker: We could add languages in separate packages like we do with aspell.

packages/hunspell.rb
uberhacker: The postinstall section can be run on demand after the package is installed.  We want the `system "install ...` lines in the install section to add to the package metadata.  Otherwise, the package would not track these files for removal when the package is uninstalled.

packages/hunspell.rb
vincowl: Why not ! No install step, just a postinstall one to allow manual periodically updates ? Is there a stub for `version`, `source_url` and `source_sha256` ?

vincowl: We could keep en_US by default in hunspell

packages/hunspell.rb
vincowl: Fine ! Do the latest commit fulfills these "requirements" ?

packages/hunspell.rb
vincowl: Example given in 58ec722

uberhacker: There is `template.rb.template` but it might be easier just to copy hunspell.rb and modify.

uberhacker:
Would this work?
```
  if ARGV[0] == 'install'
    puts
    puts "Enter your preferred language:"
    puts "1 = English"
    puts "2 = French"
    puts "3 = Spanish"
    puts "0 = Cancel"

    while version = STDIN.gets.chomp
      case version
      when '1'
        depends_on 'hunspell_en'
      when '2'
        depends_on 'hunspell_fr'
      when '3'
        depends_on 'hunspell_es'
      end
    end
  end
```

packages/hunspell.rb
vincowl: Isn't there a way to download absolutely nothing ?

vincowl:
> Would this work?

I'll try it tomorrow. It's about 5a.m here

packages/hunspell.rb
uberhacker: There is this but I don't like it since it doesn't uniquely identify the package and can cause a conflict with other packages that are being installed at the same time.
```
  source_url 'file:///dev/null'
```

uberhacker:
> How does something copied in /usr/local/tmp/crew/dest/home can be persistent ? Do you move it to $HOME once installed is done ? If so, I'll invert the modifications I have made

Yes, everything is eventually copied from the tmp folders to the correct destinations.  Take some time and study the crew command to see how the process works.

packages/hunspell.rb
vincowl: OK. Let's keep the source_url to hunspell.tar.gz so.

vincowl:
> Would this work?
> 
> ```
>   if ARGV[0] == 'install'
>     puts
>     puts "Enter your preferred language:"
>     ...
> ```

Indeed it seems to perfectly do the job. I love this solution : clean, easy to maintain and extend ! I created a spanish package and rename the other to be able to add different "version" of the languages. For instance : `hunspell_en` becomes `hunspell_en_us` to enable the installation of `hunspell_en_UK`

packages/hunspell.rb
```diff
@@ -3,32 +3,49 @@
...
+          depends_on 'hunspell_fr_fr'
+          break
+        when '3'
+          depends_on 'hunspell_es_any'
+          break
+        when '0'
+          abort
+      end
+    end
+  end
```
vincowl: It works like a charm, but as I install `hunspell` with en_US for instance and then install `hunspell_fr_fr`, the `depends_on 'hunspell'` in `hunspell_fr_fr` re-ask which language to install. Any idea how to avoid this like checking if `hunspell` is already installed (don't known how to do it in an other way than a `grep` in `crew list installed`)?

packages/hunspell_en_us.rb
```diff
@@ -0,0 +1,36 @@
...
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'hunspell'
+
+  def self.install
+    system 'git clone -b libreoffice-6-1 --depth 1 git://anongit.freedesktop.org/libreoffice/dictionaries.git'
+    system 'mkdir -p $HOME/Library/Spelling'
```
uberhacker: This won't work in single quotes.  In any event, this line is not needed since the commands below already handle the directory.

packages/hunspell_es_any.rb
```diff
@@ -0,0 +1,36 @@
...
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'hunspell'
+
+  def self.install
+    system 'git clone -b libreoffice-6-1 --depth 1 git://anongit.freedesktop.org/libreoffice/dictionaries.git'
+    system 'mkdir -p $HOME/Library/Spelling'
```
uberhacker: Please remove this line.  It is not needed.

packages/hunspell_fr_fr.rb
```diff
@@ -0,0 +1,36 @@
...
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'hunspell'
+
+  def self.install
+    system 'git clone -b libreoffice-6-1 --depth 1 git://anongit.freedesktop.org/libreoffice/dictionaries.git'
+    system 'mkdir -p $HOME/Library/Spelling'
```
uberhacker: Please remove this line.

packages/hunspell_fr_fr.rb
```diff
@@ -0,0 +1,36 @@
...
+    system "install -Dm644 dictionaries/fr_FR/fr.dic #{CREW_DEST_HOME}/Library/Spelling/fr_FR.dic"
+  end
+
+  def self.postinstall
+    system 'git clone -b libreoffice-6-1 --depth 1 git://anongit.freedesktop.org/libreoffice/dictionaries.git'
+    FileUtils.cp 'dictionaries/fr_FR/fr.aff', "#{HOME}/Library/Spelling/fr_FR.aff"
+    FileUtils.cp 'dictionaries/fr_FR/fr.dic', "#{HOME}/Library/Spelling/fr_FR.dic"
+    FileUtils.rm_rf 'dictionaries/'
+    puts
+    puts "To update the dictionaries periodically, execute 'crew postinstall hunspell_fr'".lightblue
```
uberhacker: `crew postinstall hunspell_fr_fr`?

packages/hunspell.rb
uberhacker: I suppose you could remove the `depends_on 'hunspell'` from each language package and include the build there instead of here.  Or if you don't want to repeat for each language, create a hunspell_base.rb package and depend on that instead?

vincowl: I don't like both of them : in the first one, it will duplicate the build function and complexify maintenance, and in the second one it will multiply packages to maintain. But the second one seems better to me. Indeed : language packages won't change with hunspell version upgrade (to simplify, package url in header is here only to avoid chromebrew crashes). Having build duplicated inside these files will force to update them at each new unspell version, while depending on a new package will lead to update only 2 packages at each version upgrade. Hope I am clear...

vincowl: Not sure what to put where and how to design the dependency chain...

uberhacker: Below is probably the most elegant way to solve this problem.
hunspell.rb
```
class Hunspell < Package
  description 'Hunspell is a spell checker and morphological analyzer library'
  homepage 'http://hunspell.github.io/'
  version '1.6.2'
  source_url 'https://github.com/hunspell/hunspell/archive/v1.6.2.tar.gz'
  source_sha256 '3cd9ceb062fe5814f668e4f22b2fa6e3ba0b339b921739541ce180cac4d6f4c4'
  version '1.7.0'
  source_url 'https://github.com/hunspell/hunspell/archive/v1.7.0.tar.gz'
  source_sha256 'bb27b86eb910a8285407cf3ca33b62643a02798cf2eef468c0a74f6c3ee6bc8a'

  if ARGV[0] == 'install'
    puts
    puts "Enter your preferred language:"
    puts "1 = English (US)"
    puts "2 = French (FR)"
    puts "3 = Spanish (ANY)"
    puts "0 = Cancel"

    while version = STDIN.gets.chomp
      case version
        when '1'
          depends_on 'hunspell_en_us'
          break
        when '2'
          depends_on 'hunspell_fr_fr'
          break
        when '3'
          depends_on 'hunspell_es_any'
          break
        when '0'
          abort
      end
    end
  end
end
```

uberhacker: hunspell_base.rb:
```
class Hunspell_base < Package
  description 'Hunspell is a spell checker and morphological analyzer library'
  homepage 'http://hunspell.github.io/'
  version '1.7.0'
  source_url 'https://github.com/hunspell/hunspell/archive/v1.7.0.tar.gz'
  source_sha256 'bb27b86eb910a8285407cf3ca33b62643a02798cf2eef468c0a74f6c3ee6bc8a'

  binary_url ({
  })
  binary_sha256 ({
  })

  depends_on 'readline'

  def self.build
    system 'autoreconf -vfi'
    system './configure',
           "--prefix=#{CREW_PREFIX}",
           "--include=#{CREW_PREFIX}/include",
           '--with-ui',
           '--with-readline'
    system "sed -i 's,/usr/share,#{CREW_PREFIX}/share,g' man/hunspell.1"
    system "sed -i 's,/usr/share,#{CREW_PREFIX}/share,g' src/tools/hunspell.cxx"
    system "sed -i 's,ncurses.h,#{CREW_PREFIX}/include/ncursesw/ncurses.h,' src/tools/hunspell.cxx"
    system 'make'
  end

  def self.install
  end
end
```
The instead of `depends_on 'hunspell'` in each language package, replace with `depends_on 'hunspell_base'`.
Hope that makes sense.

vincowl: Yes : that was exactly what I am designing and testing at the very moment. There will probably be a trap : it won't prevent you from installing only hunspell_base, but as it has no dictionary linked to it, its impact will be insignificant.

vincowl: Job's done. It works perfectly well here ! Thanks

packages/hunspell.rb
```diff
@@ -3,51 +3,39 @@
...
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/hunspell-1.6.2-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: '2122527ab9f428c6d88d22ea142c134bf648b933963178f368ce5e9832d93317',
-     armv7l: '2122527ab9f428c6d88d22ea142c134bf648b933963178f368ce5e9832d93317',
-       i686: '1e7bd3254580e4a211b1b8208e548675602290e9c7a1105ba66af6d4c9cee8da',
-     x86_64: '76e93eec7e0cec41707fd68f0f21351e94c25e8b993431f3bf400693d3de8251',
   })
 
   depends_on 'readline'
```
uberhacker: Don't need this line here.

uberhacker:
Nice work @vincowl!  Glad you stuck to it!

PR #3608, opened 2019-10-29
Update vim from 8.1.2225 to 8.1.2234

cstrouse:
Tested on ARM.


uberhacker:
Update gvim also?

uberhacker:
@cstrouse: Can you please include the gvim update with this PR.  Then it will be good to merge.  Actually, since it's already been a few days, a newer release is available. :)

cstrouse:
Just updated to the new release for all three vim packages.

uberhacker:
Excellent!  Thank you!

PR #3614, opened 2019-10-29
[stack] Upgrade to 2.1.3

vincowl:


vincowl:
postinstall indicates `crew remove cabal ghc stack` to remove completely stack but cabal and ghc are not part of stack's dependencies. Shall we add them or update this comment ?

uberhacker:
We can do that on the next update if you think it will help avoid confusion.

vincowl:
I am dealing with something related to it. I wanted to upgrade wkhtmltox package, which is linked to pandoc, which requires stack, cabal, ghc and ltinfo. All of these can be upgraded. Cabal depends on ghc which depends on stack. So installing cabal ask to install ghc, stack and libinfo. But ghc build step uses stack. I don't understand why stack is not installed first even if it is a "second level of dependency". Is a bug or something I missed ?

PR #3617, opened 2019-10-29
Upgrade cabal to 2.4.1.0

vincowl:


packages/cabal.rb
```diff
@@ -3,16 +3,10 @@
...
-    source_url 'https://www.haskell.org/cabal/release/cabal-install-2.2.0.0/cabal-install-2.2.0.0-i386-unknown-linux.tar.gz'
-    source_sha256 '20cc07bc7c813fac3e88bb308c972e3a28ade30dd32783d7a74cdadc74c5079e'
-  when 'x86_64'
-    source_url 'https://www.haskell.org/cabal/release/cabal-install-2.2.0.0/cabal-install-2.2.0.0-x86_64-unknown-linux.tar.gz'
-    source_sha256 '35c14663496cd18bf33cdcf4441a77eb169b0a93cfd0b79081c6816ada53f05c'
-  end
-
+  version '2.4.1.0'
+  source_url 'https://www.haskell.org/cabal/release/cabal-install-2.4.1.0/cabal-install-2.4.1.0.tar.gz'
+  source_sha256 '69bcb2b54a064982412e1587c3c5c1b4fada3344b41b568aab25730034cb21ad'
```
uberhacker: Does this work with all architectures?

vincowl: I don't know (no info and no way to test) but they removed the dedicated packages from their repo.

vincowl: OK...Don't know what's happening, it doesn't install anymore. I am trying to solve this

vincowl: Done. Wrong url to download packages. We have back the various archs.

Note that `cabal` 3.0.0.0 (the executable) is available but `ghc` 8.8.1 includes `Cabal` 3.0.0.0 whereas `ghc` 8.6.5 includes `Cabal` 2.4.0.1.  (Here `Cabal` stands for the `Cabal` library and not the executable). We are forced to use `ghc` 8.6.5 since it is the only version tested with the latest `stack` version. Hope it's clear...

PR #3618, opened 2019-10-29
Upgrade ghc to 8.6.5

vincowl:
version 8.8.1 is available but uncompliant to stack 2.1.3 (latest version) : "ghc above 8.6 has not been tested"

### DIFF
--- a/packages/ghc.rb
+++ b/packages/ghc.rb
@@ -16,7 +16,7 @@ class Ghc < Package
   depends_on 'libtinfo'
 
   def self.build
-    system "stack setup ghc-8.6.5"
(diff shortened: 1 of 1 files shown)


packages/ghc.rb
```diff
@@ -16,7 +16,7 @@ class Ghc < Package
   depends_on 'libtinfo'
 
   def self.build
-    system "stack setup ghc-8.4.2"
+    system "stack setup ghc-8.6.5"
```
uberhacker: `system "stack setup ghc-#{version}"`?  This would make the next updates easier since we would only change the version in one place above.

PR #3627, opened 2019-10-31
Update newlisp from 10.7.1 to 10.7.5

cstrouse:
Readline now works but FFI still doesn't work. Tested on ARM.

### DIFF
--- a/packages/newlisp.rb
+++ b/packages/newlisp.rb
@@ -3,24 +3,16 @@ require 'package'
 class Newlisp < Package
   description 'newLISP is a Lisp-like, general-purpose scripting language.'
   homepage 'http://www.newlisp.org/'
-  version '10.7.1'
-  source_url 'https://github.com/kosh04/newlisp/archive/10.7.1.tar.gz'
-  source_sha256 '87e3ea4bd2d42118af44f2f881302af32a990164fb47088bcad0be6dc88c5c48'
+  version '10.7.5'
+  source_url 'https://github.com/kosh04/newlisp/archive/10.7.5.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/newlisp.rb
```diff
@@ -3,24 +3,16 @@
...
   })
   binary_sha256 ({
-    aarch64: 'e18190cd85ad1bb51af5f64efe67d9262913edd4c2456fd2d27d81d9ad090e39',
-     armv7l: 'e18190cd85ad1bb51af5f64efe67d9262913edd4c2456fd2d27d81d9ad090e39',
-       i686: 'ef5d7215bff75a090e4babcc0ea1b1fb2719fe98d10ee68f6c402185c279811d',
-     x86_64: '6d99a70774a279318953fb48e6c213133426de82175060cb8a03b30b1087662e',
   })
 
-  #depends_on 'readline'
+  depends_on 'readline'
```
uberhacker: OK to delete all dependencies.  This is a core package.

packages/newlisp.rb
```diff
@@ -33,12 +25,12 @@ def self.build
     #
     # Ideally I'll figure out how to get it building properly as follows:
     # system "./configure --prefix=#{CREW_PREFIX} --enable-readline --enable-ffi"
-    system "./configure-alt --prefix=#{CREW_PREFIX}"
-    system "make"
-    system "make", "check"
+    system "./configure-alt --prefix=#{CREW_PREFIX} --enable-readline"
+    system 'make'
+    system 'make', 'check'
```
uberhacker: Move to `def self.check` section?

PR #3634, opened 2019-10-31
Update jansson from 2.11 to 2.12

cstrouse:
Tested on ARM.

### DIFF
--- a/packages/jansson.rb
+++ b/packages/jansson.rb
@@ -3,21 +3,13 @@ require 'package'
 class Jansson < Package
   description 'Jansson is a C library for encoding, decoding and manipulating JSON data.'
   homepage 'http://www.digip.org/jansson/'
-  version '2.11'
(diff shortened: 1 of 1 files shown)


packages/jansson.rb
```diff
@@ -27,6 +19,6 @@ def self.build
   end
```
uberhacker: Shouldn't this also abstract the library directory?  I guess my question is why comment the other options?

cstrouse: They were like that and I'm not familiar with the package so I left it how it was.

uberhacker: Yeah, a lot of these older packages were built before we knew about the abstractions.  I wonder if we need to create a build directory?
```
Dir.mkdir 'build'
Dir.chdir 'build' do
  system 'cmake',
              "-DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX}"
              "-DCMAKE_INSTALL_LIBDIR=#{ARCH_LIB}",
              '..'
end
```

PR #3635, opened 2019-10-31
Update libcheck from 0.11.0 to 0.13.0

cstrouse:
Tested on ARM.

### DIFF
--- a/packages/libcheck.rb
+++ b/packages/libcheck.rb
@@ -4,8 +4,8 @@ class Libcheck < Package
   description 'A unit testing framework for C'
   homepage 'https://github.com/libcheck/check'
   version '0.13.0'
-  source_url 'https://github.com/libcheck/check/archive/0.13.0.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/libcheck.rb
```diff
@@ -3,31 +3,22 @@
 class Libcheck < Package
   description 'A unit testing framework for C'
   homepage 'https://github.com/libcheck/check'
-  version '0.11.0-0'
-  source_url 'https://github.com/libcheck/check/releases/download/0.11.0/check-0.11.0.tar.gz'
-  source_sha256 '24f7a48aae6b74755bcbe964ce8bc7240f6ced2141f8d9cf480bc3b3de0d5616'
+  version '0.13.0'
+  source_url 'https://github.com/libcheck/check/archive/0.13.0.tar.gz'
```
uberhacker: Why the switch from the releases?

cstrouse: I just grabbed the link from the list of updates. I can change it.

uberhacker: Yeah, that is the problem with automated checks.

PR #3636, opened 2019-10-31
Update lz4 from 1.9.1 to 1.9.2

cstrouse:
Tested on ARM.

### DIFF
--- a/packages/lz4.rb
+++ b/packages/lz4.rb
@@ -13,10 +13,10 @@ class Lz4 < Package
   })
 
   def self.build
-    system 'make', 'BUILD_STATIC=no'
+    system 'make', "PREFIX=#{CREW_PREFIX}"
   end
 
   def self.install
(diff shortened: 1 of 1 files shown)


packages/lz4.rb
```diff
@@ -2,29 +2,21 @@
...
   binary_sha256 ({
-    aarch64: '49f2396af749a7b8a81469cdaad3d34e2aff8628b84ee7fed55af464d366544d',
-     armv7l: '49f2396af749a7b8a81469cdaad3d34e2aff8628b84ee7fed55af464d366544d',
-       i686: '5294382301199fe5f1ca42e05cae05ac19dae563d7cb3ba972eb1ae5d55eea5d',
-     x86_64: 'a9828980ed376de59fdac2dcc19dac758fc94105a3e363930e7cc81cabef0a10',
   })
 
   def self.build
-    system "make", "BUILD_STATIC=no"
+    system 'make', 'BUILD_STATIC=no'
```
uberhacker: where is `"PREFIX=#{CREW_PREFIX}"`?

uberhacker: Why not build the static libraries also?

PR #3639, opened 2019-11-02
Update wp_cli from 1.4.1 to 2.3.0

uberhacker:
Tested on all architectures.


cstrouse:
@uberhacker Is the output from the `wp` command not displaying properly for you? Works fine otherwise.

Screenshot 2019-11-03 at 2 27 12 PM

uberhacker:
I don't notice that on any of my Chromebooks.  Did you add anything to your `~/.bashrc`?

cstrouse:
Yeah I have a bunch of additions there. Must be something related to that then.

PR #3641, opened 2019-11-02
Rebuild curl binaries against libressl

uberhacker:
Fixes #3640.  Tested on all architectures.

PR #3642, opened 2019-11-02
Update php packages

uberhacker:
- Update php7 from 7.1.32 to 7.1.33
- Update php7 from 7.2.23 to 7.2.24
- Update php7 from 7.3.10 to 7.3.11
- Update php package

Depends on #3641.  Tested on all architectures.


cstrouse:
Works on ARM.

PR #3644, opened 2019-11-03
Add gambit scheme package

cstrouse:
The Gambit programming system is a full implementation of the Scheme language
which conforms to the R4RS, R5RS and IEEE Scheme standards.

Tested on ARM.

### DIFF
--- a/packages/gambit.rb
+++ b/packages/gambit.rb
@@ -1,0 +1,34 @@
+require 'package'
+
+class Gambit < Package
+  description 'Efficient implementation of the Scheme programming language'
+  homepage 'http://gambitscheme.org/wiki/index.php/Main_Page'
+  version '4.9.3'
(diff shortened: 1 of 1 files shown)


packages/gambit.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.build
+    system './configure',
+             "--prefix=#{CREW_PREFIX}",
+             "--libdir=#{CREW_LIB_PREFIX}/gambit",
```
uberhacker: Why not just `#{CREW_LIB_PREFIX}`?  This will make it easier for other programs to access the libraries.

cstrouse: Every other packager uses this path or `/gambit-c` so I did the same.

Also, I used the FreeBSD port convention for naming the interpreter and compiler commands because one of them defaults to the same name as ghostscript's binary.

PR #3645, opened 2019-11-03
Add ck4up package

cstrouse:
Ck4up is a small command-line utility primarily intended for CRUX ports
maintainers to watch http and ftp sites for updates, but may also be useful for
others.

Tested on ARM. HTTPS is not available on homepage or for the downloads.


uberhacker:
This looks useful.

PR #3656, opened 2019-11-12
[hugo] Upgrade to 0.59.1

vincowl:


uberhacker:
Thank you for tthe update @vincowl!

PR #3657, opened 2019-11-14
[emacs] Add missing libjpeg dependency

cstrouse:
Emacs wouldn't run on a fresh install due to a missing dependency. This adds libjpeg to dependencies resolving the issue. Tested on ARM.


uberhacker:
Good catch although I don't think we need new binaries.  If you install libjpeg does the binary install work?

vincowl:
@uberhacker Installing libjpeg and the emacs binary works here

cstrouse:
Anything holding this up from merging @uberhacker?

uberhacker:
Hey @cstrouse: Just keep the binaries and no need to bump the version.

PR #3663, opened 2019-11-24
Update php packages

uberhacker:
Better version handling.  Tested on all architectures.


skiman6010:
Gosh thank you for these changes. Much better.

PR #3676, opened 2019-12-01
Remove sommelier dependency from zvbi

JL2210:
This doesn't actually do anything; it's just an artifact from old Autotools. Should be safe to just remove.


uberhacker:
What about the binaries?

JL2210:
They don't need to be changed. I checked for X dependencies in the files of the package and couldn't find any.

PR #3692, opened 2019-12-10
[Hugo] Update to 0.60.1

vincowl:


uberhacker:

PR #3701, opened 2019-12-17
[Vim] Update to 8.2.0

vincowl:


uberhacker:

PR #3703, opened 2019-12-17
[Imagemagick 6 / 7] Correct HDRI troubles with gemacs

### LINKED ISSUE #3702: [gemacs] Does not display since imagemagick 6.9.10-77
While running gemacs, following error is displayed :
```
gemacs: error while loading shared libraries: libMagickWand-6.Q16.so.6: cannot open shared object file: No such file or directory
```
Trying to remove / reinstall from sources gemacs or imagemagick doesn't help.

vincowl:
Fixes #3702

### DIFF
--- a/packages/imagemagick6.rb
+++ b/packages/imagemagick6.rb
@@ -64,5 +64,12 @@ class Imagemagick6 < Package
 
   def self.install
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    system "ln -s #{CREW_LIB_PREFIX}/libMagickWand-6.Q16HDRI.so.6 #{CREW_LIB_PREFIX}/libMagickWand-6.Q16.so.6"
+    system "ln -s #{CREW_LIB_PREFIX}/libMagickCore-6.Q16HDRI.so.6 #{CREW_LIB_PREFIX}/libMagickCore-6.Q16.so.6"
+  end
+  
+  def self.remove
+    system "rm #{CREW_LIB_PREFIX}/libMagickWand-6.Q16.so.6"
+    system "rm #{CREW_LIB_PREFIX}/libMagickCore-6.Q16.so.6"
   end
 end
--- a/packages/imagemagick6.rb
+++ b/packages/imagemagick6.rb
@@ -8,14 +8,8 @@ class Imagemagick6 < Package
   source_sha256 '55b3ef6281056c728a25e178434f1e7a2e491cbd99e44dc090a82967a2df6e11'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/imagemagick6-6.9.10-77-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/imagemagick6-6.9.10-77-chromeos-armv7l.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/imagemagick6-6.9.10-77-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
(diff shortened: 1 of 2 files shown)


packages/imagemagick6.rb
```diff
@@ -64,5 +64,12 @@ def self.build
 
   def self.install
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    system "ln -s #{CREW_LIB_PREFIX}/libMagickWand-6.Q16HDRI.so.6 #{CREW_LIB_PREFIX}/libMagickWand-6.Q16.so.6"
+    system "ln -s #{CREW_LIB_PREFIX}/libMagickCore-6.Q16HDRI.so.6 #{CREW_LIB_PREFIX}/libMagickCore-6.Q16.so.6"
```
uberhacker: I believe you will need:
```
FileUtils.ln_s "#{CREW_LIB_PREFIX}/libMagickWand-6.Q16HDRI.so.6", "#{CREW_DEST_LIB_PREFIX}/libMagickWand-6.Q16.so.6"
FileUtils.ln_s "#{CREW_LIB_PREFIX}/libMagickCore-6.Q16HDRI.so.6", "#{CREW_DEST_LIB_PREFIX}/libMagickCore-6.Q16.so.6"
```

packages/imagemagick6.rb
```diff
@@ -64,5 +64,12 @@ def self.build
...
+  end
+  
+  def self.remove
```
uberhacker: I don't believe we have a self.remove method.  Generally, you won't need to remove files if the files have been tracked.  If you apply the suggestion above, you won't need to do this.

packages/imagemagick7.rb
```diff
@@ -64,5 +64,12 @@ def self.build
 
   def self.install
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    system "ln -s #{CREW_LIB_PREFIX}/libMagickWand-7.Q16HDRI.so.6 #{CREW_LIB_PREFIX}/libMagickWand-7.Q16.so.6"
+    system "ln -s #{CREW_LIB_PREFIX}/libMagickCore-7.Q16HDRI.so.6 #{CREW_LIB_PREFIX}/libMagickCore-7.Q16.so.6"
```
uberhacker: See comment in imagemagick6.rb.

packages/imagemagick7.rb
```diff
@@ -64,5 +64,12 @@ def self.build
...
+  end
+
+  def self.remove
```
uberhacker: See comment in imagemagick6.rb.

packages/imagemagick7.rb
```diff
@@ -64,5 +64,12 @@ def self.build
...
+    system "rm #{CREW_LIB_PREFIX}/libMagickWand-7.Q16.so.6"
+    system "rm #{CREW_LIB_PREFIX}/libMagickCore-7.Q16.so.6"
```
uberhacker: We will also need new binaries so please remove the existing `binary_url` and `binary_sha256` sections.

uberhacker:
Thank you for the update @vincowl.  Please see comments.

packages/imagemagick6.rb
vincowl: Could you briefly explain how files are tracked ? I don't really understand this point and that will help me to be more productive next time.

packages/imagemagick7.rb
vincowl: I will apply it to `imagemagick6.rb` too

packages/imagemagick6.rb
uberhacker: When you add files to the `CREW_BREW_DIR` and subdirectories, these files are tracked for installation and therefore are available for removal as well.  If you get a chance, study the `crew` command to get a better idea of how this works.

vincowl: Thanks !

PR #3708, opened 2019-12-19
Add skype package

uberhacker:
Skype is a telecommunications application that specializes in providing video chat and voice calls between computers, tablets, mobile devices, the Xbox One console, and smartwatches via the Internet. Skype also provides instant messaging services. Users may transmit text, video, audio and images.  See https://www.skype.com/en/.  Tested on x86_64.

### DIFF
--- a/packages/skype.rb
+++ b/packages/skype.rb
@@ -1,0 +1,37 @@
+require 'package'
+
+class Skype < Package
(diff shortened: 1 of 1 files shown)


skiman6010:
Now you're just showing off.

PR #3716, opened 2019-12-26
Update python3 from 3.8.0 to 3.8.1

cstrouse:
Tested on ARM.


uberhacker:
Thank you for the update @cstrouse and happy holidays!

PR #3721, opened 2019-12-27
[Gimp] Update to 2.10.14

vincowl:
Needs to update `babl` and `gegl`  which have to be built using `meson` / `ninja` (maybe should be added as build dependencies, or `meson` and `ninja` should be added to the chromebrew core since more and more packages seem to rely upon these two tools).

I had to add an old version of `jsonc` to get `libjson-c.so.3`. Maybe I should rename it ...

### DIFF
--- a/packages/babl.rb
+++ b/packages/babl.rb
@@ -3,33 +3,28 @@ require 'package'
 class Babl < Package
   description 'babl is a dynamic, any to any, pixel format translation library.'
   homepage 'http://gegl.org/babl/'
-  version '0.1.56'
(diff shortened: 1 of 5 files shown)


vincowl:
Too much crap inside...

uberhacker:
Yeah, this is an ambitious update.  I definitely appreciate the effort.  Maybe it makes more sense to update all the dependencies first?  Don't worry about doing it all at once.  Just update a package at a time in separate PRs.

vincowl:
Sorry : didn't see this comment before next PR... Didn't want to separate packages so closely linked.

PR #3722, opened 2019-12-27
[gimp] Update to 2.10.14

vincowl:
Needs to update `babl` and `gegl` which have to be built using `meson` / `ninja` (maybe should be added as build dependencies, or `meson` and `ninja` should be added to the chromebrew core since more and more packages seem to rely upon these two tools).

I had to add an old version of `jsonc` to get `libjson-c.so.3`. Maybe I should rename it ...


uberhacker:
Good suggestion to include `meson` / `ninja` in core.  I cringe everytime I think of adding more to core but this is a light footprint so might be worth it.  There isn't much difference between the current `jsonc` package so maybe it would be wiser to just update to the one that works?  That said, we'll need to test all the other packages that depend on `jsonc` to see if anything breaks.

vincowl:
`jsonc3` is a regression to a previous commit of `jsonc`. So it won't probably be easy to update the current one in a clean way, 

What should I do so with `meson` and `ninja` build dependency ?

uberhacker:
I'm not sure.  Maybe get a listing of all the files of the current `jsonc` package via `crew files jsonc > jsonc-files.txt` and compare that with a newly `meson` built `jsonc` to see what differences exist?

uberhacker:
For now, you can add `depends_on 'meson' => :build` for each package.

vincowl:
> I'm not sure. Maybe get a listing of all the files of the current `jsonc` package via `crew files jsonc > jsonc-files.txt` and compare that with a newly `meson` built `jsonc` to see what differences exist?

Here is the difference :
1c1
< (i) jsonc: JSON-C implements a reference counting object model that allows you to easily construct JSON objects in C, output them as JSON formatted strings and parse JSON formatted strings back into the C representation of JSON objects.
---
> (i) jsonc3: JSON-C implements a reference counting object model that allows you to easily construct JSON objects in C, output them as JSON formatted strings and parse JSON formatted strings back into the C representation of JSON objects.
3c3
< version 0.13.1-20180305
---
> version 0.13-nodoc
22,23c22,23
< /usr/local/lib64/libjson-c.so.4
< /usr/local/lib64/libjson-c.so.4.0.0
---
> /usr/local/lib64/libjson-c.so.3
> /usr/local/lib64/libjson-c.so.3.0.1
26c26

uberhacker:
What happens if you do this?
```
FileUtils.ln_s "#{CREW_LIB_PREFIX}/libjson-c.so.4", "#{CREW_DEST_LIB_PREFIX}/libjson-c.so.3"
```

vincowl:
> FileUtils.ln_s "#{CREW_LIB_PREFIX}/libjson-c.so.4", "#{CREW_DEST_LIB_PREFIX}/libjson-c.so.3"

It works well. I removed jsonc3, added this line to the install function and reinstalled jsonc. How could we be sure that this would be applied when upgrading ? Should we remove binaries or add this line to the postinstall step ?

uberhacker:
Just tack on `-1` to the end of the version and we can create new binaries.  Include the symlink as the last line in the `def self.install` section.

uberhacker:
@vincowl:  We'll need new binaries since the symlink will be added to the file metadata.  Please remove the existing binaries and bump the version.

uberhacker:
Change `version '0.13.1-20180305'` to `version '0.13.1-20180305-1'`.

vincowl:
Oops... Sorry, I thought I did it, but I may have done it on the `jsonc3` package that I finally deleted.

PR #3723, opened 2019-12-27
[codium] Update to 1.41.1

vincowl:


JL2210:
VSCodium has ARM packages now.

vincowl:
@JL2210 Fine ! We have arm and arm64 packages. Which version do we have to maintain ? I'll add it with the next codium version.

JL2210:
I would think arm, as even the arm64 userspace uses arm software.

PR #3726, opened 2019-12-29
[stack] Add armv8 source to 2.1.3 version

### LINKED ISSUE #3512: [pandoc] binaries don't work on ARMv8
At first I thought this was due to stripping behavior since it complains that it doesn't recognize the format but I think the real issue is that they only provide 64-bit binaries and while ARMv8 Chromebooks are 64-bit hardware CrOS is 32-bit.

```
chronos@localhost /usr/local/tmp $ pandoc
bash: /usr/local/bin/pandoc: cannot execute binary file: Exec format error
```

Once we fix `ghc` (see #3511) we can try creating a source build for `pandoc` and build 32-bit binaries for ARMv8.

vincowl:
Fixes #3512 or at least tries to...

### DIFF
--- a/packages/stack.rb
+++ b/packages/stack.rb
@@ -3,13 +3,13 @@ require 'package'
 class Stack < Package
   description 'The Haskell Tool Stack - Stack is a cross-platform program for developing Haskell projects. It is aimed at Haskellers both new and experienced.'
   homepage 'https://docs.haskellstack.org/en/stable/README/'
-  version '2.1.3'
+  version '2.1.3-1'
 
   case ARCH
   when 'aarch64'
     source_url 'https://github.com/commercialhaskell/stack/releases/download/v2.1.3/stack-2.1.3-linux-aarch64.tar.gz'
     source_sha256 '96394b616e29b3bd2479a3712e2c0c375a7cfd748b613f7c6c3186d7ae92977b'
-  when 'armv7l'
+  when 'armv7l', 'armv8'
     source_url 'https://github.com/commercialhaskell/stack/releases/download/v2.1.3/stack-2.1.3-linux-arm.tar.gz'
     source_sha256 '6c8a2100183368d0fe8298bc99260681f10c81838423884be885baaa2e096e78'
   when 'i686'
@@ -21,16 +21,8 @@ class Stack < Package
   end
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/stack-2.1.3-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/stack-2.1.3-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/stack-2.1.3-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/stack-2.1.3-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'fc7da39bc31870079388f77b9410c582103d160fa3bbdfddd7d3a168da9caa7e',
-     armv7l: 'fc7da39bc31870079388f77b9410c582103d160fa3bbdfddd7d3a168da9caa7e',
-       i686: 'd5e1452abdf115ac7452f2508ff8b367300fc18075af918bba5e1b9544e6337f',
-     x86_64: '96a7fd13f475e3465a35b5cd70082abfb2b1cef72474851952b9e2dda048a3c9',
   })
 
   def self.install


packages/stack.rb
```diff
@@ -3,13 +3,13 @@
...
   homepage 'https://docs.haskellstack.org/en/stable/README/'
-  version '2.1.3'
+  version '2.1.3-1'
 
   case ARCH
   when 'aarch64'
     source_url 'https://github.com/commercialhaskell/stack/releases/download/v2.1.3/stack-2.1.3-linux-aarch64.tar.gz'
     source_sha256 '96394b616e29b3bd2479a3712e2c0c375a7cfd748b613f7c6c3186d7ae92977b'
-  when 'armv7l'
+  when 'armv7l', 'armv8'
```
uberhacker: The architecture for 64-bit arm is `aarch64`.

packages/stack.rb
```diff
@@ -21,16 +21,8 @@ class Stack < Package
...
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/stack-2.1.3-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/stack-2.1.3-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/stack-2.1.3-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/stack-2.1.3-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'fc7da39bc31870079388f77b9410c582103d160fa3bbdfddd7d3a168da9caa7e',
-     armv7l: 'fc7da39bc31870079388f77b9410c582103d160fa3bbdfddd7d3a168da9caa7e',
-       i686: 'd5e1452abdf115ac7452f2508ff8b367300fc18075af918bba5e1b9544e6337f',
-     x86_64: '96a7fd13f475e3465a35b5cd70082abfb2b1cef72474851952b9e2dda048a3c9',
```
uberhacker: We shouldn't need new binaries.  The armv7l binary should work on aarch64.

uberhacker:
We are already including `aarch64`.  No need for this change.

uberhacker:
Actually, we can fix this with:
```
-when 'aarch64'
-    source_url 'https://github.com/commercialhaskell/stack/releases/download/v2.1.3/stack-2.1.3-linux-aarch64.tar.gz'
-    source_sha256 '96394b616e29b3bd2479a3712e2c0c375a7cfd748b613f7c6c3186d7ae92977b'
-  when 'armv7l'
+  when 'aarch64', 'armv7l'
```

vincowl:
Ok. So `uname -m` on arm v8 outputs `aarch64` ?

uberhacker:
Yes

packages/stack.rb
cstrouse: Also, this won't work. Google has decided that since all of the aarch64 Chromebooks only have 4GB of memory they will be shipping all devices with 32-bit userlands.

cstrouse:
Once again, aarch64 binaries will *not* run on these Chromebooks.

````bash
chronos@localhost:~ $ uname -a
Linux localhost 4.4.196-16940-geab5e6346fdb #1 SMP PREEMPT Thu Dec 12 22:39:18 PST 2019 aarch64 ARMv8 Processor rev 4 (v8l) GNU/Linux
```

packages/stack.rb
cstrouse: > Actually, we can fix this with:
> 
> ```
> -  when 'aarch64'
> -    source_url 'https://github.com/commercialhaskell/stack/releases/download/v2.1.3/stack-2.1.3-linux-aarch64.tar.gz'
> -    source_sha256 '96394b616e29b3bd2479a3712e2c0c375a7cfd748b613f7c6c3186d7ae92977b'
> -  when 'armv7l'
> +  when 'aarch64', 'armv7l'
> ```

You need to update it to use the 32-bit ARM binaries (https://get.haskellstack.org/stable/linux-arm.tar.gz)

uberhacker:
Also worth noting is `arm` usually means the same as `armv7l` or `armv7` (32 bit, not 64 bit architecture).

cstrouse:
@uberhacker is there an easy way that you can think of to make it possible to have per-platform default compilation arguments? (that is, beyond the -j args) I'd like to override -march and --host for all source builds so that I get working binaries. I think this would be better than having to constantly add conditional arch checks for configure/make just for ARMv8 machines.

Do we know from bintray analytics how many users we have on ARMv8 devices? If it's not many I can just add this to my personal fork rather than complicate the project just for a few users.

uberhacker:
I took a survey awhile back and of all the participants, I think you and I were the only ones who had aarch64 systems.  Most users are on x86_64 architecture.  I would imagine it's probably over 75% of all users.  It's quite shocking how many projects build only for x86_64 architecture now.  If we did abstract crew constants, it would be in `lib/const.rb`.  If you inspect the contents of that file, it will likely give you a better idea how to abstract new constants.

cstrouse:
The code that determines the defaults to use for the -j argument seems sort of messy and like adding more logic to it could be a pain. Should we refactor things so that it's easier for us to build a templated string of configure/make args? It would be nice to be able to clean up the package files and have the ARM-specific overrides happen somewhere else. It probably won't hurt to always specify the include/lib paths and set --march and --host. This way we'd only need things like patching Makefiles that aren't put together properly.

Another option to consider that would be OK as far as I'm concerned is to just adjust your script for building the binaries to have it always set the --host to the ARMv7 target. Or I can maybe just setup something to cross-compile ARMv7 binaries and put them on bintray.

vincowl:
Why not add a "case / when" over `uname -p` inside the  `ARCH / aarch64` one to spot the correct sources, and add a dedicated set up to your CI/CD toolchain providing binaries ?

uberhacker:
Closing due to inactivity.

PR #3727, opened 2019-12-29
[cabal] Correct URL error of i386 on 2.4.1 version

vincowl:


uberhacker:
Good catch @vincowl.  Might as well upgrade.  A new version is available.

i686: https://downloads.haskell.org/~cabal/cabal-install-latest/cabal-install-3.0.0.0-i386-unknown-linux.tar.xz
x86_64: https://downloads.haskell.org/~cabal/cabal-install-latest/cabal-install-3.0.0.0-x86_64-unknown-linux.tar.xz

uberhacker:
While we're at it, we could also update `ghc` since `cabal` needs it with the updated version:
```
https://downloads.haskell.org/~ghc/8.8.1/ghc-8.8.1-src.tar.xz.sig
```

vincowl:
As far as I remember, there was a higher constraint that forbids upgrading these packages greater than actual ones. In october, I commented an issue about that. I'll try to find back what forces this. Greater versions already existed in October.

vincowl:
Cf #3618 : stack forces cabal and ghc versions

vincowl:
It should be levered up in the next stack release since 2.1.3 has been released on July and the ghc 8.8 bug has been solved in august. Cf https://github.com/commercialhaskell/stack/issues/5018

vincowl:
It should be levered up in the next stack release since 2.1.3 has been released on July and the ghc 8.8 bug has been solved in august. Cf https://github.com/commercialhaskell/stack/issues/5018

PR #3737, opened 2020-01-02
Update README.md

skiman6010:
It's that time of the year again @uberhacker! My daily task...


uberhacker:
Daily?  Don't you mean yearly? ;)  Thanks for remembering.

PR #3740, opened 2020-01-04
Add numactl package

uberhacker:
NUMA support for Linux.  See https://github.com/numactl/numactl.  Tested on all architectures.


cstrouse:
Binaries work on ARM and I'm able to compile from source.

PR #3744, opened 2020-01-05
Add dnsmasq package

levinit:
dnsmasq is a lightweight DNS+DHCP+TFTP server.

### DIFF
--- a/packages/dnsmasq.rb
+++ b/packages/dnsmasq.rb
@@ -1,0 +1,27 @@
+require 'package'
+
+class Dnsmasq < Package
+  description 'Lightweight, easy to configure DNS forward ,TFTP server and DHCP server'
+  homepage 'http://www.thekelleys.org.uk/dnsmasq/doc.html'
+  version '2.80'
+  source_url 'http://thekelleys.org.uk/dnsmasq/dnsmasq-2.80.tar.xz'
+  source_sha256 'cdaba2785e92665cf090646cba6f94812760b9d7d8c8d0cfb07ac819377a63bb'
+
+  depends_on 'glibc'
+  depends_on 'gmp'
+  depends_on 'libidn2'
+  depends_on 'nettle'
+  #depends_on 'make' => :build
+
+  def self.build
+    system "make"
+  end
+
+  def self.install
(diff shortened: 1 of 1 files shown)


packages/dnsmasq.rb
```diff
@@ -0,0 +1,27 @@
...
+  source_sha256 'cdaba2785e92665cf090646cba6f94812760b9d7d8c8d0cfb07ac819377a63bb'
+
+  depends_on 'glibc'
+  depends_on 'gmp'
+  depends_on 'libidn2'
+  depends_on 'nettle'
+  #depends_on 'make' => :build
+
+  def self.build
+    system "make"
```
uberhacker: We generally prefer to abstract the prefix with `--prefix=#{CREW_PREFIX}` which defaults to `/usr/local`.  This is normally not needed but we include it mostly to handle cases where the prefix is `/usr` or something else.  We want to make sure the build is not putting the files in a read-only directory.  Also, if this package includes any libraries, we would want to abstract that with `--libdir=#{CREW_LIB_PREFIX}`.  This is to handle cases where the library directory is `/usr/local/lib` or `/usr/local/lib64` depending on the architecture.  Of course, if the package doesn't include any library files this can be ignored.

packages/dnsmasq.rb
```diff
@@ -0,0 +1,27 @@
...
+  homepage 'http://www.thekelleys.org.uk/dnsmasq/doc.html'
+  version '2.80'
+  source_url 'http://thekelleys.org.uk/dnsmasq/dnsmasq-2.80.tar.xz'
+  source_sha256 'cdaba2785e92665cf090646cba6f94812760b9d7d8c8d0cfb07ac819377a63bb'
+
+  depends_on 'glibc'
+  depends_on 'gmp'
+  depends_on 'libidn2'
+  depends_on 'nettle'
+  #depends_on 'make' => :build
```
uberhacker: All these dependencies are included in Chromebrew core so no need to include them here.

uberhacker:
Thank you for the new package @levinit!  Please see comments.

packages/dnsmasq.rb
levinit: thx, I am new .
Next time I will get rid of these lines.

PR #3749, opened 2020-01-05
Make crew list installed run faster

cstrouse:
Generate the list of installed packages in a more efficient manner. Similar refactoring for `crew search` to come soon.


uberhacker:
Nice!  Can you please bump the crew version in `lib/const.rb`?

PR #3755, opened 2020-01-05
Update libheif from 1.3.2 to 1.6.1

cstrouse:
Tested on ARM.


uberhacker:
Do the imagemagick and gimp binaries work with this update?

PR #3756, opened 2020-01-05
Update libseccomp from 2.4.1 to 2.4.2

cstrouse:
Tested on ARM.


uberhacker:
I'm always a bit hesitant to update libraries.  Does this work with the current bind package binaries?

cstrouse:
Actually, until I can script the build/test for these libraries and packages that depend on them I think I'll just close the update PRs.

PR #3759, opened 2020-01-06
Add brackets package

uberhacker:
A modern, open source text editor that understands web design.  See http://brackets.io/.  Tested on x86_64 and i686.  Cannot install on armv7l or aarch64.


skiman6010:
Take a break and go outside. :P

PR #3768, opened 2020-01-17
[php74] Activate zip and gd2 module

vincowl:
Compilation options have changed with php74 to activate zip and gd modules (did not test the other ones)

### DIFF
--- a/packages/php74.rb
+++ b/packages/php74.rb
@@ -3,7 +3,7 @@ require 'package'
 class Php74 < Package
   description 'PHP is a popular general-purpose scripting language that is especially suited to web development.'
   homepage 'http://www.php.net/'
-  version '7.4.1'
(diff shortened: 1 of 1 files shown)


uberhacker:
Thank you for the update @vincowl.  I wish they would make up their mind whether it is `--with-...` or `--enable-...`. LoL

uberhacker:
I'll make sure the next update includes the correct options.  It will probably be fairly soon.  There is a new release almost monthly.  I'll leave this PR open as a reminder.

vincowl:
What is totally dumb is that zip option evolved one way and gd did the opposite.... Absolutely no standardization here...

Fully agree with a latter update : I checked if a new version was already available. Didn't have any idea of the releases scheduling. Thanks for these infos !!

PR #3818, opened 2020-02-20
Update firefox from 70.0.1 to 73.0.1

uberhacker:
Tested on i686 and x86_64.  Still no sound! :(


JL2210:
No sound in Firefox on Crostini either.

PR #3819, opened 2020-02-20
Use arm64 codium on arm64, not armhf

dimkr:
Instead of using 32-bit ARM builds of Codium, this PR makes the package use 64-bit builds, which might perform better.

I haven't tested installing Codium with Chromebrew, but I know that arm64 builds work well.

### DIFF
--- a/packages/codium.rb
+++ b/packages/codium.rb
@@ -5,7 +5,11 @@ class Codium < Package
   homepage 'https://vscodium.com/'
   version '1.42.1'
   case ARCH
-  when 'aarch64', 'armv7l'
+  when 'aarch64'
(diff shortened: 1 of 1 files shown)


cstrouse:
@dimkr None of the ARMv8 Chromebooks can run 64-bit binaries.

uberhacker:
@dimkr: @cstrouse is correct.  In order to execute the 64-bit arm binaries, you would need a 64-bit architecture with 64-bit address space.  Currently ChromeOS only supports a 32-bit address space on aarch64, unfortunately.  That is why we need the armv7l binaries.

uberhacker:
Closing due to a breaking change proposal.

dimkr:
Thanks for the feedback. Just in case someone sees this PR in the future and doesn't understand the source of my confusion: I've been using Codium inside Crostini, and that's an arm64 VM.

Issue #3632, opened 2019-10-31
[hunspell_base]  Unable to spellcheck a file


uberhacker:
```
$ hunspell file.txt
Hunspell has been compiled without Ncurses user interface.
```
Interactive cli spell checker works fine:
```
$ hunspell
Hunspell 1.7.0
exampl 
& exampl 4 0: example, exam pl, exam-pl, exemplar
```

cstrouse:
The project's [README](https://github.com/hunspell/hunspell/blob/master/README.md) gives the following usage info:

After compiling and installing (see INSTALL) you can run the Hunspell spell checker (compiled with user interface) with a Hunspell or Myspell dictionary:

```
hunspell -d en_US text.txt
```

or without interface:

```
hunspell
hunspell -d en_US -l <text.txt
```

It seems that since we've compiled hunspell with the interface enabled you have to pass it a dictionary argument.

uberhacker:
You can also add `export DICTIONARY=en_US` to `~/.bashrc` and then you don't need to add the `-d en_US` option.  Interestingly, this works:
```
hunspell < file.txt
```
But this doesn't:
```
hunspell file.txt
```
According to the help, I shouldn't need the redirect.

vincowl:
Same issue here. Note that I have these infos while compiling hunspell from sources :
```
checking for tparm in -lncursesw... yes
Compiling with curses user interface.
checking curses.h usability... no
checking curses.h presence... no
checking for curses.h... no
checking ncursesw/curses.h usability... yes
checking ncursesw/curses.h presence... yes
checking for ncursesw/curses.h... yes
```
Strange.... Is there a way to "test" ncursesw" for instance ? I don't know how to use this lib.

vincowl:
Maybe an hint here : [https://bugs.gentoo.org/692614](https://bugs.gentoo.org/692614)

vincowl:
Seems like libs are not found even if LD_LIBRARY_PATH is correctly set. Absolutely no idea how this can be... `libtinfo` is installed  but `tgetent` can't find it, same for `libncurses`, `libncursesw`, ... @uberhacker @cstrouse , any idea ?

uberhacker:
@vincowl: No idea but I guess this could be another workaround?  `cat spellcheckthis.txt | hunspell`

cstrouse:
@vincowl Is that using the binaries? source builds? both? I have issues a lot of times with the binaries and symbols not being found at the expected locations.

[...]

PR #3847, opened 2020-02-28
[ncurses] Update compilation option of 6.2 release

vincowl:
Relates to #3632

### DIFF
--- a/packages/ncurses.rb
+++ b/packages/ncurses.rb
@@ -3,56 +3,47 @@ require 'package'
 class Ncurses < Package
   description 'The ncurses (new curses) library is a free software emulation of curses in System V Release 4.0 (SVr4), and more.'
   homepage 'https://www.gnu.org/software/ncurses/'
-  version '6.2'
+  version '6.2-1'
   source_url 'https://ftpmirror.gnu.org/ncurses/ncurses-6.2.tar.gz'
   source_sha256 '30306e0c76e0f9f1f0de987cf1c82a5c21e1ce6568b9227f7da5b71cbea86c9d'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/ncurses-6.2-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/ncurses-6.2-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/ncurses-6.2-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/ncurses-6.2-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'e040d9c065046e107be2712904f0452720724254cc36aa7f04056246967b41b5',
-     armv7l: 'e040d9c065046e107be2712904f0452720724254cc36aa7f04056246967b41b5',
-       i686: '715c277f836a8f8f6c40736c667f63913d6e8ee005b4eb258a7a156a6b6db95a',
-     x86_64: '7ac9b423de77550b4ac2fc82ce4d42fc694352d3911918ba265ca2472d595f38',
   })
 
   def self.build
-    # build libncursesw
-    system "mkdir -p ncursesw_build"
-    Dir.chdir("ncursesw_build") do
-      system "../configure",
-             "--prefix=#{CREW_PREFIX}",
-             "--libdir=#{CREW_LIB_PREFIX}",
-             "--without-normal",
-             "--with-shared",
-             "--with-cxx-shared",
-             "--without-debug",
-             "--enable-pc-files",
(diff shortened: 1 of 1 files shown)


packages/ncurses.rb
```diff
@@ -3,56 +3,47 @@
...
              "--libdir=#{CREW_LIB_PREFIX}",
-             "--without-normal",
+             "--with-normal",
              "--with-shared",
              "--with-cxx-shared",
              "--without-debug",
              "--enable-pc-files",
              "--with-pkg-config-libdir=#{CREW_LIB_PREFIX}/pkgconfig",
-             "--enable-widec"
+             "--without-manpages",
```
uberhacker: Why without manpages?

vincowl: I inverted the 2 builds. This option was not set by me. Indeed, why not actvating man pages... Should we set it to --with-manpages ?

uberhacker: Good question.  It might just build the manpages by default unless `--without-manpages` is provided.

uberhacker:
I have the sinking feeling this is going to break some packages that already depend on ncurses and there are A LOT of packages that have this dependency.  We need to thoroughly test this also since it's a core package and requires the binaries to install Chromebrew itself.

cstrouse:
@uberhacker Yeah, but there's a lot of things where I can't build from source on arm related to ncurses/terminfo. I'll help test things with the containers. If you put together a list of the most critical testing that needs to happen I'll script it for us.

packages/ncurses.rb
cstrouse: It'd be nice to set that globally and have everything build docs or none of them. The Android stuff eats up so much of my space on the Samsung Chromebook Plus v1 that if I build certain packages it runs out of space and starts deleting things. An option to slim things down would be nice.

We could also take the position that Crux does and just get rid of everything that's not essential to the package functioning.

uberhacker:
These are the kinds of things I test when I do it manually:
- Identify the binaries from the package, if any, and attempt to execute (something like <command> -h or <command> -v).
- If there are any libraries, make sure they are installed in the correct #{CREW_LIB_PREFIX} directory.
- If there are any manpages, attempt to run man <command> or info <command>.  If not working, check if they are installed in #{CREW_PREFIX}/man instead of the preferred #{CREW_PREFIX}/share/man.
- Execute crew files <package> to make sure files exist and it looks reasonable (count > 1 and no missing file messages).
- If it is a dependency of other packages, test to make sure the binaries of the other packages still work as expected.
- Uninstall the package and check for any residue left behind.  Removing the package should restore to a state exactly is it was before the package was installed.

cstrouse:
All of the packages that are using `#{CREW_PREFIX}/include/ncursesw` as CPPFLAGS (like `sngrep`) are now broken.

packages/ncurses.rb
vincowl: I would avoid adding manpages to save space but it remains a personal choice. Adding a global option would be great indeed, but out of this package's scope

vincowl:
It's really late here to do it by myself. Maybe try to just keep previous compilation options and juste add --with-termlib. That should be enough to make hunspell run correctly, and maybe avoid other packages crashing. Indeed, documentation states that termlib is a decorrelated part

packages/ncurses.rb
uberhacker: How about a separate package called `ncurses_manpages.rb`?  Other package managers separate out like this for manpages that require a lot of space.

vincowl:
I didn't try to test the new hunspell package with old ncurses (without this PR). Maybe it's enough and would delay this debate on a possible huge impact while having a working hunspell version

uberhacker:
The testing scope of this change is definitely greater than with most other packages.  We always need to be careful when updating libraries because the version changes tend to break things spectacularly.

uberhacker:
@cstrouse: I added the manual testing procedure [here](https://github.com/skycocker/chromebrew/wiki/Testing-procedure).  You can add some scripts here (or links to the tools directory) if you like also.

JL2210:
@uberhacker ncurses and ncursesw are ABI-compatible.

vincowl:
I performed several tests. We could just add the `--with-libterm` argument and keep previous parameters as is. But I don't understand what happened yesterday but it implies to apply the patch of configure.ac (lines to modifiy or add) we get once the hunspell install archive is extracted. Don't know how to apply a patch sinceit implies to download a patch we would host and maintain, no idea to do it easily with sed since it implies lines addition or modifications. We could use a cat configure.ac | tr '\n' '#' | sed .........| tr '#' '\n' ` But this line will be REALLY ugly....

vincowl:
What shall we do here ? Actually I added these modifications to solve hunspell troubles. But other PRs did the trick. Maybe we could close this one and avoid overthinking ?

uberhacker:
Closing but the next time this package has an update, we need to consider what is mentioned in the comments here.

vincowl:
So maybe we should decide on the cases raised here. Maybe in new issues linked to this PR ?

PR #3849, opened 2020-02-29
[hunspell] Change configuration options

### LINKED ISSUE #3632: [hunspell_base]  Unable to spellcheck a file

vincowl:
Fixes #3632

### DIFF
--- a/packages/hunspell_base.rb
+++ b/packages/hunspell_base.rb
@@ -3,28 +3,21 @@ require 'package'
 class Hunspell_base < Package
   description 'Hunspell is a spell checker and morphological analyzer library'
   homepage 'http://hunspell.github.io/'
-  version '1.7.0'
+  version '1.7.0-1'
   source_url 'https://github.com/hunspell/hunspell/archive/v1.7.0.tar.gz'
   source_sha256 'bb27b86eb910a8285407cf3ca33b62643a02798cf2eef468c0a74f6c3ee6bc8a'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/hunspell_base-1.7.0-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/hunspell_base-1.7.0-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/hunspell_base-1.7.0-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/hunspell_base-1.7.0-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: '1827aa434bd7848d03d07ddc2b76c14c052aba16643acd33f494e6d4c3d78016',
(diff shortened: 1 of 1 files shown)


vincowl:
Finally, applying the patch given in #3632 is not needed. That simplifies this PR. Updating ncurses and rebuilding hunspell_base solves the trouble on my machine

vincowl:
Modifications here ar often found in packages, can't we add in a chromebrew's core post-install step a definition of include path including ncsurses folders ?

packages/hunspell_base.rb
```diff
@@ -3,28 +3,21 @@
...
-     x86_64: '83443d8cd8e21f25f346dcd6c7ada5a8a5ef0f01a1045addb43d263ab73d9136',
   })
 
   depends_on 'readline'
   
   def self.build
     system 'autoreconf -vfi'
-    system './configure',
+    system "./configure",
+           "CPPFLAGS=-I#{CREW_PREFIX}/include -I#{CREW_PREFIX}/include/ncursesw -I#{CREW_PREFIX}/include/ncurses",
```
uberhacker: This will not work correctly because of the way ruby parses lines separated by commas.  It might work if you enclose the entire `-I...` list in quotes.  Like this:
```
"CPPFLAGS='-I#{CREW_PREFIX}/include -I#{CREW_PREFIX}/include/ncursesw -I#{CREW_PREFIX}/include/ncurses'"
```

uberhacker:
Should we mention the update in #3847 is required for this PR?

packages/hunspell_base.rb
vincowl: Doesn't work here. Got 
`hunspell_base failed to install: `./configure CPPFLAGS='-I/usr/local/include -I/usr/local/include/ncursesw -I/usr/local/include/ncurses' --prefix=/usr/local --include=/usr/local/include --libdir=/usr/local/lib64 --with-ui --with-readline` exited with 77`
It works with the syntax given in file

vincowl:
> Should we mention the update in #3847 is required for this PR?

I don't understand what's happening. Using #3847 with this PR finally doesn't work. I tried to reinstall everything several times including rebooting my laptop, but it fails if I don't apply the patch in #3632. It works with original version of ncurses 6.2 and the given hunspell_base file. So... 2 solutions :

1. Forget #3847 and use this PR. This simplifies side effects discusses in #3847 but we still don't have libtinfo produced.
2. Second one ; apply #3847 but we have to find a solution to apply the following patch  to configure.ac once the hunspell tar.gz file is extracted

Patch : 
[configure.patch.txt](https://github.com/skycocker/chromebrew/files/4271182/configure.patch.txt)

packages/hunspell_base.rb
uberhacker: I'm not sure how the parser can deal with the spaces and keep all the options for CCPFLAGS in order but I'll try it this way.  I'll fix if I have trouble compiling for all the architectures.

PR #3853, opened 2020-03-01
[hplip] Add missing dependencies and update to 3.20.2

cstrouse:
Tested on x86_64. Source builds on ARM still don't work.

### DIFF
--- a/packages/hplip.rb
+++ b/packages/hplip.rb
@@ -14,7 +14,7 @@ class Hplip < Package
 
   depends_on 'cups'
   depends_on 'dbus'
-  depends_on 'libjpeg'
+  depends_on 'libjpeg_turbo'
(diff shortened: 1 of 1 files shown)


packages/hplip.rb
```diff
@@ -3,17 +3,24 @@
...
 
   binary_url ({
   })
   binary_sha256 ({
   })
 
-  depends_on 'libjpeg'
   depends_on 'cups'
+  depends_on 'dbus'
+  depends_on 'libjpeg'
```
uberhacker: `depends_on 'libjpeg_turbo'`?

cstrouse: We should probably do a bulk update for that in it's own PR. I'll update this and then do a bulk update.

PR #3857, opened 2020-03-01
Update testdisk from 7.1 to 7.2

cstrouse:
Tested on aarch64.

### DIFF
--- a/packages/testdisk.rb
+++ b/packages/testdisk.rb
@@ -3,36 +3,28 @@ require 'package'
 class Testdisk < Package
   description 'TestDisk is powerful free data recovery software!'
   homepage 'https://www.cgsecurity.org/wiki/TestDisk'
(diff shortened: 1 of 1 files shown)


packages/testdisk.rb
```diff
@@ -3,36 +3,28 @@
...
   def self.build
     system "./configure --prefix=#{CREW_PREFIX}"
-    system "make"
+    system 'make'
   end
 
   def self.install
-    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
     system "compressdoc --gzip -9 #{CREW_DEST_PREFIX}/share/man/man8"
```
uberhacker: You can remove this line.  Manpages are compressed automatically when building the package.

PR #3862, opened 2020-03-03
Add janet package

cstrouse:
Janet is a functional and imperative programming language and bytecode
interpreter. It is a modern lisp, but replaces lists with other data
structures that provide better utility and performance (arrays, tables,
structs, tuples).

Tested on ARM.

### DIFF
--- a/packages/janet.rb
+++ b/packages/janet.rb
@@ -1,0 +1,26 @@
+require 'package'
+
+class Janet < Package
+  description 'Janet is a functional and imperative programming language and bytecode interpreter.'
+  homepage 'https://janet-lang.org'
+  version '1.7.0'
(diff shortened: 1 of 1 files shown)


packages/janet.rb
```diff
@@ -0,0 +1,26 @@
...
+  version '1.7.0'
+  source_url 'https://github.com/janet-lang/janet/archive/v1.7.0.tar.gz'
+  source_sha256 '2a119f3a79b209a858864e73ca3efda57ac044df3c89762a31480bbea386d2a3'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'meson'  => :build
```
uberhacker: This is not needed since meson is now in core.

PR #3871, opened 2020-03-08
Add alsa_tools package and fltk dependency

uberhacker:
The Advanced Linux Sound Architecture (ALSA) - tools.  See https://github.com/alsa-project/alsa-tools.  Fast Light Toolkit or FLTK (pronounced "fulltick") is a cross-platform C++ GUI toolkit.  See https://www.fltk.org/.  Tested on all architectures.

### DIFF
--- a/packages/alsa_tools.rb
+++ b/packages/alsa_tools.rb
@@ -1,0 +1,43 @@
+require 'package'
+
+class Alsa_tools < Package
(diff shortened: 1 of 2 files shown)


cstrouse:
I've had an FLTK package sitting around for a while that wasn't finished so this is awesome that you got it working!

PR #3883, opened 2020-03-13
Add mtr package

Tomo59:
Add mtr package

Works properly:
- [x] x86_64

Not tested on other archs.

### DIFF
--- a/packages/mtr.rb
+++ b/packages/mtr.rb
@@ -1,0 +1,34 @@
+require 'package'
+
+class Mtr < Package
+  description "mtr combines the functionality of the 'traceroute' and 'ping' programs in a single network diagnostic tool."
+  homepage 'https://www.bitwizard.nl/mtr/'
+  version '0.93'
+  source_url 'ftp://ftp.bitwizard.nl/mtr/mtr-0.93.tar.gz'
+  source_sha256 '229c673d637bd7dbb96471623785a47e85da0b1944978200c949994c1e6af10d'
+
+  binary_url ({
+    aarch64: '',
+     armv7l: '',
+     x86_64: '',
+       i686: ''
+  })
+  binary_sha256 ({
+    aarch64: '',
+     armv7l: '',
+     x86_64: '',
+       i686: '',
+  })
+
+  depends_on 'ncurses'
+
+  def self.build
+    system "./configure", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}", "--without-gtk",
+        "LDFLAGS=-Wl,-rpath,#{CREW_LIB_PREFIX}"
+    system "make"
+  end
+
+  def self.install
+    system "sudo", "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+end


packages/mtr.rb
```diff
@@ -0,0 +1,34 @@
+require 'package'
+
+class Mtr < Package
+  description "mtr combines the functionality of the 'traceroute' and 'ping' programs in a single network diagnostic tool."
+  homepage 'https://www.bitwizard.nl/mtr/'
+  version '0.93'
+  source_url 'ftp://ftp.bitwizard.nl/mtr/mtr-0.93.tar.gz'
```
cstrouse: Use Github releases instead of FTP for the download. https://github.com/traviscross/mtr/archive/v0.93.tar.gz

packages/mtr.rb
```diff
@@ -0,0 +1,34 @@
...
+       i686: ''
+  })
+  binary_sha256 ({
+    aarch64: '',
+     armv7l: '',
+     x86_64: '',
+       i686: '',
+  })
+
+  depends_on 'ncurses'
```
cstrouse: Remove this. `ncurses` is a core package and is not required to be listed as a dependency.

uberhacker:
Thank you for the new package @Tomo59!

packages/mtr.rb
uberhacker: One exception to using the github releases is if the ftp site version doesn't require any extra steps such as autogen.sh, autoconf or automake.  There is also https://www.bitwizard.nl/mtr/files/ but it doesn't look like the latest release is available.

packages/mtr.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def self.build
+    system "./configure", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}", "--without-gtk",
+        "LDFLAGS=-Wl,-rpath,#{CREW_LIB_PREFIX}"
+    system "make"
+  end
+
+  def self.install
+    system "sudo", "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
```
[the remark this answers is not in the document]
uberhacker: Good catch.  Yeah, we don't want to install anything with sudo since it sets root privileges.

packages/mtr.rb
Tomo59: Indeed, in the github archive there is no configure script. So I let it that way.

packages/mtr.rb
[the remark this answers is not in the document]
Tomo59: It doesn't work without "sudo make install":

make[2]: Entering directory '/usr/local/tmp/crew/mtr-0.93.tar.gz.dir/mtr-0.93'
`setcap cap_net_raw+ep /usr/local/tmp/crew/dest/usr/local/sbin/mtr-packet` \
|| chmod u+s /usr/local/tmp/crew/dest/usr/local/sbin/mtr-packet
/bin/bash: line 1: setcap: command not found

packages/mtr.rb
Tomo59: was removed in 2ef13a0703cace4b607b73fd2768e44d550f5131

Tomo59:
I see that @uberhacker fixed all the problems in 2ef13a0703cace4b607b73fd2768e44d550f5131. Thank you.

Thanks also for the postinstall note.

PR #3884, opened 2020-03-13
Update usbutils from 009 to 012

cstrouse:
Update package and add USB IDs file (only available via HTTP).

Tested on ARM.

### DIFF
--- a/packages/usbutils.rb
+++ b/packages/usbutils.rb
@@ -3,33 +3,27 @@ require 'package'
 class Usbutils < Package
   description 'Tools for examining usb devices'
   homepage 'http://linux-usb.sourceforge.net/'
-  version '009'
-  source_url 'https://www.kernel.org/pub/linux/utils/usb/usbutils/usbutils-009.tar.xz'
-  source_sha256 '8bbff0e54cb5f65a52be4feb9162fc0b022a97eb841b44784f7a89a9ea567160'
+  version '012'
+  source_url 'https://www.kernel.org/pub/linux/utils/usb/usbutils/usbutils-012.tar.xz'
+  source_sha256 '88634625f91840bc1993d2731cc081ee8d3b13d56069a95bdd6ac6ef0e063e46'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/usbutils-009-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/usbutils.rb
```diff
@@ -3,33 +3,27 @@
...
+    system "./configure \
+      --prefix=#{CREW_PREFIX} \
+      --libdir=#{CREW_LIB_PREFIX}"
   end
 
   def self.install
-    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/hwdata/"
+    system "wget http://www.linux-usb.org/usb.ids -O #{CREW_DEST_PREFIX}/share/hwdata/usb.ids"
```
uberhacker: Should we have a checksum on this download to insure file integrity?

cstrouse: It's going to be constantly changing and I didn't see a checksum for it anywhere.

uberhacker: If that is the case, should it go in the `def self.postinstall` section?

packages/usbutils.rb
```diff
@@ -23,7 +23,11 @@ def self.build
 
   def self.install
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+  end
+
+  def self.postinstall
     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/hwdata/"
```
uberhacker: `FileUtils.mkdir_p "#{CREW_PREFIX}/share/hwdata/"`?  Since this is a post install operation, no need to add to the install temp path.

packages/usbutils.rb
```diff
@@ -23,7 +23,11 @@ def self.build
...
     system "wget http://www.linux-usb.org/usb.ids -O #{CREW_DEST_PREFIX}/share/hwdata/usb.ids"
+    puts "It's recommended that you setup a cron job to update this file regularly.".lightblue
```
uberhacker: Would it help to provide an example crontab entry?  Also, might be worth mentioning to install cronie (i.e. `crew install cronie`) to get cron working.

uberhacker:
```
# Example crontab entry to update usb.ids at 3 am daily.
* 3 * * * /usr/local/bin/crew postinstall usbutils
```

uberhacker:
Nice!

PR #3892, opened 2020-03-17
Update v4l_utils from 1.16.3 to 1.18.0

uberhacker:
Tested on all architectures.


cstrouse:
I just installed this from binaries on aarch64 and the binaries dump core when I try to run them. I'll see if a source build yields the same results.

uberhacker:
@cstrouse: Which specific binaries are dumping core?  I just tested this again and it works fine on my aarch64 system.

uberhacker:
@cstrouse: I just updated to the latest ChromeOS release 80.0.3987.158.  Did you update recently?

cstrouse:
Works fine after a fresh crew install. I updated to `80.0.3987.158 (Official Build) (32-bit)` and had a bunch of issues before I reinstalled crew.

PR #3894, opened 2020-03-18
Update install.sh

KellyLSB:
Extracts libraries in the CREW_PREFIX instead of the localized root path.
Additionally notifies the user to add `${CREW_PREFIX}/usr/local/{bin,sbin}` in their path.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -127,12 +127,19 @@ function extract_install () {
     cd "${CREW_DEST_DIR}"
 
     #extract and install
-    echo "Extracting ${1} (this may take a while)..."
-    tar xpf ../"${2}"
-    echo "Installing ${1} (this may take a while)..."
-    tar cpf - ./*/* | (cd "${CREW_PREFIX}"; tar xp --keep-directory-symlink -f -)
-    mv ./dlist "${CREW_CONFIG_PATH}/meta/${1}.directorylist"
-    mv ./filelist "${CREW_CONFIG_PATH}/meta/${1}.filelist"
+    
+    if [ "${3}" == "$(cat ${CREW_CONFIG_PATH}/meta/${1}.current.sha256)" ]; then
+      echo "Current version is already extracted."
+    else 
+      echo "Extracting ${1} (this may take a while)..."
+      tar xpf ../"${2}"
+      echo "Installing ${1} (this may take a while)..."
+      tar cpf - ./*/* | (cd "${CREW_PREFIX}"; tar xp --keep-directory-symlink -f -)
+      mv ./dlist "${CREW_CONFIG_PATH}/meta/${1}.directorylist"
+      mv ./filelist "${CREW_CONFIG_PATH}/meta/${1}.filelist"
+      echo "${3}" > "${CREW_CONFIG_PATH}/meta/${1}.current.sha256"
+      exit 1;;
+    fi
 }
 
 function update_device_json () {
@@ -180,7 +187,7 @@ for i in $(seq 0 $((${#urls[@]} - 1))); do
                         # extract string between first '-' and "-chromeos"
 
   download_check "${name}" "${url}" "${tarfile}" "${sha256}"
-  extract_install "${name}" "${tarfile}"
+  extract_install "${name}" "${tarfile}" "${sha256}"
   update_device_json "${name}" "${version}"
 done
 
--- a/install.sh
+++ b/install.sh
@@ -130,7 +130,7 @@ function extract_install () {
     echo "Extracting ${1} (this may take a while)..."
     tar xpf ../"${2}"
     echo "Installing ${1} (this may take a while)..."
-    tar cpf - ./*/* | (cd /; tar xp --keep-directory-symlink -f -)
+    tar cpf - ./*/* | (cd "${CREW_PREFIX}"; tar xp --keep-directory-symlink -f -)
     mv ./dlist "${CREW_CONFIG_PATH}/meta/${1}.directorylist"
     mv ./filelist "${CREW_CONFIG_PATH}/meta/${1}.filelist"
 }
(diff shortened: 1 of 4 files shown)


KellyLSB:
I fully intend to squash these commits; unless otherwise noted.

```
curl -H 'Cache-Control: no-cache' -Ls https://raw.githubusercontent.com/KellyLSB/chromebrew/patch-2/install.sh | bash
Downloading gcc8...
Verifying gcc8...
gcc8-8.3.0-chromeos-x86_64.tar.xz: OK
Current version is already extracted.
Updating version number of gcc8 in device.json...
Downloading ruby...
Verifying ruby...
ruby-2.5.3-chromeos-x86_64.tar.xz: OK
Current version is already extracted.
Updating version number of ruby in device.json...
Downloading libiconv...
Verifying libiconv...
libiconv-1.15-1-chromeos-x86_64.tar.xz: OK
Current version is already extracted.
Updating version number of libiconv in device.json...
Downloading git...
Verifying git...
git-2.25.0-chromeos-x86_64.tar.xz: OK
Current version is already extracted.
Updating version number of git in device.json...
Downloading libssh2...
Verifying libssh2...
libssh2-1.9.0-chromeos-x86_64.tar.xz: OK
Current version is already extracted.
Updating version number of libssh2 in device.json...
git: error while loading shared libraries: libcharset.so.1: cannot open shared object file: No such file or directory
```

We're getting there.

KellyLSB:
LD_LIBRARY_PATH must be updated
GIT_EXEC_PATH must be updated

install.sh
```diff
@@ -212,7 +220,9 @@ if [[ "${CREW_PREFIX}" != "/usr/local" ]]; then
   echo "you need to run these commands to complete your installation:"
   echo "echo 'export CREW_PREFIX=${CREW_PREFIX}' >> ~/.bashrc"
   echo "echo 'export PATH=\"\${CREW_PREFIX}/bin:\${CREW_PREFIX}/sbin:\${PATH}\"' >> ~/.bashrc"
-  echo "echo 'export LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}' >> ~/.bashrc"
+  echo "echo 'export PATH=\"\${CREW_PREFIX}/usr/local/bin:\${CREW_PREFIX}/usr/local/sbin:\${PATH}\"' >> ~/.bashrc"
+  echo "echo 'export LD_LIBRARY_PATH=\"\${CREW_PREFIX}/lib\${LIB_SUFFIX}:\${LD_LIBRARY_PATH}\"' >> ~/.bashrc"
+  echo "echo 'export LD_LIBRARY_PATH=\"\${CREW_PREFIX}/usr/local/lib\${LIB_SUFFIX}:\${LD_LIBRARY_PATH}\"' >> ~/.bashrc"
```
uberhacker: I see what you are trying to do here but I don't think it's a good idea to pull from multiple library paths because it is possible the better library may not take precedence.  Also, every time this executes, it will continue to append the same path each time and create duplicate paths.  I think a better solution is to make sure each package is installing libraries in the correct `CREW_LIB_PREFIX` location.  Finally, I believe this is only a problem on x86_64 architecture.

cstrouse: I think you're right @uberhacker as I ran into library path issues in my Docker containers on x86_64 but almost never on ARM.

JL2210:
It might be easier without spaces in your SD card's name. I had trouble with installing software in paths with spaces once, too.

KellyLSB:
> It might be easier without spaces in your SD card's name. I had trouble with installing software in paths with spaces once, too.

The space and casing ensures projects account for UTF-8 file paths.

install.sh
KellyLSB: is CREW_LIB_PREFIX a currently assigned environment variable?

I'm in support of this; there is a CREW_LIB_PATH assigned at `${CREW_PREFIX}/lib/crew/` yet that seems tooling specific rather than for the running environment.

**In argument:**
CREW_PREFIX should append the LD_LIBRARY_PATHS in a `crew init` function or something since CREW_PREFIX is a localized `/usr/local` for Chromebrew specifically.

KellyLSB:
All the `${CREW_PREFIX}/usr/local/bin/*` symlinks require rebound into the CREW_PREFIX; currently the symlink destinations are preserved from the initial archive extraction on https://github.com/KellyLSB/chromebrew/blob/patch-2/install.sh#L135

install.sh
KellyLSB: OK so CREW_LIB_PREFIX is in `crew` I had been completely using the `install.sh` for reference thus far. Hmm... yah it could be better if `install.sh` setup the environment the same way that `crew` does.

KellyLSB:
> I don't think you are going to have much success installing this way on Chromebooks. Since most binaries are compiled with the default `CREW_PREFIX` (i.e. `/usr/local`), you will need to compile packages from source with the custom `CREW_PREFIX` path. Hopefully, most (if not all) the packages have abstracted `CREW_PREFIX` to make this possible.

I don't have the available disk space to install Chromebrew to `/usr/local` without hard linking the mount which will conflict with the stock Chrome OS binaries without an overlay. This is actually the biggest blocker for me. Environment agnosticism!

cstrouse:
Have you tried relocating things using `patchelf`? You can also build in a container if you have access to a Docker host. I can send you my CrOS image and/or the scripts for generating your own.

KellyLSB:
> Have you tried relocating things using `patchelf`? You can also build in a container if you have access to a Docker host. I can send you my CrOS image and/or the scripts for generating your own.

Referencing: https://github.com/skycocker/chromebrew/issues/3895

install.sh
uberhacker: Yeah, it was an afterthought to install Chromebrew on a relocated prefix.  It might even make sense to create a separate shell script to install on SD cards rather than trying to make it universal for any type of install.  I think you are going to run into issues with the binaries anyway since they're compiled with the `/usr/local` default prefix.

uberhacker:
@KellyLSB:  I appreciate the effort you put into this.  I would suggest creating a separate script (maybe relocated_install.sh?) and after you get it working for your use case, please submit a new PR.

install.sh
KellyLSB: @uberhacker I had this issue with the `crew` scripts as well.

KellyLSB:
> @KellyLSB: I appreciate the effort you put into this. I would suggest creating a separate script (maybe relocated_install.sh?) and after you get it working for your use case, please submit a new PR.

I will likely continue my current solution and reopen the PR.
The install.sh currently works with the above patches.
I was having more trouble with `crew` itself.

PR #3902, opened 2020-03-25
[codium] Open File button crashes codium

vincowl:
With 1.43.1 version, when trying to open a folder or a file, codium crashes with following message 
```
Failed to load /usr/local/share/icons/gnome/16x16/status/image-missing.png: Unable to load image-loading module: /usr/local/share/VSCodium-linux-x64/../lib64/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-png.so: /usr/local/share/VSCodium-linux-x64/../lib64/gdk-pixbuf-2.0/2.10.0/loaders/libpixbufloader-png.so: cannot open shared object file: No such file or directory (gdk-pixbuf-error-quark, 5)
```
While I could not find out how this path can be set up, I changed the installation path. Any better idea welcomed !

### DIFF
--- a/packages/codium.rb
+++ b/packages/codium.rb
@@ -3,7 +3,7 @@ require 'package'
 class Codium < Package
   description 'VSCodium is Open Source Software Binaries of VSCode with a community-driven default configuration.'
   homepage 'https://vscodium.com/'
-  version '1.43.1'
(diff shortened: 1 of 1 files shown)


uberhacker:
Nice catch @vincowl.

PR #3923, opened 2020-03-25
[xzutils] Update to 5.2.5

vincowl:


packages/xzutils.rb
```diff
@@ -3,21 +3,13 @@
 class Xzutils < Package
   description 'XZ Utils is free general-purpose data compression software with a high compression ratio.'
   homepage 'http://tukaani.org/xz/'
-  version '5.2.3-2'
-  source_url 'http://tukaani.org/xz/xz-5.2.3.tar.gz'
-  source_sha256 '71928b357d0a09a12a4b4c5fafca8c31c19b0e7d3b8ebb19622e96f26dbf28cb'
+  version '5.2.5'
+  source_url 'http://tukaani.org/xz/xz-5.2.5.tar.gz'
```
cstrouse: This is available via HTTPS.

vincowl: You are right. Corrected

PR #3924, opened 2020-03-25
[libvdpau] Clean commented old lines

vincowl:


cstrouse:
@uberhacker Do we really want to bump packages just to cleanup comments?

vincowl:
I updated this package tis afternoon. Binaries have not been generated yet. Maybe we could just set the version to 1.3 instead of 1.3-1 since it is just a correction of a file that has been merged maybe to quickly. No problem to me to let it as it was before this PR

uberhacker:
Yeah, we don't need to generated new binaries just because of comments.  Can you please leave the version the same and keep the binaries @vincowl?

uberhacker:
BTW, any PR is fair game to be merged in too quickly if it's in the queue.  You are supposed to clean things up BEFORE you submit a PR. :)

vincowl:
OK. Let's redo it from scratch. A previous PR today has been made involving libvdpau in order to update it to 1.3. It needed to move from gnutools to meson / ninja build chain. I left the old lines in comment and submitted the PR (cf #3913). PR has been merged. New binaries must be generated then.
Then I noticed those unusefull comments and submitted this PR to clean things up before new binaries generation. If new binaries have been generated since PR #3913, then let this as it was in #3913. If not, why not cleaning things up ?
@uberhacker I know that things should be cleaned up before submitting PR, but isn't it why code reviews are made for too ? ;-)

uberhacker:
If you notice, there are new binaries for this version. :)

vincowl:
....... 😲😱 Damn..... How can I missed this keypoint....  🤔😥 My apologies for this dialogue of the deaf. Closing this PR then...

vincowl:
OK. I think I finally got it... what's wrong with this one and my brain ? ;-)

uberhacker:
Actually, you don't need to bump the version and you have some trailing space after the binaries.

PR #3950, opened 2020-03-27
[lua] Update to 5.3.5-2

vincowl:


packages/lua.rb
```diff
@@ -3,25 +3,17 @@
...
   binary_sha256 ({
-    aarch64: 'e7637922564311eddb467373ce87a1220399ee38ada304d5e66555835959e9da',
-     armv7l: 'e7637922564311eddb467373ce87a1220399ee38ada304d5e66555835959e9da',
-       i686: '7062884ea29cc4b87620ce1cb4228f81cd1427684378f50e4b539aae030d95ca',
-     x86_64: '43b7e0de02fa42251c7881884c48b6593c9c78387356e1b2ea346a4487838671',
   })
 
   def self.build
-    system "make -C src all PLAT=linux SYSCFLAGS=\"-DLUA_USE_LINUX\" SYSLIBS=\"-Wl,-E -ldl -lreadline -lncurses\""
+    system "make -C src all PLAT=linux SYSCFLAGS=\"-DLUA_USE_LINUX  -fPIC\" SYSLIBS=\"-Wl,-E -ldl -lreadline -lncurses -fPIC\""
```
cstrouse: Will you reformat this so it's < 110 chars wide and easier to see which options it's being compiled with at a glance? 
It would probably look better if you change the outside quotes to single and remove the slashes escaping the inner quotes. What do you think?

vincowl: Is it OK like this ? Indeed it looks better and easier to maintain

uberhacker: Would this work?
```
    system "make -C src all \
                  PLAT=linux SYSCFLAGS='-DLUA_USE_LINUX  -fPIC' \
                  SYSLIBS='-Wl,-E -ldl -lreadline -lncurses -fPIC'"
```

uberhacker: Sorry, missed the incoming change. :)

PR #3952, opened 2020-03-27
[ocaml] Update to 4.10.0

vincowl:


cstrouse:
Were you able to compile that on all architectures? That one is usually difficult during updates.

vincowl:
I'd like to but I don't have any means to do this. How can I do this on my chromebook ?

packages/ocaml.rb
```diff
@@ -3,28 +3,20 @@
...
   })
   binary_sha256 ({
-    aarch64: '66f2bf801661c70eb4779f8075da2454b13f4b71cb3efe8680e487c54d539bb2',
-     armv7l: '66f2bf801661c70eb4779f8075da2454b13f4b71cb3efe8680e487c54d539bb2',
-       i686: 'c5942aa70696eee4c242f27a85751c7823f556eb36804f1171fde1ea64f53ba2',
-     x86_64: '3855090573fb7a460c65f2c2705d8f73cf011e79f925a30508397c9ae2376a6f',
   })
 
   depends_on 'buildessential' => :build
   depends_on 'gdbm'
```
uberhacker: These dependencies can be removed since both are in Chromebrew "core".

vincowl: Done

cstrouse:
@vincowl We really need you to at least build it successfully once before opening a PR as a smoketest. If you can't run it on your actual Chromebook for whatever reason, you can use my Docker images on another machine.

vincowl:
@cstrouse every PR is tested on my chromebook, an x86_64 architecture. I don't know how to test it on other architectures. I am not aware that docker images allow cpu architectures emulations. But if it does, I'll be happy to do deploy them on my own server

PR #3953, opened 2020-03-27
[mesa] Update to 20.0.2

vincowl:


packages/mesa.rb
```diff
@@ -27,48 +19,37 @@ class Mesa < Package
   depends_on 'llvm'
 
   def self.patch
-    system "wget -O- https://patchwork.freedesktop.org/patch/257229/raw | patch -Rp1"
+   # system "wget -O- https://patchwork.freedesktop.org/patch/257229/raw | patch -Rp1"
   end
```
uberhacker: It's fine to remove the entire patch section if this patch is no longer needed.

packages/mesa.rb
```diff
@@ -26,49 +18,34 @@ class Mesa < Package
...
   depends_on 'llvm'
 
-  def self.patch
-    system "wget -O- https://patchwork.freedesktop.org/patch/257229/raw | patch -Rp1"
-  end
-
   def self.build
     system "pip3 uninstall -y Mako MarkupSafe || :"
     system "pip3 install --prefix \"#{CREW_PREFIX}\" --root \"#{CREW_DEST_DIR}\" Mako==1.0.7"
     system "pip3 install --prefix \"#{CREW_PREFIX}\" Mako==1.0.7"
```
uberhacker: Are we certain Mako should be pinned to 1.0.7 for this release?

packages/mesa.rb
```diff
@@ -26,49 +18,34 @@ class Mesa < Package
...
-           "--enable-glx",
-           "--enable-dri",
-           "--enable-llvm"
-    system "make"
+    system "meson",
+      "--prefix=#{CREW_PREFIX}",
+      "--libdir=#{CREW_LIB_PREFIX}",
+      "-Dshared-glapi=true",
+      "-Dgallium-drivers=auto",
+      "-Ddri-drivers=auto",
```
uberhacker: Good call.  This is a lot easier than explicitly listing all the drivers.  However, I'm still a bit squeamish about this update since it affects other packages that have this as a dependency.  Please test other packages to make sure they still work as expected after installing this update when you get a chance.  Namely, the following:
```
cairo.rb:  depends_on 'mesa'
jasper.rb:  depends_on 'mesa'
libepoxy.rb:  depends_on 'mesa'
libglu.rb:  depends_on 'mesa'
libsdl.rb:  depends_on 'mesa'
libwebp.rb:  depends_on 'mesa'
sommelier.rb:  depends_on 'mesa'
wxwidgets.rb:  depends_on 'mesa'
xkbcomp.rb:  depends_on 'mesa'
```

packages/mesa.rb
vincowl: Indeed : "Python Mako module - Python Mako module is required. Version 0.8.0 or later should work." Going to change that

vincowl: OK, done. I had to add setuptools dependency to build step

PR #3954, opened 2020-03-27
[glew] Add package with version 2.1.0

vincowl:


packages/glew.rb
```diff
@@ -0,0 +1,27 @@
...
+  binary_sha256 ({
+  })
+
+  depends_on 'mesa'
+
+  def self.build
+#    system "./configure",
+#      "--prefix=#{CREW_PREFIX}",
+#      "--libdir=#{CREW_LIB_PREFIX}"
+    system "make"
```
uberhacker: Probably should include `PREFIX=#{CREW_PREFIX}` and possibly `LIBDIR=#{CREW_LIB_PREFIX}` if the configure script is not available.

vincowl: It worked as is on my machine, but it does too with those modifications. Fine to me then

PR #3955, opened 2020-03-27
[mesa-utils] Add package with version 8.4.0

vincowl:
Provides glxinfo and glxgears

### DIFF
--- a/packages/mesa_utils.rb
+++ b/packages/mesa_utils.rb
@@ -4,8 +4,8 @@ class Mesa_utils< Package
   description 'Open-source implementation of the OpenGL specification - Provides tools such as glxinfo, glxgears'
   homepage 'https://www.mesa3d.org'
   version '8.4.0'
-  source_url 'ftp://ftp.freedesktop.org/pub/mesa/demos/mesa-demos-8.4.0.tar.bz2'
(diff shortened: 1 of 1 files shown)


packages/mesa_utils.rb
```diff
@@ -0,0 +1,28 @@
+require 'package'
+
+class Mesa_utils< Package
+  description 'Open-source implementation of the OpenGL specification - Provides tools such as glxinfo, glxgears'
+  homepage 'https://www.mesa3d.org'
+  version '8.4.0'
+  source_url 'ftp://ftp.freedesktop.org/pub/mesa/demos/mesa-demos-8.4.0.tar.bz2'
```
uberhacker: Better source url? `source_url 'https://mesa.freedesktop.org/archive/demos/mesa-demos-8.4.0.tar.bz2'`

vincowl: FTP sources shall be avoided ? Why ? no problem to change this

vincowl: OK. I linked to the gitlab repository since the `mesa.freedesktop.org` didn't respond. This implies to add an additional step to the build.

uberhacker: The source url works fine for me.  Where are you located?

uberhacker: Ftp is insecure, slow and often unreliable.  We always want to try to use https, if possible, for security and reliability.

PR #3957, opened 2020-03-28
[xkeyboard_config] Update to 2.29

vincowl:


packages/xkeyboard_config.rb
```diff
@@ -27,6 +19,7 @@ class Xkeyboard_config < Package
   depends_on "gettext" => :build
 
   def self.build
+    system "sed -i 's,/usr/bin/python3,/usr/local/bin/python3,' ./rules/compat/map-variants.py"
```
uberhacker: Please change to:
```
system "sed -i 's,/usr/bin/python3,#{CREW_PREFIX}/bin/python3,' ./rules/compat/map-variants.py"
```

PR #3977, opened 2020-03-29
Update python3 from 3.8.1 to 3.8.2

uberhacker:
Tested on all architectures.


cstrouse:
Binaries work for me on ARM and AMD64. Source build worked on ARM.

PR #3986, opened 2020-04-03
Add arping package and libnet dependency

uberhacker:
ARP Ping broadcasts a who-has ARP packet on the network and prints answers. VERY useful when you are trying to pick an unused IP for a net that you don't yet have routing to.  See https://github.com/ThomasHabets/arping.  libnet is a portable framework for low-level network packet construction.  See https://github.com/libnet/libnet.

### DIFF
--- a/packages/arping.rb
+++ b/packages/arping.rb
@@ -1,0 +1,37 @@
+require 'package'
+
+class Arping < Package
+  description 'ARP Ping'
+  homepage 'https://github.com/ThomasHabets/arping'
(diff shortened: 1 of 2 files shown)


cstrouse:
When I try to build it can't find the libpcap headers.

uberhacker:
Try `crew reinstall libpcap -s` and see if that fixes it.  We may need new binaries.

cstrouse:
It works after building from source. Binaries work, too.

PR #4023, opened 2020-04-17
Revert ghostscript package to restore availability on arm

uberhacker:
Gimp depends on ghostscript which was working on arm with the earlier version.


skiman6010:
Nice catch.

PR #4032, opened 2020-04-20
[python3] pip3 removing with crew upgrade

### LINKED ISSUE #4022: Pip3 crashes and not stable
On a fresh install, pip3 command lead to errors quite randomly. 
Install python3, it will install pip3 and pip3 will run normally.
A few minutes later, or after a reboot, pip3 commands leads to :
```
  File "/usr/local/bin/pip3", line 6, in <module>
ModuleNotFoundError: No module named 'pip'
```
I know my issue is sketchy. I am investigating on it, starting with a python3 package downgrade to 3.8.1

vincowl:
Fixes #4022
Not sure in this case if version has to bumped or if binaries have to be regenerated. Please let me know...

### DIFF
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -26,11 +26,6 @@ class Python3 < Package
   depends_on 'krb5'
   depends_on 'libtirpc'
 
-  if ARGV[0] == 'install' || ARGV[0] == 'reinstall' || ARGV[0] == 'upgrade'
-    # Fix ImportError: cannot import name 'PackageFinder'.
-    # See https://stackoverflow.com/questions/59887436/importerror-cannot-import-name-packagefinder.
-    FileUtils.rm_rf Dir.glob("#{CREW_PREFIX}/lib/python3.8/site-packages/pip*")
-  end
 
   def self.build
     # IMPORTANT: Do not build with python3 already installed or pip3 will not be included.
@@ -42,6 +37,11 @@ class Python3 < Package
   end
 
   def self.install
+    if ARGV[0] == 'install' || ARGV[0] == 'reinstall' || ARGV[0] == 'upgrade'
+      # Fix ImportError: cannot import name 'PackageFinder'.
+      # See https://stackoverflow.com/questions/59887436/importerror-cannot-import-name-packagefinder.
+      FileUtils.rm_rf Dir.glob("#{CREW_PREFIX}/lib/python3.8/site-packages/pip*")
+    end
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
 
     # remove static libraries


packages/python3.rb
```diff
@@ -42,6 +37,11 @@ def self.build
   end
 
   def self.install
+    if ARGV[0] == 'install' || ARGV[0] == 'reinstall' || ARGV[0] == 'upgrade'
+      # Fix ImportError: cannot import name 'PackageFinder'.
+      # See https://stackoverflow.com/questions/59887436/importerror-cannot-import-name-packagefinder.
+      FileUtils.rm_rf Dir.glob("#{CREW_PREFIX}/lib/python3.8/site-packages/pip*")
+    end
```
uberhacker, line 44: The problem with putting this here is it only gets executed with a source install.  We want it to execute for binary and source installations.  I will look into this and try to troubleshoot.

vincowl: Right... I forgot binaries installations. Putting it in post-install doesn't work. Maybe we have to tweak the original `if` command to skip the general `upgrade` call

vincowl: Maybe letting it at its original place and changing to 
```
 if (ARGV[0] == 'install' || ARGV[0] == 'reinstall' || ARGV[0] == 'upgrade') && ARGV[1] == 'python3'
```
?

uberhacker: Ah yes.  I see the issue now.  Nice sleuthing @vincowl!  The only problem with this proposal is `crew upgrade` could be run without any package arguments.  You are definitely on the right track.  We want this to execute for python3 package requests only.

vincowl: OK. Normally this would run with `crew install python3`, `crew reinstall python3` and `crew upgrade python3`. Do you need to handle `crew upgrade` calls if python3 needs to be updated ? If so, it will be tricky since we have to know the current version installed and the file's version, or maybe `crew upgrade` finally calls `crew upgrade python3` after having found which packages need upgrade.

uberhacker: What if we just remove all the ARGV logic, get rid of the postinstall section and do this?
```
  def self.install

    # remove static libraries
    system "find #{CREW_DEST_PREFIX} -name 'libpython*.a' -print | xargs -r rm"
    # create symbolic links in lib64 for other applications which use libpython
    unless Dir.exist? "#{CREW_DEST_LIB_PREFIX}"
      FileUtils.mkdir_p "#{CREW_DEST_LIB_PREFIX}"
      system "cd #{CREW_DEST_LIB_PREFIX} && ln -s ../lib/libpython*.so* ."
    end
    puts "Upgrading pip...".lightblue
    # Fix ImportError: cannot import name 'PackageFinder'.
    # See https://stackoverflow.com/questions/59887436/importerror-cannot-import-name-packagefinder.
    FileUtils.rm_rf Dir.glob("#{CREW_PREFIX}/lib/python3.8/site-packages/pip*")
    FileUtils.rm_rf Dir.glob("#{CREW_DEST_PREFIX}/lib/python3.8/site-packages/pip*")
    system "pip3 install pip --root=#{CREW_DEST_DIR} --prefix=#{CREW_PREFIX}"
  end
```
This will make sure all the pip stuff is removed from any previous python3 install and track all the files for the pip upgrade.

uberhacker: Actually, this won't work if python3 is not installed since pip3 won't be available yet.  Let me think some on about how to handle this pip3 upgrade elegantly.

vincowl: An other easy way would be letting the original if block outside methods with it's original statement. Then add a condition based on version comparison between the version property value and the one that can be found in the json installed packages list. If the current version is greater than the installed one, we just have to adjust the logic of install and upgrade scenarios to perform the rm

Not elegant but will work

vincowl: Let's sum up the use cases we have to fire this up:
- reinstall python3
- install python3
- upgrade and new python3 version available
- upgrade python3 and new version available

And the use cases it should not be run :
- upgrade with arguments not containing python3
- postinstall python3

Constraints :
- must not be in build or install method due to binaries installation

It lets two places : postinstall and original place. I don't master the registration stuff so no idea about the postinstall case.

I'll try a logic tomorrow if no better idea  arise till then

uberhacker:
Fixed in #4035.

PR #4035, opened 2020-04-22
Add pre_install function to crew command

### LINKED ISSUE #4022: Pip3 crashes and not stable

uberhacker:
Fixes #4022.  Tested with python3.

### DIFF
--- a/crew
+++ b/crew
@@ -497,7 +497,6 @@ def build_and_preconfigure (target_dir)
     end
 
     @pkg.in_build = true
-    @pkg.preinstall
     @pkg.patch
     @pkg.build
     @pkg.in_build = false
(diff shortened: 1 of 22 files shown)


uberhacker:
@vincowl:  Here is an elegant solution.

vincowl:
Better not piss you off @uberhacker !!! 😆 Modifying `crew` logic is indeed the best we can do. That sounds great. So, content of preinstall will be run in binaries or sources install and upgrade. I didn't have a clear vision of the "under the hood" behavior. Took time now.

I now notice the `#{CREW_OPTIONS}` that would ease the `build` and `install` steps

uberhacker:
Anger is a good motivator. ;)  This has bothered me for awhile now.  I was finally fed up.

PR #4039, opened 2020-04-23
Update curl from 7.67.0 to 7.69.1

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/curl.rb
+++ b/packages/curl.rb
@@ -3,21 +3,21 @@ require 'package'
 class Curl < Package
   description 'Command line tool and library for transferring data with URLs.'
   homepage 'https://curl.haxx.se/'
-  version '7.67.0'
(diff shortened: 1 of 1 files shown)


cstrouse:
Are the extra dependencies really necessary as defaults? If the dependency bloat continues I'll have no choice but to stop using Chromebrew since I don't have any space left (even with a minimal Chromebrew package install set).

uberhacker:
Get a bigger system? ;)  These extra features were added because some users didn't have the functionality they needed.  How much space do these extra features really use?  They look small.

cstrouse:
It's not just those few things but the sum of all the things that are just wasting space. Non-essential features that other packagers leave out should probably only be enabled if the user has chosen to install them or maybe we should try split packages.

PR #4042, opened 2020-04-25
Update binutils from 2.33.1 to 2.34

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/binutils.rb
+++ b/packages/binutils.rb
@@ -2,22 +2,22 @@ require 'package'
 
 class Binutils < Package
   description 'The GNU Binutils are a collection of binary tools.'
-  homepage 'http://www.gnu.org/software/binutils/'
-  version '2.33.1'
-  source_url 'https://ftpmirror.gnu.org/binutils/binutils-2.33.1.tar.lz'
-  source_sha256 'a3239acf5c4efb7e8862a6684174c3b975766cf5f4be6fe25e424ab666ba1735'
+  homepage 'https://www.gnu.org/software/binutils/'
+  version '2.34'
+  source_url 'https://ftpmirror.gnu.org/binutils/binutils-2.34.tar.lz'
+  source_sha256 '5cec79823ef596817aa57a3f470a1afa9827bf14e3583a4e445dc046cc38d29c'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/binutils-2.33.1-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/binutils.rb
```diff
@@ -27,17 +27,16 @@ def self.build
...
-             '--disable-maintainer-mode',
-             '--enable-shared',
-             '--enable-gold',
-             '--enable-ld=default',
-             '--enable-plugins',
-             '--disable-bootstrap',
-             '--enable-64-bit-bfd',
-             '--disable-werror'
+      system "../configure \
+#{CREW_OPTIONS} \
```
cstrouse, line 31: Is this unable to be indented?

packages/binutils.rb
```diff
@@ -3,21 +3,21 @@
 class Binutils < Package
   description 'The GNU Binutils are a collection of binary tools.'
   homepage 'http://www.gnu.org/software/binutils/'
```
cstrouse, line 5: This is available via https.

packages/binutils.rb
```diff
@@ -3,21 +3,21 @@
...
-     armv7l: 'aca082ab8458f1876e7ba3ac2560a396678809780f2b3582739fb0d3ed37c486',
-       i686: '928f7b06a3f4ae28bb514c66a4805c468117489cb9318b81c8677b5ef3b36db6',
-     x86_64: '495c10f7b7b325065d26fe374274a804666d5b433fc69cd1fc3d9dc85bb2742c',
+    aarch64: 'b8194c2d9448796a2c8c5a2e0be669ab32b2a2fb832b4acccff09cb417d0d91f',
+     armv7l: 'b8194c2d9448796a2c8c5a2e0be669ab32b2a2fb832b4acccff09cb417d0d91f',
+       i686: '30a101614d39a99caebd12b3037917a428c618fd4378a04c21469d201c2e9fff',
+     x86_64: 'cc790a554f99715bd5cccefd14f965d885ef2c694afd461007a7f3b37386b91b',
   })
 
   depends_on 'filecmd'
```
cstrouse, line 23: This is a core package.

packages/binutils.rb
uberhacker: Yes, I'll make that change.

packages/binutils.rb
uberhacker: Good catch.

packages/binutils.rb
uberhacker: It is a core package but these dependencies are still needed for core packages because that is what determines all the core dependencies.  Confusing, I know. LoL

PR #4046, opened 2020-04-30
Update ruby from 2.5.3 to 2.7.1

uberhacker:
Tested on all architectures.


cstrouse:
Seems fine on aarch64. Building from source yields a working ruby. Can someone else test on Intel and comment before we merge this?

uberhacker:
Also please test on a fresh install if you can.  A new libressl dependency has been added to install.sh.

cstrouse:
Ruby works fine after a fresh install and upgrading using the binaries in this PR.

PR #4047, opened 2020-04-30
Add acpica package

cstrouse:
acpica is a set of tools for working with ACPI power management on both
32-bit and 64-bit Intel systems. It includes the Intel iASL compiler.

Tested on ARM.

### DIFF
--- a/packages/acpica.rb
+++ b/packages/acpica.rb
@@ -1,0 +1,30 @@
+require 'package'
+
+class Acpica < Package
+  description 'ACPI tools, including Intel ACPI Source Language compiler'
+  homepage 'https://acpica.org/'
+  version '20200326'
+  source_url 'https://acpica.org/sites/acpica/files/acpica-unix2-20200326.tar.gz'
+  source_sha256 '7e0d38d088ff4b2114d08e76201e000091d5a4af40ce7275950759299ba10878'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  def self.build
+    system 'make'
+  end
+
+  def self.install
+    system "install -Dm755 generate/unix/bin/iasl #{CREW_DEST_PREFIX}/bin/iasl"
+    system "install -Dm755 generate/unix/bin/acpixtract #{CREW_DEST_PREFIX}/bin/acpixtract"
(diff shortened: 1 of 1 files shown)


packages/acpica.rb
```diff
@@ -0,0 +1,30 @@
...
+  })
+  binary_sha256 ({
+  })
+
+  def self.build
+    system 'make'
+  end
+
+  def self.install
+    system "install -Dm755 generate/unix/bin/iasl #{CREW_DEST_PREFIX}/bin/iasl"
```
uberhacker, line 20: You could do something like this to avoid having to explicitly include each line and the path to each binary:
```
Dir.chdir 'generate/unix/bin' do
  system "for f in \$(ls | xargs); do install -Dm755 \$f #{CREW_DEST_PREFIX}/bin/\$f; done"
end
```
3 lines instead of 9.

packages/acpica.rb
```diff
@@ -0,0 +1,30 @@
...
+    system "install -Dm755 generate/unix/bin/acpixtract #{CREW_DEST_PREFIX}/bin/acpixtract"
+    system "install -Dm755 generate/unix/bin/acpisrc #{CREW_DEST_PREFIX}/bin/acpisrc"
+    system "install -Dm755 generate/unix/bin/acpinames #{CREW_DEST_PREFIX}/bin/acpinames"
+    system "install -Dm755 generate/unix/bin/acpihelp #{CREW_DEST_PREFIX}/bin/acpihelp"
+    system "install -Dm755 generate/unix/bin/acpiexec #{CREW_DEST_PREFIX}/bin/acpiexec"
+    system "install -Dm755 generate/unix/bin/acpiexamples #{CREW_DEST_PREFIX}/bin/acpiexamples"
+    system "install -Dm755 generate/unix/bin/acpidump #{CREW_DEST_PREFIX}/bin/acpidump"
+    system "install -Dm755 generate/unix/bin/acpibin #{CREW_DEST_PREFIX}/bin/acpibin"
```
uberhacker, line 28: Any manpages to install?

cstrouse: No

uberhacker:
Yes, it's all about the DRY method.  Nice!

PR #4048, opened 2020-05-03
Add pkgsrc package

uberhacker:
pkgsrc is a framework for building third-party software on NetBSD and other UNIX-like systems. 
 See https://github.com/NetBSD/pkgsrc.  Tested on all architectures.  @cstrouse: You should appreciate this.  Takes about 5 minutes to install everything needed to build packages from source.

### DIFF
--- a/packages/pkgsrc.rb
+++ b/packages/pkgsrc.rb
@@ -1,0 +1,131 @@
+require 'package'
+
+class Pkgsrc < Package
+  description 'pkgsrc is a framework for building third-party software on NetBSD and other UNIX-like systems'
+  homepage 'https://github.com/NetBSD/pkgsrc'
+  version '2020Q1'
+  source_url 'https://github.com/NetBSD/pkgsrc/archive/6ba7d2a6ee1388ad515177f5964681f2f6c13f13.tar.gz'
+  source_sha256 'fe2cf98d294cdd298c9191c33c5223860f73e9bed8b9d137f0c32a579ea49803'
+
+  binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/pkgsrc-2020Q1-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/pkgsrc-2020Q1-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/pkgsrc-2020Q1-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/pkgsrc-2020Q1-chromeos-x86_64.tar.xz',
+  })
+  binary_sha256 ({
+    aarch64: '3e5874fdcfe1f9bbc4379c408ea2c306938b8b81aab05a733b0431cd4560c777',
+     armv7l: '3e5874fdcfe1f9bbc4379c408ea2c306938b8b81aab05a733b0431cd4560c777',
+       i686: 'cc7393a04bcb20413e773d265c8855554bb38d4e16bb3b802029a21d9f8a2708',
+     x86_64: 'b461407fd6f86a40d4c5edae331096e57c2d9cdd0fff2e2de6f2d6ab06f53b6a',
+  })
+
+  def self.build
+    system "cat << 'EOF' > pkglocate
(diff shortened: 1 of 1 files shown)


packages/pkgsrc.rb
```diff
@@ -0,0 +1,125 @@
...
+    puts "in #{CREW_PREFIX}/doc/pkgsrc.txt."
+    puts
+    puts "Thank you for using pkgsrc!".lightgreen
+    puts
+    puts "To finish the installation, execute the following:".lightblue
+    puts "echo 'export PATH=\"#{CREW_PREFIX}/pkg/bin:#{CREW_PREFIX}/pkg/sbin:\$PATH\"' >> ~/.bashrc".lightblue
+    puts "echo 'export MANPATH=\"#{CREW_PREFIX}/pkg/man:\$MANPATH\"' >> ~/.bashrc".lightblue
+    puts "source ~/.bashrc".lightblue
+    puts
+    puts "To install packages, execute the following:".lightblue
```
cstrouse, line 116: We should probably add that packages can also be installed using binaries.

`pkg_add ftp://ftp.NetBSD.org/pub/pkgsrc/packages/Net-BSD/i386/3.1_2007Q2/shells/bash-3.2.9.tgz`

packages/pkgsrc.rb
```diff
@@ -0,0 +1,125 @@
...
+    puts
+    puts "will update the vulnerability list every day at 3AM, followed by an audit at"
+    puts "3:09AM. The result of the audit are then emailed to root. On NetBSD this may be"
+    puts "accomplished instead by adding the following line to /etc/daily.conf:"
+    puts
+    puts "fetch_pkg_vulnerabilities=YES"
+    puts
+    puts "to fetch the vulnerability list from the daily security script. The system is"
+    puts "set to audit the packages by default but can be set explicitly, if desired (not"
+    puts "required), by adding the following line to /etc/security.conf:"
```
cstrouse, line 83: This should reference `--sysconfdir=#{CREW_PREFIX}/pkg/etc`

packages/pkgsrc.rb
```diff
@@ -0,0 +1,125 @@
...
+    puts
+    puts "# Download vulnerabilities file"
+    puts "0 3 * * * #{CREW_PREFIX}/pkg/sbin/pkg_admin fetch-pkg-vulnerabilities >/dev/null 2>&1"
+    puts "# Audit the installed packages and email results to root"
+    puts "9 3 * * * #{CREW_PREFIX}/pkg/sbin/pkg_admin audit |mail -s 'Installed package audit result' \\"
+    puts "            root >/dev/null 2>&1"
+    puts
+    puts "will update the vulnerability list every day at 3AM, followed by an audit at"
+    puts "3:09AM. The result of the audit are then emailed to root. On NetBSD this may be"
+    puts "accomplished instead by adding the following line to /etc/daily.conf:"
```
cstrouse, line 77: This should reference `--sysconfdir=#{CREW_PREFIX}/pkg/etc/daily.conf`

cstrouse:
CrOS assumes you're compiling software using the hard float ABI on ARM so we should probably add `-mfloat-abi=hard` to `mk.conf`. Adding an appropriate `-mtune` option wouldn't hurt either.

packages/pkgsrc.rb
uberhacker: We could but this package already has all the source to build and install.  I suppose if you wanted a more recent version than the quarterly release, this would be beneficial.  I was thinking about adding the pkgin command in the future to download and install binary packages precompiled for Chrome OS.  I'd like to contact Benny Siegert to ask him if it could be added to the pkgsrc repo at `ftp://ftp.NetBSD.org/pub/pkgsrc/packages/Linux/<arch>/ChromeOS_2020Q2/All` or something like that.

daemonspudguy:
This is a good solution to the lack of extra software.

uberhacker:
@daemonspudguy:  I'm curious, what software is available in pkgsrc that you would use is not in Chromebrew yet?

cstrouse:
@uberhacker Are you arguing against a package that you've submitted?

uberhacker:
LoL, no but if you think it's useful, please feel free to merge if you like.  Not all the packages install cleanly with the default compiler options so it might be confusing for the casual user.

PR #4051, opened 2020-05-05
Delete build_cros_gcc

JL2210:
This file has rotten over time; it no longer works and it depends on some old files of mine, so remove it.


uberhacker:
This was a good attempt to cross-compile.  The problem with doing this is testing on the actual hardware is still needed and sometimes produces different results than compiling on the bare metal.  I understand not everybody would want or can afford to buy 4 chromebooks to support all the hardware flavors, and unfortunately, that is really the only way to fully support all the architectures.

PR #4052, opened 2020-05-05
Remove xz-java binary

JL2210:
I think these .jars were only intended to be demos anyway and the real use for this package was the library that it provides. I don't want this depending on any of my files, so remove it.


uberhacker:
Yeah, we already have xzutils which is basically the same thing.

PR #4053, opened 2020-05-05
Update fskit.rb

JL2210:
Once again removing references to my repositories; in theory this shouldn't change anything but it should be tested to see if it builds.


uberhacker:
Does it build for you?

JL2210:
I don't know, I don't have an accessible developer mode device anymore. I
just don't want to break packages by cleaning up my old repos.

PR #4055, opened 2020-05-05
Delete checkinstall.rb

JL2210:
This wasn't very useful; my patch wasn't exactly the greatest and it broke things inside the package. I doubt that it works, so just remove it. Not sure the policy to avoid breaking current installations, but it's unlikely that anybody installed it besides me.


uberhacker:
Yeah, this is a lot like dpkg.  Why do we need a package for a package manager when Chromebrew is already a package manager? LoL

PR #4059, opened 2020-05-07
[ripgrep] Update to 12.0.1

vincowl:


uberhacker:
Whoa!  What's up with the huge version jump?  Thank you for the update @vincowl!

vincowl:
Absolutely no idea ! It's a dependency to doom emacs. Still trying to find a dev environment which allows me to handle different languages and docs writing, and that would not be a pain in the arse as I disconnect the Chromebook from its dock to work outside. I have troubles with codium with several plugins. Graphical ides crashes as the Chromebook enters sleep mode due to sommelier's bug. So it finally reduces the choice to vim or emacs and their "forks" like neovim, spacevim, spacemacs and doomemacs. Nothing better than tests to choose.

uberhacker:
Sounds like good packages to add to Chromebrew.

vincowl:
Sure. I can package these tools. How should we process with doomemacs for instance ? It acts like an upper layer to emacs so maybe create a doomemacs package depending on gemacs which depends on emacs ? But doomemacs has its own upgrade system and seems to have left the GitHub release workflow for a while.... This would surely lead to a package that will be a pain to maintain.

Same thing with eclipse that can be installed as a simple copy/paste as long as jdk8 is installed. Easy to package, but this implies to maintain things we have no way to control releases and which may break at any moment. And we have no added value here.

What's the philosophy here ?

uberhacker:
All good questions.  You would have to experiment and find out what works best.  Keep in mind that most Chromebooks have lower end resources so the package should be minimal and efficient, if possible.

vincowl:
In fact I already made my opinion : no added value with a package, no release, it has its own upgrade system, so it should live aside chromebrew, not inside.  So, no package. Maybe more a word in a wiki article.

Codium or filezilla are border line to me for instance. They have their own update mechanism, so no need of a chromebrew one. But, chromebrew has an added value since it's not so easy to install and there are releases provided by their developers. So we can be sure our package won't break easily, aside dependencies upgrades troubles.

PR #4068, opened 2020-05-09
Update brackets from 1.14.1 to 1.14.1-1

uberhacker:
Tested on x86_64.  Fixes issue where aarch64 and armv7l users install all the dependencies only to be told their architecture is not supported.  Also simplified the logic.


cstrouse:
I can confirm the ARM stuff is fixed but don't have an x86_64 device with GUI environment to test otherwise.

PR #4069, opened 2020-05-10
Add check for package compatibility

uberhacker:
This PR will be difficult to review since it touches all packages.  The idea here is to provide package compatibility for specific architectures since not all packages are compatible with all architectures.  It also solves the issue of installing all dependencies for a package only to be told later the package is not available.  I took this time to clean up some packages as well since if I need to edit each file, might as well, right? :)


cstrouse:
Sounds like a good time to get the CI setup.

uberhacker:
Should I just close this @cstrouse?

uberhacker:
@cstrouse: The longer this sits here, the more times I have to update due to resolve conflicts when new PRs are merged.

vincowl:
Great job !!!! It's awesome ! How should we handle this new property when updating packages ? Do we leave it as is and let you set it up regarding the binaries generation's results ? And when adding a new package ?

@cstrouse Can you tell us more about the CI setup you are talking about ? Would this be available to us or duplicable on our own machines ?

PR #4070, opened 2020-05-13
[phpactor] Create a new package

vincowl:
This package depends on a php version >=7.3 . How can I translate it in the dependencies ? I chose to depend on `php73`

### DIFF
--- a/packages/phpactor.rb
+++ b/packages/phpactor.rb
@@ -15,7 +15,7 @@ class Phpactor < Package
   depends_on 'composer'
   php73 = `crew list installed | grep "php73" 2> /dev/null`.chomp
   php74 = `crew list installed | grep "php74" 2> /dev/null`.chomp
-  depends_on 'php73' unless "#{php73}" != "" and "#{php74}" != ""
+  depends_on 'php73' unless "#{php73}" != "" or "#{php74}" != "" or not(File.exists? "#{CREW_PREFIX}/bin/php")
 
   def self.install
     system "composer install"
--- a/packages/phpactor.rb
+++ b/packages/phpactor.rb
@@ -17,12 +17,14 @@ class Phpactor < Package
   phpMinimalVersion = "7.3"
   phpDefaultVersion = "7.4"
   phpver = `php -v 2> /dev/null | head -1 | cut -d' ' -f2`.chomp
-  if "#{phpver}" == ""
-    depends_on "php"+"#{phpDefaultVersion}".split(".").take(2).join
-  elsif Gem::Version.new("#{phpver}") >= Gem::Version.new("#{phpMinimalVersion}")
-    depends_on "php"+"#{phpver}".split(".").take(2).join
-  else
-    abort "Minimum version of php73 required".lightred 
+  if (ARGV[0] == 'install' || ARGV[0] == 'reinstall' || ARGV[0] == 'upgrade') && ARGV[1] == 'phpactor'
+    if "#{phpver}" == ""
+      depends_on "php"+"#{phpDefaultVersion}".split(".").take(2).join
+    elsif Gem::Version.new("#{phpver}") >= Gem::Version.new("#{phpMinimalVersion}")
+      depends_on "php"+"#{phpver}".split(".").take(2).join
+    else
+      abort "Minimum version of php73 required".lightred 
+    end
   end
 
 
--- a/packages/phpactor.rb
+++ b/packages/phpactor.rb
@@ -13,15 +13,18 @@ class Phpactor < Package
   })
 
   depends_on 'composer'
-  php73 = `crew list installed | grep "php73" 2> /dev/null`.chomp
-  php74 = `crew list installed | grep "php74" 2> /dev/null`.chomp
-  if "#{php74}" != ""  or not(File.exists? "#{CREW_PREFIX}/bin/php") 
-    depends_on 'php74'
-  elsif "#{php73}" != ""
-    depends_on 'php73'
+  
+  phpMinimalVersion = "7.3"
+  phpDefaultVersion = "7.4"
+  phpver = `php -v 2> /dev/null | head -1 | cut -d' ' -f2`.chomp
+  if "#{phpver}" == ""
+    depends_on "php"+"#{phpDefaultVersion}".split(".").take(2).join
+  elsif Gem::Version.new("#{phpver}") >= Gem::Version.new("#{phpMinimalVersion}")
+    depends_on "php"+"#{phpver}".split(".").take(2).join
   else
     abort "Minimum version of php73 required".lightred 
   end
+
 
   def self.install
     system "composer install"
--- a/packages/phpactor.rb
+++ b/packages/phpactor.rb
@@ -15,7 +15,13 @@ class Phpactor < Package
   depends_on 'composer'
   php73 = `crew list installed | grep "php73" 2> /dev/null`.chomp
   php74 = `crew list installed | grep "php74" 2> /dev/null`.chomp
-  depends_on 'php73' unless "#{php73}" != "" or "#{php74}" != "" or not(File.exists? "#{CREW_PREFIX}/bin/php")
+  if "#{php74}" != ""  or not(File.exists? "#{CREW_PREFIX}/bin/php") 
+    depends_on 'php74'
+  elsif "#{php73}" != ""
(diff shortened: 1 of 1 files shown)


packages/phpactor.rb
```diff
@@ -0,0 +1,32 @@
...
+  source_url 'https://github.com/phpactor/phpactor/archive/0.15.0.tar.gz'
+  source_sha256 '02f63e77aa435874a3e8ef11a293eb24a7ee259317dfaae3f058d8bc24e39e24'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'composer'
+  depends_on 'php73'
```
uberhacker, line 16: Probably don't need the php dependency unless that is a minimum version requirement since composer already has a php requirement.

packages/phpactor.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def self.preinstall
+  end
+
+  def self.build
+  end
```
uberhacker, line 22: Why the empty defs?

packages/phpactor.rb
```diff
@@ -0,0 +1,32 @@
...
+
+
+  def self.install
+    system "composer install"
+    FileUtils.mkdir_p "#{CREW_PREFIX}/share/phpactor"
+    FileUtils.cp_r '.', "#{CREW_PREFIX}/share/phpactor/."
+    FileUtils.ln_s "#{CREW_PREFIX}/share/phpactor/bin/phpactor", "#{CREW_PREFIX}/bin/phpactor"
```
uberhacker, line 29: This should be:
```
    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/phpactor"
    FileUtils.cp_r '.', "#{CREW_DEST_PREFIX}/share/phpactor/."
    FileUtils.ln_s "#{CREW_PREFIX}/share/phpactor/bin/phpactor", "#{CREW_DEST_PREFIX}/bin/phpactor"
 ```
Otherwise, crew will not be able to track the install.  Run `crew files phpactor` afterwards and you will know what I mean.  Also, probably need to copy all the files in `$HOME/.config/composer` to `#{CREW_DEST_HOME}/.config/composer` as well due to the `composer install`.

**Comment by uberhacker** [comment]
Thank you for the new package @vincowl!  Please see comments.

File: packages/phpactor.rb (hunk 3 above)
Comment (vincowl) [reply packages/phpactor.rb:29]: Right !! I'll check what's in $HOME

File: packages/phpactor.rb (hunk 1 above)
Comment (vincowl) [reply packages/phpactor.rb:16]: Indeed, as said in the PR description, it requires php >=7.3. I don't know how to express a minimum version so I chose a `php73` dependency

Comment (uberhacker) [reply packages/phpactor.rb:16]: This could be a bit tricky then.  If somebody already has another version of php installed, it might bypass the requirement altogether.  We would have to test it but I imagine if the requirement isn't met, some message should be output and the install should abort if the minimum requirement isn't met.

File: packages/phpactor.rb (hunk 3 above)
Comment (uberhacker) [reply packages/phpactor.rb:29]: Come to think of it, you likely don't need to copy anything from `$HOME/.config/composer` since this isn't a global install...I think.

Comment (vincowl) [reply packages/phpactor.rb]: Done. Indeed, nothing created in `$HOME/.config/composer`

File: packages/phpactor.rb (hunk 1 above)
Comment (vincowl) [reply packages/phpactor.rb:16]: I removed `phpactor` and my `php73` version and installed `php72` then ran `crew install phpactor`. It tryed to install `php73` but failed and stopped `phpactor` installation. Then I removed `php72`.

1. First problem : when installing `phpactor` it wanted to install `php74` and `php73`... Strange...
2. Second problem : installation fails if I install `php74` and then `phpactor`. `php73` is a minimal requirement, this would imply to create a new syntax in `crew` such as `depends_on php => "7.3"` as done in `chromebrew` : 
> Some Requirements can also take a string specifying their minimum version that the formula depends on.

Comment (uberhacker) [reply packages/phpactor.rb:16]: 1. is probably happening since composer has a php74 dependency if php is not installed.  Then the phpactor has a dependency of php73 which would make sense.  We might be able to do something like below:
```
php73 = `crew search php73 2> /dev/null`.chomp
php74 = `crew search php74 2> /dev/null`.chomp
depends_on 'php73' unless "#{php73}" != "" or "#{php74}" != ""
```
If any other version of php is installed, it will output the message php x.x is already installed.

Comment (vincowl) [reply packages/phpactor.rb:16]: This doesn't work : if I install `php72`, `phpactor` is installed but fails while running since `php73` minimal dependency is not fullfilled. Maybe try something based on checking if `php73` is in the installed packages list using `grep` and `crew list installed`. I am going to check that

Comment (vincowl) [reply packages/phpactor.rb]:
```
  php73 = `crew list installed | grep "php73" 2> /dev/null`.chomp
  php74 = `crew list installed | grep "php74" 2> /dev/null`.chomp
  depends_on 'php73' unless "#{php73}" != "" and "#{php74}" != ""
```
This works. But :
1- if php < 7.3 is installed  --> Fails : normal
2- if php73 is installed --> Success
3- if php74 is installed --> Success
4- if no php is installed --> Fails.

4 is a problematic case : since `composer` requires `php74` and `phpactor` by default depends on `php73`. So, crew will ask to install `php73` AND `php74` which fails. Maybe we should handle this case in the logic to depend on `php74` if no php is installed ?

File: packages/phpactor.rb [hunk 4]
```diff
@@ -13,8 +13,9 @@ class Phpactor < Package
   })
 
   depends_on 'composer'
-  depends_on 'php73'
-
+  php73 = `crew list installed | grep "php73" 2> /dev/null`.chomp
+  php74 = `crew list installed | grep "php74" 2> /dev/null`.chomp
+  depends_on 'php73' unless "#{php73}" != "" and "#{php74}" != ""
```
uberhacker, line 18: Does this do anything funky with other operations such as `crew search` or `crew update`?  Also, if php72 or lower is installed, does the install abort with a message something like "Minimum version of php73 required".

packages/phpactor.rb
vincowl: OK :  #cfc5566 solves this logic problems even if no php version is installed.

packages/phpactor.rb
vincowl: `crew search` interaction :
``` bash
$ crew search php
php: PHP is a popular general-purpose scripting language that is especially suited to web development.
php5: PHP is a popular general-purpose scripting language that is especially suited to web development.
php70: PHP is a popular general-purpose scripting language that is especially suited to web development.
php71: PHP is a popular general-purpose scripting language that is especially suited to web development.
php72: PHP is a popular general-purpose scripting language that is especially suited to web development.
(i) php73: PHP is a popular general-purpose scripting language that is especially suited to web development.
php74: PHP is a popular general-purpose scripting language that is especially suited to web development.
(i) phpactor: PHP completion, refactoring and introspection tool.
$ crew search php73
(i) php73: PHP is a popular general-purpose scripting language that is especially suited to web development.
$ crew search php74
php74: PHP is a popular general-purpose scripting language that is especially suited to web development.
$ crew search phpactor
(i) phpactor: PHP completion, refactoring and introspection tool.
```

vincowl:
```
$ crew update
HEAD is now at cd22336e Update ngrok from 2.3.15 to 2.3.35 (#4076)

Your software is up to date.
```

vincowl: Displaying the error message si something I must handle. Working on it

vincowl: See f672c86 : it should solve all of this

packages/phpactor.rb
vincowl: I changed the logic to display the error message and abort installation

vincowl:
> Let me think about this a bit. It might be better if we put this logic in the `crew` command and use something like:
> 
> ```
> depends_on 'php' >= 7.3
> ```
> 
> That way, we wouldn't need to duplicate this logic on other packages.

Fully agree. That's exactly what I thought in previous comments, saying it would imply to define a new crew syntax. Shall I try to implement it or do you handle this @uberhacker ?

uberhacker:
Go ahead and give it a shot if you want.  I'm still thinking about the best way to do this globally with other use cases.

PR #4079, opened 2020-05-20
php packages : error messages color

### LINKED ISSUE #4078: Message "php x.x is already installed" is not displayed as an error
While trying to install a php version, if any other version of php is installed, it will output the message php x.x is already installed. This message is displayed in green instead as an error. This leads to visual misunderstanding while installing packages with newer php dependencies

vincowl:
Fixes #4078


uberhacker:
I can see how this could create confusion.  Fair enough.

PR #4082, opened 2020-05-22
motif.rb and nedit.rb added to packages

dji-transpire:
Adds motif and nedit to packages

## Description
provide nedit and motif

## Addtional information
tested on a Lenovo Duet
Works properly:
- [NA] x86_64
- [X] aarch64 (reasons why it doesn't)

---

## That's it

### DIFF
--- a/packages/motif.rb
+++ b/packages/motif.rb
@@ -1,0 +1,38 @@
+require 'package'
+
+class Motif < Package
+  description 'Open Motif'
+  homepage 'https://github.com/libass/libass'
+  version '2.3.8'
+  source_url 'https://downloads.sourceforge.net/project/motif/Motif%202.3.8%20Source%20Code/motif-2.3.8.tar.gz'
+  source_sha256 '859b723666eeac7df018209d66045c9853b50b4218cecadb794e2359619ebce7' 
+
+  # binary_url ({
+  #    aarch64: 'file:///home/chronos/user/Downloads/motif/motif-2.3.8-chromeos-aarch64.tar.xz'
+  # })
+  # binary_sha256 ({
+  #    aarch64: 'b9a8d12260d88c0d1d40e64db579e32f7744a0e1f391c5a054a162aca392d9e9'
+  # })
+
+  #depends_on 'automake' => :build
+  #depends_on 'autoconf' => :build
+  #depends_on 'libtool' => :build
+  #depends_on 'fribidi'
+  #depends_on 'fontconfig'  
+
+  def self.build
+    system "autoconf"
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+
+  def self.check
+    system "make", "check"
+  end
+
+end
+
--- a/packages/nedit.rb
+++ b/packages/nedit.rb
@@ -1,0 +1,40 @@
+require 'package'
+
+class Nedit < Package
+  description 'Nedit 5.7'
+  homepage 'https://sourceforge.net/projects/nedit/'
+  version '5.7'
+  source_url 'https://downloads.sourceforge.net/project/nedit/nedit-source/nedit-5.7-src.tar.gz'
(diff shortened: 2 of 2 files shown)


packages/motif.rb
```diff
@@ -0,0 +1,38 @@
...
+  # })
+  # binary_sha256 ({
+  #    aarch64: 'b9a8d12260d88c0d1d40e64db579e32f7744a0e1f391c5a054a162aca392d9e9'
+  # })
+
+  #depends_on 'automake' => :build
+  #depends_on 'autoconf' => :build
+  #depends_on 'libtool' => :build
+  #depends_on 'fribidi'
+  #depends_on 'fontconfig'
```
uberhacker, line 21: If you need fribidi and fontconfig dependencies, these lines will need to be uncommented.

packages/motif.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  #depends_on 'automake' => :build
+  #depends_on 'autoconf' => :build
+  #depends_on 'libtool' => :build
+  #depends_on 'fribidi'
+  #depends_on 'fontconfig'  
+
+  def self.build
+    system "autoconf"
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
```
uberhacker, line 25: This line could be simplified to `system "./configure #{CREW_OPTIONS}"`

packages/nedit.rb
```diff
@@ -0,0 +1,40 @@
...
+  #    aarch64: 'file:///home/chronos/user/Downloads/nedit/nedit-5.7-chromeos-aarch64.tar.xz'
+  # })
+  # binary_sha256 ({
+  #    aarch64: '8c60d9a44afafc6e3be4fb4084ec42c3f36910a466bcc59cfe4a5f199c1aef7b'
+  # })
+
+  #depends_on 'automake' => :build
+  #depends_on 'autoconf' => :build
+  depends_on 'libxfont2'
+  depends_on 'sed'
```
uberhacker, line 20: The sed dependency line is not needed since that is included in "core".

packages/nedit.rb
```diff
@@ -0,0 +1,40 @@
...
+  depends_on 'motif'
+
+  def self.build
+    #system "autoconf"
+    #system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
+    system "sed -i 's/$@/linux/g' Makefile"
```
uberhacker, line 26: This line could go in the patch section like below:
```
  def self.patch
    system "sed -i 's/$@/linux/g' Makefile"
  end

  def self.build
    system "make linux"
  end
```

packages/nedit.rb
```diff
@@ -0,0 +1,40 @@
...
+    system "make linux"
+  end
+
+  def self.install
+    #system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "cp ./source/nedit /usr/local/bin"
```
uberhacker, line 32: This line will not install correctly since it doesn't go into the `#{CREW_DEST_DIR}` tracking location.  Please change to:
```
    system "install -Dm755 ./source/nedit #{CREW_DEST_PREFIX}/bin/nedit"
```

uberhacker: Any manpage included in the build?

uberhacker:
Thank you for the new packages @dji-transpire!  Please see comments.

packages/motif.rb
```diff
@@ -0,0 +1,38 @@
+require 'package'
+
+class Motif < Package
+  description 'Open Motif'
+  homepage 'https://github.com/libass/libass'
```
uberhacker, line 5: Wrong homepage?

packages/nedit.rb
```diff
@@ -0,0 +1,40 @@
+require 'package'
+
+class Nedit < Package
+  description 'Nedit 5.7'
```
uberhacker, line 4: Can we have a better description?

packages/motif.rb
```diff
@@ -0,0 +1,38 @@
+require 'package'
+
+class Motif < Package
+  description 'Open Motif'
```
uberhacker, line 4: It would be nice to include some more details to the description.

PR #4083, opened 2020-05-24
Change graphviz to cmake build

uberhacker:
Tested on all architectures.  The dot command works in this build @vincowl.  Feel free to test.


vincowl:
Works like a charm using binaries or compiling from sources ! Thanks !

PR #4088, opened 2020-05-25
Update terraform from 0.11.1 to 0.12.25

cstrouse:
Switch to official project binaries.

Tested on x86_64 and ARM.


uberhacker:
Nice!

PR #4089, opened 2020-05-27
Update stack to actually work

uberhacker:
The previous stack was unable to execute in `~/.stack`.

### DIFF
--- a/packages/stack.rb
+++ b/packages/stack.rb
@@ -3,41 +3,21 @@ require 'package'
 class Stack < Package
   description 'The Haskell Tool Stack - Stack is a cross-platform program for developing Haskell projects. It is aimed at Haskellers both new and experienced.'
   homepage 'https://docs.haskellstack.org/en/stable/README/'
-  version '2.1.3'
+  version '2.1.3-1'
 
   case ARCH
-  when 'aarch64'
-    source_url 'https://github.com/commercialhaskell/stack/releases/download/v2.1.3/stack-2.1.3-linux-aarch64.tar.gz'
-    source_sha256 '96394b616e29b3bd2479a3712e2c0c375a7cfd748b613f7c6c3186d7ae92977b'
-  when 'armv7l'
-    source_url 'https://github.com/commercialhaskell/stack/releases/download/v2.1.3/stack-2.1.3-linux-arm.tar.gz'
-    source_sha256 '6c8a2100183368d0fe8298bc99260681f10c81838423884be885baaa2e096e78'
-  when 'i686'
-    source_url 'https://github.com/commercialhaskell/stack/releases/download/v2.1.3/stack-2.1.3-linux-i386.tar.gz'
-    source_sha256 '4acd97f4c91b1d1333c8d84ea38f690f0b5ac5224ba591f8cdd1b9d0e8973807'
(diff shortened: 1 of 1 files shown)


cstrouse:
- [x] x86_64 (works properly)
- [ ] i686 (can't test)
- [ ] ARM/AARCH64 ()

When I run `stack setup` for my test, the following warning is produced:

```
Stack has not been tested with GHC versions above 8.6, and using 8.8.3, this may fail
[...]
Stack has not been tested with Cabal versions above 2.4, but version 3.0.1.0 was found, this may fail
```

Do we need to preconfigure it to use a version that they've tested?

packages/stack.rb
```diff
@@ -3,13 +3,10 @@
...
   homepage 'https://docs.haskellstack.org/en/stable/README/'
-  version '2.1.3'
+  version '2.1.3-1'
 
   case ARCH
-  when 'aarch64'
-    source_url 'https://github.com/commercialhaskell/stack/releases/download/v2.1.3/stack-2.1.3-linux-aarch64.tar.gz'
-    source_sha256 '96394b616e29b3bd2479a3712e2c0c375a7cfd748b613f7c6c3186d7ae92977b'
-  when 'armv7l'
+  when 'aarch64', 'armv7l'
```
cstrouse, line 9: The stack tool itself runs on ARM but the tools which make up the stack are still unavailable for aarch64 and possibly ARMv7l (haven't tried on a machine yet or read their issue queue in a while). We should probably just make it unavailable for aarch64 for now.

packages/stack.rb
```diff
@@ -48,7 +34,7 @@ def self.postinstall
     puts "source ~/.bashrc".lightblue
     puts
     puts "To completely uninstall stack, execute the following:".lightblue
-    puts "crew remove cabal ghc stack".lightblue
+    puts "crew remove stack".lightblue
     puts "rm -rf #{CREW_PREFIX}/share/stack".lightblue
```
cstrouse, line 38: This directory was removed for me when I ran `crew remove stack` so maybe we can get rid of this line.

uberhacker: I just included it as part of the entire process.  Probably doesn't hurt to leave it for clarity.

uberhacker:
Thanks for testing this @cstrouse.  Yeah, we should definitely include a supported version.  Not sure what's up with stack.  Apparently, it thinks ghc 8.8.3 is what needs to be installed even though that is not a supported aarch64 version.  We don't need to worry about i686 since that architecture is no longer supported and many projects now are not including 32-bit Intel/AMD binaries anymore.

packages/stack.rb
uberhacker: Yeah, this directory can be populated with other stack tools like ghc, for one.  Definitely should leave it.

uberhacker:
@cstrouse: I had the same issue with armv7l and i686 so I removed support for those also.  The only supported architecture is x86_64 at this time.

cstrouse:
There's a guy working on ARM binaries according to a an issue or PR in the upstream repo. Hopefully Google will stabilize the 64-bit userspace for Kevin soon so I can build a fully 64-bit image.

PR #4093, opened 2020-05-27
[wxwidgets] Update to 3.0.5.1

vincowl:
Added a gtk3 dependency. Was usefull to projects I had, but can remove it if needed


uberhacker:
Thank you for the update @vincowl!  Nice to see the use of `#{CREW_OPTIONS}`.

PR #4094, opened 2020-05-27
Add doom_emacs package

vincowl:
I didn't add a `gemacs` dependency since the `gemacs` -> `emacs` dependency link leads to compilation errors with release 27 which is still not declared as the reference one but marked as "recommanded" to doom-emacs. Other debate that could be opened later : are `emacs` / `gemacs` still needed or shall we merge them ?

### DIFF
--- a/packages/doom_emacs.rb
+++ b/packages/doom_emacs.rb
@@ -1,0 +1,91 @@
+require 'package'
+
+class Doom_emacs < Package
+  description 'An extensible, customizable, free/libre text editor - and more.'
+  homepage 'https://www.gnu.org/software/emacs/'
+  version '27.0.91'
+  source_url 'http://git.savannah.gnu.org/cgit/emacs.git/snapshot/emacs-27.0.91.tar.gz'
+  source_sha256 'ac7b71927dfdfa3b9c034da82e827ee28e7467d58953dafed69e9ce9d700cebd'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'giflib'
+  depends_on 'libjpeg'
+  depends_on 'libpng'
+  depends_on 'libtiff'
+  depends_on 'cairo'
+  depends_on 'libxaw'
+  depends_on 'libxpm'
+  depends_on 'wxwidgets'
+  depends_on 'imagemagick6'
+  depends_on 'sommelier'
+
+  def self.build
+    system './autogen.sh'
+    system './configure',
+            "--prefix=#{CREW_PREFIX}",
+            "--localstatedir=#{CREW_PREFIX}/share",
+            '--without-selinux',
+            '--with-x=yes',
+            '--with-x-toolkit=gtk',
+            '--with-gif=yes',
+            '--with-jpeg=yes',
+            '--with-png=yes',
+            '--with-rsvg=yes',
+            '--with-xaw=yes',
+            '--with-xpm=yes',
+            '--with-tiff=yes',
+            '--without-makeinfo',
+            '--with-pdumper',
+            '--with-gnutls',
+            '--with-cairo'
+    system 'make'
+  end
+
+  def self.install
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    system 'install', '-Dm755', 'src/emacs', "#{CREW_DEST_PREFIX}/bin/gemacs"
+    system 'install', '-Dm755', 'src/emacs-27.0.91.1', "#{CREW_DEST_PREFIX}/bin/gemacs-27.0.91"
+    if Dir.exist?("#{HOME}/.emacs.d")
+      puts "~/.emacs.d already exists. Delete it ? (y/n) ".orange
+      while answer = STDIN.gets.chomp
+        case answer
+          when 'y'
+            puts "Removing ~/.emacs.d...".orange
+            FileUtils.remove_dir("#{HOME}/.emacs.d",true) 
+            break
+          when 'n'
+            puts "Doom_emacs not installed. Please remove ~/.emacs.d".lightred
+            abort
+            break
+          else
+            puts "  Please select from one of the options."
+        end
+      end
+    end
+    system "git clone --depth 1 https://github.com/hlissner/doom-emacs #{HOME}/.emacs.d"
+    system "#{HOME}/.emacs.d/bin/doom -y install"
+  end
+
+  def self.postinstall
+    system "#{HOME}/.emacs.d/bin/doom sync"
+    system "#{HOME}/.emacs.d/bin/doom update"
+    puts
+    puts "To complete the installation, execute the following:".lightblue
+    puts "echo 'export PATH=$PATH:$HOME/.emacs.d/bin' >> ~/.bashrc".lightblue
+    puts "source ~/.bashrc".lightblue
+    puts
+    puts "Available Commands :"
+    puts " - " + "'doom sync'".blue + " to synchronize your private config with Doom by installing missing packages, removing orphaned packages, and regenerating caches. Run this whenever you modify your private init.el or packages.el, or install/remove an Emacs package through your OS package manager (e.g. mu4e or agda).".lightblue
+    puts " - " + "'doom upgrade'".blue + " to update Doom to the latest release & all installed packages.".lightblue
+    puts " - " + "'doom update'".blue + " to update all installed packages to the latest release.".lightblue
+    puts " - " + "'doom doctor'".blue + " to diagnose common issues with your system and config.".lightblue
+    puts " - " + "'doom env'".blue + " to dump a snapshot of your shell environment to a file that Doom will load at startup. This allows Emacs to inherit your PATH, among other things.".lightblue
+    puts " - " + "'doom build'".blue + " to recompile all installed packages (use this if you up/downgrade Emacs).".lightblue
+    puts
+  end
+
+end


packages/doom_emacs.rb
```diff
@@ -0,0 +1,91 @@
+require 'package'
+
+class Doom_emacs < Package
+  description 'An Emacs framework for the stubborn martian hacker'
+  homepage 'https://github.com/hlissner/doom-emacs'
+  version '27.0.91'
+  source_url 'http://git.savannah.gnu.org/cgit/emacs.git/snapshot/emacs-27.0.91.tar.gz'
```
uberhacker, line 7: Can we use https://git.savannah.gnu.org/cgit/emacs.git/snapshot/emacs-27.0.91.tar.gz?  Since https works.

packages/doom_emacs.rb
```diff
@@ -0,0 +1,91 @@
...
+            '--without-makeinfo',
+            '--with-pdumper',
+            '--with-gnutls',
+            '--with-cairo'
+    system 'make'
+  end
+
+  def self.install
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    system 'install', '-Dm755', 'src/emacs', "#{CREW_DEST_PREFIX}/bin/gemacs"
```
uberhacker, line 50: Won't this conflict with the current gemacs package since the executable name is the same?

packages/doom_emacs.rb
```diff
@@ -0,0 +1,91 @@
...
+          when 'n'
+            puts "Doom_emacs not installed. Please remove ~/.emacs.d".lightred
+            abort
+            break
+          else
+            puts "  Please select from one of the options."
+        end
+      end
+    end
+    system "git clone --depth 1 https://github.com/hlissner/doom-emacs #{HOME}/.emacs.d"
```
uberhacker, line 69: Should we checkout a tagged release here?  Otherwise, this is on the master (or default) branch which may change over time and cause potential incompatibilities.

packages/doom_emacs.rb
```diff
@@ -0,0 +1,91 @@
...
+    system "#{HOME}/.emacs.d/bin/doom -y install"
```
uberhacker, line 70: Does this add extra files we could include in the package?  Maybe something like `"#{CREW_DEST_HOME}/.emacs.d/bin/doom -y install"`?

packages/doom_emacs.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  def self.postinstall
+    system "#{HOME}/.emacs.d/bin/doom sync"
+    system "#{HOME}/.emacs.d/bin/doom update"
```
uberhacker, line 75: Can these be in the `def self.install` section to include in the package?  If that is the case, it would be `#{CREW_DEST_HOME}` instead of `#{HOME}`.

packages/doom_emacs.rb
```diff
@@ -0,0 +1,91 @@
...
+    puts "echo 'export PATH=$PATH:$HOME/.emacs.d/bin' >> ~/.bashrc".lightblue
+    puts "source ~/.bashrc".lightblue
+    puts
+    puts "Available Commands :"
+    puts " - " + "'doom sync'".blue + " to synchronize your private config with Doom by installing missing packages, removing orphaned packages, and regenerating caches. Run this whenever you modify your private init.el or packages.el, or install/remove an Emacs package through your OS package manager (e.g. mu4e or agda).".lightblue
+    puts " - " + "'doom upgrade'".blue + " to update Doom to the latest release & all installed packages.".lightblue
+    puts " - " + "'doom update'".blue + " to update all installed packages to the latest release.".lightblue
+    puts " - " + "'doom doctor'".blue + " to diagnose common issues with your system and config.".lightblue
+    puts " - " + "'doom env'".blue + " to dump a snapshot of your shell environment to a file that Doom will load at startup. This allows Emacs to inherit your PATH, among other things.".lightblue
+    puts " - " + "'doom build'".blue + " to recompile all installed packages (use this if you up/downgrade Emacs).".lightblue
```
uberhacker, line 87: This is great info.  Should we include a manpage in the package also?

uberhacker:
Thank you for the new package @vincowl!  Please see comments.  I'm thinking maybe we can include the `~/.emacs.d` folder in the package?

packages/doom_emacs.rb
uberhacker: `system "git clone --depth 1 https://github.com/hlissner/doom-emacs #{CREW_DEST_HOME}/.emacs.d"`?

vincowl:
OK, let's handle the key-point of this package. DoomEmacs is not a software on its own. It's more or less a configuration setup added to emacs. So, yes, there will be an impact on a previous emacs installation. Configuration files in emacs or often stored in ~/.emacs.d (installed packages, customizations, shortcuts, ...). Doom has its own preselected packages, installs a part of them in ~/.emacs.d but also creates a  ~/.doom folder containing the files to be modified by the user : packages activation, customizations, ... So, if I want to install emacs and doom with different configs, I will have to handle this and modify one of them. I am going to find a way to install the doom part somewhere else than ~/.emacs.d but don't know where and if it won't be overwritten by the doom commands.

Seems like doom also needs `emacs` command which will be overwritten too in the first line of the install step. So : trouble with an already installation of emacs which has an other version. That's why when we first talked about this in #4059, I said that I don't know if doom has to be a package on its own.

So including ~/.emacs.d or ~/.doom in the package would be a non-sense to me since it handles user's configurations and is managed by the git part of doom. This leads to the other point which is doom's release : there is no release notion in doom since it acts more or less like a rolling release system (a bit like chromebrew). This package would just ease doom's installation, but will never point to a stable release. The last release is from Jan 2018, with A LOT of commits since this date. Updates are handled by `doom update` and `doom upgrade`.

I think we have to define a global philosophy for this kind a "packages" before resolving reviews. It would be the same philo regarding `spacemacs` or `spacevim` that we told about in #4059

vincowl:
Additional info : I just tried to change the prodiced "emacs" binary's name, but once I `git cloned` the doom-emacs setup, I got the following message : 
```
Can't find emacs in your PATH
doom_emacs failed to install: `/home/chronos/user/.emacs.d/bin/doom -y install` exited with 1
```
So... basically, it looks like we can't install emacs AND doom-emacs in two different version since the binary will have the same name. Maybe we could install them in differents folders. But is it worth it ?

uberhacker:
Why don't you name the binary demacs instead of gemacs?

vincowl:
I could, it works. But as previously said, this isn't the part that blocks the package. I need a binary named emacs, which would interfer with our emacs package. There is no release in doom, and so on. So tell me if it is worth it to address the different remarks given the blocking points raised in my previous (long !) comment. I don't want to waste time if doom-emacs can not be seen as a package and would more likely take a place in a wiki page dedicated to such packages which are "upper-layers" with no release and requirements that exclude other packages (here, emacs and gemacs)

uberhacker:
To version "master" branches, you could checkout a recent git hash to pin it at a "version".  This at least gives you a predictable snapshot of a known good working state.  So something like:
```
system "cd #{CREW_DEST_HOME}/.emacs.d && git checkout 75b6e11f"
```

vincowl:
> To version "master" branches, you could checkout a recent git hash to pin it at a "version".  This at least gives you a predictable snapshot of a known good working state.  So something like:
> ```
> system "cd #{CREW_DEST_HOME}/.emacs.d && git checkout 75b6e11f"
> ```

I suppose the "version" field had to be set to "75b6e11f" in your example ?

What do you think of the emacs / doom mutual exclusion ?

uberhacker:
I'm not familiar enough with emacs to provide valuable feedback.  Just based on my past experience with packages, it probably makes sense to use a git hash for the version.  Including the `~/.emacs.d` directory in the install just makes it a bit easier to uninstall but if this is shared with other emacs packages, it might not make sense to do it this way.  Maybe you can target specific doom config files and just include those in the package?  I don't really know the best way to approach this issue.

vincowl:
Just found out this solution : [chemacs](https://github.com/hlissner/doom-emacs/blob/develop/docs/getting_started.org#install-doom-alongside-other-configs-with-chemacs)

This would allow us to use emacs / gemacs / spacemacs / doomemacs alongside. I'll check it out tomorrow but I should manage to install doom's layer in a separate folder (not ~/.emacs.d). I'll test doom with emacs 26.3. if all of this works, then I suggest to strongly link doom to the emacs / gemacs packages, and call `gemacs`-based aliases to run each emacs-like packages and let chemacs do the rest. 

I am still dubitative about the real added value to make doom a package on its own, but why not ? Indeed every added value is stored in the ~/.emacs.d and ~/.doom folders which should not be added to the install step because when adding emacs add-ons (like VSCode plugins), these folders will be modified as long as we modify configuration. Thus I strongly suspect these folders not being removed when deleting doom's package. Is this assertion true ?

vincowl:
Just a few words to state on this PR : after a bunch of tests, chemacs works perfectly. But :

- doom-emacs cannot play alongside emacs or spacemacs without chemacs. But I don't think it's chromebrew's scope to handle this but rather a user's prerogative. So I don't think we should handle this. Just keep a blocking warning message as actually is.
- doom-emacs is just a `git clone` and ` doom install` command and lives its life on its own.
- doom-emacs works well as an upper layer to the `gemacs` package

Then I think we shouldn't care about managing a `doom-emacs` package since it's totally trivial to install and we will have to handle a package that is constantly evolving at each commit the doom-emacs developer runs (no release available, so when should we make this package evolve ?). So, I definitely think we should close this PR and not merge it.

uberhacker:
Should we keep this open @vincowl?

vincowl:
@uberhacker I think we can close this since I strongly believe there is much more benefit to maintain only emacs / gemacs since doom-emacs has its own update mechanism

PR #4097, opened 2020-05-30
Add libsolv package

uberhacker:
This is libsolv, a free package dependency solver using a satisfiability algorithm.

The code is based on two major, but independent, blocks:

Using a dictionary approach to store and retrieve package and dependency information in a fast and space efficient manner.

Using satisfiability, a well known and researched topic, for resolving package dependencies.

The sat-solver code has been written to aim for the newest packages, record the decision tree to provide introspection, and also provides the user with suggestions on how to deal with unsolvable problems. It also takes advantage of repository storage to minimize memory usage.

See https://github.com/openSUSE/libsolv.  Tested on all architectures.

### DIFF
--- a/packages/libsolv.rb
+++ b/packages/libsolv.rb
@@ -1,0 +1,49 @@
+require 'package'
+
+class Libsolv < Package
+  description 'Library for solving packages and reading repositories'
+  homepage 'https://en.opensuse.org/openSUSE:Libzypp_satsolver'
+  version '0.7.14'
+  #compatibility 'all'
+  source_url 'https://github.com/openSUSE/libsolv/archive/0.7.14.tar.gz'
+  source_sha256 '34a67c23cedd38a949607d8cdbd867aa4b07556f18a57e17162ac962bbdbbf54'
+
+  binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/libsolv-0.7.14-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/libsolv-0.7.14-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/libsolv-0.7.14-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/libsolv-0.7.14-chromeos-x86_64.tar.xz',
+  })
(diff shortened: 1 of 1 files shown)


cstrouse:
Is bringing tcl/tk into this really necessary for a library that is likely to be used primarily by CLI utilities?

uberhacker:
Just gives the package more options in case they are needed.  I can remove if you insist.

packages/libsolv.rb
```diff
@@ -0,0 +1,50 @@
...
+  binary_sha256 ({
+    aarch64: 'baa2264b4f46c5690506e938b292ff72f6837f2bb74adbc7d5b524699d95302f',
+     armv7l: 'baa2264b4f46c5690506e938b292ff72f6837f2bb74adbc7d5b524699d95302f',
+       i686: '40baa5282bb8c1ce724285196e2a1bab68798272690ad8142fe197362d470e72',
+     x86_64: '7e7684fa8afa1f56a43963fee3360c6db9c25395f34bad1d22f448e8cd80e70c',
+  })
+
+  depends_on 'lzma'
+  depends_on 'tk'
+  depends_on 'zstd'
```
cstrouse, line 26: I was unable to build from source without installing `swig` first.

packages/libsolv.rb
```diff
@@ -0,0 +1,50 @@
...
+    Dir.chdir 'build' do
+      system 'cmake',
+             '-DCMAKE_BUILD_TYPE=Release',
+             "-DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX}",
+             '-DENABLE_BZIP2_COMPRESSION=ON',
+             '-DENABLE_LZMA_COMPRESSION=ON',
+             '-DENABLE_ZSTD_COMPRESSION=ON',
+             '-DENABLE_PYTHON=ON',
+             '-DENABLE_RUBY=ON',
+             '-DENABLE_TCL=ON',
```
cstrouse, line 39: Can we make the tcl/tk feature installation check for a pre-existing X environment or the tcl/tk packages being installed already? Otherwise, maybe we should add a configuration option to make certain feature subsets optional.

packages/libsolv.rb
uberhacker: Added.

packages/libsolv.rb
uberhacker: I just removed it.  That means we'll probably need it later.  Usually how it works.

PR #4113, opened 2020-06-05
Add pre-built binaries for girara, zathura and zathura_poppler_pdf

uberhacker:
Tested on all architectures.  @vincowl: During my testing, I found that zathura is unusable without zathura_poppler_pdf, so I made that a dependency of zathura.  You can also build zathura_poppler_pdf without zathura installed because it is only a run-time dependency.  Hopefully, that makes sense.

### DIFF
--- a/packages/girara.rb
+++ b/packages/girara.rb
@@ -8,8 +8,16 @@ class Girara < Package
   source_sha256 'e78257e4218a0f7f59cc1bea472c7c6794fa51cd9261d87affbe731c1e22c6a5'
 
   binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/girara-0.3.4-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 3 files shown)


vincowl:
Zathura is a reader based on plugins. Pdf rendering using poppler is one solution among others. That's why I did it that way. I chose pdf / poppler since poppler was an existing package but we could have use pdf / mupdf for instance or cbz plugin or ps, or....

Not sure adding zathura_poppler_pdf to zathura as a dependency is the logical way but indeed it offers a quick, direct and functional way to users who are not aware of this plugin system. So that's fine to me ! Thanks

We will think about it if we add other plugins. I think we should handle this as we did with hunspell for instance

uberhacker:
@vincowl:  Thank you for the explanation.  That makes sense.  We can make it work more like hunspell once other plugins are introduced.

PR #4116, opened 2020-06-06
[filezilla] Update to 3.48.1

vincowl:
Filezilla binaries had to be rebuild since last updates and compilation failed so upgrading it to 3.48.1 solves both troubles


uberhacker:

PR #4132, opened 2020-06-16
add roswell package

lloydhuang:
Roswell is a Lisp installer and launcher for major environment.'  . HomePage https://github.com/roswell/roswell . Please help to review it. Thank you.

### DIFF
--- a/packages/roswell.rb
+++ b/packages/roswell.rb
@@ -9,17 +9,16 @@ class Roswell < Package
   source_sha256 '798b96fdcb8c89445b36692b31570bb99882d83719d6310d969ccfcb2a35a1d4'
 
   binary_url ({
-    aarch64: '',
-     armv7l: '',
-       i686: '',
-     x86_64: '',
   })
   binary_sha256 ({
-    aarch64: '',
-     armv7l: '',
-       i686: '',
-     x86_64: '',
   })
+
+  depends_on 'brotli'
+  depends_on 'curl'
+  depends_on 'libcyrussasl'
+  depends_on 'libnghttp2'
+  depends_on 'openldap'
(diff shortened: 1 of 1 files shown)


packages/roswell.rb
```diff
@@ -0,0 +1,33 @@
...
+     armv7l: '',
+       i686: '',
+     x86_64: '',
+  })
+  binary_sha256 ({
+    aarch64: '',
+     armv7l: '',
+       i686: '',
+     x86_64: '',
+  })
```
uberhacker, line 22: You can remove this entire section.  It will be added when the pre-built binaries are generated.  No dependencies?

uberhacker:
Thank you for the new package @lloydhuang!

packages/roswell.rb
lloydhuang: Remove this entire section. Got it.
dependencies as below, Is it okay?

```
depends_on 'brotli'
depends_on 'curl'
depends_on 'krb5'
depends_on 'libcyrussasl'
depends_on 'libiconv'
depends_on 'libidn2'
depends_on 'libnghttp2'
depends_on 'libpsl'
depends_on 'libressl'
depends_on 'libssh2 '
depends_on 'libunistring'
depends_on 'openldap'
depends_on 'rtmpdump'
depends_on 'zlibpkg'
```

uberhacker: You don't need to include krb5, libiconv, libidn2, libpsl, libressl, libssh2, libunistring and zlibpkg since they are already in crew "core".

lloydhuang: Got it. Thank you.

PR #4160, opened 2020-07-07
[ghc] Allow --version option

vincowl:


packages/ghc.rb
```diff
@@ -3,7 +3,7 @@
 class Ghc < Package
   description 'The Glasgow Haskell Compiler is a state-of-the-art, open source compiler and interactive environment for the functional language Haskell.'
   homepage 'https://www.haskell.org/ghc/'
-  version '8.6.5'
+  version '8.6.5-1'
```
uberhacker, line 6: Why not upgrade while you're at it?  Latest is https://downloads.haskell.org/~ghc/8.10.1/ghc-8.10.1-src.tar.lz.

uberhacker:

packages/ghc.rb
vincowl: Please have a look at #4161 and #4162. Indeed it was necessary to upgrade ghc. But since `ghc` is a sandboxed version installed from `stack`, we need to also upgrade stack. Seems like latest stack release (2.3.1) is bound to `ghc` 8.8.3 instead of 8.10.1.

I attempted to install ghc from sources but failed as expressed on the slack channel

vincowl:
Sorry, too fast closing the PR : just want to be sure this step towards uprgade is unusefull

uberhacker:
Fixed in #4162.

PR #4170, opened 2020-07-13
Update powerstat from 0.02.22 to 0.02.24

cstrouse:
Tested on ARM.


uberhacker:

PR #4191, opened 2020-08-01
Add remmina package and dependencies

uberhacker:
remmina - The GTK Remmina Remote Desktop Client
avahi - Avahi is a system which facilitates service discovery on a local network via the mDNS/DNS-SD protocol suite.
libdaemon - libdaemon is a lightweight C library that eases the writing of UNIX daemons.

Tested on all architectures.

### DIFF
--- a/packages/avahi.rb
+++ b/packages/avahi.rb
@@ -1,0 +1,39 @@
+require 'package'
+
+class Avahi < Package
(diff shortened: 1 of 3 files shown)


leonbollerup:
Any fix for this:

chronos@localhost ~ $ remmina
remmina: error while loading shared libraries: libssl.so.47: cannot open shared object file: No such file or directory

PR #4204, opened 2020-08-11
Upgrade emacs and gemacs to 27.1

vincowl:
@uberhacker Please note that I had to remove the `emacs` dependency to the `gemacs` package. Indeed I couldn't make it work as before due to fingerprints errors (installations split). 
I think it is OK since emacs and gemacs aims two different needs : emacs is lightest as possible without X layer, whereas gemacs is "classic". One who wish to install emacs may not want gemacs and one who wants gemacs won't need emacs since he can launch gemacs without X if needed.

### DIFF
--- a/packages/emacs.rb
+++ b/packages/emacs.rb
@@ -3,22 +3,14 @@ require 'package'
 class Emacs < Package
   description 'An extensible, customizable, free/libre text editor - and more.'
   homepage 'http://www.gnu.org/software/emacs/'
-  version '26.3'
+  version '27.1'
   compatibility 'all'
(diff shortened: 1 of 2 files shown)


uberhacker:
Thank you for the updates @vincowl!  Since vim is a better editor, I think I'll do that next. ;)

vincowl:
> Thank you for the updates @vincowl!  Since vim is a better editor, I think I'll do that next. ;)

Not really true : neovim has my favours ;-) I wanted to give emacs a new chance, using upper layers like spacemacs or doom-emacs (true that emacs shines with latex, and org better suits my needs than markdown for my docs), but still cannot find such an easy way to set all this up as done with vim.

PR #4210, opened 2020-08-15
Package edit - Netbeans - Checksum was outdated

ThatGeekyWeeb:
Updated `netbeans` to use newest source package & checksum - Was outdated


ThatGeekyWeeb:
Can confirm, `netbeans` stills compiles fine without any additional edits.

uberhacker:
Thank you for the update @ThatGeekyWeeb!

PR #4212, opened 2020-08-15
[filezilla] Update to 3.49.1

vincowl:


uberhacker:
Thank you for the updates @vincowl!

PR #4213, opened 2020-08-15
New Package: librespot - Open Source Spotify Connect library

ThatGeekyWeeb:


packages/librespot.rb
```diff
@@ -0,0 +1,20 @@
+require 'package'
+
+class Librespot < Package
+  description 'Open Source Spotify connect library'
+  homepage 'https://github.com/librespot-org/librespot'
+  version '0.1.3-1'
```
cstrouse, line 6: Why did you add `-1` to the package version?

ThatGeekyWeeb: I tend to sign my scripts with `version-revision` - incase the script is revised later on, I can remove that if I that should not be done.

cstrouse:
We only add revisions to package versions when there's a change to the package definition but not a new version of the upstream software. You can remove that on the initial package. I'll test this quick and get it merged if it works. Thanks for the new package.

cstrouse:
I'm not able to compile this on a fresh install. I'm on a Pixel Slate.

Screenshot 2020-08-15 at 12 55 19 PM

ThatGeekyWeeb:
Hm, I didn't add the alsa deps as my Chromebook came with alsa<sup>(And I figured they all did)</sup>, I'll add them in case that is the issue.

cstrouse:
It installs now.

PR #4221, opened 2020-08-17
[cras] Update to 13360.B

vincowl:


packages/cras.rb
```diff
@@ -55,16 +50,21 @@ def self.build
...
              "--libdir=#{CREW_LIB_PREFIX}",
+             '--disable-alsa-plugin',
              '--disable-static',
              '--disable-webrtc-apm',
              '--enable-sse42',
              '--enable-avx',
              '--enable-avx2',
              '--enable-fma'
-      system 'make'
+      system "make -j$(nproc)"
```
uberhacker, line 67: This could be `system "make -j#{CREW_NPROC}"`

packages/cras.rb
```diff
@@ -55,16 +50,21 @@ def self.build
 }
 _EOF_'
       system './git_prepare.sh'
+      ENV['CC'] = 'clang'
+      ENV['CXX'] = 'clang++'
+      ENV['CXXFLAGS'] = "-g -O2 -std=gnu++11 -Wall"
+      ENV['CFLAGS'] = "-g -O2 -Wall"
```
uberhacker, line 56: Do we really need clang?  Wouldn't that require `depends_on 'llvm' => :build`.

vincowl: It failed while compiling with gcc. Don't know why... Indeed llvm is installed on my machine, so did not thought about that.

PR #4231, opened 2020-08-18
New Package: pavucontrol - Packaged with templator

ThatGeekyWeeb:
I'm working on packaging `pavucontrol` it's been relativity easy due to [Templator](https://github.com/ThatGeekyWeeb/templator), It takes `template`'s from the void-linux repo and converts them into 70% usable package scripts. My main issues are `gtkmm3` & `glibmm` both of which are outdated to the point where the linkers fail 100% of the time, I've repackaged them both and I'm still working on tests, I'm opening this PR to ask for help where it is needed, as there are many issues, and compiling `gtkmm3` and `glibmm` is not practical as it takes a very long time. Hopefully after the updates to `gtkmm3` and `glibmm` I provided, there will be no issues, otherwise I hope I can ask for help here! :)
~
All of the package scripts here were generated using [Templator](https://github.com/ThatGeekyWeeb/templator)
~
`.gitignore` should be ignored, I've used to it prevent pushing `templator.sh` and the templates I've sourced from void linux.

### DIFF
--- a/packages/gtkmm3.rb
+++ b/packages/gtkmm3.rb
@@ -3,38 +3,21 @@ require 'package'
 class Gtkmm3 < Package
   description 'The Gtkmm3 package provides a C++ interface to GTK+ 3.'
   homepage 'https://www.gtkmm.org/'
-  version '3.22.2'
+  version '3.24.1'
   compatibility 'all'
-  source_url 'https://ftp.gnome.org/pub/gnome/sources/gtkmm/3.22/gtkmm-3.22.2.tar.xz'
-  source_sha256 '91afd98a31519536f5f397c2d79696e3d53143b80b75778521ca7b48cb280090'
-
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gtkmm3-3.22.2-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/gtkmm3-3.22.2-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/gtkmm3-3.22.2-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/gtkmm3-3.22.2-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -0,0 +1,4 @@
+templator.sh
+template*
+temp*
+spkg*
```
uberhacker, line 4: We should not include this .gitignore.  We may exclude things we might want to include such as a package named temperature.rb, for example.  Most commits in this repo are packages and a `.gitignore` is better off ... ignored. :)

ThatGeekyWeeb: I plan to remove that after I get everything working!

packages/gtkmm3.rb
```diff
@@ -3,38 +3,24 @@
...
   depends_on 'pangomm'
 
   def self.build
     # fix the documents directory name
-    system "sed -e '/^libdocdir =/ s/$(book_name)/gtkmm-3.22.2/' \
-    -i docs/Makefile.in"
-    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
-    system "make"
+    # system "sed -e '/^libdocdir =/ s/$(book_name)/gtkmm-3.22.2/' \
+    # -i docs/Makefile.in"
```
uberhacker, line 18: It's fine to remove these lines completely.

ThatGeekyWeeb:
`pavucontrol` compiles fine with these updates, but it is not at a usable state, attempting to adjust a volume in "Output Devices" causes it to instantly crash, I'll work on fixing this.

ThatGeekyWeeb:
Still not working.... I need to somehow redirect traffic from alsa to pulseaudio, can anyone here help with that? The normal means don't work as the alsa modules are missing..

ThatGeekyWeeb:
I'm going to make those edits, at the current time note that it does not work due to pulseaudio not working without some advanced configuration.

ThatGeekyWeeb:
@uberhacker In order to use this package additional configuration is required by the user, (Configuration ik nothing about), this package works fundamentally in it's current state. `gtk3mm` and `glibmm` should be built into binaries as building them takes a relatively long time.

uberhacker:
The general "rule" is to use single quotes and only use double quotes when necessary.  This makes it easier to extract details from package contents if there is a consistent format.  Packages should use 2 space indentation as well.

PR #4261, opened 2020-08-30
New Package: Netsurf

ThatGeekyWeeb:


uberhacker:
@ThatGeekyWeeb:  Thank you for the new packages!  It looks like you put a lot of effort into this project.  The only things I noticed were all the http urls can be https instead and if you don't include `=> :build` on all the `depends_on 'netsurf_buildsystem'` lines, it will be a runtime install.  Also, do we need `depends_on 'vim'` in `netsurf.rb`?

ThatGeekyWeeb:
Alright! The `http` links we're just taken directly from the void-packages package template, and `depends_on 'vim'` was used as void-packages states netsurf depends on `xxd` which is just `vim` within their repo. I can't remember why I used `=> :build` on `depends_on 'netsurf_buildsystem'`, I'll remove that, and change `http` to `https`

ThatGeekyWeeb:
> https://github.com/void-linux/void-packages/blob/master/srcpkgs/netsurf/template
> ```sh
> <...>
> hostmakedepends="<...> xxd"
> <...>
> ```

> https://github.com/void-linux/void-packages/blob/master/srcpkgs/xxd
> ```
> Symbolic Link
> vim
> ```

PR #4271, opened 2020-09-02
New Package: elogind

ThatGeekyWeeb:


uberhacker:
Thank you for the new package @ThatGeekyWeeb!  I was working on this awhile back but never pushed it since it was a dependency of something else at the time.

packages/elogind.rb
```diff
@@ -0,0 +1,30 @@
...
+  depends_on 'linux_pam' # For _pam_macros.h
+  depends_on 'dbus'
+  depends_on 'docbook_xsl'
+  depends_on 'gperf'
+  depends_on 'libxslt'
+  depends_on 'shadow'
+  depends_on 'glib'
+
+  def self.build
+    system "meson --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} _build -Dcgroup-controller=elogind -Dman=false -Ddefault-hierarchy=legacy -Ddefault-kill-user-processes=false -Dhalt-path=/sbin/halt -Drootlibexecdir=#{CREW_PREFIX}/libexec/elogind -Dreboot-path=/sbin/reboot"
```
uberhacker, line 23: Can we split the options on separate lines to see more clearly?

packages/shadow.rb
```diff
@@ -0,0 +1,20 @@
...
+  homepage 'https://github.com/shadow-maint/shadow'
+  version '4.8.1'
+  compatibility 'all'
+  source_url 'https://github.com/shadow-maint/shadow/releases/download/4.8.1/shadow-4.8.1.tar.xz'
+  source_sha256 'a3ad4630bdc41372f02a647278a8c3514844295d36eefe68ece6c3a641c1ae62'
+
+  depends_on 'openpam'
+
+  def self.build
+      system "./configure #{CREW_OPTIONS} --bindir=#{CREW_PREFIX}/bin --sbindir=#{CREW_PREFIX}/bin --enable-shared --disable-static --with-libpam --without-selinux --with-acl --with-attr --without-su --disable-nls --enable-subordinate-ids --disable-account-tools-setuid"
```
uberhacker, line 14: Can we split the options on separate lines to see more clearly?

packages/shadow.rb
```diff
@@ -0,0 +1,20 @@
...
+      system "make -j#{CREW_NPROC}"
```
uberhacker, line 15: FYI, `-j#{CREW_NPROC}` is not needed since that is the default option.  Check out `lib/package.rb` starting on line 108.

ThatGeekyWeeb:
Heh, I packaged elogind as it's a dep for `xfce4`  (I'm working on packaging that, theres alot of work to do)

packages/shadow.rb
ThatGeekyWeeb: Good to know :+1:

PR #4267, opened 2020-08-31
New Package: libefl

ThatGeekyWeeb:
## Changes
`font_util` - Updated to `1.3.2`
`ghostscript` - Updated to `9.52` - Add `libgs` shared library - `libspectre` dependency
`libfontenc` - Updated to `1.1.4`

## Additions 
`libefl`
~
`bdftopcf` - `gsfonts` dependency
`bullet` - `libefl` dependency
`font_misc_misc` - `gsfonts` dependency
`gsfonts` - `ghostscript` dependency
`libpaper` - `ghostscript` dependency
`libraw` - `libefl` dependency
`printproto` - `libxp` dependency
`libxp` - `libefl` dependency
`libspectre` - `libefl` dependency
`libxpresent` - `libefl` dependency
`mkfontscale` - `gsfonts` dependency

PR #4272, opened 2020-09-02
Correct dependencies declaration

vincowl:
Since #4267, `ghostscript` update fails due to dependencies errors. These are not applied to the correct packages.


uberhacker:
Thank you for the fix @vincowl!  Got this covered in #4274.

PR #4277, opened 2020-09-03
New Package: Gnome Keyring

ThatGeekyWeeb:


ThatGeekyWeeb:
I've provided binaries for armv7l, just to help :)

packages/gcr.rb
```diff
@@ -0,0 +1,36 @@
...
+  compatibility 'all'
+  source_url 'https://ftp.gnome.org/pub/GNOME/sources/gcr/3.36/gcr-3.36.0.tar.xz'
+  source_sha256 'aaf9bed017a2263c6145c89a1a84178f9f40f238426463e4ae486694ef5f6601'
+
+  binary_url ({
+     armv7l: 'https://www.dropbox.com/s/ggpe4y6txpveomi/gcr-3.36.0-chromeos-armv7l.tar.xz',
+  })
+  binary_sha256 ({
+     armv7l: '84a9cb611f086c42cc920b2937ca8cb1c2287ae818ed58f61631bef421e6b41f',
+  })
```
uberhacker, line 16: You don't need to include the binary but nice to know you tested the build. :+1:

packages/gnome_keyring.rb
```diff
@@ -0,0 +1,37 @@
...
+  depends_on 'openpam'
+  depends_on 'dconf'
+  depends_on 'gcr'
+  depends_on 'docbook_xsl'
+  depends_on 'glib'
+  depends_on 'libxslt'
+  depends_on 'openssh'
+
+  def self.build
+      system "xmlcatalog /usr/local/etc/xml/catalog" # Fix dox-xml resovle issues
```
uberhacker, line 30: `system "xmlcatalog #{CREW_PREFIX}/etc/xml/catalog"`?

ThatGeekyWeeb: This doesn't need to actually be there, I just haven't removed it yet

packages/gcr.rb
ThatGeekyWeeb: Understood, I'll remove it!

PR #4284, opened 2020-09-06
New Package: picom

ThatGeekyWeeb:


packages/picom.rb
```diff
@@ -0,0 +1,36 @@
...
+    system "meson --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} _build -Dopengl=false" # Glx backend cannot be used yet
+    system 'ninja -v -C _build'
+  end
+  
+  def self.install
+    system "DESTDIR=#{CREW_DEST_DIR} ninja -C _build install"
+  end
+  def postinstall
+    puts
+    puts "Note that picom requires a WM to be running - Sommeiler does not count".lightblue
```
uberhacker, line 33: Sommelier misspelled?  Might want to give some examples of window managers to use.  Also, for clarity, it might be good to spell out WM in case someone doesn't understand the acronym.

PR #4287, opened 2020-09-07
[imagemagick7] Update to 7.0.10-28

vincowl:


packages/imagemagick7.rb
```diff
@@ -51,6 +45,7 @@ def self.patch
   end
 
   def self.build
+    FileUtils.ln_s "#{CREW_PREFIX}/include/locale.h", "#{CREW_PREFIX}/include/xlocale.h"
```
uberhacker, line 48: This should be in the `def self.install` section also so it is tracked.  See below:
```
FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/include"
FileUtils.ln_s "#{CREW_PREFIX}/include/locale.h", "#{CREW_DEST_PREFIX}/include/xlocale.h"
```

packages/imagemagick7.rb
```diff
@@ -51,6 +45,7 @@ def self.patch
   end
 
   def self.build
+    FileUtils.ln_s "#{CREW_PREFIX}/include/locale.h", "#{CREW_PREFIX}/include/xlocale.h" 
     system "CFLAGS=' -I#{CREW_PREFIX}/include/gdk-pixbuf-2.0 -I#{CREW_PREFIX}/include/c++/v1/support/xlocale' \
            ./configure \
            --prefix=#{CREW_PREFIX} \
```
uberhacker, line 51: You could replace `--prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}` with `#{CREW_OPTIONS}`.

packages/imagemagick7.rb
vincowl: It has to be defined before the build step otherwise it  won't compile. So : shall I correct this part and do the same in the install step ?

packages/imagemagick7.rb
vincowl: I'll do this. Didn't see that ... oops

packages/imagemagick7.rb
uberhacker: I'm just saying if you don't include in the `def self.install` section also, uninstalling this package won't remove that symlink.

vincowl: Ok. I'll add this in both places so with a try catch to avoid `File exists` error

uberhacker: It looks like locale.h is in the glibc package so it probably makes more sense to include the xlocale.h symlink there instead.  That said, it is a core package that would require more extensive testing.

vincowl: It would indeed be better to do so. I put this ln -s here to avoid this trouble but, if you guide me (which tests are required : packages relying on glibc ?), I will handle this since it would be a cleaner approach

uberhacker: It's not a nimble task because almost everything depends on glibc.  The most important thing is to not break install.sh.

vincowl: True, but since it is just an additional file used by nobody else but imagemagick7, files produced during glibc compilation won't be modified and then packages built upon glibc won't be affected since we don't upgrade glibc. No ?

uberhacker: In theory, yes.  In practice, maybe?  Still need to test.

vincowl (commented):
This should do the trick

packages/imagemagick7.rb
```diff
@@ -51,12 +45,13 @@ def self.patch
   end
 
   def self.build
-    system "CFLAGS=' -I#{CREW_PREFIX}/include/gdk-pixbuf-2.0 -I#{CREW_PREFIX}/include/c++/v1/support/xlocale' \
-           ./configure \
-           --prefix=#{CREW_PREFIX} \
-           --libdir=#{CREW_LIB_PREFIX} \
-           --disable-dependency-tracking \
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/include"
+    FileUtils.ln_s "#{CREW_PREFIX}/include/locale.h", "#{CREW_PREFIX}/include/xlocale.h", force: true
```
uberhacker, line 49: This could be:
```
FileUtils.ln_sf "#{CREW_PREFIX}/include/locale.h", "#{CREW_PREFIX}/include/xlocale.h"
```
This removes the `, force: true`.

packages/imagemagick7.rb
```diff
@@ -67,6 +62,8 @@ def self.build
   end
 
   def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/include"
+    FileUtils.ln_s "#{CREW_PREFIX}/include/locale.h", "#{CREW_PREFIX}/include/xlocale.h", force: true
```
uberhacker, line 66: This line should be:
```
FileUtils.ln_sf "#{CREW_PREFIX}/include/locale.h", "#{CREW_DEST_PREFIX}/include/xlocale.h"
```
That said, what happens if you patch the configure script to look for locale.h instead of xlocale.h?  Would that eliminate the need for the symlink?

vincowl: I don't know. I was thinking to patch the source files themselves... I'll try that this evening. The qtfm maintainer doesn't seem to be very active, so patching would be a better solution rather than submitting a PR

vincowl: @uberhacker I managed to avoid the xlocale.h symlink. This needs to patch imagemagick7 source files but also recompile perl.rb, which had an old dependency to xlocale.h in its binaries. This led to the need to add a symlink named libnsl.so. This lib is produced by glibc. I thought it was less impacting to insert it as close as possible to the actual need instead of insert it close to its origin.

uberhacker:
@vincowl: Thank you for the update!  Please see comments.  Have you tried to update perl?  It might be worth a shot.  Please create a separate PR if you do decide to take on the challenge.

vincowl:
OK, I'll have a look at it tomorrow. I just realize, that display and convert commands (imagemagick) are not available with this build even if the installation produced absolutely no error...

vincowl:
sed commands are no longer needed as I migrated to the github repo instead of their download site. Thus, xlocale.h have been solved. This also solves the legacy files availability troubles : old files are no more available to download once a new version is released which forbids a new installation from sources.

If I use `#{CREW_OPTIONS}` , package is built but binaries aren't available. This is due to the `--target` part in #{CREW_OPTIONS}. I think we should removed it since it is not mandatory : it is by default always set like --host. See [this url](https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Specifying-Target-Triplets.html)

vincowl:
I will correct `perl` file here and submit a PR to update it

uberhacker:
The idea behind `CREW_OPTIONS` is to tell the compiler to build for armv7l (32 bit) instead of armv8 (64 bit) on aarch64 architecture.  Even though it defaults, it will still need to be overridden for this architecture.

vincowl:
> The idea behind `CREW_OPTIONS` is to tell the compiler to build for armv7l (32 bit) instead of armv8 (64 bit) on aarch64 architecture. Even though it defaults, it will still need to be overridden for this architecture.

To be sure to understand this well : do you have a use case were `--target` content will be different to `--host` ? To me, what you intend to do only requires to set `--build` and `--host`

uberhacker:
To be clear, it is set explicitly since `config.guess` will set build to `aarch64-cros-linux-gnu` initially which sets host and then target to the same value.  We want to make sure all of the flags are set to `armv7l-cros-linux-gnuaebihf` on aarch64 architecture.

packages/imagemagick7.rb
```diff
@@ -51,18 +45,19 @@ def self.patch
...
-           --with-modules \
-           --enable-hdri \
-           --with-perl \
-           --with-rsvg \
-           --with-x"
+    system "./configure",
+           "CFLAGS=-I#{CREW_PREFIX}/include/gdk-pixbuf-2.0 -I#{CREW_PREFIX}/include/c++/v1/support/xlocale",
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}",
+           "--mandir=#{CREW_PREFIX}/share/man",
```
uberhacker, line 52: This could be `--mandir=#{CREW_MAN_PREFIX}` instead.

packages/imagemagick7.rb
```diff
@@ -51,18 +45,19 @@ def self.patch
...
+           "--with-windows-font-dir=#{CREW_PREFIX}/share/fonts/truetype/msttcorefonts",
```
uberhacker, line 53: Nice!

uberhacker:
Please remove the perl update from this PR and submit on a separate PR.  You will get more credit from Github if you do it this way. :)

vincowl:
> To be clear, it is set explicitly since `config.guess` will set build to `aarch64-cros-linux-gnu` initially which sets host and then target to the same value. We want to make sure all of the flags are set to `armv7l-cros-linux-gnuaebihf` on aarch64 architecture.

Indeed that's what I understood the first time. But here is the Manual statement : 

> configure can usually guess the canonical name for the type of system it's running on. To do so it runs a script called config.guess, which infers the name using the uname command or symbols predefined by the C preprocessor.
> Alternately, the user can specify the system type with command line arguments to configure (see System Type. Doing so is necessary when cross-compiling. In the most complex case of cross-compiling, three system types are involved. The options to specify them are:
> --build=build-type
> the type of system on which the package is being configured and compiled. It defaults to the result of running config.guess.
> --host=host-type
> the type of system on which the package runs. By default it is the same as the build machine. Specifying it enables the cross-compilation mode.
> --target=target-type
> the type of system for which any compiler tools in the package produce code (rarely needed). By default, it is the same as host.

It's been a long time since I haven't cross-compiled, but to me you should manually set --build and --host. This will skip the config.guess step and set --target to the same value as --host

vincowl:
@uberhacker Please don't merge this till we didn't update perl package

uberhacker:
Gotcha

packages/imagemagick7.rb
```diff
@@ -51,19 +45,20 @@ def self.patch
...
-           --enable-hdri \
-           --with-perl \
-           --with-rsvg \
-           --with-x"
-    system 'make'
+    system "./configure",
+           "CFLAGS=-I#{CREW_PREFIX}/include/gdk-pixbuf-2.0 -I#{CREW_PREFIX}/include/c++/v1/support/xlocale",
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_LIB_PREFIX}",
+           "--mandir=#{CREW_MAN_PREFIX}",
```
uberhacker, line 52: Does `CREW_OPTIONS` work here?

vincowl: No. It works if you remove `--target` from `#{CREW_OPTIONS}`. Don't know why....

PR #4293, opened 2020-09-09
Update kotlin from 1.2.21 to 1.4.0

cstrouse:
Tested on x86_64.


uberhacker:
Thank you for the update @cstrouse.  Guess it's been awhile for this package.

PR #4294, opened 2020-09-10
Update: git 2.26.0 -> 2.28.0

ThatGeekyWeeb:
Updated & no longer builds with a single core, tested and works.

### DIFF
--- a/packages/git.rb
+++ b/packages/git.rb
@@ -5,40 +5,18 @@ class Git < Package
   homepage 'https://git-scm.com/'
   version '2.26.0'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


uberhacker:
Thank you for the update @ThatGeekyWeeb!  Since this is a core package, we'll need binaries and install.sh will need to be updated for the new version.  I can do this since I have the ability to generate binaries for all architectures.  I'll submit a new PR and close this PR.

ThatGeekyWeeb:
Understood! :+1:

uberhacker:
Binaries added in #4312.

PR #4299, opened 2020-09-11
[imagemagick6] Update to 6.9.11-29

vincowl:


packages/imagemagick6.rb
```diff
@@ -63,7 +58,7 @@ def self.build
            --with-perl \
            --with-rsvg \
            --with-x"
-    system 'make'
+    system "make -j#{CREW_NPROC}"
```
uberhacker, line 61: This is not necessary.  The default value of `-j` is set to this value.  See `lib/package.rb`.

vincowl: Fine ! I'll remove it from imagemagick too then.

packages/imagemagick6.rb
```diff
@@ -55,6 +49,7 @@ def self.build
            ./configure \
            --prefix=#{CREW_PREFIX} \
            --libdir=#{CREW_LIB_PREFIX} \
+           --mandir=#{CREW_MAN_PREFIX} \
```
uberhacker, line 52: Does `CREW_OPTIONS` work here?

vincowl: No. It works if you remove `--target` from `#{CREW_OPTIONS}`. Don't know why....

PR #4300, opened 2020-09-11
[perl] Update to 5.32.0

vincowl:


packages/perl.rb
```diff
@@ -35,8 +28,9 @@ def self.build
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
-    system "mkdir -p #{CREW_DEST_PREFIX}/bin/"
-    system "cp cpanm #{CREW_DEST_PREFIX}/bin/"
+    FileUtils.ln_sf "#{CREW_LIB_PREFIX}/libnsl.so.1", "#{CREW_LIB_PREFIX}/libnsl.so"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin/"
```
uberhacker, line 32: This line is not necessary with the line below.  `install` will create the directory.

uberhacker:
Thank you for the update @vincowl!  Since this is a core package, I will need to generate the binaries.  On what architecture(s) have you tested this update?

vincowl:
I tested it on a x86_64 architecture. Managed to compile and install it, use it to compile imagemagick7 and run a simple helloworld test file since I don't know perl and have not installed other libs to run a more complex sample

vincowl:
Since you'll have to generate binaries, maybe it will be a good opportunity to try the --build and --host settings only and see if it works. Then we could simplify imagemagick6 & 7 configure step using `#{CREW_OPTIONS}` if it works

uberhacker:
Binaries added in #4310.

PR #4301, opened 2020-09-12
[Qt] Update packages to 5.15.1

vincowl:


packages/qttools.rb
```diff
@@ -36,7 +28,6 @@ def self.install
     FileUtils.cp_r 'examples', "#{CREW_DEST_PREFIX}/share/Qt-5"
     FileUtils.cp_r 'include', "#{CREW_DEST_PREFIX}/share/Qt-5"
     FileUtils.cp_r 'mkspecs', "#{CREW_DEST_PREFIX}/share/Qt-5"
-    FileUtils.cp_r 'plugins', "#{CREW_DEST_PREFIX}/share/Qt-5"
```
uberhacker, line 39: Why no plugins?

packages/qtx11extras.rb
```diff
@@ -28,8 +20,6 @@ def self.build
   end
 
   def self.install
-    FileUtils.mkdir_p "#{CREW_DEST_LIB_PREFIX}"
-    system "cp -a lib/* #{CREW_DEST_LIB_PREFIX}"
```
uberhacker, line 32: Why no libraries?

packages/qtbase.rb
```diff
@@ -50,7 +44,8 @@ def self.build
            '-system-zlib',
            '-system-libpng',
            '-system-libjpeg',
-           '-system-freetype'
+           '-system-freetype',
+           '-no-openssl'
```
uberhacker, line 48: I'm hesitant to remove openssl.  I'm sure we can figure out a way to include it either by updating libressl or reverting back to openssl.

packages/qttools.rb
vincowl: This folder doesn't exist anymore. It is not part of the tar file and not produced by the compilation step

packages/qtx11extras.rb
vincowl: It is not part of the tar file and not produced by the compilation step

packages/qtbase.rb
vincowl: Totally agree. But I had to remove it to make the other thing work. Updating libressl or something like it is a huge work due to its impact on the chromebrew basis. As said in a related issue, I chose to remove it in a first step, then state on libressl vs openssl and their updates and finally reactivate it in qtbase

uberhacker: Fair enough.  I'm glad you were able to compile these qt packages successfully because I couldn't do it.

PR #4302, opened 2020-09-12
[Qtfm] Update to 6.2.0

vincowl:


packages/qtfm.rb
```diff
@@ -3,29 +3,26 @@
...
   })
 
   depends_on 'qtbase'
   depends_on 'imagemagick7'
   depends_on 'ffmpeg'
   depends_on 'sommelier'
 
+  def self.patch
+    system "sed -i 's/#include <QPainter>/#include <QPainter>\\n#include <QPainterPath>/g' libfm/iconlist.h"
+    system "sed -i 's/#include <QPainter>/#include <QPainter>\\n#include <QPainterPath>/g' libfm/iconview.h"
```
uberhacker, line 23: You could simplify like below:
```
system "sed -i '12i#include <QPainterPath>' libfm/iconlist.h"
system "sed -i '14i#include <QPainterPath>' libfm/iconlist.h"
```

vincowl: Tested : it works

PR #4304, opened 2020-09-12
[Poppler] Update to 20.08.0

vincowl:


uberhacker:
Wow, quite a version jump!

vincowl:
Huuu.... Merging it will create troubles since you don't merge qt too

PR #4305, opened 2020-09-12
Update: gtk_doc

ThatGeekyWeeb:
I hate to open a WIP PR, but I am having a *very* hard time getting things to work, so, I'd love some help!
***
What's being done:
`gtk_doc` is going from `1.29` to `1.32`
Which requires `DocBook XML DTD V4.3`
> Our version of `docbook_xml` is too new

So a new package for `docbook_xml`(`43`) version 4.3 needs to be made, I did that, but, compiling for `gtk_doc` still fails-
~
Additionally: 
`xmlcatmgr` has been added and will ease the editing of `catalog.xml`\
And there are **alot** of patches that need to be applied to `gtk_doc`, and the way I have it set up to grab and apply these patches isn't ideal.
***
For reference, here is `xpbs-src`'s template for `docbook-xml` which installs `4.5`, `4.4`, `4.3`, and `4.2`
> https://github.com/void-linux/void-packages/blob/master/srcpkgs/docbook-xml/template

### DIFF
--- a/packages/docbook_xml43.rb
+++ b/packages/docbook_xml43.rb
@@ -1,0 +1,29 @@
+require 'package'
+
+class Docbook_xml43 < Package
+  description 'document type definitions for verification of XML data files against the DocBook rule set'
+  compatibility 'all'
+  homepage 'http://www.docbook.org'
+  version '4.3'
+  source_url 'http://www.oasis-open.org/docbook/xml/4.3/docbook-xml-4.3.zip'
+  source_sha256 '23068a94ea6fd484b004c5a73ec36a66aa47ea8f0d6b62cc1695931f5c143464'
+
+  depends_on 'docbook'
+  depends_on 'xmlcatmgr'
+  depends_on 'docbook_xml'
+  depends_on 'docbook_xsl'
+  
+  def self.prebuild
+    system "sed -i -e 's,<!-- .* -->,,g' #{CREW_PREFIX}/etc/xml/catalog.xml"
+    system "cat << EOF > ./remove_add.sh
+xmlcatmgr -c #{CREW_PREFIX}/xml/docbook remove public '-//OASIS//DTD DocBook XML V4.3//EN'
+xmlcatmgr -c #{CREW_PREFIX}/etc/xml/docbook add public '-//OASIS//DTD DocBook XML V4.3//EN' '#{CREW_PREFIX}/share/xml/docbook/4.3/catalog.xml'
+EOF"
+    system "bash ./remove_add.sh"
+  end
+  def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/xml/docbook/4.3/"
+    FileUtils.cp_r Dir.glob('*'), "#{CREW_DEST_PREFIX}/share/xml/docbook/4.3/"
+  end
+end
+
--- a/packages/gtk_doc.rb
+++ b/packages/gtk_doc.rb
@@ -1,38 +1,51 @@
 require 'package'
 
+# There are alot of patches to grab here ~ They've all come from Void-Packages (xbps-src)
+
 class Gtk_doc < Package
-  description 'GTK-Doc is a project which was started to generate API documentation from comments added to C code.'
-  homepage 'https://www.gtk.org/gtk-doc'
-  version '1.29'
+  description 'Documentation tool for public library API'
+  homepage 'https://www.gtk.org/gtk-doc/'
+  version '1.32'
   compatibility 'all'
-  source_url 'https://github.com/GNOME/gtk-doc/archive/GTK_DOC_1_29.tar.gz'
-  source_sha256 'fdf5051e1f164fc1454a4530d217ee2c09dcc2c8e42b93cd5d68645493319ce5'
+  source_url 'https://ftp.gnome.org/pub/GNOME/sources/gtk-doc/1.32/gtk-doc-1.32.tar.xz'
+  source_sha256 'de0ef034fb17cb21ab0c635ec730d19746bce52984a6706e7bbec6fb5e0b907c'
 
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gtk_doc-1.29-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/gtk_doc-1.29-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/gtk_doc-1.29-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/gtk_doc-1.29-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: '8641b7d74ba0de51e40347751c5022164f490703fcab9902d85843a3c6a8df41',
-     armv7l: '8641b7d74ba0de51e40347751c5022164f490703fcab9902d85843a3c6a8df41',
-       i686: '6c14eadd88bb454b2bdd1212935095d6fe478e932bc6bab301d519e767217c2f',
-     x86_64: '53adc04a62779005099e257fc2ba6df76c0b8a9dce8cc31bd6eac6f4181b69c4',
-  })
-
-  depends_on 'six'
+  depends_on 'docbook_xml'
+  depends_on 'docbook_xsl'
+  depends_on 'itstool'
   depends_on 'libxslt'
+  depends_on 'docbook_xml43'
 
-  def self.build
-    system './autogen.sh',
-           "--prefix=#{CREW_PREFIX}",
-           "--libdir=#{CREW_LIB_PREFIX}",
-           "--with-xml-catalog=#{CREW_PREFIX}/etc/xml/catalog"
-    system 'make'
+  def self.patch
+    puts
+    puts 'Grabbing patches'.lightblue
+    system 'curl --ssl -L -o "output-reproducible.patch" "https://git.io/JUlWD" --progress-bar'
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('output-reproducible.patch') ) == '701fe4124a94b124e943e9220f4ea42de7fe191379f41c349cc2f69ece37af24'
+    system 'curl --ssl -L -o "tree-structure-without-using-anytree.patch" "https://git.io/JUlOD" --progress-bar'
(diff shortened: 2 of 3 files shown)


uberhacker:
No problem.  Just remove the `[WIP]` when you are ready.

vincowl:
@ThatGeekyWeeb I've just tried to compile gtk_doc you provided, but ended with an error with docbook_xml43 : 
```
docbook_xml43 failed to install: `xmlcatmgr -c /usr/local/etc/xml/catalog.xml add rewriteSystem http://www.oasis-open.org/docbook/xml/4.3 /usr/local/share/xml/docbook/xml-dtd-4.3` exited with 127
```

ThatGeekyWeeb:
@vincowl, irrc exit code `127` means that the command doesn't exist, which means `xmlcatmgr` is missing. Install `xmlcatmgr` from this PR. If you already have, then I'll have to do some debugging.

vincowl:
Indeed, I was trusting the dependency toolchain.... Bad idea ! OK then. I successfully installed docbook_xml43 but configure step of gtk_doc fails with : 
```
checking for XML catalog... /usr/local/etc/xml/catalog.xml
checking for xmlcatalog... /usr/local/bin/xmlcatalog
checking for DocBook XML DTD V4.3 in XML catalog... not found
configure: error: could not find DocBook XML DTD V4.3 in XML catalog
```
Are you facing the same error ?

ThatGeekyWeeb:
@vincowl, yes that is the same issue I am having.

ThatGeekyWeeb:
I figured out that `./configure` runs `xmlcatalog /usr/local/etc/xml/catalog.xml -//OASIS//DTD DocBook XML V4.3//EN` to check for `DocBook XML DTD V4.3`, if the command does not exit with code 0, then it fails, I'll have to do some debugging

ThatGeekyWeeb:
I found that `catalog.xml` contains 
```xml
<delegatePublic publicIdStartString="-//OASIS//ENTITIES DocBook XML" catalog="file:///usr/local/etc/xml/docbook"/>
```
Which means `xmlcatalog` redirects `-//OASIS//DTD DocBook XML*` searches to `/usr/local/etc/xml/docbook`, and then checks for a result.
I added 
```xml
  <public publicId="-//OASIS//DTD DocBook XML V4.3//EN" uri="file:///usr/local/share/xml/dockbook/4.3/catalog.xml"/>
```
into `/usr/local/etc/xml/docbook` and `xmlcatalog /usr/local/etc/xml/catalog.xml "-//OASIS//DTD DocBook XML V4.3//EN"` run successfully! I'll push a fix soon!

ThatGeekyWeeb:
Please reinstall `docbook_xsl` before reinstalling `docbook_xml43`

ThatGeekyWeeb:
Having another package for `docbook-xml` v4.3 is bad, and `docbook_xml` should be edited to install 5.0, 5.1, 5.2, 4.5, 4.4, 4.3, and 4.2.
*** 
Additionally `docbook_xml` needs to be updated to 5.2

ThatGeekyWeeb:
Additionally `gtk_doc` compiles fine now!

vincowl:
Great !!

While php concurrent versions is bad, why would it be a problem to have several xml DTDs available ? Just be sure to provide access to both 4.3 and 5.0. It's always a pain to have to stick to an old version due to strong dependency to a slow development package.

If those are mutually exclusive, then have a look at php73, php74 and the upper layer php.

packages/docbook_xml43.rb
```diff
@@ -0,0 +1,26 @@
...
+  source_url 'http://www.oasis-open.org/docbook/xml/4.3/docbook-xml-4.3.zip'
+  source_sha256 '23068a94ea6fd484b004c5a73ec36a66aa47ea8f0d6b62cc1695931f5c143464'
+
+  depends_on 'docbook'
+  depends_on 'xmlcatmgr'
+  depends_on 'docbook_xml'
+  depends_on 'docbook_xsl' # Requires the catalog.xml created within this package
+  
+  def self.prebuild
+    system "sed -i -e 's,<!-- .* -->,,g' /usr/local/etc/xml/catalog.xml"
```
vincowl, line 17: Use `#{CREW_PREFIX}` instead of `/usr/local`

packages/docbook_xml43.rb
```diff
@@ -0,0 +1,26 @@
...
+    system 'xmlcatmgr -c /usr/local/etc/xml/docbook add public "-//OASIS//DTD DocBook XML V4.3//EN" "/usr/local/share/xml/docbook/4.3/catalog.xml"'
```
vincowl, line 18: Use `#{CREW_PREFIX}` instead of `/usr/local`

ThatGeekyWeeb:
I was planning on turning `docbook_xml` into a `is_fake` package that installs 4.2, 4.2, 4.3, 4.5, 5.0, and 5.1, since that's what most distros seem to do.

packages/gtk_doc.rb
```diff
@@ -1,38 +1,51 @@
...
+    puts
+  end
   def self.build
-    system './autogen.sh',
-           "--prefix=#{CREW_PREFIX}",
-           "--libdir=#{CREW_LIB_PREFIX}",
-           "--with-xml-catalog=#{CREW_PREFIX}/etc/xml/catalog"
-    system 'make'
+      system "./configure #{CREW_OPTIONS} --with-xml-catalog=#{CREW_PREFIX}/etc/xml/catalog.xml"
+      system "make -j#{CREW_NPROC}"
```
vincowl, line 46: `-j#{CREW_NPROC}` is useless as this value is already defined by default in package.rb. Can become `system 'make'`

packages/xmlcatmgr.rb
```diff
@@ -0,0 +1,19 @@
...
+  homepage 'https://xmlcatmgr.sourceforge.net'
+  version '2.2'
+  compatibility 'all'
+  source_url 'https://downloads.sourceforge.net/sourceforge/xmlcatmgr/xmlcatmgr-2.2.tar.gz'
+  source_sha256 'ea1142b6aef40fbd624fc3e2130cf10cf081b5fa88e5229c92b8f515779d6fdc'
+
+
+  def self.build
+      system "./configure #{CREW_OPTIONS} "
+      system "make -j#{CREW_NPROC}"
```
vincowl, line 14: `-j#{CREW_NPROC}` is useless as this value is already defined by default in package.rb

vincowl:
Sounds great to me 👍

packages/docbook_xml43.rb
```diff
@@ -0,0 +1,26 @@
...
+  source_sha256 '23068a94ea6fd484b004c5a73ec36a66aa47ea8f0d6b62cc1695931f5c143464'
+
+  depends_on 'docbook'
+  depends_on 'xmlcatmgr'
+  depends_on 'docbook_xml'
+  depends_on 'docbook_xsl' # Requires the catalog.xml created within this package
+  
+  def self.prebuild
+    system "sed -i -e 's,<!-- .* -->,,g' #{CREW_PREFIX}/etc/xml/catalog.xml"
+    system "xmlcatmgr -c #{CREW_PREFIX}/etc/xml/docbook add public '-//OASIS//DTD DocBook XML V4.3//EN' '#{CREW_PREFIX}/share/xml/docbook/4.3/catalog.xml'"
```
uberhacker, line 18: Does this command generate an output file?  If so, should this be an install step?

ThatGeekyWeeb: It edits a file installed by `docbook_xml`
See #4315

packages/gtk_doc.rb
```diff
@@ -1,38 +1,51 @@
...
+  depends_on 'docbook_xml'
+  depends_on 'docbook_xsl'
+  depends_on 'itstool'
   depends_on 'libxslt'
+  depends_on 'docbook_xml43'
 
+  def self.patch
+    puts
+    puts 'Grabbing patches'.lightblue
+    system 'curl --ssl --insecure -L -o "output-reproducible.patch" "https://git.io/JUlWD" --progress-bar'
```
uberhacker, line 22: Why do we need the `--insecure` option for https?

ThatGeekyWeeb: We don't idk exactly why I have that there, I'll remove that.

packages/docbook_xml43.rb
uberhacker: Should this command be executed in `def self.preinstall` instead?

ThatGeekyWeeb: Yes, that would be a better place, as nothing is actually built.

ThatGeekyWeeb:
reinstallation of `docbook_xml43` has a 100% chance to fail, as a file provided by another package it edited, I've added a bash script, that is both made and used inside `docbook_xml43.rb`, that will allow it to reinstall.
***
> Another idea would be to add a way for commands to fail, only if allowed.
> 
> ```
> xmlcatmgr -c /usr/local/etc/xml/docbook remove public "-//OASIS//DTD DocBook XML V4.3//EN"
> ```
> 
> Fails unless `docbook_xml43` has been installed once before, but that's fine, as long as
> 
> ```
> xmlcatmgr -c #{CREW_PREFIX}/etc/xml/docbook add public '-//OASIS//DTD DocBook XML V4.3//EN' '#{CREW_PREFIX}/share/xml/docbook/4.3/catalog.xml'"
> ```
> 
> runs afterwards its fine. But due to how `crew` works, it end as soon as one command fails.
> ***
> A final option is to use a bash script, which would execute both of them, and exit with `0` no matter what.

Please see #4315 for more info.\
~\
`/usr/local` must be used, as it is writing to a bash script.

packages/docbook_xml43.rb
```diff
@@ -0,0 +1,30 @@
...
+  version '4.3'
+  source_url 'http://www.oasis-open.org/docbook/xml/4.3/docbook-xml-4.3.zip'
+  source_sha256 '23068a94ea6fd484b004c5a73ec36a66aa47ea8f0d6b62cc1695931f5c143464'
+
+  depends_on 'docbook'
+  depends_on 'xmlcatmgr'
+  depends_on 'docbook_xml'
+  depends_on 'docbook_xsl' # Requires the catalog.xml created within this package
+  
+  def self.preinstall
```
uberhacker, line 16: Should this be `def self.prebuild`?

packages/docbook_xml43.rb
```diff
@@ -0,0 +1,30 @@
...
+    system "sed -i -e 's,<!-- .* -->,,g' #{CREW_PREFIX}/etc/xml/catalog.xml"
+    system "cat << EOF > ./remove_add.sh
+xmlcatmgr -c /usr/local/xml/docbook remove public '-//OASIS//DTD DocBook XML V4.3//EN' '/usr/local/share/xml/docbook/4.3/catalog.xml'
+xmlcatmgr -c /usr/local/etc/xml/docbook add public '-//OASIS//DTD DocBook XML V4.3//EN' '/usr/local/share/xml/docbook/4.3/catalog.xml'
```
uberhacker, line 20: Please abstract `/usr/local` with `#{CREW_PREFIX}`.

uberhacker:
Nice work!

PR #4307, opened 2020-09-12
[sass] Add new package

vincowl:
Ruby' sass gem is now obsolete and has been migrated to dart-sass. This package solves this

### DIFF
--- a/packages/sass.rb
+++ b/packages/sass.rb
@@ -1,0 +1,28 @@
+require 'package'
+
+class Sass < Package
+  description 'A Dart implementation of Sass. Sass makes CSS fun again.'
+  homepage 'https://sass-lang.com/'
+  version '1.26.10'
+  compatibility 'all'
+  source_url 'https://github.com/sass/dart-sass/archive/1.26.10.tar.gz'
+  source_sha256 '9131be72e3eb6d32265e354d6da9f9162d9b28ed5b4910f03744c31509be894f'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'dart'
+
+  def self.install
+    system 'pub get'
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/dart-sass"
+    FileUtils.cp_r '.', "#{CREW_DEST_PREFIX}/share/dart-sass"
+    system "echo '#!/bin/bash' >> sass" 
+    system "echo 'dart #{CREW_PREFIX}/share/dart-sass/bin/sass.dart' >> sass"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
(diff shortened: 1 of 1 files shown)


packages/sass.rb
```diff
@@ -0,0 +1,29 @@
...
+  depends_on 'dart'
+
+  def self.install
+    system 'pub get'
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/dart-sass"
+    FileUtils.cp_r '.', "#{CREW_DEST_PREFIX}/share/dart-sass"
+    system "echo '#!/bin/bash' >> sass" 
+    system "echo 'dart #{CREW_PREFIX}/share/dart-sass/bin/sass.dart $@' >> sass"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
+    FileUtils.cp_r 'sass', "#{CREW_DEST_PREFIX}/bin/."
```
uberhacker, line 25: Better use `system "install -Dm755 'sass', #{CREW_DEST_PREFIX}/bin/sass"` here instead to make the script executable.  You also can remove line 24 above.

packages/sass.rb
```diff
@@ -0,0 +1,29 @@
...
+  })
+
+  depends_on 'dart'
+
+  def self.install
+    system 'pub get'
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/dart-sass"
+    FileUtils.cp_r '.', "#{CREW_DEST_PREFIX}/share/dart-sass"
+    system "echo '#!/bin/bash' >> sass" 
+    system "echo 'dart #{CREW_PREFIX}/share/dart-sass/bin/sass.dart $@' >> sass"
```
uberhacker, line 23: This could be a build step.

vincowl: If I put lines 22/23 in a build step and keep the new install line (equivalent to lines 24-26) in an install step, won't we face troubles once binaries will be produced and we then won't use the build step ? install command will not find the sass script.

Putting all this in a build step doesn't work. After install sass is not found in /usr/local/bin

packages/sass.rb
```diff
@@ -0,0 +1,30 @@
...
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
+
+  depends_on 'dart'
+
+  def self.build
+    system "echo '#!/bin/bash' >> sass" 
+    system "echo 'dart #{CREW_PREFIX}/share/dart-sass/bin/sass.dart $@' >> sass"
```
uberhacker, line 20: Does the `$` need escaped?  What are the contents of sass after install?

vincowl: After installing, here is the content of `sass`. Seems it doesn't need to escape $
```
$ cat /usr/local/bin/sass 
#!/bin/bash
dart /usr/local/share/dart-sass/bin/sass.dart $@
```

Issue #4315, opened 2020-09-13
[Request] Add `self.remove` function


ThatGeekyWeeb:
I suggest adding a `self.remove` function that would be executed upon removal. This would allow packages to undo commands that cannot be undone by simply removing the file. 
***
`docbook_xml43` uses `xmlcatmgr` to edit `/usr/local/etc/xml/docbook`, which is installed by `docbook_xml`, with the way `xmlcatmgr` works, recompiling `docbook_xml43` can't be done as it adds text to `/usr/local/etc/xml/catalog.xml` which also needs to be removed by `xmlcatmgr`.
***
> ```ruby
> # NOTE: run  `xmlcatmgr -c /usr/local/etc/xml/docbook remove public "-//OASIS//DTD DocBook XML V4.3//EN"` before reinstall
> ```
***
One could just copy, edit and reinstall the edited file, but that would then require `docbook_xml` to be reinstalled as well.

uberhacker:
This is a great idea!  Several packages have postinstall removal instructions that could be added here instead.  In most cases, we should try to include the most tracked files as possible.  If files are generated from the package commands, sometimes just including a simple `touch #{CREW_DEST_PREFIX}/path/to/file` in the install section will track it in the package so it is deleted after `crew remove ...`.

ThatGeekyWeeb:
The current removal system is great but adding to files installed by another package causes issues, as these commands have to be undone to reinstall the package, otherwise, reinstallation fails.
***
This can be seen if one tries to reinstall `docbook_xml43` without running 
> `xmlcatmgr -c /usr/local/etc/xml/docbook remove public "-//OASIS//DTD DocBook XML V4.3//EN"`

As `/usr/local/etc/xml/docbook` is provided by `docbook_xml`, but a edit needs to be made to link `docbook_xml` v4.3

uberhacker:
Maybe the `xmlcatmgr` command should be executed from `docbook_xml43` instead?

ThatGeekyWeeb:
running `xmlcatmgr -c /usr/local/etc/xml/docbook remove public "-//OASIS//DTD DocBook XML V4.3//EN"` before `docbook_xml43` has been installed, would cause compilation to fail.

ThatGeekyWeeb:
Another idea would be to add a way for commands to fail, only if allowed.
```
xmlcatmgr -c /usr/local/etc/xml/docbook remove public "-//OASIS//DTD DocBook XML V4.3//EN"
[...]

PR #4316, opened 2020-09-13
Add all versions of docbook_xml

ThatGeekyWeeb:
Depends on `docbook_xml43` and `xmlcatmgr` from #4305 
Installs all versions of `docbook_xml`, `docbook_xsl` has been edited to use `xmlcatmgr` instead of a `echo` command, both `/usr/local/etc/catalog` & `/usr/local/etc/catalog.xml` are used now. 
***
> I was planning on turning `docbook_xml` into a `is_fake` package that installs 4.2, 4.2, 4.3, 4.5, 5.0, and 5.1, since that's what most distros seem to do. ~ After the PR gets merged that is, having `xmlcatmgr` makes things super easy!
> ***
> > For reference, here is xpbs-src's template for docbook-xml which installs 4.5, 4.4, 4.3, and 4.2
> > > https://github.com/void-linux/void-packages/blob/master/srcpkgs/docbook-xml/template
> ***
> It's easy to support multiple versions as the source is simply a `.zip` with files that need to be copied to `/usr/share/xml/docbook/v*/`, and an XML entry to link them.
> ***
> > ```ruby
> > def self.install
> >    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/xml/docbook/4.3/"
> >    FileUtils.cp_r Dir.glob('*'), "#{CREW_DEST_PREFIX}/share/xml/docbook/4.3/"
> >  end
> > ```
***
Instances of `/usr/local` are required as they are being redirected into a bash script please see #4315 for more info

### DIFF
--- a/packages/docbook_xml.rb
+++ b/packages/docbook_xml.rb
@@ -1,103 +1,18 @@
 require 'package'
 
-# from LFS: http://www.linuxfromscratch.org/blfs/view/cvs/pst/docbook.html
-
 class Docbook_xml < Package
-  description 'document type definitions for verification of XML data files against the DocBook rule set'
+  description 'Meta package for all versions of docbook_xml'
   compatibility 'all'
   homepage 'http://www.docbook.org'
-  version '5.1'
-  source_url 'https://docbook.org/xml/5.1/docbook-v5.1-os.zip'
-  source_sha256 'b3f3413654003c1e773360d7fc60ebb8abd0e8c9af8e7d6c4b55f124f34d1e7f'
+  version '5.1-1'
 
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/docbook_xml-5.1-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/docbook_xml-5.1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/docbook_xml-5.1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/docbook_xml-5.1-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: '9f3ef8937e0b9f16158d66fe97fea777065790f9999e24e4dc295f461efe9b4b',
-     armv7l: '9f3ef8937e0b9f16158d66fe97fea777065790f9999e24e4dc295f461efe9b4b',
-       i686: '118a652d4b192525f2400ec121747b1824292b874941b04da037b4a031107148',
-     x86_64: '19aefa5a44bdd6a0ff77072f2fe45e1b032a06d9a72faeb51ac3cae2d49e992c',
-  })
+  depends_on 'docbook_xml51'
+  depends_on 'docbook_xml50'
+  depends_on 'docbook_xml45'
+  depends_on 'docbook_xml44'
+  depends_on 'docbook_xml43'
+  depends_on 'docbook_xml42'
+  
+  is_fake
 
-  depends_on 'docbook'
-  depends_on 'sgml_common'
-
-  def self.install
-
-    xml_version = '5.1'
-    xml_dtd = "xml-dtd-#{xml_version}"
-
-    system "install -v -d -m755 #{CREW_DEST_PREFIX}/share/xml/docbook/#{xml_dtd}"
-    system "install -v -d -m755 #{CREW_DEST_PREFIX}/etc/xml"
-    system "cp -rpa . #{CREW_DEST_PREFIX}/share/xml/docbook/#{xml_dtd}/"
-    system "rm -f #{CREW_PREFIX}/etc/xml/docbook && \
-                xmlcatalog --noout --create #{CREW_DEST_PREFIX}/etc/xml/docbook && \
-            xmlcatalog --noout --add 'public' \
-                '-//OASIS//DTD DocBook XML V#{xml_version}//EN' \
-                'http://www.oasis-open.org/docbook/xml/#{xml_version}/docbookx.dtd' \
-                #{CREW_DEST_PREFIX}/etc/xml/docbook &&
-            xmlcatalog --noout --add 'public' \
-                '-//OASIS//DTD DocBook XML CALS Table Model V#{xml_version}//EN' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}/calstblx.dtd' \
-                #{CREW_DEST_PREFIX}/etc/xml/docbook &&
-            xmlcatalog --noout --add 'public' \
-                '-//OASIS//DTD XML Exchange Table Model 19990315//EN' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}/soextblx.dtd' \
-                #{CREW_DEST_PREFIX}/etc/xml/docbook &&
-            xmlcatalog --noout --add 'public' \
-                '-//OASIS//ELEMENTS DocBook XML Information Pool V#{xml_version}//EN' \
-                'file://#{CREW_PREFIX}/share/xml/docbook/#{xml_dtd}/dbpoolx.mod' \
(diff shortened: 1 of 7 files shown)


packages/docbook_xml.rb
```diff
@@ -1,103 +1,21 @@
...
-  binary_sha256 ({
-    aarch64: '9f3ef8937e0b9f16158d66fe97fea777065790f9999e24e4dc295f461efe9b4b',
-     armv7l: '9f3ef8937e0b9f16158d66fe97fea777065790f9999e24e4dc295f461efe9b4b',
-       i686: '118a652d4b192525f2400ec121747b1824292b874941b04da037b4a031107148',
-     x86_64: '19aefa5a44bdd6a0ff77072f2fe45e1b032a06d9a72faeb51ac3cae2d49e992c',
-  })
 
+  depends_on 'xmlcatmgr' # Used to edit catalog.xml
   depends_on 'docbook'
   depends_on 'sgml_common'
```
uberhacker, line 11: The 3 dependencies above are not needed since they are already included in the dependencies below.

packages/docbook_xml42.rb
```diff
@@ -0,0 +1,32 @@
...
+  source_url 'http://www.oasis-open.org/docbook/xml/4.2/docbook-xml-4.2.zip'
+  source_sha256 'acc4601e4f97a196076b7e64b368d9248b07c7abf26b34a02cca40eeebe60fa2'
+
+  depends_on 'docbook'
+  depends_on 'xmlcatmgr'
+  depends_on 'docbook_xml'
+  depends_on 'docbook_xsl' # Requires the catalog.xml created within this package
+  
+  def self.preinstall
+    system "sed -i -e 's,<!-- .* -->,,g' #{CREW_PREFIX}/etc/xml/catalog.xml"
```
uberhacker, line 17: This line should go in `def self.prebuild` unless it needs to run for every install.

packages/docbook_xml42.rb
```diff
@@ -0,0 +1,32 @@
...
+class Docbook_xml42 < Package
+  description 'Version 4.2 - document type definitions for verification of XML data files against the DocBook rule set'
+  compatibility 'all'
+  homepage 'http://www.docbook.org'
+  version '4.2'
+  source_url 'http://www.oasis-open.org/docbook/xml/4.2/docbook-xml-4.2.zip'
+  source_sha256 'acc4601e4f97a196076b7e64b368d9248b07c7abf26b34a02cca40eeebe60fa2'
+
+  depends_on 'docbook'
+  depends_on 'xmlcatmgr'
```
uberhacker, line 12: These dependencies are already included in docbook_xsl.

packages/docbook_xml42.rb
```diff
@@ -0,0 +1,32 @@
...
+  depends_on 'docbook_xml'
```
uberhacker, line 13: This dependency is not needed since it's now a fake package.

packages/docbook_xml42.rb
```diff
@@ -0,0 +1,32 @@
...
+  def self.preinstall
+    system "sed -i -e 's,<!-- .* -->,,g' #{CREW_PREFIX}/etc/xml/catalog.xml"
+    system "cat << EOF > ./remove_add.sh
+xmlcatmgr -c /usr/local/etc/xml/docbook remove public '-//OASIS//DTD DocBook XML V4.2//EN' '/usr/local/share/xml/docbook/4.2/catalog.xml'
+xmlcatmgr -c /usr/local/etc/xml/docbook remove public '-//OASIS//DTD DocBook XML V4.1.2//EN' '/usr/local/share/xml/docbook/4.2/catalog.xml'
+xmlcatmgr -c /usr/local/etc/xml/docbook add public '-//OASIS//DTD DocBook XML V4.2//EN' '/usr/local/share/xml/docbook/4.2/catalog.xml'
+xmlcatmgr -c /usr/local/etc/xml/docbook add public '-//OASIS//DTD DocBook XML V4.1.2//EN' '/usr/local/share/xml/docbook/4.2/catalog.xml'
+
+EOF"
+    system "bash ./remove_add.sh"
```
uberhacker, line 25: I'm not following.  Why do we need to remove and then add?  And why does it need to be a script?

uberhacker:
Thank you for the update @ThatGeekyWeeb!  It looks like you put a lot of effort into this and your work is appreciated.  Please see the comments.

packages/docbook_xml42.rb
ThatGeekyWeeb: We're not installing a file here, but editing on that's already been installed. The issue is, upon removal, the edited line is still there, and will cause compilation to fail upon reinstallation. In #4315 I mentioned how a simple fix would be to use a bash script, that's what I've done here.

packages/docbook_xml42.rb
ThatGeekyWeeb: Older version of `/usr/local/etc/xml/catalog.xml` have an HTML comment within them, if this line is present, it causes `xmlcatmgr` to fail.

ThatGeekyWeeb:
Due to the nature of `xmlcatmgr`, `/usr/local/etc/xml/catalog.xml` has to exist beforehand, since `docbook_xml51` is the only package which creates it, each version depends on the fact that it has been installed before.

packages/docbook_xml42.rb
ThatGeekyWeeb: Removing the line can't be done if it's not present, which would cause compilation to fail.
Please read from here to the end.
> https://github.com/skycocker/chromebrew/issues/4315#issuecomment-691685691

uberhacker: You could run the commands explicitly without a script, correct?  Since this is in the prebuild section, it is only run during the build.  Don't you want this to run during install/reinstall?  If that is the case, then these commands should be in the preinstall section instead.

ThatGeekyWeeb: if they were to both be placed in postinstall, after everything has been installed, it would work without a script.

packages/docbook_xml42.rb
```diff
@@ -0,0 +1,30 @@
...
+  depends_on 'docbook_xml51'
+  depends_on 'docbook_xsl' # Requires the catalog.xml created within this package
+  
+  def self.prebuild
+    system "sed -i -e 's,<!-- .* -->,,g' #{CREW_PREFIX}/etc/xml/catalog.xml"
+    system "cat << EOF > ./remove_add.sh
+xmlcatmgr -c /usr/local/etc/xml/docbook remove public '-//OASIS//DTD DocBook XML V4.2//EN' '/usr/local/share/xml/docbook/4.2/catalog.xml'
+xmlcatmgr -c /usr/local/etc/xml/docbook remove public '-//OASIS//DTD DocBook XML V4.1.2//EN' '/usr/local/share/xml/docbook/4.2/catalog.xml'
+xmlcatmgr -c /usr/local/etc/xml/docbook add public '-//OASIS//DTD DocBook XML V4.2//EN' '/usr/local/share/xml/docbook/4.2/catalog.xml'
+xmlcatmgr -c /usr/local/etc/xml/docbook add public '-//OASIS//DTD DocBook XML V4.1.2//EN' '/usr/local/share/xml/docbook/4.2/catalog.xml'
```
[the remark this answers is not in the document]
ThatGeekyWeeb: Since this is writing to a bash script which can't understand `#{CREW_PREFIX}`, it can't be used

packages/docbook_xml42.rb
ThatGeekyWeeb: This is creating a one-time usage script, thas only needed because the removal command can sometimes fail. The command edits a file installed by `docbook_xml51`, it really doesn't matter which section we put them in.

packages/docbook_xml42.rb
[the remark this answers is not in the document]
uberhacker: Have you tried it yet?  I think it will work.

packages/docbook_xml42.rb
uberhacker: OK, so prebuild is the correct location then.

packages/docbook_xml42.rb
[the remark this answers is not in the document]
ThatGeekyWeeb: You are correct, it did work!

PR #4322, opened 2020-09-18
New Package: clutter & cogl

ThatGeekyWeeb:


packages/clutter.rb
```diff
@@ -0,0 +1,36 @@
...
+  depends_on 'cogl'
+  depends_on 'libinput'
+  depends_on 'libevdev'
+  depends_on 'eudev'
+  depends_on 'pygtk'
+  depends_on 'glib'
+
+  def self.patch
+    system "curl --ssl --progress-bar -o clutter-x11-startup-error.patch -L https://git.io/JU0yS"
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('clutter-x11-startup-error.patch') ) == '8370bf0cf624c638edbd309d7dfc3922d726242312d7f217facff69135f56187'
```
uberhacker, line 27: Don't you need to apply the patch also?

uberhacker:
Thank you for the new packages @ThatGeekyWeeb!  These run without sommelier?

packages/clutter.rb
ThatGeekyWeeb: Opps! Looks like I forgot to make templator apply the pathes.

uberhacker: Is templator something that can be added to crew?

ThatGeekyWeeb:
iirc by using the `gdk` backend it only depends on `xdg_base`

ThatGeekyWeeb:
And these are only libraries, no executables are provided.

PR #4349, opened 2020-09-30
Update brotli from 1.0.7 to 1.0.8

cstrouse:
Tested on x86_64. All tests pass.


uberhacker:
Thank you for the update @cstrouse.

PR #4351, opened 2020-10-02
Update Heroku v7.44.0

skiman6010:
The long awaited return of the Heroku updates...


uberhacker:
Yep, it's been awhile.  Hacktoberfest must be your motivation. ;)

PR #4356, opened 2020-10-02
Update firefox from 78.0.2 to 81.0.1

uberhacker:
Tested on x86_64.  Still no sound. :(


skiman6010:
You'll get it fixed one day. :(

PR #4362, opened 2020-10-03
[codium] Update to 1.49.3

vincowl:
Still no new version for ARM archs


ThatGeekyWeeb:
You can case `ARCH` to effect the version variable.
```ruby
case ARCH
  when 'aarch64', 'armv7l'
    version: '1.49.1'
    source_url 'https://github.com/VSCodium/vscodium/releases/download/1.49.1/VSCodium-linux-arm-1.49.1.tar.gz'
    source_sha256 '6534d7b10b13a1effadf408899c90a23a8a499298d686a3e8e74535c186a6b65'
    @arch = 'arm'
when 'x86_64'
    version: '1.49.3'
    source_url 'https://github.com/VSCodium/vscodium/releases/download/1.49.3/VSCodium-linux-x64-1.49.3.tar.gz'
    source_sha256 'd8b71d50d82e8d3151f11604103a7fd3fff3a2dceec64a31e1a3683a9d2fc2a7'
    @arch = 'x64'
```
I'd recommend doing this, otherwise, for arm arches, the version variable would be wrong.

vincowl:
I was indeed implementing the same thing due to the same thougth ! ;-)

packages/codium.rb
```diff
@@ -3,16 +3,17 @@
 class Codium < Package
   description 'VSCodium is Open Source Software Binaries of VSCode with a community-driven default configuration.'
   homepage 'https://vscodium.com/'
-  version '1.49.2'
   compatibility 'aarch64,armv7l,x86_64'
   case ARCH
   when 'aarch64', 'armv7l'
+    version '1.49.1'
```
uberhacker, line 9: `version '1.49.3'`?

ThatGeekyWeeb: arm hasn't been updated, only amd64 has
> https://github.com/skycocker/chromebrew/pull/4362#issuecomment-703130291

uberhacker: Oh, I see from the earlier thread.  This is why modern coders don't have a clue.  Don't release until it works for all architectures.  I would like to hold off on this and give them time to repent from straying off the beaten path.

uberhacker:
Fixed in #4422.  Newer release has arm binary available.

PR #4365, opened 2020-10-05
Update OpenSSH to 8.4p1, fix configure error

satmandu:
## Description
configure was erroring out with this error:

```
checking if cc supports -Werror... yes
./configure: line 5480: syntax error near unexpected token `-pipe'
./configure: line 5480: `       OSSH_CHECK_CFLAG_COMPILE(-pipe)'
```

Modifying the autoconf line to ```autoreconf -fiv``` fixes the issue.

## Addtional information

Works properly:
- [x] x86_64

---


uberhacker:

PR #4366, opened 2020-10-05
Update LibreSSL to 3.1.4 (OpenBSD 6.7 rel branch)

satmandu:
Updates LibreSSL to the OpenBSD 6.7 release branch

## Description

Updating LibreSSL to keep in line with OpenBSD's current release branch.
## Addtional information

Works properly:
- [x] x86_64

---

### DIFF
--- a/packages/libressl.rb
+++ b/packages/libressl.rb
@@ -3,23 +3,11 @@ require 'package'
 class Libressl < Package
   description 'LibreSSL is a version of the TLS/crypto stack forked from OpenSSL in 2014, with goals of modernizing the codebase, improving security, and applying best practice development processes.'
   homepage 'https://www.libressl.org/'
-  version '3.0.2-2'
+  version '3.1.4-1'
   compatibility 'all'
-  source_url 'https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.0.2.tar.gz'
-  source_sha256 'df7b172bf79b957dd27ef36dcaa1fb162562c0e8999e194aa8c1a3df2f15398e'
+  source_url 'https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.1.4.tar.gz'
+  source_sha256 '414c149c9963983f805a081db5bd3aec146b5f82d529bb63875ac941b25dcbb6'
 
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/libressl-3.0.2-2-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


uberhacker:
Thank you for the update @satmandu!  The latest release is https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.2.1.tar.gz.  I'm hesitant to merge since there are several other packages that depend on this.  Do cmake, curl and python3 still work with this update?  This is also a core package so it will need binaries for the install.

satmandu:
The 3.2 branch is part of OpenBSD 6.8, which hasn't been released yet, and is this considered a "Developer version."
See https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.2.0-relnotes.txt

The 3.1 branch is based off of 6.7, which has been released. (I think OpenBSD still uses a 6 month release cadence.)

Testing rebuilding those other packages. (openssh definitely works.)

uberhacker:
Those other packages all have updates, I believe, so we might as well upgrade.  Don't worry about the binaries.  I can generate those for all architectures once we have everything stabilized.

satmandu:
current version of curl builds.

satmandu:
current version of cmake installs and runs

vincowl:
@uberhacker What is blocking this PR ? I am really eager to use this libressl update ;-)

uberhacker:
I'm having trouble with the i686 build.  Once I get that working, then all the other PRs in the queue that depend on this can be merged and the binaries included.  I want to make sure this doesn't break anything, especially with the install.  There is extensive testing since so many other packages depend on this update.  I may decide to drop support for i686 architecture altogether if I do not have success.

vincowl:
Tell us if you need help to test. Just have an x86_64 arch here...

satmandu:
Would a gcc update help?

Here's GCC-10: https://github.com/skycocker/chromebrew/pull/4379

uberhacker:
I doubt anything will help for i686 installs.  I'm probably going to drop support in the near future.  Just too difficult to manage for 0.01% of users.  Older Chromebooks will just have to become paper weights.

satmandu:
Are you compiling on an i686 host or in a VM?

ThatGeekyWeeb:
idk how much this helps, but ik Void Linux supports i686, and uses only libressl. Maybe these can be of use
> https://github.com/void-linux/void-packages/tree/master/srcpkgs/libressl/patches
> https://github.com/void-linux/void-packages/blob/master/srcpkgs/libressl/template

I found, in the code, maybe this is a issue...
```
	i686-musl) configure_args+=" --disable-hardening";;
```

uberhacker:
@satmandu: I always test on actual hardware.  The i686 Chromebook is a Samsung Series 5.

uberhacker:
Fixed in #4392.

PR #4367, opened 2020-10-05
Curl update to 7.72

satmandu:
Tested with LibreSSL update to 3.1.4

Works properly:
- [x] x86_64

---


uberhacker:
Fixed in #4392.

PR #4368, opened 2020-10-05
cmake update to 3.18.3

satmandu:
Compiles and runs with updated LibreSSL 3.1.4

Works properly:
- [x] x86_64

---


uberhacker:
Fixed in #4392.

PR #4372, opened 2020-10-05
Update nftables from 0.9.0 to 0.9.6

leoplo:
Fixes #4372.

## Description
Update for nftables and libnftnl to solve the issue above.

## Addtional information
Appears to work properly on:
- [x] x86_64


uberhacker:
Thank you for the update @leoplo!

PR #4373, opened 2020-10-05
zstd update to 1.4.5 & add lz4 support

satmandu:
Update zstd to 1.4.5

Add lz4 dependency so that it builds with lz4 support.

Works properly:
- [x] x86_64

---


satmandu:
Also, this is useful since zstd is used by current gcc (10, etc) as part of ```-lto```.

uberhacker:

PR #4374, opened 2020-10-05
nftables and libnftnl missing compatibility field

vincowl:


uberhacker:
Wait, what?  Shouldn't these already have the compatibility lines?

vincowl:
Nope 😂... Too many updates these days to be focused enough on so small details : but `crew update` is a great safeguard !

uberhacker:
It's a good thing we have users like you @vincowl to help fix these things.  Nice catch!

vincowl:
Thanks but nothing requiring great skills. The `crew update` error message was clear enough to point out the involved packages using a `grep -L` . If only everything could be so easy !

leoplo:
Sorry my bad. Is the compatibility supposed to be set to all even if it has only been tested on 86_64 architecture ?

uberhacker:
Hey @leoplo: The compatibility line should have been there already.  Not sure how your PR ended up without it but I suspect you were not branched off the latest master on your working branch.

uberhacker:
Actually, it was there.  You just removed it along with all the other lines in your PR.  That is my bad.  I should have caught it in the review.  See https://github.com/skycocker/chromebrew/pull/4372/files.

PR #4375, opened 2020-10-06
libdrm update to 2.4.102, adjusted for meson build

satmandu:
libdrm 2.4.102 needed for new mesa build.
Also this now uses meson for building.

Works properly:
- [x] x86_64

---


uberhacker:

PR #4385, opened 2020-10-07
Update gtk3 to 3.24.23 & use clang

satmandu:
I'm using clang here to compile instead of gcc since that uses old and thus one may be able to use some of the libraries which ChromeOS ships with.

For some reason I was unable to get it to use lld with gcc easily in the install.
(The end goal is to get graphics acceleration working.)

Works properly:
- [x] x86_64

### DIFF
--- a/packages/gtk3.rb
+++ b/packages/gtk3.rb
@@ -22,7 +22,7 @@ class Gtk3 < Package
   depends_on 'shared_mime_info'
   depends_on 'six' => :build
   depends_on 'xdg_base'
-  depends_on 'llvm'
+  depends_on 'llvm' => :build
 
   def self.build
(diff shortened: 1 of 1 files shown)


packages/gtk3.rb
```diff
@@ -35,8 +22,11 @@ class Gtk3 < Package
   depends_on 'shared_mime_info'
   depends_on 'six' => :build
   depends_on 'xdg_base'
+  depends_on 'llvm'
```
uberhacker, line 25: `depends_on 'llvm' => :build`?  Or is this also a runtime dependency?

satmandu: Ah yes, that's what I meant.

satmandu (commented):
so like this?

PR #4386, opened 2020-10-07
Update glew to 2.2.0-rc3

satmandu:
(This is the first update update to this package in 3 years - from January of 2020.)

Also set this to use the llvm linker since that understands the ChromeOS libraries.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/glew.rb
+++ b/packages/glew.rb
@@ -12,16 +12,21 @@ class Glew < Package
   depends_on 'mesa'
   depends_on 'llvm'
 
-  def self.build
+  def self.patch
     system 'tar fxv ../download'
+    Dir.chdir 'glew-2.2.0/build' do
+      # As per https://github.com/nigels-com/glew/pull/268
+      system "sed -i 's/OpenGL::EGL/NOT OpenGL_EGL_FOUND/' cmake/CMakeLists.txt"
+      system "sed -i 's/endif()/endif ()/g' cmake/CMakeLists.txt"
+    end
+  end
+  
+  def self.build
     ENV['CFLAGS.EXTRA'] = "-fuse-ld=lld"
     ENV['CFLAGS'] = "-fuse-ld=lld"
     ENV['CXXFLAGS'] = "-fuse-ld=lld"
     ENV['GLEW_DEST'] = "#{CREW_PREFIX}"
     Dir.chdir 'glew-2.2.0/build' do
-      # As per https://github.com/nigels-com/glew/pull/268
-      system "sed -i 's/OpenGL::EGL/NOT OpenGL_EGL_FOUND/' cmake/CMakeLists.txt"
-      system "sed -i 's/endif()/endif ()/g' cmake/CMakeLists.txt"
       system 'cmake',
              '-DCMAKE_BUILD_TYPE=Release',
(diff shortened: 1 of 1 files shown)


packages/glew.rb
```diff
@@ -3,38 +3,36 @@
...
-    Dir.chdir 'build' do
+    system 'tar fxv ../download'
+    ENV['CFLAGS.EXTRA'] = "-fuse-ld=lld"
+    ENV['CFLAGS'] = "-fuse-ld=lld"
+    ENV['CXXFLAGS'] = "-fuse-ld=lld"
+    ENV['GLEW_DEST'] = "#{CREW_PREFIX}"
+    Dir.chdir 'glew-2.2.0/build' do
+      # As per https://github.com/nigels-com/glew/pull/268
+      system "sed -i 's/OpenGL::EGL/NOT OpenGL_EGL_FOUND/' cmake/CMakeLists.txt"
+      system "sed -i 's/endif()/endif ()/g' cmake/CMakeLists.txt"
```
uberhacker, line 24: It might be good to put these patches in a separate `def self.patch` section to separate concerns.

satmandu: Does the new change look better?

packages/glew.rb
```diff
@@ -3,38 +3,41 @@
...
-  })
-  binary_sha256 ({
-    aarch64: 'd80923b6a32ec1b38df81dcd7860516d140c57157e3dea23a37d8f874c1cbbe3',
-     armv7l: 'd80923b6a32ec1b38df81dcd7860516d140c57157e3dea23a37d8f874c1cbbe3',
-       i686: '9dbd934bdf44c5109f9e52931801c2c0fb0b7806f146d4ead90a129bb31ebac1',
-     x86_64: '0496b31639402805881fa3312aaf8bea3b4239734b38935bbc53a10d48bcc840',
-  })
 
   depends_on 'mesa'
+  depends_on 'llvm'
```
uberhacker, line 13: `depends_on 'llvm' => :build`?

packages/glew.rb
```diff
@@ -3,38 +3,41 @@
 class Glew < Package
   description 'GLEW provides efficient run-time mechanisms for determining which OpenGL extensions are supported on the target platform'
   homepage 'http://glew.sourceforge.net/'
-  version '2.1.0-1'
+  version '2.2.0-rc3'
   compatibility 'all'
-  source_url 'https://github.com/nigels-com/glew/releases/download/glew-2.1.0/glew-2.1.0.tgz'
-  source_sha256 '04de91e7e6763039bc11940095cd9c7f880baba82196a7765f727ac05a993c95'
+  source_url 'https://sourceforge.net/projects/glew/files/glew/snapshots/glew-20200115.tgz/download'
```
uberhacker, line 8: Why the switch away from github to sourceforge?

satmandu: Github doesn't yet have the ready for use package yet for 2.2.0-rc3, but it is linked to from the github page.
See https://github.com/nigels-com/glew#user-content-downloads:~:text=Recent%20snapshots,-Snapshots

packages/glew.rb
satmandu: Fixed!

PR #4389, opened 2020-10-07
Update keybase.rb

aman601:
Updated packed source and sum, needs binary
https://github.com/keybase/client/archive/v5.5.1.tar.gz


uberhacker:

PR #4390, opened 2020-10-07
Update gcalculator.rb

aman601:
Updated package source and hash, needs binary


uberhacker:

PR #4395, opened 2020-10-08
Update lynis.rb

masseyzachery57:
updated version and hash, needs binaries


uberhacker:
Thank you for the update @masseyzachery57!

PR #4396, opened 2020-10-08
Update erlang.rb

aman601:
updated url and hash, needs binaries


uberhacker:

PR #4397, opened 2020-10-08
Update mpv.rb

masseyzachery57:
updated version and hash, needs binaries


uberhacker:
Thank you for the update @masseyzachery57!  Did you get sound working on any of your videos?

PR #4400, opened 2020-10-08
Let stopsommelier also pkill Xwayland

satmandu:
stopsommelier can leave Xwayland hanging. Kill it too.
startsommelier will then happily start Xwayland.

(killing Xwayland using the built-in ```pkill```.)

Works properly:
- [x] x86_64

### DIFF
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -60,6 +60,7 @@ class Sommelier < Package
       system "echo 'SOMM=\$(pidof sommelier 2> /dev/null)' >> stopsommelier"
       system "echo 'if [ ! -z \"\$SOMM\" ]; then' >> stopsommelier"
       system "echo '  killall -g sommelier' >> stopsommelier"
+      system "echo '  pkill Xwayland' >> stopsommelier"
       system "echo '  sleep 3' >> stopsommelier"
       system "echo 'fi' >> stopsommelier"
       system "echo 'SOMM=\$(pidof sommelier 2> /dev/null)' >> stopsommelier"


packages/sommelier.rb
```diff
@@ -60,6 +60,7 @@ def self.build
       system "echo 'SOMM=\$(pidof sommelier 2> /dev/null)' >> stopsommelier"
       system "echo 'if [ ! -z \"\$SOMM\" ]; then' >> stopsommelier"
       system "echo '  killall -g sommelier' >> stopsommelier"
+      system "echo '  pkill Xwayland' >> stopsommelier"
```
uberhacker, line 63: Should we test if Xwayland is running first before attempting to kill?  Otherwise, it will throw an error.

uberhacker: I thought about adding a `restartsommelier` command also.  Basically would just run `stopsommelier && startsommelier`.

ThatGeekyWeeb: this should work
```sh
if ! pgrep "xwayland"; then killall Xwayland; fi
```

satmandu: Changed to if then format as per @ThatGeekyWeeb suggestion.

packages/sommelier.rb
```diff
@@ -60,7 +60,7 @@ def self.build
       system "echo 'SOMM=\$(pidof sommelier 2> /dev/null)' >> stopsommelier"
       system "echo 'if [ ! -z \"\$SOMM\" ]; then' >> stopsommelier"
       system "echo '  killall -g sommelier' >> stopsommelier"
-      system "echo '  pkill Xwayland' >> stopsommelier"
+      system "echo '  if pgrep "Xwayland"; then killall Xwayland; fi"
```
uberhacker, line 63: We'll need to bump the version and remove the old binaries.  Just add another hex character from the git hash on the end to bump the version.

packages/sommelier.rb
```diff
@@ -3,7 +3,7 @@
 class Sommelier < Package
   description 'Sommelier works by redirecting X11 and Wayland programs to the built-in ChromeOS wayland server.'
   homepage 'https://chromium.googlesource.com/chromiumos/containers/sommelier'
-  version '1382ce084cc40'
+  version '1778a8bb16c3'
```
uberhacker, line 6: Not sure where you got that hash but I was thinking about `version '1382ce084cc407'`.

packages/sommelier.rb
uberhacker: Please remove the old binaries and this should be ready.

PR #4415, opened 2020-10-09
[Sommelier] Correct syntax error

vincowl:


packages/sommelier.rb
```diff
@@ -3,7 +3,7 @@
 class Sommelier < Package
   description 'Sommelier works by redirecting X11 and Wayland programs to the built-in ChromeOS wayland server.'
   homepage 'https://chromium.googlesource.com/chromiumos/containers/sommelier'
-  version '1382ce084cc407'
+  version '1382ce084cc408'
```
uberhacker, line 6: No need to change the version.  I haven't generated binaries yet.

uberhacker:
Good catch @vincowl!

PR #4418, opened 2020-10-09
Create package for pup

cstrouse:
Pup is a CLI utility for manipulating HTML similar to the jq tool for JSON. The developers have not been actively tagging releases but the tool has been quite stable in my experience and very useful.

Tested on x86_64 only.

### DIFF
--- a/packages/pup.rb
+++ b/packages/pup.rb
@@ -1,0 +1,24 @@
+require 'package'
+
+class Pup < Package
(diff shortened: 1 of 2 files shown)


uberhacker:
Thank you for the update @cstrouse!   I appreciate the addition to packages.yaml also.  Eventually, I would like to use this metadata to check for updates automagically with urlwatch.

PR #4434, opened 2020-10-16
Update clamav from 0.100.0 to 0.103.0

leoplo:
## Description

- Updates clamav package version
- Removes the no longer necessary patch

## Additional information

Works properly:
- [x] x86_64


uberhacker:
Thank you for the update @leoplo!  Good to know we no longer need the patch.

PR #4436, opened 2020-10-16
[zathura] Update to 0.4.7

vincowl:


uberhacker:

PR #4437, opened 2020-10-16
[poppler] Update to 20.10.0

vincowl:


uberhacker:

PR #4439, opened 2020-10-16
[xxhash] Update to 0.8.0

vincowl:


packages/xxhash.rb
```diff
@@ -3,29 +3,22 @@
...
 
   def self.build
     system 'make', "PREFIX=#{CREW_PREFIX}"
   end
 
   def self.install
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install' 
     FileUtils.mkdir_p ["#{CREW_DEST_PREFIX}/bin", "#{CREW_DEST_MAN_PREFIX}/man1"]
     FileUtils.cp 'xxhsum', "#{CREW_DEST_PREFIX}/bin/"
     FileUtils.cp 'xxhsum.1', "#{CREW_DEST_MAN_PREFIX}/man1/"
```
uberhacker, line 24: Does the `system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'` line negate the `FileUtils...` commands that follow?

vincowl: Indeed FileUtils lines are useless now. I removed them

PR #4450, opened 2020-10-19
Sommelier & Xwayland

satmandu:
Fixes Lack of hardware acceleration and updates sommelier to use near-current version.

This includes a Top of Tree version of  an xwayland package as well, which has Google Xorg patches for use with sommelier.

## Addtional information

Works properly:
- [x] x86_64

Builds properly:
- [x] armv7l


satmandu:
Methods cribbed from here: https://www.reddit.com/user/mxmrn/comments/

satmandu:
At the miinimum this needs an xserver compiled using the lld linker, and maybe also mesa too.  Checking on that now.

uberhacker:
This will be interesting to test @satmandu.  The current version is completely broken for i686 architecture.

satmandu:
The diff between my packages and the default are these:

```
Libx11
Libepoxy
Xorg_proto
Pixman
Libxext
pip-20.2.4
Python3
Mesa
Libxrandr
Xkbcomp
Xorg_server
Sommelier
Librsvg
Vala
Pango
Harfbuzz
Xcb_proto
Libxdmcp
Libva
Gcc_tools
```

Will bisect tomorrow...

satmandu:
The order I needed to install my modified files is as follows:

xorg_proto
libxdmcp
xcb_proto
libx11
libxext
harfbuzz
libxrandr
libva
pixman
libxrandr
mesa
xkbcomp
libepoxy
xorg_server
sommelier

Most of my modifications are for compiling with clang or lld. Is that an issue on non-x86_64 architectures?

uberhacker:
[Here](https://github.com/skycocker/chromebrew/issues/4403) is an example of what happens on armv7l architecture with the `-fuse-ld=lld` flags set.

satmandu:
@uberhacker Does this happen with gcc10 on arm7l or just with the older gcc8?

uberhacker:
I only tested gcc8.

satmandu:
@uberhacker It compiles fine in my arm7l container on my RPI4B.

packages/sommelier.rb
```diff
@@ -2,84 +2,127 @@
...
+# See merged at in source url to goto find actual tgz url used, which for some reason doesn't seem to have a stable sha256sum.
 
   depends_on 'mesa'
   depends_on 'xkbcomp'
-  depends_on 'xorg_server'
+  depends_on 'xorg_server' # Miminal xwayland package also works.
   depends_on 'psmisc'
   depends_on 'xsetroot'
+  depends_on 'llvm' => ':build'
+  depends_on 'meson' => ':build'
```
uberhacker, line 21: Don't need the `meson` line since it is already included in "core".

packages/sommelier.rb
```diff
@@ -107,7 +151,8 @@ def self.postinstall
     puts "Please be aware that gui applications may not work without the sommelier daemon running.".orange
     puts
     puts "If you are upgrading from an earlier version of sommelier, run".orange
-    puts "'stopsommelier && startsommelier' or logout and login again.".orange
+    puts "'stopsommelier && restartsommelier' or logout and login again.".orange
```
uberhacker, line 154: Should only need `restartsommelier`, right?

packages/sommelier.rb
```diff
@@ -2,84 +2,127 @@
...
+      ENV['CXX'] = 'clang'
+
+      system "meson",
+        "-Dprefix=#{CREW_PREFIX}",
+        "-Ddatadir=#{CREW_LIB_PREFIX}",
+        "-Dsysconfdir=#{CREW_PREFIX}/etc",
+        "-Dxwayland_path=#{CREW_PREFIX}/bin/Xwayland",
+        "-Dxwayland_gl_driver_path=#{CREW_LIB_PREFIX}/dri",
+        "-Dsharedstatedir=#{CREW_LIB_PREFIX}",
+        "-Dlocalstatedir=#{CREW_LIB_PREFIX}",
```
uberhacker, line 43: Does `#{CREW_MESON_OPTIONS}` work?

packages/sommelier.rb
```diff
@@ -2,84 +2,127 @@
...
-      system "echo 'fi' >> stopsommelier"
-      system "echo 'SOMM=\$(pidof sommelier 2> /dev/null)' >> stopsommelier"
-      system "echo 'if [ -z \"\$SOMM\" ]; then' >> stopsommelier"
-      system "echo '  echo \"sommelier process stopped\"' >> stopsommelier"
-      system "echo 'else' >> stopsommelier"
-      system "echo '  echo \"sommelier process \$SOMM is running\"' >> stopsommelier"
-      system "echo '  exit 1' >> stopsommelier"
-      system "echo 'fi' >> stopsommelier"
+
+  system 'curl https://chromium.googlesource.com/chromiumos/platform2/+archive/be4e16feb380360cabbb5d6199a09592ecaf4a42.tar.gz | tar mzx --warning=no-timestamp'
```
uberhacker, line 25: Should we have a checksum on this download?

packages/sommelier.rb
```diff
@@ -2,84 +2,127 @@
...
+  Dir.chdir ("vm_tools/sommelier") do
+      # Using method here to download single file from googlesource in lieu of 
+      # checking out the ChromeOS kernel tree submodule:
+      system 'curl "https://chromium.googlesource.com/chromiumos/third_party/kernel/+/refs/heads/master/include/uapi/linux/virtwl.h?format=TEXT" | base64 --decode > virtwl.h'
```
uberhacker, line 29: Should we have a checksum on this download?

uberhacker: Also, we should use a git hash instead of master for consistency.

packages/sommelier.rb
satmandu: restartsommelier is an alias which requires reloading .bashrc  - maybe that and startsommelier should be separate shellscripts, since these aliases are fragile?

packages/sommelier.rb
satmandu: We can use ```https://chromium.googlesource.com/chromiumos/third_party/kernel/+/5d641a7b7b64664230d2fd2aa1e74dd792b8b7bf/include/uapi/linux/virtwl.h```

Let me hack up the syntax like in mesa...

packages/sommelier.rb
satmandu: good idea. ```https://chromium.googlesource.com/chromiumos/third_party/kernel/+/5d641a7b7b64664230d2fd2aa1e74dd792b8b7bf/include/uapi/linux/virtwl.h``` is the current one. Let me work on that when I get a chance.

packages/sommelier.rb
satmandu: I can't get a stable checksummable download of this link. Something with the site. :/

packages/sommelier.rb
uberhacker: I guess we can test and find out.

satmandu:
Just tried that a short time ago and I kept getting different checksums for
the folder that download. Chromium source derived archives are weird.
Someone should setup a github mirror. (I didn't see a github mirror for
project2).

satmandu:
Also note new const ```  LD_SO_ARCH = if ARCH == 'armvll' || ARCH == 'aarch64' then 'armhf' else ARCH end``` which might be used elsewhere?

uberhacker:
It would be news to me if there was a Chrome OS device available with `armvll` architecture.

satmandu:
I'll get that fixed some time tonight. 😅

packages/sommelier.rb
```diff
@@ -20,7 +20,7 @@ class Sommelier < Package
   depends_on 'llvm' => ':build'
```
uberhacker, line 20: Remove quotes around `:build`?

satmandu: done

satmandu:
Re: ```LD_SO_ARCH = if ARCH == 'armv7l' || ARCH == 'aarch64' then 'armhf' else ARCH```
Do all armv7l & aarch64 devices use ```ld-linux-armhf.so.2 ld.so.2```?

packages/sommelier.rb
```diff
@@ -2,84 +2,138 @@
...
-      system "echo 'fi' >> stopsommelier"
+
+  system 'curl https://chromium.googlesource.com/chromiumos/platform2/+archive/be4e16feb380360cabbb5d6199a09592ecaf4a42.tar.gz | tar mzx --warning=no-timestamp'
+  Dir.chdir ("vm_tools/sommelier") do
+      # Using method here to download single file from googlesource in lieu of 
+      # checking out the ChromeOS kernel tree submodule:
+      #system 'curl "https://chromium.googlesource.com/chromiumos/third_party/kernel/+/refs/heads/master/include/uapi/linux/virtwl.h?format=TEXT" | base64 --decode > virtwl.h'
+
+
+    ######################### Download virtwl.h from Chromium 5.4 kernel tree ###########################################
```
JL2210, line 34: hmm... I'd have thought that my new linux headers package would've worked for this.

packages/sommelier.rb
JL2210: yeah, googlesource archives are probably created on-the-fly, so their timestamps are always different

packages/sommelier.rb
satmandu: This is a google modification used in the Chromium kernel fork which hasn't been submitted upstream yet, as far as I can tell.

packages/sommelier.rb
satmandu: YES!

satmandu:
Also any chance of people trying this out and getting this merged soon? It's working quite well for me.

```
chronos@localhost /usr/local/tmp/crew $ glxinfo -B
name of display: :0
display: :0  screen: 0
direct rendering: Yes
Extended renderer info (GLX_MESA_query_renderer):
    Vendor: Intel Open Source Technology Center (0x8086)
    Device: Mesa DRI Intel(R) UHD Graphics 615 (AML-KBL) (0x591c)
    Version: 20.2.1
    Accelerated: yes
    Video memory: 3058MB
    Unified memory: yes
    Preferred profile: core (0x1)
    Max core profile version: 4.6
    Max compat profile version: 3.0
    Max GLES1 profile version: 1.1
    Max GLES[23] profile version: 3.2
OpenGL vendor string: Intel Open Source Technology Center
OpenGL renderer string: Mesa DRI Intel(R) UHD Graphics 615 (AML-KBL)
OpenGL core profile version string: 4.6 (Core Profile) Mesa 20.2.1
OpenGL core profile shading language version string: 4.60
OpenGL core profile context flags: (none)
OpenGL core profile profile mask: core profile

OpenGL version string: 3.0 Mesa 20.2.1
OpenGL shading language version string: 1.30
OpenGL context flags: (none)

OpenGL ES profile version string: OpenGL ES 3.2 Mesa 20.2.1
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.20

chronos@localhost /usr/local/tmp/crew $ 
```

satmandu:
Just some data points from sommelier running in crostini:
```
pgrep -fa Xwayland
323 /opt/google/cros-containers/bin/../lib/ld-linux-x86-64.so.2 --library-path /opt/google/cros-containers/bin/../lib --inhibit-rpath  /opt/google/cros-containers/bin/Xwayland.elf :1 -nolisten tcp -rootless -displayfd 23 -wm 24 -auth /home/satadru/.Xauthority -fp /usr/share/fonts/X11/misc, /usr/share/fonts/X11/cyrillic, /usr/share/fonts/X11/100dpi/:unscaled, /usr/share/fonts/X11/75dpi/:unscaled, /usr/share/fonts/X11/Type1, /usr/share/fonts/X11/100dpi, /usr/share/fonts/X11/75dpi, built-ins
324 /opt/google/cros-containers/bin/../lib/ld-linux-x86-64.so.2 --library-path /opt/google/cros-containers/bin/../lib --inhibit-rpath  /opt/google/cros-containers/bin/Xwayland.elf -nolisten tcp -rootless -displayfd 23 -wm 24 -auth /home/satadru/.Xauthority -fp /usr/share/fonts/X11/misc, /usr/share/fonts/X11/cyrillic, /usr/share/fonts/X11/100dpi/:unscaled, /usr/share/fonts/X11/75dpi/:unscaled, /usr/share/fonts/X11/Type1, /usr/share/fonts/X11/100dpi, /usr/share/fonts/X11/75dpi, built-ins
```

```
$ ls -aFl /opt/google/cros-containers/bin/../lib
total 65384
drwxr-xr-x 2 nobody nogroup     4096 Sep 28 01:10 ./
drwxr-xr-x 6 nobody nogroup     4096 Sep 28 00:10 ../
-rwxr-xr-x 1 nobody nogroup   169552 Sep 21 21:28 ld-linux-x86-64.so.2*
-rwxr-xr-x 1 nobody nogroup  1653480 Sep 27 23:51 libbase-core-679961.so*
-rwxr-xr-x 1 nobody nogroup   285760 Sep 27 23:51 libbase-dbus-679961.so*
-rwxr-xr-x 1 nobody nogroup    24064 Sep 27 23:51 libbase-dl-679961.so*
-rwxr-xr-x 1 nobody nogroup   397064 Sep 28 00:06 libbrillo-core.so*
-rwxr-xr-x 1 nobody nogroup    71544 Sep 27 23:35 libbsd.so.0*
-rwxr-xr-x 1 nobody nogroup    60376 Sep 27 23:31 libbz2.so.1*
-rwxr-xr-x 1 nobody nogroup   160712 Sep 27 23:27 libc++abi.so.1*
-rwxr-xr-x 1 nobody nogroup    57472 Sep 27 23:46 libcares.so.2*
-rwxr-xr-x 1 nobody nogroup  2208664 Sep 27 23:45 libcrypto.so.1.1*
-rwxr-xr-x 1 nobody nogroup   751328 Sep 27 23:34 libc++.so.1*
-rwxr-xr-x 1 nobody nogroup  1795984 Sep 21 21:28 libc.so.6*
-rwxr-xr-x 1 nobody nogroup   278040 Sep 27 23:43 libdbus-1.so.3*
-rwxr-xr-x 1 nobody nogroup    14448 Sep 21 21:28 libdl.so.2*
-rwxr-xr-x 1 nobody nogroup    57000 Sep 27 23:43 libdrm.so.2*
-rwxr-xr-x 1 nobody nogroup   203256 Sep 28 00:04 libEGL.so.1*
-rwxr-xr-x 1 nobody nogroup   813344 Sep 28 00:09 libepoxy.so.0*
-rwxr-xr-x 1 nobody nogroup   130248 Sep 27 23:43 libevent_core-2.1.so.6*
-rwxr-xr-x 1 nobody nogroup   131968 Sep 27 23:34 libexpat.so.1*
-rwxr-xr-x 1 nobody nogroup    25992 Sep 27 23:32 libffi.so.6*
-rwxr-xr-x 1 nobody nogroup    54504 Sep 28 00:04 libgbm.so.1*
-rwxr-xr-x 1 nobody nogroup   173400 Sep 28 00:04 libglapi.so.0*
-rwxr-xr-x 1 nobody nogroup    33512 Sep 28 00:04 libGLESv2.so.2*
-rwxr-xr-x 1 nobody nogroup   971352 Sep 27 23:50 libglib-2.0.so.0*
-rwxr-xr-x 1 nobody nogroup   388520 Sep 28 00:04 libGL.so.1*
-rwxr-xr-x 1 nobody nogroup   308528 Sep 27 23:51 libgrpc++.so.1*
-rwxr-xr-x 1 nobody nogroup  1069736 Sep 27 23:51 libgrpc.so.6*
-rwxr-xr-x 1 nobody nogroup  1435680 Sep 21 21:28 libm.so.6*
-rwxr-xr-x 1 nobody nogroup    35224 Sep 28 00:09 libnss_compat.so.2*
-rwxr-xr-x 1 nobody nogroup    22496 Sep 21 21:28 libnss_dns.so.2*
-rwxr-xr-x 1 nobody nogroup    47072 Sep 21 21:28 libnss_files.so.2*
-rwxr-xr-x 1 nobody nogroup   440632 Sep 27 23:35 libpcre.so.1*
-rwxr-xr-x 1 nobody nogroup   524488 Sep 27 23:41 libpixman-1.so.0*
-rwxr-xr-x 1 nobody nogroup   552240 Sep 27 23:43 libprotobuf-lite.so.22*
-rwxr-xr-x 1 nobody nogroup  2384968 Sep 27 23:43 libprotobuf.so.22*
-rwxr-xr-x 1 nobody nogroup   105496 Sep 21 21:28 libpthread.so.0*
-rwxr-xr-x 1 nobody nogroup    84008 Sep 21 21:28 libresolv.so.2*
-rwxr-xr-x 1 nobody nogroup    31056 Sep 21 21:28 librt.so.1*
-rwxr-xr-x 1 nobody nogroup   471288 Sep 27 23:45 libssl.so.1.1*
-rwxr-xr-x 1 nobody nogroup    38568 Sep 27 23:43 libwayland-client.so.0*
-rwxr-xr-x 1 nobody nogroup     4536 Sep 27 23:43 libwayland-egl.so.1*
-rwxr-xr-x 1 nobody nogroup    51752 Sep 27 23:43 libwayland-server.so.0*
-rwxr-xr-x 1 nobody nogroup  1090848 Sep 27 23:52 libX11.so.6*
-rwxr-xr-x 1 nobody nogroup     3784 Sep 27 23:52 libX11-xcb.so.1*
-rwxr-xr-x 1 nobody nogroup     9888 Sep 27 23:41 libXau.so.6*
-rwxr-xr-x 1 nobody nogroup     8592 Sep 27 23:51 libxcb-composite.so.0*
-rwxr-xr-x 1 nobody nogroup    12880 Sep 27 23:51 libxcb-dri2.so.0*
-rwxr-xr-x 1 nobody nogroup    11560 Sep 27 23:51 libxcb-dri3.so.0*
-rwxr-xr-x 1 nobody nogroup    73776 Sep 27 23:51 libxcb-glx.so.0*
-rwxr-xr-x 1 nobody nogroup     8048 Sep 27 23:51 libxcb-present.so.0*
-rwxr-xr-x 1 nobody nogroup   116080 Sep 27 23:51 libxcb.so.1*
-rwxr-xr-x 1 nobody nogroup    19080 Sep 27 23:51 libxcb-sync.so.1*
-rwxr-xr-x 1 nobody nogroup    22928 Sep 27 23:51 libxcb-xfixes.so.0*
-rwxr-xr-x 1 nobody nogroup     7944 Sep 27 23:53 libXdamage.so.1*
-rwxr-xr-x 1 nobody nogroup    17320 Sep 27 23:41 libXdmcp.so.6*
-rwxr-xr-x 1 nobody nogroup    53944 Sep 27 23:52 libXext.so.6*
-rwxr-xr-x 1 nobody nogroup    16624 Sep 27 23:52 libXfixes.so.3*
-rwxr-xr-x 1 nobody nogroup   137944 Sep 27 23:43 libXfont2.so.2*
-rwxr-xr-x 1 nobody nogroup   219680 Sep 27 23:41 libxkbcommon.so.0*
-rwxr-xr-x 1 nobody nogroup   120600 Sep 27 23:52 libxkbfile.so.1*
-rwxr-xr-x 1 nobody nogroup     5752 Sep 27 23:41 libxshmfence.so.1*
-rwxr-xr-x 1 nobody nogroup    15784 Sep 27 23:53 libXxf86vm.so.1*
-rwxr-xr-x 1 nobody nogroup    74856 Sep 27 23:31 libz.so.1*
-rwxr-xr-x 2 nobody nogroup 23116224 Sep 28 00:05 swrast_dri.so*
-rwxr-xr-x 2 nobody nogroup 23116224 Sep 28 00:05 virtio_gpu_dri.so*
satadru@penguin:~$ 
```

```
satadru@penguin:~$ pgrep -fa sommelier
244 /opt/google/cros-containers/bin/../lib/ld-linux-x86-64.so.2 --library-path /opt/google/cros-containers/bin/../lib --inhibit-rpath  /opt/google/cros-containers/bin/sommelier.elf -X --x-display=1 --sd-notify=READY=1 --no-exit-with-child --x-auth=/home/satadru/.Xauthority /bin/sh -c systemctl --user set-environment DISPLAY_LOW_DENSITY=${DISPLAY};                     systemctl --user set-environment XCURSOR_SIZE_LOW_DENSITY=${XCURSOR_SIZE};                     systemctl --user import-environment SOMMELIER_VERSION;                     touch /home/satadru/.Xauthority;                     xauth -f /home/satadru/.Xauthority add :1 . $(xxd -l 16 -p /dev/urandom);                     . /etc/sommelierrc
249 /opt/google/cros-containers/bin/../lib/ld-linux-x86-64.so.2 --library-path /opt/google/cros-containers/bin/../lib --inhibit-rpath  /opt/google/cros-containers/bin/sommelier.elf --master --sd-notify=READY=1 --socket=wayland-0 /bin/sh -c systemctl --user set-environment WAYLAND_DISPLAY=${WAYLAND_DISPLAY};                     systemctl --user import-environment SOMMELIER_VERSION
253 /opt/google/cros-containers/bin/../lib/ld-linux-x86-64.so.2 --library-path /opt/google/cros-containers/bin/../lib --inhibit-rpath  /opt/google/cros-containers/bin/sommelier.elf --master --sd-notify=READY=1 --socket=wayland-1 /bin/sh -c systemctl --user set-environment WAYLAND_DISPLAY_LOW_DENSITY=${WAYLAND_DISPLAY};                     systemctl --user import-environment SOMMELIER_VERSION
254 /opt/google/cros-containers/bin/../lib/ld-linux-x86-64.so.2 --library-path /opt/google/cros-containers/bin/../lib --inhibit-rpath  /opt/google/cros-containers/bin/sommelier.elf -X --x-display=0 --sd-notify=READY=1 --no-exit-with-child --x-auth=/home/satadru/.Xauthority /bin/sh -c systemctl --user set-environment DISPLAY=${DISPLAY};                     systemctl --user set-environment XCURSOR_SIZE=${XCURSOR_SIZE};                     systemctl --user import-environment SOMMELIER_VERSION;                     touch /home/satadru/.Xauthority;                     xauth -f /home/satadru/.Xauthority add :0 . $(xxd -l 16 -p /dev/urandom);                     . /etc/sommelierrc
satadru@penguin:~$ 
```

satmandu:
FYI simply installing this and then a ```crew reinstall mesa``` is letting me have hardware acceleration at the moment on x86_64.

packages/sommelier.rb
satmandu: Those have all been moved to separate scripts, with initsommelier renamed to startsommelier.

packages/sommelier.rb
satmandu: Only a specific virtwl.h git hash  is now being pulled.

satmandu:
The current goal is to move sommelier + all deps into a separate folder so that it is standalone.

packages/sommelier.rb
```diff
@@ -2,90 +2,197 @@
...
+    # lld is needed so libraries linked to system libraries (e.g. libgbm.so) can be linked against, since those are required for graphics acceleration.
+    ENV['CFLAGS'] = "-fuse-ld=lld"
+    ENV['CXXFLAGS'] = "-fuse-ld=lld"
+    
+    system "meson #{CREW_MESON_OPTIONS} -Dxwayland_path=#{CREW_PREFIX}/bin/Xwayland -Dxwayland_gl_driver_path=#{CREW_LIB_PREFIX}/dri -Ddefault_library=both -Dxwayland_shm_driver=noop -Dshm_driver=noop -Dvirtwl_device=/dev/null -Dpeer_cmd_prefix=\"#{CREW_PREFIX}#{PEER_CMD_PREFIX}\" build"
+    system "meson configure build"
+    system "ninja -C build"
+    
+    Dir.chdir ("build") do
+      system 'curl "https://chromium.googlesource.com/chromiumos/containers/sommelier/+/refs/heads/master/sommelierrc?format=TEXT" | base64 --decode > sommelierrc'
```
uberhacker, line 62: Change `master` to a git hash instead?

packages/sommelier.rb
```diff
@@ -2,90 +2,197 @@
...
+        system "install -Dm755 stopsommelier #{CREW_DEST_PREFIX}/bin/stopsommelier"
+        system "install -Dm755 restartsommelier #{CREW_DEST_PREFIX}/bin/restartsommelier"
+        system "install -Dm755 sommelierrc #{CREW_DEST_PREFIX}/etc/sommelierrc"
+        system "install -Dm644 .sommelier/CLUTTER_BACKEND #{CREW_DEST_HOME}/.sommelier/CLUTTER_BACKEND"
+        system "install -Dm644 .sommelier/DISPLAY #{CREW_DEST_HOME}/.sommelier/DISPLAY"
+        system "install -Dm644 .sommelier/GDK_BACKEND #{CREW_DEST_HOME}/.sommelier/GDK_BACKEND"
+        system "install -Dm644 .sommelier/MESA_LOADER_DRIVER_OVERRIDE #{CREW_DEST_HOME}/.sommelier/MESA_LOADER_DRIVER_OVERRIDE"
+        system "install -Dm644 .sommelier/SCALE #{CREW_DEST_HOME}/.sommelier/SCALE"
+        system "install -Dm644 .sommelier/WAYLAND_DISPLAY #{CREW_DEST_HOME}/.sommelier/WAYLAND_DISPLAY"
+        system "install -Dm644 .sommelier/XDG_RUNTIME_DIR #{CREW_DEST_HOME}/.sommelier/XDG_RUNTIME_DIR"
```
uberhacker, line 190: Can't we include all these environment variables in one file?  Just thinking about simplicity.

satmandu: Rationale from error message I see otherwise:

https://chromium.googlesource.com/chromiumos/docs/+/master/security/noexec_shell_scripts.md

packages/sommelier.rb
```diff
@@ -2,90 +2,197 @@
...
-  depends_on 'xorg_server'
+  depends_on 'xorg_server' 
   depends_on 'psmisc'
+  depends_on 'vim' # Until there is a standalone package providing xxd
+  depends_on 'xdpyinfo'
+  depends_on 'xauth'
   depends_on 'xsetroot'
+  depends_on 'llvm' => :build
+
+  PEER_CMD_PREFIX = if ARCH == 'armv7l' || ARCH == 'aarch64' then '/lib/ld-linux-armhf.so.3' elsif ARCH == 'i686' then '/lib/ld-linux-i686.so.2' else '/lib64/ld-linux-x86-64.so.2' end
```
ThatGeekyWeeb, line 23: This would be better done using `case` as follows:
```ruby
case ARCH
 when 'armv7l', 'aarch64'
    PEER_CMD_PREFIX='/lib/ld-linux-armhf.so.3'
 when 'i686'
    PEER_CMD_PREFIX='/lib/ld-linux-i686.so.2'
 when 'x86_64'
    PEER_CMD_PREFIX='/lib64/ld-linux-x86-64.so.2'
end
```

packages/sommelier.rb
```diff
@@ -2,90 +2,197 @@
...
+      system "echo 'wayland' >> .sommelier/CLUTTER_BACKEND"
+      system "echo ':0' >> .sommelier/DISPLAY"
+      system "echo 'x11' > .sommelier/GDK_BACKEND"
+      system "echo 'wayland-1' >> .sommelier/WAYLAND_DISPLAY"
+      system "echo '/var/run/chrome' >> .sommelier/XDG_RUNTIME_DIR"
+      # Changing this to 0.5 since otherwise everything looks small on a HiDPI screen.
+      system "echo '0.5' >> .sommelier/SCALE"
+      
+      system "touch .sommelier/MESA_LOADER_DRIVER_OVERRIDE"
+      case ARCH
```
ThatGeekyWeeb, line 77: casing `ARCH` isn't ideal here, I'm sure theres a way to case the output of `uname -r` (which reports the kernel version)

ThatGeekyWeeb: also, this is only done for `i686` and `x86_64`, when there are devices under 4.16 on arm

packages/sommelier.rb
ThatGeekyWeeb: Would be smart to simply source a file with the variables, as that wouldn't cause any issues. Just source the file from `startsommelier` and export the vars.

ThatGeekyWeeb (commented):
The below changes should really be looked into.

packages/sommelier.rb
ThatGeekyWeeb: also there are non-intel devices...

satmandu: The rationale for this is that mesa doesn't enable the i965 successor driver IRIS on kernels earlier than 4.16. So this is specific to intel machines. Do non-intel machines even use iris?

The intel discrete GPUS are only coming out in the next several months anyways, and the chances of those being used with arm machines are next to nil, no?

satmandu:
Currently I am going through and seeing what the differences are between the crostini sommelier install and the chromebrew install.

I can drop the following files into the crostini lib folder to get graphical acceleration to be recognized by ```glxinfo -B```... though I'm not sure that glxinfo is the right proxy for this.

```
/usr/local/lib64/libGLEW.so.2.2
/usr/local/lib64/libGLU.so.1
/usr/local/lib64/libICE.so.6 
/usr/local/lib64/libstdc++.so.6
```
also this from the ChromeOS root partition:
```
/usr/lib64/dri/i965_dri.so
```
I also had to do a binary edit of ```Xwayland.elf``` using vim with ```:set binary``` set to replace ```/opt/google/cros-containers/bin``` with something pointing to my cros-containers folder on /usr/local.
I changed it to ```/usr/local/xcros-containers/bin```  and made a symlink.

**_In short, this is what I did:_**

First run this (after copying in everything from crostini's /opt/google/cros-containers into /usr/local/sommelier)
```
#!/bin/bash -x
# sprep
# sommelier prep
# sommelier from crostini should be extracted to the following spath:
spath=/usr/local/sommelier
cp /usr/lib64/dri/* $spath/lib/
cp $spath/bin/Xwayland.elf $spath/bin/Xwayland.elf.orig
sed -i "s/opt\/google\/cros-containers/usr\/local\/xcros-containers/" $spath/bin/Xwayland.elf
filelist="libGLEW.so.2.2 libSM.so.6 libGLU.so.1 libICE.so.6 libstdc++.so.6"
for i in ${filelist}; do cp $(realpath /usr/local/lib64/$i) $spath/lib/$i ; done
rm /usr/local/xcros-containers && ln -s $spath /usr/local/xcros-containers
```

Then run this:
```
#!/bin/bash -x
# stest
# Assumes you have glxinfo in your path (in mesa_utlls) and xxd (in vim) and that you have sommelierrc in the right place.
DISPLAY=:0
pkill -f sommelier.elf &>/dev/null
export spath=/usr/local/sommelier
export LIBGL_DRIVERS_PATH=$spath/lib
# may not be needed:
# export MESA_LOADER_DRIVER_OVERRIDE=i965
$spath/bin/sommelier --master --peer-cmd-prefix="$spath/lib/ld-linux-x86-64.so.2" --drm-device=/dev/dri/renderD128 --shm-driver=noop --data-driver=noop --display=wayland-0 --socket=wayland-1 --virtwl-device=/dev/null > /tmp/sommelier.log 2>&1 &
$spath/bin/sommelier -X --x-display=$DISPLAY  --scale=$SCALE --glamor --drm-device=/dev/dri/renderD128 --virtwl-device=/dev/null --shm-driver=noop --data-driver=noop --display=wayland-0 --xwayland-path=$spath/bin/Xwayland --xwayland-gl-driver-path=$spath/lib --peer-cmd-prefix="$spath/lib/ld-linux-x86-64.so.2 --library-path $spath/lib" --no-exit-with-child /bin/sh -c "touch ~/.Xauthority; xauth -f ~/.Xauthority add $DISPLAY . $(xxd -l 16 -p /dev/urandom); /usr/local/etc/sommelierrc"  &>>/tmp/sommelier.log &
sleep 1
LD_LIBRARY_PATH=$spath/lib glxinfo -B
```

This is what I get:
```
name of display: :0
display: :0  screen: 0
direct rendering: Yes
Extended renderer info (GLX_MESA_query_renderer):
    Vendor: Intel Open Source Technology Center (0x8086)
    Device: Mesa DRI Intel(R) UHD Graphics 615 (AML-KBL) (0x591c)
    Version: 20.1.4
    Accelerated: yes
    Video memory: 3058MB
    Unified memory: yes
    Preferred profile: core (0x1)
    Max core profile version: 4.6
    Max compat profile version: 3.0
    Max GLES1 profile version: 1.1
    Max GLES[23] profile version: 3.2
OpenGL vendor string: Intel Open Source Technology Center
OpenGL renderer string: Mesa DRI Intel(R) UHD Graphics 615 (AML-KBL)
OpenGL core profile version string: 4.6 (Core Profile) Mesa 20.1.4 (git-d6154f155b)
OpenGL core profile shading language version string: 4.60
OpenGL core profile context flags: (none)
OpenGL core profile profile mask: core profile

OpenGL version string: 3.0 Mesa 20.1.4 (git-d6154f155b)
OpenGL shading language version string: 1.30
OpenGL context flags: (none)

OpenGL ES profile version string: OpenGL ES 3.2 Mesa 20.1.4 (git-d6154f155b)
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.20
```

packages/sommelier.rb
ThatGeekyWeeb: Non-intel machines don't even use i965, and again, your just defaulting to disable iris when It's faster, you should be checking the kernel release with `uname -r`

satmandu: Suggestion for a test for kernel version > some version? Is there something builtin?

ThatGeekyWeeb: @uberhacker, I don't recall there being a builtin, how should we go abt this?

uberhacker: @ThatGeekyWeeb: I believe the way we sourced the variables in an earlier release was with a command alias to startsommelier.

uberhacker: The "ruby" way to get the kernel version is `kernel_version = ``uname -r```.

ThatGeekyWeeb: Perfect, @satmandu  simply do a `if` check with the above to see if the kernel is `4.16` or higher.

satmandu:
I've been doing more testing with just the minimal test case of using the setup in the crostini ```/opt/google/cros-containers``` and I still can't get windows to stay preserved across suspend/resume.

Any help on that would be appreciated. I'm glad to recognize that this isn't an issue with our compile of sommelier or perhaps even Xwayland.

I've been trying to suspend sommelier and Xwayland, but that doesn't seem to help. You might be able to use ```xdotool``` to get all PIDs of X apps running and suspend those first with a STOP signal, and then send a CONT signal to them upon wake, unwinding in reverse, but this is way above my skillset.

I think that crostini handles this entirely by just freezing the crosvm, and then unsuspending, which is an end-run around this entire issue. It would be nice to figure out how apps like mutter handle this, but the fixes to mutter to handle sleep seem to require a plethora of patches, e.g.: https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/506

(That set of 7 patches indicates that they had to actually implement functionality inside mutter to work around suspend/resume.)

Test script:
**(Note that running ```powerd_dbus_suspend```  requires having a functional symlink from ```/var/run/dbus``` to ```/usr/local/var/run/dbus```)**

First start sommelier (using the ```stest``` script above) and then start some xterms.

```
#!/bin/bash -x
# This tests sending stop signals to sommelier and XWayland
# and then suspending, and then sending continue signals
# in reverse order to the same after waking.
procs=$(pgrep -f "sommelier.*X" | sort -r)
rprocs=$(pgrep -f "sommelier.*X")
for i in $procs
do
kill -STOP -$i || kill -STOP $i
PGID=$(ps opgid= "$i")
echo "PID: $i, PGDID: $PGID, $(ps -p $i -o comm=)"
done
sudo powerd_dbus_suspend --delay=2 &
sleep 4
for i in $rprocs
do
echo "$(ps -p $i -o comm=)"
kill -CONT $i
kill -CONT -$i
done
```

packages/sommelier.rb
satmandu: I can do that, but can we confirm that MESA_LOADER_DRIVER_OVERRIDE is needed for arm kernels less than 4.16? Otherwise it is irrelevant.

(I can also revert the logic to load environment variables from multiple files.)

satmandu:
Ok this might be relevant: https://github.com/freedesktop/xorg-xserver/blob/0b86c0c36241989d7e9662d007c5297fe22ae8b5/hw/xfree86/os-support/linux/lnx_platform.c#L29

Systemd allows one to detect a paused drm node. Since we're not using systems, maybe that's an issue?

packages/sommelier.rb
ThatGeekyWeeb: I have no way to verify this, but I do know this may cause issues with non-Intel devices, if mesa tires to load the i965 driver on non-Intel device, it's likely going to just fail.

packages/sommelier.rb
satmandu: Now sourcing from ~/.sommelier.env again.

packages/sommelier.rb
satmandu: Done.

packages/sommelier.rb
satmandu: I've modified the logic to check for kernel at env variable load time if on an x86 or x86_64 machine from ```~/.sommelier.env```, since a kernel update might make this irrelevant on such machines.

ThatGeekyWeeb: Perfect!

packages/sommelier.rb
```diff
@@ -2,90 +2,200 @@
...
+export SCALE=0.5
+export SOMMELIER_ACCELERATORS=Super_L,<Alt>bracketleft,<Alt>bracketright
+export WAYLAND_DISPLAY=wayland-1
+export XDG_RUNTIME_DIR=/var/run/chrome
+UNAME_ARCH=$(uname -m)
+if [[ \"$UNAME_ARCH\" == 'x86_64' ]] || [[ \"$UNAME_ARCH\" == 'i686' ]]
+then
+  declare -a TEMP2
+  TEMP1=$(uname -r)
+  TEMP2=(${TEMP1//[.-]/ })
```
uberhacker, line 85: The 2 lines above could be combined into 1 with:
```
TEMP2=$(uname -r | cut -d'-' -f1)
```

packages/sommelier.rb
```diff
@@ -2,90 +2,200 @@
...
+  VERSION=$(((${TEMP2[0]} * 1000000)\
+        + (${TEMP2[1]} * 10000)\
+        + ${TEMP2[2]}))
```
uberhacker, line 88: Can we be certain every kernel version has 3 parts?

satmandu: Not at all, but chromeos kernels seem to follow that, and I think most linux non-RC kernels are a.b.c-d.e.f formatted.

packages/sommelier.rb
```diff
@@ -2,90 +2,200 @@
...
-      system "echo '  echo \"sommelier process stopped\"' >> stopsommelier"
-      system "echo 'else' >> stopsommelier"
-      system "echo '  echo \"sommelier process \$SOMM is running\"' >> stopsommelier"
-      system "echo '  exit 1' >> stopsommelier"
-      system "echo 'fi' >> stopsommelier"
-      system "echo '#!/bin/bash' > startsommelier"
-      system "echo 'set -a && source ~/.sommelier.env && set +a && initsommelier' >> startsommelier"
+    # There is no good way to checksum the googlesource tgz file, as they appear to be generated on the fly
+    # and checksums vary with each download.
+    system 'curl https://chromium.googlesource.com/chromiumos/platform2/+archive/c6f85a3472584df37fcf2d3d99474081f29256cd.tar.gz | tar mzx --warning=no-timestamp'
```
ThatGeekyWeeb, line 35: `curl -L` might be smarter here, as redirects may happen, and won't followed without `-L`

ThatGeekyWeeb (approved):
Seems good other then the comment I left for `curl -L` instead of `curl`!

packages/sommelier.rb
satmandu: Good call!

ThatGeekyWeeb (commented):
Other than uberhacker's comments this looks all good!

satmandu:
Related: If any of you can test the crouton version of this update, it would be appreciated, as apparently on some machines (e.g. octopus) sommelier windows survive suspend/resume, but not on my machine (nocturne). 

This may be due to a kernel issue, as nocturne has 4.4, and octopus has 4.14.

https://github.com/dnschneid/crouton/issues/4366#issuecomment-726180580

packages/sommelier.rb
satmandu: Fixed

satmandu:
Also @uberhacker I forget if I've asked before, but is there a way to do ```depends_on a or b```?

I have a leaner Xwayland package I'd like to be able to use as an option in lieu of xorg_server for sommelier.

uberhacker:
@satmandu: You can have conditional depends_on lines such as what is in the composer package:
```
depends_on 'php74' unless File.exists? "#{CREW_PREFIX}/bin/php"
```

satmandu:
Perfect thanks.

satmandu:
One more change, I'd like to put logs from sommelier into ```$CREW_PREFIX/var/log/sommelier.log```.

Any objections to that?

ThatGeekyWeeb:
Seems like a great idea! On objection here.

satmandu:
The sleep issues might be with older kernels and the aura compositor. I'm seeing the same weirdness with weston connecting to the built-in wayland display.

satmandu:
@uberhacker @ThatGeekyWeeb I added xwayland to this to get some xwayland changes upstream and from google in.

This should work with an older xorg_server install, so we can take out that dependency, but it might work better with it.

The sleep issues I am seeing may be unique to the 4.4 kernel, so there's no reason to hold this up for a resolution on that issue.

packages/sommelier.rb
satmandu: This actually doesn't put the major and minor version into the array...

justchen1369:
I'm interesting in porting this over to crouton.

ThatGeekyWeeb (approved):
Still looks great

uberhacker:
@satmandu: Is this still a WIP?  I noticed a few commits came trickling in the last few days.

satmandu:
@uberhacker Still doing some refactoring of the scripts, at this point to not kill the wayland sommelier proxy process if it isn't dead but the X sommelier process is.

satmandu:
Refactored startup scripts (startsommelier & sommelierd) to not restart both sommelier daemons (for wayland proxy & Xwayland) if one is already running.

Also putting sommelier instance PIDs in crewprefix/var/run/{sommelier-wayland.pid,sommelier-xwayland.pid} and using those PIDs for during the sommelier restart or stop process.

satmandu:
Any further changes I'll do as a separate PR. (Next is to use the more recent xdg-shell which was rolled out and then rolled back several months ago because it caused an issue with an older version of debian in crostini, which we shouldn't care about.)

packages/sommelier.rb
```diff
@@ -2,106 +2,256 @@
 
 class Sommelier < Package
   description 'Sommelier works by redirecting X11 and Wayland programs to the built-in ChromeOS wayland server.'
-  homepage 'https://chromium.googlesource.com/chromiumos/containers/sommelier'
-  version '1382ce084cc407'
+  homepage 'https://chromium.googlesource.com/chromiumos/platform2/+/HEAD/vm_tools/sommelier/'
+  version '20201025'
```
uberhacker, line 6: Do we have a newer version available since this has been sitting in the queue for so long?

packages/sommelier.rb
```diff
@@ -2,106 +2,256 @@
...
+    filename_virtwl = 'virtwl.h_base64'
+    sha256sum_virtwl = 'a8215f4946ccf30cbd61fcf2ecc4edfe6d05bffeee0bacadd910455274955446'
+    
+    if File.exist?(filename_virtwl) && Digest::SHA256.hexdigest( File.read("./#{filename_virtwl}") ) == sha256sum_virtwl
+      puts "Unpacking virtwl source code".yellow
+    else
+      puts "Downloading virtwl".yellow
+      system('curl', '-s', '-C', '-', '--insecure', '-L', '-#', url_virtwl, '-o', filename_virtwl)
+      abort 'Checksum mismatch. :/ Try again.'.lightred unless
+        Digest::SHA256.hexdigest( File.read("./#{filename_virtwl}") ) == sha256sum_virtwl
```
uberhacker, line 53: This should work also:
```
Digest::SHA256.hexdigest( File.read( filename_virtwl ) ) == sha256sum_virtwl
```

packages/sommelier.rb
```diff
@@ -2,106 +2,256 @@
...
+      puts "virtwl archive downloaded".lightgreen
+      system 'mkdir -p build/linux'
```
uberhacker, line 55: Rubify?
```
FileUtils.mkdir_p 'build/linux'
```

packages/sommelier.rb
```diff
@@ -2,106 +2,256 @@
...
+      system 'base64 --decode virtwl.h_base64 > build/linux/virtwl.h'
+    end    
+    
+    # lld is needed so libraries linked to system libraries (e.g. libgbm.so) can be linked against, since those are required for graphics acceleration.
+    ENV['CFLAGS'] = "-fuse-ld=lld"
+    ENV['CXXFLAGS'] = "-fuse-ld=lld"
+    
+    system "meson #{CREW_MESON_OPTIONS} -Dxwayland_path=#{CREW_PREFIX}/bin/Xwayland -Dxwayland_gl_driver_path=/usr/#{ARCH_LIB}/dri -Ddefault_library=both -Dxwayland_shm_driver=noop -Dshm_driver=noop -Dvirtwl_device=/dev/null -Dpeer_cmd_prefix=\"#{CREW_PREFIX}#{PEER_CMD_PREFIX}\" build"
```
uberhacker, line 63: Can we split this onto separate lines for readability?

packages/sommelier.rb
```diff
@@ -2,106 +2,256 @@
...
+      end
     end
   end
-
+  
   def self.postinstall
     puts
     puts "To complete the installation, execute the following:".lightblue
-    puts "echo '# Sommelier environment variables + daemon' >> ~/.bashrc".lightblue
+    puts "echo '# Sommelier envronment variables + daemon' >> ~/.bashrc".lightblue
```
uberhacker, line 237: It was spelled correctly the first time. :)

packages/sommelier.rb
```diff
@@ -2,106 +2,256 @@
...
-    puts "You may need to adjust the SCALE environment variable to get the correct screen size.".lightblue
+    puts "To start the sommelier daemon, run 'startsommelier'".lightblue
+    puts "To stop the sommelier daemon, run 'stopsommelier'".lightblue
+    puts "To restart the sommelier daemon, run 'restartsommelier'".lightblue
     puts
     puts "Please be aware that gui applications may not work without the sommelier daemon running.".orange
     puts
-    puts "If you are upgrading from an earlier version of sommelier, run".orange
-    puts "'stopsommelier && startsommelier' or logout and login again.".orange
+    puts "The sommelier daemon may also have to be restarted with 'restartsommelier' after waking your device.".red
```
uberhacker, line 252: Probably should use orange instead of red here.  Red is usually reserved to indicate an error.

packages/xwayland.rb
```diff
@@ -0,0 +1,174 @@
...
+  depends_on 'mesa'
+
+  # Patches are from Google xwayland overlay at https://source.chromium.org/chromiumos/chromiumos/codesearch/+/master:src/third_party/chromiumos-overlay/x11-base/xwayland/files/
+  def self.patch
+    url_patch1 = "https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/067ac4b5060c16e6687a97cbb4bcdbaf5a0b5639/x11-base/xwayland/files/0001-Eliminate-conflict-with-X11-Xlib.h-with-khronos-eglp.patch?format=TEXT"
+    uri_patch1 = URI.parse url_patch1
+    filename_patch1 = 'patch1_base64'
+    sha256sum_patch1 = '16f5d5d1986daf07b4e0f01eb0e68b53d8ef301d169f3cc9493158c8b4987f16'
+
+    if File.exist?(filename_patch1) && Digest::SHA256.hexdigest( File.read("./#{filename_patch1}") ) == sha256sum_patch1
```
uberhacker, line 44: This should also work:
```
if File.exist?(filename_patch1) && Digest::SHA256.hexdigest( File.read( filename_patch1 ) ) == sha256sum_patch1
```

packages/xwayland.rb
```diff
@@ -0,0 +1,174 @@
...
+      puts "Unpacking patch1 source code".yellow
+    else
```
uberhacker, line 46: Why is this `else` line here?  Not following.

packages/xwayland.rb
```diff
@@ -0,0 +1,174 @@
...
+      system 'patch -p 1 < patch1'
+    end
+    
+
+    url_patch2 = "https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/067ac4b5060c16e6687a97cbb4bcdbaf5a0b5639/x11-base/xwayland/files/0001-xwayland-Fall-back-to-gbm_bo_create-if-no-modifiers-.patch?format=TEXT"
+    uri_patch2 = URI.parse url_patch2
+    filename_patch2 = 'patch2_base64'
+    sha256sum_patch2 = 'd44bfca0e26ca06892b37b12faca2a6756cdc2455aac36ac7607bfa188856b35'
+
+    if File.exist?(filename_patch2) && Digest::SHA256.hexdigest( File.read("./#{filename_patch2}") ) == sha256sum_patch2
```
uberhacker, line 62: I guess I'm confused why you would ever need this conditional.

packages/sommelier.rb
satmandu: No new updates for the past 5 weeks:

image

packages/sommelier.rb
satmandu: I'm not going to go into how much time I wasted trying to get gtkspell working with geany yesterday...

packages/sommelier.rb
satmandu: done

packages/sommelier.rb
satmandu: done

packages/sommelier.rb
satmandu: done

packages/sommelier.rb
satmandu: Fixed.

packages/sommelier.rb
satmandu: done.

packages/sommelier.rb
```diff
@@ -2,106 +2,266 @@
...
+    # to look for the file locally.
+    ######################### Download virtwl.h from Chromium 5.4 kernel tree ###########################################
+    url_virtwl = "https://chromium.googlesource.com/chromiumos/third_party/kernel/+/5d641a7b7b64664230d2fd2aa1e74dd792b8b7bf/include/uapi/linux/virtwl.h?format=TEXT"
+    uri_virtwl = URI.parse url_virtwl
+    #filename_virtwl = File.basename(uri_virtwl.path)
+    filename_virtwl = 'virtwl.h_base64'
+    sha256sum_virtwl = 'a8215f4946ccf30cbd61fcf2ecc4edfe6d05bffeee0bacadd910455274955446'
+    
+    if File.exist?(filename_virtwl) && Digest::SHA256.hexdigest( File.read( filename_virtwl ) ) == sha256sum_virtwl
+      puts "Unpacking virtwl source code".yellow
```
uberhacker, line 48: Still a bit foggy on this logic.  What are we "unpacking" here?  How will this file ever exist until it is downloaded via curl first?

packages/xwayland.rb
satmandu: Changed!

packages/xwayland.rb
satmandu: The rationale for this was that this might prevent redownload if doing a ```crew install -k``` followed by a ```crew install```, but I've taken out that logic, as it does make it cleaner.

packages/xwayland.rb
satmandu: Agreed, and removed.

packages/sommelier.rb
satmandu: This is due to the logic being cribbed from another package file that was downloading patches, IIRC.

satmandu:
@uberhacker Unless you have any other issues, this should be good to go.

packages/sommelier.rb
uberhacker: Can we remove this conditional like the other package?

satmandu: Done.

uberhacker:
Thank you for all the hard work you put into this update @satmandu!  I might regret it, but here goes the merge. :)

wiredin:
First want to say, amazing work! Very exciting to see a recent build of sommelier running on bare metal chromeos! I've got a pretty slow laptop and crostini/container stuff just aint cutting it.

I'm wondering what is used to test that hardware acceleration is indeed working? I installed sommelier on my HP Chromebook 14 **Firmware:** snappy  **Arch**: x86_64 **Kernal:** 4.4 (Intel ApolloLake)

Sommelier seems to be working (Although i need to manually export the DISPLAY env var, as it's not set for me) I'm not sure how I can tell if hardware acceleration is working. It doesn't seem to be working with firefox, when I launch it and go to about:support I see the following:

Screenshot 2020-11-29 at 10 45 56 PM

Also get the following error message.

chronos@localhost ~ $ firefox
error: Kernel is too old for Iris. Consider upgrading to kernel v4.16.

libGL error: failed to create dri screen
libGL error: failed to load driver: iris

This may be related to the firefox package and not sommelier.

satmandu:
Did you add the lines from the postinstall section to your .bashrc? By sourcing ```.sommelier.env``` you get the right DISPLAY variable and also the mesa option to disable IRIS and use the i965 driver if the kernel is < 4.16. The fact that it is trying to use the IRIS driver to accelerate is a good sign though!

```glxinfo -B``` from ```mesa_utils``` (I think) will show you whether acceleration is enabled.

wiredin:
Ugh, yes it was just a case of not reading.... looks like you post a message after installation to do this but alas... I skipped through that. It is working now! Thanks and Great work!! Looks like there is reason to build VLC player now!!

satmandu:
The big issue I still have is sommelier not surviving a sleep/suspend cycle. I have to run ```restartsommelier``` every time I come back from a suspend, and no X apps survive the suspend.

But apparently people with kernels newer than 4.4 may not have this problem...

PR #4452, opened 2020-10-19
xorg_proto update to 2020.1, and build with meson & llvm.

satmandu:
Update to 2020.1, and use meson, clang.

(Using clang when possible as part of https://github.com/skycocker/chromebrew/pull/4450 )

Works properly:
- [x] x86_64

### DIFF
--- a/packages/xorg_proto.rb
+++ b/packages/xorg_proto.rb
@@ -3,35 +3,35 @@ require 'package'
 class Xorg_proto < Package
   description 'The xorgproto package provides the header files required to build the X Window system, and to allow other applications to build against the installed X Window system.'
   homepage 'https://www.x.org/'
-  version '2019.2'
+  version '2020.1'
   compatibility 'all'
-  source_url 'https://xorg.freedesktop.org/archive/individual/proto/xorgproto-2019.2.tar.bz2'
-  source_sha256 '46ecd0156c561d41e8aa87ce79340910cdf38373b759e737fcbba5df508e7b8e'
+  source_url 'https://xorg.freedesktop.org/archive/individual/proto/xorgproto-2020.1.tar.bz2'
+  source_sha256 '54a153f139035a376c075845dd058049177212da94d7a9707cf9468367b699d2'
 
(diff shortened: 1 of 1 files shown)


packages/xorg_proto.rb
```diff
@@ -3,35 +3,35 @@
...
-           "--datadir=#{CREW_LIB_PREFIX}",   # install *.pc in #{CREW_LIB_PREFIX}/pkgconfig
-           '--enable-strict-compilation',
-           '--enable-specs'
-    system 'make'
+    ENV['CC'] = 'clang'
+    ENV['CXX'] = 'clang'
+    system "meson",
+	    "-Dprefix=#{CREW_PREFIX}",
+	    "-Dlibdir=#{CREW_LIB_PREFIX}",
+	    "-Ddatadir=#{CREW_LIB_PREFIX}",
```
uberhacker, line 27: Is this the correct location for data?  What is the default?

uberhacker: Never mind.  I see the comment in the previous commit.

packages/xorg_proto.rb
```diff
@@ -3,35 +3,35 @@
...
-     x86_64: '14fec9e0ec5ad52bbf1035a080f2c16be9533e2090b78f3d06336990ebd2b99f',
-  })
 
+  depends_on 'meson' => ':build'
+  depends_on 'llvm' => ':build'
+  
+   def self.patch
+	  # XKB: Mark key name arrays as not NUL-terminated strings
+	  patch_url = 'https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/commit/21e05da99c855481b638a7789ad5600b5b283680.patch'
+	  patch_sha256 = 'd3c39abc92c098107b21081ff5f2971a6e43b616e2e268057f1f89e49d65856f'
```
uberhacker, line 18: This patch was never applied?

satmandu: Looks like there was a second patch upstream too. Just going to delete this patch.

PR #4454, opened 2020-10-19
libx11 Update to 1.6.12, use clang linker

satmandu:

Works properly:
- [x] x86_64

### DIFF
--- a/packages/libx11.rb
+++ b/packages/libx11.rb
@@ -3,29 +3,20 @@ require 'package'
 class Libx11 < Package
   description 'C interface to the X window system'
   homepage 'https://x.org'
-  version '1.6.9'
+  version '1.6.12'
   compatibility 'all'
-  source_url 'https://www.x.org/archive/individual/lib/libX11-1.6.9.tar.bz2'
-  source_sha256 '9cc7e8d000d6193fa5af580d50d689380b8287052270f5bb26a5fb6b58b2bed1'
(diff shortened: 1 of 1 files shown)


packages/libx11.rb
```diff
@@ -3,29 +3,20 @@
...
-  })
 
   depends_on 'xorg_proto'
   depends_on 'libxcb'
   depends_on 'libxtrans'
+  depends_on 'llvm' => ':build'
   
   def self.build
+    ENV['CFLAGS'] = "-fuse-ld=lld"
+    ENV['CXXFLAGS'] = "-fuse-ld=lld"
```
uberhacker, line 19: Does this break arm builds?

satmandu: Just got an armv7l-ish builder setup using this: https://gist.github.com/satmandu/791dd6e982d365dda6bbba5c332f0f73

Testing various compiles on it now.

packages/libx11.rb
```diff
@@ -3,29 +3,20 @@
...
     system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
```
uberhacker, line 20: Use `#{CREW_OPTIONS}`?

packages/libx11.rb
satmandu: works fine

packages/libx11.rb
satmandu: done

PR #4455, opened 2020-10-19
Libxext Update to 1.3.4, use clang linker

satmandu:

Works properly:
- [x] x86_64

### DIFF
--- a/packages/libxext.rb
+++ b/packages/libxext.rb
@@ -13,7 +13,7 @@ class Libxext < Package
   def self.build
     ENV['CFLAGS'] = "-fuse-ld=lld"
     ENV['CXXFLAGS'] = "-fuse-ld=lld"
-    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
+    system "./configure #{CREW_OPTIONS}"
     system "make"
   end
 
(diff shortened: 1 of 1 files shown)


packages/libxext.rb
```diff
@@ -3,25 +3,16 @@
...
-     x86_64: '2a8953550c8280583385f48d00a044545d01c5e5f994e5ce2171c1d39256a346',
-  })
+  source_url 'https://www.x.org/releases/individual/lib/libXext-1.3.4.tar.gz'
+  source_sha256 '8ef0789f282826661ff40a8eef22430378516ac580167da35cc948be9041aac1'
 
+  depends_on 'llvm' => ':build'
+  
   def self.build
+    ENV['CFLAGS'] = "-fuse-ld=lld"
+    ENV['CXXFLAGS'] = "-fuse-ld=lld"
```
uberhacker, line 15: Does this break arm builds?

packages/libxext.rb
```diff
@@ -3,25 +3,16 @@
...
     system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
```
uberhacker, line 16: Use `#{CREW_OPTIONS}`?

satmandu: done & tested.

PR #4459, opened 2020-10-19
Pixman Update to 0.40.0, use clang linker

satmandu:

Works properly:
- [x] x86_64

### DIFF
--- a/packages/pixman.rb
+++ b/packages/pixman.rb
@@ -13,7 +13,7 @@ class Pixman < Package
   def self.build
     ENV['CFLAGS'] = "-fuse-ld=lld"
     ENV['CXXFLAGS'] = "-fuse-ld=lld"
-    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
+    system "./configure #{CREW_OPTIONS}"
     system 'make'
   end
 
--- a/packages/pixman.rb
+++ b/packages/pixman.rb
@@ -3,25 +3,16 @@ require 'package'
 class Pixman < Package
   description 'Pixman is a low-level software library for pixel manipulation, providing features such as image compositing and trapezoid rasterization.'
   homepage 'http://www.pixman.org/'
-  version '0.34.0-0'
+  version '0.40.0-0'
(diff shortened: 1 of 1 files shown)


packages/pixman.rb
```diff
@@ -3,25 +3,16 @@
...
-    aarch64: '056ddb91ba2761a2c742fe47edb02c9a2dd2a2ae89ff06b5a4733d346ee488ff',
-     armv7l: '056ddb91ba2761a2c742fe47edb02c9a2dd2a2ae89ff06b5a4733d346ee488ff',
-       i686: '07e84363add2a3e345340243af5fe39dc7c2b1d5fd3d5252bb64080b86abf12c',
-     x86_64: 'c268902b79b7dffe7214fb0f561e9ecdb9be47309d2440da3a386cabe3ba4370',
-  })
+  depends_on 'llvm' => ':build'
 
   def self.build
+    ENV['CFLAGS'] = "-fuse-ld=lld"
+    ENV['CXXFLAGS'] = "-fuse-ld=lld"
```
uberhacker, line 15: Does this break arm builds?

satmandu: I'll wait until some version of https://github.com/skycocker/chromebrew/pull/4483 is merged to figure out if there are issues?

The armv7l image I am using from https://gist.github.com/satmandu/791dd6e982d365dda6bbba5c332f0f73 hasn't been complaining when it sees armv7l as the ARCH and #{CREW_OPTONS} is set for the apps I have tested so far.

packages/pixman.rb
```diff
@@ -3,25 +3,16 @@
...
     system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
```
uberhacker, line 16: Does `#{CREW_OPTIONS}` work?

packages/pixman.rb
satmandu: gcc10 should fix these issues.

packages/pixman.rb
satmandu: yup

PR #4460, opened 2020-10-19
libxrandr update to 1.5.2, use clang linker

satmandu:
part of https://github.com/skycocker/chromebrew/pull/4450

Works properly:
- [x] x86_64

### DIFF
--- a/packages/libxrandr.rb
+++ b/packages/libxrandr.rb
@@ -3,28 +3,19 @@ require 'package'
 class Libxrandr < Package
   description 'X.org X Resize, Rotate and Reflection extension library'
   homepage 'http://www.x.org'
-  version '1.5.1-0'
+  version '1.5.2-0'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


packages/libxrandr.rb
```diff
@@ -3,28 +3,19 @@
...
-     x86_64: '6a37502061445761f860acefaddd5bf06b7c64b4b5e468f4e3902e67baaef05e',
-  })
 
   depends_on 'libxrender'
   depends_on 'libx11'
+  depends_on 'llvm' => ':build'
 
   def self.build
+    ENV['CFLAGS'] = "-fuse-ld=lld"
+    ENV['CXXFLAGS'] = "-fuse-ld=lld"
```
uberhacker, line 18: Does this break arm builds?

satmandu: working fine with gcc10.

uberhacker: I'm wondering if we should change the default compiler from 8 to 10.

PR #4456, opened 2020-10-19
Harfbuzz update to 2.7.2, use meson

satmandu:
Part of https://github.com/skycocker/chromebrew/pull/4450 and upcoming adwaita_icon_theme PR

There MAY be circular dependencies here...   Do we need a harfbuzz_sub package?

Works properly:
- [x] x86_64

PR #4461, opened 2020-10-19
Pango update to 1.47, use clang linker

satmandu:
Part of upcoming adwaita PR.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/pango.rb
+++ b/packages/pango.rb
@@ -23,8 +23,7 @@ class Pango < Package
     ENV['CFLAGS'] = "-fuse-ld=lld"
     ENV['CXXFLAGS'] = "-fuse-ld=lld"
     system "meson",
-      "-Dprefix=#{CREW_PREFIX}",
-      "-Dlibdir=#{CREW_LIB_PREFIX}",
+      #{CREW_MESON_OPTIONS},
       "-Dinstall-tests=false",
       "builddir"
     system "ninja -C builddir"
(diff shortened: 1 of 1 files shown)


packages/pango.rb
```diff
@@ -3,34 +3,30 @@
...
   depends_on 'libxrender'
-  depends_on 'fribidi'
+  depends_on 'fribidi' # Gets built inside install automatically.
   depends_on 'six'
+  depends_on 'llvm' => ':build'
 
   def self.build
-    system "meson --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} builddir"
+    ENV['CFLAGS'] = "-fuse-ld=lld"
+    ENV['CXXFLAGS'] = "-fuse-ld=lld"
```
uberhacker, line 24: Isn't this going to break arm builds?  That has been my experience with all other packages that included these options.

satmandu: I'm happy to try these on an arm build. What is the target machine so I can pull down the recovery image and test?

uberhacker: Target would be `armv7l`.

packages/pango.rb
```diff
@@ -3,34 +3,30 @@
...
+    system "meson",
+      "-Dprefix=#{CREW_PREFIX}",
+      "-Dlibdir=#{CREW_LIB_PREFIX}",
```
uberhacker, line 27: Can we use the new `CREW_MESON_OPTIONS` here?

satmandu: done

satmandu:
FYI this needs the new harfbuzz in https://github.com/skycocker/chromebrew/pull/4456 .

PR #4462, opened 2020-10-19
Add Vulkan

satmandu:
Adds Vulkan headers & ICD, which mesa looks for.

Works properly:
- [x] x86_64


uberhacker:

packages/vulkan_icd_loader.rb
```diff
@@ -0,0 +1,52 @@
+# Adapted from Arch Linux vulkan-icd-loader PKGBUILD at:
+# https://github.com/archlinux/svntogit-packages/raw/packages/vulkan-icd-loader/trunk/PKGBUILD
+
+require "package"
+
+class Vulkan_icd_loader < Package
+  description "Vulkan Installable Client Driver ICD Loader"
+  version "1.2.153-2"
```
uberhacker, line 8: This is missing `compatibility 'all'`.  We need a quick fix for this.

PR #4463, opened 2020-10-19
Add glslang

satmandu:
Add glslang package which mesa looks for.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/glslang.rb
+++ b/packages/glslang.rb
@@ -10,10 +10,6 @@ class Glslang < Package
   source_url 'https://github.com/KhronosGroup/glslang/archive/8.13.3743.tar.gz'
   source_sha256 '639ebec56f1a7402f2fa094469a5ddea1eceecfaf2e9efe361376a0f73a7ee2f'
 
-  depends_on 'python3'
-  depends_on 'cmake' => ':build'
-  depends_on 'ninja' => ':build'
-
   def self.build
     system "./update_glslang_sources.py"
     
--- a/packages/glslang.rb
+++ b/packages/glslang.rb
@@ -1,0 +1,46 @@
+# Adapted from Arch Linux glslang PKGBUILD at:
+# https://github.com/archlinux/svntogit-packages/raw/packages/glslang/trunk/PKGBUILD
+
+require 'package'
+
+class Glslang < Package
+  description 'OpenGL and OpenGL ES shader front end and validator'
+  version '8.13.3743'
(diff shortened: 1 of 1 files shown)


packages/glslang.rb
```diff
@@ -0,0 +1,46 @@
+# Adapted from Arch Linux glslang PKGBUILD at:
+# https://github.com/archlinux/svntogit-packages/raw/packages/glslang/trunk/PKGBUILD
+
+require 'package'
+
+class Glslang < Package
+  description 'OpenGL and OpenGL ES shader front end and validator'
+  version '8.13.3743'
```
uberhacker, line 8: This is missing `compatibility 'all'`.

packages/glslang.rb
```diff
@@ -0,0 +1,46 @@
...
+  source_url 'https://github.com/KhronosGroup/glslang/archive/8.13.3743.tar.gz'
+  source_sha256 '639ebec56f1a7402f2fa094469a5ddea1eceecfaf2e9efe361376a0f73a7ee2f'
+
+  depends_on 'python3'
+  depends_on 'cmake' => ':build'
+  depends_on 'ninja' => ':build'
```
uberhacker, line 14: These are all core packages so no need to list here.

packages/glslang.rb
```diff
@@ -0,0 +1,46 @@
...
+    Dir.mkdir 'build-static'
+    Dir.chdir 'build-static' do
+    system "cmake .. \
+      -GNinja \
+      -DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX} \
+      -DCMAKE_INSTALL_LIBDIR=#{CREW_LIB_PREFIX} \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DBUILD_SHARED_LIBS=OFF"
+      end
+    system 'ninja -C build-static'
```
uberhacker, line 39: Any reason to build static libraries?

satmandu: It's cargo cult I know but I'm just copying logic from the arch pkgbuild file at https://github.com/archlinux/svntogit-packages/raw/packages/glslang/trunk/PKGBUILD

packages/glslang.rb
satmandu: fixed

packages/glslang.rb
uberhacker: I guess the minute you don't include it is the time you need it. LoL

Issue #4338, opened 2020-09-24
Need a better dependency resolver


uberhacker:
The resolver works currently, however, it needs to be optimized for dependency order and performance.  For instance, sometimes dependencies are installed in the wrong order when some dependency packages depend on other packages first.  This is not so much an issue when installing from binaries but is especially important when installing from source.  However, there are certain instances where the dependency order matters for binary installs as well.

cstrouse:
Perhaps try to leverage the solver from chef? https://github.com/berkshelf/solve

supechicken:
See #6840

PR #4466, opened 2020-10-19
Libressl update to 3.2.2 (stable)/ Fix regression

satmandu:
Sync with stable version tracking OpenBSD 6.8
Works properly:
- [x] x86_64

### DIFF
--- a/packages/libressl.rb
+++ b/packages/libressl.rb
@@ -5,21 +5,8 @@ class Libressl < Package
   homepage 'https://www.libressl.org/'
   version '3.1.4'
   compatibility 'all'
-  source_url 'https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.1.4.tar.gz'
-  source_sha256 '414c149c9963983f805a081db5bd3aec146b5f82d529bb63875ac941b25dcbb6'
-
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/libressl-3.1.4-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/libressl-3.1.4-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/libressl-3.1.4-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/libressl-3.1.4-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: '0b3da102fede122f12ec32a44cd3e543d72184cd90c9937f6fd32dea6d71c121',
-     armv7l: '0b3da102fede122f12ec32a44cd3e543d72184cd90c9937f6fd32dea6d71c121',
-       i686: '0e3e4d7be5575212286e5cc26c60b7473f448436abda50208a252afd53a1d694',
-     x86_64: '5744988a28dc957ddabd49244c203a4cc43cba5332459669a9a5928628e841e4',
-  })
+  source_url 'https://ftp.openbsd.org/pub/OpenBSD/LibreSSL/libressl-3.2.2.tar.gz'
+  source_sha256 'a9d1e1d030b8bcc67bf6428b8c0fff14a5602e2236257b9e3d77acaf12e2a7a1'
 
   def self.build
     system "./configure #{CREW_OPTIONS} --with-openssldir=/etc/ssl"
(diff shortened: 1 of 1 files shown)


uberhacker:
Didn't we just update libressl?  You do realize this update will break everything.  We'll likely need to recompile python, php (4 versions) and a slew of other packages to get this to work.

satmandu:
I think there was a discussion of keeping some of the security apps up to date? libressl would definitely be one of those.

This doesn't appear to break anything though, as the same library names are used. (But feel free to prove me wrong.)

satmandu:
@uberhacker Do we need more complicated logic in making symlinks to support older installs?

libressl.so.#.0.1 => libressl.so.1.0.0?

packages/libressl.rb
```diff
@@ -24,9 +24,15 @@ def self.install
...
+
+    # add symlink to libcrypto.so.1.0.0
+    #FileUtils.ln_s "#{CREW_LIB_PREFIX}/libcrypto.so.1.0.0", "#{CREW_DEST_LIB_PREFIX}/libcrypto.so.46"
+
+    # add symlink to libssl.so.1.0.0
+    FileUtils.ln_s "#{CREW_LIB_PREFIX}/libssl.so.1.0.0", "#{CREW_DEST_LIB_PREFIX}/libssl.so.47"
 
     # add symlink to libssl.so.1.0.0
-    FileUtils.ln_s "#{CREW_LIB_PREFIX}/libssl.so.48.0.1", "#{CREW_DEST_LIB_PREFIX}/libssl.so.1.0.0"
+    #FileUtils.ln_s "#{CREW_LIB_PREFIX}/libssl.so.1.0.0", "#{CREW_DEST_LIB_PREFIX}/libssl.so.48"
```
uberhacker, line 36: It doesn't hurt to have all these symlinks but a better solution would be to recompile all packages that rely on the older version to link to the newer version.

uberhacker:
@satmandu:  My energy level to test this update is pretty low.  Last time we updated this package, we needed to recompile or update numerous other packages.  It opens a Pandora's box.  This is a core package so we will need binaries for the install along with any other core packages this update affects.

satmandu:
@uberhacker Are you doing all of these compiles on native hardware for each of these systems?


I still think last month's libressl update was fundamentally broken because we didn't create symlinks properly.

 ```   FileUtils.ln_s "#{CREW_LIB_PREFIX}/libssl.so.48.0.1", "#{CREW_DEST_LIB_PREFIX}/libssl.so.1.0.0"``` is wrong.
That symlink is _automatically created anyways_. 

**What we need is** 
 ```   FileUtils.ln_s "#{CREW_LIB_PREFIX}/libssl.so.1.0.0", "#{CREW_DEST_LIB_PREFIX}/libssl.so.47"``` so that older installs would keep working. Instead the update borked any program looking for ```#{CREW_DEST_LIB_PREFIX}/libssl.so.47```.

If you don't want to put this in now, you might want to rebuild the libressl 3.1.4 package anyways so that symlink is there, for anyone who upgrades in the meantime.

At least, that's my thinking.

uberhacker:
A more effective solution would be to recompile the affected packages.  The main question:  How many other packages are affected?  It may be in the dozens.  The good thing is most of those packages probably need updated to newer versions anyway.

vincowl:
Just a quick intrusion : when we updated `qtbase` we had to disable openssl support since qt doesn't support libressl. Recent libressl update didn't solve the problem. Keeping stuck with libressl will imply no ssl support in the qt packages.... If everything should be recompiled, won't it be a good opportunity to roll back to openssl ? Do we have benefits that I don't get (other than philosophical ones) to stay with libressl ?

satmandu:
> A more effective solution would be to recompile the affected packages. The main question: How many other packages are affected? It may be in the dozens. The good thing is most of those packages probably need updated to newer versions anyway.

Best way would be to spin up a container and just recursively build all dependencies.

If you use my script and go to /output/ and then do a recursive build of all downstream dependencies, all the resultant files will drop in the folder you invoked the docker container from.

Ideally one would have a list of symbols  that apps and libraries in a package both use and provide, and one could easily figure out what the dependencies are. Debian does that well...

uberhacker:
We definitely need a better dependency resolver.  That is [one of the Hacktoberfest challenges](https://github.com/skycocker/chromebrew/issues/4338).  You would be a hero if you can figure it out.

satmandu:
I feel like this is a problem that has been solved by multiple distributions already. Why reinvent the wheel?

I'm in the automate or reuse code as much as possible to avoid busywork camp.

(This is also why I was happy to spend time to setup docker containers so I wouldn't have to do the same work over and over again.)

uberhacker:
OK, we won't reinvent the wheel.  Now figure out a way to leverage how other package managers handle dependencies and incorporate in this project. :)

satmandu:
See https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dependency_collector.rb
but also https://blog.jpalardy.com/posts/untangling-your-homebrew-dependencies/

Isn't chromebrew spiritually a descendant of the homebrew project? It's also in ruby, so it seems it would be _easiest_ [not actually easy] to leverage their infrastructure? At least their setup is also rpedicated upon installing alongside an existing system instead of using a chroot.

uberhacker:
Addressed in #4532.

PR #4467, opened 2020-10-19
libRsvg Update to 2.50.1

satmandu:
Update LibRsvg to 2.50.1 and fix logic in case Rust install is broken.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/librsvg.rb
+++ b/packages/librsvg.rb
@@ -1,50 +1,43 @@
 require 'package'
 
 class Librsvg < Package
-  description 'Scalable Vector Graphics (SVG) rendering library.'
+  description 'SVG library for GNOME'
   homepage 'https://wiki.gnome.org/Projects/LibRsvg'
-  version '2.42.3'
+  version '2.50.1'
   compatibility 'all'
-  source_url 'https://download.gnome.org/sources/librsvg/2.42/librsvg-2.42.3.tar.xz'
-  source_sha256 '704f2c44b9b170fc5498de36a161d01ca8f584ba9c42654b98565a7b7bcbe657'
+  source_url 'https://ftp.gnome.org/pub/GNOME/sources/librsvg/2.50/librsvg-2.50.1.tar.xz'
+  source_sha256 '9bd9799322e06cf5db19b9f7afb728edac6efcf0110baafc44f0f96f45df9a09'
 
-  binary_url ({
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/librsvg-2.42.3-chromeos-x86_64.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/librsvg.rb
```diff
@@ -1,50 +1,43 @@
...
-  description 'Scalable Vector Graphics (SVG) rendering library.'
+  description 'SVG library for GNOME'
   homepage 'https://wiki.gnome.org/Projects/LibRsvg'
-  version '2.42.3'
+  version '2.50.1'
   compatibility 'all'
-  source_url 'https://download.gnome.org/sources/librsvg/2.42/librsvg-2.42.3.tar.xz'
-  source_sha256 '704f2c44b9b170fc5498de36a161d01ca8f584ba9c42654b98565a7b7bcbe657'
+  source_url 'https://ftp.gnome.org/pub/GNOME/sources/librsvg/2.50/librsvg-2.50.1.tar.xz'
+  source_sha256 '9bd9799322e06cf5db19b9f7afb728edac6efcf0110baafc44f0f96f45df9a09'
```
uberhacker, line 9: Why the source location change?

satmandu: I think they are interchangeable. Just the first URL I found when searching for updates when building.

```https://download.gnome.org/sources/librsvg/2.50/librsvg-2.50.1.tar.xz``` works just as well. I'll change it so there is no change in source location.

File: packages/librsvg.rb [hunk 2]
```diff
@@ -1,50 +1,41 @@
...
+  depends_on 'freetype'
+  depends_on 'gdk_pixbuf'
+  depends_on 'libcroco'
+  depends_on 'pango'
   depends_on 'rust' => :build
-  depends_on 'six' => :build
-
+  depends_on 'python3'
+  depends_on 'glib'
+  depends_on 'gdk_pixbuf'
```
uberhacker, line 20: Looks like this is a duplicate of line 14.

packages/librsvg.rb
```diff
@@ -1,50 +1,41 @@
...
   depends_on 'gobject_introspection'
-  depends_on 'gtk3'
+  depends_on 'freetype'
+  depends_on 'gdk_pixbuf'
+  depends_on 'libcroco'
+  depends_on 'pango'
   depends_on 'rust' => :build
-  depends_on 'six' => :build
-
+  depends_on 'python3'
```
uberhacker, line 18: This line is not needed since python3 is already in "core".

packages/librsvg.rb
satmandu: Yup. Fixed.

packages/librsvg.rb
satmandu: Fixed.

PR #4470, opened 2020-10-20
mesa update to 20.2.1, use gcc10 for building

satmandu:
Use as many defaults as possible since meson can figure them out.

Also enable clang lld for non-arm7l since at least on x86_64 this seems to be important to enabling graphics acceleration via sommelier. I believe this is the final part of https://github.com/skycocker/chromebrew/pull/4450 for enabling graphics acceleration.

(If llvm's linker works with a newer GCC on arm7l the conditional switch should be removed and lld should be enabled for al architectures.)

Works properly:
- [x] x86_64

### DIFF
--- a/packages/mesa.rb
+++ b/packages/mesa.rb
@@ -34,9 +34,7 @@ class Mesa < Package
 
     # Just use mostly defaults.
     system "meson",
-      "-Dprefix=#{CREW_PREFIX}",
-      "-Dlibdir=#{CREW_LIB_PREFIX}",
-      "-Dbuildtype=release",
+      #{CREW_MESON_OPTIONS},
       "builddir"
     system "ninja -C builddir"
   end
--- a/packages/mesa.rb
+++ b/packages/mesa.rb
@@ -9,6 +9,7 @@ class Mesa < Package
   source_sha256 'd1a46d9a3f291bc0e0374600bdcb59844fa3eafaa50398e472a36fc65fd0244a'
   
 
+  depends_on 'gcc10' => :build
   depends_on 'setuptools' => :build
   depends_on 'libva'
   depends_on 'libvdpau'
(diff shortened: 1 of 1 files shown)


packages/mesa.rb
```diff
@@ -3,54 +3,40 @@
...
-      "-Dgles1=auto",
-      "-Dgles2=auto",
-      "-Dplatforms=x11,drm,wayland",
-      "-Dgbm=auto",
-      "-Dgallium-xa=auto",
-      "-Dglx=auto",
-      "-Dllvm=true",
+      "-Dprefix=#{CREW_PREFIX}",
+      "-Dlibdir=#{CREW_LIB_PREFIX}",
+      "-Dbuildtype=release",
```
uberhacker, line 39: Are all the drivers enabled by default then?

satmandu: So this seemed to be building all of the relevant drivers for x86_64, but I believe one can check that by adding a ```meson configure builddir``` line to show the configuration before compilation starts.

packages/mesa.rb
```diff
@@ -3,54 +3,40 @@
...
   depends_on 'elfutils'
-  depends_on 'llvm'
+  depends_on 'llvm' => ':build'
+  depends_on 'meson' => ':build'
   depends_on 'libdrm'
+  depends_on 'libunwind' => ':build'
+  depends_on 'vulkan_headers' => ':build'
+  depends_on 'vulkan_icd_loader' => ':build'
+  depends_on 'glslang' => ':build'
+  depends_on 'zstd' => ':build'
```
uberhacker, line 24: No single quotes needed around `:build`.

satmandu:
@uberhacker I took out the conditions for armv7l. It compiles just fine with gcc10. I added that as a build requirement.

packages/mesa.rb
uberhacker: Does the new `CREW_MESON_OPTIONS` constant work here?

satmandu: Let me try it. :)

satmandu: Works great!

packages/mesa.rb
satmandu: done

PR #4473, opened 2020-10-20
[firefox] Update to 82.0

vincowl:


uberhacker:
Thank you for the update @vincowl!  Do we need some extra config to get sound working?

vincowl:
Just found out how to setup sound. A few work more to add to this PR including a new package, env vars to set and firefox preferences to set via a script. Wow ! At least we found a way through this sound problem. So don't merge this PR yet

vincowl:
@uberhacker OK, that should work. Please test this and listen ;-)

packages/firefox.rb
```diff
@@ -3,27 +3,28 @@
...
-    source_sha256 '7882ce6ece0b1aeb497d68e8aeff2c58eb5a926fb5003472db493c863e3129f9'
+    source_url 'https://archive.mozilla.org/pub/firefox/releases/82.0/linux-x86_64/en-US/firefox-82.0.tar.bz2'
+    source_sha256 '4305f43e72aa46ae42882eaa4e436b22c5142a4c9b2dc9f9ccc3292e18a5e39a'
   end
 
   case ARCH
   when 'i686', 'x86_64'
     depends_on 'gtk3'
     depends_on 'sommelier'
+    depends_on 'apulse'
```
uberhacker, line 21: Can we put this dependency above sommelier?  If somebody doesn't have sommelier installed yet, we want them to see the postinstall instructions.

uberhacker:
@vincowl: This is ear candy. ;-)  Thank you!

packages/firefox.rb
vincowl: Done !

PR #4478, opened 2020-10-21
Use builtin curl & have curl retry downloads

satmandu:
I'm getting tons of download failures on a sketchy connection. It would be nice if curl retried connections from crew commands.

Crew also crashes if I try to update libiconv. Can we just use the system ```/usr/bin/curl``` from crew? Are there any systems without a working built-in curl?

Works properly:
- [x] x86_64
- [x] arm7l (only tested in build container)

### DIFF
--- a/crew
+++ b/crew
@@ -494,9 +494,9 @@ def download
   end
   Dir.chdir CREW_BREW_DIR do
     if @opt_verbose then
-      system('curl', '-v', '--progress-bar', '-C', '-', '--insecure', '-L', '-#', url, '-o', filename)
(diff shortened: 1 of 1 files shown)


uberhacker:
If we update crew, we need to bump the version in `/path/to/chromebrew/source/lib/const.rb`.  I like the idea of using the system curl for stability.  Also, I'd like to know how you are running arm in a container.  Can to share your Dockerfile?

satmandu:
https://gist.github.com/satmandu/791dd6e982d365dda6bbba5c332f0f73

satmandu:
@uberhacker Can you squash https://github.com/skycocker/chromebrew/pull/4484 https://github.com/skycocker/chromebrew/pull/4483 and this together?

uberhacker:
@satmandu: Good call.

PR #4482, opened 2020-10-21
Added instructions and all-in-one script to use i3 windows manager

supechicken666:
Added instructions and all-in-one script about setting up XServer XSDL display environment

### DIFF
--- a/packages/i3.rb
+++ b/packages/i3.rb
@@ -32,7 +32,7 @@ class I3 < Package
      system "echo 'export DISPLAY=100.115.92.2:0' >> starti3"
      system "echo 'i3 \"$@\"' >> starti3"
      system "echo 'export DISPLAY=:0' >> starti3"
-     system "echo 'initsommelier' >> starti3"
+     system "echo 'startsommelier' >> starti3"
      system "chmod +x starti3"
     end
   end
--- a/packages/i3.rb
+++ b/packages/i3.rb
@@ -47,6 +47,7 @@ class I3 < Package
       puts
       puts "To use this package, you need to download XServer XSDL from Google Play Store".lightblue
       puts "Use 'starti3' instead of 'i3' to execute this package".lightblue
+      puts "Open XServer XSDL before running 'starti3'"
       system "touch /usr/local/bin/starti3"
       system "echo '#!/bin/sh' >> /usr/local/bin/starti3"
       system "echo 'stopsommelier' >> /usr/local/bin/starti3"
--- a/packages/i3.rb
+++ b/packages/i3.rb
@@ -27,12 +27,13 @@ class I3 < Package
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
     Dir.chdir "#{CREW_DEST_PREFIX}/bin" do
       system "for f in \$(ls #{CREW_BUILD}-*); do g=\$(echo \$f | sed 's,#{CREW_BUILD}-,,'); ln -sf \$f \$g; done"
-     system "touch /usr/local/bin/starti3"
-     system "echo '#!/bin/sh' >> /usr/local/bin/starti3"
-     system "echo 'stopsommelier' >> /usr/local/bin/starti3"
-     system "echo 'export DISPLAY=100.115.92.2:0' >> /usr/local/bin/starti3"
-     system "echo 'i3 & $1 $2 $3 $4' >> /usr/local/bin/starti3"
-     system "chmod +x /usr/local/bin/starti3"
+     system "echo '#!/bin/sh' > starti3"
+     system "echo 'stopsommelier' >> starti3"
+     system "echo 'export DISPLAY=100.115.92.2:0' >> starti3"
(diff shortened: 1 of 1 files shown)


packages/i3.rb
```diff
@@ -42,4 +42,16 @@ def self.install
...
+  def self.postinstall
+      puts
+      puts "To use this package, you need to download XServer XSDL from Google Play Store".lightblue
+      puts "Use 'starti3' instead of 'i3' to execute this package".lightblue
+      system "touch /usr/local/bin/starti3"
+      system "echo '#!/bin/sh' >> /usr/local/bin/starti3"
+      system "echo 'stopsommelier' >> /usr/local/bin/starti3"
+      system "echo 'export DISPLAY=100.115.92.2:0' >> /usr/local/bin/starti3"
+      system "echo 'i3 & $1 $2 $3 $4' >> /usr/local/bin/starti3"
+      system "chmod +x /usr/local/bin/starti3"
```
uberhacker, line 55: Can we add the script to the install section?  That way it can be removed when the package is removed and requires less configuration for the user.  Might as well update the package while you're at it also.  https://i3wm.org/downloads/i3-4.18.3.tar.bz2

supechicken666: Done

packages/i3.rb
```diff
@@ -40,6 +27,19 @@ def self.install
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
     Dir.chdir "#{CREW_DEST_PREFIX}/bin" do
       system "for f in \$(ls #{CREW_BUILD}-*); do g=\$(echo \$f | sed 's,#{CREW_BUILD}-,,'); ln -sf \$f \$g; done"
+     system "touch /usr/local/bin/starti3"
+     system "echo '#!/bin/sh' >> /usr/local/bin/starti3"
+     system "echo 'stopsommelier' >> /usr/local/bin/starti3"
+     system "echo 'export DISPLAY=100.115.92.2:0' >> /usr/local/bin/starti3"
+     system "echo 'i3 & $1 $2 $3 $4' >> /usr/local/bin/starti3"
+     system "chmod +x /usr/local/bin/starti3"
```
uberhacker, line 35: You could replace lines 30-35 with:
```
     system "echo '#!/bin/sh' > starti3"
     system "echo 'stopsommelier' >> starti3"
     system "echo 'export DISPLAY=100.115.92.2:0' >> starti3"
     system "echo 'i3 & $1 $2 $3 $4' >> starti3"
     system "chmod +x starti3"
```

packages/i3.rb
```diff
@@ -40,6 +27,19 @@ def self.install
...
     end
   end
+  
+  def self.postinstall
+      puts
+      puts "To use this package, you need to download XServer XSDL from Google Play Store".lightblue
+      puts "Use 'starti3' instead of 'i3' to execute this package".lightblue
+      puts "Open XServer XSDL before running 'starti3'"
```
uberhacker, line 43: Might want to mention instructions on how to restore to using sommelier.  ie. `export DISPLAY=:0`, for example.

supechicken666: Added to the script directly

uberhacker: Have you tested this and run spacefm immediately afterwards to see if it launches?

packages/i3.rb
```diff
@@ -40,6 +27,20 @@ def self.install
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
     Dir.chdir "#{CREW_DEST_PREFIX}/bin" do
       system "for f in \$(ls #{CREW_BUILD}-*); do g=\$(echo \$f | sed 's,#{CREW_BUILD}-,,'); ln -sf \$f \$g; done"
+     system "echo '#!/bin/sh' > starti3"
+     system "echo 'stopsommelier' >> starti3"
+     system "echo 'export DISPLAY=100.115.92.2:0' >> starti3"
+     system "echo 'i3 \"$@\"' >> starti3"
+     system "echo 'export DISPLAY=:0' >> starti3"
+     system "echo 'initsommelier' >> starti3"
```
uberhacker, line 35: `system "echo 'startsommelier'` >> starti3"` will likely work better here.

uberhacker: Might want to check if sommelier is installed first in the script since this package does not include sommelier as a dependency.  Otherwise, the sommelier commands will fail ungracefully.

packages/i3.rb
supechicken666: I have tested it with `starti3 firefox` and it works well but haven't test it with spacefm yet

packages/i3.rb
supechicken666: `sommelier` isn't already in the dependencies?

supechicken666: > `system "echo 'startsommelier' >> starti3"` will likely work better here.

Done

packages/i3.rb
uberhacker: You can test with any gui app really.  I just used spacefm as an example.

PR #4484, opened 2020-10-22
Add CREW_CMAKE_OPTIONS & CREW_MESON_OPTIONS constants

satmandu:
meson builds can not use #{CREW_OPTIONS}

Note that this needs to be used like this:
```
    system "meson",
            #{CREW_MESON_OPTIONS},
            "build"
    system "meson configure build"
    system "ninja -C build" 
```

(No quotes around the line otherwise it does not work. #{CREW_OPTIONS} has the same issue.)

Works properly:
- [x] x86_64
- [x] armv7l


lib/const.rb
```diff
@@ -57,3 +57,4 @@
   CREW_BUILD = 'x86_64-cros-linux-gnu'
 end
 CREW_OPTIONS = "--prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --mandir=#{CREW_MAN_PREFIX} --build=#{CREW_BUILD} --host=#{CREW_BUILD} --target=#{CREW_BUILD}"
+CREW_MESON_OPTIONS = "-Dprefix=#{CREW_PREFIX} -Dlibdir=#{CREW_LIB_PREFIX} -Ddatadir=#{CREW_LIB_PREFIX} -Dmandir=#{CREW_MAN_PREFIX} -Dbuild=#{CREW_BUILD} -Dhost=#{CREW_BUILD} -Dtarget=#{CREW_BUILD} -Dbuildtype=release"
```
uberhacker, line 60: Nice!  We could probably add `CREW_PYTHON_OPTIONS` as well in the future.  Please bump the crew version at the top of this file.

uberhacker: We'll need to add this new constant above line 325 in the `crew` ruby script file as well.

uberhacker: Come to think of it, we could also add `CREW_CMAKE_OPTIONS`.

satmandu: I've also been wondering if it would make sense to drop an LDFLAGS option into the OPTIONS lines as well.

e.g.
```
CREW_OPTIONS = "--prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --mandir=#{CREW_MAN_PREFIX} --build=#{CREW_BUILD} --host=#{CREW_BUILD} --target=#{CREW_BUILD} LDFLAGS=\"-L#{CREW_LIB_PREFIX} -L/usr/#/{ARCH_LIB} -L/#{ARCH_LIB} -L/usr/lib -L/lib\""
```

satmandu: Constants added to ```crew``` ruby script as well.

crew
```diff
@@ -325,6 +325,8 @@ def const (var)
       'CREW_NOT_COMPRESS',
       'CREW_NOT_STRIP',
       'CREW_NPROC',
+      'CREW_CMAKE_OPTIONS',
+      'CREW_MESON_OPTIONS',
```
uberhacker, line 329: The constants array is sorted alphabetically.  Please reposition.

satmandu: done.

lib/const.rb
satmandu: FYI LDFLAGS just doesn't work there.

PR #4490, opened 2020-10-23
Gtk3: clang doesn't guarantee lld, so just use lld

satmandu:
Looks like clang isn't guaranteeing use of the lld linker, so just use lld.

Maybe don't bother bumping version unless we need to regenerate binaries.

Works properly:
- [x] x86_64
- [x] armv7l


uberhacker:
We should bump the version to `version '3.24.23-1'` since we need new binaries.

satmandu:
version bumped.

uberhacker:
Remove old binaries also?

satmandu:
Added gcc10 dep as well since we have seen issues with gcc8 & using lld.

PR #4491, opened 2020-10-23
Update Opera to 72.0.3815.148

supechicken666:
## Description
Update Opera from 71.0.3770.271 to 72.0.3815.148

Create symlink from `opera` to `/usr/bin/x-www-browser`

Tested on x86_64

### DIFF
--- a/packages/opera.rb
+++ b/packages/opera.rb
@@ -3,7 +3,7 @@ require 'package'
 class Opera < Package
   description "Opera isn't just a browser. It's an awesome browser."
   homepage 'https://www.opera.com/'
-  version '71.0.3770.271'
+  version '72.0.3815.148'
   compatibility 'x86_64'
   case ARCH
   when 'x86_64'
@@ -18,7 +18,6 @@ class Opera < Package
 
   def self.build
     system "wget https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb"
-    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("opera-stable_#{version}_amd64.deb") ) == 'cad630af33ce7d54de802abd9c360b19d330ef64759888da1ec03e5dd7e54f9a'
     system "alien -tc opera-stable_#{version}_amd64.deb"
     system "tar xvf opera-stable-#{version}.tgz"
   end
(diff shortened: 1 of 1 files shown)


packages/opera.rb
```diff
@@ -28,5 +28,7 @@ def self.install
     FileUtils.mv 'usr/bin', "#{CREW_DEST_PREFIX}"
     FileUtils.mv 'usr/lib', "#{CREW_DEST_PREFIX}"
     FileUtils.mv 'usr/share', "#{CREW_DEST_PREFIX}"
+    system "echo 'if [[ $(ls /usr/bin/ | grep x-www-browser) = x-www-browser ]]; then echo x-www-browser is already set; else sudo ln -s #{CREW_PREFIX}/bin/opera /usr/bin/x-www-browser; fi' > script.la.tmp"
+    system "bash script.la.tmp"
```
uberhacker, line 32: Why not just include the symlink in the package like below:
```
FileUtils.ln_s "#{CREW_PREFIX}/bin/opera", "#{CREW_DEST_PREFIX}/bin/x-www-browser"
```
In any event, `ln -s #{CREW_PREFIX}/bin/opera /usr/bin/x-www-browser` will fail since that directory is read-only.

uberhacker:
Thank you for the update @supechicken666!

packages/opera.rb
supechicken666: You're right, but some apps may use the absolute path to run it like `openbox`, how can we redirect it to /usr/local/bin/ ?

supechicken666:
Can we remount the /usr/bin as rw?

uberhacker:
That is not advisable.  You could corrupt the Chrome OS files and brick your Chromebook.

packages/opera.rb
uberhacker: Your `PATH` environment variable should already look for `/usr/local/bin` first so it should be fine.

supechicken666:
Now x-www-browser symlink will put on `/usr/local/bin` instead of `/usr/bin`

packages/opera.rb
supechicken666: But openbox is run the command like `/usr/bin/x-www-browser: command not found`
If can't solve I think it's ok :)

uberhacker: Can't we fix the path since we compile openbox from source?

supechicken666:
This PR is ready to merge

Issue #4486, opened 2020-10-22
[Help wanted] Building openbox


supechicken666:
Currently I am working on a new package `openbox` , but I am getting some trouble while building it. How can I fix it?

`/usr/local/lib64/gcc/x86_64-cros-linux-gnu/8.3.0/../../../../x86_64-cros-linux-gnu/bin/ld: tools/gdm-control/tools_gdm_control_gdm_control-gdm-control.o: undefined reference to symbol 'XauReadAuth'`

Here is the rb file

```shell
require 'package'

class Openbox < Package                 # The first character of the class name must be upper case
  description ''
  homepage 'http://openbox.org/'
  version '3.4.11'
  source_url 'http://openbox.org/dist/openbox/openbox-3.4.11.tar.gz'
  source_sha256 '60fe5c1a7bace88381b851cfe51e9f726b148c91aac05163fcad1a2aaad8a051'   # Use the command "sha256sum"

  depends_on 'gcc8'            # packages required by this package
  depends_on 'libev'
  depends_on 'startup_notification'
  depends_on 'xcb_util_cursor'
  depends_on 'xcb_util_keysyms'
  depends_on 'xcb_util_wm'
  depends_on 'xcb_util_xrm'
  depends_on 'yajl'
  depends_on 'wayland' => :build
  depends_on 'sommelier'
  depends_on 'xorg_proto'
  depends_on 'glib'
  depends_on 'libxfixes'
  depends_on 'libsm'

  def self.build                   # the steps required to build the package
    system "./configure", "--sysconfdir=/usr/local/etc", "--prefix=#{CREW_PREFIX}", "--libdir=#{CREW_LIB_PREFIX}"
    system "make"
  end

  def self.install                 # the steps required to install the package
    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
    system "echo '#!/bin/sh' > startopenbox"
    system "echo 'stopsommelier' >> startopenbox"
    system "echo 'export DISPLAY=100.115.92.2:0' >> startopenbox"
    system "echo 'sudo openbox-session -u $USER  \"$@\"' >> startopenbox"
    system "chmod +x startopenbox"
  end

  def self.check                   # the steps required to check if the package was built ok
    system "make", "check"
  end
  def self.postinstall
      puts
      puts "To use this package, you need to download XServer XSDL from Google Play Store".lightblue
      puts "Use 'startopenbox' instead of 'openbox' to execute this package".lightblue
[...]

PR #4495, opened 2020-10-23
New Package: Openbox

supechicken666:
## Description

Openbox is a highly configurable, next generation window manager with extensive standards support.
This is a modified version of #4486 

## Addtional information

New features compare with #4486 
>  Updated to 3,6.1
>  Now XServer XSDL will open automatically using `sudo android-sh`
>  `openbox-session` don't work anymore so switch to `openbox` 
>  `startsommelier` only works with command line ( `whereis startsommelier` and `sudo find / abc | grep "startsommelier"` can't find it also) , so uses `initsommelier`

Works properly:
- [x] x86_64

---

## That's it

### DIFF
--- a/packages/openbox.rb
+++ b/packages/openbox.rb
@@ -1,0 +1,56 @@
+require 'package'
+
+class Openbox < Package                 # The first character of the class name must be upper case
+  description 'Openbox is a highly configurable, next generation window manager with extensive standards support.'
+  homepage 'http://openbox.org/'
+  version '3.6.1'
+  compatibility 'all'
+  source_url 'http://openbox.org/dist/openbox/openbox-3.6.1.tar.xz'
+  source_sha256 'abe75855cc5616554ffd47134ad15291fe37ebbebf1a80b69cbde9d670f0e26d'   # Use the command "sha256sum"
+
+  depends_on 'libev'
+  depends_on 'startup_notification'
+  depends_on 'xcb_util_cursor'
+  depends_on 'xcb_util_keysyms'
+  depends_on 'xcb_util_wm'
+  depends_on 'xcb_util_xrm'
+  depends_on 'yajl'
+  depends_on 'wayland' => :build
+  depends_on 'sommelier'
+  depends_on 'xorg_proto'
+  depends_on 'glib'
+  depends_on 'libxfixes'
+  depends_on 'libsm'
+  depends_on 'pango' => :build
+  depends_on 'libx11' => :build
+
+  ENV['CFLAGS'] = "-lX11 -lXau"
+  def self.build                   # the steps required to build the package
+    system "./configure #{CREW_OPTIONS}"
+    system "make"
+  end
+
+  def self.install                 # the steps required to install the package
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "echo '#!/bin/bash' > startopenbox"
+    system "echo 'stopsommelier' >> startopenbox"
+    system "echo 'echo am start x.org.server/.MainActivity | sudo android-sh' >> startopenbox"
+    system "echo 'echo Starting XServer XSDL' >> startopenbox"
+    system "echo 'sleep 10' >> startopenbox"
+    system "echo 'export DISPLAY=100.115.92.2:0' >> startopenbox"
+    system "echo 'openbox' >> startopenbox"
+    system "echo 'echo XServer closed or logged out session' >> startopenbox"
+    system "echo 'initsommelier' >> startopenbox"
+    system "chmod +x startopenbox"
+    system "cp startopenbox /usr/local/bin"
+  end
+
+  def self.check                   # the steps required to check if the package was built ok
+    system "make", "check"
+  end
+  def self.postinstall
+      puts
+      puts "To use this package, you need to download XServer XSDL from Google Play Store".lightblue
+      puts "Use 'startopenbox' instead of 'openbox' to execute this package".lightblue
+  end
+end
--- a/packages/openbox.rb
+++ b/packages/openbox.rb
@@ -16,13 +16,13 @@ class Openbox < Package                 # The first character of the class name 
   depends_on 'xcb_util_xrm'
   depends_on 'yajl'
   depends_on 'wayland' => :build
-  depends_on 'sommelier'
   depends_on 'xorg_proto'
   depends_on 'glib'
   depends_on 'libxfixes'
   depends_on 'libsm'
   depends_on 'pango' => :build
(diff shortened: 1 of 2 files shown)


packages/openbox.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  depends_on 'libev'
+  depends_on 'startup_notification'
+  depends_on 'xcb_util_cursor'
+  depends_on 'xcb_util_keysyms'
+  depends_on 'xcb_util_wm'
+  depends_on 'xcb_util_xrm'
+  depends_on 'yajl'
+  depends_on 'wayland' => :build
+  depends_on 'sommelier'
```
uberhacker, line 19: Can we make `depends_on 'sommelier'` the last dependency line?

packages/openbox.rb
```diff
@@ -0,0 +1,56 @@
...
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "echo '#!/bin/bash' > startopenbox"
+    system "echo 'stopsommelier' >> startopenbox"
+    system "echo 'echo am start x.org.server/.MainActivity | sudo android-sh' >> startopenbox"
+    system "echo 'echo Starting XServer XSDL' >> startopenbox"
+    system "echo 'sleep 10' >> startopenbox"
+    system "echo 'export DISPLAY=100.115.92.2:0' >> startopenbox"
+    system "echo 'openbox' >> startopenbox"
+    system "echo 'echo XServer closed or logged out session' >> startopenbox"
+    system "echo 'initsommelier' >> startopenbox"
```
uberhacker, line 43: `startsommelier` instead of `initsommelier`?

uberhacker: We also need `export DISPLAY=:0`.

packages/openbox.rb
```diff
@@ -0,0 +1,56 @@
...
+    system "chmod +x startopenbox"
+    system "cp startopenbox /usr/local/bin"
```
uberhacker, line 45: You could replace the 2 lines above with:
```
system "install -Dm755 startopenbox #{CREW_DEST_PREFIX}/bin/startopenbox"
```

packages/opera.rb
```diff
@@ -28,5 +28,7 @@ def self.install
     FileUtils.mv 'usr/bin', "#{CREW_DEST_PREFIX}"
     FileUtils.mv 'usr/lib', "#{CREW_DEST_PREFIX}"
     FileUtils.mv 'usr/share', "#{CREW_DEST_PREFIX}"
+    system "echo 'if [[ $(ls #{CREW_PREFIX}/bin/ | grep x-www-browser) = x-www-browser ]]; then echo x-www-browser is already set; else ln -s #{CREW_PREFIX}/bin/opera #{CREW_PREFIX}/bin/x-www-browser; fi' > script.la.tmp"
+    system "bash script.la.tmp"
   end
 end
```
uberhacker, line 34: Please remove the opera update from this PR.

uberhacker:
Thank you for the new package @supechicken666!  Please see comments.

uberhacker:
I think you removed openbox instead of opera. LoL

uberhacker:
Never mind.  I think it's correct now.

packages/opera.rb
supechicken666: Done

packages/openbox.rb
```diff
@@ -0,0 +1,56 @@
...
+  depends_on 'wayland' => :build
+  depends_on 'xorg_proto'
+  depends_on 'glib'
+  depends_on 'libxfixes'
+  depends_on 'libsm'
+  depends_on 'pango' => :build
+  depends_on 'libx11' => :build
+  depends_on 'sommelier'
+
+  ENV['CFLAGS'] = "-lX11 -lXau"
```
uberhacker, line 27: Can we add a `def self.prebuild` section to change the location of `/usr/local/x-www-browser`?

packages/openbox.rb
supechicken666: The reasons of using `startsommelier` is mentioned in Addtional information section, do you know where is the `startsommelier` command path?

packages/openbox.rb
supechicken666: This problem is only happen with ubuntu deb version, the built version now doesn't need x-www-browser symlink anymore :)

packages/openbox.rb
supechicken666: Done

packages/openbox.rb
uberhacker: Just run `crew files sommelier` and it should appear as one of the binaries.

supechicken666: > Just run `crew files sommelier` and it should appear as one of the binaries.

Sorry for late reply
Still can't found `startsommelier` with `crew files sommelier` , but I noticed a line in `sommelier` post-install
```shell
echo 'alias startsommelier="set -a && source ~/.sommelier.env && set +a && initsommelier"' >> ~/.bashrc
```
Seems the `startsommelier` command was provide by `alais`

packages/openbox.rb
supechicken666: Done

packages/openbox.rb
satmandu: startsommelier  sources ~/.sommerlier.env and note that it sets ```export DISPLAY=:0``` there. But this openbox recipe is predicated upon the use of an Android Xserver, hence the ```"echo 'export DISPLAY=100.115.92.2:0' >> startopenbox"``` - so ```startsommelier``` should re replicated in a way that avoids setting the display to ```:0``` from the sourcing of ~/.bashrc.

supechicken666: > But this openbox recipe is predicated upon the use of an Android Xserver

We need to re-run `sommelier` after the connect of Xserver connection lost or `openbox` exit. Can I just add `alias startsommelier="set -a && source ~/.sommelier.env && set +a && initsommelier"` to the script directly to set up the `startsommelier`command?

uberhacker:
Maybe it makes sense to put startsommelier in a file rather than an alias.

packages/openbox.rb
satmandu: Keep in mind that you have startsommelier in your .bashrc by default... and that will run ```set -a && source ~/.sommelier.env && set +a && initsommelier``` .  It might be easier to suggest editing the .bashrc to have startsommelier mean ```set -a && source ~/.sommelier-openbox.env && set +a && initsommelier``` and have your customized DISPLAY setting in ~/.sommelier-openbox.env  .

Issue #4496, opened 2020-10-23
ruby CREW_*_OPTIONS consts not working


satmandu:
multiline CREW OPTIONS consts are not working for meson or configure, with the exception of configure when they are given in the same line, e.g. "./confgure #{CREW_OPTIONS}" . If #{CREW_OPTIONS} goes on any other line, there is an error.

#{CREW_MESON_OPTIONS} doesn't work at all. Defaults are nice, but this breaks builds.  Also --host,  --target, and --build doesn't work with meson at all.

recent updates will need to be fixed.

Also, we need to figure out if there is a way of doing this.

Here is how homebrew is doing it:

https://github.com/Homebrew/homebrew-core/blob/master/Formula/gtk+3.rb

uberhacker:
Fixed in #4515.

PR #4497, opened 2020-10-23
gtk3: fix library location

satmandu:
Helps with https://github.com/skycocker/chromebrew/issues/4496

CREW_OPTIONS, CREW_MESON_OPTIONS do not work when there are multiple arguments given alongside these constants.

This fixes the gtk3 build 

Works properly:
- [x] x86_64

### DIFF
--- a/packages/gtk3.rb
+++ b/packages/gtk3.rb
@@ -3,7 +3,7 @@ require 'package'
 class Gtk3 < Package
   description 'GTK+ is a multi-platform toolkit for creating graphical user interfaces.'
   homepage 'https://developer.gnome.org/gtk3/3.0/'
-  version '3.24.23-1'
+  version '3.24.23-2'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


packages/gtk3.rb
```diff
@@ -30,23 +30,26 @@ def self.build
...
+         '-Ddemos=false',
+         '-Dexamples=false',
+         "-Dprefix=#{CREW_PREFIX}",
+         "-Dlibdir=#{CREW_LIB_PREFIX}",
+         "-DLIB_INSTALL_DIR=#{CREW_LIB_PREFIX}",
+         "-Dmandir=#{CREW_MAN_PREFIX}",
+         "-DSYSCONFDIR=#{CREW_PREFIX}/etc",
+         "-Ddatadir=#{CREW_LIB_PREFIX}",
+         '-Dbuildtype=release',
+         'build'
```
uberhacker, line 46: Try the following:
```
system "meson #{CREW_MESON_OPTIONS} -Dbroadway_backend=true -Dgtk_doc=false -Ddemos=false -Dexamples=false build"
```

PR #4501, opened 2020-10-23
librsvg: Fix CREW_OPTIONS breakage

satmandu:
For https://github.com/skycocker/chromebrew/issues/4496

Works properly:
- [ ] x86_64 compiling...

### DIFF
--- a/packages/librsvg.rb
+++ b/packages/librsvg.rb
@@ -3,7 +3,7 @@ require 'package'
 class Librsvg < Package
   description 'SVG library for GNOME'
   homepage 'https://wiki.gnome.org/Projects/LibRsvg'
-  version '2.50.1'
+  version '2.50.1-1'
   compatibility 'all'
   source_url 'https://download.gnome.org/sources/librsvg/2.50/librsvg-2.50.1.tar.xz'
   source_sha256 '9bd9799322e06cf5db19b9f7afb728edac6efcf0110baafc44f0f96f45df9a09'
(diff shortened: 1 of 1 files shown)


packages/librsvg.rb
```diff
@@ -25,7 +25,12 @@ def self.build
...
-      #{CREW_OPTIONS},
+      "--prefix=#{CREW_PREFIX}",
+      "--libdir=#{CREW_LIB_PREFIX}",
+      "--mandir=#{CREW_MAN_PREFIX}",
+      "--build=#{CREW_BUILD}",
+      "--host=#{CREW_BUILD}",
+      "--target=#{CREW_BUILD}",
       "--enable-introspection=yes",
       "--enable-vala=yes",
       "--disable-static",
```
uberhacker, line 36: Try this instead:
```
system "./configure #{CREW_OPTIONS} --enable-introspection=yes --enable-vala=yes --disable-static"
```

uberhacker: The problem is with the way ruby parses strings separated by commas.  `CREW_OPTIONS` is space delimited.

satmandu: Right, but if you have CREW_OPTIONS  on a different line, then the other lines are ignored, and so is CREW/MESON_OPTIONS.

So all following options have to be space delimited on that line?

uberhacker: Yes, unfortunately the ruby parser gets confused when it builds the command if you mix and match space and comma delimited arguments.

PR #4511, opened 2020-10-25
graphite: freetype isn't needed after building

satmandu:
Trying to eliminate a circular dependency in building harfbuzz.

This shouldn't require a rebuild.

Works properly:
- [x] x86_64


uberhacker:
Probably not worth rebuilding binaries.  We can fix this on the next release.

satmandu:
What do you think of just updating the dependencies in the file to freetype_sub?

The issue isn't with building this, but with building harfbuzz, which needs graphite to build, which is a problem since freetype also needs harfbuzz to build.

PR #4513, opened 2020-10-25
libxinerama: Update depends

satmandu:
Does not need fontconfig to compile, but does need libx11 & libxext.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/libxinerama.rb
+++ b/packages/libxinerama.rb
@@ -1,26 +1,12 @@
-
 require 'package'
 
 class Libxinerama < Package
   description 'Xorg library, Xinerama is an X11 extension which provides support for extending a desktop across multiple displays.'
   homepage 'https://www.x.org/'
   compatibility 'all'
-  version '1.1.3-1'
-  source_url 'https://www.x.org/archive//individual/lib/libXinerama-1.1.3.tar.gz'
-  source_sha256 '0ba243222ae5aba4c6a3d7a394c32c8b69220a6872dbb00b7abae8753aca9a44'
-
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/libxinerama-1.1.3-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/libxinerama.rb
```diff
@@ -22,8 +22,9 @@ class Libxinerama < Package
      x86_64: '73f25bf7cf51787d00e2fbe46cf6589e321c465333e8d2c9dc3d97bfa93a70da',
   })
 
-  depends_on 'fontconfig'
+  depends_on 'libx11'
   depends_on 'libxcb'
+  depends_on 'libxext'
 
   def self.build
     system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
```
uberhacker, line 30: Use `#{CREW_OPTIONS}`?

packages/libxinerama.rb
```diff
@@ -22,8 +22,9 @@ class Libxinerama < Package
      x86_64: '73f25bf7cf51787d00e2fbe46cf6589e321c465333e8d2c9dc3d97bfa93a70da',
   })
 
-  depends_on 'fontconfig'
+  depends_on 'libx11'
   depends_on 'libxcb'
+  depends_on 'libxext'
```
uberhacker, line 27: Does this affect the binaries?  If not then we don't need to bump the version.

satmandu: I don't think this affects the binaries at all. This is a build issue. Those were already being pulled in from xorg_libs in any case, so the binaries would already know about them.

packages/libxinerama.rb
satmandu: done

packages/libxinerama.rb
uberhacker: Might as well update to https://www.x.org/archive/individual/lib/libXinerama-1.1.4.tar.bz2.

satmandu: Done.

PR #4514, opened 2020-10-25
libdrm & mesa: xorg_libs not needed for compile

satmandu:
libdrm uses xorg_libs, which is a hack creating circular dependencies, especially since libdrm doesn't need any of those to compile.

This moves dependencies back out to individual app dependencies.

Mesa also has dependencies added which it needs to compile which are otherwise in xorg_libs.

Works properly (both libdrm & mesa are compiling):
- [x] x86_64

There is likely breakage in downstream (of libdrm) apps, but those should be noticed when there is a version bump and/or someone decides to compile from source.

### DIFF
--- a/packages/libdrm.rb
+++ b/packages/libdrm.rb
@@ -22,7 +22,6 @@ class Libdrm < Package
   })
 
   depends_on 'libpciaccess'
-  depends_on 'xorg_lib'
   depends_on 'eudev'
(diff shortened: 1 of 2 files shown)


uberhacker:
There are actually a slew of problems with mesa that I will address in a pre-built binaries PR coming soon.

PR #4516, opened 2020-10-25
harfbuzz, cairo: Fix dependencies

satmandu:
Harfbuzz needs gobjects_introspection, which has an unnecessary cairo dependency.

This fix to both avoids pulling in full freetype as a circular dependency.

Works properly:
- [x] x86_64


uberhacker:
I have a fix for all of the issues in the queue coming shortly in the pre-built binaries PR.

satmandu:
Ah ok I'll wait for that.

PR #4520, opened 2020-10-25
Add `adwaita_icon_theme` dependency to openbox.rb

supechicken666:
This is an update of #4495 
Add `adwaita_icon_theme` as dependency to provide icon support
Use `startsommelier` instead of `initsommelier` in script


satmandu:
@supechicken666 Does my update of sommelier https://github.com/skycocker/chromebrew/pull/4450 work for you with openbox?

packages/openbox.rb
```diff
@@ -0,0 +1,57 @@
...
+    system "make"
+  end
+
+  def self.install                 # the steps required to install the package
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "echo '#!/bin/bash' > startopenbox"
+    system "echo 'stopsommelier' >> startopenbox"
+    system "echo 'echo am start x.org.server/.MainActivity | sudo android-sh' >> startopenbox"
+    system "echo 'echo Starting XServer XSDL' >> startopenbox"
+    system "echo 'sleep 10' >> startopenbox"
```
uberhacker, line 40: 10 seconds seems like too long to wait.

packages/openbox.rb
```diff
@@ -0,0 +1,57 @@
...
+    system "echo 'export DISPLAY=100.115.92.2:0' >> startopenbox"
+    system "echo 'openbox' >> startopenbox"
```
uberhacker, line 42: Should we allow for parameters like below?
```
system "echo 'openbox \"$@\"' >> startopenbox"
```

supechicken666: How about
```shell
system "echo 'openbox \"$@\ & sh"' >> startopenbox"
```
This will open a shell along with openbox
But how to exit `sh` after `openbox` close?

supechicken666:
> @supechicken666 Does my update of sommelier #4450 work for you with openbox?

The `startsommelier` command works well:) But the `adwaita_icon_theme` build fail on my laptop #4527

packages/openbox.rb
supechicken666: It takes 7 seconds on my phone (is my phone too old lol)
I think 7 seconds is enough for most chromebook

satmandu: You could also poll for xserver availability after exporting the display and before running openbox in the script, using xdpyinfo like here:  https://unix.stackexchange.com/a/418903

chromebrew doesn't have xdpyinfo, but importing a recipe for that from arch took about 5 min:

```
# Adapted from Arch Linux xorg-xdpyinfo PKGBUILD at:
# https://github.com/archlinux/svntogit-packages/raw/packages/xorg-xdpyinfo/trunk/PKGBUILD

require 'package'

class Xorg_xdpyinfo < Package
  description 'Display information utility for X'
  version '1.3.2-4'
  compatibility 'all'
  source_url 'https://xorg.freedesktop.org/archive/individual/app/xdpyinfo-1.3.2.tar.bz2'
  source_sha256 '30238ed915619e06ceb41721e5f747d67320555cc38d459e954839c189ccaf51'

  depends_on 'libx11'
  depends_on 'libxext'
  depends_on 'libxtst'
  depends_on 'libxxf86vm'
  depends_on 'libxrender'
  depends_on 'libxcomposite'
  depends_on 'libxinerama'
  depends_on 'xorg_proto' 

def self.build
system "./configure #{CREW_OPTIONS}"
system 'make'
end
def self.install
end
end
```

packages/openbox.rb
uberhacker: You can add the `& sh` as parameters in `$@`, right?

packages/openbox.rb
uberhacker: Should we add this as a new package @satmandu?

satmandu: Yes, go ahead.

Just for completeness' sake, here is roughly what I would assume would be the wait code for starting openbox:

```
system "echo 'echo Starting XServer XSDL' >> startopenbox"
system "export DISPLAY=100.115.92.2:0' >> startopenbox
system "cat <EOF>> startopenbox
# Wait for XSDL X server
MAX_ATTEMPTS=120 # About 60 seconds
COUNT=0
echo -n \"Waiting for X server to be ready...\"
while ! xdpyinfo -display ${DISPLAY} >/dev/null 2>&1; do
  echo -n \".\"
  sleep 0.50s
  COUNT=$(( COUNT + 1 ))
  if [ \"${COUNT}\" -ge \"${MAX_ATTEMPTS}\" ]; then
    echo \"  Gave up waiting for X server on ${DISPLAY}\"
    exit 1
  fi
done
echo \"  Done - X server is ready!\"
EOF"
echo 'openbox' >> startopenbox
```

(Obviously include xorg_xdpyinfo in your depends if this is the way you want to go.)

supechicken666:
Just add chromium os and i686 support
Tested platform:

- [ ] i686

- [x] Cloudready

- [x] Chrome OS with android support

packages/openbox.rb
```diff
@@ -33,25 +34,87 @@ def self.build                   # the steps required to build the package
...
     system "echo 'export DISPLAY=100.115.92.2:0' >> startopenbox"
-    system "echo 'openbox' >> startopenbox"
-    system "echo 'echo XServer closed or logged out session' >> startopenbox"
+    system "echo 'openbox \"$@\"' >> startopenbox"
     system "echo 'startsommelier' >> startopenbox"
-    system "echo 'export DISPLAY=:0' >> startopenbox"
-    system "install -Dm755 startopenbox #{CREW_DEST_PREFIX}/bin/startopenbox"
+    # For Chrome OS/Chromium OS/Cloudready without android arc container support
+    # Use 'xinit'
+    system "if [[ cat /etc/lsb-release | grep ARC != "" ]]; then echo 'Your OS have Android Support :)'; else if [[ whereis xinit | grep /usr/local/bin/xinit != /usr/local/bin/xinit ]]; then echo 'To use this package, You need to install `xinit` package'; else echo 'pkill twm && openbox &' >> /usr/local/etc/X11/xinit/xinitrc; fi"
```
uberhacker: Can this line be split into separate lines for clarity?  Also, `ARC != ""` should be `ARCH != \"\"`?  Finally, please abstract the directories and change `/usr/local` to `#{CREW_PREFIX}`.

packages/openbox.rb
```diff
@@ -33,25 +34,87 @@ def self.build                   # the steps required to build the package
...
+    #  fi
+    #done
+    #echo \"  Done - X server is ready!\"
+    #xclock &
+    #xterm &
+    #openbox &
+    #EOF"
+    # system "unzip *.zip && mv */ ~/Downloads"
+    # puts "Install the two unpack extension to chrome with entension developer mode on".lightblue  
+    # system "if [[ $(uname -m) = i686 ]]; then 'install -Dm755 startopenbox_arc-extension #{CREW_DEST_PREFIX}/bin/startopenbox'; else 'install -Dm755 startopenbox #{CREW_DEST_PREFIX}/bin/startopenbox'; fi"
```
uberhacker: No need to include all the commented lines.  Please remove.

uberhacker:
@supechicken666: This is great you are doing all this extra testing on different architectures!  Please see the comments.

packages/openbox.rb
supechicken666: > Yes, go ahead.
> 
> Just for completeness' sake, here is roughly what I would assume would be the wait code for starting openbox:
> 
> ```
> system "echo 'echo Starting XServer XSDL' >> startopenbox"
> system "export DISPLAY=100.115.92.2:0' >> startopenbox
> system "cat <EOF>> startopenbox
> # Wait for XSDL X server
> MAX_ATTEMPTS=120 # About 60 seconds
> COUNT=0
> echo -n \"Waiting for X server to be ready...\"
> while ! xdpyinfo -display ${DISPLAY} >/dev/null 2>&1; do
>   echo -n \".\"
>   sleep 0.50s
>   COUNT=$(( COUNT + 1 ))
>   if [ \"${COUNT}\" -ge \"${MAX_ATTEMPTS}\" ]; then
>     echo \"  Gave up waiting for X server on ${DISPLAY}\"
>     exit 1
>   fi
> done
> echo \"  Done - X server is ready!\"
> EOF"
> echo 'openbox' >> startopenbox
> ```
> 
> (Obviously include xorg_xdpyinfo in your depends if this is the way you want to go.)

`<<EOF>` ?

uberhacker: Sure, that may make it "easier".

satmandu:
Quick question. Any reason you just aren't installing ```xinit``` and then editing ```/usr/local/etc/X11/xinit/xinitrc``` to have ```openbox-session``` in it?

Then you can just run ```startx``` to get a window with openbox in it. (You don't even have to shut down sommelier.)

There's no need to run an xserver from Android... or am I missing something?

supechicken666:
> Quick question. Any reason you just aren't installing `xinit` and then editing `/usr/local/etc/X11/xinit/xinitrc` to have `openbox-session` in it?
> 
> Then you can just run `startx` to get a window with openbox in it. (You don't even have to shut down sommelier.)
> 
> There's no need to run an xserver from Android... or am I missing something?
Sorry for my bad English

Since it is based on `sommelier` , when the screen is not set to the original resolution of the screen in settings (like the default resolution) the screen will be cutted (The maximize button is located out of the screen) and blurred (This is also a bug on `sommelier` )

And the `xinit` package need a lot extra dependencies to work that `sommelier` not required. The XSDL is a more lightweight solution (it doesn't need any dependency as well)

packages/openbox.rb
supechicken666: the `ARC` is used to detect does the system have android container support, maybe change to `grep CHROMEOS_ARC_ANDROID_SDK_VERSION` will be more clearly

supechicken666:
I am facing some problem with the `startopenbox` script now, we need variables in the script like `${COUNT}` to count. When putting the `${COUNT}` to the startup script, `sh` will recognize and replace it with the current environment variable even with `cat` (it will replace by blank as the $COUNT doesn't set)
```shell
system "cat <<EOF> test
COUNT=0
${COUNT}
EOF"
```
How can I bypass the detection?

satmandu:
Why not just replace ```#!/bin/sh``` with ```#!/bin/bash```?

satmandu:
Try this:
```
system "cat <<'EOF'> test
COUNT=0
${COUNT}
EOF"
```

supechicken666:
> Try this:
> 
> ```
> system "cat <<'EOF'> test
> COUNT=0
> ${COUNT}
> EOF"
> ```

It works. Thanks a lot!

packages/openbox.rb
supechicken666: Done

supechicken666:
Tested on
- [x]  Chrome OS
- [x]  Cloudready

uberhacker:
@supechicken: Should we put `[WIP]` in the title or is this ready to merge?

packages/openbox.rb
```diff
@@ -0,0 +1,83 @@
...
+  ENV['CFLAGS'] = "-lX11 -lXau"
+  def self.build
+    system "./configure #{CREW_OPTIONS}"
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system "echo '#!/bin/bash' > startopenbox"
+    system "cat <<'EOF'> startopenbox
```
uberhacker, line 38: You could move the `#!/bin/bash` line here instead.

supechicken666: Done

supechicken666:
This PR is ready to merge :)

PR #4521, opened 2020-10-25
Mesa: Use CREW_MESON_OPTIONS & add direct dependencies

satmandu:
WIth the fixed CREW_MESON_OPTIONS this appears to be compiling properly on:
- [x] x86_64
- [x] i686

(Also added missing dependencies I found while compiling).


satmandu:
(no need to bump version as this is just useful for people compiling.)

PR #4528, opened 2020-10-26
Update stressng from 0.11.18 to 0.11.22

cstrouse:
Tested on x86_64


uberhacker:

PR #4529, opened 2020-10-27
libunwind: update to 1.50rc2 (fixes builds with gcc10)

satmandu:
Realized some build errors on packages I am working on were from libunwind incompatibility. Discovered I couldn't rebuild it without [patches](https://github.com/libunwind/libunwind/commit/29e17d8d2ccbca07c423e3089a6d5ae8a1c9cb6e) that went into the 1.5 series.

(I have intentionally left in the ```binary_url``` section since I don't know how this compiles on gcc8.)

Compiles properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/libunwind.rb
+++ b/packages/libunwind.rb
@@ -3,10 +3,10 @@ require 'package'
 class Libunwind < Package
   description 'libunwind is a portable and efficient C programming interface (API) to determine the call-chain of a program.'
   homepage 'https://www.nongnu.org/libunwind/'
-  version '1.3.1'
+  version '1.5.0-rc2'
   compatibility 'all'
-  source_url 'https://download.savannah.gnu.org/releases/libunwind/libunwind-1.3.1.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/libunwind.rb
```diff
@@ -3,10 +3,10 @@
 class Libunwind < Package
   description 'libunwind is a portable and efficient C programming interface (API) to determine the call-chain of a program.'
   homepage 'https://www.nongnu.org/libunwind/'
-  version '1.3.1'
+  version '1.5.0-rc2'
```
uberhacker, line 6: 1.5-rc2?

satmandu:
Nuts, I know.

Hence my suggestion for keeping the older binary section.

I'm going to try to check the the devs and ask about them tagging a 1.5...

packages/libunwind.rb
satmandu: There has to be a better way...

PR #4530, opened 2020-10-27
Add peg

satmandu:
Needed for Igt_gpu_tools PR.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/peg.rb
+++ b/packages/peg.rb
@@ -10,10 +10,9 @@ class Peg < Package
   source_sha256 '20193bdd673fc7487a38937e297fff08aa73751b633a086ac28c3b34890f9084'
 
   def self.build
-    system "sed -i 's/PREFIX  = \\/usr\\/local/PREFIX =/g' Makefile"
     system "PREFIX=#{CREW_PREFIX} make"
   end
   def self.install
-    system "make ROOT=#{CREW_DEST_DIR} install"
+    system "PREFIX=#{CREW_PREFIX} ROOT=#{CREW_DEST_DIR} make install"
   end
 end
--- a/packages/peg.rb
+++ b/packages/peg.rb
@@ -10,7 +10,8 @@ class Peg < Package
   source_sha256 '20193bdd673fc7487a38937e297fff08aa73751b633a086ac28c3b34890f9084'
 
   def self.build
-    system "make"
+    system "sed -i 's/PREFIX  = \\/usr\\/local/PREFIX =/g' Makefile"
+    system "PREFIX=#{CREW_PREFIX} make"
   end
   def self.install
     system "make ROOT=#{CREW_DEST_DIR} install"
--- a/packages/peg.rb
+++ b/packages/peg.rb
@@ -10,9 +10,10 @@ class Peg < Package
   source_sha256 '20193bdd673fc7487a38937e297fff08aa73751b633a086ac28c3b34890f9084'
 
   def self.build
-    system "PREFIX=#{CREW_PREFIX} make"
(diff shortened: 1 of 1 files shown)


packages/peg.rb
```diff
@@ -0,0 +1,18 @@
...
+require 'package'
+
+class Peg < Package
+  description 'recursive-descent parser generators for C'
+  version '0.1.18-2'
+  source_url 'http://piumarta.com/software/peg/peg-0.1.18.tar.gz'
+  source_sha256 '20193bdd673fc7487a38937e297fff08aa73751b633a086ac28c3b34890f9084'
+
+  def self.build
+    system "make"
```
uberhacker, line 13: Where is `CREW_PREFIX`?  If someone decides to change the prefix, how would they do it?

satmandu:
There's no configure script. I should hack something like that in with sed.

packages/peg.rb
satmandu: Used sed to remove PREFIX and thus force use of PREFIX during make.

packages/peg.rb
```diff
@@ -0,0 +1,19 @@
...
+require 'package'
+
+class Peg < Package
+  description 'recursive-descent parser generators for C'
+  version '0.1.18-2'
+  source_url 'http://piumarta.com/software/peg/peg-0.1.18.tar.gz'
+  source_sha256 '20193bdd673fc7487a38937e297fff08aa73751b633a086ac28c3b34890f9084'
+
+  def self.build
+    system "sed -i 's/PREFIX  = \\/usr\\/local/PREFIX =/g' Makefile"
```
uberhacker, line 13: `system "sed -i 's,PREFIX  = /usr/local,PREFIX = #{CREW_PREFIX},g' Makefile"`?

packages/peg.rb
```diff
@@ -0,0 +1,19 @@
...
+    system "PREFIX=#{CREW_PREFIX} make"
+  end
+  def self.install
+    system "make ROOT=#{CREW_DEST_DIR} install"
```
uberhacker, line 17: You could test with a different `PREFIX` than `/usr/local` but you might need `PREFIX=#{CREW_PREFIX}` here also.

packages/peg.rb
```diff
@@ -0,0 +1,19 @@
...
+
+class Peg < Package
+  description 'recursive-descent parser generators for C'
+  version '0.1.18-2'
+  source_url 'http://piumarta.com/software/peg/peg-0.1.18.tar.gz'
+  source_sha256 '20193bdd673fc7487a38937e297fff08aa73751b633a086ac28c3b34890f9084'
+
+  def self.build
+    system "sed -i 's/PREFIX  = \\/usr\\/local/PREFIX =/g' Makefile"
+    system "PREFIX=#{CREW_PREFIX} make"
```
uberhacker, line 14: Actually, you might not need the sed command.  Try with a different `PREFIX` here and see what happens.

satmandu: I was being overly anal about this. I'll delete the sed command entirely. It works fine with just the prefix before make.

packages/peg.rb
satmandu: Good point there.

packages/peg.rb
satmandu: this is perfect

satmandu:
Turns out that without the sed on the makefile, it complains:

```
mv leg-new leg
Preconfiguring package...
cp -p peg /usr/local/bin/peg
strip /usr/local/bin/peg
cp -p leg /usr/local/bin/leg
strip /usr/local/bin/leg
Performing pre-install...
Performing install...
tar: ./usr/*: Cannot stat: No such file or directory
Performing post-install...
Peg installed!
```

PR #4534, opened 2020-10-27
add xdpyinfo

satmandu:
Good for checking to see if x server is working.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/xdpyinfo.rb
+++ b/packages/xdpyinfo.rb
@@ -1,0 +1,30 @@
+# Adapted from Arch Linux xorg-xdpyinfo PKGBUILD at:
+# https://github.com/archlinux/svntogit-packages/raw/packages/xorg-xdpyinfo/trunk/PKGBUILD
+
+require 'package'
+
+class Xorg_xdpyinfo < Package
+  description 'Display information utility for X'
+  version '1.3.2-4'
+  compatibility 'all'
+  source_url 'https://xorg.freedesktop.org/archive/individual/app/xdpyinfo-1.3.2.tar.bz2'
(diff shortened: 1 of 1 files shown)


packages/xdpyinfo.rb
```diff
@@ -0,0 +1,30 @@
...
+  depends_on 'xorg_proto' 
+
+        def self.build
+                system "./configure #{CREW_OPTIONS}"
+                system 'make'
+        end
+        
+        def self.install
+                system "make DESTDIR=#{CREW_DEST_DIR} install"
+        end
```
uberhacker, line 29: Why so much indentation?

uberhacker:

packages/xdpyinfo.rb
```diff
@@ -0,0 +1,30 @@
+# Adapted from Arch Linux xorg-xdpyinfo PKGBUILD at:
+# https://github.com/archlinux/svntogit-packages/raw/packages/xorg-xdpyinfo/trunk/PKGBUILD
+
+require 'package'
+
+class Xorg_xdpyinfo < Package
+  description 'Display information utility for X'
+  version '1.3.2-4'
```
uberhacker, line 8: Why do we need `-4`?

satmandu: Sorry that was a straight import from the arch pkgbuild, and I hadn't reset the -4.

packages/xdpyinfo.rb
satmandu: my geany indents are getting reset... fixing that.

satmandu:
fixed!

PR #4535, opened 2020-10-27
libunwind: Fix GCC10 compilation & update to 1.3.2

satmandu:
This fixes compilation with gcc-10. Also compiles with gcc-8.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/libunwind.rb
+++ b/packages/libunwind.rb
@@ -3,31 +3,18 @@ require 'package'
 class Libunwind < Package
   description 'libunwind is a portable and efficient C programming interface (API) to determine the call-chain of a program.'
   homepage 'https://www.nongnu.org/libunwind/'
-  version '1.3.1'
+  version '1.3.2'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


packages/libunwind.rb
```diff
@@ -3,31 +3,18 @@
...
-     x86_64: '5b7de2da3534ee138cd258ed3202eca570cc77ef584353c011c6111fd6b0b024',
-  })
+  source_url 'https://download.savannah.gnu.org/releases/libunwind/libunwind-1.3.2.tar.gz'
+  source_sha256 '0a4b5a78d8c0418dfa610245f75fa03ad45d8e5e4cc091915d2dbed34c01178e'
 
   depends_on 'xzutils' => :build
 
+  # Fixes GCC 10 compilation.
+  # As per https://github.com/libunwind/libunwind/pull/166
+  ENV['CFLAGS'] = '-fcommon'
```
uberhacker, line 15: Does this break compilation with gcc8?

satmandu: Tested it with gcc-8 on x86_64, and it compile just fine.

PR #4536, opened 2020-10-27
xdpyinfo: fix package name detritus

satmandu:
That was sloppy. :/

Works properly:
- [x] x86_64


uberhacker:
Good catch.  I should have caught that in the review.

PR #4538, opened 2020-10-29
crew: Run ldconfig  on install; Adjust where reinstall removes package to be reinstalled.

satmandu:
Run ldconfig after package installs so apps don't complain about missing just installed libraries. 
(This happens with curl, and I have also seen it with python3 installs)

Also, do reinstalls as part of the upgrade path so that package remove happens after the package to be reinstalled has been downloaded. This fixes upgrades on curl (or its dependencies) erroring out on being unable to download the new package using curl.

Works properly:
- [x] x86_64


uberhacker:
So if I'm understanding this right you want to rework the reinstall command to work more like the upgrade command?  We'll need the crew version bumped in const.rb.

satmandu:
Exactly that. Version bumped.

uberhacker:
While we're at it, I noticed the new constants are missing after running `crew const`.

satmandu:
What constants are missing? This is reusing the ```pkg.in_upgrade``` from package.rb.

uberhacker:
`CREW_CMAKE_OPTIONS`, `CREW_CMAKE_LIBSUFFIX_OPTIONS` and `CREW_MESON_OPTIONS`.

PR #4546, opened 2020-10-30
WIP: Custom #{CREW_PREFIX} updates

satmandu:
* crew should created needed dirs in CREW_PREFIX
Otherwise one gets this error during install:
 ```ln: target '/usr/local/$CREW_PREFIX/bin/' is not a directory: No such file or directory```

* crew should run its own apps from $CREW_PREFIX/bin/ for bootstrapping

```

### CODE REVIEW

File: install.sh [hunk 1]
```diff
@@ -36,7 +36,8 @@ esac
 sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}"
 
 # prepare directories
-for dir in "${CREW_CONFIG_PATH}/meta" "${CREW_DEST_DIR}" "${CREW_PACKAGES_PATH}"; do
+DIRLIST="${CREW_CONFIG_PATH}/meta ${CREW_DEST_DIR} ${CREW_PACKAGES_PATH} ${CREW_PREFIX}/bin ${CREW_PREFIX}/lib ${CREW_PREFIX}/lib${LIB_SUFFIX} ${CREW_PREFIX}/libexec ${CREW_PREFIX}/man ${CREW_PREFIX}/share ${CREW_PREFIX}/tmp ${CREW_PREFIX}/var"
+for dir in ${DIRLIST}; do
```
uberhacker, line 40: Do we need to do this?  I haven't noticed any issues with the install.sh script.

install.sh
```diff
@@ -314,18 +315,18 @@ ln -s "../lib/crew/crew" "${CREW_PREFIX}/bin/"
 # prepare sparse checkout .rb packages directory and do it
 cd "${CREW_LIB_PATH}"
 rm -rf .git
-git init
-git remote add -f origin "https://github.com/${OWNER}/${REPO}.git"
-git config core.sparsecheckout true
+$CREW_PREFIX/bin/git init
+$CREW_PREFIX/bin/git remote add -f origin "https://github.com/${OWNER}/${REPO}.git"
+$CREW_PREFIX/bin/git config core.sparsecheckout true
```
uberhacker, line 320: There is only one git so no need to use the full path to the binary.

install.sh
```diff
@@ -314,18 +315,18 @@ ln -s "../lib/crew/crew" "${CREW_PREFIX}/bin/"
...
-git fetch origin "${BRANCH}"
-git reset --hard origin/"${BRANCH}"
-crew update
+$CREW_PREFIX/bin/git fetch origin "${BRANCH}"
+$CREW_PREFIX/bin/git reset --hard origin/"${BRANCH}"
+$CREW_PREFIX/bin/crew update
 
 # install a base set of essential packages
-yes | crew install buildessential less most
+yes | $CREW_PREFIX/bin/crew install buildessential less most
```
uberhacker, line 329: Is the idea here to make sure `crew` works if `CREW_PREFIX` changes?  In any event, how would you handle spaces in directory names?  Not that it should happen but I've seen stranger things in my time.

satmandu: Definitely an issue, but I've just changed prefixes and am noticing breakage, so adding fixes as I go.

install.sh
satmandu: the new script complains if the new prefix doesn't have files, but that could be because of breakage elsewhere.

install.sh
satmandu: I was wondering why this broke the script, and it turns out that the extract function doesn't extract to the new CREW_PREFIX. See the last commit.

satmandu:
@uberhacker How do you feel about changing the extract_install function to this:
```
function extract_install () {
    # Start with a clean slate
    rm -rf "${CREW_DEST_DIR}"
    mkdir "${CREW_DEST_DIR}"
    cd "${CREW_DEST_DIR}"

    #extract and install
    echo "Extracting & Installing ${1} (this may take a while)..."
    if tar --help | grep -q strip-components ; then
      tar xpf ../"${2}" --keep-directory-symlink --strip-components=2 -C "${CREW_PREFIX}/"
      tar xpf ../"${2}" dlist
      tar xpf ../"${2}" filelist
    else
      tar xpf ../"${2}"
      tar cpf - ./*/* | (cd /; tar xp --keep-directory-symlink -f -)
    fi
    mv ./dlist "${CREW_CONFIG_PATH}/meta/${1}.directorylist"
    mv ./filelist "${CREW_CONFIG_PATH}/meta/${1}.filelist"
}
```

The else portion is what exists now, and it is broken currently because it just doesn't take CREW_PREFIX into account at all when extracting.

satmandu:
Also using an alternate CREW_PREFIX breaks install.sh right after the git stuff:

```
 * [new tag]           v0.2.1                           -> v0.2.1
 * [new tag]           v0.2.1.1                         -> v0.2.1.1
 * [new tag]           v0.2.1.2                         -> v0.2.1.2
 * [new tag]           v0.3                             -> v0.3
 * [new tag]           v0.3.1                           -> v0.3.1
 * [new tag]           v0.4                             -> v0.4
HEAD is now at c6af2a3c Add xdotool (#4540)
        2: from /usr/local/crew-container/bin/crew:13:in `<main>'
        1: from /usr/local/crew-container/bin/crew:13:in `require_relative'
/usr/local/crew-container/lib/crew/lib/const.rb:17:in `<top (required)>': undefined method `build_from_source=' for nil:NilClass (NoMethodError)
```

This is due to this in const.rb:
```
if ENV['CREW_PREFIX'].to_s == ''
  CREW_PREFIX = '/usr/local'
else
  @pkg.build_from_source = true
end
```

That doesn't work.

satmandu:
(Taking it now doesn't help either.)

```
yes | CREW_PREFIX=/usr/local/crew-container /usr/local/crew-container/bin/crew install buildessential less most
        12: from /usr/local/crew-container/bin/crew:1089:in `<main>'
        11: from /usr/local/crew-container/bin/crew:1003:in `install_command'
        10: from /usr/local/crew-container/bin/crew:1003:in `each'
         9: from /usr/local/crew-container/bin/crew:1005:in `block in install_command'
         8: from /usr/local/crew-container/bin/crew:198:in `search'
         7: from /usr/local/lib64/ruby/2.7.0/find.rb:43:in `find'
         6: from /usr/local/lib64/ruby/2.7.0/find.rb:43:in `each'
         5: from /usr/local/lib64/ruby/2.7.0/find.rb:48:in `block in find'
         4: from /usr/local/lib64/ruby/2.7.0/find.rb:48:in `catch'
         3: from /usr/local/lib64/ruby/2.7.0/find.rb:49:in `block (2 levels) in find'
         2: from /usr/local/crew-container/bin/crew:199:in `block in search'
         1: from /usr/local/crew-container/bin/crew:121:in `set_package'
/usr/local/crew-container/bin/crew:98:in `print_package': undefined method `any?' for nil:NilClass (NoMethodError)
```

It works if I do this: ```
yes |  CREW_PREFIX=/usr/local/crew-container  LD_LIBRARY_PATH=/usr/local/crew-container/lib64 /usr/local/crew-container/bin/crew install buildessential less most
buildessential: A collection of tools essential to compile and build software.

version 1.7
binutils filecmd texinfo perl perl_locale_messages perl_text_unidecode perl_unicode_eastasianwidth ncurses gmp mpfr mpc isl glibc lz4 gcc_tools linuxheaders make pkgconfig manpages gdbm readline groff uchardet libpipeline compressdoc doxygen sed acl attr gettext libxml2 zlibpkg icu4c automake autoconf m4 libsigsegv libtool intltool perl_xml_parser expat patch diffutils bison flex util_macros wget gnutls nettle openssl libtasn1 trousers p11kit libffi libunbound gawk lzip unzip zip cmake meson ninja setuptools python27 python3 bz2 xzutils sqlite libedit krb5 mandb 
Do you agree? [Y/n] Proceeding...
```

File: install.sh [hunk 4]
```diff
@@ -4,9 +4,9 @@
 set -e
 
 #chromebrew directories
-OWNER="skycocker"
-REPO="chromebrew"
-BRANCH="master"
+OWNER="${CREW_OWNER:-skycocker}"
+REPO="${CREW_REPO:-chromebrew}"
+BRANCH="${CREW_BRANCH:-master}"
```
uberhacker, line 9: Should we add these new constants to the list?

lib/const.rb
```diff
@@ -14,9 +14,13 @@
...
 else
   CREW_PREFIX = ENV['CREW_PREFIX']
-  @pkg.build_from_source = true
+  ENV['PATH'] = "#{CREW_PREFIX}/bin:/usr/bin:/bin:/opt/bin"
+  if @pkgName
+    @pkg.build_from_source = true
+  end
 end
 CREW_LIB_PREFIX = CREW_PREFIX + '/' + ARCH_LIB
+ENV['LD_LIBRARY_PATH'] = "#{CREW_LIB_PREFIX}"
```
uberhacker, line 23: This works too.
```
ENV['LD_LIBRARY_PATH'] = CREW_LIB_PREFIX
```

install.sh
satmandu: This would be good for testing, but crew update just ends up using https://github.com/skycocker/chromebrew/blob/9e4afa7874826f3942438cd8d7d1c9a781c2b298/crew#L391
```
    system "git fetch origin master"
    system "git reset --hard origin/master"
```
so we should probably have crew save the owner/rep/branch info in prefs too.

satmandu:
@uberhacker I'm inclined to abandon this PR. There are too many edge conditions to make it worth actually supporting non-/usr/local CREW_PREFIX use.

uberhacker:
@satmandu: I understand.  This is not a nimble task for sure.

PR #4562, opened 2020-11-07
Update Opera to 72.0.3815.207, use `curl` for downloading

supechicken666:
Update opera.rb frome 72.0.3815.148 to 72.0.3815.207
Fetch sha256 checksum automatically
Hide `tar` output

### DIFF
--- a/packages/opera.rb
+++ b/packages/opera.rb
@@ -21,7 +21,7 @@ class Opera < Package
     abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("opera.deb") ) == `curl -Ls https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb.sha256sum`
     puts "Archive downloaded".lightgreen
     system "alien -tc opera.deb"
-    system "tar xvf opera-stable-#{version}.tgz"
+    system "tar xvf opera-stable-#{version}.tgz > /dev/null"
   end
 
   def self.install
--- a/packages/opera.rb
+++ b/packages/opera.rb
@@ -17,9 +17,10 @@ class Opera < Package
 
 
   def self.build
-    system "curl https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb --output opera.deb"
-    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("opera.deb") ) == `curl -L https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb.sha256sum`
-    system "alien -tc opera.deb > /dev/null"
+    system "curl https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb -#o opera.deb"
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("opera.deb") ) == `curl -Ls https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb.sha256sum`
+    puts "Archive downloaded".lightgreen
+    system "alien -tc opera.deb"
     system "tar xvf opera-stable-#{version}.tgz > /dev/null"
   end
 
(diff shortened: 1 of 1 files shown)


packages/opera.rb
```diff
@@ -17,10 +17,10 @@ class Opera < Package
 
 
   def self.build
-    system "wget https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb"
-    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("opera-stable_#{version}_amd64.deb") ) == '08ec3d1f1b142db4acaea9961687ad95db123f8567c07b0157e350aeed26a56d'
-    system "alien -tc opera-stable_#{version}_amd64.deb"
-    system "tar xvf opera-stable-#{version}.tgz"
+    system "curl https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb --output opera.deb"
```
uberhacker, line 20: `wget` is easier. :)

packages/opera.rb
```diff
@@ -17,10 +17,10 @@ class Opera < Package
...
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("opera.deb") ) == `curl -L https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb.sha256sum`
```
uberhacker, line 21: This is an interesting way to do it but it's probably easier just to insert the checksum verbatim.

packages/opera.rb
```diff
@@ -17,10 +17,10 @@ class Opera < Package
...
+    system "alien -tc opera.deb > /dev/null"
+    system "tar xvf opera-stable-#{version}.tgz > /dev/null"
```
uberhacker, line 23: I would rather not use `> /dev/null`.  This will hide any errors if they occur.

packages/opera.rb
```diff
@@ -17,9 +17,10 @@ class Opera < Package
 
 
   def self.build
-    system "curl https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb --output opera.deb"
-    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("opera.deb") ) == `curl -L https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb.sha256sum`
-    system "alien -tc opera.deb > /dev/null"
+    system "curl https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb -#o opera.deb"
```
uberhacker, line 20: Change `-#o` to `-o`?

packages/opera.rb
uberhacker: Line 24 has `> /dev/null`.

packages/opera.rb
supechicken: `#` will make `curl` show the progress bar

packages/opera.rb
supechicken: `wget` is old now, and it can reduce one dependency as well

packages/opera.rb
supechicken: It makes updating the package easier, just correct the version number and done :)

packages/opera.rb
supechicken: Only `tar` output hide now, as `tar` won't occur error for the most time

supechicken: `> /dev/null` won't hide error (it will hide `stdout` but keep `stderr`)

supechicken:
Can we run `alien` as root to get rid of the warning message?
```shell
############################################################################################################################### 100.0%
Archive downloaded
Warning: alien is not running as root!
Warning: Ownerships of files in the generated packages will probably be wrong.
opera-stable-72.0.3815.207.tgz generated
Preconfiguring package...
```

packages/opera.rb
uberhacker: We're using the latest wget in core.

packages/opera.rb
uberhacker: I get that but if there is an error, don't you need `stdout` to see where the error occurred?

packages/opera.rb
uberhacker: Fair enough. :)

uberhacker:
No need to run as root since the chronos user already has write access to `/usr/local` which is the destination directory for all the files.

PR #4577, opened 2020-11-13
Weston update to 9.0.0

satmandu:
Weston 9.0 is MUCH faster.

Works properly:
- [x] x86_64

(also does what non-accelerated sommelier does, in a window.)

### DIFF
--- a/packages/weston.rb
+++ b/packages/weston.rb
@@ -45,5 +45,9 @@ class Weston < Package
     puts "echo '[terminal]' >> ~/.config/weston.ini".lightblue
     # Cousine is the defaut ChromeOS monospace font.
     puts "echo 'font=Cousine' >> ~/.config/weston.ini".lightblue
+    puts
+    puts "To run weston with xwayland try something like this:".lightred
+    puts "export WAYLAND_DISPLAY=wayland-1".lightred
+    puts "WAYLAND_DISPLAY=wayland-0 weston -Swayland-1 --xwayland".lightred
   end
 end
--- a/packages/weston.rb
+++ b/packages/weston.rb
@@ -3,12 +3,13 @@ require 'package'
 class Weston < Package
   description 'Weston is the reference implementation of a Wayland compositor, and a useful compositor in its own right.'
   homepage 'http://wayland.freedesktop.org'
-  version '4.0.0'
+  version '9.0.0'
   compatibility 'i686,x86_64'
   case ARCH
   when 'i686', 'x86_64'
-    source_url 'https://github.com/wayland-project/weston/archive/4.0.0.tar.gz'
-    source_sha256 '46b0178cd37b0bf7471c9af12c847c7b8728699ecd5c04ce79be77ea12c98512'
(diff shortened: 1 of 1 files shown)


packages/weston.rb
```diff
@@ -19,29 +20,34 @@ class Weston < Package
...
 
   def self.install
-    system "make DESTDIR=#{CREW_DEST_DIR} install"
+    system "DESTDIR=#{CREW_DEST_DIR} ninja install -C build"
+  end
+  
+  def self.postinstall
+    puts
+    puts "To complete the installation, execute the following:".lightblue
+    puts "mkdir -p ~/.config".lightblue
```
uberhacker, line 42: This step is unnecessary if you include `depends_on 'xdg_base'`.  I was actually thinking about adding it as a dependency of sommelier since so many gui apps require config.

packages/weston.rb
```diff
@@ -19,29 +20,34 @@ class Weston < Package
...
+    puts "echo '[core]' >> ~/.config/weston.ini".lightblue
+    puts "echo 'xwayland=true' >> ~/.config/weston.ini".lightblue
+    puts "echo '[terminal]' >> ~/.config/weston.ini".lightblue
+    # Cousine is the defaut ChromeOS monospace font.
+    puts "echo 'font=Cousine' >> ~/.config/weston.ini".lightblue
```
uberhacker, line 47: Why not include this config file in a build step?  I know uninstalling will remove it but that is a good thing, right?

packages/weston.rb
```diff
@@ -19,29 +20,34 @@ class Weston < Package
...
+    puts
+    puts "To run weston with xwayland try something like this:".lightred
+    puts "export WAYLAND_DISPLAY=wayland-1".lightred
+    puts "WAYLAND_DISPLAY=wayland-0 weston -Swayland-1 --xwayland".lightred
```
uberhacker, line 51: Could we create an environment file that is sourced with the weston executable?  Maybe create a separate bash script to source the file?  That way, users wouldn't need to remember a bunch of configure options.

packages/weston.rb
satmandu: Someone may have customized those settings though. Same argument as for not automatically editing ~/.bashrc?

packages/weston.rb
satmandu: I think weston is more for debugging use. Until it has acceleration and has rootless support, I can't see most people using it as default, since it is a reference compositor.

packages/weston.rb
uberhacker: I suppose so but if we keep doing this, there will be a lot of leftover config residue when users remove packages.

packages/weston.rb
uberhacker: Fair enough.

packages/weston.rb
satmandu: Done.

packages/weston.rb
satmandu: Makes sense. Added as a builtin config file.

PR #4587, opened 2020-11-17
[composer] fix ~/.config recursive symbolic link and existing files

vincowl:


packages/composer.rb
```diff
@@ -14,7 +14,7 @@ class Composer < Package
   def self.preinstall
     if Dir.exists?("#{HOME}/.config") && !File.symlink?("#{HOME}/.config")
       # Save any existing configuration
-      FileUtils.cp_r "#{HOME}/.config", "#{CREW_PREFIX}" unless Dir.empty? "#{HOME}/.config"
+      FileUtils.cp_r "#{HOME}/.config", "#{CREW_PREFIX}", remove_destination: true unless Dir.empty? "#{HOME}/.config"
```
uberhacker, line 17: This sounds dangerous.  What if there is config from other packages?

packages/composer.rb
```diff
@@ -25,10 +25,10 @@ def self.install
     system "curl -Ls -o composer https://github.com/composer/composer/releases/download/#{version}/composer.phar"
     abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('composer') ) == '0a060e8caf1d5cde55c4562fd68369b08cf231b8fd30da9c8f994b111557ebe9'
     system "install -Dm755 composer #{CREW_DEST_PREFIX}/bin/composer"
-    FileUtils.ln_sf "#{CREW_PREFIX}/.config", "#{HOME}/.config"
   end
 
   def self.postinstall
+    FileUtils.ln_sf "#{CREW_PREFIX}/.config", "#{HOME}/.config"
```
uberhacker, line 31: Why does this line need to move?

packages/composer.rb
vincowl: It actually crashes without this option. On my machine, ~/.config is a symbolic link pointing to /usr/local/.config. It contains several configs for other packages. This line seems to use a cache or something like this since the ruby's doc says it should act like a cp -rf... It indeed works on my machine, and yes, with a bit of magic to me ...

packages/composer.rb
vincowl: Otherwise I have a ~/.config/.config/.config/.config/.config .... each time I upgrade or reinstall composer

packages/composer.rb
satmandu: huh! this is the same on my machine:
```
chronos@localhost ~ $ ls -aFl .config
lrwxrwxrwx. 1 chronos chronos 22 Oct 16 00:53 .config -> /usr/local/.config/
```

Does chromebrew make that symlink? Or is that made by something else?

uberhacker: The `xdg_base` package adds the symlink.  The reason we need to do this is some configs store executables here, otherwise the home directory is mounted noexec and these binaries will not execute.

vincowl: No pb to me with the symlink. I was forced to do this since composer crashed without this because of already existing files. The `remove_destination` option is said to act like the `force` option of a bash rm command :https://www.rubydoc.info/stdlib/fileutils/FileUtils#cp_r-class_method

packages/composer.rb
uberhacker: I guess I was asking why it was moved from `install` to the `postinstall` section.

vincowl: I understood it that way indeed. Answer is the same : if it is placed in the install section, I have recursive symbolic links which does not occure if it is placed in postinstall.... Absolutely no idea why....

PR #4599, opened 2020-11-23
glib2 update to 2.67.0

satmandu:
libiconv ldflag is needed to compile for some reason.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/glib.rb
+++ b/packages/glib.rb
@@ -3,29 +3,20 @@ require 'package'
 class Glib < Package
   description 'GLib provides the core application building blocks for libraries and applications written in C.'
   homepage 'https://developer.gnome.org/glib'
-  version '2.64.1'
+  version '2.67.0'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


packages/glib.rb
```diff
@@ -3,29 +3,20 @@
...
-  binary_sha256 ({
-    aarch64: 'ba601edad8d81993f486184431f1951e47e8300366e7655c2ea164e923fcc056',
-     armv7l: 'ba601edad8d81993f486184431f1951e47e8300366e7655c2ea164e923fcc056',
-       i686: '532128e2b88f76fd80cad43e0804e5d6549dfe02f21579158261357c9a4be159',
-     x86_64: '9f0d347094230086b43f7535bd980d1a0efb4b9e5704e74acff9c1eb584dcd04',
-  })
 
   depends_on 'util_linux'
   depends_on 'six'
+  depends_on 'libiconv'
```
uberhacker, line 14: This is already a core dependency so no need to include here.

uberhacker:

PR #4600, opened 2020-11-23
geany update to 1.37 & add geany_plugins

satmandu:
Works properly:
- [x] x86_64

### DIFF
--- a/packages/geany.rb
+++ b/packages/geany.rb
@@ -3,33 +3,19 @@ require 'package'
 class Geany < Package
   description 'Geany is a small and lightweight Integrated Development Environment.'
   homepage 'https://www.geany.org/'
-  version '1.36'
+  version '1.37'
   compatibility 'aarch64,armv7l,x86_64'
   case ARCH
   when 'aarch64', 'armv7l', 'x86_64'
-    source_url 'https://download.geany.org/geany-1.36.tar.bz2'
-    source_sha256 '9184dd3dd40b7b84fca70083284bb9dbf2ee8022bf2be066bdc36592d909d53e'
+    source_url 'https://download.geany.org/geany-1.37.tar.bz2'
+    source_sha256 'dde52584823b769c56704c27dbedddb7a6bbaf4eacb9587d10bbc387816e3d51'
     depends_on 'gtk3'
     depends_on 'xdg_base'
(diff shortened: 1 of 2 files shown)


packages/geany_plugins.rb
```diff
@@ -0,0 +1,33 @@
...
+  version '1.37'
+  compatibility 'aarch64,armv7l,x86_64'
+  case ARCH
+  when 'aarch64', 'armv7l', 'x86_64'
+    source_url 'https://download.geany.org/geany-plugins/geany-plugins-1.37.tar.bz2'
+    source_sha256 'c98f9b1303f4ab9bed7587e749cd0b5594d9136a1bf8ba110900d46a17fa9cd8'
+    depends_on 'gtk3'
+    depends_on 'xdg_base'
+    depends_on 'sommelier'
+    depends_on 'geany'
```
uberhacker, line 15: Since geany already has `gtk3`, `xdg_base` and `sommelier` dependencies, no need to add here.

uberhacker:
Nice!  Thank you for the update @satmandu!

packages/geany.rb
```diff
@@ -3,33 +3,19 @@
...
-     armv7l: 'e89045825ce51fbb6f4705f3d408d1dada843bc54ff24ad35f18cea61b4b03b3',
-     x86_64: '3d99c7c90b6ea63856fd9b469cdbdb4c2f00db6a549d7f3973e5efea25b65536',
-  })
-
   def self.build
-    system './configure',
-           "--prefix=#{CREW_PREFIX}",
-           "--libdir=#{CREW_LIB_PREFIX}",
-           '--enable-gtk3'
+    system "./configure #{CREW_OPTIONS} --enable-api-docs=no --enable-html-docs=no --enable-pdf-docs=no"
```
uberhacker, line 18: Where is the `--enable-gtk3` option?  If you don't include it, does it use gtk2 instead?

satmandu: It appears to have switched to ```--enable-gtk2``` being the optional dependency.

packages/geany_plugins.rb
satmandu: adjusted accordingly

Issue #4589, opened 2020-11-20
Need X fonts package to prevent some xorg app font complaints


satmandu:
The standard install doesn't include needed xorg fonts.
Pretty sure this is why I get errors like this as per [stackexchange](https://unix.stackexchange.com/questions/407710/xterm-can-not-load-font-misc-fixed-medium-r-semicondensed-13-120-75-75-c-60-i):
```
xterm: cannot load font "-misc-fixed-medium-r-semicondensed--13-120-75-75-c-60-iso10646-1"
```
I think we need some version of ports of these arch pkgbuilds to get the standard x fonts:
https://github.com/archlinux/svntogit-packages/blob/master/xorg-fonts-100dpi/trunk/PKGBUILD
https://github.com/archlinux/svntogit-packages/blob/master/xorg-fonts-75dpi/trunk/PKGBUILD
https://github.com/archlinux/svntogit-packages/blob/master/xorg-fonts-type1/trunk/PKGBUILD
https://github.com/archlinux/svntogit-packages/blob/master/xorg-fonts-cyrillic/trunk/PKGBUILD
https://github.com/archlinux/svntogit-packages/blob/master/xorg-fonts-misc/trunk/PKGBUILD

(I hope I'm not missing any)

We may also need an alias package like this:
https://github.com/archlinux/svntogit-packages/raw/packages/xorg-fonts-alias/trunk/PKGBUILD

I've adapted that one thus:

```
# Adapted from Arch Linux xorg-fonts-alias PKGBUILD at:
# https://github.com/archlinux/svntogit-packages/raw/packages/xorg-fonts-alias/trunk/PKGBUILD

require 'package'

class Xorg_fonts_alias < Package
  description 'X.org font alias files'
  version '1.0.4'
  compatibility 'all'
  source_url 'https://xorg.freedesktop.org/releases/individual/font/font-alias-1.0.4.tar.bz2'
  source_sha256 'f3111ae8bf2e980f5f56af400e8eefe5fc9f4207f4a412ea79637fd66c945276'

  depends_on 'xorg_util_macros'

  def self.build
    system "./configure #{CREW_OPTIONS} --with-fontrootdir=#{CREW_PREFIX}/share/fonts"
    system "make"
  end
  def self.install
    system "DESTDIR=#{CREW_DEST_DIR} make -C 100dpi install"
    system "DESTDIR=#{CREW_DEST_DIR} make -C 75dpi install"
[...]

PR #4601, opened 2020-11-24
Add adobe fonts for #4589

ThatGeekyWeeb:
PR adds adobe fonts for #4589, as well as `encodings`.
> 75dpi fonts will be added in later PR

### DIFF
--- a/packages/encodings.rb
+++ b/packages/encodings.rb
@@ -1,0 +1,19 @@
+require 'package'
+
+class Encodings < Package
+  description 'X11 Font Index Generator'
+  homepage 'https://xorg.freedesktop.org'
+  version '1.0.5'
+  compatibility 'all'
+  source_url 'https://www.x.org/releases/individual/font/encodings-1.0.5.tar.bz2'
+  source_sha256 'bd96e16143a044b19e87f217cf6a3763a70c561d1076aad6f6d862ec41774a31'
+
+  def self.build
+      system "./configure #{CREW_OPTIONS} --with-fontrootdir=#{CREW_PREFIX}/share/fonts"
+      system "sed -e 's|^\(encodings_DATA = $(DATA_FILES)\).*|\1|' -i Makefile" # Found in xbps-src  
+      system 'make'
+  end
+  def self.install
+      system "make install DESTDIR=#{CREW_DEST_DIR}"
+  end
+end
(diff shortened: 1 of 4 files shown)


packages/encodings.rb
```diff
@@ -0,0 +1,19 @@
...
+  description 'X11 Font Index Generator'
+  homepage 'https://xorg.freedesktop.org'
+  version '1.0.5'
+  compatibility 'all'
+  source_url 'https://www.x.org/releases/individual/font/encodings-1.0.5.tar.bz2'
+  source_sha256 'bd96e16143a044b19e87f217cf6a3763a70c561d1076aad6f6d862ec41774a31'
+
+  def self.build
+      system "./configure #{CREW_OPTIONS} --with-fontrootdir=#{CREW_PREFIX}/share/fonts"
+      system "sed -e 's|^\(encodings_DATA = $(DATA_FILES)\).*|\1|' -i Makefile" # Found in xbps-src
```
uberhacker, line 13: Can we put this line in a `def self.patch` section?

uberhacker:
Thank you for the new packages @ThatGeekyWeeb!

packages/encodings.rb
ThatGeekyWeeb: does `self.patch` run before or after `self.build` though? @uberhacker

uberhacker: Yes, the order of operations is below:
```
  # Function to perform patch operations prior to build from source.                                                                                                                          
  def self.patch                                                                                                                                                                              
                                                                                                                                                                                              
  end                                                                                                                                                                                         
                                                                                                                                                                                              
  # Function to perform pre-build operations prior to build from source.                                                                                                                      
  def self.prebuild                                                                                                                                                                           
                                                                                                                                                                                              
  end                                                                                                                                                                                         
                                                                                                                                                                                              
  def self.build                                                                                                                                                                              
                                                                                                                                                                                              
  end                                                                                                                                                                                         
```
Might be good to add this info in Manual.md also. :)

ThatGeekyWeeb: The `sed` command needs to be done in the make file, so `./configure` has to run before it. @uberhacker

uberhacker: No Makefile exists before the `./configure` command is run?

ThatGeekyWeeb: `./configure` generates the `Makefile`, thats the purpose of automake... @uberhacker

PR #4605, opened 2020-11-24
glib: Don't use dev version, and use option for external iconv

satmandu:
This takes out the LDFLAGS fix, which pollutes other compiles.

This is a cleaner fix for glib not finding iconv.

(This also downgrades to 2.66.3, which is the current non-dev version.)

Works properly:
- [x] x86_64


ThatGeekyWeeb (approved):
xbps-src is currently using `2.66.2`, the downgrade seems like not a bad idea. Approval from me!

uberhacker:
Thank you for the update @satmandu!  Simpler is usually better most of the time anyway.

PR #4607, opened 2020-11-24
xwayland: One Last Fix...

satmandu:
Sigh...

Works properly:
- [x] x86_64


satmandu:
@uberhacker About those regrets...

uberhacker:
Yep, that was a good catch.

PR #4613, opened 2020-11-25
WIP: BREAKING CHANGE: isl -> 0.23

satmandu:
Not sure how to best do this.

gcc occasionally complains about a missing ``` libisl.so.19``` after this is installed.

Recompiling gcc10 might be an option?

But gcc also complains about libiconv issues during compilation. (Is there a special dance with uninstalling libiconv, then reinstalling glibc, and then recompiling gcc10?)

Works properly:
- [ ] x86_64 (gcc whines about missing libisl.so.19)

### DIFF
--- a/packages/isl.rb
+++ b/packages/isl.rb
@@ -18,8 +18,8 @@ class Isl < Package
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
     system "mkdir -p #{CREW_DEST_LIB_PREFIX}"
     # For backwards compatibility:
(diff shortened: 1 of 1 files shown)


uberhacker:
I cringe at the thought of updating core packages for this very reason.  Ideally, we wouldn't need any of the backward compatibility symlinks if everything was recompiled.  Not something I'm very eager to do, however.

satmandu:
This all goes back to needing a proper dependency infrastructure checker which can tell us which symbols from which shared libraries are being used by what package. :/

uberhacker:
See #4747.  gcc10 doesn't break for me with this update.

PR #4616, opened 2020-11-25
RFC/WIP ruby wrapper for crew

satmandu:
```crew``` doesn't work if gcc is broken (uninstalled, etc.)

```crewfix``` can fix this, but why not make that automatic, so that one doesn't have to know to invoke crewfix?

By invoking ruby with a wrapper which runs ```crewfix``` if ruby is broken, we can deal with that automatically.

But there's a chicken/egg problem where ```ruby_safe``` needs to be installed before dropping it into ```crew```.

I put ```ruby_safe``` into gcc_tools, but that would have to be installed/updated before ```crew``` is changed.

Also, is there unnecessary overhead from invoking crew through a bash or shell wrapper?

Works properly:
- [ ] x86_64 (needs ruby_safe installed first!)

### DIFF
--- a/crew
+++ b/crew
@@ -1,4 +1,4 @@
-#!/usr/bin/env ruby
+#!/usr/bin/env ruby_safe
 require_relative 'lib/color'
 
 # Disallow sudo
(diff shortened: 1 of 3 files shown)


uberhacker:
I was thinking it might even make sense to update crewfix to include all missing core packages, not just gcc.  There are a lot of things that can break when core packages have been removed.

PR #4623, opened 2020-11-25
REVERT libdb to last unencumbered version

satmandu:
The 5.3 series is the last unencumbered version of libdb. Everything after that is under a different license, and distributions like debian won't even include newer versions.

Also added the ```--enable-dbm``` option which is needed for packages like linux_pam to compile properly.

Works properly:
- [x] x86_64

Yes this may cause regressions in packages, but I haven't seen any packages on my system I haven't built myself using libdb.

### DIFF
--- a/packages/libdb.rb
+++ b/packages/libdb.rb
@@ -2,10 +2,10 @@ require 'package'
 
 class Libdb < Package
   description 'Berkeley DB is a family of embedded key-value database libraries providing scalable high-performance data management services to applications.'
-  homepage 'http://www.oracle.com/technetwork/database/database-technologies/berkeleydb/overview/index.html'
+  homepage 'https://github.com/berkeleydb/libdb'
   version '5.3.28'
   compatibility 'all'
-  source_url 'http://deb.debian.org/debian/pool/main/d/db5.3/db5.3_5.3.28+dfsg1.orig.tar.xz'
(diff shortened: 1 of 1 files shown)


packages/libdb.rb
```diff
@@ -3,34 +3,22 @@
 class Libdb < Package
   description 'Berkeley DB is a family of embedded key-value database libraries providing scalable high-performance data management services to applications.'
   homepage 'http://www.oracle.com/technetwork/database/database-technologies/berkeleydb/overview/index.html'
-  version '6.2.32'
+  version '5.3.28'
   compatibility 'all'
-  source_url 'https://download.oracle.com/berkeley-db/db-6.2.32.tar.gz'
-  source_sha256 'a9c5e2b004a5777aa03510cfe5cd766a4a3b777713406b02809c17c8e0e7a8fb'
+  source_url 'http://deb.debian.org/debian/pool/main/d/db5.3/db5.3_5.3.28+dfsg1.orig.tar.xz'
```
uberhacker, line 8: I don't doubt Oracle has bastardized this package.  Does it make sense to change the homepage to https://github.com/berkeleydb/libdb?  We could probably use the release source from here also?

satmandu: Good idea.

satmandu: Done.

PR #4629, opened 2020-11-26
PCRE cleanups

satmandu:
GLIB was telling me that PCRE wasn't compiled with UTF8, so this appears to fix this.

Works properly:
- [x] x86_64


uberhacker:
Why not upgrade to https://ftp.pcre.org/pub/pcre/pcre2-10.35.tar.bz2?

uberhacker:
Never mind.  I see the other package already covers that.

PR #4639, opened 2020-11-26
glib: use CREW_MESON_OPTIONS

satmandu:
This should make this a non-debug release build.

Works properly:
- [x] x86_64


uberhacker:
Update available? https://download.gnome.org/sources/glib/2.67/glib-2.67.0.tar.xz

satmandu:
Done.

PR #4657, opened 2020-11-28
add libical

satmandu:
This isn't 3.0.9, which hasn't been tagged yet, but 3.0.8 doesn't compile.

(Trying to get gnome dependencies sorted out.)

Works properly:
- [x] x86_64

### DIFF
--- a/packages/libical.rb
+++ b/packages/libical.rb
@@ -8,21 +8,20 @@ class Libical < Package
   source_sha256 '5fdf39b0ea654589bf2e9aa88ef1b0206c18aa6987f5a4299db3748e8b1eefd6'
 
   depends_on 'glib'
-  depends_on 'icu'
+  depends_on 'icu4c'
   depends_on 'gtk_doc' => ':build'
-  depends_on 'doxygen' => ':build'
   depends_on 'vala' => ':build'
   depends_on 'gobject_introspection' => ':build'
 
-def self.build
-    Dir.mkdir 'build'
-      Dir.chdir 'build' do
-      system "cmake #{CREW_CMAKE_OPTIONS} .. -G Ninja \
-      -DGOBJECT_INTROSPECTION=true \
-      -DICAL_GLIB_VAPI=true \
-      -DICAL_BUILD_DOCS=false \
-      -DLIBICAL_BUILD_TESTING=false"
-      end
+  def self.build
+  Dir.mkdir 'build'
+    Dir.chdir 'build' do
+    system "cmake #{CREW_CMAKE_OPTIONS} .. -G Ninja \
+    -DGOBJECT_INTROSPECTION=true \
+    -DICAL_GLIB_VAPI=true \
+    -DICAL_BUILD_DOCS=false \
+    -DLIBICAL_BUILD_TESTING=false"
+    end
   system "ninja -C build"
   end
   
--- a/packages/libical.rb
+++ b/packages/libical.rb
@@ -14,18 +14,18 @@ class Libical < Package
   depends_on 'gobject_introspection' => ':build'
 
   def self.build
-  Dir.mkdir 'build'
+    Dir.mkdir 'build'
     Dir.chdir 'build' do
-    system "cmake #{CREW_CMAKE_OPTIONS} .. -G Ninja \
-    -DGOBJECT_INTROSPECTION=true \
-    -DICAL_GLIB_VAPI=true \
-    -DICAL_BUILD_DOCS=false \
-    -DLIBICAL_BUILD_TESTING=false"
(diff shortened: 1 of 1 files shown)


packages/libical.rb
```diff
@@ -0,0 +1,32 @@
...
+
+class Libical < Package
+  description 'An open source reference implementation of the icalendar data type and serialization format'
+  version '3.0.8.99' # 3.0.8 doesn't compile.
+  compatibility 'all'
+  source_url 'https://github.com/libical/libical/archive/07dc84fb0d5e50fb493005e10fa41930111da48f.zip'
+  source_sha256 '5fdf39b0ea654589bf2e9aa88ef1b0206c18aa6987f5a4299db3748e8b1eefd6'
+
+  depends_on 'glib'
+  depends_on 'icu'
```
uberhacker, line 11: Does this package exist?  I know `icu4c` does but since it's a core package it doesn't need listed anyway.

packages/libical.rb
```diff
@@ -0,0 +1,32 @@
...
+  depends_on 'gtk_doc' => ':build'
+  depends_on 'doxygen' => ':build'
```
uberhacker, line 13: `doxygen` is already in core so no need to list here.

packages/libical.rb
```diff
@@ -0,0 +1,32 @@
...
+def self.build
+    Dir.mkdir 'build'
+      Dir.chdir 'build' do
+      system "cmake #{CREW_CMAKE_OPTIONS} .. -G Ninja \
+      -DGOBJECT_INTROSPECTION=true \
+      -DICAL_GLIB_VAPI=true \
+      -DICAL_BUILD_DOCS=false \
+      -DLIBICAL_BUILD_TESTING=false"
+      end
+  system "ninja -C build"
```
uberhacker, line 26: Please fix the indentation.  It should be 2 spaces for each level with matching blocks in the same column.

uberhacker:

packages/libical.rb
satmandu: Ah! changing this.

packages/libical.rb
satmandu: I need to make a whitelist in my arch conversion script...

packages/libical.rb
satmandu: is this what you mean?
```
require 'package'

class Libical < Package
  description 'An open source reference implementation of the icalendar data type and serialization format'
  version '3.0.8.99' # 3.0.8 doesn't compile.
  compatibility 'all'
  source_url 'https://github.com/libical/libical/archive/07dc84fb0d5e50fb493005e10fa41930111da48f.zip'
  source_sha256 '5fdf39b0ea654589bf2e9aa88ef1b0206c18aa6987f5a4299db3748e8b1eefd6'

  depends_on 'glib'
  depends_on 'icu4c'
  depends_on 'gtk_doc' => ':build'
  depends_on 'vala' => ':build'
  depends_on 'gobject_introspection' => ':build'

  def self.build
  Dir.mkdir 'build'
    Dir.chdir 'build' do
    system "cmake #{CREW_CMAKE_OPTIONS} .. -G Ninja \
    -DGOBJECT_INTROSPECTION=true \
    -DICAL_GLIB_VAPI=true \
    -DICAL_BUILD_DOCS=false \
    -DLIBICAL_BUILD_TESTING=false"
    end
  system "ninja -C build"
  end
  
  def self.install
  end
end
```

uberhacker: I was thinking:
```
  def self.build
    Dir.mkdir 'build'
    Dir.chdir 'build' do
      system "cmake #{CREW_CMAKE_OPTIONS} .. -G Ninja \
        -DGOBJECT_INTROSPECTION=true \
        -DICAL_GLIB_VAPI=true \
        -DICAL_BUILD_DOCS=false \
        -DLIBICAL_BUILD_TESTING=false"
    end
    system "ninja -C build"
  end

  def self.install
  end
```

satmandu: done

packages/libical.rb
```diff
@@ -0,0 +1,31 @@
...
+
+class Libical < Package
+  description 'An open source reference implementation of the icalendar data type and serialization format'
+  version '3.0.8.99' # 3.0.8 doesn't compile.
+  compatibility 'all'
+  source_url 'https://github.com/libical/libical/archive/07dc84fb0d5e50fb493005e10fa41930111da48f.zip'
+  source_sha256 '5fdf39b0ea654589bf2e9aa88ef1b0206c18aa6987f5a4299db3748e8b1eefd6'
+
+  depends_on 'glib'
+  depends_on 'icu4c'
```
uberhacker, line 11: Not needed since this is a core package?

PR #4658, opened 2020-11-28
icu4c -> 68.1 (requires patched libxml2, recompile of wget)

satmandu:
Works properly:
- [x] x86_64

### DIFF
--- a/packages/icu4c.rb
+++ b/packages/icu4c.rb
@@ -25,6 +25,13 @@ class Icu4c < Package
   def self.install
     FileUtils.cd('source') do
       system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+      # Backwards compatibility symlinks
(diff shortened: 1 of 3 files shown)


uberhacker:
This is a core package so we will need binaries.  Did you test if this breaks gcc or any of the other compilers?

satmandu:
This does create breakage... so best to queue this up when more packages need to get changed.
e.g. libxml will need this patch to work with ice 68.1: https://gitlab.gnome.org/GNOME/libxml2/-/commit/b516ed189eb440e909f36baca1557b98e4d9ffd7.patch

uberhacker:
Yeah, maybe we should just hold off until it stabilizes a bit.  Any reason the older version needs to be updated?

satmandu:
I was just going through trying to troubleshoot issues with trying to compile gnome's evolution_data_server, and there was a complaint about this library.  68.1 IS a major release though as per https://github.com/unicode-org/icu/releases

satmandu:
This also requires wget to be recompiled after a recompiled libxml2 is installed.

uberhacker:
See #4747.

PR #4673, opened 2020-11-30
fribidi -> 1.0.10

satmandu:
Also, switch to meson.

Works properly:
- [x] x86_64


uberhacker:
Nice, meson builds are a bit more elegant.

PR #4674, opened 2020-11-30
nss: Verbose build to show errors

satmandu:
(Otherwise errors are silent.)

This shows that there is a python2 error with the build...

Works properly:
- [x] x86_64


uberhacker:
@satmandu: I can only get this to build for i686.  Both armv7l and x86_64 are dead in the water.  I literally tried 20 different ways to compile without success.

satmandu:
I think python27 is broken due to a broken openssl/libressl link?

PR #4675, opened 2020-11-30
Python2.7: fix pip

### LINKED ISSUE #4631: [nss] Failed to build
```
nss failed to build: `CPPFLAGS='-I/usr/local/include/nspr' ./build.sh --opt --gcc --system-nspr --system-sqlite --disable-tests` exited with 1
```

satmandu:
Fixes #4631 
 (NSS refuses to compile with broken python2.7, which doesn't install a working pip due to setuptools no longer supporting python 2.7)

Works properly:
- [x] x86_64

### DIFF
--- a/packages/python27.rb
+++ b/packages/python27.rb
@@ -3,34 +3,31 @@ require 'package'
 class Python27 < Package
   description 'Python is a programming language that lets you work quickly and integrate systems more effectively.'
   homepage 'https://www.python.org/'
-  version '2.7.17'
+  version '2.7.17-1'
   compatibility 'all'
   source_url 'https://www.python.org/ftp/python/2.7.17/Python-2.7.17.tar.xz'
   source_sha256 '4d43f033cdbd0aa7b7023c81b0e986fd11e653b5248dac9144d508f11812ba41'
 
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/python27-2.7.17-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


uberhacker:
This is a core package so we will need new binaries.  I'll try and do this later in the evening after work.

uberhacker:
Update available? https://www.python.org/ftp/python/2.7.18/Python-2.7.18.tar.xz

satmandu:
Good catch. Updated version.

uberhacker:
After installing, I received the following:
```
$ pip -V
bash: /usr/local/bin/pip: /usr/local/tmp/crew/dest/usr/local/bin/python: bad interpreter: No such file or directory
```

satmandu:
Same issue as with python3 pip. Adjusting that now. Also having ```pip2``` be the python2 version of pip, and have both ```pip``` and ```pip3``` be the python3 versions of pip. This way there isn't a file conflict.

cstrouse:
Python recommends using `python -m pip` rather than `pip` directly.

satmandu:
@cstrouse This is /usr/local/bin/pip:
```
#!/usr/local/bin/python3
# -*- coding: utf-8 -*-
import re
import sys
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```
If installed from python2 it puts python instead of python3 into that file, so I just have /usr/local/bin/pip removed from the python27 install so it doesn't conflict with the file installed from python3.

Is that reasonable?

satmandu:
There's another question of whether the ```/usr/local/bin/python``` should point to python27 or python3...

uberhacker:
See #4687.

PR #4676, opened 2020-11-30
Sommelier: Add .bashrc code to load sommelier if not there

satmandu:
People aren't going to remember to add the code to .bashrc to enable sommelier.

Works properly:
- [x] x86_64


uberhacker:
Good idea.  I like it!

PR #4679, opened 2020-12-01
xwayland -> 1.20.10

satmandu:
Two Xorg security advisories are fixed in just released 1.20.10 as per https://lists.x.org/archives/xorg-announce/2020-December/003066.html

Also: 
Replaced xinit dependency with graphite. (xinit creates a circular dependency.)
Removed curl continue options which aren't supported on the googlesource servers anyways.
Removed meson options not supported by 1.20.10

Works properly:
- [x] x86_64


uberhacker:
Nice, this is better!

PR #4685, opened 2020-12-03
neofetch -> 7.1.0

satmandu:
Works properly:
- [x] x86_64
Screenshot 2020-12-03 at 09 15 36


uberhacker:

PR #4692, opened 2020-12-03
Add sommelier xwayland dri path

satmandu:
Looks like this is needed for acceleration on my system, and probably others too.

Works properly:
- [x] x86_64


uberhacker:
Did I miss the changes?

PR #4693, opened 2020-12-04
gsettings_desktop_schemas: Fix compile

satmandu:
meson needed builddir passed to it.

Works properly:
- [x] x86_64


uberhacker:
Nice!  So, hopefully, this will enable us to generate binaries for all architectures.

PR #4697, opened 2020-12-04
libsoup -> 2.72

satmandu:
Switched to meson, but can't seem to compile w/o clang.

Works properly:
- [x] x86_64


uberhacker:
Thank you for the update!

PR #4698, opened 2020-12-04
Add pkgconf

satmandu:
(This is used for some packages where pkg-config can fail.)

Works properly:
- [x] x86_64


uberhacker:
LGTM but is this an alternative to pkgconfig?  What advantages does it offer, if any?

satmandu:
This is a somewhat drop-in alternative to pkgconfig, which supposedly does a better job of figuring out how dependencies should be linked.

I was messing with it because I was having some pkgconfig problems.

uberhacker:
Fair enough.

PR #4699, opened 2020-12-04
Add amtk

satmandu:
Works properly:
- [x] x86_64


packages/amtk.rb
```diff
@@ -0,0 +1,28 @@
+require 'package'
+
+class Amtk < Package
+  description 'Actions, Menus and Toolbars Kit for GTK+ applications'
+  homepage 'https://wiki.gnome.org/Projects/Amtk'
+  version '5.2.1'
```
uberhacker, line 6: Shouldn't this version be the first few characters of the git hash instead?

satmandu: Yeah that would make sense.

PR #4709, opened 2020-12-04
[WIP] Add deb file support for crew

supechicken666:
Add .deb file support for crew unarchive process

### DIFF
--- a/crew
+++ b/crew
@@ -523,6 +523,11 @@ def unpack (meta)
       puts "Unpacking archive using 'tar', this may take a while..."
       _verbopt = @opt_verbose ? 'v' : ''
       system "tar", "x#{_verbopt}f", meta[:filename], "-C", "#{extract_dir}"
+    when /\.deb$/i
+      puts "Unpacking deb using 'ar' 'tar', this may take a while..."
+      _verbopt = @opt_verbose ? 'v' : ''
+      system "ar", "x", meta[:filename]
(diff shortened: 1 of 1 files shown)


uberhacker:
@supechicken666: I don't think this is as simple as you suggest.  Some deb packages have files that extract into read-only directories in ChromeOS.  Although it's possible to install from deb packages, sometimes you have to move things around to get them to install without error.  Something like this was attempted in the past and it didn't go over well as @skiman6010 can attest.  See https://github.com/skycocker/chromebrew/pull/3052.

supechicken:
Can I add the download and unarchive `.deb` file support only? it can replace
```shell
    system 'curl https://repo.skype.com/deb/pool/main/s/skypeforlinux/skypeforlinux_8.65.0.78_amd64.deb -#o skype.deb'
    puts "Archive downloaded".lightgreen
    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('skype.deb') ) == '9026f043697d7d0783e7232701a868229d9a91e91dc0ae86509732c1c7bf0c13'
```

uberhacker:
I think it will probably create more confusion for packagers.  There are only a handful of packages that need to be extracted this way.  Once you go down this road, what about support for other package managers like rpm, zipper and pacman?  Do you see how this could end up as a slippery slope?

uberhacker:
Sorry @supechicken.  You were right about this.  Maybe you can add support for rpm, zipper, pacman and/or slackware. :)

PR #4711, opened 2020-12-04
libffi -> 3.3

satmandu:
This will need binaries.

Works properly:
- [x] x86_64


uberhacker:
This is another core package so we will need to hold off until pre-built binaries are generated.

uberhacker:
See #4747.

PR #4715, opened 2020-12-05
mandb:  Minor fix for issue with post-install on docker images.

satmandu:
Docker installs complain about the post-install section being unable to find libraries while running mandb -c.

``` "mandb: error while loading shared libraries: libgdbm.so.6:"```

This fixes that issue. **No rebuild is required for this change.**

Works properly:
- [x] x86_64

### DIFF
--- a/packages/mandb.rb
+++ b/packages/mandb.rb
@@ -56,6 +56,6 @@ class Mandb < Package
   end
 
   def self.postinstall
-    system 'mandb -c'
+    system "LD_LIBRARY_PATH=#{CREW_LIB_PREFIX} mandb -c"
   end
 end

### CODE REVIEW

File: packages/mandb.rb [hunk 1]
```diff
@@ -56,6 +56,6 @@ def self.install
   end
 
   def self.postinstall
-    system 'mandb -c'
+    system "LD_LIBRARY_PATH=#{CREW_LIB_PREFIX} mandb -c"
```
uberhacker, line 59: Not sure why you would need to do this since this is the default path.

satmandu:
You would think so, but it doesn't work without the path being specified.
(I've noticed some weird stuff with env variables not being passed to
she'll commands recently.)

satmandu:
Might be a bug with this version of mandb. There is a newer version...

satmandu:
Let's see if the new version fixes this.

PR #4716, opened 2020-12-05
mandb -> 2.9.3

satmandu:
This will need packaging.

Works properly:
- [x] x86_64


uberhacker:
Yeah, we need one big PR with all the updated core package versions including pre-built binaries.

uberhacker:
See #4747.

Issue #4700, opened 2020-12-04
[glade] Failed to build


uberhacker:
```
[84/248] Generating Gladeui-2.0.gir with a custom command
FAILED: gladeui/Gladeui-2.0.gir 
[...]

PR #4718, opened 2020-12-05
glade: Use lld

satmandu:
Fixes some gl lib issues during install.

Works properly:
- [x] x86_64


satmandu:
May or may not help with https://github.com/skycocker/chromebrew/issues/4700

uberhacker:
Worth a shot!

PR #4723, opened 2020-12-06
Use current bash prompt settings in crew installed bash

supechicken:


satmandu:
Any reason why this postinstall bit says puts ```"echo '#{CREW_PREFIX}/bin/sh' >> ~/.bashrc".lightblue``` in lieu of ```puts "echo '#{CREW_PREFIX}/bin/bash' >> ~/.bashrc".lightblue```

We're editing the .bashrc... so we're already expecting bash... and 
that's a symlink to bash anyways:
```
ls -aFl /usr/local/bin/sh
lrwxrwxrwx. 1 chronos chronos 19 Dec  4 01:06 /usr/local/bin/sh -> /usr/local/bin/bash*
```

satmandu:
Let me answer my own question: You need to invoke bash as sh since then it uses the ```--norc``` option, which does this:
```Don’t read the ~/.bashrc initialization file in an interactive shell. This is on by default if the shell is invoked as sh.```
This prevents a loop at startup.

**Comment by satmandu** [comment]
Also this also works, but the PS bit is far more elegant.

```if [[ "${BASH_VERSION}" != $(/usr/local/bin/sh --version | head -n 1 | awk '{print $4}') ]]; then /usr/local/bin/sh -il && exit ; fi```

**Comment by satmandu** [comment]
@supechicken @uberhacker Any thoughts on using 
```exec /usr/local/bin/sh ``` 
via 
```puts "echo 'exec #{CREW_PREFIX}/bin/sh' >> ~/.bashrc".lightblue```
in lieu of 
```
    puts "echo '#{CREW_PREFIX}/bin/sh' >> ~/.bashrc".lightblue
```
This way exiting the shell drops you back to crosh.

**Comment by supechicken** [comment]
@satmandu If we start `bash` with `--norc` option, the custom variables in `~/.bashrc` will load in the first `bash` shell started by `crosh`  but not the shell we start in `.bashrc`, correct?

**Comment by supechicken** [comment]
Here is my idea:
```shell
if [[ `pwdx $BASHPID` != "/usr/local/bin" ]]; then
        exec /usr/local/bin/bash
fi
```
If the current running `bash` shell isn't  located at `/usr/local`, then start `/usr/local/bin/bash`

satmandu:
I'm thinking about how these are horrible horrible hacks to get around /etc/passed being read-only so we can't just run chsh properly.

If we're going to use terrible hacks, why not just bind-mount a static bash into /bin/bash or bind-mount an edited /etc/passwd with the shell changed?

I'm doing bind mounts like this from my .bashrc:
```
if findmnt -i /etc/fonts/fonts.conf >/dev/null 2>&1; then
	sudo mount -io ro,bind ~/fonts/fonts.conf /etc/fonts/fonts.conf
fi

if findmnt -i /root  >/dev/null 2>&1; then
       sudo mount -io ro,bind /usr/local/root /root
fi
```

satmandu:
Here's my current solution in my .bashrc:
```
if [[ "${BASH_VERSION}" != $(/usr/local/bin/bash --version | head -n 1 | awk '{print $4}') ]]; then
  if findmnt -i /bin/bash  >/dev/null 2>&1; then
       sudo mount -io ro,bind /usr/local/bin/bash.static /bin/bash
  fi
fi
```

PR #4731, opened 2020-12-07
Ncurses: Update & enable static libraries, enable libtinfo generation

satmandu:
- Ncurses is currently built with ```without-normal``` which leads to other applications (such as bash) unable to be built as a static binary. Removed that line.
- Added the ```--without-tests``` line.
- Updated to 6.2-20201205 patchversion.
- Added ```--with-termlib``` to generate libtinfo
- Added symlink generation for libtinfo{,w}5.so from libtinfo{,w}6.so .

Works properly:
- [x] x86_64
- [x] armv7l

PR #4732, opened 2020-12-07
bash: Update to 5.1 final, enable static build.

satmandu:
- Update to 5.1 final
- Static Link (since setting this as shell is dangerous if libraries shift underneath it during an update.)
- Dependent upon https://github.com/skycocker/chromebrew/pull/4731 and lld from llvm, since ld seems unable to handle the static linking.

Works properly:
- [x] x86_64
- [x] armv7l

### DIFF
--- a/packages/bash.rb
+++ b/packages/bash.rb
@@ -3,26 +3,27 @@ require 'package'
 class Bash < Package
   description 'The Bourne Again SHell'
   homepage 'https://www.gnu.org/software/bash/'
-  version '5.1-rc2'
+  version '5.1'
   compatibility 'all'
-  source_url 'https://ftpmirror.gnu.org/bash/bash-5.1-rc2.tar.gz'
-  source_sha256 'f3274290260e2c74a2ec61606cc932cc0e093028d55c779cf2c7907bdb60abc0'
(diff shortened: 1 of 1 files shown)


packages/bash.rb
```diff
@@ -3,26 +3,27 @@
...
+      --with-gnu-malloc --enable-mem-scramble --enable-usg-echo-default \
+      --enable-single-help-strings --enable-select \
+      --enable-restricted --enable-progcomp --enable-process-substitution \
+      --enable-net-redirections --enable-multibyte --enable-job-control \
+      --enable-history --enable-help-builtin --enable-dparen-arithmetic \
+      --enable-directory-stack --enable-coprocesses --enable-cond-regexp \
+      --enable-cond-command --enable-command-timing --enable-casemod-expansions \
+      --enable-casemod-attributes --enable-brace-expansion --enable-bang-history \
+      --enable-array-variables --enable-arith-for-command --enable-alias \
+      --enable-static-link"
```
uberhacker, line 26: That's a lot of enable lines!

uberhacker:
Thank you for the update @satmandu.

packages/bash.rb
satmandu: Yeah that's what Arch is using, and their templates have served me well before. In any case, I get a usable binary.

PR #4734, opened 2020-12-08
wget is not longer preinstalled in Chrome os

supechicken:
```shell
chronos@localhost / $ whereis wget
wget: /usr/local/bin/wget
chronos@localhost / $ 
```


uberhacker:
Good call!

PR #4735, opened 2020-12-08
Fix constant in libtinfo

uberhacker:
Fixes `/usr/local/lib64/ruby/2.7.0/fileutils.rb:206:in 'mkdir_p': wrong number of arguments (given 0, expected 1) (ArgumentError)`.


satmandu:
Why not just add this to the ncurses package?

Also, adding ```--with-termlib``` to ncurses should generated the libtinfo files automatically as per https://github.com/mirror/ncurses/blob/master/INSTALL, I think?

PR #4736, opened 2020-12-08
Patchelf to .12+

satmandu:
This version was able to work on the fbc freebasic binary, whereas the older version just borked it.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/patchelf.rb
+++ b/packages/patchelf.rb
@@ -3,26 +3,15 @@ require 'package'
 class Patchelf < Package
   description 'PatchELF is a small utility to modify the dynamic linker and RPATH of ELF executables.'
   homepage 'http://nixos.org/patchelf.html'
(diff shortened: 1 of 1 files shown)


uberhacker:
Why do we need patchelf for freebasic?  I'm confused.

satmandu:
@uberhacker I was using patchelf to change the library requirements in fbc from libtinfo.5 to libtinfo.6 but it turns out I didn't need to do that, and just recompiling fbc using the older fbc without mucking with patchelf worked just fine.

But the older patchelf package does break fbc whereas using the newer one does not... so better to have a functional patchelf around.

uberhacker:
I will merge this after you add the binaries.

satmandu:
@uberhacker Binaries are on board.

PR #4738, opened 2020-12-10
binutils -> 2.35.1

satmandu:
- Added lto
- Added vtable-verify (llvm compile was complaining about this.)

Works properly:
- [x] x86_64
- [x] armv7l

### DIFF
--- a/packages/binutils.rb
+++ b/packages/binutils.rb
@@ -3,23 +3,11 @@ require 'package'
 class Binutils < Package
   description 'The GNU Binutils are a collection of binary tools.'
   homepage 'https://www.gnu.org/software/binutils/'
-  version '2.34'
+  version '2.35.1'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


packages/binutils.rb
```diff
@@ -37,8 +25,10 @@ def self.build
               --enable-plugins \
               --disable-bootstrap \
               --enable-64-bit-bfd \
+              --enable-lto \
+              --enable-vtable-verify \
               --disable-werror"
-      system 'make'
+      system "make -j#{CREW_NPROC}"
```
uberhacker, line 31: `-j#{CREW_NPROC}` is not necessary.  This is the default option.

uberhacker:
Thank you for the update @satmandu.  Another core package so we'll need binaries.

packages/binutils.rb
satmandu: Noted.

uberhacker:
See #4747.

Issue #4341, opened 2020-09-24
Create a PWA that launches a crew app from the desktop


uberhacker:
[Progressive Web Apps](https://web.dev/what-are-pwas/) are supposedly the new way to develop Android apps for the desktop.  See https://www.androidcentral.com/google-begins-replacing-android-apps-progressive-web-apps-chrome-os#:~:text=Google%20has%20started%20replacing%20a,also%20taking%20up%20less%20storage.  It would be really cool if we could write a PWA that launches a crew app from the desktop.  This would let us have icons on the desktop to run natively compiled apps for ChromeOS.

daemonspudguy:
Someone needs to make that happen.

uberhacker:
@daemonspudguy: Could that someone be you? :)

daemonspudguy:
@uberhacker I have no skills in HTML, CSS, WebAssembly, or Ruby.

uberhacker:
@daemonspudguy: Sounds like some good things to learn.  Might want to throw in Javascript as well.

supechicken:
In Chrome OS, only the Secure Shell App extension have privilege to access the shell, we must make the PWA shortcut ID same as the Secure Shell App.

EvidenceRepulsive151:
Not sure if that'd be possible, but `crew` could silently run a webserver and make PWAs that'll point to an endpoint that will launch the app the user wants. For example, a Firefox PWA would point to `127.0.0.1:6254/run/firefox?authentication=<random_token_used_to_prevent_websites_from_launching_apps>` and upon launhing that app, it'll close itself somehow and start the real Firefox app.

supechicken:
> Not sure if that'd be possible, but `crew` could silently run a webserver and make PWAs that'll point to an endpoint that will launch the app the user wants. For example, a Firefox PWA would point to `127.0.0.1:6254/run/firefox?authentication=<random_token_used_to_prevent_websites_from_launching_apps>` and upon launhing that app, it'll close itself somehow and start the real Firefox app.

Seems possible, but I am afraid that it may cause battery drain issue since it required a web server running on the background

Issue #4295, opened 2020-09-10
Add Pin to Shelf.


actuallyexeon:
I just got Chromebrew yesterday and I wanted the abiltiy to pin a app to the shelf.
ex: When you click Firefox run the command sommelier -X firefox
And also being able to change that command depending whether the app needs X forwarding or not.

uberhacker:
@actuallyexeon: This is not as simple as it may appear.  I _think_ the way to do it would be a PWA (progressive web app) but I haven't tried it yet.

saltedcoffii:
Crouton seems to have a [WIP article](https://github.com/dnschneid/crouton/wiki/Adding-Crouton-Icons-to-the-shelf-(WIP)) on how to do this. We may be able to hijack their process once they finish developing it.

actuallyexeon:
Hmm

supechicken:
See #6197

PR #4739, opened 2020-12-10
New package: pwashortcut

supechicken666:
Get inspired from #4341 #4295 

Scripts to create a PWA shortcut on ChromeOS dock for Chromebrew GUI Apps using Flask.

Tested on `x86_64`

### DIFF
--- a/packages/pwashortcut.rb
+++ b/packages/pwashortcut.rb
@@ -3,9 +3,9 @@ require 'package'
 class Pwashortcut < Package
   description 'Scripts to create a PWA shortcut on ChromeOS dock for Chromebrew GUI Apps.'
   homepage 'https://github.com/supechicken666/pwashortcut/'
-  version '1.0'
-  source_url 'https://github.com/supechicken666/pwashortcut/raw/master/brew_transparent.png'
-  source_sha256 'dedaa7b9dda61413ff03c823f035bebf5a006dc6a4c37eb8bbd27363e676fd56'
+  version '1.1'
+  source_url 'https://github.com/supechicken666/pwashortcut/archive/1.0-stable.tar.gz'
+  source_sha256 '9e028cb35ce6b0ba840b476d4641be496686d4ee16c73935c48d1fb30ca70d07'
 
   def self.build
     system "pip install flask"
@@ -47,14 +47,14 @@ EOF"
 v1
   \"name\"v4 \"linuxapp\"v6
   \"short_name\"v4 \"linuxapp\"v6
-  \"display\"v4 \"minimal-ui\"v6
+  \"display\"v4 \"standalone\"v6
   \"start_url\"v4 \"unixapp.app\"v6
   \"theme_color\"v4 \"#000000\"v6
   \"background_color\"v4 \"#000000\"v6
   \"icons\"v4 v2
     v1
-      \"src\"v4 \"brew.png\"v6
-      \"sizes\"v4 \"200x200\"v6
+      \"src\"v4 \"icon/brew.png\"v6
+      \"sizes\"v4 \"546x546\"v6
       \"type\"v4 \"image/png\"
     v9
   v8
@@ -72,13 +72,15 @@ EOF"
 # A simple start script
 export CREW_PREFIX=#{CREW_PREFIX}
 export PYTHONPATH=$CREW_PREFIX/lib/python3.9/site-packages/
-export tools=$CREW_PREFIX/lib/pwa/tools
-export PREFIX=$CREW_PREFIX/lib/pwa
-export FLASK_APP=$PREFIX/main.py
+export PWA_PREFIX=$CREW_PREFIX/lib/pwa
+export tools=$PWA_PREFIX/tools
+export FLASK_APP=$PWA_PREFIX/main.py
 export FLASK_ENV=development
+
 pkill flask
-if [[ $1 = '-h' ]]; then
-    echo  '
+case ${1} in
+    -h)
+          echo  '
 ===================================
     Shortcut Server Starter
   -s (Default option) Start shortcut server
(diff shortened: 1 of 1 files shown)


packages/pwashortcut.rb
```diff
@@ -0,0 +1,169 @@
...
+      flask run
+    fi
+fi
+EOF"
+    #########################################
+  end
+
+  def self.install
+    FileUtils.mkdir "#{CREW_PREFIX}/lib/pwa/" unless File.exists? "#{CREW_PREFIX}/lib/pwa/"
+    FileUtils.mkdir "#{CREW_PREFIX}/lib/pwa/tools/" unless File.exists? "#{CREW_PREFIX}/lib/pwa/tools/"
```
uberhacker, line 155: The 2 lines above could be replaced by one line:
```
FileUtils.mkdir_p "#{CREW_PREFIX}/lib/pwa/tools"
```
No need to include the `unless File.exists? ...` since I doubt any other packages are using this path...yet.

packages/pwashortcut.rb
```diff
@@ -0,0 +1,169 @@
...
+    FileUtils.mv "../brew_transparent.png", "#{CREW_PREFIX}/lib/pwa/tools/brew.png"
+    system "install -Dm755 pwashortcut #{CREW_PREFIX}/bin/pwashortcut"
+    system "cp * #{CREW_PREFIX}/lib/pwa/tools/"
```
uberhacker, line 158: `FileUtils.mv ...` instead?

packages/pwashortcut.rb
```diff
@@ -0,0 +1,169 @@
...
+    FileUtils.mv "main.py.bak", "#{CREW_PREFIX}/lib/pwa/main.py"
```
uberhacker, line 159: Might be good to put this line above the previous line.

packages/pwashortcut.rb
```diff
@@ -0,0 +1,169 @@
...
+  end
+  def self.postinstall
+      puts 
+      puts "To complete the installation, execute the following:".lightblue
+      puts "echo 'nohup pwashortcut &> /dev/null' >> ~/.bashrc".libhtblue
```
uberhacker, line 164: Not a good idea to use `&> /dev/null` since it will hide errors.

uberhacker:
Thank you for the new package @supechicken666!  This looks very interesting.  Still waiting on your email request sent to crewonslack@gmail.com.

packages/pwashortcut.rb
```diff
@@ -0,0 +1,169 @@
+require 'package'
+
+class Pwashortcut < Package
+  description 'Scripts to create a PWA shortcut on ChromeOS dock for Chromebrew GUI Apps.'
+  homepage 'https://github.com/supechicken666/pwashortcut/'
+  version '1.0'
+  source_url 'https://github.com/supechicken666/pwashortcut/raw/master/brew_transparent.png'
```
uberhacker, line 7: Probably should use a versioned source or file:///dev/null instead.

packages/pwashortcut.rb
```diff
@@ -0,0 +1,169 @@
...
+    system "sed -i 's/v9/\}/g' manifest.json.bak"
+    system "sed -i 's/v8/]/g' manifest.json.bak"
+    #######################################
+    system "cat <<'EOF'> pwashortcut
+#!/bin/bash
+# A simple start script
+export CREW_PREFIX=#{CREW_PREFIX}
+export PYTHONPATH=$CREW_PREFIX/lib/python3.9/site-packages/
+export tools=$CREW_PREFIX/lib/pwa/tools
+export PREFIX=$CREW_PREFIX/lib/pwa
```
uberhacker, line 76: Might be good to create `PWA_PREFIX` instead to avoid a potential environment variable conflict.

packages/pwashortcut.rb
```diff
@@ -0,0 +1,169 @@
...
+               flask run $OPTARG
+               ;;
+           *)  echo \"Error: unknown option '$arg'\"
+               ;;
+        esac
+    done
+    if [[ $RUN != YES ]]; then
+      flask run
+    fi
+fi
```
uberhacker, line 148: It would be more elegant to use `case ... esac` instead of `if ... elif ... elif ... fi`.

packages/pwashortcut.rb
```diff
@@ -0,0 +1,169 @@
...
+EOF"
```
uberhacker, line 149: Would it make sense to display the usage message by default if no options are provided?

supechicken666:
> Thank you for the new package @supechicken666! This looks very interesting. Still waiting on your email request sent to [crewonslack@gmail.com](mailto:crewonslack@gmail.com).

Just sent:)

packages/pwashortcut.rb
```diff
@@ -0,0 +1,185 @@
...
+    v9
+  v8
+v9
+EOF"
+    system "sed -i 's/v6/,/g' manifest.json.bak"
+    system "sed -i 's/v4/:/g' manifest.json.bak"
+    system "sed -i 's/v1/\{/g' manifest.json.bak"
+    system "sed -i 's/v2/\[/g' manifest.json.bak"
+    system "sed -i 's/v9/\}/g' manifest.json.bak"
+    system "sed -i 's/v8/]/g' manifest.json.bak"
```
uberhacker, line 68: I guess I don't understand why you need to run the sed commands.  Why not just create the manifest file verbatim?

packages/pwashortcut.rb
```diff
@@ -0,0 +1,185 @@
...
+          done
+          if [[ $RUN != YES ]]; then pwashortcut -h; fi
+          ;;
+esac
+EOF"
+    #########################################
+  end
+
+  def self.install
+    system "install -Dm755 pwashortcut #{CREW_PREFIX}/bin/pwashortcut"
```
uberhacker, line 165: `system "install -Dm755 pwashortcut #{CREW_DEST_PREFIX}/bin/pwashortcut"`?

packages/pwashortcut.rb
```diff
@@ -0,0 +1,185 @@
...
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/lib/pwa/templates"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/lib/pwa/tools/icon"
+    FileUtils.cp "installer.html.bak", "#{CREW_DEST_PREFIX}/lib/pwa/templates/installer.html"
+    FileUtils.cp "main.py.bak", "#{CREW_DEST_PREFIX}/lib/pwa/main.py"
```
uberhacker, line 169: You could replace the 4 lines above with:
```
system "install -Dm644 installer.html.bak #{CREW_DEST_PREFIX}/lib/pwa/templates/installer.html"
system "install -Dm644 main.py.bak #{CREW_DEST_PREFIX}/lib/pwa/main.py"
```

packages/pwashortcut.rb
```diff
@@ -0,0 +1,185 @@
...
+                 f)  
+                     RUN=YES
+                     flask run $OPTARG
+                     ;;
+                 *)  
+                     echo \"Error: unknown option '$arg'\"
+                     echo \"Try 'pwashortcut -h' for more options.\"
+                     break
+                     ;;
+              esac
```
uberhacker, line 155: Why do you need to nest this while loop?  Couldn't you handle these options in the parent `case...esac`?

packages/pwashortcut.rb
```diff
@@ -0,0 +1,185 @@
...
+  -s (Default option) Start shortcut server
+  -n (App Name)       Make a new shortcut
+  -h                  Show this message
+  -p                  Set py script path
+  -y                  Set PYTHONPATH
+  -f                  Pass option to Flask
+  -g                  PWA icon chooser
+  -i                  Available preinstalled icons for PWA icon chooser
+==================================='
+          ;;
```
uberhacker, line 95: I was thinking something like:
```
HELP='
===================================
    Shortcut Server Starter
  -s (Default option) Start shortcut server
  -n (App Name)       Make a new shortcut
  -h                  Show this message
  -p                  Set py script path
  -y                  Set PYTHONPATH
  -f                  Pass option to Flask
  -g                  PWA icon chooser
  -i                  Available preinstalled icons for PWA icon chooser
==================================='
case ${1} in
  -h)
    echo $HELP
    ;;
...
  *)
    echo $HELP
    ;;
```

packages/pwashortcut.rb
supechicken666: Because the `:` in manifest will break the quotation marks in ruby

packages/pwashortcut.rb
supechicken666: Done

packages/pwashortcut.rb
supechicken666: Done

packages/pwashortcut.rb
supechicken666: Removed

packages/pwashortcut.rb
supechicken666: I think the variable won't save the format of the help message
```shell
chronos@localhost ~ $ bash test.sh 
=================================== Shortcut Server Starter -s (Default option) Start shortcut server -n (App Name) Make a new shortcut -h Show this message -p Set py script path -y Set PYTHONPATH -f Pass option to Flask -g PWA icon chooser -i Available preinstalled icons for PWA icon chooser ===================================
```

supechicken666:
Screenshot 2020-12-15 at 1 12 13 AM

packages/pwashortcut.rb
uberhacker: Use heredoc instead?

supechicken666:
All features were tested

uberhacker:
Thank you for the screenshot @supechicken.  I'll test it out and give some feedback in the next day or so.

packages/pwashortcut.rb
```diff
@@ -0,0 +1,181 @@
+require 'package'
+
+class Pwashortcut < Package
+  description 'Scripts to create a PWA shortcut on ChromeOS dock for Chromebrew GUI Apps.'
+  homepage 'https://github.com/supechicken666/pwashortcut/'
+  version '1.3.1'
+  source_url 'https://github.com/supechicken666/pwashortcut/archive/1.3.1-debug.tar.gz'
+  source_sha256 'ec4d3e728733550a8cdaa8b8f19c5df01982f9e39dfc1a456e39d52934bf9d89'
+
+  HOME = Dir.home
```
uberhacker, line 10: Where is this variable used?

supechicken666: Getting the `~` directory path for `FileUtils`

packages/pwashortcut.rb
supechicken666: Just found the solution, `echo "$var"` works well:)
```shell
$ echo $var
ab cd ef
$ echo "$var" 

ab
cd
ef
```

PR #4740, opened 2020-12-10
unzip: update ubuntu maintainer patchset

satmandu:
- update "upstream" ubuntu maintainer patchset
- Remove armv7l logic, since this appears to compile just fine on armv7l with the generic target.
- Add ```-DNO_LCHMOD``` CFLAG via flags file since unzip complains occasionally about LCHMOD not being supported.

Works properly:
- [x] x86_64
- [x] armv7l

### DIFF
--- a/packages/unzip.rb
+++ b/packages/unzip.rb
@@ -3,33 +3,22 @@ require 'package'
 class Unzip < Package
   description 'UnZip is an extraction utility for archives compressed in .zip format (also called \'zipfiles\').'
   homepage 'http://www.info-zip.org/UnZip.html'
-  version '1.6_1-1'
+  version '6.0-2'
   compatibility 'all'
   source_url 'https://downloads.sourceforge.net/project/infozip/UnZip%206.x%20%28latest%29/UnZip%206.0/unzip60.tar.gz'
   source_sha256 '036d96991646d0449ed0aa952e4fbe21b476ce994abc276e49d30e686708bd37'
 
-  binary_url ({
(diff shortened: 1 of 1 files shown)


uberhacker:
Another core package.  You're killing me dude. ;)

satmandu:
Sorry. The structural problem is that we don't have a CI setup to automatically compile packages, leaving you to have to compile everything. We should fix that.

uberhacker:
Right, I'm waiting for someone smart like you to help get that going. :)

packages/unzip.rb
```diff
@@ -39,11 +28,9 @@ def self.patch
   end
 
   def self.build
-    if ARCH == "armv7l"
-      system "make -f unix/Makefile linux_noasm"
-    else
-      system "make -f unix/Makefile generic"
-    end
+    # LCHMOD is not supported in linux.
+    system 'echo "CFLAGS=-DNO_LCHMOD" >> flags'
```
uberhacker, line 32: Would this work also?
```
system "CFLAGS=-DNO_LCHMOD make -f unix/Makefile generic"
```

satmandu: That doesn't work. :( The unzip Makefile invokes the CFLAGS from a flags file. It's weird and utterly nonstandard.

uberhacker:
I was able to compile this successfully recently so if you can generate the binaries, this can be merged.

satmandu:
@uberhacker done.

PR #4744, opened 2020-12-13
Delete parsec.rb

supechicken:
This package only works on the clean debian chroot I tested last time


uberhacker:
More dependencies needed?

zhuoyang:
Hi, sorry to revive this old thread,
it seems that you tried to download rpi build for x86 architecture, maybe that's the reason why it doesn't work?
on the side note, rpi build only works with "armhf" arch, generic arm wouldn't work.

packages/parsec.rb
```diff
@@ -1,38 +0,0 @@
...
-    depends_on 'sommelier'
-
-  def self.build
-    case ARCH
-    when 'aarch64', 'armv7l'
-    	system 'curl https://builds.parsecgaming.com/package/parsec-linux.deb -#o parsec.deb'
- 	    puts "Archive downloaded".lightgreen
-   	  abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('parsec.deb') ) == '0c7207f419377fc35e30448acd00d8701300ac5a44936ac6d87a468c9afccdc1'
-    when 'x86_64', 'i686'
-      system 'curl https://builds.parsecgaming.com/package/parsec-rpi.deb -#o parsec.deb'
```
zhuoyang, line 25: this should be using parsec-linux.deb instead?

PR #4754, opened 2020-12-14
[unzip Add missing dependency

### LINKED ISSUE #4745: [unzip] Failed to build
```
cc -o unzip  -Lbzip2 unzip.o crc32.o  crypt.o envargs.o explode.o extract.o fileio.o globals.o inflate.o list.o match.o process.o ttyio.o ubz2err.o unreduce.o unshrink.o zipinfo.o unix.o  -s
/usr/local/lib64/gcc/x86_64-cros-linux-gnu/10.2.0/../../../../x86_64-cros-linux-gnu/bin/ld: unix.o: in function `charset_to_intern':
unix.c:(.text+0x20f7): undefined reference to `libiconv_open'
/usr/local/lib64/gcc/x86_64-cros-linux-gnu/10.2.0/../../../../x86_64-cros-linux-gnu/bin/ld: unix.c:(.text+0x218d): undefined reference to `libiconv'
/usr/local/lib64/gcc/x86_64-cros-linux-gnu/10.2.0/../../../../x86_64-cros-linux-gnu/bin/ld: unix.c:(.text+0x21c8): undefined reference to `libiconv_close'
collect2: error: ld returned 1 exit status
make[1]: *** [unix/Makefile:283: unzip] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/usr/local/tmp/crew/unzip60.tar.gz.dir/unzip60'
make: *** [unix/Makefile:544: generic] Error 2
unzip failed to build: `make V=0 -j2 -f unix/Makefile generic` exited with 2
```

cstrouse:
Unzip requires libiconv to build from source. Closes #4745.

Tested on x86_64.

### DIFF
--- a/packages/unzip.rb
+++ b/packages/unzip.rb
@@ -3,25 +3,18 @@ require 'package'
 class Unzip < Package
   description 'UnZip is an extraction utility for archives compressed in .zip format (also called \'zipfiles\').'
   homepage 'http://www.info-zip.org/UnZip.html'
-  version '1.6_1-1'
+  version '1.6_1-2'
   compatibility 'all'
   source_url 'https://downloads.sourceforge.net/project/infozip/UnZip%206.x%20%28latest%29/UnZip%206.0/unzip60.tar.gz'
   source_sha256 '036d96991646d0449ed0aa952e4fbe21b476ce994abc276e49d30e686708bd37'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/unzip-1.6_1-1-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/unzip-1.6_1-1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/unzip-1.6_1-1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/unzip-1.6_1-1-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-    aarch64: 'e72843178559c924341a969fe9560b56e00b4dd539085e853f0de9602240970a',
-     armv7l: 'e72843178559c924341a969fe9560b56e00b4dd539085e853f0de9602240970a',
-       i686: 'c166df14f26735426dbbbd9bfddcfbf1940057d4b3ef2713adc75f4225312af1',
-     x86_64: 'b3ef4ace8bc32020363d00183eaf19ec5f8358180292216558927f937da1daea',
   })
 
   depends_on 'compressdoc' => :build
+  depends_on 'libiconv'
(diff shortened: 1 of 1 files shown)


packages/unzip.rb
```diff
@@ -3,25 +3,18 @@
...
   })
   binary_sha256 ({
-    aarch64: 'e72843178559c924341a969fe9560b56e00b4dd539085e853f0de9602240970a',
-     armv7l: 'e72843178559c924341a969fe9560b56e00b4dd539085e853f0de9602240970a',
-       i686: 'c166df14f26735426dbbbd9bfddcfbf1940057d4b3ef2713adc75f4225312af1',
-     x86_64: 'b3ef4ace8bc32020363d00183eaf19ec5f8358180292216558927f937da1daea',
   })
 
   depends_on 'compressdoc' => :build
+  depends_on 'libiconv'
```
uberhacker, line 17: This is already a core dependency.  Here is my take.  There is a file conflict between glibc and libiconv.  If you install libiconv after glibc, then the libiconv version is used.  The problem is on a fresh install, libiconv is installed first.  I think what we need to do is remove the conflicting library file from glibc.

satmandu: 100%. I suggest making the iconv.h in libiconv primary and have the glibc remove iconv.h during the installation stage.

satmandu: I was backwards here. glibc in linux INCLUDES the iconv functions, so linux distributions use the iconv.h from glibc. So we need to REMOVE the iconv.h from libiconv. 

Here's my suggestion to fix this issue without requiring the adding of libiconv dependencies, and yes this will require a package rebuild to remove the offending ```iconv.h```:
```
require 'package'

class Libiconv < Package
  description 'GNU charset conversion library for libc which does not implement it.'
  homepage 'https://www.gnu.org/software/libiconv/'
  version '1.16-3'
  compatibility 'all'
  source_url 'https://ftpmirror.gnu.org/libiconv/libiconv-1.16.tar.gz'
  source_sha256 'e6a1b1b589654277ee790cce3734f07876ac4ccfaecbee8afa0b649cf529cc04'


  def self.build
    system "./configure #{CREW_OPTIONS} --enable-static --enable-relocatable --enable-extra-encodings"
    system 'make'
  end

  def self.install
    # Remove iconv.h which conflicts with same file from glibc
    FileUtils.rm "#{CREW_DEST_PREFIX}/include/iconv.h"
  end
end
```

satmandu: (Of course, install that, and then reinstall glibc, and then compiles will stop complaining about libiconv.)

satmandu:
This should be resolved by https://github.com/skycocker/chromebrew/pull/4760

satmandu:
Also https://github.com/skycocker/chromebrew/pull/4740 compiles just fine without libiconv.

```
chronos@localhost /usr/local/lib/crew/packages $ libs `which unzip`
libc.so.6       /usr/local/lib64/libc.so.6
```

PR #4757, opened 2020-12-15
Update graphicsmagick from 1.3.30 to 1.3.35

supechicken:
Tested on `x86_64`

### DIFF
--- a/packages/graphicsmagick.rb
+++ b/packages/graphicsmagick.rb
@@ -17,9 +17,9 @@ class Graphicsmagick < Package
 
   def self.build
     system './configure',
-           "--prefix=/usr/local",
-           "--libdir=/usr/local/lib64",
-           "--mandir=/usr/local/share/man",
+           "--prefix=#{CREW_PREFIX}",
+           "--libdir=#{CREW_PREFIX}/lib64",
+           "--mandir=#{CREW_PREFIX}/share/man",
            "--with-windows-font-dir=#{CREW_PREFIX}/share/fonts/truetype/msttcorefonts",
            "--with-perl=#{CREW_PREFIX}/bin/perl",
            '--disable-maintainer-mode',
--- a/packages/graphicsmagick.rb
+++ b/packages/graphicsmagick.rb
@@ -3,23 +3,10 @@ require 'package'
 class Graphicsmagick < Package
   description 'GraphicsMagick is the swiss army knife of image processing.'
   homepage 'http://www.graphicsmagick.org/'
-  version '1.3.30'
+  version '1.3.35'
   compatibility 'all'
-  source_url 'https://prdownloads.sourceforge.net/project/graphicsmagick/graphicsmagick/1.3.30/GraphicsMagick-1.3.30.tar.xz'
-  source_sha256 'd965e5c6559f55eec76c20231c095d4ae682ea0cbdd8453249ae8771405659f1'
-
-  binary_url ({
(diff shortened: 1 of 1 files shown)


packages/graphicsmagick.rb
```diff
@@ -3,23 +3,10 @@
...
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/graphicsmagick-1.3.30-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/graphicsmagick-1.3.30-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: 'c2a653efa6a3ee47ef4cf50fc10965dd16e3acdee0baa3668ef2899d147cf6e4',
-     armv7l: 'c2a653efa6a3ee47ef4cf50fc10965dd16e3acdee0baa3668ef2899d147cf6e4',
-       i686: '5b44100564bf0e4910c9559cc4eca9a5ed3dea91ce3f47dfceb2bf885aa2d6c2',
-     x86_64: '8554b70d6ba8428089a26b61d1534874691d55fda6fedc6d69e79af53c3c7526',
-  })
+  source_url 'http://ftp.icm.edu.pl/pub/unix/graphics/GraphicsMagick/1.3/GraphicsMagick-1.3.35.tar.gz'
```
uberhacker, line 8: What is wrong with `https://downloads.sourceforge.net/project/graphicsmagick/graphicsmagick/1.3.35/GraphicsMagick-1.3.35.tar.xz`?

supechicken: `curl` will cause a redirecting loop when auto-selecting the download mirror in `sourceforge`

packages/graphicsmagick.rb
```diff
@@ -30,8 +17,9 @@ class Graphicsmagick < Package
 
   def self.build
     system './configure',
-           "--prefix=#{CREW_PREFIX}",
-           "--libdir=#{CREW_LIB_PREFIX}",
+           "--prefix=/usr/local",
+           "--libdir=/usr/local/lib64",
+           "--mandir=/usr/local/share/man",
```
uberhacker: We should be abstracting the install directories.  Does the following work?
```
system "./configure #{CREW_OPTIONS} \
--with-windows-font-dir=#{CREW_PREFIX}/share/fonts/truetype/msttcorefonts \
--with-perl=#{CREW_PREFIX}/bin/perl --disable-maintainer-mode"
```

packages/graphicsmagick.rb
uberhacker: I didn't notice that when I tried.

packages/graphicsmagick.rb
```diff
@@ -32,6 +19,7 @@ def self.build
     system './configure',
            "--prefix=#{CREW_PREFIX}",
            "--libdir=#{CREW_LIB_PREFIX}",
+           "--mandir=#{CREW_PREFIX}/share/man",
```
uberhacker, line 22: This line could be: `--mandir=#{CREW_MAN_PREFIX}` but you could probably replace all 3 lines above with `#{CREW_OPTIONS}`.

packages/graphicsmagick.rb
supechicken: Done

packages/graphicsmagick.rb
supechicken: Just checked, only using `https://sourceforge.net/projects/graphicsmagick/files/graphicsmagick/1.3.35/GraphicsMagick-1.3.35.tar.xz/download` will cause this issue

PR #4759, opened 2020-12-15
Update Opera

supechicken:
Now Opera will fetch the newest version number automatically during install:)
- [x] Download checksum from the repo instead of hardcode
- [x] Fetch the newest version number from the index of repo during install
Tested on `x86_64`

### DIFF
--- a/packages/opera.rb
+++ b/packages/opera.rb
@@ -3,7 +3,7 @@ require 'package'
 class Opera < Package
   description "Opera isn't just a browser. It's an awesome browser."
   homepage 'https://www.opera.com/'
-  version `curl -Ls https://get.geo.opera.com/pub/opera/desktop/ | grep -r 'a' index.html | tail -1 | cut -d '"' -f 2 | sed -e 's|\/||' | perl -e 'while(<>) { $_ =~ s/[\r\n]//g; print "$_" }'`
+  version `curl -Ls https://get.geo.opera.com/pub/opera/desktop/ | grep 'a' | tail -1 | cut -d '"' -f 2 | sed -e 's|\/||' | perl -e 'while(<>) { $_ =~ s/[\r\n]//g; print "$_" }'`
   compatibility 'x86_64'
   case ARCH
   when 'x86_64'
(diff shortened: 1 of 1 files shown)


packages/opera.rb
```diff
@@ -3,7 +3,7 @@
 class Opera < Package
   description "Opera isn't just a browser. It's an awesome browser."
   homepage 'https://www.opera.com/'
-  version '72.0.3815.400'
+  version `curl -Ls https://get.geo.opera.com/pub/opera/desktop/ | grep -r 'a' index.html | tail -1 | cut -d '"' -f 2 | sed -e 's|\/||' | perl -e 'while(<>) { $_ =~ s/[\r\n]//g; print "$_" }'`
```
uberhacker, line 6: Is this idea here to make sure we always get the latest version?  I'm not sure I think this is a good idea.

supechicken: I will test this package to see if it works when every new version of Opera release:)

packages/opera.rb
```diff
@@ -5,32 +5,30 @@ class Opera < Package
...
-  end
+  source_url 'file:///dev/null'
+  source_sha256 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
+  
+  depends_on 'alien' => :build
+  depends_on 'gtk3'
+  depends_on 'gsettings_desktop_schemas'
+  depends_on 'harfbuzz'
+  depends_on 'graphite'
+  depends_on 'sommelier'
```
uberhacker, line 16: We need the `case...end` block since this package can only be installed on x86_64.  Otherwise, all the dependencies are installed for incompatible architectures.

supechicken: Done

packages/opera.rb
uberhacker: The more I think about this, the more I think it's a bad idea.  The idea is good but it probably would be better suited for a revision update script.  Thoughts?

supechicken: How about asking for the user at install (two options, the tested stable and latest stable)?

PR #4763, opened 2020-12-16
Add crew cache option

satmandu:
When developing packages with chromebrew, archive and source packages get downloaded over and over again, wasting bandwidth.

This modification lets you set a ```CREW_CACHE_OPT``` environment variable, which activates logic to both look for and save archives (package and source) during installs.

Not setting the variable leads to current behavior.

Works properly:
- [x] x86_64

### DIFF
--- a/crew
+++ b/crew
@@ -524,7 +524,7 @@ def download
 
     unless ENV["CREW_CACHE_OPT"].to_s == ''
       puts "Caching downloaded archive".lightgreen
-      FileUtils.cp "#{filename}", "#{CREW_CACHE_DIR}" 
+      FileUtils.cp filename, CREW_CACHE_DIR
     end
   end
   return {source: source, filename: filename}
--- a/crew
+++ b/crew
@@ -497,15 +497,36 @@ def download
     sha256sum = @pkg.binary_sha256[@device[:architecture]]
   end
   Dir.chdir CREW_BREW_DIR do
+    unless ENV["CREW_CACHE_OPT"].to_s == ''
+      unless File.directory?(CREW_CACHE_DIR)
+        FileUtils.mkdir_p(CREW_CACHE_DIR) 
+      end
+      if File.file?("#{CREW_CACHE_DIR}#{filename}")
+        if Digest::SHA256.hexdigest( File.read("#{CREW_CACHE_DIR}#{filename}") ) == sha256sum then
+          FileUtils.cp "#{CREW_CACHE_DIR}#{filename}", "#{CREW_BREW_DIR}"
+          puts "Archive found in cache".lightgreen
+          return {source: source, filename: filename}
+          else
+          puts 'Cached archive checksum mismatch. :/ Will download.'.lightred
+        end
+      end
+    end
+
     if @opt_verbose then
       system('curl', '-v', '--retry', '3', '--progress-bar', '-C', '-', '--insecure', '-L', '-#', url, '-o', filename)
     else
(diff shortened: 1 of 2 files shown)


crew
```diff
@@ -497,15 +497,36 @@ def download
...
       system('curl', '--retry', '3', '--progress-bar', '-C', '-', '--insecure', '-L', '-#', url, '-o', filename)
     end
+
     abort 'Checksum mismatch. :/ Try again.'.lightred unless
       Digest::SHA256.hexdigest( File.read("./#{filename}") ) == sha256sum
+    puts "Archive downloaded".lightgreen
+
+    unless ENV["CREW_CACHE_OPT"].to_s == ''
+      puts "Caching downloaded archive".lightgreen
+      FileUtils.cp "#{filename}", "#{CREW_CACHE_DIR}"
```
uberhacker, line 527: This line could be simply:
```
FileUtils.cp filename, CREW_CACHE_DIR
```
No need to enclose in quotes.

uberhacker:
This is a good idea @satmandu.  However, should we also have the ability to delete the cache?  I can see where this could fill up the disk after awhile.

satmandu:
@uberhacker This is why I made it an undocumented non-default option. There's no option to deal with filling up the disk from having too many packages, so I'm not sure this is the best place to start having logic about disk space usage?

satmandu:
(Since it is a hidden option, we could add more hidden options to clear out the folder, but I wonder if that code would just bitrot.)

uberhacker:
Well, you can always remove projects with `crew remove`.  There is no command to clear this cache.  I suppose you could add to clear_cache.rb?

satmandu:
Ah! I could just drop the cache folder into ```~/.cache/crew``` - that way```clear-cache``` wouldn't have to be changed.

crew
satmandu: done

satmandu:
@uberhacker modified the cache folder to be under home, and now also allowing CREW_CACHE_DIR to be pulled as an env variable too like CREW_PREFIX.

lib/const.rb
```diff
@@ -34,6 +34,14 @@
   HOME = CREW_PREFIX + ENV['HOME']
 end
 
+if ENV['CREW_CACHE_DIR'].to_s == ''
+  CREW_CACHE_DIR = HOME + '.cache/crewcache'
```
uberhacker, line 37: Do we need `'/.cache/crewcache'`?

lib/const.rb
```diff
@@ -34,6 +34,14 @@
...
+else
+  CREW_CACHE_DIR = ENV['CREW_CACHE_DIR']
+  @pkg.build_from_source = true
```
uberhacker, line 40: Why is this line necessary?

lib/const.rb
satmandu: ~/.cache seems like it could be used by multiple programs, so let's not clobber?

lib/const.rb
satmandu: oops. good catch.

satmandu: copypasta fixed

lib/const.rb
satmandu: Do we need a ```/``` before ```./cache```?

uberhacker: Yeah, that was my question.

satmandu: Thanks for getting me there. :)  Fixed.

PR #4768, opened 2020-12-17
zstd => 1.4.7

satmandu:
New upstream release.

Works properly:
- [x] x86_64

Yes, this will need binaries. :/


uberhacker:
LGTM but we need binaries since this is a core package.

PR #4771, opened 2020-12-19
[WIP] Remove sommelier dependency for thunar and its dependencies

supechicken:
Thunar works well with system default `wayland-0` server without `sommelier` and `xwayland`

Tested on `x86_64`

### DIFF
--- a/packages/thunar.rb
+++ b/packages/thunar.rb
@@ -39,7 +39,7 @@ class Thunar < Package
 
   def self.install
     system "make install DESTDIR=#{CREW_DEST_DIR}"
-    system "cat <<EOF> #{CREW_DEST_PREFIX}/bin/Thunar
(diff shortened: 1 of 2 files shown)


packages/thunar.rb
```diff
@@ -29,7 +29,8 @@ class Thunar < Package
   depends_on 'desktop_file_utilities'
   depends_on 'hicolor_icon_theme'
   depends_on 'xdg_base'
-  depends_on 'sommelier'
+  depends_on 'wayland'
+  depends_on 'wayland_protocols'
```
uberhacker, line 33: Should only need wayland_protocols since that already has the wayland dependency.

uberhacker:
Thank you for the update @supechicken.  Nice to know sommelier isn't needed!  This would definitely help for i686 architecture.

packages/thunar.rb
supechicken: Done

PR #4772, opened 2020-12-19
xfce4_panel need a window manager to use with it

supechicken:


packages/xfce4_panel.rb
```diff
@@ -36,4 +36,8 @@ def self.build
   def self.install
     system "make install DESTDIR=#{CREW_DEST_DIR}"
   end
+  
+  def self.postinstall
+    puts
+    puts "You need a window manager like 'openbox' to use with this package.".lightblue
```
uberhacker, line 42: I noticed openbox doesn't work with i686 but weston with xwayland does.  Maybe we should suggest that instead?  The sharpness of weston is really impressive.

supechicken: Done

PR #4773, opened 2020-12-19
Remove sommelier dependency for xfce4-terminal

supechicken:
Part of #4771

Tested on `x86_64`

### DIFF
--- a/packages/xfce4_terminal.rb
+++ b/packages/xfce4_terminal.rb
@@ -27,7 +27,8 @@ class Xfce4_terminal < Package
   depends_on 'hicolor_icon_theme'
   depends_on 'libxfce4ui'
   depends_on 'startup_notification'
-  depends_on 'sommelier'
+  depends_on 'wayland_protocols'
+  depends_on 'mesa'
 
   def self.build
     system "./configure #{CREW_OPTIONS}"
@@ -36,5 +37,13 @@ class Xfce4_terminal < Package
 
   def self.install
     system "make install DESTDIR=#{CREW_DEST_DIR}"
+    system "cat <<'EOF'> xfce4-terminal
+WAYLAND_DISPLAY=wayland-0
+GDK_BACKEND=wayland
+DISPLAY=
+#{CREW_PREFIX}/bin/xfce4-terminal $@
+EOF"
+    FileUtils.mv "#{CREW_DEST_PREFIX}/bin/xfce4-terminal", "#{CREW_DEST_PREFIX}/bin/xfce4_terminal"
+    system "install -Dm755 xfce4-terminal #{CREW_DEST_PREFIX}/bin/xfce4-terminal"
   end
 end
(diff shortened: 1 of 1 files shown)


packages/xfce4_terminal.rb
```diff
@@ -36,5 +37,13 @@ def self.build
...
   def self.install
     system "make install DESTDIR=#{CREW_DEST_DIR}"
+    system "cat <<'EOF'> xfce4-terminal
+WAYLAND_DISPLAY=wayland-0
+GDK_BACKEND=wayland
+DISPLAY=
+#{CREW_PREFIX}/bin/xfce4-terminal $@
+EOF"
+    FileUtils.mv "#{CREW_DEST_PREFIX}/bin/xfce4-terminal", "#{CREW_DEST_PREFIX}/bin/xfce4_terminal"
+    system "install -Dm755 xfce4-terminal #{CREW_DEST_PREFIX}/bin/xfce4-terminal"
```
uberhacker, line 47: So we end up with `xfce4_terminal` and `xfce4-terminal`?  What is the difference?

packages/xfce4_terminal.rb
```diff
@@ -36,5 +37,13 @@ def self.build
 
   def self.install
     system "make install DESTDIR=#{CREW_DEST_DIR}"
+    system "cat <<'EOF'> xfce4-terminal
+WAYLAND_DISPLAY=wayland-0
+GDK_BACKEND=wayland
+DISPLAY=
+#{CREW_PREFIX}/bin/xfce4-terminal $@
```
uberhacker, line 44: `#{CREW_PREFIX}/bin/xfce4_terminal`?  (Replace `-` with `_`)

packages/xfce4_terminal.rb
supechicken: The new `xfce4-terminal` script will set the Wayland display variable at start

packages/xfce4_terminal.rb
supechicken: Done

packages/xfce4_terminal.rb
```diff
@@ -36,5 +37,13 @@ def self.build
 
   def self.install
     system "make install DESTDIR=#{CREW_DEST_DIR}"
+    system "cat <<'EOF'> xfce4-terminal
+WAYLAND_DISPLAY=wayland-0
```
uberhacker, line 41: Probably should add `#!/bin/bash` at the top.

supechicken: Done

packages/xfce4_terminal.rb
```diff
@@ -36,5 +37,13 @@ def self.build
...
+GDK_BACKEND=wayland
+DISPLAY=
+#{CREW_PREFIX}/bin/xfce4_terminal $@
```
uberhacker, line 45: Probably should put quotes around `\"$@\"`.

supechicken: No need, as `'EOF'` will escape all meaning of syntax automatically

uberhacker: I'm talking about putting quotes around the arguments, not escaping the characters.

supechicken: Done

PR #4774, opened 2020-12-19
Crewfix update

satmandu:
-for the situation where curl's underlying chromebrew libraries are broken.
-for the situation where on armv8l host.
-part of this will be part of the non-libssp gcc package I'm testing.

Works properly:
- [x] x86_64
- [x] armv7l/armv8l

### DIFF
--- a/packages/gcc_tools.rb
+++ b/packages/gcc_tools.rb
@@ -18,6 +18,7 @@ set -e
 # define constants
 CREW_PREFIX="${CREW_PREFIX:-/usr/local}"
 ARCH_ACTUAL="$(uname -m)"
+[[ "$ARCH_ACTUAL" == "armv8l" ]] && ARCH_ACTUAL="armv7l"
 ARCH="${ARCH:-$ARCH_ACTUAL}"
 LIB_SUFFIX=""
 [ "${ARCH}" == "x86_64" ] && LIB_SUFFIX="64"
--- a/packages/gcc_tools.rb
+++ b/packages/gcc_tools.rb
@@ -17,12 +17,14 @@ set -e
 
 # define constants
 CREW_PREFIX="${CREW_PREFIX:-/usr/local}"
-ARCH_ACTUAL="$(uname -m)"
-[[ "$ARCH_ACTUAL" == "armv8l" ]] && ARCH_ACTUAL="armv7l"
-ARCH="${ARCH:-$ARCH_ACTUAL}"
+ARCH="$(uname -m)"
 LIB_SUFFIX=""
 [ "${ARCH}" == "x86_64" ] && LIB_SUFFIX="64"
-CURL="LD_LIBRARY_PATH=/usr/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} /usr/bin/curl"
+if curl --version &>/dev/null; then
+  CURL=curl
+else
+  CURL="LD_LIBRARY_PATH=/usr/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} /usr/bin/curl"
+fi
 
 function download_check () {
   cd /tmp
@@ -45,7 +47,7 @@ function download_check () {
 if [ ! -f "${CREW_PREFIX}/lib${LIB_SUFFIX}/libssp.so.0" ]; then
   # prepare gcc8 url and sha256
   case "${ARCH}" in
-  "aarch64"|"armv7l")
+  "aarch64"|"armv7l"|"armv8l")
     url="https://dl.bintray.com/chromebrew/chromebrew/gcc8-8.3.0-chromeos-armv7l.tar.xz"
     sha256="fbd8a589befb3d10400af6e4975d02a6940bab4907628f8fc0d6913ea89f70ae"
     ;;
@@ -68,7 +70,7 @@ fi
 if [ ! -f "${CREW_PREFIX}/bin/ruby" ]; then
   # prepare ruby url and sha256
   case "${ARCH}" in
-  "aarch64"|"armv7l")
+  "aarch64"|"armv7l"|"armv8l")
     url="https://dl.bintray.com/chromebrew/chromebrew/ruby-2.7.2-chromeos-armv7l.tar.xz"
     sha256="a435e6bf7965e1a82e8842e5ea66bdd670ec9b627d785bd720d3d2652fc89f6d"
     ;;
--- a/packages/gcc_tools.rb
+++ b/packages/gcc_tools.rb
@@ -3,7 +3,7 @@ require 'package'
 class Gcc_tools < Package
   description 'Tools for working with gcc packages'
   homepage 'https://github.com/skycocker/chromebrew'
-  version '1.1'
+  version '1.2'
   compatibility 'all'
   source_url 'file:///dev/null'
   source_sha256 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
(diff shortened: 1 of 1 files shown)


packages/gcc_tools.rb
```diff
@@ -17,16 +17,19 @@ def self.build
 
 # define constants
 CREW_PREFIX="${CREW_PREFIX:-/usr/local}"
-ARCH="$(uname -m)"
+ARCH_ACTUAL="$(uname -m)"
+[[ "$ARCH_ACTUAL" == "armv8l" ]] && ARCH_ACTUAL="armv7l"
```
uberhacker, line 21: Does this only happen with Docker images?  Most Chrome(books/boxes) (all?) return `aarch64` from the `arch/uname -m` command.

packages/gcc_tools.rb
```diff
@@ -17,16 +17,19 @@ def self.build
...
+ARCH="${ARCH:-$ARCH_ACTUAL}"
 LIB_SUFFIX=""
 [ "${ARCH}" == "x86_64" ] && LIB_SUFFIX="64"
+CURL="LD_LIBRARY_PATH=/usr/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} /usr/bin/curl"
```
uberhacker, line 25: Is the idea here to bypass the Chromebrew curl?

packages/gcc_tools.rb
satmandu: armv8l is what shows up when you run ```linux32 uname -m``` on the raspberry pi. 

Otherwise you get aarch64.

The arm64 chromebooks give aarch64 from ```uname -m``` then?

And do they put libs in /usr/lib or /usr/lib64?

uberhacker: The only architecture that has lib64 is x86_64.  All arm64/armv8/aarch64 architectures still use a 32-bit userland compatible with armv7l.

packages/gcc_tools.rb
satmandu: Yes but now that I think of it, it might be best to hide this behind a conditional checking to see if the chromebrew curl is erroring out.

uberhacker: I went through a lot of blood, sweat and tears to make sure the Chromebrew curl doesn't error out. :)

satmandu: Yes but what about when I muck about with gcc and suddenly curl can't find libraries?

(This literally happened to me yesterday.)

uberhacker: You are not supposed to be mucking around with gcc but testing on fresh installs. :)

satmandu: mucking around meaning testing gcc builds built on fresh installs on my own install to see if anything breaks.

Have to dogfood sometimes!

satmandu: Made changes to the curl logic to see if curl works, and took out the ARCH complications. Is this more bearable?

packages/gcc_tools.rb
satmandu: Noted.

satmandu: took out this logic for the most part...

packages/gcc_tools.rb
```diff
@@ -17,12 +17,14 @@ def self.build
...
-ARCH="${ARCH:-$ARCH_ACTUAL}"
+ARCH="$(uname -m)"
 LIB_SUFFIX=""
 [ "${ARCH}" == "x86_64" ] && LIB_SUFFIX="64"
-CURL="LD_LIBRARY_PATH=/usr/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} /usr/bin/curl"
+if curl --version &>/dev/null; then
+  CURL=curl
+else
+  CURL="LD_LIBRARY_PATH=/usr/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} /usr/bin/curl"
+fi
```
uberhacker, line 27: curl should always work so I wonder if this is just a Docker thing.

satmandu: There are definitely ways to break it. Try messing with glibc. (This is also a good way to deal with glibc breakage.)

satmandu: Speaking of which... this was useful when I broke glibc:
```
require 'package'

class Coreutils_static < Package
  description 'GNU coreutils built statically with musl'
  homepage 'http://www.gnu.org/software/coreutils/coreutils.html'
  version '8.32'
  compatibility 'all'
  source_url 'https://mirrors.sarata.com/gnu/coreutils/coreutils-8.32.tar.xz'
  source_sha256 '4458d8de7849df44ccab15e16b1548b285224dbba5f08fac070c1c0e0bcc4cfa'


  def self.build
  # static build script from https://github.com/luciusmagn/coreutils-static/blob/master/build.sh
system 'cat <<"EOF"> build.sh
#!/bin/bash
#
# build static coreutils because we need exercises in minimalism
# MIT licensed: google it or see robxu9.mit-license.org.
#
# For Linux, also builds musl for truly static linking.

coreutils_version="8.32"
musl_version="1.2.1"

platform=$(uname -s)

if [ -d build ]; then
  echo "= removing previous build directory"
  rm -rf build
fi

#mkdir build # make build directory
#pushd build

# download tarballs
#echo "= downloading coreutils"
#curl -LO http://ftp.gnu.org/gnu/coreutils/coreutils-${coreutils_version}.tar.xz

#echo "= extracting coreutils"
#tar xJf coreutils-${coreutils_version}.tar.xz

if [ "$platform" = "Linux" ]; then
  echo "= downloading musl"
  curl -LO http://www.musl-libc.org/releases/musl-${musl_version}.tar.gz

  echo "= extracting musl"
  tar -xf musl-${musl_version}.tar.gz

  echo "= building musl"
  working_dir=$(pwd)

  install_dir=${working_dir}/musl-install

  pushd musl-${musl_version}
  env CFLAGS="$CFLAGS -Os -ffunction-sections -fdata-sections -fPIC" LDFLAGS=\'-Wl,--gc-sections\' ./configure --prefix=${install_dir}
  make install
  popd # musl-${musl-version}

  echo "= setting CC to musl-gcc"
  export CC=${working_dir}/musl-install/bin/musl-gcc
  export CFLAGS="-static"
else
  echo "= WARNING: your platform does not support static binaries."
  echo "= (This is mainly due to non-static libc availability.)"
fi

echo "= building coreutils"

#pushd coreutils-${coreutils_version}
env FORCE_UNSAFE_CONFIGURE=1 CFLAGS="$CFLAGS -Os -ffunction-sections -fdata-sections -fPIC" LDFLAGS=\'-Wl,--gc-sections\' ./configure
make
#popd # coreutils-${coreutils_version}

popd # build

if [ ! -d releases ]; then
  mkdir releases
fi

echo "= striptease"
strip -s -R .comment -R .gnu.version --strip-unneeded build/coreutils-${coreutils_version}/coreutils
#echo "= compressing"

shopt -s extglob
# upx currently breaks the binaries
#for file in build/coreutils-${coreutils_version}/src/!(*.*)
#do
#	upx --ultra-brute $file
#done
echo "= extracting coreutils binary"
cp src/!(*.*) releases
echo "= done"
EOF'
    system 'chmod +x ./build.sh'
    system './build.sh'
    system 'curl -Lf https://salsa.debian.org/debian/debianutils/-/raw/master/which?inline=false > releases/which && chmod +x releases/which'
    system "sed -i 's#/bin/sh#/usr/local/bin/sh#g' releases/which"

  end

  def self.install
    FileUtils.cd("#{CREW_DEST_PREFIX}/bin") do
      system "echo '#!/bin/bash' > arch"
      system "echo '#{CREW_PREFIX}/bin/uname -m' >> arch"
      system 'chmod +x arch'
    end
    system "cp releases/* #{CREW_DEST_PREFIX}/bin/"
  end
end
```

satmandu: Also happy to just dump changes into a ```crewfix2``` as an alternate crewfix, or just create a separate package of utils...

PR #4776, opened 2020-12-19
Pygments: get rid of that annoying install message

satmandu:
Fixes Install Warning as per method in warning:
```
Collecting pygments
  Downloading Pygments-2.7.3-py3-none-any.whl (950 kB)
Installing collected packages: pygments
  WARNING: The script pygmentize is installed in '/usr/local/tmp/crew/dest/usr/local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
Successfully installed pygments-2.7.3
```
A version bump isn't needed here.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/pygments.rb
+++ b/packages/pygments.rb
@@ -22,6 +22,6 @@ class Pygments < Package
   })
 
   def self.install
-    system "pip3 install pygments --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR}"
+    system "pip3 install pygments --no-warn-script-location --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR}"
   end
 end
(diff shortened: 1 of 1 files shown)


packages/pygments.rb
```diff
@@ -22,6 +22,6 @@ class Pygments < Package
   })
 
   def self.install
-    system "pip3 install pygments --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR}"
+    system "pip3 install pygments --no-warn-script-location --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR}"
```
uberhacker, line 25: Remove binaries to replace with new version?

satmandu: Any reason not to get rid of the binaries entirely and just put in 
```
  source_url 'file:///dev/null'
```
(because what of the source are we really using?)

uberhacker: Yeah, no reason to download the source if it isn't being used.  It will just take longer to install without binaries.

PR #4778, opened 2020-12-19
Python 3.9.1

satmandu:
- Also using get-pip to upgrade pip, as this was the only way I was able to do this without the install throwing an error.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -3,23 +3,11 @@ require 'package'
 class Python3 < Package
   description 'Python is a programming language that lets you work quickly and integrate systems more effectively.'
   homepage 'https://www.python.org/'
-  version '3.9.0'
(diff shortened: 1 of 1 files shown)


uberhacker:
Not sure what error you are getting but upgrading pip is not a problem for a fresh install on all architectures.  Docker related maybe?

satmandu:
> Not sure what error you are getting but upgrading pip is not a problem for a fresh install on all architectures. Docker related maybe?

The error I was getting (on the local machine) was that the install failed in the postinstall stage.

Maybe the upgrade of pip is failing going from 3.9.0 to 3.9.1? ```pip3 install --upgrade pip```  threw the error.

uberhacker:
Updated with pre-built binaries in #4789.

PR #4793, opened 2020-12-21
New package: systemd

supechicken:
The systemd System and Service Manager

Tested on `x86_64`

### DIFF
--- a/packages/systemd.rb
+++ b/packages/systemd.rb
@@ -1,0 +1,30 @@
+require 'package'
+
+class Systemd < Package
+  description 'The systemd System and Service Manager'
+  homepage 'https://systemd.io/'
+  version '247'
+  compatibility 'all'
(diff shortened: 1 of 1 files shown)


satmandu:
Oh wow.

satmandu:
Wait... How are you running/starting systemd? There should probably be documentation on how you get this to actually run on ChromeOS since you can't be running systemd as the init process.

uberhacker:
Yeah, this is definitely a bold move.  I was thinking recently it would also be nice to have a boot process manager and this would be a good start in that direction.

supechicken:
> Wait... How are you running/starting systemd? There should probably be documentation on how you get this to actually run on ChromeOS since you can't be running systemd as the init process.

I am not trying to replace the system systemd at this moment :)  (must remount rootfs as rw to do it) (or GRUB?)  The purpose of
this package for me is the `libsystemd` part

PR #4795, opened 2020-12-22
Readline -> 8.1

satmandu:
Compiles/Packages properly:
- [x] x86_64
- [x] i686
- [x] armv7l

I won't be so bold as to say it works splendidly,  but I'm seeing zero issues on x86_64.

### DIFF
--- a/packages/readline.rb
+++ b/packages/readline.rb
@@ -1,4 +1,4 @@
-equire 'package'
+require 'package'
 
 class Readline < Package
   description 'The GNU Readline library provides a set of functions for use by applications that allow users to edit command lines as they are typed in.'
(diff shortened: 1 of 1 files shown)


packages/readline.rb
```diff
@@ -1,32 +1,30 @@
...
 
   def self.build
-    system './configure',
-           "--prefix=#{CREW_PREFIX}",
-           "--libdir=#{CREW_LIB_PREFIX}",
-           '--disable-static',
-           '--with-curses'
+    system "./configure #{CREW_OPTIONS} \
+    --disable-install-examples \
+    --with-curses"
```
uberhacker, line 27: Does this break anything?  There are so many packages that depend on this.

satmandu: Better safe than sorry. I'll take that line out.

uberhacker: Probably a better solution would be to recompile other packages if something breaks.  More work, I know but we don't want to sacrifice capabilities, right?

satmandu: yup. Updated with new uploads.

PR #4810, opened 2020-12-24
Delete symlinks on Cloudready before install

supechicken:
On Cloudready, `var` is a symlink which pointed to `/var` by default
```shell
lrwxrwxrwx  1 root    root        1 Nov 10 11:49 var -> /var
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -33,7 +33,12 @@ case "${ARCH}" in
 esac
 
 # This will allow things to work without sudo
-sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}"
+sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}/*/*"
+# Delete 'var' symlink on Cloudready platform
+if [[ $(cat /etc/lsb-release | grep neverware) != "" ]]; then
+  sudo rm -rf /usr/local/var
+  sudo rm -rf /usr/local/local
+fi
 
 # prepare directories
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -33,7 +33,12 @@ case "${ARCH}" in
 esac
 
 # This will allow things to work without sudo
-sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}"
+sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}/*/*"
+# Delete 'var' symlink on Cloudready platform
+if [[ $(cat /etc/lsb-release | grep neverware) != "" ]]; then
```
uberhacker, line 38: Easier logic?
```
if [[ $(grep -c neverware /etc/lsb-release) > 0 ]]; then
```

install.sh
```diff
@@ -33,7 +33,12 @@ case "${ARCH}" in
 esac
 
 # This will allow things to work without sudo
-sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}"
+sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}/*/*"
```
uberhacker, line 36: Why do we need to change this line?

supechicken: Removed

install.sh

uberhacker:
I'm not sure of the ramifications of doing this but on the surface it looks correct.

PR #4815, opened 2020-12-26
Update wine from 5.0.2-stable to 5.22

uberhacker:
Tested on:
- [x] aarch64
- [x] armv7l
- [x] x86_64


packages/wine.rb
```diff
@@ -4,47 +4,78 @@
...
   end
 
   def self.install
     case ARCH
-    when "x86_64"
+    when 'x86_64'
       FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
-      FileUtils.cd("#{CREW_DEST_PREFIX}/bin") do
-        system "ln -s wine64 wine"
+      FileUtils.cd "#{CREW_DEST_PREFIX}/bin" do
```
supechicken, line 53: Will `Dir.chdir` work?

uberhacker: Yeah, we use that in a lot of packages.

PR #4820, opened 2020-12-28
Add Grep

satmandu:
Was surprised to discover we were using the system app.

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/grep.rb
+++ b/packages/grep.rb
@@ -1,0 +1,39 @@
+
+require 'package'
+
+class Grep < Package
+  description 'A string search utility'
+  homepage 'https://www.gnu.org/software/grep/'
+  version '3.6'
+  compatibility 'all'
+  source_url 'https://ftp.gnu.org/gnu/grep/grep-3.6.tar.xz'
+  source_sha256 '667e15e8afe189e93f9f21a7cd3a7b3f776202f417330b248c2ad4f997d9373e'
+
+  binary_url ({
+     aarch64: 'file:///usr/local/tmp/packages/grep-3.6-chromeos-armv7l.tar.xz',
+      armv7l: 'file:///usr/local/tmp/packages/grep-3.6-chromeos-armv7l.tar.xz',
+        i686: 'file:///usr/local/tmp/packages/grep-3.6-chromeos-i686.tar.xz',
+      x86_64: 'file:///usr/local/tmp/packages/grep-3.6-chromeos-x86_64.tar.xz',
+  })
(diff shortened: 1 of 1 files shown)


packages/grep.rb
```diff
@@ -0,0 +1,39 @@
...
+  })
+  binary_sha256 ({
+     aarch64: '87ebec000c54b032f1e2720b03a30dfb13f14dc3a2a68a4c943daae64dc5ee34',
+      armv7l: '87ebec000c54b032f1e2720b03a30dfb13f14dc3a2a68a4c943daae64dc5ee34',
+        i686: 'a6a62d2c91c346041fc4408727e387e2cc415e304d2ca7d372a97256520d6f64',
+      x86_64: '0106cbf95ada69fc0205b4af06af4e46144156934555b5bd9376e4c00d61092f',
+  })
+
+
+  depends_on 'glibc'
```
uberhacker, line 26: This is a core package so no need to list here.

packages/grep.rb
```diff
@@ -0,0 +1,39 @@
...
+  depends_on 'pcre'
+  depends_on 'texinfo' => ':build'
```
uberhacker, line 28: Core package?

packages/grep.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def self.build
+    ENV['CFLAGS'] = '-flto'
+    system "./configure #{CREW_OPTIONS}\
+    --without-included-regex"
```
uberhacker, line 33: Why without regex?  That is one of main reasons I use grep. :)

satmandu: The rationale is that you don't add it if your libc has it.

packages/grep.rb
satmandu: removed

packages/grep.rb
uberhacker: Fair enough.  I thought for sure this would be included in coreutils but apparently not.

satmandu: from configure:
```
dnl Some installers want to be informed if we do not use our regex.
dnl For example, if the host platform uses dynamic linking and the installer
dnl knows that the grep may be invoked on other hosts with buggy libraries,
dnl then the installer should configure --with-included-regex.
```

satmandu: linuxfromscratch is not using that flag, so I am dropping it.

PR #4832, opened 2020-12-29
filezilla: add libwebp dep

### LINKED ISSUE #4807: [Filezilla] core package and lib package are incompatible
Running `/usr/local/bin/filezilla` sends :
``` 
Reading locale option from /home/chronos/user/.config/filezilla/filezilla.xml
Fatal Error: Mismatch between the program and library build versions detected.
The library used 3.0 (wchar_t,compiler with C++ ABI 1013,wx containers,compatible with 2.8),
and your program used 3.0 (wchar_t,compiler with C++ ABI 1014,wx containers,compatible with 2.8).
Aborted (core dumped)
```

satmandu:
Fixes #4807 

Add libwebp dep.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/filezilla.rb
+++ b/packages/filezilla.rb
@@ -29,8 +29,9 @@ class Filezilla < Package
   depends_on 'sqlite'
   depends_on 'wxwidgets'
   depends_on 'xdg_utils'
-  depends_on 'sommelier'
   depends_on 'libwebp'
+  depends_on 'sommelier'
+  
 
   def self.patch
     system 'filefix'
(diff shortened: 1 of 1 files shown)


packages/filezilla.rb
```diff
@@ -30,6 +30,7 @@ class Filezilla < Package
   depends_on 'wxwidgets'
   depends_on 'xdg_utils'
   depends_on 'sommelier'
+  depends_on 'libwebp'
```
uberhacker, line 33: Can we put this above sommelier so users don't miss the all-important post install message?

satmandu: done

PR #4833, opened 2020-12-29
Fix git sha256 hash in install.sh

uberhacker:
Tested fresh install on i686.


satmandu:
My docker setup still works with this bug. Are the hashes not being checked properly?

PR #4836, opened 2020-12-29
gtk3 - remove null in pkgconfig files

satmandu:
- Compiling with gtk3 is causing issues due to an errant "null" in the gtk3  pkgconfig files. I don't know where those came from, but this gets rid of those nulls in those files.
- Also graphite was needed for me to compile, so I added that dependency.

Binaries packaged for all archs, and confirmed working on x86_64.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/gtk3.rb
+++ b/packages/gtk3.rb
@@ -3,22 +3,22 @@ require 'package'
 class Gtk3 < Package
   description 'GTK+ is a multi-platform toolkit for creating graphical user interfaces.'
   homepage 'https://developer.gnome.org/gtk3/3.0/'
-  version '3.24.24'
+  version '3.24.24-1'
   compatibility 'all'
   source_url 'https://download.gnome.org/sources/gtk+/3.24/gtk+-3.24.24.tar.xz'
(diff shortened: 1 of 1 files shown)


packages/gtk3.rb
```diff
@@ -50,6 +51,7 @@ def self.build
 
   def self.install
     system "DESTDIR=#{CREW_DEST_DIR} ninja -C build install"
+    system "sed -i 's,null,,g'  #{CREW_DEST_LIB_PREFIX}/pkgconfig/gtk*.pc"
```
uberhacker, line 54: Should this be in a patch section?

satmandu: This is patching after install to the destdir, so it needs to happen after install but before package creation. Doesn't self.patch happen before build?

uberhacker: It does but I just wanted to ask if it was possible to do it before the build.

satmandu: That would be ideal. But I have **no idea** where that came from in the build. This fixes subsequent builds, so I'm not spending any more headspace on it.

PR #4837, opened 2020-12-29
ccache Fix non-x86_64 hashes

satmandu:
My initial upload and hash scripts were not generating hashes from the final packages for non-x86_64. So I just rebuilt those ccache packages.

Sorry I didn't catch this earlier.

verified hashes for:
- [x] x86_64
- [x]  i686
- [x] armv7l


uberhacker:
Good catch!

PR #4838, opened 2020-12-29
geany -> 1.37.1

satmandu:
- compiles just fine for i686, so removed i686 exclusionary logic.

Works properly:
- [x] x86_64

Compiles for everything.


uberhacker:
Nice!  Looks like you were successful with the i686 build.

PR #4839, opened 2020-12-29
libepoxy update from 1.4.3 to 1.5.5, use meson

supechicken:
Tested on `x86_64`

### DIFF
--- a/packages/libepoxy.rb
+++ b/packages/libepoxy.rb
@@ -6,7 +6,7 @@ class Libepoxy < Package
   version '1.5.5'
   compatibility 'all'
   source_url 'https://github.com/anholt/libepoxy/releases/download/1.5.5/libepoxy-1.5.5.tar.xz'
-  source_sha256 '0b808a06c9685a62fca34b680abb8bc7fb2fda074478e329b063c1f872b826f6'
+  source_sha256 '261663db21bcc1cc232b07ea683252ee6992982276536924271535875f5b0556'
 
   depends_on 'mesa'
   depends_on 'python3'
--- a/packages/libepoxy.rb
+++ b/packages/libepoxy.rb
@@ -12,12 +12,12 @@ class Libepoxy < Package
   depends_on 'python3'
 
   def self.build
-    Dir.mkdir 'build'
-    system "meson #{CREW_MESON_OPTIONS} build"
-    system "ninja -C build"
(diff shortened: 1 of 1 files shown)


packages/libepoxy.rb
```diff
@@ -3,38 +3,21 @@
...
 
   def self.build
-    system "mkdir _build"
-    Dir.chdir "_build" do
-      system "../configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX}"
-      system "make"
-    end
+    Dir.mkdir 'build'
+    system "meson #{CREW_MESON_OPTIONS} build"
+    system "ninja -C build"
```
satmandu, line 17: - You don't need a mkdir for meson build. meson will create the dir automatically.
- Can you add a ```meson configure build``` line before ninja? That lets us see what the options are during compile, which can be useful.

packages/libepoxy.rb
```diff
@@ -3,38 +3,21 @@
...
   end
 
   def self.install
-    Dir.chdir "_build" do
-      system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
-    end
+      system "meson install -C build"
```
satmandu, line 21: You need the destdir, and you can use ninja.
Do this:
```system "DESTDIR=#{CREW_DEST_DIR} ninja install -C build"```

File: packages/libepoxy.rb (hunk 1 above)
Comment (uberhacker) [reply packages/libepoxy.rb:17]: Sure, if that helps.

File: packages/libepoxy.rb (hunk 2 above)
Comment (uberhacker) [reply packages/libepoxy.rb:21]: Yep, good call.  Might want to check if there is an existing build directory.  Sometimes the `_build` directory is needed because of that situation.

Comment (satmandu) [reply packages/libepoxy.rb:21]: Yup. or builddir.

This is my goto code block for meson built packages, leading to this ending the package file:

```
  def self.build
    system "meson #{CREW_MESON_OPTIONS} builddir"
    system "meson configure builddir"
    system "ninja -C builddir"
  end

  def self.install
  end
end
```

packages/libepoxy.rb
supechicken: Done

packages/libepoxy.rb
supechicken: Done

PR #4841, opened 2020-12-29
amtk: remove clang dependency

satmandu:
- compiles just fine with gcc, with meson lld cflag.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/amtk.rb
+++ b/packages/amtk.rb
@@ -3,22 +3,22 @@ require 'package'
 class Amtk < Package
   description 'Actions, Menus and Toolbars Kit for GTK+ applications'
   homepage 'https://wiki.gnome.org/Projects/Amtk'
-  version '5.2.0-33ec'
+  version '5.2.0-33ec-1'
   compatibility 'all'
   source_url 'https://github.com/GNOME/amtk/archive/33ec171156ebc49d6dad568e6ba39470edb272e1.zip'
   source_sha256 'fc9bad18bbd5421da52e0548f9e91eda180539bb568d1e7116ca65f4f73f4b67'
 
   binary_url ({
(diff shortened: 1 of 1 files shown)


cstrouse:
@satmandu CrOS is built with llvm. What is the rationale for wanting to change packages to use gcc?

uberhacker:
@cstrouse: I think he wants to avoid the use of `ENV` variables.  @satmandu can elaborate more on this.

satmandu:
@cstrouse I have no complaints about llvm (aside from how painful I am finding builting llvm 11  with lto=thin). I still have these packages using llvm's lld linker. But I'm seeing ENV issues where ```['CC'] = 'clang'``` is set and then subsequent package installs have problems because they think the CC environment variable is STILL set to clang, even if I literally have a ```['CC'] = 'gcc'``` section in the second package file.

So I'm just going through and getting rid of all of the packages which have CC set to clang or CXX set to clang++.

If a compile with GCC fails, we can also set ```CC=clang ./configure...``` or whatever.

(But I am also wondering if ENV should be used for setting environment variables at all. Isn't that frowned upon in ruby?)


TL;DR: Not actually avoiding using all llvm components, as we are still often using lld. Just avoiding setting packages to use clang or clang++ through ENV settings where possible to rule out that being an issue with package compile failures.

cstrouse:
That makes sense. I just wanted to make sure we capture the rationale for things in the commit messages or code reviews for reference by future contributors. Thanks for the clarification @satmandu

satmandu:
Here's a place which mentions that one should only use ENV to get environment variables:
https://www.rubyguides.com/2019/01/ruby-environment-variables/#menu-item-3157:~:text=Warning%3A,like%20dry%2Dconfigurable%20for%20internal%20configuration%20options.

image

PR #4843, opened 2020-12-29
remove clang dep

satmandu:
removes clang dep & uses meson c_args for lld

Works properly:
- [x] x86_64

### DIFF
--- a/packages/xorg_proto.rb
+++ b/packages/xorg_proto.rb
@@ -3,35 +3,35 @@ require 'package'
 class Xorg_proto < Package
   description 'The xorgproto package provides the header files required to build the X Window system, and to allow other applications to build against the installed X Window system.'
   homepage 'https://www.x.org/'
-  version '2020.1'
+  version '2020.1-1'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


packages/xorg_proto.rb
```diff
@@ -3,35 +3,35 @@
...
   depends_on 'llvm' => ':build'
   
   def self.build
-    ENV['CC'] = 'clang'
-    ENV['CXX'] = 'clang'
-    system "meson #{CREW_MESON_OPTIONS} build"
-    system "meson configure build"
-    system "ninja -C build"
+    system "meson #{CREW_MESON_OPTIONS} \
+      -Dc_args='-fuse-ld=lld' \
```
uberhacker, line 28: Should we go ahead and add this to `CREW_MESON_OPTIONS`?

satmandu: I have only added lld to compiles where there is a complaint of the .relr section in a library not allowing linking.

But maybe let's wait to switch to llvm11 when I get that working well and then we can do that, since that lld is newer too.

PR #4848, opened 2020-12-30
Coloring of the output of install.sh

supechicken:
Might be better to read?

Tested on `x86_64`


supechicken:
Screenshot 2020-12-30 at 1 43 34 PM

install.sh
```diff
@@ -229,16 +230,16 @@ function download_check () {
...
     #verify
-    echo "Verifying ${1}..."
-    echo "${4}" "${3}" | sha256sum -c -
+    echo -e "${BLUE}Verifying ${1}..."
+    echo -e ${GREEN}`echo "${4}" "${3}" | sha256sum -c -`${RESET}
     case "${?}" in
     0) ;;
     *)
-      echo "${RED}Verification failed, something may be wrong with the download."
+      echo -e "${RED}Verification failed, something may be wrong with the download."
```
uberhacker, line 241: Missing `${RESET}` at the end of the string?

install.sh
```diff
@@ -250,9 +251,9 @@ function extract_install () {
     cd "${CREW_DEST_DIR}"
 
     #extract and install
-    echo "Extracting ${1} (this may take a while)..."
+    echo -e "Extracting ${1} (this may take a while)..."
```
uberhacker, line 253: Missing `${BLUE}` at the start and `${RESET}` at the end of the string?

install.sh
```diff
@@ -250,9 +251,9 @@ function extract_install () {
...
     tar xpf ../"${2}"
-    echo "Installing ${1} (this may take a while)..."
+    echo -e "Installing ${1} (this may take a while)...${RESET}"
```
uberhacker, line 255: Missing `${BLUE}` at the start of the string?

install.sh
```diff
@@ -262,17 +263,17 @@ function update_device_json () {
...
-    echo "Adding new information on ${1} to device.json..."
+    echo -e "${RESET}Adding new information on ${1} to device.json..."
     sed -i device.json -e '/^    }$/s/$/,\
     {\
       "name": "'"${1}"'",\
       "version": "'"${2}"'"\
     }/'
   else
-    echo "Adding new information on ${1} to device.json..."
+    echo -e "${RESET}Adding new information on ${1} to device.json...${RESET}"
```
uberhacker, line 275: I'm confused.  Why 2 `${RESET}`s at the start and end of the string?

install.sh
```diff
@@ -229,16 +230,16 @@ function download_check () {
     cd "${CREW_BREW_DIR}"
 
     #download
-    echo "Downloading ${1}..."
+    echo -e "${BLUE}Downloading ${1}...${RESET}"
```
uberhacker, line 232: There is already an alias to `echo -e` on line 25.  IOW, no need to add the `-e` option.

uberhacker:
Nice idea @supechicken!

install.sh
supechicken: The `echo` alias doesn't work on my laptop :(

install.sh
supechicken: Removed

install.sh
supechicken: Done

install.sh
supechicken: I think the original color (white) is better :)

install.sh
uberhacker: Did I miss the change?

install.sh
uberhacker: If the color is white already why do we need the `${RESET}`?

supechicken: Removed

install.sh
```diff
@@ -326,34 +332,34 @@ yes | crew install buildessential less most
...
-  echo "echo 'export CREW_PREFIX=${CREW_PREFIX}' >> ~/.bashrc"
-  echo "echo 'export PATH=\"\${CREW_PREFIX}/bin:\${CREW_PREFIX}/sbin:\${PATH}\"' >> ~/.bashrc"
-  echo "echo 'export LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}' >> ~/.bashrc"
-  echo 'source ~/.bashrc'
+  echo -e "${YELLOW}Since you have installed Chromebrew in a directory other than '/usr/local',"
+  echo -e "${YELLOW}you need to run these commands to complete your installation:${RESET}"
+  echo -e "${BLUE}echo 'export CREW_PREFIX=${CREW_PREFIX}' >> ~/.bashrc"
+  echo -e "${BLUE}echo 'export PATH=\"\${CREW_PREFIX}/bin:\${CREW_PREFIX}/sbin:\${PATH}\"' >> ~/.bashrc"
+  echo -e "${BLUE}echo 'export LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}' >> ~/.bashrc"
+  echo -e "${BLUE}source ~/.bashrc${RESET}"
```
uberhacker, line 340: Does this work?
```
  echo -e "${YELLOW}Since you have installed Chromebrew in a directory other than '/usr/local',"
  echo -e "you need to run these commands to complete your installation:${RESET}"
  echo -e "${BLUE}echo 'export CREW_PREFIX=${CREW_PREFIX}' >> ~/.bashrc"
  echo -e "echo 'export PATH=\"\${CREW_PREFIX}/bin:\${CREW_PREFIX}/sbin:\${PATH}\"' >> ~/.bashrc"
  echo -e "echo 'export LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}' >> ~/.bashrc"
  echo -e "source ~/.bashrc${RESET}"
```

install.sh
```diff
@@ -256,17 +262,17 @@ function update_device_json () {
...
-    echo "Adding new information on ${1} to device.json..."
+    echo -e "${RESET}Adding new information on ${1} to device.json..."
     sed -i device.json -e '/^    }$/s/$/,\
     {\
       "name": "'"${1}"'",\
       "version": "'"${2}"'"\
     }/'
   else
-    echo "Adding new information on ${1} to device.json..."
+    echo -e "Adding new information on ${1} to device.json...${RESET}"
```
uberhacker, line 275: Can we remove `${RESET}` on all the lines without a color set?  These should all be white anyway.

install.sh
supechicken: Just found a better solution
```bash
echo -e "${BLUE}To set the default PAGER environment variable to use less:
echo \"export PAGER='less'\" >> ~/.bashrc && . ~/.bashrc

Alternatively, you could use most.  Why settle for less, right?
echo \"export PAGER='most'\" >> ~/.bashrc && . ~/.bashrc

Below are some text editor suggestions.

To install 'nano', execute:
crew install nano

Or, to get an updated version of 'vim', execute:
crew install vim

You may wish to set the EDITOR environment variable for an editor default.

For example, to set 'nano' as the default editor, execute:
echo \"export EDITOR='nano'\" >> ~/.bashrc && . ~/.bashrc

To set 'vim' as the default editor, execute:
echo \"export EDITOR='vim'\" >> ~/.bashrc && . ~/.bashrc

${RESET}"
```

install.sh

uberhacker:
LGTM, nice work!

PR #4863, opened 2020-12-31
Add jdk15 package

uberhacker:
Tested on x86_64.

### DIFF
--- a/packages/jdk15.rb
+++ b/packages/jdk15.rb
@@ -1,9 +1,9 @@
 require 'package'
 
-class Jdk11 < Package
+class Jdk15 < Package
   description 'The JDK is a development environment for building applications, applets, and components using the Java programming language.'
-  homepage 'https://www.oracle.com/java/technologies/javase-jdk11-downloads.html'
-  version '11.0.9'
+  homepage 'https://www.oracle.com/java/technologies/javase-jdk15-downloads.html'
+  version '15.0.1'
   compatibility 'x86_64'
   case ARCH
   when 'x86_64'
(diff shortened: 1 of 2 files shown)


satmandu:
Why can't different java versions be installed simultaneously? 

E.g. some apps specifically need jdk8...

uberhacker:
I suppose they could but the executables would need to be renamed.  They share the same binaries, otherwise.  Same with the manpages.

packages/jdk15.rb
```diff
@@ -0,0 +1,58 @@
...
+  binary_url ({
+    x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/jdk15-15.0.1-chromeos-x86_64.tar.xz',
+  })
+  binary_sha256 ({
+    x86_64: '456f4cca19b0e5bbc89cba06b80527861647fd7f6e740b476cd2cd86cefaacdd',
+  })
+
+  case ARGV[0]
+  when 'install', 'reinstall'
+    abort "JDK8 installed.".lightgreen if Dir.exists? "#{CREW_PREFIX}/share/jdk8"
```
supechicken, line 23: Can we use `exit(0)` to replace `abort`? `abort` will make the exit code non-zero and immediately stop the installation when installing this package as a dependency.

uberhacker: It's fine to have an exit code of non-zero.  The idea here is to "abort" if any other version is installed.  So, in essence, this really means only one version of the jdk can be installed at a time.  In theory, other packages that depend on jdk8 could still work since the executables are the same.

supechicken: I mean the `abort` will exit the `package.rb` (not only the `jdk15.rb`), this will also cause packages (which depends on `jdk15`) installation stop.
e.g
```ruby
require 'jdk15'
# The installation will exit here due to the package.rb stopped
require 'vim'

def self.install
   ...
```

uberhacker: That is fine.  The way to fix that would be to remove jdk8 or jdk11 and then try again.

uberhacker:
@satmandu: Do you have any idea how to allow multiple jdk versions to coexist?  I imagine we would just rename the executables to include the version?  However, if we do this then how could we control the "default" jdk such as the javac executable?  Debian does this with symlinks called alternatives.  See https://wiki.debian.org/DebianAlternatives.

satmandu:
update-alernatives generates symlinks from versioned apps to generic app names.
e.g.  gcc-11 is symlinked to gcc, and so you can use the update-alternatives script to change the symlink for gcc to gcc-10 or whatever.

satmandu:
The update-alternatives script does some heavier lifting in debian/ubuntu, setting the symlinks to /etc/alternatives/ and then from there to the final app:

```
ls -aFl `which java`
lrwxrwxrwx 1 root root 22 Oct 20  2019 /usr/bin/java -> /etc/alternatives/java*
ls -aFl /etc/alternatives/java
lrwxrwxrwx 1 root root 46 Oct 25  2019 /etc/alternatives/java -> /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java*
```

satmandu:
I think the way to handle that would be to import the update-alternatives scripts wholesale from deban, and then just create a /usr/local/etc/alternatives folder to do exactly the same thing for java, gcc, g++, etc...

supechicken:
See #5417

PR #4869, opened 2021-01-01
Update fpc from 3.0.4 to 3.2.0

uberhacker:
Tested on all architectures.  The arm test was a bit flaky but every other platform worked fine, although the fpc executable appeared to be expecting ppc architecture?


satmandu:
Just curious, what's the rationale to install precompiled packages instead of building from source?

Debian appears to build from source.

E.g. https://launchpadlibrarian.net/494660205/fpc_3.2.0+dfsg-8build1_3.2.0+dfsg-8build2.diff.gz

PR #4881, opened 2021-01-05
Add Check

satmandu:
- newer pulseaudio asks for this.

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/check.rb
+++ b/packages/check.rb
@@ -1,0 +1,41 @@
+require 'package'
+
+class Check < Package
+  description 'A unit testing framework for C'
+  homepage 'https://libcheck.github.io/check/'
+  version '0.15.2'
+  source_url 'https://github.com/libcheck/check/releases/download/0.15.2/check-0.15.2.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/check.rb
```diff
@@ -0,0 +1,41 @@
...
+  })
+
+
+  def self.build
+    system "autoreconf -fvi"
+    Dir.mkdir 'build'
+    Dir.chdir 'build' do
+      system "env CC=gcc LD=ld \
+      cmake -G Ninja \
+      -DCMAKE_INSTALL_PREFIX=/usr/local \
```
uberhacker, line 30: Abstract with `#{CREW_PREFIX}` or better yet, `#{CREW_CMAKE_OPTIONS}`?

satmandu: I tried that and I was getting a bunch of errors, which is why I just went back to the minimal options which worked.

uberhacker: That is fine.  You don't have to recreate the binaries but please change `/usr/local` to `#{CREW_PREFIX}`.

satmandu: Whoops! Good catch.

PR #4892, opened 2021-01-08
meson -> .56.1

satmandu:
Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/meson.rb
+++ b/packages/meson.rb
@@ -21,8 +21,6 @@ class Meson < Package
       x86_64: 'b4eca2d237740bfafbc211adc8b352685335c46db29f3609e848c1931f137e1a',
   })
 
-
-
   depends_on 'ninja'
   depends_on 'setuptools'
 
--- a/packages/meson.rb
+++ b/packages/meson.rb
@@ -3,23 +3,25 @@ require 'package'
 class Meson < Package
   description 'The Meson Build System'
   homepage 'http://mesonbuild.com/'
-  version '0.56.0'
+  version '0.56.1'
   compatibility 'all'
-  source_url 'https://github.com/mesonbuild/meson/archive/0.56.0.tar.gz'
-  source_sha256 'a9ca7adf66dc69fbb7e583f7c7aef16b9fe56ec2874a3d58747e69a3affdf300'
+  source_url 'https://github.com/mesonbuild/meson/archive/0.56.1.tar.gz'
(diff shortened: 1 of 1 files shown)


packages/meson.rb
```diff
@@ -3,24 +3,26 @@
...
-       i686: 'b63ee258889c30587a784dcfd934edf6306215b57bcfd13cd0bc00c8866de34f',
-     x86_64: '0141299bd8810a52b23f5985cd4d4682b6b4207ecfebac86d4c58248acfb14ab',
+     aarch64: 'af7616839ef04f824be003d0a534a5a774ef53ed51ff6cafdd678c4744b09787',
+      armv7l: 'af7616839ef04f824be003d0a534a5a774ef53ed51ff6cafdd678c4744b09787',
+        i686: '6ad48fdd56b57d8fbbdf11a7c4a76d684937f5868dec8ebf1afe41b60a3d9655',
+      x86_64: 'b4eca2d237740bfafbc211adc8b352685335c46db29f3609e848c1931f137e1a',
   })
 
+
+
```
uberhacker, line 25: Why the extra lines?  These are nitpicky things, I know, but it helps with readability and consistency if we use only one blank line between sections.  I've noticed inconsistencies with indentation in some packages also.

satmandu: The spacing is due to my script adding lines when substituting in the URLs.
Would this function be sufficient to sanitize spacing:
```
formatfile () {
  file=${1}
  sed -i '/^$/d' "${file}"
  sed '/source_sha256/,/def self/{ /^[ \t]*$/d}' -i "${file}"
  sed '0,/require/G' -i "${file}"
  sed '/binary_url/i'$'\n' -i "${file}"
  sed '/def\ self/i'$'\n' -i "${file}"
  sed 'x;/./{x;b};x;/depends_on/h;//i'$'\n' -i "${file}"
}
```

uberhacker: Maybe?  The chkfrog.sh script already does a pretty good job of inserting the binaries.

satmandu: I generate that stuff locally for local installs from cache on the buildboxen containers, and then just use sed to swap in the bintray URLS. That way I can do testing without having uploaded to bintray.

But this should be cleaner now here on out.

I probably need to set my geany default to two spaces for indent too.

PR #4895, opened 2021-01-09
Update php packages

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/php80.rb
+++ b/packages/php80.rb
@@ -1,24 +1,24 @@
 require 'package'
 
-class Php74 < Package
+class Php80 < Package
   description 'PHP is a popular general-purpose scripting language that is especially suited to web development.'
   homepage 'http://www.php.net/'
-  version '7.4.14'
+  version '8.0.1'
   compatibility 'all'
-  source_url 'https://www.php.net/distributions/php-7.4.14.tar.xz'
-  source_sha256 'f9f3c37969fcd9006c1dbb1dd76ab53f28c698a1646fa2dde8547c3f45e02886'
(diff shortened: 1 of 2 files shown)


packages/php80.rb
```diff
@@ -0,0 +1,164 @@
...
+    #system "sed -i '98295,98296d' configure"
+    # Fix /usr/bin/file: No such file or directory
+    system 'filefix'
+  end
+
+  def self.build
+    ENV['TMPDIR'] = "#{CREW_PREFIX}/tmp"
+    ENV['CFLAGS'] = ' -liconv'
+    system './configure',
+           "--prefix=#{CREW_PREFIX}",
```
supechicken, line 71: Does `CREW_OPTIONS` works?

packages/php80.rb
```diff
@@ -0,0 +1,164 @@
...
+    system 'make'
+  end
+
+  def self.check
+    #system 'make', 'test'
+  end
+
+  def self.install
+    system "mkdir -p #{CREW_DEST_PREFIX}/log"
+    system "mkdir -p #{CREW_DEST_PREFIX}/tmp/run"
```
supechicken, line 133: Rubyize with `FileUtils`?

packages/php80.rb
uberhacker: Yes, but it requires a lot of refactoring.  I was hoping you didn't notice. :)

packages/php80.rb
uberhacker: Good call.

PR #4896, opened 2021-01-12
Add GNU MACchanger to package list

saltedcoffii:
Allow for mac address "spoofing" along with other things

Tested on x86_64. Not tested on any other architecture, however I don't see why it wouldn't work. It compiles in less than a minute so it may not need a binary package.

### DIFF
--- a/packages/macchanger.rb
+++ b/packages/macchanger.rb
@@ -5,8 +5,8 @@ class Macchanger < Package
   homepage 'http://www.gnu.org/software/macchanger/'
   version '1.7.0'
   compatibility 'all'
-  source_url 'https://ftp.gnu.org/gnu/macchanger/macchanger-1.6.0.tar.gz'
-  source_sha256 '31534f138f1d21fa247be74ba6bef3fbfa47bbcd5033e99bd10c432fe58e51f7'
+  source_url 'https://github.com/alobbs/macchanger/releases/download/1.7.0/macchanger-1.7.0.tar.gz'
+  source_sha256 'dae2717c270fd5f62d790dbf80c19793c651b1b26b62c101b82d5fdf25a845bf'
 
   def self.build
     system "./configure", "--prefix=#{CREW_PREFIX}"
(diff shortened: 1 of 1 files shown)


satmandu (changes requested):
Version and source_url don't match up?

packages/macchanger.rb
```diff
@@ -0,0 +1,22 @@
+require 'package'
+
+class Macchanger < Package
+  description 'GNU MAC Changer is an utility that makes the maniputation of MAC addresses of network interfaces easier.'
+  homepage 'http://www.gnu.org/software/macchanger/'
+  version '1.7.0'
+  compatibility 'all'
+  source_url 'https://ftp.gnu.org/gnu/macchanger/macchanger-1.6.0.tar.gz'
```
uberhacker, line 8: Better source? https://ftpmirror.gnu.org/macchanger/macchanger-1.6.0.tar.gz  This will localize the download to the nearest mirror.

packages/macchanger.rb
```diff
@@ -0,0 +1,22 @@
...
+  source_sha256 '31534f138f1d21fa247be74ba6bef3fbfa47bbcd5033e99bd10c432fe58e51f7'
+
+  def self.build
+    system "./configure", "--prefix=#{CREW_PREFIX}"
```
uberhacker, line 12: Probably better to use `system "./configure #{CREW_OPTIONS}"`.  This will abstract libraries and other machine specific options.

packages/macchanger.rb
```diff
@@ -0,0 +1,22 @@
+require 'package'
+
+class Macchanger < Package
+  description 'GNU MAC Changer is an utility that makes the maniputation of MAC addresses of network interfaces easier.'
+  homepage 'http://www.gnu.org/software/macchanger/'
+  version '1.7.0'
```
uberhacker, line 6: As @satmandu mentioned, this should be `version '1.6.0'`.

uberhacker:
Thank you for the new package @saltedcoffii!

saltedcoffii:
> Version and source_url don't match up?

Sorry, the latest release on the github was 1.7.0 but the latest release on gnu.org was 1.6.0. I chose the GNU mirror but forgot to downgrade the version.

saltedcoffii:
Alright I made some changes.
- The version number stayed the same but I used the [release from the github](https://github.com/alobbs/macchanger/releases/tag/1.7.0).
- Changed `./configure --prefix="/usr/local" to `./configure ${CREW_OPTIONS}`

Package built and works without issue.

PR #4897, opened 2021-01-13
Update youtubedl to 2021.01.08

saltedcoffii:
I also changed the homepage to the website listed on youtubedl's github

I'm not sure how you as a chromebrew community handle binary packages so I just removed them but you're welcome to add the older ones back or make new ones and add them before the file merges


uberhacker:
Good call!  We need to download videos fast before they are censored. LoL

PR #4898, opened 2021-01-13
[New Package] e2fsprogs

saltedcoffii:
Although chromeos comes with e2fsprogs in the root filesystem, this will provide updated binaries, header files and manpages.
Builds well on x86_64, hasn't been tested on other architectures. May need `./configure` flag adjustments; I took defaults and it works fine.

### DIFF
--- a/packages/e2fsprogs.rb
+++ b/packages/e2fsprogs.rb
@@ -4,6 +4,7 @@ class E2fsprogs < Package
   description 'e2fsprogs are ext2/3/4 file system utilities.'
   homepage 'http://e2fsprogs.sourceforge.net/'
   version '1.45.6'
+  compatibility 'all'
   source_url 'https://mirrors.edge.kernel.org/pub/linux/kernel/people/tytso/e2fsprogs/v1.45.6/e2fsprogs-1.45.6.tar.xz'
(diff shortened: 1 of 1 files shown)


uberhacker:
LGTM, thank you @saltedcoffii!

packages/e2fsprogs.rb
```diff
@@ -0,0 +1,22 @@
+require 'package'
+
+class E2fsprogs < Package
+  description 'e2fsprogs are ext2/3/4 file system utilities.'
+  homepage 'http://e2fsprogs.sourceforge.net/'
+  version '1.45.6'
```
uberhacker, line 6: Missing `compatibility 'all'`?

saltedcoffii: I'm not sure if it is all compatible, so I think maybe we should let the package marinate for a week or two and if any arm users report problems we can resolve them before adding compatibility all. Is that okay?

uberhacker: The fact that you don't have a line there proves to me that you haven't tested this because it will break crew. :)

saltedcoffii: I have tested it and haven't had a problem on x86_64. Unfortunately i don't own an arm chromebook :(. Has this package been tried before and had issues?

saltedcoffii:
@uberhacker added `compatibility 'all'` as you requested

PR #4901, opened 2021-01-13
[New Package] tar

saltedcoffii:
Add new package GNU tar. Although tar exists in chromium os, this should provide an updated binary, header files and manpages.

Compiles fine on x86_64 and should compile fine on arm but hasn't been tested.

### DIFF
--- a/packages/tar.rb
+++ b/packages/tar.rb
@@ -5,7 +5,7 @@ class Tar < Package
   homepage 'https://www.gnu.org/software/tar/'
   version '1.33'
   compatibility 'all'
-  source_url 'https://ftp.gnu.org/gnu/tar/tar-1.33.tar.xz'
+  source_url 'https://ftpmirror.gnu.org/tar/tar-1.33.tar.xz'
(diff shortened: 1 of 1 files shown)


packages/tar.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Tar < Package
+  description 'GNU Tar provides the ability to create tar archives, as well as various other kinds of manipulation.'
+  homepage 'https://www.gnu.org/software/tar/'
+  version '1.33'
+  compatibility 'all'
+  source_url 'https://ftp.gnu.org/gnu/tar/tar-1.33.tar.xz'
```
uberhacker, line 8: Better source? https://ftpmirror.gnu.org/tar/tar-1.33.tar.xz

uberhacker:
Thank you for the new package @saltedcoffii!  You are on a roll.

saltedcoffii:
Source changed ✅

satmandu:
Does this have a xzutils dependency, or is that in base?

PR #4902, opened 2021-01-13
Add telegram package

uberhacker:
Telegram is a messaging app with a focus on speed and security, it's super-fast, simple and free.  Great for those who have been banned from Facebook, Twitter and other tyrannical social media platforms. :) Tested on x86_64.

### DIFF
--- a/packages/telegram.rb
+++ b/packages/telegram.rb
@@ -1,0 +1,31 @@
+require 'package'
+
+class Telegram < Package
+  description "Telegram is a messaging app with a focus on speed and security, it's super-fast, simple and free."
+  homepage 'https://telegram.org/'
+  version '2.5.1'
+  compatibility 'x86_64'
+  case ARCH
+  when 'x86_64'
+    source_url 'https://updates.tdesktop.com/tlinux/tsetup.2.5.1.tar.xz'
+    source_sha256 '939275a84498dedb1a33ac3f0242b0211548ddb087f9801518db92c5d6dc0151'
+  end
+
+  def self.build
+    system "cat <<'EOF'> telegram
+#!/bin/bash
+QT_QPA_PLATFORM=wayland-egl DISPLAY= WAYLAND_DISPLAY=wayland-0 Telegram \"\$@\"
+EOF"
+  end
+
+  def self.install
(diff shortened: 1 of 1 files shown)


packages/telegram.rb
```diff
@@ -0,0 +1,24 @@
...
+class Telegram < Package
+  description "Telegram is a messaging app with a focus on speed and security, it's super-fast, simple and free."
+  homepage 'https://telegram.org/'
+  version '2.5.1'
+  compatibility 'x86_64'
+  case ARCH
+  when 'x86_64'
+    source_url 'https://updates.tdesktop.com/tlinux/tsetup.2.5.1.tar.xz'
+    source_sha256 '939275a84498dedb1a33ac3f0242b0211548ddb087f9801518db92c5d6dc0151'
+    depends_on 'sommelier'
```
supechicken, line 12: Just take a try on it:) It works with `Qt` native Wayland backend (`QT_QPA_PLATFORM=wayland-egl DISPLAY= telegram`)

supechicken: Missing `qtbase` dependency?

satmandu: Sigh. We really need a way to do ```depends_on 'a' or 'b'```.

satmandu: Also agree on avoiding a sommelier dependency where the system wayland server will suffice... but does that work on i686? I imagine the wayland server there is pretty old? Or can we update that somehow?

uberhacker: @supechicken @satmandu: I have removed the `sommelier` dependency.   Apparently, the `qtbase` dependency is not needed.

packages/telegram.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def self.build
+    system "cat <<'EOF'> telegram
+#!/bin/bash
+QT_QPA_PLATFORM=wayland-egl DISPLAY= WAYLAND_DISPLAY=wayland-0 Telegram \"\$@\"
+EOF"
+  end
+
+  def self.install
+    system "install -Dm755 Telegram #{CREW_DEST_PREFIX}/bin/Telegram"
```
supechicken, line 22: Rubyize with `FileUtils.install 'Telegram', "#{CREW_DEST_PREFIX}/bin/Telegram", mode: 0755` ?

PR #4907, opened 2021-01-13
OpenSSL 1.1.1i

### LINKED ISSUE #4888: Should we drop libressl and go back to openssl?
Gentoo is dropping it: https://archives.gentoo.org/gentoo-dev/message/55cb4dd700ff60361332b4475fd5ec00

OpenSSH & wget apparently use it too, but maybe those should be statically linked so we don't have to deal with this?

satmandu:
Fixes #4888

- Does NOT appear to conflict with libraries from libressl
- As long as this is installed after libressl, all new compiles should use the openssl headers and build with openssl
- Tested to work with curl.
- This can be added to ```install.sh``` after the libressl lines.
- Once packages using libressl have been recompiled with openssl, we can remove libressl. 

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/openssl.rb
+++ b/packages/openssl.rb
@@ -4,7 +4,7 @@ class Openssl < Package
   description 'The Open Source toolkit for Secure Sockets Layer and Transport Layer Security'
   homepage 'https://www.openssl.org'
   @_ver = '1.1.1i'
-  version _ver
+  version @_ver
   compatibility 'all'
   source_url "https://www.openssl.org/source/openssl-#{@_ver}.tar.gz"
   source_sha256 'e8be6a35fe41d10603c3cc635e93289ed00bf34b79671a3a4de64fcee00d5242'
--- a/packages/openssl.rb
+++ b/packages/openssl.rb
@@ -27,26 +27,26 @@ class Openssl < Package
 
   case ARCH
   when 'aarch64','armv7l'
-    ARCH_C_FLAGS = '-march=armv7-a -mfloat-abi=hard'
-    ARCH_CXX_FLAGS = '-march=armv7-a -mfloat-abi=hard'
+    @ARCH_C_FLAGS = '-march=armv7-a -mfloat-abi=hard'
+    @ARCH_CXX_FLAGS = '-march=armv7-a -mfloat-abi=hard'
     OPENSSL_CONFIGURE_TARGET = 'linux-generic32'
   when 'i686'
-    ARCH_C_FLAGS = ''
-    ARCH_CXX_FLAGS = ''
+    @ARCH_C_FLAGS = ''
+    @ARCH_CXX_FLAGS = ''
     OPENSSL_CONFIGURE_TARGET = 'linux-x86'
   when 'x86_64'
-    ARCH_C_FLAGS = ''
-    ARCH_CXX_FLAGS = ''
+    @ARCH_C_FLAGS = ''
+    @ARCH_CXX_FLAGS = ''
     OPENSSL_CONFIGURE_TARGET = 'linux-x86_64'
   end
-  ARCH_C_LTO_FLAGS = "#{ARCH_C_FLAGS} -flto=auto"
-  ARCH_CXX_LTO_FLAGS = "#{ARCH_CXX_FLAGS} -flto=auto"
+  @ARCH_C_LTO_FLAGS = "#{@ARCH_C_FLAGS} -flto=auto"
+  @ARCH_CXX_LTO_FLAGS = "#{@ARCH_CXX_FLAGS} -flto=auto"
 
   def self.build
     # This gives you the list of OpenSSL configure targets
     system "./Configure LIST"
     system "env  PATH=#{CREW_LIB_PREFIX}/ccache/bin:#{CREW_PREFIX}/bin:/usr/bin:/bin \
-    CFLAGS=\"#{ARCH_C_LTO_FLAGS}\" CXXFLAGS=\"#{ARCH_CXX_LTO_FLAGS}\" \
+    CFLAGS=\"#{@ARCH_C_LTO_FLAGS}\" CXXFLAGS=\"#{@ARCH_CXX_LTO_FLAGS}\" \
     ./Configure --prefix=#{CREW_PREFIX} \
     --libdir=#{CREW_LIB_PREFIX} \
     --openssldir=#{CREW_PREFIX}/etc/ssl \
--- a/packages/openssl.rb
+++ b/packages/openssl.rb
@@ -4,7 +4,7 @@ class Openssl < Package
   description 'The Open Source toolkit for Secure Sockets Layer and Transport Layer Security'
   homepage 'https://www.openssl.org'
   _ver = '1.1.1i'
-  version "#{_ver}"
+  version _ver
   compatibility 'all'
   source_url "https://www.openssl.org/source/openssl-#{_ver}.tar.gz"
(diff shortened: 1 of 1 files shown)


packages/openssl.rb
```diff
@@ -1,12 +1,64 @@
...
 class Openssl < Package
-  description 'OpenSSL is a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols.'
-  homepage 'https://www.openssl.org/'
-  version '3.0.2'
+  description 'The Open Source toolkit for Secure Sockets Layer and Transport Layer Security'
+  homepage 'https://www.openssl.org'
+  _ver = '1.1.1i'
+  version "#{_ver}"
   compatibility 'all'
+  source_url "https://www.openssl.org/source/openssl-#{_ver}.tar.gz"
```
uberhacker, line 9: I'm thinking it's not worth it to create a `_ver` variable if it's only used twice?

packages/openssl.rb
```diff
@@ -1,12 +1,64 @@
...
+    ARCH_C_FLAGS = ''
+    ARCH_CXX_FLAGS = ''
+    OPENSSL_CONFIGURE_TARGET = 'linux-x86'
+  when 'x86_64'
+    ARCH_C_FLAGS = ''
+    ARCH_CXX_FLAGS = ''
+    OPENSSL_CONFIGURE_TARGET = 'linux-x86_64'
+  end
+  ARCH_C_LTO_FLAGS = "#{ARCH_C_FLAGS} -flto=auto"
+  ARCH_CXX_LTO_FLAGS = "#{ARCH_CXX_FLAGS} -flto=auto"
```
uberhacker, line 43: Would it make sense to create crew constants in case these options need to be used for other packages?

packages/openssl.rb
supechicken: Why not using the version variable directly? (`source_url "https://www.openssl.org/source/openssl-#{version}.tar.gz"`)

packages/openssl.rb
```diff
@@ -1,12 +1,64 @@
...
+
+  def self.build
+    # This gives you the list of OpenSSL configure targets
+    system "./Configure LIST"
+    system "env  PATH=#{CREW_LIB_PREFIX}/ccache/bin:#{CREW_PREFIX}/bin:/usr/bin:/bin \
+    CFLAGS=\"#{ARCH_C_LTO_FLAGS}\" CXXFLAGS=\"#{ARCH_CXX_LTO_FLAGS}\" \
+    ./Configure --prefix=#{CREW_PREFIX} \
+    --libdir=#{CREW_LIB_PREFIX} \
```
uberhacker, line 51: I assume you tried #{CREW_OPTIONS} and it didn't work?

satmandu: Correct.

packages/openssl.rb
satmandu: Yes, but gcc supports ```-flto=auto``` and clang supports ```-flto=thin``` but thin doesn't work with gcc.
Meson also has a ```b_lto``` option but it doesn't currently support both either.

Is there a ruby += option to add something to an existing variable? That WOULD be useful.

uberhacker:
@satmandu: I have to admire your courage to take on this beast.  Last time we tried upgrading openssl, it literally broke almost everything.

packages/openssl.rb
satmandu: So this is a hack that arch is using, which could work for us too.

```_ver ``` is source package version
crew package```version``` is ```#{_ver}-1``` or ```#{_ver}-2``` or whatever.
If package sources have the source in the URL, this makes updates easier, no?

Comment (satmandu) [reply packages/openssl.rb:9]: Say we make a modification to the package file, e.g. hey clang makes this library only 3 bytes! or whatever.
In that case source_url breaks since we are incrementing ```version```.

Comment (satmandu) [reply packages/openssl.rb:9]: I'm not wedded to this approach. Just trying to make incremental improvements for maintainability.

**Comment by satmandu** [comment]
@uberhacker I was just getting sick of having to look up workarounds for libressl issues. And I say that as someone who used OpenBSD on servers for over a decade.

But maybe the pain now is worth it, before OpenSSL 3.0 comes out this year?

File: packages/openssl.rb (hunk 1 above)
Comment (uberhacker) [reply packages/openssl.rb:9]: That is a solid case for doing it this way.  I also think you could simply use `version _ver` and it would work the same (no need to put in quotes).  If we did that, then you could do something like `version _ver + '-1'`.

**Comment by satmandu** [comment]
But seriously, please try this out on your machines before this is merged. It works fine on my machine and when compiling other packages in the docker containers...

File: packages/openssl.rb (hunk 2 above)
Comment (uberhacker) [reply packages/openssl.rb:43]: Yeah, if these are just one-off changes then we can do it this way.

File: packages/openssl.rb (hunk 1 above)
Comment (satmandu) [reply packages/openssl.rb:9]: Modified accordingly.

File: packages/openssl.rb [hunk 4]
```diff
@@ -1,12 +1,64 @@
 require 'package'
 
 class Openssl < Package
-  description 'OpenSSL is a robust, commercial-grade, and full-featured toolkit for the Transport Layer Security (TLS) and Secure Sockets Layer (SSL) protocols.'
-  homepage 'https://www.openssl.org/'
-  version '3.0.2'
+  description 'The Open Source toolkit for Secure Sockets Layer and Transport Layer Security'
+  homepage 'https://www.openssl.org'
+  _ver = '1.1.1i'
+  version _ver
```
supechicken, line 7: The `_ver` variable should add the symbol `@` at first (To set the variable in this class only, not whole `package.rb`)
```ruby
@_ver = '1.1.1i'
```

satmandu: Why not whole ```package.rb```? It's conceivable that in the future ```package.rb``` might use it, no?

supechicken: If there have more than one package using the `_var` variable, it may corrupt `crew upgrade`

satmandu: Ah! I will change it.

satmandu: @uberhacker thoughts on ```s/_ver/@_ver/g``` ?

satmandu: And could ENV variable leakage be responsible for ENV variables from one package install leaking into another package install? e.g. during installs of dependencies before compiling a package with dependencies?

uberhacker: The advantage of using `@` is it makes the variable scope an instance which can be used in all methods within the class.  So, for "instance", you could use the variable in the build and install functions.  More info here: https://stackoverflow.com/questions/14319347/variables-in-ruby-on-rails

uberhacker:
Currently, openssl is a core package.  We need to remove it from https://github.com/skycocker/chromebrew/blob/master/tools/core_packages.txt and replace with libressl if you don't want this to install fresh.

uberhacker:
The way to do that, I believe, is to change nettle to depend_on libressl instead of openssl.

satmandu:
@uberhacker I want to install it in install.sh after libressl, so it should remain in core_packages.
But since libressl installs first in install.sh anyways right now, it shouldn't be a problem anyways right now, since libressl libraries will get installed anyways first before any dependencies are calculated.

uberhacker:
I think we could be running down a very slippery slope here.  Have you tested this on all architectures?

satmandu:
It works within my docker images with package compiles. As I only have a x86_64 ChromeOS machine though, I'd like some help from other people testing this too, who actually have the hardware.

PR #4910, opened 2021-01-13
[New Package] dpkg

saltedcoffii:


uberhacker:
We tried this awhile back and it was a complete disaster.  I don't think this is a good idea.

saltedcoffii:
just curious, what was disastrous about it?

uberhacker:
Your post install message explains why it's a disaster.

saltedcoffii:
with rootfs verification on, nothing too bad should happen, right? idk

supechicken:
Mounting the filesystem and #{CREW_PREFIX} together with aufs/unionfs might be a great solution for `.deb` packages :)

saltedcoffii:
That's a really interesting idea, but I think my intention with packaging `dpkg` was to allow for use of deb file manipulation through the `dpkg-deb` command. `dpkg` is also an optional dependency of `alien` which uses the `dpkg-deb` feature.

uberhacker:
In any event, a package manager should not be included in this project.  This is conceptually doing the same thing as Chromebrew but in a completely different way.  How can `dpkg` be used to assist with `alien`?  Seems like overkill to manipulate packages to work natively on Chromebooks.

saltedcoffii:
Because my chromebook is my only functioning non-mobile device I have and I have a lot of issues with crostini, I use chromebrew installed packages for a lot of development, including compiling and creating deb and rpm files, which I use `dpkg-deb` and `alien` for.

`dpkg-deb` is required for `alien` to create deb files from other files if that's what you're asking.

uberhacker:
I think the issue here is users will confuse installing with developing dev files.  Is there a way to prevent deb installs with dpkg?

saltedcoffii:
It looks like there isn't, at least without patching dpkg source code. I don't know C or perl so I wouldn't be able to do that :(.

saltedcoffii:
> 
> Using `alias`?
> 
> ```shell
> alias 'dpkg -i'='crew postinstall dpkg'
> alias 'dpkg --install'='crew postinstall dpkg'
> ```

unfortunately alias only allows the thing being aliased to be one word. 
```shell
bash: alias: `dpkg -i': invalid alias name
```

saltedcoffii:
Alright, I made https://github.com/saltedcoffii/chromebrew-dpkg-noinstall and integrated the rb file with it. Hopefully this helps :)

saltedcoffii:

PR #4915, opened 2021-01-14
[New Package] Translate Shell

saltedcoffii:
once installed, the executable is invoked with `trans`

### DIFF
--- a/packages/translate_shell.rb
+++ b/packages/translate_shell.rb
@@ -8,10 +8,8 @@ class Translate_shell < Package
   source_url 'https://github.com/soimort/translate-shell/archive/v0.9.6.12.tar.gz'
   source_sha256 '4c4843a8c66276190535b8435775ecb5d9c8286083a33cdbe2db608eba93ca97'
 
-  depends_on 'gawk'
   depends_on 'rlwrap'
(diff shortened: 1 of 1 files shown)


packages/translate_shell.rb
```diff
@@ -0,0 +1,27 @@
...
+  homepage 'https://www.soimort.org/translate-shell/'
+  version '0.9.6.12'
+  compatibility 'all'
+  source_url 'https://github.com/soimort/translate-shell/archive/v0.9.6.12.tar.gz'
+  source_sha256 '4c4843a8c66276190535b8435775ecb5d9c8286083a33cdbe2db608eba93ca97'
+
+  depends_on 'gawk'
+  depends_on 'rlwrap'
+  depends_on 'aspell' #Can also depend on hunspell instead of aspell
+  depends_on 'curl'
```
supechicken, line 14: Not need to include `curl` as it is a core package which installed with `crew` already

supechicken: Same as `gawk`

saltedcoffii: Does it do any harm to be left in?

supechicken: That's fine (just not necessary)

PR #4916, opened 2021-01-14
Add tar to core package list

saltedcoffii:
Although chromeos comes with tar, this will come with an updated binary for older chromebooks, especially chromebooks that are/are reaching EOL

### DIFF
--- a/packages/buildessential.rb
+++ b/packages/buildessential.rb
@@ -64,5 +64,8 @@ class Buildessential < Package
 
   # perl module build ?
   # depends_on 'perl_module_build'
+  
+  # archive utilities
+  depends_on 'tar'
 
 end


saltedcoffii:
Might be prudent to build binary packages for tar as it doesn't have any yet

packages/buildessential.rb
```diff
@@ -64,5 +64,8 @@ class Buildessential < Package
 
   # perl module build ?
   # depends_on 'perl_module_build'
+  
+  # archive utilities
```
supechicken, line 68: Move `tar` into `  # compression utilities` part?

uberhacker: Is there a compelling reason we need to do this?  I'm in the camp of limiting core packages.

supechicken:
I agree with @uberhacker, does there any new features/serious bug fixes in the newer version?

saltedcoffii:
features? yes. serious bugs? no, not really. Let's limit core packages then.

PR #4919, opened 2021-01-14
[New Package] busybox

saltedcoffii:
This package is a bit more complicated than the other ones I've done due to:
- The unique build procedure (it uses kConfig, the build procedure is similar to the Linux kernel)
- The sheer number of configuration options

I left basically everything enabled, including stuff that might never be used such as init, and made no applets because any busybox applets will overwrite any programs crew has already installed, including some core packages. I'm happy to compile out things like that or anything else that wouldn't work on chromeos upon request. This inital PR is more for proof of concept than a production ready package.

I considered installing busybox applets in a seperate place such as `/usr/local/bin/busybox/*` or `/usr/local/busybox/*` and making a postinstall message recommending the user modify their `PATH` however that would be a huge POSIX violation and felt weird. If someone wants me to do that though, I can easily. For the time being, any busybox applets must be called with `busybox <APPLET> [OPTIONS]`

The package.rb file only supports`x86_64` at the moment because of the install command
```
system "install -Dm 755 0_lib/libbusybox.so.1.33.0 #{CREW_DEST_PREFIX}/lib64/libbusybux.so.1.33.0"
```
I think that library needs to be in `{CREW_DEST_PREFIX}/lib` on `x86` and maybe on `arm7l` as well. If someone can confirm for me where that library goes on other architectures I can get that fixed.

Also, I didn't know where to put the `.config` file. I needed it in the repository, or at least somewhere that could be maintained by chromebrew devs and `curl`ed (so pastebin wouldn't work because it can't be edited) and I couldn't find anywhere in the repository where a config file like that would fit, so I made a new folder called `misc` in the root directory of the project and put a file called `busybox-config` in it. This can absolutely be changed.


packages/busybox.rb
```diff
@@ -0,0 +1,31 @@
...
+      system "make", "oldconfig"
+  end
+
+  def self.build
+      system "make"
+  end
+    
+  def self.install
+    system "install -Dm 755 0_lib/busybox #{CREW_DEST_PREFIX}/bin/busybox"
+    system "install -Dm 755 0_lib/libbusybox.so.1.33.0 #{CREW_DEST_PREFIX}/lib64/libbusybux.so.1.33.0"
```
ThatGeekyWeeb, line 25: Do correct me if I'm wrong but `lib64` can be set to `#{ARCH_LIB}` here iirc
```
system "install -Dm 755 0_lib/libbusybox.so.1.33.0 #{CREW_DEST_PREFIX}/#{ARCH_LIB}/libbusybux.so.1.33.0"
```

ThatGeekyWeeb (changes requested):

packages/busybox.rb
uberhacker: @ThatGeekyWeeb:  Correct, but you can take it one step further with:
```
system "install -Dm 755 0_lib/libbusybox.so.1.33.0 #{CREW_DEST_LIB_PREFIX}/libbusybox.so.1.33.0"
```
Notice I changed libbusybux to libbusybox too.  @saltedcoffii:  Please test these things before submitting a pull request.

uberhacker: You can get away with hardcoding the library path since the compatibility is currently locked to x86_64 but it doesn't hurt to abstract the library in case future architectures work.  It's good practice.

saltedcoffii: I actually don't know how it turned into libbusybux. The version I have on my computer says busybox, hence why the test worked. Must have been a copy error, sorry about that.

ThatGeekyWeeb (approved):

uberhacker:
Why do we need the config file in the repo?  Can that be generated in a post install step instead?

saltedcoffii:
I "generated" it with `make menuconfig`. I suppose we could use the default `make defconfig` and then use `diff` to make a patch file and `patch` it in `self.patch`, but that would still require the patch file to reside somewhere. There is another option I was thinking about which was storing the config file in the `busybox.rb` file itself in the form of `system "echo <line> | tee -a .config"` for every line but that doesn't seem very practical, readable or maintainable.

saltedcoffii:
Alright this package should be mergable into the master branch.

saltedcoffii:
for the config file, maybe having a directory in the root of the repository called `data` and the `.config` file can go in `data/busybox`. That way if someone generates patch files for chromebrew they can go in `data/<packagename>/<packagename>-<packageversion>-<patchname>.patch Idk, uberhacker runs this project I just get bored and make packages.

satmandu:
Should this be static? Compiled with musl? I'm just trying to think of use cases where we are using busybox, where presumably other things are fubar?

ThatGeekyWeeb:
Well, how would we even get ahold of musl to compile with? But yes, it should be compiled statically

saltedcoffii:
I don't know why we would use musl as chromeos uses glibc, however a static build wouldn't be a problem, it would just be maybe twice or three times as large as the dynamic build.

ThatGeekyWeeb:
Musl can be used to build fully static fully portable binaries that work with any libc, but I don't think we need to use it, and musl is fully compatible with busybox and needs no patches

> I don't know why we would use musl as chromeos uses glibc. Running two libcs on one machine without a vm, container or chroot sounds like a dependency and header file disaster. Musl is also posix compliant so it would likely require patches to busybox's source code (from alpine's source repos) where glibc wouldn't.
> A static build wouldn't be a problem, it would just be maybe twice or three times as large as the dynamic build.

ThatGeekyWeeb:
And more so static musl binaries are 1.5 times smaller then dynamic glibc binaries

satmandu:
I have static binaries built with musl on my system already. They live alongside the glibc binaries.

packages/busybox.rb
```diff
@@ -0,0 +1,31 @@
...
+  compatibility 'all'
+  version '1.33.0'
+  source_url 'https://www.busybox.net/downloads/busybox-1.33.0.tar.bz2'
+  source_sha256 'd568681c91a85edc6710770cebc1e80e042ad74d305b5c2e6d57a5f3de3b8fbd'
+
+  def self.patch
+      # There's probably a way to use ruby to download this file. If anyone who knows ruby wants to change it, please do so.
+      # When the package is merged, change the curl command to:
+      #system "curl -#L https://raw.github.com/skycocker/chromebrew/master/misc/busybox-config -o ./.config"
+      system "curl -#L https://raw.github.com/saltedcoffii/chromebrew/patch-16/misc/busybox-config -o ./.config"
```
uberhacker, line 15: Would a sha256sum check make sense here?  Probably should put this on a `busybox-config` branch instead of `patch-16` which doesn't make sense to a casual observer.

uberhacker: Remove `misc/busybox-config` since we're downloading here?

saltedcoffii: I guess i don't know where you want the busybox config file and in what branch in which repository (mine or the main chromebrew one). Let me know and i can make those changes/pulls, i just don't know what is wanted.

uberhacker: It probably makes sense to add the config in `/user/chronos/home` or `~`.  Maybe `#{CREW_DEST_HOME}/.config/busybox-config`?  This would require `depends_on 'xdg_base'`.

saltedcoffii:

PR #4922, opened 2021-01-15
Packages updated to work with OpenSSL

satmandu:
Packages compiled after installing openssl  in https://github.com/skycocker/chromebrew/pull/4907

(Xwayland was compiled with mesa 20.3.3 in https://github.com/skycocker/chromebrew/pull/4924 and harfbuzz in https://github.com/skycocker/chromebrew/pull/4923 )
Works properly:
- [x] x86_64
- [x] armv7l (in docker container)
- [x] i686 (in docker container)


uberhacker:
Something about that `@_ver` variable makes me feel dirty inside. :)  Update install.sh also?

satmandu:
@uberhacker FYI I'm moving git below the ssl lines in ```install.sh```, since git is dependent upon openssl.

satmandu:
Tested ```install.sh``` on x86_64 & i686 docker containers. My armv7l container is still building the stage2 llvm, so I can't test that one until some time tomorrow.

install.sh
```diff
@@ -313,23 +319,16 @@ fi
 echo -e "${BLUE}
 To set the default PAGER environment variable to use less:
 echo \"export PAGER='less'\" >> ~/.bashrc && . ~/.bashrc
-
```
supechicken, line 316: Any reason to remove spaces? I think they are better to read with spaces:)

satmandu: Just ran this through shellcheck, which found an actual legitimate bug, (probably an error in the last commit of mine), so I took their other recommendations as well.

install.sh
```diff
@@ -94,15 +94,17 @@ case "${ARCH}" in
...
-  sha256s+=('892ed67dfc19d0f5056836d3a707c1f4d5717bdf82e0d6d0facd11f9eb964fa4')
   urls+=('https://dl.bintray.com/chromebrew/chromebrew/libressl-3.2.3-chromeos-armv7l.tar.xz')
   sha256s+=('59a8e3c3cf6bd97210f04882c2b0ce94566311f360536d1174d7e04e8f9884ab')
-  urls+=('https://dl.bintray.com/chromebrew/chromebrew/curl-7.74.0-chromeos-armv7l.tar.xz')
-  sha256s+=('6c817cc8f2043982148a08713fd5439efc26e80f44788788b58bde6d5a86186c')
+  urls+=('https://dl.bintray.com/chromebrew/chromebrew/openssl-1.1.1i-chromeos-armv7l.tar.xz')
+  sha256s+=('c3033ec7b37b5d451f528df4361a348a3dde750d71a131551039bed8880b5bb0')
+  urls+=('https://dl.bintray.com/chromebrew/chromebrew/curl-7.74.0-1-chromeos-armv7l.tar.xz')
+  sha256s+=('f54a7ff2c3521535af5873b061d26514005e3e080e5b444888e25dce95981201')
+  sha256s+=('81509c5d3585f1a0c3c7d616d9f9a0d0192c6e3276c3ddea23c3905e78e56bfc')
```
uberhacker, line 105: Why do we have back to back sha256s here?

install.sh
```diff
@@ -200,7 +206,7 @@ function download_check () {
 
     #verify
     echo -e "${BLUE}Verifying ${1}...${RESET}"
-    echo -e ${GREEN}`echo "${4}" "${3}" | sha256sum -c -`${RESET}
+    echo -e "${GREEN}$(echo "${4}" "${3}" | sha256sum -c -)${RESET}"
```
uberhacker, line 209: Good call.  `$(...)` is better than backticks.

install.sh
satmandu: Good catch. Fixed.

install.sh
satmandu: Oh! thought youu were talking about some other space. (Sorry android github app is lacking here.)

I'm agnostic as to this space. I'll add it back in.

uberhacker: Run `git diff` before committing? ;)

satmandu: I was actually editing in geany, and then pasting into the web interface for uploading to the specific branch. Normally I check the diff there. I must have hit a geany command to remove spaces somehow. 

The obvious explanation is PEBKAP, but in fairness that could also be connected to Bluetooth issues with my keyboard. I'm getting a hardwired keyboard today, so maybe that will help.

satmandu: (The Pixel Slate keyboard has finally almost died due to a crappy flat ribbon cable to the USB connector shearing as is apparently common. Google's long term hardware support... not great...)

PR #4923, opened 2021-01-15
harfbuzz -> 2.7.4

satmandu:
harfbuzz compiled with llvm 11 & lto

Works properly:
- [x] x86_64

BUILDS properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/harfbuzz.rb
+++ b/packages/harfbuzz.rb
@@ -3,40 +3,47 @@ require 'package'
 class Harfbuzz < Package
   description 'HarfBuzz is an OpenType text shaping engine.'
   homepage 'https://www.freedesktop.org/wiki/Software/HarfBuzz/'
-  version '2.7.2-1'
+  @_ver = '2.7.4'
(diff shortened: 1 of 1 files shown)


packages/harfbuzz.rb
```diff
@@ -3,40 +3,47 @@
...
-  depends_on 'freetype_sub'
-  depends_on 'six' => :build
-  depends_on 'graphite' => :build
+ depends_on 'cairo' => ':build'
+ depends_on 'glib' => :build
+ depends_on 'gobject_introspection'
+ depends_on 'ragel' => :build
+ depends_on 'freetype_sub'
+ depends_on 'six' => :build
+ depends_on 'graphite' => :build
```
uberhacker, line 31: Why was the indentation changed?

satmandu: good catch. That was a manual fix of a script error on my end... which was poorly fixed.

PR #4924, opened 2021-01-15
mesa -> 20.3.3 

satmandu:
- Compiled using llvm11 lld linker
- Removed ENV variables.

Works properly:
- [x] x86_64

BUILDs properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/mesa.rb
+++ b/packages/mesa.rb
@@ -3,22 +3,23 @@ require 'package'
 class Mesa < Package
   description 'Open-source implementation of the OpenGL specification'
   homepage 'https://www.mesa3d.org'
-  version '20.3.1'
+  @_ver = '20.3.3'
+  version @_ver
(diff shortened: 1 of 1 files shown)


packages/mesa.rb
```diff
@@ -39,17 +40,20 @@ class Mesa < Package
   depends_on 'wayland_protocols'
   depends_on 'zstd'
   depends_on 'lm_sensors'
+  depends_on 'ld_default' => :build
```
uberhacker, line 43: Why do we need this dependency if we're not switching ld?

satmandu: We do NOT!

satmandu: (That was cruft from when I was playing around with ld_default to try to get lld to work during testing.)

PR #4928, opened 2021-01-16
[New Package] pypy3

saltedcoffii:
You're probably noticing that this script is a little weird. This is the first time I've ever actually used any ruby so the code might look/be a little bad. I tried for a couple days to build pypy from the source but chromebooks just don't have the power for it, so I'm using precompiled binaries. Also, a precompiled binary of pypy is necessary to build pypy from the source anyway so it seemed like a waste of time and energy to keep trying. I don't think the aarch64 binaries are compatible with armv7l, but someone is more than welcome to try!


packages/pypy3
```diff
@@ -0,0 +1,66 @@
+require 'package'
+
+class Pypy3 < Package
+  description 'pypy is fast, drop-in replacement of cpython that can execute code over 4 times faster.'
+  homepage 'https://www.pypy.org/'
+  version '7.3.3'
+  compatibility 'x86_64, x86, aarch64'
```
uberhacker, line 7: I'm not sure if we do a trim but it's probably better if we don't have spaces.  Was this tested?

saltedcoffii: This was tested. I switched text editors and i'm just noticing the mixtures of tabs and spaces. i'm fixing that right now.

uberhacker: In any event, `x86` should be `i686` instead and `armv7l` should be added if it works on `aarch64`.  I'm guessing you could replace with `compatibility 'all'`?

packages/pypy3
```diff
@@ -0,0 +1,66 @@
...
+  source_url 'https://downloads.python.org/pypy/pypy3.7-v7.3.3-src.tar.bz2'
+  source_sha256 'f6c96401f76331e474cca2d14437eb3b2f68a0f27220a6dcbc537445fe9d5b78'
+
+  def self.prebuild
+			# Delete original tarball contents
+			system "rm * -rf"
```
uberhacker, line 13: Please use spaces instead of tabs for readability.

uberhacker: I'm guessing you don't need this?

packages/pypy3
```diff
@@ -0,0 +1,66 @@
...
+      # Download prebuilt tarball
+      if `uname -m` == "x86_64\n"
+          system "curl -#LO https://downloads.python.org/pypy/pypy3.7-v7.3.3-linux64.tar.bz2"
+          if `sha256sum pypy3.7-v7.3.3-linux64.tar.bz2` != "37e2804c4661c86c857d709d28c7de716b000d31e89766599fdf5a98928b7096  pypy3.7-v7.3.3-linux64.tar.bz2\n"
+              puts "Checksum mismatch. :/ Try again.".red
+              system "exit 1"
+          end
+      elsif `uname -m` == "x86\n"
```
uberhacker, line 21: This will never be true.  It should be `elsif `uname -m` == "i686\n"`.  `case ARCH ... when ... else ... end` is a bit more elegant.

packages/pypy3
```diff
@@ -0,0 +1,66 @@
...
+          system "curl -#LO https://downloads.python.org/pypy/pypy3.7-v7.3.3-linux32.tar.bz2"
+          if `sha256sum pypy3.7-v7.3.3-linux32.tar.bz2` != "7d81b8e9fcd07c067cfe2f519ab770ec62928ee8787f952cadf2d2786246efc8  pypy3.7-v7.3.3-linux32.tar.bz2\n"
+              puts "Checksum mismatch. :/ Try again.".red
+              system "exit 1"
+          end
+      elsif `uname -m` == "aarch64\n"
```
uberhacker, line 27: If `aarch64` works then `armv7l` should work also.

packages/pypy3
```diff
@@ -0,0 +1,66 @@
...
+          system "curl -#LO https://downloads.python.org/pypy/pypy3.7-v7.3.3-aarch64.tar.bz2"
```
uberhacker, line 28: Actually, this probably won't work since `aarch64` uses a 32-bit userland.

uberhacker:
Thank you for the new package @saltedcoffii.  In the future, please test before just copying and pasting from another script.  If you look at some of the other packages, it will give you a good idea of how to build correctly.

supechicken:
`pypy` => `pypy3.rb` (filename wrong)?

saltedcoffii:
> Thank you for the new package @saltedcoffii. In the future, please test before just copying and pasting from another script. If you look at some of the other packages, it will give you a good idea of how to build correctly.

I did test this, i spent a few days figuring this out. and yes, i'll take a look at other package files

uberhacker:
Don't get me wrong.  I appreciate all your effort!  I believe you will understand a bit better if you take the time to examine other packages that deal with architecture switching logic.

uberhacker:
We need to thoroughly test if this still works with python 3.9.  I'm guessing there might be some compatibility issues with earlier binary releases.

saltedcoffii:
Alright that should be a lot more readable
Currently it installs fine but the actual `pypy3` binary is broken after the install phase and I think it has something to do with this error:
```
strip: Unable to recognise the format of the input file `./usr/local/opt/pypy3/bin/pypy3'
```
Is there a way to disable chromebrew's strip function?

uberhacker:
This is just a warning.  Don't worry about it.

satmandu:
FYI, you may not have to do a extract of the pre-built binaries. You can always get someone to try to get packages built for you in a virtualized environment.

Here's a quick & dirty conversion of the arch package file which might be used for that.
```
# Adapted from Arch Linux pypy3 PKGBUILD at:
# https://github.com/archlinux/svntogit-community/raw/packages/pypy3/trunk/PKGBUILD

require 'package'

class Pypy3 < Package
  description 'A Python3 implementation written in Python, JIT enabled'
  homepage 'https://pypy.org'
  version '7.3.3'
  compatibility 'all'
  source_url 'https://downloads.python.org/pypy/pypy3.7-v7.3.3-src.zip'
  source_sha256 '9e4756903b14c5f971989a2f5a4de6ee19b21a59f2a798b3ad2ad0e71b2582a5'

  depends_on 'expat'
  depends_on 'bzip2'
  depends_on 'gdbm'
  depends_on 'openssl'
  depends_on 'libffi'
  depends_on 'zlib'
  depends_on 'ncurses'
  depends_on 'python2' => ':build'
  depends_on 'sqlite' => ':build'
  depends_on 'tk' => ':build'
  depends_on 'python2_pycparser' => ':build'

  def self.build
    Dir.chdir 'pypy/goal' do
    system "python2 ../../rpython/bin/rpython -Ojit --shared targetpypystandalone"
    # Compile binary modules
    system "PYTHONPATH=../.. ./pypy3-c ../../lib_pypy/pypy_tools/build_cffi_imports.py"
    end
  end
  
  def self.install
  # Prepare installation
  system "python2 pypy/tool/release/package.py --archive-name pypy --targetdir ."
  system "mkdir unpacked"
  system "tar xf pypy.tar.bz2 -C unpacked"
  # Install pypy
  system "mkdir -p #{CREW_DEST_PREFIX}/pypy3"
    Dir.chdir 'unpacked' do
      system "install -Dm755 pypy/bin/pypy3 #{CREW_DEST_PREFIX}/bin/pypy3"
      system "install -Dm755 pypy/bin/libpypy3-c.so #{CREW_DEST_LIB_PREFIX}/libpypy3-c.so"
      system "cp -r pypy/include pypy/lib_pypy pypy/lib-python pypy/site-packages #{CREW_DEST_PREFIX}/pypy3"
    end
  end
end
```

saltedcoffii:
I realized the mistake I made and it's almost too stupid and embarrassing to admit to it: I had the aarch64 tarball under the `when x86_64` case and *vice versa*. Which would explain why `strip` wasn't recognizing the binary formats and why the binary wouldn't execute. I'm testing the package.rb one more time just to make sure, but I'm pretty sure it works and I'm just really... h~

saltedcoffii:
now we just need to get pip working

packages/pypy3.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def self.build
+    # Remove unnecessary files
+    system "rm -vf LICENSE README.rst site-packages/README"
+  end
+
+  def self.install
+	  system "mkdir -vp #{CREW_DEST_PREFIX}/opt/pypy3/ #{CREW_DEST_PREFIX}/bin/"
+    system "mv -v * #{CREW_DEST_DIR}#{CREW_PREFIX}/opt/pypy3/"
```
supechicken666, line 28: Change to `system "mv -v * #{CREW_DEST_PREFIX}/opt/pypy3/"` ?

packages/pypy3.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def self.postinstall
+    # Upgrade pip
+    puts
+    puts "Upgrading pip...".lightblue
+    system "curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && #{CREW_DEST_PREFIX}/opt/pypy3/bin/pypy3 get-pip.py -U --no-warn-script-location"
+    puts "You now have two python interpreters installed.\nNow when you install things with pip, be sure to use \"python3 -m pip\" or \"pypy3 -m pip\" to ensure you're using the right instance of pip.".lightblue
```
supechicken666, line 36: Why not put it into 2 lines instead of using `\n`

packages/pypy3.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def self.build
+    # Remove unnecessary files
+    system "rm -vf LICENSE README.rst site-packages/README"
+  end
+
+  def self.install
+	  system "mkdir -vp #{CREW_DEST_PREFIX}/opt/pypy3/ #{CREW_DEST_PREFIX}/bin/"
+    system "mv -v * #{CREW_DEST_PREFIX}/opt/pypy3/"
```
supechicken, line 28: Rubyize with `FileUtils.mv Dir.glob('*'), "#{CREW_DEST_PREFIX}/opt/pypy3/", verbose: true`

saltedcoffii: I don't know how to rubyize but you're welcome to edit the file and do that

supechicken: Okay

packages/pypy3.rb
uberhacker: I'll take it a step further and rubyize it:
```
FileUtils.mv '.', "#{CREW_DEST_PREFIX}/opt/pypy3/"
```

saltedcoffii:
Would somebody with an aarch64 chromebook test this package and let me know if it works? I don't know if pypy's 64 bit build will work with chromeos's 32 bit userspace.

packages/pypy3.rb
```diff
@@ -0,0 +1,46 @@
...
+  
+  case ARCH
+  when 'x86_64'
+    source_url 'https://downloads.python.org/pypy/pypy3.7-v7.3.3-linux64.tar.bz2'
+    source_sha256 '37e2804c4661c86c857d709d28c7de716b000d31e89766599fdf5a98928b7096'
+  when 'i686'
+    source_url 'https://downloads.python.org/pypy/pypy3.7-v7.3.3-linux32.tar.bz2'
+    source_sha256 '7d81b8e9fcd07c067cfe2f519ab770ec62928ee8787f952cadf2d2786246efc8'
+  when 'aarch64'
+    source_url 'https://downloads.python.org/pypy/pypy3.7-v7.3.3-aarch64.tar.bz2'
```
uberhacker, line 17: This won't work on aarch64 since it operates under a 32-bit userland.

packages/pypy3.rb
```diff
@@ -0,0 +1,46 @@
+require 'package'
+
+class Pypy3 < Package
+  description 'pypy is fast, drop-in replacement of cpython that can execute code over 4 times faster.'
+  homepage 'https://www.pypy.org/'
+  version '7.3.3'
```
uberhacker, line 6: Does this version work with python 3.9?

saltedcoffii: this version is python 3.7 i believe
pypy doesn't have python 3.8 or 3.9

packages/pypy3.rb
saltedcoffii: okay i'll remove it

saltedcoffii:

PR #4932, opened 2021-01-16
Center chromebrew logo, title and subtitle

saltedcoffii:
I think it looks better, you don't have to take it.


uberhacker:
I'll take it.  Thank you!

PR #4937, opened 2021-01-18
[New Package] libostree

saltedcoffii:
builds on x86_64
I have gtk_doc and man disabled in the configure phase because ostree.1 and some gtk_doc documentation fail to build and cause make to fail, but if someone can get them working that would be great.


packages/libostree.rb
```diff
@@ -0,0 +1,34 @@
+require 'package'
+
+class Libostree < Package
+  description 'libostree manages operating system and container binary deployment and upgrades.'
+  homepage 'https://ostreedev.github.io/ostree/'
+  version '2020.8'
```
uberhacker, line 6: Missing `compatibility 'all'`?

PR #4945, opened 2021-01-20
[New Package] rpm

saltedcoffii:
builds on x86_64. had trouble with configure script and lua, if anyone can get it to work, that'd be great :)

### DIFF
--- a/packages/rpm.rb
+++ b/packages/rpm.rb
@@ -1,0 +1,42 @@
+require 'package'
+
+class Rpm < Package
+  description 'RedHat Package Manager is a free and open-source package management system and is the baseline package format of the Linux Standard Base.'
+  homepage 'https://rpm.org'
+  version '4.16.1.2'
+  compatibility 'all'
+  source_url 'http://ftp.rpm.org/releases/rpm-4.16.x/rpm-4.16.1.2.tar.bz2'
+  source_sha256 '8357329ceefc92c41687988b22198b26f74a12a9a68ac00728f934a5c4b4cacc'
+
+  depends_on 'openssl' # Could also be libgcrypt but I chose openssl because it's a base package
+  depends_on 'popt' => :build
+  depends_on 'libarchive' => :build
+  depends_on 'libdb' => :build # I'm assuming these are build dependencies because the configure script was looking for header files not libraries or binaries, I could be wrong
+
+  def self.build
+    system "./configure #{CREW_OPTIONS} \
+            --with-crypto=openssl \
+            --enable-python \
+            --enable-sqlite \
+            --without-lua" # configure script didn't like or couldn't find lua while installed
+    system "make"
+  end
+
+  def self.install
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+
+  def self.check
+    system "make", "check"
+  end
+  
+  def self.postinstall
+    puts
+    puts 'rpm is the RedHat package manager'.lightblue
+    puts 'This is useful for those who want to create/modify rpm files'.lightblue
+    puts 'However, do NOT use rpm to install rpms on your chromeos machine'.lightblue
+    puts 'This may break your system; you may have to do a full recovery by USB'.lightblue
+    puts "You've been warned!".lightblue
+    puts
+  end
+end
--- a/packages/rpm.rb
+++ b/packages/rpm.rb
@@ -11,7 +11,7 @@ class Rpm < Package
   depends_on 'openssl' # Could also be libgcrypt but I chose openssl because it's a base package
   depends_on 'popt' => :build
   depends_on 'libarchive' => :build
-  depends_on 'libdb' => :build # I'm assuming these are build dependencies because the configure script was looking for header files not libraries or binaries, I could be wrong
+  depends_on 'libdb' => :build # I'm assuming these are build dependencies because the configure script was looking for header files not libraries or binaries, I could be wrong.
 
   def self.build
     system "./configure #{CREW_OPTIONS} \
@@ -24,6 +24,11 @@ class Rpm < Package
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    FileUtils.mv "#{CREW_DEST_PREFIX}/bin/rpm", "#{CREW_DEST_PREFIX}/bin/rpm~"
+    Dir.chdir "#{CREW_DEST_PREFIX}/bin" do
+      system "curl -#LO https://raw.github.com/saltedcoffii/chromebrew-rpm-noinstall/master/rpm"
+      system "chmod +x rpm"
+    end
   end
 
   def self.check


uberhacker:
I have the same view on rpm as I do with dpkg.  It doesn't have any place in this project since crew itself is already a package manager.  It just creates confusion for users.

saltedcoffii:
I understand where you're coming from however rpm is present in [debian](https://ftp.debian.org/debian/pool/main/r/rpm/) and [ubuntu](http://archive.ubuntu.com/ubuntu/pool/universe/r/rpm/)'s repositories, dpkg is present in [fedora](https://dl.fedoraproject.org/pub/fedora/linux/releases/33/Everything/x86_64/os/Packages/d/)'s repositories, dpkg and rpm are both found in [archlinux](https://mirrors.edge.kernel.org/archlinux/community/os/x86_64/), gentoo ([dpkg](https://github.com/gentoo/gentoo/tree/master/app-arch/dpkg),[rpm](https://github.com/gentoo/gentoo/tree/master/app-arch/rpm)) and alpine ([dpkg](https://pkgs.alpinelinux.org/package/edge/main/x86_64/dpkg),[rpm](https://pkgs.alpinelinux.org/package/edge/community/x86_64/rpm)) 's repositories. I don't see why chromebrew needs to be any different.

kjperkin:
Forgive me if I'm speaking out of turn here, but I'm really curious. Would packages from another distribution (ignoring statically compiled software) even work on ChromeOS without being compiled from source? Don't the package install scripts and dependencies expect a particular filesystem layout and a particular set of system libraries? I wouldn't be surprised if *some* packages worked okay, but I'd expect most of them to be slightly to extremely broken.

uberhacker:
@kjperkin: You are mostly correct.  There are some binaries that do work but the directory locations need to be tweaked to execute correctly.  Any files that need root access to execute cannot work unless put in a writable location under the current user.

saltedcoffii:
Deb and rpm files also package scripts, some of which wouldn't be compatible with chromeos. Chromebrew handles scripts in the self.postinstall function of a package in ruby, and deb/rpm scripts are usually in bash or similar, so any translation would likely have to be manual

kjperkin:
Right. Given that in the general case, installing "foreign" packages could cause demons to come flying out of your nose, and assuming that given the opportunity, some people won't know better than to blindly install such packages, does the benefit of having these tools readily available outweigh the risk of confusion and harm?

saltedcoffii:
If rootfs verification was enabled (it usually is) than the worst damage deb or rpm could do is destroy your chromebrew packages tree and you would have to delete everything in `/usr/local` and rerun the chromebrew install script. I personally think that the postinstall message should prevent most user error as it was heavily inspired by archlinux's postinstall message for dpkg.
Perhaps dpkg and rpm could refuse to install if rootfs verification is disabled?

kjperkin:
Well, strictly speaking, the *worst* that could happen is that rpm running as root executes a package script that then does something (probably via sysfs) that damages the hardware; but I don't think that's likely. :-)

More realistically, I'd be nervous that a package might trash something under /home/chronos or some other writable part of the hierarchy. That or a novice user installs a package that writes directly to a block device.

All of that said, there's only so much hand-holding and childproofing that is reasonable, so perhaps a warning in the package description and in the post-install message is enough.

uberhacker:
_...does the benefit of having these tools readily available outweigh the risk of confusion and harm?_
I say definitely not.

satmandu:
I would lean towards being more permissive than no with tools like dpkg and rpm. If you're using chromebrew you are already a power user, as you already have your device in developer mode, so caveat emptor.

There are definitely use cases for using dpkg or rpm to extract existing packages into a build folder for building into a chromebrew package...

uberhacker:
I would argue you can do that with `alien` without needing `dpkg` or `rpm`.

satmandu:
Does alien support chromebrew packages though?

uberhacker:
Uh, that's like asking if dpkg or rpm support Chromebrew.  Not a very strong argument for justifying dpkg or rpm.

satmandu:
IIRC alien converts from rpm to dpkg. But I can see using dpkg or rpm inside a chromebrew package script to install some software packages.

I see dpkg or rpm as souped up versions of tar, and there are some chromebrew packages which instead of building from scratch are repackaging precompiled software already.

I'm no lawyer, but using dpkg or rpm inside package scripts might also be good from the perspective of reducing liability by avoiding issues with redistribution for software that isn't licensed for repackaged redistribution.

saltedcoffii:
Alright this should fix the problem. I also made something very similar for [dpkg](https://github.com/skycocker/chromebrew/pull/4910) as well.

packages/rpm.rb
```diff
@@ -24,6 +24,11 @@ def self.build
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    FileUtils.mv "#{CREW_DEST_PREFIX}/bin/rpm", "#{CREW_DEST_PREFIX}/bin/rpm~"
+    Dir.chdir "#{CREW_DEST_PREFIX}/bin" do
+      system "curl -#LO https://raw.github.com/saltedcoffii/chromebrew-rpm-noinstall/master/rpm"
```
uberhacker, line 29: We should be using a tagged release and not master.

uberhacker: Also, there should be a sha256sum check.

uberhacker: In any event, I'd rather not use this package.

uberhacker:
@satmandu: alien doesn't do any conversion.  It simply extracts the files out of the packaged build.  That's all we're basically doing with dpkg or rpm also.  No need for these packages.

saltedcoffii:
Tar and ar can't package deb and rpm files though, only dpkg and rpm can. It would be nice, not only for chromebrew development but for just regular dpkg and rpm development. You also have/allow language package managers like gem/npm/pip/cpan that chromebrew doesn't store metadata on and therefore can't manage. Chromebrew is more directed toward powerusers and developers as less experienced users will be drawn toward crouton or crostini, and I think we should provide developing tools for the developers that use chromebrew. A postinstall warning and a script wrapper should be preventative enough for chromebrew users.

kjperkin:
Since this has been quiet for a while, I wanted to offer a few thoughts:
**alien and distros that offer each other's package managers**
IMO the environments provided by major Linux distros are more similar to each other than to chromebrew. With care, you're reasonably likely to be able to get a package from one distro working on another. Chromebooks and chromebrew offer a very different environment from your typical Linux distro, so I'm not sure the same can be said here.

**npm, pip, etc.**
The key difference between these and something like dpkg is that these are written to be distribution-agnostic. npm, pip, and friends don't care much about what distro they run on and are usually written in the language of the interpreter they support.

In the same breath, I don't really understand why distros include these packages. They become out-of-date quickly and can only be (sanely) updated by installing an updated package from the distribution.

**Creating Packages on a Chromebook***
I believe it's generally the case that packages are built either on a system running the target distribution OR inside a chroot with most of the core userspace components from that system available. In that case, you may as well put the package manager in the chroot with the rest of the target system tools and libs.

**`crew` packages that repackage binaries**
If these packages have binary executables, I would lobby for them to be converted to proper `crew` packages that can be built from source. If a package can't be built from source, I don't think it belongs here. Packages that just contain shell scripts or other interpreted scripts are much more likely to be copyable as-is.

**Other thoughts**
It looks like alien can convert packages into tarballs. tar and ar can also be used to unpack debian packages, so there are multiple paths to unpack most packages if you really want to.

satmandu:
Additionally we need to be careful about licensing for binary packages... sometimes they are not licensed for repackaging...

saltedcoffii:
A nice way to get around that is to include the rpm as a source package and make the "self.build" or "self.install" phase a repackage, and that's (as far as i know) considered a repackaging for personal use even though it's automated by a script.

kjperkin:
@saltedcoffii In your PR for `dpkg` I think you mention creating packages for crostini with this because of some issue with creating them from within crostini. How often are you doing that? To me, that seems like a far more compelling use case than repackaging binaries for use in a `crew` installation. Indeed, I'd label the repackaging case an anti-pattern and say that that's a reason to _not_ have this.

saltedcoffii:
Not super frequently, but enough to be inconvenienced when I want to do it and can't. It's not just for crostini, it's for regular debian too :). And rpm for fedora.

kjperkin:
From #4910:
> I have a lot of issues with crostini, I use chromebrew installed packages for a lot of development, including compiling and creating deb and rpm files

In those cases, wouldn't you have a chroot with the userland binaries of the target environment so that you can build the packages in an environment that strongly resembles the target system? If so, then the `dpkg` or `rpm` in the chroot should do the trick.

uberhacker:
I think one important fundamental concept you guys are not understanding is the whole reason to use Chromebrew is to run natively compiled apps on ChromeOS.  There is never any need to chroot like Crouton to run apps in a separate operating system container.  This is basically what Crostini does and it is a complete waste of resources since you are basically running a Linux operating system inside another Linux operating system.  They both run the same kernel!  Why do need the extra overhead?

saltedcoffii:

Zopolis4:
I get why people don't want rpm as an additional package manager, but I do need it in order to build a package that I'm working on adding to chromebrew. Just throwing it out there.

uberhacker:
What package are you trying to build?  I highly doubt you need rpm.

Zopolis4:
keyutils-1.6.1
```shell
chronos@localhost /usr/local/keyutils-1.6.1 $ make
ldd: /usr/bin/make: No such file or directory
grep: /etc/rpm: No such file or directory
grep: /usr/lib/rpm: No such file or directory
make: Nothing to be done for 'all'.
```

uberhacker:
Maybe this would help? https://www.linuxfromscratch.org/blfs/view/svn/general/keyutils.html

kjperkin:
Based on this output, make is trying to run ldd and grep on files in
directories that don't exist. You probably need to do several things,
including:

 * Setting the PREFIX (most Makefile are written to use a variable with
that name)
 * Choosing the right make target. The fact that this Makefile is trying to
grep files in `.../rpm` makes me suspect that it's trying to do something
unnecessary for package-related purposes
 * Patching the Makefile if needed

uberhacker:
I extracted the source archive and the Makefile has a lot of rpm build stuff in it.  I think if you strip that out and set the environment variables as @kjperkin explained, that would be sufficient to get this to build for Chromebrew.

Zopolis4:
It might build without needing to modify the makefile, it should still compile and install without it. I can't test rn due to #6765, but I now know rpm isn't an absolutely required dependency. Thanks!

PR #4946, opened 2021-01-20
Trim trailing whitespace in most files

kjperkin:
## Description
It might be somewhat pedantic of me, but I like to keep things tidy. I noticed that there were a lot of files with trailing whitespace, so I took a stab at cleaning it up. I did not alter license files in any way.

## Addtional information
I didn't bother to create an issue, especially since I think there's at least a 50% chance that this PR will get about as far as a lead balloon. 😆

I don't have time to test every single package file that was changed, but I also don't think that it's necessary. A visual inspection suffices to confirm the trimmed whitespace is non-functional. That said, it might be better to split this up into a series of commits and get them merged in over time, especially if there are any concerns about the number of changes or the possibility of merge conflicts for other contributors.

### DIFF
--- a/packages/cras.rb
+++ b/packages/cras.rb
@@ -66,7 +66,7 @@ _EOF_'
           --disable-webrtc-apm \
           --enable-sse42 \
           --enable-avx \
-          --enable-avx2 
+          --enable-avx2
           --enable-fma"
       else
         system "CFLAGS='-fuse-ld=lld' ./configure #{CREW_OPTIONS} \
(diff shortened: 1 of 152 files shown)


packages/cras.rb
```diff
@@ -66,7 +66,7 @@ def self.build
           --disable-webrtc-apm \
           --enable-sse42 \
           --enable-avx \
-          --enable-avx2 
+          --enable-avx2
```
supechicken, line 69: Thanks for your pull request! Can you help us to add the missing backslash in this line? 
```shell
--enable-avx2 \
```

uberhacker: We can address this in a separate pull request.  Thank you @kjperkin!

PR #4947, opened 2021-01-20
[New Package] spidermonkey aka mozjs

saltedcoffii:


packages/spidermonkey.rb
```diff
@@ -0,0 +1,55 @@
...
+    ENV['AS'] = 'as'
+    ENV['RANLIB'] = 'ranlib'
+    ENV['NM'] = 'nm'
+    ENV['LD'] = 'ld.gold'
+    Dir.chdir 'build' do
+      system "#{CREW_PREFIX}/bin/python3 \
+            ../configure.py --prefix=/usr/local \
+            --libdir=/usr/local/lib64 \
+            --host=x86_64-cros-linux-gnu \
+            --target=x86_64-cros-linux-gnu \
```
supechicken, line 37: We cannot hardcode this for one architecture:
```shell
      system "#{CREW_PREFIX}/bin/python3 \
            ../configure.py --prefix=#{CREW_DEST_PREFIX} \
            --libdir=#{CREW_DEST_LIB_PREFIX} \
            --host=#{ARCH}-cros-linux-gnu \
            --target=#{ARCH}-cros-linux-gnu \
```

satmandu: That actually isn't going to work either.

One way to deal with that:

```
  case ARCH
  when 'aarch64', 'armv7l'
  @gnuabi = 'gnueabihf'
  when 'x86_64', 'i686'
  @gnuabi = 'gnu'
  end

     def self.build
          system "python3.9 \
            ../configure.py --prefix=#{CREW_DEST_PREFIX} \
            --libdir=#{CREW_DEST_LIB_PREFIX} \
            --host=#{ARCH}-cros-linux-#{@gnuabi} \
            --target=#{ARCH}-cros-linux-#{@gnuabi} \
```

satmandu: Oh, and don't use CREW_DEST_PREFIX until in install...

satmandu:
This is entirely my fault for not realizing this, but there already exists a proper constant to handle gnueabihf:

```
case ARCH
when 'aarch64', 'armv7l'
  CREW_BUILD = 'armv7l-cros-linux-gnueabihf'
when 'i686'
  CREW_BUILD = 'i686-cros-linux-gnu'
when 'x86_64'
end
```

So just use CREW_BUILD.

packages/spidermonkey.rb
```diff
@@ -0,0 +1,62 @@
...
+
+class Spidermonkey < Package
+  description 'SpiderMonkey is Mozilla\'s JavaScript engine written in C and C++.'
+  homepage 'https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey'
+  version '84.0'
+  compatibility 'all'
+  source_url 'https://archive.mozilla.org/pub/firefox/releases/84.0/source/firefox-84.0.source.tar.xz'
+  source_sha256 '23273ef0165b243f5d0908c38e7854d38070282c9b526e8d93b7503cd5f69138'
+
+  depends_on 'yasm'
```
uberhacker, line 11: I'm guessing this is a build dependency also?

packages/spidermonkey.rb
```diff
@@ -0,0 +1,62 @@
...
+  depends_on 'rust' => :build
+  
+  def self.prebuild
+    puts
+    puts "Building spidermonkey requires at least 6.4 GB of space available on /mnt/stateful_partition.".lightblue
+    puts "It WILL fail without this space. Press Ctrl + C to cancel.".lightblue
```
uberhacker, line 17: Why not determine free space available and abort if true?

packages/spidermonkey.rb
```diff
@@ -0,0 +1,62 @@
...
+  def self.build
+    ENV['LC_ALL'] = 'C' # https://bugs.gentoo.org/746215
+    ENV['MACH_USE_SYSTEM_PYTHON'] = "#{CREW_PREFIX}/bin/python3"
+    ENV['CC'] = 'gcc' # Defaults to clang, but configure.py complains about oudated libstdc++ when using clang
+    ENV['CXX'] = 'g++'
+    ENV['AR'] = 'ar'
+    ENV['AS'] = 'as'
+    ENV['RANLIB'] = 'ranlib'
+    ENV['NM'] = 'nm'
+    ENV['LD'] = 'ld.gold'
```
uberhacker, line 38: We shouldn't need all these environment variables I'm thinking.

satmandu: Also if gold linker is needed, it can be invoked with ```env CFLAGS='-fuse-ld=gold' CXXFLAGS='-fuse-ld=gold' ./configure...```

packages/spidermonkey.rb
```diff
@@ -0,0 +1,62 @@
...
+            --host=#{ARCH}-cros-linux-#{@gnuabi} \
+            --target=#{ARCH}-cros-linux-#{@gnuabi} \
+            --enable-application=js \
+      		  --enable-release \
+      		  --enable-optimize \
+      		  --enable-gold \
+      		  --with-intl-api \
+	      	  --with-system-icu \
+		        --enable-jit \
+		        --enable-strip" # Mozilla's configure.py doesn't recognize --mandir and --build
```
uberhacker, line 52: Can we clean up the spacing for clarity?

uberhacker:
Thank you for the new package @saltedcoffii!

packages/spidermonkey.rb
saltedcoffii: considering that yasm only provides binaries, header files and one static library, you're probably right. I'll test that to make sure and get back in an hour when it builds.

packages/spidermonkey.rb
saltedcoffii: you're the rubyist! I didn't know how to do that in ruby, and i didn't wanna use df and sed/awk

packages/spidermonkey.rb
saltedcoffii: on it

packages/spidermonkey.rb
```diff
@@ -0,0 +1,63 @@
...
+    ENV['CXX'] = 'g++'
+    ENV['AR'] = 'ar'
+    ENV['AS'] = 'as'
+    ENV['RANLIB'] = 'ranlib'
+    ENV['NM'] = 'nm'
+    Dir.chdir 'build' do
+      system "#{CREW_PREFIX}/bin/python3 \
+            ../configure.py --prefix=#{CREW_DEST_PREFIX} \
+            --libdir=#{CREW_DEST_LIB_PREFIX} \
+            --host=#{CREW_BUILD } \
```
supechicken, line 44: Extra space?
`#{CREW_BUILD }`

saltedcoffii: I literally hate my IDE, i wish chromeos had a better one

packages/spidermonkey.rb
satmandu: Something like this might show disk free: https://stackoverflow.com/a/37988177

satmandu (commented):
The CREW_BUILD constants are in /usr/local/lib/crew/lib/const.rb and thus don't need to be defined again.

packages/spidermonkey.rb
uberhacker: What is your IDE?  Try codium if you haven't already.

saltedcoffii: https://chrome.google.com/webstore/detail/code-pad-text-editor/adaepfiocmagdimjecpifghcgfjlfmkh
this is my IDE, just make sure to go into the extension settings and disable its access to https://google-analytics.com

And does codium work on a chromebook?

packages/spidermonkey.rb
saltedcoffii: looks like yasm is a build dep, i'll make it so

packages/spidermonkey.rb
uberhacker: Try `crew install codium` and find out.

satmandu:
FYI, I added a PR for a js78 package, which is an older version of this required by gnome packages.

This is such a painful package to build.  Please make sure you have this working locally before you submit the package... :)

saltedcoffii:
the version 78 could be called spidermonkey_esr, at it is the esr version

packages/spidermonkey.rb
supechicken: Just found that the [`android_studio`](https://github.com/skycocker/chromebrew/blob/master/packages/android_studio.rb) package is doing something like this

satmandu:
> the version 78 could be called spidermonkey_esr, at it is the esr version

I've just called it [js78](https://github.com/skycocker/chromebrew/blob/master/packages/js78.rb), as other distributions label it by year.

eg. Arch's package is js78, and debian's is mozjs-78.

saltedcoffii:
Oh i pulled from gentoo which calls it spidermonkey

saltedcoffii:
Now that we have js78, I don't think we need js84. Unless anyone sees a reason why I think we should close this pr.

satmandu:
If this is useful later, we can always reopen this. Thanks for the work you put into this.

PR #4954, opened 2021-01-20
Adjust aria2 package to use CREW_BUILD

satmandu:
- remove redundant variables and just use CREW_BUILD

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/aria2.rb
+++ b/packages/aria2.rb
@@ -22,12 +22,7 @@ class Aria2 < Package
       x86_64: '2a7c0fc6c91b8b4217d7b69b0fa607a963ecbf17c689008c989841c89a610160',
   })
 
-  depends_on 'c_ares'
   depends_on 'libgcrypt'
-  depends_on 'libssh2'
-  depends_on 'libxml2'
-  depends_on 'sqlite'
-  depends_on 'zlibpkg'
   
   def self.build
     system "env CFLAGS='-fuse-ld=gold -flto' CXXFLAGS='-fuse-ld=gold -flto' \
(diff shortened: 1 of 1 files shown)


packages/aria2.rb
```diff
@@ -28,13 +28,6 @@ class Aria2 < Package
   depends_on 'libxml2'
   depends_on 'sqlite'
   depends_on 'zlibpkg'
```
uberhacker, line 30: Are c_ares, libssh2, libxml2, sqlite and zlibpkg core packages?  If so, no need to list here.

packages/aria2.rb
```diff
@@ -47,6 +40,6 @@ def self.build
 
   def self.install
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
-    FileUtils.ln_s "#{CREW_PREFIX}/bin/#{ARCH}-cros-linux-#{@gnuabi}-aria2c", "#{CREW_DEST_PREFIX}/bin/aria2c"
+    FileUtils.ln_s "#{CREW_PREFIX}/bin/#{CREW_BUILD}-aria2c", "#{CREW_DEST_PREFIX}/bin/aria2c"
```
uberhacker, line 43: Do we need new binaries?

satmandu: Nope. This is is just a simplification of the package file, with no changes to binaries.

uberhacker: Great, just wanted to verify.

packages/aria2.rb
satmandu: Core packages are defined by anything build_essential pulls in or are they the packages pullied in by install.sh before crew runs?

In any case, which core package list should we check this against?

uberhacker: https://github.com/skycocker/chromebrew/blob/master/tools/core_packages.txt

satmandu: Thanks for that. Taken care of.

PR #4958, opened 2021-01-21
crew: Simplify syntax and use fileutils

supechicken:
Tested on `x86_64`

### DIFF
--- a/crew
+++ b/crew
@@ -134,7 +134,7 @@ end
 def list_packages
   Find.find (CREW_LIB_PATH + 'packages') do |filename|
     if File.extname(filename) == '.rb'
-      print_package filename, false
+      print_package filename
     end
   end
 end
--- a/crew
+++ b/crew
@@ -14,7 +14,7 @@ require_relative 'lib/const'
 require_relative 'lib/util'
 
 # Add lib to LOAD_PATH
-$LOAD_PATH.unshift "#{CREW_PREFIX}/lib/crew/lib"
+$LOAD_PATH.unshift "#{CREW_LIB_PATH}lib"
 
 DOC = <<DOCOPT
 Chromebrew - Package manager for Chrome OS http://skycocker.github.io/chromebrew/
--- a/crew
+++ b/crew
@@ -14,7 +14,7 @@ require_relative 'lib/const'
 require_relative 'lib/util'
 
 # Add lib to LOAD_PATH
-$LOAD_PATH.unshift "#{CREW_LIB_PATH}lib"
+$LOAD_PATH.unshift "#{CREW_PREFIX}/lib/crew/lib"
 
 DOC = <<DOCOPT
 Chromebrew - Package manager for Chrome OS http://skycocker.github.io/chromebrew/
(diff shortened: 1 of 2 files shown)


crew
```diff
@@ -14,7 +14,7 @@ require_relative 'lib/const'
 require_relative 'lib/util'
 
 # Add lib to LOAD_PATH
-$LOAD_PATH.unshift "#{CREW_LIB_PATH}lib"
+$LOAD_PATH.unshift "#{CREW_PREFIX}/lib/crew/lib"
```
uberhacker, line 17: Why remove a constant already defined in `lib/const.rb`?

supechicken: `LOAD_PATH` don't works when I put `crew` in `#{CREW_PREFIX}/bin`, either the original and `"#{CREW_PREFIX}/lib/crew/lib"`. Solved when put it in `/usr/local/lib/crew/`.

supechicken: Changed back

crew
```diff
@@ -575,54 +575,42 @@ end
...
     @pkg.build
     @pkg.in_build = false
     # wipe crew destdir
-    if @opt_verbose then
-      system "rm -rvf #{CREW_DEST_DIR}/*"
-    else
-      system "rm -rf #{CREW_DEST_DIR}/*"
-    end
-    puts "Preconfiguring package..."
+    "FileUtils.rm_rf Dir.glob('#{CREW_DEST_DIR}/*')#{@fileutils_verbose}"
```
uberhacker, line 593: Why is this enclosed in double quotes?  Shouldn't it be like below?
```
FileUtils.rm_rf Dir.glob('#{CREW_DEST_DIR}/*')#{@fileutils_verbose}
```

crew
```diff
@@ -877,22 +843,16 @@ def resolve_dependencies_and_build
...
     unless @opt_keep
-      Dir.chdir CREW_BREW_DIR do
-        if @opt_verbose then
-          system "rm -rvf *"
-        else
-          system "rm -rf *"
-        end
-        system "mkdir dest" #this is a little ugly, feel free to find a better way
-      end
+      "FileUtils.rm_rf Dir.glob('#{CREW_BREW_DIR}/*')#{@fileutils_verbose}"
```
uberhacker, line 846: Change to?
```
FileUtils.rm_rf Dir.glob("#{CREW_BREW_DIR}/*") + @fileutils_verbose
```

satmandu:
This is great! thanks!

crew
supechicken: Done

crew
```diff
@@ -991,7 +943,7 @@ def build_command (args)
   args["<name>"].each do |name|
     @pkgName = name
     search @pkgName
-    print_current_package
+    print_current_package true
```
uberhacker, line 946: This should be `print_current_package @opt_verbose` instead.

supechicken666: Done

crew
```diff
@@ -1037,7 +989,7 @@ def install_command (args)
...
```
uberhacker, line 992: `print_current_package @opt_verbose`?

supechicken: The `true` option is used to show the version and homepage

uberhacker: I know but don't you think it should be optional?

supechicken: I think the `version` should show in first install (just like crew 1.5.5), maybe making a new option called `@info` ?

uberhacker: I'm not necessarily in agreement.  What do you think @satmandu?

satmandu: I don't think that showing version and homepage during a package install makes it too chatty. But I am biased because when I do installs during package testing it is really useful to have that info, without having the lest of the verbose info being dumped onto the screen.

uberhacker: So does that mean you are not in favor of keeping it optional?

satmandu: Since there are differing opinions on what level of verbosity is warranted, should  opt_verbose not be a binary option but instead a level?

PR #4962, opened 2021-01-22
fontconfig -> 2.13.93

satmandu:
- use meson, lto

Builds properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/fontconfig.rb
+++ b/packages/fontconfig.rb
@@ -3,22 +3,23 @@ require 'package'
 class Fontconfig < Package
   description 'Fontconfig is a library for configuring and customizing font access.'
   homepage 'https://www.freedesktop.org/software/fontconfig/front.html'
-  version '2.13.92'
+  @_ver = '2.13.93'
+  version @_ver
   compatibility 'all'
-  source_url 'https://www.freedesktop.org/software/fontconfig/release/fontconfig-2.13.92.tar.xz'
-  source_sha256 '506e61283878c1726550bc94f2af26168f1e9f2106eac77eaaf0b2cdfad66e4e'
+  source_url "https://github.com/freedesktop/fontconfig/archive/#{@_ver}.tar.gz"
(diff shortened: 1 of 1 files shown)


packages/fontconfig.rb
```diff
@@ -3,36 +3,55 @@
...
+  depends_on 'graphite' => :build
+  # Remove freetype and fontconfig before rebuilding this package.
+
+  def self.patch
+  # Fix failure from font directories not being writable.
+    system "cat <<'EOF'> install-cache.py
+#!/usr/bin/env python3
+import sys
+sys.exit() 
+EOF"
```
uberhacker, line 39: This qualifies as a build step rather than a patch.  Shouldn't this be installed in the `def self.install` section also?

packages/fontconfig.rb
```diff
@@ -3,36 +3,55 @@
...
   depends_on 'expat'
   depends_on 'gperf'
   depends_on 'freetype_sub'
   depends_on 'jsonc'
   depends_on 'util_linux'
+  depends_on 'graphite' => :build
+  # Remove freetype and fontconfig before rebuilding this package.
+
+  def self.patch
+  # Fix failure from font directories not being writable.
```
supechicken, line 34: Does this work?
```ruby
system "echo 'raise SystemExit' > install-cache.py"
```

satmandu: It appears to...

packages/fontconfig.rb
satmandu: I'll move it.

packages/fontconfig.rb
satmandu: Let me check...

packages/fontconfig.rb
uberhacker: I'm confused.  Is this supposed to be part of the package?  If not, how is this even used in the build process?

satmandu:
Install cache is run by various install scripts. This makes those calls
NOOPs.

PR #4960, opened 2021-01-22
Add zig package

ThatGeekyWeeb:
[Zig](https://ziglang.org) is a relatively new programming lang, that competes with C rather than being built on it.  
Zig (`zig cc`/`c++`) can be used as a drop-in replacement for `gcc` for `g++` (C/C++) and produces smaller binaries at the cost of double the compilation of that of `gcc`.  

---

I'll be using this in an upcoming project as it produces smaller binaries than `gcc`, but my Chromebooks recently bit the dust, and this requires testing, though is should build fine as this package script was generated based on the template from xbps-src, where there are also 0 patches.

PR #4775, opened 2020-12-19
llvm -> 11.1.0

satmandu:
Update to 11.1.0

- adds llvm gold plugin
- adds tools like FileCheck compiles complain about
- Major simplification of build setup using suggested LLVM best practices.

The relevant tagging of the 11.1.0rc series [has commenced](https://github.com/llvm/llvm-project/tree/release/11.x) which is ABI compatible with 10.x & 12.0.0.(11.0.x is NOT ABI compatible with either.)

Compiles & works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

PR #4964, opened 2021-01-22
Revert "Add zig package"

ThatGeekyWeeb:
Reverts skycocker/chromebrew#4960  
As @satmandu pointed out zig actually fails to compile with llvm10 on version 0.7.1  
I tested using llvm11 and it built without issues  
0.6.0 should build fine with llvm10, but I think this should be reverted and put on hold till #4775 gets merged


uberhacker:
#4775 merged.

PR #4966, opened 2021-01-22
gtk_doc -> 1.33.2, use meson, lto

satmandu:
- switched to using meson
- removed patches, but left in the section for guidance if one wants to see how the Void patches were implemented here.

Builds properly:
- [x] x86_64
- [x] i686
- [x] armv7l


uberhacker:
No patches.  I like it!

PR #4973, opened 2021-01-23
Comment LLVM_DEFAULT_TARGET_TRIPLE lines

### LINKED ISSUE #4972: [llvm] uninitialized constant
```
/usr/local/lib/crew/packages/llvm.rb:42:in `<class:Llvm>': uninitialized constant Llvm::LLVM_DEFAULT_TARGET_TRIPLE (NameError)
```

uberhacker:
Fixes #4972.

### DIFF
--- a/packages/llvm.rb
+++ b/packages/llvm.rb
@@ -39,9 +39,9 @@ class Llvm < Package
   when 'aarch64','armv7l'
     #LLVM_TARGETS_TO_BUILD = 'ARM;AArch64;AMDGPU'
     LLVM_TARGETS_TO_BUILD = 'all'
-    #@ARCH_C_FLAGS = "-fPIC -march=armv7-a -mfloat-abi=hard -ccc-gcc-name #{LLVM_DEFAULT_TARGET_TRIPLE}"
-    #@ARCH_CXX_FLAGS = "-fPIC -march=armv7-a -mfloat-abi=hard -ccc-gcc-name #{LLVM_DEFAULT_TARGET_TRIPLE}"
-    @ARCH_LDFLAGS=''
+    @ARCH_C_FLAGS = "-fPIC -march=armv7-a -mfloat-abi=hard -ccc-gcc-name #{CREW_BUILD}"
+    @ARCH_CXX_FLAGS = "-fPIC -march=armv7-a -mfloat-abi=hard -ccc-gcc-name #{CREW_BUILD}"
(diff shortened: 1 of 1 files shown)


packages/llvm.rb
```diff
@@ -39,8 +39,8 @@ class Llvm < Package
   when 'aarch64','armv7l'
     #LLVM_TARGETS_TO_BUILD = 'ARM;AArch64;AMDGPU'
     LLVM_TARGETS_TO_BUILD = 'all'
-    @ARCH_C_FLAGS = "-fPIC -march=armv7-a -mfloat-abi=hard -ccc-gcc-name #{LLVM_DEFAULT_TARGET_TRIPLE}"
-    @ARCH_CXX_FLAGS = "-fPIC -march=armv7-a -mfloat-abi=hard -ccc-gcc-name #{LLVM_DEFAULT_TARGET_TRIPLE}"
+    #@ARCH_C_FLAGS = "-fPIC -march=armv7-a -mfloat-abi=hard -ccc-gcc-name #{LLVM_DEFAULT_TARGET_TRIPLE}"
+    #@ARCH_CXX_FLAGS = "-fPIC -march=armv7-a -mfloat-abi=hard -ccc-gcc-name #{LLVM_DEFAULT_TARGET_TRIPLE}"
```
supechicken, line 43: Add two lines below ? (as it may cause errors in setting the `@ARCH_C_LTO_FLAGS` variable below if these variables not set)
```ruby
@ARCH_C_FLAGS = ' '
@ARCH_CXX_FLAGS = ' '
```

uberhacker: This is just a quick fix to get rid of the error.  Since this is a core package, users who try to install fresh or upgrade will experience a crash...not a good experience.

satmandu: I'm confused. Why does ``` @ARCH_C_FLAGS = "-fPIC -march=armv7-a -mfloat-abi=hard -ccc-gcc-name #{LLVM_DEFAULT_TARGET_TRIPLE}"``` not work?

Isn't this literally how we specify ```@_ver = 'something'``` ?

satmandu: Ah I see. @supechicken 's fix appears to be the correct one.

satmandu: The actual correct solution is to replace ```#{LLVM_DEFAULT_TARGET_TRIPLE}``` with ```#{CREW_BUILD}```

PR #4989, opened 2021-01-24
[New Package] rlottie

saltedcoffii:
a lottie processing library from samsung.

### DIFF
--- a/packages/rlottie.rb
+++ b/packages/rlottie.rb
@@ -3,18 +3,32 @@ require 'package'
 class Rlottie < Package
   description 'rlottie is a platform independent standalone c++ library for rendering vector based animations and art in realtime.'
   homepage 'https://github.com/Samsung/rlottie'
-  version '0.2'
+  @_ver = '0.2'
+  version @_ver
   compatibility 'all'
-  source_url 'https://github.com/Samsung/rlottie/archive/v0.2.tar.gz'
+  source_url "https://github.com/Samsung/rlottie/archive/v#{@_ver}.tar.gz"
   source_sha256 '030ccbc270f144b4f3519fb3b86e20dd79fb48d5d55e57f950f12bab9b65216a'
+
+  binary_url ({
+     aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/rlottie-0.2-chromeos-armv7l.tar.xz',
+      armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/rlottie-0.2-chromeos-armv7l.tar.xz',
+        i686: 'https://dl.bintray.com/chromebrew/chromebrew/rlottie-0.2-chromeos-i686.tar.xz',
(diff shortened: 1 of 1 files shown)


satmandu:
This package supports meson. Why not just try that?

Here's how one might mesonize that:
```
require 'package'

class Rlottie < Package
  description 'rlottie is a platform independent standalone c++ library for rendering vector based animations and art in realtime.'
  homepage 'https://github.com/Samsung/rlottie'
  version '0.2'
  compatibility 'all'
  source_url 'https://github.com/Samsung/rlottie/archive/v0.2.tar.gz'
  source_sha256 '030ccbc270f144b4f3519fb3b86e20dd79fb48d5d55e57f950f12bab9b65216a'

  def self.build
    system "meson #{CREW_MESON_LTO_OPTIONS} builddir"
    system "meson configure builddir"
    system "ninja -C builddir"
  end

  def self.install
  end
end
```
(I haven't tested this...)

Also check the meson_options.txt to see if there are any useful options.

Meson makes these installs so much easier!

saltedcoffii:
Alright, I think its ready to merge

satmandu:
That ```ninja -C build``` isn't a duplicate command.
ninja actually does the compile!
The ```meson configure build``` line before that just tells you what options are being used to build the package.

It's really useful to see what's being used.

satmandu:
Also literally just ```system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"``` is sufficient for the install...

(I'm using ```ninja -C builddir``` and not ```ninja -C build``` because many packages have a build folder with stuff in it, so this way we can use the same code block for multiple packages and don't have to worry about clobbering an existing ```build``` folder.)

saltedcoffii:
Alright, i fixed and tested the rb. I followed the guide on the [mesonbuild website](https://mesonbuild.com/Quick-guide.html#using-meson-as-a-distro-packager)

satmandu:
One of the advantages of using ```ninja -C builddir``` vs ```meson compile -C builddir``` is that (as I understand it) ninja automatically uses nproc to use all processors when compiling, whereas meson does not. I could be wrong on this.

Meson may automatically use ninja in any case, but see this: https://mesonbuild.com/Running-Meson.html#building-directly-with-ninja

But in any case you don't need that ```env``` in the install line as per the link you mentioned earlier. 

I'd suggest uniformity across packages as much as possible just so if at some point we want to use more macros we can.

Any thoughts @uberhacker ?

saltedcoffii:
Sure, feel free to make the necessary changes!

satmandu:
@saltedcoffii built for all 3 architectures. Can you test to see if it works properly?

saltedcoffii:
The binary and source install just fine and `crew files rlottie` shows all the files it should on x86_64. I can't test the other architectures.

uberhacker:
@satmandu: I prefer to keep things as simple as possible and stay on the beaten path. :)

PR #4992, opened 2021-01-24
[New Package] conan

saltedcoffii:
pip installs a lot of dependencies into destdir along with conan in self.install. Do these need to first be packaged individually?

### DIFF
--- a/packages/conan.rb
+++ b/packages/conan.rb
@@ -1,0 +1,20 @@
+require 'package'
+
+class Conan < Package
+  description 'conan is a open-source C/C++ package manager'
+  homepage 'https://conan.io'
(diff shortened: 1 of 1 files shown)


uberhacker:
We have to be careful not to create packages just for the sake of creating them.  See https://github.com/skycocker/chromebrew/wiki/FAQ.
```
Q: A Chromebrew package is not available for the software I need. What can I do?

A: Before you attempt to add a new package, consider there are many apps installable via gem, npm, pip or composer. A fairly comprehensive list of command-line apps can be found here.
```

saltedcoffii:
Alright, sounds good.

PR #4996, opened 2021-01-25
Add option to enable/disable colorful output

kjperkin:
## Description
Sometimes colorful output is not desirable. Other tools will disable color by default if they detect that standard output is not a tty. This is an attempt to implement the same behavior for `crew`.

I honestly have next to no idea what I'm doing. I'm so new to ruby I'm not even a novice. It's just a lot like python so I've been scraping by... 😆 

## Design considerations
 * `-d` was chosen as the short option for `--no-color` because `d` comes after `c` which is easy to remember.
 * `--color` and `--no-color` where chosen instead of an option like `--color=(always|never|auto)` because no other `crew` option takes arguments.
 * Since the colorization methods are implemented on `String`, a class instance variable seemed like the right way to have a global on/off toggle for colorized output by `crew`.
 * Repeating `[-c|--color] [-d|--no-color]` on every line of the help was super gross, so I added `[options]` to the sub-commands and removed `[--verbose]` from each help line. As described [here](https://github.com/docopt/docopt#help-message-format), any option not listed explicitly for a sub-command is allowed for `[options]`. I think this makes intuitive sense for color and verbosity even if they don't have an effect for all commands.

## Additional information
* This is a draft because I think it needs discussion and additional thought before being implemented. I really just slapped this together thinking it might be a good starting point for discussion. 

Works properly:
- [x] x86_64
- [ ] aarch64 (no test env. available)


crew
```diff
@@ -76,6 +78,7 @@ rescue Docopt::Exit => e
   exit 1
 end
 
+String.use_color = args["--color"] || !args["--no-color"]
```
uberhacker, line 81: It would be easier to use an `@` variable.

lib/color.rb
```diff
@@ -1,7 +1,17 @@
...
 class String
+  @use_color = STDOUT.isatty
+
+  class << self
+    attr_accessor :use_color
+  end
+
   def colorize(color_code, shade)
-    "\e[#{shade};#{color_code}m#{self}\e[0m"
+    if self.class.use_color
```
uberhacker, line 10: It would be easier to use an `@` variable.

lib/color.rb
```diff
@@ -1,7 +1,17 @@
...
+      return "\e[#{shade};#{color_code}m#{self}\e[0m"
+    else
+      return self
```
uberhacker, line 13: Is this necessary?  What would happen if it was ignored?

lib/color.rb
```diff
@@ -1,7 +1,17 @@
...
+  @use_color = STDOUT.isatty
+
+  class << self
+    attr_accessor :use_color
+  end
+
   def colorize(color_code, shade)
-    "\e[#{shade};#{color_code}m#{self}\e[0m"
+    if self.class.use_color
+      return "\e[#{shade};#{color_code}m#{self}\e[0m"
```
uberhacker, line 11: Where are these variables defined?

crew
```diff
@@ -20,25 +20,27 @@ DOC = <<DOCOPT
...
-  crew whatprovides <name> ...
+  crew install [options] [-k|--keep] [-s|--build-from-source] [-S|--recursive-build] <name> ...
+  crew list [options] (available|installed|compatible|incompatible)
+  crew postinstall [options] <name> ...
+  crew reinstall [options] [-k|--keep] [-s|--build-from-source] [-S|--recursive-build] <name> ...
+  crew remove [options] <name> ...
+  crew search [options] [<name> ...]
+  crew update [options]
+  crew upgrade [options] [-k|--keep] [-s|--build-from-source] [<name> ...]
+  crew whatprovides [options] <name> ...
```
uberhacker, line 36: I'm not understanding how `[options]` takes care of `[-v|--verbose]`.

uberhacker:
I like what you are trying to do here @kjperkin!

lib/color.rb
kjperkin: I had two reasons for doing it this way:
1) My thinking was that `String#colorize` shouldn't need any knowledge outside of the class in which it is defined. I didn't want to have to pass an additional parameter to every coloring method, so a class instance variable seemed to fit the bill perfectly.
2) I don't really understand variable scope in ruby. So far I've just been mimicking nearby code, so I wasn't sure how to use an `@` variable and have it be visible in crew.rb. I really need to take a weekend and start learning ruby for real 😁

lib/color.rb
kjperkin: I think so. This is the heart of the functionality. The on/off switch for colorization is a class instance variable, so it _can't_ be ignored by the colorization functions. If someone explicitly outputs a colorizing escape sequence in crew or a package.rb, I'd call that a defect. If build output for a package outputs color, I'm not sure that there's much we can do to prevent that...

crew
kjperkin: In the release notes for the python version of docopt, there's a note about `[options]` that seems to hold true for the ruby version as well:

> New in version 0.6.0:
> ...
> * Meaning of [options] shortcut slightly changed. Previously it meant "any known option". Now it means "any option not in usage-pattern". This avoids the situation when an option is allowed to be repeated unintentionally.

lib/color.rb
uberhacker: The `@` variable makes it so you don't need to pass the variable.  It's basically a global scope within the class.

uberhacker: OK, I think I understand what is happening here.  You cannot use the `@` variable since the scope is from another class.

uberhacker:
I'm going to go ahead and merge this.  If we run into issues, we can always fix it.

PR #4999, opened 2021-01-25
Add vte

satmandu:
- this is a gnome terminal dep
 
Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/vte.rb
+++ b/packages/vte.rb
@@ -22,7 +22,6 @@ class Vte < Package
   })
 
   depends_on 'gobject_introspection' => ':build'
-  depends_on 'icu4c'
   depends_on 'fribidi'
   depends_on 'gtk3'
   depends_on 'gtk4'
--- a/packages/vte.rb
+++ b/packages/vte.rb
@@ -2,6 +2,7 @@ require 'package'
 
 class Vte < Package
   description 'Virtual Terminal Emulator widget for use with GTK'
+  homepage 'https://wiki.gnome.org/Apps/Terminal/VTE'
   version '0.62.2'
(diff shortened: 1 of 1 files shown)


packages/vte.rb
```diff
@@ -0,0 +1,59 @@
+require 'package'
+
+class Vte < Package
+  description 'Virtual Terminal Emulator widget for use with GTK'
```
uberhacker, line 4: Missing homepage?

packages/vte.rb
```diff
@@ -0,0 +1,59 @@
...
+  })
+  binary_sha256 ({
+     aarch64: '948e60479e834c4cf9c3fdebab5795faf64c38122d4662f4c355b1068078f743',
+      armv7l: '948e60479e834c4cf9c3fdebab5795faf64c38122d4662f4c355b1068078f743',
+        i686: '88d7d0d1a8b3e8b86a8ff3c66aa6fc7b1ae8d57712d1659c674bc3b129fd2fcd',
+      x86_64: 'f261f381d4ff58761907737037ed60ddafff5444067b38fb28e45f02f04ed6aa',
+  })
+
+  depends_on 'gobject_introspection' => ':build'
+  depends_on 'icu4c'
```
uberhacker, line 24: This is a core package so no need to list here.

packages/vte.rb
satmandu: fixed

packages/vte.rb
satmandu: fixed

PR #5000, opened 2021-01-25
Add gnome terminal

satmandu:
Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/gnome_terminal.rb
+++ b/packages/gnome_terminal.rb
@@ -1,0 +1,44 @@
+require 'package'
+
+class Gnome_terminal < Package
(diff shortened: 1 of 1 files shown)


packages/gnome_terminal.rb
```diff
@@ -0,0 +1,44 @@
+require 'package'
+
+class Gnome_terminal < Package
+  description 'The GNOME Terminal Emulator'
```
uberhacker, line 4: Missing homepage?

satmandu: ah yup.

PR #5004, opened 2021-01-25
Fix sommelier 20210125

satmandu:
Sommelier was refusing to build for me with an aura error.
Using [RC1 for Wayland 1.19 (version 1.18.93) ](https://lists.freedesktop.org/archives/wayland-devel/2021-January/041691.html) fixes this issue.

- This is updated to the last version of sommelier which doesn't rejigger the virtwl logic. This build also uses an internal copy of virtwl.h.
- We will have to deal with the newer breakage at some point, or find a better proxy between Xwayland and the exo wayland server. The new logic doesn't let us set the virtwl device to ```/dev/null```. If ```/dev/wl0``` doesn't exist, sommelier fails with newer commits.
- Disabled the wayland proxy functionality. The wayland server in M87 is "good enough" and this way we can avoid additional breakage from things like sleep for wayland processes.

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -1,7 +1,7 @@
 require 'package'
 
 class Sommelier < Package
-  description 'Sommelier works by redirecting X11 and Wayland programs to the built-in ChromeOS wayland server.'
+  description 'Sommelier works by redirecting X11 programs to the built-in ChromeOS Exo Wayland server.'
   homepage 'https://chromium.googlesource.com/chromiumos/platform2/+/HEAD/vm_tools/sommelier/'
   version '20210109'
   compatibility 'all'
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -21,8 +21,6 @@ class Sommelier < Package
       x86_64: '5348afbfceeed2ddf5078d4d325f50ad736ea89c8249db263009c3d23cd05db9',
   })
 
-
-
   depends_on 'mesa'
   depends_on 'xkbcomp'
   depends_on 'xorg_server' unless File.exists? "#{CREW_PREFIX}/bin/Xwayland.elf"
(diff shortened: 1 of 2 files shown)


packages/sommelier.rb
```diff
@@ -44,46 +46,49 @@ class Sommelier < Package
...
+    #sha256sum_virtwl = 'a8215f4946ccf30cbd61fcf2ecc4edfe6d05bffeee0bacadd910455274955446'
+    
+    #puts "Downloading virtwl".yellow
+    #system('curl', '-s', '-C', '-', '--insecure', '-L', '-#', url_virtwl, '-o', filename_virtwl)
+    #abort 'Checksum mismatch. :/ Try again.'.lightred unless
+      #Digest::SHA256.hexdigest( File.read( filename_virtwl ) ) == sha256sum_virtwl
+    #puts "virtwl base64 downloaded".lightgreen
+    #FileUtils.mkdir_p 'build/linux'
+    #system 'base64 --decode virtwl.h_base64 > build/linux/virtwl.h'   
+
```
uberhacker, line 68: No sha265sum check?

packages/sommelier.rb
```diff
@@ -109,11 +114,11 @@ def self.build
...
   fi
 fi
-EOF"
-
+EOF"     
+      
       #Create local startup and shutdown scripts
-
-      # sommelier_sh
+
```
uberhacker, line 120: Please remove trailing whitespace.

packages/sommelier.rb
satmandu: The question is how?

The Google source servers seem to tar on the fly, which makes  for nonreproducible archives which don't have consistent sha256 sums.

satmandu: In case it isn't obvious, sommelier as of this commit is moving BACK to including virtwl.h, so no additional checksumming of that file is needed.

uberhacker:
This package is turning into a hot mess. :)

satmandu:
Tell me about it. This is why I'm desperately looking for a way to avoid it. Literally, all we need is Xwayland to not crap out when running with the Exo wayland server, and we can dispense with Sommelier.

PR #5014, opened 2021-01-26
Remove the libtool warnings while 'make install'

supechicken:
Tested on `x86_64`

### DIFF
--- a/crew
+++ b/crew
@@ -18,7 +18,6 @@ $LOAD_PATH.unshift "#{CREW_LIB_PATH}lib"
 
 DOC = <<DOCOPT
 Chromebrew - Package manager for Chrome OS http://skycocker.github.io/chromebrew/
-
 Usage:
   crew build [options] [-k|--keep] <name> ...
   crew const [options] [<name> ...]
@@ -34,7 +33,6 @@ Usage:
   crew update [options]
   crew upgrade [options] [-k|--keep] [-s|--build-from-source] [<name> ...]
   crew whatprovides [options] <name> ...
-
   -k --keep               Keep the `CREW_BREW_DIR` (#{CREW_BREW_DIR}) directory.
   -s --build-from-source  Build from source even if pre-compiled binary exists.
   -S --recursive-build    Build from source, including all dependencies, even if pre-compiled binaries exist.
@@ -43,7 +41,6 @@ Usage:
   -d --no-color           Disable colors even if standard out is a tty.
   -v --verbose            Show extra information.
   -h --help               Show this screen.
-
 version #{CREW_VERSION}
 DOCOPT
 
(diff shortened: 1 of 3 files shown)


satmandu:
Maybe I'm just tired... the libtool.txt is processed and libtool is run automatically now?

crew
```diff
@@ -619,6 +616,13 @@ def post_install (dest_dir)
     puts 'Performing post-install...'
     @pkg.postinstall
   end
+  # Run libtool warning commands
+  @warning = 'libtool: warning: remember to run '
+      @out = "#{CREW_BREW_DIR}/libtool.txt"
+  if File.exist? "#{CREW_BREW_DIR}/libtool.txt" then
```
uberhacker, line 622: `if File.exist? @out then`?  Can we fix the alignment on line 621 also?

crew
```diff
@@ -619,6 +616,13 @@ def post_install (dest_dir)
...
+    system "grep -oP \"(?<=#{@warning}\').*?(?=\')\" #{@out} | bash"
+    FileUtils.rm "#{CREW_BREW_DIR}/libtool.txt"
```
uberhacker, line 624: `FileUtils.rm @out`?  Also, I don't see where `libtool` is actually being executed.  Did I miss something?

packages/buildessential.rb
```diff
@@ -9,6 +9,9 @@ class Buildessential < Package
   is_fake
 
 
+  # For libtool patching
+  depends_on 'grep'
```
uberhacker, line 13: I don't think this is necessary.  We already have `grep` installed.

lib/package.rb
```diff
@@ -154,4 +154,12 @@ def self.system(*args)
...
     raise InstallError.new("`#{args.join(" ")}` exited with #{exitstatus}") unless exitstatus == 0
   end
+  def self.make_install (destdir = CREW_DEST_PREFIX, opt = '')
+    @warning = 'libtool: warning: remember to run '
+        @out = "#{CREW_BREW_DIR}/libtool.txt"
+
+    system "make DESTDIR=#{destdir} install #{opt} 2>&1 |\
+            tee #{@out} |\
+            grep -v \"#{@warning}\""
+  end
```
uberhacker, line 164: This is just removing the message.  This does nothing to solve the problem.

uberhacker:
@satmandu: You are not tired.  This is a pointless pull request.

crew
satmandu: I think you definitely need to have a place here where libtool is called. But if libtool is called properly, we shouldn't get these error messages in the first place, right?

uberhacker:
@supechicken: Please resubmit with an actual call to `libtool`.

lib/package.rb
supechicken: This will put the original output to the libtool.txt file (the warning will save and execute in `crew`) while deleting the warning on the screen.

crew
supechicken: @satmandu I tried to remove `libtool` but the warnings are still here

PR #5022, opened 2021-01-27
[gparted] Add xhost dependency

supechicken:
The `xhost` dependency is needed for allowing `root` to access the x server

Tested on `x86_64`

### DIFF
--- a/packages/gparted.rb
+++ b/packages/gparted.rb
@@ -44,17 +44,18 @@ class Gparted < Package
       --enable-libparted-dmraid \
       --enable-xhost-root"
     system "make"
-    system "cat <<'EOF'> gparted
-#!/bin/bash
-DISPLAY=:0
-xhost si:localuser:root
-sudo gparted.orig
-EOF"
   end
 
   def self.install
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
     FileUtils.mv "#{CREW_DEST_PREFIX}/bin/gparted", "#{CREW_DEST_PREFIX}/bin/gparted.orig"
-    FileUtils.install 'gparted', "#{CREW_DEST_PREFIX}/bin/gparted", mode: 0755
   end
+  def self.postinstall
+        system "cat <<'EOF'> gparted
+#!/bin/bash
+DISPLAY=:0
+xhost si:localuser:root
+sudo gparted.orig
+EOF"
+    FileUtils.install 'gparted', "#{CREW_PREFIX}/bin/gparted", mode: 0755
 end
--- a/packages/gparted.rb
+++ b/packages/gparted.rb
@@ -9,19 +9,6 @@ class Gparted < Package
   source_url "https://downloads.sourceforge.net/project/gparted/gparted/gparted-#{@_ver}/gparted-#{@_ver}.tar.gz"
   source_sha256 '6c90715d254d7a7ec0208b29007b64160dd9fb7df4c4aa7f8ec2c9d23114c719'
 
-  binary_url ({
-     aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gparted-1.2.0-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/gparted.rb
```diff
@@ -45,12 +47,24 @@ def self.build
...
   def self.postinstall
-    puts "To run gparted, execute the following:".orange
-    puts "xhost si:localuser:root".lightblue
-    puts "sudo gparted".lightblue
+    bashrc = "#{HOME}/.bashrc"
+    puts 'Performing script-injection...'
+    unless `grep -c "# Added by gparted package" #{bashrc} || true`.to_i > 0
+      system "cat <<'EOF'>> #{bashrc}
+# Added by gparted package
+alias sudo='sudo '
```
uberhacker, line 59: This appears to be a pointless alias.

packages/gparted.rb
```diff
@@ -45,12 +47,24 @@ def self.build
...
+alias gparted='xhost si:localuser:root; sudo gparted'
```
uberhacker, line 60: Why not create a bash script and include it in the package so it can run in any environment?

packages/gparted.rb
```diff
@@ -45,12 +47,24 @@ def self.build
...
+
+EOF"
+    end
+    puts
+    puts 'To complete the installation, execute the following:'.lightblue
+    puts
+    puts "source #{bashrc}".lightblue
+    puts
```
uberhacker, line 68: This step can be removed if a bash script is used instead.

packages/gparted.rb
supechicken: This will let `alias` work with `sudo`

packages/gparted.rb
supechicken: Moved into script

packages/gparted.rb
```diff
@@ -45,12 +34,14 @@ def self.build
...
-  
-  def self.postinstall
-    puts "To run gparted, execute the following:".orange
-    puts "xhost si:localuser:root".lightblue
-    puts "sudo gparted".lightblue
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    FileUtils.mv "#{CREW_DEST_PREFIX}/bin/gparted", "#{CREW_DEST_PREFIX}/bin/gparted.orig"
+    system "cat <<'EOF'> gparted
+#!/bin/bash
+DISPLAY=:0
```
satmandu, line 41: Do we need to set DISPLAY? What if one is connected via ssh?

Can we just use the sudo env variables by doing something like a ```sudo -E```

packages/gparted.rb
```diff
@@ -45,12 +34,14 @@ def self.build
...
-    system "make DESTDIR=#{CREW_DEST_DIR} install"
-  end
-  
-  def self.postinstall
-    puts "To run gparted, execute the following:".orange
-    puts "xhost si:localuser:root".lightblue
-    puts "sudo gparted".lightblue
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    FileUtils.mv "#{CREW_DEST_PREFIX}/bin/gparted", "#{CREW_DEST_PREFIX}/bin/gparted.orig"
+    system "cat <<'EOF'> gparted
```
uberhacker, line 39: This should be a build step.

supechicken: Moved to build step

packages/gparted.rb
supechicken: Done, use `sudo -E`

satmandu:
One last thing: the Xwayland/xorg packages are using the original binary name as binary.elf - which is what google is doing for wayland & sommelier.  Could we do the same thing here just for consistency?

That would just mean renaming ```gparted``` ```gparted.elf```

satmandu:
Also please add ```depends_on 'f2fs_tools'```

satmandu (approved):

PR #5040, opened 2021-01-28
Add Gentoo pax_utils

satmandu:
This package has ```lddtools``` which can be used to find binaries with broken library dependencies.
It requires pyelftools.

e.g.
```
chronos@localhost /usr/local/lib64 $ /usr/local/bin/lddtree -a --skip-non-elfs /usr/local/lib64/libunbound.so
/usr/local/lib64/libunbound.so (interpreter => None)
    libssl.so.1.0.0 => None
    libcrypto.so.1.0.0 => None
    libpthread.so.0 => /usr/local/lib64/libpthread.so.0
        libc.so.6 => /usr/local/lib64/libc.so.6
            ld-linux-x86-64.so.2 => /usr/local/lib64/ld-linux-x86-64.so.2
        ld-linux-x86-64.so.2 => /usr/local/lib64/ld-linux-x86-64.so.2
    libc.so.6 => /usr/local/lib64/libc.so.6
        ld-linux-x86-64.so.2 => /usr/local/lib64/ld-linux-x86-64.so.2
chronos@localhost /usr/local/lib64 $ 
```

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/pax_utils.rb
+++ b/packages/pax_utils.rb
@@ -1,0 +1,41 @@
+require 'package'
+
+class Pax_utils < Package
+  description 'ELF utils that can check files for security relevant properties'
+  homepage 'https://wiki.gentoo.org/wiki/Hardened/PaX_Utilities'
+  version '1.2.5'
+  compatibility 'all'
+  source_url 'https://dev.gentoo.org/~slyfox/distfiles/pax-utils-1.2.5.tar.xz'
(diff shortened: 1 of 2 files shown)


satmandu:
Thanks for the space fix.

PR #5042, opened 2021-01-28
Xorg updates 

satmandu:
- This is a sync of the Xwayland and Xorg_server packages. 
- Both packages are based on xorg 1.20.10
- Both are using the Xwayland script now. 
- The Xwayland package is using the google patches, while the xorg_server package is not. 
- Both packages appear to work fine with sommelier for me.
- Both packages report hardware acceleration with sommelier using ```glxinfo -B``` (despite having used gcc, LTO, & the gold linker to compile these.)

### **_I need help testing this on i686 and armv7l hardware, as I don't have either of those. We don't want anybody's X install to stop working._**

Builds properly:
- [x] x86_64
- [x] i686
- [x] armv7l

Works properly:
- [x] x86_64
- [ ] i686 **_???_**
- [ ] armv7l  **_???_**
- [ ] Cloudready devices?  **_???_**

**Here's what would be useful to test:**
- [ ] sommelier w/ acceleration: ```restartsommelier ; glxinfo -B ```
 (```glxinfo``` is in the ```mesa_utils``` package.)
- [ ] making sure your X11 app of choice opens. (Make sure this isn't a Wayland app. ```xterm``` is a fine choice for testing.)

### DIFF
--- a/packages/xorg_server.rb
+++ b/packages/xorg_server.rb
@@ -3,55 +3,103 @@ require 'package'
 class Xorg_server < Package
   description 'The Xorg Server is the core of the X Window system.'
   homepage 'https://www.x.org'
-  version '1.20.7'
+  @_ver = '1.20.10'
+  version @_ver
   compatibility 'all'
-  source_url 'https://www.x.org/releases/individual/xserver/xorg-server-1.20.7.tar.bz2'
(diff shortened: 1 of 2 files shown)


satmandu:
@uberhacker Does that mean this worked fine for you?

uberhacker:
Nothing ever X related works in i686 for me.

PR #5049, opened 2021-01-29
New Package: balena_etcher

supechicken:
Flash OS images to SD cards & USB drives, safely and easily.

Works properly:
- [x] x86_64
- [ ] i686


packages/balena_etcher.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def self.build
+    Dir.chdir('../') do
+      system "alien -tc balena-etcher-electron_#{@_ver}_*.deb"
+      system "tar xf balena-etcher-electron-#{@_ver}.tgz"
+    end
+    system "cat <<'EOF'> balena-etcher
+#!/bin/sh
+xhost si:localuser:root
+sudo -E LD_LIBRARY_PATH=#{CREW_LIB_PREFIX} balena-etcher-electron
```
uberhacker, line 37: Shouldn't need to set `LD_LIBRARY_PATH`.  This is the default path unless it was changed.

uberhacker:
Thank you for the new package @supechicken!  This looks very useful.

PR #5050, opened 2021-01-29
crew: Replace '-' with '_' in crew install <pkgName>

supechicken:
This will definitely reduce confusion for users :)

Tested with commands:
```shell
crew download xfce4-terminal
crew install xfce4-terminal
crew reinstall xfce4-terminal
crew remove xfce4-terminal
crew files xfce4-terminal
crew search xfce4-terminal
```

### DIFF
--- a/crew
+++ b/crew
@@ -949,7 +949,7 @@ end
 
 def build_command (args)
   args["<name>"].each do |name|
-    @pkgName = name
+    @pkgName = name.gsub('-', '_')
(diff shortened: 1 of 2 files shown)


uberhacker:
Good idea.  We'll need to bump the version in `lib/const.rb`.

supechicken:
> Good idea. We'll need to bump the version in `lib/const.rb`.

Done

PR #5054, opened 2021-01-30
crew: Replace '-' with '_' once instead of many times (for #5050)

kjperkin:
## Description
DRYing up #5050... instead of doing gsub on every occurence of @pkgName, just rewrite args['<name>'] the one time.
 
## Addtional information
- [x] x86_64
- [ ] aarch64 (no ability to test with this platform)

## Tested with
```
crew install pax-utils
crew remove pax-utils libseccomp pyelftools libcap
```


uberhacker:
Nice!  I like the DRY philosophy.

PR #5069, opened 2021-02-01
crew: Add alternative name for packages

supechicken:
Tested with
```shell
crew install google-chrome
balena-etcher-electron
```

### DIFF
--- a/crew
+++ b/crew
@@ -211,8 +211,19 @@ def generate_compatible
 end
 
 def search (pkgName, silent = false)
-  Find.find (CREW_LIB_PATH + 'packages') do |filename|
-    return set_package(pkgName, filename) if filename == CREW_LIB_PATH + 'packages/' + pkgName + '.rb'
+  return set_package(pkgName, "#{CREW_LIB_PATH}/packages/#{pkgName}.rb") if File.exist?("#{CREW_LIB_PATH}/packages/#{pkgName}.rb")
+  Find.find ("#{CREW_LIB_PATH}/packages/") do |filename|
+    if File.extname(filename) == '.rb'
+      packageName = File.basename filename, '.rb'
+      case packageName
+      when 'hunspell', 'php', 'imagemagick'
+      else
+      set_package packageName, filename
+      end
+      if @pkg.aka
+        return set_package(packageName, filename) if @pkg.aka.include?(pkgName)
+      end
+    end
   end
   abort "Package #{pkgName} not found. :(".lightred unless silent
 end
--- a/packages/jdk11.rb
+++ b/packages/jdk11.rb
@@ -18,9 +18,9 @@ class Jdk11 < Package
     x86_64: 'aa768cfbd031d021d6df4dc5da90db385ac144ea4d6868447a14d2b3ace058b4',
   })
 
-  case ARGV[0]
-  when 'install', 'reinstall'
+  def self.preinstall
     abort "JDK8 installed.".lightgreen if Dir.exists? "#{CREW_PREFIX}/share/jdk8"
+    abort "JDK15 installed.".lightgreen if Dir.exists? "#{CREW_PREFIX}/share/jdk15"
   end
 
   def self.install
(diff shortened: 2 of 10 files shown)


uberhacker:
If you do something like this, you will need to do it for all packages or figure out a way to ignore or set a default value.  Otherwise, you will get errors.  I doubt this was tested.

crew
```diff
@@ -211,8 +211,19 @@ def generate_compatible
...
 
 def search (pkgName, silent = false)
-  Find.find (CREW_LIB_PATH + 'packages') do |filename|
-    return set_package(pkgName, filename) if filename == CREW_LIB_PATH + 'packages/' + pkgName + '.rb'
+  return set_package(pkgName, "#{CREW_LIB_PATH}/packages/#{pkgName}.rb") if File.exist?("#{CREW_LIB_PATH}/packages/#{pkgName}.rb")
+  Find.find ("#{CREW_LIB_PATH}/packages/") do |filename|
+    if File.extname(filename) == '.rb'
+      packageName = File.basename filename, '.rb'
+      case packageName
+      when 'hunspell', 'php', 'imagemagick'
```
uberhacker, line 219: This "list" should be referenced outside of `crew`.  There are other places in `crew` we could abstract like this also.

packages/jdk11.rb
```diff
@@ -18,9 +18,9 @@ class Jdk11 < Package
     x86_64: 'aa768cfbd031d021d6df4dc5da90db385ac144ea4d6868447a14d2b3ace058b4',
   })
 
-  case ARGV[0]
-  when 'install', 'reinstall'
+  def self.preinstall
```
uberhacker, line 21: Good call.  Maybe do this as a separate PR?

crew
```diff
@@ -211,8 +211,19 @@ def generate_compatible
...
+      else
+      set_package packageName, filename
```
kjperkin, line 221: I'm not a fan of having to load every package in order to examine its `aka` attribute. This gets slow quickly.

crew
```diff
@@ -211,8 +211,19 @@ def generate_compatible
...
+      end
+      if @pkg.aka
+        return set_package(packageName, filename) if @pkg.aka.include?(pkgName)
```
kjperkin, line 224: No need to call set_package a second time in this case. You've already found the package on line [221](#diff-9cae07a2e9ceea5b88602d99f1bb35f228663649628b7a3281ac9c8203e6043aR221).

kjperkin:
Some general thoughts:
* So far, when I've not found a package by the name I expect, I've been able to find it using `crew search`. Do other package management tools support something like this? I'm concerned about the complexity this adds.
* Does ruby have a reflection mechanism that would allow us to identify and work with the package subclass in a file without knowing it's name in advance? If so, we could use that and symbolic links to implement this. Create symbolic links pointing to the original package.rb and then use reflection to work with the package class.

uberhacker:
@kjperkin: I have no idea how to answer your second question but just about every package management tool has a search feature.  In Debian based Linux distros, you can `apt search <something>`.  In RedHat based Linux distros, you can `yum search <something>`.

kjperkin:
Sorry, that first question was poorly worded. What I meant to ask was, "do other package managers support alternative names for the same package?".

I guess I need to do some research to answer my second question. 😄

supechicken:
> Some general thoughts:
> 
> * So far, when I've not found a package by the name I expect, I've been able to find it using `crew search`. Do other package management tools support something like this? I'm concerned about the complexity this adds.
> * Does ruby have a reflection mechanism that would allow us to identify and work with the package subclass in a file without knowing it's name in advance? If so, we could use that and symbolic links to implement this. Create symbolic links pointing to the original package.rb and then use reflection to work with the package class.

`apt` has a similar thing called "transitional dummy package" (like `is_fake` in crew), which also do the same thing.

Actually, we can find them easily with `crew search` or `is_fake` (as you mentioned) so I am closing this :)

PR #5072, opened 2021-02-02
psmisc -> 23.3

satmandu:
- New upstream
- reformatted with rubocop

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/psmisc.rb
+++ b/packages/psmisc.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 require 'package'
 
 class Psmisc < Package
--- a/packages/psmisc.rb
+++ b/packages/psmisc.rb
@@ -27,7 +27,8 @@ class Psmisc < Package
   def self.build
     system './autogen.sh'
     system './configure --help'
-    system "env CFLAGS='-pipe -flto=auto -I#{CREW_PREFIX}/include/ncurses -fno-stack-protector' CXXFLAGS='-pipe -flto=auto' LDFLAGS='-flto=auto -I#{CREW_PREFIX}/include/ncurses' \
+    system "env CFLAGS='-pipe -flto=auto -I#{CREW_PREFIX}/include/ncurses -fno-stack-protector' \
+      CXXFLAGS='-pipe -flto=auto' LDFLAGS='-flto=auto -I#{CREW_PREFIX}/include/ncurses' \
       ./configure #{CREW_OPTIONS}"
     system 'make'
   end
--- a/packages/psmisc.rb
+++ b/packages/psmisc.rb
@@ -1,29 +1,35 @@
+# frozen_string_literal: true
+
 require 'package'
 
 class Psmisc < Package
   description 'PSmisc is a set of some small useful utilities that use the proc filesystem.'
-  homepage 'http://psmisc.sourceforge.net/'
-  version '23.1'
+  homepage 'https://gitlab.com/psmisc/psmisc'
+  @_ver = '23.3'
+  version @_ver
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


packages/psmisc.rb
```diff
@@ -1,29 +1,35 @@
...
+               armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/psmisc-23.3-chromeos-armv7l.tar.xz',
+               i686: 'https://dl.bintray.com/chromebrew/chromebrew/psmisc-23.3-chromeos-i686.tar.xz',
+               x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/psmisc-23.3-chromeos-x86_64.tar.xz'
+             })
+  binary_sha256({
+                  aarch64: 'd951db8224299af61b2ca8fb78d64f93667f52df0901928927d8c743216a629f',
+                  armv7l: 'd951db8224299af61b2ca8fb78d64f93667f52df0901928927d8c743216a629f',
+                  i686: '00181538fdcdb8922a07ff68f046a1d5333866f76bb241ae628d369179a931c2',
+                  x86_64: '61880a1987c8295f18c44c8472a953179510318082ed92ac1ade306dbda96bd1'
+                })
```
uberhacker, line 25: Can we get rid of the tabs?

packages/psmisc.rb
```diff
@@ -1,29 +1,35 @@
...
 
   def self.build
-    system "./configure --prefix=#{CREW_PREFIX}"
-    system "make CFLAGS=' -I#{CREW_PREFIX}/include/ncurses'"
+    system './autogen.sh'
+    system './configure --help'
+    system "env CFLAGS='-pipe -flto=auto -I#{CREW_PREFIX}/include/ncurses -fno-stack-protector' CXXFLAGS='-pipe -flto=auto' LDFLAGS='-flto=auto -I#{CREW_PREFIX}/include/ncurses' \
+      ./configure #{CREW_OPTIONS}"
```
uberhacker, line 31: Can we split this into several lines to avoid horizontal scrolling?

packages/psmisc.rb
satmandu: Can we move the indentation discussion to slack?

packages/psmisc.rb
satmandu: Yes absolutely this.

packages/psmisc.rb
uberhacker: We shouldn't even be having a discussion about it. :)  Please clean it up and try to be careful not to do it next time.

satmandu: I guess we're having the discussion here @cstrouse ! This is a part of seeing what sort of linter to use for PRs. The reformatting is from using rubocop, which can be used as a linter for checking PRs on submissions, so we don't have broken submissions any more.

The auto-reformatting from rubocop creates indents with spaces (not tabs) to put those lines where they are. 

I don't know if you can modify the rubocop linter setup to not check for the indents in this area, but it might be useful regardless? There's logic to this indentation...

satmandu: The rationale for the indentation: [Correctable] Layout/FirstHashElementIndentation: Use 2 spaces for indentation in a hash, relative to the start of the line where the left curly brace is.

uberhacker: OK, I wasn't aware you were using a linter.  If we do this, we need to be consistent.  I prefer the 2 space indentation because it's the most efficient use of screen real estate and unnecessary extra whitespace.  It's also consistent with the other indentation throughout the package.

satmandu: Changed Rubocop settings.

PR #5073, opened 2021-02-02
[rtmpdump]: Pass variables to make instead of patching the Makefile

kjperkin:
## Description
Patching the Makefile for rtmpdump should not be necessary. GNU make allows us to override the variables on the command line. See [this link](https://www.gnu.org/software/make/manual/html_node/Overriding.html#Overriding)

## Addtional information
Works properly:
- [x] x86_64
- [ ] aarch64 (cannot test, no hw)

Tested by running crew build and verifying that files are packaged into the correct tree structure (i.e. under /usr/local) and that binaries are linked against GnuTLS

### DIFF
--- a/packages/rtmpdump.rb
+++ b/packages/rtmpdump.rb
@@ -7,7 +7,7 @@ class Rtmpdump < Package
   compatibility 'all'
   source_url 'https://git.ffmpeg.org/gitweb/rtmpdump.git/snapshot/c5f04a58fc2aeea6296ca7c44ee4734c18401aa3.tar.gz'
   source_sha256 'fd8c21263d93fbde8bee8aa6c5f6a657789674bb0f9e74f050651504d5f43b46'
-  @make_common_opts = ["prefix=#{CREW_PREFIX}", "libdir=#{CREW_LIB_PREFIX}", 'CRYPTO=GNUTLS']
+  @make_common_opts = ['SYS=posix', "prefix=#{CREW_PREFIX}", "libdir=#{CREW_LIB_PREFIX}", 'CRYPTO=GNUTLS']
 
   binary_url ({
      aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/rtmpdump-c5f04a58f-1-chromeos-armv7l.tar.xz',
@@ -23,7 +23,7 @@ class Rtmpdump < Package
   })
 
   def self.build
-    system 'make', *@make_common_opts
+    system "make", *@make_common_opts
   end
 
   def self.install
--- a/packages/rtmpdump.rb
+++ b/packages/rtmpdump.rb
@@ -7,6 +7,7 @@ class Rtmpdump < Package
   compatibility 'all'
   source_url 'https://git.ffmpeg.org/gitweb/rtmpdump.git/snapshot/c5f04a58fc2aeea6296ca7c44ee4734c18401aa3.tar.gz'
   source_sha256 'fd8c21263d93fbde8bee8aa6c5f6a657789674bb0f9e74f050651504d5f43b46'
+  @make_common_opts = ["prefix=#{CREW_PREFIX}", "libdir=#{CREW_LIB_PREFIX}", 'CRYPTO=GNUTLS']
 
   binary_url ({
      aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/rtmpdump-c5f04a58f-1-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/rtmpdump.rb
```diff
@@ -20,20 +21,12 @@ class Rtmpdump < Package
...
-    system "sed -i 's,prefix=/usr/local,prefix=#{CREW_PREFIX},' librtmp/Makefile"
-    system "sed -i 's,libdir=\$(prefix)/lib,libdir=#{CREW_LIB_PREFIX},' librtmp/Makefile"
-    # OpenSSL builds are broken.
-    system "sed -e 's/^CRYPTO=OPENSSL/#CRYPTO=OPENSSL/' -e 's/#CRYPTO=GNUTLS/CRYPTO=GNUTLS/' -i Makefile -i librtmp/Makefile"
-  end
- 
+
   def self.build
-    system 'make'
+    system 'make', *@make_common_opts
```
uberhacker, line 26: We have a constant for this already.  Try `system "make #{CREW_OPTIONS} CRYPTO=GNUTLS"`.

packages/rtmpdump.rb
```diff
@@ -20,20 +21,12 @@ class Rtmpdump < Package
...
   end
 
   def self.install
-    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    system 'make', *@make_common_opts, "DESTDIR=#{CREW_DEST_DIR}", 'install'
```
uberhacker, line 30: Shouldn't need these extra options.

packages/rtmpdump.rb
kjperkin: This Makefile does not follow the normal conventions. prefix and libdir are lower-case, so CREW_OPTIONS is not appropriate here.

packages/rtmpdump.rb
kjperkin: The full install paths are formed by concatenating  `DESTDIR` with variables that depend on `prefix`, so we do need to specify `prefix` and `libdir` here. In included `CRYPTO` in `@make_common_opts` because I didn't want to bother tracking down the details of how it's used. It may well suffice to put it in `self.build` only.

kjperkin: @uberhacker Let me know if you have additional questions :smiley:

uberhacker: I already made my peace.  We shouldn't need this.  Humor me and try it without and see what happens.

kjperkin: **Attempt 1: Using `#{CREW_OPTIONS}` Only in `self.build`**
```ruby
# self.build
system "make #{CREW_OPTIONS}"
```
**Result 1** 
The build fails. The options in `CREW_OPTIONS` are valid only for configuration scripts generated by tools like `autoconf`.

**Attempt 2: Using `PREFIX=... LIBDIR=...` Only in `self.build`**
```ruby
# self.build
system "make PREFIX=/usr/local LIBDIR=/usr/local/lib64 CRYPTO=GNUTLS"
```
```ruby
# self.install
```
**Result 2**
The build copies libraries to `/usr/local/lib` instead of `/usr/local/lib64`. Note that using `PREFIX` and `LIBDIR` in `self.install` would result in the output files being written directly to `/usr/local`, which is very much not what we want.

**Explanation**
If you look at the contents of `librtmp/Makefile` in the source, you can see that `libdir` and `prefix` are the variables used by the Makefile. `prefix` is not 100% necessary, because it is defaulted to `/usr/local` but I'm of the opinion that it's better to be explicit for something like that. Going with that thinking, I've added `SYS=posix` to the arguments as well. That variable is also used by the `Makefile`.
```make
# librtmp/Makefile
# line 3
prefix=/usr/local

# line 7
libdir=$(prefix)/lib

# line 11
LIBDIR=$(DESTDIR)$(libdir)

# line 51
SODIR_posix=$(LIBDIR)
SODIR=$(SODIR_$(SYS))

# line 117
install_so: librtmp$(SO_EXT)
	cp librtmp$(SO_EXT) $(SODIR)
```

uberhacker: OK, so it looks like the `def self.patch` section is needed.  What about this is different from the previous build?  If it is, we would need to bump the version.

kjperkin: There shouldn't be anything different (in terms of the result) from the previous build. This is just eliminating `self.patch` and instead passing the appropriate variables to `make` when we run it. I just have a purist mindset that was rubbed the wrong way by the idea of patching a file and calling out to external programs when all we really need to do is call `make` with the right arguments. 😅

uberhacker: Fair enough.  With that said, it appears the project maintainer strayed from the beaten path with their Makefile. :)

PR #5074, opened 2021-02-03
Add rtws package

uberhacker:
Something made especially for @satmandu. :)

### DIFF
--- a/packages/rtws.rb
+++ b/packages/rtws.rb
@@ -1,0 +1,30 @@
+require 'package'
+
+class Rtws < Package
+  description 'Script to remove trailing whitespace'
+  homepage 'https://github.com/skycocker/chromebrew/wiki/FAQ'
+  version '1.0'
+  compatibility 'all'
+  source_url 'file:///dev/null'
+  source_sha256 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
+
+  def self.build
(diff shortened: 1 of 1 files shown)


satmandu:
Rubocop also does this...

uberhacker:
Well then, where is the rubocop package?

cstrouse:
Are we doing gem wrappers?

satmandu:
I'm not sure that we need to do a rubocop package? Why create a package when ```gem install rubocop``` will suffice?

The only reason I can see for doing that would be to prepackage a .rubocop.yml file

And I think that should be part of the core install. eg. : ~/.rubocop.yml

```
# The behavior of RuboCop can be controlled via the .rubocop.yml
# configuration file. It makes it possible to enable/disable
# certain cops (checks) and to alter their behavior if they accept
# any parameters. The file can be placed either in your home
# directory or in some project directory.
#
# RuboCop will start looking for the configuration file in the directory
# where the inspected file is and continue its way up to the root directory.
#
# See https://docs.rubocop.org/rubocop/configuration

AllCops:
  NewCops: enable

Layout/HashAlignment:
  EnforcedHashRocketStyle:
    - separator
  EnforcedColonStyle:
    - separator

Layout/FirstHashElementIndentation:
  EnforcedStyle: consistent

Style/FrozenStringLiteralComment:
  Enabled: false
```

PR #5079, opened 2021-02-04
curl -> 7.75

satmandu:
- also removed libiconv dep

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/curl.rb
+++ b/packages/curl.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 require 'package'
 
 class Curl < Package
--- a/packages/curl.rb
+++ b/packages/curl.rb
@@ -1,48 +1,51 @@
+# frozen_string_literal: true
+
 require 'package'
 
 class Curl < Package
   description 'Command line tool and library for transferring data with URLs.'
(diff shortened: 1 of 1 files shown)


packages/curl.rb
```diff
@@ -1,58 +1,61 @@
...
   depends_on 'libtirpc'
   depends_on 'libunbound'
   depends_on 'openldap'
-  depends_on 'rtmpdump'
   depends_on 'zstd'
+  depends_on 'rtmpdump'
 
   def self.build
-    system "env CFLAGS='-flto=auto' CXXFLAGS='-flto=auto' \
+    raise StandardError, 'Please remove libiconv before building.' if File.exist?("#{CREW_LIB_PREFIX}/libcharset.so")
```
kjperkin, line 38: What's wrong with having libiconv installed when building curl?

satmandu:
Linux systems normally have iconv in glibc.

Having fewer required dependencies for core packages is good, right?

kjperkin:
Agree. I'm looking into if there's a way to do this without failing the build when libiconv is installed. It seems strange to me that we're forced to use it if it's there...

PR #5082, opened 2021-02-05
Waypipe and deps

satmandu:
- This adds waypipe, which lets you 
- libwacom is rebuilt since it was pulling in the system libffi
- weston is rebuilt since it was pulling in the old libwacom, and wasn't building on armv7l.

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

Example usage when the other side also has weston and waypipe (of the same version) installed:

```waypipe --compress lz4 ssh user@remotehost "weston-terminal &"```

or for a more complicated example:

```waypipe --compress lz4 ssh user@remotehost "/usr/bin/dbus-launch /usr/bin/gnome-terminal &"```

### DIFF
--- a/packages/libwacom.rb
+++ b/packages/libwacom.rb
@@ -2,34 +2,45 @@ require 'package'
 
 class Libwacom < Package
   description 'libwacom is a wrapper library for evdev devices.'
-  homepage 'http://linuxwacom.sourceforge.net/wiki/index.php/Libwacom'
-  version '0.28-0'
+  homepage 'https://github.com/linuxwacom/libwacom'
+  @_ver = 1.8
+  version @_ver
   compatibility 'all'
-  source_url 'https://downloads.sourceforge.net/linuxwacom/libwacom-0.28.tar.bz2'
-  source_sha256 'e7d632301288b221cb5af69b4c5e57fd062bafd9a9acd6f9ce271570103267ef'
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/libwacom.rb [hunk 1]
```diff
@@ -2,34 +2,45 @@
...
+  binary_sha256({
+    aarch64: '7029db156ea3a27a9024f12071cb1fb99a54f233b1df0b3c5eb6174d5e5a54fd',
+     armv7l: '7029db156ea3a27a9024f12071cb1fb99a54f233b1df0b3c5eb6174d5e5a54fd',
+       i686: '9c1f8a3548b8d374e7da99311b9ecd88781989f605ef7fa4f3b703f5bd8e6cf7',
+     x86_64: 'bd25043dcd7c6affa57e04ac2c9b6abebefa03ca7c6f83cb5c64827562ccf770'
   })
 
   depends_on 'libgudev'
   depends_on 'eudev'
+  depends_on 'libevdev'
```
uberhacker, line 27: Do we need both libgudev and libevdev?

satmandu: libgudev: a library providing GObject bindings for libudev
libevdev: a wrapper library for evdev devices

So they are different...

satmandu: They are both needed, as they come up as configure failures otherwise.)

PR #5084, opened 2021-02-06
use CAPSH wrapper scripts to avoid setuid and root ownership of inetutils {ping,ping6,traceroute}

satmandu:
- Use sudo wrapper scripts for  ping, ping6, and traceroute so that capsh can be used to give these apps the necessary ```cap_net_raw``` capabilities.
- use patchelf to set ```rpath``` so if run using sudo w/o setuid the environment doesn't change the LD_LIBRARY_PATH.

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/inetutils.rb
+++ b/packages/inetutils.rb
@@ -4,30 +4,32 @@ class Inetutils < Package
   description 'The Inetutils package contains programs for basic networking. Such as dnsdomainname, ftp, hostname, ifconfig, ping, ping6, talk, telnet, tftp, traceroute'
   homepage 'https://www.gnu.org/software/inetutils/'
   @_ver = '2.0'
-  version @_ver
+  version "#{@_ver}-1"
   compatibility 'all'
   source_url "https://ftp.gnu.org/gnu/inetutils/inetutils-#{@_ver}.tar.xz"
   source_sha256 'e573d566e55393940099862e7f8994164a0ed12f5a86c3345380842bdc124722'
 
-  binary_url ({
-     aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/inetutils-2.0-chromeos-armv7l.tar.xz',
-      armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/inetutils-2.0-chromeos-armv7l.tar.xz',
-        i686: 'https://dl.bintray.com/chromebrew/chromebrew/inetutils-2.0-chromeos-i686.tar.xz',
-      x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/inetutils-2.0-chromeos-x86_64.tar.xz',
+  binary_url({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/inetutils-2.0-1-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/inetutils-2.0-1-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/inetutils-2.0-1-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/inetutils-2.0-1-chromeos-x86_64.tar.xz'
   })
-  binary_sha256 ({
-     aarch64: '7446162e65febe771bc8c88efc9d27865e7ef0d35d03c8ca90ffd4046e7eb925',
-      armv7l: '7446162e65febe771bc8c88efc9d27865e7ef0d35d03c8ca90ffd4046e7eb925',
-        i686: 'ed0cd237ca49f613d3398704c9cde04f78af14113458cef775bc4908db9b3f38',
-      x86_64: 'cef8fbdd17ac33c31587420b5291dd9738d5fa4849140c046bb1cce9d0861d70',
+  binary_sha256({
+    aarch64: '9761fd87ab7276ef701071a52cea3eed86819c4cb6a6bbd01a0062cc51a2a6d3',
+     armv7l: '9761fd87ab7276ef701071a52cea3eed86819c4cb6a6bbd01a0062cc51a2a6d3',
+       i686: '254bdff9f045a1909111b71c248483b86671a0564ad3fd273168ef6f1f6dc595',
+     x86_64: 'bc795eea585ce49e7f041ba850450b67c74e49d9141ba1f27d58614597af208d'
   })
 
   depends_on 'linux_pam'
+  depends_on 'patchelf'
 
   def self.build
     system "env CFLAGS='-flto=auto -ltinfo' CXXFLAGS='-flto=auto' LDFLAGS='-flto=auto' \
-      ./configure #{CREW_OPTIONS} \
+      LIBRARY_PATH=#{CREW_LIB_PREFIX} ./configure #{CREW_OPTIONS} \
       --with-krb5=#{CREW_PREFIX} \
+      --disable-rpath \
       --with-wrap \
       --with-pam \
       --disable-rexec \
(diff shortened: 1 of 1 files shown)


packages/inetutils.rb
```diff
@@ -40,5 +42,27 @@ def self.build
...
+    system "patchelf --set-rpath #{CREW_LIB_PREFIX} #{CREW_DEST_PREFIX}/bin/traceroute"
+  end
+
+  def self.preinstall
+    system "sudo chown chronos #{CREW_PREFIX}/bin/ping" \
+      if File.exist? "#{CREW_PREFIX}/bin/ping"
+    system "sudo chown chronos #{CREW_PREFIX}/bin/ping6" \
+      if File.exist? "#{CREW_PREFIX}/bin/ping6"
+    system "sudo chown chronos #{CREW_PREFIX}/bin/traceroute" \
+      if File.exist? "#{CREW_PREFIX}/bin/traceroute"
```
uberhacker, line 56: I can't think of any reason why the ownership would not be the chronos user unless these were installed using sudo which should never be the case.

packages/inetutils.rb
```diff
@@ -40,5 +42,27 @@ def self.build
...
+  end
+
+  def self.postinstall
+    puts 'Settings permissions for utilities using sudo'.orange
+    system "sudo chown root #{CREW_PREFIX}/bin/ping"
+    system "sudo chown root #{CREW_PREFIX}/bin/ping6"
+    system "sudo chown root #{CREW_PREFIX}/bin/traceroute"
+    system "sudo chmod 4755 #{CREW_PREFIX}/bin/ping"
+    system "sudo chmod 4755 #{CREW_PREFIX}/bin/ping6"
+    system "sudo chmod 4755 #{CREW_PREFIX}/bin/traceroute"
```
uberhacker, line 66: Shouldn't need to change ownership to root.

uberhacker: Also, you shouldn't need to use sudo to change permissions if the ownership is already chronos.

satmandu: hmmm. inetutils' ping won't work otherwise:
```
chronos@localhost /usr/local/lib/crew/packages $ /bin/ping 8.8.8.8
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=117 time=13.2 ms
64 bytes from 8.8.8.8: icmp_seq=2 ttl=117 time=15.8 ms
^C
--- 8.8.8.8 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1001ms
rtt min/avg/max/mdev = 13.253/14.571/15.890/1.324 ms
chronos@localhost /usr/local/lib/crew/packages $ /usr/local/bin/ping 8.8.8.8
ping: Lacking privilege for icmp socket.
```

satmandu: The only alternative I see is to use setcap, which requires sudo too.

e.g. ```sudo /usr/local/sbin/setcap  cap_net_raw+ep /usr/local/bin/ping```

This is what google's default ping has done to it:
```
chronos@localhost /usr/local/lib/crew/packages $ /usr/local/sbin/getcap /bin/ping
/bin/ping = cap_net_raw+ep
```

satmandu: (This would be the cleaner solution, not requiring any use of ```chown```.

satmandu: Alternately one could implement a wrapper script which uses sudo, just like is used by the gparted package, but that is its own hack...

satmandu: e.g.

```
#!/bin/bash
sudo -E capsh --caps="cap_net_raw+eip cap_setpcap,cap_setuid,cap_setgid+ep" \
    --keep=1 --user=nobody --addamb=cap_net_raw -- \
    -c "/usr/local/bin/ping $@"
```

satmandu: This is the direction I'm going... but I'm getting capsh errors...

supechicken:
Can we use `fakeroot` to bypass the root detection?

packages/inetutils.rb
satmandu: removed

packages/inetutils.rb
satmandu: wrapper scripts appear to work with newer libcap

satmandu:
@supechicken I got it working with wrapper scripts and an updated libcap, in large thanks to copying what you did with gparted.

packages/inetutils.rb
```diff
@@ -36,9 +39,36 @@ def self.build
...
+#!/bin/bash
+sudo -E #{CREW_PREFIX}/sbin/capsh --caps=\"cap_net_raw+eip cap_setpcap,cap_setuid,cap_setgid+ep\" \\
+    --keep=1 --user=nobody --addamb=cap_net_raw -- \\
+    -c \"#{CREW_PREFIX}/bin/ping6.elf \$@\"
+EOF"
+    system "cat <<'EOF'> traceroute_
+#!/bin/bash
+sudo -E #{CREW_PREFIX}/sbin/capsh --caps=\"cap_net_raw+eip cap_setpcap,cap_setuid,cap_setgid+ep\" \\
+    --keep=1 --user=nobody --addamb=cap_net_raw -- \\
+    -c \"#{CREW_PREFIX}/bin/traceroute.elf \$@\"
```
uberhacker, line 58: Is sudo necessary?  I'm thinking maybe not?

satmandu: You have to use sudo to elevate permissions to use capsh.

supechicken (approved):

satmandu:
@uberhacker Any other concerns besides use of sudo in a wrapper script? (Which AFAI can tell can't be worked around?)

packages/inetutils.rb
uberhacker: Shouldn't need to use sudo to ping.  I'm confused.

packages/inetutils.rb
```diff
@@ -36,9 +39,36 @@ def self.build
       --enable-authentication \
       --disable-servers"
     system 'make'
+    system "cat <<'EOF'> ping_
```
kjperkin, line 42: These wrappers can be written without using `system`, which looks a bit nicer. For example:
```
PING_SH = <<EOF
#!/bin/bash
sudo -E #{CREW_PREFIX}/sbin/capsh --caps='cap_net_raw+eip cap_setpcap,cap_setuid,cap_setgid+ep' \\
    --keep=1 --user=nobody --addamb=cap_net_raw -- \\
    -c "#{CREW_PREFIX}/bin/ping.elf \$@"
EOF
IO.write('ping_', PING_SH, perm: 0755)
```

In the example above, I define `PING_SH` and then immediately use it, but IMO it should be moved out of `self.build` for readability.

uberhacker: Why not go ahead and write to `#{CREW_DEST_PREFIX}/bin/ping`?

satmandu: @uberhacker because ```system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'``` will overwrite ```#{CREW_DEST_PREFIX}/bin/ping```.

satmandu: Thanks @kjperkin. Do these changes look correct to you?

satmandu: @uberhacker Moved the heredocs to install, where as per your suggestion they get written straight to ```#{CREW_DEST_PREFIX}/bin/```.

kjperkin: @satmandu looking better. I read up on [heredocs](https://ruby-doc.org/core-3.0.0/doc/syntax/literals_rdoc.html#label-Here+Documents+-28heredocs-29) a bit more and if you put a `~` before the EOF, then you can indent everything nicely. I also think it would be good to move the constants outside of `self.install` so it's easier to read the method.

satmandu:
@uberhacker The o is for octal literals for mode:

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/NumericLiteralPrefix
See also:

https://www.dennyzhang.com/rubocop_errors

packages/inetutils.rb
satmandu: Example of how you would change the constants?

satmandu:
@uberhacker I removed IO.write. Good simplification catch!

packages/inetutils.rb
```diff
@@ -40,5 +43,30 @@ def self.build
...
+      sudo -E #{CREW_PREFIX}/sbin/capsh --caps='cap_net_raw+eip cap_setpcap,cap_setuid,cap_setgid+ep' \\
+          --keep=1 --user=nobody --addamb=cap_net_raw -- \\
+          -c "#{CREW_PREFIX}/bin/ping6.elf \$@"
+EOF
+    "#{CREW_DEST_PREFIX}/bin/traceroute" = <~EOF
+      #!/bin/bash
+      sudo -E #{CREW_PREFIX}/sbin/capsh --caps='cap_net_raw+eip cap_setpcap,cap_setuid,cap_setgid+ep' \\
+          --keep=1 --user=nobody --addamb=cap_net_raw -- \\
+          -c "#{CREW_PREFIX}/bin/traceroute.elf \$@"
+EOF
```
uberhacker, line 70: Nice simplified solution!

kjperkin:
> @uberhacker I removed IO.write. Good simplification catch!

My google-fu is failing me. How does this work?

satmandu:
We needed it. Had to create a PR to put it back in.

PR #5089, opened 2021-02-07
yelp_tools => .40.alpha

satmandu:
- better to use an official tag than a git hash

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686


uberhacker:
Alpha releases are risky. :)

satmandu:
Oh hell yes, but since we don't have any big downstream gnome packages yet, what are we breaking?

PR #5090, opened 2021-02-07
Add default Rubocop linter configuration

satmandu:
- Add a default Rubocop linter config file people can use.
- This won't change ANYTHING, since we're not running any automated scripts based on this at the moment.
 
Obviously, this is open for additions. I'm not tied to these settings, but I just wanted to start somewhere.

To get this set up on your system do this:
```
crew install libyaml
gem install rubocop
```
Once this file is at ~/.rubocop.yml, you can run ```rubocop package.rb``` to get some suggested changes.

Run ```rubocop -A package.rb``` to try to make those fixes automatically.

Please feel free to add to this PR if there are changes you think should be made.

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686


uberhacker:
Why not make a wrapper package that installs the gem and puts the config in `~/.rubocop.yml` for you?  That is the whole point of packaging, right? :)

satmandu:
I'm ok with packaging rubocop, but I wanted feedback on the rubocop config file, which since it applies to all package files, I wanted up front so people could see it. I'd have the current rubocop config file pulled from github anyways...

uberhacker:
How do you configure specific settings like 2 space indentation, for example?

satmandu:
@uberhacker I think that's in this section:
```
Layout/FirstHashElementIndentation
  EnforcedStyle: consistent
```
(Otherwise, 2 spaces is the ruby default.)

uberhacker:
I guess I wanted something more specific.  Say I wanted to change a default setting.  How is that configured...specifically?

kjperkin:
@uberhacker Looks like there is a default configuration that comes with rubocop, and then your own configuration is merged into the default, as described [here](https://docs.rubocop.org/rubocop/configuration.html#inheritance). So to override a default, you would put it in your own .rubocop.yml.

I don't think this should be a package. `gem install rubocop` is easy and having .rubocop.yml in the repository root limits the scope of the settings to files in the crew repository.

uberhacker:
That still doesn't answer my question but when if can minimize the amount of work to set things up, the better.

kjperkin:
If you really wanted to change the global defaults used by rubocop, there's
a yml file where it is installed. However, changing that would affect
everything, not just crew code.

satmandu:
Adding examples of why the rule was needed to the file as I go now...

uberhacker:
No chance this can be packaged?  I'd like to avoid manual steps to encourage more usage from other package committers (myself included).  We need a crew specific config anyway.  Don't care about their default config.

satmandu:
I'd like the default config file kept on the web site though so we can integrate that into a linter on github for PR checks.

Here is a sample package file:
```
require 'package'

class Rubocop_gem < Package
  description 'Rubocop Ruby linter.'
  homepage 'https://github.com/rubocop-hq/rubocop'
  version '1.9.1'
  compatibility 'all'
  source_url 'file:///dev/null'

  def self.postinstall
    system 'gem install rubocop'
    unless File.exist? '~/.rubocop.yml'
      puts 'Putting Chromebrew rubocop config file into ~/.rubocop.yml'.lightgreen
      system 'curl -Lf https://github.com/skycocker/chromebrew/raw/master/.rubocop.yml \
        -o ~/.rubocop.yml'
    end
  end

  def self.remove
    system 'gem uninstall rubocop'
  end
end
```

satmandu:
(I'm scared of creating a tar archive for gems since I don't want to get into the rabbit hole of packaging dependencies, and also because I don't know which databases get updated on removal/addition of gem packages.)

kjperkin:
@uberhacker As long as .rubocop.yml is in the root of the repository, rubocop will pick it up when run on any file in the repository. This allows you to have your personal rubocop configuration in your home directory and the `crew` configuration for `crew`-related work.

@uberhacker @satmandu Generally speaking, I think gem management is a solved problem with `gem` so we should avoid having packages that only run `gem install <gem>`. I'd be more comfortable if our "configuration" package were to be tailored to the use case of hacking on `crew`. For example, it might behave as follows:
* On install, the package does `gem install rubocop`
* On install, the package reconfigures the git repository in /usr/local/crew/lib so that .rubocop.yml is checked out
* On installation, the package adds useful git hooks in /usr/local/crew/lib to do things like run rubocop. The hooks should live somewhere in the repo like the `tools` directory. Bonus points if we can still do sparse checkout and not have the tools directory in /usr/local/lib/crew.
* On upgrade, all of these changes are undone and then the new package does its thing
* On removal, all of the changes are undone

uberhacker:
Maybe I need to reiterate my point.  I don't care about the default .rubocop.yml because it has no practical use.  Now if you want to talk about a config that is specific to Chromebrew, I'm all ears.  All this extra git checkout stuff makes no sense.  Just pull the stuff you need, package it and wash your hands.  We have wrapper packages for perl, python and php; gem should be no different.  The trick is to capture the metadata so the package can be uninstalled elegantly.  Maybe I'm oversimplifying the process but I'm thinking it should be easier than you describe.

satmandu:
This PR was literally meant to only to create a metadiscussion about a rubocop config that can be run on chromebrew package files when they are submitted as PRs.

The discussion of the packaging of gems seems largely off-topic here.

(Also that seems academic against the larger issue of what we're going to do about hosting our packages when bintray stops accepting packages in 3 weeks.)

uberhacker:
I think the gem has value.  Package committers can tidy things up before the GitHub linter rejects their PR.  This is ideal since it may take awhile before the GitHub actions provide feedback.  Also, packages should be checked thoroughly before submitting a PR anyway.

kjperkin:
> This PR was literally meant to only to create a metadiscussion about a rubocop config that can be run on chromebrew package files when they are submitted as PRs.

I think I may be misunderstanding something. In [this comment](https://github.com/skycocker/chromebrew/pull/5090#issuecomment-774746390) @uberhacker suggests we have a package that would either overwrite or modify a user's ~/.rubocop.yml. If we do that though, it means that users can't have their own defaults to use for projects other than chromebrew. If insteadb  .rubocop.yml lives in the repositry, the package can be installed without ruining the user's personal config.

.rubocop.yml
```diff
@@ -0,0 +1,47 @@
...
+Naming/VariableNumber:
+  Enabled: false
+  
+Naming/ClassAndModuleCamelCase:
+  Enabled: false
+
+# ncurses.rb:11:3: C: Metrics/MethodLength: Method has too many lines. [31/10]
+#  def self.build ...
+#  ^^^^^^^^^^^^^^
+Metrics/MethodLength:
```
kjperkin, line 40: Any way to disable for just the one file or method instead of globally, or if there is a sane higher limit maybe apply that instead?

cstrouse: https://docs.rubocop.org/rubocop/configuration.html#includingexcluding-files
https://docs.rubocop.org/rubocop/configuration.html#disabling-cops-within-source-code

PR #5093, opened 2021-02-08
libiconv: rebuild use lto, keep iconv.h

satmandu:
- sometimes you need to build with libiconv, so don't break that.

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/libiconv.rb
+++ b/packages/libiconv.rb
@@ -3,32 +3,39 @@ require 'package'
 class Libiconv < Package
   description 'GNU charset conversion library for libc which does not implement it.'
   homepage 'https://www.gnu.org/software/libiconv/'
-  version '1.16-3'
+  version '1.16-4'
   compatibility 'all'
   source_url 'https://ftpmirror.gnu.org/libiconv/libiconv-1.16.tar.gz'
   source_sha256 'e6a1b1b589654277ee790cce3734f07876ac4ccfaecbee8afa0b649cf529cc04'
 
   binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/libiconv-1.16-3-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/libiconv-1.16-3-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/libiconv-1.16-3-chromeos-i686.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/libiconv.rb
```diff
@@ -3,32 +3,39 @@
...
+        --enable-static \
+        --enable-relocatable \
+        --enable-extra-encodings"
     system 'make'
   end
 
   def self.install
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    # Don't remove iconv.h. Better not to have this installed if you don't
+    # want a comflict.
```
uberhacker, line 37: conflict?

packages/libiconv.rb
```diff
@@ -3,32 +3,39 @@
...
     # Remove iconv.h which conflicts with same file from glibc
-    FileUtils.rm "#{CREW_DEST_PREFIX}/include/iconv.h"
+    # FileUtils.rm "#{CREW_DEST_PREFIX}/include/iconv.h"
```
uberhacker, line 39: How do we debug if someone installs this and it breaks something else?

packages/libiconv.rb
satmandu: Yes absolutely. But there are packages like evolution_data_server. which won't compile w/o libiconv, and without the libiconv ```iconv.h``` you can't get the package to compile.

One solution might be to push those files to a different location.

For instance, this is what alpine is doing:

https://pkgs.alpinelinux.org/contents?branch=edge&name=gnu-libiconv-dev&arch=x86&repo=community

| File | Package |
| -- | -- |
| /usr/include/gnu-libiconv/iconv.h | gnu-libiconv-dev |
| /usr/include/gnu-libiconv/libcharset.h | gnu-libiconv-dev | 
| /usr/include/gnu-libiconv/localcharset.h | gnu-libiconv-dev |
| /usr/lib/libcharset.so | gnu-libiconv-dev |
| /usr/lib/libiconv.so | gnu-libiconv-dev |

Would that be acceptable?

uberhacker: I meant you misspelled conflict. :)  But yeah, if you can avoid the conflict with gcc, that would be ideal.

satmandu: headdesk.gif

packages/libiconv.rb
satmandu: now #{CREW_DEST_PREFIX}/include/iconv.h only refers to glibc's iconv.h

PR #5102, opened 2021-02-09
[New Package] SWI-Prolog

kjperkin:
## Description
Several fixes/enhancements to supporting libraries plus the package itself. This is a PR for the addition of a package for [SWI-Prolog](https://www.swi-prolog.org/). I set this up to build from the (at the time of this writing). 

## Addtional information
This has been partially tested. All the SWI-Prolog tests pass and my crew installation seems to work fine so far with the libraries built and installed as configured here. Some things I've tried:

* building and compiling things
* authoring these commits ;-)
* running pcre's tests (they pass except for one which seems to fail because of a problem with locale) 

Works properly:
- [x] x86_64

Things that still need to be done:
[ ] Build binaries
[ ] Update links
[ ] Check all versions bumped
[ ] More tests?

### DIFF
--- a/packages/pcre.rb
+++ b/packages/pcre.rb
@@ -3,41 +3,43 @@ require 'package'
 class Pcre < Package
   description 'The PCRE library is a set of functions that implement regular expression pattern matching using the same syntax and semantics as Perl 5.'
   homepage 'http://pcre.org/'
-  version '8.44-1'
+  @_ver = '8.44'
+  version "#{@_ver}-1"
   compatibility 'all'
-  source_url 'https://ftp.pcre.org/pub/pcre/pcre-8.44.tar.bz2'
+  source_url "https://ftp.pcre.org/pub/pcre/pcre-#{@_ver}.tar.bz2"
   source_sha256 '19108658b23b3ec5058edc9f66ac545ea19f9537234be1ec62b714c84399366d'
 
-  binary_url ({
+  binary_url({
     aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/pcre-8.44-1-chromeos-armv7l.tar.xz',
      armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/pcre-8.44-1-chromeos-armv7l.tar.xz',
        i686: 'https://dl.bintray.com/chromebrew/chromebrew/pcre-8.44-1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/pcre-8.44-1-chromeos-x86_64.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/pcre-8.44-1-chromeos-x86_64.tar.xz'
   })
-  binary_sha256 ({
-    aarch64: 'b6475dc3ff52f0c471d320a7c7d7a17ad015f2f08b46f4fc6805d80efb80ec10',
-     armv7l: 'b6475dc3ff52f0c471d320a7c7d7a17ad015f2f08b46f4fc6805d80efb80ec10',
-       i686: 'abb6faf10ed437b2fdef5bd6da661a2327d2ffa79dfbdf7f3bc12713da1370cf',
-     x86_64: '4f2e3865d910f15b113c59cb69ce87dedbc236a8f655304ffb3fe8ceb89ab64a',
+  binary_sha256({
+    aarch64: '6d54befc6735b9d85928c14d1f03519d25514aed4c1a1f3d12114d3585f535f3',
+     armv7l: '6d54befc6735b9d85928c14d1f03519d25514aed4c1a1f3d12114d3585f535f3',
+       i686: '092fe524f7e00bc4e19b6a5cf01e1677f9d7900c8e105adcd086cd9670c7c558',
+     x86_64: '4c62ce6b8ae532c7b0885973b50363034c2a84b281baa5542149226a38016cbe'
   })
 
   def self.build
-    system "./configure #{CREW_OPTIONS} \
-  --enable-shared \
-  --disable-static \
-  --with-pic \
-  --enable-unicode-properties \
-  --enable-pcre16 \
-  --enable-pcre32 \
-  --enable-jit \
-  --enable-pcretest-libreadline"
+    system "env CFLAGS='-flto=auto' CXXFLAGS='-flto=auto' \
+      LDFLAGS='-flto=auto' \
+      ./configure #{CREW_OPTIONS} \
+      --enable-shared \
+      --with-pic \
+      --enable-unicode-properties \
+      --enable-pcre16 \
+      --enable-pcre32 \
+      --enable-jit \
+      --enable-pcretest-libreadline"
     system 'make'
   end
 
   def self.check
     # `make check` fails because of an issue with locale.
     # All but one test passes.
-    #system 'make', 'check'
+    # system 'make', 'check'
   end
 
   def self.install
--- a/packages/pcre.rb
+++ b/packages/pcre.rb
@@ -26,7 +26,7 @@ class Pcre < Package
   --enable-shared \
   --disable-static \
   --with-pic \
-  --enable-utf \
+  --enable-unicode-properties \
   --enable-pcre16 \
   --enable-pcre32 \
   --enable-jit \
@@ -35,7 +35,9 @@ class Pcre < Package
   end
 
   def self.check
-#    system 'make', 'check'
+    # `make check` fails because of an issue with locale.
+    # All but one test passes.
+    #system 'make', 'check'
   end
 
   def self.install
--- a/packages/readline.rb
+++ b/packages/readline.rb
@@ -3,30 +3,37 @@ require 'package'
 class Readline < Package
   description 'The GNU Readline library provides a set of functions for use by applications that allow users to edit command lines as they are typed in.'
   homepage 'http://cnswww.cns.cwru.edu/php/chet/readline/rltop.html'
-  version '8.1-1'
+  @_ver = '8.1'
+  version "#{@_ver}-1"
   compatibility 'all'
-  source_url 'https://gnu.askapache.com/readline/readline-8.1.tar.gz'
+  source_url "https://gnu.askapache.com/readline/readline-#{@_ver}.tar.gz"
   source_sha256 'f8ceb4ee131e3232226a17f51b164afc46cd0b9e6cef344be87c65962cb82b02'
 
-  binary_url ({
-     aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/readline-8.1-1-chromeos-armv7l.tar.xz',
-      armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/readline-8.1-1-chromeos-armv7l.tar.xz',
-        i686: 'https://dl.bintray.com/chromebrew/chromebrew/readline-8.1-1-chromeos-i686.tar.xz',
-      x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/readline-8.1-1-chromeos-x86_64.tar.xz',
+  binary_url({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/readline-8.1-1-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/readline-8.1-1-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/readline-8.1-1-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/readline-8.1-1-chromeos-x86_64.tar.xz'
   })
-  binary_sha256 ({
-     aarch64: '4c2e3ab6488be4b58a28975f4e9f2fa7fb8649402cf4268c15c708bce9667f20',
-      armv7l: '4c2e3ab6488be4b58a28975f4e9f2fa7fb8649402cf4268c15c708bce9667f20',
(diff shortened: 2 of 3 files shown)


packages/readline.rb
```diff
@@ -3,16 +3,16 @@
 class Readline < Package
   description 'The GNU Readline library provides a set of functions for use by applications that allow users to edit command lines as they are typed in.'
   homepage 'http://cnswww.cns.cwru.edu/php/chet/readline/rltop.html'
-  version '8.1'
+  version '8.1-1'
```
supechicken, line 6: Maybe split these patches into separate PRs?

packages/pcre.rb
```diff
@@ -8,6 +8,8 @@ class Pcre < Package
   source_url 'https://ftp.pcre.org/pub/pcre/pcre-8.44.tar.bz2'
   source_sha256 '19108658b23b3ec5058edc9f66ac545ea19f9537234be1ec62b714c84399366d'
 
+  depends_on 'readline'
```
uberhacker, line 11: No need to include core package.

packages/readline.rb
```diff
@@ -21,12 +21,11 @@ class Readline < Package
       x86_64: 'ddd7cf950d05c8492834b60ff0dbb124f07158c671784eb78118e0fa71691d1b',
   })
 
-
+  depends_on 'ncurses'
```
uberhacker, line 24: No need to include core package.

packages/ncurses.rb
```diff
@@ -63,11 +63,9 @@ def self.build
   def self.install
     Dir.chdir 'ncursesw_build' do
       system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
-      FileUtils.ln_sf "#{CREW_LIB_PREFIX}/libtinfow.so.6", "#{CREW_DEST_LIB_PREFIX}/libtinfow.so.5"
     end
     Dir.chdir 'ncurses_build' do
       system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
-      FileUtils.ln_sf "#{CREW_LIB_PREFIX}/libtinfo.so.6", "#{CREW_DEST_LIB_PREFIX}/libtinfo.so.5"
```
uberhacker, line 70: We'll need new binaries to really get rid of the symlinks.  Please remove so new ones can be added.

packages/swi_prolog.rb
```diff
@@ -0,0 +1,58 @@
...
+  depends_on 'openssl'
+  depends_on 'gmp'
+  depends_on 'readline'
+  depends_on 'libarchive'
+  depends_on 'ncurses'
+  depends_on 'libdb'
+  depends_on 'acl'
+  depends_on 'ninja' => :build
+  depends_on 'cmake' => :build
+  depends_on 'llvm' => :build
```
uberhacker, line 27: No need to include most of these dependencies since they're all core packages.  Only need:
```
depends_on 'libarchive'
depends_on 'libdb'
```

packages/swi_prolog.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def self.build
+    FileUtils.remove_dir 'build', true
+    Dir.mkdir 'build'
+    Dir.chdir 'build' do
+        system "cmake -G Ninja \
+        #{CREW_CMAKE_OPTIONS} \
+        -DCURSES_CURSES_LIBRARY=/usr/local/lib64/libncurses.so.6 \
+        -DCURSES_INCLUDE_PATH=/usr/local/include/ncurses  \
+        -DCMAKE_LIBRARY_PATH=/usr/local/lib64 \
```
uberhacker, line 37: Please use crew constants to abstract instead of hardcoded directories.  For example, `/usr/local/lib64` should be `#{CREW_LIB_PREFIX}` instead since that directory only exists on x86_64 architecture.  Actually, you don't need `-DCMAKE_LIBRARY_PATH=/usr/local/lib64` since that is already included in `#{CREW_CMAKE_OPTIONS`.

packages/ncurses.rb
uberhacker: We'll also need to bump the version.

uberhacker:
We'll need to update `ncurses` in `install.sh` also.

packages/readline.rb
kjperkin: Can do

packages/pcre.rb
kjperkin: Does it hurt to do so? I feel like it's better to be explicit about which packages depend on each-other, core package or no. Certainly right now we can assume all core packages are present on the system, but what about tomorrow or next year?

packages/ncurses.rb
kjperkin: > We'll need new binaries to really get rid of the symlinks. Please remove so new ones can be added.
Can you clarify what you mean by "we'll need new binaries"

> We'll also need to bump the version.
Oops. Thought I had done that.

packages/swi_prolog.rb
kjperkin: Good catch. Sorry I missed that. I probably copy-pasted from my manual compilation and forgot to fix that up.

packages/readline.rb
kjperkin: Hmmm... actually... these all build on each-other. SWI-prolog depends on readline being linked to libtinfo and on pcre being built with --enable-unicode-properties. readline depends on ncurses which is updated too. My gut says it's better to keep the change 'atomic', and I'm unsure how to go about splitting these up nicely. I'm open to ideas!

packages/pcre.rb
kjperkin: Removed it as requested.

packages/readline.rb
kjperkin: Removed as requested.

packages/ncurses.rb
kjperkin: Turns out this isn't needed for this PR at all. Removed it and will create a separate PR later.

packages/swi_prolog.rb
kjperkin: Updated.

packages/readline.rb
kjperkin: I was able to eliminate the change to ncurses, but the the change to pcre is necessary to pass the tests and linking readline to libtinfo is needed to prevent a build failure without resorting to patching the files generated by CMake or resorting to other (IMO uglier) hacks. Let me know if you have any ideas for how to split this up further or if I can mark this resolved. :-)

packages/swi_prolog.rb
kjperkin: I removed all dependencies except libarchive, libdb, and libyaml (which I added a minute ago when I realized we had it available). Please let me know if that dependency set looks good.

satmandu: Dirty script to check for dependencies in your package file:
```
#!/bin/bash
curl -Lf https://github.com/skycocker/chromebrew/raw/master/tools/core_packages.txt > /tmp/core_packages
for i in $(grep depends_on $1 | awk '{print $2}' | sed "s/'//g"); do grep $i /tmp/core_packages ; done
```

packages/pcre.rb
```diff
@@ -26,7 +26,7 @@ def self.build
   --enable-shared \
   --disable-static \
   --with-pic \
-  --enable-utf \
+  --enable-unicode-properties \
```
uberhacker, line 29: Remove binaries and bump version?

satmandu:
Hmm.  Will this require a ncurses rebuild too?

satmandu:
FYI swi_prolog has this error during build in the check section on x86_64:
```
98% tests passed, 1 tests failed out of 65

Total Test time (real) =  28.76 sec

The following tests FAILED:
         13 - swipl:files (Failed)
Errors while running CTest
swi_prolog failed to build: `ctest -j 2` exited with 8
```

Is this an issue @kjperkin ? Or can we just comment it out?

packages/pcre.rb
```diff
@@ -3,41 +3,43 @@
 class Pcre < Package
   description 'The PCRE library is a set of functions that implement regular expression pattern matching using the same syntax and semantics as Perl 5.'
   homepage 'http://pcre.org/'
-  version '8.44-1'
+  @_ver = '8.44'
+  version "#{@_ver}-1"
```
uberhacker, line 7: Shouldn't this be `-2`?  I guess it doesn't matter now that you overwrote the previous binaries on Bintray.  But...now it's difficult to `crew upgrade`.

uberhacker: Never mind.  This is a new package that was never merged.

uberhacker: Never mind.  I guess the update to `-1` wasn't merged yet.

packages/swi_prolog.rb
```diff
@@ -0,0 +1,49 @@
...
+  homepage 'https://www.swi-prolog.org/'
+  version '8.2.4'
+  compatibility 'all'
+  source_url 'https://www.swi-prolog.org/download/stable/src/swipl-8.2.4.tar.gz'
+  source_sha256 'f4bcc78437f9080ab089762e9e6afa7071df7f584c14999b92b9a90a4efbd7d8'
+
+  binary_url ({
+  })
+  binary_sha256 ({
+  })
```
uberhacker, line 15: Missing binaries?

satmandu: Seeing a compile failure on armv7l:
```
[106/106] Generating ../home/boot.prc
FAILED: home/boot.prc
cd /usr/local/tmp/crew/swipl-8.2.4.tar.gz.dir/swipl-8.2.4/builddir/src && /usr/local/bin/cmake -E remove -f /usr/local/tmp/crew/swipl-8.2.4.tar.gz.dir/swipl-8.2.4/builddir/home/boot.prc && /usr/local/tmp/crew/swipl-8.2.4.tar.gz.dir/swipl-8.2.4/builddir/src/swipl -q -O -o /usr/local/tmp/crew/swipl-8.2.4.tar.gz.dir/swipl-8.2.4/builddir/home/boot.prc -b /usr/local/tmp/crew/swipl-8.2.4.tar.gz.dir/swipl-8.2.4/builddir/home/boot/init.pl
[FATAL ERROR: at Tue Feb  9 18:00:49 2021
        WAM Table mismatch: wam_table[8(h_mpz)] == wam_table[10(h_string)]
]
ninja: build stopped: subcommand failed.
PGO: run ninja to complete the build
[1/499] Generating VMI metadata
[2/499] Generating static atoms and functors
[3/418] Checking (git) version
[4/418] Generating ../home/boot.prc
FAILED: home/boot.prc
cd /usr/local/tmp/crew/swipl-8.2.4.tar.gz.dir/swipl-8.2.4/builddir/src && /usr/local/bin/cmake -E remove -f /usr/local/tmp/crew/swipl-8.2.4.tar.gz.dir/swipl-8.2.4/builddir/home/boot.prc && /usr/local/tmp/crew/swipl-8.2.4.tar.gz.dir/swipl-8.2.4/builddir/src/swipl -q -O -o /usr/local/tmp/crew/swipl-8.2.4.tar.gz.dir/swipl-8.2.4/builddir/home/boot.prc -b /usr/local/tmp/crew/swipl-8.2.4.tar.gz.dir/swipl-8.2.4/builddir/home/boot/init.pl
[FATAL ERROR: at Tue Feb  9 18:00:50 2021
        WAM Table mismatch: wam_table[8(h_mpz)] == wam_table[10(h_string)]
]
[5/418] Building C object packages/clib/CMakeFiles/plugin_files.dir/error.c.o
```

satmandu: Just excluding armv7l... this package moves fast enough that if someone wants to make a dev-package, it will probably compile fine on armv7l.

satmandu: Binaries added.

packages/readline.rb
satmandu: Binaries added for all of these.

satmandu:
As far as I'm concerned, this can be converted to ready for review and merged @uberhacker .

satmandu (approved):

packages/swi_prolog.rb
```diff
@@ -0,0 +1,52 @@
+require 'package'
+
+class Swi_prolog < Package
+  description 'SWI-Prolog offers a comprehensive free Prolog environment. Since its start in 1987, SWI-Prolog development has been driven by the needs of real world applications. SWI-Prolog is widely used in research and education as well as commercial applications. Join over a million users who have downloaded SWI-Prolog.'
+  homepage 'https://www.swi-prolog.org/'
+  @_ver = '8.2.4'
+  version @_ver
+  compatibility 'x86_64', 'i686'
```
uberhacker, line 8: I'm thinking this should be `'i686,x86_64'`?  Probably should change the logic to check for compatibility in crew anyway.

packages/swi_prolog.rb
```diff
@@ -0,0 +1,52 @@
...
+  })
+  binary_sha256({
+    i686: 'd8985a0a160baca9d6565d4177a43a840450ef4bb96eefae476a51ce27fec063',
+  x86_64: 'cbb431b0bd28fa1c581071aeb1eeda79b70b4ec3c28c2d07b509054e810cb62b'
+  })
+
+  depends_on 'libyaml'
+  depends_on 'libarchive'
+  depends_on 'libdb'
+  depends_on 'jdk8'
```
uberhacker, line 24: These dependencies will get installed on `aarch64` and `armv7l` even if the package is not compatible.  Eventually, I'd like to add logic in `crew` to avoid this but for now we will need logic as follows:
```
case ARCH
when 'i686', 'x86_64'
  depends_on 'libyaml'
  depends_on 'libarchive'
  depends_on 'libdb'
  depends_on 'jdk8'
end
```

satmandu: Makes sense.

kjperkin:
@satmandu that's very suspicious. I do not have any test errors when I build. I think that needs investigation.

PR #5105, opened 2021-02-09
RFC: install.sh Don't do a recursive chown on all of /usr/local 

satmandu:
- Since /usr/local is preserved across reboots, there may be other things there which Chromebrew has not put there. Additionally, there may be filesystems mounted in there. So only do a chown for the parts of /usr/local we know we will working with.
- (Yes this bit me on my first chromebrew install.)

```
crew_folders="bin cache doc docbook etc include lib lib64 libexec man sbin share tmp var"
for folder in $crew_folders ; do sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}"/"$folder" ; done
sudo chown "$(id -u)":"$(id -g)" "${CREW_PREFIX}"
```


install.sh
```diff
@@ -39,7 +39,9 @@ case "${ARCH}" in
 esac
 
 # This will allow things to work without sudo
-sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}"
+crew_folders="bin cache doc docbook etc include lib lib64 libexec man sbin share tmp var"
+for folder in $crew_folders ; do sudo chown -R "$(id -u)":"$(id -g)" "${CREW_PREFIX}"/"$folder" ; done
+sudo chown "$(id -u)":"$(id -g)" "${CREW_PREFIX}"
```
uberhacker, line 44: How is this different than doing a recursive chown?

uberhacker:
While you're at it, might as well add the recent ncurses update to install.sh.

kjperkin:
How about a `crew fixperms` or a ruby script at the end of the run that goes through the file lists of the installed crew packages and does a chown of select (which TBD) things.

RE mountpoints, if you don't go with the above, I think something like (untested and off the top of my head!!):
```
sudo find /usr/local -xdev -exec chown "$(id -u)":"$(id -g)" {} +
```

would call chown with as many arguments as possible at a time for everything in /usr/local without traversing mountpoints. If you wanted to limit this to just directories or just certain paths, that's also doable with additional options to find.

install.sh
kjperkin: The difference is that children directly under `CREW_PREFIX` that are not in `crew_folders` will not get `chown -R`'d. In particular, if I were to have a /usr/local/mnt directory where I mount my SD card or something, I definitely would not want `install.sh` to chown -R that.

uberhacker: I get that but how do we know these folders are all inclusive?  Also, `lib64` will only work on x86_64 architecture and generate an error on other architectures.  `"${CREW_PREFIX}/$folder"` should work as well.  Extra double quotes can be removed.

uberhacker: Would it be better just to ignore mounted folders and include everything else?

uberhacker: I suppose it would make sense to ignore symlinks as well.

satmandu:
@kjperkin That would be a separate PR...

install.sh
satmandu: AFAIK chown ignores symlinks by default. I don't know how to get it to ignore mounted folders.

satmandu: Changed logic to avoid lib64 when not on x86_64

install.sh
```diff
@@ -39,7 +39,8 @@ case "${ARCH}" in
 esac
 
 # This will allow things to work without sudo
-crew_folders="bin cache doc docbook etc include lib lib64 libexec man sbin share tmp var"
+crew_folders="bin cache doc docbook etc include lib libexec man sbin share tmp var"
+[ "${ARCH}" == "x86_64" ] && crew_folders="bin cache doc docbook etc include lib lib64 libexec man sbin share tmp var"
```
uberhacker, line 43: How about `crew_folders="bin cache doc docbook etc include #{ARCH_LIB} libexec man sbin share tmp var"`?

satmandu:
@uberhacker There's a readline update coming in too... Let's just put those updates together for an update later today...

install.sh
satmandu: Just changed it to the bash version of that...

satmandu:
Also @kjperkin I'd prefer a whitelist method in lieu of a blacklist method to be safer.

satmandu:
@uberhacker URLs updated.

install.sh
uberhacker: I have a bad feeling this is going to end up missing things in the future.

PR #5109, opened 2021-02-10
[SWI-Prolog] Build without jdk integration (fixes #5108)

### LINKED ISSUE #5108: [SWI-Prolog] Remove jdk dependency
## Description
I chose the dependencies to include when creating `swi_prolog.rb` on the grounds that they were either already installed or lightweight. I don't want to have to install compilers/interpreters for two languages (Java and Prolog) just to get the one (Prolog). Until `crew` has support for something like debian "source packages" from which multiple binary packages may be built, I'd rather not have a dependency on the jdk for SWI-Prolog.

kjperkin:
Fixes #5108 

## Description
When writing swi_prolog.rb, I chose dependencies that were (I thought) lightweight and likely to be useful to most users. I don't want to have to install support for two programming languages (Java and Prolog) just to get the one (Prolog). Until crew has support for something like debian "source packages" from which multiple binary packages may be built, I'd rather not have a dependency on the jdk for SWI-Prolog.

## Addtional information
All tests are passing for me, so I consider any test failures to be highly suspect. Please don't merge until all tests are passing or there is an good explanation for why tests are failing.

Works properly:
- [x] x86_64
- [ ] aarch64 (Not tested, I don't have the hardware)

### DIFF
--- a/crew
+++ b/crew
@@ -557,6 +557,10 @@ def unpack (meta)
       puts "Unpacking archive using 'tar', this may take a while..."
       FileUtils.mkdir extract_dir unless Dir.exist?(extract_dir)
       system "tar x#{@verbose}f #{meta[:filename]} -C #{extract_dir}"
+    when /\.deb$/i
+      puts "Unpacking archive using 'ar', this may take a while..."
+      FileUtils.mkdir extract_dir unless Dir.exist?(extract_dir)
+      system "ar -p #{meta[:filename]} data.tar.xz | xz -dc#{@verbose} | tar x#{@verbose} -C #{extract_dir}"
     end
     if meta[:source] == true
       # Check the number of directories in the archive
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.7.7'
+CREW_VERSION = '1.7.8'
 
 ARCH_ACTUAL = `uname -m`.strip
 # This helps with virtualized builds on aarch64 machines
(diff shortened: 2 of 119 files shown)


kjperkin:
Btw, binaries need updating... how do we handle building binaries? Are they built with a dedicated environment or is it just someone builds them and adds them to bintray?

satmandu:
[swi_prolog-8.2.4-1-x86_64.log](https://github.com/skycocker/chromebrew/files/5960357/swi_prolog-8.2.4-1-x86_64.log)
[swi_prolog-8.2.4-1-i686.log](https://github.com/skycocker/chromebrew/files/5960358/swi_prolog-8.2.4-1-i686.log)
[swi_prolog-8.2.4-1-armv7l.log](https://github.com/skycocker/chromebrew/files/5960360/swi_prolog-8.2.4-1-armv7l.log)
Here are the build logs.

Binaries are built by someone who has time to do it. I'm just building these on docker images, and then putting them on bintray. @uberhacker also does compiles and uploads.

kjperkin:
Curious. I'll have to find out how to get more detailed info about what failed and how.

uberhacker:
Are we removing the jdk8 dependency just because the tests don't pass?  Sometimes it's the tests and not the dependencies that need fixed.

kjperkin:
@uberhacker No, I want to eliminate the jdk8 dependency because it's much bigger than the other optional dependencies. The tests failing is just something that is left over from the PR that introduced the package.

uberhacker:
So it sounds like you are sacrificing functionality to save disk space.  I think that may not be the right approach.

uberhacker:
This might be a rhetorical question but is it possible to make jdk8 a build dependency instead?

satmandu:
Build dependencies get installed before the package is installed or built though...

kjperkin:
> So it sounds like you are sacrificing functionality to save disk space. I think that may not be the right approach.

I updated the description with a more detailed explanation of where I'm coming from.

> This might be a rhetorical question but is it possible to make jdk8 a build dependency instead?

> Build dependencies get installed before the package is installed or built though...

I'll give it a shot and see what happens...

kjperkin:
Attempting to load the library when the jvm is not available results in error message that seems to be non-trivial to suppress (but my prolog skills are essentially nonexistant, so I could easily be doing something wrong). I think it's best to just take it out entirely.

kjperkin:
@uberhacker, does the updated description help? I'm not excited at the prospect of taking time to figure out if the error message can be suppressed. I poked at it a little, but came up with nothing.

kjperkin:
The swipl build detects X and Qt5 and automatically builds the bindings for those if they are present. Since these are heavyweight dependencies that aren't part of the core functionality of the language, I'm opting to disable them entirely.

The `files` test began to fail for me a while back, so there seems to be something screwy going on outside of SWI-Prolog. It appears to be locale-related, and setting `LC_ALL=en_US.utf8`, the tests will pass. 

I've built the x86_64 version but don't know where I can upload it.

satmandu:
@kjperkin I added binaries.

satmandu:
@kjperkin Please let us know if they work.

supechicken:
Closing due to inactivity

PR #5112, opened 2021-02-10
[tools]: Add script to catch packages that depend_on core packages

kjperkin:
Inspired by @satmandu's quick script [here](https://github.com/skycocker/chromebrew/pull/5102#discussion_r572872648) I wrote a `gawk` program to go through the packages and find ones that `depend_on` a core package. This can and probably should be replaced with a ruby script, but for the moment, I figured I'd offer it up as a discussion-starter.

To use the script, make sure you have `gawk` installed and do `awk -f tools/check_coredeps.awk`.


uberhacker:
I'm thinking the logic should be the other way around?  We should check to see if a `depends_on '...'` line includes a core package and notify or remove if found?

satmandu:
This is the relevant part of my preflight script which does exactly what @uberhacker suggests.
```
#!/bin/bash
pkg="${1%.*}"
echo "pkg is $pkg"
for i in $(curl -Lsf https://github.com/skycocker/chromebrew/raw/master/tools/core_packages.txt)
do 
sed -i "/^[[:blank:]]*depends_on '$i'/d" /usr/local/lib/crew/packages/"${pkg}".rb
done
```

kjperkin:
I took the snippet @satmandu posted and adapted it to build a regular expression from the package list instead of looping through the package list. It's probably premature optimization, but ¯\_(ツ)_/¯. The script accepts multiple package names on the command line.

@satmandu "preflight script"? Is that something you run on any package file before you push your changes or something?

satmandu:
@kjperkin In the 90s (and maybe still?) in the newspaper world you would run "preflight" software on Quark/InDesign files before sending them to the printer. I'm using the word in the same context.

Yes I run a script which does multiple checks on the package.rb file  before sending the file to my build hosts so that binaries can be generated.

satmandu:
Also, great job documenting the script!

satmandu (approved):

timedroid:
Is there another way things in core_packages.txt get installed? I only see it used in check_updates.sh, but I don't see check_updates.sh called from anywhere.

uberhacker:
@timedroid: Two ways core packages can be installed are below:

1. Add the package to install.sh
2. Add a dependency to one of the other core packages so it is inherited during installation

supechicken:
Closing due to inactivity

PR #5113, opened 2021-02-10
Docbook Updates

satmandu:
- This fixes the rewrite hell which plagued packages.
- Also fakes out the docbook.rb which only moreutils was using
- And thus also updates moreutils.
- put together after consulting both the arch and homebrew package files

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

(everything except binaries in docbook_xsl are uploaded.)

### DIFF
--- a/packages/docbook_xsl_nons.rb
+++ b/packages/docbook_xsl_nons.rb
@@ -1,0 +1,101 @@
+require 'package'
+
+class Docbook_xsl_nons < Package
+  description 'The DocBook XSL Stylesheets package contains XSL stylesheets. These are useful for performing transformations on XML DocBook files.'
+  compatibility 'all'
+  homepage 'https://github.com/docbook/xslt10-stylesheets'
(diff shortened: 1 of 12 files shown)


packages/docbook_xsl_nons.rb
```diff
@@ -0,0 +1,101 @@
+require 'package'
+
+class Docbook_xsl_nons < Package
+  description 'The DocBook XSL Stylesheets package contains XSL stylesheets. These are useful for performing transformations on XML DocBook files.'
+  compatibility 'all'
```
uberhacker, line 5: Can we put this line below version?  I know order doesn't matter but it helps with reviews to have consistency.

satmandu: done

uberhacker:
This is some awesome work @satmandu!  Really appreciate your efforts!

PR #5116, opened 2021-02-10
Create wmctrl.rb

satmandu:
- This might be useful with non-sommelier window managers
```
wmctrl -m
Name: Sommelier
Class: N/A
PID: N/A
Window manager's "showing the desktop" mode: N/A
```

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/wmctrl.rb
+++ b/packages/wmctrl.rb
@@ -1,0 +1,47 @@
+require 'package'
+
+class Wmctrl < Package
+  description 'Control your EWMH compliant window manager from command line'
+  homepage 'http://tripie.sweb.cz/utils/wmctrl/'
+  @_ver '1.07'
+  version @_ver
+  compatibility 'all'
+  source_url "http://tripie.sweb.cz/utils/wmctrl/dist/wmctrl-#{@_ver}.tar.gz"
+  source_sha256 'd78a1efdb62f18674298ad039c5cbdb1edb6e8e149bb3a8e3a01a4750aa3cca9'
+
+  binary_url({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/wmctrl-1.07-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/wmctrl-1.07-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/wmctrl-1.07-chromeos-i686.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/wmctrl.rb
```diff
@@ -0,0 +1,47 @@
...
+  })
+
+  depends_on 'libxmu'
+
+  def self.patch
+    # Debian patch
+    patch_url = 'http://archive.debian.org/debian/pool/main/w/wmctrl/wmctrl_1.07-6.diff.gz'
+    patch_sha256 = 'c6560ba645652eab0b16d46c9dbd8e5c9881c1f9d3ac6fa3da2ee3c865b5e6a0'
+
+    system('wget', '--continue', '--no-check-certificate', patch_url, '-O', 'patch.gz')
```
supechicken, line 32: Can we use `curl`?

packages/wmctrl.rb
```diff
@@ -0,0 +1,47 @@
+require 'package'
+
+class Wmctrl < Package
+  description 'Control your EWMH compliant window manager from command line'
+  homepage 'http://tripie.sweb.cz/utils/wmctrl/'
+  @_ver '1.07'
```
supechicken, line 6: Missing `=`

uberhacker: Nice catch @supechicken!

packages/wmctrl.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def self.patch
+    # Debian patch
+    patch_url = 'http://archive.debian.org/debian/pool/main/w/wmctrl/wmctrl_1.07-6.diff.gz'
+    patch_sha256 = 'c6560ba645652eab0b16d46c9dbd8e5c9881c1f9d3ac6fa3da2ee3c865b5e6a0'
+
+    system('wget', '--continue', '--no-check-certificate', patch_url, '-O', 'patch.gz')
+    abort 'Checksum mismatch :/ try again' unless Digest::SHA256.hexdigest(File.read('./patch.gz')) == patch_sha256
+    system('gunzip', 'patch.gz')
+    system 'patch -p 1 < patch'
```
uberhacker, line 35: You could eliminate line 34 with:
````
system 'zcat patch.gz | patch -p1'
```

packages/wmctrl.rb
satmandu: Sure

packages/wmctrl.rb
satmandu: Nice.

PR #5117, opened 2021-02-11
Nss 3.6.1

satmandu:
- not quite sure if this will break anything, but nss.pc is needed for building some gnome stuff. That part is cribbed from the arch pkgbuild file.
- updated nss build variables from the build.sh file, which appear to be different from those needed for a make-based build.
- Also added gyp-next, which is python3 based and can replace gyp

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/gyp_next.rb
+++ b/packages/gyp_next.rb
@@ -1,0 +1,28 @@
+require 'package'
+
+class Gyp_next < Package
+  description 'GYP can Generate Your Projects.'
+  homepage 'https://github.com/nodejs/gyp-next'
+  version '0.8.0'
+  compatibility 'all'
+  source_url 'file:///dev/null'
+  source_sha256 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
+
+  binary_url({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gyp_next-0.8.0-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/gyp_next-0.8.0-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/gyp_next-0.8.0-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/gyp_next-0.8.0-chromeos-x86_64.tar.xz'
+  })
+  binary_sha256({
+    aarch64: '6f9637bbb731df38ebe855839e40e7c390127fc391e37428acf9ab978e73607b',
+     armv7l: '6f9637bbb731df38ebe855839e40e7c390127fc391e37428acf9ab978e73607b',
+       i686: '9662dcd813ad9d11c600d92f00cadc549ea787533c228b34bcf1b28e5cb9d180',
+     x86_64: '4e94350d25225bc3ccc8a2b32d710ccef16ea2ce637969774ccccc8b85d8edd9'
+  })
+
+  def self.install
+    system 'pip3 uninstall -y gyp-next'
+    system "pip3 install --upgrade --no-warn-script-location gyp-next --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR}"
+  end
+end
--- a/packages/nss.rb
+++ b/packages/nss.rb
@@ -24,8 +24,6 @@ class Nss < Package
 
   depends_on 'gyp_next' => :build
   depends_on 'nspr'
-  depends_on 'sqlite'
-  depends_on 'pkgconfig' => :build
 
   def self.build
     ENV['opt_build'] = '1'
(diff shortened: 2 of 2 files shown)


packages/nss.rb
```diff
@@ -3,49 +3,74 @@
...
+    aarch64: '812f36d5c1875cf8803097265483b8a558194c6f0778977d55637407682ba51c',
+     armv7l: '812f36d5c1875cf8803097265483b8a558194c6f0778977d55637407682ba51c',
+       i686: 'e724b986a9a2edce790a8bef0b89d096ff46cb93815fc2d7bf8beb950af851f5',
+     x86_64: 'ba50721b2e968c49fbd5d4a90112ff27374903d33dbcfc3f1877b2c7074adbcd'
   })
 
-  depends_on 'gyp' => :build
+  depends_on 'gyp_next' => :build
   depends_on 'nspr'
   depends_on 'sqlite'
```
uberhacker, line 27: Core package?

packages/nss.rb
```diff
@@ -3,49 +3,74 @@
...
+  depends_on 'pkgconfig' => :build
```
uberhacker, line 28: Core package?

packages/gyp_next.rb
```diff
@@ -0,0 +1,28 @@
...
+  binary_sha256({
+    aarch64: '6f9637bbb731df38ebe855839e40e7c390127fc391e37428acf9ab978e73607b',
+     armv7l: '6f9637bbb731df38ebe855839e40e7c390127fc391e37428acf9ab978e73607b',
+       i686: '9662dcd813ad9d11c600d92f00cadc549ea787533c228b34bcf1b28e5cb9d180',
+     x86_64: '4e94350d25225bc3ccc8a2b32d710ccef16ea2ce637969774ccccc8b85d8edd9'
+  })
+
+  def self.install
+    system 'pip3 uninstall -y gyp-next'
+    system "pip3 install --upgrade --no-warn-script-location gyp-next --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR}"
```
uberhacker, line 26: Ironic that nodejs needs a python build app.

packages/nss.rb
```diff
@@ -3,49 +3,74 @@
...
+    @ARCH_CFLAGS = if ARCH == 'armv7l'
+                     ''
+                   else
+                     '-flto=auto'
+                   end
+    @ARCH_LDFLAGS = if ARCH == 'armv7l'
+                      ''
+                    else
+                      '-flto=auto'
+                    end
```
uberhacker, line 46: `@ARCH_LDFLAGS = @ARCH_CFLAGS`?

packages/nss.rb
```diff
@@ -3,49 +3,74 @@
...
     end
   end
 
   def self.install
     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/include/nss"
     FileUtils.rm Dir.glob('dist/Release/lib/*.so.TOC')
-    FileUtils.mv "dist/Release/lib","dist/Release/#{ARCH_LIB}" unless "#{ARCH_LIB}" == "lib"
-    FileUtils.cp_r Dir.glob('dist/Release/*'), "#{CREW_DEST_PREFIX}"
+    FileUtils.mv 'dist/Release/lib', "dist/Release/#{ARCH_LIB}" unless ARCH_LIB.to_s == 'lib'
+    FileUtils.cp_r Dir.glob('dist/Release/*'), CREW_DEST_PREFIX.to_s
```
uberhacker, line 62: Shouldn't need the `.to_s` conversion.

packages/nss.rb
```diff
@@ -3,49 +3,74 @@
...
     FileUtils.cp_r Dir.glob('dist/public/nss/*'), "#{CREW_DEST_PREFIX}/include/nss/"
+
+    system "sed nss/pkg/pkg-config/nss.pc.in \
+    -e \"s,%libdir%,#{CREW_LIB_PREFIX},g\" \
+    -e \"s,%prefix%,#{CREW_PREFIX},g\" \
+    -e \"s,%exec_prefix%,#{CREW_PREFIX}/bin,g\" \
+    -e \"s,%includedir%,#{CREW_PREFIX}/include/nss,g\" \
+    -e \"s,%NSPR_VERSION%,$(pkg-config --modversion nspr),g\" \
+    -e \"s,%NSS_VERSION%,#{@_ver},g\" | \
+    install -Dm644 /dev/stdin #{CREW_DEST_LIB_PREFIX}/pkgconfig/nss.pc"
```
uberhacker, line 72: Huh?

packages/nss.rb
```diff
@@ -3,49 +3,74 @@
...
+    FileUtils.cp_r Dir.glob('dist/Release/*'), CREW_DEST_PREFIX.to_s
     FileUtils.cp_r Dir.glob('dist/public/nss/*'), "#{CREW_DEST_PREFIX}/include/nss/"
+
+    system "sed nss/pkg/pkg-config/nss.pc.in \
+    -e \"s,%libdir%,#{CREW_LIB_PREFIX},g\" \
+    -e \"s,%prefix%,#{CREW_PREFIX},g\" \
+    -e \"s,%exec_prefix%,#{CREW_PREFIX}/bin,g\" \
+    -e \"s,%includedir%,#{CREW_PREFIX}/include/nss,g\" \
+    -e \"s,%NSPR_VERSION%,$(pkg-config --modversion nspr),g\" \
+    -e \"s,%NSS_VERSION%,#{@_ver},g\" | \
```
uberhacker, line 71: You could use single quotes instead of escaped double quotes.

satmandu: Yup. The double-quote version is from a later stage of troubleshooting another issue with this block. I'll change it back.

packages/nss.rb
satmandu: The sed pipes into that! I was confused as well, but I guess that's how install takes piped input? It works...

packages/nss.rb
satmandu: Yeah that's the linter being overprotective.

packages/nss.rb
uberhacker: I meant this is a core package so no need to include here.

packages/nss.rb
uberhacker: I meant this is a core package so no need to include here.

satmandu:
Should be good to go?

uberhacker:
Remove core dependencies in nss.rb?

satmandu:
aye

PR #5121, opened 2021-02-11
Ruby rebuild add gem update

satmandu:
- Ruby was starting to have some weird connectivity issues, and rebuilds were failing... so I used a static build of ruby to rebuild this, and then rebuilt it again normally.
- Also updated install.sh with this ruby rebuild.
- Also added a gem update in the postinstall.
- This should really have a ```depends_on 'ca_certificates'``` to fix issues with things like ```curl -Lks 'https://git.io/rg-ssl' | ruby``` failing.

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686


uberhacker:
Good call.  Does this also fix the openssl issue?

packages/ruby.rb
```diff
@@ -3,36 +3,32 @@
...
+       i686: '44422c2fbe7c1d4d8d0088d12c1691d33de19fbe10a1461d9b3ce8cf1e2e7e1e',
+     x86_64: '20d4d18d1a184a088abb5547bc94e7f6189229565157bb35c5691cf0f1264f9f'
   })
 
-  depends_on 'readline'
-  depends_on 'ncurses'
-  depends_on 'zlibpkg' => :build
-  depends_on 'openssl' => :build
-  # depends_on 'libyaml' # This is needed to install gems
+  depends_on 'libyaml' # This is needed to install gems
```
uberhacker, line 24: Missing `depends_on 'ca_certificates'`?

satmandu: Should we then add this to install.sh as well? Core packages?

satmandu: (and yes ca_certificates fixes the openssl issue.)

uberhacker: Yes, if this is a dependency for ruby then it needs to be added to install.sh before ruby is executed.  `tools/core_packages.txt` will need to be updated as well.

uberhacker:
Please add ca_certificates and libyaml to `tools/core_packages.txt` and then I can merge this.

satmandu:
@uberhacker done.

PR #4874, opened 2021-01-03
New package: launcher

supechicken:
Changes from #4739 
- Use websocket
- Added some new features

Tested on
- Cloudready v85
- Chrome OS v87

PR #5123, opened 2021-02-12
Allow installing dependencies from language based package manager (pip, gem, npm)

supechicken:
For #4874 

Tested on `x86_64`

### DIFF
--- a/lib/package.rb
+++ b/lib/package.rb
@@ -17,23 +17,35 @@ class Package
     @dependencies
   end
 
-  def self.depends_on (dependency = nil)
+  def self.depends_on (dependency = nil, pkgmanager = nil, pkgopt = nil)
+    return unless dependency
     @dependencies = Hash.new unless @dependencies
-    if dependency
-      # add element in "[ name, [ tag1, tag2, ... ] ]" format
-      if dependency.is_a?(Hash)
-        if dependency.first[1].is_a?(Array)
-          # parse "depends_on name => [ tag1, tag2 ]"
-          @dependencies.store(dependency.first[0], dependency.first[1])
-        else
-          # parse "depends_on name => tag"
-          @dependencies.store(dependency.first[0], [ dependency.first[1] ])
-        end
+    # add element in "[ name, [ tag1, tag2, ... ] ]" format
+    if dependency.is_a?(Hash)
+      if dependency.first[1].is_a?(Array)
+        # parse "depends_on name => [ tag1, tag2 ]"
+        @dependencies.store(dependency.first[0], dependency.first[1])
       else
-        # parse "depends_on name"
-        @dependencies.store(dependency, [])
+        # parse "depends_on name => tag"
+        @dependencies.store(dependency.first[0], [ dependency.first[1] ])
       end
+    else
+      # parse "depends_on name"
+      @dependencies.store(dependency, []) unless pkgmanager
     end
+    
+    if pkgmanager and pkgopt then
+      puts "Installing dependencies using '#{pkgmanager}', please wait..."
+      puts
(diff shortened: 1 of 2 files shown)


lib/package.rb
```diff
@@ -17,23 +17,35 @@ def self.dependencies
...
+      puts "Installing dependencies using '#{pkgmanager}', please wait..."
+      puts
+      system "yes | #{pkgmanager} install #{dependency} #{pkgopt}"
+      puts
+    elsif pkgmanager
+      puts "Installing dependencies using '#{pkgmanager}', please wait..."
+      puts
+      system "yes | #{pkgmanager} install #{dependency}"
+      puts
     end
```
uberhacker, line 47: Simplify?
```
if pkgmanager then
  puts "Installing dependencies using '#{pkgmanager}', please wait..."
  puts
  if pkgopt then
    system "yes | #{pkgmanager} install #{dependency} #{pkgopt}"
  else
    system "yes | #{pkgmanager} install #{dependency}"
  end
  puts
end
```

lib/package.rb
```diff
@@ -17,23 +17,35 @@ def self.dependencies
     @dependencies
   end
 
-  def self.depends_on (dependency = nil)
+  def self.depends_on (dependency = nil, pkgmanager = nil, pkgopt = nil)
```
uberhacker, line 20: So how will this work?  Do you have an example depends_on line with all the options?

uberhacker (changes requested):

lib/package.rb
supechicken: Done

lib/package.rb
supechicken: `depends_on` will add dep to array if it is a `crew` package, otherwise install from other package manager if `pkgmanager` is provided:
```ruby
depends_on <dep> <pkgmanager> (optional) <options> (optional)
```

uberhacker: Do you have a practical example?  Something with actual parameters?

supechicken: Umm...  like:
```ruby
depends_on 'pyOpenSSL', 'pip3', "--root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
```

supechicken: (If only `dep` was provided, dependency will be added to original array)

supechicken: > Umm... like:
> 
> ```ruby
> depends_on 'pyOpenSSL', 'pip3', "--root #{CREW_DEST_DIR} --prefix #{CREW_PREFIX}"
> ```

This will be `pip3 install pyOpenSSL --root /usr/local --prefix /usr/local` in command line

uberhacker: Gotcha, so in this case the dependency will be included in the package metadata.  I'm not so sure this is the best approach, however.  The pyOpenSSL dependency might be needed with another package.  What I want to get away from is leaving a bunch of unused dependencies behind after a package is removed.  I guess it's 6 one way and half a dozen the other.  I do like your idea, however.  We have the `def self.remove` function we could use to tap into other package managers to uninstall as well.  This could let us create separate gem or python crew packages that would simply be a wrapper around those package managers.  These crew packages wouldn't need any metadata either.

kjperkin:
A quick google search turned up [this document](https://www.debian.org/doc/packaging-manuals/python-policy/index.html#contents), which I think may help shed a light on how Debian approaches things like this. If we are going to get into the business of packaging libraries that already have package managers, then I think it requires a lot of careful planning. Otherwise we're going to wind up with `pip install foo` conflicting with `crew install python-foo`.

PR #5126, opened 2021-02-12
ca_certificates: Move postinstall logic to install

satmandu:
- Postsinstall is useless for packages installed early in install.sh
- Add dated certificates.conf like everyone else does.

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l


packages/ca_certificates.rb
```diff
@@ -1,28 +1,40 @@
...
   def self.patch
+    url_patch1 = 'https://gitweb.gentoo.org/repo/gentoo.git/plain/app-misc/ca-certificates/files/ca-certificates-20150426-root.patch'
+    filename_patch1 = 'patch1'
+    sha256sum_patch1 = '0200f41d2c68b5fb0c19783ddd80d806540e136d10f9d4ee4ff3a79b48d70e73'
+    puts 'Downloading patch1'.yellow
+    system('curl', '-s', '-L', '-#', url_patch1, '-o', filename_patch1)
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless
+    Digest::SHA256.hexdigest(File.read(filename_patch1)) == sha256sum_patch1
+    puts 'patch1 archive downloaded'.lightgreen
+    system 'patch -p 3 < patch1'
```
uberhacker, line 36: Might as well use the variable here too:
```
puts filename_patch1 + ' archive downloaded'.lightgreen
system "patch -p3 < #{filename_patch1}"
```

packages/ca_certificates.rb
```diff
@@ -1,28 +1,40 @@
...
-      armv7l: '75d332ca6053e5e12d34eb5d72da6daab50d650f6733d1bd9668bed9af802e30',
-        i686: 'e61ae07f85f64f0c53fa922714ac30579c77db5f2e9fa266e4e6059b502f5475',
-      x86_64: '7a782fdba9f89b077004dab99c15367bb999ed1b19485e32a22be2e79c4f3ac3',
+     aarch64: '412f0ca92aaafb7e45543635fdfcda9bf937a8b659d0a10403f4f7e827ddd98f',
+      armv7l: '412f0ca92aaafb7e45543635fdfcda9bf937a8b659d0a10403f4f7e827ddd98f',
+        i686: '527721877820f2be4f83e13b3aa3b237e1bf1f3263e636241f122d9c001f9c08',
+      x86_64: 'f2fa591a9ec3aa7c43a1fd58081f1f6a1b9ef9f8ee4ae0c520b17202648fd395',
   })
 
+  depends_on 'bash' => :build
```
uberhacker, line 25: Why is a newer version of bash needed?

packages/ca_certificates.rb
```diff
@@ -45,13 +57,18 @@ def self.install
...
     FileUtils.mkdir_p("#{CREW_DEST_PREFIX}/share/ca-certificates/")
-    system "touch #{CREW_DEST_PREFIX}/etc/ca-certificates.conf"
     system "make DESTDIR=#{CREW_DEST_DIR} install"
-    FileUtils.ln_s Dir.glob("#{CREW_PREFIX}/share/ca-certificates/mozilla/*"),
-                   "#{CREW_DEST_PREFIX}/etc/ssl/certs/"
-  end
-
-  def self.postinstall
-    system 'update-ca-certificates --hooksdir "" --fresh'
+    @date_temp = `date -u`
```
uberhacker, line 61: You could chomp here too.  Then every variable reference doesn't need it.
```
@date_temp = `date -v`.chomp
```

satmandu: Nice.

packages/ca_certificates.rb
satmandu: It's not at all. I'm just trying to avoid as much as possible dependencies on programs that use libraries in /usr/lib64.

At some point if we switch to a new glibc that would be useful...

satmandu: Happy to remove if it is an issue.

packages/ca_certificates.rb
satmandu: Makes sense. And why not show the actual file name too...

packages/ca_certificates.rb
uberhacker: I'm thinking it may be a problem if Chrome OS is using a different version out-of-the-box.  If there is no specific reason for using it, then I think it should be removed.  Would this require new binaries?  I think maybe?

uberhacker: Make sure you remove the bash package before you create new binaries. :)

satmandu: Why new binaries? We're just changing the requirement for building here.... let me just adjust the script used too...

satmandu: ACTUALLY, I'm not even using bash any more... I was previously using a shell script I was running from bash... but that's no longer the case, and the heredoc used now dispenses with that logic entirely, so the bash dep was actually irrelevant and extraneous.

Sigh.

satmandu (commented):
accidental review...

packages/ca_certificates.rb
satmandu: Whoops. Just realized you should have put ```date-u``` here.

PR #5130, opened 2021-02-13
Add krb5 to EARLY_PACKAGES

### LINKED ISSUE #5129: curl error during install
`curl` error during install
```
Downloading curl...
Verifying curl...
curl-7.75.0-chromeos-x86_64.tar.xz: OK
Extracting curl (this may take a while)...
Installing curl (this may take a while)...
Adding new information on curl to device.json...
Downloading git...
curl: error while loading shared libraries: libcom_err.so.3: cannot open shared object file: No such file or directory
```

After some diagnosis, here's what I think is happening:
When installing before this point, the script uses `/usr/bin/curl`. After curl is bootstrapped, as shown in the output, curl is called again, but this time `/usr/local/bin/curl` is preferred, however that curl is missing `libcom_err.so.3`. To fix this error, just specify `/usr/bin/curl` instead of just `curl` throughout the install script.

satmandu:
Fixes #5129 (unless there are other libraries which are needed.)
- krb5 is needed by curl, since it provides ```libcom_err.so.3```
- krb5 needed to be recompiled with ```localstatedir``` set. That is where the krb5kdc folder is, which was defaulting to ```/run/krb5kdc```, which is not a writable folder for tar on most systems, and since install.sh exits on error, this was causing install.sh crash otherwise.
- Normal installs through crew don't show the files outside /usr/local, so we don't see this issue otherwise.
- This PR has krb5 1.19 compiled with localstatedir, and the new package appears to only use files in folders under /usr/local.

```
crew whatprovides libcom_err
e2fsprogs: /usr/local/lib64/libcom_err.a
krb5: /usr/local/lib64/libcom_err.so
krb5: /usr/local/lib64/libcom_err.so.3
krb5: /usr/local/lib64/libcom_err.so.3.0
```
@saltedcoffii  Can you try installing with this: https://github.com/satmandu/chromebrew/blob/add_k5b5_to_install.sh/install.sh

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -17,7 +17,7 @@ CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 
 EARLY_PACKAGES="gcc10 llvm brotli c_ares libcyrussasl libiconv libidn2 \
 libmetalink libnghttp2 libpsl libssh2 libtirpc libunistring openldap \
-rtmpdump zstd ncurses ca_certificates ruby libffi openssl nettle curl \
+rtmpdump zstd ncurses ca_certificates ruby libffi openssl nettle krb5 curl \
 git icu4c libedit"
 
 ARCH="$(uname -m)"
--- a/packages/krb5.rb
+++ b/packages/krb5.rb
@@ -3,38 +3,38 @@ require 'package'
 class Krb5 < Package
   description 'Kerberos is a network authentication protocol.'
   homepage 'https://web.mit.edu/kerberos'
-  version '1.18.3'
+  version '1.19'
   compatibility 'all'
-  source_url 'http://web.mit.edu/kerberos/dist/krb5/1.18/krb5-1.18.3.tar.gz'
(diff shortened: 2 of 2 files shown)


saltedcoffii:
```
Downloading krb5...
Verifying krb5...
krb5-1.18.3-chromeos-x86_64.tar.xz: OK
Extracting krb5 (this may take a while)...
Installing krb5 (this may take a while)...
tar: ./run/krb5kdc: Cannot mkdir: Permission denied
```

satmandu:
Huh. Taking a look.

satmandu:
ok that's an issue with the krb5 package not having a localstatedir set properly. Let me do a rebuild of that, as that would be the proper fix.

satmandu:
krb5 rebuild has /run folder moved, so I'll add it to this PR shortly as soon as I'm done building...

satmandu:
If this doesn't work, see https://github.com/skycocker/chromebrew/issues/5129#issuecomment-778558124

PR #5131, opened 2021-02-13
install.sh: Add more early packages

### LINKED ISSUE #5129: curl error during install

satmandu:
Fixes #5129
 - as per discussion in https://github.com/skycocker/chromebrew/issues/5129 adding ```p11kit libtasn1 gnutls``` before curl to ```EARLY_PACKAGES```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -17,8 +17,8 @@ CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 
 EARLY_PACKAGES="gcc10 llvm brotli c_ares libcyrussasl libiconv libidn2 \
 libmetalink libnghttp2 libpsl libssh2 libtirpc libunistring openldap \
-rtmpdump zstd ncurses ca_certificates ruby libffi openssl nettle krb5 curl \
-git icu4c libedit"
+rtmpdump zstd ncurses ca_certificates ruby libffi openssl nettle krb5 \
+p11kit libtasn1 gnutls curl git icu4c libedit"
 
 ARCH="$(uname -m)"
 


uberhacker:
We need to test these things before submitting PRs in the future. :)

supechicken:
I think we need a reliable dependency system instead of giving them a higher priority at install to "fix" the issue (the `p11kit` already a dependency of `gnutls`, but `gnutls` was installed first), by the way I am working on it now :)

satmandu:
@supechicken my two cents:

We should make a list of symbol exports from libraries, and then see what libraries and symbols apps and other libraries want.

That would be the ideal for building a dependency tree... So we can tell if a library replacement doesn't have the same symbols, and figure out what is going to break.

(Debian already records symbols for libraries...)

But at least doing dependencies based upon requested libraries alone with be a good first step.

satmandu:
@uberhacker testing install.sh at some level is going to be difficult, since the gold standard would be a lab with actual hardware to test on, separate from our own hardware...

nduchuyvp123:
I think we need add "dpkg" before curl to EARLY_PACKAGES

uberhacker:
@nduchuyvp123:  Why would we need dpkg?  This is not a Debian package manager.

nduchuyvp123:
> @nduchuyvp123: Why would we need dpkg? This is not a Debian package manager.

When we have dpkg, we can install .deb package

satmandu:
Since ```deb```s are neither a default package format for the Chromebrew project, it seems WIDELY out of scope to add it to ```EARLY_PACKAGES```. You're welcome to install it later, but there's no reason to make it (and it's dependencies) a mandatory early install before the package list is even updated during the Chromebrew install process.

nduchuyvp123:
I have a .deb or .rpm package, how can I install it?

satmandu:
@nduchuyvp123 I suggest opening an issue about that and trying to get some feedback there. Commenting on a merged commit isn't the best place for this sort of a discussion.

satmandu:
Actually, I missed this: https://github.com/skycocker/chromebrew/pull/5247/files/8e7e97dd9671d79c4c0e99a3e3fb28375fe63700

But this should still be an opened issue. dpkg should probably be added elsewhere...

PR #5135, opened 2021-02-14
install.sh: More carefully extract from tar in early_packages

satmandu:
- ```Install.sh``` broke late last week in part due to some packages (namely the older ```krb5``` package) having files outside ```/usr/local```. The ```extract_install ``` function in install.sh does not check to see what files are in the tar package, and tries to copy everything extracted over to ```/```.  If files are outside ```/usr/local``` this errors out if that extraction attempts to copy over to a read-only folder. This leads to the script erroring out.
- This PR adds logic to only copy over the whitelisted folders in /usr/local which we are chowning to chronos.
- Other package files may have files outside /usr/local, and this way those won't be a problem.
- This isn't a problem when crew installs packages because it just ignores the error and copies over what it can. (Well, crew also uses similar logic, but it could be fine tuned further too similarly.)
- There's an array expansion which throws an error in shellcheck here, but that error is being intentionally ignored.

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l


uberhacker:
We are making this way more complicated than necessary.  I believe we need to change the problem package, not the process as a work around.

satmandu:
Perhaps? Is it better to sanitze packages at generation time or at install time?

A package which is malformed for the purposes of the install function is a problem so long as we don't have any logic at package generation time keeping such packages from being created.

Ideally crew should error out before allowing such a package to be created... but I don't want to go through thousands of packages untarring each one to check which ones are malformed in such a manner... though I guess it wouldn't be hard to do in an automated fashion. (The chances of this happening again are pretty low... but we should never have hit this in the first place.)

(As it was it looks like this was a problem with the krb5 source... with the compile options it had it should never have produced files outside /usr/local in the first place.)

uberhacker:
Packages should be sanitized before they are packaged.  This should be done during testing anyway.  Why not solve the root problem?  We don't really need to be anal about all packages, just the core packages.  Fix krb5 and problem solved.

satmandu:
krb5 is already fixed. But there's no output that would let us know currently that the package is fixed.

Maybe just move this logic to crew then to the portion which packages?

install.sh
```diff
@@ -165,10 +173,10 @@ fi
 for i in $(seq 0 $((${#urls[@]} - 1))); do
```
supechicken, line 173: Use bash built-in method?
```shell
# print array length
for i in ${!urls[@]}; do
```

install.sh
```diff
@@ -122,7 +123,14 @@ function extract_install () {
     echo "Extracting ${1} (this may take a while)..."
     tar xpf ../"${2}"
     echo "Installing ${1} (this may take a while)..."
-    tar cpf - ./*/* | (cd /; tar xp --keep-directory-symlink -f -)
+    local TarPaths=()
+    for folder in $crew_folders; do
+      if [ -d ."${CREW_PREFIX}"/"$folder" ]; then
+        TarPaths+=(".${CREW_PREFIX}/${folder}")
```
supechicken, line 129: Can we make all folders at first? This will be more simple :)

supechicken: (we can add it to `# prepare directories` part)

install.sh
```diff
@@ -122,7 +123,14 @@ function extract_install () {
...
+      fi
+    done
+    # The following array expansion is intentionally unquoted.
+    tar cpf - ${TarPaths[@]} | (cd /; tar xp --keep-directory-symlink -f -)
```
supechicken, line 133: Actually, we can only copy the files under `./usr/local` to fix the problem:
```shell
tar cpf - ./usr/local/* | (cd /; tar xp --keep-directory-symlink -f -)
```

install.sh
satmandu: The point was to only do it to folders which have been chowned, otherwise you might get a permissions error. This protects from that!

install.sh
uberhacker: Good idea but this would be even better:
```
tar cpf - "${CREW_PREFIX}/*" | (cd /; tar xp --keep-directory-symlink -f -)
```

satmandu: Why glob at all then?  

```tar cpf - "${CREW_PREFIX}" | (cd /; tar xp --keep-directory-symlink -f -)```

Should work just as well, right?

Though again, this does nothing to prevent a package from spewing stuff all over /usr/local

uberhacker: Why are you worried about a package spewing files in `/usr/local`?  It should be a dumping grounds with no restrictions.

satmandu: Because _I_ use ```/usr/local``` as a not-dumping ground with various mounts and symlinks to other places. :)

_Also I'm trying to teach a 3 year old to not use places as dumping grounds, and there's probably some part of me trying to tilt at windmills and establish order somewhere._

uberhacker:
@satmandu: Is this still an issue?

satmandu:
I've moved on to refactoring other parts... let me see if I can pull this back in...

PR #5144, opened 2021-02-16
[New Package] flatpak

saltedcoffii:
Flatpak! I haven't quite gotten it to work yet, but I'd love any and all help. It builds fine, but running it is another issue.
Works:
- Adding flathub under the user profile `remote-add --user`
- Installing flatpaks under the user profile

Doesn't work:
- Adding flathub under the system profile `remote-add --system`
- Installing flatpaks under the system profile
- Starting flatpaks (seems to be an issue with bubblewrap)

Any and all help is welcome!

### DIFF
--- a/packages/bubblewrap.rb
+++ b/packages/bubblewrap.rb
@@ -9,32 +9,45 @@ class Bubblewrap < Package
   source_url "https://github.com/containers/bubblewrap/releases/download/v#{@_ver}/bubblewrap-#{@_ver}.tar.xz"
   source_sha256 'b9c69b9b1c61a608f34325c8e1a495229bacf6e4a07cbb0c80cf7a814d7ccc03'
 
-  binary_url({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-x86_64.tar.xz'
+  binary_url ({
+     aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-armv7l.tar.xz',
+      armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-armv7l.tar.xz',
+        i686: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-i686.tar.xz',
+      x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-x86_64.tar.xz',
   })
-  binary_sha256({
-    aarch64: 'c37463f10eaab1df36f8d2593d7cb252bc9558c34a4d9763f20418522e65e296',
-     armv7l: 'c37463f10eaab1df36f8d2593d7cb252bc9558c34a4d9763f20418522e65e296',
-       i686: 'cbc8628fc92ce28e2c3e8c790e6e0af7475486316f863d78391228e022e646ba',
-     x86_64: '8dcbe19bcac8722f5d8257bbc6e26e93d86aa959fe2b31ba78a5ce95575cea76'
+  binary_sha256 ({
+     aarch64: 'a7a003bbb434206346b0e5cf068e0d2cf6231dcafd47c2d4774e0514b04d26d4',
+      armv7l: 'a7a003bbb434206346b0e5cf068e0d2cf6231dcafd47c2d4774e0514b04d26d4',
+        i686: '32bcaff9feccd26cf1379212ea8050d99a271a4e097923aab48638d29d0aef90',
+      x86_64: '30b1458d68567c43b23251167a3488b7b05f7542b4b8b5d0278b026646e01d24',
   })
+
 
   depends_on 'libcap' => :build
 
+  def self.patch
+    system "sed -i '/SUDO_BIN/d' Makefile.in"
+  end
+
   def self.build
+    system './configure --help'
     system "env CFLAGS='-flto=auto' CXXFLAGS='-flto=auto'  LDFLAGS='-flto=auto' \
       ./configure \
       --disable-maintainer-mode \
       --disable-man \
-      --with-priv-mode=setuid \
+      --with-priv-mode=none \
       --enable-sudo"
     system 'make'
   end
 
   def self.install
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    FileUtils.install "#{CREW_DEST_PREFIX}/bin/bwrap", "#{CREW_DEST_PREFIX}/bin/bwrap.elf", mode: 0o755
+    @bwrap_sh = <<~BWRAP_HEREDOC
+      #!/bin/bash
+      sudo setcap cap_sys_admin+eip "#{CREW_PREFIX}/bin/bwrap.elf"
+      sudo -E  "#{CREW_PREFIX}/bin/bwrap.elf" -- "\$*"
+    BWRAP_HEREDOC
+    IO.write("#{CREW_DEST_PREFIX}/bin/bwrap", @bwrap_sh, perm: 0o755)
   end
 end
--- a/packages/bubblewrap.rb
+++ b/packages/bubblewrap.rb
@@ -3,7 +3,7 @@ require 'package'
 class Bubblewrap < Package
   description 'bubblewrap works by creating a new, completely empty, mount namespace'
   homepage 'https://github.com/containers/bubblewrap'
-  version '0.4.1'
+  version '0.4.1_1'
   compatibility 'all'
   source_url 'https://github.com/containers/bubblewrap/releases/download/v0.4.1/bubblewrap-0.4.1.tar.xz'
   source_sha256 'b9c69b9b1c61a608f34325c8e1a495229bacf6e4a07cbb0c80cf7a814d7ccc03'
--- a/packages/bubblewrap.rb
+++ b/packages/bubblewrap.rb
@@ -9,19 +9,18 @@ class Bubblewrap < Package
   source_url "https://github.com/containers/bubblewrap/releases/download/v#{@_ver}/bubblewrap-#{@_ver}.tar.xz"
   source_sha256 'b9c69b9b1c61a608f34325c8e1a495229bacf6e4a07cbb0c80cf7a814d7ccc03'
 
-  binary_url ({
-     aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-armv7l.tar.xz',
-      armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-armv7l.tar.xz',
-        i686: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-i686.tar.xz',
-      x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-x86_64.tar.xz',
+  binary_url({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-x86_64.tar.xz'
   })
-  binary_sha256 ({
-     aarch64: '5bb3cca606a888b7881821eeff96aa28408cb820a6acf1ad788ede5b3ded6fce',
-      armv7l: '5bb3cca606a888b7881821eeff96aa28408cb820a6acf1ad788ede5b3ded6fce',
-        i686: '071ccd273b04892dc2a633784b6cdb04edded34f9f664bf09d3b2b8228a023d5',
-      x86_64: '1fda7e2d705a6f9ef55d2158de6c172d4d78e562566b0a8177d269b3c6df520e',
+  binary_sha256({
+    aarch64: '5bb3cca606a888b7881821eeff96aa28408cb820a6acf1ad788ede5b3ded6fce',
+     armv7l: '5bb3cca606a888b7881821eeff96aa28408cb820a6acf1ad788ede5b3ded6fce',
+       i686: '071ccd273b04892dc2a633784b6cdb04edded34f9f664bf09d3b2b8228a023d5',
+     x86_64: '1fda7e2d705a6f9ef55d2158de6c172d4d78e562566b0a8177d269b3c6df520e'
   })
-
 
   depends_on 'libcap' => :build
   depends_on 'dconf'
--- a/packages/bubblewrap.rb
+++ b/packages/bubblewrap.rb
@@ -8,21 +8,12 @@ class Bubblewrap < Package
   source_url 'https://github.com/containers/bubblewrap/releases/download/v0.4.1/bubblewrap-0.4.1.tar.xz'
   source_sha256 'b9c69b9b1c61a608f34325c8e1a495229bacf6e4a07cbb0c80cf7a814d7ccc03'
 
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: 'eba5493670c286fae9c71814b32613d7c7a951b037680a3f2fbf853c69c53db7',
-     armv7l: 'eba5493670c286fae9c71814b32613d7c7a951b037680a3f2fbf853c69c53db7',
-       i686: '325f997d3d68713b29b8db744b278a9d9b3fe5649d934cfb2f3091a8b66a9b50',
-     x86_64: '4bdc8a377adb978ddd56a3600e07eed238837e39011b3c8c72f69ef1cf9dc6a8',
-  })
-
   def self.build
-    system "./configure #{CREW_OPTIONS} --disable-maintainer-mode --disable-man"
+    system "./configure #{CREW_OPTIONS} \
+            --disable-maintainer-mode \
+            --disable-man \
+            --with-priv-mode=setuid \
+            --enable-sudo"
     system 'make'
   end
 
--- a/packages/bubblewrap.rb
+++ b/packages/bubblewrap.rb
@@ -47,6 +47,7 @@ class Bubblewrap < Package
       #!/bin/bash
       sudo chown root "#{CREW_PREFIX}/bin/bwrap.elf"
       sudo chmod +s "#{CREW_PREFIX}/bin/bwrap.elf"
+      unset GDK_PIXBUF_MODULE_FILE
       #{CREW_PREFIX}/bin/bwrap.elf "\$@"
       sudo chown chronos "#{CREW_PREFIX}/bin/bwrap.elf"
     BWRAP_HEREDOC
--- a/packages/bubblewrap.rb
+++ b/packages/bubblewrap.rb
@@ -47,7 +47,6 @@ class Bubblewrap < Package
       #!/bin/bash
       sudo chown root "#{CREW_PREFIX}/bin/bwrap.elf"
       sudo chmod +s "#{CREW_PREFIX}/bin/bwrap.elf"
-      unset GDK_PIXBUF_MODULE_FILE
       #{CREW_PREFIX}/bin/bwrap.elf "\$@"
       sudo chown chronos "#{CREW_PREFIX}/bin/bwrap.elf"
     BWRAP_HEREDOC
--- a/packages/bubblewrap.rb
+++ b/packages/bubblewrap.rb
@@ -16,10 +16,10 @@ class Bubblewrap < Package
      x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-x86_64.tar.xz'
   })
   binary_sha256({
-    aarch64: '5bb3cca606a888b7881821eeff96aa28408cb820a6acf1ad788ede5b3ded6fce',
-     armv7l: '5bb3cca606a888b7881821eeff96aa28408cb820a6acf1ad788ede5b3ded6fce',
-       i686: '071ccd273b04892dc2a633784b6cdb04edded34f9f664bf09d3b2b8228a023d5',
-     x86_64: '1fda7e2d705a6f9ef55d2158de6c172d4d78e562566b0a8177d269b3c6df520e'
+    aarch64: '7c3f2c4a332342b9d7aa15d6aa776a38d1dc94a8c4c904a81a64f7e7ca45cedd',
+     armv7l: '7c3f2c4a332342b9d7aa15d6aa776a38d1dc94a8c4c904a81a64f7e7ca45cedd',
+       i686: '75795e2c15720ad488335de59b94c3577102b8ba3ca5e855b5b132f3f213cce7',
+     x86_64: '4c3f0aaec3afc4a1ba48d10dfb87208b47d11f2047fe11d6dce5056ff4dce79f'
   })
 
   depends_on 'libcap' => :build
@@ -32,7 +32,7 @@ class Bubblewrap < Package
   def self.build
     system './configure --help'
     system "env CFLAGS='-flto=auto' CXXFLAGS='-flto=auto'  LDFLAGS='-flto=auto' \
-      ./configure \
+      ./configure #{CREW_OPTIONS} \
       --disable-maintainer-mode \
       --disable-man \
       --with-priv-mode=setuid \
--- a/packages/bubblewrap.rb
+++ b/packages/bubblewrap.rb
@@ -16,14 +16,15 @@ class Bubblewrap < Package
       x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/bubblewrap-0.4.1-1-chromeos-x86_64.tar.xz',
   })
   binary_sha256 ({
-     aarch64: 'a7a003bbb434206346b0e5cf068e0d2cf6231dcafd47c2d4774e0514b04d26d4',
-      armv7l: 'a7a003bbb434206346b0e5cf068e0d2cf6231dcafd47c2d4774e0514b04d26d4',
-        i686: '32bcaff9feccd26cf1379212ea8050d99a271a4e097923aab48638d29d0aef90',
-      x86_64: '30b1458d68567c43b23251167a3488b7b05f7542b4b8b5d0278b026646e01d24',
+     aarch64: '5bb3cca606a888b7881821eeff96aa28408cb820a6acf1ad788ede5b3ded6fce',
+      armv7l: '5bb3cca606a888b7881821eeff96aa28408cb820a6acf1ad788ede5b3ded6fce',
+        i686: '071ccd273b04892dc2a633784b6cdb04edded34f9f664bf09d3b2b8228a023d5',
+      x86_64: '1fda7e2d705a6f9ef55d2158de6c172d4d78e562566b0a8177d269b3c6df520e',
   })
 
 
   depends_on 'libcap' => :build
+  depends_on 'dconf'
 
   def self.patch
     system "sed -i '/SUDO_BIN/d' Makefile.in"
@@ -35,7 +36,7 @@ class Bubblewrap < Package
       ./configure \
       --disable-maintainer-mode \
       --disable-man \
-      --with-priv-mode=none \
+      --with-priv-mode=setuid \
       --enable-sudo"
     system 'make'
   end
@@ -45,8 +46,10 @@ class Bubblewrap < Package
     FileUtils.install "#{CREW_DEST_PREFIX}/bin/bwrap", "#{CREW_DEST_PREFIX}/bin/bwrap.elf", mode: 0o755
     @bwrap_sh = <<~BWRAP_HEREDOC
       #!/bin/bash
-      sudo setcap cap_sys_admin+eip "#{CREW_PREFIX}/bin/bwrap.elf"
-      sudo -E  "#{CREW_PREFIX}/bin/bwrap.elf" -- "\$*"
+      sudo chown root "#{CREW_PREFIX}/bin/bwrap.elf"
+      sudo chmod +s "#{CREW_PREFIX}/bin/bwrap.elf"
+      #{CREW_PREFIX}/bin/bwrap.elf "\$@"
+      sudo chown chronos "#{CREW_PREFIX}/bin/bwrap.elf"
     BWRAP_HEREDOC
     IO.write("#{CREW_DEST_PREFIX}/bin/bwrap", @bwrap_sh, perm: 0o755)
   end
--- a/packages/bubblewrap.rb
+++ b/packages/bubblewrap.rb
@@ -3,17 +3,34 @@ require 'package'
 class Bubblewrap < Package
   description 'bubblewrap works by creating a new, completely empty, mount namespace'
   homepage 'https://github.com/containers/bubblewrap'
-  version '0.4.1_1'
+  @_ver = '0.4.1'
+  version "#{@_ver}-1"
   compatibility 'all'
-  source_url 'https://github.com/containers/bubblewrap/releases/download/v0.4.1/bubblewrap-0.4.1.tar.xz'
+  source_url "https://github.com/containers/bubblewrap/releases/download/v#{@_ver}/bubblewrap-#{@_ver}.tar.xz"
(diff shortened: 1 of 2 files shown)


packages/bubblewrap.rb
```diff
@@ -8,21 +8,12 @@ class Bubblewrap < Package
...
-     x86_64: '4bdc8a377adb978ddd56a3600e07eed238837e39011b3c8c72f69ef1cf9dc6a8',
-  })
-
   def self.build
-    system "./configure #{CREW_OPTIONS} --disable-maintainer-mode --disable-man"
+    system "./configure #{CREW_OPTIONS} \
+            --disable-maintainer-mode \
+            --disable-man \
+            --with-priv-mode=setuid \
+            --enable-sudo"
```
uberhacker, line 16: If we need new binaries, we will need to bump the version.

packages/flatpak.rb
```diff
@@ -0,0 +1,41 @@
...
+    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+  end
+
+  def self.postinstall
+    puts
+    puts "Configuring flathub".lightblue
+    system "flatpak remote-add --if-not-exists --user flathub https://flathub.org/repo/flathub.flatpakrepo"
+    puts
+  end
+end
```
uberhacker, line 41: I feel the same way about flatpak as I do about dpkg and rpm.  No need for another package manager.

saltedcoffii:
Flatpak has the ability to provide packages that chromebrew would be unable to provide for whatever reason. Unless you plan on porting every single package in the flathub to chromebrew, I think flatpak would be extremely advantageous. Unlike dpkg and rpm, flatpak is designed to work alongside another package manager, and is the only way, on chromeos, to sandbox proprietary apps. snapd, the other sandboxing package manager, requires systemd and chromeos uses upstart. Should we get flatpak working, I think it would enhance the chromebrew expereience and provide countless other uses for installing chromebrew. Also, flatpak doesn't in any way conflict with chromebrew's installed files like dpkg and rpm can. Flatpak installs data into `~/.local/share/flatpak` or `/usr/local/var/lib/flatpak`, nowhere else. Please reconsider.

uberhacker:
I'm certainly willing to keep this discussion going.  Ideally, I'd like to figure out how to install flatpak apps as chromebrew packages with some kind of wrapper.  Most of these other package managers need dependencies that could overlap with crew.  That would create a potentially huge unnecessary overhead burden with already notoriously small Chromebook hard drives.

packages/flatpak.rb
```diff
@@ -0,0 +1,41 @@
...
+  depends_on 'libostree'
+  depends_on 'polkit'
+  depends_on 'appstream_glib'
+  depends_on 'libseccomp'
+  depends_on 'pyparsing'
+
+  def self.build
+    system "env NOCONFIGURE=1 ./autogen.sh"
+    system "./configure #{CREW_OPTIONS} \
+            --with-system-install-dir=/usr/local/var/lib/flatpak \
```
satmandu, line 21: That should be ```--with-system-install-dir=#{CREW_PREFIX}/var/lib/flatpak \``` no?

packages/flatpak.rb
```diff
@@ -0,0 +1,41 @@
...
+            --enable-sudo \
+            --with-priv-mode=setuid \
```
satmandu, line 23: Does this work with ```--with-priv-mode=none``` ?

packages/bubblewrap.rb
```diff
@@ -3,26 +3,17 @@
...
-       i686: '325f997d3d68713b29b8db744b278a9d9b3fe5649d934cfb2f3091a8b66a9b50',
-     x86_64: '4bdc8a377adb978ddd56a3600e07eed238837e39011b3c8c72f69ef1cf9dc6a8',
-  })
-
   def self.build
-    system "./configure #{CREW_OPTIONS} --disable-maintainer-mode --disable-man"
+    system "./configure #{CREW_OPTIONS} \
+            --disable-maintainer-mode \
+            --disable-man \
+            --with-priv-mode=setuid \
```
satmandu, line 15: Does this work with ```--with-priv-mode=none``` ?

saltedcoffii: it doesn't work either way, unfortunately.

packages/flatpak.rb
saltedcoffii: it doesn't work either way, unfortunately.

packages/bubblewrap.rb
```diff
@@ -3,26 +3,17 @@
...
-  binary_sha256 ({
-    aarch64: 'eba5493670c286fae9c71814b32613d7c7a951b037680a3f2fbf853c69c53db7',
-     armv7l: 'eba5493670c286fae9c71814b32613d7c7a951b037680a3f2fbf853c69c53db7',
-       i686: '325f997d3d68713b29b8db744b278a9d9b3fe5649d934cfb2f3091a8b66a9b50',
-     x86_64: '4bdc8a377adb978ddd56a3600e07eed238837e39011b3c8c72f69ef1cf9dc6a8',
-  })
-
   def self.build
-    system "./configure #{CREW_OPTIONS} --disable-maintainer-mode --disable-man"
+    system "./configure #{CREW_OPTIONS} \
```
satmandu, line 12: Use lto?
```
system "env CFLAGS='-flto=auto' CXXFLAGS='-flto=auto'  LDFLAGS='-flto=auto' \
  ./configure #{CREW_OPTIONS} \
```

packages/flatpak.rb
```diff
@@ -0,0 +1,41 @@
...
+  depends_on 'bubblewrap'
+  depends_on 'libostree'
+  depends_on 'polkit'
+  depends_on 'appstream_glib'
+  depends_on 'libseccomp'
+  depends_on 'pyparsing'
+
+  def self.build
+    system "env NOCONFIGURE=1 ./autogen.sh"
+    system "./configure #{CREW_OPTIONS} \
```
satmandu, line 20: Use lto?
```
system "env CFLAGS='-flto=auto' CXXFLAGS='-flto=auto'  LDFLAGS='-flto=auto' \
  ./configure #{CREW_OPTIONS} \
```

saltedcoffii: if it builds, yes!

packages/bubblewrap.rb
satmandu: got this

packages/flatpak.rb
satmandu: Builds.

satmandu:
@saltedcoffii Added binaries for all so you can test...

satmandu:
Here's where this is at the moment:
```
crew install flatpak
flatpak: Flatpak is a system for building, distributing, and running sandboxed desktop applications on Linux.
https://flatpak.org
Version: 1.10.1
polkit 
Do you agree? [Y/n] y
Proceeding...
polkit: Application development toolkit for controlling system-wide privileges
############################################################################################ 100.0%
Archive downloaded
Caching downloaded archive
Performing pre-install...
Performing install...
Performing post-install...
Archive found in cache
Performing pre-install...
Performing install...
Performing post-install...

Configuring flathub

Note that the directories 

'/usr/local/var/lib/flatpak/exports/share'
'/home/chronos/user/.local/share/flatpak/exports/share'

are not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.


Flatpak installed!
chronos@localhost /usr/local/lib/crew/packages (master *%|SPARSE)$ flatpak install flathub org.gnome.Fractal

Note that the directories 

'/usr/local/var/lib/flatpak/exports/share'
'/home/chronos/user/.local/share/flatpak/exports/share'

are not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.

Looking for matches…
Required runtime for org.gnome.Fractal/x86_64/stable (runtime/org.gnome.Platform/x86_64/3.38) found in remote flathub
Do you want to install it? [Y/n]: Y

org.gnome.Fractal permissions:
    ipc      network      fallback-x11         pulseaudio      wayland
    x11      dri          dbus access [1]

    [1] org.freedesktop.Notifications, org.freedesktop.secrets


        ID                                       Branch     Op    Remote     Download
 1. [✓] org.freedesktop.Platform.GL.default      20.08      i     flathub       95.3 MB / 95.9 MB
        ID                                       Branch     Op    Remote     Download
 1. [✓] org.freedesktop.Platform.GL.default      20.08      i     flathub     95.3 MB / 95.9 MB
 2. [✓] org.freedesktop.Platform.VAAPI.Intel     20.08      i     flathub     11.4 MB / 11.5 MB
 3. [✓] org.freedesktop.Platform.ffmpeg-full     20.08      i     flathub      3.9 MB / 4.1 MB
        ID                                       Branch     Op    Remote     Download
 1. [✓] org.freedesktop.Platform.GL.default      20.08      i     flathub     95.3 MB / 95.9 MB
 2. [✓] org.freedesktop.Platform.VAAPI.Intel     20.08      i     flathub     11.4 MB / 11.5 MB
 3. [✓] org.freedesktop.Platform.ffmpeg-full     20.08      i     flathub      3.9 MB / 4.1 MB
 4. [✗] org.freedesktop.Platform.openh264        2.0        i     flathub      1.5 MB / 1.5 MB
 5. [✓] org.gnome.Fractal.Locale                 stable     i     flathub      4.8 kB / 233.8 kB
 6. [✓] org.gnome.Platform.Locale                3.38       i     flathub     17.7 kB / 326.1 MB
 7. [✓] org.gnome.Platform                       3.38       i     flathub    293.5 MB / 344.0 MB
 8. [✓] org.gnome.Fractal                        stable     i     flathub      3.5 MB / 3.6 MB

Warning: While trying to apply extra data: apply_extra script failed, exit status 256
Installation complete.
chronos@localhost /usr/local/lib/crew/packages (master *%|SPARSE)$ flatpak run org.gnome.Fractal

Note that the directories 

'/usr/local/var/lib/flatpak/exports/share'
'/home/chronos/user/.local/share/flatpak/exports/share'

are not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.

bwrap: Failed to mount tmpfs: Operation not permitted
error: ldconfig failed, exit status 256
```

saltedcoffii:
I wanna work on figuring out whether `priv-mode=setuid` is necessary in bubblewrap and flatpak. If it's not it'll make binary production and distribution much simpler. The problem I'm having is that chormeos's kernel doesn't have
`kernel.unprivileged_userns_clone=1`
I don't know if it can be enabled with `sysctl` or somewhere in `/proc` or `/sys`, or if it's just not compiled in and needs to be added with DKMS *(is that possible)* or would flatpak working require a custom kernel, basically flatpak unrealistic on chromeos.

satmandu:
Looking at the bubblewrap page, here's something we should try to get working:
```
bwrap --ro-bind /usr /usr --symlink usr/lib64 /lib64 --proc /proc --dev /dev --unshare-pid bash
```

That gives me this again:
```
bwrap: Failed to mount tmpfs: Operation not permitted
```

saltedcoffii:
Where is it trying to mount the tmpfs?

supechicken:
> I wanna work on figuring out whether `priv-mode=setuid` is necessary in bubblewrap and flatpak. If it's not it'll make binary production and distribution much simpler. The problem I'm having is that chormeos's kernel doesn't have
> `kernel.unprivileged_userns_clone=1`
> I don't know if it can be enabled with `sysctl` or somewhere in `/proc` or `/sys`, or if it's just not compiled in and needs to be added with DKMS _(is that possible)_ or would flatpak working require a custom kernel, basically flatpak unrealistic on chromeos.
> Or does chormeos's kernel support making new namespaces at all?

Do you mean:
```shell
sudo sysctl kernel.unprivileged_userns_clone=1
```

saltedcoffii:
> Do you mean:
> ```shell
> sudo sysctl kernel.unprivileged_userns_clone=1
> ```

Yes.

satmandu:
@saltedcoffii I added a wrapper script with the binaries for bubblewrap. See if that gets us closer?

packages/bubblewrap.rb
```diff
@@ -3,17 +3,34 @@
...
   homepage 'https://github.com/containers/bubblewrap'
-  version '0.4.1_1'
+  @_ver = '0.4.1'
+  version "#{@_ver}-1"
   compatibility 'all'
-  source_url 'https://github.com/containers/bubblewrap/releases/download/v0.4.1/bubblewrap-0.4.1.tar.xz'
+  source_url "https://github.com/containers/bubblewrap/releases/download/v#{@_ver}/bubblewrap-#{@_ver}.tar.xz"
   source_sha256 'b9c69b9b1c61a608f34325c8e1a495229bacf6e4a07cbb0c80cf7a814d7ccc03'
 
+  binary_url({
```
saltedcoffii: Somehow your binaries depend on dconf. Maybe try rebuilding them with dconf uninstalled?

`flatpak: error while loading shared libraries: libdconf.so.1: cannot open shared object file: No such file or directory`

saltedcoffii:
chromebrew doesn't support distributing binaries with some of the files within them setuid root

saltedcoffii:
And unfortunately the bubblewrap wrapper doesn't work. Flatpak won't build with it.

packages/bubblewrap.rb
satmandu: Is that bad? We could just add a dconf dependency..

packages/flatpak.rb
satmandu: Ugh. Looks like [setcap won't work](https://github.com/flatpak/flatpak/issues/3858#issuecomment-702016149). Alright... hold on trying something else.

packages/bubblewrap.rb
saltedcoffii: why not, we can add a dconf dependency.

saltedcoffii:
Okay so I did a bunch of research and poking around and here's the conclusion I came to. Anyone who knows more than me is welcome to correct me.
In order for bubblewrap to work, unprivileged users must have the ability to create namespaces. On most linux kernels (almost all distribution kernels), this feature is already enabled, or can be enabled until next boot by running `sudo sysctl kernel.unprivileged_userns_clone=1`. The default chromeos kernel (at least on my device) doesn't have a file at `/proc/sys/kernel/unprivileged_userns_clone`, which means, as far as I know, that the chromeos kernel doesn't support this and bubblewrap can't be supported.
This can theoretically be compiled in, however custom chromeos kernels are a pain (I don't know how to do them, I just know they can be done) and would be completely unavailable to most users. I know it requires either buying a $15 USB device or taking the bottom of your laptop off to remove the battery/a screw.
The only thing that I can think might work is a DKMS module. I don't know what part of the kernel enables `unprivileged_userns_clone` but I'm happy to do more digging. I also don't know if it's possible as a module, or if it has to be compiled into the kernel during compile time. So unless we can get that `sysctl` command working, bubblewrap, and consequently flatpak, won't work.

satmandu:
@saltedcoffii Apparently unprivileged_userns_clone isn't necessarily enabled in mainline kernels at all, and bubblewrap apparently works around it anyways?

See https://lists.debian.org/debian-kernel/2020/03/msg00237.html

satmandu:
(also flatpak builds with newest builds of bubblewrap... I'm just using the wrapper to chmod +s bwrap, and doing a chown root and chown chronos around the invocation of bwrap so it doesn't cause problems with adding or removing the package.

saltedcoffii:
if chromebrew had a `self.remove` function that was run right before the package files were uninstalled during `crew remove` it would make the setuid stuff a lot easier. and it would also allow for the deletion of flatpak's appdata.

satmandu:
Proof (no sound though)... and only wayland apps at the moment:
image


(This was using the xdg PR I just put in... I'll do a another rebuild of flatpak tomorrow morning. Currently building llvm 11.1.0... and that will take all night.)

supechicken:
Just built both with `--with-priv-mode=none` option, everything works well without `setuid` bit and scripts on my machine:
```shell
chronos@localhost /usr/local/bin $ ls -la flatpak
-rwxr-xr-x. 1 chronos chronos 1464144 Feb 18 18:09 flatpak
chronos@localhost /usr/local/bin $ ls -la bwrap 
-rwxr-xr-x. 1 chronos chronos 68480 Feb 18 17:47 bwrap
chronos@localhost /usr/local/lib/crew/packages $ file /usr/local/bin/flatpak
/usr/local/bin/flatpak: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, stripped
```
Screenshot 2021-02-18 at 6 16 16 PM

packages/bubblewrap.rb
```diff
@@ -3,30 +3,53 @@
...
   end
 
   def self.install
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    FileUtils.install "#{CREW_DEST_PREFIX}/bin/bwrap", "#{CREW_DEST_PREFIX}/bin/bwrap.elf", mode: 0o755
+    @bwrap_sh = <<~BWRAP_HEREDOC
+      #!/bin/bash
+      sudo chown root "#{CREW_PREFIX}/bin/bwrap.elf"
+      sudo chmod +s "#{CREW_PREFIX}/bin/bwrap.elf"
+      #{CREW_PREFIX}/bin/bwrap.elf "\$@"
```
supechicken, line 50: Why not use `sudo` directly? (the `suid` bit is used to run the file with permission of the file owner)

supechicken: Never mind, just found the `suid` bit will keep the home dir location while giving root permission to user

satmandu:
@supechicken That doesn't work for me.

```
flatpak run --socket=wayland org.videolan.VLC

Note that the directories 

'/usr/local/var/lib/flatpak/exports/share'
'/home/chronos/user/.local/share/flatpak/exports/share'

are not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.

bwrap: Failed to mount tmpfs: Operation not permitted
error: ldconfig failed, exit status 256
```

satmandu:
@supechicken works with the bubblewrap in this PR though (including binaries).

satmandu:
(As I understand it bubblewrap NEEDS to be setuid root... that is how it is designed to work and get around the user namespaces issues one would otherwise have.)

satmandu:
@supechicken Any special kernel you're running?

Also, I can't get DISPLAY variable into flatpak:

```
flatpak run org.mozilla.firefox

Note that the directories 

'/usr/local/var/lib/flatpak/exports/share'
'/home/chronos/user/.local/share/flatpak/exports/share'

are not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.


(firefox:3): Gtk-WARNING **: 14:52:55.291: cannot open display: :99.0
mozilla::detail::MutexImpl::~MutexImpl: pthread_mutex_destroy failed: Device or resource busy
ExceptionHandler::GenerateDump cloned child 32
ExceptionHandler::SendContinueSignalToChild sent continue signal to child
ExceptionHandler::WaitForContinueSignal waiting for continue signal...
chronos@localhost /usr/local/lib/crew/packages (master *%|SPARSE)$ DISPLAY=:0 flatpak run org.mozil
la.firefox

Note that the directories 

'/usr/local/var/lib/flatpak/exports/share'
'/home/chronos/user/.local/share/flatpak/exports/share'

are not in the search path set by the XDG_DATA_DIRS environment variable, so
applications installed by Flatpak may not appear on your desktop until the
session is restarted.


(firefox:3): Gtk-WARNING **: 14:53:08.848: cannot open display: :99.0
mozilla::detail::MutexImpl::~MutexImpl: pthread_mutex_destroy failed: Device or resource busy
ExceptionHandler::GenerateDump cloned child 32
ExceptionHandler::SendContinueSignalToChild sent continue signal to child
ExceptionHandler::WaitForContinueSignal waiting for continue signal...
```

What are you doing different? :)

satmandu:
@supechicken @saltedcoffii My x11 breakage was a problem on my end. I had ```GDK_BACKEND=wayland``` when I needed ```GDK_BACKEND=x11``` set.

Now X11 apps open fine.

Can we convert this to a Ready for Review PR?

image

satmandu:
Looks like some environment variables need to be unset before running flatpak. So I just setup a wrapper script.

This is my current setup:
```
require 'package'

class Flatpak < Package
  description 'Flatpak is a system for building, distributing, and running sandboxed desktop applications on Linux.'
  homepage 'https://flatpak.org'
  @_ver = '1.10.1'
  version @_ver
  compatibility 'all'
  source_url "https://github.com/flatpak/flatpak/releases/download/#{@_ver}/flatpak-#{@_ver}.tar.xz"
  source_sha256 'c70215792b7cbece83c489dab86adc9bfaf9b140c506affe2a48c92afa3d69b7'

  depends_on 'bubblewrap'
  depends_on 'libostree'
  depends_on 'polkit'
  depends_on 'appstream_glib'
  depends_on 'libseccomp'
  depends_on 'pyparsing'
  depends_on 'dconf'

  def self.build
    system 'env NOCONFIGURE=1 ./autogen.sh'
    system 'filefix'
    system "env BWRAP=#{CREW_PREFIX}/bin/bwrap CFLAGS='-flto=auto' CXXFLAGS='-flto=auto'  LDFLAGS='-flto=auto' \
      ./configure #{CREW_OPTIONS} \
      --with-system-install-dir=#{CREW_PREFIX}/var/lib/flatpak \
      --enable-sandboxed-triggers \
      --with-priv-mode=none \
      --enable-san \
      --without-systemd \
      --with-system-fonts-dir=#{CREW_PREFIX}/share/fonts:/usr/share/fonts \
      --with-system-font-cache-dirs=/usr/share/cache/fontconfig:#{CREW_PREFIX}/cache/fontconfig \
      --disable-documentation \
      --disable-maintainer-mode \
      --with-system-bubblewrap"
    system 'make'
  end

  def self.install
    FileUtils.install "#{CREW_DEST_PREFIX}/bin/flatpak", "#{CREW_DEST_PREFIX}/bin/flatpak.elf", mode: 0o755
    @flatpak_sh = <<~FLATPAK_HEREDOC
      #!/bin/bash
      unset GDK_PIXBUF_MODULE_FILE
      #{CREW_PREFIX}/bin/flatpak.elf "\$@"
    FLATPAK_HEREDOC
    IO.write("#{CREW_DEST_PREFIX}/bin/flatpak", @flatpak_sh, perm: 0o755)
  end

  def self.postinstall
    puts
    puts 'Configuring flathub'.lightblue
    system 'flatpak remote-add --if-not-exists --user flathub https://flathub.org/repo/flathub.flatpakrepo'
    puts
  end
end
```

saltedcoffii:
> Can we convert this to a Ready for Review PR?
Sure

satmandu:
I've uploaded binaries for all except armv7l on flatpak, since I have a slow build going on my rpi4 build box.

satmandu:
All binaries uploaded.

saltedcoffii:
Works wonderfully for me!

saltedcoffii:
chronos's home directory is on a LUKS encrypted blob, and the more data there is on it, the longer it takes to unlock. Perhaps we could symlink `~/.local/share/flatpak` to somewhere in `/usr/local` to reduce that unlock time, as flatpak's data files can take up gigabytes.

saltedcoffii:
and lastly, if we can figure out a way to get flatpak applications with wayland and x11 support to prefer wayland, that may allow them to run faster as sommelier then doesn't have to translate x directions to wayland directions.

satmandu:
> chronos's home directory is on a LUKS encrypted blob, and the more data there is on it, the longer it takes to unlock. Perhaps we could symlink `~/.local/share/flatpak` to somewhere in `/usr/local` to reduce that unlock time, as flatpak's data files can take up gigabytes.

@uberhacker was talking about this earlier.

Should we modify xdg_base.rb then to adjust this?  Suggestions?

supechicken (approved):

supechicken:
> @supechicken Any special kernel you're running?
> 
> Also, I can't get DISPLAY variable into flatpak:
> 
> ```
> flatpak run org.mozilla.firefox
> 
> Note that the directories 
> 
> '/usr/local/var/lib/flatpak/exports/share'
> '/home/chronos/user/.local/share/flatpak/exports/share'
> 
> are not in the search path set by the XDG_DATA_DIRS environment variable, so
> applications installed by Flatpak may not appear on your desktop until the
> session is restarted.
> 
> 
> (firefox:3): Gtk-WARNING **: 14:52:55.291: cannot open display: :99.0
> mozilla::detail::MutexImpl::~MutexImpl: pthread_mutex_destroy failed: Device or resource busy
> ExceptionHandler::GenerateDump cloned child 32
> ExceptionHandler::SendContinueSignalToChild sent continue signal to child
> ExceptionHandler::WaitForContinueSignal waiting for continue signal...
> chronos@localhost /usr/local/lib/crew/packages (master *%|SPARSE)$ DISPLAY=:0 flatpak run org.mozil
> la.firefox
> 
> Note that the directories 
> 
> '/usr/local/var/lib/flatpak/exports/share'
> '/home/chronos/user/.local/share/flatpak/exports/share'
> 
> are not in the search path set by the XDG_DATA_DIRS environment variable, so
> applications installed by Flatpak may not appear on your desktop until the
> session is restarted.
> 
> 
> (firefox:3): Gtk-WARNING **: 14:53:08.848: cannot open display: :99.0
> mozilla::detail::MutexImpl::~MutexImpl: pthread_mutex_destroy failed: Device or resource busy
> ExceptionHandler::GenerateDump cloned child 32
> ExceptionHandler::SendContinueSignalToChild sent continue signal to child
> ExceptionHandler::WaitForContinueSignal waiting for continue signal...
> ```
> 
> What are you doing different? :)

I am running Chrome OS with [Brunch framework kernel](https://github.com/sebanc/brunch) (also without unprivileged namespace support)

The only issue I met was the DISPLAY issue :) (solved by exporting the DISPLAY variable)

uberhacker:
@saltedcoffii: The `xdg_base` package symlinks `~/.local` to `/usr/local/.config` which should get the job done.

packages/flatpak.rb
```diff
@@ -0,0 +1,90 @@
...
+     x86_64: 'ac8e6a92a18862201c12563036f206804547d2a1d350f548d130bb38b9dfea5b'
+  })
+
+  depends_on 'bubblewrap'
+  depends_on 'libostree'
+  depends_on 'polkit'
+  depends_on 'appstream_glib'
+  depends_on 'libseccomp'
+  depends_on 'pyparsing'
+  depends_on 'dconf'
```
uberhacker, line 31: `depends_on 'xdg_base'`?

packages/flatpak.rb
```diff
@@ -0,0 +1,90 @@
...
+
+  def self.patch
+    patch_description = 'backported fixes'
+    patch_url = 'https://patch-diff.githubusercontent.com/raw/flatpak/flatpak/pull/4132.patch'
+    patch_sha256 = '8230491be909a027fdb5a03ff0fa638331c940289fa6d68a21489d78c6f7f6b0'
+    patch_filename = 'patch'
```
uberhacker, line 37: My general rule is you don't need a variable assigned unless you plan on using it more than once.  I could probably make the argument you don't need the patch_filename variable either.

packages/flatpak.rb
```diff
@@ -0,0 +1,90 @@
...
+    puts "Downloading patch: #{patch_description}".yellow
+    system('curl', '-s', '--insecure', '-L', '-#', patch_url, '-o', patch_filename)
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless
+    Digest::SHA256.hexdigest(File.read(patch_filename)) == patch_sha256
+    puts 'patch downloaded'.lightgreen
```
uberhacker, line 42: Not having variables makes the logic easier to understand as well.

packages/flatpak.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def self.build
+    system 'env NOCONFIGURE=1 ./autogen.sh'
+    system 'filefix'
```
uberhacker, line 47: Technically, this qualifies for the `def self.patch` section.

packages/flatpak.rb
```diff
@@ -0,0 +1,90 @@
...
+    system "env BWRAP=#{CREW_PREFIX}/bin/bwrap CFLAGS='-flto=auto' CXXFLAGS='-flto=auto'  LDFLAGS='-flto=auto' \
+      ./configure #{CREW_OPTIONS} \
+      --with-system-install-dir=#{CREW_PREFIX}/var/lib/flatpak \
+      --enable-sandboxed-triggers \
+      --with-priv-mode=none \
+      --enable-san \
+      --without-systemd \
+      --with-system-fonts-dir=#{CREW_PREFIX}/share/fonts:/usr/share/fonts \
+      --with-system-font-cache-dirs=/usr/share/cache/fontconfig:#{CREW_PREFIX}/cache/fontconfig \
```
uberhacker, line 56: Does it make sense to have a read-only cache dir?

packages/flatpak.rb
saltedcoffii: Sensible

packages/flatpak.rb
satmandu: The read-only cache is I think generated by google for each build for built-in fonts.

Honestly my issue is actually with the  ``` --with-system-fonts-dir=#{CREW_PREFIX}/share/fonts:/usr/share/fonts \``` line, which I don't think works. I had to make a symlink in the flatpak wrapper script to generate a symlink to ~/.local/share/fonts to get fonts in ```#{CREW_PREFIX}/share/fonts``` to be seen.

packages/flatpak.rb
satmandu: I think the issue here is that you need the configure script generated before ```filefix``` will work, and configure is in the ```def self.build``` section, no?

Unless one is talking about moving configure steps to ```def self.prebuild```...

packages/flatpak.rb
satmandu: So I'm trying to use a common code block for patching between packages.

Ideally there would be a way to add multiple patches without too much code reuse.

Case in point: We should use MOST of these patches for libsdl: https://github.com/archlinux/svntogit-packages/tree/packages/sdl/trunk

@supechicken Any chance you want to play with adding a patch framework to crew so we don't have to reinvent this for every single package which needs patches?

packages/flatpak.rb
satmandu: Yup. As per the previous comment, would love to just take all of the patch logic out of these files entirely.

packages/flatpak.rb
satmandu: Examples of how patches are put into package files elsewhere it would be nice to repicate:

Arch: https://github.com/archlinux/svntogit-packages/blob/packages/sdl/trunk/PKGBUILD
Homebrew: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/sdl.rb

(I think the arch version is more elegant, but the homebrew method might be easier to copy since their code is in Ruby.)

packages/flatpak.rb
saltedcoffii: It's fine. According to the other package.rb files, `def.patch` is almost exclusively for `sed` and `patch`.

packages/flatpak.rb
supechicken: > So I'm trying to use a common code block for patching between packages.
> 
> Ideally there would be a way to add multiple patches without too much code reuse.
> 
> Case in point: We should use MOST of these patches for libsdl: https://github.com/archlinux/svntogit-packages/tree/packages/sdl/trunk
> 
> @supechicken Any chance you want to play with adding a patch framework to crew so we don't have to reinvent this for every single package which needs patches?

I will take a try at it :)

packages/flatpak.rb
```diff
@@ -0,0 +1,92 @@
+require 'package'
+
+class Flatpak < Package
+  description 'Flatpak is a system for building, distributing, and running sandboxed desktop applications on Linux.'
+  homepage 'https://flatpak.org'
+  @_ver = '1.10.1'
+  version @_ver
+  compatibility 'all'
+  source_url "https://github.com/flatpak/flatpak/releases/download/#{@_ver}/flatpak-#{@_ver}.tar.xz"
+  source_sha256 'c70215792b7cbece83c489dab86adc9bfaf9b140c506affe2a48c92afa3d69b7'
```
saltedcoffii, line 10: Should one scroll back in flatpak's commits they'll notice that the source_sha256 field has switched back and forth between c70215792b7cbece83c489dab86adc9bfaf9b140c506affe2a48c92afa3d69b7 and
c1354f42bf3b5d51aeb4028c9b62fd4ffc673ef2ff6e583c17777f5dafdbdcb7. I'm not sure what's going on here but it appears github is giving two different files with very minor differences when requesting for the tarball? I'm not entirely sure but hopefully we can accommodate it.

satmandu: Not sure why that's happening either.

satmandu:
FYI pulseaudio works inside flatpak if pulseaudio is configured appropriately.

(I don't have sound working inside my own puleaudio configuration yet... I'm still trying to get pulseaudio to dump audio to alsa.)

saltedcoffii:
> FYI pulseaudio works inside flatpak if pulseaudio is configured appropriately.
> 
> (I don't have sound working inside my own puleaudio configuration yet... I'm still trying to get pulseaudio to dump audio to alsa.)

have you tried apulse and skip pulseaudio entirely?

satmandu:
@saltedcoffii I have not... but that sounds like its own shitshow of compatibility, since programs which expect pulseaudio might crash...

packages/flatpak.rb
satmandu: This hasn't happened in a bit...

satmandu:
@saltedcoffii I have sound working now with pulseaudio.

Please also make sure that you have the updates in https://github.com/skycocker/chromebrew/pull/5203 installed, and it might work for you. (you may have to delete your ```~/.config/pulse/``` config files if you have any.

packages/flatpak.rb
```diff
@@ -96,5 +96,9 @@ def self.postinstall
     puts 'Configuring flathub'.lightblue
     system 'flatpak.elf remote-add --if-not-exists --user flathub https://flathub.org/repo/flathub.flatpakrepo'
     puts
+    xdgconfig_in_bashrc = `grep -c "#{CREW_PREFIX}/var/lib/flatpak/exports/share" ~/.bashrc || true`
+    unless xdgconfig_in_bashrc.to_i.positive?
+      system "sed -i 's,XDG_DATA_DIRS=.*,XDG_DATA_DIRS=#{CREW_PREFIX}/share:#{HOME}/.local/share/flatpak/exports/share:#{CREW_PREFIX}/var/lib/flatpak/exports/share,g' ~/.bashrc"
```
uberhacker, line 101: Should we check if `XDG_DATA_DIRS=...` exists in `~/.bashrc` also?  It's possible that `xdg_base` is installed but the environment variables haven't been added.

satmandu: Good call.

satmandu: Made changes as requested.

satmandu (approved):

uberhacker:
I think I'm going to regret this decision but here goes anyway. :)

satmandu:
@uberhacker @supechicken How else are we going to get this working?

https://github.com/fastrizwaan/flatpak-wine32

😁

uberhacker:
The solution is simple but hard.  Persistence and determination.

UCYT5040:
Was there any solution for the bwrap error?

satmandu:
> Was there any solution for the bwrap error?

Not that I know of...

PR #5147, opened 2021-02-17
crew: Fix dependencies order

supechicken:
The bugs with the original dependencies system:
- Parent dependency will install first before sub-dependencies

Fixed by:
- Putting parent dependency to last (if have any sub-dependency)

Tested on `x86_64`


crew
```diff
@@ -725,6 +725,7 @@ def expand_dependencies
         @dependencies << dep
         search dep, true
         push_dependencies
+        @dependencies = @dependencies - [dep] + [dep]
```
uberhacker, line 728: I'm not following.  Is the intention to rearrange the dep order?

crew
```diff
@@ -739,26 +740,20 @@ def resolve_dependencies
...
-  end
-  @dependencies.concat last_deps.sort
-
-  @dependencies.each do |dep|
-    print dep + ' '
   end
 
-  puts
+  puts "\n\n"
   print 'Do you agree? [Y/n] '
```
uberhacker, line 757: You could get rid of both of the puts lines with this:
```
print "\n\nDo you agree? [Y/n] "
```

uberhacker:
We'll need to bump the crew version in `lib/const.rb` also.

crew
uberhacker: Could this be done in `push_dependencies` instead?

crew
supechicken: Done

crew
supechicken: This will move the dependencies to last after a loop of sub-dependency checking (`search dep, true` and `push_dependencies`)

supechicken: (I also tried to move `@dependencies << dep` to after `push_dependencies` but it causes an infinite loop)

supechicken:
> We'll need to bump the crew version in `lib/const.rb` also.

Done

crew
```diff
@@ -739,27 +740,20 @@ def resolve_dependencies
...
   return if @dependencies.empty?
 
+  puts "\nThe following packages also need to be installed:\n\n"
 
-  puts 'The following packages also need to be installed: '
-
-  i = 0
-  last_deps = []
-  last_packages = ["curl", "ghc", "mandb", "gtk3", "sommelier"]
+  last_packages = ['curl', 'ghc', 'mandb', 'gtk3', 'sommelier']
```
uberhacker, line 745: We can remove `curl` and `mandb` since they are already core packages.

supechicken: Done

crew
```diff
@@ -739,27 +740,20 @@ def resolve_dependencies
...
-  last_deps = []
-  last_packages = ["curl", "ghc", "mandb", "gtk3", "sommelier"]
+  last_packages = ['ghc', 'gtk3', 'sommelier']
+  
   @dependencies.each do |dep|
-    if last_packages.include?(dep)
-      @dependencies.delete_at(i)
-      last_deps.push(dep)
+    if last_packages.include?(dep) then
+      @dependencies = (@dependencies - [dep] + [dep])
```
uberhacker, line 749: I think it's also important to install gtk3 before sommelier.  Does this logic ensure that will happen?

supechicken: No, but `sommelier` is the last dep in a package at most time, so it may not a big problem :)

crew
uberhacker: Can we put this logic in `push_dependencies`?

PR #5148, opened 2021-02-17
[Update Package] elogind 243.4 => 246.9.2

saltedcoffii:
This may require a rebuild of `linux_pam`, at least on x86_64.

### DIFF
--- a/packages/elogind.rb
+++ b/packages/elogind.rb
@@ -3,10 +3,10 @@ require 'package'
 class Elogind < Package
   description 'Standalone systemd-logind fork'
   homepage 'https://github.com/elogind/elogind'
-  version '243.4'
+  version '246.9.2'
   compatibility 'all'
-  source_url 'https://github.com/elogind/elogind/archive/v243.4.tar.gz'
-  source_sha256 'f1098745863138e6270ea22e78a39baef9a0356b48246c5a53b34211992dc7db'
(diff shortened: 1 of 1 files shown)


uberhacker:
Missing binaries?

saltedcoffii:
Wym?

satmandu:
FYI, building binaries fails w/ kernels older than 4.7.

See https://github.com/elogind/elogind/issues/199

satmandu:
Also, why would a linux_pam rebuild be required?

saltedcoffii:
> Also, why would a linux_pam rebuild be required?
The binary was "broken" I don't know. All i know is that elogind failed to build with the binary linux_pam and worked when I compiled it from the source.

saltedcoffii:
> FYI, building binaries fails w/ kernels older than 4.7.
> 
> See [elogind/elogind#199](https://github.com/elogind/elogind/issues/199)

Adding a line with `uname -r` rejecting anything less than 4.7 in self.prebuild or something might work.
```ruby
abort if `uname -r` =< 4.7 end
```
or whatever that looks like, that's likely incorrect code.

satmandu:
Not building on older kernels isn't an option... most ChromeOS devices would fail this check!

As per the above link to the issue I've raised upstream, maybe there could be a workaround to fix this.

saltedcoffii:
maybe something as simple as `--without-btrfs` would fix it, as chromeos doesn't require btrfs support anyway.

PR #5155, opened 2021-02-18
[New Package] liburing

saltedcoffii:


packages/liburing.rb
```diff
@@ -0,0 +1,19 @@
...
+class Liburing < Package
+  description 'liburing provides helpers to setup and teardown io_uring instances.'
+  homepage 'https://github.com/axboe/liburing/'
+  version '0.7'
+  compatibility 'all'
+  source_url 'https://github.com/axboe/liburing/archive/liburing-0.7.tar.gz'
+  source_sha256 '8e2842cfe947f3a443af301bdd6d034455536c38a455c7a700d0c1ad165a7543'
+
+  def self.build
+    system "./configure --prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --mandir=#{CREW_MAN_PREFIX}"
```
uberhacker, line 12: `system "./configure #{CREW_OPTIONS}"`?

saltedcoffii: This particular configure script is really limited in what it does. Trying to pass CREW_OPTIONS causes it to fail.

uberhacker: Fair enough.

PR #5156, opened 2021-02-18
[New Package] libfmt

saltedcoffii:


packages/libfmt.rb
```diff
@@ -0,0 +1,19 @@
...
+class Libfmt < Package
+  description 'A modern formatting library'
+  homepage 'https://fmt.dev'
+  version '7.1.3'
+  compatibility 'all'
+  source_url 'https://github.com/fmtlib/fmt/releases/download/7.1.3/fmt-7.1.3.zip'
+  source_sha256 '5d98c504d0205f912e22449ecdea776b78ce0bb096927334f80781e720084c9f'
+
+  def self.build
+    system "cmake #{CREW_CMAKE_OPTIONS} ."
```
uberhacker, line 12: Typically, when building using cmake, a `build` directory is needed.  Did you follow the build instructions?

saltedcoffii: cmake didn't respond with any errors or warnings when I built this

uberhacker: That still doesn't answer my question. :)

saltedcoffii: Yes, and they're (here)[https://fmt.dev/latest/usage.html#building-the-library]. Although the instructions say to make a `build` directory, they also say right above "The typical workflow starts with:", implying that it isn't necessary. You're welcome to add a build dir to this package script should you like tho :)

uberhacker: The idea of the build directory is to isolate the build from the source.  That way the package doesn't end up containing extra files that are unnecessary.  I think this is a very important concept to understand.

saltedcoffii: I didn't know that, thank you! A fix is on its way.

satmandu:
Why not use Ninja with Cmake?
See https://github.com/satmandu/chromebrew/blob/master/packages/webkit2gtk.rb for an example. 

It makes for a simpler package, I think.

saltedcoffii:
Apparently ninja has a negligible speed increase (between one and five seconds)
https://hamelot.io/programming/make-vs-ninja-performance-comparison/

Otherwise I don't know how using cmake's ninja generator makes the package simpler; I'm curious to know.

PR #5159, opened 2021-02-18
[New Package] pakitheme

saltedcoffii:


packages/pakitheme.rb
```diff
@@ -0,0 +1,17 @@
...
+  version '1.0'
+  compatibility 'all'
+  source_url 'https://github.com/saltedcoffii/pakitheme/archive/1.0.tar.gz'
+  source_sha256 '32446ecc73ae95d699d6e14e5fa3dc3a8b3a9cfae02457b2b7f38fdd7d7524f3'
+
+  depends_on 'flatpak'
+
+  def self.install
+    system "mkdir #{CREW_DEST_PREFIX}/bin -pv"
+    system "install -v ./pakitheme #{CREW_DEST_PREFIX}/bin/pakitheme"
```
uberhacker, line 15: You can get rid of the mkdir line with:
```
system "install -Dm755 ./pakitheme #{CREW_DEST_PREFIX}/bin/pakitheme"
```
or...better yet.
```
FileUtils.install 'pakitheme', "#{CREW_DEST_PREFIX}/bin/pakitheme", mode: 0755
```

saltedcoffii: ooh i like the second one that's pretty. didn't know ruby did that.

saltedcoffii: looks like FileUtils.install doesn't create directories 🤔 
i'll use the first.

packages/pakitheme.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class Pakitheme < Package
+  description 'Automatically install your host GTK+ theme as a Flatpak'
+  homepage 'https://github.com/refi64/pakitheme'
+  version '1.0'
+  compatibility 'all'
+  source_url 'https://github.com/saltedcoffii/pakitheme/archive/1.0.tar.gz'
+  source_sha256 '32446ecc73ae95d699d6e14e5fa3dc3a8b3a9cfae02457b2b7f38fdd7d7524f3'
+  
+  depends_on 'flatpak'
```
uberhacker, line 11: I'll wait until flatpak is finalized before merging this.

satmandu:
Should not the source url be something off of the home page of the package?

saltedcoffii:
There weren't any releases, so I had to fork the package, create a release and use that.

PR #5162, opened 2021-02-19
[New Package] btrfsprogs

saltedcoffii:


packages/btrfsprogs.rb
```diff
@@ -0,0 +1,29 @@
+require 'package'
+
+class Btrfsprogs < Package
+  description 'BTRFS is a modern copy on write filesystem for Linux aimed at implementing advanced features while also focusing on fault tolerance, repair and easy administration.'
+  homepage 'https://github.com/kdave/btrfs-progs'
```
uberhacker, line 5: Better homepage? https://btrfs.wiki.kernel.org/index.php/Main_Page

packages/btrfsprogs.rb
```diff
@@ -0,0 +1,29 @@
...
+  version '5.10.1'
+  compatibility 'all'
+  source_url 'https://github.com/kdave/btrfs-progs/archive/v5.10.1.tar.gz'
+  source_sha256 '69788461f7076951f7235b87d0a5615683151dfbfaaa93f645279bf757e85769'
+
+  depends_on 'util_linux'
+  depends_on 'libgcrypt'
+  depends_on 'lzo'
+# depends_on 'e2fsprogs'
```
uberhacker, line 14: Can we remove the unused dependency?

saltedcoffii: I suppose we could, however I was hoping someone might be able to get `--with-convert=ext2` to which `e2fsprogs` is dependent on.

PR #5165, opened 2021-02-19
[Fix Package] fontconfig

### LINKED ISSUE #5158: fontconfig seems to require graphite as a dependency
```
fontconfig: Fontconfig is a library for configuring and customizing font access.
############################################################################ 100.0%
Archive downloaded
Performing pre-install...
Performing install...
Performing post-install...
Putting fontconfig code in ~/.bashrc
To complete the installation, execute the following:
source ~/.bashrc
fc-cache: error while loading shared libraries: libgraphite2.so.3: cannot open shared object file: No such file or directory
```

I'm not sure what you guys want to do about this, whether add `graphite` as a dependency or change fontconfig's postinstall or something else about the binary.

saltedcoffii:
Fixes #5158

### DIFF
--- a/packages/fontconfig.rb
+++ b/packages/fontconfig.rb
@@ -4,10 +4,23 @@ class Fontconfig < Package
   description 'Fontconfig is a library for configuring and customizing font access.'
   homepage 'https://www.freedesktop.org/software/fontconfig/front.html'
   @_ver = '2.13.93'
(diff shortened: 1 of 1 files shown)


uberhacker:
We'll need to remove the binaries and bump the version.

uberhacker:
Dude, you are too fast. :)

PR #5169, opened 2021-02-19
[New Package] gnuchess

saltedcoffii:


uberhacker:
If we start with games, this is going to open Pandora's box. :)

PR #5170, opened 2021-02-19
llvm -> 11.1.0 Final

satmandu:
- rebuild for 11.1.0 final
Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/llvm.rb
+++ b/packages/llvm.rb
@@ -3,47 +3,41 @@ require 'package'
 class Llvm < Package
   description 'The LLVM Project is a collection of modular and reusable compiler and toolchain technologies. The optional packages clang, lld, lldb, polly, compiler-rt, libcxx, libcxxabi, and openmp are included.'
   homepage 'http://llvm.org/'
-  @_ver = '11.1.0-rc1'
+  @_ver = '11.1.0'
   version @_ver
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


packages/llvm.rb
```diff
@@ -3,78 +3,74 @@
...
   depends_on 'ocaml' => :build
-  depends_on 'libedit'
-  depends_on 'libtirpc'
-  depends_on 'swig'
   depends_on 'pygments' => :build
   depends_on 'ccache' => :build
-  depends_on 'llvm_stage1' => :build
-  depends_on 'binutils' => :build
+  # depends_on 'llvm_stage1' => :build # Not needed with a functional
+  # prior llvm 11 build.
```
uberhacker, line 29: Are you sure we need to remove all these dependencies?

satmandu: The llvm_stage1 is commented on purpose.

Binutils is a core package, no?

uberhacker: Right but you need to be careful about removing dependencies from core packages because those dependencies actually make up the entire list of core packages. :)  I remember this package was added after binutils so yes, this line can be removed.  Just saying though.  I'm guessing you answered my question with a yes.

satmandu: hmmm... we need to modify a core packages cleaning script which doesn't touch core package files... or actually handle dependencies properly... :)

PR #5174, opened 2021-02-20
[Update] vte 0.62.3 => 0.63.91

saltedcoffii:


packages/vte.rb
```diff
@@ -3,57 +3,44 @@
...
-  })
-  binary_sha256 ({
-     aarch64: '948e60479e834c4cf9c3fdebab5795faf64c38122d4662f4c355b1068078f743',
-      armv7l: '948e60479e834c4cf9c3fdebab5795faf64c38122d4662f4c355b1068078f743',
-        i686: '88d7d0d1a8b3e8b86a8ff3c66aa6fc7b1ae8d57712d1659c674bc3b129fd2fcd',
-      x86_64: 'f261f381d4ff58761907737037ed60ddafff5444067b38fb28e45f02f04ed6aa',
-  })
-
-  depends_on 'gobject_introspection' => ':build'
+  depends_on 'gobject_introspection' => :build
```
uberhacker, line 11: Good catch.

packages/vte.rb
```diff
@@ -3,57 +3,44 @@
...
+#  depends_on 'vala' => :build
```
uberhacker, line 12: Please remove this line if vala is not a dependency.

packages/vte.rb
```diff
@@ -3,57 +3,44 @@
...
+            -Dgtk4=true \
+            -Dgir=false \
+            -Dvapi=false \
+            builddir" # TODO: get gir and vapi working
+    system "ninja -C builddir"
   end
   
   def self.install
-  system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
+    system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
```
uberhacker, line 44: Thank you for indenting.  That helps with readability.

packages/vte.rb
saltedcoffii: It's a required dependency for the TODO

PR #5176, opened 2021-02-20
gtk3 => 3.24.25, use lto

supechicken:
Builds properly on `x86_64`

### DIFF
--- a/packages/gtk3.rb
+++ b/packages/gtk3.rb
@@ -54,13 +54,13 @@ class Gtk3 < Package
     puts 'Performing configuration...'
     FileUtils.mkdir_p("#{HOME}/.config/gtk-3.0")
     @ini = "#{HOME}/.config/gtk-3.0/settings.ini"
-    @file = <<EOF
-[Settings]
-gtk-application-prefer-dark-theme = false
-gtk-icon-theme-name = hicolor
-gtk-fallback-icon-theme = gnome
-gtk-font-name = Arial 10
-EOF 
+    @file = <<~EOF
+      [Settings]
+      gtk-application-prefer-dark-theme = false
+      gtk-icon-theme-name = hicolor
+      gtk-fallback-icon-theme = gnome
+      gtk-font-name = Arial 10
+    EOF
     File.write(@ini, @file) unless File.exist?(@ini)
   end
 end
--- a/packages/gtk3.rb
+++ b/packages/gtk3.rb
@@ -39,12 +39,6 @@ class Gtk3 < Package
             builddir"
     system 'meson configure builddir'
     system 'ninja -C builddir'
-  end
-
-  def self.install
-    system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
-    system "sed -i 's,null,,g'  #{CREW_DEST_LIB_PREFIX}/pkgconfig/gtk*.pc"
-    FileUtils.mkdir_p("#{CREW_DEST_HOME}/.config/gtk-3.0")
     @file = <<~EOF
       [Settings]
       gtk-application-prefer-dark-theme = false
@@ -52,6 +46,12 @@ class Gtk3 < Package
       gtk-fallback-icon-theme = gnome
       gtk-font-name = Arial 10
     EOF
+  end
+
+  def self.install
+    system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
+    system "sed -i 's,null,,g'  #{CREW_DEST_LIB_PREFIX}/pkgconfig/gtk*.pc"
+    FileUtils.mkdir_p("#{CREW_DEST_HOME}/.config/gtk-3.0")
     File.write("#{CREW_DEST_HOME}/.config/gtk-3.0/settings.ini", @file)
   end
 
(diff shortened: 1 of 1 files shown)


packages/gtk3.rb
```diff
@@ -3,24 +3,13 @@
...
-  binary_sha256 ({
-     aarch64: '41067ff7cb6dd415c1ffaee6d9016388323a51d7bc7183fe0ef99749bf95dbea',
-      armv7l: '41067ff7cb6dd415c1ffaee6d9016388323a51d7bc7183fe0ef99749bf95dbea',
-        i686: 'e338cd7188bdd23b6675db7013098b8a13cd7fe52e9d3903560037588156f8e8',
-      x86_64: 'c9a7c3e02a3b1f283076e4dfba0e470161c3c701d0649fe5873a27f6a87b365e',
-  })
-
+  source_url "https://download.gnome.org/sources/gtk+/3.24/gtk+-#{@_ver}.tar.xz"
+  source_sha256 `curl -Ls https://download.gnome.org/sources/gtk+/3.24/gtk+-#{@_ver}.sha256sum |\
+                 tail -n1 | cut -d ' ' -f1`.tr("\n", '') if (ARGV[0] == 'install' or ARGV[0] == 'reinstall')
```
uberhacker, line 11: Would `.chomp` work better than `.tr("\n", '')`?  I'm thinking we should be checking for `-s`, `-S`, `--build-from-source` or `--recursive-build` switch options instead because technically `upgrade` could be a crew argument also.

uberhacker: Actually, I don't think we need to check for any arguments.  Just get rid of `if (ARGV[0] == 'install' or ARGV[0] == 'reinstall')` altogether and it should still work fine.

packages/gtk3.rb
```diff
@@ -59,16 +51,17 @@ def self.postinstall
...
-[Settings]
-gtk-application-prefer-dark-theme = false
-gtk-icon-theme-name = hicolor
-gtk-fallback-icon-theme = gnome
-gtk-font-name = Arial 10
-EOF".lightblue
-    puts
+    
+    puts 'Performing configuration...'
+    FileUtils.mkdir_p("#{HOME}/.config/gtk-3.0")
```
uberhacker, line 56: `#{CREW_DEST_HOME}` instead of `#{HOME}`?

uberhacker: Actually, it would probably make sense to move this to the `def self.build` section instead.

packages/gtk3.rb
```diff
@@ -59,16 +51,17 @@ def self.postinstall
...
+    @ini = "#{HOME}/.config/gtk-3.0/settings.ini"
```
uberhacker, line 57: `#{CREW_DEST_HOME}` instead of `#{HOME}`?

packages/gtk3.rb
```diff
@@ -59,16 +51,17 @@ def self.postinstall
...
+    @file = <<~EOF
+      [Settings]
+      gtk-application-prefer-dark-theme = false
+      gtk-icon-theme-name = hicolor
+      gtk-fallback-icon-theme = gnome
+      gtk-font-name = Arial 10
+    EOF
```
uberhacker, line 64: Move to `def self.build` section?

packages/gtk3.rb
```diff
@@ -59,16 +51,17 @@ def self.postinstall
...
+    File.write(@ini, @file) unless File.exist?(@ini)
```
uberhacker, line 65: Move to `def self.install` section?

uberhacker: Actually, all you need is `File.write(@ini, @file)`.

uberhacker: When all is said and done, we shouldn't need a `def self.postinstall` section.

uberhacker: Maybe it makes sense to prompt to remove `#{HOME}/.config/gtk-3.0` in `def self.remove`?

supechicken: Moved to `install` section

packages/gtk3.rb
supechicken: Done

packages/gtk3.rb
```diff
@@ -3,24 +3,13 @@
...
-      x86_64: 'c9a7c3e02a3b1f283076e4dfba0e470161c3c701d0649fe5873a27f6a87b365e',
-  })
-
+  source_url "https://download.gnome.org/sources/gtk+/3.24/gtk+-#{@_ver}.tar.xz"
+  source_sha256 `curl -Ls https://download.gnome.org/sources/gtk+/3.24/gtk+-#{@_ver}.sha256sum |\
+                 tail -n1 | cut -d ' ' -f1`.chomp
+  
   depends_on 'cups'
   depends_on 'at_spi2_atk'
   depends_on 'gcc10' => :build
```
uberhacker, line 15: Core package?

packages/gtk3.rb
```diff
@@ -41,34 +30,35 @@ class Gtk3 < Package
   depends_on 'graphite'
```
uberhacker, line 30: `depends_on 'llvm' => :build` is also a core package.

supechicken666: Removed

satmandu:
Wait, the source hash is what now? Generated on the fly? What's the point of that?

PR #5180, opened 2021-02-20
[New Package] miscfiles

saltedcoffii:


packages/miscfiles.rb
```diff
@@ -0,0 +1,18 @@
+require 'package'
+
+class Miscfiles < Package
+  description 'A collection of files not of crucial importance for system administration or operation, but which have come to be common on various systems over the years..'
```
uberhacker, line 4: Extra period (.) at the end?

uberhacker:
@saltedcoffii:  You are on a roll! :1st_place_medal:

packages/miscfiles.rb
saltedcoffii: Oops

PR #5182, opened 2021-02-20
[New Package] pyconfigure

saltedcoffii:


packages/pyconfigure.rb
```diff
@@ -0,0 +1,21 @@
+require 'package'
+
+class Pyconfigure < Package
+  description 'GNU pyconfigure provides developers with file templates for implementing standard configure scripts and Makefile recipes for their Python packages.'
+  homepage 'https://www.gnu.org/software/pyconfigure/'
+  version '3.1.1'
```
uberhacker, line 6: Doesn't match source url version.

packages/pyconfigure.rb
```diff
@@ -0,0 +1,21 @@
...
+  compatibility 'all'
+  source_url 'https://ftpmirror.gnu.org/pyconfigure/pyconfigure-0.2.3.tar.gz'
+  source_sha256 '37438ee1ae0aad7692188c7d01218cb407a1c89f3efdb5d360311b7ade4ab14f'
+  
+  depends_on 'python2_7'
```
uberhacker, line 11: `depends_on 'python27'`?

uberhacker: Actually, you can just exclude since it's a core package.

packages/pyconfigure.rb
saltedcoffii: done

packages/pyconfigure.rb
saltedcoffii: done

PR #5183, opened 2021-02-20
libcanberra: add wayland fix

satmandu:
- libcanberra dies wwhen an app using it is open in wayland. this patch fixes that error. (seen in pavucontrol opened in Wayland)
- mentioned here: http://issues.guix.gnu.org/43918

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/libcanberra.rb
+++ b/packages/libcanberra.rb
@@ -3,27 +3,26 @@ require 'package'
 class Libcanberra < Package
   description 'XDG Sound Theme and Name Specification library implementation'
   homepage 'http://0pointer.de/lennart/projects/libcanberra/'
-  version '0.30'
+  version '0.30-1'
   compatibility 'all'
   source_url 'http://pkgs.fedoraproject.org/repo/pkgs/libcanberra/libcanberra-0.30.tar.xz/34cb7e4430afaf6f447c4ebdb9b42072/libcanberra-0.30.tar.xz'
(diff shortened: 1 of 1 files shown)


packages/libcanberra.rb
```diff
@@ -3,36 +3,105 @@
...
+            +        if (!GDK_IS_X11_DISPLAY(d))
+            +                return 0;
+            +#endif
+            +
+                     if (XGetWindowProperty(GDK_DISPLAY_XDISPLAY(d), GDK_WINDOW_XID(w),
+                                            gdk_x11_get_xatom_by_name_for_display(d, "_NET_WM_DESKTOP"),
+                                            0, G_MAXLONG, False, XA_CARDINAL, &type_return,
+    LIBCANBERRA_WAYLAND_PATCH_HEREDOC
+    IO.write('libcanberra-wayland-crash.patch', @libcanberra_wayland_crash_patch, perm: 0o555)
+    system 'patch -p0 < libcanberra-wayland-crash.patch'
```
uberhacker, line 93: Epic patch!

satmandu: Yeah, but it's a package which hasn't been updated in 8 years... And the patch was mentioned on a mailing list, so I figured that we should just add it to the package instead of pulling it from somewhere where the source might get lost.

PR #5185, opened 2021-02-20
opera => 74.0.3911.154

supechicken:
Tested on `x86_64`

### DIFF
--- a/packages/opera.rb
+++ b/packages/opera.rb
@@ -14,6 +14,7 @@ class Opera < Package
     depends_on 'gsettings_desktop_schemas'
     depends_on 'harfbuzz'
     depends_on 'graphite'
+    depends_on 'cras'
     depends_on 'sommelier'
   end
 
--- a/packages/opera.rb
+++ b/packages/opera.rb
@@ -8,7 +8,7 @@ class Opera < Package
   case ARCH
   when 'x86_64'
     source_url "https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb"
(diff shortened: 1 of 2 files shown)


satmandu:
Does this work? An upgrade of opera failed for me this evening.

uberhacker:
We're ready for another update anyway.

packages/opera.rb
```diff
@@ -3,34 +3,35 @@
...
 
   def self.build
-    system "curl https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb -#o opera.deb"
-    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("opera.deb") ) == `curl -Ls https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb.sha256sum`
-    puts "Archive downloaded".lightgreen
-    system "alien -tc opera.deb"
-    system "tar xvf opera-stable-#{version}.tgz > /dev/null"
+    Dir.chdir('../') do
+      system "alien -tc opera-stable_#{version}_amd64.deb"
+      system "tar xf opera-stable-#{version}.tgz"
```
uberhacker, line 24: We don't need to do this anymore, right?

supechicken:
> Does this work? An upgrade of opera failed for me this evening.

It works before but seems #5247 broke the unarchiving process
We may need to update all deb packages for it

packages/opera.rb
supechicken: A pr for all deb packages is coming soon:)

PR #5194, opened 2021-02-20
[New Package] firefox_esr

saltedcoffii:


packages/firefox_esr.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def self.build
+    system "echo '#!/bin/bash' > firefox.sh"
+    system "echo 'cd #{CREW_PREFIX}/firefox' >> firefox.sh"
+    system "echo 'apulse ./firefox \"\$@\"' >> firefox.sh"
+  end
+
+  def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/firefox"
+    FileUtils.cp_r '.', "#{CREW_DEST_PREFIX}/firefox"
```
uberhacker, line 34: Does this interfere with the `firefox` non-esr package?

packages/firefox_esr.rb
```diff
@@ -0,0 +1,37 @@
...
+    system "install -Dm755 firefox.sh #{CREW_DEST_PREFIX}/bin/firefox"
```
uberhacker, line 35: This will interfere with the `firefox` package non-esr binary.  Should we rename to `#{CREW_DEST_PREFIX}/bin/firefox-esr` instead?

packages/firefox_esr.rb
saltedcoffii: Yes, but even if we changed everything they couldn't be installed together as the binaries don't work that way :(
at least as far as i know

uberhacker: We can't have packages that install conflicting files.

saltedcoffii: gettext and gettext_tiny

but okay, i'll close the pr

PR #5195, opened 2021-02-20
[New Package] mailutils

saltedcoffii:


saltedcoffii:
pardon the name thing, i have no clue what just happened, or why I accidentally said mailspring instead of mailutils in the commit 💀

packages/mailutils.rb
```diff
@@ -0,0 +1,36 @@
...
+  version '3.12'
+  compatibility 'all'
+  source_url 'https://ftpmirror.gnu.org/mailutils/mailutils-3.12.tar.xz'
+  source_sha256 '6d43fa217c4ac63f057de87890c562d170bb92bc402368b5fbc579e4c2b3a158'
+
+  depends_on 'python3'
+  depends_on 'libdb'
+  depends_on 'emacs'
+  depends_on 'tcpwrappers'
+  depends_on 'readline'
```
uberhacker, line 15: python3 and readline are core packages so no need to list here.

saltedcoffii: python3 isn't a core package. Or at least it doesn't install during the install script anymore.

uberhacker: That is new news to me.  It's still listed as a core package [here](https://github.com/skycocker/chromebrew/blob/master/tools/core_packages.txt).

saltedcoffii: Weird, I don't have it installed. Okay, I'll remove them.

PR #5196, opened 2021-02-20
[New Package] aha

saltedcoffii:


packages/aha.rb
```diff
@@ -0,0 +1,22 @@
...
+  compatibility 'all'
+  source_url 'https://github.com/theZiz/aha/archive/0.5.1.tar.gz'
+  source_sha256 '6aea13487f6b5c3e453a447a67345f8095282f5acd97344466816b05ebd0b3b1'
+
+  def self.patch
+    system "sed -i 's:-Wextra:-Wextra -pipe -flto -O2:g' Makefile"
+  end
+
+  def self.build
+    system "make"
```
uberhacker, line 16: How do we adjust `prefix` if `#{CREW_PREFIX}` is changed?

PR #5198, opened 2021-02-20
Correct ENV

saltedcoffii:


satmandu:
flto needs to be in LDFLAGS as well, as I understand it...

saltedcoffii:
Let's just remove -flto for now, and when we can confirm that we can add it back.

PR #5203, opened 2021-02-20
Pulsaudio updates

satmandu:
- This adds webrtc echo cancellation to pulseaudio.
- This lets flatpak output to pulseaudio
- BUT still no connection between pulseaudio and alsa.
- also adds abseil_cpp, which is needed for webrtc
- this WILL overwrite the default pulseaudio configs... but if that isn't working for anyone anyways....
- testing is reenabled in pukseaudio.

FYI
- Newer versions of webrtc are broken on i686, and also don't work with pulseaudio yet.

**Please help me get pulseaudio outputting to alsa devices...**

Works properly:
- [x] x86_64

Comnpiles properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/pulseaudio.rb
+++ b/packages/pulseaudio.rb
@@ -51,7 +51,6 @@ class Pulseaudio < Package
   def self.build
     system "meson #{CREW_MESON_LTO_OPTIONS} \
     --default-library=both \
-    -Dcpp_args='-I#{CREW_PREFIX}/include' \
     -Dsystem_user=chronos \
     -Dsystem_group=cras \
     -Daccess_group=cras \
--- a/packages/pulseaudio.rb
+++ b/packages/pulseaudio.rb
@@ -44,7 +44,6 @@ class Pulseaudio < Package
   depends_on 'gst_plugins_base'
   depends_on 'gst_plugins_good'
   depends_on 'gst_plugins_bad'
-  depends_on 'gst_plugins_bad'
   depends_on 'check' => :build
   depends_on 'webrtc_audio_processing'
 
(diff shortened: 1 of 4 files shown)


packages/pulseaudio.rb
```diff
@@ -36,7 +36,17 @@ class Pulseaudio < Package
...
   depends_on 'tdb'
+  depends_on 'cras'
+  depends_on 'orc'
+  depends_on 'jack'
+  depends_on 'avahi'
+  depends_on 'gstreamer'
+  depends_on 'gst_plugins_base'
+  depends_on 'gst_plugins_good'
+  depends_on 'gst_plugins_bad'
+  depends_on 'gst_plugins_bad'
```
uberhacker, line 47: Am I seeing double lines?

uberhacker: I thought about asking you to change to `depends_on 'gst_plugins_ugly'`. :)

packages/pulseaudio.rb
```diff
@@ -46,13 +56,74 @@ def self.build
...
+  end
+
   def self.install
     system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
+    @pulseaudio_daemon_conf = <<~PAUDIO_DAEMON_CONF_HEREDOC
+      # Replace these with the proper values
+      exit-idle-time = 10 # Exit as soon as unneeded
+      flat-volumes = yes # Prevent messing with the master volume
+    PAUDIO_DAEMON_CONF_HEREDOC
+    IO.write("#{CREW_DEST_PREFIX}/etc/pulse/daemon.conf", @pulseaudio_daemon_conf, perm: 0o555)
```
uberhacker, line 90: `perm: 0o555` makes this file read-only.  Are you sure this is correct?

packages/pulseaudio.rb
satmandu: Good catch on the dupe... I'm having to strongly resist making a Rashomon reference though...

packages/pulseaudio.rb
satmandu: good call.. changed them to 666

PR #5204, opened 2021-02-20
New Gnome terminal prereq updates

satmandu:
Compiles  properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/gtk4.rb
+++ b/packages/gtk4.rb
@@ -3,22 +3,22 @@ require 'package'
 class Gtk4 < Package
   description 'GTK+ is a multi-platform toolkit for creating graphical user interfaces.'
   homepage 'https://developer.gnome.org/gtk4/'
-  version '4.0.2'
+  version '4.1.0'
   compatibility 'all'
-  source_url 'https://download.gnome.org/sources/gtk/4.0/gtk-4.0.2.tar.xz'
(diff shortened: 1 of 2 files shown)


packages/gtk4.rb
```diff
@@ -70,9 +69,9 @@ def self.postinstall
     system "glib-compile-schemas #{CREW_PREFIX}/share/glib-2.0/schemas"
     # update mime database
     system "update-mime-database #{CREW_PREFIX}/share/mime"
-    if !File.exists?("#{HOME}/.config/gtk-4.0/settings.ini")
+    unless File.exist?("#{HOME}/.config/gtk-4.0/settings.ini")
       puts
-      puts "Adding basic gtk4 settings to ~/.config/gtk-4.0/settings.ini".lightblue
+      puts 'Adding basic gtk4 settings to ~/.config/gtk-4.0/settings.ini'.lightblue
       FileUtils.mkdir_p "#{HOME}/.config/gtk-4.0"
       system "cat << 'EOF' > #{HOME}/.config/gtk-4.0/settings.ini
```
uberhacker, line 76: Would it make sense to add this in the `def self.build` section and include the config in the package?  The drawback would be the config goes away when the package is removed.

satmandu: The problem is that one might have local customizations to these files.  You need to make sure not to overwrite the settings for these.

PR #5205, opened 2021-02-20
Split ncurses

saltedcoffii:
needed for some packages i'm working on


packages/ncurses.rb
```diff
@@ -3,73 +3,40 @@
...
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/ncurses-6.2-20210206-1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/ncurses-6.2-20210206-1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/ncurses-6.2-20210206-1-chromeos-x86_64.tar.xz'
-  })
-  binary_sha256({
-    aarch64: 'f87816a803fdbb6403ecb03ea4faec04da8c3a829a1455a2c9e0a0686818ebb6',
-     armv7l: 'f87816a803fdbb6403ecb03ea4faec04da8c3a829a1455a2c9e0a0686818ebb6',
-       i686: 'ad1fb3a7d832d4fcf4e695cd5f5f7e023581c2d0529a2bf9729acd2cd96e44cd',
-     x86_64: 'd81416072e4e4454b585495db8b7e0767867c851e512afd4655fa8174d244c02'
-  })
```
uberhacker, line 22: We'll need binaries since this is a core package.

saltedcoffii: I'll compile those for x86_64 and drop them in here. I can't do any of the other architectures.

packages/ncursesw.rb
```diff
@@ -0,0 +1,37 @@
+require 'package'
+
+class Ncursesw < Package
```
uberhacker, line 3: Would we need to add this new package as a dependency for other packages or should it be a core package?

saltedcoffii: Core package.

saltedcoffii: Although the different libraries are source compatible, they aren't binary compatible and anything built with ncursesw will break with its removal

uberhacker: In order to make it a core package, we would need to do the following:

1. Add to install.sh -or- add as a dependency of another core package so it is "inherited"
2. Add to tools/core_packages.txt

saltedcoffii:
Ncurses: https://mega.nz/file/vThASQpJ#EH2hiI2r9Hv_T8lZJ5z5G9pUhb7tPWwoQZ-v9lg80WM
sha256: 94637e04e7800cd16d6e9e777b8b06d46185a98ab9ff82bec1e7775af007e218

Ncusesw: https://mega.nz/file/LSp2UKLD#ilT9hyCZoPo-biz5KtTpEFV8uNSpmCQzABoGWsTKOdo
sha256: a7ba96516142fc465f0becc68ecd14312dcd7fa298c7dcac9d737f19cbd683c2

saltedcoffii:
Ncurses depends on ncursesw

saltedcoffii:
I think it's ready to merge

uberhacker:
@satmandu: Did you want to create binaries for this PR?  If not, I can do it after work this evening.

satmandu:
Do NOT merge this until we figure out how to deal with libtinfo getting
removed. We don't want to break any systems.

Can we merge an update to this and libtinfo simultaneously (and maybe to
crew?) such that we delete the package database info for libtinfo at the
same time?

satmandu:
FYI, the ```--with-build-cflags``` and the others don't appear to do anything. I'm not seeing ```flto``` passed at all.

saltedcoffii:
Looks like ncurses installs libtinfo and the libtinfo package isn't necessary anymore. Ncurses installs a libtinfo library, the libtinfo package overwrites it with a symlink and then when libtinfo is removed, it removes the symlink and breaks the package.

```
$ crew files ncurses | grep tinfo
/usr/local/lib64/libtinfo.a
/usr/local/lib64/libtinfo.so
/usr/local/lib64/libtinfo.so.6
/usr/local/lib64/libtinfo.so.6.2
/usr/local/lib64/libtinfow.a
/usr/local/lib64/libtinfow.so
/usr/local/lib64/libtinfow.so.6
/usr/local/lib64/libtinfow.so.6.2
/usr/local/lib64/pkgconfig/tinfo.pc
/usr/local/lib64/pkgconfig/tinfow.pc
/usr/local/share/man/man3/_nc_free_tinfo.3x.gz
```

uberhacker:
If you read the comments in the libtinfo package, the main purpose is to symlink to libtinfo.so.5.  I know the freebasic package needs this file.  See below:
```
    # See https://forums.opensuse.org/showthread.php/446927-missing-library-libtinfo-so-5.
    # See also http://www.linuxforums.org/forum/installation/6251-libtinfo-so-5-a.html.
    FileUtils.mkdir_p CREW_DEST_LIB_PREFIX
    FileUtils.ln_sf "#{CREW_LIB_PREFIX}/libncurses.so.6", "#{CREW_DEST_LIB_PREFIX}/libtinfo.so.5"
    FileUtils.ln_sf "#{CREW_LIB_PREFIX}/libncurses.so.6", "#{CREW_DEST_LIB_PREFIX}/libtinfo.so.6"
```
I think the solution is to remove the last line since that conflicts with ncurses.

uberhacker:
Can we close this @saltedcoffii?

saltedcoffii:
Yes, however I think I should redesign this and open it again. We don't need both regular and wide ncurses libraries, just wide ones. And also fix libtinfo.

PR #5207, opened 2021-02-21
make flatpak quieter by adding to XDG_DATA_DIRS...

satmandu:
- linted with rubocop.
- These added folders get brought up by flatpak at every run unless they are in ```XDG_DATA_DIRS```

Works properly:
- [x] x86_64

### DIFF
--- a/packages/xdg_base.rb
+++ b/packages/xdg_base.rb
@@ -3,18 +3,19 @@ require 'package'
 class Xdg_base < Package
   description 'XDG Base Directory Specification Configuration'
   homepage 'https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html'
-  version '0.7-6'
+  version '0.7-7'
   compatibility 'all'
   source_url 'file:///dev/null'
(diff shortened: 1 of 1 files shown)


packages/xdg_base.rb
```diff
@@ -36,12 +37,12 @@ def self.postinstall
       system "echo '# See https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html' >> ~/.bashrc"
       system "echo 'export XDG_DATA_HOME=#{HOME}/.local/share' >> ~/.bashrc"
       system "echo 'export XDG_CONFIG_HOME=#{HOME}/.config' >> ~/.bashrc"
-      system "echo 'export XDG_DATA_DIRS=#{CREW_PREFIX}/share' >> ~/.bashrc"
+      system "echo 'export XDG_DATA_DIRS=#{CREW_PREFIX}/share:#{HOME}/.local/share/flatpak/exports/share:#{CREW_PREFIX}/var/lib/flatpak/exports/share' >> ~/.bashrc"
```
uberhacker, line 40: Should these paths be added in the postinstall section of flatpak instead of this package?

satmandu: hmmm... That's a good idea.

PR #5212, opened 2021-02-21
Most is not an essential package.

saltedcoffii:


uberhacker:
I would argue less is more but more or less most is best. :)

satmandu:
As a person who was taught to use vi in the '80s as a child, I'm going to laugh if this leads to a PR stating “Emacs is not an essential package.”

uberhacker:
Emacs is not even useful when it's installed. :laughing:

saltedcoffii:
https://en.wikipedia.org/wiki/Editor_war

saltedcoffii:
I guess the reason I wanted to get rid of most is because its not necessary to build packages like buildessential, and has more dependencies and is bigger than less. Most is kinda "extra" for a core package. It would be like installing exa by default under the argument that it's better than ls (it is).

uberhacker:
Probably not worth the effort but if you really insist on removing it, you will need to also do the following:
```
Alternatively, you could use most.  Why settle for less, right?
echo "crew install most"
echo \"export PAGER='most'\" >> ~/.bashrc && . ~/.bashrc
```
And then remove `most` and `slang` from `tools/core_packages.txt`.

My guess is as soon as we do this, we'll need to start including `depends_on 'slang'` in some packages. :-D

saltedcoffii:
We could add slang to the core package list, or just make the change, get bugs and update package dependencies.

uberhacker:
Like I said, probably not worth the effort. :wink:

saltedcoffii:
Alright

uberhacker:
Closing with this:
_"Emacs is a great operating system; it just lacks a good editor"_

PR #5213, opened 2021-02-21
New Package: gnome_sudoku and its dependencies

supechicken:
Tested on `x86_64`

### DIFF
--- a/packages/clutter_gtk.rb
+++ b/packages/clutter_gtk.rb
@@ -17,8 +17,7 @@ class Clutter_gtk < Package
   depends_on 'clutter'
 
   def self.build
-    system "env CFLAGS='-fuse-ld=lld' CXXFLAGS='-fuse-ld=lld' \
-            meson #{CREW_MESON_LTO_OPTIONS} builddir"
+    system "meson #{CREW_MESON_LTO_OPTIONS} builddir"
     system 'meson configure builddir'
     system 'ninja -C builddir'
   end
--- a/packages/clutter_gtk.rb
+++ b/packages/clutter_gtk.rb
@@ -3,6 +3,7 @@ require 'package'
 class Clutter_gtk < Package
   description 'The Clutter Gtk package is a library providing facilities to integrate Clutter into GTK+ applications.'
   homepage 'https://wiki.gnome.org/Projects/Clutter'
+  compatibility 'all'
   @_app = 'clutter-gtk'
   @_ver = '1.8'
   @_patch = '.4'
--- a/packages/clutter_gtk.rb
+++ b/packages/clutter_gtk.rb
@@ -6,11 +6,10 @@ class Clutter_gtk < Package
   @_app = 'clutter-gtk'
   @_ver = '1.8'
   @_patch = '.4'
-  @_fullver = @_ver + @_patch
-  version @_fullver
-  source_url "https://download.gnome.org/sources/#{@_app}/#{@_ver}/#{@_app}-#{@_fullver}.tar.xz"
-  source_sha256 `curl -Ls https://download.gnome.org/sources/#{@_app}/#{@_ver}/#{@_app}-#{@_fullver}.sha256sum |\
-                 tail -n1 | cut -d ' ' -f1`.chomp
+  @_url = "https://download.gnome.org/sources/#{@_app}/#{@_ver}/#{@_app}-#{@_ver}#{@_patch}"
+  version @_ver + @_patch
+  source_url "#{@_url}.tar.xz"
+  source_sha256 `curl -Ls #{@_url}.sha256sum | tail -n1 | cut -d ' ' -f1`.chomp
   
   depends_on 'gtk3'
   depends_on 'libgee'
--- a/packages/clutter_gtk.rb
+++ b/packages/clutter_gtk.rb
@@ -1,0 +1,29 @@
+require 'package'
+
+class Clutter_gtk < Package
+  description 'The Clutter Gtk package is a library providing facilities to integrate Clutter into GTK+ applications.'
+  homepage 'https://wiki.gnome.org/Projects/Clutter'
+  @_app = 'clutter-gtk'
+  @_ver = '1.8'
+  @_patch = '.4'
+  @_fullver = @_ver + @_patch
+  version @_fullver
+  source_url "https://download.gnome.org/sources/#{@_app}/#{@_ver}/#{@_app}-#{@_fullver}.tar.xz"
+  source_sha256 `curl -Ls https://download.gnome.org/sources/#{@_app}/#{@_ver}/#{@_app}-#{@_fullver}.sha256sum |\
+                 tail -n1 | cut -d ' ' -f1`.chomp
+  
+  depends_on 'gtk3'
+  depends_on 'libgee'
+  depends_on 'clutter'
+
+  def self.build
+    system "env CFLAGS='-fuse-ld=lld' CXXFLAGS='-fuse-ld=lld' \
+            meson #{CREW_MESON_LTO_OPTIONS} builddir"
+    system 'meson configure builddir'
+    system 'ninja -C builddir'
+  end
+
+  def self.install
+    system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
+  end
+end
--- a/packages/libgnome_games_support.rb
+++ b/packages/libgnome_games_support.rb
@@ -1,0 +1,29 @@
+require 'package'
+
+class Libgnome_games_support < Package
+  description 'libgnome-games-support is a small library intended for internal use by GNOME Games.'
+  homepage 'https://gitlab.gnome.org/GNOME/libgnome-games-support'
+  @_app = 'libgnome-games-support'
+  @_ver = '1.8'
+  @_patch = '.0'
+  @_fullver = @_ver + @_patch
+  version @_fullver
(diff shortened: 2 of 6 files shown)


packages/clutter_gtk.rb
```diff
@@ -0,0 +1,29 @@
...
+  version @_ver + @_patch
+  source_url "#{@_url}.tar.xz"
+  source_sha256 `curl -Ls #{@_url}.sha256sum | tail -n1 | cut -d ' ' -f1`.chomp
+  
+  depends_on 'gtk3'
+  depends_on 'libgee'
+  depends_on 'clutter'
+
+  def self.build
+    system "env CFLAGS='-fuse-ld=lld' CXXFLAGS='-fuse-ld=lld' \
```
satmandu, line 20: Do we need lld for this project? Why?

packages/libgnome_games_support.rb
```diff
@@ -0,0 +1,28 @@
...
+  version @_ver + @_patch
+  source_url "#{@_url}.tar.xz"
+  source_sha256 `curl -Ls #{@_url}.sha256sum | tail -n1 | cut -d ' ' -f1`.chomp
+  
+  depends_on 'gtk3'
+  depends_on 'libgee'
+  depends_on 'clutter'
+
+  def self.build
+    system "env CFLAGS='-fuse-ld=lld' CXXFLAGS='-fuse-ld=lld' \
```
satmandu, line 19: Same qiuestion about why we might need lld for this.

packages/qqwing.rb
```diff
@@ -0,0 +1,37 @@
...
+    system 'make', "prefix=#{CREW_PREFIX}", 'tgz'
+  end
+
+  def self.install
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    FileUtils.mkdir_p CREW_DEST_LIB_PREFIX
+    FileUtils.ln_s "#{CREW_PREFIX}/lib/libqqwing.so", "#{CREW_DEST_LIB_PREFIX}/libqqwing.so"
+    FileUtils.ln_s "#{CREW_PREFIX}/lib/libqqwing.so", "#{CREW_DEST_LIB_PREFIX}/libqqwing.so.2"
+    FileUtils.ln_s "#{CREW_PREFIX}/lib/libqqwing.so", "#{CREW_DEST_LIB_PREFIX}/libqqwing.so.2.1.0"
+    FileUtils.ln_s "#{CREW_PREFIX}/lib/libqqwing.la", "#{CREW_DEST_LIB_PREFIX}/libqqwing.la"
```
uberhacker, line 26: Would it make sense to adjust the Makefile to put the libraries in the correct location.  In any event, this will crash on all non-x86_64 architectures since `#{CREW_PREFIX}/lib` is the same as `#{CREW_DEST_LIB_PREFIX}` since it will try to symlink to the same location.

packages/qqwing.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def self.check
+    system "make", "check"
+  end
+    
+  def self.remove
+    # since qqwing doesn't fully support custom prefix and DESTDIR, we need to remove them manually
+    FileUtils.rm_rf Dir.glob("#{CREW_PREFIX}/lib/libqqwing.*") if ARGV[1] == 'remove'
```
uberhacker, line 35: You don't need the `if ARV[1] == 'remove'` because this method only gets called on `crew remove ...`. :)

packages/qqwing.rb
uberhacker: You also don't need to include `libqqwing.la`.  That file is only needed during the build.

packages/clutter_gtk.rb
```diff
@@ -0,0 +1,29 @@
...
+  description 'The Clutter Gtk package is a library providing facilities to integrate Clutter into GTK+ applications.'
+  homepage 'https://wiki.gnome.org/Projects/Clutter'
+  compatibility 'all'
+  @_app = 'clutter-gtk'
+  @_ver = '1.8'
+  @_patch = '.4'
+  @_url = "https://download.gnome.org/sources/#{@_app}/#{@_ver}/#{@_app}-#{@_ver}#{@_patch}"
+  version @_ver + @_patch
+  source_url "#{@_url}.tar.xz"
+  source_sha256 `curl -Ls #{@_url}.sha256sum | tail -n1 | cut -d ' ' -f1`.chomp
```
satmandu, line 13: These need to be actual hashes, not the code which would generate a sha256 hash post-hoc.

satmandu (changes requested):
FYI: 
```
curl -Ls https://download.gnome.org/sources/clutter-gtk/1.8/clutter-gtk-1.8.4.tar.xz | sha256sum -
521493ec038973c77edcb8bc5eac23eed41645117894aaee7300b2487cb42b06  -
```

packages/clutter_gtk.rb
supechicken: I think the checksum is a normal file and won't be changed:
```
gtk+-3.24.20.sha256sum | 170 B | 2020-Apr-27 15:12
gtk+-3.24.21.sha256sum | 170 B | 2020-Jun-28 21:36
```

uberhacker: This is actually a good idea from @supechicken.  We would only need to bump the version on every update.  The problem is not all sources come with a sha256 hash so we can't use it for every package.

packages/clutter_gtk.rb
supechicken: Removed

packages/clutter_gtk.rb
satmandu: My apologies to @supechicken ... I've been too dense to realize you're downloading the sha256sums for these gnome files from the archive directory.

satmandu: Carry on...

satmandu (approved):

packages/clutter_gtk.rb
uberhacker: We could do this for sources that don't have a sha256 checksum file:
```
source_sha256  `curl -Ls #{source_url} | sha256sum | cut -d' ' -f1`.chomp
```

packages/qqwing.rb
```diff
@@ -0,0 +1,36 @@
...
+  def self.build
+    system 'make', "prefix=#{CREW_PREFIX}", 'tgz'
+  end
+
+  def self.install
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    FileUtils.mkdir_p CREW_DEST_LIB_PREFIX
+    FileUtils.ln_s "#{CREW_PREFIX}/lib/libqqwing.so", "#{CREW_DEST_LIB_PREFIX}/libqqwing.so"
+    FileUtils.ln_s "#{CREW_PREFIX}/lib/libqqwing.so", "#{CREW_DEST_LIB_PREFIX}/libqqwing.so.2"
+    FileUtils.ln_s "#{CREW_PREFIX}/lib/libqqwing.so", "#{CREW_DEST_LIB_PREFIX}/libqqwing.so.2.1.0"
```
uberhacker, line 25: This is only needed for x86_64 architecture.  Actually, have you tried `"LIBDIR=#{CREW_LIB_PREFIX}"` instead?

packages/qqwing.rb
```diff
@@ -0,0 +1,36 @@
...
+class Qqwing < Package
+  description 'QQwing is a Sudoku puzzle generator and solver.'
+  homepage 'https://qqwing.com/'
+  @_ver = '1.3.4'
+  version @_ver
+  source_url "https://github.com/stephenostermiller/qqwing/archive/v#{@_ver}.tar.gz"
+  source_sha256 'dc5d61b4470862b190d437f8143d4090639c164d34461c4caa5c081c5c91e6cc'
+
+  def self.patch
+    system 'sed -i s:@build/src-copyright-check.sh::g Makefile'
```
uberhacker, line 12: Does this pose a licensing violation issue?

packages/clutter_gtk.rb
supechicken: > We could do this for sources that don't have a sha256 checksum file:
> 
> ```
> source_sha256  `curl -Ls #{source_url} | sha256sum | cut -d' ' -f1`.chomp
> ```

Wait... The purpose of the checksum checking is to verify the file we downloaded, This command will generate a checksum according to the downloaded file, right?

packages/qqwing.rb
```diff
@@ -0,0 +1,36 @@
...
+    FileUtils.ln_s "#{CREW_PREFIX}/lib/libqqwing.so", "#{CREW_DEST_LIB_PREFIX}/libqqwing.so.2.1.0"
+  end
+
+  def self.check
+    system "make", "check"
+  end
+    
+  def self.remove
+    # since qqwing doesn't fully support custom prefix and DESTDIR, we need to remove them manually
+    FileUtils.rm_rf Dir.glob("#{CREW_PREFIX}/lib/libqqwing.*")
```
uberhacker, line 34: We shouldn't need to do this if those files are included in the package metadata.

packages/qqwing.rb
supechicken: I am not sure about it (it builds failed with that script)

packages/clutter_gtk.rb
uberhacker: Yes, but we would have to do that anyway to determine the sha256 checksum if there wasn't a `.sha256` file available, right?

packages/qqwing.rb
uberhacker: Have you read the copyright text to determine the legal restrictions, if any?

PR #5217, opened 2021-02-21
[New Package] libdeflate

saltedcoffii:


packages/libdeflate.rb
```diff
@@ -0,0 +1,18 @@
...
+class Libdeflate < Package
+  description 'Heavily optimized library for DEFLATE compression and decompression'
+  homepage 'https://github.com/ebiggers/libdeflate/'
+  version '1.7'
+  compatibility 'all'
+  source_url 'https://github.com/ebiggers/libdeflate/archive/v1.7.tar.gz'
+  source_sha256 'a5e6a0a9ab69f40f0f59332106532ca76918977a974e7004977a9498e3f11350'
+  
+  def self.build
+    system 'make', "PREFIX=#{CREW_PREFIX}"
```
uberhacker, line 12: I'm guessing you also need `"LIBDIR=#{CREW_LIB_PREFIX}"`?

saltedcoffii: Read the makefile, looks like you do!

packages/libdeflate.rb
```diff
@@ -0,0 +1,23 @@
...
+  description 'Heavily optimized library for DEFLATE compression and decompression'
+  homepage 'https://github.com/ebiggers/libdeflate/'
+  version '1.7'
+  compatibility 'all'
+  source_url 'https://github.com/ebiggers/libdeflate/archive/v1.7.tar.gz'
+  source_sha256 'a5e6a0a9ab69f40f0f59332106532ca76918977a974e7004977a9498e3f11350'
+  
+  def self.patch
+    system "sed -i 's:PREFIX ?= /usr/local:PREFIX ?= #{CREW_PREFIX}:g' Makefile"
+    system "sed -i 's:LIBDIR ?= $(PREFIX)/lib:LIBDIR ?= #{CREW_LIB_PREFIX}:g' Makefile"
```
uberhacker, line 13: This works but it's more elegant to set the variables on the make line.

saltedcoffii: It wasn't working otherwise I would've done it.

uberhacker: What happens if you do this?
```
def self.build
  system 'make', "PREFIX=#{CREW_PREFIX}", "LIBDIR=#{CREW_LIB_PREFIX}"
end

def self.install
  system 'make', "DESTDIR=#{CREW_DEST_DIR}", "PREFIX=#{CREW_PREFIX}", "LIBDIR=#{CREW_LIB_PREFIX}", 'install'
end
```

saltedcoffii: works

PR #5234, opened 2021-02-22
meson and ninja needs python3

supechicken:


packages/ninja.rb
```diff
@@ -20,6 +20,8 @@ class Ninja < Package
        i686: 'ec2996217bdf7c57d0528a8ba0eefabda15c17e59f958daebc3fe1eda6a8a9bc',
      x86_64: 'e74f2b1d374edd0ff0877939b3afb8fa80240774bf3c9126278e5a965c7a5d0b',
   })
+  
+  depends_on 'python3'
```
uberhacker, line 24: This is a core package so no need to list here.  Are you saying python3 is not installed initially?

tools/core_packages.txt
```diff
@@ -76,7 +76,6 @@ perl_text_unidecode
 perl_unicode_eastasianwidth
 perl_xml_parser
 pkgconfig
-python27
```
uberhacker, line 79: Python 2 is not going away any time soon.  Too many apps still depend on it with no upgrade path to Python 3.

packages/ninja.rb
uberhacker: @supechicken:  What are your thoughts?

uberhacker:
@supechicken is right.  python27 and python3 are no longer installed initially with the install.sh script.  @satmandu: Is this because you removed the dependency from meson and/or ninja?  Personally, I think both python packages should be core packages just because so many packages need python as a dependency and it would be overkill to have to include `depends_on 'python...` lines everywhere.

satmandu:
It is possible a recent update removed those from some core package's dependencies.

Since core packages need to be installed, can we just iterate through and crew install all packages in the core packages list right from install.sh?

satmandu:
Just like build essentials, we could have a core packages dummy package?

tools/core_packages.txt
uberhacker: @satmandu: How can we restore python27 and python3 back to core?

satmandu: See https://github.com/skycocker/chromebrew/pull/5243

satmandu:
The easiest way to fix this is to just add them as core dependencies of buildessential, hence: https://github.com/skycocker/chromebrew/pull/5243

uberhacker:
Fixed in #5243.

PR #5238, opened 2021-02-22
[Update Package] mono

saltedcoffii:
Also enabled libgdiplus support in mono.

### DIFF
--- a/packages/mono.rb
+++ b/packages/mono.rb
@@ -3,40 +3,32 @@ require 'package'
 class Mono < Package
   description 'Mono is a software platform designed to allow developers to easily create cross platform applications part of the .NET Foundation.'
   homepage 'http://www.mono-project.com/'
-  version '6.8.0.123-1'
+  version '6.12.0.107'
   compatibility 'all'
-  source_url 'https://download.mono-project.com/sources/mono/mono-6.8.0.123.tar.xz'
-  source_sha256 'e2e42d36e19f083fc0d82f6c02f7db80611d69767112af353df2f279744a2ac5'
+  source_url 'https://download.mono-project.com/sources/mono/mono-6.12.0.107.tar.xz'
+  source_sha256 '61f3cd629f8e99371c6b47c1f8d96b8ac46d9e851b5531eef20cdf9ab60d2a5f'
 
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/mono-6.8.0.123-1-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/mono-6.8.0.123-1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/mono-6.8.0.123-1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/mono-6.8.0.123-1-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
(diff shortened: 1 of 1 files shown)


packages/mono.rb
```diff
@@ -3,40 +3,32 @@
...
 
+  def self.prebuild
+    system "if [ ! -f /proc/config.gz ]; then sudo modprobe configs -v; fi"
+    system "cat /proc/config.gz | gunzip | grep SYSVIPC=y || false" # Mono build hangs without this feature enabled.
+  end
+  
   def self.build
-    system "./configure #{CREW_OPTIONS} --disable-dependency-tracking --disable-maintainer-mode --disable-silent-rules --enable-nls=no"
-    system 'make'
+    ENV['XMKMF'] = '/usr/local/bin/xmkmf'
```
uberhacker, line 22: `ENV['XMKMF'] = "#{CREW_PREFIX}/bin/xmkmf"`?  It would probably be better to include on the `./configure` line, actually.
```
system "env XMKMF='#{CREW_PREFIX}/bin/xmkmf' ./configure ...
```

packages/mono.rb
```diff
@@ -3,40 +3,32 @@
...
-    Dir.chdir "#{CREW_DEST_PREFIX}/bin" do
-      system "for f in \$(ls #{CREW_BUILD}-*); do g=\$(echo \$f | sed 's,#{CREW_BUILD}-,,'); ln -sf \$f \$g; done"
-      FileUtils.ln_sf "#{CREW_BUILD}-mono-sgen", 'mono'
-    end
-    Dir.chdir "#{CREW_DEST_MAN_PREFIX}/man1" do
-      system "for f in \$(ls #{CREW_BUILD}-*); do g=\$(echo \$f | sed 's,#{CREW_BUILD}-,,'); ln -sf \$f \$g; done"
-    end
-    Dir.chdir "#{CREW_DEST_MAN_PREFIX}/man5" do
-      system "for f in \$(ls #{CREW_BUILD}-*); do g=\$(echo \$f | sed 's,#{CREW_BUILD}-,,'); ln -sf \$f \$g; done"
-    end
```
uberhacker, line 40: Are sure this should be removed?  Do binaries and man pages still work?

packages/mono.rb
```diff
@@ -3,40 +3,32 @@
...
-  })
-  binary_sha256 ({
-    aarch64: '8e62055873099e3f943dd33789274c3db2d5436ddec704c72c8cf758d24f6ee5',
-     armv7l: '8e62055873099e3f943dd33789274c3db2d5436ddec704c72c8cf758d24f6ee5',
-       i686: 'd0442cd9a809387cf79cb6a605018244704ccd25333fcc33f1e8440fc0c52a04',
-     x86_64: '46077f0cefcfc389006e730755ec9832a7624d67b9285b27410655cd65b55083',
-  })
+  depends_on 'bc'
+  depends_on 'libgdiplus'
+  depends_on 'python3' => :build
```
uberhacker, line 13: This should be a core package.

packages/mono.rb
saltedcoffii: Okay

packages/mono.rb
saltedcoffii: Lemme rebuild and check up on that

packages/mono.rb
saltedcoffii: It isn't though, weirdly enough. Or at least it doesn't get installed by the install script anymore.

saltedcoffii:
To save everyone time here's an x86_64 binary:
https://mega.nz/file/WXxQ0TTC#TDrziqMSGi52DY3IDgvNn-xikGFmHnYddt-sJ1PmBWI
SHA256: `c7580f1e50fe7d96fbe43175fe052be02f3668288f58965796558f800ce40fc2`

PR #5239, opened 2021-02-22
Default bin/python to python3

satmandu:
- Python2 is deprecated. Set default to python3.
- We will have to fix packages as people complain, but this is better than defaulting to a package no longer updated.
- This makes no changes to the binaries in the python 2.7 package.
- Should we also default pip to pip3 and leave pip2 as the version for python2?

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/python27.rb
+++ b/packages/python27.rb
@@ -65,5 +65,9 @@ class Python27 < Package
     #puts "Upgrading pip...".lightblue
     #system 'pip2 install --upgrade pip'
     #puts
+    if File.exist? "#{CREW_PREFIX}/bin/python3" 
+      FileUtils.rm "#{CREW_PREFIX}/bin/python"
+      FileUtils.ln_s "#{CREW_PREFIX}/bin/python3","#{CREW_PREFIX}/bin/python"
+    end
   end
 end
--- a/packages/python27.rb
+++ b/packages/python27.rb
@@ -67,7 +67,7 @@ class Python27 < Package
     #puts
     FileUtils.ln_sf "#{CREW_PREFIX}/bin/python3", "#{CREW_PREFIX}/bin/python" \
       if File.exist? "#{CREW_PREFIX}/bin/python3"
-    FileUtils.cp "#{CREW_PREFIX}/bin/pip3", "#{CREW_PREFIX}/bin/pip" \
+    FileUtils.ln_sf "#{CREW_PREFIX}/bin/pip3", "#{CREW_PREFIX}/bin/pip" \
       if File.exist? "#{CREW_PREFIX}/bin/pip3"
   end
 end
--- a/packages/python27.rb
+++ b/packages/python27.rb
@@ -65,9 +65,9 @@ class Python27 < Package
     #puts "Upgrading pip...".lightblue
     #system 'pip2 install --upgrade pip'
     #puts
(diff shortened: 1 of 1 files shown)


packages/python27.rb
```diff
@@ -65,5 +65,9 @@ def self.postinstall
     #puts "Upgrading pip...".lightblue
     #system 'pip2 install --upgrade pip'
     #puts
+    if File.exist? "#{CREW_PREFIX}/bin/python3" 
+      FileUtils.rm "#{CREW_PREFIX}/bin/python"
+      FileUtils.ln_s "#{CREW_PREFIX}/bin/python3","#{CREW_PREFIX}/bin/python"
```
uberhacker, line 70: You could get rid of line 69 with `FileUtils.ln_sf ...`.  You could also make this a one-liner with:
```
FileUtils.ln_sf "#{CREW_PREFIX}/bin/python3", "#{CREW_PREFIX}/bin/python" if File.exist? "#{CREW_PREFIX}/bin/python3"
```
Does it make sense to add this logic in the python3 package also in the hypothetical case where python27 is installed after python3 and `#{CREW_PREFIX}/bin/python` is version 2?

satmandu: I have this as an automatic function in the python3 update I just submitted.

satmandu: How dioes thi look?

cstrouse:
We should be using `python -m pip` so whichever version is being used link or alias accordingly.

saltedcoffii (approved):

uberhacker:
@cstrouse: Will that overwrite any existing symlinks?

satmandu:
@cstrouse I'm setting pip to be a copy of  pip3 in the postinstall for this and pythion3 as well.  Interestingly ```/usr/bin/pip``` isn't a symlink at all.

packages/python27.rb
```diff
@@ -65,5 +65,9 @@ def self.postinstall
     #puts "Upgrading pip...".lightblue
     #system 'pip2 install --upgrade pip'
     #puts
+    FileUtils.ln_sf "#{CREW_PREFIX}/bin/python3", "#{CREW_PREFIX}/bin/python" \
+      if File.exist? "#{CREW_PREFIX}/bin/python3"
+    FileUtils.cp "#{CREW_PREFIX}/bin/pip3", "#{CREW_PREFIX}/bin/pip" \
+      if File.exist? "#{CREW_PREFIX}/bin/pip3"
```
uberhacker, line 71: Should we use `FileUtils.ln_sf "#{CREW_PREFIX}/bin/pip3", "#{CREW_PREFIX}/bin/pip" \
      if File.exist? "#{CREW_PREFIX}/bin/pip3"` instead?

saltedcoffii:
Perhaps the file at `/usr/local/bin/pip` should just be
```bash
#!/bin/sh
python3 -m pip $@
```
symlink `/usr/local/bin/pip` to `/usr/local/bin/pip3`
the file at `/usr/local/bin/pip2 should be
```bash
#!/bin/sh
python2 -m pip
```
So that way users and devs are always sure which interpreter they're using. Can someone test it and see if it works?

packages/python27.rb
satmandu: Sure. Aligning with the python3 package...

PR #5240, opened 2021-02-22
Python3 updates

satmandu:
- updates for gdbm, tcl, tk (python was complaining about them not being available.)
- Python -> 3.9.2
- compiled with lto enabled
- compiled with ```-fno-semantic-interposition``` as per [redhat](https://bugzilla.redhat.com/show_bug.cgi?id=1779341) there is a 27% speed improvement at the expense of a longer compile time.

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -108,8 +108,7 @@ class Python3 < Package
     puts 'Upgrading pip...'.lightblue
     system "curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && LD_LIBRARY_PATH=#{CREW_LIB_PREFIX} python3 get-pip.py -U"
     puts 'Setting Python3 as default python...'.lightblue
-    FileUtils.ln_s "#{CREW_PREFIX}/bin/python3", "#{CREW_PREFIX}/bin/python"
-    FileUtils.rm "#{CREW_PREFIX}/bin/pip"
-    FileUtils.cp "#{CREW_PREFIX}/bin/pip3", "#{CREW_PREFIX}/bin/pip"
+    FileUtils.ln_sf "#{CREW_PREFIX}/bin/python3", "#{CREW_PREFIX}/bin/python"
+    FileUtils.ln_sf "#{CREW_PREFIX}/bin/pip3", "#{CREW_PREFIX}/bin/pip"
   end
 end
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -108,7 +108,6 @@ class Python3 < Package
     puts 'Upgrading pip...'.lightblue
     system "curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && LD_LIBRARY_PATH=#{CREW_LIB_PREFIX} python3 get-pip.py -U"
     puts 'Setting Python3 as default python...'.lightblue
-    FileUtils.rm "#{CREW_PREFIX}/bin/python"
     FileUtils.ln_s "#{CREW_PREFIX}/bin/python3", "#{CREW_PREFIX}/bin/python"
     FileUtils.rm "#{CREW_PREFIX}/bin/pip"
     FileUtils.cp "#{CREW_PREFIX}/bin/pip3", "#{CREW_PREFIX}/bin/pip"
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -3,22 +3,23 @@ require 'package'
 class Python3 < Package
   description 'Python is a programming language that lets you work quickly and integrate systems more effectively.'
   homepage 'https://www.python.org/'
-  version '3.9.1-1'
+  @_ver = '3.9.2'
+  version @_ver
   compatibility 'all'
-  source_url 'https://www.python.org/ftp/python/3.9.1/Python-3.9.1.tar.xz'
-  source_sha256 '991c3f8ac97992f3d308fefeb03a64db462574eadbff34ce8bc5bb583d9903ff'
+  source_url "https://www.python.org/ftp/python/#{@_ver}/Python-#{@_ver}.tar.xz"
+  source_sha256 '3c2034c54f811448f516668dce09d24008a0716c3a794dd8639b5388cbde247d'
 
-  binary_url ({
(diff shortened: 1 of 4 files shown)


packages/python3.rb
```diff
@@ -90,8 +105,12 @@ def self.install
...
     puts
-    puts "Upgrading pip...".lightblue
+    puts 'Upgrading pip...'.lightblue
     system "curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && LD_LIBRARY_PATH=#{CREW_LIB_PREFIX} python3 get-pip.py -U"
-    puts
+    puts 'Setting Python3 as default python...'.lightblue
+    FileUtils.rm "#{CREW_PREFIX}/bin/python"
+    FileUtils.ln_s "#{CREW_PREFIX}/bin/python3", "#{CREW_PREFIX}/bin/python"
+    FileUtils.rm "#{CREW_PREFIX}/bin/pip"
+    FileUtils.cp "#{CREW_PREFIX}/bin/pip3", "#{CREW_PREFIX}/bin/pip"
```
uberhacker, line 114: Shouldn't need the `FileUtils.rm ...` lines with `FileUtils.ln_sf ...` and `FileUtils.cp_f ...`.

uberhacker: Should we be using `FileUtils.ln_sf "#{CREW_PREFIX}/bin/pip3", "#{CREW_PREFIX}/bin/pip"` instead?

packages/tcl.rb
```diff
@@ -40,10 +42,14 @@ def self.install
...
       FileUtils.ln_s "#{CREW_LIB_PREFIX}/tdbc1.1.1/libtdbc1.1.1.so", "#{CREW_DEST_LIB_PREFIX}/libtdbc1.1.1.so"
       FileUtils.ln_s "#{CREW_LIB_PREFIX}/thread2.8.5/libthread2.8.5.so", "#{CREW_DEST_LIB_PREFIX}/libthread2.8.5.so"
-      FileUtils.ln_s "#{CREW_LIB_PREFIX}/tdbcodbc1.1.1/libtdbcodbc1.1.1.so", "#{CREW_DEST_LIB_PREFIX}/libtdbcodbc1.1.1.so"
-      FileUtils.ln_s "#{CREW_LIB_PREFIX}/sqlite3.30.1.2/libsqlite3.30.1.2.so", "#{CREW_DEST_LIB_PREFIX}/libsqlite3.30.1.2.so"
-      FileUtils.ln_s "#{CREW_LIB_PREFIX}/tdbcmysql1.1.1/libtdbcmysql1.1.1.so", "#{CREW_DEST_LIB_PREFIX}/libtdbcmysql1.1.1.so"
-      FileUtils.ln_s "#{CREW_LIB_PREFIX}/tdbcpostgres1.1.1/libtdbcpostgres1.1.1.so", "#{CREW_DEST_LIB_PREFIX}/libtdbcpostgres1.1.1.so"
+      FileUtils.ln_s "#{CREW_LIB_PREFIX}/tdbcodbc1.1.1/libtdbcodbc1.1.1.so",
+                     "#{CREW_DEST_LIB_PREFIX}/libtdbcodbc1.1.1.so"
+      FileUtils.ln_s "#{CREW_LIB_PREFIX}/sqlite3.30.1.2/libsqlite3.30.1.2.so",
+                     "#{CREW_DEST_LIB_PREFIX}/libsqlite3.30.1.2.so"
```
uberhacker, line 48: I'm thinking this symlink should go in the `sqlite` package.  Once sqlite is upgraded, then this will break.  Not an ideal potential situation.

packages/tk.rb
```diff
@@ -27,21 +28,27 @@ class Tk < Package
...
-               "--with-tcl=#{CREW_LIB_PREFIX}",
-               '--enable-threads',
-               '--disable-64bit'
+        system "env CFLAGS='-flto=auto -pipe -fuse-ld=gold' \
+          CXXFLAGS='-flto=auto -pipe -fuse-ld=gold' \
+          LDFLAGS='-flto=auto' \
+          ./configure \
+          --prefix=#{CREW_PREFIX} \
+          --libdir=#{CREW_LIB_PREFIX} \
+          --mandir=#{CREW_PREFIX}/share/man \
```
uberhacker, line 48: Can the 3 previous lines be replaced by `#{CREW_OPTIONS}`?

packages/python3.rb
satmandu: Upon install, poip3 and pip are tiny script files, not symlinks.

Example from ubuntu:
```
#!/usr/bin/python3
# EASY-INSTALL-ENTRY-SCRIPT: 'pip==20.1.1','console_scripts','pip'
__requires__ = 'pip==20.1.1'
import re
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(
        load_entry_point('pip==20.1.1', 'console_scripts', 'pip')()
    )
```

Seems crazy to use a symlink for those, no?

packages/tcl.rb
satmandu: Agreed. All of these symlinks are preexiting. Do we need any of them?

packages/python3.rb
uberhacker: Well, the idea is to not have duplicate files.  A symlink will still work, no?

packages/tcl.rb
uberhacker: I think the idea is to make sure they are seen since everything is in a subfolder.  Otherwise, the specific paths to the shared object files need added to the library path.

packages/tk.rb
satmandu: Yup. That works fine.

packages/python3.rb
satmandu: I assume a symlink will still work? Since it is different from a normal python install though we can't be sure if something might break...

PR #5242, opened 2021-02-22
Use mawk instead of gawk

saltedcoffii:
So this might seem like a really weird change, but it has its place.
`gawk` is the GNU awk interpreter. It has features that POSIX awk doesn't have like network access, however these aren't needed by any configure script I've seen.
`mawk` is a POSIX compatible awk interpreter from the debian project. It runs, in my tests, about twice as fast as `gawk` and works in every configure script in place of `gawk`. I've been testing this for about two weeks now and haven't run into a problem.

If you're worried about packages requiring `gawk` specifically, I only found one, `translate_shell`, and it needs `gawk` for execution as its binaries are written in awk. Configure scripts and makefiles run way faster with `mawk`.

Should this be merged, I'll update `mawk`'s package file with a symlink linking `mawk` with `awk` for compatibility purposes with scripts as `gawk` is currently linked, and update `translate_shell`'s dependency list as well.

### DIFF
--- a/packages/buildessential.rb
+++ b/packages/buildessential.rb
@@ -46,7 +46,7 @@ class Buildessential < Package
   depends_on 'util_macros'
   depends_on 'gettext'
   depends_on 'wget' # in some cases, patches might be required and can be downloaded using wget
-  depends_on 'gawk'
+  depends_on 'mawk'
 
   # compression utilities
   depends_on 'lzip'
--- a/packages/buildessential.rb
+++ b/packages/buildessential.rb
@@ -3,7 +3,7 @@ require 'package'
 class Buildessential < Package
   description 'A collection of tools essential to compile and build software.'
   homepage ''
-  version '1.8'
+  version '1.8-1'
   compatibility 'all'
 
   is_fake
--- a/packages/buildessential.rb
+++ b/packages/buildessential.rb
@@ -3,7 +3,7 @@ require 'package'
 class Buildessential < Package
   description 'A collection of tools essential to compile and build software.'
   homepage ''
-  version '1.8-1'
+  version '1.9'
   compatibility 'all'
 
   is_fake
--- a/tools/core_packages.txt
+++ b/tools/core_packages.txt
@@ -16,7 +16,6 @@ doxygen
 expat
 filecmd
 flex
-gawk
 gcc10
 gcc_tools
 gdbm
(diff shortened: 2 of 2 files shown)


uberhacker:
I'm fine with this change unless somebody else objects.  We'll need to bump the version and update `tools/core_packages.txt`.

satmandu:
Does any distribution use mawk in lieu of gawk? Because if not, we're taking on a lot of technical debt if something subtly breaks?

uberhacker:
I can attest Debian-based distros symlink awk to gawk by default.

saltedcoffii:
Weird, I'm running Debian 10 in crostini and it symlinks to mawk.

saltedcoffii:
So I just debootstrapped a debian chroot with crouton (`sudo crouton -r bullseye -t core`) and did this:
```
$ ls -l $(which awk)
lrwxrwxrwx. 1 root root 21 Feb 22 17:20 /usr/bin/awk -> /etc/alternatives/awk
$ ls -l /etc/alternatives/awk
lrwxrwxrwx. 1 root root 13 Feb 22 17:20 /etc/alternatives/awk -> /usr/bin/mawk
```
Ubuntu (`sudo crouton -r focal -t core`) did the exact same thing.
This default setting can be changed with dpkg's `update-alternatives`. I wanted to see if a ubuntu with gnome or something like that installed would change that symlink, but no, it still did the same thing.

uberhacker:
Crouton was the reason I gravitated to this project.  I imagine they bastardized everything in Crostini.  Google is notorious for wanting to change everything.  Not always successfully, I might add. :)

saltedcoffii:
Just started up debian in a VM and it said the exact same thing.

satmandu:
Ubuntu appears to be using gawk as the default, and debian appears to be using mawk.

What would be nice would be for the debian ```/etc/alternatives``` setup to be able to switch between different programs for things like awk, gcc, ld, etc....

saltedcoffii:
> Ubuntu appears to be using gawk as the default, and debian appears to be using mawk.
> 
> What would be nice would be for the debian `/etc/alternatives` setup to be able to switch between different programs for things like awk, gcc, ld, etc....

try `update-alternatives`

saltedcoffii:
But aside from all that, since debian does use mawk as its default distro, we don't have any objections?

uberhacker:
My only objection is we need to bump the version and update `tools/core_packages.txt`. :)

saltedcoffii:
I think this is ready to merge :)

PR #5243, opened 2021-02-22
Add pythons to buildessential

satmandu:
- Establish pythons as a core dependency of buildessential

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l


uberhacker:
Good call.

PR #5247, opened 2021-02-23
Add deb file support to chromebrew

saltedcoffii:
This would primarily be helpful for proprietary software.

### DIFF
--- a/crew
+++ b/crew
@@ -557,6 +557,10 @@ def unpack (meta)
       puts "Unpacking archive using 'tar', this may take a while..."
       FileUtils.mkdir extract_dir unless Dir.exist?(extract_dir)
       system "tar x#{@verbose}f #{meta[:filename]} -C #{extract_dir}"
+    when /\.deb$/i
+      puts "Unpacking archive using 'ar', this may take a while..."
+      FileUtils.mkdir extract_dir unless Dir.exist?(extract_dir)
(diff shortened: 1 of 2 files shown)


crew
```diff
@@ -557,6 +557,10 @@ def unpack (meta)
       puts "Unpacking archive using 'tar', this may take a while..."
       FileUtils.mkdir extract_dir unless Dir.exist?(extract_dir)
       system "tar x#{@verbose}f #{meta[:filename]} -C #{extract_dir}"
+    when /\.deb$/i
+      puts "Unpacking archive using 'ar', this may take a while..."
+      FileUtils.mkdir extract_dir unless Dir.exist?(extract_dir)
+      system "ar -p #{meta[:filename]} data.tar.xz | xz -dc#{@verbose} | tar x#{@verbose} -C #{extract_dir}"
```
uberhacker, line 563: Does this do the same thing as alien?  I'm not sure we want to extract deb archives because of the directory layout.  I imagine a lot of archives will try to extract out to read-only directories which will not generate pretty output.  Have you tested this with a few deb packages yet?

saltedcoffii: I have. So basically what this does is it extracts the data part of the tarfile. The user can then move everything in that data hierarchy around in the `def self.build` phase to work for chormebrew.

uberhacker: We'll need to bump the crew version in `lib/const.rb`.

supechicken:
Okay, does there anything different with #4709 :)

saltedcoffii:
no... looks like uberhacker changed their mind between that PR and now.

PR #5250, opened 2021-02-23
[New Package] libglvnd

saltedcoffii:


packages/libglvnd.rb
```diff
@@ -0,0 +1,37 @@
+require 'package'
+
+class Libglvnd < Package
+  description 'The GL Vendor-Neutral Dispatch library'
+  homepage 'https://gitlab.freedesktop.org/glvnd/libglvnd'
+  version '1.2.0'
```
uberhacker, line 6: Version doesn't make source?

PR #5259, opened 2021-02-23
Add flto to packages to be built

satmandu:
- Just added flto to these packages
- These will have to be built after mono is built on actual hardware...

### DIFF
--- a/packages/libgdiplus.rb
+++ b/packages/libgdiplus.rb
@@ -20,12 +20,14 @@ class Libgdiplus < Package
   depends_on 'imake' => :build
 
   def self.build
-    system "./configure #{CREW_OPTIONS} \
-            --with-libexif \
-            --with-libtiff \
-            --with-jpeg \
-            --with-libgif \
-            --with-x11"
+    system  "CFLAGS='-flto=auto' CXXFLAGS='-flto=auto'
+      LDFLAGS='-flto=auto' \
+      ./configure #{CREW_OPTIONS} \
+      --with-libexif \
+      --with-libtiff \
+      --with-jpeg \
+      --with-libgif \
+      --with-x11"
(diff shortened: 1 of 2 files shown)


packages/libwebp.rb
```diff
@@ -17,16 +17,18 @@ class Libwebp < Package
...
+      ./configure #{CREW_OPTIONS} \
+      --enable-libwebpmux \
+      --enable-libwebpdemux \
+      --enable-libwebpdecoder \
+      --enable-libwebpextras \
+      --enable-sdl \
+      --enable-tiff \
+      --enable-jpeg \
+      --enable-png \
+      --enable-gif"
```
uberhacker, line 31: No binaries?

packages/libgdiplus.rb
```diff
@@ -20,12 +20,14 @@ class Libgdiplus < Package
...
-            --with-libgif \
-            --with-x11"
+    system  "CFLAGS='-flto=auto' CXXFLAGS='-flto=auto'
+      LDFLAGS='-flto=auto' \
+      ./configure #{CREW_OPTIONS} \
+      --with-libexif \
+      --with-libtiff \
+      --with-jpeg \
+      --with-libgif \
+      --with-x11"
```
uberhacker, line 30: No binaries?

packages/libwebp.rb
satmandu: These are dependent upon mono, and I can't build mono due to the mono build requiring NOT being built in a container.

packages/libgdiplus.rb
satmandu: These are dependent upon mono, and I can't build mono due to the mono build requiring NOT being built in a container.

PR #5263, opened 2021-02-24
[New Package] benchmark

saltedcoffii:


packages/benchmark.rb
```diff
@@ -0,0 +1,52 @@
...
+    system 'patch -p 1 -i limitsh.patch'
+  end
+  
+  def self.prebuild
+    system "git clone git://github.com/google/googletest.git -b release-1.10.0 --depth 1" # Required for build, won't interfere with the gtest package
+  end
+    
+  def self.build
+    Dir.mkdir "builddir"
+    Dir.chdir "builddir" do
```
uberhacker, line 34: Should this be?
```
Dir.chdir 'googletest' do
```
Where do you cd to the cloned repo?

saltedcoffii: I don't cd into the cloned repo, it just goes into the source like a submodule and it's compiled for tests.

uberhacker: So it clones to a folder the extracted tarball source can read already?

saltedcoffii: Yes

PR #5266, opened 2021-02-24
[New Package] civetweb

saltedcoffii:


packages/civetweb.rb
```diff
@@ -0,0 +1,36 @@
+require 'package'
+
+class Civetweb < Package
+  description 'Embedded C/C++ web server'
+  homepage 'https://github.com/civetweb/civetweb/'
+  version '0.12.1'
```
uberhacker, line 6: Version doesn't match source url?

packages/civetweb.rb
```diff
@@ -0,0 +1,36 @@
...
+  compatibility 'all'
+  source_url 'https://github.com/civetweb/civetweb/archive/v1.13.tar.gz'
+  source_sha256 'a7ccc76c2f1b5f4e8d855eb328ed542f8fe3b882a6da868781799a98f4acdedc'
+
+  depends_on 'openssl'
```
uberhacker, line 11: Core package so no need to include.

PR #5274, opened 2021-02-24
fontconfig: keep postinstall failure from breaking other builds

satmandu:
- Despite fontconfig having a graphite dependency, if fontconfig is installed before graphite, the post-install will fail.
- This breaks other installs, so keep that post-install failure from breaking builds.
- The long term solution would be to unbreak dependency installs.

Works properly:
- [x] x86_64'
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/fontconfig.rb
+++ b/packages/fontconfig.rb
@@ -77,6 +77,8 @@ class Fontconfig < Package
       puts 'To complete the installation, execute the following:'.orange
       puts 'source ~/.bashrc'.orange
     end
-    system "env FONTCONFIG_PATH=#{CREW_PREFIX}/etc/fonts fc-cache -fv"
+    # The following postinstall fails if graphite isn't installed when fontconfig
+    # is being installed.
+    system "env FONTCONFIG_PATH=#{CREW_PREFIX}/etc/fonts fc-cache -fv || true"
   end
 end


packages/fontconfig.rb
```diff
@@ -77,6 +77,8 @@ def self.postinstall
       puts 'To complete the installation, execute the following:'.orange
       puts 'source ~/.bashrc'.orange
     end
-    system "env FONTCONFIG_PATH=#{CREW_PREFIX}/etc/fonts fc-cache -fv"
+    # The following postinstall fails if graphite isn't installed when fontconfig
+    # is being installed.
+    system "env FONTCONFIG_PATH=#{CREW_PREFIX}/etc/fonts fc-cache -fv || true"
```
uberhacker, line 82: Does this mean graphite should be a fontconfig dependency?

uberhacker: Wait...graphite is already a dependency?

satmandu: During a build of a package which requires fontconfig, a bunch of depends are getting installed, but there's currently no guarantee that fontconfig's deps will get installed BEFORE fontconfig... 

In this case, fontconfig install started before graphite. postinstall fails because no graphite... the entire build fails... never getting to graphite.

This gets around that. Dependency postinstall errors are currently fatal during builds.

uberhacker: Probably a better fix is to make sure all the dependencies are installed before the build.

satmandu: A better fix is needed. This is needed to get builds working properly. 😐

PR #5284, opened 2021-02-26
Add torbrowser package

uberhacker:
See https://www.torproject.org/about/history/.  Tested on x86_64.

### DIFF
--- a/packages/torbrowser.rb
+++ b/packages/torbrowser.rb
@@ -1,0 +1,39 @@
+require 'package'
+
+class Torbrowser < Package
+  description "'The Onion Router' browser"
+  homepage 'https://www.torproject.org/'
+  compatibility 'x86_64'
+  @_ver = '10.0.12'
+  version @_ver
+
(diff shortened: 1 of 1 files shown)


satmandu:
> See https://www.torproject.org/about/history/.  Tested on x86_64.  No sound. :(

For sound something like this?

https://gitlab.torproject.org/legacy/trac/-/issues/29360

satmandu:
As suggested in that thread to fix audio with pulseaudio:

```
$ cp -pf $HOME/.esd_auth $HOME/.tor-browser/INSTALL/Browser
$ cp -pfr $HOME/.config/pulse/ $HOME/.tor-browser/INSTALL/Browser
```

supechicken:
Try to install `cras`? By the way the sound works perfectly for me :)

uberhacker:
@supechicken is correct.  Sound is working.  Not sure why but I didn't need any special fixes to get it working.  Probably user error. :)

PR #5301, opened 2021-03-01
crew: Allow to use external prebuilt binary (like .deb and .rpm)

supechicken:
Package need to execute `is_external` to use external binary

Tested on `x86_64`

### DIFF
--- a/crew
+++ b/crew
@@ -799,7 +799,7 @@ def install
     meta = download
     target_dir = unpack meta
     if meta[:source] == true
-      abort "You don't have a working C compiler. Run `crew install buildessential` to get one and try again.".lightred unless @external and system('gcc --version')
+      abort "You don't have a working C compiler. Run `crew install buildessential` to get one and try again.".lightred unless (@external or system('gcc --version'))
 
       # build from source and place binaries at CREW_DEST_DIR
       # CREW_DEST_DIR contains usr/local/... hierarchy
--- a/crew
+++ b/crew
@@ -799,7 +799,9 @@ def install
     meta = download
     target_dir = unpack meta
     if meta[:source] == true
-      (abort "You don't have a working C compiler. Run `crew install buildessential` to get one and try again.".lightred unless @external) unless system('gcc --version')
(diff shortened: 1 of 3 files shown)


crew
```diff
@@ -495,11 +495,12 @@ end
 
 def download
   url = @pkg.get_url(@device[:architecture])
+  @external = @pkg.is_external?
```
uberhacker, line 498: If you want to use a property, you will need to add it to `lib/package.rb` also.  However, if you do this then every package needs this property set to a default value if not set directly.

uberhacker:
I'm beginning to regret my decision to allow deb package support in crew. :)

crew
supechicken: Like `is_fake`, the new `is_external` function in `package.rb` will return `nil` if the variable is not set

supechicken: (which equal to `false`)

uberhacker:
Maybe I'm off base here but doesn't this issue go away if we use source_url instead of binary_url for `.deb` archives?

supechicken:
> Maybe I'm off base here but doesn't this issue go away if we use source_url instead of binary_url for `.deb` archives?

I think the package archives like `deb` are more like `prebuilt binary` instead of a `source` (although they are not built and packaged by us) The another reason that using `binary_url` is to skip the `rename *.la to *.la_tmp` part and won't show the `gcc` version (not necessary, but looks better:) )

uberhacker:
> > Maybe I'm off base here but doesn't this issue go away if we use source_url instead of binary_url for `.deb` archives?
> 
> I think the package archives like `deb` are more like `prebuilt binary` instead of a `source` (although they are not built and packaged by us) The another reason that using `binary_url` is to skip the `rename *.la to *.la_tmp` part and won't show the `gcc` version (not necessary, but looks better:) )

I would agree with that statement if deb packages were native to ChromeOS.  However, since they need to be reconfigured to work in Chromebrew, I tend to think they qualify as source instead of binary.

supechicken:
Close due to `source_url` is better

PR #5302, opened 2021-03-01
opera => 74.0.3911.160

supechicken:
depends_on #5301 

Tested on `x86_64`

### DIFF
--- a/packages/opera.rb
+++ b/packages/opera.rb
@@ -7,13 +7,6 @@ class Opera < Package
   compatibility 'x86_64'
   version '74.0.3911.160'
   
-  binary_url ({
-    x86_64: "https://get.geo.opera.com/pub/opera/desktop/74.0.3911.160/linux/opera-stable_#{version}_amd64.deb"
-  })
-  binary_sha256 ({
-    x86_64: `curl -Ls https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb.sha256sum`.chomp
-  })
-  
   is_external
   
   if ARCH == 'x86_64'
@@ -23,6 +16,8 @@ class Opera < Package
     depends_on 'graphite'
     depends_on 'cras'
     depends_on 'sommelier'
+    source_url "https://get.geo.opera.com/pub/opera/desktop/74.0.3911.160/linux/opera-stable_#{version}_amd64.deb"
+    source_sha256 `curl -Ls https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb.sha256sum`.chomp
   end
 
   def self.install
--- a/packages/opera.rb
+++ b/packages/opera.rb
@@ -30,16 +30,14 @@ class Opera < Package
     ENV['CREW_NOT_STRIP'] = '1'
     
     # since opera put the executable to library, we need to link it to CREW_PREFIX
-    # remove old symlink which link to /usr/lib/x86_64-linux-gnu/opera/opera
-    FileUtils.rm_rf 'bin/opera'
-    FileUtils.ln_s "#{CREW_LIB_PREFIX}/opera/opera", 'bin/opera'
+    FileUtils.ln_sf "#{CREW_LIB_PREFIX}/opera/opera", 'bin/opera'
       
     # place all stuff to lib64 instead of lib
-    FileUtils.mv 'lib/x86_64-linux-gnu/', 'lib64/'
+    FileUtils.mv 'lib/x86_64-linux-gnu/', 'share/'
     FileUtils.rm_rf 'lib/'
     
     FileUtils.mkdir_p CREW_DEST_PREFIX
     FileUtils.mv Dir.glob('*'), CREW_DEST_PREFIX
-    FileUtils.ln_s "#{CREW_PREFIX}/bin/opera", "#{CREW_DEST_PREFIX}/bin/x-www-browser" unless File.exist?("#{CREW_PREFIX}/bin/x-www-browser")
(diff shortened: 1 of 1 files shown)


packages/opera.rb
```diff
@@ -3,12 +3,28 @@
...
+  end
+  
+  binary_url ({
+      i686: "https://deb.opera.com/opera-stable/pool/non-free/o/opera-stable/opera-stable_#{version}_i386.deb",
+    x86_64: "https://get.geo.opera.com/pub/opera/desktop/74.0.3911.160/linux/opera-stable_#{version}_amd64.deb"
+  })
+  binary_sha256 ({
+      i686: '0fd8497f914efc47eb1ef1c7ce052487e3ca0c4c5474f01ad7e721f6ea08fea8',
+    x86_64: `curl -Ls https://get.geo.opera.com/pub/opera/desktop/#{version}/linux/opera-stable_#{version}_amd64.deb.sha256sum`.chomp
+  })
```
uberhacker, line 23: I was unable to get sommelier working on my i686 Chromebook so not sure if this version is needed.

packages/opera.rb
```diff
@@ -17,21 +33,23 @@ class Opera < Package
...
-      FileUtils.mv Dir.glob('usr/*'), CREW_DEST_PREFIX
-      FileUtils.ln_s "#{CREW_PREFIX}/bin/opera", "#{CREW_DEST_PREFIX}/bin/x-www-browser" unless File.exist?("#{CREW_PREFIX}/bin/x-www-browser")
+    # llvm-strip doesn't works with opera
+    ENV['CREW_NOT_STRIP'] = '1'
+    
+    if ARCH == 'x86_64' then
+      # since opera put the executable to library, we need to link it to CREW_PREFIX
+      # remove old symlink which link to /usr/lib/x86_64-linux-gnu/opera/opera
+      FileUtils.rm_rf 'bin/opera'
+      FileUtils.ln_s "#{CREW_LIB_PREFIX}/opera/opera", 'bin/opera'
```
uberhacker, line 44: Would it make sense to move everything from `usr/local/lib/opera` to `#{CREW_DEST_PREFIX}/share/opera`?

packages/opera.rb
```diff
@@ -17,21 +33,23 @@ class Opera < Package
...
+      
+      # place all stuff to lib64 instead of lib
+      FileUtils.mv 'lib/x86_64-linux-gnu/', 'lib64/'
```
uberhacker, line 47: How about `FileUtils.mv Dir.glob('lib/x86_64-linux-gnu/*'), 'lib64/'`?

packages/opera.rb
supechicken: Will `opt` better?

uberhacker: The `opt` location is the alternate way.  I try to keep all packages organized in `share` so we don't have apps in so many different locations.

packages/opera.rb
uberhacker: source_url, not binary_url?

packages/opera.rb
supechicken666: Change made

packages/opera.rb
satmandu: Sommelier might have a dependency on a newer version of the exo wayland server. Do wayland apps work on i686?

packages/opera.rb
```diff
@@ -3,35 +3,36 @@
 class Opera < Package
   description "Opera isn't just a browser. It's an awesome browser."
   homepage 'https://www.opera.com/'
-  version '74.0.3911.154'
+  version '74.0.3911.160'
   compatibility 'x86_64'
+  version '74.0.3911.160'
+  
+  is_external
```
uberhacker, line 10: Can we remove this and see how it works?

packages/opera.rb
```diff
@@ -3,35 +3,36 @@
 class Opera < Package
   description "Opera isn't just a browser. It's an awesome browser."
   homepage 'https://www.opera.com/'
-  version '74.0.3911.154'
+  version '74.0.3911.160'
   compatibility 'x86_64'
+  version '74.0.3911.160'
```
uberhacker, line 8: 2 version lines?

packages/opera.rb
```diff
@@ -3,35 +3,36 @@
...
-    end
+    # llvm-strip doesn't works with opera
+    ENV['CREW_NOT_STRIP'] = '1'
+    
+    # since opera put the executable to library, we need to link it to CREW_PREFIX
+    FileUtils.ln_sf "#{CREW_LIB_PREFIX}/opera/opera", 'bin/opera'
+      
+    # place all stuff to lib64 instead of lib
+    FileUtils.mv 'lib/x86_64-linux-gnu/', 'share/'
+    FileUtils.rm_rf 'lib/'
```
uberhacker, line 32: This step is unnecessary because the whole working directory will be removed after the build anyway.

packages/opera.rb
```diff
@@ -3,35 +3,36 @@
...
+    
+    FileUtils.mkdir_p CREW_DEST_PREFIX
+    FileUtils.mv Dir.glob('*'), CREW_DEST_PREFIX
+    FileUtils.ln_s "#{CREW_PREFIX}/share/opera/opera", "#{CREW_DEST_PREFIX}/bin/x-www-browser" unless File.exist?("#{CREW_PREFIX}/bin/x-www-browser")
```
uberhacker, line 36: Should this be an optional `def self.postinstall` step with a prompt?  Something like _"Would you like to make Opera your default browser? (Y/n): "_

uberhacker:
@supechicken: What is the status of this?

supechicken:
> @supechicken: What is the status of this?

Just updated :)

PR #5304, opened 2021-03-01
chrome => 88.0.4324.182

supechicken666:
depends_on #5301

Tested on `x86_64`


packages/chrome.rb
```diff
@@ -4,15 +4,20 @@ class Chrome < Package
...
+  @_ver = '88.0.4324.182'
   @_deb = "google-chrome-stable_#{@_ver}-1_amd64.deb"
   version @_ver
+  
+  binary_url ({
+    x86_64: "https://dl.google.com/linux/deb/pool/main/g/google-chrome-stable/#{@_deb}"
+  })
+  binary_sha256 ({
+    x86_64: 'ab00e9412f5f20e30c7db5dc987473248f4adf9ebf2c3f928ef62e1ffb104fe6'
+  })
```
uberhacker, line 16: Technically, this would be considered a source since we have to move files around to get it working on ChromeOS.  A true binary would include Chromebrew metadata and simply be installed by extracting all the files.

packages/chrome.rb
```diff
@@ -21,21 +26,17 @@ class Chrome < Package
...
-    end
+    ENV['CREW_NOT_STRIP'] = '1'
+    FileUtils.mkdir_p CREW_DEST_PREFIX
+    FileUtils.mkdir_p 'bin'
+    
+    FileUtils.mv 'usr/share', 'share'
+    FileUtils.rm_rf 'usr'
+    
+    FileUtils.ln_s "#{CREW_PREFIX}/opt/google/chrome/google-chrome", 'bin/google-chrome-stable'
+    FileUtils.ln_s "#{CREW_PREFIX}/opt/google/chrome/google-chrome", 'bin/google-chrome'
```
uberhacker, line 38: How about `FileUtils.mv 'opt/google/chrome/google-chrome', 'share'` and then:
```
FileUtils.ln_s "#{CREW_PREFIX}/share/google-chrome", 'bin/google-chrome-stable'
FileUtils.ln_s "#{CREW_PREFIX}/share/google-chrome", 'bin/google-chrome'
```
This is a better location than opt.

packages/google_chrome.rb
```diff
@@ -0,0 +1,13 @@
...
+class Google_chrome < Package
+  description 'Google Chrome is a fast, easy to use, and secure web browser.'
+  homepage 'https://www.google.com/chrome'
+  compatibility 'x86_64'
+  version Chrome.version
+  
+  is_fake
+  
+  depends_on 'chrome'
+end
```
uberhacker, line 13: Redundant package?

packages/chrome.rb
supechicken: I think `opt` is the best location for chromium-based browsers as programs there usually live inside a single folder that contains all of their data, libraries, etc.

packages/google_chrome.rb
supechicken: Removed

packages/chrome.rb
uberhacker: I think `share` is better than `opt` for consistency with other packages.  I don't want package files stored all over the place in different locations.  Wouldn't it be nice if you could cd to the share directory and list all the installed apps in one location?

supechicken666: Fair enough :)

supechicken666: Change made

PR #5305, opened 2021-03-01
skype => 8.69.76.76

supechicken666:
depends_on #5301

Tested on `x86_64`

### DIFF
--- a/packages/skype.rb
+++ b/packages/skype.rb
@@ -9,15 +9,10 @@ class Skype < Package
   if ARCH == 'x86_64' then
     depends_on 'gtk3'
     depends_on 'sommelier'
+    source_url "https://repo.skype.com/deb/pool/main/s/skypeforlinux/skypeforlinux_#{version}_amd64.deb"
+    source_sha256 '30182e5d940993061956a7fd188ca56fbf3bc0852f7c6e8e6a46d44460db63c4'
   end
-
-  binary_url ({
-     x86_64: "https://repo.skype.com/deb/pool/main/s/skypeforlinux/skypeforlinux_#{version}_amd64.deb"
-  })
-  binary_sha256 ({
-     x86_64: '30182e5d940993061956a7fd188ca56fbf3bc0852f7c6e8e6a46d44460db63c4'
-  })
-    
+  
   is_external
 
   def self.preinstall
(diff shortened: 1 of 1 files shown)


packages/skype.rb
```diff
@@ -3,36 +3,26 @@
...
-    depends_on 'alien' => :build
+  
+  if ARCH == 'x86_64' then
     depends_on 'gtk3'
     depends_on 'sommelier'
   end
 
   binary_url ({
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/skype-8.68.76.35-chromeos-x86_64.tar.xz',
+     x86_64: "https://repo.skype.com/deb/pool/main/s/skypeforlinux/skypeforlinux_#{version}_amd64.deb"
```
uberhacker, line 15: source_url, not binary_url?

packages/skype.rb
```diff
@@ -3,36 +3,26 @@
...
-
-  def self.build
-    system 'curl https://repo.skype.com/deb/pool/main/s/skypeforlinux/skypeforlinux_8.68.76.35_amd64.deb -#o skype.deb'
-    puts "Archive downloaded".lightgreen
-    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('skype.deb') ) == '701449b86ff1c10ee800cc017033c7c2e9b31023fcc1a9058590babf070ce3bb'
-    system 'alien -tc skype.deb'
-    system "tar xvf skypeforlinux-#{version}.tgz"
-  end
+    
+  is_external
```
uberhacker, line 21: Not sure about this.  What advantage does it provide?

packages/skype.rb
```diff
@@ -3,36 +3,26 @@
...
 
   def self.install
-    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}"
-    FileUtils.mv 'usr/bin', "#{CREW_DEST_PREFIX}"
-    FileUtils.mv 'usr/share', "#{CREW_DEST_PREFIX}"
-    FileUtils.ln_s "#{CREW_PREFIX}/bin/skypeforlinux", "#{CREW_DEST_PREFIX}/bin/skype"
+    FileUtils.mkdir_p CREW_DEST_PREFIX
+    FileUtils.ln_s "#{CREW_PREFIX}/bin/skypeforlinux", 'usr/bin/skype'
```
uberhacker, line 25: Probably should mv before ln_s for clarity.

supechicken666: Moved `ln` to `preinstall` part

packages/skype.rb
supechicken666: It provides an easier way to add different arch of deb (if arch supported) :)

supechicken:
> Can we remove this and see how it works?

Removed

PR #5306, opened 2021-03-01
Mandb -> 2.9.4

satmandu:
- removes libiconv dependency
- add PAGER variable to .bashrc 
- update libpipeline dep too.

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/libpipeline.rb
+++ b/packages/libpipeline.rb
@@ -3,9 +3,9 @@ require 'package'
 class Libpipeline < Package
   description 'libpipeline is a C library for manipulating pipelines of subprocesses in a flexible and convenient way.'
   homepage 'http://libpipeline.nongnu.org/'
-  compatibility 'all'
   @_ver = '1.5.3'
   version @_ver
+  compatibility 'all'
   source_url "https://mirror.csclub.uwaterloo.ca/nongnu/libpipeline/libpipeline-#{@_ver}.tar.gz"
   source_sha256 '5dbf08faf50fad853754293e57fd4e6c69bb8e486f176596d682c67e02a0adb0'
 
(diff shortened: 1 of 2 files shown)


packages/libpipeline.rb
```diff
@@ -3,35 +3,40 @@
 class Libpipeline < Package
   description 'libpipeline is a C library for manipulating pipelines of subprocesses in a flexible and convenient way.'
   homepage 'http://libpipeline.nongnu.org/'
-
   compatibility 'all'
```
uberhacker, line 6: Can we put this line below the version?

packages/mandb.rb
```diff
@@ -48,14 +62,23 @@ def self.install
...
       'tools/chconfig'
-    ].each { |file|
+    ].each do |file|
       system "sed -i 's,/var/cache/man,#{CREW_PREFIX}/cache/man,g' #{file}"
-    }
+    end
     system "sed -i 's,/usr/share/man,#{CREW_PREFIX}/share/man,g' tools/chconfig"
     system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    system "sed -i 's,/usr/share/man,#{CREW_PREFIX}/share/man,g' #{CREW_DEST_PREFIX}/etc/man_db.conf"
+    system "sed -i 's,/usr/man,#{CREW_PREFIX}/share/man,g' #{CREW_DEST_PREFIX}/etc/man_db.conf"
```
uberhacker, line 71: Can the sed lines be moved into `def self.patch`?

satmandu: done

PR #5311, opened 2021-03-02
Fix location of files which go in dbus-1/system.d directory in multiple package

satmandu:
- also updated deps as I went...

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/flatpak.rb
+++ b/packages/flatpak.rb
@@ -4,22 +4,22 @@ class Flatpak < Package
   description 'Flatpak is a system for building, distributing, and running sandboxed desktop applications on Linux.'
   homepage 'https://flatpak.org'
   @_ver = '1.10.1'
-  version @_ver
+  version "#{@_ver}-1"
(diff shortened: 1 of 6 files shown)


packages/flatpak.rb
```diff
@@ -43,12 +43,16 @@ def self.patch
     Digest::SHA256.hexdigest(File.read(patch_filename)) == patch_sha256
     puts 'patch downloaded'.lightgreen
     system 'patch  -p 1 --forward < patch || true'
+    # Source has libglnx repo as submodule
+    FileUtils.rm_rf('libglnx')
+    system 'git clone --depth=1 https://gitlab.gnome.org/GNOME/libglnx.git libglnx '
```
uberhacker, line 48: Should we be checking out a tagged release?

satmandu: No tags for 3 years... but I just have it checking out a specific commit now.

PR #5320, opened 2021-03-04
torbrowser => 10.0.13, use official sha256 file

supechicken:
Tested on `x86_64`

### DIFF
--- a/packages/torbrowser.rb
+++ b/packages/torbrowser.rb
@@ -4,22 +4,23 @@ class Torbrowser < Package
   description "'The Onion Router' browser"
   homepage 'https://www.torproject.org/'
   compatibility 'x86_64'
-  @_ver = '10.0.12'
+  @_ver = '10.0.13'
   version @_ver
 
   case ARCH
   when 'x86_64'
-    source_url "https://www.torproject.org/dist/torbrowser/#{@_ver}/tor-browser-linux64-#{@_ver}_en-US.tar.xz"
-    source_sha256 `curl -Ls #{@source_url} | sha256sum | cut -d ' ' -f1`.chomp
+    @_url = "https://www.torproject.org/dist/torbrowser/#{@_ver}/"
+    @_name = "tor-browser-linux64-#{@_ver}_en-US.tar.xz"
+    source_url @_url + @_name
+    source_sha256 `curl -Ls #{@_url}/sha256sums-signed-build.txt | grep '#{@_name}'`.chomp
     depends_on 'gtk3'
     depends_on 'sommelier'
   end
 
   def self.build
     system "cat <<'EOF'> tor
-#!/bin/bash
-cd #{CREW_PREFIX}/share
-./start-tor-browser.desktop \"\$@\"
+#!/bin/sh
+#{CREW_PREFIX}/share/start-tor-browser.desktop \"\$@\"
 EOF"
   end
 
--- a/packages/torbrowser.rb
+++ b/packages/torbrowser.rb
@@ -43,7 +43,7 @@ class Torbrowser < Package
         FileUtils.ln_sf 'tor', 'x-www-browser'
       end
     else
-      puts 'No change have been made.'
+      puts 'No change has been made.'
       puts
     end
     
@@ -54,7 +54,7 @@ class Torbrowser < Package
   
   def self.remove
     Dir.chdir("#{CREW_PREFIX}/bin") do
-      FileUtils.rm_rf 'x-www-browser' if File.realpath('x-www-browser') == "#{CREW_PREFIX}/bin/tor"
+      FileUtils.rm 'x-www-browser' if File.realpath('x-www-browser') == "#{CREW_PREFIX}/bin/tor"
     end    
   end
 end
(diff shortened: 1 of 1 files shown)


packages/torbrowser.rb
```diff
@@ -4,29 +4,32 @@ class Torbrowser < Package
...
     depends_on 'sommelier'
   end
 
-  def self.build
-    system "cat <<'EOF'> tor
-#!/bin/bash
-cd #{CREW_PREFIX}/share
-./start-tor-browser.desktop \"\$@\"
-EOF"
+  def self.patch
```
uberhacker, line 20: `def self.build` is actually a better section for this.

packages/torbrowser.rb
```diff
@@ -4,29 +4,32 @@ class Torbrowser < Package
...
+      ./start-tor-browser.desktop "$@"
+    EOF
+    File.write('tor', @tor)
   end
 
   def self.install
     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share"
-    FileUtils.mv 'Browser/', "#{CREW_DEST_PREFIX}/share"
+    FileUtils.mv 'Browser/', "#{CREW_DEST_PREFIX}/share/Browser"
```
uberhacker, line 32: This will create `/usr/local/share/Browser/Browser`.  Does that make sense?

packages/torbrowser.rb
```diff
@@ -4,29 +4,32 @@ class Torbrowser < Package
...
   version @_ver
 
   case ARCH
   when 'x86_64'
-    source_url "https://www.torproject.org/dist/torbrowser/#{@_ver}/tor-browser-linux64-#{@_ver}_en-US.tar.xz"
-    source_sha256 `curl -Ls #{@source_url} | sha256sum | cut -d ' ' -f1`.chomp
+    @_url = "https://www.torproject.org/dist/torbrowser/#{@_ver}/"
+    @_name = "tor-browser-linux64-#{@_ver}_en-US.tar.xz"
+    source_url @_url + @_name
+    source_sha256 `curl -Ls #{@_url}/sha256sums-signed-build.txt | grep '#{@_name}' | cut -d' ' -f1`.chomp
```
uberhacker, line 15: The earlier way was simple but this makes more sense since it's a smaller file.

satmandu (approved):

packages/torbrowser.rb
uberhacker: Since `@_url` already has a trailing slash?
```
source_sha256 `curl -Ls #{@_url}sha256sums-signed-build.txt | grep '#{@_name}' | cut -d' ' -f1`.chomp
```

packages/torbrowser.rb
uberhacker: Actually, my guess it will error out and create `/usr/local/share/Browser` anyway since we didn't create the full directory with `FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/Browser"` first.

uberhacker: In any event, the earlier way makes more sense.

packages/torbrowser.rb
supechicken: Moving back to `self.build`

packages/torbrowser.rb
```diff
@@ -29,14 +29,32 @@ def self.patch
...
   def self.postinstall
+    puts
+    puts 'Set Tor as your default browser? [Y/n]: '
+    case STDIN.gets.chomp
+    when "\n", 'Y', 'y', 'yes'
+      Dir.chdir("#{CREW_PREFIX}/bin") do
+        FileUtils.ln_sf 'tor', 'x-www-browser'
+      end
+    else
+      puts 'No change have been made.'
```
uberhacker, line 46: `puts 'Nothing has changed.'` or `puts 'No change has been made.'`?

packages/torbrowser.rb
```diff
@@ -29,14 +29,32 @@ def self.patch
...
     puts
     puts "Type 'tor' to get started.".lightblue
     puts
   end
+  
+  def self.remove
+    Dir.chdir("#{CREW_PREFIX}/bin") do
+      FileUtils.rm_rf 'x-www-browser' if File.realpath('x-www-browser') == "#{CREW_PREFIX}/bin/tor"
+    end    
+  end
```
uberhacker, line 59: Good call!

packages/torbrowser.rb
```diff
@@ -29,14 +29,32 @@ def self.patch
...
+    end
+    
     puts
     puts "Type 'tor' to get started.".lightblue
     puts
   end
+  
+  def self.remove
+    Dir.chdir("#{CREW_PREFIX}/bin") do
+      FileUtils.rm_rf 'x-www-browser' if File.realpath('x-www-browser') == "#{CREW_PREFIX}/bin/tor"
```
uberhacker, line 57: Only should need `FileUtils.rm`.  This is a symlink, not a directory.

Issue #5285, opened 2021-02-26
Cannot install through proxy


logan2k:
My network is only working behind a proxy. The Chrome OS proxy settings don't apply to shell, including the install process for Chromebrew. How can I get around that?

satmandu:
How would you suggest one get ChromeOS's proxy settings from the command line?

satmandu:
Try doing this```export https_proxy=https://proxy.server.com:6443``` before starting the install, since curl is known to respect that environment variable.

PR #5325, opened 2021-03-04
Allow modifying default variables, and allow modifying curl

satmandu:
Fixes: #5285

- prepare for repo switch
- allow for using curl with other options e.g. `CURL="LD_LIBRARY_PATH=/usr/lib curl" ./install.sh` or `CURL="curl --proxy 127.0.0.1:8080" ./install.sh`

Works properly:
- [x] x86_64

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,9 +4,9 @@
 set -e
 
 #chromebrew directories
-OWNER="skycocker"
-REPO="chromebrew"
-BRANCH="master"
+OWNER="${OWNER:-skycocker}"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -4,16 +4,17 @@
 set -e
 
 #chromebrew directories
-OWNER="skycocker"
-REPO="chromebrew"
-BRANCH="master"
+OWNER="${OWNER:-skycocker}"
```
uberhacker, line 7: Eventually this will become chromebrew so good call.

PR #5338, opened 2021-03-07
[New Package] polari

saltedcoffii:
Builds, segfaults for unknown reason when running. I don't know JavaScript and can't debug this error.

### DIFF
--- a/crew
+++ b/crew
@@ -51,6 +51,9 @@ DOCOPT
 # If CREW_XZ_OPT is defined, use it by default.  Use `-7e`, otherwise.
 ENV["XZ_OPT"] = ENV['CREW_XZ_OPT'] || "-7e -T #{CREW_NPROC}"
 
+# If CURL environment variable exists use it in lieu of curl.
+CURL = ENV['CURL'] || 'curl'
+
 # Parse arguments using docopt
 require_relative 'lib/docopt'
 begin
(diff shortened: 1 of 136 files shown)


uberhacker:
Any reason this PR was labeled Draft?  It looks good to me.

saltedcoffii:
Because it segfaults when trying to run it and I don't know why.

satmandu:
I couldn't get polari working several weeks ago, which is why I abandoned the underlying dep PRs I was working on. Please don't merge this until you have a working build on your end. You might have better luck with running this via flatpak...

uberhacker:
We really shouldn't create a PR until it is ready.

satmandu:
Also, can we stop adding sommelier as a required build dep?  If it works fine using wayland, (as many gtk apps should) then there is no reason to insist that one also install sommelier.

Just setting ```GDK_BACKEND=wayland``` is often enough.

uberhacker:
@satmandu: I don't think the average user is aware of these environment variables.  Does it make sense to create a wrapper script to set the correct values like we've done in other packages?

satmandu:
Maybe? depends upon i686?

satmandu:
@uberhacker Also, we are talking about the average user who has learned to jump through the hoops to put their Chromebook into dev mode? 

(I'd suggest making wayland the default for apps which can handle it.)

uberhacker:
Well, they can figure out from README.md.  I doubt many users will understand how to set environment variables unless it's scripted or documented somewhere.

saltedcoffii:
No :(

PR #5341, opened 2021-03-08
Add prestart function

uberhacker:
Currently, some packages have checks in the preinstall section that determine if installation should continue.  The problem is this requires a dependency check, the entire binary or source to be downloaded and then extracted unnecessarily first.  This new function will enable us to perform checks and abort the installation prior to the dependency check which should save time and unneeded extra processing.

### DIFF
--- a/crew
+++ b/crew
@@ -740,6 +740,7 @@ def expand_dependencies
 end
 
 def resolve_dependencies
+  @pkg.prestart
   expand_dependencies
 
   # leave only not installed packages in dependencies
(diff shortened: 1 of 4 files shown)


supechicken:
Will `precheck` better?

uberhacker:
Well, precheck would be confused with the check section for testing, right?  The idea here is to give us a chance to do some preflight checks before actually starting the install process.  So maybe `preflight` would work? :)

PR #5346, opened 2021-03-08
Add re2 library

supechicken666:
Tested on `x86_64`

### DIFF
--- a/packages/re2.rb
+++ b/packages/re2.rb
@@ -1,9 +1,7 @@
 require 'package'
 
 class Re2 < Package
-  description 'RE2 is a fast, safe, thread-friendly alternative to backtracking regular expression engines 
-like those used in PCRE, Perl, and Python.'
-
+  description 'RE2 is a fast, safe, thread-friendly alternative to backtracking regular expression engines like those used in PCRE, Perl, and Python.'
   homepage 'https://github.com/google/re2/'
   compatibility 'all'
   @_ver = '2021-02-02'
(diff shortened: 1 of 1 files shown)


packages/re2.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Re2 < Package
+  description 'RE2 is a fast, safe, thread-friendly alternative to backtracking regular expression engines 
+like those used in PCRE, Perl, and Python.'
```
uberhacker, line 5: Can we make this one line?  I think the wrapping may cause a problem.

packages/re2.rb
```diff
@@ -0,0 +1,23 @@
...
+  source_url "https://github.com/google/re2/archive/#{@_ver}.tar.gz"
+  source_sha256 '1396ab50c06c1a8885fb68bf49a5ecfd989163015fd96699a180d6414937f33f'
+
+  def self.build
+    # cmake version doesn't provide libre2 library, use make instead
+    system 'make'
+  end
+  
+  def self.install
+    system "make DESTDIR=#{CREW_DEST_DIR} #{CREW_OPTIONS.gsub('--', '')} install"
```
uberhacker, line 21: Are the compiler options lowercase?

supechicken: (Yeah)[https://www.gnu.org/software/make/manual/html_node/Directory-Variables.html#Directory-Variables]

packages/re2.rb
supechicken: Removed

PR #5347, opened 2021-03-08
Add preflight function

uberhacker:
Currently, some packages have checks in the preinstall section that determine if installation should continue. The problem is this requires a dependency check, the entire binary or source to be downloaded and then extracted unnecessarily first. This new function will enable us to perform checks and abort the installation prior to the dependency check which should save time and unneeded extra processing.


supechicken:
Update `Manual.md` also?

uberhacker:
Added.

satmandu (approved):

PR #5348, opened 2021-03-08
Update android_studio from 3.5.3.0 to 4.1.2.0

uberhacker:
Depends on #5347.

### DIFF
--- a/packages/android_studio.rb
+++ b/packages/android_studio.rb
@@ -39,11 +39,10 @@ class Android_studio < Package
     puts
     puts 'To start using Android Studio, type `studio`.'.lightblue
     puts
-    puts 'To completely remove Android Studio, including the'.lightblue
-    puts 'settings, SDK and tools, execute the following:'.lightblue
-    puts 'crew remove android_studio'.lightblue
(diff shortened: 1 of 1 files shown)


packages/android_studio.rb
```diff
@@ -43,7 +43,7 @@ def self.postinstall
     puts 'settings, SDK and tools, execute the following:'.lightblue
     puts 'crew remove android_studio'.lightblue
     puts "rm -rf #{CREW_PREFIX}/.config/Android".lightblue
-    puts "rm -rf #{CREW_PREFIX}/.config/.AndroidStudio3.5".lightblue
```
supechicken, line 46: Let `self.remove` do it?

supechicken (approved):

packages/android_studio.rb
supechicken: This line can be removed as the symlink installed in `CREW_DEST_HOME`

uberhacker: Good call.

uberhacker: Done.

satmandu:
Are we sure that removing a config dir automatically is a good idea?

uberhacker:
I can add a prompt.

PR #5356, opened 2021-03-08
add signal_desktop prereqs

satmandu:
- updates go and uses prebuilt binaries for i686 & x86_64, still building for armv7l.
- updates imagemagick7
-  Adds sccache

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/cfitsio.rb
+++ b/packages/cfitsio.rb
@@ -1,6 +1,3 @@
-# Adapted from Arch Linux cfitsio PKGBUILD at:
-# https://github.com/archlinux/svntogit-packages/raw/packages/cfitsio/trunk/PKGBUILD
-
 require 'package'
 
 class Cfitsio < Package
@@ -24,8 +21,11 @@ class Cfitsio < Package
      x86_64: '9957f638dd3ea9188ded8bbee7970955947784d51633033a3693ebb746158330'
   })
 
+  def self.patch
+     system "sed -e 's|LDFLAGS=.*|LDFLAGS=$LDFLAGS|g' -i configure.in # Fix LDFLAGS"
+  end
+  
   def self.build
-    system "sed -e 's|LDFLAGS=.*|LDFLAGS=$LDFLAGS|g' -i configure.in # Fix LDFLAGS"
     system 'autoreconf -vi'
     system "env CFLAGS='-pipe -fno-stack-protector -U_FORTIFY_SOURCE -flto=auto' \
       CXXFLAGS='-pipe -fno-stack-protector -U_FORTIFY_SOURCE -flto=auto' \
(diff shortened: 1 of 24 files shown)


packages/cfitsio.rb
```diff
@@ -0,0 +1,41 @@
...
+  })
+  binary_sha256({
+    aarch64: '0d24de398a86b55b04f5601e856f872cccbe1e9d347bf918f7ad18af8bae776e',
+     armv7l: '0d24de398a86b55b04f5601e856f872cccbe1e9d347bf918f7ad18af8bae776e',
+       i686: '728f1602789e0eacf81236e0739d59dd27dca649c6c57a9c9947a730d0aa2d25',
+     x86_64: '9957f638dd3ea9188ded8bbee7970955947784d51633033a3693ebb746158330'
+  })
+
+  def self.build
+    system "sed -e 's|LDFLAGS=.*|LDFLAGS=$LDFLAGS|g' -i configure.in # Fix LDFLAGS"
```
uberhacker, line 28: Can we move this to `def self.patch`?

satmandu: Done.

packages/go_tools.rb
```diff
@@ -0,0 +1,41 @@
...
+    @git_dir = 'go_tools_git'
+    @git_hash = "gopls/v#{@_ver}"
+    @git_url = 'https://github.com/golang/tools/'
+    FileUtils.rm_rf(@git_dir)
+    FileUtils.mkdir_p(@git_dir)
+    Dir.chdir @git_dir do
+      system 'git init'
+      system "git remote add origin #{@git_url}"
+      system "git fetch --depth 1 origin #{@git_hash}"
+      system 'git checkout FETCH_HEAD'
```
uberhacker, line 37: I'm guessing you are doing all this to speed things up?  Otherwise, you could replace lines 31-37 with:
```
system "git clone #{@git_url}/#{@git_dir}"
Dir.chdir @git_dir do
  system "git checkout #{@git_hash}"
end
```

packages/libaom.rb
```diff
@@ -0,0 +1,72 @@
...
+    @git_dir = 'aom_git'
+    @git_hash = '8b6eaa4d37b992efffe432fa513889e6d11ae04f'
+    @git_url = 'https://aomedia.googlesource.com/aom'
+    FileUtils.rm_rf(@git_dir)
+    FileUtils.mkdir_p(@git_dir)
+    Dir.chdir @git_dir do
+      system 'git init'
+      system "git remote add origin #{@git_url}"
+      system "git fetch --depth 1 origin #{@git_hash}"
+      system 'git checkout FETCH_HEAD'
```
uberhacker, line 36: Can't you just `git clone` and `git checkout <hash>`?  Seems like a lot of extra work.

packages/go_tools.rb
satmandu: This is to speed things up by not having to do a full git clone before a git checkout. This checks out just the git hash or tag, as I understand it...

packages/libaom.rb
satmandu: This dramatically reduces the size of data transfer because it avoids transferring all history during the clone.

PR #5373, opened 2021-03-09
Add libva-utils

supechicken666:
Need to unset `XDG_RUNTIME_DIR` for me to get it working

Tested on `x86_64`


satmandu:
Try this: 
```
  def self.patch
    system "find . -type f -exec sed -i 's,-fstack-protector,-flto=auto,g' {} +"
  end
```

supechicken:
Switched source to git snapshot, use meson

PR #5390, opened 2021-03-10
mawk: Add awk symlink

satmandu:
- Fixes issues with ChromeOS awk not functioning because of sandbox protections.

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/mawk.rb
+++ b/packages/mawk.rb
@@ -3,31 +3,35 @@ require 'package'
 class Mawk < Package
   description 'mawk is a POSIX compatible interpreter of the awk language faster than gawk.'
   homepage 'https://salsa.debian.org/debian/mawk'
-  version '1.3.4'
+  version '1.3.4-1'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


uberhacker:
Good call.

saltedcoffii:
@satmandu perhaps instead of having this automatically be the symlink, maybe have a program similar to `ld_default` called `awk_default` that allows the user to set the `awk` symlink.

satmandu:
Feel like porting Debian's etc-alternatives scripts?  I'd rather not reinvent the wheel...

saltedcoffii:
update alternatives is part of `dpkg`. I think perhaps something like gentoo's `eselect` might be more what we're looking for, however both may work.

satmandu:
> `update-alternatives` is part of `dpkg`. I think perhaps something like gentoo's `eselect` might be more what we're looking for, however both may work.

Pretty sure `update-alternatives` is shell script and `eselect` is python. Whatever language you like better, or make it ruby... Maybe homebrew has a version?

saltedcoffii:
Eselect is in bash, and update-alternatives might be in C

PR #5395, opened 2021-03-10
add gm4 symlink to m4

saltedcoffii:


packages/m4.rb
```diff
@@ -3,24 +3,11 @@
 class M4 < Package
   description 'GNU M4 is an implementation of the traditional Unix macro processor.'
   homepage 'https://www.gnu.org/software/m4/'
-  version '1.4.18'
+  version '1.4.18-1'
   compatibility 'all'
-  source_url 'https://ftp.gnu.org/gnu/m4/m4-1.4.18.tar.xz'
+  source_url 'https://ftpmirror.gnu.org/m4/m4-1.4.18.tar.xz'
```
uberhacker, line 8: Good call.

packages/m4.rb
```diff
@@ -30,6 +17,9 @@ def self.build
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    Dir.chdir "#{CREW_DEST_PREFIX}/bin" do
+      FileUtils.symlink 'm4', 'gm4'
+    end
```
uberhacker, line 22: What is the advantage of doing ths?

saltedcoffii: autotools configure scripts look for `gm4` before looking for `m4`. Also, it's a little alias for users from BSD systems used to typing `gm4` when referring to GNU m4

PR #5396, opened 2021-03-10
add gmake symlink to make

saltedcoffii:


packages/make.rb
```diff
@@ -30,6 +17,9 @@ def self.build
 
   def self.install
     system './make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    Dir.chdir "#{CREW_DEST_PREFIX}/bin" do
+      FileUtils.symlink 'make', 'gmake'
+    end
```
uberhacker, line 22: What is the advantage of doing this?  Since this is a core package, we'll need binaries before it's merged.

Issue #5398, opened 2021-03-11
glibc fails to build


saltedcoffii:
When running `crew build glibc`, it fails. Here's the last 100 lines of the failure.
[glibc.log](https://github.com/skycocker/chromebrew/files/6122952/glibc.log)

saltedcoffii:
One of the patches in [this tarball](https://dev.gentoo.org/~dilfridge/distfiles/glibc-2.27-patches-3.tar.bz2) fixed it. I'm not sure which one, as I applied all of them.

satmandu:
Bisect?

Glibc is a little weird because we are essentially just using the system glibc and only using the headers from this package, right?

saltedcoffii:
Not necessarily. We're creating some dynamic libraries that the system doesn't have like some of the nss libraries and libBrokenLocale, and we're also making static libraries and binaries like `ldd`.

satmandu:
FYI, this might be some of the patches google used here: https://source.chromium.org/chromiumos/chromiumos/codesearch/+/main:src/third_party/chromiumos-overlay/sys-libs/glibc/files/local/glibc-2.27/?q=glibc

satmandu:
Also the relevant ebuild: https://source.chromium.org/chromiumos/chromiumos/codesearch/+/main:src/third_party/chromiumos-overlay/sys-libs/glibc/glibc-2.27-r22.ebuild

satmandu:
Also if you look at our package file for glibc, you'll notice that it's really just a proxy for generating headers and some locales. We symlink the actual ld.so file... 

Feel free to figure out how to make a working actual build that doesn't immediately lead to segfaults in everything...

satmandu:
Here's an example of a non-working glibc package file which will compile but just lead to segfaults when you run something like `ls`:
```
require 'package'

class Glibc < Package
  description 'The GNU C Library project provides the core libraries for GNU/Linux systems.'
  homepage 'https://www.gnu.org/software/libc/'
  version '2.2x'
  compatibility 'all'
  case LIBC_VERSION
  when '2.23'
    source_url 'https://ftpmirror.gnu.org/glibc/glibc-2.23.tar.xz'
    source_sha256 '94efeb00e4603c8546209cefb3e1a50a5315c86fa9b078b6fad758e187ce13e9'
    binary_url ({
      aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/glibc-2.23-chromeos-armv7l.tar.xz',
[...]

PR #5399, opened 2021-03-11
Fix mixed triple issue

saltedcoffii:
Chromebrew has had the weirdest problem that I originally noticed a few months ago but didn't know what it was. Sometimes chromebrew's host triplet is what it should be: `#{ARCH}-cros-linux-gnu`. Other times, it's `#{ARCH}-pc-linux-gnu`.

This fix allows newly compiled packages to properly identify the current triplet being used and, if need be, cross compile. This fix will only cause compiling packages to cross compile if the `--build` is different than `--host`.

I've made this change and compiled multiple packages, including core packages, and I noticed no problems (except with glibc, however that's an unrelated issue that occurs without this change, relevant issue here: #5398)

After merging this, it would be prudent to recompile binaries for `binutils`, `gcc` and potentially `llvm`.


uberhacker:
I'm confused why we would need to recompile binaries if the host and target are the same.  Also, if you are going to introduce a new constant, that should also be added to `crew` so `crew const` can return the value.  That means bumping the crew version as well.

saltedcoffii:
For some reason, they aren't. Sometimes autotools configure scripts will find `#{ARCH}-cros-linux-gnu` as the build machine and host (because it was passed these arguments), so it won't cross compile. But because the triple is actually `#{ARCH}-pc-linux-gnu`, and it's not cross compiling (because the passed --build and --host are the same) it will compile with the triple `#{ARCH}-pc-linux-gnu`. Watch a configure script that hasn't been passed these arguments (run outside of chromebrew most likely) and you'll see what I'm talking about. Hopefully that makes sense.

saltedcoffii:
Also would you mind updating crew? I don't wanna break anything.

satmandu:
@saltedcoffii modified `crew` and `CREW_VERSION` for you.

Issue #4292, opened 2020-09-08
[Suggestion] Installing curl before libressl


ThatGeekyWeeb:
The installation process consists, of installing, `gcc8`, `libressl`, `ruby`, `libiconv`, `git` and `libssh2`
***
One of the main issues is replacing the systems `openssl` libs with `libressl`
***
Throughout the versions of ChromeOS, the `openssl` version used varies widely. As a result, there are many issues, usually centred around `curl` and `libressl`. I believe a simple fix is possible.
***
Upon installation of `libressl`, the systems `curl` installation breaks, as it was not compiled using the same version of `openssl`, I believe a simple fix would be to install a statically compiled version of `curl`, before installing `libressl`.
***
If we were to statically compile `curl` & `libcurl` to use a `tls` implementation, instead of `libressl`, In theory, this would fix issues caused by replacing the libraries of dynamically linked `curl` install.
***
I recommend installing `curl` before `libressl` and then adjusting the script to use `/usr/local/bin/curl` instead of `/usr/bin/curl`

uberhacker:
I have not noticed the out-of-the-box curl breaking with libressl.  Most of the issues I've noticed were after installing the Chromebrew version of curl.  This is most likely due to packages that depend on curl not being recompiled.  What specific scenarios have you noticed?

ThatGeekyWeeb:
Yes, the Chromebrew `curl` seems broken out of the box. Personally my device uses the 3.10 Kernel which doesn't have full support for ssl, (I am also running ChromeOS 83), but, supporting older version of ChromeOS becomes very hard, mainly with `libressl`, when I was using ChromeOS 77, this was a solution to issues with `libressl`, and I believe it would be a fix that would make Chromebrew more backwards compatible.

uberhacker:
FWIW, you might try adding the following in your `~/.bashrc`:
```
export LD_LIBRARY_PATH=/usr/lib:/usr/local/lib
```
This got rid of the curl error for me.

uberhacker:
Just out of curiosity, is the curl error you noticed the same as below?
```
curl: symbol lookup error: /usr/lib/libcurl.so.4: undefined symbol: RAND_egd
```

ThatGeekyWeeb:
My exact issue was
```
curl: symbol lookup error: /usr/lib/libcurl.so.4: undefined symbol: SLL_COMP_free_compression_methods
```

PR #5404, opened 2021-03-12
[WIP] Remove curl and its dependencies from install.sh

supechicken666:
Since `libressl` is replaced by `openssl` now, switch to system version of `curl` (#4292)

I also added a `puts` function for replacing color variables

Tested on `x86_64`


install.sh
```diff
@@ -16,31 +16,60 @@ CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
...
-
-if [ "${EUID}" == "0" ]; then
-  echo -e "${RED}Chromebrew should not be installed or run as root.${RESET}"
-  exit 1;
-fi
+function rootchecker() {
+  if [ ${EUID} == 0 ]; then
+    puts 'Chromebrew should not be installed or run as root.' yellow
+    puts "Trying to switch to user chronos..." yellow
+    su chronos - <<< "$(curl -L# ${URL}/install.sh)" ||\
```
uberhacker, line 61: Something doesn't feel right about this.  If you do this, don't you want to append ` && exit` afterwards?  Otherwise, the install continues after this and basically installs twice?

install.sh
```diff
@@ -16,31 +16,60 @@ CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
...
+    colorize=${colorize}'\e[1A\e[K';;
+  covercurrent)
+    colorize=${colorize}'\r';;
+  printf);;
+  *)
+    reset=${reset}'\n';;
+  esac
+  
+  printf "${colorize}${1}${reset}"
+}
```
uberhacker, line 55: Nice idea.

install.sh
```diff
@@ -16,31 +16,60 @@ CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
 CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 CURL="${CURL:-curl}"
 
-EARLY_PACKAGES="gcc10 llvm brotli c_ares libcyrussasl libiconv libidn2 \
-libmetalink libnghttp2 libpsl libssh2 libtirpc libunistring openldap \
-rtmpdump zstd ncurses ca_certificates ruby libffi openssl nettle krb5 \
-p11kit libtasn1 gnutls curl git icu4c libedit"
+EARLY_PACKAGES=(gcc10 libiconv libssh2 ncurses ruby openssl git)
```
uberhacker, line 19: Didn't we need to add curl because i686 Chromebooks were misbehaving during install?

install.sh
```diff
@@ -78,37 +107,37 @@ case "${ARCH}" in
...
 # functions to maintain packages
 function download_check () {
     cd "${CREW_BREW_DIR}"
 
     #download
-    echo -e "${BLUE}Downloading ${1}...${RESET}"
+    puts "=> Downloading ${1}..." lightblue
+    puts "=> ${BREW} ${2}" cyan
     $CURL '-#' -C - -L --ssl "${2}" -o "${3}"
+    puts ' └─> Archive download' lightgreen cover
```
uberhacker, line 132: How did you print this non-ascii character?

install.sh
```diff
@@ -163,16 +196,14 @@ if [ ! -f device.json ]; then
...
-  sha256="${sha256s["${i}"]}"
+for i in ${!urls[@]}; do
   tarfile="$(basename ${url})"
   name="${tarfile%%-*}"   # extract string before first '-'
   rest="${tarfile#*-}"    # extract string after first '-'
   version="$(echo ${rest} | sed -e 's/-chromeos.*$//')"
                         # extract string between first '-' and "-chromeos"
 
-  download_check "${name}" "${url}" "${tarfile}" "${sha256}"
+  download_check "${name}" "${urls[${i}]}" "${tarfile}" "${sha256s[${i}]}"
```
uberhacker, line 206: Is `$i` a number or the url?

install.sh
```diff
@@ -163,16 +196,14 @@ if [ ! -f device.json ]; then
 fi
 
 # extract, install and register packages
-for i in $(seq 0 $((${#urls[@]} - 1))); do
-  url="${urls["${i}"]}"
-  sha256="${sha256s["${i}"]}"
+for i in ${!urls[@]}; do
   tarfile="$(basename ${url})"
```
uberhacker, line 200: What is `${url}`?

satmandu:
The system curl can be used by install.sh now with no changes by just running `CURL="LD_LIBRARY_PATH=/usr/lib /usr/bin/curl" ./install.sh` (or using /usr/lib64) on x86_64...

install.sh
satmandu: IIRC some i686 images have a broken curl in them... which is a big reason NOT to do this at all.

uberhacker:
@supechicken666: Can we close this?

install.sh
saltedcoffii:
```bash
case $(uname -m) in ARCH
   i*86)
    early-packages+="curl"
    ;;
esac
```

Or however that would be in Ruby, maybe.

uberhacker: That could work.  I'm all for reducing the size and scope of core packages.

PR #5408, opened 2021-03-13
Update tests

uberhacker:
Improved tests.


tests/load_test
```diff
@@ -0,0 +1,19 @@
...
+require 'find'
+require_relative '../lib/const'
+require_relative '../lib/color'
+
+PKG_DIR = '../packages'
+
+# Add >LOCAL< lib to LOAD_PATH
+$LOAD_PATH.unshift '../lib'
+
+Find.find(PKG_DIR) do |filename|
```
supechicken, line 12: Will `Dir.glob("#{PKG_DIR}/*.rb").each do |filename|` better?

uberhacker: With suggested change:
```
$ time ruby load_test
...snip...
All load tests successful.

real    0m23.684s
user    0m10.799s
sys     0m0.855s
```
Without suggested change:
```
$ time ruby load_test
...snip...
All load tests successful.

real    0m24.570s
user    0m10.678s
sys     0m1.312s
```
Negligible performance improvement?

satmandu: That's a 2 second improvement!

Run it several times to get an average though...

uberhacker: It's a 2 second improvement without the suggested change. :)  I did run it several times and it varies a few seconds each time so the suggested change is effectively negligible.

uberhacker:
@satmandu: Is there a rubocop equivalent to the `cycle_test` script?  If so, how do they compare?

tests/load_test
supechicken: Try again after removing the following lines?
```
require 'find'
  if File.extname(filename) == '.rb'
```

uberhacker: Yes, now I see what you are getting at.  Good observation!

uberhacker: Definitely an improvement!
```
$ time ruby load_test
...snip...
All load tests successful.                                                                                                                                                                    

real    0m8.245s
user    0m1.082s
sys     0m0.245s
```

uberhacker: I made the same change in the cycle_test script and that was an improvement also!

supechicken (approved):

PR #5425, opened 2021-03-15
gcalculator: Only download checksum at downloading archive

supechicken666:
Keep all stuff installed at `#{CREW_PREFIX}/Applications/gcalculator/` and symlink to `CREW_PREFIX`

This will definitely help user to organize packages when all package is installed to their own dirs :)


satmandu:
This violates the [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). I'm not tied to it,  but Chromebrew is mostly compliant with it. There are amazing distributions like [NixOS](https://nixos.org/) which don't use FHS, 

We probably want more discussion before we head in that direction though, just because consistency between app installs is useful not least to end-users.

supechicken666:
> This violates the [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). I'm not tied to it, but Chromebrew is mostly compliant with it. There are amazing distributions like [NixOS](https://nixos.org/) which don't use FHS,
> 
> We probably want more discussion before we head in that direction though, just because consistency between app installs is useful not least to end-users.

I agree with some of it. We have symlinks to keep the compatibility with it so I am not worried about it (How about just create the symlinks of a package to one dir while following it?)

uberhacker:
@supechicken: We sorta have an unwritten rule to put all apps in their own isolated folder under `/usr/local/share` if they don't install themselves in some other preferred location.  I definitely understand your dilemma.

packages/gcalculator.rb
```diff
@@ -5,22 +5,26 @@ class Gcalculator < Package
...
 
   depends_on 'setuptools' => :build
   depends_on 'gtk3'
   depends_on 'gtksourceview'
   depends_on 'itstool'
   depends_on 'libgee'
   depends_on 'libsoup'
+  
+  def self.preflight
+    source_sha256 `curl -Ls #{@_url}.sha256sum | tail -n1 | cut -d' ' -f1`.chomp
```
uberhacker, line 21: Why is this here?

packages/gcalculator.rb
```diff
@@ -5,22 +5,26 @@ class Gcalculator < Package
...
+  end
 
   def self.build
-    system "meson #{CREW_MESON_LTO_OPTIONS} builddir"
+    @_prefix = "#{CREW_PREFIX}/Applications/#{self.name}/"
```
uberhacker, line 25: How about `#{CREW_PREFIX}/share/#{self.name}`?

packages/gcalculator.rb
supechicken: To get the checksum only at download (not search and update)

uberhacker:
@supechicken666: Merge conflict.

supechicken666:
> @supechicken666: Merge conflict.

Done

packages/gcalculator.rb
uberhacker: I thought https://github.com/skycocker/chromebrew/pull/5385 already handles this?

PR #5428, opened 2021-03-15
rebuild gemacs from emacs feature/pgtk branch,  add libotf dep, and hashpipe

### LINKED ISSUE #5424: emacs and gemacs binaries need to be regenerated
While running gemacs, I faced the following error 
``` shell
/usr/local/bin/gemacs: error while loading shared libraries: libMagickWand-7.Q16HDRI.so.7: cannot open shared object file: No such file or directory
```
Running `crew reinstall -s gemacs` solved the problem. I suppose it should be the same with emacs package

satmandu:
Fixes #5424 

- gemacs actually works with Wayland now
- adds libotf
- adds hashpipe for easy one line sha256sum in-pipe checking.

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/gemacs.rb
+++ b/packages/gemacs.rb
@@ -3,53 +3,90 @@ require 'package'
 class Gemacs < Package
   description 'An extensible, customizable, free/libre text editor - and more.'
   homepage 'https://www.gnu.org/software/emacs/'
-  @_ver = '27.1'
-  version "#{@_ver}-2"
+  @_ver = '28.0.50.1'
+  version "#{@_ver}-7a7b"
   compatibility 'all'
-  source_url "https://ftpmirror.gnu.org/emacs/emacs-#{@_ver}.tar.xz"
-  source_sha256 '4a4c128f915fc937d61edfc273c98106711b540c9be3cd5d2e2b9b5b2f172e41'
+  source_url 'file:///dev/null'
+  source_sha256 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
 
   binary_url({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gemacs-27.1-2-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/gemacs-27.1-2-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/gemacs-27.1-2-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/gemacs-27.1-2-chromeos-x86_64.tar.xz'
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gemacs-28.0.50.1-7a7b-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/gemacs-28.0.50.1-7a7b-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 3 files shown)


packages/gemacs.rb
```diff
@@ -3,53 +3,90 @@
...
   depends_on 'libtiff'
-  depends_on 'imagemagick7'
-  depends_on 'sommelier'
+  depends_on 'texinfo'
+  depends_on 'webkit2gtk'
+
+  def self.prebuild
+    @git_dir = 'emacs_git'
+    @git_branch = 'feature/pgtk'
+    @git_hash = '7a7bc15242896b20c7af49f77f0e22c3d78e4d88'
```
uberhacker, line 43: Can we check out with a git tag instead?  Might be able to use `#{@_ver}` with the tag.

packages/libotf.rb
```diff
@@ -0,0 +1,44 @@
...
+     armv7l: '48ea420876299f2de783c250b6def3c2a87e7bf8c16695a63b7561efbb9594f8',
+       i686: '487c98f04bdc12c06a2afd32087222150ee652d9fc3c6233daee341c41718406',
+     x86_64: '8ec1c9f4e8b99482323e3bf444cca95b5e3d0abb90b8ea278d1461b29470e340'
+  })
+
+  depends_on 'libxaw'
+  depends_on 'freetype'
+
+  def self.patch
+    system 'curl -Ls https://github.com/archlinux/svntogit-packages/raw/a67b940a19f1e7e47e7d8553ed31158ed70f1286/libotf/trunk/replace-freetype-config.patch | patch -p1 -b'
```
uberhacker, line 29: Can we patch with a tagged release instead of trunk?

uberhacker: Also, no sha256sum check?

packages/gemacs.rb
satmandu: @git_hash = '7a7bc15242896b20c7af49f77f0e22c3d78e4d88' is the git tag as I understand it.

packages/libotf.rb
satmandu: That's a tagged release, not trunk:
`https://github.com/archlinux/svntogit-packages/raw/a67b940a19f1e7e47e7d8553ed31158ed70f1286/libotf/trunk/replace-freetype-config.patch`

Not sure about the point of a sha256sum for a tagged git file downloaded from github via https? What is that trying to accomplish? Are we worried about curl being unable to download the file w/o corruption?

uberhacker: Whenever you download anything using curl, wget or whatever, there is always the possibility the download will fail or only download partially.  This is just consistent with everything else we download in all our packages.  Should be SOP.  I checked out the https://github.com/archlinux/svntogit-packages/ repo and it doesn't look like they have any tagged releases yet so a git hash is fine.

packages/gemacs.rb
uberhacker: I don't know where you are getting this version but from what I can tell, the latest tagged release is:
```
fda9d916712db6a1262cdb4539eeb4665b6cefb7	refs/tags/emacs-27.1.91
```
See https://git.savannah.gnu.org/git/emacs.git/info/refs.

packages/libotf.rb
satmandu: In general archlinux doesn't bother tagging releases, so git tags are the only option there.

satmandu: Using incoming hashpipe package now. :)

packages/gemacs.rb
satmandu: This is from the feature/pgtk branch (which enables full gtk support allowing usage from Wayland.)

https://git.savannah.gnu.org/cgit/emacs.git/log/?h=feature/pgtk

PR #5429, opened 2021-03-15
gtk3 & gtk4 update

satmandu:
- gtk3 => 3.24.27
- gtk4 => 4.1.2
- add cantarell fonts and mark them as the font to use for gtk4
- updated adwaita_icon_theme to not overwrite gtk settings file
- Also update settings file logic to conform to XDG paths

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/gtk3.rb
+++ b/packages/gtk3.rb
@@ -56,8 +56,6 @@ class Gtk3 < Package
   def self.install
     system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
     system "sed -i 's,null,,g'  #{CREW_DEST_LIB_PREFIX}/pkgconfig/gtk*.pc"
-    FileUtils.mkdir_p("#{CREW_DEST_HOME}/.config/gtk-3.0")
-    File.write("#{CREW_DEST_HOME}/.config/gtk-3.0/settings.ini", @gtk3settings)
   end
 
   def self.postinstall
--- a/packages/gtk3.rb
+++ b/packages/gtk3.rb
@@ -11,44 +11,32 @@ class Gtk3 < Package
   # source_url "https://download.gnome.org/sources/gtk/#{@_ver_prelastdot}/gtk-#{@_ver}.tar.xz"
   source_sha256 'dbc3b14ae0d8e44bc8caeac91d62b4d2403a881d837cb4e9bcfb8d138712c3a3'
 
-  binary_url({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gtk3-3.24.27-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/gtk3-3.24.27-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/gtk3-3.24.27-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/gtk3-3.24.27-chromeos-x86_64.tar.xz'
-  })
-  binary_sha256({
-    aarch64: '840b06f1ccf18e6b85e7944e1af74e9a08385053a46f0277c8f395fa02c20f2a',
-     armv7l: '840b06f1ccf18e6b85e7944e1af74e9a08385053a46f0277c8f395fa02c20f2a',
-       i686: 'f07416c233166be3cf0777be1745d61ed6f3574137a7918c300779acc39593d8',
-     x86_64: 'bdd162ae2fd29f1500e2dd6a7ed7378023b2e20d3af73c42c3017ef10f4d905f'
-  })
-
-  depends_on 'cups'
+  depends_on 'atk'
   depends_on 'at_spi2_atk'
+  depends_on 'cantarell_fonts'
+  depends_on 'cups'
+  depends_on 'gdk_pixbuf'
   depends_on 'gnome_icon_theme'
   depends_on 'gobject_introspection'
-  depends_on 'gdk_pixbuf'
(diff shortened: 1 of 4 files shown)


packages/gtk3.rb
```diff
@@ -49,26 +51,33 @@ def self.build
...
-      gtk-font-name = Arial 10
-    GTK3_CONFIG_HEREDOC
   end
 
   def self.install
     system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
     system "sed -i 's,null,,g'  #{CREW_DEST_LIB_PREFIX}/pkgconfig/gtk*.pc"
     FileUtils.mkdir_p("#{CREW_DEST_HOME}/.config/gtk-3.0")
-    File.write("#{CREW_DEST_HOME}/.config/gtk-3.0/settings.ini", @file)
+    File.write("#{CREW_DEST_HOME}/.config/gtk-3.0/settings.ini", @gtk3settings)
```
uberhacker, line 60: Where is `@gtk3settings` defined?  I see it in `postinstall` but doesn't that come after `install`?

packages/gtk3.rb
```diff
@@ -49,26 +51,33 @@ def self.build
...
     # update mime database
     system "update-mime-database #{CREW_PREFIX}/share/mime"
+    @xdg_config_home = ENV['XDG_CONFIG_HOME']
+    @gtk3settings = <<~GTK3_CONFIG_HEREDOC
+      [Settings]
+      gtk-application-prefer-dark-theme = false
+      gtk-icon-theme-name = hicolor
+      gtk-fallback-icon-theme = gnome
+      gtk-font-name = Arial 10
+    GTK3_CONFIG_HEREDOC
```
uberhacker, line 75: Move to `def self.build` instead?

packages/gtk3.rb
```diff
@@ -49,26 +51,33 @@ def self.build
...
+    unless File.exist?("#{@xdg_config_home}/gtk-3.0/settings.ini")
+      puts
+      puts 'Adding basic gtk4 settings to XDG_CONFIG_HOME/gtk-3.0/settings.ini'.lightblue
+      FileUtils.mkdir_p "#{@xdg_config_home}/gtk-3.0"
+      File.write("#{@xdg_config_home}/gtk-3.0/settings.ini", @gtk3settings)
```
uberhacker, line 80: Didn't we already install this config?  In any event, it probably should be included in the package files anyway.  Maybe add a `def self.remove` to optionally remove the config?

packages/gtk3.rb
satmandu: good catch, that's extraneous.

packages/gtk3.rb
satmandu: The heredoc isn't actually being written to disk unless the condition in postinstall is true.  (This way you can run `crew postinstall gtk3`)

uberhacker: What about creating `settings.ini.example` and include it in the package files but then copy `settings.ini.example` to `settings.ini` if it doesn't exist?  If we do this then we will want a `def self.remove` section to optionally remove `settings.ini` when the package is removed.

packages/gtk3.rb
satmandu: You don't want to overwrite local customizations, which are likely to happen with gtk. Simple things like changing the default font would get overwritten.

I would hesitate on a self-remove for this. We're not talking about a large install like you have with wine in the config directory.  The config file is measured in bytes...

satmandu: Actually, why aren't we installing the system-wide versions of these files?

satmandu: Stand by...

packages/gtk3.rb
satmandu: I think there's plenty of documentation on settings for gtk out there. By just including a default settings.ini file, we bypass all this craziness. People are free to edit their local `gtk-{3,4}/settings.ini` as they see fit

packages/gtk3.rb
satmandu: ok should be good now.

satmandu:
all done

packages/gtk3.rb
```diff
@@ -3,66 +3,71 @@
...
-    FileUtils.mkdir_p("#{CREW_DEST_HOME}/.config/gtk-3.0")
-    File.write("#{CREW_DEST_HOME}/.config/gtk-3.0/settings.ini", @file)
+    @gtk3settings = <<~GTK3_CONFIG_HEREDOC
+      [Settings]
+      gtk-icon-theme-name = Adwaita
+      gtk-fallback-icon-theme = gnome
+      gtk-theme-name = Adwaita
+      gtk-font-name = Cantarell 11
+      gtk-application-prefer-dark-theme = false
+    GTK3_CONFIG_HEREDOC
```
uberhacker, line 68: Technically, this is a build step.

packages/gtk4.rb
```diff
@@ -47,39 +53,36 @@ def self.patch
...
   end
 
   def self.install
     system "DESTDIR=#{CREW_DEST_DIR} ninja -C build install"
+    @gtk4settings = <<~GTK4_CONFIG_HEREDOC
+      [Settings]
+      gtk-icon-theme-name = Adwaita
+      gtk-theme-name = Adwaita
+      gtk-font-name = Cantarell 11
+    GTK4_CONFIG_HEREDOC
```
uberhacker, line 77: Put in `def self.build` instead?

satmandu: done

packages/gtk3.rb
satmandu: done

PR #5439, opened 2021-03-17
libjpeg_turbo => 2.0.6

satmandu:
- Yes I know this conflicts with `libjpeg`, but the key is to build packages with libjpeg installed last, no, since these are technically supposed to be binary compatible? So this just brings this package up to the current release.

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686


supechicken (approved):

uberhacker:
Does this work to replace `libjpeg` eventually?  How do other packages that depend on `libjpeg` work with this instead?

satmandu:
I think the best way out of this is to move all packages TO the regular libjpeg.

Having said that, I think it makes sense to leave the package for libjpeg_turbo up to date until that happens.

uberhacker:
Fair enough.

PR #5444, opened 2021-03-17
Crew license framework

saltedcoffii:
Chromebrew doesn't have a license field in packages, but its important to some users. This adds the framework to add the `license ''` field to chromebrew packages. When this gets accepted, I'll start going around to all the `package.rb` files and adding this field.


uberhacker:
Don't you also need to add a package property in https://github.com/skycocker/chromebrew/blob/master/lib/package.rb?  If we do this, then we need to either have a default value set in each package or a way to ignore the property if a value is not set.

saltedcoffii:
I don't know what else needs to happen. I could run a `sed` on the packages directory to add the line `license 'undefined'`. Would that work?

uberhacker:
If you did that, you might as well put in the correct license.  I think undefined would confuse users.

uberhacker:
I resolved a merge conflict in openssh.

packages/desktop_file_utils.rb
```diff
@@ -0,0 +1,33 @@
...
+  license 'GPL-2+'
+  compatibility 'all'
+  source_url 'https://www.freedesktop.org/software/desktop-file-utils/releases/desktop-file-utils-0.23.tar.xz'
+  source_sha256 '6c094031bdec46c9f621708f919084e1cb5294e2c5b1e4c883b3e70cb8903385'
+
+  binary_url ({
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/desktop_file_utilities-0.23-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/desktop_file_utilities-0.23-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/desktop_file_utilities-0.23-chromeos-i686.tar.xz',
+     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/desktop_file_utilities-0.23-chromeos-x86_64.tar.xz',
```
uberhacker, line 16: These binaries don't match the package name.  I see what you are trying to do but is it worth it just to shorten the name?  I imagine most of the time you would never explicitly install this package anyway.

saltedcoffii: I did have to explicitly install it once, I forget why, and that's how I noticed that the name changed. I'll change it back in this PR but it shud be changed later

saltedcoffii:
Omg looks like my master branch somehow got mixed into this branch and i pushed it without realizing. Lemme fix that 🤦🏻

packages/bcd.rb
```diff
@@ -0,0 +1,99 @@
+require 'package'
+
+class Bcd < Package
```
uberhacker, line 3: New package?  Can we put this in a separate PR?

saltedcoffii: That's leftover, I forgot to get it in c9bf791, I'll take care of it in the next commit.

packages/fly.rb
```diff
@@ -2,8 +2,9 @@
 
 class Fly < Package
   description 'fly is a C program that creates PNG, JPEG or GIF images on the fly from CGI and other programs.'
-  homepage 'http://martin.gleeson.com/fly/index.html'
+  homepage 'http://www.unimelb.edu.au/fly/fly.html' # 403 Forbidden
```
uberhacker, line 5: I get a 404 on this page.  Change to http://www.w3perl.com/fly/?

packages/freebasic.rb
```diff
@@ -1,16 +1,17 @@
...
-  description 'FreeBASIC is a free/open source (GPL), BASIC compiler for Microsoft Windows, DOS and Linux.'
+  description 'FreeBASIC is a free and open source, BASIC compiler for Microsoft Windows, DOS and Linux.'
   homepage 'https://www.freebasic.net/'
   version '1.07.1-1'
+  license 'LGPL-2 and GPL-2'
   compatibility 'i686,x86_64'
   case ARCH
   when 'i686'
-    source_url 'https://downloads.sourceforge.net/project/fbc/Binaries%20-%20Linux/FreeBASIC-1.07.1-linux-x86.tar.gz'
+    source_url 'https://downloads.sourceforge.net/project/fbc/Binaries%20-%20Linux/FreeBASIC-1.07.1-linux-x86.tar.gz' # 404 not found
```
uberhacker, line 11: New version at https://sourceforge.net/projects/fbc/files/Binaries%20-%20Linux/FreeBASIC-1.07.2-linux-x86.tar.gz?

packages/freebasic.rb
```diff
@@ -1,16 +1,17 @@
...
     source_sha256 'a5b8f31e38a214d63a3d8e0b1ac6608f6a304891e40a3c0755940cba2f8401f1'
   when 'x86_64'
-    source_url 'https://downloads.sourceforge.net/project/fbc/Binaries%20-%20Linux/FreeBASIC-1.07.1-linux-x86_64.tar.gz'
+    source_url 'https://downloads.sourceforge.net/project/fbc/Binaries%20-%20Linux/FreeBASIC-1.07.1-linux-x86_64.tar.gz' # 404 not found
```
uberhacker, line 14: New version at https://sourceforge.net/projects/fbc/files/Binaries%20-%20Linux/FreeBASIC-1.07.2-linux-x86_64.tar.gz?

Also, arm version available now?
https://sourceforge.net/projects/fbc/files/Binaries%20-%20Linux/FreeBASIC-1.07.2-linux-arm.tar.gz

packages/freebasic.rb
saltedcoffii: Do you think we shud get this pr merged first and then fix that? New versions requires testing

uberhacker: Yes, makes sense to include in a separate PR.

saltedcoffii:
Oh my gosh it's done 😭😭😭

PR #5450, opened 2021-03-18
Delete brackets

saltedcoffii:
Brackets is depreciated

Announcement [here](http://brackets.io/)


uberhacker:
We could probably do this to a lot of packages.  Doesn't hurt to leave it in case there are some die-hard users.

saltedcoffii:
Alright, we can leave it for now. When I get around to seperating the repo into multiple repositories, this can go in a repo for depreciated packages.

PR #5460, opened 2021-03-22
Split glibmm, pangomm, cairomm, add gnome_tweaks, gnome_terminal update

satmandu:
- Solves issues with non-overlapping newer and older libraries 
- Add gnome-terminal version using meson
- Rebuild gnome-tweaks.

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/cairomm_1_0.rb
+++ b/packages/cairomm_1_0.rb
@@ -5,6 +5,7 @@ class Cairomm_1_0 < Package
   homepage 'https://www.cairographics.org/'
   @_ver = '1.14.2'
   version @_ver
+  license 'LGPL-2+'
   compatibility 'all'
   source_url "https://www.cairographics.org/releases/cairomm-#{@_ver}.tar.xz"
   source_sha256 '0126b9cc295dc36bc9c0860d5b720cb5469fd78d5620c8f10cc5f0c07b928de3'
--- a/packages/cairomm_1_0.rb
+++ b/packages/cairomm_1_0.rb
@@ -1,0 +1,44 @@
+require 'package'
+
+class Cairomm_1_0 < Package
+  description 'The Cairomm package provides a C++ interface to Cairo.'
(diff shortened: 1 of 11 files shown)


packages/cairomm_1_0.rb
```diff
@@ -0,0 +1,45 @@
+require 'package'
+
+class Cairomm_1_0 < Package
```
uberhacker, line 3: Wouldn't a better name be `Cairomm_1_14`?

packages/pangomm_1_4.rb
```diff
@@ -0,0 +1,44 @@
+require 'package'
+
+class Pangomm_1_4 < Package
```
uberhacker, line 3: Wouldn't a better name be `Pangomm_2_46`?

packages/glibmm_2_4.rb
```diff
@@ -0,0 +1,44 @@
+require 'package'
+
+class Glibmm_2_4 < Package
```
uberhacker, line 3: Wouldn't a better name be `Glibmm_2_66`?

packages/cairomm_1_0.rb
satmandu: no. The library name is cairomm-1.0. This is important because packages demand `cairomm-1.0`.

Debian has it broken out the same way for I assume the same reason: https://packages.debian.org/sid/libcairomm-1.0-1v5

packages/pangomm_1_4.rb
satmandu: Same reason as w/ cairomm-1.0

packages/glibmm_2_4.rb
satmandu: Same reason as with cairomm-1.0.

PR #5471, opened 2021-03-23
Remove mawk from buildessential

saltedcoffii:
ChromeOS provides mawk at `/usr/bin/mawk`. We can remove mawk from the core package list and just use ChromeOS's version. Works on x86_64


satmandu:
Let's **NOT** do this system mawk is tied to the ChromeOS glibc. Better to roll our own.

uberhacker:
Agreed.  We just had a long conversation regarding mawk over gawk as well.

PR #5474, opened 2021-03-24
Update php packages

uberhacker:
- Update php74 from 7.4.15 to 7.4.16
- Update php80 from 8.0.2 to 8.0.3
- Perform check in preflight section
- Add pre-built binaries


packages/php5.rb
```diff
@@ -32,9 +33,9 @@ class Php5 < Package
   depends_on 'tidy'
   depends_on 'unixodbc'
 
-  def self.preinstall
+  def self.preflight
     phpver = `php -v 2> /dev/null | head -1 | cut -d' ' -f2`.chomp
-    abort "PHP version #{phpver} already installed.".lightred unless "#{phpver}" == ""
+    abort "PHP version #{phpver} already installed.".lightred if phpver.to_s != ''
```
supechicken, line 38: Better to use `if phpver.empty?`?

uberhacker: @supechicken: Implemented.

PR #5475, opened 2021-03-24
Update firefox from 85.0.2 to 87.0

uberhacker:


satmandu:
Are you planning on building this?

uberhacker:
You mean binaries?  We have never done that in the past.

uberhacker:
The release tarballs are already pre-compiled binaries.

satmandu:
Ah! Yes.

FYI this is the true deps list for this package based upon looking at the libraries:
```
  depends_on 'apulse'
  depends_on 'atk'
  depends_on 'cairo'
  depends_on 'dbus'
  depends_on 'dbus_glib'
  depends_on 'fontconfig'
  depends_on 'freetype'
  depends_on 'gdk_pixbuf'
  depends_on 'glib'
  depends_on 'gtk2'
  depends_on 'gtk3'
  depends_on 'libx11'
  depends_on 'libxcb'
  depends_on 'libxcomposite'
  depends_on 'libxcursor'
  depends_on 'libxdamage'
  depends_on 'libxext'
  depends_on 'libxfixes'
  depends_on 'libxi'
  depends_on 'libxrender'
  depends_on 'libxt'
  depends_on 'pango'
  depends_on 'sommelier'
```

uberhacker:
Some of these missing dependencies may explain why sound is broken.

satmandu:
Also, can we replace apulse with pulseaudio?

The have conflicting libraries... and pulseaudio sound DOES work...

satmandu:
Actually, running firefox without the apulse wrapper works fine with just the current pulseaudio installed. Can you test after removing apulse and reinstalling pulseaudio?

PR #5478, opened 2021-03-24
Update codium from 1.53.2 to 1.54.3

uberhacker:


satmandu:
Does this depend upon firefox too?

```
  depends_on 'alsa_lib'
  depends_on 'atk'
  depends_on 'at_spi2_atk'
  depends_on 'at_spi2_core'
  depends_on 'cairo'
  depends_on 'cups'
  depends_on 'dbus'
  depends_on 'firefox'
  depends_on 'gdk_pixbuf'
  depends_on 'glib'
  depends_on 'gtk3'
  depends_on 'libdrm'
  depends_on 'libx11'
  depends_on 'libxcb'
  depends_on 'libxcomposite'
  depends_on 'libxdamage'
  depends_on 'libxext'
  depends_on 'libxfixes'
  depends_on 'libxkbcommon'
  depends_on 'libxkbfile'
  depends_on 'libxrandr'
  depends_on 'mesa'
  depends_on 'nspr'
  depends_on 'nss'
  depends_on 'pango'
```
(obviously one would need sommelier too... unless this is based off of a very new version of electron.)

uberhacker:
The only reason I would think firefox could be a dependency is to run the internal browser rendering engine.  Not sure why the native Chrome wouldn't work as an alternate.

uberhacker:
How many of these dependencies come along for the ride with sommelier?

satmandu:
> How many of these dependencies come along for the ride with sommelier?

Many maybe? But sommelier seems a bad fit for that sort of dependency, since one could theoretically use another wayland/X shim such as mutter in lieu of sommelier.

PR #5481, opened 2021-03-24
crew-profile-base

saltedcoffii:
This is the start of a journey to change the way Chromebrew handles environment variables 👁️ 👄 👁️


packages/crew_profile_base.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  binary_url ({
+
+  })
+  binary_sha256 ({
+
+  })
+
+  def self.preinstall
+    @rcfile = <<~EOF
```
satmandu, line 20: The EOF for docfiles is supposed to be descriptive in ruby. so maybe `RCFILEEOF`

packages/crew_profile_base.rb
```diff
@@ -0,0 +1,40 @@
...
+      # DO NOT DELETE THIS LINE
+      # See #{CREW_PREFIX}/etc/profile for further details
+      source #{CREW_PREFIX}/etc/profile
+
+      # Put your stuff under this comment
+    EOF
+    IO.write(".bashrc", @rcfile)
```
satmandu, line 27: Would this overwrite an existing .bashrc?

saltedcoffii: Yes, it would.

packages/crew_profile_base.rb
saltedcoffii: I know it *can* be descriptive in ruby, just like it can in bash, but because the `@variable` is already descriptive, would it be okay if I left it for readability and simplicity purposes?

saltedcoffii:
A few things about the latest commit (26f892f).
- This hasn't been tested; all of this is currently theoretical code. If someone tests this branch and finds errors, please fix.
- I haven't edited daemon-y stuff like `broadway` and `cups` because I want to wait to edit those until I get `crew-init`'s skeleton down.
- @ someone who knows ruby: do constants within multi-block structures get replaced/expanded?
    Example multi-block structure:
```ruby
    @env = <<~EOF
      # Apache Ant configuration
      JAVA_HOME=#{CREW_PREFIX}/share/jdk8'
      ANT_HOME=$JAVA_HOME
    EOF
```
If not, what workarounds do you suggest?

supechicken:
> * @ someone who knows ruby: do constants within docfiles get replaced/expanded?

Yes

packages/crew_profile_base.rb
saltedcoffii: This is resolved. See https://github.com/saltedcoffii/chromebrew/blob/env-d/packages/crew_profile_base.rb

saltedcoffii:
Okay this PR is ready to be merged, however I'd like if I could get people to test it. Someone will likely also need to rebuild binaries for a lot of the modified packages like `git` and `rust`. I'd also appreciate a couple other testers.

Lastly, someone needs to modify the chromebrew binary to source the current shell's rc file between when the packages are merged from `CREW_DEST_PREFIX` to `CREW_PREFIX` and `self.postinstall`.

install.sh
```diff
@@ -16,6 +16,7 @@ CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
 CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 CURL="${CURL:-curl}"
 
+# EARLY_PACKAGES cannot depend on crew_profile_base
```
uberhacker, line 19: Isn't git considered an early package?

saltedcoffii: Yes, but only git's bash completion depends on crew-profile-base, not git itself.

saltedcoffii:
Intel media driver and sommelier seem broken. anybody wanna troubleshoot? I can't figure it out.

saltedcoffii:
Alright it's ready, we just need binaries for some packages. I'm making some right now for x86_64.

saltedcoffii:
Binaries for x86_64:
https://mega.nz/folder/7OQlAAoB#LkmmkNSjV8KzOXlvnY9shA

PR #5484, opened 2021-03-25
Remove extra dependencies in firefox

uberhacker:
All other dependencies come along for the ride with gtk2.  Just tested on a fresh install.


satmandu:
Should we expand core to include all of these?

uberhacker:
If anything, we should try to contract core instead of expand. :)

PR #5488, opened 2021-03-25
New Package: Browsh

supechicken666:
browsh: A fully-modern text-based browser, rendering to TTY and browsers

Tested on `x86_64`

### DIFF
--- a/packages/browsh.rb
+++ b/packages/browsh.rb
@@ -1,0 +1,35 @@
+require 'package'
+
+class Browsh < Package
+  description 'A fully-modern text-based browser, rendering to TTY and browsers.'
+  homepage 'https://www.brow.sh'
+  @_ver = '1.6.4'
+  version @_ver
+  license 'LGPL-2.1'
+  compatibility 'all'
+
+  depends_on 'firefox'
+    
+  case ARCH
+  when 'armv7l', 'aarch64'
+    @_arch = 'armv7'
+  when 'x86_64'
+    @_arch = 'amd64'
+  when 'i686'
+    @_arch = '386'
+  end
+    
+  @_file = "browsh_#{@_ver}_linux_#{@_arch}"
+  @_url = "https://github.com/browsh-org/browsh/releases/download/v#{@_ver}/"
+
+  source_url @_url + @_file
+  
+  def self.preflight
+    source_sha256 `curl -Ls #{@_url}/browsh_#{@_ver}_checksums.txt`.scan(/(.*)#{@_file}$/)[0][0].strip
+  end
+
+  def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin/"
+    FileUtils.install "../#{@_file}", "#{CREW_DEST_PREFIX}/bin/browsh", mode: 0755
+  end
+end
--- a/packages/browsh.rb
+++ b/packages/browsh.rb
@@ -1,7 +1,7 @@
 require 'package'
 
 class Browsh < Package
-  description 'A fully-modern text-based browser, rendering to TTY and browsers.'
+  description 'A fully-modern text-based browser with Firefox backend, rendering to TTY and browsers.'
   homepage 'https://www.brow.sh'
   @_ver = '1.6.4'
   version @_ver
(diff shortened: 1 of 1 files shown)


packages/browsh.rb
```diff
@@ -0,0 +1,35 @@
...
+    @_arch = '386'
+  end
+    
+  @_file = "browsh_#{@_ver}_linux_#{@_arch}"
+  @_url = "https://github.com/browsh-org/browsh/releases/download/v#{@_ver}/"
+
+  source_url @_url + @_file
+  
+  def self.preflight
+    source_sha256 `curl -Ls #{@_url}/browsh_#{@_ver}_checksums.txt`.scan(/(.*)#{@_file}$/)[0][0].strip
```
uberhacker, line 28: Do we need to put this in preflight?

packages/browsh.rb
```diff
@@ -0,0 +1,35 @@
...
+
+class Browsh < Package
+  description 'A fully-modern text-based browser, rendering to TTY and browsers.'
+  homepage 'https://www.brow.sh'
+  @_ver = '1.6.4'
+  version @_ver
+  license 'LGPL-2.1'
+  compatibility 'all'
+
+  depends_on 'firefox'
```
uberhacker, line 11: We need another browser to use this?

satmandu: This program is literally a text-mode proxy to a firefox instance.

uberhacker: Maybe that should be in the description?  It currently appears to be a browser.  Sounds more like a Firefox addon.

packages/browsh.rb
saltedcoffii: I feel like this could just go in the normal source_sha256 spot.

packages/browsh.rb
saltedcoffii: One could consider it a TUI browser that uses firefox as a backend.

packages/browsh.rb
```diff
@@ -0,0 +1,35 @@
+require 'package'
+
+class Browsh < Package
+  description 'A fully-modern text-based browser, rendering to TTY and browsers.'
+  homepage 'https://www.brow.sh'
+  @_ver = '1.6.4'
+  version @_ver
+  license 'LGPL-2.1'
+  compatibility 'all'
```
saltedcoffii, line 9: iirc `firefox` is only for x86_64 and i686 the way we have it packaged.

saltedcoffii: If anyone wants to build firefox from the source on arm machines and write that into the rb file, that would be nice for browsh and for regular firefox on arm purposes.

packages/browsh.rb
supechicken: It is for speeding up actions which doesn't need to download (search, update)

uberhacker: I'm not following.  Shouldn't those operations skip it anyway?

packages/browsh.rb
supechicken: I think we don't need to consider it (my purpose for adding this package is to provide a convenient way for user to install)

packages/browsh.rb
supechicken: Although they will skip the download part (and checksum), the content inside the backtick will still execute by ruby

uberhacker: Why not just make it static?

packages/browsh.rb
uberhacker: Building firefox from source is not a nimble project.

packages/browsh.rb
uberhacker: @supechicken: I think we better make all checksums static going forward.  I see what you mean when the checksum is executed during `crew search`.  This happens on the require line in `crew` as shown below:
```
def set_package (pkgName, pkgPath)
  require pkgPath
```

supechicken: Change made

packages/browsh.rb
uberhacker: This package will not allow arm users to install until the firefox package supports arm.  As dependencies go, so goes the parent package.  Should we just remove arm support for now?

PR #5489, opened 2021-03-25
Add awka

saltedcoffii:
Awk to C compiler


saltedcoffii:
Nevermind, the upstream package seems broken enough that I'm worried it will confuse Chromebrew users. If upstream gets better, I'll reopen.

PR #5490, opened 2021-03-25
libcap, mesa, libva update

satmandu:
Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/libva.rb
+++ b/packages/libva.rb
@@ -1,26 +1,26 @@
 require 'package'
 
 class Libva < Package
-  description 'Libva is an implementation for VA-API (VIdeo Acceleration API)'
+  description 'Libva is an implementation for VA-API (Video Acceleration API)'
   homepage 'https://01.org/linuxmedia'
-  @_ver = '2.10.0'
-  version "#{@_ver}-2"
+  @_ver = '2.11.0'
+  version @_ver
   license 'MIT'
   compatibility 'all'
-  source_url "https://github.com/intel/libva/releases/download/#{@_ver}/libva-#{@_ver}.tar.bz2"
-  source_sha256 'fa81e35b50d9818fce5ec9eeeeff08a24a8864ceeb9a5c8e7ae4446eacfc0236'
+  source_url "https://github.com/intel/libva/archive/refs/tags/#{@_ver}.tar.gz"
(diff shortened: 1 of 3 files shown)


packages/libva.rb
```diff
@@ -1,26 +1,26 @@
...
   homepage 'https://01.org/linuxmedia'
-  @_ver = '2.10.0'
-  version "#{@_ver}-2"
+  @_ver = '2.11.0'
+  version @_ver
   license 'MIT'
   compatibility 'all'
-  source_url "https://github.com/intel/libva/releases/download/#{@_ver}/libva-#{@_ver}.tar.bz2"
-  source_sha256 'fa81e35b50d9818fce5ec9eeeeff08a24a8864ceeb9a5c8e7ae4446eacfc0236'
+  source_url "https://github.com/intel/libva/archive/refs/tags/#{@_ver}.tar.gz"
```
uberhacker, line 10: What's wrong with the previous release tag?

packages/mesa.rb
```diff
@@ -29,22 +29,28 @@ class Mesa < Package
...
+  depends_on 'libxrandr'
   depends_on 'libxshmfence'
   depends_on 'libxv'
   depends_on 'libxvmc'
   depends_on 'libxxf86vm'
+  depends_on 'lm_sensors'
   depends_on 'valgrind'
   depends_on 'vulkan_headers' => :build
   depends_on 'vulkan_icd_loader'
+  depends_on 'wayland'
```
uberhacker, line 46: I'm going to go out on a limb and say I imagine all these extra depends_on lines aren't needed and come along for the ride with other dependencies.  No need to list?

packages/libva.rb
satmandu: They haven't generated a bz2 file on their release page.

packages/mesa.rb
satmandu: I changed everything not showing up as a runtime dep to be build deps. Hopefully that should reduce installation size a little bit without breaking anything. 

We should do the same for all of the other X11 libraries, as some of the older ones may not be used by newer packages.

PR #5491, opened 2021-03-25
sommelier refactor, add xxd_standalone, update pipewire

satmandu:
- add `xxd_standalone`
- `sommelier` refactor: replace 7Mb `vim` dependency with 15k `xxd_standalone` dependency
- update pipewire

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -41,11 +41,11 @@ class Sommelier < Package
 
   case ARCH
   when 'armv7l', 'aarch64'
-    PEER_CMD_PREFIX = '/lib/ld-linux-armhf.so.3'.freeze
+    @peer_cmd_prefix = '/lib/ld-linux-armhf.so.3'
   when 'i686'
-    PEER_CMD_PREFIX = '/lib/ld-linux-i686.so.2'.freeze
+    @peer_cmd_prefix = '/lib/ld-linux-i686.so.2'
   when 'x86_64'
-    PEER_CMD_PREFIX = '/lib64/ld-linux-x86-64.so.2'.freeze
+    @peer_cmd_prefix = '/lib64/ld-linux-x86-64.so.2'
   end
 
   def self.prebuild
@@ -81,7 +81,7 @@ class Sommelier < Package
         -Dxwayland_path=#{CREW_PREFIX}/bin/Xwayland \
         -Dxwayland_gl_driver_path=/usr/#{ARCH_LIB}/dri -Ddefault_library=both \
         -Dxwayland_shm_driver=noop -Dshm_driver=noop -Dvirtwl_device=/dev/null \
-        -Dpeer_cmd_prefix=\"#{CREW_PREFIX}#{PEER_CMD_PREFIX}\" \
+        -Dpeer_cmd_prefix=\"#{CREW_PREFIX}#{@peer_cmd_prefix}\" \
         builddir"
       system 'meson configure builddir'
       system 'ninja -C builddir'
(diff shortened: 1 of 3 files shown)


packages/sommelier.rb
```diff
@@ -3,92 +3,93 @@
 class Sommelier < Package
   description 'Sommelier works by redirecting X11 programs to the built-in ChromeOS Exo Wayland server.'
   homepage 'https://chromium.googlesource.com/chromiumos/platform2/+/HEAD/vm_tools/sommelier/'
-  version '20210109'
-  license '' # Can't find license. Probably BSD-Google or BSD-3
+  version '20210109-1'
+  license 'BSD-Google'
```
uberhacker, line 7: Nice, we needed a license for this package.

packages/sommelier.rb
```diff
@@ -3,92 +3,93 @@
...
-        i686: 'e280ae66333a01f6406333aa6a7836b6d2638ca25aeb81378f55e1bd9df9f339',
-      x86_64: '5348afbfceeed2ddf5078d4d325f50ad736ea89c8249db263009c3d23cd05db9',
+  binary_sha256({
+    aarch64: 'c076c1f99a018eb4c75de3768445e322bcfbc2221942b967a928924ec798e6b0',
+     armv7l: 'c076c1f99a018eb4c75de3768445e322bcfbc2221942b967a928924ec798e6b0',
+       i686: '98d5f94ef0f633da694b1525d1e01ad8ddf82fbb14996c52eec6b7ef51f5fed7',
+     x86_64: '45cf6f34a6fc6bd1bb5238eba9fb7dfbf5e528c6496dcad2d9048aeec4ac20ed'
   })
 
+  depends_on 'coreutils' # for readlink in wrapper script
```
uberhacker, line 25: Yikes, has this been updated recently?  It could have disastrous results.

packages/sommelier.rb
satmandu: License from https://chromium.googlesource.com/chromiumos/platform2/+/refs/heads/main/LICENSE

Which appears to be the same as 

https://developers.google.com/open-source/licenses/bsd

packages/sommelier.rb
```diff
@@ -3,92 +3,93 @@
...
-   when 'i686'
-      PEER_CMD_PREFIX='/lib/ld-linux-i686.so.2'
-   when 'x86_64'
-      PEER_CMD_PREFIX='/lib64/ld-linux-x86-64.so.2'
+  when 'armv7l', 'aarch64'
+    PEER_CMD_PREFIX = '/lib/ld-linux-armhf.so.3'.freeze
+  when 'i686'
+    PEER_CMD_PREFIX = '/lib/ld-linux-i686.so.2'.freeze
+  when 'x86_64'
+    PEER_CMD_PREFIX = '/lib64/ld-linux-x86-64.so.2'.freeze
```
uberhacker, line 48: Do we have to use constants?  Wouldn't an `@` variable remove the `.freeze` restraint?

packages/sommelier.rb
satmandu: yup. 8.32 is the current version: https://github.com/skycocker/chromebrew/blob/master/packages/coreutils.rb

packages/sommelier.rb
satmandu: I don't know if one is better than the other, but this should work fine w/ the @ variables... Let me change that.

packages/xxd_standalone.rb
```diff
@@ -0,0 +1,44 @@
+require 'package'
+
+class Xxd_standalone < Package
+  description 'Hexdump utility from vim'
+  homepage 'http://www.vim.org'
+  @_ver = '8.2.2580'
+  version @_ver
+  license 'GPL-2'
+  compatibility 'all'
+  source_url 'https://github.com/vim/vim/archive/v8.2.2580.tar.gz'
```
uberhacker, line 10: Should this be updated with all the other vim packages to keep the versions consistent?

satmandu: It is the same version as the other vim versions... Do you see different versions?

satmandu: It is the same version as the other vim versions... Do you see different versions?

PR #5493, opened 2021-03-25
unixodbc and freetds update

satmandu:
- freetds was build with Libressl, this is now built with OpenSSL.
- also updated unixodbc dependency

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/unixodbc.rb
+++ b/packages/unixodbc.rb
@@ -22,10 +22,6 @@ class Unixodbc < Package
      x86_64: '705488db5fc32fe2062da25e4650ac090983404525c9d06f32646058a76cfb2a'
   })
   
-  depends_on 'libtool'
-  depends_on 'glibc'
-  depends_on 'readline'
-  
   def self.build
     system "env CFLAGS='-pipe -fno-stack-protector -U_FORTIFY_SOURCE -flto=auto' \
       CXXFLAGS='-pipe -fno-stack-protector -U_FORTIFY_SOURCE -flto=auto' \
--- a/packages/unixodbc.rb
+++ b/packages/unixodbc.rb
@@ -3,34 +3,39 @@ require 'package'
 class Unixodbc < Package
   description 'ODBC is an open specification for providing application developers with a predictable API with which to access Data Sources.'
   homepage 'http://www.unixodbc.org/'
-  version '2.3.7'
+  version '2.3.9'
   license 'LGPL-2+'
   compatibility 'all'
-  source_url 'http://www.unixodbc.org/unixODBC-2.3.7.tar.gz'
-  source_sha256 '45f169ba1f454a72b8fcbb82abd832630a3bf93baa84731cf2949f449e1e3e77'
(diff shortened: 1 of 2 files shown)


packages/unixodbc.rb
```diff
@@ -3,34 +3,39 @@
...
+    aarch64: '6f04ee2e7fcc4e8c37db45e05fe51a5dcac345b7c61d440b791762ae7609b578',
+     armv7l: '6f04ee2e7fcc4e8c37db45e05fe51a5dcac345b7c61d440b791762ae7609b578',
+       i686: '590762ad9a2cd402ae52da0382c652d8e08a6b049f875ae99ec792f228331cd8',
+     x86_64: '705488db5fc32fe2062da25e4650ac090983404525c9d06f32646058a76cfb2a'
   })
-
+  
+  depends_on 'libtool'
+  depends_on 'glibc'
+  depends_on 'readline'
```
uberhacker, line 27: All core packages.  Let's not start listing everything.  These packages should always be installed.

satmandu: I can take them out, but I also recall you wanted to slim that core list down...

packages/unixodbc.rb
```diff
@@ -22,10 +22,6 @@ class Unixodbc < Package
      x86_64: '705488db5fc32fe2062da25e4650ac090983404525c9d06f32646058a76cfb2a'
   })
   
-  depends_on 'libtool'
-  depends_on 'glibc'
-  depends_on 'readline'
-
```
uberhacker: If we trimmed core, it wouldn't be these packages. :)

packages/unixodbc.rb
```diff
@@ -3,34 +3,35 @@
...
-       i686: '987033a506015196dee6951e25d0860a23c5deadf74068590be5b913d1bfca3d',
-     x86_64: 'ecae202a4367725ceb7124b374f5e80335cfde85b6cdc296beaff735ddea3331',
+  binary_sha256({
+    aarch64: '6f04ee2e7fcc4e8c37db45e05fe51a5dcac345b7c61d440b791762ae7609b578',
+     armv7l: '6f04ee2e7fcc4e8c37db45e05fe51a5dcac345b7c61d440b791762ae7609b578',
+       i686: '590762ad9a2cd402ae52da0382c652d8e08a6b049f875ae99ec792f228331cd8',
+     x86_64: '705488db5fc32fe2062da25e4650ac090983404525c9d06f32646058a76cfb2a'
   })
-
+
```
uberhacker, line 24: I thought we were trying to get rid of trailing whitespace?

packages/unixodbc.rb
satmandu: :)

packages/unixodbc.rb
satmandu: Weird.... That will get removed on the next cleanup...

PR #5494, opened 2021-03-27
Revert CREW_BUILD autodetect

saltedcoffii:
So it looks like automake's `config.guess` script is incapable of reporting a triple with `cros` as the vendor bytes. Apologies for the issues with compiling everyone <3

I may decide to modify the `config.guess` script so it works on chromebooks but for now this is what we have to do. :(

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.7.17'
+CREW_VERSION = '1.7.18'
 
 ARCH_ACTUAL = `uname -m`.strip
 # This helps with virtualized builds on aarch64 machines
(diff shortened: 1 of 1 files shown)


lib/const.rb
```diff
@@ -63,13 +63,15 @@
 case ARCH
 when 'aarch64', 'armv7l'
   CREW_TGT = 'armv7l-cros-linux-gnueabihf'
+  CREW_BUILD = 'armv7l-cros-linux-gnueabihf'
 when 'i686'
   CREW_TGT = 'i686-cros-linux-gnu'
+  CREW_BUILD = 'i686-cros-linux-gnu'
 when 'x86_64'
   CREW_TGT = 'x86_64-cros-linux-gnu'
+  CREW_BUILD = 'x86_64-cros-linux-gnu'
```
uberhacker, line 72: So now we have 2 constants that are the same?  Can we just revert back to only one?

saltedcoffii: I do want to leave because I'm planning on filing an upstream bug with GNU automake about the issue. This is just temporary ;)

PR #5495, opened 2021-03-28
Symbols are not quoted

uberhacker:


satmandu:
Is there a rubocop option which keeps these from being added?

PR #5496, opened 2021-03-28
[Upgrade] fop 2.2 => 2.6

saltedcoffii:
This package doesn't really need binaries since the source itself is a binary. Chromebrew rearranges the binary in less than 7 seconds on my machine.

### DIFF
--- a/packages/fop.rb
+++ b/packages/fop.rb
@@ -6,7 +6,7 @@ class Fop < Package
   version '2.6'
   license 'Apache-2.0'
   compatibility 'all'
-  source_url 'https://mirror.olnevhost.net/pub/apache/xmlgraphics/fop/binaries/fop-2.6-bin.tar.gz'
+  source_url 'https://downloads.apache.org/xmlgraphics/fop/binaries/fop-2.6-bin.tar.gz'
   source_sha256 'ccfd7a1d4e5a04e76723946efa1147ffa9a8715ce2b58d2a27085a8e744520f8'
 
   depends_on 'jdk8'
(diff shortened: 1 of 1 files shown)


packages/fop.rb
```diff
@@ -3,24 +3,11 @@
...
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/fop-2.2-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/fop-2.2-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: '5ae3c47bba798fdc6a0eec6b613fb9f25d6015b27adb5845ab90ce46b16be0da',
-     armv7l: '5ae3c47bba798fdc6a0eec6b613fb9f25d6015b27adb5845ab90ce46b16be0da',
-       i686: '06c576e9fe57ff1efec61ae76a9f82c34f5def02c658e469866ad6256ef79ef5',
-     x86_64: '84e05f40cee9b05976efe47579718b3eb6ffa5080e65a3aa98a558a8b46e1934',
-  })
+  source_url 'https://mirror.olnevhost.net/pub/apache/xmlgraphics/fop/binaries/fop-2.6-bin.tar.gz'
```
uberhacker, line 9: Why the source domain change?

saltedcoffii: The old domain kept saying '404 not found'.

uberhacker: Hmmm, works for me.

uberhacker: Does https://downloads.apache.org/xmlgraphics/fop/binaries/fop-2.6-bin.tar.gz work?

PR #5506, opened 2021-03-30
fix beav

saltedcoffii:
I don't know why beav was using binaries, but my guess is that the person packaging it didn't know how to fix the C code. This is a patched and working version of beav's source code that doesn't require a libtinfo symlink.


packages/beav.rb
```diff
@@ -3,50 +3,30 @@
...
-    system "tar xvf beav-1.40.tgz"
+    system 'make'
   end
 
   def self.install
-    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}"
-    FileUtils.mv 'usr/bin', "#{CREW_DEST_PREFIX}"
-    FileUtils.mv 'usr/share', "#{CREW_DEST_PREFIX}"
+    system "install -Dm 755 beav #{CREW_DEST_PREFIX}/bin/beav"
+    system "install -Dm 755 beav.1 #{CREW_DEST_MAN_PREFIX}/man1/beav.1"
```
uberhacker, line 30: No `FileUtils.install...`?

packages/beav.rb
```diff
@@ -3,50 +3,30 @@
...
+  def self.patch
+    system "curl -#L https://httpredir.debian.org/debian/pool/main/b/beav/beav_1.40-18.diff.gz | gunzip >> beav_1.40-18.diff"
+    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read('beav_1.40-18.diff') ) == '5139aa0bdc9424b7ad8197070e1902d8225ee78e2091d02a493422cd91fe5973'
+    system "patch -Np1 -i beav_1.40-18.diff"
+    system "sed -i 's:-lncurses:-ltinfow:g' Makefile"
+    system "sed -i '12d' basic.c"
+    system "sed -i '13d' file.c"
+    system "sed -i '76d' random.c"
+    system "sed -i '11d' symbol.c"
+    system "sed -i 's:#include <term.h>:#include <ncursesw/term.h>:' tcap.c"
```
uberhacker, line 21: These look like destructive changes.

satmandu:
@saltedcoffii If you have a new spin on this package, I can recompile & get new binaries for you...

packages/beav.rb
saltedcoffii: it looks like they are, but I know C, trust me they aren't. This package hasn't been updated since 2004, these are just things that used to be okay, but in newer versions of gcc they've been phased out.

saltedcoffii:
Those binaries look good, it looks ready to merge to me!

packages/beav.rb
```diff
@@ -42,7 +42,9 @@ def self.build
   end
 
   def self.install
-    system "install -Dm 755 beav #{CREW_DEST_PREFIX}/bin/beav"
-    system "install -Dm 755 beav.1 #{CREW_DEST_MAN_PREFIX}/man1/beav.1"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
+    FileUtils.mkdir_p "#{CREW_DEST_MAN_PREFIX}/man1/"
+    FileUtils.install 'beav', "#{CREW_DEST_PREFIX}/bin/beav", mode: 0755
+    FileUtils.install 'beav.1' "#{CREW_DEST_MAN_PREFIX}/man1/beav.1", mode: 0755
```
uberhacker, line 48: Manpages don't need to be executable.  Use `mode: 0644`.

PR #5509, opened 2021-03-30
Switch to ncurses wide character only

saltedcoffii:
The current ncurses package build `ncurses`, then `ncursesw`. The `ncursesw` binaries overwrite `ncurses`. Having both binaries on the system means that, sometimes, a source package will get confused when trying to link to one of them and raise an error saying: `error adding symbols: DSO missing from command line`.

The regular ncurses libraries and the wide ncurses libraries are source compatible, not binary compatible, meaning that every single package that links to `libform.so.6`, `libmenu.so.6`, `libncurses++.so.6`, `libncurses.so.6`, `libpanel.so.6` or `libtinfo.so.6` will need to be recompiled so they link to `libformw.so.6`, `libmenuw.so.6`, libncursesw++.so.6`, `libncursesw.so.6`,  `libpanelw.so.6` or `libtinfow.so.6` respectively. 

To avoid recompiling all of chromebrew's binaries, there is a relatively easy way to see if a binary links against one of the regular ncurses libraries. run `ldd <binary>` and if the library shows up, its entire package needs to be recompiled. To do this quickly, one can run `pushd /usr/local/bin && find . -type f -exec ldd {} \; | grep -B15 <library> && popd`. Any packages that display information to the CLI without stdout are especially suspect for using ncurses and should be checked in this manner. Examples include htop, cmatrix, a lot of the bsdgames, etc.

There are also a list of package files I had to directly modify to support ncurses wide libraries. These all need to be recompiled. That list is:
```
aspell_en
bash
biew
binclock
bmon
cdargs
ckermit
cmake
cmatrix
cpustat
enchant
eventstat
faultstat
finch
fish
gawk
gettext
goaccess
gptfdisk
gtypist
hexedit
htop
hunspell_base
iftop
inetutils
iptraf
js78
kakoune
ldc
less
libcaca
libedit
libiconv
libstfl
mdp
mg
moonbuggy
multitail
ncdc
ncdu
nethogs
nftables
nmon
pagemon
parted
percona_server
patchelf
pinentry
php73
procps
progress
psmisc
python27
python3
qttools
ranger
ruby
screen
scrollz
shadow
sl
smemstat
sngrep
swi_prolog
tack
testdisk
texinfo
tig
tina
tmux
usbutils
zile
```

Here is a list of packages I've identified via the ldd method above:
```
binutils (nm)
krb5
llvm
sqlite
```

I'm really going to need help from you guys because I can't compile any non-x86_64 binaries unless I can figure out how to virtualize ChromeOS on a rock wrapped in aluminium foil.

Some packages may not compile because they don't innately support wide ncurses. It's usually just a problem having to do with header file location or a linker argument, but if that's the case, let me know and I'll provide patches.


satmandu:
What do other distributions do?

saltedcoffii:
Debian, Ubuntu and Arch use both, Fedora, Cent, RHEL, Alpine and Linux From Scratch use wide only, and Gentoo leaves it up to the user, defaulting to wide only.

uberhacker:
This is a very ambitious endeavor.

satmandu:
With this many packages, maybe we should hold off until we get the new upload location up, so we just upload new builds of these to the new site?

saltedcoffii:
Sounds good to me! I'll make a mega folder as I usually do for x86_64 binaries so you guys only have 2 architectures to compile for.

saltedcoffii:
@uberhacker @satmandu now that we're recompiling all binaries with glibc 2.28, could we recompile them with ncursesw support as well so we can close this?

uberhacker:
glibc 2.28 will not be needed until release 90 is stable and only with x86_64 is my understanding.  We can cross that bridge when we get there.

PR #5510, opened 2021-03-30
create nautilus bookmarks file.

satmandu:
- stops console error from missing file appearing.
- don't want to do this in install, since you don't want to overwrite on reinstall.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/nautilus.rb
+++ b/packages/nautilus.rb
@@ -78,5 +78,6 @@ class Nautilus < Package
   
   def self.postinstall
     system "glib-compile-schemas #{CREW_PREFIX}/share/glib-2.0/schemas/"
+    FileUtils.touch "#{CREW_DEST_HOME}/.gtk-bookmarks"
   end
 end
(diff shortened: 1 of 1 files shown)


packages/nautilus.rb
```diff
@@ -78,5 +78,6 @@ def self.install
   
   def self.postinstall
     system "glib-compile-schemas #{CREW_PREFIX}/share/glib-2.0/schemas/"
+    FileUtils.touch "#{CREW_DEST_HOME}/.gtk-bookmarks"
```
uberhacker, line 81: Shouldn't this be `FileUtils.touch "#{HOME}/.gtk-bookmarks"`?  You would use `CREW_DEST_HOME` for packaged files in `def self.install`.

satmandu: Ah yes. I copied from slack.

satmandu: Fixed.

PR #5511, opened 2021-03-31
delete openpam

saltedcoffii:
ditch openpam. I can't get enlightenment to compile and nothing uses it as a dependency.


uberhacker:
@ThatGeekyWeeb:  What are your thoughts of nixing enlightenment?  Were you able to get it to work?

ThatGeekyWeeb:
It did at one point work, it's just very impractical to use, since it requires running it inside of xephyr inside of sommelier, which is very very slow for anything other then terminals (nested X is slow), I can't test the build as I don't have access to ChromeOS, but if @saltedcoffii can't get it to compile, it's likely broken, and since nothing else uses it, you can go ahead and drop it @uberhacker

It was just a POC that one could use a full DE in ChromeOS

saltedcoffii:
@ThatGeekyWeeb i'm happy to try to work more on enlightenment if it means that much.

PR #5512, opened 2021-03-31
remove most core package

saltedcoffii:
This set of fixes removes the need for most as a pager and in core packages. Let's face it: most is extra, and no linux distros use most as their default pager. They all opt for GNU less, busybox less, or GNU more. Using most as the default pager would be like using exa instead of ls by default. It's unnecessary.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -48,7 +48,7 @@ esac
 crew_folders="bin cache doc docbook etc include lib lib$LIB_SUFFIX libexec man sbin share tmp var"
 for folder in $crew_folders
 do
-  if [ -d "${CREW_PREFIX}"/"${folder}" ]; then 
+  if [ -d "${CREW_PREFIX}"/"${folder}" ]; then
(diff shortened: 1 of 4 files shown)


uberhacker:
Again, why do we need to be like everybody else?  Nothing wrong with giving the user the option to choose which pager they would like and that is what we do at the end of the install.  I'm thinking about reducing core in a big way.  I just might have something ready by EoW.

saltedcoffii:
I think we can, but it shouldn't be installed by default. Instead, the user can install it like any other package. My concern is that some chromebooks, especially older ones have extremely limited storage space and we don't want to have more installed by default than need be.

uberhacker:
What if we have a prompt to choose the pager in install.sh and install less or most or even nothing if users are fine with more (which is available out-of-the-box)?

uberhacker:
It would be better to prompt for the desired pager during install and only install the packages necessary for the option chosen.

PR #5503, opened 2021-03-30
Crew profile base

saltedcoffii:
For some reason the old pull request just went missing when I deleted and recreated the env-d branch to undo some commits surrounding #5497. This is a new pr but all the same stuff.

PR #5516, opened 2021-04-01
Add basic setup for Chrome OS Exo server

supechicken666:
Environment variables to use Exo Wayland server

Also added a meta-package to select between `sommelier` and `exo`

Tested on `x86_64`


packages/exo_base.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      CLUTTER_BACKEND=wayland
+      GDK_BACKEND=wayland
+
+      set +a
+    EOF
+  end
+
+  def self.install
+    system "echo '#{@env}' >> #{HOME}/.bashrc" unless `grep -c '#{@env.lines.first.chomp}' #{HOME}/.bashrc`.to_i > 0
```
uberhacker, line 36: Shouldn't this go in `def self.postinstall` since it's not in the packaged files?

packages/gui_setup.rb
```diff
@@ -0,0 +1,40 @@
...
+        depends_on 'exo_base'
+        break
+      when '2'
+        depends_on 'sommelier'
+        break
+      else
+        puts "\nPlease select from one of the options or enter 0 to cancel.".yellow
+      end
+    end
+  end
```
uberhacker, line 39: I'm thinking this entire preflight section should be wrapped in `if ARGV[0] == 'install' ... end`, otherwise upgrade operations will stop and prompt unnecessarily.

uberhacker: Actually, upgrade operations should ignore `is_fake` packages.

uberhacker:
Thank you for the new packages @supechicken!  This looks really useful.

packages/gui_setup.rb
satmandu: This might be the wrong place to ask, but is there a way to enable a setting here through a command line flag? If this is a dependency, and I'm running a compile for instance, I don't want it to hang waiting for input.

(This is also an issue with hunspell.)

uberhacker: We have the same issue with php.  I think it's a crew bug.  We should ignore `is_fake` packages during upgrades.  Maybe we should file an issue in the queue.

packages/exo_base.rb
supechicken: Done

packages/gui_setup.rb
supechicken: @satmandu How about setting a timeout for 5 seconds and making `sommelier` as default?

saltedcoffii:
@supechicken, hold this pull request until #5503 gets merged. I wanna make some changes to these scripts as they attempt to edit ~/.bashrc directly.

packages/gui_setup.rb
uberhacker: I think the main issue here is there is no need to "upgrade" is_fake packages.  The only packages that need upgrades are the dependencies.  The upgrade process should completely ignore is_fake packages because there is never a need to do it.

uberhacker: PR #5528 should resolve this issue?  No need to change anything here.

packages/thunar.rb
```diff
@@ -3,10 +3,11 @@
 class Thunar < Package
   description 'Thunar File Manager'
   homepage 'https://docs.xfce.org/xfce/thunar/Start'
-  version '4.16.1'
+  @_ver = '4.16.1'
+  version @_ver + '-1'
```
uberhacker, line 7: New version?  See https://archive.xfce.org/src/xfce/thunar/4.17/thunar-4.17.1.tar.bz2.

packages/exo_base.rb
```diff
@@ -0,0 +1,52 @@
...
+    EOF
+  end
+
+  def self.install
+    FileUtils.mkdir_p("#{CREW_DEST_PREFIX}/etc/env.d/")
+    #File.write("#{CREW_DEST_PREFIX}/etc/env.d/exo_base", @env)
+  end
+
+  def self.postinstall
+      system "echo '#{@env}' >> #{HOME}/.bashrc" unless `grep -c '#{@env.lines.first.chomp}' #{HOME}/.bashrc`.to_i > 0
```
uberhacker, line 42: Change `.to_i > 0` to `.positive?`?

supechicken: Done

packages/exo_base.rb
```diff
@@ -0,0 +1,52 @@
...
+      CLUTTER_BACKEND=wayland
+      GDK_BACKEND=wayland
+
+      set +a
+    EOF
+  end
+
+  def self.install
+    FileUtils.mkdir_p("#{CREW_DEST_PREFIX}/etc/env.d/")
+    #File.write("#{CREW_DEST_PREFIX}/etc/env.d/exo_base", @env)
```
uberhacker, line 38: Are we not installing `@env`?

supechicken: The `@env` variable is written to `bashrc` now, this line is for #5503

packages/exo_base.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def self.postinstall
+      system "echo '#{@env}' >> #{HOME}/.bashrc" unless `grep -c '#{@env.lines.first.chomp}' #{HOME}/.bashrc`.to_i.positive?
+
+      puts <<~EOT.lightblue
```
uberhacker, line 44: Shouldn't we be putting environment stuff in `"#{CREW_PREFIX}/etc/env.d"` now?

saltedcoffii: Yes.

satmandu:
So we need env.d changes, and usage of #{CREW_ENV_OPTIONS}?

satmandu:
Also, could we add the Environment variables to the `wayland` package so that it is alphabetically after `sommelier`, so that its environment variables override sommelier 's?

saltedcoffii:
> Also, could we add the Environment variables to the `wayland` package so that it is alphabetically after `sommelier`, so that its environment variables override sommelier 's?

I'm confused by what you mean, could you be more specific?

supechicken:
> Also, could we add the Environment variables to the `wayland` package so that it is alphabetically after `sommelier`, so that its environment variables override sommelier 's?

Good idea, moving to all stuff `wayland.rb`...

packages/wayland.rb
```diff
@@ -39,5 +39,7 @@ def self.build
 
   def self.install
     system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
+    FileUtils.mkdir_p("#{CREW_DEST_PREFIX}/etc/env.d/")
+    File.write("#{CREW_DEST_PREFIX}/etc/env.d/exo_base", @env)
```
uberhacker, line 43: Would `wayland` be a better name than `exo_base`?

supechicken: Changed to `wayland`

PR #5518, opened 2021-04-01
[New Package] samurai

saltedcoffii:
Samurai is basically ninja rewritten in C. I'll probably go through and convert all the package files to samurai since it does build faster than ninja does.

### DIFF
--- a/packages/samurai.rb
+++ b/packages/samurai.rb
@@ -1,0 +1,26 @@
+require 'package'
+
+class Samurai < Package
+  description 'Samurai is a ninja compatible build tool written in C.'
+  homepage 'https://github.com/michaelforney/samurai'
+  version '1.2'
+  license 'Apache-2.0 and MIT'
+  compatibility 'all'
+  source_url 'https://github.com/michaelforney/samurai/releases/download/1.2/samurai-1.2.tar.gz'
+  source_sha256 '3b8cf51548dfc49b7efe035e191ff5e1963ebc4fe8f6064a5eefc5343eaf78a5'
+
+  def self.patch
+    system "sed -i 's:PREFIX=/usr/local:PREFIX=#{CREW_PREFIX}:' Makefile"
+    system "sed -i 's:MANDIR=\$(PREFIX)/share/man:MANDIR=#{CREW_MAN_PREFIX}:g' Makefile"
+  end
+
+  def self.build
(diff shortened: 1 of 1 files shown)


satmandu:
I would hesitate on converting all packages. Speed is nice, but if software is packaged to be built with certain tools, it helps when debugging to be building with those tools... If you can get package upstream to support samurai... then that's another issue.

packages/samurai.rb
```diff
@@ -0,0 +1,26 @@
...
+  homepage 'https://github.com/michaelforney/samurai'
+  version '1.2'
+  license 'Apache-2.0 and MIT'
+  compatibility 'all'
+  source_url 'https://github.com/michaelforney/samurai/releases/download/1.2/samurai-1.2.tar.gz'
+  source_sha256 '3b8cf51548dfc49b7efe035e191ff5e1963ebc4fe8f6064a5eefc5343eaf78a5'
+
+  def self.patch
+    system "sed -i 's:PREFIX=/usr/local:PREFIX=#{CREW_PREFIX}:' Makefile"
+    system "sed -i 's:MANDIR=\$(PREFIX)/share/man:MANDIR=#{CREW_MAN_PREFIX}:g' Makefile"
```
uberhacker, line 14: Could we set these in the build section instead?  Something like:
```
system "PREFIX=#{CREW_PREFIX} MANDIR=#{CREW_MAN_PREFIX} \
            make CFLAGS='-pipe -O2 -flto=auto -fuse-ld=gold' \
            LDFLAGS='-flto=auto -fuse-ld=gold' \
            CC='#{CREW_TGT}-gcc'"
```

saltedcoffii: Not with the way the makefile is written, no. I already tried that.

saltedcoffii:
Oh sorry, I meant samurai is a drop in replacement for ninja. Some distros, like Alpine Linux, use samurai in replacement of ninja entirely. They use the same syntax and everything, it's like using llvm-strip instead of gnu strip.

packages/samurai.rb
uberhacker: I would be surprised if it didn't work.

saltedcoffii: Becuase it's `PREFIX=/usr/local` not `PREFIX:=/usr/local` it doesn't work. Same with MANDIR

uberhacker: Ah, that makes sense now.

PR #5522, opened 2021-04-01
Samuize chromebrew packages

saltedcoffii:
This is a follow up of #5518, implementing samurai in place of ninja. I've tested a sample of all these packages and they work slightly faster than with ninja and flawlessly.

Side note: f1de079 is unrelated to this, but since util-linux uses `--with-ncursesw` instead of `--with-ncurses`, `libtinfo` is no longer a dependency.

### DIFF
--- a/packages/meson.rb
+++ b/packages/meson.rb
@@ -10,7 +10,7 @@ class Meson < Package
   source_url "https://github.com/mesonbuild/meson/archive/#{@_ver}.tar.gz"
   source_sha256 '0c043c9b5350e9087cd4f6becf6c0d10b1d618ca3f919e0dcca2cdf342360d5d'
 
-  depends_on 'ninja'
+  depends_on 'samurai'
   depends_on 'ninja'
 
   binary_url({
--- a/packages/meson.rb
+++ b/packages/meson.rb
@@ -10,6 +10,7 @@ class Meson < Package
   source_url "https://github.com/mesonbuild/meson/archive/#{@_ver}.tar.gz"
   source_sha256 '0c043c9b5350e9087cd4f6becf6c0d10b1d618ca3f919e0dcca2cdf342360d5d'
 
+  depends_on 'ninja'
   depends_on 'samu'
 
   binary_url({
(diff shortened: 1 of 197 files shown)


uberhacker:
What are your thoughts about this @satmandu?

satmandu:
It is safer than switching to libressl... But I would try it only on package updates moving forward only. It doesn't make sense to recompile existing built packages, and only submit package files built with samurai if you have confirmed they do build successfully.

saltedcoffii:
There's no need to recompile builds, it does the exact same thing. And if it would make you more comfortable, I can test and report back on every single modified package @satmandu

satmandu:
My other worry is that a package file should be a direct reflection of how the binaries were created. Changing the file after binaries are generated to use a different process gets away from that.

saltedcoffii:
okay, that makes sense. i didn't think of that. however we should probably add samu to meson's dependencies and to the core package list as we transition.

saltedcoffii:
Okay, it should be ready to merge.

packages/meson.rb
```diff
@@ -10,6 +10,7 @@ class Meson < Package
   source_url "https://github.com/mesonbuild/meson/archive/#{@_ver}.tar.gz"
   source_sha256 '0c043c9b5350e9087cd4f6becf6c0d10b1d618ca3f919e0dcca2cdf342360d5d'
 
+  depends_on 'ninja'
   depends_on 'ninja'
```
uberhacker, line 14: Double dependencies?

tools/core_packages.txt
```diff
@@ -78,6 +78,7 @@ python3
 readline
 rtmpdump
 ruby
+samurai
```
uberhacker, line 81: Maybe this shouldn't be in core.  We want to make it leaner, not fatter.

saltedcoffii: Eventually we'll remove ninja as we transition from samurai to ninja, since they're both super small, would you mind allowing them for the time being?

uberhacker: I think the goal should be to exclude all build dependencies in core.

saltedcoffii: okay, let's work on that then!

PR #5530, opened 2021-04-03
[New Package] sonic

saltedcoffii:


packages/sonic.rb
```diff
@@ -0,0 +1,25 @@
...
+  version '0.2.0'
+  license 'Apache-2.0'
+  compatibility 'all'
+  source_url 'https://github.com/espeak-ng/sonic/archive/refs/tags/release-0.2.0.tar.gz'
+  source_sha256 'c7827ce576838467590ffa1f935fbe1049e896dfed6c515cf569ad3779c24085'
+
+  def self.patch
+    system "sed -i 's:CFLAGS=-Wall:CFLAGS=-Wall -pipe -flto=auto -fuse-ld=gold:g' Makefile"
+    system "sed -i 's:PREFIX=/usr:PREFIX=#{CREW_PREFIX}:' Makefile"
+    system "sed -i 's:LIBDIR=$(PREFIX}/lib:#{CREW_LIB_PREFIX}:' Makefile"
```
supechicken, line 15: Can the `prefix` and `libdir` path change with options?
```ruby
make prefix=#{CREW_PREFIX} libdir=#{CREW_LIB_PREFIX}
```

saltedcoffii: No, unfortunately because the its `PREFIX=/usr` not `PREFIX:=/usr` it won't work. Same with libdir.

PR #5531, opened 2021-04-03
[New Package] pcaudiolib

saltedcoffii:


packages/pcaudiolib.rb
```diff
@@ -0,0 +1,30 @@
...
+  def self.build
+    system './autogen.sh'
+    system "env CFLAGS='-O2 -pipe -flto=auto -fuse-ld=gold' \
+                LDFLAGS='-flto=auto' \
+            ./configure #{CREW_OPTIONS} \
+              --without-pulseaudio \
+              --with-alsa \
+              --without-qsa \
+              --without-coreaudio \
+              --without-oss"
```
uberhacker, line 23: Why so many withouts?

saltedcoffii: Coreaudio is a macos thing, I don't know what qsa and oss are (perhaps windows, i'm not sure), alsa worked and pulseaudio had too many depenedencies for the function of the package it seemed. You could add pulseaudio as a dependency and do `--with-pulseaudio` but I chose not to.

PR #5532, opened 2021-04-03
Wine to 6.5

satmandu:
- Update to Wine 6.5

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686


packages/wine.rb
```diff
@@ -11,18 +11,20 @@ class Wine < Package
...
+    depends_on 'libgphoto'
     depends_on 'libpcap'
-    depends_on 'apulse'
+    depends_on 'libunwind'
+    depends_on 'libx11'
+    depends_on 'libxext'
     depends_on 'mpg123'
-    depends_on 'ffmpeg'
+    depends_on 'openldap'
+    depends_on 'pulseaudio'
```
uberhacker, line 27: Bump version and generate new binaries?

uberhacker: Might as well upgrade to the latest release also.

satmandu: The newest binary I generated doesn't work. So...

satmandu: Just want to get rid of possibility of `apulse` getting installed and overwriting pulse audio.

We really need a central way of preventing overwriting files from existing packages during install and throwing an error... That would solve this issue...

uberhacker: Yeah, I'm sure this package requires persistence.  There are a ton of dependencies and it only makes sense to include as much as we can.

satmandu: Updated to new version. Any thoughts on moving the .wine folder location to a XDG_CONFIG_DIR location and also setting that for wine prefix? This does work for me: `WINEPREFIX=$XDG_CONFIG_HOME/.wine wine explorer`

uberhacker: Might want to add `depends_on 'xdg_base'` to the list.

satmandu: @uberhacker Can we just add that to core?

uberhacker: Maybe but how many packages does it affect?  I'm reluctant to add gui based dependencies in core.  I want to stay away from that bloat potential.

packages/wine.rb
```diff
@@ -90,6 +96,7 @@ def self.postinstall
   end
 
   def self.remove
+    @xdg_config_home = ENV['XDG_CONFIG_HOME']
```
uberhacker, line 99: What happens when the environment variable isn't set?  Should we have a default fallback value?

packages/wine.rb
```diff
@@ -103,5 +110,18 @@ def self.remove
...
+      response = $stdin.getc
+      case response
+      when 'y', 'Y'
+        FileUtils.rm_rf "#{@xdg_config_home}/.wine"
+        puts "#{@xdg_config_home}/.wine removed.".lightred
+      else
+        puts "#{@xdg_config_home}/.wine saved.".lightgreen
+      end
+      puts
+    end
```
uberhacker, line 125: Don't we want to remove the prompt above this?

packages/wine.rb
satmandu: How about `@xdg_config_home =  "#{CREW_PREFIX}/.config" if @xdg_config_home.to_s.empty?`

uberhacker: That works.  I'd hate to recursively remove the wrong `.wine` folder. :)

packages/wine.rb
satmandu: `xdg_base` is pretty slim though. Maybe we need a `gui_core`?

packages/wine.rb
satmandu: That should only show up if a `~/.wine` folder exists, right?

uberhacker: What is the difference between `#{HOME}/.wine` and `#{@xdg_config_home}/.wine`?

packages/wine.rb
```diff
@@ -4,45 +4,69 @@
...
+  depends_on 'libusb'
+  depends_on 'libx11'
+  depends_on 'libxext'
+  depends_on 'mpg123'
+  depends_on 'openal'
+  depends_on 'pulseaudio'
+  depends_on 'sommelier'
+
+  @xdg_config_home = ENV['XDG_CONFIG_HOME']
+  @xdg_config_home = "#{CREW_PREFIX}/.config" if @xdg_config_home.to_s.empty?
```
uberhacker, line 45: How about `@xdg_config_home = empty?(ENV['XDG_CONFIG_HOME']) ? "#{CREW_PREFIX}/.config" : ENV['XDG_CONFIG_HOME']`?

satmandu: done

packages/wine.rb
uberhacker: Why not do this?
```
    config_dirs = ["#{HOME}/.wine", "#{@xdg_config_home}/.wine"]
    config_dirs.each { |config_dir|                                                                                                                                                           
      if Dir.exists? config_dir                                                                                                                                                               
         print "\nWould you like to remove #{config_dir}? [y/N] "                                                                                                                                
        case STDIN.getc                                                                                                                                                                       
        when "y", "Y"                                                                                                                                                                         
          FileUtils.rm_rf config_dir                                                                                                                                                          
          puts "#{config_dir} removed.".lightred                                                                                                                                              
        else                                                                                                                                                                                  
          puts "#{config_dir} saved.".lightgreen                                                                                                                                              
        end                                                                                                                                                                                   
      end                                                                                                                                                                                     
    }
```

satmandu: This is obviously a better solution. When I get back to a desktop I can add it in, or you can.

packages/wine.rb
```diff
@@ -54,29 +78,39 @@ def self.install
...
+      # Wine configuration
+      WINEPREFIX="${XDG_CONFIG_HOME}"/.wine
+    WINE_CONFIG_EOF
+    IO.write("#{CREW_DEST_PREFIX}/etc/env.d/wine", @wine_config_env)
   end
 
   def self.postinstall
     puts
-    puts "Type 'wine explorer' to get started using the file explorer.".lightblue
+    puts "Type 'WINEPREFIX=#{@xdg_config_home}/.wine wine explorer' to get started using the file explorer.".lightblue
```
uberhacker, line 97: Would it make sense to put this in a wrapper script?

satmandu: I don't think so. That's just there in case WINEPREFIX wasn't set.

We're talking about people who put their devices in dev mode. We don't need to mollycoddle everyone!

This should indicate to people that they need WINEPREFIX set. It's wine. It's not `gemacs`.

uberhacker: Don't you realize everybody needs mollycoddled today? ;)

dciliske:
Uhmm... sorry to report: `crew update
remote: Enumerating objects: 4, done.
remote: Counting objects: 100% (4/4), done.
remote: Compressing objects: 100% (4/4), done.
remote: Total 4 (delta 0), reused 0 (delta 0), pack-reused 0
Unpacking objects: 100% (4/4), 43.69 KiB | 395.00 KiB/s, done.
   64e31d93..f21ec212  master     -> origin/master
HEAD is now at f21ec212 Wine to 6.5 (#5532)
Generating compatible packages...
############################################################################################################################ 100.0%############################################################################################################################ 100.0%
/usr/local/lib/crew/packages/wine.rb:44:in `<class:Wine>': undefined method `empty?' for Wine:Class (NoMethodError)
        from /usr/local/lib/crew/packages/wine.rb:4:in `<top (required)>'
        from <internal:/usr/local/lib64/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:/usr/local/lib64/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from /usr/local/bin/crew:138:in `set_package'
        from /usr/local/bin/crew:196:in `block in generate_compatible'
        from /usr/local/bin/crew:194:in `each'
        from /usr/local/bin/crew:194:in `generate_compatible'
        from /usr/local/bin/crew:434:in `update'
        from /usr/local/bin/crew:1089:in `update_command'
        from /usr/local/bin/crew:1117:in `<main>'`

satmandu:
hmmm... ok @uberhacker  I'll try reverting to the previous version of that line:
```
  @xdg_config_home = ENV['XDG_CONFIG_HOME']
  @xdg_config_home = "#{CREW_PREFIX}/.config" if @xdg_config_home.to_s.empty?
```

PR #5534, opened 2021-04-03
[New Package] ruby-webrick

saltedcoffii:
This package is really interesting. It's the first package in chromebrew that build with rake and install with gem, much in the way we install packages with pip. This kind of format could allow us to, instead of just installing gems in package files, to instead package gems in a much more clean way.

### DIFF
--- a/packages/ruby_webrick.rb
+++ b/packages/ruby_webrick.rb
@@ -1,0 +1,23 @@
+require 'package'
+
+class Ruby_webrick < Package
+  description 'WEBrick is an HTTP server toolkit that can be configured as an HTTPS server, a proxy server, and a virtual-host server.'
+  homepage 'https://github.com/ruby/webrick/'
+  version '1.7.0'
+  license 'BSD-2'
+  compatibility 'all'
+  source_url 'https://github.com/ruby/webrick/archive/refs/tags/v1.7.0.tar.gz'
+  source_sha256 '25df35c696301deddec8cc929954284fa3dfaa86ad90c0cce2ecc5740c476545'
+
+  def self.build
+    system 'rake', 'build'
+  end
+
+  def self.check
+    system 'rake', 'test'
+  end
+
+  def self.install
+    system 'gem', 'install', "--build=#{CREW_DEST_DIR}", 'pkg/webrick-1.7.0.gem'
+  end
+end


packages/ruby_webrick.rb
```diff
@@ -0,0 +1,23 @@
...
+  def self.build
+    system 'rake', 'build'
+  end
+
+  def self.check
+    system 'rake', 'test'
+  end
+
+  def self.install
+    system 'gem', 'install', "--build=#{CREW_DEST_DIR}", 'pkg/webrick-1.7.0.gem'
```
uberhacker, line 21: Hmmm, does this install libraries?  If so, those will need to be abstracted with `CREW_LIB_PREFIX`.  We'll also want to abstract a prefix to use `CREW_PREFIX`.

saltedcoffii: interestingly enough, it just listens to ruby's configuration, according to my testing. If ruby is installed to /usr/local and the gems are kept at /usr/local/lib64/ruby/gems/3.0.0, it'll install them there. If the libdir is just /usr/local/lib, it'll do that too.

uberhacker: It looks like there is at least one dependency.  `crew install ruby_webrick` results below:
```
rake aborted!
/usr/local/lib64/ruby/3.0.0/yaml.rb:3: warning: It seems your ruby installation is missing psych (for YAML output).
To eliminate this warning, please install libyaml and reinstall your ruby.
<internal:/usr/local/lib64/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require': libyaml-0.so.2: cannot open shared object file: No such file or directory - /usr/local/lib64/ruby/3.0.0/x86_64-linux-gnu/psych.so (LoadError)
        from <internal:/usr/local/lib64/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from /usr/local/lib64/ruby/3.0.0/psych.rb:13:in `<top (required)>'
        from <internal:/usr/local/lib64/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:/usr/local/lib64/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from /usr/local/lib64/ruby/3.0.0/yaml.rb:4:in `<top (required)>'
        from <internal:/usr/local/lib64/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:/usr/local/lib64/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from /usr/local/lib64/ruby/3.0.0/rubygems.rb:641:in `load_yaml'
        from /usr/local/lib64/ruby/3.0.0/rubygems/config_file.rb:336:in `load_file'
        from /usr/local/lib64/ruby/3.0.0/rubygems/config_file.rb:181:in `initialize'
        from /usr/local/lib64/ruby/3.0.0/rubygems/gem_runner.rb:73:in `new'
        from /usr/local/lib64/ruby/3.0.0/rubygems/gem_runner.rb:73:in `do_configuration'
        from /usr/local/lib64/ruby/3.0.0/rubygems/gem_runner.rb:38:in `run'
        from /usr/local/bin/gem:21:in `<main>'

Tasks: TOP => build
ruby_webrick failed to install: `rake build` exited with 1
```

uberhacker: Actually, I just tested and `depends_on 'libyaml'` fixes the above error.

uberhacker: It looks like it installs in `/usr/local/lib64/ruby/gems`.  This could start a whole slew of gem wrapper packages. :)

uberhacker:
I also ran `gem list webrick` before and after install and it works cleanly.

packages/ruby_webrick.rb
saltedcoffii: That's what I'm sayyinngggg

saltedcoffii: Also @uberhacker Ruby depends on libyaml so theroretically it should be installed.

uberhacker: If we get too carried away with crew ruby gem wrapper packages that means every time ruby is updated, ALL the wrapper packages need to be updated also since Ruby includes the version in the path.

saltedcoffii: It might be helpful for things that other packages depend on like ronn

PR #5535, opened 2021-04-03
Add premake package

uberhacker:
Premake is a command line utility which reads a scripted definition of a software project and, most commonly, uses it to generate project files for toolsets like Visual Studio, Xcode, or GNU Make.  See https://premake.github.io/.  Tested on all architectures.

### DIFF
--- a/packages/premake.rb
+++ b/packages/premake.rb
@@ -1,0 +1,39 @@
+require 'package'
+
+class Premake < Package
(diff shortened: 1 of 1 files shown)


satmandu:
Their versioning is a bit of a joke, no?

uberhacker:
Alpha forever? :)

PR #5538, opened 2021-04-04
crew: Add SHA512 and MD5 support

supechicken666:
Since many sources started to provide SHA512 sums, it could be convenient for packagers to copy it from sources  

Tested on `x86_64`


saltedcoffii:
@supechicken666 could you also add Blake2b and MD5 support for Gentoo and Debian respectively?

crew
```diff
@@ -916,7 +930,7 @@ def archive_package (pwd)
   Dir.chdir CREW_DEST_DIR do
     system "tar c#{@verbose}Jf #{pwd}/#{pkg_name} *"
   end
-  system "sha256sum #{pwd}/#{pkg_name} > #{pwd}/#{pkg_name}.sha256"
+  File.write("#{pwd}/#{pkg_name}.sha512", Digest::SHA512.hexdigest( File.binread("#{pwd}/#{pkg_name}") ) )
```
uberhacker, line 933: Don't we want csum and csum_type here also?

crew
```diff
@@ -529,17 +529,31 @@ def download
 
   uri = URI.parse url
   filename = File.basename(uri.path)
+  
   if source
-    sha256sum = @pkg.source_sha256
+    if @pkg.source_sha512
```
uberhacker, line 534: Will this throw an error if source_sha512 is not set?

crew
```diff
@@ -529,17 +529,31 @@ def download
...
+  elsif @pkg.binary_sha512
+    csum_type = 512
+    csum = @pkg.binary_sha512[@device[:architecture]]
+  elsif @pkg.binary_sha256
+    csum_type = 256
+    csum = @pkg.binary_sha256[@device[:architecture]]
   else
-    sha256sum = @pkg.binary_sha256[@device[:architecture]]
+    abort 'No checksum provided for this package, aborting...'.lightred
   end
```
uberhacker, line 549: `case...when...else...end` would be more elegant.

crew
supechicken666: Not needed, this is the part used to generate the checksum for packaged binary

crew
supechicken666: No, as this is a instance variable (return nil if not set)

crew
supechicken666: Do you mean `case source...else...end`?

supechicken666:
> @supechicken666 could you also add Blake2b and MD5 support for Gentoo and Debian respectively?

Sure, md5 support are also include in `digest` but blake2b require an extra [gem](https://github.com/mgomes/blake2b) to check

crew
uberhacker: You can use `when` instead of `elsif`.

crew
uberhacker: Not following.

saltedcoffii:
Package MD5 for now and I'll package Blake2B and add support for it later

crew
supechicken666: I mean this is a part only used in `crew build`, so it is fine to default `sha512sum`

crew
supechicken666: Done

crew
uberhacker: If you are going to do this then you might as well get rid of the other options for binary checksums and just go with sha512sum.

supechicken: Can we keep the compatibility for them? (since many source are still providing `sha256sum` in their checksum file)

uberhacker: Sure but the way you have it set up currently, there is no way to implement md5 or blake2b checksums.

supechicken: Removed md5 support

PR #5539, opened 2021-04-04
firefox: Add arm support

supechicken666:
Add arm support using official Debian binary

Tested on `x86_64`
**Not tested on `aarch64` or `arm` yet**


packages/firefox.rb
```diff
@@ -3,16 +3,22 @@
...
   when 'i686'
     source_url "https://archive.mozilla.org/pub/firefox/releases/#{version}/linux-i686/en-US/firefox-#{version}.tar.bz2"
     source_sha256 '9127aee106dd9f09fac0c3cb89c5d75553384da4ec9be5943b60a5f55f31fccc'
   when 'x86_64'
     source_url "https://archive.mozilla.org/pub/firefox/releases/#{version}/linux-x86_64/en-US/firefox-#{version}.tar.bz2"
     source_sha256 '3c9207bee0a998634c4fd12293acfae207d16508749ad405bf1e8717d06acf02'
+  when 'armv7l', 'aarch64'
+    @_arm = true
+    version '87.0-2' # follow debian patch
+    source_url "http://http.us.debian.org/debian/pool/main/f/firefox/firefox_#{version}_armhf.deb"
```
uberhacker, line 20: If we're going to start using Debian packages, we should be consistent and use them for i686 and x86_64 also.  It would be a maintenance nightmare to manage, otherwise.

supechicken666: Moved to `deb`

packages/firefox.rb
```diff
@@ -39,10 +47,22 @@ class Firefox < Package
   depends_on 'pulseaudio'
   depends_on 'sommelier'
 
+  def self.patch   
+    @_wrapper = <<~EOF
+      #!/bin/sh
+      # To get sound working, used : https://codelab.wordpress.com/2017/12/11/firefox-drops-alsa-apulse-to-the-rescue/
+      
+      exec apulse #{CREW_PREFIX}/lib/firefox/firefox "$@"
```
uberhacker: I thought we didn't need apulse to get sound working.  In any event, that should be a dependency and it's not listed.  See if it works without the wrapper script.

uberhacker:
Looks like it fails the arm test.
```
$ firefox
XPCOMGlueLoad error for file /usr/local/lib/firefox/libmozsandbox.so:
/usr/local/lib/libc.so.6: version `GLIBC_2.28' not found (required by /usr/local/lib/firefox/libmozsandbox.so)
Couldn't load XPCOM.
$ crew const LIBC_VERSION
LIBC_VERSION=2.27
```
An upgrade to glibc 2.28 _might_ fix it.

packages/firefox.rb
supechicken666: You're right, the sound is working without `apulse`

supechicken666:
> Looks like it fails the arm test.
> 
> ```
> $ firefox
> XPCOMGlueLoad error for file /usr/local/lib/firefox/libmozsandbox.so:
> /usr/local/lib/libc.so.6: version `GLIBC_2.28' not found (required by /usr/local/lib/firefox/libmozsandbox.so)
> Couldn't load XPCOM.
> $ crew const LIBC_VERSION
> LIBC_VERSION=2.27
> ```
> 
> An upgrade to glibc 2.28 _might_ fix it.

Upgrading `glibc.so` cause segfault on all my command (`ls`, `cat`)... Just changed the source to `ubuntu xenial` (an older system than `debian-sid`), can you try again :)

saltedcoffii:
that'll have an older version of firefox @supechicken maybe we don't want that. However you might be able to edit the binary with `patchelf` and change what it links to.

uberhacker:
It we're going to go through all this trouble and jump through this many loops to get the Debian package to work, we might as well attempt to build from source. :)

satmandu:
If you want to use the patched debian source... it would make sense to just build that. I'm not shocked that there's a glibc issue.

uberhacker:
@supechicken666: Have you had any luck building this from source?

satmandu (changes requested):
This still needs to be modified to take out the apulse stuff.

PR #5540, opened 2021-04-04
Python 3.9.4

satmandu:
- Python 3.9.4 

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -23,8 +23,21 @@ class Python3 < Package
      x86_64: 'cd3a63250e76cd887a620df69f37d349d4cea0d39e3d8bc2ceef9400a40befbb'
   })
 
+  depends_on 'bz2'
+  depends_on 'gdbm'
+  depends_on 'glibc'
+  depends_on 'krb5' => :build
+  depends_on 'libdb'
+  depends_on 'libffi'
+  depends_on 'libtirpc' => :build
+  depends_on 'ncurses'
+  depends_on 'openssl'  
+  depends_on 'readline'
+  depends_on 'sqlite'
   # depends_on 'tcl' # Needed for tkinter support
   # depends_on 'tk'  # Needed for tkinter support
+  depends_on 'xzutils'
+  depends_on 'zlibpkg'
 
   def self.preinstall
     # Fix ImportError: cannot import name 'PackageFinder'.
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -23,9 +23,22 @@ class Python3 < Package
      x86_64: 'cd3a63250e76cd887a620df69f37d349d4cea0d39e3d8bc2ceef9400a40befbb'
   })
 
+  depends_on 'fontconfig'
+  depends_on 'freetype'
+  depends_on 'glib'
+  depends_on 'graphite'
+  depends_on 'harfbuzz'
+  depends_on 'libdb'
+  depends_on 'libpng'
   depends_on 'libx11'
+  depends_on 'libxau'
+  depends_on 'libxcb'
+  depends_on 'libxext'
+  depends_on 'libxft'
+  depends_on 'libxrender'
+  depends_on 'libxss'
   depends_on 'tcl'
-  depends_on 'tk' => :build # yes this creates a circular dependency.
+  depends_on 'tk'
 
   def self.preinstall
     # Fix ImportError: cannot import name 'PackageFinder'.
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -24,26 +24,13 @@ class Python3 < Package
   })
 
   depends_on 'bz2' => :build
-  depends_on 'fontconfig'
-  depends_on 'freetype'
   depends_on 'gdbm' => :build
-  depends_on 'glib'
-  depends_on 'graphite'
-  depends_on 'harfbuzz'
   depends_on 'krb5' => :build
   depends_on 'libdb'
-  depends_on 'libpng'
   depends_on 'libtirpc' => :build
-  depends_on 'libx11'
-  depends_on 'libxau'
-  depends_on 'libxcb'
-  depends_on 'libxext'
-  depends_on 'libxft'
-  depends_on 'libxrender'
-  depends_on 'libxss'
   depends_on 'sqlite' => :build
-  depends_on 'tcl'
-  depends_on 'tk'
+  depends_on 'tcl' => :build # Also needed for tkinter support
+  depends_on 'tk' => :build # Also needed for tkinter support
   depends_on 'xzutils' => :build
 
   def self.preinstall
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -23,13 +23,17 @@ class Python3 < Package
      x86_64: 'cd3a63250e76cd887a620df69f37d349d4cea0d39e3d8bc2ceef9400a40befbb'
   })
 
+  depends_on 'bz2' => :build
   depends_on 'fontconfig'
   depends_on 'freetype'
+  depends_on 'gdbm' => :build
   depends_on 'glib'
   depends_on 'graphite'
   depends_on 'harfbuzz'
+  depends_on 'krb5' => :build
   depends_on 'libdb'
   depends_on 'libpng'
+  depends_on 'libtirpc' => :build
   depends_on 'libx11'
   depends_on 'libxau'
   depends_on 'libxcb'
@@ -37,8 +41,10 @@ class Python3 < Package
   depends_on 'libxft'
   depends_on 'libxrender'
   depends_on 'libxss'
+  depends_on 'sqlite' => :build
   depends_on 'tcl'
   depends_on 'tk'
+  depends_on 'xzutils' => :build
 
   def self.preinstall
     # Fix ImportError: cannot import name 'PackageFinder'.
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -7,8 +7,8 @@ class Python3 < Package
   version @_ver
   license 'PSF-2.0'
   compatibility 'all'
-  source_url "https://www.python.org/ftp/python/#{@_ver}/Python-#{@_ver}.tgz"
-  source_sha256 '66c4de16daa74a825cf9da9ddae1fe020b72c3854b73b1762011cc33f9e4592f'
+  source_url "https://www.python.org/ftp/python/#{@_ver}/Python-#{@_ver}.tar.xz"
+  source_sha256 '4b0e6644a76f8df864ae24ac500a51bbf68bd098f6a173e27d3b61cdca9aa134'
 
   binary_url({
     aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/python3-3.9.4-chromeos-armv7l.tar.xz',
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -3,33 +3,29 @@ require 'package'
 class Python3 < Package
   description 'Python is a programming language that lets you work quickly and integrate systems more effectively.'
   homepage 'https://www.python.org/'
-  @_ver = '3.9.2'
+  @_ver = '3.9.4'
   version @_ver
   license 'PSF-2.0'
   compatibility 'all'
-  source_url "https://www.python.org/ftp/python/#{@_ver}/Python-#{@_ver}.tar.xz"
-  source_sha256 '3c2034c54f811448f516668dce09d24008a0716c3a794dd8639b5388cbde247d'
+  source_url "https://www.python.org/ftp/python/#{@_ver}/Python-#{@_ver}.tgz"
+  source_sha256 '66c4de16daa74a825cf9da9ddae1fe020b72c3854b73b1762011cc33f9e4592f'
 
   binary_url({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/python3-3.9.2-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/python3-3.9.2-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/python3-3.9.2-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/python3-3.9.2-chromeos-x86_64.tar.xz'
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/python3-3.9.4-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/python3-3.9.4-chromeos-armv7l.tar.xz',
+       i686: 'https://dl.bintray.com/chromebrew/chromebrew/python3-3.9.4-chromeos-i686.tar.xz',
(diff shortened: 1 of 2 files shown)


saltedcoffii:
I'm working on a huge overhaul of how python works with chromebrew, i tested it with this update and it works.

packages/python3.rb
```diff
@@ -3,33 +3,48 @@
...
   description 'Python is a programming language that lets you work quickly and integrate systems more effectively.'
   homepage 'https://www.python.org/'
-  @_ver = '3.9.2'
+  @_ver = '3.9.4'
   version @_ver
   license 'PSF-2.0'
   compatibility 'all'
-  source_url "https://www.python.org/ftp/python/#{@_ver}/Python-#{@_ver}.tar.xz"
-  source_sha256 '3c2034c54f811448f516668dce09d24008a0716c3a794dd8639b5388cbde247d'
+  source_url "https://www.python.org/ftp/python/#{@_ver}/Python-#{@_ver}.tgz"
```
uberhacker, line 10: Isn't the `.tar.xz` archive smaller?

packages/python3.rb
```diff
@@ -3,33 +3,48 @@
...
+  depends_on 'libxau'
+  depends_on 'libxcb'
+  depends_on 'libxext'
+  depends_on 'libxft'
+  depends_on 'libxrender'
+  depends_on 'libxss'
+  depends_on 'sqlite' => :build
+  depends_on 'tcl'
+  depends_on 'tk'
+  depends_on 'xzutils' => :build
```
uberhacker, line 47: Does this expand the list of core dependencies?  In other words, does `tools/core_packages.txt` need updated?

packages/python3.rb
satmandu: Yes but the xz file hadn't been uploaded yet when I wanted to build it.

uberhacker: I see it's available now.

packages/python3.rb
satmandu: Tcl & tk aren't in core... But does it matter?

What changes for installs if we don't add it to core?

Nothing?

Tcl/tk will also get installed with just adding the deps to the python3 package file anyways.

uberhacker: Yes, it matters.  The list of core packages should be complete.  Since python3 is in core, the dependencies should be included in core also.

packages/python3.rb
satmandu: For clarification, the URL was there previously, but it was 404ing. I'll update the URL/hash now that it is actually available.

packages/python3.rb
satmandu: Added tcl & tk to core packages in this PR.

tools/core_packages.txt
```diff
@@ -81,6 +81,8 @@ ruby
 sed
 slang
 sqlite
+tcl
+tk
```
uberhacker, line 85: On a fresh install, I don't see the following:
```
groff
libunbound
libxdmcp
pcre2
trousers
uchardet
xorg_proto
```
Can we remove these packages?

uberhacker: Also, I attempted to install python3 after a fresh install and got the following NEW dependencies.  These will need to be added to `tools/core_packages.txt`.
````
$ crew install python3
python3: Python is a programming language that lets you work quickly and integrate systems more effectively.
https://www.python.org/
Version: 3.9.4
License: PSF-2.0
fontconfig gperf freetype_sub libpng jsonc util_linux libcap_ng libtinfo linux_pam libdb pcre2 libeconf graphite freetype harfbuzz gobject_introspection glib pcre shared_mime_info six pygments libx11 xorg_proto libxcb xcb_proto libxau pthread_stubs libxtrans libxext libxft libxrender libxss tcl tk xorg_lib libfontenc libxinerama libxdamage libxfixes libxtst libxi libpciaccess libxkbfile libxshmfence libxcursor libdmx libfs libice libbsd libsm libxaw libxmu libxt libxpm libxfont2 fop jdk8 libxrandr libxres libxv libxvmc libxxf86dga libxxf86vm libxcomposite 
Do you agree? [Y/n]
```
Not sure why we need all the X window dependencies for python3.  Can you explain?

uberhacker: After combining the 2 package lists from above, I got the complete list of changes needed as shown below:
```
+fontconfig
+fop
+freetype
+freetype_sub
+glib
+gobject_introspection
+gperf
+graphite
-groff
+harfbuzz
+jdk8
+jsonc
+libbsd
+libcap_ng
+libdb
+libdmx
+libeconf
+libfontenc
+libfs
+libice
+libpciaccess
+libpng
+libsm
+libtinfo
-libunbound
+libx11
+libxau
+libxaw
+libxcb
+libxcomposite
+libxcursor
+libxdamage
-libxdmcp
+libxext
+libxfixes
+libxfont2
+libxft
+libxi
+libxinerama
+libxkbfile
+libxmu
+libxpm
+libxrandr
+libxrender
+libxres
+libxshmfence
+libxss
+libxt
+libxtrans
+libxtst
+libxv
+libxvmc
+libxxf86dga
+libxxf86vm
+linux_pam
+pcre
+pthread_stubs
+pygments
+shared_mime_info
+six
+tcl
+tk
-trousers
-uchardet
+util_linux
+xcb_proto
+xorg_lib
+xorg_proto
```
If we can figure out a way to NOT have all the X window libraries and jdk8, that would be preferrable.

packages/python3.rb
uberhacker: Dependency list is not complete.

tools/core_packages.txt
satmandu: It's from `tk`:
_tkinter.cpython-39-x86_64-linux-gnu.so:libXrender.so.1

Hmm. Let me rebuild this w/o tk.

saltedcoffii: I'll add tkinter as its own package :)

satmandu: @saltedcoffii Thanks. I'm rebuilding with `--disable-tcl --disable-tk`

satmandu: Hmm. Still seeing this:
```
  depends_on 'fontconfig'
  depends_on 'freetype'
  depends_on 'glib'
  depends_on 'graphite'
  depends_on 'harfbuzz'
  depends_on 'libdb'
  depends_on 'libpng'
  depends_on 'libx11'
  depends_on 'libxau'
  depends_on 'libxcb'
  depends_on 'libxext'
  depends_on 'libxft'
  depends_on 'libxrender'
  depends_on 'libxss'
  depends_on 'pcre'
  depends_on 'tcl'
  depends_on 'tk'
```

Maybe I need `--with-x=no` ?

satmandu: That doesn't work either. I would suggest INTENTIONALLY avoiding including tcl/tk in base, and maybe just create a dummy package for tkinter which includes those dependencies. (and tcl/tk can be left as build dependencies for python3.)

packages/python3.rb
satmandu: Removing the tkinter library (e.g. `/usr/local/lib/python3.9/lib-dynload/_tkinter.cpython-39-i386-linux-gnu.so`)  from depenendency analysis removes all the other x11 dependencies.

So I'm just removing that file from consideration for dependencies, and now there are no new core dependencies.

tools/core_packages.txt
uberhacker: It looks like `libdb` needs to be added to `tools/core_packages.txt`.  It'd be nice to remove the other dependencies also.

uberhacker:
A good way you can check the core dependencies after a fresh install and this updated python3 package:
```
$ cd /path/to/chromebrew/source/tools
$ crew install vim
$ crew list installed -d > installed.txt
$ vimdiff installed.txt core_packages.txt
```
This should show what needs to be added/removed from `core_packages.txt`.

satmandu:
@uberhacker My getdeps script automatically compares to the core package list via 
```
# Remove original package from list.
pkgdeps=$(echo $pkgdeps | tr " " "\n" | sed "/${1}/d")
pkgdepsnocore=$pkgdeps
for i in $(curl -Ls https://github.com/skycocker/chromebrew/raw/master/tools/core_packages.txt) freetype_sub
do
  file=$(basename $i)
  pkgdepsnocore=$(echo $pkgdepsnocore | tr " " "\n" | sed "/$file/d")
done
pkgdepsnocore=$(echo $pkgdepsnocore | tr " " "\n"  | sort -u )
```
The issue was with tkinter adding the kitchen sink:
```
ldd /usr/local/lib64/python3.9/lib-dynload/_tkinter.cpython-39-x86_64-linux-gnu.so
        linux-vdso.so.1 (0x00007ffd09ba7000)
        libtk8.6.so => /usr/local/lib64/libtk8.6.so (0x00007874acb63000)
        libtcl8.6.so => /usr/local/lib64/libtcl8.6.so (0x00007874ac9aa000)
        libX11.so.6 => /usr/local/lib64/libX11.so.6 (0x00007874ac83b000)
        libpthread.so.0 => /usr/local/lib64/libpthread.so.0 (0x00007874ac81b000)
        libc.so.6 => /usr/local/lib64/libc.so.6 (0x00007874ac670000)
        libXft.so.2 => /usr/local/lib64/libXft.so.2 (0x00007874ac658000)
        libfontconfig.so.1 => /usr/local/lib64/libfontconfig.so.1 (0x00007874ac5fc000)
        libfreetype.so.6 => /usr/local/lib64/libfreetype.so.6 (0x00007874ac50b000)
        libXss.so.1 => /usr/local/lib64/libXss.so.1 (0x00007874ac505000)
        libXext.so.6 => /usr/local/lib64/libXext.so.6 (0x00007874ac4ee000)
        libdl.so.2 => /usr/local/lib64/libdl.so.2 (0x00007874ac4e8000)
        libz.so.1 => /usr/local/lib64/libz.so.1 (0x00007874ac4c9000)
        libm.so.6 => /usr/local/lib64/libm.so.6 (0x00007874ac36e000)
        libxcb.so.1 => /usr/local/lib64/libxcb.so.1 (0x00007874ac33a000)
        /usr/local/lib64/ld-linux-x86-64.so.2 (0x00007874accd5000)
        libXrender.so.1 => /usr/local/lib64/libXrender.so.1 (0x00007874ac130000)
        libexpat.so.1 => /usr/local/lib64/libexpat.so.1 (0x00007874ac0f5000)
        libbz2.so.1.0 => /usr/local/lib64/libbz2.so.1.0 (0x00007874ac0e3000)
        libpng16.so.16 => /usr/local/lib64/libpng16.so.16 (0x00007874ac09c000)
        libharfbuzz.so.0 => /usr/local/lib64/libharfbuzz.so.0 (0x00007874abfb1000)
        libbrotlidec.so.1 => /usr/local/lib64/libbrotlidec.so.1 (0x00007874abfa3000)
        libXau.so.6 => /usr/local/lib64/libXau.so.6 (0x00007874abf9d000)
        libgraphite2.so.3 => /usr/local/lib64/libgraphite2.so.3 (0x00007874abf74000)
        libglib-2.0.so.0 => /usr/local/lib64/libglib-2.0.so.0 (0x00007874abe2c000)
        libbrotlicommon.so.1 => /usr/local/lib64/libbrotlicommon.so.1 (0x00007874abe09000)
        libpcre.so.1 => /usr/local/lib64/libpcre.so.1 (0x00007874abd7f000)
```

tools/core_packages.txt
satmandu: Added `libdb` to core, and removed the other deps.

packages/python3.rb
```diff
@@ -3,33 +3,28 @@
...
+     armv7l: 'fe4fde3f9738ac76096dfdc13ff6931543e2f612035421e294f87d2bfc09773c',
+       i686: 'a9e7b61911679cda55f65cf5ca29076ee2ac4540c8dd7bda5322df66d4d6e4bd',
+     x86_64: 'cd3a63250e76cd887a620df69f37d349d4cea0d39e3d8bc2ceef9400a40befbb'
   })
 
-  depends_on 'bz2'
-  depends_on 'xzutils'
-  depends_on 'sqlite'
-  depends_on 'krb5'
-  depends_on 'libtirpc'
```
uberhacker, line 30: Are these packages included in core by other core packages?  We have to be careful about removing dependencies of core packages.

packages/python3.rb
```diff
@@ -3,33 +3,28 @@
...
-  depends_on 'tk' => :build # yes this creates a circular dependency.
-  depends_on 'gdbm' => :build
+  # depends_on 'tcl' # Needed for tkinter support
+  # depends_on 'tk'  # Needed for tkinter support
```
uberhacker, line 27: Duplicate of line 31?

packages/python3.rb
satmandu: These are the actual runtime deps which are in core:
```
  depends_on 'glibc'
  depends_on 'readline'
  depends_on 'ncurses'
  depends_on 'openssl'
  depends_on 'libdb'
  depends_on 'gdbm'
  depends_on 'zlibpkg'
  depends_on 'sqlite'
  depends_on 'bz2'
  depends_on 'xzutils'
  depends_on 'libffi'
```
As build deps, sqlite, krb5, libtirpc, and xzutils are all already in core.

Since Python3 is IN core, should I put all the build and runtime deps back in?

satmandu: That's what I did.

uberhacker: The core packages and their dependencies are what make up the core packages so if you remove any dependencies from one core package, another core package better install those dependencies too or else the list is not complete.  You can test easy enough with a fresh install to see if it matches `tools/core_packages.txt`.

uberhacker: Have the binaries been updated?

satmandu: I haven't touched the binaries. There are no changes in the compile, so I don't think they need updating.

satmandu: Also, the simple solution to issues with core packages getting installed via some hierarchy of package installs... would be to just create an initial hierarchy of core installs... and just make that in initial install. in install.sh. (That's what I have been doing using `tsort`.)

satmandu: Here's my current workaround for getting all of core installed:
```
# Use parallel to download essential packages, then install.
corefilepkgs=$(curl -Ls https://github.com/skycocker/chromebrew/raw/master/tools/core_packages.txt )
buildessentialpkgs=$(grep "^[[:blank:]] depends_on" packages/buildessential.rb | awk '{print $2}' | sed "s;';;g" )
echo "$EARLY_PACKAGES $corefilepkgs $buildessentialpkgs less most" | tr '\n' ' ' > /tmp/sorted.txt
for i in {1..7}
do
  echo "Dependency Search Pass $i"
  [[ -f /tmp/depraw.txt ]] && rm /tmp/depraw.txt
  get_deps
done
cat /tmp/sorted.txt
# Don't re-download EARLY_PACKAGES
for i in $EARLY_PACKAGES
do
  sed -i "/$i/d" /tmp/sorted.txt
done
sorted=$(cat /tmp/sorted.txt)
rm /tmp/{sorted.txt,deps.txt,depraw.txt}
echo "corefiles & buildessentials sorted in 1 level dependency order: $sorted"
parallel -j 4 CURL="$CURL" crew -d download -- "$sorted" || true
for i in  libjpeg gdk_pixbuf $sorted
do
yes |  CURL=$CURL crew install $i || true
done
```

PR #5544, opened 2021-04-05
[WIP] vivaldi => 3.7.2218.52

supechicken666:
Tested on `x86_64`


packages/vivaldi.rb
```diff
@@ -3,68 +3,48 @@
...
-    system "curl -#LO https://downloads.vivaldi.com/stable/vivaldi-stable_#{version}-1_#{arch}.deb"
-    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest( File.read("vivaldi-stable_#{version}-1_#{arch}.deb") ) == sha256
-    system "alien -tc vivaldi-stable_#{version}-1_#{arch}.deb"
-    system "tar xvf vivaldi-stable-#{version}.tgz"
+  
+  source_url "https://downloads.vivaldi.com/stable/vivaldi-stable_#{@_ver}-1_#{@_arch}.deb"
+  
+  def self.patch
+    FileUtils.ln_sf "#{CREW_PREFIX}/share/vivaldi/vivaldi", './usr/bin/vivaldi-stable'
+    FileUtils.ln_s "#{CREW_PREFIX}/share/vivaldi/vivaldi", './usr/bin/vivaldi'
```
uberhacker, line 35: Wouldn't it be easier to do this in `def self.install`?  Regardless, I don't think this qualifies as a patch per say.

packages/vivaldi.rb
```diff
@@ -3,68 +3,48 @@
...
 
   def self.install
     FileUtils.mkdir_p CREW_DEST_PREFIX
-    FileUtils.rm 'usr/bin/vivaldi-stable'
-    FileUtils.mv 'usr/bin', CREW_DEST_PREFIX
-    FileUtils.mv 'opt', CREW_DEST_PREFIX
-    FileUtils.mv 'usr/share', CREW_DEST_PREFIX
-    FileUtils.ln_s "#{CREW_PREFIX}/opt/vivaldi/vivaldi", "#{CREW_DEST_PREFIX}/bin/vivaldi"
+    FileUtils.mv './etc/', CREW_DEST_PREFIX
+    FileUtils.mv Dir['./usr/*'], CREW_DEST_PREFIX
```
uberhacker, line 41: Does this create any conflicts with existing files in `/usr/local`?  You can possibly get away with something like this in `/usr/local/share` but not `/usr/local`.

packages/vivaldi.rb
supechicken666: Moved to install part

packages/vivaldi.rb
supechicken666: Probably not, only `share` and `bin` directory are located at `./usr`

packages/vivaldi.rb
```diff
@@ -3,68 +3,58 @@
...
-    puts
-    puts "Type 'vivaldi' to get started.".lightblue
-    puts
-    puts "To completely remove, execute the following:".lightblue
-    puts "crew remove vivaldi".lightblue
-    puts "rm -rf ~/.config/vivaldi".lightblue
-    puts "rm -rf ~/.config/lib/vivaldi".lightblue
-    puts "rm -f ~/.config/share/.vivaldi_reporting_data".lightblue
-    puts
+    system "#{CREW_PREFIX}/share/vivaldi/update-ffmpeg", '--user'
```
uberhacker, line 51: Doesn't this require `depends_on 'ffmpeg'`?

packages/vivaldi.rb
```diff
@@ -3,68 +3,58 @@
...
+  end
+  
+  def self.remove
+    FileUtils.rm_rf HOME + '/.local/lib/vivaldi'
```
uberhacker, line 55: Why not add this path to the array also?

supechicken666: Done

packages/vivaldi.rb
supechicken666: Not needed? The downloaded ffmpeg patch is a static library
```shell
$ ldd /usr/local/share/vivaldi/media-codecs-89.0.4389.82/libffmpeg.so 
ldd: warning: you do not have execution permission for `/usr/local/share/vivaldi/media-codecs-89.0.4389.82/libffmpeg.so'
        linux-vdso.so.1 (0x00007ffdb2d52000)
        libpthread.so.0 => /usr/local/lib64/libpthread.so.0 (0x00007a75e008e000)
        librt.so.1 => /usr/local/lib64/librt.so.1 (0x00007a75dfe86000)
        libm.so.6 => /usr/local/lib64/libm.so.6 (0x00007a75dfb27000)
        libgcc_s.so.1 => /usr/local/lib64/libgcc_s.so.1 (0x00007a75e0d54000)
        libc.so.6 => /usr/local/lib64/libc.so.6 (0x00007a75df76e000)
        /usr/local/lib64/ld-linux-x86-64.so.2 (0x00007a75e0b49000)
```

PR #5547, opened 2021-04-06
change vim defaults to no use visual mode

satmandu:
Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686


uberhacker:
Don't we need to update gvim also?

satmandu:
> Don't we need to update gvim also?
Building...

satmandu:
gvim added.

PR #5549, opened 2021-04-06
grep rebuild

satmandu:
- Noticed that hash for grep wasn't matching download, so rebuilt it.
- Also added `pcre` dep.

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/grep.rb
+++ b/packages/grep.rb
@@ -1,35 +1,38 @@
+# Adapted from Arch Linux grep PKGBUILD at:
+# https://github.com/archlinux/svntogit-packages/raw/packages/grep/trunk/PKGBUILD
 
 require 'package'
 
 class Grep < Package
   description 'A string search utility'
   homepage 'https://www.gnu.org/software/grep/'
-  version '3.6'
+  version '3.6-1'
   license 'GPL-3+'
   compatibility 'all'
   source_url 'https://ftp.gnu.org/gnu/grep/grep-3.6.tar.xz'
   source_sha256 '667e15e8afe189e93f9f21a7cd3a7b3f776202f417330b248c2ad4f997d9373e'
 
-  binary_url ({
-     aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/grep-3.6-chromeos-armv7l.tar.xz',
-      armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/grep-3.6-chromeos-armv7l.tar.xz',
-        i686: 'https://dl.bintray.com/chromebrew/chromebrew/grep-3.6-chromeos-i686.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/grep.rb
```diff
@@ -1,35 +1,38 @@
...
-        i686: 'a6a62d2c91c346041fc4408727e387e2cc415e304d2ca7d372a97256520d6f64',
-      x86_64: '0106cbf95ada69fc0205b4af06af4e46144156934555b5bd9376e4c00d61092f',
+  binary_sha256({
+    aarch64: '15c5b97f4a77e1253220a259670d674a19e470db88a0148d79dfd6b48a74ccfa',
+     armv7l: '15c5b97f4a77e1253220a259670d674a19e470db88a0148d79dfd6b48a74ccfa',
+       i686: 'c732e8f7ce7e218553628203356d031d5a022d7a718ffa7b5529344edacead9b',
+     x86_64: 'bcb9d3d7f541fbd54170b9479a48ecb0c567a0e946a6733e1817f17f813e9260'
   })
 
+  depends_on 'pcre'
```
uberhacker, line 28: Any difference between `pcre` and `pcre2`?

packages/grep.rb
```diff
@@ -1,35 +1,38 @@
...
 
   def self.build
-    ENV['CFLAGS'] = '-flto'
-    system "./configure #{CREW_OPTIONS}\
-    --without-included-regex"
```
uberhacker, line 30: Doesn't this flag mean we want to use an external package?

packages/grep.rb
satmandu: Entirely different packages.

packages/grep.rb
satmandu: [linuxfromscratch](http://www.linuxfromscratch.org/lfs/view/10.0/chapter06/grep.html) doesn't use that flag, but [arch does](https://github.com/archlinux/svntogit-packages/blob/packages/grep/trunk/PKGBUILD). Using the builtin regex seems like the more conservative option.

uberhacker: Would we need the pcre dependency still?

packages/grep.rb
uberhacker: Does one work any better than the other and is it relevant to grep?

satmandu: I can't say. It's not an option to switch as far as I can tell.

packages/grep.rb
satmandu: rebuilt without, using debian build options,  but no change in the dependency.

packages/grep.rb
satmandu: no idea, but can't get rid of it it seems.

PR #5550, opened 2021-04-06
evince + evince/epiphany prereqs/updates

satmandu:
- Adds evince, updates 
- updates ghostscript

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686


packages/evince.rb
```diff
@@ -0,0 +1,118 @@
...
+  depends_on 'openjpeg'
+  depends_on 'orc'
+  depends_on 'pango'
+  depends_on 'pcre'
+  depends_on 'pixman'
+  depends_on 'poppler'
+  depends_on 't1lib'
+  depends_on 'texlive'
+  depends_on 'util_linux'
+  depends_on 'wayland'
```
uberhacker, line 99: I really want to get away from doing this to many more packages.

uberhacker: What I would like for us to do is create a script to find only the parent dependencies and include those only.  No need to list all.

satmandu: I think that's a great idea, so long as we do it to every package so we don't miss any.
(just using `readelf -d binary` gives just the parent dependencies...)

satmandu: (Except for core... I imagine for core you want all dependencies listed.)

satmandu: (Except for core... I imagine for core you want all dependencies listed.)

uberhacker: Yeah, definitely should include all for any core package unless there are child dependencies of other core packages.  In that case, those don't need to be listed.

satmandu: Looking for just parent runtime deps:
```
getdepsparent evince
  depends_on 'atk'
  depends_on 'cairo'
  depends_on 'djvulibre'
  depends_on 'gdk_pixbuf'
  depends_on 'glib'
  depends_on 'gnome_desktop'
  depends_on 'gst_plugins_base'
  depends_on 'gstreamer'
  depends_on 'gtk3'
  depends_on 'libarchive'
  depends_on 'libgxps'
  depends_on 'libhandy'
  depends_on 'libsecret'
  depends_on 'libspectre'
  depends_on 'libtiff'
  depends_on 'nautilus'
  depends_on 'pango'
  depends_on 'poppler'
```

A much smaller list... but this should be run on every package to make sure everything is captured.  That can be partially automated, but that's going to be tedious since that involves installing every package and running a script afterwards.

uberhacker: Run a script to install the package, inspect and then remove (if needed).

satmandu:
@uberhacker note changes to multiple packages to move to parent deps only...

packages/djvulibre.rb
```diff
@@ -0,0 +1,50 @@
...
+  })
+  binary_sha256({
+    aarch64: '2a662b5b95e3e5c24531cb4ae0f5b33c044895f33fc88c562026a4124ff15295',
+     armv7l: '2a662b5b95e3e5c24531cb4ae0f5b33c044895f33fc88c562026a4124ff15295',
+       i686: 'a0cf591f8e04fc89a0c2f45f0ad67c7c907971dcf85371a2184c3ea54e605da4',
+     x86_64: 'd17ec121e01aea869ea74b421aebcf4b8f019e8b32268859256aadd32019aab5'
+  })
+ 
+  depends_on 'libjpeg'
+  depends_on 'libtiff'
```
uberhacker, line 29: Doesn't libtiff include libjpeg as a dependency?  This means libjpeg is a child dependency and doesn't need to be listed, right?

packages/evince.rb
```diff
@@ -0,0 +1,67 @@
...
+  depends_on 'cairo'
+  depends_on 'djvulibre'
+  depends_on 'docbook_xsl' => ':build'
+  depends_on 'gdk_pixbuf'
+  depends_on 'glib'
+  depends_on 'gnome_desktop'
+  depends_on 'gobject_introspection' => ':build'
+  depends_on 'gst_plugins_base'
+  depends_on 'gstreamer'
+  depends_on 'gtk3'
```
uberhacker, line 38: The gtk3 package already includes many child dependencies listed here.

packages/ghostscript.rb
```diff
@@ -3,40 +3,80 @@
...
+  depends_on 'atk'
+  depends_on 'cairo'
+  depends_on 'cups'
   depends_on 'fontconfig'
-  depends_on 'jasper'
-  depends_on 'gsfonts'
+  depends_on 'freetype'
+  depends_on 'gdk_pixbuf'
+  depends_on 'glib'
+  depends_on 'gtk3'
```
uberhacker, line 32: The gtk3 package already includes many child dependencies.

packages/gtk3.rb
```diff
@@ -37,15 +37,9 @@ class Gtk3 < Package
   depends_on 'gdk_pixbuf'
   depends_on 'glib'
```
uberhacker, line 38: gdk_pixbuf already includes glib?

packages/libdazzle.rb
```diff
@@ -3,27 +3,30 @@
...
+    aarch64: 'c11c683320703431a0ead7f365e25e5fb7db17933967d0c23a7c12c62b269c86',
+     armv7l: 'c11c683320703431a0ead7f365e25e5fb7db17933967d0c23a7c12c62b269c86',
+       i686: 'cd1ad492b66fe8e6c52c7a67d51b117237f2b9395edfa282e63095cf916ed6dd',
+     x86_64: '132e15b726324b2d08b450637d2a7435f8dd6b471fb2218b368835fe43c05fca'
   })
 
-  depends_on 'gobject_introspection'
+  depends_on 'cairo'
+  depends_on 'gdk_pixbuf'
+  depends_on 'glib'
```
uberhacker, line 27: gdk_pixbuf already includes glib?

packages/libdazzle.rb
```diff
@@ -3,27 +3,30 @@
...
   depends_on 'gtk3'
```
uberhacker, line 28: Actually, gtk3 includes all the child dependencies listed above.

packages/libgxps.rb
```diff
@@ -0,0 +1,53 @@
...
+  depends_on 'freetype'
+  depends_on 'glib'
+  depends_on 'gobject_introspection' => ':build'
+  depends_on 'gtk_doc' => ':build'
+  depends_on 'lcms'
+  depends_on 'libarchive'
+  depends_on 'libjpeg'
+  depends_on 'libpaper'
+  depends_on 'libpng'
+  depends_on 'libtiff'
```
uberhacker, line 40: libtiff already includes libjpeg?

packages/libtiff.rb
```diff
@@ -22,11 +22,17 @@ class Libtiff < Package
...
+  depends_on 'libglu'
+  depends_on 'libice'
   depends_on 'libjpeg'
+  depends_on 'libsm'
   depends_on 'libwebp'
-  depends_on 'libdeflate'
-  depends_on 'imake' => :build
+  depends_on 'libx11'
+  depends_on 'libxi'
+  depends_on 'mesa'
```
uberhacker, line 35: mesa already includes libx11?

packages/poppler.rb
```diff
@@ -3,43 +3,38 @@
...
   depends_on 'cairo'
   depends_on 'fontconfig'
-  depends_on 'freeglut'
   depends_on 'freetype'
   depends_on 'glib'
-  depends_on 'harfbuzz'
   depends_on 'lcms'
   depends_on 'libjpeg'
   depends_on 'libpng'
   depends_on 'libtiff'
```
uberhacker, line 33: libtiff already includes libjpeg?

packages/gtk4.rb
```diff
@@ -25,67 +25,26 @@ class Gtk4 < Package
...
-  depends_on 'dav1d'
-  depends_on 'dbus'
-  depends_on 'elfutils'
-  depends_on 'eudev'
   depends_on 'ffmpeg'
   depends_on 'fontconfig'
-  depends_on 'freetype'
   depends_on 'fribidi'
   depends_on 'gdk_pixbuf'
   depends_on 'glib'
```
uberhacker, line 35: gdk_pixbuf already includes glib?

packages/djvulibre.rb
satmandu: The logic I used here is, "What parent libraries are directly asked for by libraries and binaries in this package? List the packages which have those libraries." 

I think the alternate logic you are suggesting is "What parent libraries are directly asked for by libraries and binaries in this package? List the packages which have those libraries, but don't list a package if another package listed would already provide that package."

The first algorithm guarantees packages won't break if the dependency tree implied in the second algorithm changes (through a package getting dropped from a build because of updates, etc.) I think it's a safe middle ground between "list all dependencies recursively as suggested by `ldd`" and "list the minimum possible set of dependencies."

Thoughts?

uberhacker: I lean in the direction of the minimum possible set of dependencies.  If we test packages thoroughly after each update, we should be fine.

satmandu: But not putting it there doesn't make it a smaller list of dependencies. That's just _hiding a dependency_. Listing the actual dependencies also gives insight into what the package is doing and wants, which I also think is a useful element of looking at a package file list of dependencies. Knowing that this package is calling both libjpeg and libtiff is _useful._

uberhacker: I think it's useful for package maintainers but we should be able to _see_ the child dependencies with some kind of script without the need to list verbatim, right?

satmandu: Are we not the package maintainers? Are we not the target audience for what's inside these files?

uberhacker: I just don't want us to get carried away with the "unnecessary" dependencies because it becomes difficult to maintain when every package has redundancies especially when the dependencies keep changing.

satmandu: This is literally the reason I'm trying to keep all immediate/direc/parent dependencies in each package file.

> because it becomes difficult to maintain when every package has redundancies especially when the dependencies keep changing.

This is literally the reason I'm trying to keep all immediate/direc/parent dependencies in each package file. :)

saltedcoffii:
To all of the above reviews: We should probably include dependencies even if they're dependencies of other dependencies so if we decide to rearrange dependencies, we suddenly don't get a bunch of library not found errors.

uberhacker:
I don't necessarily agree.  Especially when the package depends on metapackages like gtk# and sommelier.  It just makes it more difficult to manage.  Those packagers that rearrange dependencies have the responsibility to make sure there aren't library not found errors in other packages.

satmandu:
Sommelier is not needed for gnome apps (gtk3/gtk4), since they can use Wayland. Thus sommelier is an inappropriate metapackage for x dependencies for those apps...

uberhacker:
I'm not suggesting that you add sommelier.  I would just like the dependency list chopped down to a manageable level.

satmandu:
gtk2, gtk3, and gtk4 are also inappropriate as metapackages since there are graphical packages which may use a combination of the above packages...

We don't want to bloat a system by having unnecessary packages installed.

Honestly I think a some of the packages in gtk2, gtk3, and gtk4 are often build dependencies at best for many packages.

It would be nice to have some build dependency packages which are huge metapackages, but only go with actual dependencies the rest of the time so we never have more packages installed than actually needed by the user.

satmandu:
Let me go through the gtk3 and gtk4 deps when I get a chance? I very much want to make the dep list right to make the package manageable, but I need enough metadata to be able to figure out dependency loops, and that's hard to do when dependencies are in a metapackage.

PR #5561, opened 2021-04-07
Add `caja` package

supechicken666:
Tested on `x86_64`

### DIFF
--- a/packages/caja.rb
+++ b/packages/caja.rb
@@ -1,0 +1,46 @@
+# Adapted from Arch Linux caja PKGBUILD at:
+# https://github.com/archlinux/svntogit-community/blob/packages/caja/trunk/PKGBUILD
+
+require 'package'
+
+class Caja < Package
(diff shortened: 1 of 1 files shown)


packages/caja.rb
```diff
@@ -0,0 +1,46 @@
...
+  depends_on 'libnotify'
+  depends_on 'sommelier'
+
+  def self.build
+    system "env CFLAGS='-flto=auto' CXXFLAGS='-flto=auto' \
+                LDFLAGS='-flto=auto' \
+            ./configure #{CREW_OPTIONS}"
+    
+    # https://bugzilla.gnome.org/show_bug.cgi?id=656231
+    system "sed -i 's/ -shared / -Wl,-O1,--as-needed\\0/g' libtool"
```
uberhacker, line 30: Could this line go in `def self.patch` instead?

supechicken: No, the `libtool` file is generated in `configure`

PR #5572, opened 2021-04-08
Epiphany + webkit2gtk updates

satmandu:
- Adds epiphany
- breaks out version 4 and version 5 of webkit2gtk, as they are compiled from the same source, but have different dependencies based upon using gtk3 or gtk4.
- webkit2gtk built without bubblewrap, as that is having issues.
- Also depends upon https://github.com/skycocker/chromebrew/pull/5550

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686 (not epiphany but everything else)

### DIFF
--- a/packages/epiphany.rb
+++ b/packages/epiphany.rb
@@ -9,11 +9,8 @@ class Epiphany < Package
   version '40.0'
   license 'GPL'
   compatibility 'x86_64 aarch64 armv7l'
-  case ARCH
-  when 'x86_64', 'aarch64', 'armv7l'
-    source_url "https://gitlab.gnome.org/GNOME/epiphany/-/archive/#{version}/epiphany-#{version}.tar.bz2"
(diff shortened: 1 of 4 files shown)


uberhacker:
If you trim down the dependencies to the minimum, I'll merge this @satmandu.

packages/epiphany.rb
```diff
@@ -0,0 +1,65 @@
...
+require 'package'
+
+class Epiphany < Package
+  description 'A GNOME web browser based on the WebKit rendering engine'
+  homepage 'https://wiki.gnome.org/Apps/Web'
+  version '40.0'
+  license 'GPL'
+  compatibility 'x86_64 aarch64 armv7l'
+  case ARCH
+  when 'x86_64', 'aarch64', 'armv7l'
```
supechicken, line 13: This line can be removed

PR #5575, opened 2021-04-09
suggested formatting changes for dependencies

satmandu:
- This is what I'm thinking about as a way forward to figure out what deps are and how they should be cleaned up.

Works properly:
- [x] x86_64


uberhacker:
What is the difference between a logical and runtime dependency?  Both would technically be considered runtime, right?

satmandu:
Logical as a catchall for packages which should be installed but which are
not literal parent runtime dependencies as defined by what "readelf -d"
spits out.

In short, logical is short for dependency overrides, such as xdg_base or
sommelier for an X app.

satmandu:
Because it is useful to know why a package is listed as a dependency when no apps or libraries in the relevant package seem to want it.

Might also make sense to leave a comment as to why a logical dependency is there...

PR #5576, opened 2021-04-09
FIx flatpak, fix gdk_pixbuf, add stressng

satmandu:
- Fixes package naming for flatpak, and adds missing deps
- fix gdk_pixbuf postinstall
- Adds stressng binaries

Works properly:
- [x] x86_64


uberhacker:
Nice catches.

PR #5578, opened 2021-04-09
intel, texlive, hashpipe, gst_plugins_good , texlive updates

satmandu:
- new gmmlib allows for new intel_media_driver build.
- Hacked up hashpipe to work in docker containers which don't have `O_TMPFILE` functionality in their overlay filesystems by liberally copying code from gtk, which has a workaround.
- rebuilt gst-plugins_good to avoid libjpet_turbo
- updated texlive

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l (where applicable)
- [x] i686 (where applicable)

### DIFF
--- a/packages/gst_plugins_good.rb
+++ b/packages/gst_plugins_good.rb
@@ -4,53 +4,59 @@ class Gst_plugins_good < Package
   description 'Multimedia graph framework - good plugins'
   homepage 'https://gstreamer.freedesktop.org/'
   @_ver = '1.18.4'
-  version @_ver
+  version "#{@_ver}-1"
   license 'LGPL-2.1+'
   compatibility 'all'
   source_url "https://gstreamer.freedesktop.org/src/gst-plugins-good/gst-plugins-good-#{@_ver}.tar.xz"
   source_sha256 'b6e50e3a9bbcd56ee6ec71c33aa8332cc9c926b0c1fae995aac8b3040ebe39b0'
 
   binary_url({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/gst_plugins_good-1.18.4-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/gst_plugins_good-1.18.4-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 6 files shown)


packages/texlive.rb
```diff
@@ -40,21 +40,21 @@ class Texlive < Package
...
-      FileUtils.rm_rf 'mfpr'
-      FileUtils.rm_rf 'pixman'
-      FileUtils.rm_rf 'poppler'
-      FileUtils.rm_rf 'zlib'
-      FileUtils.rm_rf 'zziplib'
+      # FileUtils.rm_rf 'mfpr'
+      # FileUtils.rm_rf 'pixman'
+      # FileUtils.rm_rf 'poppler'
+      # FileUtils.rm_rf 'zlib'
+      # FileUtils.rm_rf 'zziplib'
```
uberhacker, line 57: Can we just remove these lines instead of commenting?

packages/gst_plugins_good.rb
```diff
@@ -4,53 +4,59 @@ class Gst_plugins_good < Package
...
+  depends_on 'libxext' # R
+  depends_on 'libxfixes' # R
+  depends_on 'mpg123' # R
+  depends_on 'orc' # R
+  depends_on 'pipewire' # R
+  depends_on 'pulseaudio' # R
+  depends_on 'speex' # R
+  depends_on 'taglib' # R
+  depends_on 'v4l_utils' # R
+  depends_on 'wavpack' # R
```
uberhacker, line 59: This dependency list is still a bit longer than necessary.  gtk3 takes care of a lot of these implicitly.

packages/hashpipe.rb
```diff
@@ -3,25 +3,203 @@
...
+       
+       LDLIBS := -lcrypto
+      -CFLAGS ?= -O3 -march=native
+      +CFLAGS ?= -O3
+       CFLAGS += -std=gnu11 -Wall -pedantic
+       
+       all: hashpipe
+    HASHPIPE_PATCHEOF
+    IO.write('hashpipe.patch', @hashpipe_patch, perm: 0o644)
+    system 'patch -Np1 -i hashpipe.patch'
```
uberhacker, line 200: Epic patch. :scream_cat:

satmandu: It seemed easier to do that than submit a patch to Jason Donenfeld...

PR #5582, opened 2021-04-10
License notice

saltedcoffii:
If one reads near the bottom of the GPL license, they'll notice some recommendations of implementing license notices. This adds an implementation of a license notice to chromebrew. Works on x86_64.

### DIFF
--- a/crew
+++ b/crew
@@ -47,7 +47,7 @@ Usage:
 version #{CREW_VERSION}
 DOCOPT
 
-LICENSE = <<~LICENSESTRING
+CREW_LICENSE = <<~LICENSESTRING
   Copyright (C) 2021 Chromebrew Authors
 
   This program is free software: you can redistribute it and/or modify
@@ -82,7 +82,7 @@ rescue Docopt::Exit => e
       puts CREW_VERSION
       exit 0
     when '-L', '--license'
-      puts LICENSE
+      puts CREW_LICENSE
       exit 0
     end
     if ARGV[0] != '-h' and ARGV[0] != '--help' then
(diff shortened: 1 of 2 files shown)


crew
```diff
@@ -64,6 +80,9 @@ rescue Docopt::Exit => e
     if ARGV[0] == '-V' or ARGV[0] == '--version' then
       puts CREW_VERSION
       exit 0
+    elsif ARGV[0] == '-L' or ARGV[0] == '--license' then
+      puts LICENSE
+      exit 0
     end
```
uberhacker, line 86: I'm thinking this might be a bit more elegant:
```
case ARGV[0]
when '-V', '--version'
  puts CREW_VERSION
  exit 0
when '-L', '--license'
  puts LICENSE
  exit 0
end
```

crew
```diff
@@ -47,6 +47,22 @@ Usage:
 version #{CREW_VERSION}
 DOCOPT
 
+LICENSE = <<~LICENSESTRING
```
supechicken, line 50: Move to `const.rb`?

supechicken, line 50: Move to `const.rb`?

saltedcoffii: Sure, if you'd like.

uberhacker: Good suggestion @supechicken.  Also change to `puts CREW_LICENSE`.

saltedcoffii: On second thought, it made more sense since the DOCOPT string is right above it

uberhacker: Yeah, it doesn't have to be a crew constant.

PR #5584, opened 2021-04-10
Busybox

saltedcoffii:
Works on x86_64.

I feel like I should provide some justification for why I did what I did in this configuration:
Busybox and its applets were dynamically linked against libbusybox, because when I tried to statically link it with glibc the binaries were massive and with musl they segfaulted.
The busybox binary itself is in /usr/local/bin, but everything else is in /usr/local/libexec/busybox/ to avoid overwriting everything. When crew-alternatives really happens this will be a big asset. Busybox provides files like patch, diff, man and a bunch of compression libraries, so I'm hoping that maybe it would allow us to get rid of some core packages.

I disabled all the tools having to do with framebuffer consoles (chromeos doesn't have framebuffer support), init (filesystem is r/o and chromeos has upstart) and klibc stuff (we're using glibc). Pretty much everything else is enabled.


packages/busybox.rb
```diff
@@ -0,0 +1,1256 @@
...
+      CONFIG_FEATURE_REMOTE_LOG=y
+      CONFIG_FEATURE_SYSLOGD_DUP=y
+      CONFIG_FEATURE_SYSLOGD_CFG=y
+      CONFIG_FEATURE_SYSLOGD_PRECISE_TIMESTAMPS=y
+      CONFIG_FEATURE_SYSLOGD_READ_BUFFER_SIZE=256
+      CONFIG_FEATURE_IPC_SYSLOG=y
+      CONFIG_FEATURE_IPC_SYSLOG_BUFFER_SIZE=16
+      CONFIG_FEATURE_KMSG_SYSLOG=y
+    BUSYBOXCONFIGEOF
+    IO.write(".config", @busyboxconfig)
```
uberhacker, line 1212: A 1200 line patch has got to be a record. :)  Although, this could qualify as a build step instead.

PR #5586, opened 2021-04-10
util_linux doesn't depend on libtinfo

saltedcoffii:
Since we use `--with-ncursesw` on util linux, it no longer depends on libtinfo.

### DIFF
--- a/packages/util_linux.rb
+++ b/packages/util_linux.rb
@@ -23,7 +23,6 @@ class Util_linux < Package
   })
 
   depends_on 'libcap_ng'
-  depends_on 'libtinfo'
   depends_on 'linux_pam'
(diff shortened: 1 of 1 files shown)


uberhacker:
I don't think you understand the purpose of libtinfo.  The only reason this package exists is to fix binaries that are looking for libtinfo.so.5 which doesn't exist in any other package.  While I agree it doesn't have to depend on libtinfo, it doesn't hurt to include it as a dependency.  The freebasic package depends on libtinfo because the binaries are looking for libtinfo.so.5.

saltedcoffii:
I understand the reason for the package perfectly, and am saying that util linux doesn't depend on the symlink. Why bloat the package more than needed?

PR #5587, opened 2021-04-10
add conflicts_with

saltedcoffii:
For a really long time we've needed the ability to have conflicting packages. Well here it is!!

I need help though, the 
```ruby
abort "#{conflict} is installed so #{**FIXME**} cannot be installed!".lightred
```
line is broken because I can't figure out the variable name of the current package trying to be installed. I tried so much. Other than that one message, it works!

### DIFF
--- a/lib/package.rb
+++ b/lib/package.rb
@@ -37,6 +37,14 @@ class Package
     @dependencies
   end
 
+  def self.conflicts_with (conflict = nil)
+    Dir[CREW_META_PATH + '*.directorylist'].sort.map do |f|
+      if File.basename(f, '.directorylist').include? conflict then
+        abort "#{conflict} is installed so #{**FIXME**} cannot be installed!".lightred
+      end
+    end
+  end
+
   def self.get_url (architecture)
     if !@build_from_source && @binary_url && @binary_url.has_key?(architecture)
       return @binary_url[architecture]


saltedcoffii:
Also, I can't figure it out but if anyone can, being able to support conflicting with a package only during buildtime, e.g. `conflicts_with 'libiconv' => :build` would be amazing.

satmandu:
This may not be the best method, but this works from inside a package file:
```
    @package = File.basename(__FILE__, File.extname(__FILE__))
    puts "Package is #{@package}"
```

satmandu:
Also, doesn't `@pkg.name` work?

supechicken:
You can use `self.superclass.name` to get the package name 

https://stackoverflow.com/a/14778831

supechicken:
> Also, doesn't `@pkg.name` work?

Probably not, as `package.rb` is required by `pkg.rb`, so it can't access variables from `crew` or `pkg.rb`

uberhacker:
I don't think it would be necessary to explicitly declare this at all.  In fact, it would probably be better if this was handled internally within crew build.

supechicken:
> I don't think it would be necessary to explicitly declare this at all. In fact, it would probably be better if this was handled internally within crew build.

But how can we access the function from packages? `require 'crew'` may not a solution

uberhacker:
I will create a PR for review.

uberhacker:
Fixed in #5588.

saltedcoffii:
@uberhacker does this mean we can now create packages that conflict?

uberhacker:
You won't be able to build any packages until conflicts are resolved first.

PR #5590, opened 2021-04-11
Fix crew whatprovides for files with underscores

### LINKED ISSUE #5589: crew whatprovides cannot find files with underscores
```
$ crew whatprovides iconv_close.3.gz
$ crew files manpages | grep iconv_close.3.gz
/usr/local/share/man/man3/iconv_close.3.gz
```

uberhacker:
Fixes #5589.  It makes sense to replace the underscores with commas for substitutions since filenames cannot have that character.

### DIFF
--- a/crew
+++ b/crew
@@ -419,7 +419,7 @@ end
 
 def whatprovides (regexPat)
   fileArray = []
-  needle = regexPat.gsub(/-/,'_').gsub(/_/,'\-')
+  needle = regexPat.gsub(/-/,',').gsub(/,/,'\-')
   Dir[CREW_META_PATH + '*.filelist'].each do |packageList|
     packageName = File.basename packageList, '.filelist'
     File.readlines(packageList).each do |line|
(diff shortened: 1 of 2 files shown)


crew
```diff
@@ -419,7 +419,7 @@ end
 
 def whatprovides (regexPat)
   fileArray = []
-  needle = regexPat.gsub(/-/,'_').gsub(/_/,'\-')
+  needle = regexPat.gsub(/-/,',').gsub(/,/,'\-')
```
supechicken, line 422: Do you mean `regexPat.gsub(/-/, '.')`? (Dots is represented to `any character` in regex)

uberhacker: No, I cannot simply `.gbus(/-/, '\-')` so I need to convert the dash to a comma first.

satmandu:
As I understand it, filenames can have any character except for  NUL and slash as per https://unix.stackexchange.com/a/155860, but POSIX compliance has stricter rules which don't allow for instance `,` but do allow `_`, so this seems like an appropriate way to go.

PR #5592, opened 2021-04-11
Binutils rebuild to stop putting files in /usr/local/CREW_TGT

satmandu:
- Previously binutils was dumping files in places like this:
```
/usr/local/armv7l-cros-linux-gnueabihf/bin/ar
/usr/local/armv7l-cros-linux-gnueabihf/bin/as
/usr/local/armv7l-cros-linux-gnueabihf/bin/ld
/usr/local/armv7l-cros-linux-gnueabihf/bin/ld.bfd
/usr/local/armv7l-cros-linux-gnueabihf/bin/ld.gold
/usr/local/armv7l-cros-linux-gnueabihf/bin/nm
/usr/local/armv7l-cros-linux-gnueabihf/bin/objcopy
/usr/local/armv7l-cros-linux-gnueabihf/bin/objdump
/usr/local/armv7l-cros-linux-gnueabihf/bin/ranlib
/usr/local/armv7l-cros-linux-gnueabihf/bin/readelf
/usr/local/armv7l-cros-linux-gnueabihf/bin/strip
```
- This uses a method from gentoo to avoid this.

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/binutils.rb
+++ b/packages/binutils.rb
@@ -4,44 +4,56 @@ class Binutils < Package
   description 'The GNU Binutils are a collection of binary tools.'
   homepage 'https://www.gnu.org/software/binutils/'
   @_ver = '2.36.1'
-  version @_ver
+  version "#{@_ver}-1"
   license 'GPL-3+'
   compatibility 'all'
   source_url "https://ftpmirror.gnu.org/binutils/binutils-#{@_ver}.tar.xz"
   source_sha256 'e81d9edf373f193af428a0f256674aea62a9d74dfe93f65192d4eae030b0f3b0'
 
   binary_url({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/binutils-2.36.1-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/binutils-2.36.1-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/binutils-2.36.1-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/binutils-2.36.1-chromeos-x86_64.tar.xz'
+    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/binutils-2.36.1-1-chromeos-armv7l.tar.xz',
+     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/binutils-2.36.1-1-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/binutils.rb
```diff
@@ -4,44 +4,56 @@ class Binutils < Package
...
     Dir.mkdir 'build'
     Dir.chdir 'build' do
       system "env CFLAGS='-pipe -flto=auto' CXXFLAGS='-pipe -flto=auto' \
         LDFLAGS='-flto=auto' \
         ../configure #{CREW_OPTIONS} \
-        --with-system-zlib \
+        --disable-bootstrap \
         --disable-maintainer-mode \
-        --enable-shared \
+        --enable-64-bit-bfd \
```
uberhacker, line 42: Only applies to x86_64?

saltedcoffii: yeah, according to binutils's gentoo ebuild, --enable-64-bit-bfd is only relevant on 64 bit processors and is much slower than the 32 bit bfd so should be omitted on 32 bit system. Maybe add this and use the variable appropriately.
```
case ARCH
when 'i686' or 'armv7l'
  @_64_bit_bfd = '--disable-64-bit-bfd'
when 'x86_64' or 'aarch64'
  @_64_bit_bfd = '--enable-64-bit-bfd'
end
```

uberhacker: You will need to disable for aarch64 also since the userland is only armv7l compatible:

satmandu: The reason I left it in was that this is apparently used on non-64-bit architectures too:

https://lfs.maru-na.net/lfs/view/systemd/chapter08/binutils.html

```
--enable-64-bit-bfd
Enables 64-bit support (on hosts with narrower word sizes). May not be needed on 64-bit systems, but does no harm.
```

(Apparently it is required on 32-bit systems if you want to build some things like grub. Src: https://schlomp.space/teldra/vp-build/commit/1dcd80236d7032da541efa1b1d8a687ab6fbc0d3 )

satmandu: The relevant gentoo bug report is from 9 years ago: https://bugs.gentoo.org/438522 

Is this slowness from memory pressure still relevant? I'm compiling i686 and armv7l packages in containers with > 4Gb of ram, so this shouldn't even be an issue.

satmandu: Debian doesn't use it at all, so maybe I should just remove it?

https://salsa.debian.org/toolchain-team/binutils/-/blob/master/debian/rules

satmandu: I'm going to try to avoid arch specific customizations here, so I'll eliminate that line. If it is done automatically for the relevant architectures, great!

Issue #5520, opened 2021-04-01
[Feature request] Force compilation from source


murkyness:
Almost everything installed with crew is precompiled. Please add a flag to force compilation from source.

BONUS: Allow passing CFlags/CXXFlags to gcc :grin:

satmandu:
`crew install -s <packagename>`

You can also edit the package file before running that command to edit CFLAGS, etc.

saltedcoffii:
Editing the package file isn't a super great way since `crew update` overwrites the merge. Also if you want to have every single package compile from the source always, you can set `CREW_PREFIX=/usr/local` in your shell's environment.

murkyness:
> if you want to have every single package compile from the source always, you can set `CREW_PREFIX=/usr/local` in your shell's environment

`$ crew const` returns that CREW_PREFIX is already set to /usr/local. I think that's the default setting?

saltedcoffii:
Oh sorry. I mean adding `export CREW_PREFIX=/usr/local` to your shell rc.

PR #5593, opened 2021-04-11
Add compiling constants

saltedcoffii:
I'm doing this to standardize the build flags we use in our packages as autotools defaults to -O2 -g, which some packages of ours keep and others trash. These constants may be customizable in the future, allowing people to customize and pass these flags to gcc (see #5520).

Here are the justifications for using the flags I did:

- `-Os` is optimizes binaries for size. I considered using `-O2` (safest optimizations) however so many chromebooks have small hard drives. We may decide to switch to `-O2` (NOTE: `-O3` and `-Ofast` are dangerous to be set globally)
- `-pipe` speeds builds by causing the compiler to use memory/swap instead of temp files on the disk. This can increase liklihood of large builds failing due to memory running out, however this is extremely rare. Given the amount of swap space on chromebooks, this almost certainly won't happen. It has no affect on the final binaries.
- `-flto=auto` causes gcc to use link-time optimization (slows builds, speeds up performance). NOTE: clang uses `-flto=full` or `-flto=thin`, `-flto=auto` will error. If a package does not support lto, it should be patched. Debian bookworm is expected to link-time optimize all packages so debian will likely have sufficient patches. If not, ask me.
- `-fuse-ld=gold` causes gcc to use the gold linker instead of bfd. Gold links much faster (about a 2x speedup) as it is built for ELF binaries and bfd only has ELF support "patched" in. We could modify gcc's build option to cause it to default to gold and binutils to `--default-gold`, however that would be more work for @satmandu and they basically do the same thing. 
- I considered using `-fomit-frame-pointer`, which causes the frame pointer to only be kept in memory where the compiler determines its needed, speeding up binaries, but gentoo cautions against setting it as a global option since it can break some compiles (e.g. glibc)
- `-g` keeps debugging stuff in the binaries, which is great for developers, but does nothing for us but make the binaries bigger, so I excluded it.

`CFLAGS` are flags passed to the c compiler, `CXXFLAGS` to the c++ compiler, and `FCFLAGS` and `FFLAGS` are used with different fortran compilers.

the `CREW_OPTIONS_ENV` constants should be set before configure is run in autotools packages, as shown:
```
system "#{CREW_ENV_OPTIONS} ./configure #{CREW_OPTIONS} \
		--enable-foo \
		--without-bar \
		--with-baz"
```

I changed meson and cmake's options to use minsize and MinSizeRel respectively, that will swap out -O2 for -Os.

In `CREW_MESON_LTO_OPTIONS`, I changed `-fuse-ld=lld` to `-fuse-ld=gold`, because, even though `lld` is faster, it produces slower binaries than gold. As a general rule, one should use clang and lld when debugging and gcc and gold (or bfd) for production. We are production.

As always, I'm open to feedback on this! Let me know if you have any concerns.


satmandu (commented):
We need a non-LTO crew meson options. Some packages won't compile with LTO.

I'd be ok with flipping the default such that CREW_MESON_OPTIONS does LTO and having a CREW_MESON_NOLTO_OPTIONS as well though.

saltedcoffii:
Okay, we can do that until I get around to patching all the files that won't compile with lto

satmandu:
But I would also check packages which have been edited in the last 3 months and are not using CREW_MESON_LTO_OPTIONS and have them changed to the non-LTO option, since probably those don't support LTO.

saltedcoffii:
I'll just do mass changes with `sed` I think, that way there aren't any problems.

crew
```diff
@@ -355,7 +355,7 @@ def const (var)
       'CREW_LIB_PREFIX',
       'CREW_LIB_SUFFIX',
       'CREW_MAN_PREFIX',
-      'CREW_MESON_LTO_OPTIONS',
```
uberhacker, line 358: Why was this removed?

lib/const.rb
```diff
@@ -72,15 +72,22 @@
...
+FCFLAGS = CREW_COMMON_FLAGS
+FFLAGS = CREW_COMMON_FLAGS
+LDFLAGS = "'-flto=auto'"
 
+CREW_ENV_OPTIONS = "CFLAGS=#{CFLAGS} CXXFLAGS=#{CXXFLAGS} FCFLAGS=#{FCFLAGS} FFLAGS=#{FFLAGS} LDFLAGS=#{LDFLAGS}"
 CREW_OPTIONS = "--prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --mandir=#{CREW_MAN_PREFIX} --build=#{CREW_BUILD} --host=#{CREW_TGT} --target=#{CREW_TGT} --program-prefix='' --program-suffix=''"
-CREW_MESON_OPTIONS = "-Dprefix=#{CREW_PREFIX} -Dlibdir=#{CREW_LIB_PREFIX} -Dmandir=#{CREW_MAN_PREFIX} -Dbuildtype=release -Dc_args='-fuse-ld=lld -pipe' -Dcpp_args='-fuse-ld=lld -pipe'"
-CREW_MESON_LTO_OPTIONS = "-Dprefix=#{CREW_PREFIX} -Dlibdir=#{CREW_LIB_PREFIX} -Dmandir=#{CREW_MAN_PREFIX} -Dbuildtype=release -Db_lto=true -Dcpp_args='-fuse-ld=gold -pipe' -Dc_args='-fuse-ld=gold -pipe'"
+CREW_MESON_OPTIONS = "-Dprefix=#{CREW_PREFIX} -Dlibdir=#{CREW_LIB_PREFIX} -Dmandir=#{CREW_MAN_PREFIX} -Dbuildtype=minsize -Db_lto=true -Dcpp_args='#{CXXFLAGS}' -Dc_args='#{CFLAGS}'"
+CREW_MESON_LTO_OPTIONS = CREW_MESON_OPTIONS # This option is depreciated, please use CREW_MESON_OPTIONS
```
uberhacker, line 85: Misspelled `deprecated`?

lib/const.rb
```diff
@@ -72,15 +72,22 @@
...
 
 # Cmake sometimes wants to use LIB_SUFFIX to install libs in LIB64, so specify such for x86_64
 # This is often considered deprecated. See discussio at https://gitlab.kitware.com/cmake/cmake/-/issues/18640
 # and also https://bugzilla.redhat.com/show_bug.cgi?id=1425064
 # Let's have two CREW_CMAKE_OPTIONS since this avoids the logic in the recipe file.
-CREW_CMAKE_OPTIONS = "-DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX} -DCMAKE_LIBRARY_PATH=#{CREW_LIB_PREFIX} -DCMAKE_BUILD_TYPE=Release --build=#{CREW_BUILD} --host=#{CREW_TGT} --target=#{CREW_TGT}"
+CREW_CMAKE_OPTIONS = "-DCMAKE_INSTALL_PREFIX=#{CREW_PREFIX} -DCMAKE_LIBRARY_PATH=#{CREW_LIB_PREFIX} -DCMAKE_C_FLAGS='#{CFLAGS}' -DCMAKE_CXX_FLAGS='#{CXXFLAGS}' -DCMAKE_BUILD_TYPE=MinSizeRel --build=#{CREW_BUILD} --host=#{CREW_TGT} --target=#{CREW_TGT}"
```
uberhacker, line 91: Won't `'#{CFLAGS}'` create double single quotes?

lib/const.rb
```diff
@@ -72,15 +72,22 @@
   CREW_BUILD = 'x86_64-cros-linux-gnu'
 end
 
+CREW_COMMON_FLAGS = "'-Os -pipe -flto=auto -fuse-ld=gold'"
+CFLAGS = CREW_COMMON_FLAGS
+CXXFLAGS = CREW_COMMON_FLAGS
+FCFLAGS = CREW_COMMON_FLAGS
+FFLAGS = CREW_COMMON_FLAGS
```
uberhacker, line 79: No need to create all these constants.  Just use `CREW_COMMON_FLAGS` for everything.

saltedcoffii: What if we need to individualize these?

uberhacker: We cross that bridge when we get there.  Conflict in `lib/const.rb`?

uberhacker: I went ahead and resolved the conflict.

saltedcoffii: Fixed.

lib/const.rb
saltedcoffii: Yes, correcting that.

crew
saltedcoffii: No longer should be removed. https://github.com/skycocker/chromebrew/pull/5593#issuecomment-817401138

uberhacker: You mean no longer should be used?

lib/const.rb
```diff
@@ -72,15 +72,20 @@
...
 
+CREW_ENV_OPTIONS = "CFLAGS=#{CREW_COMMON_FLAGS} CXXFLAGS=#{CREW_COMMON_FLAGS} FCFLAGS=#{CREW_COMMON_FLAGS} FFLAGS=#{CREW_COMMON_FLAGS} LDFLAGS=#{CREW_LDFLAGS}"
 CREW_OPTIONS = "--prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --mandir=#{CREW_MAN_PREFIX} --build=#{CREW_BUILD} --host=#{CREW_TGT} --target=#{CREW_TGT} --program-prefix='' --program-suffix=''"
-CREW_MESON_OPTIONS = "-Dprefix=#{CREW_PREFIX} -Dlibdir=#{CREW_LIB_PREFIX} -Dmandir=#{CREW_MAN_PREFIX} -Dbuildtype=release -Dc_args='-fuse-ld=lld -pipe' -Dcpp_args='-fuse-ld=lld -pipe'"
-CREW_MESON_LTO_OPTIONS = "-Dprefix=#{CREW_PREFIX} -Dlibdir=#{CREW_LIB_PREFIX} -Dmandir=#{CREW_MAN_PREFIX} -Dbuildtype=release -Db_lto=true -Dcpp_args='-fuse-ld=gold -pipe' -Dc_args='-fuse-ld=gold -pipe'"
+CREW_MESON_OPTIONS = "-Dprefix=#{CREW_PREFIX} -Dlibdir=#{CREW_LIB_PREFIX} -Dmandir=#{CREW_MAN_PREFIX} -Dbuildtype=minsize -Db_lto=true -Dcpp_args=#{CREW_COMMON_FLAGS} -Dc_args=#{CREW_COMMON_FLAGS}"
+CREW_MESON_FNO_LTO_OPTIONS = "-Dprefix=#{CREW_PREFIX} -Dlibdir=#{CREW_LIB_PREFIX} -Dmandir=#{CREW_MAN_PREFIX} -Dbuildtype=minsize -Db_lto=false -Dcpp_args=#{CREW_COMMON_FNO_LTO_FLAGS} -Dc_args=#{CREW_COMMON_FNO_LTO_FLAGS}"
```
uberhacker, line 83: Add to `crew const` output?

lib/const.rb
```diff
@@ -72,15 +72,20 @@
   CREW_BUILD = 'x86_64-cros-linux-gnu'
 end
 
+CREW_COMMON_FLAGS = "'-Os -pipe -fno-lto -fuse-ld=gold'"
+CREW_COMMON_FNO_LTO_FLAGS = "'-Os -pipe -fno-lto -fuse-ld=gold'"
```
uberhacker, line 76: Isn't this a duplicate of `CREW_COMMON_FLAGS`?

saltedcoffii: Oh thank you, one of those is supposed to say `-flto=auto`

crew
saltedcoffii: Fixed.

crew
```diff
@@ -355,8 +355,11 @@ def const (var)
...
+      'CREW_COMMON_FLAGS',
+      'CREW_COMMON_FNO_LTO_FLAGS',
+      'CREW_ENV_OPTIONS',
       'CREW_MESON_OPTIONS',
+      'CREW_MESON_FNO_LTO_OPTIONS',
```
uberhacker, line 362: Out of alphabetical order.  Should go above `CREW_MESON_OPTIONS`.

PR #5597, opened 2021-04-11
fix apr

saltedcoffii:
Apr was installing files into `/usr/local/build-1` like libtool and stuff, leftover from its build. This isn't FHS compatible and doesn't seem to be needed (at least not according to my tests). Also other fixes.

Requires #5593.


packages/apr.rb
```diff
@@ -3,34 +3,22 @@
...
+    system "#{CREW_ENV_OPTIONS} ./configure #{CREW_OPTIONS} \
+            --disable-maintainer-mode \
+            --with-devrandom \
+            --without-sendfile"
     system "make"
   end
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    FileUtils.rm_rf "#{CREW_DEST_PREFIX}/build-1" # Seems residual from build and isn't needed
```
uberhacker, line 22: Was this tested after removing this directory?  There must be some purpose the make install adds it.

PR #5598, opened 2021-04-12
Support source_urls ending with .git

saltedcoffii:
This patch allows chromebrew to clone urls ending in `.git` as the source, using `git_branch 'loremipsum'` as the branch to clone from. Hopefully this will be helpful! Here's a version of packages/dash.rb employing this feature.

```
require 'package'

class Dash < Package
  description 'The Debian Almquist Shell (dash) is a POSIX-compliant shell derived from ash that executes scripts faster than bash and has fewer library dependencies.'
  homepage 'https://salsa.debian.org/debian/dash'
  version '0.5.11.1'
  license 'BSD'
  compatibility 'all'
  source_url 'https://salsa.debian.org/debian/dash.git'
  git_branch 'upstream/0.5.11.1'

#  binary_url ({
#    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-armv7l.tar.xz',
#     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-armv7l.tar.xz',
#       i686: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-i686.tar.xz',
#     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-x86_64.tar.xz',
#  })
#  binary_sha256 ({
#    aarch64: '1b44e3da6bfc8c4e10752b14da464d8a04b748025de69423f944fac8035dbb96',
#     armv7l: '1b44e3da6bfc8c4e10752b14da464d8a04b748025de69423f944fac8035dbb96',
#       i686: 'ac582d141f234881f09860bc400688b0462d92431a5ef739c59687e09cdb9299',
#     x86_64: 'fc193b1d169b2341cd99e329552740656706ae3e22867733eac7d2f7873053b8',
#  })

  depends_on 'libedit'

  def self.build
    system "./configure #{CREW_OPTIONS} --with-libedit"
    system "make"
  end

  def self.install
  end

  def self.check
    system "make", "check"
  end
end
```


crew
```diff
@@ -570,10 +570,11 @@ def download
       end
     end
 
-    system "#{CURL} --retry 3 -#{@verbose}#LC - --insecure \'#{url}\' --output #{filename}"
+    puts File.basename(filename)
```
uberhacker, line 573: I take it this was a debug line you forgot to remove?

crew
```diff
@@ -570,10 +570,11 @@ def download
...
+    system "#{CURL} --retry 3 -#{@verbose}#LC - --insecure \'#{url}\' --output #{filename}" unless File.basename(filename).end_with? '.git'
 
     abort 'Checksum mismatch. :/ Try again.'.lightred unless
-      Digest::SHA256.hexdigest( File.read(filename) ) == sha256sum
+      Digest::SHA256.hexdigest( File.read(filename) ) == sha256sum  unless File.basename(filename).end_with? '.git'
     puts "Archive downloaded".lightgreen
```
uberhacker, line 578: How about?
unless File.basename(filename).end_with? '.git'
   abort 'Checksum mismatch. :/ Try again.'.lightred unless
      Digest::SHA256.hexdigest( File.read(filename) ) == sha256sum
    puts "Archive downloaded".lightgreen
end
```

**Comment by uberhacker** [comment]
Good idea @saltedcoffii!

File: crew (hunk 1 above)
Comment (saltedcoffii) [reply crew:573]: Oh yes it was, thank you.

**Comment by saltedcoffii** [comment]
I tried adding an `is_sourceless` method like `is_fake` but couldn't succeed. I'll try adding one later in a different PR. Should be ready to merge now.

**Comment by saltedcoffii** [comment]
Because of the lib/const.rb problem, merge this *before* #5593

File: lib/package.rb [hunk 3]
```diff
@@ -38,23 +38,31 @@ def self.depends_on (dependency = nil)
...
     end
   end
 
+  def self.get_sha256 (architecture)
+    if !@build_from_source and @binary_sha256 and @binary_sha256.has_key?(architecture)
+      return @binary_sha256[architecture]
+    else
+      return @source_sha256
+    end
+  end
```
uberhacker, line 54: This function appears to be an unnecessary abstraction since it's only called once and adds complexity.

saltedcoffii: It makes things more readable, chromebrew is quite the wreck right now 😂

saltedcoffii: If you're really sure you want it removed, I can revert 2f7c3d0

satmandu:
Can we have this sync a git tag and/or branch?

The git method I've put into several packages already does this.

This way we can just have the package version sometimes act as a proxy for the git tag being pulled.

saltedcoffii:
the `git_branch` variable pulls from a tag or branch. I didn't wanna set version and git_branch equal because sometimes people append v to tags and salsa.debian.org has its own system. Maybe I didn't understand your question though.

lib/package.rb
uberhacker: If we're going to be sticklers for following best practices then we shouldn't be using `@` variables in a function since they act as global instead of local variables.  Is there anything wrong with the current logic?  As a general practice, I don't add a function unless it is necessary to call it multiple times.  Really the only thing that is a wreck is the dependency resolver algorithm.  Most everything else has already been optimized.

satmandu:
> the `git_branch` variable pulls from a tag or branch. I didn't wanna set version and git_branch equal because sometimes people prepend v to tags and salsa.debian.org has its own system. Maybe I didn't understand your question though.

That's fair. That's all I needed.

satmandu:
@saltedcoffii @uberhacker Revised question: Can we have the setup support either a git hash or tag/branch for the git_branch variable?

Here's the code block I've basically been using.
```
    # Turn off git warnings.
    system 'git config --global advice.detachedHead false'
    system 'git config --global init.defaultBranch main'
    @git_dir = 'git_dir_for_cloning'
    @git_hash = 'some_tag_or_branch_or_hash'
    @git_url = 'https://github.com/chromebrew/crew-profile-base'
    FileUtils.rm_rf(@git_dir)
    FileUtils.mkdir_p(@git_dir)
    Dir.chdir @git_dir do
      system 'git init'
      system "git remote add origin #{@git_url}"
      system "git fetch --depth 1 origin #{@git_hash}"
      system 'git checkout FETCH_HEAD'
```

This way you can also specify a hash to be cloning as well.

satmandu:
So you would have to have (assuming this used the shorter `pkg.git_tag` in lieu of `pkg.git_branch`):
```
    when /\.git$/i
      puts "Cloning with 'git', this may take a while...".lightgreen
      FileUtils.mkdir extract_dir unless Dir.exist?(extract_dir)
      system 'git config --global advice.detachedHead false'
      system 'git config --global init.defaultBranch main'
      Dir.chdir extract_dir do
        system 'git init'
        system "git remote add origin #{@pkg.source_url}"
        system "git fetch --depth 1 origin #{@pkg.git_tag}"
        system 'git checkout FETCH_HEAD'
      end
```

saltedcoffii:
> So you would have to have (assuming this used the shorter `pkg.git_tag` in lieu of `pkg.git_branch`):
> 
> ```
>     when /\.git$/i
>       puts "Cloning with 'git', this may take a while...".lightgreen
>       FileUtils.mkdir extract_dir unless Dir.exist?(extract_dir)
>       system 'git config --global advice.detachedHead false'
>       system 'git config --global init.defaultBranch main'
>       Dir.chdir extract_dir do
>         system 'git init'
>         system "git remote add origin #{@pkg.source_url}"
>         system "git fetch --depth 1 origin #{@pkg.git_tag}"
>         system 'git checkout FETCH_HEAD'
>       end
> ```

What's the advantage of doing this over what I'm doing?

saltedcoffii:
Also I feel like we shouldn't globally turn off git warnings since some users may want those warnings from git for other purposes.

satmandu:
> What's the advantage of doing this over what I'm doing?

You can specify a git hash as well as a tag and branch.

Being able to use a specific git hash is really useful, since it is more specific than a branch, and in many repos tags don't happen very often.

satmandu:
> Also I feel like we shouldn't globally turn off git warnings since some users may want those warnings from git for other purposes.

Can we turn them off locally for just these commands?

saltedcoffii:
> > Also I feel like we shouldn't globally turn off git warnings since some users may want those warnings from git for other purposes.
> 
> Can we turn them off locally for just these commands?

Yes, after running `git init`, we can run `git config advice.detachedHead false` `git config --global init.defaultBranch master` within the repository.

saltedcoffii:
Okay, after the next push it'll be ready to merge. This demonstrates dash using `git_hash`:
```
require 'package'

class Dash < Package
  description 'The Debian Almquist Shell (dash) is a POSIX-compliant shell derived from ash that executes scripts faster than bash and has fewer library dependencies.'
  homepage 'https://salsa.debian.org/debian/dash'
  version '0.5.11.1'
  license 'BSD'
  compatibility 'all'
  source_url 'https://salsa.debian.org/debian/dash.git'
  git_tag 'master'
  git_hash 'c484ba14'

#  binary_url ({
#    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-armv7l.tar.xz',
#     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-armv7l.tar.xz',
#       i686: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-i686.tar.xz',
#     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-x86_64.tar.xz',
#  })
#  binary_sha256 ({
#    aarch64: '1b44e3da6bfc8c4e10752b14da464d8a04b748025de69423f944fac8035dbb96',
#     armv7l: '1b44e3da6bfc8c4e10752b14da464d8a04b748025de69423f944fac8035dbb96',
#       i686: 'ac582d141f234881f09860bc400688b0462d92431a5ef739c59687e09cdb9299',
#     x86_64: 'fc193b1d169b2341cd99e329552740656706ae3e22867733eac7d2f7873053b8',
#  })

  depends_on 'libedit'

  def self.build
    system './autogen.sh'
    system "./configure #{CREW_OPTIONS} --with-libedit"
    system "make"
  end

  def self.install
  end

  def self.check
    system "make", "check"
  end
end
```

This demonstrates dash using `git_tag`
```
require 'package'

class Dash < Package
  description 'The Debian Almquist Shell (dash) is a POSIX-compliant shell derived from ash that executes scripts faster than bash and has fewer library dependencies.'
  homepage 'https://salsa.debian.org/debian/dash'
  version '0.5.11.1'
  license 'BSD'
  compatibility 'all'
  source_url 'https://salsa.debian.org/debian/dash.git'
  git_tag 'upstream/0.5.11.1'

#  binary_url ({
#    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-armv7l.tar.xz',
#     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-armv7l.tar.xz',
#       i686: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-i686.tar.xz',
#     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-x86_64.tar.xz',
#  })
#  binary_sha256 ({
#    aarch64: '1b44e3da6bfc8c4e10752b14da464d8a04b748025de69423f944fac8035dbb96',
#     armv7l: '1b44e3da6bfc8c4e10752b14da464d8a04b748025de69423f944fac8035dbb96',
#       i686: 'ac582d141f234881f09860bc400688b0462d92431a5ef739c59687e09cdb9299',
#     x86_64: 'fc193b1d169b2341cd99e329552740656706ae3e22867733eac7d2f7873053b8',
#  })

  depends_on 'libedit'

  def self.build
    system "./configure #{CREW_OPTIONS} --with-libedit"
    system "make"
  end

  def self.install
  end

  def self.check
    system "make", "check"
  end
end
```

saltedcoffii:
Alright this is ready to merge!

saltedcoffii:
Couple notes:
- I switched from using a random alphanumeric string with the current time in UTC for the extract dir, it sped up the install by a couple hundredths of a second and made more logical sense.
- I moved the git clone commands from `def unpack` to `def download` as it made more sense there
- I rearranged the entire `def download` function to support adding more functions, like perhaps a version of what we do with git for `hg` and `svn`.
- I added support for caching with git repositories.

saltedcoffii:
Finally ready to merge @uberhacker. I tested every single possible thing to test.

crew
```diff
@@ -111,11 +110,13 @@ String.use_color = args["--color"] || !args["--no-color"]
...
   @verbose = 'v'
-  @mv_verbose = '-v'
+  @short_verbose = '-v'
+  @long_verbose = '--verbose'
 else
   @fileutils_verbose = false
   @verbose = ''
-  @mv_verbose = ''
+  @short_verbose = ''
+  @long_verbose = '--verbose'
```
uberhacker, line 119: Should this be `@long_version = ''`?

packages/dash.rb
```diff
@@ -6,21 +6,21 @@ class Dash < Package
...
+#     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-armv7l.tar.xz',
+#       i686: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-i686.tar.xz',
+#     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/dash-0.5.11.1-chromeos-x86_64.tar.xz',
+#  })
+#  binary_sha256 ({
+#    aarch64: '1b44e3da6bfc8c4e10752b14da464d8a04b748025de69423f944fac8035dbb96',
+#     armv7l: '1b44e3da6bfc8c4e10752b14da464d8a04b748025de69423f944fac8035dbb96',
+#       i686: 'ac582d141f234881f09860bc400688b0462d92431a5ef739c59687e09cdb9299',
+#     x86_64: 'fc193b1d169b2341cd99e329552740656706ae3e22867733eac7d2f7873053b8',
+#  })
```
uberhacker, line 23: Can restore the sourceforge binaries?

saltedcoffii: Oh, yes let me do that.

lib/const.rb
```diff
@@ -36,10 +36,13 @@
...
 
 if ENV['CREW_CACHE_DIR'].to_s == ''
-  CREW_CACHE_DIR = HOME + '/.cache/crewcache/'
+  CREW_CACHE_DIR = HOME + '/.cache/crewcache'
 else
   CREW_CACHE_DIR = ENV['CREW_CACHE_DIR']
 end
+FileUtils.mkdir_p CREW_CACHE_DIR unless Dir.exist? CREW_CACHE_DIR
+
+CREW_IS_CACHING = ENV['CREW_CACHE_OPT']
```
uberhacker, line 45: It appears to be a bit confusing to have an environment variable and crew constant with different names.  Can we just rename one or the other to match?

crew
saltedcoffii: No, but @short_verbose should be equal to '-v'

lib/const.rb
saltedcoffii: Sure.

lib/package.rb
uberhacker: Not a fun of putting this logic in a function.

crew
uberhacker: On second thought, can we just remove this variable?  It doesn't appear to be used anywhere.

saltedcoffii: nevermind it is the way it's supposed to be. And no, `--long-verbose` since the argument is --verbose not --version

saltedcoffii: I'm planning on going through all the package files at some point and making them compatible with crew's `-v` option, so we'll need it then. I'll take it out for now though.

crew
```diff
@@ -10,7 +10,6 @@ require 'uri'
 require 'digest/sha2'
 require 'json'
 require 'fileutils'
-require 'securerandom'
```
uberhacker, line 13: We want to reverse the random temp file change?

saltedcoffii: I changed `SecureRandom.alphanumeric(8)` to `Time.now.utc.strftime('%Y%m%d%H%M%S'`
The current time is entropy enough for us, and makes more sense. Because Time is built in, there's no require statement for it.

satmandu:
Can you fix conflicts so this can be merged?

saltedcoffii:
@satmandu conflicts fixed :)

bin/crew
```diff
@@ -681,7 +723,7 @@ def compress_doc (dir)
 
   if Dir.exist? dir
     system "find #{dir} -type f ! -perm -200 | xargs -r chmod u+w"
-    system "compressdoc --gzip -9#{@verbose} #{dir}"
+    system "compressdoc --gzip -9 -#{@verbose} #{dir}"
```
uberhacker, line 726: I don't think we want this change.  `#{@verbose}` could be an empty string.

packages/dash.rb
```diff
@@ -2,12 +2,12 @@
 
 class Dash < Package
   description 'The Debian Almquist Shell (dash) is a POSIX-compliant shell derived from ash that executes scripts faster than bash and has fewer library dependencies.'
-  homepage 'https://salsa.debian.org/debian/dash'
+  homepage 'https://invisible-island.net/mawk/'
```
uberhacker, line 5: Why the homepage change?  This is mawk, not dash.

bin/crew
saltedcoffii: Good point.

packages/dash.rb
saltedcoffii: because i'm dumb and my clipboard is worse 🤕

bin/crew
```diff
@@ -556,77 +561,114 @@ def download
...
   end
   return {source: source, filename: filename}
 end
 
 def unpack (meta)
-  extract_dir = meta[:filename] + '.' + SecureRandom.alphanumeric(8) + '.dir'
   target_dir = nil
   Dir.chdir CREW_BREW_DIR do
-    FileUtils.mkdir_p extract_dir, verbose: @fileutils_verbose
+    FileUtils.mkdir_p $extract_dir, verbose: @fileutils_verbose
```
uberhacker, line 641: Why the `$`?  This is ruby, not bash.

saltedcoffii: the $ makes the variable global.

uberhacker: Why does it need to be global?

saltedcoffii: Because it's called in `def download` and `def unpack`, and it needs to be the same since the variable contains entropy.

supechicken: Better to use instance variable (make it consistent with other variables)? It also shared between functions

saltedcoffii: @supechicken, the problem is that the download function isn't in a class.

uberhacker: @saltedcoffii: crew has worked for years without global variables.  Why do we suddenly need them now?

supechicken: It also works without class (see `set_package` section)

uberhacker: Please look into fixing https://github.com/skycocker/chromebrew/issues/5603.  That should be a priority.

saltedcoffii: Okay, I switched the $ to an @ and it works, thanks for the suggestion!

saltedcoffii:
I think this is ready to merge although uberhacker usually proves me wrong. After this gets merged I'm going to work on supporting multiple download urls so we can avoid the `curl -#LO Digest::SHA256 File.read == 'fasdkjfasdf'` mess

uberhacker:
I'm going to hold off on merging any more PRs until #5603 is fixed.

saltedcoffii:
okay @uberhacker #5603 is fixed!

lib/const.rb
```diff
@@ -36,10 +36,13 @@
...
 
 if ENV['CREW_CACHE_DIR'].to_s == ''
-  CREW_CACHE_DIR = HOME + '/.cache/crewcache/'
+  CREW_CACHE_DIR = HOME + '/.cache/crewcache'
 else
   CREW_CACHE_DIR = ENV['CREW_CACHE_DIR']
 end
+FileUtils.mkdir_p CREW_CACHE_DIR unless Dir.exist? CREW_CACHE_DIR
+
+CREW_IS_CACHING = ENV['CREW_IS_CACHING']
```
uberhacker, line 45: I'm thinking a better name would be `CREW_CACHE_ENABLED`.

bin/crew
```diff
@@ -556,77 +561,114 @@ def download
...
+        system 'git config init.defaultBranch master'
+        system "git remote add origin #{@pkg.source_url}", exception: true
+        if @pkg.git_hash
+          system "git fetch origin"
+          system "git checkout #{@pkg.git_hash}", exception: true
+        else
+          system "git fetch --depth 1 origin #{@pkg.git_tag}", exception: true
+        end
+        system 'git checkout FETCH_HEAD'
+        puts "Repository downloaded".lightgreen
```
uberhacker, line 625: This whole process seems like overkill.

uberhacker: I would just do `git clone --depth 1 #{@pkg.source_url} && cd dir && git checkout <branch or hash>` and wash my hands but that is just me.  Remember, this is all temporary for the build anyway.

saltedcoffii: here's my reasoning for that:

uberhacker: Yes, it makes sense for performance reasons only but all the `git init` and `git config` stuff is not needed imho.

bin/crew
```diff
@@ -556,77 +561,114 @@ def download
...
+      end
+      # Download file if not needed
+      Dir.mkdir @extract_dir
+      Dir.chdir @extract_dir do
+        system 'git init'
+        system 'git config advice.detachedHead false'
+        system 'git config init.defaultBranch master'
+        system "git remote add origin #{@pkg.source_url}", exception: true
+        if @pkg.git_hash
+          system "git fetch origin"
```
uberhacker, line 619: Doesn't this take a long time too for large repos?

bin/crew
```diff
@@ -556,77 +561,114 @@ def download
...
+        else
+          puts 'Cannot find cached repository. :/ Will download.'.lightred
+        end
+      end
+      # Download file if not needed
+      Dir.mkdir @extract_dir
+      Dir.chdir @extract_dir do
+        system 'git init'
+        system 'git config advice.detachedHead false'
+        system 'git config init.defaultBranch master'
```
uberhacker, line 616: This is not necessarily a good assumption.  Not all repos default to master and some repos don't even have a master branch.

lib/package.rb
saltedcoffii: where would you put it, the way I have it the code is much more readable. Chromebrew, although it may be optimized, it's just not very readable at all.

bin/crew
saltedcoffii: No, but it doesn't matter anyway. This is just to suppress a warning. Git won't error or cause a problem if you set the default branch to master and the repository doesn't have a master branch.

bin/crew
saltedcoffii: Yes, but it's the best solution I can think of. If you have a better idea, let me know.

bin/crew
saltedcoffii: the `git config` stuff is to suppress confusing warnings that actually aren't a problem at all for our purposes.

bin/crew
```diff
@@ -556,77 +561,114 @@ def download
...
+        system "git remote add origin #{@pkg.source_url}", exception: true
+        if @pkg.git_hash
+          system "git fetch origin"
+          system "git checkout #{@pkg.git_hash}", exception: true
```
uberhacker, line 620: Do we even need this line if the same thing happens on line 624?

lib/const.rb
saltedcoffii: I like that, yes. Lemme make that change.

packages/dash.rb
```diff
@@ -2,12 +2,12 @@
...
   description 'The Debian Almquist Shell (dash) is a POSIX-compliant shell derived from ash that executes scripts faster than bash and has fewer library dependencies.'
-  homepage 'https://salsa.debian.org/debian/dash'
+  homepage 'https://salsa.debian.org/debian/dash/'
   version '0.5.11.1'
   license 'BSD'
   compatibility 'all'
-  source_url 'https://salsa.debian.org/debian/dash/-/archive/upstream/0.5.11.1/dash-upstream-0.5.11.1.tar.bz2'
-  source_sha256 '4bb75944bb47fa6d1cf1e8a3ff941f6f1fb23497b553446e9f615d52d07ef1e7'
+  source_url 'https://salsa.debian.org/debian/dash.git'
+  git_tag 'upstream/0.5.11.2'
```
uberhacker, line 10: Can't tags, branches and hashes all use the same variable generically?

saltedcoffii: branches and tags use the same variable, but hashes need their own. One needs to specify the tag that the hash is on.

bin/crew
saltedcoffii: It's not the same thing, we do need both lines.

packages/dash.rb
uberhacker: A tag is just a human-readable alias for a hash.  I'm not understanding the logic.

bin/crew
uberhacker: Why would you need to checkout twice?  That doesn't make any sense.  Maybe line 624 should be replaced with line 620?

bin/crew
uberhacker: How about `git clone`?

packages/dash.rb
saltedcoffii: Okay. `-b` allows you to specifically clone a branch or tag. `-b` doesn't work on hashes.

`git_tag` works with either a branch or tag.
`git_hash` works with a tag or hash.

It's more bandwidth efficient to use `git_tag` for tags.

bin/crew
saltedcoffii: What if `git_hash` is empty?

bin/crew
saltedcoffii: Then it spits out a concerning warning.

bin/crew
saltedcoffii: It does the exact same thing and spits out concerning warnings, unfortunately.

bin/crew
uberhacker: How about this logic?  Trying to simplify...
```
if @pkg.git_branch.is_a?(Hash)
  system "git clone #{@pkg.source_url}"
else
  system "git clone --depth 1 #{@pkg.source_url}"
end
Dir.chdir File.basename("#{@pkg.source_url}", ".git") do
  system "git checkout #{@pkg.git_branch}"
end
```
@pkg.git_branch could be a branch, tag or hash.  Easier?

saltedcoffii: What if someone needs to specify a branch/tag on a specific branch?

bin/crew
uberhacker: If you specified master and that branch didn't exist, wouldn't you still get an error?

bin/crew
uberhacker: Well, I guess you do have a point to that argument.  The point I'm trying to make is `git clone` can replace all that `git init`, `git fetch`, `git remote add`, etc. extra unnecessary stuff.

bin/crew
saltedcoffii: Nope!

lib/package.rb
uberhacker: Now that I thought about it a bit, it does makes sense to have this function.  It is similar to the get_url function above.

bin/crew
saltedcoffii: @satmandu what do you think about that? You were the one who suggested the git init and git config sequence.

satmandu: @uberhacker git clone is simpler, but the git init, fetch, etc nonsense is **insanely powerful** because then you don't need ANY other logic to just fetch a git hash - and we really don't want to do git fetches of branches since those are moving targets, right?

saltedcoffii: That is true, it would be better to only fetch tags and hashes.

packages/dash.rb
satmandu: @saltedcoffii @uberhacker with the aforementioned logic I shared, the same variable can be used for tags, branches, and hashes.

"_logic_":
```
    # Turn off git warnings.
    system 'git config --global advice.detachedHead false'
    system 'git config --global init.defaultBranch main'
    @git_dir = 'git_dir_for_cloning'
    @git_hash = 'some_tag_or_branch_or_hash'
    @git_url = 'https://github.com/chromebrew/crew-profile-base'
    FileUtils.rm_rf(@git_dir)
    FileUtils.mkdir_p(@git_dir)
    Dir.chdir @git_dir do
      system 'git init'
      system "git remote add origin #{@git_url}"
      system "git fetch --depth 1 origin #{@git_hash}"
      system 'git checkout FETCH_HEAD'
```

The `@git_hash` variable here covers hash, tag, or branch, as I understand it. I don't really care what the variable is called.

I don't have a solution for slow internet though... Other than to say, we've all been there. 

But this should be optimized for end user building, and for the end user, there's no reason to download an entire git clone... this should be MUCH faster,

bin/crew
satmandu: Also, that could also be this in a couple of  lines,  which is really quick:

```
Dir.chdir @extract_dir do
        system 'git init'
        system 'git config advice.detachedHead false'
        system 'git config init.defaultBranch master'
        system "git fetch --depth 1 origin #{@pkg.git_hashtag}"
        system 'git checkout FETCH_HEAD'
        puts "Repository downloaded".lightgreen
```
I would actually lean against officially supporting a branch at all... You want to support commits... otherwise, you have no idea what you are downloading.

saltedcoffii: you can't clone a hash or tag like that, only a branch, at least from the testing I've done.

uberhacker: Alright, this is some really good feedback.  It looks messy but after your explanation, it makes more sense.

uberhacker: I will let @satmandu make the executive decision to merge this or not.

satmandu:
@saltedcoffii If you just want to collapse the git tag/branch/hash logic so there's one variable, we can then do some testing...

saltedcoffii:
> @saltedcoffii If you just want to collapse the git tag/branch/hash logic so there's one variable, we can then do some testing...

What if someone needs to clone a specific tag/hash on a specific branch? That's happened to me a lot on salsa.debian.org

satmandu:
@saltedcoffii I haven't had to switch branches to use a hash from a specific branch....

But a tag from a branch is just `branch/tag` no?

saltedcoffii:
Yes, but not a hash on a specific branch, unfortunately.

bin/crew
satmandu: working on this...

satmandu:
@saltedcoffii  @uberhacker Please look at the changes I made. I changed the variable to `git_hashtag` which covers both hash and tag. I did nothing with branches at all.

I also now have it automatically checking out submodules, which eliminates some extra logic in some package files!

Working package:
```
require 'package'

class Libsodium < Package
  description 'A modern, portable, easy to use crypto library'
  homepage 'https://libsodium.org'
  version '1.0.18-7168'
  license 'ISC'
  compatibility 'all'
  source_url 'https://github.com/jedisct1/libsodium.git'
  git_hashtag '7168944f16e12c1b630e66a7b9be0802d5b564dd'

  binary_url({
    aarch64: 'https://downloads.sourceforge.net/project/chromebrew/armv7l/libsodium-1.0.18-7168-chromeos-armv7l.tar.xz',
     armv7l: 'https://downloads.sourceforge.net/project/chromebrew/armv7l/libsodium-1.0.18-7168-chromeos-armv7l.tar.xz',
       i686: 'https://downloads.sourceforge.net/project/chromebrew/i686/libsodium-1.0.18-7168-chromeos-i686.tar.xz',
     x86_64: 'https://downloads.sourceforge.net/project/chromebrew/x86_64/libsodium-1.0.18-7168-chromeos-x86_64.tar.xz'
  })
  binary_sha256({
    aarch64: '8cf24abf23f1ec6b19052e29b5cc80db8f2b6917dae193827c99a02349839221',
     armv7l: '8cf24abf23f1ec6b19052e29b5cc80db8f2b6917dae193827c99a02349839221',
       i686: 'ab1a37ec01e04271deca45cfb8943dc8d379e12c4d154a4d190824e1de6c0daa',
     x86_64: 'f5ad9b26a80254a30e8aaced54eeb19e8962980b888df9d904a6377d7dce9cff'
  })

  def self.build
      system 'NOCONFIGURE=1 ./autogen.sh'
        ./configure #{CREW_OPTIONS}"
      system 'make'
  end

  def self.install
  end

  def self.check
    Dir.chdir @git_dir do
      system 'make', 'check'
    end
  end
end
```
Another working package using a tag:
```
require 'package'

class Gtk_vnc < Package
  description 'VNC viewer widget for GTK'
  homepage 'https://wiki.gnome.org/Projects/gtk-vnc'
  version '1.2.0'
  license 'LGPL-2.1+'
  compatibility 'all'
  source_url 'https://gitlab.gnome.org/GNOME/gtk-vnc.git'
  git_hashtag "v#{version}"

  binary_url({
    aarch64: 'https://downloads.sourceforge.net/project/chromebrew/armv7l/gtk_vnc-1.2.0-chromeos-armv7l.tar.xz',
     armv7l: 'https://downloads.sourceforge.net/project/chromebrew/armv7l/gtk_vnc-1.2.0-chromeos-armv7l.tar.xz',
       i686: 'https://downloads.sourceforge.net/project/chromebrew/i686/gtk_vnc-1.2.0-chromeos-i686.tar.xz',
     x86_64: 'https://downloads.sourceforge.net/project/chromebrew/x86_64/gtk_vnc-1.2.0-chromeos-x86_64.tar.xz'
  })
  binary_sha256({
    aarch64: '62d85534080b53b4b3e0a20c58979db9e550a035414821d210a45fc8965da0ab',
     armv7l: '62d85534080b53b4b3e0a20c58979db9e550a035414821d210a45fc8965da0ab',
       i686: '609017642a8be0a5d309b05d07aed5665760eca14af26075c3b719fef3736dc8',
     x86_64: '5001d552bfa4155bf2c77a1e1a7ee4714125566db0b9ce1e3e3c5e88ace8057d'
  })

  depends_on 'cairo'
  depends_on 'glib'
  depends_on 'gobject_introspection'
  depends_on 'gtk3'
  depends_on 'libgcrypt'
  depends_on 'pulseaudio'

  def self.patch
    system "sed -i 's,-fstack-protector-strong,-fno-stack-protector,' meson.build"
  end

  def self.build
    system "meson \
      #{CREW_MESON_OPTIONS} \
      builddir"
    system 'meson configure builddir'
    system 'ninja -C builddir'
  end

  def self.install
  end
end
```

uberhacker:
Yes, that looks a lot nicer with the git commands abstracted.

satmandu:
Good to merge.

saltedcoffii:
@satmandu cloning hashes is broken. See this example file:
```
require 'package'

class Ruby_docopt < Package
  description 'Docopt parses command line arguments from nothing more than a usage message.'
  homepage 'http://docopt.org/'
  version '0.6.1'
  license 'MIT'
  compatibility 'all'
  source_url 'https://github.com/docopt/docopt.rb.git'
  git_hashtag '504f310'

  depends_on 'ruby'

  def self.patch
    # https://github.com/docopt/docopt.rb/pull/37
    system "sed -i 's:0.6.0:0.6.1:' lib/docopt.rb"
    system "sed -i 's:0.6.0:0.6.1:' docopt.gemspec"
  end

  def self.build
    system 'rake build'
  end

  def self.check
    system 'rake check'
  end

  def self.install
    system "gem install --build=#{CREW_DEST_DIR} pkg/docopt-0.6.1.gem"
  end
end
```

saltedcoffii:
Hmm, the hashing thing seems to work with the long hash `504f310c5fd66e9f3c36b078156e5f44573ff1d3`

satmandu:
Always use a long hash? I've never used a short hash aside from as a programmatic short indicator of the longer hash.

I wasn't even aware it was supported for cloning!

saltedcoffii:
In the way I had the code, the short hash worked, however now it just seems that the long hash is required.

satmandu:
If you can figure out a way to make the short hash work again...

But I think we should use long hashes in our package files. 

At least that's what Arch does...

saltedcoffii:
Eh it's fine it works well enough. If it bothers you I can revert your changes but uberhacker didn't seem to like them so I think we'll struggle getting the short hash to work without delaying PRs another four days

uberhacker:
Just remember every time you take another character off the hash, you run the risk of duplicating another hash.  Better to just use the long hash.  I always cut and paste hashes anyway so it doesn't cost any more to have the whole hash.

When git hashes clash...
cFkiFMDg3iFoI

PR #5599, opened 2021-04-12
meson => 0.57.2

satmandu:
Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/meson.rb
+++ b/packages/meson.rb
@@ -22,6 +22,8 @@ class Meson < Package
        i686: 'f8a173fefc3b188208fbd6da61f3f9b0e94db62b8bbdb18573b8d740547ade82',
      x86_64: '4f6668f8cf6463fbe20e9d7de20d818548ab53ec69c7dae1e5700f57e6d6b816'
   })
+  
+  depends_on 'ninja'
 
   def self.install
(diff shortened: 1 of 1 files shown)


packages/meson.rb
```diff
@@ -3,26 +3,24 @@
...
   homepage 'http://mesonbuild.com/'
-  @_ver = '0.57.1'
+  @_ver = '0.57.2'
   version @_ver
   license 'Apache-2.0'
   compatibility 'all'
   source_url "https://github.com/mesonbuild/meson/archive/#{@_ver}.tar.gz"
-  source_sha256 '0c043c9b5350e9087cd4f6becf6c0d10b1d618ca3f919e0dcca2cdf342360d5d'
-
-  depends_on 'ninja'
```
uberhacker, line 13: Why was this dependency removed?  Don't we still need ninja with meson?

satmandu: At the time this was built, ninja was still part of `core_packages.txt`, so my scripts dutifully removed the dependency!

I'll put it back in...

uberhacker: Adjust your script to ignore any dependencies from packages in `core_packages.txt`. :)

PR #5600, opened 2021-04-12
fix directory already exists error

saltedcoffii:
Chromebrew has this issue where if two packages have the same source and are being installed during the same "session" (CREW_BREW_DIR) doesn't get cleared, it will cause an error. This was a big problem with bsdgames (which is still in the works on my end btw). This fixes that by generating eight random directory compatible ascii characters between the tarball name and .dir. Tested and works.


uberhacker:
This is (or should be) a very rare occurrence.  When a package build is done, the working directory should be removed automagically.

uberhacker:
I do agree there should be a way to handle `/dev/null` sources since packages could conflict during install if binaries are not available.

saltedcoffii:
How would you suggest fixing this bug @uberhacker? We can't, in good conscience, just leave it.

uberhacker:
Not sure but we really should try to use Ruby instead of a shell hack.  The goal should be elegance. :)

saltedcoffii:
Okay, I'll see how to generate random ASCII characters with ruby

satmandu:
```
len = 8
SecureRandom.alphanumeric(len)
=> "larHSsgL"
```
via https://stackoverflow.com/a/48521297

saltedcoffii:
So securerandom uses `/dev/urandom` as a backend (on unix systems that is) so this just rubyizes that statement and also speeds it up a little since it doesn't open a shell.

saltedcoffii:
However, we could do something without random numbers:
```ruby
  extract_dir = meta[:filename] + '.' + @pkg.name + '.dir'
```

PR #5602, opened 2021-04-12
move crew to bin/crew

saltedcoffii:
This is needed for further changes I'm going to make later such as making chromebrew a package and using standard ruby tools to install it. Works on x86_64.

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1,5 +1,5 @@
 #!/usr/bin/env ruby
-require_relative 'lib/color'
+require_relative '../lib/color'
 
 # Disallow sudo
 abort 'Chromebrew should not be run as root.'.lightred if Process.uid == 0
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -188,7 +188,7 @@ sudo ldconfig > /dev/null 2> /dev/null || true
 
 # create symlink to 'crew' in ${CREW_PREFIX}/bin/
 rm -f "${CREW_PREFIX}/bin/crew"
-ln -s "../lib/crew/crew" "${CREW_PREFIX}/bin/"
+ln -s "../lib/crew/bin/crew" "${CREW_PREFIX}/bin/"
```
uberhacker, line 191: You can effectively remove line 190 with:
```
ln -sf ...
```

bin/crew
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env ruby
-require_relative 'lib/color'
+require_relative '../lib/color'
```
uberhacker, line 2: Bump crew version in `lib/const.rb`?

uberhacker:
This PR appears to be a duplicate of #5601?

PR #5606, opened 2021-04-12
Crew Package Dev workaround

satmandu:
- Allow for overriding package conflicts when developing packages which are meant to replace other packages.
- Stop packaging "#{CREW_DEST_PREFIX}/share/info/dir" with a global exclusion during package building

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/crew
+++ b/crew
@@ -682,6 +682,14 @@ end
 
 def prepare_package (destdir)
   Dir.chdir destdir do
+    # Avoid /usr/local/share/info/dir{.gz} file conflict:
+    # The install-info program maintains a directory of installed 
+    # info documents in /usr/share/info/dir for the use of info 
+    # readers. This file must not be included in packages other 
+    # than install-info.
+    # https://www.debian.org/doc/debian-policy/ch-docs.html#info-documents
+    FileUtils.rm "#{CREW_DEST_PREFIX}/share/info/dir" if File.exist?("#{CREW_DEST_PREFIX}/share/info/dir")
+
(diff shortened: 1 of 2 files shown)


crew
```diff
@@ -682,6 +682,14 @@ end
 
 def prepare_package (destdir)
   Dir.chdir destdir do
+    # Avoid /usr/local/share/info/dir{.gz} file conflict:
+    # The install-info program maintains a directory of installed 
+    # info documents in /usr/share/info/dir for the use of info 
+    # readers. This file must not be included in packages other 
+    # than install-info.
+    # https://www.debian.org/doc/debian-policy/ch-docs.html#info-documents
+    FileUtils.rm "#{CREW_DEST_PREFIX}/share/info/dir" if File.exist?("#{CREW_DEST_PREFIX}/share/info/dir")
```
uberhacker, line 691: Should we also include `dir.gz` also in case it was already compressed?

crew
```diff
@@ -713,7 +721,7 @@ def prepare_package (destdir)
         puts f.lightred
         pp = p
       end
-      abort
+      abort unless ENV['CREW_CONFLICTS_ONLY_ADVISORY'] == '1'
```
uberhacker, line 724: I'm not a fan of allowing overrides like this.  You can easily miss conflicts if the environment variable remains set for other packages.

crew
satmandu: That's a good idea.

PR #5610, opened 2021-04-13
[New Package] flatseal

saltedcoffii:
Flatseal is a flatpak permissions manager. Although its original intention was to be installed as a flatpak, the AUR offers a non-sandboxed version. Alpine followed suit recently as well, and I thought, why not? It's smaller than a sandboxed version anyway.

Flatpak isn't a dependency because flatseal doesn't require it to run, however it's kinda useless without flatpak? (may be considered a logical dependency, but the dependency is weak enough I left it out.)

Works on x86_64, theoretically works on all arches we support.

### DIFF
--- a/packages/flatseal.rb
+++ b/packages/flatseal.rb
@@ -13,6 +13,7 @@ class Flatseal < Package
   depends_on 'desktop_file_utils'
   depends_on 'gjs'
   depends_on 'gtk3'
+  depends_on 'sommelier'
 
   def self.build
     system "meson #{CREW_MESON_OPTIONS} builddir"
--- a/packages/flatseal.rb
+++ b/packages/flatseal.rb
@@ -1,0 +1,30 @@
+require 'package'
+
+class Flatseal < Package
+  description 'Flatseal is a graphical utility to review and modify permissions from your Flatpak applications.'
+  homepage 'https://github.com/tchx84/Flatseal'
+  version '1.6.8'
+  license 'GPL-3+'
+  compatibility 'all'
+  source_url 'https://github.com/tchx84/Flatseal/archive/v1.6.8.tar.gz'
+  source_sha256 '28d05995effa2858483283cbc9fb54c55bf4dcb6267e4d7d495a32ff724806b1'
+
+  depends_on 'libhandy'
+  depends_on 'desktop_file_utils'
+  depends_on 'gjs'
+  depends_on 'gtk3'
+
+  def self.build
+    system "meson #{CREW_MESON_OPTIONS} builddir"
+    system "samu -C builddir"
+  end
+
+  def self.check
+    system "samu -C builddir test"
+  end
+
+  def self.install
+    system "DESTDIR=#{CREW_DEST_DIR} samu -C builddir install"
+    FileUtils.symlink 'com.github.tchx84.Flatseal', "#{CREW_DEST_PREFIX}/bin/flatseal"
+  end
+end
--- a/packages/meson.rb
+++ b/packages/meson.rb
@@ -22,8 +22,9 @@ class Meson < Package
        i686: 'f8a173fefc3b188208fbd6da61f3f9b0e94db62b8bbdb18573b8d740547ade82',
      x86_64: '4f6668f8cf6463fbe20e9d7de20d818548ab53ec69c7dae1e5700f57e6d6b816'
   })
-  
+
   depends_on 'ninja'
+  depends_on 'samurai'
 
   def self.install
     system "python3 setup.py install --prefix #{CREW_PREFIX} --root #{CREW_DEST_DIR}"


packages/flatseal.rb
```diff
@@ -0,0 +1,30 @@
...
+  version '1.6.8'
+  license 'GPL-3+'
+  compatibility 'all'
+  source_url 'https://github.com/tchx84/Flatseal/archive/v1.6.8.tar.gz'
+  source_sha256 '28d05995effa2858483283cbc9fb54c55bf4dcb6267e4d7d495a32ff724806b1'
+
+  depends_on 'libhandy'
+  depends_on 'desktop_file_utils'
+  depends_on 'gjs'
+  depends_on 'gtk3'
```
uberhacker, line 15: Do we need `depends_on 'samurai' => :build`?

packages/flatseal.rb
```diff
@@ -0,0 +1,30 @@
...
+    system "samu -C builddir"
+  end
+
+  def self.check
+    system "samu -C builddir test"
+  end
+
+  def self.install
+    system "DESTDIR=#{CREW_DEST_DIR} samu -C builddir install"
+    FileUtils.symlink 'com.github.tchx84.Flatseal', "#{CREW_DEST_PREFIX}/bin/flatseal"
```
uberhacker, line 28: Shouldn't `'com.github.tchx84.Flatseal'` have a path?  The working directory is different than the actual installed directory.

saltedcoffii: #TodayILearned, when making a symlink with FileUtils or `ln`, the first argument is where the symlink should point and it can be *absolute or relative*. So this just makes a symlink called `flatseal` at `/usr/local/bin` that points to `com.github.tchx64.Flatseal`. It could also be
```ruby
FileUtils.symlink './com.github.tchx84.Flatseal', "#{CREW_DEST_PREFIX}/bin/flatseal"
```
but that's just extraneous.

packages/flatseal.rb
saltedcoffii: Samurai should be a dependency of meson since we're transitioning from ninja to samurai.

uberhacker: But that currently isn't the way we have it configured.

packages/flatseal.rb
uberhacker: No, I mean the destination directory for `com.github.tchx84.Flatseal` is somewhere under `CREW_PREFIX`, not `CREW_DEST_DIR`.  Was this even tested?

packages/flatseal.rb
saltedcoffii: Fixed!

packages/flatseal.rb
saltedcoffii: This was tested and it works.

packages/flatseal.rb
uberhacker: What was fixed?

saltedcoffii: 8d4a3a7

uberhacker: Oh, I see you added samurai to meson below.

packages/flatseal.rb
uberhacker: I'm not sure how this would work since the symlink doesn't include a path.

saltedcoffii: The first argument to FileUtils.symlink is the path. `./` is assumed.

uberhacker: I don't think you understand.  That is not a path, that is just relative, not absolute.

saltedcoffii: Yes, and it works. Test it yourself and see if you really like.

uberhacker: In this case you got lucky and it works because `com.github.tchx84.Flatseal` ends up in the bin directory but I'm sure that is not always the case for most packages.

packages/flatseal.rb
uberhacker, line 15: I wasn't able to get this working without sommelier so I think we need to add that as a dependency also.

saltedcoffii: Even with GDK_BACKEND=wayland? It worked for me.

packages/flatseal.rb
saltedcoffii: ...

🤦🏻

packages/flatseal.rb
saltedcoffii: It's a gtk package so theroretically any gtk package should work

uberhacker: I don't think most users will know how to set that up so could you please provide instructions in `def self.postinstall`?

uberhacker: It might even be better to create a wrapper script to set `GDK_BACKEND=wayland`.

saltedcoffii: Just kidding, I installed it on a new install without sommelier running and it didn't work.

PR #5611, opened 2021-04-13
packages/nano.rb: add filecmd to dependencies

saltedcoffii:
Fixes
```
nano: error while loading shared libraries: libmagic.so.1: cannot open shared object file: no such file or directory
```

### DIFF
--- a/packages/nano.rb
+++ b/packages/nano.rb
@@ -24,6 +24,7 @@ class Nano < Package
   })
 
   depends_on 'xdg_base'
+  depends_on 'filecmd'
 
   def self.patch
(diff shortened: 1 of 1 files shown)


packages/nano.rb
```diff
@@ -24,6 +24,7 @@ class Nano < Package
   })
 
   depends_on 'xdg_base'
+  depends_on 'filecmd'
```
uberhacker, line 27: This is already a core package so not sure why it was removed.

saltedcoffii: Weird, it wasn't installed on my machine after a fresh install. Oh I know why, the install script is broken right now.

PR #5618, opened 2021-04-14
Install gcc10 dependencies in buildessential

uberhacker:
Because gcc10 is an early package in install.sh and dependencies are not resolved at that point, we need to include them in buildessential to get the gcc compiler working.


saltedcoffii:
Do the same for ruby as gem requires libyaml.

PR #5619, opened 2021-04-14
Replace old crew symlink when `crew update`

### LINKED ISSUE #5603: crew symlink broken
image

supechicken:
Fixes #5603


uberhacker:
@supechicken: This won't run after `crew update` if buildessential is already installed.

supechicken:
> @supechicken: This won't run after `crew update` if buildessential is already installed.

Moved to `crew_profile_base.rb`

packages/buildessential.rb
```diff
@@ -2,7 +2,6 @@
 
 class Buildessential < Package
   description 'A collection of tools essential to compile and build software.'
-  homepage ''
```
uberhacker, line 5: Why was this line removed?

packages/crew_profile_base.rb
```diff
@@ -8,6 +8,9 @@ class Crew_profile_base < Package
   compatibility 'all'
   source_url 'file:///dev/null'
   source_sha256 'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855'
+  
+  # temporary fix for crew symlink
+  FileUtils.ln_sf "../lib/crew/bin/crew", "#{CREW_PREFIX}/bin/crew" if ARGV[1].eql?('update')
```
uberhacker, line 13: This won't execute either if crew_profile_base is already installed.  Take a look at line 457 in crew.

packages/buildessential.rb
satmandu: I might have done that just now while deleting an empty space and failing. Let me fix that.

satmandu:
Do we need to have crew_profile_base version revved to get crew to notice there is an update?

(@supechicken I think this also requires a new tag for https://github.com/chromebrew/crew-profile-base too)

packages/crew_profile_base.rb
satmandu: so uprev crew_profile_base?

supechicken: I see it now, crew will read version from device.json instead of package files. Bump the version of crew profile base can make a new symlink at upgrade but not update

uberhacker:
Let's let @saltedcoffii fix this since she got us into this mess in the first place. :-P

PR #5629, opened 2021-04-15
[New Package] libssp

saltedcoffii:
I'll add more packages as I get them to work, and then remove conflicts from the main package. Hopefully this can solve a lot of our problems with gcc as a core package.

### DIFF
--- a/packages/libssp.rb
+++ b/packages/libssp.rb
@@ -1,0 +1,46 @@
+require 'package'
+
+class Libssp < Package
+  description '**FIXME**'
+  homepage 'https://gcc.gnu.org/'
+  version '10.3.0'
+  license 'GPL-3, LGPL-3, libgcc, FDL-1.2'
+  compatibility 'all'
+  source_url 'https://ftpmirror.gnu.org/gcc/gcc-10.3.0/gcc-10.3.0.tar.xz'
+  source_sha256 '64f404c1a650f27fc33da242e1f2df54952e3963a49e06e73f6940f3223ac344'
+  
+  # local depends
+#  depends_on '' #TBD (hopefully nothing, as ruby needs it)
+  
+  @gcc_name = 'libssp'
+  @gcc_global_opts = '--enable-static \
+                      --enable-shared \
+                      --disable-multilib \
+                      --disable-cet \
+                      --with-pic \
+                      --enable-symvers \
+                      --with-gcc-major-version-only \
+                      --with-system-libunwind'
+  @gcc_global_env = "#{CREW_ENV_OPTIONS}"
+
+  def self.build
+    Dir.mkdir "#{@gcc_name}-builddir"
+    Dir.chdir "#{@gcc_name}-builddir" do
+      system "#{@gcc_global_env} ../#{@gcc_name}/configure #{CREW_OPTIONS} \
+              #{@gcc_global_opts}"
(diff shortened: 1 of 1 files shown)


saltedcoffii:
Actually, @satmandu let me know if the most recent version of gcc has libssp in it. If not, this package won't conflict with libssp and we can add it.

packages/libssp.rb
```diff
@@ -0,0 +1,46 @@
...
+  @gcc_name = 'libssp'
+  @gcc_global_opts = '--enable-static \
+                      --enable-shared \
+                      --disable-multilib \
+                      --disable-cet \
+                      --with-pic \
+                      --enable-symvers \
+                      --with-gcc-major-version-only \
+                      --with-system-libunwind'
+  @gcc_global_env = "#{CREW_ENV_OPTIONS}"
```
supechicken, line 24: This can be simplify as `@gcc_global_env = CREW_ENV_OPTIONS`

satmandu:
I have been building GCC 10.3.0 w/o libssp..

saltedcoffii:
Cool! So libssp can be its own package

packages/libssp.rb
```diff
@@ -0,0 +1,43 @@
+require 'package'
+
+class Libssp < Package
+  description 'Libssp is a part of the GCC toolkit.'
+  homepage 'https://gcc.gnu.org/'
+  version '10.3.0'
+  license 'GPL-3, LGPL-3, libgcc, FDL-1.2'
```
uberhacker, line 7: How can so many licenses exist for this?  It's confusing.

packages/libssp.rb
```diff
@@ -0,0 +1,43 @@
...
+                      --disable-cet \
+                      --with-pic \
+                      --enable-symvers \
+                      --with-gcc-major-version-only \
+                      --with-system-libunwind'
+
+  def self.build
+    Dir.mkdir "#{@gcc_name}-builddir"
+    Dir.chdir "#{@gcc_name}-builddir" do
+      system "#{CREW_ENV_OPTIONS} ../#{@gcc_name}/configure #{CREW_OPTIONS} \
```
uberhacker, line 26: `system "env #{CREW_ENV_OPTIONS}`?

satmandu:
> Cool! So libssp can be its own package

Yes! But if we can get packages to compile without it, we may not have much use of it.

At some point the granularity gets to the point where we have a package per file. Where do we stop?

packages/libssp.rb
saltedcoffii: These are all the licenses for GCC, I don't know libssp's licenses specifically

packages/libssp.rb
saltedcoffii: Makes no difference. Do you really want it changed?

saltedcoffii:
Well, libssp does provide some safety stuff for ruby. Apparently the reason it links to libssp in the first place was issues with segfaulting on some platforms.

satmandu:
I still want a way to do this as a single gcc compile and then split the compiled output into multiple packages... 

Having said that, I'm letting this compile overnight for all 3 architectures...

satmandu:
Added binaries

saltedcoffii:
Okay, I think this is ready to merge

PR #5635, opened 2021-04-15
[New Package] ruby_docopt

saltedcoffii:
Step 2 of making chromebrew a gem

This needs binaries before its merged as it will become an early package.

### DIFF
--- a/packages/ruby_docopt.rb
+++ b/packages/ruby_docopt.rb
@@ -1,0 +1,27 @@
+require 'package'
+
+class Ruby_docopt < Package
+  description 'Docopt parses command line arguments from nothing more than a usage message.'
+  homepage 'http://docopt.org/'
+  version '0.6.1'
+  license 'MIT'
+  compatibility 'all'
+  source_url 'https://github.com/docopt/docopt.rb.git'
+  git_hashtag '504f310c5fd66e9f3c36b078156e5f44573ff1d3'
+
+  depends_on 'ruby'
+
+  def self.patch
+    # https://github.com/docopt/docopt.rb/pull/37
+    system "sed -i 's:0.6.0:0.6.1:' lib/docopt.rb"
+    system "sed -i 's:0.6.0:0.6.1:' docopt.gemspec"
+  end
+
+  def self.build
+    system 'rake build'
+  end
+
+  def self.install
+    system "gem install --build=#{CREW_DEST_DIR} pkg/docopt-0.6.1.gem"
+  end
+end


packages/ruby_docopt.rb
```diff
@@ -0,0 +1,27 @@
...
+    system "sed -i 's:0.6.0:0.6.1:' lib/docopt.rb"
+    system "sed -i 's:0.6.0:0.6.1:' docopt.gemspec"
+  end
+
+  def self.build
+    system 'rake build'
+  end
+
+  def self.install
+    system "gem install --build=#{CREW_DEST_DIR} pkg/docopt-0.6.1.gem"
```
uberhacker, line 25: Don't we want to handle `CREW_PREFIX` also?

uberhacker:
Don't we already use docopt in crew?
```
require_relative '../lib/docopt'
```

saltedcoffii:
Yes, I'm going to be separating it soon.

packages/ruby_docopt.rb
saltedcoffii: This does it just fine.

uberhacker: Does what just fine?  What if you want a different `CREW_PREFIX` than `/usr/local`?

saltedcoffii: Then you install ruby in a different prefix and gem follows suit when installing.

uberhacker: I think you want `--build-root` instead of `--build`?

saltedcoffii: They do the same thing

uberhacker: Well, if they are the same, it isn't documented in `gem help install`.

saltedcoffii: Let's just leave it, I used the same logic you supported so much in `ruby_webrick`

uberhacker: Well, you have a point there.  Not sure where you are learning these undocumented things.

uberhacker: One thing this package doesn't have that ruby_webrick does (which goes a long way to solidifying a solid package) is a check section.

saltedcoffii: Docopt only has one test and the test fails because of bad syntax.

uberhacker: Well...that is not good.  What is the syntax error?

saltedcoffii:
```
$ rake test
/usr/local/bin/ruby -w -I"lib:lib:test" /usr/local/lib64/ruby/gems/3.0.0/gems/rake-13.0.3/lib/rake/rake_test_loader.rb "test/test_docopt.rb" 
Loaded suite /usr/local/lib64/ruby/gems/3.0.0/gems/rake-13.0.3/lib/rake/rake_test_loader
Started

  File "/usr/local/build/docopt/docopt.rb/test/language_agnostic_tester.py", line 34
    print 'result>', result
          ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print('result>', result)?
F
========================================================================================================================================================================
      7: 
      8:   def test_docopt_reference_testcases
      9:     puts
  => 10:     assert system('python', "test/language_agnostic_tester.py", "test/testee.rb", chdir: TOPDIR)
     11:   end
     12: end
/usr/local/build/docopt/docopt.rb/test/test_docopt.rb:10:in `test_docopt_reference_testcases'
Failure: test_docopt_reference_testcases(DocoptTest): <false> is not true.
========================================================================================================================================================================

Finished in 0.11918361 seconds.
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
1 tests, 1 assertions, 1 failures, 0 errors, 0 pendings, 0 omissions, 0 notifications
0% passed
------------------------------------------------------------------------------------------------------------------------------------------------------------------------
8.39 tests/s, 8.39 assertions/s
rake aborted!
Command failed with status (1): [ruby -w -I"lib:lib:test" /usr/local/lib64/ruby/gems/3.0.0/gems/rake-13.0.3/lib/rake/rake_test_loader.rb "test/test_docopt.rb" ]

Tasks: TOP => test
```

PR #5639, opened 2021-04-16
Restore aws package

uberhacker:
- Update aws2 package description, homepage, license and source_url
- Tested on all architectures

### DIFF
--- a/packages/aws.rb
+++ b/packages/aws.rb
@@ -1,0 +1,45 @@
+require 'package'
+
+class Aws < Package
+  description 'Universal Command Line Interface for Amazon Web Services'
+  homepage 'https://github.com/aws/aws-cli'
+  version '1.19.53'
+  license 'Apache-2.0'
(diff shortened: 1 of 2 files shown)


satmandu:
Should we be using `--single-version-externally-managed` on these to avoid screwing up python's internal package database?

That's what debian uses. See https://salsa.debian.org/python-team/packages/pygments/-/blob/debian/master/debian/rules

saltedcoffii:
After I separate the repository, I'll add back and clean up all the python packages. I don't think we should be using pip to install packages anymore.

satmandu:
> After I separate the repository, I'll add back and clean up all the python packages. I don't think we should be using pip to install packages anymore.

As long as we can still use pip... We should never be entirely dependent upon packagers to get an update of we need one.

saltedcoffii:
We can, we just shouldn't. It's not a breaking issue so this PR is fine to stay merged.

PR #5641, opened 2021-04-16
Separate chromebrew repository

saltedcoffii:
The code in this PR separates chromebrew from its package repository, now located at https://github.com/chromebrew/packages. Tested on an existing install and a new install.


saltedcoffii:
To test a new installation, run:
`env OWNER=saltedcoffii REPO=chromebrew BRANCH=repo PACKAGES_OWNER=chromebrew PACKAGES_REPO=packages PACKAGES_BRANCH=testing ./install.sh`

The install.sh must be the install.sh from https://github.com/saltedcoffii/chromebrew/ as it has been modified as well.

uberhacker:
One simple question: why?

saltedcoffii:
When chromebrew is a gem, keeping packages in ruby's normal gem storage directories would be really weird and lead to a massively long gemspec file that would need to be updated with every new package. Plus this way we can, in the far future, create things like portage overlays and maybe even switch from a rolling release model to a regular release model like most linux operating systems.

saltedcoffii:
It now works on my actual chromebook as a new install (before I only tested it in a chroot)

Let me get it to work on an existing install on an actual chromebook too.

saltedcoffii:
@uberhacker I'd like your input on something: Right now I have the package directory at `/usr/local/var/crew/repo/main`. portage would put it at `/usr/local/var/db/repo/main`. Do you have a preference?

uberhacker:
I'm not sold on this idea to move everything around.  What are your thoughts @satmandu?

saltedcoffii:
Alright, this works on a chromebook on a new install now. To verify this (and maybe find a bug or two!) run:
```
$ curl -#LO https://raw.githubusercontent.com/saltedcoffii/chromebrew/repo/install.sh && chmod +x install.sh
$ env OWNER=saltedcoffii REPO=chromebrew BRANCH=repo PACKAGES_OWNER=chromebrew PACKAGES_REPO=packages PACKAGES_BRANCH=testing ./install
```

Should you need to `crew update` again, just run:
```
$ env CREW_TESTING_REPO="https://github.com/saltedcoffii/chromebrew.git" \
CREW_TESTING_BRANCH=repo \
CREW_TESTING=1 \
crew update
```

Btw, big thanks to @satmandu for making testing repositories easier, it's so helpful.

saltedcoffii:
When I originally explained my reasoning for this PR, I was making dinner and was rushed. Here's my full reasoning.
Major benefits:
- It's very rare for a package manager to be married to its repository. Chromebrew is one of the few. Should someone want to make another package repository while still using chromebrew core, they will have to modify the package manager. This is unlike if someone wants to use APT without using debian's repositories, it is completely possible (e.g. ubuntu)
- When I make chromebrew a gem, having all the packages in the main (this) repository will make making chromebrew a gem impossible. Chromebrew isn't a gem in this PR because incremental changes.
- The packages directory has 1571 packages. That's nearly unmanageable, as git and GitHub truncate output to the first 1000 entries. This PR fixes this problem by splitting up the packages into substituent directories, making package files easier to browse on GitHub as well as in a file manager.

Minor benefits: 
- This PR splits the packages into two branches, canary and testing. This will prevent broken packages from being distributed directly to users. (It'll also keep me out of trouble from @uberhacker from making broken package files so much)

Potential future benefits:
- We could move from a rolling release "distro" to an incremental release "distro" by tagging releases of the packages directory and syncing them with a tarball.
- We could open an API for third parties to make packages and distribute them in a way decentralized from chromebrew, much like portage overlays. Chromebrew overlays would also be a great way to test new/updated packages before merging them.
- We could create multiple repositories like many distros have, e.g., ubuntu's core, universe, restricted and multiverse, or arch linux's core, extra, community and multilib.
- We could restrict packages that don't have binaries to canary, meaning that end users will never compile anything locally.
- This will make it easier to make a fix where broken package files don't break Chromebrew entirely.

Detriments (cons):
- This PR will likely take a bit to merge, making package pull requests to temporarily have to stop/be moved to the new package repository. This is short-term and temporary.
- This will take effort, primarily on my part.

Potential detriments:
- The two branches (canary and testing) could become out of sync. It may benefit us to get a bot of some sort to notify us when a package has been in canary for a week so we can manually move it to testing. To be honest, it will soon be time for us to get some automation of handling things (all other distros have this automation). I will make sure packages are moved around in a timely manner.

Because this is essential to chromebrew becoming a gem, if this does not get merged in some way, I won't be able to make chromebrew a gem.

Hopefully that outlines my thought process better.

saltedcoffii:
This now works on an existing install. just run
```
env CREW_TESTING_REPO="https://github.com/saltedcoffii/chromebrew.git" CREW_TESTING_BRANCH=repo CREW_TESTING=1 crew update
```
twice. (It fails the first time, I don't know why.)

bin/crew
```diff
@@ -228,18 +228,33 @@ def generate_compatible
 end
 
 def search (pkgName, silent = false)
-  pkgPath = CREW_PACKAGES_PATH + pkgName + '.rb'
+
+  if pkgName.start_with? 'ruby_'
```
supechicken, line 232: ​```ruby
case​ ​pkgName​
​when​ ​/ruby/​
...
​else​
...
​end
```

saltedcoffii: Okay that you! I couldn't figure that out earlier

satmandu:
Having Chromebrew not be tied to a repo is a good thing... But I also like flat folders, since I like to grep for things in packages...

saltedcoffii:
Oh! Well on the new install, just run `find (wherever the package repo is) -type f -exec grep -H (what you're grepping for) {} \;

saltedcoffii:
So what's the status on this PR?

saltedcoffii:
Because this PR has been ignored, seemingly intentionally, despite other activity within the repository, I'll assume that there isn't enough interest in pushing this pull request through for the time being. When that changes, ping me and I'll update everything here and in chromebrew/packages.

satmandu:
Sorry I've been occupied with the GCC/Perl/llvm/vim updates, which are smaller scale.

It's hard to work through this until those are taken care of... You're not being ignored, the rest of us have just presumably been busy!

Thanks for all the work you have put into this.

PR #5642, opened 2021-04-16
nolibssp vim rebuild

satmandu:
- rebuild w/o libssp
- needs perl rebuild too. :/
- I'm going to stop updating `xxd_standalone`, since there are no changes to it...

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686


saltedcoffii:
Why not use the libssp at #5629?

satmandu:
@saltedcoffii The idea is to reduce core dependencies...

saltedcoffii:
This package is really tiny and builds almost instantaneously. I'm guessing there's a good reason that vim and ruby link against libssp, the library provides signal processing stuff.

satmandu:
https://wiki.osdev.org/Stack_Smashing_Protector

https://reviews.freebsd.org/D15687

saltedcoffii:
"It is worth considering enabling it by default if your operating system is security conscious and you provide support."

This is now a question of whether we care about security. I would err on the side of security, as ChromeOS definitely values security. Thanks for those links by the way, I couldn't find any documentation on it.

uberhacker:
We have needed to relax security a bit with `xdg_base` and `sed` to get around the tyrannical and heavy-handed ChromeOS security measures.

saltedcoffii:
Yes, their security measures are quite harsh. Since libssp does no harm and is such a tiny package, I think we should use it.

satmandu:
> Yes, their security measures are quite harsh. Since libssp does no harm and is such a tiny package, I think we should use it.

My initial skepticism was due to this pulling in all of GCC as a runtime dependency. But splitting off libssp into its own package which could be added to core does seem to make a lot of sense.

(LLVM depending upon GCC was/is especially annoying.)

satmandu:
`vim` & `vim_runtime` here no longer need `perl` installed to work. `gvim` is built against the updated `perl` 5.32.1 in https://github.com/skycocker/chromebrew/pull/5665

satmandu:
Merging this before https://github.com/skycocker/chromebrew/pull/5665 will break `gvim`, but then merging https://github.com/skycocker/chromebrew/pull/5665 before updating all perl packages will... you know.

satmandu:
This now works with the current version of perl, and the new version, so this should be good to go.

packages/xxd_standalone.rb
```diff
@@ -3,24 +3,24 @@
 class Xxd_standalone < Package
   description 'Hexdump utility from vim'
   homepage 'http://www.vim.org'
-  @_ver = '8.2.2725'
+  @_ver = '8.2.2771'
```
uberhacker, line 6: Can this be updated to 8.2.2783?

satmandu: Yes, but the source file hasn't changed. It compiled to literally the same binary.

PR #5655, opened 2021-04-18
Add spellcheck_test script

uberhacker:
How I figured out the spelling errors from the last few PRs.

### DIFF
--- a/tests/spellcheck_test
+++ b/tests/spellcheck_test
@@ -1,0 +1,15 @@
+#!/usr/bin/env ruby
+
+# This test checks for spelling errors and provides suggested changes.
+
+require_relative '../lib/const'
+require_relative '../lib/color'
+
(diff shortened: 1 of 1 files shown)


supechicken:
Better to write this command to documents like `useful commands`?

uberhacker:
Not sure what you mean.

supechicken:
Since this is a wrapper script for `codespell`, it could be make sense to run it directly (`cd #{CREW_PREFIX}/lib/crew/packages; codespell`) in shell

uberhacker:
The purpose of the wrapper script is to run from the correct location.  Otherwise, the script does not catch everything.  Also, so we can consolidate testing scripts in one area.

saltedcoffii:
Putting all the testing scripts into tests/*.rb will make it easier to run `rake test` once we have a Rakefile

PR #5656, opened 2021-04-18
Static checksum for some packages

supechicken:
Tested on `x86_64`


uberhacker:
Good call.  The calculated checksums were running everywhere whether installed or not.

PR #5657, opened 2021-04-18
Fix wayland

supechicken:
Tested on `x86_64`

### DIFF
--- a/packages/wayland.rb
+++ b/packages/wayland.rb
@@ -29,10 +29,11 @@ class Wayland < Package
       set +a
     EOF
     
-    system "meson #{CREW_MESON_FNO_LTO_OPTIONS} \
-
-    -Ddocumentation=false \
-    builddir"
+    system <<~CONFIGURE
+      meson #{CREW_MESON_FNO_LTO_OPTIONS} \
+      -Ddocumentation=false \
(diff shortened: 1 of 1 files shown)


packages/wayland.rb
```diff
@@ -29,10 +29,11 @@ def self.build
...
     
-    system "meson #{CREW_MESON_FNO_LTO_OPTIONS} \
-
-    -Ddocumentation=false \
-    builddir"
+    system <<~CONFIGURE
+      meson #{CREW_MESON_FNO_LTO_OPTIONS} \
+      -Ddocumentation=false \
+      builddir
+    CONFIGURE
```
uberhacker, line 36: What exactly is fixed here?  This CONFIGURE string is not referenced anywhere else in the package.

supechicken: I removed the empty line in `line 33`

satmandu: Why use `<<~CONFIGURE` instead of just using double quotes?

uberhacker: I guess one advantage would be you don't have to escape double quotes, if needed.

uberhacker:
Fixed in #5660.

PR #5658, opened 2021-04-18
crew: Hide all non-fatal warnings during build

supechicken:
The warnings are useful for package developers but not helpful for debugging (like `warning: variable ‘abc’ set but not used`).

Tested on `x86_64`


lib/const.rb
```diff
@@ -75,20 +62,63 @@
...
-CREW_OPTIONS = "--prefix=#{CREW_PREFIX} --libdir=#{CREW_LIB_PREFIX} --mandir=#{CREW_MAN_PREFIX} --build=#{CREW_BUILD} --host=#{CREW_TGT} --target=#{CREW_TGT} --program-prefix='' --program-suffix=''"
-CREW_MESON_OPTIONS = "-Dprefix=#{CREW_PREFIX} -Dlibdir=#{CREW_LIB_PREFIX} -Dmandir=#{CREW_MAN_PREFIX} -Dbuildtype=minsize -Db_lto=true -Dcpp_args=#{CREW_COMMON_FLAGS} -Dc_args=#{CREW_COMMON_FLAGS}"
-CREW_MESON_FNO_LTO_OPTIONS = "-Dprefix=#{CREW_PREFIX} -Dlibdir=#{CREW_LIB_PREFIX} -Dmandir=#{CREW_MAN_PREFIX} -Dbuildtype=minsize -Db_lto=false -Dcpp_args=#{CREW_COMMON_FNO_LTO_FLAGS} -Dc_args=#{CREW_COMMON_FNO_LTO_FLAGS}"
+CREW_ENV_OPTIONS = <<~OPT
+  CFLAGS=#{CREW_COMMON_FLAGS} \
+  CXXFLAGS=#{CREW_COMMON_FLAGS} \
+  FCFLAGS=#{CREW_COMMON_FLAGS} \
+  FFLAGS=#{CREW_COMMON_FLAGS} \
+  LDFLAGS=#{CREW_LDFLAGS}
+OPT
```
uberhacker, line 76: Doesn't this add extra spaces which makes the actual output even longer on a single line?

bin/crew
```diff
@@ -769,21 +771,23 @@ def prepare_package (destdir)
...
-    conflicts.reject!(&:empty?)
-    unless conflicts.empty?
+    conflicts = {}
+    File.readlines('filelist').each do |p|
+      next unless File.exist?(p.chomp)
+      Dir['/usr/local/etc/crew/meta/*.filelist'].each do |f|
+        pkgName = File.basename(f, '.filelist')
+        c = File.foreach(f).grep( Regexp.new('^' + p + '$') )[0]
+        conflicts[c] = pkgName if c
+      end
```
uberhacker, line 781: Nice!  Does this add any extra time to the process?

bin/crew
```diff
@@ -769,21 +771,23 @@ def prepare_package (destdir)
...
-        puts "\n#{p} package conflicts below:".lightred if p != pp
-        puts f.lightred
+      pp = nil
+      conflicts.each_pair do |f, p|
+        puts nil, "#{p} package conflicts below:".lightred if p != pp
+        puts f.chomp.lightred
         pp = p
       end
-      abort unless ENV['CREW_CONFLICTS_ONLY_ADVISORY'] == '1'
+      abort unless ENV['CREW_CONFLICTS_ONLY_ADVISORY']
```
uberhacker, line 790: Does this abort with `CREW_CONFLICTS_ONLY_ADVISORY=false` or `CREW_CONFLICTS_ONLY_ADVISORY='0'`?

lib/const.rb
```diff
@@ -1,16 +1,19 @@
...
 
-CREW_VERSION = '1.8.9'
+CREW_VERSION = '1.8.10'
 
-ARCH_ACTUAL = `uname -m`.strip
+ARCH_ACTUAL = `uname -m`.chomp
 # This helps with virtualized builds on aarch64 machines
 # which report armv8l when linux32 is run.
-ARCH = if ARCH_ACTUAL == 'armv8l' then 'armv7l' else ARCH_ACTUAL end
+ARCH = ENV['CREW_ARCH'] || if ARCH_ACTUAL == 'armv8l' then 'armv7l' else ARCH_ACTUAL end
```
uberhacker, line 8: Why would we ever want the architecture in an environment variable?  Shouldn't that go with the territory?

lib/const.rb
```diff
@@ -1,16 +1,19 @@
...
+
+CREW_LIB_SUFFIX = if ARCH == 'x86_64' then '64' else '' end
```
uberhacker, line 10: This would also work:
```
CREW_LIB_SUFFIX = (ARCH == 'x86_64') ? '64' : ''
```

lib/const.rb
```diff
@@ -1,16 +1,19 @@
...
 
 ARCH_LIB = if ARCH == 'x86_64' then 'lib64' else 'lib' end
 LIBC_VERSION = if File.exist? "/#{ARCH_LIB}/libc-2.27.so" then '2.27' else '2.23' end
+CHROMEOS_RELEASE = File.read('/etc/lsb-release').scan(/CHROMEOS_RELEASE_CHROME_MILESTONE=(.*)/)[0][0]
```
uberhacker, line 14: Nice!

bin/crew
```diff
@@ -642,11 +642,13 @@ def download
       # Stow file in cache if requested
       if CREW_CACHE_ENABLED
         puts 'Caching downloaded archive...'
-          Dir.chdir "#{@extract_dir}" do
-            system "tar c#{@verbose}Jf #{CREW_CACHE_DIR}/#{filename}.tar.xz \
-              $(find -mindepth 1 -maxdepth 1 -printf '%P\n')"
-          end
-        system "sha256sum #{CREW_CACHE_DIR}/#{filename}.tar.xz > #{CREW_CACHE_DIR}/#{filename}.tar.xz.sha256"
+        Dir.chdir @extract_dir do
```
uberhacker, line 645: Not that it ever should happen but what if there are spaces in the directory path?

bin/crew
```diff
@@ -4,9 +4,6 @@ require_relative '../lib/color'
 # Disallow sudo
 abort 'Chromebrew should not be run as root.'.lightred if Process.uid == 0
 
-require 'find'
-require 'net/http'
-require 'uri'
```
uberhacker, line 9: Good call.  Less is more.

lib/const.rb
```diff
@@ -75,20 +62,63 @@
   CREW_BUILD = 'x86_64-cros-linux-gnu'
 end
 
-CREW_COMMON_FLAGS = "'-Os -pipe -flto=auto -fuse-ld=gold'"
-CREW_COMMON_FNO_LTO_FLAGS = "'-Os -pipe -fno-lto -fuse-ld=gold'"
+CREW_COMMON_FLAGS = "'-w -Os -pipe -flto=auto -fuse-ld=gold'"
```
saltedcoffii, line 65: Maybe instead of doing -w, doing `-Wno-error -Wno-dev` would be helpful. When I'm building packages I do use some of those package warnings to figure out why a package isn't building. -Wno-error prevents the compiler from fatally failing and -Wno-dev hides warnings meant for upstream developers.

lib/const.rb
saltedcoffii: Not with the ~ after OPT. That causes ruby to cut excess indents from the string.

lib/const.rb
saltedcoffii: I think what we have right now makes it more readable.

lib/const.rb
```diff
@@ -1,16 +1,19 @@
...
+ARCH_ACTUAL = `uname -m`.chomp
 # This helps with virtualized builds on aarch64 machines
 # which report armv8l when linux32 is run.
-ARCH = if ARCH_ACTUAL == 'armv8l' then 'armv7l' else ARCH_ACTUAL end
+ARCH = ENV['CREW_ARCH'] || if ARCH_ACTUAL == 'armv8l' then 'armv7l' else ARCH_ACTUAL end
+
+CREW_LIB_SUFFIX = if ARCH == 'x86_64' then '64' else '' end
 
 ARCH_LIB = if ARCH == 'x86_64' then 'lib64' else 'lib' end
 LIBC_VERSION = if File.exist? "/#{ARCH_LIB}/libc-2.27.so" then '2.27' else '2.23' end
```
saltedcoffii, line 13: Is there a way to read what the libc version is and use that as the variable?

lib/const.rb
```diff
@@ -75,20 +62,63 @@
...
+  -DCMAKE_LIBRARY_PATH=#{CREW_LIB_PREFIX} \
+  -DCMAKE_C_FLAGS=#{CREW_COMMON_FLAGS} \
+  -DCMAKE_CXX_FLAGS=#{CREW_COMMON_FLAGS} \
+  -DCMAKE_BUILD_TYPE=MinSizeRel \
+  --build=#{CREW_BUILD} \
+  --host=#{CREW_TGT} \
+  --target=#{CREW_TGT}
+OPT
+
+CREW_CMAKE_LIBSUFFIX_OPTIONS = "-DLIB_SUFFIX=#{CREW_LIB_SUFFIX}" + CREW_CMAKE_OPTIONS
```
saltedcoffii, line 124: Good call

lib/const.rb
```diff
@@ -29,39 +32,23 @@
...
 CREW_DEST_MAN_PREFIX = CREW_DEST_DIR + CREW_MAN_PREFIX
 
-if ENV['CREW_PREFIX'].to_s == ''
-  HOME = ENV['HOME']
-else
-  HOME = CREW_PREFIX + ENV['HOME']
-end
+HOME = ENV['HOME'] || CREW_PREFIX + ENV['HOME']
+USER = `whoami`.chomp
+CREW_CACHE_DIR = ENV['CREW_CACHE_DIR'] || HOME + '/.cache/crewcache'
```
saltedcoffii, line 37: The crew cache dir should be CREW_PREFIX + "/var/cache/crew"

saltedcoffii (changes requested):

lib/const.rb
saltedcoffii: To test out multiarching stuff, or to run 32 bit code on a 64 bit machine

bin/crew
saltedcoffii: I don't even think homebrew supports spaces in directory paths.

bin/crew
saltedcoffii: You can also take out `securerandom` as we don't need it anymore.

bin/crew
satmandu: Do NOT change this section unless it is faster.

There is a 3 or 4 order of magnitude change in speed from the last version of this code block, and you had better be **damn** sure that this is going to be faster.

Just looking at this code you are iterating a grep for every single file in the file list, which will read the entire list of filelists in CREW_META_PATH for every line in filelist.

On my system that is 19M of files being read for every line in a filelist.
llvm.filelist is ~ 4000 lines.

That's 76Gb of files having to be read from slow ssd just to do a conflict check.

`grep --exclude #{CREW_META_PATH}#{@pkg.name}.filelist \
      -Fxf filelist #{CREW_META_PATH}*.filelist | tr ':' ' ' | \
      sed 's,.filelist,,g' | sed 's,#{CREW_META_PATH},,g'` reads that 19Mb from the filesystem once.

bin/crew
satmandu: Leave the `== 1` check in. It's much simpler, and leaves it in env for when you want to grep to see what the relevant env variables are. It's much easier to do a `export CREW_CONFLICTS_ONLY_ADVISORY=0` or whatever when you want to go back to regular mode, and then flip back as necessary without having to look up the relevant env variable again.

lib/const.rb
satmandu: CREW_ARCH looks useful. I'm literally modifying ARCH in install.sh with sed for all of my installs in containers because I'm running on a RPI4B with 8gb of ram on a 64 bit kernel...

lib/const.rb
satmandu: It would definitely be good to have `-Wno-dev` as a minimum part of CMAKE variables. I think ARCH already does this.

lib/const.rb
uberhacker: If you need to change the architecture then you are probably not operating in the correct environment.

bin/crew
uberhacker: Just because homebrew doesn't support it doesn't mean we need to follow suit.  We're not copycats.

bin/crew
supechicken: > You can also take out `securerandom` as we don't need it anymore.

Done

lib/const.rb
supechicken: There is no way to do it in ruby, but you can use a command to get it:
```shell
$ /lib64/libc.so.6 --version
GNU C Library (Gentoo 2.27-r20 p3) stable release version 2.27.
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.
Compiled by GNU CC version 4.9.x 20150123 (prerelease).
libc ABIs: UNIQUE IFUNC
For bug reporting instructions, please see:
<http://crbug.com/new>.
```

lib/const.rb

lib/const.rb

bin/crew
supechicken: Reverted change

bin/crew
supechicken: It will be faster than `grep` if there is no conflict in the package (because of `File.exist?` wrapper)

lib/const.rb
supechicken: @uberhacker Ruby will delete the spaces at start automatically when in `~` mode
@saltedcoffii That's fine as we have space already before the backsplash

lib/const.rb
saltedcoffii: We could probably use sed and tr to get it

Or we could just tr the string of the name of the library.

bin/crew
satmandu: Let's assume that one has llvm 11 installed, and one is building llvm 12.
Most of those files in llvm12 exist in llvm11, so let's say 4000 of the files in llvm 12 also exist in llvm 11.

https://github.com/skycocker/chromebrew/blob/0013840470a2d4b47438fc7a10ea9ea11ecaca54/bin/crew#L758

So File exists, and we continue onto the code block, where `File.foreach(f).grep( Regexp.new('^' + p + '$') )[0]` has to go through every file from `Dir['/usr/local/etc/crew/meta/*.filelist'].each do |f|`, and that has to happen x the number of files in the llvm 12 filelist which were also in the 11vm 11 filelist.

The total disk read for this algorithm will be... ~ 19Mb x 4000... so the same as before.

Maybe the best method here would be to break these files into little ruby scripts we could test directly and compare?

llvm would be the logical extreme.

`perf stat -r 5 -d <program>` should give us some small examples.

Can we build a test case for this?

satmandu: (I'm building a quick test case.)

satmandu: testcase attached. Also, your change to the algorithm does not seem to work right yet, as it is not showing the actual conflicts, and seems to error out after the first conflict found, which should during comparisons ignore the `@pkg.name.filelist`.
```
perf stat -r 5 -d ruby oldconflicts.rb 
Unable to complete this build since there is a conflict with the same file in another package.

gcc10 package conflicts below:
/usr/local/lib64/libgomp.so
Unable to complete this build since there is a conflict with the same file in another package.

gcc10 package conflicts below:
/usr/local/lib64/libgomp.so
Unable to complete this build since there is a conflict with the same file in another package.

gcc10 package conflicts below:
/usr/local/lib64/libgomp.so
Unable to complete this build since there is a conflict with the same file in another package.

gcc10 package conflicts below:
/usr/local/lib64/libgomp.so
Unable to complete this build since there is a conflict with the same file in another package.

gcc10 package conflicts below:
/usr/local/lib64/libgomp.so

 Performance counter stats for 'ruby oldconflicts.rb' (5 runs):

            339.06 msec task-clock                #    0.962 CPUs utilized            ( +-  1.97% )
                73      context-switches          #    0.214 K/sec                    ( +-  7.73% )
                 8      cpu-migrations            #    0.024 K/sec                    ( +- 28.96% )
             4,305      page-faults               #    0.013 M/sec                    ( +-  0.05% )
       996,739,450      cycles                    #    2.940 GHz                      ( +-  2.10% )  (50.15%)
     1,351,523,505      instructions              #    1.36  insn per cycle           ( +-  0.25% )  (62.60%)
       237,242,963      branches                  #  699.705 M/sec                    ( +-  0.43% )  (62.50%)
         2,608,311      branch-misses             #    1.10% of all branches          ( +-  2.12% )  (62.66%)
       511,345,096      L1-dcache-loads           # 1508.119 M/sec                    ( +-  0.39% )  (62.72%)
         5,076,733      L1-dcache-load-misses     #    0.99% of all L1-dcache hits    ( +-  2.28% )  (62.87%)
         1,387,491      LLC-loads                 #    4.092 M/sec                    ( +-  4.20% )  (50.36%)
           303,797      LLC-load-misses           #   21.90% of all LL-cache hits     ( +-  7.08% )  (50.24%)

            0.3524 +- 0.0133 seconds time elapsed  ( +-  3.76% )
```

```
perf stat -r 5 -d ruby newconflicts.rb 
Unable to complete this build since there is a conflict with the same file in another package.

llvm package conflicts below:
/usr/local/include/llvm/ADT/APFloat.h
Unable to complete this build since there is a conflict with the same file in another package.

llvm package conflicts below:
/usr/local/include/llvm/ADT/APFloat.h
Unable to complete this build since there is a conflict with the same file in another package.

llvm package conflicts below:
/usr/local/include/llvm/ADT/APFloat.h
Unable to complete this build since there is a conflict with the same file in another package.

llvm package conflicts below:
/usr/local/include/llvm/ADT/APFloat.h
Unable to complete this build since there is a conflict with the same file in another package.

llvm package conflicts below:
/usr/local/include/llvm/ADT/APFloat.h

 Performance counter stats for 'ruby newconflicts.rb' (5 runs):

            609.27 msec task-clock                #    0.996 CPUs utilized            ( +-  0.32% )
                26      context-switches          #    0.043 K/sec                    ( +- 40.06% )
                 3      cpu-migrations            #    0.005 K/sec                    ( +- 49.74% )
             2,912      page-faults               #    0.005 M/sec                    ( +-  1.97% )
     1,855,294,971      cycles                    #    3.045 GHz                      ( +-  0.34% )  (49.95%)
     2,844,404,546      instructions              #    1.53  insn per cycle           ( +-  0.47% )  (62.45%)
       522,401,396      branches                  #  857.422 M/sec                    ( +-  0.36% )  (62.52%)
         3,166,502      branch-misses             #    0.61% of all branches          ( +-  0.67% )  (62.59%)
     1,000,120,822      L1-dcache-loads           # 1641.508 M/sec                    ( +-  0.24% )  (62.62%)
         6,968,953      L1-dcache-load-misses     #    0.70% of all L1-dcache hits    ( +-  0.72% )  (62.59%)
         2,091,469      LLC-loads                 #    3.433 M/sec                    ( +-  1.13% )  (50.06%)
           348,255      LLC-load-misses           #   16.65% of all LL-cache hits     ( +-  3.49% )  (49.93%)

           0.61168 +- 0.00220 seconds time elapsed  ( +-  0.36% )
```
[testcase.zip](https://github.com/skycocker/chromebrew/files/6337597/testcase.zip)

satmandu: (The `filelist` in that test case is the file list from a llvm12 build.)

bin/crew
supechicken: It is fine since URL will convert spaces to `%20` string

bin/crew
supechicken: Separating the check logic into a script...

lib/const.rb
satmandu: This is still a good idea right here...

bin/crew
supechicken: @satmandu I tried to make the logic faster but it is still a bit (~0.2 secs) slower than `grep`, so I restored the original one

satmandu: I wish there was a better way to make it faster... but grep is Really Fast when you don't have to read that stuff over and over again. :)

lib/const.rb

PR #5660, opened 2021-04-18
fix wayland.rb

ThatGeekyWeeb:
This empty newline causes the an error that prevents compiling...

> This is not my screenshot
image


uberhacker:
Thank you for the fix @ThatGeekyWeeb!

PR #5661, opened 2021-04-18
Fix aria2

### LINKED ISSUE #5648: [aria2] broken, needs libssh2.so
## Description
After a fresh install of chromebrew, installed aria2.
Ran:
```
$ aria2c
aria2c: error while loading shared libraries: libssh2.so.1: cannot open shared object file: No such file or directory
```

Potential solutions:
1. Compile aria2 without libssh2
2. Put libssh2 in aria2's dependencies.

saltedcoffii:
Fixes #5648

### DIFF
--- a/packages/aria2.rb
+++ b/packages/aria2.rb
@@ -4,34 +4,21 @@ class Aria2 < Package
   description 'aria2 is a lightweight multi-protocol & multi-source, cross platform download utility operated in command-line. It supports HTTP/HTTPS, FTP, SFTP, BitTorrent and Metalink.'
   homepage 'https://aria2.github.io/'
   @_ver = '1.35.0'
-  version "#{@_ver}-2"
+  version "#{@_ver}-3"
   license 'GPL-2+-with-openssl-exception'
   compatibility 'all'
   source_url "https://github.com/aria2/aria2/releases/download/release-#{@_ver}/aria2-#{@_ver}.tar.xz"
   source_sha256 '1e2b7fd08d6af228856e51c07173cfcf987528f1ac97e04c5af4a47642617dfd'
 
-  binary_url({
-    aarch64: 'https://downloads.sourceforge.net/project/chromebrew/armv7l/aria2-1.35.0-2-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/aria2.rb
```diff
@@ -4,34 +4,21 @@ class Aria2 < Package
...
-     x86_64: 'https://downloads.sourceforge.net/project/chromebrew/x86_64/aria2-1.35.0-2-chromeos-x86_64.tar.xz'
-  })
-  binary_sha256({
-    aarch64: 'b8082b1a4f9b8941f2891564046415abf544f113a452650d139c64253f4378f1',
-     armv7l: 'b8082b1a4f9b8941f2891564046415abf544f113a452650d139c64253f4378f1',
-       i686: '0e3b881cd0505eed0edb1ab898e5a272a3d9d950f9e9c9fd5ebc4d1d9a8c66d9',
-     x86_64: '20cb509c94ac4776c39172fbf8b13c80088249ab6deffce4c4eb6d40081bbd71'
-  })
-
   depends_on 'libgcrypt'
```
uberhacker, line 13: Don't we also need `depends_on 'libssh2'`?

packages/aria2.rb
```diff
@@ -4,34 +4,21 @@ class Aria2 < Package
...
   def self.build
-    system "env CFLAGS='-fuse-ld=gold -flto' CXXFLAGS='-fuse-ld=gold -flto' \
-     LDFLAGS='-flto' \
+    system "env #{CREW_ENV_OPTIONS}
      ./configure #{CREW_OPTIONS} \
       --without-libnettle \
       --with-libgcrypt \
-      --disable-dependency-tracking"
+      --disable-dependency-tracking \
+      --without-libssh2"
```
uberhacker, line 21: What's wrong with including this?

packages/aria2.rb
saltedcoffii: No, because of `--without-libssh2`

packages/aria2.rb
saltedcoffii: Adds another dependency we don't need.

uberhacker: You mean like using ssh 1 instead of 2 which is less secure?

saltedcoffii: It just doesn't link against ssh

PR #5665, opened 2021-04-19
perl -> 5.32.1, gcc -> 11.1, llvm -> 12.0.0, add pixz to core

satmandu:
- Perl update to 5.32.1 (plus all packages)
- Gcc10 -> 10.3.0
- Add gcc11 (11.1.0) and set buildessential to use it.
- llvm -> 12.0.0
- Add pixz to core (needed for tpxz package installs.)

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l


uberhacker:
I get it.  Perl updates are a PITA.

packages/perl.rb
```diff
@@ -3,23 +3,23 @@
...
+  source_url 'http://www.cpan.org/src/5.0/perl-5.32.1.tar.gz'
+  source_sha256 '03b693901cd8ae807231b1787798cf1f2e0b8a56218d07b7da44f784a7caeb2c'
 
-  binary_url ({
-    aarch64: 'https://downloads.sourceforge.net/project/chromebrew/armv7l/perl-5.32.0-chromeos-armv7l.tar.xz',
-     armv7l: 'https://downloads.sourceforge.net/project/chromebrew/armv7l/perl-5.32.0-chromeos-armv7l.tar.xz',
-       i686: 'https://downloads.sourceforge.net/project/chromebrew/i686/perl-5.32.0-chromeos-i686.tar.xz',
-     x86_64: 'https://downloads.sourceforge.net/project/chromebrew/x86_64/perl-5.32.0-chromeos-x86_64.tar.xz',
+  binary_url({
+    aarch64: 'https://downloads.sourceforge.net/project/chromebrew/perl-5.32.1-chromeos-armv7l.tar.xz',
```
uberhacker, line 13: Shouldn't the aarch64 and armv7l downloads be identical?

satmandu: Fixed. Not sure what happened there. Looks like an errant upload.

uberhacker:
There are many more perl modules.  I assume this PR will include those packages also?

satmandu:
Yes that's the plan... but this is planned for after we deal with gcc & llvm...

packages/lzma.rb
```diff
@@ -3,31 +3,13 @@
 class Lzma < Package
   description 'LZMA Utils are legacy data compression software with high compression ratio.'
   homepage 'https://tukaani.org/lzma/'
-  version '4.32.7-1'
+  version '4.98'
   license 'public-domain'
   compatibility 'all'
-  source_url 'https://tukaani.org/lzma/lzma-4.32.7.tar.bz2'
-  source_sha256 '618e54513993b3a153fa1c150fccdf25788c72b36e84ab4db71911083531cf6a'
+  source_url 'SKIP'
```
uberhacker, line 9: Fake packages shouldn't need this line.

packages/xzutils.rb
```diff
@@ -3,28 +3,22 @@
...
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/xzutils/5.2.5_x86_64/xzutils-5.2.5-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: 'bbb38dfbc5f7845dde5d4e6c1aa46be40d4d94f5db26b0aa20f1603a14c6e61d',
-     armv7l: 'bbb38dfbc5f7845dde5d4e6c1aa46be40d4d94f5db26b0aa20f1603a14c6e61d',
-       i686: '34db46c8dc58db4c212ba83e36eba17c8852eab14964a6a1d7ab75aaf23be91c',
-     x86_64: '05f0e86514ed8fefa240e0fc2d40cb2b95414d628f670b19a230a3867e992a24',
-  })
+  source_url 'https://github.com/xz-mirror/xz.git'
+  git_hashtag 'e7da44d5151e21f153925781ad29334ae0786101'
```
uberhacker, line 10: Binaries coming?

packages/perl_locale_gettext.rb
```diff
@@ -1,32 +1,28 @@
-
 require 'package'
 
 class Perl_locale_gettext < Package
   description 'Locale::gettext - message handling functions'
   homepage 'https://metacpan.org/pod/Locale::gettext'
-  compatibility 'all'
   license 'GPL-1+ or Artistic'
-  version '1.07-1'
+  version '1.07-2'
```
uberhacker, line 7: Can we put version above license for consistency?

packages/perl_sgmls.rb
```diff
@@ -1,36 +1,34 @@
-
 require 'package'
 
 class Perl_sgmls < Package
   description 'a set of Perl5 routines for processing the output from the onsgmls SGML parsers.'
   homepage 'http://search.cpan.org/dist/SGMLSpm/'
-  compatibility 'all'
   license 'GPL-1+ or Artistic'
-  version '1.1-1'
+  version '1.1-2'
```
uberhacker, line 7: Can we put version above license?

packages/stow.rb
```diff
@@ -9,6 +9,12 @@ class Stow < Package
   source_url 'https://ftpmirror.gnu.org/stow/stow-2.3.1.tar.gz'
   source_sha256 '09d5d99671b78537fd9b2c0b39a5e9761a7a0e979f6fdb7eabfa58ee45f03d4b'
 
+  depends_on 'perl'
```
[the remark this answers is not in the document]
satmandu: This should really just be named perl_stow. Maybe I should just make that change, since this is a perl module?

I added the dependency so that people would notice when rebuilding all perl packages.

packages/perl_locale_gettext.rb
satmandu: That's my default for new packages. This was just a bunch of straight rebuilds with newer perl... Feel free to change that in any packages here you notice that on, as that will also save me work.

satmandu: (by default I meant that all I was changing was the version number as I was updating these packages...)

uberhacker: I meant swap the lines to be consistent with all the other packages.

satmandu: No I got that... it's on my list... there are just a lot of these files which could have that issue.

packages/perl_sgmls.rb
satmandu: There has to be an awk or sed construction which should make this easy, right?

satmandu: This does a mediocre job in checking all files.
`for i in perl*.rb ; do gawk -i inplace '/^\slicense/{t=$0;next}1;/^\sversion/{print t}' $i ; rubocop -A $i;  done`

packages/gcc10.rb
```diff
@@ -66,131 +100,292 @@ def self.build
...
+    # Remove all conflicting files from conflicting packages / reinstall
+    # This should be at the very end.
+    conflict_packages = %w[libssp]
+    conflict_packages.each do |package|
+       file = File.open("#{CREW_META_PATH}#{package}.filelist").read
+       file.each_line do |line|
+         FileUtils.rm "/#{line}" if File.exist?("/#{line}")
+       end
+      # Reinstall these conflicting packages since we may have overwritten them.
+      system "crew reinstall #{package}"
```
uberhacker, line 388: I'm thinking the symlinks should go in the install section instead so they can be tracked?  Seems like there could be a more elegant way to do this?  Why don't we just not install the conflicting files from gcc in the first place?

packages/libgcc_s1.rb
```diff
@@ -0,0 +1,14 @@
...
+  homepage 'https://www.gnu.org/software/gcc/'
+  version '11.1.0-1'
+  license 'GPL-3, LGPL-3, libgcc, FDL-1.2'
+  compatibility 'all'
+
+  depends_on 'gcc11'
+  
+  is_fake
+  
+end
```
uberhacker, line 14: One question: why?

packages/gcc11.rb
```diff
@@ -0,0 +1,391 @@
...
+    # This should be at the very end.
+    conflict_packages = %w[libssp]
+    conflict_packages.each do |package|
+       file = File.open("#{CREW_META_PATH}#{package}.filelist").read
+       file.each_line do |line|
+         FileUtils.rm "/#{line}" if File.exist?("/#{line}")
+       end
+      # Reinstall these conflicting packages since we may have overwritten them.
+      system "crew reinstall #{package}"
+    end
```
uberhacker, line 389: Same questions apply from the gcc10 package.

packages/libssp.rb
```diff
@@ -3,41 +3,92 @@
...
+  --with-pic \
+  --with-system-libunwind \
+  --with-system-zlib'
+
+  @cflags = '-fPIC -pipe'
+  @cxxflags = '-fPIC -pipe'
+  @languages = 'c,c++,jit,objc,fortran,go'
+  case ARCH
+  when 'armv7l', 'aarch64'
+    @archflags = '--with-arch=armv7-a --with-float=hard --with-fpu=neon --with-tune=cortex-a15'
```
uberhacker, line 63: Should we be assuming all arm systems use the cortex a15 chipset?

packages/llvm.rb
```diff
@@ -3,75 +3,94 @@
...
   source_url "https://github.com/llvm/llvm-project/archive/llvmorg-#{@_ver}.tar.gz"
-  source_sha256 '53a0719f3f4b0388013cfffd7b10c7d5682eece1929a9553c722348d1f866e79'
+  source_sha256 '8e6c99e482bb16a450165176c2d881804976a2d770e0445af4375e78a1fbf19c'
 
   binary_url({
-    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/llvm/11.1.0_armv7l/llvm-11.1.0-chromeos-armv7l.tar.xz',
-     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/llvm/11.1.0_armv7l/llvm-11.1.0-chromeos-armv7l.tar.xz',
-       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/llvm/11.1.0_i686/llvm-11.1.0-chromeos-i686.tar.xz',
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/llvm/11.1.0_x86_64/llvm-11.1.0-chromeos-x86_64.tar.xz'
+
```
uberhacker, line 14: Extra line?

packages/perl.rb
```diff
@@ -3,41 +3,58 @@
...
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/perl/5.32.1_x86_64/perl-5.32.1-chromeos-x86_64.tar.xz',
+     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/perl/5.32.1_x86_64/perl-5.32.1-chromeos-x86_64.tar.xz'
   })
-  binary_sha256 ({
+  binary_sha256({
     aarch64: '8f96c4ac4e4927ff643670ba362568e5483ba4ef970acd742877ef49b48829df',
      armv7l: '8f96c4ac4e4927ff643670ba362568e5483ba4ef970acd742877ef49b48829df',
        i686: '2193da9719f91be5aa261471aae0314b19da430ae3f09006c30d7d276b5f230d',
-     x86_64: '8c12ba41052f0f49f1dccc7cc1de795d4b4c55ba70c4ed1eff866580405e300a',
+     x86_64: '8c12ba41052f0f49f1dccc7cc1de795d4b4c55ba70c4ed1eff866580405e300a'
```
uberhacker, line 23: Maybe I'm just tired but shouldn't these binaries be new to this version?

packages/llvm.rb
```diff
@@ -3,75 +3,94 @@
...
     # This keeps system libraries from being linked in as dependencies on a build host image, such as in docker.
     # Uncomment the rest of this section if you need to.
-    # system "sudo rm /lib#{CREW_LIB_SUFFIX}/libncurses.so.5 || true"
-    # system "sudo rm /usr/lib#{CREW_LIB_SUFFIX}/libform.so || true"
-    # system "sudo ln -s #{CREW_LIB_PREFIX}/libncurses.so.6 /lib#{CREW_LIB_SUFFIX}/libncurses.so.5 || true"
-    # system "sudo ln -s #{CREW_LIB_PREFIX}/libform.so /usr/lib#{CREW_LIB_SUFFIX}/libform.so || true"
+    system "sudo rm /lib#{CREW_LIB_SUFFIX}/libncurses.so.5 || true"
+    system "sudo rm /usr/lib#{CREW_LIB_SUFFIX}/libform.so || true"
+    system "sudo ln -s #{CREW_LIB_PREFIX}/libncurses.so.6 /lib#{CREW_LIB_SUFFIX}/libncurses.so.5 || true"
+    system "sudo ln -s #{CREW_LIB_PREFIX}/libform.so /usr/lib#{CREW_LIB_SUFFIX}/libform.so || true"
```
uberhacker, line 87: Why do we need to change system files?  Can't `LD_LIBRARY_PATH` be used to override system shared libraries?

tools/core_packages.txt
```diff
@@ -46,6 +46,7 @@ p11kit
 pcre
 pcre2
 perl
+pixz
```
uberhacker, line 49: Should we be setting `CREW_USE_PIXZ = 1` in `lib/const.rb`?

packages/libssp.rb
satmandu: The a15 is from here: https://www.chromium.org/chromium-os/how-tos-and-troubleshooting/chromiumos-board-porting-guide
```
-mtune=<arch> -- see the GCC manual for complete lists of possible values here
This selects the processor that the code will be optimized best for without requiring it
If you don't specify this, then the -march setting will be used
cortex-a8 or cortex-a9 or cortex-a15 are common for ARM parts
```

(I think _optimized best for without requiring it_ keeps it safe.)

packages/gcc10.rb
satmandu: This particular code block is from when I had libgcc_s1 also pulled out to a different package. libssp should not conflict with gcc. 

In regards to the symlinks... they're there in postinstall as a method of avoiding conflicts with other gcc installs which could in theory be installed alongside this.

With the postinstall as it exists now, that is possible... at least for the executables, but not for the libraries (which are per-gcc -version and are all in the common lib folder.)

Why not also consider the alternative of appending files to the CREW_META filelist from post-install code blocks?

We need to create an array of files that are being created, and then append them to the filelist...

packages/llvm.rb
satmandu: This is only for building on the container. I'll comment this out. I think some packages are happy to ignore `LD_LIBRARY_PATH` and do their own thing. :/

tools/core_packages.txt
satmandu: I was thinking it might make sense to invert the logic and use PIXZ unless `CREW_USE_PIXZ=0`?

uberhacker: That would work too.

packages/gcc10.rb
uberhacker: You don't have to do all that if you put everything in install instead of postinstall. :)  The problem with not tracking the symlinks is they become dangling after package removal.

packages/libssp.rb
satmandu: I'll remove it from here. Better safe.

packages/gcc10.rb
```diff
@@ -66,131 +100,297 @@ def self.build
...
+    end
+    puts 'If you do not need gcc you can uninstall gcc to save space with'.lightgreen
+    puts '"crew update ; crew upgrade ; crew remove gcc10 gcc11"'.lightgreen
+    # Add postinstall symlinks to filelist
+    filelist = File.readlines("#{CREW_META_PATH}#{@name}.filelist")
+    add_filelist.each do |new_file|
+      filelist.append(new_file) unless filelist.include?(new_file)
+    end
+    File.open("#{CREW_META_PATH}#{@name}.filelist", 'w+') do |f|
+      f.puts(filelist)
```
uberhacker, line 393: I'm still not understanding why this filelist appending is necessary if the same logic could be in the install section instead of the postinstall section.

packages/perl.rb
satmandu: They SHOULD be. But it looks like you inadvertently edited in the wrong versions when adjusting for gitlab URLs?

https://github.com/skycocker/chromebrew/blob/f7c9a99f2d7b897ef85b6a17511aa5a6c89252f0/packages/perl.rb

packages/gcc10.rb
satmandu: rebuilding with symlinks in install. See you Wednesday. :)

satmandu:
@uberhacker Also see install.sh changes here to allow .tpxz package usage...

install.sh
```diff
@@ -127,7 +129,11 @@ function extract_install () {
 
     #extract and install
     echo "Extracting ${1} (this may take a while)..."
-    tar xpf ../"${2}"
+    if ! LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} pixz -h > /dev/null; then 
+      tar xpf ../"${2}"
+    else
+    LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} tar -Ipixz -xpf ../"${2}"
```
uberhacker, line 135: Not sure why we would ever need to set LD_LIBRARY_PATH.  Are there system libraries needed for pixz and tar?

install.sh
```diff
@@ -17,12 +17,12 @@ CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
...
 # EARLY_PACKAGES cannot depend on crew_profile_base for their core operations (completion scripts are fine)
-EARLY_PACKAGES="gcc10 brotli c_ares libcyrussasl libidn2 libmetalink libnghttp2 libpsl \
+EARLY_PACKAGES="libarchive openssl zstd xzutils lz4 bz2 libxml2 pixz libssp brotli c_ares libcyrussasl libidn2 libmetalink libnghttp2 libpsl \
 libtirpc libunistring openldap rtmpdump zstd ncurses ca_certificates libyaml ruby libffi \
-openssl nettle krb5 p11kit libtasn1 gnutls curl git icu4c"
+nettle krb5 p11kit libtasn1 gnutls curl git icu4c "
 
 LATE_PACKAGES="binutils crew_profile_base less most manpages filecmd mawk readline perl pcre pcre2 \
-python27 python3 py3_pip sed bz2 lz4 lzip unzip xzutils zip"
+python27 python3 py3_pip sed lzip unzip zip"
```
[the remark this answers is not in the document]
satmandu: pixz has those dependencies, and I've given up on trying to get pixz compiled with `-static`.

packages/gcc10.rb
satmandu: This was taken out.

uberhacker:
I still don't see `lib/const.rb` with the version bump.

install.sh
```diff
@@ -203,14 +203,24 @@ git remote add -f origin "https://github.com/${OWNER}/${REPO}.git"
...
 
 [[ "$devtools" == "y" || "$devtools" == "Y" ]] && yes | crew install buildessential
+# Install core packages
+echo -e "${BLUE}
+Installing core packages...
+${RESET}"
+for i in $(cat tools/core_packages.txt  )
+do
+( yes | crew upgrade $i || true )
+done
```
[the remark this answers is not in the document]
satmandu: It does attempt to install packages that are already installed. It's a little messy. Let me just take this out for now.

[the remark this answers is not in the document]
satmandu: I'm just going to leave tools in though, so we can run this locally later on without having to redownload the `core_packages.txt` file.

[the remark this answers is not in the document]
satmandu: backed out..

PR #5666, opened 2021-04-19
Update opera package from 75.0.3969.93 to 75.0.3969.171

uberhacker:
Change to static sha256 checksum to avoid downloading for other commands.  Tested on x86_64.

### DIFF
--- a/packages/opera.rb
+++ b/packages/opera.rb
@@ -20,10 +20,11 @@ class Opera < Package
     # llvm-strip doesn't works with opera
     ENV['CREW_NOT_STRIP'] = '1'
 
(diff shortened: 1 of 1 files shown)


packages/opera.rb
```diff
@@ -33,21 +34,25 @@ def self.install
 
   def self.postinstall
     puts
-    puts 'Set Opera as your default browser? [Y/n]: '
-    case STDIN.gets.chomp
-    when "\n", 'Y', 'y', 'yes'
+    print 'Set Opera as your default browser? [Y/n]: '
```
satmandu, line 37: That's a sensible change!

PR #5667, opened 2021-04-19
rebuild ruby w/o dependence on gcc10 libssp

satmandu:
- This works for me with `gcc10` 10.3.0 built w/o libssp.
- This removes the `gcc10` dependence for `ruby` at install time,

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/ruby.rb
+++ b/packages/ruby.rb
@@ -3,23 +3,23 @@ require 'package'
 class Ruby < Package
   description 'Ruby is a dynamic, open source programming language with a focus on simplicity and productivity.'
   homepage 'https://www.ruby-lang.org/en/'
-  version '3.0.1'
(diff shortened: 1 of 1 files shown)


saltedcoffii:
What about the libssp standalone package?

uberhacker:
Are you confident enough to remove gcc10 from core packages?

satmandu:
I don't think we should remove gcc10 until I get the new gcc10 uploaded, and figure out a way to get libssp installed at the same time as the gcc10 update.

(How do we replace 1 package with two packages as part of an upgrade?)

The order needs to be:
1. Install new ruby
2. Remove old gcc10
3. Install new gcc10
4. Install libssp

supechicken (approved):

PR #5671, opened 2021-04-20
GCC 10.3.0 w/o libssp

satmandu:
- built w/o libssp, which is now a separate package (now merged)

- ruby built w/o libssp is already merged.
- openssh 8.6 is built w/o libssp is merged
- [perl 5.32.1](https://github.com/skycocker/chromebrew/pull/5665) is built w/o libssp

- core crew should not be broken with this installed, but this obviously needs testing before merge...

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686


packages/gcc10.rb
```diff
@@ -69,128 +76,338 @@ def self.build
...
+          --enable-plugin \
+          --enable-shared \
+          --enable-static \
+          --enable-threads=posix \
+          --program-suffix=-#{gcc_version} \
+          --with-arch-32=#{ARCH} \
+          --with-build-config=bootstrap-lto-lean \
+          --with-isl \
+          --with-pic \
+          --with-system-zlib"
```
uberhacker, line 173: I'm wondering instead of including all the same configure options for each architecture if it's possible to have one set of options for every architecture with variables?

packages/gcc10.rb
```diff
@@ -69,128 +76,338 @@ def self.build
...
+      # many packages expect this symlink
+      puts "Symlinking gcc-#{gcc_version} to cc"
+      FileUtils.ln_sf "gcc-#{gcc_version}", 'cc'
+    end
+    # make sure current version of gcc LTO plugin for Gold linker is installed.
+    FileUtils.mkdir_p "#{CREW_LIB_PREFIX}/bfd-plugins/"
+    puts "Symlinking #{CREW_PREFIX}/libexec/#{gcc_dir}/liblto_plugin.so to #{CREW_LIB_PREFIX}/bfd-plugins/"
+    FileUtils.ln_sf "#{CREW_PREFIX}/libexec/#{gcc_dir}/liblto_plugin.so", "#{CREW_LIB_PREFIX}/bfd-plugins/"
+    # binutils makes a symlink here, but just in case it isn't there.
+    unless CREW_LIB_PREFIX.to_s == "#{CREW_PREFIX}/lib"
```
uberhacker, line 407: Would `if ARCH == 'x86_64'` be easier to understand?

packages/gcc10.rb
satmandu: Just like llvm, sure...

That's a lot easier to do post hoc once we know that compiles work...

packages/gcc10.rb
satmandu: For sure... I just wanted a bit about the reasoning, but probably simpler just to have that in there.

packages/gcc10.rb
```diff
@@ -69,128 +76,338 @@ def self.build
...
-        --disable-werror \
-        --disable-libmpx \
-        --enable-static \
-        --enable-shared \
-        --with-pic \
-        --program-suffix=-#{gcc_version} \
-        --with-arch-64=x86-64 \
-        --with-build-config=bootstrap-lto-lean"
+        system "env CFLAGS=#{@cflags} CXXFLAGS=#{@cxxflags}  \
+          LD_LIBRARY_PATH=#{CREW_LIB_PREFIX} \
```
supechicken, line 118: This line can be removed since it is exported already

packages/gcc10.rb
```diff
@@ -69,128 +76,338 @@ def self.build
...
+          --with-arch-32=#{ARCH} \
+          --with-build-config=bootstrap-lto-lean \
+          --with-isl \
+          --with-pic \
+          --with-system-zlib"
       end
-      system 'make'
+      system "env LD_LIBRARY_PATH=#{CREW_LIB_PREFIX} \
+          LIBRARY_PATH=#{CREW_LIB_PREFIX} PATH=#{@path} \
+          make -j4"
```
supechicken, line 177: Will `make -j#{CREW_NPROC}` works?

satmandu: `-j#{CREW_NPROC}` is actually built-in! But for some reason I was only effectively getting `-j1` performance. Either a bug or an issue with how my container was seeing this....

saltedcoffii: try `crew const CREW_NPROC` maybe the nproc variable wasn't getting set correctly

packages/gcc10.rb
saltedcoffii: without LD_LIBRARY_PATH being exported, ruby won't even run so you don't have to worry about it.

packages/gcc10.rb
satmandu: `CREW_NPROC` does look like it is populated properly. However, it wasn't being pulled in during make for some reason. I plan on reverting this code to the default before it is moved from draft though.

packages/gcc10.rb
satmandu: Thanks all, That code is a kludge from when I wasn't sure if the environment variables were being passed correctly.

packages/gcc10.rb
satmandu: Collapsed all arch versions into one section.

packages/libssp.rb
```diff
@@ -10,10 +10,10 @@ class Libssp < Package
...
 
   binary_url({
-    aarch64: 'https://github.com/chromebrew/binaries/raw/main/armv7l/libssp-10.3.0-chromeos-armv7l.tar.xz',
-     armv7l: 'https://github.com/chromebrew/binaries/raw/main/armv7l/libssp-10.3.0-chromeos-armv7l.tar.xz',
-       i686: 'https://github.com/chromebrew/binaries/raw/main/i686/libssp-10.3.0-chromeos-i686.tar.xz',
-     x86_64: 'https://github.com/chromebrew/binaries/raw/main/x86_64/libssp-10.3.0-chromeos-x86_64.tar.xz'
+    aarch64: 'https://github.com/chromebrew/binaries/releases/download/v_gcc10_10.3.0/libssp-10.3.0-chromeos-armv7l.tar.xz',
+     armv7l: 'https://github.com/chromebrew/binaries/releases/download/v_gcc10_10.3.0/libssp-10.3.0-chromeos-armv7l.tar.xz',
+       i686: 'https://github.com/chromebrew/binaries/releases/download/v_gcc10_10.3.0/libssp-10.3.0-chromeos-i686.tar.xz',
+     x86_64: 'https://github.com/chromebrew/binaries/releases/download/v_gcc10_10.3.0/libssp-10.3.0-chromeos-x86_64.tar.xz'
```
uberhacker, line 16: I'm not sold on the releases idea.  This is going to make the repo huge over time.  As you know, space costs.

satmandu: Discussion on slack...

packages/gcc10.rb
```diff
@@ -1,63 +1,110 @@
...
+
+    # Install newer version of mpfr
+    mpfr_url = "https://www.mpfr.org/mpfr-current/mpfr-#{@mpfr_ver}.tar.xz"
+    mpfr_sha256 = '0c98a3f1732ff6ca4ea690552079da9c597872d30e96ec28414ee23c95558a7f'
+    system "curl -Ls #{mpfr_url} | hashpipe sha256 #{mpfr_sha256} | tar xJ"
+    Dir.chdir "mpfr-#{@mpfr_ver}" do
+      system 'curl -Ls "https://gforge.inria.fr/scm/viewvc.php/mpfr/misc/www/mpfr-4.1.0/allpatches?revision=14491&view=co" | \
+        hashpipe sha256 dfa7d8a14ec7cb3b344cb81cfd7bd7e22aba62379941cc9110759f11172ac013 | patch -NZp1 --binary'
+    end
+    system "ln -sf ../mpfr-#{@mpfr_ver} mpfr"
```
uberhacker, line 99: Why aren't we using packages `gmp`, `isl`, `mpc` and `mpfr` instead?

satmandu: I should comment this out. This was before I updated those packages.

satmandu:
Now that gcc 11.1.0 is out, I'm going to respin this for gcc 11.1, still breaking out the libssp, but doing that more cleanly.

satmandu:
This is obsoleted by https://github.com/skycocker/chromebrew/pull/5697

PR #5682, opened 2021-04-23
Update gcc prereqs: gmp, mpc, mpfr

satmandu:
- Updates gmp, mpc, mpfr
- gcc 10.3.0 has been built with these updates.
- We can't use `CREW_ENV_OPTIONS` because `lto` breaks these builds.

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l


packages/mpfr.rb
```diff
@@ -3,41 +3,50 @@
...
-    system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
     system "mkdir -p #{CREW_DEST_LIB_PREFIX}"
     system "ln -s #{CREW_LIB_PREFIX}/libmpfr.so.6 #{CREW_DEST_LIB_PREFIX}/libmpfr.so.4"
   end
 
   def self.check
-    system "make", "check"
+    # Check only works after an install on x86_64
+    system 'PWD=$(pwd) ; LD_LIBRARY_PATH=$pwd/src/.libs:$LD_LIBRARY_PATH make check'
```
uberhacker, line 50: No need to assign `pwd` to a variable.  You probably want:
```
system "LD_LIBRARY_PATH=\$(pwd)/src/.libs:\$LD_LIBRARY_PATH make check"
```

packages/mpfr.rb
```diff
@@ -3,41 +3,50 @@
...
+       i686: '2950b94cbcbcb11b0bb34a6e31d7e6e9982eb74d37edd73a17a3700c2b6abe41',
+     x86_64: 'a7e0a584a6d54b852d4869ef3ba8f9fdb076f48d6d3f2cb1f46842bc8975583a'
   })
 
+  depends_on 'autoconf_archive' => :build
+
+  def self.patch
+    puts 'Applying current rolling patchset. See https://www.mpfr.org/mpfr-current/#bugs'
+    puts 'for more information.'
+    system 'curl -Ls https://www.mpfr.org/mpfr-current/allpatches | patch -NZp1 --binary'
```
uberhacker, line 30: No checksum validation?

packages/gmp.rb
```diff
@@ -3,36 +3,36 @@
...
+    system "./configure \
+      #{CREW_OPTIONS} \
+      --disable-maintainer-mode \
+      --enable-cxx"
     system 'make'
   end
 
   def self.check
+    # Run tests on armv7l after install
     system 'make', 'check'
```
uberhacker, line 36: This will attempt to run tests for all architectures, not just armv7l.

satmandu: Yes. There's no workaround I could find aside from commenting out the check on armv7l, doing the install, and then taking out the hash and running the make check again.

satmandu: (same issue as with mpfr on x86_64)

uberhacker: What is the output of the failed tests?  Maybe we need to work on WHY the tests failed instead of trying to bypass?

PR #5685, opened 2021-04-24
Handle git branch only source

uberhacker:
Sometimes you might want to build from source from a git branch without a git_hashtag.

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -630,9 +630,13 @@ def download
         unless @pkg.git_branch.nil? || @pkg.git_branch.empty?
           # Leave a message because this step can be slow.
           puts "Downloading src from a git branch. This may take a while..."
-          system "git remote set-branches origin #{@pkg.git_branch}", exception: true
-          system "git fetch --progress origin #{@pkg.git_branch}", exception: true
-          system "git checkout #{@pkg.git_hashtag}", exception: true
+          unless @pkg.git_hashtag.nil? || @pkg.git_hashtag.empty?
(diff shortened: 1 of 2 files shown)


saltedcoffii:
That's rediculous, branches are moving targets and we can't support that. It would be like supporting `master` in a repository.

satmandu:
May I suggest a compromise?

I have bought into the idea that we should be using for tags and hashes instead of branches since that gets us closer to deterministic builds (that is, we should be able to build a package later and get the exact same build.)

However, lots of packages would benefit from being able to target git branches, since patches are often applied to non-master branches.

How about then being able to tag a package with the git hash of the most recent commit in a branch we are pulling from?

Isn't there a `git short log` command or something to that effect which might give that to us?

I would suggest creating a `package.build` file in the crew meta folder which would have a copy of the package file used, and we could use `sed` to substitute in the hash tag from `git log` which was used to build the file.

(You could store that file in the package archive during builds right next to filelist and float.)


TL;DR: I think building from branches should be supported, but then at the minimum we should record the actual git hash used during the build, so we can go back and figure out what was needed to create the build.

uberhacker:
Well, currently, if `git_branch` is provided in the package and `git_hashtag` is missing, the resulting error is not very graceful.  Maybe we should abort the build with a human readable message if one or the other properties are missing in the package?  Sometimes GitHub projects don't have release tags and the branch is the closest thing to a release.

saltedcoffii:
This goes back to my original code where one could specify a specific hash on a specific branch. I think that would be the most handy and something we can support.

PR #5687, opened 2021-04-24
Add darktable package

uberhacker:
darktable is an open source photography workflow application and raw developer.  See https://www.darktable.org/.  Tested on x86_64.  Depends on #5686.


saltedcoffii:
@uberhacker is there a reason this doesn't work with non em64t architectures?

uberhacker:
It will report when you try to compile on non-x86_64 architectures.

PR #5694, opened 2021-04-26
Crew Cache Cleanup

satmandu:
- Adds trailing slash to CREW_CACHE_DIR using technique from https://stackoverflow.com/a/12393692
- Link cached archive to and from CREW_BREW_DIR using hard links if possible, which is much faster than copying.
- Don't copy archive to cache if it is from cache

Works properly:
- [x] x86_64

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -578,26 +578,31 @@ def download
     case File.basename(filename)
     # Sources that download with curl
     when /\.zip$/i, /\.(tar(\.(gz|bz2|xz|lz))?|tgz|tbz|txz)$/i, /\.deb$/i
-      # Recall file from cache if requeseted
+      # Recall file from cache if requested
       if CREW_CACHE_ENABLED
+        puts "Looking for archive in cache".orange if @opt_verbose
         cachefile = CREW_CACHE_DIR + filename
-        if File.file?(cachefile)
+        if ! File.exist?(cachefile)
+          puts 'Cannot find cached archive. 😔 Will download.'.lightred
+          cachefile = ''
+        else
+          puts "Archive file exists in cache".lightgreen if @opt_verbose
           if Digest::SHA256.hexdigest( File.read(cachefile) ) == sha256sum then
             begin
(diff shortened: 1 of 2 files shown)


bin/crew
```diff
@@ -578,36 +578,58 @@ def download
...
+          cachefile = ''
+        else
+          puts "Archive file exists in cache".lightgreen if @opt_verbose
           if Digest::SHA256.hexdigest( File.read(cachefile) ) == sha256sum then
-            FileUtils.cp cachefile, CREW_BREW_DIR, verbose: @fileutils_verbose
-            puts "Archive found in cache".lightgreen
-            return {source: source, filename: filename}
+            begin
+              # Hard link cached file if possible.
+              FileUtils.ln cachefile, CREW_BREW_DIR, verbose: @fileutils_verbose
```
uberhacker, line 593: Why a hard link and why call it cachefile when you are linking to a directory?

lib/const.rb
```diff
@@ -35,11 +35,13 @@
   HOME = CREW_PREFIX + ENV['HOME']
 end
 
+# File.join ensures a trailing slash if one does not exist.
 if ENV['CREW_CACHE_DIR'].to_s == ''
```
uberhacker, line 39: `if ENV['CREW_CACHE_DIR'].to_s.empty`?

bin/crew
satmandu: CREW_BREW_DIR is a dir, but `ln filename dir` should make a hard linked file in `dir`, no?

lib/const.rb
satmandu: Sure, but that was the existing code...

bin/crew
uberhacker: I suppose that is true if the directory exists.

lib/const.rb
satmandu: I changed all  `.to_s == ''` to `.to_s.empty` .

PR #5697, opened 2021-04-27
GCC 11.1

satmandu:
- This has all the changes from the GCC 10.3.0 PR.
- The goal is to just go directly from GCC 10.2.1 to GCC 11.1, using the lessons learned from building GCC 10.3.0.
- Appears to be successfully building all software.
- There is an incompatibility with compiling mesa, but that is due to an llvm issue I'm trying to address with an llvm rebuild...

- Now to test upgrades from gcc10...

Works properly:
- [x] x86_64 
- [x] armv7l
- [x] i686


satmandu:
(Currently building.)

satmandu:
@uberhacker @saltedcoffii @supechicken If you would give this a try?

```export CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git
export CREW_TESTING_BRANCH=gcc11
CREW_TESTING=1 crew update
crew upgrade gcc10 libssp ; crew install gcc11
```

To undo:
```crew remove gcc11 ; CREW_TESTING=0 crew update ; crew install gcc10 libssp```

File: packages/gcc10.rb [hunk 1]
```diff
@@ -3,194 +3,12 @@
...
-        i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/gcc10/10.2.1-0f64_i686/gcc10-10.2.1-0f64-chromeos-i686.tar.xz',
-      x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/gcc10/10.2.1-0f64_x86_64/gcc10-10.2.1-0f64-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-     aarch64: 'dc0760c6fc123a48a72a6f5492f56d2be4273c29be3dfce6e7b202521c659365',
-      armv7l: 'dc0760c6fc123a48a72a6f5492f56d2be4273c29be3dfce6e7b202521c659365',
-        i686: '52850f19acb5c7def3b7116a5873a5c68c4f9a7fa254b5198deacff36f1e15aa',
-      x86_64: '438ae78481cbbfcc38c6d917498f6b6f0681627a798b263bb9b8ef222ee38b35',
-  })
+    depends_on 'libssp'
```
uberhacker, line 12: I guess I don't quite understand why we need a fake package for gcc10 if it only depends on libssp.

satmandu: With this upgrading gcc10 should:
- Remove the current gcc10
- Install libssp
So effectively this leads to a big drop in disk space used on most systems, without hopefully causing any compatibility issues for programs which might need libssp from gcc10. (This was the big issue with needing to use crewfix, no?)

saltedcoffii: maybe we can use the libssp from gcc11?

satmandu: @saltedcoffii We are in this PR. :)

uberhacker: OK, so I think I understand now why this is needed.  We eventually want to replace gcc10 with gcc11 but this provides an upgrade path.  Here is a wrinkle.  Why not just leave gcc10 with users that haven't upgraded yet?  New installs can have gcc11 but what does it hurt to allow users to stay at gcc10?  I'm more on the side of isolating gcc10, gcc11 and libssp than worrying about an upgrade.  Would it make sense to concentrate on an update-alternatives solution to allow switching between gcc versions?  We could provide a suggestion to upgrade to gcc11 from `crew update` but it would have to swap out gcc10.

satmandu: What I've done is not build gcc11 w/o libssp, but just pull libssp out of the install for gcc11. That seems to be much more stable. I think it could be fine to rebuild gcc10 this way, but we just then need to get gcc10 built that way, and leave a dependency for libssp in gcc 10 10.3.0.

That lets people keep gcc11. Let me see if I can make that happen...

satmandu: (but we need to figure out a way to have buildessential be ok with either `gcc10` or `gcc11`, otherwise any build will make crew install gcc10 or gcc11.

uberhacker: Yeah, we want to avoid the overlap of gcc10 and gcc11.  Both cannot be installed at the same time since they share the same executables and probably the same libraries.

satmandu: I still think some sort of provides ___ is the ideal way forward.

If crew is already building a list of compatible packages, it should be able to figure out a list of provides too, add that to the list of packages, and figure out if depends are satisfied...

saltedcoffii: on the contrary. Check out portage's slotting of gcc, it's quite impressive.

https://github.com/gentoo/gentoo/tree/master/sys-devel/gcc/

satmandu: @saltedcoffii those ebuild files are basically skeletons... there's so much abstracted away that I can't make heads or tails of what they're doing there.

(But I don't know whether one can emerge multiple gcc installs at the same time though on gentoo.)

saltedcoffii: The binutils packages might be easier to see slotting then, sorry about that.

uberhacker:
Does this PR obsolete #5671 because there appears to be a bit of an overlap?

satmandu:
> I think it's already included in `lib/packages.rb`

I'm going to keep the NPROC lines in until there's a patch to `packages.rb` to fix nproc when `make` isn't the first argument to `system`. Otherwise I forget to add this back in and builds take forever.

packages/gcc10.rb
```diff
@@ -1,63 +1,84 @@
...
-      x86_64: '438ae78481cbbfcc38c6d917498f6b6f0681627a798b263bb9b8ef222ee38b35',
-  })
-
-  depends_on 'unzip' => :build
-  depends_on 'mawk' => :build
-  depends_on 'dejagnu' => :build # for test
-  depends_on 'icu4c' => :build
-  depends_on 'python2' => :build
-  depends_on 'python3' => :build
+  source_url 'https://gcc.gnu.org/pub/gcc/releases/gcc-10.3.0/gcc-10.3.0.tar.xz'
```
uberhacker, line 10: Better url? `https://ftpmirror.gnu.org/gcc/gcc-10.3.0/gcc-10.3.0.tar.xz`

packages/libssp.rb
```diff
@@ -3,41 +3,92 @@
 class Libssp < Package
   description 'Libssp is a part of the GCC toolkit.'
   homepage 'https://gcc.gnu.org/'
-  version '10.3.0'
+  version '11.1.0'
   license 'GPL-3, LGPL-3, libgcc, FDL-1.2'
   compatibility 'all'
-  source_url 'https://ftpmirror.gnu.org/gcc/gcc-10.3.0/gcc-10.3.0.tar.xz'
-  source_sha256 '64f404c1a650f27fc33da242e1f2df54952e3963a49e06e73f6940f3223ac344'
+  source_url 'https://sourceware.org/pub/gcc/releases/gcc-11.1.0/gcc-11.1.0.tar.xz'
```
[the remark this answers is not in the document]
satmandu: When I built these the gnu mirrors hadn't synced the new release yet. :)

packages/libgcc_s1.rb
```diff
@@ -0,0 +1,355 @@
...
+    #cut -c2- ../dlist > dlistcut
+    #tail -n +2 dlistcut > dlist
+    ## Create new archive
+    #[[ -e "$startdir"/${newfile} ]] && mv  "$startdir"/${newfile} "$startdir"/${newfile}.old
+    #tar -c * --owner=chronos:1000 --group=chronos:1000 --use-compress-program='xz -9 -T 0' -f "$startdir"/${newfile}
+    #cd "$startdir"
+    #rm -rf $tmpdir
+    #sha256sum ${newfile} > ${newfile}.sha256
+  # extracttar gcc11-11.1.0-chromeos-x86_64.tar.xz libgcc_s1-11.1.0-chromeos-x86_64.tar.xz /usr/local/lib64/libgcc_s.so.1
+  end
```
uberhacker, line 354: Did you see my earlier comment regarding commenting everything out?

packages/gcc11.rb
```diff
@@ -0,0 +1,392 @@
...
+                                                "#{CREW_PREFIX}/bin/gcc -v 2>&1 | tail -1 | cut -d' ' -f3")
+    if status.success?
+      installed_gccver = stdout_and_stderr.chomp
+      # One gets "1:" with no gcc installed.
+      unless installed_gccver.to_s == '1:' ||
+             installed_gccver.to_s == 'No' ||
+             installed_gccver.to_s == 'not' ||
+             installed_gccver.to_s == 'gcc:' ||
+             installed_gccver.to_s == gcc_version.to_s
+        abort "GCC version #{installed_gccver} already installed.".lightgreen
```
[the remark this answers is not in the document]
uberhacker: What happens if you get rid of the `2>&1`?  Does it return an error status?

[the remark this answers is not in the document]
satmandu: Also this...
```
    if File.exist?("#{CREW_PREFIX}/bin/gcc")
      stdout_and_stderr, status = Open3.capture2e('bash', '-c',
                                                  "#{CREW_PREFIX}/bin/gcc -dumpversion | awk '{print $1}'")
      if status.success?
        @gccver = stdout_and_stderr.chomp
        unless @gccver.to_s == 'ccache' ||
               @gccver.to_s == ''
        end
      end
      else gccver = '11'
    end
```

packages/libgcc_s1.rb
satmandu: yup. It's not a great solution. We just need to figure out a way to do this properly...

Realisically we need to get this done before llvm 12 gets merged too, since just some of the runtime libs are all many people will need.

packages/gcc11.rb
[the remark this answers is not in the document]
uberhacker: It might make sense to add generic logic to crew and then set a property like `install 'strict'` or something to indicate this version cannot be installed concurrently with another version.  So the idea is you basically wouldn't need to add this logic section to every package with similar restrictions.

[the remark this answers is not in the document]
satmandu: Does not renaming gcc10 -> gcc version 10 and gcc11 -> gcc version 11 give us that for free?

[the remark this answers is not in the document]
satmandu: If they are both named gcc, then one in theory could upgrade to gcc version 10, or to gcc version 11, and it would replace the other, since crew does an uninstall before an upgrade.

Luckily now an uninstall of gcc doesn't kill crew. :)

uberhacker:
This PR is not tagged with Draft but since it conflicts with another pending PR, does this mean it's not ready to merge?

satmandu:
Leaving it a draft at the moment untilI figure out a way to deal with libgcc_s1...

PR #5699, opened 2021-04-27
modify crew: Strip at build time, use fewer CMAKE options, uninstall pkg if filelist exists

satmandu:
- strip at build time too
- shrink CREW_CMAKE_OPTIONS
- allow uninstall of packages if CREW_META/package.filelist exists but package is not listed as installed.

Works properly:
- [x] x86_64


bin/crew
```diff
@@ -816,6 +816,14 @@ def prepare_package (destdir)
 
     # remove temporary files
     FileUtils.rm_rf ['dlistcut', '../dlist', '../filelist'], verbose: @fileutils_verbose
+
+    # Strip libraries with -S
+    strip_find_files "find . -type f -name 'lib*.a' -print", "-S"
+    strip_find_files "find . -type f -name 'lib*.so*' -print", "-S"
+
+    # Strip binaries but not compressed archives
+    strip_find_files "find . -type f ! -iname '*\.bz2' ! -iname '*\.gz' ! -iname '*\.lha' ! -iname '*\.lz' ! -iname '*\.rar' ! -iname '*\.tar' ! -iname '*\.tbz' ! -iname '*\.tgz' ! -iname '\*.txz' ! -iname '*\.xz' ! -iname '*\.zip' -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d'"
```
uberhacker, line 825: Should we put these lines in a function since the same lines are called in install_package?  Or do we even need to call the lines in install_package?

bin/crew
```diff
@@ -1069,7 +1077,7 @@ end
 def remove (pkgName)
 
   #make sure the package is actually installed
-  unless @device[:installed_packages].any? { |pkg| pkg[:name] == pkgName }
+  unless @device[:installed_packages].any? { |pkg| pkg[:name] == pkgName } || File.file?("#{CREW_META_PATH}#{pkgName}.filelist")
```
uberhacker, line 1080: Should this be `File.exist?` instead of `File.file?`?

lib/const.rb
```diff
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.8.11'
+CREW_VERSION = '1.8.11.1'
```
uberhacker, line 3: It's fine to go ahead and increment to `1.8.12`.  I don't want us to get into crazy versioning.

bin/crew
satmandu: `strip_find_files` is already a function though. I think we will need a transition period to strip files at install for packages which weren't already stripped at build time, after which we can remove the install portion.

bin/crew
satmandu: `File.file?` is used elsewhere, which is why I used it there.  `File.exist?` would work fine too. 

If it is worth it I can change it...

lib/const.rb
satmandu: Sure. :)

bin/crew
uberhacker: Yes but the parameters are not.  What happens if we need to exclude another extension?  This way we would need to change in 2 places instead of just 1 if it was encapsulated in a function.  I'm just thinking like a DRY kind of guy.

bin/crew
uberhacker: Doesn't `File.file?` just determine if it's a file and then fallback to false if it doesn't find it?  `File.exist?` returns a boolean true or false based on the true intention of what we are trying to determine.

bin/crew
satmandu: If you want to take a stab at converting that to a function, go ahead? I'm just trying to have a light touch here... I'm not sure making a function out of this would make it any faster...

uberhacker: I'm not concerned about speed, only not repeating the same lines multiple times.

uberhacker: How about?
```
def strip_binaries()
  extensions = [ 'bz2', 'gz', 'lha', 'lz', 'rar', 'tar', 'tbz', 'tgz', 'txz', 'xz', 'zip' ]
  inames = extensions.join(' ! -iname *\.')
  strip_find_files "find . -type f#{inames} -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d'"
end
```
Then just call `strip_binaries` in prepare_package and install_package.

saltedcoffii: if we have .LHA files, we might also want to add .Z and .LHZ

uberhacker: Thank you @saltedcoffii.  This is exactly why we need a function for this @satmandu. :)

satmandu: Done,

bin/crew
```diff
@@ -828,6 +831,18 @@ def strip_find_files (find_cmd, strip_option = "")
   system "#{find_cmd} | xargs -r sh -c 'for i in \"$0\" \"$@\"; do case \"$(head -c 4 $i)\" in ?ELF|\!?ar) echo \"$i\";; esac ; done' | xargs -r llvm-strip #{strip_option}"
 end
 
+def strip_dir (dir)
+  Dir.chdir dir do
+    # Strip libraries with -S
+    puts "Stripping libraries..."
+    strip_find_files "find . -type f  \\( -name 'lib*.a' -o -name 'lib*.so*' \\) -print", "-S"
```
uberhacker, line 838: Why is the `-o` in the middle of `-name`s?

bin/crew
```diff
@@ -828,6 +831,18 @@ def strip_find_files (find_cmd, strip_option = "")
...
+
+    # Strip binaries but not compressed archives
+    puts "Stripping binaries..."
+    strip_find_files "find . -type f ! -iname '*\.bz2' ! -iname '*\.gz' ! -iname '*\.lha' ! -iname '*\.lz' ! -iname '*\.rar' ! -iname '*\.tar' ! -iname '*\.tbz' ! -iname '*\.tgz' ! -iname '\*.txz' ! -iname '*\.xz' ! -iname '*\.zip' -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d'"
```
uberhacker, line 842: An array with join would be a bit more elegant.  It would be easier to add extensions also.

bin/crew
satmandu: Either lib*.a or lib*.so* ...

satmandu: combines two lines into one.

bin/crew
satmandu: Suggested code?

uberhacker: I already added suggested code but you force pushed and got rid of it.

bin/crew
uberhacker: Gotcha, nice!

bin/crew
uberhacker:
```
extensions = [ 'bz2', 'gz', 'lha', 'lz', 'rar', 'tar', 'tbz', 'tgz', 'txz', 'xz', 'zip' ]
inames = extensions.join(' ! -iname *\.')
strip_find_files "find . -type f ! -iname *\.#{inames} -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d'"
```

satmandu: Thanks! And sorry about that!

satmandu: Added.

bin/crew
```diff
@@ -828,6 +831,20 @@ def strip_find_files (find_cmd, strip_option = "")
...
+  Dir.chdir dir do
+    # Strip libraries with -S
+    puts "Stripping libraries..."
+    strip_find_files "find . -type f  \\( -name 'lib*.a' -o -name 'lib*.so*' \\) -print", "-S"
+
+    # Strip binaries but not compressed archives
+    puts "Stripping binaries..."
+    extensions = [ 'bz2', 'gz', 'lha', 'lz', 'rar', 'tar', 'tbz', 'tgz', 'txz', 'xz', 'zip' ]
+    inames = extensions.join(' ! -iname *\.')
+    strip_find_files "find . -type f ! -iname *\.#{inames} -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d'"
```
uberhacker, line 844: Too fast. :)  Was this tested?

satmandu: Tested insofar as it is not throwing any errors:

```
-- Installing: /usr/local/tmp/crew/dest/usr/local/share/vulkan/registry/spec_tools
-- Installing: /usr/local/tmp/crew/dest/usr/local/share/vulkan/registry/spec_tools/util.py
-- Installing: /usr/local/tmp/crew/dest/usr/local/share/vulkan/registry/reg.py
Checking...
Stripping libraries...
Stripping binaries...
Archiving...
vulkan_headers is built!
```

uberhacker: Also, did you catch @saltedcoffii's suggested additional extensions?

PR #5705, opened 2021-04-28
add make to lsb_release deps

### LINKED ISSUE #5703: Add dependency for make in lsb_release.rb
## Before you submit an issue
Please look up our issues, maybe your problem was mentioned before.

## Description
Provide a short description of your problem here.

lsb_release fails to install without make installed

```
chronos@localhost ~/tmp $ crew install lsb_release
lsb_release: Linux Standard Base
https://wiki.linuxfoundation.org/lsb/start
Version: 1.4-1
License: GPL-2
########################################################################################################### 100.0%
########################################################################################################### 100.0%
Archive downloaded.
lsb_release failed to install: `make V=0 -j4` exited with 127

```

- [ ] Add make as dependency for lsb_release.rb
- [ ] profit!

---

## That's it
Thank you for submitting your issue.

satmandu:
Fixes #5703
- Add `make` to deps for `lsb_release`
- make both `make` and `help2man` build deps

Works properly:
- [x] x86_64

### DIFF
--- a/packages/lsb_release.rb
+++ b/packages/lsb_release.rb
@@ -14,8 +14,8 @@ class Lsb_release < Package
   binary_sha256 ({
   })
 
-  depends_on 'help2man' => :build
-  depends_on 'make' => :build
+  depends_on 'help2man'
+  depends_on 'make'
 
   def self.build
     system "cp /etc/lsb-release /tmp"
--- a/packages/lsb_release.rb
+++ b/packages/lsb_release.rb
@@ -15,6 +15,7 @@ class Lsb_release < Package
   })
 
   depends_on 'help2man'
+  depends_on 'make'
 
   def self.build
     system "cp /etc/lsb-release /tmp"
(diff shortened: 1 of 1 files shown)


packages/lsb_release.rb
```diff
@@ -14,7 +14,8 @@ class Lsb_release < Package
   binary_sha256 ({
   })
 
-  depends_on 'help2man'
+  depends_on 'help2man' => :build
+  depends_on 'make' => :build
```
uberhacker, line 18: Don't these both already get installed from buildessential which is necessary before any build?

supechicken (approved):

packages/lsb_release.rb
satmandu: In theory? @esprout could you try a `crew update && crew install lsb_release` and see if that works?

uberhacker: I see the issue here.  Since this doesn't have pre-built binaries, these need to be runtime dependencies since make needs to execute with `crew install` in addition to `crew build`.  The build process is specific to each system so pre-built binaries are not possible with this package.

satmandu: Changed these to runtime deps.

PR #5707, opened 2021-04-29
Add uninstall script

supechicken:
Although `rm -rf /usr/local/*` is a quick way to do it, it will also delete files created by others

This script will delete `crew` itself and all installed packages

Tested on `x86_64`


saltedcoffii:
pip isn't registered by anything in chromebrew's filelist right now, are you sure it works after running `crew postinstall python3`?

uberhacker:
This will only partially remove crew.  There will be some residue left behind.  The simplest way to uninstall (barring there are no mounts or chroots) is the following:
```
$ rm -rf /usr/local
```
If there are mounts or chroots, you will need to remove everything else and leave those untouched.

satmandu (commented):
I would remove directories in the crew meta directory list files only if they are empty.

PR #5711, opened 2021-04-29
ninja and re2c update

satmandu:
Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/ninja.rb
+++ b/packages/ninja.rb
@@ -32,7 +32,6 @@ class Ninja < Package
     system "env #{CREW_ENV_OPTIONS} \
       AR=gcc-ar \
       python3 configure.py --bootstrap"
-    system 'ninja'
   end
 
   def self.install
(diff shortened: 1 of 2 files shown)


packages/ninja.rb
```diff
@@ -3,30 +3,40 @@
...
+  def self.patch
+    system 'filefix'
+  end
+
   def self.build
-    system 'python3 configure.py --bootstrap'
+    system "env #{CREW_ENV_OPTIONS} \
+      AR=gcc-ar \
+      python3 configure.py --bootstrap"
+    system 'ninja'
```
supechicken, line 35: Do you mean `system './ninja'`?

satmandu: Let me check...

satmandu: Turns out we don't need that line at all!

supechicken (approved):

PR #5715, opened 2021-05-01
Add cpu_x package

uberhacker:
CPU-X is a Free software that gathers information on CPU, motherboard and more.  See https://x0rg.github.io/CPU-X/.  Depends on #5714.

### DIFF
--- a/packages/cpu_x.rb
+++ b/packages/cpu_x.rb
@@ -1,0 +1,42 @@
+require 'package'
+
+class Cpu_x < Package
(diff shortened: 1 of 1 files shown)


satmandu:
If you're going to set and unset an environment variable, maybe store the previous state of the variable and restore it after done instead?

uberhacker:
> If you're going to set and unset an environment variable, maybe store the previous state of the variable and restore it after done instead?

Done

PR #5720, opened 2021-05-03
allow verbose crew list installed

satmandu:
- Adds option to list packages with versions.
- `crew list -v installed`
- The big advantage here is the ability to grep to get the version of an installed package quickly, e.g. 
```
crew list -v installed | grep meson
meson                      0.58.0
```

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686


uberhacker:
`crew search meson -v` already does this without needing to grep. :)

PR #5726, opened 2021-05-04
Lzma Cleanup & replacement with newer xzutils, Add pixz support

satmandu:
- Replaces lzma with newer xzutils, which also supports pixz (allows indexed xz archives, which allows one to extract specific files without reading an entire archive.)
- Adds pixz
- Adds logic to crew to handle pixz archives (this will need some updating once I figure out a good way to pass libraries we need for packages from other packages to crew).
- Adds the logic from the other day for easier perl package builds.

Working:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -595,7 +595,7 @@ def download
 
     case File.basename(filename)
     # Sources that download with curl
-    when /\.zip$/i, /\.(tar(\.(gz|bz2|xz|lz))?|tgz|tbz|txz)$/i, /\.deb$/i, /\.AppImage$/i
+    when /\.zip$/i, /\.(tar(\.(gz|bz2|tpxz|xz|lz))?|tgz|tbz|txz)$/i, /\.deb$/i, /\.AppImage$/i
       # Recall file from cache if requested
       if CREW_CACHE_ENABLED
         puts "Looking for archive in cache".orange if @opt_verbose
@@ -725,6 +725,9 @@ def unpack (meta)
       Dir.chdir @extract_dir do
         system "../#{meta[:filename]} --appimage-extract", exception: true
       end
+    when /\.tpxz$/i
+      puts "Unpacking 'tpxz' archive using 'tar', this may take a while..."
+      system "pixz -d #{meta[:filename]} | tar x#{@verbose} -C #{@extract_dir}", exception: true
     end
     if meta[:source] == true
       # Check the number of directories in the archive
@@ -810,6 +813,12 @@ def prepare_package (destdir)
     # https://www.debian.org/doc/debian-policy/ch-docs.html#info-documents
     FileUtils.rm "#{CREW_DEST_PREFIX}/share/info/dir" if File.exist?("#{CREW_DEST_PREFIX}/share/info/dir")
 
+    # Remove all perl module files which will conflict
+    if @pkg.name.include? 'perl_'
+      puts "Removing .packlist and perllocal.pod files to avoid conflicts with other perl packages.".orange
+      system "find #{CREW_DEST_DIR} -type f \\( -name '.packlist' -o -name perllocal.pod \\) -delete" 
+    end
+
     # compress manual files
     compress_doc "#{CREW_DEST_PREFIX}/man"
     compress_doc "#{CREW_DEST_PREFIX}/info"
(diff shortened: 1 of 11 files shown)


bin/crew
```diff
@@ -234,9 +234,13 @@ def generate_compatible
   Dir[CREW_PACKAGES_PATH + '*.rb'].each do |filename|
     pkgName = File.basename filename, '.rb'
     set_package pkgName, filename
+    puts "Checking #{pkgName} for compatibility.".orange if @opt_verbose
     if @pkg.compatibility.include? 'all' or @pkg.compatibility.include? ARCH
       #add to compatible packages
+      puts "Adding #{pkgName} to compatible packages.".lightgreen if @opt_verbose
       @device[:compatible_packages].push(name: @pkg.name)
+    else
+      puts "#{pkgName} not a compatible packages.".lightred if @opt_verbose
```
uberhacker, line 243: `not a compatible package.` singular?

packages/pixz.rb
```diff
@@ -0,0 +1,46 @@
...
+require 'package'
+
+class Pixz < Package
+  description 'Parallel, indexed xz compressor'
+  homepage 'https://github.com/vasi/pixz'
+  version '1.0.7-0829'
+  license 'BSD'
+  compatibility 'all'
+  source_url 'https://github.com/vasi/pixz.git'
+  git_hashtag '0829c7315c804a4e40abd63a9d624194dc1e4f0a'
```
saltedcoffii, line 13: why are you using a different hashtag instead of v1.0.7?

packages/pixz.rb
```diff
@@ -0,0 +1,46 @@
...
+      manpage=true \
+      ./configure #{CREW_OPTIONS}"
+    system 'make'
+  end
+
+  def self.install
+    system "make DESTDIR=#{CREW_DEST_DIR} install"
+  end
+  unless File.exist?("#{CREW_PREFIX}/bin/pixz")
+    puts 'Pixz is not installed. Please install it with \'crew install pixz\''.lightred
```
saltedcoffii, line 44: What's the point of this review?

saltedcoffii, line 44: What's the point of this line?

bin/crew
uberhacker: This seems like a bit of an overkill to print these messages for every package.  We're talking over 1500 lines.

bin/crew
```diff
@@ -1122,9 +1138,18 @@ def build_package (pwd)
 end
 
 def archive_package (pwd)
-  pkg_name = "#{@pkg.name}-#{@pkg.version}-chromeos-#{@device[:architecture]}.tar.xz"
-  Dir.chdir CREW_DEST_DIR do
-    system "tar c#{@verbose}Jf #{pwd}/#{pkg_name} *"
+  unless ENV['CREW_USE_PIXZ'] == '1'
```
uberhacker, line 1141: Why do we need an environment variable set to do this?

satmandu: Default should be (for now at least) tar.xz. But especially for large packages, one could set that to generate `.tpxz` archives instead using `pixz`.  If there's a better way to do that...

bin/crew
uberhacker: I guess since it only happens with the verbose option, it's fine.

packages/pixz.rb
satmandu: pixz is needed for extracting tpxz files from crew... so until we make it part of core... this will be a tiny annoyance.

Is that bad?

packages/pixz.rb
satmandu: Great question. There are actually some bugfixes more recently which we should have, e.g.
https://github.com/vasi/pixz/commit/1aeb09b868dcac05830c513c02dcbe587d3f44e1

bin/crew
satmandu: Thanks!

bin/crew
satmandu: (trying to be conservative here with changes.)

satmandu: Later we can flip to default or not... Let's see compression ratios and how useful tpxz ends up being?

uberhacker: You are supposed to be doing that in your testing already. :)

satmandu: This is an inital framework for getting packages stored using tpxz.

I don't want to create bugs by using a standalone crew version for testing... I have to dogfood this, which means using it production with feature flags hiding things. The env variables are our version of feature flags, no?

uberhacker: Don't you always create a standalone crew version for testing when you work locally?  It's fine to make this the default if has been fully tested and provides a benefit to improve the compression scheme.

packages/pixz.rb
uberhacker: Wait, what?  This will run for every command, won't it?  What happens with `crew search pixz`?  Also, you should probably explain to the user why they would need to install pixz.

saltedcoffii:
Maybe instead of using pixz we could just use gnu ar like deb files do? That might be simpler, I don't know.

bin/crew
satmandu: Yes I create a local version.... using CREW_TESTING...

We need the option to create two separate archive formats since there's a bootstrapping issue with new installs.

packages/pixz.rb
satmandu: I can get rid of this, since crew prompts if pixz isn't installed.

satmandu: Removed.

satmandu:
@saltedcoffii The advantage of pixz is that we could possibly do file installs from pixz packages without downloading the entire archive file. That could be much faster than doing installs of other types of archives.

PR #5727, opened 2021-05-04
Fix dependency algorithm

supechicken:
Fixed the dependency algorithm by rearranging the steps of adding dependencies into the `@dependency` array          

Tested on `x86_64`


bin/crew
```diff
@@ -1130,53 +1091,40 @@ def archive_package (pwd)
...
-        end
-      end
-
-      #remove the file and directory list
-      File.unlink "meta/#{pkgName}.filelist"
-      File.unlink "meta/#{pkgName}.directorylist"
-
+  # make sure the package is actually installed
+  unless installed?(pkgName, mode: 'filelist')
+    return puts "Package #{pkgName} isn't installed.".lightred
```
uberhacker, line 1096: Why are returning a value when this isn't in a function?

bin/crew
```diff
@@ -1256,9 +1209,9 @@ end
 
 def postinstall_command (args)
   args["<name>"].each do |name|
-    @pkgName = name
-    search @pkgName, true
-    if @device[:installed_packages].any? do |elem| elem[:name] == @pkgName end
+    if installed?(name)
+      @pkgName = name
```
uberhacker, line 1213: Do you even need to assign `@pkgName`?  You could just use name throughout.

bin/crew
supechicken: The `return` block is used to jump out the function (it will return `nil` as `puts` will returns nothing)

lib/dependencyutils.rb
```diff
@@ -0,0 +1,36 @@
...
+          @checking << @pkg.name
+          push_dependencies
+          @dependencies << dep
+        end
+      end
+    end
+    push_dependencies
+    remove_instance_variable(:@checking)
+  end
+end
```
uberhacker, line 36: How does this ensure the dependencies are installed in the correct order?  For instance, is libpng installed before shared_mime_info or sommelier before its dependencies?

bin/crew
supechicken: Change made

bin/crew
```diff
@@ -964,31 +950,6 @@ def resolve_dependencies
   end_packages = []
   first_packages = ["shared_mime_info"]
   last_packages = ["curl", "ghc", "mandb", "gtk3", "gtk4", "sommelier"]
```
uberhacker, line 952: If you removed all the code below, you might as well delete the 5 lines above also.

bin/crew
uberhacker: But this isn't a function.

lib/dependencyutils.rb
supechicken: Does `libpng` a dependency of them?

If current package depends on package A depends on pkgB depends on pkgC:
```
dependencies of current one:
pkgA
|- pkgB
   |- pkgC
```

The old algorithm will handle like this:
- Add pkgA to dependency array
- set current package to pkgA deps,each
- go back to step 1 but pkgA changed to pkgB
===snip===

Now it will handle like this:
- set current package to pkgA deps,each
- go back to step 1 but pkgA changed to pkgB
===snip===
- Add pkgA to dependency array

bin/crew
supechicken: > This line is under `def remove`

bin/crew
supechicken: Do we need to move sommelier to the last dependency now?

uberhacker: Um...sommelier needs to be the last dependency because there are postinstall instructions that users shouldn't miss and if your algorithm is as foolproof as you claim, it should already be the last package installed.

lib/dependencyutils.rb
uberhacker: shared_mime_info is a dependency of libpng so the install order should make shared_mime_info ahead of libpng.

bin/crew
satmandu: It would be nice to create a class of postinstall message that could be queued and then dumped at the very end after installs are done.

You could add to an array from each install, and when the queue of installs is empty, then print that array out.  That way you don't lose the message if packages during installs need to be built, especially if any packages depend on sommelier and install sommelier before building another package.

uberhacker: Sounds like a good next pull request. ;)

bin/crew
```diff
@@ -1130,53 +1091,40 @@ def archive_package (pwd)
...
-        rescue => exception #swallow exception
-        end
-      end
-
-      #remove the file and directory list
-      File.unlink "meta/#{pkgName}.filelist"
-      File.unlink "meta/#{pkgName}.directorylist"
-
+  # make sure the package is actually installed
+  unless installed?(pkgName, mode: 'filelist')
```
satmandu, line 1095: Please add back the `|| File.exist?("#{CREW_META_PATH}#{pkgName}.filelist")` - since it useful to remove packages for which package files have disappeared.

bin/crew
```diff
@@ -1202,54 +1157,41 @@ def archive_package(pwd)
...
+      verbose: @fileutils_verbose
 
+    # remove all directories installed by the package
+    puts "Removing directories".lightred if @opt_verbose
+    begin
+      FileUtils.rmdir \
+        File.readlines("meta/#{pkgName}.directorylist", chomp: true),
+        verbose: @fileutils_verbose
+    rescue Errno::EROFS # Read-only file system @ dir_s_rmdir
     end
```
uberhacker, line 1181: How is this better?  You got rid of all the verbosity so there is no way to see all the files and directories that are removed.

lib/dependencyutils.rb
```diff
@@ -0,0 +1,36 @@
...
```
uberhacker, line 36: Does this resolve in the correct order?  That would be the only reason to rewrite everything.

PR #5728, opened 2021-05-04
Keep crew from crashing due to a malformed package file.

satmandu:
- "Thou shalt not let invalid input crash thy program"
- Make `generated_compatible` threaded, letting crew withstand malformed package files during `crew update`
- Try removing a compatible line from a package file and see how crew doesn't crash.
image

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -266,9 +266,9 @@ def generate_compatible
         puts "Thread Error: #{e}".orange unless e.to_s.include?('queue empty') if @opt_verbose
       end
     end
-  end; "ok"
+  end
   begin
-    workers.map(&:join); "ok"
+    workers.map(&:join)
   rescue => exception
     puts "Explicitly caught - #{exception.class}: #{exception.message}" if @opt_verbose
   end
--- a/bin/crew
+++ b/bin/crew
@@ -140,7 +140,6 @@ def print_package(pkgPath, extra = false)
 end
 
 def print_current_package(extra = false)
-  pkgName = @pkg.name
   status = ''
   status = 'installed' if @device[:installed_packages].any? do |elem| elem[:name] == @pkg.name end
   status = 'incompatible' unless @device[:compatible_packages].any? do |elem| elem[:name] == @pkg.name end
@@ -229,7 +228,7 @@ def list_compatible(compat = true)
       if ( File.exist? CREW_META_PATH + pkgName + '.filelist' ) && compat
         puts pkgName.lightgreen
       else
-        puts pkgName unless !compat
+        puts pkgName if compat
       end
     else
       puts pkgName.lightred
--- a/bin/crew
+++ b/bin/crew
@@ -217,7 +217,11 @@ end
 def list_compatible(compat = true)
   Dir[CREW_PACKAGES_PATH + '*.rb'].each do |filename|
     pkgName = File.basename filename, '.rb'
-    set_package pkgName, filename
+    begin
+      set_package pkgName, filename
+    rescue => e
+      puts "Error with #{pkgName}.rb: #{e}".red unless e.to_s.include?('uninitialized constant')
+    end
     if compat
       if @pkg.compatibility&.include? 'all' or @pkg.compatibility&.include? ARCH
         if File.exist? CREW_META_PATH + pkgName + '.filelist'
--- a/bin/crew
+++ b/bin/crew
@@ -259,7 +259,7 @@ def generate_compatible
               puts "#{pkgName} is not a compatible package.".lightred if @opt_verbose
             end
           rescue => e
-            puts "Error with #{pkgName}.rb: #{e}".red
+            puts "Error with #{pkgName}.rb: #{e}".red unless e.to_s.include?('uninitialized constant') 
           end
         end
       rescue => e
--- a/bin/crew
+++ b/bin/crew
@@ -142,8 +142,8 @@ end
 def print_current_package(extra = false)
   pkgName = @pkg.name
   status = ''
-  status = 'installed' if @device[:installed_packages].any? do |elem| elem[:name] == pkgName end
-  status = 'incompatible' unless @pkg.name.in? @device[:compatible_packages]
+  status = 'installed' if @device[:installed_packages].any? do |elem| elem[:name] == @pkg.name end
+  status = 'incompatible' unless @device[:compatible_packages].any? do |elem| elem[:name] == @pkg.name end
   case status
   when 'installed'
     print @pkg.name.lightgreen
@@ -225,7 +225,7 @@ def list_compatible(compat = true)
   generate_compatible
   Dir[CREW_PACKAGES_PATH + '*.rb'].each do |filename|
     pkgName = File.basename filename, '.rb'
-    if pkgName.in? @device[:compatible_packages]
+    if @device[:compatible_packages].any? do |elem| elem[:name] == pkgName end
       @compatibility = true
     else
       @compatibility = false
@@ -999,7 +999,7 @@ def expand_dependencies
 end
 
 def resolve_dependencies
-  abort "Package #{@pkg.name} is not compatible with your device architecture (#{ARCH}) :/".lightred unless @pkg.name.in? @device[:compatible_packages]
+  abort "Package #{@pkg.name} is not compatible with your device architecture (#{ARCH}) :/".lightred unless @device[:compatible_packages].any? do |elem| elem[:name] == @pkg.name end
 
   @dependencies = []
   if @pkg.build_from_source
--- a/bin/crew
+++ b/bin/crew
@@ -275,6 +275,7 @@ def generate_compatible
     output = JSON.parse @device.to_json
     file.write JSON.pretty_generate(output)
   end
+  puts 'Generating compatible packages done.'.orange if @opt_verbose
 end
 
 def search (pkgName, silent = false)
@@ -529,7 +530,6 @@ def update
 
   #update compatible packages
   generate_compatible
-  puts 'Generating compatible packages done.'.orange if @opt_verbose
   #check for outdated installed packages
   puts 'Checking for package updates...'
 
--- a/bin/crew
+++ b/bin/crew
@@ -306,7 +306,7 @@ def regexp_search(pkgPat)
       end
     end
   end
-  abort "Package #{pkgPat} not found. :(".lightred unless results.length.positive?
+  abort "Package #{pkgPat} not found. :(".lightred if results.empty?
 end
 
 def help(pkgName)
@@ -594,7 +594,7 @@ def upgrade
       end
     end
 
-    if toBeUpdated.length.positive?
+    unless toBeUpdated.empty?
       puts 'Updating packages...'
       toBeUpdated.each do |package|
         search package
--- a/bin/crew
+++ b/bin/crew
@@ -850,7 +850,7 @@ def prepare_package (destdir)
     FileUtils.rm "#{CREW_DEST_PREFIX}/share/info/dir" if File.exist?("#{CREW_DEST_PREFIX}/share/info/dir")
     
     # Remove all perl module files which will conflict
-    if @pkg.name.include? 'perl_'
+    if @pkg.name ~= /^perl_/
       puts "Removing .packlist and perllocal.pod files to avoid conflicts with other perl packages.".orange
       system "find #{CREW_DEST_DIR} -type f \\( -name '.packlist' -o -name perllocal.pod \\) -delete" 
     end
--- a/bin/crew
+++ b/bin/crew
@@ -246,7 +246,6 @@ def generate_compatible
   puts 'Generating compatible packages...'
   @device[:compatible_packages] = []
   Dir[CREW_PACKAGES_PATH + '*.rb'].each do |filename|
-
     pkgName = File.basename filename, '.rb'
     begin
       set_package pkgName, filename
--- a/bin/crew
+++ b/bin/crew
@@ -777,7 +777,7 @@ def unpack(meta)
       # Check the number of directories in the archive
       entries = Dir["#{@extract_dir}/*"]
       entries = Dir[@extract_dir] if entries.empty?
-      if entries.length.zero?
+      if entries.empty?
         abort "Empty archive: #{meta[:filename]}".lightred
       elsif entries.length == 1 && File.directory?(entries.first)
         # Use `extract_dir/dir_in_archive` if there is only one directory.
--- a/bin/crew
+++ b/bin/crew
@@ -183,7 +183,7 @@ def list_available
     notInstalled = false if File.exists? CREW_META_PATH + pkgName + '.filelist'
     if notInstalled
       set_package pkgName, filename
-      if @pkg.compatibility.include? 'all' or @pkg.compatibility.include? ARCH
+      if @pkg.compatibility&.include? 'all' or @pkg.compatibility&.include? ARCH
         puts pkgName
       else
         puts pkgName.lightred
@@ -219,7 +219,7 @@ def list_compatible(compat = true)
     pkgName = File.basename filename, '.rb'
     set_package pkgName, filename
     if compat
-      if @pkg.compatibility.include? 'all' or @pkg.compatibility.include? ARCH
+      if @pkg.compatibility&.include? 'all' or @pkg.compatibility&.include? ARCH
         if File.exist? CREW_META_PATH + pkgName + '.filelist'
           puts pkgName.lightgreen
         else
@@ -227,7 +227,7 @@ def list_compatible(compat = true)
         end
       end
     else
-      unless @pkg.compatibility.include? 'all' or @pkg.compatibility.include? ARCH
+      unless @pkg.compatibility&.include? 'all' or @pkg.compatibility&.include? ARCH
         puts pkgName.lightred
       end
     end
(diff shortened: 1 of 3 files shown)


uberhacker:
But don't we want crew to crash if there is a missing compatibility line?  How else would we know there is a problem and what is the true compatibility of the package?  I think we may be overthinking this a bit.

satmandu:
> But don't we want crew to crash if there is a missing compatibility line? How else would we know there is a problem and what is the true compatibility of the package? I think we may be overthinking this a bit.

Hard no.

You want an error message.  But you never want `crew update` to crash, because the end user doesn't actually know if syncing down new changes has failed when they see the crash message. This means that they don't know if there's a problem that can be fixed at all by running it again. It's one thing to say one file is broken, or that the network is down. Those are useful.

But I think a crash just says the software has insufficient error handling.

satmandu:
With this update broken files are ignored, but you still get an error message about the broken files.

satmandu:
Also a "Please file a bug at ..." message could be added.

uberhacker:
This goes back to my earlier point that we should not be adding broken package files to the repo in the first place. :)  But point taken about error messages versus hard crashes.

uberhacker:
We could get carried away with `begin...rescue` blocks for error trapping if we wanted.

bin/crew
```diff
@@ -231,14 +236,42 @@ end
...
+              #add to compatible packages
+              puts "Adding #{pkgName} to compatible packages.".lightgreen if @opt_verbose
+              @semaphore.synchronize do
+                @device[:compatible_packages].push(name: @pkg.name)
+              end
+            else
+              puts "#{pkgName} is not a compatible package.".lightred if @opt_verbose
+            end
+          rescue => e
+            puts "Error with #{pkgName}.rb: #{e}".red unless e.to_s.include?('uninitialized constant')
```
uberhacker, line 262: Why is `'uninitialized constant'` not considered an error?

bin/crew
```diff
@@ -809,6 +848,12 @@ def prepare_package (destdir)
     # than install-info.
     # https://www.debian.org/doc/debian-policy/ch-docs.html#info-documents
     FileUtils.rm "#{CREW_DEST_PREFIX}/share/info/dir" if File.exist?("#{CREW_DEST_PREFIX}/share/info/dir")
+    
+    # Remove all perl module files which will conflict
+    if @pkg.name.include? 'perl_'
```
uberhacker, line 853: I'm wondering if we should make this a regex check like `if @pkg.name ~= /^perl_/` to make sure the package name starts with perl and not some rouge package named `this_aint_perl_dude` (not that anybody would create a package with that name but just to emphasize the point).

bin/crew
satmandu: That's what comes up here when there is a syntax error reported from `set_package`. This way you don't get an erroneous duplicate notification.

In reality we just need to deal with the compatibility check failing here.

bin/crew
satmandu: Sounds good.

satmandu:
> We could get carried away with `begin...rescue...end` blocks for error trapping if we wanted.

As we see new errors. This was just the low-hanging fruit so to speak...

bin/crew
satmandu: done

uberhacker:
Bump crew version in `lib/const.rb`?

uberhacker:
This conflicts with #5726 which was merged recently.

satmandu:
There may be a couple more commits to fix some of the other places syntax errors could crash crew. I'll take a look at those tomorrow, and resolve conflicts then.

bin/crew
supechicken: Better to move the `rescue` block to the `set_package` part?

bin/crew
```diff
@@ -231,14 +240,42 @@ end
...
-    pkgName = File.basename filename, '.rb'
-    set_package pkgName, filename
-    if @pkg.compatibility.include? 'all' or @pkg.compatibility.include? ARCH
-      #add to compatible packages
-      @device[:compatible_packages].push(name: @pkg.name)
+  @work_q = Queue.new
+  @semaphore = Mutex.new
+  Dir[CREW_PACKAGES_PATH + '*.rb'].each{|file| @work_q << file}
+  # This creates a 10 worker thread pool
+  workers = (0...10).map do
```
supechicken, line 247: Are the threads for improving the checking time?

bin/crew
```diff
@@ -4,6 +4,7 @@ require_relative '../lib/color'
 # Disallow sudo
 abort 'Chromebrew should not be run as root.'.lightred if Process.uid == 0
 
+require 'thread'
```
supechicken, line 7: `Thread` is a core module so we don't need to require it

bin/crew
```diff
@@ -212,17 +217,21 @@ end
...
+      if @pkg.compatibility&.include? 'all' or @pkg.compatibility&.include? ARCH
         if File.exist? CREW_META_PATH + pkgName + '.filelist'
           puts pkgName.lightgreen
         else
           puts pkgName
         end
       end
     else
-      unless @pkg.compatibility.include? 'all' or @pkg.compatibility.include? ARCH
+      unless @pkg.compatibility&.include? 'all' or @pkg.compatibility&.include? ARCH
```
supechicken, line 234: What is the purpose of the `&` symbol?

bin/crew
```diff
@@ -32,7 +32,7 @@ Usage:
   crew reinstall [options] [-k|--keep] [-s|--build-from-source] [-S|--recursive-build] <name> ...
   crew remove [options] <name> ...
   crew search [options] [<name> ...]
-  crew update [options]
+  crew update [options] [<compatible>]
```
uberhacker, line 35: Why do we need this?

bin/crew
```diff
@@ -210,33 +222,59 @@ def list_installed
 end
 
 def list_compatible(compat = true)
+  generate_compatible
```
uberhacker, line 225: Is this necessary every time `crew list compatible` is run?  I'm thinking it's only needed once during `crew update`.

bin/crew
```diff
@@ -334,6 +380,8 @@ def help (pkgName)
     puts "Update crew."
     puts "Usage: crew update"
     puts "This only updates crew itself.  Use `crew upgrade` to update packages."
+    puts "Usage: crew update compatible"
+    puts "This updates the crew package compatibility list."
```
uberhacker, line 384: Why do we need to do this manually?  Shouldn't it be part of the `crew update` process?

bin/crew
satmandu: https://stackoverflow.com/a/64089398

bin/crew
satmandu: Without this you can't test `generate_compatible` without first putting your crew changes up on a git repo!

Essentially it saves a great deal of time in the debug cycle.

bin/crew
```diff
@@ -940,7 +999,7 @@ def expand_dependencies
 end
 
 def resolve_dependencies
-  abort "Package #{@pkg.name} is not compatible with your device architecture (#{ARCH}) :/".lightred unless @pkg.compatibility.include?('all') or @pkg.compatibility.include?(ARCH)
+  abort "Package #{@pkg.name} is not compatible with your device architecture (#{ARCH}) :/".lightred unless @device[:compatible_packages].any? do |elem| elem[:name] == pkgName end
```
uberhacker, line 1002: Faster? `unless @pkg.name.in? @device[:compatible_packages]`

uberhacker: This might not work.  Just a thought.

lib/package.rb
```diff
@@ -45,6 +45,14 @@ def self.get_url (architecture)
     end
   end
 
+  def self.get_binary_url (architecture)
+    if @binary_url.has_key?(architecture)
+      return @binary_url[architecture]
+    else
+      return nil
+    end
```
uberhacker, line 53: `return @binary_url.has_key?(architecture) ? @binary_url[architecture] : null`

bin/crew
satmandu: I refactored `list_compatible` to not duplicate code with `generate_compatible`. Now the read of all package files happens during generate_compatible,which then outputs into the JSON device file. `list_compatible` now just reads the JSON file.

The entire process takes about 3 seconds.
crew update compatible:
```
real    0m1.490s
user    0m1.405s
sys     0m0.085s
```

crew list compatible:
```
real    0m2.072s
user    0m1.968s
sys     0m0.104s
```

One could just read the json file, but in case you have updated local packages, I think it makes sense to spend the extra 1.5 seconds to check to see if those changes fixed anything?

bin/crew
satmandu: It still is part of that process. This just allows one to run it separately without a git pull, which is incredibly useful when you are making local changes to files on a git repo and want to see if there are going to be issues with the package files being read correctly.

If we had had this before, we would have avoided many issues with crew breaking. :)

bin/crew
satmandu: Definitely cleaner!

bin/crew
uberhacker: It takes 3 seconds in your Docker container but what about the many underpowered Chromebooks in existence?

bin/crew
uberhacker: I guess I don't understand why it's important for development.

bin/crew
```diff
@@ -131,15 +131,19 @@ end
...
   print_current_package extra
 end
 
-def print_current_package (extra = false)
+def print_current_package(extra = false)
   pkgName = @pkg.name
   status = ''
   status = 'installed' if @device[:installed_packages].any? do |elem| elem[:name] == pkgName end
-  status = 'incompatible' unless @device[:compatible_packages].any? do |elem| elem[:name] == pkgName end
+  status = 'incompatible' unless @pkg.name.in? @device[:compatible_packages]
```
uberhacker, line 146: I guess just leave the existing logic if this doesn't work?

bin/crew
```diff
@@ -203,40 +215,66 @@ def list_installed
...
-    set_package pkgName, filename
-    if compat
-      if @pkg.compatibility.include? 'all' or @pkg.compatibility.include? ARCH
-        if File.exist? CREW_META_PATH + pkgName + '.filelist'
-          puts pkgName.lightgreen
-        else
-          puts pkgName
-        end
-      end
+    if pkgName.in? @device[:compatible_packages]
```
uberhacker, line 228: Return to `if @pkg.compatibility.include? 'all' or @pkg.compatibility.include? ARCH`?

bin/crew
```diff
@@ -203,40 +215,66 @@ def list_installed
...
+      @compatibility = true
     else
-      unless @pkg.compatibility.include? 'all' or @pkg.compatibility.include? ARCH
-        puts pkgName.lightred
+      @compatibility = false
+    end
+    if @compatibility
```
uberhacker, line 233: Why not put this entire block in the condition above and avoid setting the `@compatibility` variable altogether?

bin/crew
satmandu: The rationale is to just pull the compatibility list from device.json for the `list_compatible` function, otherwise you're just reading all the package files again and that's what `generate_compatible` is for.

bin/crew
satmandu: No longer threading! so removed! :)

bin/crew
satmandu: This entire section got refactored to remove the threading....

satmandu:
I think I am done with this.

bin/crew
```diff
@@ -131,15 +131,19 @@ end
...
+    set_package pkgName, pkgPath
+  rescue => e
+    puts "Error with #{pkgName}.rb: #{e}".red unless e.to_s.include?('uninitialized constant')
+  end
   print_current_package extra
 end
 
-def print_current_package (extra = false)
+def print_current_package(extra = false)
   pkgName = @pkg.name
```
uberhacker, line 143: We can remove this line if you are changing lines 145 and 146 below.

bin/crew
```diff
@@ -490,7 +529,6 @@ def update
   puts 'Package lists, crew, and library updated.'
 
   #update compatible packages
-  puts 'Generating compatible packages...'
   generate_compatible
   puts 'Generating compatible packages done.'.orange if @opt_verbose
```
uberhacker, line 533: Does it make sense to add this message to `generate_compatible` since it's called multiple times?

bin/crew
```diff
@@ -203,39 +215,56 @@ def list_installed
...
-        if File.exist? CREW_META_PATH + pkgName + '.filelist'
-          puts pkgName.lightgreen
-        else
-          puts pkgName
-        end
+    if @device[:compatible_packages].any? do |elem| elem[:name] == pkgName end
+      if ( File.exist? CREW_META_PATH + pkgName + '.filelist' ) && compat
+        puts pkgName.lightgreen
+      else
+        puts pkgName unless !compat
```
uberhacker, line 232: Why not just `puts pkgName if compat`?

bin/crew
satmandu: `generate_compatible` is only called only once during crew operations, unless you are seeing something I'm missing.
once during `crew update`
once during `crew update compatible`
once during `crew list compatible`
once during `crew list incompatible`

uberhacker: Right but my point is this message will only print for `crew update` and not `crew list ...`.

bin/crew
```diff
@@ -725,7 +762,7 @@ def unpack (meta)
       system "ar -p #{meta[:filename]} data.tar.xz | xz -dc#{@verbose} | tar x#{@verbose} -C #{@extract_dir}", exception: true
     when /\.AppImage$/i
       puts "Unpacking 'AppImage' archive, this may take a while..."
-      FileUtils.chmod 0755, meta[:filename], verbose: @fileutils_verbose
+      FileUtils.chmod 0o755, meta[:filename], verbose: @fileutils_verbose
```
uberhacker, line 765: Not sure why the `o` is necessary but whatever...

bin/crew
satmandu: Is that a problem? If you're asking crew to list compatible or incompatible packages, you really can't complain if it then does that...

Let me put it behind a verbose debug flag...

bin/crew
satmandu: rubocop whines about octal... so I'm just adding it everywhere to shut it up.

uberhacker: Gotcha

bin/crew
uberhacker: I just thought it makes sense so the message is displayed whenever it's called.

satmandu: I misunderstood. It **was** being printed whenever it was being called, as I just moved that puts line to generate_compatible.
If you do `crew -v` something that invokes it, it will print that line...

uberhacker: Right but it only prints that line with `crew update -v`; not with `crew list compatible -v`.

bin/crew
```diff
@@ -740,7 +777,7 @@ def unpack (meta)
       # Check the number of directories in the archive
       entries = Dir["#{@extract_dir}/*"]
       entries = Dir[@extract_dir] if entries.empty?
-      if entries.length == 0
+      if entries.length.zero?
```
uberhacker, line 780: Would `if entries.empty?` work also?

satmandu: Yup. Is there any reason to use one over the other?

satmandu: rubocop suggested both, IIRC, and I just used one of them.

I'll be pretty pissed if neither is actually faster than `== 0`

uberhacker: I imagine `.empty?` would be faster than `.length.zero?` by nanoseconds. :)

bin/crew
satmandu: I'm seeing different behavior:
```
chronos@localhost ~/src/satadru/chromebrew (crewrescue)$ crew -v list compatible
Generating compatible packages...
Checking a2png for compatibility.
```

Also
```
crew list compatible -v
Generating compatible packages...
Checking a2png for compatibility.
```

bin/crew
```diff
@@ -263,17 +295,21 @@ def regexp_search(pkgPat)
...
+        puts "Error with #{pkgName}.rb: #{e}".red unless e.to_s.include?('uninitialized constant')
+      end
       if ( @pkg.description =~ /#{pkgPat}/i )
         print_current_package @opt_verbose
         results.push(packageName)
       end
     end
   end
-  abort "Package #{pkgPat} not found. :(".lightred unless results.length > 0
+  abort "Package #{pkgPat} not found. :(".lightred unless results.length.positive?
```
uberhacker, line 309: How about `if results.empty?` instead of `unless results.length.positive?`?

bin/crew
```diff
@@ -557,7 +594,7 @@ def upgrade
       end
     end
 
-    if toBeUpdated.length > 0
+    if toBeUpdated.length.positive?
```
uberhacker, line 597: How about `unless toBeUpdated.empty?` instead of `if toBeUpdated.length.positive?`?

bin/crew
satmandu: done

bin/crew
satmandu: done

bin/crew
uberhacker: Do you see `Generating compatible packages done.` with `crew -v list compatible`?

satmandu: Good catch! Fixed that by moving that inside `generate_compatible`.

PR #5734, opened 2021-05-06
Add mysql package

uberhacker:
MySQL Community Edition is a freely downloadable version of the world's most popular open source database. 
 See https://www.mysql.com/.  This has been a long awaited package and the latest version is small enough to have a reasonably sized install.  Tested on x86_64.


satmandu:
Is there going to be an issue with MySQL just being killed on a reboot instead of cleanly shut down with a shutdown script.. especially since we have no shutdown script functionality?

uberhacker:
Good question.  It really should be shutdown gracefully with `mysql.server stop`.

satmandu:
Maybe you want to have a big red caveat emptor warning about how shutdowns are not likely to be clean and to expect data loss then...

uberhacker:
Or we could add startup/shutdown scripts to do this elegantly.

satmandu:
@uberhacker Are shutdown scripts even possible?  Does putting `trap '' HUP` in a script keep it backgrounded and allow a shutdown script to run?  Can we hook into elogind?

packages/mysql.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  depends_on 'numactl'
+
+  def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/mysql"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share"
+    FileUtils.mv 'man/', "#{CREW_DEST_PREFIX}/share"
+    FileUtils.mv Dir.glob('*'), "#{CREW_DEST_PREFIX}/mysql"
+    Dir.chdir "#{CREW_DEST_PREFIX}/mysql/bin" do
```
supechicken, line 25: You can rubyize it with:
```ruby
Dir["#{CREW_DEST_PREFIX}/mysql/bin/*"].map {|f| f.sub("#{CREW_DEST_DIR}", '').each do |f|
  FileUtils.ln_s f, "#{CREW_DEST_PREFIX}/bin/"
end
```

uberhacker: I get the following error:
```
SyntaxError: /usr/local/lib/crew/packages/mysql.rb:33: syntax error, unexpected `end', expecting '}'
  end
  ^~~
/usr/local/lib/crew/packages/mysql.rb:77: syntax error, unexpected end-of-input, expecting `end'
```
I'm in agreement with the spirit of your suggestion, however.

PR #5735, opened 2021-05-07
Add blender package

uberhacker:
Blender is the free and open source 3D creation suite.  See https://www.blender.org/.  Tested on x86_64.


satmandu:
Too bad we're not building from source.

uberhacker:
You're welcome to try.  It has a ton of dependencies.

PR #5745, opened 2021-05-08
Update codium from 1.54.3 to 1.55.2

uberhacker:


satmandu:
1.56 should have the new electron with Wayland support!

PR #5755, opened 2021-05-08
Symbols are not quoted

uberhacker:
I've allowed these through before.  Easy to miss.


satmandu:
Aside from style, this is all still valid ruby, right?

uberhacker:
With the quotes, it makes it a string, no?

PR #5767, opened 2021-05-12
Fix gcc11 issues, and upgrade issues with gcc10, modify preflight

satmandu:
- Fix GCC11 & GCC10 packages (issue with dangling symbolic links being created and incorrect ruby `.partition` usage.)
- moved `@pkg.preflight` to `resolve_dependencies_and_install` which is processed before package downloads in `install`, avoiding a download if the logic dictates the package is not compatible.

image

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l


uberhacker:
I feel like we're creating a problem for an existing solution.  This is essentially the same as preflight.  Not sure why we need a separate function.

satmandu:
|   |  Build | Install  | 
|---|---|---|
| preflight  |  Can Block | Can Block  |  
| preinstallcheck  | nop  | Can Block  |

packages/gcc10.rb
```diff
@@ -82,7 +86,10 @@ def self.preinstallcheck
              installed_gccver.to_s == 'bash:' ||
              installed_gccver.to_s == @gcc_version.to_s ||
              installed_gccver.partition('.')[0].to_s == @gcc_version.partition('.')[0].to_s
-        abort "GCC version #{installed_gccver} is already installed. Please run: crew remove gcc#{installed_gccver} && crew install #{self.to_s.downcase}".lightgreen
+        $stderr.puts "GCC version #{installed_gccver} is currently installed.".lightred
+        $stderr.puts "To use #{self.to_s.downcase} please run:".lightgreen
+        $stderr.puts "crew remove gcc#{installed_gccver} && crew install #{self.to_s.downcase}".lightgreen
+        exit 1
```
uberhacker, line 92: This is still too late in the process.  The warning should happen before the download.

satmandu: Is it not happening before the process?

Inside `crew` this is what is happening:
```
def install
  if !@pkg.in_upgrade && @device[:installed_packages].any? { |pkg| pkg[:name] == @pkg.name }
    return
  end

  unless @pkg.is_fake?
    # perform pre_install_check process to see if package should even
    # be downloaded.
    pre_install_check CREW_DEST_DIR
  end

  unless @pkg.is_fake?
    meta = download
    target_dir = unpack meta
    if meta[:source] == true
```

`meta = download` is where downloads happen, right? I put `pre_install_check` before that point.

uberhacker: You could do the same exact thing in preflight without needing to add the exit 1.

satmandu: The ` $stderr.puts + exit 1` is there instead of using `abort` which equals `warn + exit 1`  as I understand it. I wanted the multiline output.

(Also rejiggering this to use preflight...)

satmandu: done with moving preflight around...

satmandu:
The ordering of the install process seems correct:
image

PR #5772, opened 2021-05-13
Handle edge case of cachefile already in CREW_BREW_DIR when build restarts

satmandu:
- if a build is interrupted, a cachefile may already be in CREW_BREW_DIR. This avoids an error in that case.

Works properly:
- [x] x86_64


uberhacker:
Good call.  Stale cache is not fun.

PR #5775, opened 2021-05-14
Assume binaries are already stripped during install

uberhacker:
This speeds up installs significantly.  No reason to strip binaries.


satmandu:
Binaries in packages were not previously stripped during packaging, so this will only work for packages created in the last month, but sure.

uberhacker:
It really should be up to the packager to determine if stripping is necessary.  Some packages break after stripping so we shouldn't assume everything needs stripped.

satmandu:
Maybe we make stripping default with an opt-out?

Maybe via package-specific opt-out options?

uberhacker:
I thought it already was the default for builds and the opt-out would be to set `CREW_NOT_STRIP`?

satmandu:
I thought that was only for not stripping at install. I think I just added stripping at build, but I could be mistaken...

PR #5801, opened 2021-05-20
Add upx and ucl

satmandu:
- Adds upx (and dependency ucl) 
- This may be one option for reducing binary size @saltedcoffii 
- older git tags and release version had issues with broken generated binaries that this fixes.

Builds properly:
- [x] x86_64
- [x] i686
- [x] armv7l


uberhacker:
Just what we need...more compression options as if we aren't already confused enough. :)

PR #5815, opened 2021-05-27
Update libidn2 from 2.3.0 to 2.3.1

uberhacker:
Tested on all architectures.


satmandu:
Why disable static? Isn't that useful in case one needs to statically link?

_Stares in `curl`..._

uberhacker:
Why is static linking necessary?  Shared linking should still work fine.

satmandu:
I was only asking in the case that someone wants to link statically...

PR #5823, opened 2021-05-29
Fix crew list available and crew list compatible

### LINKED ISSUE #5821: crew list available|compatible lists all packages including incompatible
I see red files with both `crew list available` and `crew list compatible`.  The expectation is both commands should show only packages that can be installed.

uberhacker:
Fixes #5821.


satmandu:
Thanks for fixing this!

PR #5837, opened 2021-06-04
#2. Convert curl, git, pixz to static, and update dependencies...

satmandu:
- Change `git`, `curl`, and `pixz` to static apps
  - This should allow for massively streamlining `EARLY_PACKAGES` in `install.sh` as well as keep dependency changes for these packages from screwing up installs. (Getting ruby to have fewer dependencies is next on my list.)
- Add `rdfind` and a crew option to use it (with `symlinks`) to eliminate duplicate files during archiving.
  - Accordingly added `rdfind` and `symlinks` to `buildessential`
- Update dependencies for git, curl, and pixz to enable static library generation.
- Switch `krb5` to use `e2fsprogs` as a dependency. (and thus add `e2fsprogs` to core)
- `curl` and `libcurl` are now two separate packages. Note that there is also a rebuilt `openssl` which you need to properly link to `libcurl`...
- Updated icu to 69. Packages which may need an update (in progress) include js78, evolution_data_server, webkit2gtk_4, webkit2gtk_5).

- Using `upx` to compress application binaries (butnot libraries) at the moment, as there appears to be library compression issues at the moment. (But this saves a LOAD of space for packages like git.)

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/curl.rb
+++ b/packages/curl.rb
@@ -24,6 +24,7 @@ class Curl < Package
   })
 
   depends_on 'ca_certificates' => :build
+  depends_on 'libcurl' # L (This is only here to make sure upgrades of curl pull in libcurl.)
   depends_on 'libunbound' => :build
   depends_on 'musl' => :build
   depends_on 'py3_pip' => :build
--- a/packages/curl.rb
+++ b/packages/curl.rb
@@ -229,7 +229,7 @@ class Curl < Package
 
   def self.check
     system 'src/curl -Lf https://github.com/skycocker/chromebrew/raw/master/install.sh -o /tmp/install.sh'
-    FileUtil.rm '/tmp/install.sh'
+    FileUtils.rm '/tmp/install.sh'
   end
 
   def self.install
--- a/packages/curl.rb
+++ b/packages/curl.rb
@@ -3,99 +3,236 @@ require 'package'
 class Curl < Package
   description 'Command line tool and library for transferring data with URLs.'
   homepage 'https://curl.se/'
-  @_ver = '7.77.0'
+  @_ver = '7.77.0-git'
   version @_ver
   license 'curl'
   compatibility 'all'
-  source_url "https://curl.se/download/curl-#{@_ver}.tar.xz"
-  source_sha256 '0f64582c54282f31c0de9f0a1a596b182776bd4df9a4c4a2a41bbeb54f62594b'
+  source_url 'https://github.com/curl/curl.git'
+  git_hashtag 'curl-7_77_0'
 
   binary_url({
-    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0_armv7l/curl-7.77.0-chromeos-armv7l.tpxz',
-     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0_armv7l/curl-7.77.0-chromeos-armv7l.tpxz',
-       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0_i686/curl-7.77.0-chromeos-i686.tpxz',
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0_x86_64/curl-7.77.0-chromeos-x86_64.tpxz'
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0-git_armv7l/curl-7.77.0-git-chromeos-armv7l.tpxz',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0-git_armv7l/curl-7.77.0-git-chromeos-armv7l.tpxz',
+       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0-git_i686/curl-7.77.0-git-chromeos-i686.tpxz',
+     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0-git_x86_64/curl-7.77.0-git-chromeos-x86_64.tpxz'
   })
   binary_sha256({
-    aarch64: '0273d3cf0ffb3d18777bb4d5da6a20c00970ad3787687e48c023f406aa42d4bc',
-     armv7l: '0273d3cf0ffb3d18777bb4d5da6a20c00970ad3787687e48c023f406aa42d4bc',
-       i686: 'ead35292354673e20806e70a9359306bd63758a04db3888eb5eca7c99cd1904a',
-     x86_64: '81b8641d60b222248a75f910e2bf82ce7b9d793a84558466e7670215bbd9b8d1'
+    aarch64: '268771c547ef80761b937aedafe786ca701524e20b4737263aee9791514189d0',
+     armv7l: '268771c547ef80761b937aedafe786ca701524e20b4737263aee9791514189d0',
+       i686: '45796384ef70732ab7d4df71f09d2860e22bee776f86f7ff60c9863ff5bc28c2',
+     x86_64: 'ee99c0558b9dc9e64e4a2619706224daa3e7a0d4b100af135ac5b0a6a3b68ab4'
   })
 
-  depends_on 'libunbound' # ?
-  depends_on 'openldap' # R
-  depends_on 'openssl' # R
+  depends_on 'ca_certificates' => :build
+  depends_on 'libunbound' => :build
+  depends_on 'musl' => :build
   depends_on 'py3_pip' => :build
   depends_on 'rust' => :build
   depends_on 'valgrind' => :build
-  depends_on 'zlibpkg' # R
-  depends_on 'zstd' # R
 
   def self.build
     raise StandardError, 'Please remove libiconv before building.' if File.exist?("#{CREW_LIB_PREFIX}/libcharset.so")
 
-    @krb5_static_libs = "#{CREW_LIB_PREFIX}/libkrb5support.a " + "#{CREW_LIB_PREFIX}/libgssapi_krb5.a " + "#{CREW_LIB_PREFIX}/libkrb5.a " + "#{CREW_LIB_PREFIX}/libk5crypto.a " + "#{CREW_LIB_PREFIX}/libcom_err.a "
+    @krb5_static_libs = '-l:libkrb5support.a -l:libgssapi_krb5.a -l:libkrb5.a -l:libk5crypto.a -l:libcom_err.a'
 
-    FileUtils.mkdir 'ssl'
-    FileUtils.ln_s "#{CREW_PREFIX}/include/openssl", 'ssl/include'
-    FileUtils.ln_s CREW_LIB_PREFIX, 'ssl/lib'
-    @sslpath = "#{`pwd`.chomp}/ssl"
+    FileUtils.mkdir_p 'build'
+    FileUtils.mkdir_p 'deproot/include'
+    FileUtils.mkdir_p 'deproot/lib'
+    Dir.chdir 'deproot' do
+      FileUtils.ln_s 'lib', 'lib64'
+    end
+    @deppath = "#{`pwd`.chomp}/deproot"
+    FileUtils.ln_s "#{CREW_PREFIX}/include/linux", "#{@deppath}/include/"
+    FileUtils.ln_s "#{CREW_PREFIX}/include/asm", "#{@deppath}/include/"
+    FileUtils.ln_s "#{CREW_PREFIX}/include/asm-generic", "#{@deppath}/include/"
+
+    @arch_ssp_cflags = ''
+    case ARCH
+    when 'i686'
+      @arch_ssp_cflags = '-fno-stack-protector'
+    end
+
+    @curldep_cmake_options = "CC=musl-gcc \
+          CFLAGS='-flto -pipe -O3 -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
+          CXXFLAGS='-flto -pipe -O3 -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
+          CPPFLAGS='-I#{@deppath}/include' \
+          LDFLAGS='-L#{@deppath}/lib' \
+          cmake \
+          -DCMAKE_INSTALL_PREFIX='#{@deppath}' \
+          -DCMAKE_LIBRARY_PATH='#{@deppath}/lib' \
+          -DCMAKE_C_COMPILER=`which musl-gcc` \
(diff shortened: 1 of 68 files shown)


satmandu:
I will have a dedicated PR just for the curl, git, pixz upgrade, since that needs to happen first.

satmandu:
Method of testing: 
```
CREW_TESTING_BRANCH=initial_static CREW_TESTING=1 crew update
crew upgrade
CREW_TESTING_BRANCH=static CREW_TESTING=1 crew update
crew upgrade
```

satmandu:
@uberhacker Ready for review.

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
+    case ARCH
+    when 'i686'
+      @arch_ssp_cflags = '-fno-stack-protector'
+    end
+
+    @curldep_cmake_options = "CC=musl-gcc \
+          CFLAGS='-flto -pipe -O3 -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
+          CXXFLAGS='-flto -pipe -O3 -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
+          CPPFLAGS='-I#{@deppath}/include' \
+          LDFLAGS='-L#{@deppath}/lib' \
```
uberhacker, line 60: What's wrong with just using `#{@deppath}/#{ARCH_LIB}`?  Then you don't need any crazy symlink madness.

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
+          cmake \
+          -DCMAKE_INSTALL_PREFIX='#{@deppath}' \
+          -DCMAKE_LIBRARY_PATH='#{@deppath}/lib' \
```
uberhacker, line 63: What's wrong with just using `#{@deppath}/#{ARCH_LIB}`?

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
+          -DCMAKE_STATIC_LINKER_FLAGS='-flto' \
+          -DCMAKE_MODULE_LINKER_FLAGS='-flto' \
+          -DPROPERTY_INTERPROCEDURAL_OPTIMIZATION=TRUE \
+          -DCMAKE_BUILD_TYPE=Release"
+
+    @curldep_env_options = "CC=musl-gcc \
+        CFLAGS='-flto -pipe -O3 -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
+        CXXFLAGS='-flto -pipe -O3 -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans#{@arch_ssp_cflags}' \
+        CPPFLAGS='-I#{@deppath}/include' \
+        LDFLAGS='-L#{@deppath}/lib'"
```
uberhacker, line 79: `-L#{@deppath}/#{ARCH_LIB}`?

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
+        -DZSTD_BUILD_SHARED=OFF \
+        ../ -G Ninja"
+      end
+      system 'ninja -C build/cmake/builddir'
+      system 'ninja -C build/cmake/builddir install'
+    end
+    FileUtils.mkdir_p 'build/brotli'
+    Dir.chdir 'build/brotli' do
+      puts 'Building Brotli.'.lightgreen
+      system 'curl -Ls https://github.com/google/brotli/archive/v1.0.9.tar.gz | tar --strip-components=1 -zxf -'
```
uberhacker, line 107: Why not use the `brotli` package instead?

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
+        @brotlilibs = %w[libbrotlidec libbrotlienc libbrotlicommon]
+        @brotlilibs.each do |lib|
+          FileUtils.ln_s "#{lib}-static.a", "#{lib}.a"
+        end
+      end
+    end
+    FileUtils.mkdir_p 'build/idn2'
+    Dir.chdir 'build/idn2' do
+      puts 'Building IDN2.'.lightgreen
+      system 'curl -Ls http://ftp.gnu.org/gnu/libidn/libidn2-2.3.1.tar.gz | tar --strip-components=1 -zxf -'
```
uberhacker, line 127: Why not use the `libidn2` package instead?

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
+      @arch_cxx_flags = ''
+      @openssl_configure_target = 'linux-x86'
+    when 'x86_64'
+      @arch_c_flags = ''
+      @arch_cxx_flags = ''
+      @openssl_configure_target = 'linux-x86_64'
+    end
+    Dir.chdir 'build/ssl' do
+      puts 'Building OpenSSL.'.lightgreen
+      system 'curl -Ls https://www.openssl.org/source/openssl-1.1.1k.tar.gz | tar --strip-components=1 -zxf -'
```
uberhacker, line 150: Why do we need to build openssl?  Why not just reference the package instead?

packages/curl.rb
satmandu: This way it is the same across all architectures. It's only used during build anyways, since these are only built to create the static libraries.

The symlink is needed in any case, because some of these packages are happy to use lib, and some want to use lib64 anyways, but at least this way you don't have to police them to do the right thing.

(same applies to the other similar comments. :) )

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
+      #-DHAVE_GLOB=0 \
+      # ../ -G Ninja"
+      # end
+      # system 'ninja -C builddir'
+      # system 'ninja -C builddir install'
+    # end
+    FileUtils.mkdir_p 'build/nghttp2'
+    Dir.chdir 'build/nghttp2' do
+      puts 'Building Nghttp2.'.lightgreen
+      system 'curl -Ls https://github.com/nghttp2/nghttp2/releases/download/v1.43.0/nghttp2-1.43.0.tar.gz | tar --strip-components=1 -zxf -'
```
uberhacker, line 185: Why not use the `libnghttp2` package instead?

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
-      CXXFLAGS='-flto -pipe -O3' \
-      LDFLAGS='-flto -static -Wl,--no-as-needed -ldl' LIBS='-l:libresolv.a #{@krb5_static_libs} -l:libzstd.a -l:libssh.a -lm -lbrotlicommon -lbrotlidec -lz -lssl -lcrypto -lsasl2 -lxml2 -lpthread' \
-      ./configure #{CREW_OPTIONS} \
+    system "env CC=musl-gcc \
+      PKG_CONFIG='pkg-config --static' \
+      CPPFLAGS='-I#{@deppath}/include' \
+      CFLAGS='-flto -pipe -O3 -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
+      CXXFLAGS='-flto -pipe -O3 -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
+      LDFLAGS='-L#{@deppath}/lib -flto -static -Wl,--no-as-needed -ldl' \
+      LIBS='#{@deppath}/lib/libssl.a #{@deppath}/lib/libcrypto.a #{@deppath}/lib/libbrotlicommon.a #{@deppath}/lib/libbrotlidec.a #{@deppath}/lib/libidn2.a #{@deppath}/lib/libnghttp2.a  #{@deppath}/lib/libz.a #{@deppath}/lib/libzstd.a -L#{@deppath}/lib' \
```
uberhacker, line 203: Replace `/lib/` with `/#{ARCH_LIB}/` instead?

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
+      CURL_LIBRARY_PATH=#{@deppath}/lib \
```
uberhacker, line 204: `#{@deppath}/#{ARCH_LIB}`?

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
-   def self.check
-     # Python package impacket needed for testing.
-     # 1094 tests out of 1097 reported OK: 99% on 10/25/2020
-     # The 3 tests that failed were FTP, SMB and GOPHER.
-     system 'pip3 install impacket'
-     system 'make check || true'
-     system 'pip3 uninstall -y impacket'
-   end
+  def self.check
+    system 'src/curl -Lf https://github.com/skycocker/chromebrew/raw/master/install.sh -o /tmp/install.sh'
```
uberhacker, line 232: Remove `/tmp/install.sh` after testing?

packages/curl.rb
satmandu: These are built with musl. Static libraries built with gcc or clang appear to sometimes still invoke glibc 2.2x, and that's creating subtle libnss crashes during dns lookups with curl. Compiling these separately with musl avoids that problem entirely.

Also, curl no longer segfaults.

See https://github.com/curl/curl/issues/7194 for the issue I opened before I gave up on compiling with the chromebrew static libraries.

packages/curl.rb
satmandu: Ah yes, I should definitely do that.

packages/libidn2.rb
```diff
@@ -9,17 +9,17 @@ class Libidn2 < Package
...
-  binary_sha256 ({
-    aarch64: '9533e091572a2fb3aff874794327ffb752967b9952608cbccbd06321a1098323',
-     armv7l: '9533e091572a2fb3aff874794327ffb752967b9952608cbccbd06321a1098323',
-       i686: 'f65ae2734d93e883e32a7602bee7e494d396406addd38d675346d41c5eb34943',
-     x86_64: '3a54e6d6c60865885257e3f1383a1b1d957483a97ef5c0ff400611f58908f28e',
+  binary_sha256({
+    aarch64: 'd5451f73189a53bbdd500b23c781750d688d9b8ee4763eecca8323eace221566',
+     armv7l: 'd5451f73189a53bbdd500b23c781750d688d9b8ee4763eecca8323eace221566',
+       i686: '03df0cdc3848415a7e08e2026eab5a3a50c6943c1980223dac8d0579cdbdc3be',
+     x86_64: '2ade8b8171841b7a911b1802d2dbee0f890537687131e16515a4155c9167187a'
```
uberhacker, line 22: Is this just because the original sha256sums were wrong?  Normally, we would need to bump the version if that wasn't the case.

packages/libssh.rb
```diff
@@ -4,46 +4,51 @@ class Libssh < Package
   description 'libssh is a multiplatform C library implementing the SSHv2 and SSHv1 protocol on client and server side.'
   homepage 'https://www.libssh.org/'
   @_ver = '0.9.5'
-  version "#{@_ver}-1"
+  version "#{@_ver}-2"
   license 'LGPL-2.1'
-  compatibility 'all'
   @_ver_prelastdot = @_ver.rpartition('.')[0]
+  compatibility 'all'
```
uberhacker, line 10: Can we put license right above compatibility so the version lines are together?

packages/libssh.rb
```diff
@@ -4,46 +4,51 @@ class Libssh < Package
...
+      -DWITH_STATIC_LIB=ON \
+      ../ -G Ninja"
+    end
+    system 'ninja -C builddir'
+    Dir.chdir('builddir') do
+      FileUtils.cp 'src/libssh.a', '../' if File.exist?('src/libssh.a')
+      system "cmake #{CREW_CMAKE_OPTIONS} \
+      -DWITH_EXAMPLES=OFF \
+      -DBUILD_SHARED_LIBS=ON \
+      -DWITH_STATIC_LIB=OFF \
```
uberhacker, line 44: No static library?

tools/core_packages.txt
```diff
@@ -7,6 +7,7 @@ c_ares
 ca_certificates
 crew_profile_base
 curl
+e2fsprogs
```
uberhacker, line 10: Can we add libunbound also or is it not needed now that curl is a fully static build?

packages/curl.rb
uberhacker: Why don't we build the crew packages with musl?  Seems like a maintenance nightmare to do it this way.

packages/libssh.rb
satmandu: See the `FileUtils.cp 'src/libssh.a', '../' if File.exist?('src/libssh.a')` bit there?

packages/curl.rb
uberhacker: The whole idea of `#{ARCH_LIB}` is to abstract the library so you don't have to police. :)

packages/libssh.rb
uberhacker: You wouldn't have to do any of that if you enabled static, no?

tools/core_packages.txt
satmandu: I don't know the rationale for putting it there in the first place, but if the reason was "because curl needs it" then use it can be removed...

satmandu: libunbound is not a runtime dependency of curl now that curl is static.

uberhacker: Please do a fresh install and compare the base packages with core_packages.txt.  The way I do it is with vimdiff.
```
cd /path/to/tools
crew list installed -d > installed.txt
vimdiff installed.txt core_packages.txt
```
Adjust any differences.

packages/curl.rb
satmandu: Musl is ONLY being used for the static curl dependencies, because it's not breaking openssl in the way that glibc 2.27 appears to be doing.

Even libcurl (and static git, which uses it) are being built against glibc.

This is probably a decent explainer on why not to build everything against musl: https://www.madetech.com/blog/a-tale-in-adopting-alpine-linux-for-docker-problems-we-faced-with-rspec-testing/

(Also, if we decide to build everything against a newer libc, we should probably just use a newer glibc...)

packages/libidn2.rb
satmandu: This is a rebuild I'll bump the version.

packages/curl.rb
satmandu: Let's see if this works. :)

satmandu: Works!

packages/libidn2.rb
satmandu: adjusted

packages/curl.rb
satmandu: We need to link to openssl static libs for https to work. The segfault I was getting was from accessing https URLs.

Building all of these dependencies separately, and with musl, ensured no dependence upon glibc.

It's really hard to debug curl issues unless you go dependency by dependency, and doing everything with musl not only took glibc out of the equation, but made debugging this issue a hell of a lot easier.

Curl is by no means the longest build of any package anyways, even with these dependency builds.

PR #5843, opened 2021-06-07
Fix segfault issue in static curl, add libcurl to break out shared libraries, update git (built with libcurl)

satmandu:
- Dramatically lowers the size of the curl package for install.
- Also used upx to further shrink the binary.
- Also updated git and built with this libcurl.
- Git is now shrunk with upx too, reducing the file size to ~ 300Mb when extracted.)

(These are the same versions as in https://github.com/skycocker/chromebrew/pull/5837 )

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/curl.rb
+++ b/packages/curl.rb
@@ -3,63 +3,212 @@ require 'package'
 class Curl < Package
   description 'Command line tool and library for transferring data with URLs.'
   homepage 'https://curl.se/'
-  @_ver = '7.77.0'
+  @_ver = '7.77.0-git'
   version @_ver
   license 'curl'
   compatibility 'all'
-  source_url "https://curl.se/download/curl-#{@_ver}.tar.xz"
-  source_sha256 '0f64582c54282f31c0de9f0a1a596b182776bd4df9a4c4a2a41bbeb54f62594b'
+  source_url 'https://github.com/curl/curl.git'
+  git_hashtag 'curl-7_77_0'
 
   binary_url({
-    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0_armv7l/curl-7.77.0-chromeos-armv7l.tpxz',
-     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0_armv7l/curl-7.77.0-chromeos-armv7l.tpxz',
-       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0_i686/curl-7.77.0-chromeos-i686.tpxz',
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0_x86_64/curl-7.77.0-chromeos-x86_64.tpxz'
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0-git_armv7l/curl-7.77.0-git-chromeos-armv7l.tpxz',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/curl/7.77.0-git_armv7l/curl-7.77.0-git-chromeos-armv7l.tpxz',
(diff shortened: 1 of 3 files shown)


packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
+    @curldep_env_options = "CC=musl-gcc \
+        CFLAGS='-flto -pipe -O3 -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
+        CXXFLAGS='-flto -pipe -O3 -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans#{@arch_ssp_cflags}' \
+        CPPFLAGS='-I#{@deppath}/include' \
+        LDFLAGS='-L#{@deppath}/#{ARCH_LIB}'"
+
+    FileUtils.mkdir_p 'build/zlib'
+    Dir.chdir 'build/zlib' do
+      puts 'Building Zlib.'.lightgreen
+      system 'curl -Ls http://zlib.net/zlib-1.2.11.tar.gz | tar --strip-components=1 -zxf -'
```
uberhacker, line 84: Should we checksum these downloads to have an integrity check?

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
...
+
+    @curldep_env_options = "CC=musl-gcc \
+        CFLAGS='-flto -pipe -O3 -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
+        CXXFLAGS='-flto -pipe -O3 -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans#{@arch_ssp_cflags}' \
+        CPPFLAGS='-I#{@deppath}/include' \
+        LDFLAGS='-L#{@deppath}/#{ARCH_LIB}'"
+
+    FileUtils.mkdir_p 'build/zlib'
+    Dir.chdir 'build/zlib' do
+      puts 'Building Zlib.'.lightgreen
```
uberhacker, line 83: The color green typically indicates success.  Would yellow be better for a status message?

packages/curl.rb
```diff
@@ -3,108 +3,237 @@
 class Curl < Package
   description 'Command line tool and library for transferring data with URLs.'
   homepage 'https://curl.se/'
-  @_ver = '7.77.0'
+  @_ver = '7.77.0-git'
```
uberhacker, line 6: Should we remove the curl update from #5837?

packages/git.rb
```diff
@@ -3,33 +3,35 @@
 class Git < Package
   description 'Git is a free and open source distributed version control system designed to handle everything from small to very large projects with speed and efficiency.'
   homepage 'https://git-scm.com/'
-  @_ver = '2.31.1'
-  version "#{@_ver}-1"
+  @_ver = '2.32.0'
```
uberhacker, line 6: Should we remove the git update from #5837?

packages/curl.rb
satmandu: It's literally the same package file in both PRs. And they're static. So no change is needed.

packages/git.rb
satmandu: It's literally the same package file in both PRs. And they're static. So no change is needed.

packages/curl.rb
satmandu: Sure. Let me add those with hashpipe.

satmandu: Let me add those with hashpipe.

packages/curl.rb
uberhacker: I guess my question is why do we need to update the packages in both PRs?

packages/curl.rb
satmandu: yes

packages/curl.rb
satmandu: This is only in case you want more testing of the other larger PR. We don't want to leave curl partially broken while that is happening.

PR #5844, opened 2021-06-07
Update openldap from 2.4.57 to 2.5.5

### LINKED ISSUE #5840: openldap and libcurl version mismatch?
`libcurl` seems to be looking for `libldap-2.5`, when `openldap` is providing `libldap-2.4`.

```bash
$ ls /usr/local/lib64/libldap-*
/usr/local/lib64/libldap-2.4.so.2
/usr/local/lib64/libldap-2.4.so.2.11.5

$ ldd /usr/local/lib64/libcurl.so.4 | grep ldap
      libldap-2.5.so.0 => not found
```

edit: same is true of `liblber`, which is in `openldap`

edit: looks like the stable version ([`openldap` 2.5.4](https://git.openldap.org/openldap/openldap/-/tags/OPENLDAP_REL_ENG_2_5_4)) released about a month ago, so we could potentially update the `openldap` package?

uberhacker:
Fixes #5840.  Tested on all architectures.

### DIFF
--- a/packages/openldap.rb
+++ b/packages/openldap.rb
@@ -3,27 +3,31 @@ require 'package'
 class Openldap < Package
   description 'OpenLDAP Software is an open source implementation of the Lightweight Directory Access Protocol.'
   homepage 'https://www.openldap.org/'
-  @_ver = '2.4.57'
+  @_ver = '2.5.5'
   version @_ver
   license 'OpenLDAP and GPL-2'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


satmandu:
Unless you built this with the new krb5/e2fsprogs setup we need to rebuild this, since it needs com_err, and krb5 now uses the libcom_err.so.2 from e2fsprogs...

satmandu:
I'll add the rebuild to my PR...

PR #5848, opened 2021-06-07
install.sh: Put pixz, git, curl at the front of EARLY_PACKAGES

satmandu:
- Put pixz, git, curl at the front of the package install section in `install.sh`.

Works properly:
- [x] x86_64

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -903,6 +903,9 @@ def prepare_package(destdir)
 
     strip_dir destdir 
 
+    # make hard linked files symlinks and use upx on executables
+    shrink_dir destdir
+
   end
 end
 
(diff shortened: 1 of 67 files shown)


uberhacker:
Can we remove some of the curl dependencies that are now included in the static build?

satmandu:
> Can we remove some of the curl dependencies that are now included in the static build?

Yes, but I wanted to leave that for refactoring, as we really don't want to accidentally remove something which ruby needs.

At some point we should just bisect these dependencies...

uberhacker:
Can we bisect the dependencies in this PR?  Shouldn't be too hard, right?

PR #5859, opened 2021-06-10
Install.sh cleanup & simplification, using jq

satmandu:
- This works much faster for me.
- Added a line for container armv8l usage so I dogfood this script more. :)
- still broken on i686, but that's due to git being broken on there.

Works properly:
- [x] x86_64 (M92 Beta container)
- [x] armv7l

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -122,12 +122,14 @@ done
 # functions to maintain packages
 function download_check () {
     cd "$CREW_BREW_DIR"
-    if [[ -n "$CREW_CACHE_ENABLED" ]] && [[ -f "$CREW_CACHE_DIR/${3}" ]] ; then
+    # use cached file if available and caching enabled
+    if [ -n "$CREW_CACHE_ENABLED" ] && [[ -f "$CREW_CACHE_DIR/${3}" ]] ; then
       echo -e "${BLUE}Verifying cached ${1}...${RESET}"
       echo -e "${GREEN}$(echo "${4}" "$CREW_CACHE_DIR/${3}" | sha256sum -c -)${RESET}"
       case "${?}" in
       0)
-        ln -sf "$CREW_CACHE_DIR/${3}" "$CREW_BREW_DIR/${3}" && return
+        ln -sf "$CREW_CACHE_DIR/${3}" "$CREW_BREW_DIR/${3}" || true
+        return
         ;;
       *)
         echo -e "${RED}Verification of cached ${1} failed, downloading.${RESET}"
@@ -141,7 +143,11 @@ function download_check () {
     echo -e "${BLUE}Verifying ${1}...${RESET}"
     echo -e "${GREEN}$(echo "${4}" "${3}" | sha256sum -c -)${RESET}"
     case "${?}" in
-    0) [[ -n "$CREW_CACHE_ENABLED" ]] && cp "${3}" "$CREW_CACHE_DIR/${3}"
+    0)
+      if [ -n "$CREW_CACHE_ENABLED" ] ; then
+        cp "${3}" "$CREW_CACHE_DIR/${3}" || true
+      fi
+      return
       ;;
     *)
       echo -e "${RED}Verification failed, something may be wrong with the download.${RESET}"
--- a/install.sh
+++ b/install.sh
@@ -11,22 +11,29 @@ URL="https://raw.githubusercontent.com/${OWNER}/${REPO}/${BRANCH}"
 CREW_PREFIX="${CREW_PREFIX:-/usr/local}"
 CREW_LIB_PATH="${CREW_PREFIX}/lib/crew"
 CREW_CONFIG_PATH="${CREW_PREFIX}/etc/crew"
-CREW_BREW_DIR="${CREW_PREFIX}/tmp/crew/"
+CREW_BREW_DIR="${CREW_PREFIX}/tmp/crew"
 CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
 CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 CURL="${CURL:-curl}"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -198,6 +187,7 @@ ln -sfv "../lib/crew/bin/crew" "${CREW_PREFIX}/bin/"
 # prepare sparse checkout .rb packages directory and do it
 cd "${CREW_LIB_PATH}"
 rm -rf .git
+git config --global init.defaultBranch main
```
uberhacker, line 190: Wouldn't the default branch be master?

install.sh
```diff
@@ -91,8 +88,22 @@ case "${ARCH}" in
...
+    '. | .[$key0]=$value0 | .[$key1]=[]' <<<'{}' > device.json
+fi
+
+echo -e "${YELLOW}Downloading information for Bootstrap packages.${BLUE}"
+
+(cd "${CREW_LIB_PATH}"/packages && curl -#OL "${URL}"/packages/{"${BOOTSTRAP_PACKAGES// /,}"}.rb)
+echo -e "${RESET}"
+
+for package in $BOOTSTRAP_PACKAGES; do
+  pkgfile="$(cat "${CREW_LIB_PATH}"/packages/"$package".rb)"
```
uberhacker, line 106: Do nested double quotes work or do you need to escape?

uberhacker: I guess it would upon further review.

install.sh
```diff
@@ -208,9 +198,14 @@ echo crew >> .git/info/sparse-checkout
 echo tools   >> .git/info/sparse-checkout
 git fetch origin "${BRANCH}"
 git reset --hard origin/"${BRANCH}"
+echo -e "${YELLOW}Updating crew package information.${RESET}"
 crew update
 
-yes | crew install $LATE_PACKAGES 
+crew_installed_packages=$(jq  '.installed_packages[] .name' "${CREW_CONFIG_PATH}/device.json" | sed -e 's/"//g'   | sort)
+crew_core_packages=$(cat "${CREW_LIB_PATH}/tools/core_packages.txt" | sort)
```
uberhacker, line 205: If we do this and core packages are updated, we need to be watchful to update core_packages.txt as needed.

install.sh
```diff
@@ -208,9 +198,14 @@ echo crew >> .git/info/sparse-checkout
...
+# The comm tool gives the set of packages in core not already installed.
+# Install those remaining core packages.
+yes | crew install $(comm -23 <(echo $crew_core_packages) <(echo $crew_installed_packages))
```
uberhacker, line 208: Very interesting.  The only drawback I can see of doing it this way is if order matters during install.

install.sh
satmandu: This gets rid of the warning message, and we're pulling from master. The local branch name can be anything. (Do we hard code that local branch name anywhere?)

install.sh
satmandu: Yes. We should update logic in crew's `crew update` for this.

install.sh
satmandu: Crew handles the order anyways though... And we're not breaking crew, git, or curl any more with ordering...

install.sh
uberhacker: Doing this would require us to execute `git pull origin master` rather than just `git pull`, correct?

satmandu: changed git init to clone so this line is no longer needed.

satmandu:
Added cache functionality.
This now finishes in 2 minutes 5 seconds in a container, and in 45 seconds when run on an existing install.

uberhacker:
I'll wait until you remove the Draft label to add my review.

install.sh
```diff
@@ -49,12 +51,15 @@ esac
 
 # install a base set of development packages for compiling and building software
 echo -e -n "${BLUE}Install development tools? (not needed for most users) [y/N]: ${RESET}"; read -n1 devtools
-
+echo -e "\n\n${YELLOW}Doing initial setup for install in ${CREW_PREFIX}.${RESET}"
+echo -e "${YELLOW}This may take a while if there are preexisting files in ${CREW_PREFIX}...${RESET}"
```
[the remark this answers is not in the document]
satmandu: Status quo is not aborting though...

I see the argument for aborting, but we're not deleting files, and this is much safer than a year ago. 😅

install.sh
```diff
@@ -191,28 +204,59 @@ done
...
+rm -rf "${CREW_LIB_PATH}"/{.git,.crewsetuptmp}
+
+# Do a minimal clone, which also sets origin to the master/main branch
+# by default. For more on why this setup might be useful see:
+# https://github.blog/2020-01-17-bring-your-monorepo-down-to-size-with-sparse-checkout/
+git clone --depth=1 --filter=blob:none --no-checkout "https://github.com/${OWNER}/${REPO}.git" "${CREW_LIB_PATH}"/.crewsetuptmp
 
-# prepare sparse checkout .rb packages directory and do it
+# move .git directory to final location and remove tmp folder.
+mv "${CREW_LIB_PATH}"/.crewsetuptmp/.git "${CREW_LIB_PATH}"/ && rm -rf "${CREW_LIB_PATH}"/.crewsetuptmp
```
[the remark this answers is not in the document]
uberhacker: I think we're spending too much effort to support existing installs.  It's SOOO much easier just to blow away the old and start fresh rather than working around already installed files which are most likely corrupt anyway if another install is attempted.

satmandu:
I'll modify the logic.

satmandu:
Logic modified @uberhacker .

install.sh
satmandu: Opened https://github.com/skycocker/chromebrew/issues/5865 to track this.

PR #5866, opened 2021-06-11
Improvements to README.md

uberhacker:
Add Latest Packages section, organize commands in a table and rearrange sections.


satmandu:
Nice. Can we automate the latest packages listing?

uberhacker:
We could probably write something to extract info from git.

PR #5867, opened 2021-06-12
Add Core as a metapackage

### LINKED ISSUE #5865: crew should check on update to see if there are uninstalled packages in core.
- Since `install.sh` is installing `core`, we should make sure that any changes to core are reflected later too in the installation state.

satmandu:
Fixes #5865 (Except `crew` modifications)

- Adds `core.rb` as a metapackage
- Add libyaml to Bootstrap packages (otherwise ruby can complain.)
- Update `install.sh` to install `core`.

Works properly:
- [x] x86_64


packages/core.rb
```diff
@@ -0,0 +1,86 @@
+require 'package'
+
+class Core < Package
+  description 'Core Chromebrew Packages.'
+  homepage ''
```
uberhacker, line 5: `homepage 'https://github.com/skycocker/chromebrew'`?

packages/core.rb
```diff
@@ -0,0 +1,86 @@
...
+  depends_on 'libidn2'
+  depends_on 'libmetalink'
+  depends_on 'libnghttp2'
+  depends_on 'libpipeline'
+  depends_on 'libpsl'
+  depends_on 'libseccomp'
+  depends_on 'libsigsegv'
+  depends_on 'libssp'
+  depends_on 'libtasn1'
+  depends_on 'libtirpc'
```
uberhacker, line 49: `depends_on 'libunbound'`?

packages/core.rb
```diff
@@ -0,0 +1,86 @@
...
+  depends_on 'pixz'
+  depends_on 'py3_pip'
+  depends_on 'py3_setuptools'
+  depends_on 'py3_wheel'
+  depends_on 'python2'
+  depends_on 'python3'
+  depends_on 'readline'
+  depends_on 'rtmpdump'
+  depends_on 'ruby'
+  depends_on 'sed'
```
uberhacker, line 78: Should we remove acl, attr, filecmd and sed since they are already included in buildessential?

satmandu: Sure.

PR #5871, opened 2021-06-13
Geany rebuild

satmandu:
- Rebuild, (recent startup speed fix & gtk2 cleanup)
- modify postinstall to make sure geany icons get properly registered.

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/geany.rb
+++ b/packages/geany.rb
@@ -3,32 +3,37 @@ require 'package'
 class Geany < Package
   description 'Geany is a small and lightweight Integrated Development Environment.'
   homepage 'https://www.geany.org/'
-  version '1.37.1'
+  version '1.37.1-b524'
   license 'GPL-2+ HPND'
   compatibility 'all'
-    source_url 'https://download.geany.org/geany-1.37.1.tar.bz2'
-    source_sha256 '18c5756444c1d8bcd737c8ecfd4ef0b3607c924fc02560d4e8b78f6121531a18'
+  source_url 'https://github.com/geany/geany.git'
+  git_hashtag 'b524a58e12e85c94a32f64fb72615978b2628af1'
 
-  binary_url ({
-     aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/geany/1.37.1_armv7l/geany-1.37.1-chromeos-armv7l.tar.xz',
-      armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/geany/1.37.1_armv7l/geany-1.37.1-chromeos-armv7l.tar.xz',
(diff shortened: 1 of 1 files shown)


packages/geany.rb
```diff
@@ -41,12 +46,16 @@ def self.install
   end
 
   def self.postinstall
+    # This is needed to aoid "Error loading theme icon 'geany-build' for stock" messages.
```
uberhacker, line 49: avoid?  Missing v?  What package provides `gtk-update-icon-cache`?  If it is a dependency then no need to check if the binary exists first.

packages/geany.rb
```diff
@@ -41,12 +46,16 @@ def self.install
...
+    end
     puts
     puts "To get started, type 'geany'.".lightblue
     puts
-    puts "To completely uninstall, execute the following:".lightblue
-    puts "crew remove geany".lightblue
-    puts "rm -rf ~/.config/geany".lightblue
+    puts 'To completely uninstall, execute the following:'.lightblue
+    puts 'crew remove geany'.lightblue
+    puts 'rm -rf ~/.config/geany'.lightblue
```
uberhacker, line 58: Could these steps be added to `def self.remove` with a prompt instead?

packages/geany.rb
satmandu: True too. I think gtk3 has that, so just adding the dependency should take care of it. And yes that's a typo.

packages/geany.rb
satmandu: Yes, that could be handled in such a manner. If I get to it before I leave tomorrow I will add that.

packages/geany.rb
satmandu: Looks like gtk3 & gtk2 have it, so this is already covered...

packages/geany.rb
satmandu: Added self.remove

packages/geany.rb
```diff
@@ -3,32 +3,40 @@
...
   })
 
-    depends_on 'gtk3'
-    depends_on 'xdg_base'
-    depends_on 'sommelier'
+  depends_on 'gtk3'
+  depends_on 'xdg_base'
+  depends_on 'sommelier'
 
+  @xdg_config_home = ENV['XDG_CONFIG_HOME']
```
uberhacker, line 29: Doesn't this make an assumption the environment variable is set?  What if it isn't set?

packages/geany.rb
```diff
@@ -3,32 +3,40 @@
...
+  @xdg_config_home = "#{CREW_PREFIX}/.config" if @xdg_config_home.to_s.empty?
```
uberhacker, line 30: Never mind.  I see you set it here is it's empty.

PR #5895, opened 2021-06-23
Run postinstalls together, after all package installs, and run them in a tempdir instead of CREW_DEST_DIR

satmandu:
- Currently, postinstalls are run after each install, which is a problem when multiple package installs are being done together and programs being run in one postinstall depend upon files in the other packages which have not yet been extracted.
- This extracts files in all packages to be installed together first and then runs the postinstalls for those packages.
- `post_install` also uses a tempdir, and no longer assumes that they will be run in `CREW_DEST_DIR`
- Also adjusted coloring of output to more easily tell which packages are being installed in a multi-package install.
- Also removes the exclusion for disallowing `is_fake` packages from having postinstalls... just to make things simpler.
- `crew upgrade` will still do postinstalls in chunks of packages. Maybe that doesn't matter since packages being installed together are still having postinstalls done together?  e.g.
image
- Let a `SKIP` argument be used for `source_sha256`. (needed for current Minecraft pkg.)

Works properly:
- [x] x86_64

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -656,7 +656,7 @@ def download
           cachefile = ''
         else
           puts "#{@pkg.name.capitalize} archive file exists in cache".lightgreen if @opt_verbose
-          if Digest::SHA256.hexdigest( File.read(cachefile) ) == sha256sum
+          if Digest::SHA256.hexdigest( File.read(cachefile) ) == sha256sum || sha256sum =~ /^SKIP$/i
             begin
               # Hard link cached file if possible.
               FileUtils.ln cachefile, CREW_BREW_DIR, force: true, verbose: @fileutils_verbose unless File.identical?(cachefile,CREW_BREW_DIR + '/' + filename)
@@ -677,7 +677,7 @@ def download
       # Download file if not cached.
       system "#{CURL} --retry 3 -#{@verbose}#LC - --insecure \'#{url}\' --output #{filename}"
       abort 'Checksum mismatch. 😔 Try again.'.lightred unless
-        Digest::SHA256.hexdigest( File.read(filename) ) == sha256sum
+        Digest::SHA256.hexdigest( File.read(filename) ) == sha256sum || sha256sum =~ /^SKIP$/i
       puts "#{@pkg.name.capitalize} archive downloaded.".lightgreen
       # Stow file in cache if requested (and if file is not from cache).
       if CREW_CACHE_ENABLED and cachefile.to_s.empty?
--- a/bin/crew
+++ b/bin/crew
@@ -1127,7 +1127,7 @@ def resolve_dependencies
       print_current_package
       install
     end
-    if @resolve_dependencies_and_install == 1
+    if @resolve_dependencies_and_install == 1 or @resolve_dependencies_and_build == 1
       @to_postinstall = @dependencies
     else
       @dependencies.each do |dep|
@@ -1177,7 +1177,7 @@ def install
     # perform install process
     install_package dest_dir
 
-    unless @resolve_dependencies_and_install == 1
+    unless @resolve_dependencies_and_install == 1 or @resolve_dependencies_and_build == 1
       # perform post-install process
       post_install
     end
@@ -1195,13 +1195,18 @@ def install
 end
 
 def resolve_dependencies_and_build
+  @resolve_dependencies_and_build = 1
+  @to_postinstall = []
   begin
     origin = @pkg.name
 
     # mark current package as which is required to compile from source
     @pkg.build_from_source = true
     resolve_dependencies
-
+    @to_postinstall.each do |dep|
+      search dep
+      post_install
+    end
     search origin, true
     build_package Dir.pwd
   rescue InstallError => e
@@ -1214,6 +1219,7 @@ def resolve_dependencies_and_build
     end
   end
   puts "#{@pkg.name} is built!".lightgreen
+  @resolve_dependencies_and_build = 0
 end
 
 def build_package(pwd)
--- a/bin/crew
+++ b/bin/crew
@@ -11,6 +11,7 @@ require 'digest/sha2'
 require 'json'
 require 'fileutils'
 require 'securerandom'
+require 'tmpdir'
 require_relative '../lib/const'
 require_relative '../lib/util'
 
@@ -845,9 +846,11 @@ def pre_install(dest_dir)
 end
 
 def post_install
-  Dir.chdir CREW_DEST_DIR do
-    puts "Performing post-install for #{@pkg.name}...".lightblue
-    @pkg.postinstall
+  Dir.mktmpdir do |post_install_tempdir|
+    Dir.chdir post_install_tempdir do
+      puts "Performing post-install for #{@pkg.name}...".lightblue
+      @pkg.postinstall
+    end
   end
 end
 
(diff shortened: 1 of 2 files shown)


bin/crew
```diff
@@ -844,10 +845,12 @@ def pre_install(dest_dir)
...
-  Dir.chdir dest_dir do
-    puts 'Performing post-install...'
-    @pkg.postinstall
+def post_install
+  Dir.mktmpdir do |post_install_tempdir|
+    Dir.chdir post_install_tempdir do
+      puts "Performing post-install for #{@pkg.name}...".lightblue
+      @pkg.postinstall
+    end
   end
```
uberhacker, line 854: What is the justification for running postinstall in a temporary directory?

satmandu: post_install was previously being passed CREW_DEST_DIR, and the contents of that could change between package installs. Since postinstall is now serialized, one can not assume that CREW_DEST_DIR will stay the same between running the package extract and the postinstall, so this makes it explicit that post_install does not have access to the package extraction CREW_DEST_DIR.

uberhacker: This has been an issue for a long time but wouldn't it make sense to clear out `CREW_DEST_DIR` after each package is processed?  I think you need to remove the temp dir after processing also, correct?  When installing packages that have a lot of dependencies, it can potentially require a lot of disk space if it isn't cleared out each time, right?

uberhacker: I can see a potential problem with delaying postinstalls until the end.  Some packages may "depend" on a postinstall process as a prerequisite.

satmandu: The temp directory is automatically removed at the end of the function block.

CREW_DEST_DIR is only used for builds, yes?

That CREW_DEST_DIR is what post_install was being pointed to is its own issue, which is now addressed.

I think we are talking about `CREW_BREW_DIR` though, and that is handled already by 
```
    # cleanup
    unless @opt_keep
      FileUtils.rm_rf Dir.glob("#{CREW_BREW_DIR}/*")
      FileUtils.mkdir_p "#{CREW_BREW_DIR}/dest" # this is a little ugly, feel free to find a better way
    end
```

`@extract_dir` appears to be the location for packages being extracted, 

(One could also do the same for CREW_BREW_DIR, by using the tempdir function for that, but that might get hairy since we need to make sure that the mktempdir function block encompasses all use of that folder.)

Maybe that's a separate PR?

uberhacker: Is the idea behind delaying the postinstalls is so the messages don't scroll by as other dependencies are installed?  If that is the only reason, we could create a new section just for messages.  Would something like `def self.messages` work?

satmandu: The primary reason behind delaying postinstalls is to make sure all binaries from dependencies are installed before they can be invoked from postinstall sections of packages. This way we avoid issues with worrying about whether for instance an icon cache for `geany` can be updated by `gtk-update-icon-cache` before `gtk-update-icon-cache` is available from having installed the `gtk3` package.

Having postinstall messages all show up at the end is just a side benefit.

satmandu:
Noticed this issue:
```
Preconfiguring package...
/usr/local/lib64/ruby/3.0.0/fileutils.rb:494:in `realpath': No such file or directory @ rb_check_realpath_internal - opt (Errno::ENOENT)
        from /usr/local/lib64/ruby/3.0.0/fileutils.rb:494:in `copy_entry'
        from /usr/local/lib64/ruby/3.0.0/fileutils.rb:471:in `block in cp_r'
        from /usr/local/lib64/ruby/3.0.0/fileutils.rb:1597:in `block in fu_each_src_dest'
        from /usr/local/lib64/ruby/3.0.0/fileutils.rb:1611:in `fu_each_src_dest0'
        from /usr/local/lib64/ruby/3.0.0/fileutils.rb:1595:in `fu_each_src_dest'
        from /usr/local/lib64/ruby/3.0.0/fileutils.rb:470:in `cp_r'
        from /usr/local/lib/crew/packages/minecraft.rb:20:in `install'
        from /usr/local/bin/crew:827:in `block in build_and_preconfigure'
        from /usr/local/bin/crew:808:in `chdir'
        from /usr/local/bin/crew:808:in `build_and_preconfigure'
        from /usr/local/bin/crew:1154:in `install'
        from /usr/local/bin/crew:1019:in `resolve_dependencies_and_install'
        from /usr/local/bin/crew:1388:in `block in install_command'
        from /usr/local/bin/crew:1383:in `each'
        from /usr/local/bin/crew:1383:in `install_command'
        from /usr/local/bin/crew:1477:in `<main>'
```

uberhacker:
I think it's dangerous to make the assumption that ALL postinstalls are safe to delay.

satmandu:
The issue was with the `minecraft` package, which is now broken due to an update to `minecraft-launcher`...

satmandu:
> I think it's dangerous to make the assumption that ALL postinstalls are safe to delay.

This is literally what Debian/Ubuntu does. All package scripts are run after all packages are unpacked to their target locations.

satmandu:
Are there any specific postinstalls that are sensitive to ordering you can think of?

uberhacker:
Well, we did have that issue with `shared_mime_info` but that was solved by adding it to `CREW_FIRST_PACKAGES`.  #5897 is not related but I suppose if `glib` is not installed first, it could be an issue.  Maybe we should add `glib` to `CREW_FIRST_PACKAGES` also just in case?

satmandu:
It doesn't matter if `glib` isn't installed first. As long as `glib` is listed as a dependency for packages which need it, glib will be extracted  with all other dependencies such as gtk3, etc _before_ postinstalls are called which would run binaries such as `glib-compile-schemas`.

(The problem with glib is that it isn't in core and it isn't listed as a dependency inside gtk3, etc...)

satmandu:
This won't fix https://github.com/skycocker/chromebrew/issues/5897 because that's a segfault issue, but this will fix installing gtk3, shared_mime_info, and so on, since those will because of dependencies all be extracted together before postinstalls.

bin/crew
```diff
@@ -1158,8 +1177,10 @@ def install
     # perform install process
     install_package dest_dir
 
-    # perform post-install process
-    post_install dest_dir
+    unless @resolve_dependencies_and_install == 1 or @resolve_dependencies_and_build == 1
```
uberhacker, line 1180: I'm confused why we would even need this conditional.

satmandu: Otherwise, postinstall is run twice for the first package we are trying to install.

uberhacker: This whole process seems overly complicated to me.

satmandu: It's an artifact of the current decision tree.

Both `resolve_dependencies_and_install` & `resolve_dependencies_and_build` will invoke `post_install` on a notional `package 1` after finding all dependencies, since it creates an array of all _dependencies for_ `package 1`, and also adds `package 1` to that array, which it can then use to run `post_install`s for everything in that array in sequence.  So we need to exclude `package 1` here, otherwise, `package 1` gets `post_install` run twice.

uberhacker: Would it make sense to remove the entire code block since post_install is called in another place?

satmandu: No, because you need if there if you are just calling install on a package w/o dependencies which need installing.

Just trying to take a light touch here with a minimum of refactoring.

uberhacker: It just doesn't feel elegant but I guess it works for now.

satmandu: Did a clean install with buildessential in a container and found no issues.

PR #5901, opened 2021-06-24
geany plugins with gcc11 compile fix

satmandu:
- prior version would not build with gcc 11

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l

### DIFF
--- a/packages/geany_plugins.rb
+++ b/packages/geany_plugins.rb
@@ -3,33 +3,50 @@ require 'package'
 class Geany_plugins < Package
   description 'Geany plugins.'
   homepage 'https://www.geany.org/'
-  version '1.37'
+  version '1.37-7c54'
   license 'GPL-2'
-  compatibility 'aarch64,armv7l,x86_64'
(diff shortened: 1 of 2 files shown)


packages/geany_plugins.rb
```diff
@@ -3,33 +3,50 @@
...
+  depends_on 'gdk_pixbuf' # R
+  depends_on 'geany' # R
+  depends_on 'glib' # R
+  depends_on 'gtk3' # R
+  depends_on 'harfbuzz' # R
+  depends_on 'libsoup2' # R
+  depends_on 'libxml2' # R
+  depends_on 'pango' # R
+  depends_on 'webkit2gtk_4' # R
+  depends_on 'zlibpkg' # R
```
uberhacker, line 38: Doesn't `depends_on 'geany'` already cover a lot of these dependencies already with gtk3 and sommelier?  Not sure why we need to include all of these.

satmandu: Ah, yes, this is an excuse to fixup geany dependencies too.

PR #5902, opened 2021-06-25
gtk4 -> 4.3.1

satmandu:
Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l

### DIFF
--- a/packages/gtk4.rb
+++ b/packages/gtk4.rb
@@ -3,25 +3,23 @@ require 'package'
 class Gtk4 < Package
   description 'GTK+ is a multi-platform toolkit for creating graphical user interfaces.'
   homepage 'https://developer.gnome.org/gtk4/'
-  @_ver = '4.2.1'
+  @_ver = '4.3.1'
   @_ver_prelastdot = @_ver.rpartition('.')[0]
   version @_ver
   license 'LGPL-2.1'
-  compatibility 'all'
+  compatibility 'x86_64 aarch64 armv7l'
   source_url 'https://gitlab.gnome.org/GNOME/gtk.git'
   git_hashtag @_ver
 
   binary_url({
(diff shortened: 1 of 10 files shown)


packages/gtk4.rb
```diff
@@ -3,25 +3,23 @@
...
   })
   binary_sha256({
-    aarch64: '131b1ef2ee52e9e4e19ac99a2973f2a6b0322d9bb02b7141f792bfe9f6f0b7e8',
-     armv7l: '131b1ef2ee52e9e4e19ac99a2973f2a6b0322d9bb02b7141f792bfe9f6f0b7e8',
-       i686: '49d26942d98c39a093fd341cfb5f62c97d869c20246304ccd8f9ff1cee5c4b92',
-     x86_64: '47520c1d162a3f626e7df1dd50a936b737ce4cf09e76ed1cec228e806fc3e321'
+    aarch64: 'd714a0b5c0318ee50ad3e7bd581e4d69435aafcf3800b640030f940e5accb1cd',
+     armv7l: 'd714a0b5c0318ee50ad3e7bd581e4d69435aafcf3800b640030f940e5accb1cd',
+     x86_64: '986e9290a4ba1e216db1f7d6c89ce4d86cf9ecae1af3b90c1ad43fd385237c75'
   })
```
uberhacker, line 23: If we are going to remove i686, we need to also remove i686 binaries for all packages that depend on gtk4 also, otherwise kludgy errors will occur.

satmandu: good call. Direct dependencies on gtk4 have had i686 binaries removed.

uberhacker:
This could be opening Pandora's box because now any package that depends on any of these needs the i686 binaries removed also.

satmandu:
For sure... the complaints we get will let us know what our i686 gui userbase is too though...

uberhacker:
I don't expect many but if we do the cleanup now, we won't be generating i686 binaries unnecessarily for future updates.

PR #5913, opened 2021-06-29
Update komodo from 11.1.1 to 12.0.1

uberhacker:
Komodo IDE is now free.  See https://www.activestate.com/products/komodo-ide/.  Tested on x86_64.

### DIFF
--- a/packages/komodo.rb
+++ b/packages/komodo.rb
@@ -1,24 +1,13 @@
 require 'package'
 
 class Komodo < Package
-  description 'Komodo Edit is a great editor if you\'re looking for something powerful, yet simple.'
-  homepage 'https://www.activestate.com/products/komodo-edit/'
-  version '11.1.1'
-  license '' # Can't find license for project
-  compatibility 'all'
-  case ARCH
-  when 'i686'
-    source_url 'https://downloads.activestate.com/Komodo/releases/11.1.1/Komodo-Edit-11.1.1-18206-linux-x86.tar.gz'
-    source_sha256 'b2879342b3fcda0688bd461ad9545b178a3315c6ab7229b350db58b794f7c9a5'
-  when 'x86_64'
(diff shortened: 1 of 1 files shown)


packages/komodo.rb
```diff
@@ -26,12 +22,27 @@ class Komodo < Package
...
-    system './install.sh'
+    system "./install.sh -I #{CREW_DEST_PREFIX}"
+  end
+
+  def self.remove
+    print "Would you like to remove the config directories? [y/N] "
+    response = STDIN.getc
+    config_dirs = ["#{HOME}/.komodoide", "#{HOME}/.activestate"]
+    config_dirs.each { |config_dir|
+      if Dir.exists? config_dir
```
supechicken, line 37: This line can be removed as `rm_rf` will not produce error when file not exist

saltedcoffii:
Does the EULA permit us to distribute binaries? I would hate for chromebrew  to get in trouble

satmandu:
@saltedcoffii has a valid concern here. I'm not seeing anything in the EULA on the web site about allowing distribution of binaries: https://www.activestate.com/products/komodo-ide/license-agreement/

uberhacker:
I think the main concern of the license is to restrict selling or reverse engineering the software.  We are not doing any of that.  It would be to ActiveState's benefit that we make their software easily available on Chromebooks because it promotes more users to sign up for subscriptions.

packages/komodo.rb
uberhacker: True but if the directory doesn't exist then the command is executed unnecessarily.

satmandu:
@uberhacker Respectfully, I'm not sure that's how licensing works when the lawyers get involved. The conservative approach would be to only have binaries that have a clear indication (through licensing or otherwise) that distribution is allowed.

If there's a shadow of a doubt... don't distribute it.

uberhacker:
Can you point out the specific paragraph in the license that casts a shadow of doubt because I'm not seeing it?

satmandu:
Screenshot 2021-06-30 11 56 12

PR #5915, opened 2021-06-30
Add valgrind to buildessential

satmandu:
- many builds require `valgrind` for building. Let's just add it to `buildessential`.

Works properly:
- [x] x86_64


uberhacker:
I suppose this is a good addition.

PR #5921, opened 2021-07-02
Update fpc from 3.2.0 to 3.2.2

uberhacker:
Tested on all architectures.


donaldysmith:
Thank you very much for the package. I love to program with pascal.

My system is Acer C720p with Chromebook flex, developer mode enabled. 

When I run fpc, I got the error message
Error: ppcx64 can't be executed, error message: Failed to execute "ppcx64", error code: 127

PR #5922, opened 2021-07-04
sommelier: Add `xhost` and remove `coreutils` dependency

supechicken:
`xhost` is used to allow root access in `sommelierd`
`readlink` is in system version of `coreutils` already but don’t have a symlink to PATH

### DIFF
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -84,7 +84,7 @@ class Sommelier < Package
 
           case "$(uname -m)" in
           x86_64|i686)
-            sort_result = "$(sort -V <<< "$(uname -r)"$'\\n'"4.16.0" | tail -n1)"
+            sort_result="$(sort -V <<< "$(uname -r)"$'\\n'"4.16.0" | tail -n1)"
             if [[ "${sort_result}" == "4.16.0" ]]; then
               MESA_LOADER_DRIVER_OVERRIDE=i965
             fi
@@ -125,7 +125,7 @@ class Sommelier < Package
           basedir=${base%/*}
           LD_ARGV0_REL="../bin/sommelier" \
             exec "${basedir}/..#{@peer_cmd_prefix}" \
-              ../bin/sommelier \
+              --argv0 "$0" \
               --library-path \
               "${basedir}/../#{ARCH_LIB}" \
               --inhibit-rpath '' \
@@ -285,8 +285,7 @@ class Sommelier < Package
     # all tasks are done by sommelier.env now
     puts
     puts 'Removing old sommelier env variable loading code in ~/.bashrc'.lightblue
-    system 'sed -i "s,set -a && source ~/.sommelier.env && set +a,," -i.backup ~/.bashrc'
-    system 'sed -i "s,set -a ; source ~/.sommelier-default.env ; source ~/.sommelier.env ; set +a,," -i.backup ~/.bashrc'
+    system 'sed -i "/[sS]ommelier/d" -i.backup ~/.bashrc'
     puts 'To complete the installation, execute the following:'.orange
     puts 'source ~/.bashrc'.orange
 
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -296,7 +296,7 @@ class Sommelier < Package
       To restart the sommelier daemon, run 'restartsommelier'
     
     EOT
-    puts <<EOT.orange
+    puts <<~EOT.orange
       Please be aware that gui applications may not work without the
       sommelier daemon running.
     
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -99,7 +99,7 @@ class Sommelier < Package
         # This file via:
         # crostini: /opt/google/cros-containers/bin/sommelier
         # https://source.chromium.org/chromium/chromium/src/+/master:third_party/chromite/third_party/lddtree.py;drc=46da9a8dfce28c96765dc7d061f0c6d7a52e7352;l=146
-        IO.write 'sommelier_sh' <<~EOF
+        IO.write 'sommelier_sh', <<~EOF
           #!/bin/bash
           function readlink(){
             coreutils --coreutils-prog=readlink "$@"
@@ -128,7 +128,7 @@ class Sommelier < Package
         EOF
 
         # sommelierd
-        IO.write 'sommelierd' <<~EOF
+        IO.write 'sommelierd', <<~EOF
           #!/bin/bash -a
 
           source ${CREW_PREFIX}/etc/env.d/sommelier.env &>/dev/null
@@ -182,7 +182,7 @@ class Sommelier < Package
         EOF
         
         # startsommelier
-        IO.write 'startsommelier' <<~EOF
+        IO.write 'startsommelier', <<~EOF
           #!/bin/bash -a
 
           source ~/.sommelier-default.env &>/dev/null
@@ -228,7 +228,7 @@ class Sommelier < Package
         EOF
 
         # stopsommelier
-        IO.write 'stopsommelier' <<~EOF
+        IO.write 'stopsommelier', <<~EOF
           #!/bin/bash
           SOMM="$(pgrep -fc sommelier.elf 2> /dev/null)"
           if [[ "${SOMM}" -gt "0" ]]; then
@@ -247,7 +247,7 @@ class Sommelier < Package
         EOF
         
         # restartsommelier
-        IO.write 'restartsommelier' <<~EOF
+        IO.write 'restartsommelier', <<~EOF
           #!/bin/bash
           stopsommelier && startsommelier
         EOF
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -5,7 +5,7 @@ class Sommelier < Package
   homepage 'https://chromium.googlesource.com/chromiumos/platform2/+/HEAD/vm_tools/sommelier/'
   version '20210109-4'
   license 'BSD-Google'
-  compatibility 'armv7l aarch64 x86_64'
+  compatibility 'all'
   source_url 'https://chromium.googlesource.com/chromiumos/platform2.git'
   git_hashtag 'f3b2e2b6a8327baa2e62ef61036658c258ab4a09'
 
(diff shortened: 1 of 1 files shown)


packages/sommelier.rb
```diff
@@ -109,6 +96,10 @@ def self.build
         # https://source.chromium.org/chromium/chromium/src/+/master:third_party/chromite/third_party/lddtree.py;drc=46da9a8dfce28c96765dc7d061f0c6d7a52e7352;l=146
         system "cat <<'EOF'> sommelier_sh
 #!/bin/bash
+function readlink(){
+  coreutils --coreutils-prog=readlink "$@"
```
satmandu, line 100: Doesn't doing this still require coreutils?

supechicken: `coreutils` is a built-in command for GNU based Linux (`/usr/bin/coreutils`)

uberhacker: I have to agree with @supechicken on this.  We shouldn't need coreutils.

uberhacker: If we're going to update, we might as well get the latest code?  https://chromium.googlesource.com/chromiumos/platform2.git/+/be7553c1d04f95cae6d6e7c3c2292fb0440ed2db

supechicken: > If we're going to update, we might as well get the latest code? 

Done

packages/sommelier.rb
```diff
@@ -230,14 +221,14 @@ def self.install
...
+        FileUtils.install 'startsommelier', "#{CREW_DEST_PREFIX}/bin/startsommelier", mode: 0o755
         FileUtils.ln_sf 'startsommelier', "#{CREW_DEST_PREFIX}/bin/initsommelier"
-        system "install -Dm755 stopsommelier #{CREW_DEST_PREFIX}/bin/stopsommelier"
-        system "install -Dm755 restartsommelier #{CREW_DEST_PREFIX}/bin/restartsommelier"
-        system "install -Dm755 sommelierrc #{CREW_DEST_PREFIX}/etc/sommelierrc"
-        system "install -Dm644 .sommelier-default.env #{CREW_DEST_HOME}/.sommelier-default.env"
+        FileUtils.install 'stopsommelier', "#{CREW_DEST_PREFIX}/bin/stopsommelier", mode: 0o755
+        FileUtils.install 'restartsommelier', "#{CREW_DEST_PREFIX}/bin/restartsommelier", mode: 0o755
+        FileUtils.install 'sommelierrc', "#{CREW_DEST_PREFIX}/etc/sommelierrc", mode: 0o755
+        FileUtils.install '.sommelier-default.env', "#{CREW_DEST_HOME}/.sommelier-default.env", mode: 0o644
```
uberhacker, line 231: If we use FileUtils.install, don't we have to create directories first?  Does this compile without errors?

packages/sommelier.rb
```diff
@@ -230,14 +221,14 @@ def self.install
...
+        FileUtils.install 'sommelierd', "#{CREW_DEST_PREFIX}/sbin/sommelierd", mode: 0o755
+        FileUtils.install 'startsommelier', "#{CREW_DEST_PREFIX}/bin/startsommelier", mode: 0o755
         FileUtils.ln_sf 'startsommelier', "#{CREW_DEST_PREFIX}/bin/initsommelier"
-        system "install -Dm755 stopsommelier #{CREW_DEST_PREFIX}/bin/stopsommelier"
-        system "install -Dm755 restartsommelier #{CREW_DEST_PREFIX}/bin/restartsommelier"
-        system "install -Dm755 sommelierrc #{CREW_DEST_PREFIX}/etc/sommelierrc"
-        system "install -Dm644 .sommelier-default.env #{CREW_DEST_HOME}/.sommelier-default.env"
+        FileUtils.install 'stopsommelier', "#{CREW_DEST_PREFIX}/bin/stopsommelier", mode: 0o755
+        FileUtils.install 'restartsommelier', "#{CREW_DEST_PREFIX}/bin/restartsommelier", mode: 0o755
+        FileUtils.install 'sommelierrc', "#{CREW_DEST_PREFIX}/etc/sommelierrc", mode: 0o755
```
uberhacker, line 230: Since this is not an executable, we can use `mode: 0o644`.

packages/sommelier.rb
satmandu: Newer platform2 sommelier doesn't work due to issues with the virtual device not showing up when sommelier isn't tunneling through crosvm.

I have no problems with updating to newer git so long as you build and confirm sommelier works!

(Otherwise revert to the older version.). 😅

supechicken: Just check it in my system, it built successfully on chroot but failed on my system

packages/sommelier.rb
supechicken: Done

packages/sommelier.rb
uberhacker: Should we revert or try an older git hash?

packages/sommelier.rb
```diff
@@ -21,8 +21,7 @@ class Sommelier < Package
        i686: 'f00ace87c154fa1c4f786656db369004612b51c27236dbdca9c4f0fd230496eb',
      x86_64: '29b161ee80fbc1f334bfaf7d191ce8b37c0b7cff414d3e639934b10de64280d7'
   })
```
uberhacker, line 23: We will need new binaries so remove and bump the version?

satmandu: @supechicken Can you confirm that this sommelier works when built?

uberhacker: @supechicken: What is the status of this PR?

supechicken: I need to get back my laptop from my friend for testing

packages/sommelier.rb
supechicken: The current version is the latest before dropping virtwl support and I can get it working

packages/sommelier.rb
supechicken: > We will need new binaries so remove and bump the version?

Done

supechicken:
- Reverted sommelier git commit to the current
- Updated `sommelier_sh` wrapper
- Use heredoc

Tested on `x86_64`

packages/sommelier.rb
```diff
@@ -47,246 +39,281 @@ class Sommelier < Package
     @peer_cmd_prefix = '/lib64/ld-linux-x86-64.so.2'
   end
 
+  def self.preflight
+    unless File.socket?('/var/run/chrome/wayland-0')
+      abort "This package is not compatible with your device :/".lightred
+    end
+  end
+
+  def self.prebuild
```
satmandu, line 48: Why this instead of self.patch?

packages/sommelier.rb
```diff
@@ -47,246 +39,281 @@ class Sommelier < Package
...
+    # Patch to avoid error with GCC > 9.x
+    # ../sommelier.cc:3238:10: warning: ‘char* strncpy(char*, const char*, size_t)’ specified bound 108 equals destination size [-Wstringop-truncation]
+    Kernel.system "sed -i 's/sizeof(addr.sun_path))/sizeof(addr.sun_path) - 1)/' sommelier.cc", chdir: 'vm_tools/sommelier'
```
satmandu, line 51: Is there an advantage to using `Kernel.system` over just `system`?

( Asking because I have no idea! ) :)

packages/sommelier.rb
```diff
@@ -47,246 +39,281 @@ class Sommelier < Package
...
-        -Db_lto_mode=thin \
-        -Dc_args='-fuse-ld=lld' \
-        -Dcpp_args='-fuse-ld=lld' \
-        -Dcpp_link_args='-fuse-ld=lld' \
-        -Dxwayland_path=#{CREW_PREFIX}/bin/Xwayland \
-        -Dxwayland_gl_driver_path=/usr/#{ARCH_LIB}/dri -Ddefault_library=both \
-        -Dxwayland_shm_driver=noop -Dshm_driver=noop -Dvirtwl_device=/dev/null \
-        -Dpeer_cmd_prefix=\"#{CREW_PREFIX}#{@peer_cmd_prefix}\" \
-        builddir"
+      system <<~BUILD
```
satmandu, line 58: Nice.

packages/sommelier.rb
```diff
@@ -47,246 +39,281 @@ class Sommelier < Package
...
       end
     end
   end
 
   def self.install
     Dir.chdir('vm_tools/sommelier') do
-      system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
+      system "DESTDIR=#{CREW_DEST_DIR} samu -C builddir install"
       Dir.chdir('builddir') do
+        FileUtils.mkdir_p ["#{CREW_DEST_PREFIX}/bin", "#{CREW_DEST_PREFIX}/sbin", "#{CREW_DEST_PREFIX}/etc", CREW_DEST_HOME]
```
satmandu, line 273: This also works and is slightly shorter:
`FileUtils.mkdir_p %W[#{CREW_DEST_PREFIX}/bin #{CREW_DEST_PREFIX}/sbin #{CREW_DEST_PREFIX}/etc CREW_DEST_HOME]`

packages/sommelier.rb
supechicken: It is for bypassing the patched `system` function in `package.rb`, The `package.rb` one won't work as it will throw an error when trying the `chdir` hash

packages/sommelier.rb
supechicken: According to the `Manual.md`:
```
self.prebuild - Only sed commands should be run during this stage - if provided
```
(but `self.patch` 's meaning is more accurate :) )

supechicken: Changed to `self.patch`

packages/sommelier.rb
satmandu: @uberhacker Any suggestions on making `def self.system(*args)` in `package.rb` work for this?

satmandu:
@uberhacker Do we need a i686 build for this?

uberhacker:
If you can build a i686 binary that works then more power to you.  There was a time in the past when this actually worked.

satmandu:
Does this build for you on i686?

satmandu:
@uberhacker Does this work at all on i686 now?

uberhacker:
I think sommelier is officially dead on i686.  Unless the code is changed, there is no hope.

packages/sommelier.rb
```diff
@@ -35,258 +34,296 @@ class Sommelier < Package
...
 
     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/etc/env.d"
-    @sommelierenv = <<~SOMMELIERENVEOF
-      # Sommelier loading code
-      set -a
-      source ~/.sommelier-default.env
-      source ~/.sommelier.env
-      set +a
-    SOMMELIERENVEOF
     IO.write("#{CREW_DEST_PREFIX}/etc/env.d/sommelier", @sommelierenv)
```
uberhacker, line 296: Can we add `/usr/local/etc/bash.d/sommelier` and insert `startsommelier` so it autostarts on login?  This removes the need to add it to `~/.bashrc` and the autostart goes away after removing the package.

satmandu: Done, and binaries have been updated.

PR #5929, opened 2021-07-07
tar => 1.34

supechicken:
Tested on `x86_64`

### DIFF
--- a/packages/tar.rb
+++ b/packages/tar.rb
@@ -6,7 +6,7 @@ class Tar < Package
   version '1.34'
   license 'BSD'
   compatibility 'all'
-  source_url 'https://ftpmirror.gnu.org/gnu/tar/tar-1.34.tar.gz'
+  source_url 'https://ftpmirror.gnu.org/gnu/tar/tar-1.34.tar.xz'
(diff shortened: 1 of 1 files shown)


packages/tar.rb
```diff
@@ -3,24 +3,13 @@
 class Tar < Package
   description 'GNU Tar provides the ability to create tar archives, as well as various other kinds of manipulation.'
   homepage 'https://www.gnu.org/software/tar/'
-  version '1.33'
+  version '1.34'
   license 'BSD'
   compatibility 'all'
-  source_url 'https://ftpmirror.gnu.org/tar/tar-1.33.tar.xz'
-  source_sha256 '66a8344b1dc83a411d311bd1547e0176e56cc311f28ee94a30f84dafb3d9067e'
+  source_url 'https://ftpmirror.gnu.org/gnu/tar/tar-1.34.tar.gz'
```
uberhacker, line 9: The url should work without the `gnu` also.

uberhacker: What is wrong with the `.tar.xz` extension?  Should be a smaller tarball with better compression.

supechicken: Since this file is auto-generated by my [update checker](https://github.com/supechicken/crew-package-updater), some of the sources may be changed

PR #5933, opened 2021-07-07
Allow crew to handle git urls

satmandu:
allows `git://` `source_url`s

(Works on `x86_64`)

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -642,8 +642,13 @@ def download
     puts "No precompiled binary available for your platform, downloading source..."
   end
 
-  Dir.chdir CREW_BREW_DIR do
+  if uri.scheme =~ /git/
+    @git = true
+  else
+    @git = false
+  end
 
+  Dir.chdir CREW_BREW_DIR do
     case File.basename(filename)
     # Sources that download with curl
     when /\.zip$/i, /\.(tar(\.(gz|bz2|xz|lz))?|tgz|tbz|tpxz|txz)$/i, /\.deb$/i, /\.AppImage$/i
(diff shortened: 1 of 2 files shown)


saltedcoffii:
I would highly advise against supporting `git://` URLs as they use a different port than `http://` (80) and `https://` (443). I don't remember what that port is, but it's specific to the protocol. This means that if a chromebook is behind a network firewall, such as an enterprise one (many chromebooks are) this would prevent them from installing that package from the source. The `git://` protocol is also insecure so it's equivalent to using `http://` as far as encryption goes. All in all, let's just stick to `https://` for `.git` URLs.

satmandu:
@saltedcoffii Does ```git config --global url."https://".insteadOf git://``` work for forcing https?
Alternately maybe adding ```git config --global url."https://git.launchpad.net".insteadOf git://git.launchpad.net```

The problem situation is sites like this: https://git.launchpad.net/ubuntu/+source/libdbusmenu

Note the lack of a `.git` suffix for the site. How would you suggest we create git `source_url`s for that?

satmandu:
(git config info is from https://gist.github.com/taoyuan/bfa3ff87e4b5611b5cbe )

saltedcoffii:
> @saltedcoffii Does `git config --global url."https://".insteadOf git://` work for forcing https?
> Alternately maybe adding `git config --global url."https://git.launchpad.net".insteadOf git://git.launchpad.net`
> 
> The problem situation is sites like this: https://git.launchpad.net/ubuntu/+source/libdbusmenu
> 
> Note the lack of a `.git` suffix for the site. How would you suggest we create git `source_url`s for that?

Perhaps the presence of a `git_hashtag` parameter would tell chromebrew that the `source_url` is a git url?

satmandu:
I don't have time at the moment, but if you wanted to submit a PR to do just that...

PR #5935, opened 2021-07-07
Remmina => 1.4.19 w/ new deps

satmandu:
- Now uses older libsodium.
- Has new required deps added.

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l

### DIFF
--- a/packages/libappindicator_gtk3.rb
+++ b/packages/libappindicator_gtk3.rb
@@ -1,0 +1,57 @@
+# Adapted from Arch Linux libappindicator PKGBUILD at:
+# https://github.com/archlinux/svntogit-community/raw/packages/libappindicator/trunk/PKGBUILD
+
+require 'package'
+
+class Libappindicator_gtk3 < Package
+  description 'Allow applications to extend a menu via Ayatana indicators in Unity, KDE or Systray'
+  homepage 'https://launchpad.net/libappindicator'
+  version '12.10.0'
+  license 'LGPL2.1 LGPL3'
+  compatibility 'x86_64 aarch64 armv7l'
+  source_url 'https://launchpad.net/libappindicator/12.10/12.10.0/+download/libappindicator-12.10.0.tar.gz'
+  source_sha256 'd5907c1f98084acf28fd19593cb70672caa0ca1cf82d747ba6f4830d4cc3b49f'
+
+  binary_url({
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/libappindicator_gtk3/12.10.0_armv7l/libappindicator_gtk3-12.10.0-chromeos-armv7l.tpxz',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/libappindicator_gtk3/12.10.0_armv7l/libappindicator_gtk3-12.10.0-chromeos-armv7l.tpxz',
+     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/libappindicator_gtk3/12.10.0_x86_64/libappindicator_gtk3-12.10.0-chromeos-x86_64.tpxz'
+  })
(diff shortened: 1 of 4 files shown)


packages/libappindicator_gtk3.rb
```diff
@@ -0,0 +1,57 @@
...
+  depends_on 'gobject_introspection'
+  depends_on 'gtk_doc'
+  depends_on 'libindicator_gtk3'
+  depends_on 'pygtk'
+  depends_on 'vala'
+  depends_on 'libdbusmenu_gtk3'
+
+  def self.patch
+    system 'NOCONFIGURE=1 ./autogen.sh'
+    system "grep -rl Werror . | xargs sed -i 's,-Wall -Werror,-Wall,g'"
```
uberhacker, line 37: If we do this, should we have a `def self.check` for testing?

packages/libdbusmenu_gtk3.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  depends_on 'gnome_common'
+  depends_on 'gobject_introspection'
+  depends_on 'gtk2'
+  depends_on 'gtk3'
+  depends_on 'intltool'
+  depends_on 'vala'
+
+  def self.patch
+    system './autogen.sh'
```
uberhacker, line 34: No `NOCONFIGURE=1`?

packages/libindicator_gtk3.rb
```diff
@@ -0,0 +1,52 @@
...
+     armv7l: 'fcfdbe22d476d4fdef9e740149026e867e049f2c9770c7e332206653d002752d',
+     x86_64: '854c2183a47a9ab88db302edb47ac155eb275961020a97eeb10aeac556a07936'
+  })
+
+  depends_on 'gtk3'
+
+  def self.patch
+    system 'NOCONFIGURE=1 autoreconf -fi'
+    system 'filefix'
+    system "grep -rl Werror . | xargs sed -i 's,-Wall -Werror,-Wall,g'"
```
uberhacker, line 31: Do we need `def self.check` for testing?

packages/remmina.rb
```diff
@@ -3,43 +3,51 @@
...
+
   def self.build
     Dir.mkdir 'builddir'
     Dir.chdir 'builddir' do
       system "cmake \
         -G Ninja \
-        #{CREW_CMAKE_OPTIONS} \
-        -DWITH_APPINDICATOR=OFF \
+        #{CREW_CMAKE_OPTIONS.sub("C_FLAGS='",
+                                 "C_FLAGS='-Wno-unused-function ")} \
```
uberhacker, line 50: Is this spacing tabs or spaces?  Seems a bit extreme.

packages/libappindicator_gtk3.rb
satmandu: That's actually done in the ubuntu patches themselves. See https://git.launchpad.net/ubuntu/+source/libappindicator/commit/?id=60c196f25f7afe9fa3bb0360b8ede3035f1c5b47

packages/remmina.rb
satmandu: It is what rubocop suggested. :/

packages/libindicator_gtk3.rb
satmandu: ubuntu's own patches seem to be doing this...

packages/libdbusmenu_gtk3.rb
satmandu: good call

packages/remmina.rb
satmandu: (and that's spaces...) 

What's best practices here?

uberhacker: I don't think you need to break the line unless it exceeds 80 chars?

PR #5941, opened 2021-07-09
glibc 2.32 revision

### LINKED ISSUE #5966: Inkscape cannot be installed or built
## Description
Installation of inkscape fails due to imagemagick6 & imagemagick7 dependency, which cannot be installed together.
Removing the imagemagick6 dep succeeds with installation but then libgsl.so.23 is not found which seems to be outdated since libgsl.so.25 is installed.
Building results in the glibc 2.32 issue mentioned in #5936 

```shell
crew install inkscape
---
inkscape
---
crew build inkscape
```
Output:
```shell
install:

imagemagick6 
Do you agree? [Y/n] Y
Proceeding...
imagemagick6: Use ImageMagick to create, edit, compose, or convert bitmap images.
########################################################################################## 100.0%
Imagemagick6 archive downloaded.
Performing pre-install...
ImageMagick version 7.0.11-2 already installed.
---

running after installation without imemageick6:

inkscape: error while loading shared libraries: libgsl.so.23: cannot open shared object file: No such file or directory
---

build:

/usr/local/bin/ld: cannot find Scrt1.o: No such file or directory
/usr/local/bin/ld: cannot find crti.o: No such file or directory

```

chromebrew: `crew --version`
1.12

ChromeOS: `cat /etc/lsb-release`
CHROMEOS_ARC_ANDROID_SDK_VERSION=28
CHROMEOS_ARC_VERSION=7487557
CHROMEOS_BOARD_APPID={0C1E39B7-DAE6-4972-8004-E96F60D9342C}
CHROMEOS_DEVSERVER=
CHROMEOS_RELEASE_APPID={0C1E39B7-DAE6-4972-8004-E96F60D9342C}
CHROMEOS_RELEASE_BOARD=fizz-signed-mp-v13keys
CHROMEOS_RELEASE_BRANCH_NUMBER=39
CHROMEOS_RELEASE_BUILDER_PATH=fizz-release/R92-13982.39.0
CHROMEOS_RELEASE_BUILD_NUMBER=13982
CHROMEOS_RELEASE_CHROME_MILESTONE=92
CHROMEOS_RELEASE_DESCRIPTION=13982.39.0 (Official Build) beta-channel fizz 
CHROMEOS_RELEASE_KEYSET=mp-v13
CHROMEOS_RELEASE_TRACK=beta-channel
CHROMEOS_RELEASE_UNIBUILD=1
CHROMEOS_RELEASE_VERSION=13982.39.0
DEVICETYPE=CHROMEBOX
GOOGLE_RELEASE=13982.39.0

Model: Asus Chromebox i7

Architecture: `uname -m`
x86_64

satmandu:
Fixes #5966 
- Use glibc 2.32 with glibc 2.32
- Set version depending upon LIBC_VERSION, so users with an older GLIBC aren't bothered.
- This appears to let compilation work with `glibc 2.32`, without adding a dependency on `libgcc_s.so.1`, which causes segfaults.
- On glibc 2.32 this uses the fedora locale pruning method, which was used up until F30. This lets us just use the default install commands for locales, with pruning happening during postinstall. (I couldn't find a way to easily move this up into `install`, since the `build-locale-archive` tool is not DESTROOT aware.
- The key to getting this to work was:
```
system "sed 's,/usr/#{ARCH_LIB}/libc_nonshared.a,#{CREW_LIB_PREFIX}/libc_nonshared.a,g' /usr/#{ARCH_LIB}/libc.so > #{CREW_DEST_LIB_PREFIX}/libc.so"
```

Test with:
```
CREW_TESTING=1 CREW_TESTING_BRANCH=glibc_2.32-1 \
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew \
crew update ; crew reinstall glibc ; crew update
```
Nothing should break. :)

Works properly:
- [x] x86_64 (`glibc 2.32`)
- [x] x86_64 (`glibc 2.27`) 
- [x] armv7l (`glibc 2.27`)
- [ ] armv7l (`glibc 2.32`) (NOT TESTED)

### DIFF
--- a/packages/glibc.rb
+++ b/packages/glibc.rb
@@ -4,6 +4,7 @@ class Glibc < Package
   description 'The GNU C Library project provides the core libraries for GNU/Linux systems.'
   homepage 'https://www.gnu.org/software/libc/'
   license 'LGPL-2.1+, BSD, HPND, ISC, inner-net, rc, and PCRE'
+  compatibility 'all'
 
   depends_on 'gawk' => :build
   depends_on 'libgd' => :build
@@ -17,7 +18,6 @@ class Glibc < Package
   case LIBC_VERSION
   when '2.23'
     version LIBC_VERSION
-    compatibility 'i686'
     source_url 'https://ftpmirror.gnu.org/glibc/glibc-2.23.tar.xz'
     source_sha256 '94efeb00e4603c8546209cefb3e1a50a5315c86fa9b078b6fad758e187ce13e9'
 
@@ -29,7 +29,6 @@ class Glibc < Package
     })
   when '2.27'
     version LIBC_VERSION
-    compatibility 'x86_64 aarch64 armv7l'
     source_url 'https://ftpmirror.gnu.org/glibc/glibc-2.27.tar.xz'
     source_sha256 '5172de54318ec0b7f2735e5a91d908afe1c9ca291fec16b5374d9faadfc1fc72'
 
@@ -45,7 +44,6 @@ class Glibc < Package
     })
   when '2.32' # All architectures with updates past M92.
     version '2.32-1'
-    compatibility 'x86_64 aarch64 armv7l'
     source_url 'https://ftpmirror.gnu.org/glibc/glibc-2.32.tar.xz'
     source_sha256 '1627ea54f5a1a8467032563393e0901077626dc66f37f10ee6363bb722222836'
 
--- a/packages/glibc.rb
+++ b/packages/glibc.rb
@@ -4,19 +4,23 @@ class Glibc < Package
   description 'The GNU C Library project provides the core libraries for GNU/Linux systems.'
   homepage 'https://www.gnu.org/software/libc/'
   license 'LGPL-2.1+, BSD, HPND, ISC, inner-net, rc, and PCRE'
-  compatibility 'all'
 
   depends_on 'gawk' => :build
   depends_on 'libgd' => :build
   depends_on 'libidn2' => :build
   depends_on 'texinfo' => :build
+  depends_on 'hashpipe' => :build
 
   # enable-obsolete-rpc should NOT be included as a build option as
   # this functionality is provided by libtirpc
 
-  case LIBC_VERSION
+  @LIBC_VERSION = LIBC_VERSION
+  # Set the following to build newer glibc on an older architecture
+  # @LIBC_VERSION = '2.32'
+  case @LIBC_VERSION
   when '2.23'
-  version LIBC_VERSION
+    version LIBC_VERSION
+    compatibility 'i686'
     source_url 'https://ftpmirror.gnu.org/glibc/glibc-2.23.tar.xz'
     source_sha256 '94efeb00e4603c8546209cefb3e1a50a5315c86fa9b078b6fad758e187ce13e9'
 
@@ -27,7 +31,8 @@ class Glibc < Package
       i686: '52145b65cb49c2751f69d4c46636f0685f2abb6685d8a080b71b2f091595a950'
     })
   when '2.27'
-  version LIBC_VERSION
+    version LIBC_VERSION
+    compatibility 'x86_64 aarch64 armv7l'
     source_url 'https://ftpmirror.gnu.org/glibc/glibc-2.27.tar.xz'
     source_sha256 '5172de54318ec0b7f2735e5a91d908afe1c9ca291fec16b5374d9faadfc1fc72'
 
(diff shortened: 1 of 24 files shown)


packages/glibc.rb
```diff
@@ -41,19 +42,22 @@ class Glibc < Package
...
-  })
-  binary_sha256({
-    aarch64: '1fb970e490c988e49672e3e5d1edcaea795323e4fbb74559dea3bd35d7e6c63c',
-    armv7l: '1fb970e490c988e49672e3e5d1edcaea795323e4fbb74559dea3bd35d7e6c63c',
-    x86_64: '794745d17eae977dccf6d5635f67e5bb1465286a3888bc5e69eda487e87654ae'
-  })
+    binary_url({
+      aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/glibc/2.27_armv7l/glibc-2.27-chromeos-armv7l.tar.xz',
+       armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/glibc/2.27_armv7l/glibc-2.27-chromeos-armv7l.tar.xz',
+       x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/glibc/2.27_x86_64/glibc-2.27-chromeos-x86_64.tar.xz'
```
uberhacker, line 54: We're changing from 2.32 to 2.27?  Was this intended?

packages/glibc.rb
```diff
@@ -263,9 +283,35 @@ def self.install
...
+      if LIBC_VERSION == '2.32'
+        FileUtils.ln_sf "/#{ARCH_LIB}/libresolv-#{LIBC_VERSION}.so", 'libresolv.so.2'
+        FileUtils.ln_sf "/#{ARCH_LIB}/libc.so.6", '.'
+        case ARCH
+        when 'armv7l', 'aarch64'
+          FileUtils.ln_sf "/#{ARCH_LIB}/ld-linux-armhf.so.3", 'ld-linux-armhf.so.3'
+        when 'x86_64'
+          FileUtils.ln_sf "/#{ARCH_LIB}/ld-linux-x86-64.so.2", 'ld-linux-x86-64.so.2'
+        end
+      end
```
uberhacker, line 314: Could the symlinks fix issues with the 2.32 binaries?

saltedcoffii:
Have you tried looking at the gentoo and chromeos patches for glibc? Those may get it working

packages/glibc.rb
satmandu: Yes. The crt files used for compiles in the globe 2.27 package don't appear to pull in `libgcc-s.so.1` as a dependency.

satmandu:
> Have you tried looking at the gentoo and chromeos patches for glibc? Those may get it working

That's already what is being done.

packages/glibc.rb
satmandu: They do not. I think the issue is with the crt binaries bring generated.

satmandu: They do not. I think the issue is with the crt binaries bring generated.

satmandu:
Copying /usr/lib64/libc.so to #CREW_LIB_PREFIX and changing the path inside for libc_nonshared.a needs to be a part of this too... 2.32 may be a frankenglibc..

packages/glibc.rb
```diff
@@ -3,20 +3,24 @@
...
   depends_on 'libgd' => :build
   depends_on 'libidn2' => :build
   depends_on 'texinfo' => :build
+  depends_on 'hashpipe' => :build
 
   # enable-obsolete-rpc should NOT be included as a build option as
   # this functionality is provided by libtirpc
 
-  case LIBC_VERSION
+  @LIBC_VERSION = LIBC_VERSION
```
uberhacker, line 17: Shouldn't need to do this since `LIBC_VERSION` is a constant.

packages/glibc.rb
```diff
@@ -76,6 +85,17 @@ def self.patch
...
+      # Patch to enable build-local-archive
+      system 'curl -Ls https://src.fedoraproject.org/rpms/glibc/raw/f30/f/glibc-fedora-locarchive.patch | \
+      hashpipe sha256 0acccf57d8c6e7de82871c61ccb845f7a1ae13ae1fbc65995d347de8367c7bb1 | \
+      patch -Np1 --binary'
+      system 'curl -Ls https://src.fedoraproject.org/rpms/glibc/raw/f30/f/build-locale-archive.c | \
+      hashpipe sha256 6834e8b8f2a987bf8adfd265c0e01665f102c7115db94b99ec36376b68e9d3fa > fedora/build-locale-archive.c'
+      system "sed -i 's,/lib/locale,/lib#{CREW_LIB_SUFFIX}/locale,g' fedora/build-locale-archive.c"
+      system "sed -i 's,/usr/sbin/tzdata-update,/bin/true,g' fedora/build-locale-archive.c"
+      system "sed -i 's,verbose,locale_verbose,g' fedora/build-locale-archive.c"
+      system "sed -i 's,be_quiet,locale_be_quiet,g' fedora/build-locale-archive.c"
```
uberhacker, line 98: I feel like we're making this way too complicated.  Why are we using the fedora sources?

satmandu: See one rationale here, via another project using the same tool: https://github.com/Mrm3/image-build/blob/4a878611f813f1a859e48747ecf68897e7dd9015/diskimage_builder/elements/yum-minimal/pre-install.d/03-yum-cleanup#L38

In short, using the Fedora tool lets us avoid this:

image

Also, we get the benefit of users being able to modify the postinstall locally to add any other locales they might need with a quick reinstall of glibc.

packages/glibc.rb
satmandu: Yes, but the plumbing for this is just in there for bootstrapping builds of newer glibc on older glibc. Otherwise it's a huge pain.

satmandu: As suggested, only using LIBC_VERSION.

packages/glibc.rb
```diff
@@ -3,20 +3,21 @@
 class Glibc < Package
   description 'The GNU C Library project provides the core libraries for GNU/Linux systems.'
   homepage 'https://www.gnu.org/software/libc/'
-  version LIBC_VERSION
   license 'LGPL-2.1+, BSD, HPND, ISC, inner-net, rc, and PCRE'
-  compatibility 'all'
```
uberhacker, line 8: I think you can leave these lines.

packages/glibc.rb
```diff
@@ -3,20 +3,21 @@
...
   depends_on 'texinfo' => :build
+  depends_on 'hashpipe' => :build
 
   # enable-obsolete-rpc should NOT be included as a build option as
   # this functionality is provided by libtirpc
 
   case LIBC_VERSION
   when '2.23'
+    version LIBC_VERSION
+    compatibility 'i686'
```
uberhacker, line 20: These lines don't need to be here.

packages/glibc.rb
```diff
@@ -27,6 +28,8 @@ class Glibc < Package
       i686: '52145b65cb49c2751f69d4c46636f0685f2abb6685d8a080b71b2f091595a950'
     })
   when '2.27'
+    version LIBC_VERSION
+    compatibility 'x86_64 aarch64 armv7l'
```
uberhacker, line 32: These lines don't need to be here.

packages/glibc.rb
```diff
@@ -41,19 +44,22 @@ class Glibc < Package
        x86_64: '5fe94642dbbf900d22b715021c73ac1a601b81517f0da1e7413f0af8fbea7997'
     })
   when '2.32' # All architectures with updates past M92.
+    version '2.32-1'
+    compatibility 'x86_64 aarch64 armv7l'
```
uberhacker, line 48: These lines don't need to be here.  Since release 92 is not stable yet, I think we can get away with not incrementing the version.

packages/glibc.rb
satmandu: Versioning should be per Glibc version, no?

Otherwise Glibc 2.27 users will get update suggestions when we update Glibc 2.32...

packages/glibc.rb
satmandu: I'm not sure we should set a precedent of not incrementing a deployed version?

It's one thing to not officially support non-release channel, it's another to intentionally break non-release channel by not showing appropriate updates.

(Also, with the M91 breakage this week, it has been often suggested that people move to beta channel, so we might have users with M92 who need to get the newer build of Glibc 2.32.)

packages/glibc.rb
uberhacker: That's what I'm saying.  `version LIBC_VERSION` can be set at the top.  The way I envision it working is `LIBC_VERSION` is what drives the decision and not compatibility.  That can be set at the top to all.

packages/glibc.rb
uberhacker: OK, I think we can keep the version down in the case block because 2.32-1, although I think we could get away with not incrementing since it hasn't hit the stable branch yet.

packages/glibc.rb
uberhacker: It only sets a bad precedent if it's on the stable branch.  Since it isn't, I think we can get away with not incrementing.  I can merge this but you could clean this up on the next update by going back to just 2.32 for the version and set it at the top.

PR #5953, opened 2021-07-12
Add natron package

uberhacker:
Open Source Compositing Software For VFX and Motion Graphics.  See https://natrongithub.github.io/.  Tested on x86_64.

### DIFF
--- a/packages/natron.rb
+++ b/packages/natron.rb
@@ -1,0 +1,57 @@
+require 'package'
+
+class Natron < Package
+  description 'Open Source Compositing Software For VFX and Motion Graphics.'
+  homepage 'https://natrongithub.github.io/'
+  version '2.4.0'
+  license 'GPL-2'
+  compatibility 'x86_64'
+  source_url 'https://github.com/NatronGitHub/Natron/releases/download/v2.4.0/Natron-2.4.0-Linux-64.tgz'
+  source_sha256 '7e8f2ec343f553799f34dce89aa250c410024f17e2c9ccfb5e22544db3e46bb4'
+
+  binary_url ({
+    x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/natron/2.4.0_x86_64/natron-2.4.0-chromeos-x86_64.tpxz'
+  })
+  binary_sha256 ({
+    x86_64: '35b3c22b67f74d285e71f619179cf36a3a475d0df37dd7dcb104d22a318edce4'
+  })
+
+  depends_on 'desktop_file_utils'
+  depends_on 'libglu'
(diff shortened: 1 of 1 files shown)


packages/natron.rb
```diff
@@ -0,0 +1,55 @@
...
+    shrink = ENV['CREW_SHRINK_ARCHIVE']
+    ENV['CREW_NOT_STRIP'] = '1'
+    ENV['CREW_SHRINK_ARCHIVE'] = '0'
+    # IMPORTANT BUILD/INSTALL FROM SOURCE INSTRUCTIONS BELOW:
+    # Enter /usr/local/tmp/crew/dest/usr/local/share/Natron-#{version} for the install directory.
+    # Uncheck 'Run Natron now' at the end of the install wizard.
+    system "./Natron-#{version}-Linux-64"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
+    FileUtils.ln_s "#{CREW_PREFIX}/share/Natron-#{version}/Natron", "#{CREW_DEST_PREFIX}/bin/natron"
+    ENV['CREW_SHRINK_ARCHIVE'] = shrink
```
satmandu, line 34: I think the order of operations here is:
`build_and_preconfigure`
`@pkg.install` [run `def self.install`]
`prepare_package`
`strip_dir` [check for `CREW_NOT STRIP` here]

So if we set  reset the env variable via `ENV['CREW_SHRINK_ARCHIVE'] = shrink` inside `def self.install` that's going to run `strip_dir` based upon the original setting, no? (I don't think that's what you want?)

packages/natron.rb
```diff
@@ -0,0 +1,57 @@
...
+  ENV['CREW_NOT_STRIP'] = '1'
+  ENV['CREW_SHRINK_ARCHIVE'] = '0'
+
+  def self.install
+    # IMPORTANT BUILD/INSTALL FROM SOURCE INSTRUCTIONS BELOW:
+    # Enter /usr/local/tmp/crew/dest/usr/local/share/Natron-#{version} for the install directory.
+    # Uncheck 'Run Natron now' at the end of the install wizard.
+    system "./Natron-#{version}-Linux-64"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
+    FileUtils.ln_s "#{CREW_PREFIX}/share/Natron-#{version}/Natron", "#{CREW_DEST_PREFIX}/bin/natron"
```
supechicken, line 34: Use relative symlink?
```
../share/Natron-#{version}/Natron
```

uberhacker: I prefer the absolute symlink for clarity.

PR #5954, opened 2021-07-12
Disallow Crostini and non-stable channel device from install

supechicken:
Tested on `armv7l`


install.sh
```diff
@@ -3,6 +3,9 @@
 # exit on fail
 set -e
 
+# load system info
+source /etc/lsb-release
```
uberhacker, line 7: Do we know if `lsb-release` is available on all Chrom(e/ium)OS-like environments?  I haven't checked Cloudready recently but there are others like Gallium that might have a different file or location.  Maybe we should check if it exists first?

install.sh
```diff
@@ -33,6 +36,29 @@ MAGENTA='\e[1;35m';
 RESET='\e[0m'
 
 echo -e "${GREEN}Welcome to Chromebrew!${RESET}\n"
+
+# disallow non Chrome OS devices
+if [ -z "${CHROMEOS_RELEASE_VERSION}" ]; then
+  echo -e "${RED}Only Chrome OS developer shell is supported by Chromebrew :/${RESET}"
+  exit 1
+fi
```
uberhacker, line 44: This could end up locking out a lot of users if `CHROMEOS_RELEASE_VERSION` is not set.  Could we make a better assumption?

install.sh
```diff
@@ -33,6 +36,29 @@ MAGENTA='\e[1;35m';
...
+
+# disallow non-stable channels Chromr OS
+if [ "${CHROMEOS_RELEASE_TRACK}" != 'stable-channel' ]; then
+  echo -e "${YELLOW}The beta, dev, and canary channel are unsupported by Chromebrew"
+  echo -e "Install anyway? [Y/n]"
+  echo -e "(This may cause major issues with your Chromebrew installation)${RESET}"
+  # `<&1`: tell bash read from console not stdin (curl)
+  read -n1 response <&1
```
uberhacker, line 52: This might hinder automated installs.  Have you tried to see what happens?

install.sh
```diff
@@ -33,6 +36,29 @@ MAGENTA='\e[1;35m';
...
+if [ -z "${CHROMEOS_RELEASE_VERSION}" ]; then
+  echo -e "${RED}Only Chrome OS developer shell is supported by Chromebrew :/${RESET}"
+  exit 1
+fi
+
+# disallow non-stable channels Chromr OS
+if [ "${CHROMEOS_RELEASE_TRACK}" != 'stable-channel' ]; then
+  echo -e "${YELLOW}The beta, dev, and canary channel are unsupported by Chromebrew"
+  echo -e "Install anyway? [Y/n]"
+  echo -e "(This may cause major issues with your Chromebrew installation)${RESET}"
```
uberhacker, line 50: It might be better to swap the 2 lines above so the cursor isn't below everything.

satmandu: We should also have a way to override this, with an env variable or whatever. We want to be able to dogfood on testing channels ourselves without huge hacks.

supechicken: Replaced `read` to variable

install.sh
supechicken: This variable is set by `source /etc/lsb_release`, it is set unless non chrome os system or file not exist

install.sh
supechicken: Now it will check existent of crosh command if lsb-release is not found

install.sh
```diff
@@ -33,6 +33,30 @@ MAGENTA='\e[1;35m';
 RESET='\e[0m'
 
 echo -e "${GREEN}Welcome to Chromebrew!${RESET}\n"
+
+# load system info
+if [ -f /etc/lsb-release ]; then
+  source /etc/lsb-release
+else
+  echo -e "${YELLOW}Unable to detect system information, installation will continue.${RESET}"
+  CHROMEOS_RELEASE_VERSION=0.0-00000
```
uberhacker, line 42: You could probably use just `CHROMEOS_RELEASE_VERSION=0`. :)

install.sh
```diff
@@ -33,6 +33,30 @@ MAGENTA='\e[1;35m';
...
+fi
+
+# disallow non Chrome OS devices
+if [ -z "${CHROMEOS_RELEASE_VERSION}" -a "${CREW_FORCE_INSTALL}" != '1' ]; then
```
uberhacker, line 46: This will never be true, correct?

daemonspudguy:
I feel that this might end up breaking everything if, say, Google accidentally pushes pre-release software into Stable and retracts it again.

install.sh
supechicken: It means if both CHROMEOS_RELEASE (check if it is chrome os) and FORCE_INSTALL (skip switch) not set then exit

install.sh

supechicken:
> I feel that this might end up breaking everything if, say, Google accidentally pushes pre-release software into Stable and retracts it again.

If so it is a serious mistake :) I mean it is nearly impossible to happen in a such big company

install.sh
uberhacker: The logic above always sets `CHROMEOS_RELEASE_VERSION`, right?

satmandu:
Possibly related: https://github.com/skycocker/chromebrew/issues/5962

daemonspudguy:
> If so it is a serious mistake :) I mean it is nearly impossible to happen in a such big company

You would think, but they have done it in the past. Remember the Chrome OS 81 fiasco back in April? Pepperidge Farm remembers. (sorry, I had to)

install.sh
supechicken: The script only check existent of crosh now

supechicken:
> You would think, but they have done it in the past. Remember the Chrome OS 81 fiasco back in April? Pepperidge Farm remembers. (sorry, I had to)

But we can’t fetch the release note of Chrome OS to make sure that is a regular release (we can’t extract the latest version for each device), also chromebrew may working fine with that 'mistake' version

supechicken:
Tested chroot:
  [x] armv7l (`kukui`)
  [x] i686 (`alex`)
  [x] x86_64 (`eve`)

supechicken:
I will merge this pr with #5967 later

PR #5955, opened 2021-07-13
Update codium from 1.57.1 to 1.58.0

uberhacker:
Tested on:
- [x] x86_64
- [x] aarch64
- [x] armv7l

The symlink command removes libcom_err.so.3 which defeats the whole purpose of creating the libcom_err package.  So we need to disable `CREW_SHRINK_ARCHIVE`.  Also disabled for this package since we don't have binaries and upx takes an eternity to compress.  We don't need to restore `CREW_SHRINK_ARCHIVE` to the original value because `ENV` doesn't export.

### DIFF
--- a/packages/libcom_err.rb
+++ b/packages/libcom_err.rb
@@ -11,6 +11,7 @@ class Libcom_err < Package
   depends_on 'e2fsprogs'
 
   def self.install
+    ENV['CREW_SHRINK_ARCHIVE'] = '0'
     FileUtils.mkdir_p CREW_DEST_LIB_PREFIX
     Dir.chdir CREW_DEST_LIB_PREFIX do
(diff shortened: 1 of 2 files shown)


packages/libcom_err.rb
```diff
@@ -11,6 +11,7 @@ class Libcom_err < Package
   depends_on 'e2fsprogs'
 
   def self.install
+    ENV['CREW_SHRINK_ARCHIVE'] = '0'
```
satmandu, line 14: Any reason for this here? (and why not put this in arch specific CASE if it is i686 specific?)

uberhacker: Yes, because symlink deletes what I want this package to contain. :)

PR #5957, opened 2021-07-13
Ruby 3.0.2

satmandu:
- Security fixes therein...
- needs `i686` build added before merge.

Works properly:
- [x] x86_64
- [x] armv7l


uberhacker:
Added i686 binary.

PR #5959, opened 2021-07-14
libcom_err: add binaries

satmandu:
- avoids the need for going through the install from source functions.

Works properly:
- [x] x86_64
- [x] i686 (probably!)
- [x] armv7l

### DIFF
--- a/packages/libcom_err.rb
+++ b/packages/libcom_err.rb
@@ -3,18 +3,30 @@ require 'package'
 class Libcom_err < Package
   description 'Common error description library'
   homepage 'http://e2fsprogs.sourceforge.net/'
-  version '1.0'
+  version '1.1'
   license 'GPL-2'
   compatibility 'all'
   source_url 'SKIP'
 
+  binary_url({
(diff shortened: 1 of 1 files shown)


uberhacker:
I don't think we need binaries for a single symlink.  What is going to happen next time we build if the libcom_err.so version changes?  Don't we have the same situation?

satmandu:
We have the same situation with a libcom_err.so version change, sure... but this is faster. The entire package file is less than 1 k, and installing that is faster than going through the logic of doing the install process.

uberhacker:
Shouldn't we leave the `ENV['CREW_SHRINK_ARCHIVE'] = '0'` line in there then?

satmandu:
It's no longer needed. the symlinks app doesn't mess things up if the symlink is to a file in the same directory. Sure it's dangling, but the symlink stays.

uberhacker:
A lot of effort to save 10 sec. :)

satmandu:
Multiply that ten seconds by the number of users and installs... It adds up. 😂

PR #5960, opened 2021-07-14
Don't let crew break if CREW_PREFIX is set.

satmandu:
If `CREW_PREFIX` is set, `crew` breaks, complaining about `@pkg.build_from_source` being set in `const.rb`, with a `/usr/local/lib/crew/lib/const.rb:20:in `<top (required)>': undefined method `build_from_source=' for nil:NilClass (NoMethodError)
` error.

image

Works properly:
- [x] x86_64

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.11.9'
+CREW_VERSION = '1.12.0'
 
 ARCH_ACTUAL = `uname -m`.strip
 # This helps with virtualized builds on aarch64 machines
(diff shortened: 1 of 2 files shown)


lib/const.rb
```diff
@@ -13,11 +13,11 @@
 @libcvertokens=  %x[/#{ARCH_LIB}/libc.so.6].lines.first.chomp.split(/[\s]/)
 LIBC_VERSION = @libcvertokens[@libcvertokens.find_index("version") + 1].sub!(/[[:punct:]]?$/,'')
 
-if ENV['CREW_PREFIX'].to_s.empty?
+if ENV['CREW_PREFIX'].to_s.empty? or ENV['CREW_PREFIX'] == '/usr/local'
```
uberhacker, line 16: I fail to see how this logic solves the problem.  It should still be set in the else block if the environment variable is set.

satmandu: Case 1: CREW_PREFIX is empty
Case 2: CREW_PREFIX = /usr/local
Case 3: CREW_PREFIX is something else

If 1 or 2, set CREW_PREFIX=/usr/local
if 3, leave CREW_PREFIX alone.

Currently Case 2 or 3 breaks `crew`.

uberhacker: Shouldn't case 2 and 3 still work with the old logic?

satmandu: Case 2 doesn't work. Try it: `CREW_PREFIX=/usr/local crew const CREW_PREFIX`

uberhacker: The only difference with the original logic is it doesn't set `CREW_BUILD_FROM_SOURCE = 1` but I'm thinking better logic would be the following?
```
CREW_PREFIX = '/usr/local'
if ENV['CREW_PREFIX'] and ENV['CREW_PREFIX'] != CREW_PREFIX
  CREW_BUILD_FROM_SOURCE = 1
end
```

satmandu: Sounds good to me, and it appears to work too!

PR #5961, opened 2021-07-14
Add adobe_source_code_pro_fonts

satmandu:
- Adds monospace font needed for gnome-terminal.

- The single install line is much cleaner than ruby-fying it, no?

Works properly:
- [x] x86_64

Packages properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/adobe_source_code_pro_fonts.rb
+++ b/packages/adobe_source_code_pro_fonts.rb
@@ -1,0 +1,37 @@
+# Adapted from Arch Linux adobe-source-code-pro-fonts PKGBUILD at:
+# https://github.com/archlinux/svntogit-packages/raw/packages/adobe-source-code-pro-fonts/trunk/PKGBUILD
+
+require 'package'
+
+class Adobe_source_code_pro_fonts < Package
+  description 'Monospaced font family for user interface and coding environments'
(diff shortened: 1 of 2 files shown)


packages/adobe_source_code_pro_fonts.rb
```diff
@@ -0,0 +1,37 @@
...
+    aarch64: '22858b9d32b3911f42ce9d38d762a0bda58500dc936bec1a6d15600c3918c27b',
+     armv7l: '22858b9d32b3911f42ce9d38d762a0bda58500dc936bec1a6d15600c3918c27b',
+       i686: 'f5675e683e54148c6cb6a731913b43ac1be36f20b95828398618dcb7bce04762',
+     x86_64: '572b83ce1c1370f590800a253b5b69dac29953f3dd29e5177ac658894987d265'
+  })
+
+  depends_on 'fontconfig'
+
+  def self.install
+    system "install -Dt #{CREW_DEST_PREFIX}/share/fonts/adobe-source-code-pro-fonts -m644 OTF/*.otf VAR/*.otf"
```
uberhacker, line 31: I like using the `install` command with system because it creates the directory if it doesn't exist.

PR #5965, opened 2021-07-15
Add core to buildessential

satmandu:
Older installs may not have installed core, so make sure at least `buildessential` has a core dependency.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/buildessential.rb
+++ b/packages/buildessential.rb
@@ -3,11 +3,14 @@ require 'package'
 class Buildessential < Package
   description 'A collection of tools essential to compile and build software.'
   homepage ''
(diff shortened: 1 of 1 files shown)


packages/buildessential.rb
```diff
@@ -3,12 +3,15 @@
...
   description 'A collection of tools essential to compile and build software.'
   homepage ''
-  version '1.17'
+  version '1.18'
   license 'GPL-3+'
   compatibility 'all'
 
   is_fake
 
+  # Makre sure core is installed
```
uberhacker, line 12: Make?

PR #5971, opened 2021-07-19
Update brave from 1.28.33 to 1.29.8

uberhacker:
Tested on x86_64.

### DIFF
--- a/packages/brave.rb
+++ b/packages/brave.rb
@@ -3,17 +3,19 @@ require 'package'
 class Brave < Package
   description 'Next generation Brave browser for macOS, Windows, Linux, Android.'
   homepage 'https://brave.com/'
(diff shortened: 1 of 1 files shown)


packages/brave.rb
```diff
@@ -3,17 +3,19 @@
 class Brave < Package
   description 'Next generation Brave browser for macOS, Windows, Linux, Android.'
   homepage 'https://brave.com/'
-  version '1.28.33'
```
supechicken, line 6: Seems this is a nightly version, can we revert back to the latest stable?

uberhacker: The last version was nightly too.

uberhacker: I reverted to pre-release 1.27.106.

PR #5977, opened 2021-07-23
Adjust dbus deps & don't require for non-X11 installs

### LINKED ISSUE #5976: If X and other things aren't installed, a new shell raises errors.
```
$ source ~/.bashrc
dbus-launch: error while loading shared libraries: libSM.so.6: cannot open shared object file: No such file or directory
dbus-update-activation-environment: error: unable to connect to D-Bus: Unable to autolaunch a dbus-daemon without a $DISPLAY for X11
```

```
$ crew list installed
acl
attr
autoconf
automake
bash
binutils
bison
brotli
bz2
c_ares
ca_certificates
cmake
compressdoc
crew_profile_base
dbus
diffutils
docbook_xml412
docbook_xml42
docbook_xml43
docbook_xml44
docbook_xml45
docbook_xml50
docbook_xml51
docbook_xsl
docbook_xsl_nons
doxygen
e2fsprogs
elfutils
expat
filecmd
flex
fop
gcc11
gcc_tools
gdbm
gettext
git
glibc
gmp
gnutls
groff
gtk_doc
help2man
icu4c
intltool
isl
itstool
jdk8
jq
krb5
less
libarchive
libcyrussasl
libdb
libedit
libffi
libgcrypt
libgpgerror
libidn2
libmetalink
libnghttp2
libpipeline
libpsl
libseccomp
libsigsegv
libssp
libtasn1
libtirpc
libtool
libunbound
libunistring
libxml2
libxslt
libyaml
linuxheaders
llvm
lz4
lzip
m4
make
mandb
manpages
mawk
meson
most
mpc
mpfr
ncurses
nettle
ninja
openldap
openssl
p11kit
patch
pcre
pcre2
perl
perl_locale_messages
perl_text_unidecode
perl_unicode_eastasianwidth
perl_xml_parser
pixz
pkgconfig
py3_libxml2
py3_nuitka
py3_pip
py3_setuptools
py3_wheel
python2
python3
rdfind
readline
rtmpdump
ruby
samurai
sed
slang
sqlite
symlinks
texinfo
trousers
uchardet
ucl
unzip
upx
util_macros
valgrind
xmlcatmgr
xzutils
zip
zlibpkg
zstd
```

Perhaps the dbus fixes should be added to a chromebrew package and not to crew_profile_base itself.

satmandu:
Fixes #5976 

- Dbus deps are now appropriately enumerated
- Moves the dbus env.d file to the dbus package.
- This keeps dbus from pulling in a bunch of X11 deps on a non-X11 install of chromebrew
- I am not updating dbus binaries since I don't want to create a broken i686 binary (hence the use of postinstall.)
- https://github.com/chromebrew/crew-profile-base can be updated in a separate PR.
- Thanks to @saltedcoffii for noticing this issue and pointing it out.

Works properly:
- [x] x86_64

### DIFF
--- a/packages/crew_profile_base.rb
+++ b/packages/crew_profile_base.rb
@@ -8,7 +8,6 @@ class Crew_profile_base < Package
   compatibility 'all'
   source_url 'SKIP'
 
-  depends_on 'dbus'
 
   def self.install
     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/etc/env.d"
@@ -45,6 +44,10 @@ class Crew_profile_base < Package
       # Remove stale files from the last install
       FileUtils.rm "#{HOME}/.bashrc.bak" if File.exist? "#{HOME}/.bashrc.bak"
       FileUtils.rm "#{HOME}/.zshrc.bak" if File.exist? "#{HOME}/.zshrc.bak"
+
+      # dbus file moved to dbus package, so remove it.
+      FileUtils.rm_f './src/env.d/env.d/03-dbus'
+      FileUtils.rm_f "#{CREW_PREFIX}/etc/env.d/03-dbus"
 
       # Don't overwrite a custom shell rc
       @_str = "source #{CREW_PREFIX}/etc/profile"
--- a/packages/dbus.rb
+++ b/packages/dbus.rb
@@ -3,37 +3,66 @@ require 'package'
 class Dbus < Package
   description 'D-Bus is a message bus system, a simple way for applications to talk to one another.'
   homepage 'https://www.freedesktop.org/wiki/Software/dbus/'
-  version '1.13.18'
+  version '1.13.18-1'
   license 'Apache-2.0'
   compatibility 'all'
   source_url 'https://dbus.freedesktop.org/releases/dbus/dbus-1.13.18.tar.xz'
   source_sha256 '8078f5c25e34ab907ce06905d969dc8ef0ccbec367e1e1707c7ecf8460f4254e'
 
-  binary_url ({
-    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/dbus/1.13.18_armv7l/dbus-1.13.18-chromeos-armv7l.tar.xz',
-     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/dbus/1.13.18_armv7l/dbus-1.13.18-chromeos-armv7l.tar.xz',
-       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/dbus/1.13.18_i686/dbus-1.13.18-chromeos-i686.tar.xz',
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/dbus/1.13.18_x86_64/dbus-1.13.18-chromeos-x86_64.tar.xz',
+  binary_url({
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/dbus/1.13.18-1_armv7l/dbus-1.13.18-1-chromeos-armv7l.tpxz',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/dbus/1.13.18-1_armv7l/dbus-1.13.18-1-chromeos-armv7l.tpxz',
+       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/dbus/1.13.18-1_i686/dbus-1.13.18-1-chromeos-i686.tpxz',
+     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/dbus/1.13.18-1_x86_64/dbus-1.13.18-1-chromeos-x86_64.tpxz'
   })
-  binary_sha256 ({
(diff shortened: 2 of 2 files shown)


packages/dbus.rb
```diff
@@ -36,4 +39,20 @@ def self.build
...
+      [[ "\$DBUS_SESSION_BUS_ADDRESS"=="disabled:" ]] && unset DBUS_SESSION_BUS_ADDRESS
+      if [ -z "\${DBUS_SESSION_BUS_ADDRESS}" ]; then
+        # if not found, launch a new one
+        eval `dbus-launch --sh-syntax`
+        # echo "D-Bus per-session daemon address is: \$DBUS_SESSION_BUS_ADDRESS"
+      fi
+      dbus-update-activation-environment --all
+    DBUSCONFIGEOF
+    IO.write("#{CREW_PREFIX}/etc/env.d/dbus", @dbusconfigenv)
+  end
```
uberhacker, line 57: This is definitely better to put in install rather than postinstall.  Of course, that would require new binaries.

satmandu: Let's do that then.

satmandu: Could you rebuild for i686?

satmandu: Actually, this needs work. I get a `libaudit.h` not found error on x86_64... Might have to turn off selinux, and maybe we should tweak the runstatedir setting too...

satmandu: Compilation fixed. If you can get an i686 build done @uberhacker. I just need to get my armv7l server connected again before I can do that build.

packages/crew_profile_base.rb
```diff
@@ -46,6 +45,10 @@ def self.postinstall
       FileUtils.rm "#{HOME}/.bashrc.bak" if File.exist? "#{HOME}/.bashrc.bak"
       FileUtils.rm "#{HOME}/.zshrc.bak" if File.exist? "#{HOME}/.zshrc.bak"
 
+      # dbus file moved to dbus postinstall, so remove it.
+      FileUtils.rm_f './src/env.d/env.d/03-dbus'
+      FileUtils.rm_f "#{CREW_PREFIX}/etc/env.d/03-dbus"
+
```
uberhacker, line 51: Do we need this since the environment has been moved into install instead of postinstall?

packages/dbus.rb
```diff
@@ -3,37 +3,60 @@
...
-     x86_64: 'c777f379f48d88844a5f9eb1f9e38174df8645c6e76845ab0fc8f1ee58034c52',
+  binary_sha256({
+    x86_64: '3a459724ba6e7f14eacc089fb5500b20b7857aa3034cfc58edf2450ad8ac1c02'
   })
 
-  depends_on 'expat'
+  depends_on 'expat' # R
+  depends_on 'libice' # R
+  depends_on 'libsm' # R
+  depends_on 'libx11' # R
```
uberhacker, line 22: We need libx11 for dbus?  :face_with_head_bandage:

packages/crew_profile_base.rb
satmandu: This will clean up the older installed dbus file to avoid running it twice.

packages/dbus.rb
satmandu: oh yeah. Check out the packages which depend upon it.

packages/crew_profile_base.rb
uberhacker: This file wasn't installed by dbus initially? :face_with_head_bandage:

packages/dbus.rb
```diff
@@ -3,37 +3,60 @@
...
+        eval `dbus-launch --sh-syntax`
+        # echo "D-Bus per-session daemon address is: \$DBUS_SESSION_BUS_ADDRESS"
+      fi
+      dbus-update-activation-environment --all
+    DBUSCONFIGEOF
+    IO.write("#{CREW_DEST_PREFIX}/etc/env.d/dbus", @dbusconfigenv)
+  end
+
+  def self.postinstall
+    FileUtils.rm_f "#{CREW_PREFIX}/etc/env.d/03-dbus"
```
uberhacker, line 60: I don't understand.  We're installing it and then removing it again in postinstall?

packages/crew_profile_base.rb
satmandu: This is cleaning up the env.d/03-dbus file from crew_profile_base

packages/dbus.rb
satmandu: (or not... if none of the packages which depend upon dbus work on i686... maybe it doesn't matter on there?)

packages/dbus.rb
satmandu: removing the file from crew_profile_base if it exists.

packages/crew_profile_base.rb
```diff
@@ -46,6 +45,10 @@ def self.postinstall
       FileUtils.rm "#{HOME}/.bashrc.bak" if File.exist? "#{HOME}/.bashrc.bak"
       FileUtils.rm "#{HOME}/.zshrc.bak" if File.exist? "#{HOME}/.zshrc.bak"
 
+      # dbus file moved to dbus postinstall, so remove it.
```
uberhacker, line 48: Moved to install, not postinstall?

PR #5985, opened 2021-07-25
🎉 bsdgames 🎉 (2)

saltedcoffii:
For some reason, the first bsdgames pull request didn't actually add any games when it was merged. This PR actually adds the games.

This is also a proof of concept for package files that share a lot of similar material. I plan on expanding this concept to common build systems, namely meson, cmake and autotools. This would significantly shrink the Chromebrew repository and standardize Chromebrew packages.

Because this is a POC, I am looking for feedback on the mechanisms. I'd really really like to avoid using the global variables (variables starting with $) but I couldn't find another way to do it. If anyone knows of another way, please let me know. 

This also fixes a small bug where, when caching, if a `git_hashtag` contains a `/`, the caching code fails with an error from `tar`.


bin/crew
```diff
@@ -708,7 +708,7 @@ def download
       # Recall repository from cache if requested
       if CREW_CACHE_ENABLED
         if @pkg.git_branch.nil? || @pkg.git_branch.empty?
-          cachefile = CREW_CACHE_DIR + filename + @pkg.git_hashtag + '.tar.xz'
+          cachefile = CREW_CACHE_DIR + filename + @pkg.git_hashtag.gsub('/', '_') + '.tar.xz'
```
uberhacker, line 711: A git hashtag should only include hex characters.  Not sure why this would be needed unless the git hashtag was used to indicate an upstream/branch.  But then it shouldn't be called a hashtag in that instance.

lib/packages/libbsdgames2.rb
```diff
@@ -0,0 +1,85 @@
...
+    system "echo bsd_games_cfg_dictionary_src='#{CREW_PREFIX}/share/dict/words' >> config.params"
+    system "echo bsd_games_cfg_dm_configfile='#{CREW_PREFIX}/etc/dm/dm.conf' >> config.params"
+    system "echo bsd_games_cfg_dm_nogamesfile='#{CREW_PREFIX}/etc/dm/nogames' >> config.params"
+    system "echo bsd_games_cfg_hangman_wordsfile='#{CREW_PREFIX}/share/dict/words' >> config.params"
+    system "echo bsd_games_cfg_wtf_acronymfile='#{CREW_PREFIX}/share/misc/acryonyms' >> config.params"
+    system "echo bsd_games_cfg_ncurses_lib='$(pkg-config --libs-only-l ncurses)' >> config.params" # Fix ncurses support
+  end
+
+  def self.build
+    system "./configure"
```
uberhacker, line 74: Shouldn't we be using `#{CREW_OPTIONS}`?

lib/packages/libbsdgames2.rb
```diff
@@ -0,0 +1,85 @@
+require 'package'
+
+class Bsdgame < Package
+  $_bsdgame_homepage = 'https://www.polyomino.org.uk/computer/software/bsd-games/'
```
supechicken, line 4: Use class variable? (only can be read in relative classes)
`@@_bsdgame_homepage = '...'`

packages/adventure.rb
```diff
@@ -0,0 +1,14 @@
+require 'package'
```
supechicken, line 1: We can remove this line since it will load in `libbsdgames2.rb`

lib/packages/libbsdgames2.rb
```diff
@@ -0,0 +1,85 @@
+require 'package'
```
supechicken, line 1: We need to create a `Libbsdgames2` class otherwise user will get errors when `crew install libbsdgames2`

packages/backgammon.rb
```diff
@@ -0,0 +1,14 @@
+require 'package'
+require 'packages/libbsdgames2'
+
+class Backgammon < Bsdgame
+  description 'A text-based backgammon game; play against the computer or a friend.'
+  homepage $_bsdgame_homepage
+  version $_bsdgame_version
```
supechicken, line 7: We can set them in `libbsdgames2`

lib/packages/libbsdgames2.rb
saltedcoffii: This isn't an autotools configure script, it was written custom for bsdgames. #{CREW_OPTIONS} won't work with it.

lib/packages/libbsdgames2.rb
saltedcoffii: libbsdgames2 isn't supposed to be installed. It's supposed to provide centralized information for a bunch of other package files.

packages/backgammon.rb
saltedcoffii: If you know how, please tell me.

packages/adventure.rb
saltedcoffii: Good point.

lib/packages/libbsdgames2.rb
saltedcoffii: I tried that and it didn't work. How would you go about that?

bin/crew
saltedcoffii: A lot of tags from https://salsa.debian.org/ have this pattern.

packages/backgammon.rb
supechicken: `homepage 'https://www.polyomino.org.uk/computer/software/bsd-games/'`?
(since ruby doesn’t care the variable is set by which file)

lib/packages/libbsdgames2.rb
supechicken: Make a message for telling user about that?

packages/backgammon.rb
saltedcoffii: It's not a variable, it's a property function defined in `lib/package_helpers.rb` as far as I understand it.

saltedcoffii: It didn't work @supechicken

lib/packages/libbsdgames2.rb
saltedcoffii: Actually it looks like it's not even an issue. It'll just say that the package isn't found.
```
$ crew install libbsdgames2
Package libbsdgames2 not found. :(
$ crew install bsdgame
Package bsdgame not found. :(
```

packages/adventure.rb
saltedcoffii: 621edfc

satmandu:
Can we find a way to collapse this ALL into a single file?

Other distributions such as arch will just spit out multiple packages from a single package file, and when you have ~ 40 almost dummy packages here, it would seem that the best way to move forward would be to make that happen so that there is a single bsdgames package which is responsible for all the sub-packages.

Or... maybe just make it a single package to add or remove all of these. How much space do these all take up anyways?

saltedcoffii:
> Can we find a way to collapse this ALL into a single file?
> 
> Other distributions such as arch will just spit out multiple packages from a single package file, and when you have ~ 40 almost dummy packages here, it would seem that the best way to move forward would be to make that happen so that there is a single bsdgames package which is responsible for all the sub-packages.
> 
> Or... maybe just make it a single package to add or remove all of these. How much space do these all take up anyways?

I like your first option as it's much needed in this package manager anyway. LLVM and GCC badly need package separation, and many other packages could do well with it.

PR #5990, opened 2021-07-26
Update codium from 1.58.0 to 1.58.2

### LINKED ISSUE #5986: FHS3 enforcement causing issues on packages using outside binaries such as vscodium
The recent push of #5983 has caused an issue with the vscodium package, the package is not built from source, but rather it's binary is simply taken from it's repo and slapped onto the system; I figure it would be proper to add a way to install binaries from other sources without it being detected as building from source, or to actually compile it....

Issue was reported to the discord guild.

uberhacker:
Fixes #5986.


drocsid:
still broken https://github.com/skycocker/chromebrew/issues/5986#issuecomment-888138117

PR #5992, opened 2021-07-27
Add crew constants from environment variables

uberhacker:
Something that was bothering me.

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.12.0'
+CREW_VERSION = '1.13.0'
 
 ARCH_ACTUAL = `uname -m`.strip
 # This helps with virtualized builds on aarch64 machines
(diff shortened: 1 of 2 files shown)


supechicken:
Will this break some package setting `CREW_NOT_STRIP` or `CREW_CONFLICTS_ONLY_ADVISORY`like `chrome`? (those env variable are fixed on start, not check every time)

lib/const.rb
```diff
@@ -48,25 +52,44 @@
...
+
+CREW_SHRINK_ARCHIVE = ENV['CREW_SHRINK_ARCHIVE']
+
+# Set testing constants from environment variables
+crew_testing_repo = ENV['CREW_TESTING_REPO']
+crew_testing_branch = ENV['CREW_TESTING_BRANCH']
+crew_testing = ENV['CREW_TESTING']
+crew_testing = '0' if crew_testing_repo.nil? || crew_testing_repo.empty?
+crew_testing = '0' if crew_testing_branch.nil? || crew_testing_branch.empty?
+CREW_TESTING = crew_testing
```
supechicken, line 80: Does we need both variables?

satmandu:
Would it be easier to just import all CREW_ env variables as CREW constants?

Or import from a whitelist? That way we don't need to update const.rb?

satmandu:
"Easier" 

i.e. If a CREW_* env variable exists, set it as a ruby constant, and override any other setting of the constant in const.rb, unless in a blacklist of constants. (could go whitelist or blacklist.)

lib/const.rb
satmandu: let's merge this and then we can back out a duplicate if necessary....

satmandu:
> Will this break some package setting `CREW_NOT_STRIP` or `CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY` like `chrome`? (those env variable are fixed on start now, not checked by every time)

The variable names are not changing, so I think it should be ok?

PR #6015, opened 2021-07-30
OpenSSH Fido2  Support (plus libfido2 and libcbor formulae/recipes)

chromakey-io:
Hi guys,

I've been trying to find a way to get security keys to work on my chromebook with openssh.

This seemed like the best way, though perhaps not the easiest.

I managed to get all of the  software working and looking for available keys, unfortunately something unique with chrome seems to be blocking access to the keys.

Here's the debug output:

>ssh-keygen -t ecdsa-sk -f ~/.ssh/id_mykey_sk -vv

>Generating public/private ecdsa-sk key pair.

>You may need to touch your authenticator to authorize key generation.

>debug1: start_helper: starting /usr/local/libexec/ssh-sk-helper 

>debug1: sshsk_enroll: provider "internal", device "(null)", application "ssh:", userid "(null)", flags 0x01, challenge len 0

>debug1: sshsk_enroll: using random challenge

**>debug1: sk_probe: 0 device(s) detected**

>debug1: ssh_sk_enroll: failed to find sk

>debug1: sshsk_enroll: provider "internal" failure -1

>debug1: ssh-sk-helper: Enrollment failed: invalid format

>debug1: main: reply len 8

>debug1: client_converse: helper returned error -4

>Key enrollment failed: invalid format`

So the "internal provider" works just fine.  Unfortunately, "sk_probe" finds no devices :(  I've tried adding some udev rules, but I've reached the end of the  line wwith  trying to debug this and need to move onto some other projects.

If anyone can get the rest of this working I'll chip in and document and  help clean up the code :)

My code/recipes are a little rough, quick, and dirty ... but i've added 2 other libraries in addition to the modications I  made to openssh.  LibFido2 and LibCBOR.  Use at your own peril :)


uberhacker:
Thank you for the update @chromakey-io!  @satmandu: Did you want to generate binaries since this new dependency will end up being a core package?

satmandu:
So here's a related issue for crouton: https://github.com/dnschneid/crouton/issues/1162

Note that they appear to be bind mounting `/run/udev` as per https://github.com/dnschneid/crouton/wiki/udev:-manage-inserted-devices which we are not doing.... so that might have to be done to get udev to detect devices...

chromakey-io:
Also FYI .... for anyone trying this at home you need to put:

`SecurityKeyProvider=internal`

In one of your ssh config  files (*not* sshd, ssh_config or what-have-you)

chromakey-io:
I got it working with the instructions here:

https://github.com/dnschneid/crouton/issues/1162

I tried everything they  thought might work, but I'll take the time to narrow it down to *exactly* what needs  to be done.

I think it is just the udev rule they recommend lower down that worked for the OP ... regardless nearly all of this  stuff could be added to postinstall so it works out of the box :)

Thansk for the help!

chromakey-io:
SO it's just the udev rule put in /run/udev/rules.d

`SUBSYSTEMS=="usb", ATTRS{idVendor}=="1ea8", ATTRS{idProduct}=="fc25", OWNER="chronos", GROUP="wheel", MODE="0660"`

Two problems:

1) it's outside of the crew /usr/local space

2) it requires a specific manufacturer and device ID (maybe this can be fixed, don't know too much about the rules system here)

chromakey-io:
Okay .... mistakes were made.


The only thing that needs to change is the permissions on the hidraw devices.

This doesn't require any knowledge of the specific manufacturer or device ID.

The file that manages this in chrome os:

/lib/udev/rules.d/99-hidraw.rules

I tried adding chronos to the  group "hidraw" with no success.  Though changing the permissions from 0660 to 0666 worked just fine.  That may not be optimal from a security  perspective ...

_This is the only system modification required:  change MODE=0666 in /lib/udev/rules.d/99-hidraw.rules_

... but I believe the "permission_broker" is doing something magical behind the scenes  blocking access.  I think it is setup to prevent hidraw devices from being past through to  the VM's .... but I've not really figured out how that mess works yet (if i did, i'd have setup passthrough for crostini :)

uberhacker:
Can you install a new file in `/usr/local/lib/udev/rules.d/99-hidraw.rules` and reference the new location in the config?

satmandu:
Realizing we may need to compile eudev with `--with-rootrundir=/run ` as per https://github.com/dnschneid/crouton/blob/14ca6bc7f0a1314f3972ce0cbf915525ece609a1/host-bin/enter-chroot#L367

chromakey-io:
@uberacker yes there *has* to be a way to include other udev directories .... i'm just not familiar with it.  Right putting  the config file in either the /run/udev or the /lib/udev locations works .... but obviously it would be ideal if we could configure things to get someting in /usr/local working or we'd probably be left with requiring the user to do that final step with some instructions during the install ...  as many users likely don't have RW privs on their rootfs and boot drives  .... and this really complicates things obviously.

chromakey-io:
1) eudev and it's associated binaries all need to be run as root.  It's a replacement for udev when your system is not running systemd.  It's not something that can run in userspace, but perhaps something that could run in a chroot environment.

I'd argue it be removed from chromebrew (along with the udev WSL stub), unless someone can explain a use-case for either.

2) I've obviously now tried adding udev rules to the /usr/local/etc/udev/rules.d space .... and even running eudev as root.  It does not work, as there is already a udev process running with a lower PID capturing all of these events well before eudev ever could.

3) the only way to make fido keys and this associated stuff work is by modifying system files within the existing udev system.  There is nothing that can be done in user-space ... unless chromebrew decides to switch towards a chroot model like crouton has.

uberhacker:
All great observations @chromakey-io.  Maybe @janbaudisch or @cstrouse can explain it in more detail?  I believe @janbaudisch created the package and @cstrouse updated it in the past.

chromakey-io:
*Maybe* it could work in a chroot environment or a well-crafted container system.

Though if we want this to work *just* in user space there's no escaping the modification of a root udev file (but only *very* slightly, changing permissions from 0660 to 0666 in *one* file :)

packages/openssh.rb
```diff
@@ -22,11 +22,16 @@ class Openssh < Package
      x86_64: '8c04f059c2d1f7683185d795fd87adad1988cd46546ab0c445957c2ed593781c'
   })
```
uberhacker, line 23: Remove binaries?

uberhacker: Bump version?

packages/openssh.rb
```diff
@@ -22,11 +22,16 @@ class Openssh < Package
...
+  depends_on 'zlibpkg'
+  depends_on 'openssl'
+  depends_on 'libfido2'
+
   def self.build
     system 'autoreconf -fiv'
     system 'autoheader'
     system "env #{CREW_ENV_OPTIONS} \
-      ./configure #{CREW_OPTIONS}"
+      ./configure #{CREW_OPTIONS} --with-ssl-engine --with-ssl-dir=/usr/local/include/openssl --with-zlib=/usr/local/include/zlib.h --with-security-key-builtin"
```
uberhacker, line 34: The include directories should be abstracted as shown below:
```
./configure #{CREW_OPTIONS} --with-ssl-engine \
--with-ssl-dir=#{CREW_PREFIX}/include/openssl \
--with-zlib=#{CREW_PREFIX}/include/zlib.h \
--with-security-key-builtin"
```

PR #6020, opened 2021-07-31
Add a CI bot for updating readme packages

supechicken:
Let’s hire a bot for doing that automatically:)

Tested on my fork

### DIFF
--- a/.github/workflows/readme-ci.yml
+++ b/.github/workflows/readme-ci.yml
@@ -1,0 +1,29 @@
+name: Readme Package Updater
+
+on:
+  push:
+    branch: 'master'
+    paths:
+      - 'packages/**.rb'
+
+jobs:
+  update_readme:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


uberhacker:
This is interesting but I'm not following what it actually does during the update.

supechicken:
At every commit pushes, it will check modified/added files and add them to `Recently updated/Latest Package` table in readme

uberhacker:
How do you tell the difference between recently updated and latest packages?

supechicken:
> How do you tell the difference between recently updated and latest packages?

By running `git diff-tree`, we can get the path of modified/added file and status  (added/modified) to know whatever they new package or updated.

Also in `README.md`, I added some comments for helping the script find the part of new/recently update packages

saltedcoffii:
I'd also love to see the bot check for package updates, automatically update them and make a PR. That wud save us so much time :P

PR #6023, opened 2021-08-02
Split crew warnings and error messages

uberhacker:


bin/crew
```diff
@@ -918,11 +925,20 @@ def prepare_package(destdir)
     conflicts << %x[#{@conflictscmd}].chomp.split(" ")
     conflicts.reject!(&:empty?)
     unless conflicts.empty?
-      puts "Unable to complete this build since there is a conflict with the same file in another package.".lightred
+      if CREW_CONFLICTS_ONLY_ADVISORY == '1'
```
supechicken, line 928: Can we make warning as default? Since it doesn't break the package

uberhacker: I would argue making the default a warning instead would potentially cause more breakage since conflicts can create a problem if packages are removed.  It would encourage laziness.

supechicken: Sorry for pointing wrong, I mean the FHS3 part

uberhacker: Thoughts @saltedcoffii or @satmandu?

saltedcoffii: We should definitely default to error with FHS3 problems. Chromebrew packages need to be standardized.

satmandu: @saltedcoffii Do you have the (literal) bandwitdth to script scanning the package binaries to see which packages have FHS3 problems? It would be good to know how many packages need to be dealt with before we make this an error vs a warning.

uberhacker: Just so you guys know this PR is only to provide intuitive feedback in the process.  We can set defaults in lib/const.rb.

PR #6025, opened 2021-08-02
eudev -> update & change rootrundir

satmandu:
- `rootrundir` should be `/run`

Works properly:
- [x] x86_64
- [x] armv7l (built with M91 container)
- [x] i686 (@uberhacker a build please?)

### DIFF
--- a/packages/eudev.rb
+++ b/packages/eudev.rb
@@ -3,41 +3,39 @@ require 'package'
 class Eudev < Package
   description 'Gentoo standalone udev'
   homepage 'https://wiki.gentoo.org/wiki/Project:Eudev'
-  version '3.2.9'
+  version '3.2.10'
   license 'LGPL-2.1, MIT and GPL-2'
   compatibility 'all'
   source_url 'https://github.com/gentoo/eudev/archive/v3.2.9.tar.gz'
   source_sha256 '7d281276b480da3935d1acb239748c2c9db01a8043aad7e918ce57a223d8cd24'
 
-  binary_url ({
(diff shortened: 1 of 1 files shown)


packages/eudev.rb
```diff
@@ -28,16 +26,15 @@ class Eudev < Package
 
   def self.build
     system 'sed -i s,/usr/bin/xsltproc,xsltproc,g man/make.sh'
```
uberhacker, line 28: Move to `def self.patch` section?

packages/eudev.rb
```diff
@@ -28,16 +26,15 @@ class Eudev < Package
...
-           "--with-rootlibexecdir=#{CREW_PREFIX}/libexec"
+    system 'NOCONFIGURE=1 ./autogen.sh'
+    system "#{CREW_ENV_OPTIONS} ./configure \
+              #{CREW_OPTIONS} \
+              --enable-hwdb \
+              --enable-rule-generator \
+              --with-rootprefix=#{CREW_PREFIX} \
+              --with-rootlibdir=#{CREW_LIB_PREFIX} \
+              --with-rootrundir=/run \
+              --with-rootlibexecdir=#{CREW_PREFIX}/libexec"
```
uberhacker, line 37: Don't all these options default to the correct locations already?

satmandu: You are right.

packages/eudev.rb
satmandu: done

uberhacker:
Tested on armv7l and i686.  The udevadm executable is not segfaulting.

PR #6026, opened 2021-08-03
Add package for opening URL on Chrome OS

supechicken:
Create a new package for calling Chrome OS to open an URL through dbus 

Tested on `x86_64`

### DIFF
--- a/packages/crew_url_handler.rb
+++ b/packages/crew_url_handler.rb
@@ -1,7 +1,7 @@
 require 'package'
 
 class Crew_url_handler < Package
-  description 'Script for opening an URL in Chrome OS browser'
+  description 'Script for opening an URL in your default browser'
   homepage 'https://github.com/skycocker/chromebrew'
   version '1.0'
   license 'GPL-3'
--- a/packages/crew_url_opener.rb
+++ b/packages/crew_url_opener.rb
@@ -1,0 +1,56 @@
+require 'package'
+
+class Crew_url_opener < Package
+  description 'Script for opening an URL in Chrome OS browser'
+  homepage 'https://github.com/skycocker/chromebrew'
+  version '1.0'
+  license 'GPL-3'
+  compatibility 'all'
+  source_url 'SKIP'
+  source_sha256 'SKIP'
+  
+  def self.install
+    @script = <<SCRIPT
+#!/bin/bash
+
+case ${1} in
+https://|http://)
+  exec dbus-send \
+    --system \
+    --type=method_call \
+    --print-reply \
+    --dest=org.chromium.UrlHandlerService \
+    /org/chromium/UrlHandlerService \
+    org.chromium.UrlHandlerServiceInterface.OpenUrl \
+    "string:${1}"
+;;
+-h|--help)
+  cat <<EOF
+open-url: Open an URL in Chrome OS browser.
+
+Usage:
+  open-url <option> [URL]
+  (URL must start with 'http://' or 'https://')
+
+  -h --help      Show this message
+EOF
+
+;;
+*)
+  cat <<EOF
+Error: only support URL start with 'http://' or 'https://'
+
+Type 'open-url --help' for usage
+EOF
+
+  exit 1
+;;
+esac
+SCRIPT
+
+  end
+
+  def self.postinstall
+    puts "Type 'url-opener' for usage"
+  end
+end
--- a/packages/crew_url_opener.rb
+++ b/packages/crew_url_opener.rb
@@ -1,6 +1,6 @@
 require 'package'
 
-class Crew_url_opener < Package
+class Crew_url_handler < Package
   description 'Script for opening an URL in Chrome OS browser'
   homepage 'https://github.com/skycocker/chromebrew'
   version '1.0'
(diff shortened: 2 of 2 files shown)


packages/crew_url_opener.rb
```diff
@@ -0,0 +1,56 @@
+require 'package'
+
+class Crew_url_opener < Package
+  description 'Script for opening an URL in Chrome OS browser'
+  homepage 'https://github.com/skycocker/chromebrew'
+  version '1.0'
+  license 'GPL-3'
+  compatibility 'all'
+  source_url 'SKIP'
+  source_sha256 'SKIP'
```
uberhacker, line 10: I believe you can remove this line and it should still skip the source.

packages/crew_url_opener.rb
```diff
@@ -0,0 +1,56 @@
...
+  cat <<EOF
+Error: only support URL start with 'http://' or 'https://'
+
+Type 'open-url --help' for usage
+EOF
+
+  exit 1
+;;
+esac
+SCRIPT
```
uberhacker, line 49: Where is this script wrote out to a file?  The `@script = <<SCRIPT...SCRIPT` should probably go in `def self.build` instead of `def self.install`.

uberhacker: Is there a way to set the default browser or can it only open in Chrome?

packages/crew_url_opener.rb
```diff
@@ -0,0 +1,56 @@
...
+  compatibility 'all'
+  source_url 'SKIP'
+  source_sha256 'SKIP'
+  
+  def self.install
+    @script = <<SCRIPT
+#!/bin/bash
+
+case ${1} in
+https://|http://)
```
uberhacker, line 17: Does this work for partial strings?

saltedcoffii:
We should get this to work with `xdg-open` somehow

uberhacker:
Interesting script idea @supechicken.  I'm thinking there should be some dependencies?

packages/crew_url_opener.rb
supechicken: Done

packages/crew_url_opener.rb
supechicken: I think I accidentally deleted that line😅 Added default browser part in postinstall

packages/crew_url_opener.rb
supechicken: Now it will work with urls without http prefix

supechicken:
> We should get this to work with `xdg-open` somehow

We may need a desktop file for that

supechicken:
> Interesting script idea @supechicken. I'm thinking there should be some dependencies?

No, the `dbus-send` command is already built-in to system :)

supechicken:
@saltedcoffii: Confirmed it works with `xdg-open`
@uberhacker: This pr is ready

packages/crew_url_opener.rb
```diff
@@ -0,0 +1,104 @@
+require 'package'
+
+class Crew_url_handler < Package
```
uberhacker, line 3: `Crew_url_handler` class doesn't match the filename `crew_url_opener.rb`?

packages/crew_url_opener.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+  end
+
+  def self.build
+    @_script = <<~SCRIPT
+      #!/bin/bash
+      case "${1}" in
+      ''|-h|--help)
+        cat <<EOF
+      ${0##*/}: Open an URL in Chrome OS browser.
```
uberhacker, line 35: Should this say `Open an URL in your default browser`?

packages/crew_url_opener.rb
```diff
@@ -0,0 +1,104 @@
...
+      ;;
+      esac
+    SCRIPT
+
+    # reference:
+    # https://chromium.googlesource.com/chromiumos/containers/cros-container-guest-tools/+log/refs/heads/main/cros-garcon/garcon_host_browser.desktop
+
+    @_desktop = <<~EOF
+      [Desktop Entry]
+      Name=Chrome OS Host Browser
```
uberhacker, line 63: `Name=Default Host Browser`?

packages/crew_url_opener.rb
```diff
@@ -0,0 +1,104 @@
...
+    FileUtils.mkdir_p ["#{CREW_DEST_PREFIX}/bin/", "#{CREW_DEST_PREFIX}/share/application/"]
+    File.write "#{CREW_DEST_PREFIX}/share/application/cros_host_browser.desktop", @_desktop
+    File.open "#{CREW_DEST_PREFIX}/bin/crew-url-handler", 'w' do |file|
+      file.write(@_script)
+      file.chmod(0755)
+    end
+  end
+
+  def self.postinstall
+    print 'Set Chrome OS browser as default? [Y/n]: '
```
uberhacker, line 82: Should this be flexible for any browser?  Would it make sense to loop over all the installed browser packages and provide a list of options with the current default selected?

packages/crew_url_opener.rb
```diff
@@ -0,0 +1,104 @@
+require 'package'
+
+class Crew_url_handler < Package
+  description 'Script for opening an URL in Chrome OS browser'
```
uberhacker, line 4: `Script for opening an URL in your default browser`?

packages/crew_url_opener.rb
supechicken: But users may select a different browser as the default

packages/crew_url_opener.rb
supechicken: Corrected to `crew_url_opener`

packages/crew_url_opener.rb
supechicken: Done

packages/crew_url_opener.rb
supechicken: Good idea, do we need to create a list for all browser packages in our repo?

uberhacker: AFAIK, the browsers in the repo include: brave, chrome, elinks, epiphany, firefox, links, netsurf, opera, torbrowser, vivaldi and w3m

packages/crew_url_handler.rb
```diff
@@ -0,0 +1,104 @@
...
+      MimeType=x-scheme-handler/http;x-scheme-handler/https;x-scheme-handler/ftp;x-scheme-handler/mailto;
+      Type=Application
+      NoDisplay=true
+      OnlyShowIn=Never
+    EOF
+  end
+
+  def self.install
+    FileUtils.mkdir_p ["#{CREW_DEST_PREFIX}/bin/", "#{CREW_DEST_PREFIX}/share/application/"]
+    File.write "#{CREW_DEST_PREFIX}/share/application/cros_host_browser.desktop", @_desktop
```
uberhacker, line 74: Do desktop icons work?  That would be really awesome if they did.

packages/crew_url_handler.rb
```diff
@@ -0,0 +1,104 @@
...
+    File.open "#{CREW_DEST_PREFIX}/bin/crew-url-handler", 'w' do |file|
+      file.write(@_script)
+      file.chmod(0755)
+    end
```
uberhacker, line 78: Interesting but seemingly more complicated way to save the file.

packages/crew_url_handler.rb
supechicken: Do you mean an icon at the dock?

packages/crew_url_handler.rb
supechicken: It might be better as it only opens the file once

packages/crew_url_handler.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  def self.postinstall
+    print 'Set Chrome OS browser as default? [Y/n]: '
+
+    case STDIN.getc
+    when "\n", 'Y', 'y'
+      @_www_browser_symlinks.each do |symlink|
+        FileUtils.ln_sf 'crew-url-handler', symlink
+      end
+      puts 'Chrome OS browser is now default.'.lightgreen
```
uberhacker, line 89: Should this be limited to Chrome?  Are other options possible?

Issue #6027, opened 2021-08-05
[nano] Missing libmagic.so.1 shared library


leoplo:
Similar to https://github.com/skycocker/chromebrew/issues/5827.

## Description
Installing nano worked fine but running it failed with message " Missing libmagic.so.1 shared library".

If you executed a specific command:
```shell
crew install nano
touch tmp
nano tmp
```

Output:
```shell
Missing libmagic.so.1 shared library
```

## Versions and system information
chromebrew: `crew --version`
1.11.9

ChromeOS: `cat /etc/lsb-release`
CHROMEOS_RELEASE_BUILD_NUMBER=14092

Model: HP Chromebook 13 G1

Architecture: `uname -m`
x86_64

PR #6028, opened 2021-08-05
packages/nano.rb : fix Missing libmagic.so.1

leoplo:
Add filecmd dependency to fix "Missing libmagic.so.1 shared library".

Fixes #

#6027 

## Description
Add filecmd dependency to fix Missing libmagic.so.1

Works properly:
- [x] x86_64

### DIFF
--- a/packages/nano.rb
+++ b/packages/nano.rb
@@ -24,6 +24,7 @@ class Nano < Package
   })
 
   depends_on 'xdg_base'
+  depends_on 'filecmd'
 
   def self.patch
     system "sed -i '/SIGWINCH/d' src/nano.c"


satmandu:
Packages listed in https://github.com/skycocker/chromebrew/blob/master/packages/core.rb should always be installed, as such we don't include those in other package files.

`readline` is one such package.

Having said that, the package which has `libmagic.so.1` is actually `filecmd`, as shown by the command 
```
chronos@cheekon-x86_64 /usr/local/lib/crew/packages $ crew whatprovides libmagic.so.1
filecmd: /usr/local/lib64/libmagic.so.1
filecmd: /usr/local/lib64/libmagic.so.1.0.0

Total found: 2
```

So if you just want to edit this PR to make sure that `filecmd` is one of the dependencies for nano, that would be great, since `filecmd` is not part of `core`.

Alternately @uberhacker @saltedcoffii @supechicken what do you think about filecmd being a part of core?

supechicken:
I think it is okay to put it in `core`, since `file` is a built-in command in lots of Linux distros

satmandu:
Superseded by https://github.com/skycocker/chromebrew/pull/6032

PR #6031, opened 2021-08-05
meson -> 0.59.0

satmandu:
- @uberhacker i686 build please?

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686


uberhacker:
Wow, finally an update that actually works on i686.

PR #6039, opened 2021-08-08
Update blender from 2.92.0 to 2.93.2

uberhacker:


packages/blender.rb
```diff
@@ -19,7 +19,7 @@ class Blender < Package
   def self.install
     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/blender"
-    FileUtils.mv Dir.glob('*'), "#{CREW_DEST_PREFIX}/share/blender"
+    FileUtils.mv Dir['*'], "#{CREW_DEST_PREFIX}/share/blender"
     FileUtils.ln_s "#{CREW_PREFIX}/share/blender/blender", "#{CREW_DEST_PREFIX}/bin/blender"
```
supechicken, line 23: Use relative symlink?
```
FileUtils.ln_s '../share/blender/blender', "#{CREW_DEST_PREFIX}/bin/blender"
```

uberhacker: The absolute symlink works the same and is easier to understand.

PR #6046, opened 2021-08-08
Update gcloud from 346.0.0 to 351.0.0

uberhacker:


packages/gcloud.rb
```diff
@@ -3,16 +3,16 @@
...
+    source_url 'https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-351.0.0-linux-x86.tar.gz'
+    source_sha256 '8562bda2cc815d0ff6c0bddfc3f5e3546bf9519b3a372af8827848c0f8b7e193'
   when 'x86_64'
-    source_url 'https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-346.0.0-linux-x86_64.tar.gz'
-    source_sha256 '9d21569ecd66e4278c45594e3ce10fcb2591704756aeb2b0f04bed14f86f9658'
+    source_url 'https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-351.0.0-linux-x86_64.tar.gz'
+    source_sha256 'f91a23607af59ac76f23276a184b400c227ba1c0681b6800546adbe260c10a42'
   end
 
   depends_on 'xdg_base'
```
supechicken, line 18: Write to `env.d` instead of `.bashrc`?

uberhacker: I believe the installer script modifies `~/.bashrc` already but I can try to update to use the `env.d` location instead.

uberhacker: @supechicken: I _think_ I was able to relocate the init in `env.d` instead.

PR #6050, opened 2021-08-09
Add jdk16 package

uberhacker:
Include preflight fake packages in dependency resolver.  Tested on x86_64.


satmandu:
Should GCC be part of this as well?

PR #6051, opened 2021-08-09
Curl => 7.78.0

saltedcoffii:
- [x] x86_64
- [ ] i686
- [ ] armv7l/aarch64

Also upgraded supporting libraries.

### DIFF
--- a/packages/curl.rb
+++ b/packages/curl.rb
@@ -25,7 +25,7 @@ class Curl < Package
       FileUtils.mkdir_p "build/#{builddir}"
     end
 
-    puts "Downloading supporting libraries...".yellow
+    puts "Downloading supporting libraries.".yellow
     Dir.chdir 'build/zlib' do
       system 'curl -#L https://zlib.net/zlib-1.2.11.tar.gz | \
         hashpipe sha256 c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1 | \
@@ -56,7 +56,7 @@ class Curl < Package
        hashpipe sha256 acffef2da98e761fc1fd9c4fddde0f3af60ab44c4f5af05cd1b2d60a3fa08718 | \
         tar xJ --strip-components=1'
     end
-    Dir.chdir 'build/idn2' do
+    Dir.chdir 'build/nghttp2' do
       system 'curl -#L https://github.com/nghttp2/nghttp2/releases/download/v1.44.0/nghttp2-1.44.0.tar.bz2 | \
         hashpipe sha256 989971276517a1c9ed330b779c34cf02d99da3b85d156eb297f42b1b7227b297 | \
         tar xj --strip-components=1'
@@ -81,8 +81,8 @@ class Curl < Package
     end
 
     @curldep_cmake_options = "CC=musl-gcc \
-          CFLAGS='-flto -pipe -Os -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
-          CXXFLAGS='-flto -pipe -Os -fPIC -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
+          CFLAGS='-flto -pipe -Os -fPIC -fuse-ld=gold -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
+          CXXFLAGS='-flto -pipe -Os -fPIC -fuse-ld=gold -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans #{@arch_ssp_cflags}' \
           CPPFLAGS='-I#{@deppath}/include' \
           LDFLAGS='-L#{@deppath}/#{ARCH_LIB}' \
           cmake \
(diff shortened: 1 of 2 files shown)


satmandu:
Nice. Setting up builds for x86_64 & armv7l.

satmandu:
I think we might want to use the source links and not git links, because otherwise we see this:
```
chronos@cheekon-x86_64 /usr/local/lib/crew/packages $ curl --version
curl 7.78.0-DEV (x86_64-pc-linux-musl) libcurl/7.78.0-DEV OpenSSL/1.1.1k zlib/1.2.11 brotli/1.0.9 zstd/1.5.0 libidn2/2.3.2 nghttp2/1.44.0
Release-Date: [unreleased]
Protocols: dict file ftp ftps gopher gophers http https mqtt pop3 pop3s smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IDN IPv6 Largefile libz NTLM NTLM_WB SSL TLS-SRP UnixSockets zstd
```

satmandu:
Also metalink linking appears to be deprecated:
`configure: error: --with-libmetalink no longer works!`

packages/curl.rb
```diff
@@ -3,40 +3,69 @@
...
+    Dir.chdir 'build/libssh' do
+      system 'curl -#L https://www.libssh.org/files/0.9/libssh-0.9.5.tar.xz | \
+       hashpipe sha256 acffef2da98e761fc1fd9c4fddde0f3af60ab44c4f5af05cd1b2d60a3fa08718 | \
+        tar xJ --strip-components=1'
+    end
+    Dir.chdir 'build/nghttp2' do
+      system 'curl -#L https://github.com/nghttp2/nghttp2/releases/download/v1.44.0/nghttp2-1.44.0.tar.bz2 | \
+        hashpipe sha256 989971276517a1c9ed330b779c34cf02d99da3b85d156eb297f42b1b7227b297 | \
+        tar xj --strip-components=1'
+    end
```
uberhacker, line 63: Why not just use crew packages instead?

satmandu: Because these are for building against musl...

satmandu:
Added some binaries which need testing as I have no native hardware at the moment.

@uberhacker i686 binaries please?

satmandu:
Whoops! Didn't include an armv7l build for libcurl. Starting that now.

satmandu:
ok, all armv7l binaries are now uploaded (but unchecked) for this.

uberhacker:
Overnight i686 build locked up at:
```
checking if getifaddrs can be linked... yes
checking if getifaddrs is prototyped... yes
checking if getifaddrs is compilable... yes
checking if getifaddrs seems to work... 
```

uberhacker:
On armv7l, I get:
```
$ crew search curl -v
curl: Command line tool and library for transferring data with URLs.
https://curl.se/
Version: 7.78.0
License: curl
$ curl -L# git.io/vddgY -o install.sh
bash: /usr/local/bin/curl: No such file or directory
$ ls -l /usr/local/bin/curl
-rwxr-xr-x 1 chronos chronos 3514344 Aug 10 01:53 /usr/local/bin/curl
```

satmandu:
> On armv7l, I get:
> 
> ```
> $ crew search curl -v
> curl: Command line tool and library for transferring data with URLs.
> https://curl.se/
> Version: 7.78.0
> License: curl
> $ curl -L# git.io/vddgY -o install.sh
> bash: /usr/local/bin/curl: No such file or directory
> $ ls -l /usr/local/bin/curl
> -rwxr-xr-x 1 chronos chronos 3514344 Aug 10 01:53 /usr/local/bin/curl
> ```

Thanks. debugging...

uberhacker:
Also, on x86_64, I get:
```
$ curl -V
$ curl -L# git.io/vddgY -o install.sh
$ ls install.sh
ls: cannot access 'install.sh': No such file or directory
```
Doesn't say file not found but doesn't execute either, apparently.

satmandu:
Looks like there's a musl dependency, which defeats the point of using musl, as we were using musl to get a static build.

(install musl to fix this, but this package build needs to be fixed @saltedcoffii )

satmandu:
My initial thought would be that gold linker might break this...

packages/curl.rb
saltedcoffii: Honestly I've been thinking of switching entirely to musl with Chromebrew. Like wouldn't that fix so many glibc problems and allow chromebrew, and allow binaries to be much smaller and run faster? @satmandu let me know you're opinion, but I'd really like to start doing that.

satmandu:
I'm redoing this based on the git package file, since that did produce a working static musl curl.

satmandu:
FYI if testing this breaks curl, just do this:
`cp /usr/local/bin/curl_musl_static /usr/local/bin/curl`

satmandu:
ok this is due to a bug in the 7.78.0 sources....

satmandu:
Got it working on x86_64:
```
chronos@cheekon-x86_64 ~ $ curl --version
curl 7.78.0 (x86_64-pc-linux-gnu) libcurl/7.78.0 OpenSSL/1.1.1k zlib/1.2.11 brotli/1.0.9 zstd/1.5.0 c-ares/1.17.2 libidn2/2.3.2 nghttp2/1.44.0
Release-Date: 2021-07-21
Protocols: dict file ftp ftps gopher gophers http https mqtt pop3 pop3s smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli HSTS HTTP2 HTTPS-proxy IDN IPv6 Largefile libz NTLM NTLM_WB SSL TLS-SRP UnixSockets zstd
chronos@cheekon-x86_64 ~ $ ldd `which curl`
        not a dynamic executable
```

satmandu:
Static builds should now be fixed. I imported from the git build, and then realized that this was a bug in 7.78.0, but sunk costs and all it now all appears to work in the containers.

Please see if this works on real hardware, please?

packages/curl.rb
uberhacker: This PR is way too complex.  It also won't build on i686.  Would it make sense to split it out into dependency packages?

satmandu: Sure... That makes sense.

Let me create some musl_ packages, which will put stuff in the non-FHS compliant (but standard for musl!) #{CREW_PREFIX}/musl 

What do you think @saltedcoffii ?

saltedcoffii:
Works on real x86_64 hardware as far as I can tell

packages/curl.rb
saltedcoffii: I think we have much bigger priorities. Many packages don't compile with musl without void or alpine patches, so easy patch downloading is a must, hence the extra url PR. I also think we need (and have needed for a while) a way to track and remove orphaned dependencies, much like `apt autoremove`. Is that reasonable @satmandu ?

uberhacker:
It works fine on x86_64 for me too.

uberhacker:
This also works on armv7l.  At least on my outdated release 84 EOL'd system.

uberhacker:
@satmandu: Does this PR have any hope or should we start over?

satmandu:
> @satmandu: Does this PR have any hope or should we start over?

I should have the curl 1.79.1 up shortly...

satmandu:
7.79.1 is up here: https://github.com/skycocker/chromebrew/pull/6065

uberhacker:
Superceded by #6065.

PR #6052, opened 2021-08-09
Always filefix

saltedcoffii:
Always run filefix on all source packages.

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -827,6 +827,10 @@ def build_and_preconfigure(target_dir)
     @pkg.in_build = true
     @pkg.patch
     @pkg.prebuild
+
+    puts 'Applying filefix'.lightblue
+    system 'filefix'
+
     @pkg.build
     @pkg.in_build = false
     # wipe crew destdir


satmandu:
I agree with the _intent_ of this very much.

But filefix doesn't work if a `configure` script doesn't exist yet, and also isn't needed if a `configure` `make` sequence isn't part of the build.

I would suggest adding `filefix` before `make` is run, but also only running it if a `configure` script exists.

saltedcoffii:
Okay, didn't think about configure not being present. How would you recommend doing that?

satmandu:
> Okay, didn't think about configure not being present. How would you recommend doing that?

At the minimum `if File.exist?('configure')` should wrap that code, but I think you also want that attached to the code which detects `make` since you don't need this if for instance meson or cmake are being used. I think that's in package.rb, where the code also adds `-j NPROC` to make.

satmandu:
Whoops, obviously filefix should go before a `configure` line, not a `make` line... so this should really check to see if the file `configure` exists and that filefix is prepended before a line with `configure` in it.

uberhacker:
Actually, it doesn't hurt to run it if the configure scripts are missing.  The find command will only operate on existing files.

satmandu:
Would it be better (though possibly slower) to just have the entire thing run as a system line instead of calling filefix and requiring the external package?

e.g.
```
system "for f in \$(find . -name configure); do sed -i 's,/usr/bin/file,#{CREW_PREIX}/bin/file,g' ${f} ; done"
```

uberhacker:
We could do that but it would be nice to keep the filefix command so builds can be done manually, if necessary.

uberhacker:
We would still need the filecmd package, otherwise, the file command doesn't exist. :)

uberhacker:
@saltedcoffii: We can add this but we'll need to bump the crew version in `lib/const.rb`.

saltedcoffii:
I will, but I wanna make it work when the configure file isn't initially present, i.e. it must be generated with autotools.

uberhacker:
It doesn't hurt to run it on all source even if no configure scripts are found.  The find command will ignore executing the sed command which essentially does the same thing.

PR #6056, opened 2021-08-10
gtest 1.10.0 => 1.11.0

saltedcoffii:
Tested:
- [x] x86_64
- [x] i686
- [x] armv7l/aarch64

### DIFF
--- a/packages/gtest.rb
+++ b/packages/gtest.rb
@@ -3,52 +3,36 @@ require 'package'
 class Gtest < Package
   description 'Google Test - C++ testing utility'
   homepage 'https://opensource.google/projects/googletest'
-  version '1.10.0-389c'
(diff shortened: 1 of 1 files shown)


satmandu (approved):
Just need i686 binaries from @uberhacker and someone should test this on the other architectures as well...

uberhacker:
After running `crew files gtest`, I only see static libraries.  Should we be including shared libraries also?
```
/usr/local/lib64/libgmock.a
/usr/local/lib64/libgmock_main.a
/usr/local/lib64/libgtest.a
/usr/local/lib64/libgtest_main.a
```

uberhacker:
I updated to include shared libaries, rebuilt and tested all the binaries.

PR #6062, opened 2021-08-11
Musl native toolchain + all packages necessary for static curl, git builds

satmandu:
- Part of the curl package breakup, to allow building static.
- Please see if `/usr/local/musl/bin/gcc -dumpversion` works.
- Also see if `crew build musl_zlib` works.
- Only tested **in container**.

Works properly:
- [x] x86_64 
- [ ] armv7l
- [ ] i686


satmandu:
@ThatGeekyWeeb does this work for you in i686?

packages/musl_native_toolchain.rb
```diff
@@ -0,0 +1,51 @@
...
+    ENV['CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY'] = '1'
+    ENV['CREW_NOT_STRIP'] = '1'
+    ENV['CREW_SHRINK_ARCHIVE'] = '0'
+    warn_level = $VERBOSE
+    $VERBOSE = nil
+    load "#{CREW_LIB_PATH}lib/const.rb"
+    $VERBOSE = warn_level
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/musl/lib"
+    Dir.chdir("#{CREW_DEST_PREFIX}/musl") do
+      FileUtils.ln_s 'lib', 'lib64'
```
uberhacker, line 45: This only needs to be done on x86_64, correct?

satmandu: Yup. Added a condition.

ThatGeekyWeeb:
iirc musl does not build for i686

ThatGeekyWeeb:
hmm but you seem to have a package, but I have no way to test it, don't have an i686 chromebook; mine is armv7

ThatGeekyWeeb:
should be noted that it is impossible to build a static curl with ssh support; or rather I tested and attempted to for months with 0 success

ThatGeekyWeeb:
(this also means libcurl will lack ssh support and pkgs such as git that use libcurl will be unable to use ssh)

satmandu:
> should be noted that it is impossible to build a static curl with ssh support; or rather I tested and attempted to for months with 0 success 

I want to take a gander at this. 😂

ThatGeekyWeeb:
> > should be noted that it is impossible to build a static curl with ssh support; or rather I tested and attempted to for months with 0 success
> 
> I want to take a gander at this. joy

I'm p sure it's due to the fact that libssh2 and libssh both just can't be used statically; I tired with both when I did my testing

packages/musl_native_toolchain.rb
satmandu: Removed that entirely...

satmandu:
@ThatGeekyWeeb 
```
chronos@cheekon-x86_64 /usr/local/musl/lib $ curl --version
curl 7.78.0 (x86_64-pc-linux-gnu) libcurl/7.78.0 OpenSSL/1.1.1k zlib/1.2.11 brotli/1.0.9 zstd/1.5.0 c-ares/1.17.2 libidn2/2.3.2 libssh/0.9.5/openssl/zlib nghttp2/1.44.0
Release-Date: 2021-07-21
Protocols: dict file ftp ftps gopher gophers http https mqtt pop3 pop3s scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB SPNEGO SSL TLS-SRP UnixSockets zstd
chronos@cheekon-x86_64 /usr/local/musl/lib $ ldd `which curl`
        not a dynamic executable
```

uberhacker:
On i686, I get:
```
$ crew build musl_zlib
musl_zlib: zlib is a massively spiffy yet delicately unobtrusive compression library.
Downloading source...
Musl_zlib archive downloaded.
Compiler error reporting is too harsh for ./configure (perhaps remove -Werror).
** ./configure aborting.
musl_zlib failed to build: `PATH=/usr/local/musl/bin:/home/chronos/user/.cargo/bin:/usr/local/pkg/bin:/usr/local/pkg/sbin:/usr/local/bin:/usr/bin:/bin:/opt/bin:/usr/local/libexec/busybox/bin:/usr/local/libexec/busybox/sbin       CC='/usr/local/musl/bin/i686-linux-musl-gcc'       CXX='/usr/local/musl/bin/i686-linux-musl-g++'       LD=/usr/local/musl/bin/i686-linux-musl-ld.gold       PKG_CONFIG_LIBDIR=/usr/local/musl/lib/pkgconfig       CFLAGS='-B/usr/local/musl/include -flto -pipe -O3 -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans  -fno-stack-protector -fcommon'       CXXFLAGS='-B/usr/local/musl/include -flto -pipe -O3 -ffat-lto-objects -fipa-pta -fno-semantic-interposition -fdevirtualize-at-ltrans  -fno-stack-protector -fcommon'       CPPFLAGS='-I/usr/local/musl/include -fcommon'       LDFLAGS='-L/usr/local/musl/lib -flto -static' ./configure --prefix=/usr/local/musl           --static` exited with 1
```

satmandu:
@uberhacker  Can you post the part of `configure.log` from that build so we can see what was setting that off?

ThatGeekyWeeb:
> @ThatGeekyWeeb
> 
> ```
> chronos@cheekon-x86_64 /usr/local/musl/lib $ curl --version
> curl 7.78.0 (x86_64-pc-linux-gnu) libcurl/7.78.0 OpenSSL/1.1.1k zlib/1.2.11 brotli/1.0.9 zstd/1.5.0 c-ares/1.17.2 libidn2/2.3.2 libssh/0.9.5/openssl/zlib nghttp2/1.44.0
> Release-Date: 2021-07-21
> Protocols: dict file ftp ftps gopher gophers http https mqtt pop3 pop3s scp sftp smb smbs smtp smtps telnet tftp
> Features: alt-svc AsynchDNS brotli GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB SPNEGO SSL TLS-SRP UnixSockets zstd
> chronos@cheekon-x86_64 /usr/local/musl/lib $ ldd `which curl`
>         not a dynamic executable
> ```

I'd have to look into this more

---

But I think it might be smarter to, use some kinda of config variable to set which libc is being used and compile based on that config var instead of build custom musl packges; most packages should work instantly with musl and other that don't will require patches; should be noted that these patches will have no effect on glibc builds

none the less whats being done here seems fine unless you want to look into the above

ThatGeekyWeeb:
That is, unless we are simply adding the musl cross build tool chain and not adding musl in it's entirety; if we are just doing a cross build tool chain addition; maybe this is also a good time to look into the major advantages musl would have over glibc for chromebrew as a whole

satmandu:
> That is, unless we are simply adding the musl cross build tool chain and not adding musl in it's entirety; if we are just doing a cross build tool chain addition; maybe this is also a good time to look into the major advantages musl would have over glibc for chromebrew as a whole

This musl tool chain is really only useful for building static linked apps, I think.

Which is nice, since in theory this bypasses the orphaned Glibc on older EOL machines, but also because it is so much easier to get static builds of apps with musl than with Glibc which are also in theory backwards compatible with older Glibc installations.

Any thoughts on why we might be getting breakage on i686 with this though?

Do we know anyone besides @uberhacker who has i686 machines we can have test these i686 builds?

@uberhacker do you have more than one i686 machine? I just want to rule out hardware failure...

ThatGeekyWeeb:
I would consider musl with i686 a lost cause; void doesn't provide a i686 port of musl; I don't recall if alpine provides one, but if they do; might be smart to dig around in the APORTS repo to figure out how

ThatGeekyWeeb:
I checked alpine and aports; zlib should work fine with i686; there is likely an issue with the toolchain

satmandu:
@ThatGeekyWeeb I think I got the static git & static curl with ssh working. See https://github.com/skycocker/chromebrew/pull/6078 and https://github.com/skycocker/chromebrew/pull/6065

PR #6064, opened 2021-08-12
Rust -> 1.54.0

satmandu:
- Only tested in container and used in curl build. Needs testing on hardware.

Builds properly:
- [x] x86_64
- [x] i686
- [x] armv7l

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/rust.rb
+++ b/packages/rust.rb
@@ -70,16 +70,16 @@ class Rust < Package
   def self.remove
     config_dirs = %W[#{HOME}/.rustup #{CREW_PREFIX}/share/rustup #{HOME}/.cargo #{CREW_PREFIX}/share/cargo]
     config_dirs.each do |config_dir|
-      next unless Dir.exist? config_dir
-
-      system "echo #{config_dir}; ls #{config_dir}"
-      print "\nWould you like to remove the config directories above? [y/N] "
-      case $stdin.getc
-      when 'y', 'Y'
-        FileUtils.rm_rf config_dir
-        puts "#{config_dir} removed.".lightred
-      else
-        puts "#{config_dir} saved.".lightgreen
+      if Dir.exist?(config_dir)
+        system "echo #{config_dir}; ls #{config_dir}"
+        print "\nWould you like to remove the config directories above? [y/N] "
+        case $stdin.getc
+        when 'y', 'Y'
+          FileUtils.rm_rf config_dir
+          puts "#{config_dir} removed.".lightred
+        else
+          puts "#{config_dir} saved.".lightgreen
+        end
       end
     end
   end
--- a/packages/rust.rb
+++ b/packages/rust.rb
@@ -68,6 +68,18 @@ class Rust < Package
   end
   
   def self.remove
-    FileUtils.rm_f "~/.rustup"
+    config_dirs = %W[#{HOME}/.rustup #{CREW_PREFIX}/share/rustup #{HOME}/.cargo #{CREW_PREFIX}/share/cargo]
+    config_dirs.each do |config_dir|
+      next unless Dir.exist? config_dir
+
+      print "\nWould you like to remove #{config_dir}? [y/N] "
+      case $stdin.getc
+      when 'y', 'Y'
+        FileUtils.rm_rf config_dir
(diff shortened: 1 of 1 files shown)


packages/rust.rb
```diff
@@ -3,37 +3,37 @@
...
-      default_host = 'armv7-unknown-linux-gnueabihf'
-    else
-      default_host = "#{ARCH}-unknown-linux-gnu"
-    end
+    default_host = case ARCH
+                   when 'aarch64', 'armv7l'
+                     'armv7-unknown-linux-gnueabihf'
+                   else
+                     "#{ARCH}-unknown-linux-gnu"
+                   end
```
uberhacker, line 34: This could also work as an immediate if statement:
```
default_host = ARCH == 'aarch64' || ARCH == 'armv7' ? 'armv7-unknown-linux-gnueabihf' : "#{ARCH}-unknown-linux-gnu"
```

uberhacker:
I can test the precompiled binaries after work this evening.

uberhacker (approved):

packages/rust.rb
satmandu: done

uberhacker:
On x86_64, I get:
```
$ rustc -V
error: no override and no default toolchain set
$ cargo -V
error: no override and no default toolchain set
```
Not sure why but I didn't need to do anything special to get these executables to work on i686 and armv7l.

uberhacker:
After `crew reinstall rust`, I now get:
```
$ rustc -V
error: could not create home directory: '/home/chronos/user/.rustup': File exists (os error 17)
```
Do we need to add a `def self.remove` section?

satmandu:
I'm wondering if there is an issue with `NOSYMFOLLOW` for `x86_64` and the .rustup file? You're not seeing it on the other architectures because they are pre-M90 and therefore don't have that `feature`?

satmandu:
This might be problematic?
```
chronos@cheekon-x86_64 ~ $ ls -aFl ~/.rustup
lrwxrwxrwx 1 chronos chronos 23 Aug 12 17:04 /home/chronos/user/.rustup -> /usr/local/share/rustup/
chronos@cheekon-x86_64 ~ $
```

uberhacker:
Yes, I believe you are correct about the pre-90 release.  armv7l is 84 and i686 is 58.  I wonder if we can do something with xdg_base to fix this?

satmandu:
@saltedcoffii @supechicken thoughts?

uberhacker:
FWIW, I was able to fix the issue with `rm -rf /usr/local/share/rustup ~/.rustup && crew reinstall rust` so it looks like we just need a `def self.remove` section to clean things up after `crew remove rust`.

satmandu:
@uberhacker Does this work for you?

uberhacker:
I think `def self.remove` should also include (with a prompt, of course) the following:
```
rm -rf #{HOME}/.rustup #{CREW_PREFIX}/rustup #{HOME}/.cargo #{CREW_PREFIX}/cargo
```

satmandu:
@uberhacker Done.

packages/rust.rb
```diff
@@ -71,4 +66,20 @@ def self.install
...
+      print "\nWould you like to remove #{config_dir}? [y/N] "
+      case $stdin.getc
+      when 'y', 'Y'
+        FileUtils.rm_rf config_dir
+        puts "#{config_dir} removed.".lightred
+      else
+        puts "#{config_dir} saved.".lightgreen
+      end
+    end
+  end
```
uberhacker, line 84: This doesn't work like you might expect.  The first prompt works but the second prompt is bypassed.  Check for yourself to see what I mean.  Take a look at github_desktop.rb for an example of what works.

uberhacker: Also, a single prompt would be less verbose than four separate prompts. :)

satmandu: I can't actually test this as this is dependent upon NOSYMFOLLOW keeping those dirs from being deleted. Does this work for you?

uberhacker: Nice!

PR #6074, opened 2021-08-15
Update gtk2 from 2.24.33-1 to 2.24.33-2

### LINKED ISSUE #6070: [gtk3] Conflict with gtk2
When building gtk3 from source with gtk2 installed, I get the following:
```
Checking for FHS3 compliance...
Exiting due to above errors.
Error: There is a conflict with the same file in another package.

gtk2 package conflicts below:
/usr/local/bin/gtk-update-icon-cache
```
How should we handle this conflict?  Does it make sense to rename to `gtk2-update-icon-cache`?

uberhacker:
- Add binaries for gtk2
- Add i686 binaries for gtk3 and gtk4
- Add i686 binary for gtkmm4

Fixes #6070.

### DIFF
--- a/packages/gtk2.rb
+++ b/packages/gtk2.rb
@@ -3,23 +3,23 @@ require 'package'
 class Gtk2 < Package
   description 'GTK+ is a multi-platform toolkit for creating graphical user interfaces.'
   homepage 'https://www.gtk.org/'
-  version '2.24.33-1'
+  version '2.24.33-2'
   license 'LGPL-2.1'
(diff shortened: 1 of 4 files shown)


satmandu:
Shouldn't self-remove only remove gtk2-icon-cache... in case gtk3 is still installed?

uberhacker:
No, it should delete the symlink since that is not in the package metadata.  gtk2-update-icon-cache is removed when the package is removed.

PR #6075, opened 2021-08-16
Grep 3.7

satmandu:
- Apparently fixes a huge performance regression in 3.6

Builds & Works properly (in container):
- [x] x86_64
- [x] armv7l
- [x] i686 (@uberhacker build please?)

Known to work properly on actual hardware:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/grep.rb
+++ b/packages/grep.rb
@@ -6,30 +6,29 @@ require 'package'
 class Grep < Package
   description 'A string search utility'
   homepage 'https://www.gnu.org/software/grep/'
-  version '3.6-2'
+  version '3.7'
   license 'GPL-3+'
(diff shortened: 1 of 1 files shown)


packages/grep.rb
```diff
@@ -9,7 +9,7 @@ class Grep < Package
   version '3.7'
   license 'GPL-3+'
   compatibility 'all'
-  source_url 'https://ftp.gnu.org/gnu/grep/grep-3.7.tar.xz'
+  source_url 'https://ftpmirror.gnu.org/grep/grep-3.7.tar.xz'
```
uberhacker, line 12: :100:

PR #6080, opened 2021-08-19
Use more standard compile flags

satmandu:
- We are now using upx to reduce package size. I suspect this is a vast improvement over using compile-time shrinking flags (at the expense of initial slowdown for upx decompression.)
- So use more standard compile flags of `-O2` and for meson/cmake the `Release' build target, which should give us faster binaries and hopefully reduce compile issues.

Works properly:
- [x] x86_64
- [x] armv7l


uberhacker:
I agree release should be used over minsize.

PR #6091, opened 2021-08-22
Only need qtbase for qt apps

satmandu:
- qtbase is huge. Don't pull it in via avahi.

Works properly:
- [x] x86_64


uberhacker:
Good call.  We could probably change on every package that has this dependency.

PR #6099, opened 2021-08-22
xfce4_devtool => 4.17.0

supechicken:
Built successfully on `x86_64`.  Also builds on armv7l and i686.  All pre-built binaries added.


satmandu:
This looks fine to me but @uberhacker will have to generate binaries.

PR #6119, opened 2021-08-30
Delete crew_profile.rb

supechicken:
This package is no longer needed as `crew-profile-base` replaced it


uberhacker:
Good call.

PR #6130, opened 2021-09-01
libsdl2 -> 2.0.16

satmandu:
Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686 (@uberhacker  build please?)


packages/libsdl2.rb
```diff
@@ -29,9 +27,11 @@ class Libsdl2 < Package
   depends_on 'nasm' => :build
 
   def self.build
+    system 'filefix'
```
uberhacker, line 30: Can we move this to `def self.patch`?

satmandu: done

PR #6120, opened 2021-08-30
libcurl -> 7.78.0

satmandu:
- Just an update of non-static libcurl
- needs armv7l, i686 builds

Works properly:

- [x] armv7l
- [x] i686
- [x] x86_64

PR #6131, opened 2021-09-01
Tesseract rebuild with newer libcurl

satmandu:
- this is dependent upon https://github.com/skycocker/chromebrew/pull/6120

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [ ] i686 (@uberhacker  build please?)

### DIFF
--- a/packages/tesseract.rb
+++ b/packages/tesseract.rb
@@ -34,7 +34,6 @@ class Tesseract < Package
   depends_on 'libpng'
   depends_on 'libtiff'
   depends_on 'pango'
-  depends_on 'pango' => :build
 
   def self.build
     system '[ -x configure ] || ./autogen.sh'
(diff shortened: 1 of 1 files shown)


packages/tesseract.rb
```diff
@@ -39,6 +38,7 @@ class Tesseract < Package
```
uberhacker, line 38: Duplicate pango lines?

packages/tesseract.rb
```diff
@@ -31,6 +29,7 @@ class Tesseract < Package
   depends_on 'glib'
```
uberhacker, line 29: Duplicate cairo lines?

packages/tesseract.rb
satmandu: done

packages/tesseract.rb
satmandu: done

PR #6132, opened 2021-09-01
mpv rebuild, and made hw accel default.

satmandu:
- Enabled sdl2 in build, and also enabled hw acceleration by default, which makes a dramatic difference in performance on x86_64.
- this is dependent upon https://github.com/skycocker/chromebrew/pull/6120 https://github.com/skycocker/chromebrew/pull/6130 and https://github.com/skycocker/chromebrew/pull/6131

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [ ] i686 (@uberhacker  build please?)

### DIFF
--- a/packages/mpv.rb
+++ b/packages/mpv.rb
@@ -48,6 +48,7 @@ class Mpv < Package
   depends_on 'libxrandr' # R
   depends_on 'libxss' # R
   depends_on 'libxv' # R
+  depends_on 'llvm' # R
   depends_on 'luajit' # R
   depends_on 'mesa' # R
   depends_on 'mujs' # R
--- a/packages/mpv.rb
+++ b/packages/mpv.rb
@@ -3,23 +3,22 @@ require 'package'
 class Mpv < Package
   description 'Video player based on MPlayer/mplayer2'
   homepage 'https://mpv.io/'
-  version '0.33.1'
+  @_ver = '0.33.1'
+  version "#{@_ver}-1"
   license 'LGPL-2.1+, GPL-2+, BSD, ISC and GPL-3+'
   compatibility 'all'
-  source_url "https://github.com/mpv-player/mpv/archive/v#{version}.tar.gz"
+  source_url "https://github.com/mpv-player/mpv/archive/v#{@_ver}.tar.gz"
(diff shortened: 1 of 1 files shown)


packages/mpv.rb
```diff
@@ -3,23 +3,22 @@
 class Mpv < Package
   description 'Video player based on MPlayer/mplayer2'
   homepage 'https://mpv.io/'
-  version '0.33.1'
+  @_ver = '0.33.1'
+  version "#{@_ver}-1"
   license 'LGPL-2.1+, GPL-2+, BSD, ISC and GPL-3+'
   compatibility 'all'
```
uberhacker, line 9: I think we can drop i686 support since sommelier doesn't work and this project requires it.

satmandu: Should we leave it as compatible and just not add binaries?

I don't know that it won't build... but let the user deal with that.

As we move to deprecate i686, it might make sense to leave i686 as a specifically excluded outlier, rather than manually change packages to everything BUT i686.

uberhacker: I agree with this line of thinking.  It would be really nice if we could get sommelier working in i686 again.  Might have to go way back in the repo tree to find a commit that works, however.

satmandu:
(if we're not adding i686 binaries, then this is ready to go.)

packages/mpv.rb
```diff
@@ -39,6 +38,7 @@ class Mpv < Package
   depends_on 'libdvdnav' # R
```
uberhacker, line 38: Duplicate libcdio lines?

packages/mpv.rb
```diff
@@ -48,6 +48,7 @@ class Mpv < Package
   depends_on 'libxrandr' # R
   depends_on 'libxss' # R
   depends_on 'libxv' # R
+  depends_on 'llvm' # R
```
uberhacker, line 51: Are you sure this is a runtime dependency?

satmandu: Yes. Just did a fresh install and discovered this.

image

packages/mpv.rb
satmandu: done

packages/mpv.rb
uberhacker: We really need a llvm_runtime package.  The llvm proper package is UUUGGGE.

PR #6136, opened 2021-09-04
Add chafa

satmandu:
Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/chafa.rb
+++ b/packages/chafa.rb
@@ -8,18 +8,20 @@ class Chafa < Package
   homepage 'https://hpjansson.org/chafa/'
   version '1.6.1'
   license 'LGPL'
-  compatibility 'x86_64 aarch64 armv7l'
+  compatibility 'all'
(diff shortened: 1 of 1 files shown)


packages/chafa.rb
```diff
@@ -0,0 +1,45 @@
...
+# https://github.com/archlinux/svntogit-community/raw/packages/chafa/trunk/PKGBUILD
+
+require 'package'
+
+class Chafa < Package
+  description 'Image-to-text converter supporting a wide range of symbols and palettes, transparency, animations, etc.'
+  homepage 'https://hpjansson.org/chafa/'
+  version '1.6.1'
+  license 'LGPL'
+  compatibility 'x86_64 aarch64 armv7l'
```
uberhacker, line 11: Since this is a cli program, shouldn't this be 'all'?

satmandu: Yes, but it doesn't have binaries for i686. Feel free to add and modify...

PR #6144, opened 2021-09-09
Add `cros_adapta` package

supechicken:
This is the default gtk+ themes on Crostini

Tested on `x86_64`

### DIFF
--- a/packages/cros_adapta.rb
+++ b/packages/cros_adapta.rb
@@ -9,6 +9,19 @@ class Cros_adapta < Package
   source_url 'https://chromium.googlesource.com/chromiumos/third_party/cros-adapta.git'
   git_hashtag 'eb6d8c1832b9181926df107faf41a80887fd982c'
 
+  binary_url({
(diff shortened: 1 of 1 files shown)


packages/cros_adapta.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def self.install
+    @_dest_dir = "#{CREW_DEST_PREFIX}/share/themes/CrosAdapta/"
+    FileUtils.mkdir_p [ @_dest_dir, "#{CREW_DEST_PREFIX}/etc/env.d/" ]
+    FileUtils.mv Dir['gtk*/'], @_dest_dir
+    File.write "#{CREW_DEST_PREFIX}/etc/env.d/cros_adapta", @_env
+  end
+
+  def self.postinstall
+    puts nil, <<~EOT.lightblue
```
uberhacker, line 28: Is the nil necessary?  Does `puts <<~EOT.lightblue` work?  Just curious.

satmandu:
Also, just saying, but it would be useful to have a no-arch option for this...

packages/cros_adapta.rb
supechicken: It acts like a new line like `\n`

PR #6145, opened 2021-09-09
crew: modify upx logic to use concurrent-ruby

satmandu:
- Using `upx` in crew on multi-core machines is slow when multiple binaries need to be compressed.
- This PR adds the use of `concurrent-ruby` to create a thread pool to spawn multiple simultaneous upx instances, up to `CREW_NPROC`. This way more than one core is used during this process, which should speed up builds.
- This also cleans up the error messages from the long shell code being used in the upx code block.

Works properly:
- [x] x86_64
- [x] armv7l (in container)

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1011,10 +1011,36 @@ def shrink_dir(dir)
         @execfiles = %x[find . -executable -type f ! \\( -name \"*.so*\" -o -name \"*.a\" -o -name \"Xwayland.elf\" -o -name \"sommelier.elf\" \\) -exec sh -c \"file -i \'{}\' | grep -q \'executable; charset=binary\'\" \\; -print].chomp
         unless @execfiles.empty? or @execfiles.nil?
           puts "Using upx to shrink binaries."
-          @execfiles.each_line do |execfile|
-            system "upx --best -k --overlay=skip #{execfile} && \
-            \( upx -t #{execfile} && rm #{execfile}.~ || mv #{execfile}.~ #{execfile}\)"
+          begin
+              gem 'concurrent-ruby'
+          rescue Gem::LoadError
+              puts ' -> install gem ' + 'concurrent-ruby'
+              Gem.install('concurrent-ruby')
+              gem 'concurrent-ruby'
           end
+          require 'concurrent-ruby'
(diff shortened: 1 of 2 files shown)


bin/crew
```diff
@@ -1011,10 +1011,36 @@ def shrink_dir(dir)
...
+          begin
+              gem 'concurrent-ruby'
+          rescue Gem::LoadError
+              puts ' -> install gem ' + 'concurrent-ruby'
+              Gem.install('concurrent-ruby')
+              gem 'concurrent-ruby'
+          end
+          require 'concurrent-ruby'
+          pool = Concurrent::ThreadPoolExecutor.new(
+            min_threads: CREW_NPROC,
```
uberhacker, line 1023: Would it be safer to put `min_threads: 1,` here?

bin/crew
```diff
@@ -1011,10 +1011,36 @@ def shrink_dir(dir)
...
+            max_threads: CREW_NPROC,
+            max_queue: 0, # unbounded work queue
+            fallback_policy: :caller_runs
+          )
           @execfiles.each_line do |execfile|
-            system "upx --best -k --overlay=skip #{execfile} && \
-            \( upx -t #{execfile} && rm #{execfile}.~ || mv #{execfile}.~ #{execfile}\)"
+            puts "Attempting to compress #{execfile}".lightblue
```
uberhacker, line 1029: Add `...`?

bin/crew
```diff
@@ -1011,10 +1011,36 @@ def shrink_dir(dir)
...
+            pool.post do
+              begin
+                system "upx --best -k --overlay=skip #{execfile}"
+                system "upx -t #{execfile}" and FileUtils.rm "#{execfile}.~"
```
uberhacker, line 1033: Replace `and` with `&&`?  Not sure `and` works like you might expect in the shell.

bin/crew
satmandu: Sure.

bin/crew
satmandu: done

bin/crew
uberhacker: Doesn't the system command execute a shell?  Maybe I'm missing something?

PR #6177, opened 2021-09-13
Autoconf213: Add m4 files

satmandu:
- `js78` build complains of missing `autoconf.m4` in the autoconf213 package, so this fixes that.

Works properly:
- [x] x86_64

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/autoconf213.rb
+++ b/packages/autoconf213.rb
@@ -3,23 +3,21 @@ require 'package'
 class Autoconf213 < Package
   description 'A GNU tool for automatically configuring source code Legacy 2.1x version'
   homepage 'https://www.gnu.org/software/autoconf/'
-  version '2.13'
+  version '2.13-1'
   license 'GPL-2'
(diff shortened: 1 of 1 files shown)


packages/autoconf213.rb
```diff
@@ -3,35 +3,38 @@
...
+     x86_64: '98f744a5a7ae84583997e57ec03dcc6ce1f962b6316bd6d4c54a6c3ebaad5a36'
   })
 
   def self.build
     system "env #{CREW_ENV_OPTIONS} \
       ./configure #{CREW_OPTIONS} \
       --infodir=#{CREW_PREFIX}/share/info \
       --program-suffix=-2.13 \
-      -datadir=#{CREW_PREFIX}/share/autoconf213"
+      --datadir=#{CREW_PREFIX}/share/autoconf213"
```
uberhacker, line 28: Good catch.

PR #6183, opened 2021-09-13
combine nss & nspr (-> 3.69.1, 4.32) as they share source package

satmandu:
- Underlies build for `js78`
- Since they share a source package this combined build also lets us avoid issues with the packages being out of sync.
- fix `exec_prefix` in nss package-config file.

Works properly:
- [x] x86_64

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/nss.rb
+++ b/packages/nss.rb
@@ -25,11 +25,7 @@ class Nss < Package
   depends_on 'gyp_next' => :build
 
   def self.build
-    @build_64 = if ARCH == 'x86_64'
-                  '1'
-                else
-                  '0'
-                end
+    @build_64 = ARCH == 'x86_64' ? '1' : '0'
     @arch_cflags = ''
     @arch_ldflags = @arch_cflags
 
--- a/packages/nss.rb
+++ b/packages/nss.rb
@@ -14,13 +14,13 @@ class Nss < Package
   binary_url({
     aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/nss/nss.3.69.1.nspr.4.32_armv7l/nss-nss.3.69.1.nspr.4.32-chromeos-armv7l.tpxz',
      armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/nss/nss.3.69.1.nspr.4.32_armv7l/nss-nss.3.69.1.nspr.4.32-chromeos-armv7l.tpxz',
-       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/nss/nss.3.69.1.nspr.3.69.1_i686/nss-nss.3.69.1.nspr.3.69.1-chromeos-i686.tar.xz',
+       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/nss/nss.3.69.1.nspr.4.32_i686/nss-nss.3.69.1.nspr.4.32-chromeos-i686.tar.xz',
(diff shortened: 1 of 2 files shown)


packages/nss.rb
```diff
@@ -3,70 +3,66 @@
...
-    @arch_cflags = if ARCH == 'armv7l'
-                     ''
-                   else
-                     '-flto=auto'
-                   end
+    @build_64 = if ARCH == 'x86_64'
+                  '1'
+                else
+                  '0'
+                end
```
uberhacker, line 32: Simplify to: `@build_64 = ARCH == 'x86_64' ? '1' : '0'`?

satmandu: done

packages/nss.rb
```diff
@@ -3,70 +3,64 @@
...
-                      else
-                        '0'
-                      end
-    @arch_cflags = if ARCH == 'armv7l'
-                     ''
-                   else
-                     '-flto=auto'
-                   end
+    @build_64 = ARCH == 'x86_64' ? '1' : '0'
+    @arch_cflags = ''
```
supechicken, line 31: `@arch_ldflags = @arch_cflags = ''`?

packages/nss.rb
```diff
@@ -3,70 +3,64 @@
...
     system "sed nss/pkg/pkg-config/nss.pc.in \
     -e 's,%libdir%,#{CREW_LIB_PREFIX},g' \
     -e 's,%prefix%,#{CREW_PREFIX},g' \
-    -e 's,%exec_prefix%,#{CREW_PREFIX}/bin,g' \
+    -e 's,%exec_prefix%,#{CREW_PREFIX},g' \
     -e 's,%includedir%,#{CREW_PREFIX}/include/nss,g' \
-    -e 's,%NSPR_VERSION%,$(pkg-config --modversion nspr),g' \
-    -e 's,%NSS_VERSION%,#{@_ver},g' | \
+    -e 's,%NSPR_VERSION%,#{@nspr_ver},g' \
+    -e 's,%NSS_VERSION%,#{@nss_ver},g' | \
```
supechicken, line 63: Redirect to file instead of using install command?

packages/nss.rb
satmandu: Sure? That's debug code at this point, since LTO was previously being used for non-armv7l, but it just causes issues for everything. Maybe it can be revisited for a later revision.

satmandu: (At which point it would likely have to be broken out again if there is an arch where it does work right.)

packages/nss.rb
satmandu: Just suggest the replacement code. :)

packages/nss.rb
supechicken: Got it, never mind :)

packages/nss.rb
supechicken: Like this
`sed … > #{CREW_DEST_LIB_PREFIX}/pkgconfig/nss.pc`

uberhacker: Can fix this on the next version?  I don't want to build the i686 binary again.  It takes forever. :)

supechicken: It doesn’t require a rebuild 😅

uberhacker: But a true test would require it. ;)

PR #6194, opened 2021-09-15
[`docbook*`]: Remove bash dependency

supechicken:
Seems the newer version of `bash` is only used for executing scripts from the package, might not be necessary.

**Not tested yet**


satmandu:
Nice. Feel free to also remove the debug output code like `puts "#{CREW_PREFIX}/etc/xml/catalog exists"` or wrap that with a verbose flag... we really don't need that output by default.

satmandu (approved):

packages/docbook_xsl.rb
```diff
@@ -25,10 +25,9 @@ class Docbook_xsl < Package
 
   depends_on 'docbook_xml'
   depends_on 'xmlcatmgr'
-  depends_on 'bash'
 
   def self.patch
-    system 'curl -OLf "https://github.com/archlinux/svntogit-packages/raw/packages/docbook-xsl/trunk/765567_non-recursive_string_subst.patch"'
+    system 'curl -OLsf "https://github.com/archlinux/svntogit-packages/raw/packages/docbook-xsl/trunk/765567_non-recursive_string_subst.patch"'
```
uberhacker, line 30: Would it make sense to exclude `-s` if `-v` is added to any command?  Adding `#` might be helpful too, although this is a tiny patch so it probably wouldn't make a difference.  Just a thought for future packages.

supechicken:
### New changes after edd594b
- Merge multiple `system` calls into a single call with heredoc
- Use `downloader.rb` instead of `curl`

```
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=remove_bash_dep CREW_TESTING=1 crew update
crew reinstall -s docbook_xsl docbook_xml50 docbook_xml51 docbook_xsl_nons
```

packages/docbook_xml50.rb
```diff
@@ -189,20 +206,21 @@ def self.install
       done
     ADDFILES_HEREDOC
     IO.write('add_files.sh', @ADDFILES_SH, perm: 0o755)
-    system "#{CREW_PREFIX}/bin/bash -x ./add_files.sh | true"
+    system 'bash -x ./add_files.sh || true'
```
uberhacker, line 209: Is calling bash necessary?  I suppose it would be necessary if `add_files.sh` is not executable.

supechicken: Original yes (although `add_files.sh` is executable, a shebang line is missing, so it need to be involved by a shell)

The missing shebang lines were added now (the file will be executed directly now)

supechicken:
Be noticed that there are 2 conflicts between `docbook_xml50` and `docbook_xml51`:
```
Error: There is a conflict with the same file in another package.
docbook_xml51:/usr/local/share/xml/docbook/stylesheet/docbook5/db4-upgrade.xsl
docbook_xml51:/usr/local/bin/db4-entities.pl
Exiting due to above errors.
```

satmandu:
> Be noticed that there are 2 conflicts between `docbook_xml50` and `docbook_xml51`:
> 
> ```
> Error: There is a conflict with the same file in another package.
> docbook_xml51:/usr/local/share/xml/docbook/stylesheet/docbook5/db4-upgrade.xsl
> docbook_xml51:/usr/local/bin/db4-entities.pl
> Exiting due to above errors.
> ```

Maybe just keep the newer one?

satmandu:
Maybe we should also put the puts in these files behind debug flags? We don't need to see them during normal installs.

satmandu:
Is this otherwise ready? It would be nice to get this merged...

satmandu:
Made my suggested edits and added binaries. I think this is good to go...

PR #6223, opened 2021-09-22
Mesa -> 21.2.2

### LINKED ISSUE #6218: [Suggestion] Add mesa patch for nyan boards
There are a few chromebooks using the nyan board, and until just recently I have been searching for a way to get hardware acceleration working on them. I recently came across https://github.com/hexdump0815/linux-mainline-tegra-k1-kernel/issues/4; which features a relatively new patch that makes mesa/opengl/glamor work _at all_ on these boards.

The patch is just a hack in it's current state, and there are still some issues, but it's better then nothing at all.

I know the nyan boards are no longer offically supported, but adding the patch wouldn't cause any issues, and I have confirmed proof that it does infact fix the issue.

The proposed patch is found below
> https://github.com/hexdump0815/mesa-etc-build/blob/master/tegra-hack.patch

Additionally a issue was opened on the official freedesktop mesa repo
> https://gitlab.freedesktop.org/mesa/mesa/-/issues/3505

---

Adding the patch alone wouldn't be enough to get hardware acceleration with sommelier fully working, but it would bring us a lot closer. With some work to make sommelier use the nouveau drive with glamor, (and possibly an updated kernel) it might work.

But I still see no reason to not add this patch, as it would not cause any issues, and only improve support and usability of the Nyan boards.

satmandu:
Fixes #6218 

- Update and suggested patch for armv7l

Works properly:
- [x] x86_64
- [x] armv7l? (Can you check this @ThatGeekyWeeb )
- [x] i686 (needs build)

Testing:
```
export CREW_TESTING_BRANCH=mesa_21.2.2
export CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git
CREW_TESTING=1 crew update
yes | crew install mesa_utils
crew upgrade mesa
glxinfo -B
crew update
```

### DIFF
--- a/packages/mesa.rb
+++ b/packages/mesa.rb
@@ -13,11 +13,13 @@ class Mesa < Package
   binary_url({
     aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/mesa/21.2.2_armv7l/mesa-21.2.2-chromeos-armv7l.tpxz',
      armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/mesa/21.2.2_armv7l/mesa-21.2.2-chromeos-armv7l.tpxz',
+       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/mesa/21.2.2_i686/mesa-21.2.2-chromeos-i686.tar.xz',
      x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/mesa/21.2.2_x86_64/mesa-21.2.2-chromeos-x86_64.tpxz'
   })
   binary_sha256({
     aarch64: '0cbce907c2515100c624643c1243876c3c0cc47c6c531ca253e6d97b3d13f438',
      armv7l: '0cbce907c2515100c624643c1243876c3c0cc47c6c531ca253e6d97b3d13f438',
+       i686: '7d28e75839baed8ceb15e677eba566b6501819c03cbe9f8067857d79a0bb9a7e',
      x86_64: '82f5de3eb10852a5fdd40ac38b85ed3c4c17c9f7a67199417a9ebe8021552d2e'
   })
 
--- a/packages/mesa.rb
+++ b/packages/mesa.rb
@@ -3,7 +3,7 @@ require 'package'
 class Mesa < Package
   description 'Open-source implementation of the OpenGL specification'
   homepage 'https://www.mesa3d.org'
-  @_ver = '21.2.1'
+  @_ver = '21.2.2'
   version @_ver
   license 'MIT'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


ThatGeekyWeeb:
I'll test this when I can, but for the moment I am working on getting hardware rendering to work at all via xorg

---
None the less the repo from which the patch came, contains the proof itself, I just need to figure out *how* to get xorg to work...

Also it should be noted that this patch will do nothing unless `MESA_LOADER_DRIVER_OVERRIDE` is set to `nouveau` for testing with nyan boards; I know I'm likely the only one with a nyan board in this repo however; so I will likely have to do this testing myself

ThatGeekyWeeb:
From what I have read, xorg-server will need to basically be set to build from git in order for the patch to make any difference

See: 
> https://github.com/hexdump0815/imagebuilder/blame/main/systems/chromebook_nyan/readme.md#L50

---

I didn't see it being added this fast; there is still alot of work I need to do since all the info is heavily scattered

satmandu:
@ThatGeekyWeeb 

See  https://www.cnx-software.com/2019/09/01/how-to-sandbox-an-arm64-gcc-on-aarch64-hardware-with-armv7-userspace/ 

Looks like it wouldn't be too hard to bootstrap an aarch64 chroot for aarch64 chromeos machines. (Also might be a workable way to do a hw accelerated mesa maybe on aarch64 machines, if just sommelier & mesa were run out of an aarch64 chroot.)

ThatGeekyWeeb:
> @ThatGeekyWeeb
> 
> See https://www.cnx-software.com/2019/09/01/how-to-sandbox-an-arm64-gcc-on-aarch64-hardware-with-armv7-userspace/
> 
> Looks like it wouldn't be too hard to bootstrap an aarch64 chroot for aarch64 chromeos machines. (Also might be a workable way to do a hw accelerated mesa maybe on aarch64 machines, if just sommelier & mesa were run out of an aarch64 chroot.)

huh? nyan boards are armv7l only; and all the issues are coming from the fact that the nouveau driver is broken / doesn't work properly / needs xorg-server-git

and I'm doing my work from a raw alpine install using the kpart signed kernel setup; the issue is that if the driver won't work for raw xorg; it's highly unlikely it will work for sommelier; I'm going to setup a temp crew install via a mounted drive to /usr/local and do the proposed test; but I doubt acceleration will work right away; unless by some miracle it works right away, I suggest this pr is converted to a draft as it currently only allows to the driver to work if _EVERYTHING ELSE_ works - which from my raw xorg testing; it doesn't.

satmandu:
Ah! My mistake.

ThatGeekyWeeb:
All good, I just need a bit more time to test everything, and figure out why not even raw xorg works

ThatGeekyWeeb:
I've given another look at the imagebuilder repo (hexdump0815/imagebuilder); and it appears the issue is this;

```text
  - seems to be related to drm initialization order
    - working:
      ```
      [drm] Initialized tegra 0.0.0 20120330 for drm on minor 0
      drm drm: [drm] fb0: tegradrmfb frame buffer device
      ```
    - not working:
      ```
      drm drm: [drm] fb0: tegradrmfb frame buffer device
      [drm] Initialized tegra 0.0.0 20120330 for drm on minor 0
      ```
    - next would be to find a way to force proper order
```

> https://github.com/hexdump0815/imagebuilder/blob/3369bf93a44ca0c5550b9f983739f2f25675036a/systems/chromebook_nyan/todo.txt#L6 
This is referencing dmesg and how the drm drivers seem to initialize; I believe the nouveau driver will work fine with chromeos; if there isn't an issue with the xorg-server version

uberhacker:
@ThatGeekyWeeb: Is this ready or do you want to do some testing?

ThatGeekyWeeb:
Go ahead and merge it, there might still be some kernel related issues but thats nothing I can fix here, this is a good step, I think it's fine to merge.

ThatGeekyWeeb (approved):

PR #6225, opened 2021-09-23
llvm 12.0.1:  new i686 binaries

satmandu:
- rebuild of i686 binaries

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686 ?


uberhacker:
Works!  You nailed it.

PR #6229, opened 2021-09-23
crew: Handle attempted removal of essential files

satmandu:
- Upgrading libssp, gcc10, gcc11, etc breaks crew because removing the old package removes libraries ruby is using.
- This PR allows crew to ignore deletions of those files.
- Reinstalling or upgrading such packages no longer breaks crew.

Works properly:
- [x] x86_64

e.g.
previously:
```
bash-5.1$ crew remove libssp
Libssp removed!
bash-5.1$ crew install libssp
ruby: error while loading shared libraries: libssp.so.0: cannot open shared object file: No such file or directory
bash-5.1$ 
```
Now:
image

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.16.7'
+CREW_VERSION = '1.16.8'
 
 ARCH_ACTUAL = `uname -m`.strip
 # This helps with virtualized builds on aarch64 machines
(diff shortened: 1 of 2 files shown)


lib/const.rb
```diff
@@ -126,3 +126,7 @@
 
 CREW_FIRST_PACKAGES = %w[libssh curl git pixz shared_mime_info]
 CREW_LAST_PACKAGES = %w[ghc mandb gtk3 gtk4 sommelier]
```
uberhacker, line 128: Bump version?

satmandu: done

PR #6236, opened 2021-09-24
binutils: remove upx workaround

satmandu:
- remove armv7l upx workaround, as this is a binutils bug, not a upx bug which is causing the issues with `size` reporting incorrect information for armv7l upx compressed binaries.

Works properly:
- [x] armv7l

### DIFF
--- a/packages/binutils.rb
+++ b/packages/binutils.rb
@@ -66,13 +66,6 @@ class Binutils < Package
   end
 
   def self.install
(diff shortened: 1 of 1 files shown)


uberhacker:
Does this require a new armv7l binary?

satmandu:
We could rebuild it, but I am not sure it is necessary.

PR #6244, opened 2021-09-25
Add Android SDK CLI tools package

uberhacker:
Tested on x86_64.

### DIFF
--- a/packages/android_sdk_tools.rb
+++ b/packages/android_sdk_tools.rb
@@ -17,7 +17,7 @@ class Android_sdk_tools < Package
     FileUtils.mv Dir['*'], CREW_DEST_PREFIX + '/share/android-sdk-tools'
     Dir["#{CREW_DEST_PREFIX}/share/android-sdk-tools/bin/*"].each do |filename|
       binary = File.basename(filename)
-      FileUtils.ln_s "#{CREW_PREFIX}/share/android-sdk-tools/bin/#{binary}", "#{CREW_DEST_PREFIX}/bin/#{binary}"
+      FileUtils.ln_s "../share/android-sdk-tools/bin/#{binary}", "#{CREW_DEST_PREFIX}/bin/#{binary}"
     end
   end
 
--- a/packages/android_sdk_tools.rb
+++ b/packages/android_sdk_tools.rb
@@ -1,0 +1,28 @@
+require 'package'
+
+class Android_sdk_tools < Package
+  description 'Android SDK CLI tools'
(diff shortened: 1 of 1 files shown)


packages/android_sdk_tools.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  depends_on 'jdk11'
+
+  def self.install
+    FileUtils.mkdir_p CREW_DEST_PREFIX + '/bin'
+    FileUtils.mkdir_p CREW_DEST_PREFIX + '/share/android-sdk-tools'
+    FileUtils.mv Dir['*'], CREW_DEST_PREFIX + '/share/android-sdk-tools'
+    Dir["#{CREW_DEST_PREFIX}/share/android-sdk-tools/bin/*"].each do |filename|
+      binary = File.basename(filename)
+      FileUtils.ln_s "#{CREW_PREFIX}/share/android-sdk-tools/bin/#{binary}", "#{CREW_DEST_PREFIX}/bin/#{binary}"
```
satmandu, line 20: will doing
```
Dir.chdir "#{CREW_DEST_PREFIX}/bin" do
  FileUtils.ln_s "../share/android-sdk-tools/bin/#{binary}", "#{binary}"
end
```
avoid symlinks issues?

uberhacker: I'm not seeing an issue with the binaries that have the absolute symlinks.  Did you get a chance to give it a try yet?

uberhacker: You are right.  I had `export CREW_SHRINK_ARCHIVE=0` in my `~/.bashrc`.  Once I removed, the symlinks broke.

uberhacker: Do you think it makes sense to not run symlink in `CREW_DEST_DIR` while building or installing packages?

uberhacker: I went ahead and changed to relative symlinks for the time being.

satmandu: symlinks I think fixes issues for some packages? But the relative symlinks avoids the issue.

uberhacker: What exactly does it fix?  Like upx, it introduces more problems than it solves.

satmandu: That's a valid question. symlinks came in through https://github.com/skycocker/chromebrew/pull/5837

The idea was to remove absolute symlinks and make them relative, so that if CREW_PREFIX is changed, programs still work.

upx is another issue.... but it does dramatically reduce the size of packages.

PR #6248, opened 2021-09-26
coreutils -> 9.0

satmandu:
Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686 (@uberhacker needs verification)


uberhacker:
I tested a few binaries and they all worked.  Good to go!

PR #6261, opened 2021-09-27
libxml2 needs gcc11 dep for installs to succeed

satmandu:
-  xltproc is used by the install process, and it depends upon libxml2. In turn, libxml2 depends upon gcc11's `#{CREW_LIB_PREFIX}/libgcc_s.so.1`
- While this adds gcc11 installs, hopefully, this can be reduced out soon, as this library is small relative to the size of gcc11.

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/libxml2.rb
+++ b/packages/libxml2.rb
@@ -24,6 +24,7 @@ class Libxml2 < Package
 
   depends_on 'zlibpkg'
   depends_on 'readline'
+  depends_on 'gcc11'
 
   def self.patch
     # Fix encoding.c:1961:31: error: ‘TRUE’ undeclared (first use in this function)
--- a/packages/buildessential.rb
+++ b/packages/buildessential.rb
@@ -14,7 +14,7 @@ class Buildessential < Package
 
   # install first to get ldconfig
   depends_on 'glibc'
-  depends_on 'gcc11'
+  # depends_on 'gcc11'
   depends_on 'gmp'
   depends_on 'mpfr'
   depends_on 'mpc'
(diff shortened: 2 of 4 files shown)


packages/libxml2.rb
```diff
@@ -24,6 +24,7 @@ class Libxml2 < Package
 
   depends_on 'zlibpkg'
   depends_on 'readline'
+  depends_on 'gcc11'
```
uberhacker, line 27: Since libxml2 is a core package, that means gcc11 is now in core too.  Can you please update `tools/core_packages.txt`?  That also means you can remove it from buildessential.  Try a fresh install and then run `crew list installed` to see if it matches.  Better yet - figure out a way to avoid gcc11 as a dependency. :)

uberhacker: I also forgot to mention it needs to be added to core.rb.

satmandu: I'm going to do a fresh install, but my build environment script adds some packages, so I can not at the moment compare `crew list installed` to the core packages list...

uberhacker: Why wouldn't you want to match your build environment?  That seems like a potential recipe for false positive tests.

satmandu: meaning that I also install: `util_linux` `coreutils` `psmisc` `bash` on my build environment, because for instance, I use `linux32` for the armv7l container...

satmandu: (not having to add more packages after I start the container speeds up builds.)

uberhacker: Those have the potential to have a meaningful impact for testing packages.

uberhacker: We're supposed to be reducing the size of core, not expanding it. :)

PR #6266, opened 2021-09-29
Default Rubocop linter config for Chromebrew Take 2.5

satmandu:
- Add a default Rubocop linter config file people can use.
- This won't change ANYTHING, since we're not running any automated scripts based on this at the moment.
 
Obviously, this is open for additions. I'm not tied to these settings, but I just wanted to start somewhere.

To get this set up on your system do this:
```
crew install libyaml
gem install rubocop
```
Once this file is at ~/.rubocop.yml, you can run ```rubocop package.rb``` to get some suggested changes.

Run ```rubocop -A package.rb``` to try to make those fixes automatically.

Please feel free to add to this PR if there are changes you think should be made.

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

### DIFF
--- a/packages/ruby_rubocop.rb
+++ b/packages/ruby_rubocop.rb
@@ -50,7 +50,7 @@ class Ruby_rubocop < Package
   end
 
   def self.postinstall
-    @gem_name = self.name.sub('ruby_','')
+    @gem_name = name.sub('ruby_', '')
     system "gem install -N #{@gem_name} --conservative"
 
     puts "Installing Chromebrew rubocop config file at #{@xdg_config_home}/rubocop/config.yml".lightblue
@@ -63,9 +63,12 @@ class Ruby_rubocop < Package
   end
 
   def self.remove
-    @gem_name = self.name.sub('ruby_','')
+    @gem_name = name.sub('ruby_', '')
     @gems_deps = `gem dependency ^#{@gem_name}\$ | awk '{print \$1}'`.chomp
-    @gems = @gems_deps.split("\n").drop(1).append("#{@gem_name}")
+    # Delete the first line and convert to an array.
+    @gems = @gems_deps.split("\n").drop(1).append(@gem_name)
+    # bundler never gets uninstalled, though gem dependency lists it for
+    # every package, so delete it from the list.
     @gems.delete('bundler')
     @gems.each do |gem|
       system "gem uninstall -Dx --force --abort-on-dependent #{gem} || true"
--- a/packages/ruby_rubocop.rb
+++ b/packages/ruby_rubocop.rb
@@ -1,0 +1,49 @@
+# Adapted from Arch Linux ruby-rubocop PKGBUILD at:
+# https://aur.archlinux.org/cgit/aur.git/plain/PKGBUILD?h=ruby-rubocop
+
+require 'package'
+
+class Ruby_rubocop < Package
+  description 'A Ruby static code analyzer and formatter'
+  homepage 'https://rubocop.org'
+  version '1.22.0'
+  compatibility 'all'
+  source_url 'SKIP'
+
+  depends_on 'libyaml'
+
+  @xdg_config_home = ENV['XDG_CONFIG_HOME']
+  @xdg_config_home = "#{CREW_PREFIX}/.config" if @xdg_config_home.to_s.empty?
+
+  def self.install
+    ENV['CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY'] = '1'
+    warn_level = $VERBOSE
+    $VERBOSE = nil
+    load "#{CREW_LIB_PATH}lib/const.rb"
+    $VERBOSE = warn_level
+    FileUtils.mkdir_p "#{CREW_DEST_DIR}#{@xdg_config_home}/rubocop"
+  end
+
+  def self.postinstall
+    system 'gem install -N rubocop --conservative'
+    puts "Installing Chromebrew rubocop config file at #{@xdg_config_home}/rubocop/config.yml".lightblue
+    puts "This can be overridden by a ~/.rubocop.yml".lightblue
+    system "curl -Ls https://github.com/chromebrew/chromebrew/raw/master/.rubocop.yml > #{@xdg_config_home}/rubocop/config.yml"
+  end
+
+  def self.remove
+    config_dirs = %W[#{@xdg_config_home}/rubocop]
+    config_dirs.each do |config_dir|
+      next unless Dir.exist? config_dir
+
+      print "\nWould you like to remove #{config_dir}? [y/N] "
+      case $stdin.getc
+      when 'y', 'Y'
+        FileUtils.rm_rf config_dir
+        puts "#{config_dir} removed.".lightred
+      else
+        puts "#{config_dir} saved.".lightgreen
+      end
+    end
+  end
+end
--- a/packages/ruby_rubocop.rb
+++ b/packages/ruby_rubocop.rb
@@ -6,33 +6,59 @@ require 'package'
 class Ruby_rubocop < Package
   description 'A Ruby static code analyzer and formatter'
   homepage 'https://rubocop.org'
-  version '1.22.0'
+  version '1.26.1'
   compatibility 'all'
-  source_url 'SKIP'
+  source_url 'https://github.com/rubocop/rubocop.git'
+  git_hashtag "v#{version}"
+
+
+  binary_url({
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby_rubocop/1.26.1_armv7l/ruby_rubocop-1.26.1-chromeos-armv7l.tar.zst',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby_rubocop/1.26.1_armv7l/ruby_rubocop-1.26.1-chromeos-armv7l.tar.zst',
+       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby_rubocop/1.26.1_i686/ruby_rubocop-1.26.1-chromeos-i686.tar.zst',
+     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby_rubocop/1.26.1_x86_64/ruby_rubocop-1.26.1-chromeos-x86_64.tar.zst'
+  })
+  binary_sha256({
+    aarch64: 'c81b0e5bee7f31d632c5e5a6159331eef36a839aa56d3cd98217161ea1ba1b6e',
+     armv7l: 'c81b0e5bee7f31d632c5e5a6159331eef36a839aa56d3cd98217161ea1ba1b6e',
+       i686: '61e75cfeaa580f4f78e94dfbee12186331b412721244cfc0d7660bd9338040d9',
+     x86_64: '66c9134d10ee390f53449edf863cdc1a58be3af8ed8fda6571aa4228244d45a3'
+  })
+
+  no_fhs
 
   depends_on 'libyaml'
+  depends_on 'ruby'
   depends_on 'xdg_base'
 
   @xdg_config_home = ENV['XDG_CONFIG_HOME']
   @xdg_config_home = "#{CREW_PREFIX}/.config" if @xdg_config_home.to_s.empty?
 
+  def self.build
+    system 'bundle install'
+    system 'rake build'
+  end
+
   def self.install
-    ENV['CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY'] = '1'
-    warn_level = $VERBOSE
-    $VERBOSE = nil
-    load "#{CREW_LIB_PATH}lib/const.rb"
-    $VERBOSE = warn_level
-    FileUtils.mkdir_p "#{CREW_DEST_DIR}#{@xdg_config_home}/rubocop"
+    system "gem install --build=#{CREW_DEST_DIR} pkg/rubocop-#{version}.gem"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/.config/rubocup"
+    system "curl -Ls https://github.com/satmandu/chromebrew/raw/rubocop_yaml/.rubocop.yml > #{CREW_DEST_PREFIX}/.config/rubocup/config.yml"
+    # Uncomment after merge
+    # system "curl -Ls https://github.com/chromebrew/chromebrew/raw/master/.rubocop.yml > #{CREW_DEST_PREFIX}/.config/rubocup/config.yml"
   end
 
   def self.postinstall
-    system 'gem install -N rubocop --conservative'
+    # system 'gem install -N rubocop --conservative'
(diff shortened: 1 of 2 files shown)


uberhacker:
Based on
```
RuboCop will start looking for the configuration file in the directory
where the inspected file is and continue its way up to the root directory.
```
This tells me you can have multiple config files.  The question is what is the order of precedence?  If you "installed" a file in `/usr/local/etc/.rubocop.yml`, for instance, would your local `~/.rubocop.yml` override (aka take precedence over) the installed config file?  Even if that doesn't work and you created a wrapper package to install the gem with postinstall instructions, that would be better than "hiding" the file in the project root.  Also, how do you know the project root file doesn't take precedence over your `~/.rubocop.yml`?

satmandu:
The idea is to only add the rubocop config to the root of the chromebrew repo so people know about it.

I'd hate to overwrite a local config file people are using. This is only being put on the site for the sake of communication with users of the "hey, this is what we use as a baseline for our package files."

uberhacker:
I get that.  I'm not suggesting we install to `~/.rubocop.yml`.  What I'm asking is can we install the "default" `.rubocop.yml` somewhere and still have `~/.rubocop.yml` take precedence?  Even if you put it in the project root after you clone the repo locally, there is nothing preventing the project root config from taking precedence over your `~/.rubocop.yml` depending on where you execute the analysis.

satmandu:
I see what you're saying.

As per https://docs.rubocop.org/rubocop/1.22/configuration.html

```
Config file locations
RuboCop will start looking for the configuration file in the directory where the inspected file is and continue its way up to the root directory.

If it cannot be found until reaching the project’s root directory, then it will be searched for in the user’s global config locations, which consists of a dotfile or a config file inside the XDG Base Directory specification.

~/.rubocop.yml

$XDG_CONFIG_HOME/rubocop/config.yml (expands to ~/.config/rubocop/config.yml if $XDG_CONFIG_HOME is not set)

If both files exist, the dotfile will be selected.

As an example, if RuboCop is invoked from inside /path/to/project/lib/utils, then RuboCop will use the config as specified inside the first of the following files:

/path/to/project/lib/utils/.rubocop.yml

/path/to/project/lib/.rubocop.yml

/path/to/project/.rubocop.yml

/.rubocop.yml

~/.rubocop.yml

~/.config/rubocop/config.yml
```

Would dropping a copy in `$XDG_CONFIG_HOME/rubocop/config.yml` be useful, since that maps to `/usr/local/.config/rubocop/config.yml`. It looks like that would not supersede a `~/.rubocop.yml` if such a file exists.

uberhacker:
Yes, that sounds reasonable.  If this was packaged, you could remove it for a more elegant way to control the config.

satmandu:
@uberhacker package added, pointing to new repo...

packages/ruby_rubocop.rb
```diff
@@ -0,0 +1,49 @@
...
+require 'package'
+
+class Ruby_rubocop < Package
+  description 'A Ruby static code analyzer and formatter'
+  homepage 'https://rubocop.org'
+  version '1.22.0'
+  compatibility 'all'
+  source_url 'SKIP'
+
+  depends_on 'libyaml'
```
uberhacker, line 13: Add `depends_on 'xdg_base'`?  This will create the `~/.config` directory.

packages/ruby_rubocop.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  @xdg_config_home = ENV['XDG_CONFIG_HOME']
+  @xdg_config_home = "#{CREW_PREFIX}/.config" if @xdg_config_home.to_s.empty?
+
+  def self.install
+    ENV['CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY'] = '1'
+    warn_level = $VERBOSE
+    $VERBOSE = nil
+    load "#{CREW_LIB_PATH}lib/const.rb"
+    $VERBOSE = warn_level
```
uberhacker, line 23: Why do you need the previous 5 lines?

packages/ruby_rubocop.rb
```diff
@@ -0,0 +1,49 @@
...
+    FileUtils.mkdir_p "#{CREW_DEST_DIR}#{@xdg_config_home}/rubocop"
```
uberhacker, line 24: What is installed?

packages/ruby_rubocop.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def self.postinstall
+    system 'gem install -N rubocop --conservative'
```
uberhacker, line 28: Shouldn't this go in the `self.install` section so files can be tracked for removal?  Otherwise, removing the package leaves the gem installed.  Not ideal.

packages/ruby_rubocop.rb
```diff
@@ -0,0 +1,49 @@
...
+    puts "Installing Chromebrew rubocop config file at #{@xdg_config_home}/rubocop/config.yml".lightblue
+    puts "This can be overridden by a ~/.rubocop.yml".lightblue
+    system "curl -Ls https://github.com/chromebrew/chromebrew/raw/master/.rubocop.yml > #{@xdg_config_home}/rubocop/config.yml"
```
uberhacker, line 31: Shouldn't this be installed in `#{CREW_DEST_PREFIX}/.config/rubocup` so it can be tracked for removal?

packages/ruby_rubocop.rb
```diff
@@ -0,0 +1,49 @@
...
+      print "\nWould you like to remove #{config_dir}? [y/N] "
+      case $stdin.getc
+      when 'y', 'Y'
+        FileUtils.rm_rf config_dir
+        puts "#{config_dir} removed.".lightred
+      else
+        puts "#{config_dir} saved.".lightgreen
+      end
+    end
+  end
```
uberhacker, line 48: Would you even need this section if the config was installed in `#{CREW_DEST_PREFIX}/.config/rubocup`?

packages/ruby_rubocop.rb
satmandu: It complains.

packages/ruby_rubocop.rb
satmandu: I added self.remove for this.  I don't want to interfere with the gem installed database.

packages/ruby_rubocop.rb
satmandu: isn't that what  `#{@xdg_config_home}/rubocop/` is?

packages/ruby_rubocop.rb
satmandu: removed it.

uberhacker: Should this also remove the config at `#{@xdg_config_home}/rubocop/config.yml`?  If this file was installed in `#{CREW_DEST_PREFIX}/.config/rubocop/config.yml`, it could be included in the package metadata file.

packages/ruby_rubocop.rb
uberhacker: `#{@xdg_config_home}` is using `CREW_PREFIX`, not `CREW_DEST_PREFIX` so it won't be tracked in the package.

uberhacker: Also, shouldn't we be using a hash or tag instead of master?

satmandu:
@uberhacker How does this look?

packages/ruby_rubocop.rb
```diff
@@ -0,0 +1,64 @@
...
+  @xdg_config_home = "#{CREW_PREFIX}/.config" if @xdg_config_home.to_s.empty?
+
+  def self.build
+    system 'bundle install'
+    system 'rake build'
+  end
+
+  def self.install
+    system "gem install --build=#{CREW_DEST_DIR} pkg/rubocop-#{version}.gem"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/.config/rubocup"
```
uberhacker, line 44: rubocup?

packages/ruby_rubocop.rb
```diff
@@ -0,0 +1,64 @@
...
+    system "curl -Ls https://github.com/satmandu/chromebrew/raw/rubocop_yaml/.rubocop.yml > #{CREW_DEST_PREFIX}/.config/rubocup/config.yml"
+    # Uncomment after merge
+    # system "curl -Ls https://github.com/chromebrew/chromebrew/raw/master/.rubocop.yml > #{CREW_DEST_PREFIX}/.config/rubocup/config.yml"
```
uberhacker, line 47: Should this be versioned?

satmandu: I don't think it should be. We should be pulling the most recent version of the linter config from our repo, in the same way that `install.sh` isn't versioned.

satmandu:
```
chronos@cheekon-x86_64 /usr/local/lib/crew/packages $ crew install ruby_rubocop
ruby_rubocop: A Ruby static code analyzer and formatter
https://rubocop.org
Version: 1.26.1

xdg_base
Do you agree? [Y/n] y
Proceeding...
xdg_base: XDG Base Directory Specification Configuration
[###############################################################]  524.00 B 100%
Xdg_base archive downloaded.
Performing pre-install...
Performing install...
Using rdfind to convert duplicate files to hard links.
Now scanning ".", found 1 files.
Now have 1 files in total.
Removed 0 files due to nonunique device and inode.
Total size is 422 bytes or 422 B
Removed 1 files due to unique sizes from list. 0 files left.
Now eliminating candidates based on first bytes: removed 0 files from list. 0 files left.
Now eliminating candidates based on last bytes: removed 0 files from list. 0 files left.
Now eliminating candidates based on sha1 checksum: removed 0 files from list. 0 files left.
It seems like you have 0 files that are not unique
Totally, 0 B can be reduced.
Now making hard links.
Making 0 links.
[###############################################################]  753.00 B 100%
Ruby_rubocop archive downloaded.
Performing pre-install...
Performing install...
Using rdfind to convert duplicate files to hard links.
Now scanning ".", found 1 files.
Now have 1 files in total.
Removed 0 files due to nonunique device and inode.
Total size is 1002 bytes or 1002 B
Removed 1 files due to unique sizes from list. 0 files left.
Now eliminating candidates based on first bytes: removed 0 files from list. 0 files left.
Now eliminating candidates based on last bytes: removed 0 files from list. 0 files left.
Now eliminating candidates based on sha1 checksum: removed 0 files from list. 0 files left.
It seems like you have 0 files that are not unique
Totally, 0 B can be reduced.
Now making hard links.
Making 0 links.
Performing post-install for xdg_base...
Performing post-install for ruby_rubocop...
Fetching rubocop-1.26.1.gem
Successfully installed rubocop-1.26.1
1 gem installed
Installing Chromebrew rubocop config file at /usr/local/.config/rubocop/config.yml
This can be overridden by a ~/.rubocop.yml
Ruby_rubocop installed!
```
```
chronos@cheekon-x86_64 /usr/local/lib/crew/packages $ crew remove ruby_rubocop
Successfully uninstalled parallel-1.22.1
Removing ruby-parse
Removing ruby-rewrite
Successfully uninstalled parser-3.1.1.0
Successfully uninstalled rainbow-3.1.1
Successfully uninstalled regexp_parser-2.2.1
Successfully uninstalled rexml-3.2.5
Successfully uninstalled rubocop-ast-1.16.0
Successfully uninstalled ruby-progressbar-1.11.0
Successfully uninstalled unicode-display_width-2.1.0
Removing rubocop
Successfully uninstalled rubocop-1.26.1
Ruby_rubocop removed!
chronos@cheekon-x86_64 /usr/local/lib/crew/packages $ 
```

packages/ruby_rubocop.rb
satmandu: This is fixed.

packages/ruby_rubocop.rb
```diff
@@ -0,0 +1,79 @@
...
+  def self.remove
+    @gem_name = name.sub('ruby_', '')
+    @gems_deps = `gem dependency ^#{@gem_name}\$ | awk '{print \$1}'`.chomp
+    # Delete the first line and convert to an array.
+    @gems = @gems_deps.split("\n").drop(1).append(@gem_name)
+    # bundler never gets uninstalled, though gem dependency lists it for
+    # every package, so delete it from the list.
+    @gems.delete('bundler')
+    @gems.each do |gem|
+      system "gem uninstall -Dx --force --abort-on-dependent #{gem} || true"
```
uberhacker, line 74: That `--abort-on-dependent` option is nice.

packages/ruby_rubocop.rb
```diff
@@ -0,0 +1,79 @@
...
+    # system 'bundle install'
+    # system 'rake build'
+  end
+
+  def self.install
+    # system "gem install --build=#{CREW_DEST_DIR} pkg/rubocop-#{version}.gem"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/.config/rubocop"
+
+    # Remove following line after merge of this PR.
+    system "curl -Lsf https://github.com/satmandu/chromebrew/raw/rubocop_yaml/.rubocop.yml -o #{CREW_DEST_PREFIX}/.config/rubocop/config.yml"
```
uberhacker, line 47: `downloader 'https://github.com/satmandu/chromebrew/raw/rubocop_yaml/.rubocop.yml', <hash>, "#{CREW_DEST_PREFIX}/.config/rubocop/config.yml"`?

packages/ruby_rubocop.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  def self.postinstall
+    @gem_name = name.sub('ruby_', '')
+    system "gem install -N #{@gem_name} --conservative"
+
+    puts "Installing Chromebrew rubocop config file at #{@xdg_config_home}/rubocop/config.yml".lightblue
+    puts 'This can be overridden by a ~/.rubocop.yml'.lightblue
+    FileUtils.mkdir_p "#{@xdg_config_home}/rubocop"
+    # Remove following line after merge of this PR.
+    system "curl -Lsf https://github.com/satmandu/chromebrew/raw/rubocop_yaml/.rubocop.yml -o #{@xdg_config_home}/rubocop/config.yml"
```
uberhacker, line 60: `downloader 'https://github.com/satmandu/chromebrew/raw/rubocop_yaml/.rubocop.yml', <hash>, "#{@xdg_config_home}/rubocop/config.yml"`?

satmandu: The point is not to use a hash since we are trusting the Chromebrew repo. If the repo is compromised, then the hash for `rubocop.yml` linter config file in the `ruby_rubofop.rb` can get compromised at the same time as the linter config file. So there really isn't any point to doing a further hash check here.

satmandu: Having said that, @supechicken can downloader handle not having a hash provided to it?

packages/ruby_rubocop.rb
satmandu: Same argument as in https://github.com/skycocker/chromebrew/pull/6266#discussion_r840182516

packages/ruby_rubocop.rb
uberhacker: I believe you can use the keyword `'SKIP'`.

satmandu: Changes made!

PR #6271, opened 2021-10-02
crew: check missing libraries while installation

supechicken:
Requires #6275 

Although it can't fix the `missing shared object` errors, it can let us know and fix it more quickly

Tested on `x86_64`


satmandu:
So this is a compendium of several changes, some of which look great, and some of which I have questions about.

Questions:
Why change using upx to not default?  (This saves a ton of room.)

Why not use symlinks (honestly open to taking that out if not useful.)

Also some of the changes here could be classified as cleanups. Perhaps put those in a separate PR, so we can get those merged more quickly?

bin/crew
```diff
@@ -5,12 +5,10 @@ require_relative '../lib/color'
 abort 'Chromebrew should not be run as root.'.lightred if Process.uid == 0
 
 require 'find'
-require 'net/http'
 require 'uri'
 require 'digest/sha2'
 require 'json'
 require 'fileutils'
-require 'securerandom'
```
uberhacker, line 13: Are you sure these modules are not needed?

lib/const.rb
```diff
@@ -51,8 +51,8 @@
 CREW_CACHE_ENABLED = ENV['CREW_CACHE_ENABLED']
 
 CREW_CONFLICTS_ONLY_ADVISORY = ENV['CREW_CONFLICTS_ONLY_ADVISORY']
-
-CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY = ENV['CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY']
+CREW_FHS_NONCOMPLIANCE_ERROR = ENV['CREW_FHS_NONCOMPLIANCE_ERROR']
```
uberhacker, line 54: If you are going to replace `CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY`, you will also need to change all the packages that reference this constant.  Suggestion: Don't change. :)

lib/const.rb
```diff
@@ -62,12 +62,13 @@
 end
 
 # Set CREW_NOT_COMPRESS from environment variable
-CREW_NOT_COMPRESS = ENV['CREW_NOT_COMPRESS']
+CREW_NOT_COMPRESS = ENV['CREW_COMPRESS']
```
uberhacker, line 65: Why was this changed?  Don't we want the environment variable to match the constant?

lib/const.rb
```diff
@@ -62,12 +62,13 @@
...
 
 # Set CREW_NOT_STRIP from environment variable
 CREW_NOT_STRIP = ENV['CREW_NOT_STRIP']
 
 CREW_SHRINK_ARCHIVE = ENV['CREW_SHRINK_ARCHIVE']
+CREW_USE_UPX = ENV['CREW_USE_UPX']
```
uberhacker, line 71: Good call.  I could have sworn we already made this a constant.

lib/const.rb
```diff
@@ -83,7 +84,7 @@
 
 USER = `whoami`.chomp
 
-CHROMEOS_RELEASE = `grep CHROMEOS_RELEASE_CHROME_MILESTONE= /etc/lsb-release | cut -d'=' -f2`.chomp
+CHROMEOS_RELEASE = File.read('/etc/lsb-release')[/CHROMEOS_RELEASE_CHROME_MILESTONE=(.+)/, 1]
```
uberhacker, line 87: Rubyize for the win!

bin/crew
```diff
@@ -1002,16 +1013,16 @@ def shrink_dir(dir)
...
         system "#{CREW_PREFIX}/bin/rdfind -removeidentinode true -makesymlinks true -makeresultsfile false ."
       end
-      if File.exist?("#{CREW_PREFIX}/bin/symlinks")
-        puts "Using symlinks tool to make absolute symlinks relative"
-        system 'symlinks -cr .' if File.exist?("#{CREW_PREFIX}/bin/symlinks")
-      end
+      #if File.exist?("#{CREW_PREFIX}/bin/symlinks")
+      #  puts "Using symlinks tool to make absolute symlinks relative"
+      #  system 'symlinks -cr .' if File.exist?("#{CREW_PREFIX}/bin/symlinks")
+      #end
```
uberhacker, line 1019: I kinda agree symlinks creates more problems than it solves.

bin/crew
```diff
@@ -907,19 +905,32 @@ def prepare_package(destdir)
     system 'find . -type l >> ../filelist'
     system 'cut -c2- ../filelist > filelist'
 
+    # check existence of required libraries
+    missing_lib = `find . -executable -type f -exec env LD_TRACE_LOADED_OBJECTS=1 {} \\;`.scan(/\t(.+) => not found/).flatten
```
uberhacker, line 909: Can this be Rubyized further?

lib/const.rb
```diff
@@ -84,7 +84,7 @@
 
 USER = `whoami`.chomp
 
-CHROMEOS_RELEASE = File.read('/etc/lsb-release')[/CHROMEOS_RELEASE_CHROME_MILESTONE=(.+)/, 1]
+CHROMEOS_RELEASE = ( File.exist?('/etc/lsb-release') )? File.read('/etc/lsb-release')[/CHROMEOS_RELEASE_CHROME_MILESTONE=(.+)/, 1]
```
uberhacker, line 87: If you are going to use an immediate if, it would be good to add `: ''` at the end.

lib/const.rb
supechicken: Corrected back to `ENV['CREW_NOT_COMPRESS']`

lib/const.rb
supechicken: Changed back

bin/crew
supechicken: Yes, @saltedcoffii told me that in the previous pr

supechicken:
> Why change using upx to not default? (This saves a ton of room.)
Because it is very slow with larger packages even with threading (like `android-studio`, `opera`)

> Why not use symlinks (honestly open to taking that out if not useful.)
As it messes up all absolute symlinks at this time (we need to fix it before enabling in `crew` again)

> Also some of the changes here could be classified as cleanups. Perhaps put those in a separate PR, so we can get those merged more quickly?
Okay :)

bin/crew

bin/crew
saltedcoffii: Yes, securerandom is not needed.

saltedcoffii:
@supechicken what's the progress on this?

supechicken:
> @supechicken what's the progress on this?

This branch was deleted by me accidentally, I will reopen this pr with the latest changes later

PR #6274, opened 2021-10-04
Fix conflict between findutils and mlocate

saltedcoffii:
This pull request fixes multiple things between `findutils` and `mlocate`.

- `locate` and `updatedb` both exist in packages `findutils` and `mlocate`. `mlocate` is a more secure implementation, so the `findutils` tools must be accessed at `locate.findutils` and `updatedb.findutils`. This is in accordance with Debian and Ubuntu systems, however Debian and Ubuntu use `plocate` instead of `mlocate`. This change may come later, as `plocate` is faster.
- to keep ownership consistent throughout `/usr/local` in accordance with (currently unwritten) Chromebrew standards, `updatedb` cannot be run as root
- The old location of the `mlocate.db` file, found in `/usr/local/db/mlocate/mlocate.db` is not FHS compliant. It can now be found at `/usr/local/var/mlocate/mlocate.db`. If requested, I'm happy to move to `/usr/local/var/db/mlocate/mlocate.db`
- Update environment and `CFLAGS` for `mlocate`


packages/mlocate.rb
```diff
@@ -2,35 +2,39 @@
...
-    aarch64: '69e02f44a498cd38a8f4bc2e3410f934506426d670664c402a09bfb34d51418c',
-     armv7l: '69e02f44a498cd38a8f4bc2e3410f934506426d670664c402a09bfb34d51418c',
-       i686: '74acf4453947b3570f68c43e485cb8e9f2da688e4a50336d593bbba1b6ebd2ee',
-     x86_64: 'f943fe00ad4666f5a9c7bf12d710cd8f2e136a33ad94568863b92dd45526d7eb',
-  })
+  def self.patch
+      @updatedb_wrapper = <<~EOF
+      #!/bin/bash -e
+      if [ "${EUID}" = 0 ]; then
+        echo "updatedb should not be run as root on ChromeOS."
```
uberhacker, line 16: Why not?

saltedcoffii: "to keep ownership consistent throughout `/usr/local` in accordance with (currently unwritten) Chromebrew standards, `updatedb` cannot be run as root"

packages/mlocate.rb
```diff
@@ -2,35 +2,39 @@
...
-    system "./configure"
-    system "make"
+    system "sed -i 's:groupname = mlocate:groupname = #{USER}:g' Makefile.*"
+    system "#{CREW_ENV_OPTIONS} ./configure #{CREW_OPTIONS}"
+    system 'make'
   end
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "install"
+    FileUtils.mv "#{CREW_DEST_PREFIX}/bin/updatedb", "#{CREW_DEST_PREFIX}/bin/updatedb.real"
```
uberhacker, line 33: I don't think this is a good idea.  We should probably just remove the binaries from findutils or remove the package altogether.

packages/mlocate.rb
saltedcoffii: Minus that consistency, should someone run `updatedb` as root, and then attempt to run it as a normal user, they will receive an error. Also, running as a normal user is a very slight security improvement, as it doesn't save `/proc` or `/sys` in the database.

packages/mlocate.rb
saltedcoffii: That is how Debian does it. It would be nice to have updated versions of `find` and `xargs`, so let's not remove the `findutils` package. Also, when I finally get to working on the `crew-alternatives` package, our own little implementation of `dpkg`'s `update-alternatives`, we can offer `findutils`'s implementation as an option.

packages/mlocate.rb
uberhacker: How about instead of restricting it altogether, reset the ownership back afterwards?  I think this restriction is too limiting.

saltedcoffii: We could, but that would require the use of `sudo`, as `chown` seems to require `sudo` all the time. So running `updatedb` would then require a password.

saltedcoffii: @satmandu, @supechicken second opinions?

uberhacker: I don't think sudo should be a requirement.  You should be able to run `updatedb` or `sudo updatedb` separately.

packages/mlocate.rb
```diff
@@ -2,35 +2,39 @@
...
-  binary_sha256 ({
-    aarch64: '69e02f44a498cd38a8f4bc2e3410f934506426d670664c402a09bfb34d51418c',
-     armv7l: '69e02f44a498cd38a8f4bc2e3410f934506426d670664c402a09bfb34d51418c',
-       i686: '74acf4453947b3570f68c43e485cb8e9f2da688e4a50336d593bbba1b6ebd2ee',
-     x86_64: 'f943fe00ad4666f5a9c7bf12d710cd8f2e136a33ad94568863b92dd45526d7eb',
-  })
+  def self.patch
+      @updatedb_wrapper = <<~EOF
+      #!/bin/bash -e
+      if [ "${EUID}" = 0 ]; then
```
supechicken, line 15: Instead of a wrapper script, I would prefer to patching the source by digging into it (e.g add a header file for rejecting root)

packages/mlocate.rb
supechicken: I think our target is making them exist at the system if possible but not keeping once only, a better option is to isolate them from /use/local filesystems and make symlinks to /usr/local like homebrew does

uberhacker: There are a lot of things I don't like about Homebrew including excessive symlinks and the overuse of non-standard PATH locations.  This trap is something I want to get away from in Chromebrew.

saltedcoffii: @supechicken, I have a solution that will satisfy both you and @uberhacker, but that's out of the scope of this PR.

PR #6275, opened 2021-10-05
crew: Clean up

supechicken:
Simply codes :)

Tested on `x86_64`

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -2,7 +2,7 @@
 
 CREW_VERSION = '1.16.10'
 
-ARCH_ACTUAL = `uname -m`.strip
+ARCH_ACTUAL = `uname -m`.chomp
 # This helps with virtualized builds on aarch64 machines
 # which report armv8l when linux32 is run.
 ARCH = ( ARCH_ACTUAL == 'armv8l' ) ? 'armv7l' : ARCH_ACTUAL
@@ -55,7 +55,7 @@ CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY = ENV['CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY
 
 # Set CREW_NPROC from environment variable or `nproc`
 if ENV['CREW_NPROC'].to_s.empty?
-  CREW_NPROC = `nproc`.strip
+  CREW_NPROC = `nproc`.chomp
 else
   CREW_NPROC = ENV['CREW_NPROC']
 end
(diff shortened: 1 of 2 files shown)


uberhacker:
Bump crew version?

lib/const.rb
```diff
@@ -1,21 +1,20 @@
...
-ARCH_LIB = if ARCH == 'x86_64' and Dir.exist?('/lib64') then 'lib64' else 'lib' end
+ARCH_LIB = ( ARCH == 'x86_64' and Dir.exist?('/lib64') ) ? 'lib64' : 'lib'
 
 # Glibc version can be found from the output of libc.so.6
-@libcvertokens=  %x[/#{ARCH_LIB}/libc.so.6].lines.first.chomp.split(/[\s]/)
-LIBC_VERSION = @libcvertokens[@libcvertokens.find_index("version") + 1].sub!(/[[:punct:]]?$/,'')
+LIBC_VERSION = %x[/#{ARCH_LIB}/libc.so.6][/version ([\d.]+)/, 1]
 
-if ENV['CREW_PREFIX'] and ENV['CREW_PREFIX'] != '' and ENV['CREW_PREFIX'] != '/usr/local'
+unless ENV['CREW_PREFIX'].to_s.empty? or ENV['CREW_PREFIX'] == '/usr/local'
```
uberhacker, line 17: Does this throw an error if the environment variable `CREW_PREFIX` is not set?

lib/const.rb
```diff
@@ -70,20 +69,25 @@
...
-CREW_TESTING_BRANCH = crew_testing_branch
-CREW_TESTING_REPO = crew_testing_repo
+CREW_TESTING_BRANCH = ENV['CREW_TESTING_BRANCH']
+CREW_TESTING_REPO = ENV['CREW_TESTING_REPO']
+
+CREW_TESTING = unless CREW_TESTING_BRANCH.to_s.empty? or CREW_TESTING_REPO.to_s.empty?
+  ENV['CREW_TESTING']
+else
+  '0'
+end
```
uberhacker, line 79: Lines 75-79 could be turned into an immediate if one-liner.

lib/const.rb
```diff
@@ -70,20 +69,25 @@
...
 
 USER = `whoami`.chomp
 
-CHROMEOS_RELEASE = `grep CHROMEOS_RELEASE_CHROME_MILESTONE= /etc/lsb-release | cut -d'=' -f2`.chomp
+CHROMEOS_RELEASE = if File.exist?('/etc/lsb-release')
+  File.read('/etc/lsb-release')[/CHROMEOS_RELEASE_CHROME_MILESTONE=(.+)/, 1]
+else
+  # newer version of Chrome OS exports info to env by default
+  ENV['CHROMEOS_RELEASE_CHROME_MILESTONE']
+end
```
uberhacker, line 90: Turn into an immediate if one-liner?

lib/const.rb
```diff
@@ -105,18 +109,65 @@
...
+  -Dprefix=#{CREW_PREFIX} \
+  -Dlibdir=#{CREW_LIB_PREFIX} \
+  -Dmandir=#{CREW_MAN_PREFIX} \
+  -Dbuildtype=release \
+  -Db_lto=true \
+  -Dstrip=true \
+  -Db_pie=true \
+  -Dcpp_args='-O2' \
+  -Dc_args='-O2'
+OPT
```
uberhacker, line 140: I like this.  Much cleaner and easier to add or remove options.

lib/const.rb
supechicken: The `to_s` function handles it if it is not set (from nil to an empty string if not set)

lib/const.rb

lib/const.rb
supechicken: It would be too long if making this in a line

uberhacker: How about something like below?
```
CHROMEOS_RELEASE = ( File.exist?('/etc/lsb-release') ) ? \
  File.read('/etc/lsb-release')[/CHROMEOS_RELEASE_CHROME_MILESTONE=(.+)/, 1] : \
  # newer version of Chrome OS exports info to env by default
  ENV['CHROMEOS_RELEASE_CHROME_MILESTONE']
```
Continued one-liner. :)

supechicken: I think if/unless is better then escaping :)

PR #6291, opened 2021-10-11
Disable symlinks usage in shrink_archive function

satmandu:
- removes symlinks from crew

Works properly:
- [x] x86_64
- [x] armv7l

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.16.9'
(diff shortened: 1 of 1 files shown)


uberhacker:
This is a welcome change.  Challenge for the future: I think the issue with symlinks is it is run in `CREW_DEST_PREFIX` and doesn't account for the relocated install `CREW_PREFIX` base.  If it can simulate or rewrite to the correct location, that may be all that is needed to fix this.  It might be as simple as `sed 's,../../../../,,' <target>`.

PR #6301, opened 2021-10-14
Remove newlines from some lib/const.rb variables

saltedcoffii:
This fixes a lot of building errors I was running into surrounding `./configure`.


saltedcoffii:
Alright, this works now!

PR #6306, opened 2021-10-17
Package functions: a framework for making package building a whole lot simpler

saltedcoffii:
We repeat a LOT of code in our package files. Many are identical past their headers. This aims to add a simple way to simplify and DRY a lot of our code in package files. This should also help reduce lots of typing, especially for Python packages. It also builds in an extensible way to configure how the framework interacts with the package file through called functions. Naturally this won't work with every package.

I know this is very novel, so let me know what you think! Ideas and improvements very welcome.

```
CREW_TESTING=1 CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew CREW_TESTING_BRANCH=package-functions crew update
```


saltedcoffii:
Seems like we all like the idea. I'll revert the examples to their original forms since I would like to get this PR through tonight so I can incorporate it into tomorrow's work.

uberhacker:
I guess a good way to test this is to build some binaries.  Has this been attempted yet?

saltedcoffii:
@uberhacker @supechicken  please re-review, I would like to work on this more :)

lib/buildsystems/cmake_v1.rb
```diff
@@ -0,0 +1,36 @@
...
+    Dir.mkdir 'builddir'
+    Dir.chdir 'builddir' do
+      system "#{CREW_ENV_OPTIONS} cmake -G Ninja #{CREW_CMAKE_OPTIONS} \
+            #{@cmake_options} .."
+    end
+    system 'samu -C builddir'
+  end
+
+  def self.install
+    system "DESTDIR=#{CREW_DEST_DIR} samu -C builddir install"
```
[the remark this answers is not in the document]
satmandu: I'm perfectly fine with `samu`...

satmandu:
My brain is tired at the moment, but I'm all about using functions.

To test this, I think this should work:
```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=package-functions CREW_TESTING=1 crew update
```

Followed by an attempt at building something...

satmandu:
Looks like there are still some problems?

image

lib/buildsystems/meson/v1.rb
```diff
@@ -0,0 +1,33 @@
...
+  @_buildsystems_meson_version = '1.0.0'
+
+  def self.meson_options (options = '')
+    if options
+      @meson_options = options
+    end
+    return @meson_options
+  end
+
+  def self.check? (bool = false)
```
supechicken, line 13: Change it to without argument required?
```ruby
def self.run_check
  @check = true
end
```
```ruby
# in packages
use_check
```

lib/buildsystems/setup_py/v1.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def self.python_version (pyversion = '3')
+    @python = 'python' + pyversion
+    return @python
+  end
+
+  def self.build
+    if @python == nil
+      @python = 'python3'
```
supechicken, line 18: How about:
```ruby
# set @python if not set
@python ||= 'python3'
```

lib/buildsystems/setup_py/v1.rb
```diff
@@ -0,0 +1,31 @@
+require 'package'
+
+class Setup_py_v1 < Package
+  @_buildsystems_setup_py_version = '1.0.0'
+
+  def self.svem? (bool = true)
```
supechicken, line 6: Change it to without argument required?
```ruby
self.use_svem
```

Zopolis4:
Testing this on armv7l, py3_magic installs fine, but the build has a different sha256 to the regular build, but it looks like they should be identical.

lib/buildsystems/autotools/v1.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def self.check? (bool = false)
+    @check = bool
+    return @check
+  end
+
+  def self.build
+    # Run autoreconf if necessary
+    unless File.executable? './configure'
```
Zopolis4, line 20: Would it not make more sense to regenerate no matter what? That would, among other things, remove the need for `filefix` (after we update to libtool 2.4.7, that is).

lib/buildsystems/autotools/v1.rb
```diff
@@ -0,0 +1,43 @@
...
+      if File.executable? './autogen.sh'
+        system 'NOCONFIGURE=1 ./autogen.sh'
```
Zopolis4, line 22: `NOCONFIGURE` is an environment variable mostly accepted by freedesktop projects, it makes sense to have it just in case but `--no-configure` is more commonly used.

lib/buildsystems/autotools/v1.rb
```diff
@@ -0,0 +1,43 @@
...
+      elsif File.executable? './bootstrap'
+        system 'NOCONFIGURE=1 ./bootstrap'
```
Zopolis4, line 24: Same here.

Zopolis4 (changes requested):

lib/buildsystems/autotools/v1.rb
supechicken: `--no-configure` is a parameter and should be placed after `./autogen.sh`

lib/buildsystems/autotools/v1.rb
supechicken: `--no-configure` is a parameter and should be placed after `./autogen.sh`

Zopolis4: Good point, I'm embarrassed I missed that. I've updated both of the suggestions.

uberhacker:
@saltedcoffii: Merge conflicts.

saltedcoffii:
> @saltedcoffii: Merge conflicts.

@uberhacker is this package ready for merge? if it is yes I'll definitely take care of merge conflicts

uberhacker:
It's fine to fix the conflicts now.  I can add the arm binaries once resolved.

lib/buildsystems/autotools/v1.rb
uberhacker: The line could be:
```
system 'NOCONFIGURE=1 ./bootstrap'
```
or:
```
system './bootstrap --no-configure'
```
but it doesn't need to be both.

saltedcoffii: Let's use NOCONFIGURE=1, because the script may react with an error if passed an incorrect parameter, but environment variables are ignored if not used.

lib/buildsystems/autotools/v1.rb
saltedcoffii: I want to say yes, however we've had some instances where packages compile from tarballs with a ./configure script already there, however have failed when re`autoreconf`ing or `autoreconf`ing from a git repository. Maybe a variable like `autoreconf?` that normally defaults to false but can be set to true if need be, something like `svem?` as seen in `lib/buildsystems/setup_py/v1.rb`

Zopolis4: If it fails, then that's probably an upstream issue which we can report or update to fix. A variable sounds good as a fallback, but I'd prefer to keep the average package as small as possible.

lib/buildsystems/autotools/v1.rb
Zopolis4: But, from what I've checked, not all scripts honor both. So it makes more sense to put both in just in case.

Zopolis4: I think that makes sense, its not beautiful but it's in the backend so it dosent really matter.

lib/buildsystems/autotools/v1.rb
Zopolis4: Maybe something that backs up `./configure` and friends, then runs autoreconf, restoring the original backups if it fails?

Zopolis4:
Given that a new hacktoberfest has rolled around, maybe it's time to dust off this PR, resolve the outstanding conversations and give this one last look before merge?

Also instead of merge commits can you please rebase the CI dosen't run unless a branch is rebased ontop of master to the point where the CI works (i.e. the Fix CI commit). 

Also, merge commits are ugly, imo. (And fix commits, no need to bulk out the log with a bunch of typo and lint fixes)

saltedcoffii:
> Given that a new hacktoberfest has rolled around, maybe it's time to dust off this PR, resolve the outstanding conversations and give this one last look before merge?
> 
> Also instead of merge commits can you please rebase the CI dosen't run unless a branch is rebased ontop of master to the point where the CI works (i.e. the Fix CI commit).
> 
> Also, merge commits are ugly, imo. (And fix commits, no need to bulk out the log with a bunch of typo and lint fixes)

I am totally in support of that! unfortunately, i'm a full time med student so I can't help much :(
Anyway it warms my heart how much you guys work on chromebrew, and who knows maybe i'll come back to help update packages and what not on my breaks.

lib/buildsystems/autotools/v1.rb
saltedcoffii: that honestly sounds like too much effort for the job we're trying to do here: specific versions of packages are meant to be built (and have been tested with) certain versions of autotools. If we use a different version, we're introducing more instability unnecessarily into a system that, not to poop on any of the devs, myself, included, but chromebrew is quite unstable.

lib/buildsystems/autotools/v1.rb
saltedcoffii: see [e5ce5a1](https://github.com/chromebrew/chromebrew/pull/6306/commits/e5ce5a12cfa4f6a0720bf3040087a1855b00baa4)

lib/buildsystems/autotools/v1.rb
saltedcoffii: see [e5ce5a1](https://github.com/chromebrew/chromebrew/pull/6306/commits/e5ce5a12cfa4f6a0720bf3040087a1855b00baa4)

lib/buildsystems/autotools/v1.rb
saltedcoffii: see e5ce5a1

satmandu:
Superseded by merged https://github.com/chromebrew/chromebrew/pull/8401  Thanks for getting this started @saltedcoffii !

PR #6310, opened 2021-10-17
new package bsdfetch

saltedcoffii:
- [x] x86_64
- [x] i686
- [x] armv7l/aarch64


uberhacker:
Rather than replacing the package, why not update it with the bsdfetch?

saltedcoffii:
OHHH. bsdfetch and go-fetch are two completely different packages that do two completely different things.

packages/go_fetch.rb
```diff
@@ -0,0 +1,25 @@
...
+  description 'fetch makes it easy to download files, folders, or release assets from a specific commit, branch, or tag of a public or private GitHub repo.'
+  homepage 'https://github.com/gruntwork-io/fetch/'
+  @_ver = '0.4.2'
+  version @_ver
+  license 'LGPL-2.1'
+  compatibility 'all'
+  source_url 'https://github.com/gruntwork-io/fetch.git'
+  git_hashtag 'v' + @_ver
+
+  depends_on 'go'
```
uberhacker, line 13: Build only dependency?

packages/fetch.rb
```diff
@@ -1,37 +1,32 @@
...
+              -DUSE_SYSTEM_SSL=ON .."
+    end
+    system 'samu -C builddir'
   end
 
   def self.install
-    system "mkdir -p #{CREW_DEST_PREFIX}/bin"
-    system "go build -ldflags \"-X main.VERSION=v0.1.1\" -o #{CREW_DEST_PREFIX}/bin/fetch"
+    system "DESTDIR=#{CREW_DEST_DIR} samu -C builddir install"
+    if CREW_LIB_SUFFIX != ''
```
uberhacker, line 26: Change to `if CREW_LIB_SUFFIX == '64'` for clarity and in case a new suffix is introduced?

saltedcoffii: If a new suffix is introduced, this logic will handle it. Although I can see your clarity point. Ping me if you'd still like this changed to `if CREW_LIB_SUFFIX == '64'`

uberhacker: It may or may not handle it correctly, though.  I don't think it's a good assumption.

saltedcoffii: okay! see 7c541b66

packages/go_fetch.rb
saltedcoffii: Turns out it was, thank you

PR #6313, opened 2021-10-18
Switch upx to just use lzma to speed up builds

satmandu:
- upx is very slow testing all compression formats, but lzma tends to be give some of the best compression anyway, so just stick to that, which should speed up builds with large executable binaries.

Works properly:
- [x] x86_64


uberhacker:
Anything to help speed things up is welcome.

PR #6345, opened 2021-10-31
New packages: zopfli, gif2apng, apng2gif, patch libpng with APNG support

saltedcoffii:
Works on x86_64. These packages need binaries.

### DIFF
--- a/packages/apng2gif.rb
+++ b/packages/apng2gif.rb
@@ -24,7 +24,10 @@ class Apng2gif < Package
 
   def self.build
     system 'make'
-    system "help2man -s 1 -N -h '' --version-string='#{@_ver}' ./apng2gif -o apng2gif.1"
+    system "help2man -s 1 -N -h '' \
+            -n '#{self.description.downcase.delete! '.'}' \
+            --version-string='#{@_ver}' \
+            ./apng2gif -o apng2gif.1"
   end
 
   def self.install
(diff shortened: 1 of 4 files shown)


packages/apng2gif.rb
```diff
@@ -0,0 +1,36 @@
...
+    system "sed -i 's:CFLAGS     = -Wall -pedantic:CFLAGS     = -Wall -pedantic #{CREW_COMMON_FLAGS}:' Makefile"
+    system "sed -i 's:CFLAGS_OPT = -O2:CFLAGS_OPT =:' Makefile"
+    # zlib is unused, remove the header and library link
+    system "sed -i 's:LIBS       = -lstdc++ -lm -lpng -lz:LIBS       = -lstdc++ -lm -lpng:' Makefile"
+    system "sed -i 's:#include \"zlib.h\"::' Makefile"
+  end
+
+  def self.build
+    system 'make'
+    system "help2man -s 1 -N -h '' --version-string='#{@_ver}' ./apng2gif -o apng2gif.1"
```
uberhacker, line 27: This is a great consideration.  How is the rendering of the help text?

packages/libpng.rb
```diff
@@ -4,39 +4,25 @@ class Libpng < Package
   description 'libpng is the official PNG reference library.'
   homepage 'http://libpng.org/pub/png/libpng.html'
   @_ver = '1.6.37'
-  version "#{@_ver}-1"
+  version "#{@_ver}+apng"
```
uberhacker, line 7: It would be better to use a numeric value here because there might be future iterations.

packages/apng2gif.rb
saltedcoffii: It's not _great_, but it gets the point across.

packages/libpng.rb
saltedcoffii: I was thinking the next iteration could be "#{@_ver}+apng-1"

uberhacker: Normally, features should be included in the description, not the version.

satmandu: Clearly, we need a better way of describing build variants...

PR #6379, opened 2021-11-08
Crew cache git granularity

satmandu:
- Please check the boolean logic here?
- Adds commit granularity for cached git branch source_url downloads, if a commit was specified. Previously a single cachefile would be used even if a new commit on the branch was specified the same day the cache was created.

Works properly:
- [x] x86_64

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=crew_cache_git_granularity CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -700,11 +700,18 @@ def download
     if @git == true
       # Recall repository from cache if requested
       if CREW_CACHE_ENABLED
+        # No git branch specified, just a git commit or tag
         if @pkg.git_branch.nil? || @pkg.git_branch.empty?
+          abort("No Git branch, commit, or tag specified!").lightred if @pkg.git_hashtag.nil? || @pkg.git_hashtag.empty
           cachefile = CREW_CACHE_DIR + filename + @pkg.git_hashtag + '.tar.xz'
           puts "cachefile is #{cachefile}".orange if @opt_verbose
+        # Git branch and git commit specified
+        elsif ( ! @pkg.git_hashtag.nil? || ! @pkg.git_hashtag.empty )  and ( ! @pkg.git_branch.nil? || ! @pkg.git_branch.empty?)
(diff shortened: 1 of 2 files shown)


bin/crew
```diff
@@ -700,11 +700,18 @@ def download
...
           puts "cachefile is #{cachefile}".orange if @opt_verbose
+        # Git branch and git commit specified
+        elsif ( ! @pkg.git_hashtag.nil? || ! @pkg.git_hashtag.empty )  and ( ! @pkg.git_branch.nil? || ! @pkg.git_branch.empty?)
+          cachefile = CREW_CACHE_DIR + filename + @pkg.git_branch.gsub(/[^0-9A-Za-z.\-]/, '_') + '_' + @pkg.git_hashtag + '.tar.xz'
+          puts "cachefile is #{cachefile}".orange if @opt_verbose
+        # Git branch specified, without a specific git commit.
         else
-          # Use to the day granularity for a branch timestamp.
+          # Use to the day granularity for a branch timestamp with no specific commit specified.
           cachefile = CREW_CACHE_DIR + filename + @pkg.git_branch.gsub(/[^0-9A-Za-z.\-]/, '_') + Time.now.strftime("%m%d%Y") + '.tar.xz'
```
uberhacker, line 715: You could include hours and minutes in the timestamp for more granularity.

satmandu: For sure, but the proper fix is probably to just add the commit SHA if a commit hash is provided, no?

That way you can modify the package file to point to a new commit in a branch, and you don't have to worry about granularity.

(I figure just pointing to the branch is something we are less likely to do, since it's nice to have an exact commit to make builds reproducible.)

satmandu: (To be clear, the addition here is really just to add a case for when git branch and git hashtag are both provided.)

PR #6382, opened 2021-11-10
Add edge package

uberhacker:
I know, just what we need, right?  Yet another web browser.  Microsoft Edge is the fast and secure browser that helps you protect your data and save time and money.  See https://www.microsoft.com/en-us/edge.

### DIFF
--- a/packages/edge.rb
+++ b/packages/edge.rb
@@ -1,0 +1,44 @@
+require 'package'
+
+class Edge < Package
+  description 'Microsoft Edge is the fast and secure browser'
+  homepage 'https://www.microsoft.com/en-us/edge'
+  @_ver = '95.0.1020.44'
+  version @_ver
+  compatibility 'x86_64'
+  license 'MIT'
+
+  source_url 'https://packages.microsoft.com/repos/edge/pool/main/m/microsoft-edge-stable/microsoft-edge-stable_95.0.1020.44-1_amd64.deb'
+  source_sha256 'dadb1ab841f49b660d4fb9dc764900766ed1ffa6d3bd56f3c9d2e4a3dc6097f9'
+
+  depends_on 'libcom_err'
+  depends_on 'sommelier'
+
+  def self.patch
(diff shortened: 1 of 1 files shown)


supechicken:
Fun fact: most of our gui packages are browsers😅

packages/edge.rb
```diff
@@ -0,0 +1,44 @@
...
+    FileUtils.mv './opt/microsoft/msedge', "#{CREW_DEST_PREFIX}/share"
+    FileUtils.ln_sf "#{CREW_PREFIX}/share/msedge/msedge", "#{CREW_DEST_PREFIX}/bin/edge"
+    FileUtils.ln_sf "#{CREW_PREFIX}/share/msedge/msedge", "#{CREW_DEST_PREFIX}/bin/msedge"
+    FileUtils.ln_sf "#{CREW_PREFIX}/share/msedge/msedge", "#{CREW_DEST_PREFIX}/bin/microsoft-edge"
+    FileUtils.ln_sf "#{CREW_PREFIX}/share/msedge/msedge", "#{CREW_DEST_PREFIX}/bin/microsoft-edge-stable"
+    FileUtils.ln_sf "#{CREW_MAN_PREFIX}/man1/microsoft-edge.1.gz", "#{CREW_DEST_MAN_PREFIX}/man1/edge.1.gz"
+    FileUtils.ln_sf "#{CREW_MAN_PREFIX}/man1/microsoft-edge.1.gz", "#{CREW_DEST_MAN_PREFIX}/man1/msedge.1.gz"
+
+    # Add icon for use with crew-launcher
+    system 'curl -L# https://cdn.icon-icons.com/icons2/2552/PNG/128/edge_browser_logo_icon_152998.png -o microsoft-edge.png'
```
supechicken, line 34: The icons are already included in the deb archive?

uberhacker: Unfortunately, no.  I looked already.  It would be nice if they did include them.  The .desktop file is there but no icons.

PR #6388, opened 2021-11-15
Zig 0.7.1 => 0.8.1

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/zig.rb
+++ b/packages/zig.rb
@@ -3,22 +3,29 @@ require 'package'
 class Zig < Package
   description 'Programming language designed for robustness, optimality, and clarity'
   homepage 'https://ziglang.org/'
-  version '0.7.1'
+  version '0.8.1'
   license 'MIT'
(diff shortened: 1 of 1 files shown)


packages/zig.rb
```diff
@@ -3,22 +3,29 @@
 class Zig < Package
   description 'Programming language designed for robustness, optimality, and clarity'
   homepage 'https://ziglang.org/'
-  version '0.7.1'
+  version '0.8.1'
   license 'MIT'
-  compatibility 'x86_64 aarch64'
-  source_url 'https://ziglang.org/download/0.7.1/zig-0.7.1.tar.xz'
```
supechicken, line 9: Building from source as the current version does?

uberhacker: No need.  The binaries provided work fine.  I always appreciate it when projects provide working binaries for all architectures.

uberhacker: And...compatibility is all now.

PR #6391, opened 2021-11-17
neofetch add binaries

### LINKED ISSUE #6383: Neofetch depends on make
Hi,
I got this error after trying to install neofetch:

```
chronos@localhost / $ crew install neofetch
neofetch: A fast, highly customizable system info script
https://github.com/dylanaraps/neofetch
Version: 7.1.0
License: MIT-with-advertising
#### 100.0% -=#=- #      #       #                                                                                                                                                
Neofetch archive downloaded.
Preconfiguring package...
neofetch failed to install: `make DESTDIR=/usr/local/tmp/crew/dest PREFIX=/usr/local install` exited with 127
```

I've solved it installing make. 

`chronos@localhost / $ crew install make
`

Neofetch should be marked as dependent on make.

chromebrew: 1.18.2 x86_64

Regards

satmandu:
Fixes #6383 

Works properly:
- [x] x86_64
- [x] armv7l
- [ ] i686 ?
```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=neofetch CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/neofetch.rb
+++ b/packages/neofetch.rb
@@ -3,11 +3,24 @@ require 'package'
 class Neofetch < Package
   description 'A fast, highly customizable system info script'
   homepage 'https://github.com/dylanaraps/neofetch'
-  version '7.1.0'
+  version '7.1.0-24b1'
   license 'MIT-with-advertising'
   compatibility 'all'
-  source_url 'https://github.com/dylanaraps/neofetch/archive/7.1.0.tar.gz'
-  source_sha256 '58a95e6b714e41efc804eca389a223309169b2def35e57fa934482a6b47c27e7'
+  source_url 'https://github.com/dylanaraps/neofetch.git'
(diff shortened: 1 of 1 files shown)


saltedcoffii:
Neofetch is pure bash, so there's no reason we can't use the same binary for all architectures.

saltedcoffii:
Maybe add an `is_arch_independent` to a package file like `is_fake` or something like that.

satmandu:
@uberhacker thoughts on having an`is_arch_independent` option?

uberhacker:
Maybe it's overkill to have `is_arch_independent`?  The Makefile basically doesn't compile but installs all the files in the correct location.  Still doesn't hurt to have binaries which can speed up installs.  In this case, the speed is probably negligible but it removes the need for build dependencies.

PR #6423, opened 2021-11-30
Add wayland_proxy_virtwl (a sommelier alternative)

satmandu:
- This is an alternative to `sommelier`: https://github.com/talex5/wayland-proxy-virtwl
- Wayland & X11 proxying works, albeit unaccelerated.
- Example usage: `/usr/local/opam/bin/wayland-proxy-virtwl  --wayland-display wayland-2 --x-display=2 --xrdb Xft.dpi:150`  
  - This should create a proxied wayland display at `wayland-2` which can be used by prefixing a wayland command with `WAYLAND_DISPLAY=wayland-2`.
  - e.g. `WAYLAND_DISPLAY=wayland-2 weston-terminal`
  - This also creates a proxied X display at `:2` which can be used by prefixing a x11 command with `DISPLAY=:2`.
  - e.g. `DISPLAY=:2 xterm`

Builds properly:
- [x] x86_64
- [x] armv7l 
- [x] i686 (X11 issues... ah well.)

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=wayland_proxy_virtwl_2 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/opam.rb
+++ b/packages/opam.rb
@@ -7,68 +7,65 @@ class Opam < Package
   description 'OCaml package manager'
   homepage 'https://opam.ocaml.org/'
   @_ver = '2.1.1'
-  version @_ver
+  version "#{@_ver}-2"
   compatibility 'all'
   source_url 'https://github.com/ocaml/opam.git'
   git_hashtag @_ver
 
   binary_url({
-    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/opam/2.1.1_armv7l/opam-2.1.1-chromeos-armv7l.tpxz',
-     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/opam/2.1.1_armv7l/opam-2.1.1-chromeos-armv7l.tpxz',
-       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/opam/2.1.1_i686/opam-2.1.1-chromeos-i686.tar.xz',
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/opam/2.1.1_x86_64/opam-2.1.1-chromeos-x86_64.tpxz'
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/opam/2.1.1-2_armv7l/opam-2.1.1-2-chromeos-armv7l.tpxz',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/opam/2.1.1-2_armv7l/opam-2.1.1-2-chromeos-armv7l.tpxz',
+       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/opam/2.1.1-2_i686/opam-2.1.1-2-chromeos-i686.tpxz',
+     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/opam/2.1.1-2_x86_64/opam-2.1.1-2-chromeos-x86_64.tpxz'
   })
   binary_sha256({
-    aarch64: '53d3b4e436c1a70f5772523f0af5454aa54ad7dfb777d50fe7495d31160b7462',
-     armv7l: '53d3b4e436c1a70f5772523f0af5454aa54ad7dfb777d50fe7495d31160b7462',
-       i686: '4d8e3979729315249b0a9dabf795108c91d16618dd5f38285dc758eabf3896d1',
-     x86_64: 'f0cced8d7504be9b5bbd2f98d346e28b9cd18cd8b02985722643854a5cd5d51f'
+    aarch64: 'ef542a60a56abda5549b9e0a07e76898315dd07149bda008f0498b5244ec79b8',
+     armv7l: 'ef542a60a56abda5549b9e0a07e76898315dd07149bda008f0498b5244ec79b8',
+       i686: '380e81c2ce976bf42ccfc18050598c50bf282b157a269f4dd179d0410ab2dbb9',
+     x86_64: 'cb95bae566579fcf5d1952afe41c89d40932a10458e90c00a0f812ae9ef0f43f'
   })
 
   depends_on 'bubblewrap'
   depends_on 'ocaml'
   depends_on 'rsync'
 
+  @OPAMROOT = "#{CREW_PREFIX}/share/opam"
+
   def self.build
     system "#{CREW_ENV_OPTIONS} ./configure #{CREW_OPTIONS}"
     system "make lib-ext all -j1 \
       OCAMLC='ocamlc -unsafe-string' \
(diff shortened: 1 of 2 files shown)


packages/wayland_proxy_virtwl.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  @OPAMROOT = "#{CREW_PREFIX}/share/opam"
+
+  def self.patch
+    # For more info see https://github.com/talex5/wayland-proxy-virtwl/issues/24
+    @xwayland_patch = <<~'PATCH_XWAYLAND_EOF'
+      --- a/wayland-proxy-virtwl/xwayland.ml     2021-11-18 20:34:27.000000000 +0000
+      +++ b/wayland-proxy-virtwl/xwayland.ml     2021-11-24 19:27:16.940505733 +0000
+      @@ -129,9 +129,7 @@ module Selection = struct
+           x11 : X11.Display.t Lwt.t;
```
supechicken, line 37: I’d more prefer using tools like `.sub` for patches, since it doesn’t need to re-generate a patch file after upstream updates (as the exact line number may be changed)

packages/opam.rb
```diff
@@ -7,68 +7,65 @@ class Opam < Package
...
+    if Dir.exist? @OPAMROOT.to_s
       puts
-      print "Would you like to remove #{CREW_PREFIX}/opam? [y/N] "
+      print "Would you like to remove #{@OPAMROOT}? [y/N] "
       response = $stdin.getc
       case response
       when 'y', 'Y'
-        FileUtils.rm_rf "#{CREW_PREFIX}/opam"
-        puts "#{CREW_PREFIX}/opam removed.".lightred
+        FileUtils.rm_rf @OPAMROOT.to_s
```
supechicken, line 65: Does we need the `.to_s` part (it is already a string)

packages/wayland_proxy_virtwl.rb
satmandu: How do I do that?

packages/opam.rb
satmandu: Rubocop added that. Probably not needed!

packages/wayland_proxy_virtwl.rb
satmandu: (In this case, though, I'm hoping that there will be sufficient upstream changes in an eventual tagged release making these patches unnecessary.)

supechicken: Unlike sed, the `.sub` function accepts multi line strings as the replace target, even using a mutil line regex :) (with //m regex)

packages/wayland_proxy_virtwl.rb
```diff
@@ -0,0 +1,143 @@
...
+    end
+  end
+
+  def self.postinstall
+    puts <<~'EOFSTRING'.lightblue
+      Note that there is not yet hardware acceleration in wayland-proxy-virtwl.
+      wayland-proxy-virtwl example usage (before running a gui program):
+    EOFSTRING
+    puts <<~'EOFSHELLCODE'.lightcyan
+      wayland-proxy-virtwl --wayland-display wayland-2 --x-display=2 --xrdb Xft.dpi:150 &
```
supechicken, line 138: Get the dpi value using xdpyinfo like sommelier does?
https://github.com/skycocker/chromebrew/blob/cb9a768917622f73ea0513010fb5ac4e8fd1009c/packages/sommelier.rb#L179

packages/opam.rb
```diff
@@ -7,68 +7,65 @@ class Opam < Package
...
       OCAMLOPT='ocamlopt -unsafe-string'"
-    @bashd_opam = <<~EOF
-      export OPAMROOT=#{CREW_PREFIX}/opam
-      eval \$(opam env --root=#{CREW_PREFIX}/opam --switch=default)
-      test -r #{CREW_PREFIX}/opam/opam-init/init.sh && . #{CREW_PREFIX}/opam/opam-init/init.sh > /dev/null 2> /dev/null || true
-    EOF
+    @bashd_opam = <<~OPAMEOF
+      export OPAMROOT=#{@OPAMROOT}
+      eval \$(opam env --root=#{@OPAMROOT} --switch=default)
+      test -r #{@OPAMROOT}/opam-init/init.sh && . #{@OPAMROOT}/opam-init/init.sh > /dev/null 2> /dev/null || true
```
supechicken, line 42: `&> /dev/null`?

packages/wayland_proxy_virtwl.rb
satmandu: That would be a good idea for a shell script for automatic setup, for sure.

(Also that's a chicken and egg issue... `xdpyinfo` won't work until a X11 Display is setup no?)

satmandu: I think the point of this PR is to offer a proof of concept... not to immediately replace `sommelier`.

packages/opam.rb
satmandu: This is intentionally echoing the format in that file:
```
cat /usr/local/share/opam/opam-init/init.sh
if [ -t 0 ]; then
  test -r /usr/local/share/opam/opam-init/complete.sh && . /usr/local/share/opam/opam-init/complete.sh > /dev/null 2> /dev/null || true

  test -r /usr/local/share/opam/opam-init/env_hook.sh && . /usr/local/share/opam/opam-init/env_hook.sh > /dev/null 2> /dev/null || true
fi

test -r /usr/local/share/opam/opam-init/variables.sh && . /usr/local/share/opam/opam-init/variables.sh > /dev/null 2> /dev/null || true
```

satmandu:
Rebuilding as per https://github.com/talex5/wayland-proxy-virtwl/issues/24#issuecomment-986012287

packages/wayland_proxy_virtwl.rb
supechicken: > That would be a good idea for a shell script for automatic setup, for sure.
> 
> (Also that's a chicken and egg issue... `xdpyinfo` won't work until a X11 Display is setup no?)

You’re right😅 Never mind

uberhacker:
@satmandu: Is this PR ready or still a WIP?

satmandu:
@uberhacker this is good to merge...

packages/opam.rb
uberhacker: I think what @supechicken means is you can replace `> /dev/null 2> /dev/null` with just `&> /dev/null`.

satmandu:
(just tested it on `x86_64`)

packages/opam.rb
satmandu: Understood. I'll respin.

satmandu: new binaries uploaded.

satmandu:
@uberhacker Take 2...

PR #6431, opened 2021-12-04
New musl toolchain w/ static git, curl, lzip (and gcc11 rebuilds to fix brokenness on i686)

satmandu:
- The goal of this PR is to have all early packages during install be statically linked to avoid dependencies on packages to be downloaded later, like newer libgcc or llvm libraries, using the musl libc, which unlike glibc allows binaries to be statically linked without dependence on a particular version of glibc sitting around.
- `curl` is no longer needed at install, so the need for a static `curl` has dropped out of the scope of this PR. A smaller static libcurl is being built for usage in linking to `git`.
- Static binary packages from this PR are `curl`, `git`, and `lzip`
- The musl toolchain is `musl_native_toolchain`
  - Toolchains for `x86_64` and `armv7l`, and `i686` are built from scratch.
- lzip is only dependent upon `musl_native_toolchain` for build.
- Build order for git: `musl_native_toolchain musl_zlib musl_libunistring musl_libidn2 musl_ncurses musl_openssl musl_zstd musl_brotli musl_expat musl_curl git`
- Other packages built as mostly static for musl, in rough order of necessary build: `musl_bz2 musl_xz lzip musl_cyrus_sasl musl_krb5 musl_c_ares musl_libnghttp2 musl_libssh  musl_libbacktrace musl_libucontext musl_libunwind musl_obstack musl_lz4 musl_wolfssl musl_perl`
- I had to revert to an older version of musl, since the current version has a `getaddrinfo` issue with the 3.8 i686 kernel.
- New versions of musl can be tested on the i686 musl by adjusting the musl git commit line in the `musl_native_toolchain` file `MUSL_VER = git-d1395c43c019aec6b855cf3c656bf47c8a719e7f` (I picked [this commit](https://git.musl-libc.org/cgit/musl/commit/src/network/getaddrinfo.c?id=d1395c43c019aec6b855cf3c656bf47c8a719e7f) because I wanted to try an older version of the `getaddrinfo.c` file to try to address this issue. But bringing that version of the file into  a current musl tree doesn't work to fix the issue. Feel free to bisect away!)
-  To test for a working musl, install a musl toolchain with that `MUSL_VER` changed, and then try to build `musl_getaddrinfo_test` on i686 hardware or in a VM running a 3.8.x kernel.
- A new `lib/musl.rb` is being used to abstract away all the musl variables.

- **curl -> 7.81.0**
- **git -> 2.34.1**

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=new_musl_toolchain CREW_TESTING=1 crew update
```


satmandu:
@saltedcoffii the `lzip` package here has a functioning static `lzip` for all architectures.

saltedcoffii:
Maybe using -O3 isn't the best idea. See https://wiki.gentoo.org/wiki/GCC_optimization#-O

satmandu:
https://github.com/skycocker/chromebrew/pull/6630 Has been added to this PR.

lib/musl.rb
```diff
@@ -0,0 +1,98 @@
...
+  class << self
+    def patchelf
+      puts 'Running Musl.patchelf'.lightblue
+      abort('No Patchelf found!').lightred unless File.exist?("#{CREW_PREFIX}/bin/patchelf")
+
+      @execfiles = `find . -executable -type f ! \\( -name \"*.a\" \\) -exec sh -c \"file -i \'{}\' | grep -q \'executable; charset=binary\'\" \\; -exec ls -1i {} \\; | sort -u -n -s -k1,1 | awk '{print $2}'`.chomp
+      return if @execfiles.empty? || @execfiles.nil?
+
+      puts 'Running patchelf to patch binaries for musl paths'.lightblue
+      @execfiles.each_line do |execfiletopatch|
```
supechicken, line 80: Use `chomp` option in `each_line` to pervent newlines?
```ruby
@execfiles.each_line(chomp: true) do
```

lib/musl.rb
```diff
@@ -0,0 +1,98 @@
...
+        execfiletopatch.slice! '.'
+        execfiletopatch = `pwd`.chomp + execfiletopatch.chomp
```
supechicken, line 82: Rubyize with `Dir.pwd`?

lib/musl.rb
supechicken: Sorry, should be "prevent"😅

lib/musl.rb
```diff
@@ -0,0 +1,98 @@
...
+# Defines common musl build constants used in musl builds
+MUSL_ABI = ''
+MUSL_ARCH = ARCH
+MUSL_ARCH_SSP_CFLAGS = ''
+MUSL_ARCH_C_FLAGS = ''
+MUSL_ARCH_CXX_FLAGS = ''
+case ARCH
+when 'aarch64', 'armv7l'
+  MUSL_ABI = 'eabihf'
+  MUSL_ARCH = 'armhf'
```
supechicken, line 14: Constant redefined?

lib/musl.rb
```diff
@@ -0,0 +1,98 @@
...
+      return if @execfiles.empty? || @execfiles.nil?
+
+      puts 'Running patchelf to patch binaries for musl paths'.lightblue
+      @execfiles.each_line do |execfiletopatch|
+        execfiletopatch.slice! '.'
+        execfiletopatch = `pwd`.chomp + execfiletopatch.chomp
+        system "patchelf --set-interpreter #{CREW_MUSL_PREFIX}/lib/libc.so #{execfiletopatch}"
+        system "patchelf --set-rpath #{CREW_MUSL_PREFIX}/lib #{execfiletopatch}"
+        @neededlibs = `patchelf --print-needed #{execfiletopatch}`.chomp
+        next if @neededlibs.empty? || @neededlibs.nil?
```
supechicken, line 86: `next if @neededlibs.to_s.empty?`

lib/musl.rb
```diff
@@ -0,0 +1,98 @@
...
+
+        @neededlibs.each_line do |neededlibspatch|
```
supechicken, line 88: Use `chomp` option in `each_line` to prevent newlines?
```ruby
@neededlibs.each_line(chomp: true) do
```

lib/musl.rb
```diff
@@ -0,0 +1,98 @@
...
+    def patchelf
+      puts 'Running Musl.patchelf'.lightblue
+      abort('No Patchelf found!').lightred unless File.exist?("#{CREW_PREFIX}/bin/patchelf")
+
+      @execfiles = `find . -executable -type f ! \\( -name \"*.a\" \\) -exec sh -c \"file -i \'{}\' | grep -q \'executable; charset=binary\'\" \\; -exec ls -1i {} \\; | sort -u -n -s -k1,1 | awk '{print $2}'`.chomp
+      return if @execfiles.empty? || @execfiles.nil?
+
+      puts 'Running patchelf to patch binaries for musl paths'.lightblue
+      @execfiles.each_line do |execfiletopatch|
+        execfiletopatch.slice! '.'
```
supechicken, line 81: Will `.delete_prefix` better (since it will not break files that contains `.` in the filename)

satmandu: Let me try that!

satmandu:
@supechicken's suggested changes will be in a new PR.

lib/musl.rb
satmandu: `execfiletopatch = Dir.pwd + execfiletopatch.delete_prefix('.')` looks like it does the trick!

PR #6442, opened 2021-12-05
Gcloud 362.0.0 => 365.0.1

uberhacker:


satmandu:
Sigh. Sad how Google doesn't even bother to support armv7l userspace any more. Maybe we should offer a `aarch64`  userspace crostini alternative which isn't as locked down as Crostini for newer Chromebook users running arm64 kernels?

PR #6295, opened 2021-10-13
libffi => 3.4.2 & python3 => 3.10.1, and all downstream packages requiring updates

saltedcoffii:
Upgrade of libffi to 3.4.2 & python to 3.10
Before we merge this pull request, it is imperative that **ALL** packages depending on python3 for build or runtime rebuild and preferably upgraded to their latest version. I can help with this but would love backup, either from a human, or if anyone knows of an update checker script, that would be really helpful as well.

All packages requiring libffi also need updating.

This will tell you what needs updating on a `x86_64` system for the libffi update:
```for i in /usr/local/lib64/*.so ; do ldd $i 2>/dev/null | grep libffi.so.7 &>/dev/null && crew whatprovides $i | head -n 1 | sed 's/://' | awk '{print $1'} ; done | sort -u```

**All packages in the list that command gives need to be updated in this PR.**

That list may include at a minimum the following, but this list may shrink as common dependencies get updated.

- [x] cairo
- [x] cups
- [x] ffmpeg
- [x] gjs
- [x] glib
- [x] gnutls
- [x] gobject_introspection
- [x] llvm
- [x] p11kit
- [x] rtmpdump
- [x] wayland

### Run the following to get this pull request's changes locally for testing.
```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=python3.10 CREW_TESTING=1 crew update
```

PR #6464, opened 2021-12-08
crew: use hard links in packaging; after git download fetch git tags

satmandu:
- This changes crew to use `rdfind` to hard link duplicate files.
  - Upx now copies a file to be compressed to a tmp file, and then overwrites the original only if  the compressed tmp file can also be extracted successfully.
  - The search for files to be compressed now excludes more than one copy of a hard linked file.
  - This leads to the git taking up ~ 30Mb of space after install.
- The git change fixes the compile of `py3_numpy` when installed from git source, since the `py3_numpy` install extracts version information from git tags. (There is other software which does this too.) Relevant to https://github.com/skycocker/chromebrew/pull/6295#issuecomment-988890330

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686
```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=newcrew CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -786,7 +786,9 @@ def download
       if CREW_CACHE_ENABLED and File.writable?(CREW_CACHE_DIR)
         puts 'Caching downloaded git repo...'
         Dir.chdir "#{@extract_dir}" do
-          system "tar c#{@verbose}Jf #{cachefile} --exclude-vcs \
+          # Do not use --exclude-vcs to exclude .git
+          # because some builds will use that information.
+          system "tar c#{@verbose}Jf #{cachefile} \
             $(find -mindepth 1 -maxdepth 1 -printf '%P\n')"
         end
         system "sha256sum #{cachefile} > #{cachefile}.sha256"
--- a/bin/crew
+++ b/bin/crew
@@ -779,6 +779,7 @@ def download
           system 'git checkout FETCH_HEAD'
         end
         system 'git submodule update --init --recursive'
+        system 'git fetch --tags', exception: true
         puts 'Repository downloaded.'.lightgreen
       end
       # Stow file in cache if requested and cache is writable.
@@ -1035,8 +1036,8 @@ def shrink_dir(dir)
   unless CREW_SHRINK_ARCHIVE == '0'
     Dir.chdir dir do
       if File.exist?("#{CREW_PREFIX}/bin/rdfind")
-        puts "Using rdfind to find duplicate or hard linked files."
-        system "#{CREW_PREFIX}/bin/rdfind -removeidentinode true -makesymlinks true -makeresultsfile false ."
+        puts "Using rdfind to convert duplicate files to hard links."
+        system "#{CREW_PREFIX}/bin/rdfind -removeidentinode true -makesymlinks false -makehardlinks true -makeresultsfile false ."
       end
       # Issues with non-x86_64 in compressing libraries, so just compress
       # non-libraries. Also note that one needs to use "upx -d" on a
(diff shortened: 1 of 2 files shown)


saltedcoffii:
Correct me if I'm wrong. A hard link causes two duplicate files to reference the same inode, saving space on the disk. However, should those files be copied, for instance, into a tarball and onto another computer, like we do when we distribute binaries, I don't think those hard links will maintain, instead becoming duplicate files again. Can we check for duplicate files when installing binaries as well as when installing from source?

satmandu:
tar appears to preserve hard links by default.

satmandu:
>  Can we check for duplicate files when installing binaries as well as when installing from source?

We could add `system "#{CREW_PREFIX}/bin/rdfind -removeidentinode true -makesymlinks false -makehardlinks true -makeresultsfile false ."` to `def install_package(pkgdir)`

satmandu:
(To be clear... hard linking during build saves a bunch of time during packaging, since multiple copies of the same binary only have to get compressed once.)

saltedcoffii:
As long as hard links are preserved in tarballs, I'm fine, I don't see any changes that need to be made :)

saltedcoffii (approved):

saltedcoffii:
Be sure to bump version before merge

bin/crew
```diff
@@ -1035,30 +1038,44 @@ def shrink_dir(dir)
   unless CREW_SHRINK_ARCHIVE == '0'
     Dir.chdir dir do
       if File.exist?("#{CREW_PREFIX}/bin/rdfind")
-        puts "Using rdfind to find duplicate or hard linked files."
-        system "#{CREW_PREFIX}/bin/rdfind -removeidentinode true -makesymlinks true -makeresultsfile false ."
+        puts "Using rdfind to convert duplicate files to hard links."
+        system "#{CREW_PREFIX}/bin/rdfind -removeidentinode true -makesymlinks false -makehardlinks true -makeresultsfile false ."
```
uberhacker, line 1042: Shouldn't we be removing duplicates instead?  The idea is to reduce clutter and save space, correct?

saltedcoffii: This does remove duplicates. What behavior specifically are you not seeing that you would like to see?

satmandu: Hard linking solves the space issue because hard linked files don't take up any more space.

Symlinking the duplicate files ends up wreaking havoc because creating proper relative symlinks isn't trivial, and absolute symlinks get broken during package and install...

uberhacker: I guess I'm wondering why there are any duplicates in the first place?  Why is it necessary to have hard links?

satmandu: Packages like git will link a bunch of apps to one binary.

I think another package which does this is BusyBox?

When we use upx or another compression program on such a binary, it will replace the binary with a compressed one, but break the hard link.

E.g.

```
ln git git2
upx --lzma git
```
(Now you have two files — a compressed git binary and an uncompressed git2 binary.)

However
```
ln git git2
cp git git.tmp
upx --lzma git.tmp
cp git.tmp git
rm git.tmp
```
Now you have a compressed `git` and `git2`, and both are hard linked to each other.

uberhacker:
I don't know what happened to your comment (appeared in email but disappeared here) but that was an excellent explanation to help me understand what is going on here with these hard links.

satmandu:
> @satmandu: I don't know what happened to your comment (appeared in email but disappeared here) but that was an excellent explanation to help me understand what is going on here with these hard links.

Here: https://github.com/skycocker/chromebrew/pull/6464#discussion_r765413793

PR #6470, opened 2021-12-09
curl: Use dynamic curl on i686

satmandu:
- Current musl-derived build of curl is having issues on `i686`, so copy over the curl.nonstatic from the `libcurl` package for now at install via postinstall.
- Now properly using ruby method for overwriting the file...
- This keeps further package upgrades from breaking on i686 subsequent to install.

Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=curl_i686_fix CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/curl.rb
+++ b/packages/curl.rb
@@ -219,8 +219,8 @@ class Curl < Package
 
   def self.postinstall
     if ARCH == 'i686'
-      puts 'Static curl is broken on i686, so use dynamic curl from libcurl package.'.orange
-      FileUtils.cp "#{CREW_PREFIX}/bin/curl.nonstatic", "#{CREW_PREFIX}/bin/curl"
+      puts 'Static curl is broken on i686, so using curl from the libcurl package.'.orange
+      FileUtils.cp_r "#{CREW_PREFIX}/bin/curl.nonstatic", "#{CREW_PREFIX}/bin/curl", remove_destination: true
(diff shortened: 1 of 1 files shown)


packages/curl.rb
```diff
@@ -219,8 +219,8 @@ def self.install
 
   def self.postinstall
     if ARCH == 'i686'
-      puts 'Static curl is broken on i686, so use dynamic curl from libcurl package.'.orange
-      FileUtils.cp "#{CREW_PREFIX}/bin/curl.nonstatic", "#{CREW_PREFIX}/bin/curl"
+      puts 'Static curl is broken on i686, so using curl from the libcurl package.'.orange
+      FileUtils.cp_r "#{CREW_PREFIX}/bin/curl.nonstatic", "#{CREW_PREFIX}/bin/curl", remove_destination: true
```
uberhacker, line 223: What is recursive here?  Maybe `cp_f` is what you need to force overwrite?

satmandu: `cp_f` doesn't exist, as I discovered. See https://ruby-doc.org/stdlib-2.4.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-cp

PR #6471, opened 2021-12-09
Keep crew from breaking w/ missing package files for installed packages

satmandu:
- This is a common occurrence when developing packages.

Works properly:
- [x] x86_64

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=crew_missing_packages_do_not_break CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -594,6 +594,10 @@ def upgrade
     # Make an installed packages list belong to the dependency order
     dependencies = []
     @device[:installed_packages].each do |package|
+      if not File.exist?(CREW_PACKAGES_PATH + package[:name] + '.rb')
+        puts "Package file for installed package #{package[:name]} is missing.".lightred
+        next
+      end
       # skip package if it is dependent other packages previously checked
       next if dependencies.include? package[:name]
       # add package itself
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.19.2'
+CREW_VERSION = '1.19.3'
 
 ARCH_ACTUAL = `uname -m`.chomp
 # This helps with virtualized builds on aarch64 machines


uberhacker:
So this just notifies if there is a package missing.  How would we "sync up" the metadata?

satmandu:
I'm not sure we should throw away the data?

If a package is reported as installed, but the package file is missing, then it is useful to keep that info in case the package file reappears.

I think this gets complicated quickly...

Having said that, this is another way to get that info:
```
#!/bin/bash
# crewtidy
crewconst() {
local var="$1"
local const=$(crew const | grep $var | awk -F= '{print $2}')
echo ${const%/}
}
CREW_CONFIG_PATH=$(crewconst CREW_CONFIG_PATH)
CREW_LIB_PATH=$(crewconst CREW_LIB_PATH)
CREW_META_PATH=$(crewconst CREW_META_PATH)

echo "Making sure all core packages are installed."
crew_installed_packages=$(cat "${CREW_CONFIG_PATH}/device.json" | jq  '.installed_packages[] .name')
for i in $(cat "${CREW_LIB_PATH}/tools/core_packages.txt")
do
( [[ $crew_installed_packages == *\"$i\"* ]] || yes |  crew install $i || true)
done

echo "Checking if there are any orphaned packages."
for i in $(cd ${CREW_META_PATH} && ls *.filelist)
do
if [[ $crew_installed_packages != *\"${i%.filelist}\"* ]]; then
  conflicts=$(grep --exclude ${CREW_META_PATH}/${i}.filelist \
      -Fxf ${CREW_META_PATH}/${i}.filelist ${CREW_META_PATH}/*.filelist | \
      tr ':' ' ' | \
      sed 's,.filelist,,g' | sed "s,${CREW_META_PATH},,g")
  echo "${i%.filelist} has a filelist but not recorded as installed."
  echo "You can remove orphaned files from that package by running:"
  echo "crew remove ${i%.filelist}"
  [[ -n $conflicts ]] && echo "crew reinstall ${conflicts}"
fi
done
crew_installed_packages=$(cat "${CREW_CONFIG_PATH}/device.json" | jq  '.installed_packages[] .name')

echo "Checking if there are non-'is_fake' packages recorded as installed"
echo "which do not have filelists."
for i in $(echo $crew_installed_packages | tr -d '"')
do
if ! ls ${CREW_META_PATH}/$i.filelist &>/dev/null ; then
  if ! grep -q "is_fake" $CREW_LIB_PATH/packages/${i}.rb; then
    echo "${i%.filelist} is recorded as installed but has no filelist."
    echo "You can reinstall that package by running:"
    echo "crew reinstall ${i%.filelist}"
  fi
fi
done
```

uberhacker:
It would be great to have a tidy script like this to sync up the metadata.  I have also noticed on occasion this metadata can become corrupted if attempting to install a package that fills up the disk space.  Once the metadata is hosed in this way, currently, the only solution is to reinstall crew.

satmandu:
Maybe the writes to the metadata file could be atomic?

Write to a new file, then copy over the old file?

satmandu:
That would require storing uncompressed & extracted package size metadata for each package, which we are not doing at the moment...

PR #6472, opened 2021-12-11
Allow customization of chromebrew's downloader

saltedcoffii:
Allow an environment variable to influence what software Chromebrew uses to download files. Currently supported downloaders are `aria2c`, `curl`, `wget` and the system `curl`. I plan to open further pull requests to support fetch and httpie.

@supechicken can you help fix my really bad ruby? Thank you.


uberhacker:
This is a great idea @saltedcoffii, however, I don't think filename should be a required parameter for `download_file`.  For example, with `curl -#LO <url>`, it will default to the filename from the url.  Just my 2 cents.

uberhacker:
It might make sense to allow for additional options too, like verbosity for debugging as an example.  Also, there probably should be a check to see if the executable exists first and suggest to install the package like aria2, for instance, if not found.  Rather than aborting altogether if the executable is missing, I'm thinking it should fallback to /usr/bin/curl and continue with a message indicating the download executable needs installed.  Should we add a constant in `lib/const.rb` also?

lib/downloader.rb
```diff
@@ -0,0 +1,35 @@
...
+
+def notimplemented(package)
+  puts "#{package} has not been implemented as a downloader for Chromebrew yet. Defaulting to curl.".orange
+  @_crew_downloader = "curl -#LC -"
+  def download_file(url, filename)
+    system "#{@_crew_downloader} \'#{url}\' --output #{filename}"
+  end
+end
+
+case ENV["CREW_DOWNLOADER"]
```
supechicken, line 12: We can use `.to_s` so that the value is always a string even if it is not set

lib/downloader.rb
```diff
@@ -0,0 +1,35 @@
...
+when 'aria2', 'aria2c'
+  def download_file(url, filename)
```
supechicken, line 14: Put the case block in the function?
```ruby
def download_file (url, filename)
  case …
  when …
  end
end
```

lib/downloader.rb
```diff
@@ -0,0 +1,35 @@
...
+when 'wget'
+  def download_file(url, filename)
+    system "wget -c \'#{url}\' --output-document #{filename}"
+  end
+when 'system curl', 'system'
+  def download_file(url, filename)
+    system "/usr/bin/curl -#LC - \'#{url}\' --output #{filename}"
+  end
+when 'fetch', 'bsdfetch', 'libfetch'
+  notimplemented 'BSD Fetch'
```
supechicken, line 30: Use this message for all unknown/unsupported downloaders? (replace the `unknown value` message with this in `else` block)

lib/downloader.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+when 'curl', '', nil
+  def download_file(url, filename)
+    system "curl -#LC - \'#{url}\' --output #{filename}"
+  end
+when 'wget'
+  def download_file(url, filename)
+    system "wget -c \'#{url}\' --output-document #{filename}"
+  end
+when 'system curl', 'system'
```
supechicken, line 25: Combine this with the `curl` option?
```ruby
when 'curl', '/usr/bin/curl', ''
  cmd = ( ENV['CREW_DOWNLOADER'] || 'curl' )
  system "#{cmd} -#LC - '#{url}' -o #{filename}"
```

lib/downloader.rb
```diff
@@ -0,0 +1,35 @@
...
+when 'aria2', 'aria2c'
+  def download_file(url, filename)
+    system "aria2c -c \'#{url}\' --out #{filename}"
+  end
+when 'curl', '', nil
+  def download_file(url, filename)
+    system "curl -#LC - \'#{url}\' --output #{filename}"
+  end
+when 'wget'
+  def download_file(url, filename)
```
supechicken, line 22: Make the filename argument optional? (extract filename from url if not given):
```ruby
def download_file ( url, filename = File.basename(url) )
```

lib/downloader.rb
supechicken: We may also unset the LD_LIBRARY_PATH to get rid of segfails when libs in CREW_PREFIX are corrupted

lib/downloader.rb
```diff
@@ -0,0 +1,35 @@
...
+    system "/usr/bin/curl -#LC - \'#{url}\' --output #{filename}"
+  end
+when 'fetch', 'bsdfetch', 'libfetch'
+  notimplemented 'BSD Fetch'
+when 'httpie'
+  notimplemented 'HTTPie'
+else
+  abort "CREW_DOWNLOADER is set to an unknown value. Exiting.".lightred
+end
```
supechicken, line 35: Maybe add a rescue block that will retry if the download command above failed:
```ruby
@retry_times = 0
unless $?.exitstatus == 0
  if @retry_times == 3
    abort 'Download failed :/'.lightred
  end
  puts "Downloader exited with status #{$?.exitstatus}, retrying…".yellow
  @retry_times += 1
  download_file(url, filename)
end
```

lib/downloader.rb
saltedcoffii: Good call, thank you.

lib/downloader.rb
saltedcoffii: That's why I have the `nil` in there.

lib/downloader.rb
saltedcoffii: That makes sense. thanks.

lib/downloader.rb
saltedcoffii: I can see why that would make sense DRY wise, however it doesn't work the `system` or `system curl` parameters.

lib/downloader.rb
saltedcoffii: If I recall correctly, this is built into every downloader we're using except fetch.

lib/downloader.rb
saltedcoffii: I did this a little differently, but it works :)

bin/crew
```diff
@@ -652,6 +652,24 @@ def upgrade
   end
 end
 
+def download_file(url, filename = url.split('/')[-1])
```
supechicken, line 655: Be notice that if the function is defined in the main script, it cannot called by the package recipe (put it in a separate file and `require` it in package.rb)

lib/downloader.rb
supechicken: You’re right but the `—retry` option in `curl` will not handle some server errors.

bin/crew
saltedcoffii: really? I didn't have a problem with it. just call the function `download_file`

supechicken: Oh, looks like I am wrong😅

satmandu (approved):

uberhacker:
Can we replace `download_file` with @supechicken's `downloader` from #6491?

saltedcoffii:
> Can we replace `download_file` with @supechicken's `downloader` from #6491?
sure whatever works for you. however it would be nice to use the internal net/http downloader by default and still offer an option to use curl, aria2 or whatever via the same environment variable.

uberhacker:
> > Can we replace `download_file` with @supechicken's `downloader` from #6491?
> 
> sure whatever works for you. however it would be nice to use the internal net/http downloader by default and still offer an option to use curl, aria2 or whatever via the same environment variable.

If you can figure out how to trap errors like `net/http` does with the other download options then it might be justified.  Otherwise, we are just doing more of the same of what we have now (no error trapping).

saltedcoffii:
fair. alright let's have error tapping :3

saltedcoffii:
closing because #6491 is better

PR #6475, opened 2021-12-11
Check if downloaded file is empty

saltedcoffii:
Sometimes curl will successfully connect to a server but then not download anything and exist successfully. I have no idea why, but it happens on my Wi-Fi. This will describe this error more specifically, instead of just saying “Checksum mismatch, try again.”

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -558,7 +558,7 @@ def update
     if @pkg.nil?
       puts "Package file for #{package[:name]} not found. :(".lightred if @opt_verbose
       next
-    end 
+    end
     if package[:version].to_s != @pkg.version
       canBeUpdated += 1
       puts @pkg.name + ' could be updated from ' + package[:version].to_s + ' to ' + @pkg.version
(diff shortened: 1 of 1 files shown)


bin/crew
```diff
@@ -708,6 +708,8 @@ def download
       end
       # Download file if not cached.
       system "#{CURL} --retry 3 -#{@verbose}#LC - --insecure \'#{url}\' --output #{filename}"
+      abort 'Nothing was downloaded. Check your internet or try again later.'.lightred unless
```
uberhacker, line 711: It's also possible the url cannot be found.  In that case, the error message is still a bit confusing.  Maybe retrieving the error code and displaying the message generated from the response is even more helpful and accurate?  It might make sense to use a ruby library.  See https://ruby-doc.org/stdlib-3.0.3/libdoc/net/http/rdoc/Net/HTTP.html.

saltedcoffii: I honestly don't even know where to start with that.

saltedcoffii:
Closing because #6491 is better

PR #6502, opened 2021-12-18
new package info2man

saltedcoffii:
Info2man converts GNU info pages to manpages or to POD (perl documentation) pages. Nobody seems to know the license (not even Debian), and it seems to be abandonware. It, however, works. It also doesn't install like most perl packages via makefile.PL, so unfortunately this install is a bit hacky.

Works on x86_64, and is pure perl, so should work on all architectures. I'm not sure if it needs binaries because its just copying stuff around, nothing is being built, and its only dependency, perl, is a core package.


packages/info2man.rb
```diff
@@ -0,0 +1,26 @@
...
+  description 'Convert GNU info files to POD or man pages.'
+  homepage 'https://www.cskk.ezoshosting.com/cs/css/info2pod.html'
+  version '3.9.8'
+  license 'unknown'
+  compatibility 'all'
+  source_url 'https://salsa.debian.org/debian/info2man.git'
+  git_hashtag 'debian/1.1-10'
+
+  def self.patch
+    system "sed -i 's:#!/usr/bin/perl:#!#{CREW_PREFIX}/bin/perl:g' 0README.txt info2pod"
```
uberhacker, line 13: Should there be a `0` in front of `README.txt`?

packages/info2man.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin/"
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/lib/perl5/site_perl/#{%x[echo "printf '%vd\n', $^V;" | perl].chomp}/"
+    FileUtils.mkdir_p "#{CREW_DEST_MAN_PREFIX}/man1/"
+    FileUtils.install 'info2man', "#{CREW_DEST_PREFIX}/bin/info2man", mode: 0o755
+    FileUtils.install 'info2pod', "#{CREW_DEST_PREFIX}/bin/info2pod", mode: 0o755
+    FileUtils.cp_r 'cs', "#{CREW_DEST_PREFIX}/lib/perl5/site_perl/#{%x[echo "printf '%vd\n', $^V;" | perl].chomp}/"
```
uberhacker, line 22: Would it make sense to assign something like `perl_version = #{%x[echo "printf '%vd\n', $^V;" | perl].chomp}` and use `#{perl_version}` here since it's being referenced more than once?

packages/info2man.rb
uberhacker: If there is, it would make sense to rename to `README.txt`.

packages/info2man.rb
```diff
@@ -0,0 +1,26 @@
+require 'package'
+
+class Info2man < Package
+  description 'Convert GNU info files to POD or man pages.'
+  homepage 'https://www.cskk.ezoshosting.com/cs/css/info2pod.html'
```
uberhacker, line 5: I get a 404 for this url.

packages/info2man.rb
saltedcoffii: The README.txt isn't installed, it's just used in the building of the package. I wouldn't worry about it.

packages/info2man.rb
saltedcoffii: Good call.

packages/info2man.rb
uberhacker: Then you shouldn't worry about correcting it either. :)

packages/info2man.rb
saltedcoffii: Me too, that's the one used by Debian and the AUR. The next best URL is the salsa.debian.org one? @uberhacker let me know what you'd prefer me to use

uberhacker: Good question.  The one I found that is least like a man or package page is http://freshmeat.sourceforge.net/projects/info2man.

saltedcoffii: I'm having an issue with my HTTPS certificate on that site.

uberhacker: That means a cert is not installed at that domain.

saltedcoffii:
@uberhacker @satmandu what's the status on this pull request?

PR #5967, opened 2021-07-15
install.sh: Fetch tarball of repo directly and simply

supechicken:
Fix #5839 

Only tested for new parts separately, haven’t tried with full script

PR #6519, opened 2021-12-25
install.sh: Improve stability of `curl`, disallow crostini/beta/canary channel

supechicken:
Merged from #5954 and #5967 

### Changes
---
- Added a retry loop for `curl` (the `--retry 3 --retry-all-errors` parameter does not work with the `Couldn't connect to server ` error)

Examples:
```
Downloading ca_certificates...
curl: (7) Couldn't connect to server                                          

Retrying, 3 retries left.
```
```
Downloading gmp...
curl: (6) Could not resolve host: gitlab.com
Retrying, 3 retries left.
curl: (6) Could not resolve host: gitlab.com
Retrying, 2 retries left.
curl: (6) Could not resolve host: gitlab.com
Retrying, 1 retries left.
curl: (6) Could not resolve host: gitlab.com
Retrying, 0 retries left.
Download failed :/ Please check your network settings.
```
---
- Force `curl` to use system libraries (`unset LD_LIBRARY_PATH`), use `ssl`
- Enable parallel download on `curl` if available
- Optimize the `url/sha256` extract logic
---
- Disallow Linux containers (`crostini`) and devices on `beta/dev/canary` update channels
```
$ curl -Ls https://raw.githubusercontent.com/supechicken666/chromebrew/install_sh/install.sh | bash
Crostini containers are not supported by Chromebrew :/
Run 'curl -Ls git.io/vddgY | CREW_FORCE_INSTALL=1 bash' to perform install anyway
```
---
Tested on `x86_64`

```
curl -Ls https://raw.githubusercontent.com/supechicken666/chromebrew/install_sh/install.sh | bash
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,18 +4,18 @@
 set -e
 
 #chromebrew directories
-: ="${OWNER:=skycocker}"
+: "${OWNER:=skycocker}"
 : "${REPO:=chromebrew}"
 : "${BRANCH:=master}"
 URL="https://raw.githubusercontent.com/${OWNER}/${REPO}/${BRANCH}"
-CREW_PREFIX="${CREW_PREFIX:-/usr/local}"
+: "${CREW_PREFIX:=/usr/local}"
 CREW_LIB_PATH="${CREW_PREFIX}/lib/crew"
 CREW_CONFIG_PATH="${CREW_PREFIX}/etc/crew"
 CREW_BREW_DIR="${CREW_PREFIX}/tmp/crew"
 CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
 CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 : "${CURL:=/usr/bin/curl}"
-CREW_CACHE_DIR="${CREW_CACHE_DIR:-$CREW_PREFIX/tmp/packages}"
+: "${CREW_CACHE_DIR:=$CREW_PREFIX/tmp/packages}"
 # For container usage, where we want to specify i686 arch
 # on a x86_64 host by setting ARCH=i686.
 : "${ARCH:=$(uname -m)}"
@@ -156,10 +156,10 @@ fi
 for package in $BOOTSTRAP_PACKAGES; do
   pkgfile="${CREW_PACKAGES_PATH}/${package}.rb"
 
-  [[ "$(sed -n '/binary_sha256/,/}/p' "${pkgfile}")" =~ .*x86_64:[[:blank:]]*[\'\"]([^\'\"]*) ]]
+  [[ "$(sed -n '/binary_sha256/,/}/p' "${pkgfile}")" =~ .*${ARCH}:[[:blank:]]*[\'\"]([^\'\"]*) ]]
     sha256s+=("${BASH_REMATCH[1]}")
 
-  [[ "$(sed -n '/binary_url/,/}/p' "${pkgfile}")" =~ .*x86_64:[[:blank:]]*[\'\"]([^\'\"]*) ]]
+  [[ "$(sed -n '/binary_url/,/}/p' "${pkgfile}")" =~ .*${ARCH}:[[:blank:]]*[\'\"]([^\'\"]*) ]]
     urls+=("${BASH_REMATCH[1]}")
 done
 
--- a/install.sh
+++ b/install.sh
@@ -47,7 +47,7 @@ fi
 # disallow non-stable channels Chrome OS
 if [ -f /etc/lsb-release ]; then
   if [[ ! "$(< /etc/lsb-release)" =~ CHROMEOS_RELEASE_TRACK=stable-channel$'\n' && "${CREW_FORCE_INSTALL}" != '1' ]]; then
-    echo -e "${YELLOW}The beta, dev, and canary channel are unsupported by Chromebrew${RESET}"
+    echo -e "${RED}The beta, dev, and canary channel are unsupported by Chromebrew${RESET}"
     echo -e "${YELLOW}Run 'curl -Ls git.io/vddgY | CREW_FORCE_INSTALL=1 bash' to perform install anyway${RESET}"
     exit 1
   fi
@@ -55,11 +55,12 @@ else
   echo -e "${YELLOW}Unable to detect system information, installation will continue.${RESET}"
 fi
 
-echo -e "\n${GREEN}Welcome to Chromebrew!${RESET}\n"
 if [ "${EUID}" == "0" ]; then
   echo -e "${RED}Chromebrew should not be installed or run as root.${RESET}"
   exit 1;
 fi
+
+echo -e "\n${GREEN}Welcome to Chromebrew!${RESET}\n"
 
 # prompt user to enter the sudo password if it set
 # if the PASSWD_FILE specified by chromeos-setdevpasswd exist, that means a sudo password is set
@@ -80,7 +81,7 @@ function curl () {
       return 0 || \
       echo -e "${YELLOW}Retrying, $((3-$i)) retries left.${RESET}"
   done
-  # the download failed if we still here
+  # the download failed if we're still here
   echo -e "${RED}Download failed :/ Please check your network settings.${RESET}"
   return 1
 }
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -30,16 +30,61 @@ RED='\e[1;91m';    # Use Light Red for errors.
...
+  if [[ ! "$(< /etc/lsb-release)" =~ CHROMEOS_RELEASE_TRACK=stable-channel$'\n' && "${CREW_FORCE_INSTALL}" != '1' ]]; then
+    echo -e "${YELLOW}The beta, dev, and canary channel are unsupported by Chromebrew${RESET}"
+    echo -e "${YELLOW}Run 'curl -Ls git.io/vddgY | CREW_FORCE_INSTALL=1 bash' to perform install anyway${RESET}"
+    exit 1
+  fi
+else
+  echo -e "${YELLOW}Unable to detect system information, installation will continue.${RESET}"
+fi
+
+echo -e "\n${GREEN}Welcome to Chromebrew!${RESET}\n"
```
uberhacker, line 58: This line should probably go beneath lines 59-62 below.

install.sh
```diff
@@ -30,16 +30,61 @@ RED='\e[1;91m';    # Use Light Red for errors.
...
+if [[ -d /opt/google/cros-containers && "${CREW_FORCE_INSTALL}" != '1' ]]; then
+  echo -e "${RED}Crostini containers are not supported by Chromebrew :/${RESET}"
+  echo -e "${YELLOW}Run 'curl -Ls git.io/vddgY | CREW_FORCE_INSTALL=1 bash' to perform install anyway${RESET}"
+  exit 1
+fi
+
+# disallow non-stable channels Chrome OS
+if [ -f /etc/lsb-release ]; then
+  if [[ ! "$(< /etc/lsb-release)" =~ CHROMEOS_RELEASE_TRACK=stable-channel$'\n' && "${CREW_FORCE_INSTALL}" != '1' ]]; then
+    echo -e "${YELLOW}The beta, dev, and canary channel are unsupported by Chromebrew${RESET}"
```
uberhacker, line 50: `{RED}` instead of `{YELLOW}`?

install.sh
```diff
@@ -30,16 +30,61 @@ RED='\e[1;91m';    # Use Light Red for errors.
...
   exit 1;
 fi
 
+# prompt user to enter the sudo password if it set
+# if the PASSWD_FILE specified by chromeos-setdevpasswd exist, that means a sudo password is set
+if [[ "$(< /usr/sbin/chromeos-setdevpasswd)" =~ PASSWD_FILE=\'([^\']+) ]] && [ -f "${BASH_REMATCH[1]}" ]; then
+  echo -e "${BLUE}Please enter the developer mode password${RESET}"
+  # reset sudo timeout
+  sudo -k
+  sudo /bin/true
```
uberhacker, line 70: Are 2 sudo lines necessary?

install.sh
```diff
@@ -30,16 +30,61 @@ RED='\e[1;91m';    # Use Light Red for errors.
...
+# force curl to use system libraries
+function curl () {
+  # retry if download failed
+  # the --retry/--retry-all-errors parameter in curl will not work with the 'curl: (7) Couldn't connect to server'
+  # error, a for loop is used here
+  for (( i = 0; i < 4; i++ )); do
+    env LD_LIBRARY_PATH='' ${CURL} --ssl -C - "${@}" && \
+      return 0 || \
+      echo -e "${YELLOW}Retrying, $((3-$i)) retries left.${RESET}"
+  done
```
uberhacker, line 82: Just so you know, this will be 4 retries instead of 3.

install.sh
```diff
@@ -30,16 +30,61 @@ RED='\e[1;91m';    # Use Light Red for errors.
...
+  # the download failed if we still here
```
uberhacker, line 83: `if we're still here`?

install.sh
supechicken: The first `sudo` is used to reset the password timeout, so we can confirm that a password prompt will be shown in next sudo command

A password prompt will show in the second `sudo`, then the following `sudo` will not ask for the password (unless the `sudo` password timestamp expired)

install.sh
```diff
@@ -64,11 +109,9 @@ do
...
 sudo chown "$(id -u)":"$(id -g)" "${CREW_PREFIX}"
 
-# Delete 'var' symlink on Cloudready platform
-if [[ $(grep neverware /etc/lsb-release) != "" ]]; then
-  [ -L /usr/local/var ] && sudo rm -f /usr/local/var
-  [ -L /usr/local/local ] && sudo rm -f /usr/local/local
-fi
+# Delete ${CREW_PREFIX}/{var,local} symlink on some Chromium OS distro if exist
+[ -L ${CREW_PREFIX}/var ] && sudo rm -f #{CREW_PREFIX}/var
+[ -L ${CREW_PREFIX}/local ] && sudo rm -f #{CREW_PREFIX}/local
```
uberhacker, line 114: Probably should enclose in double quotes in case `#{CREW_PREFIX}` has a space.  Not that it ever should happen but it could potentially lead to the wrong files being deleted depending on the directory name.

install.sh
supechicken: It means 1 initial try, then 3 retries😅

install.sh
```diff
@@ -77,21 +120,25 @@ for dir in "${CREW_CONFIG_PATH}/meta" "${CREW_DEST_DIR}" "${CREW_PACKAGES_PATH}"
   fi
 done
 
+echo -e "\n${YELLOW}Downloading information for Bootstrap packages...${RESET}"
+echo -en "${GRAY}"
```
uberhacker, line 124: Is this line necessary?

install.sh
```diff
@@ -77,21 +120,25 @@ for dir in "${CREW_CONFIG_PATH}/meta" "${CREW_DEST_DIR}" "${CREW_PACKAGES_PATH}"
...
+# use parallel mode if available
+if [[ "$(curl --help curl)" =~ --parallel ]]; then
```
uberhacker, line 126: `$(${CURL}...`?

install.sh
```diff
@@ -136,7 +180,7 @@ function download_check () {
     fi
     #download
     echo -e "${BLUE}Downloading ${1}...${RESET}"
-    $CURL '-#' -L "${2}" -o "${3}"
+    curl '-#' -L "${2}" -o "${3}"
```
uberhacker, line 183: Why was this line changed?  Don't we want to use the environment variable instead?

install.sh
```diff
@@ -192,8 +236,8 @@ for i in $(seq 0 $((${#urls[@]} - 1))); do
   tarfile="$(basename ${url})"
   name="${tarfile%%-*}"   # extract string before first '-'
   rest="${tarfile#*-}"    # extract string after first '-'
-  version="$(echo ${rest} | sed -e 's/-chromeos.*$//')"
-                        # extract string between first '-' and "-chromeos"
+  version="${rest%%-chromeos*}"
```
uberhacker, line 239: Good call.

install.sh
supechicken: The `${CURL}` variable is already applied to the `curl` command in `function curl()`:
https://github.com/skycocker/chromebrew/blob/ecfc304e46e082c5f2490f097f5f2acf1e86977c/install.sh#L74-L79

install.sh
supechicken: See https://github.com/skycocker/chromebrew/pull/6519#discussion_r775123014

install.sh
supechicken: It is used to set the output color of `curl`

install.sh
supechicken: Changed

install.sh

install.sh
uberhacker: The initial try is still a retry. ;)

uberhacker: If we're defining a curl function here, why do we need the `CURL` environment variable?

uberhacker: OK, I think I understand what is going on here.  The `CURL` environment variable is used inside this function.

install.sh
```diff
@@ -104,18 +152,15 @@ if [ ! -f device.json ]; then
...
-  temp_sha256="$(echo "$pkgfile" | grep -m 3 "$ARCH": | tail -n 1 | cut -d\' -f2 | tr -d \' | tr -d \" | sed 's/,//g')"
-  urls[k]="$temp_url"
-  sha256s[k]="$temp_sha256"
-  k=$((k+1))
+  pkgfile="${CREW_PACKAGES_PATH}/${package}.rb"
+
+  [[ "$(sed -n '/binary_sha256/,/}/p' "${pkgfile}")" =~ .*x86_64:[[:blank:]]*[\'\"]([^\'\"]*) ]]
+    sha256s+=("${BASH_REMATCH[1]}")
+
+  [[ "$(sed -n '/binary_url/,/}/p' "${pkgfile}")" =~ .*x86_64:[[:blank:]]*[\'\"]([^\'\"]*) ]]
```
uberhacker, line 162: Shouldn't we abstract the arch with `$ARCH` instead of `x86_64`?

supechicken: Change made

PR #6520, opened 2021-12-25
downloader.rb: add `file://` protocol support, fallback to `curl` when the protocol is not `http(s)://`/`file://`

### LINKED ISSUE #6518: Unable to install local binaries
A package with a local binary cannot be installed.

```
  binary_url({                                                                                                                                       
    x86_64: 'file://...'                                        
  })
```
```
/usr/local/lib64/ruby/3.0.0/net/http.rb:987:in `initialize': Failed to open TCP connection to :80 (Connection refused - connect(2) for "" port 80) (Errno::ECONNREFUSED)
        from /usr/local/lib64/ruby/3.0.0/net/http.rb:987:in `open'
        from /usr/local/lib64/ruby/3.0.0/net/http.rb:987:in `block in connect'
        from /usr/local/lib64/ruby/3.0.0/timeout.rb:97:in `block in timeout'
        from /usr/local/lib64/ruby/3.0.0/timeout.rb:107:in `timeout'
        from /usr/local/lib64/ruby/3.0.0/net/http.rb:985:in `connect'
        from /usr/local/lib64/ruby/3.0.0/net/http.rb:970:in `do_start'
        from /usr/local/lib64/ruby/3.0.0/net/http.rb:959:in `start'
        from /usr/local/lib64/ruby/3.0.0/net/http.rb:621:in `start'
        from /usr/local/lib/crew/lib/downloader.rb:24:in `downloader'
```

supechicken:
Fixes #6518 

Copy from filesystem if the protocol in the given URL is `file://`
If the protocol is not `http(s)://` nor `file://`, use `curl`

Tested on `x86_64`

```
CREW_TESTING_REPO=https://github.com/supechicken666/chromebrew.git CREW_TESTING_BRANCH=downloader_file_protocol CREW_TESTING=1 crew update
```


lib/downloader.rb
```diff
@@ -21,6 +21,28 @@ def downloader (url, filename = File.basename(url), retry_count = 0, verbose = f
...
   uri = URI(url)
 
+  case uri.scheme
+  when /http?/
+  when 'file'
+    # use FileUtils to copy if it is a local file (the url protocol is file://)
+    if File.exist?(uri.path)
+      return FileUtils.cp uri.path, filename
+    else
+      abort "#{uri.path}: File not exist :/".lightred
```
uberhacker, line 31: `File not found` is better?

lib/downloader.rb
```diff
@@ -21,6 +21,28 @@ def downloader (url, filename = File.basename(url), retry_count = 0, verbose = f
   trap('WINCH') { setTermSize }
   uri = URI(url)
 
+  case uri.scheme
+  when /http?/
```
uberhacker, line 25: What do we do about `ftp` and other protocols?

saltedcoffii:
can we use the net library to support ftp:// URLs?

lib/downloader.rb
supechicken: Fallback to `curl`:

https://github.com/skycocker/chromebrew/pull/6520/files#diff-46c0af844c07a052d4a9dd139bf6e2d494a7032fed363960eaedaf4bc82c9e8dR33-R44

supechicken:
> can we use the net library to support ftp:// URLs?

Yes, there is a library called [`net/ftp`](https://ruby-doc.org/stdlib-2.4.0/libdoc/net/ftp/rdoc/Net/FTP.html), maybe we can implement it in the next PR

lib/downloader.rb
supechicken: Change made

lib/downloader.rb
uberhacker: It would be more elegant to use net/http for ftp.  Possible?

lib/const.rb
```diff
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.20.0'
+CREW_VERSION = '1.20.1'
```
uberhacker, line 3: Bump to `1.20.3`?

lib/downloader.rb
satmandu: Is https://ruby-doc.org/stdlib-2.7.0/libdoc/open-uri/rdoc/OpenURI.html another option? Since it wraps HTTP and FTP?

lib/const.rb

lib/downloader.rb
supechicken: It is a much better solution :)

PR #6524, opened 2021-12-26
Upgrade some gnu packages

saltedcoffii:
all need binaries. works on x86_64.

```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=update-and-fix-gnu CREW_TESTING=1 crew update
```

Updated packages are: `bison gnuchess libsigsegv mailutils sed texinfo`

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.20.0'
+CREW_VERSION = '1.20.1'
 
 ARCH_ACTUAL = `uname -m`.chomp
 # This helps with virtualized builds on aarch64 machines
@@ -122,8 +122,8 @@ end
 
 CREW_COMMON_FLAGS = '-O2 -pipe -flto -ffat-lto-objects -fPIC -fuse-ld=gold'
 CREW_COMMON_FNO_LTO_FLAGS = '-O2 -pipe -fno-lto -fPIC -fuse-ld=gold'
-CREW_FNO_LTO_LDFLAGS = '-fno-lto'
 CREW_LDFLAGS = '-flto'
+CREW_FNO_LTO_LDFLAGS = '-fno-lto'
 
 CREW_ENV_OPTIONS = <<~OPT.chomp
   CFLAGS='#{CREW_COMMON_FLAGS}' \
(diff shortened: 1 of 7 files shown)


packages/bison.rb
```diff
@@ -33,6 +20,7 @@ def self.install
   end
 
   def self.check
-    system 'make check'
+    puts 'Information: Bison checks take a very long time.'.lightblue
```
uberhacker, line 23: How about `Notice: ...'.orange`?

packages/libsigsegv.rb
```diff
@@ -3,32 +3,16 @@
...
   def self.build
-    system './configure',
-           "--prefix=#{CREW_PREFIX}",
-           "--libdir=#{CREW_LIB_PREFIX}",
-           '--enable-shared',
-           '--disable-static',
-           '--with-pic'
+    system 'autoreconf -fiv'
+    # libsigsegv fails to build with LTO.
+    system "#{CREW_ENV_FNO_LTO_OPTIONS} ./configure #{CREW_OPTIONS}"
```
uberhacker, line 15: Does it build without setting any environment options?

uberhacker:
Thank you for the updates @saltedcoffii!

packages/libsigsegv.rb
saltedcoffii: yes, but it would still be helpful to have `-O2` and `-pipe` and `-fPIC` and `-fuse-ld=gold`, etc.

packages/bison.rb
saltedcoffii: Isn't orange for warnings? What about green or yellow?

uberhacker: This sorta is like a warning.  Yellow would be better than green since that would indicate success.

saltedcoffii: yellow sounds good. Warning indicates something has gone wrong, this is just a notice.

saltedcoffii: 4ca78cc

packages/mailutils.rb
```diff
@@ -3,35 +3,25 @@
...
+    system "#{CREW_ENV_OPTIONS} ./configure #{CREW_OPTIONS} \
             --enable-ipv6 \
+            --with-gnutls \
+            --with-tcpwrappers \
             --with-berkeley-db \
             --without-guile \
             --with-gdbm \
+            --with-fribidi \
             --with-unistring \
             --with-tcp-wrappers \
```
uberhacker, line 26: 2 `--with-tcp(-)wrappers`?

lib/const.rb
```diff
@@ -122,8 +122,8 @@
 
 CREW_COMMON_FLAGS = '-O2 -pipe -flto -ffat-lto-objects -fPIC -fuse-ld=gold'
 CREW_COMMON_FNO_LTO_FLAGS = '-O2 -pipe -fno-lto -fPIC -fuse-ld=gold'
-CREW_FNO_LTO_LDFLAGS = '-fno-lto'
 CREW_LDFLAGS = '-flto'
+CREW_FNO_LTO_LDFLAGS = '-fno-lto'
```
satmandu, line 126: This is the most minor nitpick, but why move this?

saltedcoffii: It makes more sense. before it was
yes
no
no
yes

Now it's
yes
no
yes
no

I don't care, change it back if you like. it just looked better to me.

packages/mailutils.rb
saltedcoffii: looks like the correct one is `--with-tcp-wrappers`, thanks for catching that.

PR #6538, opened 2021-12-29
package.rb: Accept environment variables and options in `system` function

supechicken:
Duplicate of #6495 
Tested on `x86_64`

### Fixed issues
The patched `system` function will add `-j#{CREW_NPROC}` to the command name instead of passing a new parameter to `Kernel.system` when the command is passed as multi parameters (like `system 'make', 'install'`)

```
CREW_TESTING_REPO=https://github.com/supechicken666/chromebrew.git CREW_TESTING_BRANCH=package_rb CREW_TESTING=1 crew update
crew reinstall -s screen
```


satmandu (approved):
Appears to work and also builds`libcap` successfully.

lib/const.rb
```diff
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.20.0'
+CREW_VERSION = '1.20.1'
```
uberhacker, line 3: We cannot reuse `1.20.1`.  Bump to `1.20.2`?

uberhacker:
Thank you for the update @supechicken.

PR #6544, opened 2021-12-30
package.rb: Use a function to create function placeholders

supechicken:
Tested on `x86_64`

### Changes
- Create a new function in `lib/package_helpers.rb` for creating function placeholders (Line 8-17)
- Put all escape sequences into a regex (Line 180)

```
CREW_TESTING_REPO=https://github.com/supechicken666/chromebrew.git CREW_TESTING_BRANCH=crew_buildessential CREW_TESTING=1 crew update
crew reinstall -s screen
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.20.3'
+CREW_VERSION = '1.20.5'
 
 ARCH_ACTUAL = `uname -m`.chomp
 # This helps with virtualized builds on aarch64 machines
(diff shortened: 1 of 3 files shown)


lib/package.rb
```diff
@@ -167,9 +131,7 @@ def self.system(*args)
     # Add -j arg to build commands.
     cmd_args.sub(/\b(?<=make)(?=\b)/, " -j#{CREW_NPROC}") unless cmd_args =~ /-j\s*\d+/
     # Escape special bash characters.
-    cmd_args.gsub!('"','\\\\"')
-    cmd_args.gsub!('$','\\\\$')
-    cmd_args.gsub!('`','\\\\`')
+    cmd_args.gsub!(/("|\$|`)/, '\\\\\1')
```
uberhacker, line 134: Can we just come up with something better than escaping?  This is extremely hackish.

supechicken: Some possible solutions:
- Create an `IO` object based on the command and pass it to bash through stdin
- Pass the command as a separate parameter (`Kernel.system 'bash', '-c', '-e', @command`) so that `/bin/bash` will not be called before `bash`
- use heredoc:
```shell
system <<~BASH
  bash -e -c <<'CMD'
  #{@command}
  CMD
BASH
```

supechicken: @satmandu Which one do you like better?

satmandu: > @satmandu Which one do you like better?

I think the arguments to `bash -e -c` need to be passed in quotes, no?

lib/package.rb
```diff
@@ -5,6 +5,16 @@ class Package
            :binary_url, :binary_sha256, :source_url, :source_sha256,
            :git_branch, :git_hashtag, :is_fake
 
+  create_placeholder :preflight,   # Function for checks to see if install should occur.
+                     :patch,       # Function to perform patch operations prior to build from source.
+                     :prebuild,    # Function to perform pre-build operations prior to build from source.
+                     :build,       # Function to perform build from source.
```
uberhacker, line 11: Add `:postbuild`?

uberhacker:
@supechicken: Merge conflict with package.rb.

lib/package.rb
uberhacker: I was just thinking do we even need bash?  The directory this is run in has exec permissions so it could (should?) be able to run without bash, right?

satmandu:
Also, we need to fix ``` -j #{CREW_NPROC} ``` being added to lines with `make`. This is currently broken.

lib/package.rb
```diff
@@ -118,28 +119,22 @@ def self.system(*args)
...
     end
 
+    # after removing the env hash, all remaining args must be command args
+    cmd_args = args.join(' ')
+
     # Add -j arg to build commands.
-    cmd_args.sub(/\b(?<=make)(?=\b)/, " -j#{CREW_NPROC}") unless cmd_args =~ /-j\s*\d+/
-    # Escape special bash characters.
-    cmd_args.gsub!(/("|\$|`)/, '\\\\\1')
+    cmd_args.sub!(/\b(?<=make)(?=\b)/, " -j#{CREW_NPROC}") unless cmd_args =~ /-j\s*\d+/
```
satmandu, line 134: Please confirm that this works. :)

lib/const.rb
```diff
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.20.3'
+CREW_VERSION = '1.20.5'
```
uberhacker, line 3: Did you merge master into your working branch?  I'm thinking the old line should be `CREW_VERSION = '1.20.4'`.

lib/package.rb
```diff
@@ -118,28 +119,22 @@ def self.system(*args)
...
-    # extract Kernel.system options (if provided)
-    system_options = if args[-1].is_a?(Hash)
-      { exception: true }.merge(args[-1])
+    # extract env hash
+    if args[0].is_a?(Hash)
+      env = CREW_ENV_OPTIONS_HASH.merge(args[0])
+      args.delete_at(0) # remove env hash from args array
     else
-      { exception: true }
+      env = CREW_ENV_OPTIONS_HASH
```
satmandu, line 127: Can you add a check along the lines of `abort unless env.is_a?(Hash)` so we don't break what we're passing to Kernel.system in the future?

lib/const.rb
supechicken: The version is bumped

lib/package.rb
supechicken: I think the line 126 already handle it?

supechicken:
I have just run some tests on it:
```ruby
    system 'echo', 'make', 'install'
    system 'echo make install'
    system({'a' => '1'}, 'echo $a')
    system 'pwd', chdir: '/home'
    system({'a' => '1'}, 'echo $a; pwd', chdir: '/home')
```
Output:
```
make -j4 install
make -j4 install
1
/home
1
/home
```

Looks like it is working perfectly :)

lib/package.rb
supechicken: Done

lib/package.rb
supechicken: Actually `bash` is not necessary, but it allows us to use some extra expressions in `bash` like regex

lib/const.rb
supechicken: Merged just now (seems I misunderstood your meaning, never mind😅)

satmandu:
Have built gtk3 & gtk4 with this on multiple architectures.

PR #6560, opened 2022-01-06
add c89 and c99 executables to gcc11

saltedcoffii:
The POSIX Single UNIX Specification, Version 2 requires the c89 executable to compile ANSI C programs.
The POSIX Open Group Base Specifications, Issue 6 requires the c99 executable to compile C99 programs.

This pull request adds both executables to gcc11. I honestly don't think we need to recompile the binaries. We just need to extract the already existing binaries, put the scripts in their proper places with 0755 permissions and retarball and compress them.

```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=c89\&c99 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/gcc11.rb
+++ b/packages/gcc11.rb
@@ -10,6 +10,19 @@ class Gcc11 < Package
   source_url 'https://ftpmirror.gnu.org/gcc/gcc-11.2.0/gcc-11.2.0.tar.xz'
   source_sha256 'd08edc536b54c372a1010ff6619dd274c0f1603aa49212ba20f7aa2cda36fa8b'
 
+  binary_url ({
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/gcc11/11.2.0-1_armv7l/gcc11-11.2.0-1-chromeos-armv7l.tpxz',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/gcc11/11.2.0-1_armv7l/gcc11-11.2.0-1-chromeos-armv7l.tpxz',
+       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/gcc11/11.2.0-1_i686/gcc11-11.2.0-1-chromeos-i686.tpxz',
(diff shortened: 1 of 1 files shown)


satmandu:
I think @uberhacker just added functionality to add files properly from post-install...

Also, this is essentially the same as the file from debian, no?

satmandu:
Ubuntu c89:
```
#! /bin/sh

# Call the appropriate C compiler with options to accept ANSI/ISO C
# The following options are the same (as of gcc-2.95):
#       -ansi
#       -std=c89
#       -std=iso9899:1990

extra_flag=-std=c89

for i; do
    case "$i" in
        -ansi|-std=c89|-std=iso9899:1990)
            extra_flag=
            ;;
        -std=*)
            echo >&2 "`basename $0` called with non ANSI/ISO C option $i"
            exit 1
            ;;
    esac
done

exec gcc $extra_flag ${1+"$@"}
```

saltedcoffii:
I probably should have put it in the original message, but these scripts are from archlinux.

satmandu:
Can you put this in the postinstall so that we don't have to rebuild this? (And use the logic @uberhacker setup to add the files to the filelist.)

saltedcoffii:
I'll make binaries.

satmandu:
Also, FYI, a generic`crew` modification to allow uncompressing and recompressing packages would be useful too. Something along the lines of extracting a given package to destroot, and then repackaging it (and generating the appropriate filelist and dlist files too, and also running the steps to compress the contents.)  This way you could programmatically compress older packages already built with a different compression format, which might be nice for batch processing older package files.

saltedcoffii:
https://github.com/saltedcoffii/chromebrew/releases/tag/v0.4

satmandu:
Also, please don't put ampersands in your branch names. :)

saltedcoffii:
> Also, please don't put ampersands in your branch names. :)

Of course, why?

satmandu:
They are hard to make work with the command line since they need eacaping.

PR #6576, opened 2022-01-09
crew: add CREW_DISABLE_ENV_OPTION option

satmandu:
- `package.rb` currently automatically adds `CREW_ENV_OPTIONS` whereas we previously added it manually.
- With this modification we can set an env variable which sets the env variables passed with `make` to be just `CREW_DISABLE_ENV_OPTION=1`
- This also modifies the error message to include `CREW_ENV_OPTIONS` since that may have been involved in causing the error.

Works properly:
- [x] x86_64

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=CREW_NO_ADD_ENV_TO_MAKE CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -127,13 +127,18 @@ CREW_COMMON_FNO_LTO_FLAGS = '-O2 -pipe -fno-lto -fPIC -fuse-ld=gold'
 CREW_LDFLAGS = '-flto'
 CREW_FNO_LTO_LDFLAGS = '-fno-lto'
 
-CREW_ENV_OPTIONS_HASH = {
-  'CFLAGS'   => CREW_COMMON_FLAGS,
-  'CXXFLAGS' => CREW_COMMON_FLAGS,
-  'FCFLAGS'  => CREW_COMMON_FLAGS,
-  'FFLAGS'   => CREW_COMMON_FLAGS,
-  'LDFLAGS'  => CREW_LDFLAGS
-}
+CREW_DISABLE_ENV_OPTIONS = ENV['CREW_DISABLE_ENV_OPTIONS']
+unless CREW_DISABLE_ENV_OPTIONS
+  CREW_ENV_OPTIONS_HASH = {
+    'CFLAGS'   => CREW_COMMON_FLAGS,
+    'CXXFLAGS' => CREW_COMMON_FLAGS,
+    'FCFLAGS'  => CREW_COMMON_FLAGS,
+    'FFLAGS'   => CREW_COMMON_FLAGS,
+    'LDFLAGS'  => CREW_LDFLAGS
+  }
+else
+  CREW_ENV_OPTIONS_HASH = { "CREW_DISABLE_ENV_OPTIONS" => '1' }
+end
 # parse from hash to shell readable string
 CREW_ENV_OPTIONS = CREW_ENV_OPTIONS_HASH.map {|k, v| "#{k}=\"#{v}\"" } .join(' ')
 
@@ -224,5 +229,3 @@ CREW_LAST_PACKAGES = %w[ghc mandb gtk3 gtk4 sommelier]
 # libssp is in the libssp package
 # libatomic is in the gcc package
 CREW_ESSENTIAL_FILES = %x[LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/ruby].scan(/\t([^ ]+)/).flatten
-
-CREW_DISABLE_ENV_OPTION = ENV['CREW_DISABLE_ENV_OPTION']
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.21.1'
+CREW_VERSION = '1.21.2'
 
 ARCH_ACTUAL = `uname -m`.chomp
 # This helps with virtualized builds on aarch64 machines
@@ -195,7 +195,7 @@ CREW_CMAKE_OPTIONS = <<~OPT.chomp
   -DCMAKE_SHARED_LINKER_FLAGS='#{CREW_LDFLAGS}' \
   -DCMAKE_STATIC_LINKER_FLAGS='#{CREW_LDFLAGS}' \
   -DCMAKE_MODULE_LINKER_FLAGS='#{CREW_LDFLAGS}' \
-  -DCMAKE_INTERPROCEDURAL_OPTIMIZATION \
+  -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=TRUE \
   -DCMAKE_BUILD_TYPE=Release
 OPT
 CREW_CMAKE_FNO_LTO_OPTIONS = <<~OPT.chomp
@@ -224,3 +224,5 @@ CREW_LAST_PACKAGES = %w[ghc mandb gtk3 gtk4 sommelier]
 # libssp is in the libssp package
 # libatomic is in the gcc package
 CREW_ESSENTIAL_FILES = %x[LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/ruby].scan(/\t([^ ]+)/).flatten
+
+CREW_NO_ADD_ENV_TO_MAKE = ENV['CREW_NO_ADD_ENV_TO_MAKE']
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -225,4 +225,4 @@ CREW_LAST_PACKAGES = %w[ghc mandb gtk3 gtk4 sommelier]
 # libatomic is in the gcc package
 CREW_ESSENTIAL_FILES = %x[LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/ruby].scan(/\t([^ ]+)/).flatten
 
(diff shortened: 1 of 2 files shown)


lib/package.rb
```diff
@@ -129,6 +129,7 @@ def self.system(*args, **opt_args)
     else
       env = CREW_ENV_OPTIONS_HASH
     end
+    if CREW_NO_ADD_ENV_TO_MAKE then env = { "CREW_NO_ADD_ENV_TO_MAKE " => '1' } end
```
uberhacker, line 132: `env` is not a key word?  Also, how is this applied?  Don't we need an update to crew also?

satmandu: I can't speak to it being a key word, but I see `env` being used thus in `package.rb`:

```
# extract env hash
    if args[0].is_a?(Hash)
      env = CREW_ENV_OPTIONS_HASH.merge(args[0])
      args.delete_at(0) # remove env hash from args array
    else
      env = CREW_ENV_OPTIONS_HASH
    end
    if CREW_NO_ADD_ENV_TO_MAKE then env = { "CREW_NO_ADD_ENV_TO_MAKE " => '1' } end
```
and also here:
```
begin
      Kernel.system(env, cmd_args, **opt_args)
    rescue => e
      exitstatus = $?.exitstatus
      # print failed line number and error message
      puts "#{e.backtrace[1]}: #{e.message}".orange
      raise InstallError, "`#{env} #{cmd_args}` exited with #{exitstatus}"
    end
```

No update to crew is needed. We got rid of crew having to be updated for new constants some time back. Just `const.rb` needs to be updated.

I believe the way to have this applied is thus:
```
  def self.build
    ENV['CREW_NO_ADD_ENV_TO_MAKE'] = '1'
    warn_level = $VERBOSE
    $VERBOSE = nil
    load "#{CREW_LIB_PATH}lib/const.rb"
    $VERBOSE = warn_level
    FileUtils.mkdir('builddir')
    Dir.chdir('builddir') do
      system "#{@musldep_cmake_options} \
      -DOPENSSL_INCLUDE_DIR=#{CREW_MUSL_PREFIX}/include \
      -DCARES_STATIC=ON \
      -DCARES_STATIC_PIC=ON \
      -DCARES_BUILD_TESTS=ON \
      -DCARES_BUILD_CONTAINER_TESTS=ON \
      -DCARES_BUILD_TOOLS=ON \
      -DCARES_SHARED=OFF \
      ../ -G Ninja"
    end
    system 'samu -C builddir'
  end
```

or 
```
  def self.build
    ENV['CREW_NO_ADD_ENV_TO_MAKE'] = '1'
    warn_level = $VERBOSE
    $VERBOSE = nil
    load "#{CREW_LIB_PATH}lib/const.rb"
    $VERBOSE = warn_level
    system 'make'
  end
```

uberhacker: Am I missing something?  I don't see where this takes effect if set.

satmandu: Try a package file with 
```
def self.build
    ENV['CREW_NO_ADD_ENV_TO_MAKE'] = '1'
    warn_level = $VERBOSE
    $VERBOSE = nil
    load "#{CREW_LIB_PATH}lib/const.rb"
    $VERBOSE = warn_level
    system 'exit 1'
  end
```

You should see only `CREW_NO_ADD_ENV_TO_MAKE=1` being set in the error output,instead of having all of `CREW_ENV_OPTIONS` in there...

satmandu: Line 132 is where `env` is redefined to be a nop if the env variable exists.

satmandu:
```
crew build test
test: Test Package.
Downloading source...
/usr/local/lib/crew/lib/package.rb:141:in `system': Command failed with exit 1: exit 1
test failed to build: `{"CREW_NO_ADD_ENV_TO_MAKE "=>"1"} exit 1` exited with 1
```

satmandu: test package: `test.rb`
```
require 'package'

class Test < Package
  description 'Test Package.'
  homepage 'http://chromebrew'
  @_ver = '1.0'
  version @_ver
  license 'BSD and GPL-2'
  compatibility 'all'
  source_url 'SKIP'


  def self.build
   ENV['CREW_NO_ADD_ENV_TO_MAKE'] = '1'
    warn_level = $VERBOSE
    $VERBOSE = nil
    load "#{CREW_LIB_PATH}lib/const.rb"
    $VERBOSE = warn_level
    system "exit 1"
  end

  def self.check
    ENV['CREW_NO_ADD_ENV_TO_MAKE'] = '1'
    warn_level = $VERBOSE
    $VERBOSE = nil
    load "#{CREW_LIB_PATH}lib/const.rb"
    $VERBOSE = warn_level
    system 'make', 'test'
  end

  def self.install
    ENV['CREW_NO_ADD_ENV_TO_MAKE'] = '1'
    warn_level = $VERBOSE
    $VERBOSE = nil
    load "#{CREW_LIB_PATH}lib/const.rb"
    $VERBOSE = warn_level
  end
end
```

satmandu: I'm not printing out the hash in the error message command line output, but open to suggestions on how to do that best...

supechicken:
Would `CREW_DISABLE_ENV_OPTION` better?

「Satadru Pramanik ***@***.***>」:

satmandu:
> Would `CREW_DISABLE_ENV_OPTION` better?

Yes!

lib/package.rb
uberhacker: Oh, so you are just overriding the `env` value?  I'm thinking it might be better to just not set `env` in the first place.  That would require a change to `crew`, however.

satmandu: I was trying to have a light touch. `env` needs to be a hash, and it's being used for every `system` line via that mechanism at the end of `package.rb`.  So it seemed best to just set the hash to something minimal which indicated what had been done.

satmandu: But if you think that it is better to edit `crew` feel free to offer alternative code?

uberhacker: It would be easier to understand what is going on if the logic was all in one place.  ie. Add a conditional to set based on the const value.

satmandu: Where would you put it though in crew? `system` could be called from anywhere.

Maybe we could add a comment line right above `if CREW_NO_ADD_ENV_TO_MAKE then env = { "CREW_NO_ADD_ENV_TO_MAKE " => '1' } end`?

supechicken:
> Would `CREW_DISABLE_ENV_OPTION` better?

Sorry, should be `CREW_DISABLE_ENV_OPTIONS`😅

lib/const.rb
```diff
@@ -224,3 +224,5 @@
 # libssp is in the libssp package
 # libatomic is in the gcc package
 CREW_ESSENTIAL_FILES = %x[LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/ruby].scan(/\t([^ ]+)/).flatten
+
+CREW_DISABLE_ENV_OPTION = ENV['CREW_DISABLE_ENV_OPTION']
```
supechicken, line 228: Combine with `CREW_ENV_OPTIONS` to make these changes globally (some packages are using `CREW_ENV_OPTIONS` in `./configure`)?
```ruby
unless CREW_DISABLE_ENV_OPTIONS
  CREW_ENV_OPTIONS_HASH = ...
else
  CREW_ENV_OPTIONS_HASH = { "CREW_DISABLE_ENV_OPTION" => '1' }
end
```

supechicken: (by that way we don't have to modify `self.system`)

lib/package.rb
uberhacker: Sorry, I said `crew` earlier but I should have said `const.rb`.
```
CREW_DISABLE_ENV_OPTIONS = ENV['CREW_DISABLE_ENV_OPTIONS']
CREW_ENV_OPTIONS_HASH = {                                                                                                                                                                 
  'CFLAGS'   => CREW_COMMON_FLAGS,                                                                                                                                                        
  'CXXFLAGS' => CREW_COMMON_FLAGS,                                                                                                                                                        
  'FCFLAGS'  => CREW_COMMON_FLAGS,                                                                                                                                                        
  'FFLAGS'   => CREW_COMMON_FLAGS,                                                                                                                                                        
  'LDFLAGS'  => CREW_LDFLAGS                                                                                                                                                              
}                                                                                                                                                                                         
# parse from hash to shell readable string                                                                                                                                                
CREW_ENV_OPTIONS = CREW_DISABLE_ENV_OPTIONS ? 'CREW_DISABLE_ENV_OPTIONS=1' : CREW_ENV_OPTIONS_HASH.map {|k, v| "#{k}=\"#{v}\"" } .join(' ')
```

lib/package.rb
```diff
@@ -129,6 +129,7 @@ def self.system(*args, **opt_args)
     else
       env = CREW_ENV_OPTIONS_HASH
```
uberhacker, line 130: Should this line be `env = CREW_ENV_OPTIONS`?  No `_HASH`?

satmandu: @supechicken ?

satmandu: no. env needs to be a hash. That's the format that system takes. See https://ruby-doc.org/core-3.0.2/Kernel.html#method-i-system

lib/const.rb
satmandu: Brilliant. Yes. Done.

PR #6610, opened 2022-01-11
libsndfile => 1.0.28-1

saltedcoffii:
Works on x86_64. needs binaries. depends on #6581 , #6579 and #6583

```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=libsndfile_1.0.28-1 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/libsndfile.rb
+++ b/packages/libsndfile.rb
@@ -3,34 +3,31 @@ require 'package'
 class Libsndfile < Package
   description 'Libsndfile is a C library for reading and writing files containing sampled sound (such as MS Windows WAV and the Apple/SGI AIFF format) through one standard library interface.'
   homepage 'http://www.mega-nerd.com/libsndfile/'
-  version '1.0.28'
+  version '1.0.28-1'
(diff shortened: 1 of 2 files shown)


packages/libsndfile.rb
```diff
@@ -3,34 +3,31 @@
...
-     x86_64: '65b4f4f13f020620801e43bbfe9027fc7208d18575727bda56d2ebe75c80e7fc',
-  })
+  depends_on 'libvorbis'
+  depends_on 'speex'
+  depends_on 'libogg'
+  depends_on 'opus'
+  depends_on 'flac'
+  depends_on 'sqlite'
+  depends_on 'alsa_lib'
+  depends_on 'nasm'
```
uberhacker, line 19: Interesting.  We were missing all these dependencies in the previous version?

saltedcoffii: it seems we were yes

PR #6617, opened 2022-01-11
new package libmtp

saltedcoffii:
"libmtp is a library to access MTP (Media Transfer Protocol) Devices."

```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=libmtp CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/libmtp.rb
+++ b/packages/libmtp.rb
@@ -1,0 +1,35 @@
+require 'package'
+
+class Libmtp < Package
+  description 'libmtp is a library to access MTP (Media Transfer Protocol) Devices.'
+  homepage 'https://github.com/libmtp/libmtp/'
+  @_ver = '1.1.19'
+  version @_ver
+  license 'LGPL-2.1'
+  compatibility 'all'
+  source_url 'https://github.com/libmtp/libmtp.git'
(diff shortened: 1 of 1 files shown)


packages/libmtp.rb
```diff
@@ -0,0 +1,35 @@
...
+  compatibility 'all'
+  source_url 'https://github.com/libmtp/libmtp.git'
+  git_hashtag 'v' + @_ver
+
+  depends_on 'libusb'
+  depends_on 'eudev'
+
+  def self.patch
+    system "sed -e 's:read IN:IN=y:' \
+                -e 's:wget -O:curl -#Lo:g' \
```
uberhacker, line 18: I think you want `-#LO` and not `-#Lo`.  Capital 'O'.

saltedcoffii: I want a lowercase o here because -O writes the filename from the server and -o allows you to specify your own filename. As does -O from wget.

uberhacker: OK, I didn't realize you were trying to write to a file.  It's a bit confusing because `-O` and `-o` almost mean the exact opposite between curl and wget. :)

PR #6618, opened 2022-01-11
libnfs => 4.0.0-1

saltedcoffii:

```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=libnfs_4.0.0-1 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/libnfs.rb
+++ b/packages/libnfs.rb
@@ -1,40 +1,27 @@
-# Adapted from Arch Linux libnfs PKGBUILD at:
-# https://github.com/archlinux/svntogit-community/raw/packages/libnfs/trunk/PKGBUILD
-
 require 'package'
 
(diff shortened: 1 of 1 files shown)


packages/libnfs.rb
```diff
@@ -1,40 +1,27 @@
-# Adapted from Arch Linux libnfs PKGBUILD at:
-# https://github.com/archlinux/svntogit-community/raw/packages/libnfs/trunk/PKGBUILD
-
```
uberhacker, line 3: Why was this note removed?  It might be helpful to keep this.

saltedcoffii: because i rewrote the file so much that it no longer resembles that PKGBUILD file.

PR #6620, opened 2022-01-11
mpg123 => 1.29.3

saltedcoffii:

```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=mpg123_1.29.3 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/mpg123.rb
+++ b/packages/mpg123.rb
@@ -1,39 +1,28 @@
 require 'package'
 
 class Mpg123 < Package
-  description 'Fast console MPEG Audio Player and decoder library'
+  description 'Fast console MPEG Audio Player and decoder library.'
   homepage 'http://www.mpg123.org/'
-  version '1.25.10'
+  version '1.29.3'
   license 'GPL-2 and LGPL-2.1'
   compatibility 'all'
-  source_url 'https://downloads.sourceforge.net/project/mpg123/mpg123/1.25.10/mpg123-1.25.10.tar.bz2'
-  source_sha256 '6c1337aee2e4bf993299851c70b7db11faec785303cfca3a5c3eb5f329ba7023'
+  source_url 'https://sourceforge.net/projects/mpg123/files/mpg123/1.29.3/mpg123-1.29.3.tar.bz2'
+  source_sha256 '963885d8cc77262f28b77187c7d189e32195e64244de2530b798ddf32183e847'
 
-  binary_url ({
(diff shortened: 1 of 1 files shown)


packages/mpg123.rb
```diff
@@ -1,39 +1,28 @@
...
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/mpg123/1.25.10_x86_64/mpg123-1.25.10-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: '25786bf03b27a76815ab30e8f3c8e864e7f2f58548b84e092e93f51a074e2515',
-     armv7l: '25786bf03b27a76815ab30e8f3c8e864e7f2f58548b84e092e93f51a074e2515',
-       i686: 'efe58119727f1d513309661bc643250ce3dff4dfd313a7e8e0b8b3bbdba00baf',
-     x86_64: '28edfc1ab64e472a915f6561cb810ed4f69a6ec06d38d7209cef400b823d9328',
-  })
-
-  depends_on 'alsa_plugins'
```
uberhacker, line 25: Why were alsa plugins removed?

saltedcoffii: mpg123 doesn't link against anything from alsa_plugins, only things from alsa_lib

packages/mpg123.rb
```diff
@@ -1,39 +1,28 @@
...
-  depends_on 'cras'
+  depends_on 'alsa_lib'
+#  depends_on 'cras'
 
   def self.build
-    system './configure',
-           "--prefix=#{CREW_PREFIX}",
-           "--libdir=#{CREW_LIB_PREFIX}",
-           '--with-default-audio=oss'
+    system "#{CREW_ENV_OPTIONS} ./configure #{CREW_OPTIONS} \
```
supechicken, line 16: We don't need to add `CREW_ENV_OPTIONS` to the command now (it will be inserted by package.rb automatically now)

uberhacker: If `CREW_ENV_OPTIONS` is added does only one set of config options get added?

supechicken: The newly passed `CREW_ENV_OPTIONS` in the command will cover it (kind of redefine) (so there is always one CFLAGS only)

PR #6621, opened 2022-01-12
install.sh: Add rsync as a core package

### LINKED ISSUE #6516: [crew]: `rsync: ACLs are not supported on this client`
This error was reported when installing any package:
```
rsync: ACLs are not supported on this client
rsync error: syntax or usage error (code 1) at main.c(1746) [client=3.2.3]
```

supechicken:
Fixes #6516 #6633

The version of `rsync` provided by CloudReady is compiled without ACL support, but the `-A` param we are using, requires ACL support, thus an error will be thrown when using the system provided `rsync`
```
rsync: ACLs are not supported on this client
rsync error: syntax or usage error (code 1) at main.c(1746) [client=3.2.3]
```

### Changes
- Check if the available `rsync` command supports ACLs before using it

```
CREW_TESTING_REPO=https://github.com/supechicken666/chromebrew.git CREW_TESTING_BRANCH=cloudready_fix CREW_TESTING=1 crew update
```


bin/crew
```diff
@@ -1126,24 +1126,27 @@ def install_package(pkgdir)
         end
       end
     end
-    if File.exist?("#{CREW_PREFIX}/bin/rdfind")
-        puts "Using rdfind to convert duplicate files to hard links."
-        system "#{CREW_PREFIX}/bin/rdfind -removeidentinode true -makesymlinks false -makehardlinks true -makeresultsfile false ."
+    if system 'command -v rdfind', out: '/dev/null'
```
uberhacker, line 1129: Why not Rubyize?

bin/crew
```diff
@@ -1126,24 +1126,27 @@ def install_package(pkgdir)
...
+        puts 'Using rdfind to convert duplicate files to hard links.'
+        system 'rdfind -removeidentinode true -makesymlinks false -makehardlinks true -makeresultsfile false .'
     end
+
+    # check if the available rsync command support ACLs
+    rsync_available = system 'command -v rsync && [ "$(rsync -V)" != *"no ACLs"* ]', out: '/dev/null'
```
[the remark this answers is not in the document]

bin/crew
supechicken: Oh, I think you mean using the previous `File.exist?` one. `command -v` can check in the whole $PATH, why not :)

uberhacker: No Ruby way to do the same thing?  Is rsync available oob?  If not, the only place it would be located is at `#{CREW_PREFIX}/bin`, right?

supechicken: The reason for search it in whole PATH is to let users can specify and use their executable in other locations

satmandu:
Is there a good way to check to see if a system is Cloudready?
If we can add a check to `const.rb`, we can add `rsync` to `core.rb` with an `if cloudready? ` sort of check, no?

That way we fix this issue during install?

supechicken:
> Is there a good way to check to see if a system is Cloudready? If we can add a check to `const.rb`, we can add `rsync` to `core.rb` with an `if cloudready? ` sort of check, no?
> 
> That way we fix this issue during install?

But any installing action will call `rsync`...

satmandu:
> But every installing action will call `rsync`... (also during `rsync` installation)
For sure. I mean in addition to your changes.  We should also add the logic to `install.sh` since this issue will crop up during installs, right? Ideally, we address the problem automatically then.

install.sh
```diff
@@ -24,7 +24,7 @@ CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 ARCH="${ARCH/armv8l/armv7l}"
 
 # BOOTSTRAP_PACKAGES cannot depend on crew_profile_base for their core operations (completion scripts are fine)
-BOOTSTRAP_PACKAGES="pixz jq ca_certificates git gmp ncurses libyaml ruby openssl"
```
satmandu, line 27: Why remove `jq`? jq is used in `install.sh` (I think the `jq` in most installs is much older.)

bin/crew
uberhacker: If users are relocating executables, they should be using `CREW_PREFIX` to do it.

install.sh
supechicken: The preinstalled version works well with our `jq` arguments in `install.sh`

bin/crew
supechicken: Change reverted

bin/crew
```diff
@@ -1127,24 +1127,35 @@ def install_package(pkgdir)
...
+    rsync_available = File.exist?("#{CREW_PREFIX}/bin/rsync")
+    puts 'rsync is not working. Please install it with \'crew install rsync\''.lightred unless rsync_available
+
+    if Dir.exists? "#{pkgdir}/#{HOME}" then
+      if rsync_available
         system "rsync -ahHAXW --remove-source-files ./#{HOME.delete_prefix('/')}/ #{HOME}"
       else
-        puts 'rsync is not working. Please install it with \'crew install rsync\''.lightred if @opt_verbose
-        system "tar -c#{@verbose}f - ./usr/* ./home/* | (cd /; tar xp --keep-directory-symlink -f -)"
+        system "cp -a#{@verbose} ./#{HOME}/* #{HOME}"
```
uberhacker, line 1142: I'm VERY hesitant to change line 1138.  What happened to `./usr/*`?

bin/crew
```diff
@@ -1127,24 +1127,35 @@ def install_package(pkgdir)
...
     end
     if Dir.exists? "#{pkgdir}/usr/local" then
-      if system("#{CREW_PREFIX}/bin/rsync -V > /dev/null 2>&1")
+      if rsync_available
         # Adjust "./usr/local" if the build CREW_PREFIX ever changes.
         system "rsync -ahHAXWx --remove-source-files ./usr/local/ #{CREW_PREFIX}"
       else
-        puts 'rsync is not working. Please install it with \'crew install rsync\''.lightred if @opt_verbose
-        system "tar -c#{@verbose}f - ./usr/* | (cd /; tar xp --keep-directory-symlink -f -)"
+        system "cp -a#{@verbose} ./usr/local/* #{CREW_PREFIX}"
```
uberhacker, line 1150: I'm not sure about this change.

bin/crew
uberhacker: Oh, I see now.  The `./usr/*` files are handled below.

supechicken: The reason of `tar` was replaced by `cp` is `cp` is more suitable for copying files, the `-a` param is used to tell `cp` to perform like `tar` (keep owner/exec/group bit and don't touch symlinks)

install.sh
```diff
@@ -213,7 +213,7 @@ function extract_install () {
       LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} tar -Ipixz -xpf ../"${2}"
     fi
     echo -e "${BLUE}Installing ${1} ...${RESET}"
-    tar cpf - ./*/* | (cd /; tar xp --keep-directory-symlink -f -)
+    cp -a ./usr/local/* "${CREW_PREFIX}"
```
satmandu, line 216: Why make this change? Tar is already on the system?

supechicken: [See above](https://github.com/skycocker/chromebrew/pull/6621#discussion_r784531492_)

satmandu: At the minimum, does `cp` preserve directory symlinks like the tar usage does?

supechicken: Yes, the `-a` option tell cp to preserve everything

bin/crew
satmandu: Apparently, tar is much faster than cp. See https://superuser.com/a/985080/1646144

But maybe that's old information?

Could you do some testing by seeing how long installing something large like gcc11 or llvm takes with the existing `tar` setup vs the `cp -a` setup you are suggesting?  I have no problem switching especially if it is faster.

uberhacker: I imagine `mv` would be the fastest and most space efficient. :)

satmandu: Not necessarily! rsync can be faster...

satmandu:
As per https://github.com/skycocker/chromebrew/issues/6633 It looks like it isn't just cloudready which has this rsync issue.

bin/crew
satmandu: I'd be much more comfortable with the change from `tar` if an install with this were tested first.

Until that happens, I'd be ok with merging this with just the `rsync_available` changes and the fallback being the use of tar in both `crew` and `install.sh`.

supechicken: You are right,  `tar` is much more faster than `cp -a`, switching back to `tar`

satmandu:
Are you switching `cp` back to `tar` in `crew` too?

PR #6623, opened 2022-01-13
Patchelf -> 0.14.3

satmandu:
Works properly:
- [x] x86_64
- [x] armv7;
- [x] i686
```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=patchelf CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/patchelf.rb
+++ b/packages/patchelf.rb
@@ -3,33 +3,28 @@ require 'package'
 class Patchelf < Package
   description 'PatchELF is a small utility to modify the dynamic linker and RPATH of ELF executables.'
   homepage 'http://nixos.org/patchelf.html'
-  version '0.13'
+  version '0.14.3'
   license 'GPL-3'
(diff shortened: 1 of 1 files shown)


saltedcoffii:
Since we're linking this statically, should we be using musl?

satmandu:
> Since we're linking this statically, should we be using musl?

We don't need to, so it appears. (But good thinking on this... a bunch of recent commits this past month to the patchelf repo indicate tweaks for building with musl!)

saltedcoffii:
why don't we need to?

satmandu:
`patchelf` as built here is already static. I think it's the more complicated programs which have huge problems with building statically with glibc.

satmandu:
(I think generally if you need to pull in other non-libc libraries, then you need to use musl to make a static build.)

PR #6634, opened 2022-01-19
make env variables used by crew mostly booleans

satmandu:
- Currently variables in `const.rb` set from env variables are treated as both booleans and strings inside `crew`.
- empty env variables are still leading to the crew variables being set as true, which is a problem.
- This change converts most of these variables to booleans, such that an empty variable is treated as false.
- Parsing of these variables was also fixed in `crew`, and relevant packages were also updated.

Works properly:
- [x] x86_64

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=fix_la_rename_enabled CREW_TESTING=1 crew update
```


uberhacker:
Good call!

PR #6640, opened 2022-01-20
Add git_fetchtags, simplify (.nil? or .empty?) to .to_s.empty?

satmandu:
- We have been fetching git tags because some packages require git tags to pull version information.
- However, doing this massively increases the size of the download and resultant cached file (if caching is enabled.) (@saltedcoffii pointed this out.)
- Instead, mark packages which need git tags with `git_fetchtags` and only fetch git tags for such packages.
- The `py3_numpy` package requires git tags, and has been so marked.

- ALSO, replaced instances of `.nil? or .empty?` with `.to_s.empty?` which handles this more simply.

Works properly:
- [x] x86_64

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=crew_cache CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -729,7 +729,7 @@ def download
           cachefile = CREW_CACHE_DIR + filename + @pkg.git_hashtag.gsub('/', '_') + '.tar.xz'
           puts "cachefile is #{cachefile}".orange if @opt_verbose
         # Git branch and git commit specified
-        elsif ! @pkg.git_hashtag.to_s.empty? and ! @pkg.git_branch.to_s.empty?
+        elsif ! @pkg.git_hashtag.to_s.empty?
           cachefile = CREW_CACHE_DIR + filename + @pkg.git_branch.gsub(/[^0-9A-Za-z.\-]/, '_') + '_' + @pkg.git_hashtag.gsub('/', '_') + '.tar.xz'
           puts "cachefile is #{cachefile}".orange if @opt_verbose
         # Git branch specified, without a specific git commit.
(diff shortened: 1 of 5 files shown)


bin/crew
```diff
@@ -724,12 +724,12 @@ def download
...
         # No git branch specified, just a git commit or tag
-        if @pkg.git_branch.nil? || @pkg.git_branch.empty?
-          abort("No Git branch, commit, or tag specified!").lightred if @pkg.git_hashtag.nil? || @pkg.git_hashtag.empty?
+        if @pkg.git_branch.to_s.empty?
+          abort("No Git branch, commit, or tag specified!").lightred if @pkg.git_hashtag.to_s.empty?
           cachefile = CREW_CACHE_DIR + filename + @pkg.git_hashtag.gsub('/', '_') + '.tar.xz'
           puts "cachefile is #{cachefile}".orange if @opt_verbose
         # Git branch and git commit specified
-        elsif ( ! @pkg.git_hashtag.nil? || ! @pkg.git_hashtag.empty )  and ( ! @pkg.git_branch.nil? || ! @pkg.git_branch.empty?)
+        elsif ! @pkg.git_hashtag.to_s.empty? and ! @pkg.git_branch.to_s.empty?
```
supechicken, line 732: `@pkg.git_branch.to_s.empty?` already checked above?

satmandu: That's looking for `! @pkg.git_branch.to_s.empty?`, so the reverse condition...

PR #6642, opened 2022-01-21
crew: Replace `binutils/ar` with a new library

supechicken:
The only usage of including `binutils` in `core` is to provide `ar` for `crew` (all linker libraries are not necessary for core)

This PR adds a new library that handles all `.deb` archive files, compared to `ar`, regex can be used to specify the name of the file to extract.

### Changes
- Remove `binutils` from `core.rb`
- Add `lib/deb_utils.rb`
- Replace `ar` with the new library

Tested with `opera`

```
CREW_TESTING_REPO=https://github.com/supechicken666/chromebrew.git CREW_TESTING_BRANCH=ar_replace CREW_TESTING=1 crew update
```


lib/deb_utils.rb
```diff
@@ -0,0 +1,43 @@
...
+    elsif target
+      next
+    end
+
+    File.open(name, 'wb') do |dst_fileIO|
+      # write to filesystem
+      dst_fileIO.write(fileContent)
+    end
+  end
+  abort "#{target}: File not found in archive :/" if target and !file_found
```
uberhacker, line 42: Simplify `if target and !file_found` with `unless file_found`?

lib/deb_utils.rb
```diff
@@ -0,0 +1,43 @@
...
+  # extract_deb: unarchive .deb files
+  # Usage: extract_deb(<file>, <target (optional)>)
+  #   file: a .deb archive file
+  #   target: (optional) a specified file to extract from archive, should be passed in regex
+  #
+  # Example:
+  #   extract_deb('example.deb', /data\..*/) # extract files from example.deb with filenames matching the /data\..*/ regex
+  #   extract_deb('example.deb')             # extract all files in example.deb
+  #
+  file_found = false if target
```
uberhacker, line 11: Simply `file_found = false` is good enough?

lib/deb_utils.rb
```diff
@@ -0,0 +1,43 @@
...
+  src_fileIO = File.open(file, 'rb')
+  # get first line of the given file, should be a signature string (`!<arch>\n`) if it is a valid deb file
+  signature = src_fileIO.gets
+
+  abort 'Malformed archive :/' unless signature == "!<arch>\n"
```
uberhacker, line 16: This seems like an overly simplistic way to determine if an archive is malformed.

lib/deb_utils.rb
uberhacker: OK, nevermind.  I see what you are trying to do here.

lib/deb_utils.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    # check ending byte
+    abort "Malformed archive :/" unless end_char == '`'
+
+    # capture file in archive with given offset bytes (file size)
+    fileContent = src_fileIO.read(size.to_i)
+
+    # filter filename if a target file is specified
+    if target and name =~ target
+      file_found = true
```
uberhacker, line 32: This could be more efficient if there was a break if target was not using a wildcard regex expression.

lib/deb_utils.rb
```diff
@@ -0,0 +1,43 @@
...
+  abort 'Malformed archive :/' unless signature == "!<arch>\n"
+
+  # process each file in archive
+  while (line = src_fileIO.gets) do
+    # read file meta
+    name, modtime, uid, gid, mode, size, end_char = \
+      line.scan(/(.{16})(.{12})(.{6})(.{6})(.{8})(.{10})(.{1})/).flatten.map(&:strip)
+
+    # check ending byte
+    abort "Malformed archive :/" unless end_char == '`'
```
uberhacker, line 25: This is good enough to determine if an archive is malformed?

lib/deb_utils.rb
uberhacker: Better message?
```
abort "#{target} not found in archive. :/".lightred
```

supechicken: Change made

lib/deb_utils.rb
supechicken: An `ar` archive is formed by readable texts only (except data stored in the archive, like `.gz` files), the ``` symbol will added to the end of every file metadata line, see https://en.wikipedia.org/wiki/Ar_(Unix)

lib/deb_utils.rb
supechicken: This extracts all files that match the regex instead of one of the matched file

lib/deb_utils.rb
supechicken: See below

lib/deb_utils.rb
uberhacker: I get that but if only one file is requested without any regex wildcards to look for multiple files, it makes sense to break when found rather than looping through all the files.

satmandu:
Do we care about the scripts which often come with Deb's? Such as pre-install and post-install scripts?

lib/deb_utils.rb
supechicken: Since I can't find a good way that checks whatever the passed regex contains wildcards, added the support of using strings as the target now (when `target` is `String`, exit function when file found)

supechicken:
> Do we care about the scripts which often come with Deb's? Such as pre-install and post-install scripts?

Functions for patching and running those scripts can be added in the next PR :)

lib/deb_utils.rb
```diff
@@ -0,0 +1,49 @@
...
+      # capture file in archive with given offset bytes (file size)
+      fileContent = src_fileIO.read(size.to_i)
+
+      # filter filename if a target file is specified
+      if target.is_a?(String) and name == target
+        # if target is passed as string, write matched file to filesyetem and exit function
+        # write to filesystem
+        File.open(name, 'wb') {|dst_fileIO| dst_fileIO.write(fileContent) }
+        # exit function
+        return true
```
uberhacker, line 38: Don't you also want `file_found = true` here too?

PR #6644, opened 2022-01-22
Move all runtime options to `lib/runtime_var.rb`

supechicken:
The begin of refactoring crew :)

Tested on `x86_64`

### Changes
- Add `lib/function.rb` (see `lib/function.rb` for more info)
- Move all runtime options to `lib/runtime_var.rb`
- All runtime options were wrapped into a global variable `$OPT`, which means we can overwrite them from packages without reloading `const.rb`

```
CREW_TESTING_REPO=https://github.com/supechicken666/chromebrew.git CREW_TESTING_BRANCH=runtime_var CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -9,6 +9,7 @@ require 'digest/sha2'
 require 'json'
 require 'fileutils'
 require 'tmpdir'
+require_relative '../lib/function'
 require_relative '../lib/const'
 require_relative '../lib/util'
 require_relative '../lib/convert_size'
@@ -104,9 +105,9 @@ end
 
 String.use_color = args["--color"] || !args["--no-color"]
 @opt_keep = args["--keep"]
-@opt_verbose = args["--verbose"]
+Verbose = args["--verbose"]
 
-if @opt_verbose then
+if Verbose then
   @fileutils_verbose = true
   @verbose = 'v'
   @short_verbose = '-v'
@@ -193,7 +194,7 @@ def list_available
 end
 
 def list_installed
-  unless @opt_verbose
+  unless Verbose
     Dir[CREW_META_PATH + '*.directorylist'].sort.map do |f|
       File.basename(f, '.directorylist').lightgreen
     end
@@ -232,7 +233,7 @@ def list_compatible(compat = true)
 end
 
 def generate_compatible
-  puts 'Generating compatible packages...'.orange if @opt_verbose
+  verbose_puts 'Generating compatible packages...'.orange
   @device[:compatible_packages] = []
   Dir[CREW_PACKAGES_PATH + '*.rb'].each do |filename|
     pkgName = File.basename filename, '.rb'
@@ -241,7 +242,7 @@ def generate_compatible
     rescue => e
       puts "Error with #{pkgName}.rb: #{e}".red unless e.to_s.include?('uninitialized constant')
     end
-    puts "Checking #{pkgName} for compatibility.".orange if @opt_verbose
+    verbose_puts "Checking #{pkgName} for compatibility.".orange
     # If compatibility property does not exist, check if a binary package
     # exists, and if not, see if at least a source url exists.
     @compatibility = true
@@ -255,22 +256,22 @@ def generate_compatible
         #puts "url: #{@url}".green
         # If no source package is available, then package is not compatible.
         @compatibility = false unless @url
-        puts "#{pkgName} compatibility is #{@compatibility}" if @opt_verbose
+        verbose_puts "#{pkgName} compatibility is #{@compatibility}"
       end
     end
     if ( @pkg.compatibility&.include? 'all' or @pkg.compatibility&.include? ARCH or @pkg.compatibility.nil? ) and @compatibility
       #add to compatible packages
-      puts "Adding #{pkgName} to compatible packages.".lightgreen if @opt_verbose
+      verbose_puts "Adding #{pkgName} to compatible packages.".lightgreen
       @device[:compatible_packages].push(name: @pkg.name)
     else
-      puts "#{pkgName} is not a compatible package.".lightred if @opt_verbose
+      verbose_puts "#{pkgName} is not a compatible package.".lightred
     end
   end
   File.open(CREW_CONFIG_PATH + 'device.json', 'w') do |file|
     output = JSON.parse @device.to_json
     file.write JSON.pretty_generate(output)
   end
-  puts 'Generating compatible packages done.'.orange if @opt_verbose
+  verbose_puts 'Generating compatible packages done.'.orange
 end
 
 def search (pkgName, silent = false)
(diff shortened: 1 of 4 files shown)


satmandu:
Can you explain usage with some examples?

For instance, how would one replace the usage of `ENV['CREW_DISABLE_ENV_OPTIONS'] = '1'`?

And how would this change using env variables such as in `CREW_TESTING_REPO=https://github.com/supechicken666/chromebrew.git CREW_TESTING_BRANCH=runtime_var CREW_TESTING=1 crew update` ?

bin/crew
```diff
@@ -104,9 +105,9 @@ end
 
 String.use_color = args["--color"] || !args["--no-color"]
 @opt_keep = args["--keep"]
-@opt_verbose = args["--verbose"]
+Verbose = args["--verbose"]
```
uberhacker, line 108: Lowercase `v`?

supechicken: This is a constant

supechicken:
> For instance, how would one replace the usage of `ENV['CREW_DISABLE_ENV_OPTIONS'] = '1'`?
```ruby
# setvar({ varName, defaultValue })
 setvar({ DISABLE_ENV_OPTIONS: false })
```

The variable name `DISABLE_ENV_OPTIONS` will be passed to the `setvar` function, which defines the value based on the given default value (`false` in this case). If the env variable (`ENV['CREW_DISABLE_ENV_OPTIONS']`) is valid, the value of the env variable will be used.

Reading the variable we just defined:
```
# $OPT is the new placeholder for all runtime variables, can be modified by packages without redefining

p $OPT[:DISABLE_ENV_OPTIONS]
# == or ==
# Using the const name CREW_#{varName}
# set in function `setvar`: `eval "CREW_#{varName} = #{value}"`

p CREW_DISABLE_OPTIONS
```

> And how would this change using env variables such as in `CREW_TESTING_REPO=https://github.com/supechicken666/chromebrew.git CREW_TESTING_BRANCH=runtime_var CREW_TESTING=1 crew update` ?

```ruby
# pass these variables to `setvar` with a default value (`nil, nil, false`)
setvar({ TESTING_REPO: nil,  TESTING_BRANCH: nil, TESTING: false })

# if the corresponding env variable is defined, the `$OPT` placeholder looks like this
$OPT[:TESTING_REPO] # => "<value of the CREW_TESTING_REPO env variable>"
CREW_TESTING_REOP # => same as $OPT[:TESTING_REPO] 
```

bin/crew
uberhacker: I get that but this change is not consistent with the other options.  If we are going to make this kind of a change, we should make it the same for all the other arguments too.  Ex: `@opt_keep`

supechicken: The changes of other arguments can be added in further refactoring

satmandu: Is it not ruby style to use (lowercase) `snake_case` for symbols, methods, and variables?

satmandu:
Can we abstract some of `setvar({ varName, defaultValue })` away?

For instance, right now we can set a variable in a package thus: `is_fake` which can be checked with `if @pkg.is_fake?`

Similarly we could just add `:no_env_options` to `package.rb` and then just be able to add `no_env_options` to a package file.

Could that be wired up in runtime_var or something so it translates to `setvar({ DISABLE_ENV_OPTIONS: false })`?

I think making packages have as little text as possible would be ideal.

satmandu:
I make a LOT of typos... so it would be beneficial for me to have to type fewer brackets and parentheses into files...

supechicken:
> Can we abstract some of `setvar({ varName, defaultValue })` away?
> 
> For instance, right now we can set a variable in a package thus: `is_fake` which can be checked with `@pkg.is_fake.to_s.empty?`
> 
> Similarly we could just add `:no_env_options` to `package.rb` and then just be able to add `no_env_options` to a package file.
> 
> Could that be wired up in runtime_var or something so it translates to `setvar({ DISABLE_ENV_OPTIONS: false })`?
> 
> I think making packages have as little text as possible would be ideal.

Good idea, this can be done by modifying the `def property` part (`lib/package_helper.rb`)

bin/crew
supechicken: I don't think there is a standard for variable names? (like the `pkgName` variable in `bin/crew`)

satmandu:
> > Can we abstract some of `setvar({ varName, defaultValue })` away?
> > For instance, right now we can set a variable in a package thus: `is_fake` which can be checked with `@pkg.is_fake.to_s.empty?`
> > Similarly we could just add `:no_env_options` to `package.rb` and then just be able to add `no_env_options` to a package file.
> > Could that be wired up in runtime_var or something so it translates to `setvar({ DISABLE_ENV_OPTIONS: false })`?
> > I think making packages have as little text as possible would be ideal.
> 
> Good idea, this can be done by modifying the `def property` part (`lib/package_helper.rb`)

Any idea how to do this properly?

I'm wondering if it would make sense to get rid of this entirely: 
```
def property (*properties)
  # property: like attr_accessor, but `=` is not needed to define a value
  # Examples:
  #   {prop_name}('example') # set {prop_name} to 'example'
  #   {prop_name}            # return the value of {prop_name}
  properties.each do |prop|
    self.class_eval <<~EOT
      def self.#{prop} (#{prop} = nil)
        @#{prop} = #{prop} if #{prop}
        @#{prop}
      end
    EOT
  end
end
```
(Unless we can figure out a way to also add `def self.#{prop}?` here somehow so that we don't have to also keep adding stuff like 
```
  def self.is_fake?
    @is_fake
  end
```
to `package.rb`. )

Also, we have overlap between runtime variables/options and package properties, like `is_fake`. Can we to the extent possible combine the two, at least for the boolean ones? Maybe that means making the `attr_accessor` properties/symbols read/write and adjustable/overridable from the command line, but more importantly, be able to set those variables simply as properties inside a package file.

If we can assign a property for each of those, then we can just check like with `@pkg.is_fake?`

Also, maybe then we could covert `$OPT[:DISABLE_ENV_OPTIONS]` to two booleans: `env_options` and `disable_env_options` or more simply `no_env_options` for packages.

That way we can have a global default, and just put one of those into a package file to deviate from the default.

It would be nice to be able to do that for something like`use_lto/no_use_lto`, `is_musl`, and `build_static` as well.

After all, you're suggesting: `setvar({ DISABLE_ENV_OPTIONS: false })`
I'm suggesting `disable_env_options` or `no_env_options`

I think that's clearer than a double negative?

bin/crew
uberhacker: I don't believe variables are capitalized because they can get confused with text (like words from a sentence).  Environment variables are generally all caps but other variables start with lowercase as @satmandu mentioned.  camelCase is acceptable too, although, I'm old school and prefer underscores for clarity.

satmandu:
Setting those symbols thus doesn't work:
`package.rb`:
```
class Package
  property :description, :homepage, :version, :license, :compatibility,
           :binary_url, :binary_sha256, :source_url, :source_sha256,
           :git_branch, :git_hashtag
  boolean_property :git_fetchtags, :is_fake,
           :is_musl, :is_static, :no_patchelf
```

`package_helpers.rb`:
```
def boolean_property (*properties)
  properties.each do |prop|
    self.class_eval <<~EOT
      def self.#{prop}
        @#{prop} = #{prop} unless #{prop}.to_s.empty?
        @#{prop}
      end
      def self.#{prop}?
        unless @#{prop}.to_s.empty?
          true
        else
          false
        end
      end
    EOT
  end
end
```

PR #6655, opened 2022-01-25
add mold linker package and deps

satmandu:
- Mold is allegedly faster than gold & uses fallback to gold for `lto`, which it doesn't support yet.
- `-fuse-ld=mold` will be in gcc12
- this just adds the mold package.

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686 (needs to be built on a machine with a 3.x kernel)

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=mold CREW_TESTING=1 crew update
```


uberhacker:
Not something you want to tell your mother.  "I use mold to do my linking.  What do you use?"

satmandu:
I mean... she might be a mycologist...

PR #6657, opened 2022-01-26
polkit fixes for CVE-2021-4034

satmandu:
- Fix for [CVE-2021-4034](https://www.qualys.com/2022/01/25/cve-2021-4034/pwnkit.txt) 
- removed `js_` dep and replaced with `duktape` and its build dep `setconf`

Builds properly:
- [x] x86_64
- [x] armv7l
- [x] i686

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=polkit CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/duktape.rb
+++ b/packages/duktape.rb
@@ -9,7 +9,7 @@ class Duktape < Package
   version '2.6.0'
   compatibility 'all'
   license 'MIT'
-  source_url 'https://duktape.org/duktape-2.6.0.tar.xz'
+  source_url "https://duktape.org/duktape-#{version}.tar.xz"
   source_sha256 '96f4a05a6c84590e53b18c59bb776aaba80a205afbbd92b82be609ba7fe75fa7'
 
   binary_url({
@@ -48,7 +48,8 @@ class Duktape < Package
   end
 
   def self.install
-    system "install -Dm644 duktape.pc #{CREW_DEST_LIB_PREFIX}/pkgconfig/duktape.pc"
+    FileUtils.mkdir_p %W[#{CREW_DEST_LIB_PREFIX}/pkgconfig]
+    FileUtils.install 'duktape.pc', "#{CREW_DEST_LIB_PREFIX}/pkgconfig/duktape.pc", mode: 0o644
     system 'make install'
   end
 end
--- a/packages/duktape.rb
+++ b/packages/duktape.rb
@@ -6,10 +6,11 @@ require 'package'
 class Duktape < Package
   description 'Embeddable Javascript engine'
   homepage 'https://duktape.org/'
-  version '2.6.0'
+  @_ver = '2.6.0'
+  version @_ver
   compatibility 'all'
   license 'MIT'
-  source_url "https://duktape.org/duktape-#{version}.tar.xz"
+  source_url "https://duktape.org/duktape-#{@_ver}.tar.xz"
   source_sha256 '96f4a05a6c84590e53b18c59bb776aaba80a205afbbd92b82be609ba7fe75fa7'
 
   binary_url({
@@ -40,7 +41,7 @@ class Duktape < Package
 
       Name: duktape
       Description: Embeddable Javascript engine
-      Version: 2.6.0
+      Version: #{@_ver}
       Libs: -L\${libdir} -lduktape
       Cflags: -I\${includedir}
     DUKTAPEPCEOF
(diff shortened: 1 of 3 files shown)


packages/polkit.rb
```diff
@@ -3,35 +3,42 @@
...
 
-  depends_on 'js91'
+  depends_on 'duktape'
   depends_on 'elogind'
   depends_on 'gtk_doc' => :build
   depends_on 'gobject_introspection' => :build
 
+  def self.patch
+    # Fix meson 0.60+ compatibility
+    system 'curl -OLf https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/99.patch'
```
supechicken, line 32: Use `downloader.rb`?
```ruby
downloader('https://gitlab.freedesktop.org/polkit/polkit/-/merge_requests/99.patch')
```

satmandu: Might be easier to just put the patch in as a ruby here doc?

satmandu: done

packages/setconf.rb
```diff
@@ -0,0 +1,34 @@
...
+     armv7l: 'd8106e68200d60bd648b24cf982e38e4a02b4ef06fc4f59b861925f1259e4788',
+       i686: 'ae85bdab10a698f6bfed275dc4b6cb560efcfc83d96ceb91035fc27a96e4911e',
+     x86_64: '152a3f9c7d5a4a7a505913cef5dec91820427faf5a50096c4f42fc36b27e1398'
+  })
+
+  depends_on 'python3'
+
+  def self.install
+    system "install -Dm755 setconf.py #{CREW_DEST_PREFIX}/bin/setconf"
+    system "install -Dm644 setconf.1.gz #{CREW_DEST_MAN_PREFIX}/man1/setconf.1.gz"
```
uberhacker, line 32: Rubyize?

packages/duktape.rb
```diff
@@ -0,0 +1,54 @@
...
+      Description: Embeddable Javascript engine
+      Version: 2.6.0
+      Libs: -L\${libdir} -lduktape
+      Cflags: -I\${includedir}
+    DUKTAPEPCEOF
+    File.write('duktape.pc', @duktapepc)
+  end
+
+  def self.install
+    system "install -Dm644 duktape.pc #{CREW_DEST_LIB_PREFIX}/pkgconfig/duktape.pc"
```
uberhacker, line 51: Rubyize?

packages/duktape.rb
```diff
@@ -0,0 +1,54 @@
...
+    system "setconf Makefile INSTALL_PREFIX #{CREW_DEST_PREFIX}"
+    @duktapepc = <<~DUKTAPEPCEOF
+      prefix=#{CREW_PREFIX}
+      exec_prefix=\${prefix}
+      libdir=#{CREW_LIB_PREFIX}
+      includedir=\${prefix}/include
+
+      Name: duktape
+      Description: Embeddable Javascript engine
+      Version: 2.6.0
```
uberhacker, line 43: Use `#{version}` instead of `2.6.0`?

packages/setconf.rb
satmandu: Stylistically I see the rationale for maximum rubyfication. But in the case of `install`, I think we should make an exception, since Ruby's install doesn't include `mkdir -p`.

I suggest either an upstream submission to add that to FileUtils.install, a modification to `system` (@supechicken ?) to add the necessary `FileUtils.mkdir_p`, or we keep using `system "install`...

satmandu: Why use two lines when one will suffice?

uberhacker: Only one additional line is needed with `FileUtils.mkdir_p ["#{CREW_DEST_PREFIX}/bin", "#{CREW_DEST_MAN_PREFIX}/man1"]`?  I think you can even get rid of the `"`s with `%x`?

uberhacker: I definitely understand the rationale behind reducing lines of code, however, we should be consistent with all our packages with whatever method we use.

packages/duktape.rb
uberhacker: Why was this marked resolved?

saltedcoffii: Maybe we should use @_ver instead of version

satmandu: @uberhacker This is weird. I go to edit on the web, and it shows #{version}.  Something isn't refreshing...

satmandu: Also, I was apparently looking at the beginning now where you highlighted.  Both should now be resolved.

PR #6661, opened 2022-01-28
patchelf option during package creation, revamp of boolean symbol processing

satmandu:
- boolean symbols in `package.rb` are automatically handled and can be placed in any package files, and are recognized by simply checking `@pkg.symbol?`
- The obvious follow-up would be to add plumbing to connect to the boolean env variables in `const.rb`
- allows use of `patchelf` to hardcode library paths if `patchelf` is set in a package (and explicitly disables if `no_patchelf` is set in a package. (disabled on `armv7l` currently because hardcoding to use our glibc's libc.so.6 breaks binaries on there.)
- Also let's `is_musl` in a package automatically load the `musl.rb` file.

Works properly:
- [x] x86_64
- [x] armv7l (no because libc.so.6 is broken on armv7l)
- [x] i686

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=crew_patchelf CREW_TESTING=1 crew update
```


bin/crew
```diff
@@ -983,11 +983,54 @@ def prepare_package(destdir)
...
 end
 
+def patchelf_set_need_paths(dir)
+  return unless @pkg.no_patchelf.to_s.empty?
+
+  Dir.chdir dir do
+    puts 'Running patchelf'.lightblue
+    abort('No Patchelf found!').lightred unless File.exist?("#{CREW_PREFIX}/bin/patchelf")
+
+    @execfiles = `find . -executable -type f ! \\( -name \"*.a\" \\) -exec sh -c \"file -i \'{}\' | grep -q \'executable; charset=binary\'\" \\; -exec ls -1i {} \\; | sort -u -n -s -k1,1 | awk '{print $2}'`.chomp
```
supechicken, line 1004: How about checking the ELF file header directly instead of passing it to `file` and `grep` for ELFs?
```bash
bash -c "[[ \"\$(head -c4 {})\" == \$'\x7FELF' ]]"
```
(also it is much faster then `file+grep`):
```
chronos@localhost /usr/local/bin $ time bash -c "for f in *; do file -i \$f | grep -q 'executable; charset=binary'; done"

real    0m4.581s
user    0m7.016s
sys     0m1.639s
chronos@localhost /usr/local/bin $ time bash -c "for f in *; do [[ \"\$(head -c4 \$f)\" == \$'\x7FELF' ]]; done"

real    0m1.790s
user    0m0.790s
sys     0m1.103s
```

bin/crew
```diff
@@ -983,11 +983,54 @@ def prepare_package(destdir)
...
+       patchelf_set_need_paths destdir
+    end
+
     # use upx on executables
     shrink_dir destdir
   end
 end
 
+def patchelf_set_need_paths(dir)
+  return unless @pkg.no_patchelf.to_s.empty?
```
supechicken, line 998: `@pkg.no_patchelf` should be already a boolean?

bin/crew
```diff
@@ -983,11 +983,54 @@ def prepare_package(destdir)
...
+
+      @neededlibs.each_line(chomp: true) do |neededlibspatch|
+        next if neededlibspatch.include?("#{@patchelf_lib_prefix}")
+
+        @neededlib_basename = File.basename(neededlibspatch)
+        @neededlibspatchednamepath = "#{@patchelf_lib_prefix}/" + @neededlib_basename
+        # The first check here can be changed to just check the dest_dir
+        # hierarchy for @neededlib_basename if the intent is to allow 
+        # using a different CREW_PREFIX during package installs.
+        if File.exist?("#{@neededlibspatchednamepath}") or File.exist?(Dir.pwd + @neededlibspatchednamepath)
```
supechicken, line 1023: `File.exist?(@neededlibspatchednamepath)`?

bin/crew
satmandu: It is `nil` if it doesn't exist. There is code in `package.rb` to make `is_fake` a Boolean, but I think we can avoid having to do that for everything if just use the `.to_s.empty?` trick!

supechicken: As I know `nil` acts like `false` in comparing (`if nil` is the same as `if false`)?

As an alternative, we can use `!!@pkg.no_patchelf` to force convert the value to boolean (All object except `nil` and `false` will convert to `true`, `nil` and `false` will convert to `false`)

uberhacker: Unfortunately, environment variables don't work this way.  `ENV['SOME_VARIABLE']=false` is actually true.  In order to make an environment variable "false", it would need to be unset like `ENV['SOME_VARIABLE']=`.

bin/crew
satmandu: I will have to see if this works...

bin/crew
satmandu: If this works we should use if everywhere...

bin/crew
```diff
@@ -983,11 +983,54 @@ def prepare_package(destdir)
 
     strip_dir destdir
 
+    # Use patchelf to set need paths for all binaries
+    # currently broken due to glibc issue on armv7l
+    unless ARCH == 'armv7l' || ARCH == 'aarch64'
```
uberhacker, line 988: Would something like this work?
```
['aarch64', 'armv7l'].include?(ARCH)
```

bin/crew
```diff
@@ -983,11 +983,54 @@ def prepare_package(destdir)
...
 end
 
+def patchelf_set_need_paths(dir)
+  return if !!@pkg.no_patchelf?
+
+  Dir.chdir dir do
+    puts 'Running patchelf'.lightblue
+    abort('No Patchelf found!').lightred unless File.exist?("#{CREW_PREFIX}/bin/patchelf")
+
+    @execfiles = `find . -executable -type f ! \\( -name \"*.a\" \\) -exec sh -c \"[[ \"\$(head -c4 {})\" == \$'\x7FELF' ]]" \\; -exec ls -1i {} \\; | sort -u -n -s -k1,1 | awk '{print $2}'`.chomp
```
supechicken, line 1004: Notice that the `[[ str1 == str2 ]]` expression and the `$'\x7F'` ascii escape only works on `bash`.

satmandu: I think `sh` invokes `bash` on ChromeOS, no?

supechicken: When `sh` called, `/usr/local/bin/sh` will be in the first priority of target executable, and the `/usr/local/bin/sh` file is a symlink provided by `bash` package. `/bin/sh` is a symlink to `/bin/dash`

satmandu: @uberhacker looks like we should be using `bash` in lieu of `sh` since `bash` is in neither `core` nor `buildessential`.

bin/crew
satmandu: @supechicken I'm seeing this:
```
head: cannot open '{}' for reading: No such file or directory
bash: -c: line 1: conditional binary operator expected
bash: -c: line 1: syntax error near `$'ELF''
bash: -c: line 1: `[[  == $'ELF' ]]'
```

satmandu: with this line:
```
@execfiles = `find . -executable -type f ! \\( -name \"*.a\" \\) -exec bash -c \"[[ \"\$(head -c4 {})\" == \$'\x7FELF' ]]" \\; -exec ls -1i {} \\; | sort -u -n -s -k1,1 | awk '{print $2}'`.chomp
```

satmandu: I couldn't get the subshell to work. But this works:
```
    @execfiles = `find . -executable -type f ! \\( -name \"*.a\" \\) -exec head -c4 {} 2>/dev/null \\; -printf ' %i %p\\n'| grep ^.ELF | sort -u -n -s -k2,2 | awk '{print $3}'`.chomp

```

bin/crew
satmandu: @supechicken I can't seem to get this to work. `!!@pkg.no_patchelf` is always false. In fact, the other properties I set aren't showing up either. Do we need to be modifying package_helpers to make sure that properties are set?

What's the simplest way to put a property in a file and then be able to see if it is there?

And why are we setting all these properties in package_helpers?

Should we be using `include DataMapper::Resource` like [here](https://www.rubydoc.info/github/datamapper/dm-core/DataMapper/Property)?

bin/crew
```diff
@@ -1000,11 +1003,53 @@ def prepare_package(destdir)
...
+  return if @pkg.no_patchelf? || !@pkg.patchelf?
+
+  Dir.chdir dir do
+    puts 'Running patchelf'.lightblue
+    abort('No Patchelf found!').lightred unless File.exist?("#{CREW_PREFIX}/bin/patchelf")
+    @execfiles = `find . -executable -type f ! \\( -name \"*.a\" \\) -exec head -c4 {} 2>/dev/null \\; -printf ' %i %p\\n'| grep ^.ELF | sort -u -n -s -k2,2 | awk '{print $3}'`.chomp
+    return if @execfiles.to_s.empty?
+
+    @patchelf_lib_prefix = @pkg.is_musl? ? "#{CREW_MUSL_PREFIX}/lib" : CREW_LIB_PREFIX
+    puts "@patchelf_lib_prefix is #{@patchelf_lib_prefix}"
```
uberhacker, line 1027: Should this be visible only with `-v` or `--verbose` options?

bin/crew
```diff
@@ -1000,11 +1003,53 @@ def prepare_package(destdir)
...
+      @neededlibs.each_line(chomp: true) do |neededlibspatch|
+        next if neededlibspatch.include?(@patchelf_lib_prefix.to_s)
+
+        @neededlib_basename = File.basename(neededlibspatch)
+        @neededlibspatchednamepath = "#{@patchelf_lib_prefix}/" + @neededlib_basename
+        # The first check here can be changed to just check the dest_dir
+        # hierarchy for @neededlib_basename if the intent is to allow
+        # using a different CREW_PREFIX during package installs.
+        if File.exist?(@neededlibspatchednamepath) || File.exist?(Dir.pwd + @neededlibspatchednamepath)
+          puts "patchelf --replace-needed #{neededlibspatch} #{@neededlibspatchednamepath} #{execfiletopatch}"
```
uberhacker, line 1043: Should this be visible only with `-v` or `--verbose` options?

bin/crew
```diff
@@ -1000,11 +1003,53 @@ def prepare_package(destdir)
...
+          system "patchelf --replace-needed #{neededlibspatch} #{@neededlibspatchednamepath} #{execfiletopatch}"
+        else
+          puts "#{execfiletopatch} needed library #{@neededlib_basename} not found in #{@patchelf_lib_prefix} or #{Dir.pwd + @neededlibspatchednamepath}."
```
uberhacker, line 1046: Use `.lightred` to indicate an error?

lib/package.rb
```diff
@@ -162,7 +158,7 @@ def self.system(*args, **opt_args)
     rescue => e
       # print failed line number and error message
       puts "#{e.backtrace[1]}: #{e.message}".orange
-      raise InstallError, "`#{env.map {|k, v| "#{k}=\"#{v}\"" } .join(' ')} #{cmd_args.join(' ')}` exited with #{$?.exitstatus}"
+      raise InstallError, "`#{env.map { |k, v| "#{k}=\"#{v}\"" }.join(' ')} #{cmd_args.join(' ')}` exited with #{$?.exitstatus}"
```
uberhacker, line 161: Use `.lightred` to indicate an error?

satmandu:
changes made.

bin/crew
```diff
@@ -1000,11 +1003,53 @@ def prepare_package(destdir)
...
+        @neededlib_basename = File.basename(neededlibspatch)
+        @neededlibspatchednamepath = "#{@patchelf_lib_prefix}/" + @neededlib_basename
+        # The first check here can be changed to just check the dest_dir
+        # hierarchy for @neededlib_basename if the intent is to allow
+        # using a different CREW_PREFIX during package installs.
+        if File.exist?(@neededlibspatchednamepath) || File.exist?(Dir.pwd + @neededlibspatchednamepath)
+          puts "patchelf --replace-needed #{neededlibspatch} #{@neededlibspatchednamepath} #{execfiletopatch}" if @opt_verbose
+          system "patchelf --replace-needed #{neededlibspatch} #{@neededlibspatchednamepath} #{execfiletopatch}"
+        else
+          puts "#{execfiletopatch} needed library #{@neededlib_basename} not found in #{@patchelf_lib_prefix} or #{Dir.pwd + @neededlibspatchednamepath}." if @opt_verbose
```
uberhacker, line 1046: Missing `.lightred`?

PR #6686, opened 2022-01-29
Add reload_constants command

uberhacker:
Also reduces clutter in packages.

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -69,7 +69,24 @@ CREW_LICENSE = <<~LICENSESTRING
 LICENSESTRING
 
 # All available crew commands.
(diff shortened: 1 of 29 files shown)


satmandu:
Does it make sense to use this as a `crew reload` command?

Don't the variables get reloaded based upon current env variables every time one runs crew?

uberhacker:
Does this work?
```
$ export CREW_NOT_STRIP='1'
$ crew const CREW_NOT_STRIP
```
If so, then we don't need the `crew reload` command.

satmandu:
> Does this work?
> ```
> $ export CREW_NOT_STRIP='1'
> $ crew const CREW_NOT_STRIP
> ```
> If so, then we don't need the `crew reload` command.

uberhacker:
We don't need it.  I'll remove.

PR #6691, opened 2022-01-30
Refactor colorizing in install.sh

saltedcoffii:
In install.sh, there's a lot of repeated `echo -e "${COLOR}Message. ${RESET}"`. This pull request simplifies those messages five functions, one for each color. 

This also adds some pretty Chromebrew ASCII art for when Chromebrew finishes installing. 
Tested on actual hardware, works.


satmandu:
Could we convert those functions to snake_case?

e.g.
```
echo_error() { echo -e "${RED}${*}${RESET}" >&2; }
echo_info() { echo -e "${YELLOW}${*}${RESET}" >&1; }
echo_success() { echo -e "${GREEN}${*}${RESET}" >&1; }
echo_intra() { echo -e "${BLUE}${*}${RESET}" >&1; }
echo_out() { echo -e "${GRAY}${*}${RESET}" >&1; }
```
I think that helps readability.

saltedcoffii:
can we use camelCase, it's easier to type

install.sh
```diff
@@ -277,39 +284,66 @@ git sparse-checkout set packages lib bin crew tools install.sh
...
+             .xMNO0X0OOkkkkkkkkkkXMWKXKOx;.
+             .0MXOOOOOOkkkkkkkkkOKM0..
+             'NMWNXXKK000000KKXNNMMX.
+             .;okk0XNWWMMMMWWNKOkdc'.
+                .....'cc:cc:''..."
+echo "  ___  _                               _
+ / (_)| |                             | |
+|     | |     ,_    __   _  _  _    _ | |   ,_    _
+|     |/ \   /  |  /  \_/ |/ |/ |  |/ |/ \_/  |  |/  |  |  |_
+ \___/|   |_/   |_/\__/   |  |  |_/|__/\_/    |_/|__/ \/ \/
```
uberhacker, line 326: Whoa, that is some fancy ascii art!  It might make sense to add more of a hump on the `h`?

uberhacker:
> can we use camelCase, it's easier to type

camelCase is not normally used in bash scripts.  See https://google.github.io/styleguide/shellguide.html#naming-conventions.

install.sh
saltedcoffii: yeah what would that look like?

uberhacker: I was thinking similar to the `r`.

saltedcoffii: oh okay

satmandu:
> > can we use camelCase, it's easier to type
> 
> camelCase is not normally used in bash scripts. See https://google.github.io/styleguide/shellguide.html#naming-conventions.

Looks like I was mischaracterizing what I was suggesting!

`Lower-case, with underscores to separate words.` (from that page) is what I was asking for. :)

install.sh
satmandu: I think there is an ascii art generator I used ~ 20+ years ago to generate a signature in my email.... There's zero chance of me finding that now, alas..

saltedcoffii: it's okay i used figlet, it worked well enough.

uberhacker: That looks better but I'm confused by the extra vertical line between the `C` and `h`.

saltedcoffii: this is a cursive script, look at a cursive h
also because of the `echo '` line, the top line of characters is moved over

uberhacker: Fair enough.  I probably don't know how to write in cursive now anyway since it's been so long since I did any actual writing.  Does that make me illiterate?

PR #6698, opened 2022-02-02
Vivaldi 5.0.2497.48 => 5.0.2497.51

uberhacker:


packages/vivaldi.rb
```diff
@@ -18,14 +17,14 @@ class Vivaldi < Package
 
   case ARCH
   when 'aarch64', 'armv7l'
-    @_arch = 'armhf'
-    source_sha256 'f0603cc97800b040879c69af05d0264385d07ea9a3ec886c14ec6fb7f0473302'
+    arch = 'armhf'
```
satmandu, line 20: Does this mean that `arch` is set for all packages? Should we be using `@arch` here?

uberhacker: This was changed to a local variable as opposed to instance because the instance scope is not needed in this context.

PR #6702, opened 2022-02-02
Add memory lower limit for using pixz for package compression

satmandu:
- Fixes issue with `pixz` failing to compress on systems with less ram
- Solution is to just avoid using `pixz` for compression on such systems.

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=pixz_ram_check CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1435,9 +1435,8 @@ end
 
 def archive_package(pwd)
   @ramfree = %x(free).split(" ")[7]
-  # Set minimum ram for running pixz to be ~ 8Gb. (This number is just
-  # below that on a RPI4B used for building armv7l packages.)
-  unless @ramfree < '7999000' || CREW_NOT_USE_PIXZ || !File.exist?("#{CREW_PREFIX}/bin/pixz")
+  # Set minimum ram for running pixz to be ~ 4 Gb.
+  unless @ramfree < '3999500' || CREW_NOT_USE_PIXZ || !File.exist?("#{CREW_PREFIX}/bin/pixz")
     puts "Using pixz to compress archive."
     pkg_name = "#{@pkg.name}-#{@pkg.version}-chromeos-#{@device[:architecture]}.tpxz"
     Dir.chdir CREW_DEST_DIR do
--- a/bin/crew
+++ b/bin/crew
@@ -1434,7 +1434,10 @@ def build_package(pwd)
 end
 
 def archive_package(pwd)
-  unless CREW_NOT_USE_PIXZ || !File.exist?("#{CREW_PREFIX}/bin/pixz")
+  @ramfree = %x(free).split(" ")[7]
+  # Set minimum ram for running pixz to be ~ 8Gb. (This number is just
+  # below that on a RPI4B used for building armv7l packages.)
+  unless @ramfree < '7999000' || CREW_NOT_USE_PIXZ || !File.exist?("#{CREW_PREFIX}/bin/pixz")
     puts "Using pixz to compress archive."
     pkg_name = "#{@pkg.name}-#{@pkg.version}-chromeos-#{@device[:architecture]}.tpxz"
     Dir.chdir CREW_DEST_DIR do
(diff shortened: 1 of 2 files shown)


saltedcoffii:
can you show an example of pixz failing on a computer with less ram? i haven't had an issue with that building on actual hardware (4GB of RAM)

uberhacker:
Bump version in `lib/const.rb`?

uberhacker:
> can you show an example of pixz failing on a computer with less ram? i haven't had an issue with that building on actual hardware (4GB of RAM)

I imagine this depends on the size of the file being compressed too.  I was having trouble with large files with only 2 gb of ram.

satmandu:
version bumped

satmandu:
> can you show an example of pixz failing on a computer with less ram? i haven't had an issue with that building on actual hardware (4GB of RAM)

I can lower the limit to 4Gb... what do you see from `%x(free).split(" ")[7]` on your system @saltedcoffii ?

On low ram systems, pixz just fails. See https://github.com/vasi/pixz/issues/97

satmandu:
@saltedcoffii As long as your ram is > `3999500`, which I just set the max to, pixz may be fine?

saltedcoffii:
I'd be happy if chromebrew emitted a warning letting the user know that RAM is too low to compress using pixz, that pixz can be forced using an environment variable, and that chromebrew is falling back to xz.

uberhacker:
If pixz requires more that 4 gb of ram, you can disable with an architecture conditional.  32-bit systems can only physically address 4 gb max.

satmandu:
Honestly, since I don't have time to implement the individual file decompression stuff with pixz I was planning on, I think it best to just limit this here. I think doing something like zstd for compression might be much better anyways, and I think zstd might handle low memory situations better anyways...

For decompression of course it doesn't matter.

satmandu:
> If pixz requires more that 4 gb of ram, you can disable with an architecture conditional. 32-bit systems can only physically address 4 gb max.

Yes, but when running in a container on a `x86_64` system, the i686 apps are PAE aware, and can throw 4Gb at each process, so it is the total ram available which matters.

uberhacker:
> > If pixz requires more that 4 gb of ram, you can disable with an architecture conditional. 32-bit systems can only physically address 4 gb max.
> 
> Yes, but when running in a container on a `x86_64` system, the i686 apps are PAE aware, and can throw 4Gb at each process, so it is the total ram available which matters.

But you are talking about Docker containers.  We probably should be making decisions based on Chromebook hardware instead.

satmandu:
> 
> But you are talking about Docker containers. We probably should be making decisions based on Chromebook hardware instead.

We're talking about building packages. Since most packages are being generated in containers that should take priority. Decompression is of course not an issue at all on any hardware.

uberhacker:
> > But you are talking about Docker containers. We probably should be making decisions based on Chromebook hardware instead.
> 
> We're talking about building packages. Since most packages are being generated in containers that should take priority. Decompression is of course not an issue at all on any hardware.

I kind of disagree since there are those who have a Chromebook as their only option and a failure on a large build will be disappointing and confusing.

uberhacker:
What if we replace pixz with something else as discussed in the Slack channel?

saltedcoffii:
like with tarlz?

satmandu:
> I kind of disagree since there are those who have a Chromebook as their only option and a failure on a large build will be disappointing and confusing.

Oh yes, I agree with you here.  On ChromeOS hardware `crew` should never break due to insufficient hardware. This is why I think we should try to be safe and not present options _to the end-user_ which may result in breakage for them.

satmandu:
> like with tarlz?

I think at the moment my problem with tarlz is that we are already using xz, which is also a lzma derived compression format. 

I don't see any major benefits from the switch?

satmandu:
Resolved by https://github.com/skycocker/chromebrew/pull/6703

PR #6703, opened 2022-02-03
crew: use zstd

satmandu:
- pixz is having issues on low memory systems.
- This PR implements using `zstd` as a default compression for chromebrew packages.
- zstd compression settings were copied from ARCH based upon this analysis: https://lists.archlinux.org/pipermail/arch-dev-public/2019-March/029542.html
- A static musl tar is also added to this PR to compensate for machines which might have older versions of tar which do not support zstd.
- `install.sh`'s bootstrap_packages have been set to use `no_zstd` when building so we don't have a chicken/egg issue when doing installs. (This may actually not be an issue since I put `tar` at the beginning of `bootstrap_packages`.
- `gcc11` compressed with zstd on a 2Gb RAM `i686` machine with the settings in crew, so that's good!
- Added `tar` to `core` and `buildessential`
- Added update to rust, compressed with `zstd`. `crew upgrade rust`

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686
```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=zstd_crew CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1086,7 +1086,7 @@ def strip_dir(dir)
 
       # Strip binaries but not compressed archives
       puts "Stripping binaries..."
-      extensions = [ 'bz2', 'gz', 'lha', 'lz', 'lzh', 'rar', 'tar', 'tbz', 'tgz', 'tpxz', 'txz', 'xz', 'Z', 'zip' ]
+      extensions = %w[ bz2 gz lha lz lzh rar tar tbz tgz tpxz txz xz Z zip zst ]
       inames = extensions.join(' ! -iname *\.')
       strip_find_files "find . -type f ! -iname *\.#{inames} -perm /111 -print | sed -e '/lib.*\.a$/d' -e '/lib.*\.so/d'"
     end
--- a/bin/crew
+++ b/bin/crew
@@ -1479,7 +1479,7 @@ def build_package(pwd)
 end
 
 def archive_package(pwd)
-  unless %x[tar --version | head -n 1 | awk '{print $4}'] > '1.31' || !File.exist?("#{CREW_PREFIX}/bin/zstd") || @pkg.no_zstd?
+  unless %x[tar --version | head -n 1 | awk '{print $4}'] < '1.31' || !File.exist?("#{CREW_PREFIX}/bin/zstd") || @pkg.no_zstd?
     puts "Using zstd to compress archive."
     pkg_name = "#{@pkg.name}-#{@pkg.version}-chromeos-#{@device[:architecture]}.zst"
     Dir.chdir CREW_DEST_DIR do
--- a/bin/crew
+++ b/bin/crew
@@ -828,7 +828,7 @@ def unpack(meta)
         abort 'zstd is needed for this install. Please install it first with \'crew install zstd\''.lightred
       end
       puts "Unpacking archive using 'tar', this may take a while..."
-      system "tar x#{@verbose}f #{meta[:filename]} -C #{@extract_dir}", exception: true
+      system "PATH=#{CREW_MUSL_PREFIX}/bin:\$PATH tar x#{@verbose}f #{meta[:filename]} -C #{@extract_dir}", exception: true
     when /\.deb$/i
       puts "Unpacking '.deb' archive, this may take a while..."
       DebUtils::extract_deb(meta[:filename], /data\..*/)
@@ -1505,7 +1505,7 @@ def archive_package(pwd)
       # decompression speed over compression speed.
       # See https://lists.archlinux.org/pipermail/arch-dev-public/2019-March/029542.html
       # Use nice so that user can (possibly) do other things during compression.
-      system "tar c#{@verbose} * | nice -n 20 zstd -c -T0 --ultra -20 - > #{pwd}/#{pkg_name}"
+      system "PATH=#{CREW_MUSL_PREFIX}/bin:\$PATH tar c#{@verbose} * | nice -n 20 zstd -c -T0 --ultra -20 - > #{pwd}/#{pkg_name}"
     end
   else
     puts "Using xz to compress package. This may take some time.".lightblue
--- a/bin/crew
+++ b/bin/crew
@@ -974,7 +974,7 @@ def prepare_package(destdir)
     Dir.foreach(CREW_DEST_PREFIX) do |filename|
       next if filename == '.' or filename == '..'
       unless @fhs_compliant_prefix.include?(filename)
-        if CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY
+        if CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY || @pkg.no_fhs?
           puts "Warning: #{CREW_PREFIX}/#{filename} in #{@pkg.name} is not FHS3 compliant.".orange
         else
           puts "Error: #{CREW_PREFIX}/#{filename} in #{@pkg.name} is not FHS3 compliant.".lightred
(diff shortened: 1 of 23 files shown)


uberhacker:
Fixes #6702?

bin/crew
```diff
@@ -1479,14 +1485,18 @@ def build_package(pwd)
 end
 
 def archive_package(pwd)
-  unless CREW_NOT_USE_PIXZ || !File.exist?("#{CREW_PREFIX}/bin/pixz")
-    puts "Using pixz to compress archive."
-    pkg_name = "#{@pkg.name}-#{@pkg.version}-chromeos-#{@device[:architecture]}.tpxz"
+  unless %x[tar --version | head -n 1 | awk '{print $4}'] < '1.31' || !File.exist?("#{CREW_PREFIX}/bin/zstd") || @pkg.no_zstd?
```
uberhacker, line 1488: Would it make sense to convert this into a crew constant?  Something like `CREW_TAR_VERSION`?

satmandu: We only need to know that when we are using tar. Why do that check every single time crew is run?

satmandu: Getting that value will make crew 0.010s to 0.157s slower/run, comparing i686 and x86_64 runs of that command.

packages/buildessential.rb
```diff
@@ -119,6 +119,10 @@ class Buildessential < Package
   # Samurai
   #depends_on 'samurai'
 
+  # Tar
+  # Version > 1.30 needed for zstd support
+  depends_on 'tar'
```
uberhacker, line 124: I'm not exactly a fan of needing an updated tar to compress files.  It would be better to use all of the out-of-the-box versions instead but...

packages/git.rb
```diff
@@ -37,6 +37,7 @@ class Git < Package
   depends_on 'musl_expat' => :build
 
   is_static
+  no_zstd
```
uberhacker, line 40: Are we going to need to add this to every package that needs it now?

satmandu: It's really only needed for packages installed before zstd and a tar which supports zstd. So that's a very very small list.

And it may not be needed at all, since I think tar was built with zstd... I should check that...

satmandu: I need to add a check for zstd...

uberhacker: So many compression options and so little time. :)

packages/buildessential.rb
satmandu: Only `i686` needs an updated tar. Other systems appear to have a new enough tar. Let me change that...

satmandu: The updated tar is needed for `decompressing` though...

satmandu: Changed the dependency to i686 only

packages/musl_zstd.rb
```diff
@@ -3,46 +3,54 @@
...
 
   binary_url({
-    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/musl_zstd/1.5.1_armv7l/musl_zstd-1.5.1-chromeos-armv7l.tpxz',
-     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/musl_zstd/1.5.1_armv7l/musl_zstd-1.5.1-chromeos-armv7l.tpxz',
-       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/musl_zstd/1.5.1_i686/musl_zstd-1.5.1-chromeos-i686.tpxz',
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/musl_zstd/1.5.1_x86_64/musl_zstd-1.5.1-chromeos-x86_64.tpxz'
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/musl_zstd/1.5.2_armv7l/musl_zstd-1.5.2-chromeos-armv7l.tar.xz',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/musl_zstd/1.5.2_armv7l/musl_zstd-1.5.2-chromeos-armv7l.tar.xz',
+       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/musl_zstd/1.5.2_i686/musl_zstd-1.5.2-chromeos-i686.tar.xz',
+     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/musl_zstd/1.5.2_x86_64/musl_zstd-1.5.2-chromeos-x86_64.tar.xz'
```
uberhacker, line 17: So we're turned full circle and now back to `.tar.xz`? :confounded:

bin/crew
```diff
@@ -675,7 +675,7 @@ def download
   Dir.chdir CREW_BREW_DIR do
     case File.basename(filename)
     # Sources that download with curl
-    when /\.zip$/i, /\.(tar(\.(gz|bz2|xz|lz))?|tgz|tbz|tpxz|txz)$/i, /\.deb$/i, /\.AppImage$/i
+    when /\.zip$/i, /\.(tar(\.(gz|bz2|xz|lz|zst))?|tgz|tbz|tpxz|txz)$/i, /\.deb$/i, /\.AppImage$/i
```
uberhacker, line 678: Add the `zst` extension to ` extensions = [ 'bz2', 'gz', 'lha', 'lz', 'lzh', 'rar', 'tar', 'tbz', 'tgz', 'tpxz', 'txz', 'xz', 'Z', 'zip' ]` in `def strip_dir`?

bin/crew
uberhacker: Fair enough.  I generally like to abstract kludgy logic like this, however.

packages/musl_zstd.rb
satmandu: Only for the bootstrap packages... but literally, only `tar` and `musl_zstd` now have to be `.tar.xz`

bin/crew
satmandu: done

bin/crew
```diff
@@ -1068,7 +1086,7 @@ def strip_dir(dir)
 
       # Strip binaries but not compressed archives
       puts "Stripping binaries..."
-      extensions = [ 'bz2', 'gz', 'lha', 'lz', 'lzh', 'rar', 'tar', 'tbz', 'tgz', 'tpxz', 'txz', 'xz', 'Z', 'zip' ]
+      extensions = %w[ bz2 gz lha lz lzh rar tar tbz tgz tpxz txz xz Z zip zst ]
```
uberhacker, line 1089: Good call.  That reduced the size of that array almost in half!

packages/musl_zstd.rb
uberhacker: This is going to be difficult to remember when we do updates.  It'd be nice if we could be consistent with a generally accepted compression scheme to be used everywhere.

tools/gl.sh
```diff
@@ -39,6 +41,7 @@ if test $1; then
 else
   packages=$(find ../release/*/*.xz -exec basename -s .tar.xz {} + | cut -d'-' -f1 | sort | uniq | xargs)" "
   packages+=$(find ../release/*/*.tpxz -exec basename -s .tpxz {} + | cut -d'-' -f1 | sort | uniq | xargs)
```
uberhacker, line 43: Add `" "` to the end?

packages/musl_zstd.rb
satmandu: Unless PKG name includes `tar` or `musl_zstd` use zstd...

satmandu: I can remove the no_zstd from all of the other packages.

PR #6716, opened 2022-02-09
Provide feedback for ruby gem updates

uberhacker:


packages/ruby.rb
```diff
@@ -63,6 +63,8 @@ def self.install
   end
 
   def self.postinstall
-    system 'gem update --silent -N --system'
+    puts 'Updating ruby gems. This may take a while...'
```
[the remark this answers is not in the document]
uberhacker: Added.

packages/ruby.rb
```diff
@@ -63,6 +63,11 @@ def self.install
...
+    if `grep -c "gem: --no-document" #{HOME}/.gemrc`.to_i.zero?
```
supechicken, line 67: Check for exit status instead?
```ruby
if system 'grep', '-q', "gem: --no-document", "#{HOME}/.gemrc"
```

packages/ruby.rb
```diff
@@ -63,6 +63,11 @@ def self.install
...
+      system "echo 'gem: --no-document' >> #{HOME}/.gemrc"
```
supechicken, line 68: Rubyize with `File.write("#{HOME}/.gemrc", "gem: --no-document\n", mode: 'a')`

satmandu:
Added @supechicken's added changes. @supechicken please test and merge if that works...

supechicken:
Tested on `x86_64`

PR #6724, opened 2022-02-11
dbus_glib => 0.112

saltedcoffii:
builds on x86_64. needs binaries.

```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=dbus_glib_0.112 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/dbus_glib.rb
+++ b/packages/dbus_glib.rb
@@ -1,36 +1,23 @@
 require 'package'
 
 class Dbus_glib < Package
-  description 'The D-Bus GLib package contains GLib interfaces to the D-Bus API.'
-  homepage 'http://www.linuxfromscratch.org/blfs/view/svn/general/dbus-glib.html'
-  version '0.110'
+  description 'An obsolete, primarily unmaintained glib binding for libdbus.'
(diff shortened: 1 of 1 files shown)


packages/dbus_glib.rb
```diff
@@ -1,36 +1,23 @@
 require 'package'
 
 class Dbus_glib < Package
-  description 'The D-Bus GLib package contains GLib interfaces to the D-Bus API.'
-  homepage 'http://www.linuxfromscratch.org/blfs/view/svn/general/dbus-glib.html'
-  version '0.110'
+  description 'An obsolete, primarily unmaintained glib binding for libdbus.'
```
uberhacker, line 4: Why the description change?  The old one seems to be more descriptive?

saltedcoffii: because the wiki says its unmantained

uberhacker: It's OK to mention that but maybe just append to the current description?

uberhacker: This is the description from the new homepage.
```
dbus-glib is a deprecated API for use of D-Bus from GLib applications.
```

PR #6743, opened 2022-02-16
crew: Download `source_url` as a normal file even if it is not an archive

supechicken:
Currently, the download algorithm will not download the `source_url` if it is not an archive or `git` URL. Added a part to this PR that handles normal files.

Tested on `x86_64`

```shell
CREW_TESTING=1 CREW_TESTING_REPO=https://github.com/supechicken/chromebrew CREW_TESTING_BRANCH=download_file crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -731,10 +731,20 @@ def download
 
     when /^SKIP$/i
       Dir.mkdir @extract_dir
-
-    # Source URLs which end with .git are git sources.
-    when /\.git$/i
+    when /\.git$/i # Source URLs which end with .git are git sources.
       @git = true
+    else
+      Dir.mkdir @extract_dir
+      # use curl if CURL environment variable set
+      if CURL == 'curl'
+        downloader url, filename, @opt_verbose
+      else
+        system "#{CURL} --retry 3 -#{@verbose}#LC - --insecure \'#{url}\' --output #{filename}"
+      end
+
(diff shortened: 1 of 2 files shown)


bin/crew
```diff
@@ -731,10 +731,22 @@ def download
...
     when /^SKIP$/i
       Dir.mkdir @extract_dir
-
-    # Source URLs which end with .git are git sources.
-    when /\.git$/i
+    when /\.git$/i # Source URLs which end with .git are git sources.
       @git = true
+    else
+      Dir.mkdir @extract_dir
+      # use curl if CURL environment variable set
```
uberhacker, line 738: Shouldn't this say `use downloader if CURL environment variable set`?

supechicken: If the `CURL` variable is set by the user, it might contain some `curl` specific environment variables and `curl` will be called instead.

supechicken: Renamed it to `CREW_USE_CURL`

bin/crew
```diff
@@ -731,10 +731,22 @@ def download
...
+      if CURL == 'curl'
+        downloader url, filename, @opt_verbose
+      else
+        system "#{CURL} --retry 3 -#{@verbose}#LC - --insecure \'#{url}\' --output #{filename}"
+      end
```
uberhacker, line 743:
```
      unless CREW_USE_CURL
        downloader url, filename, @opt_verbose
      else
      end
```
?

PR #6748, opened 2022-02-17
crew: Allow package name starts with digits

supechicken:
Due to the restriction with ruby's `class` naming system (a capital letter must be the first character of a `class` name), we are unable to name packages with a name that starts with digits (like `class 7_zip < Package`).

This PR allows package names with digits as the beginning, but the package class needs to be named with the prefix `PKG_`.

Tested on `x86_64`

### Example:
```ruby
# an `PKG_` prefix is not needed in the package recipe's filename
# /usr/local/lib/crew/packages/7_zip.rb

# an `PKG_` prefix is needed in the package's class name
class PKG_7_zip < Package
```

```
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=pkgname_start_with_number CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -181,10 +181,14 @@ def set_package(pkgName, pkgPath)
     puts "#{e.class}: #{e.message}".red
   end
 
-  # add `PKG_` to the beginning of pkgName if it starts with a digit
-  pkgName = 'PKG_' + pkgName if pkgName =~ /^[0-9]/
+  # add `PKG_` to the beginning of class name if it starts with a digit
+  if pkgName =~ /^[0-9]/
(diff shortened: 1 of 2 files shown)


uberhacker:
@supechicken: I'm liking these updates!  In the future, can you please include more descriptive commit messages instead of just `Update crew` or `Update const.rb`?

supechicken:
> @supechicken: I'm liking these updates! In the future, can you please include more descriptive commit messages instead of just `Update crew` or `Update const.rb`?

Okay :)

PR #6772, opened 2022-02-23
Update core_packages.txt

supechicken:


supechicken:
> Thank you for the update @supechicken. Can you please sort the packages in alphabetical order? This is so we can avoid duplicates.

Done

tools/core_packages.txt
```diff
@@ -1,74 +1,104 @@
...
+libssh
 libssp
 libtasn1
 libtirpc
+libunbound
 libunistring
 libxml2
+libxslt
 libyaml
+llvm
```
uberhacker, line 61: This is in core now?  Are you sure?  If it is, we need to remove it immediately.

tools/core_packages.txt
```diff
@@ -1,74 +1,104 @@
...
 py3_pip
 py3_setuptools
 py3_wheel
 python2
 python3
 readline
+rsync
 rtmpdump
 ruby
-sed
```
uberhacker, line 67: This should be core.  The out-of-the-box sed is worthless.

tools/core_packages.txt
```diff
@@ -1,74 +1,104 @@
...
+docbook_xsl_nons
 e2fsprogs
 elfutils
 expat
 filecmd
 flex
-gdbm
+gcc
 gcc11
+gdbm
```
uberhacker, line 23: Good call.  gdbm was out of sequence.

tools/core_packages.txt
supechicken: The list is generated by `crew deps core`, so it might be correct?

tools/core_packages.txt
supechicken: AFAIK `sed` was moved away to `buildessential` times ago (only `patch` and `build` section needs it)

tools/core_packages.txt
```diff
@@ -1,74 +1,104 @@
-acl
-attr
-binutils
```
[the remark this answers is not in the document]
supechicken: See #6642

tools/core_packages.txt
```diff
@@ -1,74 +1,104 @@
...
 brotli
 bz2
-c_ares
 ca_certificates
+c_ares
```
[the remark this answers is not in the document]
supechicken: Seems the `sort` command doesn't think so 😅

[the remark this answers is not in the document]
supechicken: Fixed

tools/core_packages.txt
uberhacker: And how many packages use it?  About half?  I think we should add it back.

tools/core_packages.txt
uberhacker: We need to remove it then.  There should be no reason to include this in core.  Not only is it massive but it's only needed for builds so it really should be in buildessential.

tools/core_packages.txt
uberhacker: After further thought, it's fine to leave this in buildessential.  I think we should move llvm to buildessential, if possible.  That should NOT be in core.

satmandu: > After further thought, it's fine to leave this in buildessential. I think we should move llvm to buildessential, if possible. That should NOT be in core.

As soon as we can break out the llvm libraries into another package!

uberhacker: Requiring llvm in core is cringeworthy.

uberhacker:
I did a fresh install and added missing packages from this PR.  Packages I would like to make an effort to remove from core and ideally into buildessential instead in the future include:
```
autoconf
automake
cmake
gcc11
jdk8
llvm
make
patch
patchelf
pkgconfig
valgrind
```

PR #6774, opened 2022-02-24
Add upm package

uberhacker:
Universal Password Manager (UPM) allows you to store usernames, passwords, URLs and generic notes in an encrypted database protected by one master password.  See http://upm.sourceforge.net.  Tested on all architectures.  Only aarch64, armv7l and x86_64 work, however.

### DIFF
--- a/packages/upm.rb
+++ b/packages/upm.rb
@@ -1,0 +1,30 @@
+require 'package'
+
+class Upm < Package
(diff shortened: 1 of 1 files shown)


satmandu:
Do we need to leave a note about where the database is stored? This is the sort of file that's going to be important to back up in some manner, no?

uberhacker:
Good point.  Also we should probably offer to delete the database on removal.  I'll work on it.

PR #6796, opened 2022-02-26
Mysqltuner 1.7.21 => 1.8.3

uberhacker:


satmandu (commented):
Would it make sense to just have multiple `source_url` files recognized in crew instead of working around that with `source_urls`?

uberhacker:
There is a way to simplify it, however.  I'll work on it.

satmandu:
I don't have a problem with the setup here at all. I just wish this was a part of crew so we could have multiple downloads for a given package...

PR #6803, opened 2022-02-28
Add volume-key package

Zopolis4:
Adds the volume-key package. Depends on #6805, and is part of the the same larger package as it.

Works properly:
- [x] armv7l

### DIFF
--- a/packages/volume_key.rb
+++ b/packages/volume_key.rb
@@ -1,0 +1,35 @@
+require 'package'
+
+class Volume_key < Package
+  description 'The volume_key project provides a libvolume_key, a library for manipulating
+storage volume encryption keys and storing them separately from volumes, and an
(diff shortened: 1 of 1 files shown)


satmandu:
Please use more of the PR template next time...

packages/volume_key.rb
```diff
@@ -0,0 +1,48 @@
...
+  depends_on 'nss' => :build
+  depends_on 'swig' => :build
+
+  def self.build
+    system 'autopoint'
+    system 'aclocal'
+    system 'libtoolize'
+    system 'autoconf'
+    system 'autoheader'
+    system 'automake --add-missing'
```
satmandu, line 40: FYI I think most of this can just be replaced with `system 'autoreconf -fvi'`  .

PR #6808, opened 2022-03-02
Remove Bintray and Sourceforge images

Zopolis4:
These images were added a long time ago, and are no longer relevant as neither of these platforms are used for hosting binaries anymore.


uberhacker:
Good call.

PR #6810, opened 2022-03-02
Add ndctl package

Zopolis4:
## Description
Adds the ndctl package (part of the same larger package as #6766).

## Additional information
This depends on #6809

- also updates `libkmod`
- build is dependent upon using a fixed `docbook_xsl` and also a rebuilt `asciidoctor`

Works properly:
- [x] x86_64
- [x] armv7l
- [ ] i686 (kernel too old to have a `linux/ndctl.h`)

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=ndctl CREW_TESTING=1 crew update
```
---

### DIFF
--- a/packages/ndctl.rb
+++ b/packages/ndctl.rb
@@ -1,0 +1,32 @@
+require 'package'
+
+class Ndctl < Package
+  description 'Helper tools and libraries for managing the libnvdimm (non-volatile memory device) sub-system in the Linux kernel'
+  homepage 'https://github.com/pmem/ndctl'
+  version 'v72.1'    
+  license 'GPL-2.0'
+  compatibility 'all'
+  source_url 'https://github.com/pmem/ndctl/archive/refs/tags/v72.1.tar.gz'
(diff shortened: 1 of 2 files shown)


packages/ndctl.rb
```diff
@@ -0,0 +1,32 @@
...
+  depends_on 'asciidoctor' => :build
+  depends_on 'bash_completion' => :build
+  depends_on 'iniparser' => :build
+  depends_on 'jsonc' => :build
+  depends_on 'keyutils' => :build
+  depends_on 'xmlto' => :build
+  
+  def self.build
+    system "./autogen.sh"
+    system "./configure CFLAGS='-g -O2' #{CREW_OPTIONS} --without-systemd"
```
satmandu, line 21: CFLAGS are automatically added. See https://github.com/skycocker/chromebrew/blob/master/lib/const.rb

satmandu:
Fixed up dependency issues, and added an update to `libkmod`, which needs to be built first once we fix `docbook_xsl`...

satmandu:
Ok. Binaries added. `i686` is incompatible because the kernels are too old.

PR #6836, opened 2022-03-07
Remove build dependencies found in buildessential

Zopolis4:
Since the dependencies removed are already in buildessential, their removal should make no difference and clean up the codebase a bit.

### DIFF
--- a/packages/amtk.rb
+++ b/packages/amtk.rb
@@ -24,8 +24,6 @@ class Amtk < Package
 
   depends_on 'gtk3'
   depends_on 'gobject_introspection' => :build
-  depends_on 'gtk_doc' => :build
-  depends_on 'llvm' => :build
 
 
   def self.build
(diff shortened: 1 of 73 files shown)


supechicken666 (approved):

satmandu:
One question here: does this change make it easier or harder to tell what dependencies are needed for a package?

For instance, "what packages need python2"?

I think this change makes it harder.

It is _useful_ to be able to tell which packages are needed for a build, even if those are buildessential packages.

Otherwise we have the situation of not being able to tell which packages can be removed from buildessential as unnecessary.

Zopolis4:
I think most of the codebase conforms to the idea of not adding build dependencies listed in buildessential, given that i only found ~40 offenders out of ~1000

satmandu:
That's absolutely true, and I think that's been a big problem over time. Until we fix our dependency algorithm, I don't think we should remove any data which might be needed to build a new dependency algorithm...

After that happens...

The simple fact is that we NEED to have that info about which packages depend upon which package somewhere. Deleting this info from these packages means that information just isn't around any more.

PR #6839, opened 2022-03-09
GCC11 rebuilt w/ mold support, add mold to buildessentials, add libst…

satmandu:
- `gcc11` rebuilt w/ `mold` support (i.e. support for `-fuse-ld=mold` which has also been accepted into GCC12), off of the GCC 11 stable branch.
- `mold` added to `buildessentials` for non-armv7l architectures (adds ~ 3mb, but may be order of magnitude faster than existing linkers)
- Add `libstdc++.so.6` to1 `CREW_ESSENTIAL_FILES`, since rdfind uses it.
- This does _NOT_ replace `-fuse-ld=gold` with `-fuse-ld=mold` in `const.rb`, but that would be the next step, once mold is updated to support `armv7l`.

- Also fixes circular dependencies in `graphite`, `freetype`, `chafa`, and `harfbuzz`, with `freetype` modified to use `mold` except on `armv7l` due to  upstream issue reported with `-ld-fuse=mold` https://github.com/rui314/mold/issues/389
- Updates packages which use `freetype_sub` to have dependency on `harfbuzz`.
- Updates `gtk3` and `gtk4` to fix `:installed_packages` algorithm used to be in line with what works inside `crew`

Works properly:
- [x] x86_64 
- [x] armv7l
- [x] i686

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=gcc_mold CREW_TESTING=1 crew update ; crew upgrade
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -220,5 +220,6 @@ CREW_LAST_PACKAGES = %w[ghc mandb gtk3 gtk4 sommelier]
 
 CREW_ESSENTIAL_FILES = `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/ruby`.scan(/\t([^ ]+)/).flatten +
                        `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/rsync`.scan(/\t([^ ]+)/).flatten +
+                       `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/rdfind`.scan(/\t([^ ]+)/).flatten +
                        %w[libzstd.so.1]
 CREW_ESSENTIAL_FILES.uniq!
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -220,6 +220,5 @@ CREW_LAST_PACKAGES = %w[ghc mandb gtk3 gtk4 sommelier]
 
 CREW_ESSENTIAL_FILES = `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/ruby`.scan(/\t([^ ]+)/).flatten +
                        `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/rsync`.scan(/\t([^ ]+)/).flatten +
-                       `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/rdfind`.scan(/\t([^ ]+)/).flatten +
-                       %w[libzstd.so.1]
+                       %w[libzstd.so.1 libstdc++.so.6]
 CREW_ESSENTIAL_FILES.uniq!
--- a/packages/bz2.rb
+++ b/packages/bz2.rb
@@ -12,14 +12,14 @@ class Bz2 < Package
   binary_url({
     aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8-1_armv7l/bz2-1.0.8-1-chromeos-armv7l.tar.zst',
      armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8-1_armv7l/bz2-1.0.8-1-chromeos-armv7l.tar.zst',
-    i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8-1_i686/bz2-1.0.8-1-chromeos-i686.tar.zst',
-  x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8-1_x86_64/bz2-1.0.8-1-chromeos-x86_64.tar.zst'
+       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8-1_i686/bz2-1.0.8-1-chromeos-i686.tar.zst',
+     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8-1_x86_64/bz2-1.0.8-1-chromeos-x86_64.tar.zst'
   })
   binary_sha256({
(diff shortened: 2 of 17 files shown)


lib/const.rb
```diff
@@ -220,5 +220,6 @@
 
 CREW_ESSENTIAL_FILES = `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/ruby`.scan(/\t([^ ]+)/).flatten +
                        `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/rsync`.scan(/\t([^ ]+)/).flatten +
+                       `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/rdfind`.scan(/\t([^ ]+)/).flatten +
```
supechicken, line 223: The `rdfind` command is provided by `buildessential`, maybe move this check to `crew`? (check when `rdfind` is exist and being called)

satmandu: `rdfind` is used during package installs. without this bit, during a gcc11 install there is an error about a `libstdc++.so.6` missing, which will happen if any package has ever been built, because then rdfind is installed.

(rdfind is also only ~ 80k)

satmandu: Actually... no I think you are right... let me just add the `libstdc++.so.6` dep to CREW_ESSENTIAL FILES.

packages/freetype_sub.rb
```diff
@@ -3,44 +3,66 @@
...
 
   def self.install
-    ENV['CREW_CONFLICTS_ONLY_ADVISORY'] = '1'
-    reload_constants
-    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    system "DESTDIR=#{CREW_DEST_DIR} samu -C builddir install"
   end
 
   def self.postinstall
     system "find #{CREW_BREW_DIR}/* -name freetype*.tar | xargs rm -rf"  # make sure to delete downloaded files
```
supechicken, line 55:
```ruby
system "find #{CREW_BREW_DIR}/* -name freetype*.tar -exec rm -rf {} \+"
```

packages/freetype.rb
```diff
@@ -3,40 +3,56 @@
...
   description 'FreeType is a freely available software library to render fonts.'
   homepage 'https://www.freetype.org/'
-  version '2.11.0'
+  version '2.11.1'
   license 'FTL or GPL-2+'
   compatibility 'all'
-  source_url 'https://download.savannah.gnu.org/releases/freetype/freetype-2.11.0.tar.xz'
-  source_sha256 '8bee39bd3968c4804b70614a0a3ad597299ad0e824bc8aad5ce8aaf48067bde7'
+  source_url 'https://gitlab.freedesktop.org/freetype/freetype.git'
+  git_hashtag "VER-#{version.gsub('.', '-')}"
```
supechicken, line 10: `.tr` is better (and a *little* bit faster) in single word global substitution:
```ruby
git_hashtag "VER-#{version.tr('.', '-')}"
```

satmandu: There are a bunch of packages which have that. Feel free to use sed to change that in multiple packages after we get this merged. ;)

packages/bz2.rb
```diff
@@ -3,61 +3,71 @@
...
 
-  binary_url ({
-    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8_armv7l/bz2-1.0.8-chromeos-armv7l.tar.xz',
-     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8_armv7l/bz2-1.0.8-chromeos-armv7l.tar.xz',
-       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8_i686/bz2-1.0.8-chromeos-i686.tar.xz',
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8_x86_64/bz2-1.0.8-chromeos-x86_64.tar.xz',
+  binary_url({
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8-1_armv7l/bz2-1.0.8-1-chromeos-armv7l.tar.zst',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8-1_armv7l/bz2-1.0.8-1-chromeos-armv7l.tar.zst',
+    i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/bz2/1.0.8-1_i686/bz2-1.0.8-1-chromeos-i686.tar.zst',
```
supechicken, line 15: Malformed layout?

packages/freetype_sub.rb
```diff
@@ -1,46 +1,13 @@
 require 'package'
 
 class Freetype_sub < Package
-  description 'Freetype_sub is a version without harfbuzz. It is intended to handle circular dependency betwwen freetype and harfbuzz.'
+  description 'Freetype_sub was a version w/o harfbuzz, now handled within harfbuzz'
   homepage 'https://www.freetype.org/'
-  version '2.11.0'
+  version '1.0'
```
supechicken, line 6: Follow `freetype` version? (otherwise this may brake `crew update`)

packages/bz2.rb
satmandu: Fixed.

packages/freetype_sub.rb
satmandu: This is intentional, to force anyone with freetype_sub installed to be updated to the full freetype version.

satmandu: (Freetype_sub is no longer a dependency for any package.)

supechicken: Oh it should be "break"😅

satmandu: As far as I can tell, `crew update` doesn't do monotonic versioning. It just cares that the version has changed _at all_ from what it was before.

PR #6841, opened 2022-03-10
Fix `=> :build` tags on some packages

supechicken:
The `:build` tag should be a symbol and no quotes are needed, extra quotes were removed.


uberhacker:
Good call.

PR #6842, opened 2022-03-10
Make mold default x86 linker, plus harfbuzz + initial libpng cleanup

satmandu:
- make mold linker default for x86, leave gold as default for arm
- Fix git cached builds also caching all tags in crew
- libpng is now generated by harfbuzz as part of that package's build (yes this removes the animated png patch @saltedcoffii but it makes the builds of many other packages simpler.)
- ragel is now generated by harfbuzz as part of that package's build, 
- update libpng dependent packages, which means a whole bunch of x11 and gnome stuff, including `gtk3` and `gtk4`

Works properly:
- [x] x86_64
- [x] armv7l
- [x] i686

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=libpng CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.23.0'
+CREW_VERSION = '1.23.1'
 
 ARCH_ACTUAL = `uname -m`.chomp
 # This helps with virtualized builds on aarch64 machines
@@ -115,8 +115,14 @@ when 'x86_64'
   CREW_BUILD = 'x86_64-cros-linux-gnu'
 end
 
-CREW_COMMON_FLAGS = '-O2 -pipe -flto -ffat-lto-objects -fPIC -fuse-ld=gold'
-CREW_COMMON_FNO_LTO_FLAGS = '-O2 -pipe -fno-lto -fPIC -fuse-ld=gold'
+case ARCH
+when 'aarch64', 'armv7l'
+  CREW_LINKER = 'gold'
+when 'i686', 'x86_64'
+  CREW_LINKER = 'mold'
+end
+CREW_COMMON_FLAGS = "-O2 -pipe -flto -ffat-lto-objects -fPIC -fuse-ld=#{CREW_LINKER}"
+CREW_COMMON_FNO_LTO_FLAGS = "-O2 -pipe -fno-lto -fPIC -fuse-ld=#{CREW_LINKER}"
 CREW_LDFLAGS = '-flto'
 CREW_FNO_LTO_LDFLAGS = '-fno-lto'
 
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -115,7 +115,7 @@ when 'x86_64'
   CREW_BUILD = 'x86_64-cros-linux-gnu'
 end
 
-unless ENV['CREW_LINKER'].to_s.empty?
+if ENV['CREW_LINKER'].to_s.empty?
   case ARCH
   when 'aarch64', 'armv7l'
     CREW_LINKER = 'gold'
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -119,15 +119,18 @@ unless ENV['CREW_LINKER'].to_s.empty?
   case ARCH
   when 'aarch64', 'armv7l'
     CREW_LINKER = 'gold'
+    CREW_LINKER_FLAGS = "-Wl,--threads -Wl,--thread-count,#{CREW_NPROC}"
   when 'i686', 'x86_64'
     CREW_LINKER = 'mold'
+    CREW_LINKER_FLAGS = ''
   end
 else
   CREW_LINKER = ENV['CREW_LINKER']
 end
-CREW_COMMON_FLAGS = "-O2 -pipe -flto -ffat-lto-objects -fPIC -fuse-ld=#{CREW_LINKER}"
-CREW_COMMON_FNO_LTO_FLAGS = "-O2 -pipe -fno-lto -fPIC -fuse-ld=#{CREW_LINKER}"
-CREW_LDFLAGS = '-flto'
+
+CREW_COMMON_FLAGS = "-O2 -pipe -flto -ffat-lto-objects -fPIC -fuse-ld=#{CREW_LINKER} #{CREW_LINKER_FLAGS}"
+CREW_COMMON_FNO_LTO_FLAGS = "-O2 -pipe -fno-lto -fPIC -fuse-ld=#{CREW_LINKER} #{CREW_LINKER_FLAGS}"
+CREW_LDFLAGS = "-flto #{CREW_LINKER_FLAGS}"
 CREW_FNO_LTO_LDFLAGS = '-fno-lto'
 
 unless CREW_DISABLE_ENV_OPTIONS
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -215,8 +215,8 @@ PY2_SETUP_BUILD_OPTIONS = "--executable=#{CREW_PREFIX}/bin/python2"
 PY_SETUP_INSTALL_OPTIONS_NO_SVEM = "--root=#{CREW_DEST_DIR} --prefix=#{CREW_PREFIX} -O2 --compile"
 PY_SETUP_INSTALL_OPTIONS = "#{PY_SETUP_INSTALL_OPTIONS_NO_SVEM} --single-version-externally-managed"
 
-CREW_FIRST_PACKAGES = %w[libssh curl git pixz shared_mime_info]
-CREW_LAST_PACKAGES = %w[ghc mandb gtk3 gtk4 sommelier]
+CREW_FIRST_PACKAGES = %w[libssh curl git pixz shared_mime_info harfbuzz]
+CREW_LAST_PACKAGES = %w[ghc mandb gtk3 gtk4 sommelier freetype]
 
 CREW_ESSENTIAL_FILES = `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/ruby`.scan(/\t([^ ]+)/).flatten +
                        `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/rsync`.scan(/\t([^ ]+)/).flatten +
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -115,11 +115,15 @@ when 'x86_64'
   CREW_BUILD = 'x86_64-cros-linux-gnu'
 end
 
-case ARCH
-when 'aarch64', 'armv7l'
-  CREW_LINKER = 'gold'
-when 'i686', 'x86_64'
-  CREW_LINKER = 'mold'
+unless ENV['CREW_LINKER'].to_s.empty?
+  case ARCH
+  when 'aarch64', 'armv7l'
+    CREW_LINKER = 'gold'
+  when 'i686', 'x86_64'
+    CREW_LINKER = 'mold'
+  end
+else
+  CREW_LINKER = ENV['CREW_LINKER']
 end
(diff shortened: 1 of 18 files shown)


satmandu:
Currently building:
fontconfig.rb cairo.rb pango.rb gdk_pixbuf.rb gtk3.rb gtk4.rb

lib/const.rb
```diff
@@ -115,8 +115,14 @@
...
 end
 
-CREW_COMMON_FLAGS = '-O2 -pipe -flto -ffat-lto-objects -fPIC -fuse-ld=gold'
-CREW_COMMON_FNO_LTO_FLAGS = '-O2 -pipe -fno-lto -fPIC -fuse-ld=gold'
+case ARCH
+when 'aarch64', 'armv7l'
+  CREW_LINKER = 'gold'
+when 'i686', 'x86_64'
+  CREW_LINKER = 'mold'
+end
```
uberhacker, line 123: Is this something we could override with an environment variable?

satmandu: I intentionally labeled it `CREW_LINKER` with that in mind...

lib/const.rb
```diff
@@ -215,8 +221,8 @@
 PY_SETUP_INSTALL_OPTIONS_NO_SVEM = "--root=#{CREW_DEST_DIR} --prefix=#{CREW_PREFIX} -O2 --compile"
 PY_SETUP_INSTALL_OPTIONS = "#{PY_SETUP_INSTALL_OPTIONS_NO_SVEM} --single-version-externally-managed"
 
-CREW_FIRST_PACKAGES = %w[libssh curl git pixz shared_mime_info]
-CREW_LAST_PACKAGES = %w[ghc mandb gtk3 gtk4 sommelier]
+CREW_FIRST_PACKAGES = %w[libssh curl git pixz shared_mime_info harfbuzz]
+CREW_LAST_PACKAGES = %w[ghc mandb gtk3 gtk4 sommelier freetype]
```
uberhacker, line 225: The last package should be sommelier to prevent the postinstall instructions for scrolling past the screen.

packages/libxdmcp.rb
```diff
@@ -3,30 +3,31 @@
...
+    aarch64: '591a059b27c7095f45adf57a4d1ac4b045fb2e6be8ea55bac8aa2b2eef1b3c16',
+     armv7l: '591a059b27c7095f45adf57a4d1ac4b045fb2e6be8ea55bac8aa2b2eef1b3c16',
+       i686: 'a91f416ab907be4f4183b4c5da505e4d25586736b22185e3f58b64c36a1f9ca1',
+     x86_64: '553304325808a09bc564a989a1e727046000b892ce75d1dec437df7d67ade648'
   })
 
   depends_on 'xorg_proto'
 
   def self.build
     system './autogen.sh'
```
uberhacker, line 29: Add `NOCONFIGURE=1`?

lib/const.rb
satmandu: But you can always do a `no_env_options` and then do a `CREW_ENV_OPTIONS.tr('mold','lld')` or whatever...

uberhacker: Too hackish. :)

lib/const.rb
uberhacker: Do we even need this now with the new dependency resolver?

satmandu: I haven't tested it yet, but I will say that harfbuzz needs to be after freetype, otherwise the more limited freetype files get left on the system... I don't see how the new resolver fixes that.  @supechicken?

supechicken: I remember that this is a part of mitigations and I have completely removed it. Since the post-install part will run after all package's installation now, we can already ensure that the user won't miss any information

lib/const.rb
```diff
@@ -115,8 +115,14 @@
   CREW_BUILD = 'x86_64-cros-linux-gnu'
 end
 
-CREW_COMMON_FLAGS = '-O2 -pipe -flto -ffat-lto-objects -fPIC -fuse-ld=gold'
-CREW_COMMON_FNO_LTO_FLAGS = '-O2 -pipe -fno-lto -fPIC -fuse-ld=gold'
+case ARCH
+when 'aarch64', 'armv7l'
+  CREW_LINKER = 'gold'
```
supechicken, line 120: Enable multithread support for `gold` by appending the following to `*FLAGS`? (theoretically would be slightly faster, but I don't see any difference in my system😅, but should be better than none)
```
-Wl,--threads -Wl,--thread-count,8
```

lib/const.rb
satmandu: Looks like it is still in crew @supechicken ?

```
    # Adjust package install ordering for upgrades.
    CREW_FIRST_PACKAGES.each do |pkg|
      # Add package to beginning.
      toBeUpdated.insert(0,toBeUpdated.delete(pkg)) if toBeUpdated.include? pkg
    end
    CREW_LAST_PACKAGES.each do |pkg|
      if toBeUpdated.include? pkg
        # Add package to beginning.
        toBeUpdated.insert(0,toBeUpdated.delete(pkg))
        # Now rotate first package to last package.
        toBeUpdated=toBeUpdated.rotate(1)
      end
    end
```

supechicken: It is in `def upgrade`, which I forgot to remove😅

lib/const.rb
satmandu: is the new logic better?

lib/const.rb
satmandu: Go ahead and modify this PR to make those changes?

lib/const.rb
supechicken: Like this:
```ruby
unless ENV['CREW_LINKER'].to_s.empty?
  case ARCH
  when 'aarch64', 'armv7l'
    CREW_LINKER = 'gold'
    CREW_LINKER_FLAGS = "-Wl,--threads -Wl,--thread-count,#{CREW_NPROC}"
  when 'i686', 'x86_64'
    CREW_LINKER = 'mold'
    CREW_LINKER_FLAGS = ''
  end
else
  CREW_LINKER = ENV['CREW_LINKER']
end

CREW_COMMON_FLAGS = "-O2 -pipe -flto -ffat-lto-objects -fPIC -fuse-ld=#{CREW_LINKER} #{CREW_LINKER_FLAGS}"
CREW_COMMON_FNO_LTO_FLAGS = "-O2 -pipe -fno-lto -fPIC -fuse-ld=#{CREW_LINKER} #{CREW_LINKER_FLAGS}"
CREW_LDFLAGS = "-flto #{CREW_LINKER_FLAGS}"
CREW_FNO_LTO_LDFLAGS = '-fno-lto'
```

satmandu: Do those work with llvm?

If not, that could cause problems with packages which use clang.

supechicken: Yes, according to the reference of `clang`:
```
-Wl,<arg>,<arg2>...[¶](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-wl-arg-arg2)
Pass the comma separated arguments in <arg> to the linker
```

satmandu: I get this error on i686:
```
/usr/local/lib/crew/lib/const.rb:131:in `<top (required)>': uninitialized constant CREW_LINKER_FLAGS (NameError)                                                                CREW_COMMON_FLAGS = "-O2 -pipe -flto -ffat-lto-objects -fPIC -fuse-ld=#{CREW_LINKER} #{CREW_LINKER_FLAGS}"
                                                                                       ^^^^^^^^^^^^^^^^^
Did you mean?  CREW_LIB_PATH
        from /usr/local/bin/crew:12:in `require_relative'
        from /usr/local/bin/crew:12:in `<main>'
```
Why is setting `CREW_LINKER_FLAGS` to an empty string causing a problem?

supechicken: We should also set the flags under the `CREW_LINKER = ENV['CREW_LINKER']` part😅

satmandu: Actually, this was my fault. I had the logic backward!

packages/glew.rb
```diff
@@ -3,29 +3,29 @@
...
-  binary_url ({
-    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/glew/2.2.0-rc3_armv7l/glew-2.2.0-rc3-chromeos-armv7l.tar.xz',
-     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/glew/2.2.0-rc3_armv7l/glew-2.2.0-rc3-chromeos-armv7l.tar.xz',
-       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/glew/2.2.0-rc3_i686/glew-2.2.0-rc3-chromeos-i686.tar.xz',
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/glew/2.2.0-rc3_x86_64/glew-2.2.0-rc3-chromeos-x86_64.tar.xz',
+  binary_url({
+    aarch64: 'file:///usr/local/tmp/packages/glew-2.2.0-chromeos-armv7l.tar.zst',
+     armv7l: 'file:///usr/local/tmp/packages/glew-2.2.0-chromeos-armv7l.tar.zst',
+       i686: 'file:///usr/local/tmp/packages/glew-2.2.0-chromeos-i686.tar.zst',
+     x86_64: 'file:///usr/local/tmp/packages/glew-2.2.0-chromeos-x86_64.tar.zst'
```
uberhacker, line 16: Local files?

satmandu: fixed.

packages/glew.rb
```diff
@@ -34,17 +34,10 @@ def self.patch
   end
 
   def self.build
-    #ENV['CFLAGS.EXTRA'] = "-fuse-ld=lld"
-    #ENV['CFLAGS'] = "-fuse-ld=lld"
-    #ENV['CXXFLAGS'] = "-fuse-ld=lld"
-    ENV['GLEW_DEST'] = "#{CREW_PREFIX}"
+    ENV['GLEW_DEST'] = CREW_PREFIX.to_s
```
supechicken, line 37: `.to_s` should not be required

packages/fontconfig.rb
```diff
@@ -38,7 +39,7 @@ def self.build
       import sys
       sys.exit()
```
supechicken, line 40: Instead of calling the exit function manually, just left the file empty will be faster :)

packages/glew.rb
satmandu: changed.

packages/fontconfig.rb
satmandu: It's a hack to work around install issues. If I recall correctly this way we bypass getting errors that stop the install from succeeding.

satmandu: I do wish there was a better way. Feel free to try alternatives!

supechicken: Oh, mostly because this file is imported as a module so we need to exit manually

packages/libpng.rb
```diff
@@ -1,49 +1,13 @@
 require 'package'
 
 class Libpng < Package
-  description 'libpng is the official PNG reference library. Patched with APNG support.'
+  description 'libpng is the official PNG reference library. Now built inside harfbuzz'
```
supechicken, line 4: "Now bundled with harfbuzz"?

packages/ragel.rb
```diff
@@ -1,33 +1,13 @@
 require 'package'
 
 class Ragel < Package
-  description 'Ragel compiles executable finite state machines from regular languages.'
+  description 'Ragel compiles executable finite state machines from regular languages. Now built inside harfbuzz'
```
supechicken, line 4: "Now bundled with harfbuzz"?

packages/libpng.rb
satmandu: changed

packages/ragel.rb
satmandu: changed

PR #6848, opened 2022-03-11
lib/deb_utils.rb: Improve `.deb` unarchive algorithm

supechicken:
### About
In `lib/deb_utils.rb`, all contents in the `.deb` archive should be files (folders are not supported in `ar` format archive), so any trailing slash in the filename is not expected (and no reason to be appended), but some packages are exceptions:
```
$ head -n4 signal-desktop_5.34.0_amd64.deb 
!<arch>                  
debian-binary/  0           0     0     644     4         `
2.0
control.tar.gz/ 0           0     0     644     6020      `
```
(all files in this archive have an `/` suffix, which should not be added)

### Changes
- Automatically remove `/` suffix for filenames (if any)

```
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=deb_utils_fix CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/deb_utils.rb
+++ b/lib/deb_utils.rb
@@ -15,7 +15,7 @@ module DebUtils
     # get first line of the given file, should be a signature string (`!<arch>\n`) if it is a valid deb file
     signature = src_fileIO.gets
 
-    abort 'Malformed archive :/' unless signature == "!<arch>\n"
+    abort 'Malformed archive :/'.lightred unless signature == "!<arch>\n"
 
     # process each file in archive
     while (line = src_fileIO.gets) do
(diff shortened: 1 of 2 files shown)


lib/deb_utils.rb
```diff
@@ -23,8 +23,12 @@ def self.extract_deb(file, target)
       name, modtime, uid, gid, mode, size, end_char = \
         line.scan(/(.{16})(.{12})(.{6})(.{6})(.{8})(.{10})(.{1})/).flatten.map(&:strip)
 
+      # remove slash suffix from filename (if any)
+      # (a `.deb` ar archive does not support any directories, so we can confirm that all entries are normal files)
+      name.sub!(/\/$/, '')
+
       # check ending byte
-      abort "Malformed archive :/" unless end_char == '`'
+      abort 'Malformed archive :/' unless end_char == '`'
```
uberhacker, line 31: Add color?

PR #6872, opened 2022-03-17
meson => 0.61.3

saltedcoffii:
Changes:
- Upgrade meson

```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=meson_0.61.3 CREW_TESTING=1 crew update
```

Needs armv7l binaries, however meson is arch independent, so maybe we don't need to compile more, we can just use the x86_64 or i686 ones? Sidenote: chromebrew should have a way to handle arch independent packages.


saltedcoffii:
@uberhacker arm binaries? 🥺

PR #6873, opened 2022-03-17
envvar to control use of samurai or ninja

saltedcoffii:
Changes:
- Recompile samurai to use the latest CREW_COMMON_FLAGS, etc.
- Create an environment variable called CREW_NINJA to control use of samurai or ninja
- Create a chromebrew constant called CREW_NINJA controlled by the environment variable

Example use case:
```ruby
def self.build
  Dir.mkdir "builddir"
  Dir.chdir "builddir" do
    system "cmake -G Ninja #{CREW_COMMON_FLAGS} .."
  end
  system "#{CREW_NINJA} -C builddir"
end

def self.install
  system "#{CREW_NINJA} -C builddir install"
end

def self.check
  system "#{CREW_NINJA} -C builddir test"
end
```

Run the following to get this pull request's changes locally for testing:
```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=samuorninja CREW_TESTING=1 crew update
```

samurai needs armv7l binaries.

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -197,6 +197,13 @@ CREW_MESON_FNO_LTO_OPTIONS = <<~OPT.chomp
   -Dc_args='-O2'
 OPT
 
+# Use ninja or samurai
+if ENV['CREW_NINJA'].to_s.downcase == 'ninja'
+  CREW_NINJA = 'ninja'
+else
+  CREW_NINJA = 'samu'
+end
+
 # Cmake sometimes wants to use LIB_SUFFIX to install libs in LIB64, so specify such for x86_64
 # This is often considered deprecated. See discussio at https://gitlab.kitware.com/cmake/cmake/-/issues/18640
 # and also https://bugzilla.redhat.com/show_bug.cgi?id=1425064
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.23.2'
+CREW_VERSION = '1.23.3'
 
 ARCH_ACTUAL = `uname -m`.chomp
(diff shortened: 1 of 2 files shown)


lib/const.rb
```diff
@@ -197,6 +197,13 @@
   -Dc_args='-O2'
 OPT
 
+# Use ninja or samurai
+if ENV['CREW_NINJA'].to_s.downcase == 'ninja'
+  CREW_NINJA = 'ninja'
+else
+  CREW_NINJA = 'samu'
+end
+
```
uberhacker, line 206: Bump version in this file too?

saltedcoffii: aa94d03

packages/samurai.rb
```diff
@@ -28,12 +25,16 @@ def self.patch
   end
 
   def self.build
-    system "make CFLAGS='-pipe -O2 -flto=auto -fuse-ld=gold' \
-                LDFLAGS='-flto=auto -fuse-ld=gold' \
+    system "make CFLAGS='#{CREW_COMMON_FLAGS}' \
```
supechicken, line 28: Use `CREW_ENV_OPTIONS`?

supechicken:
Does there something missing (in function) in `samu` but in `ninja`? If not, I think it is better to force users to use `samu`

saltedcoffii:
the only missing feature of samu that's in ninja is job control based on load average. chromebrew does not implement this feature.

packages/samurai.rb
saltedcoffii: aren't crew env options automatically added

saltedcoffii:
Merging because 2/3 reviewers approved

uberhacker:
I don't see any approvals yet and we need arm binaries.

PR #6878, opened 2022-03-19
Fix downloader for file:// scheme

### LINKED ISSUE #6877: Error in downloader with file:// scheme
```
/usr/local/lib/crew/lib/downloader.rb:43:in `downloader': undefined local variable or method `filename' for main:Object (NameError)

        return FileUtils.cp uri.path, filename
                                      ^^^^^^^^
```

uberhacker:
Fixes #6877.

### DIFF
--- a/lib/downloader.rb
+++ b/lib/downloader.rb
@@ -40,6 +40,11 @@ def downloader (*args)
     when 'file'
       # use FileUtils to copy if it is a local file (the url protocol is file://)
       if File.exist?(uri.path)
+        if args[1].to_s.empty?
+          filename = File.basename(uri.path)
+        else
+          filename = args[1]
+        end
         return FileUtils.cp uri.path, filename
       else
         abort "#{uri.path}: File not found :/".lightred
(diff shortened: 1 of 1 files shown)


lib/downloader.rb
```diff
@@ -40,6 +40,11 @@ def downloader (*args)
     when 'file'
       # use FileUtils to copy if it is a local file (the url protocol is file://)
       if File.exist?(uri.path)
+        if args[1].to_s.empty?
+          filename = File.basename(uri.path)
+        else
+          filename = args[1]
+        end
         return FileUtils.cp uri.path, filename
```
supechicken, line 48: If `args[1]` is a string, we don't need to check if it is valid (easier for debugging, also since this won't involve env variables, it is unlikely that an empty string will be passed instead of `nil` (why pass it if it is empty :) ) )

As above, we can combine the `if else` part with the `return` part:
```ruby
return FileUtils.cp uri.path, args[1] || File.basename(uri.path)
```

uberhacker: I'm all for simplifying...if it works. :)

PR #6890, opened 2022-03-20
sysinfo: Fix `not a git repository` error

supechicken:


lib/const.rb
```diff
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.23.3'
+CREW_VERSION = '1.23.3-1'
```
uberhacker, line 3: Can we just bump the lowest number instead of `-1`?

Issue #6886, opened 2022-03-20
Catch 22 installing Android Studio


adithv08:
## Description
Catch 22 installing Android Studio. jdk11 is a dependency but requires to be built from source. Building from source brings up overwriting issues with jdk8. Uninstalling jdk8 and building jdk11 installs jdk8 as a dependency. Forgetting all of that and installing Android studio and jdk11 as a dependency brings up 404 not found

## Versions and system information
fatal: not a git repository (or any parent up to mount point /media/fuse)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
- Architecture: `x86_64` (`x86_64`)
- Kernel version: `5.4.172-17736-gfddbc46e9604`

- Chromebrew version: `1.23.3`
- Chromebrew prefix: `/usr/local`
- Chromebrew libdir: `/usr/local/lib64`

- Last update in local repository: ``

- OS variant: `Chrome OS`
- OS version: `nami-release/R99-14469.41.0`
- OS channel: `stable-channel`
## Additional information
also updated using crew update

uberhacker:
Thank you for the feedback @adithv08.  This is definitely a bug because there are no dependencies in the jdk11 package.  @supechicken?

PR #6891, opened 2022-03-20
crew: Move checksum checking part to lib/downloader.rb

supechicken:
Tested on `x86_64`

Moved checksum checking part from `crew` to `lib/downloader`, so this can be used in packages also:
```ruby
# before
system 'curl -#Lo cf.bash https://raw.githubusercontent.com/cloudfoundry/cli/v6.36.1/ci/installers/completion/cf'
unless Digest::SHA256.hexdigest(File.read('cf.bash')) == 'f3f05a2414075c00b101b05f73cf260b9eec9966659adf2957c1b2937bd4c48e'
  abort 'Checksum mismatch. :/ Try again.'.lightred
end

# after
downloader 'https://raw.githubusercontent.com/cloudfoundry/cli/v6.36.1/ci/installers/completion/cf', 
 'f3f05a2414075c00b101b05f73cf260b9eec9966659adf2957c1b2937bd4c48e','cf.bash'
```

```
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=downloader_checksum CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -712,10 +712,12 @@ def download
         end
       end
       # Download file if not cached.
-      downloader url, filename, @opt_verbose
+      unless sha256sum =~ /^SKIP$/i
+        downloader url, filename, @opt_verbose, sha256sum: sha256sum
+      else
+        downloader url, filename, @opt_verbose
+      end
 
-      abort 'Checksum mismatch. 😔 Try again.'.lightred unless
-        Digest::SHA256.hexdigest( File.read(filename) ) == sha256sum || sha256sum =~ /^SKIP$/i
       puts "#{@pkg.name.capitalize} archive downloaded.".lightgreen
       # Stow file in cache if requested, if file is not from cache,
       # and cache is writable.
--- a/bin/crew
+++ b/bin/crew
@@ -712,11 +712,7 @@ def download
         end
       end
       # Download file if not cached.
-      unless sha256sum =~ /^SKIP$/i
-        downloader url, filename, @opt_verbose, sha256sum: sha256sum
-      else
-        downloader url, filename, @opt_verbose
-      end
+      downloader url, filename, sha256sum, @opt_verbose
 
       puts "#{@pkg.name.capitalize} archive downloaded.".lightgreen
       # Stow file in cache if requested, if file is not from cache,
--- a/bin/crew
+++ b/bin/crew
@@ -712,7 +712,7 @@ def download
         end
       end
       # Download file if not cached.
-      downloader url, filename, sha256sum, @opt_verbose
+      downloader url, sha256sum, filename, 3, @opt_verbose
 
       puts "#{@pkg.name.capitalize} archive downloaded.".lightgreen
       # Stow file in cache if requested, if file is not from cache,
--- a/bin/crew
+++ b/bin/crew
@@ -712,7 +712,7 @@ def download
         end
       end
       # Download file if not cached.
-      downloader url, sha256sum, filename, 3, @opt_verbose
+      downloader url, sha256sum, filename, @opt_verbose
 
       puts "#{@pkg.name.capitalize} archive downloaded.".lightgreen
       # Stow file in cache if requested, if file is not from cache,
@@ -736,10 +736,8 @@ def download
       @git = true
     else
       Dir.mkdir @extract_dir
-      downloader url, filename, @opt_verbose
+      downloader url, sha256sum, filename, @opt_verbose
 
-      abort 'Checksum mismatch. 😔 Try again.'.lightred unless
-        Digest::SHA256.hexdigest( File.read(filename) ) == sha256sum || sha256sum =~ /^SKIP$/i
       puts "#{filename}: File downloaded.".lightgreen
 
       FileUtils.mv filename, "#{@extract_dir}/#{filename}"
--- a/lib/downloader.rb
+++ b/lib/downloader.rb
@@ -28,8 +28,7 @@ end
 
 def downloader (url, sha256sum, filename = File.basename(url), verbose = false)
   # downloader: wrapper for all Chromebrew downloaders (`net/http`,`curl`...)
-  # Usage: downloader <url>, <sha256sum>, <filename::optional>, <retry_count::optional>,
-  #                   <verbose::optional>
+  # Usage: downloader <url>, <sha256sum>, <filename::optional>, <verbose::optional>
   #
   #           <url>: URL that points to the target file
   #     <sha256sum>: SHA256 checksum, verify downloaded file with given checksum
@@ -96,7 +95,7 @@ def http_downloader (url, filename = File.basename(url), verbose = false)
         redirect_uri.scheme ||= uri.scheme
         redirect_uri.host ||= uri.host
 
-        return downloader(redirect_uri.to_s, filename, verbose)
+        return send(__method__, redirect_uri.to_s, filename, verbose)
       else
         abort "Download failed with error #{response.code}: #{response.msg}".lightred
       end
--- a/lib/downloader.rb
+++ b/lib/downloader.rb
@@ -65,7 +65,7 @@ def downloader (url, sha256sum, filename = File.basename(url), retry_count = 0, 
   end
 
   # verify with given checksum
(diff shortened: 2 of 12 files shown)


lib/downloader.rb
```diff
@@ -26,11 +27,33 @@ def setTermSize
...
 
   uri = URI(args[0]) # read url from given params
+  filename = args[1] || File.basename(uri.path)
+
+  # verify with given checksum if checksum is provided in optional args (`downloader "url", sha256sum: "..."`)
+  if args[-1].is_a?(Hash) and args[-1].has_key?(:sha256sum)
+    check_sha256 = true
+    sha256sum = args[-1][:sha256sum]
+  else
+    check_sha256 = false
```
uberhacker, line 52: Shouldn't we do a sha256sum check on everything?  Throw an error instead?

supechicken: The real check is in lines 80-82

uberhacker: Right, but if `check_sha256` is false, those lines are skipped.

supechicken: Yeah, if `sha256sum` is not provided, then skip those checks

uberhacker: I don't think you understand what I'm suggesting.  If those checks are skipped then there is no way to verify if the download is corrupt.

uberhacker:
Fix all packages that use downloader with the old method?

lib/downloader.rb
supechicken: I mean if the `sha256sum` parameter is provided (should be provided by `crew` is sha256sum != `SKIP`), then `check_sha256` will be set to `true`, then the checksum check will be performed.

supechicken: (we don't have any way to verify the file if the `sha256sum` param is not provided :) )

supechicken:
> Fix all packages that use downloader with the old method?

I will make the changes in a separate pr later

lib/downloader.rb
uberhacker: Right, which is why I'm suggesting we make that a required parameter.  If the parameter is not provided then throw an error.

supechicken: The reason that I make it an optional parameter is for external use (like using for downloading in command line), and it takes lots of work to add a required parameter between the `<url>` and `<verbose>` arg :)

uberhacker: Don't we already have a `crew download` command?  I'm just wondering about package usage.  I'm not sure we should be using it on the command line because the use case is to check if files are downloaded correctly.

supechicken: Changed `sha256sum` to a required parameter (it is much simpler than I think...)

lib/downloader.rb
```diff
@@ -25,32 +26,47 @@ def setTermSize
   return true
 end
 
-def downloader (*args)
+def downloader (url, sha256sum, filename = File.basename(url), retry_count = 0, verbose = false)
+  # downloader: wrapper for all Chromebrew downloaders (`net/http`,`curl`...)
+  # Usage: downloader <url>, <filename::optional>, <retry_count::optional>,
+  #                   <verbose::optional>, sha256sum: <optional>
```
uberhacker, line 32: Change usage to mention sha256sum check is required?

bin/crew
```diff
@@ -712,10 +712,8 @@ def download
         end
       end
       # Download file if not cached.
-      downloader url, filename, @opt_verbose
+      downloader url, filename, sha256sum, @opt_verbose
```
uberhacker, line 715: Doesn't the option order matter?

uberhacker:
```
downloader url, sha256sum, filename, 0, @opt_verbose
```

packages/cunit.rb
```diff
@@ -24,12 +24,14 @@ class Cunit < Package
...
-      abort 'Checksum mismatch. :/ Try again.'.lightred
-    end
+    downloader 'https://httpredir.debian.org/debian/pool/main/c/cunit/cunit_2.1-3-dfsg-2.4.debian.tar.xz', 'e7a09a24c7db0e2aa9feb444fe38957286ebfc63b355c308957794f064b5881d'
+
     system 'tar xf cunit_2.1-3-dfsg-2.4.debian.tar.xz'
-    system "for i in \$(cat debian/patches/series); do patch -Np1 -i debian/patches/\${i}; done"
+
+    File.foreach 'debian/patches/series' do |patch|
+      system "patch -Np1 -i debian/patches/#{patch}"
+    end
```
uberhacker, line 33: Nice rubification!

packages/libspatialaudio.rb
```diff
@@ -28,10 +28,10 @@ class Libspatialaudio < Package
...
-      downloader 'https://github.com/greekgoddj/mit-hrtf-lib/raw/658657aebc58480d5511bd726b8dfa73b4f49ac7/source/normal/mit_hrtf_normal_44100.h'
-      downloader 'https://github.com/greekgoddj/mit-hrtf-lib/raw/658657aebc58480d5511bd726b8dfa73b4f49ac7/source/normal/mit_hrtf_normal_48000.h'
-      downloader 'https://github.com/greekgoddj/mit-hrtf-lib/raw/658657aebc58480d5511bd726b8dfa73b4f49ac7/source/normal/mit_hrtf_normal_88200.h'
-      downloader 'https://github.com/greekgoddj/mit-hrtf-lib/raw/658657aebc58480d5511bd726b8dfa73b4f49ac7/source/normal/mit_hrtf_normal_96000.h'
+      downloader 'https://github.com/greekgoddj/mit-hrtf-lib/raw/658657aebc58480d5511bd726b8dfa73b4f49ac7/source/normal/mit_hrtf_normal_44100.h', 'SKIP'
+      downloader 'https://github.com/greekgoddj/mit-hrtf-lib/raw/658657aebc58480d5511bd726b8dfa73b4f49ac7/source/normal/mit_hrtf_normal_48000.h', 'SKIP'
+      downloader 'https://github.com/greekgoddj/mit-hrtf-lib/raw/658657aebc58480d5511bd726b8dfa73b4f49ac7/source/normal/mit_hrtf_normal_88200.h', 'SKIP'
+      downloader 'https://github.com/greekgoddj/mit-hrtf-lib/raw/658657aebc58480d5511bd726b8dfa73b4f49ac7/source/normal/mit_hrtf_normal_96000.h', 'SKIP'
```
uberhacker, line 34: Why not replace the `@sha256sums` section below instead of `SKIP`?

uberhacker: Also the for loop could be rubified?

bin/crew
supechicken: Fixed

lib/downloader.rb
supechicken: Done

packages/libspatialaudio.rb
supechicken: > Why not replace the `@sha256sums` section below instead of `SKIP`?

~Because of lazy :)~

supechicken: Done

supechicken:
@satmandu I am seeing the prebuilt binaries in [packages/libspatialaudio.rb](https://github.com/skycocker/chromebrew/pull/6891/files#diff-4222267f2917bedb628174ce845072b933cc8e8c5ba5837d8b5ac920ee6430ec) started with `file://` prefix :)

bin/crew
```diff
@@ -712,10 +712,8 @@ def download
         end
       end
       # Download file if not cached.
-      downloader url, filename, @opt_verbose
+      downloader url, sha256sum, filename, 3, @opt_verbose
```
uberhacker, line 715: Should we make the retry count a variable?  It'd be nice if it was configurable.

supechicken: Now it can be adjust with `CREW_DOWNLOADER_RETRY` env variable

uberhacker: Great work @supechicken.  Is this ready?

supechicken: > Great work @supechicken. Is this ready?

Yes

uberhacker: Great, can we work on #6886 next? :)

Issue #6894, opened 2022-03-23
baobab needs postinstall modification


satmandu:
- `baobab` needs to have `system "glib-compile-schemas #{CREW_PREFIX}/share/glib-2.0/schemas"` in its postinstall.

Otherwise we see this: 
```
(baobab:12873): GLib-GIO-ERROR **: 11:37:53.214: Settings schema 'org.gnome.baobab.preferences' is not installed
```

PR #6904, opened 2022-03-26
Add rhythmbox package and totem_pl_parser dependency

uberhacker:
Rhythmbox is a music playing application for GNOME.  See https://wiki.gnome.org/Apps/Rhythmbox.

Tested on:
- [x] aarch64
- [x] armv7l
- [x] x86_64

### DIFF
--- a/packages/rhythmbox.rb
+++ b/packages/rhythmbox.rb
@@ -1,0 +1,42 @@
+require 'package'
+
+class Rhythmbox < Package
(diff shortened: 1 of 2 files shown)


satmandu:
Nice.
Does https://github.com/skycocker/chromebrew/issues/6894 apply since this is a gnome package?

uberhacker:
> Nice. Does #6894 apply since this is a gnome package?

Already included. :)

PR #6909, opened 2022-03-28
Gnome macro

### LINKED ISSUE #6894: baobab needs postinstall modification

satmandu:
Fixes #6894 
- Use macro for gnome package postinstalls.
- I expect this might require some revisions.. .suggestions welcome!

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=gnome_macro CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -941,6 +941,7 @@ def post_install
       puts "Performing post-install for #{@pkg.name}...".lightblue
       @pkg.postinstall
       if @pkg.gnome?
+        puts "Performing Gnome post-installs for #{@pkg.name}...".lightblue
         # Is this needed here?
         @device = JSON.parse(File.read("#{CREW_CONFIG_PATH}device.json"), symbolize_names: true)
         # generate schemas
--- a/bin/crew
+++ b/bin/crew
@@ -940,6 +940,22 @@ def post_install
     Dir.chdir post_install_tempdir do
       puts "Performing post-install for #{@pkg.name}...".lightblue
       @pkg.postinstall
+      if @pkg.gnome?
+        # Is this needed here?
+        @device = JSON.parse(File.read("#{CREW_CONFIG_PATH}device.json"), symbolize_names: true)
+        # generate schemas
+        if @device[:installed_packages].any? { |elem| elem[:name] == 'glib' }
+          system "#{CREW_PREFIX}/bin/glib-compile-schemas #{CREW_PREFIX}/share/glib-2.0/schemas"
+        end
+        # update mime database
+        if @device[:installed_packages].any? { |elem| elem[:name] == 'shared_mime_info' }
+          system "#{CREW_PREFIX}/bin/update-mime-database #{CREW_PREFIX}/share/mime"
+        end
+        # update icon cache, but only if gdk_pixbuf is already installed.
+        if @device[:installed_packages].any? { |elem| elem[:name] == 'gdk_pixbuf' }
+          system "#{CREW_PREFIX}/bin/gtk-update-icon-cache -ft #{CREW_PREFIX}/share/icons/* || true"
+        end
+      end
     end
   end
 end
@@ -1294,6 +1310,13 @@ def resolve_dependencies
   abort "Package #{@pkg.name} is not compatible with your device architecture (#{ARCH}) :/".lightred unless @device[:compatible_packages].any? {|elem| elem[:name] == @pkg.name }
 
   @dependencies = []
+  if @pkg.gnome?
+    GNOME_DEPS.each do |dep|
+      return if @pkg.name == dep
+
+      @dependendencies.push(dep) unless @device[:installed_packages].any? { |pkg| pkg[:name] == dep }
+    end
+  end
   expand_dependencies
 
   # leave only not installed packages in dependencies
--- a/bin/crew
+++ b/bin/crew
@@ -1313,7 +1313,7 @@ def resolve_dependencies
   @dependencies = []
   if @pkg.gnome?
     GNOME_DEPS.each do |dep|
-      return if @pkg.name == dep
+      continue if @pkg.name == dep
 
       @dependendencies.push(dep) unless @device[:installed_packages].any? { |pkg| pkg[:name] == dep }
     end
--- a/bin/crew
+++ b/bin/crew
@@ -941,9 +941,7 @@ def post_install
       puts "Performing post-install for #{@pkg.name}...".lightblue
       @pkg.postinstall
       if @pkg.gnome?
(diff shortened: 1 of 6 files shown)


bin/crew
```diff
@@ -1294,6 +1311,13 @@ def resolve_dependencies
   abort "Package #{@pkg.name} is not compatible with your device architecture (#{ARCH}) :/".lightred unless @device[:compatible_packages].any? {|elem| elem[:name] == @pkg.name }
 
   @dependencies = []
+  if @pkg.gnome?
+    GNOME_DEPS.each do |dep|
+      return if @pkg.name == dep
```
uberhacker, line 1316: Should this be `continue if`?

packages/baobab.rb
```diff
@@ -36,8 +37,4 @@ def self.build
   def self.install
     system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
   end
-
-  def self.postinstall
-    system "update-mime-database #{CREW_PREFIX}/share/mime"
-  end
```
uberhacker, line 42: This is the only package we do this?

satmandu: I didn't want to edit the others until I got consensus on the approach.

bin/crew
satmandu: This is meant to NOT add the package itself to its own dependencies, but I forget if logic for that already exists... @supechicken?

satmandu: Also, I'm not actually sure if this logic block adds these `GNOME_DEPS` from `const.rb` to the dependencies list for packages marked with `gnome`. Which is the goal here.

uberhacker: I guess I'm not understanding.  There are 3 `GNOME_DEPS` but this could abort after the second or third loop and miss any further dependencies.

satmandu: Fixed.

bin/crew
```diff
@@ -1294,6 +1311,13 @@ def resolve_dependencies
   abort "Package #{@pkg.name} is not compatible with your device architecture (#{ARCH}) :/".lightred unless @device[:compatible_packages].any? {|elem| elem[:name] == @pkg.name }
 
   @dependencies = []
+  if @pkg.gnome?
+    GNOME_DEPS.each do |dep|
```
supechicken, line 1315: If there are some gnome specific dependencies missing in some gnome packages, we should add those dependencies to the package recipes but not modify the dependency list.

bin/crew
supechicken: I think `next` is better than `continue` (in the meaning of the word itself) (although they do the same thing)

bin/crew
```diff
@@ -940,6 +940,23 @@ def post_install
     Dir.chdir post_install_tempdir do
       puts "Performing post-install for #{@pkg.name}...".lightblue
       @pkg.postinstall
+      if @pkg.gnome?
+        puts "Performing Gnome post-installs for #{@pkg.name}...".lightblue
+        # Is this needed here?
+        @device = JSON.parse(File.read("#{CREW_CONFIG_PATH}device.json"), symbolize_names: true)
```
supechicken, line 946:
```ruby
@device = JSON.load_file("#{CREW_CONFIG_PATH}device.json", symbolize_names: true)
```

bin/crew
```diff
@@ -940,6 +940,23 @@ def post_install
     Dir.chdir post_install_tempdir do
       puts "Performing post-install for #{@pkg.name}...".lightblue
       @pkg.postinstall
+      if @pkg.gnome?
+        puts "Performing Gnome post-installs for #{@pkg.name}...".lightblue
```
supechicken, line 944: Show it only at verbose?

bin/crew
satmandu: My point was that this is already being done in crew at line 143:
```
@device = JSON.parse(File.read(CREW_CONFIG_PATH + 'device.json'), symbolize_names: true)
```
Should we change that to `load_file`?

bin/crew
satmandu: The argument here is to just make sure the gnome deps needed for the gnome postinstall get added in case they aren't installed. These are logical dependencies, not necessarily build or runtime dependencies. (Though in this case some are runtime deps too.)

Is that a mistake?

packages/baobab.rb
satmandu: more can be added as we update...

bin/crew
uberhacker: I agree with @supechicken here.  We should include the dependencies in the packages instead.  There may be an edge case where we don't want the extra dependencies.  Better to control at the package level.

satmandu: Sounds good. I removed that entirely.

packages/meson.rb
```diff
@@ -3,24 +3,24 @@
 class Meson < Package
```
uberhacker, line 3: How did meson end up in this PR?  Didn't we just merge that recently?

satmandu: Ugh. My master merges must be screwed up.

satmandu: Fixed.

Zopolis4:
Correct me if I'm wrong, but don't gnome packages already do this?

`gnome_nibbles` (https://gitlab.gnome.org/GNOME/gnome-nibbles/-/blob/master/meson.build#L58-63)
```meson
# Postinstall
gnome.post_install(
  glib_compile_schemas: true,
  gtk_update_icon_cache: true,
  update_desktop_database: true,
)
```

`gtk4` (https://gitlab.gnome.org/GNOME/gtk/-/blob/main/meson.build?ref_type=heads#L861-869)
```meson
if not meson.is_cross_build()
  gnome.post_install(
    glib_compile_schemas: true,
    gio_querymodules: gio_module_dirs,
    gtk_update_icon_cache: get_option('build-demos'),
  )
else
  message('Not executing post-install steps automatically when cross compiling')
endif
```

See also https://github.com/mesonbuild/meson/issues/8268.

satmandu:
Looks like they do!

Do you want to add a PR to remove this?

Zopolis4:
I'll submit a PR in a little while.

PR #6938, opened 2022-04-02
Packages which need modification of installs to allow for partial installs.

satmandu:
- `llvm14` (which we need to modify the install of to allow only library install option.)
- Update `harfbuzz`, `fontconfig`, `freetype`, `cairo`, `librsvg`, `glib` and modify dependencies
- pixman also rebuilt to allow it to be used earlier in dependencies.
- Multiple packages here have overlapping files!
  - `harfbuzz` now contains , `ragel`, a `freetype` stub, and a `cairo` stub.
  - `fontconfig` now contains `libpng`, as fontconfig is a dep of `harfbuzz`.
- This PR is on HOLD until we have a way of upgrading all of these packages such that for instance freetype files from harfbuzz don't overwrite files from the full build of freetype. (see slack discussion) - because we can't ask people to run `crew upgrade ; crew reinstall pixman fontconfig harfbuzz librsvg freetype cairo`

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=fonts CREW_TESTING=1 crew update
crew upgrade
crew install pixman fontconfig harfbuzz librsvg freetype cairo
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -933,14 +933,15 @@ end
 def pre_install(dest_dir, provides)
   Dir.chdir dest_dir do
     puts 'Performing pre-install...'
-    @pkgProvidesDefault = if @pkg.provides&.nil?
+    puts @pkg.provides.inspect
+    @pkgProvidesDefault = if @pkg.provides&.nil? or @pkg.provides&.empty?
                             'default'
                           else
                             @pkg.provides.split(
                               / /, -1
                             )[0]
                           end
-    abort('invalid provides!') if !@pkg.provides.includes?(provides) && !@pkgProvidesDefault&.includes?(provides)
+    abort('invalid provides!') if !@pkg.provides.include?(provides) && !@pkgProvidesDefault&.include?(provides)
     provides = '' if (provides == 'default') && (@pkgProvidesDefault == 'default')
     provides = @pkgProvidesDefault if provides.nil?
     @pkg.preinstall(provides)
@@ -1440,7 +1441,7 @@ def install
 
   unless @pkg.is_fake?
     # perform pre-install process
-    pre_install dest_dir @pkgProvides
+    pre_install(dest_dir,@pkgProvides)
 
     # perform install process
     install_package dest_dir
@@ -1800,10 +1801,10 @@ end
 
 def install_command(args)
   args["<name>"].each do |name|
-    if name.includes?(':')
+    if name.include?(':')
       @PkgArgs = name.split(/:/, 2)
       @pkgName = @PkgArgs[0]
-      @pkgProvides = @pkgArgs[1]
+      @pkgProvides = @PkgArgs[1]
     else
       @pkgName = name
       @pkgProvides = 'default'
@@ -1841,7 +1842,14 @@ end
 
 def reinstall_command(args)
   args["<name>"].each do |name|
-    @pkgName = name
+    if name.include?(':')
+      @PkgArgs = name.split(/:/, 2)
+      @pkgName = @PkgArgs[0]
+      @pkgProvides = @PkgArgs[1]
+    else
+      @pkgName = name
+      @pkgProvides = 'default'
+    end
     search @pkgName
     print_current_package
     @pkg.build_from_source = true if @opt_src or @opt_recursive or CREW_BUILD_FROM_SOURCE == '1'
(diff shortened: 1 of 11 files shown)


bin/crew
```diff
@@ -1404,6 +1410,9 @@ def install
     # perform pre-install process
     pre_install dest_dir
 
+    # perform install_multiple process
+    install_multiple_package dest_dir
+
     # perform install process
     install_package dest_dir
```
uberhacker, line 1417: Is `install_multiple_package` an either/or for `install_package`?  If so, do we need an if...else...end?

satmandu: We don't have an if/then for preinstall...

uberhacker: What I'm asking is do we need to run both `install_multiple_package` and `install_package` or just one or the other?

satmandu: Right now I'm working on using both.

The install_multiple modifies filelist and dlist, and install_package then installs all files in (the resultant modified or unmodified) filelist and dlist.

That way the existing install algorithm stays largely intact.

uberhacker: Actually, do you even need a separate function?  Can't that go in preinstall instead?

satmandu: You're right.

bin/crew
```diff
@@ -1243,29 +1243,53 @@ def install_package(pkgdir)
...
+    
+    # Create filelists for rsync and tar to use. Use standard paths since 
+    # builds always use these paths.
+    FileUtils.touch %w[
+      dflist.rsync.home dflist.rsync.usrlocal dflist.tar.home 
+      dflist.tar.usrlocal
+    ]
+    # Create filelists for rsync to use.
+    system "grep -s \"^/home/chronos/user\" #{CREW_META_PATH + @pkg.name + '.directorylist'} | cut -c 19- > dflist.rsync.home"
+    system "grep -s \"^/home/chronos/user\" #{CREW_META_PATH + @pkg.name + '.filelist'} | cut -c 19- >> dflist.rsync.home"
```
uberhacker, line 1255: Abstract with `#{HOME}`?

bin/crew
```diff
@@ -1243,29 +1243,53 @@ def install_package(pkgdir)
...
+    
+    system "grep -s \"^/usr/local\" #{CREW_META_PATH + @pkg.name + '.directorylist'} | cut -c 11- > dflist.rsync.usrlocal"
+    system "grep -s \"^/usr/local\" #{CREW_META_PATH + @pkg.name + '.filelist'} | cut -c 11- >> dflist.rsync.usrlocal"
```
uberhacker, line 1258: Abstract with `#{CREW_PREFIX}`?

packages/fontconfig.rb
```diff
@@ -79,6 +79,14 @@ def self.install
     File.write("#{CREW_DEST_PREFIX}/etc/env.d/fontconfig", @env)
   end
 
+  def self.preinstall
+    @device = JSON.parse(File.read("#{CREW_CONFIG_PATH}device.json"), symbolize_names: true)
+    if @device[:installed_packages].any? { |elem| elem[:name] == 'freetype' }
+      system "sed -i '/*freetype*/d' filelist"
+      system "sed -i '/*freetype*/d' dlist"
+    end
+  end
```
uberhacker, line 88: Does it make sense to put this above `def self.install` to better represent the progression/flow of the packaging process?

bin/crew
satmandu: Why? The path in package binaries should never change, right?

bin/crew
satmandu: Same why as before. The crew prefix in built packages should always be `/usr/local`, otherwise installing binary packages will break, right?

We could specify the home and crew prefix used in a binary package in a file in the package though, and use that if it is there...

packages/fontconfig.rb
satmandu: Yes and no?

Yes, because it happens _before_ the actual install (for which `def self.install` is irrelevant.)
No, because it happens _after_ the packaging process which uses `def self.install`.

(This is another argument I suppose for renaming `def self.install` def self.package`)

satmandu:
Superseded by https://github.com/skycocker/chromebrew/pull/7002 and https://github.com/skycocker/chromebrew/pull/7006

PR #6943, opened 2022-04-03
mold -> 1.2.1, enable mold on armv7l

satmandu:
- Mold is now supportd on armv7l.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=mold_arm CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -123,18 +123,11 @@ when 'x86_64'
 end
 
 if ENV['CREW_LINKER'].to_s.empty?
-  case ARCH
-  when 'aarch64', 'armv7l'
-    CREW_LINKER = 'gold'
-    CREW_LINKER_FLAGS = ''
-  when 'i686', 'x86_64'
-    CREW_LINKER = 'mold'
-    CREW_LINKER_FLAGS = ''
-  end
+  CREW_LINKER = 'mold'
 else
   CREW_LINKER = ENV['CREW_LINKER']
-  CREW_LINKER_FLAGS = ENV['CREW_LINKER_FLAGS']
 end
+CREW_LINKER_FLAGS = ENV['CREW_LINKER_FLAGS']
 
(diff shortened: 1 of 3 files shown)


lib/const.rb
```diff
@@ -125,7 +125,7 @@
 if ENV['CREW_LINKER'].to_s.empty?
   case ARCH
   when 'aarch64', 'armv7l'
-    CREW_LINKER = 'gold'
+    CREW_LINKER = 'mold'
     CREW_LINKER_FLAGS = ''
   when 'i686', 'x86_64'
```
supechicken, line 130: Maybe just remove the `case` block?

packages/mold.rb
```diff
@@ -6,18 +6,22 @@
 class Mold < Package
   description 'A Modern Linker'
   homepage 'https://github.com/rui314/mold'
-  version '1.1.1'
-  compatibility 'i686, x86_64'
+  version '615a24d1dde2e7d0756a53897ed7f46842f0c3a1'
```
supechicken, line 9: Show the first 7 letters of the commit only?

satmandu:
@supechicken This is just for testing the armv7l functionality in mold. Once we have it confirmed working, we can ask for a tagged release...

lib/const.rb
satmandu: Once we know it works...

supechicken:
The new release of `mold` with armv7 support triggered two days ago :) (`v1.2.0`)

satmandu:
> The new release of `mold` with armv7 support triggered two days ago :) (`v1.2.0`)

Thanks for that! Building `1.2.0` now.

packages/mold.rb
satmandu: no longer relevant. :)

satmandu:
Looks like building patchelf on `armv7l` with this mold fails.

satmandu:
`patchelf` successfully built with mold 1.2.1, so let's get mold set as the default and go from there?

PR #6957, opened 2022-04-09
crew: Only perform postinstall only if it is defined by package, hide error when `ctrl+c`

supechicken:
### Changes
- Only perform `@pkg.postinstall` and show `post-install` message only if the `self.postinstall` function is defined (used) by the package recipe
- Hide backtrace when interrupted (`ctrl+c`)

```
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=postinstall_check CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -938,6 +938,9 @@ def pre_install(dest_dir)
 end
 
 def post_install
+  # return unless the postinstall function was defined by the package recipe
+  return if @pkg.method(:postinstall).source_location[0].include?(@pkg.name)
+
   Dir.mktmpdir do |post_install_tempdir|
     Dir.chdir post_install_tempdir do
       puts "Performing post-install for #{@pkg.name}...".lightblue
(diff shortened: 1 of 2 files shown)


bin/crew
```diff
@@ -1877,6 +1880,10 @@ def is_command(name)
   return !!!name[/^[-<]/]
 end
 
+trap('INT') do
+  abort 'Interrupted.'.lightred
+end
```
uberhacker, line 1885: Why is this needed?

supechicken: This is used to hide the additional backtrace message when `ctrl+c` is pressed (since this is not an error, backtrace messages are unnecessary):
```
# without trap
$ ruby -e 'sleep 10'
^C-e:1:in `sleep': Interrupt
        from -e:1:in `<main>'

# with trap
$ ruby -e 'trap("INT") { abort "This hide the backtrace" }; sleep 10'
^CThis hide the backtrace
```

uberhacker: I'm not sure this is a good idea.  There might be an interrupt internally and traceback would be helpful in those cases.

supechicken: As I know the interrupt signal is reserved as a user signal (should be fired by user only)

uberhacker: Alright then I'll just pass the folks who want to troubleshoot apps to you then. :)

PR #6994, opened 2022-04-18
Removed unnecesary dependencies in zathura package

5hif7y:
# Fixes

removed fish shell and bash_completion from zathura dependencies.

Tested in:
- [x] `x86_64`


uberhacker:
I have to agree with this suggestion.  Thank you for the update @5hif7y!

PR #7002, opened 2022-04-20
Fonts package updates (w/o reworking crew) + qemu -> 7.0 (also update qemu deps)

### LINKED ISSUE #6959: [qemu] Incorrect libpng library version
```
$ crew search ^qemu -v
qemu: QEMU is a generic and open source machine emulator and virtualizer.
http://www.qemu.org/
Version: 6.2.0
$ qemu-system-x86_64
qemu-system-x86_64: /usr/local/lib64/libpng16.so.16: version `PNG16_0' not found (required by qemu-system-x86_64)
```

satmandu:
Fixes #6959 

- Also `qemu`->7.0

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=fonts3 CREW_TESTING=1 crew update
crew upgrade pixman fontconfig harfbuzz librsvg freetype cairo glib libxdmcp
```

### DIFF
--- a/packages/freetype.rb
+++ b/packages/freetype.rb
@@ -3,23 +3,23 @@ require 'package'
 class Freetype < Package
   description 'FreeType is a freely available software library to render fonts.'
   homepage 'https://www.freetype.org/'
-  version '2.11.1' # Update freetype in harfbuzz when updating freetype
+  version '2.12.0' # Update freetype in harfbuzz when updating freetype
   license 'FTL or GPL-2+'
   compatibility 'all'
   source_url 'https://gitlab.freedesktop.org/freetype/freetype.git'
   git_hashtag "VER-#{version.tr('.', '-')}"
 
   binary_url({
-    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/freetype/2.11.1_armv7l/freetype-2.11.1-chromeos-armv7l.tar.zst',
-     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/freetype/2.11.1_armv7l/freetype-2.11.1-chromeos-armv7l.tar.zst',
-       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/freetype/2.11.1_i686/freetype-2.11.1-chromeos-i686.tar.zst',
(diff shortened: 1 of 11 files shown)


packages/freetype.rb
```diff
@@ -62,5 +56,29 @@ def self.install
...
+      end
+      conflicts.each do |conflict|
+        conflict.each do |thisconflict|
+          singleconflict = thisconflict.split(':',-1)
+          if @override_allowed.include?(singleconflict[0])
+            system "sed -i '\\\?^#{singleconflict[1]}?d'  #{CREW_META_PATH}/#{singleconflict[0]}.filelist"
+          end
+        end
+      end
+    end
```
uberhacker, line 82: Should this logic replace the conflict handling code we have in crew?  Seems like it could be used universally?

packages/harfbuzz.rb
```diff
@@ -64,4 +67,39 @@ def self.build
...
+      conflicts.each do |conflict|
+        conflict.each do |thisconflict|
+          singleconflict = thisconflict.split(':',-1)
+          if @override_allowed.include?(singleconflict[0])
+            system "sed -i '\\\?^#{singleconflict[1]}?d'  #{CREW_META_PATH}/#{singleconflict[0]}.filelist"
+          end
+        end
+      end
+    end
+  end
```
uberhacker, line 104: Should this logic replace the conflict handling code we have in crew? Seems like it could be used universally?

packages/freetype.rb
satmandu: I think this function should ultimately be integrated into crew, yes.

But I was waiting on @supechicken 's work on making packages smaller in https://github.com/supechicken/chromebrew/tree/llvm_lib_diff to get done, but I haven't had time to go through that yet.

uberhacker: Does this logic override what crew executes or does it run afterwards?  I'm guessing there are 2 conflicts checks running?

uberhacker: But are there 2 checks or just the 1?  Ultimately, I think we should have just one check.

satmandu: There is an existing check at the moment, and this adds a second check in the post install.

As this second check isn't universal, the first (existing) check is still useful to tell us how the second check should be written. (I.e., we need to know which conflicting files, from which package, should be considered to be superseding.)

PR #7018, opened 2022-04-28
crew: Add multi-component support in one package

supechicken:
### Concept
- A package can have multiple "components", each "component" can have its own `description`
- Each component needs a file filter to tell `crew` which file(s) need to be extracted/installed from the main package
- **The component `all` represent the main package**
- Use `<pkgName>::<componentName>` in cmdline/`depends_on` if you want to specify a component
- If component `all` (the whole package) is installed, then all components in that package will be marked as `installed`
- If component `all` (the whole package) is installed, then `crew remove <pkgName>::<componentsOtherThanALL>` will not work
- `crew` uses the new `installedComponents` entry (see `Major changes`) to determine whatever a component is installed (see line 590 for more info)

### Major changes
- New "package": `llvm::libs`, `gcc11::libs`
- Modify commands to accept the new `<pkgName>::<componentName>` format
- A new `device.json` entry called `installedComponents` will be added to newly installed packages, it contains all installed components of the current package (note: If the main package is installed, only component `all` will be added to the list (not every component name in the package))

### Other changes
- Added `package_installed?` function (used to check if a package/component installed)
- The `set_package` function no longer require the full package path

## Additional information

Works properly:
- [x] `x86_64`

```
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=multi_component CREW_TESTING=1 crew update
```

# **Further tests needed**


saltedcoffii:
Because this platform is semiprofessional or at least we like to think we are, I'm going to avoid writing in all caps, however, I have been so excited for so long for this to happen. Here's what I'd like to see each package be broken down into:

First, packages that have multiple different packages within them, like llvm, should probably be broken down into their component parts, e.g., `llvm::llvm`, `llvm::clang`, etc, and `man` can have `man::man`, `man::apropos`, etc.

Second, I'd like to see the following base number of components in every package:
- `lib` for shared object libraries, e.g., `gcc::lib`
- `bin` for binaries, e.g., `gcc::bin`
- `doc` for documentation, like manpages, info pages, gtk-doc, etc., e.g., `gcc::doc`
- `dev` for headers and the like, e.g., `gcc::dev`
- `static` for static libraries `gcc::static`
This can even apply to packages with `lib` in their name, we can have `libostree::lib` and `libostree::bin`.
Can anyone else think of anymore categories?

Thirdly, can we make it possible to specify components of components, so we can do, e.g., `llvm::clang::lib` and `man::apropos::doc` ?

The idea in my head for the first three is to break down packages as small as possible (much like Alpine Linux) so only what is needed is installed.

Fourthly, I would like each component to be able to have its own dependencies, and packages to depend on specific components. For example, `gcc::lib` would be necessary for almost every package, however `gcc::bin` is only necessary if you want to compile C code, and `buildessential` can depend on `gcc::bin` and `core` can depend on `gcc:lib`.

Lastly, I'd like to see all of `core` and `buildessential` made into this new form before this PR is merged. I'm happy to help with any and all building. @.satmandu check slack.

uberhacker:
`etc` for config like `gcc::etc`?
If we're going to split out static libraries should there also be a `gcc::shared` subpackage too?  Needless to say, this could end up being a complete nightmare to maintain.

packages/buildessential.rb
```diff
@@ -14,13 +14,11 @@ class Buildessential < Package
 
   # install first to get ldconfig
   depends_on 'glibc'
-  # depends_on 'gcc11'
   depends_on 'gmp'
   depends_on 'mpfr'
   depends_on 'mpc'
   depends_on 'isl'
   depends_on 'libyaml'
-  depends_on 'gcc_tools'
```
satmandu, line 23: Nice.

satmandu:
Looking at this code, it looks like package:arbitrary_sub_package_name all depends upon code within the package file itself?

That sounds fine to me. You can add whatever you want, or not, to packages....

Let's see how this works!

satmandu (approved):

satmandu:
This looks good to me.

After this is merged, we should update mesa to also have `depends_on 'llvm::libs'`.

(or merge master and modify before merge :) )

satmandu:
@saltedcoffii Do you still think that all of core needs to be modified before this is merged? Just having llvm and GCC use this format gives users immediate savings of disk space...

uberhacker:
Run `sed -i 's,slient,silent,g' bin/crew`.

uberhacker:
Use `Dir[...]` instead of `Dir.glob(...)` in `bin/crew`?

bin/crew
```diff
@@ -149,40 +149,51 @@ end
...
   end
-  puts ""
 end
 
-def set_package(pkgName, pkgPath)
-  begin
-    require_relative pkgPath
-  rescue SyntaxError => e
-    puts "#{e.class}: #{e.message}".red
+def set_package(pkgName, componentName = 'all', slient: false)
```
uberhacker, line 183: Change `slient` to `silent`?

bin/crew
```diff
@@ -149,40 +149,51 @@ end
...
+  # set_package: load package #{pkgName} and bind the package with variable @pkg
+  #
+  #   pkgName: name of the package
+  #   componentName: component to be operated, only used to set @pkg.target_component here (default: 'all')
+  #
+  pkgPath = "#{CREW_PACKAGES_PATH}/#{pkgName}.rb"
+
+  unless File.exist?(pkgPath)
+    unless slient
```
uberhacker, line 192: Change `slient` to `silent`?

bin/crew
```diff
@@ -302,40 +330,71 @@ def generate_compatible
...
-      packageName = File.basename packagePath, '.rb'
-      begin
-        set_package packageName, packagePath
-      rescue => e
-        puts "Error with #{pkgName}.rb: #{e}".red unless e.to_s.include?('uninitialized constant')
+def regexp_search(pattern)
+  regex = Regexp.new(pattern, Regexp::IGNORECASE)
+  any_matches = nil
+
+  Dir.glob("#{CREW_PACKAGES_PATH}/*.rb") do |pkgPath|
```
uberhacker, line 368: Use `Dir[...]` instead of `Dir.glob(...)`?

bin/crew
```diff
@@ -302,40 +330,71 @@ def generate_compatible
...
       end
-      if ( @pkg.description =~ /#{pkgPat}/i )
-        print_current_package @opt_verbose
-        results.push(packageName)
+    end
+  end
+
+  return if any_matches
+
+  Dir.glob("#{CREW_PACKAGES_PATH}/*.rb") do |pkgPath|
```
uberhacker, line 385: Use `Dir[...]` instead of `Dir.glob(...)`?

bin/crew
```diff
@@ -1249,33 +1380,10 @@ def install_package(pkgdir)
...
-      end
-    end
-    if Dir.exist? "#{pkgdir}/#{CREW_PREFIX}"
-      if rsync_available
-        # Adjust "./usr/local" if the build CREW_PREFIX ever changes.
-        system "rsync -ahHAXWx --remove-source-files ./#{CREW_PREFIX}/ #{CREW_PREFIX}"
-      else
-        system "cp -a#{@verbose} ./#{CREW_PREFIX}/* #{CREW_PREFIX}"
-      end
-    end
```
uberhacker, line 1278: I take it you don't like the rsync? :)

satmandu: It got moved to the `copy_to_rootfs` function.

satmandu:
Made @uberhacker 's suggested changes.  Please confirm nothing is broken...

uberhacker:
I'm going to hold off on this until we have thoroughly tested.  Nice update @supechicken!

packages/llvm.rb
```diff
@@ -27,10 +27,13 @@ class Llvm < Package
   depends_on 'py3_pygments' => :build
   depends_on 'ccache' => :build
   depends_on 'elfutils' # R
-  depends_on 'gcc' # R
+  depends_on 'gcc11::libs' # R
```
uberhacker, line 30: If we're going to include version specific packages, we need to make sure if gcc is updated to version 12, these dependency lines are updated too.

satmandu: We need to figure out how to make this work with `gcc::libs` so we don't have to specify the version number.

supechicken:
> Use `Dir[...]` instead of `Dir.glob(...)` in `bin/crew`?

**`Dir[...]` will not work with blocks...**

satmandu:
> > Use `Dir[...]` instead of `Dir.glob(...)` in `bin/crew`?
> 
> **`Dir[...]` will not work with blocks...**

Sorry about that...

saltedcoffii:
> @saltedcoffii Do you still think that all of core needs to be modified before this is merged? Just having llvm and GCC use this format gives users immediate savings of disk space...

obviously its not necessary but I'd like to get all packages multi-componented, and starting with core and buildessential is a great way to do that

bin/crew
uberhacker: This should work.  There is no difference.

uberhacker: Try `Dir["..."].each`.  We should be consistent with the way we use `Dir` throughout.

bin/crew
uberhacker: Use `Dir[...].each`?

bin/crew
uberhacker: I think this is too much abstraction.  If it was called more than once...

bin/crew
supechicken: If there have a cleaner way to write it why not choose it :)

uberhacker: All I'm saying is let's be consistent.  There are many other places in this file that use `Dir[...].each`.  Both methods are fine but if they do the same thing we should try to be consistent throughout to avoid confusion.

satmandu:
It's been 9 days, has anyone had time to multi-component further packages in this PR? If not I think we should move to doing that in separate PRs... I'd really like to start testing this with installs.

packages/llvm.rb
uberhacker: This is no longer an issue if we rename gcc12.rb to gcc.rb and remove the version?
```
depends_on 'gcc::libs'
```

satmandu:
FYI it would also make sense to split `libgccgit` out of the GCC package since that is rarely going to be needed.

packages/llvm.rb
uberhacker: Are we still trying to abstract the gcc version?

satmandu:
We should just make the gcc.rb package like lvm2 and have versioning inside
reflect the compiler version.

I haven't had a chance to do that yet.

packages/core.rb
```diff
@@ -19,7 +19,7 @@ class Core < Package
   depends_on 'expat'
   depends_on 'filecmd'
   depends_on 'flex'
-  depends_on 'gcc'
+  depends_on 'gcc11::libs'
```
uberhacker, line 22: `gcc::libs`?

packages/icu4c.rb
```diff
@@ -22,6 +22,8 @@ class Icu4c < Package
     x86_64: '868a783d335eb13ff06d29b5221448a50e68a728ee1c35f3dc4c16c816716d68'
   })
 
+  depends_on 'gcc11::libs'
```
uberhacker, line 25: `depends_on 'gcc::libs'`?

packages/core.rb
supechicken: Change made

supechicken:
### Changes in [55d8591](https://github.com/skycocker/chromebrew/pull/7018/commits/55d8591a7e0256927488cc092c2e448993fd4995)...[5ad0247](https://github.com/skycocker/chromebrew/pull/7018/commits/5ad0247ea4290d4d8acb6ed60cc3fc323834d451)
- The `provides` function now accepts multiple filter types, including glob, regex and array containing filenames
- Switch from `gcc11::libs` to `gcc::libs`
- Merge upstream

uberhacker:
I hope I don't regret merging this. :)

supechicken:
### Changes in [da0fdd0](https://github.com/skycocker/chromebrew/pull/7018/commits/da0fdd0d5795bb5ec68dd85d947f6c34a9a6985b)
- Now the `CREW_PREFIX` in filenames will be deleted before passing to the filter function, for example:
```ruby
# before
=> "/usr/local/lib64/foo/a"
"#{CREW_LIB_PREFIX}/foo/*" # will match

# after
=> "lib64/foo/a"
"#{CREW_LIB_PREFIX}/foo/*" # will not match
"#{CREW_LIB_SUFFIX}/foo/*" # will match
```

Issue #6133, opened 2021-09-01
Move Primary repo to https://github.com/chromebrew/chromebrew


satmandu:
**To-do list.** _(Open for discussion)_
- [ ] Set date for changeover. _September 30, 2021?_
- [x] Push changes to new repo during transition period. (Started by @satmandu )
  - [ ] Use main as default branch on new repo?
- [x] Change URLs in `install.sh`
- [x] Update `crew` to modify git repo sync settings to point to new repo
- [x] Sync issues and PRs
- [x] Move Wiki (remove deprecated FAQ sections and clean up)
- [x] Update Readme
- [x] Modify text in packages which refers to the old repo.
- [x] Redirect https://github.com/skycocker/chromebrew to https://github.com/chromebrew/chromebrew (possible?)
- [ ] ~Archive old repo~
- [x] Move [Github pages](https://skycocker.github.io/chromebrew/)
- [ ] **Deprecation for i686, `x86_64 w/ glibc 2.27` (EOL before M92), `armv7l w/ glibc 2.27` (EOL before M92)**
  - This will affect:
  - (Start compiling all software on glibc 2.32 based systems/containers. _Glibc 2.32 was added in M92._)

Suggested deprecated devices (not receiving M92):
big
blaze
buddy
butterfly
clapper
daisy
falco-li
gandof
glimmer
guado
jacuzzi
jaq
jerry
kitty
leon
link
lulu
lumpy
mario
mccloud
mickey
mighty
minnie
monroe
paine
panther
parrot
peppy
pi
pit
puff
quawks
rikku
samus
skate
speedy
spring
squawks
stout
stumpy
terra
tidus
tricky
volteer
wolf
x86-alex-he
x86-zgb-he
yuna
zako
zork

</body>

</html>

supechicken:
The homepage has just moved to https://chromebrew.github.io

Zopolis4:
> The homepage has just moved to https://chromebrew.github.io

To that end, we should probably update the github website.

saltedcoffii:
Alright let's get this done! It's super easy.
@uberhacker, can you:
- delete the https://github.com/chromebrew/chromebrew repository

uberhacker:
It's actually easier than that.  All we need to do is add a remote to git@github.com:skycocker/chromebrew.git and `git pull` everything over.  We just need @skycocker to add a redirect to the new repo immediately afterwards and that should do it.

saltedcoffii:
[...]

PR #7044, opened 2022-05-09
Deprecation of i686 architecture

supechicken:
Part of #6133 

Maybe it's time to do it :)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -70,6 +70,21 @@ if [ "${EUID}" == "0" ]; then
   exit 1;
 fi
 
+case "${ARCH}" in
+"i686")
+  echo_info 'The support for legacy 32-bit userspace (i686) is dropped on 2022-05-10,'
+  echo_info 'Consider switching to a 64-bit Chromium OS.'
+  exit 1
+  ;;
+"x86_64"|"armv7l"|"aarch64")
+  LIB_SUFFIX=
+  [ "${ARCH}" == "x86_64" ] && LIB_SUFFIX='64'
+  ;;
+*)
+  echo_error "Your device is not supported by Chromebrew yet :/"
+  exit 1;;
+esac
+
 echo_success "Welcome to Chromebrew!"
 
 # prompt user to enter the sudo password if it set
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -70,6 +70,21 @@ if [ "${EUID}" == "0" ]; then
   exit 1;
 fi
 
+case "${ARCH}" in
+"i686")
+  echo_info 'The support for legacy 32-bit userspace (i686) is dropped on 2022-05-10,'
+  echo_info 'Consider switching to a 64-bit Chromium OS.'
+  exit 1
```
uberhacker, line 77: Would a warning be better?  IOW, no `exit 1`.  Also, since arm Chromebooks still operate in a 32-bit userspace, maybe it would be better to say:
```
Support for i686 architecture has been dropped as of May 10, 2022.
```
?

satmandu:
I'd be in favor of a soft deprecation. Just stop bothering with most i686 packages. Stop adding binaries for i686 as we add updates...There's no reason to delete binaries yet.

And crew should definitely start showing a message that it is deprecated when someone does a `crew update` from `i686` machines...

uberhacker:
@saltedcoffii: I think you are overlooking the fact that i686 Chromebooks will become paper weights without Chromebrew.  There are many cli programs that still work fine so I think this suggestion is a bit extreme.  In fact, older i686 systems make great digital photo galleries.  Without Chromebrew, this wouldn't be possible.

satmandu:
@uberhacker The counterpoint is of course that by providing software, there is an implicit assumption that people are relying on us to keep that software up to date and safe, and as we are already not updating large chunks of the OS (glibc, etc) that becomes more and more an increasingly false assumption.  Nobody is going to get a i686 Chromebook to do CLI stuff when they can get a cheap x86_64 Chromebook to do the same where more is supported. The price differential is de minimis at this point...

uberhacker:
That sounds like a good idea.  What do you think @supechicken?

supechicken:
Soft deprecation is definitely the way to go :)

However, if we are going to go ahead with soft deprecation, here are something we might need to do:
- Version inconsistent between package script and prebuilt binary: Although we can solve it by adding some new functions to `crew`, a better way is to create a separate branch from `master` for i686
- Keep updating crew itself: If we decided to create a separate branch from `master` for i686, the second problem we meet is how to update the `crew` script itself only. Instead of using something like `git sparse-checkout`, I'd prefer to separate the `crew` scripts and packages into two standalone repo (I remember that @saltedcoffii has been working with it times ago) and update `crew` from its own repo only when `crew update`
- A new place for i686 prebuilt binaries: (recommend) Keeping those outdated i686 packages in an active repo might make new contributors confusing. Maybe move all i686 package binaries to a separate repo?

uberhacker:
That seems like a lot of extra work just to serve less than 1% of our user base.  What is wrong with just removing the i686 binaries altogether from the packages?  We have compatibility just for this purpose.  Like I've mentioned before, there are still a lot of cli apps that work fine.  It's mostly the gui apps and their dependencies that are broken in i686.

satmandu:
> What is wrong with just removing the i686 binaries altogether from the packages?

All packages or some packages?

uberhacker:
It would only be the gui-related packages since most cli programs work fine?  So...some packages which should roughly be only around 20%.

PR #7048, opened 2022-05-10
openssl rebuilds, add gcc12 to install.sh

satmandu:
- Install is currently broken due to rsync and ruby complaining about the just rebuilt openssl, so `rsync`, `ruby`, and `popt` (dep of rsync) are rebuilt in this PR. (rsync and popt are actually updated.)
- Also, `gcc12` is added to `install.sh` since gcc12's libgcc may be needed for ruby at install time.
- (This can hopefully be adjusted to just have gcc12::libs once we port that function to work in `install.sh` @supechicken 

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=openssl_rebuilds  CREW_TESTING=1 crew update
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -24,7 +24,7 @@ CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 ARCH="${ARCH/armv8l/armv7l}"
 
 # BOOTSTRAP_PACKAGES cannot depend on crew_profile_base for their core operations (completion scripts are fine)
-BOOTSTRAP_PACKAGES="musl_zstd pixz ca_certificates git gmp ncurses xxhash lz4 popt libyaml openssl zstd rsync ruby"
+BOOTSTRAP_PACKAGES="musl_zstd pixz ca_certificates git gmp ncurses xxhash lz4 popt libyaml openssl zstd gcc12 rsync ruby"
 
 # Add musl bin to path
 PATH=/usr/local/share/musl/bin:$PATH
--- a/packages/popt.rb
+++ b/packages/popt.rb
@@ -23,7 +23,7 @@ class Popt < Package
   })
 
   depends_on 'glibc' # R
-  depends_on 'gcc12' if ARCH == 'i686' # R
+  depends_on 'gcc12' unless ARCH == 'x86_64' # R
 
   def self.build
     system '[ -x configure ] || NOCONFIGURE=1 ./autogen.sh'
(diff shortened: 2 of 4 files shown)


install.sh
```diff
@@ -24,7 +24,7 @@ CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 ARCH="${ARCH/armv8l/armv7l}"
 
 # BOOTSTRAP_PACKAGES cannot depend on crew_profile_base for their core operations (completion scripts are fine)
-BOOTSTRAP_PACKAGES="musl_zstd pixz ca_certificates git gmp ncurses xxhash lz4 popt libyaml openssl zstd rsync ruby"
+BOOTSTRAP_PACKAGES="musl_zstd pixz ca_certificates git gmp ncurses xxhash lz4 popt libyaml openssl zstd gcc12 rsync ruby"
```
uberhacker, line 27: Would it be better to rename gcc12.rb to gcc.rb so we don't have to update this file every time for a new major gcc version?

packages/popt.rb
```diff
@@ -3,25 +3,28 @@
...
-       i686: '611be7ffbadab77c9da3492f13fbc8860522ec01adefbfbb462fb9c1afc76530',
-     x86_64: '591ff491561a48dc2a993b718a554dc303fab823929dc622963d6144db6e5fe1'
+    aarch64: '5bc51a9e367fac643a4c39acf250d8989c606c3d49c5ee41a662911acfc6863d',
+     armv7l: '5bc51a9e367fac643a4c39acf250d8989c606c3d49c5ee41a662911acfc6863d',
+       i686: '3ef4682d25a75e2d3464a5479744d9bd57473e9e303d1161fa4a2945752d3eb5',
+     x86_64: '93a2f1c02bda8ceee0c4cec070eab58bb997ba1a80e5cd0c37d4052464da4ce2'
   })
 
+  depends_on 'glibc' # R
+  depends_on 'gcc12' if ARCH == 'i686' # R
```
uberhacker, line 26: Why is this only needed for i686?

install.sh
satmandu: Yes. In another PR though?

packages/popt.rb
satmandu: Turns out that armv7l needs it too...

uberhacker:
I don't understand the `if` and `unless` conditions in the dependencies but whatever...

PR #7076, opened 2022-05-22
Merge `gcc12.rb` and `gcc.rb`, cleanup, remove any `gcc` packages older than `12`

supechicken:
Merge `gcc12.rb` and `gcc.rb` for #7018, also removes all outdated `gcc` versions (it is not necessary to keep multiple outdated `gcc` versions.)

### Changes
- Remove `gcc{7,8,10,11}`
- Remove `gcc-tools`
- Merge `gcc12.rb` and `gcc.rb`
- Cleanup **( review needed :) )**
- Fix `exception` logic in `def self.system` (`package.rb`)

Tested on `x86_64`


uberhacker:
This looks great.  Thank you for the update @supechicken!

PR #7082, opened 2022-05-24
Derive binary_url in package.rb

Zopolis4:
## Description
When finished, this PR will have package.rb find the binary_url of packages using the variables already present, instead of having a bunch of unnecessary lines in each package. 

Current problems include the lack of consistency between package compression (`tar.gz`, `tar.xz`, `tpxz`) and how some packages still don't use gitlab. Also-- is there any reason why we denote aarch64 and armv7l as different platforms internally if they have identical binaries? The current setup would need multiple ugly `if` statements to catch the aarch64 platform and replace it with armv7l.

Once that's resolved, I plan to create a sed script which goes through every package, and (if they use gitlab) replaces the binary_url values with a simple enum detailing what platforms it supports. (i.e. `binaries ({aarch64, armv7l, i686, x86_64})`). Alternatively, this could be merged with binary_sha256, and have one enum containing the hashes and the platforms supported. 

Note: I also haven't tested any of these changes as I don't have easy access to a ChromeOS machine atm.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=binary_url crew update
```


satmandu:
Is the lack of consistency in package compression a problem?

We have tried multiple compression formats, and we need to support multiple compression formats for installs to work. (e.g. we need to boostrap zstd support with zstd from a tar.xz or tar.gz package for architectures where recent zstd support is missing.)

Zopolis4:
Fair enough on both points, I'll adjust the changes and modify an example package in a bit.

lib/package.rb
```diff
@@ -184,8 +184,8 @@ def self.depends_on (dependency = nil)
   end
 
   def self.get_url (architecture)
-    if !@build_from_source and @binary_url and @binary_url.has_key?(architecture)
-      return @binary_url[architecture]
+    if !@build_from_source and @binary_sha256 and @binary_sha256.has_key?(architecture)
```
uberhacker, line 187: `@binary_sha256` twice?

Zopolis4:
I'm not too experienced with `sed`, so if someone who knows it well could write something to automate the conversion that'd be great.

Zopolis4:
I'm also not sure what to do about [this line](https://github.com/skycocker/chromebrew/blob/master/install.sh#L172) in `install.sh`.

supechicken:
> I'm also not sure what to do about [this line](https://github.com/skycocker/chromebrew/blob/master/install.sh#L172) in `install.sh`.

https://github.com/skycocker/chromebrew/blob/729178cae73bdd4e98fa95ed58a4622bd813e2c3/install.sh#L172

It is used to extract the `binary_url` of the current architecture. For example:
```
# a random package recipe with prebuilt binaries, `ruby` in this example
$ cat /usr/local/lib/crew/packages/ruby.rb
...snip...
  binary_url({
    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby/3.1.2-2_armv7l/ruby-3.1.2-2-chromeos-armv7l.tar.zst',
     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby/3.1.2-2_armv7l/ruby-3.1.2-2-chromeos-armv7l.tar.zst',
       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby/3.1.2-2_i686/ruby-3.1.2-2-chromeos-i686.tar.zst',
     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby/3.1.2-2_x86_64/ruby-3.1.2-2-chromeos-x86_64.tar.zst'
  })
  binary_sha256({
    aarch64: '6a7f6815e3c945afa520415d82caed3a277af294d323713ff3d3afb2e9201374',
     armv7l: '6a7f6815e3c945afa520415d82caed3a277af294d323713ff3d3afb2e9201374',
       i686: '8c1c82df986e4038b1c6e4a17d1807542e336c0517269f5398351dc896ab2343',
     x86_64: '70bae7cde10fadde59f6b2ab5c2017fdfa390385820a4154df49533527b6e3ff'
  })
...snip...


# this is our system architecture
$ uname -m
x86_64


# we are going to extract the corresponding `binary_url` from `ruby.rb`, using the expression in `install.sh`, line 172
[[ "$(sed -n '/binary_url/,/}/p' ruby.rb)" =~ .*${ARCH}:[[:blank:]]*[\'\"]([^\'\"]*) ]] 
#
# "$(sed -n ...)": extract the `binary_url` part from `ruby.rb`, the output of the command will be like this:
#    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby/3.1.2-2_armv7l/ruby-3.1.2-2-chromeos-armv7l.tar.zst',
#     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby/3.1.2-2_armv7l/ruby-3.1.2-2-chromeos-armv7l.tar.zst',
#       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby/3.1.2-2_i686/ruby-3.1.2-2-chromeos-i686.tar.zst',
#     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby/3.1.2-2_x86_64/ruby-3.1.2-2-chromeos-x86_64.tar.zst'
#
# `=~ .*${ARCH}:[[:blank:]]*[\'\"]([^\'\"]*)`: extract corresponding `binary_url` for our system architecture using bash built-in regex support,
# matched result (the actual download link) will be stored in ${BASH_REMATCH[1]}


# now ${BASH_REMATCH[1]} contains the download url
$ echo "${BASH_REMATCH[1]}"
https://gitlab.com/api/v4/projects/26210301/packages/generic/ruby/3.1.2-2_x86_64/ruby-3.1.2-2-chromeos-x86_64.tar.zst
```

Zopolis4:
I know what it does, not so much how to replace it, given that it finds the binary url through sed and doesn't derive it through ruby.

supechicken:
> I know what it does, not so much how to replace it, given that it finds the binary url through sed and doesn't derive it through ruby.

In order to actually "parse" the values of `binary_*` (instead of "extracting" it), we need to let `bash` understand the format of those urls, a better way is to convert all these hashes to JSON-compatible formats (so that we can parse it from `jq` in `bash`)

Like this (but it still requires sed for extracting the `binary_*` part):
```
  binary_sha256({
    "aarch64": "6a7f6815e3c945afa520415d82caed3a277af294d323713ff3d3afb2e9201374",
     "armv7l": "6a7f6815e3c945afa520415d82caed3a277af294d323713ff3d3afb2e9201374",
       "i686": "8c1c82df986e4038b1c6e4a17d1807542e336c0517269f5398351dc896ab2343",
     "x86_64": "70bae7cde10fadde59f6b2ab5c2017fdfa390385820a4154df49533527b6e3ff"
  })
```

Zopolis4:
The hashes aren't the issue, its the lack of a precomposed binary_url that is the problem. Afaik, you aren't able to neatly parse ruby files with sed/bash in the same way that I do in package.rb. Maybe https://github.com/whitequark/parser would work?

saltedcoffii:
> The hashes aren't the issue, its the lack of a precomposed binary_url that is the problem. Afaik, you aren't able to neatly parse ruby files with sed/bash in the same way that I do in package.rb. Maybe https://github.com/whitequark/parser would work?

Perhaps I'm missing a piece of something, but why can't we just use ruby itself to extract the variables?
Also, using `sed` and `bash` shouldn't be a problem at all. Other useful parsing tools include `tr`, `awk` (we use `mawk` with Chromebrew, so don't use `gawk` specific code), `expr`, `grep`, `xargs`, and likely more. Perhaps using `grep` to find the line and `sed` to remove the unwanted parts?

Zopolis4:
We could probably use ruby, but the only issue is I don't think we have ruby at this stage in the install.

Zopolis4:
Ok, I've done some research and there are a few options to deal with the `binary_url` stuff in `install.sh`.

1. Make a blood sacrifice to the `sed` gods and write an incoherent, hacky mess of `sed` commands that should hopefully occasionally work.
2. Hardcode a couple of urls in `install.sh`
3. Rewrite much of `install.sh` to be a smaller script designed to setup the bare minimum of crew and then use crew to install the rest of itself, eliminating the need to get (most of the) `binary_url` values through sed.

Zopolis4:
I'm leaning towards option 3, which would probably be separated into a different PR.

uberhacker:
Option 3 is already part of install.sh.

Zopolis4:
Not enough for my purposes (using ruby to extract the binary_url values).

Zopolis4:
I've added the `binary_compression` value to each package using this _beautiful_ script, working on deleting the binary_url values next.

```bash
#!/bin/bash

for package in *.rb; do
  [[ "$(sed -n '/binary_url/,/}/p' "${package}")" =~ .*${ARCH}:[[:blank:]]*[\'\"]([^\'\"]*) ]]
    bn1="$(basename "${BASH_REMATCH[1]}")"

  if [[ $bn1 == *.tpxz ]]; then
    bnc=".tpxz"
  elif [[ $bn1 == *.tar.xz ]]; then
    bnc=".tar.xz"
  elif [[ $bn1 == *.tar.gz ]]; then
    bnc=".tar.gz"
  elif [[ $bn1 == *.tar.zstd ]]; then
    bnc=".tar.zstd"
  elif [[ $bn1 == *.tar.zst ]]; then
    bnc=".tar.zst"
  else
    bnc=""
  fi

  if [[ $bnc != "" ]]; then
    export inject="  binary_compression '${bnc}'"
    if grep -q "Adapted from Arch Linux" ${package}; then
      # Thanks tshiono! (https://stackoverflow.com/questions/73439455/how-to-edit-the-second-blank-line-in-a-file-using sed)
      gawk -i inplace '/^[[:blank:]]*$/ && ++c == 3 {print ENVIRON["inject"]} 1' ${package}
    else
      gawk -i inplace '/^[[:blank:]]*$/ && ++c == 2 {print ENVIRON["inject"]} 1' ${package}
    fi
  fi

done
```

Zopolis4:
Every time a package is updated it generates a bunch of conflicts, could we hold off on package updates for a bit until this is merged? 
I've finished the changes and 7297 works now, so once the discussions are resolved on that and its merged this'll be ready to go, so it won't be a super long wait.

satmandu:
Is this ready to be reviewed?

Zopolis4:
#7297 has been merged (woo!) so this is good to go.

lib/package.rb
```diff
@@ -227,7 +227,15 @@ def self.get_url(architecture)
   end
 
   def self.get_binary_url(architecture)
-    return @binary_url.key?(architecture) ? @binary_url[architecture] : nil
+    if @binary_sha256.key?(architecture)
+      if architecture == x86_64 || i686 || armv7l
+        return "https://gitlab.com/api/v4/projects/26210301/packages/generic/#{name}/#{version}_#{architecture}/#{name}-#{version}-chromeos-#{architecture}.#{binary_compression}"
+      elsif architecture == aarch64
+        return "https://gitlab.com/api/v4/projects/26210301/packages/generic/#{name}/#{version}_armv71/#{name}-#{version}-chromeos-armv71.#{binary_compression}"
```
uberhacker, line 234: Where is `#{binary_compression}` set?

uberhacker (commented):
Asking a question.

lib/package.rb
Zopolis4: In packages-- I just missed the commit last rebase.

packages/broadway.rb
```diff
@@ -8,7 +8,6 @@ class Broadway < Package
   compatibility 'all'
   source_url 'SKIP'
 
-  binary_url({})
   binary_sha256({})
```
uberhacker, line 11: Remove both lines and add `compile_not_needed`?

uberhacker (commented):
I'm unable to review this pr due to too many files.  Please check all other packages similar to the one mentioned with the comment.

Zopolis4:
Done.

Zopolis4:
@satmandu @uberhacker Can we please get this merged? 

It's very difficult to keep up-to-date and rebased (I've got a custom script and four local branches and it still requires manual intervention), and all the reviews have been adressed some time ago.

satmandu:
Does `install.sh` need to be updated? Or did you do that already?

Zopolis4:
I already did that in #7297-- might be good to just quickly test an install, i'm not on my laptop rn so I can't test.

Zopolis4:
I tested an install-- it broke on `glibc_lib235`. Looking into it further, there are some packages which have binary urls which are inconsistent with the way the rest of the packages work. 

I wrote a script to detect these:
```ruby
# Add >LOCAL< lib to LOAD_PATH
$LOAD_PATH.unshift '../lib'

  next unless pkg.is_binary?(:x86_64)
  hardcoded_binary_url = "#{pkg.get_binary_url(:x86_64).gsub(/\.t\S+/, '')}"
  generated_binary_url = "https://gitlab.com/api/v4/projects/26210301/packages/generic/#{pkg.name}/#{pkg.version}_x86_64/#{pkg.name}-#{pkg.version}-chromeos-x86_64"
  if hardcoded_binary_url != generated_binary_url
    puts "#{pkg.name}"
  end
end
```

Running it outputs this:
```
anagram
apr_iconv
apr_util
ascii
asmc
biew
brackets
gimp
glibc_build227
glibc_build233
glibc_build235
glibc_dev235
glibc_lib235
imgur
linux_sources
netavark
nss
pinentry
qtfm
rust
sed
wput
xorg_intel_driver
```

So these are the packages which need to have new binaries uploaded that conform with the rest of the codebase.

satmandu:
The older glibc builds are extremely fragile. You're going to have to regenerate those packages in a manner that they work for this PR to get merged.

We absolutely can not break installs.

Zopolis4:
All the packages have been fixed, and I have tested and verified that the install works on all the architectures. We are good to go.

satmandu (approved):

PR #7104, opened 2022-05-31
gzip -> 1.12

saltedcoffii:
Needs armv7l binaries. i686 is broken, something about gzip won't know what time it is after january 2038. won't compile.

```
CREW_TESTING=1 CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew CREW_TESTING_BRANCH=gzip_1.12 crew update
```

### DIFF
--- a/packages/gzip.rb
+++ b/packages/gzip.rb
@@ -3,27 +3,15 @@ require 'package'
 class Gzip < Package
   description 'GNU Gzip is a popular data compression program originally written by Jean-loup Gailly for the GNU project.'
   homepage 'https://www.gnu.org/software/gzip/'
-  version '1.11'
+  version '1.12'
   license 'GPL-3'
   compatibility 'all'
-  source_url 'https://ftpmirror.gnu.org/gzip/gzip-1.11.tar.xz'
-  source_sha256 '9b9a95d68fdcb936849a4d6fada8bf8686cddf58b9b26c9c4289ed0c92a77907'
-
-  binary_url({
-    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/gzip/1.11_armv7l/gzip-1.11-chromeos-armv7l.tpxz',
-     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/gzip/1.11_armv7l/gzip-1.11-chromeos-armv7l.tpxz',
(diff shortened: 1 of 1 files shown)


satmandu:
no idea how to fix i686

uberhacker:
It doesn't look like the build is much different than the earlier version.  Have you tried the build on actual hardware yet?

saltedcoffii:
I'll dig around for patches, January 2038 isn't for a while, and I'm sure there's a way to fix this.

satmandu:
Maybe this might help for that build error?

https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1851491.html

(But also I don't see how this would fix things as we are already passing `--build=i686-cros-linux-gnu --host=i686-cros-linux-gnu ` to the build on i686.)

satmandu:
`LDFLAGS=-static` might also help fix this...

satmandu:
This is actually a recent gnulib bug with i686. Attempting fix now...

satmandu (approved):

satmandu:
This now builds on i686.

packages/gzip.rb
```diff
@@ -3,28 +3,59 @@
...
+       _GL_CXXALIASWARN (strndup);
+       #else
+      -# if __GNUC__ >= 11
+      +# if __GNUC__ >= 11 && !defined strndup
+       /* For -Wmismatched-dealloc: Associate strndup with free or rpl_free.  */
+       _GL_FUNCDECL_SYS (strndup, char *,
+                         (char const *__s, size_t __n)
+    PATCH_EOF
+    File.write('gnulib.patch', @gnulibpatch)
+    system 'patch -p 1 -i gnulib.patch' if ARCH == 'i686'
```
uberhacker, line 52: Would it make sense to encapsulate the entire patch logic in `if ARCH == 'i686'`?  No need to create for arm or x86_64, correct?

uberhacker (changes requested):
Only create patch for i686?

packages/gzip.rb
satmandu: Is the `return unless ARCH == 'i686'` line sufficient?

uberhacker: Yes, sorry, I didn't know you made a change until now.

uberhacker (approved):

packages/gzip.rb
satmandu: I literally just made that change today. Sorry for forgetting about it!

PR #7106, opened 2022-05-31
python buildsystem, second PR

saltedcoffii:
Changes:
- `python -m build`, via `setuptools-scm` can get versions wrong for git packages. I added an environment variable to correct this.
- upgraded dependencies of `py3_cryptography` and the package itself, most of which use python build and the associated file, `pyproject.toml`
- added two new features under `@pkg`, `git_no_submodules` and `git_clone_deep`. `git_no_submodules` omits cloning submodules when invoked, and `git_clone_deep` clones the whole git repository instead of just our little `--depth 1`, as is needed by some python packages and potentially others.

Note: needs `armv7l` binaries.

```
CREW_TESTING=1 CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew CREW_TESTING_BRANCH=python_buildsystem_2 crew update
```

To build binaries, update the packages, then run:
```
crew reinstall py3_{pyparsing,setuptools_scm,setuptools,packaging,wheel,flit_core,build,installer}
/output/crewb py3_{setuptools,pyparsing,tomli,tomli-w,pep517,packaging,wheel,setuptools_scm,flit_core,build,installer,typing_extensions,pycparser,flit,semantic_version,setuptools_rust,cryptography}`
```
I know it's not normal to reinstall things before building binaries, however the newer versions of the reinstalled packages need to be present to properly build binaries. This buildsystem is a little screwy to figure out.


lib/package.rb
```diff
@@ -6,8 +6,9 @@ class Package
            :git_branch, :git_hashtag
 
   boolean_property = %i[conflicts_ok git_fetchtags gnome is_fake is_musl
-                        is_static no_compile_needed no_env_options no_fhs 
-                        no_patchelf no_zstd patchelf]
+                        is_static no_compile_needed no_env_options no_fhs
+                        no_patchelf no_zstd patchelf git_clone_deep
+                        git_no_submodules]
```
uberhacker, line 11: Would `no_git_submodules` be better to be consistent with the others that start with `no_`?

uberhacker: It would also be nice if we had a command to print all these boolean properties.  At the very least, it should be documented somewhere other than here.

packages/py3_build.rb
```diff
@@ -4,31 +4,27 @@ class Py3_build < Package
...
+  x86_64: '820bc24d5da0c2394fa587b3b41f62b5bf4903a7eb91263ef555380f367bc9a3'
   })
 
   depends_on 'py3_packaging'
   depends_on 'py3_pep517'
   depends_on 'py3_tomli'
 
   def self.build
-    system "python3 -m build #{PY3_BUILD_OPTIONS}"
+    system "SETUPTOOLS_SCM_PRETEND_VERSION=#{@_ver} python3 -m build #{PY3_BUILD_OPTIONS}"
```
uberhacker, line 27: This is really an environment variable option?

packages/py3_pycparser.rb
```diff
@@ -0,0 +1,29 @@
...
+    i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/py3_pycparser/2.21_i686/py3_pycparser-2.21-chromeos-i686.tar.zst',
+  x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/py3_pycparser/2.21_x86_64/py3_pycparser-2.21-chromeos-x86_64.tar.zst'
+  })
+  binary_sha256({
+    i686: 'a43749ca7aceccda4465573ae593fbd4056a6ec57d331da5b8c48d3151d45af4',
+  x86_64: 'b3a54f1e0a3dd3d27a6cd3cc3a578f91cd91507d8e6e15c85b5ad29d57bfe49c'
+  })
+
+  def self.build
+    system "python3 setup.py build #{PY3_SETUP_BUILD_OPTIONS}"
```
uberhacker, line 23: Why no `python3 -m build` here?

saltedcoffii: if you look at the source code, the `python3 -m build` requires the presence of a `pyproject.toml` file. if that doesn't exist in the project, either yet or never, we use `setup.py` instead. hope that helps :)

lib/package.rb
saltedcoffii: > Would `no_git_submodules` be better to be consistent with the others that start with `no_`?

yes, i'll change that in my next commit.

packages/py3_tomli_w.rb
```diff
@@ -4,27 +4,14 @@ class Py3_tomli_w < Package
...
-       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/py3_tomli_w/1.0.0_i686/py3_tomli_w-1.0.0-chromeos-i686.tar.zst',
-     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/py3_tomli_w/1.0.0_x86_64/py3_tomli_w-1.0.0-chromeos-x86_64.tar.zst'
-  })
-  binary_sha256({
-    aarch64: '446079f23706d1e23af65ff654b8e373aec8c3202c99978faea6accc9d7c706b',
-     armv7l: '446079f23706d1e23af65ff654b8e373aec8c3202c99978faea6accc9d7c706b',
-       i686: '1f7ab0a43dd6168dee084f3f293cbcf405c060331bb819cdfac867dd654001c2',
-     x86_64: '6a386b76da03628b1a46f24996370e0a8a9eb6c9c35d0952c9bd63c122e316dc'
-  })
-
```
uberhacker, line 25: Missing binaries?

saltedcoffii: [1b26c82](https://github.com/skycocker/chromebrew/pull/7106/commits/1b26c828cbf2cfd3445e71a25a8ceda0def3779e)

PR #7109, opened 2022-06-01
Add ddgr package

uberhacker:
ddgr is a cmdline utility to search [DuckDuckGo](https://html.duckduckgo.com/html/) from the terminal.  Tested on all architectures.

Examples
DuckDuckGo hello world:

$ ddgr hello world
I'm Feeling Ducky search:

$ ddgr -j lucky ducks
DuckDuckGo Bang search hello world in Wikipedia:

$ ddgr !w hello world
$ ddgr \!w hello world // bash-specific, need to escape ! on bash
Bangs work at the omniprompt too. To look up bangs, visit https://duckduckgo.com/bang?#bangs-list.

Bang alias to fire from the cmdline, open results in a GUI browser and exit:

alias bang='ddgr --gb --np'
$ bang !w hello world
$ bang \!w hello world // bash-specific, need to escape ! on bash
Website specific search:

$ ddgr -w amazon.com digital camera
Site specific search continues at omniprompt.

Search for a specific file type:

$ ddgr instrumental filetype:mp3
Fetch results on IPL cricket from India in English:

$ ddgr -r in-en IPL cricket
To find your region parameter token visit https://duckduckgo.com/params.

Search quoted text:

$ ddgr it\'s a \"beautiful world\" in spring
Show complete urls in search results (instead of only domain name):

$ ddgr -x ddgr
Use a custom color scheme, e.g., one warm color scheme designed for Solarized Dark:

$ ddgr --colors bjdxxy hello world
$ DDGR_COLORS=bjdxxy ddgr hello world
Tunnel traffic through an HTTPS proxy, e.g., a local Privoxy instance listening on port 8118:

$ ddgr --proxy localhost:8118 hello world
By default the environment variable https_proxy (or HTTPS_PROXY) is used, if defined.

Look up n, p, o, O, q, d keywords or a result index at the omniprompt: as the omniprompt recognizes these keys or index strings as commands, you need to prefix them with d, e.g.,

d n
d g keywords
d 1


saltedcoffii:
Can we use a `.git` URL here?

uberhacker:
Does it make a difference?  This is just a binary and the man page.

saltedcoffii:
> Does it make a difference? This is just a binary and the man page.

touché. It doesn't.

uberhacker:
It doesn't make any difference to me.

packages/ddgr.rb
```diff
@@ -0,0 +1,33 @@
+require 'package'
+
+class Ddgr < Package
+  description 'Search DuckDuckGo and use !bangs from the terminal.'
+  homepage 'https://github.com/jarun/ddgr'
+  @_ver = '2.0'
+  version @_ver
+  license 'GPL-3'
+  compatibility 'all'
+  source_url 'https://github.com/jarun/ddgr,git'
```
uberhacker, line 10: `.git`, not `,git`?

PR #7115, opened 2022-06-02
rfkill -> 1.0

saltedcoffii:
needs binaries.
NOTE: rfkill is also provided by util-linux, and that seems to be the standard binary throughout Linux distros, notably, Debian, Fedora and Arch Linux. Should we make this package fake and redirect it to util-linux?

```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=rfkill_1.0 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/rfkill.rb
+++ b/packages/rfkill.rb
@@ -3,26 +3,11 @@ require 'package'
 class Rfkill < Package
   description 'Tool for enabling and disabling wireless devices.'
   homepage 'http://linuxwireless.org/en/users/Documentation/rfkill'
-  version '0.5'
+  version '1.0'
   license 'ISC'
   compatibility 'all'
-  source_url 'https://www.kernel.org/pub/software/network/rfkill/rfkill-0.5.tar.xz'
-  source_sha256 'e0ae3004215e39a6c5c36e0726558740728d16f67ebdb8bea621250f6091d86a'
-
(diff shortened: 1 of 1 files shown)


packages/rfkill.rb
```diff
@@ -3,33 +3,20 @@
...
+  source_sha256 'dffc631c611520478b8a286f57c67a35e8cb5802d376c6ca13b057365432389c'
 
   def self.build
     system "make"
   end
 
   def self.install
     system "make", "DESTDIR=#{CREW_DEST_DIR}", "SBINDIR=#{CREW_PREFIX}/sbin", "MANDIR=#{CREW_PREFIX}/share/man", "install"
+    # Manpage is automatically gzipped. Chromebrew double-gzips it when packaging.
+    system "gunzip #{CREW_DEST_MAN_PREFIX}/man8/rfkill.8.gz"
```
uberhacker, line 19: This shouldn't be necessary.  The crew build command handles this automagically.

uberhacker: Oh, well that shouldn't happen.  We should fix this then.

saltedcoffii:
closing because rfkill is in util_linux. I don't know why we have this package.

PR #7116, opened 2022-06-03
`lib/downloader`: Add an option for disabling progress bar

supechicken:
This allows the user to hide the download progress bar by setting `CREW_DOWNLOADER_SHOW_PROGBAR` to `0`, which is useful when redirecting output to log files (when outputting to log files, the `\r` control char will be ignored by text editors and show many lines of progress bar)

Tested on `x86_64`

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,6 +1,6 @@
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.23.18'
+CREW_VERSION = '1.23.19'
 
 ARCH_ACTUAL = `uname -m`.chomp
 # This helps with virtualized builds on aarch64 machines
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -81,7 +81,7 @@ else
 end
 
 # If CREW_USE_CURL environment variable exists use curl in lieu of net/http.
-CREW_USE_CURL = ENV['CREW_USE_CURL'] == '1'
+CREW_USE_CURL = ENV['CREW_USE_CURL'].eql?('1')
 
 # Use an external downloader instead of net/http if CREW_DOWNLOADER is set, see lib/downloader.rb for more info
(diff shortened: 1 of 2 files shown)


lib/downloader.rb
```diff
@@ -133,15 +133,17 @@ def http_downloader (url, filename = File.basename(url), verbose = false)
...
-          downloaded_size += chunk.size # record downloaded size, used for showing progress bar
-          if file_size.positive?
-            # calculate downloading progress percentage with the given file size
-            percentage = (downloaded_size / file_size) * 100
-            # show progress bar, file size and progress percentage
-            printf "\r""[%-#{@progBarW}.#{@progBarW}s] %9.9s %3d%%",
-                   '#' * ( @progBarW * (percentage / 100) ),
-                   human_size(file_size),
-                   percentage
+          if CREW_DOWNLOADER_SHOW_PROGBAR
```
uberhacker, line 136: `unless CREW_HIDE_PROGBAR`?

lib/const.rb
```diff
@@ -89,6 +89,8 @@
 
 # Downloader maximum retry count
 CREW_DOWNLOADER_RETRY = ( ENV['CREW_DOWNLOADER_RETRY'].to_s.empty? ) ? 3 : ENV['CREW_DOWNLOADER_RETRY'].to_i
+# Downloader show progress bar or not
+CREW_DOWNLOADER_SHOW_PROGBAR = ENV.fetch('CREW_DOWNLOADER_SHOW_PROGBAR', '1').eql?('1')
```
uberhacker, line 93: Shouldn't this default to 1?  The initial value will be nil (ie. same as zero or false) if the environment variable is not set.  Maybe you can change to `CREW_HIDE_PROGBAR` instead?

lib/downloader.rb
supechicken: Done

uberhacker:
Bump crew version in `lib/const.rb`?

supechicken:
Done

Issue #7097, opened 2022-05-30
JDK8 package does not have any binaries


saltedcoffii:
As JDK8 is a buildessential package, I am marking this as high priority. @satmandu can you work on building binaries or send me the source code on slack? I'm having trouble getting the oracle java sources.

satmandu:
The solution would be to remove it from `buildessential`.  See https://github.com/skycocker/chromebrew/issues/7094

kjperkin:
Seems like it is a transitive dependency coming in from `gtk-doc`, which depends on `fop`, which depends on `jdk8`. `fop` is needed for PDF support in `gtk-doc`.

It seems to me like there's too many packages in `buildessential`. If you take a look at debian and Ubuntu, there's only six direct dependencies, which amount to a C/C++ compiler, GNU make, and a libc.

satmandu:
Is jdk8 a build or runtime dependency for fop and gtk-doc?

uberhacker:
jdk8 is a runtime dependency of fop and gtk_doc has a runtime dependency of fop.  I agree with @kjperkin.  We need to reduce the build dependencies.

satmandu:
Why not just delete gtk_doc from `buildessential`?

saltedcoffii:
We've needed to pare down buildessential for a looong time anyway

saltedcoffii:
Okay, here's what I'm thinking: right now our Java SE from Oracle requires a login, and we can't distribute their binaries. Thankfully, OpenJDK is open source, and we can compile some binaries, but there are a couple different ways we could do this, and I wanted input from @satmandu @uberhacker @Zopolis4 and @supechicken about what y'all thought. I've also been looking at [some](https://github.com/alpinelinux/aports/blob/master/community/openjdk8/APKBUILD) [various](https://github.com/archlinux/svntogit-packages/blob/master/java8-openjdk/trunk/PKGBUILD) [build](https://github.com/gentoo/gentoo/blob/master/dev-java/openjdk-bin/openjdk-bin-8.332_p09.ebuild) [files](https://github.com/void-linux/void-packages/blob/master/srcpkgs/openjdk8/template) to see how other distros are doing it.
Building OpenJDK is a bit awkward because it requires the existence of another already-built JRE of the same version or one major version prior.
Here are a few of our options:
[...]

PR #7125, opened 2022-06-06
fix java requirements for Minecraft package

cobalt2727:
## Description

Needed to run MC versions >= 1.17, see https://github.com/MultiMC/Launcher/wiki/Using-the-right-Java for an overview

## Additional information

Works properly:
- [x] `x86_64`
- [x] `i686`
- [ ] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/cobalt2727/chromebrew.git CREW_TESTING_BRANCH=patch-1 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/minecraft.rb
+++ b/packages/minecraft.rb
@@ -12,7 +12,6 @@ class Minecraft < Package
   source_sha256 'SKIP'
 
   depends_on 'libx11'
-  depends_on 'openjdk8'
   depends_on 'openjdk17'
   depends_on 'gtk3'
   depends_on 'libcom_err'
--- a/packages/minecraft.rb
+++ b/packages/minecraft.rb
@@ -13,6 +13,7 @@ class Minecraft < Package
 
   depends_on 'libx11'
   depends_on 'jdk8'
+  depends_on 'jdk17'
   depends_on 'gtk3'
   depends_on 'libcom_err'
   depends_on 'libsecret'
--- a/packages/minecraft.rb
+++ b/packages/minecraft.rb
@@ -12,8 +12,8 @@ class Minecraft < Package
   source_sha256 'SKIP'
 
   depends_on 'libx11'
-  depends_on 'jdk8'
-  depends_on 'jdk17'
+  depends_on 'openjdk8'
+  depends_on 'openjdk17'
   depends_on 'gtk3'
   depends_on 'libcom_err'
   depends_on 'libsecret'


packages/minecraft.rb
```diff
@@ -13,6 +13,7 @@ class Minecraft < Package
 
   depends_on 'libx11'
   depends_on 'jdk8'
+  depends_on 'jdk17'
```
uberhacker, line 16: We can only have one jdk requirement.

cobalt2727: 1.16 and below need Java 8, so unless it's feasible to rework your setup to include both, I'd say switching to 17 and dropping support for older versions is the way to go

uberhacker: Have you tried removing the jdk8 dependency from this package yet?  If so, just remove the `depends_on 'jdk8'` line and submit it with this pull request.

cobalt2727: Nope, I actually don't currently have a Chromebook readily available for me to test with until the end of the summer - hence why I wanted to avoid making any possibly breaking changes like that.

saltedcoffii: Let's wait to make any changes on the Minecraft package until # #7097 is fixed. after that it will likely be that we can have multiple java versions installed concurrently

satmandu: @saltedcoffii looks like #7097 is now a won't fix, right?

cobalt2727: Where does that leave us on this PR, then? I've got an old Samsung Chromebook in my hands now, so I can run tests as needed once I get this thing set up.

taroumaru52: I appreciate this great package.
It works very good jdk8+1.165 version perfect.
Soddium Mod also works good
Minecraft cannot be installed without jdk8
I removed jdk8 and jdk17
But it doesnt work on 1.192 version

cobalt2727:
...TIL GitHub only lets you re-request a review from one person at a time - sorry for the notification spam.

packages/minecraft.rb
uberhacker: Would `depends_on 'openjdk17'` work?

cobalt2727: I've switched both Java packages to their openjdk variants.

taroumaru52: Minecraft always crash because of conflict java
Even I remove jdk8 and install jdk17,
It still keep jdk8 environment.
I hope update to jdk17 dependencies.
If I can remove jdk8 completely, it will work.

I will try openjdk Thanks

saltedcoffii (approved):
Seems to work!

packages/minecraft.rb
saltedcoffii: Marking as resolved per b12e215

packages/minecraft.rb
```diff
@@ -12,7 +12,8 @@ class Minecraft < Package
   source_sha256 'SKIP'
 
   depends_on 'libx11'
-  depends_on 'jdk8'
+  depends_on 'openjdk8'
+  depends_on 'openjdk17'
```
uberhacker, line 16: We only need one or the other.  These packages are mutually exclusive of each other.

cobalt2727: Hmm. That's a problem then - see the opening comment on this issue:

> Needed to run MC versions >= 1.17, see https://github.com/MultiMC/Launcher/wiki/Using-the-right-Java for an overview

Just to clarify, versions older than 1.17 _will still need Java 8_. Worst case scenario I say switch to 17 anyways and ditch the old versions, but being able to support both would be ideal...

uberhacker: This Minecraft version requires 17 so I say remove the `depends_on 'openjdk8'` line.

cobalt2727: I can live with that, but you're 100% sure there's no way multiple versions of Java can be supported at once, like on other Linux distros?

uberhacker: Not unless the executables have different names but that is not easy to do.  Are you 100% sure this Mindcraft release requires jdk8?

cobalt2727: Yeah, modern versions of Minecraft (and all future versions going forward) require JDK 17 and will error out or something if they're launched with version 8.

uberhacker (approved):

PR #7146, opened 2022-06-15
Update homepage URL in packages from skycocker/chromebrew to chromebrew/chromebrew

supechicken:


uberhacker:
Good call!  Bump crew version in `lib/const.rb`?

supechicken:
Done :)

PR #7147, opened 2022-06-17
Remove `glibc` dependency in all packages

supechicken:
Since all libraries provided by the `glibc` package are also available from the system as well, there is no need to include `glibc` as a runtime dependency (the only usage of this package is to provide header files for building so I moved it to `buildessential.rb`)

### Changes
- Remove `glibc` dependency from `core` and all packages
- Add `glibc` to `buildessential.rb` as a dependency

### DIFF
--- a/packages/buildessential.rb
+++ b/packages/buildessential.rb
@@ -12,8 +12,10 @@ class Buildessential < Package
   # Make sure core is installed
   depends_on 'core'
 
-  # install first to get ldconfig
+  # install libc header
   depends_on 'glibc'
+
+  # install first to get ldconfig
   depends_on 'gcc'
   depends_on 'gmp'
   depends_on 'mpfr'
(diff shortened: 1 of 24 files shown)


satmandu:
The problem here is that having glibc in package files as a runtime dependency allows us to track which packages are dependent upon glibc, and which are not. (For instance musl packages.)

How else will the dependency resolver tell us which packages are dependent upon glibc and which are not?

supechicken:
How about checking for the `is_musl` property set in package? (the reason that I want to remove `glibc` is to prevent duplicated files if possible)

satmandu:
Can you explain where the duplicate files issue happens?

I just want the dependency tree to be accurate and list which libc package a package depends upon.

You can add logic to create a hack for musl, but there are theoretically other options for a libc out there, so why create a hack which won't work for those? 

I think it is fine to add glibc to `buildessential` though, since so many build apps depend upon it.

uberhacker:
@supechicken: What is the status of this PR?

supechicken:
Closing because I have a better way to remove duplicate files while keeping `glibc` as a dependency in packages

PR #7153, opened 2022-06-19
Add curtail package and dependencies

uberhacker:
Curtail (previously ImCompressor) is an useful image compressor, supporting PNG, JPEG and WEBP file types.  Tested on x86_64.  Unable to test on arm since sommelier is broken but the builds were successful.

New dependencies include:
pngquant - Command-line utility and a library for lossy compression of PNG images.
jpegoptim - Utility to optimize/compress JPEG files.

### DIFF
--- a/packages/curtail.rb
+++ b/packages/curtail.rb
@@ -1,0 +1,41 @@
+require 'package'
+
+class Curtail < Package
+  description 'Curtail (previously ImCompressor) is an useful image compressor, supporting PNG, JPEG and WEBP file types.'
+  homepage 'https://github.com/Huluti/Curtail'
+  version '1.3.0'
+  license 'GPL-3'
+  compatibility 'all'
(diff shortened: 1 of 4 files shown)


saltedcoffii:
let's use a .git url since we're just downloading a tarball anyway

uberhacker:
@saltedcoffii: I'm not convinced that's the best approach.  A git clone doesn't have a checksum and it's a slower way to get the source.

saltedcoffii:
git clone has a built in checksum

uberhacker:
There is always the unlikely possibility of a sha1 hash attack.  Having the checksum in the package is safer.  Also, if a repository is big, it will definitely take longer to pull down the source no matter how fast is your internet connection.

satmandu (approved):

satmandu (approved):

saltedcoffii:
...thankfully git doesn't use SHA1 anymore

supechicken (approved):

PR #7164, opened 2022-06-23
Add gosu package

uberhacker:
This is a simple tool grown out of the simple fact that su and sudo have very strange and often annoying TTY and signal-forwarding behavior. They're also somewhat complex to setup and use (especially in the case of sudo), which allows for a great deal of expressivity, but falls flat if all you need is "run this specific application as this specific user and get out of the pipeline".

The core of how gosu works is stolen directly from how Docker/libcontainer itself starts an application inside a container (and in fact, is using the /etc/passwd processing code directly from libcontainer's codebase).
```
$ gosu
Usage: gosu user-spec command [args]
   eg: gosu tianon bash
       gosu nobody:root bash -c 'whoami && id'
       gosu 1000:1 id

gosu version: 1.14 (go1.16.7 on linux/amd64; gc)
gosu license: Apache-2.0 (full text at https://github.com/tianon/gosu)
```
Once the user/group is processed, we switch to that user, then we exec the specified process and gosu itself is no longer resident or involved in the process lifecycle at all. This avoids all the issues of signal passing and TTY, and punts them to the process invoking gosu and the process being invoked by gosu, where they belong.

See https://github.com/tianon/gosu.

### DIFF
--- a/packages/gosu.rb
+++ b/packages/gosu.rb
@@ -1,0 +1,28 @@
+require 'package'
+
+class Gosu < Package
+  description 'Simple Go-based setuid+setgid+setgroups+exec'
+  homepage 'https://github.com/tianon/gosu'
+  version '1.14'
+  license 'Apache-2.0'
+  compatibility 'all'
+  source_url ({
+    aarch64: 'https://github.com/tianon/gosu/releases/download/1.14/gosu-armhf',
+     armv7l: 'https://github.com/tianon/gosu/releases/download/1.14/gosu-armhf',
+       i686: 'https://github.com/tianon/gosu/releases/download/1.14/gosu-i386',
+     x86_64: 'https://github.com/tianon/gosu/releases/download/1.14/gosu-amd64'
+  })
+  source_sha256 ({
+    aarch64: 'abb1489357358b443789571d52b5410258ddaca525ee7ac3ba0dd91d34484589',
+     armv7l: 'abb1489357358b443789571d52b5410258ddaca525ee7ac3ba0dd91d34484589',
+       i686: 'fa12da9c07783fe5fb2c2a81b5d4277eec7615c957a0b0cce3f14802ff9372e7',
+     x86_64: 'bd8be776e97ec2b911190a82d9ab3fa6c013ae6d3121eea3d0bfd5c82a0eaf8c'
+  })
+
+  no_compile_needed
+
+  def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
+    FileUtils.install Dir['gosu-*'][0], "#{CREW_DEST_PREFIX}/bin/gosu", mode: 0o755
+  end
+end
--- a/tools/packages.yaml
+++ b/tools/packages.yaml
@@ -2399,6 +2399,11 @@ url: https://ftp.gnome.org/pub/gnome/sources/gobject-introspection/
 activity: medium
 ---
 kind: url
+name: gosu
+url: https://github.com/tianon/gosu/releases
+activity: low
+---
+kind: url
 name: gox
 url: https://github.com/mitchellh/gox/releases
 activity: low


satmandu (approved):

saltedcoffii:
@uberhacker have you tested it on a device. our versions of `sudo` are notorious for not working, does this work? do you have to be root to use it properly?

uberhacker:
Yes, you can try the examples if you want.  This worked for me on x86_64:
```
$ sudo gosu cros:cros bash -c 'whoami && id'
```

saltedcoffii:
okay it does seem to work, what's the point of this. like how would it be useful to chromebrew users

uberhacker:
The usefulness is you can run as another user without switching accounts.

saltedcoffii:
okay cool. why are we using precompiled releases instead of compiling binaries ourselves with our triple and other config?

satmandu:
I have no problem using pre-built binaries, especially for small binaries which are built for all of our architectures!

The sudo/su packages are already landmines...

packages/gosu.rb
```diff
@@ -0,0 +1,27 @@
+require 'package'
+
+class Gosu < Package
+  description 'Simple Go-based setuid+setgid+setgroups+exec'
+  homepage 'https://github.com/tianon/gosu'
+  version '1.14'
+  license 'Apache-2.0'
+  compatibility 'all'
+  source_url 'SKIP'
+
```
supechicken, line 10: Add `no_compile_needed` flag?

packages/gosu.rb
```diff
@@ -0,0 +1,27 @@
...
+  def self.install
+    case ARCH
+    when 'aarch64', 'armv7l'
+      url = 'https://github.com/tianon/gosu/releases/download/1.14/gosu-armhf'
```
supechicken, line 14: Use `source_url` and install the file with `Dir['gosu-*']`?

packages/gosu.rb
uberhacker: Good call.  Added!

packages/gosu.rb
uberhacker: No sure what you mean.  Is that a simplified way to install?

uberhacker: That won't work because the binaries are not in a compressed format.  They simply need to be downloaded and intsalled.

supechicken: > Not sure what you mean. Is that a simplified way to install?

Kind of :) It takes lesser lines of code to achieve
```ruby
source_url ({
  x86_64: '...',
  ...
})

def self.install
  FileUtils.install Dir['gosu-*'], ...
end
```

> That won't work because the binaries are not in a compressed format. They simply need to be downloaded and intsalled.
The non-archive file support for `source_url` has been added in #6743

packages/gosu.rb
saltedcoffii: Will the `no_compile_needed` flag disable UPX?

supechicken: No, it will only disable `strip`

packages/gosu.rb
uberhacker: It doesn't work. :)  See below:
```
require 'package'

class Gosu < Package
  description 'Simple Go-based setuid+setgid+setgroups+exec'
  homepage 'https://github.com/tianon/gosu'
  version '1.14'
  license 'Apache-2.0'
  compatibility 'all'
  source_url ({
    aarch64: 'https://github.com/tianon/gosu/releases/download/1.14/gosu-armhf',
     armv7l: 'https://github.com/tianon/gosu/releases/download/1.14/gosu-armhf',
       i686: 'https://github.com/tianon/gosu/releases/download/1.14/gosu-i386',
     x86_64: 'https://github.com/tianon/gosu/releases/download/1.14/gosu-amd64'
  })
  source_sha256 ({
    aarch64: 'abb1489357358b443789571d52b5410258ddaca525ee7ac3ba0dd91d34484589',
     armv7l: 'abb1489357358b443789571d52b5410258ddaca525ee7ac3ba0dd91d34484589',
       i686: 'fa12da9c07783fe5fb2c2a81b5d4277eec7615c957a0b0cce3f14802ff9372e7',
     x86_64: 'bd8be776e97ec2b911190a82d9ab3fa6c013ae6d3121eea3d0bfd5c82a0eaf8c'
  })

  no_compile_needed

  def self.install
    FileUtils.install Dir['gosu-*'], "#{CREW_DEST_PREFIX}/bin/gosu", mode: 0o755
  end
end
```
With the changes above, I get:
```
$ crew install gosu
gosu: Simple Go-based setuid+setgid+setgroups+exec
https://github.com/tianon/gosu
Version: 1.14
License: Apache-2.0
[##############################################################################################################################################################################################]   2.18 MB 100%
gosu-amd64: File downloaded.
Preconfiguring package...
/usr/local/lib64/ruby/3.1.0/fileutils.rb:1395:in `initialize': No such file or directory @ rb_sysopen - /usr/local/tmp/crew/dest/usr/local/bin/gosu/gosu-amd64 (Errno::ENOENT)
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:1395:in `open'
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:1395:in `block in copy_file'
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:1394:in `open'
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:1394:in `copy_file'
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:504:in `copy_file'
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:868:in `block in install'
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:1577:in `block in fu_each_src_dest'
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:1586:in `block in fu_each_src_dest0'
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:1584:in `each'
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:1584:in `fu_each_src_dest0'
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:1575:in `fu_each_src_dest'
	from /usr/local/lib64/ruby/3.1.0/fileutils.rb:864:in `install'
	from /usr/local/lib/crew/packages/gosu.rb:26:in `install'
	from /usr/local/bin/crew:927:in `block in build_and_preconfigure'
	from /usr/local/bin/crew:901:in `chdir'
	from /usr/local/bin/crew:901:in `build_and_preconfigure'
	from /usr/local/bin/crew:1399:in `install'
	from /usr/local/bin/crew:1307:in `resolve_dependencies_and_install'
	from /usr/local/bin/crew:1791:in `block in install_command'
	from /usr/local/bin/crew:1786:in `each'
	from /usr/local/bin/crew:1786:in `install_command'
	from /usr/local/bin/crew:1908:in `<main>'
```

supechicken: Oops, it should be `FileUtils.install Dir['gosu-*'][0], "#{CREW_DEST_PREFIX}/bin/gosu", mode: 0o755`😅

uberhacker: Nice, changes made.

supechicken:
> The usefulness is you can run as another user without switching accounts.

Sounds like it does the same thing as `sudo -u <USER> <COMMAND>`😀

uberhacker:
Read the above again about the behavior of su and sudo.

supechicken:
> $ sudo gosu cros:cros bash -c 'whoami && id'

But it still needs to use `sudo` to `setuid()` as another user properly...

Would it be a better way to gain the `suid` bit to the `gosu` binary so that we don't need `sudo` to run it anymore?

uberhacker:
> > $ sudo gosu cros:cros bash -c 'whoami && id'
> 
> But it still needs `sudo` to `setuid()` as another user properly...
> 
> Would it be a better way to gain the `suid` bit to the `gosu` binary so that we don't need `sudo` to run it anymore?

Not sure what question you are asking.  I think this command was created as a way to avoid the issues with su or sudo.  My understanding it can be used in Docker builds to streamline the process of scripts.

supechicken:
> Not sure what question you are asking. I think this command was created as a way to avoid the issues with su or sudo. My understanding it can be used in Docker builds to streamline the process of scripts.

The `gosu` command needs to combine with `sudo` in order to run the command as another user (`sudo` needs to be involved anyways), right?

If we give the `suid` bit (similar to the executable bit, but the binary can run as `root` with this bit (thats how the `sudo` binary works) ) to the `gosu` binary, so that we don't need `sudo` anymore (but we need root to gain the `suid` bit at installation)

uberhacker:
I don't think I have to explain the security hole that would open if we allowed the chronos user to run `gosu` as root.

supechicken:
> I don't think I have to explain the security hole that would open if we allowed the chronos user to run `gosu` as root.

The point is `sudo` is necessary in order to use `gosu`... (those TTY behavior are still exists)

PR #7180, opened 2022-07-04
Update libpng, fontconfig and harfbuzz packages

uberhacker:
These packages build fine without any special tweaking.  I was missing libpng too many times and had to fix it. :)

### DIFF
--- a/packages/fontconfig.rb
+++ b/packages/fontconfig.rb
@@ -2,90 +2,46 @@ require 'package'
 
 class Fontconfig < Package
   description 'Fontconfig is a library for configuring and customizing font access.'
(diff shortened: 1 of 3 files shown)


satmandu:
I think we will also have to adjust depends elsewhere which list harfbuzz as a dep when they mean libpng...

uberhacker:
Yes, good point.  Do we need the extra cache stuff for fontconfig?

uberhacker:
@satmandu: The fontconfig and harfbuzz packages have the libpng dependency now so no need to change for other packages?

satmandu (approved):

PR #7193, opened 2022-07-11
[WIP] Ldc 1.7.0 => 1.29.0

uberhacker:
Currently unable to build the standalone d compiler required to build ldc.

### DIFF
--- a/packages/ldc.rb
+++ b/packages/ldc.rb
@@ -1,49 +1,38 @@
 require 'package'
 
-class Ldc < Package                 # The first character of the class name must be upper case
+class Ldc < Package
   description 'D language compiler using LLVM.'
   homepage 'https://github.com/ldc-developers/ldc'
-  version '1.7.0'
+  version '1.29.0'
   license 'BSD-3, Boost-1.0, MIT, NCSA and Artistic-1.0 or GPL-2+'
   compatibility 'all'
-  source_url 'https://github.com/ldc-developers/ldc/releases/download/v1.7.0/ldc-1.7.0-src.tar.gz'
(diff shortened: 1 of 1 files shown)


satmandu:
As per https://github.com/ldc-developers/ldc/issues/3138 it appears that debian now builds ldc with gdmd using a multi-stage compile, which is now supported with the `-DD_COMPILER=/usr/bin/gdmd` flags... See https://salsa.debian.org/d-team/ldc/-/blob/debian/master/debian/rules

But that would require building a gdmd package... 

That's... a commitment.

uberhacker:
Apparently, the gcc build needs gdc before it can enable the d language.  Another chicken before the egg scenario.

saltedcoffii:
What's the default D compiler on Debian and Arch?

Zopolis4:
> 
> But that would require building a gdmd package...
> 
> That's... a commitment.

Building [this](https://github.com/D-Programming-GDC/gdmd)???

Doesn't seem too hard, unless I'm missing something.

uberhacker:
> > But that would require building a gdmd package...
> > That's... a commitment.
> 
> Building [this](https://github.com/D-Programming-GDC/gdmd)???
> 
> Doesn't seem too hard, unless I'm missing something.

I didn't have much luck but you are welcome to give it go.

Zopolis4:
I have figured it out. gdmd is simple to build, gdc-12 is a little harder. One must first build gdc-9, gdc-10 or gdc-11, then use that to build gdc-12. I will build these once gcc 12.3 releases, as to avoid using `@_ver`.

PR #7215, opened 2022-07-19
gl.sh: Add progress bar to uploads.

satmandu:
- Make `gl.sh` output more verbose, to indicate when uploads are occurring.
- We need the `| cat` in the `curl` commandline because otherwise `curl` doesn't show the progress bar.

Works properly:
- [x] `x86_64`

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=gl_edit CREW_TESTING=1 crew update
```

### DIFF
--- a/tools/gl.sh
+++ b/tools/gl.sh
@@ -84,11 +84,11 @@ for package in ${packages}; do
     fi
     echo "Generating sha256sum ..."
     new_sha256=$(sha256sum ${new_tarfile} | cut -d' ' -f1)
-    echo
     noname="${new_tarfile#*-}"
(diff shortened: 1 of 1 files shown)


uberhacker:
Nice!  There is another issue with this script.  The binary hashes are always generated regardless since it expects .tar.xz instead of .zst extension.

uberhacker (approved):

satmandu:
> Nice! There is another issue with this script. The binary hashes are always generated regardless since it expects .tar.xz instead of .zst extension.

Yes but it just always generates hashes now...

PR #7231, opened 2022-07-25
Add xcurses package

uberhacker:
This is a port of PDCurses for X11, aka XCurses. It is designed to allow existing curses programs to be re-compiled with PDCurses, resulting in native X11 programs.  See https://github.com/wmcbrine/PDCurses/blob/master/x11/README.md.

### DIFF
--- a/packages/xcurses.rb
+++ b/packages/xcurses.rb
@@ -1,0 +1,45 @@
+require 'package'
+
+class Xcurses < Package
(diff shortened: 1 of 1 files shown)


saltedcoffii:
the packaging itself looks good 👍😁 but what will this package do for chromebrew? like why add it?

satmandu (approved):

saltedcoffii:
okay awesome

PR #7280, opened 2022-08-16
protobuf -> 21.5

satmandu:
Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=protobuf CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/protobuf.rb
+++ b/packages/protobuf.rb
@@ -1,41 +1,47 @@
+# Adapted from Arch Linux protobuf PKGBUILD at:
+# https://github.com/archlinux/svntogit-packages/raw/packages/protobuf/trunk/PKGBUILD
+
 require 'package'
 
 class Protobuf < Package
-  description 'Protocol buffers are a language-neutral, platform-neutral extensible mechanism for serializing structured data.'
+  description 'Protocol Buffers - Googles data interchange format'
   homepage 'https://developers.google.com/protocol-buffers/'
-  version '3.9.2'
+  @_ver = '21.5'
+  version @_ver
   license 'BSD'
   compatibility 'all'
-  source_url 'https://github.com/protocolbuffers/protobuf/archive/v3.9.2.tar.gz'
-  source_sha256 '1fbf1c2962af287607232b2eddeaec9b4f4a7a6f5934e1a9276e9af76952f7e0'
+  source_url "https://github.com/protocolbuffers/protobuf/archive/v#{@_ver}/protobuf-#{@_ver}.tar.gz"
+  source_sha256 '4a7e87e4166c358c63342dddcde6312faee06ea9d5bb4e2fa87d3478076f6639'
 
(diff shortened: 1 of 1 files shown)


packages/protobuf.rb
```diff
@@ -1,41 +1,47 @@
...
+# https://github.com/archlinux/svntogit-packages/raw/packages/protobuf/trunk/PKGBUILD
+
 require 'package'
 
 class Protobuf < Package
-  description 'Protocol buffers are a language-neutral, platform-neutral extensible mechanism for serializing structured data.'
+  description 'Protocol Buffers - Googles data interchange format'
   homepage 'https://developers.google.com/protocol-buffers/'
-  version '3.9.2'
+  @_ver = '21.5'
```
uberhacker, line 9: Wow, that is quite a jump.

uberhacker (approved):

packages/protobuf.rb
satmandu: Right? They totally changed versioning...

supechicken (approved):

packages/protobuf.rb
```diff
@@ -1,41 +1,47 @@
...
-           "--libdir=#{CREW_LIB_PREFIX}"
-    system 'make'
+    FileUtils.mkdir('builddir')
+    Dir.chdir('builddir') do
+      system "cmake #{CREW_CMAKE_OPTIONS} \
+      -Dprotobuf_BUILD_SHARED_LIBS=ON \
+      -Dprotobuf_BUILD_TESTS=OFF \
+      ../ -G Ninja"
+    end
+    system 'samu -C builddir'
```
supechicken, line 41: Move `samu` into the `Dir.chdir` block?

supechicken: (or use `chdir: 'builddir'` argument in `cmake` line)

satmandu: Looking at usage in other packages:

```
/usr/local/lib/crew/packages $ grep "'samu -C builddir'" * | wc -l
34
```
So this is standard usage for many packages... why change it?

supechicken: Now we have 2 `chdir` calls, one in `Dir.chdir` and one is done by `samu` (according by the `-C` argument)

My through is to reduce the `chdir` system call if possible (although that might not make it faster)

satmandu: `samu` is already opening in a subshell. I suspect the increase in speed from removing an internal chdir call for samu is de minimus... and is that even a chdir? samu is merely being told which dir to start working from when it is invoked.

@uberhacker @saltedcoffii Thoughts?

uberhacker: I think the code is fine the way it is now.

Issue #7224, opened 2022-07-23
Sommelier not working on arm device


justinkkwan:
## Description

After installing Sommelier through Chromebrew on my Lenovo 10e, Sommelier can not start.

If you executed a specific command:
```shell
crew install sommelier
source ~/.bashrc
```

Output:
```
some sommelier processes failed to start
sommelier processess running:

```

## Versions and system information

***
fatal: ambiguous argument 'HEAD^..HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
- Architecture: `aarch64` (`aarch64`)
- Kernel version: `5.10.114-16025-ge75506b9d98e`

- Chromebrew version: `1.24.0`
- Chromebrew prefix: `/usr/local`
- Chromebrew libdir: `/usr/local/lib`

- Last update in local repository: ``

- OS variant: `Chrome OS`
- OS version: `kukui-release/R103-14816.131.0`
- OS channel: `stable-channel`
***

## Additional information

After this failed, I tried `crew reinstall sommelier` but that didn't help.
Then I did powerwashed (twice, actually) and did nothing other than the installation steps from the readme and installing sommelier.
When it still failed, I tried another `crew reinstall sommelier` which did not help.
In every case, the same thing I pasted above was printed to the console.

uberhacker:
@justinkkwan: Can you please provide the output from `/usr/local/var/log/sommelier.log`?

justinkkwan:
Contents of `/usr/local/var/log/sommelier.log`:
```
MESA-LOADER: failed to open vgem: /usr/local/lib/dri/vgem_dri.so: cannot open shared object file: No such file or directory (search paths /usr/local/lib/dri, suffix _dri)
failed to load driver: vgem
MESA-LOADER: failed to open kms_swrast: /usr/local/lib/libLLVM-14.so: cannot open shared object file: No such file or directory (search paths /usr/local/lib/dri, suffix _dri)
failed to load driver: kms_swrast
MESA-LOADER: failed to open swrast: /usr/local/lib/libLLVM-14.so: cannot open shared object file: No such file or directory (search paths /usr/local/lib/dri, suffix _dri)
failed to load swrast driver
error: couldn't get display device
MESA-LOADER: failed to open vgem: /usr/local/lib/dri/vgem_dri.so: cannot open shared object file: No such file or directory (search paths /usr/local/lib/dri, suffix _dri)
[...]

PR #7285, opened 2022-08-17
crew: Update `git` command used in `crew sysinfo`

supechicken:
Seems the `git` command that uses to get the latest commit time does not work on every system, weird...

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1853,7 +1853,7 @@ def sysinfo_command(args)
     lsb_release = ENV
   end
 
-  git_commit_message_format = '%C(yellow)%h%C(reset)  %s  %C(brightcyan)(%cr)%C(reset)'
+  git_commit_message_format = '%h  %s  (%cr)'
 
   puts <<~MD
     - Architecture: `#{ARCH_ACTUAL}` (`#{ARCH}`)
@@ -1863,7 +1863,7 @@ def sysinfo_command(args)
     - Chromebrew prefix: `#{CREW_PREFIX}`
     - Chromebrew libdir: `#{CREW_LIB_PREFIX}`
 
-    - Last update in local repository: `#{`git show --color=always -s --format='#{git_commit_message_format}'`.chomp}`
+    - Last update in local repository: `#{`git show -s --format='#{git_commit_message_format}' '#{CREW_LIB_PATH}'`.chomp}`
 
     - OS variant: `#{lsb_release['CHROMEOS_RELEASE_NAME']}`
     - OS version: `#{lsb_release['CHROMEOS_RELEASE_BUILDER_PATH']}`
--- a/bin/crew
+++ b/bin/crew
@@ -1853,6 +1853,8 @@ def sysinfo_command(args)
     lsb_release = ENV
   end
 
+  git_commit_message_format = '%C(yellow)%h%C(reset)  %s  %C(brightcyan)(%cr)%C(reset)'
+
   puts <<~MD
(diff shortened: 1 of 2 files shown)


satmandu:
Where does it break?

satmandu:
Seems that we REALLY want the last update time though, no?

Could this just be due to really out of date chromebrew setups?

supechicken:
> Seems that we REALLY want the last update time though, no?
> 
> Could this just be due to really out of date chromebrew setups?

But #7224 is from 24 days ago...

Finding another way to get the latest commit time...

satmandu:
Something like `git show -s --format=%H` would show the last commit...

supechicken:
> Something like git show -s --format=%H would show the last commit...

It also works with colors :)
```shell
git show --color=always -s --format='%C(yellow)%h%C(reset)  %s  %C(brightcyan)(%cr)%C(reset)' | cat
```

supechicken:
Added a few changes and now it looks like this:
- Kernel version: `5.10.120-16471-g167ad2e5dcba`

- Chromebrew version: `1.24.1`

- Last update in local repository: 8ae1b0a `Remove color as markdown does not support (70 seconds ago)`

- OS variant: `Chrome OS`
- OS version: `reven-release/R104-14909.100.0`

bin/crew
```diff
@@ -1863,7 +1863,7 @@ def sysinfo_command(args)
     - Chromebrew prefix: `#{CREW_PREFIX}`
     - Chromebrew libdir: `#{CREW_LIB_PREFIX}`
 
-    - Last update in local repository: `#{`git show -s --format='#{git_commit_message_format}' '#{CREW_LIB_PATH}'`.chomp}`
+    - Last update in local repository: #{`git show -s --format='#{git_commit_message_format}' '#{CREW_LIB_PATH}'`.chomp}
```
uberhacker, line 1866: Is the outer `#{...}` necessary?

bin/crew
```diff
@@ -1853,7 +1853,7 @@ def sysinfo_command(args)
     lsb_release = ENV
   end
 
-  git_commit_message_format = '%h  %s  (%cr)'
+  git_commit_message_format = '%h `%s (%cr)`'
```
uberhacker, line 1856: Why the backticks?

uberhacker (changes requested):

bin/crew
supechicken: It is for markdown styling (\`abc\` shows `abc`)

bin/crew
supechicken: Seems this syntax highlight of github is a bit misleading😅, maybe this more clear:
```ruby
"- Last update in local repository: #{IO.popen("git show -s --format='#{git_commit_message_format}' '#{CREW_LIB_PATH}").read.chomp}"

bin/crew
uberhacker: Ummm, I don't think that is going to work like you might think.   Backticks are used to EXECUTE commands.

supechicken: This is not `bash`... In ruby only `#{}` in `""` will be executed (that means the \` does not have any special meaning in strings)

supechicken: Oh I see what you mean😅 The `#{git_commit_message_format}` in the git command is wrapped with `''` so it will not be executed

uberhacker: Backticks work in ruby too.  I think you are right about the single quotes.

uberhacker (approved):

PR #7308, opened 2022-08-20
crew_profile_base => 0.0.5

supechicken:
Tested on `x86_64`

### DIFF
--- a/packages/crew_profile_base.rb
+++ b/packages/crew_profile_base.rb
@@ -3,8 +3,8 @@ require 'package'
 class Crew_profile_base < Package
   description 'Crew-profile-base sets up Chromebrew\'s environment capabilities.'
   homepage 'https://github.com/chromebrew/crew-profile-base'
-  @_ver = '0.0.4'
-  version @_ver + '-1'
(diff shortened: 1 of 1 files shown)


satmandu:
What's the reason for this change?

supechicken:
> What's the reason for this change?

Since we are not using the `master` branch in this recipe, the git tag here needs to be bumped after each pr in `crew-profile-base` repo

satmandu:
But why is `system 'sed', '-i', 's,${TMPDIR},${TMPDIR:-/usr/local/tmp},g', './src/env.d/00-path'` removed?

supechicken:
> But why is `system 'sed', '-i', 's,${TMPDIR},${TMPDIR:-/usr/local/tmp},g', './src/env.d/00-path'` removed?

I committed the fix directly after https://github.com/chromebrew/crew-profile-base/pull/9 merged (https://github.com/chromebrew/crew-profile-base/commit/c68720df6d84b8e45c159f51de3699fdcf372fd5)

uberhacker (approved):
I like when patches are removed.

PR #7318, opened 2022-08-22
libsdl2 -> 2.24.0

satmandu:
BUILDS properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

```
CREW_TESTING_REPO=https://github.comsatmandu/chromebrew.git CREW_TESTING_BRANCH=libsdl2-2.24.0 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/libsdl2.rb
+++ b/packages/libsdl2.rb
@@ -5,8 +5,8 @@ class Libsdl2 < Package
   homepage 'http://www.libsdl.org'
   @_ver = '2.24.0'
   version @_ver
-  compatibility 'all'
   license 'ZLIB'
+  compatibility 'all'
   source_url 'https://github.com/libsdl-org/SDL.git'
   git_hashtag "release-#{@_ver}"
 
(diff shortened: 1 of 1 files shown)


packages/libsdl2.rb
```diff
@@ -3,24 +3,24 @@
 class Libsdl2 < Package
   description 'Simple DirectMedia Layer is a cross-platform development library designed to provide low level access to audio, keyboard, mouse, joystick, and graphics hardware via OpenGL and Direct3D.'
   homepage 'http://www.libsdl.org'
-  @_ver = '2.0.22'
+  @_ver = '2.24.0'
   version @_ver
-  license 'ZLIB'
   compatibility 'all'
+  license 'ZLIB'
```
uberhacker, line 9: Why was this line moved?

uberhacker (dismissed):
Return license line to the original location to be consistent with other packages.

packages/libsdl2.rb
satmandu: Rubocop...

satmandu: I changed it back.

supechicken (approved):

satmandu:
> Return license line to the original location to be consistent with other packages.

Ideally we can have a CI app to set the right order for these?

@Zopolis4 any ideas?

Zopolis4:
> > Return license line to the original location to be consistent with other packages.
> 
> Ideally we can have a CI app to set the right order for these?
> 
> @Zopolis4 any ideas?

I'll have a look, I can't think of any particularly neat solution right now though.

PR #7341, opened 2022-08-24
Fix glibc symlinks not being made properly

### LINKED ISSUE #7328: remmina is missing rdp plugin
## Description

remmina has rdp plugin by default in linux but its missing here. i tried the -s option too but it bugged out. i tried searching for the plugin with crew but couldnt find anything
```shell
# your command here
```
remmina
Output:
```
# output of your command here
```

***
***

satmandu:
Fixes #7328

- Previously the glibc postinstall was making symlinks for the system provided libraries in here:
```
@libraries = %w[ld libBrokenLocale libSegFault libanl libc libcrypt
                    libdl libm libmemusage libmvec libnsl libnss_compat libnss_db
                    libnss_dns libnss_files libnss_hesiod libpcprofile libpthread
                    libresolv librlv librt libthread_db-1.0 libutil]
```
exclusively from `/#{ARCH_LIB}/` but many of these files are actually in `/usr/#{ARCH_LIB}/`, so we need to symlink from there.
BUT we also have to make sure we select only the libraries we want and don't wildcard glob onto everything... AND we need to avoid selecting folders.

- Upon crew upgrade this should force glibc to reinstall, which will generate the symlinks properly.

Works properly:
- [x] `x86_64` 
- [x] `i686`
- [x] `armv7l`

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=glibc_fix2 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/glibc.rb
+++ b/packages/glibc.rb
@@ -599,14 +599,14 @@ class Glibc < Package
       puts 'Creating symlinks to system glibc version to prevent breakage.'.lightblue
       @libraries.each do |lib|
         # Reject entries which aren't libraries ending in .so, and which aren't files.
-        Dir.glob("/#{ARCH_LIB}/#{lib}.so*").reject { |f| File.directory?(f) }.each do |f|
+        Dir["/#{ARCH_LIB}/#{lib}.so*"].reject { |f| File.directory?(f) }.each do |f|
           if `file #{f} | grep "shared object"`
             g = File.basename(f)
             FileUtils.ln_sf f.to_s, g.to_s
           end
         end
         # Reject entries which aren't libraries ending in .so, and which aren't files.
-        Dir.glob("/usr/#{ARCH_LIB}/#{lib}.so*").reject { |f| File.directory?(f) }.each do |f|
+        Dir["/usr/#{ARCH_LIB}/#{lib}.so*"].reject { |f| File.directory?(f) }.each do |f|
           if `file #{f} | grep "shared object"`
             g = File.basename(f)
             FileUtils.ln_sf f.to_s, g.to_s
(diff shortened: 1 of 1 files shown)


satmandu:
@kunver400 

Please try this:
```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=glibc_fix2 CREW_TESTING=1 crew update ; crew upgrade
```

packages/glibc.rb
```diff
@@ -598,7 +598,15 @@ def self.postinstall
       puts "System glibc version is #{LIBC_VERSION}.".lightblue
       puts 'Creating symlinks to system glibc version to prevent breakage.'.lightblue
       @libraries.each do |lib|
-        Dir.glob("/#{ARCH_LIB}/#{lib}*").each do |f|
+        # Reject entries which aren't libraries ending in .so, and which aren't files.
+        Dir.glob("/#{ARCH_LIB}/#{lib}.so*").reject { |f| File.directory?(f) }.each do |f|
```
uberhacker, line 602: Use `Dir["/#{ARCH_LIB}/#{lib}.so*"]` instead?

uberhacker (changes requested):
See comments.

packages/glibc.rb
```diff
@@ -598,7 +598,15 @@ def self.postinstall
...
+          if `file #{f} | grep "shared object"`
+            g = File.basename(f)
+            FileUtils.ln_sf f.to_s, g.to_s
+          end
+        end
+        # Reject entries which aren't libraries ending in .so, and which aren't files.
+        Dir.glob("/usr/#{ARCH_LIB}/#{lib}.so*").reject { |f| File.directory?(f) }.each do |f|
```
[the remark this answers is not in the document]
satmandu: done.

packages/glibc.rb
satmandu: done.

uberhacker (approved):

PR #7347, opened 2022-08-24
adjustments for new install script

satmandu:
- Just adds the `ca_certiticates` and `core` from https://github.com/chromebrew/chromebrew/pull/7297
- Also one minor rubocop fixup for `glibc`

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=install_refactor_deps  CREW_TESTING=1 crew update
```


packages/core.rb
```diff
@@ -79,6 +81,7 @@ class Core < Package
   depends_on 'uchardet'
   depends_on 'unzip'
   depends_on 'xzutils'
+  depends_on 'xxhash'
```
uberhacker, line 84: We're supposed to be reducing the core dependencies. :)

uberhacker (approved):

packages/core.rb
satmandu: Yes but that's an rsync dep!

PR #7376, opened 2022-09-06
Add script and action for generating JSON for Repology

Zopolis4:
When finished, will resolve repology/repology-updater#1236.

Having a couple of issues with loading package.rb, would welcome some help. Ruby is not my specialty.

### DIFF
--- a/tools/json.rb
+++ b/tools/json.rb
@@ -1,0 +1,16 @@
+require 'json'
+
+# Add >LOCAL< lib to LOAD_PATH
+$LOAD_PATH.unshift '../lib'
+
+require_relative '../lib/const'
+require_relative '../lib/package'
+
+output = Array.new
+
+Dir.glob('../packages/*.rb').each do |filename|
+  pkg = Package.load_package(filename)
+  output << {name: File.basename(filename, '.rb'), description: pkg.description, homepage: pkg.homepage, version: pkg.version, license: pkg.license, compatibility: pkg.compatibility}
+end
+
+File.write('repology.json', JSON.generate(output))
--- a/.github/workflows/Repology.yml
+++ b/.github/workflows/Repology.yml
@@ -1,0 +1,23 @@
+name: Generate Repology JSON
+
+on:
+  schedule:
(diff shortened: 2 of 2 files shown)


tools/json.rb
```diff
@@ -0,0 +1,11 @@
...
+require 'fileutils'
+require_relative '../lib/package'
+
+Dir['../packages/*.rb'].each do |filename|
+  puts filename.to_s
+  pkgName = File.basename filename, '.rb'
+  puts pkgName.to_s
+  @pkg = Package.load_package(filename, pkgName)
+  puts pkg.name.to_s
+end
```
uberhacker, line 11: What exactly does this do?

Zopolis4: The aim is to read each package file and print certain variables (name, url, version, etc) into a json file for repology. The various `puts` calls were for debugging.

uberhacker: How does this hep with updates?  I can see where it could be used for testing whether the filename and class name match but just printing the package info is for what purpose?  Is this supposed to replace packages.yaml?

uberhacker: By the way, we should not be submitting pull requests with debug lines without setting the Draft status.

Zopolis4:
Sorry, got distracted :p
Feel free to have a go!

Zopolis4:
Sorry, got distracted. :p
Feel free to have a go!

satmandu:
@supechicken @Zopolis4 What's the status on this?

Zopolis4:
I haven't made any progress, working on other stuff.

Zopolis4:
Progress-- `tools/json.rb` is now fully and properly working, and the only remaining issue is with the CI. I'm using docker because crew doesn't work easily on non-chromeos systems, but `docker run` seems to be having some issues. I'd use the built in docker support, but that doesn't work with upload-artifact when running in the crewbuild docker containers, and we need upload-artifact to output json regularly for Repology.

Zopolis4:
Figured it out! (Note that the current workflow will not run sucessfully when it is not merged, as `json.rb` is not yet in upstream chromebrew.

PR #7398, opened 2022-09-13
mold: modify to fix mold -run issue with mold-wrapper.so location on x86_64

satmandu:
Works around issue reported upstream at https://github.com/rui314/mold/issues/697 .

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=mold_run  CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/mold.rb
+++ b/packages/mold.rb
@@ -6,22 +6,22 @@ require 'package'
 class Mold < Package
   description 'A Modern Linker'
   homepage 'https://github.com/rui314/mold'
-  version '1.4.2-7ffdcd4'
+  version '1.4.2-9fd8313'
   compatibility 'all'
   source_url 'https://github.com/rui314/mold.git'
-  git_hashtag '7ffdcd49d8cc30c34dbadbed1e93b4d15b3d9c15'
+  git_hashtag '9fd831365738b58c4e5d5b3593aedc940941315b'
 
   binary_url({
-    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/mold/1.4.2-7ffdcd4_armv7l/mold-1.4.2-7ffdcd4-chromeos-armv7l.tar.zst',
-     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/mold/1.4.2-7ffdcd4_armv7l/mold-1.4.2-7ffdcd4-chromeos-armv7l.tar.zst',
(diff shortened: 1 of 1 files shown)


packages/mold.rb
```diff
@@ -46,5 +46,10 @@ def self.build
 
   def self.install
     system "DESTDIR=#{CREW_DEST_DIR} samu -C builddir install"
+    # See https://github.com/rui314/mold/issues/697
+    return unless ARCH == 'x86_64'
+
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/lib/mold"
+    FileUtils.ln "#{CREW_DEST_LIB_PREFIX}/mold/mold-wrapper.so", "#{CREW_DEST_PREFIX}/lib/mold/mold-wrapper.so"
```
uberhacker, line 53: `FileUtils.ln "#{CREW_LIB_PREFIX}/mold/mold-wrapper.so", "#{CREW_DEST_PREFIX}/lib/mold/mold-wrapper.so"`?

packages/mold.rb
```diff
@@ -46,5 +46,10 @@ def self.build
 
   def self.install
     system "DESTDIR=#{CREW_DEST_DIR} samu -C builddir install"
+    # See https://github.com/rui314/mold/issues/697
+    return unless ARCH == 'x86_64'
```
uberhacker, line 50: `if ARCH == 'x86_64'` would be clearer.

satmandu: This was a rubocop stylistic suggestion. I switched it from the `if` format.

uberhacker: I guess it doesn't matter.  Just a bit more confusing.

packages/mold.rb
satmandu: No , as this is making a hard link, not a symlink, so you need the actual path to find the file.

Is that bad?  I think our system handles hard links just fine...

supechicken: Will the hard link point to `/usr/local/tmp/crew/dest/...` after installation? If so using `cp` or symlink might be a better choice

uberhacker: What benefit is there to a hardlink vs a symlink?

uberhacker (approved):

PR #7401, opened 2022-09-13
Add binutils dependency to gcc.rb

Zopolis4:
Needed for `as`, which gcc will fail a sanity check without.

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=fix1 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/gcc.rb
+++ b/packages/gcc.rb
@@ -22,6 +22,7 @@ class Gcc < Package
      x86_64: '01efdcdab6bebc32fd5608f6a061c7ceebd69e9f96084293a903f5af9c3576f3'
   })
 
+  depends_on 'binutils'
   depends_on 'ccache' => :build
   depends_on 'dejagnu' => :build # for test
   depends_on 'glibc' => :build


satmandu (approved):

packages/gcc.rb
```diff
@@ -22,6 +22,7 @@ class Gcc < Package
      x86_64: '01efdcdab6bebc32fd5608f6a061c7ceebd69e9f96084293a903f5af9c3576f3'
   })
 
+  depends_on 'binutils'
```
uberhacker, line 25: Does this require `binutils` in core.rb?

Zopolis4: I thought about it, but I decided against it. Whether binutils should be in core.rb independently of this is something else altogether, and I don't have any strong opinions on that either way.

uberhacker: Since this gcc is a core package this dependency will get inherited so I think it should be added.  Also, it would be nice to update `tools/core_packages.txt` too.

PR #7405, opened 2022-09-15
Composer 2.4.0 => 2.4.1

uberhacker:


packages/composer.rb
```diff
@@ -25,7 +25,7 @@ def self.preinstall
 
   def self.install
     downloader "https://github.com/composer/composer/releases/download/#{version}/composer.phar",
```
supechicken, line 27: Use `source_url`?

uberhacker: Does that work with `.phar` files?

supechicken: Yes, any non-archive files should work

uberhacker: Done.  I also abstracted the path from `~/.bashrc`.

satmandu (approved):

PR #7426, opened 2022-09-24
hunspell: automatic selection of language if no input

satmandu:
Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=hunspell  CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/hunspell.rb
+++ b/packages/hunspell.rb
@@ -23,7 +23,7 @@ class Hunspell < Package
         puts '2 = Français'
         puts '3 = Español'
         puts '0 = Cancel'
-        $stdin.gets.chomp
+        $stdin.getc.chomp.to_i
       end
     rescue Timeout::Error
       @user_input = 1
--- a/packages/hunspell.rb
+++ b/packages/hunspell.rb
@@ -24,9 +24,9 @@ class Hunspell < Package
 
     EOT
 
-    if STDIN.isatty # check if stdin is a terminal
+    if $stdin.isatty # check if stdin is a terminal
       @stdin_thread = Thread.new do
-        @user_input = STDIN.getc.to_i
+        @user_input = $stdin.getc.to_i
 
         Thread.kill(@countdown_thread) # stop countdown when user input
       end
--- a/packages/hunspell.rb
+++ b/packages/hunspell.rb
@@ -31,6 +31,6 @@ class Hunspell < Package
     depends_on 'hunspell_en_us' if @user_input == 1
     depends_on 'hunspell_fr_fr' if @user_input == 2
     depends_on 'hunspell_es_any' if @user_input == 3
-    abort if @user_input == 0
+    abort if @user_input.zero?
   end
 end
--- a/packages/hunspell.rb
+++ b/packages/hunspell.rb
@@ -15,7 +15,7 @@ class Hunspell < Package
     puts <<~EOT
 
       Enter your preferred language:
-      (The default, American English, will be selected in #{@timeout} seconds if there is no input.)
+      (The default, American English, will be selected if there is no input.)
 
         1 = American English
         2 = Français
--- a/packages/hunspell.rb
+++ b/packages/hunspell.rb
@@ -14,10 +14,7 @@ class Hunspell < Package
 
     puts <<~EOT
 
-      Enter your preferred language:
-      (The default, American English, will be selected if there is no input.)
-
-        1 = American English
+        1 = American English (default)
         2 = Français
         3 = Español
         0 = Cancel
@@ -26,7 +23,8 @@ class Hunspell < Package
 
     if $stdin.isatty # check if stdin is a terminal
       @stdin_thread = Thread.new do
-        @user_input = $stdin.getc.to_i
+        @user_input = $stdin.getc
+        @user_input = 1 if @user_input == "\n"
 
         Thread.kill(@countdown_thread) # stop countdown when user input
       end
@@ -34,7 +32,7 @@ class Hunspell < Package
       @countdown_thread = Thread.new do
         (0..@timeout).each do |i|
           remaining_time = @timeout - i
-          print "\r#{remaining_time} second(s) left: "
+          print "\rDefault selected in #{remaining_time} second(s).  Enter your preferred language [1 = American English]: "
 
           if remaining_time.zero?
             Thread.kill(@stdin_thread) # stop reading from terminal if timeout
@@ -49,17 +47,15 @@ class Hunspell < Package
     end
 
     case @user_input
-    when 0
+    when '0'
       abort "\nCancelled.".lightred
-    when 1
+    when '1'
       depends_on 'hunspell_en_us'
-    when 2
+    when '2'
       depends_on 'hunspell_fr_fr'
-    when 3
+    when '3'
       depends_on 'hunspell_es_any'
     else
       warn "\nSelected 'American English' by default.".yellow
       depends_on 'hunspell_en_us'
     end
-  end
-end
--- a/packages/hunspell.rb
+++ b/packages/hunspell.rb
@@ -14,10 +14,10 @@ class Hunspell < Package
 
     puts <<~EOT
 
-        1 = American English (default)
-        2 = Français
-        3 = Español
-        0 = Cancel
+      1 = American English (default)
+      2 = Français
+      3 = Español
+      0 = Cancel
 
     EOT
 
@@ -59,3 +59,5 @@ class Hunspell < Package
       warn "\nSelected 'American English' by default.".yellow
       depends_on 'hunspell_en_us'
     end
+  end
+end
--- a/packages/hunspell.rb
+++ b/packages/hunspell.rb
@@ -1,4 +1,5 @@
 require 'package'
+require 'timeout'
 
 class Hunspell < Package
   description 'Hunspell is a spell checker and morphological analyzer library'
@@ -12,27 +13,24 @@ class Hunspell < Package
   is_fake
 
   if ARGV[0] == 'install'
-    puts
-    puts 'Enter your preferred language:'
-    puts '1 = American English'
-    puts '2 = Français'
-    puts '3 = Español'
-    puts '0 = Cancel'
-
-    while version = $stdin.gets.chomp
-      case version
-      when '1'
-        depends_on 'hunspell_en_us'
-        break
-      when '2'
-        depends_on 'hunspell_fr_fr'
-        break
-      when '3'
-        depends_on 'hunspell_es_any'
-        break
-      when '0'
-        abort
+    begin
+      @timeout = 10
+      @user_input = Timeout.timeout(@timeout) do
+        puts
+        puts 'Enter your preferred language:'
+        puts "(The default, American English, will be selected in #{@timeout} seconds if there is no input.)"
+        puts '1 = American English'
+        puts '2 = Français'
+        puts '3 = Español'
+        puts '0 = Cancel'
+        $stdin.gets.chomp
       end
+    rescue Timeout::Error
+      @user_input = 1
     end
+    depends_on 'hunspell_en_us' if @user_input == 1
+    depends_on 'hunspell_fr_fr' if @user_input == 2
+    depends_on 'hunspell_es_any' if @user_input == 3
+    abort if @user_input == 0
   end
 end
--- a/packages/hunspell.rb
+++ b/packages/hunspell.rb
@@ -1,36 +1,65 @@
 require 'package'
-require 'timeout'
 
 class Hunspell < Package
   description 'Hunspell is a spell checker and morphological analyzer library'
-  homepage 'http://hunspell.github.io/'
+  homepage 'https://hunspell.github.io'
   version '1.7.0'
   license 'MPL-1.1, GPL-2 and LGPL-2.1'
   compatibility 'all'
-  source_url 'https://github.com/hunspell/hunspell/archive/v1.7.0.tar.gz'
-  source_sha256 'bb27b86eb910a8285407cf3ca33b62643a02798cf2eef468c0a74f6c3ee6bc8a'
 
   is_fake
 
-  if ARGV[0] == 'install'
-    begin
-      @timeout = 10
-      @user_input = Timeout.timeout(@timeout) do
-        puts
-        puts 'Enter your preferred language:'
-        puts "(The default, American English, will be selected in #{@timeout} seconds if there is no input.)"
-        puts '1 = American English'
-        puts '2 = Français'
-        puts '3 = Español'
-        puts '0 = Cancel'
-        $stdin.getc.chomp.to_i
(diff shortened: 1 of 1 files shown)


packages/hunspell.rb
```diff
@@ -12,27 +13,24 @@ class Hunspell < Package
...
-      when '3'
-        depends_on 'hunspell_es_any'
-        break
-      when '0'
-        abort
+    begin
+      @timeout = 10
+      @user_input = Timeout.timeout(@timeout) do
+        puts
+        puts 'Enter your preferred language:'
```
supechicken, line 20: Use heredoc?

packages/hunspell.rb
```diff
@@ -12,27 +13,24 @@ class Hunspell < Package
...
+        puts '2 = Français'
+        puts '3 = Español'
+        puts '0 = Cancel'
+        $stdin.gets.chomp
       end
+    rescue Timeout::Error
+      @user_input = 1
     end
+    depends_on 'hunspell_en_us' if @user_input == 1
+    depends_on 'hunspell_fr_fr' if @user_input == 2
```
supechicken, line 32: Better to use `case`?

packages/hunspell.rb
satmandu: That makes it longer unnecessarily?

packages/hunspell.rb
satmandu: case causes issues with `depends_on`.

packages/hunspell.rb
```diff
@@ -12,27 +13,24 @@ class Hunspell < Package
...
+        puts
+        puts 'Enter your preferred language:'
+        puts "(The default, American English, will be selected in #{@timeout} seconds if there is no input.)"
+        puts '1 = American English'
+        puts '2 = Français'
+        puts '3 = Español'
+        puts '0 = Cancel'
+        $stdin.gets.chomp
       end
+    rescue Timeout::Error
```
supechicken, line 28: How about combining the `do...end` block and `rescue` block?
```ruby
@user_input = Timeout.timeout(10) do
  ...
rescue
  ...
end
```

packages/hunspell.rb
```diff
@@ -12,27 +13,24 @@ class Hunspell < Package
...
-        depends_on 'hunspell_es_any'
-        break
-      when '0'
-        abort
+    begin
+      @timeout = 10
+      @user_input = Timeout.timeout(@timeout) do
+        puts
+        puts 'Enter your preferred language:'
+        puts "(The default, American English, will be selected in #{@timeout} seconds if there is no input.)"
```
uberhacker, line 21: Can `@timeout` display a countdown in real time?  If that can be done then we don't need `if there is no input`?

packages/hunspell.rb
supechicken: It is worth to add two more line for a more clear look :)

packages/hunspell.rb
supechicken: That sounds weird...

packages/hunspell.rb
```diff
@@ -12,27 +13,24 @@ class Hunspell < Package
...
+        puts '1 = American English'
+        puts '2 = Français'
+        puts '3 = Español'
+        puts '0 = Cancel'
+        $stdin.gets.chomp
```
supechicken, line 26: Do you mean `$stdin.gets.chomp.to_i` (otherwise the variable will be `"1"`)

Also `getc`would be more suitable for single word input

packages/hunspell.rb
supechicken: No, as we can't get the current countdown from `Timeout`

satmandu: Alternate code to do that would be welcomed!

packages/hunspell.rb
satmandu: Can you verify if that works?

packages/hunspell.rb
satmandu: I think the issue is accepting input while the countdown timer is running.

packages/hunspell.rb
satmandu: done

packages/hunspell.rb
satmandu: Please just suggest replacement code to do that. :)

packages/hunspell.rb
uberhacker: We used case initially and it worked fine.

uberhacker: I'll make it easy for you.
```
      case @user_input
      when '1'
        depends_on 'hunspell_en_us'
        break
      when '2'
        depends_on 'hunspell_fr_fr'
        break
      when '3'
        depends_on 'hunspell_es_any'
        break
      when '0'
        abort
```

packages/hunspell.rb
uberhacker: What happens if an incorrect value is entered?

packages/hunspell.rb
satmandu: More specifically, it was breaking inside the begin statement while I was testing that.

You're right, I should test this now that logic is outside that again.

satmandu: Here is what I see @uberhacker:
```
  if ARGV[0] == 'install'
    begin
      @timeout = 10
      @user_input = Timeout.timeout(@timeout) do
        puts
        puts 'Enter your preferred language:'
        puts "(The default, American English, will be selected in #{@timeout} seconds if there is no input.)"
        puts '1 = American English'
        puts '2 = Français'
        puts '3 = Español'
        puts '0 = Cancel'
        $stdin.getc.chomp.to_i
      end
    rescue Timeout::Error
      @user_input = 1
    end
    case @user_input
    when '1'
      depends_on 'hunspell_en_us'
      break
    when '2'
      depends_on 'hunspell_fr_fr'
      break
    when '3'
      depends_on 'hunspell_es_any'
      break
    when '0'
      abort
    end
  end
end
```

```
SyntaxError: /usr/local/lib/crew/packages/hunspell.rb:34: Can't escape from eval with break
/usr/local/bin/crew:171:in `block in print_current_package': undefined method `name' for nil:NilClass (NoMethodError)

  status = 'installed' if @device[:installed_packages].any? { |elem| elem[:name] == @pkg.name }
                                                                                        ^^^^^
        from /usr/local/bin/crew:171:in `any?'
        from /usr/local/bin/crew:171:in `print_current_package'
        from /usr/local/bin/crew:1752:in `block in install_command'
        from /usr/local/bin/crew:1749:in `each'
        from /usr/local/bin/crew:1749:in `install_command'
```

satmandu: (That's the same issue I was seeing before, which I why I moved away from using `case`.

uberhacker: Maybe remove the quotes on the numbers since they are integers?

satmandu: That doesn't help. The issue is with the `break`s in there.
```
SyntaxError: /usr/local/lib/crew/packages/hunspell.rb:34: Can't escape from eval with break
```

supechicken:
Here is all my suggestions, with countdown in real-time (tested)

Remember to remove comments start with triple `#`
```ruby
require 'package'

class Hunspell < Package
  description 'Hunspell is a spell checker and morphological analyzer library'
  homepage 'https://hunspell.github.io' ### use https
  version '1.x'
  license 'MPL-1.1, GPL-2 and LGPL-2.1'
  compatibility 'all'
  ### source is not required for fake packages

  is_fake

  if ARGV.include?('install') ### fix when 'install' isn't in ARGV[0] (like `crew -k install`)
    @timeout = 10

    puts <<~EOT

      Enter your preferred language:
      (The default, American English, will be selected in #{@timeout} seconds if there is no input.)

        1 = American English
        2 = Français
        3 = Español
        0 = Cancel

    EOT

    if STDIN.isatty ### check if stdin is a terminal
      ### reworked countdown logic
      @stdin_thread = Thread.new do
        @user_input = STDIN.getc.to_i

        Thread.kill(@countdown_thread) # stop countdown when user input
      end

      @countdown_thread = Thread.new do
        (0..@timeout).each do |i|
          remaining_time = @timeout - i
          print "\r#{remaining_time} second(s) left: "

          if remaining_time.zero?
            Thread.kill(@stdin_thread) # stop reading from terminal if timeout
          else
            sleep 1
          end
        end
      end

      @stdin_thread.join
      @countdown_thread.join
    end

    unless @user_input&.between(0, 3)
      warn "\nSelected 'American English' by default.".yellow
      @user_input = 1
    end

    ### this is not javascript, the `break` statement is not necessary in `case` at all...
    case @user_input
    when 1
      depends_on 'hunspell_en_us'
    when 2
      depends_on 'hunspell_fr_fr'
    when 3
      depends_on 'hunspell_es_any'
    else
      abort "\nCancelled.".lightred
    end
  end
end
```

packages/hunspell.rb
satmandu: With @supechicken's changes, an incorrect value uses the default.

satmandu:
@supechicken ` if STDIN.isatty` is false, do we need to set `@user_input`?

packages/hunspell.rb
```diff
@@ -2,37 +2,64 @@
...
-        break
-      when '2'
-        depends_on 'hunspell_fr_fr'
-        break
-      when '3'
-        depends_on 'hunspell_es_any'
-        break
-      when '0'
-        abort
+  if ARGV.include?('install') # fix when 'install' isn't in ARGV[0] (like `crew -k install`)
```
supechicken, line 12: This comment can be removed :)

supechicken:
> @supechicken if `STDIN.isatty` is false, do we need to set `@user_input`?

No, `@user_input` will be `nil` if `STDIN.isatty` is false and will fallback to the `else` part in the `case` statement

packages/hunspell.rb
satmandu: It is a good comment to leave in.

supechicken (approved):

packages/hunspell.rb
```diff
@@ -2,37 +2,64 @@
...
+      @stdin_thread = Thread.new do
+        @user_input = $stdin.getc.to_i
+
+        Thread.kill(@countdown_thread) # stop countdown when user input
       end
+
+      @countdown_thread = Thread.new do
+        (0..@timeout).each do |i|
+          remaining_time = @timeout - i
+          print "\r#{remaining_time} second(s) left: "
```
uberhacker, line 37: Why `\r` and not `\n`?  Remove `:`?

packages/hunspell.rb
```diff
@@ -2,37 +2,64 @@
...
-        break
-      when '0'
-        abort
+  if ARGV.include?('install') # fix when 'install' isn't in ARGV[0] (like `crew -k install`)
+    @timeout = 10
+
+    puts <<~EOT
+
+      Enter your preferred language:
+      (The default, American English, will be selected in #{@timeout} seconds if there is no input.)
```
uberhacker, line 18: No need for `#{@timeout}` with `#{remaining_time}`?

packages/hunspell.rb
satmandu: @supechicken?

packages/hunspell.rb
satmandu: @supechicken?

packages/hunspell.rb
supechicken: The `\r` char represents to "reprint the current line" (see `downloader.rb` for more info), and `:` is for prompting user to input

packages/hunspell.rb
supechicken: Countdown is in line 37...

satmandu: Adjusted the message.

uberhacker:
To avoid confusion for the user, how about?
```
  if ARGV.include?('install') # fix when 'install' isn't in ARGV[0] (like `crew -k install`)
    @timeout = 10

    puts <<~EOT

        1 = American English (default)
        2 = Français
        3 = Español
        0 = Cancel

    EOT

    if $stdin.isatty # check if stdin is a terminal
      @stdin_thread = Thread.new do
        @user_input = $stdin.getc
        @user_input = 1 if @user_input == "\n"

        Thread.kill(@countdown_thread) # stop countdown when user input
      end

      @countdown_thread = Thread.new do
        (0..@timeout).each do |i|
          remaining_time = @timeout - i
          print "\rDefault selected in #{remaining_time} second(s).  Enter your preferred language [1 = American English]: "

          if remaining_time.zero?
            Thread.kill(@stdin_thread) # stop reading from terminal if timeout
          else
            sleep 1
          end
        end
      end

      @stdin_thread.join
      @countdown_thread.join
    end

    case @user_input
    when '0'
      abort "\nCancelled.".lightred
    when '1'
      depends_on 'hunspell_en_us'
    when '2'
      depends_on 'hunspell_fr_fr'
    when '3'
      depends_on 'hunspell_es_any'
    else
      warn "\nSelected 'American English' by default.".yellow
      depends_on 'hunspell_en_us'
    end
```

supechicken:
How about including the whole countdown+select logic as a library in `crew`? So that other fake packages like `jdk` can implement this also without duplicated codes

packages/hunspell.rb
```diff
@@ -2,37 +2,62 @@
...
+      2 = Français
+      3 = Español
+      0 = Cancel
+
+    EOT
+
+    if $stdin.isatty # check if stdin is a terminal
+      @stdin_thread = Thread.new do
+        @user_input = $stdin.getc
+        @user_input = 1 if @user_input == "\n"
```
supechicken, line 27: This can be moved to the `case` statement:
```ruby
case ...
when '1', "\n"
  ...
```

supechicken:
> I think that would be a good idea...

Working on this now :)

satmandu:
> How about including the whole countdown+select logic as a library in `crew`? So that other fake packages like `jdk` can implement this also without duplicated codes

Do you want to work on making this a library for crew, and then we can implement this using that?

supechicken:
> > How about including the whole countdown+select logic as a library in `crew`? So that other fake packages like `jdk` can implement this also without duplicated codes
> 
> Do you want to work on making this a library for crew, and then we can implement this using that?

Yup

packages/hunspell.rb
uberhacker: Good call!

uberhacker: @satmandu: What do you think about @supechicken's suggestion?

satmandu: I think that's fine... but I thought we were just waiting for @supechicken to get this working as a module in crew...

uberhacker: Fair enough.

supechicken:
See #7473

PR #7428, opened 2022-09-25
Firefox 105.0.1 fix checksums

MadBrad:
Firefox will not install due to incorrect checksums. I edited the sha256 checksums to be correct according to http://download-origin.cdn.mozilla.net/pub/firefox/releases/105.0.1/SHA256SUMS


satmandu:
Thanks for the fix!

satmandu (approved):

PR #7430, opened 2022-09-27
Update package template

supechicken:
### Changes
- Merge with package template in [chromebrew/chromebrew.github.io](https://github.com/supechicken/chromebrew.github.io/blob/main/index.html)
- Add `#!/usr/bin/env ruby` to enable syntax highlighting

### DIFF
--- a/packages/template.rb.template
+++ b/packages/template.rb.template
@@ -1,5 +1,6 @@
-# It is recommended that you investigate other files in the package directory
-# to help get an idea of how existing packages are being installed through chromebrew
+#!/usr/bin/env ruby
+# It is recommended that you investigate other files in the 'packages/' directory
+# to help get an idea of how existing packages are being installed through Chromebrew.
 
 require 'package'
 
@@ -7,23 +8,24 @@ class Template < Package
   description ''
   homepage ''
   version ''
-  license ''
-  compatibility ''
-  source_url '*.tar.gz'
+  license '' # Package license
+  compatibility '' # Package compatibility
+  source_url '*.tar.{gz,bz2,lz,etc}'
   source_sha256 ''
 
-  binary_url ({
-    aarch64: 'https://dl.bintray.com/chromebrew/chromebrew/*-chromeos-armv7l.tar.xz',
-     armv7l: 'https://dl.bintray.com/chromebrew/chromebrew/*-chromeos-armv7l.tar.xz',
-       i686: 'https://dl.bintray.com/chromebrew/chromebrew/*-chromeos-i686.tar.xz',
-     x86_64: 'https://dl.bintray.com/chromebrew/chromebrew/*-chromeos-x86_64.tar.xz',
-  })
-  binary_sha256 ({
-    aarch64: '',
-     armv7l: '',
-       i686: '',
-     x86_64: '',
-  })
+  # Optional: Prebuilt binary
+  #  binary_url ({
+  #    aarch64: '',
+  #     armv7l: '',
+  #       i686: '',
+  #     x86_64: '',
+  #  })
+  #  binary_sha256 ({
+  #    aarch64: '',
+  #     armv7l: '',
+  #       i686: '',
+  #     x86_64: '',
+  #  })
 
 
   # Register dependencies (use the following line as a basis)
@@ -31,8 +33,13 @@ class Template < Package
   # depends_on '*'
   #
 
-  # Function to perform pre-flight operations prior to dependency checks.
+  # Function for checks to see if install should occur.
   def self.preflight
+
+  end
+
+  # Function to perform pre-install operations prior to install for both source build and binary distribution.
+  def self.preinstall
 
   end
 
@@ -42,31 +49,20 @@ class Template < Package
 
   end
 
-
   # Function to perform pre-build operations prior to build from source.
   def self.prebuild
 
   end
-
 
   # Function to perform build from source.
   def self.build
 
   end
 
-
-  # Function to perform check from source build.
-  # This executes only during `crew build`.
-  def self.check
-
-  end
-
-
   # Function to perform pre-install operations prior to install.
   def self.preinstall
 
   end
-
 
   # Function to perform install from source build.
   def self.install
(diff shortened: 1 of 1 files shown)


satmandu:
Do we want to have `#!/usr/bin/env ruby` in every package file? Those files are set to 0644 on purpose so they aren't considered executable.

@uberhacker ?

uberhacker:
> Do we want to have `#!/usr/bin/env ruby` in every package file? Those files are set to 0644 on purpose so they aren't considered executable.
> 
> @uberhacker ?

I generally prefer `#!/usr/local/bin/ruby` but that might mess up GitHub language detection.

packages/template.rb.template
```diff
@@ -1,73 +1,69 @@
...
   #
 
-  # Function to perform pre-flight operations prior to dependency checks.
+  # Function for checks to see if install should occur.
   def self.preflight
 
   end
 
+  # Function to perform pre-install operations prior to install for both source build and binary distribution.
+  def self.preinstall
```
uberhacker, line 42: Already on line 66?

packages/template.rb.template
```diff
@@ -80,4 +76,9 @@ class Template < Package
   end
 
 
-end
+  # Function to perform check from source build.
+  # This is execute if and only if `crew build`.
+  def self.check
```
uberhacker, line 81: I arranged these in chronological sequence.  I like to have the sequence in packages too but so many packagers like to change it around. :)

uberhacker (changes requested):

packages/template.rb.template
```diff
@@ -1,73 +1,69 @@
...
 
 class Template < Package
   description ''
   homepage ''
   version ''
-  license ''
-  compatibility ''
-  source_url '*.tar.gz'
+  license '' # Package license
+  compatibility '' # Package compatibility
```
uberhacker, line 12: Maybe show some examples too?
```
license '' # Package license (eg. GPL-3, MIT, etc.)
compatibility '' # Package compatibility (eg. aarch64, armv7l, i686, x86_64 or all)
```

packages/template.rb.template
```diff
@@ -1,73 +1,69 @@
...
+  source_url '*.tar.{gz,bz2,lz,etc}'
```
uberhacker, line 13: How about?
```
source_url '' # The url to the source code (eg. https://ftpmirror.gnu.org/sed/sed-4.8.tar.xz)
```

uberhacker (changes requested):

uberhacker:
> > Do we want to have `#!/usr/bin/env ruby` in every package file? Those files are set to 0644 on purpose so they aren't considered executable.
> > @uberhacker ?
> 
> I generally prefer `#!/usr/local/bin/ruby` but that might mess up GitHub language detection.

Thinking outside the box.  Why do we care about language detection?  Maybe it would be less confusing if we don't add the line?

supechicken:
> Thinking outside the box. Why do we care about language detection? Maybe it would be less confusing if we don't add the line?

Because I prefer syntax highlighting :)

uberhacker:
> > Thinking outside the box. Why do we care about language detection? Maybe it would be less confusing if we don't add the line?
> 
> Because I prefer syntax highlighting :)

Isn't that controlled in your editor?

packages/template.rb.template
supechicken: Added

packages/template.rb.template
supechicken: Reverted to the original sequence

supechicken:
> Can you mention in the comment that the `#!/usr/bin/env ruby` is only for the template file, and that it does not go into a normal package file?

Now the shebang line looks like this:
```ruby
#!/usr/bin/env ruby <- This line is not necessary, just for GitHub's language detection
```

packages/template.rb.template
```diff
@@ -1,73 +1,69 @@
-# It is recommended that you investigate other files in the package directory
-# to help get an idea of how existing packages are being installed through chromebrew
+#!/usr/bin/env ruby <- This line is not necessary, just for GitHub's language detection
```
uberhacker, line 1: Do we need another `#` before `<-` to indicate the rest of the line is a comment?

packages/template.rb.template
```diff
@@ -1,73 +1,69 @@
...
-
   # Function to perform build from source.
   def self.build
 
   end
 
-
   # Function to perform check from source build.
-  # This executes only during `crew build`.
+  # This is execute if and only if `crew build`.
```
uberhacker, line 57: `This is executed ...`?  Proper verb usage. :)

packages/template.rb.template
supechicken: The shebang line won't work anyways as `env` does not support comments

packages/template.rb.template
```diff
@@ -1,73 +1,69 @@
...
+  #  })
 
 
   # Register dependencies (use the following line as a basis)
   #
   # depends_on '*'
   #
 
-  # Function to perform pre-flight operations prior to dependency checks.
+  # Function for checks to see if install should occur.
```
uberhacker, line 36: How about?
```
# Function to check if install should occur.
```

packages/template.rb.template
supechicken: Change made

packages/template.rb.template
```diff
@@ -1,73 +1,69 @@
...
-
   # Function to perform build from source.
   def self.build
 
   end
 
-
   # Function to perform check from source build.
-  # This executes only during `crew build`.
+  # This is executed if and only if `crew build`.
```
uberhacker, line 57: How about?
```
# This is executed only during `crew build`.
```

packages/template.rb.template
supechicken: Change made

supechicken:
Please also take a look at https://github.com/chromebrew/chromebrew.github.io/pull/2 :)

uberhacker (approved):

PR #7436, opened 2022-09-28
Git w/o musl, install.sh modified to work w/o git in bootstrap.

satmandu:
- Dependencies have been added to core packages to make sure the order at installation doesn't cause problems.
 - Also includes the glibc fix from #7449 in case that was causing your problem.
- waiting for @uberhacker to test the installer with: 
```
cd /usr/local
curl -OLf https://github.com/satmandu/chromebrew/raw/git_no_musl/install.sh
chmod +x install.sh
OWNER=satmandu BRANCH=git_no_musl ./install.sh
```
 
Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=git_no_musl  CREW_TESTING=1 crew update
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -24,7 +24,11 @@ CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 ARCH="${ARCH/armv8l/armv7l}"
 
 # BOOTSTRAP_PACKAGES cannot depend on crew_profile_base for their core operations (completion scripts are fine)
-BOOTSTRAP_PACKAGES="zstd pixz ca_certificates git gmp ncurses xxhash lz4 popt libyaml openssl gcc rsync ruby c_ares libnghttp2 libidn2 libssh libpsl openldap brotli zlibpkg krb5 e2fsprogs libunistring pcre2 libcyrussasl libcurl glibc"
+# rsync requires openssl xxhash lz4 popt
+BOOTSTRAP_PACKAGES="zstd pixz ca_certificates ruby rsync openssl xxhash lz4 popt"
+# i686 requires gcc and openssl
+[ "${ARCH}" == "i686" ] && BOOTSTRAP_PACKAGES+=" gcc"
+
 RED='\e[1;91m';    # Use Light Red for errors.
 YELLOW='\e[1;33m'; # Use Yellow for informational messages.
 GREEN='\e[1;32m';  # Use Green for success messages.
@@ -276,6 +280,25 @@ echo_info "Setup and synchronize local package repo..."
 echo -e "${GRAY}"
 
 # Remove old git config directories if they exist
+rm -rf "${CREW_LIB_PATH}" && mkdir "${CREW_LIB_PATH}"
+
+cd "${CREW_LIB_PATH}" && curl -L https://github.com/"${OWNER}"/"${REPO}"/tarball/"${BRANCH}" | tar -xz --strip-components=1
+
+# Set LD_LIBRARY_PATH so crew doesn't break on i686 and
+# the mandb install doesn't fail
+export LD_LIBRARY_PATH="${CREW_PREFIX}/lib${LIB_SUFFIX}"
+
+# Since we just downloaded the package repo, just update package compatibility information.
+crew update compatible
+
+echo_info "Installing core Chromebrew packages...\n"
+yes | crew install core
+
+echo_info "\nRunning Bootstrap package postinstall scripts...\n"
+crew postinstall $BOOTSTRAP_PACKAGES
+
+echo_info "Synchronizng local package repo..."
+cd "${CREW_LIB_PATH}"/..
 rm -rf "${CREW_LIB_PATH}"
 
 # Do a minimal clone, which also sets origin to the master/main branch
@@ -295,20 +318,6 @@ git checkout "${BRANCH}"
 git sparse-checkout set packages lib bin crew tools
 git reset --hard origin/"${BRANCH}"
 echo -e "${RESET}"
-
-echo_info "Updating crew package information...\n"
-# set LD_LIBRARY_PATH so crew doesn't break on i686 and the mandb 
-# install doesn't fail
-export LD_LIBRARY_PATH="${CREW_PREFIX}/lib${LIB_SUFFIX}"
-
-# Since we just ran git, just update package compatibility information.
-crew update compatible
-
-echo_info "Installing core Chromebrew packages...\n"
-yes | crew install core
-
-echo_info "\nRunning Bootstrap package postinstall scripts...\n"
-crew postinstall $BOOTSTRAP_PACKAGES
 
 echo "                       . .
                    ..,:;;;::'..
--- a/install.sh
+++ b/install.sh
@@ -215,11 +215,11 @@ function extract_install () {
     echo_intra "Extracting ${1} ..."
     if [[ "$2" == *".zst" ]];then
       if [[ -e /usr/bin/zstd ]]; then
-        PATH=/usr/bin:$PATH tar -Izstd -xpf ../"${2}"
+        tar -I /usr/bin/zstd -xpf ../"${2}"
       elif (LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} "${CREW_PREFIX}"/bin/zstd --version &> /dev/null); then
-        LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} tar -Izstd -xpf ../"${2}"
+        LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} tar -I "${CREW_PREFIX}"/bin/zstd -xpf ../"${2}"
       else
-        echo "Zstd is broken or nonfunctional, and packages can not be extracted properly."
+        echo_error "Zstd is broken or nonfunctional, and packages can not be extracted properly."
         exit 1
       fi
     elif [[ "$2" == *".tpxz" ]];then
@@ -298,7 +298,7 @@ echo -e "${RESET}"
 echo_info "Updating crew package information...\n"
 # Without setting LD_LIBRARY_PATH, the mandb postinstall fails
 # from not being able to find the gdbm library.
-export LD_LIBRARY_PATH=$(LD_LIBRARY_PATH=/usr/local/lib$LIB_SUFFIX:/usr/lib$LIB_SUFFIX:/lib$LIB_SUFFIX crew const CREW_LIB_PREFIX | sed -e 's:CREW_LIB_PREFIX=::g')
+export LD_LIBRARY_PATH=$(LD_LIBRARY_PATH=/usr/local/lib${LIB_SUFFIX}:/usr/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} crew const CREW_LIB_PREFIX | sed -e 's:CREW_LIB_PREFIX=::g')
 # Since we just ran git, just update package compatibility information.
 crew update compatible
 
--- a/install.sh
+++ b/install.sh
@@ -37,24 +37,24 @@ GRAY='\e[0;37m';   # Use Gray for program output.
 MAGENTA='\e[1;35m';
 RESET='\e[0m'
 
-# simplify colors and print errors to stderr (2)
+# Simplify colors and print errors to stderr (2).
 echo_error() { echo -e "${RED}${*}${RESET}" >&2; }
 echo_info() { echo -e "${YELLOW}${*}${RESET}" >&1; }
 echo_success() { echo -e "${GREEN}${*}${RESET}" >&1; }
 echo_intra() { echo -e "${BLUE}${*}${RESET}" >&1; }
 echo_out() { echo -e "${GRAY}${*}${RESET}" >&1; }
 
-# skip all checks if running on a docker container
+# Skip all checks if running on a docker container.
 [[ -f "/.dockerenv" ]] && CREW_FORCE_INSTALL=1
 
-# reject crostini
+# Reject crostini.
 if [[ -d /opt/google/cros-containers && "${CREW_FORCE_INSTALL}" != '1' ]]; then
   echo_error "Crostini containers are not supported by Chromebrew :/"
   echo_info "Run 'curl -Ls git.io/vddgY | CREW_FORCE_INSTALL=1 bash' to perform install anyway"
   exit 1
 fi
 
-# disallow non-stable channels Chrome OS
+# Disallow non-stable channels Chrome OS.
 if [ -f /etc/lsb-release ]; then
   if [[ ! "$(< /etc/lsb-release)" =~ CHROMEOS_RELEASE_TRACK=stable-channel$'\n' && "${CREW_FORCE_INSTALL}" != '1' ]]; then
     echo_error "The beta, dev, and canary channel are unsupported by Chromebrew"
@@ -72,26 +72,27 @@ fi
 
 echo_success "Welcome to Chromebrew!"
 
-# prompt user to enter the sudo password if it set
-# if the PASSWD_FILE specified by chromeos-setdevpasswd exist, that means a sudo password is set
+# Prompt user to enter the sudo password if it set.
+# If the PASSWD_FILE specified by chromeos-setdevpasswd exist, that means a sudo password is set.
 if [[ "$(< /usr/sbin/chromeos-setdevpasswd)" =~ PASSWD_FILE=\'([^\']+) ]] && [ -f "${BASH_REMATCH[1]}" ]; then
   echo_intra "Please enter the developer mode password"
-  # reset sudo timeout
+  # Reset sudo timeout.
   sudo -k
   sudo /bin/true
 fi
 
-# force curl to use system libraries
+# Force curl to use system libraries.
 function curl () {
-  # retry if download failed
-  # the --retry/--retry-all-errors parameter in curl will not work with the 'curl: (7) Couldn't connect to server'
-  # error, a for loop is used here
+  # Retry if download failed.
+  # The --retry/--retry-all-errors parameter in curl will not work with 
+  # the 'curl: (7) Couldn't connect to server' error, a for loop is used
+  # here.
   for (( i = 0; i < 4; i++ )); do
     env LD_LIBRARY_PATH='' ${CURL} --ssl -C - "${@}" && \
       return 0 || \
       echo_info "Retrying, $((3-$i)) retries left."
   done
-  # the download failed if we're still here
+  # The download failed if we're still here.
   echo_error "Download failed :/ Please check your network settings."
   return 1
 }
@@ -109,7 +110,7 @@ esac
 echo_info "\n\nDoing initial setup for install in ${CREW_PREFIX}."
 echo_info "This may take a while if there are preexisting files in ${CREW_PREFIX}...\n"
 
-# This will allow things to work without sudo
+# This will allow things to work without sudo.
 crew_folders="bin cache doc docbook etc include lib lib$LIB_SUFFIX libexec man sbin share tmp var"
 for folder in $crew_folders
 do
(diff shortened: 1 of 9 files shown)


install.sh
```diff
@@ -218,7 +214,14 @@ function extract_install () {
     #extract and install
     echo_intra "Extracting ${1} ..."
     if [[ "$2" == *".zst" ]];then
-      LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} tar -Izstd -xpf ../"${2}"
+      if [[ -e /usr/bin/zstd ]]; then
+        PATH=/usr/bin:$PATH tar -Izstd -xpf ../"${2}"
+      elif (LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} "${CREW_PREFIX}"/bin/zstd --version &> /dev/null); then
+        LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} tar -Izstd -xpf ../"${2}"
+      else
+        echo "Zstd is broken or nonfunctional, and packages can not be extracted properly."
```
supechicken, line 222: Use `echo_error`?

install.sh
```diff
@@ -218,7 +214,14 @@ function extract_install () {
     #extract and install
     echo_intra "Extracting ${1} ..."
     if [[ "$2" == *".zst" ]];then
-      LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} tar -Izstd -xpf ../"${2}"
+      if [[ -e /usr/bin/zstd ]]; then
+        PATH=/usr/bin:$PATH tar -Izstd -xpf ../"${2}"
```
supechicken, line 218: Will `tar -I/usr/bin/zstd -xpf ../"${2}"` better?

install.sh
```diff
@@ -218,7 +214,14 @@ function extract_install () {
...
+      elif (LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} "${CREW_PREFIX}"/bin/zstd --version &> /dev/null); then
+        LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} tar -Izstd -xpf ../"${2}"
+      else
+        echo "Zstd is broken or nonfunctional, and packages can not be extracted properly."
+        exit 1
+      fi
     elif [[ "$2" == *".tpxz" ]];then
       if ! LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} pixz -h &> /dev/null; then
```
supechicken, line 226: Do we have any core package that is still compressed with `.tpxz`?

packages/libssh.rb
```diff
@@ -3,52 +3,50 @@
...
       FileUtils.cp 'src/libssh.a', '../' if File.exist?('src/libssh.a')
-      system "cmake #{CREW_CMAKE_OPTIONS} \
+      system "mold -run cmake #{CREW_CMAKE_OPTIONS} \
       -DWITH_EXAMPLES=OFF \
       -DBUILD_SHARED_LIBS=ON \
-      -DWITH_STATIC_LIB=OFF \
       ../ -G Ninja"
     end
-    system 'ninja -C builddir'
+    system 'mold -run samu -C builddir'
```
supechicken, line 45: Will it make the `CREW_LINKER` constant meaningless if we use `mold -run` in every package?

install.sh
```diff
@@ -298,7 +298,7 @@ echo -e "${RESET}"
 echo_info "Updating crew package information...\n"
 # Without setting LD_LIBRARY_PATH, the mandb postinstall fails
 # from not being able to find the gdbm library.
-export LD_LIBRARY_PATH=$(crew const CREW_LIB_PREFIX | sed -e 's:CREW_LIB_PREFIX=::g')
+export LD_LIBRARY_PATH=$(LD_LIBRARY_PATH=/usr/local/lib$LIB_SUFFIX:/usr/lib$LIB_SUFFIX:/lib$LIB_SUFFIX crew const CREW_LIB_PREFIX | sed -e 's:CREW_LIB_PREFIX=::g')
```
uberhacker, line 301: I don't think this line should be here.  The default `LD_LIBRARY_PATH` will be either `/usr/local/lib` or `/usr/local/lib64` depending on the architecture.  If something doesn't work then we need to move to the correct default directory.  This is just curing a symptom of the underlying issue.

install.sh
uberhacker: We shouldn't be setting `LD_LIBRARY_PATH` in the first place.

install.sh
satmandu: This is only for i686, where the default LD_LIBRARY_PATH doesn't include /usr/local/lib yet when we first run crew, and the wrong libgcc gets used, which kills ruby, since ruby expects a newer libgcc.

packages/libssh.rb
satmandu: Yes.

install.sh
satmandu: I don't think so, but in case we add one, safer to leave the code in.

install.sh
satmandu: It's a very narrow situation where that is happening.

install.sh
satmandu: done.

packages/libssh.rb
satmandu: Yes, but this also lets us avoid hacking up everything with `-fuse-ld=mold`. It would be nice for CREW_LINKER to automatically add `mold -run` to the command line...

install.sh
satmandu: If not i686, this basically does nothing, since ruby doesn't complain about the older libgcc.

install.sh
```diff
@@ -295,7 +298,7 @@ echo -e "${RESET}"
 echo_info "Updating crew package information...\n"
 # Without setting LD_LIBRARY_PATH, the mandb postinstall fails
 # from not being able to find the gdbm library.
-export LD_LIBRARY_PATH=$(crew const CREW_LIB_PREFIX | sed -e 's:CREW_LIB_PREFIX=::g')
+export LD_LIBRARY_PATH=$(LD_LIBRARY_PATH=/usr/local/lib${LIB_SUFFIX}:/usr/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} crew const CREW_LIB_PREFIX | sed -e 's:CREW_LIB_PREFIX=::g')
```
uberhacker, line 301: Can you give me a convincing reason why we need this line?

uberhacker (changes requested):

install.sh
satmandu: On i686, at least in the container, install.sh starts with LD_LIBRARY_PATH to /usr/lib:/lib.
Invoking crew in this line invokes ruby. Ruby depends upon the newer libgcc which is in /usr/local/lib. 
However, owing to the LD_LIBRARY_PATH, if first looks in /usr/lib, where it finds an older libgcc. 
This causes crew to fail running `crew const CEW_LIB_PREFIX`, which in turn causes the install script to fail.
Merely having the LD_IBRARY_PATH fixed for this one line avoids this problem.

satmandu: On newer systems, the system libgcc doesn't cause an issue here.

uberhacker: It will still find the newer libgcc first since `/usr/local/lib` is the higher priority path.  I guess I still don't understand why we need the extra `:/usr/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX}` added to the path.

satmandu: I thought it would too... but it doesn't. My build logs show a libgcc error.

Zopolis4: Why do we need anything except `export LD_LIBRARY_PATH=$(LD_LIBRARY_PATH=/usr/local/lib${LIB_SUFFIX})`?

satmandu: > Why do we need anything except `export LD_LIBRARY_PATH=$(LD_LIBRARY_PATH=/usr/local/lib${LIB_SUFFIX})`?

We may be calling system libraries at this point since core isn't installed yet, and those libraries are not in /usr/local...

Zopolis4: Packages already know how to find system libraries, otherwise they wouldn't work at all.

satmandu: It's a question of library priority. We do NOT want the system libraries preferred to chromebrew priorities, since the system libraries for libgcc are much much older on i686.

satmandu: > It will still find the newer libgcc first since `/usr/local/lib` is the higher priority path. I guess I still don't understand why we need the extra `:/usr/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX}` added to the path.

@uberhacker This is wrong. The `/etc/ld.so.conf` shows what the default library search path priorities are. 
on armv7l:
```
chronos@rpi4b-armv7l /usr/local/lib/crew/packages $ cat /etc/ld.so.conf 
# ld.so.conf autogenerated by env-update; make all changes to
# contents of /etc/env.d directory
/lib
/usr/lib
/usr/local/lib
```
on i686:
```
chronos@cheekon-i686 /usr/local/lib/crew/packages $ cat /etc/ld.so.conf 
# ld.so.conf autogenerated by env-update; make all changes to
# contents of /etc/env.d directory
/lib
/usr/lib
/usr/local/lib
```
on x86_64:
```
chronos@cheekon-x86_64 /usr/local/lib/crew/packages $ cat /etc/ld.so.conf 
# ld.so.conf autogenerated by env-update; make all changes to
# contents of /etc/env.d directory
/lib64
/usr/lib64
/usr/local/lib64
/lib
/usr/lib
/usr/local/lib
```
On ALL of these images, the system libs are higher in the search hierarchy. I'm just trying to override this once when first running crew to get the crew library folder.

satmandu: This is the error I see: 

image

satmandu: (And again, this issue is ONLY for i686.)

satmandu: This is now simplified.

uberhacker: Nice!

satmandu:
@uberhacker is going to test this tonight, so don't merge before he does that.

uberhacker:
On x86_64, I get:
```
chronos@localhost /usr/local $ OWNER=satmandu BRANCH=git_no_musl ./install.sh

,,,snip...

crew_profile_base: Crew-profile-base sets up Chromebrew's environment capabilities.
Initialized empty Git repository in /usr/local/tmp/crew/crew_profile_base.20221002133701.dir/.git/
remote: Enumerating objects: 19, done.
remote: Counting objects: 100% (19/19), done.
remote: Compressing objects: 100% (16/16), done.
remote: Total 19 (delta 0), reused 6 (delta 0), pack-reused 0
Unpacking objects: 100% (19/19), 16.69 KiB | 4.17 MiB/s, done.
From https://github.com/chromebrew/crew-profile-base
 * tag               0.0.5      -> FETCH_HEAD
HEAD is now at c68720d Fix `stat: cannot statx '': No such file or directory` error

...snip...

Performing post-install for mandb...
mandb: warning: $MANPATH set, ignoring /usr/local/etc/man_db.conf
Processing manual pages under /usr/local/share/man...
Updating index cache for path `/usr/local/share/man/man8'. Wait...mandb: warning: /usr/local/share/man/man8/saslauthd.8.gz: whatis parse for saslauthd(8) failed
Updating index cache for path `/usr/local/share/man/man3'. Wait...mandb: warning: /usr/local/share/man/man3/pcre2demo.3.gz: whatis parse for pcre2demo(3) failed
mandb: warning: /usr/local/share/man/man3/pcredemo.3.gz: whatis parse for pcredemo(3) failed
done.

...snip...

Creating symlinks to system glibc version to prevent breakage.
Paring locales to a minimal set.
/usr/local/lib64/ruby/3.1.0/fileutils.rb:1446: [BUG] Bus Error at 0x00007b8d9e711b0c
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux-gnu]

-- Control frame information -----------------------------------------------
c:0022 p:---- s:0104 e:000103 CFUNC  :initialize
c:0021 p:---- s:0101 e:000100 CFUNC  :unlink
c:0020 p:0014 s:0096 e:000095 BLOCK  /usr/local/lib64/ruby/3.1.0/fileutils.rb:1446
c:0019 p:0007 s:0093 e:000092 METHOD /usr/local/lib64/ruby/3.1.0/fileutils.rb:1451
c:0018 p:0004 s:0087 e:000086 METHOD /usr/local/lib64/ruby/3.1.0/fileutils.rb:1445
c:0017 p:0019 s:0083 e:000082 METHOD /usr/local/lib64/ruby/3.1.0/fileutils.rb:788
c:0016 p:0007 s:0077 e:000076 BLOCK  /usr/local/lib64/ruby/3.1.0/fileutils.rb:575 [FINISH]
c:0015 p:---- s:0073 e:000072 CFUNC  :each
c:0014 p:0052 s:0069 e:000068 METHOD /usr/local/lib64/ruby/3.1.0/fileutils.rb:574
c:0013 p:0011 s:0060 e:000059 METHOD /usr/local/lib64/ruby/3.1.0/fileutils.rb:589
c:0012 p:0024 s:0052 e:000051 BLOCK  /usr/local/lib/crew/packages/glibc.rb:637 [FINISH]
c:0011 p:---- s:0048 e:000047 CFUNC  :each
c:0010 p:0015 s:0044 e:000043 BLOCK  /usr/local/lib/crew/packages/glibc.rb:634 [FINISH]
c:0009 p:---- s:0041 e:000040 CFUNC  :chdir
c:0008 p:0014 s:0036 e:000035 BLOCK  /usr/local/lib/crew/packages/glibc.rb:633 [FINISH]
c:0007 p:---- s:0032 e:000031 CFUNC  :each
c:0006 p:0434 s:0028 e:000027 METHOD /usr/local/lib/crew/packages/glibc.rb:632
c:0005 p:0032 s:0024 e:000023 BLOCK  /usr/local/bin/crew:1775 [FINISH]
c:0004 p:---- s:0020 e:000019 CFUNC  :each
c:0003 p:0008 s:0016 e:000015 METHOD /usr/local/bin/crew:1771
c:0002 p:0611 s:0011 E:001158 EVAL   /usr/local/bin/crew:1873 [FINISH]
c:0001 p:0000 s:0003 E:001590 (none) [FINISH]

...snip...
```

satmandu:
Wow. That's a new one. ;/

satmandu:
@uberhacker Can you try this installer again? This is now modified to avoid using git during the early install. It works in an i686 container.

install.sh
```diff
@@ -272,6 +280,25 @@ echo_info "Setup and synchronize local package repo..."
...
+crew update compatible
+
+echo_info "Installing core Chromebrew packages...\n"
+yes | crew install core
+
+echo_info "\nRunning Bootstrap package postinstall scripts...\n"
+crew postinstall $BOOTSTRAP_PACKAGES
+
+echo_info "Synchronizng local package repo..."
+cd "${CREW_LIB_PATH}"/..
```
uberhacker, line 301: Is this line needed?

install.sh
```diff
@@ -272,6 +280,25 @@ echo_info "Setup and synchronize local package repo..."
 echo -e "${GRAY}"
 
 # Remove old git config directories if they exist
+rm -rf "${CREW_LIB_PATH}" && mkdir "${CREW_LIB_PATH}"
+
+cd "${CREW_LIB_PATH}" && curl -L https://github.com/"${OWNER}"/"${REPO}"/tarball/"${BRANCH}" | tar -xz --strip-components=1
```
supechicken, line 285: Use `-C` in `tar`?
```shell
curl -L https://github.com/"${OWNER}"/"${REPO}"/tarball/"${BRANCH}" | tar -xz --strip-components=1 -C "${CREW_LIB_PATH}"
```

supechicken: (also append the `-#` option to `curl` to show the download progress bar)

install.sh
```diff
@@ -272,6 +280,25 @@ echo_info "Setup and synchronize local package repo..."
 echo -e "${GRAY}"
 
 # Remove old git config directories if they exist
+rm -rf "${CREW_LIB_PATH}" && mkdir "${CREW_LIB_PATH}"
```
supechicken, line 283: Use `find` to delete all stuff including hidden files:
```shell
find "${CREW_LIB_PATH}" -mindepth 1 -delete
```

packages/crew_profile_base.rb
```diff
@@ -11,6 +11,7 @@ class Crew_profile_base < Package
   source_url 'https://github.com/chromebrew/crew-profile-base.git'
   git_hashtag @_ver
 
+  depends_on 'git'
```
supechicken, line 14: Better to remain it in `core.rb`? (since it is required for git source functionality in Chromebrew)

packages/glibc.rb
```diff
@@ -612,7 +612,7 @@ def self.postinstall
     end
     if @libc_version.to_f >= 2.32
       puts 'Paring locales to a minimal set.'.lightblue
-      system 'localedef --list-archive | grep -v -i -e ^en -e ^cs -e ^de -e ^es -e ^fa -e ^fe -e ^it -e ^ja -e ^ru -e ^tr -e ^zh -e ^C| xargs localedef --delete-from-archive'
+      system 'localedef --list-archive | grep -v -i -e ^en -e ^cs -e ^de -e ^es -e ^fa -e ^fe -e ^it -e ^ja -e ^ru -e ^tr -e ^zh -e ^C| xargs localedef --delete-from-archive || true'
```
supechicken, line 615: Rubyize?
```ruby
system 'localedef --list-archive | grep -v -i -e ^en -e ^cs -e ^de -e ^es -e ^fa -e ^fe -e ^it -e ^ja -e ^ru -e ^tr -e ^zh -e ^C| xargs localedef --delete-from-archive', exception: true
```

install.sh
```diff
@@ -218,12 +218,20 @@ function extract_install () {
     #extract and install
     echo_intra "Extracting ${1} ..."
     if [[ "$2" == *".zst" ]];then
```
supechicken, line 220: Use `case`?
```shell
case "${2}" in
*.zst)
  ...
;;
*.tpxz)
  ...
;;
*)
;;
esac
```

install.sh
```diff
@@ -24,10 +24,10 @@ CREW_PACKAGES_PATH="${CREW_LIB_PATH}/packages"
 ARCH="${ARCH/armv8l/armv7l}"
 
 # BOOTSTRAP_PACKAGES cannot depend on crew_profile_base for their core operations (completion scripts are fine)
-BOOTSTRAP_PACKAGES="musl_zstd pixz ca_certificates git gmp ncurses xxhash lz4 popt libyaml openssl zstd gcc rsync ruby"
-
-# Add musl bin to path
-PATH=/usr/local/share/musl/bin:$PATH
+# rsync requires openssl xxhash lz4 popt
+BOOTSTRAP_PACKAGES="zstd pixz ca_certificates ruby rsync openssl xxhash lz4 popt"
+# i686 requires gcc and openssl
```
Zopolis4, line 29:
```
# i686 requires gcc and openssl
```
What for? (i.e. why does i686 *need* these packages when bootstrapping)

install.sh
```diff
@@ -218,12 +218,20 @@ function extract_install () {
...
+        exit 1
+      fi
     elif [[ "$2" == *".tpxz" ]];then
-      if ! LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} pixz -h &> /dev/null; then
-        tar xpf ../"${2}"
+      if "${CREW_PREFIX}"/bin/pixz -h &> /dev/null; then
+        tar -I "${CREW_PREFIX}"/bin/pixz -xpf ../"${2}"
       else
-        LD_LIBRARY_PATH=${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX} tar -Ipixz -xpf ../"${2}"
+        echo "Pixz is broken or nonfunctional, and some packages can not be extracted properly."
```
Zopolis4, line 233: Why are we using pixz over system now?

packages/crew_profile_base.rb
satmandu: Without this dependency listed, `crew_profile_base` tries to install before `git` is installed. That breaks installs.

install.sh
satmandu: i686 installs have such an old system they have an older libgcc and libssl than what our packages are built against.

install.sh
satmandu: We have packages with no updates that we haven't had to rebuild.

Zopolis4: What was wrong with the previous tar -xvf method?

uberhacker (approved):
Works on x86_64.

install.sh
Zopolis4: So does linking fail for ruby, git and ca_certificates?

Zopolis4: (i.e. before we install core packages)

satmandu: I don't understand the question. Apps complain about the older libraries if only those are available, and then the apps crash. This is the case for ruby for sure. Git is no longer an issue.

PR #7455, opened 2022-10-04
Run preflight check on dependencies also, remove `autoremove` command

supechicken:
The `File.exist?` check of `jdk*.rb` will be moved to `self.preflight` in my next PR, and the current installation logic does not run preflight on dependencies, which breaks the `jdk` fake package.

Also removed the `autoremove` command since it is broken for a long while + cleanup

Tested on `x86_64`

### Changes
- Run compatibility check for `@pkg` earlier
- Run preflight check on dependencies
- Add a new function (`@pkg.compatible?`) for compatibility check (`lib/package.rb`)
- Remove `crew autoremove` command

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1318,7 +1318,7 @@ def resolve_dependencies
   puts 'The following packages also need to be installed: '
 
   @dependencies.each do |dep|
-    abort "Dependency #{dep} was not found.".lightred unless File.exist?("#{CREW_PACKAGES_PATH}/#{dep}.rb")
+    abort "Dependency #{dep} was not found.".lightred unless File.exist?( File.join(CREW_PACKAGES_PATH, "#{dep}.rb") )
   end
 
   puts @dependencies.join(' ')
--- a/bin/crew
+++ b/bin/crew
@@ -1317,7 +1317,7 @@ def resolve_dependencies
 
   # run preflight check for dependencies
   @dependencies.each do |depName|
-    dep_pkgPath = File.join(CREW_PACKAGES_PATH, depName)
+    dep_pkgPath = File.join(CREW_PACKAGES_PATH, "#{depName}.rb")
     Package.load_package(dep_pkgPath, depName).preflight
   end
 
--- a/bin/crew
+++ b/bin/crew
@@ -154,13 +154,6 @@ end
   end
 end
 
-def is_compatible?(pkgName)
-  pkgPath = File.join(CREW_PACKAGES_PATH, "#{pkgName}.rb")
-  pkg     = Package.load_package(pkgPath, pkgName)
-
-  return pkg.compatible?
-end
-
 def print_package(pkgPath, extra = false)
   pkgName = File.basename pkgPath, '.rb'
   begin
--- a/bin/crew
+++ b/bin/crew
@@ -1340,7 +1340,7 @@ def resolve_dependencies
     print_current_package
     install
   end
-  if @resolve_dependencies_and_install.eql?(1) or @resolve_dependencies_and_build.eql?(1)
+  if @resolve_dependencies_and_install.eql?(1) || @resolve_dependencies_and_build.eql?(1)
     @to_postinstall = @dependencies
   else
     @dependencies.each do |dep|
--- a/bin/crew
+++ b/bin/crew
@@ -154,7 +154,7 @@ end
   end
 end
 
-def is_compatible? (pkgName)
+def is_compatible?(pkgName)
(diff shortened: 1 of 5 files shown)


satmandu:
Is autoremove not fixable? It would be nice to have...

bin/crew
```diff
@@ -156,6 +154,13 @@ end
   end
 end
 
+def is_compatible?(pkgName)
```
uberhacker, line 157: Unused function?

uberhacker (changes requested):
I agree with @satmandu that we should try to fix rather than remove.

bin/crew
supechicken: Removed

supechicken:
The current `autoremove` logic is nearly unfixable, nearly every package's `is_dep?` key in `device.json` are `true`, which makes `autoremove` thinks they are no longer required by any package and can be removed.

So remove it for a while to prevent further damage to user's `crew` installation before I have ideas on how to reimplement it :)

uberhacker (changes requested):
OK, so if we're going to remove it, README.md should be updated too.

supechicken:
> OK, so if we're going to remove it, README.md should be updated too.

Done

lib/package.rb
```diff
@@ -157,6 +159,15 @@ def self.#{prop}
     end
   end
 
+  def self.compatible?
+    if @compatibility
+      return (@compatibility.casecmp?('all') or @compatibility.include?(ARCH))
+    else
+      warn "#{name}: Missing `compatibility` field.".yellow
+      return true
```
uberhacker, line 167: `return false`?  We do not want to continue if `@compatibility` is not set or false, right?

supechicken: Changed to `abort`

lib/package.rb
```diff
@@ -157,6 +159,14 @@ def self.#{prop}
     end
   end
 
+  def self.compatible?
+    if @compatibility
+      return (@compatibility.casecmp?('all') or @compatibility.include?(ARCH))
+    else
+      abort "#{name}: Missing `compatibility` field.".lightred
```
uberhacker, line 166: Return `false` would be better since:
```
abort "Package #{@pkg.name} is not compatible with your device architecture (#{ARCH}) :/".lightred unless @pkg.compatible?
```
?

uberhacker (changes requested):

lib/package.rb
supechicken: Change made

uberhacker (changes requested):

bin/crew
```diff
@@ -1301,24 +1301,24 @@ def expand_dependencies
...
+    Package.load_package(dep_pkgPath, depName).preflight
   end
 
   return if @dependencies.empty?
 
   puts 'The following packages also need to be installed: '
 
   @dependencies.each do |dep|
-    abort "Dependency #{dep} was not found.".lightred unless File.exist?("#{CREW_PACKAGES_PATH}#{dep}.rb")
+    abort "Dependency #{dep} was not found.".lightred unless File.exist?("#{CREW_PACKAGES_PATH}/#{dep}.rb")
```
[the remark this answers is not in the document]
supechicken: It was added to clarify that they are not a single filename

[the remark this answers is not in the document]
uberhacker: But it adds an extra `/` unnecessarily.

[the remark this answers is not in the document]
supechicken: Okay I think I found a better way to do it:
```ruby
File.join(CREW_PACKAGES_PATH, "#{dep}.rb")
```

uberhacker (approved):

uberhacker:
Rubocop doesn't like spaces between parenthesis.

supechicken:
But it does make the code more clear when in nested brackets :)

uberhacker:
Can we change or ignore the Rubocop rule?

supechicken:
Yeah, just disable `Layout/SpaceInsideParens` in `.rubocop.yml` like in #7457

uberhacker:
If you create a PR for that then this PR can be merged. :)

uberhacker (approved):

uberhacker:
For some reason, the space inside parentheses check is still in effect.

supechicken:
> For some reason, the space inside parentheses check is still in effect.

Because the check have not re-run yet

Zopolis4:
*because you havent rebased

PR #7457, opened 2022-10-05
Re-enable And/Or cop, lint tree

Zopolis4:
@supechicken, please actually lint the tree when you change the rubocop file, otherwise it will break all new prs made on it, and prs that rebase.

In regards to the And/Or cop, `&&` and `||` have been used as boolean operators since the dawn of time, and are very easy to find the meaning of if a beginner does not understand. Also, using `or` and `and` will only help English-speaking beginners. Furthermore, allowing `or` and `and` to be used alongside `&&` and `||` will make things even worse, as a beginner would presume they have a different meaning to the english variants.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -124,9 +124,6 @@ Layout/FirstHashElementIndentation:
 Layout/IndentationStyle:
   EnforcedStyle: spaces
 
-Style/AndOr:
-  Enabled: false
-
 Style/RedundantReturn:
   Enabled: false
 
--- a/lib/progress_bar.rb
+++ b/lib/progress_bar.rb
@@ -57,7 +57,7 @@ class ProgressBar
         return false
       end
     end
- 
+
     if @start_time
       @elapsed_time = (Time.now - @start_time).to_i
     else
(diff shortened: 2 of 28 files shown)


supechicken:
> using or and and will only help English-speaking beginners.

I don’t think there would be a contributor that doesn't understand basic English like `and`…

@uberhacker @satmandu What are your thoughts on this?

Zopolis4:
I'm not saying they wont understand it, but its not going to be particularly more helpful than `&&` or similar. It also has the converse effect, as it makes it more difficult for experienced programmers (or at least me) to understand, because they are not used to seeing logical (or other) operators written in english.

uberhacker:
@Zopolis4: I'm fine with this but, in general, I'm more favorable to less restrictions than more; especially if it has to done with style instead of function.  Thank you for removing the trailing white space.  Packagers should set their editor/ide to do this also.

Zopolis4:
Personally, I prefer consistency over excessive choice. Especially when it comes to a binary choice.

uberhacker:
Consistency is great but if something fails to get past the rubocop there should be a clear way forward to fix.  Otherwise, packagers will get discouraged.  We need to make it as easy as possible for new contributors.

PR #7473, opened 2022-10-08
lib/selector: Add selector library, apply to `jdk` and `hunspell`

supechicken:
`lib/selector.rb`: Show a menu with countdown for selecting providers in fake packages

Also applied to `packages/{jdk,hunspell}.rb`

Tested on `x86_64`


supechicken:
```shell
CREW_TESTING=1 CREW_TESTING_REPO=https://github.com/supechicken/chromebrew CREW_TESTING_BRANCH=selector crew update
crew install jdk
```

satmandu (approved):

packages/jdk.rb
```diff
@@ -1,43 +1,43 @@
 require 'package'
 
 class Jdk < Package
-  description 'The JDK is a development environment for building applications, applets, and components using the Java programming language.'
-  homepage 'https://www.oracle.com/java/technologies/javase/javase-jdk8-downloads.html'
-  version '8u333-18.0.1.1'
+  description 'The Oracle JDK is a development environment for building applications, applets, and components using the Java programming language.'
+  homepage 'https://www.oracle.com/java'
+  version 'LATEST'
```
uberhacker, line 6: Can we keep the old version so it can be incremented with newer releases?

packages/hunspell.rb
```diff
@@ -2,37 +2,20 @@
...
-        depends_on 'hunspell_fr_fr'
-        break
-      when '3'
-        depends_on 'hunspell_es_any'
-        break
-      when '0'
-        abort
-      end
-    end
+    depends_on Selector.new(options).show_prompt if ARGV.include?('install')
```
uberhacker, line 19: Can we wrap the whole block inside `if ARGV.include?('install') ... end`?  That way options is not assigned for every argument.

packages/jdk.rb
```diff
@@ -1,43 +1,43 @@
...
+        majver     = jdkver.split('.')[0]
+        majver     = '8' if majver == '1'
+        pkg_suffix = is_openjdk ? 'openjdk' : 'jdk'
+
+        puts "Package #{pkg_suffix}#{majver} already installed.".lightgreen
+        warn "Run `crew remove #{pkg_suffix}#{majver} && crew install jdk` to install a different version.".yellow
+        return false
+      end
+
+      jdk_latest_ver = %w[8 11 17 18].to_h do |jdk_majver|
```
uberhacker, line 27: How about instead of hardcoding the versions, search for `jdk#.rb` in packages and use whatever is found instead?  That way never packages are added without modifying the code.

packages/jdk.rb
```diff
@@ -1,43 +1,43 @@
...
+        pkg = Package.load_package("#{CREW_PACKAGES_PATH}/jdk#{jdk_majver}.rb")
+
+        [jdk_majver, pkg.version]
       end
+
+      options = [
+        { value: 'jdk8',  description: 'Oracle JDK 8'  },
+        { value: 'jdk11', description: 'Oracle JDK 11' },
+        { value: 'jdk17', description: 'Oracle JDK 17' },
+        { value: 'jdk18', description: 'Oracle JDK 18' }
```
uberhacker, line 37: See earlier comment about hardcoding.

uberhacker:
This looks very interesting @supechicken!  Nice work.

uberhacker:
It'd be nice if we did this for the php and openjdk packages as well.

Zopolis4:
Would it not make more sense to make this a new class ala #6306?

satmandu:
@supechicken Any updates on this? Also we should use this on `wxwidgets{,30,31}.rb`...

packages/jdk.rb
supechicken: That's what this block initially do :)

packages/hunspell.rb
supechicken: Change made

packages/jdk.rb
```diff
@@ -1,43 +1,51 @@
...
-  description 'The JDK is a development environment for building applications, applets, and components using the Java programming language.'
-  homepage 'https://www.oracle.com/java/technologies/javase/javase-jdk8-downloads.html'
-  version '8u333-18.0.1.1'
+  description 'The Oracle JDK is a development environment for building applications, applets, and components using the Java programming language.'
+  homepage 'https://www.oracle.com/java'
   license 'Oracle-BCLA-JavaSE'
   compatibility 'all'
 
+  # full version number extracted from jdk*.rb recipes
+  @avail_jdk_ver = [8, 11, 17, 18].to_h do |jdk_majver|
```
uberhacker, line 10: How about instead of an array of versions (still hardcoded), why not do something like this:
```
Dir["#{CREW_PACKAGES_PATH}/jdk*.rb"].each do |package|
  pkg = Package.load_package("#{CREW_PACKAGES_PATH}/#{package}")
  ...
```

supechicken:
> Would it not make more sense to make this a new class ala #6306?

What do you mean? This is already a class:
https://github.com/chromebrew/chromebrew/blob/c361337df40c00268bdb5ee5cc70011d53ae7426/lib/selector.rb#L5

packages/jdk.rb
```diff
@@ -1,43 +1,51 @@
...
+          Run "crew remove #{pkg_prefix}#{jdk_major_ver}; crew install #{name}" to install another version of JDK
+        EOT
       end
+
+      options = [
+        { value: 'jdk8',  description: "Oracle JDK #{@avail_jdk_ver[8]}"  },
+        { value: 'jdk11', description: "Oracle JDK #{@avail_jdk_ver[11]}" },
+        { value: 'jdk17', description: "Oracle JDK #{@avail_jdk_ver[17]}" },
+        { value: 'jdk18', description: "Oracle JDK #{@avail_jdk_ver[18]}" }
+      ]
```
uberhacker, line 46: Build `options` based on the package file versions found in `#{CREW_PACKAGES_PATH}/jdk*.rb`?

packages/jdk.rb
supechicken: Change made

packages/jdk.rb
supechicken: Change made

uberhacker: Nice!

PR #7478, opened 2022-10-10
asciidoc => 10.2.0

supechicken:
All binaries are built on real machines and uploaded :)

Build properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`


supechicken:
The x86_64 binary need to rebuild since it was linked with glibc 2.33😅

satmandu (approved):

PR #7482, opened 2022-10-11
lib/downloader: Fix progress bar length on older CrOS terminals

supechicken:
On older CrOS terminals (mainly `i686`), the progress bar will overflow since a width bug will occur when using the box-drawing char with bold character color form (`1;36m`)

Changed the front/background color code to its non-bold equivalent in this PR

```shell
CREW_TESTING=1 CREW_TESTING_REPO=https://github.com/supechicken/chromebrew CREW_TESTING_BRANCH=fix_progbar_color crew update
```

### DIFF
--- a/lib/progress_bar.rb
+++ b/lib/progress_bar.rb
@@ -11,10 +11,9 @@ class ProgressBar
     # character used to fill the progress bar, one of the box-drawing character in unicode
     @bar_char = "\u2501"
 
-    # color scheme of progress bar, can be changed
-    # see color.rb for more available colors
-    @bar_front_color = :lightcyan
-    @bar_bg_color = :gray
+    # color scheme of progress bar (in ANSI color codes), can be changed
+    @bar_front_color = [ 96, 0 ]
+    @bar_bg_color = [ 90, 0 ]
 
     # all info blocks with space taken
     @info_before_bar = { downloaded_size_in_str: 20 }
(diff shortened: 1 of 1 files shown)


lib/progress_bar.rb
```diff
@@ -11,10 +11,9 @@ def initialize(total_size)
     # character used to fill the progress bar, one of the box-drawing character in unicode
     @bar_char = "\u2501"
 
-    # color scheme of progress bar, can be changed
-    # see color.rb for more available colors
-    @bar_front_color = :lightcyan
-    @bar_bg_color = :gray
+    # color scheme of progress bar (in ANSI color codes), can be changed
+    @bar_front_color = [96, 0]
+    @bar_bg_color = [90, 0]
```
uberhacker, line 16: Why the change?  What is wrong with the more descriptive symbol notation?

supechicken: Since the `90-99` color code does not exist in `color.rb`

supechicken: (thinking what should they name if we decided to add them to `color.rb` since `light*` prefix are already assigned to the bold form

uberhacker: Prefix with `lite`?  Add to `color.rb`?

uberhacker (approved):
Testing works fine.

PR #7488, opened 2022-10-13
add uutils_coreutils package

satmandu:
Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=uutils CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/uutils_coreutils.rb
+++ b/packages/uutils_coreutils.rb
@@ -40,7 +40,7 @@ class Uutils_coreutils < Package
       abort "crew remove #{cutils} && crew install uutils_coreutils".lightblue
     end
   end
- 
+
   def self.build
     system 'make PROFILE=release MULTICALL=y'
   end
--- a/packages/uutils_coreutils.rb
+++ b/packages/uutils_coreutils.rb
@@ -1,0 +1,42 @@
+# Adapted from Arch Linux uutils-coreutils PKGBUILD at:
+# https://github.com/archlinux/svntogit-community/raw/packages/uutils-coreutils/trunk/PKGBUILD
+
+require 'package'
+
+class Uutils_coreutils < Package
+  description 'Cross-platform Rust rewrite of the GNU coreutils'
+  homepage 'https://github.com/uutils/coreutils'
+  version '0.0.16'
+  license 'MIT'
+  compatibility 'all'
+  source_url 'https://github.com/uutils/coreutils.git'
+  git_hashtag version
+
+  binary_url({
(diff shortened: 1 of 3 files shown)


packages/uutils_coreutils.rb
```diff
@@ -0,0 +1,42 @@
...
+     armv7l: '819bf74dc4d1cdbe7fb45cb711b73f2494c14a7779644905a8b50e8bc8739451',
+       i686: '9b6f7f69737350b817fb88ac97f918f8db25ea1418f045446a63783f99c8db7d',
+     x86_64: '082a81168509e3c1a0ff43f8f89e82fb46f4e2c2aa89d3dd5a53a16a7317b8a4'
+  })
+
+  depends_on 'rust' => :build
+  depends_on 'sphinx' => :build
+  depends_on 'gcc' # R
+  depends_on 'glibc' # R
+  conflicts_ok # conflicts with coreutils
```
uberhacker, line 32: What are the conflicts and will it create a problem if both are installed and one is removed afterwards?

satmandu: The uutils coreutils is a rust-written alternative to the gnu coreutils, and replicating all functionality is a working goal of the project.

ChromeOS already includes a baseline set of core utils, so removing either of these packages should not break the system.

uberhacker: I have noticed removing packages like this makes the bash session explode in epic fashion since it expects to find executables in `/usr/local/bin`.  Something like coreutils has a lot of essential binaries too.

uberhacker: Also, if both of these packages are installed and one is removed it will delete files from the other package with the expectation those files are still installed if the other package is installed.  It would be best to install one or the other but not both like what we do with the jdk and wxwidgets packages.

satmandu: Just added that logic to both this file and coreutils...

uberhacker: Remove `conflicts_ok` line?

satmandu:
That's for building. It should be ok to have coreutils installed when
building the rust based version.

packages/uutils_coreutils.rb
satmandu: That's for building. It should be ok to have coreutils installed when
building the rust based version. (Since coreutils might be newer than what the system provides.) Otherwise we can't build this with coreutils installed.

We have the check now for during install.

uberhacker: Makes sense.  Fair enough.

uberhacker (approved):

PR #7494, opened 2022-10-14
rubocop: Disable some cops

supechicken:
### Re: `rubocop` checks in my PR
For me, the `rubocop` check result is more like a suggestion, and I may ignore some of them when I think the current one is better.

For example:
```ruby
# trying to align a hash table to make it looks better

# first try
example = {
      a: 1,
    bcd: 2,
   asdf: 3
}

# rubocop doesn't like it, however
# C: [Correctable] Layout/FirstHashElementIndentation: Use 2 spaces for indentation in a hash, ...
# second try, and this time I followed rubocop's suggestion
example = {
  :a    => 1,
  :bcd  => 2,
  :asdf => 3
}

# rubocop doesn't like it either
# [Correctable] Style/HashSyntax: Use the new Ruby 1.9 hash syntax.
```

What I want to say is to compare the current version and `rubocop`'s one first, and if you think the `rubocop` one is better, tell me and I will change it :)

#### `Layout/CommentIndentation, Layout/FirstHashElementIndentation`
- It stops me from aligning the hash tables...
#### `Layout/TrailingEmptyLines`
- Github will do it for us when the PR merge

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -128,7 +128,13 @@ Layout/FirstHashElementIndentation:
 Layout/IndentationStyle:
   EnforcedStyle: spaces
 
+Layout/SpaceInsideBlockBraces:
+  Enabled: false
+
 Layout/SpaceInsideParens:
+  Enabled: false
+
+Layout/TrailingEmptyLines:
   Enabled: false
 
 Style/RedundantReturn:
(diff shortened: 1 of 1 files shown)


uberhacker (approved):
Agreed.

PR #7498, opened 2022-10-15
jdk*: Make prompt more clearly + add openjdk detection

supechicken:
### Changes
- jdk8 => 8u341
- jdk17 => 17.0.4.1
- jdk18 => 18.0.2.1
- The jdk download prompt will tell you the version and exact filename of the archive now (to prevent misleading)
```
Oracle now requires an account to download the JDK.

You must login at https://login.oracle.com/mysso/signon.jsp and then visit:
https://www.oracle.com/java/technologies/javase/jdk18-archive-downloads.html

Download "jdk-18.0.2.1_linux-x64_bin.tar.gz" (Linux x64 Compressed Archive) to Chrome OS download folder to continue.
```
- The preflight message is updated and able to detect `openjdk*` now
```
OpenJDK 17.0.4.1 installed.

Run "crew remove openjdk17; crew install jdk8" to install this version of JDK
```
- Added backup link in case the version is no longer exists on the main download page
- Refactor to use `source_url` over `downloader()`

Tested on `x86_64`

### DIFF
--- a/packages/jdk11.rb
+++ b/packages/jdk11.rb
@@ -26,7 +26,7 @@ class Jdk11 < Package
       pkg_branding = is_openjdk ? 'OpenJDK' : 'Oracle JDK'
         pkg_prefix = is_openjdk ? 'openjdk' : 'jdk'
 
-      abort <<~EOT.yellow unless jdk_major_ver == '11'
+      abort <<~EOT.yellow unless jdk_major_ver == name.delete_prefix('jdk')
 
         #{pkg_branding} #{jdk_ver} installed.
 
--- a/packages/jdk11.rb
+++ b/packages/jdk11.rb
@@ -4,11 +4,11 @@ require 'uri'
 class Jdk11 < Package
   description 'The Oracle JDK is a development environment for building applications, applets, and components using the Java programming language.'
   homepage 'https://www.oracle.com/java'
-  version '11.0.16.1'
+  version "#{@_ver = '11.0.16.1'}-1"
   license 'Oracle-BCLA-JavaSE'
   compatibility 'x86_64'
 
-  source_url File.join('file://', HOME, 'Downloads', "jdk-#{version}_linux-x64_bin.tar.gz")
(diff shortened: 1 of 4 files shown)


satmandu:
Rubocop still not happy! 😅

packages/jdk11.rb
```diff
@@ -1,46 +1,85 @@
 require 'package'
+require 'uri'
 
 class Jdk11 < Package
-  description 'The JDK is a development environment for building applications, applets, and components using the Java programming language.'
-  homepage 'https://www.oracle.com/java/technologies/javase-jdk11-downloads.html'
+  description 'The Oracle JDK is a development environment for building applications, applets, and components using the Java programming language.'
+  homepage 'https://www.oracle.com/java'
   version '11.0.16.1'
```
uberhacker, line 7:
```
@_ver = '11.0.16.1'
version "#{@_ver}-1"
```
Replace all instances of `#{version}` with `#{@_ver}` since we need to bump the version?

uberhacker (changes requested):
Nice @supechicken, this is great!  And you even updated the releases.

packages/jdk11.rb
supechicken: Done

packages/jdk11.rb
```diff
@@ -4,11 +4,11 @@
 class Jdk11 < Package
   description 'The Oracle JDK is a development environment for building applications, applets, and components using the Java programming language.'
   homepage 'https://www.oracle.com/java'
-  version '11.0.16.1'
+  version "#{@_ver = '11.0.16.1'}-1"
```
uberhacker: I didn't know you could do this.  I might have some luck extracting the version in a script afterall.

supechicken: The `=` expression can return its value :) But `rubocop` doesn't like it anyways:
```
irb(main):005:0> a=1
=> 1
```

uberhacker (approved):

packages/jdk11.rb
uberhacker: Sounds like another tyrannical rubocop rule we can bypass?

PR #7506, opened 2022-10-17
Fix gem install command line

satmandu:
- Removing `--conservative` from the gem install command line fixes the installs of dependencies.
- Replace postinstall regex usage in ruby gem packages which doesn't work.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=gem_noconservative CREW_TESTING=1 crew update ; crew upgrade
```

### DIFF
--- a/packages/ruby_rubocop.rb
+++ b/packages/ruby_rubocop.rb
@@ -6,7 +6,7 @@ require 'package'
 class Ruby_rubocop < Package
   description 'A Ruby static code analyzer and formatter'
   homepage 'https://rubocop.org'
-  version '1.26.1'
+  version '1.36'
   compatibility 'all'
   source_url 'SKIP'
   # source_url 'https://github.com/rubocop/rubocop.git'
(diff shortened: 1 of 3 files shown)


packages/ruby_rubocop.rb
```diff
@@ -6,7 +6,7 @@
 class Ruby_rubocop < Package
   description 'A Ruby static code analyzer and formatter'
   homepage 'https://rubocop.org'
-  version '1.26.1'
+  version '1.36'
```
uberhacker, line 9: Why the big jump?

uberhacker (approved):

packages/ruby_rubocop.rb
satmandu: The version number never got updated...

PR #7515, opened 2022-10-19
fop: Fix JAVA_HOME

satmandu:
- as per slack discussion

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=fop_2 CREW_TESTING=1 crew update
```


saltedcoffii:
why is openjdk8's home /usr/local/jre anyway? shouldn't it be in lib or lib64?

satmandu:
The use of `lib64` is an abomination and we should avoid it wherever it isn't needed. Here it isn't needed.

Just my two bits...

uberhacker (approved):

PR #7519, opened 2022-10-19
mold -> 1.6.0

satmandu:
Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=mold160  CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/mold.rb
+++ b/packages/mold.rb
@@ -8,7 +8,7 @@ class Mold < Package
   homepage 'https://github.com/rui314/mold'
   version '1.6.0'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


saltedcoffii:
why are we using a tarball instead of a git repository?

satmandu:
It doesn't matter? It's all from the same place. I would consider them equivalent, especially when we are using a tagged release. (The tarball is slightly faster for large repos since we don't have to download and cache it, but of course this is a small repo for mold.)

uberhacker (approved):

PR #7530, opened 2022-10-22
Add ruby and yamllint syntax check to Rubocop.yml

uberhacker:
The idea here is to only run rubocop on ruby files.  Any other file type can be ignored.  Also we can sneak in simple yaml and ruby syntax checks.

### DIFF
--- a/.github/workflows/Rubocop.yml
+++ b/.github/workflows/Rubocop.yml
@@ -1,17 +1,47 @@
+---
 name: Rubocop
-
-on:
+on: # yamllint disable-line rule:truthy
   pull_request:
-    branches: [ master ]
-
+    branches: [master]
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
-    - name: Rubocop
-      uses: Freshly/Octocop@v0.0.2
-      with:
-        github_token: ${{ secrets.github_token }}
-        reporter: github-pr-check
-        rubocop_flags: "-c .rubocop.yml"
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Install ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 3.1
+      - name: Install yamllint
+        run: |
+          sudo apt install yamllint -y
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v33
+      - name: Syntax check
+        run: |
+          ruby=
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            ext="${file##*.}"
+            type="$(file -b $file | cut -d' ' -f1)"
+            [[ "$type" == "Ruby" ]] && ext="rb"
+            case $ext in
+              yml|yaml)
+                yamllint $file;;
+              rb)
+                ruby=true
+                ruby -wcWlevel=2 $file;;
+              *)
+                echo "Unable to check syntax of $file.";;
+            esac
+          done
+          echo "ruby=$ruby" >> $GITHUB_ENV
+      - name: Rubocop
+        if: env.ruby
+        uses: Freshly/Octocop@v0.0.2
+        with:
+          github_token: ${{ secrets.github_token }}
+          reporter: github-pr-check
+          rubocop_flags: "-c .rubocop.yml"


uberhacker:
I even used the linting to check the syntax of Rubocop.yml.

satmandu:
Github's own linter is flagging line 4. 😅

uberhacker:
> Github's own linter is flagging line 4. sweat_smile
Actually, that is a yamllint warning.  The error was too many blank lines. :face_with_spiral_eyes:

Zopolis4:
Lets put yml syntax checking in its own workflow, shall we?

Zopolis4:
Rubocop is already pretty smart, by the way, and won't run itself on non-ruby files. This applies to both locally and via Octocop.

In fact, looking at this ruby check, it seems this would break linting for all our ruby files that dont end in .rb. Think `bin/crew`, `tests/cycle_test`, etc.

uberhacker:
> Rubocop is already pretty smart, by the way, and won't run itself on non-ruby files. This applies to both locally and via Octocop.
> 
> In fact, looking at this ruby check, it seems this would break linting for all our ruby files that dont end in .rb. Think `bin/crew`, `tests/cycle_test`, etc.

I have the cases of files not ending in `.rb` covered.  Take a look again.

uberhacker:
> Lets put yml syntax checking in its own workflow, shall we? I think its great to have, but I think people getting yamlint failures from Rubocop.yml would be confusing.

Putting the syntax check in a separate workflow would slow the process down significantly because a whole new ubuntu instance would need to be spun up.  The name of the file doesn't mean anything from the front end.  All they would see is the error.

Zopolis4:
It's entirely possible that I am just reading the bash wrong, but I don't see it. Regardless, there's no need to have a check like this when rubocop already handles it, so the only needed part here is the yaml stuff, which should be in another pr.
```
            type="$(file -b $file | cut -d' ' -f1)"
            [[ "$type" == "Ruby" ]] && ext="rb"
            case $ext in
              rb)
                ruby=true
                ruby -wcWlevel=2 $file;;
              *)
                echo "Unable to check syntax of $file.";;
```

uberhacker:
> It's entirely possible that I am just reading the bash wrong, but I don't see it. Regardless, there's no need to have a check like this when rubocop already handles it, so the only needed part here is the yaml stuff, which should be in another pr.
> 
> ```
>             type="$(file -b $file | cut -d' ' -f1)"
>             [[ "$type" == "Ruby" ]] && ext="rb"
>             case $ext in
>               rb)
>                 ruby=true
>                 ruby -wcWlevel=2 $file;;
>               *)
>                 echo "Unable to check syntax of $file.";;
> ```
This would abort the process early on if there is a syntax error and not run rubocop.  Again, the idea is to speed up the checks.

Zopolis4:
What happens if someone only modifies the .rubocop.yml file? Changes that would cause lint to fail could sneak by this way, no? 

I feel dubious about conditionally running checks regardless. It's not like we're actually building anything, the average Rubocop CI check takes under 100 seconds. This change does seem to save about a minute on PR's that don't modify any ruby, which is good, but I personally don't feel the risk that it lets something through worth it given that almost every PR modifies ruby files.

But if you add something to catch .rubocop.yml modifications and split the yaml stuff into another workflow, then I don't see any objective issues with this PR.
image

uberhacker:
> What happens if someone only modifies the .rubocop.yml file? Changes that would cause lint to fail could sneak by this way, no?
> 
> I feel dubious about conditionally running checks regardless. It's not like we're actually building anything, the average Rubocop CI check takes under 100 seconds. This change does seem to save about a minute on PR's that don't modify any ruby, which is good, but I personally don't feel the risk that it lets something through worth it given that almost every PR modifies ruby files.
> 
> But if you add something to catch .rubocop.yml modifications and split the yaml stuff into another workflow, then I don't see any objective issues with this PR. image

Again, the goal is to speed things up.  Separate workflows slow things down.

uberhacker:
@supechicken: Any thoughts?

Zopolis4:
> Separate workflows slow things down. 

I'm not sure if they do, as they run in parallel, but even if they slowed things down they'd still be worth saving the confusion of getting yaml errors from a rubocop CI.

> Again, the goal is to speed things up.

But that's not the main point here. I'm all for speeding things up, but we're talking about 1 minute as opposed 2 minutes here. This isn't a meaningful amount of time, and its not like anything is ever waiting on these CI's. You can always just hook up a pre-commit and then they'll always be green.


> A pull requests that modifies .rubocop.yml will receive a yamllint check.

But rubocop won't be rerun, which is the issue. If a modification re-enables or changes a cop, that will almost certainly lead to the rubocop CI having different results, which needs to be picked up.

Zopolis4:
In regards to speed, once everything is tidied up and merged here I'll add caching to the workflows which will significantly cut down on the time as well.

uberhacker:
> > Separate workflows slow things down.
> 
> I'm not sure if they do, as they run in parallel, but even if they slowed things down they'd still be worth saving the confusion of getting yaml errors from a rubocop CI.
>
But they shouldn't run in parallel.  The idea is to run the syntax checks first to see if rubocop is needed.  If not, the workflow is skipped.

> > Again, the goal is to speed things up.
> 
> But that's not the main point here. I'm all for speeding things up, but we're talking about 1 minute as opposed 2 minutes here. This isn't a meaningful amount of time, and its not like anything is ever waiting on these CI's. You can always just hook up a pre-commit and then they'll always be green.
> 
Not sure what you mean here.

> > A pull requests that modifies .rubocop.yml will receive a yamllint check.
> 
> But rubocop won't be rerun, which is the issue. If a modification re-enables or changes a cop, that will almost certainly lead to the rubocop CI having different results, which needs to be picked up.

Why would a new pull request for `.rubocop.yml` need to rerun rubocop?  This pull request will ignore rubocop if the only file modified was `.rubocop.yml`.  If you want to test the new rubocop rules, then you would need to add a `.rb` package file or follow up with a separate pull request.

Zopolis4:
> Why would a new pull request for .rubocop.yml need to rerun rubocop?

Because if it modifies the rubocop config, it will modify the effects of rubocop. If I re-enable a cop, that will affect what rubocop does. That could make it so the CI now fails. But If we don't rerun rubocop on a modified .rubocop.yml, we won't notice until innocent PR's start having their CI fail.

> This pull request will ignore rubocop if the only file modified was .rubocop.yml.

IMO, this is a bug. See above.

>  If you want to test the new rubocop rules, then you would need to add a .rb package file or follow up with a separate pull request.

Why? (I'm not asking why in terms of code, I'm asking why you think this is a good idea)

** **

> But they shouldn't run in parallel. The idea is to run the syntax checks first to see if rubocop is needed. If not, the workflow is skipped.

I'm talking about making the yaml workflow and the rubocop workflow separate for the sake of clarity.

> Not sure what you mean here.

I'm saying we are talking about speeding up an already very fast CI. It's fast enough, in fact, that we don't need to sacrifice clarity to make it faster. We can have two CI's, one for rubocop and one for yamlint. They're both very fast CI's already.

.github/workflows/Rubocop.yml
```diff
@@ -1,17 +1,47 @@
+---
 name: Rubocop
-
 on:
```
satmandu, line 3: Can you replace this line with this to avoid the linter warning:
```
on:  # yamllint disable-line rule:truthy
```

satmandu (approved):
I'm ok with both ruby and yaml linting happening in the same workflow, as that does speed up the workflow. I think it is meaningful to minimize the usage of server resources when possible, even at the risk of making complexity slightly higher.

satmandu (approved):

PR #7557, opened 2022-10-26
zlib & minizip => 1.2.13

saltedcoffii:
## Description
upgrade zlib and minizip to their latest versions

## Additional information

Works properly:
- [ ] `x86_64`
- [ ] `i686`
- [ ] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=zlib_1.2.13 CREW_TESTING=1 crew update
```


packages/minizip.rb
```diff
@@ -1,40 +1,34 @@
 require 'package'
+require_relative 'zlibpkg'
 
 class Minizip < Package
   description 'Minizip is a simple package to zip/unzip files, from zlib.'
-  homepage 'https://www.zlib.net/'
-  @_ver = '1.2.11'
+  homepage Zlibpkg.homepage
```
uberhacker, line 6: Not a valid url?

uberhacker, line 6: Not a valid url?

packages/minizip.rb
```diff
@@ -1,40 +1,34 @@
...
   # When upgrading minzip, be sure to upgrade zlibpkg in tandem.
-  version @_ver
-  license 'zlib'
+  version Zlibpkg.version
+  license Zlibpkg.license
   compatibility 'all'
-  source_url "https://www.zlib.net/zlib-#{@_ver}.tar.gz"
-  source_sha256 'c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1'
+  source_url Zlibpkg.source_url
+  source_sha256 Zlibpkg.source_sha256
```
uberhacker, line 13: Are these supposed to be replaceable parameters?

uberhacker (changes requested):
See comments.

packages/minizip.rb
saltedcoffii: what do you mean?

packages/minizip.rb
saltedcoffii: it is a valid url...

uberhacker: Doesn't resolve for me.  You are confusing me.

uberhacker: Doesn't resolve for me.  You are confusing me.

saltedcoffii: Oh! `homepage` is a property of the class `Zlibpkg` which i've required in the line `require_relative zlibpkg` at the top of the file. Because `zlibpkg` and `minizip` all use the same source, i've dried up the code a bit and made future updates easier by using class properties, e.g., chromebrew will interpret `Zlibpkg,homepage` as the `homepage` parameter in the `Zlibpkg` class. The same goes for `Zlibpkg.source_url` and `Zlibpkg.source_sha256`.

packages/minizip.rb
saltedcoffii: See my other comment 
https://github.com/chromebrew/chromebrew/pull/7557#discussion_r1006138811

packages/minizip.rb
uberhacker: Doesn't that overcomplicate things?  You still need to manage the versions in property values.

saltedcoffii: doesn't having to copy and paste things around overcomplicate things? also minizip and zlib use the same tarball so naturally the versions will be the same.

supechicken:
@saltedcoffii I see what you are trying to do. However, it is better to use `Package::Zlibpkg` to make it self-explanatory

packages/minizip.rb
supechicken: @uberhacker: @saltedcoffii is reusing the defined `homepage` and `source_{url,sha256}` from `zlibpkg.rb` (thats why `require_relative 'zlibpkg'` in line 2) and access it with the class const

packages/minizip.rb
```diff
@@ -1,40 +1,34 @@
 require 'package'
+require_relative 'zlibpkg'
```
supechicken, line 2: Use `Package.load_package()`? (a wrapper for loading package templates)

packages/minizip.rb
uberhacker: I gathered but what happens when we use `crew search zlibpkg -v`?  Is it going to display the correct homepage, version, etc. properties.  These are the kinds of things that have to be considered before willy-dilly doing something this drastic.

saltedcoffii: yes it does display the correct things.

uberhacker:
Would it make sense to combine these packages into one?

saltedcoffii:
No, `zlib` is a core library and `minizip` is just a cutesy little program that some people like having the binaries around for, or some programs require. mixing these two programs would just make `core` bigger.

packages/minizip.rb
satmandu: `Package.load_package('zlibpkg')` doesn't work instead of `require_relative 'zlibpkg'`...

satmandu (approved):

uberhacker (approved):

PR #7571, opened 2022-10-29
Bump all py3_* package versions

uberhacker:
In preparation for python 3.11.0 upgrade.


uberhacker:
Combined with python311 pr branch.

PR #7594, opened 2022-11-04
Torbrowser 11.5.1 => 11.5.7

uberhacker:
Tested on x86_64 with a resolution of 4800.

### DIFF
--- a/packages/torbrowser.rb
+++ b/packages/torbrowser.rb
@@ -3,23 +3,34 @@ require 'package'
 class Torbrowser < Package
   description "'The Onion Router' browser"
   homepage 'https://www.torproject.org/'
-  @_ver = '11.5.1'
+  @_ver = '11.5.7'
   version @_ver
   license 'BSD, custom, MPL-2.0 and MIT'
   compatibility 'x86_64'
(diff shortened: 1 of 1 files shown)


packages/torbrowser.rb
```diff
@@ -3,23 +3,34 @@
...
   depends_on 'gtk3'
   depends_on 'sommelier'
 
   def self.build
-    @tor = <<~EOF
-      #!/bin/sh -e
+    tor = <<~EOF
+      #!/bin/bash
+      SCALE=1
+      RESOLUTION=$(xdpyinfo | awk '/dimensions:/ { print $2 }' | cut -d'x' -f1)
```
satmandu, line 20: Nice. But do we want to add this sort of logic to sommelier?

satmandu (approved):

packages/torbrowser.rb
uberhacker: This logic is specific to gtk3 so maybe `/usr/local/etc/gtk-3.0/settings.ini`?

uberhacker: Actually, it might make sense to include it in the X11 config.

PR #7635, opened 2022-11-14
py3_codespell: Wrap variables in `self.preflight`

### LINKED ISSUE #7601: crew install issue
```
Setup and synchronize local package repo...

     -=O=-#     #       #         #                                                                                                                 
sh: 1: python3: not found
sh: 1: python3: not found
Installing core Chromebrew packages...
```
Looks like something is trying to execute python before it was installed.

supechicken:
Fix #7601 

Please define the variable inside the `def self.preflight` function next time if external commands are involved, otherwise the command will be triggered always (i.e when updating/removing)

### DIFF
--- a/packages/py3_codespell.rb
+++ b/packages/py3_codespell.rb
@@ -25,9 +25,11 @@ class Py3_codespell < Package
 
   depends_on 'py3_setuptools' => :build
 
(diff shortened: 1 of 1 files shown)


uberhacker:
I don't see how this fixes #7601.  We're still executing python3.

PR #7641, opened 2022-11-17
glibc: modify postinstall to make proper ld-linux symlinks

### LINKED ISSUE #7554: glibc: ldd executable broken
## Description:
The `ldd` shebang points to an executable that doesn't always exist,

## Output:
```
$ ldd
bash: /usr/local/bin/ldd: /usr/local/bin/bash: bad interpreter: No such file or directory

$ head -n 1 $(command -v ldd)
#! /usr/local/bin/bash
```

## Possible solutions:
- Change the shebang on ldd from `#! /usr/local/bin/bash` to `#! /usr/bin/env bash` (in the glibc binaries)
- Add fix in `def self.patch` and rebuild glibc
- make `bash` a dependency of `glibc` (therefore expanding core)

satmandu:
Fixes #7554 for glibc 2.33

- Fixes issue on `aarch64` with sommelier.
- Postinstall section adding symlinks also added to install section.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=glibc233 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/glibc.rb
+++ b/packages/glibc.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'package'
 
 class Glibc < Package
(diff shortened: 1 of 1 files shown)


packages/glibc.rb
```diff
@@ -590,6 +590,14 @@ def self.postinstall
...
       puts "System glibc version is #{LIBC_VERSION}.".lightblue
       puts 'Creating symlinks to system glibc version to prevent breakage.'.lightblue
+      case ARCH
+      when 'aarch64', 'armv7l'
+        FileUtils.ln_sf "/lib/ld-#{LIBC_VERSION}.so", 'ld-linux-armhf.so.3'
+      when 'i686'
+        FileUtils.ln_sf "/lib/ld-#{LIBC_VERSION}.so", 'ld-linux-i686.so.2'
+      when 'x86_64'
+        FileUtils.ln_sf "/lib64/ld-#{LIBC_VERSION}.so", 'ld-linux-x86-64.so.2'
+      end
```
uberhacker, line 600: Why are we doing this in postinstall instead of install so the symlinks can be deleted after the package is removed?

uberhacker (changes requested):
I think the postinstall section should be move to install?

packages/glibc.rb
satmandu: The postinstall handles situations where we don't have a binary package for a glibc build yet, same as the rest of the symlinks created in the postinstall.

satmandu:
We always need symlinks pointing to the _currently installed by google_ version of glibc.

packages/glibc.rb
uberhacker: I guess the next question is why don't we have binary packages yet?  I'm just thinking about potential dangling symlinks during upgrades and the like.

satmandu: Just added binaries.

uberhacker (approved):
I approved this reluctantly.  We should try to simplify in the future.

satmandu:
We're following Google and supporting EOL machines. This is going to stay complex. 😐

PR #7653, opened 2022-11-18
Update sommelier to 20221117

satmandu:
- Updated to the 20221117 version @supechicken noticed.
- Cleaned up output and error messages.

Works properly:
- [x] `x86_64`

Builds properly:
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=sommelier_new CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/sommelier.rb
+++ b/packages/sommelier.rb
@@ -16,10 +16,10 @@ class Sommelier < Package
      x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/sommelier/20221117-1_x86_64/sommelier-20221117-1-chromeos-x86_64.tar.zst'
   })
   binary_sha256({
-    aarch64: 'bebd8e71c33b4ad5e75addcf27c29c38ebf8c232c2ef0698c701a7984930d944',
-     armv7l: 'bebd8e71c33b4ad5e75addcf27c29c38ebf8c232c2ef0698c701a7984930d944',
-       i686: 'cc65f33b8cb9369dea7ef4aec0230e668fabe5ea0fa2daa032a49af8725696de',
-     x86_64: 'c8b0e73e7cfbe631dead10678204950e7f4f163985b8a3f5ca60c314ca48a6ca'
+    aarch64: '131485e6c60091254347fd906d729b1a3280cec40c40341459cad38caa574b1e',
+     armv7l: '131485e6c60091254347fd906d729b1a3280cec40c40341459cad38caa574b1e',
+       i686: '541474d7ab640e71990a6ac3dfe35fa28774b47af337e38757a7913c619be3b4',
+     x86_64: '5e553206ed88581d3948dd6920ab42331e80100976016d59b37bc5a66c444a9c'
   })
 
   depends_on 'libdrm'
@@ -187,17 +187,6 @@ class Sommelier < Package
         File.write 'sommelierd', <<~SOMMELIERDEOF
           #!/bin/bash -a
 
-          source ${CREW_PREFIX}/etc/env.d/sommelier.env &>/dev/null
-          set +a
-          mkdir -p #{CREW_PREFIX}/var/{log,run}
-          checksommelierwayland () {
-            [[ -f "#{CREW_PREFIX}/var/run/sommelier-wayland.pid" ]] || return 1
-            /sbin/ss --unix -a -p | grep "\b$(cat #{CREW_PREFIX}/var/run/sommelier-wayland.pid)" | grep wayland &>/dev/null
-          }
-          checksommelierxwayland () {
-            DISPLAY="${DISPLAY}" timeout 1s xset q &>/dev/null
-          }
-
           # get a list of all available DRM render nodes
           DRM_DEVICES_LIST=( /sys/class/drm/renderD* )
 
@@ -217,6 +206,17 @@ class Sommelier < Package
             # if only one node available, use it directly
             SOMMELIER_DRM_DEVICE="/dev/dri/${DRM_DEVICES_LIST[0]##*/}"
           fi
+
+          source ${CREW_PREFIX}/etc/env.d/sommelier.env &>/dev/null
+          set +a
+          mkdir -p #{CREW_PREFIX}/var/{log,run}
+          checksommelierwayland () {
+            [[ -f "#{CREW_PREFIX}/var/run/sommelier-wayland.pid" ]] || return 1
+            /sbin/ss --unix -a -p | grep "\b$(cat #{CREW_PREFIX}/var/run/sommelier-wayland.pid)" | grep wayland &>/dev/null
+          }
+          checksommelierxwayland () {
+            DISPLAY="${DISPLAY}" timeout 1s xset q &>/dev/null
+          }
 
           ## As per https://www.reddit.com/r/chromeos/comments/8r5pvh/crouton_sommelier_openjdk_and_oracle_sql/e0pfknx/
           ## One needs a second sommelier instance for wayland clients since at some point wl-drm was not implemented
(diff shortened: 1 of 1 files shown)


packages/sommelier.rb
```diff
@@ -75,10 +75,9 @@ def self.build
           -Db_asneeded=false \
           -Db_lto=true \
           -Db_lto_mode=thin \
+          -Dwith_tests=false \
           -Dxwayland_path=#{CREW_PREFIX}/bin/Xwayland \
           -Dxwayland_gl_driver_path=/usr/#{ARCH_LIB}/dri -Ddefault_library=both \
```
supechicken, line 80: Use the `crew` version of `mesa`? (some ARM Chromebooks (like mine) do not come across with the `mesa` drivers preinstalled, all drivers are merged into the chrome binary instead)

packages/sommelier.rb
```diff
@@ -200,15 +199,17 @@ def self.build
...
 
             sommelier -X \
-              --x-display=${DISPLAY}  \
-              --scale=${SCALE} --glamor \
-              --drm-device=/dev/dri/renderD128 \
-              --virtwl-device=/dev/null \
-              --shm-driver=noop --data-driver=noop \
+              --x-display=${DISPLAY} \
+              --scale=${SCALE} \
+              --direct-scale \
```
supechicken, line 204: Notice the `--direct-scale` option only works on recent version of Chrome OS:
```
# on my ARM chromebook, chrome 84
sommelier.elf: ../sommelier-output.cc:180: void sl_output_init_dimensions_direct(struct sl_host_output *, int *, int *, int *, int *, int *): Assertion `host->ctx->xdg_output_manager' failed.
```

supechicken:
Works on `x86_64` (`qemu+virgl` and intel hardware), **with `/dev/dri/renderD129` as DRM device**:

Failed on my ARM Chromebook, this patch I figured out last week fixed it:
```
sommelier.elf: ../sommelier-window.cc:412: void sl_window_update(struct sl_window *): Assertion `ctx->xdg_shell' failed.
(EE) failed to write to Xwayland fd: Broken pipe
Aborted (core dumped)
```

```patch
diff -Nur a/sommelier.cc b/sommelier.cc
--- a/sommelier.cc	2022-11-06 19:29:16.580361574 +0800
+++ b/sommelier.cc	2022-11-06 19:37:28.830367176 +0800
@@ -616,7 +616,7 @@
       data_device_manager->host_global =
           sl_data_device_manager_global_create(ctx);
     }
-  } else if (strcmp(interface, "xdg_wm_base") == 0) {
+  } else if ((strcmp(interface, "xdg_wm_base") == 0) || (strcmp(interface, "zxdg_shell_v6") == 0)) {
     struct sl_xdg_shell* xdg_shell =
         static_cast<sl_xdg_shell*>(malloc(sizeof(struct sl_xdg_shell)));
     assert(xdg_shell);
```

supechicken:
Move changes to `crew-package-sommelier` for better maintenance?

satmandu:
@supechicken do you want to hack this up to get it to the finish line?

packages/sommelier.rb
satmandu: Do you know which ChromeOS Milestone this starts working with? And is that a fatal error on the older versions?
Any ideas for a milestone version check we can use in the sommelierd script to select whether to use `--direct-scale`?

packages/sommelier.rb
satmandu: Ah yes, I suppose that makes sense!

satmandu:
Rebuilt this. @supechicken @uberhacker Would you please see if this works now on arm for you?

uberhacker (changes requested):
Please remove the `--direct-scale` option.  I tested this on my armv7l system (stuck at release 84 like @supechicken) and it started working with that one change.

packages/sommelier.rb
```diff
@@ -283,7 +308,7 @@ def self.build
           }
           if ! checksommelierwayland || ! checksommelierxwayland ; then
             [ -f  #{CREW_PREFIX}/bin/stopbroadway ] && stopbroadway
-            #{CREW_PREFIX}/sbin/sommelierd &
+            #{CREW_PREFIX}/sbin/sommelierd &>/dev/null
```
uberhacker, line 311: What is wrong with seeing the output?

uberhacker (commented):
Do we need new binaries?

packages/sommelier.rb
satmandu: The output is already being put into a log file from inside `sommelierd`

satmandu:
```
            sommelier -X \
              --x-display=${DISPLAY} \
              --scale=${SCALE} \
              ${SOMMELIER_DIRECT_SCALE} \
              --glamor \
              --force-drm-device=${SOMMELIER_DRM_DEVICE} \
              --display=wayland-0 \
              --xwayland-path=/usr/local/bin/Xwayland \
              --xwayland-gl-driver-path=#{CREW_LIB_PREFIX}/dri \
              --peer-cmd-prefix="#{CREW_PREFIX}#{@peer_cmd_prefix}" \
              --enable-xshape \
              --noop-driver \
              --no-exit-with-child \
              /bin/sh -c "touch ~/.Xauthority
                xauth -f ~/.Xauthority add ${DISPLAY} . $(xxd -l 16 -p /dev/urandom)
                source #{CREW_PREFIX}/etc/sommelierrc" \
            &>> #{CREW_PREFIX}/var/log/sommelier.log
```

packages/sommelier.rb
satmandu: ok I inserted code to handle this.

satmandu:
Rebuilt and added new binaries.

Please check with 
```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=sommelier_new CREW_TESTING=1 crew update ; crew reinstall sommelier
```

packages/sommelier.rb
uberhacker: Gotcha.

uberhacker (approved):

myelin:
Nice spotting... I was wondering if I should ping someone at Chromebrew to see if you were interested in switching to the newer Sommelier, but it looks like @supechicken found it almost immediately after it landed.

I'm curious about the zxdg_shell_v6 change; it looks like support for this was dropped in https://crrev.com/c/3205214 in Oct 2021, because it broke apps on Debian Stretch.  @supechicken what software were you running that caused Sommelier to crash with the ctx->xdg_shell assertion failure?

supechicken:
I found your commit while tracing which commit breaks the noop driver patch (due to the new `/*is_drm=*/true` option in `sl_create_host_buffer()`). Anyways thank you for re-adding the noop driver to sommelier :)

Without modifying the `strcmp(interface, "xdg_wm_base")` line, every X program that requires opening a window triggers the `assert(ctx->xdg_shell)` line (commands that do not require creating an X window like `xdpyinfo` and `glxinfo -B` works well):
```
$ sommelier -X --noop-driver glxgears
sommelier.elf: ../sommelier-window.cc:412: void sl_window_update(struct sl_window *): Assertion `ctx->xdg_shell' failed.
(EE) failed to write to Xwayland fd: Broken pipe
Aborted (core dumped)
```

I tried adding `printf("%s", interface);` before the if statement [in `sommelier.cc`, line 619](https://chromium.googlesource.com/chromiumos/platform2/+/main/vm_tools/sommelier/sommelier.cc#619), it prints `zxdg_shell_v6` instead of the v7 `xdg_wm_base`

It seems that the exo server on CrOS v84 was compiled before https://crrev.com/c/2237428 was merged, and thus using the older v6 interface (surprisingly sommelier works without replacing the new `xdg_shell.xml` with the old v6_unstable one)

satmandu:
Honestly the other lingering issue with the exo server is that `wl_compositor` version is still `3` instead of `4`. This is continuing to be an issue with building software which can run directly w/o sommelier. e.g. [mpv](https://github.com/mpv-player/mpv/issues/10868#issuecomment-1314584031).

PR #7678, opened 2022-11-25
Add dvtm package

uberhacker:
dvtm brings the concept of tiling window management, popularized by X11-window managers like dwm to the console. As a console window manager it tries to make it easy to work with multiple console based programs.  See https://www.brain-dump.org/projects/dvtm/.  Tested on all architectures.

### DIFF
--- a/packages/dvtm.rb
+++ b/packages/dvtm.rb
@@ -1,0 +1,37 @@
+require 'package'
+
+class Dvtm < Package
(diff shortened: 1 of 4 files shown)


satmandu (approved):

satmandu:
Complaint about single quoting strings...

uberhacker:
I switched to draft to do more testing.

PR #7708, opened 2022-12-02
initial attempt at functionalizing replaces in python3 package

satmandu:
- Adds generalizable code for removing packages in preinstall. (This should move into crew at some point.)

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=python3_tweak  CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/python3.rb
+++ b/packages/python3.rb
@@ -49,11 +49,15 @@ class Python3 < Package
   def self.preinstall
     @device = JSON.load_file("#{CREW_CONFIG_PATH}/device.json", symbolize_names: true)
     @replaces = %w[py3_pip py3_setuptools]
+    @replaces_installed = []
     @replaces.each do |package|
       if @device[:installed_packages].any? { |elem| elem[:name] == package }
-        puts "Removing superseded package #{package}...".orange
-        system "crew remove #{package}", exception: false
+        @replaces_installed.push(package)
       end
+    end
+    unless @replaces_installed.empty?
+      puts "Removing superseded package(s): #{@replaces_installed.join(' ')}...".orange
+      system "crew remove #{@replaces_installed.join(' ')}", exception: false
     end
   end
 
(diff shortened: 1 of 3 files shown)


packages/python3.rb
```diff
@@ -47,7 +47,14 @@ class Python3 < Package
   conflicts_ok
 
   def self.preinstall
-    system 'crew remove py3_setuptools py3_pip', exception: false
+    @device = JSON.load_file("#{CREW_CONFIG_PATH}/device.json", symbolize_names: true)
+    @replaces = %w[py3_pip py3_setuptools]
+    @replaces.each do |package|
+      if @device[:installed_packages].any? { |elem| elem[:name] == package }
+        puts "Removing superseded package #{package}...".orange
+        system "crew remove #{package}", exception: false
```
uberhacker, line 55: Does this work?  It would be recursive crew commands.

satmandu: It works!

uberhacker: Then you don't need to loop over each package.  This would work too:
```
system "crew remove pip3_pip py3_setuptools", exception: false
```

uberhacker: So does this really work or are there issues with nesting crew commands?

satmandu:
It's only partially functional. It effectively removes the package files,
but not the package from the installed section of the device.json.

packages/python3.rb
satmandu: The nested solution now works with a `crew` change to reload `device.json` after preinstall finishes.

satmandu: Also modified the code to run `crew remove` only once.

uberhacker (approved):

PR #7725, opened 2022-12-08
crew: Add support for dependency version ranges

supechicken:
This PR adds the functionality to check whether the installed version of a dependency fulfills the requirement specified by packages

To use this feature, just add a range after the name of dependency
```ruby
depends_on 'python3', '>= 3.12.0' # this version doesn't exist at this moment :)
```

When the installed version of the specified dependency doesn't fulfill the range given, an error message will be thrown:
```
Package py3_setuptools depends on 'python3' (>= 3.12.0), however version '3.11.0' is currently installed :/

Run `crew update && crew upgrade` and try again?
```

Tested on `x86_64`

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1357,19 +1357,19 @@ def resolve_dependencies
 
   # compare dependency version with required range (if installed)
   @dependencies.each do |dep|
-    next unless version_check
-
     depName  = dep.keys[0]
-    dep_info = @device[:install_package].select {|pkg| pkg[:name] == depName }
+    dep_info = @device[:installed_packages].select {|pkg| pkg[:name] == depName } [0]
 
     # skip if dependency is not installed
     next if dep_info.empty?
 
-    tags, version_check = deps.values[0]
+    tags, version_check = dep.values[0]
     installed_version   = dep_info[:version]
 
+    next unless dep.version_check
+
     # abort if the range is not fulfilled
-    abort unless version_check.call(installed_version)
+    abort unless dep.version_check.call(installed_version)
   end
 
   # leave only dependency names (remove all package attributes returned by @pkg.get_deps_list)
--- a/bin/crew
+++ b/bin/crew
@@ -1357,6 +1357,8 @@ def resolve_dependencies
 
   # compare dependency version with required range (if installed)
   @dependencies.each do |dep|
+    next unless version_check
+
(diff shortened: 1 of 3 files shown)


lib/package.rb
```diff
@@ -87,11 +88,11 @@ def self.get_deps_list(pkgName = name, hash: false, include_build_deps: 'auto',
        !exclude_buildessential && \
        !@checked_list.keys.include?('buildessential')
 
-      deps = { 'buildessential' => [:build] }.merge(deps)
+      deps = { 'buildessential' => [[:build]] }.merge(deps)
     end
 
     # parse dependencies recursively
-    expandedDeps = deps.uniq.map do |dep, depTags|
+    expandedDeps = deps.uniq.map do |dep, (depTags, ver_check)|
```
uberhacker, line 95: We should decide if we're using camelCase or not and be consistent.

bin/crew
```diff
@@ -1349,16 +1349,31 @@ def resolve_dependencies_and_install
...
+    # abort if the range is not fulfilled
+    abort unless version_check.call(installed_version)
+  end
+
+  # leave only dependency names (remove all package attributes returned by @pkg.get_deps_list)
+  @dependencies.map!(&:keys).flatten!
 
   # leave only not installed packages in dependencies
-  @dependencies.reject! { |name| @device[:installed_packages].any? { |pkg| pkg[:name] == name } }
+  @dependencies.reject! { |depName| @device[:installed_packages].any? { |pkg| pkg[:name] == depName } }
```
uberhacker, line 1376: Bump version in const.rb?

uberhacker (changes requested):
See comments.

lib/package.rb
supechicken: Change made

bin/crew
supechicken: Done

satmandu (approved):

uberhacker (approved):

Zopolis4:
What is the syntax for declaring version ranges on a build dependency?
(i.e. has a build dependency on a python version equal to 3.12)

satmandu:
Wasn't this reverted?

Zopolis4:
The code's still in `lib/package.rb`, so I don't think so...

PR #7558, opened 2022-10-26
go => 1.19.2

saltedcoffii:
## Description
update go to latest version

## Additional information

Works properly:
- [ ] `x86_64`
- [ ] `i686`
- [ ] `armv7l` 

### Run the following to get this pull request's changes locally for testing.
```
CREW_TESTING_REPO=https://github.com/saltedcoffii/chromebrew.git CREW_TESTING_BRANCH=go_1.19.2 CREW_TESTING=1 crew update
```

PR #7741, opened 2022-12-15
Go 1.18 => 1.19.4

uberhacker:
Why build from source when binaries are available?  Tested on all architectures.  Fixes #7558.


satmandu:
The armv6l binaries work fine on armv7l? Does armv6l support hard float?

uberhacker:
Yes, as I mentioned it was tested.

satmandu:
> Yes, as I mentioned it was tested. Not sure about hard float. What is that?

hardware floating point support...

satmandu (approved):

PR #7751, opened 2022-12-17
crew: Use `rename()` syscall to install files from destdir for faster speed

supechicken:
Currently, we use `rsync` to copy files from `CREW_DEST_DIR` to `CREW_PREFIX` and delete them after copying (`--remove-source-files`)

Sounds like `mv src/* dst` is more suitable for this task (`rsync` need to copy all file contents under the source and remove the source while `mv` only needs to update the path stored in the filesystem table, which is definitely faster) (see [here](https://unix.stackexchange.com/a/454404) for further explained)

However, neither the `mv` command nor `FileUtils.mv` supports overriding existing files/directories in the destination, errors will be returned when a file/directory exists in both `src` and `dst`

So I wrote a C program for `mv` with file overriding support (so that it can be used in `install.sh`), and it is **way** faster than `rsync` without a doubt.

Here is its performance (compared with `tar` and `rsync`) in `crew install llvm`:
```
> time rsync -ahHAXW --remove-source-files ./usr/local/ /usr/local/

real    0m1.874s
user    0m0.419s
sys     0m1.940s
Performing post-install for llvm...


> time { cd ./usr/local/; tar -cf - ./* | (cd /usr/local/; tar -xp --keep-directory-symlink -f -); }

real    0m1.572s
user    0m0.104s
sys     0m1.835s
Performing post-install for llvm...


> crew-mvdir  ./usr/local/ /usr/local/

real    0m0.087s
user    0m0.003s
sys     0m0.084s
Performing post-install for llvm...
```

### Changes
- New file: `src/mvdir.c`
- New package: `crew_mvdir`
- Cleanup

Tested on `x86_64`

```
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=use_rename CREW_TESTING=1 crew update
```


bin/crew
```diff
@@ -1279,38 +1289,17 @@ def install_package(pkgdir)
...
-    if Dir.exist? "#{pkgdir}/#{CREW_PREFIX}"
-      if rsync_available
-        # Adjust "./usr/local" if the build CREW_PREFIX ever changes.
-        system "rsync -ahHAXWx --remove-source-files ./#{CREW_PREFIX}/ #{CREW_PREFIX}"
-      else
-        system "cp -a#{@verbose} ./#{CREW_PREFIX}/* #{CREW_PREFIX}"
-      end
-    end
+    # Adjust "./usr/local" if the build CREW_PREFIX ever changes.
+    install_files('./usr/local', CREW_PREFIX) if Dir.exist?('./usr/local') && CREW_PREFIX != '/usr/local'
```
uberhacker, line 1302: What is wrong with simply running `install_files(CREW_PREFIX)`?

packages/crew_mvdir.rb
```diff
@@ -0,0 +1,23 @@
...
+
+class Crew_mvdir < Package
+  description 'Alternative for "rsync --remove-source-files dir1/ dir2/", with faster speed'
+  homepage 'https://chromebrew.github.io'
+  version '0.1'
+  compatibility 'all'
+  license 'GPL-3'
+
+  source_url 'https://github.com/supechicken/chromebrew/raw/use_rename/src/mvdir.c'
+  source_sha256 'SKIP'
```
uberhacker, line 11: Why not use `source_sha256`?

packages/crew_mvdir.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Crew_mvdir < Package
+  description 'Alternative for "rsync --remove-source-files dir1/ dir2/", with faster speed'
+  homepage 'https://chromebrew.github.io'
+  version '0.1'
+  compatibility 'all'
+  license 'GPL-3'
+
+  source_url 'https://github.com/supechicken/chromebrew/raw/use_rename/src/mvdir.c'
```
uberhacker, line 10: Should this be versioned and use `https://github.com/chromebrew` repo?

packages/crew_mvdir.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  no_shrink # upx: CantPackException: bad DT_GNU_HASH (1+ max_bucket)=0x1 < symbias=0x14
+
+  def self.build
+    system "mold -run cc #{CREW_COMMON_FLAGS} mvdir.c -o crew-mvdir"
+  end
+
+  def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
+    FileUtils.install 'crew-mvdir', "#{CREW_DEST_PREFIX}/bin/crew-mvdir", mode: 0o755
```
uberhacker, line 21: How does this compare with the ChromeOS `mv` command?  Seems like reinventing the wheel?

packages/crew_mvdir.rb
supechicken: Checksum added

packages/crew_mvdir.rb
supechicken: Now it is pointed to crew repo, however I am still using `master` since we cannot get the commit hash before this pr merge

packages/crew_mvdir.rb
supechicken: One of the biggest disadvantages of `mv -f` is it cannot deal with non-empty directories:
```
$ /usr/local/bin/tree src dst
src
└── test
    └── file
dst
└── test
    └── file2

2 directories, 2 files

$ mv -f src/* dst
mv: cannot move 'src/test' to 'dst/test': Directory not empty
```

bin/crew
supechicken: This line is used to move `/usr/local` to custom `CREW_PREFIX` in case `CREW_PREFIX != /usr/local` (inherited from the original logic)

uberhacker: Why don't we just install directly in CREW_PREFIX instead of moving things around?

supechicken: As this part also handles copying prebuilt binaries after moving out `filelist/directorylist` (which assume CREW_PREFIX is `/usr/local` when packaging)

uberhacker: Why make that assumption at all?  Just extract directly to CREW_PREFIX.

uberhacker: Oh, so you are saying the `usr/local` path is in the archive?  Why don't we build binaries relative to `CREW_PREFIX` then?

supechicken: Let's make changes in the another pr?

satmandu:
@supechicken Is this good to go?

supechicken:
> @supechicken Is this good to go?

Yep :) And we need to adjust the `git_hashtag` and prebuilt binary after this pr merged

lib/const.rb
```diff
@@ -136,6 +136,9 @@
                      ENV.fetch('CHROMEOS_RELEASE_CHROME_MILESTONE', nil)
                    end
 
+# If CREW_DISABLE_MVDIR environment variable exists use rsync/tar to install files in lieu of crew-mvdir.
+CREW_DISABLE_MVDIR = ENV['CREW_DISABLE_MVDIR'].eql?('1')
+
 # If CREW_USE_CURL environment variable exists use curl in lieu of net/http.
 CREW_USE_CURL = ENV['CREW_USE_CURL'].eql?('1')
```
uberhacker, line 143: Bump crew version?

satmandu: Done

satmandu:
So we need binaries built?

uberhacker:
Rubocop is sounding the alarm. ;)

satmandu:
@supechicken Would you prefer to have a separate repo for the mvdir.c? We need to have the package built before we do testing... and that is easier if the source file is in another repo, like the other chromebew repos which are used for other parts of packages...

satmandu:
(in which case we would close https://github.com/chromebrew/chromebrew/pull/7774 )

satmandu:
> Rubocop is sounding the alarm. ;)

Shellcheck this time. ;)

supechicken:
> @supechicken Would you prefer to have a separate repo for the mvdir.c? We need to have the package built before we do testing... and that is easier if the source file is in another repo, like the other chromebew repos which are used for other parts of packages...

But we have only single file for this😅

satmandu:
> > @supechicken Would you prefer to have a separate repo for the mvdir.c? We need to have the package built before we do testing... and that is easier if the source file is in another repo, like the other chromebew repos which are used for other parts of packages...
> 
> But we have only single file for this😅

Yes but it is ancillary to the project. It should be in its own repo so you can tag releases...

supechicken:
@satmandu See [chromebrew/crew-mvdir](https://github.com/chromebrew/crew-mvdir)

satmandu:
> @satmandu See [chromebrew/crew-mvdir](https://github.com/chromebrew/crew-mvdir)

That works!

uberhacker:
@supechicken: Does it make sense to build binary releases at https://github.com/chromebrew/crew-mvdir?

packages/crew_mvdir.rb
```diff
@@ -0,0 +1,23 @@
+require 'package'
+
+class Crew_mvdir < Package
+  description 'Alternative for "rsync --remove-source-files dir1/ dir2/", with faster speed'
+  homepage 'https://chromebrew.github.io'
```
uberhacker, line 5: `homepage 'https://github.com/chromebrew/crew-mvdir'`?

supechicken:
> @supechicken: Does it make sense to build binary releases at https://github.com/chromebrew/crew-mvdir?

Can we still include this as a bootstrap package in `install.sh` without `binary_url`? (or `binary_url` pointing to github release instead of gitlab package registry)

bin/crew
Zopolis4: Let's make that PR, though, as this breaks installing on non `/usr/local` `CREW_PREFIX`'s.

Both on my refactor and the original.

satmandu:
Are we closer to using this?

Zopolis4:
> Yes, it can be a github binary_url but don't you want to include the source_url in the package too?

Can we please not use alternative providers for binary urls? That'll break #7082.

supechicken:
> Can we please not use alternative providers for binary urls? That'll break https://github.com/chromebrew/chromebrew/pull/7082.

I am okay with both options...

FYI binaries are pushed to https://github.com/chromebrew/crew-mvdir/releases/tag/0.1

uberhacker:
> > Can we please not use alternative providers for binary urls? That'll break #7082.
> 
> I am okay with both options (gitlab package or github release)
> 
> FYI binaries are pushed to https://github.com/chromebrew/crew-mvdir/releases/tag/0.1

Yeah, this is no different than any other project 

> > Can we please not use alternative providers for binary urls? That'll break #7082.
> 
> I am okay with both options (gitlab package or github release)
> 
> FYI binaries are pushed to https://github.com/chromebrew/crew-mvdir/releases/tag/0.1

Yeah, this is no different than any other project on GitHub.  We'll just have to include it as a core package since the executable is used in crew.

packages/crew_mvdir.rb
supechicken: Change made

satmandu:
Please rebased to master and give me a CREW_TESTING command so this can be tested in a container...

supechicken:
```
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=use_rename CREW_TESTING=1 crew update
```

satmandu:
The fallback here ensures nothing breaks if `crew_mvdir` isn't installed, but is it a bug that there is no prompt to install `crew_mvdir` despite core having had `crew_mvdir` added to it?

Should part of this PR be making sure that `crew_mvdir` gets installed for updates and not just fresh installs?

supechicken:
> The fallback here ensures nothing breaks if `crew_mvdir` isn't installed, but is it a bug that there is no prompt to install `crew_mvdir` despite core having had `crew_mvdir` added to it?
> 
> Should part of this PR be making sure that `crew_mvdir` gets installed for updates and not just fresh installs?

Added a prompt for it

bin/crew
```diff
@@ -1271,10 +1271,14 @@ end
 def install_files(src, dst = src.delete_prefix('.'))
   if File.executable?("#{CREW_PREFIX}/bin/crew-mvdir") && !CREW_DISABLE_MVDIR
     system "crew-mvdir #{@short_verbose} #{src} #{dst}"
-  elsif File.executable?("#{CREW_PREFIX}/bin/rsync")
-    system "rsync -ah#{@verbose}HAXW --remove-source-files #{src}/ #{dst}"
   else
-    system "cd #{src}; tar -cf - ./* | (cd #{dst}; tar -x#{@verbose}p --keep-directory-symlink -f -)"
+    warn 'crew-mvdir is not installed. Please install it with \'crew install crew_mvdir\''.yellow
```
satmandu, line 1275: @uberhacker Is this reasonable or do you think there needs to be more explanation?

uberhacker: You could mention that installing crew-mvdir can help improve performance?

satmandu: Maybe this would work:
```
warn 'crew-mvdir is not installed. Please install it with \'crew install crew_mvdir\' for increased installation performance'.yellow
```

satmandu (approved):

bin/crew
uberhacker: This line is no longer necessary if we adjust the `dst` parameter as commented above?

bin/crew
```diff
@@ -1268,6 +1268,20 @@ def shrink_dir(dir)
   end
 end
 
+def install_files(src, dst = src.delete_prefix('.'))
```
uberhacker, line 1271: How about?
```
def install_files(src, dst = File.join(CREW_PREFIX, src.delete_prefix('./usr/local')))
```
That way we can pass any destination directory from the binary and it automagically installs in the correct location.

bin/crew
```diff
@@ -1300,33 +1314,11 @@ def install_package(pkgdir)
       end
     end
 
-    # check if the available rsync command support ACLs
+    install_files(File.join('.', HOME)) if Dir.exist?(File.join('.', HOME))
```
uberhacker, line 1317:
```
install_files(File.join('.', HOME), HOME) if Dir.exist?(File.join('.', HOME))
```
?

bin/crew
uberhacker: I'm thinking lines 1318 and 1321 can be one call as below?
```
install_files('./usr/local')
```
This is assuming the `dst` parameter is changed as suggested above in the `install_files` function.

bin/crew
uberhacker: Actually, why don't we put the `Dir.exist?` logic inside the `install_files` function?  Something like:
```
if Dir.exist?(src)
  ...
else
  puts "#{src} directory does not exist.".lightred
end
```

bin/crew
supechicken: This will not work for `install_files(CREW_DEST_HOME, HOME)`

uberhacker: Right, because source and destination are reversed?

uberhacker: Right because it needs to be `install_files(File.join('.', CREW_DEST_HOME), HOME)`?

uberhacker: The `HOME` parameter should override the `dst` default.

bin/crew
```diff
@@ -1273,7 +1273,7 @@ def install_files(prefix, path, dst = File.join( CREW_PREFIX, src.delete_prefix(
   return false unless Dir.exist?(src)
 
   if File.executable?("#{CREW_PREFIX}/bin/crew-mvdir") && !CREW_DISABLE_MVDIR
-    system 'crew-mvdir', @short_verbose, src, dst, exception: true
+    system 'crew-mvdir', *(@verbose ? [@short_verbose, src] : src), dst, exception: true
```
uberhacker, line 1276: Why the `*`?

satmandu: I think `*` is meant to collect the arguments as a list.

satmandu:
I'm getting this error at the end of an install:
```
/usr/local/bin/crew:1271:in `install_files': undefined local variable or method `src' for main:Object (NameError)

def install_files(prefix, path, dst = File.join( CREW_PREFIX, src.delete_prefix(".#{CREW_PREFIX}") ))
                                                              ^^^
        from /usr/local/bin/crew:1320:in `block in install_package'
        from /usr/local/bin/crew:1289:in `chdir'
        from /usr/local/bin/crew:1289:in `install_package'
        from /usr/local/bin/crew:1474:in `install'
        from /usr/local/bin/crew:1344:in `resolve_dependencies_and_install'
        from /usr/local/bin/crew:1787:in `block in install_command'
        from /usr/local/bin/crew:1782:in `each'
        from /usr/local/bin/crew:1782:in `install_command'
        from /usr/local/bin/crew:1914:in `<main>'
```

bin/crew
```diff
@@ -1268,6 +1268,23 @@ def shrink_dir(dir)
   end
 end
 
+def install_files(prefix, path, dst = File.join( CREW_PREFIX, src.delete_prefix(".#{CREW_PREFIX}") ))
+  src = File.join(prefix, path)
```
satmandu, line 1272: You are declaring src inside the function, so you can't use it in the argument.

However, this doesn't work:
```
def install_files(prefix, path, dst = File.join( CREW_PREFIX, (File.join(prefix, path)).delete_prefix(".#{CREW_PREFIX}") ))
  src = File.join(prefix, path)
  puts "dst is #{dst} and src is #{src}".lightblue
  return false unless Dir.exist?(src)
```

I get this: 
image

bin/crew
satmandu: In any case it doesn't work...

satmandu: This works: ` system "crew-mvdir #{@short_verbose} #{src} #{dst}",exception: true`

But worth noting that @verbose and @short_verbose are undefined in this function no matter whether `-v` is passed to install or not...

bin/crew
satmandu: This works:
`def install_files(prefix, path, dst = File.join( CREW_PREFIX, (File.join(prefix, path)).delete_prefix(".#{CREW_PREFIX}") ))`

bin/crew
```diff
@@ -1268,6 +1268,23 @@ def shrink_dir(dir)
...
+  return false unless Dir.exist?(src)
+
+  if File.executable?("#{CREW_PREFIX}/bin/crew-mvdir") && !CREW_DISABLE_MVDIR
+    system 'crew-mvdir', *(@verbose ? [@short_verbose, src] : src), dst, exception: true
+  else
+    warn 'crew-mvdir is not installed. Please install it with \'crew install crew_mvdir\' for improved installation performance'.yellow
+
+    if File.executable?("#{CREW_PREFIX}/bin/rsync") && system("#{CREW_PREFIX}/bin/rsync --version > /dev/null")
+      system 'rsync', "-ah#{@verbose}HAXW", '--remove-source-files', src, dst, exception: true
```
satmandu, line 1281: You need this before the rsync line:
```
 # rsync src path needs a trailing slash
      src << '/' unless src.end_with?('/')
```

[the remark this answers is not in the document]
satmandu: Fixed....

bin/crew
satmandu: The existing setup is better, because the code block is used for BOTH `CREW_PREFIX` and `HOME` and there may be no `HOME` in the binary archive we are installing. There's no need to complain if that's the case.

satmandu:
> I'm getting this error at the end of an install:
> 
> ```
> /usr/local/bin/crew:1271:in `install_files': undefined local variable or method `src' for main:Object (NameError)
> 
> def install_files(prefix, path, dst = File.join( CREW_PREFIX, src.delete_prefix(".#{CREW_PREFIX}") ))
>                                                               ^^^
>         from /usr/local/bin/crew:1320:in `block in install_package'
>         from /usr/local/bin/crew:1289:in `chdir'
>         from /usr/local/bin/crew:1289:in `install_package'
>         from /usr/local/bin/crew:1474:in `install'
>         from /usr/local/bin/crew:1344:in `resolve_dependencies_and_install'
>         from /usr/local/bin/crew:1787:in `block in install_command'
>         from /usr/local/bin/crew:1782:in `each'
>         from /usr/local/bin/crew:1782:in `install_command'
>         from /usr/local/bin/crew:1914:in `<main>'
> ```

I resolved this...

bin/crew
satmandu: I fixed the `@verbose` issue.

satmandu (approved):

bin/crew
```diff
@@ -1268,6 +1268,25 @@ def shrink_dir(dir)
   end
 end
 
+def install_files(prefix, path, dst = File.join( CREW_PREFIX, File.join(prefix, path).delete_prefix(".#{CREW_PREFIX}") ))
```
uberhacker, line 1271: Change `".#{CREW_PREFIX}"` to `"./usr/local"`?  This will be needed if `CREW_PREFIX != '/usr/local'`?

satmandu: I only fixed the code, without commenting on the logic. 🫣

uberhacker: @supechicken?

satmandu: @supechicken can you explain the logic here?

bin/crew
```diff
@@ -1300,33 +1319,11 @@ def install_package(pkgdir)
...
-    if Dir.exist? "#{pkgdir}/#{CREW_PREFIX}"
-      if rsync_available
-        # Adjust "./usr/local" if the build CREW_PREFIX ever changes.
-        system "rsync -ahHAXWx --remove-source-files ./#{CREW_PREFIX}/ #{CREW_PREFIX}"
-      else
-        system "cp -a#{@verbose} ./#{CREW_PREFIX}/* #{CREW_PREFIX}"
-      end
-    end
+    # Adjust "./usr/local" if the build CREW_PREFIX ever changes.
+    install_files('.', '/usr/local', CREW_PREFIX) unless CREW_PREFIX == '/usr/local'
```
uberhacker, line 1326: Not needed?

bin/crew
uberhacker: `prefix` parameter not needed?  should only need src and dst?

bin/crew
```diff
@@ -1268,6 +1268,25 @@ def shrink_dir(dir)
   end
 end
 
+def install_files(prefix, path, dst = File.join( CREW_PREFIX, File.join(prefix, path).delete_prefix(".#{CREW_PREFIX}") ))
+  src = File.join(prefix, path)
```
uberhacker, line 1272: Not needed if `prefix` parameter removed?

bin/crew
```diff
@@ -1268,6 +1268,25 @@ def shrink_dir(dir)
...
+  return false unless Dir.exist?(src)
```
uberhacker, line 1273: Return with error message rather than silent confusion?

packages/crew_mvdir.rb
```diff
@@ -0,0 +1,35 @@
+require 'package'
+
+class Crew_mvdir < Package
+  description 'Alternative for "rsync --remove-source-files dir1/ dir2/", with faster speed'
```
uberhacker, line 4: `Faster alternative for "rsync --remove-source-files dir1/ dir2/"`?

satmandu: Added "Faster" to description.

packages/crew_mvdir.rb
```diff
@@ -0,0 +1,35 @@
+require 'package'
+
+class Crew_mvdir < Package
+  description 'Faster alternative to "rsync --remove-source-files dir1/ dir2/", with faster speed'
```
uberhacker, line 4: Don't need `, with faster speed`?

satmandu: Fixed. 🫣

bin/crew
uberhacker: Adding the logic inside the function is more elegant because currently there is no way to validate the directory unless the conditional is outside the function and we definitely want to throw an error message for an invalid directory rather than the kludgy error we'll get with the first command that attempts to operate on it.

uberhacker: But I see what you are saying about `HOME`.  It's fine to have both?

satmandu: I guess it makes sense to have a conditional throw an error if the function is passed a nonexistent `src` dir, and then also not invoke the function if the `src` dir doesn't exist.

satmandu: How about...
```
def install_files(prefix, path, dst = File.join( CREW_PREFIX, File.join(prefix, path).delete_prefix(".#{CREW_PREFIX}") ))
  src = File.join(prefix, path)
  if Dir.exist?(src)
    if File.executable?("#{CREW_PREFIX}/bin/crew-mvdir") && !CREW_DISABLE_MVDIR
      system "crew-mvdir #{@short_verbose} #{src} #{dst}", exception: true
    else
      warn 'crew-mvdir is not installed. Please install it with \'crew install crew_mvdir\' for improved installation performance'.yellow unless @pkg.name == 'crew_mvdir'
  
      if File.executable?("#{CREW_PREFIX}/bin/rsync") && system("#{CREW_PREFIX}/bin/rsync --version > /dev/null")
        # rsync src path needs a trailing slash
        src << '/' unless src.end_with?('/')
        system 'rsync', "-ah#{@verbose}HAXW", '--remove-source-files', src, dst, exception: true
      else
        system "cd #{src}; tar -cf - ./* | (cd #{dst}; tar -x#{@verbose}p --keep-directory-symlink -f -)", exception: true
      end
    end
  else
    puts "#{src} directory does not exist.".lightred
    return false 
  end
end
```

satmandu: (leaving aside the logic question about the parameters to this function.)

lib/const.rb
uberhacker: Lost new crew version?

satmandu: Yup. Let me fix that.

satmandu: fixed

uberhacker:
I definitely noticed the speed improvement during my tests.  Great work @supechicken!

bin/crew
```diff
@@ -1268,6 +1268,26 @@ def shrink_dir(dir)
   end
 end
 
+def install_files(src, dst = File.join( CREW_PREFIX, src.delete_prefix(".#{CREW_PREFIX}") ))
```
uberhacker, line 1271: It makes no sense to have `".#{CREW_PREFIX}"`.  That should be `"./usr/local"`, otherwise, why even delete the prefix?

uberhacker (approved):

PR #7753, opened 2022-12-19
Update electric, far, logisim, pmd and sejda_console packages

uberhacker:
Updates to change dependencies from jdk8 to openjdk8.

### DIFF
--- a/packages/pmd.rb
+++ b/packages/pmd.rb
@@ -17,18 +17,14 @@ class Pmd < Package
   def self.build
     cpd = <<~EOF
       #!/bin/bash
-      PWD=$(pwd)
       cd #{CREW_PREFIX}/share/pmd
-      bin/run.sh cpd "$@"
-      cd $PWD
+      exec bin/run.sh cpd "$@"
     EOF
     File.write('cpd', cpd)
     pmd = <<~EOF
       #!/bin/bash
-      PWD=$(pwd)
       cd #{CREW_PREFIX}/share/pmd
-      bin/run.sh pmd "$@"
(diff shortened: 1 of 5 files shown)


packages/pmd.rb
```diff
@@ -3,45 +3,45 @@
...
 
+  no_compile_needed
+
+  def self.build
+    cpd = <<~EOF
+      #!/bin/bash
+      PWD=$(pwd)
+      cd #{CREW_PREFIX}/share/pmd
+      bin/run.sh cpd "$@"
+      cd $PWD
```
supechicken, line 23: We don't need to `cd` back since this is independent from the shell process

packages/pmd.rb
```diff
@@ -3,45 +3,45 @@
...
   depends_on 'unzip'
 
+  no_compile_needed
+
+  def self.build
+    cpd = <<~EOF
+      #!/bin/bash
+      PWD=$(pwd)
+      cd #{CREW_PREFIX}/share/pmd
+      bin/run.sh cpd "$@"
```
supechicken, line 22: `exec bin/run.sh cpd "$@"` to prevent bash waiting in background?

packages/pmd.rb
uberhacker: Good call.  Fixed.

packages/pmd.rb
uberhacker: Fixed.

supechicken (approved):

satmandu (approved):

PR #7785, opened 2022-12-30
Fix crew build issue

### LINKED ISSUE #7784: 🐞 Bug: Unable to build packages (no compressed archive generated)
This is a consistent bug.  Running `crew build zstd` results in the following:
```
Using zstd to compress package. This may take some time.
sha256sum: /home/chronos/user/chromebrew/release/x86_64/zstd-1.5.2-1-chromeos-x86_64.tar.zst: No such file or directory
```
Restoring the repo to hash 850c53f2e82b26562e81a0b0fe8547dca4e79596 fixes this problem.

uberhacker:
Fixes #7784.  Restored `def archive_package` function in crew.

Explanation:
We cannot remove `@crew_prefix_zstd_available` and `@crew_musl_prefix_zstd_available` variables because they are used later in the function.

Note: Merge this after #7783.

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1468,10 +1468,19 @@ def build_package(pwd)
 end
 
 def archive_package(pwd)
-  if @pkg.no_zstd?
+  # Check to see that there is a working zstd
+  if File.file?("#{CREW_PREFIX}/bin/zstd")
+    @crew_prefix_zstd_available = `#{CREW_PREFIX}/bin/zstd --version`.include?('zstd command line interface') ? true : nil
+  end
+  if File.file?("#{CREW_MUSL_PREFIX}/bin/zstd")
+    @crew_musl_prefix_zstd_available = `#{CREW_MUSL_PREFIX}/bin/zstd --version`.include?('zstd command line interface') ? true : nil
+  end
+  if @pkg.no_zstd? || (!@crew_prefix_zstd_available && !@crew_musl_prefix_zstd_available)
     puts 'Using xz to compress package. This may take some time.'.lightblue
     pkg_name = "#{@pkg.name}-#{@pkg.version}-chromeos-#{@device[:architecture]}.tar.xz"
-    system "tar c#{@verbose}Jf #{pwd}/#{pkg_name} *", chdir: CREW_DEST_DIR
+    Dir.chdir CREW_DEST_DIR do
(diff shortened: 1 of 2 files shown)


bin/crew
```diff
@@ -1468,10 +1468,19 @@ def build_package(pwd)
...
+  if File.file?("#{CREW_MUSL_PREFIX}/bin/zstd")
+    @crew_musl_prefix_zstd_available = `#{CREW_MUSL_PREFIX}/bin/zstd --version`.include?('zstd command line interface') ? true : nil
+  end
+  if @pkg.no_zstd? || (!@crew_prefix_zstd_available && !@crew_musl_prefix_zstd_available)
     puts 'Using xz to compress package. This may take some time.'.lightblue
     pkg_name = "#{@pkg.name}-#{@pkg.version}-chromeos-#{@device[:architecture]}.tar.xz"
-    system "tar c#{@verbose}Jf #{pwd}/#{pkg_name} *", chdir: CREW_DEST_DIR
+    Dir.chdir CREW_DEST_DIR do
+      system "tar c#{@verbose}Jf #{pwd}/#{pkg_name} *"
+    end
```
supechicken, line 1483: Can we delete the `if` logic below instead? (as the result generated with `musl` build and regular build should be identical) (this also combined the `Dir.chdir` block)
```ruby
      # Using same zstd compression options as Arch, which privilege
      # decompression speed over compression speed.
      # See https://lists.archlinux.org/pipermail/arch-dev-public/2019-March/029542.html
      # Use nice so that user can (possibly) do other things during compression.
      system "tar c#{@verbose} * | nice -n 20 zstd -c -T0 --ultra -20 - > #{pwd}/#{pkg_name}", chdir: CREW_DEST_DIR
```

uberhacker: I just restored the entire function without too much analysis.  We have to do a better job of testing this stuff before making wholesale changes.  Maybe it would make sense on the next PRs to break down the changes into smaller chunks to keep some sanity?  I appreciate the enthusiasm to make significant changes but we have to pay more attention to detail going forward.  That said, can we just merge this and you can change it in another PR?  This was your mess I cleaned up in the first place. :-P

supechicken: Sorry about that😅 I think I should test each command next time...

supechicken (approved):

PR #7788, opened 2022-12-30
File => 5.44

satmandu:
Fixes #7786

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

### Run the following to get this pull request's changes locally for testing.
```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=filecmd544 CREW_TESTING=1 crew update
```

PR #7789, opened 2022-12-30
Filecmd 5.43 => 5.44

uberhacker:
Add lzlib package

Fixes #7788.

### DIFF
--- a/packages/filecmd.rb
+++ b/packages/filecmd.rb
@@ -3,37 +3,39 @@ require 'package'
 class Filecmd < Package
   description 'file and libmagic determine file type'
   homepage 'http://ftp.astron.com/'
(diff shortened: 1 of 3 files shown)


satmandu:
You also need the zstd and GLIBC deps like in my PR...

satmandu:
(run the getrealdeps script to verify...)

satmandu (commented):
Also add `no_strip` to filecmd?

uberhacker:
Rather than using `no_strip`, why don't we fix the syntax error?

satmandu:
> Rather than using `no_strip`, why don't we fix the syntax error?

You're right. Just realized I'm getting it on other packages too.

satmandu (approved):

PR #7297, opened 2022-08-19
Refactor install.sh

Zopolis4:
Option 3.

In order for this to work, the ruby binary needs to be reuploaded as `.tar.xz`. I am aware that changing ruby to `.tar.xz` will lead to slower decompression, but it is the neatest and simplest way to make this work.

This refactors some of install.sh to be much slimmer, and makes some minor style changes. `install.sh` is pretty inconsistent when it comes to style overall, so these style changes were only for comments. I do think we should have a consistent style across the board, but that is for another PR.

The main purpose of this PR is to enable #7082, which is what the changes in `core.rb` and `ruby.rb` are for. 

Not even remotely tested.
### Run the following to get this pull request's changes locally for testing.
```
exec bash --init-file <(curl -Ls https://raw.githubusercontent.com/Zopolis4/chromebrew/betterstall/install.sh)
```

PR #7797, opened 2023-01-03
Do not use @_ver for gcc, git, openssl and ruby.

Zopolis4:
I'm being the change I want to see in the world. Dependency for #7297.

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git
CREW_TESTING_BRANCH=naver CREW_TESTING=1 crew update
```


satmandu:
Linter failed...

uberhacker:
I can get onboard with this request.  However, keep in mind that we may need to return to `@_ver` if the binaries need to be rebuilt with the same source.

satmandu (approved):

PR #7798, opened 2023-01-03
Lint tree with rubocop, again.

Zopolis4:
While attempting to locally run rubocop, I encountered all of these errors. They seemed to mostly be of the Style/RedundantStringEscape style, so I presume its just a new rubocop cop.

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=nodiff CREW_TESTING=1 crew update
```


satmandu (approved):

uberhacker:
Please don't merge this PR.  We CANNOT make changes like this.  The `system "sed ...` lines need to have the `\` escape characters so the variables are not evaluated.  Otherwise, the lines need to change to `system 'sed ...`.  Single quote `'` instead of double quote `"`.

uberhacker:
These changes are not correct.  Please review and test before creating a PR.

satmandu:
> Please don't merge this PR. We CANNOT make changes like this. The `system "sed ...` lines need to have the `\` escape characters so the variables are not evaluated. Otherwise, the lines need to change to `system 'sed ...`. Single quote `'` instead of double quote `"`.

Are we sure that these are breaking changes, and not something that ruby automagically ignores?

Zopolis4:
If they are breaking, this should be reported to rubocop. I can't easily test this without installing all of the packages, which is not quick to do, and something that I prefer to avoid for a simple lint check.

PR #7800, opened 2023-01-04
Move @_ver comments to same line as version value.

Zopolis4:
Having these on a separate line pushes down the version value and breaks sed.


satmandu (approved):

satmandu:
Next time add reviewers to your PR so we get to it faster. :)

PR #7807, opened 2023-01-05
Add @_ver comment to ca_certificates.rb

Zopolis4:
I dislike sed.

Anyways, this comment is necessary for the install to work, and I cannot be bothered fixing it, and fixing it would make the install.sh sed even more complicated.


Zopolis4:
@satmandu because I don't have the perms to request reviewers.

uberhacker (approved):

PR #7813, opened 2023-01-05
fixup prompts

### LINKED ISSUE #7812: 🐞 Bug Crew does not wait for responses to postinstall prompts when an install also requires dependencies to be installed
Queries in the `postinstall` give users agency over their install. Automatically defaulting to answers is a bug, unless that default intentionally has a timeout connected automatic selection.

Working prompt behavior in a new container or install:
```
yes | crew install `comm -13 <(crew -d list installed | sort) <(crew deps -b -d "firefox" | sort)`
crew install firefox
```
Broken prompt behavior in a new container or install:
```
crew install firefox
```
Output as seen:
image

(This also affects `vim` and `neovim` and probably all other installs with queries in the `postinstall`.)

satmandu:
Fixes #7812 

- Get string, chomp it, downcase it, and then look for `''`, `y`, and `yes` in input string. This allows one to hit enter after the input without having that newline passed to another prompt.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=fix_prompts CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/php.rb
+++ b/packages/php.rb
@@ -28,7 +28,7 @@ class Php < Package
     puts '8.2 = PHP 8.2.1'
     puts '  0 = Cancel'
 
-    while version = $stdin.gets.chomp
+    while version = $stdin.gets.chomp.downcase
       case version
       when '5.6'
         depends_on 'php5'
--- a/bin/crew
+++ b/bin/crew
@@ -1405,11 +1405,11 @@ def resolve_dependencies
   puts @dependencies.join(' ')
 
   print 'Do you agree? [Y/n] '
-  response = $stdin.getc
+  response = $stdin.gets.chomp.downcase
   case response
-  when 'n'
+  when 'n', 'no'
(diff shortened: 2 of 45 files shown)


packages/php.rb
```diff
@@ -28,7 +28,7 @@ def self.preflight
     puts '8.2 = PHP 8.2.1'
     puts '  0 = Cancel'
 
-    while version = $stdin.gets.chomp
+    while version = $stdin.gets.chomp.downcase
```
uberhacker, line 31: No need for `.downcase` here since all the expected input will be numeric.

uberhacker (approved):

packages/php.rb
satmandu: Ah well, you can fix it on the next update. :)

PR #7816, opened 2023-01-07
Update version checks in check_updates.sh

uberhacker:
Still not perfect but WAY better than it was prior to this update.


satmandu:
Linter...

satmandu (approved):

PR #7817, opened 2023-01-07
glibc 2.33 rebuild, add glibc 2.35

satmandu:
- This should support M108 properly, and also fix an issue with libnsl.so.2 getting deleted during the build process, which breaks compatibility with software built on the M9x containers.

Works properly _in containers and on M108 x86_64 hardware_:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.comsatmandu/chromebrew.git CREW_TESTING_BRANCH=glibc_rebuild CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/glibc.rb
+++ b/packages/glibc.rb
@@ -18,7 +18,7 @@ class Glibc < Package
   @libc_version = LIBC_VERSION
   # Uncomment following line to build a version of glibc different
   # from the one ChromeOS ships with.
-  # @libc_version = '2.33'
+  # @libc_version = '2.35'
   if @libc_version == '2.23'
     version '2.23-6'
     source_url 'https://ftpmirror.gnu.org/glibc/glibc-2.23.tar.xz'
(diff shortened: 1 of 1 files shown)


packages/glibc.rb
```diff
@@ -323,6 +342,27 @@ def self.patch
...
+      system "sed -i 's,be_quiet,locale_be_quiet,g' fedora/build-locale-archive.c"
+      @googlesource_branch = 'release-R109-15236.B'
+      system "git clone --depth=1 -b  #{@googlesource_branch} https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay googlesource"
+      # Remove conflicting patches.
+      FileUtils.rm 'googlesource/sys-libs/glibc/files/local/glibc-2.35/0005-Add-dlopen_with_offset-to-glibc.patch'
+      FileUtils.rm 'googlesource/sys-libs/glibc/files/local/glibc-2.35/0006-glibc-add-clang-style-FORTIFY.patch'
+      Dir.glob('googlesource/sys-libs/glibc/files/local/glibc-2.35/*.patch').each do |patch|
+        puts "patch -Np1 < #{patch} || true" if @opt_verbose
+        system "patch -Np1 -F 10 -i #{patch} || true"
+      end
```
uberhacker, line 365: This is so ugly.

uberhacker (approved):

packages/glibc.rb
satmandu: The `|| true` isn't actually necessary. The patches apply just fine.

PR #7828, opened 2023-01-11
Docker

satmandu:
- This is a work in progress docker package.
- The primary remaining issue with this is figuring out group settings and startup scripts.
- All pre-reqs have been merged.
- The binaries RUN... but there is nothing currently in place in regards to documentation for trying to get a docker daemon running. Perhaps someone wants to take a go at that since the binaries are available?

BUILDS properly:
- [x] `x86_64`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=docker2023-1 CREW_TESTING=1 crew update
```


uberhacker:
This would be a very welcomed addition if you can work it out.

satmandu:
Maybe it makes sense to merge as is at the moment... and let people figure out if there is a good way to make this work.

packages/docker.rb
```diff
@@ -0,0 +1,177 @@
...
+  depends_on 'glibc' # R
+  depends_on 'go' => :build
+  depends_on 'go_md2man' => :build
+  depends_on 'iproute2'
+  depends_on 'lvm2' # R
+  depends_on 'sqlite'
+
+  no_env_options
+  no_fhs
+
```
uberhacker, line 40: Add `def self.preflight` section to determine compatibility?  IOW, abort and display an error message to Chromebook users that do not have support for Linux Apps?

packages/docker.rb
```diff
@@ -0,0 +1,177 @@
...
+    })
+    binary_sha256({
+      aarch64: '923b0eb95c947dde422e196abd0f1ddbc0958792db2807e2c55c3048b700702a',
+       armv7l: '923b0eb95c947dde422e196abd0f1ddbc0958792db2807e2c55c3048b700702a',
+       x86_64: '21f9db6ec0086ab6452be9cd50a6bc980eb00caafb439a3387802c5363085ec1'
+    })
+
+    @libnetwork_version = '3f0048413d95802b9c6c836eba06bfc54f9dbd03'
+    @tini_version = '0b44d3665869e46ccbac7414241b8256d6234dc4'
+    @buildx_version = 'v0.10.5'
```
uberhacker, line 70: Should these be separate dependency packages?

uberhacker (changes requested):
See comments.

packages/docker.rb
satmandu: Done!

packages/docker.rb
satmandu: Nothing else uses them... and they all get built together...

uberhacker: We'll just have to remember to increment the versions on updates, if necessary.

packages/docker.rb
uberhacker: Simple enough!

uberhacker (approved):

ryugenxd (commented):
🤓

satmandu:
> 🤓

Feel free to provide feedback if you get it working!

PR #7843, opened 2023-01-16
Add transmission dep to fragments

satmandu:
Fixes https://gitlab.gnome.org/World/Fragments/-/issues/179

Works properly:
- [x] `x86_64`
- [x] `armv7l`

### DIFF
--- a/packages/fragments.rb
+++ b/packages/fragments.rb
@@ -36,6 +36,7 @@ class Fragments < Package
   depends_on 'openssl' # R
   depends_on 'pango' # R
   depends_on 'rust' => :build
+  depends_on 'transmission' # L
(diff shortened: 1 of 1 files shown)


packages/fragments.rb
```diff
@@ -36,6 +36,7 @@ class Fragments < Package
   depends_on 'openssl' # R
   depends_on 'pango' # R
   depends_on 'rust' => :build
+  depends_on 'transmission' # L
```
uberhacker, line 39: Do we need new binaries?

satmandu: No. The software just doesn't throw an error when transmission isn't installed. It's an upstream bug.

uberhacker (approved):

PR #7894, opened 2023-02-04
Update php packages

uberhacker:
- php81 8.1.14 => 8.1.15
- php82 8.2.1 => 8.2.2


satmandu (approved):
I'm okay with getting rid of the `@_ver` here...

uberhacker:
Yeah, I figured it was easier to change the urls for future updates.  It might make sense to continue to do this with packages that include the version more than once in the url or throughout the package.

PR #7907, opened 2023-02-08
Update Xterm and add some X app deps and utils

satmandu:
Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=x2 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/ncompress.rb
+++ b/packages/ncompress.rb
@@ -13,16 +13,16 @@ class Ncompress < Package
   source_sha256 '96ec931d06ab827fccad377839bfb91955274568392ddecf809e443443aead46'
 
   binary_url({
-     aarch64: 'file:///usr/local/tmp/packages/ncompress-5.0-chromeos-armv7l.tar.zst',
-      armv7l: 'file:///usr/local/tmp/packages/ncompress-5.0-chromeos-armv7l.tar.zst',
-        i686: 'file:///usr/local/tmp/packages/ncompress-5.0-chromeos-i686.tar.zst',
-      x86_64: 'file:///usr/local/tmp/packages/ncompress-5.0-chromeos-x86_64.tar.zst'
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ncompress/5.0_armv7l/ncompress-5.0-chromeos-armv7l.tar.zst',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/ncompress/5.0_armv7l/ncompress-5.0-chromeos-armv7l.tar.zst',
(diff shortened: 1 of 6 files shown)


packages/ncompress.rb
```diff
@@ -0,0 +1,38 @@
...
+  license 'PublicDomain'
+  compatibility 'all'
+  source_url 'https://github.com/vapier/ncompress/archive/v5.0.tar.gz'
+  source_sha256 '96ec931d06ab827fccad377839bfb91955274568392ddecf809e443443aead46'
+
+  binary_url({
+     aarch64: 'file:///usr/local/tmp/packages/ncompress-5.0-chromeos-armv7l.tar.zst',
+      armv7l: 'file:///usr/local/tmp/packages/ncompress-5.0-chromeos-armv7l.tar.zst',
+        i686: 'file:///usr/local/tmp/packages/ncompress-5.0-chromeos-i686.tar.zst',
+      x86_64: 'file:///usr/local/tmp/packages/ncompress-5.0-chromeos-x86_64.tar.zst'
```
uberhacker, line 19: Local files?

packages/xterm.rb
```diff
@@ -3,46 +3,83 @@
...
+      --enable-wide-chars \
+      --with-app-defaults=#{CREW_PREFIX}/share/X11/app-defaults/ \
+      --with-pcre \
+      --with-pkg-config \
+      --with-tty-group=tty \
+      --with-utempter \
+      --with-x \
+      --with-xinerama \
+      --with-xpm=#{CREW_PREFIX}"
+    system "sed -i '/^LDFLAGS/ s/$/ -DUSE_TERMINFO=1 -ltinfo/' Makefile"
```
uberhacker, line 82: Move to `def self.patch` section?

satmandu: Can't, as configure needs to run first.

But I'm having a discussion upstream about how to do this cleanly...

packages/ncompress.rb
satmandu: Fixed!

packages/xterm.rb
uberhacker: Can it be patched in `Makefile.in` prior to the `configure`?

satmandu: Possibly... but I'm waiting for an upstream fix to resolve the issue without needing patching...

I've sent my config.status and config.log to the maintainer as requested...

satmandu: I just went ahead and added a patch section to modify the Makefile.in...

uberhacker (approved):

PR #7927, opened 2023-02-12
Adding package: helix_editor

ncdm-stldr:
## Description
Adding *helix editor*, a vim inspired editor that supports language servers (LSP). See [here](https://helix-editor.com/) for more information.

## Additional information

Works properly:
- [ ] `x86_64` (not tested)
- [ ] `i686` (not tested)
- [x] `armv7l` 

No install binaries at the moment.

```
CREW_TESTING_REPO=https://github.com/ncdm-stldr/chromebrew.git CREW_TESTING_BRANCH=add-helix_editor CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -49,7 +49,7 @@ class Helix_editor < Package
     command_status = system 'hx --version', exception: false
     puts 'Warning: hx is not in PATH'.lightred unless command_status == true
     # Check if helix can find its runtime path
-    if command_status == true
+    if command_status
       command_output = `hx --health`
       unless command_output.include? @helix_runtime_dir
         puts "Warning: helix cannot find its runtime dir. \
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -30,12 +30,12 @@ class Helix_editor < Package
     # Copy executable
     helix_executable_dest_dir = "#{CREW_DEST_PREFIX}/bin"
     FileUtils.mkdir_p helix_executable_dest_dir.to_s
-    FileUtils.install ".#{@build_folder_suffix}/hx", helix_executable_dest_dir, mode: 0o755
+    FileUtils.install ".#{@build_folder_suffix}/hx", helix_executable_dest_dir.to_s, mode: 0o755
     # Copy runtime dir (without the sources)
     FileUtils.rm_rf './runtime/grammars/sources' # remove the sources
     helix_runtime_dest_dir = "#{CREW_DEST_DIR}#{@helix_runtime_dir}"
-    FileUtils.mkdir_p helix_runtime_dest_dir
-    FileUtils.cp_r './runtime', helix_runtime_dest_dir
+    FileUtils.mkdir_p helix_runtime_dest_dir.to_s
+    FileUtils.cp_r './runtime', helix_runtime_dest_dir.to_s
   end
 
   def self.check
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -12,10 +12,10 @@ class Helix_editor < Package
   depends_on 'rust' => :build
   depends_on 'xdg_base'
 
-  @no_fhs = true
-  @no_shrink = true
-  @no_strip = true
-  @no_patchelf = true
+  @no_fhs
+  @no_shrink
+  @no_strip
+  @no_patchelf
 
   @xdg_config_home = ENV.fetch 'XDG_CONFIG_HOME', "#{CREW_PREFIX}/.config"
   @helix_runtime_dir = "#{@xdg_config_home}/helix"
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -43,12 +43,12 @@ class Helix_editor < Package
   def self.postinstall
     # This will print a warning if "hx" is not a valid command
     command_status = system 'hx --version', exception: false
-    puts 'Warning: hx is not in path'.red unless command_status == true
+    puts 'Warning: hx is not in path'.lightred unless command_status == true
     # Check if helix can find its runtime path
     command_output = `hx --health`, exception: false
-    puts 'Warning: helix cannot find its runtime dir'.red unless command_output.include? @helix_runtime_dir
+    puts 'Warning: helix cannot find its runtime dir'.lightred unless command_output.include? @helix_runtime_dir
 
-    puts <<~EOT2.orange
+    puts <<~EOT2.lightblue
       Use the 'hx' command to start helix.
       Use 'hx --health' to see if helix can find its runtime and to see which LSP
       servers are detected.
@@ -67,9 +67,9 @@ class Helix_editor < Package
     case $stdin.gets.chomp.downcase
     when 'y', 'yes'
       FileUtils.rm_rf config_dir
-      puts "#{config_dir} removed.".lightred
+      puts "#{config_dir} removed.".lightgreen
     else
-      puts "#{config_dir} was not removed.".lightgreen
+      puts "#{config_dir} was not removed.".lightblue
     end
   end
 end
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -49,9 +49,13 @@ class Helix_editor < Package
     command_status = system 'hx --version', exception: false
     puts 'Warning: hx is not in PATH'.lightred unless command_status == true
     # Check if helix can find its runtime path
-    command_output = `hx --health`
-    puts 'Warning: helix cannot find its runtime dir'.lightred unless command_output.include? @helix_runtime_dir
-
+    if command_status == true
+      command_output = `hx --health`
+      unless command_output.include? @helix_runtime_dir
+        puts "Warning: helix cannot find its runtime dir. \
+        The environment variable HELIX_RUNTIME may need to be set.".lightred
+      end
+    end
     puts <<~EOT2.lightblue
       Use the 'hx' command to start helix.
       Use 'hx --health' to see if helix can find its runtime and to see which LSP
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -1,52 +1,62 @@
 require 'package'
 
-class Helix_editor < Package                 # The first character of the class name must be upper case
+class Helix_editor < Package
   description 'A vim inspired editor with LSP support'
   homepage 'https://helix-editor.com/'
   version '22.12'
   license 'MPL-2.0' # license of source
   compatibility 'all'
-  source_url 'https://github.com/helix-editor/helix/archive/refs/tags/22.12.tar.gz'
-  source_sha256 'edae8af46401b45c3e71c38b4fa99f931c4458127978ccd1b29aaae79331d972'   # Use the command "sha256sum"
-  
+  source_url 'https://t.ly/-vk7'
+  source_sha256 'edae8af46401b45c3e71c38b4fa99f931c4458127978ccd1b29aaae79331d972' # Use the command "sha256sum"
+  source_url 'file:///usr/local/home/helix/deleteme.tar.gz'
+  source_sha256 'c3a0e022a7bcd6b139ae7abf1e0dae33c8f9ec82dc680076d9e5ea6593b85bc6'
+
   depends_on 'rust' => :build
+  depends_on 'xdg_base'
 
   @no_fhs = true
-  
+
+  @xdg_config_home = ENV.fetch 'XDG_CONFIG_HOME', '/usr/local/.config'
+  @helix_runtime_dir = "#{@xdg_config_home}/helix"
+
   def self.build
-   puts "Building. This may be long."
-   system "cargo build \
+    puts 'Building. This may be long.'
+    puts 'cargo build \
       --release \
       --locked \
-     "
+     '
   end
+
+  def self.check
+    # This will raise an error if "hx" is not a valid command
+    `hx --version`
+    # Check if helix can find its runtime path
+    command_output = `hx --health`
+    if !command_output.include?(@helix_runtime_dir.to_s)
+      raise 'Helix cannot find its runtime dir'
+    end
+  end
+
 
   def self.install
     # Copy executable
     helix_executable_dest_dir = "#{CREW_DEST_PREFIX}/bin"
-    FileUtils.mkdir_p("#{helix_executable_dest_dir}")
-    FileUtils.cp("./target/release/hx", "#{helix_executable_dest_dir}")
+    FileUtils.mkdir_p helix_executable_dest_dir.to_s
+    FileUtils.install './target/release/hx', helix_executable_dest_dir.to_s, :mode => 0755
     # Copy runtime dir
-    helix_runtime_dest_dir = "#{CREW_DEST_DIR}#{ENV['XDG_CONFIG_HOME']}/helix"
-    FileUtils.mkdir_p("#{helix_runtime_dest_dir}")
-    FileUtils.cp_r("./runtime", "#{helix_runtime_dest_dir}")
-  end
-
-  def self.check
-    puts "self.check is called"
-    return false
-    # Check "hx --version" status code and if hx found its runtime directory
-    # return (system "hx --version") && (`hx --health`.include?("#{ENV['XDG_CONFIG_HOME']}"))
+    helix_runtime_dest_dir = "#{CREW_DEST_DIR}#{@helix_runtime_dir}"
+    FileUtils.mkdir_p helix_runtime_dest_dir.to_s
+    FileUtils.cp_r './runtime', helix_runtime_dest_dir.to_s
   end
 
   def self.postinstall
     puts <<~EOT2.orange
       Use the 'hx' command to start helix.
-    
-      Use 'hx --health' to see if helix detects its runtime and to see which LSP 
+
+      Use 'hx --health' to see if helix can find its runtime and to see which LSP
       servers are detected.
-    
-      To be able to load some themes, helix neefds to be started in a terminal it recognizes 
+
+      To be able to load some themes, helix needs to be started in a terminal it recognizes#{' '}
       as supporting true colors.
     EOT2
   end
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -6,10 +6,8 @@ class Helix_editor < Package
   version '22.12'
   license 'MPL-2.0' # license of source
   compatibility 'all'
-  source_url 'https://t.ly/-vk7'
+  source_url 'https://github.com/helix-editor/helix/archive/22.12.tar.gz' #'https://t.ly/-vk7'
   source_sha256 'edae8af46401b45c3e71c38b4fa99f931c4458127978ccd1b29aaae79331d972' # Use the command "sha256sum"
-  source_url 'file:///usr/local/home/helix/deleteme.tar.gz'
-  source_sha256 'c3a0e022a7bcd6b139ae7abf1e0dae33c8f9ec82dc680076d9e5ea6593b85bc6'
 
   depends_on 'rust' => :build
   depends_on 'xdg_base'
@@ -21,10 +19,10 @@ class Helix_editor < Package
 
   def self.build
     puts 'Building. This may be long.'
-    puts 'cargo build \
+    system "cargo build \
       --release \
       --locked \
-     '
+     "
   end
 
   def self.check
@@ -59,5 +57,21 @@ class Helix_editor < Package
       To be able to load some themes, helix needs to be started in a terminal it recognizes#{' '}
       as supporting true colors.
     EOT2
+  end
+
+  def self.remove
+    # Some files in the directory are write protected...
+    FileUtils.rm_rf @helix_runtime_dir
+    # If the user added a configuration dir, we ask if he wishes to remove it as well
+    config_dir = "#{HOME}/.config/helix"
+    Dir.exist? config_dir
+      puts "\nWould you like to remove the configuration folder: #{config_dir}? [y/N] "
+      case $stdin.gets.chomp.downcase
+      when 'y', 'yes'
+        FileUtils.rm_rf config_dir
+        puts "#{config_dir} removed.".lightred
+      else
+        puts "#{config_dir} was not removed.".lightgreen
+      end
   end
 end
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -13,6 +13,9 @@ class Helix_editor < Package
   depends_on 'xdg_base'
 
   @no_fhs = true
+  @no_shrink = true
+  @no_strip = true
+  @no_patchelf = true
 
   @xdg_config_home = ENV.fetch 'XDG_CONFIG_HOME', "#{CREW_PREFIX}/.config"
   @helix_runtime_dir = "#{@xdg_config_home}/helix"
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -58,8 +58,7 @@ class Helix_editor < Package
     end
     puts <<~EOT2.lightblue
       Use the 'hx' command to start helix.
-      Use 'hx --health' to see if helix can find its runtime and to see which LSP
-      servers are detected.
+      Use 'hx --health' to see if helix can find its runtime and to see which LSP servers are detected.
       Note that to be able to load some themes, helix needs to be started in a terminal it recognizes
       as supporting true colors.
     EOT2
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -8,7 +8,7 @@ class Helix_editor < Package                 # The first character of the class 
   compatibility 'all'
   source_url 'https://github.com/helix-editor/helix/archive/refs/tags/22.12.tar.gz'
   source_sha256 'edae8af46401b45c3e71c38b4fa99f931c4458127978ccd1b29aaae79331d972'   # Use the command "sha256sum"
-
+  
   depends_on 'rust' => :build
 
   @no_fhs = true
@@ -33,10 +33,10 @@ class Helix_editor < Package                 # The first character of the class 
   end
 
   def self.check
-    # TODO: check seems to be never called
-    # system "hx --version"
-    # helix_health_output = capture(:stdout) { system("hx --health") }
-    # return helix_health_output.include?("#{CREW_PREFIX}#{ENV['XDG_CONFIG_HOME']}/helix")
+    puts "self.check is called"
+    return false
+    # Check "hx --version" status code and if hx found its runtime directory
+    # return (system "hx --version") && (`hx --health`.include?("#{ENV['XDG_CONFIG_HOME']}"))
   end
 
   def self.postinstall
@@ -46,7 +46,7 @@ class Helix_editor < Package                 # The first character of the class 
       Use 'hx --health' to see if helix detects its runtime and to see which LSP 
       servers are detected.
     
-      To use the default theme, helix has to be started in a terminal it recognizes 
+      To be able to load some themes, helix neefds to be started in a terminal it recognizes 
       as supporting true colors.
     EOT2
   end
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -41,13 +41,13 @@ class Helix_editor < Package
   def self.check
     # Ensure hx is executable
     command_status = system ".#{@build_folder_suffix}/hx --version", exception: false
-    raise 'hx is not executable'.lightred unless command_status == true
+    raise 'hx is not executable'.lightred unless command_status
   end
 
   def self.postinstall
     # This will print a warning if "hx" is not a valid command
     command_status = system 'hx --version', exception: false
-    puts 'Warning: hx is not in PATH'.lightred unless command_status == true
+    puts 'Warning: hx is not in PATH'.lightred unless command_status
     # Check if helix can find its runtime path
     if command_status
       command_output = `hx --health`
--- a/packages/helix_editor.rb
+++ b/packages/helix_editor.rb
@@ -69,14 +69,15 @@ class Helix_editor < Package
     FileUtils.rm_rf @helix_runtime_dir
     # If the user added a configuration dir, we ask if he wishes to remove it as well
     config_dir = "#{HOME}/.config/helix"
-    Dir.exist? config_dir
-    puts "\nWould you like to remove the configuration folder: #{config_dir}? [y/N] "
-    case $stdin.gets.chomp.downcase
-    when 'y', 'yes'
-      FileUtils.rm_rf config_dir
-      puts "#{config_dir} removed.".lightgreen
-    else
-      puts "#{config_dir} was not removed.".lightblue
+    if Dir.exist? config_dir
+      puts "\nWould you like to remove the configuration folder: #{config_dir}? [y/N] "
+      case $stdin.gets.chomp.downcase
+      when 'y', 'yes'
+        FileUtils.rm_rf config_dir
+        puts "#{config_dir} removed.".lightgreen
+      else
+        puts "#{config_dir} was not removed.".lightblue
+      end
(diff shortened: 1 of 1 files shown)


packages/helix_editor.rb
```diff
@@ -0,0 +1,73 @@
...
+  #source_url 'file:///usr/local/home/helix/deleteme.tar.gz'
+  #source_sha256 'c3a0e022a7bcd6b139ae7abf1e0dae33c8f9ec82dc680076d9e5ea6593b85bc6'
+
+  depends_on 'rust' => :build
+
+  # TODO is it the appropriate location?
+  # By default, helix expects runtime to be in ~/.config or XDG_CONFIG_HOME.
+  # The first location is not appropriate because helix will not be installed in ~/
+  # With the second location crew complaints that it is not  FHS3 compliant.
+  @runtime_dest_dir_suffix = "/var/lib/helix"
```
satmandu, line 20: Everything ChromeOS touches needs to be under `/usr/local`...

uberhacker: Yes and this extra directory is not even necessary if the output of the build is a single executable.

packages/helix_editor.rb
```diff
@@ -0,0 +1,73 @@
...
+    FileUtils.mkdir_p("#{helix_executable_dest_dir}")
+    FileUtils.cp("./target/release/hx", "#{helix_executable_dest_dir}")
+    # Copy runtime dir
+    FileUtils.mkdir_p("#{CREW_DEST_PREFIX}#{@runtime_dest_dir_suffix}")
+    FileUtils.cp_r("./runtime", "#{CREW_DEST_PREFIX}#{@runtime_dest_dir_suffix}")
+    # Set env var for runtime dir
+    env_dir = "#{CREW_DEST_PREFIX}/etc/env.d"
+    FileUtils.mkdir_p "#{env_dir}"
+    File.write("#{env_dir}/helix",
+               "export HELIX_RUNTIME=#{CREW_PREFIX}#{@runtime_dest_dir_suffix}/runtime")
```
uberhacker, line 42: If this is a single executable without any dependencies, it could be installed directly in `"#{CREW_DEST_PREFIX}/bin"`.  The installed location which is `/usr/local/bin` by default is already in the path.

packages/helix_editor.rb
```diff
@@ -0,0 +1,73 @@
...
+    # system "source #{env_dir}/helix"
+  end
+
+  def self.check
+    # # TODO (verify that this check works, and if it fails when the system language
+    # #  is not english)
+    # helix_health = capture(:stdout) { system("hx --health") }
+    # return !helix_health.contains('Runtime directory does not exist')
+    return true
```
uberhacker, line 51: A simple test could be to run the executable to see if it provides any output.  Something like `helix --version`.

packages/helix_editor.rb
ncdm-stldr: Yes all the files normally go under `/usr/local/`.

Maybe the code comments are a bit unclear, but `@runtime_dest_dir_suffix` is not used as a full path. It is used as a suffix, and appended after a prefix as in: `"#{CREW_DEST_PREFIX}#{@runtime_dest_dir_suffix}"`.

Also, *helix editor* is in two parts:
- The `hx` executable, that would go inside `/usr/local/bin` I suppose.
- The `runtime` directory that is needed by `hx` in order to work properly. Without it, the themes will not be available, and the code syntax highlighting will not work either.

By default, helix expects the runtime folder to be found either under `$XDG_CONFIG_HOME/helix` (which corresponds to `/usr/local/.config/helix`), either under `~/.config/helix`, or both.
`~/.config/helix` is outside of `/usr/local` so I should not touch it. Crew is not happy with `/usr/local/.config/helix` because it is not "FHS3 compliant".
This is why I am unsure of where to put this runtime folder. Currently I use `/usr/local/var/lib/helix`.
To allow helix to find its runtime folder I set the environment variable HELIX_RUNTIME.

satmandu: There is a FHS override flag you can put in the package file.

ncdm-stldr: Ok. I use the `no_fhs` flag now. The script is also simpler.

packages/helix_editor.rb
ncdm-stldr: I removed this code now. I do not need to set the environment variable anymore.

packages/helix_editor.rb
ncdm-stldr: I tried 'hx --version'. However it seems the self.check method is never called. I don't really understand why.

ncdm-stldr: I tried 'hx --version'. However it seems the self.check method is never called. I don't really understand why.

satmandu: self.check is only used during `crew build`, not `crew install -s`...

uberhacker: Sorry @ncdm-stldr:
I probably should have explained better how to do it.  Something like below:
```
def self.check
  hx_ver = `hx --version 2>&1`
  return (hx_ver.empty? || hx_ver.include?('No such file or directory') || hx_ver.include?('not found')) ? false : true
end
```

ncdm-stldr: Ok, I tried a `crew build` and saw it was indeed called there. Thanks.

I tried something along the lines of the suggestion. However even if  self.check returns false, the build is not aborted. I think that I will remove the self.check method; it is not very important.

packages/helix_editor.rb
```diff
@@ -0,0 +1,53 @@
...
+      --release \
+      --locked \
+     "
+  end
+
+  def self.install
+    # Copy executable
+    helix_executable_dest_dir = "#{CREW_DEST_PREFIX}/bin"
+    FileUtils.mkdir_p("#{helix_executable_dest_dir}")
+    FileUtils.cp("./target/release/hx", "#{helix_executable_dest_dir}")
```
uberhacker, line 28: It might be better to use `FileUtils.install` to ensure the binary is executable.  Something like:
```
FileUtils.install './target/release/hx', helix_executable_dest_dir, mode: 0o755
```

packages/helix_editor.rb
```diff
@@ -0,0 +1,53 @@
...
+    # Copy runtime dir
+    helix_runtime_dest_dir = "#{CREW_DEST_DIR}#{ENV['XDG_CONFIG_HOME']}/helix"
```
uberhacker, line 30: This makes the assumption that environment variable `XDG_CONFIG_HOME` is set which might not be the case.  Probably should check and set to a default value if not set.

packages/helix_editor.rb
```diff
@@ -0,0 +1,53 @@
+require 'package'
+
+class Helix_editor < Package                 # The first character of the class name must be upper case
+  description 'A vim inspired editor with LSP support'
+  homepage 'https://helix-editor.com/'
+  version '22.12'
+  license 'MPL-2.0' # license of source
+  compatibility 'all'
+  source_url 'https://github.com/helix-editor/helix/archive/refs/tags/22.12.tar.gz'
```
uberhacker, line 9: A shortened version of the url, `https://github.com/helix-editor/helix/archive/22.12.tar.gz` should work too.

packages/helix_editor.rb
```diff
@@ -0,0 +1,53 @@
...
+  source_sha256 'edae8af46401b45c3e71c38b4fa99f931c4458127978ccd1b29aaae79331d972'   # Use the command "sha256sum"
+  
+  depends_on 'rust' => :build
```
uberhacker, line 12: Include `depends_on 'xdg_base'` when using the `XDG_CONFIG_HOME` environment variable.

packages/helix_editor.rb
```diff
@@ -0,0 +1,53 @@
...
+    FileUtils.mkdir_p("#{helix_executable_dest_dir}")
+    FileUtils.cp("./target/release/hx", "#{helix_executable_dest_dir}")
+    # Copy runtime dir
+    helix_runtime_dest_dir = "#{CREW_DEST_DIR}#{ENV['XDG_CONFIG_HOME']}/helix"
+    FileUtils.mkdir_p("#{helix_runtime_dest_dir}")
+    FileUtils.cp_r("./runtime", "#{helix_runtime_dest_dir}")
+  end
+
+  def self.check
+    puts "self.check is called"
```
uberhacker, line 36: The linter is going to scream about the double quotes so better use single quotes instead.

packages/helix_editor.rb
uberhacker: Try explicitly throwing an error if the test fails.

packages/helix_editor.rb
satmandu: When you're in your git directory, do `rubocop -c .rubocop.yml -A packages/packagefile.rb`

packages/helix_editor.rb
```diff
@@ -1,52 +1,62 @@
...
+  source_url 'file:///usr/local/home/helix/deleteme.tar.gz'
+  source_sha256 'c3a0e022a7bcd6b139ae7abf1e0dae33c8f9ec82dc680076d9e5ea6593b85bc6'
+
   depends_on 'rust' => :build
+  depends_on 'xdg_base'
 
   @no_fhs = true
-  
+
+  @xdg_config_home = ENV.fetch 'XDG_CONFIG_HOME', '/usr/local/.config'
```
uberhacker, line 19: Abstract `/usr/local` with `#{CREW_PREFIX}`?

packages/helix_editor.rb
```diff
@@ -1,52 +1,62 @@
...
+    puts 'cargo build \
       --release \
       --locked \
-     "
+     '
+  end
+
+  def self.check
+    # This will raise an error if "hx" is not a valid command
+    `hx --version`
```
uberhacker, line 32: At this stage, it will still be located at `./target/release/hx`, correct?

packages/helix_editor.rb
```diff
@@ -1,52 +1,62 @@
...
       Use the 'hx' command to start helix.
-    
-      Use 'hx --health' to see if helix detects its runtime and to see which LSP 
+
+      Use 'hx --health' to see if helix can find its runtime and to see which LSP
       servers are detected.
-    
-      To be able to load some themes, helix neefds to be started in a terminal it recognizes 
+
+      To be able to load some themes, helix needs to be started in a terminal it recognizes#{' '}
```
uberhacker, line 59: Why the `#{' '}`?

uberhacker (changes requested):
Please see comments.

packages/helix_editor.rb
ncdm-stldr: Ok. Yes it works with the raise error.
(I tried this once but I forgot to copy the helix_editor.rb at the correct location...)

packages/helix_editor.rb
ncdm-stldr: done

packages/helix_editor.rb
ncdm-stldr: I created one (see comment in code) but I don't know why: "error: could not find `Cargo.toml` in `/usr/local/tmp/crew/helix_editor.20230216163702.dir` or any parent directory".
By the way, what is the motive of using a shortened path here?

packages/helix_editor.rb
ncdm-stldr: ok done

packages/helix_editor.rb
ncdm-stldr: Ok I'll use rubocop now.

packages/helix_editor.rb
ncdm-stldr: indeed. done

packages/helix_editor.rb
uberhacker: Shorter is better?  No reason other than saving characters. :)

packages/helix_editor.rb
ncdm-stldr: At this stage, it seems to me that the files are located as if the installation was finished. So /usr/local/bin/hx should be valid I think.
I don't know if ./target/release/hx is still valid; I haven't tried to access it.

packages/helix_editor.rb
uberhacker: You may need to `FileUtils.touch 'Cargo.toml'` somewhere.

packages/helix_editor.rb
ncdm-stldr: I typed them by mistake. Removed now

packages/helix_editor.rb
ncdm-stldr: Maybe but why does it work with the original link? I don't get it.

packages/helix_editor.rb
ncdm-stldr: Or maybe I am wrong and hx was already installed when I self.check was executed? ^^

packages/helix_editor.rb
```diff
@@ -6,7 +6,7 @@ class Helix_editor < Package
   version '22.12'
   license 'MPL-2.0' # license of source
   compatibility 'all'
-  source_url 'https://github.com/helix-editor/helix/archive/22.12.tar.gz' #'https://t.ly/-vk7'
+  source_url 'https://t.ly/-vk7'
```
satmandu: Please don't use a shortened url.

You can either use 
```
source_url 'https://github.com/helix-editor/helix.git'
git_hashtag version
```
or
```
source_url 'https://github.com/helix-editor/helix/archive/22.12.tar.gz'
source_sha256 'edae8af46401b45c3e71c38b4fa99f931c4458127978ccd1b29aaae79331d972'
```

packages/helix_editor.rb
ncdm-stldr: You are right it is at `./target/release/hx` (what I said previously doesn't make sense).
I will only try './target/release/hx --version' then.

Would it be right to move the current code inside postinstall? I wanted to check the installation was working properly...

packages/helix_editor.rb
ncdm-stldr: Yes ok. I did it because of another comment.
I prefer the normal path as it is more user friendly (or reader friendly I should say).

satmandu: Shorter is better, but not at the expense of obfuscating the source...

packages/helix_editor.rb
ncdm-stldr: I removed the shortened link. The original link is more understandable and direct...

packages/helix_editor.rb
ncdm-stldr: Fixed self.check. I moved the old code in postinstall to print warning in case the installation doesn't work as expected.

packages/helix_editor.rb
ncdm-stldr: ok removed now

packages/helix_editor.rb
```diff
@@ -0,0 +1,78 @@
...
+     '
+  end
+
+  def self.install
+    # Copy executable
+    helix_executable_dest_dir = "#{CREW_DEST_PREFIX}/bin"
+    FileUtils.mkdir_p helix_executable_dest_dir.to_s
+    FileUtils.install ".#{@build_folder_suffix}/hx", helix_executable_dest_dir, mode: 0o755
+    # Copy runtime dir
+    helix_runtime_dest_dir = "#{CREW_DEST_DIR}#{@helix_runtime_dir}"
```
uberhacker, line 35: Better to store in `"#{CREW_DEST_PREFIX}/#{@helix_runtime_dir}"`?  This will be ignored otherwise, correct @satmandu?

uberhacker: Nevermind.  I think this is correct.

packages/helix_editor.rb
```diff
@@ -0,0 +1,78 @@
...
+  def self.check
+    # Ensure hx is executable
+    command_status = system ".#{@build_folder_suffix}/hx --version"
+    raise 'hx is not executable' unless command_status == true
+  end
+
+  def self.postinstall
+    # This will print a warning if "hx" is not a valid command
+    command_status = system 'hx --version'
+    puts 'Warning: hx is not path'.red unless command_status == true
```
uberhacker, line 49: `is not in PATH`?

packages/helix_editor.rb
```diff
@@ -0,0 +1,78 @@
...
+  version '22.12'
+  license 'MPL-2.0' # license of source
+  compatibility 'all'
+  source_url 'https://github.com/helix-editor/helix/archive/22.12.tar.gz'
+  source_sha256 'edae8af46401b45c3e71c38b4fa99f931c4458127978ccd1b29aaae79331d972' # Use the command "sha256sum"
+
+  depends_on 'rust' => :build
+  depends_on 'xdg_base'
+
+  @no_fhs = true
```
uberhacker, line 15: Only need `no_fhs`?

uberhacker:
@ncdm-stldr: Thank you for all the effort you are putting into this new package.  I tested this on x86_64 and it works great but there are a ton of files in the package.  We can probably safely remove the sources directory at `/usr/local/.config/helix/runtime/grammars/sources` and reduce the size dramatically.
```
$crew files helix_editor
...snip...
/usr/local/bin/hx

Total found: 15066
Disk usage: 788.57 MB
$ crew files helix_editor|grep /usr/local/.config/helix/runtime/grammars/sources|wc -l
14522
```

packages/helix_editor.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  @xdg_config_home = ENV.fetch 'XDG_CONFIG_HOME', "#{CREW_PREFIX}/.config"
+  @helix_runtime_dir = "#{@xdg_config_home}/helix"
+  @build_folder_suffix = '/target/release'
+
+  def self.build
+    puts 'Building. This may be long.'
+    system 'cargo build \
```
supechicken, line 23: If the build options are few we can put them in a single line

packages/helix_editor.rb
```diff
@@ -0,0 +1,78 @@
...
+    FileUtils.install ".#{@build_folder_suffix}/hx", helix_executable_dest_dir, mode: 0o755
+    # Copy runtime dir
+    helix_runtime_dest_dir = "#{CREW_DEST_DIR}#{@helix_runtime_dir}"
+    FileUtils.mkdir_p helix_runtime_dest_dir
+    FileUtils.cp_r './runtime', helix_runtime_dest_dir
+  end
+
+  def self.check
+    # Ensure hx is executable
+    command_status = system ".#{@build_folder_suffix}/hx --version"
```
supechicken, line 42: Use `system ..., exception: false` to return a boolean (in `crew`, `system` will raise an error on non-zero exit status by default)

packages/helix_editor.rb
```diff
@@ -0,0 +1,78 @@
...
+
+  def self.postinstall
+    # This will print a warning if "hx" is not a valid command
+    command_status = system 'hx --version'
+    puts 'Warning: hx is not path'.red unless command_status == true
+    # Check if helix can find its runtime path
+    command_output = `hx --health`
+    puts 'Warning: helix cannot find its runtime dir'.red unless command_output.include? @helix_runtime_dir
+
+    puts <<~EOT2.orange
```
uberhacker, line 54: We typically use `.lightblue` for messages, `.orange` for warnings and `.lightred` for errors.

packages/helix_editor.rb
ncdm-stldr: Maybe library shrinking / stripping and patchelf should be disabled?
The libraries are standalone and compiled during the build process, so patchelf is not useful I guess.
I don't really understand what the shrinking/stripping is exactly.

uberhacker: You can specify `no_shrink` or `no_strip`.  See `crew help prop no_shrink` or `crew help prop no_strip` for more information.  To display a list of all available properties, enter `crew prop`.

packages/helix_editor.rb
```diff
@@ -0,0 +1,83 @@
...
+      Use 'hx --health' to see if helix can find its runtime and to see which LSP
+      servers are detected.
+      Note that to be able to load some themes, helix needs to be started in a terminal it recognizes
+      as supporting true colors.
+    EOT2
+  end
+
+  def self.remove
+    # Some files in the directory are write protected...
+    FileUtils.rm_rf @helix_runtime_dir
```
uberhacker, line 70: Prompt to remove?

packages/helix_editor.rb
ncdm-stldr: yes. done.

packages/helix_editor.rb
ncdm-stldr: Got it. Done. I also changed code of self.postinstall to take account on this.

packages/helix_editor.rb
ncdm-stldr: Indeed. Done.

packages/helix_editor.rb
ncdm-stldr: ok. I added no_shrink and no_strip. I also added no_patchelf. Is patchelf required for libraries that are build with crew?

packages/helix_editor.rb
ncdm-stldr: Ok. Changed now.

packages/helix_editor.rb
ncdm-stldr: The user should not modify this folder. If the user wants to override the config, he should do it in `~/.config/helix` instead.

ncdm-stldr:
> @ncdm-stldr: Thank you for all the effort you are putting into this new package. I tested this on x86_64 and it works great but there are a ton of files in the package. We can probably safely remove the sources directory at `/usr/local/.config/helix/runtime/grammars/sources` and reduce the size dramatically.
> 
> ```
> $ crew files helix_editor
> ...snip...
> /usr/local/bin/hx
> 
> Total found: 15066
> Disk usage: 788.57 MB
> $ crew files helix_editor|grep /usr/local/.config/helix/runtime/grammars/sources|wc -l
> 14522
> ```

Yes definitely. I am glad you noticed this.
The self.install step now remove the sources before copying the runtime.

packages/helix_editor.rb
uberhacker: Just a note in case you didn't catch it.  All these properties do not have to be explicitly set to true,  You could also add them like below:
```
no_fhs
no_shrink
no_strip
no_patchelf
```

uberhacker: The `no_patchelf` can be used to bypass this process and is optional.

packages/helix_editor.rb
uberhacker: Is this safe to delete if the user decides to install the module later?  In other words, will they need to set up and configure the package again if this is removed?

uberhacker: If there are no additional files added to this folder then could those be included in the package meta data to be removed during `crew remove helix_editor`?

packages/helix_editor.rb
```diff
@@ -0,0 +1,83 @@
...
+    FileUtils.rm_rf './runtime/grammars/sources' # remove the sources
+    helix_runtime_dest_dir = "#{CREW_DEST_DIR}#{@helix_runtime_dir}"
+    FileUtils.mkdir_p helix_runtime_dest_dir
+    FileUtils.cp_r './runtime', helix_runtime_dest_dir
+  end
+
+  def self.check
+    # Ensure hx is executable
+    command_status = system ".#{@build_folder_suffix}/hx --version", exception: false
+    raise 'hx is not executable' unless command_status == true
```
uberhacker, line 44: Use `.lightred` to indicate an error?

packages/helix_editor.rb
```diff
@@ -0,0 +1,83 @@
...
+  end
+
+  def self.postinstall
+    # This will print a warning if "hx" is not a valid command
+    command_status = system 'hx --version', exception: false
+    puts 'Warning: hx is not in PATH'.lightred unless command_status == true
+    # Check if helix can find its runtime path
+    if command_status == true
```
uberhacker, line 52: I _think_ `if command_status` is all you would need?

packages/helix_editor.rb
ncdm-stldr: Ok. I get your point. 

I removed this code now. Crew already removes the files itself using the package metadata.

In fact, I added this code because some files in the runtime directory were write protected, and crew failed to remove the runtime folder properly. But these write protected files were in fact all in the source folder (in the `.git database`). Since the sources are not anymore part of the package I can remove this code safely.

packages/helix_editor.rb
ncdm-stldr: done

packages/helix_editor.rb
ncdm-stldr: done

packages/helix_editor.rb
ncdm-stldr: Good to know. I am not familiar with ruby. I removed the explicit assignment.
I still have no clue if I should set no_patchelf or not.
```
readelf -d hx | grep -i needed
 0x00000001 (NEEDED)                     Shared library: [libgcc_s.so.1]
 0x00000001 (NEEDED)                     Shared library: [libm.so.6]
 0x00000001 (NEEDED)                     Shared library: [libc.so.6]
 0x00000001 (NEEDED)                     Shared library: [ld-linux-armhf.so.3]
```

packages/helix_editor.rb
ncdm-stldr: Should I still promp the user in case crew left some files in `@helix_runtime_dir`?

uberhacker: It probably wouldn't hurt to add a prompt in case there is some residue left over that didn't get added to the package files.  If we know that no user data is added there then it would fine to exclude.

packages/helix_editor.rb
```diff
@@ -0,0 +1,81 @@
...
+  source_url 'https://github.com/helix-editor/helix/archive/22.12.tar.gz'
+  source_sha256 'edae8af46401b45c3e71c38b4fa99f931c4458127978ccd1b29aaae79331d972' # Use the command "sha256sum"
+
+  depends_on 'rust' => :build
+  depends_on 'xdg_base'
+
+  @no_fhs
+  @no_shrink
+  @no_strip
+  @no_patchelf
```
uberhacker, line 18: Actually, the `@` is not needed either.

packages/helix_editor.rb
```diff
@@ -0,0 +1,81 @@
...
+    FileUtils.rm_rf './runtime/grammars/sources' # remove the sources
+    helix_runtime_dest_dir = "#{CREW_DEST_DIR}#{@helix_runtime_dir}"
+    FileUtils.mkdir_p helix_runtime_dest_dir
+    FileUtils.cp_r './runtime', helix_runtime_dest_dir
+  end
+
+  def self.check
+    # Ensure hx is executable
+    command_status = system ".#{@build_folder_suffix}/hx --version", exception: false
+    raise 'hx is not executable'.lightred unless command_status == true
```
uberhacker, line 44: Only need `unless command_status`?

packages/helix_editor.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def self.postinstall
+    # This will print a warning if "hx" is not a valid command
+    command_status = system 'hx --version', exception: false
+    puts 'Warning: hx is not in PATH'.lightred unless command_status == true
```
uberhacker, line 50: Only need `unless command_status`?

packages/helix_editor.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  def self.install
+    # Copy executable
+    helix_executable_dest_dir = "#{CREW_DEST_PREFIX}/bin"
+    FileUtils.mkdir_p helix_executable_dest_dir.to_s
+    FileUtils.install ".#{@build_folder_suffix}/hx", helix_executable_dest_dir, mode: 0o755
+    # Copy runtime dir (without the sources)
+    FileUtils.rm_rf './runtime/grammars/sources' # remove the sources
+    helix_runtime_dest_dir = "#{CREW_DEST_DIR}#{@helix_runtime_dir}"
+    FileUtils.mkdir_p helix_runtime_dest_dir
```
uberhacker, line 37: `helix_runtime_dest_dir.to_s`?

packages/helix_editor.rb
```diff
@@ -0,0 +1,81 @@
...
+    FileUtils.cp_r './runtime', helix_runtime_dest_dir
```
uberhacker, line 38: `helix_runtime_dest_dir.to_s`?

packages/helix_editor.rb
```diff
@@ -0,0 +1,81 @@
...
+  def self.build
+    puts 'Building. This may be long.'
+    system 'cargo build --release --locked'
+  end
+
+  def self.install
+    # Copy executable
+    helix_executable_dest_dir = "#{CREW_DEST_PREFIX}/bin"
+    FileUtils.mkdir_p helix_executable_dest_dir.to_s
+    FileUtils.install ".#{@build_folder_suffix}/hx", helix_executable_dest_dir, mode: 0o755
```
uberhacker, line 33: `helix_executable_dest_dir.to_s`?

uberhacker (changes requested):
See comments.

packages/helix_editor.rb
uberhacker: Any ideas @satmandu?

packages/helix_editor.rb
uberhacker: The linter doesn't like it either:
```
packages/helix_editor.rb:15:3: W: Lint/Void: Variable @no_fhs used in void context.
  @no_fhs
  ^^^^^^^
packages/helix_editor.rb:16:3: W: Lint/Void: Variable @no_shrink used in void context.
  @no_shrink
  ^^^^^^^^^^
packages/helix_editor.rb:17:3: W: Lint/Void: Variable @no_strip used in void context.
  @no_strip
  ^^^^^^^^^
packages/helix_editor.rb:18:3: W: Lint/Void: Variable @no_patchelf used in void context.
  @no_patchelf
  ^^^^^^^^^^^^
```

packages/helix_editor.rb
ncdm-stldr: What do you have in mind?

packages/helix_editor.rb
ncdm-stldr: done

packages/helix_editor.rb
ncdm-stldr: done

packages/helix_editor.rb
ncdm-stldr: done

packages/helix_editor.rb
ncdm-stldr: done

packages/helix_editor.rb
ncdm-stldr: Oh sorry you meant the "== true" can be ommited. (I thought I needed an additional condition)

uberhacker: Yes, the ` == true` is what I meant.

ncdm-stldr: ok done

packages/helix_editor.rb
ncdm-stldr: done

packages/helix_editor.rb
ncdm-stldr: ok done

packages/helix_editor.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+    puts <<~EOT2.lightblue
+      Use the 'hx' command to start helix.
+      Use 'hx --health' to see if helix can find its runtime and to see which LSP servers are detected.
+      Note that to be able to load some themes, helix needs to be started in a terminal it recognizes
+      as supporting true colors.
+    EOT2
+  end
+
+  def ask_to_remove_user_defined_config_files_in(dir)
```
uberhacker, line 66: I like this idea.  This is a good candidate for a utility function for all packages.  We could even pass the prompt message as a parameter too.

uberhacker (approved):

uberhacker (approved):

PR #7930, opened 2023-02-13
Update gimp and dependencies

uberhacker:
Reverting to the stable version.

- gimp 2.99.14 => 2.10.32
- libavif 0.10.1-1 => 0.11.1
- libffi 3.4.2-1 => 3.4.4
- Add py2_pycairo build dependency

Tested on all architectures.


satmandu:
This is technically a downgrade. The 2.99 series is based upon gtk3, and the 2.10 series is based upon gtk2.

There is for instance no native Wayland support in gtk2...

uberhacker:
Stick with the "unstable" release.

PR #7943, opened 2023-02-15
Crew remove: Do not remove outside CREW_PREFIX hierarchy

satmandu:
- Only store files in `CREW_PREFIX` during package builds.
- Do not (attempt to) remove files or directories from an installed package outside of `CREW_PREFIX`  during `crew remove packagename`.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=crew_remove CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1616,7 +1616,8 @@ def remove(pkgName)
 
       # remove all directories installed by the package
       File.foreach("meta/#{pkgName}.directorylist", chomp: true) do |line|
-        next unless Dir.exist?(line) && Dir.empty?(line)
+        puts "directorylist contains #{line}".lightred if @opt_verbose && !line.include?(CREW_PREFIX)
+        next unless Dir.exist?(line) && Dir.empty?(line) && line.include?(CREW_PREFIX)
 
         puts "Removing directory #{line}".lightred if @opt_verbose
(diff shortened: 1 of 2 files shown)


uberhacker:
This is not really fixing a bug.  I think what it will end up doing is hiding problems with packages that are not built correctly.  We shouldn't be building binaries that include paths that are not writable.

satmandu:
It's a security risk. Anybody could drop a file into dirlist or filelist and it will get removed with a crew remove...

satmandu:
There is a SEPARATE packaging issue of not putting files and folders outside CREW_PREFIX into directorylist and filelist, which should be handled in a separate PR.  But I think that we shouldn't do wrong things because of malformed filelists and directorylists...

satmandu:
@uberhacker This now also fixes package generation for new packages.

uberhacker (approved):
This is a better solution, for sure.  For a future PR, could this be Rubyized?

satmandu:
> This is a better solution, for sure. For a future PR, could this be Rubyized?

Perhaps... but `find` is really fast, so a ruby solution might slow down builds...

PR #7954, opened 2023-02-17
Add dmd package

uberhacker:
D is a general-purpose programming language with static typing, systems-level access, and C-like syntax. With the D Programming Language, write fast, read fast, and run fast.  See https://dlang.org/.  Tested on x86_64.

### DIFF
--- a/packages/dmd.rb
+++ b/packages/dmd.rb
@@ -1,0 +1,39 @@
+require 'package'
+
+class Dmd < Package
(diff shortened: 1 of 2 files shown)


uberhacker:
You might be asking the question "Why didn't you put phobos in a separate package"?  The TL;DR answer is because phobos requires the dmd sources to build.

satmandu (approved):
Amazing ...

PR #7960, opened 2023-02-18
Add dmd.conf, tools and manpages to dmd

uberhacker:
Tested partially.  Some tests fail to run but that is due mainly to bugs in the testing code.  No need to bump the version since I already overwrote the old binary.


satmandu (approved):

satmandu:
Linter failing...

PR #7963, opened 2023-02-19
Update librsvg and gdk_pixbuf

satmandu:
Builds properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=gnome_updates CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/gdk_pixbuf.rb
+++ b/packages/gdk_pixbuf.rb
@@ -3,24 +3,24 @@ require 'package'
 class Gdk_pixbuf < Package
   description 'GdkPixbuf is a library for image loading and manipulation.'
   homepage 'https://developer.gnome.org/gdk-pixbuf'
-  @_ver = '2.42.10-6550872'
+  @_ver = '2.42.10'
   version @_ver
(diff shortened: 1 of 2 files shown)


packages/gdk_pixbuf.rb
```diff
@@ -41,6 +41,7 @@ class Gdk_pixbuf < Package
   depends_on 'glibc' # R
   depends_on 'libpng' # R
   depends_on 'zlibpkg' # R
+
   gnome
```
uberhacker, line 45: Should we add the environment variables in postinstall to gdk_base.rb?

satmandu: They're already getting added to `env.d` though... so as long as a new shell is opened, the variables would be there.

uberhacker: Gotcha.

uberhacker (approved):

PR #7965, opened 2023-02-19
Move completion lines from ~/.bashrc to bash.d so it is removable

uberhacker:


uberhacker:
Not sure what happened here.

satmandu (approved):

PR #8001, opened 2023-02-26
Antlr4 4.7.1-1 => 4.12.0

uberhacker:
Works properly:
- [x] x86_64
- [x] i686
- [x] armv7l

### DIFF
--- a/packages/antlr4.rb
+++ b/packages/antlr4.rb
@@ -3,39 +3,35 @@ require 'package'
 class Antlr4 < Package
   description 'ANTLR (ANother Tool for Language Recognition) is a powerful parser generator for reading, processing, executing, or translating structured text or binary files.'
   homepage 'https://www.antlr.org/'
-  version '4.7.1-1'
+  version '4.12.0'
   license 'BSD'
   compatibility 'all'
-  source_url 'https://raw.githubusercontent.com/antlr/antlr4/4.7.1/README.md'
-  source_sha256 '70a58ea4c4f5ed23306313782bc13f36c3529d9a990e95ab273d5deed9286d4f'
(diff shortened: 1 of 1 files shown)


packages/antlr4.rb
```diff
@@ -3,39 +3,35 @@
...
-    system "install -Dm644 antlr-4.7.1-complete.jar #{CREW_DEST_LIB_PREFIX}/antlr-4.7.1-complete.jar"
-
-    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/etc/env.d/"
+  def self.build
     @antlrenv = <<~ANTLR_EOF
       # ANTLR (ANother Tool for Language Recognition) configuration
-      export CLASSPATH=\".:#{CREW_LIB_PREFIX}/antlr-4.7.1-complete.jar:\$CLASSPATH\"
-      alias antlr4=\"java -jar #{CREW_LIB_PREFIX}/antlr-4.7.1-complete.jar\"
-      alias grun=\"java org.antlr.v4.gui.TestRig\"
+      export CLASSPATH=".:#{CREW_PREFIX}/share/antlr/antlr-#{version}-complete.jar:$CLASSPATH"
```
satmandu, line 19: We don't need `export` for variable declarations in env.d, since those operate under `set -a`.

uberhacker: It doesn't hurt to include it for clarity but I can change if you insist.

uberhacker: I just tested without the export and it still works as expected.

satmandu: Fine either way.

satmandu (approved):

PR #8006, opened 2023-02-27
epiphany => 43.1

supechicken:
I need to figure out this error first... (installing `libiconv` does not help)
```
Run-time dependency libarchive found: NO (tried cmake)

meson.build:96:0: ERROR: Dependency lookup for libarchive with method 'pkgconfig' failed: Could not generate cargs for libarchive:
Package iconv was not found in the pkg-config search path.
Perhaps you should add the directory containing `iconv.pc'
Package 'iconv', required by 'libarchive', not found
```

### DIFF
--- a/packages/epiphany.rb
+++ b/packages/epiphany.rb
@@ -27,6 +27,7 @@ class Epiphany < Package
   depends_on 'at_spi2_core' # R
   depends_on 'docbook_xml' => :build
   depends_on 'freetype' => :build
+  depends_on 'gcc' # R
   depends_on 'gcr_3' # R
   depends_on 'gdk_pixbuf' # R
   depends_on 'glibc' # R
--- a/packages/epiphany.rb
+++ b/packages/epiphany.rb
@@ -8,42 +8,50 @@ class Epiphany < Package
   homepage 'https://wiki.gnome.org/Apps/Web'
   version '43.1'
   license 'GPL'
-  compatibility 'x86_64 aarch64 armv7l'
+  compatibility 'all'
   source_url 'https://gitlab.gnome.org/GNOME/epiphany.git'
   git_hashtag version
 
+  binary_url({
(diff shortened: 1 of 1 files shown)


satmandu:
iconv is part of glibc... are you building with an older glibc?

satmandu:
useful to check the meson build logs. looking at /usr/local/tmp/crew/epiphany.20230227220330.dir/builddir/meson-log/meson-log.txt...

```
Determining dependency 'libarchive' with pkg-config executable '/usr/local/bin/pkg-config'
env[PKG_CONFIG_PATH]:
Called `/usr/local/bin/pkg-config --modversion libarchive` -> 0
3.6.2
env[PKG_CONFIG_PATH]:
Called `/usr/local/bin/pkg-config --cflags libarchive` -> 1
```
Now let's try that command:
```
chronos@cheekon-x86_64 /usr/local/tmp/crew/epiphany.20230227220330.dir/builddir/meson-logs $ /usr/local/bin/pkg-config --cflags libarchive
Package iconv was not found in the pkg-config search path.
Perhaps you should add the directory containing `iconv.pc'
Package 'iconv', required by 'libarchive', not found
```
hmmm...

satmandu:
Arch is building libarchive with these flags: 
```
--without-xml2 \
      --without-nettle \
      --disable-static
```
NOTE ALSO that arch does this:
```
  sed -i "s/iconv //" "$pkgdir"/usr/lib/pkgconfig/libarchive.pc
```
Maybe we should at least be doing at least that last bit?

See https://github.com/archlinux/svntogit-packages/blob/packages/libarchive/trunk/PKGBUILD

Can you build libarchive with that locally and see if that fixes the issue?

satmandu (approved):

satmandu:
44.beta requires webkitgtk 2.40, so that's not an option at the moment...

uberhacker (approved):

PR #8017, opened 2023-02-28
Ccache 4.6.3 => 4.7.4

uberhacker:
Works properly:
- [x] armv7l
- [x] i686
- [x] x86_64

### DIFF
--- a/packages/ccache.rb
+++ b/packages/ccache.rb
@@ -3,34 +3,33 @@ require 'package'
 class Ccache < Package
   description 'Compiler cache that speeds up recompilation by caching previous compilations'
   homepage 'https://ccache.samba.org/'
-  @_ver = '4.6.3'
+  @_ver = '4.7.4'
   version @_ver
   license 'GPL-3 and LGPL-3'
   compatibility 'all'
   source_url "https://github.com/ccache/ccache/releases/download/v#{@_ver}/ccache-#{@_ver}.tar.xz"
-  source_sha256 '1e3a251bb112632553b8255a78661fe526c3a16598496d51128c32b218fd8b22'
+  source_sha256 'df0c64d15d3efaf0b4f6837dd6b1467e40eeaaa807db25ce79c3a08a46a84e36'
 
   binary_url({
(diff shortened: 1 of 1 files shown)


packages/ccache.rb
```diff
@@ -3,41 +3,46 @@
...
       #{CREW_CMAKE_OPTIONS} \
       -DCMAKE_INSTALL_SYSCONFDIR=#{CREW_PREFIX}/etc \
       -DZSTD_FROM_INTERNET=ON \
       -DHIREDIS_FROM_INTERNET=ON \
       .."
       system 'mold -run ninja'
     end
+    @ccacheenv = <<~CCACHEEOF
+      # ccache configuration
+      if ! [[ $PATH == *"ccache/bin"* ]]; then
```
supechicken, line 42: Use `!=`?

uberhacker: That might work but this is a regex evaluation.

packages/ccache.rb
```diff
@@ -3,41 +3,46 @@
...
+        export PATH="#{CREW_LIB_PREFIX}/ccache/bin:$PATH"
```
supechicken, line 43: `export` is not necessary on env files

packages/ccache.rb
supechicken: This is globbing... (regex uses `~=` and `!=` as operator)

uberhacker: Right, I just tested with `!=` and it works.  Rebuilding binaries... :sleeping:

satmandu: Does using `ccache` make the rebuild go faster? 🤣

uberhacker: You asked too late.  Already did the rebuilds.  How do you use this anyway?

satmandu (approved):

packages/ccache.rb
satmandu: You add the ccache dir to the path first, and then ccache intercepts calls to the compiler and maintains/checks a cache of compiled code for every compile request.

PR #8030, opened 2023-03-02
Reland of #7018: crew: Add multi-component support in one package

satmandu:
-This is a reland of @supechicken 's  https://github.com/chromebrew/chromebrew/pull/7018 .
- This is unfinished. Updates are needed to `def upgrade_command` and `def upgrade` to handle component upgrades (due to filelist conflicts between components and all).
- `preflight` is also not implemented for components. (_How do we want to handle that?_)
- `preremove` is set to only be triggered when `all` components are being removed. (_Is that what we want?_)
- The `copy_to_rootfs` function needs to be merged with the `install_files` so that `crew-mvdir` can be used. (Or `crew-mvdir` needs to have file list support.) Currently tar is used in `copy_to_rootfs` because it supports the use of file lists.
- `crew files` is currently broken: 
```
chronos@cheekon-x86_64 /usr/local/lib/crew/packages $ crew files glibc
glibc: The GNU C Library project provides the core libraries for GNU/Linux systems.glibc: no filelist available for component 'all' as it is installed as a part of component 'all'
```

Works properly:
- [ ] `x86_64`
- [ ] `i686`
- [ ] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=multi_component_rebase CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -689,7 +689,7 @@ def package_installed?(pkgName, componentName = 'all', output_if_not_installed =
   # that means all components were installed
   all_components_installed = File.exist?("#{CREW_META_PATH}/#{pkgName}.filelist")
 
-  unless all_components_installed || \
+  unless all_components_installed ||
          @device[:installed_packages].select {|pkg| pkg[:name] == pkgName } [0][:installedComponents]&.include?(componentName)
 
     puts "#{pkgName}: component #{componentName} isn't installed.".lightred if output_if_not_installed


supechicken:
I have several new thoughts on this:
- Can we use JSON on filelists to easily combine lists of different components in one file instead of `basePackage::component.filelist`?
- Would splitting binaries into component-based a good idea? So that we don't need to download the full package when installing components. Like this:
```ruby
class Base < Packages
  class Component1
    binary_url({
      x86_64: 'binary_component1'
    })
  end

  class Component2
    binary_url({
      x86_64: 'binary_component2'
    })
  end
  # base::all will download all binaries from components to form a full package
end
```
- Rewrite it from scratch (instead of keep patching and debugging on it) would make our work a lot easier :)

If we decided to go ahead with patching this, I am ready to help with this... (finally get some free time aside from school works, although this might be a bit late😅)

satmandu:
What do you mean by using json on the filelists?

At the moment we have components being split out manually into `_lib` and `_dev` packages from a `_build` package. It's not incredibly elegant, but it does give us the advantage of being able to use grep on the raw filelists.

One question with components is "How granular do we take it?" It could go down to the file level, or even to the level of having a document of exports from libraries, and dynamically pulling down the files from packages which are needed.

That would be insane but at peak theoretical efficiency in regards to saving space. ;)

PR #8051, opened 2023-03-07
gtk3 => 3.24.37

satmandu:
- Also sets `crew` to do the `sommelier` postinstall LAST.

Works properly:
- [x] `x86_64`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=gtk3 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/gtk3.rb
+++ b/packages/gtk3.rb
@@ -4,7 +4,6 @@ class Gtk3 < Package
   description 'GTK+ is a multi-platform toolkit for creating graphical user interfaces.'
   homepage 'https://developer.gnome.org/gtk3/3.0/'
   @_ver = '3.24.37'
-  @_ver_prelastdot = @_ver.rpartition('.')[0]
   version @_ver
   license 'LGPL-2.1'
   compatibility 'x86_64 aarch64 armv7l'
@@ -17,9 +16,9 @@ class Gtk3 < Package
      x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/gtk3/3.24.37_x86_64/gtk3-3.24.37-chromeos-x86_64.tar.zst'
   })
   binary_sha256({
-    aarch64: 'e201202002119422cc80a59f8868f6da5a0cef1933617b3f9ec97119adee209c',
-     armv7l: 'e201202002119422cc80a59f8868f6da5a0cef1933617b3f9ec97119adee209c',
-     x86_64: '56bb3d925fddd32df243f858a65f1aefb9296dacc854fda4c3f9854cd712ea81'
+    aarch64: '3dbf8f2b97f1ea7e0b58bfd07a95be8d8288674a92e084f26688b9e8233d96bc',
+     armv7l: '3dbf8f2b97f1ea7e0b58bfd07a95be8d8288674a92e084f26688b9e8233d96bc',
+     x86_64: 'ef6decef4d5d47c29a99777af6fc20ac9f9fcc3490a0974e512569cd3f75dc9c'
   })
 
   # L = Logical Dependency, R = Runtime Dependency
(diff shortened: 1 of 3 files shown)


packages/gtk3.rb
```diff
@@ -3,25 +3,23 @@
 class Gtk3 < Package
   description 'GTK+ is a multi-platform toolkit for creating graphical user interfaces.'
   homepage 'https://developer.gnome.org/gtk3/3.0/'
-  @_ver = '3.24.34-80d8970'
+  @_ver = '3.24.37'
   @_ver_prelastdot = @_ver.rpartition('.')[0]
```
uberhacker, line 7: Why do we need this?  It's not referenced anywhere in the package.

packages/gtk3.rb
```diff
@@ -62,13 +63,12 @@ class Gtk3 < Package
   depends_on 'libxrandr' # R
   depends_on 'mesa' => :build
   depends_on 'pango' # R
-  depends_on 'rest' # R
+  depends_on 'rest' => :build
   depends_on 'shared_mime_info' # L
   depends_on 'sommelier' # L
```
uberhacker, line 68: Can we put sommelier last?

uberhacker: Can we also add `depends_on 'xdg_base'` and replace the following:
```
    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/etc/gtk-3.0"
    File.write("#{CREW_DEST_PREFIX}/etc/gtk-3.0/settings.ini", @gtk3settings)
```
with:
```
    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/.config/gtk-3.0"
    File.write("#{CREW_DEST_PREFIX}/.config/gtk-3.0/settings.ini", @gtk3settings)
```
?

uberhacker: See https://developer-old.gnome.org/gtk3/stable/GtkSettings.html.

satmandu: I sort the list alphabetically with `sort -u` in my editor. Does it matter if sommelier is at the bottom? That's not going to affect the order of the installation of dependencies...

uberhacker: I'm just thinking about the postinstall message that will scroll past the screen from newly installed gui apps.

satmandu: > I'm just thinking about the postinstall message that will scroll past the screen from newly installed gui apps.

Yes, but if we want sommelier last, I think we should address that by modifying crew to make the sommelier post-install show up last, since installing sommelier last definitely shouldn't break anything.

packages/gtk3.rb
satmandu: It was once... I can get rid of it...

packages/gtk3.rb
satmandu: > Can we also add `depends_on 'xdg_base'` and replace the following:
> 
> ```
>     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/etc/gtk-3.0"
>     File.write("#{CREW_DEST_PREFIX}/etc/gtk-3.0/settings.ini", @gtk3settings)
> ```
> 
> with:
> 
> ```
>     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/.config/gtk-3.0"
>     File.write("#{CREW_DEST_PREFIX}/.config/gtk-3.0/settings.ini", @gtk3settings)
> ```
> 
> ?

Making that change... but xdg_base is already in the depends...

satmandu: Change made.

uberhacker (approved):

PR #8094, opened 2023-03-20
Curl => 8.0.1

satmandu:
- Also shifts dependencies back to `curl` from `libcurl`...
- Also updates `ca_certificates`, with `curl` built against the new version.

Works properly:
- [x] `x86_64`
- [X] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=curl8 CREW_TESTING=1 crew update
```


packages/curl.rb
```diff
@@ -3,10 +3,67 @@
...
+  depends_on 'openldap' # R
+  depends_on 'openssl' # R
+  depends_on 'python3' => :build
+  depends_on 'rust' => :build
+  depends_on 'valgrind' => :build
+  depends_on 'zlibpkg' # R
+  depends_on 'zstd' # R
+
+  no_patchelf
+  conflicts_ok
```
uberhacker, line 45: What are the conflicts?

satmandu: `libcurl` is the conflict during install. Which is fine, since we are replacing libcurl with curl...

packages/libcurl.rb
```diff
@@ -1,68 +1,18 @@
 require 'package'
+require_relative 'curl'
 
 class Libcurl < Package
   description 'Command line tool and library for transferring data with URLs.'
   homepage 'https://curl.se/'
-  @_ver = '7.88.1'
-  version @_ver
+  @curlversion = Curl.version
+  version @curlversion
```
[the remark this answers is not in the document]
satmandu: If I recall correctly, that didn't work in other packages. Hence this workaround...

packages/curl.rb
uberhacker: Uninstall libcurl prior to the build?  That would be better than adding `conflicts_ok`, right?

packages/libcurl.rb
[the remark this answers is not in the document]
uberhacker: I cannot see any reason why it shouldn't work.  What does "didn't work" mean?

packages/curl.rb
```diff
@@ -3,10 +3,67 @@
...
+  depends_on 'openldap' # R
+  depends_on 'openssl' # R
+  depends_on 'python3' => :build
+  depends_on 'rust' => :build
+  depends_on 'valgrind' => :build
+  depends_on 'zlibpkg' # R
+  depends_on 'zstd' # R
+
+  no_patchelf
+  conflicts_ok # Only conflict is with the older libcurl.
```
uberhacker, line 45: Remove?

packages/curl.rb
satmandu: Worry is that might leave a situation without a working curl. This avoids that.

Ideally we would have crew be able to specify which packages conflicts_ok will allow conflicts from...

packages/libcurl.rb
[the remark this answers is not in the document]
satmandu: Crew didn't recognize the version unless it was a string, either inside quotes, or via something like @_ver already specified earlier in the file.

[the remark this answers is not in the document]
satmandu: (I spent some time tearing my hair out over this at some point.)

packages/curl.rb
satmandu: Remove which?

uberhacker: The `conflicts_ok` line.  We don't need it if we `crew remove libcurl` first, right?

packages/libcurl.rb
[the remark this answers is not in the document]
satmandu: Yes! Changes made!

packages/curl.rb
uberhacker: I guess we can let it percolate awhile with this due to that issue but eventually, I'd like to remove it.

satmandu: Let me check...

PR #8101, opened 2023-03-22
Update x86_64 sha256sum hash for coreutils

uberhacker:
The eariler binary included kill and uptime because I built the new binary in the wrong directory. :face_with_spiral_eyes:


uberhacker:
> Should we maybe add a linter to check hashes?

This was just a case of not building the binary in the `release/x86_64` directory.

PR #8117, opened 2023-03-26
fswatch => 1.17.1

supechicken:
Tested on `x86_64`

### DIFF
--- a/packages/fswatch.rb
+++ b/packages/fswatch.rb
@@ -3,30 +3,14 @@ require 'package'
 class Fswatch < Package
   description 'fswatch is a file change monitor that receives notifications when the contents of the specified files or directories are modified.'
   homepage 'https://github.com/emcrisostomo/fswatch'
-  version '1.11.2'
+  version '1.17.1'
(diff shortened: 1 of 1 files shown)


satmandu (approved):

packages/fswatch.rb
```diff
@@ -3,30 +3,25 @@
 class Fswatch < Package
   description 'fswatch is a file change monitor that receives notifications when the contents of the specified files or directories are modified.'
   homepage 'https://github.com/emcrisostomo/fswatch'
-  version '1.11.2'
+  version '1.17.1'
   license 'GPL-3+'
   compatibility 'all'
```
uberhacker, line 8: Shouldn't this have changed to `compatibility 'x86_64 aarch64 armv7l'`?

supechicken: i686 (if any?) users should be able to build it from the source as the package still supports i686

uberhacker: Then why didn't we generate a binary? ;)

PR #8167, opened 2023-04-10
Change fake packages to use properties of the dependency package(s).

uberhacker:
This can probably be simplified further to call a function to set all the properties but there are certain instances where it makes sense to only use certain dependency package properties.  The plus side of doing this is fake versions bump automagically when the dependency package(s) are updated.

### DIFF
--- a/packages/a52.rb
+++ b/packages/a52.rb
@@ -1,12 +1,12 @@
 require 'package'
+require_relative 'liba52'
 
 class A52 < Package
-  description 'Legacy compatibility package for liba52'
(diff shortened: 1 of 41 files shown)


satmandu (approved):
Great idea!

satmandu:
Just need to fix the rubocop issue...

uberhacker:
> Just need to fix the rubocop issue...

Fixed.  Lesson learnt: `.to_s` is not needed for quoted string interpolation values.

PR #8196, opened 2023-04-18
Add a new entry for our custom version bump

supechicken:
**Not tested yet**

Currently, we use `@_ver` to store the upstream version and put our custom subversion inside of `version "#{@_ver}-..."`, but having two different variables for the version might confuse new contributors.

This PR adds a new entry called `revision` for our custom subversion and is optional for backward compatibility, here are some examples:
```ruby
# before
@_ver = '1.1'
version "#{@_ver}-1"

# after
version '1.1'
revision 1
```

I only applied it to some packages as there are so many packages that use `@_ver` for versioning 😅

```
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=add_release CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/package.rb
+++ b/lib/package.rb
@@ -5,7 +5,7 @@ require_relative 'package_helpers'
 require_relative 'selector'
 
 class Package
-  property :description, :homepage, :version, :license, :compatibility,
+  property :description, :homepage, :version, :revision, :license, :compatibility,
            :binary_url, :binary_sha256, :source_url, :source_sha256,
            :git_branch, :git_hashtag
 
@@ -216,6 +216,14 @@ class Package
     @dependencies.store(depName, [dep_tags, ver_check])
   end
 
+  def self.full_version
+    if @release && @revision != 0
+      return "#{@version}-#{@revision}"
+    else
+      return @version
+    end
+  end
+
   def self.get_url(architecture)
     if !@build_from_source && @binary_url && @binary_url.key?(architecture)
       return @binary_url[architecture]
@@ -245,23 +253,15 @@ class Package
   end
 
   def self.get_extract_dir
-    "#{name}.#{Time.now.utc.strftime('%Y%m%d%H%M%S')}.dir"
+    return "#{name}.#{Time.now.utc.strftime('%Y%m%d%H%M%S')}.dir"
   end
 
   def self.is_binary?(architecture)
-    if !@build_from_source && @binary_url && @binary_url.key?(architecture)
-      return true
-    else
-      return false
-    end
+    return (!@build_from_source && @binary_url && @binary_url.key?(architecture))
   end
 
   def self.is_source?(architecture)
-    if is_binary?(architecture) || is_fake?
-      return false
-    else
-      return true
-    end
+    return !(is_binary?(architecture) || is_fake?)
   end
 
   def self.system(*args, **opt_args)
--- a/packages/abseil_cpp.rb
+++ b/packages/abseil_cpp.rb
@@ -3,11 +3,10 @@ require 'package'
 class Abseil_cpp < Package
   description 'Abseil Common Libraries C++'
   homepage 'https://abseil.io/'
-  @_ver = '20200923.3'
-  version @_ver
+  version '20200923.3'
   license 'Apache-2.0'
   compatibility 'all'
-  source_url "https://github.com/abseil/abseil-cpp/archive/#{@_ver}.tar.gz"
+  source_url "https://github.com/abseil/abseil-cpp/archive/#{@version}.tar.gz"
   source_sha256 'ebe2ad1480d27383e4bf4211e2ca2ef312d5e6a09eba869fd2e8a5c5d553ded2'
 
   binary_url({
--- a/packages/apng2gif.rb
+++ b/packages/apng2gif.rb
@@ -3,11 +3,10 @@ require 'package'
 class Apng2gif < Package
   description 'Convert APNG animations into animated GIF format.'
   homepage 'https://sourceforge.net/projects/apng2gif/'
-  @_ver = '1.8'
-  version @_ver
+  version '1.8'
   license 'ZLIB'
   compatibility 'all'
-  source_url "https://sourceforge.net/projects/apng2gif/files/#{@_ver}/apng2gif-#{@_ver}-src.zip"
+  source_url "https://sourceforge.net/projects/apng2gif/files/#{@version}/apng2gif-#{@version}-src.zip"
   source_sha256 '9a07e386017dc696573cd7bc7b46b2575c06da0bc68c3c4f1c24a4b39cdedd4d'
 
   binary_url({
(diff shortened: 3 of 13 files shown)


lib/package.rb
```diff
@@ -216,6 +216,14 @@ def self.depends_on(dependency, ver_range = nil)
     @dependencies.store(depName, [dep_tags, ver_check])
   end
 
+  def self.full_version
+    if @release && @revision != 0
```
uberhacker, line 220: I think `if @revision` might be enough?

uberhacker (changes requested):
I like this idea!  Bump crew version?  Obviously, we'll need to update all the affected packages before this can be merged.  Another thing to consider is all the packages that use `@_ver` throughout the package need to use that version.

satmandu:
So we're also doing some tagging of python3 packages with the python3 version, and i think we're headed towards doing the same for perl.

But there are other version sub-tags which might make sense too. For instance, when building with LTO, packages complain when a library is built with a different version of GCC.

Might it make sense to also be able to tag packages with the GCC version used to build it? Obviously a revision tag would make sense for the rebuild, but it would also be nice to independently be able to scan packages to see what version of GCC was used to build it.

satmandu:
There's an ur-question here of how much package metadata should go in the package file, as opposed to being put elsewhere.

packages/apng2gif.rb
```diff
@@ -3,11 +3,10 @@
 class Apng2gif < Package
   description 'Convert APNG animations into animated GIF format.'
   homepage 'https://sourceforge.net/projects/apng2gif/'
-  @_ver = '1.8'
-  version @_ver
+  version '1.8'
   license 'ZLIB'
   compatibility 'all'
-  source_url "https://sourceforge.net/projects/apng2gif/files/#{@_ver}/apng2gif-#{@_ver}-src.zip"
+  source_url "https://sourceforge.net/projects/apng2gif/files/#{@version}/apng2gif-#{@version}-src.zip"
```
uberhacker, line 9: `@` not needed?

packages/abseil_cpp.rb
```diff
@@ -3,11 +3,10 @@
 class Abseil_cpp < Package
   description 'Abseil Common Libraries C++'
   homepage 'https://abseil.io/'
-  @_ver = '20200923.3'
-  version @_ver
+  version '20200923.3'
   license 'Apache-2.0'
   compatibility 'all'
-  source_url "https://github.com/abseil/abseil-cpp/archive/#{@_ver}.tar.gz"
+  source_url "https://github.com/abseil/abseil-cpp/archive/#{@version}.tar.gz"
```
Zopolis4, line 9: Wouldn't `#{version}` work here?

supechicken: Both `version` and `@version` should be identical

packages/apng2gif.rb
supechicken: Both `version` and `@version` should be identical, will remove `@` as it is simpler...

supechicken: Done

packages/abseil_cpp.rb
supechicken: (Removing `@` as it is simpler...)

lib/package.rb
supechicken: Yeah, it is redundant. Removed

supechicken:
Current progress on replacing `@_ver` in all packages: 119/549

It might take me a week to complete it... 💀

supechicken:
Thanks to @Zopolis4, now we have 362/549 left...

Zopolis4:
Speaking of-- it looks like you're manually removing `@_ver` from packages. What I've been doing is writing and using a series of sed scripts, which requires much less busywork. In the interests of saving time and making sure we don't double up on work, would it make more sense for me to just continue on as I am using my scripts? I plan to submit PRs for every `@_ver` usage that doesn't need this PR.

After that, I'll submit PRs that remove the usages of `@_ver` that do need this PR, and we can merge those after this is merged.

Is that ok? I don't want to step on your toes, but I also don't want you to have to go through ~400 packages manually when I can automate the process.

Zopolis4:
I've got some fallbacks for that, and I do test the packages.

packages/linuxheaders.rb
```diff
@@ -3,8 +3,8 @@
 class Linuxheaders < Package
   description 'Linux headers for Chrome OS.'
   homepage 'https://kernel.org/'
-  @version = CREW_KERNEL_VERSION == '4.14' ? "#{CREW_KERNEL_VERSION}-1" : CREW_KERNEL_VERSION
-  version @version
+  version = CREW_KERNEL_VERSION == '4.14' ? "#{CREW_KERNEL_VERSION}-1" : CREW_KERNEL_VERSION
+  version version
```
uberhacker, line 7:
```
revision 1
version CREW_KERNEL_VERSION == '4.14' ? "#{CREW_KERNEL_VERSION}-#{revision}" : CREW_KERNEL_VERSION
```
?

supechicken: Done

packages/linux_sources.rb
```diff
@@ -3,8 +3,8 @@
 class Linux_sources < Package
   description 'Sources for the Linux kernel'
   homepage 'https://kernel.org/'
-  @version = CREW_KERNEL_VERSION == '4.14' ? "#{CREW_KERNEL_VERSION}-1" : CREW_KERNEL_VERSION
-  version @version
+  version = CREW_KERNEL_VERSION == '4.14' ? "#{CREW_KERNEL_VERSION}-1" : CREW_KERNEL_VERSION
+  version version
```
uberhacker, line 7:
```
  revision 1
  version CREW_KERNEL_VERSION == '4.14' ? "#{CREW_KERNEL_VERSION}-#{revision}" : CREW_KERNEL_VERSION
```
?

lib/package.rb
```diff
@@ -245,23 +253,15 @@ def self.get_sha256(architecture)
...
+    return (!@build_from_source && @binary_url && @binary_url.key?(architecture))
   end
 
   def self.is_source?(architecture)
-    if is_binary?(architecture) || is_fake?
-      return false
-    else
-      return true
-    end
+    return !(is_binary?(architecture) || is_fake?)
```
Zopolis4, line 264: If these are just general improvements to the code,
does it make more sense to seperate these into another PR?

packages/gittools.rb
```diff
@@ -3,11 +3,12 @@
...
   description 'Tools for analyzing .git repositories'
   homepage 'https://github.com/internetwache/GitTools'
-  version '24eaef0d11e590643e65d188b017b49414d81cc2'
+  @commit = '24eaef0d11e590643e65d188b017b49414d81cc2'
+  version @commit[0...7]
   license 'MIT'
   compatibility 'all'
   source_url 'https://github.com/internetwache/GitTools.git'
-  git_hashtag version
+  git_hashtag @commit
```
Zopolis4, line 11: I'm conscious of the fact that this is still quite similar to using @_ver-- is there any other way to do this which doesn't introduce an additional variable?

PR #8209, opened 2023-04-21
qemu => 8.0.0

satmandu:
Builds properly:
- [x] `x86_64`
- [x] `armv7l`

### DIFF
--- a/packages/qemu.rb
+++ b/packages/qemu.rb
@@ -3,27 +3,28 @@ require 'package'
 class Qemu < Package
   description 'QEMU is a generic and open source machine emulator and virtualizer.'
   homepage 'http://www.qemu.org/'
-  @_ver = '7.1.0'
+  @_ver = '8.0.0'
   version @_ver
-  compatibility 'armv7l aarch64 x86_64'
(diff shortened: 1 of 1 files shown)


packages/qemu.rb
```diff
@@ -60,21 +61,22 @@ class Qemu < Package
...
     system "sed -i 's,<linux/usbdevice_fs.h>,\"linux/usbdevice_fs.h\",g' hw/usb/host-libusb.c"
   end
 
   def self.build
     FileUtils.mkdir_p 'build'
     Dir.chdir 'build' do
-      system "../configure #{CREW_OPTIONS.sub(/--target.*/, '')} \
+      system "mold -run ../configure #{CREW_OPTIONS.sub(/--target.*/, '')} \
         --enable-lto"
       system 'make || make -j1'
```
uberhacker, line 81: No `mold -run make`?

satmandu: Not necessarily needed. The LD flags should in theory make mold get used automatically. I try to use `mold -run` when it looks like that might not be happening.

uberhacker (approved):

PR #8216, opened 2023-04-23
Add shotwell package

uberhacker:
Shotwell is a personal photo manager.

Features
- Import from disk or camera
- Organize by time-based Events, Tags (keywords), Folders, and more
- View your photos in full-window or fullscreen mode
- Crop, rotate, color adjust, straighten, and enhance photos
- Slideshow
- Video and RAW photo support
- Share to major Web services, including Flickr, Tumblr and YouTube

See https://wiki.gnome.org/Apps/Shotwell.

Tested on:
- [x] aarch64
- [x] armv7l
- [x] x86_64

### DIFF
--- a/packages/shotwell.rb
+++ b/packages/shotwell.rb
@@ -1,0 +1,60 @@
+require 'package'
+
+class Shotwell < Package
+  description 'An open-source digital photo organiser for GNOME'
+  homepage 'https://wiki.gnome.org/Apps/Shotwell'
+  version '0.32.0'
+  license 'GPL-2.1+'
+  compatibility 'x86_64 aarch64 armv7l'
+  source_url 'https://download.gnome.org/sources/shotwell/0.32/shotwell-0.32.0.tar.xz'
+  source_sha256 '1772b2fa1e7a5479d528028bc9a303f2855e1d959a5b3c917922e63a98a90b19'
+
+  binary_url({
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/shotwell/0.32.0_armv7l/shotwell-0.32.0-chromeos-armv7l.tar.zst',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/shotwell/0.32.0_armv7l/shotwell-0.32.0-chromeos-armv7l.tar.zst',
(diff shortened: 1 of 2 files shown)


packages/shotwell.rb
```diff
@@ -0,0 +1,61 @@
...
+  depends_on 'libportal' # R
+  depends_on 'libraw' # R
+  depends_on 'libsecret' # R
+  depends_on 'libsoup' # R
+  depends_on 'libwebp' # R
+  depends_on 'libxml2' # R
+  depends_on 'json_glib' # R
+  depends_on 'sqlite' # R
+  depends_on 'webkit2gtk' # R
+  depends_on 'sommelier'
```
satmandu, line 44: Doesn't need sommelier since it has gtk3.

packages/shotwell.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  gnome
+
+  def self.build
+    system "meson setup #{CREW_MESON_OPTIONS} build"
+    system 'meson configure build'
+    system 'ninja -C build'
+  end
```
satmandu, line 52: s/ninja/#{CREW_NINJA}/

uberhacker: What if I want to use ninja instead of samu?

packages/shotwell.rb
uberhacker: The idea here is to make sure sommelier is installed last to make sure the post install instructions are not missed.

uberhacker: BTW, I could probably exclude a lot of the dependencies here that are installed implicitly. ;)

satmandu: > The idea here is to make sure sommelier is installed last so the post install instructions are not missed.

Remember how sommelier's post-install  is already set to be shown last because of code in `crew`?

satmandu: > BTW, I could probably exclude a lot of the dependencies here that are installed implicitly. ;)

I'm of course only talking about the non-library dependencies. 😅🫡

uberhacker: OK, fair enough.  Removed.

packages/shotwell.rb
satmandu: You can set `CREW_NINJA=ninja` in that case, no?

uberhacker: But then I have to set the environment variable.  Too much of a hassle. :)

satmandu (approved):

PR #8226, opened 2023-04-28
Update `crew whatprovides` to include packages not installed

uberhacker:
This update enables `crew whatprovides` to search for files in all packages; including packages that are not installed.

```
CREW_TESTING_REPO=https://github.com/uberhacker/chromebrew.git CREW_TESTING_BRANCH=update-crew-whatprovides CREW_TESTING=1 crew update
CREW_TESTING_ACCOUNT=uberhacker CREW_TESTING_BRANCH=update-crew-whatprovides CREW_TESTING=1 crew updatefiles
crew whatprovides /usr/local/bin/mysql
```


satmandu:
Zstd instead of gz?

satmandu:
Also, I'm not sure a single monolithic `filelist_all.gz` really works. This means the file has to be regenerated for every single PR, and every single PR needs to be rebased on top of every PR just merged, with the rebase updating the `filelist_all.gz` file.

It would be better for PRs to only handle the filelists for the packages in that PR, and put the file lists in a `metadata/filelist/package.filelist.zst` in a place in the git repo, so that an updated package PR will only touch the filelists for that PR. (Otherwise we're going to see the `filelist_all.gz` being iterated on every single package PR, which will lead to a LOT of cruft in the git history...)

This is probably also a good time to implement git garbage collection as part of `crew update` btw, since anything we do here will lead to more cruft in the .git folder on local installs...

uberhacker:
> Zstd instead of gz?

Does that work with `zgrep`?

supechicken:
I would suggest putting all filelist (without compressing) into a separate git repo, here are the advantages:

- Github will compress it for you during download
- History of filelist can be traced back (useful for the user to figure out when a file was removed)
- We can make use of the [Github's rest API](https://docs.github.com/en/rest/search?apiVersion=2022-11-28#about-search) to search through filelists without cloning the whole repo (which means we can implement it on our homepage)

satmandu:
I'm excited for putting the `meta` folder into git, because the next thing I want to put in there is a dump of dependencies for all binaries for each package, along the lines of what the `getrealdeps.sh` script builds... so that we can (more) easily tell if an update will break dependencies.

It would be nice to be able to avoid situations like in https://github.com/chromebrew/chromebrew/issues/8233 , where a package changes a library name, and the packages dependent upon that library suddenly break.

satmandu (approved):
Works for me! Thanks for making this happen!

PR #8236, opened 2023-05-02
OpenSSL 3.0.8 updates & rebuilds, plus update PAGER logic, pagers...

satmandu:
- The OpenSSL 3.0.8 package includes the libraries from OpenSSL 1.1.1t, so existing packages do not break.
- Mold rebuild has the MOLD_JOBS variable only set for systems with less than 4Gb of free RAM.
- Libraries are set to only build as shared, and instructions to build static libraries have been removed.
- Updated `less` and `most`.
- `crew_profile_base` now asks for `PAGER` selection in the postinstall.
- `crew` is adjusted to install openssl first in a set of upgrades. (Really though, the upgrades should walk a dependency hierarchy to figure this order out automatically.)
- All relevant packages in a default install have been updated.
- The following code will show what other packages need updating:
```
for i in *.so ; do lddtree $i 2>/dev/null | grep "libcrypto.so.1.1\|libssl.so.1.1" && crew whatprovides $i ; done | grep -v "Total\|=" | awk -F : '{print $1}' | sort -u
``` 
Here's a list of packages on my system that are not broken but need to be rebuilt against OpenSSL 3.x based on running that command:
```
audacious_plugins
bind
cups
ffmpeg
freerdp
freetds
ghostscript
graphicsmagick
gstreamer
libevent
libgit2
libgxps
libimobiledevice
libkmod
libspectre
libssh2
libvncserver
mpv
neon
qtwebengine
spice_gtk
srt
tesseract
```

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=openssl308 CREW_TESTING=1 crew update ; crew upgrade ; crew upgrade

```

### DIFF
--- a/packages/crew_profile_base.rb
+++ b/packages/crew_profile_base.rb
@@ -3,15 +3,14 @@ require 'package'
 class Crew_profile_base < Package
   description 'Crew-profile-base sets up Chromebrew\'s environment capabilities.'
   homepage 'https://github.com/chromebrew/crew-profile-base'
-  @_ver = '0.0.8'
-  version "#{@_ver}-1"
+  @_ver = '0.0.10'
+  version @_ver
   license 'GPL-3+'
   compatibility 'all'
-  source_url "https://github.com/chromebrew/crew-profile-base/archive/#{@_ver}.tar.gz"
-  source_sha256 '95642b0a8c7fe83ade2338ca68425ebf4598705633e8c0655cce8991b04697b7'
+  source_url "https://github.com/chromebrew/crew-profile-base/archive/refs/tags/#{@_ver}.tar.gz"
+  source_sha256 '73c60cabbae1050a55f031dc3c88738c3dbe52b516d5daca54f4d8890856a5f3'
 
   no_compile_needed
-  no_patchelf
 
   def self.install
     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/etc/"
--- a/packages/crew_profile_base.rb
+++ b/packages/crew_profile_base.rb
@@ -4,7 +4,7 @@ class Crew_profile_base < Package
   description 'Crew-profile-base sets up Chromebrew\'s environment capabilities.'
   homepage 'https://github.com/chromebrew/crew-profile-base'
   @_ver = '0.0.8'
-  version @_ver
+  version "#{@_ver}-1"
   license 'GPL-3+'
   compatibility 'all'
   source_url "https://github.com/chromebrew/crew-profile-base/archive/#{@_ver}.tar.gz"
@@ -70,5 +70,27 @@ class Crew_profile_base < Package
       # save changes
       File.write rc_path, rc_file.join("\n")
     end
+    puts 'Choose your default pager:'.yellow
+    @pager_options = [
+      { value: 'most', description: 'Most Pager' },
+      { value: 'more',  description: 'More Pager' },
+      { value: 'less',  description: 'Less Pager' }
+    ]
+
+    @pager_default = Selector.new(@pager_options).show_prompt
+    File.write 'pagerenv', <<~PAGER_ENV_EOF
+      # The user's preferred pager is set here by the crew_profile_base
+      # postinstall.
+
+      # PAGER from container PAGER variable is passed through into the
+      # CONTAINER_PAGER variable.
+      if [ -z "$CONTAINER_PAGER" ]; then
+        PAGER="#{@pager_default}"
+      else
+        PAGER="$CONTAINER_PAGER"
+      fi
+    PAGER_ENV_EOF
+    FileUtils.install 'pagerenv', "#{CREW_PREFIX}/etc/env.d/03-pager", mode: 0o644
+    puts "The default PAGER has been set to #{@pager_default}.".lightblue
   end
 end
--- a/packages/openssl.rb
+++ b/packages/openssl.rb
@@ -3,30 +3,27 @@ require 'package'
 class Openssl < Package
   description 'The Open Source toolkit for Secure Sockets Layer and Transport Layer Security'
   homepage 'https://www.openssl.org'
-  version '1.1.1t' # Do not use @_ver here, it will break the installer.
-  license 'openssl'
+  version '3.0.8' # Do not use @_ver here, it will break the installer.
+  license 'Apache-2.0'
   compatibility 'all'
-  source_url 'https://www.openssl.org/source/openssl-1.1.1t.tar.gz'
-  source_sha256 '8dee9b24bdb1dcbf0c3d1e9b02fb8f6bf22165e807f45adeb7c9677536859d3b'
+  source_url 'https://www.openssl.org/source/openssl-3.0.8.tar.gz'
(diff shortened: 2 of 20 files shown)


satmandu:
I would love to have a way to do this in an interactive manner for a set of important packages, i.e. `ruby, openssl`
```
  # Only upgrade openssl if openssl is in the upgrade list, as other
  # package upgrades may break until crew is rerun with the new
  # version of openssl.
  if to_be_upgraded.include?('openssl')
    to_be_upgraded = ['openssl']
    @rerun_upgrade = true
  end
```

supechicken:
No need as the OpenSSL shared library are loaded into the memory already before `ruby` runs. (which means it will not break the running `ruby` after the library was updated)

satmandu:
> No need as the OpenSSL shared library are loaded into the memory already before `ruby` runs. (which means it will not break the running `ruby` process whatever the library is being deleted/updated)

The rationale for loading OpenSSL independently (well, first) is so that other packages (built against the newer OpenSSL) don't break during postinstalls which require the newer OpenSSL.

Maybe I should adjust this to just pop OpenSSL to the front of the install queue.

satmandu:
Adjusted the `crew` logic to just add some code to make `openssl` install first during an upgrade.

packages/openssl.rb
```diff
@@ -47,15 +44,18 @@ class Openssl < Package
   @ARCH_CXX_LTO_FLAGS = "#{@arch_cxx_flags} -flto=auto"
 
   def self.build
+    @no_tests_target = `openssl version | awk '{print $2}'`.chomp == version.to_s ? 'no-tests' : ''
+    puts "@no_tests_target is #{@no_tests_target}"
```
uberhacker, line 48:
```
puts "@no_tests_target is #{@no_tests_target}" unless @no_tests_target == ''
```
?

packages/openssl.rb
```diff
@@ -65,10 +65,32 @@ def self.check
...
   end
 
   def self.install
     system "make DESTDIR=#{CREW_DEST_DIR} install_sw install_ssldirs"
+    # Extract OpenSSL 1.1.1t libraries for backwards compatibility purposes.
+    # Builds and rebuilds of packages against OpenSSL should automatically
+    # build against OpenSSL 3.x and not against OpenSSL 1.1.1x.
+    File.write 'openssl111_files', <<~EOF
+      usr/local/#{ARCH_LIB}/libcrypto.so.1.1
+      usr/local/#{ARCH_LIB}/libssl.so.1.1
```
uberhacker, line 78: What happens if `CREW_PREFIX` is something other than `/usr/local`?

packages/openssl.rb
```diff
@@ -65,10 +65,32 @@ def self.check
...
+    case ARCH
+    when 'aarch64', 'armv7l'
+      downloader 'https://gitlab.com/api/v4/projects/26210301/packages/generic/openssl/1.1.1t_armv7l/openssl-1.1.1t-chromeos-armv7l.tar.xz',
+                 '88a36d1539c7c01af1f5e469b64c2760f43126bb75c0e63b53d3d61c2a6fbe7f', 'openssl-1.1.1t-chromeos.tar.xz'
+    when 'i686'
+      downloader 'https://gitlab.com/api/v4/projects/26210301/packages/generic/openssl/1.1.1t_i686/openssl-1.1.1t-chromeos-i686.tar.xz',
+                 'ce98c1898e57df1cbcceab08912e219fb5f27b0e4585315f7babdf524fa844dc', 'openssl-1.1.1t-chromeos.tar.xz'
+    when 'x86_64'
+      downloader 'https://gitlab.com/api/v4/projects/26210301/packages/generic/openssl/1.1.1t_x86_64/openssl-1.1.1t-chromeos-x86_64.tar.xz',
+                 'd6583dc2c7566da33402cb8d7f9025189d65b8ba6300e752edcb0d74ebcb1f68', 'openssl-1.1.1t-chromeos.tar.xz'
```
uberhacker, line 90: There are a lot of redundant versions here.  Would it make sense to use `#{version}`?

packages/openssl.rb
satmandu: I think we have torn out the plumbing to enable that in any case... but open to code to just remove the first slash from CREW_PREFIX...

packages/openssl.rb
satmandu: Code suggestion to replace this block?

packages/openssl.rb
uberhacker: This is definitely an example of tearing out the plumbing. ;)

packages/openssl.rb
uberhacker:
```
    @legacy_version = '1.1.1t'
    case ARCH
    when 'aarch64', 'armv7l'
      downloader "https://gitlab.com/api/v4/projects/26210301/packages/generic/openssl/#{@legacy_version}_armv7l/openssl-#{@legacy_version}-chromeos-armv7l.tar.xz",
                 '88a36d1539c7c01af1f5e469b64c2760f43126bb75c0e63b53d3d61c2a6fbe7f', "openssl-#{@legacy_version}-chromeos.tar.xz"
    when 'i686'
      downloader "https://gitlab.com/api/v4/projects/26210301/packages/generic/openssl/#{@legacy_version}_i686/openssl-#{@legacy_version}-chromeos-i686.tar.xz",
                 'ce98c1898e57df1cbcceab08912e219fb5f27b0e4585315f7babdf524fa844dc', "openssl-#{@legacy_version}-chromeos.tar.xz"
    when 'x86_64'
      downloader "https://gitlab.com/api/v4/projects/26210301/packages/generic/openssl/#{@legacy_version}_x86_64/openssl-#{@legacy_version}-chromeos-x86_64.tar.xz",
                 'd6583dc2c7566da33402cb8d7f9025189d65b8ba6300e752edcb0d74ebcb1f68', "openssl-#{@legacy_version}-chromeos.tar.xz"
...
system "tar -xv --files-from #{@cur_dir}/openssl111_files -f #{@cur_dir}/openssl-#{@legacy_version}-chromeos.tar.xz"
```

packages/openssl.rb
satmandu: `CREW_PREFIX.chr` might fix this...

packages/openssl.rb
satmandu: Done.

packages/crew_profile_base.rb
```diff
@@ -4,7 +4,7 @@ class Crew_profile_base < Package
   description 'Crew-profile-base sets up Chromebrew\'s environment capabilities.'
   homepage 'https://github.com/chromebrew/crew-profile-base'
   @_ver = '0.0.8'
-  version @_ver
+  version "#{@_ver}-1"
```
uberhacker, line 7: Why not just increment from 0.0.8 to 0.0.9?  This is our package.

uberhacker (approved):
Nice job.  I couldn't find anything wrong with these updates.

PR #8261, opened 2023-05-16
Fix CREW_TESTING conditional logic

uberhacker:
Hopefully, this is the last fix needed. :)


satmandu (approved):

uberhacker:
I exited my docker container and launched again and it worked so it must have been a glitch.  Shouldn't matter if the environment variable is initialized or not because the condition would work as intended anyway.

PR #8311, opened 2023-05-30
Nautilus 44.0 => 44.2

uberhacker:
No changes to manifest files.

### DIFF
--- a/packages/nautilus.rb
+++ b/packages/nautilus.rb
@@ -6,22 +6,21 @@ require 'package'
 class Nautilus < Package
   description 'Default file manager for GNOME'
   homepage 'https://wiki.gnome.org/Apps/Files'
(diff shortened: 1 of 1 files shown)


packages/nautilus.rb
```diff
@@ -53,6 +52,7 @@ class Nautilus < Package
   depends_on 'pango' # R
   depends_on 'tracker3_miners'
   depends_on 'tracker3' # R
+  depends_on 'vulkan_headers' # R
```
satmandu, line 55: This is a runtime dependency? (Isn't this a build dependency?)

uberhacker: I'm not sure.  I just noticed an error when compiling.

satmandu (approved):

PR #8328, opened 2023-06-02
Repackaging of llvm into llvm_lib16 and llvm_dev16

satmandu:
- Needs testing with packages which indicate a `llvm` dependency but really only depend upon the `libLLVM-16.so` file.
- Remove `openmp` from the `gettext` runtime deps, since only `msgmerge` depends upon `openmp`. (I put `openmp` in `buildessential`, and maybe we can just not build `gettext` with `openmp` when we next build it...)

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=llvm_libs CREW_TESTING=1 crew update
```


satmandu:
Not sure this needs to be in packages.yaml?

packages/llvm_libs.rb
```diff
@@ -0,0 +1,44 @@
+require 'package'
+require_relative 'llvm'
+
+class Llvm_libs < Package
+  description 'LLVM Libraries'
+  homepage Llvm.homepage
+  version Llvm.version
+  # When upgrading llvm, be sure to upgrade llvm_libs in tandem.
+  puts "#{self} version differs from llvm version #{Llvm.version}".orange if version != Llvm.version.to_s
```
uberhacker, line 9: Will this ever print?  You're setting it on the line above.

packages/llvm_libs.rb
```diff
@@ -0,0 +1,44 @@
...
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/llvm_libs/16.0.4_armv7l/llvm_libs-16.0.4-chromeos-armv7l.tar.zst',
+       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/llvm_libs/16.0.4_i686/llvm_libs-16.0.4-chromeos-i686.tar.zst',
+     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/llvm_libs/16.0.4_x86_64/llvm_libs-16.0.4-chromeos-x86_64.tar.zst'
+  })
+  binary_sha256({
+    aarch64: 'b6e5091f724e334ca371ac9b6e391496b2875b56e927f5f8cebccdf7e4f899bb',
+     armv7l: 'b6e5091f724e334ca371ac9b6e391496b2875b56e927f5f8cebccdf7e4f899bb',
+       i686: '35cc9ee58f6fb140cc3babfe96c0f472c04949f01439b06eee455cac3a9e647b',
+     x86_64: '7c2a38326d08d6d69ccad8c47906d67956111b784359ca50dd67b7735d355027'
+  })
```
uberhacker, line 25: Binaries not necessary?

packages/llvm_libs.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  depends_on 'llvm' => :build
+  conflicts_ok
```
uberhacker, line 28: Wouldn't a better approach be to create the llvm_libs.filelist from llvm during the build?  That way there will be no conflicts.

packages/llvm_libs.rb
```diff
@@ -0,0 +1,44 @@
...
+    abort 'File list for llvm does not exist!'.lightred unless File.file?(@filelist_path)
+    @filelist = File.readlines(@filelist_path, chomp: true).sort
+
+    @filelist.each do |filename|
+      next unless filename.include?(CREW_LIB_PREFIX) && filename.include?('.so') && filename.include?('libLLVM')
+
+      @destpath = "#{CREW_DEST_DIR.chomp('/')}#{filename}"
+      @filename_target = File.realpath(filename)
+      FileUtils.install @filename_target, @destpath
+    end
```
uberhacker, line 42: No need to extra out if llvm generates llvm_libs.filelist?

packages/llvm_libs.rb
satmandu: Yes, crew can't generate two binary packages per architecture from one build now.

What I did is possible now with no changes.

packages/llvm_libs.rb
satmandu: Useful during installation to save space...

packages/llvm_libs.rb
uberhacker: Right, it can't now but we can update to allow it for the llvm package.

packages/llvm_libs.rb
uberhacker: But the problem is when llvm is updated, these need to be updated separately.

packages/llvm_libs.rb
satmandu: Ah!

satmandu: fixed.

packages/llvm_libs.rb
satmandu: Yes, but the rationale is that during installs, you don't need the extra Gb of temporary space to do the install.

We already have to rebuild openmp when building llvm...

packages/llvm_libs.rb
satmandu: I agree that once someone updates `crew` to handle that, this approach can be deprecated...

packages/llvm_libs.rb
satmandu: This goes back to the ur-question of someone modifying `crew` to generate multiple binary packages from a single build, which hasn't happened in the past year in a mergable form.

packages/llvm_bin.rb
```diff
@@ -0,0 +1,44 @@
+require 'package'
+require_relative 'llvm_build'
+
+class Llvm_bin < Package
+  description 'LLVM Libraries'
```
uberhacker: `LLVM Binaries`?

packages/llvm_build.rb
```diff
@@ -0,0 +1,207 @@
...
+version=$(gcc -dumpversion)
+gnuc_lib=#{CREW_LIB_PREFIX}/gcc/${machine}/${version}
+clang -B ${gnuc_lib} -L ${gnuc_lib} \"$@\"' > builddir/clc"
+      system "echo '#!/bin/bash
+machine=$(gcc -dumpmachine)
+version=$(gcc -dumpversion)
+cxx_sys=#{CREW_PREFIX}/include/c++/${version}
+cxx_inc=#{CREW_PREFIX}/include/c++/${version}/${machine}
+gnuc_lib=#{CREW_LIB_PREFIX}/gcc/${machine}/${version}
+clang++ -fPIC  -rtlib=compiler-rt -stdlib=libc++ -cxx-isystem ${cxx_sys} -I ${cxx_inc} -B ${gnuc_lib} -L ${gnuc_lib} \"$@\"' > builddir/clc++"
```
uberhacker, line 125: Convert wrapper scripts to use `<<~EOF ... EOF` instead of `system "echo ...`?

satmandu: Good idea.

packages/gettext.rb
```diff
@@ -32,12 +32,10 @@ class Gettext < Package
   depends_on 'openjdk8' => :build
   depends_on 'openmp' => :build
   depends_on 'zlibpkg' # R
-  depends_on 'openmp' # R
+  # depends_on 'openmp' # R This is only needed for the msgmerge binary.
```
uberhacker, line 35: Why is this listed twice?

packages/llvm_lib16.rb
```diff
@@ -0,0 +1,53 @@
...
+  depends_on 'libffi' # R
+  depends_on 'libxml2' # R
+  depends_on 'ncurses' # R
+  depends_on 'zlibpkg' # R
+  depends_on 'zstd' # R
+
+  conflicts_ok
+
+  def self.install
+    puts 'Installing llvm_build16 to pull files for build...'.lightblue
```
uberhacker, line 40: Isn't the main goal to save space so we don't have to install the huge llvm_build16 package?

packages/llvm_dev16.rb
```diff
@@ -0,0 +1,59 @@
...
+  depends_on 'llvm_lib16' # R
+  depends_on 'ncurses' # R
+  depends_on 'xzutils' # R
+  depends_on 'zlibpkg' # R
+  depends_on 'zstd' # R
+
+  conflicts_ok
+
+  def self.install
+    puts 'Installing llvm_build16 to pull files for build...'.lightblue
```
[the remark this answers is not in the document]
satmandu: Yes. one only needs `llvm_build16` as a build_dep. It isn't normally needed.

packages/gettext.rb
satmandu: Removed!

uberhacker (approved):

PR #8339, opened 2023-06-06
LHa 6f6cbc1 -> 2695022

Zopolis4:
LHa 6f6cbc1 -> 2695022

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=update-lha CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/lha.rb
+++ b/packages/lha.rb
@@ -9,6 +9,21 @@ class Lha < Package
   source_url 'https://github.com/jca02266/lha.git'
   git_hashtag '26950220c9c7590fd603ecaa54a12a52371affed'
 
+  binary_url({
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/lha/2695022_armv7l/lha-2695022-chromeos-armv7l.tar.zst',
+     armv7l: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/lha/2695022_armv7l/lha-2695022-chromeos-armv7l.tar.zst',
+       i686: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/lha/2695022_i686/lha-2695022-chromeos-i686.tar.zst',
+     x86_64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/lha/2695022_x86_64/lha-2695022-chromeos-x86_64.tar.zst'
+  })
+  binary_sha256({
+    aarch64: 'a9e93c81eef64714252d453f98843a2d6b00c43bbcc9b4c67c3aabeb8365db50',
+     armv7l: 'a9e93c81eef64714252d453f98843a2d6b00c43bbcc9b4c67c3aabeb8365db50',
+       i686: 'f3f0e76166650dafe0874af2b10b650f3e826503e1be6cacae4275d03fe6a918',
+     x86_64: 'b5ac02857046c1673acdb928c7a189ee350d9b0babdaaea1661c39fc81328590'
+  })
+
+  depends_on 'glibc_lib' # R
+
   def self.build
     system 'autoreconf -sif'
     system "./configure #{CREW_OPTIONS}"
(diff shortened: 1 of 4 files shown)


packages/lha.rb
```diff
@@ -3,34 +3,15 @@
...
-     armv7l: '8fa5dd3a57a9da3ffe245f1417f746da3c6d712e4436dcad67a71ec34e1c2e94',
-       i686: '8b5d6ed5ffb6118b36ab7a1ab93519003694fc18b5203ca4f885447aa023965b',
-     x86_64: 'e1894e33a4ed0932a493d049007e8d4ac70d94ec5e46247a04d448f23bf32c5d'
-  })
-
-  def self.patch
-    system "sed -i 's/lha_LDADD.*/& -liconv/' src/Makefile.am"
-  end
+  source_url 'https://github.com/jca02266/lha.git'
+  git_hashtag 'ee7fbac2c6a10530d5784db8f0c5aef86948dee0'
```
uberhacker, line 10: Latest hash is 26950220c9c7590fd603ecaa54a12a52371affed.  Change version to `20230328`?

Zopolis4: The latest timestamp is `20220213`-- jca02266/lha@ee7fbac2c6a10530d5784db8f0c5aef86948dee0.

uberhacker: That's not what I'm seeing.  Where are you getting your information?

Zopolis4: The timestamp in the configure.ac file at the root of the repository.

uberhacker: Take a look at the project page and notice the `src` directory was changed 3 months ago.

Zopolis4: Yeah, but that's not a version bump.

uberhacker: The version bump was missed.  We can use sed to change in configure.ac.

Zopolis4: The timestamp bump there is linked to a [tag](https://github.com/jca02266/lha/tags) and a [github release](https://github.com/jca02266/lha/releases), so I felt it made more sense to stick with that, as its an actual release by all metrics.

uberhacker: The version you suggest is not a tag or release.  My guess is they forgot or neglected to bump the release and tag it.

Zopolis4: Ah.

Zopolis4: imo we should still stick to the latest timestamp, as distributing a package with a version which doesn't exist could cause confusion.

uberhacker: The version can be the git hash.  We have several packages that have repos with no tags or releases so that is the only alternative.

uberhacker: Thoughts @satmandu?

satmandu: The latest commit is fine ...

Zopolis4: Updated.

packages/lha.rb
```diff
@@ -3,34 +3,15 @@
...
-  })
-
-  def self.patch
-    system "sed -i 's/lha_LDADD.*/& -liconv/' src/Makefile.am"
-  end
+  source_url 'https://github.com/jca02266/lha.git'
+  git_hashtag '26950220c9c7590fd603ecaa54a12a52371affed'
 
   def self.build
     system 'autoreconf -sif'
```
uberhacker, line 13: Bump release version to match the latest hash?
```
def self.patch
  system "sed -i 's,20220213,20230328,' configure.ac"
end
```

Zopolis4: I updated this so there wouldn't be a patch section :p

At any rate, I'd rather not have a downstream patch, espescially one which upstream isin't interested in carrying.

satmandu:
Do you need help setting up the builders so you can generate binaries?

uberhacker (approved):

Zopolis4:
Sorry I didn't get around to it, thanks for adding binaries!

PR #8355, opened 2023-06-09
Refactor workflow files

Zopolis4:
Required for some of my future ideas, mainly proper CI testing for builds and `install.sh`, as well as possibly automated building.


Zopolis4:
Apparently there is simply no way to use environment variables in a `uses` call, see https://github.com/orgs/community/discussions/25246.

Zopolis4:
I feel obliged to point out how [I gradually went insane](https://github.com/Zopolis4/chromebrew/actions) debugging issues related to this, but it works now.

.github/workflows/Handoff.yml
```diff
@@ -0,0 +1,57 @@
...
+        uses: tj-actions/changed-files@v36
+      - name: Get extensions of changed files
+        id: changed-file-extensions
+        run: |
+          workflow=
+          for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+            ext="${file##*.}"
+            # In order for files such as bin/crew to be recognised
+            type="$(file -b $file | cut -d' ' -f1)"
+            [[ "$type" == "Ruby" ]] && ext="rb"
```
uberhacker, line 26: I think you better keep this code block:
```
            if ! [[ "$ext" =~ ^(md|rb|sh|yml|yaml)$ ]]; then
              type="$(file -b "$file" | cut -d' ' -f1)"
              case $type in
                Bourne-Again)
                  ext="sh"
                  ;;
                HTML)
                  ext="md"
                  ;;
                Ruby)
                  ext="rb"
                  ;;
              esac
            fi
```

Zopolis4: Why? All the bash files we have end in `.sh`, and all the markdown files we have end in `.md`. It's only ruby which has files that don't end in `.rb`, and I kept that functionality intact.

uberhacker: Think about it.  It's possible any one of these types may not have an extension.  This was inserted for a reason.

Zopolis4: It's possible, yes, but its currently not true. If for some reason we have another markdown file added without an extension, or a shell script without an extension, and there is a reason for them to not have an extension, I'll add the relevant part of the block back in.

uberhacker: If you can avoid force pushing changes that would be good too.  It removes all comment history for previous commits.  We can squash the commits in the end prior to merging.

Zopolis4: We aren't missing those checks anymore, though.

Force pushing commits is kinda how I use git, but I'll try to do it less often.

satmandu: > This logic was added because we were missing checks for those types of files.  Better leave it in.  Thoughts @satmandu?

As long as `file` is able to identify file types, I have no issues with eliminating the use of file extensions to identify file types.

Zopolis4: It's the other way around-- we used to have code for using file as well as using file extensions, I've removed the file code (apart from the Ruby code which we still need), and kept the extension code.

The issue with using file is that it can't identify YAML files, as it just identifies them as ASCII.

I'm happy to use either the file code or the file extension code, but I don't want both because then it feels messy and unessecary.

uberhacker:
Is this ready?  Thoughts @satmandu?

Zopolis4:
Yeah, it's ready.

satmandu:
Let's see how it works

satmandu (approved):

PR #8356, opened 2023-06-09
Run Layout/SpaceInsideHashLiteralBraces cop on tree

Zopolis4:
Looks nicer, is more compact, and will help with #7082.

Tested and working on x86_64.

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=literallyaspaceinsidehashbraces CREW_TESTING=1 crew update
```


satmandu:
Do we need to add this cop to the standard rubocop `.rubocop.yml` file? (in which case we need to update the `ruby_rubocop.rb` package too.)

Zopolis4:
It's already enabled by default there, but the way our rubocop actions are set up now means that if a new cop gets added we won't get it applied across the entire tree.

Which is fine, I can just run rubocop locally every so often and add the new results, like what I'm doing here.

satmandu (approved):

PR #8360, opened 2023-06-09
Remove file:/// binary_urls from packages (and a bintray url)

Zopolis4:
How did those sneak in there?


uberhacker (approved):
Good catch!

PR #8368, opened 2023-06-11
Update unicode_cldr to 43.0

supechicken:


satmandu (commented):
Does this change filelists in the manifest directory?

packages/unicode_cldr.rb
```diff
@@ -2,27 +2,14 @@
 
 class Unicode_cldr < Package
   description 'Unicode Common Locale Data Repository'
-  homepage 'http://cldr.unicode.org/'
-  @_ver = '38.0'
+  homepage 'http://cldr.unicode.org'
```
uberhacker, line 5: Change to `https://cldr.unicode.org/`?

supechicken:
> Does this change filelists in the manifest directory?

Does it have filelist before? (as I know only packages with non-prebuilt sources have filelist generated)

packages/unicode_cldr.rb
supechicken: Done

satmandu:
> > Does this change filelists in the manifest directory?
> 
> Does it have filelist before? (as I know only packages with non-prebuilt sources have filelist generated)

Didn't realize binaries weren't needed!

satmandu (approved):

PR #7472, opened 2022-10-08
add missing runtime deps to ALL core packages, add getrealdeps script

satmandu:
- Also adds the `getrealdeps` script to tools.
- (I'd note that the script doesn't actually check to see what binaries are being run from the package script itself as those could themselves be considered install or build-time dependencies.)

image

```
chronos@cheekon-x86_64 /usr/local/lib/crew/packages $ crew deps -t uchardet
Walking through dependencies recursively, this may take a while...

 : satisfied dependency
 : build dependency
 : runtime dependency

└───┬─uchardet
    ├───┬─gcc
    │   ├─────glibc
    │   ├─────gmp
    │   ├─────isl
    │   ├─────libssp
    │   ├───┬─mpc
    │   │   └─────mpfr
    │   ├─────zlibpkg
    │   └───┬─zstd
    │       └─────musl_zstd
    └─────glibc
```

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

### Run the following to get this pull request's changes locally for testing.
```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=gcc_fix CREW_TESTING=1 crew update
```

PR #7402, opened 2022-09-13
Add filecmd dependency to glibc.rb

Zopolis4:
Fixes creating symlinks on a fresh install.
See line 597 and 606.

### Run the following to get this pull request's changes locally for testing.
```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=fix2 CREW_TESTING=1 crew update
```

PR #8369, opened 2023-06-12
Re-add filecmd dependency to glibc

Zopolis4:
Added in #7402, removed in #7472. 

Not sure why it was removed, and I didnt notice it was removed at the time.
Until now, when I encountered the error again today while testing.

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=fix2v2eb CREW_TESTING=1 crew update
```


satmandu (approved):
Thanks for catching this!

PR #8377, opened 2023-06-12
Do not remove files in other packages during package removal.

### LINKED ISSUE #8375: 🐞 is_fake package issues
1.`crew reinstall` doesn't work. (as reported by @Zopolis4 on slack)
2. upgrades to packages which have been converted to `is_fake` should happen before all other package upgrades, to prevent the removal process of the prior non-`is_fake` package causing the deletion of files installed during the upgrade. (as reported by @uberhacker on slack)
e.g.,
1. `packageA` is installed and has a `packageA.filelist`
2. `packageA` converted to an `is_fake` package depending on `packageA_lib` and `packageA_dev`, but the new `is_fake` `packageA` has an empty filelist.
3. `crew upgrade` is run, which causes:
    - `packageA_lib` and `packageA_dev` get installed, replacing files from `packageA`
    - we now get to the `packageA` upgrade, which causes first a `packageA` removal, i.e., removal of all files in `packageA.filelist`. This removes the files just installed from `packageA_lib` and `packageA_dev`... though the system still thinks those files are installed from `packageA_lib`, and `packageA_dev`, and the filelists for those two packages are left untouched.
   - The new `packageA` is now marked as being installed (with no files in its filelist.)

We need to deal with this... maybe one way would be to not remove filelists of packages being converted to `is_fake`, or just not delete files that exist in other filelists?

satmandu:
Fixes #8375 
- Does NOT fix reinstall of `is_fake` packages.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=crew_is_fake CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1082,7 +1082,7 @@ def determine_conflicts(dir, pkg)
     conflicts = conflictscmd.gsub(/(\.filelist|#{CREW_META_PATH})/, '').split("\n")
     conflicts.reject!(&:empty?)
   end
-  unless conflicts.empty?
+  if conflicts.any?
     puts 'There is a conflict with the same file in another package:'.orange
     puts conflicts.to_s.orange
   end
@@ -1160,8 +1160,8 @@ def prepare_package(destdir)
     end
 
     # check for conflicts with other installed files
-    conflicts = determine_conflicts(`pwd`.chomp, @pkg.name)
-    unless conflicts.empty?
+    conflicts = determine_conflicts(Dir.pwd, @pkg.name)
+    if conflicts.any?
       if CREW_CONFLICTS_ONLY_ADVISORY || @pkg.conflicts_ok?
         puts 'Warning: There is a conflict with the same file in another package.'.orange
       else
@@ -1697,7 +1697,7 @@ def remove(pkgName)
     end
   end
 
-  conflicts = determine_conflicts(`pwd`.chomp, @pkg.name)
+  conflicts = determine_conflicts(Dir.pwd, @pkg.name)
 
   # if the filelist exists, remove the files and directories installed by the package
   if File.file?("#{CREW_META_PATH}#{pkgName}.filelist")
@@ -1713,7 +1713,7 @@ def remove(pkgName)
         else
           puts "Removing file #{line}".lightred if @opt_verbose
           puts "filelist contains #{line}".lightred if @opt_verbose && !line.include?(CREW_PREFIX)
-          if line.include?(CREW_PREFIX)
+          if line.start_with?(CREW_PREFIX)
             if conflicts.include?(line)
               puts "#{line} is in another package. It will not be removed during the removal of #{pkgName}".orange
             else
--- a/bin/crew
+++ b/bin/crew
@@ -1072,7 +1072,6 @@ def determine_conflicts(dir, pkg)
   if File.file?("#{dir}/filelist")
     if File.file?("#{CREW_META_PATH}#{pkg}.filelist")
       puts 'Checking for conflicts with files from installed packages...'
-      puts "#{CREW_META_PATH}#{pkg}.filelist found"
       conflictscmd = `grep --exclude #{CREW_META_PATH}#{pkg}.filelist --exclude #{CREW_META_PATH}*_build.filelist -Fxf filelist #{CREW_META_PATH}*.filelist`
       conflicts = conflictscmd.gsub(/(\.filelist|#{CREW_META_PATH})/, '').split("\n")
       conflicts.reject!(&:empty?)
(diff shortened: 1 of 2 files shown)


satmandu:
@Zopolis4 Can you also test this?

bin/crew
```diff
@@ -1067,6 +1067,30 @@ def compress_doc(dir)
   end
 end
 
+def determine_conflicts(dir, pkg)
+  conflicts = []
+  if File.file?("#{dir}/filelist")
+    if File.file?("#{CREW_META_PATH}#{pkg}.filelist")
+      puts 'Checking for conflicts with files from installed packages...'
+      conflictscmd = `grep --exclude #{CREW_META_PATH}#{pkg}.filelist --exclude #{CREW_META_PATH}*_build.filelist -Fxf filelist #{CREW_META_PATH}*.filelist`
```
uberhacker, line 1075: Should it be `-Fxf #{dir}/filelist`?

bin/crew
```diff
@@ -1067,6 +1067,30 @@ def compress_doc(dir)
...
+      conflicts = conflictscmd.gsub(/(\.filelist|#{CREW_META_PATH})/, '').split("\n")
+      conflicts.reject!(&:empty?)
+    end
+  elsif File.file?("#{CREW_META_PATH}#{pkg}.filelist")
+    puts "Checking for conflicts of #{pkg} with files from installed packages...".orange
+    conflictscmd = `grep --exclude #{CREW_META_PATH}#{pkg}.filelist --exclude #{CREW_META_PATH}*_build.filelist -Fxf #{CREW_META_PATH}#{pkg}.filelist #{CREW_META_PATH}*.filelist`
+    conflicts = conflictscmd.gsub(/(\.filelist|#{CREW_META_PATH})/, '').split("\n")
+    conflicts.reject!(&:empty?)
+  end
+  puts "removing #{pkg}".orange
```
uberhacker, line 1085: Why is this line necessary?

bin/crew
```diff
@@ -1690,8 +1713,14 @@ def remove(pkgName)
...
-          puts "fileist contains #{line}".lightred if @opt_verbose && !line.include?(CREW_PREFIX)
-          FileUtils.rm_rf line if line.include?(CREW_PREFIX)
+          puts "filelist contains #{line}".lightred if @opt_verbose && !line.include?(CREW_PREFIX)
+          if line.include?(CREW_PREFIX)
+            if conflicts.include?(line)
+              puts "#{line} is in another package. It will not be removed during the removal of #{pkgName}".orange
+            else
+              FileUtils.rm_rf line
+            end
+          end
```
uberhacker, line 1723: Would `file` be more descriptive of the variable than `line`?

satmandu: Yes, but note that line is used for both files and _directories_ in the code block below this...

bin/crew
```diff
@@ -1067,6 +1067,29 @@ def compress_doc(dir)
   end
 end
 
+def determine_conflicts(dir, pkg)
+  conflicts = []
+  if File.file?("#{dir}/filelist")
+    if File.file?("#{CREW_META_PATH}#{pkg}.filelist")
+      puts 'Checking for conflicts with files from installed packages...'
+      conflictscmd = `grep --exclude #{CREW_META_PATH}#{pkg}.filelist --exclude #{CREW_META_PATH}*_build.filelist -Fxf #{dir}/filelist #{CREW_META_PATH}*.filelist`
+      conflicts = conflictscmd.gsub(/(\.filelist|#{CREW_META_PATH})/, '').split("\n")
```
supechicken, line 1076:
```ruby
conflicts = File.basename(conflictscmd, '.filelist').lines(chomp: true)
# conflicts.reject!(&:empty?) can be removed
```

bin/crew
```diff
@@ -1690,8 +1712,14 @@ def remove(pkgName)
           puts "Removing #{line} will break crew. It was #{'NOT'.lightred} deleted." if @opt_verbose
         else
           puts "Removing file #{line}".lightred if @opt_verbose
-          puts "fileist contains #{line}".lightred if @opt_verbose && !line.include?(CREW_PREFIX)
-          FileUtils.rm_rf line if line.include?(CREW_PREFIX)
+          puts "filelist contains #{line}".lightred if @opt_verbose && !line.include?(CREW_PREFIX)
+          if line.include?(CREW_PREFIX)
```
supechicken, line 1716: More specific in matching?
```ruby
if line.start_with?(CREW_PREFIX)
```

bin/crew
```diff
@@ -1067,6 +1067,29 @@ def compress_doc(dir)
...
+      conflicts = conflictscmd.gsub(/(\.filelist|#{CREW_META_PATH})/, '').split("\n")
+      conflicts.reject!(&:empty?)
+    end
+  elsif File.file?("#{CREW_META_PATH}#{pkg}.filelist")
+    puts "Checking for conflicts of #{pkg} with files from installed packages...".orange
+    conflictscmd = `grep --exclude #{CREW_META_PATH}#{pkg}.filelist --exclude #{CREW_META_PATH}*_build.filelist -Fxf #{CREW_META_PATH}#{pkg}.filelist #{CREW_META_PATH}*.filelist`
+    conflicts = conflictscmd.gsub(/(\.filelist|#{CREW_META_PATH})/, '').split("\n")
+    conflicts.reject!(&:empty?)
+  end
+  unless conflicts.empty?
```
supechicken, line 1085: We can use `.any?` since this is an array:
```ruby
if conflicts.any?
```

bin/crew
```diff
@@ -1677,6 +1696,9 @@ def remove(pkgName)
       FileUtils.ln(@canonicalized_file, "#{CREW_LIB_PREFIX}/#{file}", force: true)
     end
   end
+
+  conflicts = determine_conflicts(`pwd`.chomp, @pkg.name)
```
supechicken, line 1700: Rubyize...
```ruby
conflicts = determine_conflicts(Dir.pwd, @pkg.name)
```

bin/crew
satmandu: Is there an advantage of `.any?` vs `.empty?` ?

supechicken: Not much, it might have a better performance because of not having double negation

bin/crew
satmandu: I don't think that accounts for BOTH gsub replacements though...

And I think the point of `conflicts.reject!(&:empty?)` is to then remove any lines which end up blank... though I agree that I can't imagine a situation where this would lead to a blank line... but better to be safe?

satmandu: Still need to remove CREW_META_PATH from the fields, because that will always conflict...
`gsub(/(\.filelist|#{CREW_META_PATH})/, '')`

satmandu: We need the path in each line though. Every file + path has to be unique... because the `def remove` code is checking to see if a pathname + file in the conflicts array matches a file about to be deleted.

supechicken: Never mind, it seems that I completely misunderstand the usage of this line (I thought it is used to handle the file paths but not the message outputted by `grep`) 😅

PR #8379, opened 2023-06-13
crew: Add glob support to FileUtils

supechicken:
This PR adds glob support to FileUtils methods with multiple argument support, which means we can use globbing on `FileUtils` without the need of `Dir[]` (just like regular `FileUtils` call) and thus improve the consistency of code

Tested on `x86_64`

### Example
```ruby
# before
FileUtils.rm_rf Dir['*']
FileUtils.cp_r Dir.glob("#{HOME}/*"), '/'

# after
FileUtils.rm_rf '*'
FileUtils.cp_r '~/*', '/'
```

```
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=fileutils_extended CREW_TESTING=1 crew update
```


satmandu:
Shouldn't this be submitted upstream to the ruby fileutils repo?

Otherwise we're maintaining a patch against upstream, which is a maintenance hassle for us.

(It's not terribly hard to submit a PR upstream to fileutils... I did it this year!)

supechicken:
I don't think there would be a chance of merging this to ruby upstream as they need to ensure that all changes need to be backward-compatible (this change will break with files that have symbols like `~` or `*` in the filename, although this was very rare)

But anyway we can give a try on it :)

uberhacker:
Can we convert this to draft and fix the conflicting file?

uberhacker:
Any upstream updates on this?

satmandu:
Please submit this to upstream! Otherwise we will be maintaining a downstream patch to FileUtils indefinitely!

(Upstream DOES accept patches! I've done it!)

Zopolis4:
> I don't think there would be a chance of merging this to ruby upstream as they need to ensure that all changes need to be backward-compatible.

They will merge backwards-incompatible changes-- here is the [list of compatibility issues in Ruby 3.4](https://docs.ruby-lang.org/en/master/NEWS_md.html#label-Compatibility+issues).

image

We probably wont be able to get this change into Ruby 3.4, but the earlier we submit it upstream the sooner we'll be able to use it.

PR #8414, opened 2023-06-20
Add automatic builds

Zopolis4:
When finished, this PR will automatically build changed packages, update the filelists, upload new binaries and update the sha256s, and commit them to the PR.

Right now, it only does that on x86_64. Or, at least, it could. It needs a gitlab secret (I've discussed this further on the slack channel), and currently artifacts are broken. However, it does work. For a period of 2 builds uploading artifacts worked, before a change in the docker file broke them.

I've got an issue for the broken uploading artifacts over at actions/upload-artifact#409, but it may be a better solution to fix the docker file, as I doubt I'm gonna get a response on that issue any time soon.

In terms of fixing the docker file, I might need @satmandu's help here, given that its their docker file. 

Additionally, I need help getting `crewbuild:386` and `crewbuild:arm32v7` working. In particular, I need to be able to use them the way I'm using `satmandu/crewbuild:amd64`, using the github actions integration.

Once those are working, I'll adjust this PR and the workflow accordingly.

### DIFF
--- a/.github/workflows/Build.yml
+++ b/.github/workflows/Build.yml
@@ -1,0 +1,72 @@
+---
+name: Build
+on:
+  workflow_call:
+    inputs:
+      packages:
+        required: true
+        type: string
+jobs:
+  build_x86_64:
+    runs-on: ubuntu-latest
+    container:
+      # This is using an older version of the dockerfile to avoid issues with actions/upload-artifact
+      image: satmandu/crewbuild:amd64@sha256:8ce37cddc8950a6717c2b6a335580aa4142393926fff0640ec64b6b78ee530b8
+      options: -u chronos
+    steps:
+      - name: Setup environment
+        run: |
+          mkdir /usr/local/manifests
+          mkdir /usr/local/build
+      - name: Build package
+        run: |
+          export CREW_LOCAL_MANIFEST_PATH=/usr/local/manifests
+          cd /usr/local/build
+          for package in ${{ inputs.packages }}; do
+            yes | crew build $package
+          done
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: artifacts_x86_64
+          path: |
+            /usr/local/build
+            /usr/local/manifests
+          retention-days: 1
+  commit:
+    needs: build_x86_64
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repo
+        uses: actions/checkout@v3
+      - name: Download artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: artifacts_x86_64
+          path: ${{ github.workspace }}
+      - name: Update filelists
+        run: |
+          for package in ${{ inputs.packages }}; do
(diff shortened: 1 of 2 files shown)


Zopolis4:
Looked into the upload-artifact issue a little more, it works on `satmandu/crewbuild:amd64@sha256:8ce37cddc8950a6717c2b6a335580aa4142393926fff0640ec64b6b78ee530b8`, but fails on `satmandu/crewbuild:amd64@sha256:11987d7f9da04ca647097fec0bd0cb7386741eb6344c98077a668afe663a7276`.

satmandu, do you know what changed between these two versions?

.github/workflows/Build.yml
```diff
@@ -0,0 +1,72 @@
...
+      image: satmandu/crewbuild:amd64@sha256:8ce37cddc8950a6717c2b6a335580aa4142393926fff0640ec64b6b78ee530b8
+      options: -u chronos
+    steps:
+      - name: Setup environment
+        run: |
+          mkdir /usr/local/manifests
+          mkdir /usr/local/build
+      - name: Build package
+        run: |
+          export CREW_LOCAL_MANIFEST_PATH=/usr/local/manifests
```
uberhacker, line 23: Not FHS3 compliant?

Zopolis4: FHS3?

Zopolis4:
Been thinking some more about this PR, there might be a major refactoring, might not, etc. Anyways, what you see here is very WIP, so don't review it or anything until I mark this pr as ready for review. For now, consider this PR here just to raise awareness.

satmandu:
> Looked into the upload-artifact issue a little more, it works on `satmandu/crewbuild:amd64@sha256:8ce37cddc8950a6717c2b6a335580aa4142393926fff0640ec64b6b78ee530b8`, but fails on `satmandu/crewbuild:amd64@sha256:11987d7f9da04ca647097fec0bd0cb7386741eb6344c98077a668afe663a7276`.
> 
> satmandu, do you know what changed between these two versions?

Not sure... but there have been a bunch of crew changes in the past week...

.github/workflows/Build.yml
satmandu: https://refspecs.linuxfoundation.org/FHS_3.0/index.html

Putting stuff under /usr/local that shouldn't be there...

Zopolis4: It's only there for the life of the workflow, and its an easily accessible and writable directory on ChromeOS.

Zopolis4:
Closing because the docker situation is not very good for using github actions built-in docker support when we're trynig to run `armv7l` and `aarch64`, and the number of upstream changes have made it easier for me to build so there's less of a need for this, and my custom build script is also incompatible with upstream due to a number of changes, and would need not insignificant reworking. (script still available [here](https://gist.github.com/Zopolis4/02a5e900068680c55cd521ee663a28d2), the build workflow on this branch is nothing special, and the build workflow [here](https://github.com/Zopolis4/chromebrew/blob/builds1_test/.github/workflows/Build.yml) is also nothing particularly notable, with half of it being available in equivalent form in the Repology.yml workflow, and the rest probably still going to be available with the way github handles deleted branches and the like.)

PR #8415, opened 2023-06-20
crew: Add manifest support to `crew files`, remove `crew updatefiles` command

supechicken:
### Changes
- Use `git` to fetch manifest during `crew update` for consistency, remove `crew updatefiles` as it is no longer needed.
- Remove variables related to `crew updatefiles`
- Add manifest support to `crew files`

```
# run it twice
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=manifest_use_git CREW_TESTING=1 crew update
```


satmandu:
`crew updatefiles` is needed to inform crew about local updates to package files. We saw this during install before git is available, if I recall correctly.

It is also used to test locally when the network is unavailable when making local changes. 😅

Zopolis4:
It's triggering on Contributing.md and MANUAL.md, and because those have never been linted before it's failing. I would have thought they'd at least be linted manually before the original CI was added, but that's fine.

Anyways, since I'm the one advocating for this CI system i'll deal with the underlying problems my changes have exposed-- gimme a day or two to make a PR

install.sh
```diff
@@ -355,7 +355,7 @@ else
   git checkout "${BRANCH}"
 
   # Set sparse-checkout folders.
-  git sparse-checkout set packages lib bin crew tools
+  git sparse-checkout set packages "manifest/${ARCH}" lib bin crew tools
```
Zopolis4, line 358: This won't work on `aarch64` machines, btw. Better just to use `manifest` and then we can move back to `"manifest/${ARCH}"` when we support `aarch64` fully.

uberhacker: It makes sense to use `ARCH` here.  We don't need to be concerned about manifest files that are not matching the architecture.  I believe @satmandu has already added logic that determines if lib64 is needed for aarch64 or not.

Zopolis4: The problem is armv7l installs on aarch64 userspaces-- those will need manifests in manifest/armv7l but will only try and checkout a manifest/aarch64 directory.

uberhacker: Currently, we're mapping aarch64 to armv7l so that would never happen.

Zopolis4: What do you mean?

uberhacker: I suppose we don't have a symlink from `metatag/armv7l` to `metatag/aarch64` yet so I guess you are right.
```
ARCH = KERN_ARCH.eql?('armv8l') ? 'armv7l' : KERN_ARCH
```
This logic in const.rb does change for armv8l but not for aarch64.

Zopolis4: And at any rate that logic is in crew-- the ARCH variable being drawn from here is the output of uname -m, in the installer.

bin/crew
```diff
@@ -616,11 +599,7 @@ def prop(silent = false)
 end
 
 def whatprovides(regexPat)
-  @crew_manifest_search_dir = ENV['CREW_TESTING'] == '1' ? "#{CREW_MANIFEST_CACHE_DIR}/chromebrew-#{CREW_TESTING_BRANCH}/manifest/#{ARCH}" : "#{CREW_MANIFEST_CACHE_DIR}/chromebrew-#{CREW_GITHUB_BRANCH}/manifest/#{ARCH}"
-
-  updatefiles unless Dir.exist?(@crew_manifest_search_dir) && !Dir.empty?(@crew_manifest_search_dir)
-
-  matchedList = `grep -r "#{regexPat}" #{@crew_manifest_search_dir}`.lines(chomp: true).flat_map do |result|
+  matchedList = `grep -R "#{regexPat}" #{CREW_LIB_PATH}/manifest/#{ARCH}`.lines(chomp: true).flat_map do |result|
```
uberhacker, line 602: Add?
```
userspace_arch = ARCH.include('aarch64') ? 'armv7l' : ARCH
matchedList = `grep -R "#{regexPat}" #{CREW_LIB_PATH}/manifest/#{userspace_arch}`.lines(chomp: true).flat_map do |result|
```
Might make sense to add a new value in constant.rb too.

uberhacker: See https://github.com/chromebrew/chromebrew/pull/8425.

install.sh
uberhacker: See #8427.

bin/crew
supechicken: Change made

supechicken:
> `crew updatefiles` is needed to inform crew about local updates to package files. We saw this during install before git is available, if I recall correctly.
>
> It is also used to test locally when the network is unavailable when making local changes. 😅

But `crew updatefiles` also requires network connection :) This PR just moves the manifest directory to a different location.

> I'm not opposed to integrating the `crew updatefiles` command into `crew update` but we need a way to let the user know it should be refreshed occasionally. Like @satmandu mentioned, it doesn't update locally unless it pulls down the updated filelists.

Users can do a `crew update` if they want to ensure the manifest is up to date. (especially downloading package scripts along with `manifest` does not take any extra time compared to downloading package scripts only)

satmandu:
@supechicken Sorry I had `crew updatefiles` confused with `crew update compatible`. I withdraw my objection.

uberhacker:
> @supechicken Sorry, I had `crew updatefiles` confused with `crew update compatible`. I withdraw my objection.

But even that can be merged into `crew update`. :slightly_smiling_face:

bin/crew
```diff
@@ -557,46 +551,35 @@ def const(var = nil)
...
-      size += File.size(filename) if File.file?(filename) && !File.symlink?(filename)
-    end
-
-    puts filelist, <<~EOT.lightgreen
-
-      Total found: #{lines}
-      Disk usage: #{human_size(size)}
-    EOT
+  local_filelist    = File.join(CREW_META_PATH, "#{pkgName}.filelist")
+  manifest_filelist = File.join(CREW_LIB_PATH, "manifest/#{ARCH}/#{pkgName[0]}/#{pkgName}.filelist")
```
uberhacker, line 555: `s,ARCH,USER_SPACE_ARCH,`?

uberhacker:
Should we also clean up?
```
FileUtils.rm_rf "#{HOME}/.cache/crewcache/manifest" if Dir.exists?("#{HOME}/.cache/crewcache/manifest")
```

lib/const.rb
```diff
@@ -133,23 +133,6 @@
...
-  puts e.message.to_s.orange
-rescue Errno::EACCES => e
-  # no write access
-  puts @crew_manifest_cache_error.lightred
-  puts e.message.to_s.orange
-rescue Errno::ENOENT => e
-  # weird fs e.g., /proc
-  puts @crew_manifest_cache_error.lightred
-  puts e.message.to_s.orange
-end
```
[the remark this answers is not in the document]
uberhacker: We can remove the `require 'fileutils'` in const.rb along with this now.

satmandu:
> Should we also clean up?
> 
> ```ruby
> FileUtils.rm_rf "#{HOME}/.cache/crewcache/manifest" if Dir.exists?("#{HOME}/.cache/crewcache/manifest")
> ```

As in, change it to:
```ruby
FileUtils.rm_rf "#{HOME}/.cache/crewcache/manifest"
```
?

uberhacker:
> > `crew updatefiles` is needed to inform crew about local updates to package files. We saw this during install before git is available, if I recall correctly.
> > It is also used to test locally when the network is unavailable when making local changes. sweat_smile
> 
> But `crew updatefiles` also requires network connection :) This PR just moves the manifest directory to a different location.

Actually, this is using the cloned repo directory that already exists locally.  I just wasn't smart enough to figure this out when I first set up the manifest search. :blush:

uberhacker:
> > Should we also clean up?
> > ```ruby
> > FileUtils.rm_rf "#{HOME}/.cache/crewcache/manifest" if Dir.exists?("#{HOME}/.cache/crewcache/manifest")
> > ```
> 
> As in, change it to:
> 
> ```ruby
> FileUtils.rm_rf "#{HOME}/.cache/crewcache/manifest"
> ```
> 
> ?

Yeah, I suppose it will run silently even if the directory doesn't exist.

bin/crew
```diff
@@ -640,10 +619,13 @@ def update
   abort "'crew update' is used to update crew itself. Use 'crew upgrade <package1> [<package2> ...]' to update specific packages.".orange if @pkgName
   # update package lists
   Dir.chdir(CREW_LIB_PATH) do
+    # Set sparse-checkout folders.
+    system "git sparse-checkout set packages manifest/#{ARCH} lib bin crew tools"
```
uberhacker, line 623: `s,ARCH,USER_SPACE_ARCH,`?

Zopolis4:
Do we have an upgrade path from current installations to this? (Other than manually setting new sparse-checkout folders or re-installing crew).

bin/crew
supechicken: Done

lib/const.rb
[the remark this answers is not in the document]
supechicken: Done

supechicken:
> Do we have an upgrade path from current installations to this? (Other than manually setting new sparse-checkout folders or re-installing crew).

No, we don't have a `postinstall` section for `crew update` currently :/

supechicken:
> How about git checkout manifest? This will create the folder if it doesn't exist locally yet. Even if the folder exists, this will pull down the latest changes so it doesn't hurt to include it without any conditions.

Umm... That's what `git sparse-checkout` does. (also does `git checkout <path>` works😅)

uberhacker:
@supechicken: Bump crew version and resolve merge conflict and this is ready?

satmandu (approved):

PR #8425, opened 2023-06-21
Add USER_SPACE_ARCH constant

uberhacker:
Could be helpful for logic that needs to switch user space for aarch64.


satmandu (approved):
That's a nice way to deal with it...

PR #8429, opened 2023-06-22
Remove basic miscellaneous uses of @_ver

Zopolis4:
Following on from #8428, I am removing 27 uses of `@_ver` from packages that use it like this:
```
@_ver = '1.2.3'
version @_ver
...
@git_hash = "gopls/v#{@_ver}"
```
which now becomes:
```
version '1.2.3'
...
@git_hash = "gopls/v#{version}"
```

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=laver6 CREW_TESTING=1 crew update
```

### DIFF
--- a/packages/apng2gif.rb
+++ b/packages/apng2gif.rb
@@ -3,11 +3,10 @@ require 'package'
 class Apng2gif < Package
   description 'Convert APNG animations into animated GIF format.'
   homepage 'https://sourceforge.net/projects/apng2gif/'
-  @_ver = '1.8'
-  version @_ver
+  version '1.8'
   license 'ZLIB'
   compatibility 'all'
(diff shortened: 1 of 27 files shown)


Zopolis4:
In other news, this (when combined with 8428) leaves only 24 basic uses of `@_ver`!
(basic usage):
```
version @_ver
```

That leaves around 285 uses of `@_ver`, of which around 160 are using `version "#{@_ver}-py3.11"`, which will be solved by 7713. The remaining uses are almost all ones which require 8196 to be merged, and will be relativley simple replacements after that.

satmandu (approved):

PR #8431, opened 2023-06-22
Add fileutils dependency to lib/const.rb

Zopolis4:
`lib/const.rb` needs `fileutils`, and this becomes apparent when trying to run any of the tests. I presume it worked fine in regular usage because whatever file that included it already had `fileutils` as a dependency.

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=uangle CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,3 +1,5 @@
+require 'fileutils'
+
 # Defines common constants used in different parts of crew
 
 CREW_VERSION = '1.34.5'


supechicken:
Theoretically `const.rb` should not contain any part that doesn't relate to defining constants, including `FileUtils.mkdir_p` if the directory does not exist.

Zopolis4:
That's fair, but right now this PR fixes a problem that exists. We can look into refactoring the code later, of course.

Zopolis4:
All the tests are failing. `load_test`, `cycle_test` and `spellcheck_test`.

As for the usage of FileUtils in `lib/const.rb`...
```ruby
begin
  FileUtils.mkdir_p CREW_MANIFEST_CACHE_DIR
rescue Errno::EROFS => e
```

satmandu:
> ```ruby
> begin
>   FileUtils.mkdir_p CREW_MANIFEST_CACHE_DIR
> rescue Errno::EROFS => e
> ```
This is a really strong argument for needing `require 'fileutils'` in the file!

satmandu:
Also, I didn't know you could do
```
```ruby
```
to identify ruby code in Markdown!

uberhacker:
> All the tests are failing. `load_test`, `cycle_test` and `spellcheck_test`.
> 
> As for the usage of FileUtils in `lib/const.rb`...
> 
> ```ruby
> begin
>   FileUtils.mkdir_p CREW_MANIFEST_CACHE_DIR
> rescue Errno::EROFS => e
> ```

My bad, missed that on initial inspection.

uberhacker (approved):

satmandu:
> Theoretically `const.rb` should not contain any part that doesn't relate to defining constants, including `FileUtils.mkdir_p` if the directory does not exist. (that should be done in `crew` instead)

Would it be cleaner to put the `FileUtils.mkdir_p` inside the const section of `crew?

```ruby
def const(var = nil)
  if var
    value = Object.const_get(var.to_sym)
    puts "#{var}=#{value}"
  else
    # Get all constants defined by const.rb
    constants = Object.constants.select do |const|
      Object.const_source_location(const)[0] == File.join(CREW_LIB_PATH, 'lib/const.rb')
    end.sort

    # Print a sorted list of the remaining constants used by crew.
    constants.each do |var|
      value = Object.const_get(var.to_sym)
      puts "#{var}=#{value}"
    end
  end
end
```
That would be more complicated. This code block only handles the constants set by the `const.rb` file.

PR #8446, opened 2023-06-27
Add CREW_ARCHIVE_DEST constant for use during builds

satmandu:
`crew` currently builds package files in the dir that it was started from when `crew build` is invoked. This breaks packages that are referencing files in the `/usr/local/lib/crew/packages` dir. But invoking `crew build` in `/usr/local/crew/packages` also drops the built package files in `/usr/local/crew/packages`... which can be unwanted.

- This PR lets one set an ENV variable of `CREW_ARCHIVE_DEST` to specify where one wants those generated packages put. It falls back to `Dir.pwd`, the previous location.
- Add a check to bail out if `Dir.pwd` or `CREW_ARCHIVE_DEST` is not writable has also been added.
- Make `buildsystems` `{meson,make,configure}_options` messages more verbose.
- Add `filefix` to `buildsystems`' `autotools`.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=crew_build_dir CREW_TESTING=1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1602,7 +1602,8 @@ def resolve_dependencies_and_build
       post_install
     end
     search origin, true
-    build_package Dir.pwd
+    abort "CREW_ARCHIVE_DEST (#{CREW_ARCHIVE_DEST}) is not writable!".lightred unless File.writable?(CREW_ARCHIVE_DEST)
+    build_package CREW_ARCHIVE_DEST
   rescue InstallError => e
     abort "#{@pkg.name} failed to build: #{e}".lightred
   ensure
@@ -1616,7 +1617,7 @@ def resolve_dependencies_and_build
   @resolve_dependencies_and_build = 0
 end
 
-def build_package(pwd)
+def build_package(crew_archive_dest)
   abort 'It is not possible to build a fake package'.lightred if @pkg.is_fake?
   abort 'It is not possible to build without source'.lightred unless @pkg.is_source?(@device[:architecture])
   abort "Unable to locate CREW_LOCAL_MANIFEST_PATH. Please try again in the chromebrew/release/#{ARCH} directory.".lightred if CREW_LOCAL_MANIFEST_PATH.empty?
@@ -1643,10 +1644,10 @@ def build_package(pwd)
 
   # build package from filelist, dlist and binary files in CREW_DEST_DIR
   puts 'Archiving...'
-  archive_package pwd
+  archive_package crew_archive_dest
 end
 
-def archive_package(pwd)
+def archive_package(crew_archive_dest)
   # Check to see that there is a working zstd
   if File.file?("#{CREW_PREFIX}/bin/zstd")
     @crew_prefix_zstd_available = File.file?("#{CREW_PREFIX}/bin/zstd") ? true : nil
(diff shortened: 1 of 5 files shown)


bin/crew
```diff
@@ -1616,7 +1617,7 @@ def resolve_dependencies_and_build
   @resolve_dependencies_and_build = 0
 end
 
-def build_package(pwd)
+def build_package(crew_archive_dest)
```
uberhacker, line 1620: Why not capitalized to reference the constant?

satmandu: good point. changed.

satmandu: Turns out that doesn't work as it fails the linter check:
```ruby
bin/crew:1620:19: F: Lint/Syntax: formal argument cannot be a constant
(Using Ruby 3.1 parser; configure using TargetRubyVersion parameter, under AllCops)
def build_package(CREW_ARCHIVE_DEST)
```

lib/buildsystems/autotools.rb
```diff
@@ -21,6 +21,10 @@ def self.build
         system 'autoreconf -fiv'
       end
     end
+    if `grep -q /usr/bin/file configure`
```
uberhacker, line 24:
```
if `grep -q /usr/bin/file`
```
Would this work too?  Not sure why configure would be needed.

lib/const.rb
```diff
@@ -62,6 +62,8 @@
   HOME = CREW_PREFIX + Dir.home
 end
 
+CREW_ARCHIVE_DEST = ENV.fetch('CREW_ARCHIVE_DEST', Dir.pwd)
```
uberhacker, line 65: Does `Dir.pwd` called from here return the correct current working directory?

satmandu: Yes:
```bash
chronos@mbp113-x86_64 /usr/local/lib/crew/packages $ crew const CREW_ARCHIVE_DEST
CREW_ARCHIVE_DEST=/usr/local/lib/crew/packages
chronos@mbp113-x86_64 /usr/local/lib/crew/packages $ cd ..
chronos@mbp113-x86_64 /usr/local/lib/crew $ crew const CREW_ARCHIVE_DEST
CREW_ARCHIVE_DEST=/usr/local/lib/crew
```

satmandu: and setting it works correctly too:
```
chronos@mbp113-x86_64 /usr/local/lib/crew $ CREW_ARCHIVE_DEST=/usr/local/tmp crew const CREW_ARCHIVE_DEST
CREW_ARCHIVE_DEST=/usr/local/tmp
```

uberhacker: Nice!

lib/buildsystems/autotools.rb
```diff
@@ -21,6 +22,12 @@ def self.build
         system 'autoreconf -fiv'
       end
     end
+    abort "configure script not found!".lightred if File.file?('configure')
```
uberhacker, line 25: Will this fail the linter because of double quotes instead of single quotes?

uberhacker: Apparently, not. :)

uberhacker (approved):

lib/buildsystems/autotools.rb
satmandu: I'll fix it in my next PR...

PR #8465, opened 2023-07-10
Meson buildsystem fix

satmandu:
Fixes this message during meson builds:
```
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
```

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=meson_buildsystem  CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,5 +1,5 @@
 # Defines common constants used in different parts of crew
-CREW_VERSION = '1.34.9'
+CREW_VERSION = '1.35.0'
 
 # kernel architecture
 KERN_ARCH = `uname -m`.chomp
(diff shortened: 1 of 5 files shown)


lib/const.rb
```diff
@@ -1,5 +1,5 @@
 # Defines common constants used in different parts of crew
-CREW_VERSION = '1.34.9'
+CREW_VERSION = '1.35.0'
```
uberhacker, line 2: This is the correct move even if we didn't make any direct changes to the crew executable.  But we should probably clearly define reasons for bumping the version.  Should we do it for document changes and/or only for files that reside inside /usr/local/lib/crew/lib and the crew ruby script?

uberhacker (approved):

lib/const.rb
satmandu: I think anything inside lib being changed and also any changes to the `crew` script should trigger a version change.

PR #8486, opened 2023-07-19
Fix no_lto detection in meson and cmake buildsystems

Zopolis4:
Discovered this while working on some other stuff.

As an aside, setting `-DCMAKE_SHARED_LINKER_FLAGS` to `-fno_lto` breaks builds as `ar` interprets it as an argument. Not sure if this actually occurs after this fix though, or whether it was just in my specific environment, and so on. Just noting it here.

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=flixesto CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/buildsystems/cmake.rb
+++ b/lib/buildsystems/cmake.rb
@@ -11,9 +11,9 @@ class CMake < Package
 
   def self.build
     puts "Additional cmake_options being used: #{@cmake_options.nil? || @cmake_options.empty? ? '<no cmake_options>' : @cmake_options}".orange
-    @crew_cmake_options = no_lto ? CREW_CMAKE_FNO_LTO_OPTIONS : CREW_CMAKE_OPTIONS
-    @mold_linker_prefix_cmd = CREW_LINKER == 'mold' ? 'mold -run ' : ''
-    system "#{@mold_linker_prefix_cmd}cmake -B builddir -G Ninja #{@crew_cmake_options} #{@cmake_options}"
+    @crew_cmake_options = @no_lto ? CREW_CMAKE_FNO_LTO_OPTIONS : CREW_CMAKE_OPTIONS
+    @mold_linker_prefix_cmd = CREW_LINKER == 'mold' ? 'mold -run' : ''
+    system "#{@mold_linker_prefix_cmd} cmake -B builddir -G Ninja #{@crew_cmake_options} #{@cmake_options}"
     system "#{CREW_NINJA} -C builddir"
   end
 
--- a/lib/buildsystems/meson.rb
+++ b/lib/buildsystems/meson.rb
@@ -11,9 +11,9 @@ class Meson < Package
 
   def self.build
     puts "Additional meson_options being used: #{@meson_options.nil? || @meson_options.empty? ? '<no meson_options>' : @meson_options}".orange
-    @crew_meson_options = no_lto ? CREW_MESON_FNO_LTO_OPTIONS : CREW_MESON_OPTIONS
-    @mold_linker_prefix_cmd = CREW_LINKER == 'mold' ? 'mold -run ' : ''
-    system "#{@mold_linker_prefix_cmd}meson setup #{@crew_meson_options} #{@meson_options} builddir"
+    @crew_meson_options = @no_lto ? CREW_MESON_FNO_LTO_OPTIONS : CREW_MESON_OPTIONS
+    @mold_linker_prefix_cmd = CREW_LINKER == 'mold' ? 'mold -run' : ''
+    system "#{@mold_linker_prefix_cmd} meson setup #{@crew_meson_options} #{@meson_options} builddir"
     system 'meson configure builddir'
     system "#{CREW_NINJA} -C builddir"
   end
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,5 +1,5 @@
 # Defines common constants used in different parts of crew
-CREW_VERSION = '1.35.1'
+CREW_VERSION = '1.35.2'
 
 # kernel architecture
 KERN_ARCH = `uname -m`.chomp


lib/buildsystems/cmake.rb
```diff
@@ -11,9 +11,9 @@ def self.check?(bool = true)
 
   def self.build
     puts "Additional cmake_options being used: #{@cmake_options.nil? || @cmake_options.empty? ? '<no cmake_options>' : @cmake_options}".orange
-    @crew_cmake_options = no_lto ? CREW_CMAKE_FNO_LTO_OPTIONS : CREW_CMAKE_OPTIONS
-    @mold_linker_prefix_cmd = CREW_LINKER == 'mold' ? 'mold -run ' : ''
-    system "#{@mold_linker_prefix_cmd}cmake -B builddir -G Ninja #{@crew_cmake_options} #{@cmake_options}"
+    @crew_cmake_options = @no_lto ? CREW_CMAKE_FNO_LTO_OPTIONS : CREW_CMAKE_OPTIONS
```
satmandu, line 14: How does this handle detection of `no_lto` which would be set in a package?

satmandu:
@supechicken Can you help with this? `no_lto` is defaulting to `true` in `buildsystems/cmake` for some reason...

satmandu:
This doesn't work:
```ruby
require 'package'

class CMake < Package
  def self.cmake_options(options = '')
    return (@cmake_options = options if options)
  end

  def self.check?(bool = true)
    return (@check = bool)
  end

  def self.build
    puts "Additional cmake_options being used: #{@cmake_options.nil? || @cmake_options.empty? ? '<no cmake_options>' : @cmake_options}".orange
    if @pkg.no_lto?
      @no_lto = true
    else
      @no_lto = false
    end
    puts "no_lto is #{no_lto}, @no_lto is #{@no_lto}"
    @crew_cmake_options = @no_lto ? CREW_CMAKE_FNO_LTO_OPTIONS : CREW_CMAKE_OPTIONS
    @mold_linker_prefix_cmd = CREW_LINKER == 'mold' ? 'mold -run ' : ''
    system "#{@mold_linker_prefix_cmd}cmake -B builddir -G Ninja #{@crew_cmake_options} #{@cmake_options}"
    system "#{CREW_NINJA} -C builddir"
  end

  def self.install
    system "DESTDIR=#{CREW_DEST_DIR} #{CREW_NINJA} -C builddir install"
  end

  if @check
    def self.check
      system "#{CREW_NINJA} -C builddir test"
    end
  end
end
```

satmandu:
For the record, this PR does fix the symptom but not the underlying cause, which is causing `no_lto` to be detected as `true` when it is actually `false`...

supechicken:
> @supechicken Can you help with this? `no_lto` is defaulting to `true` in `buildsystems/cmake` for some reason...

The reason that the current setup does not work is because `no_lto` is used:
```ruby
@crew_cmake_options = no_lto ? CREW_CMAKE_FNO_LTO_OPTIONS : CREW_CMAKE_OPTIONS
```

(`no_lto` is used to **set** the flag to `true`, we should use `no_lto?` (notice the `?`) or `@no_lto` to **read** the flag)

That also explains why this PR fixes it :)

satmandu:
@supechicken where is `@no_lto` being set though?

satmandu:
I tried using this but it doesn't appear to set it correctly.

```ruby
if @pkg.no_lto?
   @no_lto = true
else
   @no_lto = false
end
```

supechicken:
> @supechicken where is `@no_lto` being set though?

In `package.rb` (under the `no_lto` method):

https://github.com/chromebrew/chromebrew/blob/e970f2a51dc721b83893cd96eb5446adf31a31eb/lib/package.rb#L147-L159


> I tried using this but it doesn't appear to set it correctly.
> 
> ```ruby
> if @pkg.no_lto?
>    @no_lto = true
> else
>    @no_lto = false
> end
> ```

It is because you called `no_lto` in the next line:
```ruby
puts "no_lto is #{no_lto}, @no_lto is #{@no_lto}"
```
This will set `no_lto` to `true` (you should use `no_lto?` here)

Zopolis4:
So... is this PR the correct solution? A bit of that went over my head, I'll admit, but I'm happy to change this PR if I should do it another way.

supechicken:
> So... is this PR the correct solution? A bit of that went over my head, I'll admit, but I'm happy to change this PR if I should do it another way.

Yes, it is the correct way to deal with it (at least for me 😀)

lib/buildsystems/cmake.rb
satmandu, line 14: This needs to be 
```ruby
@crew_cmake_options = no_lto? ? CREW_CMAKE_FNO_LTO_OPTIONS : CREW_CMAKE_OPTIONS
```

lib/buildsystems/meson.rb
```diff
@@ -11,9 +11,9 @@ def self.check?(bool = true)
 
   def self.build
     puts "Additional meson_options being used: #{@meson_options.nil? || @meson_options.empty? ? '<no meson_options>' : @meson_options}".orange
-    @crew_meson_options = no_lto ? CREW_MESON_FNO_LTO_OPTIONS : CREW_MESON_OPTIONS
-    @mold_linker_prefix_cmd = CREW_LINKER == 'mold' ? 'mold -run ' : ''
-    system "#{@mold_linker_prefix_cmd}meson setup #{@crew_meson_options} #{@meson_options} builddir"
+    @crew_meson_options = @no_lto ? CREW_MESON_FNO_LTO_OPTIONS : CREW_MESON_OPTIONS
```
satmandu, line 14: This needs to be 
```ruby
@crew_meson_options = no_lto? ? CREW_MESON_FNO_LTO_OPTIONS : CREW_MESON_OPTIONS
```

satmandu (changes requested):
Using `no_lto?` instead of `no_lto` should fix things...
We don't set `@no_lto` with the value of `no_lto` so I don't see how _that_ would work.

lib/buildsystems/cmake.rb
satmandu: Looks like I am wrong here!

lib/buildsystems/meson.rb
satmandu: I was wrong!

satmandu (approved):

satmandu:
Just putting `instance_variables` into a package file shows that `@no_lto` exists when `no_lto` is added to the package.

PR #8495, opened 2023-07-21
added package repo

Franceschinie:
## Additional information

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

### DIFF
--- a/packages/repo.rb
+++ b/packages/repo.rb
@@ -22,8 +22,6 @@ class Repo < Package
 
   def self.install
     FileUtils.install 'repo', "#{CREW_DEST_PREFIX}/bin/repo", mode: 0o755
-    system("chmod a+x $CREW_PREFIX/bin/repo")
- 
   end
  
  
--- a/packages/repo.rb
+++ b/packages/repo.rb
@@ -1,0 +1,33 @@
+
+require 'package'
+ 
+class Repo < Package
+  description 'Android Repo'
+  homepage 'https://gerrit.googlesource.com/git-repo/'
+  version '1.0'
+  license 'Apache License 2'      
+  compatibility 'x86_64' 
+  source_url 'https://storage.googleapis.com/git-repo-downloads/repo'
+  source_sha256 '974992a8bc1c787979f3eb7702a803b051deddc3cd75726c4de52e09e93b798d '
+
+  depends_on 'gpgme'
+  depends_on 'curl'
+  depends_on 'python3'
+  depends_on 'chromebeer'
+ 
+  def self.preflight
+    free_space = `echo $(($(stat -f --format="%a*%S" #{CREW_PREFIX})))`.chomp.to_i
+    abort 'Not enough free disk space.  You need at least 1 MB to install.'.lightred if free_space < 1000000
+ 
+  end
+
+  def self.install
+    FileUtils.mv 'repo', "/usr/local/bin/"
+    system("gpg --recv-key 8BB9AD793E8E6153AF0F9A4416530D5E920F5C65")
+    system("chmod a+x usr/local/bin/repo")
+ 
+  end
+ 
+ 
+
+end
--- a/packages/repo.rb
+++ b/packages/repo.rb
@@ -3,16 +3,13 @@ require 'package'
  
 class Repo < Package
   description 'Android Repo'
-  homepage 'https://gerrit.googlesource.com/git-repo/'
-  version '1.0'
+  version '2.32'
   license 'Apache License 2'      
-  compatibility 'x86_64' 
-  source_url 'https://storage.googleapis.com/git-repo-downloads/repo'
-  source_sha256 '974992a8bc1c787979f3eb7702a803b051deddc3cd75726c4de52e09e93b798d'
-
-  depends_on 'curl'
+  compatibility 'all' 
+  source_url 'https://github.com/GerritCodeReview/git-repo.git'
+  git_hashtag "v#{version}"
+ 
   depends_on 'python3'
-  depends_on 'chromebeer'
  
   def self.preflight
     free_space = `echo $(($(stat -f --format="%a*%S" #{CREW_PREFIX})))`.chomp.to_i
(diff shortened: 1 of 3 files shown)


packages/repo.rb
```diff
@@ -0,0 +1,33 @@
...
+  def self.preflight
+    free_space = `echo $(($(stat -f --format="%a*%S" #{CREW_PREFIX})))`.chomp.to_i
+    abort 'Not enough free disk space.  You need at least 1 MB to install.'.lightred if free_space < 1000000
+ 
+  end
+
+  def self.install
+    FileUtils.mv 'repo', "/usr/local/bin/"
+    system("gpg --recv-key 8BB9AD793E8E6153AF0F9A4416530D5E920F5C65")
+    system("chmod a+x usr/local/bin/repo")
```
uberhacker, line 27: Not sure why we would need the gpg key but would this work to reduce 3 lines to 1?
```
FileUtils.install 'repo', "#{CREW_DEST_PREFIX}/bin/repo", mode: 0o755
```

packages/repo.rb
```diff
@@ -0,0 +1,33 @@
...
+class Repo < Package
+  description 'Android Repo'
+  homepage 'https://gerrit.googlesource.com/git-repo/'
+  version '1.0'
+  license 'Apache License 2'      
+  compatibility 'x86_64' 
+  source_url 'https://storage.googleapis.com/git-repo-downloads/repo'
+  source_sha256 '974992a8bc1c787979f3eb7702a803b051deddc3cd75726c4de52e09e93b798d '
+
+  depends_on 'gpgme'
```
uberhacker, line 13: How is this used?

uberhacker (changes requested):
Thank you for the new package @Franceschinie!

packages/repo.rb
Franceschinie: gpgme is for gpg key

uberhacker: Right, I guess my question is how is gpg used for the install?

Franceschinie: for verity repo
export REPO=$(mktemp /tmp/repo.XXXXXXXXX)
curl -o ${REPO} https://storage.googleapis.com/git-repo-downloads/repo
gpg --recv-key 8BB9AD793E8E6153AF0F9A4416530D5E920F5C65
curl -s https://storage.googleapis.com/git-repo-downloads/repo.asc | gpg --verify - ${REPO} && install -m 755 ${REPO} ~/bin/repo
from
https://source.android.com/docs/setup/develop?hl=it#installing-repo

Franceschinie (commented):
added prefix support

packages/repo.rb
uberhacker: Doesn't `source_sha256` do that too?

packages/repo.rb
```diff
@@ -0,0 +1,33 @@
...
+require 'package'
+ 
+class Repo < Package
+  description 'Android Repo'
+  homepage 'https://gerrit.googlesource.com/git-repo/'
+  version '1.0'
+  license 'Apache License 2'      
+  compatibility 'x86_64' 
+  source_url 'https://storage.googleapis.com/git-repo-downloads/repo'
+  source_sha256 '974992a8bc1c787979f3eb7702a803b051deddc3cd75726c4de52e09e93b798d '
```
uberhacker, line 11: Extra space at the end?

packages/repo.rb
```diff
@@ -0,0 +1,31 @@
...
+ 
+  def self.preflight
+    free_space = `echo $(($(stat -f --format="%a*%S" #{CREW_PREFIX})))`.chomp.to_i
+    abort 'Not enough free disk space.  You need at least 1 MB to install.'.lightred if free_space < 1000000
+ 
+  end
+
+  def self.install
+    FileUtils.install 'repo', "#{CREW_DEST_PREFIX}/bin/repo", mode: 0o755
+    system("chmod a+x $CREW_PREFIX/bin/repo")
```
uberhacker, line 25: This line is not needed.  The `mode: 0o755` above makes it executable.

packages/repo.rb
```diff
@@ -0,0 +1,29 @@
+
+require 'package'
+ 
+class Repo < Package
+  description 'Android Repo'
+  homepage 'https://gerrit.googlesource.com/git-repo/'
+  version '1.0'
+  license 'Apache License 2'      
+  compatibility 'x86_64' 
+  source_url 'https://storage.googleapis.com/git-repo-downloads/repo'
```
satmandu, line 10: Would it be better to use `repo` from the github mirror?

It appears to have tagged releases there.

And then you could have something like
```ruby
  version '2.32'
  license 'Apache License 2'      
  compatibility 'all' 
  source_url "https://github.com/GerritCodeReview/git-repo/raw/v#{version}/repo"
```

satmandu (changes requested):

packages/repo.rb
satmandu: Another way to do it would be to use the git functionality we have:

```ruby
  version '2.32'
  license 'Apache License 2'      
  compatibility 'all' 
  source_url "https://github.com/GerritCodeReview/git-repo.git"
  git_hashtag "v#{version}"
```
(That way, you don't need to calculate the sha256 of the file.)

satmandu:
Just need to update `tools/packages.yaml`!

packages/repo.rb
```diff
@@ -0,0 +1,24 @@
+require 'package'
+ 
+class Repo < Package
+  description 'The Multiple Git Repository Tool'
+  homepage 'https://gerrit.googlesource.com/git-repo/'
+  version '2.35'
+  license 'Apache 2.0'
+  compatibility 'x86_64'
```
satmandu, line 8: This can also just be
```ruby
Compatibility 'all'
```
as there is nothing that makes this x86_64 only...

satmandu (commented):
Almost there!

tools/packages.yaml
```diff
@@ -7266,6 +7266,11 @@ url: https://x.org/archive/individual/app
 activity: none
 ---
 kind: url
+name: repo
+url: https://gerrit.googlesource.com/git-repo
```
satmandu, line 7270: @uberhacker does this URL work for your update check script?

Does `https://gerrit.googlesource.com/git-repo/+refs` ?

(Working with his update check script is the important for the urls we have in packages.yaml.)

satmandu (approved):

PR #8511, opened 2023-07-25
Add support for crew build without setting a CREW_LOCAL_MANIFEST_PATH

Zopolis4:
Useful for #8414, and allows you to run `crew build` on a local version of crew without needing to set a `CREW_LOCAL_MANIFEST_PATH`.

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=neemfisk CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,5 +1,5 @@
 # Defines common constants used in different parts of crew
-CREW_VERSION = '1.35.2'
+CREW_VERSION = '1.35.3'
 
 # kernel architecture
 KERN_ARCH = `uname -m`.chomp
@@ -107,7 +107,7 @@ CREW_GITHUB_REPO = 'https://github.com/chromebrew/chromebrew.git'
 
 # Local constants for contributors.
 repo_root = `git rev-parse --show-toplevel 2> /dev/null`.chomp.to_s
-if repo_root.empty? || File.basename(repo_root) != 'chromebrew'
+if repo_root.empty?
   Dir.chdir '../..' do
     repo_root = `git rev-parse --show-toplevel 2> /dev/null`.chomp.to_s
   end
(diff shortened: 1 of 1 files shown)


lib/const.rb
```diff
@@ -107,15 +107,15 @@
...
     repo_root = `git rev-parse --show-toplevel 2> /dev/null`.chomp.to_s
   end
 end
 CREW_LOCAL_REPO_ROOT = repo_root
 CREW_LOCAL_REPO_BASE = CREW_LOCAL_REPO_ROOT.empty? ? '' : File.basename(CREW_LOCAL_REPO_ROOT)
 CREW_LOCAL_MANIFEST_PATH = if ENV['CREW_LOCAL_MANIFEST_PATH'].to_s.empty?
-                             CREW_LOCAL_REPO_BASE == CREW_GITHUB_ACCOUNT ? "#{CREW_LOCAL_REPO_ROOT}/manifest" : ''
+                             "#{CREW_LOCAL_REPO_ROOT}/manifest"
                            else
                              ENV.fetch('CREW_LOCAL_MANIFEST_PATH', nil)
```
satmandu, line 120: The goal is to have a place where the manifest files can be put after a build, where they won't be overwritten on a crew update, right?

Maybe it is better to have an error message if `CREW_LOCAL_REPO_BASE` is empty to have the user specify a `CREW_LOCAL_MANIFEST_PATH` to put those files?

Zopolis4: This is simply for users running `crew build`. imo, the odds that they then unpack and install those builds, making it so the manifests are then relevant, are low. This PR just aims to reduce the barrier to entry and have `crew build <package>` "just work". (Also, I have a faint memory of being unable to install packages from source due to this issue, which would just affect regular users and is something we probably want to avoid)

uberhacker: I think @Zopolis4 is correct.  There is no need for the conditional here since we already set `CREW_GITHUB_ACCOUNT`.

satmandu (approved):

PR #8531, opened 2023-08-03
Detect rebuilds via changes in binary_sha256 when updating

Zopolis4:
I noticed that we sometimes bump the revision of a package for no other reason than a rebuild, i.e. 1c8249a. 

Now, rebuilt packages will simply be detected as needing an update, with no need to bump the revision of a package. 
Of course, in cases where we are patching the package or otherwise modifying it, a revision bump is still advisable.

Tested quite extensively on `x86_64`.

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=quintessence CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,5 +1,5 @@
 # Defines common constants used in different parts of crew
-CREW_VERSION = '1.35.4'
+CREW_VERSION = '1.35.5'
 
 # kernel architecture
 KERN_ARCH = `uname -m`.chomp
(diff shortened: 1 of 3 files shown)


lib/const.rb
```diff
@@ -1,5 +1,5 @@
 # Defines common constants used in different parts of crew
-CREW_VERSION = '1.35.3'
+CREW_VERSION = '1.35.4'
```
uberhacker, line 2: This needs to be synced with master because we're currently on `1.35.4` now.

Zopolis4: Fixed.

uberhacker (approved):

PR #8532, opened 2023-08-03
Xzutils 5.4.3 => 5.4.4

uberhacker:
Currently with autotools builds there is no way to control whether tests should run or not.  This update adds a `run_tests` property (demonstrated in the xzutils update) to execute the self.check section if set, otherwise the check will be skipped.  This would be desirable if the autotools configuration does not include tests or if `make test` is needed instead.  I also removed the `Running tests...` during builds which previously printed the message regardless if tests were run or not.


satmandu (approved):
If it compiled fine with the changes...

PR #8541, opened 2023-08-04
Fix autotools testing buildsystem error

Zopolis4:
Not really sure what happened here to cause an issue in the first place, but this seems to fix the testing issues in the autotools buildsystem. 

Tested on `x86_64`.

```
CREW_TESTING_REPO=https://github.com/Zopolis4/chromebrew.git CREW_TESTING_BRANCH=dediv CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/buildsystems/autotools.rb
+++ b/lib/buildsystems/autotools.rb
@@ -2,9 +2,7 @@ require 'fileutils'
 require 'package'
 
 class Autotools < Package
-  def self.configure_options(options = '')
-    return (@configure_options = options if options)
-  end
+  property :configure_options
 
   def self.build
     puts "Additional configure_options being used: #{@configure_options.nil? || @configure_options.empty? ? '<no configure_options>' : @configure_options}".orange
@@ -34,9 +32,7 @@ class Autotools < Package
   end
 
   def self.check
-    if run_tests
-      puts 'Testing with make check.'.orange
-      system 'make', 'check'
-    end
+    puts 'Testing with make check.'.orange if @run_tests
+    system 'make', 'check' if @run_tests
   end
 end
--- a/lib/buildsystems/cmake.rb
+++ b/lib/buildsystems/cmake.rb
@@ -1,13 +1,7 @@
 require 'package'
 
 class CMake < Package
-  def self.cmake_options(options = '')
-    return (@cmake_options = options if options)
-  end
-
-  def self.check?(bool = true)
-    return (@check = bool)
-  end
+  property :cmake_options
 
   def self.build
     puts "Additional cmake_options being used: #{@cmake_options.nil? || @cmake_options.empty? ? '<no cmake_options>' : @cmake_options}".orange
@@ -21,9 +15,8 @@ class CMake < Package
     system "DESTDIR=#{CREW_DEST_DIR} #{CREW_NINJA} -C builddir install"
   end
 
-  if @check
-    def self.check
-      system "#{CREW_NINJA} -C builddir test"
-    end
+  def self.check
+    puts "Testing with #{CREW_NINJA} test.".orange if @run_tests
+    system "#{CREW_NINJA} -C builddir test" if @run_tests
   end
 end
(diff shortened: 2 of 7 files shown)


uberhacker:
But this doesn't run the tests though; hence the change.  Try building xzutils.

Zopolis4:
I've taken the opportunity to do a larger refactoring, but this does fix the issue.

bin/crew
```diff
@@ -434,8 +434,6 @@ def help(pkgName = nil)
         puts "Use the 'no_zstd' property for the alternate xz compression algorithm."
       when 'patchelf'
         puts "Use the 'patchelf' property for patchelf execution."
-      when 'run_tests'
-        puts "Use the 'run_tests' property to execute make check tests."
```
uberhacker, line 438: We'll still need this for help, right?

lib/buildsystems/autotools.rb
```diff
@@ -2,12 +2,10 @@
...
 
 class Autotools < Package
-  def self.configure_options(options = '')
-    return (@configure_options = options if options)
-  end
+  property :configure_options
+  boolean_property :run_tests
 
   def self.build
-    puts "Additional configure_options being used: #{@configure_options.nil? || @configure_options.empty? ? '<no configure_options>' : @configure_options}".orange
```
uberhacker, line 10: This is still a nice confirmation message though, correct?

lib/buildsystems/autotools.rb
```diff
@@ -34,9 +32,6 @@ def self.install
   end
 
   def self.check
-    if run_tests
-      puts 'Testing with make check.'.orange
```
uberhacker, line 38: Needed to confirm tests are running?

lib/buildsystems/cmake.rb
```diff
@@ -1,16 +1,10 @@
...
 class CMake < Package
-  def self.cmake_options(options = '')
-    return (@cmake_options = options if options)
-  end
-
-  def self.check?(bool = true)
-    return (@check = bool)
-  end
+  property :cmake_options
+  boolean_property :run_tests
```
uberhacker, line 5: Would it make sense to define the `run_tests` property in package.rb since it's used generically with all build systems?

Zopolis4: I thought it'd be better to keep it in each individual buildsystem, since it's not used by non-buildsystem packages. Fine with either, though.

uberhacker: It just follows the DRY (don't repeat yourself) principle.

Zopolis4: Fair enough. I'll update the PR when I get the chance.

lib/buildsystems/autotools.rb
uberhacker: We could compromise here and only display with the `-v` or `--verbose` option.

Zopolis4: My (petty) issue was with the extra line in the buildsystem, not with the display while building. Your logic makes sense, I'll re-add this line (and the others like it)

lib/buildsystems/cmake.rb
```diff
@@ -21,9 +15,7 @@ def self.install
     system "DESTDIR=#{CREW_DEST_DIR} #{CREW_NINJA} -C builddir install"
   end
 
-  if @check
-    def self.check
-      system "#{CREW_NINJA} -C builddir test"
-    end
+  def self.check
+    system "#{CREW_NINJA} -C builddir test" if @run_tests
```
uberhacker, line 19: Add `puts "Testing with #{CREW_NINJA} test.".orange if @run_tests`?

lib/buildsystems/meson.rb
```diff
@@ -22,9 +16,7 @@ def self.install
     system "DESTDIR=#{CREW_DEST_DIR} #{CREW_NINJA} -C builddir install"
   end
 
-  if @check
-    def self.check
-      system "#{CREW_NINJA} -C builddir test"
-    end
+  def self.check
+    system "#{CREW_NINJA} -C builddir test" if @run_tests
```
uberhacker, line 20: Add `puts "Testing with #{CREW_NINJA} test.".orange if @run_tests`?

uberhacker (approved):

PR #8547, opened 2023-08-09
Remove obsolete tools

Zopolis4:
We've got a bunch of tools, and we don't need most of them anymore. So I got rid of them.

TODO:
- What is `getrealdeps.sh`? What do we use it for? Is it still relevant?
- Do we still use `licenses.txt` anymore? Is it up-to-date?
- What about the remaining group of things: `check.rb`, `check_updates.sh`, `compare_packages.sh`, `core_packages.txt`, `packages.yaml`? I know for sure that once I get around to implementing a project of mine, they'll be obsolete and we wont need them. But what about now? Do we need all of them? What *exactly* do we use them for?

Given that the scripts and files I'm wondering about were written by you guys, I'm hoping you'll be able to answer my questions.


satmandu:
https://github.com/chromebrew/chromebrew/pull/7472

Zopolis4:
Thanks. So, as I understand it, `getrealdeps.sh` just basically adds missing dependencies to packages via checking the libraries they need? (I also see that it adds `glibc` as a dependency to most packages, which feels a little redundant-- couldn't we just always install `glibc`, or if we really need to, have `glibc` as an implied dependency for all packages without `is_musl`?)

satmandu:
I use `getrealdeps.sh` as part of my build scripts to automatically add missing dependencies.

It's nice not to imply dependencies when they don't exist, so let's not add glibc to packages that don't need it. Packages with no executables may, for instance, not require a glibc dependency.

uberhacker:
> I use `getrealdeps.sh` as part of my build scripts to automatically add missing dependencies.
> 
> It's nice not to imply dependencies when they don't exist, so let's not add glibc to packages that don't need it. Packages with no executables may, for instance, not require a glibc dependency.

Is it possible to Rubify this script and integrate it in with the other crew commands?

satmandu:
The script heavily makes use of `grep`... so maybe? It's not a high priority for me.

Zopolis4:
I'm in favour of rubifying it eventually, although I feel its better suited as a script rather than as part of crew. At any rate, now I know what it does, and I'll update the PR soon.

(I still am unsure about the other ones tho)

satmandu (approved):
Looks fine to me. Do we need to keep any of the older scripts at all @uberhacker ? Do you need them?

PR #8571, opened 2023-08-19
Mutt 2.2.10 => 2.2.11

uberhacker:
All tests passed.

### DIFF
--- a/packages/mutt.rb
+++ b/packages/mutt.rb
@@ -1,50 +1,46 @@
-require 'package'
+require 'buildsystems/autotools'
 
-class Mutt < Package
+class Mutt < Autotools
   description 'Mutt is a small but very powerful text-based mail client for Unix operating systems.'
   homepage 'http://mutt.org/'
-  version '2.2.10'
+  version '2.2.11'
   license 'GPL-2'
   compatibility 'all'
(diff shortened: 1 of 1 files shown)


packages/mutt.rb
```diff
@@ -1,50 +1,46 @@
...
+    --enable-hcache"
+
+  run_tests
+
+  def self.install
+    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/etc/mail"
     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/etc/env.d/"
-    @muttenv = <<~MUTTEOF
+    File.write "#{CREW_DEST_PREFIX}/etc/env.d/mutt", <<~MUTTEOF
```
satmandu, line 41: Should we use FileUtils.install for this file?

uberhacker: I suppose but how would that work with a heredoc?  I figured since we don't need execute permissions on this file, the default read/write perms would be acceptable.

satmandu: Perms should be 0644, right?

uberhacker: Yes, I believe so.

satmandu (approved):

PR #8587, opened 2023-08-24
Improve logic for detecting package updates

Zopolis4:
The previous logic had a number of issues, so I fixed them. I've also added a warning for when packages have updated versions but no updated binaries, and I'll be using this in 8414 (eventually). 

I also touched up the logic in the installer script.
```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=touchup crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -665,18 +665,25 @@ def update
       puts "Package file for #{package[:name]} not found. :(".lightred if @opt_verbose
       next
     end
-    if package[:binary_sha256] && package[:binary_sha256] != @pkg.get_binary_sha256(@device[:architecture])
-      canBeUpdated += 1
-      puts "#{@pkg.name} could be updated (rebuild)"
-    end
-    if package[:version] != @pkg.version
-      canBeUpdated += 1
+    differentVersion = (package[:version] != @pkg.version)
+    hasSha = !(@pkg.get_binary_sha256(@device[:architecture]).to_s.empty? || package[:binary_sha256].to_s.empty?)
+    differentSha = hasSha && package[:binary_sha256] != @pkg.get_binary_sha256(@device[:architecture])
+
+    canBeUpdated += 1 if differentVersion || differentSha
+
+    if differentVersion && !differentSha && hasSha
+      canBeUpdated -= 1
+      puts "#{@pkg.name} has a version change but does not have updated binaries".yellow
+    elsif differentVersion
       puts "#{@pkg.name} could be updated from #{package[:version]} to #{@pkg.version}"
+    elsif !differentVersion && differentSha
+      puts "#{@pkg.name} could be updated (rebuild)"
     end
   end
 
   if canBeUpdated.positive?
-    puts "\nRun `crew upgrade` to update all packages or `crew upgrade <package1> [<package2> ...]` to update specific packages."
+    puts "\n#{canBeUpdated} packages can be updated."
+    puts 'Run `crew upgrade` to update all packages or `crew upgrade <package1> [<package2> ...]` to update specific packages.'
   else
     puts 'Your software is up to date.'.lightgreen
   end
--- a/install.sh
+++ b/install.sh
@@ -209,14 +209,12 @@ for package in $BOOTSTRAP_PACKAGES; do
 
 done
 
-# Create the device.json file if it doesn't exist.
-if [ ! -f "${CREW_CONFIG_PATH}/device.json" ]; then
-  cd "${CREW_CONFIG_PATH}"
-  echo_info "\nCreating new device.json."
-  jq --arg key0 'architecture' --arg value0 "${ARCH}" \
-    --arg key1 'installed_packages' \
-    '. | .[$key0]=$value0 | .[$key1]=[]' <<<'{}' > device.json
-fi
+# Create the device.json file.
+cd "${CREW_CONFIG_PATH}"
+echo_info "\nCreating device.json."
+jq --arg key0 'architecture' --arg value0 "${ARCH}" \
+  --arg key1 'installed_packages' \
+  '. | .[$key0]=$value0 | .[$key1]=[]' <<<'{}' > device.json
 
 # Functions to maintain packages.
 
@@ -281,16 +279,9 @@ function extract_install () {
 
 function update_device_json () {
   cd "${CREW_CONFIG_PATH}"
-
-  if [[ $(jq --arg key "$1" -e '.installed_packages[] | select(.name == $key )' device.json) ]]; then
-    echo_intra "Updating version number of ${1} in device.json..."
-    ver_num=$(jq --arg key0 "$1" --arg value0 "$2" '(.installed_packages[] | select(.name == $key0) | .version) |= $value0' device.json)
-    cat <<< "${ver_num}" > device.json
-  else
-    echo_intra "Adding new information on ${1} to device.json..."
-    new_info=$(jq --arg key0 "$1" --arg value0 "$2" '.installed_packages |= . + [{"name": $key0, "version": $value0}]' device.json)
-    cat <<< "${new_info}" > device.json
-  fi
+  echo_intra "Adding new information on ${1} to device.json..."
+  new_info=$(jq --arg name "$1" --arg version "$2" --arg sha256 "$3" '.installed_packages |= . + [{"name": $name, "version": $version, "binary_sha256": $sha256}]' device.json)
+  cat <<< "${new_info}" > device.json
 }
 
 echo_info "Downloading Bootstrap packages...\n"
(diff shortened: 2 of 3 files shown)


bin/crew
```diff
@@ -655,18 +655,25 @@ def update
...
-      canBeUpdated += 1
-      puts "#{@pkg.name} could be updated (rebuild)"
-    end
-    if package[:version] != @pkg.version
-      canBeUpdated += 1
+    differentVersion = package[:version] != @pkg.version ? true : false
+    hasSha = @pkg.get_binary_sha256(@device[:architecture]).to_s.empty? || package[:binary_sha256].to_s.empty? ? false : true
+    differentSha = hasSha && package[:binary_sha256] != @pkg.get_binary_sha256(@device[:architecture]) ? true : false
+
+    canBeUpdated += differentVersion || differentSha ? 1 : 0
```
supechicken, line 662: More straightforward?
```ruby
canBeUpdated += 1 if differentVersion || differentSha
```

bin/crew
```diff
@@ -655,18 +655,25 @@ def update
       puts "Package file for #{package[:name]} not found. :(".lightred if @opt_verbose
       next
     end
-    if package[:binary_sha256] && package[:binary_sha256] != @pkg.get_binary_sha256(@device[:architecture])
-      canBeUpdated += 1
-      puts "#{@pkg.name} could be updated (rebuild)"
-    end
-    if package[:version] != @pkg.version
-      canBeUpdated += 1
+    differentVersion = package[:version] != @pkg.version ? true : false
```
supechicken, line 658: Do not use `if` with boolean results, the condition is already a boolean:
```ruby
differentVersion = (package[:version] != @pkg.version)
hasSha = @pkg.get_binary_sha256(@device[:architecture]).nil? || package[:binary_sha256].nil?
differentSha = (hasSha && package[:binary_sha256] != @pkg.get_binary_sha256(@device[:architecture]))
```

Zopolis4:
Your proposed changes break the detection of rebuilds, both in the underlying systems for `crew update` and `crew upgrade`. I'll look into what specifically happens later, but just letting you know-- the current solutions I settled on came after quite an arduous testing process, and this area of code in particular seems to be quite temperamental.

Zopolis4:
Specifically, your changes to the detection of `hasSha` break things. Everything else seems to work fine.

Zopolis4:
I've taken all of your changes except for the ones to `hasSha`. If you can find what the issue was, I'd be happy to accept a revised version, but as for now I'm labelling this pr as good to merge. 

This PR fixes a pretty annoying bug, and I need the changes in it for 8414, so I think it makes sense to merge it now and work on code refactoring later.

install.sh
```diff
@@ -184,14 +184,12 @@ for package in $BOOTSTRAP_PACKAGES; do
...
-# Create the device.json file if it doesn't exist.
-if [ ! -f "${CREW_CONFIG_PATH}/device.json" ]; then
-  cd "${CREW_CONFIG_PATH}"
-  echo_info "\nCreating new device.json."
-  jq --arg key0 'architecture' --arg value0 "${ARCH}" \
-    --arg key1 'installed_packages' \
-    '. | .[$key0]=$value0 | .[$key1]=[]' <<<'{}' > device.json
-fi
+# Create the device.json file.
+cd "${CREW_CONFIG_PATH}"
```
satmandu, line 188: This overwrites any existing package.json file, which means that running the installer over an older install (in case something is broken) will now remove all info about existing packages. 

Is that good?

Can we pick up the existing installed packages when doing this and import into a device.json after doing some sanitizing?

We should at the minimum move package.json into a package.json.`date` if it already exists?

Zopolis4: The current installer already does this by clearing out `CREW_PREFIX`. There's no particular value in keeping a device.json when all packages have been deleted.

bin/crew
```diff
@@ -646,18 +646,25 @@ def update
...
+
+    canBeUpdated += 1 if differentVersion || differentSha
+
+    if differentVersion
       puts "#{@pkg.name} could be updated from #{package[:version]} to #{@pkg.version}"
+    elsif !differentVersion && differentSha
+      puts "#{@pkg.name} could be updated (rebuild)"
+    elsif differentVersion && !differentSha && hasSha
+      canBeUpdated -= 1
+      puts "#{@pkg.name} has a version change but does not have updated binaries".yellow
```
uberhacker, line 661: This condition is not possible since the first condition passes.  What you likely want is:
```
    if differentVersion
      puts "#{@pkg.name} could be updated from #{package[:version]} to #{@pkg.version}"
    elsif differentSha
      puts "#{@pkg.name} could be updated (rebuild)"
    end
```

Zopolis4: Fixed.

bin/crew
```diff
@@ -647,18 +647,25 @@ def update
...
       next
     end
-    if package[:binary_sha256] && package[:binary_sha256] != @pkg.get_binary_sha256(@device[:architecture])
-      canBeUpdated += 1
-      puts "#{@pkg.name} could be updated (rebuild)"
-    end
-    if package[:version] != @pkg.version
-      canBeUpdated += 1
+    differentVersion = (package[:version] != @pkg.version)
+    hasSha = @pkg.get_binary_sha256(@device[:architecture]).to_s.empty? || package[:binary_sha256].to_s.empty? ? false : true
```
uberhacker, line 651: No need for the immediate if.  IOW, ` ? false : true` can be removed.

Zopolis4: Assuming you mean turning 
```ruby
hasSha = @pkg.get_binary_sha256(@device[:architecture]).to_s.empty? || package[:binary_sha256].to_s.empty? ? false : true
```

into 
```ruby
hasSha = @pkg.get_binary_sha256(@device[:architecture]).to_s.empty? || package[:binary_sha256].to_s.empty?
```
then no, it cannot be removed. That breaks it.

bin/crew
```diff
@@ -681,18 +681,25 @@ def update
...
```
supechicken, line 685: Don't use `if` with boolean result, remove `? false : true` and use `!` to flip the boolean:
```ruby
hasSha = !(@pkg.get_binary_sha256(@device[:architecture]).to_s.empty? || package[:binary_sha256].to_s.empty?)
```

bin/crew
```diff
@@ -681,18 +681,25 @@ def update
...
+    elsif differentVersion
       puts "#{@pkg.name} could be updated from #{package[:version]} to #{@pkg.version}"
+    elsif !differentVersion && differentSha
+      puts "#{@pkg.name} could be updated (rebuild)"
     end
   end
 
   if canBeUpdated.positive?
-    puts "\nRun `crew upgrade` to update all packages or `crew upgrade <package1> [<package2> ...]` to update specific packages."
+    puts "\n#{canBeUpdated} packages can be updated."
```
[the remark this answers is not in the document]
Zopolis4: For only two lines, I don't really see the benefit.

bin/crew
Zopolis4: This version of `differentSha` is triggering rubocop now...

Zopolis4:
@satmandu would you mind taking a look at this? Without this bugfix, #9201 will trigger unnecessary reinstalls of the packages it changed.

install.sh
```diff
@@ -209,14 +209,12 @@ for package in $BOOTSTRAP_PACKAGES; do
...
 
-# Create the device.json file if it doesn't exist.
-if [ ! -f "${CREW_CONFIG_PATH}/device.json" ]; then
-  cd "${CREW_CONFIG_PATH}"
-  echo_info "\nCreating new device.json."
-  jq --arg key0 'architecture' --arg value0 "${ARCH}" \
-    --arg key1 'installed_packages' \
-    '. | .[$key0]=$value0 | .[$key1]=[]' <<<'{}' > device.json
-fi
+# Create the device.json file.
```
satmandu, line 212: We want to be able to do a repair install, which only updates the device.json. why break that functionality?

Zopolis4: We don't have any files in `CREW_PREFIX` by this stage of the installer, so there's no point checking for a `device.json` as we know there will not be one.

satmandu: That's not true if you're doing a reinstall...

satmandu: @uberhacker are we recapitulating a previous discussion here?

Zopolis4: We clear out all of `CREW_PREFIX` as part of the installation process. See [here](https://github.com/chromebrew/chromebrew/blob/master/install.sh#L50-L58).

I'm happy to discuss *whether* we should be doing that at a later date, but this PR is not the place to discuss it. 

All I'm doing here is removing a redundant check.

satmandu (approved):

PR #8646, opened 2023-09-12
Fix installation on aarch64 devices

### LINKED ISSUE #8635: 🐞 Chromebrew install fails. (Chrome os 116)
**Describe the bug**

Chromebrew install fails with the error "Your device is not supported by Chromebrew yet, installing as armv7l." and then exits out to crosh
**To Reproduce**

1. I executed this command:

      ```shell
      chronos@juniper-rev4 / $ exec bash --init-file <(curl -Ls git.io/vddgY)

      ```

      Output:

Welcome to Chromebrew!
Please enter the developer mode password
Password: 
Your device is not supported by Chromebrew yet, installing as armv7l.

**Expected behavior**
Chromebrew installs correctly
**Screenshots**

**Environment: Paste the output of `crew sysinfo -v` between the two `***` below**

***
bash: crew: command not found
***

**Additional context**
If my memory serves correctly, chromebrew installed correctly in ~2021 on the same chromebook

Zopolis4:
oops

Resolves #8635.

Unable to test because I don't have an aarch64 docker script, although I am confident this change will solve the issue.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -104,7 +104,6 @@ function curl () {
 
 if [[ "$USER_SPACE_ARCH" == 'armv7l' ]] && [[ "$ARCH" == 'aarch64' ]]; then
     echo_error "Your device is not supported by Chromebrew yet, installing as armv7l."
-    exit 1
 fi
 
 if [ -d "${CREW_PREFIX}" ]; then


uberhacker:
Actually, what really makes sense is something like below:
```
if [[ "$USER_SPACE_ARCH" == 'aarch64' ]] && [[ "$ARCH" == 'aarch64' ]]; then
  echo_error "Your device is not supported by Chromebrew.  Binaries are not available yet."
  exit 1
fi
```

Zopolis4:
As I understand it, we still do support aarch64 userspace, we just install armv7l binaries. @satmandu wrote the armv7l on aarch64 support, though, so they might know what the correct solution is.

uberhacker:
> As I understand it, we still do support aarch64 userspace, we just install armv7l binaries. @satmandu wrote the armv7l on aarch64 support, though, so they might know what the correct solution is.

No, we only support armv7l userspace.  The architecture can still be aarch64, however.

satmandu:
> > As I understand it, we still do support aarch64 userspace, we just install armv7l binaries. @satmandu wrote the armv7l on aarch64 support, though, so they might know what the correct solution is.
> 
> No, we only support armv7l userspace.  The architecture can still be aarch64, however.

We do experimentally support aarch64 userspace, since a stub armv7l userspace is also installed therein...

uberhacker:
> > > As I understand it, we still do support aarch64 userspace, we just install armv7l binaries. @satmandu wrote the armv7l on aarch64 support, though, so they might know what the correct solution is.
> > 
> > 
> > No, we only support armv7l userspace.  The architecture can still be aarch64, however.
> 
> We do experimentally support aarch64 userspace, since a stub armv7l userspace is also installed therein...

No, if we supported aarch64 userspace then we would provide aarch64 binaries.  As soon as ChromeOS supports it then Chromebrew can support it.

satmandu:
> > > > As I understand it, we still do support aarch64 userspace, we just install armv7l binaries. @satmandu wrote the armv7l on aarch64 support, though, so they might know what the correct solution is.
> > > 
> > > 
> > > No, we only support armv7l userspace.  The architecture can still be aarch64, however.
> > 
> > We do experimentally support aarch64 userspace, since a stub armv7l userspace is also installed therein...
> 
> No, if we supported aarch64 userspace then we would provide aarch64 binaries.  As soon as ChromeOS supports it then Chromebrew can support it.

> We don't support aarch64 userspace.

We _do unofficially support_ installs on aarch64 userspace machines, since they include an armv7l userspace stub and GLIBC...

satmandu (approved):
This should at least get around this error...

PR #8755, opened 2023-10-07
crew: Retry `make` with single thread if failed

supechicken:
Suggested by @satmandu:

> It would be nice to automatically retry commands which have a -j#{CREW_NPROC} passed to it which fail automatically with -j1 so we can see the command that failed...

Tested on `x86_64`

```
CREW_TESTING_REPO=https://github.com/supechicken/chromebrew.git CREW_TESTING_BRANCH=retry_make CREW_TESTING=1 crew update
```

### DIFF
--- a/lib/package.rb
+++ b/lib/package.rb
@@ -259,7 +259,7 @@ class Package
 
     # add "-j#" argument to "make" at compile-time, if necessary
 
-    # Order of precedence to assign the number of processors:
+    # Order of precedence to assign the number of threads:
     # 1. The value of '-j#' from the package make argument
     # 2. The value of ENV["CREW_NPROC"]
     # 3. The value of `nproc`.strip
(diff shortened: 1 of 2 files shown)


lib/package.rb
```diff
@@ -276,19 +276,39 @@ def self.system(*args, **opt_args)
...
+        # replace placeholder with '-j#' arg and execute the actual command
+        Kernel.system(env, *cmd_args.map { |arg| arg.sub('<<<CREW_NPROC>>>', "-j#{make_threads}") }, **opt_args)
+      else
+        Kernel.system(env, *cmd_args, **opt_args)
+      end
+    rescue RuntimeError => e
+      if modded_make_cmd && make_threads != 1
+        # retry with single thread if command is `make` and is modified by crew
+        warn "Command \"#{cmd_args.map { |arg| arg.sub('<<<CREW_NPROC>>>', "-j#{make_threads}") } .join(' ')}\" failed, retrying with \"-j1\"...".yellow
+        make_threads = 1
```
uberhacker, line 306: What does setting this value do if it's not being used later?

satmandu: @supechicken Do you want to answer this so we can get it merged?

supechicken: Sorry for late reply😅 This variable will be reused in line 296-301 after calling `retry` (line 307) in this `rescue` part.

uberhacker: OK, so this acts like a loop then.

uberhacker: What can we do to prevent an endless loop?  Should we limit the number of retries?

supechicken: This retry logic will be skipped when `make_threads` is `1`

satmandu: @uberhacker Any other objections?

satmandu (approved):

lib/package.rb
uberhacker: Bump crew version?

PR #8772, opened 2023-10-13
Add --no-pager to meson configure lines

satmandu:
- Meson builds pause on `meson configure` to show the options being used, unless one is running `meson configure` from inside a script. But the pause halts the build until one pages through the options display. This adds the `--no-pager` option so that the options are still displayed, but the build isn't paused.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=meson_no_pager CREW_TESTING=1 crew update
```


uberhacker (approved):
I'm fine with this change but shouldn't we be converting these packages over to use the buildsystem?  I guess we can cross that bridge on the next updates.

PR #8805, opened 2023-10-19
Nftables 1.0.8 => 1.0.9

uberhacker:
All tests passed.

### DIFF
--- a/packages/nftables.rb
+++ b/packages/nftables.rb
@@ -1,26 +1,26 @@
 # Adapted from Arch Linux nftables PKGBUILD at:
 # https://github.com/archlinux/svntogit-packages/raw/packages/nftables/trunk/PKGBUILD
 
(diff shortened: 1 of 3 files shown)


packages/nftables.rb
```diff
@@ -1,26 +1,26 @@
...
 
-require 'package'
+require 'buildsystems/autotools'
 
-class Nftables < Package
+class Nftables < Autotools
   description 'Netfilter tables userspace tools'
   homepage 'https://netfilter.org/projects/nftables/'
-  version '1.0.8'
+  version '1.0.9'
```
satmandu, line 9: We should tag this with `-py3.12`...

uberhacker: How do we include the python files?  Do we need to update the package?

satmandu (approved):

PR #8820, opened 2023-10-23
Fix Repology workflow

Zopolis4:
I hate github actions :(

Anyways, fixed the workflow again by ensuring that crew does not try to take everything as an argument.

### DIFF
--- a/.github/workflows/Repology.yml
+++ b/.github/workflows/Repology.yml
@@ -12,8 +12,8 @@ jobs:
         run: |
             sudo docker pull satmandu/crewbuild:amd64
             sudo docker run -t -v $(pwd):/usr/local/json satmandu/crewbuild:amd64 sudo -i -u chronos /bin/bash -c "
-            crew update
-            ruby ../tools/json.rb
+            crew update &&
+            ruby ../tools/json.rb &&
             cp repology.json /usr/local/json"
       - name: Upload JSON arifact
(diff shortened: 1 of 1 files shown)


.github/workflows/Repology.yml
```diff
@@ -12,8 +12,8 @@ jobs:
         run: |
             sudo docker pull satmandu/crewbuild:amd64
             sudo docker run -t -v $(pwd):/usr/local/json satmandu/crewbuild:amd64 sudo -i -u chronos /bin/bash -c "
-            crew update
-            ruby ../tools/json.rb
+            crew update &&
+            ruby ../tools/json.rb &&
             cp repology.json /usr/local/json"
```
satmandu, line 17: Does this need to be
```
crew update && \
ruby ../tools/json.rb && \
cp repology.json /usr/local/json"
```

Zopolis4: It'll work without them

uberhacker: Just so you know, the `&&` means if any command fails the following commands will not execute.

Zopolis4: In this case I'm just using it to separate the commands so crew doesn't misunderstand. Although, the behaviour of not continuing from a fail is probably desired here.

satmandu (approved):

PR #8827, opened 2023-10-24
Fix Repology workflow pt.2

Zopolis4:
done.
working.
tested.
confirmed.

gonna be fun when i need to rework this after fixing `lib/const.rb`


.github/workflows/Repology.yml
```diff
@@ -12,11 +12,12 @@ jobs:
         run: |
             sudo docker pull satmandu/crewbuild:amd64
             sudo docker run -t -v $(pwd):/usr/local/json satmandu/crewbuild:amd64 sudo -i -u chronos /bin/bash -c "
-            crew update &&
-            ruby ../tools/json.rb &&
-            cp repology.json /usr/local/json"
+            crew update && \
```
satmandu, line 15: I knew it!

satmandu (approved):

Issue #8779, opened 2023-10-15
🐞 Bug latest ChromeOS versions block sudo in Crosh terminal, you have to go to vty2 and once crew is installed there you can't source the .bashrc because it utilizes sudo somewhere in the chain


espoelstra:
**Describe the bug**
The latest ChromeOS versions block sudo in Crosh terminal, you have to go to vty2 to login directly as root.

Downloading crew via the pipe to bash also seems to fail, so I've had to use `cd /tmp; curl -Ls git.io/vddgY -o chromebrew.sh; sudo -u chronos bash chromebrew.sh` to grab it to a file and execute it as not setuid 0 (root) but still allow root to make changes that would have required sudo for a non-root user.

BUT once crew is installed there you can't source the .bashrc because it utilizes sudo somewhere in the chain.

I was able to temporarily set an alias for the root user of `alias crew='sudo -u chronos crew '` so that I could install things which the regular user _should_ still be able to use, but the inability to sudo directly from crosh is going to make things a PITA.

**To Reproduce**

1. I executed this command:

      ```shell
      $ cd /tmp; curl -Ls git.io/vddgY -o chromebrew.sh; bash chromebrew.sh
      ```

      Output:

      ```
      sudo: The "no new privileges" flag is set, which prevents sudo from running as root.
      sudo: If sudo is running in a container, you may need to adjust the container configuration to disable the flag.
      ```

**Expected behavior**

Crew probably needs to update its behavior and warn people of the temporary workarounds required to get things working.

**Workaround**
Ctrl+Alt+F2 to get to vty2, then login as `root` with no password required.

    ```shell
      $ cd /tmp; curl -Ls git.io/vddgY -o chromebrew.sh; sudo -u chronos bash chromebrew.sh
    ```

Then set an alias for the root user of `alias crew='sudo -u chronos crew '` so that you can install things which the regular user _should_ still be able to use.

**Environment: Paste the output of `crew sysinfo -v` between the two `***` below**

***
crew sysinfo -v
<details><summary>Expand</summary>

- Architecture: `x86_64` (`x86_64`)
- Processor vendor: `GenuineIntel`
- User space: `64-bit`
- Chromebrew Kernel version: `5.15`
- Chromebrew Running in Container: `false`

- Chromebrew version: `1.36.2`
- Chromebrew prefix: `/usr/local`
- Chromebrew libdir: `/usr/local/lib64`

[...]

PR #8832, opened 2023-10-25
New package: crew-sudo

supechicken:
`crew-sudo`: Workaround for using `sudo` on ChromeOS `crosh` shell (ChromeOS v117+)

See https://github.com/supechicken/crew-sudo for more information.

Might help #8779, and this package **needs further testing** :) (I will transfer this project to `chromebrew` org once it is ready)

### Usage
Run this in VT-2 shell first to start the sudo daemon
```shell
# the bashrc should do it automatically once you login the VT-2 shell
crew-sudo --daemon
```

And `sudo` should work in `crosh` now (be aware that the original `sudo` command is covered with `crew-sudo` now because of the `PATH` priority, but it should be fine in most cases)

---

Tested on `x86_64`

```
CREW_REPO=https://github.com/supechicken/chromebrew CREW_BRANCH=crew-sudo crew update
```

### DIFF
--- a/packages/crew_sudo.rb
+++ b/packages/crew_sudo.rb
@@ -24,7 +24,7 @@ class Crew_sudo < Package
     FileUtils.ln_s "#{path}/crew-sudo", "#{bindir}/sudo"
     FileUtils.ln_s "#{path}/crew-sudo", "#{bindir}/sudod"
 
-    FileUtils.cp 'autostart/crew-sudo.sh', envdir
+    FileUtils.cp 'autostart/crew-sudo.sh', "#{envdir}/crew_sudo"
   end
 
   def self.postinstall
--- a/packages/crew_sudo.rb
+++ b/packages/crew_sudo.rb
@@ -17,7 +17,7 @@ class Crew_sudo < Package
     envdir = File.join(CREW_DEST_PREFIX, 'etc/env.d')
     path   = File.join(CREW_PREFIX, 'lib/crew-sudo')
 
-    FileUtils.mkdir_p [bindir, File.join(CREW_DEST_DIR, path)]
+    FileUtils.mkdir_p [bindir, envdir, File.join(CREW_DEST_DIR, path)]
     FileUtils.cp_r Dir['*'], File.join(CREW_DEST_DIR, path)
 
     FileUtils.ln_s "#{path}/crew-sudo", "#{bindir}/crew-sudo"
--- a/packages/crew_sudo.rb
+++ b/packages/crew_sudo.rb
@@ -8,7 +8,7 @@ class Crew_sudo < Package
   compatibility 'all'
 
   source_url 'https://github.com/supechicken/crew-sudo.git'
-  git_hashtag 'v1.0'
+  git_hashtag '1.0'
 
   no_compile_needed
 
--- a/packages/crew_sudo.rb
+++ b/packages/crew_sudo.rb
@@ -30,7 +30,7 @@ class Crew_sudo < Package
   def self.postinstall
     warn <<~EOT.lightblue
 
-      In order to make sudo works properly, the crew-sudo daemon need to
+      In order to make sudo work properly, the crew-sudo daemon needs to
       start in the VT-2 shell every boot by:
 
         - Press Ctrl + Alt + -> to enter VT-2 shell
--- a/packages/crew_sudo.rb
+++ b/packages/crew_sudo.rb
@@ -17,7 +17,7 @@ class Crew_sudo < Package
     envdir = File.join(CREW_DEST_PREFIX, 'etc/env.d')
     path   = File.join(CREW_PREFIX, 'lib/crew-sudo')
 
-    FileUtils.mkdir_p bindir, File.join(CREW_DEST_DIR, path)
+    FileUtils.mkdir_p [bindir, File.join(CREW_DEST_DIR, path)]
     FileUtils.cp_r Dir['*'], File.join(CREW_DEST_DIR, path)
 
     FileUtils.ln_s "#{path}/crew-sudo", "#{bindir}/crew-sudo"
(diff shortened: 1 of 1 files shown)


satmandu:
The post install here would be good for the work @uberhacker and I were just talking about for some sort of array inside crew of messages to dump at the very end of all installs...

packages/crew_sudo.rb
```diff
@@ -0,0 +1,43 @@
...
+  version '1.0'
+  license 'GPL-3'
+  compatibility 'all'
+
+  source_url 'https://github.com/supechicken/crew-sudo.git'
+  git_hashtag '1.0'
+
+  no_compile_needed
+
+  def self.install
```
Zopolis4, line 15: Could we offload the install logic to the actual package itself? It feels more appropriate there...

satmandu:
There is a chicken and egg issue here, right?

This needs to be installed before crew can be installed?

satmandu:
@supechicken Here is one way of dumping that message at the end of crew running:

https://github.com/chromebrew/chromebrew/pull/8851

packages/crew_sudo.rb
supechicken: The logic here just copying the files, it sounds better to me to leave it here (an installation script will be needed if we move this logic to the `crew-sudo` repo)

supechicken:
> What is wrong with just typing `shell` from the crosh prompt and continuing from there? :) Ultimately, I'd like to figure out a way to bypass crosh and go directly to a bash shell session. Possible?

@uberhacker Seems you misunderstand the usage of this package😅 This package is for fixing `sudo` in newer ChromeOS milestones, here is the background of it posted by @satmandu on slack:

> Just noticed this in crosh after updating to M117:
> ```
> Sudo commands will not succeed by default.
> If you want to use sudo commands, use the VT-2 shell
> (Ctrl-Alt-{F2/Right arrow/Refresh}) or build the image with the
> login_enable_crosh_sudo USE flag:
>
> $ USE=login_enable_crosh_sudo emerge-$BOARD chromeos-login
> or
> $ USE=login_enable_crosh_sudo cros build-packages --board=$BOARD
> ```
> 

> About the sudo issue, looks like google blocked the suid bit in crosh via the NO_NEW_PRIVS exec flag (resulting in an unusable sudo), which cannot be bypassed unlike nosymfollow.
> So I created a simple script to redirect sudo calls in crosh to a daemon running in VT-2 and call the actual sudo by the daemon.
> Fun fact: most android root solutions work in a similar way (as the suid bit is also blocked in android)
> If anyone is interested in it:
> https://gist.github.com/supechicken/58d530210620d24eea327043b7cb9b3b

packages/crew_sudo.rb
```diff
@@ -0,0 +1,44 @@
...
+class Crew_sudo < Package
+  description 'Workaround for using sudo on ChromeOS crosh shell (ChromeOS v117+)'
+  homepage 'https://github.com/supechicken/crew-sudo'
+  version '1.0'
+  license 'GPL-3'
+  compatibility 'all'
+
+  source_url 'https://github.com/supechicken/crew-sudo.git'
+  #git_hashtag '1.0'
+  git_branch 'main'
```
Zopolis4, line 12: We don't need to set `git_branch` here, given that there's only one.

packages/crew_sudo.rb
Zopolis4: That's kind of the idea... 😅

If we have the choice (which we do), I'd prefer that we offload as much logic to packages as possible.

spoelstraethan:
I just tested this using the syntax provided in the first message but I had to comment out the `postinstall.message` because it wasn't happy with it, but another install command after that tweak it installed. Then after logging out of VT2 and logging back in I saw crew-sudo started a daemon and I was immediately able to switch back to the GUI session and run some commands with sudo and got prompted for my password and they succeeded without error.

This is going to be awesome!

NichoGamerTv:
it asks me for a password when typing the command, im new to chromebrew

packages/crew_sudo.rb
supechicken: Change made. Now the install logic is located at [crew-sudo/install.sh](https://github.com/supechicken/crew-sudo/blob/v1.1/install.sh)

supechicken:
This PR is ready :)

spoelstraethan:
@supechicken has the script itself been updated? I've noticed some weird behaviors like not being able to touch stuff in /usr/local on the Chromebrew as the regular crosh shell user even though all the file permissions are set correctly and I shouldn't have to `sudo`.

supechicken:
@spoelstraethan This is a bug of ChromeOS, see https://github.com/chromebrew/chromebrew/issues/8919#issuecomment-1890331190 for solutions

spoelstraethan:
Using gnome terminal looks like a decent workaround and adding the kernel option to the bootloader shouldn't require breaking dm-verity hopefully.

I'm not surprised some of these bugs only hit those using developer mode to make Chromebooks even more useful.

I did notice the sommelier issue while doing something else today, it gets pulled in with gtk3 even when it isn't requested and seems to have its own issues with permissions. I'm wondering if just having the shortcut for it use `env` to set some alternate paths where it'll have permissions would make the failures go away.

Zopolis4:
The `crew_sudo.rb` package file looks good to me, but I can't test this on real hardware. so someone should probably do that. But after someone tests that and everything looks good, we should be good to go for merge.

satmandu:
@supechicken Can we merge this?

satmandu (approved):

Arewafly:
Hello i find this threat exactly as issue i'm fatching when running a shell command

I try to use the solution of this threat by pressing ctrl+alt+f2 and add the following commad

exec bash --init-file <(curl -is git.io/vddgy)

crew_repo=https://github.com/supechicken/chromebrew crew_branch=crew-sudo crew update 

But, still it's show command not found

Please I need solution, i'm new in chrome os and i want bot it from external drive

PR #8836, opened 2023-10-27
Support running crew on non ChromeOS distros (partially)

Zopolis4:
Was going to rework `lib/const.rb`, but it was too daunting. I still did this though-- it's not perfect, but it will allow us to run scripts that use `lib/const.rb` while not being on chromeos.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=stine crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,7 +1,7 @@
 # lib/const.rb
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.40.2'
+CREW_VERSION = '1.40.1'
 
 # kernel architecture
 KERN_ARCH = `uname -m`.chomp
@@ -50,7 +50,7 @@ CREW_LIB_SUFFIX = ARCH.eql?('x86_64') && Dir.exist?('/lib64') ? '64' : ''
 ARCH_LIB        = "lib#{CREW_LIB_SUFFIX}"
 
 # Glibc version can be found from the output of libc.so.6
-LIBC_VERSION = `/#{ARCH_LIB}/libc.so.6`[/Gentoo ([^-]+)/, 1]
+LIBC_VERSION = `#{`whereis libc.so.6`.split[1]}`[/Gentoo ([^-]+)/, 1]
 
 CREW_PREFIX = ENV.fetch('CREW_PREFIX', '/usr/local')
 


lib/const.rb
```diff
@@ -46,19 +46,21 @@
 # downloaded, e.g. in a docker container.
 ARCH_LIB = ARCH.eql?('x86_64') && Dir.exist?('/lib64') ? 'lib64' : 'lib'
 
+libc_path = `whereis libc.so.6`[11, 1000]
+
 # Glibc version can be found from the output of libc.so.6
-LIBC_VERSION = `/#{ARCH_LIB}/libc.so.6`[/Gentoo ([^-]+)/, 1]
+LIBC_VERSION = `#{libc_path}`[/Gentoo ([^-]+)/, 1]
```
uberhacker, line 52: It's fine to leave this the way it is currently.

Zopolis4: This change is the point of the PR-- it lets crew be run on non ChromeOS distros-- very helpful for CI and testing.

lib/const.rb
```diff
@@ -46,19 +46,21 @@
...
 
 if ENV['CREW_PREFIX'].to_s.empty? || (ENV['CREW_PREFIX'] == '/usr/local')
-  CREW_BUILD_FROM_SOURCE = ENV.fetch('CREW_BUILD_FROM_SOURCE', nil)
   CREW_PREFIX = '/usr/local'
   HOME = Dir.home
 else
-  CREW_BUILD_FROM_SOURCE = 1
```
uberhacker, line 57: Don't we always want to build from source if `CREW_PREFIX != '/usr/local'`?

Zopolis4: Why would that be the case?

satmandu: All current binaries are built with `--prefix=/usr/local` - so if you change CREW_PREFIX, none of the binaries will work...

lib/const.rb
uberhacker: Is the location different on non ChromeOS distros?  I'm thinking probably not?  Do we have an example where it wouldn't be in the `ARCH_LIB` location?

Zopolis4: Debian

uberhacker: Why would anyone in their right mind want to run crew on a Debian system?  You already have a superior package manager available.

Zopolis4: It's helpful for creating scripts and tests that use lib/const.rb, without needing to spin up a Docker instance.

lib/const.rb
```diff
@@ -50,7 +50,8 @@
 ARCH_LIB        = "lib#{CREW_LIB_SUFFIX}"
 
 # Glibc version can be found from the output of libc.so.6
-LIBC_VERSION = `/#{ARCH_LIB}/libc.so.6`[/Gentoo ([^-]+)/, 1]
+libc_path = `whereis libc.so.6`[11, 1000]
```
supechicken, line 53: Use `partition(':')` or `split(':')`?
```ruby
libc_path = `whereis libc.so.6`.partition(': ')[2]
```

Zopolis4:
Closing due to the increased obstacles to running crew on non ChromeOS distros since this PR was opened

PR #8839, opened 2023-10-27
restart crew update if there is a new crew version

satmandu:
image

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=fixupversion crew update
```

### DIFF
--- a/lib/fixup.rb
+++ b/lib/fixup.rb
@@ -135,7 +135,7 @@ NEW_CONST_GIT_COMMIT = `git log -n1 --oneline ../lib/const.rb`.chomp.split.first
 
 unless NEW_CONST_GIT_COMMIT.to_s == CREW_CONST_GIT_COMMIT.to_s
   puts 'Restarting crew update since there is an updated crew version'.lightcyan
-  puts "CREW_REPO=#{CREW_REPO} CREW_BRANCH=#{CREW_BRANCH} crew update".orange
+  puts "CREW_REPO=#{CREW_REPO} CREW_BRANCH=#{CREW_BRANCH} crew update".orange if @opt_verbose
   exec "CREW_REPO=#{CREW_REPO} CREW_BRANCH=#{CREW_BRANCH} crew update"
 end
 
--- a/lib/fixup.rb
+++ b/lib/fixup.rb
@@ -1,3 +1,4 @@
+# lib/fixup.rb
 # Add fixups to be run during crew update here.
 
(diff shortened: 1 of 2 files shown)


lib/fixup.rb
```diff
@@ -124,3 +125,19 @@
     puts "#{pkg[:pkg_name].capitalize} not removed.".lightblue
   end
 end
+
+# Restart crew update if the git commit of ../lib/const.rb loaded in 
+# const.rb is different from the git commit of the potentially updated
+# ../lib/const.rb loaded here after a git pull.
+
+# Handle case of const.rb not yet defining CREW_CONST_GIT_COMMIT.
+CREW_CONST_GIT_COMMIT = '0000' unless defined?(CREW_CONST_GIT_COMMIT)
```
uberhacker, line 134: Is this necessary?

lib/fixup.rb
```diff
@@ -124,3 +125,19 @@
...
+
+NEW_CONST_GIT_COMMIT = `git log -n1 --oneline ../lib/const.rb`.chomp.split.first
```
uberhacker, line 136: Add to const.rb?

satmandu: Local only...

lib/fixup.rb
satmandu: Yes. Because it isn't defined otherwise.

satmandu: If you haven't crew updated to have the variable in const.rb..

lib/fixup.rb
satmandu: Changed variable type...

lib/fixup.rb
uberhacker: So if the constant isn't defined yet doesn't that mean it will always restart `crew update`?  Why not just use regular variables here instead of constants to ensure the correct value is calculated?

satmandu: The variable is only unset when `const.rb` is too old to have set the variable, so really only the first time the user loads this code.

If we don't do the define check for this situation we get a NameError exception...

satmandu: So essentially, when const.rb hasn't defined the variable with the correct git commit information, we should definitely be trying to run `crew update`.

uberhacker (approved):

PR #8845, opened 2023-10-29
fix ambiguous argument '../lib/const.rb' error

satmandu:
- Use absolute path for const.rb in check for git commit.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=const_git crew update
```


uberhacker (approved):
I imagined it was a bit more complicated than I originally thought. :)

PR #8849, opened 2023-10-30
Installation changes for newer Milestones.

### LINKED ISSUE #8823: 🐞 Bug Multiple binaries segfaulting on AMD StoneyRidge when ASLR is enabled.
As reported many times before, various binaries (e.g. `sed`, `bison`, `lzip`, etc.) are segfaulting on AMD StoneyRidge machines.
This seems being related to ASLR, enabled by default on ChromeOS, as disabling it the issue disappears.

Running `echo 0 | sudo tee /proc/sys/kernel/randomize_va_space` those binaries are no longer segfaulting (while they do with `1` or `2`), but obviously is it not an optimal solution from a security standpoint.

https://askubuntu.com/questions/318315/how-can-i-temporarily-disable-aslr-address-space-layout-randomization

satmandu:
Fixes #8823
- Adjust documentation for installation on newer milestone (> M118) devices.
- Rebuild both `lzip`, and `tar` without musl.
- Adjust `repo_root` logic to always work wherever `crew` is run from.
- Add logic to disable ASLR on AMD machines during install.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=amd crew update
```


install.sh
```diff
@@ -51,6 +50,37 @@ echo_out() { echo -e "\e[0;37m${*}${RESET}" >&1; } # Use Gray for program output
 # Skip all checks if running on a docker container.
 [[ -f "/.dockerenv" ]] && CREW_FORCE_INSTALL=1
 
+if [[ "$CREW_FORCE_INSTALL" ]]; then
+  echo_info "In container, so skipping hardware checks"
+elif tty | grep -q /dev/pts1; then
+  echo_success "You are in VT-2."
```
uberhacker, line 56: I could be in VT-3, 4 or 5 too. :)  Maybe we should call it Developer Console.

install.sh
```diff
@@ -51,6 +50,37 @@ echo_out() { echo -e "\e[0;37m${*}${RESET}" >&1; } # Use Gray for program output
...
+  if [ "${EUID}" == "0" ]; then
```
uberhacker: How about if `${USER} != 'chronos'`?

install.sh
satmandu: The entire check is only needed for sudo use... so without using sudo, we may not need it at all...

install.sh
satmandu: We could do that, but then I think there is another PR playing around with using this NOT on chromeos... and the existing code is simpler...

install.sh
```diff
@@ -51,6 +50,37 @@ echo_out() { echo -e "\e[0;37m${*}${RESET}" >&1; } # Use Gray for program output
...
+    echo_info "Please login as 'chronos' and restart the install."
+    echo_info "Note that you can set the chronos password if you login as root and run 'chromeos-setdevpasswd'."
+    exit 1
+  fi
+else
+  # Error out if not in VT-2
+  echo_error "You need to be in the VT-2 shell for the install to work."
```
uberhacker: You need to be in the Developer Console?

install.sh
Zopolis4: To be fair, my PR isin't dealing with actually installing crew-- just being able to run some things in `tools/`.

But, I might look into that in future, and it makes sense not to throw up additional roadblocks.

install.sh
satmandu: Resolving this because I commented out the code...

install.sh
```diff
@@ -51,6 +50,42 @@ echo_out() { echo -e "\e[0;37m${*}${RESET}" >&1; } # Use Gray for program output
 # Skip all checks if running on a docker container.
 [[ -f "/.dockerenv" ]] && CREW_FORCE_INSTALL=1
 
+if [[ "$CREW_FORCE_INSTALL" ]]; then
```
Zopolis4, line 53: Would it not make sense to gate all of the `CREW_FORCE_INSTALL` checks behind one big `if` statement instead of having multiple?

install.sh
```diff
@@ -125,7 +156,7 @@ else
 fi
```
uberhacker, line 156: Remove `sudo`?

satmandu: Done.

install.sh
satmandu: The other conditions ask if `CREW_FORCE_INSTALL` is set manually, so I have left those alone.

But I have combined the parts that make sense combined.

supechicken:
Here are my two cents, since copying/pasting the installation command is impossible in VT-2 (probably hard to remember for new users), we can turn our bootstrap script into two parts:
- The first part will be running in `crosh`, replace the `.bashrc` with `install.sh`
- Prompt user to log in VT-2 and the installation should start automatically
- Prompt user to switch back to ChromeOS UI once installation done

satmandu:
> About the `sudo` issue: Here are my two cents, since copying/pasting the installation command is impossible in VT-2 (probably hard to remember for new users), we can turn our bootstrap script into two parts:
> 
> * The first part will be running in `crosh`, replace the `.bashrc` with `install.sh`
> * Prompt user to log in VT-2 and the installation should start automatically
> * Prompt user to switch back to ChromeOS UI once installation done

That's an interesting strategy...  Maybe it might be better to have some code we could replace `~/.bashrc` with that would offer to source `install.sh` if the shell is opening not in crosh?

satmandu:
(That could be a separate PR we should test...)

install.sh
```diff
@@ -113,19 +143,22 @@ if [ -d "${CREW_PREFIX}" ]; then
...
+        find "${CREW_PREFIX}" -mindepth 1 -delete
         break 2
       else
         exit 1
       fi
     done
   fi
 else
-  sudo mkdir "${CREW_PREFIX}"
+  echo_error "${CREW_PREFIX} does not exist. (It should be mounted!)"
```
uberhacker, line 156: Good point.  Maybe say `It should be created or mounted prior to install`?

satmandu: done.

uberhacker (approved):

PR #8851, opened 2023-10-30
Add exitmessages

satmandu:
- Allow one to use `ExitMessage.add 'message to be printed before crew exits'` in a package, such as in a postinstall.
- `fixup.rb` and `sommelier.rb` have both been adjusted to use `ExitMessage.new` in lieu of `puts`.
- One can still use a color override, e.g.
```
    ExitMessage.add <<~EOT1.lightblue

      The default environment is stored in #{CREW_PREFIX}/etc/env.d/sommelier.
      DO NOT EDIT THIS FILE SINCE UPDATES WILL OVERWRITE YOUR CHANGES.
      To override environment variables set above, edit ~/.sommelier.env instead.
      Information about those environment variables may be found on the
      Chromebrew wiki: https://github.com/chromebrew/chromebrew/wiki

      To start the sommelier daemon, run 'startsommelier'
      To stop the sommelier daemon, run 'stopsommelier'
      To restart the sommelier daemon, run 'restartsommelier'
    EOT1
```

image

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=exitmessages crew update
```


bin/crew
```diff
@@ -124,6 +124,25 @@ String.use_color = args['--color'] || !args['--no-color']
 @verbose           = @opt_verbose ? 'v' : ''
 @short_verbose     = @opt_verbose ? '-v' : ''
 
+class ExitMessage
+  attr_accessor :msg
+
+  @obj_all = []
+
+  def initialize(msg)
```
supechicken, line 132: This can be deleted as the `.new` function has been overwritten

satmandu: You're saying I don't need lines 129-132?

bin/crew
```diff
@@ -124,6 +124,25 @@ String.use_color = args['--color'] || !args['--no-color']
...
+    @msg = msg
+  end
+
+  class << self
+    attr_reader :obj_all
+
+    def new(msg)
+      obj_all << super
+      obj_all.last
```
supechicken, line 141: Do you mean `obj_all << msg`? (the message string instead of the `ExitMessage` class)

This line can be removed as the line above will return the message itself.

supechicken: (actually, it is better to use `.add` as the method name to prevent confusing)

satmandu: I'm modifying this:

https://stackoverflow.com/a/53877094/21935208

bin/crew
supechicken: The `def initialize` function can be deleted (from line 132 to 134)

satmandu:
@supechicken if there's a better way to do this, I'm happy to entirely change the approach...

bin/crew
```diff
@@ -124,6 +124,25 @@ String.use_color = args['--color'] || !args['--no-color']
 @verbose           = @opt_verbose ? 'v' : ''
 @short_verbose     = @opt_verbose ? '-v' : ''
 
+class ExitMessage
+  attr_accessor :msg
```
supechicken, line 128: Making the `msg` variable available outside the class should not be necessary (this line is not needed)

supechicken:
> @supechicken if there's a better way to do this, I'm happy to entirely change the approach...

I am thinking something like this:
```ruby
class ExitMessage
  @messages = []

  def self.add(msg, print_last: false)
    # add an option to allow important messages (like sommelier) to print at the bottom
    @messages << [msg. print_last]
  end

  def self.print # print all messages
    @messages.reject(&:print_last).map(&:first)
    @messages.select(&:print_last).map(&:first)
  end
end

**Comment by supechicken** [comment]
> ok, so we could then use `ExitMessage.add 'message'` and `ExitMessage.print` - is that right?

Yes

**Comment by satmandu** [comment]
@supechicken This does make doing color harder...

Is there a way to make the puts have a default color here?

**Comment by supechicken** [comment]
> @supechicken This does make doing color harder...
> 
> Is there a way to make the puts have a default color here?

We can just apply a default color here:
```ruby
@messages << [msg.lightcyan, print_last]
```

satmandu:
@supechicken Just dropped in your code. Seems to work well!

uberhacker (approved):

PR #8853, opened 2023-10-31
Misc updates

satmandu:
- Rebuild ncurses to not build twice. (confirmed working with `vim`)
- Add `disable-dependency-tracking` to CREW_OPTIONS.

Builds properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=misc_updates crew update
```


packages/mpg123.rb
```diff
@@ -1,41 +1,28 @@
...
+class Mpg123 < Autotools
   description 'Fast console MPEG Audio Player and decoder library.'
   homepage 'http://www.mpg123.org/'
-  version '1.29.3'
-  compatibility 'all'
+  version '1.32.3'
   license 'GPL-2 and LGPL-2.1'
-  source_url 'https://sourceforge.net/projects/mpg123/files/mpg123/1.29.3/mpg123-1.29.3.tar.bz2'
-  source_sha256 '963885d8cc77262f28b77187c7d189e32195e64244de2530b798ddf32183e847'
+  compatibility 'x86_64 aarch64 armv7l'
```
Zopolis4, line 8: I had issues building `mpg123` in your `i686` docker image, but there are `i686` [packages](https://centos.pkgs.org/7/centos-x86_64/mpg123-devel-1.25.6-1.el7.i686.rpm.html) for `mpg123`, so maybe its a docker issue? Could be something with the emulation or cross-compilation or something. 

Did you compile this on real hardware? If not, could you try it?

packages/speex.rb
```diff
@@ -1,43 +1,30 @@
...
+  configure_options '--disable-dependency-tracking \
       --disable-maintainer-mode \
-      --disable-examples"
-    system 'make'
-  end
-
-  def self.install
-    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
-  end
+      --disable-examples'
```
Zopolis4, line 29: The `--disable-examples` option is only relevant to `speexdsp`, it's not present in `speex`.

packages/mpg123.rb
Zopolis4: As a last resort, compilation on the docker image worked when adding `--with-cpu=i386`, but this looks like a symptom of a root issue with the docker image, as it appears to be presenting itself as `x86_64` in some way.

packages/speex.rb
```diff
@@ -1,43 +1,30 @@
...
-  end
+  depends_on 'glibc' # R
 
-  def self.build
-    system 'NOCONFIGURE=1 ./autogen.sh'
-    system "env #{CREW_ENV_OPTIONS} \
-      ./configure \
-      #{CREW_OPTIONS} \
-      --disable-dependency-tracking \
+  configure_options '--disable-dependency-tracking \
```
Zopolis4, line 27: Given that `--disable-dependency-tracking` [appears](https://www.gnu.org/software/automake/manual/html_node/Dependency-Tracking.html#index-_002d_002ddisable_002ddependency_002dtracking) to be relevant and useful for all autotools packages, should this be in `CREW_OPTIONS` in `lib/const.rb`?

packages/speexdsp.rb
```diff
@@ -1,45 +1,31 @@
...
-    system 'filefix'
-  end
-
-  def self.build
-    system 'NOCONFIGURE=1 ./autogen.sh'
-    system "env #{CREW_ENV_OPTIONS} \
-      ./configure \
-      #{CREW_OPTIONS} \
-      --disable-dependency-tracking \
+  configure_options '--disable-dependency-tracking \
```
Zopolis4, line 28: See the comment about `--disable-dependency-tracking` on `speex.rb`.

packages/alsa_lib.rb
```diff
@@ -3,25 +3,28 @@
...
+       i686: '9be05bc6dcc67cdb8d02ca168612f06baf429cd84e24533b99a9b0f9e0b9a907',
+     x86_64: '6faec883fafbf426ef0570e976a24e4755878ce77c997f346283258130124982'
   })
 
-  depends_on 'python3' # L
   depends_on 'glibc' # R
+  depends_on 'python3' # L
 
   def self.build
     @py_ver = `python -c "import sys; version = '.'.join(map(str, sys.version_info[:2])) ; print(version)"`.chomp
```
Zopolis4, line 30: It [looks](https://github.com/alsa-project/alsa-lib/blob/master/configure.ac#L462C11-L464) like `alsa-lib` will automatically do this for us if we have `python3-config` installed. 

This would allow us to clean up this code, and convert `alsa-lib` to buildsystems.

Zopolis4 (changes requested):
Couple relatively minor changes, few slightly larger ones.

packages/alsa_lib.rb
Zopolis4: And it looks like that automatic code already works. Double check this though, my testing was a little rough.

packages/speex.rb
satmandu: Good Idea.

packages/alsa_lib.rb
satmandu: The python stuff doesn't get in without the py_ver stuff...

uberhacker (approved):

packages/mpg123.rb
Zopolis4: The `i686` docker container reports an arch of `x86_64` from both the `arch` command and `uname -m`. Manually overriding the configure script to use a target of `i686-cros-linux-gnu`, which is what `crew` would use on an `i686` arch, works. 

Given that `--with-cpu=i386` disables some optimisations, I think it'd make sense to revisit this in future when we find out the issue with the docker image.

PR #8869, opened 2023-11-01
buildsystems/autotools: only run configure if Makefile does not exist

satmandu:
- When restarting a cached build, it is useful not to run configure again if Makefile already exists, for instance when working with adjusting the install section during build testing.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=cached_autotools crew update
```

### DIFF
--- a/lib/buildsystems/autotools.rb
+++ b/lib/buildsystems/autotools.rb
@@ -5,25 +5,27 @@ class Autotools < Package
   property :pre_configure_options, :configure_options
 
   def self.build
-    puts "Additional configure_options being used: #{@pre_configure_options.nil? || @pre_configure_options.empty? ? '<no pre_configure_options>' : @pre_configure_options} #{@configure_options.nil? || @configure_options.empty? ? '<no configure_options>' : @configure_options}".orange
-    # Run autoreconf if necessary
-    unless File.executable? './configure'
-      if File.executable? './autogen.sh'
-        system 'NOCONFIGURE=1 ./autogen.sh --no-configure || NOCONFIGURE=1 ./autogen.sh'
-      elsif File.executable? './bootstrap'
-        system 'NOCONFIGURE=1 ./bootstrap --no-configure || NOCONFIGURE=1 ./bootstrap'
-      else
-        system 'autoreconf -fiv'
+    unless File.file?('Makefile')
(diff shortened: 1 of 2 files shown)


lib/buildsystems/autotools.rb
```diff
@@ -5,25 +5,27 @@ class Autotools < Package
...
       end
+      abort 'configure script not found!'.lightred unless File.file?('configure')
+      FileUtils.chmod('+x', 'configure')
+      if `grep -q /usr/bin/file configure`
+        puts 'Using filefix.'.orange
+        system 'filefix'
+      end
+      @mold_linker_prefix_cmd = CREW_LINKER == 'mold' ? 'mold -run ' : ''
+      system "#{@pre_configure_options} #{@mold_linker_prefix_cmd}./configure #{CREW_OPTIONS} #{@configure_options}"
     end
```
uberhacker, line 28: What do you do in cases when there is no Makefile?  Run automake?

uberhacker: Also, woundn't you want to run configure if the Makefile settings were wrong?  ie. `prefix = /usr`?

satmandu: Isn't that where `def self.patch` comes in?

satmandu: Generally with autotools, you don't have a Makefile until configure runs....

uberhacker: Edits in `def self.patch` is not the elegant way to do it.  The whole point of `configure` is so you can control the settings without patching.

satmandu: For sure, but if the Makefile is prepopulated, that's non-standard, no? With autotools you are supposed to have a configure script to generate the Makefile.

Yes, there are cases where a Makefile is provided, but generally those are hacks..., and in those cases we can't use the buildtools/autotools mechanism anyways.

satmandu: now gated behind `CREW_CACHE_BUILD`...

uberhacker (approved):

lib/buildsystems/autotools.rb
uberhacker: I have seen several projects that provide a pre-generated Makefile but only some of the time are the settings correct.  I hope the configure script will be able to run the first time without any messing with the default environment settings.

PR #8871, opened 2023-11-02
Inetutils => 2.4

satmandu:
- Also fixup `iputils` filelists.
- Update `alive` with preflight section

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=inetutils crew update
```


packages/alive.rb
```diff
@@ -16,4 +16,11 @@ class Alive < Autotools
 
   depends_on 'guile' # R
   depends_on 'inetutils' # L
+
+  def self.preflight
+    puts "You need to be in the VT-2 shell for the install of alive to work.".lightblue
+    puts "Make sure you have used Ctrl-Alt-{F2/Right arrow/Refresh} to switch to VT-2.".lightblue
+    puts "Then login as 'chronos' and install alive with 'crew install alive'.".lightblue
+    puts "You can then switch back to your regular Chromeos windows with Ctrl-Alt-{F1/Left arrow}.".lightblue
```
uberhacker, line 24: Why is this necessary?

satmandu: The configure script calls ping from inetutils, and that only works with sudo now, so configure needs to be run in VT2 where sudo works...

uberhacker (approved):

PR #8889, opened 2023-11-04
lib/package: Remove `reload_constants` and its usage

supechicken:
Compared to `reload_constants`, the `no_*` package flags did the job nicely, so let's remove it for simplicity :)


uberhacker (approved):
I'm fine with this change.

PR #8890, opened 2023-11-04
lib/const: Simplify + cleanup

supechicken:
This PR replaced some `if ENV[].to_s.empty?` blocks with `ENV.fetch()` and simplified some other parts (**the behavior of the new code should be the same as before**)

Also, boolean constants will switch to `true` only if `ENV['const_name'] == '1'` for simplicity now.

### Other changes
- Use `File.join` on paths to prevent duplicated slashes
- Use `.eql?` instead of `==` on one-line if statements
- Use Ruby's new one-line method syntax on some functions in `lib/packages.rb`

```
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=syntax_cleanup crew update
```


supechicken:
About the `Lint/RedundantSafeNavigation` rubocop error: The `ENV[]` object doesn't like any other hash objects in ruby, the returned value can be `nil` if the key does not exist...

satmandu:
> About the `Lint/RedundantSafeNavigation` rubocop error: The `ENV[]` object doesn't like any other hash objects in ruby, the returned value can be `nil` if the key does not exist...

You can change the `.rubocop.yaml`, and put the new hash in the `ruby_rubocop.rb` if you want to update the rubocop config to bypass this check.

.rubocop.yml
```diff
@@ -105,6 +105,9 @@ Lint/ShadowingOuterLocalVariable:
 Lint/SuppressedException:
   Enabled: false
 
+Lint/RedundantSafeNavigation:
+  Enabled: false
```
satmandu, line 109: Also make sure to update the hash in the ruby_rubocop.rb file, otherwise that package will no longer install.

supechicken: Done

satmandu:
This breaks `crew install vim`

lib/const.rb
```diff
@@ -50,31 +50,31 @@
 # Glibc version can be found from the output of libc.so.6
 LIBC_VERSION = `/#{ARCH_LIB}/libc.so.6`[/Gentoo ([^-]+)/, 1]
 
-if ENV['CREW_PREFIX'].to_s.empty? || (ENV['CREW_PREFIX'] == '/usr/local')
-  CREW_BUILD_FROM_SOURCE = ENV.fetch('CREW_BUILD_FROM_SOURCE', nil)
-  CREW_PREFIX = '/usr/local'
+CREW_PREFIX = ENV.fetch('CREW_PREFIX', '/usr/local')
+
+if CREW_PREFIX == '/usr/local'
+  CREW_BUILD_FROM_SOURCE = !!ENV['CREW_BUILD_FROM_SOURCE']&.eql?('1')
```
uberhacker, line 56: What is wrong with `ENV.fetch` or simply `CREW_BUILD_FROM_SOURCE = ENV['CREW_BUILD_FROM_SOURCE']`?

satmandu:
> This breaks `crew install vim`

It would make sense to have a GitHub action that runs some crew commands to make sure they don't break for every PR. Installing and installing a package with dependencies for instance should probably be a part of that.

lib/const.rb
```diff
@@ -50,31 +50,31 @@
...
-  CREW_BUILD_FROM_SOURCE = 1
-  CREW_PREFIX = ENV.fetch('CREW_PREFIX', nil)
+  CREW_BUILD_FROM_SOURCE = true
   HOME = CREW_PREFIX + Dir.home
 end
 
 CREW_ARCHIVE_DEST = ENV.fetch('CREW_ARCHIVE_DEST', Dir.pwd)
 
-CREW_IN_CONTAINER = File.exist?('/.dockerenv') || !ENV['CREW_IN_CONTAINER'].to_s.empty?
+CREW_IN_CONTAINER = File.exist?('/.dockerenv') || !!ENV['CREW_IN_CONTAINER']&.eql?('1')
```
uberhacker, line 65: `|| ENV['CREW_IN_CONTAINER']`?

lib/const.rb
```diff
@@ -50,31 +50,31 @@
...
 CREW_CPU_VENDOR = CPUINFO['vendor_id'] || 'unknown'
 # The cpuinfo vendor_id may not exist on non-x86 platforms, or when a
 # container is virtualized on non-x86 platforms. Default to
 # CREW_IS_INTEL for x86 architectures. Note that a QEMU_EMULATED check
 # is not relevant here since qemu can be configured to pass through a
 # cpuinfo vendor_id.
-CREW_IS_AMD = ARCH == 'x86_64' ? ( CREW_CPU_VENDOR != 'unknown' and CPUINFO['vendor_id'].include?('AuthenticAMD') ) : false
-CREW_IS_INTEL = ARCH == 'x86_64' || ARCH == 'i686' ? ( CREW_CPU_VENDOR == 'unknown' or CPUINFO['vendor_id'].include?('GenuineIntel') ) : false
+CREW_IS_AMD = CREW_CPU_VENDOR.eql?('AuthenticAMD')
+CREW_IS_INTEL = %w[x86_64 i686].include?(ARCH) && %w[unknown GenuineIntel].include?(CREW_CPU_VENDOR)
```
uberhacker, line 74: I like this simplification.

lib/const.rb
```diff
@@ -50,31 +50,31 @@
...
 
 # Use sane minimal defaults if in container and no override specified.
-if CREW_IN_CONTAINER && ENV['CREW_KERNEL_VERSION'].to_s.empty?
+if CREW_IN_CONTAINER && ENV['CREW_KERNEL_VERSION'].nil?
```
uberhacker, line 77: Does `!ENV['CREW_KERNEL_VERSION']` work?

lib/const.rb
```diff
@@ -87,74 +87,59 @@
...
-CREW_DEST_MAN_PREFIX = CREW_DEST_DIR + CREW_MAN_PREFIX
+CREW_LIB_PREFIX      = File.join(CREW_PREFIX, ARCH_LIB)
+CREW_MAN_PREFIX      = File.join(CREW_PREFIX, 'share/man/')
+CREW_LIB_PATH        = File.join(CREW_PREFIX, 'lib/crew/')
+CREW_PACKAGES_PATH   = File.join(CREW_LIB_PATH, 'packages/')
+CREW_CONFIG_PATH     = File.join(CREW_PREFIX, 'etc/crew/')
+CREW_META_PATH       = File.join(CREW_CONFIG_PATH, 'meta/')
+CREW_BREW_DIR        = File.join(CREW_PREFIX, 'tmp/crew/')
+CREW_DEST_DIR        = File.join(CREW_BREW_DIR, 'dest/')
+CREW_DLL_PREFIX      = File.join(CREW_LIB_PREFIX, 'wine/')
```
uberhacker, line 98: Better name? `CREW_WINE_PREFIX`?

supechicken:
> This breaks `crew install vim`

This is fixed by adding a `!` to `def is_source`

lib/const.rb
supechicken: Yes

lib/const.rb
supechicken: All boolean constants will switch to true only if `ENV['const_name'] == '1'` for simplicity now.

lib/const.rb
```diff
@@ -87,74 +87,59 @@
...
 
 # Local constants for contributors.
-repo_root = `git rev-parse --show-toplevel 2> /dev/null`.chomp.to_s
-if repo_root.empty?
-  Dir.chdir '../..' do
-    repo_root = `git rev-parse --show-toplevel 2> /dev/null`.chomp.to_s
-  end
-end
+repo_root = `git rev-parse --show-toplevel 2> /dev/null`.chomp
+repo_root = `cd ../..; git rev-parse --show-toplevel 2> /dev/null`.chomp if repo_root.empty?
```
uberhacker, line 106: `cd ../..` is a bad assumption.  What if you are in `/` or `/usr` to start?  A better approach would be to `cd #{CREW_LIB_PATH} && ...`.  This should always provide `repo_root` unless someone deliberately deletes the `.git` directory.

uberhacker: Also `&&` is better than `;` because it will abort if the directory doesn't exist.

lib/const.rb
supechicken: Some people might be uncomfortable with actually unsetting a variable with `unset` (like me :) ), they would probably run something like `VAR=''` to "unset" them (the env variable still exists but being empty).

Also, this prevents unexpected results and misunderstanding, like `export CREW_BUILD_FROM_SOURCE=false` being evaluated as `true`

lib/const.rb
supechicken: This part is modified from the current code (I am not very familiar with the usage of this variable actually😅)

supechicken: It will run inside `CREW_LIB_PATH` now

lib/const.rb

bin/crew
```diff
@@ -1924,7 +1924,7 @@ def reinstall_command(args)
     @pkgName = name
     search @pkgName
     print_current_package
-    @pkg.build_from_source = true if @opt_source || @opt_recursive || (CREW_BUILD_FROM_SOURCE == '1')
+    @pkg.build_from_source = true if @opt_source || @opt_recursive || CREW_BUILD_FROM_SOURCE
```
uberhacker, line 1927: Does `export CREW_BUILD_FROM_SOURCE=false` still build from source?

uberhacker: I think we could omit the `true if` too, correct?

lib/const.rb
```diff
@@ -332,7 +306,7 @@
   -DCMAKE_BUILD_TYPE=Release
 OPT
 
-CREW_LIB_SUFFIX = ARCH == 'x86_64' ? '64' : ''
+CREW_LIB_SUFFIX = ARCH.eql?('x86_64') ? '64' : ''
 CREW_CMAKE_LIBSUFFIX_OPTIONS = "#{CREW_CMAKE_OPTIONS} -DLIB_SUFFIX=#{CREW_LIB_SUFFIX}"
 
 PY3_SETUP_BUILD_OPTIONS = "--executable=#{CREW_PREFIX}/bin/python3"
```
uberhacker, line 312: Bump crew version?

bin/crew
supechicken: IIRC this parts means overriding the original value with `true` if `@opt_source || @opt_recursive || CREW_BUILD_FROM_SOURCE` (the `true if` is necessary)

supechicken: > Does `export CREW_BUILD_FROM_SOURCE=false` still build from source?

No, only the exact `1` value will trigger the flag

uberhacker: This contradicts what you said earlier about the reason it is necessary to set the environment variable to 1.

uberhacker: `@opt_source || @opt_recursive || CREW_BUILD_FROM_SOURCE` will return true anyways, right?

supechicken: > This contradicts what you said earlier about the reason it is necessary to set the environment variable to 1.

Umm, you didn't get my meaning... For example:
```
export CREW_BUILD_FROM_SOURCE=false  # will be `false` in const.rb
export CREW_BUILD_FROM_SOURCE=true   # will be `false` in const.rb
export CREW_BUILD_FROM_SOURCE=""     # will be `false` in const.rb
export CREW_BUILD_FROM_SOURCE=1      # only this will make the const `true`
```

> @opt_source || @opt_recursive || CREW_BUILD_FROM_SOURCE will return true anyways, right?

`CREW_BUILD_FROM_SOURCE` isn't always true...

lib/const.rb
supechicken: Done

bin/crew
uberhacker: Well, if it isn't true and it doesn't build from source, that is the expected behavior, right?

supechicken: Yes, unless the `crew install -s` option (`@opt_source`) is given

uberhacker (approved):

uberhacker (approved):

PR #8893, opened 2023-11-05
Vscodium 1.83.1.23285 => 1.84.0.23306

uberhacker:


satmandu (approved):
Do we need to add any file lists with the new package name?

uberhacker:
> Do we need to add any file lists with the new package name?

Yes, updates pushed.

satmandu (approved):

PR #8896, opened 2023-11-05
Glib_networking 2.76.0 => 2.78.0

uberhacker:
Fixes /usr/local/lib/libproxy.so.1: version `LIBPROXY_0.4.16' not found (required by /usr/local/lib/gio/modules/libgiolibproxy.so)
Failed to load module: /usr/local/lib/gio/modules/libgiolibproxy.so

Unable to build for i686 since libproxy is a dependency and not compatible.

### DIFF
--- a/packages/glib_networking.rb
+++ b/packages/glib_networking.rb
@@ -1,25 +1,23 @@
-require 'package'
+require 'buildsystems/meson'
 
-class Glib_networking < Package
+class Glib_networking < Meson
   description 'Network extensions for GLib'
   homepage 'https://github.com/GNOME/glib-networking'
-  version '2.76.0'
+  version '2.78.0'
   license 'LGPL-2.1+'
-  compatibility 'all'
+  compatibility 'x86_64 aarch64 armv7l'
   source_url 'https://gitlab.gnome.org/GNOME/glib-networking.git'
   git_hashtag version
 
   binary_url({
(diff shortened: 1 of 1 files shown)


satmandu (approved):

Zopolis4:
`libproxy` seems to have i686 binaries (see https://packages.debian.org/sid/libproxy-dev), is there a particular reason it doesn't work on chromeos?

satmandu:
> `libproxy` seems to have i686 binaries (see https://packages.debian.org/sid/libproxy-dev), is there a particular reason it doesn't work on chromeos?

We are spending zero time on fixing dependencies for i686 comnect to X11 since we can't get any X or Wayland apps working on the last milestone with i686 support, M58...

Zopolis4:
`neon.rb` uses libproxy, and is not a gui application.

Although, `neon.rb` has `compatibility 'all'`, which is probably not ideal. Perhaps we should have some sort of script to check the compatibility of packages with their dependencies, as this is incorrect behavior regardless of the true compatibility of `libproxy`.

satmandu:
> `neon.rb` uses libproxy, and is not a gui application.
> 
> Although, `neon.rb` has `compatibility 'all'`, which is probably not ideal. Perhaps we should have some sort of script to check the compatibility of packages with their dependencies, as this is incorrect behavior regardless of the true compatibility of `libproxy`.

Go to town. 😅

Issue #8894, opened 2023-11-05
🐞 Bug crew install from source is broken


uberhacker:
```
$ crew install girara -s
girara: Girara is a library that implements a user interface that focuses on simplicity and minimalism
https://pwmt.org/projects/girara/
Version: 0.4.0
License: ZLIB
Performing pre-flight checks...
Package girara already installed, skipping...
Girara installed!
```

uberhacker:
Not a but if using `crew reinstall girara -s`.

satmandu:
Not being able to install from source is still a bug!

uberhacker:
I think the logic to determine if the package is already installed precedes the decision to install from binary or source.

supechicken:
Huh? It is expected and shouldn't be considered a bug. (same behavior after #8895...)

> Not being able to install from source is still a bug!

@uberhacker is trying to install an already installed package, `crew` will block it of course...

PR #8901, opened 2023-11-06
Reimplementing #8890 (`lib/const: Simplify + cleanup`)

supechicken:
#8894 is not a bug, and `crew install -s` on non-installed packages works just fine:

> Huh? It is expected and shouldn't be considered a bug. (same behavior after #8895...)
>
> > Not being able to install from source is still a bug!
>
> @uberhacker is trying to install an already installed package, `crew` will block it of course...

---

Also, can we **stop** "fixing" bugs by reverting? Sometimes debugging is not that hard...


satmandu:
Debugging a regression when you didn't write the code can be difficult.

You don't want to break existing code with changes!

Can we wait on this large change until @Zopolis4 has his GitHub action unit test ready to go?

It shouldn't be that complicated, and this would be an excellent test of that...

supechicken:
I am okay with it...

uberhacker:
These look like the exact changes from the previous PR that broke everything. :)

Zopolis4:
> Can we wait on this large change until @Zopolis4 has his GitHub action unit test ready to go?
> 
> It shouldn't be that complicated, and this would be an excellent test of that...

(Not sure if github shows pronouns properly, but its *her* Github action unit test, btw)

Anyways, I can spin up a pretty crude one by the end of today, although I have plans for proper unit tests later.

Those plans are here on my list of todo's, though, so...
- Derive binary_url in package.rb (7082)
- Improve logic for detecting package updates (8587)
- Remove obsolete tools (8547)
- Add automatic builds (8414)
- better unit tests
- better docker

satmandu:
Sorry for wrong pronouns!

supechicken:
> These look like the exact changes from the previous PR that broke everything. :)

#8894 is not a bug, please read the description... (this PR is actually identical to #8890)

satmandu:
I really want to see if this PR passes the tests just added in https://github.com/chromebrew/chromebrew/pull/8907

satmandu (approved):
Unit Tests Passed!

PR #8906, opened 2023-11-07
Fix installer on all milestones and first-time installs.

Zopolis4:
After discussions on Slack with the team and discussions with users on Discord, this should fix the issues.

Testing would be apprciated, in particular tests on milestones before and after the `crosh` change.

```
exec bash --init-file <(curl -Ls https://raw.githubusercontent.com/Zopolis4/chromebrew/aisda/install.sh)
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,41 +3,6 @@
 # Exit on fail.
 set -e
 
-# Default chromebrew repo values.
-
-: "${OWNER:=chromebrew}"
-: "${REPO:=chromebrew}"
-: "${BRANCH:=master}"
-
-# Chromebrew directories
-: "${CREW_PREFIX:=/usr/local}"
-CREW_LIB_PATH="${CREW_PREFIX}/lib/crew"
-CREW_CONFIG_PATH="${CREW_PREFIX}/etc/crew"
-CREW_BREW_DIR="${CREW_PREFIX}/tmp/crew"
-CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
-: "${CURL:=/usr/bin/curl}"
-: "${CREW_CACHE_DIR:=$CREW_PREFIX/tmp/packages}"
-
-# Architecture
-
-# For container usage, where we want to specify i686 arch
-# on a x86_64 host by setting ARCH=i686.
-: "${ARCH:=$(uname -m)}"
-# For container usage, when we are emulating armv7l via linux32
-# uname -m reports armv8l.
-ARCH="${ARCH/armv8l/armv7l}"
-
-# Determine if there is a difference between kernel and user space
-if [[ "${ARCH}" == "aarch64" && ! -d '/lib64' ]]; then
-  USER_SPACE_ARCH='armv7l'
-else
-  USER_SPACE_ARCH="${ARCH}"
-fi
-
-if [[ "$ARCH" == "x86_64" ]]; then
-  LIB_SUFFIX='64'
-fi
-
 RESET='\e[0m'
 
 # Simplify colors and print errors to stderr (2).
@@ -47,44 +12,12 @@ echo_success() { echo -e "\e[1;32m${*}${RESET}" >&1; } # Use Green for success m
 echo_intra() { echo -e "\e[1;34m${*}${RESET}" >&1; } # Use Blue for intrafunction messages.
 echo_out() { echo -e "\e[0;37m${*}${RESET}" >&1; } # Use Gray for program output.
 
-# Skip all checks if running on a docker container.
-if [[ -f "/.dockerenv" ]]; then
-  CREW_FORCE_INSTALL=1
-  echo_info "In container, so skipping hardware checks"
-fi
-
-#
-# Attempt install without sudo.
-#elif tty | grep -q /dev/pts1; then
-#  echo_success "You are in VT-2."
-#  if [ "${EUID}" == "0" ]; then
-#    echo_info "Please login as 'chronos' and restart the install."
-#    echo_info "Note that you can set the chronos password if you login as root and run 'chromeos-setdevpasswd'."
-#    exit 1
-#  fi
-#else
-#  # Error out if not in VT-2
-#  echo_error "You need to be in the VT-2 shell for the install to work."
-#  echo_info "Please use Ctrl-Alt-{F2/Right arrow/Refresh}) to switch to VT-2."
-#  echo_info "Then login as 'chronos' and restart the install."
-#  echo_info "Note that you can set the chronos password if you login as root and run 'chromeos-setdevpasswd'."
-#  exit 1
-#fi
-
 # Check if the script is being run as root.
 if [ "${EUID}" == "0" ]; then
   echo_error "Chromebrew should not be installed or run as root."
-  exit 1;
+  echo_info "Please login as 'chronos' and restart the install."
+  exit 1
 fi
-
-# Uncomment this section if you have a stoneyridge machine and your
-# install fails.
-# if grep -s "AuthenticAMD" /proc/cpuinfo ; then
-#  echo_info "Need to disable randomize_va_space on AMD machines."
-#  # Otherwise one may get segfaults during install on stoneyridge
-#  # devices. See https://github.com/chromebrew/chromebrew/issues/8823
-#  echo 0 | sudo tee /proc/sys/kernel/randomize_va_space
-#fi
 
 # Reject crostini.
 if [[ -d /opt/google/cros-containers && "${CREW_FORCE_INSTALL}" != '1' ]]; then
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -200,7 +186,7 @@ for package in $BOOTSTRAP_PACKAGES; do
   nln=$(nl -b a -n ln "${package}.rb" | sed -n '/class/p' | head -1 | cut -c 1)
   vln=$(nl -b a -n ln "${package}.rb" | sed -n '/  version/p' | head -1 | cut -c 1)
   name=$(sed -n "${nln}s/class\(.*\)<.*/\L\1/;${nln}s/^[ \t]*//;${nln}s/[ \t]*$//p" "${package}.rb")
-  version=$(sed -n "${vln}p" "${package}.rb" | awk '{print $2}' | tr -d "'")
+  version=$(sed -n "${vln}s/version \(.*\)/\1/;${vln}s/^[ \t]*//;${vln}s/[ \t]*$//;${vln}s/ *#.*//p" "${package}.rb" | tr -d "'")
```
uberhacker, line 189: Why are we going back to the more complex way that didn't work?

install.sh
```diff
@@ -237,8 +223,7 @@ function download_check () {
     # Use cached file if available and caching is enabled.
     if [ -n "$CREW_CACHE_ENABLED" ] && [[ -f "$CREW_CACHE_DIR/${3}" ]] ; then
       mkdir -p "$CREW_CACHE_DIR"
-      # sudo chown -R "$(id -u)":"$(id -g)" "$CREW_CACHE_DIR" || true
-      chown -R "$(id -u)":"$(id -g)" "$CREW_CACHE_DIR" || true
+      sudo chown -R "$(id -u)":"$(id -g)" "$CREW_CACHE_DIR" || true
```
uberhacker, line 226: Why not just use `chronos` verbatim since we are requiring a login with this user anyway?

install.sh
```diff
@@ -104,17 +69,44 @@ else
...
+  if tty | grep -q '/dev/pts/1'; then
+    echo_info "Please login as 'chronos' and restart the install."
+    exit 1;
+  fi
+  exit 1;
+fi
+
+# Warn users of the AMD segfault issue and allow them to work around it
+if grep "AuthenticAMD" /proc/cpuinfo; then
+  echo_info "If the install fails, try running 'CREW_AMD_SEGFAULT=1 exec bash --init-file <(curl -Ls git.io/vddgY)'"
```
uberhacker, line 92: I think a better (less techie) name could be `CREW_AMD_INSTALL`.  Mentioning `SEGFAULT` sounds a bit too hackish.

install.sh
```diff
@@ -181,14 +170,11 @@ BOOTSTRAP_PACKAGES="zstd crew_mvdir ruby git ca_certificates openssl"
...
+
 libc_version=$(/lib"$LIB_SUFFIX"/libc.so.6 | head -n 1 | sed -E 's/.*(stable release version.*) (.*)./\2/')
 case ${libc_version} in
   # Append the correct packages for M113 onwards systems.
-  "2.35")
-    BOOTSTRAP_PACKAGES+=' glibc_lib235 zlibpkg gmp'
-    # Recent Arm systems have a cut down system.
-    [[ "${USER_SPACE_ARCH}" == "armv7l" ]] && BOOTSTRAP_PACKAGES+=' bzip2 ncurses readline pcre2 gcc_lib'
-  ;;
+  "2.35") BOOTSTRAP_PACKAGES="${BOOTSTRAP_PACKAGES} glibc_lib235 zlibpkg gmp";;
```
uberhacker, line 177: What about older armv7l systems that need the extra packages?

uberhacker (changes requested):
I think we also need to update README.md with new installation instructions for the developer console too?

install.sh
Zopolis4: Possibly because the system may have a custom configuration, or other users?

To be honest I'm not entirely sure, but it doesn't look like we've used chronos verbatim for 7 years (if my git grep skills are to be trusted, which is not a given).

At any rate, its a total difference of 6 characters, and keeping it as is may help some obscure setups (custom user on ChromiumOS, maybe).

supechicken: I prefer not to hardcoding it also.

Zopolis4:
The script will tell people what to do if it's relevant-- no need to overcomplicate things for people on older milestones.

install.sh
```diff
@@ -104,17 +69,37 @@ else
   echo_info "Unable to detect system information, installation will continue."
 fi
 
+# Check if the user is on ChromeOS v117+ and not in the VT-2 console, sudo will not work
+if sudo | grep -q 'no new privileges'; then
```
supechicken, line 73: Calling `sudo` with no arguments will print the help message, this might works:
```shell
if [[ "$(sudo /bin/true 2>&1)" == *'no new privileges'* ]]
```

Zopolis4: When the relevant flag is set, it will print the warning instead (at least in my testing in docker containers)

satmandu:
The current installer works just fine in `crosh` as long as:
1. The machine is not AMD (those issues are getting fixed.)
2. Someone has run either `chown -R chronos /usr/local` as `root` on VT-2 or when logged in as `chronos` on VT-2 run  `sudo chown -R chronos`.

So... would it not make sense just to leave a message to run  `chown -R chronos /usr/local` as `root` on VT-2 before running the install, as that's the only invocation of sudo really needed?

Zopolis4:
> So... would it not make sense just to leave a message to run chown -R chronos /usr/local as root on VT-2 before running the install, as that's the only invocation of sudo really needed?

I guess? Personally, I think it is better if the user only has to run a single command.

This way, for users pre-M117, they can simply run the installer in `crosh` and all will work normally. Users after M117, if necessary, will be notified to switch to VT-2, and run the installer again.

install.sh
```diff
@@ -104,17 +68,37 @@ else
...
```
supechicken, line 72: This line will just print the error to terminal and will always `false`, you need to redirect stderr with `2>&1`

install.sh
```diff
@@ -322,7 +300,7 @@ done
 
 # Work around https://github.com/chromebrew/chromebrew/issues/3305.
 # shellcheck disable=SC2024
-# sudo ldconfig &> /tmp/crew_ldconfig || true
+sudo ldconfig &> /tmp/crew_ldconfig || true
```
supechicken, line 303: Is this workaround still necessary? We can remove it if it is no longer needed.

Zopolis4: I'm not sure...

uberhacker:
@Zopolis4: I think what you have here is great.  Can we update the README.md install section to mention the need to login to VT-2 instead of typing 'shell' from crosh?  I'd also like to move all the conditions that would cause an exit 1 to the top.  No need to set a bunch of values prior to exiting.

install.sh
```diff
@@ -48,43 +48,7 @@ echo_intra() { echo -e "\e[1;34m${*}${RESET}" >&1; } # Use Blue for intrafunctio
...
-#  exit 1
-#fi
-
-# Check if the script is being run as root.
-if [ "${EUID}" == "0" ]; then
-  echo_error "Chromebrew should not be installed or run as root."
-  exit 1;
-fi
-
-# Uncomment this section if you have a stoneyridge machine and your
```
satmandu, line 80: Please also leave this stoney ridge stuff in in some capacity so people still get the error message.

install.sh
```diff
@@ -143,22 +128,19 @@ if [ -d "${CREW_PREFIX}" ]; then
...
         break 2
       else
         exit 1
       fi
     done
   fi
 else
-  echo_error "${CREW_PREFIX} does not exist. (It should be created or mounted prior to install!)"
-  exit 1
+  sudo mkdir "${CREW_PREFIX}"
```
uberhacker, line 139: Might want to add `-p` in case `CREW_PREFIX` is several subdirectories.

install.sh
uberhacker: How about a compromise?  
```
: "${CREW_USER:=chronos}"
sudo chown -R "$CREW_USER":"$CREW_USER" "$CREW_CACHE_DIR" || true
```

Zopolis4:
I've got to rework a couple things, and I don't have the time to do so right now.

Zopolis4:
> Can we update the README.md install section to mention the need to login to VT-2 instead of typing 'shell' from crosh? 
Not everyone will need to use `VT-2`, so I'd prefer to keep things simpler for those who don't need to.

> I'd also like to move all the conditions that would cause an exit 1 to the top. No need to set a bunch of values prior to exiting.

We need to setup the `echo_*` functions to give people nice error messages before exiting.

install.sh
Zopolis4: This seems like it would go against the user's expectations, as if they are not the user `chronos` they will not be able to use crew without setting this variable.

uberhacker: The VT-2 session requires the user to login as chronos.  Unless they create a new user, there is no reason not to require chronos to login.

install.sh
```diff
@@ -143,48 +139,28 @@ if [ -d "${CREW_PREFIX}" ]; then
...
-  echo_info "Please use Ctrl-Alt-{F2/Right arrow/Refresh}) to switch to VT-2."
-  echo_info "Then login as 'root' and run: 'chown -R chronos ${CREW_PREFIX}'"
-  echo_info "or login as 'chronos' and run: 'sudo chown -R chronos ${CREW_PREFIX}'"
-  echo_info "Then switch back to crosh in the main ChromeOS screen with Ctrl-Alt-{F1/Left arrow}) and restart the install."
-  exit 1
-}
-
-
+# Do not redundantly use sudo if the user already owns the directory.
+if [ "$(stat -c '%u' "${CREW_PREFIX}")" == "$(id -u)" ]; then
```
uberhacker, line 154: What if the directory is not empty and subdirectory ownership needs to change?

install.sh
```diff
@@ -143,48 +139,28 @@ if [ -d "${CREW_PREFIX}" ]; then
...
+  # This will allow things to work without sudo.
+  sudo chown "$(id -u)":"$(id -g)" "${CREW_PREFIX}"
```
uberhacker, line 156: Use `-R` to recursively change all subdirectories?

install.sh
```diff
@@ -143,48 +139,28 @@ if [ -d "${CREW_PREFIX}" ]; then
...
+fi
 
-# This will create the directories
+# This will create the directories.
 crew_folders="bin cache doc docbook include lib/crew/packages lib$LIB_SUFFIX libexec man sbin share var etc/crew/meta etc/env.d tmp/crew/dest"
 # shellcheck disable=SC2086
 # Quoting crew_folders leads to breakage
-(cd "${CREW_PREFIX}"; mkdir -p ${crew_folders} || perm_error )
+(cd "${CREW_PREFIX}"; mkdir -p ${crew_folders})
```
uberhacker, line 163: `&&` would be better than `;` in case the cd command fails and we create crew folders in some random place.

install.sh
```diff
@@ -143,48 +139,28 @@ if [ -d "${CREW_PREFIX}" ]; then
...
         break 2
       else
         exit 1
       fi
     done
   fi
 else
-  echo_error "${CREW_PREFIX} does not exist. (It should be created or mounted prior to install!)"
-  exit 1
+  mkdir -p "${CREW_PREFIX}" || sudo mkdir -p "${CREW_PREFIX}"
```
uberhacker, line 150: It's fine to use `sudo mkdir -p "${CREW_PREFIX}"`.  We change the perms later anyway.

install.sh
```diff
@@ -133,7 +129,7 @@ function curl () {
 }
 
 if [[ "$USER_SPACE_ARCH" == 'armv7l' ]] && [[ "$ARCH" == 'aarch64' ]]; then
-    echo_error "Your device is not supported by Chromebrew yet, installing as armv7l."
+  echo_error "Your device is not supported by Chromebrew yet, installing as armv7l."
```
uberhacker, line 132: echo_info instead of echo_error?

uberhacker:
> > Can we update the README.md install section to mention the need to login to VT-2 instead of typing 'shell' from crosh?
> > Not everyone will need to use `VT-2`, so I'd prefer to keep things simpler for those who don't need to.
> 
> > I'd also like to move all the conditions that would cause an exit 1 to the top. No need to set a bunch of values prior to exiting.
> 
> We need to setup the `echo_*` functions to give people nice error messages before exiting.

Yes, put these above the exit conditions.  We are already doing this.  Just need to move every condition that exits higher and the other variable assignments and continuing logic lower.

install.sh
Zopolis4: We'll have deleted all the contents of CREW_PREFIX by this point, so there's nothing else to recurse over.

install.sh
Zopolis4: And if they have created a new user, the install will be broken by default.

uberhacker: That's what I'm saying.  Just require chronos user.

Zopolis4: Why? What's wrong with supporting other users?

install.sh
uberhacker: The install will not continue if the user refuses to remove all existing files so that makes sense.

install.sh
uberhacker: You cannot login to VT-2 and use sudo unless you are chronos.  If somebody creates a new user, they will need to add to sudo users or the install will fail.

Zopolis4: True, but if the new user does have sudo perms then there's no reason for the install to arbitrarily fail, and if the new user does not have sudo perms then the install will fail before that. 

And besides, we still have users who are pre M117 and do not need the VT-2 workaround.

uberhacker: I think the VT-2 workaround should be the standard installation instructions for all users.  It works on older releases too and most of the users will upgrade to the latest ChromeOS milestone anyway (as long as their hardware is not EOL).  It would be nice to autologin to a VT-2 session.  Maybe @supechicken can figure out how to do it. :smiley:

install.sh
Zopolis4: We only need ownership of the directory to delete all the files in the directory (tested).

Zopolis4:
We need a couple of the other variables we set as well-- notably `CREW_PREFIX`. I can still split them apart, but that feels like it would make the code significantly uglier.

install.sh
```diff
@@ -104,17 +67,49 @@ else
...
+# Check if the user is on ChromeOS v117+ and not in the VT-2 console, as sudo will not work.
+if [[ "$(stat -c '%u' "${CREW_PREFIX}")" == "$(id -u)" ]] && sudo 2>&1 | grep -q 'no new privileges'; then
+  echo_error "Please run the installer in the VT-2 shell."
+  exit 1
+fi
+
+# Check if the script is being run as root.
+if [ "${EUID}" == "0" ]; then
+  echo_error "Chromebrew should not be installed or run as root."
+  echo_info "Please login as 'chronos' and restart the install."
```
uberhacker, line 86: `Please login to VT-2 as 'chronos' and restart the install.`?  Might be good to provide instructions too.  `To start the VT-2 session, type Ctrl + Alt + ->`.

Zopolis4:
@uberhacker Following up from your comment
> If you can add instructions to README.md, that should make this ready to merge.

I've already got instructions in `install.sh`, I haven't added README instructions because they arent relevant to everyone and I don't want to potentially confuse users or have them do an unnecessarily complex setup.

Zopolis4:
@uberhacker, In future, I would prefer it if you did not force-push and overwrite my changes on my PR without at the very least asking me first.

Zopolis4:
For the sake of getting this PR merged and avoiding an argument, I'll keep these changes, but still.

uberhacker:
> @uberhacker, In future, I would prefer it if you did not force-push and overwrite my changes on my PR without at the very least asking me first.

This is an important update and I went ahead and made the changes I had suggested to get this moving.  Too many users are confused and we need to clear up the installation instructions.

uberhacker:
> It is easier to use `crosh` than `VT-2`. For users who don't need to use `VT-2`, the instructions in the README are the simplest and quickest option. For users who do need to use `VT-2`, the installer will provide instructions if necessary.
> 
> Replacing the `crosh` instructions with the `VT-2` instructions will mean that `crosh` users are using a more complex method for no reason.

You could make the argument that a VT-2 session install is easier; especially if there is no chronos password set.

Zopolis4:
> This is an important update and I went ahead and made the changes I had suggested to get this moving.

Yes, but you could also have accepted that these changes were not strictly necessary for this PR, or we could have reached an agreement.

Zopolis4:
> You could make the argument that a VT-2 session install is easier; especially if there is no chronos password set.

True, but VT-2 is fullscreen and lacks highlighting and clipboard support.

uberhacker:
> > This is an important update and I went ahead and made the changes I had suggested to get this moving.
> 
> Yes, but you could also have accepted that these changes were not strictly necessary for this PR, or we could have reached an agreement.

This PR hasn't been merged yet so we still have time to come to an agreement.  The reason I think the README.md instructions are absolutely necessary is because 90%+ of all users cannot run sudo from a crosh shell session now.  The VT-2 instructions should also work on older releases too.  I just wanted universal instructions to make it more clear and to avoid confusion.  As you may have already noticed, there were several issues submitted where users ran across errors due to the inability to run sudo commands.

uberhacker:
> > You could make the argument that a VT-2 session install is easier; especially if there is no chronos password set.
> 
> True, but VT-2 is fullscreen and lacks highlighting and clipboard support.

Then why don't we update README.md to explain that once installation is complete, we can run inside the crosh shell?  We just shouldn't be installing in crosh.

Zopolis4:
As it stands, the instructions in the README are fine. #8832 is still necessary for a proper experience in crosh, and until the status of #8919 is clarified we probably should hold off on recommending crosh.

Let's just get this PR merged.

uberhacker:
> As it stands, the instructions in the README are fine. #8832 is still necessary for a proper experience in crosh, and until the status of #8919 is clarified we probably should hold off on recommending crosh.
> 
> Let's just get this PR merged.

What are your thoughts @satmandu?

satmandu:
Merge it.

satmandu (approved):

Zopolis4:
In all that discussion it appears none of us tested the actual installer, so gimme a sec to fix some issues.

satmandu (approved):

PR #8913, opened 2023-11-08
Fix Github Action Handoff linter tests

satmandu:
- Fix handoff.yml
- Allow `Unit-Test.yml` and `Repology.yml` to pass yaml linter.
- Make rubocop changes to `lib/const.rb`

### DIFF
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches: [master]
 jobs:
-  generate:
+  container_tests:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -9,10 +9,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Dump github context
-        run:   echo "$GITHUB_CONTEXT"
+        run: echo "$GITHUB_CONTEXT"
         shell: bash
         env:
-         GITHUB_CONTEXT: ${{ toJson(github) }}
+          GITHUB_CONTEXT: ${{ toJson(github) }}
       - name: Unit Tests
         run: |
             sudo docker run -t satmandu/crewbuild:amd64 sudo -i -u chronos /bin/bash -c "
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -1,5 +1,5 @@
+---
 name: Run Unit Tests on PR
-
 on:
   pull_request:
     branches: [master]
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -1,8 +1,7 @@
 ---
 name: Run Unit Tests on PR
 on:
-  pull_request:
-    branches: [master]
+  workflow_call
 jobs:
   container_tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -1,7 +1,6 @@
 ---
 name: Run Unit Tests on PR
-on:
-  workflow_call
+on: workflow_call
 jobs:
   container_tests:
     runs-on: ubuntu-latest
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,7 +1,7 @@
 # lib/const.rb
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.39.0'
+CREW_VERSION = '1.39.1'
 
 # kernel architecture
 KERN_ARCH = `uname -m`.chomp
@@ -55,7 +55,7 @@ LIBC_VERSION = `/#{ARCH_LIB}/libc.so.6`[/Gentoo ([^-]+)/, 1]
 CREW_PREFIX = ENV.fetch('CREW_PREFIX', '/usr/local')
 
 if CREW_PREFIX == '/usr/local'
-  CREW_BUILD_FROM_SOURCE = !!ENV['CREW_BUILD_FROM_SOURCE']&.eql?('1')
+  CREW_BUILD_FROM_SOURCE = !ENV['CREW_BUILD_FROM_SOURCE']&.eql?('1').nil?
   HOME = Dir.home
 else
   CREW_BUILD_FROM_SOURCE = true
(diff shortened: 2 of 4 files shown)


Zopolis4:
This should work, although it kinda bypasses Handoff.yml and means we'll be running the unit tests more than we need to (e.g. when markdown files are changed)

lib/const.rb
```diff
@@ -55,7 +55,7 @@
 CREW_PREFIX = ENV.fetch('CREW_PREFIX', '/usr/local')
 
 if CREW_PREFIX == '/usr/local'
-  CREW_BUILD_FROM_SOURCE = !!ENV['CREW_BUILD_FROM_SOURCE']&.eql?('1')
+  CREW_BUILD_FROM_SOURCE = !ENV['CREW_BUILD_FROM_SOURCE']&.eql?('1').nil?
```
uberhacker, line 58: No easier way to do this?  Seems like overly complicated logic.

satmandu: I don't know? These were the automated rubocop replacements using our current rubocop linter config, so in theory this should have been triggered by the changes in any case...

satmandu:
> This should work, although it kinda bypasses Handoff.yml and means we'll be running the unit tests more than we need to (e.g. when markdown files are changed)

This is true, but 90% of our chances from PR to PR are ruby, and this also happens in parallel with the other linter stuff, so it is pretty quick.

Zopolis4:
True, I'm just thinking about our total CI minutes usage. I'm still okay with this as is, let's just keep an eye on our CI usage to make sure.

lib/const.rb
uberhacker: It appears we are mixing and matching methods and not being consistent.  I see nothing wrong with using `ENV.fetch` throughout.  Would this work?
```
 ENV.fetch('CREW_CACHE_BUILD', nil)&.eql?('1')
```

satmandu: We could try it!

satmandu: @supechicken what do you think?

lib/const.rb
```diff
@@ -120,21 +120,21 @@
...
 # Set following as boolean if environment variables exist.
-CREW_CACHE_ENABLED                   = !!ENV['CREW_CACHE_ENABLED']&.eql?('1')
-CREW_CONFLICTS_ONLY_ADVISORY         = !!ENV['CREW_CONFLICTS_ONLY_ADVISORY']&.eql?('1')         # or use conflicts_ok
-CREW_DISABLE_ENV_OPTIONS             = !!ENV['CREW_DISABLE_ENV_OPTIONS']&.eql('1')              # or use no_env_options
-CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY = !!ENV['CREW_FHS_NONCOMPLIANCE_ONLY_ADVISORY']&.eql?('1') # or use no_fhs
-CREW_NOT_COMPRESS                    = !!ENV['CREW_NOT_COMPRESS']&.eql?('1')                    # or use no_compress
-CREW_NOT_LINKS                       = !!ENV['CREW_NOT_LINKS']&.eql?('1')                       # or use no_links
-CREW_NOT_STRIP                       = !!ENV['CREW_NOT_STRIP']&.eql?('1')                       # or use no_strip
-CREW_NOT_SHRINK_ARCHIVE              = !!ENV['CREW_NOT_SHRINK_ARCHIVE']&.eql?('1')              # or use no_shrink
+CREW_CACHE_ENABLED                   = !ENV['CREW_CACHE_ENABLED']&.eql?('1').nil?
```
supechicken, line 130: Note that `.eql?('1')` will return `false` when it is not empty and not equal to `1`, adding `!` and `.nil?` will return `true` anyways

satmandu: Can you adjust the logic such that it doesn't trip up rubocop? Or alternately add an exception to the rubocop config so rubocop doesn't trip on the original code?

Zopolis4: Are we not able to simply fetch these from the environment like so?
```ruby
CREW_CACHE_ENABLED = ENV.fetch('CREW_CACHE_ENABLED', false)
```
As I understand it, this results in `CREW_CACHE_ENABLED` being false by default, and true if it has an environment variable set (given that Ruby only views `false` and `nil` as false, and everything else as true)

lib/const.rb
supechicken: Like my view towards `Dir[]` and `Dir.glob()`, I prefer mixing them. (but I am okay with using one of them only :) )

lib/const.rb
supechicken: @Zopolis4 See my comment in #8889:
```
What is wrong with `ENV.fetch` or simply `CREW_BUILD_FROM_SOURCE = ENV['CREW_BUILD_FROM_SOURCE']`?

Some people might be uncomfortable with actually unsetting a variable with unset
(like me :) ), they would probably run something like `VAR=''` to `"unset"` them
(the env variable still exists but being empty).

Also, this prevents unexpected results and misunderstanding,
like `export CREW_BUILD_FROM_SOURCE=false` being evaluated as true
```

@satmandu How about this:
```ruby
CREW_CACHE_ENABLED = ENV.fetch('CREW_CACHE_ENABLED', false).eql?('1')
```

Zopolis4: Perhaps we could document somewhere that we require variables to be properly unset? It just feels like a lot of extra code for what feels like a boilerplate task.

supechicken: I think we should check and prevent those "edge cases" (not every user will read the docs, they will open an issue first)

However, we can still make a suggestion on it

lib/const.rb
uberhacker: The DRY method does not apply here. :)

satmandu:
@supechicken I'm also happy to revert to the current working version of `const.rb` if you just want to add rules to `.rubocop.yml` to make it not complain!

I have no problems otherwise with the existing file.

satmandu:
> This should work, although it kinda bypasses Handoff.yml and means we'll be running the unit tests more than we need to (e.g. when markdown files are changed)

@Zopolis4 I modified the code to start the container unit tests from handoff...

.github/workflows/Unit-Test.yml
```diff
@@ -1,18 +1,17 @@
+---
 name: Run Unit Tests on PR
-
 on:
-  pull_request:
-    branches: [master]
+  workflow_call
```
Zopolis4, line 4: You can do this as one line
```yaml
on: workflow_call
```

lib/const.rb
satmandu: @supechicken Made as much as possible use that format with 
```ruby
CREW_VARIABLE = ENV.fetch('CREW_VARIABLE', '0').eql?('1')
```

lib/const.rb
satmandu: As mentioned in the other comment, I've used this format as much as possible:

```ruby
CREW_VARIABLE = ENV.fetch('CREW_VARIABLE', '0').eql?('1')
```

uberhacker (approved):

PR #8916, opened 2023-11-10
crew: Remove `.tpxz` support

supechicken:
Since we have switched to `.tar.zst` as our prebuilt binary format for a while, the `.tpxz` logic should be no longer needed.

For existing packages with the `.tpxz` format, the `tar.xz` logic will be used instead (`.tpxz` files are backward compatible with normal `xz`).

This PR also included some cleanups.

```
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=refactor-1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -962,28 +962,21 @@ def unpack(meta)
       case File.basename meta[:filename]
       when /\.zip$/i
         puts "Unpacking archive using 'unzip', this may take a while..."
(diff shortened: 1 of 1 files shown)


Zopolis4:
We can also remove the need to install `pixz` from `install.sh`

satmandu:
> We can also remove the need to install `pixz` from `install.sh`

That can be a future PR....

satmandu (approved):

PR #8918, opened 2023-11-10
Rubyize on some packages

supechicken:
This PR converts file operations with commands into the ruby way on some packages.


packages/xdotool.rb
```diff
@@ -35,7 +35,5 @@ def self.build
 
   def self.install
     system "LDCONFIG=#{CREW_PREFIX}/sbin/ldconfig DESTDIR=#{CREW_DEST_DIR} PREFIX=#{CREW_PREFIX} INSTALLLIB=#{CREW_LIB_PREFIX} make install"
-    # Remove execute bit from header file'
-    system "chmod -x #{CREW_DEST_PREFIX}/include/xdo.h"
```
satmandu, line 39: Was this going to be rubyized too?

satmandu (approved):

packages/xdotool.rb
supechicken: I removed this part as it seems not necessary

PR #8920, opened 2023-11-12
rebuild bison,sed; update cpio, strace

satmandu:
Rebuilds of packages and updates to address issues brought up in https://github.com/chromebrew/chromebrew/issues/8823

Works properly:
- [x] `x86_64`
- [ ] `i686`
- [ ] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=some_updates crew update
```


Zopolis4:
We can drop the code in `install.sh` for this case now, yeah?

satmandu:
This is a subset of the rebuilds...

Zopolis4:
Oh...

packages/elinks.rb
```diff
@@ -1,40 +1,49 @@
...
-    system 'make'
-  end
+  meson_options '-Dcgi=true \
+    -D256-colors=true \
+    -Dhtml-highlight=true \
+    -Didn=false \
+    -Dgemini=true \
+    -Dbrotli=true \
+    -Dbzlib=true \
+    -Dzlib=true'
```
Zopolis4, line 44: `zlib` is true by default in `0.16.1.1`

https://github.com/rkd77/elinks/blob/v0.16.1.1/meson_options.txt#L44

packages/elinks.rb
```diff
@@ -1,40 +1,49 @@
...
 
-  def self.install
-    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
-    FileUtils.rm_f "#{CREW_DEST_PREFIX}/share/locale/locale.alias"
+  def self.patch
+    system "sed -i 's,/usr/bin/perl,#{CREW_PREFIX}/bin/perl,g' doc/tools/help2xml"
```
Zopolis4, line 47: There are a couple instances of us replacing `/usr/bin/perl` shebangs-- Is there a upstream-acceptable solution for this? Is `#! /usr/bin/env perl` or similar acceptable? Alternatively, is there a problem with using the system perl?

packages/r.rb
```diff
@@ -46,25 +45,7 @@ class R < Package
   depends_on 'xzutils' # R
   depends_on 'zlibpkg' # R
 
-  def self.build
-    @x = ARCH == 'i686' ? '' : '--with-x'
-    system "./configure #{CREW_OPTIONS} \
-           --disable-maintainer-mode \
+  configure_options "--disable-maintainer-mode \
```
Zopolis4, line 48: Maintainer mode is disabled by default in `4.3.2`.

https://github.com/wch/r-source/blob/tags/R-4-3-2/configure.ac#L321-L327

packages/openjpeg.rb
```diff
@@ -1,41 +1,36 @@
...
+  depends_on 'libtiff' # R
+  depends_on 'libwebp' # R
+  depends_on 'xzutils' # R
+  depends_on 'zlibpkg' # R
+  depends_on 'zstd' # R
 
-  def self.install
-    system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
-  end
+  cmake_options "-DBUILD_SHARED_LIBS=ON \
```
Zopolis4, line 34: `DBUILD_SHARED_LIBS` is `ON` by default in `2.5.0`.

https://github.com/uclouvain/openjpeg/blob/v2.5.0/CMakeLists.txt#L186

packages/libjxl.rb
```diff
@@ -6,21 +6,22 @@
 class Libjxl < Package
   description 'JPEG XL image format reference implementation'
   homepage 'https://jpeg.org/jpegxl/'
-  version '0.8.2'
+  @_ver = '0.8.2'
```
Zopolis4, line 9: Same thing with `@_ver` and not needing to set a version suffix anyways...

packages/graphviz.rb
```diff
@@ -3,59 +3,53 @@
...
   depends_on 'libx11' # R
   depends_on 'libxrender' # R
   depends_on 'pango' # R
-  depends_on 'qt5_base' unless ARCH == 'i686'
+  depends_on 'qt5_base' # R
   depends_on 'zlibpkg' # R
-  depends_on 'ghostscript' # R
 
   def self.patch
     system "sed -i 's|/usr|#{CREW_PREFIX}|g' cmake/config_checks.cmake"
```
Zopolis4, line 44: Assuming I'm reading this sed correctly, `/usr` no longer exists in `cmake/config_checks.cmake` in `9.0.0`.

https://gitlab.com/graphviz/graphviz/-/blob/5733d3a95898f1380424ab15f966ace9a283d506/cmake/config_checks.cmake

packages/graphviz.rb
```diff
@@ -3,59 +3,53 @@
...
-    # system 'patch -Np1 -i graphviz_cmake.patch'
   end
 
   def self.build
-    # gvedit pulls in the QT dependency.
-    @gvedit = ARCH == 'i686' ? 'OFF' : 'ON'
     system "AWK=#{CREW_PREFIX}/bin/mawk \
```
Zopolis4, line 48: Any way to convert this to buildsystems? The possible obstacles I see are the need to set `AWK` to `mawk`, and the sed patching (see previous comment).

packages/graphviz.rb
```diff
@@ -3,59 +3,53 @@
...
         cmake -B builddir \
         -G Ninja \
         #{CREW_CMAKE_OPTIONS} \
-        -Dwith_gvedit=#{@gvedit}"
+        -Dwith_gvedit=ON"
```
Zopolis4, line 52: `with_gvedit` is `ON` by default in `9.0.0`

https://gitlab.com/graphviz/graphviz/-/blob/5733d3a95898f1380424ab15f966ace9a283d506/CMakeLists.txt#L10

Zopolis4 (changes requested):
A brief look-- havent reviewed everything yet.

packages/elinks.rb
satmandu: system doesn't have perl

satmandu: Feel free to suggest a patch to upstream though. :)

packages/graphviz.rb
satmandu: We don't have a pre_configure_flags yet here for cmake... we should add it in another PR?

packages/elinks.rb
Zopolis4: Happy to, just not sure what to suggest. Will `#!/usr/bin/env perl` work? That is what seems to be suggested for portability from what I can find, so...

packages/anagram.rb
```diff
@@ -3,52 +3,46 @@
 class Anagram < Package
   description 'finds anagrams or permutations of words in the target phrase'
   homepage 'https://www.fourmilab.ch/anagram/'
-  version '1.5'
+  version '1.5-1'
```
Zopolis4, line 6: No need for version suffix? (Same with other packages: `gegl`, `gimp`, `openjpeg`, `sed`)

packages/gdal.rb
```diff
@@ -1,63 +1,59 @@
-require 'package'
+require 'buildsystems/cmake'
 
-class Gdal < Package
+class Gdal < CMake
   description 'The Geospatial Data Abstraction Library is a translator for raster and vector geospatial data formats.'
   homepage 'http://www.gdal.org/'
```
Zopolis4, line 5: Redirects to `https://gdal.org/`, change?

packages/graphviz.rb
satmandu: done

Zopolis4 (changes requested):
few more reviews, just waiting to see if graphviz builds without the pre-configure option. (still good to keep that preconfigure code for the cmake buildsystem, regardless)

Zopolis4:
Not sure what happened-- github doubled up my review, but I think i deleted both of them?

At any rate, my reviews were just to change the homepage in `gegl.rb` to what it redirects to, and remove version suffixes in the other packages: `gegl gimp openjpeg sed anagram`

packages/imagemagick7.rb
```diff
@@ -67,26 +68,15 @@ class Imagemagick7 < Package
...
 
-  def self.patch
-    system 'filefix'
-  end
-
-  def self.build
-    system "./configure \
-      #{CREW_OPTIONS} \
-      --mandir=#{CREW_MAN_PREFIX} \
+  configure_options "--mandir=#{CREW_MAN_PREFIX} \
```
Zopolis4, line 77: `magick-plus-plus` and `enable-hdri` are enabled by default.

packages/graphviz.rb
```diff
@@ -1,63 +1,46 @@
...
-    # gvedit pulls in the QT dependency.
-    @gvedit = ARCH == 'i686' ? 'OFF' : 'ON'
-    system "AWK=#{CREW_PREFIX}/bin/mawk \
-        cmake -B builddir \
-        -G Ninja \
-        #{CREW_CMAKE_OPTIONS} \
-        -Dwith_gvedit=#{@gvedit}"
-    system "#{CREW_NINJA} -C builddir"
-  end
+  pre_cmake_options "AWK=#{CREW_PREFIX}/bin/mawk"
```
Zopolis4, line 43: In my testing this is not needed.

satmandu:
> Not sure what happened-- github doubled up my review, but I think i deleted both of them?
> 
> At any rate, my reviews were just to change the homepage in `gegl.rb` to what it redirects to, and remove version suffixes in the other packages: `gegl gimp openjpeg sed anagram`

gegl.org does not redirect?

Zopolis4:
> gegl.org does not redirect?

Oops, I meant `http://www.gdal.org/`, which goes to `https://gdal.org/`.

Nothing major on either front, but just nice to keep things up-to-date.

satmandu:
versions fixed up

satmandu:
One problem with not using `-1` in versions is that package installs are broken until this PR is merged...

Zopolis4:
Not entirely sure what you mean?

If you mean that not using -1 won't work until 8587 is merged, I see a pretty clear solution to that 😉

satmandu:
> Not entirely sure what you mean?
> 
> If you mean that not using -1 won't work until 8587 is merged, I see a pretty clear solution to that 😉 
> 
> 

Without unique versions for packages, old packages won't work because the hashes therein are now broken. As some of these packages are now required for installs, Chromebrew installs are now broken until this PR is merged.

Zopolis4:
Oh-- uploading the binaries replaces the old ones, with different sha256's, and Chromebrew doesn't expect these new sha256 files.

Zopolis4:
In future, it would probably make sense to not upload the binaries until after the PR is merged, as this will result in a much smaller period of time (a few minutes, depending on the binary) during which these packages cannot be upgraded.

Zopolis4:
Although-- the only packages which have the same binary urls (i.e. no version change and no difference in binary compression) are gegl, openjpeg and libjxl, which aren't critical for the installation.

satmandu:
> Although-- the only packages which have the same binary urls (i.e. no version change and no difference in binary compression) are gegl, openjpeg and libjxl, which aren't critical for the installation.

Buildessential requires `bison`...

Zopolis4:
Bison is changed from .tpxz to .tar.zst, so we haven't replaced the old binaries.

satmandu:
Ah you're right! We got lucky there!

packages/sed.rb
```diff
@@ -16,27 +16,14 @@ class Sed < Package
...
-  def self.check
-    # Remove selinux tests since we're building without it.
-    system "sed -i 's,testsuite/inplace-selinux.sh ,,' Makefile"
-    system 'make', 'check'
-  end
-
-  def self.install
-    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
-  end
+  configure_options '--without-selinux'
```
Zopolis4, line 28: Not entirely confident, but it looks like ChromeOS has SELinux? https://www.chromium.org/chromium-os/developer-library/reference/security/selinux/

lib/buildsystems/cmake.rb
```diff
@@ -1,13 +1,13 @@
 require 'package'
 
 class CMake < Package
-  property :cmake_options
+  property :cmake_options, :pre_cmake_options
 
   def self.build
-    puts "Additional cmake_options being used: #{@cmake_options.nil? || @cmake_options.empty? ? '<no cmake_options>' : @cmake_options}".orange
+    puts "Additional cmake_options being used: #{@pre_cmake_options.nil? || @pre_cmake_options.empty? ? '<no pre_cmake_options>' : @pre_cmake_options} #{@cmake_options.nil? || @cmake_options.empty? ? '<no cmake_options>' : @cmake_options}".orange
```
supechicken, line 7: The `pre_cmake_options` will not be empty? (no need to check if it is an empty string)

packages/sed.rb
satmandu: Historically this has led to build issues...

lib/buildsystems/cmake.rb
satmandu: This should be the same for all of the `pre_` and non-`pre_` options in all of the buildsystems files.

Do you have suggested code for all of it?

satmandu: I will try removing `@cmake_options.empty?` and so on to see if anything breaks. :)

satmandu: Changes made!

supechicken (approved):

PR #8939, opened 2023-11-13
Convert workflows to use reviewdog, run rubocop on tree

Zopolis4:
Should make dealing with lint failures much nicer.


satmandu (approved):
Let's see how this works!

Zopolis4:
This means we can both delete our respective Octocop forks-- the company behind Octocop, Freshly, seems to have gone out of business, so I doubt our PRs will get merged anytime soon.

PR #8945, opened 2023-11-15
Add warnings for Grunt (AMD StoneyRidge) devices

### LINKED ISSUE #8823: 🐞 Bug Multiple binaries segfaulting on AMD StoneyRidge when ASLR is enabled.

Juma7C9:
Fixes #8823

## Description
Add a warning of possible broken packages if crew is installed or run on an AMD StoneyRidge device.

## Additional information

* Added a warning on `install.sh` at the end of the installation process, with a link to the GitHub page;
* Added a test on `lib/const.rb` to detect the CPU family;
* Added a warning on `bin/crew` at the end of the package installation routine, which also proposes alternate solutions (rebuilding from source and disabling ASLR).

As I am not familiar with the project, nor the style (or even the language :P), I don't know if that's the best place to put such a warning.
Also the warning may be a bit obnoxious as it is a bit verbose, as I tried to convey the information as exhaustively as possible. It may be a good idea to add some logic to disable or reduce it, but I didn't want to clutter the code too much.

Works properly:
- [x] `x86_64`
- [ ] `i686`
- [ ] `armv7l` 

```
CREW_REPO=https://github.com/Juma7C9/chromebrew.git CREW_BRANCH=grunt_warn crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,7 +1,7 @@
 # lib/const.rb
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.39.2'
+CREW_VERSION = '1.39.3'
 
 # kernel architecture
 KERN_ARCH = `uname -m`.chomp
@@ -74,9 +74,6 @@ CREW_CPU_VENDOR = CPUINFO['vendor_id'] || 'unknown'
 # cpuinfo vendor_id.
 CREW_IS_AMD   = CREW_CPU_VENDOR.eql?('AuthenticAMD')
 CREW_IS_INTEL = %w[x86_64 i686].include?(ARCH) && %w[unknown GenuineIntel].include?(CREW_CPU_VENDOR)
-# Used to identify Grunt (AMD StoneyRidge, family 21) boards;
-# May also be used to identify other sub-variants, exp. AMD or ARM (using some other key).
-CREW_CPU_FAMILY = %w[x86_64 i686].include?(ARCH) ? CPUINFO['cpu family'] : ''
 
 # Use sane minimal defaults if in container and no override specified.
 if CREW_IN_CONTAINER && ENV['CREW_KERNEL_VERSION'].nil?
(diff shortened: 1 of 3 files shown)


satmandu:
Note the rubucop failure. You can install `ruby_rubocop` and do a `rubocop -A file.rb` to fix the file in question so it will pass the linter...

satmandu:
In your git tree you can also do a `rubocop -c .rubocop.yml -A bin/crew`...

lib/const.rb
satmandu, line 1: Please bump `CREW_VERSION`.

satmandu (changes requested):
Otherwise I don't have any complaints.

lib/const.rb
```diff
@@ -74,6 +74,9 @@
 # cpuinfo vendor_id.
 CREW_IS_AMD   = CREW_CPU_VENDOR.eql?('AuthenticAMD')
 CREW_IS_INTEL = %w[x86_64 i686].include?(ARCH) && %w[unknown GenuineIntel].include?(CREW_CPU_VENDOR)
+# Used to identify Grunt (AMD StoneyRidge, family 21) boards;
+# May also be used to identify other sub-variants, exp. AMD or ARM (using some other key).
+CREW_CPU_FAMILY = %w[x86_64 i686].include?(ARCH) ? CPUINFO['cpu family'] : ''
```
supechicken, line 79: No need to create a separate constant if it is available under `CPUINFO`

lib/const.rb
Juma7C9: Should I increment the revision number, i.e. `1.39.2` -> `1.39.3`?

satmandu (approved):
Looks Good To Me!

PR #8955, opened 2023-11-17
Add difftastic package

uberhacker:
Difftastic is a structural diff tool that understands syntax. It supports [over 30 programming languages](https://difftastic.wilfred.me.uk/languages_supported.html) and when it works, it's fantastic.  See https://difftastic.wilfred.me.uk/.

### DIFF
--- a/packages/difft.rb
+++ b/packages/difft.rb
@@ -1,0 +1,14 @@
+require 'package'
+require_relative 'difftastic'
+
+class Difft < Package
+  description Difftastic.description
+  homepage Difftastic.homepage
+  version Difftastic.version
+  license Difftastic.license
+  compatibility Difftastic.compatibility
+
+  is_fake
+
+  depends_on 'difftastic'
+end
--- a/packages/difftastic.rb
+++ b/packages/difftastic.rb
@@ -1,0 +1,37 @@
+require 'package'
+
+class Difftastic < Package
+  description 'Difftastic is a structural diff tool that compares files based on their syntax.'
+  homepage 'https://github.com/Wilfred/difftastic'
+  version '0.52.0'
+  license 'MIT'
+  compatibility 'all'
+  source_url 'https://github.com/Wilfred/difftastic.git'
+  git_hashtag version
+
+  binary_url({
+    aarch64: 'https://gitlab.com/api/v4/projects/26210301/packages/generic/difftastic/0.52.0_armv7l/difftastic-0.52.0-chromeos-armv7l.tar.zst',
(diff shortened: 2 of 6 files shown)


packages/difft.rb
```diff
@@ -0,0 +1,14 @@
+require 'package'
+require_relative 'difftastic'
```
Zopolis4, line 2: Do we really need an alias for this? After all, it's named upstream as difftastic.

packages/difftastic.rb
```diff
@@ -0,0 +1,35 @@
+require 'package'
+
+class Difftastic < Package
+  description 'Difftastic is a structural diff tool that compares files based on their syntax.'
+  homepage 'https://github.com/Wilfred/difftastic'
+  version '0.52.0'
+  license 'MIT'
+  compatibility 'x86_64 aarch64 armv7l'
```
Zopolis4, line 8: This is a terminal app, does it  it work on i686?

packages/difftastic.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def self.install
+    system "cargo install \
+      --locked difftastic \
+      --path . --root #{CREW_DEST_PREFIX}"
+    FileUtils.rm_rf Dir["#{CREW_DEST_PREFIX}/.crate*"]
+  end
+
+  def self.postinstall
+    puts "\nType 'difft' to get started.\n".lightblue
```
Zopolis4, line 33: There's no setup or anything, do we need this postinstall message?

Zopolis4 (dismissed):

packages/difft.rb
satmandu: It would be nice to have some alias functionality in crew, like we have our table of adjustments in `fixup.rb` such that crew knows when there is an alias without having to create a separate package file.
maybe an aliases line in the package?

satmandu: Also, `diift` is the name of the binary, so people might look for a `difft` package.

supechicken: > Also, `difft` is the name of the binary, so people might look for a `difft` package.

User can use `crew whatprovides` to search the package with the command name

packages/difftastic.rb
uberhacker: It might work but is it really worth the effort?

packages/difft.rb
uberhacker: Most users don't know how to use `crew whatprovides`.  My policy is to always provide a post install message if the package name differs from the cli command.  It helps avoid confusion.

packages/difftastic.rb
satmandu: I think it is entirely reasonable to mention the name of the relevant binary in the `postinstall`, especially when it differs from the package name,

satmandu (approved):

PR #8987, opened 2023-11-26
Remove spellcheck_test and rework load_test into prop_test

Zopolis4:
Add some relatively basic testing for missing variables, should allow us to remove some handling for this in `crew` itself.

Someone more experienced with ruby than me-- does loading all of the packages here mean we can also remove checks like [this](https://github.com/chromebrew/chromebrew/blob/master/bin/crew#L172)? If not, what CI testing will allow us to remove them?

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=smest crew update
```


.github/workflows/Unit-Test.yml
```diff
@@ -19,4 +19,8 @@ jobs:
             CREW_REPO=${{ github.event.pull_request.head.repo.clone_url }} CREW_BRANCH=${{ github.head_ref }} crew update && \
             yes | crew upgrade && \
             yes | crew install vim && \
-            yes | crew remove vim"
+            yes | crew remove vim && \
+            git sparse-checkout add ../tests && \
```
satmandu, line 23: We can avoid this by just adding `tests` to `install.sh`. Could we do that?

We just have to change these lines:
```
  git sparse-checkout set packages "manifest/${USER_SPACE_ARCH}" lib bin crew tools
  git reset --hard origin/"${BRANCH}"
```
to
```
  git sparse-checkout set packages "manifest/${USER_SPACE_ARCH}" lib bin crew tests tools
  git reset --hard origin/"${BRANCH}"
```

satmandu: (Then the next image build will pick that up, and the image used for the workflows will have it.)

satmandu: Let me do a separate PR for that...

satmandu:
Please merge master!

.github/workflows/Unit-Test.yml
satmandu, line 23: This sparse checkout of tests should no longer be needed, as it is added from `install.sh`.

satmandu:
Also, the unit tests are failing...

satmandu:
The unit test is reporting some sort of issues with missing licenses for a bunch of packages...

Zopolis4:
> The unit test is reporting some sort of issues with missing licenses for a bunch of packages...
Thats the intention-- `prop_test` is a basic test designed to catch missing licenses, homepages, compatibilities, etc. I'll fix the relevant packages soon.

satmandu:
Also, `pipewire.rb` does have a version, it is just dependent upon `LIBC_VERSION`, so any version check really needs to be in ruby, or understand the ruby logic.

Zopolis4:
The test is written in ruby and works with crew directly, so I presume the failure with `pipewire.rb` is the assignment of `version` using `=` when that's not how we're supposed to use it. I've gotten it to work with `@_ver`, but...

Zopolis4:
nvm, found better solution for `pipewire`

.github/workflows/Unit-Test.yml
satmandu: ok, this line should really not be needed after #8994 is merged and I rebuild containers! I will ping when that is done...

satmandu: Ok, this should be good to be removed. Both crew and install.sh are now adding the `tests` dir...

satmandu:
Does it help for pipewire if we change the version code to 
```ruby
  @_ver = if Gem::Version.new(CREW_KERNEL_VERSION.to_s) < Gem::Version.new('3.9')
              '0.3.29'
            elsif Gem::Version.new(CREW_KERNEL_VERSION.to_s) <= Gem::Version.new('5.4')
              '0.3.60'
            else
              '1.0.0'
            end
  version @_ver
```

uberhacker:
> Does it help for `pipewire.rb` if we change the version code to the following?
> 
> ```ruby
>   @_ver = if Gem::Version.new(CREW_KERNEL_VERSION.to_s) < Gem::Version.new('3.9')
>               '0.3.29'
>             elsif Gem::Version.new(CREW_KERNEL_VERSION.to_s) <= Gem::Version.new('5.4')
>               '0.3.60'
>             else
>               '1.0.0'
>             end
>   version @_ver
> ```

Maybe?  How do we version it currently?

satmandu (approved):
Is this good to be merged or are you still adding commits?

Zopolis4:
Good to be merged.

PR #8994, opened 2023-11-27
Add perl buildsystem

satmandu:
## Description
- Adds perl buildsystem. Note that the package name internally is `PERL` so as not to conflict with the `perl` package.
- Rebuilds perl packages that had files in the `/usr/local/lib/perl5/5.38.0` dir.

## Additional information

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=perl_buildsystem crew update
```


bin/crew
```diff
@@ -657,7 +657,7 @@ def update
   # update package lists
   Dir.chdir(CREW_LIB_PATH) do
     # Set sparse-checkout folders.
-    system "git sparse-checkout set packages manifest/#{USER_SPACE_ARCH} lib bin crew tools"
+    system "git sparse-checkout set packages manifest/#{USER_SPACE_ARCH} lib bin crew tests tools"
```
Zopolis4, line 660: I've already got this change in #8993, best not to double up.

Zopolis4 (dismissed):

lib/buildsystems/perl.rb
```diff
@@ -0,0 +1,20 @@
+require 'package'
+
+class PErl < Package
```
uberhacker, line 3: How about `PERL` all caps?  This mixed case is a bit confusing.

satmandu: Sure.

lib/buildsystems/perl.rb
```diff
@@ -0,0 +1,20 @@
...
+  property :pre_perl_options
+
+  def self.prebuild
+    puts "Additional pre_perl_options being used: #{@pre_perl_options.nil? ? '<no pre_perl_options>' : @pre_perl_options}".orange
+    system "#{@pre_perl_options} perl Makefile.PL"
+    system "sed -i 's,/usr/local,#{CREW_PREFIX},g' Makefile"
```
uberhacker, line 9: Could this be in `def self.patch` or does that run before prebuild?

packages/perl_data_dumper.rb
```diff
@@ -3,23 +3,23 @@
 class Perl_data_dumper < Package
```
uberhacker, line 3: No buildsystems/perl?

packages/perl_file_spec.rb
```diff
@@ -3,23 +3,23 @@
 class Perl_file_spec < Package
```
uberhacker, line 3: No buildsystems/perl?

packages/perl_data_dumper.rb
satmandu: The more complicated packages I did not change.

packages/perl_file_spec.rb
satmandu: The more complicated packages I did not change.

lib/buildsystems/perl.rb
satmandu: patch runs before prebuild

packages/perl_data_dumper.rb
satmandu: changed

packages/perl_file_spec.rb
satmandu: changed

uberhacker (approved):

PR #8979, opened 2023-11-23
Fix to get true local repo root

uberhacker:
We don't need to install the manifest files in `/usr/local/lib/crew/manifest`.  We want to store them in `/path/to/local/repo/root/manifest` instead.  The caveat here is the crew commands must be run from inside the local repo.  We only really need `CREW_LOCAL_REPO_ROOT` for `crew build`.  The builds should be run inside the repo in the `release/#{ARCH}` directories anyway.

### Run the following to get this pull request's changes locally for testing.
```
CREW_REPO=https://github.com/uberhacker/chromebrew.git CREW_BRANCH=update-const.rb crew update
```

PR #8995, opened 2023-11-27
Change local repo constants for building packages

uberhacker:
- Adjust logic to allow multiple builds/installs without aborting
- Ensure builds are performed in the correct location

This PR attempts to build upon #8979 and improve the build//install process to fix things that have been bothering me for awhile.  There is no need to set `CREW_LOCAL_MANIFEST_PATH` since that constant has been removed.  Hopefully, this workflow makes sense.


lib/const.rb
```diff
@@ -104,15 +102,13 @@
 CREW_DEST_MAN_PREFIX  = File.join(CREW_DEST_DIR, CREW_MAN_PREFIX)
 
 # Local constants for contributors.
-CREW_LOCAL_REPO_ROOT = `cd #{CREW_LIB_PATH} && git rev-parse --show-toplevel 2> /dev/null`.chomp
+CREW_LOCAL_REPO_ROOT = `git rev-parse --show-toplevel 2> /dev/null`.chomp
+CREW_LOCAL_BUILD_DIR = "#{CREW_LOCAL_REPO_ROOT}/release/#{USER_SPACE_ARCH}"
```
satmandu, line 106: My problem here is that this requires the build to be started in a git dir, and that's not my workflow. I start my build in a tmpdir outside of a git tree.

This PR ensures that manifest files are only copied out of the build into `#{CREW_LOCAL_REPO_ROOT}/manifest/`, but then `CREW_LOCAL_REPO_ROOT` is only defined if the crew is run inside a git dir...

If crew build isn't started inside a git clone of a chromebrew repo, builds just don't work, right?

uberhacker:
> These changes break my workflow, and it looks like they break satmandu's too.
> 
> Given that, what is the rationale behind them?

The rational is to make it easier to build packages and submit prs without jumping through loops with setting environment variables and copying files all over the place.  How will this break your workflow?

Zopolis4:
My workflow involves building packages inside a git tree, and then committing the new manifests. Because the manifests are already in the correct position, all I need to run is 'git commit -am <message>', and I don't need to copy any files or set any environment variables.

I also update the package files with the new binary urls and sha256s in the same go, using the names and hashes of the just-built binaries to do so.

uberhacker:
> My workflow involves building packages inside a git tree, and then committing the new manifests. Because the manifests are already in the correct position, all I need to run is 'git commit -am ', and I don't need to copy any files or set any environment variables.
> 
> I also update the package files with the new binary urls and sha256s in the same go, using the names and hashes of the just-built binaries to do so.

You can still do that with these changes.  Everything is in the git repo.

Zopolis4:
I get this when trying to run my script:
```
Unable to locate local build directory /usr/local/lib/crew/release/x86_64.
```

I don't use the release/#{ARCH} structure, so requiring it breaks my workflow.

uberhacker:
> I get this when trying to run my script:
> 
> ```
> Unable to locate local build directory /usr/local/lib/crew/release/x86_64.
> ```
> 
> I don't use the release/#{ARCH} structure, so requiring it breaks my workflow.

There is a way to resolve this issue.  We all need to agree on a workflow.  The `release/ARCH` directory structure is used by gl.sh.  This is what @satmandu and myself use to build all binaries and push upstream.  We don't need to rewrite multiple workflows.  That is just reinventing the wheel.  Let's work together to come up with something we can all agree upon.

Zopolis4:
It wouldn't be too hard for me to rework my `binaries.sh` script to work with `release/ARCH`-- that's the only holdup for this working with my workflow, so I'm fine with this.

Thank you for developing on a solution that will work for everyone.

Zopolis4:
@supechicken I'm not sure what your workflow for building packages is, will it work with these changes?

Zopolis4:
Nitpicking-- would it make more sense for the directory to be `builds/ARCH`? 

We'd need to change the `.gitignore` as well as `tools/gl.sh`, but that's about it if we were to go down that path.

I don't really mind, just throwing the idea out there.

supechicken:
> @supechicken I'm not sure what your workflow for building packages is, will it work with these changes?

I don't have it actually😅 I gave up the idea of building packages by myself after I realized that all binaries generated by myself are segfaulting on other user's systems.

uberhacker:
> > @supechicken I'm not sure what your workflow for building packages is, will it work with these changes?
> 
> I don't have it actually😅 I gave up the idea of building prebuilt binaries by myself after I realized that all binaries generated by myself are segfaulting on other user's systems.

Yikes, have you tried @satmandu's Docker images yet?

satmandu (approved):

satmandu (approved):

PR #9038, opened 2023-12-03
Add thorium package

uberhacker:
Requested by @DarkevilPT over at https://github.com/uberhacker/chromebrew-scripts/issues/2. Tested on x86_64.


DarkevilPT:
I also think they have an arm64 version here: https://github.com/Alex313031/Thorium-Raspi/releases

uberhacker:
Unfortunately, since ChromeOS runs in a 32-bit user space, aarch64 binaries won't work.

DarkevilPT:
> Unfortunately, since ChromeOS runs in a 32-bit user space, aarch64 binaries won't work.

But ChromiumOS does...  check about this:
https://community.fydeos.io/t/topic/24891
https://openfyde.com/

FydeOS provides us the Chromium Fork for ARM64 on RK3588 and we can use Chromebrew on it as well as most arm64 devices running chromiumOS forks.

Those compatible versions of arm64 can be used and I can test them. Please.. keep this door open it would be nice to always have options.

PR #9040, opened 2023-12-03
Add missing quote to flutter.rb

Zopolis4:
Not sure how this snuck past rubocop in #9036, but it's fixed now.


uberhacker (approved):
Nice find.

PR #9041, opened 2023-12-04
Run Rubocop 1.58 on tree

Zopolis4:
Not tested fully yet, but just putting this here because otherwise PR's will begin to fail.

Just linting the tree with the new cops.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=nubocop crew update
```


satmandu (approved):
Looks like nothing breaks.

PR #9085, opened 2023-12-14
Traceroute 2.1.0 => 2.1.4

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/traceroute.rb
+++ b/packages/traceroute.rb
@@ -3,23 +3,23 @@ require 'package'
 class Traceroute < Package
   description 'Traceroute tracks the route packets taken from an IP network on their way to a given host.'
   homepage 'http://traceroute.sourceforge.net/'
-  version '2.1.0'
+  version '2.1.4'
   license 'GPL-2 and LGPL-2.1'
   compatibility 'all'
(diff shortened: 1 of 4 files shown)


packages/traceroute.rb
```diff
@@ -3,23 +3,23 @@
 class Traceroute < Package
   description 'Traceroute tracks the route packets taken from an IP network on their way to a given host.'
   homepage 'http://traceroute.sourceforge.net/'
-  version '2.1.0'
+  version '2.1.4'
   license 'GPL-2 and LGPL-2.1'
   compatibility 'all'
-  source_url 'https://downloads.sourceforge.net/project/traceroute/traceroute/traceroute-2.1.0/traceroute-2.1.0.tar.gz'
-  source_sha256 '3669d22a34d3f38ed50caba18cd525ba55c5c00d5465f2d20d7472e5d81603b6'
+  source_url 'https://downloads.sourceforge.net/project/traceroute/traceroute/traceroute%202.1.4/traceroute-2.1.4.tar.gz'
```
Zopolis4, line 9: Just checking that the change to a (url-encoded) space in the URL is intended?

uberhacker: Yes, for some reason they decided to add the space instead of dash.  Must be Windoze users!  Try it yourself.

satmandu (approved):

PR #9094, opened 2023-12-15
Yt_dlp 2023.09.24 => 2023.11.16

uberhacker:
Tested on all architectures.


satmandu (changes requested):
This appears to include a bunch of dependencies looking at the file lists... Should those be broken out into separate dependencies?

uberhacker:
> This appears to include a bunch of dependencies looking at the file lists... Should those be broken out into separate dependencies?

It's a rabbit hole.  The dependencies likely change every time the yt_dlp package is updated.  I just update everything as I go because the manifest files change.

satmandu:
Can you just update py3_certifi and the other one and add them as deps so we don't accidentally overwrite them if they get installed separately?

uberhacker:
> Can you just update py3_certifi and py3_websockets and add them as deps so we don't accidentally overwrite them if they get installed separately?



> Can you just update py3_certifi and py3_websockets and add them as deps so we don't accidentally overwrite them if they get installed separately?

That's fair.

satmandu (approved):

PR #9101, opened 2023-12-18
Add install_extras properties to buildsystems

uberhacker:
I needed this for python so I decided to add to all other buildsystems just in case I might have a need in the future.  Makes everything more flexible.


satmandu (approved):

satmandu:
Can you leave an example here of the usage?

PR #9102, opened 2023-12-18
Firefox 120.0.1 => 121.0

uberhacker:
No filelist updates.


satmandu:
Firefox 121 allegedly supports WAYLAND. Do you want to see if running with wayland works on ChromeOS? (It might need a newer compositor...)

uberhacker:
> Firefox 121 allegedly supports WAYLAND. Do you want to see if running with wayland works on ChromeOS? (It might need a newer compositor...)

Can you please explain how that works in more detail?  Do I need to stop sommelier?

satmandu:
We can check for wayland compatibility later...

satmandu (approved):

PR #9109, opened 2023-12-20
Gstreamer 1.22.6 => 1.22.8

uberhacker:


packages/gstreamer.rb
```diff
@@ -1,24 +1,25 @@
-require 'package'
+require 'buildsystems/meson'
 
-class Gstreamer < Package
+class Gstreamer < Meson
   description 'GStreamer is a library for constructing graphs of media-handling components.'
   homepage 'https://gstreamer.freedesktop.org/'
-  @_ver = '1.22.6'
-  version "#{@_ver}-1"
+  @_ver = '1.22.8'
```
Zopolis4, line 6: Do we still need to be using `@_ver` here?

uberhacker: It will be needed for `"#{@_ver}-1"` again without affecting the original version.

Zopolis4: Do we anticipate the need for #{@_ver}-1 in the future? Most packages don't use @_ver by default, and we only use it when we need it.

Zopolis4: Never mind, I didn't see the updated changes.

Zopolis4 (approved):

PR #9133, opened 2023-12-28
Ruby 3.2.2-1 => 3.3.0

uberhacker:
Tested on all architectures.


Zopolis4:
We should also update the various parts of crew that are affected by the ruby version-- off the top of my head there is one in our rubocop configuration.

Zopolis4:
We also no longer need this section:
```ruby
  def self.patch
    # Get the 1.71 version of fileutils, which is not in Ruby 3.2.2.
    downloader 'https://github.com/ruby/fileutils/raw/7138d851562688267f6dc064fbceca72cb96a53e/lib/fileutils.rb',
               'c758e5c1967ec8841930fe42c7ac7bc26db72e28f208b9b175a6b6074002afe1', 'lib/fileutils.rb'
  end
```

and we also need to update this line:
```ruby
gempath: #{CREW_LIB_PREFIX}/ruby/gems/3.2.0
```

In future, when updating complex packages with large `self.build` or `self.install` sections, and so on, please check more carefully.

uberhacker:
> We also no longer need this section:
> 
> ```ruby
>   def self.patch
>     # Get the 1.71 version of fileutils, which is not in Ruby 3.2.2.
>     downloader 'https://github.com/ruby/fileutils/raw/7138d851562688267f6dc064fbceca72cb96a53e/lib/fileutils.rb',
>                'c758e5c1967ec8841930fe42c7ac7bc26db72e28f208b9b175a6b6074002afe1', 'lib/fileutils.rb'
>   end
> ```
> 
> and we also need to update this line:
> 
> ```ruby
> gempath: #{CREW_LIB_PREFIX}/ruby/gems/3.2.0
> ```
> 
> In future, when updating complex packages with large `self.build` or `self.install` sections, and so on, please check more carefully.

Thank you for the thorough check.  This is why we shouldn't be hardcoding versions and should make them variables instead.  That said, this doesn't have to be a complex package but it turned into that.  I would like to simplify this whole process.

Zopolis4:
I am always in favour of simplifying complex packages.

uberhacker:
Removed the ruby_debug package since all debug executables are already in ruby itself.

uberhacker:
@satmandu: Would love to hear your feedback on this proposal.

satmandu (commented):
I would leave out the gem versioning stuff. We aren't doing that for the python py3 packages, and we don't have the time or energy to constantly do point updates to the ruby_package files just so gem can be forced to do an update.

Just make the ruby buildsystems install the current version of a gem.

satmandu (commented):
`gem install -N` is useful, yes?

satmandu:
Also, do we need the complicated gem uninstall stuff that the gem packages used to have?

Do we need to plumb that in?

uberhacker:
> Also, do we need the complicated gem uninstall stuff that the gem packages used to have?
> 
> Do we need to plumb that in?

Ruby lets you install multiple versions of the same gem.  Uninstalling prior to installing a new version ensures all previous versions are uninstalled and any dependencies that might need updates or removed altogether.

uberhacker:
> `gem install -N` is useful, yes?

Already there?

uberhacker:
> I would leave out the gem versioning stuff. We aren't doing that for the python py3 packages, and we don't have the time or energy to constantly do point updates to the ruby_package files just so gem can be forced to do an update.
> 
> Just make the ruby buildsystems install the current version of a gem.

How can you maintain a package without versioning?

satmandu (approved):

satmandu:
necroposting here... I'm re-adding the ruby_debug package in https://github.com/chromebrew/chromebrew/pull/10436 because when there is an update to ruby's debug gem, we get annoying messages. It can in fact be updated outside of ruby's own updates...

PR #9142, opened 2024-01-03
Remove reference to ruby_debug

uberhacker:
Add conflicts_ok to ruby_concurrent_ruby

### DIFF
--- a/packages/buildessential.rb
+++ b/packages/buildessential.rb
@@ -3,7 +3,7 @@ require 'package'
 class Buildessential < Package
   description 'A collection of tools essential to compile and build software.'
   homepage 'SKIP'
-  version '1.26'
+  version '1.27'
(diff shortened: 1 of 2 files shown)


Zopolis4 (approved):

Zopolis4:
Going to assume there's a good reason for adding `conflicts_ok` to `ruby_concurrent_ruby`, and the `ruby_debug` change is self-explanatory.

uberhacker:
> Going to assume there's a good reason for adding `conflicts_ok` to `ruby_concurrent_ruby`, and the `ruby_debug` change is self-explanatory.

This is a legitimate concern since removing one package could break another.  The idea for the latest ruby update was to pass off as much of the dependency management as possible to ruby gem but still integrate with crew so we can list and search for files in gem packages.

PR #9151, opened 2024-01-06
Add starship package

uberhacker:
The minimal, blazing-fast, and infinitely customizable prompt for any shell!  See https://starship.rs/.  Tested on all architectures.

### DIFF
--- a/packages/starship.rb
+++ b/packages/starship.rb
@@ -1,0 +1,34 @@
+require 'package'
+
+class Starship < Package
+  description 'Cross-shell prompt for astronauts'
+  homepage 'https://starship.rs'
+  version '1.17.1'
+  license 'ISC'
+  compatibility 'all'
+
+  source_url({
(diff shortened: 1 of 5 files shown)


packages/starship.rb
```diff
@@ -0,0 +1,40 @@
...
+       i686: '950551fecc321e5d7834f489eba06e23456d3e646fe04bf7b1f4d8460d7a13c0',
+     x86_64: '4f325e520c4a95116a613b09aeb37f5b99265ca2b9dbf167465db52e836dbcf4'
+  })
+
+  no_compile_needed
+  print_source_bashrc
+
+  def self.build
+    File.write '99-starship', <<~EOF
+      eval "$(starship init bash)"
```
supechicken, line 28: This can be simplified to `File.write '99-starship', '$(starship init bash)'`

uberhacker: Good to know!  Always open to simplify.

satmandu (approved):

PR #9176, opened 2024-01-14
Add pocl package

uberhacker:
PoCL is a portable open source (MIT-licensed) implementation of the OpenCL standard. In addition to being an easily portable/layered multi-device open-source OpenCL implementation, a major goal of this project is improving interoperability of diversity of OpenCL-capable devices by integrating them to a single centrally orchestrated platform. Also one of the key goals is to enhance performance portability of OpenCL programs across device types utilizing runtime and compiler techniques.  See http://portablecl.org/.  Unable to build on arm.


Zopolis4:
It looks like it supports arm, what was the cause of the build failure?

uberhacker:
> It looks like it supports arm, what was the cause of the build failure?

You're welcome to give it a try yourself but it appears there is some detection logic issue.
```
-- Autodetected CPU (unknown) overridden by user to x86-64
-- Checking clang -march vs. -mcpu flag
CMake Error at cmake/LLVM.cmake:710 (message):
  Could not determine whether to use -march or -mcpu with clang
Call Stack (most recent call first):
  CMakeLists.txt:525 (include)


-- Configuring incomplete, errors occurred!
There was a build error.
```

Zopolis4:
Looked into it further-- the detection logic is [here](https://github.com/pocl/pocl/blob/main/cmake/LLVM.cmake#L648), which uses `llc --version`. 

`llc --version` outputs `Host CPU: (unknown)` on an armv7l docker container, but output my host CPU on an `x86_64` docker container. 

I presume this is why you hardcoded the CPU detection in the package file with `cmake_options '-DLLC_HOST_CPU=x86-64'`, although this obviously won't work for arm.

Long story short, I think this is one of the packages we can't build with docker, and need to use real hardware for.

satmandu (approved):

PR #9193, opened 2024-01-17
update spirv_ vulkan_ gtk3 gtk4 glslang

satmandu:
Builds properly:
- [x] `x86_64`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=vulkan_spirv_gtk_updates crew update
```


packages/spirv_tools.rb
```diff
@@ -5,54 +5,43 @@
 class Spirv_tools < Package
```
uberhacker, line 5: Use Cmake buildsystem?

packages/vulkan_icd_loader.rb
```diff
@@ -3,21 +3,21 @@
 class Vulkan_icd_loader < Package
```
uberhacker, line 3: Use CMake buildsystem?

packages/vulkan_tools.rb
```diff
@@ -6,21 +6,21 @@
 class Vulkan_tools < Package
```
uberhacker, line 6: Use CMake buildsystem?

uberhacker (dismissed):
Use CMake buildsystem?

packages/vulkan_tools.rb
satmandu: This doesn't really work. It's too complicated, as it is a dual build...

uberhacker (approved):

PR #9203, opened 2024-01-22
Update APR and APR-util, follow up on apr-iconv rename

Zopolis4:
Update APR and APR-util, `apr_iconv` needs an upstream bugfix before we can convert it to buildsystems, so just fix the rename on that one for now.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=binarycleanup1 crew update
```


packages/apr_iconv.rb
```diff
@@ -6,14 +6,14 @@ class Apr_iconv < Package
   version '1.2.2'
```
uberhacker, line 6: Bump version?

uberhacker: So you renamed the binaries too?

uberhacker: Can you please confirm no updates to the manifest files occur after a build?

Zopolis4: Binaries are renamed and reuploaded, same hashes and manifests.

uberhacker (approved):

PR #9218, opened 2024-01-25
lib/const: Rubyize external commands

supechicken:
Included #8836 in another way

This PR makes use of the `etc` built-in module to get things like `uname`, `nproc` and `glibc` version, no need to call external commands anymore.

Some cleanups are included also

```
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=rubyize_const crew update
```


lib/const.rb
```diff
@@ -102,17 +103,17 @@
 CREW_DEST_MAN_PREFIX  = File.join(CREW_DEST_DIR, CREW_MAN_PREFIX)
 
 # Local constants for contributors.
-CREW_LOCAL_REPO_ROOT = `git rev-parse --show-toplevel 2> /dev/null`.chomp
+CREW_LOCAL_REPO_ROOT = Dir.exist?('.git') ? `git rev-parse --show-toplevel`.chomp : nil
```
uberhacker, line 106: Is there another way to execute beside the backticks?  Would a system call be more Ruby-like?

satmandu: The backticks do work...

uberhacker: I know they work but the whole gist of this PR is to Rubyize and there may be an alternative "best practice" way to do it in Ruby.

uberhacker: Would `exec` work?  See https://apidock.com/ruby/Kernel/exec.

satmandu: > Would `exec` work? See https://apidock.com/ruby/Kernel/exec.

No. Exec **Replaces the current process by running the given external _command_**.

See https://stackoverflow.com/a/18623297

I'm not sure it makes sense to fiddle with this. backticks do the job, and it's easy to see what's being invoked...

satmandu: Also:
image
from https://stackoverflow.com/a/31572431

uberhacker: How about https://apidock.com/ruby/v2_5_5/Kernel/syscall?  That returns the output.

supechicken: > How about https://apidock.com/ruby/v2_5_5/Kernel/syscall? That returns the output.

The `syscall` function is for calling low-level kernel system calls like `read()` and `exec()`, not for external executables😅

We can use `Open3.capture2('cmd')[0]` or `%x[cmd]` to rubyize it, the former is better (it will not call `/bin/sh` if the command is passed in array) but longer

satmandu:
@supechicken Do you want to merge master on this?

uberhacker:
@supechicken: There are merge conflicts that need to be resolved.  Probably the crew version.

supechicken:
@uberhacker @satmandu Done

satmandu:
> @uberhacker @satmandu Done

Rubocop failing...

satmandu:
> @uberhacker @satmandu Done

```
lib/const.rb:80:3: C: Style/ConditionalAssignment: Use the return of the conditional for variable assignment and comparison.
```

Just do a `rubocop -c .rubocop.yml -A lib/const.rb`

satmandu:
@supechicken Unit tests are also failing.

lib/const.rb
```diff
@@ -42,13 +43,16 @@
 # When we have proper aarch64 support, remove this.
 ARCH = %w[aarch64 armv8l].include?(KERN_ARCH) ? 'armv7l' : KERN_ARCH
 
+# This helps determine if there is a difference between kernel and user space
+USER_SPACE_ARCH = RUBY_DESCRIPTION[/\[(.+?)-linux-gnu/, 1]
```
uberhacker, line 47: Are we adding this back? :-D

satmandu (changes requested):
Just make it pass the linter and the unit tests. :)

Zopolis4 (changes requested):
Needs a rebase and a lint, but apart from that everything looks good.

supechicken:
Fixes for container test are in #9649

satmandu (approved):

PR #9219, opened 2024-01-25
Gnupg 2.4.3 => 2.4.4

uberhacker:
No filelist updates.  All tests passed.

### DIFF
--- a/packages/gnupg.rb
+++ b/packages/gnupg.rb
@@ -1,20 +1,20 @@
-require 'package'
+require 'buildsystems/autotools'
 
-class Gnupg < Package
+class Gnupg < Autotools
   description 'GnuPG is a complete and free implementation of the OpenPGP standard as defined by RFC4880 (also known as PGP).'
   homepage 'https://gnupg.org/'
-  version '2.4.3'
+  version '2.4.4'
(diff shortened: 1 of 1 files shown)


satmandu (approved):

packages/gnupg.rb
```diff
@@ -34,20 +34,10 @@ class Gnupg < Package
...
 
-  def self.build
-    system "./configure #{CREW_OPTIONS} \
-              --enable-maintainer-mode \
-              --with-zlib \
-              --with-bzip2 \
-              --with-readline"
-    system "make -j #{CREW_NPROC} || make -j 1"
-  end
+  configure_options '--enable-maintainer-mode \
```
Zopolis4, line 37: Why are we enabling maintainer mode? We normally disable it, if it's not already disabled by default.

uberhacker: I don't think it really matters in all practical terms.  It is there to help maintainers.

Zopolis4: True, but its not useful for our usecase-- it's just an extra line of code.

Zopolis4 (approved):

PR #9220, opened 2024-01-25
Unrar 5.6.3 => 7.0.6

uberhacker:
Tested on all architectures.

### DIFF
--- a/packages/unrar.rb
+++ b/packages/unrar.rb
@@ -3,24 +3,24 @@ require 'package'
 class Unrar < Package
   description 'UnRAR is a powerful archive extractor.'
   homepage 'http://www.rarlab.com/'
-  version '5.6.3'
+  version '7.0.6'
   license 'unRAR'
   compatibility 'all'
-  source_url 'https://www.rarlab.com/rar/unrarsrc-5.6.3.tar.gz'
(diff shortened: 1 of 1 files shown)


satmandu (approved):

packages/unrar.rb
```diff
@@ -3,24 +3,24 @@
...
+     armv7l: '9f3f6421b5570cb695f841cce31b47b7585ea351e48d43844c12466700c88a21',
+       i686: '71e544df5bf2ada5f716cff0d516c81c63e883eb7dac07bb1fa5c371026fb44e',
+     x86_64: '911088681e176d910c5401c82b4982f43df1541e97c9d54761f318c96ae76cf7'
   })
 
   def self.build
     # force to compile in sequential since unrar Makefile doesn't work in parallel
-    system 'make', '-j1', 'all'
-    system 'make', '-j1', 'lib'
+    system "CPPFLAGS='-fPIC' make -j1 all"
```
Zopolis4, line 22: Is the unrar makefile still unable to compile in parallel?

uberhacker: It was failing without the `-fPIC` but I never tried to remove the `-j1` option.

PR #9223, opened 2024-01-26
install.sh: Simplify syntax and improvements

supechicken:
### Changes
- Check for `sudo` availability in a more elegant way
- Group constants in a place
- Import `/etc/lsb-release` directly instead of using `grep`
- Replace all `$(id -u)` calls with `${EUID}`
- Align comments
- Put repository URLs into variables
- Remove `pixz` since `.pixz` should be backward compatible with `xz`
- Only extract files specified in `git sparse-checkout` from the repo tarball
- Simplify syntax
- Use array on `BOOTSTRAP_PACKAGES` instead of string

### Not tested yet

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -19,8 +19,8 @@ CREW_BREW_DIR="${CREW_PREFIX}/tmp/crew"
 CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
 : "${CREW_CACHE_DIR:=$CREW_PREFIX/tmp/packages}"
 
-BOOTSTRAP_PACKAGES="zstd crew_mvdir ruby git ca_certificates openssl"
-SPRASE_CHECKOUT="packages manifest/${USER_SPACE_ARCH} lib bin crew tests tools"
+BOOTSTRAP_PACKAGES=(zstd crew_mvdir ruby git ca_certificates openssl)
+SPRASE_CHECKOUT=(packages manifest/${USER_SPACE_ARCH} lib bin crew tests tools)
 
 # Simplify colors and print errors to stderr (2).
 echo_error()   { echo -e "\e[1;91m${*}\e[0m" >&2; }  # Use Light Red for errors.
@@ -176,18 +176,18 @@ urls=()
 sha256s=()
 
 # Older i686 systems.
-[[ "${ARCH}" == "i686" ]] && BOOTSTRAP_PACKAGES+=" gcc_lib"
+[[ "${ARCH}" == "i686" ]] && BOOTSTRAP_PACKAGES+=(gcc_lib)
 
 if [[ -n "${CHROMEOS_RELEASE_CHROME_MILESTONE}" ]] && (( "${CHROMEOS_RELEASE_CHROME_MILESTONE}" > "112" )); then
   # Append the correct packages for systems running v113 onwards.
-  BOOTSTRAP_PACKAGES+=' glibc_lib235 zlibpkg gmp'
+  BOOTSTRAP_PACKAGES+=(glibc_lib235 zlibpkg gmp)
 
   # Recent Arm systems have a cut down system.
-  [[ "${USER_SPACE_ARCH}" == "armv7l" ]] && BOOTSTRAP_PACKAGES+=' bzip2 ncurses readline pcre2 gcc_lib'
+  [[ "${USER_SPACE_ARCH}" == "armv7l" ]] && BOOTSTRAP_PACKAGES+=(bzip2 ncurses readline pcre2 gcc_lib)
 fi
 
 # Get the URLs and hashes of the bootstrap packages.
-for package in $BOOTSTRAP_PACKAGES; do
+for package in "${BOOTSTRAP_PACKAGES[@]}"; do
   cd "${CREW_LIB_PATH}/packages"
   [[ "$(sed -n '/binary_sha256/,/}/p' "${package}.rb")" =~ .*${ARCH}:[[:blank:]]*[\'\"]([^\'\"]*) ]]
     sha256s+=("${BASH_REMATCH[1]}")
@@ -333,10 +333,7 @@ echo_info "Installing core Chromebrew packages...\n"
 yes | crew install core
 
 echo_info "\nRunning Bootstrap package postinstall scripts...\n"
-# Due to a bug in crew where it accepts spaces in package files names rather than
-# splitting strings at spaces, we cannot quote ${BOOTSTRAP_PACKAGES}.
-# shellcheck disable=SC2086
-crew postinstall ${BOOTSTRAP_PACKAGES}
+crew postinstall "${BOOTSTRAP_PACKAGES[@]}"
 
 if ! "${CREW_PREFIX}"/bin/git version &> /dev/null; then
   echo_error "\nGit is broken on your system, and crew update will not work properly."
--- a/install.sh
+++ b/install.sh
@@ -19,6 +19,7 @@ CREW_BREW_DIR="${CREW_PREFIX}/tmp/crew"
 CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
 : "${CREW_CACHE_DIR:=$CREW_PREFIX/tmp/packages}"
 
+BOOTSTRAP_PACKAGES="zstd crew_mvdir ruby git ca_certificates openssl"
 SPRASE_CHECKOUT="packages manifest/${USER_SPACE_ARCH} lib bin crew tests tools"
 
 # Simplify colors and print errors to stderr (2).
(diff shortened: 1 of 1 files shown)


satmandu:
Please try in containers!

install.sh
```diff
@@ -3,56 +3,81 @@
 # Exit on fail.
 set -e
 
-RESET='\e[0m'
+# Default chromebrew repo values.
+: "${GTIHUB_OWNER:=chromebrew}"
+: "${GITHUB_REPO:=chromebrew}"
+: "${GITHUB_BRANCH:=master}"
+: "${GITLAB_REPO:=26210301}"
```
Zopolis4, line 10: To be honest, I don't know if there's any practical value to changing the gitlab repo, I don't know of any other repos that mirror our package repository or would be otherwise useful to switch to.

install.sh
```diff
@@ -102,15 +109,17 @@ fi
 
 if [[ "$ARCH" == "x86_64" ]]; then
   LIB_SUFFIX='64'
+else
+  LIB_SUFFIX=''
```
Zopolis4, line 113: Is LIB_SUFFIX not already empty by default, since we don't set the variable?

install.sh
```diff
@@ -194,16 +200,16 @@ for package in $BOOTSTRAP_PACKAGES; do
   # This is really ugly, FIXME after #7082 is merged.
```
Zopolis4, line 200: Obviously this isn't something you have to do, but if you wanted you could change the code to not have two copies of each url based on the tar compression, and instead use the value of binary_compression now that that's a possibility.

Zopolis4: If not, I'll get around to it soon.

Zopolis4 (changes requested):

install.sh
```diff
@@ -3,56 +3,81 @@
...
+  else
+    echo_info "Unable to detect system information, installation will continue."
+  fi
+fi
 
 # Check if the script is being run as root.
 if [ "${EUID}" == "0" ]; then
   echo_error "Chromebrew should not be installed or run as root."
-  echo_info "Please login as 'chronos' and restart the install."
+  echo_info  "Please login as 'chronos' and restart the install."
```
uberhacker, line 44: Extra space?

install.sh
```diff
@@ -3,56 +3,81 @@
...
   exit 1
 fi
 
 # Reject crostini.
 if [[ -d /opt/google/cros-containers && "${CREW_FORCE_INSTALL}" != '1' ]]; then
   echo_error "Crostini containers are not supported by Chromebrew :/"
-  echo_info "Run 'CREW_FORCE_INSTALL=1 exec bash --init-file <(curl -Ls git.io/vddgY)' to perform install anyway"
+  echo_info  "Run 'CREW_FORCE_INSTALL=1 exec bash --init-file <(curl -Ls git.io/vddgY)' to perform install anyway"
```
uberhacker, line 51: Extra space?

install.sh
```diff
@@ -3,56 +3,81 @@
...
 fi
 
-# Check if the user owns the CREW_PREFIX directory, as sudo is unnecessary if this is the case.
 # Check if the user is on ChromeOS v117+ and not in the VT-2 console, as sudo will not work.
-: "${CREW_PREFIX:=/usr/local}"
-if [[ "$(stat -c '%u' "${CREW_PREFIX}")" == "$(id -u)" ]] && sudo 2>&1 | grep -q 'no new privileges'; then
+if grep -qx 'NoNewPrivs:[[:blank:]]\+1' /proc/self/status; then
   echo_error "Please run the installer in the VT-2 shell."
-  echo_info "To start the VT-2 session, type Ctrl + Alt + ->"
+  echo_info  "To start the VT-2 session, press Ctrl + Alt + -> and login with user 'chronos'"
```
uberhacker, line 65: Extra space?

uberhacker: End sentence with a `.`?

install.sh
```diff
@@ -3,56 +3,81 @@
...
+
 # Make sure installation directory is clean.
 if [ -d "${CREW_PREFIX}" ]; then
-  if [ "$(ls -A "${CREW_PREFIX}")" ]; then
+  if [ -e "$(ls -A "${CREW_PREFIX}")" ]; then
     echo_error "${CREW_PREFIX} is not empty, would you like it to be cleared?"
-    echo_info "This will delete ALL files in ${CREW_PREFIX}!"
-    echo_info "Continue?"
+    echo_info  "This will delete ALL files in ${CREW_PREFIX}!"
+    echo_info  "Continue?"
```
uberhacker, line 80: Extra space?

install.sh
```diff
@@ -3,56 +3,81 @@
...
-if [ -f /etc/lsb-release ]; then
-  if [[ ! "$(< /etc/lsb-release)" =~ CHROMEOS_RELEASE_TRACK=stable-channel$'\n' && "${CREW_FORCE_INSTALL}" != '1' ]]; then
-    echo_error "The beta, dev, and canary channel are unsupported by Chromebrew."
-    echo_info "Run 'CREW_FORCE_INSTALL=1 exec bash --init-file <(curl -Ls git.io/vddgY)' to perform install anyway."
-    exit 1
-  fi
-  export "$(grep CHROMEOS_RELEASE_CHROME_MILESTONE /etc/lsb-release)"
-else
-  echo_info "Unable to detect system information, installation will continue."
+if ! [[ "${CHROMEOS_RELEASE_TRACK}" == 'stable-channel' || "${CREW_FORCE_INSTALL}" == '1' ]]; then
```
uberhacker, line 56: `if [ -n "${CREW_FORCE_INSTALL}" ]; then` is good enough?

install.sh
uberhacker: This is used to set the `PREBUILT_URL` variable below to reduce duplicate long assignments.  This is actually good because if this value changes in the future, we only have to change in one place...here.

install.sh
```diff
@@ -3,56 +3,81 @@
...
+# Chromebrew directories.
+: "${CREW_PREFIX:=/usr/local}"
+CREW_LIB_PATH="${CREW_PREFIX}/lib/crew"
+CREW_CONFIG_PATH="${CREW_PREFIX}/etc/crew"
+CREW_BREW_DIR="${CREW_PREFIX}/tmp/crew"
+CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
+: "${CREW_CACHE_DIR:=$CREW_PREFIX/tmp/packages}"
+
+BOOTSTRAP_PACKAGES=(zstd crew_mvdir ruby git ca_certificates openssl)
+SPARSE_CHECKOUT=(packages manifest/${USER_SPACE_ARCH} lib bin crew tests tools)
```
uberhacker, line 23: `USER_SPACE_ARCH` not defined yet?

install.sh
uberhacker: It doesn't hurt to set it explicitly for clarity and in the off chance it's set outside of crew.

install.sh
```diff
@@ -102,15 +109,17 @@ fi
...
 # Warn users of the AMD segfault issue and allow them to work around it.
 # The easiest way to distinguish StoneyRidge platorms is to check for the FMA4
 # instruction, as it was first introduced in Bulldozer and later dropped in Zen.
-if grep -s "fma4" /proc/cpuinfo ; then
+if grep -q "fma4" /proc/cpuinfo ; then
   echo_info "Notice: You are running an AMD StoneyRidge device; due to some bugs some packages may fail with a segmentation fault and need to be rebuilt."
-  echo_info "If this happens, please report them to: https://github.com/chromebrew/chromebrew/issues"
-  echo_info "If the install fails, try running 'CREW_AMD_INSTALL=1 exec bash --init-file <(curl -Ls git.io/vddgY)'"
+  echo_info "If this happens, please report them to: ${REPO_URL}/issues"
+  echo_info "If the install fails, try running 'CREW_AMD_INSTALL=1 exec bash --init-file <(curl -L git.io/vddgY)'"
```
uberhacker, line 122: Now you are starting to think like me.  I don't like when ANY command is silenced.

install.sh
uberhacker: Reduce?
```
LIB_SUFFIX=''
[[ "${ARCH}" == "x86_64" ]] && LIB_SUFFIX='64'
```

install.sh
```diff
@@ -194,16 +200,16 @@ for package in $BOOTSTRAP_PACKAGES; do
...
+      urls+=("${PREBUILT_URL}/${name}/${version}_${ARCH}/${name}-${version}-chromeos-${ARCH}.tar.xz")
     fi
   else
     [[ "${package}" == "glibc_lib235" ]] && name="glibc_lib"
     if [[ "$ARCH" == "aarch64" ]]; then
-      urls+=("https://gitlab.com/api/v4/projects/26210301/packages/generic/${name}/${version}_armv7l/${name}-${version}-chromeos-armv7l.tar.zst")
+      urls+=("${PREBUILT_URL}/${name}/${version}_armv7l/${name}-${version}-chromeos-armv7l.tar.zst")
     else
-      urls+=("https://gitlab.com/api/v4/projects/26210301/packages/generic/${name}/${version}_${ARCH}/${name}-${version}-chromeos-${ARCH}.tar.zst")
+      urls+=("${PREBUILT_URL}/${name}/${version}_${ARCH}/${name}-${version}-chromeos-${ARCH}.tar.zst")
```
uberhacker, line 212: Reduce `"${PREBUILT_URL}/${name}/${version}_${ARCH}/${name}-${version}-chromeos-${USER_SPACE_ARCH}.tar.${ARCHIVE_EXTENSION}"` to a signal variable and simply assign `ARCHIVE_EXTENSION` instead?

Zopolis4: And we don't need to assign ARCHIVE_EXTENSION, we can just use the binary_compression value in each package.

install.sh
Zopolis4: Wouldn't that be the same if we just had the value set in PREBUILT_URL? I'm not a fan of single-use variables.

install.sh
```diff
@@ -3,56 +3,81 @@
 # Exit on fail.
 set -e
 
-RESET='\e[0m'
+# Default chromebrew repo values.
+: "${GTIHUB_OWNER:=chromebrew}"
```
uberhacker, line 7: `GITHUB_OWNER`?

install.sh
Zopolis4: (once this is rebased onto master, this review is no longer relevant)

uberhacker:
@supechicken: Did you want to merge master into this PR branch and resolve merge conflicts?

PR #9225, opened 2024-01-28
Add crew upload command

uberhacker:
This moves and refactors the logic in `tools/gl.sh` into crew.  Tested for all architectures.

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -54,6 +54,7 @@ DOC = <<~DOCOPT
     crew sysinfo [options]
     crew update [options] [<compatible>]
     crew upgrade [options] [-k|--keep] [-s|--source] [<name> ...]
+    crew upload [options] [<name> ...]
     crew whatprovides [options] <pattern> ...
 
     -b --include-build-deps  Include build dependencies in output.
@@ -515,6 +516,15 @@ def help(pkgName = nil)
       Use `crew update` to update crew itself.
       If `-s` or `--source` is present, the package(s) will be compiled instead of upgraded via binary.
       If `-v` or `--verbose` is present, extra information will be displayed.
+    EOT
+  when 'upload'
+    puts <<~EOT
+      Upload binaries.
+      Usage: crew upload [<package1> <package2> ...]
+      This will update the binary_sha256 hashes in each package and upload binaries to GitLab.
+      The GITLAB_TOKEN environment variable must be set to access the upstream repository.
+      If no package(s) are provided, all binaries in `#{CREW_LOCAL_REPO_ROOT}/release/<arch>` will be uploaded.
+      If `-v` or `--verbose` is present, additional debug information will be displayed.
     EOT
   when 'whatprovides'
     puts <<~EOT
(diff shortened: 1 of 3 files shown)


bin/crew
```diff
@@ -1847,6 +1857,111 @@ def print_deps_tree(args)
...
+
+  base_url = 'https://gitlab.com/api/v4/projects/26210301/packages/generic'
+
+  packages.strip!
+
+  [packages].each do |package|
+    %w[x86_64 i686 armv7l].each do |arch|
+      puts "Package: #{package}, Arch: #{arch}".yellow
+      pkgfile = "#{CREW_LOCAL_REPO_ROOT}/packages/#{package}.rb"
+      new_tarfile = Dir["#{CREW_LOCAL_REPO_ROOT}/release/#{arch}/#{package}-*-chromeos-#{arch}.{tar.xz,tpxz,tar.zst}"].max_by { |f| File.mtime(f) }
```
Zopolis4, line 1893: We aren't creating .tpxz files anymore.

bin/crew
```diff
@@ -1847,6 +1857,111 @@ def print_deps_tree(args)
...
+      if new_tarfile.nil?
+        puts "#{CREW_LOCAL_REPO_ROOT}/release/#{arch}/#{package}-#-chromeos-#{arch}.(tar.xz|tpxz|tar.zst) not found.\n".lightred
```
Zopolis4, line 1895: Same here.

bin/crew
```diff
@@ -1847,6 +1857,111 @@ def print_deps_tree(args)
...
+      puts "curl -# --header \"DEPLOY-TOKEN: #{gitlab_token}\" --upload-file \"#{new_tarfile}\" \"#{new_url}\" | cat" if @opt_verbose
+      output = `curl -# --header "DEPLOY-TOKEN: #{gitlab_token}" --upload-file "#{new_tarfile}" "#{new_url}" | cat`.chomp
+      unless output.include?('201 Created')
+        puts output.lightred
+        puts "Unable to upload. Skipping binary_sha256 update in #{pkgfile}...".lightred
+        next
+      else
+        puts output.lightgreen
+      end
+      old_sha256 = `grep -m 1 #{arch}: #{pkgfile} 2> /dev/null`.chomp
```
Zopolis4, line 1913: Can't we just query this via get_binary_sha256?

bin/crew
```diff
@@ -1847,6 +1857,111 @@ def print_deps_tree(args)
...
+          puts "Updating binary_sha256 in #{pkgfile}..."
+          puts "from: #{arch}: '#{old_sha256}'"
+          puts "  to: #{arch}: '#{new_sha256}'"
+          puts "sed -i 's/#{old_sha256}/#{new_sha256}/g' #{pkgfile}" if @opt_verbose
+          system "sed -i 's/#{old_sha256}/#{new_sha256}/g' #{pkgfile}"
+        else
+          puts "Skipping binary_sha256 update in #{pkgfile}..."
+        end
+      else
+        unless File.readlines("#{CREW_LOCAL_REPO_ROOT}/packages/#{package}.rb").grep(/binary_sha256/).any?
```
Zopolis4, line 1926: Can't we use is_binary or something similar for this?

Zopolis4 (changes requested):
Just some initial thoughts, mostly focusing on the possibility of increased usage of crew built in functions now that we have that ability.

bin/crew
uberhacker: Doesn't hurt to include just in case.

bin/crew
uberhacker: No, this logic is here to check if the `binary_sha256` section exists and if not, inserts the lines in the package file.

bin/crew
uberhacker: Maybe?

bin/crew
Zopolis4: On the contrary-- if we are creating .tpxz files it's a bug, the installer will break (in its current state) and some of my scripts will break. 

The creation and uploading of new .tpxz files is a bug, and if we catch it at the upload stage it will save us hassle in the future.

bin/crew
Zopolis4: Ah, for packages which are lacking builds entirely.

bin/crew
```diff
@@ -1847,6 +1857,111 @@ def print_deps_tree(args)
...
+    %w[x86_64 i686 armv7l].each do |arch|
+      puts "Package: #{package}, Arch: #{arch}".yellow
+      pkgfile = "#{CREW_LOCAL_REPO_ROOT}/packages/#{package}.rb"
+      new_tarfile = Dir["#{CREW_LOCAL_REPO_ROOT}/release/#{arch}/#{package}-*-chromeos-#{arch}.{tar.xz,tpxz,tar.zst}"].max_by { |f| File.mtime(f) }
+      if new_tarfile.nil?
+        puts "#{CREW_LOCAL_REPO_ROOT}/release/#{arch}/#{package}-#-chromeos-#{arch}.(tar.xz|tpxz|tar.zst) not found.\n".lightred
+        next
+      end
+      puts 'Generating sha256sum ...'
+      new_sha256 = `sha256sum #{new_tarfile} | cut -d' ' -f1`.chomp
```
Zopolis4, line 1899: I believe there is a built in ruby function for this, which we should be using instead of system commands.

Zopolis4:
Just wondering-- does it make more sense to not have logic to create the new binary_sha256 arrays in crew? 

i.e. leave it to be manually created on the rare occasion an existing package does not have any binaries and needs them.

I say this simply because the logic is not particularly clean (no way around that), so might be better to leave out if we don't use that often.

Just wondering, if this has already been thought about then don't worry about it.

Zopolis4:
Is there logic for updating the `binary_compression` value? i.e. when a package that was previously built as anything other than `.tar.zst` and doesn't have the `no_zstd` value, which will need a new `binary_compression` value of `.tar.zst`

I wrote some code (in fish) which does this, in case you need an example. (obviously when rewritten in ruby it could be a lot cleaner and doesn't need to use `grep`, and so on)
```fish
function update_binary_compression
    # If the package has the no_zstd property set, it will be compressed as .tar.xz
    # If the property is not set, it will be compressed as .tar.zst
    if grep -q 'no_zstd' ../packages/$argv.rb
        sed -i "s/binary_compression.*/binary_compression 'tar.xz'/" ../packages/$argv.rb
    else
        sed -i "s/binary_compression.*/binary_compression 'tar.zst'/" ../packages/$argv.rb
    end
end
```

bin/crew
uberhacker: I would argue it's not a bug.  What is a bug (or lacking feature, rather) is not being flexible and allowing and coding for ANY compression format.  IOW, we shouldn't be setting limits.  In the future, there could be a better compression format that may be preferable.

Zopolis4: And in the future, we can switch to that. Flexibility may be easy to account for here, in ruby, but in the installer and in helper scripts it can make things much more complex.

essentially, imo we shouldn't be checking for binaries compressed in a format we deprecated

satmandu:
Robocop failing?

bin/crew
uberhacker: I wasn't able to find a command.  Feel free to suggest something else.

satmandu:
Most of these rubucop errors can be corrected automatically.

`rubocop -c .Robocop.yml -A bin/crew` I think is the syntax?

Zopolis4:
this should probably work:

rubocop -A -c .rubocop.yml bin/crew

bin/crew
Zopolis4: We use Digest::SHA256.hexdigest(File.read('<path>')) in a number of places around the codebase, so probably that.

uberhacker: That won't work.  We need to look for the hash of a specific architecture.

bin/crew
uberhacker: I couldn't get this to work with anything else.

Zopolis4:
Should we choose between either taking a list of packages as an argument or checking the `release` directory? Given that only a few people will be using this, we can just decide which one to use and not have code for two ways of doing things.

uberhacker:
> Should we choose between either taking a list of packages as an argument or checking the `release` directory? Given that only a few people will be using this, we can just decide which one to use and not have code for two ways of doing things.

Absolutely need the ability to pass arguments.  Why is it a problem to have added flexibility?

satmandu (approved):

PR #9226, opened 2024-01-29
Fix repology workflow

Zopolis4:
Hopefully this works-- we can't use `crew update` because satmandu's containers are root by default, but my containers have a broken sudo right now, so we can't use those either, and we have to use a container until #9218 is merged (and maybe even after that), so this is my best option.

### DIFF
--- a/.github/workflows/Repology.yml
+++ b/.github/workflows/Repology.yml
@@ -11,6 +11,7 @@ jobs:
       - name: Generate JSON
         run: |
           git config --global safe.directory /usr/local/lib/crew
+          git pull
           git config --global user.name "github-actions[bot]"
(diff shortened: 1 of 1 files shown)


.github/workflows/Repology.yml
```diff
@@ -11,6 +11,8 @@ jobs:
       - name: Generate JSON
         run: |
           git config --global safe.directory /usr/local/lib/crew
+          git fetch
+          git pull
```
uberhacker, line 15: These commands are redundant.  `git pull` already does a fetch.  See https://git-scm.com/docs/git-pull.

uberhacker (approved):

PR #9134, opened 2023-12-28
Rust 1.74.1 => 1.75.0

uberhacker:
Tested on all architectures.

PR #9227, opened 2024-01-29
Fix unnecessary warning about packages with changed versions but not updated binaries

Zopolis4:
While it would prevent issues like what happened in #9134, overall the legitimate occurences of changing a version without new binaries, namely removing unnecessary revision tags, outweight the bugs it catches.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=beautifulbasilisk crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -673,7 +673,6 @@ def update
 
     if differentVersion && !differentSha && hasSha
       canBeUpdated -= 1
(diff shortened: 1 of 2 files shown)


uberhacker:
My vote is to  keep this warning.  IOW, I feel it is necessary.

PR #9233, opened 2024-01-31
Fix broken gnu_time and linuxheaders packages on i686

uberhacker:
The binaries had the incorrect compression extension.

### DIFF
--- a/packages/gnu_time.rb
+++ b/packages/gnu_time.rb
@@ -11,12 +11,17 @@ class Gnu_time < Package
   compatibility 'all'
   source_url 'https://git.savannah.gnu.org/git/time.git'
   git_hashtag "v#{version}"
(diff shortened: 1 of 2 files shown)


Zopolis4 (approved):
obviously the fix is good, but it does feel a little ugly to have code like this

it'll get fixed by the next update/rebuild, though, so not too big of a deal

uberhacker:
> obviously the fix is good, but it does feel a little ugly to have code like this
> 
> it'll get fixed by the next update/rebuild, though, so not too big of a deal

The real problem is the inconsistent compression schemes.  All binaries should be the same for every package.

Zopolis4:
I agree-- not sure what produced these inconsistent results back when these binaries were being compiled, but at least the current build setup won't let that happen.

PR #8362, opened 2023-06-10
Add armv7l install on aarch64 userspace systems

satmandu:
Fixes #8326 #8329 
- Refactors glibc installs for new installs to reduce disk space used on M113 and later systems, and also moves glibc to versioned packages.
- Includes `armv7l` glibc files for ARM machines with `aarch64` userspace. (This creates an `armv7l` Chromebrew install on such systems, which is a stopgap until we get `aarch64` bootstrapped on these systems.)

**(There is NO upgrade path currently setup to go from this to a future aarch64 Chromebrew without a clean reinstall of Chromebrew!)**
- Also adds minor fix for `git` postinstall...
- Also adds the crew update from https://github.com/chromebrew/chromebrew/pull/8377 to allow for upgrades of is_fake packages properly.

- Please try this on any fresh install. It should now work on any system. (But note that this install hard codes the repo to my own repo... is there a good way to change that back?)
```
cd /usr/local
curl -Ls https://github.com/satmandu/chromebrew/raw/glibc_lib/install.sh | \
OWNER=satmandu REPO=chromebrew BRANCH=glibc_lib bash
```

On an aarch64 userspace container built off of M113, these are the steps I used after entering the container as root:
```
passwd -d chronos
echo "chronos ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
cd /usr/local
chown chronos:chronos .
chown chronos:chronos tmp
sudo -u chronos -s
curl -Ls https://github.com/satmandu/chromebrew/raw/glibc_lib/install.sh | \
OWNER=satmandu REPO=chromebrew BRANCH=glibc_lib bash
```
- _**Note that the postinstall steps NEED to be followed after the install is complete.**_

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
- [x] `aarch64` (as `armv7l`)

### Run the following to get this pull request's changes locally for testing.
```
CREW_TESTING_REPO=https://github.com/satmandu/chromebrew.git CREW_TESTING_BRANCH=glibc_lib CREW_TESTING=1 crew update
```

PR #9241, opened 2024-02-01
Rework installer, change USER_SPACE_ARCH

Zopolis4:
This fixes, among other things, installs on `aarch64` userspaces.

For the rework to the installer, I just worked in using the `binary_compression` value, then noticed that we could unify some code.

Regarding the  `USER_SPACE_ARCH` removal: 
1. If `USER_SPACE_ARCH` is `aarch64`, a bunch of things break. 
2. At any rate, we treat `aarch64` userspaces like `armv7l` ones, so there's no need to make a distinction.
3. Once we properly support `aarch64` userspaces, there won't be a need for special handling.

This will require a rebase from #9223, but apart from that most of the changes there still apply.

Tested and working on all architectures, except `i686`, where the missing curl binary breaks **all** installs, not just this one. The install on `i686` does proceed flawlessly up until that point, so it'll work once the curl issue is fixed.

```
exec bash --init-file <(curl -Ls https://raw.githubusercontent.com/Zopolis4/chromebrew/quifix/install.sh)
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -93,8 +93,8 @@ CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
 # uname -m reports armv8l.
 ARCH="${ARCH/armv8l/armv7l}"
 
-# Determine if there is a difference between kernel and user space.
-if [[ "${ARCH}" == "aarch64" && ! -d '/lib64' ]]; then
+# This does not represent the actual userspace architecture (if there is an aarch64 userspace), but we behave as if it does
+if [[ "${ARCH}" == "aarch64" ]]; then
   USER_SPACE_ARCH='armv7l'
 else
   USER_SPACE_ARCH="${ARCH}"
@@ -163,10 +163,6 @@ find "${CREW_LIB_PATH}" -mindepth 1 -delete
 # Download the chromebrew repository.
 curl -L --progress-bar https://github.com/"${OWNER}"/"${REPO}"/tarball/"${BRANCH}" | tar -xz --strip-components=1 -C "${CREW_LIB_PATH}"
 
-# Prepare urls and sha256s variables.
-urls=()
-sha256s=()
-
 BOOTSTRAP_PACKAGES="zstd crew_mvdir ruby git ca_certificates openssl"
 
 # Older i686 systems.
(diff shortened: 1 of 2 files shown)


uberhacker:
I'm not sure I'm in agreement with these changes.  The whole point of `USER_SPACE_ARCH` is to set to armv7l if the architecture is aarch64.  We don't want to change the architecture from aarch64.  We just need to show there is a distinction between the architecture and user space.  Rather than remove, why not make sure `USER_SPACE_ARCH` is always set to armv7l which is what it was indented to do in the first place.  The changes would be less dramatic than what this PR is doing.

Zopolis4:
But there's absolutely no need for `USER_SPACE_ARCH` to exist. There's no problem with setting `ARCH` to `armv7l` on `aarch64`, because we treat `aarch64` and `armv7l` identically. We don't need a second variable for that.

satmandu:
I think the simplest reason not to remove `USER_SPACE_ARCH` is that in adding support for aarch64 systems, we will need to handle machines that have an aarch64 kernel and aarch64 userspace, but where we have also installed armv7l chromebrew... and we're going to want to identify such systems... so being able to note where `ARCH` != `USER_SPACE_ARCH` will probably be useful.

The current issue with bootstrapping aarch64 is that we haven't had packages for glibc and GCC for the aarch64 systems... but now that gentoo has binary packages, maybe we can use those to get that working...

Zopolis4:
> we will need to handle machines that have an aarch64 kernel and aarch64 userspace, but where we have also installed armv7l chromebrew

I thought that there was no supported upgrade path from an `armv7l` installation to an `aarch64` one? (At least, thats what the description of #8362 said)

At any rate, if we do need to make that distinction, we could just do something like this:
(I presume this will be in `lib/fixup.rb`)
```ruby
if `uname -m`.chomp == 'aarch64'
  puts "You are running an `armv7l` installation of chromebrew on an `aarch64` kernel.\n"
  puts "Please run "exec bash --init-file <(curl -Ls git.io/vddgY)" to reinstall chromebrew.\n"
  puts "You may want to copy any files you wish to keep out of #{CREW_PREFIX}, as the installation process will delete them.\n"
  puts "You may also want to take a list of the installed packages via "crew list --installed" and save this to install later."
end
```

Zopolis4:
As per the discussion on slack, I'm dropping the `USER_SPACE_ARCH` changes for the sake of getting the issue fixed. 

I still think this is the wrong way to approach this, we have a variable which doesn't need to exist, and that misrepresents the actual architecture. 

But, that discussion shouldn't hold up fixing installations for all users with `aarch64` userspaces.

bin/crew
```diff
@@ -582,7 +582,7 @@ end
 
 def files(pkgName)
   local_filelist    = File.join(CREW_META_PATH, "#{pkgName}.filelist")
-  manifest_filelist = File.join(CREW_LIB_PATH, "manifest/#{USER_SPACE_ARCH}/#{pkgName[0]}/#{pkgName}.filelist")
+  manifest_filelist = File.join(CREW_LIB_PATH, "manifest/#{ARCH}/#{pkgName[0]}/#{pkgName}.filelist")
```
uberhacker, line 585: Change back to `USER_SPACE_ARCH`?

uberhacker (approved):
This looks good.

PR #9244, opened 2024-02-02
Fix crew upload to work with all gitlab tokens

Zopolis4:
As per the discussion on slack, this should fix `crew upload` for all gitlab tokens created relatively recently (~2 months ago, maximum upper bound of 6) while still allowing old tokens to work.


uberhacker (approved):
New tokens are longer and the older tokens expire yearly anyway so this makes sense.

PR #9248, opened 2024-02-02
Add snd package

uberhacker:
Snd is a sound editor.  See https://ccrma.stanford.edu/software/snd/.


packages/snd.rb
```diff
@@ -0,0 +1,47 @@
+require 'buildsystems/autotools'
+
+class Snd < Autotools
+  description 'Sound editor'
+  homepage 'https://ccrma.stanford.edu/software/snd/'
+  version '24.1'
+  license 'Copyright 1996-2016 Bill Schottstaedt'
```
Zopolis4, line 7: `license 'TCL/TK`? 

https://spdx.org/licenses/TCL.html

uberhacker: I couldn't find a license but I did find the copyright notice in COPYING from the source.

Zopolis4: Yes-- that COPYING file is the `TCL` license. I looked up the actual text of the file, beyond the copyright notice, and got this. See the link I sent.

uberhacker: Why should we remove the author's copyright notice?  If it was my code, I'd want the copyright notice.

Zopolis4: Because this is the same practice we follow for every package? The `license` field is for what license the package has. This package has the `TCL` license.

uberhacker: I'm not a lawyer but this is taking away credit from the author.

Zopolis4: I think there's a misunderstanding here. That COPYING file is equivalent to a LICENSE file. Most of the LICENSE files we use also have the copyright of the author embedded in them (for example, the MIT license). We are treating this package the same way we would treat any other.

satmandu: Why is SPDX relevant? As per https://sourceforge.net/projects/snd/ they are using the `Artistic License`...

Zopolis4: Where did you see that? I was going off of the `COPYING` file which had the same text as the TCL License (so I linked to SPDX to show that text)

uberhacker: Did you click on the link?  The sourceforge repo is reporting `Artistic License`.

Zopolis4: Thank you-- I wasn't sure where exactly it was, I'm not familiar with sourceforge.

satmandu (approved):

PR #9250, opened 2024-02-03
pipewire => 1.0.3

satmandu:
Builds properly:
- [x] `x86_64`
- [x] `armv7l`


Zopolis4 (approved):
maybe the package could be simplified more depending on how upstream has changed, but that can be a future PR, no need to hold this update back

PR #9270, opened 2024-02-06
Add Ruby Getrealdeps

satmandu:
- Converted getrealdeps to ruby.
- Have crew run it when packaging.
- Fixup template package to note that the binary section is needed during builds, even as a dummy section, as otherwise the dependency checking tries to build again because crew doesn't try to find local binaries from the build.
- Fix upload handling of no binary_compression section in package files.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=getrealdeps crew update
```


tools/getrealdeps.rb
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env ruby
+# getrealdeps for Chromebrew version 1.2
+# Author: Satadru Pramanik (satmandu) satadru at gmail dot com
+require 'fileutils'
```
uberhacker, line 4: Do we need to include the other libraries for color, etc.?

satmandu: no color needed.

tools/getrealdeps.rb
```diff
@@ -0,0 +1,116 @@
...
+unless File.exist?("#{CREW_PREFIX}/etc/crew/meta/#{pkg}.filelist")
+  echo_error "Package #{pkg} either does not exist or does not contain any libraries."
+  exit 1
+end
+
+# Speed up grep.
+ENV['LC_ALL'] = 'C'
+
+# Install grep if a functional local copy does not exist.
+if system('grep --version > /dev/null 2>&1')
```
uberhacker, line 34: Does `system` have a return value?

satmandu: Yes.

tools/getrealdeps.rb
```diff
@@ -0,0 +1,116 @@
...
+  GREP = 'grep'
+else
+  system('crew install grep')
+  GREP = "#{CREW_PREFIX}/bin/grep"
+end
+
+# Gawk is needed for adding dependencies.
+unless system('gawk -W version > /dev/null 2>&1')
+  puts "\nGawk is needed for adding dependencies to package files."
+  system('crew install gawk') unless system('gawk -W version > /dev/null 2>&1')
```
uberhacker, line 44: `unless ...` not needed?

satmandu: oops!

satmandu: fixed

tools/getrealdeps.rb
satmandu: Good call. I adjusted that to `puts` to make things simpler.

tools/getrealdeps.rb
```diff
@@ -0,0 +1,116 @@
...
+pkgfiles = crewfilesfxn(pkg).split("\n").uniq
+
+FileUtils.rm_rf("/tmp/deps/#{pkg}")
+# Remove files we don't care about, such as man files and non-binaries.
+pkgfiles = pkgfiles.reject { |i| !File.file?(i.chomp) || !File.read(i.chomp, 4) == "\x7FELF" || i.include?('.zst') }
+
+# Use readelf to determine library dependencies, as
+# this doesn't almost run a program like using ldd would.
+pkgdepsfiles = pkgfiles.map do |i|
+  system("upx -d #{i} > /dev/null 2>&1") if File.exist?('/.dockerenv')
```
uberhacker, line 72: Should there be a check to see if `upx` is installed first?

tools/getrealdeps.rb
uberhacker: I see you removed `echo_error` so this should no longer be an issue.

tools/getrealdeps.rb
satmandu: That makes sense.

satmandu: Fixed

uberhacker (dismissed):
Add `crew getrealdeps` command and include in README.md?

satmandu:
Readme updated.

tools/getrealdeps.rb
```diff
@@ -0,0 +1,121 @@
...
+if ARGV[0].nil? || ARGV[0].empty? || ARGV[0].include?('#')
+  puts 'Getrealdeps checks for the dependencies of a package.'
+  puts 'Dependencies are added if the package file is missing them.'
+  puts 'Usage: getrealdeps.rb <packagename>'
+  exit 1
+end
+
+pkg = ARGV[0].chomp('.rb')
+
+CREW_PREFIX = ENV['CREW_PREFIX'] || `crew const CREW_PREFIX`.split('=')[1].chomp
```
Zopolis4, line 16: Would it not make more sense to include lib/const and draw the variables from there?

tools/getrealdeps.rb
```diff
@@ -0,0 +1,121 @@
...
+end
+
+# Gawk is needed for adding dependencies.
+unless system('gawk -W version > /dev/null 2>&1')
+  puts "\nGawk is needed for adding dependencies to package files."
+  system('crew install gawk')
+end
+
+# upx is needed to expand compressed binaries to check for dependencies.
+unless system('upx --version > /dev/null 2>&1')
```
Zopolis4, line 48: If we've gotten to this stage in the build process and we have compressed binaries, we must have used upx to compress them right before, so we're guaranteed to have it

Zopolis4 (dismissed):

tools/getrealdeps.rb
satmandu: This can be run independently of the build though...

tools/getrealdeps.rb
satmandu: This can be run independently of the build. We don't need all the `const.rb` infrastructure...

bin/crew
```diff
@@ -43,6 +43,7 @@ DOC = <<~DOCOPT
     crew deps [options] [-t|--tree] [-b|--include-build-deps] [--exclude-buildessential] <name> ...
     crew download [options] [-s|--source] <name> ...
     crew files [options] <name> ...
+    crew getrealdeps <name> ...
```
uberhacker, line 46: Would it make sense to integrate getrealdeps into the deps command?  I'm thinking having 2 commands might be a bit confusing.  At least for me. :blush:

satmandu: Deps is much quicker, because it doesn't need to extract and decompress everything... But that's also the downside...

uberhacker: I was thinking we could add a switch.  Something like `-d` or `--deep` for a deep dive into dependencies.  It would do the same thing without the need to have a separate seemingly related command.

bin/crew
```diff
@@ -1711,6 +1718,8 @@ def archive_package(crew_archive_dest)
     end
   end
   system "sha256sum #{crew_archive_dest}/#{pkg_name} > #{crew_archive_dest}/#{pkg_name}.sha256"
+  # Make sure the package file has runtime dependencies added properly.
+  system "#{CREW_LIB_PATH}/tools/getrealdeps.rb #{@pkgName}"
```
supechicken, line 1722: Use `require` for ruby code?

tools/getrealdeps.rb
uberhacker: We could implement the 3 or more rule.  If you have to use this method 3 or more times, include const.rb?

tools/packages.yaml
```diff
@@ -5346,16 +5346,6 @@ url: https://gitlab.com/cespedes/ltrace
...
-name: lttng_tools
-url: https://lttng.org/download/#build-from-source
-activity: medium
----
-kind: url
-name: lttng_ust
-url: https://lttng.org/download/#build-from-source
-activity: medium
----
-kind: url
```
uberhacker, line 5358: Merge master into this PR branch?

satmandu: ugh... it is broken...

bin/crew
satmandu: Do you have a good way of making this happen while also allowing it to run standalone?

bin/crew
satmandu: I have added a `crew deps --deep` option.

Example: `crew deps --deep mpv`

tools/getrealdeps.rb
satmandu: We check for upx just in case, but if run standalone, buildessential may not be installed.

bin/crew
satmandu: Let's hold off on this for now, as this will require significant refactoring...

bin/crew
```diff
@@ -40,9 +40,10 @@ DOC = <<~DOCOPT
   Usage:
     crew build [options] [-k|--keep] <name> ...
     crew const [options] [<name> ...]
-    crew deps [options] [-t|--tree] [-b|--include-build-deps] [--exclude-buildessential] <name> ...
+    crew deps [options] [--deep] [-t|--tree] [-b|--include-build-deps] [--exclude-buildessential] <name> ...
```
uberhacker, line 43: Add `-d|--deep` short option too?

bin/crew
uberhacker: Do we need `crew getrealdeps` if we have `crew deps --deep`?

bin/crew
satmandu: There is already a `-d` option for crew, so I don't want to create a conflict.

bin/crew
satmandu: Removed.

uberhacker (approved):

PR #9276, opened 2024-02-07
Git_prompt 2.32.0 => 2.43.0

uberhacker:
We should probably get in the habit of updating this along with git.


satmandu:
Feel free to add code to remind us like we have in zlibpkg/minizip?

satmandu (approved):

PR #9280, opened 2024-02-08
fix aarch64 installs, Remove USER_SPACE_ARCH

Zopolis4:
This fixes, among other things, installs on aarch64 userspaces.

I am removing `USER_SPACE_ARCH` because there is no need to make a distinction between kernel and userspace. 

We treat `aarch64` userspaces as if they were `armv7l`, and if we do not do this, things (like installs!) will break.

This PR fixes the issue by ensuring we always treat `aarch64` userspaces as `armv7l` userspaces.

Yes, this does mean that the value of `ARCH` does not always correspond to the actual architecture. Instead, it corresponds to the architecture we treat it as. Currently, the value of USER_SPACE_ARCH` does not always correspond to the actual userspace architecture. There is no way to make `armv7l-on-aarch64` installs work without at some point, lying about the architecure. 

Yes, an alternative method to fixing this issue would be to find out where specifically telling the truth about the current architecture breaks things, and add another special-case exception for that. I don't see any value in doing that, rather than just having one value for architecrure and have it represent the architecture we are treating it as.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -89,15 +89,12 @@ CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
 # For container usage, where we want to specify i686 arch
 # on a x86_64 host by setting ARCH=i686.
 : "${ARCH:=$(uname -m)}"
-# For container usage, when we are emulating armv7l via linux32
-# uname -m reports armv8l.
-ARCH="${ARCH/armv8l/armv7l}"
 
-# This does not represent the actual userspace architecture (if there is an aarch64 userspace), but we behave as if it does
-if [[ "${ARCH}" == "aarch64" ]]; then
-  USER_SPACE_ARCH='armv7l'
-else
-  USER_SPACE_ARCH="${ARCH}"
+# For container usage, when we are emulating armv7l via linux32, where uname -m will report armv8l.
+# Additionally, if the architecture is aarch64, set it to armv7l, as we treat as if it was armv7l.
+# When we have proper support for aarch64, remove this.
+if [[ "${ARCH}" = "armv8l" ]] || [[ "${ARCH}" = "aarch64" ]]; then
+  ARCH='armv7l'
 fi
 
 if [[ "$ARCH" == "x86_64" ]]; then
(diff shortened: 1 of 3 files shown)


lib/const.rb
```diff
@@ -37,12 +37,14 @@
...
-# This helps with virtualized builds on aarch64 machines
-# which report armv8l when linux32 is run.
-ARCH = KERN_ARCH.eql?('armv8l') ? 'armv7l' : KERN_ARCH
-
-# This doesn't actually represent the userspace architecture (if the userspace architecture is aarch64), but we behave as if it does
-USER_SPACE_ARCH = ARCH.eql?('aarch64') ? 'armv7l' : ARCH
+# This helps with virtualized builds on aarch64 machines which report armv8l when linux32 is run.
+# We also report aarch64 machines as armv7l for now, as we treat them as if they were armv7l.
+# When we have proper aarch64 support, remove this.
+ARCH = if KERN_ARCH == 'armv8l' || KERN_ARCH == 'aarch64'
```
satmandu, line 43: I like this here...

uberhacker: `ARCH = %w[aarch64 armv8l].include?(KERN_ARCH)`?

Zopolis4: There's no amazing solution for this, but it is nicer to have it in one line.

install.sh
```diff
@@ -89,15 +89,12 @@ CREW_DEST_DIR="${CREW_BREW_DIR}/dest"
...
-# This does not represent the actual userspace architecture (if there is an aarch64 userspace), but we behave as if it does
-if [[ "${ARCH}" == "aarch64" ]]; then
-  USER_SPACE_ARCH='armv7l'
-else
-  USER_SPACE_ARCH="${ARCH}"
+# For container usage, when we are emulating armv7l via linux32, where uname -m will report armv8l.
+# Additionally, if the architecture is aarch64, set it to armv7l, as we treat as if it was armv7l.
+# When we have proper support for aarch64, remove this.
+if [[ "${ARCH}" = "armv8l" ]] || [[ "${ARCH}" = "aarch64" ]]; then
+  ARCH='aarch64'
```
uberhacker, line 97: `ARCH='armv7l'`?

Zopolis4: Oops :sweat_smile:

PR #9295, opened 2024-02-09
Getrealdeps Process Refactor

satmandu:
- only install package after install if `-f` option is added.
- Do not do an install as part of using `getrealdeps.rb` after a build completes.
- Add `--use_crew_dest_dir` option to `getrealdeps.rb` and use it from inside `crew`.
- Added `gawk` to `buildessential` since `getrealdeps.rb` uses it.
- Tested by attempting rebuild if `xxd_standalone` (Which worked.)

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=getrealdeps2 crew update
```


packages/core.rb
```diff
@@ -52,10 +52,10 @@ class Core < Package
   depends_on 'libyaml'
   depends_on 'lz4'
   depends_on 'lzip'
+  depends_on 'mawk'
   depends_on 'm4'
   depends_on 'mandb'
   depends_on 'manpages'
-  depends_on 'mawk'
```
uberhacker, line 58: Keep in alphabetical order?

tools/getrealdeps.rb
```diff
@@ -1,30 +1,39 @@
 #!/usr/bin/env ruby
-# getrealdeps for Chromebrew version 1.2
+# getrealdeps for Chromebrew version 1.3
 # Author: Satadru Pramanik (satmandu) satadru at gmail dot com
 require 'fileutils'
 
+if ARGV.include?('--use-crew-dest-dir')
+  ARGV.delete('--use-crew-dest-dir')
```
uberhacker, line 7: Why does the argument need to be deleted?

tools/getrealdeps.rb
```diff
@@ -1,30 +1,39 @@
...
+  puts 'Usage: getrealdeps.rb [--use_crew_dest_dir] <packagename>'
   exit 1
 end
 
 pkg = ARGV[0].chomp('.rb')
+puts "Checking for the runtime dependencies of #{pkg}..."
 
 CREW_PREFIX = ENV['CREW_PREFIX'] || `crew const CREW_PREFIX`.split('=')[1].chomp
+CREW_DEST_DIR = ENV['CREW_DEST_DIR'] || `crew const CREW_DEST_DIR`.split('=')[1].chomp
 CREW_LIB_PREFIX = ENV['CREW_LIB_PREFIX'] || `crew const CREW_LIB_PREFIX`.split('=')[1].chomp
```
uberhacker, line 24: Doesn't pass the 3 or more test to justify `require 'lib/const'`?

packages/core.rb
satmandu: Fixed!

tools/getrealdeps.rb
satmandu: plumbing for allowing getrealdeps to handle multiple package arguments. I just didn't implement a for loop to do that.

tools/getrealdeps.rb
satmandu: We should make it a greater than 3 rule... because `require 'lib/const'` needs the path to be right, and we might be anywhere in a cloned git tree.

tools/getrealdeps.rb
```diff
@@ -1,30 +1,39 @@
...
-  exit 1
+if @opt_use_crew_dest_dir
+  define_singleton_method('pkgfilelist') {File.join(CREW_DEST_DIR, 'filelist')}
+  abort('Pkg was not built.') unless File.exist?(pkgfilelist)
+else
+  define_singleton_method('pkgfilelist') {"#{CREW_PREFIX}/etc/crew/meta/#{pkg}.filelist"}
+  # Package needs to be installed for package filelist to be populated.
+  unless File.exist?(pkgfilelist)
+    puts "Installing #{pkg} because it is not installed."
+    system("yes | crew install #{pkg}")
```
uberhacker, line 34: Check to see if local binary files exist and the version is updated prior to install?

tools/getrealdeps.rb
uberhacker: Why not?  That would be the elegant way to do it, right?

tools/getrealdeps.rb
uberhacker: Doesn't `CREW_LOCAL_REPO_ROOT` account for that?
```
require CREW_LOCAL_REPO_ROOT.to_s + '/lib/const'
```
?

uberhacker: Actually, now that I thought about it a bit, `CREW_LOCAL_REPO_ROOT` is loaded from const.rb. :blush:

tools/getrealdeps.rb
satmandu: That code is for the case of `getrealdeps.rb` being run standalone. It isn't used in a build any more, since builds are shunted to `@opt_use_crew_dest_dir`

tools/getrealdeps.rb
satmandu: Done.

tools/getrealdeps.rb
satmandu: Because that will require refactoring to wrap everything in a function, which I haven't done yet.

But also, removing the argument allows argv[0] to be the package that is looked at.

satmandu: Wrapped!

satmandu:
Assistance with avoiding this issue would be appreciated:
```
tools/getrealdeps.rb:63:3: W: Lint/NestedMethodDefinition: Method definitions must not be nested. Use lambda instead.
  def whatprovidesfxn(pkgdepslcl, pkg) ...
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

tools/getrealdeps.rb
```diff
@@ -1,121 +1,141 @@
...
-  GREP = 'grep'
+@crew_local_repo_root = `git rev-parse --show-toplevel 2> /dev/null`.chomp
+# When invoked from crew, pwd is CREW_DEST_DIR, so @crew_local_repo_root
+# is empty.
+if @crew_local_repo_root.to_s.empty?
+  require_relative '../lib/const'
 else
-  system('crew install grep')
-  GREP = "#{CREW_PREFIX}/bin/grep"
+  require File.join(@crew_local_repo_root, 'lib/const')
```
uberhacker, line 12: `/lib/const`?

satmandu: File.join handles this just fine adding necessary slashes.

satmandu: https://ruby-doc.org/3.2.2/File.html#method-c-join

uberhacker: These functions try to be too smart.  The more you know.

tools/getrealdeps.rb
uberhacker: For a future PR then?

tools/getrealdeps.rb
satmandu: It's great because it adds a slash without adding an extra slash if one of the strings ends in a slash!

tools/getrealdeps.rb
```diff
@@ -1,121 +1,141 @@
...
+  # Look at files in CREW_DEST_DIR instead of assuming the package is
+  # normally installed, which lets us avoid installing the package if it
+  # was just built.
+  pkgfiles.map! {|item| item.prepend(CREW_DEST_DIR)} if @opt_use_crew_dest_dir
 
-# Look for missing runtime dependencies.
-missingpkgdeps = pkgdeps.reject { |i| File.read("#{CREW_PREFIX}/lib/crew/packages/#{pkg}.rb").include?("depends_on '#{i}'") unless File.read("#{CREW_PREFIX}/lib/crew/packages/#{pkg}.rb").include?("depends_on '#{i}' => :build") }
+  FileUtils.rm_rf("/tmp/deps/#{pkg}")
+  # Remove files we don't care about, such as man files and non-binaries.
+  pkgfiles = pkgfiles.reject { |i| !File.file?(i.chomp) || !File.read(i.chomp, 4) == "\x7FELF" || i.include?('.zst') }
```
uberhacker, line 88: Should the compression algorithm be abstracted with a variable?

satmandu: The zst files are from man pages... otherwise it would be a gz file... but in any case, readelf will reject the files anyways.

satmandu: So I don't think it makes sense to make it any more complicated. This is a 99.9% solution for the usage of this tool, and the other uses would be for checking existing packages ~ 2 years old or older...

tools/getrealdeps.rb
```diff
@@ -1,121 +1,141 @@
 #!/usr/bin/env ruby
-# getrealdeps for Chromebrew version 1.2
+# getrealdeps for Chromebrew version 1.3
```
uberhacker, line 2: `version 1.43.0`?

tools/getrealdeps.rb
uberhacker: This entire PR is full of complexities. :sweat:

tools/getrealdeps.rb
```diff
@@ -1,121 +1,141 @@
...
-  puts "\n Adding deps: #{adddep}"
-  gawk_cmd = if File.foreach("#{CREW_PREFIX}/lib/crew/packages/#{pkg}.rb").grep(/depends_on/).any?
-    # This files contains dependencies already, so add new deps after existing dependencies.
-               "gawk -i inplace -v dep=\"  depends_on '#{adddep}' # R\" 'FNR==NR{ if (/depends_on/) p=NR; next} 1; FNR==p{ print dep }' #{CREW_PREFIX}/lib/crew/packages/#{pkg}.rb #{CREW_PREFIX}/lib/crew/packages/#{pkg}.rb"
-             else
-    # This files doesn't contain deps, so just add new deps.
-               "gawk -i inplace -v dep=\"  depends_on '#{adddep}' # R\" 'FNR==NR{ if (/})/) p=NR; next} 1; FNR==p{ print \"\\n\" dep }' #{CREW_PREFIX}/lib/crew/packages/#{pkg}.rb #{CREW_PREFIX}/lib/crew/packages/#{pkg}.rb"
-             end
-  system(gawk_cmd)
+ARGV.each do |package|
```
uberhacker, line 139: How does the script handle no arguments?

satmandu: That's handled at the top of the script.
```
 $ tools/getrealdeps.rb
Getrealdeps checks for the runtime dependencies of a package.
The runtime dependencies are added if the package file is missing them.
Usage: getrealdeps.rb [--use_crew_dest_dir] <packagename>
```

tools/getrealdeps.rb
```diff
@@ -1,121 +1,141 @@
...
+  ARGV.delete('--use-crew-dest-dir')
+  @opt_use_crew_dest_dir = true
 end
 
-# upx is needed to expand compressed binaries to check for dependencies.
-unless system('upx --version > /dev/null 2>&1')
-  puts "\nUpx is needed to expand compressed binaries."
-  system('crew install upx')
+# Exit quickly if an invalid package name is given.
+if ARGV[0].nil? || ARGV[0].empty? || ARGV[0].include?('#')
```
uberhacker, line 21: This answers my earlier question about how the script handles no arguments.

uberhacker (approved):

PR #9301, opened 2024-02-10
Fix install part 2: Add no_compile_needed to all ruby_packages

satmandu:
- This address a race condition during installs. `crew` wants to run `getrealdeps.rb` on `ruby_mdl`, which isn't marked as `no_compile_needed`, so it wants to install `upx`, which `getrealdeps.rb` needs... but that install wipes out `CREW_DEST_DIR`, so everything breaks.
- Adding `no_compile_needed` to the `ruby_` packages should fix this.
- The root cause is that each install should use a separate `CREW_DEST_DIR` , but that's an issue for another day.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=getrealdeps3 crew update
```

### DIFF
--- a/packages/ruby_asciidoctor.rb
+++ b/packages/ruby_asciidoctor.rb
@@ -7,4 +7,6 @@ class Ruby_asciidoctor < RUBY
   license 'MIT'
   compatibility 'all'
   source_url 'SKIP'
+
+  no_compile_needed
 end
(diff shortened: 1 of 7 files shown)


Zopolis4 (approved):
makes sense to get urgent fixes in now, and we can deal with what the proper thing is to do later

PR #9302, opened 2024-02-10
crew: Skip SHA256 check for official prebuilt binary

supechicken:
### Why?
When a prebuilt binary is downloaded, the `binary_sha256` section is used to verify the integrity, however:

- `https` can ensure the binary comes from verified sources
- `zstd` archives already include an XXH64 checksum for detecting corruption
- In our case, each binary file will remain unchanged once uploaded

... which does the job well and makes `binary_sha256` redundant.

We can also apply this to `install.sh` for getting rid of those `sed` hacks :)

```
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=disable_sha256 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -790,12 +790,12 @@ def upgrade(*pkgs, build_from_source: false)
 end
 
 def download
-  url = @pkg.get_url(@device[:architecture])
+  url, is_trusted = @pkg.get_url(@device[:architecture])
   source = @pkg.is_source?(@device[:architecture])
 
   uri = URI.parse url
   filename = File.basename(uri.path)
-  sha256sum = @pkg.get_sha256(@device[:architecture])
+  sha256sum = is_trusted ? 'SKIP' : @pkg.get_sha256(@device[:architecture])
   @extract_dir = @pkg.get_extract_dir
 
   @build_cachefile = File.join(CREW_CACHE_DIR, "#{@pkg.name}-#{@pkg.version}-build-#{@device[:architecture]}.tar.zst")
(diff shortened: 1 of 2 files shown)


satmandu:
Doesn't the sha256 help us notice when the binaries have changed? Say a malicious actor steals gitlab credentials and starts replacing binaries. We would have no way of knowing unless the sha256sums on GitHub also started changing...

satmandu:
In short, the sha256 is used primarily to verify that the binary hasn't changed since it was uploaded. Zst itself can provide the integrity check to show that the binary wasn't corrupted during download.

supechicken:
> Doesn't the sha256 help us notice when the binaries have changed? Say a malicious actor steals gitlab credentials and starts replacing binaries. We would have no way of knowing unless the sha256sums on GitHub also started changing...

That's our responsibility... If a random guy could steal our gitlab credentials then he is likely to have our github credentials as well😅

supechicken:
> In short, the sha256 is used primarily to verify that the binary hasn't changed since it was uploaded. Zst itself can provide the integrity check to show that the binary wasn't corrupted during download.

Oh, I forgot that it can also prevent accidental modifications to the file, closing this...

satmandu:
> > Doesn't the sha256 help us notice when the binaries have changed? Say a malicious actor steals gitlab credentials and starts replacing binaries. We would have no way of knowing unless the sha256sums on GitHub also started changing...
> 
> That's our responsibility... If a random guy could steal our gitlab credentials then he is likely to have our github credentials as well😅

Yes this is true, but forcing someone to change the hashes leaves an audit trail which helps track when things have changed.

This is why most distributions still maintain hashes or signatures for files.

I'd also be fine replacing hashes with gpg signatures. 😉

PR #9310, opened 2024-02-11
Remove a number of unnecessary instance variables

Zopolis4:
To be honest, I'm suprised Rubocop doesn't have an automated check for this sort of thing.

At any rate, this PR removes a number of unnecessary instance variables (i.e. those that work perfectly fine as regular variables).

I have removed all of the easy ones (those that could be changed in-place) in `bin/crew`, `tools/getrealdeps.rb` and `lib/musl.rb`. There are some still in `bin/crew` that could be removed with minimal refactoring, but that is for another PR.

There are still very many instance variables that do not need to be instance variables, particularly in packages. I did not remove them, simply because of how much time it would take to do this way. 

I ask that when you update a package, please check if it has any unnecessary instance variables, and if so, remove them.

In addition, I request that when you are creating new code, please try and avoid using unnecessary instance variables.

The only instance variables that we should have, by the end, should be `@pkg`, `@device`, and probably the `@opt_*` variables. The `@*_verbose` variables could probably use a refactor, but that is also a PR for another time.

I'm working on a refactor of `bin/crew`, and unnecessary instance variables pose a pretty big issue to that-- I've seperated out my changes here, so they can be merged with less friction. The instance variables in package files aren't an obstacle to the refactor, but it is good practice to not have variables with unnecessary scope.

All functionality should be identical, I've done some testing-- as I do not have a build caching setup, and a not insignificant chunk of that code is affected, it would make sense for someone to test that.

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -241,17 +241,17 @@ end
 
 def list_installed
   if @opt_verbose
-    @installed_packages = []
+    installed_packages = []
     @device[:installed_packages].each do |package|
       search package[:name], true
-      @installed_packages.append("#{package[:name]} #{package[:version]}")
+      installed_packages.append("#{package[:name]} #{package[:version]}")
     end
-    @sorted_installed_packages = @installed_packages.sort
-    @sorted_installed_packages.unshift('======= =======')
-    @sorted_installed_packages.unshift('Package Version')
-    @first_col_width = @sorted_installed_packages.map(&:split).map(&:first).max_by(&:size).size + 2
-    @sorted_installed_packages.map(&:strip).each do |line|
-      puts "%-#{@first_col_width}s%s".lightgreen % line.split
+    sorted_installed_packages = installed_packages.sort
+    sorted_installed_packages.unshift('======= =======')
+    sorted_installed_packages.unshift('Package Version')
+    first_col_width = sorted_installed_packages.map(&:split).map(&:first).max_by(&:size).size + 2
+    sorted_installed_packages.map(&:strip).each do |line|
+      puts "%-#{first_col_width}s%s".lightgreen % line.split
     end
     print "\n"
   else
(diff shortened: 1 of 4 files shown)


bin/crew
```diff
@@ -241,17 +241,17 @@ end
...
-    @first_col_width = @sorted_installed_packages.map(&:split).map(&:first).max_by(&:size).size + 2
-    @sorted_installed_packages.map(&:strip).each do |line|
-      puts "%-#{@first_col_width}s%s".lightgreen % line.split
+    sorted_installed_packages = installed_packages.sort
+    sorted_installed_packages.unshift('======= =======')
+    sorted_installed_packages.unshift('Package Version')
+    first_col_width = sorted_installed_packages.map(&:split).map(&:first).max_by(&:size).size + 2
+    sorted_installed_packages.map(&:strip).each do |line|
+      puts "%-#{first_col_width}s%s".lightgreen % line.split
     end
```
uberhacker, line 255: You don't need `sorted_installed_packages` with `installed_packages.sort!`.  For example:
```
irb(main):001> lcm = ['larry','curly','moe']
=> ["larry", "curly", "moe"]
irb(main):002> lcm.sort
=> ["curly", "larry", "moe"]
irb(main):004> puts lcm.to_s
["larry", "curly", "moe"]
irb(main):005> lcm.sort!
=> ["curly", "larry", "moe"]
irb(main):006> puts lcm.to_s
["curly", "larry", "moe"]
```

Zopolis4: Good point! I'll put that in the next PR, so that this one stays as exclusively demoting variable scope.

uberhacker (changes requested):
Bump crew version in const.rb?

Zopolis4:
Forgot about that-- fixed.

uberhacker (approved):

satmandu:
Wasn't there an advantage to using instance variables @supechicken ?

supechicken:
I think you are referring https://github.com/chromebrew/chromebrew/pull/4907#discussion_r557326680 😀

Yes, but only in classes and the variable needs to be accessible in all functions within the class.

In most cases, this PR's variables are defined and used only within the same function, and instance variables (`@var`) have the same effect as global variables (`$var`). As @Zopolis4 mentioned, limiting the scope is a good practice and abusing instance variables might bring confusion to other maintainers.

PR #9311, opened 2024-02-11
Miscellaneous crew changes

Zopolis4:
Just a bunch of changes from my large refactor that I'm splitting out into this PR.

The changes are split up into seperate commits for readability and clarity of rationale.

Functionality should be identical, I've tested what I can think of.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=crewcleanup crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -161,8 +161,7 @@ end
 
 def load_json
   # load_json(): (re)load device.json
-  json_path = File.join(CREW_CONFIG_PATH, 'device.json')
-  @device   = JSON.load_file(json_path, symbolize_names: true)
+  @device = JSON.load_file(File.join(CREW_CONFIG_PATH, 'device.json'), symbolize_names: true)
 
   # symbolize also values
   @device.transform_values! {|val| val.is_a?(String) ? val.to_sym : val }
@@ -1604,7 +1603,7 @@ def install
   # If this fails, the install should fail before we create any
   # damage, and we should roughly be at maximal disk space usage at this
   # point anyways.
-  FileUtils.cp File.join(CREW_CONFIG_PATH, 'device.json'), "#{CREW_CONFIG_PATH}/device.json.tmp"
+  FileUtils.cp File.join(CREW_CONFIG_PATH, 'device.json'), File.join(CREW_CONFIG_PATH, 'device.json.tmp')
 
   # remove it just before the file copy
   if @pkg.in_upgrade
@@ -1627,13 +1626,13 @@ def install
 
   # add to installed packages
   @device[:installed_packages].push(name: @pkg.name, version: @pkg.version, binary_sha256: @pkg.get_binary_sha256(@device[:architecture]))
-  File.open("#{CREW_CONFIG_PATH}/device.json.tmp", 'w') do |file|
+  File.open(File.join(CREW_CONFIG_PATH, 'device.json.tmp'), 'w') do |file|
     output = JSON.parse @device.to_json
     file.write JSON.pretty_generate(output)
   end
   # Copy over original if the write to the tmp file succeeds.
-  FileUtils.cp "#{CREW_CONFIG_PATH}/device.json.tmp", File.join(CREW_CONFIG_PATH, 'device.json')
-  FileUtils.rm "#{CREW_CONFIG_PATH}/device.json.tmp"
+  FileUtils.cp File.join(CREW_CONFIG_PATH, 'device.json.tmp'), File.join(CREW_CONFIG_PATH, 'device.json')
+  FileUtils.rm File.join(CREW_CONFIG_PATH, 'device.json.tmp')
 end
 
 def resolve_dependencies_and_build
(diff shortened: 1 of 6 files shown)


bin/crew
```diff
@@ -1403,12 +1350,12 @@ def install_files(src, dst = File.join( CREW_PREFIX, src.delete_prefix('./usr/lo
         # Check for ACLs support.
         rsync_version = `rsync --version`.chomp
         if rsync_version.include?('ACLs') && !rsync_version.include?('no ACLs')
-          system 'rsync', "-ah#{@verbose}HAXW", '--remove-source-files', src, dst, exception: true
+          system 'rsync', "#{@verbose} -ahHAXW", '--remove-source-files', src, dst, exception: true
```
supechicken, line 1353: This might not work as `"#{@verbose} -ahHAXW"` is treated as a single parameter, equivalent of `rsync "-v -ahHAXW" --remove-source...` in shell. The `@verbose` needs to be a standalone argument passed to `system`

lib/package.rb
```diff
@@ -197,38 +191,30 @@ def self.depends_on(dependency, ver_range = nil)
...
-  def self.get_sha256(architecture)
-    if !@build_from_source && @binary_sha256 && @binary_sha256.key?(architecture)
-      return @binary_sha256[architecture]
-    elsif @source_sha256.respond_to?(:has_key?)
-      return @source_sha256.key?(architecture) ? @source_sha256[architecture] : nil
+  def self.get_sha256
+    if is_binary?
+      return @binary_sha256[ARCH.to_sym]
+    elsif @source_sha256.is_a?(Hash)
+      return @source_sha256&.dig(ARCH.to_sym)
```
supechicken, line 208: Since we know it is a hash, we can assume it is not `nil` now (`.dig` is not necessary in this case):
```ruby
 return @source_sha256[ARCH.to_sym]
```

bin/crew
```diff
@@ -718,13 +667,12 @@ def upgrade(*pkgs, build_from_source: false)
...
     end
-    pkgVer_latest     = Package.load_package(pkgFile, pkgName).version
+    set_package File.join(CREW_PACKAGES_PATH, "#{pkgName}.rb")
     pkgVer_installed  = @device[:installed_packages].select { |pkg| pkg[:name] == pkgName } [0][:version]
-    pkgHash_latest    = Package.load_package(pkgFile, pkgName).get_binary_sha256(@device[:architecture])
     pkgHash_installed = @device[:installed_packages].select { |pkg| pkg[:name] == pkgName } [0][:binary_sha256]
 
-    return pkgHash_latest != pkgHash_installed unless !pkgHash_installed || pkgHash_latest == ''
-    return pkgVer_latest != pkgVer_installed
+    return @pkg.binary_sha256&.dig(ARCH.to_sym) != pkgHash_installed unless !pkgHash_installed || @pkg.binary_sha256&.dig(ARCH.to_sym) == 'null'
```
supechicken, line 674: Does `@pkg.binary_sha256&[ARCH.to_sym]` works? (`.dig` seems unnecessary in this case)

supechicken, line 674: Is there any case that `binary_sha256` is `'null'`?

bin/crew
Zopolis4: Having rsync "-v -ahHAXW" --remove-source... is the intent, though

bin/crew
Zopolis4: Yes-- if binary_sha256 is not found when writing to json, the ruby nil will become a string literall null in the json file

Zopolis4: I'm not sure-- this particular piece of code is very fragile, so I'm trying to avoid changing behaviour

lib/package.rb
Zopolis4: That's true, but we don't know if it has the architecture we need, so we still need dig to return nil instead of [] which returns an error. We can drop the safe navigation tho

bin/crew
```diff
@@ -224,18 +217,12 @@ end
 
 def list_available
   Dir["#{CREW_PACKAGES_PATH}/*.rb"].each do |filename|
-    pkgName      = File.basename(filename, '.rb')
     filelist     = File.join(CREW_META_PATH, "#{pkgName}.filelist")
```
uberhacker, line 220: Where is `pkgName` set?

Zopolis4: oops-- will fix when i next have the chance

bin/crew
```diff
@@ -206,11 +199,11 @@ def print_current_package(extra = false)
...
 
-def set_package(pkgName, pkgPath)
+def set_package(pkgPath)
   begin
-    @pkg = Package.load_package(pkgPath, pkgName)
-  rescue SyntaxError => e
-    warn "#{e.class}: #{e.message}".red
+    @pkg = Package.load_package(pkgPath)
+  rescue StandardError => e
+    puts "Error with #{pkgName}.rb: #{e}".red
```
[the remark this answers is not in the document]
Zopolis4: Will adjust when possible

bin/crew
Zopolis4: Yes-- I have removed the version of verbose without the leading `-`, and so I am making these changes to the previous usecases which relied on not having a leading `-`.

uberhacker (changes requested):
Rubocop complaints.

satmandu (approved):

bin/crew
satmandu: It's passing the github action, which does an install, so it works well enough to do that...

PR #9314, opened 2024-02-12
Add missing py3_* manifest filelist updates

uberhacker:
Finishes the remaining manifest filelist updates from the last python3 update.

Method used to install all py3_* packages to update the manifest filelists below:
```
cd /path/to/chromebrew/packages
for p in $(ls -1 py3_*|cut -d. -f1); do yes|crew install $p; done
git add ../manifest/*/p/py3_*
git commit -m 'Update py3 manifest filelists'
```


Zopolis4 (approved):

satmandu:
If we could combine this with a GitHub action to update the py3_ package versions from pypi and automatically update the manifests too...

PR #9316, opened 2024-02-12
Add build test source file to tools dir & hello_world_chromebrew build test package

satmandu:
- This adds one source file for the `hello_world_chromebrew` build test package, and the  `hello_world_chromebrew.rb` build test package, which will subsequently be added to the github actions for testing builds.


uberhacker (approved):
Good idea!

PR #9335, opened 2024-02-14
Mutter 41.0 => 45.4

uberhacker:
Add libei 1.2.1 dependency


satmandu:
Is there a glibc limitation?

uberhacker:
> Is there a glibc limitation?

Good question.  I tested the mutter executable with the early glibc versions (m90 and m91 containers) and it worked so maybe not?

satmandu (approved):

PR #9341, opened 2024-02-15
Have crew_profile_base remove the path env.d file from install.sh

satmandu:
- The chromebrew installer generates a `/usr/local/etc/env.d/path` file, but the `/usr/local/etc/env.d/00-path` file supersedes it, so remove the first file.
- Don't update the version number since this only needs to be fixed for new installs.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=path crew update
```

### DIFF
--- a/packages/crew_profile_base.rb
+++ b/packages/crew_profile_base.rb
@@ -30,7 +30,7 @@ class Crew_profile_base < Package
   def self.postinstall
     # Remove install.sh provided path file since we supersede it.
     if File.exist?("#{CREW_PREFIX}/etc/env.d/00-path")
-      puts "Removing #{CREW_PREFIX}/etc/env.d/path installed by the Chromebrew installer.".orange
+      puts "Removing #{CREW_PREFIX}/etc/env.d/path installed by the Chromebrew installer.\n".orange
       FileUtils.rm "#{CREW_PREFIX}/etc/env.d/path"
     end
 
--- a/packages/crew_profile_base.rb
+++ b/packages/crew_profile_base.rb
@@ -29,7 +29,10 @@ class Crew_profile_base < Package
 
   def self.postinstall
     # Remove install.sh provided path file since we supersede it.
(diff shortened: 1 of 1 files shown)


packages/crew_profile_base.rb
```diff
@@ -28,6 +28,12 @@ def self.install
   end
 
   def self.postinstall
+    # Remove install.sh provided path file since we supersede it.
```
Zopolis4, line 31: Put in lib/fixup.rb? Every installation of crew should have crew_profile_base anyways

install.sh
```diff
@@ -280,9 +280,6 @@ echo_out "Set up and synchronize local package repo..."
 echo "LD_LIBRARY_PATH=$CREW_PREFIX/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX}" >> "$CREW_PREFIX"/etc/env.d/00-library
 export LD_LIBRARY_PATH="${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX}"
 
-# Add the CREW_PREFIX bin and musl bin directories to PATH.
-echo -e "## Inserted by Chromebrew's install.sh\nPATH=$CREW_PREFIX/bin:$CREW_PREFIX/sbin:$CREW_PREFIX/share/musl/bin:\$PATH" > "$CREW_PREFIX"/etc/env.d/path
-
 echo "export CREW_PREFIX=${CREW_PREFIX}" >> "${CREW_PREFIX}/etc/env.d/profile"
```
uberhacker, line 283: Should this be in a package instead?  Otherwise, there is no way to uninstall on crew removal.

packages/crew_profile_base.rb
satmandu: moved!

install.sh
satmandu: it's in crew_profile_base...

Zopolis4 (approved):

PR #9391, opened 2024-02-26
Prompt to copy local repo package to crew prior to the build

uberhacker:
Removes one more manual process necessary for builds.

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1639,9 +1639,6 @@ end
 def resolve_dependencies_and_build
   @resolve_dependencies_and_build = 1
 
-  # Process preflight block to see if package should be built
-  pre_flight
-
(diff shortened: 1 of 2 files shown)


uberhacker:
> Allow a way to bypass the prompt so this is easily scripted? Use the force option or whatever...

`yes | crew build package`?

satmandu:
I have done that before, but `yes | program` has always seemed dangerous to me... especially as it messes with scripting.

Could you just add a timeout to automatically select the default if there is no input after say 10 seconds?

Ideally, I'd like to have this happen for the dialog to install dependent packages during builds, too... Too many times I start a build and it just gets stuck on a prompt...

satmandu (approved):

PR #9394, opened 2024-02-27
Enable additional temporarily disabled cops

Zopolis4:
When I added rubocop to crew, I temporarily disabled a number of cops in order to get the changes merged.

I'm now going back and re-enabling them.

The temporary cops I have left disabled are left because I am not immediately sure how to apply the fixes, apart from `Layout/HashAlignment`, which I am not sure on the style to use yet.

There are still some changes I'd like to make to our rubocop configuration, including enabling some other cops, but for now I'm just putting this out there.

Untested as of yet.
```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=cleanhop crew update
```


Zopolis4:
Due to the nature of github actions and rubocop, once this PR is merged all outstanding PRs will need to rebase onto master in order to get these new checks, as otherwise they may sneak into master with rubocop errors, and then new PRs based on master will be affected by those and have undeserved CI fails.

Zopolis4:
Tested everything I can think of, should be good to go.

lib/const.rb
```diff
@@ -76,9 +76,7 @@
   case ARCH
   when 'i686'
     CREW_KERNEL_VERSION = '3.8'
-  when 'aarch64', 'armv7l'
-    CREW_KERNEL_VERSION = '5.10'
-  when 'x86_64'
+  when 'aarch64', 'armv7l', 'x86_64'
```
uberhacker, line 79: How about just `else`?

lib/deb_utils.rb
```diff
@@ -11,25 +11,25 @@ def self.extract_deb(file, target)
     #   extract_deb('example.deb', /*/)        # extract all files from example.deb
     #
     file_found = false
-    src_fileIO = File.open(file, 'rb')
-    file_size = src_fileIO.size
+    src_file_io = File.open(file, 'rb')
+    file_size = src_file_io.size
```
uberhacker, line 15: I'm not a camel case fan either, especially when using underscores too.

lib/const.rb
Zopolis4: Doing that triggers some additional rubocop demands, but I let it autocorrect those

satmandu (approved):

PR #9429, opened 2024-03-03
Switch to sorting with LC_ALL=C for consistency and apply that to pre-existing filelists

Zopolis4:
Big diff, I know.

Just ensuring we get the same results from `sort` each time by setting `LC_ALL=C`, and running `LC_ALL=C sort` on all our filelists.

This should fix the issue where some characters have the same `sort` values, and thus we get different sort orders each time we build a package.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=evenbeforethat crew update
```


Zopolis4:
https://unix.stackexchange.com/a/87763
```
A few cases where you typically need to set LC_ALL=C:

sort -u or sort ... | uniq.... In many locales other than C, on some systems (notably GNU ones), some characters have the same sorting order. sort -u doesn't report unique lines, but one of each group of lines that have equal sorting order. 
So if you do want unique lines, you need a locale where characters are byte and all characters have different sorting order (which the C locale guarantees).
```

manifest/armv7l/a/a2ps.filelist
```diff
@@ -71,32 +72,32 @@
 /usr/local/share/a2ps/encoding/ms-cp1250.edf
 /usr/local/share/a2ps/encoding/ms-cp1251.edf
 /usr/local/share/a2ps/fonts/pcfont.pfa
+/usr/local/share/a2ps/ppd/README
 /usr/local/share/a2ps/ppd/level1.ppd
```
uberhacker, line 76: Shouldn't README come after level1.ppd?

Zopolis4: I guess it being in all caps means it is sorted first?

uberhacker: I suppose the ascii value is lower so that would make sense.

Zopolis4: We could use `LC_ALL=C sort -f` if we want to ignore case when sorting, I think.

uberhacker (approved):

PR #9430, opened 2024-03-03
Remove usage and creation of .directorylist files

Zopolis4:
This one was pretty simple, slightly more code complexity in `bin/crew` but it saves 800kb on a fresh install, with that increasing per packages installed. Should also slim down the sizes of our binary packages. Looking into `.filelist` files as well, but these ones were much more simple so I thought i'd get these changes out there.

Tested and working on `x86_64` (uninstalling `zoneinfo` works as before and removes empty directories)

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=cc0 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -253,8 +253,8 @@ def list_installed
     end
     print "\n"
   else
-    Dir["#{CREW_META_PATH}/*.directorylist"].map do |f|
-      File.basename(f, '.directorylist').lightgreen
+    @device[:installed_packages].each do |package|
+      puts package[:name].lightgreen
     end
   end
 end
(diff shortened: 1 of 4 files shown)


bin/crew
```diff
@@ -253,8 +253,8 @@ def list_installed
     end
     print "\n"
   else
-    Dir["#{CREW_META_PATH}/*.directorylist"].map do |f|
-      File.basename(f, '.directorylist').lightgreen
+    @device[:installed_packages].each do |package|
+      puts package[:name]
```
uberhacker, line 257: Missing `.lightgreen`?

Zopolis4: oops, fixed

bin/crew
```diff
@@ -1587,8 +1583,20 @@ def remove(pkg_name)
...
+      directories = []
+      filepaths.uniq.each do |directory|
+        current_directory = directory
+        loop do
+          directories << current_directory
+          break if current_directory == CREW_PREFIX
+          current_directory = File.dirname(current_directory)
+        end
+      end
+      directories.uniq!.map!(&:to_s)
```
uberhacker, line 1596: Isn't this reinventing the wheel?

uberhacker (changes requested):
This seems like a major change for a minor issue.

bin/crew
Zopolis4: By doing this at runtime, we are able avoid creating static lists of directories in addition to lists of files and thus save space.

uberhacker: What if we added directories to the filelists and removed everything in one fell swoop?  The directory would have to be the last entry after every file in that directory, however.

Zopolis4: As in, added directories to our existing filelists in `manifests`?

uberhacker: Yes, that way we could loop through all filelist entries and do a wholesale delete.

Zopolis4: I'm open to the idea-- but unless we are going to regenerate the filelist for every package, we'll still need to keep some code around to handle legacy cases.

Zopolis4: Although, we'd need some way to ensure that we remove directories last.

uberhacker: That was the intention of the .directorylist file.  You could append the directory list to the filelist and still have only one filelist per package.

Zopolis4: True-- that does mean we don't get a bunch of the space savings, though.

satmandu:
Why not just add code to compress the files using zstd if we want to save space?

Zopolis4:
It saves more space to delete the files altogether :wink:

Zopolis4:
Perhaps it will make sense to revisit this at a later date-- for now, I'll incorporate the non-contentious changes from this into some other PRs later.

PR #9435, opened 2024-03-04
update source_highlight, gdb, ctags

satmandu:
- Update `ctags` to `universal-ctags`.
- `source_highlight` updated to work with current `boost` 
- `gdb` => 14.2

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=gdb crew update
```


packages/source_highlight.rb
```diff
@@ -6,25 +6,25 @@
 class Source_highlight < Autotools
   description 'Convert source code to syntax highlighted document'
   homepage 'https://www.gnu.org/software/src-highlite/'
-  version '3.1.9-e4cf32d-2'
+  version '3.1.9-98bcd3f'
```
Zopolis4, line 9: This is a different git tag than the git_hashtag

uberhacker: Also, from the home page, you could also add ctags support.  Since version 2.2, source-highlight can also generate cross references; in order to do this it relies on GNU Ctags, http://ctags.sourceforge.net/.

`depends_on 'ctags'`

satmandu: Rebuilding `ctags`, but this should also automatically be part of the build since it is in build-deps...

satmandu: > Also, from the home page, you could also add ctags support. Since version 2.2, source-highlight can also generate cross references; in order to do this it relies on GNU Ctags, http://ctags.sourceforge.net/.
> 
> `depends_on 'ctags'`

Rebuilt with newer `ctags` on the system.

uberhacker (approved):

PR #9439, opened 2024-03-05
Add tmate & gnome updates

satmandu:
- Adds new `tmate` package, which is a variant of tmux with a remote shared screen.
image

- Updates `atkmm`, `cairomm_1_0`, `cairomm_1_16`, `glibmm_2_68`, `gstreamer`, `gtk4`, `gtkmm4`, `libevent`, `mm_common`, `msgpack_c`, `pangomm_1_4`, `pangomm_2_48`, and `transmission`

Builds properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=tmate crew update
```


Zopolis4:
What is the purpose of splitting `cairomm` into multiple versions? Is it still needed?

packages/msgpack_c.rb
```diff
@@ -1,50 +1,38 @@
...
 
-  def self.build
-    FileUtils.mkdir('builddir')
-    Dir.chdir('builddir') do
-      system "cmake #{CREW_CMAKE_OPTIONS} \
-      -DCMAKE_C_STANDARD='17' \
+  cmake_options "-DCMAKE_C_STANDARD='17' \
       -DBUILD_SHARED_LIBS=ON \
+      -DMSGPACK_32BIT=#{ARCH == 'i686' ? 'ON' : 'OFF'} \
       -DMSGPACK_ENABLE_STATIC=OFF \
```
Zopolis4, line 35: Already disabled by our earlier passing of `-DBUILD_SHARED_LIBS=ON`.
https://github.com/msgpack/msgpack-c/blob/c-6.0.0/CMakeLists.txt#L90-L93

packages/msgpack_c.rb
```diff
@@ -1,50 +1,38 @@
...
     system "sed -i 's,CMAKE_INSTALL_PREFIX}/lib,CMAKE_INSTALL_PREFIX}/#{ARCH_LIB},g' CMakeLists.txt"
     system "sed -i 's,CMAKE_INSTALL_LIBDIR lib,CMAKE_INSTALL_LIBDIR #{ARCH_LIB},g' CMakeLists.txt"
   end
 
-  def self.build
-    FileUtils.mkdir('builddir')
-    Dir.chdir('builddir') do
-      system "cmake #{CREW_CMAKE_OPTIONS} \
-      -DCMAKE_C_STANDARD='17' \
+  cmake_options "-DCMAKE_C_STANDARD='17' \
```
Zopolis4, line 32: Any reason for doing this?

packages/msgpack_c.rb
```diff
@@ -1,50 +1,38 @@
...
+       i686: '23f44da7a26b62bc6b38be11b7c2a5cefc1b2ac11a5d87982f73a4a2edb8ad03',
+     x86_64: '4171f5ad1d9e9b4e40b06e0e9f68173edace4f25d7d95e8b1f9e581e45a87f7d'
   })
 
-  depends_on 'cmake' => :build
   depends_on 'doxygen' => :build
-  depends_on 'graphviz' => :build
+  depends_on 'glibc' # R
 
   def self.patch
```
Zopolis4, line 26: Send these improvements upstream? (i.e. add support for passing `-DCMAKE_INSTALL_PREFIX` the standard way)

packages/tmate.rb
```diff
@@ -0,0 +1,37 @@
...
+  depends_on 'msgpack_c' # R
+  depends_on 'ncurses' # R
+
+  def self.patch
+    # Patch for msgpack_c 6.x
+    downloader 'https://github.com/tmate-io/tmate/pull/281.patch', '9f4f315a92b959606d42d672cf8c7312afaf4a93fba6c0e21bfee0b84e2d4667'
+    system 'patch -p1 -i 281.patch'
+  end
+
+  pre_configure_options "CFLAGS='-I#{CREW_PREFIX}/include/ncurses' CPPFLAGS='-I#{CREW_PREFIX}/include/ncurses'"
```
Zopolis4, line 36: It's [this issue](https://gitlab.kitware.com/cmake/cmake/-/issues/23236) again...

Zopolis4 (changes requested):
Also, why are we disabling building documentation? That's not what we normally do...


Same thing with tests-- if it doesnt result in additional installed binaries or anything, unlike examples, we should be building and running them.

satmandu:
Documentation is available on the web for libraries. We don't need to install it locally. The goal is to minimize size for library packages.

satmandu:
> What is the purpose of splitting `cairomm` into multiple versions? Is it still needed?

There are different version libraries.... With packages dependent upon particular versions...

packages/msgpack_c.rb
satmandu: Didn't change the existing option...

packages/msgpack_c.rb
satmandu: Feel free!

Zopolis4:
> Documentation is available on the web for libraries. We don't need to install it locally. The goal is to minimize size for library packages.

We install documentation for other packages, and we compress it to save size...

Let's keep the status quo for these packages for now, though.

Zopolis4:
What about building tests?

satmandu:
> > Documentation is available on the web for libraries. We don't need to install it locally. The goal is to minimize size for library packages.
> 
> We install documentation for other packages, and we compress it to save size...
> 
> Let's keep the status quo for these packages for now, though.

Respectfully, I'd like to move in the other direction. We should set the `gnome` flagged packages to automatically stop building documentation. NOBODY uses the enclosed documentation for gnome libraries. Everyone just googles for documentation anyways.

satmandu:
> What about building tests?

I'm honestly not sure the tests in these libraries are meaningful. They're not command line apps. Are we really going to follow up upstream with failed tests on widely used and tested gnome packages?

I'd be perfectly happy having tests run by default, and then set the gnome packages to not run tests by default. I just think it's a waste of CPU time...

Zopolis4:
We run tests for packages for the purpose of ensuring we haven't broken anything-- a package can build fine but be broken at runtime, and tests help to catch some of that.

satmandu:
I have removed _the don't build tests_ lines from the files in this PR, so the next time these are built, they can be tested. You can add a PR to make the buildsystems builds run tests by default if you have time. :)

packages/gtkmm4.rb
```diff
@@ -3,23 +3,34 @@
...
+  depends_on 'cairo' # R
+  depends_on 'gcc_lib' # R
+  depends_on 'gdk_pixbuf' # R
+  depends_on 'glibc' # R
+  depends_on 'glibmm_2_68' # R
+  depends_on 'glib' # R
+  depends_on 'gobject_introspection' => :build
+  depends_on 'graphene' # R
+  depends_on 'gtk4' => :build
+  depends_on 'gtk4' # R
```
Zopolis4, line 31: `gtk4` is listed twice.

Zopolis4 (dismissed):

supechicken:
> Documentation is available on the web for libraries. We don't need to install it locally. The goal is to minimize size for library packages.

I agree with @satmandu on this point. Especially when most web documents have a nicer format compared with the `man` one

packages/gtkmm4.rb
satmandu: Good catch! The new build system automatically adds the runtime dep if missing! Hmm. Maybe I will write code at some point to also switch the build dep to runtime dep if it notices the build dep is already there.

uberhacker:
> > Documentation is available on the web for libraries. We don't need to install it locally. The goal is to minimize size for library packages.
> 
> I agree with @satmandu on this point. Especially when most web documents have a nicer format compared with the `man` one

Not to mention the savings on valuable disk space.

packages/atkmm.rb
```diff
@@ -1,20 +1,20 @@
...
   source_url 'https://gitlab.gnome.org/GNOME/atkmm.git'
   git_hashtag version
   binary_compression 'tar.zst'
 
   binary_sha256({
-    aarch64: '049d1efb8c411ac36328983605a54d54e3bc45587aa3ac9157f4b629be75f720',
-     armv7l: '049d1efb8c411ac36328983605a54d54e3bc45587aa3ac9157f4b629be75f720',
+    aarch64: '9224b82b6db207ebcb652edeb5403eec1be918f7b395d449c2b4a6bd04298a37',
+     armv7l: '9224b82b6db207ebcb652edeb5403eec1be918f7b395d449c2b4a6bd04298a37',
        i686: '4ceaac05ebc9ae13009b10103249356131e260b7ce8ed3a3e57ad91f73f174e2',
```
uberhacker, line 16: Missing i686 update?

satmandu: i686 removed... it doesn't make sense to do gnome updates for i686...

packages/gtkmm4.rb
```diff
@@ -3,23 +3,31 @@
...
+  depends_on 'gdk_pixbuf' # R
+  depends_on 'glibc' # R
+  depends_on 'glibmm_2_68' # R
+  depends_on 'glib' # R
+  depends_on 'gobject_introspection' => :build
+  depends_on 'graphene' # R
+  depends_on 'gtk4' # R
+  depends_on 'libsigcplusplus3' # R
+  depends_on 'pangomm_2_48' # R
+  depends_on 'vulkan_headers' => :build
```
uberhacker, line 32: Disable documentation?  Filelists are huge.

packages/atkmm.rb
uberhacker: I see it was deleted.  Great!

uberhacker:
> > > Documentation is available on the web for libraries. We don't need to install it locally. The goal is to minimize size for library packages.
> > 
> > 
> > I agree with @satmandu on this point. Especially when most web documents have a nicer format compared with the `man` one
> 
> Not to mention the savings on valuable disk space.

It might be helpful to provide a link to the online documentation in postinstall?

satmandu:
> > > > Documentation is available on the web for libraries. We don't need to install it locally. The goal is to minimize size for library packages.
> > > 
> > > 
> > > I agree with @satmandu on this point. Especially when most web documents have a nicer format compared with the `man` one
> > 
> > Not to mention the savings on valuable disk space.
> 
> It might be helpful to provide a link to the online documentation in postinstall?

It's elementary to Google documentation on a library...

Zopolis4:
:woman_shrugging: the room is against me on this one, let's disable documentation.

packages/msgpack_c.rb
```diff
@@ -1,50 +1,38 @@
...
-require 'package'
+require 'buildsystems/cmake'
 
-class Msgpack_c < Package
+class Msgpack_c < CMake
   description 'An efficient object serialization library'
   homepage 'https://msgpack.org/'
-  version '4.0.0'
+  version '6.0.0'
   license 'custom:Boost'
```
Zopolis4, line 10: From what I can see on the repo, its just the version 1.0 of the Boost Software License (BSL-1.0), nothing custom.

packages/gtkmm4.rb
satmandu: Done.

uberhacker (approved):

uberhacker (approved):

PR #9444, opened 2024-03-06
Add no_source_build property

satmandu:
- As used in https://github.com/chromebrew/chromebrew/pull/9329 for `llvm*_dev` and `llvm*_lib`.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=no_source_build crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1904,7 +1904,7 @@ def build_command(args)
     # Process preflight block to see if package should be built
     pre_flight
 
-    if !@pkg.is_fake? && @pkg.compatible? && @pkg.source?(ARCH) && @pkg.source_url.to_s.upcase != 'SKIP' && !@pkg.no_compile_needed?
+    if !@pkg.is_fake? && @pkg.compatible? && @pkg.source?(ARCH) && ( @pkg.no_source_build? || @pkg.source_url.to_s.upcase != 'SKIP' ) && !@pkg.no_compile_needed?
       resolve_dependencies_and_build
     else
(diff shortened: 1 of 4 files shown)


Zopolis4:
Why do we need this? `source_url 'SKIP'` does the same thing, and more cleanly.

In fact, you add the code for this right next to the code for `source_url 'SKIP'` which does the same thing...

satmandu:
Doesn't work any more, as I discovered earlier today doing the llvm18_dev and llvm18_lib builds. Try to build and you will see, and if you look at the changes you see how I fixed it.

satmandu:
Also yes I agree with you ....

Zopolis4:
It probably makes more sense to fix the bug instead of adding another property :sweat_smile:

uberhacker:
> It probably makes more sense to fix the bug instead of adding another property 😅

I actually don't exactly consider this a bug.  Every package we "build" should be from source.  We're just bastardizing the process to make things more modular.  I'm thinking there might be a more elegant way to handle these pseudo packages.

satmandu:
I'm currently using this to build the LLVM18 packages, so it would help to get this in for now... I don't know where the regression is.

uberhacker:
> I'm currently using this to build the LLVM18 packages, so it would help to get this in for now... I don't know where the regression is.

The regression is in build_command.  The build is rejected because there is no source.  The assumption is there must be source to actually perform a build.

uberhacker (approved):

PR #9445, opened 2024-03-06
Split const command out into separate file

Zopolis4:
**Split out `const` command to `commands/const.rb`.**

Behaviour changes:
- Added error catching in the case of an invalid constant.

Refactoring changes:
- Change handling of no argument being passed to use `end.empty?` the same way `def search_command` does. This will be the way I handle similar cases.
 
Other changes:
- Add testing in `tests/commands/const.rb`

Tested and working on `x86_64`, and tests pass.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula1 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -4,6 +4,7 @@ require 'digest/sha2'
 require 'json'
 require 'fileutils'
 require 'tmpdir'
+require_relative '../commands/const'
 require_relative '../commands/help'
 require_relative '../lib/docopt'
 require_relative '../lib/color'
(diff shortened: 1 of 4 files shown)


uberhacker:
This looks great @Zopolis4!  How do you run the tests?

satmandu (approved):

satmandu:
> This looks great @Zopolis4! How do you run the tests?

```
cd /usr/local/lib/crew/tests/commands
ruby const.rb
```

PR #9451, opened 2024-03-08
Mandb: background cache rebuild from postinstall

satmandu:
- the i686 installs hang for a very long time on the mandb cache generation process, so background the process to try to let installs continue.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=mandb crew update
```

### DIFF
--- a/packages/mandb.rb
+++ b/packages/mandb.rb
@@ -62,10 +62,11 @@ class Mandb < Package
   end
 
   def self.postinstall
-    puts 'Starting backgrounded mandb cache rebuild. This will be logged to /tmp/man-db-rebuild.log.'.yellow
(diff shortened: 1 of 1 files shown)


packages/mandb.rb
```diff
@@ -62,10 +62,10 @@ def self.install
   end
 
   def self.postinstall
-    puts 'Started mandb cache rebuild. (Errors from this can either be ignored or reported upstream to the relevant package maintainers.)'.yellow
+    puts 'Starting backgrounded mandb cache rebuild. This will be logged to /tmp/man-db-rebuild.log.'.yellow
```
uberhacker, line 65: Log to `/usr/local/var/log` instead?  The `/tmp` location is not a "best practice" location for log files.

satmandu: Sure.

uberhacker (approved):

PR #9453, opened 2024-03-08
Adjust git sparse checkout options

satmandu:
- This is all in service of trying to get rid of the i686 container sparse checkout error messages on opening a container.
image

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=git_sparse crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -434,9 +434,14 @@ def update
   abort "'crew update' is used to update crew itself. Use 'crew upgrade <package1> [<package2> ...]' to update specific packages.".orange if @pkg_name
 
   # update package lists
+  @other_arches = %w[i686 x86_64 armv7l]
+  @other_arches.reject! { |n| n == ARCH }
   Dir.chdir(CREW_LIB_PATH) do
     # Set sparse-checkout folders.
-    system "git sparse-checkout set packages manifest/#{ARCH} lib commands bin crew tests tools"
+    system "git sparse-checkout set --no-cone \"/*\" packages \"/manifest/#{ARCH}/*\" lib commands bin crew tests tools"
+    @other_arches.each do |other_arch|
+      system "git sparse-checkout add \"!/manifest/#{other_arch}/\""
+    end
     system 'git sparse-checkout reapply'
 
     system "git fetch #{CREW_REPO} #{CREW_BRANCH}", exception: true
--- a/bin/crew
+++ b/bin/crew
@@ -438,7 +438,7 @@ def update
   @other_arches.reject! { |n| n == ARCH }
   Dir.chdir(CREW_LIB_PATH) do
     # Set sparse-checkout folders.
-    system "git sparse-checkout set --no-cone \"/*\" packages \"/manifest/#{ARCH}/*\" lib commands bin crew tests tools"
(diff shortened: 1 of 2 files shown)


Zopolis4:
Does this fix the issue? Fundamentally the only cause should be that the ARCH value is wrong-- messing with the way we set it is not an ideal solution.

satmandu:
> Does this fix the issue? Fundamentally the only cause should be that the ARCH value is wrong-- messing with the way we set it is not an ideal solution.

I can't tell yet if it fixes the issue. It only happens when we first open the container.

Do the changes seem sane?

bin/crew
```diff
@@ -434,9 +434,14 @@ def update
   abort "'crew update' is used to update crew itself. Use 'crew upgrade <package1> [<package2> ...]' to update specific packages.".orange if @pkg_name
 
   # update package lists
+  @other_arches = %w[i686 x86_64 armv7l]
+  @other_arches.reject! { |n| n == ARCH }
   Dir.chdir(CREW_LIB_PATH) do
     # Set sparse-checkout folders.
-    system "git sparse-checkout set packages manifest/#{ARCH} lib commands bin crew tests tools"
+    system "git sparse-checkout set --no-cone \"/*\" bin commands crew lib \"/manifest/#{ARCH}/*\" packages tests tools"
```
uberhacker, line 441: `--no-clone`?

satmandu: No it's `cone` referring to the new sparse algorithm we are avoiding.

satmandu: https://stackoverflow.com/questions/9572407/git-sparse-checkout-with-exclusion#26129711

satmandu:
Also, `ARCH` shouldn't be setup wrong:
```
chronos@cheekon-i686 /usr/local/lib/crew $ crew const | grep ARCH
ARCH=i686
ARCH_LIB=lib
CPU_SUPPORTED_ARCH=["i686", "x86_64"]
CREW_ARCH_FLAGS=
CREW_ARCH_FLAGS_OVERRIDE=
CREW_NOT_SHRINK_ARCHIVE=false
KERN_ARCH=i686
chronos@cheekon-i686 /usr/local/lib/crew $ uname -m
i686
chronos@cheekon-i686 /usr/local/lib/crew $
```

Zopolis4:
I'm thinking the issue might be in the setup of the docker container, before setarch is available. The error message suggests that manifest/x86_64 was originally set, then later manifest/i686 was set, but manifest/x86_64 still remains. This would be consistent with the issue occurring in the initial installation.

satmandu:
> I'm thinking the issue might be in the setup of the docker container, before setarch is available. The error message suggests that manifest/x86_64 was originally set, then later manifest/i686 was set, but manifest/x86_64 still remains. This would be consistent with the issue occurring in the initial installation.

I think you're right. This is probably an interaction with the `chromebrewstart` command, which might be running `crew update` from `~/.bashrc` before `/usr/local/bin/setarch \\\${SETARCH_ARCH} sudo -i -u chronos` is invoked.

satmandu:
Will work on fixing the container startup scripts...

PR #9454, opened 2024-03-09
Split list command into separate file

Zopolis4:
**Split out `const` command to `commands/const.rb`.**

No behaviour changes.

Refactoring changes:
- Merge commands back into one function, to allow for re-use of the installed_packages hash
- Use `pkg.compatible?` instead of `@device[:compatible_packages]`
  - Load packages to test compatibility
- Check if packages are installed by accessing `installed_packages`, rather than checking for a filelist in `CREW_META_PATH`
- Do not rescue errors on loading packages (these will be caught by `prop_test` in CI before they get merged)
- Find the width of the first column of installed packages through a `.size` based `.max` comparison
- Align the installed packages via `.ljust` and two manual spaces in `puts` 

The code could probably be simplified a bit further if it shared the `Dir["#{CREW_PACKAGES_PATH}/*.rb"].each do |filename|` loop, but this seems like it would dramatically impact performance (not sure).

No testing added, simply because the current testing method is not well set up for testing cases like these where the expected output is hundreds of lines long. I may go back in future and work out a testing method.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula2 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -6,6 +6,7 @@ require 'fileutils'
 require 'tmpdir'
 require_relative '../commands/const'
 require_relative '../commands/help'
+require_relative '../commands/list'
 require_relative '../lib/docopt'
 require_relative '../lib/color'
 require_relative '../lib/const'
@@ -219,62 +220,6 @@ end
 def list_packages
   Dir["#{CREW_PACKAGES_PATH}/*.rb"].each do |filename|
     print_package filename
-  end
-end
-
-def list_available
-  Dir["#{CREW_PACKAGES_PATH}/*.rb"].each do |filename|
-    pkg_name     = File.basename(filename, '.rb')
-    filelist     = File.join(CREW_META_PATH, "#{pkg_name}.filelist")
-    not_installed = !File.file?(filelist)
-
-    next unless not_installed
-
-    begin
-      set_package pkg_name, filename
-    rescue StandardError => e
-      puts "Error with #{pkg_name}.rb: #{e}".red unless e.to_s.include?('uninitialized constant')
-    end
-    puts pkg_name if @pkg.compatible?
-  end
-end
-
-def list_installed
-  if @opt_verbose
-    installed_packages = []
-    @device[:installed_packages].each do |package|
-      search package[:name], true
-      installed_packages.append("#{package[:name]} #{package[:version]}")
-    end
-    sorted_installed_packages = installed_packages.sort
-    sorted_installed_packages.unshift('======= =======')
-    sorted_installed_packages.unshift('Package Version')
-    first_col_width = sorted_installed_packages.map(&:split).map(&:first).max_by(&:size).size + 2
-    sorted_installed_packages.map(&:strip).each do |line|
-      puts "%-#{first_col_width}s%s".lightgreen % line.split
-    end
-    print "\n"
-  else
-    Dir["#{CREW_META_PATH}/*.directorylist"].map do |f|
-      File.basename(f, '.directorylist').lightgreen
-    end
-  end
-end
-
-def list_compatible(compat = true)
-  Dir["#{CREW_PACKAGES_PATH}/*.rb"].each do |filename|
-    pkg_name = File.basename filename, '.rb'
-    if @device[:compatible_packages].any? { |elem| elem[:name] == pkg_name }
-      if compat
-        if File.file? "#{CREW_META_PATH}/#{pkg_name}.filelist"
-          puts pkg_name.lightgreen
-        else
-          puts pkg_name
-        end
-      end
-    elsif !compat
-      puts pkg_name.lightred
-    end
   end
 end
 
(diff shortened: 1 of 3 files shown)


commands/list.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    if available
+      Dir["#{CREW_PACKAGES_PATH}/*.rb"].each do |filename|
+        pkg_name = File.basename(filename, '.rb')
+        next if installed_packages.key?(pkg_name)
+        pkg = Package.load_package(filename)
+        puts pkg_name if pkg.compatible?
+      end
+    elsif installed
```
uberhacker, line 21: Use `case ... when` instead of `if ... elsif`?

Zopolis4: Because they're all passed as separate arguments (in order to avoid a second case statement in `bin/crew`), we cant use a `case` statement.

commands/list.rb
```diff
@@ -0,0 +1,49 @@
+require 'fileutils'
+require_relative '../lib/color'
+require_relative '../lib/const'
+require_relative '../lib/package'
+
+class Command
+  def self.list(available, installed, compatible, incompatible, verbose)
```
uberhacker, line 7: Use one argument?

bin/crew
```diff
@@ -1964,15 +1909,7 @@ def install_command(args)
...
-  if args['available']
-    list_available
-  elsif args['installed']
-    puts list_installed
-  elsif args['compatible']
-    list_compatible true
-  elsif args['incompatible']
-    list_compatible false
-  end
+  Command.list(args['available'], args['installed'], args['compatible'], args['incompatible'], @opt_verbose)
```
uberhacker, line 1912: Shouldn't need to pass `@opt_verbose` since it is a global instance variable.

Zopolis4: Global variables and instance variables are different-- `@opt_verbose` is accessible to everything in `bin/crew`, but we are crossing a class barrier and thus instance variables do not propagate.

commands/list.rb
Zopolis4: This saves us having to use an additional `if` statement in `bin/crew`:

One argument:
```ruby
# in bin/crew
def list_command(args)
  if args['available']
    Command.list(args['available'])
  elsif args['installed']
    Command.list(args['compatible'], @opt_verbose)
  elsif args['compatible']
    Command.list(args['compatible'])
  elsif args['incompatible']
    Command.list(args['incompatible'])
end
  
# in commands/list.rb
def self.list(args, verbose)
  when args
  case 'available'
    ...
  case 'installed'
    if verbose
      ...
    else
      ...
    end
  case 'compatible'
    ...
  case 'incompatible'
    ...
  end
end
```

Multiple arguments:
```ruby
# in bin/crew
def list_command(args)
  Command.list(args['available'], args['installed'], args['compatible'], args['incompatible'], @opt_verbose)
end
  
# in commands/list.rb
def self.list(available, installed, compatible, incompatible, verbose)
  if available
    ...
  elsif installed
    if verbose
      ...
    else
      ...
    end
  elsif compatible
    ...
  elsif incompatible
    ...
  end
end
```

uberhacker: I don't see how that is less complicated.  Just pass `args` and then use the case statement.  No need to list out all the arguments as parameters.

commands/list.rb
uberhacker: You get rid of the if statement altogether if you simply pass args and then make the decision in the function instead.

commands/list.rb
Zopolis4: My end goal is that every `*_command(args)` function is a one-line call to the relevant command. 

I want to have the minimum possible amount of logic in `bin/crew`.

I can see no downside to having multiple parameters, but I do see the upside of saving lines of code and complexity that they allow.

uberhacker:
```
# in bin/crew
def list_command(args)
  Command.list(args)
end
  
# in commands/list.rb
def self.list(args)
  args.each do |arg|
    case arg
    when '-v', '--verbose'
      verbose = true
    end
  end
  args.each do |arg|
    case arg
    when 'available'
      ...
    when 'installed'
      if verbose
        ...
      else
        ...
      end
    when 'compatible'
      ...
    when 'incompatible'
      ...
    end
  end
end
```

Zopolis4: We use docopt for a reason-- moving argument handling logic to each command file and duplicating it does not seem like a good solution.

uberhacker: I guess I'm not understanding the docopt process.  Seems counter intuitive.  Passing an entire list of parameters instead of a single parameter array is certainly  not a best practice.

Zopolis4: Parts of docopt are certainly not self-explanatory, this is the best solution I found that optimises for code simplicity and length. I agree the list of parameters is not ideal, but I'd prefer it over the other options.

lib/const.rb
```diff
@@ -1,7 +1,7 @@
 # lib/const.rb
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.44.8'
+CREW_VERSION = '1.44.9'
```
satmandu, line 4: Bump to `1.45.0`?

satmandu (approved):

satmandu (approved):

Issue #7645, opened 2022-11-17
Deprecation of i686 based systems


supechicken:
### Overview
The last official release of `i686` Chrome OS (excluding CloudReady) was released in 2017 (5 years ago). That means the lack of security patches and an older system base, which makes it more and more unreliable in daily use.

Currently, some online features (like extension store and Chrome history syncing) on these systems are no longer working, even some of the Chromebrew applications are broken due to a lack of tests.

Although there are not many efforts required for us to maintain existing `i686` packages, there are very very few users (or nobody) using `crew` on those platforms or even the OS itself (which means our efforts are meaningless)

So this might be a perfect time to say goodbye :)

### Timeline

> [!NOTE]
> All plans below have been put on hold, everything remains currently.
>
> However, please notice that the support of i686 might be dropped soon.

- Now: No changes
- 01/2023: New packages will not provide prebuilt binary for `i686`  (can still build from source)
- 02/2023: Disable i686 compatibility for `sommelier` and all X11 applications.
- 05/2023: New updates will not provide prebuilt binary for `i686`  (except core packages/compilers)  (can still build from source)
- 09/2023: Stop providing prebuilt binary for `i686` on all packages (existing binaries will be kept on gitlab)
- 12/2023: No new bug fixes on crew command regarding `i686`
- ~~01/2024~~ future plan: Completely remove support and prebuilt binaries for `i686`

uberhacker:
It's even worse that you mentioned.  The Chrome browser has many more issues that make it incompatible with several sites.  Even Google Drive doesn't work on older releases anymore.  But...a lot of cli apps still work great.  I definitely think it's not worth the effort to continue to struggle with the gui stuff but cli apps are fine to continue to build.

alexsch01:
@supechicken I agree that 32-bit x86 systems should eventually be depreciated.
However, how does this work for things like Wine and Steam that require 32-bit x86 packages?

satmandu:
Wine emulation of x86 requires multilib (x86 libraries on x86_64) which Chromebrew has never supported...

alexsch01:
Google has also dropped support for all ARMv7 chromebooks

satmandu:
> Google has also dropped support for all ARMv7 chromebooks

I think that just means Chromebooks with an `armv7l` kernel.  All arm Chromebooks still use an `armv7l` userspace...

satmandu:
[...]

PR #9458, opened 2024-03-09
Remove compatibility from packages with incompatible dependencies

Zopolis4:
If a package has incompatible dependencies on an architecture, the installation logically should fail on that architecture. 

It is important to note that this PR does not remove compatibility from any packages that actually had it before-- trying to install all of these on i686 would trigger a build from source on incompatible packages, which would have failed. 

If that build would not have failed, the package shouldn't have been marked as `i686` incompatible (assuming it has no `i686` incompatible dependencies)

I am aware this removes compatibility from a lot of packages (around 340 for `i686`, and 2 for `armv7l`)

However, in the majority of these cases, the package had simply not been rebuilt since the dropping of GUI support for `i686`, and so would not have worked anyways.

In some cases, namely with `cairo`, non-GUI packages that have a dependency on packages with optional GUI functionality lose `i686` compatibility despite technically being able to build. 

It is important to note that they still could not have built under the current `crew` configuration, just that if we changed the way `cairo` and other packages worked they could. 

However, as per my comments in #7645, this would negatively affect the functionality of chromebrew, as we would have to introduce significant additional complexity to support `i686`.

Here is the code I used to find which packages have incompatible dependencies:
```ruby
#!/usr/bin/env ruby

# Add >LOCAL< lib to LOAD_PATH so that packages can be loaded
$LOAD_PATH.unshift '../lib'

  dependencies = pkg.get_deps_list
  incompatible_dependencies = dependencies.any? { |dep| !Package.load_package("#{CREW_PACKAGES_PATH}/#{dep}.rb").compatible? }
  puts "#{pkg.name} should be marked as incompatibile with #{ARCH}" if pkg.compatible? && incompatible_dependencies
end
```

I am not adding this to the tests because it only works if you are on the relevant architecture, and we only test on `x86_64` at the moment. However, I have added code to `crew` itself to handle this scenario in the meantime, and when we can ensure this will not happen due to it already being tested in CI we can remove that code.

All architectures tested, and have no remaining packages falsely marked as compatible.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=pendepdepdpe crew update
```


bin/crew
```diff
@@ -1308,6 +1308,9 @@ def resolve_dependencies
   # leave only dependency names (remove all package attributes returned by @pkg.get_deps_list)
   @dependencies.map!(&:keys).flatten!
 
+  # abort if we have incompatible dependencies
+  abort "Some dependencies are not compatible with your device architecture (#{ARCH}). Skipping install".lightred if @dependencies.any? { |dep| !Package.load_package("#{CREW_PACKAGES_PATH}/#{dep}.rb").compatible? }
```
uberhacker, line 1312: This is not isolated to just installs but builds also.  Maybe change `Skipping install` to `Unable to continue`?

uberhacker: It might be helpful to list the dependencies which are incompatible.

bin/crew
```diff
@@ -1308,6 +1308,9 @@ def resolve_dependencies
   # leave only dependency names (remove all package attributes returned by @pkg.get_deps_list)
   @dependencies.map!(&:keys).flatten!
 
+  # abort if we have incompatible dependencies
+  abort "Some dependencies are not compatible with your device architecture (#{ARCH}). Unable to continue.".lightred if @dependencies.any? { |dep| !Package.load_package("#{CREW_PACKAGES_PATH}/#{dep}.rb").compatible? }
```
uberhacker, line 1312: It might be helpful to list the incompatible dependencies.

bin/crew
Zopolis4: We don't know what they are-- It's possible to rework the code so that we do, but that'll slow things down.

Besides, this is only there in a worst-case scenario, we should only hit this code if someone forgets to change compatibility for a package.

bin/crew
Zopolis4: To do that would mean reworking the code to be slower and more complex. 

In this case, given that this is only here as a worst-case scenario backup, I think makes more sense not to.

PR #9459, opened 2024-03-09
crew: allow override of ARCH const in crew with env variable

satmandu:
- The Chromebrew installer creates the wrong manifest directories during installs on `i686` because it runs without `setarch` on a `x86_64` host. Rather than go through setting up `setarch` to download first during the container installs, allow overrides of the internal `ARCH` constant if the `ARCH` env variable is set.
image

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=arch crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -1,7 +1,7 @@
 # lib/const.rb
 # Defines common constants used in different parts of crew
 
-CREW_VERSION = '1.44.8'
+CREW_VERSION = '1.44.9'
 
 # kernel architecture
 KERN_ARCH = `uname -m`.chomp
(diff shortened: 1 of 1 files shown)


lib/const.rb
```diff
@@ -40,7 +40,7 @@
 # This helps with virtualized builds on aarch64 machines which report armv8l when linux32 is run.
 # We also report aarch64 machines as armv7l for now, as we treat them as if they were armv7l.
 # When we have proper aarch64 support, remove this.
-ARCH = %w[aarch64 armv8l].include?(KERN_ARCH) ? 'armv7l' : KERN_ARCH
+ARCH = ENV.fetch('ARCH', %w[aarch64 armv8l].include?(KERN_ARCH) ? 'armv7l' : KERN_ARCH)
```
uberhacker, line 43: Won't this set ARCH to aarch64 when you want it armv7l?

satmandu: This literally only modifies the existing code to override only if an `ARCH` env variable exists. So if you do a `export ARCH=this_does_not_make_sense` then yes that is what `crew const ARCH` will report.

satmandu: If `ARCH` isn't set, the existing code applies.

satmandu: I have no problem with doing that instead.

satmandu: done.

Zopolis4:
If we're doing this, does it make sense to remove all of the other existing container detection code around `ARCH`? (i.e. `KERN_ARCH`, now that we can just override `ARCH` in the container)

satmandu:
Let me just see if this fixes the issue first, and then we can work on cleanup?

Zopolis4 (approved):
fair enough

PR #9460, opened 2024-03-09
Telegram 4.15.0 => 4.15.1

uberhacker:


manifest/x86_64/t/telegram.filelist
```diff
@@ -1,2 +1,2 @@
-/usr/local/bin/telegram
 /usr/local/bin/Telegram
```
Zopolis4, line 1: This shouldnt be happening... we should have consistent sorting orders now. Is there anything you are doing to change the way filelists are created?

uberhacker: No, it's just sorting the way YOU changed in https://github.com/chromebrew/chromebrew/pull/9429. :)

satmandu (approved):

PR #9471, opened 2024-03-11
Add git_mestrelion_tools package

satmandu:
- This uses the _upstream_ package name from `debian` as we already have a `gittools` package.
- From debian's description:
   * git-restore-mtime - set timestamps to the date of a file's last commit
   * git-clone-subset - clones only some files from a repository (inc. history)
   * git-find-uncommitted-repos - recursively searches for unclean git repos
   * git-strip-merge - filters away some files during a merge

Builds properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=gittoolsgit crew update
```


packages/git_restore_mtime.rb
```diff
@@ -0,0 +1,34 @@
+# Adapted from Arch Linux git-tools-git PKGBUILD at:
+# https://aur.archlinux.org/cgit/aur.git/plain/PKGBUILD?h=git-tools-git
+
+require 'package'
+
+class Git_restore_mtime < Package
```
uberhacker, line 6: Isn't there an existing gittools package?

satmandu: Yes that's my point with a different name. But the existing package in Chromebrew is entirely different.

uberhacker: I just checked and that is a different repository.  That said, since this includes more executables than just git-restore-mtime, would it make more sense to give it a more generic name like maybe git_utils?

satmandu: I was considering that, but thought that the best of mediocre solutions would just be to replicate a major upstream package name, so that people could easily find it if they were looking.

See https://packages.debian.org/sid/git-restore-mtime

satmandu: Actually, looking at that, how about `git-mestrelion-tools`?

uberhacker: Not sure what mestrelion means but I guess it would be fine.

uberhacker (approved):

PR #9474, opened 2024-03-12
crew: run git-restore-mtime on crew update

satmandu:
- We only run `git fetch`, and so we cannot use a git hook for running `git-restore-mtime`. ([No git hook exists](https://git-scm.com/docs/githooks) for `git fetch`.)
- The simpler way to handle this is to just run `git-restore-mtime` during crew update from crew. (Hence the addition of the tool to `core.rb`.)
- Note that STDERR is removed on purpose when running this command, because otherwise the output is littered with sparse-checkout excluded folder information. See https://github.com/MestreLion/git-tools/issues/70
- Also use `git-restore-mtime` in the installer.
- Replaces approach in https://github.com/chromebrew/chromebrew/pull/9473

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=crew_add_git-restore-mtime crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -386,6 +386,7 @@ def update
 
     system "git fetch #{CREW_REPO} #{CREW_BRANCH}", exception: true
     system 'git reset --hard FETCH_HEAD', exception: true
+    system 'git-restore-mtime -sq 2>/dev/null', exception: true if File.file?("#{CREW_PREFIX}/bin/git-restore-mtime")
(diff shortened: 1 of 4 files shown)


uberhacker (approved):
Good solution.

PR #9484, opened 2024-03-14
Add 6.1 builds to linuxheaders

Zopolis4:
Milestone 122 builds are running kernel 6.1, so...

Tested and working on `x86_64` and `armv7l`.

The `rsync` build dep is because the build, at least on `6.1` and `x86_64`, fails without it.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=6sandwich crew update
```


Zopolis4:
Also the postinstall is definitely overmatching, it threw errors both times I ran this despite the installed headers matching the current kernel version.

Not sure if there's any purpose to it or if it makes more sense to just rip it out, but thats not exactly in the scope of this PR.

satmandu (approved):

satmandu (approved):

PR #9490, opened 2024-03-15
add neessesary packages.

yeesterPlus:
Adding etcher packages

### DIFF
--- a/packages/core.rb
+++ b/packages/core.rb
@@ -8,7 +8,18 @@ class Core < Package
   compatibility 'all'
 
   is_fake
-
+  depends_on 'gtk3'
+  depends_on 'freetype'
+  depends_on 'cairo'
+  depends_on 'xzutils'
+  depends_on 'libnotify'
+  depends_on 'nspr'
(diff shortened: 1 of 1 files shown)


packages/core.rb
```diff
@@ -8,7 +8,18 @@ class Core < Package
...
+  depends_on 'cairo'
+  depends_on 'xzutils'
+  depends_on 'libnotify'
+  depends_on 'nspr'
+  depends_on 'libgconf'
+  depends_on 'libxss'
+  depends_on 'libcom_err'
+  depends_on 'gsettings_desktop_schemas'
+  depends_on 'xhost'
+  depends_on 'sommelier'
```
uberhacker, line 22: Including this in core is going to add a ton of additional dependencies that would require a significant amount of free disk space.  Some users may not want gui apps due to space limitations.

Zopolis4:
These dependencies are already in `balena_etcher.rb`, they should not be in core. If you are having an issue with installing etcher or the dependencies or something else feel free to open an issue.

PR #9493, opened 2024-03-16
Remove activesupport gem package from core, adjust installer to insta…

satmandu:
- The installer here succeeds to the point that `core` installs, so remove ruby_activesupport from core.


uberhacker (approved):
All this because Ruby doesn't ship with `.blank` yet?

PR #9496, opened 2024-03-16
readd old packages

yeesterPlus:
just in case people want old stuff


satmandu:
We should add some sort of flag or message that points out that these are outdated and cannot be updated.

yeesterPlus:
I'll make it echo in build script

satmandu:
Yes but I mean add some code to crew so that we know not to bother trying to update packages that have been deprecated, and perhaps mention that to people as well who install them.

@supechicken666 @Zopolis4 @uberhacker Thoughts?

satmandu:
Also, you can run `rubocop -c .rubocop.yml -A packages/<package>.rb` to help make sure a package file passes the ruby linter.

yeesterPlus:
added deprecation support

Zopolis4:
We ship a number of packages that are either explicitly deprecated upstream or haven't been updated in decades-- I see no problem with adding packages that may not get updated. 

In fact, I'd argue having a deprecation flag will lead to situations where it is not clear whether a package has been deprecated, and thus we might accidentally mark it as deprecated and thus miss an update.

Zopolis4:
In regards to the actual packages, we can convert these to meson buildystem and alphabetize the dependencies at a quick glance.

yeesterPlus:
when deprecating upstream, add flag!

satmandu:
@Zopolis4 This is for packages I just deprecated because either they or underlying dependencies can no longer be updated due to for instance dependence on systems.

As such, we need a way of flagging these packages so we don't bother trying to update them to later versions.

Some sort of "Deprecated: Newer versions are no longer compatible with Chromebrew due to dependence on systemd." text might be sufficient?

satmandu:
@yeesterPlus Could you just add a line to these to files at the top saying:
```ruby
# Deprecated: Newer versions are no longer compatible with Chromebrew due to dependence on systems.
```
I literally think that's all we need...
@uberhacker Do you have anything you want to add to keep from being bugged by your script about updating these packages?

yeesterPlus:
Y E S

yeesterPlus:
I did it, by using a flag check `obj.flag?`.

satmandu:
Just take out the crew modifications.

packages/gnome_settings_daemon.rb
```diff
@@ -0,0 +1,74 @@
...
+  description 'GNOME Settings Daemon'
+  homepage 'https://gitlab.gnome.org/GNOME/gnome-settings-daemon'
+  version '43.0'
+  license 'GPL-2+ and LGPL-2+'
+  compatibility 'x86_64 aarch64 armv7l' # not compatible with i686 upstream
+  source_url 'https://gitlab.gnome.org/GNOME/gnome-settings-daemon.git'
+  git_hashtag version
+  binary_compression 'tar.zst'
+
+  deprecated
```
satmandu, line 15: Just take this out.

packages/gnome_tweaks.rb
```diff
@@ -0,0 +1,30 @@
...
+  homepage 'https://wiki.gnome.org/Apps/Tweaks'
+  @_ver = '42.0'
+  version "#{@_ver}-py3.12"
+  license 'GPL-3+ and CC0-1.0'
+  compatibility 'x86_64 aarch64 armv7l'
+  source_url 'https://gitlab.gnome.org/GNOME/gnome-tweaks.git'
+  git_hashtag @_ver
+  binary_compression 'tar.zst'
+  
+  deprecated
```
satmandu, line 16: Just take this out too.

bin/crew
```diff
@@ -1328,6 +1328,14 @@ def install
...
     return
   end
+    
+  if @pkg.deprecated?
+    puts "Package #{@pkg.name} is deprecated".orange
+    if @pkg.in_upgrade
+       puts "\b, skipping...".orange
+       return
+    end
+  end
```
satmandu, line 1338: Just take this out... We would want to add a boolean for this in `lib/package.rb` in any case before we did this... but I don't think marking this with a deprecated flag when some larger proportion of our software might be abandoned upstream...

packages/gnome_settings_daemon.rb
uberhacker: To explain this a bit better, we currently don't have a `depecated` property so not sure why this was added.

satmandu (changes requested):
Once those bits are taken out, I think this is fine.

packages/gnome_settings_daemon.rb
```diff
@@ -0,0 +1,74 @@
+# Deprecated: Newer versions are no longer compatible with Chromebrew due to dependence on systemd.
+
+require 'package'
+
+class Gnome_settings_daemon < Package
```
uberhacker, line 5: Use `< Meson` instead of `< Package`?  This will require `require 'buildsystems/meson'` above as well.

packages/gnome_settings_daemon.rb
satmandu: It was part of the discussion.... But agree as per discussion it isn't needed at this time.

packages/gnome_settings_daemon.rb
```diff
@@ -0,0 +1,74 @@
...
+  depends_on 'nss' # R
+  depends_on 'pango' # R
+  depends_on 'pulseaudio' # R
+  depends_on 'wayland' # R
+
+  def self.build
+    system "meson setup #{CREW_MESON_OPTIONS} \
+    -Dsystemd=false \
+    -Dcolord=false \
+    builddir"
```
uberhacker, line 66: The entire `def self.build` section can be replaced with `meson_options '-Dsystemd=false -Dcolord=false'`

bin/crew
```diff
@@ -1328,6 +1328,14 @@ def install
     puts "Package #{@pkg.name} already installed, skipping...".lightgreen
     return
   end
+    
+  if @pkg.deprecated?
```
uberhacker, line 1332: We'll need to add this property in `lib/package.rb` in order to use it.

uberhacker: Also, the crew version should be bumped in `lib/const.rb`.

uberhacker:
Thank you for the new packages @yeesterPlus!  Please don't take these criticisms the wrong way.  We appreciate your contribution!

yeesterPlus:
Done

yeesterPlus:
Stupid github!

packages/gnome_tweaks.rb
yeesterPlus: support addel

lib/package.rb
```diff
@@ -12,7 +12,7 @@ class Package
   boolean_property :arch_flags_override, :conflicts_ok, :git_clone_deep, :git_fetchtags, :gnome, :is_fake, :is_musl, :is_static,
                    :no_compile_needed, :no_compress, :no_env_options, :no_fhs, :no_git_submodules,
                    :no_links, :no_lto, :no_patchelf, :no_shrink, :no_source_build, :no_strip, :no_zstd, :patchelf,
-                   :print_source_bashrc, :run_tests
+                   :print_source_bashrc, :run_tests, :deprecated
```
uberhacker, line 15: Can we add this in alphabetical order?  The reason is to avoid potential duplicates.

satmandu, line 15: Just put all the boolean properties in alphabetical order again.

packages/gnome_settings_daemon.rb
```diff
@@ -0,0 +1,67 @@
...
+  depends_on 'pango' # R
+  depends_on 'pulseaudio' # R
+  depends_on 'wayland' # R
+
+  meson_options '-Dsystemd=false -Dcolord=false'
+
+  def self.install
+    system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
+  end
+end
```
satmandu, line 67: With `buildsystems/meson`, the standard self.install section isn't needed any more, unless it is different from the default, so you can take this entire section out.

packages/gnome_settings_daemon.rb
```diff
@@ -0,0 +1,67 @@
...
+  depends_on 'nss' # R
+  depends_on 'pango' # R
+  depends_on 'pulseaudio' # R
+  depends_on 'wayland' # R
+
+  meson_options '-Dsystemd=false -Dcolord=false'
+
+  def self.install
+    system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
+  end
```
uberhacker, line 66: The entire `def self.install` section can be removed too.

packages/gnome_settings_daemon.rb
satmandu: (The last `end` needs to stay there.)

bin/crew
```diff
@@ -1328,6 +1328,14 @@ def install
...
+    puts "Package #{@pkg.name} is deprecated".orange
+    if @pkg.in_upgrade
+       puts "\b, skipping...".orange
```
uberhacker, line 1335: Why the backspace?

packages/gnome_settings_daemon.rb
yeesterPlus: Added to packages.rb

packages/gnome_tweaks.rb
yeesterPlus: NO

yeesterPlus: I'll review your pull requests

satmandu:
Run rubocop on crew too...

bin/crew
yeesterPlus: to remove the new line

uberhacker: You can use `print` instead of `puts` to avoid the new line at the end.

yeesterPlus:
<details><summary>[REDACTED]</summary>
merged it w/ mine! Closed!
<details>

PR #9497, opened 2024-03-16
Fix i686 install

satmandu:
Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -160,10 +160,10 @@ find "${CREW_LIB_PATH}" -mindepth 1 -delete
 # Download the chromebrew repository.
 curl -L --progress-bar https://github.com/"${OWNER}"/"${REPO}"/tarball/"${BRANCH}" | tar -xz --strip-components=1 -C "${CREW_LIB_PATH}"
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -282,6 +282,12 @@ export LD_LIBRARY_PATH="${CREW_PREFIX}/lib${LIB_SUFFIX}:/lib${LIB_SUFFIX}"
 
 echo "export CREW_PREFIX=${CREW_PREFIX}" >> "${CREW_PREFIX}/etc/env.d/profile"
 
+# Install activesupport gem for ruby
+echo_info 'Installing essential ruby gems.'
+gem update -N --system
+gem install -N activesupport --conservative
+gem install -N concurrent-ruby --conservative
```
uberhacker, line 289: Doesn't this assume Ruby is installed on the host machine?

Zopolis4: We've got crew at this point, so we've got ruby.

satmandu: This is just before we run crew, so yes we have ruby at this point. This PR basically adds the deps necessary to get gem updating working as well at this point.

uberhacker (approved):

PR #9501, opened 2024-03-18
Add check command

uberhacker:
Tested and working.  Version check inspired by [version.rb](https://github.com/Zopolis4/chromebrew/blob/vcheck/tools/version.rb).  Fixed a bug with `crew search -v` and provided better error handling for missing or invalid command arguments.

### DIFF
--- a/tools/version.rb
+++ b/tools/version.rb
@@ -1,0 +1,151 @@
+#!/usr/bin/env ruby
+
+if ARGV.include?('-h') || ARGV.include?('--help')
+  abort "Usage: ./version.rb [<package>] [-h, --help, -v, --verbose]
+Example: ./version.rb abcde -v
+The <package> can contain '*': ./version.rb xorg_*
+If <package> is omitted, all packages will be checked."
+end
+
+require 'json'
+require 'net/http'
+
+# Add >LOCAL< lib to LOAD_PATH
+$LOAD_PATH.unshift '../lib'
+
+require_relative '../lib/color'
+require_relative '../lib/package'
+
+def get_version(name, homepage)
+  anitya_id = get_anitya_id(name, homepage)
+  # If we weren't able to get an Anitya ID, return early here to save time and headaches
+  return '' if anitya_id.nil?
+  # Get the latest version of the package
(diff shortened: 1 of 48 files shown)


Zopolis4:
`version.rb` is unfinished (still should be mostly functional) and this is not how I would have implemented things-- happy to let this get merged and used until I sort things out, though.

uberhacker:
> `version.rb` is unfinished (still should be mostly functional) and this is not how I would have implemented things-- happy to let this get merged and used until I sort things out, though.

How would you implement this then?  What do you need to sort out?

Zopolis4:
I probably wouldn't have a check command at all-- both of these files are in `tools` for a reason, and especially `prop_test`, which is already part of our CI. The version of `version.rb` you have is missing the libversion integration (the one on my `vcheck` branch has that), and I haven't finished working on properly excluding packages or handling missing packages. 

Again, though, I'm happy for this to be merged. I'm not going to hold up a feature from being merged now just because I plan to improve it in the future.

tools/version.rb
```diff
@@ -0,0 +1,190 @@
...
+  'gtksourceview_3',
+  'gtksourceview_4',
+  'gtksourceview_5',
+  'imagemagick6',
+  'imagemagick7',
+  'jdk11',
+  'jdk17',
+  'jdk18',
+  'jdk8',
+  'js102',
```
satmandu, line 118: Add js115?

satmandu (approved):

tools/version.rb
uberhacker: Replaced list with `no_upstream_update` property.

Zopolis4:
These changes to prop_test are really slowing down CI (see some of the runs on #9514)-- I'm not entirely sure why, and it only seems to be happening sometimes.

uberhacker:
> These changes to prop_test are really slowing down CI (see some of the runs on #9514)-- I'm not entirely sure why, and it only seems to be happening sometimes.

There are over 2k packages now.  The last run took only a min so not too bad...faster than I would have expected.

PR #9504, opened 2024-03-18
Add squid package

Zopolis4:
https://www.squid-cache.org/

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=coilingtendrils crew update
```


packages/squid.rb
```diff
@@ -0,0 +1,39 @@
+require 'buildsystems/autotools'
+
+class Squid < Autotools
+  description 'Squid Web Proxy Cache'
+  homepage 'https://github.com/squid-cache/squid'
```
uberhacker, line 5: Better home page? https://www.squid-cache.org

packages/squid.rb
```diff
@@ -0,0 +1,39 @@
...
+  version '6.8'
+  license 'GPL-2.0'
+  compatibility 'all'
+  source_url 'https://www.squid-cache.org/Versions/v6/squid-6.8.tar.xz'
```
uberhacker, line 9:
```
source_url 'https://github.com/squid-cache/squid.git'
git_hashtag 'SQUID_6_8'
```
?

packages/squid.rb
```diff
@@ -0,0 +1,39 @@
...
+  depends_on 'libxml2' # R
+  depends_on 'linux_pam' # R
+  depends_on 'nettle' # R
+  depends_on 'openldap' # R
+  depends_on 'glibc' # R
+
+  # https://github.com/squid-cache/squid/pull/1118
+  pre_configure_options "CXXFLAGS='-Wno-alloc-size-larger-than'"
+  configure_options '--disable-arch-native --without-systemd'
+end
```
uberhacker, line 39: Add to `tools/packages.yaml`?

uberhacker:
Thank you for the new package @Zopolis4!  See comments for suggestions.

packages/squid.rb
Zopolis4: It's faster to download tarfiles than to download from git

uberhacker: A git clone is a one-time build from source and you don't need to calculate the sha256sum hash.

Zopolis4: My apologies, I timed both of them and downloading from git is indeed quite a bit faster.

uberhacker: Good to know but there are certainly times where it is slower to clone the repo.

satmandu:
(script....)

Zopolis4:
I'm not sure-- at any rate, is it not possible for us to at least run an init manager as user for basic startup scripts? Rolling them ourselves seems unnecessary in the majority of cases (bar `sommelier`, where there is no upstream script)

satmandu:
> I'm not sure-- at any rate, is it not possible for us to at least run an init manager as user for basic startup scripts? Rolling them ourselves seems unnecessary in the majority of cases (bar `sommelier`, where there is no upstream script)

It's a huge pain because the first crosh process we open isn't PID 1, which systemd needs.

Zopolis4:
Oh, I was thinking of using a systemd alternative-- I think there are a few which can handle the service management and also don't require PID 1.

satmandu (approved):

PR #9505, opened 2024-03-19
Set default 'binary_compression' to 'tar.zst', allow verbose option for all crew commands.

satmandu:
- Set a sane default for 'binary_compression' so we don't need it in package files.
- Also allows for verbose mode for other crew options.

Before: 
image
After:
image

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=binary_compression_optional crew update
```


satmandu:
This code in `package.rb` doesn't seem to work... 

```ruby
  def self.get_binary_url(architecture)
    architecture = 'armv7l' if architecture == 'aarch64'
    puts "binary_compression is #{binary_compression}"
    puts "binary_compression.nil? is #{binary_compression.nil?}"
    binary_compression = binary_compression.nil? ? 'tar.zst' : binary_compression
    puts "binary_compression is #{binary_compression}"
    return "https://gitlab.com/api/v4/projects/26210301/packages/generic/#{name}/#{version}_#{architecture}/#{name}-#{version}-chromeos-#{architecture}.#{binary_compression}"
  end
```

```bash
$ crew install procps
procps: procps is a set of command line and full-screen utilities that provide information out of the pseudo-filesystem most commonly located at /proc.
https://gitlab.com/procps-ng/procps
Version: 3.3.16
License: GPL-2
binary_compression is tar.xz
binary_compression.nil? is false
binary_compression is tar.zst
Cannot find cached archive. 😔 Will download.
Download failed with error 404: Not Found
```

Zopolis4:
Couldn't we just check for `no_zstd`? 

At any rate, once all the packages are either `.tar.xz` (because of no_zstd) or `.tar.zst`, we wont need the binary_compression value.

`crew upload` also relies on a binary_compression value, so you might want to update that as well.

satmandu:
> Couldn't we just check for `no_zstd`? 
> 
No because older packages not packaged with zstd are still packaged without zstd.

In any case, the code isn't working, and I can't figure out why. The ternary for some reason sets `binary_compression` to `nil`.

uberhacker (approved):

PR #9510, opened 2024-03-20
Add pre-commit hooks

Zopolis4:
Add pre-commit hooks via https://pre-commit.com/.

Also add `py3-pre-commit` package.

The changes to `.md` files and `caja.rb` are the pre-commit hooks in action.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=clawed crew update
```


packages/py3_pre_commit.rb
```diff
@@ -0,0 +1,15 @@
...
+  homepage 'https://pre-commit.com/'
+  @_ver = '3.6.2'
+  version "#{@_ver}-py3.12"
+  license 'MIT'
+  compatibility 'all'
+  source_url 'SKIP'
+
+  depends_on 'python3'
+
+  no_compile_needed
```
uberhacker, line 14: Add  postinstall  to execute  pre-commit?
```
def self.postinstall
  system 'pre-commit install'
end
```
Likewise, it might make sense to add `def self.remove` to clean up the pre-commit hooks installed locally unless they can be included with this package?

Zopolis4: We can't ship pre-commit hooks, and you have to run pre-commit install in the directory of a git repo.

packages/buildessential.rb
```diff
@@ -90,9 +90,6 @@ class Buildessential < Package
   # depends_on 'openjdk11'
   # depends_on 'openjdk17'
 
-  # Linter
-  depends_on 'linter'
-
```
uberhacker, line 95: Might be good to replace linter with py3_pre_commit.

packages/linter.rb
```diff
@@ -1,91 +0,0 @@
...
-    File.write('.mdl_style.rb', mdl_style)
-  end
-
-  def self.install
-    FileUtils.mkdir_p CREW_DEST_HOME.to_s
-    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
-    FileUtils.install 'linter', "#{CREW_DEST_PREFIX}/bin", mode: 0o755
-    FileUtils.install ['.mdlrc', '.mdl_style.rb'], CREW_DEST_HOME.to_s, mode: 0o644
-  end
-end
```
uberhacker, line 91: All that work is now thrown away?  That makes me sad. :crying_cat_face:

bin/crew
```diff
@@ -1736,10 +1736,6 @@ def upload(pkg_name = nil)
           system "sed -i 's/#{old_sha256}/#{new_sha256}/g' #{pkg_file}"
         end
       end
-      # Use rubocop to sanitize package file, and let errors get flagged.
-      system 'yes | crew install ruby_rubocop' unless @device[:installed_packages].any? { |elem| elem[:name] == 'ruby_rubocop' }
-      puts "Using rubocop to sanitize #{pkg_file} .".orange
-      system "rubocop -c #{File.join(CREW_LOCAL_REPO_ROOT, '.rubocop.yml')} -A #{pkg_file}", exception: true
```
uberhacker, line 1742: This is not a bad thing.

packages/linter.rb
Zopolis4: We've still got a bunch of the equivalent code in our CI, and at any rate the linter package was long out of date.

uberhacker: The nice thing about this script is it's a one-stop-shop for all things lint.

bin/crew
Zopolis4: We shouldn't be doing this in `crew`-- odds are this package file has been commited to a repo for us to get here, so it'll have been linted. If it hasn't been commited to a repo yet, it'll get flagged when it is.

packages/buildessential.rb
Zopolis4: `py3_pre_commit` is only really useful to chromebrew developers-- no need to install it for users.

packages/linter.rb
```diff
@@ -1,91 +0,0 @@
...
-            shellcheck "$file"
-            ;;
-          yml|yaml)
-            yaml-lint "$file"
-            ;;
-          *)
-            echo "Unable to check syntax of $file."
-            ;;
-        esac
-        codespell "$file"
```
uberhacker, line 61: This goes away too with this script.

packages/linter.rb
Zopolis4: It's out of date with our linting configs-- I don't see any value in shipping a package for our convenience when it doesn't do its job properly anymore (and its job is done better by the pre-commit hooks).

packages/linter.rb
Zopolis4: We haven't used codespell since 2022... #7542

packages/buildessential.rb
uberhacker: The entire essence of the `buildessential` package is for those who want to build from source and check packages.

packages/py3_pre_commit.rb
uberhacker: Why not?  What is the harm in doing so?  Nobody is going to know they need to run `pre-commit install`.  Might as well do it for them because, otherwise, this package is useless.  The linter package was created to make it EASIER for those who want to lint stuff.  This makes it harder.  One of my big pet peeves about changing things.

packages/buildessential.rb
Zopolis4: The `buildessential` packages are automatically installed if we build a package from source-- there are instances where a user needs to build a package from source but is not planning to contribute it to chromebrew (packages missing binaries, packages requiring local rebuilds for glibc reasons, etc)

packages/linter.rb
uberhacker: Tell me exactly how you think this is out of date and what specifically it is not doing correctly?

packages/py3_pre_commit.rb
Zopolis4: running `pre-commit install` like this just doesn't work. thats not how the package works.

you run `pre-commit install` in a specific git directory when you want to set up the pre-commit hooks for that project.

And even still-- nobody is going to be installing this package without knowing to run `pre-commit install` in the relevant directory. If they're installing the `py3-pre-commit` package, they know what it does. This is a developer tool-- trying to guess what the user is aiming to do and doing it for them is not a good idea.

packages/linter.rb
uberhacker: Well, we should start using it again.  I've seen a lot of poorly spelled comments.  Where do you get these ideas about ditching stuff?  FYI, this is the linter script, not to be run in Github Actions but rather, locally.

packages/linter.rb
Zopolis4: The markdownlint configuration is out-of-date, and would in fact override our current one due to the way it is set up.

packages/linter.rb
Zopolis4: You're welcome to write the codespell CI and fix the issues that led to you removing it last time...

> FYI, this is the linter script, not to be run in Github Actions but rather, locally.

What is the actual usecase for this script? If its for developer convenience, the pre-commit hooks replace that.

If its for users, it probably shouldn't be-- why would we ship a package which runs four random linters, one of them with an opinionated configuration, in the current (git) directory?

packages/py3_pre_commit.rb
uberhacker: The `CREW_LOCAL_REPO_ROOT` constant takes the guesswork out of the equation.

uberhacker: Nobody is going to understand that they need to run `pre-commit install` unless you explain it to them.  Again, try and make things easier, not harder.

Zopolis4: But then we're assuming that users are installing this for the purpose of contributing to crew-- if they arent, we've installed those hooks (and the linters they use!) for no reason.

Zopolis4: The fundamental point of what I'm saying is that we shouldn't assume what a user wants to do and then do it for them.

packages/linter.rb
uberhacker: The pre-commit hooks are not convenient.  For one, the package is not installed with buildessential and even if it was, there is no explanation of the need to run `pre-commit install` to make it useful other than a comment that nobody will read unless they look in the package.  That should be at the very least a postinstall message.

packages/py3_pre_commit.rb
Zopolis4: > Nobody is going to understand that they need to run `pre-commit install` unless you explain it to them. Again, try and make things easier, not harder.

Nobody is going to install the `py3-pre-commit` package without knowing what it does, and if they don't know and want to find out, they can use the description, or the homepage, or google.

We don't do this for any other package-- and neither do any other package managers.

Why should we do it for this one?

packages/linter.rb
uberhacker: There is nothing random or opinionated about these linters.  You are just like my wife who finds it easy to throw away things that belong to others. :scream_cat:

Zopolis4: The `linter` package shouldn't be a package at all-- this is not a program available anywhere else, this is an out-of-date convenience script. If you find this convenient, keep it on your local machine.

uberhacker: You mean just like you added 2 more opinionated exclusions to the GitHub Actions rules?  It doesn't hurt to keep this package.

packages/py3_pre_commit.rb
uberhacker: We're not doing anything for them other than aiding the process.  Again, this package is USELESS without running `pre-commit install`.  And we don't have to include it in buildessential.  We've established that, at least.

packages/linter.rb
Zopolis4: > It doesn't hurt to keep this package.

Again, this is an outdated convenience script. This should not be a package.

packages/py3_pre_commit.rb
Zopolis4: > We're not doing anything for them other than aiding the process.

No, we're assuming that they want to have linters installed in `CREW_LOCAL_REPO_ROOT` and doing that automatically for them. That's not aiding the process, that's at best doing something that they didn't ask for and don't understand, and at worst doing something completely unnecessary that runs counter to what they were trying to do.

> Again, this package is USELESS without running `pre-commit install`.

What exactly do you mean useless? This package provides a tool-- it is not useless just because we don't automatically use it.

bin/crew
satmandu: Slow on the uptake here. Do the pre-commit hooks allow for `rubocop -c .rubocop -A #{pkg_file}` to happen in some other manner before the commit happens?

I don't care how it happens. I just want it to happen.

Zopolis4: Yup! That's exactly what they do.

Zopolis4: You'll have to set them up
locally first by runnng pre-commit install, btw.

Zopolis4:
As per discussions on slack, I am keeping the code to run `rubocop` in `crew upload`, as satmandu uses it in his workflow and it makes contributing easier for new contribors.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -24,7 +24,7 @@ Works properly:
 - [ ] `armv7l` <!-- (reasons why it doesn't) -->
 
 ### Run the following to get this pull request's changes locally for testing.
-```
+```text
```
satmandu, line 27: This should actually be **```bash**

satmandu: That way stuff like this works:
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=distcc crew update ; yes | crew install distcc
```

vs
```text
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=distcc crew update ; yes | crew install distcc
```

satmandu: (bash gives the correct color highlighting.)

satmandu (dismissed):

Zopolis4:
I've added a section to `CONTRIBUTING.md` on how to install the pre-commit hooks.

PR #9514, opened 2024-03-21
Split prop command into separate file

Zopolis4:
**Split out `prop` command to `commands/prop.rb`.**

Behavior changes:
- The functionality of `crew help prop <property>` has been replaced by `crew prop <property>`
  - `crew help prop` now returns a description of what the `crew prop` command does, in line with the other commands.
- Add support for `crew prop no_upstream_update`.
  
Refactoring changes:
- Make `crew prop` a real command that actually does something.
- Remove the random verbose argument to `crew prop`.
- Have `commands/prop.rb` handle what was previously handled by `help_prop` in `commands/help.rb`.

Testing changes:
- Add tests for `crew prop` in `tests/commands/prop.rb`.
- Update the `crew help` tests to reflect the new behavior of `crew help`.
- Add testing for the new behaviour `crew help prop` in `tests/commands/help.rb` specifically.

Tested and working on `x86_64`, and all tests pass.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula4 crew update
```


commands/prop.rb
```diff
@@ -0,0 +1,53 @@
...
+    when 'no_patchelf'
+      puts "Use the 'no_patchelf' property to bypass patchelf execution."
+    when 'no_shrink'
+      puts "Use the 'no_shrink' property to bypass upx binary compression."
+    when 'no_source_build'
+      puts "Use the 'no_source_build' property to build even if source is unspecified or unavailable."
+    when 'no_strip'
+      puts "Use the 'no_strip' property to bypass strip execution."
+    when 'no_lto'
+      puts "Use the 'no_lto' property to bypass lto usage."
```
uberhacker, line 35: This is my fault but we need to add:
```
when 'no_upstream_update'
  puts "Use the 'no_upstream_update' property for packages that cannot be updated from an upstream repository."
```

commands/prop.rb
```diff
@@ -0,0 +1,53 @@
...
+      puts "Use the 'patchelf' property for patchelf execution."
+    when 'print_source_bashrc'
+      puts "Prints 'source ~/.bashrc' message at end of install."
+    when 'run_tests'
+      puts "Use the 'run_tests' property to execute make check tests."
+    else
+      puts <<~EOT
+        Explain the purpose of various package boolean properties.
+        Usage: crew prop <property>
+        Available properties: conflicts_ok, git_fetchtags, is_fake, is_musl, is_static, no_compile_needed, no_compress, no_env_options, no_fhs, no_links, no_patchelf, no_shrink, no_source_build, no_strip, no_lto, no_zstd, patchelf, print_source_bashrc, run_tests
```
uberhacker, line 48: Add `no_upstream_update` here too.

tests/commands/prop.rb
```diff
@@ -0,0 +1,32 @@
+require 'minitest/autorun'
+require_relative '../../commands/prop'
+
+class PropCommandTest < Minitest::Test
+  def test_no_arguments
+    expected_output = <<~EOT
+      Explain the purpose of various package boolean properties.
+      Usage: crew prop <property>
+      Available properties: conflicts_ok, git_fetchtags, is_fake, is_musl, is_static, no_compile_needed, no_compress, no_env_options, no_fhs, no_links, no_patchelf, no_shrink, no_source_build, no_strip, no_lto, no_zstd, patchelf, print_source_bashrc, run_tests
```
uberhacker, line 9: Add `no_upstream_update` here too.

commands/help.rb
```diff
@@ -75,7 +75,11 @@ def self.help(command, property)
         The package(s) must be currently installed.
       EOT
     when 'prop'
-      help_prop(property)
+      puts <<~EOT
+        Explain the purpose of various package boolean properties.
+        Usage: crew prop <property>
+        Available properties: conflicts_ok, git_fetchtags, is_fake, is_musl, is_static, no_compile_needed, no_compress, no_env_options, no_fhs, no_links, no_patchelf, no_shrink, no_source_build, no_strip, no_lto, no_zstd, patchelf, print_source_bashrc, run_tests
```
uberhacker, line 81: Add `no_upstream_update` here too.

commands/prop.rb
Zopolis4: On it.

Zopolis4:
Added testing for the new behavior of `crew help prop` specifically, as we've got to change that in a number of places, so best to test it in all those places.

Zopolis4:
I don't know what's going on with those unit test failures, but they certainly aren't related to this PR.

The command tests pass, btw.
image

tests/commands/prop.rb
```diff
@@ -0,0 +1,32 @@
+require 'minitest/autorun'
+require_relative '../../commands/prop'
+
+class PropCommandTest < Minitest::Test
+  def test_no_arguments
+    expected_output = <<~EOT
+      Explain the purpose of various package boolean properties.
+      Usage: crew prop <property>
+      Available properties: conflicts_ok, git_fetchtags, is_fake, is_musl, is_static, no_compile_needed, no_compress, no_env_options, no_fhs, no_links, no_patchelf, no_shrink, no_source_build, no_strip, no_lto, no_upstream_update, no_zstd, patchelf, print_source_bashrc, run_tests
```
uberhacker, line 9: I'm wondering if it makes sense to calculate based on what exists add the list of properties to a variable or constant and then reuse that variable here instead?  I only say that because it's being used in 4 different places now.

bin/crew
```diff
@@ -359,22 +360,6 @@ def files(pkg_name)
...
 end
 
-def prop(silent = false)
-  props = []
-  pkg = Package.new
-  excluded_methods = %w[compatible binary source json_creatable autoload include const_defined class_variable_defined singleton_class method_defined public_method_defined private_method_defined protected_method_defined instance_variable_defined instance_of kind_of is_a frozen nil eql respond_to equal]
-  all_methods = pkg.class.methods.grep(/\?$/).to_s.gsub(/([?:,\[\]])/, '').split
-  all_methods.each do |method|
-    props.push(method) unless excluded_methods.include?(method)
-  end
```
uberhacker, line 369: Add this to `lib/const.rb` to be used in 4 different places?

tests/commands/prop.rb
Zopolis4: I looked into that-- the code for doing it is pretty ugly, and it also has the problem that we actually don't have documentation for a lot of those properties :sweat_smile:

It'd also mean we need to either duplicate the code, or add more dependencies, or pass around another argument.

I agree the current situation is not ideal, but when we add a new property we already have to manually add it to a list in `lib/package.rb`, a few more lists isn't too big of a deal.

bin/crew
uberhacker: How about this?
```
... calculated as above ...
CREW_BOOLEAN_PROPERTIES = props.join(', ')
```

Zopolis4:
(A re-run fixed the unit tests, probably not great that we have a non-deterministic CI or that we aren't able to gracefully handle a terminated download, but thats a problem for another time)

tests/commands/prop.rb
uberhacker: Remember the DRY method.  Don't Repeat Yourself.  It's a cardinal sin for developers.

bin/crew
Zopolis4: The whole point of this is to make `bin/crew` smaller and simpler by splitting it up into standalone (as much as possible) files-- doing this would work, but it would add `lib/package.rb` and `lib/const.rb` dependencies to `commands/prop.rb` and `commands/help.rb` and their tests, not to mention adding more complexity in what is supposed to be a simple list of static constants for easy reuse.

I think the downsides of having to manually update the property list are small (we aren't adding new properties very often, and when we do add them there's still manual work involved in doing so) compared to the penalties this incurs in code complexity and increased inter-dependency.

If you very strongly disagree, or if both you and @satmandu think this is the better path foward, that's fine (enough) by me.

tests/commands/prop.rb
uberhacker: Added benefit: New properties are added to the list automagically.

Zopolis4: Probably not amazing if we add properties to the list independent of writing documentation-- I can see a scenario in which someone forgets to add documentation for a property but `crew` thinks we have documentation (In fact, this is the case right now with `git_clone_deep`, `arch_flags_override` and `gnome` (off the top of my head), so its not an imaginary concern.

(right now this PR doesn't have that issue-- it only says it has documentation for the properties we have documentation for-- master has this issue, and implementing things like this would cause the issue)

I think it probably makes sense for people to be adding to this list manually, at the same time they're adding the documentation for the property.

bin/crew
uberhacker: I'm open to a better property calculation algorithm.

Zopolis4: The problem isn't this particular algorithm, its using one at all, unfortunately.

tests/commands/help.rb
```diff
@@ -8,25 +8,29 @@ def test_no_arguments
       Available commands: build, check, const, deps, download, files, help, install, license, list, postinstall, prop, reinstall, remove, search, sysinfo, test, update, upgrade, upload, version, whatprovides
```
uberhacker, line 8: I think this list could be calculated too.

Zopolis4: We add new commands even less frequently than we add new properties-- the case for doing this is weaker than the case for doing it for properties.

tests/commands/prop.rb
uberhacker: The fact that you have split out these files is making things more complicated simply by abstraction and it would be very easy to miss something new.  I don't think making this many manual changes repeatedly is good for anyone's sanity.

Zopolis4:
Would you be okay with merging this now and having the discussion about dynamically generating the list of properties (and maybe commands) at a later date?

This PR doesn't change the status quo on generation, so decisions about that shouldn't hold it up.

uberhacker:
> Would you be okay with merging this now and having the discussion about dynamically generating the list of properties (and maybe commands) at a later date?
> 
> This PR doesn't change the status quo on generation, so decisions about that shouldn't hold it up.

I will leave that decision up to @satmandu.

tests/commands/prop.rb
Zopolis4: It's one identical change that has to be done four times, whenever someone adds a new property (which has been happening once every month recently, and that's a faster pace than usual). A find and replace would get the job done. 

If you just really do not want to have to make these changes whenever you add a property, I'm happy to write a script that you can use locally to avoid doing so, or just make the changes for you myself whenever you add a property.

My goal here is not to make things more difficult for you, and I'm happy with any compromise that preserves code simplicity and reduces inter-dependencies.

satmandu: > It's one identical change that has to be done four times,

Please figure out a way to do this only once. I agree with @uberhacker that making an identical change four times is problematic because it can be missed. 

Figure out a way to automate the change automatically. I don't care it it goes in a config file or whatever. `crew` should be able to figure out the rest on its own.

Zopolis4: By design, the tests ensure that if the change is made in `commands/prop.rb` they will fail unless they are also updated, so it can't be missed.

Personally, I think making an identical change 4 times is not particularly difficult, especially when it does not occur very often.

Zopolis4: But the most important thing here is that, again, I do not think it is a good idea to make these changes automatically. 

That introduces the possibility of `crew help prop` reporting that it has documentation for properties that it currently does not. 

In master, we generate this list of properties automatically. And in master, this is wrong. Multiple properties are automatically added to the list despite not having documentation available. 

This PR fixes that bug-- I do not want to re-introduce it.

bin/crew
uberhacker: Not a problem, a solution.

tests/commands/help.rb
uberhacker: Not weak if it helps make sure tests don't fail when new properties are added and this test update was missed.

commands/help.rb
```diff
@@ -1,7 +1,7 @@
 require_relative '../lib/const'
```
uberhacker, line 1: Add?
```
```

tests/commands/help.rb
```diff
@@ -8,25 +8,31 @@ def test_no_arguments
...
 
-  def test_const_command
+  def test_prop_command
     expected_output = <<~EOT
-      Display constant(s).
-      Usage: crew const [<const1> <const2> ...]
-      If no constants are provided, all constants will be displayed.
+      Explain the purpose of various package boolean properties.
+      Usage: crew prop <property>
+      Available properties: conflicts_ok, git_fetchtags, is_fake, is_musl, is_static, no_compile_needed, no_compress, no_env_options, no_fhs, no_links, no_lto, no_patchelf, no_shrink, no_source_build, no_strip, no_upstream_update, no_zstd, patchelf, print_source_bashrc, run_tests
```
uberhacker, line 19: Replace with?
```
Available properties: #{Package.print_boolean_properties}
```

tests/commands/prop.rb
```diff
@@ -0,0 +1,32 @@
+require 'minitest/autorun'
+require_relative '../../commands/prop'
+
+class PropCommandTest < Minitest::Test
+  def test_no_arguments
+    expected_output = <<~EOT
+      Explain the purpose of various package boolean properties.
+      Usage: crew prop <property>
+      Available properties: conflicts_ok, git_fetchtags, is_fake, is_musl, is_static, no_compile_needed, no_compress, no_env_options, no_fhs, no_links, no_lto, no_patchelf, no_shrink, no_source_build, no_strip, no_upstream_update, no_zstd, patchelf, print_source_bashrc, run_tests
```
uberhacker, line 9: Replace with?
```
Available properties: #{Package.print_boolean_properties}
```

commands/prop.rb
```diff
@@ -0,0 +1,55 @@
...
+      puts "Use the 'patchelf' property for patchelf execution."
+    when 'print_source_bashrc'
+      puts "Prints 'source ~/.bashrc' message at end of install."
+    when 'run_tests'
+      puts "Use the 'run_tests' property to execute make check tests."
+    else
+      puts <<~EOT
+        Explain the purpose of various package boolean properties.
+        Usage: crew prop <property>
+        Available properties: conflicts_ok, git_fetchtags, is_fake, is_musl, is_static, no_compile_needed, no_compress, no_env_options, no_fhs, no_links, no_lto, no_patchelf, no_shrink, no_source_build, no_strip, no_upstream_update, no_zstd, patchelf, print_source_bashrc, run_tests
```
uberhacker, line 50: Replace with?
```
Available properties: #{Package.print_boolean_properties}
```

commands/help.rb
```diff
@@ -166,63 +170,3 @@ def self.help(command, property)
     end
   end
 end
-
-def help_prop(property)
-  if property
-    case property
```
uberhacker, line 172: Add?
```
when 'arch_flags_override'
  puts "Use 'arch_flags_override' to override architecture specific flags."
```

commands/help.rb
```diff
@@ -166,63 +170,3 @@ def self.help(command, property)
...
-    when 'conflicts_ok'
-      puts "The 'conflicts_ok' property bypasses checks for other package file conflicts."
-    when 'git_clone_deep'
-      puts "The 'git_clone_deep' property clones the repository without a depth value."
-      puts "Applicable only when 'source_url' is a git repository."
-    when 'git_fetchtags'
-      puts "The 'git_fetchtags' property gets the repository tags."
-      puts "Applicable only when 'source_url' is a git repository."
```
uberhacker, line 180: Add?
```
when 'gnome'
  puts "Use 'gnome' to include options for gnome specific packages."
```

commands/prop.rb
```diff
@@ -0,0 +1,55 @@
+class Command
+  def self.prop(property)
+    case property
```
uberhacker, line 3: Add?
```
when 'arch_flags_override'
  puts "Use 'arch_flags_override' to override architecture specific flags."
```

commands/prop.rb
```diff
@@ -0,0 +1,55 @@
...
+    when 'conflicts_ok'
+      puts "The 'conflicts_ok' property bypasses checks for other package file conflicts."
+    when 'git_fetchtags'
+      puts "The 'git_fetchtags' property gets the repository tags."
+      puts "Applicable only when 'source_url' is a git repository."
```
uberhacker, line 8: Add?
```
when 'gnome'
  puts "Use 'gnome' to include options for gnome specific packages."
```

commands/prop.rb
```diff
@@ -0,0 +1,55 @@
+class Command
+  def self.prop(property)
+    case property
+    when 'conflicts_ok'
+      puts "The 'conflicts_ok' property bypasses checks for other package file conflicts."
```
uberhacker, line 5: Add?
```
    when 'git_clone_deep'
      puts "The 'git_clone_deep' property clones the repository without a depth value."
      puts "Applicable only when 'source_url' is a git repository."
```

commands/prop.rb
```diff
@@ -0,0 +1,55 @@
+class Command
```
uberhacker, line 1: Add?
```
require 'package'
```

tests/commands/help.rb
```diff
@@ -8,25 +8,31 @@ def test_no_arguments
       Available commands: build, check, const, deps, download, files, help, install, license, list, postinstall, prop, reinstall, remove, search, sysinfo, test, update, upgrade, upload, version, whatprovides
```
uberhacker, line 8: Add?
```
require_relative '../../lib/package'
```

tests/commands/prop.rb
```diff
@@ -0,0 +1,32 @@
+require 'minitest/autorun'
+require_relative '../../commands/prop'
```
uberhacker, line 2: Add?
```
require_relative '../../lib/package'
```

commands/prop.rb
```diff
@@ -0,0 +1,55 @@
...
+    when 'no_patchelf'
+      puts "Use the 'no_patchelf' property to bypass patchelf execution."
+    when 'no_shrink'
+      puts "Use the 'no_shrink' property to bypass upx binary compression."
+    when 'no_source_build'
+      puts "Use the 'no_source_build' property to build even if source is unspecified or unavailable."
+    when 'no_strip'
+      puts "Use the 'no_strip' property to bypass strip execution."
+    when 'no_lto'
+      puts "Use the 'no_lto' property to bypass lto usage."
```
uberhacker, line 35: Move up to sort alphabetically?

commands/prop.rb
```diff
@@ -0,0 +1,55 @@
...
+      puts "Use the 'is_static' property for packages which do not require shared dependencies."
+    when 'no_compile_needed'
+      puts "Use the 'no_compile_needed' property for packages that do not require pre-built binaries."
+    when 'no_compress'
+      puts "Use the 'no_compress' property for packages that do not need compressed files."
+    when 'no_env_options'
+      puts "Use the 'no_env_options' property for packages that do not require"
+      puts 'environment options or to override the default options.'
+    when 'no_fhs'
+      puts "The 'no_fhs' property bypasses FHS3 compliance checks."
```
uberhacker, line 23: Add?
```
    when 'no_git_submodules'
      puts "Use the 'no_git_submodules' property for repositories without git submodules."
      puts "Applicable only when 'source_url' is a git repository."
```

commands/prop.rb
uberhacker: There should be (on last check) 24 properties so we should have that many help lines.

bin/crew
uberhacker: Resolved in #9519.

commands/prop.rb
```diff
@@ -0,0 +1,67 @@
...
+    when 'no_compile_needed'
+      puts "Use the 'no_compile_needed' property for packages that do not require pre-built binaries."
+    when 'no_compress'
+      puts "Use the 'no_compress' property for packages that do not need compressed files."
+    when 'no_env_options'
+      puts "Use the 'no_env_options' property for packages that do not require"
+      puts 'environment options or to override the default options.'
+    when 'no_git_submodules'
+      puts "Use the 'no_git_submodules' property for repositories without git submodules."
+      puts "Applicable only when 'source_url' is a git repository."
```
uberhacker, line 33: Put this below `when 'no_fhs'`?  It would be nice to keep all the properties in alphabetical order to prevent potential duplicate lines.

uberhacker (changes requested):

commands/prop.rb
Zopolis4: Got the alphabetical order wrong :sweat_smile:

commands/prop.rb
Zopolis4: Thanks for adding documentation for this and the other properties, as well as implementing a better way to get the list of boolean properties dynamically.

tests/commands/help.rb
Zopolis4: That's the whole point of tests-- to make sure we don't miss the updates.

commands/prop.rb
uberhacker: No problem.  I think we will all sleep better once this is done.

tests/commands/help.rb
uberhacker: We can address this on a future PR.

tests/commands/prop.rb
```diff
@@ -0,0 +1,32 @@
...
+
+class PropCommandTest < Minitest::Test
+  def test_no_arguments
+    expected_output = <<~EOT
+      Explain the purpose of various package boolean properties.
+      Usage: crew prop <property>
+      Available properties: arch_flags_override, conflicts_ok, git_clone_deep, git_fetchtags, gnome, is_fake, is_musl, is_static, no_compile_needed, no_compress, no_env_options, no_fhs, no_git_submodules, no_links, no_lto, no_patchelf, no_shrink, no_source_build, no_strip, no_upstream_update, no_zstd, patchelf, print_source_bashrc, run_tests
```
uberhacker, line 9: Replace with?
```
Available properties: #{Package.print_boolean_properties}
```

tests/commands/help.rb
```diff
@@ -8,25 +8,29 @@ def test_no_arguments
...
 
-  def test_const_command
+  def test_prop_command
     expected_output = <<~EOT
-      Display constant(s).
-      Usage: crew const [<const1> <const2> ...]
-      If no constants are provided, all constants will be displayed.
+      Explain the purpose of various package boolean properties.
+      Usage: crew prop <property>
+      Available properties: arch_flags_override, conflicts_ok, git_clone_deep, git_fetchtags, gnome, is_fake, is_musl, is_static, no_compile_needed, no_compress, no_env_options, no_fhs, no_git_submodules, no_links, no_lto, no_patchelf, no_shrink, no_source_build, no_strip, no_upstream_update, no_zstd, patchelf, print_source_bashrc, run_tests
```
uberhacker, line 19: Replace with?
```
Available properties: #{Package.print_boolean_properties}
```

tests/commands/prop.rb
Zopolis4: These tests are designed to ensure that things such as `Package.print_boolean_properties` do not break-- If we use the behavior of `Package.print_boolean_properties` as the expected behavior of itself, we have no way of knowing if it has broken.

uberhacker: And what if the test fails because the hardcoded list is out of date?  The argument goes both ways.  It is too difficult to manage duplicate content in multiple places; especially if it is buried several subdirectories deep like these test commands.  I would rather keep it dynamic.  We can have a separate test for `Package.print_boolean_properties` run right before this test if that is your concern.

Zopolis4: > And what if the test fails because the hardcoded list is out of date?

We update the test, and this will have been caught by CI.

> The argument goes both ways.

This is true, and I see your point-- I think testing this behavior is worth having to update the tests whenever we change the behavior, as this will be the case for any test of any sort.


> We can have a separate test for Package.print_boolean_properties run right before this test if that is your concern.

This is a good suggestion, but it'll still have the same structure of being hardcoded, which is the thing you're trying to avoid.

uberhacker: You could run a test as simple as counting the properties generated by `Package.print_boolean_properties` vs. the `when 'something'` lines in help.rb and fail if the counts don't match.  Still doesn't have to be hardcoded.

Zopolis4: I wrote the tests this way because they ensure maximum coverage-- I don't think its a good idea to have them catch less mistakes.

uberhacker: If you want to get fancy, you could compare all the `when 'something'` lines to actual properties generated by `Package.print_boolean_properties` and make the test more thorough.

Zopolis4: I'm not sure I entirely understand what you mean.

uberhacker: See #9531.  If this test fails, then you can know Package.print_boolean_properties is not generating the expected output.

Zopolis4: Its a start, but it still doesn't cover alphabetical order, properties being added and removed such that the number is the same but the contents are different, just at a glance.

Now we're relying on the behavior of a bunch of system calls and pipes to remain the same (pretty likely, but even still).

Zopolis4: At this point, you've put in more effort and time trying to avoid having to manually update two lines occasionally than you would have simply updating the lines :wink: 

This is a test for a command that prints out strings-- the best test is to hardcode those strings and check they are the same.

uberhacker: You are still missing the point and @satmandu agrees with me.  Since you separated out the files and abstracted them, it will be very easy for new properties to be missed if it was hardcoded.  Why not be smart about it?  Future updates will be easier.  Think outside the box.

Zopolis4: > Since you separated out the files and abstracted them, it will be very easy for new properties to be missed if it was hardcoded.

Actually, it will be impossible to miss new properties, because the tests will fail. 

I have already conceded on dynamically generating the list of packages when running the commands, because all that did was introduce an additional dependency to them.

I am not going to reduce the effectiveness of these tests in order to avoid having to occasionally update the tests. As the behavior of a program changes, you update the tests to reflect that. This is normal practice.

uberhacker: Think about how much easier it will be knowing the tests will work as expected without having to update them manually each time something changes.  That is working smarter, not harder.

Zopolis4: > without having to update them manually each time something changes. 

This is the whole point of a test. It detects if something changes. 

Sometimes, we want something to change (i.e. adding a new property), and in that case we update the test.

Otherwise, we didn't want something to change, and in that case we fix the bug.

You seem to be trying to get the benefits of a test (detecting if something changes) while trying to stop it from detecting certain changes (adding a new package). This is not a good idea, and will at best result in a test with bad coverage.

Zopolis4: Can we just agree to disagree on this one?

I've capitulated and let the list be dynamic in the command files, can you let it be static in the test files?

uberhacker: I provided help_test to detect changes and output what is missing.  This test will fail without knowing what went wrong or how to fix it.  Maybe it makes sense to come up with a better test?

Zopolis4: I really don't see the logic behind the test for a method that outputs a string being anything other than checking that the string is the same as what we expect it to be.

uberhacker: The expected output for `Available properties: ...` is the output generated by `Package.print_boolean_properties`, otherwise, the test will fail if a new property is added and this string is not updated.  There will be no obvious way to know how to fix the test.  A test should only fail based on the expected output.  I honestly think this is a worthless test anyways even if it passes.  Think about that logic for a moment.

Zopolis4: > The expected output for `Available properties: ...` is the output generated by `Package.print_boolean_properties`

Yes. If a new property is added, `Package.print_boolean_properties` will have a different output. Or, if something breaks, `Package.print_boolean_properties` will have a different output. 

> There will be no obvious way to know how to fix the test.

The test literally prints out the expected output and the actual output. 
In the case of a new string being added, you can just copy the actual output into the test file. 
It could not be more obvious, especially as when you are adding a new string you expect the output to change.

> A test should only fail based on the expected output.

Which this one does?

Zopolis4: > I honestly think this is a worthless test anyways even if it passes.

Then clearly there is no value in continuing this discussion further.
I want this test to be hardcoded so it detects if the output changes, and you don't want this test to exist at all.

uberhacker: The point I'm trying to make which you don't seem to understand is you shouldn't need to change the test to fix the expected output.  It should be the other way around.  The expected output should always be correct in a test.  The reason I think this is a worthless test is it doesn't provide any value as a test.  I guess you think making a manual change to a test any time we add a new property is fine.  It makes me cringe thinking about it; especially if that changes needs to be done manually in 4 different places.  That is the definition of insanity.

Zopolis4: > The point I'm trying to make which you don't seem to understand is you shouldn't need to change the test to fix the expected output.

When we are deliberately changing the behavior, it makes sense to me that we then change the expected behavior.

uberhacker: Why do it manually?  I think at this point it probably makes sense to drop this test since it adds no value even if it passes.

Zopolis4: > I guess you think making a manual change to a test any time we add a new property is fine.

When we add a new property we make a manual change in:
`lib/package.rb`, to add the property to the list of properties there.
`commands/help.rb`, to add the property to the list of properties there.

And that's not counting the changes to `bin/crew` and whatever packages need that property.

> It makes me cringe thinking about it; especially if that change needs to be done manually in 4 different places.

2 places, actually.

So yes, I do think making two additional manual changes (which are impossible to forget, and the test gives you the right change to make), is something that is fine when adding a new property (which again, we do how often? monthly, at most?).

> The reason I think this is a worthless test is it doesn't provide any value as a test.

If you don't think this test provides value, I don't understand how you see value in testing at all.

This test ensures that `Package.print_boolean_properties` doesn't break unnoticed. 
Admittedly, if it was just a test checking two hardcoded values, like it was when I originally opened this PR, then yes, this test would have less value as it is unlikely that we would break `puts` or similar.

uberhacker: A test like help_test provides value because it gives helpful output if the test fails.  This test just fails only because you forgot to do a manual change, otherwise, it will always pass and prove what? 
 `Package.print_boolen_properties` should have a separate test.  It shouldn't rely on this test which would make no sense at all.

Zopolis4: > This test just fails only because you forgot to do a manual change, otherwise, it will always pass and prove what?

Again, it also fails if the behavior of `Package.print_boolean_properties` changes.

> Package.print_boolean_properties should have a separate test. It shouldn't rely on this test which would make no sense at all.

This is literally the only place in the code `Package.print_boolean_properties` is used. I cannot see any possible reason to not test it here.

Zopolis4: > A test like help_test provides value because it gives helpful output if the test fails.

I agree, it is better to have a test provide additional information on what the problem is. 

But `help_test` only fails if someone forgot to make a manual change and add their property to `commands/help.rb`.

It is good that we cover that specifically, but it is also important to ensure the behavior of `Package.print_boolean_properties` does not change.

tests/commands/help.rb
uberhacker: Replace with?
```
Available commands: #{CREW_COMMANDS}
```

Zopolis4:
This is ridiculous. I opened this PR to make the codebase better, cleaner, more readable. In your inability to compromise, you have worsened the situation to the point where merging this PR will still leave us worse off than where we started.

I have tried every possible avenue to reach an agreement. I compromised down from 4 to 2. I offered to do the work you did not want to do. I explained over and over again why I was doing what I was doing. 

In return, you:
- Added a test that does not run as part of CI, is largely composed of system commands, and covers only one aspect of my test.
- Abstracted away the help string from `bin/crew`, where common practice dictates it should be, and moved it into `lib/const.rb`, despite it only being used once in the codebase, adding 65 lines to a file that is composed almost exclusively of single-line statements for constants we reuse multiple times throughout the codebase.
Had I been awake when these PRs were opened, I would have objected, but somehow I doubt that would have changed anything.

It is obvious that exponentially more time has been spent trying to avoid occasionally changing two strings then actually doing that simple task would have required. 

And, might I add, neither of your tests actually test properly-- they just catch the number of commands/constants. Perhaps ditching established testing practice in favor of a jumble of `system()` calls is not a good idea. 

I really don't see what else I could have done in this situation-- attempts to compromise failed, attempts to explain failed, attempts to literally do the work that you didn't want to do failed. When I offer to solve your concerns by doing the work you are concerned about, and you reject this in favor of adding 35 lines of code that don't actually catch real bugs, I don't see any further value in trying to reason with you.

I'm closing this PR, enough damage has been done as it is in your quest to avoid changing two strings every few months. If you want to reach an agreement that doesn't involve whatever [this](https://github.com/chromebrew/chromebrew/blob/master/tests/help_test) is, perhaps the one where I offered to do the work that you are trying to avoid, feel free to let me know.

I'm taking a break for now, I don't see the value in trying to improve the codebase if you are damaging it faster than I can fix it.

uberhacker:
> This is ridiculous. I opened this PR to make the codebase better, cleaner, more readable. In your inability to compromise, you have worsened the situation to the point where merging this PR will still leave us worse off than where we started.
> 
> I have tried every possible avenue to reach an agreement. I compromised down from 4 to 2. I offered to do the work you did not want to do. I explained over and over again why I was doing what I was doing.
> 
> In return, you:
> 
> * Added a test that does not run as part of CI, is largely composed of system commands, and covers only one aspect of my test.
> * Abstracted away the help string from `bin/crew`, where common practice dictates it should be, and moved it into `lib/const.rb`, despite it only being used once in the codebase, adding 65 lines to a file that is composed almost exclusively of single-line statements for constants we reuse multiple times throughout the codebase.
>   Had I been awake when these PRs were opened, I would have objected, but somehow I doubt that would have changed anything.
> 
> It is obvious that exponentially more time has been spent trying to avoid occasionally changing two strings then actually doing that simple task would have required.
> 
> And, might I add, neither of your tests actually test properly-- they just catch the number of commands/constants. Perhaps ditching established testing practice in favor of a jumble of `system()` calls is not a good idea.
> 
> I really don't see what else I could have done in this situation-- attempts to compromise failed, attempts to explain failed, attempts to literally do the work that you didn't want to do failed. When I offer to solve your concerns by doing the work you are concerned about, and you reject this in favor of adding 35 lines of code that don't actually catch real bugs, I don't see any further value in trying to reason with you.
> 
> I'm closing this PR, enough damage has been done as it is in your quest to avoid changing two strings every few months. If you want to reach an agreement that doesn't involve whatever [this](https://github.com/chromebrew/chromebrew/blob/master/tests/help_test) is, perhaps the one where I offered to do the work that you are trying to avoid, feel free to let me know.
> 
> I'm taking a break for now, I don't see the value in trying to improve the codebase if you are damaging it faster than I can fix it.

A bit of an overreaction don't you think?  You were almost there.  Here is the thing.  Your test does not provide the correct expected values after stuff is added.  This is the reverse of how it should work.  The expected value should always be the expected value, not some outdated value.  Making the test string dynamic avoids the need to update tests after something changes.  This makes tests more full proof because the expected values change with the variable changes.  FYI, I moved the constants out of bin/crew and into lib/const.rb because I wanted to create the `CREW_COMMANDS` constant to be used in other places in the code, including your tests!  I appreciate all the work you have put into this project and I by no means meant to offend.

uberhacker:
You don't think this is helpful?  The test explains everything that needs to be updated WITHOUT needing to update the tests manually.  This will work for future tests as well.
```
chronos@mx-x86_64 ~/chromebrew/tests $ ./help_test 
Help is missing for the crew commands in commands/help.rb below:
help
Help is missing for the properties in commands/help.rb below:
arch_flags_override
gnome
no_upstream_update
Help test failed.
```

uberhacker:
I like everything about this PR with only the dynamic changes needed that both @satmandu and I have suggested.

Zopolis4:
I'll re-open the PR and make your changes, but a few things--

Fundamentally, the reason I think the tests should be static and require manual updating is because when we add a new property we change the behavior of the command, so obviously we need to update the tests to now expect that new behavior.

If you disagree with that idea then that's one thing, but if your main concern is the labor required in manually updating the tests, would you be open to a solution whereby I provide an external tool (perhaps a shell script, maybe something in the CI) that does that work for you? 

We can reach an agreement on this another time, let's get this PR merged first.

uberhacker (approved):

PR #9516, opened 2024-03-21
Automatically source ~/.bashrc and similar

Zopolis4:
The "trick" is to use `exec()` to replace the ruby process with a new instance of the parent shell on exit, which sources the relevant configuration file(s).

Tested and working on `x86_64` with both `bash` and `fish`-- I couldn't get `zsh` working (at all) in the container, so that hasn't been tested.

Also, given that we don't need it anymore, I removed `print_source_bashrc` and the handling around that, as well as any other leftover calls to `puts source ~/.bashrc` and the like.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=whisk crew update
```


Zopolis4:
Note that this will not run if you just call `crew`, as docopt handles the exit in that case and none of the crew `at_exit` code runs.

Zopolis4:
This will be pretty verbose on the containers because they print `This is the .bashrc` whenever the `.bashrc` is sourced, btw.

Zopolis4:
Ah-- this hangs the unit tests, because we start a new shell.

hm

bin/crew
```diff
@@ -175,6 +175,8 @@ at_exit do
   GnomePostinstall.run unless GnomePostinstall.gnome_packages.blank?
   # Print exit messages.
   ExitMessage.print
+  # Replace the currently running Ruby process with a new instance of the parent shell (thus sourcing the relevant configuration)
+  exec(`ps -p #{Process.ppid} -o comm=`.chomp)
```
uberhacker, line 179: I highly doubt this works the way you think.  Sourcing has to be done in the current working session or it won't work.

Zopolis4: It works. I've tested it.

yeesterPlus (changes requested):
please don't do this, shutdown pull request, until there is a way

Zopolis4:
Fixed-- set `CREW_SCRIPTING` to 1 if you are running crew commands in a script from now on.

supechicken:
Sorry for late response😅

I don't think this is a good approach, as it will spawn "nested" shells after each `crew install` call (the original shell is still running unless with `exec crew install`), which is a waste of memory.

If I remember correctly, you cannot simply kill the parent shell as it will cause the crosh window close immediately.

FYI https://github.com/chromebrew/crew-profile-base/pull/6 was my previous attempt on auto-sourcing bashrc after `crew install`

Zopolis4:
Yeah, the additional shells aren't great, but this is the best solution I could find.

Ideally, we'd just replace the shell process that spawned crew, but we can't do that if it'll kill the crosh window.

In theory, could we spawn one additional shell process, then continuously replace that?

Zopolis4:
I'm sure there's a way to check if there have been any changes in the environment that need to be sourced (i.e. try starting a new shell and check if its environment setup is the same as the old one), and then we could only start a new shell based off of that?

That wouldn't eliminate the problem of additional shells, but it would drastically reduce it.

supechicken:
The window will not close if we run `exec bash` in the shell spawned with `crosh> shell`, it closes only when we trying to `kill -9` the shell.

> Ideally, we'd just replace the shell process that spawned crew, but we can't do that if it'll kill the crosh window.

Why we need to replace the shell at all :) If we can run `exec bash` in the parent shell then we should be able to `source /usr/local/etc/profile` while keeping the current shell (variables will be override anyways).

We can use something like `inotifywait` to monitor if there are any changes in `/usr/local/etc/env.d` or with `sha256sum /usr/local/etc/env.d/*` and compare the result

Zopolis4:
The problem is that we can't run `exec bash` in the parent shell since we're still in `crew`, as far as I'm aware.

supechicken:
> The problem is that we can't run `exec bash` in the parent shell since we're still in `crew`, as far as I'm aware.

Although we cannot call `exec bash` or `source ~/.bashrc` directly in `crew`, we can call the parent shell to do it by using custom signal traps.

We can use `SIGUSR1` as the signal for `source ~/.bashrc`, like this:
```bash
# Add this line to /usr/local/etc/profile
trap "source ~/.bashrc" USR1
```

In `crew`, all we need to do is send a signal to the parent shell:
```ruby
Process.kill('USR1', Process.ppid)
```

satmandu:
> ```shell
> # Add this line to /usr/local/etc/profile (crew-profile-base)
> trap "source ~/.bashrc" USR1
> ```
Don't we sometimes source /usr/local/etc/profile from ~/.bashrc? This would create a loop...

supechicken:
> > ```shell
> > # Add this line to /usr/local/etc/profile (crew-profile-base)
> > trap "source ~/.bashrc" USR1
> > ```
> 
> Don't we sometimes source /usr/local/etc/profile from ~/.bashrc? This would create a loop...

No, as the trap will trigger once only after someone try to `kill -USR1` the shell process, then the trap will overrides itself during `source ~/.bashrc`

satmandu:
@Zopolis4 How do you feel about modifying this PR to use `SIGUSR1` with `trap` like @supechicken suggested?

satmandu:
@supechicken The approach you suggested kills the parent process of crew, which will close `crosh`.

Try this in a container:
```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=autosource crew update ; crew upgrade
```

satmandu:
In https://github.com/chromebrew/chromebrew/pull/9649 I'm just execing a shell at the end of the crew process if the trap isn't set... but that should only happen once, I think.

satmandu:
(If that's ok, you should be able to merge master and use 95% of the changes here...)

satmandu:
I would like to merge https://github.com/chromebrew/chromebrew/pull/9649 into this.

I don't think `CREW_SCRIPTING` is necessary the way you have implemented it, since if we use trap we can just tell the shell to reload `~/.bashrc` without reloading the shell, but that needs a variable set in the parent shell that sets the trap, which is why I'm doing that inside `/usr/local/etc/profile`.

But I think the essence of that would be helpful in multiple places, maybe as a `CREW_NONINTERACTIVE`, so we could also tell non-interactive usage to just bypass waiting around or doing any fancy graphical stuff.

zsh not working ought to be a separate bug... but also, if nobody uses it... maybe we should also just mark is as broken... maybe somebody wants to take that up as their own project.

Zopolis4:
Updated and rebased to only have the removal of `source ~/.bashrc` messages.

bin/crew
```diff
@@ -130,6 +130,8 @@ at_exit do
 
   # Print exit messages.
   ExitMessage.print
+  # Replace the currently running Ruby process with a new instance of the parent shell (thus sourcing the relevant configuration)
+  exec(`ps -p #{Process.ppid} -o comm=`.chomp) unless CREW_SCRIPTING
```
uberhacker, line 134: Bump crew version in `lib/const.rb`?

satmandu: So this approach doesn't work, right? So do we need to just rename the print_source_bashrc line with something else since the autosource PR handles this? Maybe `'s/print_source_bashrc/source_bashrc/'` or  `'s/print_source_bashrc/autosource_bashrc/'` ?

Zopolis4: i'll see if i can profile that, actually. some numbers might help in the decision.

.github/workflows/Unit-Test.yml
```diff
@@ -13,19 +13,20 @@ jobs:
...
+            yes | crew remove vim
+            ruby ../tests/prop_test
+            ruby ../tests/buildsystem_test
+            ruby ../tests/commands/const.rb
+            ruby ../tests/commands/help.rb
+            ruby ../tests/commands/prop.rb
+            cd ~
+            git clone --depth=1 https://github.com/chromebrew/chromebrew.git
+            cd chromebrew
+            yes | crew build -f packages/hello_world_chromebrew.rb
```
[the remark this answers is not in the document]
Zopolis4: Looking into this further, setting `CREW_CACHE_ENABLED=1` just papers over the bug that `crew build` is actually broken as it attempts to install packages after they are built, despite new binaries not having been uploaded-- we probably shouldn't be silencing bugs in our unit tests...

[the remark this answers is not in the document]
satmandu: That's not a bug. That's intentional.

Installing just built packages as part of a sequence of building multiple interdependent packages is an intentional feature of the `-f` flag to `crew build` in that context.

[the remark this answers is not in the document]
satmandu: Otherwise you can't test if a just built package installs correctly.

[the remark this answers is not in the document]
Zopolis4: But that code fails by default if CREW_CACHE_ENABLED is not set.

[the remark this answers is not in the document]
satmandu: I enable have `CREW_CACHE_ENABLED=1` in all of my builds...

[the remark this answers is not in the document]
satmandu: But you're right, we should have this work without that...

bin/crew
Zopolis4: looks like each new bash process adds around 8MB. That's not acceptable, I'll look into ways of dealing with that.

Zopolis4: Slightly more advanced testing setup (copilot-written script, so take it with a grain of salt)
```bash
#!/bin/bash

while true; do
    bash -c 'echo $$; ps -p $$ -o rss= | awk "{ printf \"%.2f MB\n\", \$1/1024 }"'
    sleep 3
done
```

Gives around 3.5 MB per new process.

Zopolis4:
Just checking I've got the right understanding here, but we now `source ~/.bashrc` completely automatically, right? So we're good to delete all the `print_source_bashrc` stuff as per this PR.

bin/crew
uberhacker: +1 for renaming `print_source_bashrc` to `source_bashrc`.  This will enable us to control at a package level which installation requires a `source ~/.bashrc`.  That said, this PR has a lot of changes to make that happen.

Zopolis4: If we can `source ~/.bashrc` on crew exit, which we can in master, why don't we just do that unconditionally?

uberhacker: It doesn't make sense to source unless it would be necessary.

Zopolis4: Better to be safe than sorry, and its not like sourcing costs us anything. Maybe only run after installing a package?

uberhacker: Sourcing after every install could have unintended consequences.  For instance, sometimes you might want to set environment variables for a specific use case and the sourcing could revert those settings.  Better to have control on a package level.

Zopolis4: By that logic, sourcing after any install could have unintended consequences-- better to consistently do something than have seemingly random failures of environment variables and the like.

uberhacker: You're not understanding my point.  Having the boolean property lets you have control over the process in cases where you might not want to source.  It also makes packages that require sourcing self-documented so you know when it is needed.  How is that logic?

Zopolis4: How about checking if anything in `/usr/local/etc/env.d/` was changed, so that sourcing is only applied when it is needed?

satmandu: Is that not harder to do? And how do you know if ~/.bashrc has been sourced yet at all? Isn't that the larger problem?

Zopolis4: > Is that not harder to do? 
You check the latest file modification timein `/usr/local/etc/env.d/`, compare it with current, and done. Shouldn't be too bad. At any rate, I'm trying to avoid repeating lots of code in package files, and this is a good place to improve.

> And how do you know if ~/.bashrc has been sourced yet at all?
Not really sure what you mean here.

satmandu: >> And how do you know if ~/.bashrc has been sourced yet at all?
> Not really sure what you mean here.

Sometimes we want to make sure that the bashrc is sourced after the crew command, irrespective of whether files have changed in env.d (because scripts there might change some state when being run again, even if the scripts themselves haven't changed.)

Does that make sense?

Zopolis4: I think so? We can set it up to source after every command, but we currently don't. If we want to do it unconditionally, it shouldn't be too hard.

satmandu: I think there is a middle ground where we don't want the running to be unconditional. See for instance the bash scripts for distcc.

We don't need those run after every single crew command. They only need to be run after distcc is installed.

satmandu: (but they can be run after every shell invocation... That's fine... If run with every crew command we will essentially get logspam with every crew command.)

PR #9519, opened 2024-03-22
Simplify crew prop command

uberhacker:
Helps improve #9514 by providing a solution to avoid repeating the properties output in 4 different places.  Caveat is `require 'package'` is a prerequisite for each file requiring the method reference.

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -360,17 +360,11 @@ def files(pkg_name)
 end
 
 def prop(silent = false)
-  props = []
-  pkg = Package.new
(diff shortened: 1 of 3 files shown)


Zopolis4:
If you're going to do this, would you mind adding documentation for the properties this exposes? (gnome, arch_flags_override and git_clone_deep)

uberhacker:
> If you're going to do this, would you mind adding documentation for the properties this exposes? (gnome, arch_flags_override and git_clone_deep)

I thought you could do that in your PR?  We'll need to split out the help to a separate file and update with this method call anyway, right?

satmandu (approved):

PR #9529, opened 2024-03-23
Split remove command into separate file

Zopolis4:
**Split out `remove` command to `commands/remove.rb`.**

Behavior changes:
- Remove random `[options]` in the docopt EOF for `crew remove`-- it only supports the already listed verbose option.
- Don't print `filelist contains <line>` when line does not start with `CREW_PREFIX` and the verbose option is passed.
  
Refactoring changes:
- Use a list of the essential packages and refuse to remove those, rather than using `CREW_ESSENTIAL_FILES`.
  - I've tested this on all architectures and glibc versions and `CREW_ESSENTIAL_FILES` is the same every time.
  - The essential packages provide all these libraries, and I have checked that this is the case with all the glibc versions.
  - I didn't use the list of core packages here because that is not what is essential for the functionality of crew-- If we have all these essential packages we can install all the others.
  - There is a possibility of adding packages like `crew_mvdir` and removing the checks for those, but that's outside the scope of this PR.
- Don't use `determine_conflicts`-- keep it in `bin/crew` as `prepare_package` still uses it.
- ~Don't use `@is_fixup`, have `lib/fixup.rb` use a system call to `crew remove <package>`~
  - ~`lib/fixup.rb` should really just be a regular ruby script that we happen to run from `crew`, as integrating it risks a breakage in `crew` being something that we can't solve with `lib/fixup.rb`.~
  - ~Fixing that more is for a later time, though.~
- Don't use `@is_fixup`, create a minimal Package object in `lib/fixup.rb` with stub methods for preremove and remove.
- Don't symbolize names when loading `device.json`.
- Error out instead of silently doing nothing in the case of a missing filelist.
- Clean up and simplify the actual removal of files and directories, and the various checks that occur while doing so.
- Use `FileUtils.remove_file` for removing files and `Dir.rmdir` for removing directories, for security and readability reasons.
- Use `File.join` instead of string interpolation.
- Load the `Package` Object through `search name` so that we can check if the package exists.
  - This is done instead of passing in the name and loading the package in `commands/remove.rb`.

No tests at the moment as I need to split out the install command first.

Still tested and working on `x86_64`, though.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula5 crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -8,6 +8,7 @@ require_relative '../commands/const'
 require_relative '../commands/help'
 require_relative '../commands/list'
 require_relative '../commands/prop'
+require_relative '../commands/remove'
 require_relative '../commands/sysinfo'
 require_relative '../lib/color'
 require_relative '../lib/const'
@@ -1284,7 +1285,7 @@ def install
   # remove it just before the file copy
   if @pkg.in_upgrade
     puts 'Removing since upgrade or reinstall...'
-    remove @pkg.name
+    Command.remove(@pkg, @opt_verbose)
   end
 
   unless @pkg.is_fake?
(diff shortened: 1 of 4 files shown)


Zopolis4:
It'd be nice to rubify the grep call on 35 of `commands/remove.rb`:
```ruby
system("grep --exclude #{pkg.name}.filelist -Fxq #{line} ./meta/*.filelist")
```

To be honest, though-- I have a feeling re-implementing this in ruby will make it longer and less readable for relatively minimal benefits (akin to using `File.join` but worse).

At any rate-- I didn't think it was important enough to hold up this PR, and I'll have a look at how feasible it is later.

lib/fixup.rb
```diff
@@ -128,14 +128,7 @@
...
   when 'y', 'yes'
-    @in_fixup = true
-    begin
-      remove pkg[:pkg_name]
-    rescue NoMethodError # This won't work the first time crew update happens, since this requires an update to crew.
-      puts "Please rerun 'crew update' to allow for removal of #{pkg[:pkg_name].capitalize}.".orange
-      puts "#{pkg[:pkg_name].capitalize} not removed.".lightblue
-    end
-    @in_fixup = false
+    system("crew remove #{pkg[:pkg_name]}")
```
uberhacker, line 131: Didn't we have to refactor due to environment issues in the containers because we ran crew within crew @satmandu?

lib/const.rb
```diff
@@ -304,7 +304,3 @@
 PY_SETUP_INSTALL_OPTIONS         = "#{PY_SETUP_INSTALL_OPTIONS_NO_SVEM} --single-version-externally-managed"
 PY3_BUILD_OPTIONS                = '--wheel --no-isolation'
 PY3_INSTALLER_OPTIONS            = "--destdir=#{CREW_DEST_DIR} --compile-bytecode 2 dist/*.whl"
-
-CREW_ESSENTIAL_FILES = `LD_TRACE_LOADED_OBJECTS=1 #{CREW_PREFIX}/bin/ruby`.scan(/\t([^ ]+)/).flatten +
-                       %w[libzstd.so.1 libstdc++.so.6]
-CREW_ESSENTIAL_FILES.uniq!
```
uberhacker, line 310: Bump crew version?  Make sure you merge updated master into your working branch first.

lib/fixup.rb
Zopolis4: We do a similar thing at line 1665 of `bin/crew`:
```ruby
system 'yes | crew install ruby_rubocop' unless @device[:installed_packages].any? { |elem| elem[:name] == 'ruby_rubocop' }
```

Though I'm not aware of these issues so not sure if this would trigger it.

bin/crew
```diff
@@ -55,7 +56,7 @@ DOC = <<~DOCOPT
     crew postinstall [options] [-v|--verbose] <name> ...
     crew prop [-v|--verbose]
     crew reinstall [options] [-k|--keep] [-s|--source] [-S|--recursive-build] [-v|--verbose] <name> ...
-    crew remove [options] [-v|--verbose] <name> ...
+    crew remove [-v|--verbose] <name> ...
```
uberhacker, line 59: I'm still foggy on what exactly `[options]` is used for inside docopt.

Zopolis4: Me too-- it seems safe to remove it so far though, haven't encountered any problems.

Zopolis4: image
Looks like we definitely don't need it here.

lib/fixup.rb
satmandu: Yes that was a problem! We should fix line 1665! invoking crew within crew needs to be fixed, but currently it breaks things...

yeesterPlus (approved):
More friendley

lib/fixup.rb
Zopolis4: Admittedly I don't think line 1665 ever gets called, as rubocop is in buildessential.

uberhacker: @satmandu: What is your recommendation?  This is the only thing I could find in this PR where there might be a potential problem.

Zopolis4: Fixed.

commands/remove.rb
```diff
@@ -0,0 +1,67 @@
...
+        Dir.rmdir line
+      end
+
+      # Remove the file and directory lists.
+      FileUtils.remove_file File.join(CREW_META_PATH, "#{pkg.name}.filelist")
+      FileUtils.remove_file File.join(CREW_META_PATH, "#{pkg.name}.directorylist")
+    end
+
+    # Remove the package from the list of installed packages in device.json.
+    puts "Removing package #{pkg_name} from device.json".lightred if verbose
```
uberhacker, line 56: `.lightred` indicates an error.  Would `.yellow` make more sense?

commands/remove.rb
```diff
@@ -0,0 +1,67 @@
...
+    device_json['installed_packages'].delete_if { |entry| entry['name'] == pkg.name }
+
+    # Update device.json with our changes.
+    File.write File.join(CREW_CONFIG_PATH, 'device.json'), JSON.pretty_generate(JSON.parse(device_json.to_json))
+
+    # Perform any operations required after package removal.
+    pkg.remove
```
uberhacker, line 63: What happened to the `@in_fixup` condition?

Zopolis4: We don't need it anymore-- the same functionality is achieved through other means.

uberhacker (approved):

uberhacker:
Resolve merge conflicts?

Zopolis4:
Resolved.

supechicken (approved):

PR #9530, opened 2024-03-23
Add buildsystem_test and check

uberhacker:
Tested and working both as a standalone script and with the `crew check` command.  Also added check to see if there were any changes to warrant a copy to crew from the local working repo.

Example in action:
```
$ crew check perl_test
Checking perl_test package ...
Property tests for perl_test passed.
Checking perl_test package ...
Perl is an invalid buildsystem. Valid buildsystems include Autotools, CMake, Meson, PERL, Package, Pip, Python, Qmake, RUBY.
Buildsystem test for perl_test failed.
```
Notice `Perl` is not all capital letters.  This is a common mistake when building packages.

### DIFF
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -21,6 +21,7 @@ jobs:
             yes | crew install vim && \
             yes | crew remove vim && \
             ruby ../tests/prop_test && \
+            ruby ../tests/buildsystem_test && \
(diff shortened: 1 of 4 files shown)


Zopolis4:
Add this to the CI?

uberhacker:
> Add this to the CI?

Great, just another test to slow things down. LoL  Done.

Zopolis4 (approved):

PR #9594, opened 2024-04-03
Split upload command into separate file

Zopolis4:
**Split out `upload` command to `commands/upload.rb`.**

Behavior changes:
- Remove support for uploading without passing a package name.
  - This wasn't working anyways, and I can't conceive of a situation where you are uploading packages without knowing their names or having just built them.
  - If this functionality is still needed, I can write a bash harness to implement this functionality externally.
- Remove support for uploading multiple architectures.
  - This functionality was working, but it would have dramatically complicated the code to continue to implement.
  - If this functionality is needed, just setting the `ARCH` environment variable should take care of things.
- Remove `[options]` from the docopt section for `crew upload`.
  - Same rationale as the last times I've done this.
- Remove verbose argument.
  - Because of the refactor, the code that previously would have used this is now gone. 
  
Refactoring changes:
- Rework `load_package` in `lib/package.rb` to only take one argument, being the path to the package file.
  - Derive the package name in the function itself, rather than the weird setup we had before with a default argument value (when would we ever want to pass a different value to that?).
- Rework `set_package` in to only take one argument, , being the path to the package file.
  - All it was using the previous argument for was to unnecessarily pass it to `load_package`.
- Rework `search` to accept a package name, path, and a value for silent.
  - Use named parameters for the last two to save boilerplate.
  - Use a default value for `pkg_path`, as it is a named parameter rather than just being an optional one.
- Move the rubocop section that was previously in `def upload` to `def copy_package`.
  - Better to sanitize the package file earlier in the process.
- Load the package file from `CREW_LOCAL_REPO_ROOT` when passing the package object to `Command.upload`, as this is the package file that we will be working with.
- Rework the way we look for `GITLAB_TOKEN` to not need a trailing `if` and `.nil?` by using a block in the fallback of `ENV.fetch` to abort if we can't find it.
- Check `no_zstd` to find the extension of the newly built binaries and set the local variable `binary_compression` to that.
  - Since we just built the binary, we can be confident in the extension, as chromebrew will only use one of two compression programs.
- Use `net/http` instead of `curl` to upload the binary.
- Use `.sub!` and `.gsub!` instead of `sed` to update the package file.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula6 crew update
```


bin/crew
```diff
@@ -202,10 +203,9 @@ def generate_compatible
   puts 'Generating compatible packages done.'.orange if @opt_verbose
 end
 
-def search(pkg_name, silent = false)
-  pkg_path = File.join(CREW_PACKAGES_PATH, "#{pkg_name}.rb")
+def search(pkg_name, pkg_path: File.join(CREW_PACKAGES_PATH, "#{pkg_name}.rb"), silent: false)
```
uberhacker, line 206: Why does this need to be a parameter?  Keep it simple!

bin/crew
```diff
@@ -1964,10 +1838,9 @@ end
 def upgrade_command(args) = upgrade(*args['<name>'], build_from_source: @opt_source)
 
 def upload_command(args)
-  upload if args['<name>'].empty?
   args['<name>'].each do |name|
-    search name
-    upload name if @pkg.name
+    search name, pkg_path: File.join(CREW_LOCAL_REPO_ROOT, 'packages', "#{name}.rb")
```
uberhacker, line 1842: No need to pass the package path.  Let the function figure it out from the predefined constant.  Also, we do not want to use `CREW_LOCAL_REPO_ROOT`.  We should be using `CREW_PACKAGES_PATH`.

uberhacker (changes requested):
I'm of the opinion we should only pass the package name to any function.  The function itself should figure out the path.  We also need to be consistent with the calls.  Right now, some functions require the path and others just the package name.  Let's just send the package name only.

commands/upload.rb
```diff
@@ -0,0 +1,50 @@
...
+    gitlab_token = ENV.fetch('GITLAB_TOKEN') { abort 'GITLAB_TOKEN environment variable not set.'.lightred }
+
+    # If the package has the no_zstd property, the binaries will be compressed as .tar.xz
+    # Otherwise, the binaries will be compressed as .tar.zst
+    binary_compression = pkg.no_zstd? ? 'tar.xz' : 'tar.zst'
+
+    # Get the name of the compressed binary.
+    pkg_tarfile = "#{pkg.name}-#{pkg.version}-chromeos-#{ARCH}.#{binary_compression}"
+    # Get the absolute path to the compressed binary, aborting if there is no file there.
+    abort "No compiled binary found for #{pkg.name}.".lightred unless File.file?(pkg_tarfile_path = File.join(CREW_LOCAL_BUILD_DIR, pkg_tarfile))
```
uberhacker, line 19: The `File.file?` parameter is an assignment, not a file.

commands/help.rb
```diff
@@ -149,8 +149,6 @@ def self.help(command)
         Usage: crew upload [<package1> <package2> ...]
         This will update the binary_sha256 hashes in each package and upload binaries to GitLab.
         The GITLAB_TOKEN environment variable must be set to access the upstream repository.
-        If no package(s) are provided, all binaries in `#{CREW_LOCAL_REPO_ROOT}/release/<arch>` will be uploaded.
-        If `-v` or `--verbose` is present, additional debug information will be displayed.
```
uberhacker, line 153: Why remove and require 3 separate uploads for all the architectures?

bin/crew
Zopolis4: Please read the reasons I gave in the PR description...

Zopolis4:
I'm not entirely sure what you mean by this-- we only pass the package path to search, which loads the package file.

commands/upload.rb
Zopolis4: It both assigns the value and checks that the file exists.

commands/help.rb
Zopolis4: The added code complexity wasn't worth it, and I couldn't think of a scenario where uploading all architectures was useful.

At any rate, if this is a functionality you do use, please see the PR description for how you can replicate it.

bin/crew
uberhacker: You have to copy the package from your local repo to the crew packages directory BEFORE you can even test it so I'm not sure why you think this needs a refactor.  Also, please stop resolving conversations.

commands/upload.rb
```diff
@@ -0,0 +1,50 @@
...
+    pkg_file = File.join(CREW_LOCAL_REPO_ROOT, 'packages', "#{pkg.name}.rb")
+
+    # Read the package file into memory.
+    file = File.read(pkg_file)
+
+    # Update the package file with the new binary_compression value.
+    file.sub!(/binary_compression.*/, "binary_compression '#{binary_compression}'")
+
+    # Update the package file with the new binary_sha256 value.
+    file.gsub!(pkg.binary_sha256[ARCH.to_sym], Digest::SHA256.hexdigest(File.read(pkg_tarfile_path)))
```
uberhacker, line 45: This handles updating the package if the sha256sum hash exists but what about new packages that don't have the section added yet?

Zopolis4: New packages will need a dummy section something like this one:
```ruby
  binary_sha256({
    aarch64: 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
     armv7l: 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
       i686: 'iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii',
     x86_64: 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
  })
```

uberhacker: All the code to not require this is already there.  You are digressing the code and making the process harder instead of making it better.

bin/crew
Zopolis4: > I'm not sure why you think this wasn't working when I've been using it for over a month now without an issue.

Probably because it's been broken whenever I tried to do it.

> No, the whole purpose of adding this function was to remove the need for bash scripts and incorporate it as a crew command instead.

Again, I do not understand how you can be in a scenario where you don't know the names of the binaries you are uploading. I genuinely do not see a usecase for not passing a package name.

> Then add it back. We'll want to control verbosity.

Because we aren't using system commands anymore, we can't just print the `curl` command we are using. If it was possible to add this code back I would have.

> It was already written and simplifies the process for contributors. Because the code is complicated is not a valid reason for removing it altogether.
> This functionality was added for a reason

Again, what is the reason? When are you in a scenario where you aren't on the same architecture as the binaries you've built?

> Again, package name makes more sense than package path
> No need to do this at all.

Please see my other comment where I explained why this is not the case.

bin/crew
uberhacker: This is NOT my workflow.  This is harder than a simple `crew build libnvme` and NOT requiring the path.  The path is abstracted which is the whole point of why I wrote this originally.

Zopolis4: > This is NOT my workflow.

I'm not centering my refactor around your personal workflow.

uberhacker: Obviously because you don't understand my workflow.  I'm trying to make it EASIER, not HARDER, which is apparently the route you want to take.

commands/upload.rb
Zopolis4: > All the code to not require this is already there.

This code is broken, at least in my experience.

> You are digressing the code and making the process harder instead of making it better.

Would you be open to the addition of something like a `crew create` command which creates a package template file? My intention is not to make things more difficult, and I'd be happy to implement this functionality elsewhere where it is more suited.

bin/crew
Zopolis4: `crew` already requires that you be in a local repo-- I don't understand why it would be easier to then edit the actual crew installation?

> Obviously because you don't understand my workflow.

Apparently not. Would you mind explaining it so I can ensure it continues working?

uberhacker: I'm abstracting the path so if you are making changes to the package in your local repo, crew will detect that it differs from the crew repo package and offer to copy it over so it can run it inside crew.  So simply running `crew build/install/reinstall <package>` will enable you to test the package without needed to copy the package file over manually.  My vision is for contributors to be able to do everything they need to do to create and update packages inside their own forked repo instead of working inside crew directly.  This way changes are easier to manage and removes the potential human error element from the process.  I know @satmandu has a different workflow and that is fine but I still think mine is easier. :)

Zopolis4: So your goal is that the user does all the changes in their local repository, but crew then copies it to the installation, works with that, then copies back any changes to the local repository?

bin/crew
Zopolis4: Actually, I can probably re-implement some of the verbosity, I'll do that when I've got the time.

commands/upload.rb
uberhacker: Sure, that sounds like a great idea.  Maybe provide a few basic prompts like the package name, description, version, etc.  It would be nice to search through all the packages to determine if there is a name conflict also.

Zopolis4: Sounds good! I'll mark this PR as draft until then because I don't want to remove this functionality until an option is available.

bin/crew
Zopolis4: I'll point out that the workflow I am enabling here does not require any manual copying, and in fact just removes the requirement for a package file to be inside the crew installation. (I'm also trying to reach an agreement on this because it will significantly affect how I refactor `crew check`, which I'd like to get started on soon)

bin/crew
Zopolis4: Added back the verbosity that was practical to do.

commands/upload.rb
Zopolis4: See #9624.

bin/crew
uberhacker: I still don't think you fully understand my workflow.  The way it currently works (by design) is any change you make in your local working repo will be prompted to copy to crew for installation (no manual copying).  I don't think a refactor is necessary because it will over complicate the process and the package has to end up in crew anyway to test the install.  You can always do a crew update to restore back to the original version if you are updating existing packages.

Zopolis4: The point of these changes to `search` is to ensure that there is no need for any copying, whether automatic or manual. I'd argue that this simplifies the process, in fact.

lib/const.rb
```diff
@@ -364,7 +364,7 @@
     crew test [-v|--verbose] [<name> ...]
     crew update [options] [-v|--verbose] [<compatible>]
     crew upgrade [options] [-k|--keep] [-s|--source] [-v|--verbose] [<name> ...]
-    crew upload [options] [-v|--verbose] [<name> ...]
+    crew upload <name> ...
```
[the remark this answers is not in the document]
Zopolis4: The old verbose option just printed the system commands being used-- no system commands are used anymore, so I wasn't sure how to re-implement it.

Zopolis4:
What would people think if I added the automatic adding of manifests of the newly built packages? This would simplify the workflow to `crew build -f <package> && crew upload <package> && git commit && git push` in all cases.

uberhacker:
> What would people think if I added the automatic adding of manifests of the newly built packages? This would simplify the workflow to `crew build -f <package> && crew upload <package> && git commit && git push` in all cases.

Maybe a bit of automation overload?  I can certainly think of cases where you might not want to do everything at once.

Zopolis4:
I'm not saying that needs to be the workflow used, just that there's no need to run git add manifest/ARCH/p/package.filelist after guiding and uploading a package anymore. I find myself doing this quite often, and was wondering if it'd be helpful to automate.

uberhacker:
> I'm not saying that needs to be the workflow used, just that there's no need to run git add manifest/ARCH/p/package.filelist after guiding and uploading a package anymore. I find myself doing this quite often, and was wondering if it'd be helpful to automate.

Yes, some assistance with manifest files might be useful.

Zopolis4:
Added, tested, and working.

Zopolis4:
I believe this already incorporates the changes from #10240, although testing wouldn't hurt.

uberhacker:
Resolve conflicts?  We had quite a few changes to the upload command since this PR was introduced so tread with care.

PR #9599, opened 2024-04-04
Simplify check_package function

uberhacker:
Tested and working.


bin/crew
```diff
@@ -1731,20 +1731,12 @@ def check_package(pkg_name)
   end
 
   # Compare local repo package to the crew repo package and prompt to copy if necessary to prepare for the operation.
-  crew_package_updated = ''
+  package_updated = ''
   Dir.chdir CREW_PACKAGES_PATH do
-    crew_package_updated = `git diff #{CREW_PACKAGES_PATH}/#{pkg_name}.rb`.chomp
+    package_updated = `diff "#{CREW_LOCAL_REPO_ROOT}/packages/#{pkg_name}.rb" "#{CREW_PACKAGES_PATH}/#{pkg_name}.rb"`.chomp
```
Zopolis4, line 1736: We can probably Rubify this with `File.identical?`

uberhacker: Good call!  That is definitely more elegant.  I also discovered and fixed another bug if the package doesn't exist.

Zopolis4 (approved):

PR #9601, opened 2024-04-04
Split files command into separate file

Zopolis4:
**Split out `files` command to `commands/files.rb`.**

Behavior changes:
- Do not print files for packages that aren't installed.
  - This is consistent with the documentation, and the previous behavior would falsely report a size of 0 bytes.
  - If the consensus is that this behavior should be brought back in some way, I'm happy to implement that.
- Remove `[options]` and `[-v|--verbose]` in the docopt EOF for `crew files`.
  - Same rationale for removing `[options]` as usual, and we never actually did anything with the verbose option here.
- No newline printed after `Disk usage: ...`
  
Refactoring changes:
- Check if the package is installed before checking if it has a filelist.
- Don't use `.sort` after reading the filelist.
  - We already sort the filelists upon creation, and its not a good idea to have `crew files` possibly display a different sorting order.
- Slightly clean up the various checks on file existence and symlinks.
- Instead of using the weird previous setup of a `puts` call with multiple arguments and an EOT, separate the printing of the filelist, total number of files, and disk usage into their own `puts` calls. 

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula7 crew update
```

### DIFF
--- a/commands/files.rb
+++ b/commands/files.rb
@@ -1,0 +1,42 @@
+require 'json'
+require_relative '../lib/const'
+require_relative '../lib/convert_size'
+
+class Command
+  def self.files(pkg)
+    # Check if the package is even installed first, as this is the most likely reason we cannot find a filelist.
+    device_json = JSON.load_file(File.join(CREW_CONFIG_PATH, 'device.json'))
+    if device_json['installed_packages'].none? { |entry| entry['name'] == pkg.name }
+      puts "Package #{pkg.name} is not installed.".lightred
+      return
+    end
+
+    # We can't do anything if we don't have the filelist.
+    unless File.file?(filelist_path = File.join(CREW_META_PATH, "#{pkg.name}.filelist"))
+      puts "Package #{pkg.name} does not have a filelist :(".lightred
+      return
+    end
+
+    # Print the name and description of the package.
+    puts pkg.name.lightgreen + ": #{pkg.description}".lightblue
+
+    # Read the file into memory as an array of lines.
+    filelist = File.readlines(filelist_path, chomp: true)
+    size = 0
+
+    filelist.each do |filename|
+      # Skip calculating the filesize if the file doesn't exist.
+      next unless File.file?(filename)
+      # Ignore symlinks to prevent duplicating calculation.
+      next if File.symlink?(filename)
+      # Add the size of the file to the total size.
+      size += File.size(filename)
+    end
+
+    # Print the filelist, the total number of files, and the total size of those files.
+    puts filelist
+    puts "\nTotal found: #{filelist.count}".lightgreen
+    puts "Disk usage: #{human_size(size)}".lightgreen
+  end
+end
+
--- a/bin/crew
+++ b/bin/crew
@@ -5,6 +5,7 @@ require 'json'
 require 'fileutils'
 require 'tmpdir'
 require_relative '../commands/const'
+require_relative '../commands/files'
 require_relative '../commands/help'
 require_relative '../commands/list'
 require_relative '../commands/prop'
@@ -12,7 +13,6 @@ require_relative '../commands/remove'
 require_relative '../commands/sysinfo'
 require_relative '../lib/color'
 require_relative '../lib/const'
-require_relative '../lib/convert_size'
 require_relative '../lib/deb_utils'
 require_relative '../lib/docopt'
 require_relative '../lib/downloader'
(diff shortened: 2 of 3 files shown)


commands/files.rb
```diff
@@ -0,0 +1,42 @@
...
+require_relative '../lib/convert_size'
+
+class Command
+  def self.files(pkg)
+    # Check if the package is even installed first, as this is the most likely reason we cannot find a filelist.
+    device_json = JSON.load_file(File.join(CREW_CONFIG_PATH, 'device.json'))
+    if device_json['installed_packages'].none? { |entry| entry['name'] == pkg.name }
+      puts "Package #{pkg.name} is not installed.".lightred
+      return
+    end
```
uberhacker, line 12: We use this conditional in several places.  What are your thoughts of creating a generic function something like `if package_installed(pkg.name)` to abstract this kludgy logic?

uberhacker: Another method to determine if a package is installed is if the meta filelist exists.

uberhacker: In fact, it might make sense to include both checks in case device.json is corrupted.  If you want to take it a step further, you could also do some kind of self-healing to fix a corrupted device.json if the filelist is missing.  I would do a `crew remove` of the package before trying to update device.json, however, in cases where the files of the package are still hanging around.

commands/files.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    # We can't do anything if we don't have the filelist.
+    unless File.file?(filelist_path = File.join(CREW_META_PATH, "#{pkg.name}.filelist"))
+      puts "Package #{pkg.name} does not have a filelist :(".lightred
+      return
+    end
```
uberhacker, line 18: See comments above.

commands/files.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    # Print the name and description of the package.
+    puts pkg.name.lightgreen + ": #{pkg.description}".lightblue
```
uberhacker, line 21: I would think you would want the `:` lightgreen instead of lightblue.

commands/files.rb
```diff
@@ -0,0 +1,42 @@
...
+      next unless File.file?(filename)
+      # Ignore symlinks to prevent duplicating calculation.
+      next if File.symlink?(filename)
+      # Add the size of the file to the total size.
+      size += File.size(filename)
+    end
+
+    # Print the filelist, the total number of files, and the total size of those files.
+    puts filelist
+    puts "\nTotal found: #{filelist.size}".lightgreen
```
uberhacker, line 38: `.length` instead of `.size`?  I know they call the same function under the hood but `.length` is a more pragmatic method.

commands/files.rb
Zopolis4: No, this mirrors the current behaviour and is how print_current_package does it.

uberhacker: Just sayin'...

commands/files.rb
uberhacker: I'm saying abstract the logic into a single function instead of including both checks here.

commands/files.rb
uberhacker: Come to think of it, even `.length` is incorrect because it gives a byte count instead of line count.  Can we restore the previous logic?

Zopolis4: `filelist` is an array. This is the same logic as before. Would you prefer `.length` or `.size`?

commands/files.rb
Zopolis4: We only use it in two places right now-- I don't think the savings are worth it. 

I'll keep an eye on it as I add more commands though, and abstract if it makes sense.

uberhacker: It's not about savings so much as reuse.  We don't want to keep adding the same logic anytime we need to determine if a package is installed.  Is it too difficult to add?  I think @satmandu needed it for something recently.

commands/files.rb
uberhacker: How about `.count`?

Zopolis4: Yeah, that reads better.

commands/files.rb
Zopolis4: Fair enough-- that gives me an excuse for doing a `lib/package.rb` refactor that's been on the back of my mind, so I'll do that in another PR. I can update these files after I create and merge that PR, so don't feel the need to wait on it.

PR #9609, opened 2024-04-06
Add vulkansdk package

uberhacker:
Vulkan is a new generation graphics and compute API that provides high-efficiency, cross-platform access to modern GPUs used in a wide variety of devices from PCs and consoles to mobile phones and embedded platforms.  See https://vulkan.lunarg.com/doc/sdk.

Tested all binaries to make sure nothing segfaults in m121 container.


uberhacker:
This has a conflict with vulkan_tools.

packages/vulkansdk.rb
```diff
@@ -0,0 +1,64 @@
...
+      VK_LAYER_PATH=#{CREW_PREFIX}/share/vulkan/explicit_layer.d
+    EOF
+    Dir.chdir 'x86_64' do
+      # Remove conflicts with vulkan_icd_loader
+      FileUtils.rm Dir['lib/cmake/VulkanLoader/*.cmake']
+      FileUtils.rm Dir['lib/libvulkan.so*']
+      FileUtils.mv 'lib', 'lib64'
+      FileUtils.mv Dir['include/vulkan/*'], 'include'
+      FileUtils.rm_rf 'include/vulkan'
+      %w[bin include lib64 share].each do |dir|
```
satmandu, line 40: Do we need to worry about `lib` at all here? 

I.e. should that be `%w[bin include #{ARCH_LIB} share].each`...

satmandu (approved):

packages/vulkansdk.rb
uberhacker: No need since the only compatible architecture is x86_64.

satmandu (approved):

satmandu (approved):

PR #9611, opened 2024-04-06
Libxext 1.3.4 => 1.3.6

uberhacker:
Removed i686 since libx11 is no longer supported.  All tests passed.


packages/libxext.rb
```diff
@@ -1,32 +1,29 @@
...
   })
 
   depends_on 'llvm16_lib' => :build
+  depends_on 'xmlto' => :build
+  depends_on 'glibc' # R
+  depends_on 'libx11' # R
+  depends_on 'libxau' # R
+  depends_on 'libxcb' # R
+  depends_on 'libxdmcp' # R
+  depends_on 'xorg_proto' # R
```
satmandu, line 26: Alphabetize the depends?

satmandu (approved):

packages/libxext.rb
uberhacker: Done

packages/libxext.rb
```diff
@@ -1,32 +1,29 @@
...
+     armv7l: '2f74453c3b6d4dee2cd194c59134e0fd14f1dc3fcd240edfbd2976021de0c27d',
+     x86_64: '99bd31ebf5bf87c2587ac10bcafc7719a70629660dfb366b0362604a9228d6a1'
   })
 
+  depends_on 'glibc' # R
+  depends_on 'libx11' # R
+  depends_on 'libxau' # R
+  depends_on 'libxcb' # R
+  depends_on 'libxdmcp' # R
   depends_on 'llvm16_lib' => :build
```
satmandu, line 24: Oops! This should be updated too to `llvm18_lib`

satmandu (approved):

packages/libxext.rb
uberhacker: That would require new binaries, right?

satmandu: If getrealdeps.rb doesn't complain, then no rebuild is necessary.

uberhacker: Rebuilt binaries anyway to make sure it works.

satmandu (approved):

PR #9617, opened 2024-04-09
Add lib/package_utils.rb to provide a number of package convenience functions.

Zopolis4:
As discussed in #9601.

Over time, a number of convenience functions have collected in `lib/package.rb`, and more importantly, in the `Package` class. 

They shouldn't be there-- given that almost every operation we do entails passing one or more `Package` objects around, its a good idea not to include random chunks of logic in there. 

I'm splitting out the functions related to urls and sha256s for now (and `get_extract_dir`), as well as adding the `installed?` convenience method discussed previously.

Everything related to dependencies I will handle in my refactor of `crew deps`, and I'll deal with the redefinition of `system` in my `crew build` refactor.

`.source?` and by extension `.binary?` are more difficult, partiually due to the internal usage of `.source?` in `.get_deps_list`, and also due to the actual useage of `.source?` as opposed to `.build_from_source`. I'll take a look at those another time.

Tested and working on all architectures, with all tests passing.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=ontesid crew update
```

### DIFF
--- a/lib/package.rb
+++ b/lib/package.rb
@@ -150,15 +150,6 @@ class Package
     end
   end
 
-  def self.compatible?
-    if @compatibility
-      return @compatibility.casecmp?('all') || @compatibility.include?(ARCH)
-    else
-      warn "#{name}: Missing `compatibility` field.".lightred
-      return false
-    end
-  end
-
   def self.depends_on(dependency, ver_range = nil)
     @dependencies ||= {}
     ver_check = nil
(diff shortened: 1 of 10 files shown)


lib/package_convenience.rb
```diff
@@ -0,0 +1,33 @@
+require 'json'
+require_relative 'const'
+
+class PackageFunctions
```
supechicken, line 4: Maybe a shorter name is better for classes? Like `PackageUtils`

Zopolis4:
unit tests are breaking because somewhere along the line the changes made it so crew upgrade now checks fake packages, i think

lib/package_convenience.rb
uberhacker: Good idea and the class name should also match the filename to avoid confusion.

lib/package_convenience.rb
```diff
@@ -0,0 +1,33 @@
...
+  def self.installed?(pkg)
```
uberhacker, line 5: I'm thinking simply passing the package name would be better?  I only say this because currently it requires loading the package object prior to passing the parameter to the function and in this case, the object is not needed, only the package name.

tests/lib/package_convenience.rb
```diff
@@ -0,0 +1,127 @@
+require 'minitest/autorun'
+require_relative '../../lib/const'
+require_relative '../../lib/package'
+require_relative '../../lib/package_convenience'
+
+class PackageFunctionsTest < Minitest::Test
```
uberhacker, line 6: Keep the class name consistent with the filename?

lib/package_convenience.rb
Zopolis4: I'd prefer using the entire object for consistency, but in this case it does make more sense to just pass the package name.

lib/package.rb
```diff
@@ -202,38 +193,6 @@ def self.depends_on(dependency, ver_range = nil)
...
-      return @binary_sha256[architecture]
-    elsif @source_sha256.respond_to?(:has_key?)
-      return @source_sha256.key?(architecture) ? @source_sha256[architecture] : nil
-    else
-      return @source_sha256
-    end
-  end
-
-  def self.get_binary_sha256(architecture) = @binary_sha256&.key?(architecture) ? @binary_sha256[architecture] : ''
-  def self.get_extract_dir = "#{name}.#{Time.now.utc.strftime('%Y%m%d%H%M%S')}.dir"
```
uberhacker, line 235: This is not needed?

Zopolis4: nope

uberhacker (dismissed):
Resolve merge conflict with crew version?

uberhacker (approved):

PR #9623, opened 2024-04-09
Rsync 3.2.7 => 3.3.0

uberhacker:
No filelist updates.  All tests passed.

### DIFF
--- a/packages/rsync.rb
+++ b/packages/rsync.rb
@@ -1,20 +1,20 @@
-require 'package'
+require 'buildsystems/autotools'
 
-class Rsync < Package
+class Rsync < Autotools
   description 'rsync is an open source utility that provides fast incremental file transfer.'
   homepage 'https://rsync.samba.org/'
-  version '3.2.7'
+  version '3.3.0'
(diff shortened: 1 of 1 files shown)


packages/rsync.rb
```diff
@@ -30,16 +30,7 @@ class Rsync < Package
   no_patchelf
   no_zstd
 
-  def self.build
-    system "./configure #{CREW_OPTIONS} --disable-maintainer-mode"
-    system 'make'
-  end
+  configure_options '--disable-maintainer-mode'
```
Zopolis4, line 33: It seems like maintainer mode is disabled already (lines 122-126 of the configure.ac file in the prvoided tarball):
```
AC_ARG_ENABLE(maintainer-mode,
    AS_HELP_STRING([--enable-maintainer-mode],[enable to turn on extra debug features]))
if test x"$enable_maintainer_mode" = x"yes"; then
    CFLAGS="$CFLAGS -DMAINTAINER_MODE"
fi
```

Zopolis4 (approved):

PR #9624, opened 2024-04-10
Add create command

Zopolis4:
As discussed in #9594.

Pretty simple at the moment, only supports buildsystems, and you will still manually have to add dependencies yourself.

A step in the right direction though, and enough to provide the convenience that I'm taking away from `crew upload`.

It's hanging right now, and I'd welcome help figuring out why.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=sintered crew update
```


commands/create.rb
```diff
@@ -0,0 +1,42 @@
+class Command
+  def self.create
+    puts 'Example package names: libnvme, crew-profile-base'.lightblue
+    name = gets('Enter the package name:').chomp.capitalize.tr('-', '_')
```
uberhacker, line 4: Check for invalid package names?   First character must be a letter, only alphanumeric and - or _ allowed, search packages for duplicate names, etc.

commands/create.rb
```diff
@@ -0,0 +1,42 @@
...
+    puts 'Example package description: C Library for NVM Express on Linux.'.lightblue
+    description = gets('Enter the package description:').chomp
+    puts 'Example package homepage: https://github.com/linux-nvme/libnvme'.lightblue
+    homepage = gets('Enter the package homepage:').chomp
```
uberhacker, line 8: Must be a url?

commands/create.rb
```diff
@@ -0,0 +1,42 @@
...
+    puts 'Example package version: v1.0-rc4'.lightblue
+    version = gets('Enter the package version:').chomp
+    puts 'Example package license: LGPL-2.1+'.lightblue
+    license = gets('Enter the package license:').chomp
+    puts 'Example package source_url: https://github.com/linux-nvme/libnvme/archive/refs/tags/v1.0-rc4.tar.gz'.lightblue
+    source_url = gets('Enter the package source_url:').chomp
```
uberhacker, line 14: Must be a url?

commands/create.rb
```diff
@@ -0,0 +1,42 @@
...
+    puts 'Example package source_sha256: 82dc150c09e5e211d5200a37fabbbe25c69f5b0f22582956e51587a5a54463b5'.lightblue
+    source_sha256 = gets('Enter the package source_sha256:').chomp
```
uberhacker, line 16: Only hex characters allowed?

commands/create.rb
```diff
@@ -0,0 +1,42 @@
...
+    puts 'Example package buildsystem: meson'.lightblue
```
uberhacker, line 17: List all possible buildsystems?

uberhacker: Select from a list of options?

commands/create.rb
```diff
@@ -0,0 +1,42 @@
...
+    buildsystem = gets('Enter the package buildsystem:').chomp
```
uberhacker, line 18: Select from a list of options?

lib/const.rb
```diff
@@ -335,6 +335,7 @@
     crew build [options] [-k|--keep] [-v|--verbose] <name> ...
     crew check [-V|--version] [-v|--verbose] <name> ...
     crew const [-v|--verbose] [<name> ...]
+    crew create
```
uberhacker, line 338: Allow switch options to bypass interactive prompts?

commands/create.rb
```diff
@@ -0,0 +1,42 @@
...
+        source_url "'#{source_url}'"
+        source_sha256 "'#{source_sha256}'"
+        binary_compression 'tar.zst'
+
+        binary_sha256({
+         aarch64: 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
+          armv7l: 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
+            i686: 'iiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiii',
+          x86_64: 'xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
+        })
```
uberhacker, line 38: What if `no_compile_needed` is the intention?

commands/create.rb
```diff
@@ -0,0 +1,42 @@
...
+    license = gets('Enter the package license:').chomp
+    puts 'Example package source_url: https://github.com/linux-nvme/libnvme/archive/refs/tags/v1.0-rc4.tar.gz'.lightblue
+    source_url = gets('Enter the package source_url:').chomp
+    puts 'Example package source_sha256: 82dc150c09e5e211d5200a37fabbbe25c69f5b0f22582956e51587a5a54463b5'.lightblue
+    source_sha256 = gets('Enter the package source_sha256:').chomp
+    puts 'Example package buildsystem: meson'.lightblue
+    buildsystem = gets('Enter the package buildsystem:').chomp
+
+    puts <<~EOF
+      require "'buildsystems/#{buildsystem}'"
```
uberhacker, line 21: What about a standard package?

Zopolis4: As I said in the PR description, this only supports buildsystems for now-- a significant chunk of packages are buildsystems anyways, and those that aren't buildsystems will have unique `def self.build` sections that benefit less from a template.

commands/create.rb
Zopolis4: I don't see how anyone would enter an invalid sha256 value in this way.

uberhacker: Just another validation possibility.

commands/create.rb
uberhacker: `package` could be a choice and `buildsystems/` can be added to the others as a replacement pattern.  There are many packages that don't require buildsystems too.

commands/create.rb
uberhacker: Maybe have a prompt to include binaries?  Something like `Will this package require prebuilt binaries? (Y/n):`?

PR #9643, opened 2024-04-13
Ruby rebuild + cmake => 3.29.2

satmandu:
- Rebuilds `ruby` from `ruby_3_3` branch to fix https://github.com/advisories/GHSA-592j-995h-p23j
- Add `crewlog(string)` function to `crew`.
- Refactor json package removal for upgraded packages.
- Add `save_json` function to safely save the json file.
- Automatically rerun `crew upgrade` when `ruby` is being upgraded.
- Added `CREW_ESSENTIAL_PACKAGES = %w[gcc_lib glibc gmp ruby zlibpkg zstd]` to `const.rb`
- If `distcc` is installed, use `distcc -j` for `CREW_NPROC` calculation.
- grep usage in `commands/remove.rb` needs quoting for file names with spaces.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=ruby_cmake crew update ; yes | crew upgrade -v
```


satmandu:
Currently there's a problem where `CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=ruby_cmake crew update ; yes | crew upgrade` doesn't work:

```bash
chronos@rpi52-armv7l /usr/local/lib/crew/packages $ CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=ruby_cmake crew update ; yes | crew upgrade
remote: Enumerating objects: 43, done.
remote: Counting objects: 100% (43/43), done.
remote: Compressing objects: 100% (4/4), done.
remote: Total 26 (delta 22), reused 26 (delta 22), pack-reused 0
Unpacking objects: 100% (26/26), 15.37 KiB | 215.00 KiB/s, done.
From https://github.com/satmandu/chromebrew
 * branch                ruby_cmake -> FETCH_HEAD
HEAD is now at 111cbde75 cmake => 3.29.2
ruby could be updated from 3.3.0 to 3.3.0-2f65458
cmake could be updated from 3.29.0 to 3.29.2
                                                                                                                                                                                                              2 packages can be updated.
Run `crew upgrade` to update all packages or `crew upgrade <package1> [<package2> ...]` to update specific packages.
Updating packages...
ruby: Ruby is a dynamic, open source programming language with a focus on simplicity and productivity.
Cannot find cached archive. 😔 Will download.
Ruby archive downloaded.
Removing since upgrade or reinstall...
Refusing to remove essential package ruby.
Performing pre-install...
Performing install...
Using rdfind to convert duplicate files to hard links.
Now scanning ".", found 17612 files.
Now have 17612 files in total.
Removed 15 files due to nonunique device and inode.
Total size is 40856842 bytes or 39 MiB
Removed 1997 files due to unique sizes from list. 15600 files left.
Now eliminating candidates based on first bytes: removed 15481 files from list. 119 files left.
Now eliminating candidates based on last bytes: removed 68 files from list. 51 files left.
Now eliminating candidates based on sha1 checksum: removed 51 files from list. 0 files left.
It seems like you have 0 files that are not unique
Totally, 0 B can be reduced.
Now making hard links.
Making 0 links.
Performing post-install for ruby...
Updating ruby gems. This may take a while...
Ruby installed!
Ruby was upgraded. Please run 'crew upgrade' again to make sure upgrades are complete.
chronos@rpi52-armv7l /usr/local/lib/crew/packages $ crew upgrade
Updating packages...
ruby: Ruby is a dynamic, open source programming language with a focus on simplicity and productivity.
Archive found in cache
Removing since upgrade or reinstall...
Refusing to remove essential package ruby.
Performing pre-install...
Performing install...
Using rdfind to convert duplicate files to hard links.
Now scanning ".", found 17612 files.
Now have 17612 files in total.
Removed 15 files due to nonunique device and inode.
Total size is 40856842 bytes or 39 MiB
Removed 1997 files due to unique sizes from list. 15600 files left.
Now eliminating candidates based on first bytes: removed 15481 files from list. 119 files left.
Now eliminating candidates based on last bytes: removed 68 files from list. 51 files left.
Now eliminating candidates based on sha1 checksum: removed 51 files from list. 0 files left.
It seems like you have 0 files that are not unique
Totally, 0 B can be reduced.
Now making hard links.
Making 0 links.
Performing post-install for ruby...
Updating ruby gems. This may take a while...
Ruby installed!
Ruby was upgraded. Please run 'crew upgrade' again to make sure upgrades are complete.
chronos@rpi52-armv7l /usr/local/lib/crew/packages $
```

packages/ruby.rb
```diff
@@ -3,44 +3,43 @@
...
-  depends_on 'openssl' # R
   depends_on 'libyaml' # R
-  depends_on 'readline' # R
+  depends_on 'openssl' # R
   depends_on 'rust' => :build
-  depends_on 'ca_certificates'
-  depends_on 'libyaml' # This is needed to install gems
+  depends_on 'zlibpkg' # R
+
+  conflicts_ok
```
uberhacker, line 31: What is conflicting?  Is it really OK?  If there is another ruby package that has conflicting files, removing that package could brick crew?

packages/cmake.rb
```diff
@@ -3,23 +3,24 @@
...
+    aarch64: '06d2f226d1eff483d29ec8d861fb2c47a0200f82d30f121d872f89975ff70c2f',
+     armv7l: '06d2f226d1eff483d29ec8d861fb2c47a0200f82d30f121d872f89975ff70c2f',
+       i686: 'c4715d535d0637d37f4c5c7d7ae4a9f903baea88d8221848f9cc8ae9625781af',
+     x86_64: '5ad0c9ee32f7c058a6e31954b74d47cfa339f3c1184f00e68a9fa1100f1254af'
   })
 
   depends_on 'bzip2' => :build
   depends_on 'cppdap' # R
   depends_on 'curl' # R
+  depends_on 'distcc' => :build
```
Zopolis4, line 23: Does this really need to be the case?

packages/ruby.rb
satmandu: Only conflicts with Ruby ...

satmandu: Conflict code is wonky ... 🤦‍♂️😔

packages/cmake.rb
satmandu: Just trying to speed up the build. Tests take forever... Using distcc helps...

satmandu:
Added some debug code which shows this happening:
```bash
Removing since upgrade or reinstall...
Refusing to remove essential package ruby.
Performing pre-install...
Performing install...
Using rdfind to convert duplicate files to hard links.
Now scanning ".", found 17612 files.
Now have 17612 files in total.
Removed 15 files due to nonunique device and inode.
Total size is 43133583 bytes or 41 MiB
Removed 2002 files due to unique sizes from list. 15595 files left.
Now eliminating candidates based on first bytes: removed 15478 files from list. 117 files left.
Now eliminating candidates based on last bytes: removed 68 files from list. 49 files left.
Now eliminating candidates based on sha1 checksum: removed 49 files from list. 0 files left.
It seems like you have 0 files that are not unique
Totally, 0 B can be reduced.
Now making hard links.
Making 0 links.
before json change
{
  "name": "ruby",
  "version": "3.3.0",
  "binary_sha256": "a97a17336b4d233129fb5a55c82d32b38b981d017cda70ba3cbb51012bc129ca"
}
{
  "name": "ruby",
  "version": "3.3.0-2f65458",
  "sha256": "c554a429edbc1d104be2f7a149f91b3892e81605dc550f26a918c2f8bb749c22"
}
after json change
{
  "name": "ruby",
  "version": "3.3.0",
  "binary_sha256": "a97a17336b4d233129fb5a55c82d32b38b981d017cda70ba3cbb51012bc129ca"
}
{
  "name": "ruby",
  "version": "3.3.0-2f65458",
  "sha256": "c554a429edbc1d104be2f7a149f91b3892e81605dc550f26a918c2f8bb749c22"
}
{
  "name": "ruby",
  "version": "3.3.0-2f65458",
  "sha256": "c554a429edbc1d104be2f7a149f91b3892e81605dc550f26a918c2f8bb749c22"
}
Performing post-install for ruby...
```

Zopolis4:
Is this happening for any other package?

bin/crew
```diff
@@ -1283,15 +1283,18 @@ def install
   # Debug code...
   puts 'before json change'
   system "/usr/bin/jq --arg key ruby -e '.installed_packages[] | select(.name == \$key )' /usr/local/etc/crew/device.json"
+
+  # Clean out package from json before adding, especially if package may be in CREW_ESSENTIAL_PACKAGES.
+  @device[:installed_packages].delete_if { |entry| entry['name'] == @pkg.name }
```
uberhacker, line 1288: Why is this necessary?  Doesn't it just get added back in again?  Wouldn't a better approach be to check if the package exists and if not, add it?

satmandu: A second one gets added in if it isn't first removed.

We need
`If the name is in the array
  remove it
then add name
...
`

uberhacker: A second one is added only because you're not checking if it exists beforehand.

satmandu: Adjusted code to remove the entry if it is in `CREW_ESSENTIAL_PACKAGES`.

satmandu:
Everything seems to work now.

bin/crew
```diff
@@ -115,15 +115,34 @@ at_exit do
...
 
+def crewlog(str)
+  caller_locations(1, 1).first.tap {|loc| puts "#{loc.path}:#{loc.lineno}:#{caller_locations(3).first.label}:#{str}".lightpurple if @opt_verbose}
+end
+
 def load_json
   # load_json(): (re)load device.json
   json_path = File.join(CREW_CONFIG_PATH, 'device.json')
   @device   = JSON.load_file(json_path, symbolize_names: true)
+  @device_json = JSON.load_file(json_path)
```
uberhacker, line 126: I sure would like to keep this simple.  Why load twice?

bin/crew
```diff
@@ -1250,12 +1274,23 @@ def install
...
+  end
 
   # remove it just before the file copy
   if @pkg.in_upgrade
     puts 'Removing since upgrade or reinstall...'
     Command.remove(@pkg, @opt_verbose)
+
+    if CREW_ESSENTIAL_PACKAGES.include?(@pkg.name)
+      crewlog "Removing package #{@pkg.name} from device.json"
+      @device_json['installed_packages'].delete_if { |entry| entry['name'] == @pkg.name }
```
uberhacker, line 1291: Still don't quite understand why the package needs to be removed from device.json.

bin/crew
satmandu: Because there is different behavior whether `symbolize_names: true` is set or not. That's the prior behavior, and @Zopolis4 has been doing stuff that works without it being set, so I kept both until we figure out if we actually do need `symbolize_names: true` set everywhere...

bin/crew
```diff
@@ -1274,14 +1309,8 @@ def install
...
   # add to installed packages
-  @device[:installed_packages].push(name: @pkg.name, version: @pkg.version, sha256: PackageUtils.get_sha256(@pkg, build_from_source: @opt_source))
-  File.open("#{CREW_CONFIG_PATH}/device.json.tmp", 'w') do |file|
-    output = JSON.parse @device.to_json
-    file.write JSON.pretty_generate(output)
-  end
-  # Copy over original if the write to the tmp file succeeds.
-  FileUtils.cp "#{CREW_CONFIG_PATH}/device.json.tmp", File.join(CREW_CONFIG_PATH, 'device.json')
-  FileUtils.rm "#{CREW_CONFIG_PATH}/device.json.tmp"
+  @device_json['installed_packages'].push(name: @pkg.name, version: @pkg.version, sha256: PackageUtils.get_sha256(@pkg, build_from_source: @opt_source))
```
uberhacker, line 1312: Check if package is already in installed_packages prior to push?

lib/const.rb
```diff
@@ -115,8 +117,12 @@
...
 CREW_CACHE_FAILED_BUILD = ENV.fetch('CREW_CACHE_FAILED_BUILD', '0').eql?('1')
 
-# Set CREW_NPROC from environment variable or `nproc`
-CREW_NPROC = ENV.fetch('CREW_NPROC', `nproc`.chomp)
+# Set CREW_NPROC from environment variable, `distcc -j`, or `nproc`.
+CREW_NPROC = if File.file?("#{CREW_PREFIX}/bin/distcc")
+               ENV.fetch('CREW_NPROC', `distcc -j`.chomp)
+             else
+               ENV.fetch('CREW_NPROC', `nproc`.chomp)
+             end
```
uberhacker, line 125: What's the difference?

bin/crew
satmandu: Upgrade is the union of remove and install.

Remove takes out the package's line from the json file, except when one of the `CREW_ESSENTIAL_PACKAGES` is being dealt with, in which case remove returns early and the json file never gets updated.

This works around that, as otherwise, a new line gets added to the json file
This is a problem because the version detection for packages looks for the first package info returned from the json file, and since the old info is still there after an upgrade due to this bug, crew thinks that the package still needs to be upgraded after the new package is installed.

packages/cmake.rb
```diff
@@ -44,6 +44,9 @@ class Cmake < CMake
   # CMakeLib.testDebuggerNamedPipe-Script (armv7l,i686,x86_64)
   # RunCMake.CMakeRelease (armv7l,i686,x86_64)
   def self.check
+    @current_installed_cmake = `cmake --version | head -n 1 | awk '{print \$3}'`.chomp
```
uberhacker, line 47: What happens if cmake is not installed?

lib/const.rb
satmandu: With multiple distcc hosts, the number can be much higher than what nproc gives, as work can be offloaded to other machines...

bin/crew
satmandu: If the package is already installed we still need to delete the package info... We could do a check for that, if that would make more sense?

packages/ruby.rb
```diff
@@ -3,44 +3,41 @@
 class Ruby < Package
   description 'Ruby is a dynamic, open source programming language with a focus on simplicity and productivity.'
   homepage 'https://www.ruby-lang.org/en/'
-  version '3.3.0' # Do not use @_ver here, it will break the installer.
+  version '3.3.0-2f65458' # Do not use @_ver here, it will break the installer.
```
uberhacker, line 6: Does it make sense to wait until the next stable release?  I only say this because the maintainers probably have a good reason for not tagging a new release.

uberhacker (commented):
See comments.

bin/crew
Zopolis4: We don't need it set at all-- it's just a question of the way we access values (i.e. :name vs 'name')

bin/crew
Zopolis4: Removing the package should remove the package from the skin file, though.

packages/ruby.rb
Zopolis4: This is for a security fix, so it might make sense to get this through before a new stable.

bin/crew
Zopolis4: Ah, I see-- the removal is getting skipped, so the package isn't getting removed.

Zopolis4:
How does upgrading an essential package work then, given that we don't (to my knowledge) remove the currently running ruby binary?

bin/crew
uberhacker: What is the default for symbolize_names?  If we don't need to set it, let's not set it.  I'd be more inclined to want it to be flexible for symbols or names if that is possible.

packages/ruby.rb
satmandu: Also, this is from their stable branch... the update will trigger a gem rebuild for any vulnerable versions installed.

satmandu:
> How does upgrading an essential package work then, given that we don't (to my knowledge) remove the currently running ruby binary?

Crew should overwrite the package without removing it, thus never hopefully creating a moment when there aren't working binaries for the package on disk. (Which isn't entirely true, especially if doing a massive number of updates and dependencies get mucked up during the update... but that's arguably an edge case.)

bin/crew
satmandu: I've added a package installed check right before it is added.  (Right now we DO have a situation where the package keeps getting added...)

bin/crew
Zopolis4: It is not set by default (neither is our custom method to symbolize values), and no method for supporting both is available. 

Feel free to change it and all the code that relies on it, or I'll eventually refactor all of the code.

satmandu: Looks like just flipping it to not set it breaks things:
```
chronos@cheekon-x86_64 /usr/local/lib/crew/packages $ crew upgrade
/usr/local/bin/crew:394:in `upgrade': undefined method `each' for nil (NoMethodError)

    @device[:installed_packages].each do |pkg|
                                ^^^^^
        from /usr/local/bin/crew:1878:in `upgrade_command'
        from /usr/local/bin/crew:1908:in `<main>'
chronos@cheekon-x86_64 /usr/local/lib/crew/packages $
```

satmandu: uh... well, maybe I can fix that...

Zopolis4: by and large you'll have to replace :foo with 'foo' whenever json is concerned.

satmandu: not easily... I think this should be left for a cleanup in a future PR...

satmandu: @Zopolis4 Going in reverse, code like `@device[:installed_packages].delete_if { |entry| entry['name'] == @pkg.name }` doesn't work. Any idea how to fix that? (That's only used twice, so it might be easier to take care of those...)

satmandu: I think I got it working...

satmandu: Ok I removed it...

bin/crew
```diff
@@ -421,7 +439,12 @@ def upgrade(*pkgs, build_from_source: false)
   end
 
   puts 'Packages have been updated.'.lightgreen unless rerun_upgrade
-  puts "Ruby was upgraded. Please run 'crew upgrade' again to make sure upgrades are complete.".lightblue if rerun_upgrade
+  if rerun_upgrade
+    at_exit do
+      puts "Ruby was upgraded. Rerunning 'crew upgrade' to make sure upgrades are complete.".lightblue
+      exec 'crew upgrade'
+    end
```
uberhacker, line 446: Can we move line 441 to this conditional so it makes more logical sense?
```
else
  puts 'Packages have been updated.'.lightgreen
end
```

satmandu: Done.

uberhacker (approved):

PR #9649, opened 2024-04-14
Autosource ~/.bashrc

satmandu:
Inspired by @Zopolis4 's https://github.com/chromebrew/chromebrew/pull/9516, but without all the changes to `print_source_bashrc` in package files, so hopefully that PR can be merged after this since it has all the relevant changes in those files.

As suggested by @supechicken in https://github.com/chromebrew/chromebrew/pull/9516#issuecomment-2045854314

This is also dependent upon the enabling changes in https://github.com/chromebrew/crew-profile-base/pull/22

- With this, `~/.bashrc` no longer has to be manually sourced, except the first time this updated `crew_profile_base` is installed.
- Sets `selector.rb` timeout to `1` when run non-interactively.
- Fixed the `progress_bar` errors that were causing issues in the non-interactive unit tests.
- Moved `crewlog` to `lib/crewlog.rb` so I could call it from other files.
- Added `CREW_VERBOSE` constant so I wouldn't have to pass verbose to `crewlog`...

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=autosource crew update ; crew upgrade -v  ; setsid bash -c "crew reinstall crew_profile_base -v"
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -733,6 +733,7 @@ def post_install
     PRINT_SOURCE_BASHRC_EOT
       #To finish the installation, please execute the following:
       #source ~/.bashrc
+    crewlog "@pkg.print_source_bashrc?:#{@pkg.print_source_bashrc?} @pkg.gnome?:#{@pkg.gnome?}"
     if CREW_SOURCE_BASHRC_ENABLED
       Process.kill('USR1', Process.ppid)
       puts '~/.bashrc was automatically sourced.'.lightblue
--- a/bin/crew
+++ b/bin/crew
@@ -729,13 +729,17 @@ end
 def post_install
   if @pkg.print_source_bashrc? || @pkg.gnome?
     ExitMessage.add <<~PRINT_SOURCE_BASHRC_EOT.lightblue, print_last: true
-    ~/.bashrc will automatically be sourced.
+      ~/.bashrc was automatically sourced.
     PRINT_SOURCE_BASHRC_EOT
       #To finish the installation, please execute the following:
       #source ~/.bashrc
-
-    at_exit do
+    if CREW_SOURCE_BASHRC_ENABLED
       Process.kill('USR1', Process.ppid)
+      puts '~/.bashrc was automatically sourced.'.lightblue
+    else
+      at_exit do
+        exec 'bash'
+      end
     end
   end
 
--- a/bin/crew
+++ b/bin/crew
@@ -77,6 +77,7 @@ String.use_color = args['--color'] || !args['--no-color']
 @fileutils_verbose = @opt_verbose
 @verbose           = @opt_verbose ? 'v' : ''
 @short_verbose     = @opt_verbose ? '-v' : ''
+$verbose = @verbose
 
 class ExitMessage
   @messages = []
--- a/bin/crew
+++ b/bin/crew
@@ -116,8 +116,7 @@ at_exit do
     crewlog "@pkg.print_source_bashrc?:#{@pkg.print_source_bashrc?} @pkg.gnome?:#{@pkg.gnome?}"
     # Check to see if the trap was set, because otherwise, USR1 will
     # kill crosh.
-    # check if /tmp/.crew-autosource is opened in the parent shell as file descriptor 20
-    if File.file?("/proc/#{Process.ppid}/fd/20") && File.readlink("/proc/#{Process.ppid}/fd/20") == '/tmp/.crew-autosource'
+    if File.read("/proc/#{Process.ppid}/comm").include?('[trap set]')
       crewlog 'USR1 trap exists. Sourcing ~/.bashrc .'
       Process.kill('USR1', Process.ppid)
     else
--- a/bin/crew
+++ b/bin/crew
@@ -728,15 +728,12 @@ end
 
 def post_install
   if @pkg.print_source_bashrc? || @pkg.gnome?
-    ExitMessage.add <<~PRINT_SOURCE_BASHRC_EOT.lightblue, print_last: true
-      ~/.bashrc was automatically sourced.
-    PRINT_SOURCE_BASHRC_EOT
-      #To finish the installation, please execute the following:
-      #source ~/.bashrc
     crewlog "@pkg.print_source_bashrc?:#{@pkg.print_source_bashrc?} @pkg.gnome?:#{@pkg.gnome?}"
+    # Check to see if the env variable was set, because without the trap
+    # set, USR1 will kill crosh.
     if CREW_SOURCE_BASHRC_ENABLED
       Process.kill('USR1', Process.ppid)
-      puts '~/.bashrc was automatically sourced.'.lightblue
+      crewlog '~/.bashrc was automatically sourced.'
     else
       at_exit do
         exec 'bash'
--- a/bin/crew
+++ b/bin/crew
@@ -68,15 +68,14 @@ String.use_color = args['--color'] || !args['--no-color']
 
 @opt_force     = args['--force']
 @opt_keep      = args['--keep']
-@opt_verbose   = args['--verbose']
 @opt_source    = args['--source']
 @opt_recursive = args['--recursive-build']
 @opt_version   = args['--version']
 
 # Verbose options
-@fileutils_verbose = @opt_verbose
-@verbose           = @opt_verbose ? 'v' : ''
-@short_verbose     = @opt_verbose ? '-v' : ''
+@fileutils_verbose = CREW_VERBOSE
+@verbose           = CREW_VERBOSE ? 'v' : ''
+@short_verbose     = CREW_VERBOSE ? '-v' : ''
 
 class ExitMessage
   @messages = []
@@ -211,7 +210,7 @@ def list_packages
 end
 
 def generate_compatible
-  puts 'Generating compatible packages...'.orange if @opt_verbose
+  puts 'Generating compatible packages...'.orange if CREW_VERBOSE
   @device[:compatible_packages] = []
   Dir["#{CREW_PACKAGES_PATH}/*.rb"].each do |filename|
     pkg_name = File.basename filename, '.rb'
@@ -220,12 +219,12 @@ def generate_compatible
     rescue StandardError => e
       puts "Error with #{pkg_name}.rb: #{e}".red unless e.to_s.include?('uninitialized constant')
     end
-    puts "Checking #{pkg_name} for compatibility.".orange if @opt_verbose
+    puts "Checking #{pkg_name} for compatibility.".orange if CREW_VERBOSE
     if PackageUtils.compatible?(@pkg)
       # add to compatible packages
-      puts "Adding #{pkg_name} #{@pkg.version} to compatible packages.".lightgreen if @opt_verbose
+      puts "Adding #{pkg_name} #{@pkg.version} to compatible packages.".lightgreen if CREW_VERBOSE
       @device[:compatible_packages].push(name: @pkg.name)
-    elsif @opt_verbose
+    elsif CREW_VERBOSE
       puts "#{pkg_name} is not a compatible package.".lightred
     end
   end
@@ -233,7 +232,7 @@ def generate_compatible
     output = JSON.parse @device.to_json
     file.write JSON.pretty_generate(output)
   end
-  puts 'Generating compatible packages done.'.orange if @opt_verbose
+  puts 'Generating compatible packages done.'.orange if CREW_VERBOSE
 end
 
 def search(pkg_name, silent = false)
@@ -254,7 +253,7 @@ def regexp_search(pkg_pat)
   re = Regexp.new(pkg_pat, true)
   results = Dir["#{CREW_PACKAGES_PATH}/*.rb"] \
             .select  { |f| File.basename(f, '.rb') =~ re } \
-            .each    { |f| print_package(f, @opt_verbose) }
+            .each    { |f| print_package(f, CREW_VERBOSE) }
   if results.empty?
     Dir["#{CREW_PACKAGES_PATH}/*.rb"].each do |package_path|
       package_name = File.basename package_path, '.rb'
@@ -264,7 +263,7 @@ def regexp_search(pkg_pat)
         puts "Error with #{pkg_name}.rb: #{e}".red unless e.to_s.include?('uninitialized constant')
       end
       if @pkg.description =~ /#{pkg_pat}/i
-        print_current_package @opt_verbose
+        print_current_package CREW_VERBOSE
         results.push(package_name)
       end
     end
@@ -342,7 +341,7 @@ def update
   @device[:installed_packages].each do |package|
     search package[:name], true
     unless @pkg
-      puts "Package file for #{package[:name]} not found. :(".lightred if @opt_verbose
+      puts "Package file for #{package[:name]} not found. :(".lightred if CREW_VERBOSE
       next
     end
     different_version = (package[:version] != @pkg.version)
@@ -446,7 +445,7 @@ def upgrade(*pkgs, build_from_source: false)
     print_current_package
     @pkg.build_from_source = (build_from_source || CREW_BUILD_FROM_SOURCE)
 
-    puts "Updating #{@pkg.name}..." if @opt_verbose
+    puts "Updating #{@pkg.name}..." if CREW_VERBOSE
 
     @pkg.in_upgrade = true
     resolve_dependencies_and_install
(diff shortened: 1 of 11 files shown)


satmandu:
The unit tests appear to barf on the pager setting dialog in crew_profile_base...

lib/const.rb
```diff
@@ -133,6 +133,7 @@
 CREW_NOT_LINKS                       = ENV.fetch('CREW_NOT_LINKS', '0').eql?('1')                       # or use no_links
 CREW_NOT_STRIP                       = ENV.fetch('CREW_NOT_STRIP', '0').eql?('1')                       # or use no_strip
 CREW_NOT_SHRINK_ARCHIVE              = ENV.fetch('CREW_NOT_SHRINK_ARCHIVE', '0').eql?('1')              # or use no_shrink
+CREW_SOURCE_BASHRC_ENABLED           = ENV.fetch('CREW_SOURCE_BASHRC_ENABLED', '0').eql?('1')
```
uberhacker, line 136: If we're not using a property, why do we even need a constant?

satmandu: It is being used in the postinstall:
```ruby
def post_install
  if @pkg.print_source_bashrc? || @pkg.gnome?
    crewlog "@pkg.print_source_bashrc?:#{@pkg.print_source_bashrc?} @pkg.gnome?:#{@pkg.gnome?}"
    # Check to see if the env variable was set, because without the trap
    # set, USR1 will kill crosh.
    if CREW_SOURCE_BASHRC_ENABLED
      Process.kill('USR1', Process.ppid)
      crewlog '~/.bashrc was automatically sourced.'
    else
      at_exit do
        exec 'bash'
      end
    end
  end

  GnomePostinstall.add @pkg.name if @pkg.gnome?

  # return unless the postinstall function was defined by the package recipe
  return unless @pkg.method(:postinstall).source_location[0].include?("#{@pkg.name}.rb")

  Dir.mktmpdir do |post_install_tempdir|
    Dir.chdir post_install_tempdir do
      puts "Performing post-install for #{@pkg.name}...".lightblue
      @pkg.postinstall
    end
  end
end
```

satmandu: I will move it out of that section though...

uberhacker: So if you disable it, how does the user know they need to execute `source ~/.bashrc` manually?

satmandu: Variable renamed to `CREW_SOURCE_BASHRC_ACTIVATED` to be used for when `crew_profile_base` is upgraded, and the new USR1 trap isn't set yet because the new profile hasn't been loaded yet.

satmandu:
ruby invokes /bin/sh for shells, which does not support `trap -p`...

satmandu:
Ok this works by just explicitly running  `bash --noprofile --norc -c 'trap -p USR1|grep -v "source ~/.bashrc"'` from ruby.

bin/crew
```diff
@@ -728,11 +728,19 @@ end
...
   if @pkg.print_source_bashrc? || @pkg.gnome?
-    ExitMessage.add <<~PRINT_SOURCE_BASHRC_EOT.lightblue, print_last: true
-
-      To finish the installation, please execute the following:
-      source ~/.bashrc
-    PRINT_SOURCE_BASHRC_EOT
+    crewlog "@pkg.print_source_bashrc?:#{@pkg.print_source_bashrc?} @pkg.gnome?:#{@pkg.gnome?}"
+    # Check to see if the trap was set, because otherwise, USR1 will
+    # kill crosh.
+    if `bash --noprofile --norc -c 'trap -p USR1|grep -v "source ~/.bashrc"'`.chomp
```
supechicken, line 734: This will not work as signal traps will not inherit to child process like environment variables do.

satmandu: Sigh. You're right. Any other ways to do this aside from using an env variable?

supechicken: We can open a specific file in `bash` as an indication of having the `SIGUSR1` trap:
```bash
# open /tmp/.crew-autosource as file descriptor 20
exec 20> /tmp/.crew-autosource
```

... and then we can check which processes opened `/tmp/.crew-autosource` with `lsof`:
```shell
# lsof /tmp/.crew-autosource

COMMAND   PID        USER   FD   TYPE DEVICE SIZE/OFF   NODE NAME
bash    21825 supechicken   20w   REG   0,51        0 498210 /tmp/.crew-autosource
bash    21917 supechicken   20w   REG   0,51        0 498210 /tmp/.crew-autosource
```

As we can see, two shells opened the file. And we got their PID also :)

satmandu: > We can open a specific file in `bash` as an indication of having the `SIGUSR1` trap:

Is this simpler than just using an env variable?

@uberhacker?

satmandu: We could also dump `trap -p` into an env variable, which would be apparent in child shells...

Zopolis4:
I'm weighing up the increasing complexity of doing this with the memory cost of my more simple approach-- are we sure there is no way to replace the parent shell using only Ruby?

bin/crew
supechicken: The benefit of this approach over environment variables is that we can send the signal to all currently running shells instead of the parent shell only

supechicken:
> I'm weighing up the increasing complexity of doing this with the memory cost of my more simple approach-- are we sure there is no way to replace the parent shell using only Ruby?

AFAIK no :/ The only way to replace a process from outside of that process is with debugger, but that requires root

bin/crew
satmandu: I don't think we really need to worry about keeping the PID of the parent bash shell process? As such I have reimplemented the env variable based trap detection, as that is simple and is passed down through descendant shells.

satmandu:
OK all, this definitely works... I'm using the env variable method again for that reason:

Run this in a container:
```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=autosource crew update ; crew upgrade -v ; crew reinstall crew_profile_base -v
```

satmandu:
> I'm weighing up the increasing complexity of doing this with the memory cost of my more simple approach-- are we sure there is no way to replace the parent shell using only Ruby?

Replacing the parent shell sounds a lot lot more complicated... Can you do that anywhere?

bin/crew
supechicken: The point is not only on getting the parent shell PID, but getting the PID of **all** currently opened shell processes. That enables us to update the environment of all running shells, not only the parent shell.

If we decide to go ahead with using environment variables, then the `CREW_SCRIPTING` environment variable is necessary. (the `SIGUSR1` trap is not exist when the parent process is not a shell (e.g shell scripts), and sending `SIGUSR1` will likely kill them)

satmandu: Do we even want to update the env of all running shells?

Is that risky? I'm honestly not sure that crew is architecture such that if there are multiple spawned shells we want bashrc reloaded in all of them, because we don't want startup scripts run from those run again.

bin/crew
```diff
@@ -113,6 +113,21 @@ at_exit do
...
+    crewlog "@pkg.print_source_bashrc?:#{@pkg.print_source_bashrc?} @pkg.gnome?:#{@pkg.gnome?}"
+    # Check to see if the trap was set, because otherwise, USR1 will
+    # kill crosh.
+    if CREW_SOURCE_BASHRC_ACTIVATED
+      crewlog 'USR1 trap exists. Sourcing ~/.bashrc .'
+      Process.kill('USR1', Process.ppid)
+    else
+      at_exit do
+        crewlog 'USR1 trap does not exist. Spawning shell to source ~/.bashrc .'
+        exec 'bash'
```
supechicken, line 127: This would block the parent process if the parent process is not a shell.

bin/crew
satmandu: Doing exactly that, except doing a File.file? check on the fd first. :)

install.sh
```diff
@@ -370,3 +370,5 @@ You may wish to edit the ${CREW_PREFIX}/etc/env.d/02-editor file for an editor d
 Chromebrew provides nano, vim and emacs as default TUI editor options."
 
 echo_success "Chromebrew installed successfully and package lists updated."
+# Exit into a new shell so that the source_bashrc trap in ${CREW_PREFIX}/etc/profile is loaded.
+exec "${CREW_PREFIX}/bin/bash"
```
supechicken, line 374: This should be no longer needed after #9654

lib/progress_bar.rb
```diff
@@ -86,7 +86,9 @@ def show
         sleep 0.15 # update progress bar after each 0.15 seconds
 
         completed_length = (@bar_width * (@percentage / 100)).to_i
+        puts "completed_length: #{completed_length}"
         uncompleted_length = @bar_width - completed_length
+        puts "uncompleted_length: #{uncompleted_length}"
```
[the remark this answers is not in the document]
satmandu: I'm adding some debug code just to figure out where the negative values are coming from... I'm getting closer to tracking this down...

satmandu:
> Oh no, I forgot that fds are inherited by default...
> 
> Ugh. We need to find another way to do it...

I think it is working though?

supechicken:
> > Oh no, I forgot that fds are inherited by default...
> > Ugh. We need to find another way to do it...
> 
> I think it is working though?

It is working but not properly :D
```shell
$ exec 20>/dev/null
$ ls /proc/$$/fd
0  1  2  20  255  32

# fd 20 inherited to the child...
$ bash -c 'ls /proc/$$/fd/'
0  1  2  20  3  32
```
I am digging in `/proc` to see if there are any other solutions to this...

supechicken:
> @supechicken See the handoff log:
> 
> ```
> @terminal_w is 0
> @bar_width is -38
> @info_before_bar.merge(@info_after_bar).values.sum is 32
> (@info_before_bar.merge(@info_after_bar).length * 2) is 6
> 25l@bar_width is -38
> @percentage is 100.0
> completed_length: -38
> uncompleted_length: 0
> ```

`@terminal_w` should not be `0` in normal case, are you running in docker?

satmandu:
> > @supechicken See the handoff log:
> > ```
> > @terminal_w is 0
> > @bar_width is -38
> > @info_before_bar.merge(@info_after_bar).values.sum is 32
> > (@info_before_bar.merge(@info_after_bar).length * 2) is 6
> > 25l@bar_width is -38
> > @percentage is 100.0
> > completed_length: -38
> > uncompleted_length: 0
> > ```
> 
> `@terminal_w` should not be `0` in normal case, are you running in docker?

This is from the github unit test run...

supechicken:
> I am digging in /proc to see if there are any other solutions to this...

Found it😃 We can identify if a trap is set by changing the process name:
```shell
$ cat /proc/$$/comm
# original process name here
bash

# now if we change the process name...
$ echo -n "bash [autosource trap set]" > /proc/$$/comm

$ ps -o comm $$
COMMAND
bash [autosource trap set]

# you can get the exact same result with
$ cat /proc/$$/comm
bash [autosource trap set]
```

In ruby:
```ruby
if File.read("/proc/#{Process.ppid}/comm").include?('[autosource trap set]')
```

satmandu:
> > I am digging in /proc to see if there are any other solutions to this...
> 
> Found it😃 We can identify if a trap is set by changing the process name:
> 
> ```shell
> $ cat /proc/$$/comm
> # original process name here
> bash
> 
> # now if we change the process name...
> $ echo -n "bash [autosource trap set]" > /proc/$$/comm
> 
> $ ps -o comm $$
> COMMAND
> bash [autosource trap set]
> 
> # you can get the exact same result with
> $ cat /proc/$$/comm
> bash [autosource trap set]
> ```
> 
> In ruby:
> 
> ```ruby
> if File.read("/proc/#{Process.ppid}/comm").include?('[autosource trap set]')
> ```

Is... it dangerous to change the bash process name? Does anything depend upon that?

supechicken:
`bash` also works with different executable names, so it should be fine:
```shell
$ ln -s /bin/bash test-exe
$ exec ./test-exe
$ cat /proc/$$/comm
test-exe
```

But notice that the maximum length of `comm` is 15 characters, the best option is to use `bash [trap set]` as the new title name.

satmandu:
Hmm... everything seems to be working, but the unit tests are now broken?

supechicken:
Umm, is this expected to trigger an error?

https://github.com/chromebrew/chromebrew/blob/3ce7c0854fd956a377fa79994ba705f8c7e831a0/packages/hello_world_chromebrew.rb#L11-L20

satmandu:
This is also failing on a standard container at  the moment:
```
git clone --depth=1 https://github.com/chromebrew/chromebrew.git build_test && \
            cd build_test && \
            yes | crew build -f packages/hello_world_chromebrew.rb
```

Hmmm...

satmandu:
I enabled `CREW_CACHE_ENABLED=1` for the unit test package build, since that's also the default for the containers I build with. Is that a problem for anyone? @uberhacker ?

Zopolis4:
> I enabled `CREW_CACHE_ENABLED=1` for the unit test package build, since that's also the default for the containers I build with. Is that a problem for anyone? @uberhacker ?

It won't speed anything up (in fact it might marginally slow things down), so I don't really see a reason to use it?

Zopolis4:
> > Umm, is this expected to trigger an error?
> 
> > 
> 
> > https://github.com/chromebrew/chromebrew/blob/3ce7c0854fd956a377fa79994ba705f8c7e831a0/packages/hello_world_chromebrew.rb#L11-L20
> 
> 
> 
> It hasn't been an issue before... so why is it failing now?

Hm, I encountered this issue previously, but I thought I fixed it.

Zopolis4 (dismissed):

bin/crew
```diff
@@ -76,6 +77,7 @@ String.use_color = args['--color'] || !args['--no-color']
 @fileutils_verbose = @opt_verbose
 @verbose           = @opt_verbose ? 'v' : ''
 @short_verbose     = @opt_verbose ? '-v' : ''
+$verbose = @verbose
```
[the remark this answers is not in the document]
satmandu: We set verbose once globally with `-v`. It seems to be painful to constantly pass it around?

@uberhacker @supechicken thoughts?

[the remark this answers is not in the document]
Zopolis4: Global variables aren't good practice, though. (We have them disabled in rubocop right now but that was a temporary measure)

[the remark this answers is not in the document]
supechicken: I would suggest setting it as a constant and put it in `const.rb` in this case. As @Zopolis4 said, we should prevent using global variables in most cases.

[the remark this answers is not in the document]
satmandu: I might be able to work around setting a global variable for verbose by just using `unless %w[-v --verbose].include?(ARGV)` which is basically the same thing, and is, for the record, a hack. :)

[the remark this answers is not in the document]
satmandu: I took out the global variable.

satmandu:
> > I enabled `CREW_CACHE_ENABLED=1` for the unit test package build, since that's also the default for the containers I build with. Is that a problem for anyone? @uberhacker ?
> 
> It won't speed anything up (in fact it might marginally slow things down), so I don't really see a reason to use it?

The install of the just built package with `-f` breaks otherwise...

The ur-problem here is that this is a workaround to get just-built packages installed using the just-built archive file.

lib/progress_bar.rb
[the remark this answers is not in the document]
satmandu: This should now be resolved as those debug lines are no longer there.

bin/crew
```diff
@@ -129,7 +142,7 @@ def load_json
 end
 
 def save_json(json_object)
-  crewlog 'Saving device.json...'
+  # crewlog 'Saving device.json...'
```
uberhacker, line 145: Uncomment?  If verbose is not set, this won't print anyway, right?

satmandu: I was just try to clean up stuff I wasn't using to debug any more... I should just delete these...

satmandu: I'll just uncomment them...

lib/selector.rb
```diff
@@ -10,7 +12,15 @@ class Selector
 
   def initialize(options, prompt = @@default_prompt, timeout = 10)
     @options = options
-    @timeout = timeout
+    # Set timeout to zero if a non-interactive console.
+    # Check noninteractive usage with `setsid command`.
+    @timeout = if !IO.console&.console_mode || IO.console&.winsize == [0, 0]
+      # crewlog "IO.console&.console_mode is #{IO.console&.console_mode}"
+      # crewlog "IO.console&.winsize is #{IO.console&.winsize}"
```
uberhacker, line 19: Uncomment?

satmandu: Removed! They were there to help debug a progress_bar issue.

uberhacker (approved):

PR #9654, opened 2024-04-15
install.sh: Add message on error exit and simplify installation command

supechicken:
## Changes
- Added an error trap that prints a message when the script exits on error occurs
- Remove silent flag in `curl` so that network errors can be shown
- Makes the bootstrap command more straightforward by removing the `exec` hack and source `~/.bashrc` in the parent shell

```shell
bash <(curl -L https://raw.githubusercontent.com/supechicken/chromebrew/install_sh_error_handling/install.sh) && . ~/.bashrc
```


supechicken:
Umm... Should I take care of shellcheck suggestions? Those `|| exit 1` suggestions sounds unnecessary.

uberhacker:
Shellcheck is complaining @supechicken.

supechicken:
> Shellcheck is complaining @supechicken.

Read my previous comment...

uberhacker:
> > Shellcheck is complaining @supechicken.
> 
> Read my previous comment...

Can we disable rules in the .yml config?

supechicken:
> Can we disable rules in the .yml config?

Done

Zopolis4:
SC2164 is supposed to be disabled by set -e, which I thought we had.

supechicken:
> SC2164 is supposed to be disabled by set -e, which I thought we had.

I think you are right, as I replaced `set -e` with `set -E`. Let me add it back...

Zopolis4:
If we source ~/.bashrc in the install command then we don't need to tell people to do it manually.

supechicken:
> If we source ~/.bashrc in the install command then we don't need to tell people to do it manually.

Removed `source ~/.bashrc` prompt

install.sh
```diff
@@ -12,6 +12,9 @@ echo_success() { echo -e "\e[1;32m${*}${RESET}" >&1; } # Use Green for success m
 echo_intra() { echo -e "\e[1;34m${*}${RESET}" >&1; } # Use Blue for intrafunction messages.
 echo_out() { echo -e "\e[0;37m${*}${RESET}" >&1; } # Use Gray for program output.
 
+# Print a message before exit on error
+trap "echo_error 'An error occured during the installation :/'" ERR
```
uberhacker, line 16: Add color red?

uberhacker: Nevermind.  That's what `echo_error` does already. :+1:

README.md
```diff
@@ -57,10 +57,10 @@ You will need a Chromebook with developer mode enabled. To do so, select your de
 
 Open a VT-2 terminal session with <kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>-></kbd> and login with the `chronos` user and password if set [above](#set_passwd). *(if you are unable to do this, please have a second look at the prerequisites and make sure your Chromebook is in developer mode)*
 
-Then run the installation script below:
+Then download and run the installation script:
```
uberhacker, line 60: We're just running it, right?  If we weren't redirecting it to bash then it would be considered a  download.

satmandu (approved):

satmandu:
Please merge master and modify line 60 as per @uberhacker 's request so we can merge this?

README.md
supechicken: Change reverted

satmandu:
Please merge master again, as we just updated this for M123...

uberhacker (approved):

PR #9658, opened 2024-04-16
Rework load_package, set_package, search and copy_package

Zopolis4:
Pulling out my changes from #9594 because I need them for refactoring `crew check`.

Including my description of those changes here:
- Rework `load_package` in `lib/package.rb` to only take one argument, being the path to the package file.
  - Derive the package name in the function itself, rather than the weird setup we had before with a default argument value (when would we ever want to pass a different value to that?).
- Rework `set_package` in to only take one argument, , being the path to the package file.
  - All it was using the previous argument for was to unnecessarily pass it to `load_package`.
- Rework `search` to accept a package name, path, and a value for silent.
  - Use named parameters for the last two to save boilerplate.
  - Use a default value for `pkg_path`, as it is a named parameter rather than just being an optional one.
- Move the rubocop section that was previously in `def upload` to `def copy_package`.
  - Better to sanitize the package file earlier in the process.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=teparate crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -158,7 +158,7 @@ end
 def print_package(pkg_path, extra = false)
   pkg_name = File.basename pkg_path, '.rb'
   begin
-    set_package pkg_name, pkg_path
+    set_package pkg_path
   rescue StandardError => e
     warn "Error with #{pkg_name}.rb: #{e}".red unless e.to_s.include?('uninitialized constant')
   end
@@ -193,9 +193,9 @@ def print_current_package(extra = false)
   puts ''
 end
 
-def set_package(pkg_name, pkg_path)
+def set_package(pkg_path)
   begin
-    @pkg = Package.load_package(pkg_path, pkg_name)
+    @pkg = Package.load_package(pkg_path)
   rescue SyntaxError => e
     warn "#{e.class}: #{e.message}".red
   end
@@ -215,7 +215,7 @@ def generate_compatible
   Dir["#{CREW_PACKAGES_PATH}/*.rb"].each do |filename|
     pkg_name = File.basename filename, '.rb'
     begin
-      set_package pkg_name, filename
+      set_package filename
     rescue StandardError => e
       puts "Error with #{pkg_name}.rb: #{e}".red unless e.to_s.include?('uninitialized constant')
     end
@@ -235,10 +235,9 @@ def generate_compatible
   puts 'Generating compatible packages done.'.orange if CREW_VERBOSE
 end
 
-def search(pkg_name, silent = false)
-  pkg_path = File.join(CREW_PACKAGES_PATH, "#{pkg_name}.rb")
+def search(pkg_name, pkg_path: File.join(CREW_PACKAGES_PATH, "#{pkg_name}.rb"), silent: false)
   begin
-    return set_package(pkg_name, pkg_path) if File.file?(pkg_path)
+    return set_package(pkg_path) if File.file?(pkg_path)
   rescue StandardError => e
     puts "Error with #{pkg_name}.rb: #{e}".lightred unless e.to_s.include?('uninitialized constant')
   end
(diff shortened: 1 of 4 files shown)


Zopolis4:
The problem with deriving the path in the function is that means we can't set a different path if we need to (which we need to do for crew upload, build and check).

And passing the path as a second optional parameter is what we did earlier, but given that you can derive the name from the path there's no reason to have two parameters.

uberhacker:
> The problem with deriving the path in the function is that means we can't set a different path if we need to (which we need to do for crew upload, build and check).
> 
> And passing the path as a second optional parameter is what we did earlier, but given that you can derive the name from the path there's no reason to have two parameters.

I don't think you are understanding what I'm suggesting.  The second parameter for the path is optional in case you want to override the derived path.  It's easier to pass the package name as the first parameter.  I'm just saying that should be the first, not second parameter like we have now.

bin/crew
```diff
@@ -193,9 +193,9 @@ def print_current_package(extra = false)
   puts ''
 end
 
-def set_package(pkg_name, pkg_path)
+def set_package(pkg_path)
```
supechicken, line 196: Maybe we can go one step further and merge `def set_package()` and `def search()`? iirc `def search()` is just a wrapper for `def set_package()` with `rescue` block

uberhacker: I'm with you.  I like making things simpler.

Zopolis4: And we don't even need the rescue block, since it's covered by prop_test.

Zopolis4: I think we sometimes use it without the rescue block, but I'm not sure if there's a valid reason to do this.

bin/crew
```diff
@@ -1625,10 +1623,18 @@ def update_sha256(package, key, value)
 end
 
 def copy_package(pkg_name, prompt_msg = '')
+  pkg_file = File.join(CREW_LOCAL_REPO_ROOT, 'packages', "#{pkg_name}.rb")
+  # Use rubocop to sanitize package file, and let errors get flagged.
+  if @device[:installed_packages].any? { |elem| elem[:name] == 'ruby_rubocop' }
+    puts "Using rubocop to sanitize #{pkg_file} .".orange
+    system "rubocop -c #{File.join(CREW_LOCAL_REPO_ROOT, '.rubocop.yml')} -A #{pkg_file}", exception: true
```
uberhacker, line 1630: After thinking about this for a bit, I'm not so sure the `-A` option to actually change the file is a good idea.  Maybe a check and report would be better?  The reason I say this is because rubocop might get it wrong and it's not obvious to the user that the package was updated.

bin/crew
```diff
@@ -1625,10 +1623,18 @@ def update_sha256(package, key, value)
 end
 
 def copy_package(pkg_name, prompt_msg = '')
+  pkg_file = File.join(CREW_LOCAL_REPO_ROOT, 'packages', "#{pkg_name}.rb")
+  # Use rubocop to sanitize package file, and let errors get flagged.
+  if @device[:installed_packages].any? { |elem| elem[:name] == 'ruby_rubocop' }
```
uberhacker, line 1628: I sure wish we had something like `if package_installed?('ruby_rubocop')`.  That would be more elegant.

bin/crew
Zopolis4: I thought the point of this was to make things easier for new users-- I assume automatically fixing errors would be better, but I'm not too opinionated either way.

uberhacker: It is definitely an opinionated approach but we to need to think about the best way to do this without making it confusing.  I'll go ahead and merge this to see how it works out and if it becomes an issue, we can change the process then.

bin/crew
uberhacker: This can be the next pull request?

lib/package.rb
```diff
@@ -35,9 +35,10 @@ class << self
     attr_accessor :name, :cached_build, :in_build, :build_from_source, :in_upgrade
   end
 
-  def self.load_package(pkg_file, pkg_name = File.basename(pkg_file, '.rb'))
+  def self.load_package(pkg_file)
```
uberhacker, line 38: After analyzing this a bit more, I think passing the full path to the file is fine in this case since it is only used internally.

uberhacker (approved):

PR #9673, opened 2024-04-18
Update crew_profile_base => 0.0.18

satmandu:
- Updates some ruby gems.
- Sets ruby gems to update to latest version
- fixes crew bash loading in /usr/local/etc/profile
- Adds workaround that _may work_ in some cases when glibc gets updated from under crew.
- removes extra dbus env.d file from `crew_profile_base` that has been moved to the dbus package
- Fixes crew upgrade of `no_compile_needed` packages.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=crewprofilebase crew update
```

### DIFF
--- a/lib/buildsystems/ruby.rb
+++ b/lib/buildsystems/ruby.rb
@@ -15,7 +15,7 @@ class RUBY < Package
     @gem_ver = version.split('-', 2).first
     @ruby_ver = version.split('-', 3).last
     system "yes | gem uninstall -a #{@gem_name}", exception: false
-    system "gem install -i #{CREW_DEST_LIB_PREFIX}/ruby/gems/#{@ruby_ver}.0 -n #{CREW_DEST_PREFIX}/bin -N -v #{@gem_ver} #{@gem_name}", exception: false
+    system "gem install -i #{CREW_DEST_LIB_PREFIX}/ruby/gems/#{@ruby_ver}.0 -n #{CREW_DEST_PREFIX}/bin -N #{@gem_name}", exception: false
     eval @ruby_install_extras if @ruby_install_extras
   end
 end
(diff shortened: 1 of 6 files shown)


lib/buildsystems/ruby.rb
```diff
@@ -15,7 +15,7 @@ def self.install
     @gem_ver = version.split('-', 2).first
     @ruby_ver = version.split('-', 3).last
     system "yes | gem uninstall -a #{@gem_name}", exception: false
-    system "gem install -i #{CREW_DEST_LIB_PREFIX}/ruby/gems/#{@ruby_ver}.0 -n #{CREW_DEST_PREFIX}/bin -N -v #{@gem_ver} #{@gem_name}", exception: false
+    system "gem install -i #{CREW_DEST_LIB_PREFIX}/ruby/gems/#{@ruby_ver}.0 -n #{CREW_DEST_PREFIX}/bin -N #{@gem_name}", exception: false
```
uberhacker, line 18: Why remove `@gem_ver`?

satmandu: Without that, the most current version gets installed. This is how our pip installs work.

uberhacker: So this means, potentially, the package version might not match the actual installed version?

uberhacker: What happens in the case where you might want to lock a version if the most recent version causes an issue?

satmandu: We can modify `buildtools/ruby.rb` if that's ever an issue... but it certainly hasn't been an issue for any of the pip packages, and this has in fact hurt us by letting us miss a rubocop update that would have let us avoid an issue with the previous version of rubocop where it failed to check the glibc files properly.

satmandu: (Having said that, we desperately need a script that automatically updates versions in package files and automatically creates PRs based on version updates.)

uberhacker: Next task on the list?

uberhacker (approved):

PR #9696, opened 2024-04-20
add future chromeos compatablility

koolkid327:
Fixes # copy glibc to /usr/local

## Additional information

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/koolkid327/chromebrew.git CREW_BRANCH=patch-1 crew update
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -169,9 +169,9 @@ BOOTSTRAP_PACKAGES='zstd crew_mvdir ruby git ca_certificates libyaml openssl'
 [[ "${ARCH}" == "i686" ]] && BOOTSTRAP_PACKAGES+=' zlibpkg gcc_lib'
 
 if [[ -n "${CHROMEOS_RELEASE_CHROME_MILESTONE}" ]]; then
-  sudo cp /lib$LIB_SUFFIX/libc.so* /usr/local/
-  sudo chown chronos /usr/local/lib$LIB_SUFFIX/libc.so*
-  sudo chmod 777 /usr/local/lib$LIB_SUFFIX/libc.so*
+  sudo cp /lib$LIB_SUFFIX/libc.so* $CREW_PREFIX/lib$LIB_SUFFIX
+  sudo chown chronos $CREW_PREFIX/lib$LIB_SUFFIX/libc.so*
+  sudo chmod 777 $CREW_PREFIX/lib$LIB_SUFFIX/libc.so*
   if (( "${CHROMEOS_RELEASE_CHROME_MILESTONE}" > "112" )); then
     # Recent Arm systems have a cut down system.
     [[ "${ARCH}" == "armv7l" ]] && BOOTSTRAP_PACKAGES+=' bzip2 ncurses readline pcre2 gcc_lib'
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -151,7 +151,7 @@ function curl () {
 }
 
 # This will create the directories.
-crew_folders="bin cache doc docbook include lib/crew/packages lib$LIB_SUFFIX libexec man sbin share var etc/crew/meta etc/env.d tmp/crew/dest"
+crew_folders="lib64 bin cache doc docbook include lib/crew/packages lib$LIB_SUFFIX libexec man sbin share var etc/crew/meta etc/env.d tmp/crew/dest"
```
uberhacker, line 154: `$LIB_SUFFIX` abstracts the library folder which is not `lib64` on arm and i686 architecture.  Is `$LIB_SUFFIX` not set to `64` on your system?

install.sh
```diff
@@ -169,6 +169,9 @@ BOOTSTRAP_PACKAGES='zstd crew_mvdir ruby git ca_certificates libyaml openssl'
 [[ "${ARCH}" == "i686" ]] && BOOTSTRAP_PACKAGES+=' zlibpkg gcc_lib'
 
 if [[ -n "${CHROMEOS_RELEASE_CHROME_MILESTONE}" ]]; then
+  sudo cp /lib64/libc.so* /usr/local/
+  sudo chown chronos /usr/local/lib64/libc.so*
+  sudo chmod 777 /usr/local/lib64/libc.so*
```
uberhacker, line 174: This only works for x86_64 architecture and we need to abstract `/usr/local` with `$CREW_PREFIX`.  What issue are you experiencing to require this change?  Maybe an issue should be opened to further explain your installation trouble?

install.sh
koolkid327: done!

install.sh
koolkid327: the one that you just fixed

install.sh
```diff
@@ -169,9 +169,9 @@ BOOTSTRAP_PACKAGES='zstd crew_mvdir ruby git ca_certificates libyaml openssl'
 [[ "${ARCH}" == "i686" ]] && BOOTSTRAP_PACKAGES+=' zlibpkg gcc_lib'
 
 if [[ -n "${CHROMEOS_RELEASE_CHROME_MILESTONE}" ]]; then
-  sudo cp /lib$LIB_SUFFIX/libc.so* /usr/local/
-  sudo chown chronos /usr/local/lib$LIB_SUFFIX/libc.so*
-  sudo chmod 777 /usr/local/lib$LIB_SUFFIX/libc.so*
+  sudo cp /lib$LIB_SUFFIX/libc.so* $CREW_PREFIX/lib$LIB_SUFFIX
+  sudo chown chronos $CREW_PREFIX/lib$LIB_SUFFIX/libc.so*
+  sudo chmod 777 $CREW_PREFIX/lib$LIB_SUFFIX/libc.so*
```
uberhacker, line 174: Can we change the 777 to 644?  These libraries should not need execute or other group write permission.

koolkid327: oh

uberhacker: What are your thoughts about this proposed change @satmandu and @supechicken?

koolkid327: so that we will have to make less every time google makes 100 chromeos updates

install.sh
satmandu: With `lib$LIB_SUFFIX` we don't need `lib64`...

satmandu (approved):
Seems like it shouldn't break anything...

uberhacker:
@koolkid327:  Not sure if you noticed but shellcheck is complaining about missing double quotes.

uberhacker (approved):

PR #9707, opened 2024-04-22
Split check command into separate file

Zopolis4:
**Split out `check` command to `commands/check.rb`.**

As I have split out many of these changes into separate PRs, the ones that are already in master/no longer relevant have been crossed out.

Behavior changes:
- ~~Replace `crew test` with a Rakefile and `crew check`.~~
- Rewrite the message printed when `ruby_rubocop` is not installed.
- Remove the `crew check -v` behaviour.
  - Calling `version.rb` with the name of the package works the same, and dropping this wrapper allows us to drop the `-V` parameter which was at cross purposes with the same parameter when used like so: `crew -V`.
- ~~Make `prop_test` and `buildsystem_test` less verbose by default.~~
  - ~~This also makes it a lot easier to read the unit test logs.~~
- Fix the help message for `crew check`.
- Remove `zstd` instead of `zlib` in `test_force_remove_essential_package` in `tests/commands/remove.rb`, as a missing zlib causes rubocop failures on i686.
  
Refactoring changes:
(`prop_test` and `buildsystem_test`)
- ~~Return literal '1' and '0' and add those rather than using `@tofail`.~~
- ~~Use `CREW_PACKAGES_PATH` and `CREW_LIB_PATH` rather than relative directory paths.~~
- ~~Take an actual package object in `check_properties` rather than using a name and an instance variable.~~
- ~~Use less single-use variables in and around `check_buildsystem`.~~

(`bin/crew` and `commands/check.rb`)
- Merge `check_package` and `copy_package` into a single function, using the `to_copy` variable to replicate the previous method of outsourcing both the prompting for input and subsequent copying to `copy_package`.
  - As `crew build`, `crew install` and `crew reinstall` all now call `Command.check`, we rework the case of there being no local package to check to result in a silent ignore when we are calling it from another method (so you can install packages without having to worry about being in `CREW_LOCAL_REPO_ROOT`, but we let the user know about this happening when it is a call from `crew check`.
- Use `CREW_LIB_PATH` instead of relative paths.
- Use `File.join` instead of string interpolation.

(`prop_test`, `buildsystem_test` and `dep_test`)
- Return a failure code (when tests fail) when ran against a single file.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula8 crew update
```


Zopolis4:
The unit tests are failing because the docopt string for `crew build` (and also `crew install` and `crew reinstall`) are broken and don't have anything for `[-f|--force]`. I'm not sure why this is only becoming an issue now, but on that note, does it even make sense to be running `crew check` when building, installing or reinstalling?

I get that `crew build` should have `[-f|--force]` because of the `release/#{ARCH}` stuff, and i'll add it because of that, but I'm questioning the logic behind running this when we build or install and so on.

lib/const.rb
```diff
@@ -346,8 +346,8 @@
   Chromebrew - Package manager for Chrome OS https://chromebrew.github.io
 
   Usage:
-    crew build [options] [-k|--keep] [-v|--verbose] <name> ...
-    crew check [-V|--version] [-v|--verbose] <name> ...
+    crew build [options] [-f|--force] [-k|--keep] [-v|--verbose] <name> ...
+    crew check [-f|--force] <name> ...
```
uberhacker, line 350: Keep the verbose option?

Zopolis4: It never did anything, so removing it just makes things less misleading.

README.md
```diff
@@ -94,7 +94,6 @@ Where available commands are:
 | remove       | remove package(s) |
 | search       | look for package(s) |
 | sysinfo      | show system information in markdown style |
-| test         | test crew command(s) |
```
uberhacker, line 97: What if we want to test prior to submitting a PR?  If this is removed, we would have to submit a PR and see if the tests fail first?  If that is true, not ideal.

tests/cycle_test
```diff
@@ -1,68 +0,0 @@
...
-end
-
-@cycles = 'cycles'
-@cycles = 'cycle' if @failed == 1
-
-if @failed.positive?
-  abort "\n#{@failed} dependency #{@cycles} found.".lightred
-else
-  puts "\nNo dependency cycles found.".lightgreen
-end
```
uberhacker, line 68: Does this no longer have any value?  Someone put a lot of effort into creating this test.

README.md
Zopolis4: If we want to test prior to submitting a PR, we now run `rake test`.

tests/cycle_test
Zopolis4: Unfortunately, it does not. As per the PR description, we have a number of packages with dependency cycles, including unavoidable ones such as cairo, harfbuzz, freetype and the rest. 

If this test still worked, it would fail on behavior that we cannot change.

Zopolis4: It will live on in the git history, though :sweat_smile:

README.md
Zopolis4: This also allows easier testing on non-ChromeOS environments, as you do not need to have crew installed to test every command at once.

Zopolis4:
CI is failing on i686 because the system zlib is not recent enough to act as a proper fallback for ruby, so we remove it successfully but are unable to re-install it because ruby, and thus crew, is now broken.

Zopolis4:
Fixed by removing zstd instead, which doesn't appear to trigger the same failures.

lib/const.rb
uberhacker: Actually, we need to add `-V|--version`.
```
$ crew help check
Check package(s) for syntax errors and upstream updates.
Usage: crew check [-V|--version] [-v|--verbose] <package1> [<package2> ...]
If `-V` or `--version` is present, it will search for an upstream update.
If `-v` or `--verbose` is present, up to date packages will be displayed.
```
Wouldn't hurt to add `-f|--force` to the help command too.

Zopolis4: I've dropped the `-V|--version` functionality because it is out of place with the rest of the command, but I'll add `-f|--force`.

Zopolis4: Actually, `-f|--force is already in the help` (see line 21 of the modifications to commands/help.rb in this PR)

satmandu (approved):

PR #9736, opened 2024-04-29
Update git_prompt, have install use new faster git reftable support

satmandu:
- Adds the faster git reftable support only during installs as per https://github.blog/2024-04-29-highlights-from-git-2-45/
- install change tested on `i686`.

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=git-install crew update
```

### DIFF
--- a/packages/git_prompt.rb
+++ b/packages/git_prompt.rb
@@ -4,18 +4,17 @@ require_relative 'git'
 class Git_prompt < Package
   description 'Display the git branch and status in the command prompt'
   homepage 'https://github.com/git/git'
-  version '2.44.0'
+  version Git.version
   # When upgrading git_prompt, be sure to upgrade git in tandem.
-  puts "#{self} version differs from Git version #{Git.version}".orange if version != Git.version
+  # puts "#{self} version differs from Git version #{Git.version}".orange if version != Git.version
   license 'GPL-2'
   compatibility 'all'
-  source_url 'SKIP'
+  source_url Git.source_url
+  source_sha256 Git.source_sha256
 
   print_source_bashrc
 
   def self.build
-    downloader "https://raw.githubusercontent.com/git/git/v#{version}/contrib/completion/git-prompt.sh",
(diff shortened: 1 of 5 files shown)


uberhacker (commented):
Good call!  No x86_64 filelist?

satmandu:
> Good call! No x86_64 filelist?

Already there!

packages/git_prompt.rb
```diff
@@ -4,18 +4,25 @@
...
+  source_url Git.source_url
+  source_sha256 Git.source_sha256
+  binary_compression 'tar.zst'
+
+  binary_sha256({
+    aarch64: '1541e237594840c458b5a234278cb2150dc1f01d5e7d0c7932a8397e598a6837',
+     armv7l: '1541e237594840c458b5a234278cb2150dc1f01d5e7d0c7932a8397e598a6837',
+       i686: 'caa10c0cfa40b0c20454151433599482eee9a431a06856eeea1d4c0d455caa33',
+     x86_64: 'dab1b1d174feb393035fb6e3ba38cb53282ec827916580e46f43b2eaaff83554'
+  })
```
uberhacker, line 19: Is this necessary?  We don't need to compile since we're just copying files, right?  That said, we'll also need:
```
no_compile_needed
```

satmandu: It's faster to build in advance so you don't have to download the entire git source tree?

This just simplifies things to not have to have version changes automatically picked up, but yes, we could remove the binaries and change it to no_compile_needed...

satmandu: done.

uberhacker: Does it take longer than 1 minute to install without binaries?  Actually, we could probably combine this with the git package eventually but it is nice to make it more modular.

satmandu: It just takes time to download the git package to pull out one file... No biggie, I think.

uberhacker (approved):

PR #9756, opened 2024-05-06
update libxml2, libunwind

satmandu:
Builds properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=libxml2 crew update
```


uberhacker (approved):

packages/libxml2.rb
```diff
@@ -29,23 +29,12 @@ class Libxml2 < Autotools
...
-      --with-threads \
-      --with-history'
-
-  def self.check
     # Remove EBCDIC test since it fails.
     # Check https://mail.gnome.org/archives/xml/2010-April/msg00010.html for details.
-    system 'rm', 'test/ebcdic_566012.xml'
-
-    system 'make', 'check'
+    FileUtils.rm 'test/ebcdic_566012.xml'
```
Zopolis4, line 34: If we aren't even testing anymore, is there a need to remove a failing test?

satmandu: I used `run_tests` but building with tests led to the install including the tests too, which we don't want.

uberhacker: Running some tests is better than running no tests at all because one test is failing the whole suite.  Totally in agreement with this decision.

Zopolis4: Yes, but since `run_tests` has been removed, we aren't running any tests at all. Hence, why would we modify the tests if we don't even run them?

uberhacker: AFAIK, `run_tests` is a valid boolean property and still  works with `buildsystems/autotools` builds.

Zopolis4: I mean it has been removed from the libxml2 package file.

uberhacker: Oh, was it ever there?

Zopolis4: There used to be an equivalent `self.check` section, which this PR removed while still leaving the test modification line behind.

PR #9772, opened 2024-05-08
Remove wkhtmltox

### LINKED ISSUE #3620: Upgrade wkhtmltox to 0.12.5-1
It seems it will become REALLY hard to upgrade wkhtmltox since developers totally changed their approach between 0.12.4 and 0.12.5 (or at least what they provide). i386-linux and amd64-linux are no more available and you need to install an installation platform to build wkhtmltopdf.  This platform runs a python script that uses the wkhtmltopdf repository as an argument. Dependencies to this installation platform are heavy and require docker..... Bad news... This package won't be able to be upgraded as previously done unless someone has a genious idea. It is required by pandoc to produce PDFs using html preconversion instead of latex.

----------------------------------

Packaging wkhtmltopdf releases
==============================

Packaging wkhtmltopdf is a challenge because of the need for using a patched
Qt to provide additional functionality and the cross-platform targets.
Especially for Linux, the approach for packaging has changed multiple times,
so it is best to decouple it from the releases itself.

This will allow creation of packages as per latest best practices, using the
latest dependent libraries in static builds and for targets to be added long
after the release has been made.

All targets are built in a separate container or VM to ensure that nothing
from the build machine leaks into the output package and it can be reproduced
by anyone.

Requirements
============

The software requirements on the build machine are:

* `git` and `p7zip`
* `python` and `PyYAML`
* `docker` >= 17.05 (for linux targets)
  * build-time args in FROM were introduced in this version
* Linux kernel >= 4.8 (for linux targets -- foreign archs only)
  * for seamless foreign arch emulation via qemu-user-static
* `vagrant` with `virtualbox` (for non-linux targets)

On Ubuntu 18.04, this can be installed via a single command:

    sudo apt install -y python-yaml docker.io vagrant virtualbox p7zip-full

Zopolis4:
Fixes #3620

Given that `wkhtmltox` is deprecated [upstream](https://github.com/wkhtmltopdf/wkhtmltopdf), and we currently ship a version vulnerable to [CVE-2020-21365](https://nvd.nist.gov/vuln/detail/CVE-2020-21365), and the latest upstream release is vulnerable to [CVE-2022-35583](https://nvd.nist.gov/vuln/detail/CVE-2022-35583), and [the majority of distros](https://repology.org/project/wkhtmltopdf/versions) haven't patched it either, let's just cut our losses.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=listve crew update
```


satmandu:
Wget2 rename is part of this too?

Zopolis4:
> Wget2 rename is part of this too?

Adding a comma at the end of the line.

uberhacker:
Isn't this a bit extreme?  Not sure I'm in agreement with this decision.  Think about what functionality this package provides.  It simply converts HTML to PDF.  Of course, a rouge HTML document will generate an insecure PDF.  But if you have control over your content, this is not a big deal at all.

Zopolis4:
I don't feel comfortable knowingly shipping a security vulnerability.
image
image

Zopolis4:
> But if you have control over your content, this is not a big deal at all.

I don't expect people to be manually inspecting the html of every website they convert to PDFs-- do you?

uberhacker:
> I don't feel comfortable knowingly shipping a security vulnerability. image image

Well, doesn't every other package manager keep it in their repo?  Don't overreact.  The security issue is only a problem if you don't control your content.

uberhacker:
> > But if you have control over your content, this is not a big deal at all.
> 
> I don't expect people to be manually inspecting the html of every website they convert to PDFs-- do you?

Yes, I do actually.  That is part of doing your due diligence.  The vulnerability is only a problem if you allow iframes in your html.  Just don't include or accept iframe content and you should be fine.

Zopolis4:
> Don't overreact. 
We are not shipping a package that exposes our users to a 7.5 CVE. I am shocked that this is even something that needs to be debated.

> The security issue is only a problem if you don't control your content.

I really don't get what you are trying to say here. Websites are inherently untrusted content. Users do not know whether a website will be a security risk or not.

uberhacker:
> > Don't overreact.
> > We are not shipping a package that exposes our users to a 7.5 CVE. I am shocked that this is even something that needs to be debated.
> 
> > The security issue is only a problem if you don't control your content.
> 
> I really don't get what you are trying to say here. Websites are inherently untrusted content. Users do not know whether a website will be a security risk or not.

Then they deserve to be hacked if they accept untrusted content.  Simple solution is to not accept it.

Zopolis4:
> Then they deserve to be hacked if they accept untrusted content. 

I will not expose users to security risks simply because you think they "deserve to be hacked".

uberhacker:
> > Then they deserve to be hacked if they accept untrusted content.
> 
> I will not expose users to security risks simply because you think they "deserve to be hacked".

You are not understanding what I'm saying.  Can we add a postinstall message with a warning about the security issue instead of removing the whole package please?

Zopolis4:
I think we might have lost the plot a little bit here-- we are currently shipping a vulnerable version of a package. Upgrading is not a solution, because that introduces an even more severe vulnerability. 

Since the package is deprecated upstream and all versions are vulnerable, for security reasons the only option is to remove the package.

satmandu (approved):

satmandu:
Looks like this package works?

Just tested it.

Please add the missing libxrender and libxext deps and a warning about being careful about what HTML pages are converted?

ZentarXD:
Ja nie wiem jak to mam zrobić,proszę to zrozumieć

czw., 9 maj 2024, 19:30 użytkownik Satadru Pramanik, DO, MPH, MEng <
***@***.***> napisał:

PR #9773, opened 2024-05-10
Re-add wkhtmltox

satmandu:
Reverts chromebrew/chromebrew#9772
- Adds dependencies
- Updates compatibility to remove i686
- Adds postinstall warning message

### DIFF
--- a/lib/fixup.rb
+++ b/lib/fixup.rb
@@ -61,8 +61,7 @@ pkg_update_arr = [
   { pkg_name: 'qtwebsockets', pkg_rename: 'qt5_websockets', pkg_deprecated: nil, comments: nil },
   { pkg_name: 'qtx11extras', pkg_rename: 'qt5_x11extras', pkg_deprecated: nil, comments: nil },
   { pkg_name: 'ruby_debug', pkg_rename: nil, pkg_deprecated: true, comments: 'Integrated into ruby package.' },
-  { pkg_name: 'wget', pkg_rename: 'wget2', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
-  { pkg_name: 'wkhtmltox', pkg_rename: nil, pkg_deprecated: true, comments: 'Deprecated upstream with unpatched CVE-2022-35583.' }
+  { pkg_name: 'wget', pkg_rename: 'wget2', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' }
 ]
 
 pkg_update_arr.each do |pkg|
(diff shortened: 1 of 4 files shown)


Zopolis4 (changes requested):
I still don't understand the rationale for doing this-- shipping vulnerable packages is not a good idea.

Zopolis4:
i.e. we rebuilt `xz` to not have the security vulnerability in #9568. That's not an option here, so let's just not ship the package. Am I missing something?

uberhacker (approved):

uberhacker:
> i.e. we rebuilt `xz` to not have the security vulnerability in #9568. That's not an option here, so let's just not ship the package. Am I missing something?

Yes, you're missing that having a vulnerability for iframes is not a good enough reason to remove a package.  The security implications are not that severe.

satmandu:
xz is HEAVILY used. we have to keep that patched. It's comparing apples and oranges, I think.

Zopolis4:
the security implication is a 9.8 on the CVE scale. The security implications are nearly as severe as possible.

from the cve page:
```
This allows the attacker to takeover the whole infrastructure by accessing their internal assets.
```

Zopolis4:
> xz is HEAVILY used. we have to keep that patched. It's comparing apples and oranges, I think.

is the implication that because wkhtmltox is not heavily used we are ok to ship an exploitable version?
I'm struggling to understand the reasoning here.

uberhacker:
> the security implication is a 9.8 on the CVE scale. The security implications are nearly as severe as possible.
> 
> from the cve page:
> 
> ```
> This allows the attacker to takeover the whole infrastructure by accessing their internal assets.
> ```

Only if you allow iframes.  Most responsible website owners will block iframe content if it is coming from users, otherwise, they can control the content of their own html.

uberhacker:
> this isnt about website owners, wkhtmltox allows users to convert websites into a pdf.
> 
> the majority of users will simply use the tool to convert whatever website they decide to, and will not be manually inspecting the html for an iframe or the other exploit. Again, this is a 9.8 CVE, there is no world in which we should be shipping this package.
> 
> not shipping a 9.8 CVE is common sense

How do you know what the majority of users will do?  This is the way script kiddies think, not responsible content owners.  Again, the vulnerability is with iframes.  I feel like I'm not getting my point across.

Zopolis4:
There are two things here.

One, we disagree on how likely users are to be exploited by this vulnerability. 

Thats fine. We don't need to agree on that.

Second, we disagree on what we should do about security vulnerabilities.

My belief is that we should fix security vulnerabilities.
Espescially ones that are rated as a 9.8 (Critical) out of a possible 10.
Whether you personally think this vulnerability will only afffect people "who deserve to be hacked" should not affect this.
This is common-sense security policy. You don't knowingly ship exploitable software. 
As a package manager, one of our responsiblities is to ensure we don't expose people to vulnerable software.

Zopolis4:
Perhaps it would be a good idea to agree on a chromebrew security policy, to prevent situations like this occuring.

PR #9784, opened 2024-05-13
Rework lib/fixup.rb to not use an array for renaming and deleting packages

### LINKED ISSUE #9783: `lib/fixup.rb` does not handle removing deprecated packages properly.
**Describe the bug**
When removing a deprecated package, `lib/fixup.rb` removes the files, but does not remove them from the list of installed packages.
This is due to an error in the code for writing to `device.json`

**To Reproduce**

`satmandu/crewbuild:amd64` docker image (`Digest: sha256:c650cf0911bd7b9f4cca4d3289386dbb1d9f68e2765e37958459338096d152a3`

      ```
      crew install js91
      crew update # Accept the removal of js91 here
      crew list installed | grep "js91"
      ```

      Output:

      ```
      chronos@67adfbd614ad /usr/local/lib/crew/packages $ crew update
      From https://github.com/chromebrew/chromebrew
       * branch                master     -> FETCH_HEAD
      HEAD is now at e62d1ce08 Vscodium 1.87.2.24072 => 1.89.1.24130 (#9780)
      
      Js91 found in package fixup list
      Js91 is deprecated and should be removed.
      Js91: Replaced by newer versions of js.
      
      Would you like to remove deprecated package Js91? [y/N] y
      js91 removed!
      Your software is up to date.
      
      js91 # from the checking of crew list installed | grep "js91"
      ```
**Expected behavior**
That `js91` is removed from the list of installed packages.

- User space: `64-bit`
- Chromebrew Kernel version: `5.10`
- Chromebrew Running in Container: `true`

- Chromebrew version: `1.48.1`

- Last update in local repository: e62d1ce08 `Vscodium 1.87.2.24072 => 1.89.1.24130 (#9780) (33 minutes ago)`

- OS variant: `Chrome OS`
- OS version: `nocturne-release/R124-15823.40.0`

</details>

Zopolis4:
Fixes #9783.

`lib/fixup.rb` could use more work, but this is a start.

Also, `js91` is deprecated but the package file is still in-tree?

```
# Reset to an older version of crew and install some renamed and deprecated packages first
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=grout crew update
```

### DIFF
--- a/commands/remove.rb
+++ b/commands/remove.rb
@@ -58,7 +58,7 @@ class Command
     device_json['installed_packages'].delete_if { |entry| entry['name'] == pkg.name }
 
     # Update device.json with our changes.
-    save_json(device_json)
+    File.write File.join(CREW_CONFIG_PATH, 'device.json'), JSON.pretty_generate(device_json)
 
     # Perform any operations required after package removal.
     pkg.remove
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -2,7 +2,7 @@
 # Defines common constants used in different parts of crew
 require 'etc'
 
-CREW_VERSION = '1.49.1'
+CREW_VERSION = '1.49.2'
 
 # kernel architecture
 KERN_ARCH = Etc.uname[:machine]
--- a/lib/fixup.rb
+++ b/lib/fixup.rb
@@ -23,142 +23,139 @@ system(" sed -i 's/binary_sha256/sha256/g' #{File.join(CREW_CONFIG_PATH, 'device
 
 # Check for renamed and deprecated packages, and handle them.
 
-pkg_update_arr = [
-  { pkg_name: 'acli', pkg_rename: 'acquia_cli', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
-  { pkg_name: 'agrind', pkg_rename: 'angle_grinder', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
-  { pkg_name: 'apriconv', pkg_rename: 'apr_iconv', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
-  { pkg_name: 'aprutil', pkg_rename: 'apr_util', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
-  { pkg_name: 'arpscan', pkg_rename: 'arp_scan', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
-  { pkg_name: 'aws', pkg_rename: 'aws_cli', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
-  { pkg_name: 'btrfsprogs', pkg_rename: 'btrfs_progs', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
-  { pkg_name: 'bz2', pkg_rename: 'bzip2', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
-  { pkg_name: 'bz3', pkg_rename: 'bzip3', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
-  { pkg_name: 'codium', pkg_rename: 'vscodium', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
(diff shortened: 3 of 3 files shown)


commands/remove.rb
```diff
@@ -54,7 +54,7 @@ def self.remove(pkg, verbose)
     device_json['installed_packages'].delete_if { |entry| entry['name'] == pkg.name }
 
     # Update device.json with our changes.
-    save_json(device_json)
+    File.write File.join(CREW_CONFIG_PATH, 'device.json'), JSON.pretty_generate(device_json.to_json)
```
satmandu, line 57: Why change this?

Zopolis4: The previous code was failing in certain cases-- see attatched issue.

satmandu:
I don't understand why using an array is bad? Isn't that simpler and cleaner?

Zopolis4:
It means we have to have an unnecessary  `pkg_rename: nil` or `pkg_deprecated: nil` for each package, and also we need to pass an unnecessary `comments: nil` in most cases.

Zopolis4:
Compare:
```ruby
  { pkg_name: 'qtwebglplugin', pkg_rename: 'qt5_webglplugin', pkg_deprecated: nil, comments: nil },
```

```ruby
pkg_rename('qtwebglplugin', 'qt5_webglplugin')
```

satmandu:
Unit tests are failing.

commands/remove.rb
satmandu: So the `save_json` function isn't working and that's the root cause?

Zopolis4:
Fixed.

commands/remove.rb
Zopolis4: The `save_json` function isn't working in this specific case-- not sure why it was working when we weren't running things from `lib/fixup.rb`. If I had to guess, its probably something to do with the fact that `save_json` isn't actually available in `commands/remove.rb` and is only there due to the leakage of us calling it from `bin/crew`, and that doesn't happen when we call it from `lib/fixup.rb` inside `bin/crew` despite not `require`'ing the function.

This is why I'm trying to split everything out into files, and then use files in `lib` for shared function implementations. Otherwise debugging these sort of issues gets pretty complicated.

satmandu:
Honestly I'd be happier with the array being put in another file and just read in, as opposed to putting a bunch of `deprecrated` or whatever calls in a script file.

Is it not preferred to split data and code?

_Also, please discuss refactoring in slack before spending a bunch of time implementing. This is a group project, and you need sign-off on how something is being architected from stakeholders. You can't just assume that because you put in the time, that people are going to be cool with it._

Zopolis4:
> Honestly I'd be happier with the array being put in another file and just read in, as opposed to putting a bunch of `deprecrated` or whatever calls in a script file.

I don't see how moving the array into a different file changes anything?

> Is it not preferred to split data and code?

My thinking was that this:
```ruby
pkg_rename('qtwebglplugin', 'qt5_webglplugin')
```
was preferable to this:
```ruby
  { pkg_name: 'qtwebglplugin', pkg_rename: 'qt5_webglplugin', pkg_deprecated: nil, comments: nil },
```

> This is a group project, and you need sign-off on how something is being architected from stakeholders. You can't just assume that because you put in the time, that people are going to be cool with it.

Not really sure what you're getting at-- this is a pull request, of course I am asking for sign-off here. This is me asking for feedback and comments on my changes.

commands/remove.rb
Zopolis4: I don't think there's a problem with `save_json`-- the problem is that we're using `save_json` at all. 

As I said:
> If I had to guess, its probably something to do with the fact that save_json isn't actually available in commands/remove.rb and is only there due to the leakage of us calling it from bin/crew, and that doesn't happen when we call it from lib/fixup.rb inside bin/crew despite not require'ing the function.

Zopolis4:
This fixes a pretty important outstanding issue-- @satmandu, if you're willing to accept this solution then let's merge it, otherwise let me know and I'll go back to the drawing board-- no sense in me keeping this updated if you don't want it to be merged.

PR #9808, opened 2024-05-15
Update xorg_proto, xwayland

satmandu:
Builds properly:
- [x] `x86_64`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=x crew update ; yes | crew upgrade
```


Zopolis4 (approved):
For what its worth-- running the supplied testing command does nothing because neither of these packages are installed.

satmandu:
> For what its worth-- running the supplied testing command does nothing because neither of these packages are installed. 

I assumed that you would be running it on an actual install, not in a container to test, as xwayland is installed if you are using sommelier.

uberhacker (approved):

PR #9818, opened 2024-05-17
Add ytnef package and switch gspell to autotools

Zopolis4:
I'm working on adding geary` but it fails with a bunch of linking errors in webkit2gtk.

However, I thought it best to PR these changes now, so that no work is duplicated.

ytnef is a geary dependency, and the switch to the autotools buildsystem in gspell is to add the required vala files. (The CMake build file for gspell doesn't add them)

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=leaf crew update
```


uberhacker (changes requested):
Add ytnef to tools/packages.yaml?

Zopolis4:
Done. That being said, what do we use `tools/packages.yml` for anymore?

uberhacker:
> Done. That being said, what do we use `tools/packages.yaml` for anymore?

The original intent was for use with [urlwatch](https://urlwatch.readthedocs.io) but I was never able to get it to work.

uberhacker (approved):

Zopolis4:
`version.rb` should have that covered now, right? 

It might make sense to keep `check.rb` around until I can verify that `version.rb` can handle all the packages `check.rb` can, so I can use the regexes to add the package to Anitya.

But `packages.yaml` and `check_updates.sh` aren't needed anymore, right?

PR #9819, opened 2024-05-17
Minor source_url cleanups

Zopolis4:
Just going through some url's and fixing anything I see that pops up.

Apart from `libdbusmenu`, every other git `source_url` ends in `.git`, which is correct most of the time and harmless other times, but `libdbusmenu` cannot end in `.git`, as this triggers a login request.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=tenor crew update
```

### DIFF
--- a/packages/musl_xz.rb
+++ b/packages/musl_xz.rb
@@ -6,7 +6,7 @@ class Musl_xz < Package
   version '5.2.5-e7da-1'
   license 'public-domain, LGPL-2.1+ and GPL-2+'
   compatibility 'all'
-  source_url 'https://github.com/xz-mirror/xz.git'
+  source_url 'https://github.com/tukaani-project/xz.git'
   git_hashtag 'e7da44d5151e21f153925781ad29334ae0786101'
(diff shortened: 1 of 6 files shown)


packages/musl_xz.rb
```diff
@@ -6,7 +6,7 @@ class Musl_xz < Package
   version '5.2.5-e7da-1'
   license 'public-domain, LGPL-2.1+ and GPL-2+'
   compatibility 'all'
-  source_url 'https://github.com/xz-mirror/xz.git'
+  source_url 'https://github.com/tukaani-project/xz.git'
```
uberhacker, line 9: It might be better to keep the previous source url.  A mirror site will localize the download.

Zopolis4: The previous url is archived and tells you to go to the other url, so I switched.
image

uberhacker: Fair enough.

uberhacker (approved):

PR #9830, opened 2024-05-21
Handbrake 1.7.0 => 1.8.0

uberhacker:
Tested and working in x86_64 container.

### DIFF
--- a/packages/handbrake.rb
+++ b/packages/handbrake.rb
@@ -3,7 +3,7 @@ require 'package'
 class Handbrake < Package
   description 'HandBrake is a tool for converting video from nearly any format to a selection of modern, widely supported codecs.'
   homepage 'https://handbrake.fr/'
-  version '1.7.0'
+  version '1.8.0'
   license 'GPL-2'
   compatibility 'x86_64'
   source_url 'https://github.com/HandBrake/HandBrake.git'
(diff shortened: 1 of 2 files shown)


packages/handbrake.rb
```diff
@@ -94,20 +94,15 @@ def self.install
     Dir.chdir 'x86_64-cros-linux-gnu' do
       system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
     end
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
     FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/etc/env.d"
+    FileUtils.ln_s "#{CREW_PREFIX}/bin/HandBrakeCLI", "#{CREW_DEST_PREFIX}/bin/hb"
```
Zopolis4, line 99: Doesn't this draw from the current install instead of the files in `CREW_DEST_PREFIX`?

Zopolis4 (changes requested):

packages/handbrake.rb
Zopolis4: Also, why are we creating this symlink?

uberhacker: The first argument is the source; the second argument is the target.  The target is what we want to create so that when it installs, it has the correct path.

Zopolis4 (approved):

PR #9832, opened 2024-05-22
Split update command into separate file

Zopolis4:
This just moves the code out of `bin/crew` and into `commands/update.rb`, and only makes the most basic changes to not use unnecessary instance variables and to not rely on `bin/crew` instance variables.

~~Behaviour changes:~~
~~- Drop unused `generate_compatible` function.~~
~~- Let docopt handle incorrect arguments.~~
~~- Fix the docopt string for `crew update`.~~
~~- Only mark the truly essential packages in `CREW_ESSENTIAL_PACKAGES`.~~
~~- We do not need `xzutils` for `.xz` support in `zstd` to run ruby, for example.~~

~~Refactoring changes:~~
~~- Use `File.join` more often.~~
~~- Use `PackageUtils.installed?` to check if we have `git-restore-mtime`.~~

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula11 crew update
```

### DIFF
--- a/commands/update.rb
+++ b/commands/update.rb
@@ -1,0 +1,100 @@
+require 'json'
+require_relative '../lib/color'
+require_relative '../lib/const'
+require_relative '../lib/package'
+require_relative '../lib/package_utils'
+
+class Command
+  def self.update(last_update_check)
+    unless CREW_NO_GIT
+      unless Dir.exist?(File.join(CREW_LIB_PATH, '.git'))
+        puts 'Fixing Chromebrew system git repo clone...'.orange
+        system(<<~GIT_REPAIR_COMMANDS, chdir: CREW_LIB_PATH, %i[out err] => File::NULL)
+          ## Run the git setup commands used in install.sh.
+          # Make the git default branch error messages go away.
+          git config --global init.defaultBranch main
+          # Setup the dir with git information.
+          git init --ref-format=reftable
+          git remote add origin #{CREW_REPO}
+          # Help handle situations where GitHub is down.
+          git config --local http.lowSpeedLimit 1000
+          git config --local http.lowSpeedTime 5
+          # Checkout, overwriting local files.
+          git fetch --all
+          git checkout -f master
+          git reset --hard origin/#{CREW_BRANCH}
+        GIT_REPAIR_COMMANDS
+      end
+
+      ## Update crew from git.
+      # Set sparse-checkout folders.
+      silent = CREW_UNATTENDED ? '&>/dev/null' : ''
+      system "git sparse-checkout set packages manifest/#{ARCH} lib commands bin crew tests tools #{silent}", chdir: CREW_LIB_PATH, exception: true
+      system "git sparse-checkout reapply  #{silent}", chdir: CREW_LIB_PATH, exception: true
+      system "git fetch #{CREW_REPO} #{CREW_BRANCH}  #{silent}", chdir: CREW_LIB_PATH, exception: true
+      # Now that we've fetched all the new changes, see if lib/const.rb was changed.
+      # We do this before resetting to FETCH_HEAD because we lose the original HEAD when doing so.
+      to_update = `cd #{CREW_LIB_PATH} && git show --name-only HEAD..FETCH_HEAD`.include?('lib/const.rb')
+      system("git reset --hard FETCH_HEAD  #{silent}", chdir: CREW_LIB_PATH, exception: true)
+
+      if Time.now.to_i - last_update_check > (CREW_UPDATE_CHECK_INTERVAL * 3600 * 24)
+        puts 'Updating RubyGems.'.orange
+        system "gem update -N --system  #{silent}"
+        system "gem cleanup  #{silent}"
+      end
+
+      puts 'Package lists, crew, and library updated.' unless CREW_UNATTENDED
+
+      # If lib/const.rb was changed, CREW_VERSION was bumped, so we re-run crew update.
+      if to_update
+        puts 'Restarting crew update since there is an updated crew version.'.lightcyan unless CREW_UNATTENDED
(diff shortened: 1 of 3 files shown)


commands/update.rb
```diff
@@ -0,0 +1,59 @@
...
+  def self.update
+    # Update package lists.
+    Dir.chdir(CREW_LIB_PATH) do
+      # Set sparse-checkout folders.
+      system "git sparse-checkout set packages manifest/#{ARCH} lib commands bin crew tests tools"
+      system 'git sparse-checkout reapply'
+
+      system "git fetch #{CREW_REPO} #{CREW_BRANCH}", exception: true
+      system 'git reset --hard FETCH_HEAD', exception: true
+      system 'git-restore-mtime -sq 2>/dev/null', exception: true if PackageUtils.installed?('git_mestrelion_tools')
```
uberhacker, line 17: Why is this necessary?

uberhacker: I guess this was carried over so...

Zopolis4: I'm not exactly sure what question you're asking.

commands/help.rb
```diff
@@ -131,8 +131,6 @@ def self.help(command)
         Update crew.
         Usage: crew update
         This only updates crew itself.  Use `crew upgrade` to update packages.
-        Usage: crew update compatible
-        This updates the crew package compatibility list.
```
uberhacker, line 135: I'm not sure of the ramifications of removing this.

Zopolis4: None. The last remaining usage was in `crew whatprovides`, and I've already removed that. We just use `PackageUtils.compatible?` instead.

uberhacker: Should we purge it from device.json too?

Zopolis4: We could, its not doing any harm by being there and it won't show up on new installs. I don't really mind either way, so I didn't implement it.

satmandu (dismissed):
`crew update compatible` is used to refresh package information without a git pull.

It looks like this PR breaks that functionality...

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula11 crew update compatible
Could not understand "crew update compatible".
Usage:
  crew build [options] [-k|--keep] [-v|--verbose] <name> ...
  crew check [-V|--version] [-v|--verbose] <name> ...
  crew const [-v|--verbose] [<name> ...]
  crew deps [options] [--deep] [-t|--tree] [-b|--include-build-deps] [--exclude-buildessential] [-v|--verbose] <name> ...
  crew download [options] [-s|--source] [-v|--verbose] <name> ...
  crew files <name> ...
  crew help [<command>] [-v|--verbose] [<subcommand>]
  crew install [options] [-k|--keep] [-s|--source] [-S|--recursive-build] [-v|--verbose] <name> ...
  crew list [options] [-v|--verbose] (available|installed|compatible|incompatible)
  crew postinstall [options] [-v|--verbose] <name> ...
  crew prop [<property>]
  crew remove [-v|--verbose] <name> ...
  crew search [-v|--verbose] <name> ...
  crew sysinfo [-v|--verbose]
  crew test [-v|--verbose] [<name> ...]
  crew update
  crew upgrade [options] [-k|--keep] [-s|--source] [-v|--verbose] [<name> ...]
  crew upload [options] [-v|--verbose] [<name> ...]
  crew whatprovides <pattern> ...
chronos@cheekon-x86_64 /usr/local/lib/crew/packages (master >|SPARSE)$
```

Zopolis4:
All crew update compatible does is run PackageUtils.compatible? on every package and save the result. No part of crew uses the result anymore, so we can safely remove it.

satmandu:
> All crew update compatible does is run PackageUtils.compatible? on every package and save the result. No part of crew uses the result anymore, so we can safely remove it.

So if I change a package file in `/usr/local/lib/crew/packages` and run `crew upgrade` it will automatically be picked up now?

Zopolis4:
crew update compatible never did that, unless I am missing something. But yes, that should work.

satmandu:
I guess you will also have to add in the essential packages code that creates `@device[:essential_deps]` I just dropped into `generate_compatible`...

🫣

Zopolis4:
It's kinda hard to keep all these branches rebased and up to date when theyre not getting reviewed... :sweat_smile: 

maybe a review in less than 3 months this time? :pray:

Zopolis4:
@satmandu thoughts on the implementation? This avoids the need for most of the problems in #10363, so if we can get this merged you'll have a much simpler job to do.

lib/const.rb
```diff
@@ -40,7 +40,10 @@
   HOME = File.join(CREW_PREFIX, Dir.home)
 end
 
-CREW_ESSENTIAL_PACKAGES = %w[gcc_lib glibc gmp lz4 ruby xzutils zlib zstd]
+# These are packages that crew needs to run-- only packages that the bin/crew needs should be required here.
+# lz4, for example, is required for zstd to have lz4 support, but this is not required to run bin/crew.
+# The LIBC_VERSION ternary is to reflect the change from glibc_build to glibc_lib as the source of essential libraries starting at glibc 2.35.
+CREW_ESSENTIAL_PACKAGES = %W[gcc_lib #{LIBC_VERSION.to_f > 2.34 ? "glibc_lib#{LIBC_VERSION.delete('.')}" : "glibc_build#{LIBC_VERSION.delete('.')}"} gmp ruby zlib zstd]
```
uberhacker, line 46: Use `Gem::Version.new(LIBC_VERSION) > Gem::Version.new('2.34')`?

Zopolis4: We don't need to use it here, so better to avoid the unnecessary complexity.

Zopolis4:
The codebase has changed somewhat since I last rebased this-- I made most of the changes that I would have liked to have made here in instead in other PRs over the course of a couple years. New things I would like to refactor have also shown up since then, but those will be in a later PR as per previous discussions.

satmandu (approved):

PR #9839, opened 2024-05-23
Mesa => 24.1.0

satmandu:
Builds properly:
- [x] `x86_64`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=master crew update
```


Zopolis4:
Given that `libclc` uses the llvm sources, would it make more sense to integrate it into the llvm packages?

packages/libclc.rb
```diff
@@ -0,0 +1,78 @@
+# Adapted from Arch Linux libclc PKGBUILD at:
+# https://github.com/archlinux/svntogit-packages/raw/packages/libclc/trunk/PKGBUILD
+
+require 'package'
+require_relative 'llvm18_build'
+
+class Libclc < Package
```
uberhacker, line 7: Not possible to use CMake buildsystem?

packages/spirv_llvm_translator.rb
```diff
@@ -14,9 +14,9 @@ class Spirv_llvm_translator < CMake
   binary_compression 'tar.zst'
 
   binary_sha256({
-    aarch64: '09ffe51c28f2839522c74bc1180d7a8b6862f3e360ae1736ddb4f8601311f5e5',
-     armv7l: '09ffe51c28f2839522c74bc1180d7a8b6862f3e360ae1736ddb4f8601311f5e5',
-     x86_64: '89bfa4e3cbb0fc50aba0024932ce6345b08e57b5dc90fd9b85d9e0c3dc172648'
+    aarch64: 'f98669e96f7168605997bc17bf3403931a585a38d5159d9ece7c01f9938e8bcb',
+     armv7l: 'f98669e96f7168605997bc17bf3403931a585a38d5159d9ece7c01f9938e8bcb',
+     x86_64: '85be281a6bd28b33f6987112137e32da1e846fef48103f5c2df0496c90d63a5f'
```
uberhacker, line 19: No version bump?

Zopolis4: Looking at the commit message, its just a rebuild.

uberhacker: Yes, but rebuilds should also have a version bump because this overwrites the previous (likely working) build.

Zopolis4: Oh, it overwrites the uploaded build. (Not exactly overwrites, its still [here](https://gitlab.com/chromebrew/binaries/-/packages/23451211), but the url now downloads the new binary.)

We should probably find out a way to fix that so we aren't continuously bumping revisions.

Maybe not uploading until right before the PR is merged? Perhaps there is a smarter solution on the gitlab end.

satmandu:
> Given that `libclc` uses the llvm sources, would it make more sense to integrate it into the llvm packages?

No? Similar situation to openmp.

packages/spirv_llvm_translator.rb
satmandu: No version bump because rebuild.

It would nice nice for `crew` to be able to iterate through all the downloads in `https://gitlab.com/chromebrew/binaries/-/packages/234512111` until the hash matches, so we don't have to create a separate version bump.

uberhacker (approved):

PR #9856, opened 2024-05-27
Various package renames and fixes for version checking

Zopolis4:
Working through getting feature parity between `version.rb` and `check.rb`.

First step, ensuring every listed package in `check.rb` works in `version.rb`. (Technically most of those packages don't work even in `check.rb`, but still)

I've left out the `zlibpkg` -> `zlib` rename because it breaks minizip, I'll look at that separately.

This leaves a few things which are listed in `check.rb` but don't work in `version.rb`:
- bcif (doesn't work in `check.rb` anyways)
- gdal (upstream anitya issue that I can't do anything to fix, have admin flagged it)
- go_bootstrap (we just need to fix the version in the package, probably needs a rebuild)
- gperf (upstream gperf issue because they havent actually released 3.2, issue report already present upstream)
- jdk8 (doesn't work in `check.rb` but does have a regex, although this package is gonna be replaced by openjdk8)
- nethack4 (also doesn't work in `check.rb` anyways)
- zlibpkg (working on the rename for this one)

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=shiny crew update
```


Zopolis4:
Also, I've got some refactoring in mind for `version.rb`, as well as implementing libversion support so we don't get things like this:
```
Package                            Status              Current             Upstream
-------                            ------              -------             --------
a2png                              outdated            0.1.5-1             0.1.5
```

uberhacker (approved):
I wasn't able to find anything wrong so LGTM.

PR #9859, opened 2024-05-28
Rename zlibpkg to zlib

Zopolis4:
Tested (including fresh installs) and working on all architectures. 

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=gleam crew update
```


packages/zlib.rb
```diff
@@ -1,10 +1,13 @@
 require 'buildsystems/cmake'
 
-class Zlibpkg < CMake
+# Remove the defenition of Ruby's Zlib module here to avoid conflicts.
+Object.send(:remove_const, :Zlib) if Object.constants.include?(:Zlib)
```
uberhacker, line 4: This appears to be a potential issue.  What if Ruby needs to use Zlib internally?  This is a major change for a minor issue just because you don't like the package name.

Zopolis4: We don't use the Zlib module ourselves, and I didn't encounter any issues in my testing.

supechicken: This is handled by #7283 already, same issue happened on `readline.rb` previously

Zopolis4: If this line is not there, things break...

uberhacker: So how can we fix this without adding this line?

Zopolis4: If there was a way to do it I would have used it-- this is how we fix it, unless someone finds something else.

supechicken: > If this line is not there, things break...

For me it works after renaming the existing `zlibpkg.rb` to `zlib.rb` and updating the class name

uberhacker: So are you suggesting any changes here @supechicken?

packages/zlib.rb
```diff
@@ -1,10 +1,13 @@
 require 'buildsystems/cmake'
 
-class Zlibpkg < CMake
+# Remove the defenition of Ruby's Zlib module here to avoid conflicts.
```
uberhacker, line 3: definition?

packages/zlib.rb
supechicken: > So are you suggesting any changes here @supechicken?

Yes, my suggestion is to remove this line.

Zopolis4: image

supechicken: I think I found the root cause of the error above. We should not use `require_relative` to refer to other package files, as `require_relative` does all the things on the global scope instead of the isolated `Package` class.

https://github.com/chromebrew/chromebrew/blob/81a34411f1d19166cf231b6f9d4c7f2962497105/packages/minizip.rb#L1-L5

A correct way to deal with this issue is to replace `require_relative` to `Package.load_package()` in all package scripts

uberhacker: Sounds like a lot more changes. LoL

packages/zlib.rb
Zopolis4: good catch!

uberhacker (approved):

PR #9910, opened 2024-06-05
Fixup gcc_lib and crew update for very old milestone installs

### LINKED ISSUE #9901: 🐞 Bug Lib_GCC error
**Describe the bug**
During the install.sh script, when it does the `crew install core` command, i get this error when its trying to install gcc_lib

```gcc_lib: GCC shared libs except libgccjit
Skipping source download...
Preconfiguring package...
Installing Gcc_build to pull files for build...
File list for Gcc_build does not exist!
An error occured during the installation :/
```

**To Reproduce**

My environment was (crosh, bash or VT-2): crosh

      ```shell
      bash <(curl -L git.io/vddgY) && . ~/.bashrc
      ```
**Expected behavior**

Not have an error there.

**Screenshots**

***

- User space: `64-bit`
- Chromebrew Kernel version: `4.19`

- Chromebrew version: `1.48.4`

- Last update in local repository: 

- OS variant: `Chrome OS`
- OS version: `edgar-release/R103-14816.131.0`

</details>
***

satmandu:
Probably Fixes #9901 

- These issues come up when I open the M106 container...

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=old_install_fix crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -4,7 +4,14 @@ require 'digest/sha2'
 require 'json'
 require 'fileutils'
 require 'tmpdir'
-require_relative '../commands/const'
+begin
+  require_relative '../commands/const'
+rescue LoadError
+  # Older crew installs won't have the commands dir in the sparse checkout,
+  # so disable sparse checkout if those files are missing.
+  system 'git sparse-checkout disable'
+  require_relative '../commands/const'
+end
 require_relative '../commands/files'
 require_relative '../commands/help'
 require_relative '../commands/list'
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -2,7 +2,7 @@
 # Defines common constants used in different parts of crew
 require 'etc'
 
-CREW_VERSION = '1.48.4'
+CREW_VERSION = '1.48.5'
 
 # kernel architecture
 KERN_ARCH = Etc.uname[:machine]
@@ -50,7 +50,7 @@ CREW_LIB_SUFFIX = ARCH.eql?('x86_64') && Dir.exist?('/lib64') ? '64' : ''
 ARCH_LIB        = "lib#{CREW_LIB_SUFFIX}"
 
 # Glibc version can be found from the output of libc.so.6
-LIBC_VERSION = Etc.confstr(Etc::CS_GNU_LIBC_VERSION).split.last
+LIBC_VERSION = ENV.fetch('LIBC_VERSION', Etc.confstr(Etc::CS_GNU_LIBC_VERSION).split.last )
 
 CREW_PREFIX = ENV.fetch('CREW_PREFIX', '/usr/local')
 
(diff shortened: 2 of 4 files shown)


bin/crew
```diff
@@ -4,7 +4,14 @@ require 'digest/sha2'
 require 'json'
 require 'fileutils'
 require 'tmpdir'
-require_relative '../commands/const'
+begin
+  require_relative '../commands/const'
+rescue LoadError
+  # Older crew installs won't have the commands dir in the sparse checkout,
```
Zopolis4, line 10: Don't we already handle this in lib/fixup.rb?

lib/const.rb
```diff
@@ -50,7 +50,7 @@
 ARCH_LIB        = "lib#{CREW_LIB_SUFFIX}"
 
 # Glibc version can be found from the output of libc.so.6
-LIBC_VERSION = Etc.confstr(Etc::CS_GNU_LIBC_VERSION).split.last
+LIBC_VERSION = ENV.fetch('LIBC_VERSION', Etc.confstr(Etc::CS_GNU_LIBC_VERSION).split.last )
```
Zopolis4, line 53: What was the problem with the old method?

Zopolis4 (changes requested):

bin/crew
satmandu: The error happens when loading crew, before we get to fixup.

lib/const.rb
satmandu: Trying to set `LIBC_VERSION` manually for when we need to override it, as when we are building the various packages with that used in the version field.

satmandu:
What's the requested change @Zopolis4 ?

bin/crew
Zopolis4: But how will we have updated to a newer state without running fixup?

satmandu: Git pull has already happened at this point. Disabling the sparse checkout unhides all the files needed.

satmandu: The git pull leaves crew in a broken state, because the commands dir is not yet part of the sparse checkout.

Zopolis4 (approved):
Not entirely sure about the sparse-checkout stuff, but i'll look into that without holding up an install fix.

jacobh12349:
So, do i just run the install script again to fix it?

jacobh12349:
> So do i just run the install script again to fix it?

I'm just gonna do that now, and see if it works, and come back if it doesn't.

jacobh12349:
I've also seen this while the script is running too, but the script still continues running though, so i dont know if its that important.

```
fatal: not a git repository (or any parent up to mount point /usr)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

jacobh12349:
Ya, gcc_lib installed fine now, and the rest of the install script all worked fine now too, thx

PR #9912, opened 2024-06-05
Handle case where ruby needs an upgrade, but openssl also needs to be upgraded, and other M106 fixes.

satmandu:
- Trying to fixup the M106 container.
- This fixes 99% of issues opening the M106 container and updating/upgrading to current.
- Left for another day is fixing the upgrade of older versions of installed is_fake packages like `gcc`, `llvm`, etc...

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=ruby_openssl_upgrade crew update ; yes | crew upgrade
```

### DIFF
--- a/lib/buildsystems/pip.rb
+++ b/lib/buildsystems/pip.rb
@@ -39,7 +39,7 @@ class Pip < Package
       @pip_path = File.expand_path("#{@pip_files_base}#{pip_file}")
       @destpath = File.join(CREW_DEST_DIR, @pip_path)
       # Handle older FileUtils from older ruby versions.
-      FileUtils.mkdir_p File.join(CREW_DEST_DIR, @pip_files_base) if Gem::Version.new(RUBY_VERSION.to_s) < Gem::Version.new('3.3')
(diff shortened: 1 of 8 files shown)


lib/buildsystems/pip.rb
```diff
@@ -37,7 +37,9 @@ def self.install
     @pip_files_lines = @pip_files[/(?<=Files:\n)[\W|\w]*/, 0].split
     @pip_files_lines.each do |pip_file|
       @pip_path = File.expand_path("#{@pip_files_base}#{pip_file}")
-      @destpath = "#{CREW_DEST_DIR.chomp('/')}#{@pip_path}"
+      @destpath = File.join(CREW_DEST_DIR, @pip_path)
+      # Handle older FileUtils from older ruby versions.
+      FileUtils.mkdir_p File.dirname(@destpath) if Gem::Version.new(RUBY_VERSION.to_s) < Gem::Version.new('3.3')
```
Zopolis4, line 42: Do we run this code before we have upgraded ruby?

uberhacker: Doesn't hurt to have the check anyway.

uberhacker (approved):
LGTM

PR #9913, opened 2024-06-05
Swap version.rb to `ruby_libversion` instead of `Gem::Version`

Zopolis4:
Still got a couple of ideas for `version.rb`, but this should help quite a bit for version checking logic, especially when we patch upstream releases or have rebuilds.

See:
image

We now have a few false positives because we're accepting higher version levels as uptodate rather than a mismatch, but this is correct in the vast majority of cases and is the best choice.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=vcheck2 crew update
```

### DIFF
--- a/tools/version.rb
+++ b/tools/version.rb
@@ -9,6 +9,7 @@ end
 
 require 'json'
 require 'net/http'
+require 'ruby_libversion'
 
 # Add >LOCAL< lib to LOAD_PATH
 $LOAD_PATH.unshift '../lib'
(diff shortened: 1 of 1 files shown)


tools/version.rb
```diff
@@ -117,35 +113,24 @@ def valid_semantic_version?(version)
...
+    # Bail out if we arent verbose and so dont want to print packages that are up to date.
+    next if Libversion.version_compare2(pkg.version, upstream_version) >= 0 && !verbose
+    # Print the package name.
+    print pkg.name.ljust(35)
+    # Print the package update status.
+    if Libversion.version_compare2(pkg.version, upstream_version) >= 0
+      print 'uptodate'.ljust(20).lightgreen
+    elsif Libversion.version_compare2(pkg.version, upstream_version) == -1
+      print 'outdated'.ljust(20).yellow
+    end
```
uberhacker, line 132: What do you do in cases where the updated version is less than the current version.  Not just outdated in those cases.

Zopolis4: As per the PR description:
```
We now have a few false positives because we're accepting higher version levels as uptodate rather than a mismatch, but this is correct in the vast majority of cases and is the best choice.
```

See `a2png`, `antiword

Zopolis4:
Converting to draft just so I can iron out a few issues in the algorithm.

Zopolis4:
Current ideas for improving the algorithm:
- Fallback homepage matching (strip trailing slashes, `http` v `https`)
- Local version normalization (drop `-llvm`, `-py3.12`

As of now though, this is a marked improvement.

satmandu (approved):

PR #9915, opened 2024-06-06
Update homepages of packages from a-c

Zopolis4:
Going through packages and following any homepage redirects.

Also, in some cases, I've updated the `source_url` if I noticed that it wasn't working.

I've left the mirror redirects that go to different homepages at different times intact.

Some packages have homepages that are dead upstream:
- `anagram`, where I can't access the homepage normally but it works fine if I'm in Germany, so not sure whats going on there-- I've updated the `source_url` to the github, which will always work, but kept the homepage for now.
- `apg`, which is just dead upstream entirely. We should swap to the https://github.com/jabenninghoff/apg fork when possible.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=homepage0 crew update
```

The script:
```ruby
require 'net/http'

# Add >LOCAL< lib to LOAD_PATH
$LOAD_PATH.unshift '../lib'

def get_final_destination(url)
  uri = URI(url)
  response = Net::HTTP.get_response(uri)

  if response.is_a?(Net::HTTPRedirection)
    redirect_url = response['location']
    get_final_destination(redirect_url)
  else
    uri.to_s
  end
end

  next if pkg.is_fake?
  redirected_homepage = get_final_destination(pkg.homepage)
  next unless redirected_homepage != pkg.homepage
  puts "#{pkg.name}\n"
  puts "Current homepage: #{pkg.homepage}"
  puts "Redirected homepage: #{redirected_homepage}"
  print 'Commit these changes? [Y/n] '
  response = $stdin.gets.chomp.downcase
  case response
  when 'n', 'no'
    puts "Skipping..."
    next
  when '', 'y', 'yes'
    puts 'Proceeding...'
    system ("sed -i -e 's|#{pkg.homepage}|#{redirected_homepage}|' #{filename}")
    system ('git commit -va --dry-run')
    system ("git commit -m 'Update #{pkg.name} homepage' -o #{filename}")
  else
    puts "I don't understand `#{response}`. :(".lightred
    abort 'No changes made.'
  end
end
```

### DIFF
--- a/packages/acpica.rb
+++ b/packages/acpica.rb
@@ -2,12 +2,12 @@ require 'package'
 
 class Acpica < Package
   description 'ACPI tools, including Intel ACPI Source Language compiler'
-  homepage 'https://acpica.org/'
+  homepage 'https://www.intel.com/content/www/us/en/developer/topic-technology/open/acpica/overview.html'
   version '20200925'
   license 'GPL-2'
   compatibility 'all'
-  source_url 'https://acpica.org/sites/acpica/files/acpica-unix-20200925.tar.gz'
-  source_sha256 'd44388e21e3d2e47c6d39e9c897935d3f775f04fec76271dcba072c74f834589'
+  source_url 'https://github.com/acpica/acpica.git'
+  git_hashtag 'R09_25_20'
   binary_compression 'tar.xz'
 
   binary_sha256({
--- a/packages/alpine.rb
+++ b/packages/alpine.rb
@@ -2,11 +2,11 @@ require 'package'
 
 class Alpine < Package
   description 'The continuation of the Alpine email client from University of Washington.'
-  homepage 'http://alpine.x10host.com/alpine'
+  homepage 'https://alpineapp.email/'
   version '2.25'
   license 'Apache-2.0'
   compatibility 'all'
-  source_url 'http://alpine.x10host.com/alpine/release/src/alpine-2.25.tar.xz'
+  source_url 'https://alpineapp.email/alpine/release/src/Old/alpine-2.25.tar.xz'
   source_sha256 '658a150982f6740bb4128e6dd81188eaa1212ca0bf689b83c2093bb518ecf776'
   binary_compression 'tar.zst'
 
--- a/packages/anagram.rb
+++ b/packages/anagram.rb
@@ -4,10 +4,10 @@ class Anagram < Package
   description 'finds anagrams or permutations of words in the target phrase'
   homepage 'https://www.fourmilab.ch/anagram/'
   version '1.5'
-  license 'public-domain'
+  license 'CC0-1.0'
   compatibility 'all'
-  source_url 'https://www.fourmilab.ch/anagram/anagram-1.5.tar.gz'
-  source_sha256 '62eca59318782e889118a0e130d454e1c397aedd99fc59b2194393bf0eff5348'
+  source_url 'https://github.com/Fourmilab/anagram.git'
+  git_hashtag '5466ccd28c1b8f46bd8af420d17801bdf1da51a8'
   binary_compression 'tar.zst'
 
   binary_sha256({
--- a/packages/avahi.rb
+++ b/packages/avahi.rb
@@ -2,7 +2,7 @@ require 'buildsystems/autotools'
 
 class Avahi < Autotools
   description 'Avahi is a system which facilitates service discovery on a local network via the mDNS/DNS-SD protocol suite.'
-  homepage 'http://www.avahi.org/'
+  homepage 'https://avahi.org/'
   version '0.9-rc1'
   license 'LGPL-2.1'
   compatibility 'all'
--- a/packages/aide.rb
+++ b/packages/aide.rb
@@ -2,7 +2,7 @@ require 'package'
 
 class Aide < Package
   description 'Advanced Intrusion Detection Environment'
-  homepage 'http://aide.sourceforge.net/'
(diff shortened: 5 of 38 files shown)


packages/acpica.rb
```diff
@@ -2,11 +2,11 @@
...
   description 'ACPI tools, including Intel ACPI Source Language compiler'
-  homepage 'https://acpica.org/'
+  homepage 'https://www.intel.com/content/www/us/en/developer/topic-technology/open/acpica/overview.html'
   version '20200925'
   license 'GPL-2'
   compatibility 'all'
-  source_url 'https://acpica.org/sites/acpica/files/acpica-unix-20200925.tar.gz'
+  source_url 'https://downloadmirror.intel.com/774848/acpica-unix-20191213.tar.gz'
   source_sha256 'd44388e21e3d2e47c6d39e9c897935d3f775f04fec76271dcba072c74f834589'
   binary_compression 'tar.xz'
```
satmandu, line 11: Should we not rebuild these packages of we are updating the sources? And also change the `binary_compression` field then to `tar.zst`?

satmandu: Also, this replacement source appears to be older?

Zopolis4: Oops, I copied the wrong url-- will fix when I wake up.

packages/avahi.rb
```diff
@@ -2,7 +2,7 @@
 
 class Avahi < Autotools
   description 'Avahi is a system which facilitates service discovery on a local network via the mDNS/DNS-SD protocol suite.'
-  homepage 'http://www.avahi.org/'
+  homepage 'http://avahi.org/'
```
uberhacker, line 5: Change to `https://avahi.org/`?

packages/acpica.rb
uberhacker: https://github.com/acpica/acpica/files/14718333/acpica-unix-20240321.tar.gz ?

packages/acpica.rb
```diff
@@ -2,12 +2,12 @@
...
   description 'ACPI tools, including Intel ACPI Source Language compiler'
-  homepage 'https://acpica.org/'
+  homepage 'https://www.intel.com/content/www/us/en/developer/topic-technology/open/acpica/overview.html'
   version '20200925'
   license 'GPL-2'
   compatibility 'all'
-  source_url 'https://acpica.org/sites/acpica/files/acpica-unix-20200925.tar.gz'
-  source_sha256 'd44388e21e3d2e47c6d39e9c897935d3f775f04fec76271dcba072c74f834589'
+  source_url 'https://github.com/acpica/acpica.git'
+  git_hashtag 'R09_25_20'
```
uberhacker, line 10: This is old.  Why not update?

packages/alpine.rb
```diff
@@ -2,11 +2,11 @@
 
 class Alpine < Package
   description 'The continuation of the Alpine email client from University of Washington.'
-  homepage 'http://alpine.x10host.com/alpine'
+  homepage 'https://alpineapp.email/'
   version '2.25'
   license 'Apache-2.0'
   compatibility 'all'
-  source_url 'http://alpine.x10host.com/alpine/release/src/alpine-2.25.tar.xz'
+  source_url 'https://alpineapp.email/alpine/release/src/Old/alpine-2.25.tar.xz'
```
uberhacker, line 9: Update to https://alpineapp.email/alpine/release/src/alpine-2.26.tar.xz?

uberhacker: This will prove the new domain compiles correctly with the current package configuration.

packages/anagram.rb
```diff
@@ -4,10 +4,10 @@ class Anagram < Package
   description 'finds anagrams or permutations of words in the target phrase'
   homepage 'https://www.fourmilab.ch/anagram/'
   version '1.5'
-  license 'public-domain'
+  license 'CC0-1.0'
   compatibility 'all'
-  source_url 'https://www.fourmilab.ch/anagram/anagram-1.5.tar.gz'
-  source_sha256 '62eca59318782e889118a0e130d454e1c397aedd99fc59b2194393bf0eff5348'
+  source_url 'https://github.com/Fourmilab/anagram.git'
+  git_hashtag '5466ccd28c1b8f46bd8af420d17801bdf1da51a8'
```
uberhacker, line 10: Definitely should rebuild the binaries to prove this update doesn't break the current package.

packages/alpine.rb
Zopolis4: image
Its the same file with the same hash.

packages/acpica.rb
Zopolis4: As I said-- I'm fixing the URLs. That's all.

packages/anagram.rb
Zopolis4: image

Zopolis4: Git version:
image

Tar version:
image

packages/alpine.rb
uberhacker: The amount of effort to prove the code is the same could be better spent going ahead with the update but I digress...

packages/anagram.rb
Zopolis4: The old url is inconsistent-- I wasn't able to access it yesterday. Github will always work.

uberhacker: Not sure what that means but whatever...

uberhacker (approved):
Thank you for finding good repos.

PR #9917, opened 2024-06-07
Nethack 3.6.0 => 3.6.7

uberhacker:
Normally I don't give much attention to any game but I took on this challenge as a good exercise to consider every angle to get a successful build.  It's likely we will run into other packages like this so this serves as a potential template.  Tested and working on all platforms.


Zopolis4:
Nice! Are these patches applicable upstream?

uberhacker:
> Nice! Are these patches applicable upstream?

No, the latest release already has the update.

Zopolis4 (approved):

PR #9927, opened 2024-06-07
Stop passing linker flags to ar via -DCMAKE_STATIC_LINKER_FLAGS

Zopolis4:
As the [CMake documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_STATIC_LINKER_FLAGS.html) says, static libraries don't actually link, so we're just passing linker flags to `ar` for no reason.

Passing `-fno-lto` breaks builds:
image

And I presume `-flto=auto` doesn't trigger a failure by chance.

```
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=mountain crew update
```


uberhacker:
LGTM but I will trust @satmandu's judgement here.

satmandu (approved):

PR #9944, opened 2024-06-10
Add buildessential dependency to py3_mercurial

uberhacker:


supechicken:
But... Why? It shouldn't be necessary to include the whole compiling toolset as an runtime dependency, we should only include those essential dependencies it need, preventing bloated installation

PR #9955, opened 2024-06-12
Fix Repology workflow

Zopolis4:
Should work, hopefully.


uberhacker (approved):
Not sure I understand what is going on here but I'll trust you know what you are doing.

PR #9976, opened 2024-06-16
Remove CREW_CMAKE_FNO_LTO_OPTIONS

Zopolis4:
Removing `CREW_LINKER_FLAGS` was pretty simple, nothing of note there.

However, I have marked this PR as draft to get additional feedback on the `CREW_CMAKE_FNO_LTO_OPTIONS` changes-- there's more manual modification than I'd like, and thats because the affected packages aren't on buildsystems yet.

I could probably switch most of them over to buildsystems, but I'd like @satmandu's thoughts first, because he's the one who builds them.

Also, why do these packages select `CREW_CMAKE_FNO_LTO_OPTIONS` but then delete `-fno-lto`? If I knew why they were doing that and if that needed to be done, things could probably be simpler.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=bowling crew update
```

### DIFF
--- a/lib/buildsystems/cmake.rb
+++ b/lib/buildsystems/cmake.rb
@@ -5,7 +5,7 @@ class CMake < Package
 
   def self.build
     @cmake_build_relative_dir ||= '.'
-    @crew_cmake_options = @no_lto ? CREW_CMAKE_FNO_LTO_OPTIONS : CREW_CMAKE_OPTIONS
+    @crew_cmake_options = @no_lto ? CREW_CMAKE_OPTIONS.gsub('-flto=auto', '-fno-lto').sub('-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=TRUE', '') : CREW_CMAKE_OPTIONS
     puts 'Additional cmake options being used:'.orange
     method_list = methods.grep(/cmake_/).delete_if { |i| send(i).blank? }
     method_list.each do |method|
--- a/packages/libclc.rb
+++ b/packages/libclc.rb
@@ -52,7 +52,7 @@ class Libclc < Package
   def self.build
     @cmake_options = case ARCH
                      when 'i686', 'x86_64'
-                       CREW_CMAKE_FNO_LTO_OPTIONS.gsub('-fno-lto', '')
+                       CREW_CMAKE_OPTIONS.gsub('-flto=auto', '').sub('-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=TRUE', '')
                      else
                        CREW_CMAKE_OPTIONS
                      end
--- a/packages/webkitgtk_6.rb
+++ b/packages/webkitgtk_6.rb
@@ -148,7 +148,7 @@ class Webkitgtk_6 < Package
       @arch_linker_flags = ARCH == 'x86_64' ? '' : '-Wl,--no-keep-memory'
       system "CREW_LINKER_FLAGS='#{@arch_linker_flags}' CC='#{@workdir}/bin/gcc' CXX='#{@workdir}/bin/g++' \
           cmake -B builddir -G Ninja \
-          #{CREW_CMAKE_FNO_LTO_OPTIONS.gsub('-DCMAKE_LINKER_TYPE=MOLD', '').sub('-pipe', '-pipe -Wno-error').gsub('-fno-lto', '')} \
+          #{CREW_CMAKE_OPTIONS.sub('-pipe', '-pipe -Wno-error').gsub('-flto=auto', '').sub('-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=TRUE', '')} \
           -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON \
           -DENABLE_BUBBLEWRAP_SANDBOX=OFF \
           -DENABLE_DOCUMENTATION=OFF \
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -268,7 +268,6 @@ CREW_NINJA ||= ENV.fetch('CREW_NINJA', 'ninja') unless defined?(CREW_NINJA)
 # Cmake sometimes wants to use LIB_SUFFIX to install libs in LIB64, so specify such for x86_64
 # This is often considered deprecated. See discussio at https://gitlab.kitware.com/cmake/cmake/-/issues/18640
 # and also https://bugzilla.redhat.com/show_bug.cgi?id=1425064
-# Let's have two CREW_CMAKE_OPTIONS since this avoids the logic in the recipe file.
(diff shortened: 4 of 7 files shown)


lib/buildsystems/cmake.rb
```diff
@@ -5,7 +5,7 @@ class CMake < Package
 
   def self.build
     puts "Additional cmake_options being used: #{@pre_cmake_options.nil? ? '<no pre_cmake_options>' : @pre_cmake_options} #{@cmake_options.nil? ? '<no cmake_options>' : @cmake_options}".orange
-    @crew_cmake_options = @no_lto ? CREW_CMAKE_FNO_LTO_OPTIONS : CREW_CMAKE_OPTIONS
+    @crew_cmake_options = @no_lto ? CREW_CMAKE_OPTIONS.gsub('-flto=auto', '-fno-lto').sub('-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=TRUE', '') : CREW_CMAKE_OPTIONS
```
satmandu, line 8: Why not change `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=TRUE'` to `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=FALSE'` ?

Zopolis4: It's disabled by default, so no need.

packages/libclc.rb
```diff
@@ -51,7 +51,7 @@ def self.patch
   def self.build
     @cmake_options = case ARCH
                      when 'i686', 'x86_64'
-                       CREW_CMAKE_FNO_LTO_OPTIONS.gsub('-fno-lto', '')
+                       CREW_CMAKE_OPTIONS.gsub('-flto=auto', '').sub('-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=TRUE', '')
```
satmandu, line 54: Why can't we make `no_lto` work for this sort of thing?

Zopolis4: This behaviour isn't what `no_lto` normally does. `no_lto` will cause `-fno-lto` to be passed instead of `-flto=auto`, but this is just removing `-flto=auto`. 

I'm not exactly sure why this is being done, so I didn't change it to normal.

satmandu: Can we change no_lto to both disable lto and also disable flto=auto?

Zopolis4: `flto=auto` is lto, not sure what you mean.

satmandu: I mean that `no_lto` should remove all added `lto` lines such as `-flto=auto` and also set the other lto lines such as the interprocedural cmake option to false. We should not have to edit cmake_options at the package level for every package. Just putting `no_lto` in the package file should be sufficient.

Zopolis4: `no_lto` does that. These packages have other behavior, which is why `no_lto` does not work for them. As you added the packages, you have the best understanding of why they do this and if it is necessary.

satmandu: I'm just saying that we should make things so that in package files we never need to adjust for no_lto beyond saying no_lto...

satmandu: The custom modifications are from before we had a functional no_lto...

Zopolis4: Ah, I see.

Zopolis4:
Unfortunately, `no_lto` does not work with `no_env_options`, so even if the majority of affected packages did not further modify `CREW_CMAKE_OPTIONS`, they could not be converted to use `no_lto`.

satmandu:
FYI I use `CREW_LINKER_FLAGS` in https://github.com/chromebrew/chromebrew/pull/10272 to get the newer ChromeOS glibc to work for building...

Zopolis4:
> FYI I use `CREW_LINKER_FLAGS` in #10272 to get the newer ChromeOS glibc to work for building...

I've dropped that change, now this is just removing `CREW_CMAKE_FNO_LTO_OPTIONS`.

packages/webkitgtk_6.rb
```diff
@@ -148,7 +148,7 @@ def self.build
       @arch_linker_flags = ARCH == 'x86_64' ? '' : '-Wl,--no-keep-memory'
       system "CREW_LINKER_FLAGS='#{@arch_linker_flags}' CC='#{@workdir}/bin/gcc' CXX='#{@workdir}/bin/g++' \
           cmake -B builddir -G Ninja \
-          #{CREW_CMAKE_FNO_LTO_OPTIONS.gsub('-DCMAKE_LINKER_TYPE=MOLD', '').sub('-pipe', '-pipe -Wno-error').gsub('-fno-lto', '')} \
+          #{CREW_CMAKE_OPTIONS.gsub('mold', 'gold').sub('-pipe', '-pipe -Wno-error').gsub('-flto=auto', '').sub('-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=TRUE', '')} \
```
satmandu, line 151: Gold linker is deprecated: https://www.phoronix.com/news/GNU-Gold-Linker-Deprecated

We shouldn't add it back in anywhere.

satmandu:
Aside from not adding `gold` linker back in anywhere, I am fine with this.

packages/webkitgtk_6.rb
Zopolis4: Good point. I've replaced it with the generic `ld`.

satmandu:
Would you please just modify all the `.gsub('mold', 'gold')` and `.gsub('mold', 'ld')` lines here to `.gsub('mold', '')`?

satmandu:
> Would you please just modify all the `.gsub('mold', 'gold')` and `.gsub('mold', 'ld')` lines here to `.gsub('mold', '')`?

Actually, you can just remove all the `.gsub('mold', ...` bits because `mold` in lower case doesn't even come up:
```
crew const CREW_CMA
KE_OPTIONS
CREW_CMAKE_OPTIONS=-DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_LIBRARY_PATH=/usr/local/lib64 -DCMAKE_C_FLAGS='-O3 -pipe -ffat-lto-objects -fPIC   -flto=auto' -DCMAKE_CXX_FLAGS='-O3 -pipe -ffat-lto-objects -fPIC   -flto=auto' -DCMAKE_EXE_LINKER_FLAGS='-flto=auto ' -DCMAKE_LINKER_TYPE=MOLD -DCMAKE_SHARED_LINKER_FLAGS='-flto=auto ' -DCMAKE_MODULE_LINKER_FLAGS='-flto=auto ' -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=TRUE -DCMAKE_BUILD_TYPE=Release
```

Zopolis4:
Fixed.

satmandu (approved):

PR #9980, opened 2024-06-16
Enable more rubocop cops

Zopolis4:
Trying to cut down on the amount of cops we disable that aren't stylistic choices.

Would have enabled `Style/OptionalBooleanParameter`, but that triggers `Style/HashSyntax` on keyword arguments with the same name as the value we are passing to them (i.e. `verbose: verbose`), so we need rubocop/rubocop#12132 to be fixed before that works.

@supechicken, would appreciate your help on in these areas of code, I'm not confident enough in the behaviour to make the changes needed to fix the lint errors.
Screenshot_20240616_115515

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=tubo crew update
```


satmandu (approved):
_LGTM_

uberhacker:
Less is more, right?  Especially when it comes to heavy-handed, boots on the neck cops.

Zopolis4:
I agree-- that is why I only chose relatively minor cops with desirable behavior to re-enable. If you disagree with any of my choices, I'm happy to discuss.

PR #9981, opened 2024-06-16
Simplify ARCH detection and rework reading /proc/cpuinfo into CPUINFO

Zopolis4:
Now that we can set `ARCH` via an environment variable, there's no need to have all the awkward workarounds for containers, as they can simply set the desired architecture.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=back crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -2,47 +2,18 @@
 # Defines common constants used in different parts of crew
 require 'etc'
 
-CREW_VERSION = '1.49.2'
+CREW_VERSION = '1.49.3'
 
-# kernel architecture
+# Kernel architecture.
 KERN_ARCH = Etc.uname[:machine]
 
-# read and parse processor information from /proc/cpuinfo
-CPUINFO = File.read('/proc/cpuinfo') \
-              .partition("\n\n")[0] \
-              .scan(/^(.+?)\t*: (.+)$/).to_h \
-              .transform_keys(&:downcase)
+# Read and parse processor information from /proc/cpuinfo
+CPUINFO = File.readlines('/proc/cpuinfo').map { |line| line.chomp.split(/\t+: /) if line.include?("\t") }.compact.to_h
 
-# get architectures supported by the processor natively
-CPU_SUPPORTED_ARCH = \
-  if CPUINFO.key?('flags')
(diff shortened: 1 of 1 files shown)


satmandu:
The detection code is useful, I think, since we use minor corrections to what the containers give us to emulate what shows up on actual hardware.

It's not just a matter of forcing an ARCH value...

Zopolis4:
Isn't it? To me this just seems like leftover code from the days when we couldn't set ARCH via an environment variable, so this can all be replaced by just setting ARCH as armv7l in a container that would otherwise report armv8l.

satmandu:
The code may be useful, and I'm not comfortable getting rid of it while we are still having architecture specific issues that might be more granular than we assume...

Zopolis4:
Do we have any of those issues outstanding? I was under the impression that most of this code wasn't getting used anyways...

satmandu:
> Do we have any of those issues outstanding? 

We still don't have aarch64 working!

>I was under the impression that most of this code wasn't getting used anyways...

I think it is premature to assume that when we still have architectures not working properly. And aarch64 user space definitely fits that.

That is I think reason enough to keep this plumbing in for now...

Zopolis4:
This code is only for containers, though. Thats not particularly related to the aarch64 situation.

satmandu:
99% of the testing (not just for new architectures) happens in containers... And that goes for aarch64 too.

I literally don't have an aarch64 ChromeOS device ... Everything I do is in containers.

Zopolis4:
Just to be clear-- all the code around `CPU_SUPPORTED_ARCH` is unused. Same with `QEMU_EMULATED`. 

Neither of those variables are used.

The only change I'm making is that `armv8l` (only encountered in certain container setups) is no longer automatically converted to `armv7l`, and now the `ARCH` environment variable has to be used.

lib/const.rb
```diff
@@ -4,45 +4,16 @@
 
 CREW_VERSION = '1.49.2'
```
uberhacker, line 5: Bump version?

satmandu (approved):

satmandu (approved):

PR #9990, opened 2024-06-18
Update sommelier => 20240607

satmandu:
- This updates `sommelier` to the latest `chromiumos/platform2` vm_tools git tag.

Builds properly:
- [x] `x86_64`
- [x] `armv7l` 

_Confirmed working on:_
- [x] `x86_64`
- [x] `armv7l` 

```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=sommelier crew update
yes | crew upgrade ; restartsommelier
```


packages/sommelier.rb
```diff
@@ -360,10 +361,10 @@ def self.build
             # Set default SCALE to 1 if unset.
             SCALE=${SCALE:-1}
             # Allow overriding environment variables before starting sommelier daemon.
-            [ -f "$HOME/.sommelier.env" ] && source ~/.sommelier.env 2>> #{CREW_PREFIX}/var/log/sommelier.log
+            [ -f "#{CREW_PREFIX}/.config/.sommelier.env" ] && source #{CREW_PREFIX}/.config/.sommelier.env 2>> #{CREW_PREFIX}/var/log/sommelier.log
```
uberhacker, line 364: How about add to `#{CREW_PREFIX}/etc/env.d`?

satmandu: This is for user overrides to the defaults that won't be overwritten by package updates, so that would be inappropriate.

packages/sommelier.rb
```diff
@@ -459,18 +460,31 @@ def self.postinstall
...
-    FileUtils.touch "#{HOME}/.sommelier.env" unless File.exist? "#{HOME}/.sommelier.env"
+    unless File.exist? "#{CREW_PREFIX}/.config/.sommelier.env"
+      FileUtils.mkdir_p "#{CREW_PREFIX}/.config"
+      if File.exist? "#{HOME}/.sommelier.env"
+        FileUtils.cp "#{HOME}/.sommelier.env", "#{CREW_PREFIX}/.config/.sommelier.env"
+        ExitMessage.add <<~EOT0.lightblue
+
+          Overrides of the default sommelier configuration must now be stored in:
+          #{CREW_PREFIX}/.config/.sommelier.env
+        EOT0
```
uberhacker, line 476: Not needed if added to `#{CREW_PREFIX}/etc/env.d`?

satmandu: This is for local user modifications to the config, so we don't want to put those into config files in env.d that the user expects to be overwritten.

The move to .config is because the HOME dir has noexec set in newer ChromeOS milestones...

packages/sommelier.rb
```diff
@@ -489,11 +503,13 @@ def self.postinstall
...
       Please open a github issue at
       https://github.com/chromebrew/chromebrew/issues/new/choose
-      with the output of both
+      with the output of
        readlink -f '/sys/class/drm/renderD129/device/driver'
       and
        readlink -f '/sys/class/drm/renderD128/device/driver'
-      if sommelier does not start properly on your arm ChromeOS device.
+      and
+       tail #{CREW_PREFIX}/var/log/sommelier.log
```
uberhacker, line 511: Surprised rubocop didn't complain about the indentation here.

satmandu: It doesn't care about text blocks?

packages/sommelier.rb
uberhacker: Makes sense for overrides.

uberhacker (approved):

PR #9998, opened 2024-06-19
Fix error with autotools builds

### LINKED ISSUE #9997: 🐞 Bug Issue with autotools buildsystem
```
Additional configure_options being used: <no pre_configure_options> --localstatedir=/usr/local/tmp
Command "bash -c grep -q /usr/bin/file configure" failed, retrying with "-j1"...
There was a build error.
```

uberhacker:
Fixes #9997.

### DIFF
--- a/lib/buildsystems/autotools.rb
+++ b/lib/buildsystems/autotools.rb
@@ -19,7 +19,7 @@ class Autotools < Package
       end
       abort 'configure script not found!'.lightred unless File.file?('configure')
       FileUtils.chmod('+x', 'configure')
-      if system('grep -q /usr/bin/file configure')
+      if Kernel.system('grep -q /usr/bin/file configure')
         puts 'Using filefix.'.orange
         system 'filefix'
       end
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -2,7 +2,7 @@
 # Defines common constants used in different parts of crew
 require 'etc'
 
-CREW_VERSION = '1.49.1'
+CREW_VERSION = '1.49.2'
 
 # kernel architecture
 KERN_ARCH = Etc.uname[:machine]


lib/buildsystems/autotools.rb
```diff
@@ -19,7 +19,7 @@ def self.build
       end
       abort 'configure script not found!'.lightred unless File.file?('configure')
       FileUtils.chmod('+x', 'configure')
-      if system('grep -q /usr/bin/file configure')
+      if File.executable? "#{CREW_PREFIX}/bin/file"
         puts 'Using filefix.'.orange
```
satmandu, line 23: Shouldn't we only report we are running filefix if `/usr/bin/file` is actually in the configure script?

Sure checking to see if a local copy of `file` exists first too is a good idea, but that check can come afterwards only if we actually need to run filefix...

uberhacker: I think it should be fine?  It aborts if configure is not found 2 lines above that.

lib/buildsystems/autotools.rb
```diff
@@ -19,7 +19,7 @@ def self.build
       end
       abort 'configure script not found!'.lightred unless File.file?('configure')
       FileUtils.chmod('+x', 'configure')
-      if system('grep -q /usr/bin/file configure')
```
Zopolis4, line 22: The problem here is that our custom system command defined in `lib/package.rb` will retry on failure, where here we actually don't want it do do that. @supechicken, any ideas on how to access the ruby `system` here? Maybe `Kernel.system()`?

I've got plans to refactor that system code out into `commands/build.rb` to stop leakages like this, but that won't happen until most other commands are refactored, and isn't a timely solution to this problem.

uberhacker: This is not simple to Rubyize.  If someone has a better solution, I'm all ears.
```
     search = false
      Dir['**/*'].grep(/config$/).each do |filename|
        File.open(filename) do |file|
          file.each_line do |line|
            if line =~ /\/usr\/bin\/file/  
              search = true
              break
            end
          end
        end
      end
      if search
        puts 'Using filefix.'.orange
        system 'filefix'
      end
```
It does no harm to run filefix on every config file even if `/usr/bin/file` doesn't exist.  The filefix program itself can do the checking and fixing in one go.

Zopolis4: No, I'm simply saying is there a way to run the default `system()` command rather than the one from `lib/package.rb`

Zopolis4: What package did you encounter the bug on? I'd like to test `Kernel.system()`.

uberhacker: I was trying to update a2ps.rb.

Zopolis4: For what its worth, I'm working on that as well: https://github.com/Zopolis4/chromebrew/tree/0a. Might be best to avoid doubling up. I should PR my branches sooner :sweat_smile:.

Zopolis4: Tested, using `Kernel.system()` works.

lib/buildsystems/autotools.rb
```diff
@@ -19,7 +19,7 @@ def self.build
...
+      if File.executable? "#{CREW_PREFIX}/bin/file"
```
Zopolis4, line 22: `if Kernel.system('grep -q /usr/bin/file configure')`?

Zopolis4 (changes requested):

lib/buildsystems/autotools.rb
satmandu: > Tested, using `Kernel.system()` works.

Yeah I noticed that too. I mentioned it to @uberhacker and as I understand it he's going to update the PR accordingly.

satmandu: We should probably figure out though why `system()` is broken though. 😅

Zopolis4: As I said, its because we define a custom `system()` in `lib/package.rb` that restarts with `make -j1` on errors instead of failing.

satmandu: Yes. We should fix the custom `system()`...

That's the root cause of this issue.

Zopolis4: The custom `system()` is working as intended, that's what its supposed to do. Although, it might be nice to find a way to isolate that to package files only, rather than having to use `Kernel.system()` to override it.

Zopolis4 (approved):

PR #10014, opened 2024-06-23
Blender 4.0.2 -> 4.1.1

### LINKED ISSUE #9626: 🐞 Bug Blender giving error about libtbb.so.2
**Describe the bug**
Running `blender` gives me `blender: error while loading shared libraries: libtbb.so.2: ELF load command address/offset not page-aligned`

**To Reproduce**

1. I executed this command:

      ```shell
      blender
      ```

      Output:

      ```
      blender: error while loading shared libraries: libtbb.so.2: ELF load command address/offset not page-aligned
      ```

**Expected behavior**
I expect Blender to run without any issues

**Environment: Paste the output of `crew sysinfo -v` between the two `***` below**

***

- User space: `64-bit`
- Chromebrew Kernel version: `5.15`

- Chromebrew version: `1.46.6`

- Last update in local repository: f2fae48a7 `Rsync 3.2.7 => 3.3.0 (#9623) (7 hours ago)`

- OS variant: `Chrome OS`
- OS version: `octopus-release/R122-15753.55.0`

</details>
***

**Screenshots**

***
the error
***

**Additional context**
None

      - [ ] N/A

Zopolis4:
This took quite some time. Fixes #9626.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=biversio crew update
```


packages/blender.rb
```diff
@@ -1,45 +1,66 @@
...
+  depends_on 'zlibpkg'
+  depends_on 'zstd'
 
-  def self.install
-    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/bin"
-    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/share/blender"
-    FileUtils.mv Dir['*'], "#{CREW_DEST_PREFIX}/share/blender"
-    FileUtils.ln_s "#{CREW_PREFIX}/share/blender/blender", "#{CREW_DEST_PREFIX}/bin/blender"
+  def self.patch
+    # These fixes haven't made their way to to a release yet, so we just patch them in.
```
uberhacker, line 53: 2 tos?

packages/openimageio.rb
```diff
@@ -0,0 +1,35 @@
+require 'buildsystems/cmake'
+
+class Openimageio < CMake
+  description 'OpenImageIO is a toolset for reading, writing, and manipulating image files of any image file format relevant to VFX / animation via a format-agnostic API with a feature set, scalability, and robustness needed for feature film production.'
+  homepage 'https://github.com/AcademySoftwareFoundation/OpenImageIO'
+  version '2.5.12.0'
```
uberhacker, line 6: Use `-py3.12` since it depends on a python version?

packages/openimageio.rb
```diff
@@ -0,0 +1,35 @@
...
+  depends_on 'libpng'
+  depends_on 'libtiff'
+  depends_on 'libwebp' # R
+  depends_on 'openexr'
+  depends_on 'pugixml'
+  depends_on 'py3_pybind11'
+  depends_on 'zlibpkg' # R
+
+  # The setup for finding pybind11 is custom, so it's hard to tell where exactly it breaks to fix it properly.
+  cmake_options "-Dpybind11_ROOT=#{CREW_PREFIX}/lib/python3.12/site-packages/pybind11 -DOIIO_BUILD_TESTS=OFF"
```
uberhacker, line 34: We need to be aware of the python version here for future updates.  Ideally, it would be variable based on `version`.

packages/py3_pybind11.rb
```diff
@@ -0,0 +1,15 @@
+require 'buildsystems/pip'
+
+class Py3_pybind11 < Pip
+  description 'Seamless operability between C++11 and Python'
+  homepage 'https://github.com/pybind/pybind11'
+  @_ver = '2.12.0'
+  version "#{@_ver}-py3.12"
```
uberhacker, line 7: `@_ver` is unused here so we could just set `version '2.12.0-py3.12'`.

packages/fftw.rb
```diff
@@ -1,37 +1,25 @@
...
+class Fftw < Autotools
+  description 'FFTW is a C subroutine library for computing the discrete Fourier transform (DFT) in one or more dimensions, of arbitrary input size, and of both real and complex data (as well as of even/odd data, i.e. the discrete cosine/sine transforms or DCT/DST).'
+  homepage 'https://www.fftw.org/'
+  version '3.3.10'
   license 'GPL-2+'
   compatibility 'all'
-  source_url "http://www.fftw.org/fftw-#{version}.tar.gz"
-  source_sha256 'bf2c7ce40b04ae811af714deb512510cc2c17b9ab9d6ddcf49fe4487eea7af3d'
-  binary_compression 'tar.xz'
+  source_url 'https://fftw.org/fftw-3.3.10.tar.gz'
```
uberhacker, line 9: Keeping `#{version}` makes it a bit easier to do updates, actually.

uberhacker (changes requested):
Ambitious!

packages/openimageio.rb
uberhacker: Maybe do something like `#{version.split('y').last}` to get `3.12`?

packages/fftw.rb
uberhacker: That said, if we include `-py3.12` in the version then we would need to revert back to `@_ver` or use `#{version.split('-').first}`.

packages/blender.rb
```diff
@@ -1,45 +1,66 @@
...
+    system 'patch -Np1 -i 861536b24c866d09857ea4b92717f822da4bc75e.patch'
   end
 
-  def self.postinstall
-    ExitMessage.add "Type 'blender' to get started.".lightblue
-  end
+  # Blender needs the single-precision build of fftw, which we do not provide.
+  # WITH_SYSTEM_AUDASPACE is broken in 4.1.1, should be resolved in 4.2
+  # WITH_INSTALL_PORTABLE needs to be disabled to get a a FHS-compatible install.
+  cmake_options '-DWITH_FFTW3=OFF -DWITH_SYSTEM_EIGEN3=ON -DPYTHON_VERSION=3.12 -DWITH_PYTHON_INSTALL=OFF -DWITH_INSTALL_PORTABLE=OFF'
```
uberhacker, line 65: I just noticed we might want to do the `#{version.split('y').last}` trick here too.  That would mean `version '4.1.1-py3.12'` for this package.

Zopolis4: Blender doesn't always require a hard-defined python version, as it does officially support python 3.11-- Once it upgrades to 3.12, there is no need for this define.

So, my thinking is that it makes sense for this version to be manually updated, so it can be removed if unneeded.

uberhacker: I guess that reasoning makes sense but we'll just need to remember to bump the version on the next release, if available.  Not a simple thing to remember.

uberhacker (approved):

PR #10032, opened 2024-06-26
Slack_cli 2.22.0 => 2.26.0

uberhacker:


Zopolis4 (approved):
Interesting how the ostensibly 64-bit version is working on `i686`, but if you've tested it that's good enough for me.

PR #10044, opened 2024-06-28
Packer 1.7.10 => 1.11.0

uberhacker:

### DIFF
--- a/packages/packer.rb
+++ b/packages/packer.rb
@@ -3,26 +3,26 @@ require 'package'
 class Packer < Package
   description 'Packer is an open source tool for creating identical machine images for multiple platforms from a single source configuration.'
   homepage 'https://www.packer.io/'
-  version '1.7.10'
+  version '1.11.0'
   license 'Apache-2.0, BSD-2, BSD-4, MIT, MPL-2.0 and unicode'
   compatibility 'all'
   source_url({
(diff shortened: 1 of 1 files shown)


packages/packer.rb
```diff
@@ -3,26 +3,26 @@
...
-       i686: '9ebe8ce5b7e89c1caf645327ec654dc5de6c27ecaebcc6c536e07cc6cd1052e1',
-     x86_64: '1c8c176dd30f3b9ec3b418f8cb37822261ccebdaf0b01d9b8abf60213d1205cb'
+    aarch64: '644ff27386993ef09b29169c383b70c0c92aef6bce53bdd9f87b945c4a6a31bb',
+     armv7l: '644ff27386993ef09b29169c383b70c0c92aef6bce53bdd9f87b945c4a6a31bb',
+       i686: 'b3f51d82341c155c520d530178ffdf5cfe48cd001cf9a760ece11e7813132ca0',
+     x86_64: 'dcac06a4c671bbb71e916da5abe947ebf4d6aa35c197e21c7c7b1d68cb8b7cad'
   })
 
   no_compile_needed
+  no_shrink
```
Zopolis4, line 23: Does the shrinking break things? If so, we should file a bug report.

uberhacker: No, it just takes forever to install for little benefit.

satmandu (approved):

PR #10076, opened 2024-07-04
fpc 3.2.2 -> 3.2.3

### LINKED ISSUE #7593: fpc (free pascal compiler) with error, possibly the file system permission problem
I have a Ace C720p with chromebook flex developer mode enabled, inside developer mode installed chromebrew, then

crew install fpc

smoothly.  When I run "fpc", the message is

Error: ppcx64 can't be executed, error message: Failed to execute "ppcx64", error code: 127

I ever tried g++ with "HelloWorld" code. The compiled output "a.out" is not executable. I have to remount the directory

sudo mount -o remount,rw -o exec /home/chronos/user 

So I can run "a.out". 

So I guess whether the file system causes the fpc problem?

Any ideas? Thanks a lot.

Zopolis4:
This took me the better part of two days to figure out, and then bootstrapping wasn't particularly quick either. 
(Once an upstream issue is resolved, things  should be a bit cleaner, as well what looks to be an issue with our docker images.)

At any rate, we now have a working `fpc` on all platforms, which should make the next update significantly easier.

Fixes #7593.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=sidel crew update
```


packages/fpc.rb
```diff
@@ -3,56 +3,48 @@
...
-    config_dirs.each do |config_dir|
-      case response
-      when 'y', 'yes'
-        FileUtils.rm_rf config_dir
-        puts "#{config_dir} removed.".lightred
-      else
-        puts "#{config_dir} saved.".lightgreen
-      end
-    end
+    ExitMessage.add "Type 'fp' to start.".lightblue
```
uberhacker, line 48: We still want the user to have the option to remove the config directories if they still exist in this release.  If you're like me, I absolutely hate when I uninstall a package and it leaves residue behind.

packages/fpc.rb
```diff
@@ -3,56 +3,48 @@
...
+
+  def self.build
+    # https://gitlab.com/freepascal.org/fpc/source/-/issues/40841
+    # fpc isn't fooled by our docker emulation, and thus disables reading from the config because it thinks we're cross-compiling.
+    # See https://gitlab.com/freepascal.org/fpc/source/-/commit/898a4debcab843b2f6f5c29b795a517485582dfb
+    # Because we don't read from the config, we don't get the correct library paths.
+    # The correct thing to do here would be to find the fault in our docker setup that reveals we're running under emulation, and fix it.
+    # In the meantime, though, just pass the correct library path manually.
+    # Chromebrew messes things up if we don't pass this in one string, adding single quotes around OPT which break things.
+    system "make OVERRIDEVERSIONCHECK=1 OPT='#{"-Fl#{CREW_LIB_PREFIX}" unless ARCH.eql?('x86_64')} -k-L#{CREW_LIB_PREFIX}' all"
```
uberhacker, line 30: The nested double quotes should be escaped or removed altogether.  That might be your issue here.

packages/fpc.rb
Zopolis4: They are in `CREW_PREFIX/etc` now, so they'll be handled by crew remove.

packages/fpc.rb
Zopolis4: The nested double quotes work, so I'm not entirely sure what you mean?

uberhacker: The system command will only run `system "make OVERRIDEVERSIONCHECK=1 OPT='#{"` and ignore the rest of the line unless you escape it.  That's what I mean.

Zopolis4: This does not happen, though, and I was able to build the package successfully.

uberhacker: I wouldn't know how that is possible.

Zopolis4: If I had to guess, its probably because the quotes are inside a `#{}`, so only the result of that segment of code is then substituted in, and the result of that code does not include any double quotes.

satmandu:
Does this need rebuilding on an older container?

Zopolis4:
I believe so.

Zopolis4:
Rebuilt.

uberhacker (approved):

PR #10078, opened 2024-07-04
shrinkpdf: update homepage and fetch urls, fix fetch, update to 1.1

### LINKED ISSUE #10077: 🐞 Bug shrinkpdf URLs broken
**Describe the bug**
Just noticed that in https://github.com/chromebrew/chromebrew/blob/master/packages/shrinkpdf.rb both source_url and url fetched by curl are non existing, so the package is probably broken. You might want to switch to fetch from github.

**To Reproduce**
Not tried, since not using chromebrew.

AMDmi3:
Fixes #10077, but this needs to be tested first and checksums need to be updated.

### DIFF
--- a/packages/shrinkpdf.rb
+++ b/packages/shrinkpdf.rb
@@ -6,18 +6,14 @@ class Shrinkpdf < Package
   version '1.1'
   license 'BSD-3'
   compatibility 'x86_64 aarch64 armv7l'
-  source_url 'https://raw.githubusercontent.com/aklomp/shrinkpdf/v#{version}/shrinkpdf.sh'
-  source_sha256 'bd23c0917741a6a63694bcb99315c4afabdac0d24fef2be896c5be5c7b3aed3c'
-
-  binary_sha256({
-    aarch64: '?',
-     armv7l: '?',
-     x86_64: '?'
-  })
+  source_url 'https://github.com/aklomp/shrinkpdf.git'
+  git_hashtag "v#{version}"
 
   depends_on 'ghostscript'
 
+  no_compile_needed
+
   def self.install
-    system "install -Dm755 shrinkpdf.sh #{CREW_DEST_PREFIX}/bin/shrinkpdf"
+    FileUtils.install 'shrinkpdf.sh', "#{CREW_DEST_PREFIX}/bin/shrinkpdf", mode: 0o755
   end
 end
--- a/packages/shrinkpdf.rb
+++ b/packages/shrinkpdf.rb
@@ -2,25 +2,22 @@ require 'package'
 
 class Shrinkpdf < Package
   description 'A simple wrapper around Ghostscript to shrink PDFs (as in reduce filesize) under Linux.'
-  homepage 'http://alfredklomp.com/programming/shrinkpdf/'
-  version '2014'
+  homepage 'https://github.com/aklomp/shrinkpdf'
+  version '1.1'
   license 'BSD-3'
   compatibility 'x86_64 aarch64 armv7l'
-  source_url 'http://alfredklomp.com/programming/shrinkpdf/index.html'
-  source_sha256 'cdc74b28d8ad757fa2cc9d8782031fe2325a17d3f4fd6080c05ca62790e5c690'
-  binary_compression 'tar.xz'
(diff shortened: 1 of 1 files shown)


packages/shrinkpdf.rb
```diff
@@ -2,25 +2,22 @@
...
+  source_sha256 'bd23c0917741a6a63694bcb99315c4afabdac0d24fef2be896c5be5c7b3aed3c'
 
   binary_sha256({
-    aarch64: '19c53e4be836957a64613eb60e329506cfafde2891c54155b99ba16de0855ba6',
-     armv7l: '19c53e4be836957a64613eb60e329506cfafde2891c54155b99ba16de0855ba6',
-     x86_64: '92e7bddf8ead1324927547c1591e7f6192c2b195a8137a3365597c08d802f950'
+    aarch64: '?',
+     armv7l: '?',
+     x86_64: '?'
   })
```
uberhacker, line 16: This bash executable is small so we don't really need binaries.  Probably can remove the whole section.  That said, you can replace the binary section with `no_compile_needed`.  Typically, we put these "boolean properties" below any depends_on lines.

packages/shrinkpdf.rb
```diff
@@ -2,25 +2,22 @@
...
   description 'A simple wrapper around Ghostscript to shrink PDFs (as in reduce filesize) under Linux.'
-  homepage 'http://alfredklomp.com/programming/shrinkpdf/'
-  version '2014'
+  homepage 'https://github.com/aklomp/shrinkpdf'
+  version '1.1'
   license 'BSD-3'
   compatibility 'x86_64 aarch64 armv7l'
-  source_url 'http://alfredklomp.com/programming/shrinkpdf/index.html'
-  source_sha256 'cdc74b28d8ad757fa2cc9d8782031fe2325a17d3f4fd6080c05ca62790e5c690'
-  binary_compression 'tar.xz'
```
satmandu, line 11: You can change this to `binary_compression 'tar.zst'`

packages/shrinkpdf.rb
```diff
@@ -2,25 +2,22 @@
...
+     armv7l: '?',
+     x86_64: '?'
   })
 
   depends_on 'ghostscript'
 
   def self.install
-    system 'curl -#LO http://alfredklomp.com/programming/shrinkpdf/shrinkpdf.sh'
-    abort 'Checksum mismatch. :/ Try again.'.lightred unless Digest::SHA256.hexdigest(File.read('shrinkpdf.sh')) == 'bf313e47e81ce4c088704840554f2f643597db1b3d6937a0af5ed43368b1ccc7'
     system "install -Dm755 shrinkpdf.sh #{CREW_DEST_PREFIX}/bin/shrinkpdf"
```
uberhacker, line 21: Replace with?
```
FileUtils.install 'shrinkpdf.sh', "#{CREW_DEST_PREFIX}/bin/shrinkpdf", mode: 0o755
```

satmandu, line 21: This can be changed to use `FileUtils.install`.

Just grep the other package files install sections for the usage...

satmandu:
And thanks for the update!

packages/shrinkpdf.rb
uberhacker: The updated package with the least amount of maintenance going forward is below:
```
require 'package'

class Shrinkpdf < Package
  description 'A simple wrapper around Ghostscript to shrink PDFs (as in reduce filesize) under Linux.'
  homepage 'https://github.com/aklomp/shrinkpdf'
  version '1.1'
  license 'BSD-3'
  compatibility 'x86_64 aarch64 armv7l'
  source_url 'https://github.com/aklomp/shrinkpdf.git'
  git_hashtag "v#{version}"

  depends_on 'ghostscript'

  no_compile_needed

  def self.install
    FileUtils.install 'shrinkpdf.sh', "#{CREW_DEST_PREFIX}/bin/shrinkpdf", mode: 0o755
  end
end
```

AMDmi3:
Should I update the PR with @uberhacker version? IMO it's not quite fair.

uberhacker:
> Should I update the PR with @uberhacker version? IMO it's not quite fair.

Yes, I will allow you to get full credit for making my proposed changes.

AMDmi3:
Updated

uberhacker (approved):

PR #10082, opened 2024-07-04
Add quick armv7l support to unit testing CI

Zopolis4:
And I do mean quick-- there a number of improvements to be made to this action, but most of them are waiting on #9707, so lets just get more coverage for now.

### DIFF
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -11,9 +11,30 @@ jobs:
         shell: bash
         env:
           GITHUB_CONTEXT: ${{ toJson(github) }}
-      - name: Unit Tests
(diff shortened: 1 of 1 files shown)


Zopolis4:
quick as in how long it took to implement, not how long it takes to run now... :sweat_smile:

uberhacker:
This is great but shouldn't the priority be to fix the broken arm build process first?  A recent update broke it and I will submit a revert PR if a fix is not implemented soon.

Zopolis4:
working on that as well

satmandu (approved):

satmandu:
Thank you for putting this together! This should hopefully prevent breakage in the future...

PR #10086, opened 2024-07-05
fix building armv7l binaries inside armv8l container regression

### LINKED ISSUE #10081: 🐞 Building packages on arm is broken.
```
chronos@rpi51-armv7l /output/git/satmandu (mesa *)$ crew build -f packages/mesa.rb
/usr/local/lib/crew/lib/const.rb:174:in `<top (required)>': uninitialized constant CREW_ARCH_FLAGS (NameError)

CREW_CORE_FLAGS           = "-O2 -pipe -ffat-lto-objects -fPIC #{CREW_ARCH_FLAGS} -fuse-ld=#{CREW_LINKER} #{CREW_LINKER_FLAGS}"
                                                                 ^^^^^^^^^^^^^^^
Did you mean?  CREW_USE_CURL
        from /usr/local/lib/crew/commands/const.rb:1:in `require_relative'
        from /usr/local/lib/crew/commands/const.rb:1:in `<top (required)>'
        from /usr/local/bin/crew:15:in `require_relative'
        from /usr/local/bin/crew:15:in `<main>'
chronos@rpi51-armv7l /output/git/satmandu (mesa *)$
```

satmandu:
Fixes #10081 

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=rpi_build_fix crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -2,7 +2,7 @@
 # Defines common constants used in different parts of crew
 require 'etc'
 
-CREW_VERSION = '1.49.3'
+CREW_VERSION = '1.49.4'
 
 # Kernel architecture.
 KERN_ARCH = Etc.uname[:machine]
@@ -10,10 +10,17 @@ KERN_ARCH = Etc.uname[:machine]
 # Read and parse processor information from /proc/cpuinfo
 CPUINFO = File.readlines('/proc/cpuinfo').map { |line| line.chomp.split(/\t+: /) if line.include?("\t") }.compact.to_h
 
-# We report aarch64 machines as armv7l for now, as we treat them as if they were armv7l.
+# We report aarch64 & armv8l machines as armv7l for now, as we treat
+# them as if they were armv7l.
 # When we have proper aarch64 support, remove this.
 # Also, we allow ARCH to be changed via the ARCH environment variable.
-ARCH = KERN_ARCH.eql?('aarch64') ? 'armv7l' : ENV.fetch('ARCH', KERN_ARCH)
+ARCH = ENV.fetch('ARCH') do |_name|
+  case KERN_ARCH
+  when 'aarch64', 'armv8l'
+    'armv7l'
+  else KERN_ARCH
+  end
+end
 
 # Allow for edge case of i686 install on a x86_64 host before linux32 is
 # downloaded, e.g. in a docker container.


Zopolis4:
Again, I think it makes more sense for a workaround for your specific hardware setup to be something in your configuration, rather than in chromebrew itself.

Zopolis4:
That is, after all, what the ability to override the architecture via an environment variable is for. So specific workarounds don't need to be in chromebrew, and instead you can just pass `-e ARCH=armv7l` to your personal docker invocation.

lib/const.rb
```diff
@@ -2,18 +2,25 @@
...
+CREW_VERSION = '1.49.4'
 
 # Kernel architecture.
 KERN_ARCH = Etc.uname[:machine]
 
 # Read and parse processor information from /proc/cpuinfo
 CPUINFO = File.readlines('/proc/cpuinfo').map { |line| line.chomp.split(/\t+: /) if line.include?("\t") }.compact.to_h
 
-# We report aarch64 machines as armv7l for now, as we treat them as if they were armv7l.
+# We report aarch64 & armv8l machines as armv7l for now, as we treat
```
Zopolis4, line 13: There are no chromeos `armv8l` machines, so at the very least mention that this is a container-specific workaround when running the containers on certain hardware.

Zopolis4 (changes requested):

lib/const.rb
uberhacker: armv8l is just another way to say aarch64.

Zopolis4: No, it is not. No chromeos machines report `armv8l`.

uberhacker (approved):
Note: We need to build binaries inside containers so let's not remove this logic just because it's not related to ChromeOS devices.

uberhacker (approved):

Zopolis4:
We're a ChromeOS project...

uberhacker:
> We're a ChromeOS project...

You're not understanding my point.

uberhacker (approved):

Zopolis4:
Evidently not. My logic is that we shouldn't have a workaround specific to satmandu's hardware in chromebrew, and that it makes more sense to have the workaround as something in the dockerfile or satmandu's docker invocation.

Zopolis4:
Also, please stop marking my conversation as resolved.

lib/const.rb
uberhacker: The point is Docker containers do report armv8l.  We can't just remove logic that isn't related to ChromeOS devices.

uberhacker (approved):

lib/const.rb
Zopolis4: Docker containers don't report it, raspberry pi's report it. 

I don't think this logic should be located as part of crew, i think it makes more sense to have it as part of docker, because it only affects docker containers.

uberhacker:
> > You're not understanding my point.
> 
> Evidently not. My logic is that we shouldn't have a workaround specific to satmandu's hardware in chromebrew, and that it makes more sense to have the workaround as something in the dockerfile or satmandu's docker invocation.
> 
> Why do you disagree with this?

Listen.  This is not about @satmandu's hardware.  This is about the ability to run ChromeOS in Docker containers that report armv8l for the architecture.  There is NO DIFFERENCE between armv8l and aarch64.  They both mean 64-bit arm architecture.  Just a different way to saying the same thing.

Zopolis4:
This is literally about satmandu's hardware. The rasberry pi is reporting armv8l, which is passed to docker.

All I'm saying is that this workaround shouldn't be in `lib/const.rb`. It's an issue that only affects containers running on armv8l hardware. So, this workaround should only be for that setup.

i.e. passing `-e ARCH=armv7l` to the docker container.

uberhacker:
> This is literally about satmandu's hardware. The rasberry pi is reporting armv8l, which is passed to docker.
> 
> All I'm saying is that this workaround shouldn't be in `lib/const.rb`. It's an issue that only affects containers running on armv8l hardware. So, this workaround should only be for that setup.
> 
> i.e. passing `-e ARCH=armv7l` to the docker container.

I'm not understanding why it should be a big deal to leave this logic here.  What if someone else like you or me want to build on a Raspberry Pi device?  Are you saying we all need a special config outside of crew to do this?

Zopolis4:
> I'm not understanding why it should be a big deal to leave this logic here.

A workaround specific to one configuration should not be in chromebrew. That is bad practice.

> Are you saying we all need a special config outside of crew to do this?

I'm proposing one of two solutions:
- A workaround in the dockerfile to export `ARCH=armv7l` on armv7l containers. This means that no configuration is required, and everything works without needing specific logic in chromebrew.
- Alternatively, we just use the existing workaround plumbing for this. If your specific setup reports the wrong architecture, override it with an environment variable.

Zopolis4:
Actually, the dockerfile already has code for the first solution:
```dockerfile
ENV CREW_KERNEL_VERSION=$CREW_KERNEL_VERSION
ENV ARCH=$ARCH
ENV PATH=/usr/local/bin:/usr/local/sbin:/usr/bin:/bin
```

I'll try and fix that

uberhacker:
> > I'm not understanding why it should be a big deal to leave this logic here.
> 
> A workaround specific to one configuration should not be in chromebrew. That is bad practice.
> 
> > Are you saying we all need a special config outside of crew to do this?
> 
> I'm proposing one of two solutions:
> 
> * A workaround in the dockerfile to export `ARCH=armv7l` on armv7l containers. This means that no configuration is required, and everything works without needing specific logic in chromebrew.
> * Alternatively, we just use the existing workaround plumbing for this. If your specific setup reports the wrong architecture, override it with an environment variable.

Bad practice would be overriding the "wrong" architecture with an environment variable setting.  Again, there is no difference between armv8l and aarch64.  They both mean 64-bit arm.

Zopolis4:
> Bad practice would be overriding the "wrong" architecture with an environment variable setting.

As opposed to overriding it in chromebrew? I really don't understand what you mean here. We're overriding it either way, I'm just asking if we could override it in the dockerfile, which is already trying to do that.

uberhacker:
> > Bad practice would be overriding the "wrong" architecture with an environment variable setting.
> 
> As opposed to overriding it in chromebrew? I really don't understand what you mean here. We're overriding it either way, I'm just asking if we could override it in the dockerfile, which is already trying to do that.

You have it in your mind that `armv8l` is wrong.  Think of it as another way of saying aarch64.

Zopolis4:
> Bad practice would be overriding the "wrong" architecture with an environment variable setting.

We already do this?

Every single command in the dockerfile is prefixed with an environment variable overriding the architecture. 

It's just not currently exported to the lifetime of the container, which is why this issue is happening.

Zopolis4:
I have attached a patch that should fix the issue in the dockerfile.

```diff
diff --git a/chromeos_docker37 b/chromeos_docker37
index 4dbd381..cb32e26 100644
--- a/chromeos_docker37
+++ b/chromeos_docker37
@@ -186,6 +186,7 @@ chown chronos:chronos /home/chronos/user/.profile
 tee -a /home/chronos/user/.bashrc <<TEEBASHRCEOF
 echo "This is the .bashrc"
 set -a
+ARCH=\\\$ARCH
 QEMU_CPU=max
 CFLAGS="-march=$MARCH"
 : "\\\${LANG:=en_US.UTF-8}"
@@ -359,7 +360,7 @@ build_docker_image () {
   $buildx_cmd  || echo "Docker Build Error."
 }
 make_docker_image_script () {
-  dockercmd="docker run --init --platform ${PLATFORM} --rm --net=host \${PAGER_PASSTHROUGH} \${TZ_PASSTHROUGH} \${X11} -e LOCALRC=\"\${LOCALRC}\" -v \$(pwd)/pkg_cache:/usr/local/tmp/packages -v \$(pwd):/output -h \$(hostname)-${ARCH} -it ${REPOSITORY}/crewbuild:${name}-${ARCH}.m${milestone}"
+  dockercmd="docker run --init --platform ${PLATFORM} --rm --net=host \${PAGER_PASSTHROUGH} \${TZ_PASSTHROUGH} \${X11} -e LOCALRC=\"\${LOCALRC}\" -e ARCH=${ARCH} -v \$(pwd)/pkg_cache:/usr/local/tmp/packages -v \$(pwd):/output -h \$(hostname)-${ARCH} -it ${REPOSITORY}/crewbuild:${name}-${ARCH}.m${milestone}"
   [[ -f $(abspath "${outdir}")/crewbuild-${name}-${ARCH}.m${milestone}.sh ]] && rm $(abspath "${outdir}")/crewbuild-"${name}"-"${ARCH}".m"${milestone}".sh
   cat <<IMAGESCRIPTEOF > $(abspath "${outdir}")/crewbuild-"${name}"-"${ARCH}".m"${milestone}".sh
 #!/bin/bash
```

satmandu:
It's entirely reasonable to have philosophical disagreements about how things should work.

We just have to sometimes accept that people may disagree with us on some things, and that we have to find compromises.

My argument is that running `armv7l` binaries on systems booted into `aarch64` kernels that also support `armv7l` binaries will sometimes expose the architecture via `uname -m` as the 32-bit superset of `armv7l` — `armv8l`.

So we should support that condition. The raspberry pi [3,4,5] when booted into the non-16k page size 64-bit kernels also shows that behavior.

Also, I believe that we should support running the containers we create without external twiddling (through the use of environment variables) for basic functionality.

The containers work fine without any environment variables at all right now.

We shouldn't have to add an environment variable to get minimal functionality.

I believe this PR falls under that rationale, as this fixes a regression (that I merged) that broke that functionality.

satmandu:
Also, to reiterate, there's a longstanding policy in the Linux kernel dev community to never break user space.

This is sort of like that ... Don't break existing functionality with changes.

Zopolis4:
Fair enough. What you've said makes sense, and I'm happy to support this configuration-- that being said, as this only occurs in containers, could we put the fix in the containers?

> Also, I believe that we should support running the containers we create without external twiddling (through the use of environment variables) for basic functionality.

That's fair enough-- if we have the arch exported in the dockerfile, it won't require any configuration or runtime arguments to work.

> The containers work fine without any environment variables at all right now.

I thought the reason you added that override in #9459 was because they didnt?

> We shouldn't have to add an environment variable to get minimal functionality.

I think an environment variable set in a container is preferable to container workarounds being in general code.

Zopolis4:
> This is sort of like that ... Don't break existing functionality with changes.

Sorry :sweat_smile: I didn't realize that anything was still reporting `armv8l`.

uberhacker (approved):

PR #10091, opened 2024-07-07
libxml2 2.12.7-icu74.2 -> 2.13.2,  py3_libxml2 2.12.7-py3.12 -> 2.13.2-py3.12

Zopolis4:
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=onetree crew update
```


packages/libxml2.rb
```diff
@@ -2,39 +2,27 @@
...
+    aarch64: 'eb69071220e11d4f0fd61895b6c46e4be71ec34a0c8a33ff991180437c83f2b9',
+     armv7l: 'eb69071220e11d4f0fd61895b6c46e4be71ec34a0c8a33ff991180437c83f2b9',
+       i686: '199c9ec45327f3be3f402f08023fea2cd9dc1f04ca65e5eac0a2018a26dfd41b',
+     x86_64: '91e68d3cf4eb408735aca931cbe23d7c52dc5f9d3e9084255910d203443507fe'
   })
 
   depends_on 'gcc_lib' # R
   depends_on 'glibc' # R
-  depends_on 'icu4c' # R
+  depends_on 'icu4c'
```
satmandu, line 22: Not a runtime dep?

packages/py3_libxml2.rb
```diff
@@ -1,28 +1,26 @@
...
+    aarch64: '6248e0710ebec3690a92a5a762776e8aaf31d4f4266d6298c565532d6fb92678',
+     armv7l: '6248e0710ebec3690a92a5a762776e8aaf31d4f4266d6298c565532d6fb92678',
+       i686: 'f8e328879434dd77a4094b64cbd95e816d0b2d4ce6b639ac51c2f531873724df',
+     x86_64: '6e1e839c0118536a04baf16153d14adc8b07f09058fb6dacc21dc364215541ec'
   })
 
   depends_on 'glibc' # R
-  depends_on 'libxml2' # R
-  depends_on 'libxslt' => :build
+  depends_on 'libxml2'
```
satmandu, line 21: This is a runtime dep, no?

packages/py3_libxml2.rb
```diff
@@ -1,28 +1,26 @@
...
-     x86_64: '456edd3f290f759a30b49e875e8e40bea9351702a07170c55329ed1c19cbb3f7'
+    aarch64: '6248e0710ebec3690a92a5a762776e8aaf31d4f4266d6298c565532d6fb92678',
+     armv7l: '6248e0710ebec3690a92a5a762776e8aaf31d4f4266d6298c565532d6fb92678',
+       i686: 'f8e328879434dd77a4094b64cbd95e816d0b2d4ce6b639ac51c2f531873724df',
+     x86_64: '6e1e839c0118536a04baf16153d14adc8b07f09058fb6dacc21dc364215541ec'
   })
 
   depends_on 'glibc' # R
-  depends_on 'libxml2' # R
-  depends_on 'libxslt' => :build
```
satmandu, line 23: Not a build dep?

Zopolis4: Not a dependency at all.

packages/libxml2.rb
Zopolis4: I thought the '# R' represented that the dependency was set using getrealdeps.rb-- given that we can confirm this is a dependency without doing that, I removed the comment.

satmandu: `# R` was for being added for runtime deps, which the getrealdeps.rb was adding automatically.

It doesn't catch all of them though, and it does help distinguish deps added before we started using that script... I would keep the `# R` on deps that are runtime deps... Otherwise I was mostly marking them as ` # L` logical steps which aren't detected by the script.

Zopolis4: Aren't runtime/build deps already delineated by '=> :build'? 

Not too fussed either way, just wondering about the exact implications of '# R' so we can be on the same page here.

satmandu: Yes but before the script was used runtime deps were added manually, and because builds weren't done in clean containers for every build (amazing that was only a couple of years ago), we had no idea of knowing if the deps list was complete. So adding the # R or # L (separately from build only deps ) is a way to know that the dep list has been recently vetted.

Zopolis4: Fair enough.

satmandu (approved):

PR #10092, opened 2024-07-07
Add smplayer package

uberhacker:
SMPlayer is a free media player for Windows and Linux with built-in codecs that can play virtually all video and audio formats.  This is basically mpv with a graphical frontend but does not interfere with the mpv package since the binaries are isolated from each other.  See https://www.smplayer.info/.  Tested and working on an x86_64 Samsung Chromebook running milestone 126.


packages/smplayer.rb
```diff
@@ -0,0 +1,64 @@
...
+  version '24.5.0'
+  license 'GPL-2'
+  compatibility 'x86_64'
+  source_url "https://github.com/smplayer-dev/smplayer/releases/download/v#{version}/SMPlayer-#{version}-x86_64.AppImage"
+  source_sha256 '85fb5a2322f48a298b7784fc3516e672525593c017ca504310bff05e1330457b'
+
+  no_compile_needed
+
+  depends_on 'gtk3'
+  depends_on 'gdk_base'
```
Zopolis4, line 15: Dependencies in alphabetical order?

satmandu (approved):

packages/smplayer.rb
```diff
@@ -0,0 +1,64 @@
...
+  depends_on 'sommelier'
+
+  def self.build
```
Zopolis4, line 18: Probably move this to the install section if you're setting `no_compile_needed`.

packages/smplayer.rb
```diff
@@ -0,0 +1,64 @@
...
+      If you have trouble playing videos, try the steps below:
+      Select Options > Preferences
+      Click the 'Audio' tab
+      Uncheck 'Use software volume control'
+      Click OK
+    EOM
+  end
+
+  def self.remove
+    config_dir = "#{CREW_PREFIX}/.config/smplayer"
```
Zopolis4, line 52: No need to use a variable if there's only a single value and we don't modify it.

packages/smplayer.rb
uberhacker: Let's not get too anal now. :)

packages/smplayer.rb
uberhacker: This technically is a build step.

tools/packages.yaml
```diff
@@ -8136,6 +8136,11 @@ url: http://kernel.ubuntu.com/~cking/tarballs/smemstat
 activity: medium
 ---
 kind: url
+name: smplayer
+url: https://github.com/smplayer-dev/smplayer/releases
```
Zopolis4, line 8140: I thought we were using `version.rb` instead of `tools/packages.yaml`.

packages/smplayer.rb
uberhacker: No but this section is a pattern for other packages with the same removal needs.  Actually, it might make sense to eventually turn this into a generic function which accepts an array of config directories and abstract it from the package files.

packages/smplayer.rb
Zopolis4: :woman_shrugging: It'll be fixed whenever I write the cop for it, so its not that big of a deal.

tools/packages.yaml
uberhacker: `version.rb` is still not usable for me.  Some package lookups are not the same project and many times I get no results at all.  I did have it working quite well at one point but then SOMEONE decided to change it and make it work not so well after that. 😛

packages/smplayer.rb
Zopolis4: > No but this section is a pattern for other packages with the same removal needs.

Yeah, but a variable only makes sense in the situations where there are multiple config directories. It makes sense for those packages, but we can do without it here.

tools/packages.yaml
Zopolis4: image

I've tested it, and it works for smplayer-- If you still want to add it to `tools/packages.yaml` go ahead i guess

packages/smplayer.rb
uberhacker: Actually, a variable makes a lot of sense in this case since it's reused 4 times.

tools/packages.yaml
uberhacker: I think we need to use multiple sources for package versions.  The way to make it more comprehensive would be to use another source if no results are found with the first repository.

packages/smplayer.rb
Zopolis4: fair enough

packages/smplayer.rb
uberhacker: I get that it makes sense to sort in alphabetical order to identify duplicate lines and that definitely makes sense for packages with a lot of dependencies.

packages/smplayer.rb
Zopolis4: Sorry, I misunderstood the code-- `no_compile_needed` skips building altogether:
image

uberhacker: Right, build means to compile source into binaries.

Zopolis4 (approved):

PR #10115, opened 2024-07-10
Speed up Github Actions workflow unit tests

uberhacker:
We'll have to test this to see if the file path is working correctly.

### DIFF
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -11,6 +11,17 @@ jobs:
         shell: bash
         env:
           GITHUB_CONTEXT: ${{ toJson(github) }}
+      - name: Get all changed package files
+        id: changed-ruby-files
+        uses: tj-actions/changed-files@v44
+        with:
+          files: |
+             packages/*.rb
+      - name: List all changed package files
+        uses: tj-actions/changed-files@v44
+        if: steps.changed-ruby-files.outputs.any_changed == 'true'
+        env:
+          ALL_CHANGED_FILES: ${{ steps.changed-ruby-files.outputs.all_changed_files }}
       - name: Unit Tests (x86_64)
         run: |
             sudo docker run --rm -t satmandu/crewbuild:amd64 sudo -i -u chronos /bin/bash -c "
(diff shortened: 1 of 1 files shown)


.github/workflows/Unit-Test.yml
```diff
@@ -11,6 +11,18 @@ jobs:
         shell: bash
         env:
           GITHUB_CONTEXT: ${{ toJson(github) }}
+        # Only test relevant files that have changed.
+      - name: Get all changed package files
+        id: changed-ruby-files
+        uses: tj-actions/changed-files@v44
```
Zopolis4, line 17: We could do this in Handoff.yml and pass it here as a variable

Zopolis4: I've got something pretty close to that in https://github.com/Zopolis4/chromebrew/blob/ed7412a4c41a65fd534c7914d5da1469a8daa313/.github/workflows/Handoff.yml

uberhacker: I'm fine with using your changes instead of mine.  As long as it speeds it up.  Currently, the tests take 10 min to run.

uberhacker: The unit tests only need to check for changes to the package files and nothing else.

Zopolis4:
I've looked a little more into the speed situation-- obviously the main issue is running under software armv7l emulation, which will be the case for quite a while until github releases arm runners to the public. Beyond that, `buildsystem_test` is taking up the most time by quite a large margin:
```
$ time ./prop_test
real    0m0.653s
user    0m0.601s
sys     0m0.053s
```

```
$ time ./buildsystem_test
real    0m5.844s
user    0m7.356s
sys     0m1.090s
```

This is on an x86_64 container, which is about 5 times faster than the `armv7l` containers under software emulation.

However, the main time is spent in the `hello_world_chromebrew` test-- cloning the repo and building the package takes a significant portion of time. Updating, upgrading, installing and removing vim also seem to contribute.

We could save more time here if we enabled caching-- i.e. set `CREW_CACHE_ENABLED` and then used github action's caching support to cache package downloads and git clones.

Zopolis4 (approved):
I'll move the file change stuff over later, no point holding this change up.

PR #10124, opened 2024-07-11
aide 0.15.1 -> 0.18.8-f445264

Zopolis4:
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=ultimatebattle crew update
```

### DIFF
--- a/manifest/x86_64/a/aide.filelist
+++ b/manifest/x86_64/a/aide.filelist
@@ -1,6 +1,3 @@
 /usr/local/bin/aide
-/usr/local/etc/aide.conf
-/usr/local/etc/doc/aide.db
-/usr/local/man/man5/aide.conf.5.gz
-/usr/local/share/man/man1/aide.1.gz
-/usr/local/share/man/man5/aide.conf.5.gz
+/usr/local/share/man/man1/aide.1.zst
+/usr/local/share/man/man5/aide.conf.5.zst
--- a/packages/aide.rb
+++ b/packages/aide.rb
@@ -1,54 +1,34 @@
-require 'package'
+require 'buildsystems/autotools'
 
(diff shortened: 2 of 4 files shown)


manifest/x86_64/a/aide.filelist
```diff
@@ -1,6 +1,3 @@
 /usr/local/bin/aide
-/usr/local/etc/aide.conf
-/usr/local/etc/doc/aide.db
```
uberhacker, line 3: What happened to the database and config?

Zopolis4: The config was actually a manpage, so shouldn't be installed like that, and the database was an empty file.

uberhacker: What?  This is what makes this thing work.

Zopolis4: `aide.conf.5` is a manual page, not a configuration file. The `doc/aide.conf` file that was present in `0.15.1` is no longer there.

packages/aide.rb
```diff
@@ -1,54 +1,34 @@
...
-    puts
-    puts "The configuration file is located at #{CREW_PREFIX}/etc/aide.conf".lightblue
-    puts
-    puts 'To initialize the database, execute the following:'.lightblue
-    puts "mkdir #{CREW_PREFIX}/etc/src".lightblue
-    puts "cd #{CREW_PREFIX}/etc/doc".lightblue
-    puts 'aide -i'.lightblue
-    puts 'mv aide.db.new aide.db'.lightblue
-    puts
-  end
```
uberhacker, line 53: Take a look at this postinstall message and see if you can get this working with your updates.

PR #10131, opened 2024-07-11
Revert "libpsl 0.21.2 -> 0.21.5"

satmandu:
Reverts chromebrew/chromebrew#10125

Needs to be built on an older container:

```
Initialized empty Git repository in /usr/local/tmp/crew/mesonbuild.20240711173915.dir/.git/
/usr/local/libexec/git-core/git-remote-https: /usr/local/lib64/libc.so.6: version `GLIBC_2.33' not found (required by /usr/local/lib64/libpsl.so.5)
/usr/local/bin/crew:550:in `system': Command failed with exit 128: git (RuntimeError)  
```

### DIFF
--- a/manifest/armv7l/l/libpsl.filelist
+++ b/manifest/armv7l/l/libpsl.filelist
@@ -3,7 +3,7 @@
 /usr/local/include/libpsl.h
 /usr/local/lib/libpsl.so
 /usr/local/lib/libpsl.so.5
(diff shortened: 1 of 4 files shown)


uberhacker (approved):
Yes, good catch.  @Zopolis4: All builds should be done on the lowest glibc version possible unless it doesn't work.

PR #10132, opened 2024-07-11
Update meson, rebuild libpsl

satmandu:
Builds properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=meson crew update
```

### DIFF
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -24,7 +24,7 @@ jobs:
           ALL_CHANGED_FILES: ${{ steps.changed-ruby-files.outputs.all_changed_files }}
       - name: Unit Tests (x86_64)
         run: |
-            sudo docker run --rm -t satmandu/crewbuild:m90-x86_64 sudo -i -u chronos /bin/bash -c "
+            sudo docker run --rm -t satmandu/crewbuild:m90-x86_64 sudo -i -u chronos /usr/local/bin/bash -c "
             echo \"CREW_REPO is ${{ github.event.pull_request.head.repo.clone_url }}\" && \
             echo \"CREW_BRANCH is ${{ github.head_ref }}\" && \
             CREW_REPO=${{ github.event.pull_request.head.repo.clone_url }} CREW_BRANCH=${{ github.head_ref }} crew update && \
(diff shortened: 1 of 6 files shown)


packages/libpsl.rb
```diff
@@ -3,19 +3,19 @@
 class Libpsl < Meson
   description 'C library for the Public Suffix List'
   homepage 'https://github.com/rockdaboot/libpsl'
-  @_ver = '0.21.2'
-  version @_ver.to_s
+  @_ver = '0.21.5'
+  version "#{@_ver}-1"
```
uberhacker, line 7: How about just simply the following?
```
version '0.21.5'
...
git_hashtag version
```

satmandu: This is replacing existing "0.21.5" binaries...

satmandu: Ok that could work...

satmandu: Done.

packages/mesonbuild.rb
```diff
@@ -3,7 +3,7 @@
 class Mesonbuild < Python
   description 'Meson is an open source build system meant to be both extremely fast and user friendly.'
   homepage 'https://mesonbuild.com/'
-  @_ver = '1.4.0'
+  @_ver = '1.5.0'
   version "#{@_ver}-py3.12"
```
uberhacker, line 7:
```
version '1.5.0-py3.12'
...
git_hashtag version.split('-').first
```
?

satmandu: Done.

.github/workflows/Unit-Test.yml
```diff
@@ -57,7 +57,7 @@ jobs:
             ruby ../tests/commands/const.rb && \
             ruby ../tests/commands/help.rb && \
             ruby ../tests/commands/prop.rb && \
-            ( [[ -z "${ALL_CHANGED_FILES}" ]] && ( for file in ${ALL_CHANGED_FILES}; do
+            ( [[ -v ALL_CHANGED_FILES ]] && ( for file in ${ALL_CHANGED_FILES}; do
```
uberhacker, line 60: Don't we want `[[ -n "${ALL_CHANGED_FILES}" ]] && ...`?  We're checking if there are files, correct?

satmandu: The variable doesn't exist if it is empty.

uberhacker:
```
    String operators:
    
      -z STRING      True if string is empty.
    
      -n STRING
         STRING      True if string is not empty.
```

uberhacker: We can try your way but I know that `-n` definitely should work.

uberhacker (approved):

PR #10152, opened 2024-07-15
Download the correct file when there are multiple versions of a package binary.

Zopolis4:
Tested and working locally, and extended the unit tests to cover this.

Combined with the changes I made to `crew update` a few years ago, this should allow us to rebuild packages without needing to increment the  patchlevel (`<version>` becomes `<version>-1`, `<version>-1` becomes `<version>-2`, etc).

In addition to simplifying the rebuild process, this eliminates the need for the patchlevel at all.
- We can rebuild packages without incrementing the patchlevel
- We can remove existing patchlevels 
  - relevant for #8196 given that we don't actually need a patchlevel anymore
  - allows us to clean up our version data for Repology. ([This](https://repology.org/projects/?inrepo=chromebrew&problematic=1) tracks all the ignored versions, although this currently only applies to where we have applied patchlevels on otherwise up-to-date packages.)

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=sweep crew update
```

### DIFF
--- a/lib/package_utils.rb
+++ b/lib/package_utils.rb
@@ -1,4 +1,5 @@
 require 'json'
+require 'net/http'
 require_relative 'const'
 
 class PackageUtils
@@ -13,7 +14,7 @@ class PackageUtils
 
   def self.get_url(pkg, build_from_source: false)
     if !build_from_source && pkg.binary_sha256&.key?(ARCH.to_sym)
-      return "https://gitlab.com/api/v4/projects/26210301/packages/generic/#{pkg.name}/#{pkg.version}_#{ARCH}/#{pkg.name}-#{pkg.version}-chromeos-#{ARCH}.#{pkg.binary_compression}"
+      return get_binary_url(pkg)
     elsif pkg.source_url.is_a?(Hash) && pkg.source_url&.key?(ARCH.to_sym)
       return pkg.source_url[ARCH.to_sym]
     else
@@ -28,6 +29,25 @@ class PackageUtils
       return pkg.source_sha256[ARCH.to_sym]
     else
       return pkg.source_sha256
+    end
+  end
+
+  def self.get_binary_url(pkg)
+    # List all the packages with the name and version of the package file.
+    # The name search is fuzzy, so we need to refine it further (otherwise packages like vim, gvim and vim_runtime would break).
+    packages = JSON.parse(Net::HTTP.get(URI("https://gitlab.com/api/v4/projects/26210301/packages?package_name=#{pkg.name}&package_version=#{pkg.version}_#{ARCH}")))
+    # Loop over each result until we get an exact name match, then return the package ID for that match.
+    package_id = 0
+    (0..packages.count - 1).each do |i|
+      next unless packages[i]['name'] == pkg.name
+      package_id = packages[i]['id']
+    end
+    # List all the package files for that package ID.
+    package_files = JSON.parse(Net::HTTP.get(URI("https://gitlab.com/api/v4/projects/26210301/packages/#{package_id}/package_files")))
+    # Loop over each result until we find a matching file_sha256 to our binary_sha256.
+    (0..package_files.count - 1).each do |i|
+      next unless package_files[i]['file_sha256'] == pkg.binary_sha256[ARCH.to_sym]
+      return "https://gitlab.com/chromebrew/binaries/-/package_files/#{package_files[i]['id']}/download"
     end
   end
 
--- a/packages/hello_world_chromebrew.rb
+++ b/packages/hello_world_chromebrew.rb
@@ -12,10 +12,10 @@ class Hello_world_chromebrew < Package
 
   # These are needed to successfully build and check for dependencies.
   binary_sha256({
-    aarch64: '000',
-     armv7l: '000',
-       i686: '111',
-     x86_64: '222'
+    aarch64: '827f9794864aa76f4c99fd31f989077f1fa65771386f73db30a7681842f8736d',
+     armv7l: '827f9794864aa76f4c99fd31f989077f1fa65771386f73db30a7681842f8736d',
+       i686: 'ff0942c505b04982fed187bcda123adead37b3ac2dcfd7e2f0543ca179e81df6',
+     x86_64: '3081f1f25950c91f75db41095b644a2783987a3a7ef2832fc2b85bf138bb006f'
   })
 
   # Register dependencies (use the following line as a basis)
--- a/tests/lib/package_utils.rb
+++ b/tests/lib/package_utils.rb
@@ -50,15 +50,38 @@ class PackageUtilsTest < Minitest::Test
     refute(PackageUtils.compatible?(pkg))
   end
 
-  def test_get_binary_url
+  def test_get_binary_url_old_hash
     pkg = Class.new(Package)
-    pkg.name = 'test_package'
+    pkg.name = 'hello_world_chromebrew'
     pkg.instance_eval do
-      version '1.0'
+      version '1.1'
       binary_compression 'tar.zst'
-      binary_sha256({ ARCH.to_sym => '0000000000000000000000000000000000000000000000000000000000000000' })
+      binary_sha256({
+        aarch64: '29185a6c4a8ecc3532606649f5831e37cd977c24bfd4fb1c12328ca5ba966ff1',
+         armv7l: '29185a6c4a8ecc3532606649f5831e37cd977c24bfd4fb1c12328ca5ba966ff1',
+           i686: '6e46c31245e9e17eeecf03d61454bb7312b6e769bfaf352ced527ee93cc62518',
+         x86_64: '83f674b3e8fe29c7e761ce6adc27fd7df927d985140ff6d1ae64a0046339322b'
+      })
     end
-    assert_equal("https://gitlab.com/api/v4/projects/26210301/packages/generic/test_package/1.0_#{ARCH}/test_package-1.0-chromeos-#{ARCH}.tar.zst", PackageUtils.get_url(pkg))
+    package_file_id = { aarch64: '137956464', armv7l: '137956464', i686: '137956442', x86_64: '137956370' }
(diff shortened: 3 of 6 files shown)


packages/hello_world_chromebrew.rb
```diff
@@ -12,10 +12,10 @@ class Hello_world_chromebrew < Package
...
   # These are needed to successfully build and check for dependencies.
   binary_sha256({
-    aarch64: '000',
-     armv7l: '000',
-       i686: '111',
-     x86_64: '222'
+    aarch64: '827f9794864aa76f4c99fd31f989077f1fa65771386f73db30a7681842f8736d',
+      armv7l: '827f9794864aa76f4c99fd31f989077f1fa65771386f73db30a7681842f8736d',
+        i686: 'ff0942c505b04982fed187bcda123adead37b3ac2dcfd7e2f0543ca179e81df6',
+      x86_64: '3081f1f25950c91f75db41095b644a2783987a3a7ef2832fc2b85bf138bb006f'
```
uberhacker, line 18: The alignment was right earlier.

Zopolis4: Fixed.

uberhacker:
Tests failed with:
```
/usr/local/lib64/ruby/3.3.0/uri/rfc3986_parser.rb:79:in `split': undefined method `to_str' for an instance of Range (NoMethodError)

        uri = uri.to_str
                 ^^^^^^^
Did you mean?  to_set
               to_s
```

Zopolis4:
That's because the unit test is cloning from the chromebrew repo instead of this branch for the hello_world_chromebrew test. Working on a fix.

lib/package_utils.rb
```diff
@@ -30,4 +31,23 @@ def self.get_sha256(pkg, build_from_source: false)
...
+    # The name search is fuzzy, so we need to refine it further (otherwise packages like vim, gvim and vim_runtime would break).
+    packages = JSON.parse(Net::HTTP.get(URI("https://gitlab.com/api/v4/projects/26210301/packages?package_name=#{pkg.name}&package_version=#{pkg.version}_#{ARCH}")))
+    # Loop over each result until we get an exact name match, then return the package ID for that match.
+    package_id = 0
+    (0..packages.count - 1).each do |i|
+      next unless packages[i]['name'] == pkg.name
+      package_id = packages[i]['id']
+    end
+    # List all the package files for that package ID.
+    package_files = JSON.parse(Net::HTTP.get(URI("https://gitlab.com/api/v4/projects/26210301/packages/#{package_id}/package_files")))
```
uberhacker, line 46: Can `https://gitlab.com/api/v4/projects/26210301/packages` be a variable since it is used more than once?  The reason I say this is because the url may change in the future and it would be good to only change on one line.

tests/lib/package_utils.rb
```diff
@@ -50,15 +50,38 @@ def test_not_compatible_arch
...
+      binary_sha256({
+        aarch64: '827f9794864aa76f4c99fd31f989077f1fa65771386f73db30a7681842f8736d',
+         armv7l: '827f9794864aa76f4c99fd31f989077f1fa65771386f73db30a7681842f8736d',
+           i686: 'ff0942c505b04982fed187bcda123adead37b3ac2dcfd7e2f0543ca179e81df6',
+         x86_64: '3081f1f25950c91f75db41095b644a2783987a3a7ef2832fc2b85bf138bb006f'
+      })
     end
-    assert_equal("https://gitlab.com/api/v4/projects/26210301/packages/generic/test_package/1.0_#{ARCH}/test_package-1.0-chromeos-#{ARCH}.tar.zst", PackageUtils.get_url(pkg))
+    package_file_id = { aarch64: '137956527', armv7l: '137956527', i686: '137956521', x86_64: '137956520' }
+    assert_equal("https://gitlab.com/chromebrew/binaries/-/package_files/#{package_file_id[ARCH.to_sym]}/download", PackageUtils.get_url(pkg))
```
uberhacker, line 84: Use a variable for `https://gitlab.com/chromebrew/binaries/-/package_files`?

lib/package_utils.rb
Zopolis4: If the URL changes to anything other than gitlab, this whole thing needs to be rewritten-- I don't think a variable makes sense just to store `26210301` twice.

tests/lib/package_utils.rb
Zopolis4: Same situation as before-- if the url changes to anything other than gitlab, this whole thing is useless, and I don't see us changing from the `chromebrew/binaries` repository to anything else anytime soon? Also, tests are supposed to be isolated from one another so sharing a variale would break that.

uberhacker: This url is actually used more than twice.  A best practice in any development work is to follow the DRY (don't repeat yourself) principle.

Zopolis4: It's only used in this file twice-- I get what you're saying, but I think repeating something twice is acceptable.

uberhacker: It's also used in `lib/package_utils.rb`.  This is even more reason to make it a constant.

Zopolis4: I strongly disagree. Adding an additional source-wide constant for a string that appears three times in two files is an unnecessary abstraction that would negatively impact code quality. Avoiding repetition is good, but this is an overzealous application of that principle.

uberhacker: I wouldn't consider following best practices overzealous.  Think about the next time you might need to use the same url somewhere else and it might make a bit more sense.

Zopolis4: Fair enough. As it stands, I still believe an abstraction is unnecessary and would worsen code quality. However, if this changes in the future, i.e. we start referring to the gitlab repo url more frequently, I will add an abstraction if it is beneficial.

Zopolis4:
Are we good to merge, variable re-use discussions aside? `parted` needs a rebuild, and I'd like to be able to do it without a revision.

satmandu:
Is there a way to get a list of binaries given a partial version name to gitlab?

i.e.,
This works:
https://gitlab.com/api/v4/projects/26210301/packages?package_name=gcc_lib&package_version=14.2.0-glibc2.32_x86_64

But this doesn't:
https://gitlab.com/api/v4/projects/26210301/packages?package_name=gcc_lib&package_version=14.2.0*_x86_64

It would be nice to do a wildcard query to get a list of binaries and then for instance automatically select the lower glibc version, e.g., 2.27 as a fallback if the glibc 2.32 version isn't available.

satmandu:
I guess the question is partly, should we be leveraging gitlab's query interface better... does that require us to upload files in a different directory structure, since the directory structure is what gives us the query options?

i.e., currently we do `#{base_url}/#{package}/#{new_version}_#{arch}/#{new_tarfile}`
But should we be doing `#{base_url}/#{package}/#{arch}/#{glibc_version}/#{gcc_version}/#{new_version}/#{new_tarfile}`

I'm not sure if this is just a hack on top of gitlab's interface... so if there is a more efficient way to handle this I'm all for it... The ability to support an arbitrary number of tags would be useful... the ones that immediately come to mind are: `perl_version`, `python3_version`, `icu_version`, `gcc_version`, `glibc_version`, `llvm_version`, `ruby_version`, `boost_version`.

Zopolis4:
I'm not entirely sure off the top of my head-- I do think we shouldn't be leaning too heavily into the version tagging stuff, given that the majority of it is temporary (`perl_version`, `python3_version` and `ruby_version` will be obsolete once we integrate those languages package managers, `gcc_version` only seems to be useful in cases where we're accidentally linking against static libraries, not sure about `llvm_version`, `glibc_version` is only here until I finish working on musl support and libgcompat, `icu_version` might be unnecessary (see image), which leaves only `boost_version` as a possible issue.)

image


Also, as I discussed on slack, we already have support for versioned dependencies, which would be a more graceful way of handling these sort of things-- we could ensure that a specific version of boost is installed if a package was built with that version, and so on.

satmandu:
> image
> 

Looking at the ICU stuff...  we just need to build with `--disable-renaming` ?  Starting a PR for that RIGHT NOW, since 75.1 is out...

satmandu:
sigh...
```
chronos@cheekon-x86_64 /output/git/satmandu/packages (icu *)$ grep "depends_on 'icu4c" * | wc -l
54
```

supechicken:
I prefer keeping patchlevel for an easier way to ensure the consistency between prebuilt binaries and package script. Some packages like `bash` do care about patchlevel

Also, how can we check for patch-level updates? `curl api.gitlab.com` is a way but it is kinda hacky😅

Zopolis4:
I'm not saying we need to ditch packages which have patchlevels themselves, just that now we don't need to increment a revision for chromebrew rebuilds.

> Also, how can we check for patch-level updates? curl api.gitlab.com is a way but it is kinda hacky😅

the logic to to detect pure rebuild updates is already in chromebrew, and has been for quite some time-- this just allows us to not break existing stuff. 

as for things outside of chromebrew, not sure. :woman_shrugging: that's the best i can think of, check for a new sha256 than the existing one

supechicken:
Oh nevermind, I think I missed that `crew update` PR (but this still doesn't allow us to get the latest patchlevel without gitlab api):
https://github.com/chromebrew/chromebrew/blob/560947f53709c8735496186458ec5198235ff1de/bin/crew#L290-L296

uberhacker (approved):

uberhacker:
@Zopolis4: Merge conflicts.

tests/lib/package_utils.rb
Zopolis4: Thank you, I appreciate the compromise. Let's see how this goes.

Zopolis4:
Fixed.

uberhacker (approved):

uberhacker:
> Fixed.

I just noticed we went from crew version 1.49.8 to 1.45.0.  I think you fixed conflicts the reverse of how they should have been resolved. :face_with_head_bandage:

Zopolis4:
It didn't give me any merge conflicts in `lib/const.rb`, so not sure what happened there :sweat_smile:

Zopolis4:
image
this is the commit from the merge...

PR #10160, opened 2024-07-15
Dbeaver 24.1.0 => 24.1.2

uberhacker:
Tested and working in x86_64 container.


packages/dbeaver.rb
```diff
@@ -3,14 +3,14 @@
...
-  version '24.1.0'
+  version '24.1.2'
   license 'Apache-2.0'
   compatibility 'x86_64'
   source_url({
-    x86_64: 'https://github.com/dbeaver/dbeaver/releases/download/24.1.0/dbeaver-ce-24.1.0-linux.gtk.x86_64.tar.gz'
+    x86_64: "https://github.com/dbeaver/dbeaver/releases/download/#{version}/dbeaver-ce-#{version}-linux.gtk.x86_64.tar.gz"
   })
   source_sha256({
-    x86_64: 'c5018c3e2b3f746c5d5f04886431c653fc8569bde4417ceff3122e46c9f7cd1e'
```
Zopolis4, line 13: If we only support `x86_64`, there's no need to have an architecture-specific `source_url` and `source_sha256`.

uberhacker: Yes, but this also makes it easier to add new architectures if there is another supported in the future.

Zopolis4 (approved):

packages/dbeaver.rb
uberhacker: Fixed.

PR #10197, opened 2024-07-23
New package: LXC

supechicken:
> [!NOTE]
> This package will not work on ChromeOS stock kernel, user need to rebuild their kernel with `SECURITY_CHROMIUMOS=n` to use it (see https://github.com/containers/crun/issues/1121 for more info)

```bash
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=lxc crew update
```


packages/lxc.rb
```diff
@@ -0,0 +1,57 @@
...
+    # FIXME: Remove `date` binary from `zoneinfo`
+    # meson.build:310:16: ERROR: Command `/usr/local/bin/date --utc --date=@1719046192 +%Y-%m-%d` failed with status 1.
+    system 'sed', '-i', "s,find_program('date'),'date',g", 'meson.build'
+    system 'sed', '-i', "s,'date','/bin/date',g", 'meson.build'
+  end
+
+  meson_install_extras do
+    Dir.chdir(File.join(CREW_DEST_PREFIX, 'bin')) do
+      FileUtils.mv 'lxc-start', 'lxc-start.real'
+      File.write 'lxc-start', lxc_start_wrapper, perm: 0o755
```
uberhacker, line 54: Doesn't the move above make this line fail with a file not found?

packages/libseccomp.rb
```diff
@@ -1,33 +1,14 @@
...
-  source_sha256 '96bbadb4384716272a6d2be82801dc564f7aab345febfe9b698b70fc606e3f75'
-  binary_compression 'tar.zst'
-
-  binary_sha256({
-    aarch64: 'd4aafa04c27719b2d5d5be94a89846d0a118710effd5f1144e2139a0b8f9c09d',
-     armv7l: 'd4aafa04c27719b2d5d5be94a89846d0a118710effd5f1144e2139a0b8f9c09d',
-       i686: '96743fd080af70581781c4e0b17bdb3f5c91b765c1355803193527984176f5f9',
-     x86_64: '1097b0549a2f0210fa15ca54f7916b9927363534913dbc3030270c39c2559d3d'
-  })
+  source_url "https://github.com/seccomp/libseccomp.git"
```
uberhacker, line 9: Rubocop doesn't like double quotes unless there is interpolation.

uberhacker:
If lxc actually works, I'll be impressed.

packages/libseccomp.rb
supechicken: Change made

packages/lxc.rb
supechicken: `File.write` will create that file if it does not exist

supechicken:
It works well on my x86 laptop with Brunch installed, but not my ARM Chromebook with stock ChromeOS kernel (see my first comment)

packages/lxc.rb
uberhacker: Which file?  You just renamed lxc-start to lxc-start.real on the previous line so lxc-start no longer exists.

supechicken: > Which file? You just renamed lxc-start to lxc-start.real on the previous line so lxc-start no longer exists.

Yes, the `lxc-start` file no longer exists after this line, but the `File.create 'lxc-start', ...` line will recreate the file `lxc-start` with new content specified in variable `lxc_start_wrapper`

uberhacker:
> It works well on my x86 laptop with Brunch installed, but not my ARM Chromebook with stock ChromeOS kernel (see my first comment)

Were you able to get docker working with this kernel?

packages/lxc.rb
uberhacker: > > Which file? You just renamed lxc-start to lxc-start.real on the previous line so lxc-start no longer exists.
> 
> Yes, the `lxc-start` file no longer exists after this line, but the `File.write 'lxc-start', ...` line will recreate the file `lxc-start` with new content specified in variable `lxc_start_wrapper`

Gotcha, I was disoriented with the source and destination arguments for File.write.

supechicken:
> > It works well on my x86 laptop with Brunch installed, but not my ARM Chromebook with stock ChromeOS kernel (see my first comment)
> 
> Were you able to get docker working with this kernel?

No on both kernels. Even if I managed to bypass the CGroup error mentioned in https://github.com/containers/crun/issues/1115 by mounting a CGroup V2 filesystem in `/sys/fs/cgroup`, docker still complains about the missing `mqueue` support (which isn't enabled in ChromeOS kernels by default)

uberhacker:
> It works well on my x86 laptop with Brunch installed, but not my ARM Chromebook with stock ChromeOS kernel (see my first comment)

We'll need binaries before these can be merged.

packages/lxc.rb
```diff
@@ -0,0 +1,57 @@
...
+    -Dapparmor=false \
+    -Dselinux=false \
+    -Dexamples=false \
+    -Dtests=false \
+    -Dman=false \\
+  EOF
+
+  lxc_start_wrapper = <<~EOF
+    #!/bin/bash -eu
+    mkdir -p #{CREW_PREFIX}/run /sys/fs/cgroup/systemd
```
uberhacker, line 30: Does this require sudo?

packages/lxc.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    if ! mountpoint -q /sys/fs/cgroup/systemd; then
+      # systemd-based distros will not boot without this
+      mount cgroup -t cgroup -o none,name=systemd /sys/fs/cgroup/systemd
```
uberhacker, line 34: Does this require sudo?

packages/lxc.rb
```diff
@@ -0,0 +1,57 @@
...
+    fi
+
+    exec lxc-start.real "${@}"
+  EOF
+
```
Zopolis4, line 39: Should we have a preflight check in order to ensure that the kernel has been recompiled as required?

packages/libseccomp.rb
```diff
@@ -1,33 +1,14 @@
...
-  binary_compression 'tar.zst'
-
-  binary_sha256({
-    aarch64: 'd4aafa04c27719b2d5d5be94a89846d0a118710effd5f1144e2139a0b8f9c09d',
-     armv7l: 'd4aafa04c27719b2d5d5be94a89846d0a118710effd5f1144e2139a0b8f9c09d',
-       i686: '96743fd080af70581781c4e0b17bdb3f5c91b765c1355803193527984176f5f9',
-     x86_64: '1097b0549a2f0210fa15ca54f7916b9927363534913dbc3030270c39c2559d3d'
-  })
+  source_url 'https://github.com/seccomp/libseccomp.git'
+  git_hashtag "v#{@version}"
```
Zopolis4, line 10:
```suggestion
  git_hashtag "v#{version}"
```

packages/libcap.rb
```diff
@@ -24,22 +16,25 @@ class Libcap < Package
...
     # add includes option to make it work with gperf-3.1
     system 'sed -e "/gperf --/s/gperf --/gperf --includes --/" -e "/gperf --/s/cap_lookup_name(/cap_dummy(/" -i libcap/Makefile'
-    # change the path to ld
-    system "sed -i 's,/usr/bin/ld,#{CREW_PREFIX}/bin/ld,g' Make.Rules"
+
+    # change the path to mold
+    system "sed -i 's,/usr/bin/ld,#{CREW_PREFIX}/bin/mold,g' Make.Rules"
+
     # change prefix
     system "sed -i 's,prefix=/usr,prefix=#{CREW_PREFIX},g' Make.Rules"
```
Zopolis4, line 24: Can't we use environment variable overrides for this?

packages/libcap.rb
```diff
@@ -24,22 +16,25 @@ class Libcap < Package
   def self.patch
     # add includes option to make it work with gperf-3.1
     system 'sed -e "/gperf --/s/gperf --/gperf --includes --/" -e "/gperf --/s/cap_lookup_name(/cap_dummy(/" -i libcap/Makefile'
-    # change the path to ld
-    system "sed -i 's,/usr/bin/ld,#{CREW_PREFIX}/bin/ld,g' Make.Rules"
+
+    # change the path to mold
+    system "sed -i 's,/usr/bin/ld,#{CREW_PREFIX}/bin/mold,g' Make.Rules"
```
Zopolis4, line 21:
```make
ifndef lib
lib=$(shell ldd /usr/bin/ld|grep -E "ld-linux|ld.so"|cut -d/ -f2)
endif
```
We should be able to bypass this by just overriding the lib variable to LIB_SUFFIX.

packages/libcap.rb
```diff
@@ -24,22 +16,25 @@ class Libcap < Package
...
-    if ARCH == 'x86_64'
-      # use system libdir
-      system "sed -i 's,^lib_prefix=\$(exec_prefix),lib_prefix=#{CREW_LIB_PREFIX},g' Make.Rules"
-    end
+
+    # use system libdir
+    system "sed -i 's,^lib_prefix=\$(exec_prefix),lib_prefix=#{CREW_LIB_PREFIX},g' Make.Rules"
+
     # http://git.yoctoproject.org/cgit.cgi/poky/plain/meta/recipes-support/libcap/files/0001-ensure-the-XATTR_NAME_CAPS-is-defined-when-it-is-use.patch
     system 'sed -i "s,^\#ifdef VFS_CAP_U32,\#if defined (VFS_CAP_U32) \&\& defined (XATTR_NAME_CAPS),g" libcap/cap_file.c'
```
Zopolis4, line 33: A fix for this issue seems to have been upstreamed a few years ago: https://bugzilla.kernel.org/show_bug.cgi?id=208375

packages/libcap.rb
```diff
@@ -2,20 +2,12 @@
...
   description 'Libcap implements the user-space interfaces to the POSIX 1003.1e capabilities available in Linux kernels.'
-  homepage 'https://directory.fsf.org/wiki/Libcap'
-  version '2.66'
+  homepage 'https://directory.fsf.org/wiki/Libcap/'
+  version '2.70'
   license 'GPL-2 or BSD'
   compatibility 'all'
-  source_url 'https://git.kernel.org/pub/scm/libs/libcap/libcap.git'
+  source_url 'https://kernel.googlesource.com/pub/scm/libs/libcap/libcap.git'
   git_hashtag "libcap-#{version}"
```
Zopolis4, line 10: It looks like the `libcap-2.70` and the `v1.2.70` tags are identical: 

- https://git.kernel.org/pub/scm/libs/libcap/libcap.git/tag/?h=v1.2.70
- https://git.kernel.org/pub/scm/libs/libcap/libcap.git/tag/?h=libcap-2.70

Although, not sure if `v1.#{version}` is more readable or makes more sense than `libcap-#{version}`

packages/libcap.rb
```diff
@@ -2,20 +2,12 @@
...
 class Libcap < Package
   description 'Libcap implements the user-space interfaces to the POSIX 1003.1e capabilities available in Linux kernels.'
-  homepage 'https://directory.fsf.org/wiki/Libcap'
-  version '2.66'
+  homepage 'https://directory.fsf.org/wiki/Libcap/'
+  version '2.70'
   license 'GPL-2 or BSD'
   compatibility 'all'
-  source_url 'https://git.kernel.org/pub/scm/libs/libcap/libcap.git'
+  source_url 'https://kernel.googlesource.com/pub/scm/libs/libcap/libcap.git'
```
Zopolis4, line 9: Any particular reason to swap to google's mirror? It requires a google account to see the gitweb, which isn't great.

Zopolis4, line 9: Any particular reason to swap to google's mirror? It requires a google account to see the gitweb, which isn't great.

packages/lxc.rb
Zopolis4, line 54: There's no need to use the `lxc_start_wrapper` here, right? 

We can just do something like this:
```ruby
File.write 'waylandenv', <<~WAYLAND_ENV_EOF
  # environment set-up for Chrome OS built-in Wayland server
  : "${XDG_RUNTIME_DIR:=/var/run/chrome}"
  : "${XDG_SESSION_TYPE:=wayland}"
  : "${WAYLAND_DISPLAY:=wayland-0}"
  : "${CLUTTER_BACKEND:=wayland}"
  : "${GDK_BACKEND:=wayland}"
WAYLAND_ENV_EOF
FileUtils.install 'waylandenv', "#{CREW_DEST_PREFIX}/etc/env.d/wayland", mode: 0o644
 ```

packages/lxc.rb
supechicken: Yes, but the `lxc-start` command requires root anyway. So it should be fine?

packages/lxc.rb
supechicken: See above

packages/lxc.rb
Zopolis4: In that case, should we add a dependency on `crew-sudo` if we're above the relevant milestone?

packages/libseccomp.rb
supechicken: I prefer using `@version` (a variable) instead of `version` (a function that reads and returns `@version`) for better performance.

packages/libcap.rb
supechicken: I am not sure about it...

Zopolis4: All good, I'll tidy up the build section in the next libcap update then.

packages/lxc.rb
supechicken: The purpose of putting them in a wrapper script is to "do it only when the user needs to run it", also minimizing the impact on the environment :)

packages/libcap.rb
supechicken: Just because `git.kernel.org` is very slow on my network (probably crappy ISP), using Google's mirror is way more faster for me 😅

uberhacker:
Did you give up on this @supechicken?  It'd be nice to have something like this in addition to docker but the main issue is the kernel needs to support virtualization.

satmandu:
It doesn't make sense to leave this compatible with `armv7l`, right? 32-bit host-based virtualization is even being torn out of the Linux kernel...

supechicken:
@uberhacker I have found a workaround for using LXC on CrOS without rebuilding the kernel (see https://github.com/containers/crun/issues/1121#issuecomment-2403280341), but I don't have a Chromebook for testing it at this moment.

@satmandu Containers do not involve virtualization (things like KVM are not needed), they use namespace/cgroup isolation instead (which is available on all Chromebook's kernel by default)

uberhacker:
> @uberhacker I have found a workaround for using LXC on CrOS without rebuilding the kernel (see [containers/crun#1121 (comment)](https://github.com/containers/crun/issues/1121#issuecomment-2403280341)), but I don't have a Chromebook for testing it at this moment.
Can you move out of the country to get outside the Great Firewall of China? :grin: 

> 
> @satmandu Containers do not involve virtualization (things like KVM are not needed), they use namespace/cgroup isolation instead (which is available on all Chromebook's kernel by default)
If this is true, will we be able to get docker working too or does it only apply to lxcs?

supechicken:
> Can you move out of the country to get outside the Great Firewall of China? 😁

The Great Firewall of China has never been deployed in Hong Kong... I could still access the worldwide Internet without using a VPN 😅


> If this is true, will we be able to get docker working too or does it only apply to lxcs?

Theoretically yes, but docker requires `mqueue` and I am not sure if CrOS's kernel supports it.

buhron:
try this
- go to vt-2
- run `sudo /usr/share/vboot/bin/make_dev_ssd.sh -i /dev/mmcblk0 --partitions 2 --save_config /home/chronos/user/kernel.cfg`
- edit `/home/chronos/user/kernel.cfg` and add `lsm=selinux,landlock` to the file
- run `sudo /usr/share/vboot/bin/make_dev_ssd.sh -i /dev/mmcblk0 --partitions 2 --set_config /home/chronos/user/kernel.cfg`
- reboot (I haven't done that yet)
- credits: https://github.com/containers/crun/issues/1121#issuecomment-2403280341, https://www.chromium.org/chromium-os/developer-library/guides/kernel/kernel-development/#kernel-configuration

supechicken:
> it seems that this does not work in M132 (newest version) :/ chromeos prob did this to get more users to switch to their limited-compared-to-a-regular-lxc-container cRoStInI, or to get into their secure mode, getting the people using the dev mode to be in that corner

Really? But there are no significant modifications to CrOS LSM [since the last few milestones](https://chromium.googlesource.com/chromiumos/third_party/kernel/+log/HEAD/security/chromiumos).

Did you reapply the command line to the correct slot after updating? (should be `--partitions 4` if the update was installed to slot B)

Interested-Deving-1896:
https://wiki.gentoo.org/wiki/Incus/Gentoo_Github_pullrequest_testing

Interested-Deving-1896:
https://wiki.gentoo.org/wiki/Incus/Gentoo_Github_pullrequest_testing

supechicken:
@buhron Have you confirmed that the `lsm=...` flag has been correctly passed into the kernel? You can check it by running `cat /proc/cmdline`

Also, you can check the activated LSM list by running `cat /sys/kernel/security/lsm`

buhron:
I have `lsm=none` in cmdline
image
When I `cat /sys/kernel/security/lsm` I get `capability,landlock,yama,loadpin,safesetid,chromiumos,selinux,bpf`, which is not right...

buhron:
> I have `lsm=none` in cmdline image When I `cat /sys/kernel/security/lsm` I get `capability,landlock,yama,loadpin,safesetid,chromiumos,selinux,bpf`, which is not right... I want to rebuild the kernel but I tried once and failed, and don't have a recovery USB to recover... so yeah Edit: Apparently the new ChromeOS update seemed to fix it, or the command line thing only works on slot A

hi how's everyone doing

PR #10198, opened 2024-07-23
Xfsprogs 6.5.0 => 6.9.0, Liburcu 0.14.0-1 => 0.14.0-2

uberhacker:
No longer builds successfully on i686.  All tests passed.


packages/liburcu.rb
```diff
@@ -1,38 +1,23 @@
-require 'package'
+require 'buildsystems/autotools'
 
-class Liburcu < Package
+class Liburcu < Autotools
   description 'Userspace RCU (read-copy-update) library.'
   homepage 'https://liburcu.org/'
-  version '0.14.0-1'
+  version '0.14.0-gcc14'
```
Zopolis4, line 6: Why are we tagging it with the gcc version here?

uberhacker: @satmandu?

uberhacker: I assume this is to indicate the gcc version that compiled the binaries.

Zopolis4: I would have thought that is only nessecary in certain cases, like things interacting with the compiler itself.

uberhacker: The associated libraries are built with the gcc version embedded also.  Below is what can result from different gcc versions.
```
lto1: fatal error: bytecode stream in file '/usr/local/lib64/liburcu.a' generated with LTO version 13.0 instead of the expected 14.0
```

Zopolis4: Sorry to keep moving the goalposts :sweat_smile:, but would building shared libraries instead solve this?

satmandu: Wait, why isn't liburcu being built with shared libraries!

Good catch!

uberhacker: I removed the static libraries and rebuilt the binaries.  Thank you for the feedback.

Zopolis4 (approved):
Out of curiosity, what is the error encountered when trying to build `xfsprogs` for i686? Wondering if that's an intended drop in support or not.

satmandu (approved):

uberhacker:
> Out of curiosity, what is the error encountered when trying to build `xfsprogs` for i686? Wondering if that's an intended drop in support or not.

```
xfile.c:9:10: fatal error: linux/memfd.h: No such file or directory
    9 | #include <linux/memfd.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [../include/buildrules:75: xfile.lo] Error 1
make[1]: *** [include/buildrules:36: libxfs] Error 2
make: *** [Makefile:84: default] Error 2
Command "bash -c make -j12" failed, retrying with "-j1"...
   [HEADERS] include
   [HEADERS] libxfs
Building include
Building libfrog
Building libxfs
    [CC]     xfile.lo
xfile.c:9:10: fatal error: linux/memfd.h: No such file or directory
    9 | #include <linux/memfd.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [../include/buildrules:75: xfile.lo] Error 1
make[1]: *** [include/buildrules:36: libxfs] Error 2
make: *** [Makefile:84: default] Error 2
There was a build error.
```

Zopolis4:
Looks like that's a kernel version thing again (see containers/buildah#1602), I just can't actually find a place to report the bug or send a patch.

PR #10202, opened 2024-07-24
Use Package.load_package instead of require_relative in package files to avoid global scope conflicts

Zopolis4:
Following on from https://github.com/chromebrew/chromebrew/pull/9859#discussion_r1633641170.

I haven't updated `openjdk.rb` or `php.rb` because this change breaks them, meaning they're probably relying on global scope when they shouldn't be.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=matilda crew update
```

### DIFF
--- a/packages/a52.rb
+++ b/packages/a52.rb
@@ -1,5 +1,5 @@
 require 'package'
-require_relative 'liba52'
+Package.load_package("#{__dir__}/liba52.rb")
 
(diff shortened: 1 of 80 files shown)


supechicken (approved):
LGTM

uberhacker (approved):

PR #10219, opened 2024-07-29
Add metadata_cleaner package and py3_mat2 dependency

uberhacker:
Tested and working in x86_64 container.  Unable to build on arm.


packages/metadata_cleaner.rb
```diff
@@ -0,0 +1,26 @@
...
+  depends_on 'desktop_file_utils' # R
+  depends_on 'gtk4' # R
+  depends_on 'itstool' # R
+  depends_on 'libadwaita' # R
+  depends_on 'mesonbuild' => :build
+  depends_on 'py3_mat2' # R
+  depends_on 'pygobject' # R
+  depends_on 'python3' # R
+
+  no_compile_needed
```
Zopolis4, line 21: This is a meson package, right? Why are we setting `no_compile_needed`?

uberhacker: Because we don't need to compile binaries.

Zopolis4: Oh, its entirely implemented in Python. That makes sense.

PR #10240, opened 2024-08-02
crew upload should only upload package files with the right package versions

satmandu:
- Now this should work:
```
for i in 2.23 2.27 2.32 2.33 2.35; do LIBC_VERSION=$i crew upload gcc_lib ; done
```
- This still doesn't fix the issue of updating hashes correctly for all of these variants in a ruby case structure, but that's yet another reason to move to using a json hash or whatever to store available package version and filename info, because the above command will still mangle many hashes because `crew upload` just uses a sed command instead of using jq or something to update a proper data structure correctly. 
- Also, this doesn't fix the issue of crew (and install.sh) knowing to be able to access the same URL for different package versions when one binary is just an alias to another. 

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=gcc crew update
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1458,7 +1458,7 @@ def print_deps_tree(args)
   puts tree_view
 end
 
-def upload(pkg_name = nil)
+def upload(pkg_name = nil, pkg_version = nil)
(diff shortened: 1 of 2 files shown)


uberhacker (approved):
Good call.

PR #10243, opened 2024-08-03
Add zed package

uberhacker:
I know.  Just what we need.  Yet another editor.  Zed is a high-performance, multiplayer code editor from the creators of Atom and Tree-sitter.  It's also open source.  See https://zed.dev/.  Tested and working on x86_64 Chromebook.


Zopolis4 (approved):
i'm gonna be able to edit so much code with this

PR #10244, opened 2024-08-03
Upgrade f* packages

Zopolis4:
Most of them, anyways.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=1muck crew update
```


satmandu (approved):
This is an insane amount of work!

We should probably get a script written to automate some of this if we can...

Zopolis4:
Yeah, that's part of the reason why I'm doing this-- the simple packages that I've converted to use buildsystems and git version tags should be pretty easy to automate.

packages/libspiro.rb
```diff
@@ -0,0 +1,21 @@
...
+  homepage 'https://github.com/fontforge/libspiro'
+  version '20221101'
+  license 'GPL3+'
+  compatibility 'all'
+  source_url 'https://github.com/fontforge/libspiro.git'
+  git_hashtag version
+  binary_compression 'tar.zst'
+
+  binary_sha256({
+    aarch64: 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa',
```
uberhacker, line 14: Missing binary hash?

uberhacker:
Forgot to add new packages and update renamed packages in `tools/packages.yaml`.

uberhacker:
I appreciate all the effort you put into these updates @Zopolis4!  I was feeling alone for a long time.  In the future, it probably makes sense to split up into multiple PRs to make the process more manageable and easier to review.

Zopolis4:
I'm trying to find a balance between a new PR for every package, because then I spend more time opening branches and writing PRs than doing actual work, but I agree this was probably too many in one go.

packages/libspiro.rb
Zopolis4: I have no idea how that even happened, given that `crew upload` should replace both `aarch64` and `armv7l` given that they have the same base hash.

uberhacker: The process probably shouldn't rely on a template but rather generate the entire section if it isn't present.

Zopolis4:
New and renamed packages are covered by `version.rb`.
image

PR #10246, opened 2024-08-03
icu4c => 75.1 + updates

satmandu:
- The `--disable-renaming` option for icu is horribly broken. It requires adding `-DU_DISABLE_RENAMING=1` to all sorts of builds, and even that fails. (libxml2 refuses to build with cmake when that is enabled.) For more on this see a discussion from over ten years ago: https://bugzilla.redhat.com/show_bug.cgi?id=856594

There is a LAUNDRY LIST of packages that needed/need to be updated:
- [x] appstream (updating because other packages here depend upon it.)
- [x] bind
- [x] boost
- [x] caja
- [x] dehtml
- [x] evolution_data_server
-   _freerdp ([broken](https://github.com/FreeRDP/FreeRDP/issues/10466))_
- [x] geany_plugins
- [x] gettext
- [x] glib => 2.81.1 (updating because other packages here depend upon it.)
-   _gnome_docking_library (Currently [broken upstream](https://gitlab.gnome.org/GNOME/gdl/-/issues/9).)_
- [x] gnome_text_editor
- [x] gobject_introspection => 1.80.1 (updating because other packages here depend upon it.)
- [x] graphicsmagick
- [x] gspell
- [x] gtk3 => 3.24.42 (updating because other packages here depend upon it.)
- [x] gtk4 => 4.15.4 (updating because other packages here depend upon it.)
-   _gtksourceview_3 (deprecated - no dependencies.)_
- [x] gtksourceview_5 => 5.13.1 (updating because other packages here depend upon it.)
-   _handbrake (wait on ffpmeg7 update for this.)_
- [x] harfbuzz
-   _icecream (appears abandoned upstream)_
- [x] imagemagick7
- [x] imlib2
-   _js102 (deprecated - no dependencies.)_
- [x] _js115 (not being updated, might not need updating, but update does not build due to python issues)_
- [x] ldb (updating because other packages here depend upon it.)
- [x] libadwaita => 1.6.beta (updating because other packages here depend upon it)
- [x] libabigail
- [x] libarchive
- [x] libcdr
- [x] libdbusmenu_gtk3 (updating because other packages here depend upon it & remove `gtk2` dep.)
- [x] libgedit_gfls (Adding because other packages here depend upon it.)
- [x] libgedit_gtksourceview (Updating because other packages here depend upon it. )
- [x] libgsf
- [x] libical
- [x] libmetalink
- [x] librsvg
- [x] libsass => 3.6.6 (updating because other packages here depend upon it)
- [x] libvisio
- [x] libxml2_autotools => 2.13.3
- [x] libxml2 => 2.13.3
- [x] libxslt
- [x] mold (updating because other packages here depend upon it)
- [x] mysql
- [x] nautilus => 46.2 (updating because deps had to be updated anyways)
- [x] ncurses => 6_5_20240727 (updating because other packages here depend upon it.)
-   php74 (consider updating after new containers are built after this PR is merged.)
-   php80 (consider updating after new containers are built after this PR is merged.)
-  php81 (consider updating after new containers are built after this PR is merged.)
-   php82 (consider updating after new containers are built after this PR is merged.)
-   php83 (consider updating after new containers are built after this PR is merged.)
- [x] postgresql
- [x] python3 => 3.12.5
- [x] py3_pyicu
-  qt5_base (consider updating after new containers are built after this PR is merged.)
-  qt5_location (consider updating after new containers are built after this PR is merged.)
-  qt5_webengine (consider updating after new containers are built after this PR is merged.)
- [x] r
- [x] ruby_pry_byebug (used in debugging...)
- [x] sdl2_tff (adding because other packages here depend upon it.)
- [x] sdl3 (adding because other packages here depend upon it.)
- [x] sdl3_tff (adding because other packages here depend upon it.)
- [x] shared_mime_info
- [x] smbclient
- [x] tdb (updating because other packages here depend upon it.)
-   _tepl_5 (deprecated - no dependencies.)_
- [x] tepl_6
- [x] tesseract
- [x] localsearch _renamed from tracker3_miners_
- [x] tinysparql _renamed from tracker3_
- [x] uriparser (adding because other packages here depend upon it.)
- [x] vte
- [x] wayland
- [x] xwayland => 24.1.2 (updating because other packages here depend upon it.)
-  webkit2gtk_4_1 (consider updating after new containers are built after this PR is merged.)
-  webkit2gtk_4 (consider updating after new containers are built after this PR is merged.)
-  webkitgtk_6 (consider updating after new containers are built after this PR is merged.)
- [x] xfsprogs
- [x] yelp

Works properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=icu crew update \
&& yes | crew upgrade
```


packages/libxml2.rb
```diff
@@ -35,8 +43,13 @@ def self.patch
   end
 
   cmake_options '-DLIBXML2_WITH_PYTHON=OFF \
```
uberhacker, line 45: Why is python off?

satmandu: That's built in `py3_libxml2.rb` .

No reason that has to happen though...

satmandu: I removed py3_libxml2...

packages/libxml2_autotools.rb
```diff
@@ -6,18 +6,18 @@
 class Libxml2_autotools < Autotools
   description 'Libxml2 is the XML C parser and toolkit developed for the Gnome project.'
   homepage 'http://xmlsoft.org/'
-  version Libxml2.version
+  version Libxml2.version.gsub(/-py.*$/, '')
   license 'MIT'
   compatibility 'all'
   source_url 'https://gitlab.gnome.org/GNOME/libxml2.git'
   git_hashtag "v#{version.split('-').first}"
```
uberhacker, line 13: No need to split here if version is already transformed?

satmandu: It turns out that the libxml2 internal build of the python bindings drop files in the wrong place, so I'm backing out this change.

Zopolis4:
Their solution to the problems of requiring mass rebuilds is just doing mass rebuilds :sob: At least we tried `--disable-renaming`, and thanks for putting in the rebuild work here!

packages/libxml2_autotools.rb
```diff
@@ -0,0 +1,46 @@
+Package.load_package("#{__dir__}/libxml2.rb")
+# cmake builds are dependent upon libxml2, so we need a libxml2 build
```
Zopolis4, line 2: In that case, why are we adding another version and not just using autotools on the main build?

satmandu: A bunch of cmake specific files get added to the build if we use cmake, which I guess are used by other builds...

satmandu:
```
crew files libxml2 | grep cmake
/usr/local/lib64/cmake/libxml2-2.13.3/libxml2-config-version.cmake
/usr/local/lib64/cmake/libxml2-2.13.3/libxml2-config.cmake
/usr/local/lib64/cmake/libxml2-2.13.3/libxml2-export-release.cmake
/usr/local/lib64/cmake/libxml2-2.13.3/libxml2-export.cmake
```

Zopolis4: But cmake needs libxml2 to build and circular dependencies don't work with libraries with versioned symbols... great.

Zopolis4: Although-- https://gitlab.gnome.org/GNOME/libxml2/-/issues/709 and https://gitlab.gnome.org/GNOME/libxml2/-/issues/708 seem to suggest we might be able to get away with not using icu?

Zopolis4:
Given that libxml2 is talking about dropping icu4c support altogether, I wonder if there are any other packages where we can avoid using it-- anything to lessen this rebuild workload would be a boon.

satmandu:
> Given that libxml2 is talking about dropping icu4c support altogether, I wonder if there are any other packages where we can avoid using it-- anything to lessen this rebuild workload would be a boon.


Feel free to do the research! But libxml2 is at least taken care of for version 75.1 of ICU...

Zopolis4:
> The --disable-renaming option for icu is horribly broken. It requires adding -DU_DISABLE_RENAMING=1 to all sorts of builds, and even that fails. (libxml2 refuses to build with cmake when that is enabled.) For more on this see a discussion from over ten years ago: https://bugzilla.redhat.com/show_bug.cgi?id=856594

Have you reported this upstream as a fresh issue? Might be worth a shot, at the very least.

satmandu:
> > The --disable-renaming option for icu is horribly broken. It requires adding -DU_DISABLE_RENAMING=1 to all sorts of builds, and even that fails. (libxml2 refuses to build with cmake when that is enabled.) For more on this see a discussion from over ten years ago: https://bugzilla.redhat.com/show_bug.cgi?id=856594
> 
> Have you reported this upstream as a fresh issue? Might be worth a shot, at the very least.

Absolutely not. ICU is a huge package, with a long history, and I really don't want to deal with that right now.

Zopolis4:
If we're adding all the `-icu75.1` tags, can you extend `PackageUtils.get_clean_version` to handle that?

Something like this should work, but i'd recommend adding tests just to be sure.(Also, the check should occur before we clean the git tag suffixes.)
```ruby
# Delete -icu75.1, futureproofed until icu 100
pkg_version.gsub!(/-icu\d{2}\.\d/, '')
```

satmandu:
> If we're adding all the `-icu75.1` tags, can you extend `PackageUtils.get_clean_version` to handle that?
> 
> Something like this should work, but i'd recommend adding tests just to be sure.(Also, the check should occur before we clean the git tag suffixes.)
> 
> ```ruby
> # Delete -icu75.1, futureproofed until icu 100
> pkg_version.gsub!(/-icu\d{2}\.\d/, '')
> ```

Sure. But I maintain that using `-` delimiters between fields in version is a weak hack and we should **_really_** just use different variables for this stuff to generate unique versions for packages.

Take for instance this version entry:

`chromaprint.rb:  version '1.5.1-aa67c95-1'`

```ruby
  # Delete git version tags (1.2.4-qnd73k6), avoiding overmatching and hitting things that arent git hashtags.
  pkg_version.gsub!(/-[\w]{7}$/, '')
```
Currently that gives `nil`...

Zopolis4:
Well we currently don't need to be doing any dependency version tagging at all because we have versioned dependency support, and I'm working on removing the need for rebuild version increments, which only leaves commit information.

Zopolis4:
if you set the version in the dependency, you can probably just grep all depends_on icu lines and then grep that for 75.1-- i think the icu-dependent packages would need to be tagged with depends_on 'icu4c','== 75.1' if they were built with 75.1, so just grep for that.

satmandu:
Currently hacking up the glibc 2.37 containers and fixing builds on them by:
moving our built `libc.so.6` to `/usr/local/lib64/libC.so.6`
copying the system `/lib64/libc.so.6` to `/usr/local/lib64/`
using `patchelf --add-needed libC.so.6` for: libc.so.6, libm.so.6, libstdc++.so.6 (which is from gcc_lib)
building packages with `CREW_LINKER_FLAGS=' /usr/local/lib64/libC.so.6 '`

Zopolis4:
The rubocop is passing even though it's reporting warnings in the CI-- are you able to reproduce those checks locally?

satmandu:
> The rubocop is passing even though it's reporting warnings in the CI-- are you able to reproduce those checks locally?

I'm running Robocop on them locally too and they tend to pass ... The CI has been hit and miss with this PR...

Zopolis4:
What version of Rubocop?

satmandu:
> What version of Rubocop?

Whatever is installed by default in a newly built Chromebrew container...

Zopolis4:
Ah-- thats `1.63.2`, whereas the latest is `1.65.1`. that might be it

satmandu:
> Ah-- thats `1.63.2`, whereas the latest is `1.65.1`. that might be it

Nope. Using 1.65.1

```
rubocop --version
1.65.1
```

satmandu:
https://github.com/chromebrew/chromebrew/pull/10272 is basically already a part of this, so if that gets merged, this should hopefully be good to go.

bin/crew
```diff
@@ -676,6 +677,10 @@ def build_and_preconfigure(target_dir)
     puts 'Preconfiguring package...'
     cache_build if CREW_CACHE_BUILD
     @pkg.install
+    build_end_time = Time.now.to_i
+    build_time_elapsed = build_end_time - build_start_time
+    crewlog "Build & install time for #{@pkg.name} was #{build_time_elapsed} seconds."
+    puts "Build & install time for #{@pkg.name} was #{build_time_elapsed} seconds. Please ask for #{ARCH} binaries to be generated for #{@pkg.name}.".lightpurple if build_time_elapsed > 60
```
uberhacker, line 683: Include hours and minutes in the time?  Some builds can require all night and it will be hard to determine the actual time without converting.

tools/packages.yaml
```diff
@@ -8906,6 +8941,11 @@ url: https://github.com/upx/upx/releases/
 activity: low
 ---
 kind: url
+name: uriparser
+url:https://github.com/uriparser/uriparser/releases/
+activity: low
+---
```
uberhacker, line 8947: The yaml "error" is a bit deceiving.  The reason is there needs to be a space after `url:`.

bin/crew
satmandu: Added hours and minutes...

tools/packages.yaml
satmandu: Fixed!

.github/workflows/Unit-Test.yml
```diff
@@ -41,7 +41,7 @@ jobs:
             cd ~ && \
             git clone --depth=1 https://github.com/chromebrew/chromebrew.git build_test && \
             cd build_test && \
-            yes | CREW_CACHE_ENABLED=1 crew build -f packages/hello_world_chromebrew.rb"
+            yes | CREW_CACHE_ENABLED=1 crew build -vf packages/hello_world_chromebrew.rb"
```
uberhacker, line 44: Make verbose optional?

lib/package_utils.rb
```diff
@@ -44,6 +44,8 @@ def self.get_clean_version(pkg_version)
     pkg_version.delete_suffix!("-glibc#{LIBC_VERSION}")
     # Delete git version tags (1.2.4-qnd73k6), avoiding overmatching and hitting things that arent git hashtags.
     pkg_version.gsub!(/-[\w]{7}$/, '')
+    # Delete -icu75.1, futureproofed until icu 100
+    pkg_version.gsub!(/-icu\d{2}\.\d/, '')
```
uberhacker, line 48: Would `pkg_version.split('-').first` allow unlimited future proofing?

.github/workflows/Unit-Test.yml
satmandu: It is just there to make sure verbose mode works.

(Verbose mode is disabled for the armv7l run of the unit test.)

lib/package_utils.rb
satmandu: No. There are packages like imagemagick that literally use `-number` as a part of their versioning...

.github/workflows/Unit-Test.yml
uberhacker: Yeah, that makes sense.  I just realized this was for tests.

packages/icu4c.rb
```diff
@@ -24,16 +24,15 @@ def self.build
...
         --enable-shared \
         --disable-samples \
         --disable-tests"
       system 'make'
     end
   end
 
-  @icuver = '74.2'
-  @oldicuver = %w[73]
+  @icuver = '75.1'
```
uberhacker, line 34: Use `version` here instead?  Not sure how to handle incrementing old version.  Maybe `@oldicuver = version.split('.').first.to_i - 1`?

packages/py3_identify.rb
```diff
@@ -0,0 +1,14 @@
...
+  description 'File identification library for Python'
+  homepage 'https://github.com/pre-commit/identify'
+  version '2.6.0-py3.12'
+  license 'MIT'
+  compatibility 'all'
+  source_url 'SKIP'
+
+  depends_on 'python3' => :build
+
+  no_compile_needed
```
uberhacker, line 13: I'm thinking why not add `depends_on 'python3' => :build` and `no_compile_needed` to the pip buildsystem since all python packages installed via pip need these anyway?  It would reduce redundancy.

packages/python3.rb
```diff
@@ -3,18 +3,18 @@
 class Python3 < Package
   description 'Python is a programming language that lets you work quickly and integrate systems more effectively.'
   homepage 'https://www.python.org/'
-  version '3.12.4'
+  version '3.12.5'
   license 'PSF-2.0'
   compatibility 'all'
-  source_url "https://www.python.org/ftp/python/#{version}/Python-#{version}.tar.xz"
-  source_sha256 'f6d419a6d8743ab26700801b4908d26d97e8b986e14f95de31b32de2b0e79554'
+  source_url "https://www.python.org/ftp/python/#{version}/Python-#{version.split('-').first}.tar.xz"
```
uberhacker, line 9: `version` doesn't need to be split?

tools/packages.yaml
```diff
@@ -4217,12 +4222,17 @@ activity: low
...
+kind: url
+name: libgedit_gfls
+url: https://gitlab.gnome.org/World/gedit/libgedit-gfls/-/tags
 activity: medium
 ---
 kind: url
 name: libgedit_gtksourceview
-url: https://github.com/gedit-technology/libgedit-gtksourceview/releases
+url: https://gitlab.gnome.org/World/gedit/libgedit-gtksourceview/-/tags
```
uberhacker, line 4235: Missing new package `libxml2_autotools`?

tools/packages.yaml
```diff
@@ -7331,6 +7351,11 @@ url: https://github.com/pypa/hatch/releases
 activity: medium
 ---
 kind: url
+name: py3_identify
+url: https://github.com/pre-commit/identify/tags
+activity: medium
+---
+kind: url
```
uberhacker, line 7358: Missing new package `ruby_pry_byebug`?

packages/py3_identify.rb
Zopolis4: Also, most of them probably depend on python for more than building.

tools/packages.yaml
satmandu: No this is just a stub install for autotools.

satmandu:
Also added the updates in https://github.com/chromebrew/chromebrew/pull/10293 ...

packages/crew_sudo.rb
```diff
@@ -6,9 +8,8 @@ class Crew_sudo < Package
   version '1.1'
   license 'GPL-3'
   compatibility 'all'
-
-  source_url 'https://github.com/chromebrew/crew-sudo.git'
-  git_hashtag "v#{@version}"
+  source_url 'https://github.com/chromebrew/crew-sudo/archive/refs/tags/v1.1.tar.gz'
+  source_sha256 'f71d6a24bc6d1271a7ebbe255a53d08f55fd95d2430d9c14b8d10f82f29c610f'
```
uberhacker, line 12: Missing crew_sudo.filelist(s) in the manifest directory?

satmandu: Not building binaries...

bin/crew
```diff
@@ -1493,6 +1493,7 @@ def print_deps_tree(args)
 end
 
 def upload(pkg_name = nil, pkg_version = nil)
+  abort "\nPackage to be uploaded was not specified.\n".lightred if pkg_name.nil?
```
uberhacker, line 1496: There is already logic a few lines down that loops through all binaries and uploads all if no package argument(s) are provided.  I think the issue is it aborts if there are no .tar.xz binaries.

bin/crew
```diff
@@ -1492,35 +1492,30 @@ def print_deps_tree(args)
...
-  gitlab_token = ENV.fetch('GITLAB_TOKEN', nil)
-  base_url = 'https://gitlab.com/api/v4/projects/26210301/packages/generic'
 
-  if pkg_name.nil?
-    %w[.tar.xz .tar.zst].each do |ext|
-      packages += `find #{CREW_LOCAL_REPO_ROOT}/release/*/*#{ext} -exec basename -s #{ext} {} + | cut -d- -f1 | LC_ALL=C sort | uniq | xargs`
-      packages += ' '
-    end
-    abort 'No package binaries found.'.lightred if packages.empty?
-  end
```
uberhacker, line 1473: Why are we removing this logic?  It would be helpful to upload in bulk for multiple binaries, right?

uberhacker (approved):

bin/crew
satmandu: There is already upload being invoked for each package. There should be one upload per package per architecture...

bin/crew
satmandu: This still looks for the right binaries though, no?

satmandu: This code is there to abort early if no pkg_name is specified, as when one just invokes `crew upload` without anything else.

The code later on separately looks for whether binaries are available.

uberhacker: I don't think you understand how this was originally intended to work.  If no package arguments are provided then it should still upload all the binaries it can find in the release directories.

bin/crew
```diff
@@ -1509,25 +1540,29 @@ def upload(pkg_name = nil, pkg_version = nil)
...
-        puts "curl -Ls #{new_url} | sha256sum" if CREW_VERBOSE
-        upstream_sha256 = `curl -Ls #{new_url} | sha256sum`.chomp.split.first
-        if upstream_sha256 == new_sha256
-          puts output.lightgreen
-        else
-          if CREW_VERBOSE
-            puts "expected sha256 hash=#{new_sha256}"
-            puts "upstream sha256 hash=#{upstream_sha256}"
+      if `curl -sI #{new_url}`.lines.first.split[1] == '200'
+        abort "#{new_tarfile} has already been uploaded. Please change the #{package} package version from #{new_version} and try again.".lightred
```
uberhacker, line 1544: I don't think this is a good idea.  I've built binaries and found out later they had issues but there was no need to bump the version since the package changes were not committed yet.  It doesn't do any harm to overwrite existing binaries as long as the correct sha256 hashes accompany them in the package.

lib/downloader.rb
```diff
@@ -59,9 +59,12 @@ def downloader(url, sha256sum, filename = File.basename(url), verbose = false)
...
 
   unless (sha256sum =~ /^SKIP$/i) || (calc_sha256sum == sha256sum)
-    if @opt_force
-      puts "Updating checksum for #{filename} from #{sha256sum} to #{calc_sha256sum} in #{CREW_LOCAL_REPO_ROOT}/packages/#{@pkg_name}.rb .".lightblue
-      system "sed -i 's/#{sha256sum}/#{calc_sha256sum}/' #{CREW_LOCAL_REPO_ROOT}/packages/#{@pkg_name}.rb"
+    if CREW_FORCE
+      pkg_name = @pkg_name.empty? ? name : @pkg_name
+      puts "Updating checksum for #{filename}".lightblue
+      puts "from #{sha256sum} to #{calc_sha256sum}".lightblue
+      puts "in #{CREW_LOCAL_REPO_ROOT}/packages/#{pkg_name}.rb .".lightblue
```
uberhacker, line 66: I know we didn't have this earlier but it might be a good idea to remind the package contributor they need to commit their changes and submit a PR?

satmandu: That's a good idea!

uberhacker (approved):

PR #10247, opened 2024-08-03
Fix missing aarch64 hash in libspiro

Zopolis4:
Oh, I see the problem-- I didn't hold down the `a` key for long enough when creating the blank `aarch64` hash so it didn't replace it. oops

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=howeven crew update
```

### DIFF
--- a/packages/libspiro.rb
+++ b/packages/libspiro.rb
@@ -11,7 +11,7 @@ class Libspiro < Autotools
   binary_compression 'tar.zst'
 
   binary_sha256({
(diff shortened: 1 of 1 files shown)


uberhacker (approved):

uberhacker:
The current `crew upload` doesn't rely on that line to be there so I'm not sure how different it is on your end.

Zopolis4:
That functionality has never worked for me-- I've got to manually populate an empty binary_sha256 each time. I should probably just get one to copy and paste.

PR #10249, opened 2024-08-04
Add get_clean_version in PackageUtils and use it in json.rb and version.rb

Zopolis4:
Should reduce the number of bad versions on Repology, as well as get some nicer matches for version.rb (In cases where git hashtags start with `a` and are counted as alpha versions, for example).

There are some suffixes that I haven't taken care of:
- `boost1.85` 
  - `./source_highlight.rb:  version '3.1.9-894cacd-boost1.85'`
- `gcc14`
  - `./binutils.rb:  version '2.42-gcc14'`
  - `./gdb.rb:  version '15.1-gcc14-py3.12'`
- `perl-5.36` (as opposed to the usual `perl5.36` or `5.40` or whatever)
  - `./perl_sgmls.rb:  version '1.1-perl-5.36'`
  - `./perl_text_charwidth.rb:  version '0.04-perl-5.36'`
  - `./perl_text_unidecode.rb:  version '1.30-perl-5.36'`

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=thesource crew update
```

### DIFF
--- a/.github/workflows/Repology.yml
+++ b/.github/workflows/Repology.yml
@@ -18,6 +18,7 @@ jobs:
       - name: Generate repology.json
         env:
           CREW_KERNEL_VERSION: 5.1
+          LIBC_VERSION: 2.37
         run: |
           ruby -Ctools json.rb
       - name: Create Pull Request
--- a/tests/lib/package_utils.rb
+++ b/tests/lib/package_utils.rb
@@ -120,4 +120,27 @@ class PackageUtilsTest < Minitest::Test
     end
     assert_equal('4444444444444444444444444444444444444444444444444444444444444444', PackageUtils.get_sha256(pkg, build_from_source: true))
   end
+
+  def test_get_clean_python_version
+    assert_equal('1.2.3', PackageUtils.get_clean_version('1.2.3-py3.12'))
+  end
+
+  def test_get_clean_perl_version
+    assert_equal('0.004.2', PackageUtils.get_clean_version('0.004.2-perl5.40'))
+  end
+
+  def test_get_clean_llvm_version
+    assert_equal('72.93', PackageUtils.get_clean_version('72.93-llvm18'))
+  end
+
+  def test_get_clean_glibc_version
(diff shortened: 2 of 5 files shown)


Zopolis4:
The main remaining source of "bad versions" is our rebuild suffix tags, which are taken care of by #10152, and then once both of these PRs are merged I can work on tracking down individual problems.

.github/workflows/Repology.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       - name: Generate repology.json
         env:
           CREW_KERNEL_VERSION: 5.1
+          LIBC_VERSION: 2.37
```
uberhacker, line 21: Why are these environment variables necessary?

tests/lib/package_utils.rb
```diff
@@ -120,4 +120,27 @@ def test_get_sha256_build_from_source
...
+  def test_get_clean_glibc_version
+    assert_equal('9.5.18', PackageUtils.get_clean_version('9.5.18-glibc2.39'))
+  end
+
+  def test_get_clean_git_version
+    assert_equal('0.0.7', PackageUtils.get_clean_version('0.0.7-8ab26so'))
+    assert_equal('8.2.4-2', PackageUtils.get_clean_version('8.2.4-2-zh725k9'))
+    assert_equal('579-4', PackageUtils.get_clean_version('579-4-1628457'))
+    assert_equal('2.1.5-20220429', PackageUtils.get_clean_version('2.1.5-20220429'))
+  end
```
uberhacker, line 145: It's great we have test functions but how are they executed?

Zopolis4: They're run by minitest when you execute the file.

.github/workflows/Repology.yml
Zopolis4: Otherwise the github actions runner reports a kernel and libc version we don't support, which causes some problems.

uberhacker (approved):

PR #10250, opened 2024-08-05
Fix PackageUtils.get_binary_url when trying to install a non-uploaded package

Zopolis4:
oops

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=ansc crew update
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -2,7 +2,7 @@
 # Defines common constants used in different parts of crew
 require 'etc'
 
-CREW_VERSION = '1.45.0'
+CREW_VERSION = '1.45.1'
 
 # Kernel architecture.
 KERN_ARCH = Etc.uname[:machine]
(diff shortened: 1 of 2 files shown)


lib/const.rb
```diff
@@ -2,7 +2,7 @@
 # Defines common constants used in different parts of crew
 require 'etc'
 
-CREW_VERSION = '1.45.0'
+CREW_VERSION = '1.45.1'
```
uberhacker, line 5: This should be 1.49.9 since the last version was 1.49.8.

uberhacker:
Let's revert the last commit and resolve conflicts correctly the next time.  See #10253.

lib/const.rb
Zopolis4: i don't understand what you mean-- the current version is 1.50.0 in this PR, and the previous PR went from 1.49.9 to 1.50.0

uberhacker: This PR downgraded crew from 1.49.9 to 1.45.0.

Zopolis4: ohhhh i can't read and i did the version number wrong oops

Issue #10251, opened 2024-08-05
🐞 [Bug] Unable to get prebuilt binary URL on older ChromeOS versions


supechicken:
**Describe the bug**
---
The new `get_binary_url(pkg)` function in #10152 does not work on older ChromeOS versions (one of my Chromebooks is on M84 and unable to do a clean install)

```
lzlib: Lzlib is a data compression library providing in-memory LZMA compression and decompression functions, including integrity checking of the decompressed data.
/usr/local/lib/ruby/3.3.0/net/protocol.rb:46:in `connect_nonblock': SSL_connect returned=1 errno=0 peeraddr=172.65.251.78:443 state=error: certificate verify failed (self-signed certificate in certificate chain) (OpenSSL::SSL::SSLError)
        from /usr/local/lib/ruby/3.3.0/net/protocol.rb:46:in `ssl_socket_connect'
        from /usr/local/lib/ruby/3.3.0/net/http.rb:1674:in `connect'
        from /usr/local/lib/ruby/3.3.0/net/http.rb:1580:in `do_start'
        from /usr/local/lib/ruby/3.3.0/net/http.rb:1569:in `start'
        from /usr/local/lib/ruby/3.3.0/net/http.rb:1029:in `start'
        from /usr/local/lib/ruby/3.3.0/net/http.rb:822:in `get_response'
        from /usr/local/lib/ruby/3.3.0/net/http.rb:803:in `get'
        from /usr/local/lib/crew/lib/package_utils.rb:38:in `get_binary_url'
        from /usr/local/lib/crew/lib/package_utils.rb:17:in `get_url'
        from /usr/local/bin/crew:418:in `download'
        from /usr/local/bin/crew:1231:in `install'
        from /usr/local/bin/crew:1209:in `block in resolve_dependencies'
        from /usr/local/bin/crew:1206:in `each'
        from /usr/local/bin/crew:1206:in `resolve_dependencies'
        from /usr/local/bin/crew:1089:in `resolve_dependencies_and_install'
        from /usr/local/bin/crew:1760:in `block in install_command'
        from /usr/local/bin/crew:1745:in `each'
        from /usr/local/bin/crew:1745:in `install_command'
        from /usr/local/bin/crew:1885:in `<main>'
```

**Additional context**
---
- `cat /etc/lsb-release`
```
CHROMEOS_AUSERVER=https://tools.google.com/service/update2
[...]

PR #10254, opened 2024-08-05
lib/downloader: Refactor and add git support

supechicken:
Part of fixing #10251 

## Changes
- Refactor `downloader.rb`, put all related functions into a class
- Add git support
- Remove external downloader support for simplifying codes
- Support returning the downloaded content directly without saving it to filesystem first

### DIFF
--- a/lib/downloader.rb
+++ b/lib/downloader.rb
@@ -1,182 +1,168 @@
 require 'io/console'
 require 'digest/sha2'
+require 'securerandom'
+require 'resolv-replace'
+require 'net/http'
+require 'stringio'
 require 'uri'
 require_relative 'const'
 require_relative 'color'
 require_relative 'progress_bar'
 
-begin
-  require 'securerandom'
-  require 'resolv-replace'
-  require 'net/http'
-rescue RuntimeError => e
-  # hide the error message and fallback to curl if securerandom raise an error
-  if e.message == 'failed to get urandom'
-    Object.send(:remove_const, :CREW_USE_CURL)
-    CREW_USE_CURL = true
-  else
-    raise
-  end
-end
+class Downloader
+  def self.downloader(url, sha256sum, destination = File.basename(url), verbose = false)
+    # downloader: wrapper for all Chromebrew downloaders (`net/http`,`curl`...)
+    # Usage: downloader <url>, <sha256sum>, <dest::optional>, <verbose::optional>
+    #
+    #           <url>: URL that points to the target file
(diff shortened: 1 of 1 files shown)


lib/const.rb
```diff
@@ -128,20 +128,13 @@
...
-
 # Downloader maximum retry count
 CREW_DOWNLOADER_RETRY = ENV.fetch('CREW_DOWNLOADER_RETRY', 3).to_i
 
-# show download progress bar or not (only applied when using the default ruby downloader)
+# show download progress bar or not (only applied when using the default ruby Downloader.download)
 CREW_HIDE_PROGBAR = ENV.fetch('CREW_HIDE_PROGBAR', '0').eql?('1')
 
-# set certificate file location for lib/downloader.rb
+# set certificate file location for lib/Downloader.download.rb
```
Zopolis4, line 137: Bit of an overzealous find/replace here, i think

lib/downloader.rb
```diff
@@ -1,182 +1,178 @@
...
+    url_params    = URI.decode_www_form(uri.query).to_h
+    git_branch    = url_params['branch']
+    git_hashtag   = url_params['hashtag']
+    no_submodules = url_params['no_submodules'].eql?('1')
+
+    uri.scheme = 'https' if uri.scheme == 'git+https'
+
+    FileUtils.mkdir_p destination
+    Dir.chdir(destination) do
+      system 'git', 'init'
```
Zopolis4, line 146: We don't need to split stuff here, do we? We aren't passing `-j8` or anything...

lib/downloader.rb
```diff
@@ -1,182 +1,178 @@
...
+      system 'git', 'config', 'advice.detachedHead', 'false' # suppress "You are in 'detached HEAD' state" warning
+      system 'git', 'remote', 'add', 'origin', uri.to_s.rpartition('?')[0]
+      system 'git', 'fetch', 'origin', (git_hashtag || git_branch)
```
Zopolis4, line 149: We've lost support for fetchtags here-- If we're going to be moving the git download code to here, we need to be passing the package object or at the very least all of the git variables in it.

Zopolis4:
It looks like we're moving a lot of the logic into `lib/downloader.rb`-- this isn't necessarily an issue, but it does mess with my refactor: https://github.com/Zopolis4/chromebrew/blob/d8e194a167c212f3b84100087820a265c9d0ff92/commands/download.rb

personally, I think the git download code fits a lot better in here, because we have all the package information here. If you're able to implement it cleanly, then thats fine by me, though.

PR #10258, opened 2024-08-06
Upgrade serf, scons, sngrep, peg, srt, sshuttle

Zopolis4:
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=around1 crew update
```


uberhacker (approved):
This all looks good, although it still pushes the limit of how much should go in a single PR.  The main reason we should make incremental changes is if something goes wrong, it's easier to revert and fix one thing rather than having to roll back the entire collection of a massive PR to fix one wrong thing.  Just my 2 cents of the matter.

uberhacker:
An obvious exception to the "rule" is #10246.

Zopolis4:
I'm aiming for 5 per PR, this one had 6 because the version of SCons was too old to work to build the package.

satmandu:
As a followup to the discussion of [how many licks a tootsie pop needs to get to the center](https://www.youtube.com/watch?v=5ZtbCOpx8Sk) I mean how many packages a PR ought to have, may I suggest that this is a structural problem.

We should have a command to list all the downstream packages that are affected by a package change.

Debian and Ubuntu use abidiff from libabigail to handle this, and maybe alongside manifests we should dump libabigail-derived info too to figure out if stuff has sufficiently changed that we should rebuild downstream packages.

Ideally, crew (or a chromebrew tool) should be able to tell us that a package update/rebuild has changed the apis sufficiently that downstream packages that depend upon those should be rebuilt... and here are the packages that as a result need to be updated.

uberhacker:
Is it possible to leverage abidiff in crew?  Could we implement it like Debian-based distros?

Zopolis4:
Although if debian or similar tracks abi differences and things requiring a rebuild, we could just piggyback off of that and just check whether the version we're upgrading to is one that requires rebuilds by looking at the debian database.

PR #10281, opened 2024-08-08
make xz an xz package

satmandu:
- as per prior PRs...


uberhacker (approved):
How can we make sure these stay as .xz instead of .zst on future updates?

satmandu:
> How can we make sure these stay as .xz instead of .zst on future updates?

The `no_zstd` flag handles that.

PR #10283, opened 2024-08-09
Add crew-sudo as a core package on relevant milestones

Zopolis4:
We encounter enough issues where the solution is crew-sudo that it seems not enough people know about this package, so just add it as a core package on the relevant milestones. Sudo support is needed for some packages (i..e docker) anyways.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=3twell crew update
```

### DIFF
--- a/packages/core.rb
+++ b/packages/core.rb
@@ -3,7 +3,7 @@ require 'package'
 class Core < Package
   description 'Core Chromebrew Packages.'
   homepage 'https://github.com/chromebrew/chromebrew'
-  version '2.5'
+  version '2.6'
   license 'GPL-3+'
   compatibility 'all'
 
@@ -17,6 +17,7 @@ class Core < Package
   depends_on 'ca_certificates'
   depends_on 'crew_mvdir'
   depends_on 'crew_profile_base'
+  depends_on 'crew_sudo' if CHROMEOS_RELEASE.to_i > 116
   depends_on 'e2fsprogs'
   depends_on 'elfutils'
   depends_on 'expat'


packages/core.rb
```diff
@@ -17,6 +17,7 @@ class Core < Package
   depends_on 'ca_certificates'
   depends_on 'crew_mvdir'
   depends_on 'crew_profile_base'
+  depends_on 'crew_sudo' if CHROMEOS_RELEASE.to_i > 116
```
uberhacker, line 20: I don't know about this.

Zopolis4: i'm not sure what you mean?

satmandu (commented):
Are there packages that specifically depend upon this for functionality?

If so, it makes sense. (But why not just add it as a dependency then for such systems?)

Zopolis4:
docker, and i think i remember a few others using it. i think it makes more sense to add it as a core dependency because the ability to use sudo is generally expected as a functionality.

satmandu:
> docker, and i think i remember a few others using it. i think it makes more sense to add it as a core dependency because the ability to use sudo is generally expected as a functionality.

Docker doesn't actually run though...

satmandu:
> we can at least make progress on fixing it...

I think docker is fundamentally broken with how google is locking down these devices...

But there are many apps that require root access to function properly... for instance apps like gparted...

satmandu (approved):
Let's see how this works...

PR #10293, opened 2024-08-10
install fixups

satmandu:
- This hopefully fixes installs on current machines, _**which are failing**_ on `x86_64` from the `glibc_lib237` changes and on `armv7l` from the addition of `crew_sudo` to `core.rb`.

- `crew_sudo` installs from `git`, but has no dependencies listed, and also installs from source. This means it installs almost first, before the `git` dependencies are installed, and because it installs from source, `crew` might think it may require `buildessential` too! _I generated binaries for it._ Hopefully this fixes the install on recent arm machines, since otherwise the install suddenly needs to run `git`, but `git`’s dependency `curl` isn’t installed yet...

- `glibc_lib237` appears to be having its postinstall segfaults on `x86_64`, so let us do those postinstall steps more carefully.

- Also modify `install.sh` so a postinsall failure in `BOOTSTRAP_PACKAGES` does not take down the entire install.


supechicken:
Isn't `git` one of the bootstrap packages?

packages/crew_sudo.rb
```diff
@@ -6,9 +8,8 @@ class Crew_sudo < Package
   version '1.1'
   license 'GPL-3'
   compatibility 'all'
-
-  source_url 'https://github.com/chromebrew/crew-sudo.git'
-  git_hashtag "v#{@version}"
+  source_url 'https://github.com/chromebrew/crew-sudo/archive/refs/tags/v1.1.tar.gz'
+  source_sha256 'f71d6a24bc6d1271a7ebbe255a53d08f55fd95d2430d9c14b8d10f82f29c610f'
```
uberhacker, line 12: I also found out there is no crew_sudo.filelist in the manifest directory.  Can we add that too?  Probably the cleanest way to add is at https://github.com/chromebrew/crew-sudo along with a new release.

packages/glibc_lib237.rb
```diff
@@ -77,7 +77,9 @@ def self.postinstall
       if File.file?(File.join(CREW_LIB_PREFIX, 'libC.so.6'))
         Dir.chdir(CREW_LIB_PREFIX) do
           libc_patch_libraries.each do |lib|
-            Kernel.system "patchelf --add-needed libC.so.6 #{lib}" and Kernel.system "patchelf --remove-needed libc.so.6 #{lib}"
+            FileUtils.cp lib, "#{lib}.tmp"
+            Kernel.system "patchelf --add-needed libC.so.6 #{lib}.tmp" and Kernel.system "patchelf --remove-needed libc.so.6 #{lib}.tmp"
+            FileUtils.mv "#{lib}.tmp", lib
```
uberhacker, line 82: Why is a `.tmp` file necessary?

packages/crew_sudo.rb
uberhacker: Actually, you can probably just remove the updates from crew_sudo.rb here and just make them in #10246 instead.

satmandu: Crew_sudo doesn't need binaries.

packages/glibc_lib237.rb
satmandu: Maybe running patchelf on a running binary was causing segfaults....

satmandu:
> Isn't `git` one of the bootstrap packages (if git have deps they should be added to to bootstrap package list also)?

Ah, yes, but curl is the dep it was complaining about, and that has tons of deps...

manifest/armv7l/c/crew_sudo.filelist
```diff
@@ -0,0 +1,59 @@
...
+/usr/local/lib/crew-sudo/.git/objects/9b/08247f6646266d482672f074fcd8048a81f000
+/usr/local/lib/crew-sudo/.git/objects/aa/0b8edd03cbe3ed4954c6f48ae5619b3071b0c5
+/usr/local/lib/crew-sudo/.git/objects/bb/1aa3cad22c877af8357be686f44e4aa07420ed
+/usr/local/lib/crew-sudo/.git/objects/bc/5d987d52abadbe69ceff8d622ff5a22a7c0a69
+/usr/local/lib/crew-sudo/.git/objects/c4/ab9ed0894d37355cdf6fea1aa7bd6d8f49bd6c
+/usr/local/lib/crew-sudo/.git/objects/ec/05ac75a5033312bcb29d0a815663689240737a
+/usr/local/lib/crew-sudo/.git/objects/f2/88702d2fa16d3cdf0035b15a9fcbc552cd88e7
+/usr/local/lib/crew-sudo/.git/objects/f2/98aed20e064e68611ed0160ee0d993ffe26753
+/usr/local/lib/crew-sudo/.git/objects/f8/d717d879ee3fc1e26b3e7a0e6dbdabfa0fef13
+/usr/local/lib/crew-sudo/.git/shallow
```
uberhacker, line 46: We wouldn't need the .git directory if we used the source download to build, right?

satmandu: We should modify that to remove that... let me do that...

packages/crew_sudo.rb
```diff
@@ -6,14 +6,19 @@ class Crew_sudo < Package
   version '1.1'
   license 'GPL-3'
   compatibility 'all'
```
uberhacker, line 8: `compatibility 'x86_64 aarch64 armv7l'`?

uberhacker (approved):

PR #10294, opened 2024-08-11
Refactor ruby gems in packages, crew cleanup of debugging, install.sh refactor for gems

### LINKED ISSUE #10295: 🐞 Bug Checksum mismatch :/ Try again?
**Describe the bug**
When trying to install the GParted on chromebrew I get the above error.  

**To Reproduce**
Steps to reproduce the behavior: Try installing GParted and the error might  reproduce itself.

My environment was (crosh, bash or VT-2): These commands where executed both in the VT-2 shell and the crosh shell.

      ```shell
      # crew install gparted
      ```

      Output:
`gparted: A Partition Magic clone, frontend to GNU Parted
https://gparted.org/
Version: 1.6.0
License: GPL-2+ and FDL-1.2+
parted libsigcplusplus glibmm_2_4 atkmm16 cairomm_1_0 librsvg libdecor libxtrans libfontenc libxfont2 libxfont libxkbfile rendercheck xwayland sommelier gtk3 pangomm_1_4 gtkmm3 libinih liburcu xfsprogs exfatprogs f2fs_tools gpart mtools
Do you agree? [Y/n] y
Proceeding...
parted: Create, destroy, resize, check, copy partitions and file systems.
Checksum mismatch :/ Try again?

                      Filename: parted-3.6-chromeos-x86_64.tar.zst
    Expected checksum (SHA256): 9f5c1053aa58626f1d6d3cdad28957c1dbc4e417c1e3d3a3af06160433ddda84
  Calculated checksum (SHA256): 01d8bdaf0944aa74d362c4fd00de3fbf2f0c006e74093afe391e4f99bdbb29ea`

**Expected behavior**
I expected it to install.

**Screenshots**
image

***
- User space: `64-bit`
- Chromebrew Kernel version: `6.6`

- Chromebrew version: `1.50.1`

- Last update in local repository: 8dbc3925d `icu4c => 75.1 + updates (#10246) (17 hours ago)`

- OS variant: `Chrome OS`
- OS version: `brya-release/R126-15886.74.0`

***
I think the checksums have to be updated. Is there a way I can do that.

satmandu:
Fixes #10295 
- Use highline for more prompts in `crew`
- Refactored gem loading in `crew`, and move require of gems to where they are needed, so we don't have to load them early in crew when they might even be needed.
  - The rationale here is that `crew` shouldn't have to load gems before they are installed properly.
  - But also modify `install.sh` to load gem binaries early, not build them, since building them requires `buildessential` to be installed. ( And while we are at it, let's add the gem info to `device.json`.
- Largely refactors ruby gem installs through crew.
- Uses git to check for the last time `crew update` was run, and allows a gem update check every two weeks when `crew update` is run, and also prompts to run `crew update` if it has been more than a week since the last time `crew update` was run. There is also now a `CREW_UPDATE_CHECK_INTERVAL` const for this that can be set, defaulting to 7 days.
image
- Also adds a `time_difference` function to crew to make it easier to report times, which is also used for the reports of how long installs take...
- Adjusts crew to call remove `def self.remove` and changes the old sections to be `def self.postremove`.
- Expand gems installed, properly add `pry` gem, and move the `~.pryrc` that stores the single character step/next shortcuts intto the new `ruby_pry` gem package.
- Refactor `upload` in `crew` to better handle replacing existing uploads.
- Have `buildsystems/ruby` only run a gem uninstall during install if the gem is already installed.
- Add missing `librsvg` binaries.
- Looks like we don't need to run `crew update compatible` in `install.sh`...
- `Install.sh` tested in `i686` container.
- Adds builds for `webkitgtk_6`.
- Add `update_ruby_gem_packages.rb` script in `tools/` to update the `ruby_` gem packages.
  - @Zopolis4 Can we have a scheduled task to run something like this routinely and automatically generate PRs at some point?
- Rebuilds `parted`.
- Updates `libsdl2`.
- Updates `freerdp`.

##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
##
##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=webkit crew update \
&& yes | crew upgrade
```


satmandu:
I guess the question is whether adding `ruby_highline` to `core.rb` is going to cause a problem in installs due to buildessential not being installed beforehand...

bin/crew
```diff
@@ -1540,38 +1538,34 @@ def upload(pkg_name = nil, pkg_version = nil, gitlab_token = nil, binary_compres
...
-        begin
-          gem 'highline'
-        rescue Gem::LoadError
-          puts ' -> install gem highline...'
-          Gem.install('highline')
-          gem 'highline'
+        puts "\n#{new_tarfile} has already been uploaded.\nPlease change the #{package} package version from #{new_version} and try again.\n".lightred
+        require_gem('highline/import')
+        unless agree('Do you want to overwrite the existing upload instead? ')
+          puts 'Will NOT overwite the existing upload.'.lightred
```
uberhacker, line 1546: Not an error.  Use `.orange` for warning instead?

install.sh
```diff
@@ -270,6 +272,23 @@ function update_device_json () {
...
+  for gem in "$@"; do
+    ruby_gem="${gem}"
+    echo_intra "Installing ${ruby_gem} gem..."
+    gem install -N "${ruby_gem}" --conservative
+    gem_version="$(gem list "${ruby_gem}" | grep "${ruby_gem}" | awk '{print $2}' | tr -d '()')"
+    json_gem_version="${gem_version}-ruby-${rubymajorversion}"
+    crew_gem_package="ruby_${ruby_gem//-/_}"
+    update_device_json "${crew_gem_package}" "${json_gem_version}" ""
+    echo_success "${ruby_gem^} gem installed."
+    BOOTSTRAP_PACKAGES+=" ${crew_gem_package} "
```
uberhacker, line 288: No need for trailing space?
```
 BOOTSTRAP_PACKAGES+=" ${crew_gem_package}"
```

install.sh
```diff
@@ -308,18 +327,12 @@ sudo ldconfig &> /tmp/crew_ldconfig || true
...
 
-# Install activesupport gem for ruby
-echo_info 'Installing essential ruby gems.'
-gem update -N --system
-gem install -N activesupport --conservative
-gem install -N concurrent-ruby --conservative
-
-# Since we downloaded the package repo, just update package compatibility information.
-crew update compatible
+echo_info 'Installing essential ruby gems...\n'
```
uberhacker, line 332: Does `\n` inside single quotes print the literal and not a linefeed?

install.sh
```diff
@@ -308,18 +327,12 @@ sudo ldconfig &> /tmp/crew_ldconfig || true
...
+BOOTSTRAP_GEMS='activesupport concurrent-ruby highline'
+# shellcheck disable=SC2086
+install_ruby_gem ${BOOTSTRAP_GEMS}
```
uberhacker, line 335: Should these be packages and added to core.rb instead?

uberhacker: Oh wait, they already are packages?  Why do we need to do this then?

satmandu: They are packages already using the `buildsystems/ruby` gem install system. This is just a way to force the install early without building from source, which using the ruby buildsystem forces, at least on i686.

install.sh
satmandu: Fixed.

tools/update_ruby_gem_packages.rb
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env ruby
+# update_ruby_gem_packages version 1.0 (for Chromebrew)
+# Author: Satadru Pramanik (satmandu) satadru at gmail dot com
+# Usage in packages dir: ../tools/update_ruby_gem_packages.rb
+Dir['ruby_*.rb'].each do |package|
+  gem_name = package.gsub('.rb', '').gsub('ruby_', '').gsub('_', '-')
+  gem_version = Gem.latest_spec_for(gem_name).version.to_s
+  puts "Updating #{gem_name} package file to #{gem_version}"
+  system "sed -i \"s,.*version.*,\ \ version '#{gem_version}-ruby-#{RUBY_VERSION.slice(/(?:.*(?=\.))/)}',\" #{package}"
+end
```
uberhacker, line 10: Could this be a crew command instead of a script?  Seems like something we could use often.

uberhacker: Maybe it makes sense to integrate into the `crew upgrade` process?

satmandu: Yes, in theory it could be a part of `crew update` but that would really slow that down.

This is better done in batch mode occasionally externally maybe?

bin/crew
```diff
@@ -1643,17 +1637,11 @@ def copy_package(pkg_name, prompt_msg = '')
...
     operation = caller_locations(1, 2)[1].to_s.split[3].split('_')[0]
     puts prompt_msg.yellow
-    print "\nWould you like to copy #{pkg_name}.rb to crew and start the #{operation}? [Y/n] ".yellow
-    response = $stdin.gets.chomp.downcase
-    case response
-    when 'n', 'no'
-      puts "#{operation.capitalize} skipped."
-      next_pkg = true
-    when '', 'y', 'yes'
+    require_gem('highline/import')
```
uberhacker, line 1640: Should this be at the top of the file since it is used multiple times?  Also, I see it referenced twice here and on line 1236.

install.sh
uberhacker: Maybe we need to rethink the ruby packages then?  It sort of makes sense to let ruby handle the package management heavy-lifting with a minimal crew wrapper.

tools/update_ruby_gem_packages.rb
satmandu: But I do think adding more to crew for that might be another PR... And you then also get sync issues with packages in the repo... So maybe we think about ignoring the versioning entirely, and only looking for version numbers during `crew update` for installed packages.

bin/crew

bin/crew
```diff
@@ -52,6 +57,8 @@ require_relative '../lib/package'
 require_relative '../lib/package_utils'
 require_relative '../lib/util'
 
+require_gem('highline/import')
```
uberhacker, line 60: Would it make sense to group all the require* lines together above everything at the top?  It looks a bit disorganized when spread out like this.

bin/crew
```diff
@@ -52,6 +57,8 @@ require_relative '../lib/package'
...
+
 # Disallow sudo
 abort 'Chromebrew should not be run as root.'.lightred if Process.uid.zero?
```
uberhacker, line 63: This should be above everything else and just below the require* lines too.  No need to do any processing if we need to exit early.

bin/crew
satmandu: No. I want the debug require pry-byrbug to come as early as possible, right after const.rb is loaded.

bin/crew
```diff
@@ -1,16 +1,34 @@
...
+    Gem.install(gem_name)
+    gem gem_name
+  end
+  requires = if require_override.nil?
+               gem_name_and_require.split('/')[1].nil? ? gem_name_and_require.split('/')[0] : gem_name_and_require
+             else
+               require_override
+             end
+  require requires
 end
```
uberhacker, line 23: This def block can be below all the require* lines, correct?

satmandu: It's up there on purpose. The color require is needed in that def block for `require_gem`, and `require_gem` enables the three gems below that, which are needed to enable debugging shortly below where the debugging enabledment is right after `const.rb` is loaded.

satmandu: The goal is to allow exit as soon as possible if run with sudo... ;)

install.sh
```diff
@@ -197,8 +199,9 @@ fi
 cd "${CREW_CONFIG_PATH}"
 echo_info "\nCreating device.json."
 jq --arg key0 'architecture' --arg value0 "${ARCH}" \
-  --arg key1 'installed_packages' \
-  '. | .[$key0]=$value0 | .[$key1]=[]' <<<'{}' > device.json
+  --arg key1 'last_update_check' --arg value1 "`date +%s`" \
+  --arg key2 'installed_packages' \
```
uberhacker, line 203: Rubocop doesn't like backticks.

satmandu: Shellcheck!

packages/crew_launcher.rb
```diff
@@ -34,7 +34,7 @@ def self.install
     EOF
   end
 
-  def self.remove
+  def self.postremove
     FileUtils.rm_rf "#{CREW_PREFIX}/share/crew-launcher/"
```
uberhacker, line 38: Should this operation have a prompt?

install.sh
satmandu: Fixed.

packages/crew_launcher.rb
satmandu: Prompt for all postremove sections?

packages/nginx.rb
```diff
@@ -79,7 +79,7 @@ def self.postinstall
     EOM
   end
 
-  def self.remove
+  def self.postremove
     if Dir.exist? "#{CREW_PREFIX}/share/nginx"
       puts "\nWARNING: This will remove all hosting files and configuration.".orange
       print "Would you like to remove #{CREW_PREFIX}/share/nginx? [y/N] "
```
uberhacker, line 85: Use `agree(...)`?

packages/netbeans.rb
```diff
@@ -27,7 +27,7 @@ def self.postinstall
     puts "\nType 'netbeans' to get started.\n".lightblue
   end
 
-  def self.remove
+  def self.postremove
     config_dir = "#{HOME}/.netbeans"
     if Dir.exist? config_dir
       print "Would you like to remove the config directory #{config_dir}? [y/N] "
```
uberhacker, line 33: Use `agree(...)`?

packages/natron.rb
```diff
@@ -33,7 +33,7 @@ def self.postinstall
     puts "\nType 'natron' to get started.\n".lightblue
   end
 
-  def self.remove
+  def self.postremove
     config_dir = "#{HOME}/.Natron"
     if Dir.exist? config_dir
       print "Would you like to remove the config directory #{config_dir}? [y/N] "
```
uberhacker, line 39: Use `agree(...)`?

tools/update_ruby_gem_packages.rb
uberhacker: Is this still needed since we're updating gems weekly in crew now?

uberhacker: Actually, now that I think about it, maybe it makes sense to set an environment variable or constant for the update period instead of hard coding it to a week?  That way we can set the duration period to zero if we want to update right away.

bin/crew
uberhacker: Fair enough.

bin/crew
```diff
@@ -285,6 +282,11 @@ def update
     system 'git-restore-mtime -sq 2>/dev/null', exception: true if File.file?("#{CREW_PREFIX}/bin/git-restore-mtime")
   end
 
+  # Only update gems once a week.
+  if Time.now.to_i - @last_update_check > 604800
```
uberhacker, line 286: Make the 604800 period configurable with a variable?

install.sh
uberhacker: Yes, shellcheck instead of rubocop since this is a shell script. :)

lib/const.rb
```diff
@@ -130,6 +130,9 @@
 # If CREW_DISABLE_MVDIR environment variable exists and is equal to 1 use rsync/tar to install files in lieu of crew-mvdir.
 CREW_DISABLE_MVDIR = ENV.fetch('CREW_DISABLE_MVDIR', '0').eql?('1')
 
+# Days between crew prompting to run 'crew update' and doing gem updates.
+CREW_UPDATE_CHECK_INTERVAL = 7
```
uberhacker, line 134: Shouldn't this be set based on whether an environment variable is set?
```
CREW_UPDATE_CHECK_INTERVAL = ENV.fetch('CREW_UPDATE_CHECK_INTERVAL',  7)
```

satmandu: I was considering that!

satmandu: Done.

bin/crew
```diff
@@ -132,14 +137,6 @@ class ExitMessage
...
 at_exit do
-  begin
-    gem 'activesupport'
-  rescue Gem::LoadError
-    puts ' -> install gem activesupport'
-    Gem.install('activesupport')
-    gem 'activesupport'
-  end
-  require 'active_support/core_ext/object/blank'
   GnomePostinstall.run unless GnomePostinstall.gnome_packages.blank?
```
Zopolis4, line 140: can't we just use `.empty?` here

bin/crew
```diff
@@ -1922,5 +1925,14 @@ Signal.trap('INT') do
...
 
 load_json
+if @device[:last_update_check].blank?
+  @device.store('last_update_check', '0')
+  @last_update_check = 0
+  save_json(@device)
+else
+  @last_update_check = @device[:last_update_check]
+  crewlog("The last update was #{time_difference(@last_update_check, Time.now.to_i)} ago.")
+  puts "It has been more than #{CREW_UPDATE_CHECK_INTERVAL} day#{CREW_UPDATE_CHECK_INTERVAL < 2 ? '' : 's'} since crew was last updated. Please run 'crew update'".lightpurple if Time.now.to_i - @last_update_check > (CREW_UPDATE_CHECK_INTERVAL * 3600 * 24)
```
Zopolis4, line 1935: Can't we just check the last time the git repository was fetched?

bin/crew
```diff
@@ -1230,18 +1238,10 @@ def resolve_dependencies
 
   if @opt_force
     puts 'Proceeding with dependency package installation...'.orange
+  elsif agree('Do you agree? ')
```
Zopolis4, line 1241: This feels like we're adding a large gem dependency and significantly increasing the dependencies and complexity of bin/crew in order to avoid 9 lines of code-- couldn't we just take the existing input prompt code we have and split it out into a file in `lib`, then call that?

bin/crew
satmandu: The great thing about `blank?` is that it avoids `NoMethodError`s.

See https://stackoverflow.com/a/26877095

satmandu: And yes, I used to think the same thing, and then just got sick of it and accepted that adding `.blank?` from the `activesupport` gem really makes life so much easier.

bin/crew
satmandu: Ah!

I guess we could use `stat -c %Y .git/FETCH_HEAD` ?

bin/crew
satmandu: We could, but highline is so much better at asking for an answer than our code and makes our code simpler. Should we really be worried about loading an extra gem for a single threaded app on standalone systems, especially if it makes our code much more readable?

Look at it:

https://github.com/JEG2/highline

Zopolis4: Our code isn't being made overall simpler by having to work around a gem dependency-- If we were using it in tons of places for tons of things, sure, but we're only using it for an input prompt, which we already have the code for.

I just don't see the value in doing all this compared to splitting out the 10 lines of code into a helper file and calling that.

satmandu:
I removed the json changes, as we can much more simply get the last update time from looking at git, specifically `CREW_LIB_PATH/.git/FETCH_HEAD` as @Zopolis4 suggested.

bin/crew
satmandu: I guess in part this is a stylistic question? I really don't like our prompt code. I think it is fragile and doesn't work well, and as such I'm happy to outsource to a gem that does it better.

The "I don't understand" bit there now is especially annoying because clearly there is code that can do a better job at understanding edge case inputs.

But of course I'm happy to let other people vote as to whether they think this is a better approach.

**My TL;DR here would be:** _It's ok if a slight increase in complexity much improves usability._

tools/update_ruby_gem_packages.rb
satmandu: This is also what I was thinking as I went to bed... so I changed it accordingly.

bin/crew
uberhacker: We can (and probably should) potentially change to use `agree(...)` in tons of packages in the postremove block.  In agreement with @satmandu here.

tools/update_ruby_gem_packages.rb
satmandu: 100%. If `crew update` is our proxy for version checks for our packages, we need to be automating getting version updates into our packages, especially for packages that we don't have to build ourselves, as that's low hanging fruit...

uberhacker (approved):

uberhacker (approved):

uberhacker (approved):

Zopolis4:
This breaks `tools/json.rb`: 
image

Zopolis4:
actually presumably the solution would be to just install highline in the action. i'll get to that in a bit

PR #10309, opened 2024-08-13
Use grep better to speed up crew remove of packages.

### LINKED ISSUE #10102: Significantly slower crew remove performance
The current method of checking for conflicts when removing a package is much slower than the previous version, probably because we use ruby to iterate through the file rather than passing it to grep. 

Master:
```
$ time crew remove qt5_base
qt5_base removed!

real    0m17.970s
user    0m8.586s
sys     0m9.207s
```

Before #9529:
```
$ time crew remove qt5_base
Checking for conflicts of qt5_base with files from installed packages...
Qt5_base removed!

real    0m1.669s
user    0m1.021s
sys     0m0.278s
```

satmandu:
Fixes #10102 
- Uses grep features to speed up package removal by roughly 25x.
- Simplifies `@last_update_check` in `crew` to use `File.mtime`.
- Cleanup in `package.rb`, fixing spelling and also fixing `.load_package` so `.get_deps_list` works when invoked from places like `remove.rb`
- Moves json functions out to `PackageUtils` so they can be called from other files. If this isn't the best place for it happy for it to go elsewhere.
- Stores `:essential_deps` in json since it only needs to be regenerated when `generate_compatible` happens.
- Refactors `command/remove.rb` slightly
- Finishes process of GitHub action CI Unit Tests... moving as much as possible to `tests/unit_tests.sh` since it is a PITA to debug one long line of shell code.
- Adds a unit test for `remove.rb` and put it in `tests/unit_tests.sh`
- Fixes Docopt so things like `crew -d <command>` work. (Every line needed `[options]`.) 
  - Yes Docopt could use tuneups, but this was a regression.) We should probably have a unit test for docopt... 
  - And maybe move to `docopt_ng`, which is still updated.

##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
##
##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=slow_remove crew update \
&& yes | crew upgrade ; crew install xxd_standalone ; crew remove xxd_standalone -v ; crew install glibc_lib ; crew remove glibc_lib
```
This gives:
```
Xxd_standalone installed!
/usr/local/bin/crew:1928:<main>:The last update was 9 seconds ago.
Removing file /usr/local/bin/xxd
Removing file /usr/local/share/man/man1/xxd.1.zst
Removing package xxd_standalone from device.json
/usr/local/bin/crew:156:save_json:Saving device.json...
xxd_standalone removed!
chronos@mbp113-i686 /usr/local/lib/crew/packages (master >|SPARSE)$
```

Also:
```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=slow_remove crew update \
&& yes | crew upgrade ; yes | crew install qt5_base ; time crew remove qt5_base | tail
```
gives me this:
```
qt5_base removed!

real    0m0.972s
user    0m0.784s
sys     0m0.146s
```


commands/remove.rb
```diff
@@ -34,16 +34,30 @@ def self.remove(pkg, verbose)
...
+
+          unique_to_package_files = package_files.flatten - all_other_files.flatten
+          package_files_that_overlap = all_other_files.flatten & package_files.flatten
+
+          unless package_files_that_overlap.flatten.empty?
+            puts "The following file(s) in other packages will not be deleted during the removal of #{pkg.name}:".orange
+            puts package_files_that_overlap.join("\n").orange
+          end
+          unique_to_package_files.flatten.each do |file|
+            if CREW_ESSENTIAL_FILES.include?(File.basename(file))
```
Zopolis4, line 56: Couldn't we check if the package is included in the dependencies of `CREW_ESSENTIAL_PACKAGES`? i.e. using `pkg.get_deps_list`

or, given that the dependency problem only applies to `glibc_*` to my knowledge, we could just add a wildcard check for that

satmandu: I think that would be ideal!

Can you figure out how to do that properly?

I couldn't figure out how to use that to pull the package d'EPS of stuff like glibc_lib.rb from inside remove.rb...

supechicken: > Seriously though, how is one supposed to invoke `pkg.get_deps_list` for other packages from elsewhere in `crew`?

We can do it with one line:
```ruby
Package.load_package('pkg.rb').get_deps_list
```

satmandu: @supechicken Looks like that works if I make sure `pkg_file` includes the path to the file. 

E.g., it works if I change `self.load_package` in `package.rb` to have this before `class_eval(File.read(pkg_file`:
```
    # read and eval package script under 'Package' class, using the
    # newest file available.
    pkg_file = ["#{CREW_LOCAL_REPO_ROOT}/packages/#{pkg_name}.rb", "#{CREW_PACKAGES_PATH}/#{pkg_name}.rb"].max { |a, b| File.mtime(a) <=> File.mtime(b) }

```

commands/remove.rb
```diff
@@ -19,6 +19,13 @@ def self.remove(pkg, verbose)
       return
     end
 
+    # List dependencies of packages in CREW_ESSENTIAL_PACKAGES
+    essential_deps = ''
+    CREW_ESSENTIAL_PACKAGES.each do |pkg|
+      essential_deps += Package.load_package("#{pkg}.rb").get_deps_list
+    end
+    puts "essential_deps are #{essential_deps}".orange
```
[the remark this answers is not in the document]
satmandu: Yes, was still debugging...

commands/remove.rb
satmandu: Ok I got it working.

uberhacker:
I noticed a lot of the glibc* packages have "conflicts" so those would be a good test scenario.

satmandu:
> I noticed a lot of the glibc* packages have "conflicts" so those would be a good test scenario.

Looks like that works:
image

uberhacker:
> > I noticed a lot of the glibc* packages have "conflicts" so those would be a good test scenario.
> 
> Looks like that works: image

What happens with glibc_dev237?

satmandu:
> > > I noticed a lot of the glibc* packages have "conflicts" so those would be a good test scenario.
> > 
> > 
> > Looks like that works: image
> 
> What happens with glibc_dev237?

```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=slow_remove crew update \
&& yes | crew upgrade ; crew install glibc_dev237 ; crew remove glibc_dev237
```
Gives this:
image

(That appears to be what we want.)

commands/remove.rb
```diff
@@ -34,17 +44,46 @@ def self.remove(pkg, verbose)
...
+
+          # When making a list of all files from crew filelists we again
+          # ignore all files from packages and dependent packages of
+          # CREW_ESSENTIAL_PACKAGES.
+          essential_deps_excludes = essential_deps.map { |i| File.file?("#{File.join(CREW_META_PATH, i.to_s)}.filelist") ? "--exclude=#{File.join(CREW_META_PATH, i.to_s)}.filelist" : '' }.join(' ')
+
+          package_files_cmd = "grep -h #{essential_deps_exclude_froms} \"^#{CREW_PREFIX}\\|^#{HOME}\" #{flist}"
+          package_files = `#{package_files_cmd}`.split("\n").map(&:to_s)
+          package_files.uniq!
+          package_files.sort!
```
uberhacker, line 68: Can this be a chained command instead of separate lines?
```
package_files = `#{package_files_cmd}`.split("\n").map(&:to_s).uniq.sort!
```

commands/remove.rb
```diff
@@ -34,17 +44,46 @@ def self.remove(pkg, verbose)
...
+          all_other_files_cmd = "grep -h --exclude #{pkg.name}.filelist #{essential_deps_excludes} \"^#{CREW_PREFIX}\\|^#{HOME}\" #{CREW_META_PATH}/*.filelist"
+          all_other_files = `#{all_other_files_cmd}`.split("\n").map(&:to_s)
+          all_other_files.uniq!
+          all_other_files.sort!
```
uberhacker, line 72: Can this be a chained command instead of separate lines?
```
all_other_files = `#{all_other_files_cmd}`.split("\n").map(&:to_s).uniq.sort!
```

lib/package.rb
```diff
@@ -48,9 +51,9 @@ def self.load_package(pkg_file)
   end
 
   def self.dependencies
-    # We need instance variable in derived class, so not define it here,
-    # base class.  Instead of define it, we initialize it in a function
-    # called from derived classees.
+    # We need instance variable in derived class, so do not define it here,
+    # base class.  Instead of defining it, we initialize it in a function
+    # called from derived classes.
```
uberhacker, line 56: Kudos on attention to detail with the grammar and spell check.

lib/package.rb
```diff
@@ -36,7 +36,10 @@ def self.load_package(pkg_file)
     pkg_name = File.basename(pkg_file, '.rb')
     class_name = pkg_name.capitalize
 
-    # read and eval package script under 'Package' class
+    # Read and eval package script under 'Package' class, using the
+    # newest file available.
+    pkg_file = File.file?("#{CREW_LOCAL_REPO_ROOT}/packages/#{pkg_name}.rb") ? ["#{CREW_LOCAL_REPO_ROOT}/packages/#{pkg_name}.rb", "#{CREW_PACKAGES_PATH}/#{pkg_name}.rb"].max { |a, b| File.mtime(a) <=> File.mtime(b) } : "#{CREW_PACKAGES_PATH}/#{pkg_name}.rb"
```
supechicken, line 41: Simplify with `Dir.glob`?
```ruby
# Dir[] will also check the existence of file
pkg_file = Dir["/{#{CREW_LOCAL_REPO_ROOT}/packages,#{CREW_PACKAGES_PATH}}/#{pkg_name}.rb"].max { |a, b| File.mtime(a) <=> File.mtime(b) }
```

commands/remove.rb
supechicken: Also we don't need `.map(&:to_s)` here as the output of `.split("\n")` must be strings

commands/remove.rb
```diff
@@ -13,9 +13,19 @@ def self.remove(pkg, verbose)
...
-    # Don't remove any of the packages ruby (and thus crew) needs to run.
-    if CREW_ESSENTIAL_PACKAGES.include?(pkg.name)
-      puts "Refusing to remove essential package #{pkg.name}.".lightred
+    # Determine dependencies of packages in CREW_ESSENTIAL_PACKAGES, as
+    # those are needed for ruby and crew to run, and thus should not be
+    # removed.
+    essential_deps = CREW_ESSENTIAL_PACKAGES.map do |pkg|
+      Package.load_package("#{pkg}.rb").get_deps_list
+    end
+    essential_deps.flatten!.uniq!
```
supechicken, line 22: Use `.flat_map` on the loop above?
```ruby
essential_deps = CREW_ESSENTIAL_PACKAGES.flat_map do |pkg|
  Package.load_package("#{pkg}.rb").get_deps_list
end.uniq
```

commands/remove.rb
satmandu: yes but uniq and sort can't be combined.

lib/package.rb
satmandu: The privilege of growing up immersed in it...

lib/package.rb
satmandu: That doesn't work when `CREW_LOCAL_REPO_ROOT` is empty.

```ruby
pkg_name = 'ninja'
CREW_LOCAL_REPO_ROOT = '/output/git/satmandu'
CREW_PACKAGES_PATH = '/usr/local/lib/crew/packages'
Dir["#{CREW_LOCAL_REPO_ROOT}/packages/#{pkg_name}.rb","#{CREW_PACKAGES_PATH}}/#{pkg_name}.rb"].max { |a, b| File.mtime(a) <=> File.mtime(b) }
```
gives:
```ruby
=> "/output/git/satmandu/packages/ninja.rb"
```
But
```ruby
pkg_name = 'ninja'
CREW_LOCAL_REPO_ROOT = ''
CREW_PACKAGES_PATH = '/usr/local/lib/crew/packages'
Dir["#{CREW_LOCAL_REPO_ROOT}/packages/#{pkg_name}.rb","#{CREW_PACKAGES_PATH}}/#{pkg_name}.rb"].max { |a, b| File.mtime(a) <=> File.mtime(b) }
```
gives:
```ruby
=> nil
```

supechicken: You have an excess bracket here...
```ruby
"#{CREW_PACKAGES_PATH}>>>}<<</#{pkg_name}.rb"
```

satmandu:
That last change broke essential package protections...


```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=slow_remove crew update && yes | crew upgrade ; crew remove  glibc_lib237
From https://github.com/satmandu/chromebrew
 * branch                slow_remove -> FETCH_HEAD
HEAD is now at a6f4ba6d5 suggested changes
Your software is up to date.
Your software is already up to date.
The following file(s) in other packages will not be deleted during the removal of glibc_lib237:
/usr/local/bin/zdump
glibc_lib237 removed!
```

satmandu:
We should have a unit test to check for that...

satmandu:
Maybe also we should have an `exit 1` if an avoided remove is not in a pkg upgrade situation...

commands/remove.rb
```diff
@@ -13,9 +13,18 @@ def self.remove(pkg, verbose)
...
-    # Don't remove any of the packages ruby (and thus crew) needs to run.
-    if CREW_ESSENTIAL_PACKAGES.include?(pkg.name)
-      puts "Refusing to remove essential package #{pkg.name}.".lightred
+    # Determine dependencies of packages in CREW_ESSENTIAL_PACKAGES, as
+    # those are needed for ruby and crew to run, and thus should not be
+    # removed.
+    essential_deps = CREW_ESSENTIAL_PACKAGES.flat_map.uniq do |pkg|
+      Package.load_package("#{pkg}.rb").get_deps_list
+    end
+    if essential_deps.include?(pkg.name)
```
supechicken, line 22: We should check for the package that we are currently operating also, not just its dependencies

satmandu:
Ok that is fixed...

```
chronos@cheekon-x86_64 /output/git/satmandu (slow_remove *+)$ crew remove glibc_lib
  Glibc_lib is considered an essential package needed for
  Chromebrew to function and thus cannot be removed.

chronos@cheekon-x86_64 /output/git/satmandu (slow_remove *+)$
```

commands/remove.rb
satmandu: ok hopefully this makes more sense:
```ruby
    # Determine dependencies of packages in CREW_ESSENTIAL_PACKAGES and
    # their dependencies, as those are needed for ruby and crew to run,
    # and thus should not be removed.
    essential_deps = CREW_ESSENTIAL_PACKAGES.flat_map { |essential_pkg| Package.load_package("#{essential_pkg}.rb").get_deps_list }.push(*CREW_ESSENTIAL_PACKAGES).uniq.sort
    essential_deps2 = essential_deps
    loops = 0
    until loops == 5
      essential_deps = essential_deps2.flat_map { |essential_pkg| Package.load_package("#{essential_pkg}.rb").get_deps_list }.push(*essential_deps).uniq.sort

      break if essential_deps == essential_deps2
      essential_deps2 = essential_deps
      loops += 1
    end

    if essential_deps.include?(pkg.name)
      puts <<~ESSENTIAL_PACKAGE_WARNING_EOF.gsub(/^(?=\w)/, '  ').lightred
        #{pkg.name.capitalize} is considered an essential package needed for
        Chromebrew to function and thus cannot be removed.
      ESSENTIAL_PACKAGE_WARNING_EOF

      return
    end
```

satmandu: Let me know if there is a more efficient way of doing this.

lib/const.rb
```diff
@@ -307,25 +307,25 @@
...
-    crew prop [<property>]
+    crew prop [options] [<property>]
     crew reinstall [options] [-k|--keep] [-s|--source] [-S|--recursive-build] [-v|--verbose] <name> ...
-    crew remove [-v|--verbose] <name> ...
-    crew search [-v|--verbose] <name> ...
-    crew sysinfo [-v|--verbose]
-    crew test [-v|--verbose] [<name> ...]
+    crew remove [options] [-v|--verbose] <name> ...
+    crew search [options] [-v|--verbose] <name> ...
+    crew sysinfo [options] [-v|--verbose]
```
Zopolis4, line 323: Why are we adding all these back? For example, `crew sysinfo` certainly does not take any other options than the ones already listed.

Zopolis4 (dismissed):

commands/remove.rb
```diff
@@ -13,10 +14,20 @@ def self.remove(pkg, verbose)
...
+    # and thus should not be removed.
+    essential_deps = CREW_ESSENTIAL_PACKAGES.flat_map { |essential_pkg| Package.load_package("#{essential_pkg}.rb").get_deps_list }.push(*CREW_ESSENTIAL_PACKAGES).uniq.sort
+
+    if essential_deps.include?(pkg.name)
+      return if Package.load_package("#{pkg.name}.rb").in_upgrade
+
+      puts <<~ESSENTIAL_PACKAGE_WARNING_EOF.gsub(/^(?=\w)/, '  ').lightred
+        #{pkg.name.capitalize} is considered an essential package needed for
+        Chromebrew to function and thus cannot be removed.
+      ESSENTIAL_PACKAGE_WARNING_EOF
```
Zopolis4, line 28: I'm confused-- why can't we just use one or to `puts` lines here?

commands/remove.rb
```diff
@@ -13,10 +14,20 @@ def self.remove(pkg, verbose)
...
-    if CREW_ESSENTIAL_PACKAGES.include?(pkg.name)
-      puts "Refusing to remove essential package #{pkg.name}.".lightred
-      return
+    # Determine dependencies of packages in CREW_ESSENTIAL_PACKAGES and
+    # their dependencies, as those are needed for ruby and crew to run,
+    # and thus should not be removed.
+    essential_deps = CREW_ESSENTIAL_PACKAGES.flat_map { |essential_pkg| Package.load_package("#{essential_pkg}.rb").get_deps_list }.push(*CREW_ESSENTIAL_PACKAGES).uniq.sort
+
+    if essential_deps.include?(pkg.name)
+      return if Package.load_package("#{pkg.name}.rb").in_upgrade
```
Zopolis4, line 23: we shouldn't need to load the package again, the loaded package object was already passed to `Command.remove`

```suggestion
      return if pkg.in_upgrade
```

commands/remove.rb
```diff
@@ -34,17 +45,42 @@ def self.remove(pkg, verbose)
     # Remove the files and directories installed by the package.
     unless pkg.is_fake?
       Dir.chdir CREW_CONFIG_PATH do
-        # Remove all files installed by the package.
+        # Remove all files installed by the package in CREW_PREFIX and
+        # HOME.
```
Zopolis4, line 49: Won't this conflict with our current config directory code which allows the user to choose if they want to remove it or not?

commands/remove.rb
```diff
@@ -34,17 +45,42 @@ def self.remove(pkg, verbose)
...
+          # all packages and dependent packages of CREW_ESSENTIAL_PACKAGES.
+          essential_deps_exclude_froms = essential_deps.map { |i| File.file?("#{File.join(CREW_META_PATH, i.to_s)}.filelist") ? "--exclude-from=#{File.join(CREW_META_PATH, i.to_s)}.filelist" : '' }.join(' ')
+
+          # When making a list of all files from crew filelists we again
+          # ignore all files from packages and dependent packages of
+          # CREW_ESSENTIAL_PACKAGES.
+          essential_deps_excludes = essential_deps.map { |i| File.file?("#{File.join(CREW_META_PATH, i.to_s)}.filelist") ? "--exclude=#{File.join(CREW_META_PATH, i.to_s)}.filelist" : '' }.join(' ')
+
+          package_files_cmd = "grep -h #{essential_deps_exclude_froms} \"^#{CREW_PREFIX}\\|^#{HOME}\" #{flist}"
+          package_files = `#{package_files_cmd}`.split("\n").uniq.sort
```
Zopolis4, line 67: Is there any reason to have the `package_files_cmd` variable here instead of just doing it in one line?
i.e.
```
package_files = `#{"grep -h #{essential_deps_exclude_froms} \"^#{CREW_PREFIX}\\|^#{HOME}\" #{flist}"}`.split("\n").uniq.sort
```

commands/remove.rb
```diff
@@ -34,17 +45,42 @@ def self.remove(pkg, verbose)
...
-            next unless line.start_with?(CREW_PREFIX)
-            if system("grep --exclude #{pkg.name}.filelist -Fxq '#{line}' ./meta/*.filelist")
-              puts "#{line} is in another package. It will not be removed during the removal of #{pkg.name}.".orange
-            else
-              puts "Removing file #{line}".yellow if verbose
-              FileUtils.remove_file line, exception: false
-            end
+          # When searching for files to delete we exclude the files from
+          # all packages and dependent packages of CREW_ESSENTIAL_PACKAGES.
+          essential_deps_exclude_froms = essential_deps.map { |i| File.file?("#{File.join(CREW_META_PATH, i.to_s)}.filelist") ? "--exclude-from=#{File.join(CREW_META_PATH, i.to_s)}.filelist" : '' }.join(' ')
```
Zopolis4, line 59: Also, should we maybe move this logic into a separate private function in `commands/remove.rb`?

Zopolis4: i.e. have a function which takes the package name and list of essential dependencies and returns the `unique_to_package_files` and `package_files_that_overlap` variables.

commands/remove.rb
satmandu: It can be in the same line now that it is simplified!

commands/remove.rb
satmandu: Honestly maybe it should actually go in package or package_helpers?

I'm having to use it in the testing file for remove I am trying to get working...

commands/remove.rb
satmandu: Hmmm... what do you think @uberhacker?

Just grepping for `/home/chronos/user` in the manifests dir it does look like there are a decent number of packages that put files in `#{HOME}`.

commands/remove.rb
satmandu: That's also going to get complicated quickly. This logic is tied tightly to excluding files not to delete, in favor of doing it fast, so maybe moving too much to other places might not make sense.

But feel free to refactor...

satmandu:
@supechicken @Zopolis4 Here's a conundrum. This test app I wrote to test the remove function keeps breaking, because the essential packages list that gets generated keeps changing.

Can you run this in a container (so you **DON't BREAK YOUR SYSTEM**) and tell me what I'm doing wrong?

```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=slow_remove crew update \
&& yes | crew upgrade ; ruby ../tests/commands/remove.rb
```

lib/const.rb
satmandu: docopt is broken without the [options] fields.
try this. It uses the `-d` flag to enable `--no-color`.
```
crew -d sysinfo
```
Now try this:
```
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=slow_remove crew update ; crew -d sysinfo
```
Only the latter usage works. docopt has been erroring out with these flags for a while... It's an old regression!

commands/remove.rb
satmandu: Thanks! Changed!

satmandu:
(I'm leaving this in draft until that test issue is figured out... as I want to be really conservative here.)

commands/remove.rb
Zopolis4: I'm trying to minimise the amount of stuff in `lib/package.rb` because otherwise we're loading all that code as part of a package object which will probably lead to performance headaches in the future-- package_helpers would be fine but I don't see where we'd need to use this code in any other place apart from here?

lib/const.rb
Zopolis4: That's because we didn't list `[-d |--no-color]` in crew sysinfo-- adding `[options]` will make docopt automatically accept all the listed options we have i.e. passing `--recursive-build` to `crew sysinfo` will be accepted. 

also-- `crew sysinfo` doesn't print any color, so it shouldn't need to accept the `--no-color` option. 

if there are some commands which need to support specific arguments which they don't currently, then by all means add those arguments to those commands, but adding `[options]` makes every command accept every argument, which is misleading.

commands/remove.rb
satmandu: I'm thinking that at least for generating the list of essential dependencies that should only happen when we do a crew update. Maybe we just add it to device.json then and add it to a variable then? Then we can read it whenever we need to access it?

The other stuff, yeah, better to just leave it in remove...

satmandu:
I'm hoping this isn't wildly inefficient:
```ruby
@last_update_check = Dir["#{CREW_LIB_PATH}/{.git/FETCH_HEAD,lib/const.rb}"].compact.map { |i| File.mtime(i).utc.to_i }.max
```

commands/remove.rb
satmandu: It looks really nice!

commands/remove.rb
uberhacker: Yes, we have `HOME` stuff in the manifest filelists so we need to remove that too during package removal.

commands/remove.rb
uberhacker: The doc block method is easier to read and adjust the text, if necessary.

uberhacker:
Rubocop is still complaining.

bin/crew
```diff
@@ -223,11 +223,12 @@ def generate_compatible
...
     end
   end
-  File.open(File.join(CREW_CONFIG_PATH, 'device.json'), 'w') do |file|
-    output = JSON.parse @device.to_json
-    file.write JSON.pretty_generate(output)
-  end
-  puts 'Generating compatible packages done.'.orange if CREW_VERBOSE
+  puts 'Determining essential dependencies from CREW_ESSENTIAL_PACKAGES...'.orange if CREW_VERBOSE
+  @device[:essential_deps] = []
+  @device[:essential_deps].concat(CREW_ESSENTIAL_PACKAGES.flat_map { |i| Package.load_package("#{i}.rb").get_deps_list }.push(*CREW_ESSENTIAL_PACKAGES).uniq.sort)
```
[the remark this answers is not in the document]
satmandu: It allows all elements of CREW_ESSENTIAL_PACKAGES to be pushed to the array.

lib/const.rb
satmandu: Feel free to submit a PR or add a change to this PR to handle that. Just make sure `crew -d <command>` doesn't break things.

commands/remove.rb
satmandu: I tried moving more of this stuff out, but it made things slower...

lib/const.rb
satmandu: To add further impetus... NOT having `[options]` there for everything BREAKS Docopt... so we need to find a way around that, otherwise stuff like `--no-color` causes crew to error out in docopt before args get passed to `crew` from it.

install.sh
```diff
@@ -113,7 +114,7 @@ LIBC_VERSION=$(/lib"$LIB_SUFFIX"/libc.so.6 2>/dev/null | awk 'match($0, /Gentoo
 if grep -s "fma4" /proc/cpuinfo ; then
   echo_info "Notice: You are running an AMD StoneyRidge device; due to some bugs some packages may fail with a segmentation fault and need to be rebuilt."
   echo_info "If this happens, please report them to: https://github.com/chromebrew/chromebrew/issues"
-  echo_info "If the install fails, try running 'CREW_AMD_INSTALL=1 bash <(curl -Ls git.io/vddgY) && . ~/.bashrc'"
+  echo_info "If the install fails, try running 'CREW_AMD_INSTALL=1 exec bash --init-file <(curl -Ls git.io/vddgY)'"
```
uberhacker, line 117: Wasn't this just changed recently in #10317?

satmandu: Fixed.

uberhacker (approved):
Great work!

PR #10311, opened 2024-08-14
Update gtk4, libdrm, mesa, test buildsystems changes, make unit tests easier to debug

satmandu:
- Fixes GitHub CI Unit Tests and makes them easier to change.
- Updates `gtk4`, `libdrm`, and `mesa`.
- Adds build extras and install extras to all build systems.
- Adds `git` config options to handle GitHub outages such that `crew update` will hopefully return quicker.

##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=unit_test crew update \
&& yes | crew upgrade
```

### DIFF
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -28,10 +28,10 @@ jobs:
             echo "ALL_CHANGED FILES is/are ${ALL_CHANGED_FILES}." && \
             export CHANGED_PACKAGES="$(echo ${ALL_CHANGED_FILES} | sed -e 's,packages/,,g' -e 's,\.rb,,g' | sort)" && \
             echo "CHANGED PACKAGES is/are ${CHANGED_PACKAGES}." && \
-            export X86_64_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do echo $(grep -l \"compatibility.*all\|x86_64\" packages/$p.rb); done | xargs)" && \
-            echo "PR #${PR_NUMBER} has these x86_64 packages: ${X86_64_PACKAGES}" && \
-            export ARMV7L_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do echo $(grep -l \"compatibility.*all\|armv7l\" packages/$p.rb); done | xargs)" && \
-            echo "PR #${PR_NUMBER} has these ARMv7L packages: ${ARMV7L_PACKAGES}" && \
+            export X86_64_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do grep -q "compatibility.*all\|x86_64" packages/$p.rb && echo $p; done | xargs)" && \
+            echo "PR #${PR_NUMBER} has these x86_64 compatible packages: ${X86_64_PACKAGES}" && \
+            export ARMV7L_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do grep -q "compatibility.*all\|armv7l" packages/$p.rb && echo $p; done | xargs)" && \
+            echo "PR #${PR_NUMBER} has these ARMv7L compatible packages: ${ARMV7L_PACKAGES}" && \
             sudo docker run --rm \
             --platform linux/amd64 -i satmandu/crewbuild:nocturne-x86_64.m90 \
             sudo -i -u chronos /bin/bash -c " \
@@ -52,10 +52,10 @@ jobs:
             echo "ALL_CHANGED FILES is/are ${ALL_CHANGED_FILES}." && \
             export CHANGED_PACKAGES="$(echo ${ALL_CHANGED_FILES} | sed -e 's,packages/,,g' -e 's,\.rb,,g' | sort)" && \
             echo "CHANGED PACKAGES is/are ${CHANGED_PACKAGES}." && \
-            export X86_64_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do echo $(grep -l \"compatibility.*all\|x86_64\" packages/$p.rb); done | xargs)" && \
-            echo "PR #${PR_NUMBER} has these x86_64 packages: ${X86_64_PACKAGES}" && \
-            export ARMV7L_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do echo $(grep -l \"compatibility.*all\|armv7l\" packages/$p.rb); done | xargs)" && \
-            echo "PR #${PR_NUMBER} has these ARMv7L packages: ${ARMV7L_PACKAGES}" && \
+            export X86_64_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do grep -q "compatibility.*all\|x86_64" packages/$p.rb && echo $p; done | xargs)" && \
+            echo "PR #${PR_NUMBER} has these x86_64 compatible packages: ${X86_64_PACKAGES}" && \
+            export ARMV7L_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do grep -q "compatibility.*all\|armv7l" packages/$p.rb && echo $p; done | xargs)" && \
+            echo "PR #${PR_NUMBER} has these ARMv7L compatible packages: ${ARMV7L_PACKAGES}" && \
             [[ $ARMV7L_PACKAGES ]] || exit 0 && \
             sudo docker run --rm \
             --platform linux/arm/v7 -t satmandu/crewbuild:fievel-armv7l.m91 \
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -24,11 +24,11 @@ jobs:
         env:
           ALL_CHANGED_FILES: ${{ steps.changed-ruby-files.outputs.all_changed_files }}
         run: |
-            echo "pwd is $(pwd)" && \
-            ls -aFl && \
             echo "ALL_CHANGED FILES is ${ALL_CHANGED_FILES}." && \
             export CHANGED_PACKAGES="$(echo ${ALL_CHANGED_FILES#packages/} | tr -d '.rb' | sort)" && \
             echo "CHANGED PACKAGES is ${CHANGED_PACKAGES}." && \
+            export X86_64_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do echo $(grep -l \"compatibility.*all\|x86_64\" packages/$p.rb); done | xargs)" && \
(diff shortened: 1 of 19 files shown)


.github/workflows/Unit-Test.yml
```diff
@@ -36,18 +36,20 @@ jobs:
...
-            export compatible_packages=$(comm -12 <(crew list -d compatible | sort) <(echo ${changed_packages}) | tr -s \"\n\" ' ') ; \
-            [[ -n ${compatible_packages-} ]] && (for pkg in ${compatible_packages} ;\
-            do \
-            echo \"Testing install of compatible package ${pkg}\"; \
-            yes | time crew install ${pkg}; \
+            [[ -n ${changed_packages-} ]] && ( \
+            for pkg in ${changed_packages}; \
+              do \
+            export all_compatible_packages=$(crew list -d compatible) ;\
+            if $(echo ${all_compatible_packages} | grep ${pkg}); then ;\
```
uberhacker, line 43: Might want to add `| grep ^${pkg}$`.  Example packages that would throw off the results:
```
gtksourceview
gtksourceview_{3|4|5}
google_chrome
google_chrome{_beta|_stable}
```

.github/workflows/Unit-Test.yml
```diff
@@ -20,11 +20,19 @@ jobs:
...
-      - name: Unit Tests (x86_64)
         run: |
-            sudo docker run --rm -t satmandu/crewbuild:m90-x86_64 sudo -i -u chronos /bin/bash -c "
+            echo "ALL_CHANGED_FILES is ${ALL_CHANGED_FILES}" && \
+            export CHANGED_PACKAGES="$(echo ${ALL_CHANGED_FILES#packages/} | tr -d '.rb' | sort)" \
+            echo "CHANGED_PACKAGES is ${CHANGED_PACKAGES}" \
+            sudo docker run --rm \
+            -e ALL_CHANGED_FILES="${ALL_CHANGED_FILES}" \
+            -e CHANGED_PACKAGES="$(CHANGED_PACKAGES}" \
+            -t satmandu/crewbuild:m90-x86_64 \
```
uberhacker, line 33: No need to run this unless changed package compatibility is all or x86_64?

satmandu: I think we still want to know that core functionality is not broken for arm...

uberhacker: No need if the affected file changes are not including that architecture?

uberhacker: Example:
There would be no need to run arm tests if the sbcl package was updated since it is only compatible with x86_64 and i686.

.github/workflows/Unit-Test.yml
```diff
@@ -20,48 +20,38 @@ jobs:
...
         if: steps.changed-ruby-files.outputs.any_changed == 'true'
+      - name: Unit Tests (x86_64)
         env:
           ALL_CHANGED_FILES: ${{ steps.changed-ruby-files.outputs.all_changed_files }}
-      - name: Unit Tests (x86_64)
         run: |
-            sudo docker run --rm -t satmandu/crewbuild:m90-x86_64 sudo -i -u chronos /bin/bash -c "
+            echo "ALL_CHANGED FILES is ${ALL_CHANGED_FILES}." && \
+            export CHANGED_PACKAGES="$(echo ${ALL_CHANGED_FILES#packages/} | tr -d '.rb' | sort)" && \
+            echo "CHANGED PACKAGES is ${CHANGED_PACKAGES}." && \
```
uberhacker, line 29: Add?
```
export X86_PACKAGES="$for p in ${CHANGED_PACKAGES}; do echo $(grep -l compatibility.*x86_64.* $p); done | xargs)"
```
Then if any packages exist with x86_64 architecture you can launch the container.  Otherwise, it would be pointless, correct?  Same logic could be used with arm and i686 containers as well.

uberhacker: Of course, this logic would be exempt if any of the changed files affect crew itself or are outside the packages directory.

install.sh
```diff
@@ -358,6 +358,9 @@ else
 
   # Make the git default branch error messages go away.
   git config --global init.defaultBranch main
+  # Help handle situations where GitHub is down.
+  git config --global http.lowSpeedLimit 1000
+  git config --global http.lowSpeedTime 5
```
uberhacker, line 363: What happens if the internet itself is down?

uberhacker:
Rubocop doesn't like the indentation.

install.sh
satmandu: That timeout should still hit!

.github/workflows/Unit-Test.yml
satmandu: Hmmm, we need to see what the working directory is when a github action starts...

uberhacker: Actually, we need to include `all` in the filtering too.
```
export X86_64_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do echo $(grep -l \"compatibility.*[all|x86_64]\" packages/$p.rb); done | xargs)"
```

satmandu: Got that already... :)

satmandu: Do we need the brackets though around all and x86_64?

uberhacker: Try and find out.  It worked for me with them.

satmandu: Looks like I got it, and early escape from loading the armv7l container if the package doesn't contain any armv7l packages is in there too.

install.sh
```diff
@@ -358,6 +358,9 @@ else
 
   # Make the git default branch error messages go away.
   git config --global init.defaultBranch main
+  # Help handle situations where GitHub is down.
+  git config --global http.lowSpeedLimit 1000
```
Zopolis4, line 362: If possible we should avoid setting global configuration settings-- can we set these in `CREW_PREFIX/lib/crew` instead?

lib/fixup.rb
```diff
@@ -38,6 +38,10 @@ def require_gem(gem_name_and_require = nil, require_override = nil)
   system 'git sparse-checkout reapply'
 end
 
+# Set git timeout values for situations where GitHub is down.
+system 'git config --global http.lowSpeedLimit 1000' if `git config --local http.lowSpeedLimit`.empty?
+system 'git config --global http.lowSpeedTime 5' if `git config --local http.lowSpeedTime`.empty?
```
Zopolis4, line 43: Again, let's set these locally.

tests/unit_test.sh
```diff
@@ -0,0 +1,36 @@
...
+ruby ../tests/commands/const.rb
+ruby ../tests/commands/help.rb
+ruby ../tests/commands/prop.rb
+if [[ -n ${ALL_CHANGED_FILES-} ]]; then
+  # for file in ${ALL_CHANGED_FILES}; do
+    # ruby ../tests/prop_test "$file"
+    # ruby ../tests/buildsystem_test "$file"
+  # done
+  if [[ -n ${CHANGED_PACKAGES-} ]]; then
+    all_compatible_packages=$(crew list -d compatible)
```
Zopolis4, line 18: Do we need to pass `-d here`? doesn't that just disable colors?

install.sh
satmandu: Sure.

tests/unit_test.sh
satmandu: Yes, because otherwise it spits out color ansi codes... 

It's really useful to be able to do local testing where the TTY will be detected and hence color will automatically be enabled, but that will much up the env variables...

.github/workflows/Unit-Test.yml
```diff
@@ -13,55 +13,56 @@ jobs:
...
         env:
           ALL_CHANGED_FILES: ${{ steps.changed-ruby-files.outputs.all_changed_files }}
-      - name: Unit Tests (x86_64)
         run: |
-            sudo docker run --rm -t satmandu/crewbuild:m90-x86_64 sudo -i -u chronos /bin/bash -c "
+            export PR_NUMBER="${{ github.event.pull_request.number || github.event.issue.number }}" && \
+            echo "ALL_CHANGED FILES is/are ${ALL_CHANGED_FILES}." && \
+            export CHANGED_PACKAGES="$(echo ${ALL_CHANGED_FILES} | sed -e 's,packages/,,g' -e 's,\.rb,,g' | sort)" && \
+            echo "CHANGED PACKAGES is/are ${CHANGED_PACKAGES}." && \
+            export X86_64_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do grep -q "compatibility.*all\|x86_64" packages/$p.rb && echo $p; done | xargs)" && \
```
uberhacker, line 31: I think we want `grep -l` instead of `grep -q` because we want to return the package names, correct?

uberhacker: Oh, I see you `echo $p`.  Nevermind!

lib/buildsystems/python.rb
```diff
@@ -12,24 +19,29 @@ def self.build
...
-    if File.file?('setup.py')
-      puts "Python build options being used: #{PY3_SETUP_BUILD_OPTIONS} #{@python_build_options}".orange
-      system "MAKEFLAGS=-j#{CREW_NPROC} python3 setup.py build #{PY3_SETUP_BUILD_OPTIONS} #{@python_build_options}"
-    else
-      puts "Python build options being used: #{PY3_BUILD_OPTIONS}".orange
-      system "MAKEFLAGS=-j#{CREW_NPROC} python3 -m build #{PY3_BUILD_OPTIONS}"
+    Dir.chdir(@python_build_relative_dir) do
+      if File.file?('setup.py')
+        puts "Python build options being used: #{PY3_SETUP_BUILD_OPTIONS} #{@python_build_options}".orange
+        system "MAKEFLAGS=-j#{CREW_NPROC} python3 setup.py build #{PY3_SETUP_BUILD_OPTIONS} #{@python_build_options}"
```
uberhacker, line 25: Add `puts "MAKEFLAGS=... if VERBOSE`?

uberhacker (approved):
LGTM.  Really looking forward to bypassing arm tests when not needed.

PR #10317, opened 2024-08-16
Fix depreciated mentions of `--init-file` and update mentions of running the installer script again

VenomousSteam81:
Remove references of depreciated `--init-file`

## Description

There are no issues as far as I know, but I'm pretty sure running the install command with `--init-file` was depreciated some time ago. This will still work fine on all systems as I have only removed instances of where it tells you to run the install command with `--init-file`.
However, I'm not entirely sure if running these updated commands with the variables will work, and I don't have any ways of testing this.


uberhacker (approved):
Nice!  Thank you for the update!

VenomousSteam81:
You are welcome! Surprised it got pushed, especially that quick lol.

PR #10328, opened 2024-08-20
Cleanup glibc packages, rebuild glibc 2.32, fix broken git repo clone in installs, rebuild OpenSSL to fix pkgconfig breakage in 3.3.1

satmandu:
- Remove old cruft from older glibc packages, and rebuild glibc 2.32 (used in M97) as a build test.
- This fixes the M97 container glibc 2.32 installs.
- Have crew be able to try to fixup its own git repo clone.
- Adjust `fixup.rb` to be able to be run standalone... this helped track down some issues with error messages from dependencies... e.g., git errors from inside `const.rb`, which have also been addressed.
- OpenSSL 3.3.1 has broken pkgconfig files because `--libdir` configure behavior changed and now requires `--libdir=#{ARCH_LIB}` instead of `--libdir=#{CREW_LIB_PREFIX}`. Also there is a cmake file generation fix later than 3.3.1, so I rebuilt OpenSSL from the 3.3 branch. Upstream issue opened at https://github.com/openssl/openssl/issues/25249

Broken 3.3.1 `/usr/local/lib/pkgconfig/openssl.pc`:
```
libdir=/usr/local
includedir=/usr/local/include

Name: OpenSSL
Description: Secure Sockets Layer and cryptography libraries and tools
Version: 3.3.1
Requires: libssl libcrypto
```
Fixed `/usr/local/lib/pkgconfig/openssl.pc`:
```
prefix=/usr/local
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: OpenSSL
Description: Secure Sockets Layer and cryptography libraries and tools
Version: 3.3.2-dev
Requires: libssl libcrypto
```
##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
##
##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=more_glibc crew update \
&& yes | crew upgrade
```


lib/fixup.rb
```diff
@@ -21,8 +28,28 @@ def require_gem(gem_name_and_require = nil, require_override = nil)
 end
 require_gem('highline/import')
 
+# All needed constants & variables should be defined here in case they
+# have not yet been loaded or or fixup is being run standalone.
```
uberhacker, line 32: Two `or`s?

lib/fixup.rb
```diff
@@ -21,8 +28,28 @@ def require_gem(gem_name_and_require = nil, require_override = nil)
...
+CREW_CONST_GIT_COMMIT = '72d807aac' unless defined?(CREW_CONST_GIT_COMMIT)
+CREW_REPO   = 'https://github.com/chromebrew/chromebrew.git' unless defined?(CREW_REPO)
+CREW_BRANCH = 'master' unless defined?(CREW_BRANCH)
+unless defined?(ARCH)
+  KERN_ARCH = Etc.uname[:machine] unless defined? KERN_ARCH
+  ARCH = %w[aarch64 armv8l].include?(KERN_ARCH) ? 'armv7l' : KERN_ARCH
+end
+LIBC_VERSION = Etc.confstr(Etc::CS_GNU_LIBC_VERSION).split.last unless defined?(LIBC_VERSION)
+CREW_PACKAGES_PATH = File.join(CREW_LIB_PATH, 'packages') unless defined?(CREW_PACKAGES_PATH)
+@device = PackageUtils.load_json unless defined? @device
```
uberhacker, line 51: Wouldn't it be easier just to `require 'const'`?

lib/fixup.rb
```diff
@@ -215,8 +233,8 @@ def self.postremove; end
 FileUtils.rm "#{CREW_PACKAGES_PATH}/pagerenv" if File.file?("#{CREW_PACKAGES_PATH}/pagerenv")
 
 # Handle broken system glibc affecting gcc_lib on newer x86_64 ChromeOS milestones.
-if (ARCH == 'x86_64') && (LIBC_VERSION.to_f >= 2.36)
-  abort("patchelf is needed. Please run: 'crew install patchelf && crew update'") unless File.file?(File.join(CREW_PREFIX, 'bin/patchelf'))
+if (ARCH == 'x86_64') && (LIBC_VERSION >= '2.36')
```
uberhacker, line 236: Better to compare numbers than strings?

packages/glibc_build223.rb
```diff
@@ -335,7 +321,7 @@ def self.install
     end
     # Only save libnsl.so.2, since libnsl.so.1 is provided by perl
     # For this to work, build on a M107 or newer container.
-    FileUtils.cp File.realpath("#{CREW_DEST_LIB_PREFIX}/libnsl.so.1"), "#{CREW_DEST_LIB_PREFIX}/libnsl.so.2"
+    FileUtils.ln_sf File.realpath("#{CREW_DEST_LIB_PREFIX}/libnsl.so.1"), "#{CREW_DEST_LIB_PREFIX}/libnsl.so.2"
```
uberhacker, line 324:
```
 FileUtils.ln_sf File.realpath("#{CREW_LIB_PREFIX}/libnsl.so.1"), "#{CREW_DEST_LIB_PREFIX}/libnsl.so.2
```
?

packages/glibc_build223.rb
```diff
@@ -377,26 +364,14 @@ def self.postinstall
...
-          if ['shared object', 'symbolic link'].any? { |type| @filetype.include?(type) }
-            g = File.basename(f)
-            FileUtils.ln_sf f.to_s, "#{CREW_LIB_PREFIX}/#{g}"
-          end
-        end
         # Reject entries which aren't libraries ending in .so, and which aren't files.
         # Reject text files such as libc.so because they points to files like
         # libc_nonshared.a, which are not provided by ChromeOS
-        Dir["/usr/#{ARCH_LIB}/#{lib}.so*"].reject { |f| File.directory?(f) }.each do |f|
+        Dir["{,/usr}/#{ARCH_LIB}/#{lib}.so*"].compact.reject { |f| File.directory?(f) }.each do |f|
```
uberhacker, line 374: `Dir["#{CREW_LIB_PREFIX}/#{lib}.so*"]`?

packages/glibc_build227.rb
```diff
@@ -327,30 +327,20 @@ def self.install
...
-          if ['shared object', 'symbolic link'].any? { |type| @filetype.include?(type) }
-            g = File.basename(f)
-            FileUtils.ln_sf f.to_s, "#{CREW_DEST_LIB_PREFIX}/#{g}"
-          end
-        end
         # Reject entries which aren't libraries ending in .so, and which aren't files.
         # Reject text files such as libc.so because they points to files like
         # libc_nonshared.a, which are not provided by ChromeOS
-        Dir["/usr/#{ARCH_LIB}/#{lib}.so*"].reject { |f| File.directory?(f) }.each do |f|
+        Dir["{,/usr}/#{ARCH_LIB}/#{lib}.so*"].compact.reject { |f| File.directory?(f) }.each do |f|
```
uberhacker, line 343: `Dir["#{CREW_LIB_PREFIX}/#{lib}.so*"]`?

packages/glibc_build227.rb
```diff
@@ -364,7 +354,7 @@ def self.install
     end
     # Only save libnsl.so.2, since libnsl.so.1 is provided by perl
     # For this to work, build on a M107 or newer container.
-    FileUtils.cp File.realpath("#{CREW_DEST_LIB_PREFIX}/libnsl.so.1"), "#{CREW_DEST_LIB_PREFIX}/libnsl.so.2"
+    FileUtils.ln_sf File.realpath("#{CREW_DEST_LIB_PREFIX}/libnsl.so.1"), "#{CREW_DEST_LIB_PREFIX}/libnsl.so.2"
```
uberhacker, line 357:
```
FileUtils.ln_sf File.realpath("#{CREW_LIB_PREFIX}/libnsl.so.1"), "#{CREW_DEST_LIB_PREFIX}/libnsl.so.2"
```
?

packages/glibc_build227.rb
```diff
@@ -396,25 +387,15 @@ def self.postinstall
...
-          if ['shared object', 'symbolic link'].any? { |type| @filetype.include?(type) }
-            g = File.basename(f)
-            FileUtils.ln_sf f.to_s, "#{CREW_LIB_PREFIX}/#{g}"
-          end
-        end
         # Reject entries which aren't libraries ending in .so, and which aren't files.
         # Reject text files such as libc.so because they points to files like
         # libc_nonshared.a, which are not provided by ChromeOS
-        Dir["/usr/#{ARCH_LIB}/#{lib}.so*"].reject { |f| File.directory?(f) }.each do |f|
+        Dir["{,/usr}/#{ARCH_LIB}/#{lib}.so*"].compact.reject { |f| File.directory?(f) }.each do |f|
```
uberhacker, line 398: `Dir["#{CREW_LIB_PREFIX}/#{lib}.so*"]`?

packages/glibc_build232.rb
```diff
@@ -180,43 +181,29 @@ def self.install
...
-            if ['shared object', 'symbolic link'].any? { |type| @filetype.include?(type) }
-              g = File.basename(f)
-              FileUtils.ln_sf f.to_s, "#{CREW_DEST_LIB_PREFIX}/#{g}"
-            end
-          end
           # Reject entries which aren't libraries ending in .so, and which aren't files.
           # Reject text files such as libc.so because they points to files like
           # libc_nonshared.a, which are not provided by ChromeOS
-          Dir["/usr/#{ARCH_LIB}/#{lib}.so*"].reject { |f| File.directory?(f) }.each do |f|
+          Dir["{,/usr}/#{ARCH_LIB}/#{lib}.so*"].compact.reject { |f| File.directory?(f) }.each do |f|
```
uberhacker, line 206: `Dir["#{CREW_LIB_PREFIX}/#{lib}.so*"]`?

lib/fixup.rb
```diff
@@ -1,5 +1,12 @@
+#!/usr/bin/env ruby
```
supechicken, line 1: Wrong file permission?

satmandu: Looks right?
```bash
chronos@cheekon-i686 /output/git/satmandu (more_glibc *$%)$ ls -aFl lib/fixup.rb
-rwxrwxr-x 1 chronos chronos 16284 Aug 20 19:53 lib/fixup.rb*
chronos@cheekon-i686 /output/git/satmandu (more_glibc *$%)$
```

lib/fixup.rb
satmandu: For sure. This is an exercise in throwing out dependencies to see what happens. This was useful in tracking down an issue in `const.rb`, so maybe that's useful?

lib/fixup.rb
satmandu: Apparently it works for both. Is it faster not to have to do a conversion to float?

packages/glibc_build223.rb
satmandu: No. We're looking for what the symlink we may have just created in dest_lib_prefix actually points to.

packages/openssl.rb
```diff
@@ -47,7 +47,7 @@ def self.build
       CFLAGS=\"#{@ARCH_C_LTO_FLAGS}\" CXXFLAGS=\"#{@ARCH_CXX_LTO_FLAGS}\" \
       LDFLAGS=\"#{@ARCH_LDFLAGS}\" \
       mold -run ./Configure --prefix=#{CREW_PREFIX} \
-      --libdir=#{CREW_LIB_PREFIX} \
+      --libdir=#{ARCH_LIB} \
```
uberhacker, line 50: So I'm guessing they use `prefix` and `libdir` together to get the actual library path?

bin/crew
```diff
@@ -249,8 +249,28 @@ end
...
+        `git config --global init.defaultBranch main`
+        # Setup the dir with git information.
+        `git init --ref-format=reftable`
+        `git remote add origin #{CREW_REPO}`
+        # Help handle situations where GitHub is down.
+        `git config --local http.lowSpeedLimit 1000`
+        `git config --local http.lowSpeedTime 5`
+        # Checkout, overwriting local files.
+        system 'git fetch --all', %i[out err] => File::NULL
+        system 'git checkout -f master', %i[out err] => File::NULL
```
uberhacker, line 269: Why use `system` instead of backticks?

packages/openssl.rb
satmandu: In the last six months they broke their previous behavior.... I've opened an issue, which reopens an earlier issue...

bin/crew
satmandu: So I can use `%i[out err] => File::NULL`.

lib/fixup.rb
uberhacker: Does this need to be executable?

lib/fixup.rb
uberhacker: Why don't we have tests for const.rb instead?

lib/fixup.rb
satmandu: I set it so that it could be run independently.  It can also be run with `ruby fixup.rb`.

I'm happy to remove the executable bit. Should I?

lib/fixup.rb
uberhacker: Here is the potential problem with string comparisons of version numbers.
```
irb(main):005> str1 = '2.4'
=> "2.4"
irb(main):006> str2 = '2.36'
=> "2.36"
irb(main):007> str1 > str2
=> true
```
But is it really true for versions?

lib/fixup.rb
satmandu: done.

bin/crew
uberhacker: Would it be possible (and maybe even faster) if all the git commands were executed in a single system call?

lib/fixup.rb
uberhacker: Rubocop is complaining about permissions.  Unless we have a need to run as an executable, maybe it would be better to remove this line?

packages/glibc_build233.rb
```diff
@@ -262,37 +246,20 @@ def self.postinstall
...
-          elsif @opt_verbose
-            puts "#{f} excluded because #{@filetype}"
-          end
+        Dir["{,/usr}/#{ARCH_LIB}/#{lib}.so*"].compact.select { |i| ['shared object', 'symbolic link'].any? { |j| `file #{i}`.chomp.include? j } }.each do |k|
+          FileUtils.ln_sf k, File.join(CREW_LIB_PREFIX, File.basename(k))
         end
       end
     end
-    return unless @libc_version.to_f >= 2.32 && Gem::Version.new(CREW_KERNEL_VERSION.to_s) >= Gem::Version.new('5.10')
+    return unless @libc_version.to_f > 2.32 && Gem::Version.new(CREW_KERNEL_VERSION.to_s) >= Gem::Version.new('5.10')
```
uberhacker, line 262: Why don't we use `Gem::Version.new` with both version comparisons?  In fact, shouldn't we use `Gem::Version.new` throughout?

satmandu: I suspect that that's great for version comparisons with all sorts of subversions, but the internal GLIBC version is just a float, so this is probably quite a bit faster.

lib/fixup.rb
satmandu: This file gets loaded on crew update via git pull, and thus there may be new constants in const.rb that don't exist yet because the const.rb that was loaded is older, so this is really just a superset of the prior approach of checking for the define of a constant before it is first used in `fixup.rb`.

lib/fixup.rb
satmandu: That's fair. I'll change it.

packages/glibc_build223.rb
satmandu: This is for symlinking system libraries.

The first argument to FileUtils.ln_sf is the source file, which is actually a system file in /lib{64} or /usr/lib{64}, and we are looking at symlinks to such files that are already in CREW_DEST_LIB_PREFIX, which with `realpath()` will resolve to the /lib{64} or /usr/lib{64} paths.

The second argument is for where we are making the symlink, which is in CREW_DEST_LIB_PREFIX. We want that there so archiving saves the new symlink.

bin/crew
uberhacker: What would be the readability expense?  Maybe the commands could be grouped together in a doc block?

packages/glibc_build233.rb
satmandu: Done.

packages/glibc_build227.rb
satmandu: Same as https://github.com/chromebrew/chromebrew/pull/10328#discussion_r1725388749

packages/glibc_build227.rb
satmandu: Same as https://github.com/chromebrew/chromebrew/pull/10328#discussion_r1724955226

packages/glibc_build232.rb
satmandu: Same as https://github.com/chromebrew/chromebrew/pull/10328#discussion_r1724955226

lib/fixup.rb
```diff
@@ -21,8 +27,28 @@ def require_gem(gem_name_and_require = nil, require_override = nil)
...
+
 CREW_VERBOSE = ARGV.intersect?(%w[-v --verbose]) unless defined?(CREW_VERBOSE)
 
+HOME = '/home/chronos/user' unless defined?(HOME)
+CREW_PREFIX = '/usr/local' unless defined?(CREW_PREFIX)
+CREW_LIB_PATH = File.join(CREW_PREFIX, 'lib/crew') unless defined?(CREW_LIB_PATH)
+CREW_CONFIG_PATH = File.join(CREW_PREFIX, 'etc/crew') unless defined?(CREW_CONFIG_PATH)
+CREW_META_PATH = File.join(CREW_CONFIG_PATH, 'meta') unless defined?(CREW_CONFIG_PATH)
+# via git log --reverse --oneline lib/const.rb | head -n 1
+CREW_CONST_GIT_COMMIT = '72d807aac' unless defined?(CREW_CONST_GIT_COMMIT)
```
uberhacker, line 41: Why are we hardcoding a commit here if not available?

bin/crew
```diff
@@ -249,8 +249,29 @@ end
...
+        # Checkout, overwriting local files.
+        git fetch --all
+        git checkout -f master
+        git reset --hard origin/#{CREW_BRANCH}
+      GIT_REPAIR_COMMAND
+    end
+
     # Set sparse-checkout folders.
     system "git sparse-checkout set packages manifest/#{ARCH} lib commands bin crew tests tools"
     system 'git sparse-checkout reapply'
```
uberhacker, line 277: Can we doc block these git commands and the following lines too?

lib/fixup.rb
satmandu: That's just the first commit that lists const.rb ...
```
git/chromebrew$ git log --reverse --oneline lib/const.rb | head -n 5
72d807aac Refactoring crew
ef2a9302c Change default option for xz to use multi thread
dd3222416 Add crew -V|--version and -v|--verbose options
acb470333 Add CHROMEOS_RELEASE constant
33540a449 Add recursive build from source option to crew (#2499)
```

satmandu: (It's entirely silly...)

bin/crew

uberhacker (approved):

PR #10335, opened 2024-08-22
Add libc version and chromeos milestone to container bash prompts

satmandu:
## Description
- It's hard to tell when you have multiple chromeos containers open which one you are in, so this will put the Glibc version and the container milestone right in the bash prompt, using the env variables that are now created from `crew_profile_base`.

## Additional information
- This also fixes up the gcc packages to handle GLIBC fallthrough (You can test this with `CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=libc_git_prompt crew update && yes | LIBC_VERSION=2.50 crew upgrade` in a container.
- This also shifts the `git` build to use the cmake buildsystem, while also modifying the cmake buildsystem to work with how cmake is packaged.
- Also adds some missing manifests.

##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
##

##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=libc_git_prompt crew update \
&& yes | crew upgrade
```


lib/buildsystems/cmake.rb
```diff
@@ -1,24 +1,37 @@
...
+      system "#{@pre_cmake_options} #{@mold_linker_prefix_cmd} cmake -B builddir -G Ninja #{@crew_cmake_options} #{@cmake_options}"
+      system "#{CREW_NINJA} -C builddir"
+    end
     @cmake_build_extras&.call
   end
 
   def self.install
-    system "DESTDIR=#{CREW_DEST_DIR} #{CREW_NINJA} -C builddir install"
+    Dir.chdir(@cmake_build_relative_dir) do
+      system "DESTDIR=#{CREW_DEST_DIR} #{CREW_NINJA} -C builddir install"
```
uberhacker, line 26: Doesn't the `-C builddir` already build in the specified directory?  Would this work?
```
system "DESTDIR=#{CREW_DEST_DIR} #{CREW_NINJA} -C #{@cmake_build_relative_dir}/builddir install"
```

uberhacker:
Good catch on the missing filelists.

packages/git.rb
```diff
@@ -67,13 +62,18 @@ def self.build
...
       GIT_PS1_SHOWCOLORHINTS=yes
 
-      PS1='\\[\\033[1;34m\\]\\u@\\H \\[\\033[1;33m\\]\\w \\[\\033[1;31m\\]$(__git_ps1 "(%s)")\\[\\033[0m\\]\\$ '
+      # Add LIBC_VERSION and CHROMEOS_RELEASE_CHROME_MILESTONE set in
+      # crew_profile_base to prompt if in a container.
+      if [[ -e /.dockerenv ]] && [ -n "${LIBC_VERSION+1}" ] && [ -n "${CHROMEOS_RELEASE_CHROME_MILESTONE+1}" ]; then
+        PS1='\\[\\033[1;34m\\]\\u@\\H:$LIBC_VERSION M$CHROMEOS_RELEASE_CHROME_MILESTONE \\[\\033[1;33m\\]\\w \\[\\033[1;31m\\]$(__git_ps1 "(%s)")\\[\\033[0m\\]\\$ '
+      else
+        PS1='\\[\\033[1;34m\\]\\u@\\H \\[\\033[1;33m\\]\\w \\[\\033[1;31m\\]$(__git_ps1 "(%s)")\\[\\033[0m\\]\\$ '
+      fi
```
uberhacker, line 71: :+1:

packages/libssp.rb
```diff
@@ -4,7 +4,17 @@
...
   description 'Libssp is a part of the GCC toolkit.'
   homepage 'https://gcc.gnu.org/'
-  version "14.2.0-glibc#{LIBC_VERSION}" # Do not use @_ver here, it will break the installer.
+  @gcc_libc_version = if %w[2.23 2.27 2.32 2.33 2.35 2.37].any? { |i| LIBC_VERSION.include? i }
+                        LIBC_VERSION
+                      else
+                        case ARCH
+                        when 'i686'
+                          '2.23'
+                        when 'x86_64', 'aarch64', 'armv7l'
```
uberhacker, line 13: This line could also be replaced with `else`.

uberhacker: Or you could just use tertiary logic?
```
ARCH.eql?('i686') ? '2.23' : '2.27'
```

lib/buildsystems/cmake.rb
satmandu: Yes that would work too!  Which is preferable?

uberhacker: I'll let you decide but the `-C` method reduces lines which is always a good thing.

packages/libssp.rb
satmandu: Ternary is definitely the way to go here.

uberhacker: Yes, ternary is a better word than tertiary.

uberhacker (approved):

PR #10336, opened 2024-08-22
Reland of 10152: Download the correct file when there are multiple versions of a package binary

Zopolis4:
Needs testing on real hardware:
- [ ] `x86_64`
- [ ] `i686`
- [x] `armv7l`

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=valian crew update \
&& yes | crew upgrade
```

### DIFF
--- a/lib/package_utils.rb
+++ b/lib/package_utils.rb
@@ -1,6 +1,7 @@
 # lib/package_utils.rb
 # Utility functions that take either a package object or a component of a package object as primary input.
 require 'json'
+require 'net/http'
 require_relative 'const'
 
 class PackageUtils
@@ -23,7 +24,7 @@ class PackageUtils
 
   def self.get_url(pkg, build_from_source: false)
     if !build_from_source && pkg.binary_sha256&.key?(ARCH.to_sym)
-      return "https://gitlab.com/api/v4/projects/26210301/packages/generic/#{pkg.name}/#{pkg.version}_#{ARCH}/#{pkg.name}-#{pkg.version}-chromeos-#{ARCH}.#{pkg.binary_compression}"
+      return get_binary_url(pkg)
     elsif pkg.source_url.is_a?(Hash) && pkg.source_url&.key?(ARCH.to_sym)
       return pkg.source_url[ARCH.to_sym]
     else
@@ -39,6 +40,34 @@ class PackageUtils
     else
       return pkg.source_sha256
     end
+  end
+
+  def self.get_binary_url(pkg)
+    # Fall back to the old method if querying the gitlab API doesn't work for whatever reason.
+    fallback_url = "https://gitlab.com/api/v4/projects/26210301/packages/generic/#{pkg.name}/#{pkg.version}_#{ARCH}/#{pkg.name}-#{pkg.version}-chromeos-#{ARCH}.#{pkg.binary_compression}"
+    # List all the packages with the name and version of the package file.
+    # The name search is fuzzy, so we need to refine it further (otherwise packages like vim, gvim and vim_runtime would break).
+    packages = JSON.parse(Net::HTTP.get(URI("https://gitlab.com/api/v4/projects/26210301/packages?package_name=#{pkg.name}&package_version=#{pkg.version}_#{ARCH}")))
+    # Loop over each result until we get an exact name match, then return the package ID for that match.
+    package_id = 0
+    (0..packages.count - 1).each do |i|
+      next unless packages[i]['name'] == pkg.name
+      package_id = packages[i]['id']
+    end
+    # Return early if we weren't able to find the package ID, so that the CREW_CACHE_ENABLED hack to test packages without uploading them still works.
+    # When we're doing that, we're calling download knowing that there isn't an actual file to download, but relying on CREW_CACHE_ENABLED to save us before we get there.
+    return fallback_url if package_id.zero?
+    # List all the package files for that package ID.
+    package_files = JSON.parse(Net::HTTP.get(URI("https://gitlab.com/api/v4/projects/26210301/packages/#{package_id}/package_files")))
+    # Bail out if we weren't actually able to find a package.
+    return fallback_url if package_files.is_a?(Hash) && package_files['message'] == '404 Not found'
+    # Loop over each result until we find a matching file_sha256 to our binary_sha256.
+    (0..package_files.count - 1).each do |i|
+      next unless package_files[i]['file_sha256'] == pkg.binary_sha256[ARCH.to_sym]
+      return "https://gitlab.com/chromebrew/binaries/-/package_files/#{package_files[i]['id']}/download"
+    end
+    # If we're still here, the likely cause is that the file sha256s are mismatched.
+    return fallback_url
   end
 
   def self.get_clean_version(pkg_version)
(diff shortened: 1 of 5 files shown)


Zopolis4:
Dug up my old C100PA and tested it to the best of my ability, everything looks fine.

lib/package_utils.rb
```diff
@@ -41,6 +48,32 @@ def self.get_sha256(pkg, build_from_source: false)
...
+    package_files = JSON.parse(Net::HTTP.get(URI("https://gitlab.com/api/v4/projects/26210301/packages/#{package_id}/package_files")))
+    # Bail out if we weren't actually able to find a package.
+    return if package_files.is_a?(Hash) && package_files['message'] == '404 Not found'
+    # Loop over each result until we find a matching file_sha256 to our binary_sha256.
+    (0..package_files.count - 1).each do |i|
+      next unless package_files[i]['file_sha256'] == pkg.binary_sha256[ARCH.to_sym]
+      return "https://gitlab.com/chromebrew/binaries/-/package_files/#{package_files[i]['id']}/download"
+    end
+    # If we're still here, the likely cause is that the file sha256s are mismatched. Let the user figure it out.
+    return
```
uberhacker, line 74: Providing a return value helps with debugging.  If you return false then you can make decisions from the calling function.

lib/package_utils.rb
```diff
@@ -23,7 +24,13 @@ def self.incompatible_reason(pkg)
...
   def self.get_url(pkg, build_from_source: false)
     if !build_from_source && pkg.binary_sha256&.key?(ARCH.to_sym)
-      return "https://gitlab.com/api/v4/projects/26210301/packages/generic/#{pkg.name}/#{pkg.version}_#{ARCH}/#{pkg.name}-#{pkg.version}-chromeos-#{ARCH}.#{pkg.binary_compression}"
+      # Fall back to the old method if querying the gitlab API doesn't work for whatever reason.
+      possible_url = get_binary_url(pkg)
+      if possible_url.to_s.empty?
+        return "https://gitlab.com/api/v4/projects/26210301/packages/generic/#{pkg.name}/#{pkg.version}_#{ARCH}/#{pkg.name}-#{pkg.version}-chromeos-#{ARCH}.#{pkg.binary_compression}"
+      else
+        return possible_url
+      end
```
uberhacker, line 33: Why not move this whole section into `get_binary_url`?

lib/package_utils.rb
Zopolis4: I'm using an empty return here rather than return false for brevity (we have 3 possible early return places, so it's easier to specify `if possible_url.to_s.empty?` in the calling function rather than saying `return false` three times in order to only have to write `if possible_url` in the calling function.

lib/package_utils.rb
Zopolis4: I'm handling this here so we only have to state the fallback URL once, rather than three times in `get_binary_url`. 

trying to avoid repeating myself :wink:

lib/package_utils.rb
uberhacker: I don't think you understand what I'm trying to suggest here.  If this function returns nothing and you "let the user figure it out", they would have no idea why this function failed to get a url.  If you returned an error code with a message, on the other hand, that would be much more helpful to figure out what is wrong.

lib/package_utils.rb
uberhacker: Who said you needed to return before the end of all the decisions?

lib/package_utils.rb
Zopolis4: Oh, that's a documentation error on my part. What I mean is that this situation will be handled by the calling function.

lib/package_utils.rb
Zopolis4: I don't really understand your point here-- I've got multiple early exits in `get_binary_url` in order to avoid the function crashing, which is why there are multiple returns. This code in `get_url` is here in order to streamline the early exit process, rather than having to return the fallback url in each instance we want to exit early in `get_binary_url`.

lib/package_utils.rb
```diff
@@ -41,6 +48,32 @@ def self.get_sha256(pkg, build_from_source: false)
     end
   end
 
+  def self.get_binary_url(pkg)
```
uberhacker, line 51:
```
fallback_url = "https://gitlab.com/api/v4/projects/26210301/packages/generic/#{pkg.name}/#{pkg.version}_#{ARCH}/#{pkg.name}-#{pkg.version}-chromeos-#{ARCH}.#{pkg.binary_compression}"
```

lib/package_utils.rb
```diff
@@ -41,6 +48,32 @@ def self.get_sha256(pkg, build_from_source: false)
...
+    packages = JSON.parse(Net::HTTP.get(URI("https://gitlab.com/api/v4/projects/26210301/packages?package_name=#{pkg.name}&package_version=#{pkg.version}_#{ARCH}")))
+    # Loop over each result until we get an exact name match, then return the package ID for that match.
+    package_id = 0
+    (0..packages.count - 1).each do |i|
+      next unless packages[i]['name'] == pkg.name
+      package_id = packages[i]['id']
+    end
+    # Return early if we weren't able to find the package ID, so that the CREW_CACHE_ENABLED hack to test packages without uploading them still works.
+    # When we're doing that, we're calling download knowing that there isn't an actual file to download, but relying on CREW_CACHE_ENABLED to save us before we get there.
+    return if package_id.zero?
```
uberhacker, line 63:
```
return fallback_url if package_id.zero?
```

lib/package_utils.rb
```diff
@@ -41,6 +48,32 @@ def self.get_sha256(pkg, build_from_source: false)
...
+    # List all the package files for that package ID.
+    package_files = JSON.parse(Net::HTTP.get(URI("https://gitlab.com/api/v4/projects/26210301/packages/#{package_id}/package_files")))
+    # Bail out if we weren't actually able to find a package.
+    return if package_files.is_a?(Hash) && package_files['message'] == '404 Not found'
```
uberhacker, line 67:
```
return fallback_url if package_files.is_a?(Hash) && package_files['message'] == '404 Not found'
```

lib/package_utils.rb
uberhacker:
```
return fallback_url
```

lib/package_utils.rb
uberhacker: I added code snippets for moving this section into `get_binary_url`.

lib/package_utils.rb
```diff
@@ -23,7 +24,7 @@ def self.incompatible_reason(pkg)
 
   def self.get_url(pkg, build_from_source: false)
     if !build_from_source && pkg.binary_sha256&.key?(ARCH.to_sym)
-      return "https://gitlab.com/api/v4/projects/26210301/packages/generic/#{pkg.name}/#{pkg.version}_#{ARCH}/#{pkg.name}-#{pkg.version}-chromeos-#{ARCH}.#{pkg.binary_compression}"
+      get_binary_url(pkg)
```
uberhacker, line 27: `return get_binary_url(pkg)`?

Zopolis4: There's no need for a return there, as it will return that value anyways, but I can add one if you'd prefer that.

uberhacker: Probably should for clarity and consistency.  The elsif line below uses a return.

uberhacker (approved):

PR #10340, opened 2024-08-23
Refactor unit tests

Zopolis4:
Was trying to fix the CI issues caused by cloning upstream instead of the pr branch for the hello_world_chromebrew test, and it was complicated enough that I figured now was a good a time as any to give the CI a makeover.

No logic/behavioral changes apart from the `hello_world_chromebrew` fixes, as well as fixing the grep statement for the  compatibility checks (it was checking for `compatibility 'all'` or any line including x86_64 instead of `compatibility 'all'` or `compatibility '*x86_64*'`).

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=conseq crew update \
&& yes | crew upgrade
```


satmandu:
Please add some package changes to this PR for testing!

Zopolis4:
Just did-- I fixed the install for `ttf_hanazono`, which was missing a comma in its `FileUtils.install` call so was actually interpolating strings-- rubocop now autocorrects `Lint/ImplicitStringConcatenation` offenses which broke it further.

satmandu:
And this is why I wanted packages added. :)

.github/workflows/Unit-Test.yml
```diff
@@ -3,84 +3,76 @@ name: Run Unit Tests on PR
...
-            echo $container_cmdline && \
-            sudo docker run --rm \
-            --platform linux/arm/v7 \
-            -i ${armv7l_CONTAINER} \
-            /bin/bash -c "$container_cmdline"
+            echo sudo docker run \
+             --platform $PLATFORM \
+             -v "${{ github.workspace }}/tests/unit_test.sh:/unit_test.sh" \
+             -u chronos \
+             -e CHANGED_PACKAGES="$CHANGED_PACKAGES" \
```
satmandu, line 73: I've discovered that passing environment variables via docker's `-e` mechanism doesn't actually work in github actions... :/

.github/workflows/Unit-Test.yml
```diff
@@ -3,84 +3,76 @@ name: Run Unit Tests on PR
...
+             -e LD_LIBRARY_PATH="/usr/local/lib$LIB_SUFFIX" \
+             -e CREW_REPO="${{ github.event.pull_request.head.repo.clone_url }}" \
+             -e CREW_BRANCH="${{ github.head_ref }}" \
+             "satmandu/crewbuild:$CONTAINER" \
+             /usr/local/bin/bash /unit_test.sh
```
satmandu, line 78: You need the full path to the unit_test.sh...

satmandu (commented):
You're reinventing the wheel here in part... there's some reason to the madness... passing env variables to docker is hacky. See what I did with creating a cmdline variable and passing that to docker...

satmandu:
Try going back to my working Unit-Test.yml, then hacking away one change at a time so you know what breaks...

satmandu:
But I would definitely change the Unit-Test.yml file LAST, as that's really fragile, sadly...

satmandu:
(Also, it would be nice to just be able to use ruby in the Unit-Test.yml... you can do an `apt-get install -y ruby` in there...

.github/workflows/Unit-Test.yml
Zopolis4: no, because we're mounting it with the earlier `-v` argument.

Zopolis4:
We can use `ruby/setup-ruby` for that, although I don't see a place where ruby is particularly helpful here.

Zopolis4:
I'll work on caching the docker pulls at some point in a separate pr, as that'll speed things up as well.

Zopolis4:
as far as i can tell the unit tests are working, but ttf_hanozono is broken :p

Zopolis4:
working! now to test high glibc packages.

satmandu:
You're also missing the early escape logic for skipping the armv7l container if there are no changed packages with arm compatibility....

tests/unit_test.sh
```diff
@@ -1,13 +1,13 @@
 #!/bin/bash -e
 # This is for use as a Github CI Pull Request Unit Test.
-echo "ALL_CHANGED FILES: ${ALL_CHANGED_FILES}."
-echo "CHANGED_PACKAGES: ${CHANGED_PACKAGES}."
+set -e
```
satmandu, line 3: This is already set on the shebang line.

tests/unit_test.sh
```diff
@@ -1,13 +1,13 @@
...
+echo "CHANGED_PACKAGES: ${CHANGED_PACKAGES}"
 cd /usr/local/lib/crew/packages/
+yes | crew update
 yes | crew upgrade
-(cd ~ && \
-git clone --depth=1 https://github.com/chromebrew/chromebrew.git build_test && \
-cd build_test && \
-yes | CREW_CACHE_ENABLED=1 crew build -vf packages/hello_world_chromebrew.rb)
+git clone --depth=1 --branch="$CREW_BRANCH" "$CREW_REPO" ~/build_test
+# crew wont let you build if you're in the installation directory.
```
satmandu, line 9: Thanks for documenting this!

tests/unit_test.sh
Zopolis4: we aren't using the shebang though, as we're running it via directly calling `/usr/local/bin/bash`.

Zopolis4:
> You're also missing the early escape logic for skipping the armv7l container if there are no changed packages with arm compatibility....

doesn't it still make sense to test if any core functionality was broken on armv7l?

.github/workflows/Unit-Test.yml
Zopolis4: that's only an issue when you're running an interactive docker session `-i`, iirc.

Zopolis4:
now handbrake is broken...

Zopolis4:
The culprit: #10246, where the new handbrake binary was never uploaded.
image

Zopolis4:
In any event, the unit test framework looks like its working in all cases, so we're good to go.

tests/unit_test.sh
uberhacker: Would it make sense to clone into `/tmp/build_test` to indicate this is a temporary location?  Also, remove when tests are complete?

Zopolis4: We don't really need to clean up after ourselves given that this is being run in a docker image, which will be destroyed at the end of the CI.

.github/workflows/Unit-Test.yml
```diff
@@ -3,84 +3,76 @@ name: Run Unit Tests on PR
...
+            echo "PR #${{ github.event.pull_request.number }} has these possibly Glibc 2.27 compatible packages: $(for i in ${CHANGED_PACKAGES} ; do if ! grep -q min_glibc packages/$i.rb; then echo $i ; else grep min_glibc packages/$i.rb | tr -d \' | awk '{exit $2 <= 2.27}' || echo $i ; fi ; done | xargs)"
+
+            # If a package doesnt have a min_glibc value, or if its below 2.37, add it to GLIBC_237_COMPATIBLE_PACKAGES
+            echo "GLIBC_237_COMPATIBLE_PACKAGES=$(for i in ${CHANGED_PACKAGES} ; do grep min_glibc packages/$i.rb | tr -d \' | awk '{exit $2 <= 2.37}' || echo $i ; done | xargs)" >> $GITHUB_ENV
+            echo "PR #${{ github.event.pull_request.number }} has these possibly Glibc 2.37 compatible packages: ${GLIBC_237_COMPATIBLE_PACKAGES}"
+
+            # If a package has a compatibility of 'all' or one that includes 'x86_64', add it to the list.
+            echo "PR #${{ github.event.pull_request.number }} has these x86_64 compatible packages: $(for p in ${CHANGED_PACKAGES}; do grep -q "compatibility 'all'\|compatibility '.*x86_64.*'" packages/${p}.rb && echo ${p}; done | xargs)"
```
satmandu, line 37: What variable is this going into?

.github/workflows/Unit-Test.yml
```diff
@@ -3,84 +3,76 @@ name: Run Unit Tests on PR
...
+
+            # If a package has a compatibility of 'all' or one that includes 'armv7l', add it to the list.
+            echo "PR #${{ github.event.pull_request.number }} has these armv7l compatible packages: $(for p in ${CHANGED_PACKAGES}; do grep -q "compatibility 'all'\|compatibility '.*armv7l.*'" packages/${p}.rb && echo ${p}; done | xargs)"
```
satmandu, line 40: What variabele is this going into such that we can replicate the early exit in 
```bash
export ARMV7L_PACKAGES="$(for p in ${CHANGED_PACKAGES}; do grep -q "compatibility.*all\|armv7l" packages/${p}.rb && echo ${p}; done | xargs)" && \
            echo "PR #${PR_NUMBER} has these ARMv7L compatible packages: ${ARMV7L_PACKAGES}" && \
            [[ $ARMV7L_PACKAGES ]] || exit 0
```
?

satmandu (changes requested):
Almost there!

.github/workflows/Unit-Test.yml
Zopolis4: again, doesn't it still make sense to test if any core functionality has broken on armv7l? `unit_test.sh` itself will ensure that no packages are tested if there arent any compatible ones, so we don't have to worry about that.

.github/workflows/Unit-Test.yml
Zopolis4: its not going into a variable, because we didn't actually use the variable before aside from printing it so there's no point to creating one.

.github/workflows/Unit-Test.yml
satmandu: That's fair. And I guess they are happening in parallel now.

satmandu:
The handbrake binaries weren't uploaded because handbrake was never built... in any case I have a build going now...

satmandu (approved):

PR #10343, opened 2024-08-23
Properly separate ConvenienceFunctions out of PackageUtils and move additional functions into MiscFunctions, and document the various roles of each class

Zopolis4:
## Description

`save_json` and `load_json` should not be in PackageUtils, so I moved them out and documented the role of PackageUtils to hopefully prevent something like this happening again. Also, while I was at it, unified some of our disparate function implementations into MiscFunctions and brought them under scope.

ideally at some point i'd take the highline stuff out of `lib/package.rb` and also properly deal with selector.rb and then `Include` them into the package class, but thats for a later date.

## Tested & Working properly:
- [ ] `x86_64`
- [ ] `i686`
- [ ] `armv7l`

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=stree crew update && yes | crew upgrade
```


lib/package_utils.rb
```diff
@@ -1,27 +1,12 @@
...
-    end
-
-    # Copy over original if the write to the tmp file succeeds.
-    FileUtils.cp("#{CREW_CONFIG_PATH}/device.json.tmp", File.join(CREW_CONFIG_PATH, 'device.json')) && FileUtils.rm("#{CREW_CONFIG_PATH}/device.json.tmp")
-  end
-
   def self.installed?(pkg_name)
-    return load_json[:installed_packages].any? { |elem| elem[:name] == pkg_name }
+    device_json = JSON.load_file(File.join(CREW_CONFIG_PATH, 'device.json'))
+    return device_json['installed_packages'].any? { |elem| elem['name'] == pkg_name }
```
satmandu, line 9: Just use this?
```ruby
return ConvenienceFunctions.load_symbolized_json[:installed_packages].any? { |elem| elem['name'] == pkg_name }
```

satmandu: Just so our usage is synchronized. As a ruby newbie I was throughly confused by flitting between the symbolized and non-zymbolized usage.

Zopolis4: that's why i'm trying to swap entirely to strings :sweat_smile:

satmandu:
Thanks for moving stuff logically into separate files. I was hoping you would do this. :)

lib/package_utils.rb
satmandu: Is there an advantage to using the string versions over the symbolized versions?

I'd be happy to just entirely use one or the other. Let's just not use a mix.

Thoughts @uberhacker @supechicken ?

Zopolis4: i'm moving to string versions just because it means we don't have to manually modify the returned array to use symbols-- there's no real reason otherwise, but theres no particular advantage to symbolized versions

uberhacker: Agree that we need to be consistent.  I'm fine with all strings.  It's a bit less confusing.

lib/misc_functions.rb
```diff
@@ -0,0 +1,75 @@
+# lib/misc_functions.rb
+# Generic implementations of various functions/algorithms that are not crew-specific.
+require 'matrix'
+
+class MiscFunctions
```
uberhacker, line 5: How about `ConversionFunctions`?  Then the file could be conversion_functions.rb.  Misc is a bit too generic and the name should represent the general purpose of the functions.

lib/convenience_functions.rb
```diff
@@ -0,0 +1,24 @@
+# lib/convenience_functions.rb
+# Extracted bits of crew-specific code that we use frequently enough that it makes sense to split them out to here.
+require 'json'
+require_relative 'const'
+require_relative 'crewlog'
+
+class ConvenienceFunctions
```
uberhacker, line 7: `JSONFunctions`?

Zopolis4: I forsee this expanding beyond just json-related functions in the future

lib/misc_functions.rb
Zopolis4: i agree misc is generic, but these are just a collection of generic functions, thats kinda the point-- only a few of them convert things, so i don't think thats an appropriate name overall.

lib/convenience_functions.rb
uberhacker: Well, I don't see the point in creating ConvenienceFunctions and MiscFunctions if both functions will be used generically.  If we are going to organize into separate functions then we need to really refactor with a sensible plan.

Zopolis4: the purposes of both files are explained at the top-- ConvenienceFunctions is for repeated bits of crew-related code, while MiscFunctions is for implementations of algorithms and conversions and anything else that isnt crew-specific.

lib/package_utils.rb
satmandu: Can we switch everything over to string in this PR, and just modify fixup to redefine @device with a call to the load_json code so that it is no longer symbolized?

Zopolis4: Can we save that for a second PR? I'd be happy to do it, it just feels out of scope for this one.

satmandu (approved):
LFG...

PR #10344, opened 2024-08-23
Handbrake => 1.8.2, update unit test workflow.

satmandu:
Fixes #10340 (Specifically missing handbrake binaries as mentioned.)

- Mea culpa... I hadn't built the binaries when I updated the handbrake package file, because I was focused upon fixing builds on the Glibc 2.37 systems, and then forgot.

- This also adjusts the unit test workflow to only run the arm container if there are armv7l compatible packages changes or changes to non-package files.

##
Builds:
- [x] `x86_64`
- [ ] `i686`
- [ ] `armv7l`
##
Tested & Working properly:
- [ ] `x86_64`
- [ ] `i686`
- [ ] `armv7l` 
##
##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=handbrake crew update \
&& yes | crew upgrade
```

### DIFF
--- a/.github/workflows/Unit-Test.yml
+++ b/.github/workflows/Unit-Test.yml
@@ -46,21 +46,21 @@ jobs:
             fi
 
             # If a package has a compatibility of 'all' or one that includes 'x86_64', add it to x86_64_PACKAGES.
-            export x86_64_PACKAGES="$(for i in ${CHANGED_PACKAGES}; do grep -q \"compatibility.*all\|x86_64\" packages/${i}.rb && echo ${i}; done | xargs)"
+            export x86_64_PACKAGES="$(for i in ${CHANGED_PACKAGES}; do grep -q "compatibility.*all\|x86_64" packages/${i}.rb && echo ${i}; done | xargs)"
             if [[ -n ${x86_64_PACKAGES} ]]; then
               echo "x86_64_PACKAGES=${x86_64_PACKAGES}" >> $GITHUB_ENV
               echo "PR #${{ github.event.pull_request.number }} has these x86_64 compatible packages: ${x86_64_PACKAGES}"
             fi
 
             ## If a package has a compatibility of 'all' or one that includes 'armv7l', add it to ARMV7L_PACKAGES.
-            export ARMV7L_PACKAGES="$(for i in ${CHANGED_PACKAGES}; do grep -q \"compatibility.*all\|armv7l\" packages/${i}.rb && echo ${i}; done | xargs)"
+            export ARMV7L_PACKAGES="$(for i in ${CHANGED_PACKAGES}; do grep -q "compatibility.*all\|armv7l" packages/${i}.rb && echo ${i}; done | xargs)"
             if [[ -n ${ARMV7L_PACKAGES} ]]; then
               echo "ARMV7L_PACKAGES=${ARMV7L_PACKAGES}" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


satmandu:
@Zopolis4 I think I've changed my mind on this again. The Unit Tests should not load the armv7l container at all if no changed packages have compatibility with armv7l...

satmandu:
Converting to draft because I'm going to use this PR to test some unit test changes..., because we do want to test non-package file updates on the armv7l containers...

.github/workflows/Unit-Test.yml
```diff
@@ -66,6 +74,11 @@ jobs:
             esac
       - name: Run unit tests
         run: |
+             if [[ -z ${NON_PKG_CHANGED_FILES} ]] && [[ $PLATFORM == 'linux/arm/v7' ]] && [ -n "${ARMV7L_PACKAGES+1}" ]; then
```
uberhacker, line 77: Won't this condition always be true?  `[ -n "${ARMV7L_PACKAGES+1}" ]`

satmandu: From https://linuxsimply.com/bash-scripting-tutorial/conditional-statements/if/set-variable-if-not/

```
Similarly, you can employ the parameter expansion with the -n option to check if a variable is set. In this case, use the syntax: if [[ -n "${variable+value}" ]] where “-n” returns true only when the variable is set.
```

satmandu: Changing the logic...

.github/workflows/Unit-Test.yml
```diff
@@ -14,30 +14,57 @@ jobs:
...
+            export ARMV7L_PACKAGES="$(for i in ${CHANGED_PACKAGES}; do grep -q "compatibility.*all\|armv7l" packages/${i}.rb && echo ${i}; done | xargs)"
+            if [[ -n ${ARMV7L_PACKAGES} ]]; then
+              echo "ARMV7L_PACKAGES=${ARMV7L_PACKAGES}" >> $GITHUB_ENV
+              echo "PR #${{ github.event.pull_request.number }} has these armv7l compatible packages: ${ARMV7L_PACKAGES}"
+            fi
 
-            # If a package has a compatibility of 'all' or one that includes 'armv7l', add it to the list.
-            echo "PR #${{ github.event.pull_request.number }} has these armv7l compatible packages: $(for p in ${CHANGED_PACKAGES}; do grep -q "compatibility 'all'\|compatibility '.*armv7l.*'" packages/${p}.rb && echo ${p}; done | xargs)"
+            ## If a package has a compatibility of 'all' or one that includes 'i686', add it to i686_PACKAGES.
+            export i686_PACKAGES="$(for i in ${CHANGED_PACKAGES}; do grep -q "compatibility.*all\|i686" packages/${i}.rb && echo ${i}; done | xargs)"
```
Zopolis4, line 63: this grep logic is broken, as i mentioned in the previous pr:
image

Zopolis4 (changes requested):

.github/workflows/Unit-Test.yml
satmandu: Yeah agreed... Still futzing with it...

satmandu: (That's why it is in draft...)

satmandu: Good catch!

uberhacker (approved):
Nice work!

PR #10355, opened 2024-08-25
Remove usage of .blank? and dependency on activesupport

Zopolis4:
## Description
yeah so we don't actually need to defensively program against our own code-- `gnome_packages` is initialized at line 2 of `lib/gnome.rb`, so it will never be nil, and we can just convert `@pkg_name` to a string so we don't have the `NoMethodError` problems.

also why do we have code for editing package files in `lib/downloader.rb`???

Tested & Working properly:
- [x] `x86_64`
- [ ] `i686`
- [ ] `armv7l`

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=arid crew update \
&& yes | crew upgrade
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -24,7 +24,6 @@ end
 require_gem('highline')
 # The json gem can break when upgrading from a much older version of ruby.
 require_gem('json')
-require_gem('activesupport', 'active_support/core_ext/object/blank')
 require 'digest/sha2'
 require 'fileutils'
 require 'tmpdir'
@@ -137,7 +136,7 @@ class ExitMessage
 end
 
 at_exit do
-  GnomePostinstall.run unless GnomePostinstall.gnome_packages.blank?
+  GnomePostinstall.run unless GnomePostinstall.gnome_packages.empty?
 
   # Print exit messages.
   ExitMessage.print
--- a/install.sh
+++ b/install.sh
@@ -334,7 +334,7 @@ ln -sfv "../lib/crew/bin/crew" "${CREW_PREFIX}/bin/"
 echo "export CREW_PREFIX=${CREW_PREFIX}" >> "${CREW_PREFIX}/etc/env.d/profile"
 
 echo_info "Installing essential ruby gems...\n"
-BOOTSTRAP_GEMS='activesupport concurrent-ruby highline'
+BOOTSTRAP_GEMS='concurrent-ruby highline'
 # shellcheck disable=SC2086
 install_ruby_gem ${BOOTSTRAP_GEMS}
 
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -3,7 +3,7 @@
 require 'etc'
 
 OLD_CREW_VERSION ||= defined?(CREW_VERSION) ? CREW_VERSION : '1.0'
-CREW_VERSION ||= '1.52.0' unless defined?(CREW_VERSION) && CREW_VERSION == OLD_CREW_VERSION
+CREW_VERSION ||= '1.52.1' unless defined?(CREW_VERSION) && CREW_VERSION == OLD_CREW_VERSION
 
 # Kernel architecture.
 KERN_ARCH ||= Etc.uname[:machine]
--- a/lib/downloader.rb
+++ b/lib/downloader.rb
@@ -1,4 +1,3 @@
-require 'active_support/core_ext/object/blank'
 require 'digest/sha2'
 require 'io/console'
 require 'uri'
@@ -61,7 +60,7 @@ def downloader(url, sha256sum, filename = File.basename(url), verbose = false)
 
   unless (sha256sum =~ /^SKIP$/i) || (calc_sha256sum == sha256sum)
     if CREW_FORCE
-      pkg_name = @pkg_name.blank? ? name : @pkg_name
+      pkg_name = @pkg_name.to_s.empty? ? name : @pkg_name
       puts "Updating checksum for #{filename}".lightblue
       puts "from #{sha256sum} to #{calc_sha256sum}".lightblue
       puts "in #{CREW_LOCAL_REPO_ROOT}/packages/#{pkg_name}.rb .".lightblue
--- a/packages/core.rb
+++ b/packages/core.rb
@@ -78,12 +78,12 @@ class Core < Package
   depends_on 'readline'
   depends_on 'rtmpdump'
   depends_on 'ruby'
-  depends_on 'ruby_activesupport'
+  # For use in simultaneous upx compression.
   depends_on 'ruby_concurrent_ruby'
   # For use in ruby prompts.
   depends_on 'ruby_highline'
   # This contains the debugger config files.
-  depends_on 'ruby_pry'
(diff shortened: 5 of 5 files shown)


satmandu (changes requested):
Disagree on principle with avoiding active support's inclusion of .`blank?`. Blank is really useful and I'm sick of programming against failures of the core ruby developers to actually avoid situations where blank will work around ruby crapping out.

https://stackoverflow.com/a/20663389

Zopolis4:
There are two usecases of it in our codebase-- one of which is entirely unnecessary as it is on a method that will never return nil, and another can be trivially replaced. Is it really that useful?

satmandu:
I have personally had it bite me multiple times on a bunch of stuff. Enough that I would have been using blank a lot more than I have been using it if we had it earlier.

Feel free to add the entire code block for it into one of the extra functions blocks if it makes you feel better, but I prefer to have another organization maintaining that code...

satmandu:
A function that routinely throws nomethod errors like `.empty?` does for stuff where it really shouldn't is IMHO broken.

satmandu:
<insert Joker meme...>

Zopolis4:
fwiw, this is the [entire definition](https://justincypret.com/blog/activesupport-blank-vs-empty#blank-is-basically-an-alias-for-empty) of `.blank?`:
```ruby
# lib/active_support/core_ext/object/blank.rb
class Object
  def blank?
    respond_to?(:empty?) ? !!empty? : !self
  end
end
```

Zopolis4:
but again-- there are two instances of it, one of which is unnecessary and one of which was trivial to replace.

satmandu:
> fwiw, this is the [entire definition](https://justincypret.com/blog/activesupport-blank-vs-empty#blank-is-basically-an-alias-for-empty) of `.blank?`:
> ```ruby
> # lib/active_support/core_ext/object/blank.rb
> class Object
>   def blank?
>     respond_to?(:empty?) ? !!empty? : !self
>   end
> end
> ```

Please add it to the functions block, because I don't ever want to get bitten by using `empty?` ever again.

satmandu:
> also why do we have code for editing package files in `lib/downloader.rb`???
> 
That's to automatically update hashes when building packages with version number updates. It's behind a force flag...

Zopolis4:
> > fwiw, this is the [entire definition](https://justincypret.com/blog/activesupport-blank-vs-empty#blank-is-basically-an-alias-for-empty) of `.blank?`:
> > ```ruby
> > # lib/active_support/core_ext/object/blank.rb
> > class Object
> >   def blank?
> >     respond_to?(:empty?) ? !!empty? : !self
> >   end
> > end
> > ```
> 
> Please add it to the functions block, because I don't ever want to get bitten by using `empty?` ever again.

i'm happy to add it if or when we need it, but right now we don't need it.

Zopolis4:
> > also why do we have code for editing package files in `lib/downloader.rb`???
> 
> That's to automatically update hashes when building packages with version number updates. It's behind a force flag...

ah. that probably makes more sense to have somewhere else, i'll take a look at it later

satmandu:
> > > fwiw, this is the [entire definition](https://justincypret.com/blog/activesupport-blank-vs-empty#blank-is-basically-an-alias-for-empty) of `.blank?`:
> > > ```ruby
> > > # lib/active_support/core_ext/object/blank.rb
> > > class Object
> > >   def blank?
> > >     respond_to?(:empty?) ? !!empty? : !self
> > >   end
> > > end
> > > ```
> > 
> > Please add it to the functions block, because I don't ever want to get bitten by using `empty?` ever again.
> 
> i'm happy to add it if or when we need it, but right now we don't need it.

Please do not remove the blank require without adding a function block for it.

I want it to be available.

Zopolis4:
i do not see the logic in adding a method we do not use, but whatever

satmandu:
You're removing the usage of the `.blank?` method in favor of streamlining. My logic is that I am insisting on keeping the method around because I have every intention of continuing to use it, and furthermore I don't want people to avoid using it, because I think it is a safer method to use than `.empty?`

uberhacker:
Would it make sense to propose an upstream improvement to `empty` so it behaves like `blank`?  It's confusing to have multiple functions when the proper expected behavior of a single function would keep it simple.

Zopolis4:
my logic is that there's no need to use a helper method if we don't need to use it, and that it makes sense to limit checks to what is relevant. 

if something will never be nil, it is misleading and unnecessary to check it with `.blank?` instead of `.empty?` or whatever is appropriate in this scenario. 

if a variable may be nil, there still tends to be a way to handle it using ruby's built in features. 

if there is a situation where `.blank?` is unequivocally required, then i see no problem in using it, but if it is not then there's no need.

Zopolis4:
> Would it make sense to propose an upstream improvement to `empty` so it behaves like `blank`? It's confusing to have multiple functions when the proper expected behavior of a single function would keep it simple.

i'll have a look if anyone has asked, but I don't love our odds of changing the ruby upstream language if rails wasn't able to. still, always worth a shot!

satmandu:
> my logic is that there's no need to use a helper method if we don't need to use it, and that it makes sense to limit checks to what is relevant. 
> 
> if something will never be nil, it is misleading and unnecessary to check it with `.blank?` instead of `.empty?` or whatever is appropriate in this scenario. 
> 
> if a variable may be nil, there still tends to be a way to handle it using ruby's built in features. 
> 
> if there is a situation where `.blank?` is unequivocally required, then i see no problem in using it, but if it is not then there's no need.


Yes, but sometimes you end up with an unexpected `nil`.

Blank handles that. Handling unexpected cases is safer.

satmandu:
> > Would it make sense to propose an upstream improvement to `empty` so it behaves like `blank`? It's confusing to have multiple functions when the proper expected behavior of a single function would keep it simple.
> 
> i'll have a look if anyone has asked, but I don't love our odds of changing the ruby upstream language if rails wasn't able to. still, always worth a shot!

Agreed on the odds, which is why i didn't try to upstream it.

Zopolis4:
agree to disagree on this one-- i prefer to avoid defensive programming whenever possible, and i'd rather know exactly where my codebase is interacting with external data and the source of unknowns, whereas it looks like you'd prefer to optimise for eliminating the possibility of these kinds of bugs, which is fair enough.

Zopolis4:
also in terms of more practical concerns-- i can't add blank as a helper function without the stuff from #10343

satmandu:
> agree to disagree on this one-- i prefer to avoid defensive programming whenever possible, and i'd rather know exactly where my codebase is interacting with external data and the source of unknowns, whereas it looks like you'd prefer to optimise for eliminating the possibility of these kinds of bugs, which is fair enough.

This is my medical and public health and engineering training kicking in... Systemic fixes help prevent downstream issues.

Zopolis4:
With the disagreements over the value of `.blank?`, and the significant increase in usage of external gems in the codebase anyways, this isn't a battle worth fighting.

Issue #10352, opened 2024-08-24
🐞 Bug Unable to `crew remove`


uberhacker:
```
$ crew remove qt5_base
/usr/local/lib/crew/commands/remove.rb:64:in `split': invalid byte sequence in US-ASCII (ArgumentError)

          all_other_files = `grep -h --exclude #{[pkg.name](http://pkg.name/)}.filelist \"^#{CREW_PREFIX}\\|^#{HOME}\" #{CREW_META_PATH}/*.filelist`.split("\n").uniq.sort
                                                                                                                                       ^^^^
        from /usr/local/lib/crew/commands/remove.rb:64:in `block in remove'
        from /usr/local/lib/crew/commands/remove.rb:49:in `chdir'
        from /usr/local/lib/crew/commands/remove.rb:49:in `remove'
        from /usr/local/bin/crew:1765:in `block in remove_command'
        from /usr/local/bin/crew:1763:in `each'
        from /usr/local/bin/crew:1763:in `remove_command'
        from /usr/local/bin/crew:1846:in `<main>'
```

satmandu:
Why is `#{[pkg.name](http://pkg.name/)}.filelist` happening?

satmandu:
This is the relevant line 64:
```ruby
all_other_files = `grep -h --exclude #{pkg.name}.filelist \"^#{CREW_PREFIX}\\|^#{HOME}\" #{CREW_META_PATH}/*.filelist`.split("\n").uniq.sort
```

satmandu:
I'm having trouble replicating this. In the hatch M126 container, I add and then remove qt5_base.

(I added a debug line to print what `pkg.name` is right before the relevant complaint.)

![Screenshot_20240825-101011.png](https://github.com/user-attachments/assets/3061fc93-faa9-49f3-a1c7-5982d18c942c)

uberhacker:
Some more context.  This happens because I'm unable to `crew upgrade` due to qt5_base which depends on mysql.  Because mysql is not compatible with my glibc version, the whole process aborts.  And because I'm unable to remove qt5_base, I cannot continue with the updates.  Happens on more than one chromebook that requires updates.

Zopolis4:
Working on mysql updates over at https://github.com/Zopolis4/chromebrew/tree/1really, but as CMake builds are currently broken not making much progress.

[...]

Issue #10350, opened 2024-08-24
🐞 Bug Unable to `crew update`


uberhacker:
```
$ crew update
remote: Enumerating objects: 934, done.
remote: Counting objects: 100% (729/729), done.
remote: Compressing objects: 100% (261/261), done.
remote: Total 934 (delta 544), reused 565 (delta 462), pack-reused 205 (from 1)
Receiving objects: 100% (934/934), 1.38 MiB | 116.00 KiB/s, done.
Resolving deltas: 100% (619/619), completed with 218 local objects.
From https://github.com/chromebrew/chromebrew
 * branch                master     -> FETCH_HEAD
Updating files: 100% (4474/4474), done.
HEAD is now at 36aaefd78 Handle use of version variable in gcc_lib when in install.sh. (#10348)
Package lists, crew, and library updated.
Restarting crew update since there is an updated crew version.
From https://github.com/chromebrew/chromebrew
 * branch                master     -> FETCH_HEAD
HEAD is now at 36aaefd78 Handle use of version variable in gcc_lib when in install.sh. (#10348)
Package lists, crew, and library updated.
<internal:/usr/local/lib64/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:136:in `require': cannot load such file -- highline/import (LoadError)
        from <internal:/usr/local/lib64/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:136:in `require'
        from /home/chronos/user/chromebrew/packages/ruby_pry.rb:2:in `load_package'
        from /usr/local/lib/crew/lib/package.rb:104:in `class_eval'
        from /usr/local/lib/crew/lib/package.rb:104:in `load_package'
        from /usr/local/bin/crew:175:in `set_package'
        from /usr/local/bin/crew:189:in `block in generate_compatible'
        from /usr/local/bin/crew:186:in `each'
        from /usr/local/bin/crew:186:in `generate_compatible'
        from /usr/local/bin/crew:294:in `update'
        from /usr/local/bin/crew:1807:in `update_command'
        from /usr/local/bin/crew:1846:in `<main>'
```

Zopolis4:
also having this issue on some older containers of mine that i havent updated in a few weekrs-- `gem install highline` solves the issue, but i thought the point of all the code at the start of `bin/crew` and in `lib/fixup.rb` was to prevent the need for this...

uberhacker:
[...]

PR #10363, opened 2024-08-27
Fix for require issues with highline and glibc 2.37 refactor

### LINKED ISSUE #10359: Unable to build CMake packages
**Describe the bug**
The cmake buildsystem was broken by #10335.

**To Reproduce**

`satmandu/crewbuild:m90-x86_64` container

      ```shell
      crew reinstall -s libxml2
      ```

      Output:

      ```
      libxml2: Libxml2 is the XML C parser and toolkit developed for the Gnome project.
      Downloading source...
      Cannot find cached git repository. 😔 Will download.
      Initialized empty Git repository in /usr/local/tmp/crew/libxml2.20240826002845.dir/.git/
      remote: Enumerating objects: 3452, done.
      remote: Counting objects: 100% (3452/3452), done.
      remote: Compressing objects: 100% (2167/2167), done.
      remote: Total 3452 (delta 865), reused 2850 (delta 707), pack-reused 0 (from 0)
      Receiving objects: 100% (3452/3452), 3.39 MiB | 2.74 MiB/s, done.
      Resolving deltas: 100% (865/865), done.
      From https://gitlab.gnome.org/GNOME/libxml2
       * tag               v2.13.3    -> FETCH_HEAD
      HEAD is now at 3b1742b Release v2.13.3
      Repository downloaded.
      Caching downloaded git repo...
      Git repo cached.
      Additional cmake_options being used: <no pre_cmake_options> -DLIBXML2_WITH_PYTHON=OFF \
          -DLIBXML2_WITH_LZMA=OFF \
          -DLIBXML2_WITH_ICU=ON
      CMake Error: -C must be followed by a file name.
      CMake Error: Run 'cmake --help' for all supported options.
      Command "bash -c  mold -run cmake -C  -B builddir -G Ninja -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_LIBRARY_PATH=/usr/local/lib64 -DCMAKE_C_FLAGS='-O2 -pipe -ffat-lto-objects -fPIC   -flto=auto' -DCMAKE_CXX_FLAGS='-O2 -pipe -ffat-lto-objects -fPIC   -flto=auto' -DCMAKE_EXE_LINKER_FLAGS='-flto=auto ' -DCMAKE_LINKER_TYPE=MOLD -DCMAKE_SHARED_LINKER_FLAGS='-flto=auto ' -DCMAKE_MODULE_LINKER_FLAGS='-flto=auto ' -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=TRUE -DCMAKE_BUILD_TYPE=Release -DLIBXML2_WITH_PYTHON=OFF \
          -DLIBXML2_WITH_LZMA=OFF \
          -DLIBXML2_WITH_ICU=ON" failed, retrying with "-j1"...
      CMake Error: -C must be followed by a file name.
      CMake Error: Run 'cmake --help' for all supported options.
      There was a build error.
      ```
**Expected behavior**
The build to work successfully

***
/usr/local/bin/crew:1843:<main>:The last update was 3 min, 35 seconds ago.

- User space: `64-bit`
- Chromebrew Kernel version: `5.10`
- Chromebrew Running in Container: `true`

- Chromebrew version: `1.51.1`

- Last update in local repository: 8bf154910 `Fix repology workflow again by installing highline (#10354) (20 hours ago)`

- OS variant: `Chrome OS`
- OS version: `nocturne-release/R90-13816.82.0`

</details>

***

satmandu:
Fixes #10359 #10350 #10352 

## Description
- Refactors `fixup.rb` and `remove.rb` to make them work together more robustly when updating from a much older crew. (Also this uses the non-symbolized json, and then updates the crew json in memory whether symbolized or not.)
- Fixup also uses an `@installed_packages` `set` that can be searched much faster with `@installed_packages.include?(pkg_name)`. We should consider making this global as this should be much faster for checking if stuff is installed, which we do a lot of, since we don't have to interate through all the packages each time.
- `const.rb` was refactored to allow it to be loaded again without more annoying warnings.
- Adjusts commands/list .rb to use the non-symbolized json.
- Adjusts `commands/remove.rb` to not require a verbose argument since `const.rb` is now being pulled in.
- Refactors the glibc 2.37 packages.
- Fixes issues with `highline` includes causing problems with updates from older crew releases.

## And one more thing...
- Also, we should really have code to check to see if the device.json has been updated, and if so to reload the json accordingly, especially since we are now invoking outside programs that might save the json. (This took me a while to debug.)

Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
##

```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=requirefix crew update \
&& yes | crew upgrade && \
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=requirefix crew update \
&& yes | crew upgrade 
```


lib/buildsystems/cmake.rb
```diff
@@ -5,16 +5,16 @@ class CMake < Package
 
   attr_accessor :cmake_build_relative_dir
```
Zopolis4, line 6: Why is this not a property?

lib/fixup.rb
```diff
@@ -3,30 +3,52 @@
...
 
-CREW_VERBOSE = ARGV.intersect?(%w[-v --verbose]) unless defined?(CREW_VERBOSE)
+CREW_VERBOSE ||= ARGV.intersect?(%w[-v --verbose])
 
-HOME = '/home/chronos/user' unless defined?(HOME)
-CREW_PREFIX = '/usr/local' unless defined?(CREW_PREFIX)
-CREW_LIB_PATH = File.join(CREW_PREFIX, 'lib/crew') unless defined?(CREW_LIB_PATH)
-CREW_CONFIG_PATH = File.join(CREW_PREFIX, 'etc/crew') unless defined?(CREW_CONFIG_PATH)
-CREW_META_PATH = File.join(CREW_CONFIG_PATH, 'meta') unless defined?(CREW_CONFIG_PATH)
+HOME ||= '/home/chronos/user'
```
Zopolis4, line 33: Why do we need to do this complex dance of conditional assignment for these variables? Why can't we just require const.rb?

.rubocop.yml
```diff
@@ -24,6 +24,10 @@ Lint/Debugger:
   DebuggerMethods:
     Pry: ~
 
+# Need to allow reloading const.rb from fixup.rb
+Lint/OrAssignmentToConstant:
```
Zopolis4, line 28: If we end up needing to use conditional assignment in const.rb, then disable this rule for that file only, rather than the entire codebase.

lib/buildsystems/cmake.rb
```diff
@@ -5,16 +5,16 @@ class CMake < Package
...
-    super
-    @cmake_build_relative_dir = '.'
-  end
-
   def self.build
-    puts "Additional cmake_options being used: #{@pre_cmake_options.nil? ? '<no pre_cmake_options>' : @pre_cmake_options} #{@cmake_options.nil? ? '<no cmake_options>' : @cmake_options}".orange
+    @cmake_build_relative_dir ||= '.'
     @crew_cmake_options = @no_lto ? CREW_CMAKE_FNO_LTO_OPTIONS : CREW_CMAKE_OPTIONS
+    puts "Additional cmake options being used:".orange
+    method_list = methods.grep(/cmake_/).delete_if { |i| send(i).blank? }
```
Zopolis4, line 12: this feels like a much more complicated solution than the previous method of a puts statement, and also looks like it doesn't provide as much information as before in cases where the option is nil?

lib/package.rb
```diff
@@ -27,7 +27,6 @@ def require_gem(gem_name_and_require = nil, require_override = nil)
 
 def agree_with_default(yes_or_no_question_msg, character = nil, default:)
   yes_or_no_question = yes_or_no_question_msg.lightpurple
-  require_gem('highline')
```
Zopolis4, line 30: We use highline in the lines immediately following, so we should probably still require it in `lib/package.rb`

Zopolis4 (changes requested):

lib/buildsystems/cmake.rb
satmandu: Good call. Should probably only be a property...

lib/fixup.rb
satmandu: It's a good question!

We `load` fixup.rb... but since we already have const.rb required, as I understand it... it won't be loaded again unless we use `load`, which gives us a ton of `warning: already initialized constant ` errors.

The problem is that we NEED const.rb reloaded, because otherwise we will call remove.rb for a package that fixup wants to remove, but then remove.rb will die if crew update was run so long ago the in-memory const.rb doesn't have an essential_packages array constant loaded.

.rubocop.yml
satmandu: Good idea.

lib/package.rb
satmandu: Agreed, but I'm loading highline from `crew` now, so it should already be loaded. (Otherwise I was getting issues with stuff complaining about highline not being loaded.)

lib/buildsystems/cmake.rb
satmandu: By nil you mean the option isn't set? Because in that case it just doesn't show the method.

Zopolis4: Whereas before it would print something like <no_cmake_options>, right?

lib/fixup.rb
satmandu: I am still getting the warning: already initialized constant  errors though... Any suggestions on how to deal with this?

Zopolis4: Oh, that is tricky. Couple things: I've removed the need for essential_deps in #9832, so that's a possible solution.

In the short term, couldn't we load essential_deps from device.json in commands/remove.rb and bypass the need for it to be present in lib/const.rb?

lib/package.rb
Zopolis4: Fair, although I think it's still a good idea to require the necessary files rather than relying on them being implicitly present. (i.e. if we only load lib/package.rb and it's not run through bin/crew, which is the case for all our tests.)

lib/fixup.rb
Zopolis4: > I am still getting the warning: already initialized constant  errors though... Any suggestions on how to deal with this?

My suggestion is to take the warning at face value-- the constants are already initialised, so we don't need to set them again.

lib/package.rb
```diff
@@ -27,7 +27,6 @@ def require_gem(gem_name_and_require = nil, require_override = nil)
```
Zopolis4, line 27: We also don't need to define `require_gem` in `lib/package.rb` anymore, as a regular require should work here as highline will have been installed before we get here.

commands/remove.rb
```diff
@@ -1,5 +1,6 @@
 require 'fileutils'
 require_relative '../lib/const'
+load '../lib/const.rb' if CREW_ESSENTIAL_PACKAGES.blank?
```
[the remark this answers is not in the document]
Zopolis4: image
from testing on a local docker installation without the essential_deps field in device.json

commands/remove.rb
```diff
@@ -18,7 +39,12 @@ def self.remove(pkg, verbose)
     # their dependencies, as those are needed for ruby and crew to run,
     # and thus should not be removed.
     # essential_deps = recursive_deps(CREW_ESSENTIAL_PACKAGES)
-    essential_deps = device_json[:essential_deps]
+    if device_json[:essential_deps].blank?
```
Zopolis4, line 42: is there a situation in which this will be empty? it'll be nil if the field isnt in device.json, which does happen and is the cause of the issue, but I can't think of a scenario in which it'll be empty.

lib/fixup.rb
satmandu: Yes, but it looks really ugly... and we don't want to throw unnecessary warnings. :/

commands/remove.rb
satmandu: I'm just trying to handle the nil error... .`nil?` is complaining... but that could be a race condition from having older code in the mix. I'm just opening the M121 container for testing...

lib/fixup.rb
Zopolis4: Hence why I am suggesting not defining the constants again in lib/fixup.rb (i.e. why do we need to define HOME again?)

lib/buildsystems/qmake.rb
```diff
@@ -4,7 +4,7 @@ class Qmake < Package
   property :qmake_build_extras, :qmake_install_extras
 
   def self.build
-    system "QMAKE_CXX='g++ #{ARCH == 'x86_64' && LIBC_VERSION.to_f >= 2.35 ? File.join(CREW_LIB_PREFIX, 'libC.so.6').to_s : ''}' qmake"
+    system "QMAKE_CXX='g++ #{ARCH == 'x86_64' && Gem::Version.new(LIBC_VERSION.to_s) >= Gem::Version.new('2.35') ? File.join(CREW_LIB_PREFIX, 'libC.so.6').to_s : ''}' qmake"
```
Zopolis4, line 7: Why this change? Isn't it just a more complicated way of doing the same thing?

satmandu: There was a discussion earlier of version comparisons possibly breaking in the glibc builds, so as per that I've been unifying the usage with the gem version comparisons.

uberhacker: Yes, when comparing versions 2.4 is lower than 2.37, however 2.40 is higher than 2.37.  That is where the potential problem may occur.

Zopolis4: Yes, but the lowest glibc version we will encounter is 2.23, so we don't need to worry about that. All the glibc versions we deal with have double digit minor versions.

satmandu:
FYI I'm really mucking around here, so this branch isn't usable at the moment...

Zopolis4:
On that note, in future maybe split up separate changes to PRs? This PR is attempting to fix at least 3 issues by my count, all of which are pretty urgent, and because all the changes are combined, difficulties in solving one is holding up the fixes for all the others.

satmandu:
Yeah that's on me. Sorry about that.

lib/package_utils.rb
```diff
@@ -3,10 +3,20 @@
...
+  keepers = keeper_keys.to_set
+  arr.map { |h| h.select { |k, _| keepers.include?(k) } }
+end
+
 class PackageUtils
-  def self.installed?(pkg_name)
+  def self.generate_installed_packages
     device_json = JSON.load_file(File.join(CREW_CONFIG_PATH, 'device.json'))
-    return device_json['installed_packages'].any? { |elem| elem['name'] == pkg_name }
+    return keep_keys(device_json['installed_packages'], ['name']).flat_map(&:values).to_set
```
Zopolis4, line 14: Given that we have to iterate over the array anyways to convert it to a set, we aren't actually getting speedups here, unfortunately:
image
```ruby
require 'json'
require 'set'
require 'benchmark'

CREW_CONFIG_PATH = '/usr/local/etc/crew'

# Method 1: Direct check for package name
def method_1(pkg_name)
  device_json = JSON.load_file(File.join(CREW_CONFIG_PATH, 'device.json'))
  device_json['installed_packages'].any? { |elem| elem['name'] == pkg_name }
end

# Method 2: Using keep_keys to filter and check
def keep_keys(arr, keeper_keys)
  keepers = keeper_keys.to_set
  arr.map { |h| h.select { |k, _| keepers.include?(k) } }
end

def method_2(pkg_name)
  device_json = JSON.load_file(File.join(CREW_CONFIG_PATH, 'device.json'))
  target = keep_keys(device_json['installed_packages'], ['name']).flat_map(&:values).to_set
  target.include?(pkg_name)
end

# Package name to search for
pkg_name = 'example_package_name'
installed_pkg_name = 'ruby'

# Benchmarking the two methods
Benchmark.bm do |x|
  x.report("Method 1, not installed package: ") { 1000.times { method_1(pkg_name) } }
  x.report("Method 2, not instapped package: ") { 1000.times { method_2(pkg_name) } }
  x.report("Method 1, installed package: ") { 1000.times { method_1(installed_pkg_name) } }
  x.report("Method 2, instapped package: ") { 1000.times { method_2(installed_pkg_name) } }
end
```

satmandu: Trying to only iterate over the array once... That means creating sets for installed packages once.

satmandu: So the idea is to have the one installed_packages set being used for multiple `installed?` checks.

Can you benchmark that?

satmandu:
Argh. Package rename is still broken. I'm probably reloading the json before writing it...

Zopolis4:
you've got stuff still symbolised in command/remove.rb

maybe a quick grep for 'device_json[:' to ensure u don't miss anything

satmandu:
> you've got stuff still symbolised in command/remove.rb
> 
> maybe a quick grep for 'device_json[:' to ensure u don't miss anything

But that shouldn't matter, because we aren't passing symbolized stuff back and forth between remove.rb and everything else.

satmandu:
Oh, we also need a superseded_by field in our update array.

If py3_pre_commit is installed, then we can remove linter.

Also, if a package file isn't there, prompt to remove package if marked as installed, especially if there aren't any file lists present?

Seeing that issue now with incomplete removal...

lib/fixup.rb
satmandu: HOME is no longer defined again.

commands/remove.rb
satmandu: Fixed the root cause.

lib/buildsystems/qmake.rb
satmandu: Honestly I don't care which way we go on this. I just want it to be uniform if possible. I think that glibc versions are a special case, and we should avoid special cases when possible...

lib/fixup.rb
Zopolis4: It's not just HOME-- why do we need to define most of these constants again, and why do we need to set all the constants in `lib/const.rb` as `||=` when we only redefine a few of them?

and again, all of this could be bypassed if we didn't load `essential_deps` from device.json

lib/buildsystems/qmake.rb
Zopolis4: Let's unify it to just the float comparison-- `Gem::Version.new` is overkill here.

uberhacker: I just explained why we need to compare versions semantically a few comments earlier.

uberhacker: Bedtime reading material: https://semver.org/

lib/fixup.rb
satmandu: You need the `||=` construction because we are using ` load 'lib/const.rb'` instead of using `require 'lib/const.rb'` to reload `const.rb` from `fixup.rb`, and that triggers tons of errors about trying to redefine constants.

(We reload const.rb from fixup.rb right after a git pull, so we may have new or changed constants there we need to check for.)

`||=` does a _define if not already defined_, which appears to work for most stuff, except anything we are defining with a call to check an environment variable, so that's mostly ideal for a file of constant definitions that we are about to reload.

Zopolis4: We could try a system call to `lib/fixup.rb` so its not dealing with any scope problems and loads everything afresh? (i.e. `system ./lib/fixup.rb` and just have it be a regular script)

lib/buildsystems/qmake.rb
Zopolis4: > I just explained why we need to compare versions semantically a few comments earlier.

And I explained a few comments later why that is not relevant in the case of comparing glibc versions.

glibc version 2.4 was released in 2005, 5 years before the first chromebook released. and again, the lowest glibc version we will encounter is 2.23.

lib/fixup.rb
```diff
@@ -4,30 +4,107 @@
...
+             end
+  require requires
+end
+require_gem('highline')
 
 # All needed constants & variables should be defined here in case they
 # have not yet been loaded or fixup is being run standalone.
 
-CREW_VERBOSE = ARGV.intersect?(%w[-v --verbose]) unless defined?(CREW_VERBOSE)
+CREW_VERBOSE ||= ARGV.intersect?(%w[-v --verbose]) unless defined?(CREW_VERBOSE)
```
Zopolis4, line 32: You don't need `unless defined?(CREW_VERBOSE)` if you're using `||=`, do you?

satmandu: You would think not, but for some reason for the way these variables are being set ruby complains otherwise. 🤷🏿

satmandu: I thought it was annoying too...

Every time I have both is because the `||=` wasn't sufficient.

satmandu:
I'm getting this only on the arm container:

```
chronos@rpi51-armv7l:2.27 M91 /usr/local/lib/crew/packages (master >|SPARSE)$ crew whatprovides bin/convert
/usr/local/lib/crew/lib/package.rb:104:in `read': no implicit conversion of nil into String (TypeError)

    class_eval(File.read(pkg_file, encoding: Encoding::UTF_8), pkg_file) unless const_defined?("Package::#{class_name}")
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        from /usr/local/lib/crew/lib/package.rb:104:in `load_package'
        from /usr/local/lib/crew/commands/whatprovides.rb:11:in `block in whatprovides'
        from /usr/local/lib/crew/commands/whatprovides.rb:7:in `each'
        from /usr/local/lib/crew/commands/whatprovides.rb:7:in `flat_map'
        from /usr/local/lib/crew/commands/whatprovides.rb:7:in `whatprovides'
        from /usr/local/bin/crew:1804:in `block in whatprovides_command'
        from /usr/local/bin/crew:1803:in `each'
        from /usr/local/bin/crew:1803:in `whatprovides_command'
        from /usr/local/bin/crew:1826:in `<main>'
```

Any idea where this is from?

satmandu:
@Zopolis4 @supechicken Any ideas?

satmandu:
(I don't think there are any changes relevant to package.rb that would cause that in this PR?, and it isn't an issue on x86_64 or i686...)

satmandu:
I need to add a unit test for `crew whatprovides`...

satmandu:
Ok `crew whatprovides` is broken on arm on master, so this PR isn't to blame, so I'm flipping this off of draft again. That's another issue that can be fixed in the future...

lib/package.rb
satmandu: Resolving since a `require_gem('highline')` is still in `lib/package.rb`.

lib/package_utils.rb
satmandu: Let's figure this portion out another day, since this is only in `lib/fixup.rb` for now...

lib/buildsystems/qmake.rb
uberhacker: Remember, versions can be of the form 2.37.1.  So, if LIBC_VERSION is converted to float in this instance, the value would be 2.37 which compared to an earlier version may or may not yield the desired result.  We cannot assume glibc will always have one decimal point.  A lot of the earlier glibc versions had 2 decimal points so we better use the proper version comparison.

packages/glibc_lib237.rb
```diff
@@ -32,7 +33,7 @@ def self.install
     @filelist = File.readlines(@filelist_path, chomp: true).sort
 
     @filelist.each do |filename|
-      next unless filename.include?('.so') || filename.include?('bin/')
+      next unless (filename.include?('.so') || filename.include?('bin/')) && File.file?(filename)
```
uberhacker, line 36: Should `File.file?(filename)` be the first comparison?
```
next unless File.file?(filename) && (filename.include?('.so') || filename.include?('bin/'))
```

satmandu: Good point.

packages/ruby_pry.rb
```diff
@@ -1,5 +1,4 @@
 require 'buildsystems/ruby'
-require 'highline/import'
```
uberhacker, line 2: This is a good fix and much needed.

packages/glibc_lib237.rb
satmandu: Is it better to do a string comparison before we look for a file on disk?

satmandu: (Because the string comparison is faster than checking to see if the file is there?)

uberhacker (approved):

PR #10394, opened 2024-09-01
Remove need for essential_deps field in device.json

Zopolis4:
## Description
The definition of what is an 'essential' package has been confused. Essential packages are those that are absolutely required for chromebrew to run, and only chromebrew. If you remove lz4, zstd can no longer decompress `.tar.lz` binaries-- this is not required for ruby to run, and thus is not an essential package.

As such, the list of essential packages is actually quite short, and the only complexity required is in ensuring we have the right `glibc_lib_*` or `glibc_build_*` package.

##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=ughhhh crew update \
&& yes | crew upgrade
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -3,7 +3,7 @@
 require 'etc'
 
 OLD_CREW_VERSION ||= defined?(CREW_VERSION) ? CREW_VERSION : '1.0'
-CREW_VERSION ||= '1.56.9' unless defined?(CREW_VERSION) && CREW_VERSION == OLD_CREW_VERSION
+CREW_VERSION ||= '1.57.0' unless defined?(CREW_VERSION) && CREW_VERSION == OLD_CREW_VERSION
 
 # Kernel architecture.
 KERN_ARCH ||= Etc.uname[:machine]
@@ -41,7 +41,10 @@ else
   HOME ||= File.join(CREW_PREFIX, Dir.home)
 end
 
-CREW_ESSENTIAL_PACKAGES ||= %w[curl gcc_lib glibc gmp lz4 ruby xzutils zlib zstd]
+# These are packages that crew needs to run-- only packages that the bin/crew needs should be required here.
+# lz4, for example, is required for zstd to have lz4 support, but this is not required to run bin/crew.
+# The LIBC_VERSION ternary is to reflect the change from glibc_build to glibc_lib as the source of essential libraries starting at glibc 2.35.
+CREW_ESSENTIAL_PACKAGES ||= %W[gcc_lib #{LIBC_VERSION.to_f > 2.34 ? "glibc_lib#{LIBC_VERSION.delete('.')}" : "glibc_build#{LIBC_VERSION.delete('.')}"} gmp ruby zlib zstd]
 
 CREW_IN_CONTAINER ||= File.exist?('/.dockerenv') || ENV.fetch('CREW_IN_CONTAINER', false) unless defined?(CREW_IN_CONTAINER)
 
@@ -354,7 +357,7 @@ CREW_DOCOPT ||= <<~DOCOPT
     crew remove [options] [-v|--verbose] <name> ...
     crew search [options] [-v|--verbose] <name> ...
     crew sysinfo [options] [-v|--verbose]
-    crew update [options] [-v|--verbose] [<compatible>]
+    crew update [options] [-v|--verbose]
     crew upgrade [options] [-k|--keep] [-s|--source] [-v|--verbose] [<name> ...]
     crew upload [options] [-v|--verbose] [<name> ...]
     crew whatprovides [options] <pattern> ...
--- a/tests/commands/remove.rb
+++ b/tests/commands/remove.rb
@@ -1,7 +1,6 @@
 require 'minitest/autorun'
 require_relative '../../commands/remove'
 require_relative '../../lib/const'
-require_relative '../../lib/convenience_functions'
 require_relative '../../lib/package_utils'
 
 # Add lib to LOAD_PATH
@@ -12,11 +11,10 @@ String.use_color = false
 
 class RemoveCommandTest < Minitest::Test
   def setup
-    essential_deps = ConvenienceFunctions.load_symbolized_json[:essential_deps]
-    @random_essential_package_name = essential_deps[rand(0...(essential_deps.length - 1))]
+    @random_essential_package_name = CREW_ESSENTIAL_PACKAGES.sample
     puts <<~ESSENTIAL_PACKAGE_REMOVAL_TEST_EOF
 
-      Testing the removal of essential package #{@random_essential_package_name}, which was picked at random from one the essential packages: #{essential_deps.join(', ')}
+      Testing the removal of essential package #{@random_essential_package_name}, which was picked at random from one the essential packages: #{CREW_ESSENTIAL_PACKAGES.join(', ')}
       (This should fail.)
 
     ESSENTIAL_PACKAGE_REMOVAL_TEST_EOF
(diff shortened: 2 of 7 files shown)


satmandu (changes requested):
I would define essential packages as those required to keep Chromebrew functions from breaking.

Being able to install packages requires zstd and its dependencies, so I don't understand why zstd isn't essential...

satmandu:
Also, we want to make sure zstd isn't removed, because otherwise we are going to get support issues with people who didn't realize they were breaking their systems.

Zopolis4:
> Being able to install packages requires zstd and its dependencies, so I don't understand why zstd isn't essential

zstd is essential, which is why I've kept it as an essential package.

Zopolis4:
image
as you can see, removing lz4 does not break the basic functionality of crew, hence why it is not an essential package.

satmandu:
```
chronos@cheekon-x86_64:2.27 M90 /usr/local/lib/crew/packages (master >|SPARSE)$ upx -d `which zstd`
                       Ultimate Packer for eXecutables
                          Copyright (C) 1996 - 2024
UPX 4.2.4       Markus Oberhumer, Laszlo Molnar & John Reiser    May 9th 2024

        File size         Ratio      Format      Name
   --------------------   ------   -----------   -----------
   1102781 <-    382512   34.69%   linux/amd64   zstd

Unpacked 1 file.
chronos@cheekon-x86_64:2.27 M90 /usr/local/lib/crew/packages (master >|SPARSE)$ ../tools/getrealdeps.rb zstd
Checking for the runtime dependencies of zstd...

Package zstd has runtime library dependencies on these packages:
  depends_on 'gcc_lib' # R
  depends_on 'glibc' # R
  depends_on 'lz4' # R
  depends_on 'xzutils' # R
  depends_on 'zlib' # R
chronos@cheekon-x86_64:2.27 M90 /usr/local/lib/crew/packages (master >|SPARSE)$ lddtree `which zstd`
/usr/local/bin/zstd (interpreter => /lib64/ld-linux-x86-64.so.2)
    libz.so.1 => /usr/local/lib64/libz.so.1
    liblzma.so.5 => /usr/local/lib64/liblzma.so.5
    liblz4.so.1 => /usr/local/lib64/liblz4.so.1
    libpthread.so.0 => /usr/local/lib64/libpthread.so.0
    libc.so.6 => /usr/local/lib64/libc.so.6
    ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2
chronos@cheekon-x86_64:2.27 M90 /usr/local/lib/crew/packages (master >|SPARSE)$
```

satmandu:
All dependencies of zstd need to be installed...

satmandu:
Otherwise installs break.

Zopolis4:
> All dependencies of zstd need to be installed...

This is not the case. As I have shown, `lz4` can be removed and crew will still function.

satmandu:
> > All dependencies of zstd need to be installed...
> 
> This is not the case. As I have shown, `lz4` can be removed and crew will still function.

My experience is that installs break.

Please confirm that installs do not break with your changes by passing the REPO and BRANCH flags to the installer after deleting Chromebrew in a container.

We should really just have a unit test for this triggered with any changes in `install.sh` or to stuff like packages in `buildessential` and `core` when packages in those have their dependencies change, just to make sure installs don't break. Would that be overkill?

(Maybe a locally triggered unit test that doesn't burden GitHub's Action servers would make the most sense? Doing a test install takes about 15 min on my systems at most.)

uberhacker:
> > All dependencies of zstd need to be installed...
> 
> This is not the case. As I have shown, `lz4` can be removed and crew will still function.

Yes, but do fresh installs still work with it removed?  We cannot break fresh installs under any circumstances or we will antagonize new users and create a stampede of incoming issues.

satmandu:
(I'm not trying to be dogmatic for the sake of being dogmatic — just being overly protective because it is annoying to track down breakage after the fact. It would be nice to delegate that protectiveness to some automatic mechanism like we have done with our other unit tests, because refactoring can be really nice to streamline and cut out excess.)

Zopolis4:
I have just tested a fresh install using this branch, everything works.

satmandu:
> I have just tested a fresh install using this branch, everything works. 

Just keep in mind that dependencies are lowest common denominator for all of our systems for the most part.

Do installs work on i686 + pre-M126/post-M126 systems on both x86_64 + Arm...

There is a reason that essential packages is currently set to expand to include all dependencies. 

If zstd doesn't have dependencies installed, it will try to use system libraries. If those aren't available... I think crew then tries to use the system zstd... That isn't available on the oldest systems.

Zopolis4:
What i'm trying to convey here is that zstd doesn't need some dependencies in order to function for chromebrew's usage of it. zstd is able to decompress `.tar.zst` files and run ruby without `lz4` installed, so we don't need it for basic functionality.

Zopolis4:
i.e. we aren't breaking zstd by removing lz4. we might be removing its ability to decompress `.tar.lz` files, but it still works for ruby's purposes.

satmandu:
Also, I expanded essential deps because installs were failing...

Zopolis4:
> Also, I expanded essential deps because installs were failing...

In what situations were they failing? If I can test this PR and it works in that case, then we should be pretty good.

Zopolis4:
any update on this? I'd like to get this PR in before it joins the pile of "out of sight out of mind" pull requests that have been waiting on reviews for months

and i've got some cleanup changes after this, including dealing with the warning constant message on crew startup, so i'd like to get those in as well

satmandu:
Will look at this again after I finish this massive python cleanup/update I'm on final approach for...

Zopolis4:
Re: `generate_compatible`-- I don't see anything in that code that does anything other than check essential deps and package compatibility-- what exactly was your usecase?

I'll check out that other bug soon, swapped to a new laptop and distro and so am still setting up my development environment.

tests/commands/remove.rb
```diff
@@ -56,7 +54,7 @@ def test_remove_essential_package
   def test_remove_package_with_essential_file
     assert_output(true) do
       until PackageUtils.installed?(@package_with_essential_file)
-        system "crew install -d #{@package_with_essential_file}", %i[out err] => File::NULL
+        system "yes | crew install -d #{@package_with_essential_file}", %i[out err] => File::NULL
```
uberhacker, line 57: We can use `-f` now instead of `yes | ...`.

Zopolis4: Fixed, thanks.

lib/const.rb
```diff
@@ -41,7 +41,10 @@
   HOME ||= File.join(CREW_PREFIX, Dir.home)
 end
 
-CREW_ESSENTIAL_PACKAGES ||= %w[curl gcc_lib glibc gmp lz4 ruby xzutils zlib zstd]
+# These are packages that crew needs to run-- only packages that the bin/crew needs should be required here.
+# lz4, for example, is required for zstd to have lz4 support, but this is not required to run bin/crew.
+# The LIBC_VERSION ternary is to reflect the change from glibc_build to glibc_lib as the source of essential libraries starting at glibc 2.35.
+CREW_ESSENTIAL_PACKAGES ||= %W[gcc_lib #{LIBC_VERSION.to_f > 2.34 ? "glibc_lib#{LIBC_VERSION.delete('.')}" : "glibc_build#{LIBC_VERSION.delete('.')}"} gmp ruby zlib zstd]
```
uberhacker, line 47: Bump crew version?

Zopolis4: Fixed.

uberhacker (dismissed):

Zopolis4:
Unit tests appear to be hanging-- I'll look into this when I get back.

satmandu:
It looks like removing packages previously deemed essential, such as `gcc_lib,` is breaking the container unit tests that check to ensure that trying to remove essential packages doesn't break things.

tests/commands/remove.rb
Zopolis4: This causes hangs, so I've reverted it.

Zopolis4:
FIxed-- the issue was just about using `yes | ...` instead of `crew -f` in the unit tests.

> It looks like removing packages previously deemed essential, such as `gcc_lib,` is breaking the container unit tests that check to ensure that trying to remove essential packages doesn't break things.

`gcc_lib` is still listed as an essential package and cannot be removed.

satmandu (approved):

PR #10396, opened 2024-09-02
mysql 8.4.2-icu75.1 -> 9.0.1

Zopolis4:
## Description
Updates mysql to 9.0.1, moves the icu version requirement to the depends_on call.

## Additional information
We should try and avoid using `min_glibc` unless it is 100% necessary, as this cuts off tons of devices. We should really only be using `min_glibc` on precompiled proprietary binary packages where we can't do anything about the glibc requirements. I'll go through packages and fix them, but i'd appreciate it if we stopped adding these requirements when there is no reason for them.

##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

##
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=1really crew update \
&& yes | crew upgrade
```

### DIFF
--- a/packages/mysql.rb
+++ b/packages/mysql.rb
@@ -3,22 +3,21 @@ require 'buildsystems/cmake'
 class Mysql < CMake
   description "MySQL Community Edition is a freely downloadable version of the world's most popular open source database"
   homepage 'https://www.mysql.com/'
-  version '8.4.2-icu75.1'
+  version '9.0.1'
   license 'GPL-2'
   compatibility 'x86_64' # Only 64-bit platforms are supported, so this will work on aarch64 userspaces once those are supported.
-  min_glibc '2.37'
-  source_url 'https://cdn.mysql.com/Downloads/MySQL-8.4/mysql-8.4.2.tar.gz'
-  source_sha256 '5657a78dc86bf0bf2227e0b05f8de5a2c447a816a112ffa26fa70083bcbe9814'
+  source_url 'https://github.com/mysql/mysql-server.git'
+  git_hashtag "mysql-#{version}"
   binary_compression 'tar.zst'
 
   binary_sha256({
-     x86_64: '5ce6d8be4a2660d05b7ec4892b09f17a7236881fcc34f9756c6102c5d2539791'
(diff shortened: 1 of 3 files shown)


packages/mysql.rb
```diff
@@ -3,22 +3,21 @@
...
   binary_sha256({
-     x86_64: '5ce6d8be4a2660d05b7ec4892b09f17a7236881fcc34f9756c6102c5d2539791'
+     x86_64: '78f197a0cc31d6657bc682cbbb5be958326872eef0b35d93d7633f59646265d8'
   })
 
   depends_on 'boost' => :build
   depends_on 'gcc_lib' # R
   depends_on 'glibc_lib' # R
-  depends_on 'icu4c' => :build
+  depends_on 'icu4c', '== 75.1'
```
uberhacker, line 20: Does this work?

packages/mysql.rb
```diff
@@ -93,4 +84,10 @@ def self.postremove
         -DWITH_UNIT_TESTS=OFF" # Since we don't currently run the tests, there's no point in building them.
   # 52/136 Test  #52: merge_small_tests ..................................***Failed   62.42 sec
   # run_tests
+  cmake_install_extras do
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/etc/bash.d"
+    File.write "#{CREW_DEST_PREFIX}/etc/bash.d/01-mysql", <<~EOF
+      [ -x #{CREW_PREFIX}/bin/mysql.server ] && #{CREW_PREFIX}/bin/mysql.server start
+    EOF
+  end
```
uberhacker, line 92: Was this built with the lowest glibc container?

Zopolis4: It was built on the M90 container, which has glibc 2.27

packages/mysql.rb
Zopolis4: yes

uberhacker: Great

packages/mysql.rb
uberhacker: Excellent.

uberhacker (approved):

PR #10400, opened 2024-09-03
v4l-utils 1.22.1 -> 1.28.1

Zopolis4:
## Description
Update v4l-utils, and drop the qt5_base dependency so we don't have recursion via `v4l-utils- -> qt5_base -> ffmpeg -> v4l-utils`.

The circular dependency situation still is not perfect, but this makes progress on avoiding them in situations where that is possible (unlike, say, harfbuzz and that debacle.)

##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
##
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=bird crew update \
&& yes | crew upgrade
```


packages/v4l_utils.rb
```diff
@@ -1,42 +1,25 @@
...
 
-  depends_on 'alsa_lib'
-  depends_on 'eudev'
-  depends_on 'libglu'
   depends_on 'libjpeg_turbo'
-  depends_on 'mesa' unless ARCH == 'i686'
-  depends_on 'qt5_base' unless ARCH == 'i686'
-  depends_on 'sdl2_image'
-  depends_on 'gcc_lib' # R
-  depends_on 'glibc' # R
```
satmandu, line 27: Doesn't depend upon glibc?

uberhacker: Are we certain all these other dependencies are not needed?

Zopolis4: yup

uberhacker (approved):

PR #10403, opened 2024-09-03
Revert PR #10336.

uberhacker:
Revert to fix broken installation.


satmandu:
This won't be needed. A small change to `install.sh` should fix installs. We just need to load the SSL env variables before we start using `crew` in the installer.

PR #10426, opened 2024-09-05
Properly rename pygobject to py3_pygobject in depends_on calls

Zopolis4:
## Description
Cleaning up after #10414.

Fixes the crash that happens when any package depending on pygobject is loaded:
image

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=floor crew update \
&& yes | crew upgrade
```


satmandu (approved):
Thanks for fixing this.

uberhacker (approved):

Zopolis4:
i686 unit tests arent running because one non-i686 compatible package was changed... should probably fix that although installing every changed package is taking long enough already

satmandu (approved):

PR #10431, opened 2024-09-06
Remove old runtime dependencies when running getrealdeps.rb

Zopolis4:
## Description
the lists of runtime dependencies for some packages are pretty long-- some of them r just left over from when the package used to depend on something at runtime.

we don't touch anything without `# R`, because the packager added that manually so we don't know enough to delete it, but anything added by `getrealdeps.rb` can be deleted by it because we have the same amount of information.

i have ran this on ffmpeg as an example, and also manually deleted the `libfdk_aac` dependency because it was, to the best of my understanding, triggering a bug where if a dependency is present both as a build and runtime dependency `getrealdeps.rb` fails to see that it is already there.

I have WIP plans for a larger refactor and integration of `getrealdeps.rb`, but I thought it best to split this out as it will have beneficial effects the quicker it gets into the build process.

##
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=flow crew update \
&& yes | crew upgrade
```

### DIFF
--- a/packages/ffmpeg.rb
+++ b/packages/ffmpeg.rb
@@ -32,19 +32,16 @@ class Ffmpeg < Package
   depends_on 'intel_media_sdk' if ARCH == 'x86_64' && CREW_IS_INTEL # R
   depends_on 'jack' # R
   depends_on 'ladspa' # ?
-  depends_on 'leptonica' # R
   depends_on 'libaom' # R
   depends_on 'libass' # R
   depends_on 'libavc1394' # R
   depends_on 'libbluray' # R
   depends_on 'libdc1394' => :build
   depends_on 'libdrm' # R
-  depends_on 'libfdk_aac' => :build
   depends_on 'libfdk_aac' # R
   depends_on 'libfrei0r' => :build
   depends_on 'libglvnd' # R
   depends_on 'libiec61883' # R
-  depends_on 'libjpeg_turbo' # R
   depends_on 'libjxl' # R
   depends_on 'libmodplug' # R
   depends_on 'libmp3lame' # R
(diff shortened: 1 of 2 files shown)


Zopolis4:
well, at least we know rubocop is working :p

Zopolis4:
a little too well, I think Style/CombinableLoops is being overzealous

Zopolis4:
Reported as rubocop/rubocop#13202-- in the meantime, I'll sort this out manually.

tools/getrealdeps.rb
```diff
@@ -107,6 +107,9 @@ def main(pkg)
   pkgdeps = pkgdeps.map { |i| i.gsub(/glibc_build.*/, 'glibc') }.uniq
   pkgdeps = pkgdeps.map { |i| i.gsub(/glibc_lib.*/, 'glibc_lib') }.uniq.map(&:strip).reject(&:empty?)
 
+  # Leave early if we didn't find any dependencies.
+  return if pkgdeps.empty?
```
satmandu, line 111: Won't this exit if a package incorrectly has no dependencies listed?

That sounds like a bad idea?

Zopolis4: No this exits if we didn't find any runtime dependencies.

packages/ffmpeg.rb
```diff
@@ -25,26 +25,22 @@ class Ffmpeg < Package
   depends_on 'fontconfig' # R
   depends_on 'freetype' # R
   depends_on 'fribidi' # R
-  depends_on 'gcc_lib' # R
```
satmandu, line 28: Do you get gcc_lib as a dependency if you run the script in an arm container?

Zopolis4: Ah, I do. hm

Zopolis4: Fixed.

satmandu (approved):

PR #10436, opened 2024-09-08
Python => 3.12.6

satmandu:
- Add a script to check py3_ packages for updates. (@Zopolis4 Can we create a GitHub action to run this and the one for ruby scripts to automatically generate PRs?)
- Fix i686 container builds in `installer.sh`
- Migrate most py3_ packages to use `pip`.
- Adds `ruby_debug` gem so that we can try to avoid the error messages for it being out of date...
- Have `lib/package.rb` specify which file is missing when it tries to load a missing package file.
- Since we can now update py3 package file versions with a script, have `buildsystems/pip` install the specific version of the pip package in the package file.

[Pypi cache](https://docs.gitlab.com/ee/user/packages/pypi_repository/) I am integrating can be set thus:
```
pip config --user set local.index-url https://gitlab.com/api/v4/projects/26210301/packages/pypi/simple
pip config --user set local.extra-index-url https://pypi.org/simple
pip config --user set local.trusted-host gitlab.com
```

##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [ ] `x86_64`
- [ ] `i686`
- [ ] `armv7l` 

##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=python crew update \
&& yes | crew upgrade
```


lib/buildsystems/pip.rb
```diff
@@ -9,6 +9,11 @@ class Pip < Package
   property :pip_install_extras, :pre_configure_options
 
   def self.install
+    # Make sure Chromebrew pypi variables are set:
+    Kernel.system 'pip config --user set local.index-url https://gitlab.com/api/v4/projects/26210301/packages/pypi/simple' unless `pip config get local.index-url`.chomp == 'https://gitlab.com/api/v4/projects/26210301/packages/pypi/simple'
+    Kernel.system 'pip config --user set local.extra-index-url https://pypi.org/simple' unless `pip config get local.extra-index-url`.chomp == 'https://pypi.org/simple'
+    Kernel.system 'pip config --user set local.trusted-host gitlab.com' unless `pip config get local.trusted-host`.chomp == 'gitlab.com'
```
uberhacker, line 15: Should we also be doing something with `requirements.txt` files that contain the package dependencies on projects?

satmandu: Possibly? But that's a LOT more complicated, especially with versioning...

commands/remove.rb
```diff
@@ -40,12 +40,18 @@ def self.remove(pkg, verbose = CREW_VERBOSE)
...
-      if Kernel.system "gem list -i \"^#{@gem_name}\$\""
-        puts "Uninstalling #{@gem_name} before updating. It's ok if this fails.".orange
+      if Kernel.system "gem list -i \"^#{@gem_name}\$\"", %i[out err] => File::NULL
+        puts "Uninstalling #{@gem_name} before removing gem files. It's ok if this fails.".orange
         system "gem uninstall -aIx --abort-on-dependent #{@gem_name}", exception: false
       end
     end
 
+    # Use pip to first try to remove python packages if package uses pip.
+    if pkg.name.start_with?('py3_')
```
Zopolis4, line 50:
```suggestion
    if pkg.superclass.eql?('Pip')
```

lib/buildsystems/pip.rb
```diff
@@ -2,15 +2,22 @@
 Encoding.default_internal = Encoding::UTF_8
 require 'json'
 require 'package'
+require 'package_utils'
```
Zopolis4, line 5: Why has this dependency been added?

bin/crew
```diff
@@ -1232,7 +1237,7 @@ def resolve_dependencies
 end
 
 def install
-  if !@pkg.in_upgrade && PackageUtils.installed?(@pkg.name)
+  if !@pkg.in_upgrade && PackageUtils.installed?(@pkg.name) && !@pkg.name.start_with?('ruby_')
```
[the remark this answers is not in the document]
satmandu: Does this work?

Running the unit test takes 30 min... 😅

lib/buildsystems/pip.rb
satmandu: Ah, forgot to take that out!  I was trying to use PackageUtils.clean but it just borked things up.

bin/crew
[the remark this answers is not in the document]
Zopolis4: It works in a simple local test I've got, although you might wanna double check locally before pushing a change.

lib/buildsystems/pip.rb
satmandu: (I'll push commits once I confirm that this PR actually passes the unit tests...)

bin/crew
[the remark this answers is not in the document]
satmandu: Great. I'll try to take a look later tonight...

satmandu:
@Zopolis4 if the suggested changes are minor, would you mind if we handle those in a small followup cleanup PR?

The hour it is taking to run unit tests where every touched python package gets installed and uninstalled could be avoided then...

satmandu:
> @Zopolis4 if the suggested changes are minor, would you mind if we handle those in a small followup cleanup PR?
> 
> The hour it is taking to run unit tests where every touched python package gets installed and uninstalled could be avoided then...


(That's assuming that the unit tests actually succeed this time...)

bin/crew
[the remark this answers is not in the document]
satmandu: Ok so that doesn't work.

```ruby
Making 2 links.
Attempting removal since this is an upgrade or reinstall...
/usr/local/lib/crew/commands/remove.rb:41:in `remove': undefined method `superclass' for nil (NoMethodError)

    if @pkg.superclass.eql?('RUBY')
           ^^^^^^^^^^^
        from /usr/local/bin/crew:1283:in `install'
        from /usr/local/bin/crew:1117:in `resolve_dependencies_and_install'
        from /usr/local/bin/crew:424:in `block in upgrade'
        from /usr/local/bin/crew:416:in `each'
        from /usr/local/bin/crew:416:in `upgrade'
        from /usr/local/bin/crew:1791:in `upgrade_command'
        from /usr/local/bin/crew:1826:in `<main>'
```

commands/remove.rb
satmandu: Not doings this because it breaks things...

bin/crew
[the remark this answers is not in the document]
Zopolis4: How do we have nil @pkg values? That seems like a pretty big issue

Zopolis4:
You have a good point with how long the unit tests take-- I haven't fully reviewed this yet, but if I find any nitpicks we can save those changes for a followup PR.

packages/py3_urllib3.rb
```diff
@@ -7,6 +7,7 @@ class Py3_urllib3 < Pip
   license 'MIT'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: this package still has `no_compile_needed` and doesn't have any binaries, so either this line should be removed or binaries should be uploaded.

packages/py3_unsupported_python.rb
```diff
@@ -1,16 +1,17 @@
 require 'buildsystems/pip'
 
 class Py3_unsupported_python < Pip
-  description 'This package can be used as a conditional dependency to indicate lack of support for particular versions of Python.'
```
Zopolis4, line 4: where did the description go?

packages/py3_unsupported_python.rb
```diff
@@ -1,16 +1,17 @@
...
   license 'GPL-3+'
   compatibility 'all'
   source_url 'SKIP'
-
-  binary_sha256({})
+  binary_compression 'tar.zst'
 
   depends_on 'python3' => :build
 
+  no_compile_needed
```
Zopolis4, line 14: Adding both a binary_compression value and no_compile_needed seems misleading-- they're essentially conflicting values, after all.

packages/py3_twine.rb
```diff
@@ -0,0 +1,27 @@
+# Adapted from Arch Linux python-twine PKGBUILD at:
+# https://aur.archlinux.org/cgit/aur.git/plain/PKGBUILD?h=python-twine
+
+require 'buildsystems/pip'
+
+class Py3_twine < Pip
+  description 'A utility for interacting with PyPI'
+  homepage 'https://pypi.python.org/pypi/twine'
+  version '5.1.1-py3.12'
+  license 'APACHE'
```
Zopolis4, line 10: Usually we don't have the licenses in all caps

packages/py3_typogrify.rb
```diff
@@ -7,6 +7,7 @@ class Py3_typogrify < Pip
   license 'BSD'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: This package still has `no_compile_needed`, so a binary compression value here is spurious.

packages/py3_tomli_w.rb
```diff
@@ -7,4 +7,5 @@ class Py3_tomli_w < Pip
   license 'MIT'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: binary_compression value but no binaries?

packages/py3_sphinxcontrib_serializinghtml.rb
```diff
@@ -7,6 +7,7 @@ class Py3_sphinxcontrib_serializinghtml < Pip
   license 'BSD'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: this package still has no_compile_needed, so no point in a `binary_compression`  value

packages/py3_snowballstemmer.rb
```diff
@@ -7,6 +7,7 @@ class Py3_snowballstemmer < Pip
   license 'BSD'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_sortedcontainers.rb
```diff
@@ -7,6 +7,7 @@ class Py3_sortedcontainers < Pip
   license 'Apache-2.0'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_sphinxcontrib_qthelp.rb
```diff
@@ -7,6 +7,7 @@ class Py3_sphinxcontrib_qthelp < Pip
   license 'BSD'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_sphinxcontrib_jsmath.rb
```diff
@@ -7,6 +7,7 @@ class Py3_sphinxcontrib_jsmath < Pip
   license 'BSD'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_sphinxcontrib_htmlhelp.rb
```diff
@@ -7,6 +7,7 @@ class Py3_sphinxcontrib_htmlhelp < Pip
   license 'BSD'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_sphinxcontrib_applehelp.rb
```diff
@@ -7,6 +7,7 @@ class Py3_sphinxcontrib_applehelp < Pip
   license 'BSD'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_sphinxcontrib_devhelp.rb
```diff
@@ -7,6 +7,7 @@ class Py3_sphinxcontrib_devhelp < Pip
   license 'BSD'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_slugify.rb
```diff
@@ -1,23 +1,23 @@
-require 'buildsystems/python'
+require 'buildsystems/pip'
 
-class Py3_slugify < Python
+class Py3_slugify < Pip
   description 'Python-slugify returns unicode slugs.'
   homepage 'https://github.com/un33k/python-slugify/'
-  @_ver = '8.0.1'
-  version "#{@_ver}-py3.12"
+  version '0.0.1-py3.12'
```
Zopolis4, line 6: 8.0.1 to 0.0.1?

packages/py3_smartypants.rb
```diff
@@ -7,6 +7,7 @@ class Py3_smartypants < Pip
   license 'BSD'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_semantic_version.rb
```diff
@@ -7,6 +7,7 @@ class Py3_semantic_version < Pip
   license 'BSD-2'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_requests.rb
```diff
@@ -7,6 +7,7 @@ class Py3_requests < Pip
   license 'Apache-2.0'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_pygobject.rb
```diff
@@ -23,7 +24,12 @@ class Py3_pygobject < Pip
...
   depends_on 'harfbuzz' # R
   depends_on 'libffi' # R
+  depends_on 'py3_meson_python' => :build
   depends_on 'py3_pycairo' => :build
   depends_on 'python3' # R
   depends_on 'wayland' => :build
+
+  meson_options '-Dtests=false'
+
+  no_source_build
```
Zopolis4, line 34: There's a `source_url` value, so this is probably wrong.

packages/py3_pygments.rb
```diff
@@ -7,6 +7,7 @@ class Py3_pygments < Pip
   license 'BSD-2'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_pycparser.rb
```diff
@@ -7,4 +7,5 @@ class Py3_pycparser < Pip
   license 'BSD-3'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values?

packages/py3_py.rb
```diff
@@ -7,6 +7,7 @@ class Py3_py < Pip
   license 'MIT'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_ply.rb
```diff
@@ -7,6 +7,7 @@ class Py3_ply < Pip
   license 'BSD'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_cfgv.rb
```diff
@@ -7,6 +7,7 @@ class Py3_cfgv < Pip
   license 'MIT'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_devedeng.rb
```diff
@@ -3,19 +3,25 @@
...
+     x86_64: 'f5b0e921aca08f0c46b02ff95e5fca125c015be55025148ca5fddc692b7ed6cb'
+  })
 
   depends_on 'python3' => :build
   depends_on 'py3_pycairo'
   depends_on 'py3_pygobject'
   depends_on 'gtk3'
 
-  no_compile_needed
+  no_source_build
```
Zopolis4, line 24: There's still a `source_url`, so we shouldn't have a `no_source_build` value.

packages/py3_dool.rb
```diff
@@ -7,6 +7,7 @@ class Py3_dool < Pip
   license 'GPL3+'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

Zopolis4:
perhaps the fact that CI is taking so long on such large PRs is a subtle hint to us that we should split up pull requests more :wink:

packages/py3_identify.rb
```diff
@@ -7,6 +7,7 @@ class Py3_identify < Pip
   license 'MIT'
   compatibility 'all'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_mat2.rb
```diff
@@ -7,6 +7,7 @@ class Py3_mat2 < Pip
   license 'MIT'
   compatibility 'x86_64'
   source_url 'SKIP'
+  binary_compression 'tar.zst'
```
Zopolis4, line 10: Why is there a `binary_compression` value in this package if it does not have any `binary_sha256` values and it still has `no_compile_needed`?

packages/py3_libxml2.rb
```diff
@@ -37,4 +37,6 @@ def self.install
       system "python3 setup.py install #{PY_SETUP_INSTALL_OPTIONS_NO_SVEM}"
     end
   end
+
+  no_source_build
```
Zopolis4, line 41: libxml2 has a source_url, so presumably this package recieves that one?

Zopolis4 (commented):
couple things here-- mostly the setting of `binary_compression` when `no_compile_needed` is still set, as well as `no_source_url` when there is still a source_url value.

Zopolis4:
I don't wanna burn extra CI time, so if you can guarantee you'll quickly adress these changes in a follow-up PR then we can merge this as-is.

bin/crew
```diff
@@ -249,6 +249,10 @@ end
 def update
   abort "'crew update' is used to update crew itself. Use 'crew upgrade <package1> [<package2> ...]' to update specific packages.".orange if @pkg_name
 
+  # The following is used in fixup.rb to determine if crew update needs to
+  # be run again.
+  @crew_const_git_commit = `git -C #{CREW_LIB_PATH} log -n1 --oneline #{CREW_LIB_PATH}/lib/const.rb`.split.first
```
uberhacker, line 254: Why are we assigning a variable (and calling it a constant) that is not even used anywhere?

Zopolis4: It's used in fixup.rb, as per the comment in the changed lines.

lib/fixup.rb
```diff
@@ -34,27 +34,27 @@ def require_gem(gem_name_and_require = nil, require_override = nil)
...
+current_crew_const_git_commit = `git -C #{CREW_LIB_PATH} log -n1 --oneline #{CREW_LIB_PATH}/lib/const.rb`.split.first
+unless current_crew_const_git_commit == @crew_const_git_commit
   puts 'Restarting crew update since there is an updated crew version.'.lightcyan
   puts "CREW_REPO=#{CREW_REPO} CREW_BRANCH=#{CREW_BRANCH} crew update".orange if CREW_VERBOSE
   exec "CREW_REPO=#{CREW_REPO} CREW_BRANCH=#{CREW_BRANCH} crew update"
 end
 
+load "#{CREW_LIB_PATH}/lib/const.rb"
+load "#{CREW_LIB_PATH}/lib/package.rb"
+load "#{CREW_LIB_PATH}/lib/convenience_functions.rb"
```
uberhacker, line 56: Why `load` and not `require` and why was this section moved down in the file?

tools/core_packages.txt
```diff
@@ -80,6 +85,7 @@ libtool
 libunbound
 libunistring
 libuv
+libversion
```
uberhacker, line 88: Does this work with ruby scripts like version.rb?  Currently, version.rb complains that the ruby-libversion gem is not available when running commands such as `crew check -V cmake`.

uberhacker:
```
$ crew check -V cmake
<internal:/usr/local/lib64/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:136:in `require': cannot load such file -- ruby_libversion (LoadError)
Did you mean?  rubygems/version
	from <internal:/usr/local/lib64/ruby/site_ruby/3.3.0/rubygems/core_ext/kernel_require.rb>:136:in `require'
	from ../tools/version.rb:12:in `<main>'
```

Zopolis4: Libversion is the C library, `ruby-libversion` is the Ruby bindings to that library that version.rb requires. `crew install libversion && gem install ruby-libversion` will satisfy the dependencies of `version.rb`.

bin/crew
uberhacker: I read that.  We're not using it in crew though so why not assign it there instead?

satmandu: It is assigned in crew and also in fixup if not already defined.

tools/core_packages.txt
satmandu: Do we need a `ruby_libversion` gem package?

Zopolis4: Using `gem` on its own works fine, so I don't think there's a particularly strong need, but I don't mind either way.

satmandu: I only added libversion here because the container builds noted that it was being installed by default.

Zopolis4: It's already in core.rb, so...

satmandu: Yeah, I think we were supposed to keep those roughly in sync?

Zopolis4: oh, for some reason i thought we were talking about buildessential.rb, oops

satmandu:
> couple things here-- mostly the setting of `binary_compression` when `no_compile_needed` is still set, as well as `no_source_url` when there is still a source_url value.

Ah yes, my grep-fu and sed-fu is sometimes weak...

That's an easy fix that can be in a subsequent PR...

bin/crew
uberhacker: So what is confusing me is why are we defining it here if there is no logic to use it?  Why not just define in fixup.rb since that is where it's needed?

uberhacker (approved):

satmandu:
> > couple things here-- mostly the setting of `binary_compression` when `no_compile_needed` is still set, as well as `no_source_url` when there is still a source_url value.
> 
> Ah yes, my grep-fu and sed-fu is sometimes weak...
> 
> That's an easy fix that can be in a subsequent PR...

@Zopolis4 I see this:
```bash
for i in $(grep -l no_compile_needed py3*); do grep -q binary_compression $i && echo $i ; done
py3_cfgv.rb
py3_dool.rb
py3_identify.rb
py3_mat2.rb
py3_ply.rb
py3_pygments.rb
py3_py.rb
py3_requests.rb
py3_semantic_version.rb
py3_smartypants.rb
py3_sortedcontainers.rb
py3_sphinxcontrib_applehelp.rb
py3_sphinxcontrib_devhelp.rb
py3_sphinxcontrib_htmlhelp.rb
py3_sphinxcontrib_jsmath.rb
py3_sphinxcontrib_qthelp.rb
py3_sphinxcontrib_serializinghtml.rb
py3_typogrify.rb
py3_unsupported_python.rb
py3_urllib3.rb
```

(Is that ok? That lets us build binaries for those easily...)

Also from doing this:
```bash
for i in $(grep -l no_source_build py3*); do grep source_url $i && grep -v -q "source_url 'SKIP'" && echo $i ; done | less
```
I see that only `py3_devedeng.rb` has an extraneous no_source_build. (The rest have a `source_url 'SKIP'`, which I assume should be ok?)

Zopolis4:
`no_compile_needed` means crew will refuse to create binaries for the package, so there's no need for a `binary_compression` value in that case. Do you intend to upload binaries for these packages? If so, you should remove the  `no_compile_needed` value and upload binaries. If not, remove the `binary_compression` variable.

packages/py3_unsupported_python.rb
Zopolis4: Actually, how did this get past CI? This is removing a description, which should cause `tests/prop_test` to fail.

satmandu:
Hmm... should we rename `no_compile_needed` to `no_binaries`?

Because it's nice to have making binaries be optional... Or maybe we allow binaries to be created for packages that have `no_compile_needed` set?

satmandu:
Yeah I screwed up a bunch of those. Do you want to fix that and CI's test
too?

Zopolis4:
> Hmm... should we rename `no_compile_needed` to `no_binaries`?
> 
> Because it's nice to have making binaries be optional... Or maybe we allow binaries to be created for packages that have `no_compile_needed` set?

The whole point of `no_compile_needed` is that it means binaries aren't required-- if binaries are wanted, then you remove it and upload binaries. Unless I'm misunderstanding you?

PR #10442, opened 2024-09-09
Rename autotools constants to fall in line with the rest of crew

Zopolis4:
## Description
As far as I can tell, these are legacies from when crew only had constants for autotools, so there was no need to name them specially.

## Additional information
oh the CI is gonna take FOREVER on this because its gonna have to reinstall every package huh
maybe we should get a solution that only reinstalls packages if they have new binaries or smth

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=parity crew update \
&& yes | crew upgrade
```


Zopolis4:
honestly the CI on this might take so long that we burn a lot of our free hours, hmm

Zopolis4:
I've manually cancelled this in order to avoid that-- would people mind just taking a look over this and visually confirming its ok?

uberhacker (approved):
A bit of an OCD change but it looks good. :)

PR #10452, opened 2024-09-10
More python cleanups

Zopolis4:
## Description
Also added unit tests to prevent stuff like this from happening again (hopefully).

@satmandu, in regards to the 20 or so packages you added binary_compression values to, are you planning to upload binaries for them? 

##
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=slider crew update \
&& yes | crew upgrade
```

### DIFF
--- a/packages/micro.rb
+++ b/packages/micro.rb
@@ -6,10 +6,6 @@ class Micro < Package
   version '2.0.13'
   license 'MIT'
   compatibility 'all'
-  binary_compression 'tar.zst'
-
-  no_compile_needed
-
   case ARCH
   when 'aarch64', 'armv7l'
     source_url 'https://github.com/zyedidia/micro/releases/download/v2.0.13/micro-2.0.13-linux-arm.tar.gz'
@@ -21,6 +17,8 @@ class Micro < Package
     source_url 'https://github.com/zyedidia/micro/releases/download/v2.0.13/micro-2.0.13-linux64.tar.gz'
     source_sha256 'a50e405d3d09d58f6b2c182429c18537a05f317dc0c3c9cb834b3271362e4781'
   end
+
+  no_compile_needed
 
   def self.install
     system "install -Dm755 micro #{CREW_DEST_PREFIX}/bin/micro"
--- a/packages/natron.rb
+++ b/packages/natron.rb
@@ -10,8 +10,6 @@ class Natron < Package
   source_sha256 '7e8f2ec343f553799f34dce89aa250c410024f17e2c9ccfb5e22544db3e46bb4'
   binary_compression 'tpxz'
 
-  no_compile_needed
-
   binary_sha256({
     x86_64: '35b3c22b67f74d285e71f619179cf36a3a475d0df37dd7dcb104d22a318edce4'
   })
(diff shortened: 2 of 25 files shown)


satmandu:
I haven't build-tested those binaries. But I'd like to leave those as build-optional, as that makes it easier to just update the package versions without a build...

(The issue is that some py3_ packages don't build in the github action container, but do build natively, so those HAVE to have binaries for the unit tests to succeed.)

packages/micro.rb
```diff
@@ -25,4 +21,6 @@ class Micro < Package
   def self.install
     system "install -Dm755 micro #{CREW_DEST_PREFIX}/bin/micro"
   end
+
+  no_compile_needed
```
uberhacker, line 25: Why the move down here?

Zopolis4: The normal layout of a package has 
```
# package information
source_url
version
binary_sha256

# package dependencies
depends_on
depends_on
depends_on

# package actions
def self.patch
  ...
end

def self.install
  ...
end

# package properties
no_lto
no_compile_needed
configure_options
```

I'll try and set up a rubocop cop to enforce this at some point

Zopolis4:
> I haven't build-tested those binaries. But I'd like to leave those as build-optional, as that makes it easier to just update the package versions without a build...

I'll drop the binary_compression values, then, and you can add them back and remove the `no_compile_needed` if required.

satmandu:
> > I haven't build-tested those binaries. But I'd like to leave those as build-optional, as that makes it easier to just update the package versions without a build...
> 
> I'll drop the binary_compression values, then, and you can add them back and remove the `no_compile_needed` if required.

But why have `no_compile_needed`  block building binaries? The plain meaning is no compile is needed...

Zopolis4:
Good point-- maybe we should rename it in a second PR.

satmandu:
(also, just a pet peeve, but an absent `binary_compression` already defaults to `tar.zst` so it shouldn't be needed for uploads that just use `tar.zst`... I think I was annoyed that builds were being done, but without the `binary_compression` set I couldn't use `crew upload`...)

Zopolis4:
binary_compression only defaults to `.tar.zst` when being created in `crew upload` (although that functionality is also broken)-- we specify explicit binary_compression values for clarity.

at any rate, once every package is rebuilt to use `.tar.zst` except the few `no_zstd` ones, we'll be able to drop it anyways, so binary_compression is a temporary solution.

packages/micro.rb
uberhacker: Most of our packages have the boolean properties above the def functions.

Zopolis4: Ah, you're right. I mixed them up with buildsystem properties. Boolean properties above def functions, buildsystem properties below.

satmandu (approved):

packages/micro.rb
```diff
@@ -22,6 +18,8 @@ class Micro < Package
     source_sha256 'a50e405d3d09d58f6b2c182429c18537a05f317dc0c3c9cb834b3271362e4781'
   end
 
+  no_compile_needed
+
```
satmandu, line 22: Rubocop complaining about trailing whitespace.

Can we add a `rubocop -c .rubocop.yml -A .` to whatever pre-commit hooks we have? (Not needed for this PR... Just saying that it seems that rubocop can just do these cleanups for us...)

satmandu (approved):

PR #10453, opened 2024-09-11
Add Rakefile to replace crew test and simplify unit testing.

Zopolis4:
## Description
As per discussions on slack, I'm splitting up my refactor PRs into smaller chunks in order to help them get reviewed.

Behavior changes:
- Replace `crew test` with a Rakefile.
  - `rake` replaces the previous behaviour of `crew test` when in the crew installation directory.

### Run the following to get this pull request's changes locally for testing.
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula8.1 crew update \
&& yes | crew upgrade
```

PR #10456, opened 2024-09-11
Revert #10453

uberhacker:
Tests are failing for everything now with the following:
```
rake aborted!
No Rakefile found (looking for: rakefile, Rakefile, rakefile.rb, Rakefile.rb)

Error: Process completed with exit code 1.
```


Zopolis4:
For what it's worth, the "fix" for this was just to rebase your branch on master. This is good practice by default, and a step that should be taken before resorting to a revert.

satmandu:
Shouldn't we just have the unit test check to see if the .GitHub folder is newer on master, and if so refuse to run?


That would force a merge master...

Zopolis4:
There's a setting in github somewhere to require PRs be up-to-date with master before merge

PR #10461, opened 2024-09-11
Refactor prop_test and buildsystem_test to not rely on instance variables.

Zopolis4:
## Description
more smaller chunks of refactoring 

## Additional information
Performance comparison in `buildsystem_test`, running it on all files.
```
# Before
...
All packages passed buildsystem tests.

real    0m6.341s
user    0m7.919s
sys     0m1.257s
```

```
# After
All packages passed buildsystem tests.

real    0m0.498s
user    0m0.442s
sys     0m0.056s
```

the speed boost here comes from dropping `get_buildsystems` as the method for finding the class of a file, using built-in ruby functionality instead. additionally, a static list of the available buildsystems saves us from the remaining grep call, as well as dropping the need to sort it on errors.

we will need to update the list of buildsystems if we add a new one, but i think its worth a 20x speedup.

##
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula8.2 crew update \
&& yes | crew upgrade
```

### DIFF
--- a/tests/buildsystem_test
+++ b/tests/buildsystem_test
@@ -1,42 +1,49 @@
 #!/usr/bin/env ruby
 
+require_relative '../lib/const'
 require_relative '../lib/color'
+require_relative '../lib/package'
 
-def get_buildsystems(file, field)
-  return `grep ^class #{file} | cut -d' ' -f#{field} | xargs`.chomp
-end
+# Add >LOCAL< lib to LOAD_PATH so that packages can be loaded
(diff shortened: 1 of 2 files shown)


satmandu (approved):

Zopolis4:
Realised that `buildsystem_test` will tend to throw errors when attempting to load a malformed package, and while that would still technically successfully test for that, I thought it best to wrap that nicely and rescue the errors.

uberhacker (approved):
Always welcome speed improvements!

PR #10462, opened 2024-09-12
Remove cycle_test

Zopolis4:
## Description
Unfortunately, this test no longer works, and on top of that we have several packages with cyclical dependencies (i.e. the whole cairo and harfbuzz situation) that we cannot avoid.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=projectregula8.3 crew update \
&& yes | crew upgrade
```


uberhacker:
Why not debug and fix instead of deleting?

Zopolis4:
As per the pr description-- we have unavoidable cyclical dependencies, so a test that checked for them would always fail.

uberhacker:
> As per the pr description-- we have unavoidable cyclical dependencies, so a test that checked for them would always fail.

That is just one example.  What if there are other instances that do matter?

Zopolis4:
> That is just one example. What if there are other instances that do matter?

I don't see the purpose of a test that we know will fail due to reasons we can't fix.

uberhacker:
> > That is just one example. What if there are other instances that do matter?
> 
> I don't see the purpose of a test that we know will fail due to reasons we can't fix.

Again, you are basing your general conclusion on one scenario.

Zopolis4:
now time to fix the errors...
```
chronos@28d82349c8f5 /usr/local/lib/crew/tests $ ruby dep_test
csvkit depends on removed package py3_agate_dfb
Removed dependency test for csvkit failed.
mutter depends on removed package gnome_settings_daemon
Removed dependency test for mutter failed.
textract depends on removed package py3_pptx
Removed dependency test for textract failed.

3 packages failed dependency tests.
```

PR #10463, opened 2024-09-12
Some pip refactoring...

satmandu:
- Now has pip wheel upload to gitlab enabled whenever `crew upload package` is run for a `buildessential/pip` package. (This requires a `GITLAB_TOKEN_USERNAME` env variable, which isn't used for the non-python package uploads. The same `GITLAB_TOKEN` variable can be used though if you already have one generated with a username...
- Adds back the prompt to overwrite package uploads that was *ahem* _accidentally_ removed by a prior PR.
- Adds a `prerelease` property for packages that explicitly excludes the package from being included in the repology json.
- The `py3_twine` package is marked as `prerelease` as well since the release version has some breakage in uploading some wheels.
- pip downloads now look first to the gitlab repo, as we have cached wheels for `i686` and `armv8l` architectures since they aren't provided by default..
- This allows `no_source_build` packages to have binaries generated. This was broken before. This allows me to quickly spin up binaries for packages that take forever to build on arm or i686 without having to mess around with the package files.
- Adds a wrapper script `build_updated_packages.rb` to check for new packages (currently just pip and ruby) and automatically build and then try to upload them. (This won't work unless the relevant env variables for gitlab are set.)
  -  _Yes this is hackish. Feel free to refactor it. I have it working locally to build packages and I find it useful!_
- Also updates some python packages and also builds some new binaries for the ones that already had binaries.
- Also has `lib/downloader.rb` report broken `uri`s that throw errors.

##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=pip crew update \
&& yes | crew upgrade
```


packages/py3_twine.rb
```diff
@@ -1,27 +1,27 @@
...
-
+# Twine 5.1.1 currently breaks on some uploads.
+# Install twine from https://github.com/pypa/twine/pull/1123 instead.
 require 'buildsystems/pip'
 
 class Py3_twine < Pip
   description 'A utility for interacting with PyPI'
   homepage 'https://pypi.python.org/pypi/twine'
-  version '5.1.1-py3.12'
+  version '5.1.2.dev26+gc577765-py3.12'
```
Zopolis4, line 8: Why all this instead of just a commit hash and a url patch?

Zopolis4 (changes requested):

Zopolis4:
> Adds back the prompt to overwrite package uploads that was ahem accidentally removed by a prior PR.

I deliberately removed this, as overwriting package uploads is no longer a problem. Is there a reason you added it back?

packages/py3_twine.rb
satmandu: Because it's much easier to have pip install directly from git than actually download the source and patch ourselves.

This way we just follow the relevant patch branch...

satmandu:
> > Adds back the prompt to overwrite package uploads that was ahem accidentally removed by a prior PR.
> 
> I deliberately removed this, as overwriting package uploads is no longer a problem. Is there a reason you added it back?

Do you mean this?

```ruby
 if output.include?('201 Created')
        puts "curl -Ls #{new_url} | sha256sum" if CREW_VERBOSE
        upstream_sha256 = `curl -Ls #{new_url} | sha256sum`.chomp.split.first
        if upstream_sha256 == new_sha256
          puts output.lightgreen
        else
          if CREW_VERBOSE
            puts "expected sha256 hash=#{new_sha256}"
            puts "upstream sha256 hash=#{upstream_sha256}"
          end
          puts "#{output}. Checksum mismatch. Skipping binary_sha256 update in #{pkg_file}...".lightred
          next
        end
      else
        puts output.lightred
        puts "#{output}. Unable to upload. Skipping binary_sha256 update in #{pkg_file}...".lightred
        next
      end
```

This just skips the upload if there is a hash mismatch, but doesnt ask if you want to replace the file uploaded, right?

Am I missing something?

Zopolis4:
> > > Adds back the prompt to overwrite package uploads that was ahem accidentally removed by a prior PR.
> > 
> > 
> > I deliberately removed this, as overwriting package uploads is no longer a problem. Is there a reason you added it back?
> 
> Do you mean this?
> 
> ```ruby
>  if output.include?('201 Created')
>         puts "curl -Ls #{new_url} | sha256sum" if CREW_VERBOSE
>         upstream_sha256 = `curl -Ls #{new_url} | sha256sum`.chomp.split.first
>         if upstream_sha256 == new_sha256
>           puts output.lightgreen
>         else
>           if CREW_VERBOSE
>             puts "expected sha256 hash=#{new_sha256}"
>             puts "upstream sha256 hash=#{upstream_sha256}"
>           end
>           puts "#{output}. Checksum mismatch. Skipping binary_sha256 update in #{pkg_file}...".lightred
>           next
>         end
>       else
>         puts output.lightred
>         puts "#{output}. Unable to upload. Skipping binary_sha256 update in #{pkg_file}...".lightred
>         next
>       end
> ```
> 
> This just skips the upload if there is a hash mismatch, but doesnt ask if you want to replace the file uploaded, right?
> 
> Am I missing something?

No, what this does is check that the file you just uploaded, when downloaded and hashed again, has the hash value that you expect it to. By the point this code has been ran, the file has been uploaded, hence the checking for the `201 Created` as part of the returned http status from uploading the file.

satmandu:
What happens when there is not a 201 code?

satmandu:
Switching this to draft, as it is breaking on i686....

bin/crew
```diff
@@ -1768,10 +1805,11 @@ def upgrade_command(args) = upgrade(*args['<name>'], build_from_source: @opt_sou
 
 def upload_command(args)
   gitlab_token = ENV.fetch('GITLAB_TOKEN', nil)
+  gitlab_token_username = ENV.fetch('GITLAB_TOKEN_USERNAME', nil)
   upload if args['<name>'].empty?
   args['<name>'].each do |name|
     search name
-    upload(name, @pkg.version, gitlab_token, @pkg.binary_compression)
+    upload(name, @pkg.version, gitlab_token, gitlab_token_username, @pkg.binary_compression)
```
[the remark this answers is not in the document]
satmandu: It's a requirement specifically for the girls pypi server. Maybe the twine pip package used to upload can't handle a blank username?

In any case it's what the gitlab docs say we need...

tools/build_updated_packages.rb
```diff
@@ -0,0 +1,177 @@
...
+    gem gem_name
+  end
+  requires = if require_override.nil?
+               gem_name_and_require.split('/')[1].nil? ? gem_name_and_require.split('/')[0] : gem_name_and_require
+             else
+               require_override
+             end
+  require requires
+end
+require_gem 'highline'
```
uberhacker, line 35: Move to the top with the other require lines?

uberhacker: Or will it not work until after the function is defined?

tools/build_updated_packages.rb
```diff
@@ -0,0 +1,177 @@
...
+      next
+    else
+      puts "#{name.capitalize} #{@version} needs builds uploaded for: #{build.join(' ')}".lightblue
+    end
+    if build.include?(ARCH) && agree_default_yes("\nWould you like to build #{name} #{@version}")
+      system "yes | crew build -f #{pkg}"
+      system "crew upload #{name}"
+    end
+  end
+end
```
uberhacker, line 177: I have to admit, this is great work and I'm sure it took some time.

tools/update_python_pip_packages.rb
```diff
@@ -8,10 +10,20 @@
 py3_files = Dir['py3_*.rb']
 pip_files = `grep -l "^require 'buildsystems/pip'" *.rb`.chomp.split
 relevant_pip_files = (py3_files + pip_files).uniq!
+
+pip_config = `pip config list`.chomp
+system 'pip config --user set global.index-url https://gitlab.com/api/v4/projects/26210301/packages/pypi/simple', %i[err out] => File::NULL unless pip_config.include?("global.index-url='https://gitlab.com/api/v4/projects/26210301/packages/pypi/simple'")
```
uberhacker, line 15: We have this `https://gitlab.com/api/v4/projects/26210301/packages` url hardcoded in several places throughout the repo.  Would it make sense to add a constant?

tools/update_python_pip_packages.rb
```diff
@@ -8,10 +10,20 @@
...
 Parallel.map(relevant_pip_files) do |package|
   pip_name = package.gsub('.rb', '').sub('py3_', '').gsub('_', '-')
-  pip_version = `python -m pip index versions #{pip_name} 2>/dev/null | head -n 1 | awk '{print $2}'`.chomp.delete('()')
-  puts "#{pip_name} version is #{pip_version}"
+  if system("grep -q '^\ \ prerelease' #{package}")
+    puts "#{pip_name.capitalize} is configured to look for a pre-release version."
+    pip_version = `python -m pip index versions --pre #{pip_name} 2>/dev/null | head -n 1 | awk '{print $2}'`.chomp.delete('()')
+  else
+    pip_version = `python -m pip index versions #{pip_name} 2>/dev/null | head -n 1 | awk '{print $2}'`.chomp.delete('()')
+  end
```
uberhacker, line 26: Simplify?
```
  prerelease = ''
  if system("grep -q '^\ \ prerelease' #{package}")
    puts "#{pip_name.capitalize} is configured to look for a pre-release version."
    prerelease = '--pre'
  end
  pip_version = `python -m pip index versions #{prerelease} #{pip_name} 2>/dev/null | head -n 1 | awk '{print $2}'`.chomp.delete('()')
```

tools/build_updated_packages.rb
satmandu: The function is required before we use it...

tools/build_updated_packages.rb
satmandu: Yeah, this is a case of `burying the lede` in journalism terms. This entire PR is mostly in service of making this script work...

bin/crew
```diff
@@ -1506,6 +1524,25 @@ def upload(pkg_name = nil, pkg_version = nil, gitlab_token = nil, binary_compres
         end
       end
     end
+    # Upload python wheels if we are dealing with a pip package, but only
+    # if a gitlab token username is set. (The generic repo does not
+    # require a gitlab token username.)
+    next unless Kernel.system "grep -q \"^require 'buildsystems/pip'\" #{pkg_file}"
+    if gitlab_token_username.nil?
+      puts "\nGITLAB_TOKEN_USERNAME environment variable not set, so generated python wheels will not be uploaded.\n".lightred
+      next package
```
uberhacker, line 1533: Should we abort here instead of next to avoid a stampede of the same messages?

satmandu: No? Because then at least builds will continue?

Generated python wheels can always be uploaded later....

tools/update_python_pip_packages.rb
satmandu: Does `CREW_GITLAB_PKG_REPO` work?

tools/update_python_pip_packages.rb
satmandu: Done.

tools/update_python_pip_packages.rb
satmandu: That's what I did...

bin/crew
uberhacker: What does uploading later really mean?  My thought was if `GITLAB_TOKEN_USERNAME` is not set, it should fail early so the wheels are included?

tools/update_python_pip_packages.rb
```diff
@@ -8,10 +10,21 @@
 py3_files = Dir['py3_*.rb']
 pip_files = `grep -l "^require 'buildsystems/pip'" *.rb`.chomp.split
 relevant_pip_files = (py3_files + pip_files).uniq!
+
+CREW_GITLAB_PKG_REPO = 'https://gitlab.com/api/v4/projects/26210301/packages'
```
uberhacker, line 14: Use `require_relative '../lib/const'` instead?

bin/crew
satmandu: The wheels get saved in the container in `/usr/local/.cache/pip`.

As long as the container isn't closed, the wheels are still there. But at this point yes, it probably makes sense to just set an abort... It's easy enough to make sure everyone generates a GITLAB_TOKEN_USERNAME...

uberhacker (approved):

Issue #9456, opened 2024-03-09
Packages with removed dependencies


Zopolis4:
Came across this while trying to write a script-- I'm sure its indicative of a current lack of thorough testing, and tests could certainly be written to catch this.

Right now, let's just get the issue fixed.

Code I hacked together to get some nice error messages, could be used as the foundation of a test if need be. At some point I do have plans for comprehensive testing of packages themselves (as opposed to the in-progress `crew` unit tests), but there's no harm in increasing our spotty coverage until then.
```ruby
#!/usr/bin/env ruby

# Add >LOCAL< lib to LOAD_PATH so that packages can be loaded
$LOAD_PATH.unshift '../lib'

require_relative '../lib/package'

Dir.glob('../packages/*.rb').each do |filename|
  pkg = Package.load_package(filename)
  begin
    dependencies = pkg.get_deps_list
  rescue Errno::ENOENT => e
    puts "#{pkg.name} depends on removed package #{e.to_s[/\/packages\/(\w+)\.rb/, 1]}"
  end
end
```

Affected packages:
```
dexed depends on removed package dub
obs depends on removed package qtsvg
plplot depends on removed package qtsvg
stellarium depends on removed package qtlocation
```

I ran the code on all architectures, so these should be the only broken packages.

yeesterPlus:
nautilus has a missing dependency called: `64/tracker-3.0/libtracker-parser-libicu.so` in old tracker-3.0

yest5xjquvf2:
somehow you blocked non spam

PR #10464, opened 2024-09-12
Add test for removed dependencies

Zopolis4:
Given that this issue has reared its head twice in the past week, I thought it prudent to add my test from #9456 so these issues are caught earlier.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=mop crew update \
&& yes | crew upgrade
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -19,6 +19,7 @@ end
 
 # Package tests
 task :package_test do
+  ruby 'tests/dep_test'
   ruby 'tests/prop_test'
   ruby 'tests/buildsystem_test'
 end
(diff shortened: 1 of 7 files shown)


Zopolis4:
I have checked, and mutter works without `gnome_settings_daemon`: https://gitlab.gnome.org/GNOME/mutter/-/blob/main/meson.build?ref_type=heads#L124

Zopolis4:
I have checked, and mutter works without `gnome_settings_daemon`: https://gitlab.gnome.org/GNOME/mutter/-/blob/main/meson.build?ref_type=heads#L124

uberhacker (approved):
Glad you made the script work with arguments.

PR #10478, opened 2024-09-14
Require package.rb from buildsystems properly.

satmandu:
- Properties couldn't be read from buildsystems files because `package_helpers` wasn't being loaded... but `package_helpers` is already loaded from `require 'package'`!
- Well, `require 'package'` doesn't work from buildsystems because `package.rb` is one directory down... so let's fix that with `require_relative '../package.rb'`.
- Update packages with the `tools/build_updated_packages.rb` to make sure nothing broke.
##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 

##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=pip_helpers crew update \
&& yes | crew upgrade
```


uberhacker (approved):
Good call!

PR #10481, opened 2024-09-15
Subversion 1.9.12 => 1.14.3

uberhacker:
Tested & Working properly:
- [x] `x86_64`
- [ ] `i686` No longer supported
- [x] `armv7l`
```bash
CREW_REPO=https://github.com/uberhacker/chromebrew.git CREW_BRANCH=update-subversion crew update \
&& yes | crew upgrade
```


uberhacker:
> 86 support-- could you point me to the release notes or otherwise the error you encountered?

The issue is not with Subversion, per say, but rather the dependencies.  In particular, util_linux does not provide a libuuid.la file necessary to build.

satmandu:
There is I believe a macro to generate `.la` files in the graphicsmagick.rb file which anyone is welcome to extract into package_helpers or wherever...

Zopolis4:
How do other distributions deal with this? or the package developers themselves? depending on a file that the dependency doesn't create doesn't make very much sense

satmandu:
> How do other distributions deal with this? or the package developers themselves? depending on a file that the dependency doesn't create doesn't make very much sense

Feel free to do the deep dive on the libtool archive issue. I just gave up at some point and just started generating libtool archives for libraries when packages needed them to build...

It's just not an issue with cmake or meson builds as far as I have seen...

satmandu:
Also, I'm not going to lose sleep over this... nobody is using subversion on i686...
@Zopolis4 if you want to figure out i686 support in the next day, great, if not let's just merge this...

satmandu (approved):

Zopolis4:
i'll take a look at it at some point, ig

Zopolis4:
ok so looking into it-- we already have the macro to generate the libuuid.la file in util_linux, we just don't run it on i686 for some reason.

Zopolis4:
i think the issue is because we use the meson buildsystem instead of autotools for util_linux, and only autotools generates .la files

Zopolis4:
so either we swap to autotools for util_linux, report an issue upstream with meson to try and get them to have an option to generate .la files, or we patch imagemagick and subversion and any other package that requires .la files to not need them

PR #10494, opened 2024-09-19
Add gem binary build functionality and also refactor upload to use regex in lieu of sed

### LINKED ISSUE #10483: 🐞 Some ruby gems have extensions that require compiling, so installs need to be done after a compiler is available
- This is a [problem ](https://github.com/chromebrew/chromebrew/issues/10465)when built-in gems such as `debug` which are installed with `ruby` have an update, as then trying to update all gems (as from the current `ruby.rb` postinstall) requires a compiler to be installed, which we can't do until `buildessential` is installed.
- This is also a problem because we don't want to require users to have `buildessential` installed...

Quick hackish ways to determine if our gem packages have such extensions is to do the following. (The second method works better.)

1. Search for `.so` files in `gem contents <gem>`.
```bash
for i in ruby_*.rb
do
  export gem_pkg=${i%.rb}
  crew install ${gem_pkg} &>/dev/null
  export gem_cleaner_pkg=${gem_pkg//ruby_/}
  export gem_cleaned_pkg=${gem_cleaner_pkg//_/-}
  gem contents ${gem_cleaned_pkg} | grep -q "\.so" && echo "${gem_pkg} has extensions needing compiling."
done
```

Method 1 reports:

```
debug has extensions needing compiling.
```

2. Search for `spec.extensions` in the unpacked gem file.
```ruby
Dir.chdir('/tmp') do
  Dir['/usr/local/lib/crew/packages/ruby_*.rb'].each do |package|
    gem_name = package.sub('ruby_', '').sub('_', '-').gsub('/usr/local/lib/crew/packages/', '').sub('.rb', '')
    version = `sed -n -e 's/^\ \ version //p' #{package}`.chomp.delete("'").delete('"')
    gem_ver = version.split('-', 2).first.to_s
    ruby_ver = version.split('-', 3).last.to_s
    # Need to check if a gem has extensions. If it does, we need
    # buildessential.
    system "gem fetch #{gem_name} --platform=ruby --version=#{gem_ver}", %i[out err] => File::NULL
    system "gem unpack #{gem_name}-*.gem", %i[out err] => File::NULL
    gem_compile_needed = system "grep -q -r spec.extensions  #{gem_name}-#{gem_ver}/*.gemspec", %i[out err] => File::NULL
    puts "#{gem_name} has extensions" if gem_compile_needed
  end
end
```

Method 2 reports:
```
debug has extensions
ruby-libversion has extensions
```

These gem packages should in some way _not have_ `no_compile_needed` marked in those packages.

These gems can have pre-compiled binaries generated and installed by doing:

```ruby
system 'gem install gem-compiler'
system "gem compile --strip --prune #{gem_name}-#{gem_ver}.gem"
gem_arch = case ARCH
                 when 'x86_64'
                   'x86_64-linux'
                 when 'i686'
                   'x86-linux'
                 when 'aarch64', 'armv7l'
                   'armv8l-linux-eabihf'
                 end
gem_file = "#{gem_name}-#{gem_ver}-#{gem_arch}.gem"
system "gem install --local #{gem_file}"
```

Questions:

- Should we package binary gems in this case?
- If so, how best to do that?

satmandu:
Fixes #10483 

- `def upload` was refactored to handle binary gems, and also to use regex and gsub to add and replace the binary_sha256 blocks in package files.
- `crew` was refactored to add functionality to generate, install, and upload binary gems for gems that have extensions which need compiling. (This includes `ruby_ruby_libversion` and `ruby_debug`.)
- `buildsystems/ruby` was modified accordingly to also 
- Adds the `ruby_gem_compiler` package which is needed for building binary gems.
- Moves `ruby_ruby_libversion` from `buildessential` to `core` as it can now be installed without a compiler present.
- Removes the unecessary `require_relative` from buildsystems files.
- Removes `gem pristine` from the ruby postinstall since the root cause of the debug gem messages are from the ruby debug gem being newer than the version bundled with ruby.

##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=gem crew update \
&& yes | crew upgrade
```

### DIFF
--- a/lib/buildsystems/ruby.rb
+++ b/lib/buildsystems/ruby.rb
@@ -2,63 +2,56 @@ require 'active_support/core_ext/object/blank'
 require 'package'
 require_relative '../package_helpers'
 
+def check_gem_binary_build_needed(gem_name = nil, gem_ver = nil)
+  puts "Checking to see if gem compile for #{gem_name} is needed..."
+  @extract_dir = "#{name}.#{Time.now.utc.strftime('%Y%m%d%H%M%S')}.dir"
+  FileUtils.mkdir_p File.join(CREW_BREW_DIR, @extract_dir)
+  Dir.chdir(File.join(CREW_BREW_DIR, @extract_dir)) do
+    # Need to check if the gem has extensions. If it does, we need
+    # either a compiler or a pre-compiled binary gem.
+    system "gem fetch #{gem_name} --platform=ruby --version=#{gem_ver}"
+    system "gem unpack #{gem_name}-#{gem_ver}.gem"
+    return system "grep -q -r spec.extensions  #{gem_name}-#{gem_ver}/*.gemspec", %i[out err] => File::NULL
+  end
+end
+
+def set_vars(passed_name = nil, passed_version = nil)
+  # crewlog "Setting gem variables... name: #{passed_name}, version: #{passed_version}"
+  # This assumes the package class name starts with 'Ruby_' and
+  # version is in the form '(gem version)-ruby-(ruby version)'.
+  # For example, name 'Ruby_awesome' and version '1.0.0-ruby-3.3'.
+  @gem_name = passed_name.sub('ruby_', '').sub('_', '-')
+  @gem_ver = passed_version.split('-').first.to_s
+  @pkg_ruby_ver = passed_version.split('ruby-').last.to_s
+end
+
 class RUBY < Package
-  property :ruby_install_extras
+  property :gem_name, :gem_ver, :pkg_ruby_ver, :ruby_install_extras
 
   depends_on 'ruby'
 
+  def self.preflight
+    set_vars(name, version)
+    # crewlog "@gem_name: #{@gem_name}, @gem_ver: #{@gem_ver}, @pkg_ruby_ver: #{@pkg_ruby_ver}"
+  end
+
   def self.preinstall
-    # This assumes the package class name starts with 'Ruby_' and
-    # version is in the form '(gem version)-ruby-(ruby version)'.
-    # For example, name 'Ruby_awesome' and version '1.0.0-ruby-3.3'.
-    # puts "@pkg.name is #{@pkg.name}"
-    # puts name
     @extract_dir = "#{name}.#{Time.now.utc.strftime('%Y%m%d%H%M%S')}.dir"
-    @gem_name = name.sub('ruby_', '').sub('_', '-')
-    @gem_ver = version.split('-', 2).first.to_s
-    @ruby_ver = version.split('-', 3).last.to_s
-    puts "no_compile_needed?: #{no_compile_needed?}, compile_needed?: #{compile_needed?}"
-    unless no_compile_needed? || compile_needed?
-      puts "Checking to see if gem compile for #{@gem_name} is needed..."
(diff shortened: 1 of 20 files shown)


lib/buildsystems/ruby.rb
```diff
@@ -1,22 +1,58 @@
...
   depends_on 'ruby'
 
-  no_compile_needed
+  def self.preflight
+    set_vars(name, version)
+    crewlog "@gem_name: #{@gem_name}, @gem_ver: #{@gem_ver}"
+  end
+
+  def self.preinstall
+    @gem_binary_build_needed = check_gem_binary_build_needed(@gem_name, @gem_version) unless no_compile_needed? || gem_compile_needed?
```
uberhacker, line 38: `unless no_compile_needed? || !gem_compile_needed?`

lib/const.rb
```diff
@@ -286,6 +286,15 @@
 
 CREW_CMAKE_LIBSUFFIX_OPTIONS ||= "#{CREW_CMAKE_OPTIONS} -DLIB_SUFFIX=#{CREW_LIB_SUFFIX}"
 
+GEM_ARCH ||= case ARCH
+             when 'x86_64'
+               'x86_64-linux'
+             when 'i686'
+               'x86-linux'
+             when 'aarch64', 'armv7l'
+               'armv8l-linux-eabihf'
```
uberhacker, line 295: This is not 32-bit?

packages/buildessential.rb
```diff
@@ -172,13 +172,13 @@ class Buildessential < Package
   depends_on 'ruby_concurrent_ruby'
   # Needed for irb
   depends_on 'ruby_debug'
+  # Needed to build binary gems
+  depends_on 'ruby_gem_compiler'
```
uberhacker, line 176: Not needed here if included in core?

lib/buildsystems/ruby.rb
satmandu: The logic is to not run the check if we already know (as specified in the package file) that we need to (or don't need to) build the binary gem.

uberhacker: So if `gem_compile_needed` is set, no `@gem_binary_build_needed`?

lib/const.rb
uberhacker: I thought the `7` meant 32-bit and the `8` meant 64-bit?  I was under the impression the `l` meant low power.

uberhacker: Were the gem binaries tested in arm containers and confirmed working?

satmandu: Yes. They're working in my arm containers.

Feel free to test on your hardware to confirm.

satmandu: Armv7l and armv8l are both 32-bit.

Arm64 and aarch64 are the 64-bit modes.

lib/buildsystems/ruby.rb
satmandu: Right. The latter is a variable set when we don't know whether the gem needs binaries yet.

uberhacker (approved):

PR #10500, opened 2024-09-19
Fix build without source logic.

satmandu:
- I needed an `||` instead of an `&&`.


uberhacker (approved):
All this boolean logic makes my head spin.

PR #10503, opened 2024-09-20
Py updates, fix env variable for timeout, add option to skip remote updates in build_updates script

satmandu:
- also rebuilds `parted` and updates `nconvert`.
##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=py_updates crew update \
&& yes | crew upgrade
```


tools/build_updated_packages.rb
```diff
@@ -15,7 +15,9 @@
 
 abort "\nGITLAB_TOKEN environment variable not set.\n".lightred if ENV['GITLAB_TOKEN'].nil?
 abort "\nGITLAB_TOKEN_USERNAME environment variable not set.\n".lightred if ENV['GITLAB_TOKEN_USERNAME'].nil?
+puts "Setting the CREW_AGREE_TIMEOUT_SECONDS environment variable to less than the default of #{CREW_AGREE_TIMEOUT_SECONDS} may speed this up...".orange if ENV['CREW_AGREE_TIMEOUT_SECONDS'].nil?
 
+SKIP_UPDATE_CHECKS = ARGV.include?('--skip')
```
uberhacker, line 20: Should this be documented somewhere?

tools/build_updated_packages.rb
```diff
@@ -161,13 +167,12 @@ def self.agree_default_yes(message = nil)
     puts "#{name.capitalize} #{@version} has no binaries and may not need them.".lightgreen
     next pkg
   else
-    build = compatibility == 'all' ? %w[x86_64 armv7l i686] : compatibility.split
-    build.each do |arch|
+    check_builds = compatibility == 'all' ? %w[x86_64 armv7l i686] : compatibility.split
```
uberhacker, line 170: This assumes all the compatibility lines are separated by spaces.
```
cage.rb:  compatibility 'aarch64, armv7l, x86_64'
```

tools/build_updated_packages.rb
satmandu: Done.

uberhacker: I mean like in help.   It sorta already is self-documented here.

satmandu: It's only being used in that one script. It's not an addition to `crew` or to `const.rb`.

tools/build_updated_packages.rb
satmandu: Good call.

Now it handles that case.

tools/build_updated_packages.rb
uberhacker: Would it make sense to add to crew eventually?

tools/build_updated_packages.rb
```diff
@@ -161,13 +170,12 @@ def self.agree_default_yes(message = nil)
     puts "#{name.capitalize} #{@version} has no binaries and may not need them.".lightgreen
     next pkg
   else
-    build = compatibility == 'all' ? %w[x86_64 armv7l i686] : compatibility.split
-    build.each do |arch|
+    check_builds = compatibility == 'all' ? %w[x86_64 armv7l i686] : compatibility.delete(',').split
```
uberhacker, line 173: That probably doesn't work for this?
```
snort.rb:  compatibility 'aarch64,armv7l,x86_64'
```

tools/build_updated_packages.rb
satmandu: Maybe? If we eventually move the script into crew...

But I don't think it is needed at the moment.

tools/build_updated_packages.rb
satmandu: Was easier just to change the packages...

uberhacker (approved):

uberhacker (approved):

PR #10551, opened 2024-10-01
Ffmpeg => 7.1 & Deps Updates

satmandu:
- We need to make sure all the packages that depend upon ffmpeg don't break from this.
Builds:
- [x] `x86_64`
- [x] `armv7l`
##
Tested & Working properly:
- [ ] `x86_64`
- [ ] `armv7l` 
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=ffmpeg crew update \
&& yes | crew upgrade
```


uberhacker:
Wow, ambitious undertaking!

VenomousSteam81:
ffmpeg works on arm64.

satmandu:
Let's just merge this and see what happens. We can update packages as people have issues ... But this way packages don't keep getting built against an older ffmpeg...

packages/avisynthplus.rb
```diff
@@ -1,37 +1,25 @@
...
-        -DBUILD_SHARED_LIBS=ON \
-        .."
-    end
-    system 'ninja -C builddir'
-  end
-
-  def self.install
-    system "DESTDIR=#{CREW_DEST_DIR} ninja -C builddir install"
-  end
+  cmake_options "-DCMAKE_BUILD_TYPE:STRING='None'"
```
uberhacker, line 24: Shouldn't this be `Release`?

satmandu: I copied from here:

https://gitlab.archlinux.org/archlinux/packaging/packages/avisynthplus/-/blob/main/PKGBUILD?ref_type=heads

uberhacker (approved):

PR #10600, opened 2024-10-13
Fix buildsystems/autotools install_extras,  Update mpc, mpfr, gawk, autoconf_archive, sphinx

satmandu:
- Update mpc, mpfr, gawk, autoconf_archive, sphinx
- Also Update reviewdog to add reviews to PR.

Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=fix crew update \
&& yes | crew upgrade
```


bin/crew
```diff
@@ -883,7 +883,7 @@ def prepare_package(destdir)
     abort 'Exiting due to above errors.'.lightred if errors
 
     # Make sure the package file has runtime dependencies added properly.
-    system "#{CREW_LIB_PATH}/tools/getrealdeps.rb --use-crew-dest-dir #{@pkg.name}", exception: true unless @pkg.no_compile_needed?
+    system "#{CREW_LIB_PATH}/tools/getrealdeps.rb --use-crew-dest-dir #{@pkg.name}", exception: true if Kernel.system('which gawk', %i[out err] => File::NULL) && Kernel.system('which upx', %i[out err] => File::NULL) && !@pkg.no_compile_needed?
```
uberhacker, line 886: I'm wondering if `File.file? "#{CREW_PREFIX}/bin/gawk"` would be a more reliable way to determine if the file exists?

satmandu: In theory there could be an existing gawk in the path?

uberhacker: Doesn't the package need to be installed for gawk to be available?

satmandu: Not just installed but available. This appears to break during gawk installs, and the getrealdeps script tries to install gawk if it isn't installed, so it seemed best to just check if a gawk is available?

uberhacker (approved):

PR #10626, opened 2024-10-17
Pip updates, ruby buildsystem tweak, install update to report cache usage, pip updates

satmandu:
- Also adds activestate gem deps right in install.sh so that those gems get reported as installed early.

Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=updates crew update \
&& yes | crew upgrade
```


bin/crew
```diff
@@ -520,7 +520,7 @@ def download
             cachefile = ''
           end
         else
-          puts 'Cannot find cached archive. 😔 Will download.'.lightred
+          puts "Cannot find cached archive at #{cachefile}. 😔 Will download.".lightred
```
uberhacker, line 523: Not really an error but rather a warning?  Use yellow or orange instead of red?

uberhacker (approved):

PR #10719, opened 2024-11-07
Use M130 containers for unit tests, and use unit_test_stub.sh to invoke unit tests.

satmandu:


tests/unit_test.sh
```diff
@@ -1,10 +1,8 @@
-#!/bin/bash -e
-# This is for use as a Github CI Pull Request Unit Test.
+#!/bin/bash
+# This is for use as a Github CI Unit Test.
+# Version 1.0
 set -e
-echo "CHANGED_PACKAGES: ${CHANGED_PACKAGES}"
 cd /usr/local/lib/crew/packages/
-yes | crew update
-yes | crew upgrade
```
uberhacker, line 7: This just means the latest images always need to be up to date, correct?

satmandu: No. The `unit_tests_stub.sh` script (already merged) invokes `crew update` and crew upgrade`, and then runs `unit_tests.sh` so that we have the newest `unit_tests.sh` script run...

PR #10738, opened 2024-11-13
Add taskwarrior package

uberhacker:
Taskwarrior is Free and Open Source Software that manages your TODO list from the command line.
It is flexible, fast, and unobtrusive. It does its job then gets out of your way.  See https://taskwarrior.org/.

Tested & Working properly:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l` 
```bash
CREW_REPO=https://github.com/uberhacker/chromebrew.git CREW_BRANCH=add-taskwarrior-package crew update \
&& yes | crew upgrade
```


packages/taskwarrior.rb
```diff
@@ -0,0 +1,63 @@
...
+  depends_on 'glibc' # R
+  depends_on 'util_linux' # R
+
+  def self.postinstall
+    ExitMessage.add "\nType 'man task' for help to get started.\n"
+  end
+
+  def self.postremove
+    taskrc = "#{HOME}/.taskrc"
+    if File.exist? taskrc
```
satmandu, line 31: We have a postinstall remove macro now!

uberhacker: Does it work for files and directories?

satmandu: I believe so? Do you want to check?

uberhacker: I examined remove.rb and didn't notice any special logic to handle the cleanup.  Also, many packages still have logic similar to this.

satmandu: See this in `package.rb`:
```ruby
  def self.agree_to_remove(config_object = nil)
    if File.file? config_object
      identifier = 'file'
    elsif File.directory? config_object
      identifier = 'directory'
    else
      puts "Cannot identify #{config_object}.".lightred
      return
    end
    if agree_default_no("Would you like to remove the config #{identifier}: #{config_object}")
      FileUtils.rm_rf config_object
      puts "#{config_object} removed.".lightgreen
    else
      puts "#{config_object} saved.".lightgreen
    end
  end
```

In `ruby_pry.rb`:
```ruby
  def self.postremove
    Package.agree_to_remove("#{HOME}/.pryrc")
  end
```

In `stellarium.rb`:
```ruby
  def self.postremove
    Package.agree_to_remove("#{HOME}/.stellarium")
  end
```

Is that not cleaner?

uberhacker: Gotcha, didn't notice that.

satmandu (approved):

PR #10747, opened 2024-11-15
Add versioned linuxheaders dependency to armv7l ruby installs.

satmandu:
- Tweak to force dependence on the older linuxheaders for a container build.

### DIFF
--- a/packages/ruby.rb
+++ b/packages/ruby.rb
@@ -24,6 +24,7 @@ class Ruby < Package
   depends_on 'gmp' # R
   depends_on 'libffi' # R
   depends_on 'libyaml' # R
+  depends_on 'linuxheaders', '== 3.10' => :build if %w[armv7l aarch64].include?(ARCH) && CREW_IN_CONTAINER
   depends_on 'openssl' # R
   depends_on 'rust' => :build
   depends_on 'zlib' # R


uberhacker (approved):

satmandu:
@supechicken It appears that the code in `lib/package.rb` can't handle a versioned build dependency such as 
```ruby
depends_on 'linuxheaders', '== 3.10' => :build
```

That throws this error:
```
/usr/local/lib/crew/lib/package.rb:324:in `depends_on': undefined method `split' for an instance of Hash (NoMethodError)

      operator, target_ver = ver_range.split(' ', 2)
                                      ^^^^^^
	from /usr/local/lib/crew/packages/ruby.rb:27:in `<class:Ruby>'
	from /usr/local/lib/crew/packages/ruby.rb:3:in `load_package'
	from /usr/local/lib/crew/lib/package.rb:118:in `class_eval'
	from /usr/local/lib/crew/lib/package.rb:118:in `load_package'
	from /usr/local/bin/crew:374:in `block in upgrade'
	from /usr/local/bin/crew:395:in `block in upgrade'
	from /usr/local/bin/crew:393:in `each'
	from /usr/local/bin/crew:393:in `upgrade'
	from /usr/local/bin/crew:1898:in `upgrade_command'
	from /usr/local/bin/crew:1934:in `<main>'
```

This is the relevant code in `lib/package.rb`:

```ruby
  def self.depends_on(dependency, ver_range = nil)
    @dependencies ||= {}
    ver_check = nil
    dep_tags  = []

    # Add element in "[ name, [ tag1, tag2, ... ] ]" format.
    if dependency.is_a?(Hash)
      # Parse "depends_on name => <tags: Symbol|Array>".
      dep_name, tags = dependency.first

      # Convert `tags` to array in case `tags` is a symbol.
      dep_tags += [tags].flatten
    else
      # Parse "depends_on name".
      dep_name = dependency
    end

    # Process dependency version range if specified.
    # example:
    #   depends_on name, '>= 1.0'
    #
    # Operators can be: '>=', '==', '<=', '<', or '>'
    if ver_range
      operator, target_ver = ver_range.split(' ', 2)
```

Any thoughts on how best to fix this? (Asking because this is your code.)

Zopolis4:
When I last tested versioned dependencies they worked for me, perhaps its because glibc is a fake package? Also, we don't have a glibc 2.26 version available, not that i can see.

satmandu:
> When I last tested versioned dependencies they worked for me, perhaps its because glibc is a fake package? Also, we don't have a glibc 2.26 version available, not that i can see.

Yes that was the point. I was trying to get the build to fail.

PR #10751, opened 2024-11-17
Use latest ubuntu LTS for CI workflows, and make sure gems are update…

satmandu:
- Try to fix the current breakage in the ruby workflow as well...


.github/workflows/Linter-Handoff.yml
```diff
@@ -6,7 +6,7 @@ on:
   merge_group:
 jobs:
   handoff:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
Zopolis4, line 9: Do we need to make this change immediately rather than just waiting for https://github.com/actions/runner-images/issues/10636?

satmandu: We're already using ubuntu-24.04 based images for our armv7l runner, since that requires .NET 9, which isn't supported on older Ubuntu releases.

So this should actually sync up all of our actions to run on similar containers.

Zopolis4: Fair enough.

uberhacker (dismissed):

uberhacker (approved):

PR #10769, opened 2024-11-21
Update glib, gtk[3,4], libadwaita, and fix gtk3 build dependency loops.

satmandu:
- Also removes `broadway` from gtk[3,4] since it is likely not used and also not wired up for use.

##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=gtk crew update \
&& yes | crew upgrade
```


packages/rest.rb
```diff
@@ -16,26 +16,28 @@ class Rest < Meson
      x86_64: 'dcb417551add89f25c24a9f3cdcd3f43c4fa0ac78606775fe252dd6662d3056c'
```
uberhacker, line 16: No new binaries?

satmandu: No. Just updated dependency logic.

PR #10908, opened 2024-12-07
Only have tools/version.rb report stable version updates when using 'crew check -V <package>'.

satmandu:
```bash
for pkg in $(grep -l no_compile_needed *) ; do echo ${pkg%.rb} ; crew check -V ${pkg%.rb} | grep outdated ; done
abcl
acquia_cli
act
act                                outdated            0.2.34              0.2.70
angle_grinder
apktool
atom
audacity
autojump
aws2
balena_etcher
bash_snippets
bat
bazel
```
...


uberhacker (approved):
Nice!

PR #11064, opened 2024-12-31
add libsamplerate as an dependency for QEMU

buhron:
I added libsamplerate as an dependency for QEMU.


satmandu:
Was the binary broken without this installed or is a rebuild needed?

buhron:
Qemu was broken without libsamplerate, without it, it would say something like "qemu-system-x86_64: error while loading shared libraries: libsamplerate.so.2: No such file or directory"

uberhacker (approved):
Thank you for the update @buhron!

PR #11112, opened 2025-01-09
Remove duplicate `zlib` install and use proper parsing for /etc/lsb-release

s0urce-c0de:
Fixes N/A

## Description

This PR removes a duplicate `zlib` in install  proper `/etc/lsb-release` parsing (source: [Chromium OS Developer Library /etc/lsb-release File Format](https://www.chromium.org/chromium-os/developer-library/reference/infrastructure/lsb-release/#:~:text=If%20you're%20writing,code%20in%20a%20better%20language.)).

##
Builds:
- [ ] `x86_64`
- [ ] `i686`
- [ ] `armv7l`
##
Tested & Working properly:
- [ ] `x86_64`
- [ ] `i686`
- [ ] `armv7l` 
##
##
```bash
rm -rf /usr/local/* && bash <(curl -Ls github.com/s0urce-c0de/chromebrew/raw/no_duplicate_zlib_proper_lsb_parsing/install.sh)
```


install.sh
```diff
@@ -31,12 +49,12 @@ fi
 
 # Reject non-stable Chrome OS channels.
 if [ -f /etc/lsb-release ]; then
-  if [[ ! "$(< /etc/lsb-release)" =~ CHROMEOS_RELEASE_TRACK=stable-channel$'\n' && "${CREW_FORCE_INSTALL}" != '1' ]]; then
+  if [[ ! "$(lsbval CHROMEOS_RELEASE_TRACK)" =~ stable-channel && "${CREW_FORCE_INSTALL}" != '1' ]]; then
```
uberhacker, line 52: Where is `CHROMEOS_RELEASE_TRACK` set initially?  If not set, `lsbval` will always return 1.

uberhacker: What we're trying to accomplish here is to find out if the stable channel is active for ChromeOS.  What is the output of `cat /etc/lsb-release` on your system?

satmandu: Stable channel will still be marked as true if updates are rejected, right?

satmandu:
In any case, this doesn't appear to entirely solve the problem on the M124 container. (But the zlib changes may help...) There's a race condition where all the libraries needed to extract both xz and zst formats aren't installed at the same time when the system zstd binary exists, but our zstd isn't installed yet.

satmandu:
(I had a solution briefly... I think it involves selectively choosing which LD_LIBRARY_PATH to use at the right time...)

Will take a look again when I get a chance, but that's my checkpoint for trying to get this to work...

install.sh
s0urce-c0de: > Stable channel will still be marked as true if updates are rejected, right?

`/etc/lsb-release` should stay read only by default so it will be on the version that the current booted OS is on.

s0urce-c0de: > Where is `CHROMEOS_RELEASE_TRACK` set initially? If not set, `lsbval` will always return 1.

CHROMEOS_RELEASE_TRACK should be set in /etc/lsb-release unless the user removed it.

uberhacker: The conditional logic is trying to find out what the value of `CHROMEOS_RELEASE_TRACK` is set to in ``/etc/lsb-release`.  I'm still trying to figure out what needs to be done here.

s0urce-c0de: > The conditional logic is trying to find out what the value of `CHROMEOS_RELEASE_TRACK` is set to in `/etc/lsb-release`. I'm still trying to figure out what needs to be done here.

CHROMEOS_RELEASE_TRACK makes sure that the user is on stable channel because beta/dev/canary are unsupported.

install.sh
```diff
@@ -12,6 +12,24 @@ echo_success() { echo -e "\e[1;32m${*}${RESET}" >&1; } # Use Green for success m
...
+# Reference: https://www.chromium.org/chromium-os/developer-library/reference/infrastructure/lsb-release/
+lsbval() {
+  local key="$1"
+  local lsbfile="${2:-/etc/lsb-release}"
+
+  if ! echo "${key}" | grep -Eq '^[a-zA-Z0-9_]+$'; then
+    return 1
+  fi
+
+  sed -E -n -e \
```
supechicken, line 25: For me, I prefer parsing all variables into the shell first instead of scanning and extracting the value from the file every time

Maybe we could make use of the `cut` command here? Like this:
```shell
cat <<EOF | cut -d'=' -f2
DOUBLE_QUOTES = "double"
SINGLE_QUOTES = 'sin gle'
EOF

# "double"
# 'sin gle'
```

supechicken: Never mind, didn't realize this cannot handle trailing spaces😅

satmandu:
Ok I got this to work for M124.

Here's the working container I built from modifying the install script in this PR as per https://github.com/chromebrew/chromebrew/issues/10981#issuecomment-2583112790 :
```bash
#!/bin/bash
# crewbuild-octopus-x86_64.m124.sh
if [ -n "$SSH_CLIENT" ] || [ -n "$SSH_TTY" ]; then
  SESSION_TYPE=remote/ssh
elif pstree -p | egrep --quiet --extended-regexp ".*sshd.*\($$\)"; then
  SESSION_TYPE=remote/ssh
else
  case $(ps -o comm= -p $PPID) in
    sshd|*/sshd) SESSION_TYPE=remote/ssh;;
  esac
fi
if [ -z ${PAGER+x} ]; then 
  echo "PAGER is not set."
else 
  PAGER_PASSTHROUGH=-e
  PAGER_PASSTHROUGH+=" "
  PAGER_PASSTHROUGH+=CONTAINER_PAGER=${PAGER}
fi
if [ -z ${TZ+x} ]; then 
  echo "TZ is not set."
else 
  TZ_PASSTHROUGH=-e
  TZ_PASSTHROUGH+=" "
  TZ_PASSTHROUGH+=TZ=${TZ}
fi
X11=-e
X11+=" "
X11+=DISPLAY=${DISPLAY:-:0.0}
X11+=" "
if ! [[ $SESSION_TYPE == remote/ssh ]] && [ -d /tmp/.X11-unix ]; then
  X11+=" -v /tmp/.X11-unix:/tmp/.X11-unix "
fi
if [ -f "$HOME"/.Xauthority ]; then
  X11+=--volume=$HOME/.Xauthority:/home/chronos/user/.Xauthority:rw
  X11+=" "
  X11+=--volume=$HOME/.Xauthority:/home/chronos/.Xauthority:rw
fi
docker pull --platform linux/amd64 satmandu/crewbuild:octopus-x86_64.m124
docker pull tonistiigi/binfmt
docker run --privileged --rm tonistiigi/binfmt --install all
docker run --init --platform linux/amd64 --rm --net=host ${PAGER_PASSTHROUGH} ${TZ_PASSTHROUGH} ${X11} -e LOCALRC="${LOCALRC}" -v $(pwd)/pkg_cache:/usr/local/tmp/packages -v $(pwd):/output -h $(hostname)-x86_64 -it satmandu/crewbuild:octopus-x86_64.m124 $@
```

satmandu:
(Feel free to pull those changes into this PR...)

satmandu:
I added changes to build M124, M130, and M131 container images successfully.

satmandu:
This does break i686 container image installs, so I need to debug that...

satmandu:
Looks like i686 is working now...

uberhacker (approved):

PR #11116, opened 2025-01-10
Manually set HOME for container usage.

satmandu:
- In some containers, `Dir.home`, which is used to set `HOME`, gets set to `~root/` instead of `/home/chronos/user`, so set it manually if usage is in a container.


uberhacker (approved):
Good call.  Is there any way we can run processes in the container exclusively under the chronos user?

satmandu:
> Good call.  Is there any way we can run processes in the container exclusively under the chronos user?

`id` shows `chronos`, but for some reason `$HOME` isn't getting set properly. It's probably a sudo or bash invocation setting we are missing...

PR #11143, opened 2025-01-15
Firefox 134.0 => 134.0.1

uberhacker:
Tested & Working properly:
- [x] `x86_64`
##
##
```bash
CREW_REPO=https://github.com/uberhacker/chromebrew.git CREW_BRANCH=update-firefox crew update \
&& yes | crew upgrade
```


packages/firefox.rb
```diff
@@ -100,4 +100,18 @@ def self.postinstall
     end
     ExitMessage.add "\nType 'firefox' to get started.\n"
   end
+
+  def self.postremove
+    Dir.chdir("#{CREW_PREFIX}/bin") do
+      if File.exist?('x-www-browser') && File.symlink?('x-www-browser') \
```
supechicken, line 106: `File.exist?` can be removed here as `File.symlink?` also handles it

supechicken (dismissed):

packages/firefox.rb
```diff
@@ -100,4 +100,18 @@ def self.postinstall
     end
     ExitMessage.add "\nType 'firefox' to get started.\n"
   end
+
+  def self.postremove
+    Dir.chdir("#{CREW_PREFIX}/bin") do
+      if File.symlink?('x-www-browser') \
+        && (File.realpath('x-www-browser') == "#{CREW_PREFIX}/bin/firefox")
```
supechicken, line 107: My fault 😅 Doesn't know `File.realpath` will fail on invalid symlink. Use `File.readlink` instead?

uberhacker: That didn't work.  I think the main issue here is we should be using `self.preremove` instead of `self.postremove` because in the postremove stage, the firefox executable has already been removed.

satmandu (approved):

PR #10376, opened 2024-08-30
Update docbook-xsl-nons to 2020-06-03.

satmandu:
##
Builds:
- [x] `x86_64`
- [x] `i686`
- [x] `armv7l`
##

##
### Run the following to get this pull request's changes locally for testing.
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=docbook crew update \
&& yes | crew upgrade
```

PR #11266, opened 2025-02-08
xerxesc 3.2.3 -> 3.2.5, zeromq 4.3.2 -> 4.3.5

Zopolis4:
## Description
Had these lying around on an old branch-- I think I was waiting to update more packages so I could do a bulk PR, but I've forgotten what those packages would have been, so I'm just sending these as they are.

No filelist updates for `xerxesc`.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=spanner1 crew update \
&& yes | crew upgrade
```


packages/zeromq.rb
```diff
@@ -1,37 +1,28 @@
...
-  def self.check
-    system 'make', 'check'
-  end
-
-  def self.install
-    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
-  end
+  # warning: failed to load external entity "http://docbook.sourceforge.net/release/xsl/current/manpages/docbook.xsl"
+  # compilation error: file /tmp/xmlto-xsl.bNfymk line 4 element import
+  configure_options '--without-docs'
```
uberhacker, line 26: We should try to restore the man pages, if possible.  Rework docbook packages?

Zopolis4: This has been an issue for as long as I can remember-- I'm leaving this comment here so that I can find it when we fix it, but that's a bigger issue for later.

Zopolis4: See #10376.

satmandu: I don't have time to rework the doc book packages. Consider opening an issue for that and asking for help for it...

satmandu (approved):

PR #11283, opened 2025-02-10
Split out definition of require_gem to standalone file

Zopolis4:
## Description
Not sure why we have this defined identically in so many places, but now we don't.

I tried using ConvenienceFunctions, but that led to circular requires and other issues, so its a standalone file. I am trying to avoid increasing the number of files in `lib`, but there are a couple suited for merging so this hopefully won't be too big of an issue.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=lowfru crew update \
&& yes | crew upgrade
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1,26 +1,9 @@
 #!/usr/bin/env ruby
+# We require color here to get access to .lightred.
 require_relative '../lib/color'
-# Disallow sudo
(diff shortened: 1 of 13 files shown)


satmandu (approved):
Looks like nothing breaks!

PR #11305, opened 2025-02-12
WIP break apart builds to avoid race conditions.

satmandu:


Zopolis4:
What race condition is this resolving? would be good to know what the specific problem being solved is before I start to review

satmandu:
> What race condition is this resolving? would be good to know what the specific problem being solved is before I start to review 

Right now we can't fire up builds on multiple architectures in parallel, because each build could independently update a package file, and then you have conflicting commits.

I suppose another approach would be to set a lock on the package file using git lfs...

bin/crew
```diff
@@ -1535,16 +1509,85 @@ def upload(pkg_name = nil, pkg_version = nil, gitlab_token = nil, gitlab_token_u
...
+        end
+      end
+    end
+    puts "\e[1A\e[K🎉 Uploads complete for #{package}. 🎉\r\n".lightgreen
+    puts "\e[1A\e[K🎉 Updating package file for #{package}. 🎉\r\n".lightgreen
+    update_package_binary_hashes(pkg_name, pkg_version, binary_compression)
+  end
+end
+
+def update_package_binary_hashes(pkg_name = nil, pkg_version = nil, binary_compression = nil)
```
Zopolis4, line 1547: This feels much better suited as a script in tools, given that it's quite developer specific, and beyond that pretty CI specific.

bin/crew
```diff
@@ -1535,16 +1509,85 @@ def upload(pkg_name = nil, pkg_version = nil, gitlab_token = nil, gitlab_token_u
...
+
+  [packages].each do |package|
+    pkg_file = "#{CREW_LOCAL_REPO_ROOT}/packages/#{package}.rb"
+    binary_compression_not_in_file = binary_compression.nil?
+    if binary_compression_not_in_file
+      binary_compression = @pkg.superclass.to_s == 'RUBY' ? 'gem' : 'tar.zst'
+      binary_compression_line = "  binary_compression '#{binary_compression}'"
+      puts "Setting binary compression in #{pkg_file} to '#{binary_compression}'..."
+      # Add binary compression setting, and add the line if it doesn't exist.
+      file = File.read(pkg_file)
```
Zopolis4, line 1559: Load the package file instead of parsing it?

tools/github_actions_source_hash_updater.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
Zopolis4, line 1: If we're going to be splitting everything out into smaller scripts, would it make sense to have a tools/actions directory? Right now the tools directory is getting pretty cluttered with scripts that are really only used in CI.

tools/update_package_source_hashes.rb
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env ruby
```
Zopolis4, line 1: There's a lot of duplicate code between this and its binary equivalent-- would it be possible to combine them into one script?

Zopolis4 (commented):
Just some preliminary comments, mainly about general structural and architectural choices that it'd be best to resolve early.

Zopolis4:
> > What race condition is this resolving? would be good to know what the specific problem being solved is before I start to review 
> 
> Right now we can't fire up builds on multiple architectures in parallel, because each build could independently update a package file, and then you have conflicting commits.
> 
> I suppose another approach would be to set a lock on the package file using git lfs...

Ah. Perhaps we could also store the new hash in memory and rebase after a build is finished? i.e. the x86_64 build completes first and so the armv7l build pulls the new x86_64 hashes in before making its own edits.

satmandu:
> > > What race condition is this resolving? would be good to know what the specific problem being solved is before I start to review
> > 
> > 
> > Right now we can't fire up builds on multiple architectures in parallel, because each build could independently update a package file, and then you have conflicting commits.
> > I suppose another approach would be to set a lock on the package file using git lfs...
> 
> Ah. Perhaps we could also store the new hash in memory and rebase after a build is finished? i.e. the x86_64 build completes first and so the armv7l build pulls the new x86_64 hashes in before making its own edits.

The only shared memory between the containers is the GitHub environment though.

And we have to assume that the build could start partially finished for any particular architecture, so hashes may not already be generated.

It gets complicated quickly, especially if the hashes need to be generated for existing binaries from for instance Glibc variable specific per architecture.

Zopolis4:
Hence why every build would pull in the latest changes before they write their new hashes-- then they can work from a clean git base without any risk of conflict.

PR #11314, opened 2025-02-13
Add Extreme Tux Racer

FinnBaltazar1111:
By far one of the best Linux games to ever exist.

## Description

Adds `extremetuxracer`

## Additional information

As you can see, it's just a `.deb` package that I built. But it works, on the architectures that I've tested it on!
By the way, the `.deb` packages all include `libsfml_system`, `libsfml_audio`, `libsfml_window`, `libsfml_graphics`, and `libFLAC` because I didn't wanna take the time to port each of those individual packages to Chromebrew (and I was convinced it probably wouldn't have worked if I did). This may or may not be an issue, but I'm, of course, willing to work with everyone here to fix any issues!

##
Builds:
- [x] `x86_64`
- [ ] `i686`
- [x] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [ ] `armv7l` 
```bash
CREW_REPO=https://github.com/FinnBaltazar1111/chromebrew.git CREW_BRANCH=extremetuxracer crew update \
&& yes | crew upgrade
```


packages/extremetuxracer.rb
```diff
@@ -0,0 +1,26 @@
+require 'package'
+
+class Extremetuxracer < Package
+  description 'High speed arctic racing game based on Tux Racer'
+  homepage 'https://sourceforge.net/projects/extremetuxracer/'
+  version "0.8.4-1"
+  compatibility 'x86_64 armv7l'
+  license 'GPLv2'
+  source_url "https://github.com/FinnBaltazar1111/extremetuxracer/releases/latest/download/extremetuxracer_#{version}_#{ARCH}.deb"
+  source_sha256 'SKIP'
```
FinnBaltazar1111, line 10: I've skipped the source checksum because the way I've linked the different `.deb` packages for each different architectures depends on `source_url`'s value itself (which will, again, be different on all architectures). Please tell me if there's any way to insert a verifiable checksum!

uberhacker: Typically, that would be handled like below:
```
  source_url({
    aarch64: "https://github.com/FinnBaltazar1111/extremetuxracer/releases/latest/download/extremetuxracer_#{version}_armv7l.deb",
     armv7l: "https://github.com/FinnBaltazar1111/extremetuxracer/releases/latest/download/extremetuxracer_#{version}_armv7l.deb",
       i686: "https://github.com/FinnBaltazar1111/extremetuxracer/releases/latest/download/extremetuxracer_#{version}_i686.deb",
     x86_64: "https://github.com/FinnBaltazar1111/extremetuxracer/releases/latest/download/extremetuxracer_#{version}_x86_64.deb"
  })  
  source_sha256({
    aarch64: '4b0550239c2cd884cc03ddeb2b9934708f4b073ad59a96fccbfe09f7e4f54233',
     armv7l: '4b0550239c2cd884cc03ddeb2b9934708f4b073ad59a96fccbfe09f7e4f54233',
       i686: '876159cc1e15efb571e61843b39a2327f8925951d48b9a7a03048c36f72180f7',
     x86_64: 'd45312e61ebcc48032b77bc4cf7fd6915c11fa16e4aad116b66c9468211230ea'
  })
```
Note that aarch64 needs armv7l architecture due to the ChromeOS 32-bit runtime.

FinnBaltazar1111: Alright, I fixed it! I feel pretty confident with the `amd64` and `armhf` binaries, but I might need someone with an `arm64`/`aarch64` Chromebook to test that binary.

packages/extremetuxracer.rb
```diff
@@ -0,0 +1,34 @@
...
+class Extremetuxracer < Package
+  description 'High speed arctic racing game based on Tux Racer'
+  homepage 'https://sourceforge.net/projects/extremetuxracer/'
+  version "0.8.4-1"
+  compatibility 'x86_64 armv7l aarch64'
+  license 'GPLv2'
+  source_url ({
+    "armv7l": "https://github.com/FinnBaltazar1111/extremetuxracer/releases/download/0.8.4/extremetuxracer_#{version}_armhf.deb",
+    "x86_64": "https://github.com/FinnBaltazar1111/extremetuxracer/releases/download/0.8.4/extremetuxracer_#{version}_amd64.deb",
+    "aarch64": "https://github.com/FinnBaltazar1111/extremetuxracer/releases/download/0.8.4/extremetuxracer_#{version}_arm64.deb"
```
uberhacker, line 12: This line needs to be armhf (32-bit) even though the architecture is aarch64.  Also, we'll need manifest files.  These are the files listed in `manifest/<arch>/e/extremetuxracer.filelist` on your local repo.  In your case, there should be 2 (1 for armv7l and 1 for x86_64).  We'll also need an entry in `tools/packages.yaml`.  Take a look at some of the closed PRs to get an idea of what is needed.

FinnBaltazar1111: > This line needs to be armhf (32-bit) even though the architecture is aarch64. Also, we'll need manifest files. These are the files listed in `manifest/<arch>/e/extremetuxracer.filelist` on your local repo. In your case, there should be 2 (1 for armv7l and 1 for x86_64). We'll also need an entry in `tools/packages.yaml`. Take a look at some of the closed PRs to get an idea of what is needed.

I'm a bit confused. Do I need to change the architecture tag from `aarch64` to `arm32hf`, or do I need to completely recompile the binary? And do I need to put the file list inside the package script somehow, or should I just upload it here? Sorry, I couldn't find a package update or package added PR that mentioned the file list.

uberhacker: Sorry, I probably should have given you an example instead of sending you searching.  Here is an example: https://github.com/chromebrew/chromebrew/pull/11237/files.  Any time you install a package, if the filelists are not present, it will generate them.  We need to include these so we can search for files in packages regardless if they are installed or not with `crew whatprovides /usr/local/bin/rsync`, for example.  Also notice the entry in `tools/packages.yaml` in that PR.

packages/extremetuxracer.rb
```diff
@@ -0,0 +1,34 @@
...
+    "armv7l": "47e14334f142d6adc2094d5b9304b0c9129c2fa94e25765ad2c53523ac635202",
+    "x86_64": "d081c1b4fcc8f4586f734c9c9f4c23a5f00909eda2596729d141328a637b99ea",
+    "aarch64": "47e14334f142d6adc2094d5b9304b0c9129c2fa94e25765ad2c53523ac635202"
+  })
+  
+  depends_on 'libsdl2'
+  depends_on 'mesa'
+  depends_on 'mesa_utils'
+  depends_on 'tcl'
+
```
uberhacker, line 24: Since this package doesn't require a build, add the following line here:
```
  no_compile_needed
```

packages/extremetuxracer.rb
```diff
@@ -0,0 +1,34 @@
...
+class Extremetuxracer < Package
+  description 'High speed arctic racing game based on Tux Racer'
+  homepage 'https://sourceforge.net/projects/extremetuxracer/'
+  version "0.8.4-1"
+  compatibility 'x86_64 armv7l aarch64'
+  license 'GPLv2'
+  source_url ({
+    "armv7l": "https://github.com/FinnBaltazar1111/extremetuxracer/releases/download/0.8.4/extremetuxracer_#{version}_armhf.deb",
+    "x86_64": "https://github.com/FinnBaltazar1111/extremetuxracer/releases/download/0.8.4/extremetuxracer_#{version}_amd64.deb",
+    "aarch64": "https://github.com/FinnBaltazar1111/extremetuxracer/releases/download/0.8.4/extremetuxracer_#{version}_armhf.deb"
```
uberhacker, line 12: This looks correct, although we sort the architectures in all our other packages.

packages/extremetuxracer.rb
FinnBaltazar1111: I just add `no_compile_needed`,  or do I delete all of the `self.install` functions too?

uberhacker: No, just add that line.

packages/extremetuxracer.rb
FinnBaltazar1111: Thanks for the context.

FinnBaltazar1111: What should I put for `activity` in `tools/packages.yaml`?

uberhacker (approved):
Looks good!  Thank you for the new package.

FinnBaltazar1111:
> Looks good! Thank you for the new package.

***AWESOME!!!*** I'm cooking on the PRs. Thanks for all the help!

FinnBaltazar1111:
...are the failing checks a problem?

uberhacker:
> ...are the failing checks a problem @uberhacker?

Yes, it looks like there is a linter issue.

FinnBaltazar1111:
> > ...are the failing checks a problem @uberhacker?
> 
> Yes, it looks like there is a linter issue.
> 
> ```
> packages/extremetuxracer.rb:7:17: C: Chromebrew/OrderedCompatibility: Compatibility values should be in alphabetical order.
> ```

so i guess I should just put them in alphabetical order...? i'll do that now

FinnBaltazar1111:
there we go
it feels unnecessary to force the architecture names to be in alphabetical order but whatever

uberhacker:
> there we go it feels unnecessary to force the architecture names to be in alphabetical order but whatever

I somewhat agree but you can duke that out with @Zopolis4.

uberhacker (approved):

PR #11365, opened 2025-02-19
Refactor tests/commands/remove.rb

Zopolis4:
## Description
Some general refactoring that I'm splitting out here rather than adding to #9707.

Changes:
- The puts statements for each test are now included in the test body itself, rather than all being run at the start and thus out of order with the actual running of the test itself.
- Use `Command.remove` instead of a system call, and use `capture_io` to ensure it remains silent.
- Eliminate unnecessary variables and move the remaining ones into their respective tests.
- Simplify assertions and use the correct methods for the correct purposes.

Note that you aren't really supposed to print at all in a minitest file, so there are still some minor visual issues when running the test:
image

Personally, I'd rather remove the printing entirely, but that's a more opinionated change that I didn't want to hold up this refactoring with.
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=unearth crew update \
&& yes | crew upgrade
```

### DIFF
--- a/tests/commands/remove.rb
+++ b/tests/commands/remove.rb
@@ -1,7 +1,6 @@
 require 'minitest/autorun'
 require_relative '../../commands/remove'
 require_relative '../../lib/const'
-require_relative '../../lib/package_utils'
 
 # Add lib to LOAD_PATH
 $LOAD_PATH << File.join(CREW_LIB_PATH, 'lib')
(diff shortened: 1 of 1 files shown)


tests/commands/remove.rb
```diff
@@ -10,68 +9,32 @@
...
-
   def test_remove_essential_package
-    # expected_output = %(  #{@random_essential_package_name.capitalize} is considered an essential package needed for
-    # Chromebrew to function and thus cannot be removed.
-    # )
-    # assert_output(expected_output, nil) do
-    # Command.remove(@random_essential_pkg, true)
-    # end
+    random_essential_package = CREW_ESSENTIAL_PACKAGES.sample
+    puts "Testing the removal of essential package #{random_essential_package}, which was picked at random from one the essential packages: #{CREW_ESSENTIAL_PACKAGES.join(', ')}. This should fail."
```
uberhacker, line 14: from one `of` the?

tests/commands/remove.rb
```diff
@@ -10,68 +9,32 @@
...
-        system "yes | crew install -d #{@normal_package_name}", %i[out err] => File::NULL
-        sleep 2
-      end
-      Command.remove(@normal_pkg, true)
+    puts 'Testing the removal of normal package xxd_standalone. This should succeed.'
+
+    expected_output = "xxd_standalone removed!\n"
+    assert_output expected_output, nil do
+      system 'crew install xxd_standalone', %i[out err] => File::NULL
+      Command.remove(Package.load_package('xxd_standalone.rb'), false)
```
uberhacker, line 37: Should we also check the `-f` switch?  This should override protection and remove an essential package.

Zopolis4: As in, add another test? I'm happy to do that, but I'd rather address it in a separate PR.

uberhacker: Fair enough.

uberhacker (approved):

PR #11368, opened 2025-02-19
Add tests for force and verbose arguments to Command.remove

Zopolis4:
## Description
As discussed in https://github.com/chromebrew/chromebrew/pull/11365#discussion_r1961111269. 

I also added a test for verbose behavior.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=lightlegs crew update \
&& yes | crew upgrade
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -1297,7 +1297,7 @@ def install
   # remove it just before the file copy
   if @pkg.in_upgrade
     puts 'Attempting removal since this is an upgrade or reinstall...'
-    Command.remove(@pkg)
+    Command.remove(@pkg, verbose: CREW_VERBOSE, force: @opt_force)
   end
 
   if @pkg.is_fake?
@@ -1844,7 +1844,7 @@ end
 def remove_command(args)
   args['<name>'].each do |name|
     search name
-    Command.remove(@pkg)
+    Command.remove(@pkg, verbose: CREW_VERBOSE, force: @opt_force)
   end
 end
 
--- a/commands/remove.rb
+++ b/commands/remove.rb
@@ -6,7 +6,7 @@ require_relative '../lib/package'
 require_relative '../lib/package_utils'
 
 class Command
-  def self.remove(pkg, verbose = CREW_VERBOSE)
+  def self.remove(pkg, verbose: false, force: false)
     device_json = JSON.load_file(File.join(CREW_CONFIG_PATH, 'device.json'))
 
     # Make sure the package is actually installed before we attempt to remove it.
@@ -16,7 +16,7 @@ class Command
     end
 
     # Do not remove any packages in CREW_ESSENTIAL_PACKAGES, as those are needed for ruby and thus crew to run.
-    if CREW_ESSENTIAL_PACKAGES.include?(pkg.name) && !CREW_FORCE
+    if CREW_ESSENTIAL_PACKAGES.include?(pkg.name) && !force
       return if pkg.in_upgrade
 
       puts <<~ESSENTIAL_PACKAGE_WARNING_EOF.gsub(/^(?=\w)/, '  ').lightred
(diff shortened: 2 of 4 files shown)


commands/remove.rb
```diff
@@ -16,7 +16,7 @@ def self.remove(pkg, verbose = CREW_VERBOSE)
     end
 
     # Do not remove any packages in CREW_ESSENTIAL_PACKAGES, as those are needed for ruby and thus crew to run.
-    if CREW_ESSENTIAL_PACKAGES.include?(pkg.name) && !CREW_FORCE
+    if CREW_ESSENTIAL_PACKAGES.include?(pkg.name) && !force
```
uberhacker, line 19: `&& (!force || !CREW_FORCE)`?  Otherwise, the environment variable serves no purpose.

commands/remove.rb
```diff
@@ -60,7 +60,7 @@ def self.remove(pkg, verbose = CREW_VERBOSE)
 
           # When searching for files to delete we exclude the files from CREW_ESSENTIAL_PACKAGES.
           essential_packages_exclude_froms = ''
-          unless CREW_FORCE
+          unless force
```
uberhacker, line 63: `unless force || CREW_FORCE`?

commands/remove.rb
Zopolis4: CREW_FORCE isn't set by an environment variable-- it's set by checking for the `-f` argument, and right now its only use is in `lib/downloader.rb`.

Zopolis4: i.e. these changes do not affect user-facing behavior at all.

uberhacker: I should have said constant instead of environment variable, actually.  Wouldn't it make sense to pass parameters like below?
```
def self.remove(pkg, verbose = CREW_VERBOSE, force = CREW_FORCE)
```

uberhacker: You could still do that with:
```
 def self.remove(pkg, verbose: CREW_VERBOSE, force: CREW_FORCE)
```

Zopolis4: True, but I'd rather we pass the information through parameters than have a default value set by an unpredictable constant in an external file. Ideally, once we've split every command out into a file, we won't have any need for these constants, because we'll pass the parameters through in calls to every command.

uberhacker: Any thoughts on this @satmandu?

bin/crew
```diff
@@ -1844,7 +1844,7 @@ end
 def remove_command(args)
   args['<name>'].each do |name|
     search name
-    Command.remove(@pkg)
+    Command.remove(@pkg, verbose: CREW_VERBOSE, force: @opt_false)
```
uberhacker, line 1847: Why not simply `force: false`?

Zopolis4: oops, typo. that's meant to say @opt_force

bin/crew
```diff
@@ -1297,7 +1297,7 @@ def install
   # remove it just before the file copy
   if @pkg.in_upgrade
     puts 'Attempting removal since this is an upgrade or reinstall...'
-    Command.remove(@pkg)
+    Command.remove(@pkg, verbose: CREW_VERBOSE, force: @opt_force)
```
uberhacker, line 1300: Why not use `@opt_verbose` for consistency?

Zopolis4: As it stands we don't have an `@opt_verbose` variable, only the constant. (something to do with `lib/fixup.rb`, maybe? i can't remember why the constant was added in favour of the variable)

uberhacker: Constants hold their value globally but the scope of instance variables are limited.

Zopolis4: true-- it was probably because we didn't pass anything to lib/fixup.rb, so the only way to make it verbose was through a constant. I'll see what I can do about that when i refactor `lib/fixup.rb`.

Zopolis4:
@uberhacker thank you for your help on this PR.

PR #11430, opened 2025-02-27
Vivaldi 7.1.3570.54-1 => 7.1.3570.58-1

uberhacker:
Tested & Working properly:
- [x] `x86_64` Unable to launch in hatch m132 container
- [x] `armv7l` Unable to launch in strongbad m132 container
```bash
CREW_REPO=https://github.com/uberhacker/chromebrew.git CREW_BRANCH=update-vivaldi crew update \
&& yes | crew upgrade
```


Zopolis4:
Got it. Unit tests are failing because of the removal of zlib in `test_force_remove_essential_package` causing the arm container to break.
```
FTesting the removal of normal package xxd_standalone. This should succeed.
ruby: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

#9707 fixes this by changing the removed essential package to zstd (I encountered a similar issue to this one on that PR, brought about by rubocop issues on i686, not sure why things started breaking on arm here), although I'd also be just as happy to drop the test because its a pretty risky idea to remove a critical package and hope everything works.

uberhacker:
> Got it. Unit tests are failing because of the removal of zlib in `test_force_remove_essential_package` causing the arm container to break.
> 
> ```
> FTesting the removal of normal package xxd_standalone. This should succeed.
> ruby: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
> ```
> 
> #9707 fixes this by changing the removed essential package to zstd (I encountered a similar issue to this one on that PR, brought about by rubocop issues on i686, not sure why things started breaking on arm here), although I'd also be just as happy to drop the test because its a pretty risky idea to remove a critical package and hope everything works.

Yes, we shouldn't be testing the removal of core packages.  We have 2k+ other packages we could test instead. :)

Zopolis4:
> Yes, we shouldn't be testing the removal of core packages. We have 2k+ other packages we could test instead. :)

It was your idea... :sweat_smile: https://github.com/chromebrew/chromebrew/pull/11365#discussion_r1961111269

Zopolis4:
> We absolutely should test the removal of core packages.
> 
> Removing core packages should fail. That's the point of testing to make sure an attempt at removing them doesn't break anything!

I agree-- What is causing the issue here is testing the forced removal of core packages, which works successfully but then (at least in the case of zlib) that package being missing causes crew to break and so it can't reinstall the package to restore functionality and then the rest of the tests break.

satmandu:
> > We absolutely should test the removal of core packages.
> > 
> > Removing core packages should fail. That's the point of testing to make sure an attempt at removing them doesn't break anything!
> 
> I agree-- What is causing the issue here is testing the forced removal of core packages, which works successfully but then (at least in the case of zlib) that package being missing causes crew to break and so it can't reinstall the package to restore functionality and then the rest of the tests break.

Feel free to bring back the essential files logic that prevented this situation! 😅👴

Zopolis4:
> > > We absolutely should test the removal of core packages.
> > > Removing core packages should fail. That's the point of testing to make sure an attempt at removing them doesn't break anything!
> > 
> > 
> > I agree-- What is causing the issue here is testing the forced removal of core packages, which works successfully but then (at least in the case of zlib) that package being missing causes crew to break and so it can't reinstall the package to restore functionality and then the rest of the tests break.
> 
> Feel free to bring back the essential files logic that prevented this situation! 😅👴

The essential files logic is still there, and works. Again, this is the forced removal of an essential package. This is expected to break things, and its just a lucky coincidence that the way we test this hasn't broken stuff until now.

satmandu:
> > > > We absolutely should test the removal of core packages.
> > > > Removing core packages should fail. That's the point of testing to make sure an attempt at removing them doesn't break anything!
> > > 
> > > 
> > > I agree-- What is causing the issue here is testing the forced removal of core packages, which works successfully but then (at least in the case of zlib) that package being missing causes crew to break and so it can't reinstall the package to restore functionality and then the rest of the tests break.
> > 
> > Feel free to bring back the essential files logic that prevented this situation! 😅👴
> 
> The essential files logic is still there, and works. Again, this is the forced removal of an essential package. This is expected to break things, and its just a lucky coincidence that the way we test this hasn't broken stuff until now.

Respectfully, the essential _files_ logic was designed to prevent the removal of files in essential packages (_and the packages they depend upon_.)

That's no longer there, right?

An attempt to remove an essential package should fail, but also, the attempt should be harmless because it should be impossible to delete the files in the package.

Sounds like this is no longer the case?

satmandu:
> What is being discussed here is the **forced removal** of an essential package. This was added in #10818. 

I stand corrected!

Why are we removing essential packages during testing?

We should not be forcing the removal of essential packages during testing? The point of an essential package is that they're supposed to be essential and forcing removal is expected to cause breakage?

Zopolis4:
> > What is being discussed here is the **forced removal** of an essential package. This was added in #10818.
> 
> I stand corrected!
> 
> Why are we removing essential packages during testing?
> 
> We should not be forcing the removal of essential packages during testing? The point of an essential package is that they're supposed to be essential and forcing removal is expected to cause breakage?

It was uberhackers request in https://github.com/chromebrew/chromebrew/pull/11365#discussion_r1961111269. I agree that the idea of testing a feature that is expected to break the entire program is a dubious one, but we can make it work, at least for now. I'm unbothered as to whether this should be fixed or removed, but we should decide one one option quickly as to fix unit tests.

uberhacker:
> We absolutely should test the removal of core packages.
> 
> Removing core packages should fail. That's the point of testing to make sure an attempt at removing them doesn't break anything!

Expected failed tests should not result in an overall failed result.  That is what we need to fix.

Zopolis4:
> Expected failed tests should not result in an overall failed result. That is what we need to fix.

They don't. That is not the issue. The issue is that the side effects of `test_force_remove_essential_package` break other tests. We can fix that by merging 9707 or splitting out changes from that and merging those, or we can remove the problematic test.

satmandu (approved):

PR #11456, opened 2025-03-03
Add comments to every package in lib/fixup.rb and simplify comment printing + misc fixes to lib/fixup.rb

Zopolis4:
## Description
Working on simplifying lib/fixup.rb, starting with adding comments to every package and assuming they're going to be there (I'm pretty sure the resulting code would still work if comments were nil, but its best practice to document the reason for the removal/rename of the package anyways)

Also, in testing this PR, I discovered that lib/fixup.rb is broken when updating in a number of ways, so I fixed the issues i ran across. We should probably set up CI for it, which I'll take a look at after this PR.

Tested and working on `x86_64` with both deprecation and renames.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=ghost crew update \
&& yes | crew upgrade
```

### DIFF
--- a/lib/fixup.rb
+++ b/lib/fixup.rb
@@ -127,14 +127,14 @@ pkg_update_arr = [
   { pkg_name: 'gnome_shell', pkg_rename: nil, pkg_deprecated: true, comments: 'No longer compatible with any architecture, requires systemd.' },
   { pkg_name: 'gnome_tweaks', pkg_rename: nil, pkg_deprecated: true, comments: 'No longer compatible with any architecture, requires systemd in dep package gnome_settings_daemon.' },
   { pkg_name: 'ilmbase', pkg_rename: nil, pkg_deprecated: true, comments: 'Included in openexr.' },
-  { pkg_name: 'imagemagick6', pkg_rename: nil, pkg_deprecated: true, comments: nil },
-  { pkg_name: 'imagemagick', pkg_rename: nil, pkg_deprecated: true, comments: nil },
+  { pkg_name: 'imagemagick6', pkg_rename: nil, pkg_deprecated: true, comments: 'Obsolete version of package, no longer required.' },
+  { pkg_name: 'imagemagick', pkg_rename: nil, pkg_deprecated: true, comments: 'Metapackage made redundant by the removal of imagemagick6.' },
   { pkg_name: 'itstool', pkg_rename: 'py3_itstool', pkg_deprecated: false, comments: 'Renamed to indicate switch to pip buildsystem.' },
   { pkg_name: 'js91', pkg_rename: nil, pkg_deprecated: true, comments: 'Replaced by newer versions of js.' },
   { pkg_name: 'jsonc', pkg_rename: 'json_c', pkg_deprecated: nil, comments: 'Renamed to better match upstream.' },
   { pkg_name: 'komodo', pkg_rename: nil, pkg_deprecated: true, comments: 'Abandoned upstream and no longer functional.' },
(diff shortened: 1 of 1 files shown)


lib/fixup.rb
```diff
@@ -127,14 +127,14 @@ def refresh_crew_json
   { pkg_name: 'gnome_shell', pkg_rename: nil, pkg_deprecated: true, comments: 'No longer compatible with any architecture, requires systemd.' },
   { pkg_name: 'gnome_tweaks', pkg_rename: nil, pkg_deprecated: true, comments: 'No longer compatible with any architecture, requires systemd in dep package gnome_settings_daemon.' },
   { pkg_name: 'ilmbase', pkg_rename: nil, pkg_deprecated: true, comments: 'Included in openexr.' },
-  { pkg_name: 'imagemagick6', pkg_rename: nil, pkg_deprecated: true, comments: nil },
-  { pkg_name: 'imagemagick', pkg_rename: nil, pkg_deprecated: true, comments: nil },
+  { pkg_name: 'imagemagick6', pkg_rename: nil, pkg_deprecated: true, comments: 'Obsolete version of package, no longer required.' },
+  { pkg_name: 'imagemagick', pkg_rename: nil, pkg_deprecated: true, comments: 'Metapackage made redundant by the removal of imagemagick6.' },
```
uberhacker, line 131: How about making this a fake package for imagemagick7 instead?

lib/fixup.rb
```diff
@@ -160,24 +160,24 @@ def refresh_crew_json
   { pkg_name: 'py3_pptx', pkg_rename: 'py3_python_pptx', pkg_deprecated: nil, comments: 'Fix to match upstream name.' },
   { pkg_name: 'pygobject', pkg_rename: 'py3_pygobject', pkg_deprecated: nil, comments: 'Renamed to use pip packaging.' },
   { pkg_name: 'q', pkg_rename: nil, pkg_deprecated: true, comments: 'No longer suitable for packaging.' },
-  { pkg_name: 'qtbase', pkg_rename: 'qt5_base', pkg_deprecated: nil, comments: 'Qt packages renamed to qt5_*' },
-  { pkg_name: 'qtcharts', pkg_rename: 'qt5_charts', pkg_deprecated: nil, comments: nil },
+  { pkg_name: 'qtbase', pkg_rename: 'qt5_base', pkg_deprecated: nil, comments: 'Renamed to allow qt6 packages.' },
+  { pkg_name: 'qtcharts', pkg_rename: 'qt5_charts', pkg_deprecated: nil, comments: 'Renamed to allow qt6 packages.' },
```
uberhacker, line 164: I actually think renaming them back to qt5_* ALLOWS for qt6 packages.  Once we have both, we'll need to distinguish between them.

lib/fixup.rb
Zopolis4: The imagemagick situation is a little complicated because of the way the metapackage could have installed things-- I think my original plan was to eventually rename imagemagick7 to imagemagick, but this is outside of the scope of this PR regardless.

lib/fixup.rb
Zopolis4: Exactly-- I figured specifying that here would make things more clear to users.

uberhacker: The only problem is users won't know to look here to find that out.

Zopolis4: These comments are printed when the user updates from an older version before the rename, so it will be made known to users that it applies to.

satmandu (approved):

PR #11523, opened 2025-03-12
Add FCEUX package

FinnBaltazar1111:
## Description

Add FCEUX package (it's a great NES emulator)

## Additional information
At the time of writing, the package only supports `x86_64` because, even though FCEUX supports other architectures, I don't know how to cross-compile and I don't have an `armhf`/`armv7l` 32-bit OS to compile on. But it may be supported in the future!

##
Builds:
- [x] `x86_64`
- [ ] `i686`
- [ ] `armv7l`
##
Tested & Working properly:
- [x] `x86_64`
- [ ] `i686`
- [ ] `armv7l`
##
##
```bash
CREW_REPO=https://github.com/FinnBaltazar1111/chromebrew.git CREW_BRANCH=fceux crew update \
&& yes | crew upgrade
```


manifest/x86_64/f/fceux.filelist
```diff
@@ -0,0 +1,111 @@
+/usr/local/bin/fceux
+/usr/local/lib64/liblua5.1-c++.so.0
+/usr/local/lib64/liblua5.1-c++.so.0.0.0
+/usr/local/lib64/liblua5.1.so.0
+/usr/local/lib64/liblua5.1.so.0.0.0
+/usr/local/lib64/libx264.so.155
+/usr/local/lib64/libx265.so.179
```
uberhacker, line 7: I'm wondering if these will create a potential conflict with other library packages.  Would it make sense to move into library packages instead?

packages/fceux.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def self.install
+    FileUtils.mkdir_p "#{CREW_DEST_PREFIX}/"
+    FileUtils.cp_r ".", "#{CREW_DEST_PREFIX}"
+  end
+
+  def self.postinstall
+    ExitMessage.add "\nInstallation complete. FCEUX executable is located in \"#{CREW_PREFIX}/bin/fceux\".\n"
+  end
+end
```
uberhacker, line 31: We'll need an entry in `tools/packages.yaml`.

manifest/x86_64/f/fceux.filelist
FinnBaltazar1111: FCEUX specifically requires Lua 5.1, and I'm pretty sure the existing Lua package contains the latest version. It also specifically looks for `libx264.so.155` and `libx265.so.179` (which aren't provided by Chromebrew packages `libx264` and `libx265` (they're different)), and creating symbolic links doesn't work, nor does changing the file name (doing so throws some kind of unknown symbol error when running `fceux`). I guess I could create packages for the specific versions of those libraries that are needed, but I'd like to finish up everything else in this PR before I potentially move to that.

packages/fceux.rb
FinnBaltazar1111: Right, right, I forgot that. I'll do that now.

FinnBaltazar1111: Done!

packages/fceux.rb
```diff
@@ -0,0 +1,31 @@
...
+  source_sha256 "7b85f7682ac89a148b8c2a43dd4b8c71e8effd719a3d8f251f119e56ba63052f"
+  
+  depends_on 'qt5_base'
+  depends_on 'minizip'
+  depends_on 'libsdl2'
+  depends_on 'ffmpeg'
+  depends_on 'libarchive'
+  depends_on 'zlib'
+  depends_on 'mesa_utils'
+  depends_on 'numactl'
```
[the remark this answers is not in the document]
FinnBaltazar1111: I'm assuming you meant sorting alphabetically, because there are no build dependencies (only runtime) as the package "source" is a deb.

manifest/x86_64/f/fceux.filelist
FinnBaltazar1111: I'm pretty sure that no library packages provide those specific shared object files. But, of course, I could be wrong.

uberhacker: What you could do is move these from `/usr/local/lib64` to `/usr/local/share/fceux/lib` and create a wrapper script to include the library path.  Something like below:
```
def self.build
    File.write 'fceux.sh', <<~EOF
      #!/bin/bash
      export LD_LIBRARY_PATH="#{CREW_PREFIX}/share/fceux/lib:$LD_LIBRARY_PATH"
      #{CREW_PREFIX}/bin/fceux "$@"
    EOF
end
```
This will enable you to use these specific library versions without potentially treading on other library packages.  Of course, you would need to remember to run `fceux.sh` to launch or you could rename `#{CREW_PREFIX}/bin/fceux` and use the new executable name in `fceux.sh` minus the `.sh` if that makes any sense.

packages/fceux.rb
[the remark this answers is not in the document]
uberhacker: Yes, the sorting helps us to identify duplicate depends_on lines.  It becomes more apparent with packages that have many dependencies.

PR #11564, opened 2025-03-17
Add qortal_hub package

uberhacker:
## Description
Qortal Hub is the newest interface for Qortal, part of the 'Qortal Trifecta' series of new User Interfaces for the platform/network.   See https://qortal.dev/.
##
Tested & Working properly:
- [x] `x86_64`
```bash
CREW_REPO=https://github.com/uberhacker/chromebrew.git CREW_BRANCH=add-qortal_hub-package crew update \
&& yes | crew upgrade
```

### DIFF
--- a/packages/qortal_hub.rb
+++ b/packages/qortal_hub.rb
@@ -1,0 +1,40 @@
+require 'package'
+
+class Qortal_hub < Package
+  description 'Desktop interface for Qortal'
+  homepage 'https://qortal.dev/'
+  version '0.5.2'
+  license 'MIT'
+  compatibility 'x86_64'
+  source_url "https://github.com/Qortal/Qortal-Hub/releases/download/v#{version}/Qortal-Hub_#{version}.AppImage"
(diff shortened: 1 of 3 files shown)


packages/qortal_hub.rb
```diff
@@ -0,0 +1,40 @@
+require 'package'
+
+class Qortal_hub < Package
+  description 'Desktop interface for Qortal'
+  homepage 'https://qortal.dev/'
+  version '0.5.2'
+  license 'Unknown'
```
Zopolis4, line 7:
```suggestion
  license 'MIT'
```

https://github.com/Qortal/Qortal-Hub/blob/feature/initial-conversion/electron/package.json#L24

Zopolis4 (changes requested):

packages/qortal_hub.rb
uberhacker: That is Electron only.  There is also Android code too.

Zopolis4 (approved):

PR #11566, opened 2025-03-17
Clean up rubocop configuration pt.1

Zopolis4:
## Description
Our rubocop configuration has been bugging me for a while, so I'm fixing it.

The main thing here is that every disabled cop now requires a comment explaining why we've done it, so that we aren't looking at our configuration guessing what each cop does and why we don't want that.

I am aware that I'm enabling some additional cops here, but I have looked through them all carefully and I believe these should be uncontroversial. If I am wrong, please let me know and we can discuss things.

I'm also splitting these changes into 3 prs:
- The cops that are easy to enable and have easily understood changes (thats this PR!)
- The cops that raise manual offenses that require refactoring of complicated code
- The cops that are likely to create massive diffs and hide all other changes

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=spectre crew update \
&& yes | crew upgrade
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -3,7 +3,7 @@
 require 'etc'
 
 OLD_CREW_VERSION ||= defined?(CREW_VERSION) ? CREW_VERSION : '1.0'
-CREW_VERSION ||= '1.57.8' unless defined?(CREW_VERSION) && CREW_VERSION == OLD_CREW_VERSION
+CREW_VERSION ||= '1.57.9' unless defined?(CREW_VERSION) && CREW_VERSION == OLD_CREW_VERSION
 
 # Kernel architecture.
 KERN_ARCH ||= Etc.uname[:machine]
@@ -61,7 +61,7 @@ else
 end
 
 # Use sane minimal defaults if in container and no override specified.
-CREW_KERNEL_VERSION ||= \
+CREW_KERNEL_VERSION ||=
   if CREW_IN_CONTAINER && ENV.fetch('CREW_KERNEL_VERSION', nil).nil?
     ARCH.eql?('i686') ? '3.8' : '5.10'
   else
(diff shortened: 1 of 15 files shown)


Zopolis4:
i love when you make a behaviour-identical change to a package file so it then gets run through CI and it turns out the package has been broken for who knows how long so now you need to fix the package before the PR passes ci :star_struck:

Zopolis4:
Disabling `Style/MultilineBlockChain` so it doesn't require me to modify `packages/jdk.rb` so that way CI doesn't try to run it and it breaks. i'll figure out why jdk.rb works on an x86_64 container but not the armv7l CI later.

uberhacker:
> Disabling `Style/MultilineBlockChain` so it doesn't require me to modify `packages/jdk.rb` so that way CI doesn't try to run it and it breaks. i'll figure out why jdk.rb works on an x86_64 container but not the armv7l CI later.

We can probably get rid of all the jdk*.rb packages.  We're using openjdk*.rb now since Oracle has essentially made it closed instead of open source.

lib/const.rb
```diff
@@ -61,7 +61,7 @@
 end
 
 # Use sane minimal defaults if in container and no override specified.
-CREW_KERNEL_VERSION ||= \
+CREW_KERNEL_VERSION ||=
```
uberhacker, line 64: This is a line continuation indicator and probably should remain.

packages/qemacs.rb
```diff
@@ -20,12 +20,12 @@ class Qemacs < Package
...
-           '--disable-xv', \
-           '--disable-xrender', \
-           '--disable-html', \
+    system './configure',
+           "--prefix=#{CREW_PREFIX}",
+           '--disable-x11',
+           '--disable-xv',
+           '--disable-xrender',
+           '--disable-html',
            '--disable-png'
```
uberhacker, line 29: Why not remove all the quotes too?

lib/const.rb
Zopolis4: This change is from `Style/RedundantLineContinuation`: https://docs.rubocop.org/rubocop/cops_style.html#styleredundantlinecontinuation

From what I can tell, this is safe to autocorrect and does not change behavior at all.

Zopolis4:
> > Disabling `Style/MultilineBlockChain` so it doesn't require me to modify `packages/jdk.rb` so that way CI doesn't try to run it and it breaks. i'll figure out why jdk.rb works on an x86_64 container but not the armv7l CI later.
> 
> We can probably get rid of all the jdk*.rb packages. We're using openjdk*.rb now since Oracle has essentially made their jdk closed instead of open source.

I think that's waiting on https://github.com/chromebrew/chromebrew/issues/7476

packages/qemacs.rb
Zopolis4: I agree! I'll tackle that in a different PR.

uberhacker (approved):

PR #11568, opened 2025-03-18
Simplify checking for updated crew version when updating

Zopolis4:
## Description
This removes one of the main pieces of shared state between `bin/crew` and `lib/fixup.rb`.

Probably most of the git fixup logic in `def update` should be in `lib/fixup.rb`, but this bit in particular works much better when we are actually doing the updating and have all the requisite git information.

##
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=tion crew update \
&& yes | crew upgrade
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -219,10 +219,6 @@ def update
   abort "'crew update' is used to update crew itself. Use 'crew upgrade <package1> [<package2> ...]' to update specific packages.".orange if @pkg_name
 
   unless CREW_NO_GIT
-    # The following is used in fixup.rb to determine if crew update needs to
-    # be run again.
-    @crew_const_git_commit = `git -C #{CREW_LIB_PATH} log -n1 --oneline #{CREW_LIB_PATH}/lib/const.rb`.split.first
-
     unless Dir.exist?(File.join(CREW_LIB_PATH, '.git'))
       puts 'Fixing Chromebrew system git repo clone...'.orange
       system(<<~GIT_REPAIR_COMMANDS, chdir: CREW_LIB_PATH, %i[out err] => File::NULL)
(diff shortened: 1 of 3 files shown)


satmandu:
Does this work from a really old container that hasn't been updated in forever?

Since you're working with logic meant to deal with really old installs...

Zopolis4:
Yup-- tested it on a version a few weeks old, with all combinations of const.rb commit updates and orders and so on.

satmandu:
> Can you run it on an image older than a year or two? Do we have those?


Example image to test with:

Note that this requires `gem update` to be run as part of the update process...

```bash
#!/bin/bash
if [ -n "$SSH_CLIENT" ] || [ -n "$SSH_TTY" ]; then
  SESSION_TYPE=remote/ssh
elif pstree -p | egrep --quiet --extended-regexp ".*sshd.*\($$\)"; then
  SESSION_TYPE=remote/ssh
else
  case $(ps -o comm= -p $PPID) in
    sshd|*/sshd) SESSION_TYPE=remote/ssh;;
  esac
fi
if [ -z ${PAGER+x} ]; then
  echo "PAGER is not set."
else
  PAGER_PASSTHROUGH=-e
  PAGER_PASSTHROUGH+=" "
  PAGER_PASSTHROUGH+=CONTAINER_PAGER=${PAGER}
fi
X11+=" "
X11=-e
X11+=" "
X11+=DISPLAY=${DISPLAY:-:0.0}
X11+=" "
if ! [[ $SESSION_TYPE == remote/ssh ]] && [ -d /tmp/.X11-unix ]; then
  X11+=" -v /tmp/.X11-unix:/tmp/.X11-unix "
fi
if [ -f "$HOME"/.Xauthority ]; then
  X11+=--volume=$HOME/.Xauthority:/home/chronos/user/.Xauthority:rw
  X11+=" "
  X11+=--volume=$HOME/.Xauthority:/home/chronos/.Xauthority:rw
fi
docker pull --platform linux/amd64 satmandu/crewbuild:nocturne-x86_64.m116
docker pull tonistiigi/binfmt
docker run --privileged --rm tonistiigi/binfmt --install all
docker run --platform linux/amd64 --rm --net=host ${PAGER_PASSTHROUGH} ${X11} -e LOCALRC="${LOCALRC}" -v $(pwd)/pkg_cache:/usr/local/tmp/packages -v $(pwd):/output -h $(hostname)-x86_64 -it satmandu/crewbuild:nocturne-x86_64.m116
```

Zopolis4:
> Can you run it on an image older than a year or two? Do we have those?

This only modifies the logic for checking if const.rb was updated, and the solution I used is very self-contained and shouldn't be affected by anything else. If you really think it's necessary to test on a version multiple years old, I will, I just want to double check why first.

satmandu (approved):

PR #11588, opened 2025-03-22
Refactor tools/update_python_pip_packages.rb to not be loaded and split into functions

Zopolis4:
## Description
Running this via `system()` instead of `load()` lets us simplify some of the logic, as well as cleaning up shared state for a larger simplification later. Additionally, the splitting into functions here is a step towards implementing automatic updates for generic packages.

The git diff looks a little aggressive (at least when viewing whitespace) due to the functions, but I've only really changed about 10 lines or so.

Tested and working by going to an older container where the python/pip packages are not up to date and running it, both standalone and via `tools/build_updated_packages.rb`.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=pool crew update \
&& yes | crew upgrade
```

### DIFF
--- a/tools/build_updated_packages.rb
+++ b/tools/build_updated_packages.rb
@@ -59,7 +59,7 @@ if SKIP_UPDATE_CHECKS
   puts 'Skipping pip and gem remote update checks.'.orange
 else
   puts 'Checking for pip package version updates...'.orange
-  load 'tools/update_python_pip_packages.rb'
+  Kernel.system 'tools/update_python_pip_packages.rb'
   puts 'Checking for ruby gem package version updates...'.orange
   load 'tools/update_ruby_gem_packages.rb'
 end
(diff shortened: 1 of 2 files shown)


tools/build_updated_packages.rb
```diff
@@ -59,7 +59,7 @@ def self.check_build_uploads(architectures_to_check = nil, name = nil)
   puts 'Skipping pip and gem remote update checks.'.orange
 else
   puts 'Checking for pip package version updates...'.orange
-  load 'tools/update_python_pip_packages.rb'
+  system 'tools/update_python_pip_packages.rb'
```
satmandu, line 62: Do you want to use Kernel.system since we override system?

Zopolis4: Good point. Fixed.

satmandu (approved):

Issue #7476, opened 2022-10-10
Check to see if packages can be switched from jdk8 to openjdk8


satmandu:
Check to see if packages which have dependencies on `jdk8` also build/run with `openjdk8` and if so switch the dependencies.

satmandu:
As of today:
```
antlr4.rb:  depends_on 'jdk8'
jobscheduler.rb:  depends_on 'jdk8'
plplot.rb:  depends_on 'jdk8'
projectlibre.rb:  depends_on 'jdk8'
pycharm.rb:  depends_on 'jdk8'
rhino.rb:  depends_on 'jdk8'
rsu_client.rb:  depends_on 'jdk8'
sbt.rb:  depends_on 'jdk8'
selenium_server_standalone.rb:  depends_on 'jdk8'
swi_prolog.rb:  depends_on 'jdk8'
upm.rb:  depends_on 'jdk8'
vuze.rb:  depends_on 'jdk8'
xz_java.rb:  depends_on 'jdk8'
```

Zopolis4:
As of now:
```
./jobscheduler.rb:  depends_on 'jdk8'
./projectlibre.rb:  depends_on 'jdk8'
./pycharm.rb:  depends_on 'jdk8'
./rhino.rb:  depends_on 'jdk8'
./rsu_client.rb:  depends_on 'jdk8'
./sbt.rb:  depends_on 'jdk8'
./selenium_server_standalone.rb:  depends_on 'jdk8'
./upm.rb:  depends_on 'jdk8'
./vuze.rb:  depends_on 'jdk8' unless CREW_IN_CONTAINER
./xz_java.rb:  depends_on 'jdk8'
```

PR #11616, opened 2025-03-26
Switch java packages without new versions to openjdk8

Zopolis4:
## Description
Working on #7476. I've switched all the packages that haven't had new releases and so don't need to be updated to openjdk8. Admittedly some of these packages may need additional inspection (vuze was forked and upstream development has stopped, rsu_client was archived upstream due to no longer working with rs3) but I've confirmed to the best of my ability that the switch to openjdk8 has not broken these packages any more than they already are.

##
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=sofar crew update \
&& yes | crew upgrade
```


uberhacker (approved):
Good catch!

PR #11623, opened 2025-03-28
Refactor tools/update_python_pip_packages.rb to use package objects

Zopolis4:
## Description
Behavior should be identical apart from a minor grammar change in a print, and possibly increased accuracy in version comparisons from using libversion.

In terms of refactoring, I switched from using sed and grep to loading packages, fixing the `$LOAD_PATH` in order to do so. (We should probably take a look at all the LOAD_PATH modifications and see if we can make them work when the program is ran from any directory). Also some other rubifications, mainly for text extraction and insertion.

Also had a look at the concurrency code, concluded I do not understand it, and thus left it alone.

Tested and working inside an `x86_64` container. Had libversion load path issues locally, and then different issues related to `LD_LIBRARY_PATH` not including `/usr/local` on a github codespace. Both of these were fixed by environment variables and I've confirmed the code works after setting the requisite variables. Haven't been able to test in a workflow run, but I believe it will work.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=wist crew update \
&& yes | crew upgrade
```


tools/update_python_pip_packages.rb
```diff
@@ -6,47 +6,57 @@
...
+$LOAD_PATH.unshift './lib'
+
 require_relative '../lib/color'
 require_relative '../lib/const'
+require_relative '../lib/package'
+require_relative '../lib/package_utils'
 require_relative '../lib/require_gem'
-require_gem('activesupport', 'active_support/core_ext/object/blank')
 require_gem 'concurrent-ruby'
+require_gem 'ruby-libversion', 'ruby_libversion'
```
uberhacker, line 18: Why do we need 2 different gems?

Zopolis4: We need to install the gem ruby-libversion, and then require it as ruby_libversion

.github/workflows/Updater.yml
```diff
@@ -27,8 +27,16 @@ jobs:
...
+      - name: Install ruby-libversion  # Hopefully this will get added as an Ubuntu/Debian package so we don't have to do this manually.
+        run: |
+          git clone https://github.com/repology/libversion
+          cd libversion
+          mkdir build
+          cd build
+          cmake ..
+          make -j $(nproc)
+          sudo make install
+          sudo gem install ruby-libversion
```
uberhacker, line 39: How long does this build process take?

Zopolis4: probably about 30 seconds, but i'll swap it to a shallow clone which should cut the total time to around 15

satmandu: Do you know how to create an Ubuntu ppa for a debian package? That would the easiest way of getting a pre-built Ubuntu package here... Well, maybe that would only save ~ 15 seconds...

satmandu:
I'm really happy for another set of eyes on this!

Zopolis4:
> I'm really happy for another set of eyes on this!

Phew! I was worried I was stepping on your toes a little bit here... :sweat_smile:

.github/workflows/Updater.yml
Zopolis4: neither ubuntu nor debian package ruby-libversion or the dependent libversion library, so we'd be starting from scratch. I looked into adding those packages to debian, but its more trouble than its worth, imo.

satmandu:
> > I'm really happy for another set of eyes on this!
> 
> Phew! I was worried I was stepping on your toes a little bit here... :sweat_smile: 

It was a pain to setup, but I'm glad that people understand the process enough to iterate on it and improve it.

satmandu (approved):

PR #11671, opened 2025-04-01
Refactor tools/update_ruby_gem_packages.rb to use package objects, the rubygems.org api, and more rubification

Zopolis4:
## Description
Pretty similar to last time, but I swapped to directly querying the rubygems API, which lets us simplify a lot of things, lets us check the versions of all ruby packages (i.e. we can check rubocop-chromebrew now), and might be a little bit faster.

Also, I added support for stripping ruby versions in `PackageUtils.get_clean_version` so we can use that here, and it will also stop all of our ruby packages being marked as out of date in Repology (thats an additional ~130 packages marked as up to date now).

Tested and working in container and in codespace.
```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=portrait crew update \
&& yes | crew upgrade
```

### DIFF
--- a/.github/workflows/Updater.yml
+++ b/.github/workflows/Updater.yml
@@ -79,7 +79,7 @@ jobs:
         env:
           UPDATE_BRANCH_NAME: ${{ steps.set-variables.outputs.UPDATE_BRANCH_NAME }}
         run: |
-          ruby tools/update_ruby_gem_packages.rb
+          LD_LIBRARY_PATH=/usr/local/lib ruby tools/update_ruby_gem_packages.rb
           # Create a new branch with the updated package files only
           # if there are updated packages. Otherwise exit early.
           if [ -n "$(git status --porcelain)" ]; then
--- a/tools/update_ruby_gem_packages.rb
+++ b/tools/update_ruby_gem_packages.rb
@@ -5,45 +5,68 @@
 # tools/update_ruby_gem_packages.rb
 
 # Add >LOCAL< lib to LOAD_PATH
-$LOAD_PATH.unshift '../lib'
+$LOAD_PATH.unshift './lib'
+
+require 'json'
+require 'net/http'
 require_relative '../lib/color'
 require_relative '../lib/const'
-require_relative '../lib/gem_compact_index_client'
+require_relative '../lib/package'
+require_relative '../lib/package_utils'
 require_relative '../lib/require_gem'
 require_gem 'concurrent-ruby'
+require_gem 'ruby-libversion', 'ruby_libversion'
 
 def check_for_updated_ruby_packages
-  gems = BasicCompactIndexClient.new.gems
-
+  # Create a thread pool for parallelization.
   pool = Concurrent::ThreadPoolExecutor.new(
     min_threads: 1,
     max_threads: CREW_NPROC.to_i + 1,
     max_queue: 0, # unbounded work queue
     fallback_policy: :caller_runs
   )
+
+  # Currently, the only packages that use the ruby buildsystem are ruby_* packages, so this finds all the packages we need to check.
   relevant_gem_packages = Dir['packages/ruby_*.rb']
+
+  # Get the total number of files to check, and then the length of that number, so status updates can be formatted.
   total_files_to_check = relevant_gem_packages.length
   numlength = total_files_to_check.to_s.length
   updateable_packages = {}
+  packages_without_gem_versions = []
   relevant_gem_packages.each_with_index do |package, index|
     pool.post do
-      untested_package_name = package.gsub(%r{^packages/ruby_}, '').gsub(/.rb$/, '')
-      gem_test = gems.grep(/#{"^#{untested_package_name}\\s.*$"}/).last.blank? ? gems.grep(/#{"^\(#{passed_name.gsub(/^ruby_/, '').gsub('_', ')*.(')}\\s\).*$"}/).last : gems.grep(/#{"^#{untested_package_name}\\s.*$"}/).last
-      abort "Cannot find #{passed_name} gem to install.".lightred if gem_test.blank?
-      gem_test_name = gem_test.split.first
-      gem_test_versions = gem_test.split[1].split(',')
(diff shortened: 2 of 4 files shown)


Zopolis4:
There's also a json API for pypi, which we could switch to using, but our current solution for querying python versions is already quite simple, so it wouldn't be as big of a difference

tools/update_ruby_gem_packages.rb
```diff
@@ -5,55 +5,63 @@
...
-      # https://github.com/rubygems/rubygems/blob/b5798efd348935634d4e0e2b846d4f455582db48/lib/rubygems/version.rb#L305
-      gem_test_versions.delete_if { |i| i.match?(/[a-zA-Z]/) }
-      gem_test_version = gem_test_versions.max
-      gem_name = gem_test_name.blank? ? Gem::SpecFetcher.fetcher.suggest_gems_from_name(untested_package_name).first : gem_test_name
-      gem_version = gem_test_name.blank? ? Gem.latest_version_for(untested_package_name).to_s : gem_test_version
-      next package if gem_version.blank?
-
-      relevant_gem_packages.delete(package)
+      pkg = Package.load_package(package)
+      gem_name = pkg.name.sub('ruby_', '').gsub('_', '-')
```
satmandu, line 40: This doesn't replace the logic above, which is there for a reason: some gems have a mix of underscores and dashes in their names!

satmandu (changes requested):

tools/update_ruby_gem_packages.rb
Zopolis4: Ah. Some packages do benefit from this logic, like rubocop-chromebrew, so I'll see about setting up some special cases.

uberhacker: How about?
```
gem_name = pkg.name.sub('ruby_', '').gsub('_', '-') unless pkg.name.include?('-') && pkg.name.include?('_')
```

Zopolis4: oooh

Zopolis4: Fixed! I had to use manual special cases, because we don't actually know whether the gem originally used underscores or dashes. Still, we now check every package:
```
Done checking rubygems for updates to 133 ruby packages.
Updated versions were not listed in rubygems for: 
```

tools/update_ruby_gem_packages.rb
```diff
@@ -5,55 +5,77 @@
...
+        # These gems have an underscore then a dash, but there's only one, so we hardcode the logic for now.
+        gem_name = 'language_server-protocol'
+      elsif gem_name == 'unicode_display_width'
+        # These gems have a dash then an underscore, but there's only one, so we hardcode the logic for now.
+        gem_name = 'unicode-display_width'
+      else
+        # In the common case, the gem name used only dashes, which we all replaced with underscores.
+        gem_name.gsub!('_', '-')
+      end
+
```
satmandu, line 55: I don't think that whitelisting all gems with variant usage of dashes & underscores is resilient to a new package having the same issue?

What's wrong with just having a code path that checks for all possibilities if a dash or underscore may be in the gem name, like before?

satmandu (changes requested):

tools/update_ruby_gem_packages.rb
Zopolis4: The previous codepath didn't work:
```
Done checking for updates to 133 ruby gems.
Updated versions were could not be found for: concurrent-ruby gem-compiler io-console io-nonblock io-wait language-server-protocol net-ftp net-http net-imap net-pop net-protocol net-smtp net-telnet open-uri pry-byebug readline-ext resolv-replace rubocop-ast rubocop-chromebrew ruby-libversion ruby-progressbar rubygems-update test-unit unicode-display-width unicode-emoji yaml-lint
```

Zopolis4: Actually, that raises the question of whether the logic for installing these gems works properly, given that its the one I replaced here.

Zopolis4: Would it make more sense to just have a property in gem packages denoting their upstream name where it differs from a simple `.gsub('_','-')`? (Obviously that'd be something for a different PR)

satmandu: Maybe? Or maybe we should figure out a way to allow dashes and underscores in our package names?

Zopolis4: That would also be a good solution-- something for a second PR, perhaps?

.github/workflows/Updater.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         env:
           UPDATE_BRANCH_NAME: ${{ steps.set-variables.outputs.UPDATE_BRANCH_NAME }}
         run: |
-          ruby tools/update_ruby_gem_packages.rb
+          LD_LIBRARY_PATH=/usr/local/lib ruby tools/update_ruby_gem_packages.rb
```
uberhacker, line 82: This shouldn't be necessary and not advisable for x86_64.

Zopolis4: This is a requirement to load ruby-libversion in the github actions (not inside the chromeos image) container. We already do this for python packages.

uberhacker: Did we install ruby-libversion incorrectly then?  All libraries for x86_64 should go in a `lib64` directory.

Zopolis4: Some distributions don't make that distinction, including the ones our actions run on.

uberhacker: That is fair but in this case libversion was installed from source so the real reason is because ruby-libversion needs to find libversion.so in `/usr/local/lib`.  Typically Debian-based distros store libraries in `/lib` or `/usr/lib`.

Zopolis4: Yes.

satmandu (dismissed):

PR #11719, opened 2025-04-15
Adjust ruby buildessential deps

satmandu:


uberhacker (commented):
This is just temporary, correct?

uberhacker (approved):

PR #11766, opened 2025-04-19
Remove Skype package

supechicken:
Microsoft has announced the plan to sunset Skype [earlier this year](https://www.microsoft.com/en-us/microsoft-365/blog/2025/02/28/the-next-chapter-moving-from-skype-to-microsoft-teams/), and it will stop working in May.

Let's remove it from our repository now...


uberhacker:
This is disappointing but not surprising due to M$'s long history of using the [embrace, enhance and extinguish](https://en.wikipedia.org/wiki/Embrace,_extend,_and_extinguish) strategy.  You'll need to add an entry to `lib/fixup.rb`.

supechicken:
> You'll need to add an entry to lib/fixup.rb
Done

uberhacker (approved):

PR #11789, opened 2025-04-21
Adjust pip updater to have update exclusion list.

satmandu:
- Some pip packages are broken or don't need to be constantly updated, so allow those to be excluded from the update checks.

### DIFF
--- a/tools/update_python_pip_packages.rb
+++ b/tools/update_python_pip_packages.rb
@@ -18,9 +18,9 @@ require_gem 'concurrent-ruby'
 require_gem 'ruby-libversion', 'ruby_libversion'
 
 def check_for_updated_python_packages
-  # Some packages should be excluded from the
-  # version update check.
-  # Some packages have daily updates (awscli) and some just don't build.
+  # Some packages should be excluded from the version update check,
+  # because some packages have daily updates (awscli) and some packages
+  # have broken builds. (ldapdomaindump)
   ignored_packages = %w[ldapdomaindump awscli]
 
   # Sets the correct pip configuration values if they are not already set.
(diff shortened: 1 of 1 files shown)


tools/update_python_pip_packages.rb
```diff
@@ -18,6 +18,11 @@
 require_gem 'ruby-libversion', 'ruby_libversion'
 
 def check_for_updated_python_packages
+  # Some packages should be excluded from the version update check,
+  # because some packages have daily updates (awscli) and some packages
+  # have broken builds. (ldapdomaindump)
+  ignored_packages = %w[ldapdomaindump awscli]
```
uberhacker, line 24: Does this mean these packages are no longer checked?

satmandu: Exactly that. That's the intent. We need a way of manually excluding packages from the automatic update check.

uberhacker: That is a problem when you do want to eventually update, right?

uberhacker: Novel idea: Why don't we fix the broken packages?

satmandu: The broken packages hold up the rest of the updates, and the list of packages that needs updates grows.

We really need to be firing off a separate PR and branch for each updated package... But even in that case, we want to avoid firing off a PR and new branch each day for a package that we don't currently have a fix for.

Somewhere we need to maintain a broken package list...

uberhacker: Alright but hopefully, this is only a temporary fix.  My concern is we'll forget about it and not even know we are missing updates.

uberhacker (approved):

PR #11790, opened 2025-04-21
Abstract the latest llvm version

uberhacker:
This PR abstracts the latest llvm version to enable major release updates in one place (core.rb) rather than nearly 20 packages.  Tested and working.

```bash
CREW_REPO=https://github.com/uberhacker/chromebrew.git CREW_BRANCH=abstract-llvm-version crew update \
&& yes | crew upgrade
```

### DIFF
--- a/packages/distcc.rb
+++ b/packages/distcc.rb
@@ -23,15 +23,10 @@ class Distcc < Autotools
   depends_on 'avahi' # R
   depends_on 'ccache' # L
   depends_on 'gcc_dev' # L
-  depends_on 'gcc_lib' # R
   depends_on 'glibc' # R
-  depends_on 'glib' # R
-  depends_on 'llvm20_dev' # L
-  depends_on 'llvm20_lib' # R
-  depends_on 'nss_mdns' # R
+  depends_on 'llvm_dev' # L
   depends_on 'popt' # R
   depends_on 'python3' => :build
-  depends_on 'zlib' # R
 
   print_source_bashrc
 
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -3,7 +3,7 @@
 require 'etc'
 
 OLD_CREW_VERSION ||= defined?(CREW_VERSION) ? CREW_VERSION : '1.0'
-CREW_VERSION ||= '1.58.6' unless defined?(CREW_VERSION) && CREW_VERSION == OLD_CREW_VERSION
+CREW_VERSION ||= '1.58.7' unless defined?(CREW_VERSION) && CREW_VERSION == OLD_CREW_VERSION
 
 # Kernel architecture.
 KERN_ARCH ||= Etc.uname[:machine]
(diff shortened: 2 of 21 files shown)


satmandu:
I don't think this is going to work... because packages, when built against llvm, are linked to versioned llvm libraries.

See mesa:
```
lddtree /usr/local/lib64/libEGL_mesa.so 
 /usr/local/lib64/libEGL_mesa.so (interpreter => None)
      libgallium-25.0.3.so => /usr/local/lib64/libgallium-25.0.3.so
          libLLVM.so.20.1 => /usr/local/lib64/libLLVM.so.20.1
...
```

Having said that, running tools/getrealdeps.rb on several of the packages in this PR shows that they don't actually have runtime deps on any llvm_lib packages, so the llvm_lib dependencies can be removed from those packages!

uberhacker:
> I don't think this is going to work... because packages, when built against llvm, are linked to versioned llvm libraries.
> 
> See mesa:
> 
> ```
> lddtree /usr/local/lib64/libEGL_mesa.so 
>  /usr/local/lib64/libEGL_mesa.so (interpreter => None)
>       libgallium-25.0.3.so => /usr/local/lib64/libgallium-25.0.3.so
>           libLLVM.so.20.1 => /usr/local/lib64/libLLVM.so.20.1
> ...
> ```
> 
> Having said that, running tools/getrealdeps.rb on several of the packages in this PR shows that they don't actually have runtime deps on any llvm_lib packages, so the llvm_lib dependencies can be removed from those packages!

That is fine.  This just abstracts the version from the package name.  When building a new major release, you would simply build all the llvm#_* packages and then bump the version in core.rb.  The new version should just pass through the abstraction packages.  The reason the mesa package failed here is because the version was inadvertently bumped.

satmandu:
We should check every one of these non-llvm packages to make sure that they still need a llvm_lib runtime dependency at all. Some of them definitely do not. Instead of substituting in ` depends_on 'llvm_lib' # R`, that line should just be removed if it isn't needed.

`tools/getrealdeps.rb <package>` should give you that information.

uberhacker:
> We should check every one of these non-llvm packages to make sure that they still need a llvm_lib runtime dependency at all. Some of them definitely do not. Instead of substituting in ` depends_on 'llvm_lib' # R`, that line should just be removed if it isn't needed.
> 
> `tools/getrealdeps.rb <package>` should give you that information.

Done

packages/distcc.rb
```diff
@@ -22,16 +22,9 @@ class Distcc < Autotools
 
   depends_on 'avahi' # R
   depends_on 'ccache' # L
-  depends_on 'gcc_dev' # L
```
satmandu, line 25: The logical deps lines are intentionally here...

packages/distcc.rb
```diff
@@ -22,16 +22,9 @@ class Distcc < Autotools
...
-  depends_on 'gcc_lib' # R
   depends_on 'glibc' # R
-  depends_on 'glib' # R
-  depends_on 'llvm20_dev' # L
```
satmandu, line 29: Same here.

packages/distcc.rb
uberhacker: Resolved.

satmandu (approved):

Issue #9612, opened 2024-04-06
🐞 Bug ruby and other programs cause a stack smashing error on ChromeOS 123, limiting chromebrew operation


noxxwk:
**Describe the bug**
Ruby and other programs (e.g. Firefox or crew which runs on ruby) cause a stack smashing error and terminates. Very little chromebrew functionality still works.

I tried reinstalling from scratch but the issue remains.

**To Reproduce**

1. I executed this command:

      ```shell
      crew
      ```

      Output:

      ```
      *** stack smashing detected ***: terminated
      Segmentation fault (core dumped)
      ```
     Sometimes you'll get an aborted error:
     ```
     Aborted (core dumped)
      ```

**Expected behavior**
crew and other programs execute.

**Environment: Paste the output of `crew sysinfo -v` between the two `***` below**

crew no longer runs so it's not possible to get the output of `crew sysinfo -v`, but this is a brya device (Asus Vibe CX34) on ChromeOS 123 stable.

I tested and the same issue also occurs on an eve device (Pixelbook).

The behavior began after updating to ChromeOS 123 stable.

luhgarlicbread:
Yeah I've gotten this error too. I just decided to downgrade to ChromeOS v116, so I could have sudo in Crosh still.

luhgarlicbread:
You can downgrade by going on https://cros.tech , searching for your Chromebook's board, and downloading a v116 recovery image.
Flash it to your USB by hitting the gear icon and selecting "use local image" on the Chromebook Recovery Utility extension for Chromium based browsers.

yestgoqp9yd1:
cuz of glibc_235

supechicken:
@yestgoqp9yd1 Oh sorry, I think I accidentally deleted your comment, here is the original comment:
```text
cuz of glibc_235
```

supechicken:
Quick fix:
```shell
unset LD_LIBRARY_PATH
cat /usr/local/etc/crew/meta/glibc_lib235.filelist | grep .so | grep -v gconv | xargs rm -rf
export LD_LIBRARY_PATH=/usr/local/lib64
```

VenomousSteam81:
@supechicken It breaks when you try to install atm, and the second command gives you an error saying that `/usr/local/etc/crew/meta/glibc_lib235.filelist` doesn't exist. To undo that, you would need to run:
```
/usr/bin/sudo /bin/bash 
sudo rm -rf /usr/local/*
```

[...]

PR #11806, opened 2025-04-23
Add a standalone glibc package

supechicken:
Instead of compiling full glibc builds against the system glibc version, why don't we have our installation of glibc that is completely separated from the system one?

By having a standalone glibc installation, we don't need to refresh our glibc builds each time ChromeOS updates its glibc version anymore (and thus no more glibc breakage like #9612 😀)

To prevent conflicts with system glibc, all libraries will be placed in `#{CREW_PREFIX}/opt/glibc-libs`, and any value set in `LD_LIBRARY_PATH` will be ignored.

To test it, set interpreter path to `/usr/local/opt/glibc-libs/ld-linux-x86-64.so.2`:
```shell
patchelf --set-interpreter /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2 /usr/local/bin/ruby
```

```bash
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=standalone_glibc crew update \
&& yes | crew upgrade
```


supechicken:
> Are you reinventing a chroot?

There is no denying that having another libc makes Chromebrew closer to a chroot, but it does make our maintenance easier and provides better compatibility for older systems...

The biggest characteristic of chroot is that it is fully isolated from the host, making command calls from chroot to host commands impossible. On the other hand, we are still able to access built-in commands from our packages without any trouble, and vice versa.

packages/glibc_standalone.rb
```diff
@@ -0,0 +1,101 @@
...
+    File.write 'hardcode-library-path.patch', <<~EOF
+      diff -Nur a/elf/dl-support.c b/elf/dl-support.c
+      --- a/elf/dl-support.c    2025-04-23 01:58:17.265974714 +0800
+      +++ b/elf/dl-support.c    2025-04-23 14:38:27.373033131 +0800
+      @@ -296,7 +296,7 @@
+
+         /* Initialize the data structures for the search paths for shared
+            objects.  */
+      -  _dl_init_paths (getenv ("LD_LIBRARY_PATH"), "LD_LIBRARY_PATH",
+      +  _dl_init_paths ("#{CREW_PREFIX}/opt/glibc-libs:#{CREW_LIB_PREFIX}:/#{ARCH_LIB}", "LD_LIBRARY_PATH",
```
satmandu, line 35: Why `#{CREW_LIB_PREFIX}:/#{ARCH_LIB}`? Isn't that redundant?

satmandu:
Don't get me wrong, this is a _clever_ hack.

Let's call this a Chromebrew2 environment for argument's sake.

For this to work we need to figure out how to:
1. Crosh window opens
2. `.bashrc` and/or `.bash_profile` loads.
3. We use that to immediately invoke our special shell modified with patchelf to point to our glibc. (_Is this not effectively a chroot?_)  e.g., `[[ $LD_LIBRARY_PATH = *opt/glibc-libs* ]] || LD_LIBRARY_PATH=#{CREW_PREFIX}/opt/glibc-libs:$LD_LIBRARY_PATH #{CREW_PREFIX}/bin/bash -l`

We still want to be able to access system commands which are linked to the system glibc. (Could we just have a special shell to do that?)

I guess I don't get the advantage of this approach vs just directly invoking a chroot, maybe with an option when you start `crosh` to either open a regular ChromeOS shell, or open a chroot.

Once you have set your LIBRARY_PATH to include another directory to supersede the ChromeOS library directories, why not take advantage of a chroot and get access to all the software already written for a standard Linux install in that chroot installed in the usual Linux install directories?

packages/glibc_standalone.rb
supechicken: It is ChromeOS's default `LD_LIBRARY_PATH`:
```
$ env | grep LD_LIBRARY
LD_LIBRARY_PATH=/usr/local/lib64:/lib64
```

supechicken:
I thought you were against this idea when I first read your comment 😅

> We still want to be able to access system commands which are linked to the system glibc. (Could we just have a special shell to do that?)

To use our glibc, just change the ELF interpreter path to our `ld-linux` and we are all set. No extra modifications required :)

Here is a quick demonstration of what is going on:

1. A dynamic (non-static build) executable with ELF interpreter path set to `${CREW_PREFIX}/opt/glibc-libs/ld-linux-x86-64.so.2` was being run
2. Since the interpreter was pointed to our glibc installation, our dynamic linker will handle it (instead of the system one, located at `/lib64/ld-linux-x86-64.so.2`)
3. To prevent ChromeOS built-in glibc libraries from being called (`/lib64/libc.so.6`, `/lib64/libm.so.6` and etc), our dynamic linker will always override `LD_LIBRARY_PATH` so that glibc libraries in `${CREW_PREFIX}/opt/glibc-libs` will have the highest priority.
4. Executable is now running with our glibc libraries :)

And here is how a normal (unmodified) executable works:

1. ELF interpreter path is `/lib64/ld-linux-x86-64.so.2`, the system's dynamic linker will be used.
2. Search all required libraries from `LD_LIBRARY_PATH` 
3. Search all required libraries from `/etc/ld.so.cache` if not found in `LD_LIBRARY_PATH`

> But why not just Crouton is also a meaningful response to this question. :)

The answer would be "no root is required", especially when there are no more `sudo` in crosh now 🗡️

supechicken:
> We would have to make sure all binaries are compatible with the glibc version installed or nothing may work. We are treading very sensitive ground here. Changing the glibc version would break all system binaries and unless there was a crew package equivalent installed, breakage will occur.

It should be fine as glibc is backward compatible. Anything compiled against an older glibc version should also work on newer glibc releases.

Also, I have configured it to put all of its libraries under `#{CREW_PREFIX}/opt/glibc-libs` (instead of `#{CREW_LIB_PREFIX}`), so that our newer glibc libraries won't be used when the system's glibc dynamic linker is used

uberhacker:
> > We would have to make sure all binaries are compatible with the glibc version installed or nothing may work. We are treading very sensitive ground here. Changing the glibc version would break all system binaries and unless there was a crew package equivalent installed, breakage will occur.
> 
> It should be fine as glibc is backward compatible. Anything compiled against an older glibc version should also work on newer glibc releases.
> 
> Also, I have configured it to put all of its libraries under `#{CREW_PREFIX}/opt/glibc-libs` (instead of `#{CREW_LIB_PREFIX}`), so that our newer glibc libraries won't be used when the system's glibc dynamic linker is used

If you can get glibc 2.37+ working on older ChromeOS images, that would be a huge win.

supechicken:
It doesn't build on i686 because of broken `gawk` :/

supechicken:
Latest version of glibc is running on i686!
```shell
chronos@574e933c28e4 ~ $ /usr/local/opt/glibc-libs/ld-linux.so.2 --version
ld.so (GNU libc) stable release version 2.41.
Copyright (C) 2025 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

chronos@574e933c28e4 ~ $ ./glibc 
GNU libc version: 2.23
chronos@574e933c28e4 ~ $ /usr/local/opt/glibc-libs/ld-linux.so.2 ./glibc 
GNU libc version: 2.41
```

supechicken:
Just realized that this might be the last missing piece of enabling aarch64 userspace support on supported devices (i.e ARM Chromebooks with 64-bit SoC and kernel). But the problem is we don't have an aarch64 compiler currently...

packages/glibc_standalone.rb
```diff
@@ -0,0 +1,101 @@
...
+
+class Glibc_standalone < Package
+  description 'The GNU C Library project provides the core libraries for GNU/Linux systems.'
+  homepage 'https://www.gnu.org/software/libc/'
+  version '2.41'
+  license 'LGPL-2.1+, BSD, HPND, ISC, inner-net, rc, and PCRE'
+  compatibility 'aarch64 armv7l x86_64'
+  source_url "https://ftpmirror.gnu.org/glibc/glibc-#{version}.tar.xz"
+  source_sha256 'a5a26b22f545d6b7d7b3dd828e11e428f24f4fac43c934fb071b6a7d0828e901'
+  binary_compression 'tar.zst'
```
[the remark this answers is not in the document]
supechicken: Binaries uploaded

packages/glibc_standalone.rb
satmandu: Oops. I misread that.

satmandu:
Does this, perhaps, fix building with `mold`? (You would probably have to run `autoreconf—fiv` as well in the build section.)
```
diff -Npaur a/configure.ac b/configure.ac
--- a/configure.ac	2025-04-24 11:06:13.148443082 -0400
+++ b/configure.ac	2025-04-24 11:11:43.340286165 -0400
@@ -540,6 +540,13 @@ case $($LD --version) in
 		    LD=: critic_missing="$critic_missing LLD")
     libc_cv_with_lld=yes
     ;;
+  "mold"*)
+  # Accept mold 2.37.1 or higher
+    AC_CHECK_PROG_VER(LD, $LD, --version,
+		    [mold.* \([0-9][0-9]*\.[0-9.]*\)],
+		    [2.37[1-9]*|2.3[8-9][0-9]*|2.[4-9][0-9]*],
+		    LD=: critic_missing="$critic_missing mold")
+    ;;
   *)
     AC_CHECK_PROG_VER(LD, $LD, --version,
 		    [GNU ld.* \([0-9][0-9]*\.[0-9.]*\)],
```
[mold_glibc_configure.patch.txt](https://github.com/user-attachments/files/19894731/mold_glibc_configure.patch.txt)

satmandu:
Do we need to immediately set `
```bash
if [ -d "${CREW_PREFIX}/opt/glibc-lib" ]; then
  export LD_LIBRARY_PATH=${CREW_PREFIX}/opt/glibc-libs:$LD_LIBRARY_PATH
fi
```
in `/usr/local/etc/profile` which is in `crew_profile_base`?

Or do we need to invoke a shell using our patched glibc immediately? Otherwise, will we get weird edge cases where a binary that doesn't have `LD_LIBRARY_PATH` set as we need it doesn't work right because it tries to use the system glibc, which is likely older?

satmandu:
I presume that anything built after this will automatically just work being built with the new glibc... so I'd like to get this all worked out before we do final builds of GCC 15.1 in https://github.com/chromebrew/chromebrew/pull/11807 , because then I can dispense with all this craziness with multiple builds for different glibc versions...

supechicken:
Modifying the version check does work (I used `sed -i s/GNU ld/mold/g` directly on `configure`), but `mold` still segfaults in the middle of compiling :/

---

> Do we need to immediately set
> 
> ```shell
> if [ -d "${CREW_PREFIX}/opt/glibc-lib" ]; then
>   export LD_LIBRARY_PATH=${CREW_PREFIX}/opt/glibc-libs:$LD_LIBRARY_PATH
> fi
> ```
> 
> in `/usr/local/etc/profile` which is in `crew_profile_base`?
> 
> Or do we need to invoke a shell using our patched glibc immediately? Otherwise, will we get weird edge cases where a binary that doesn't have `LD_LIBRARY_PATH` set as we need it doesn't work right because it tries to use the system glibc, which is likely older?

Not needed, our newer glibc should work out of the box. The only thing required is to run `patchelf --set-interpreter` on existing executables, telling them to use our glibc installation (all unmodified/system commands still run flawlessly with system glibc, see https://github.com/chromebrew/chromebrew/pull/11806#issuecomment-2824942741)

supechicken:
> I presume that anything built after this will automatically just work being built with the new glibc... so I'd like to get this all worked out before we do final builds of GCC 15.1 in #11807 , because then I can dispense with all this craziness with multiple builds for different glibc versions...

Do we have to delete all existing `glibc_*` packages now? Since most files between `glibc_*` and `glibc_standalone` have conflict.

satmandu:
> Not needed, our newer glibc should work out of the box. The only thing required is to run `patchelf --set-interpreter` on existing executables, telling them to use our glibc installation (all unmodified/system commands still run flawlessly with system glibc, see [#11806 (comment)](https://github.com/chromebrew/chromebrew/pull/11806#issuecomment-2824942741))

We should avoid using patchelf to hack in our interpreter on every binary. All sorts of weird stuff can happen. And this also causes breakage with our upx usage.

Once we have our glibc loaded, it should transparently prefix our glibc search paths automatically right?

We just need to make sure that `rpath` is not hardcoded in the app?

supechicken:
But changing the interpreter path is necessary... All `LD` related environment variables are being resolved **after** the interpreter initializes. And the interpreter version must match the glibc library going to load (i.e `ld-linux` from glibc 2.23 + `lib{c,m}.so.6` from glibc 2.41 will not work)

> We should avoid using patchelf to hack in our interpreter on every binary. All sorts of weird stuff can happen. And this also causes breakage with our upx usage.

It should be fine as only executables need to be patched, shared libraries don't require any modification. But `upx` is one problem...

supechicken:
I have given up on that since it cannot handle `upx` executable as well... Another solution would be adding `-Wl,--dynamic-linker` to `LDFLAGS` during build. That way `upx` still works and no further hacks are required.

satmandu:
@supechicken Please figure out what other files need to be modified for this approach to work... One merged, we don't want things to break.

supechicken:
> Please figure out what other files need to be modified for this approach to work... One merged, we don't want things to break.

To do that, we will have to:
- (For **autotool**) Add `LDFLAGS='-Wl,--dynamic-linker,#{CREW_GLIBC_PREFIX}/ld-linux.so'` to `CREW_ENV_OPTIONS`
- (For **meson**) Set `-D{c,cpp}_link_args=${LDFLAGS}` in `CREW_MESON_OPTIONS` in case `meson` ignores `LDFLAGS`
- (For **cmake**) Set `-DCMAKE_{EXE,SHARED}_LINKER_FLAGS` to `LINKER:--dynamic-linker,#{CREW_GLIBC_PREFIX}/ld-linux.so`

Alternatively, create a wrapper for `mold` to ensure the `--dynamic-linker` option is always appended.

For prebuilt sources (`no_compile_needed`), we will need some hacks (this also works all executables) :)

- Create a wrapper for it, which will re-exec the current executable using our dynamic linker
- Embed the original executable inside the wrapper, compile it
- Inside the wrapper, run the embed executable with `#{CREW_GLIBC_PREFIX}/ld-linux-x86-64.so.2 --argv0 executable /path/to/executable` (another way to involve our linker without setting the ELF interpreter path)
- Archive it with `upx`

supechicken:
Demonstration of the latter one:
```
$ gcc glibc.c -o glibc-checker
$ ./glibc-checker
GNU libc version: 2.37

# now embed it in the wrapper
$ cp glibc-checker executable
$ objcopy -I binary -O default executable executable.o
$ gcc -fPIC executable.o wrapper.c -fuse-ld=mold -o result

# compress it with upx
$ upx result 

# run it
$ ./result
[+] Now re-exec with ld-linux...
GNU libc version: 2.41
```

Source code of the wrapper can be found [here](https://gist.github.com/supechicken/4c26ae331b404ea309c2ebb3e0d93c02)

satmandu (approved):

PR #11817, opened 2025-04-25
crew: Set dynamic linker path in linker

### LINKED ISSUE #5777: Chromebrew sets and relies on LD_LIBRARY_PATH which breaks some system binaries
## Description and Analysis
The chromebrew binaries use some chromebrew-shipped libraries stored in /usr/local/lib64. In order for them to find their libraries chromebrew adds an **LD_LIBRARY_PATH** environment variable with just this path. This is bad because when the run-time dynamic linker sees that variable its contents override / take precedence over the system's previously configured library search path. When this variable is set and the user runs a system provided binary this causes those binaries to find and use some chromebrew version of libraries instead of the Chrome OS version. This breaks some system binaries and could potentially cause more serious system issues if the binary interacts with other system services.

For example **`/usr/bin/powerd_dbus_suspend`** is one such binary. When run with this environment variable set it uses a mix of system and chromebrew supplied libraries with fatal results:

```
root / # powerd_dbus_suspend 
2021-05-14T22:06:18.706380Z ERROR powerd_dbus_suspend: [bus.cc(393)] Failed to connect to the bus: Failed to connect to socket /usr/local/var/run/dbus/system_bus_socket: No such file or directory
2021-05-14T22:06:18.706430Z FATAL powerd_dbus_suspend: [powerd_dbus_suspend.cc(176)] Check failed: bus->Connect(). 
/usr/lib64/libbase-core.so(base::debug::CollectStackTrace(void**, unsigned long)+0x9) [0x7d26c6f69659]

Aborted (core dumped)
```

As we can see this system binary is pulling in several unintended libraries:

```
root / # ldd `which powerd_dbus_suspend`
        linux-vdso.so.1 (0x00007ffdab786000)
        libbrillo-core.so => /usr/lib64/libbrillo-core.so (0x00007a99a4633000)
        libbase-core.so => /usr/lib64/libbase-core.so (0x00007a99a44bb000)
        libbase-dl.so => /usr/lib64/libbase-dl.so (0x00007a99a44b5000)
        libbase-dbus.so => /usr/lib64/libbase-dbus.so (0x00007a99a4470000)
        libprotobuf-lite.so.22 => /usr/lib64/libprotobuf-lite.so.22 (0x00007a99a43e7000)
        libc++.so.1 => /usr/local/lib64/libc++.so.1 (0x00007a99a430f000)
        libc++abi.so.1 => /usr/local/lib64/libc++abi.so.1 (0x00007a99a42cb000)
        libc.so.6 => /usr/local/lib64/libc.so.6 (0x00007a99a4120000)
        libpthread.so.0 => /usr/local/lib64/libpthread.so.0 (0x00007a99a4100000)
        libcrypto.so.1.1 => /usr/local/lib64/libcrypto.so.1.1 (0x00007a99a3dc3000)
        libm.so.6 => /usr/local/lib64/libm.so.6 (0x00007a99a3c6a000)
        libdouble-conversion.so.3 => /usr/lib64/libdouble-conversion.so.3 (0x00007a99a3c57000)
        librt.so.1 => /usr/local/lib64/librt.so.1 (0x00007a99a3c4b000)
        libglib-2.0.so.0 => /usr/local/lib64/libglib-2.0.so.0 (0x00007a99a3b05000)
        libevent_core-2.1.so.6 => /usr/lib64/libevent_core-2.1.so.6 (0x00007a99a3adf000)
        /lib64/ld-linux-x86-64.so.2 => /usr/local/lib64/ld-linux-x86-64.so.2 (0x00007a99a46b9000)
        libdl.so.2 => /usr/local/lib64/libdl.so.2 (0x00007a99a3ad9000)
        libdbus-1.so.3 => /usr/local/lib64/libdbus-1.so.3 (0x00007a99a3a75000)
        libgcc_s.so.1 => /usr/local/lib64/libgcc_s.so.1 (0x00007a99a3a59000)
        libatomic.so.1 => /usr/local/lib64/libatomic.so.1 (0x00007a99a3a4f000)
        libpcre.so.1 => /usr/local/lib64/libpcre.so.1 (0x00007a99a39c5000)
```

One particular area of concern is the the standard C library (libc.so.6 _et al_.). Because Chrome OS runs a modified Linux kernel with some modified system calls its C library is also modified to take advantage of it. Unless the version of the C library shipped with chromebrew has the exact same patches this could potentially lead to system stability or security issues if system binaries are run in a library environment they don't expect.

Another area which at first glance would appear to be concerning is the run-time dynamic linker (ld-linux-x86-64.so.2). Because Chrome OS has a strong performance and security focus it seems likely to me that Google has modified the dynamic linker or will do so in the future. Running system binaries without the system-supplied dynamic linker could add performance issues or potential security vulnerabilities unnecessarily. However, I see on closer inspection that the path `/usr/local/lib64/ld-linux-x86-64.so.2` is a symlink that ultimately points to the system supplied dynamic linker at `/lib64/ld-2.27.so` so this is not an issue.

## Possible Solutions

Instead of using **LD_LIBRARY_PATH** as it currently does chromebrew could:

- Use the `-Wl,rpath,/usr/local/lib64` compiler linker option. (Or just the `-rpath` linker option if `ld` is run separately from the compiler which is unlikely.) This option causes the linker to hard-code the chromebrew library search path directly into the resulting ELF binary.

  This is the preferred and safest solution because it ensures that the library environment used by system binaries does not accidentally contain chromebrew-supplied libraries. I strongly recommend this option. A potential drawback of this option is that it means that all binary packages would need to be rebuilt and re-downloaded by users.

  Another advantage of this method is that should chromebrew ever want to ship `setuid` binaries this method will still work but **LD_LIBRARY_PATH** is ignored.

- Change **LD_LIBRARY_PATH** to add the chromebrew library path to the **end** of the systems standard library search path like this:

  ```export LD_LIBRARY_PATH="$(grep '^/' /etc/ld.so.conf | tr '\n' ':')/usr/local/lib64"```

  This has the advantage of being a small change that does not require rebuilding the binaries and re-downloading them by the users. A disadvantage is that it would still be possible for system binaries to accidentally use the chromebrew version of a library if the system version is missing or broken. Another disadvantage is that this does not work for `setuid` binaries.

chromebrew: 1.9.8

Model: Brunch-based Franken-Chromebook on upgraded LG Gram 17 platform (works awesomely).

ChromeOS:
```
chronos / $ cat /etc/lsb-release
CHROMEOS_ARC_ANDROID_SDK_VERSION=30
CHROMEOS_ARC_VERSION=7347114
CHROMEOS_BOARD_APPID={95EE134E-B47F-43FB-9835-32C276865F9A}
CHROMEOS_DEVSERVER=
CHROMEOS_RELEASE_APPID={95EE134E-B47F-43FB-9835-32C276865F9A}
CHROMEOS_RELEASE_BOARD=hatch-signed-mp-v6keys
CHROMEOS_RELEASE_BRANCH_NUMBER=31
CHROMEOS_RELEASE_BUILDER_PATH=hatch-release/R91-13904.31.0
CHROMEOS_RELEASE_BUILD_NUMBER=13904
CHROMEOS_RELEASE_CHROME_MILESTONE=91
CHROMEOS_RELEASE_DESCRIPTION=13904.31.0 (Official Build) beta-channel hatch 
CHROMEOS_RELEASE_KEYSET=mp-v6
CHROMEOS_RELEASE_TRACK=beta-channel
CHROMEOS_RELEASE_UNIBUILD=1
CHROMEOS_RELEASE_VERSION=13904.31.0
DEVICETYPE=CHROMEBOOK
GOOGLE_RELEASE=13904.31.0
```

Architecture:
```
chronos / $ uname -r -m
5.10.26-brunch-sebanc x86_64
```

## Additional information

Chromebrew is awesome. **You** are awesome. Thank you. Is there a way to donate?

supechicken:
Fix #5777 

_Part of #11806 [merged]_ 

_Depends on chromebrew/crew-profile-base#31 [merged]_

_Do not merge this before #11806 [merged] is ready_

```bash
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=set_dynamic_linker crew update \
&& yes | crew upgrade
```


lib/const.rb
```diff
@@ -202,13 +206,12 @@
...
 
 CREW_LINKER ||= ENV.fetch('CREW_LINKER', 'mold') unless defined?(CREW_LINKER)
-CREW_GLIBC_OVERRIDE_LINKER_FLAGS ||= ARCH == 'x86_64' && LIBC_VERSION.to_f >= 2.35 ? " #{File.join(CREW_LIB_PREFIX, 'libC.so.6')} " : ''
-CREW_LINKER_FLAGS ||= ENV.fetch('CREW_LINKER_FLAGS', CREW_GLIBC_OVERRIDE_LINKER_FLAGS) unless defined?(CREW_LINKER_FLAGS)
+CREW_LINKER_FLAGS ||= ENV.fetch('CREW_LINKER_FLAGS', '') unless defined?(CREW_LINKER_FLAGS)
 
 CREW_CORE_FLAGS           ||= "-O3 -pipe -ffat-lto-objects -fPIC #{CREW_ARCH_FLAGS} -fuse-ld=#{CREW_LINKER} #{CREW_LINKER_FLAGS}"
 CREW_COMMON_FLAGS         ||= "#{CREW_CORE_FLAGS} -flto=auto"
 CREW_COMMON_FNO_LTO_FLAGS ||= "#{CREW_CORE_FLAGS} -fno-lto"
-CREW_LDFLAGS              ||= "-flto=auto #{CREW_LINKER_FLAGS}"
```
satmandu, line 211: Is the `-flto=auto` omission intentional?

lib/const.rb
```diff
@@ -202,13 +206,12 @@
...
+CREW_LDFLAGS              ||= "#{CREW_LINKER_FLAGS} -Wl,--dynamic-linker,#{CREW_GLIBC_INTERPRETER}"
```
uberhacker, line 214: Trailing `,` if `CREW_GLIBC_INTERPRETER` is nil?

satmandu:
Also, you should probably adjust the logic for `CREW_GLIBC_OVERRIDE_LINKER_FLAGS` in this PR to be set only for `2.35 < GLIBC < 2.41`...

lib/const.rb
supechicken: I think I deleted it accidentally...

lib/const.rb
```diff
@@ -76,17 +76,21 @@
...
+CREW_DEST_WINE_PREFIX ||= File.join(CREW_DEST_PREFIX, CREW_WINE_PREFIX)
 
 # Local constants for contributors.
 CREW_LOCAL_REPO_ROOT ||= `git rev-parse --show-toplevel 2> /dev/null`.chomp
 CREW_LOCAL_BUILD_DIR ||= "#{CREW_LOCAL_REPO_ROOT}/release/#{ARCH}"
 CREW_GITLAB_PKG_REPO ||= 'https://gitlab.com/api/v4/projects/26210301/packages'
 
+# Glibc related constants
+CREW_GLIBC_PREFIX      ||= File.join(CREW_LIB_PREFIX, 'opt/glibc-libs')
+CREW_GLIBC_INTERPRETER ||= File.symlink?("#{CREW_PREFIX}/bin/ld.so") ? File.readlink("#{CREW_PREFIX}/bin/ld.so") : nil
```
satmandu, line 92: This doesn't work, as it doesn't give an absolute path to `CREW_GLIBC_INTERPRETER`:

```ruby
CREW_LINKER_FLAGS = 'DUMMY_LINKER_FLAGS'
CREW_PREFIX = '/usr/local'
CREW_GLIBC_INTERPRETER ||= File.symlink?("#{CREW_PREFIX}/bin/ld.so") ? File.readlink("#{CREW_PREFIX}/bin/ld.so") : nil
CREW_LDFLAGS ||= "#{CREW_LINKER_FLAGS} #{CREW_GLIBC_INTERPRETER.nil? ? '' : "-Wl,--dynamic-linker,#{CREW_GLIBC_INTERPRETER}"}"
puts "CREW_GLIBC_INTERPRETER: #{CREW_GLIBC_INTERPRETER}"
puts "CREW_LDFLAGS: #{CREW_LDFLAGS}"
```

gives:
```
chronos@buppie-x86_64:2.27 M90 ~ $ ruby test.rb
CREW_GLIBC_INTERPRETER: ../opt/glibc-libs/ld-linux-x86-64.so.2
CREW_LDFLAGS: DUMMY_LINKER_FLAGS -Wl,--dynamic-linker,../opt/glibc-libs/ld-linux-x86-64.so.2
```

You want to use `File.realpath`:
```ruby
CREW_LINKER_FLAGS = 'DUMMY_LINKER_FLAGS'
CREW_PREFIX = '/usr/local'
CREW_GLIBC_INTERPRETER ||= File.symlink?("#{CREW_PREFIX}/bin/ld.so") ? File.realpath("#{CREW_PREFIX}/bin/ld.so") : nil
CREW_LDFLAGS ||= "#{CREW_LINKER_FLAGS} #{CREW_GLIBC_INTERPRETER.nil? ? '' : "-Wl,--dynamic-linker,#{CREW_GLIBC_INTERPRETER}"}"
puts "CREW_GLIBC_INTERPRETER: #{CREW_GLIBC_INTERPRETER}"
puts "CREW_LDFLAGS: #{CREW_LDFLAGS}"
```

```
chronos@buppie-x86_64:2.27 M90 ~ $ ruby test.rb
CREW_GLIBC_INTERPRETER: /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2
CREW_LDFLAGS: DUMMY_LINKER_FLAGS -Wl,--dynamic-linker,/usr/local/opt/glibc-libs/ld-linux-x86-64.so.2
```

lib/const.rb
satmandu: You want this:
```ruby
CREW_LDFLAGS ||= "#{CREW_LINKER_FLAGS} #{CREW_GLIBC_INTERPRETER.nil? ? '' : "-Wl,--dynamic-linker,#{CREW_GLIBC_INTERPRETER}"}"
```

supechicken:
> Also, you should probably adjust the logic for `CREW_GLIBC_OVERRIDE_LINKER_FLAGS` in this PR to be set only for `2.35 < GLIBC < 2.41`...

But it should work on every system regardless of the glibc version

lib/const.rb

supechicken:
@satmandu I have merged `CREW_LDFLAGS` and `CREW_LINKER_FLAGS` to avoid confusion

satmandu:
@supechicken You accidentally added a bunch of bsdgames stuff...

satmandu:
Please move that to another PR, and rebase? sorry I just rebased for you...

satmandu:
Otherwise, I think this is good... we should just check to see how this works with builds and then make sure nothing breaks!

supechicken:
> Otherwise, I think this is good... we should just check to see how this works with builds and then make sure nothing breaks!

After this is merged, we will have to rebuild Ruby first to fix #11818

supechicken:
Now moving bsdgames stuff to a separate branch...

lib/const.rb
```diff
@@ -76,17 +76,21 @@
...
+CREW_WINE_PREFIX      ||= File.join(CREW_LIB_PREFIX, 'wine')
+CREW_DEST_WINE_PREFIX ||= File.join(CREW_DEST_PREFIX, CREW_WINE_PREFIX)
 
 # Local constants for contributors.
 CREW_LOCAL_REPO_ROOT ||= `git rev-parse --show-toplevel 2> /dev/null`.chomp
 CREW_LOCAL_BUILD_DIR ||= "#{CREW_LOCAL_REPO_ROOT}/release/#{ARCH}"
 CREW_GITLAB_PKG_REPO ||= 'https://gitlab.com/api/v4/projects/26210301/packages'
 
+# Glibc related constants
+CREW_GLIBC_PREFIX      ||= File.join(CREW_LIB_PREFIX, 'opt/glibc-libs')
```
satmandu, line 91: This needs to be : 
```ruby
CREW_GLIBC_PREFIX ||= File.join(CREW_PREFIX, 'opt/glibc-libs')
```

satmandu:
To be addressed before merge, tested in the hatch M134 container...

After running `CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=glibc_bootstrap crew update \
&& yes | crew upgrade`

System bash doesn't load with glibc_standalone.
```
chronos@buppie-x86_64:2.37 M134 /usr/local/lib/crew/packages (master >|SPARSE)$ /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2 /bin/bash
/bin/bash: error while loading shared libraries: /bin/bash: DT_RELR without GLIBC_ABI_DT_RELR dependency
```
glibc_standalone isn't playing well with existing upx compressed binaries.
```
chronos@buppie-x86_64:2.37 M134 /usr/local/lib/crew/packages (master >|SPARSE)$ /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2 /usr/local/bin/bash
/usr/local/bin/bash: error while loading shared libraries: /usr/local/bin/bash: object file has no dynamic section
```
Appending the `LD_LIBRARY_PATH` doesn't help.
```
chronos@buppie-x86_64:2.37 M134 /usr/local/lib/crew/packages (master >|SPARSE)$ LD_LIBRARY_PATH=/usr/local/opt/glibc-libs:$LD_LIBRARY_PATH /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2 /bin/bas
h
/bin/bash: error while loading shared libraries: /bin/bash: DT_RELR without GLIBC_ABI_DT_RELR dependency
chronos@buppie-x86_64:2.37 M134 /usr/local/lib/crew/packages (master >|SPARSE)$ which bash
/usr/local/bin/bash
chronos@buppie-x86_64:2.37 M134 /usr/local/lib/crew/packages (master >|SPARSE)$ LD_LIBRARY_PATH=/usr/local/opt/glibc-libs:$LD_LIBRARY_PATH /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2 /usr/loc
al/bin/bash
/usr/local/bin/bash: error while loading shared libraries: /usr/local/bin/bash: object file has no dynamic section
```
Decompressing our `bash` with `upx -d` gets us a little further...
```
chronos@buppie-x86_64:2.37 M134 /usr/local/lib/crew/packages (master >|SPARSE)$ upx -d `which bash`
                       Ultimate Packer for eXecutables
                          Copyright (C) 1996 - 2025
UPX 5.0.0       Markus Oberhumer, Laszlo Molnar & John Reiser   Feb 20th 2025

        File size         Ratio      Format      Name
   --------------------   ------   -----------   -----------
   1143520 <-    432572   37.83%   linux/amd64   bash

Unpacked 1 file.
chronos@buppie-x86_64:2.37 M134 /usr/local/lib/crew/packages (master >|SPARSE)$ LD_LIBRARY_PATH=/usr/local/opt/glibc-libs:$LD_LIBRARY_PATH /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2 /usr/local/bin/bash
bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8): No such file or directory
.🐣🐚rc
Segmentation fault (core dumped)
```

lib/const.rb
uberhacker: Still  deleted?

satmandu:
Please see the patches in: https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/refs/heads/release-R136-16238.B/sys-libs/glibc/files/local/glibc-2.39

Maybe we need to use https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/refs/heads/release-R136-16238.B/sys-libs/glibc/files/local/glibc-2.39/0009-Revert-Add-GLIBC_ABI_DT_RELR-for-DT_RELR-support.patch since I assume that many scripts will call `/bin/bash` which will break?

lib/const.rb
uberhacker: You could also take out the `opt/`.

satmandu:
Just the 009 patch is insufficient, but it does get us further. upx still breaks things.

Maybe https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/refs/heads/release-R136-16238.B/sys-libs/glibc/files/local/glibc-2.39/0004-Add-dlopen_with_offset-to-glibc.patch?pli=1%2F is needed? Though that might conflict with your patch...

satmandu:
We need to make sure that nothing breaks on upgrades... and I'm pretty sure that means making sure that at least `/bin/bash` doesn't break... Worst case scenario you might have to hack up glibc to replace `/bin/bash` and `/bin/sh` with calls to our own bash...

In any case, currently trying to build this to get around the previous errors from `/usr/local/opt/glibc-libs/ld-linux-x86-64.so.2 /bin/bash`.
```
require 'package'

class Glibc_standalone < Package
  version '2.41-1'
  license 'LGPL-2.1+, BSD, HPND, ISC, inner-net, rc, and PCRE'
  compatibility 'all'
  source_url "https://ftpmirror.gnu.org/glibc/glibc-#{version.split('-').first}.tar.xz"
  source_sha256 'a5a26b22f545d6b7d7b3dd828e11e428f24f4fac43c934fb071b6a7d0828e901'
  binary_compression 'tar.zst'

  binary_sha256({
    aarch64: '4d34f1aea3003156670485deecf5da4ff3f377de5260082edd1e27e3e8aa907e',
     armv7l: '4d34f1aea3003156670485deecf5da4ff3f377de5260082edd1e27e3e8aa907e',
       i686: '77a95bfa9bdf58c02e9534645eb45886a8bc0a0043d4ba0bbf39bd862bd63aeb',
     x86_64: '595548c1f923322526925698a4b2fec064112f78878375388fb5889583619f2f'
  })

  depends_on 'gawk' => :build
  depends_on 'filecmd' # L Fixes creating symlinks on a fresh install.
  depends_on 'libidn2' => :build
  depends_on 'llvm' => :build
  depends_on 'texinfo' => :build
  depends_on 'libxcrypt' # Latest glibc removed libcrypt.so, add this for backward compatibility
  depends_on 'patchelf' # L
  depends_on 'glibc' # R

  conflicts_ok
  no_env_options
  no_shrink

  def self.patch
    File.write 'hardcode-library-path.patch', <<~EOF
      diff -Nur a/elf/dl-support.c b/elf/dl-support.c
      --- a/elf/dl-support.c    2025-04-23 01:58:17.265974714 +0800
      +++ b/elf/dl-support.c    2025-04-23 14:38:27.373033131 +0800
      @@ -296,7 +296,7 @@

         /* Initialize the data structures for the search paths for shared
            objects.  */
      -  _dl_init_paths (getenv ("LD_LIBRARY_PATH"), "LD_LIBRARY_PATH",
      +  _dl_init_paths ("#{CREW_PREFIX}/opt/glibc-libs:#{CREW_LIB_PREFIX}:/#{ARCH_LIB}", "LD_LIBRARY_PATH",
       \t\t  /* No glibc-hwcaps selection support in statically
       \t\t     linked binaries.  */
       \t\t  NULL, NULL);
      diff -Nur a/elf/rtld.c b/elf/rtld.c
      --- a/elf/rtld.c    2025-04-23 04:04:02.230377659 +0800
      +++ b/elf/rtld.c    2025-04-23 14:37:52.298031258 +0800
      @@ -2647,7 +2647,7 @@
       \t  /* The library search path.  */
       \t  if (memcmp (envline, "LIBRARY_PATH", 12) == 0)
       \t    {
      -\t      state->library_path = &envline[13];
      +\t      state->library_path = "#{CREW_PREFIX}/opt/glibc-libs:#{CREW_LIB_PREFIX}:/#{ARCH_LIB}";
       \t      state->library_path_source = "LD_LIBRARY_PATH";
       \t      break;
       \t    }
    EOF

    system 'filefix'
    system 'patch', '-p1', '-i', 'hardcode-library-path.patch'
    # See https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/refs/heads/release-R136-16238.B/sys-libs/glibc/files/local/glibc-2.39/0009-Revert-Add-GLIBC_ABI_DT_RELR-for-DT_RELR-support.patch?pli=1%2F
    File.write '0009-Revert-Add-GLIBC_ABI_DT_RELR-for-DT_RELR-support.patch', <<~GLIBC_ABI_DT_RELR_PATCH_EOF
      diff --git a/elf/dl-version.c b/elf/dl-version.c
      index 8966d612cc79..7fb5cb16d47b 100644
      --- a/elf/dl-version.c
      +++ b/elf/dl-version.c
      @@ -359,30 +359,6 @@ _dl_check_map_versions (struct link_map *map, int verbose, int trace_mode)
       \t}
           }

      -  /* When there is a DT_VERNEED entry with libc.so on DT_NEEDED, issue
      -     an error if there is a DT_RELR entry without GLIBC_ABI_DT_RELR
      -     dependency.  */
      -  if (dyn != NULL
      -      && map->l_info[DT_NEEDED] != NULL
      -      && map->l_info[DT_RELR] != NULL
      -      && __glibc_unlikely (!map->l_dt_relr_ref))
      -    {
      -      const char *strtab = (const void *) D_PTR (map, l_info[DT_STRTAB]);
      -      const ElfW(Dyn) *d;
      -      for (d = map->l_ld; d->d_tag != DT_NULL; ++d)
      -\tif (d->d_tag == DT_NEEDED)
      -\t  {
      -\t    const char *name = strtab + d->d_un.d_val;
      -\t    if (strncmp (name, "libc.so.", 8) == 0)
      -\t      {
      -\t\t_dl_exception_create
      -\t\t  (&exception, DSO_FILENAME (map->l_name),
      -\t\t   N_("DT_RELR without GLIBC_ABI_DT_RELR dependency"));
      -\t\tgoto call_error;
      -\t      }
      -\t  }
      -    }
      -
         return result;
       }

      --
      2.41.0
    GLIBC_ABI_DT_RELR_PATCH_EOF
    system 'patch', '-p1', '-i', '0009-Revert-Add-GLIBC_ABI_DT_RELR-for-DT_RELR-support.patch'
  end

  def self.build
    glibc_libdir = File.join(CREW_PREFIX, 'opt/glibc-libs')

    build_env = {
      CFLAGS:   '-O3 -pipe -fPIC -fno-lto -fuse-ld=lld',
      CXXFLAGS: '-O3 -pipe -fPIC -fno-lto -fuse-ld=lld',
      LD:       'ld.lld' # use lld here as mold will segfault
    }

    config_opts = %W[
      --prefix=#{CREW_PREFIX}
      --libdir=#{glibc_libdir}
      --libexecdir=#{CREW_PREFIX}/libexec
      --mandir=#{CREW_PREFIX}/man
      --with-headers=#{CREW_PREFIX}/include
      --with-bugurl=https://github.com/chromebrew/chromebrew/issues/new
      --enable-bind-now
      --enable-fortify-source
      --enable-kernel=#{ARCH.eql?('i686') ? '3.2' : '4.4'}
      --disable-nscd
      --disable-profile
      --disable-sanity-checks
      --disable-werror
      --without-cvs
      --without-selinux
    ]

    config_opts << '--enable-cet' unless ARCH == 'i686'

    FileUtils.mkdir_p 'builddir'
    Dir.chdir('builddir') do
      File.write 'configparms', <<~EOF
        slibdir=#{glibc_libdir}
        rtlddir=#{glibc_libdir}
        sbindir=#{CREW_PREFIX}/bin
        rootsbindir=#{CREW_PREFIX}/bin
      EOF

      system build_env.transform_keys(&:to_s), '../configure', *config_opts
      system 'make', "PARALLELMFLAGS='-j #{CREW_NPROC}'"
    end
  end

  def self.install
    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install', chdir: 'builddir'
    system "make -j1 DESTDIR=#{CREW_DEST_DIR} localedata/install-locales || make -j1 DESTDIR=#{CREW_DEST_DIR} localedata/install-locales", chdir: 'builddir'
  end
end
```

lib/const.rb
supechicken: Re-added in the next line

supechicken:
> Just the 009 patch is insufficient, but it does get us further. upx still breaks things.
> 
> Maybe https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/refs/heads/release-R136-16238.B/sys-libs/glibc/files/local/glibc-2.39/0004-Add-dlopen_with_offset-to-glibc.patch?pli=1%2F is needed? Though that might conflict with your patch...

`ld-linux-x86-64.so <upx packed program>` will not work. Anything after `upx` will become a "static" executable (`ld-linux` is executed after `upx` bootstrap, so it is impossible to run it with `ld-linux` directly):
```shell
$ file /usr/local/bin/bash
/usr/local/bin/bash: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), statically linked, no section header

$ upx -d /usr/local/bin/bash
$ file /usr/local/bin/bash
/usr/local/bin/bash: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, stripped

**Comment by satmandu** [comment]
So without upx, we do have this once we use the 009 patch and install locales:
```
chronos@buppie-x86_64:2.27 M90 /output/git/chromebrew (glibc_standalone_rebuild *$)$ /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2 /bin/bash -c "echo hello"
hello
```

File: lib/const.rb (hunk 4 above)
Comment (supechicken) [reply lib/const.rb]: Fixed

Comment (satmandu) [reply lib/const.rb]: That would require rebuilds of glibc_standalone and also trip the warnings for FHS compatibility.

**Comment by satmandu** [comment]
FYI https://github.com/chromebrew/chromebrew/pull/11828 needs to be merged before this...

**Comment by satmandu** [comment]
> 
> `ld-linux-x86-64.so <upx packed program>` will not work. Anything after `upx` will become a "static" executable (`ld-linux` is executed (by `upx`, instead of the kernel) after `upx` bootstraps itself, so it is impossible to run it with `ld-linux` directly):
> 
> ```shell
> $ file /usr/local/bin/bash
> /usr/local/bin/bash: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), statically linked, no section header
> 
> $ upx -d /usr/local/bin/bash
> $ file /usr/local/bin/bash
> /usr/local/bin/bash: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, stripped
> ```
Ok what's the best solution here?

Do we just stop using `upx`, and run a script to uncompress every binary in every package using `upx -d` and then reupload it?  (And do we also while doing that run patchelf on the binaries?)

Futher hacking things up with upx seems to be a level of unnecessary complexity.

**Comment by satmandu** [comment]
@uberhacker @supechicken @Zopolis4 Thoughts?

**Comment by satmandu** [comment]
Also, we should hold off on building `mold` 2.38 until we know if we are going to be patching it to automatically add the library path for new builds.

**Comment by supechicken** [comment]
> Ok what's the best solution here?
> 
> Do we just stop using upx, and run a script to uncompress every binary in every package using upx -d and then reupload it? (And do we also while doing that run patchelf on the binaries?)

See my previous comment in https://github.com/chromebrew/chromebrew/pull/11806#issuecomment-2829758380, I have created a wrapper program for re-executing the program with our interpreter once `upx` completes bootstraping

**Comment by satmandu** [comment]
> > Ok what's the best solution here?
> > Do we just stop using upx, and run a script to uncompress every binary in every package using upx -d and then reupload it? (And do we also while doing that run patchelf on the binaries?)
> 
> See my previous comment in [#11806 (comment)](https://github.com/chromebrew/chromebrew/pull/11806#issuecomment-2829758380), I have created a wrapper program for re-executing the program with our interpreter once `upx` completes bootstraping

Hmm... Is there a way to do it without modifying the binaries to use another wrapper script besides `upx`? Can we use `LD_PRELOAD`?

File: lib/const.rb (hunk 1 above)
Comment (uberhacker) [reply lib/const.rb:211]: I don't see where it was re-added.  Are we removing the `CREW_LDFLAGS` constant?

**Comment by supechicken** [comment]
Hmm... Is there a way to do it without modifying the binaries to use another wrapper script besides upx? Can we use LD_PRELOAD?

I have tried it before :) But I soon gave it up since there is no way to get the decompressed code after `upx` initializes. (We can't use `ld-linux` directly on upx compressed programs, as mentioned above)

**Comment by satmandu** [comment]
@supechicken I should have binaries uploaded for glibc built with the 009 patch and locales generated shortly.

**Comment by satmandu** [comment]
Maybe we just abandon upx? Adding another wrapper on top of a wrapper seems like a good way to slow things down and have unintended issues?

**Comment by supechicken** [comment]
> @supechicken I should have binaries uploaded for glibc built with the 009 patch and locales generated shortly.

I have included another patch from ChromiumOS that speeds up glibc on ARM platforms in my previous commit. Please include it as well😅

**Comment by supechicken** [comment]
> Maybe we just abandon upx? Adding another wrapper on top of a wrapper seems like a good way to slow things down and have unintended issues?

I am on the same page with you :)

**Comment by supechicken** [comment]
Do we need to use `tar.xz` for prebuilt binaries? Since this will become one of the bootstrap packages

**Comment by satmandu** [comment]
I'm updating gawk, and the update failed to build when glibc_standalone was installed. So there might be a chicken and egg issue here.

(Also I was having a memory exhaustion issue with `gawk` when trying to build on i686, but maybe I just need to reboot my server...)

**Comment by satmandu** [comment]
gawk newer than 5.3.0 is broken on i686.

Changes here:

in io.c:
```

/*
 * See the thread starting at
 * https://lists.gnu.org/archive/html/bug-gawk/2023-12/msg00011.html.
 *
 * We do our version of popen for write pipes in order
 * to be able to reset SIGPIPE. Bleah.
 */
```

**Comment by satmandu** [comment]
forcibly downgraded and trying to build locally...

**Comment by satmandu** [comment]
Looks like that build is working... But we need to patch gawk for i686 to revert whatever changed popen functionality back in December 2023, since that is breaking pipe usage on i686.

Weirdly, starting the build on the arm host in the i686 container was working, so this could be either a kernel or docker issue on my end...

**Comment by satmandu** [comment]
FYI this is the build command that caused the error:
```bash
LC_ALL=C gawk -f scripts/versionlist.awk /usr/local/tmp/crew/glibc_standalone.20250428004517.dir/glibc-2.41/builddir/Versions.v > /usr/local/tmp/crew/glibc_standalone.20250428004517.dir/glibc-2.41/builddir/Versions.defT
gawk: scripts/versionlist.awk:58: (FILENAME=/usr/local/tmp/crew/glibc_standalone.20250428004517.dir/glibc-2.41/builddir/Versions.v FNR=3737) fatal: io.c:4625:gawk_popen_write: open_pipes: cannot allocate -64 bytes of memory: Cannot allocate memory
```

Zopolis4:
> @uberhacker @supechicken @Zopolis4 Thoughts?

Still in the middle of exams so won't be much help, and i'm not knowledgeable enough on how glibc operates to be able to provide a particularly relevant opinion. 

From a quick look at this PR and the other related ones (correct me if im wrong at any point in this), my understanding is that this will let us use the latest glibc on all chromeos versions, which is great, while also allowing all the system libraries to work? However, in order to do this, we need to hardcode the interpreter path (either when building via arguments to the compilers/buildsystems, or via patchelf after things have been built (i think this also works for precompiled binaries?)).

Of course, this whole situation with our glibc vs system glibc is always going to be a little messy, but I think this new approach is better.

Just to check my understanding:
Benefits:
- Can use latest glibc on all systems
  - If all packages are rebuilt, allows us to not have to worry about `min_glibc`?
Disadvantages:
- No more `upx`?
- Additional required arguments when building


As an aside, I still see musl as the eventual solution, but that doesn't mean we can't improve our glibc handling until then.

supechicken:
> or via patchelf after things have been built (i think this also works for precompiled binaries?)

Yep, it works on all dynamically linked binaries.

> As an aside, I still see musl as the eventual solution, but that doesn't mean we can't improve our glibc handling until then.

We will need to go through the same steps if we want to use `musl` as our primary C runtime library (changing the interpreter path on existing binaries, specifying interpreter path during compilation, putting musl libraries elsewhere besides `CREW_LIB_PREFIX` to prevent conflicts)

Compiling all binaries statically would be another solution, but it would take a ton of extra storage spaces.

satmandu:
Some stuff just doesn't work with musl, especially the GUI stuff. Remember that I tried adding musl several years ago... Some of that work is still in Chromebrew...

Zopolis4:
I've been doing my own experiments with musl as well. There are a few distros that only use musl, and support has gotten better over the past several years. It's still a pretty distant goal of mine, though.

supechicken:
Succeed with building `gawk` and `htop`. (No idea why the minimum kernel version is set to `5.10`, maybe `glibc` needs rebuilding with a lower `--enable-kernel` option)
```
$ file /usr/local/bin/htop
/usr/local/bin/htop: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2, for GNU/Linux 5.10.0, stripped

$ file /usr/local/bin/gawk
/usr/local/bin/gawk: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2, for GNU/Linux 5.10.0, stripped
```

supechicken:
We need to find a way to block `libC.so.6` from loading when using our interpreter, as it makes some binaries segfault

uberhacker:
> We need to find a way to block `libC.so.6` from loading when using our interpreter, as it makes some binaries segfault
> 
> EDIT: Now figuring out how to create an [audit library](https://man7.org/linux/man-pages/man7/rtld-audit.7.html) for intercepting and ignoring it...

Include the directory containing the new libC.so.6 first in `LD_LIBRARY_PATH`?

supechicken:
> Include the directory containing the new libC.so.6 first in LD_LIBRARY_PATH?

`glibc` does not respect `LD_LIBRARY_PATH` in this case:
```
chronos@brunch /usr/local/opt/glibc-libs $ ls libC.so.6 
libC.so.6
chronos@brunch /usr/local/opt/glibc-libs $ LD_LIBRARY_PATH=/usr/local/opt/glibc-libs:/usr/local/lib64 LD_TRACE_LOADED_OBJECTS=1 /usr/local/bin/gcc
        linux-vdso.so.1 (0x00007c7563e4c000)
        libm.so.6 => /usr/local/lib64/libm.so.6 (0x00007c7563d5c000)
        libc.so.6 => /usr/local/lib64/libc.so.6 (0x00007c7563b8c000)
        /lib64/ld-linux-x86-64.so.2 (0x00007c7563e4e000)
        libC.so.6 => /usr/local/lib64/libC.so.6 (0x00007c7563993000)
```

supechicken:
> EDIT: Now figuring out how to create an [audit library](https://man7.org/linux/man-pages/man7/rtld-audit.7.html) for intercepting and ignoring it...

Just completed and it works well :) ([source code here](https://gist.github.com/supechicken/54839c560fac14e22e1dfaca9c6f4586))
```
 $ LD_AUDIT=/tmp/audit /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2 /usr/local/bin/mold
crew-audit: Initialized on glibc 2.41 with interface version 2
crew-audit: Chromebrew prefix: /usr/local
crew-audit: Library found in /usr/local/opt/glibc-libs/libdl.so.2, using it instead...
crew-audit: Library found in /usr/local/opt/glibc-libs/libpthread.so.0, using it instead...
crew-audit: Library found in /usr/local/opt/glibc-libs/librt.so.1, using it instead...
crew-audit: Library found in /usr/local/opt/glibc-libs/libm.so.6, using it instead...
crew-audit: Library found in /usr/local/opt/glibc-libs/libc.so.6, using it instead...
crew-audit: libC.so.6 being requested, replacing it with /usr/local/opt/glibc-libs/libc.so.6...
mold: fatal: -m option is missing
```

supechicken:
## Brief description of the new changes

- All patches are moved to [chromebrew/crew-package-glibc](https://github.com/chromebrew/crew-glibc-patches)
- Minimum kernel version is set to 3.2.0 for all architectures
- A [glibc audit library](https://github.com/chromebrew/crew-package-glibc/blob/main/crew-audit.c) is added, mainly used for blocking `libC.so.6` from load and fixing system commands (#5777)
  - All libraries from `CREW_LIB_PREFIX` will NOT be loaded when running system commands:
  ```
  $ LD_TRACE_LOADED_OBJECTS=1 CREW_AUDIT_VERBOSE=1 /bin/ls 
  crew-audit: Initialized on glibc 2.37 with interface version 2
  crew-audit: Running with ChromeOS's dynamic linker...
  crew-audit: System command detected, libraries provided by Chromebrew will not be used.
  crew-audit: Library libselinux.so.1 is being requested...
  ...snip...
  ```

- `LD_AUDIT` will be set to `${CREW_GLIBC_PREFIX}/crew-audit.so` during shell startup

Tested with building `htop`, it is working well so far :)
```
$ file /usr/local/bin/htop
/usr/local/bin/htop: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /usr/local/opt/glibc-libs/ld-linux-x86-64.so.2, for GNU/Linux 3.2.0, stripped
```

packages/glibc_standalone.rb
```diff
@@ -198,19 +79,28 @@ def self.build
...
+        rtlddir=#{CREW_GLIBC_PREFIX}
         sbindir=#{CREW_PREFIX}/bin
         rootsbindir=#{CREW_PREFIX}/bin
       EOF
 
       system build_env.transform_keys(&:to_s), '../configure', *config_opts
       system "make PARALLELMFLAGS='-j #{CREW_NPROC}'"
     end
+
+    # compile crew-audit with system's glibc version, as we want it to work on system's glibc also
```
satmandu, line 92: This means we need to continue shipping a Glibc package (with Glibc headers) for the system Glibc on the machine?

supechicken: No, the older glibc package is only used to build the audit library. This ensures the compatibility with system's glibc runtime library (`/lib64/libc.so.6`)

satmandu: Right, but do we need to build a Glibc 2.39 package to build an audit library for new milestones running Glibc 2.39?

satmandu:
Would it make sense to pull https://github.com/chromebrew/chromebrew/pull/11828 into here?

satmandu:
Hmmm:
```
 CREW_REPO=https://github
.com/supechicken/chromebrew.git CREW_BRANCH=set_dynamic_linker crew update && yes | crew upgrade
From https://github.com/supechicken/chromebrew
 * branch                set_dynamic_linker -> FETCH_HEAD
HEAD is now at fe9efde79 bump version
/usr/local/lib/crew/lib/const.rb:92: warning: already initialized constant CREW_GLIBC_INTERPRETER
/usr/local/lib/crew/lib/const.rb:92: warning: previous definition of CREW_GLIBC_INTERPRETER was here
```

satmandu:
Fixed.
```
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=
set_dynamic_linker crew update && yes | crew upgrade
remote: Enumerating objects: 7, done.
remote: Counting objects: 100% (7/7), done.
remote: Compressing objects: 100% (1/1), done.
remote: Total 4 (delta 3), reused 4 (delta 3), pack-reused 0 (from 0)
Unpacking objects: 100% (4/4), 609 bytes | 609.00 KiB/s, done.
From https://github.com/supechicken/chromebrew
 * branch                set_dynamic_linker -> FETCH_HEAD
HEAD is now at 935dea71f Fix aleady initialized constant error.
Restarting crew update since there is an updated crew version.
From https://github.com/supechicken/chromebrew
 * branch                set_dynamic_linker -> FETCH_HEAD
HEAD is now at 935dea71f Fix aleady initialized constant error.
Your software is up to date.
Your software is already up to date.
```

bin/crew
```diff
@@ -883,71 +883,30 @@ def prepare_package(destdir)
...
-          puts "#{execfiletopatch} needed library #{neededlib_basename} not found in #{patchelf_lib_prefix} or #{Dir.pwd + neededlibspatchednamepath}.".lightred
-        end
-      end
-      # Do not set interpreter for non-musl, as this can break apps if there
-      # is an issue with the crew glibc.
-      next unless @pkg.is_musl?
-
       puts 'Running patchelf to patch binary interpreter paths'.lightblue
-      system "patchelf --set-interpreter #{patchelf_interpreter} #{execfiletopatch}"
+      system "upx -d #{execfiletopatch} > /dev/null 2> /dev/null"
```
satmandu, line 906: Use this:
```
system "upx -qq -d #{execfiletopatch} > /dev/null 2> /dev/null"
```

install.sh
```diff
@@ -323,8 +305,30 @@ function extract_install () {
...
+
+      # setup ld.so.conf and update cache
+      cp /etc/ld.so.conf "${CREW_PREFIX}/etc/ld.so.conf"
+      echo "${CREW_PREFIX}/opt/glibc-libs" >> "${CREW_PREFIX}/etc/ld.so.conf"
+      ldconfig
+    else
+      # decompress and switch to glibc_standalone for existing binaries
+      if command -v upx &> /dev/null; then
+        echo_intra "Running upx on ${1}..."
+        grep "/usr/local/\(bin\|lib\|lib${LIB_SUFFIX}\)" < filelist | xargs -P "$(nproc)" -n1 upx -d 2> /dev/null || true
```
satmandu, line 323: This will probably be less verbose:
```bash
grep "/usr/local/\(bin\|lib\|lib${LIB_SUFFIX}\)" < filelist | xargs -P "$(nproc)" -n1 upx -qq -d 2> /dev/null || true
```

bin/crew
satmandu:
```
chronos@buppie-x86_64 /usr/local/lib/crew/packages $ upx -qq -d `which bash`
   1143520 <-    432572   37.83%   linux/amd64   bash                                                   chronos@buppie-x86_64 /usr/local/lib/crew/packages $ upx -qq -d `which bash`
chronos@buppie-x86_64 /usr/local/lib/crew/packages $
```

install.sh
supechicken: Change made

packages/glibc_standalone.rb
```diff
@@ -29,162 +29,44 @@ class Glibc_standalone < Package
...
-       \t/* Magic dust alert!  Force VFP on Cortex-A9.  Experiments show
-       \t   that the FP pipeline is much better at streaming loads and
-      --
-      2.49.0
-    LIBC_SPEED_UP_PATCH_EOF
-    system 'patch', '-p1', '-i', '0002-libc-Speed-up-large-memcpy-on-Cortex-A7-A15.patch'
+
+    Dir.glob('crew-package-glibc/patches/*.patch') do |patch|
+      puts "Applying #{patch}...".yellow
+      system 'patch', '-p1', '-i', patch
```
uberhacker, line 40: Rather than patching here, why not patch at crew-package-glibc instead?  Or better yet just move everything over here instead of referencing crew-package-glibc?

packages/glibc_standalone.rb
```diff
@@ -29,162 +29,44 @@ class Glibc_standalone < Package
...
       LD:       'ld.lld' # use lld here as mold will segfault
     }
 
     config_opts = %W[
       --prefix=#{CREW_PREFIX}
-      --libdir=#{glibc_libdir}
+      --libdir=#{CREW_GLIBC_PREFIX}
       --libexecdir=#{CREW_PREFIX}/libexec
-      --mandir=#{CREW_PREFIX}/man
+      --mandir=#{CREW_PREFIX}/share/man
```
uberhacker, line 64: Replace `#{CREW_PREFIX}/share/man` with `#{CREW_MAN_PREFIX}`?

lib/fixup.rb
```diff
@@ -254,30 +254,18 @@ def self.postremove; end
...
+  puts 'Decompressing binaries with upx...'.yellow
+  system "find #{CREW_PREFIX}/bin -type f -executable -print | xargs -P#{CREW_NPROC} -n1 upx -qq -d 2> /dev/null"
+  system "find #{CREW_LIB_PREFIX} -type f -executable -print | xargs -P#{CREW_NPROC} -n1 upx -qq -d 2> /dev/null"
+
+  # Switch to glibc_standalone if installed
+  puts 'Switching to glibc_standalone for all installed executables...'.yellow
+  system "find #{CREW_PREFIX}/bin -type f -executable -print | xargs -P#{CREW_NPROC} -n1 patchelf --set-interpreter #{CREW_GLIBC_INTERPRETER} 2> /dev/null"
+  system "find #{CREW_PREFIX}/bin -type f -executable -print | xargs -P#{CREW_NPROC} -n1 patchelf --remove-rpath 2> /dev/null"
+else
+  abort 'Please install upx and patchelf first by running \'crew install upx patchelf\'.'.lightred
```
uberhacker, line 268: Are we preventing any future compressed binaries with upx?  Seems pointless to compress and then uncompress when the end result is uncompressed.

satmandu: Yes. We are avoiding upx compression. It's another level of overhead that conflicts with the Glibc patching...

satmandu:
The ldconfig line in glibc_standalone.rb is broken. There is no `#{CREW_GLIBC_PREFIX}/ldconfig`.

Do this in a container and you will see:
```
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=set_dynamic_linker crew update
sed -i "s,#{CREW_GLIBC_PREFIX}/ldconfig,ldconfig," glibc_standalone.rb
yes | crew upgrade
```

Even then you get a list like this:
```
ldconfig: /usr/local/lib64/libstdc++.so.6 is not a symbolic link

ldconfig: /usr/local/lib64/liblsan.so.0 is not a symbolic link

ldconfig: /usr/local/lib64/libitm.so.1 is not a symbolic link

ldconfig: /usr/local/lib64/libubsan.so.1 is not a symbolic link

ldconfig: /usr/local/lib64/libcc1.so.0 is not a symbolic link

ldconfig: /usr/local/lib64/libquadmath.so.0 is not a symbolic link

ldconfig: /usr/local/lib64/libhwasan.so.0 is not a symbolic link

ldconfig: /usr/local/lib64/libobjc.so.4 is not a symbolic link

ldconfig: /usr/local/lib64/libasan.so.8 is not a symbolic link

ldconfig: /usr/local/lib64/libgfortran.so.5 is not a symbolic link

ldconfig: /usr/local/lib64/libtsan.so.2 is not a symbolic link
```

packages/glibc_standalone.rb
```diff
@@ -198,19 +80,35 @@ def self.build
...
+
+    # install crew-audit
+    FileUtils.install 'crew-package-glibc/crew-audit.so', File.join(CREW_DEST_DIR, CREW_GLIBC_PREFIX, 'crew-audit.so'), mode: 0o755
+    File.write "#{CREW_DEST_PREFIX}/etc/env.d/10-glibc", "LD_AUDIT=#{File.join(CREW_GLIBC_PREFIX, 'crew-audit.so')}\n"
+  end
+
+  def self.postinstall
+    # update search cache for ld.so
+    FileUtils.cp '/etc/ld.so.conf', "#{CREW_PREFIX}/etc/ld.so.conf"
+    File.write "#{CREW_PREFIX}/etc/ld.so.conf", "#{CREW_GLIBC_PREFIX}\n", mode: 'a'
```
satmandu, line 111: Shouldn't `/usr/local/opt/glibc-libs` go to the top of the list here?

packages/glibc_standalone.rb
supechicken: We need to apply the patches here anyways. If we created a patch script in crew-package-glibc, we still need to call it here.

packages/glibc_standalone.rb
supechicken: Chnage made

satmandu (approved):

PR #11821, opened 2025-04-25
Refactor upload

satmandu:
- The goal here is to refactor crew's upload to handle adjusting package files for remote binaries, and better document the algorithm.
- I have also added a port of the lockfile gem, to use to create lockfiles for downloaded binaries and cache files which may be on shared network spaces, to avoid race conditions with uploads during parallel builds.

##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=refactor_upload crew update \
&& yes | crew upgrade
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -32,6 +32,7 @@ require_relative '../commands/search'
 require_relative '../commands/sysinfo'
 require_relative '../commands/whatprovides'
 require_relative '../lib/const'
+require_relative '../lib/crew_lockfile'
 require_gem 'pry-byebug' if CREW_DEBUG
 # Debugging starts at the following breakpoint when enabled via crew
 # being started with a --debug flag.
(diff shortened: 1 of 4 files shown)


bin/crew
```diff
@@ -203,8 +204,14 @@ def cache_build
       FileUtils.mv build_cachefile, "#{build_cachefile}.bak", force: true if File.file?(build_cachefile)
       FileUtils.mv "#{build_cachefile}.sha256", "#{build_cachefile}.sha256.bak", force: true if File.file?("#{build_cachefile}.sha256")
       Dir.chdir(CREW_BREW_DIR) do
-        system "tar c#{@verbose} #{pkg_build_dirname} \
-        | nice -n 20 zstd -c  --ultra --fast -f -o #{build_cachefile} -"
+        @build_cachefile_lockfile = CrewLockfile.new "#{build_cachefile}.lock"
+        begin
+          @build_cachefile_lockfile.lock
```
uberhacker, line 209: Does this line do anything?  AFAICT, it only outputs the lock file.  Should it be preceded with a command like to check if it exists?

lib/crew_lockfile.rb
```diff
@@ -0,0 +1,568 @@
...
+      begin
+        lock
+        yield
+      ensure
+        unlock
+      end
+    end
+
+    def lock
+      raise StackingLockError, "<#{@path}> is locked!" if @locked
```
uberhacker, line 205: Add `.lightred` since this is an error?

bin/crew
uberhacker: EDIT: OK, I see this is calling the lock method of the CrewLockfile class.

satmandu: I think `../lib/crew_lockfile` handles that automatically:
```ruby
      begin
        new(path, opts).lock
      rescue LockError
        raise Errno::EEXIST, path
      end
      File.open(path, *options, &)
```

lib/crew_lockfile.rb
```diff
@@ -0,0 +1,568 @@
...
+              i = 0
+              begin
+                trace { "polling attempt <#{i}>..." }
+                begin
+                  File.link tmp_path, @path
+                rescue Errno::ENOENT
+                  try_again!
+                end
+                lock_stat = File.lstat @path
+                raise StatLockError, 'stats do not agree' unless
```
uberhacker, line 229: Add `.lightred`?

lib/crew_lockfile.rb
```diff
@@ -0,0 +1,568 @@
...
+            FileUtils.rm_f path
+          end
+        else
+          trace { "ignoring <#{path}> generated by <#{host}>" }
+        end
+      rescue StandardError
+        next
+      end
+    rescue StandardError => e
+      warn(errmsg(e))
```
uberhacker, line 347: Add `.lightred`?

satmandu: Added `.lightred` to the errmsg function.

uberhacker (approved):

PR #11822, opened 2025-04-25
Adjust counter for crew while loops.

satmandu:
- Move counter variable outside of whilte loops.


uberhacker (commented):
Bump version in `lib/const.rb`?

uberhacker (approved):

PR #11861, opened 2025-05-01
crew: Fix more SSL errors

supechicken:
Fixes https://github.com/chromebrew/chromebrew/pull/11858#issuecomment-2842985890

```bash
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=fix-ssl-part2 crew update \
&& yes | crew upgrade
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -50,6 +50,10 @@ Layout/FirstHashElementIndentation:
 Layout/LineLength:
   Enabled: false
 
+# Adding a space after {} block improves consistency.
+Layout/SpaceAroundMethodCallOperator:
+  Enabled: false
+
 # Having an explicit return can often improve readability.
 Style/RedundantReturn:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -50,10 +50,6 @@ Layout/FirstHashElementIndentation:
 Layout/LineLength:
   Enabled: false
 
-# Adding a space after {} block improves consistency.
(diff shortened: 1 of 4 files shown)


supechicken:
The linter keeps claiming there is an extra indentation, but I don't see any issue...

.rubocop.yml
```diff
@@ -50,6 +50,10 @@ Layout/FirstHashElementIndentation:
 Layout/LineLength:
   Enabled: false
 
+# Adding a space after {} block improves consistency.
+Layout/SpaceAroundMethodCallOperator:
```
Zopolis4, line 54: Disabling this means code like this is now allowed:
```ruby
foo. bar
foo &.bar
```
https://docs.rubocop.org/rubocop/1.75/cops_layout.html#layoutspacearoundmethodcalloperator

I think this cop improves overall readability, even if it might hurt your specific case.

satmandu: @supechicken Just add a specific override of this cop in your code for your specific use?

supechicken: This cop is not configurable. Now finding a way to satisfy the check without disabling it...

supechicken: @Zopolis4 @satmandu Does the new syntax in [b661668](https://github.com/chromebrew/chromebrew/pull/11861/commits/b6616686d6107f961f41fd462b249d8da7f052be) looks better?

satmandu (approved):

PR #11866, opened 2025-05-02
glibc_standalone: Switch to `tar.xz` for prebuilt binaries

supechicken:


uberhacker (approved):
Is this because older releases don't work with .zst?

PR #11867, opened 2025-05-02
Add zstd_static package, use it if necessary in install.sh

satmandu:


install.sh
```diff
@@ -191,7 +191,14 @@ echo_out 'Set up the local package repo...'
 # Download the chromebrew repository.
 curl_wrapper -L --progress-bar https://github.com/"${OWNER}"/"${REPO}"/tarball/"${BRANCH}" | tar -xz --strip-components=1 -C "${CREW_LIB_PATH}"
 
-BOOTSTRAP_PACKAGES='glibc_standalone libxcrypt upx patchelf lz4 zlib xzutils zstd zlib_ng crew_mvdir ruby git ca_certificates libyaml openssl gmp'
+BOOTSTRAP_PACKAGES=
+ZSTD_STATUS=
+if ! zstd --help &>/dev/null; then
+  ZSTD_STATUS="broken"
+  BOOTSTRAP_PACKAGES+='zstd_static'
```
uberhacker, line 198: Add a space before zstd_static?

install.sh
```diff
@@ -191,7 +191,14 @@ echo_out 'Set up the local package repo...'
 # Download the chromebrew repository.
 curl_wrapper -L --progress-bar https://github.com/"${OWNER}"/"${REPO}"/tarball/"${BRANCH}" | tar -xz --strip-components=1 -C "${CREW_LIB_PATH}"
 
-BOOTSTRAP_PACKAGES='glibc_standalone libxcrypt upx patchelf lz4 zlib xzutils zstd zlib_ng crew_mvdir ruby git ca_certificates libyaml openssl gmp'
+BOOTSTRAP_PACKAGES=
+ZSTD_STATUS=
+if ! zstd --help &>/dev/null; then
+  ZSTD_STATUS="broken"
```
uberhacker, line 197: Surprised rubocop didn't complain about the double quotes.

satmandu: This is shellcode, not ruby.

packages/zstd_static.rb
```diff
@@ -0,0 +1,1545 @@
...
+
+  def self.install
+    case ARCH
+    when 'i686'
+      @cross = 'i686-linux-musl'
+    when 'x86_64'
+      @cross = 'x86_64-linux-musl'
+    when 'aarch64', 'armv7l'
+      @cross = 'arm-linux-musleabihf'
+    end
```
uberhacker, line 1542: Why not add this logic outside all the class methods but still inside the class instead of duplicating in both build and install methods?

satmandu: Doesn't that mean that gets processed every time the file is read by crew? Otherwise it is only processed when a build or install happens.

install.sh
uberhacker: > Surprised rubocop didn't complain about the double quotes.

Didn't realize this was a bash shell script.

PR #11869, opened 2025-05-03
Adjust libc version detection.

satmandu:
- Need to adjust LIBC version detection during install, otherwise older glibc overwrites glibc_standalone, and that wreaks havoc.
- This uses a variant of the older LIBC detection code we used to use.
- Bypassing container checks until this is merged and new containers get built...

##
##
```bash
CREW_REPO=https://github.com/satmandu/chromebrew.git CREW_BRANCH=master crew update \
&& yes | crew upgrade
```

### DIFF
--- a/packages/glibc.rb
+++ b/packages/glibc.rb
@@ -1,10 +1,4 @@
 require 'package'
-Package.load_package("#{__dir__}/glibc_build223.rb")
-Package.load_package("#{__dir__}/glibc_build227.rb")
-Package.load_package("#{__dir__}/glibc_build232.rb")
-Package.load_package("#{__dir__}/glibc_build233.rb")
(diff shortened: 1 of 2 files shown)


packages/glibc.rb
```diff
@@ -38,6 +39,10 @@ class Glibc < Package
     version Glibc_build237.version
     compatibility Glibc_build237.compatibility
     depends_on 'glibc_lib237'
+  when '2.41'
```
uberhacker, line 42: Does this mean we need to update when the glibc version bumps to 2.42?  Would a `>=` comparison prevent requiring another update here?

satmandu: Logic changed.

uberhacker (approved):

PR #11872, opened 2025-05-03
Adjust gcc versions for current  limited LIBC_VERSION variation.

satmandu:


packages/gcc_lib.rb
```diff
@@ -4,11 +4,7 @@
 class Gcc_lib < Package
   description 'GCC shared libs except libgccjit'
   homepage Gcc_build.homepage
-  @gcc_libc_version = if %w[2.23 2.27 2.32 2.33 2.35 2.37].any? { |i| LIBC_VERSION.include? i }
-                        LIBC_VERSION
-                      else
-                        ARCH.eql?('i686') ? '2.23' : '2.27'
-                      end
+  @gcc_libc_version = ARCH.eql?('i686') ? '2.23' : '2.27'
```
uberhacker, line 7: Why not just use `LIBC_VERSION` for everything?

satmandu: This is a stopgap measure. It will go away in gcc15.

uberhacker (approved):

uberhacker (approved):

PR #11911, opened 2025-05-19
glibc-standalone: Rename to `glibc`, add a LD_PRELOAD library with useful features

supechicken:
Requires https://github.com/chromebrew/crew-profile-base/pull/33

## Changes
- **All non-shared libraries are moved into `CREW_LIB_PREFIX` to make sure `gcc` can find them**
- **`glibc_standalone` is renamed to `glibc`**
- All existing `glibc_{lib,dev,build}` packages are removed, as they are no longer needed.
- `LD_LIBRARY_PATH` is no longer hardcoded in the patches, instead, it appends `CREW_GLIBC_PREFIX` to the existing `LD_LIBRARY_PATH` environment variable
- Removed `crew-audit.so`, which is superseded by `crew-preload.so`
- An `LD_PRELOAD` library called `crew-preload.so` has been added for multiple usages.

## About `crew-preload.so`
This library has multiple functions that are helpful in Chromebrew environments:
- Redirect all linker calls to `mold` (just like what `mold -run` does)
- Append `--dynamic-linker ...` parameter to linkers automatically, ensuring the ELF interpreter path is always set to our dynamic linker
- Fix hardcoded shebang/command paths and redirect them to `CREW_PREFIX` if possible (things like `filefix` are not needed anymore)
- Always redirect `#!/bin/sh` to Chromebrew version of shell
- Unset `LD_LIBRARY_PATH` when running system commands to prevent unexpected issues

Tested on `i686`

```shell
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=crew_preload crew update
crew reinstall -s glibc
```


satmandu:
The arm build of `glibc` is missing. Building that now...

supechicken:
> The arm build of `glibc` is missing. Building that now...

Thanks a lot :) Btw the build workflow seems to have some issues, the build completed but the workflow still fails

lib/const.rb
```diff
@@ -284,7 +280,7 @@
   -DCMAKE_C_FLAGS='#{CREW_COMMON_FLAGS.gsub(/-fuse-ld=.{2,4}\s/, '')}' \
   -DCMAKE_CXX_FLAGS='#{CREW_COMMON_FLAGS.gsub(/-fuse-ld=.{2,4}\s/, '')}' \
   -DCMAKE_EXE_LINKER_FLAGS='#{CREW_LINKER_FLAGS}' \
-  -DCMAKE_LINKER_TYPE=#{CREW_LINKER.upcase} \
+  -DCMAKE_LINKER_TYPE=MOLD \
```
uberhacker, line 283: Not variable?

satmandu:
Also, do you want this built with `--disable-crypt`? Redhat has apparently been doing this for years, and we're already shipping `libxcrypt`...

supechicken:
> Also, do you want this built with `--disable-crypt`? Redhat has apparently been doing this for years, and we're already shipping `libxcrypt`...

It seems to be disabled by default in 2.41 (there are no `libcrypt.so` in the filelist)

lib/const.rb
supechicken: I have removed it due to the lack of use cases. Why use other linkers if there is a much faster one :)

satmandu: Reverted.

.github/workflows/Unit-Test.yml
```diff
@@ -130,25 +130,27 @@ jobs:
...
                 # Export the x86_64 container depending on whether this PR updates packages with appropriate minimum glibc.
-                if [[ $GLIBC_232_COMPATIBLE_PACKAGES ]]; then
-                    echo "CONTAINER=nocturne-x86_64.m97" >> "$GITHUB_ENV"
-                elif [[ $GLIBC_237_COMPATIBLE_PACKAGES ]]; then
-                    echo "CONTAINER=hatch-x86_64.m135" >> "$GITHUB_ENV"
-                else
-                    echo "CONTAINER=nocturne-x86_64.m90" >> "$GITHUB_ENV"
-                fi
+                # if [[ $GLIBC_232_COMPATIBLE_PACKAGES ]]; then
+                #     echo "CONTAINER=nocturne-x86_64.m97" >> "$GITHUB_ENV"
```
supechicken, line 134: @satmandu We don't need to care about the `glibc` version of the container anymore, as I have extracted some glibc libraries from the oldest available ChromeOS images (link with those libraries if we want binaries also works on system's `glibc`, see [chromebrew/crew-package-glibc](https://github.com/chromebrew/crew-package-glibc/tree/main/prebuilt)). This also means we can now use the latest version of ChromeOS for building things

satmandu: That's fine... It was just using the M97 containers for tests, which was a waste of time...

satmandu: The older chromeos based containers are smaller, so probably load faster?

supechicken: Maybe try stripping out the Android subsystem (`/opt/google/vms/android`) and Chrome (`/opt/google/chrome`) when building a new container image next time? That would save a ton of space for us.

satmandu:
Current breakage is from this after `crew update` before anything else happens on arm:
```
CREW_GLIBC_INTERPRETER=/usr/local/opt/glibc-libs/ld-linux-aarch64.so.1
```

satmandu:
The Arm issue is fixed. I think that the other unit test failures are due to specific issues regarding packages that are unconnected to the glibc changes in this PR.

satmandu (approved):

.github/workflows/Unit-Test.yml
satmandu: This is in the container image script being used starting tonight ...

lib/const.rb
```diff
@@ -31,8 +31,9 @@
 CREW_PREFIX ||= ENV.fetch('CREW_PREFIX', '/usr/local') unless defined?(CREW_PREFIX)
 
 # Glibc related constants
-CREW_GLIBC_PREFIX      ||= File.join(CREW_PREFIX, 'opt/glibc-libs')
-CREW_GLIBC_INTERPRETER ||= File.symlink?("#{CREW_PREFIX}/bin/ld.so") ? File.realpath("#{CREW_PREFIX}/bin/ld.so") : nil unless defined?(CREW_GLIBC_INTERPRETER)
+CREW_GLIBC_PREFIX ||= File.join(CREW_PREFIX, 'opt/glibc-libs')
+@crew_glibc_interpreter = File.join(CREW_GLIBC_PREFIX, File.basename(File.realpath("#{CREW_PREFIX}/bin/ld.so"))) unless defined?(CREW_GLIBC_INTERPRETER)
+CREW_GLIBC_INTERPRETER ||= File.file?(@crew_glibc_interpreter) ? @crew_glibc_interpreter : nil unless defined?(CREW_GLIBC_INTERPRETER)
```
uberhacker, line 36: Bump version?  Although no changes to `bin/crew`, there are changes to this file. :)

supechicken: Done

satmandu (approved):

PR #11926, opened 2025-05-21
iperf3 3.7 => 3.19

supechicken:
Binaries needed, and [the build workflow refuses to work](https://github.com/chromebrew/chromebrew/actions/runs/15168929846)😅

Tested and built on `x86_64`


satmandu:
I686 should be removed from compatibility.

packages/iperf3.rb
```diff
@@ -17,16 +17,5 @@ class Iperf3 < Package
...
-           "--prefix=#{CREW_PREFIX}",
-           '--disable-dependency-tracking',
-           '--disable-maintainer-mode'
-    system 'make'
-  end
-
-  def self.install
-    system 'make', "DESTDIR=#{CREW_DEST_DIR}", 'install'
-  end
+  configure_options '--disable-maintainer-mode'
```
Zopolis4, line 20: maintainer mode is already disabled by default for iperf:
https://github.com/esnet/iperf/blob/master/configure.ac#L44C1-L44C19
https://www.gnu.org/software/automake/manual/html_node/maintainer_002dmode.html

uberhacker:
> I686 should be removed from compatibility.

I removed the binary too.

satmandu:
Also fixed the buil workflow here...

satmandu (approved):
Someone else will have to approve this too...

uberhacker (approved):

PR #11943, opened 2025-05-24
Cleanup const.rb and add CMAKE_INSTALL_LIBDIR option to cmake builds.

satmandu:
- Another attempt at fixing the workflows.
- Also adds `CMAKE_INSTALL_LIBDIR=#{ARCH_LIB}` as some cmake packages are using that.
- Built `libarchive` with this PR and the changes in https://github.com/libarchive/libarchive/pull/2509 to make sure that worked.


uberhacker (approved):

lib/const.rb
```diff
@@ -282,6 +280,7 @@
 # CMAKE_LIBRARY_PATH is the build LIBRARY_PATH, as opposed to
 # CMAKE_INSTALL_LIBDIR which defaults to lib or lib64.
 CREW_CMAKE_OPTIONS ||= <<~OPT.chomp
+  -DCMAKE_INSTALL_LIBDIR=#{ARCH_LIB} \
```
Zopolis4, line 283: As I recall this should only be used for certain packages, as this overrides the correct value of CMAKE_INSTALL_LIBDIR as set by GNUInstallDirs

Zopolis4: I'm not sure-- I have been going around fixing packages that don't use GNUInstallDirs, so I'd rather know which ones need to be fixed.

Zopolis4: fwiw, libarchive now uses `GNUInstallDirs`: https://github.com/libarchive/libarchive/pull/2676, so we can drop the postinstall patch (because thats how our package file works around it) once a new version is released and we upgrade to it.

PR #11998, opened 2025-06-06
Run rubocop 1.76.0 on tree

Zopolis4:
## Description
Fittingly, most of the new offense are from the `Style/EmptyStringInsideInterpolation` cop I added based on patterns I kept seeing in chromebrew. The only other new offense is from `Lint/EmptyInterpolation`, which I also changed :grin:.

There's also some new `Naming/PredicateMethod` offenses that arent autocorrectable and I don't think are valid, so I'll look at that later.

```bash
CREW_REPO=https://github.com/Zopolis4/chromebrew.git CREW_BRANCH=pride crew update \
&& yes | crew upgrade
```


Zopolis4:
idk whats going on with yamllint but those configuration lines def should be 2 indented rather than 1, which is what it wants apparently

satmandu (approved):

PR #12037, opened 2025-06-14
lib/const: Wrap all constant definitions with a function

supechicken:
This PR simplifies all `CONST = ... unless defined?(...)` syntax into `crew_const :CONST, ...` for convenience:
 
```ruby
# before
CREW_CONST ||= '1.0' unless defined?(CREW_CONST)

# after
crew_const :CREW_CONST, '1.0'
```

```bash
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=wrap_const crew update \
&& yes | crew upgrade
```

### DIFF
--- a/lib/const.rb
+++ b/lib/const.rb
@@ -6,7 +6,6 @@ require 'open3'
 def crew_const(const_name, default_value = nil) = (Object.const_set(const_name, default_value) unless Object.const_defined?(const_name))
 
 crew_const :CREW_VERSION, '1.61.10'
-crew_const :OLD_CREW_VERSION, CREW_VERSION
 
 # Kernel architecture.
 crew_const :KERN_ARCH, Etc.uname[:machine]


lib/const.rb
```diff
@@ -3,256 +3,250 @@
 require 'etc'
 require 'open3'
 
-OLD_CREW_VERSION ||= defined?(CREW_VERSION) ? CREW_VERSION : '1.0'
-CREW_VERSION ||= '1.61.9' unless defined?(CREW_VERSION) && CREW_VERSION == OLD_CREW_VERSION
+def crew_const(const_name, default_value = nil) = (Object.const_set(const_name, default_value) unless Object.const_defined?(const_name))
+
+crew_const :CREW_VERSION, '1.61.10'
+crew_const :OLD_CREW_VERSION, CREW_VERSION
```
uberhacker: Don't you want to assign the old version prior to assigning the new version?

lib/const.rb
```diff
@@ -3,256 +3,250 @@
...
+             #{'glibc' unless CREW_GLIBC_INTERPRETER.nil?}
+             #{
+               if LIBC_VERSION.to_f > 2.34 && LIBC_VERSION.to_f < 2.41
+                 "#{"glibc_lib#{LIBC_VERSION.delete('.')}" if File.file?(File.join(CREW_PREFIX, "etc/crew/meta/glibc_lib#{LIBC_VERSION.delete('.')}.filelist"))}
+                  #{"glibc_build#{LIBC_VERSION.delete('.')}" if File.file?(File.join(CREW_PREFIX, "etc/crew/meta/glibc_build#{LIBC_VERSION.delete('.')}.filelist"))}"
+               else
+                 File.file?(File.join(CREW_PREFIX, "etc/crew/meta/glibc_build#{LIBC_VERSION.delete('.')}.filelist")) ? "glibc_build#{LIBC_VERSION.delete('.')}" : ''
+               end
+             }
+           ].reject!(&:empty?)
```
uberhacker, line 68: Too much indentation?

lib/const.rb
```diff
@@ -310,36 +304,34 @@
...
-CREW_LICENSE ||= <<~LICENSESTRING
+crew_const :CREW_CMAKE_LIBSUFFIX_OPTIONS, "#{CREW_CMAKE_OPTIONS} -DLIB_SUFFIX=#{CREW_LIB_SUFFIX}"
+
+crew_const :GEM_ARCH, case ARCH
+                      when 'x86_64'
+                        'x86_64-linux'
+                      when 'i686'
+                        'x86-linux'
+                      when 'aarch64', 'armv7l'
+                        'armv8l-linux-eabihf'
```
uberhacker, line 314: armv7l (32 bit) should not be associated with armv8l (64 bit).

lib/const.rb
uberhacker: Just mentioning this since rubocop might balk.

uberhacker (changes requested):
See comments...

Zopolis4:
We'll be able to revert back to normal constant assignment once I finish up some changes to `lib/fixup.rb`, but I'm fine with improving it in the meantime.

lib/const.rb
supechicken: Please read the comment above...
```
# We report aarch64 & armv8l machines as armv7l for now, as we treat
# them as if they were armv7l.
```

lib/const.rb
supechicken: I think so, but rubocop wants this 😅

lib/const.rb
supechicken: The `OLD_CREW_VERSION` has `CREW_VERSION` as its default value, so it must be set after `CREW_VERSION` is set

lib/const.rb
uberhacker: Then it should be `armv7l-linux-eabihf` instead of `armv8l-linux-eabihf`. :)

lib/const.rb
uberhacker: Doesn't `OLD_CREW_VERSION` get set to `1.61.10` instead of the previous crew version?

lib/const.rb
uberhacker: Rubocop rules are...confusing. :grimacing:

lib/const.rb
supechicken: Just realized this constant is unused, removing it now...

lib/const.rb
supechicken: I am not sure about this, maybe @satmandu can answer this?

satmandu:
Sorry, I've been traveling internationally... 😅

Let me know if you still want me to look at this.

supechicken:
> Sorry, I've been traveling internationally... 😅
> 
> Let me know if you still want me to look at this.

I didn't give up on this 😅 Just looking for a better way to do this (something like wrapping with a class)

PR #12081, opened 2025-06-20
crew: Remove `activesupport` gem, implement `.blank?` logic in `lib/convenience_functions`

supechicken:
It makes no sense to load a ~1MB external gem library just for one single function, especially when it can be easily reimplemented within a few lines...

```bash
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=no_activesupport crew update \
&& yes | crew upgrade
```

### DIFF
--- a/lib/convenience_functions.rb
+++ b/lib/convenience_functions.rb
@@ -6,12 +6,12 @@ require_relative 'const'
 require_relative 'crewlog'
 require_relative 'downloader'
 
-class NilClass; def blank? = true; end
-class Numeric;  def blank? = false; end
-class Array;    def blank? = self.empty?; end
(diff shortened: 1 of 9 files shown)


lib/convenience_functions.rb
```diff
@@ -6,6 +6,13 @@
 require_relative 'crewlog'
 require_relative 'downloader'
```
Zopolis4, line 8: Add a comment here explaining that this is a reimplementation of ActiveSupport's blank? it's a little confusing without context

supechicken: Change made

Zopolis4 (approved):
Nice work!

PR #12083, opened 2025-06-20
crew: Fix `.blank?` call for proc blocks

supechicken:
```bash
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=fix_blank crew update \
&& yes | crew upgrade
```


satmandu:
Is this good to replace our active support calls everywhere? I think maybe we were using one more thing from there, but I could be wrong...

PR #12104, opened 2025-06-25
Add ruby_matrix to core

satmandu:


packages/core.rb
```diff
@@ -94,6 +94,8 @@ class Core < Package
   depends_on 'ruby_gem_compiler'
   # For use in ruby prompts.
   depends_on 'ruby_highline'
+  # Needed by lib/misc_functions.
+  depends_on 'ruby_matrix'
```
uberhacker, line 98: Does the latest version cause an error?

satmandu: It does not in the container I am opening.

uberhacker (approved):

PR #12119, opened 2025-06-28
Remove need to install in VT-2 session.

uberhacker:
VT-2 session login should not be required to install.

```
LD_PRELOAD= CREW_PY_VER=3.14 OWNER=uberhacker BRANCH=update-install.sh /bin/bash <(/usr/bin/curl -L https://raw.githubusercontent.com/uberhacker/chromebrew/refs/heads/update-install.sh/install.sh) && . ~/.bashrc
```


install.sh
```diff
@@ -63,46 +63,40 @@ else
...
-# Check if the user is on ChromeOS v117+ and not in the VT-2 console, as sudo will not work.
-: "${CREW_PREFIX:=/usr/local}"
-if [[ "$(stat -c '%u' "${CREW_PREFIX}")" == "$(id -u)" ]] && sudo 2>&1 | grep -q 'no new privileges'; then
-  echo_error "Please run the installer in the VT-2 shell."
-  echo_info "To start the VT-2 session, type Ctrl + Alt + ->"
+# If the user does not own the installation directory, exit with instructions how to fix.
+if [ "$(stat -c '%u' "${CREW_PREFIX}")" != "$(id -u)" ]; then
+  echo_error "${CREW_PREFIX} is not owned by $(id -u)."
+  echo_info "Execute the following and try again:"
+  echo "sudo chown -R $(id -u):$(id -g) ${CREW_PREFIX}"
```
satmandu: We should leave the VT-2 instructions here so people know how to use sudo if it is necessary, right?

uberhacker: No, I want users to realize VT-2 sessions are never a requirement to install.  The only reason to do this would be because you want to use sudo but there isn't (and should never be) a reason to do so.  When you start adding root permissions to files and folders, it becomes a nightmare to manage the bash session and makes it even harder to debug errors when they occur.

satmandu: I'm just asking for “Execute the following and try again” to include detailed instructions because otherwise they will try sudo _as per the instruction_ and it won't work...

uberhacker: Anyone who is in this situation has done something out of the box to change the permissions which in most cases would never need to be done.  I understand what you are suggesting but it's highly likely they already know how to use sudo to get themselves in this predicament in the first place.  I can certainly add instructions "sudo for dummies" back in, however.

satmandu:
Any luck with this?

uberhacker:
> Any luck with this?

Unfortunately, not yet.  Still get segfaults and a weird insufficient randomness error when trying to access git during the install.  I think we need to do a better job of isolating Docker container environments from actual hardware.  There are several Docker specific sections that crash on hardware.

uberhacker:
There really is no way around the need to use sudo (and therefore VT-2) if permissions are changed in `/usr/local`.  The current installer minimizes the usage as much as possible.

PR #12375, opened 2025-08-07
Update version.rb to only update versions with -u or --update-package-files

uberhacker:
This fixes a bug that updated the versions even if the `--update-package-files` switch was not provided.

### DIFF
--- a/tools/version.rb
+++ b/tools/version.rb
@@ -1,13 +1,15 @@
 #!/usr/bin/env ruby
-# version.rb version 1.2 (for Chromebrew)
+# version.rb version 1.3 (for Chromebrew)
+
+OPTIONS = %w[-h --help -u --update-package-files -v --verbose]
 
 if ARGV.include?('-h') || ARGV.include?('--help')
   abort <<~EOM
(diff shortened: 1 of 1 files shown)


tools/version.rb
```diff
@@ -25,7 +27,8 @@
 require_relative '../lib/package'
 require_relative '../lib/package_utils'
 
-UPDATE_PACKAGE_FILES = ARGV.include?('--update-package-files')
+UPDATE_PACKAGE_FILES = ARGV.include?('-u') || ARGV.include?('--update-package-files')
```
Zopolis4, line 30: If you wanted to get fancy, you could just have `ARGV.include?('-u')`, as `-u` is contained within `--update-package-files`. Same with `-v` and `--verbose`.

Whether that would actually be an improvement in terms of code clarity, I'm doubtful.

uberhacker: In this context, ARGV is an array, not a string so I'm not sure it would work the way you might think.

Zopolis4: ah

Zopolis4 (approved):

PR #12442, opened 2025-08-14
Auto-generated update for signal_desktop (7.66.0)

supechicken:
### Package information
- Package name: signal_desktop
- Old version: 7.64.0
- New version: 7.66.0

This PR is generated by [crew-update-checker](https://github.com/supechicken/crew-update-checker)

### DIFF
--- a/packages/signal_desktop.rb
+++ b/packages/signal_desktop.rb
@@ -46,4 +46,3 @@ class Signal_desktop < Package
     ExitMessage.add "\nType 'signal-desktop' to get started.\n"
   end
 end
-
--- a/packages/signal_desktop.rb
+++ b/packages/signal_desktop.rb
@@ -3,12 +3,17 @@ require 'package'
 class Signal_desktop < Package
   description 'Private Messenger for Windows, Mac, and Linux'
   homepage 'https://signal.org/'
-  version '7.64.0'
+  version '7.66.0'
   license 'AGPL-3.0'
   compatibility 'x86_64'
-  min_glibc '2.29'
-  source_url "https://updates.signal.org/desktop/apt/pool/s/signal-desktop/signal-desktop_#{version}_amd64.deb"
-  source_sha256 '02ef930e19f4b5d11c3e83cb17aef85b4ed895539e4fd8c9b817c45c07adc06e'
+
+  source_url({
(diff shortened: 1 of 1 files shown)


satmandu (changes requested):
Can you collapse the single source `source_url`s so they're not hashes?

packages/signal_desktop.rb
```diff
@@ -3,12 +3,12 @@
...
   homepage 'https://signal.org/'
-  version '7.64.0'
+  version '7.66.0'
   license 'AGPL-3.0'
   compatibility 'x86_64'
-  min_glibc '2.29'
-  source_url "https://updates.signal.org/desktop/apt/pool/s/signal-desktop/signal-desktop_#{version}_amd64.deb"
-  source_sha256 '02ef930e19f4b5d11c3e83cb17aef85b4ed895539e4fd8c9b817c45c07adc06e'
+
+  source_url 'https://updates.signal.org/desktop/apt/pool/s/signal-desktop/signal-desktop_7.66.0_amd64.deb'
```
satmandu, line 10: Do we want to use `#{version}` in the url here?

supechicken: No, that makes automation harder😅

satmandu: `tools/version.rb` handles it just fine...

supechicken: The point of using `#{version}` before is to reduce the effort of updating packages, but that's no longer an advantage if we have a fully automated update workflow like this :)

Additionally, adding `#{version}` support to `source_url` brings extra complexity to the update script...

satmandu: Counterpoint: https://github.com/chromebrew/chromebrew/pull/12445#issuecomment-3189104393

satmandu: There are now two separate universal updater scripts. Please don't break the one I've gotten working over the past week that is in active use.

supechicken: I see... Now putting `#{version}` back and putting all changes into one PR

uberhacker: There are special cases for excluding `#{version}`.  Sometimes the updated url has a git commit hash along with the version.  In these situations, simply abstracting the version is not enough.

PR #12443, opened 2025-08-14
Auto-generated update for vivaldi (7.5.3735.62-1)

supechicken:
### Package information
- Package name: vivaldi
- Old version: 7.5.3735.58-1
- New version: 7.5.3735.62-1

This PR is generated by [crew-update-checker](https://github.com/supechicken/crew-update-checker)


supechicken:
It should work with our glibc. Is `glibc_standalone` available in our docker images?

satmandu:
> It should work with our glibc. Is `glibc_standalone` available in our docker images?


You just need to set a min_glibc for packages that need a higher glibc so the unit tests use a higher glibc container.


That's all that's needed here.

supechicken:
> > It should work with our glibc. Is `glibc_standalone` available in our docker images?
> 
> You just need to set a min_glibc for packages that need a higher glibc so the unit tests use a higher glibc container.
> 
> That's all that's needed here.

Doesn't setting `min_glibc` here block some old systems from installing?

PR #12441, opened 2025-08-14
Auto-generated update for opera (120.0.5543.161)

supechicken:
### Package information
- Package name: opera
- Old version: 120.0.5543.128
- New version: 120.0.5543.161

This PR is generated by [crew-update-checker](https://github.com/supechicken/crew-update-checker)

PR #12440, opened 2025-08-14
Auto-generated update for google_chrome (139.0.7258.127-1)

supechicken:
### Package information
- Package name: google_chrome
- Old version: 139.0.7258.66-1
- New version: 139.0.7258.127-1

This PR is generated by [crew-update-checker](https://github.com/supechicken/crew-update-checker)

PR #12439, opened 2025-08-14
Auto-generated update for edge (139.0.3405.86-1)

supechicken:
### Package information
- Package name: edge
- Old version: 138.0.3351.121-1
- New version: 139.0.3405.86-1

This PR is generated by [crew-update-checker](https://github.com/supechicken/crew-update-checker)

PR #12438, opened 2025-08-14
Auto-generated update for brave (1.81.135)

supechicken:
### Package information
- Package name: brave
- Old version: 1.81.131
- New version: 1.81.135

This PR is generated by [crew-update-checker](https://github.com/supechicken/crew-update-checker)

PR #12449, opened 2025-08-14
Update most `.deb` based packages

supechicken:
Merge of #12437, #12438, #12439, #12440, #12441, #12442, #12443

```bash
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=update_debs crew update \
&& yes | crew upgrade
```


uberhacker:
If we cannot build the hello world package, we definitely have an issue. :grimacing:

satmandu:
https://github.com/chromebrew/chromebrew/pull/12450

packages/edge.rb
```diff
@@ -1,25 +1,24 @@
...
-require 'convenience_functions'
 
 class Edge < Package
   description 'Microsoft Edge is the fast and secure browser'
   homepage 'https://www.microsoft.com/en-us/edge'
-  version '138.0.3351.121-1'
+  version '139.0.3405.86-1'
   license 'MIT'
   compatibility 'x86_64'
-  min_glibc '2.29'
```
uberhacker, line 10: Are you sure it works with the lower glibc versions?

supechicken:
Just discovered that `brave` hangs when `LD_PRELOAD` is set. Sounds like something goes wrong in `crew-preload.so`...

packages/signal_desktop.rb
```diff
@@ -3,12 +3,12 @@
 class Signal_desktop < Package
   description 'Private Messenger for Windows, Mac, and Linux'
   homepage 'https://signal.org/'
-  version '7.64.0'
+  version '7.66.0'
   license 'AGPL-3.0'
   compatibility 'x86_64'
-  min_glibc '2.29'
```
uberhacker, line 9: Are you sure it works with the lower glibc versions?

packages/vivaldi.rb
```diff
@@ -1,14 +1,28 @@
...
-  version '7.5.3735.58-1'
+  version '7.5.3735.62-1'
   license 'Vivaldi'
   compatibility 'aarch64 armv7l x86_64'
   min_glibc '2.37'
 
+  source_url({
+    aarch64: "https://repo.vivaldi.com/stable/deb/pool/main/vivaldi-stable_#{version}_armhf.deb",
+    armv7l: "https://repo.vivaldi.com/stable/deb/pool/main/vivaldi-stable_#{version}_armhf.deb",
+    x86_64: "https://repo.vivaldi.com/stable/deb/pool/main/vivaldi-stable_#{version}_amd64.deb"
```
uberhacker: Align colons?

packages/edge.rb
supechicken: Probably not, but it should work using our glibc version

packages/signal_desktop.rb
satmandu: The highest `min_glibc` in this PR is 2.39, which is causing the unit tests to fail.

Move those higher `min_glibc` packages from this PR, and I suspect the unit tests will pass, since then this PR will run unit tests on a lower glibc container.

Maybe building on the glibc 2.41 containers is broken? @supechicken?

supechicken: Can we run the unit tests with our glibc? That way glibc version is not an issue anymore

Or I can add back those `min_glibc` lines if that causes issues...

satmandu: It makes sense to only use the newer glibc containers (which are much larger) for unit tests when we need them since loading a smaller container makes everything go faster.

satmandu: But we should definitely run unit tests once a day or so on all containers to make sure the containers aren't broken!

supechicken: > It makes sense to only use the newer glibc containers (which are much larger) for unit tests when we need them since loading a smaller container makes everything go faster.

Does stripping out things help in this case? Something like removing the `/opt/google` directory entirely (that thing takes 1.5GB in M139), `/usr/share/{fonts,chromeos-assets}` also takes a ton of space (~500MB) and could also be removed.

I am adding back the `min_glibc` lines for now...

satmandu: You only need the min_glibc for packages that complain when installing in the lower glibc containers.


Realistically, nobody is running anything on systems with an older glibc in the real world.

uberhacker: The reason we need to test on lower glibc versions is because older Chromebooks have reached EoL and likely don't exceed glibc 2.27 in many cases.

packages/vivaldi.rb
supechicken: Done

packages/signal_desktop.rb
Zopolis4: > Realistically, nobody is running anything on systems with an older glibc in the real world.

Well, apart from all the EOL devices...

Ideally we should try and keep as much compatibility with older devices as possible, especially since they don't really have an alternative.

satmandu: I still think the solution for THIS PR is to just separate out the packages into separate PRs, so that the unit tests don't fail.

The unit tests for the GUI apps are artificial in any case... and we're not actually testing the binaries to see if they work in the unit tests. (And when they DO complain, that's when we bump the `min_glibc` for that.

Essentially we have a `min_glibc` for build, and a `min_glibc` for running, and those are two entirely different things, and the builds from source DO test the former, and the unit tests SOMETIMES test the latter... but really not...

supechicken: The only package at this moment that might fail to install on older glibc is `vivaldi`, and if the workflow respects `min_glibc` then it should be good to go (all of these packages are marked as `no_compile_needed`, so we don't need to care about if it works during build)

uberhacker: Any package that depends on sommelier will fail because that doesn't run in containers.

satmandu:
@supechicken Do you want to get this PR taken care of?

supechicken:
> @supechicken Do you want to get this PR taken care of?

`gcc` is still segfaulting... (and now also `git` in armv7l workflow) Let's put it aside at this moment.

PR #12450, opened 2025-08-14
Revert "updater-zlib_ng-2.2.5 — zlib_ng => 2.2.5"

satmandu:
Reverts chromebrew/chromebrew#12444

Looks like this failed the unit tests.


uberhacker:
> Reverts #12444
> 
> Looks like this failed the unit tests.

Why not try to fix the update?

satmandu:
Not sure what the issue is. Better to debug and not break things until that happens.

(Don't have time to debug until later today...)

satmandu:
Actually, the issue might be building with the glibc 2.41 installed.

Issue #4834, opened 2020-12-29
Proposal: Check free disk space before install


uberhacker:
It would be nice if Chromebrew was smart enough to know the disk space requirement for each package before installation and then provide a friendly message accordingly in order to avoid the dreaded out of space message during install.  In order to do this, we would need to know the size of the source or binary tarball before and after extraction.  We would also need to know (if it was a source build) the size of all the files after install.  So basically, we would need to know the sum of the following:
- Size of tarball (source or binary)
- Size of extracted tarball
- Total size of all files in the package

satmandu:
Does any package manager do this?

uberhacker:
If others don't, does that mean we don't need to do it?

satmandu:
It would be useful, but if others have done it, it would also be useful to see how they do it so we don't reinvent the wheel.

satmandu:
Might also be nice to generate a space estimation when updating install.sh with package updates, so install can error out if there is insufficient free space.

uberhacker:
Yes, good thought.

cstrouse:
OpenSUSE's zypper has a few references to disk space checking before installing packages but is thin on how best to achieve this.

- https://github.com/openSUSE/zypper/issues/245
- https://bugzilla.opensuse.org/show_bug.cgi?id=1124282

Another package manager full of interesting and well-thought out features is [spack](https://spack.readthedocs.io/en/latest/command_index.html) for HPC but can be relocated and ran without root so many parallels exist with our constraints.

satmandu:
Actually just ran across a reference to spack while searching out ways to fix llvm pulling in system libraries: https://reviews.llvm.org/D85820

satmandu:
Also amazing how convergent evolution has basically led spack and chromebrew and arch, etc to similar package file formats:
e.g.
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/ghostscript/package.py

saltedcoffii:
> Does any package manager do this?
portage does for some of its ebuilds

uberhacker:
This would be worth adding to our base preflight method.

Zopolis4:
We can query the download url of a package and use the content-length header to find out how big the tarball would be, but I'm not sure how we'd find out the decompressed size.

uberhacker:
> We can query the download url of a package and use the content-length header to find out how big the tarball would be, but I'm not sure how we'd find out the decompressed size.

[...]

PR #12515, opened 2025-08-16
crew; Include file size in file list

supechicken:
Preparation of fixing #4834

In order to get the installed size before installation, the total size of all files will be included in the generated filelist:
```
$ cat /usr/local/etc/crew/meta/brave.filelist
 
# Total size: 419008441
/usr/local/bin/brave
/usr/local/share/brave/LICENSE
...snip...
```

```bash
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=filelist_embed_size crew update \
&& yes | crew upgrade
```


satmandu:
Why not create a separate file for package metadata like package size that can sit in the metadata dir?

supechicken:
> Why not create a separate file for package metadata like package size that can sit in the metadata dir?

In my view, it is better to keep it inside the filelist, since the total size is bound with the file content (the size changes when the content changes)

Creating a separate place just for it might be redundant in this case😅

satmandu:
So we have any logic being used to grep the filelists where changing the format might mess things up?

One of the getrealdeps scripts might be using those files?

Just check to make sure that you fix any of that logic too.

Zopolis4:
We can just skip the first line of the file list wherever we parse it, i assume

satmandu:
Agreed. Just maybe make such changes atomically with this PR so nothing breaks is all I'm saying.

uberhacker (approved):
Good idea!  We'll need to update all filelists eventually in order to use this to calculate free disk space needed to install packages.

supechicken:
> So we have any logic being used to grep the filelists where changing the format might mess things up?

> One of the getrealdeps scripts might be using those files?

It should be fine as scripts should check for the existence of file before processing (`File.exist?` will return `false` when it is not a file path at all)

For most places, I added some logic to ignore lines that start with `#` to ensure nothing breaks

satmandu:
bump version?

supechicken:
> bump version?

Done

PR #12522, opened 2025-08-17
Update snowflake

uberhacker:
Tested & Working properly:
- [x] `x86_64`
```bash
CREW_REPO=https://github.com/uberhacker/chromebrew.git CREW_BRANCH=update-snowflake crew update \
&& yes | crew upgrade
```

### DIFF
--- a/packages/snowflake.rb
+++ b/packages/snowflake.rb
@@ -3,53 +3,37 @@ require 'package'
 class Snowflake < Package
   description 'Snowflake (aka Muon) is a graphical SSH client.'
   homepage 'https://github.com/subhra74/snowflake'
-  version '1.3.0'
+  version '1.0.4'
   license 'GPL-3'
   compatibility 'x86_64'
-  source_url 'SKIP'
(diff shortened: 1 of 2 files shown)


packages/snowflake.rb
```diff
@@ -3,53 +3,37 @@
 class Snowflake < Package
   description 'Snowflake (aka Muon) is a graphical SSH client.'
   homepage 'https://github.com/subhra74/snowflake'
-  version '1.0.4'
+  version '1.0.4-1'
```
Zopolis4: We shouldn't need the version bump here, and if we do, something is broken which we should fix.

uberhacker: This does fix an issue with the jdk version.  It also prepares for the next update.

Zopolis4: I'm not saying we shouldn't upgrade the package, just that we don't need to bump the version

Zopolis4 (approved):

PR #12563, opened 2025-08-19
Add z_library package

uberhacker:
## Description
Introducing the alpha version of Z-Library launcher for desktop. Say goodbye to the hassle of searching for a working website link, as our desktop launcher takes care of everything for you. What`s more, it operates seamlessly over the TOR network, ensuring uninterrupted access to the library while keeping your browsing experience anonymous and your privacy protected.
##
Tested & Working properly:
- [x] `x86_64` Unable  to launch in hatch m138 container
- [x] `armv7l` Unable to launch in strongbad m138 container
```bash
CREW_REPO=https://github.com/uberhacker/chromebrew.git CREW_BRANCH=add-z_library-package crew update \
&& yes | crew upgrade
```


packages/z_library.rb
```diff
@@ -0,0 +1,35 @@
+require 'package'
+
+class Z_library < Package
+  description 'Z-Library launcher for desktop'
+  homepage 'https://z-lib.fm/z-access#desktop_app_tab'
+  version '2.4.3'
+  license 'Apache-2.0, BSD, MIT, others'
+  compatibility 'aarch64 armv7l x86_64'
+  min_glibc '2.29'
+  source_url 'https://s3proxy.cdn-zlib.sk/te_public_files/soft/linux/zlibrary-setup-latest.tar.gz'
```
supechicken, line 10: Do we have a static (version fixed) URL for this? This link will always point to the latest release

uberhacker: I wasn't able to locate it.  We can update if somebody complains.

supechicken: Can we set `source_sha256` to `SKIP` in this case so that it won't break when a new version comes out?

uberhacker: The problem with that is it won't bump the package version.

PR #12579, opened 2025-08-21
crew: Do `strip_dir` before calculating file size 

supechicken:
Otherwise the calculated size would be inaccurate, as binaries will be stripped after calculating the size.

```bash
CREW_REPO=https://github.com/supechicken/chromebrew.git CREW_BRANCH=fixup_filelist crew update \
&& yes | crew upgrade
```

### DIFF
--- a/bin/crew
+++ b/bin/crew
@@ -868,6 +868,8 @@ def prepare_package(destdir)
     # Allow postbuild to override the filelist contents
     @pkg.postbuild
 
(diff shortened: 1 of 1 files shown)


satmandu (approved):

uberhacker:
Forgot to bump the version in const.rb but we have had so many changes lately it probably doesn't make a difference.

